PR #773, opened 2016-11-07
Count bids with function on adapter if present

protonate:
## Type of change
- [x] Other

## Description of change
We count bids expected so we know when to invoke callbacks for a placement and for the overall request. To support counting bids in different ways this PR will use a function `countBids`, if available, in place of the default behavior.

### DIFF
--- a/src/adapters/indexExchange.js
+++ b/src/adapters/indexExchange.js
@@ -530,8 +530,14 @@ var IndexExchangeAdapter = function IndexExchangeAdapter() {
   }
 
   return {
-    callBids: _callBids
+    callBids: _callBids,
+    countBids: _countBids
   };
 };
+
+function _countBids(bid) {
+  const bidCount = bid.sizes
+  return bid && bid.sizes && bid.sizes.length || null;
+}
 
 module.exports = IndexExchangeAdapter;
--- a/src/adapters/indexExchange.js
+++ b/src/adapters/indexExchange.js
@@ -536,7 +536,6 @@ var IndexExchangeAdapter = function IndexExchangeAdapter() {
 };
 
 function _countBids(bid) {
-  const bidCount = bid.sizes
   return bid && bid.sizes && bid.sizes.length || null;
 }
 
--- a/src/bidmanager.js
+++ b/src/bidmanager.js
@@ -1,4 +1,4 @@
-import { uniques, flatten } from './utils';
+import { uniques, flatten, add } from './utils';
 import {getPriceBucketString} from './cpmBucketManager';
 
 var CONSTANTS = require('./constants.json');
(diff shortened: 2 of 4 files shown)


src/adapters/indexExchange.js
```diff
@@ -530,8 +530,14 @@ var IndexExchangeAdapter = function IndexExchangeAdapter() {
...
 
   return {
-    callBids: _callBids
+    callBids: _callBids,
+    countBids: _countBids
   };
 };
 
+function _countBids(bid) {
+  const bidCount = bid.sizes
```
dugwood: Why do you have this unused variable?

protonate: cruft, will remove.

src/bidmanager.js
```diff
@@ -40,34 +41,41 @@ function getBidders(bid) {
   return bid.bidder;
 }
 
+function add(a, b) {
```
mkendall07: let's put into `utils`

src/bidmanager.js
```diff
@@ -40,34 +41,41 @@ function getBidders(bid) {
...
+  return a + b;
+}
+
+/**
+ * Bidders handle multiple requests and multiple sizes in different ways.
+ * A Bidder Adapter can export a `countBids` function if this is anything
```
mkendall07: should we call this "countExpectedBids" ?

mkendall07:
I restarted travis and it still is failing with ``` Uncaught TypeError: object is not a function
  at /home/travis/build/prebid/Prebid.js/test/spec/adapters/aardvark_spec.js:10281```

Seems random but maybe not...

protonate:
@mkendall07 review notes addressed

src/prebid.js
```diff
@@ -847,4 +847,12 @@ $$PREBID_GLOBAL$$.getHighestCpmBids = function (adUnitCode) {
   return getWinningBids(adUnitCode);
 };
 
+/**
+ * Returns the adaptermanager.bidderRegistry
+ * @returns {*}
+ */
+$$PREBID_GLOBAL$$.getBidderRegistry = function() {
```
mkendall07: Why is this added to the global? We shouldn't need to expose this. If we need a means to store objects let's put it into a separate file that does not have global access.

protonate:
The current IX adapter bid count [workaround](https://github.com/prebid/Prebid.js/blob/master/src/bidmanager.js#L49), while inelegant, is working. I've put this is on hold for the time being. We should determine the best way to either support custom bid counting or conforming to a standard way of doing this.

cc: @prebid/indexexchange @indexexchange

protonate:
Stale PR, closing, if there is further interest here please reopen.

PR #778, opened 2016-11-08
remove nyan cat reporter as it slows down tests considerably

### LINKED ISSUE #774: Remove Nyan Karma reporter as it slows down testing 4x
## Type of issue
Question, build process, performance

## Description
Running unit tests without Nyan Karma reporter takes 1/4 of the time then when the Nyan reporter is enabled. I would advise removing it. To me, it doesn't give much value either.

## Steps to reproduce
Run `$ gulp test`.
- With Nyan: ~ 2mins, 2.5 mins for my colleague
- Without Nyan: ~ 30s

## Platform details
OS X El Capitan, Node.js 5 and 6

protonate:
## Type of change
- [x] Build related changes

## Description of change
Removes test coverage Nyan Cat reporter due to performance issues.

## Other information
fixes #774

### DIFF
--- a/karma.conf.js
+++ b/karma.conf.js
@@ -64,7 +64,7 @@ module.exports = function (config) {
     // test results reporter to use
     // possible values: 'dots', 'progress'
     // available reporters: https://npmjs.org/browse/keyword/karma-reporter
-    reporters: CI_MODE ? ['junit', 'coverage'] : ['progress', 'html', 'nyan', 'coverage'],
+    reporters: CI_MODE ? ['junit', 'coverage'] : ['progress', 'html', 'coverage'],
 
     // junit reporter config
     junitReporter: {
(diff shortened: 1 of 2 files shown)


marian-r:
@protonate looks fine, thanks :)

yairraz:
You would probably want to remove the karma-nyan-reporter plugin from the karma.conf.js plugins section and from package.json dev dependencies

protonate:
@mkendall07 @yairraz PR updated

mkendall07:
This didn't actually speed up running `gulp test` or `gulp serve` for me on a mac but I don't see any need for it so I'm good.

marian-r:
@mkendall07 I tried it now, and it wasn't that much slower with Nyan reporter, but still about 10 seconds slower than without it. I'm not sure how I measured it last time. Maybe the difference is related to Chrome update? I don't see anything else being changed on my machine. Anyway, I found Nyan reporter quite useless and 10s speed up (25%) is still good :)

mkendall07:
adios kitty.

protonate:
I believe that incremental rebuilds on watched files is faster as well, haven't measured it.

Studnicky:
Github does not have sad enough emojis to express how I feel about this.

PR #779, opened 2016-11-08
Prebid config passed as amp-ad json attribute

protonate:
## Type of change
- [x] Feature

## Description of change
Pass Prebid config as a JSON attribute to an `<amp-ad>` component.


mkendall07:
This LGTM. Are we good to merge @protonate ?

PR #787, opened 2016-11-11
sendStandardTargeting must be explicitly false

protonate:
## Type of change
- [x] Bugfix

## Description of change
sendStandardTargeting property filters out bids if value is `false` explicitly, but if 'undefined' filter should return key. Solution here is to not use the "double bang" `!!` operator to coerce the boolean.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -5,6 +5,7 @@ build
 # Test Files
 test/app
 gpt.html
+gpt-mlane.html
 gpt-each-bidder3.html
 
 # Selenium files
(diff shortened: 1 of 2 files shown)


protonate:
@mkendall07 @matthewlane for review

@snapwich please check this out for your use cases -- the property should be left "undefined" if it is so.

snapwich:
This looks like a correct change :+1:

mkendall07:
LGTM.

PR #788, opened 2016-11-11
wrap getTopWindowUrl in try/catch for iframed cases

protonate:
## Type of change
- [x] Bugfix

## Description of change
If Prebid is running in a x-domain iframe a call to `utils.getTopWindowUrl` will fail silently. Fix is to wrap the call in a try/catch at the adapter level.

## Other information
This will fix the jsfiddle.net examples on Prebid.org once a build is pushed to `not-for-prod` on CDN.

### DIFF
--- a/src/adapters/appnexus.js
+++ b/src/adapters/appnexus.js
@@ -118,11 +118,7 @@ AppNexusAdapter = function AppNexusAdapter() {
 
     //append referrer
     if (referrer === '') {
-      try {
-        referrer = utils.getTopWindowUrl();
-      } catch (err) {
-        referrer = 'iframed';
-      }
+      referrer = utils.getTopWindowUrl();
     }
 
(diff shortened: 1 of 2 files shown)


protonate:
@matthewlane @mkendall07 for review

src/adapters/appnexus.js
```diff
@@ -118,7 +118,11 @@ AppNexusAdapter = function AppNexusAdapter() {
 
     //append referrer
     if (referrer === '') {
-      referrer = utils.getTopWindowUrl();
+      try {
+        referrer = utils.getTopWindowUrl();
```
mkendall07: I'm not sure everyone will realize this can throw. Is there a way to swallow the error in `utils.getTopeWindowUrl` ?

protonate:
@mkendall07 review notes addressed

mkendall07:
LGTM.

PR #807, opened 2016-11-16
support IE userLanguage Navigator property

protonate:
## Type of change
- [x] Bugfix

## Description of change
IE does not support `Navigator.language` so add a check for this and support `Navigator.userLanguage` in case of IE.


protonate:
@matthewlane @mkendall07 for review

PR #810, opened 2016-11-17
fixes for bugs in test suite

protonate:
## Type of change
- [x] Bugfix

### DIFF
--- a/src/ajax.js
+++ b/src/ajax.js
@@ -15,64 +15,63 @@ const XHR_DONE = 4;
  */
 
 export function ajax(url, callback, data, options = {}) {
+  try {
+    let x;
+    let useXDomainRequest = false;
+    let method = options.method || (data ? 'POST' : 'GET');
 
-  let x;
-  let useXDomainRequest =false;
-  let method = options.method || (data ? 'POST' : 'GET');
-  if(!window.XMLHttpRequest){
-    useXDomainRequest = true;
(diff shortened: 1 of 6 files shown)


src/ajax.js
```diff
@@ -15,64 +15,63 @@ const XHR_DONE = 4;
  */
 
 export function ajax(url, callback, data, options = {}) {
+  try {
```
mkendall07: 👍

src/ajax.js
```diff
@@ -15,64 +15,63 @@ const XHR_DONE = 4;
...
+    let x;
+    let useXDomainRequest = false;
+    let method = options.method || (data ? 'POST' : 'GET');
```
mkendall07: We should probably just drop this and let the caller specify explicitly. Ok for now tho.

src/ajax.js
```diff
@@ -15,64 +15,63 @@ const XHR_DONE = 4;
...
+          callback(x.responseText, x);
+        }
+      };
     }
-    if (options.preflight) {
-      x.setRequestHeader('X-Requested-With', 'XMLHttpRequest');
+
+    if (method === 'GET' && data) {
+      let urlInfo = parseURL(url);
+      Object.assign(urlInfo.search, data);
```
mkendall07: What's this doing exactly?

PR #811, opened 2016-11-17
Add new adapter districtmDMX 

stevealliance:
## Type of change

- [ x] New bidder adapter

## Description of change

added districtmDMX to adapters.json
added districtmDMX.js to folder src/adapters
added districtm_specs.js to test for unit testing 

```
{
  bidder: 'districtmDMX',
  params: {
        id:109800, // required
        floor: 1.10,  // not required
        revShare: 0.90  // not required
        currency: 'cad' // not required will default to 'usd'
  }
}
```
- [steve@districtm.ca ] official adapter submission

## Other information

test @97%, two files added one updated

### DIFF
--- a/src/adapters/districtmDMX.js
+++ b/src/adapters/districtmDMX.js
@@ -3,8 +3,8 @@ var bidmanager = require('../bidmanager.js');
 var adLoader = require('../adloader');
 
 var DistrictmAdaptor = function districtmAdaptor(){
-   let districtmUrl = window.location.protocol + '//prebid.districtm.ca/lib.js';
-   this.callBids = params =>{
+  let districtmUrl = window.location.protocol + '//prebid.districtm.ca/lib.js';
+  this.callBids = params =>{
     if(!window.hb_dmx_res){
       adLoader.loadScript(districtmUrl,()=>{
         this.sendBids(params);
--- a/src/adapters/districtmDMX.js
+++ b/src/adapters/districtmDMX.js
@@ -1,0 +1,56 @@
+var bidfactory = require('../bidfactory.js');
+var bidmanager = require('../bidmanager.js');
+var adLoader = require('../adloader');
+
+var DistrictmAdaptor = function districtmAdaptor(){
+   let districtmUrl = window.location.protocol + '//prebid.districtm.ca/lib.js';
+   this.callBids = params =>{
+    if(!window.hb_dmx_res){
+      adLoader.loadScript(districtmUrl,()=>{
+        this.sendBids(params);
+      });
+    }else{
+      this.sendBids(params);
+    }
+    return params;
+  };
+
+
+  this.handlerRes = function(response, bidObject){
+    let bid;
+    if(parseFloat(response.result.cpm) > 0){
+      bid = bidfactory.createBid(1);
+      bid.bidderCode = bidObject.bidder;
+      bid.cpm = response.result.cpm;
+      bid.width = response.result.width;
+      bid.height = response.result.height;
+      bid.ad = response.result.banner;
+      bidmanager.addBidResponse(bidObject.placementCode, bid);
+    }else{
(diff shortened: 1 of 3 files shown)


test/spec/adapters/districtm_spec.js
```diff
@@ -0,0 +1,293 @@
...
+       let districtm;
+
+        beforeEach(()=>{
+            districtm = new Adaptor();
+        })
+
+        it('it\'s now time to play with the response ...', ()=>{
+            let result = districtm.handlerRes(PREBID_RESPONSE(), PREBID_PARAMS);
+            _each(result, function(k, v){
+                console.log(11, `${k} value is ${v}`);
```
matthewlane: In the `test` directory we try to avoid logging to the console to avoid having non-build related activity displayed during the build process. Please convert this to an assertion if it makes sense or remove the line

src/adapters/districtmDMX.js
```diff
@@ -0,0 +1,56 @@
+var bidfactory = require('../bidfactory.js');
+var bidmanager = require('../bidmanager.js');
+var adLoader = require('../adloader');
+
+var DistrictmAdaptor = function districtmAdaptor(){
+   let districtmUrl = window.location.protocol + '//prebid.districtm.ca/lib.js';
+   this.callBids = params =>{
```
matthewlane: The spacing on this line 6 and this line is causing a jscs error: `Expected indentation of 2 characters`. Please use two spaces on these lines rather than 3, or alternatively, place `// jscs:disable` at the top of this file to avoid these errors from displaying during the build process

stevealliance:
I just make some change please review

matthewlane:
Thanks for the update. It looks like this adapter is making a call to the same /jpt endpoint as the AppNexus adapter, any reason to not alias the AppNexus adapter instead?

stevealliance:
Yes, we are adding addtional features such as the currency parameter which allows us to send the bid in foreign currency in the ad server. We take care of updating currencies on a daily basis on behalf of the publisher

mkendall07:
@stevealliance 
So you are taking the bid returned by the Bidder and applying a currency conversion to it? How does it compete in the auction with other bidders of different currencies? Just curious how this works. Thanks

stevealliance:
If the publisher has an ad server setup in Canadian as an example, we apply convertion to it.
This is in the case that other end points would also be able to convert, and/or are applying a bid multiplier to reflect the same currency.

Many publishers are having Canadian ad servers, we tought this feature would be an option to simplify the process for them.

They have to be aware that this have to be used in a context of comparing apples to apples.

A little bit like gross vs net.

Thanks

internetchris:
@stevealliance/ @mkendall07  - I'm upgrading prebid from 0.24. to 1.10 and don't see an adapter listed in the modules folder for `districtmDMX`. Subsequently, I'm getting this error now in my prebid output: `Adapter trying to be called which does not exist: districtmDMX adaptermanager.callBids`

Can you advise if this adapter is still supported and the proper module I should be including in my build. FWIW, I did compile prebid with all of the adapters and still received the same message.

alandinglei:
Can someone answer @internetchris's question? And I have the same issue right now. Thanks!

lizardslair:
@stevealliance  Any news on if there will be a DistrictmDMX module compatable with > 1.10? We are currently performing an (overdue) upgrade from PB 0.21 and while DistrictmDMX is still listed in the official list of bidders, there is no districtmDMXBidAdapter.js file under /modules. Are you the point person for this update or is the community of devs supposed to chime in? We have several properties that use DistrictM and don't want to push our upgrade live until we get them working again. @matthewlane

stevealliance:
@lizardslair I have a PR in review for next release https://github.com/prebid/Prebid.js/pull/2765
You can also use our AppNexus alias, but please confirm with you account manager before doing the switch.

Thanks

PR #826, opened 2016-11-25
Ix endpoint upgrade

indexexchange:
## Type of change

## Description of change
*    Changing IX end point from headertag to cygnus
*    Deprecating unused variable – `index_slots`.
*    Additional IX custom targets and render functions.
*    sandbox mode now checks parent frames for query string parameter


wrosenbergatappnexus:
Is it normal behavior for Index to return multiple bids per slot? 

Additionally, I am seeing a situation where Index returns multiple bids per slot, and the higher bid is submitted with hb_pb etc. keys, and the lower bid is passed in through the hb_pb_indexexchange key. Thus, the lower bid is sent to the DFP auction.

I saw the behavior happening on this site: http://bit.ly/2nt0A9Q; prebid.js: 0.19.0; Google Chrome 56.0.2924.87

protonate:
It is expected in cases for multi-size ad units as IX makes a request per size. The highest IX bid should always be set as auction winner keys when IX wins. As for the bid landscape keys, I'm not sure that those keys would always be the second lower bid.

wrosenbergatappnexus:
Ok thanks for confirming they send multiple bids. 

The client is using "Send All Bids". I'd expect if an IX bid wins, it should be inserted for auction winner keys _and_ bid landscape keys, since the bid landscape keys are targeted by Index specific DFP line items, but that didn't seem to always be the case. 
screen shot 2017-03-06 at 1 15 12 pm

PR #828, opened 2016-11-27
Truncate bids requested on clearPlacements (#825)

### LINKED ISSUE #825: adUnit code based refresh times out without setting targets
## Type of issue
Bug

## Description
Prebid key-value targets are always empty on the ad-unit code based refreshes.

## Steps to reproduce
* have multiple slots
* refresh just a single slot, e.g. 
```
  pbjs.requestBids({
    adUnitCodes: ['div-gpt-ad-1438287399331-0'],
    bidsBackHandler: function() {
      pbjs.setTargetingForGPTAsync(['div-gpt-ad-1438287399331-0']);
      googletag.pubads().refresh([rightSlot]);
    }
  });
```

### Expected results

Targets on adUnit code based refresh:

```
> hb_bidder: indexExchange
> hb_adid: 1548157f46ea2e3
> hb_pb: 8.00
> hb_size: 300x250
```

### Actual results

Targets on adUnit code based refresh:

```
> hb_bidder: 
> hb_adid: 
> hb_pb: 
> hb_size: 
```
## Platform details
Versions 0.15.0, 0.15.1, 0.15.2 are all affected; v0.14.0 is not affected.

## Other information

`bid.requestTimestamp` is not necessarily the timestamp of the slot being refreshed.  Because the previous request's timestamp is older than request timestamp of the refresh, the auction always times out as soon as the first response is processed and therefore none of the bids participate in auctions:

```
exports.addBidResponse = function (adUnitCode, bid) {
  if (bid) {
    ...
 
    bid.timeToRespond = bid.responseTimestamp - bid.requestTimestamp;
 
    if (bid.timeToRespond > pbjs.cbTimeout + pbjs.timeoutBuffer) {
      var timedOut = true;
 
      exports.executeCallback(timedOut);
    }
    ...
    pbjs._bidsReceived.push(bid);
```

protonate:
## Type of change
- [x] Bugfix

## Description of change
The `clearPlacements` function is used to clear prior bids for an ad slot during a request bids for that slot while retaining bids for slots not refreshed. We do this to avoid removing bids that might be expected for any in-flight ad server call.

However we were also retaining `pbjs._bidsRequested` in certain cases, and this caused bugs where the new request did not get updated properties, so the timeout fired early and the targeting was not set. This issue occurs when calling `pbjs.requestBids` twice or more for the same ad slot without a full page refresh.

This PR simply truncates `pbjs._bidsRequested` to an empty array in `clearPlacements`, as bid request data is not needed for any in-flight ad server calls. This leaves the `pbjs._bidsReceived` logic unchanged.

## Other information
fixes #825


internetchris:
@protonate - just checking in on this PR to see when it might get reviewed and merged. Thanks!

mkendall07:
@internetchris 
We are working on unit tests and will commit shortly. After that it will be included in 0.16.0 release. Thanks

internetchris:
Sounds good! I love the sound of unit tests, always worth it.

protonate:
@mkendall07 a test is supplied using the new `faker` approach for mock data.

test/spec/integration/bug-825-truncate-bidsRequested_spec.js
```diff
@@ -0,0 +1,68 @@
...
+
+      resetPrebid();
+
+      adapters.forEach(adapter => {
+        adaptermanager.bidderRegistry[[adapter.bidder]] = { callBids: adapter.callBids };
+      });
+    });
+
+    after(() => {
+      window.$$PREBID_GLOBAL$$ = pbjsBackup;
```
protonate: This restores the context that the current set of unit tests is expecting.

PR #832, opened 2016-11-28
Allow changing of 'ga' global variable

dantarion:
## Type of change

- [x] Bugfix

## Description of change
The documentation uses this snippet of code to instruct people how to set up the Google Analytics adapter
```
    pbjs.enableAnalytics({
        provider: 'ga',
        options: {
            global: 'ga', // <string> name of GA global. Default is 'ga'
            enableDistribution: false,
        }
    });
```

However, nothing actually references `options.global`. `src/adapters/analytics/ga.js` sets `_gaGlobal` to `ga` always.

Am I missing something here? Is this a bug, or is the documentation wrong?


matthewlane:
Good catch, thanks @dantarion

PR #837, opened 2016-11-29
Detect browser width instead of the screen width

agabel:
## Type of change
- [X] Bugfix

## Description of change
Currently the size mapping is detecting the user's screen width instead of the browser. This change will use the browser width to determine which size mapping should be used.

### DIFF
--- a/src/sizeMapping.js
+++ b/src/sizeMapping.js
@@ -44,8 +44,16 @@ function isSizeMappingValid(sizeMapping) {
 
 function getScreenWidth(win) {
   var w = win || _win || window;
-  if(w.screen && w.screen.width) {
-    return w.screen.width;
+  var d = w.document;
+
+  if (w.innerWidth) {
+    return w.innerWidth;
+  }
+  else if (d.body.clientWidth) {
+    return d.body.clientWidth;
+  }
+  else if (d.documentElement.clientWidth) {
+    return d.documentElement.clientWidth;
   }
   return 0;
 }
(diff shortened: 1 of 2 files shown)


mkendall07:
It looks like this was the original code I had written. We noticed that this wasn't working with device emulation on chrome. I'll revisit this.

mkendall07:
Thanks for the PR. The problem here is that I'm confused with how `innerWidth` is calculated on mobile. For example, visit http://client-test.devnxs.net/demos/display.html from your phone (fullscreen browser) and see the differences in measurement.

agabel:
I see what you mean. I'll take another look into this and see if there is a better way to handle it.

mkendall07:
It seems to be related to the mobile "visual viewport" : http://www.quirksmode.org/mobile/viewports2.html

I haven't digested all of this yet, but we might need to just pick one and go with it. I'm not sure which way is more "standard".

agabel:
I have been looking into this further and it appears to me that the current PR should be the preferred way to detect viewport size. The [innerWidth method](https://www.w3.org/TR/cssom-view-1/#dom-window-innerwidth) will return the browser viewport. A fallback to the [clientWidth method](https://www.w3.org/TR/cssom-view-1/#dom-element-clientwidth) is acceptable to detect the width of the document itself. There are cases where the document could expand past the browser's actual viewport. The viewport meta tag should be used to set the viewport width to match the device width. The [http://client-test.devnxs.net/demos/display.html](http://client-test.devnxs.net/demos/display.html) site does not set a viewport meta tag so it will use the browsers default layout width. In Safari on iPhone that is 980px, but other browsers will vary. The innerWidth/clientWidth method would give us this value which would still be technically correct in regard to the page width.

mkendall07:
@agabel 
Ok thanks, I tend to agree with you. This LGTM.

PR #845, opened 2016-12-01
Add usersyncing to AppNexus adapters

matthewlane:
## Description of change
Adds usersyncing to appnexus and appnexusAst adapters after endpoint responses come back. Does not reload iframe on subsequent requests.


src/adapters/appnexus.js
```diff
@@ -201,6 +202,14 @@ AppNexusAdapter = function AppNexusAdapter() {
         bidmanager.addBidResponse(placementCode, bid);
       }
 
+      if (!usersync) {
+        var iframe = utils.createInvisibleIframe();
+        iframe.src = '//acdn.adnxs.com/ib/static/usersync/v3/async_usersync.html';
+        document.body.appendChild(iframe);
```
mkendall07: there's probably a small chance that `document.body` is undefined, so wrap in try/catch.

mkendall07:
LGTM

Issue #797, opened 2016-11-15
Post-bid issues with event emit order (BID_WON before AUCTION_END)


mpereira96:
## Type of issue
Question

## Description
It seems in a post-bid scenario:

![image](https://cloud.githubusercontent.com/assets/16920897/20315301/4fb5a044-ab55-11e6-92c5-fdf5843a6c00.png)

The standard is to call `pbjs.renderAd` inside the `pbjs.requestBids` handler. The problem with this is that the `AUCTION_END` event is only dispatched after the `pbjs.requestBids` handler is executed so the `BID_WON`  event will be dispatched first from `pbjs.renderAd`. This can lead to some issues if we are to assume the correct order is to have `AUCTION_END` before `BID_WON`.

Potentially there could be a better way to implement postbid on the frontend and avoid this.

Is this something you were aware? We would like to know your thoughts. We are happy to contribute to fix this issue.

mkendall07:
It's an interesting callout, not really intentional, it's just the way it works with postbid. Is this causing any issues? I suppose we could move `AUCTION_END` up before executing the callback, not sure what side effects (if any) that would cause.

marian-r:
@mkendall07 moving `AUCTION_END` up before executing the callback would be a perfect solution for us. We are happy to contribute a PR. Are we good to go?

mkendall07:
@matthewlane @protonate any objections?

protonate:
Makes sense to me.

mkendall07:
@marian-r 
Feel free to submit a PR. Thanks

marian-r:
Thank you. It might take a while as it's lower in our priorities, but me and @mpereira96 wanted to know your opinions first.

protonate:
AUCTION_END is intended as the final event of the auction, just prior to starting the next, if there is one. However BID_WON events are not tied to the AUCTION_END event in that the `renderAd` event is called in the response from the ad server, and these could begin before and continue after AUCTION_END. Closing and see comment on #852 about possibly adding another event.

PR #852, opened 2016-12-05
Make sure AUCTION_END event is dispatched before BID_WON event in a post-bid scenario.

mpereira96:
## Type of change

- [x] Bugfix
- [x] Refactoring

## Description of change
Updated to have the AUCTION_END event dispatched before BID_WON event in a post-bid scenario where `pbjs.renderAd` is called inside the `pbjs.requestBids` handler.

## Other information
https://github.com/prebid/Prebid.js/issues/797

### DIFF
--- a/src/bidmanager.js
+++ b/src/bidmanager.js
@@ -257,13 +257,13 @@ exports.executeCallback = function (timedOut) {
 
   //execute one time callback
   if (externalCallbacks.oneTime) {
+    $$PREBID_GLOBAL$$.clearAuction();
     try {
       processCallbacks([externalCallbacks.oneTime]);
     }
     finally {
       externalCallbacks.oneTime = null;
       externalCallbacks.timer = false;
-      $$PREBID_GLOBAL$$.clearAuction();
     }
   }
 };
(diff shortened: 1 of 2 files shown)


src/bidmanager.js
```diff
@@ -257,13 +257,13 @@ exports.executeCallback = function (timedOut) {
 
   //execute one time callback
   if (externalCallbacks.oneTime) {
+    $$PREBID_GLOBAL$$.clearAuction();
```
mkendall07: We may have some timing issues with moving this up. @protonate to review

mpereira96:
Hey guys any update on this?

protonate:
Actually AUCTION_END is intended to fire after BID_WON events. I apologize for not catching this sooner. As it is the `clearAuction()` function will initiate a new "auction" if there is a pending request in `bidsRequestQueue`, so this change would introduce timing issues as @mkendall07 mentioned.

Perhaps what is wanted is an AUCTION_CLOSED event which could be emitted at the top of `bidmanager.executeCallback()`:
https://github.com/prebid/Prebid.js/blob/master/src/bidmanager.js#L239

protonate:
See https://github.com/prebid/Prebid.js/pull/884 for possible solution.

PR #853, opened 2016-12-05
Smart AdServer adapter

Mordhak:
## Type of change
- [x] New bidder adapter

## Description of change
This is the official Smart AdServer adapter for header bidding using prebid library.

- test parameters for validating bids : not yet available
- contact email of the adapter’s maintainer : prebid@smartadserver.com

## Other information

Add Smart AdServer adapter with tests


Mordhak:
Hi,

We can't reproduce this issue in our side. All tests, inlcuding smartadserver_spec.js, are green, and we don't have the "undefined is not a function" error.

Besides, the error line number does not match any method (line 90 refers to an empty line in smartadserver_spec.js) so we can't know exactly which function is concerned.

Could you help us to reproduce this issue locally or tell us which method is concerned ?

Regards,

test/spec/adapters/smartadserver_spec.js
```diff
@@ -0,0 +1,117 @@
...
+  };
+
+  it("set url parameters", function () {
+    var stubLoadScript = sinon.stub(adLoader, "loadScript");
+
+    adapter().callBids(DEFAULT_PARAMS);
+
+    var smartCallback;
+    for (var k in $$PREBID_GLOBAL$$) {
+      if (k.startsWith("sas_")) {
```
matthewlane: Here's the line causing the failure. [This page](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/startsWith) has a polyfill that could be used directly or adapted to make this work cross-browser

matthewlane:
Great, tests are passing now. Do you have test parameters available for validating bids?

Mordhak:
Hi Matthew,

Here are the params for bid validation :

```javascript
{
  sizes: [[320, 250]],
  bids: [{
    bidder: "smartadserver",
    params: {
      siteId: 132432,
      pageId: 742455,
      formatId: 24977,
      domain: 'http://diff.smartadserver.com'
    }
  }]
}
```

Tell me if you need something else.

Regards,

aneuway2:
@Mordhak what params are required to be provided in your adaptor? Thanks!

Mordhak:
Hi @aneuway2,

The required parameters are : domain, siteId, pageId, formatId.
only "target" is optional.

Regards,

PR #861, opened 2016-12-06
Catch cross-origin DOMException

matthewlane:
## Type of change
- [x] Bugfix

## Description of change
Accessing `global.top` from the GumGum adapter is causing `Uncaught DOMException: Blocked a frame with origin "<url>" from accessing a cross-origin frame.` in setups such as the prebid.org [example pages](http://prebid.org/dev-docs/examples/basic-example.html). This change catches that exception.

## Other information
cc @davidmh

### DIFF
--- a/src/adapters/gumgum.js
+++ b/src/adapters/gumgum.js
@@ -10,8 +10,16 @@ const GumgumAdapter = function GumgumAdapter() {
 
   const bidEndpoint = `https://g2.gumgum.com/hbid/imp`;
 
-  const WINDOW = global.top;
-  const SCREEN = WINDOW.screen;
+  let WINDOW;
+  let SCREEN;
+
+  try {
+    WINDOW = global.top;
+    SCREEN = WINDOW.screen;
+  } catch (error) {
+    utils.logError(error);
+    return;
+  }
 
   function _callBids({ bids }) {
     const browserParams = {


src/adapters/gumgum.js
```diff
@@ -10,8 +10,16 @@ const GumgumAdapter = function GumgumAdapter() {
 
   const bidEndpoint = `https://g2.gumgum.com/hbid/imp`;
 
-  const WINDOW = global.top;
-  const SCREEN = WINDOW.screen;
+  let WINDOW;
```
mkendall07: Why caps? Probably should make this more consistent with our style.

src/adapters/gumgum.js
```diff
@@ -10,8 +10,16 @@ const GumgumAdapter = function GumgumAdapter() {
...
+  let SCREEN;
+
+  try {
+    WINDOW = global.top;
+    SCREEN = WINDOW.screen;
+  } catch (error) {
+    utils.logError(error);
+    return;
```
mkendall07: Does the adapter require access to `window.top` or can it continue if it fails to get access? Maybe a question for @davidmh

davidmh: We do need to access `window.top`, since the service was built with the idea of having the top being the host of the script.

Especially with the In-Screen service which uses the placement a 1x1 pixel, because our In-Screen uses fluid designs across the window width. If we do not have access to `top`, the In-Screen should not run.

@mkendall07 I can work on this if you like.

src/adapters/gumgum.js
davidmh: To keep a clear reference to the variable in a local scope. But I'm open to suggestions. What would be the Prebid approach to this?

src/adapters/gumgum.js
mkendall07: @davidmh 
We'd like to merge this code for now to fix our jsfiddle examples for the 0.16.x release. (see here http://jsfiddle.net/prebid/bhn3xk2j/4/? for example). 

Basically, because JsFiddle sandboxes the execution environment, access to window.top is blocked. Typically on a publisher page this is *not* the case but it causes our examples to stop working. The above code catches this error and will not run GumGum adapter in that case. 

Feel free to work on a solution that would support an sandboxed environment though for the future.

src/adapters/gumgum.js
mkendall07: Well we normally reserve caps for constants, anything else should just be camelCase. Also as an FYI we have some helper functions that safely attempt tot access window.top.* variables that might be helpful: https://github.com/prebid/Prebid.js/blob/master/src/utils.js#L193

mkendall07:
Confirmed this resolves the cross-orgin error and ads are loading on JsFiddle pages again. Thanks

PR #865, opened 2016-12-08
Add Sharethrough adapter

rizhang:
## Type of change

- [ ] Bugfix
- [ ] Feature
- [x] New bidder adapter
- [ ] Build related changes
- [ ] CI related changes
- [ ] Other

## Description of change

```
{
  bidder: 'sharethrough',
  params: {
      pkey: 'DfFKxpkRGPMS7A9f71CquBgZ'
  }
}
```
- [rzhang@sharethrough.com] official adapter submission

## Other information

Added Sharethrough Adapter


src/adapters/sharethrough.js
```diff
@@ -0,0 +1,149 @@
...
+      _handleInvalidBid(bidObj);
+    }
+  };
+
+  function _handleInvalidBid(bidObj) {
+    const bid = bidfactory.createBid(2, bidObj);
+    bidmanager.addBidResponse(bidObj.placementCode, bid);
+  }
+
+  str.bidWon = function() {
```
mkendall07: We currently do not allow bidders to listen to other bids. If a bidder wants to do this, they should add an analytics adapter. This is a publisher opt in only process. https://github.com/prebid/Prebid.js/blob/master/src/adapters/analytics/example2.js

src/adapters/sharethrough.js
```diff
@@ -0,0 +1,149 @@
...
+var bidfactory = require('../bidfactory.js');
+
+const STR_BIDDER_CODE = "sharethrough";
+const STR_VERSION = "0.1.0";
+
+var SharethroughAdapter = function SharethroughAdapter() {
+
+  const str = {};
+  str.STR_BTLR_HOST = document.location.protocol + "//btlr.sharethrough.com";
+  str.STR_TEST_HOST =  document.location.protocol + "//btlr-prebid-test.sharethrough.com";
```
mkendall07: Using a different endpoint for test is not ideal as real issues are discovered in prod. Is there a way to not use a test endpoint here?

src/adapters/sharethrough.js
```diff
@@ -0,0 +1,149 @@
...
+    $$PREBID_GLOBAL$$.onEvent('bidWon', str.bidWon);
+
+    addEventListener("message", _receiveMessage, false);
+    
+    // cycle through bids
+    for (let i = 0; i < bids.length; i += 1) {
+      const bidRequest = bids[i];
+      str.placementCodeSet[bidRequest.placementCode] = bidRequest;
+      const scriptUrl = _buildSharethroughCall(bidRequest);
+      str.loadIFrame(scriptUrl);
```
mkendall07: Don't load JSONP inside an iframe. Use `ajax.js` or `adLoader.loadScript()` instead

src/adapters/sharethrough.js
```diff
@@ -0,0 +1,149 @@
...
+                </div>
+                <script>var ${windowLocation} = ${bidJsonString}</script>
+                <script src="//native.sharethrough.com/assets/sfp-set-targeting.js"></script>
+                <script type='text/javascript'>
+                (function() {
+                    var sfp_js = document.createElement('script');
+                    sfp_js.src = "//native.sharethrough.com/assets/sfp.js";
+                    sfp_js.type = 'text/javascript';
+                    sfp_js.charset = 'utf-8';
+                    window.top.document.getElementsByTagName('body')[0].appendChild(sfp_js);
```
mkendall07: `window.top` isn't safe to call directly, it should be in a try/catch. Additionally it should be avoided if possible to support SafeFrame in the future.

src/adapters/sharethrough.js
rizhang: We have to use a test endpoint here to guarantee that a creative is always return from our exchange. If we used the real endpoint there's no way to guarantee that a DSP will always bid on the impression request.

src/adapters/sharethrough.js
rizhang: We need to put the url inside an iframe so we can drop tracking cookies on publisher's page. This method work for iOS Safari since they block 3rd party cookies.

src/adapters/sharethrough.js
```diff
@@ -0,0 +1,132 @@
...
+  const str = {};
+  str.STR_BTLR_HOST = document.location.protocol + "//btlr.sharethrough.com";
+  str.STR_TEST_HOST =  document.location.protocol + "//btlr-prebid-test.sharethrough.com";
+  str.STR_BEACON_HOST = document.location.protocol + "//b.sharethrough.com/butler?";
+  str.placementCodeSet = {};
+
+  function _callBids(params) {
+    const bids = params.bids;
+
+    $$PREBID_GLOBAL$$.onEvent('bidWon', str.bidWon);
```
mkendall07: This line can be removed now that you added your analytics adapter.

src/adapters/sharethrough.js
```diff
@@ -0,0 +1,132 @@
...
+    iframe.src = url;
+    iframe.style.cssText = 'display:none;';
+
+    document.body.appendChild(iframe);
+  };
+
+  function _receiveMessage(event) {
+    if(event.origin === str.STR_BTLR_HOST || event.origin === str.STR_TEST_HOST) {
+      try {
+        $$PREBID_GLOBAL$$.strcallback(JSON.parse(event.data).response);
```
mkendall07: this doesn't need to a public function because it's scoped within the adapter itself. Only use a public function if you use JSONP.

src/adapters/sharethrough.js
mkendall07: Ok

src/adapters/sharethrough.js
mkendall07: Usually other bidders setup a test creative /placement on the live exchange. Is this possible?

src/adapters/analytics/sharethrough_analytics.js
```diff
@@ -0,0 +1,63 @@
+import adapter from 'AnalyticsAdapter';
+const utils = require('../../utils');
+
+const url = 'https://httpbin.org/post';
```
mkendall07: You probably don't want to be sending data to `httpbin.org`

package.json
```diff
@@ -17,7 +17,7 @@
     "prebid"
   ],
   "globalVarName": "pbjs",
-  "analytics": [],
+  "analytics": ["sharethrough_analytics"],
```
matthewlane: Analytics is explicitly turned on by the publisher so this array should be blank by default

rizhang:
I updated the test pkey paramter

mkendall07:
@matthewlane just need final review for bids back on this one.

Issue #858, opened 2016-12-06
Emit event when targeting is sent to DFP (setTargetingForGPTAsync)


itayo155:
## Type of issue
feature request

## Description
Currently, there's no reliable way to know when an auction has actually ended, i.e. the point after which no bids would participate in the auction.
The existing AUCTION_END event is, in many cases, unreliable for this use because most setups do not use prebid's timeout field for ending the auction - they use external setTimeout instead, and/or are calling setTargetingForGPTAsync directly.

The suggestion is to emit a new event ("SET_TARGETING_CALLED") when setTargetingForGPTAsync is called. This indicates that all bids have been collected and actually sent away.
Of course, there are some implementations in which this function is never called/ is not a reliable indicator, however it should work in the vast majority of cases.

This event would be useful for bidders, allowing them to optimize response times, which in turn would maximize bid-rate for publishers. 
Since the event would contain no sensitive information regarding the auction itself or its result, it should be ok for access by bidders.

## Other information
Related: 
https://github.com/prebid/Prebid.js/issues/494 - details the current conditions for emitting AUCTION_END

itayo155:
Following up - any comments here? would be happy to submit a PR if this makes sense.

protonate:
@itayo155 PRs are always welcome, thanks.

matthewlane:
Closing with #873 now in master

PR #873, opened 2016-12-13
Emit event when setTargetingForGPTAsync is called

itayo155:
## Type of change
- [x] Feature

## Description of change
Emit a new event ('setTargeting') when setTargetingForGPTAsync is called. This indicates that all bids have been collected and actually sent away. 
For reference regarding why this is needed and intended use, please see issue #858. 

## Other information
Original issue: https://github.com/prebid/Prebid.js/issues/858


itayo155:
Following up - any comments, or thoughts when it could be merged?

mkendall07:
@itayo155 
can you rebase off master to resolve conflicts?

itayo155:
Sure, done. Please let me know if there are any other comments.

jaiminpanchal27:
@itayo155 
I don't think we should add this "SET_TARGETING" event for 2 reasons
1) We will be adding another targeting function for Appnexus adserver and maybe in future some more. So we should not be emitting this event from all the other public functions.
2) It is very similar to AUCTION_END event. Hence if its unreliable, we should fix that instead of adding new similar event.

itayo155:
@jaiminpanchal27 
OK. If it is agreed that the AUCTION_END event should be emitted when bidding ends (i.e. when additional bids will not participate in current auction anymore), I suggest to close this PR and open a new one, for "making the AUCTION_END more accurate" or something like that.

However, given the way things are usually set-up by publishers, I see only two ways of implementing this change:
1. Hook the individual setTargeting functions (e.g. for DFP, AN, etc.), similarly to the way its done in the current PR. If there are multiple functions expected, this can be done by calling a single function which would emit the event.
2. Force a change in the way publishers interact with Prebid, e.g. by changing the API.

The reason is that setTargetging is usually being called directly from publisher's code, and more specifically - this happens when the auction times out at a time arbitrarily selected by publisher.
E.g. (adapted from a live implementation, comments are mine):
```
//start the auction
  pbjs.que.push(function() {
      pbjs.addAdUnits(adUnits);
      //...
      pbjs.requestBids({
      });

//set a timeout - arbitrarily selected by publisher
 setTimeout(function() {
      sendAdserverRequest();
  }, some_publisher_set_timeout);

//end the auction - note the setTargetingForGPTAsync is being called directly; and only at this point the auction actually ends
    //...      
          pbjs.que.push(function() {
              //auction ends now, and we'd like the AUCTION_END event to somehow be emitted
              //...
          });
      });
  }
```

Would be happy for additional ideas for how to implement this.
Out of the two options above, I personally find (1) preferable.

Would be happy to hear your thoughts.

mkendall07:
@itayo155 
Thanks for the PR. We'll review internally and get back to you.
Could you also send me an email directly? My email is in my profile. Thanks

jaiminpanchal27:
@itayo155 
Looks good. Merging the PR.

PR #881, opened 2016-12-15
remove hb_pb targeting key for deal bids with no cpm

protonate:
## Type of change
- [x] Feature

### DIFF
--- a/src/bidmanager.js
+++ b/src/bidmanager.js
@@ -129,19 +129,15 @@ exports.addBidResponse = function (adUnitCode, bid) {
 
     //if there is any key value pairs to map do here
     var keyValues = {};
-    if (bid.bidderCode && bid.cpm !== 0) {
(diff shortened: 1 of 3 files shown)


src/bidmanager.js
```diff
@@ -129,16 +129,18 @@ exports.addBidResponse = function (adUnitCode, bid) {
 
     //if there is any key value pairs to map do here
     var keyValues = {};
-    if (bid.bidderCode && bid.cpm !== 0) {
+    if (bid.bidderCode) {
```
mkendall07: Should we check for dealId or comments?

mkendall07: DealId or CPM*

protonate: 👍

protonate:
@mkendall07 notes addressed

mkendall07:
LGTM.

PR #882, opened 2016-12-16
GumGum adapter - include new parameter on the request

davidmh:
## Type of change

- [ ] Bugfix
- [ ] Feature
- [ ] New bidder adapter
- [ ] Build related changes
- [ ] CI related changes
- [x] Other

## Description of change
Including a new parameter for subsequent requests on the same page.

dmejorado@gumgum.com

## Other information
Changed case on a couple of variables.

### DIFF
--- a/src/adapters/gumgum.js
+++ b/src/adapters/gumgum.js
@@ -10,12 +10,13 @@ const GumgumAdapter = function GumgumAdapter() {
 
   const bidEndpoint = `https://g2.gumgum.com/hbid/imp`;
 
-  let WINDOW;
-  let SCREEN;
+  let topWindow;
+  let topScreen;
+  let pageViewId;
 
   try {
-    WINDOW = global.top;
-    SCREEN = WINDOW.screen;
+    topWindow = global.top;
+    topScreen = topWindow.screen;
   } catch (error) {
     utils.logError(error);
     return;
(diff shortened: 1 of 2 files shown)


davidmh:
Hi @matthewlane, could you help us review this update?

matthewlane:
@davidmh Gladly. I have these bidder params for testing GumGum
```
{
  bidder: 'gumgum',
  params: {
    inScreen: 'ggumtest'
  }
}
```
but currently getting back empty bids. Are there different params I should be using?

davidmh:
Thank you @matthewlane, that's expected behavior. Everything is fine as long as you get a healthy response.

mkendall07:
@davidmh We prefer having permanent test params that return a valid bid. Is that possible?

davidmh:
@mkendall07 we have a filter for unprocessed pages, so the endpoint won't return a bid until we have processed the page contents.

So if you're testing with a local domain the filter won't go away because the processor won't be able to reach it. We can bypass the filter by marking the URL as processed manually, can you share the full local URL you are using to test?

mkendall07:
we usually just test with localhost, but if that's not possible, prebid.org can be the host. If you need a full URL, (http://prebid.org/all_bidders_test.html).

davidmh:
Localhost works, we just need the full URL e.g. `http://localhost:9999/integrationExamples/gumgum.html`

davidmh:
Done, you should be able to test with `http://prebid.org/all_bidders_test.html`

matthewlane:
Can the `localhost:9999` address be accepted as well? I'm struggling to configure my test page as `prebid.org` without a port number

davidmh:
Of course, I can ask our team to add `http://localhost: 9999/all_bidders_test.html`

Would that work for you?

matthewlane:
I think it will, thank you!

davidmh:
Done, seems to be working on my test.

test

matthewlane:
That did the trick, thanks a lot

PR #885, opened 2016-12-16
Support SafeFrame / x-domain on prebid creatives

protonate:
## Type of change
- [x] Feature

## Description of change
A creative served in a cross domain iframe can request the ad from the auction winner via `postMessage`.


integrationExamples/gpt/safeframe/creative.html
```diff
@@ -0,0 +1,38 @@
...
+  var requestAdFromPrebid = function () {
+    var message = JSON.stringify({
+      message: 'Prebid creative requested: %%PATTERN:hb_adid%%',
+      adId: '%%PATTERN:hb_adid%%'
+    });
+    window.parent.postMessage(message, '*');
+  };
+
+  var listenAdFromPrebid = function () {
+    window.addEventListener("message", renderAd, false);
```
mkendall07: IE support

integrationExamples/gpt/safeframe/creative.html
```diff
@@ -0,0 +1,38 @@
...
+      }
+    }
+  };
+
+  var requestAdFromPrebid = function () {
+    var message = JSON.stringify({
+      message: 'Prebid creative requested: %%PATTERN:hb_adid%%',
+      adId: '%%PATTERN:hb_adid%%'
+    });
+    window.parent.postMessage(message, '*');
```
mkendall07: No way to avoid * here I guess?

src/prebid.js
```diff
@@ -289,6 +289,35 @@ function setRenderSize(doc, width, height) {
   }
 }
 
+function listenAdRequestFromCreative() {
+  addEventListener('message', sendAdToCreative, false);
```
mkendall07: missing window?

src/prebid.js
```diff
@@ -289,6 +289,35 @@ function setRenderSize(doc, width, height) {
...
+  }
+
+  if (data.adId) {
+    var adObject = $$PREBID_GLOBAL$$._bidsReceived.find(function (bid) {
+      return bid.adId === data.adId;
+    });
+
+    var ad = adObject.ad;
+    var adUrl = adObject.adUrl;
+    var message = JSON.stringify({
```
mkendall07: prefer a factory pattern for creating messages.

integrationExamples/gpt/x-domain/creative.js
```diff
@@ -0,0 +1,56 @@
...
+    }
+
+    if (adObject.ad || adObject.adUrl) {
+      var height = adObject.height;
+      var width = adObject.width;
+      var url = adObject.adUrl;
+      var ad = adObject.ad;
+      var doc = window.document;
+
+      // if (doc === document || adObject.mediaType === 'video') {
```
mkendall07: conversation piece?

integrationExamples/gpt/x-domain/creative.js
```diff
@@ -0,0 +1,56 @@
...
+      } else {
+        // utils.logError('Error trying to write ad. No ad for bid response id: ' + id);
+      }
+    }
+  }
+
+function requestAdFromPrebid() {
+  var message = JSON.stringify({
+    message: 'Prebid Request',
+    adId: window.prebidAdId
```
mkendall07: Where does `prebidAdId` come from?

integrationExamples/gpt/x-domain/creative.js
```diff
@@ -0,0 +1,56 @@
...
+  });
+  window.parent.postMessage(message, '*');
```
mkendall07: Should we add a marco for hostname? This could be set as a keyword by Prebid and then communication would be limited between those domains.

integrationExamples/gpt/x-domain/creative.js
protonate: Yep. This code is running in the x-domain iframe, so `doc === document`. Also `utils` module is not available so need to solve for a local `console.log`.

integrationExamples/gpt/x-domain/creative.js
protonate: The DFP code snippet needs to set the `adId` for the creative.js, and while developing using an external creative.js file this simply sticks it on the x-domain iframe window object.

integrationExamples/gpt/x-domain/creative.js
protonate: Yes, indeed, I'll make the change.

protonate:
@mkendall07 this is ready for further review.

integrationExamples/gpt/x-domain/creative.js
```diff
@@ -0,0 +1,53 @@
+// this script can be returned by an ad server delivering a cross domain iframe, into which the
+// creative will be rendered, e.g. DFP delivering a SafeFrame
+
+const publisherDomain = 'http://localhost:9999';
+const adServerDomain = 'http://tpc.googlesyndication.com';
```
mkendall07: Let's add a note about making sure to match the protocol. I assume that this becomes HTTPS on secure pages?

integrationExamples/gpt/x-domain/creative.js
```diff
@@ -0,0 +1,53 @@
+// this script can be returned by an ad server delivering a cross domain iframe, into which the
+// creative will be rendered, e.g. DFP delivering a SafeFrame
```
mkendall07: Can make a note about using a macro here instead of a hardcoded string. Many creatives will be delivered to multiple domains.

src/prebid.js
```diff
@@ -288,6 +291,58 @@ function setRenderSize(doc, width, height) {
   }
 }
 
+function listenMessagesFromCreative() {
```
mkendall07: Let's move this function into a separate file

protonate:
@mkendall07 review notes addressed.

StaymanHou:
Would the doc also be updated regarding the safeframe option?

[Here](http://prebid.org/adops/step-by-step.html#step-2-add-a-creative) it still says the safeframe box has to be unchecked. And the code snippet doesn't work for safeframe.

PR #892, opened 2016-12-23
Add pixel size to memeglobal adatper

AlexStanovsky:
## Type of change

- [x] Bugfix

## Description of change
Fix pixel size in win notification url


jaiminpanchal27:
Looks good

PR #916, opened 2017-01-11
Add StickyAdsTV Bidder adapter

guillaume-sticky:
## Type of change

- [ ] Bugfix
- [ ] Feature
- [ x] New bidder adapter
- [ ] Build related changes
- [ ] CI related changes
- [ ] Other

## Description of change

Add StickyAdsTV Bidder Adaptor

```
{
  bidder: 'stickyadstv',
  params: {
    zoneId : 277225  //only this parameter is mandatory
  }
}
```
- contact email of the adapter’s maintainer : gandouard@freewheel.tv

## Other information
Contact: gandouard@freewheel.tv

Test page: http://sandbox.stickyadstv.com/qa/prebid.html


mkendall07:
Please fix test failures.

guillaume-sticky:
Done by merging in master.

src/adapters/stickyadstv.js
```diff
@@ -0,0 +1,265 @@
...
+  }
+  
+  var formatInBannerHTML = function(bid,size){
+    var placementCode = bid.placementCode;
+    
+    var divHtml = "<div id=\"stickyadstv_prebid_target\"></div>";
+
+    var script = "<script type='text/javascript'>"+
+       
+    "var vast =  window.top.stickyadstv_cache[\""+placementCode+"\"];"+
```
protonate: Use of `window.top` will not work when running in a cross domain iframe, which is increasingly more common. Delivery of creatives into a SafeFrame, for example, or when integrating Prebid with an AMP Page.

src/adapters/stickyadstv.js
```diff
@@ -0,0 +1,265 @@
...
+    }
+    
+    return priceData;
+  }
+
+  function formatBidObject(valid, priceData, html, width, height){
+    var bidObject;
+    if(valid && priceData) {
+      // valid bid response
+      bidObject = bidfactory.createBid(1);
```
protonate: Pass in the bid requested here so the request/response pair can be matched.

src/adapters/stickyadstv.js
```diff
@@ -0,0 +1,265 @@
...
+      bidObject.cpm = priceData.price;
+      bidObject.currencyCode = priceData.currency;
+      bidObject.ad = html;
+      bidObject.width = width;
+      bidObject.height = height;
+      
+    }
+    else {
+      // invalid bid response
+      bidObject = bidfactory.createBid(2);
```
protonate: Pass in the bid requested here as well.

src/adapters/stickyadstv.js
guillaume-sticky: Indeed,  I'll remove it.

src/adapters/stickyadstv.js
guillaume-sticky: How? the documentation says to use  "var bidObject = bidfactory.createBid(1);"
see: http://prebid.org/dev-docs/bidder-adaptor.html

guillaume-sticky: OK,  seen an example in bidder adaptor AppNexus.js:
bidfactory.createBid(1, bidObj);

You should add it in the documentation, I didn't find an API reference for the bidfactory.
(did I miss it?)

protonate: Ah, thanks, created an issue on docs repo:
https://github.com/prebid/prebid.github.io/issues/168

protonate:
@guillaume-sticky please let us know the status on this, thanks.

guillaume-sticky:
I'm sorry I was quite busy those days.

I'll push the requested modifications soon.

src/adapters/stickyadstv.js
guillaume-sticky: Is there a way I can get the context where my script has been loaded before the bid?

When I use "adloader.loadScript" I load a full API that I would need to access again inside the "Ad" (html/javascript loaded if my bid wins) which is loaded inside an iframe.

That is what "window.top" was for.

guillaume-sticky:
Updated with requested modifications.

protonate:
This appears to be a video adapter. Please have a look at the docs on video support: http://prebid.org/dev-docs/show-video-with-a-dfp-video-tag.html

This PR can be updated to incorporate the video config as described.

guillaume-sticky:
Can you be more specific on what is needed here ? We don't want to use your video player example as we are already setup our player as we need it.

I guess I have to declare the supported medie type video ?
"supportedMediaTypes" : ["video"]

Is that it ?

protonate:
@guillaume-sticky yes and also set the `bid.mediaType` to `"video"`. This way the adapter will be registered as a video adapter and the bid will be handled as a video bid.

mkendall07:
@guillaume-sticky 
Please hold off on changes for now. You only need to designate bid.mediaType to "video" if the bid returns a VAST response. 

This format appears to be in banner video, is that correct?

guillaume-sticky:
Yes it is,  we never return a VAST response, we give the html ad in a bid object.

protonate:
@guillaume-sticky This is good as-is, thanks for the contribution!

guillaume-sticky:
Thank you for adding stickyadstv bidder adapter.

I wonder if we should worry about the 20$ CPM cap? As we could see some CPM over 20$ in our video platform.
And if we need to work around this, can we do it from inside the bidder adapter or is the "custom price bucket" this something the publisher needs to setup?

PR #921, opened 2017-01-12
rhythmone: reset auctionEnded and requestCompleted on new auction

kitwestneat:
## Type of change
- [x] Bugfix

## Description of change
The RhythmOne adapter tracks the auction end and request completion, but doesn't reset the tracking variables on new bid requests. This patch resets those variables so that new bids are not ignored.


jstocker76:
Original dev here.  Is it a known practice to invoke CallBids more than once per page view?

mkendall07:
@jstocker76 
Yes it's possible via `pbjs.requestBids()` function to be invoked multiple times.

jstocker76:
Aah, good catch.

PR #937, opened 2017-01-19
PubMatic to execute tracker pixel in iframe.

pm-harshad-mane:
## Type of change

- [x] Bugfix
- [ ] Feature
- [ ] New bidder adapter
- [ ] Build related changes
- [ ] CI related changes
- [ ] Other

## Description of change
PubMatic impression tracker pixel needs to be executed in iframe to avoid multiple times execution.
So we have added a new function createTrackPixelIframeHtml in utils and have used the same in PubMatic adapter.


pm-harshad-mane:
Hello @mkendall07 ,

Do you need anything from me for code-review ?

mkendall07:
@pm-harshad-mane 
Can you explain the need for the iframe vs image embed? I don't see the difference here.

src/utils.js
```diff
@@ -471,6 +471,19 @@ exports.createTrackPixelHtml = function (url) {
...
+ * Creates a snippet of Iframe HTML that retrieves the specified `url`
+ * @param  {string} url plain URL to be requested
+ * @return {string}     HTML snippet that contains the iframe src = set to `url`
+ */
+exports.createTrackPixelIframeHtml = function (url) {
+  if (!url) {
+    return '';
+  }
+
+  return '<iframe frameborder="0" allowtransparency="true" marginheight="0" marginwidth="0" width="0" hspace="0" vspace="0" height="0" style="height:0p;width:0p;display:none;" scrolling="no" src="' + encodeURI(url) + '"></iframe>';
```
mkendall07: Please use string template literals here: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Template_literals

pm-harshad-mane:
Hello @mkendall07 ,

Agree with you that there is no much difference in executing a pixel in an image tag or an iframe tag.
But in our past-experience we have observed some discrepancies in case when our tracker call is executed in image tag as image tag with zero byte content reloads in some cases.
So at PubMatic we recommend to execute our tracker call in an iframe tag than image tag.

mkendall07:
Thanks. LGTM.

pm-harshad-mane:
Thank you @mkendall07

PR #953, opened 2017-01-25
Pubgears Header Bidding Adapter

Walexander:
## Type of change

- [ ] Bugfix
- [ ] Feature
- [x] New bidder adapter
- [ ] Build related changes
- [ ] CI related changes
- [ ] Other

## Description of change

Adapter and specs for Pubgears header bidder.

```
{
  bidder: 'pubgears',
  adUnitCode: 'id123/header-bid-tag-0',
  params: {
    publisherName: "integration",
    pubZone: "i.testpub.com/multibidder"
  }
}
```
- walexander@mtheorydigital.com

### DIFF
--- a/src/adapters/pubgears.js
+++ b/src/adapters/pubgears.js
@@ -1,6 +1,7 @@
 var bidfactory = require('../bidfactory.js');
 var bidmanager = require('../bidmanager.js');
 var consts = require('../constants.json');
+var utils = require('../utils.js');
 var d = document;
 var SCRIPT = 'script';
 var PARAMS = 'params';
@@ -26,6 +27,8 @@ var PUBLISHER_ATTRIBUTE = 'pub';
 var FLAG_ATTRIBUTE = 'flag';
 
 var AD_UNIT_CODE = 'adUnitCode';
+var PLACEMENT_CODE = 'placementCode';
+var BID_ID = 'bidId'
 var PUBLISHER_PARAM = 'publisherName';
 var PUB_ZONE_PARAM = 'pubZone';
 
@@ -62,7 +65,7 @@ function PubGearsAdapter() {
     bids.forEach(function(bid) {
 
       var name = getSlotFromBidParam(bid);
-      pendingSlots[ name ] = bid[AD_UNIT_CODE] ;
+      pendingSlots[ name ] = bid;
     });
 
     proxy = proxy || getScript(SCRIPT_ID) || makeScript(slots, publisher, SCRIPT_ID, TAG_URL);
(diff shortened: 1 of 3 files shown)


protonate:
Is it possible to implement without using additional scripts?

protonate:
Try rebasing onto current master branch, or if you prefer it may be easier to start with a new branch cut from current master branch and reintroduce your changes there.

Walexander:
The rebase has been completed.  Apologies for the delay.

Walexander:
It would be difficult to implement this without loading our external library. The adapter is built around receiving events from the script element itself. The code that publishes these events is implemented in the external library.

src/adapters/pubgears.js
```diff
@@ -0,0 +1,194 @@
...
+
+var BIDDER_CODE = 'pubgears';
+
+var SCRIPT_ID = 'pg-header-tag';
+
+var ATTRIBUTE_PREFIX = 'data-bsm-';
+var SLOT_LIST_ATTRIBUTE = 'slot-list';
+var PUBLISHER_ATTRIBUTE = 'pub';
+var FLAG_ATTRIBUTE = 'flag';
+
```
mkendall07: please remove all these extra new lines.

src/adapters/pubgears.js
```diff
@@ -0,0 +1,194 @@
...
+      pendingSlots[ name ] = bid;
+    });
+
+    proxy = proxy || getScript(SCRIPT_ID) || makeScript(slots, publisher, SCRIPT_ID, TAG_URL);
+    if(!initialized)
+        registerEventListeners(proxy);
+    initialized = true;
+  }
+
+  function loadScript(script) {
```
mkendall07: Please use the standard `adLoader.loadScript` instead

src/adapters/pubgears.js
```diff
@@ -0,0 +1,194 @@
...
+
+    response[CPM] = price / 1e3 ;
+    response[WIDTH] = dims[0];
+    response[HEIGHT] = dims[1];
+    return response;
+  }
+
+  function getCreative(resource) {
+
+    var bookends = '%%';
```
mkendall07: You can use `utils.replaceTokenInString` for this functionality or a string template.

Walexander: Thanks for the suggestion. This has been done.

src/adapters/pubgears.js
Walexander: Done.

src/adapters/pubgears.js
Walexander: We use the `script` element to obtain the bid response data from our API. The ```adLoader.loadScript``` does not provide the element either via its return value or callback.

mkendall07: @Walexander can you give an example? That seems like a strange pattern to me.

Walexander: Our API uses DOM events instead of method calls. The adapter registers event listeners with the script element [here](https://github.com/MbidIO/Prebid.js/blob/ebc6db3272dd5124fbcddc36d5684db0c1ce21bb/src/adapters/pubgears.js#L116-L117)
 
[API Documentation](http://c.pubgears.com/eg/nucleus/api/index.html)

mkendall07:
@Walexander thanks for the updates. It looks like your adapter is adding 2 bid responses using the test params. See screenshot here: 
image

You should only add a `no_bid` in case there isn't any demand. Please update. 

Also - it doesn't appear that ` adUnitCode: 'id123/header-bid-tag-0',` does anything? What's it's purpose?

test/spec/adapters/pubgears_spec.js
```diff
@@ -215,7 +215,7 @@ describe('PubGearsAdapter', () => {
 			expect(bidmanager.addBidResponse.calledOnce).to.be.ok
 		})
 
-		it('should not completely fucking fail', () => {
```
mkendall07: lol

mkendall07:
LGTM

protonate:
A note here that the use of `script.dispatchEvent` in tests is failing in some version of IE with "Object doesn't support this action". This is limited to unit tests in IE.

Issue #954, opened 2017-01-25
Updates needed for integrationExamples/gpt/x-domain/creative.js


brondsem:
## Type of issue
Bug

## Description
* the `JSON.stringify` call has a dangling `adServerDomain` all by itself
* In IE, the ad doesn't render and console has this error:  `SEC7111: HTTPS security is compromised by (null)        https://tpc.googlesyndication.com/safeframe/1-0-5/html/container.html`
* In IE < 11, `const` raises a syntax error, `var` should be used instead.
* The `id` var isn't defined anywhere
* Per https://github.com/prebid/Prebid.js/issues/927#issuecomment-273507533 `adServerDomain` should check the current protocol (or just default to https)
* Instead of `%%PUBLISHER_DOMAIN%%`, `%%PATTERN:url%%` exists already and can be parsed to get the domain.
* would be nice to have `<script>....</script>` around the whole JS (and then save the file as .html) so it is a full example ready to paste directly in.

I can provide a merge request shortly for most of these items, but I don't know how to fix the IE HTTPS error.

brondsem:
http://joji.me/en-us/blog/how-to-resolve-sec7111-https-security-is-compromised-by-null-error-in-ie11 indicates its the `doc.write` calls that IE doesn't like.

brondsem:
Fixed with #955 and #971

PR #955, opened 2017-01-25
Make x-domain safe frame example work out of the box

brondsem:
* Nest in <script> tags
* Use garden-variety JS
* Dynamically parse host domain
* Use https for ad server

This addresses some of the #954 issues.  I didn't know what a valid replacement for `id` would be, and the IE HTTPS issue is outstanding as well.


brondsem:
I've added a commit which avoids the IE error, but read the commit message for all the gory details.  I do not have super high confidence in this yet, but it is more successful than the doc.write.

mkendall07:
Great stuff, thanks!

PR #962, opened 2017-01-31
Use package dependencies for ES6 Array shims

outoftime:
The shims that are vendored in to the library break in IE; also, no reason to reinvent the wheel. Instead, use the `array.prototype.find` and `array-includes` modules to ensure compatibility.

Original error message in IE:

### DIFF
--- a/package.json
+++ b/package.json
@@ -96,5 +96,8 @@
     "webpack-stream": "^3.1.0",
     "yargs": "^1.3.1"
   },
-  "dependencies": {}
+  "dependencies": {
+    "array-includes": "^3.0.2",
+    "array.prototype.find": "^2.0.3"
+  }
 }
--- a/src/polyfill.js
+++ b/src/polyfill.js
@@ -2,59 +2,15 @@
 Misc polyfills
 */
 /*jshint -W121 */
-if (!Array.prototype.find) {
-  Object.defineProperty(Array.prototype, "find", {
-    value: function(predicate) {
-      if (this === null) {
-        throw new TypeError('Array.prototype.find called on null or undefined');
-      }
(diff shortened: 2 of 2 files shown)


mkendall07:
LGTM. We should figure out why our unit tests didn't catch this.

nanek:
@outoftime Do you recall what version of IE? I'm testing our site in IE 9 and it works fine in Prebid 0.18 without this PR. Could this have been something else in your app?

mkendall07:
Good point, the screenshot references another script file and we test extensively in IE.

outoftime:
The problem was observed in IE 11. I can tell you that the change in this PR, in isolation, fixed the error. You’re correct that the error was being thrown from our main application bundle, so it’s possible that the original polyfill differed subtly from real browser behavior in terms of the property descriptor for the polyfilled methods, which caused another library in our bundle to break.

Anyway, at this point we bundle Prebid into a JS bundle that we control, so if you roll this back we can still work around the problem by importing an appropriate polyfill before importing Prebid.

mkendall07:
We are gonna to roll back this change, however, I'm not sure it's causing the issue you think it is. I just ran this code in IE 11: https://jsfiddle.net/tn3d87zy/ and don't see any exception. 

Also - note that the same code is used in the polyfill includes as it was with the manual includes we had (it does a check on `!Array.prototype.find` and replaces if not found).

outoftime:
@mkendall07 yeah it’s hard to say at this point, although maybe I can get more info on the error once this is rolled back. I think it’s probably more subtle than a transparent error in redefining `find`, but again, this PR alone made the difference between broken and not-broken.

PR #966, opened 2017-02-02
Added referrer parameter

bjorn-wo:
## Type of change

- [x ] Feature

## Description of change

Added a new parameter "referrer" where publisher can (optionally) insert the referrer domain if the ad is running within an iframe where the url can't be detected.

bandersson@wideorbit.com


integrationExamples/gpt/pbjs_example_gpt.html
```diff
@@ -343,6 +343,13 @@
                         params: {
                             inScreen: 'ggumtest' // REQUIRED str Tracking Id
                         }
+                    },
+                    {
+                        bidder: 'wideorbit',
+                        params: {
+                            pbId: 123, // REQUIRED Publisher Id,
+                            pId: 123456 // REQUIRED Placement Id
+                        }
```
matthewlane: `pbjs_example_gpt.html` already includes the same wideorbit params further up the page, are these needed or can they be removed?

src/adapters/wideorbit.js
```diff
@@ -1,213 +1,215 @@
-var bidfactory = require('../bidfactory.js'),
```
matthewlane: The line endings in this commit are CRLF (Windows-style), but the existing wideorbit commit uses LF (Unix-style) line endings. Can you keep LF line endings so the diff is easier to read?

integrationExamples/gpt/pbjs_example_gpt.html
bjorn-wo: Thank you for spotting this. It looks like there was a problem when mergin in the latest code.

aneuway2:
@bjorn-wo can you add information on the new `referrer` parameter to the [docs repo](https://github.com/prebid/prebid.github.io) ?

http://prebid.org/dev-docs/bidders.html#wideorbit

PR #967, opened 2017-02-02
Use new key descriptionUrl instead of vastUrl

jaiminpanchal27:
## Type of change

- [ x] Bugfix

## Description of change

Changed adserver code to to use descriptionUrl instead of vastUrl

## Other information

#950
@mkendall07 @dbridges12


mkendall07:
LGTM.

PR #633, opened 2016-09-15
add Roxot adapter

kir-roxot:
## Type of change

- [ ] Bugfix
- [ ] Feature
- [x] New bidder adapter
- [ ] Code style update (formatting, local variables)
- [ ] Refactoring (no functional changes, no api changes)
- [ ] Build related changes
- [ ] CI related changes
- [ ] Other

## Description of change

- test parameters for validating bids
```
{
  bidder: '<bidder name>',
  params: {
    "publisherId": "{publisher_id}"
  }
}
```
- contact email : it@roxot.com
- [x] official adapter submission

PR #969, opened 2017-02-03
Roxot analytic adapter

kir-roxot:
## Type of change

- [ ] Bugfix
- [ ] Feature
- [x] New bidder adapter
- [ ] Build related changes
- [ ] CI related changes
- [ ] Other

```
{
  bidder: 'roxotanalytic',
  params: {
  }
}
```


kir-roxot:
Hey @mkendall07  @matthewlane !

I represent Roxot. We have two approved prebid.js adapters: a bid request adapter and an analytics adapter which was built according to example.js -(https://github.com/prebid/Prebid.js/blob/master/src/adapters/analytics/example2.js).

We recommend all our publishers to use your prebid.js builder on prebid.org. But, unfortunately, our publishers have difficulties with the analytics adapter installation - they have to manually build prebid.js if they want to use it. There is no option to include it into prebid.js when building it from your website - http://prebid.org/download.html.

We decided to follow your suggestion (https://github.com/prebid/Prebid.js/pull/633#issuecomment-252120325) to resolve the issue. We are developing an additional Roxot Analytics adapter that combines the functionality of both adapters: it requests bids and sends analytics data. The analytics part of our new Roxot Analytics adapter is identical to previous standalone analytics adapter.

Issue #919, opened 2017-01-12
Reduce of empty array with no initial value


rafaguerrero:
## Type of issue
Bug

## Description
I've been trying to update to the latest version of prebid and I've found that some of our users are finding a error: "Reduce of empty array with no initial value".
As far as I can see the error comes from the class bidmanager, both functions bidsBackAdUnit and bidsBackAll have a reduce in there that has no initial value.

## Steps to reproduce
I've not been able to reproduce it neither in my local environment nor on production. We only have errors logged with this information.

## Platform details
We've been using v0.8.1 but we wanted to update to the latest. When we try the upgrade is when we found the error.

tedrand:
What adapters are you using?

rafaguerrero:
We had: sovrn, pubmatic, pulsepoint, aol & rubicon.
But we wanted to add indexExchange to the list.

rafaguerrero:
Any luck?

protonate:
Is there a page we can access that demonstrates the issue?  I'm not able to reproduce.

The reduce without initial value on the `flatten` operations is not expected to cause an issue unless the array passed to flatten is itself empty. In the `bidsBack...` functions this should not be the case.

rafaguerrero:
I had to revert the latest changes so I don't have a page right now that I can show.
Anyway I wasn't able to reproduce it myself, we only have information on the error logs.

rafaguerrero:
Here's a link with the configuration as I was explaining:
http://lab.marfeel.com/www.maduradas.com/index.html

As I said in the previous comment the error is not clear and I was not able to reproduce it on my own.

PR #981, opened 2017-02-10
Fix error when refreshing individual Ad Units and bidder responds slowly

ceresnam:
## Type of change

- [x] Bugfix

## Description of change
Bugfix for race condition.

Calling .reduce(flatten) in bidsBackAdUnit() and bidsBackAll() throws an exception when refreshing Ad unit.

That happens when some bidder takes long time to respond and second requestBids() (i.e. ad unit refresh) is called in meanwhile.

Here is a sequence diagram explaining in more detail why that happend. https://goo.gl/LPfkXc

### DIFF
--- a/src/bidmanager.js
+++ b/src/bidmanager.js
@@ -46,7 +46,7 @@ function bidsBackAdUnit(adUnitCode) {
     .map(request => request.bids
       .filter(adUnitsFilter.bind(this, $$PREBID_GLOBAL$$._adUnitCodes))
       .filter(bid => bid.placementCode === adUnitCode))
-    .reduce(flatten)
+    .reduce(flatten, [])
     .map(bid => {
       return bid.bidder === 'indexExchange' ?
           bid.sizes.length :
@@ -64,7 +64,7 @@ function add(a, b) {
 function bidsBackAll() {
   const requested = $$PREBID_GLOBAL$$._bidsRequested
     .map(request => request.bids)
-    .reduce(flatten)
+    .reduce(flatten, [])
     .filter(adUnitsFilter.bind(this, $$PREBID_GLOBAL$$._adUnitCodes))
     .map(bid => {
       return bid.bidder === 'indexExchange' ?


mkendall07:
@ceresnam 
Is there any observed bug that happens besides the thrown exception?

protonate:
Here is a related issue: #919

ceresnam:
@mkendall07 
this callback has not been called
https://github.com/prebid/Prebid.js/blob/master/src/bidmanager.js#L142

protonate:
Also possibly related issue #985.

protonate:
Thanks for the PR, if you would rebase onto current master it should resolve the build check failure, this was due to an unrelated PR.

protonate:
I pulled this change into a new branch to resolve the build failure, see #989, closing.

dmitriyshashkin:
I was trying to figure out if this bug could have caused a discrepancy. So I've spent quite some time trying to reproduce this issue. And the only way I was able to do that is by calling pbjs.requestBids with an adUnit that has it's bids array empty. eg `pbjs.requestBids({'adUnits': [{'code': 'a_code', 'bids': []}]});` or with an empty list of adUnits eg `pbjs.requestBids({'adUnits': []});`

If I reproduce the sequence described above, then even after the first auction is cleared and the second auction has started, and clearPlacements() has been called, the pbjs._bidsRequested still has a list of requests for the second auction. Yes, indeed they are for another ad unit, and they will be filtered-out inside bidsBackAdUnit, but what is left is not an empty array but an array with one element which is empty array. So the following code:
`
$$PREBID_GLOBAL$$._bidsRequested
  .map(request => request.bids
  .filter(adUnitsFilter.bind(this, $$PREBID_GLOBAL$$._adUnitCodes))
  .filter(bid => bid.placementCode === adUnitCode))
`

produces something like this: `[[]]`

And while `[].reduce(flatten)` will cause an error, `[[]].reduce(flatten)` works just fine. So unless I'm missing something, the only way to reproduce the error is to have empty _bidsRequested array, which can only be achieved if you call requestBids with empty list of bids or empty list of adUnits.

So again if I'm not missing something, the issue above can not cause any discrepancy unless you pass erroneous parameters to requestBids.

And perhaps the proper way to adress this was to check the parameters that are passed to pbjs.requestBids

It would be great if someone could confirm or correct me.

PR #990, opened 2017-02-13
check truthiness of adUnitCode

protonate:
## Type of change
- [x] Bugfix

## Description of change
`adUnitCode` is expected to be passed to `bidmanager.addBidResponse` and if it is not the lookup for originating bid request fails, the time to respond value is incorrect and can end an auction early.


mkendall07:
LGTM

Issue #977, opened 2017-02-09
Missing http-referrer for img tracking pixels on iOS


ckbo3hrk:
## Type of issue
[x] Bug ?

## Description
During tests on iOS we've noticed that http request for tracking pixel in rendered creatives doesn't contain referrer http-header and that may cause a discrepancy in statistics.  The referrer header is absent at least for Safari and Chrome mobile browsers. After some tests I've found that approach with rendering creative in iframe with iframe.srcdoc solves this issue. I beleive that this issue may affect any of ad platforms that relying on http-referrer information from theirs tracking pixels.

Here is POC that fixes this issue:
[/src/prebid.js:renderAd](https://github.com/prebid/Prebid.js/blob/82fd5a59c967eb398e1d5331c2234c0753fa7d9b/src/prebid.js#L285)
```javascript
if (doc === document || adObject.mediaType === 'video') {
    utils.logError('Error trying to write ad. Ad render call ad id ' + id + ' was prevented from writing to the main document.');
} else if (ad) {
  if (doc.defaultView && 'srcdoc' in doc.defaultView.frameElement) {
    doc.defaultView.frameElement.srcdoc = ad;
  } else {
    doc.write(ad);
    doc.close();
  }
  setRenderSize(doc, width, height);
}
```
## Steps to reproduce

1. Open http://jsfiddle.net/ckbo3hrk/u9qq2ee1/ url with Chrome or Safari on IOS
2. In intercepted requests (I used fiddler with proxy option enabled in order to intercept traffic from mobile device) there is no http referrer header for http://cpm.metaadserving.com/win?i=UBY6QS45GBA_0 request
3.  Open http://jsfiddle.net/ckbo3hrk/baqedx28/ url (it uses an updated version of prebid.js with proposed solution) with Chrome or Safari on IOS
4. Http referrer header is presented in requests as expected 

## Platform details
IOS 10.2.1 on Ipad Air,  Chrome 56.0.2924.79, Safari 10.0, Prebid 0.17.0

mkendall07:
Great stuff! I'm all for using HTML 5 when we can. Want to open up a PR? 

How did you use a mobile device to test with fiddler? Was it emulation or a real device?

ckbo3hrk:
> Want to open up a PR?

Once I've done some in-house tests I'll open up a PR.

> How did you use a mobile device to test with fiddler? Was it emulation or a real device?

[...]

PR #996, opened 2017-02-17
iOS Referrer fix

ckbo3hrk:
## Type of change

- [x] Bugfix
- [x] Feature

## Description of change

1. Implemented srcdoc ad rendering appoach as suggested in https://github.com/prebid/Prebid.js/issues/977#issue-206471234 for all browsers that support iframe.srcdoc property except  Firefox as it [doesn't send proper http referrer](https://bugzilla.mozilla.org/show_bug.cgi?id=1265961)  for such approach.
2. Fixed url.parse function to initialize pathname property in uniform way for all browsers (with leading /. Current version set pathname without leading / on IE11)


mkendall07:
@ckbo3hrk 
Can you add a unit test that asserts for the `url.parse` leading / change?

ckbo3hrk:
@mkendall07 I've changed the existing one to cover `url.parse` changed behaviour

src/prebid.js
```diff
@@ -133,6 +133,15 @@ function setRenderSize(doc, width, height) {
   }
 }
 
+/**
+ * Check if parent iframe of passed document supports content rendering via 'srcdoc' property
+ * @param {HTMLDocument} doc document to check support of 'srcdoc'
+ */
+function isSrcdocSupported(doc) {
```
mkendall07: can you move into `utils.js` please

ckbo3hrk: done

mkendall07:
LGTM. Thanks!

src/prebid.js
```diff
@@ -299,15 +300,19 @@ $$PREBID_GLOBAL$$.renderAd = function (doc, id) {
...
-        if (doc===document || adObject.mediaType === 'video') {
-          utils.logError('Error trying to write ad. Ad render call ad id ' + id + ' was prevented from writing to the main document.');
+        
+        if (doc === document || adObject.mediaType === 'video') {
+          utils.logError(`Error trying to write ad. Ad render call ad id ${id} was prevented from writing to the main document.`);
         } else if (ad) {
-          doc.write(ad);
-          doc.close();
+          if (isSrcdocSupported(doc)) {
+            doc.defaultView.frameElement.srcdoc = ad;
```
mkendall07: It looks like `srcdoc` completely wipes out any other content present in the iframe. This is causing DFP to fail to report on viewable impressions. I will be reverting this change until we can find a proper solution.

mkendall07: @ckbo3hrk fyi

ckbo3hrk: @mkendall07, I'd like to investigate this bug. Could you provide me a sample dfp tag to reproduce this issue ?

mkendall07: @ckbo3hrk 
Here is an example:
Working (filter by "activeview" in network)
http://mjacobson.devnxs.net/~mjacobson/discrepancy_testing/test_page_20.html

Not working:
http://mjacobson.devnxs.net/~mjacobson/discrepancy_testing/test_page_21.html

PR #1006, opened 2017-02-23
Fix bug with supporting non-cpm dealId bids.

mkendall07:
## Type of change

- [x] Bugfix
- [ ] Feature
- [ ] New bidder adapter
- [ ] Build related changes
- [ ] CI related changes
- [ ] Other

## Description of change
Support non-cpm dealId bids. Including support for Xaxis adapter. 

## Other information
@prebid/core for review

### DIFF
--- a/src/adapters/xhb.js
+++ b/src/adapters/xhb.js
@@ -8,6 +8,19 @@ const bidfactory = require('../bidfactory.js');
 
 const XhbAdapter = function XhbAdapter() {
 
+  var _defaultBidderSettings = {
+    alwaysUseBid: true,
+    adserverTargeting: [
+      {
+        key: 'hb_xhb_deal',
+        val: function (bidResponse) {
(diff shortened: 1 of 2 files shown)


daniel-hoffmann:
@mkendall07 Hey! Looks good to me in regards to the changes in the adapter.

mkendall07:
@snapwich 
want to give this a quick look as I think you refactored some of this code before?

snapwich:
LGTM.  :+1: 

With that change the deal should come through automatically as `hb_deal` if you want to use `alwaysUseBid` or `enableSendAllBids` rather than manually passing in adapter.

mkendall07:
@snapwich thanks. Yeah actually I don't think `hb_deal` comes through because it's considered a default key now, which is omitted on non-winning bids because we don't want to collide with the winning bid.

PR #1019, opened 2017-03-01
Create bids for requests without mutating ad units

protonate:
## Type of change

- [x] Bugfix

## Description of change
`Object.assign` should be passed first an empty object when creating bids for bid requests. Currently ad units are mutated with bid request data.


mkendall07:
Yeah good catch. LGTM.

PR #1022, opened 2017-03-02
Change identification of JavaScript user matching

bjorn-wo:
## Type of change
- [x] Bugfix

## Description of change
User matchings that are done by outputting a JavaScript were previously identified by using the string "javascript". This has been modified to instead compare to the string "js".


jaiminpanchal27:
Looks good

Issue #1025, opened 2017-03-02
sharethrough uses document.body before it exists


kitwestneat:
## Type of issue
bug

## Description
We load prebid in the document head section. Occasionally it can happen that prebid runs before document.body is created. The sharethrough adapter, however, uses document.body to load an iframe without checking if it actually exists.

## Steps to reproduce
Load prebid in the <head> section and request bids with sharethrough, do something time consuming in head.

### Expected results
sharethrough waits for document.body to exist before attempting to add an iframe

### Actual results
sharethrough tries to use document.body and throws an error

## Platform details
Prebid commit 3e50de14842c1fa7ff1e7fa5f74d04ae7b9e12ff, chrome browser

rizhang:
#1042 should fix this issue as well, since the adapter will be using ajax to make requests instead of iframes

matthewlane:
Closing as fixed by #1042

PR #1026, opened 2017-03-02
sharethrough: wait for document.body to exist

kitwestneat:
## Type of change
- [x ] Bugfix

## Description of change

If document.body doesn't exist yet, wait for content loaded event
from document before adding iframe.

## Other information

Fixes issue #1025

### DIFF
--- a/src/adapters/sharethrough.js
+++ b/src/adapters/sharethrough.js
@@ -51,7 +51,12 @@ var SharethroughAdapter = function SharethroughAdapter() {
     iframe.src = url;
     iframe.style.cssText = 'display:none;';
 
-    document.body.appendChild(iframe);
+    if (document.body)
+      document.body.appendChild(iframe);
+    else
+      document.addEventListener("DOMContentLoaded", function() {
+        document.body.appendChild(iframe);
+      });
   };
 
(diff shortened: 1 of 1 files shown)


snapwich:
Is there a reason you use an iFrame to load JSONP content rather than a script tag?  Also, Prebid.js provides `adloader#loadScript` to load JSONP data.

I'm pretty sure an iFrame is going to both slow down the auction as well as put you guys at a disadvantage as far as response time.

kitwestneat:
@snapwich I'm not a sharethrough developer, so I can't comment on their loading procedure. Using JSONP directly instead of their iframe solution seems a lot simpler to me as well. With the current parameters, their bid server returns HTML containing a script tag with the callback, so not JSONP per se (they use the term ijson).

It looks like @rizhang made the initial commit, so perhaps they can comment on it?

rizhang:
We used an iframe because that is how are endpoint is meant to be consumed and we didn't want to do any backend work.

I made a PR to use ajax instead of an iframe hoping that this will speed up our response times. #1042

kitwestneat:
@rizhang thanks for your PR and backend change, it seems to work so I'll close this

PR #1034, opened 2017-03-07
Update Roxot prebid analytic adapter

kir-roxot:
- [x] Feature

## Description of change
Hi Guys,

We have changed our analytics adapter logics. AUCTION_END and REQUEST_BIDS events were also added in AnalyticsAdapter to get it full.


kir-roxot:
@matthewlane @mkendall07 Hi guys, i've added test publisher for you. You can test analytic that sent at our endpoint with this config, and you will get 200 ok response:
```
pbjs.que.push(function () {
  pbjs.enableAnalytics({
    provider: 'roxot',
    options: {
      publisherIds: ['bea17352-fb94-490d-b206-45096c96dbbc']
    }
  });
});
```

src/adapters/analytics/roxot.js
```diff
@@ -1,19 +1,94 @@
...
+let roxotAdapter = utils.extend(adapter({url, analyticsType}),
   {
-    track({ eventType, args }) {
-      ajax(url, (result) => utils.logInfo('Event ' + eventType + ' sent to roxot analytics with result ' + result), JSON.stringify({ eventType, args }));
+    track({eventType, args}) {
+      if (!checkOptions()) {
+        return;
+      }
+
+      args.ad = "";
```
matthewlane: Probably best to check for the existence of `args` first, if `args` is undefined this causes a `TypeError` originating in event.js to be thrown

test/spec/roxot_analytic_spec.js
```diff
@@ -1,4 +1,3 @@
-let assert = require('assert');
 let analytic = require('../../src/adapters/analytics/roxot');
```
matthewlane: For the assertions, was meaning that the test doesn't check for anything in it's current state. As it's structured could check that the `track` function was called in response to the emitted events. This can be done by setting up a sinon spy before the emits with `sinon.spy(analytic, 'track');`, then after the emits something like `sinon.assert.callCount(analytic.track, 7);`. To make this work the analytic import here would need to change to use ES2015 `import` statement, or to `let analytic = require('../../src/adapters/analytics/roxot').default;` because of how the default export is compiling. Feel free to test this according to your preferences though

Also please move this into the analytics test directory at `test/spec/adapters/analytics`, thanks

kir-roxot:
Hey @matthewlane !

We added the tests you described.
Is it possible to jump the queue and release our updated analytics adapter as soon as possible? 

Currently, the adapter on http://prebid.org/download.html is outdated and doesn't collect data as it is supposed to. We are concerned that publishers might be confused as the old version of our adapter doesn't work as described.

matthewlane:
Thanks for the tests. We'll get the PR merged as soon as possible. There's not currently a plan to do a patch release but it would be included whenever the next release is

kir-roxot:
@matthewlane Sorry, fixed.

PR #1037, opened 2017-03-08
hb_adid should be uppercase in all cases

jaiminpanchal27:
## Type of change

- [x] Bugfix

## Description of change

hb_adid should be in uppercase for all cases


mkendall07:
LGTM.

Issue #877, opened 2016-12-14
Need to generate unique impression ID


pierre-nicolas:
## Type of issue
Feature request

## Description
As a buyer integrated with several adaptors, I want to be able to know when the same impression is being offered by two or more of them. This can be done by generating a unique impression ID in prebid.js and then distributing it through adaptors and their own integrations.

Note that OpenRTB has already introduced the notion of "source" in [its 2.5 version](https://www.iab.com/guidelines/real-time-bidding-rtb-project/), and in particular a "tid" field to relay this impression identifier.

cotarmanach:
+1

pieroxy:
+1

BenjaminCh:
+1

mkendall07:
We do have an impression ID already, I don't think anyone is actively using to to present to the buyer. (it might be included in some of the OpenRTB complaint bidders). Your best bet would be to initiate conversations with those demand partners to see how they can include it. 

For reference, this would be the `bid.bidId` that gets passed through when `callBids()` function is invoked on each adapter. Here's an example of Sovrn using it: https://github.com/prebid/Prebid.js/blob/master/src/adapters/sovrn.js#L47

Swiiip:
The bidRequest.requestId is shared accross all bidRequests on all adUnits Is there a unique ID shared between all bidRequests for a given adUnit ?

mkendall07:
> Is there a unique ID shared between all bidRequests for a given adUnit ?
Currently no. We can add one though. This should be added into the [1.0.0 proposal ](https://github.com/prebid/Prebid.js/issues/891)

mkendall07:
Will implement as part of https://github.com/prebid/Prebid.js/issues/891

PR #1040, opened 2017-03-09
Add a new generated field transactionId to each adunits.

rcassan:
Transaction ID that must be common across all participants in
this bid request (e.g., potentially multiple exchanges) (OpenRTB v2.5).

## Type of change
- [x] Feature

## Description of change
As a buyer integrated with several adaptors, I want to be able to know when the same impression is being offered by two or more of them. The goal of this PR is to generate a unique transaction ID in prebid.js and then distributing it through adaptors and their own integrations.

Note that OpenRTB has already introduced the notion of "source" in its 2.5 version, and in particular a "tid" field to relay this impression identifier.

## Other information
This feature was first requested by https://github.com/prebid/Prebid.js/issues/877

### DIFF
--- a/src/adaptermanager.js
+++ b/src/adaptermanager.js
@@ -29,6 +29,7 @@ function getBids({bidderCode, requestId, bidderRequestId, adUnits}) {
         return Object.assign(bid, {
           placementCode: adUnit.code,
           mediaType: adUnit.mediaType,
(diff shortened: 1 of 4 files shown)


mkendall07:
LGTM.

mkendall07:
@protonate 
can you give this a quick look?

PR #1043, opened 2017-03-10
Add alias freewheel-ssp to stickyadstv bidder adapter 

guillaume-sticky:
## Type of change

- [ ] Bugfix
- [ ] Feature
- [ ] New bidder adapter
- [ ] Build related changes
- [ ] CI related changes
- [x] Other

## Description of change
As new may know StickyAdsTV is now FreeWheel (www.stickyads.tv)
So I've made changes to use the alias freewheel-ssp pointing to the stickyadstv bidder adaptor

```
{
  bidder: 'freewheel-ssp',
  params: {
      zoneId: 277225
  }
}
```
gandouard@freewheel.tv

## Other information
Test page: http://sandbox.stickyadstv.com/qa/prebid.html

We are using prebid to serve inbanner video ads and we are wondering if this case is concerned by the 20$ CPM CAP?  
(http://prebid.org/dev-docs/show-video-with-a-dfp-video-tag.html#implement-custom-price-buckets-to-work-around-the-default-cpm-cap-of-20)

When I run tests using a price over 20$ everything seems fine.

If we need to work around this, can we do it from inside the bidder adapter or is the "custom price bucket" something the publisher needs to setup?


mkendall07:
LGTM.

PR #1057, opened 2017-03-20
Add GourmetAds AppNexus Alias

tmoerschbacher:
Adding Gourmet Ads as an alias of AppNexus adapter.

### DIFF
--- a/adapters.json
+++ b/adapters.json
@@ -73,6 +73,11 @@
       }
     },
     {
+      "appnexus": {
+        "alias": "gourmetads"
+      }
+    },
+    {
       "appnexusAst": {
         "supportedMediaTypes": ["video"]
       }


mkendall07:
LGTM.

chefbenjamin:
@matthewlane @mkendall07 @tmoerschbacher Given the addition, will this be added to the front page of prebid.org ? What else do we need to provide ?

tmoerschbacher:
Hi @chefbenjamin, no other info is required.  Would you like to have Gourmet show as an option on the download page (http://prebid.org/download.html)?  It is not something included here, but we can make that happen if you'd like.

chefbenjamin:
@tmoerschbacher I see you've done it already. thanks mate.

viddyrizkiyansah:
Hi,

Sorry for jump in, please advice if we want to be a bidder on appnexus adapter like gourmet ads, what we have to do ?

so our publisher who already using appnexus can add us in their Header Bidding as aliases

Thank you

PR #1066, opened 2017-03-22
allow writing into current document if prebid is loaded inside an iframe

surensilva:
## Type of change

- [ ] Bugfix
- [x] Feature
- [ ] New bidder adapter
- [ ] Build related changes
- [ ] CI related changes
- [ ] Other

## Description of change

For PostBid implementations, the secondary iframe ('postbid_if') where the ad is inserted into can cause view-ability metrics to drop. This enhancement allows the passing of the current document to the renderAd provided that prebid is loaded inside an iframe.

Updated PostBid Example: https://gist.github.com/surensilva/bdeec34abf1027acd8a87eb48a79f144

### DIFF
--- a/src/prebid.js
+++ b/src/prebid.js
@@ -301,7 +301,7 @@ $$PREBID_GLOBAL$$.renderAd = function (doc, id) {
         var url = adObject.adUrl;
         var ad = adObject.ad;
 
-        if (!utils.inIframe() || adObject.mediaType === 'video') {
+        if ((doc === document && !utils.inIframe()) || adObject.mediaType === 'video') {
           utils.logError(`Error trying to write ad. Ad render call ad id ${id} was prevented from writing to the main document.`);
         } else if (ad) {
           if (isSrcdocSupported(doc)) {
--- a/src/prebid.js
+++ b/src/prebid.js
@@ -301,7 +301,7 @@ $$PREBID_GLOBAL$$.renderAd = function (doc, id) {
         var url = adObject.adUrl;
         var ad = adObject.ad;
 
-        if (!utils.isInIframe() || adObject.mediaType === 'video') {
+        if (!utils.inIframe() || adObject.mediaType === 'video') {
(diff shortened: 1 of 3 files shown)


mkendall07:
This is a great catch, thank you!

src/prebid.js
```diff
@@ -301,7 +301,7 @@ $$PREBID_GLOBAL$$.renderAd = function (doc, id) {
         var url = adObject.adUrl;
         var ad = adObject.ad;
 
-        if (doc === document || adObject.mediaType === 'video') {
+        if (!utils.inIframe() || adObject.mediaType === 'video') {
```
mkendall07: Unfortunately this logic is wrong, it fails for the standard prebid case (always evaluates to true when not in an iframe).

src/utils.js
```diff
@@ -588,4 +588,12 @@ export function isSrcdocSupported(doc) {
 
 export function cloneJson(obj) {
   return JSON.parse(JSON.stringify(obj));
+}
+
+export function inIframe() {
```
mkendall07: I think the problem here is we need to check a passed in window param, which we don't have. I think we have to change the `renderAd` API to accept the window instead of doc :(

src/prebid.js
snapwich: Yes you are correct, this breaks my test page;  we should have tested that.

src/utils.js
snapwich: Hmm; I would hate to break backwards compatibility.  Suren and I will brainstorm some techniques to see if we can come up with a backwards compatible solution that still solves this problem, and if so we'll update the pull-request.  Thanks!

snapwich:
@mkendall07 updated this to be more specific for our case so it doesn't break the general Prebid.js case.  Also updated `isSrcdocSupported` as it was breaking when ran inside a cross-domain iFrame.

snapwich:
@prebid/core can we get another review on this to see if it's good to go?  We'd like to merge this sooner than later.  Thanks!

mkendall07:
LGTM

PR #1082, opened 2017-03-27
Outstream Video Support

protonate:
## Type of change
- [x] Feature

## Description of change
As media formats proliferate a Renderer type is proposed. This provides a set of functions to specify, load, configure and utilize a custom 3rd party renderer for a given bid's media format.

An initial use case for Renderer is support for outstream video as shown in this change. Here we are assuming the `renderer_url` is a value returned on a Bid response object, and if present the Bid response is assumed to be an outstream video ad in need of a custom renderer.

A renderer source could also be specified at the adapter or publisher level, which can be done by setting the `rendererInstance.url` value appropriately.

To implement in a bidder adapter, import Renderer module:
```
import { Renderer } from 'src/renderer';
```

A `renderer` can be installed on a bid response in a bidder adapter by using the static `Renderer.install` method. This takes an object with the following properties:
```
  {
    url: 'path/to/renderer.js'
    config: { example: 'an object to configure renderer options' },
    id: 1, // an id may be present in the renderer info returned on a bid
    callback: () => return 'function to call once renderer has loaded'
  }
```

An example looks like:
```
bid.renderer = Renderer.install({
  id: ad.renderer_id,
  url: ad.renderer_url,
  config: { adText: `AppNexus Outstream Video Ad via Prebid.js` },
  callback: () => onOutstreamRendererLoaded.call(null, bid)
});
```
Once a renderer is installed on a bid, the renderer can set it's render function using `setRender()`:
```
  function onOutstreamRendererLoaded(bid) {
    bid.renderer.setRender(outstreamRender);
  }
```
The render function will receive the bid response object and the renderer can be called with it:
```
  function outstreamRender(bid) {
    window.ANOutstreamVideo.renderAd({
      tagId: bid.adResponse.tag_id,
      sizes: [bid.getSize().split('x')],
      targetId: bid.adUnitCode, // target div id to render video
      uuid: bid.adResponse.uuid,
      adResponse: bid.adResponse,
      rendererOptions: bid.renderer.getConfig()
    }, handleOutstreamRendererEvents.bind(bid));
  }
```
For the AppNexus Outstream Video renderer, the second argument takes a function that is invoked on events. We use this to call the Renderer's `handleVideoEvent` method:
```
  function handleOutstreamRendererEvents(id, eventName) {
    const bid = this;
    bid.renderer.handleVideoEvent({ id, eventName });
  }
```
A `renderer` can register event handlers using a hash with keys mapped to the renderers event names:
```
bid.renderer.setEventHandlers({
  impression: () => utils.logMessage('AppNexus outstream video impression event'),
  loaded: () => utils.logMessage('AppNexus outstream video loaded event'),
  ended: () => {
    utils.logMessage('AppNexus outstream renderer video event');
    document.querySelector(`#${bid.adUnitCode}`).style.display = 'none';
  }
});
```


protonate:
@mkendall07 steel thread version for review

src/prebid.js
```diff
@@ -332,6 +334,38 @@ $$PREBID_GLOBAL$$.renderAd = function (doc, id) {
...
+    tagId: adObject.adResponse.tag_id,
+    sizes: [adObject.getSize().split('x')],
+    targetId: adObject.adUnitCode, // target div id to render video
+    uuid: adObject.adResponse.uuid, // is this the correct UUID
+    adResponse: adObject.adResponse
+  });
+};
+
+function performRenderViaRenderer(doc, adObject) {
+  window.apntag = { debug: true };
```
mkendall07: this logic is specific to appnexus, we should move it into the adapter or into a renderer interface type.

src/prebid.js
```diff
@@ -332,6 +334,38 @@ $$PREBID_GLOBAL$$.renderAd = function (doc, id) {
...
+  window.apntag.registerRenderer = function(id, cb) {
+    renderOutstream(cb.renderAd, adObject);
+  };
+
+  // use renderer defined by creative or default to ANOutstreamVideo.js
+  loadScript(
+    adObject.rendererUrl ||
+    'http://cdn.adnxs.com/renderer/video/ANOutstreamVideo.js'
```
mkendall07: make protocoless to support https

test/spec/e2e/gpt-examples/gpt_outstream.html
```diff
@@ -0,0 +1,234 @@
...
+    var videoAdUnit = {
+      code: 'video1',
+      sizes: [ 640, 480 ],
+      mediaType: 'video-outstream',
+      bids: [
+        {
+          bidder: 'appnexusAst',
+          params: {
+            placementId: '5768085',
+            video: {
```
mkendall07: I wonder if we could move these params up to the adunit level?

src/prebid.js
```diff
@@ -293,23 +293,25 @@ $$PREBID_GLOBAL$$.renderAd = function (doc, id) {
...
+
         //emit 'bid won' event here
         events.emit(BID_WON, adObject);
 
-        var height = adObject.height;
-        var width = adObject.width;
-        var url = adObject.adUrl;
-        var ad = adObject.ad;
+        const { height, width, ad, mediaType } = adObject;
+        const url = adObject.adUrl;
```
mkendall07: This can be included in the destructuring above

src/prebid.js
```diff
@@ -293,23 +293,25 @@ $$PREBID_GLOBAL$$.renderAd = function (doc, id) {
...
 
-        if (doc === document || adObject.mediaType === 'video') {
+        if (doc === document) {
           utils.logError(`Error trying to write ad. Ad render call ad id ${id} was prevented from writing to the main document.`);
+        } else if (mediaType === 'video' || mediaType === 'video-outstream') {
```
mkendall07: I don't think we want to render a pure `video` response yet, since that requires a video player

integrationExamples/gpt/amp/amp_page_with_taboola_html.html
```diff
@@ -0,0 +1,180 @@
+<!doctype html>
```
mkendall07: drop this file.

src/Renderer.js
```diff
@@ -0,0 +1,37 @@
+import { loadScript } from 'src/adloader';
+const utils = require('src/utils');
```
mkendall07: nit: why not import?

src/adapters/appnexusAst.js
```diff
@@ -242,6 +244,45 @@ function AppnexusAstAdapter() {
...
+            config: {},
+            callback: () => {
+              bid.adResponse.ad = bid.adResponse.ads[0];
+              bid.adResponse.ad.video = bid.adResponse.ad.rtb.video;
+              bid.renderer.setRender(() => {
+                window.ANOutstreamVideo.renderAd({
+                  tagId: bid.adResponse.tag_id,
+                  sizes: [bid.getSize().split('x')],
+                  targetId: bid.adUnitCode, // target div id to render video
+                  uuid: bid.adResponse.uuid, // is this the correct UUID ?
```
mkendall07: should work fine.

src/adapters/appnexusAst.js
```diff
@@ -242,6 +244,45 @@ function AppnexusAstAdapter() {
...
+                  adResponse: bid.adResponse,
+                  rendererOptions: {
+                    adText: 'Prebid Outstream Video Ad'
```
mkendall07: I think we might want to default this to empty text.

src/adapters/appnexusAst.js
```diff
@@ -242,6 +244,45 @@ function AppnexusAstAdapter() {
...
+                  }
+                }, (id, eventName) => {
```
mkendall07: Is `id` here `uuid`? If so, maybe we should make this `adUnit.code` instead.

src/adapters/appnexusAst.js
```diff
@@ -242,6 +244,45 @@ function AppnexusAstAdapter() {
...
+                  bid.renderer.handleVideoEvent({ id, eventName });
+                });
+              });
+              bid.renderer.setEventHandlers({
+                impression: () => utils.logMessage('AppNexus outstream video impression event'),
```
mkendall07: who's firing these events?

mkendall07:
Looks great! Just a few minor changes/suggestions

mkendall07:
@protonate looks like there is a conflict.

jaiminpanchal27:
LGTM.

src/adapters/appnexusAst.js
```diff
@@ -225,6 +227,26 @@ function AppnexusAstAdapter() {
     return tag && tag.ads && tag.ads.length && tag.ads.find(ad => ad.rtb);
   }
 
+  function outstreamRender(bid) {
+    window.ANOutstreamVideo.renderAd({
```
mkendall07: Exception would be thrown here if `window.ANOutstreamVideo` isn't defined but should be caught in `renderAd` so I guess it's alright for now.

mkendall07:
LGTM

richardhhh:
Hey @mkendall07    quick question: If you want to render outstream without DFP just straight by calling and rendering on the page, I took the example from postbid, modifying it to video, then getting bids returned, but don't get the ad to render. Not sure if this works for outstream in the same fashion as for display using the iFrame. Appreciate a short note.

```
              // If any bidders return any creatives
          if (params && params['hb_adid']) {
                    console.log("creatives returned");
            pbjs.renderAd(iframeDoc, params['hb_adid']);
          }
        }
      });
    });

  </script>

  <iframe id='dynamic_ad' FRAMEBORDER="0" SCROLLING="no" MARGINHEIGHT="0" MARGINWIDTH="0" TOPMARGIN="0" LEFTMARGIN="0" ALLOWTRANSPARENCY="true" WIDTH="0" HEIGHT="0"></iframe>
```

spormeon:
is there a full list of renderer options anywhere, i cant find it in an docs only the adText one:

renderer: { options: { adText: 'Advertisement', skippable: true } },

PR #1096, opened 2017-03-30
Adapter bug fix

bjorn-wo:
## Type of change
- [x] Bugfix

## Description of change
We had a an issue server side that output the wrong identifier string to our adapter which caused the adapter to fail with an exception.

In order to avoid this in the future, we made the adapter more robust for this kind of errors.


test/spec/adapters/wideorbit_spec.js
```diff
@@ -489,6 +489,25 @@ describe('wideorbit adapter tests', function () {
...
+                    {
+                        Type: 'unrecognized',
+                        Url: 'http%3A%2F%2Fwww.admeta.com%2F1.js'
+                    }
+                ],
+                Placements: placements
+            };
+
+            $$PREBID_GLOBAL$$.handleWideOrbitCallback(response);
+
```
mkendall07: shouldn't you assert that `addBidResponse` wasn't invoked here?

bjorn-wo: It will actually call addBidResponse. It will however not throw an exception as it did before. I agree the test isn't very good, but it did catch my particular issue.

mkendall07:
LGTM

PR #1108, opened 2017-04-06
Sharethrough: Change to using a closure for the callback from ajax to preserve bidObj

rizhang:
## Type of change

- [x] Bugfix
- [ ] Feature
- [ ] New bidder adapter   
- [ ] Build related changes
- [ ] CI related changes
- [ ] Other

## Description of change

Passing the bid request object inside the callback function, instead of relying on fetching it with a bid Id. When multiple auctions are ran, old bid requests will be deleted, which causes the adapter to error out.

rzhang@sharethrough.com


test/spec/adapters/sharethrough_spec.js
```diff
@@ -117,8 +108,8 @@ describe('sharethrough adapter', () => {
                               "stxUserId": ""
                             };
 
-      pbjs.strcallback(JSON.stringify(bidderReponse1));
-      pbjs.strcallback(JSON.stringify(bidderReponse2));
+      adapter.callback(bidderRequest.bids[0], JSON.stringify(bidderReponse1));
```
harpere: rather than exposing the callback just for testing, it's preferred that you test using the `sinon.fakeServer` with a mock response and calling `adapter.callBids()`.  You can see an example in [here](https://github.com/prebid/Prebid.js/blob/master/test/spec/adapters/rubicon_spec.js#L466).

src/adapters/sharethrough.js
```diff
@@ -39,10 +45,9 @@ var SharethroughAdapter = function SharethroughAdapter() {
     return url;
   }
 
-  $$PREBID_GLOBAL$$.strcallback = function(bidResponse) {
+  function _strcallback(bidObj, bidResponse) {
     bidResponse = JSON.parse(bidResponse);
```
harpere: I see there was another change (causing a conflict) where you wrapped this in a separate try/catch.  I'd still recommend wrapping it in the try block below so that the bid is added to the bidManager (in your `_handleInvalidBid()` function) so it can still be tracked.

rizhang:
Made the changes!

src/adapters/sharethrough.js
```diff
@@ -102,6 +101,7 @@ var SharethroughAdapter = function SharethroughAdapter() {
   return {
     callBids: _callBids,
     str : str,
+    callback: _strcallback
```
harpere: no need to expose `callback` any longer

harpere:
thanks @rizhang!  made one more minor comment.

rizhang:
Made the change!

PR #1115, opened 2017-04-10
AOL Adapter - ONE Mobile endpoint implemented.

mpereira96:
## Type of change
- [x] Feature

## Description of change
This PR is to implement the ONE Mobile endpoint, which is a second endpoint where bid requests can be made from depending on the parameters set on the bid config. This also adds a feature where if enabled by choice the adapter will drop sync pixels on bid response. 

Is there a formal procedure for updating the bidder params for the AOL adapter?

### DIFF
--- a/src/adapters/aol.js
+++ b/src/adapters/aol.js
@@ -2,17 +2,130 @@ const utils = require('../utils.js');
 const ajax = require('../ajax.js').ajax;
 const bidfactory = require('../bidfactory.js');
 const bidmanager = require('../bidmanager.js');
+const constants = require('../constants.json');
 
 const AolAdapter = function AolAdapter() {
 
   let showCpmAdjustmentWarning = true;
   const pubapiTemplate = template`${'protocol'}://${'host'}/pubapi/3.0/${'network'}/${'placement'}/${'pageid'}/${'sizeid'}/ADTECH;v=2;cmd=bid;cors=yes;alias=${'alias'}${'bidfloor'};misc=${'misc'}`;
+  const nexageBaseApiTemplate = template`${'protocol'}://${'host'}/bidRequest?`;
+  const nexageGetApiTemplate = template`dcn=${'dcn'}&pos=${'pos'}&cmd=bid${'ext'}`;
   const BIDDER_CODE = 'aol';
-  const SERVER_MAP = {
+  const MP_SERVER_MAP = {
     us: 'adserver-us.adtech.advertising.com',
     eu: 'adserver-eu.adtech.advertising.com',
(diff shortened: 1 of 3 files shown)


src/adapters/aol.js
```diff
@@ -2,17 +2,130 @@ const utils = require('../utils.js');
...
+        }
+        return idempotentFn();
+      };
+      if (document.readyState === "complete") {
+        return idempotentFn();
+      }
+      if (document.addEventListener) {
+        document.addEventListener("DOMContentLoaded", idempotentFn, false);
+        window.addEventListener("load", idempotentFn, false);
+      } else if (document.attachEvent) {
```
snapwich: Prebid.js only supports IE9+, so you should be fine with using `addEventListener` only.

src/adapters/aol.js
```diff
@@ -2,17 +2,130 @@ const utils = require('../utils.js');
...
+    iframe.width = 1;
+    iframe.height = 1;
+    iframe.style.display = 'none';
+    iframe.src = pixelsItem.src;
+    if (document.readyState === 'interactive' ||
+      document.readyState === 'complete') {
+      document.body.appendChild(iframe);
+    } else {
+      domReady(() => {
+        document.body.appendChild(iframe);
```
snapwich: Since you're checking the readyState above, I think removing `domReady` and just using `DOMContentLoaded` should be sufficient here.

src/adapters/aol.js
```diff
@@ -2,17 +2,130 @@ const utils = require('../utils.js');
...
+    }
+
+    return pixelsItems;
+  }
+
+  function renderPixelElements(pixelsElements) {
+    pixelsElements.forEach((element) => {
+      switch (element.tagName) {
+        case SYNC_TYPES.img:
+          return renderPixelImage(element);
```
jaiminpanchal27: Check if you can reuse these functions 
https://github.com/prebid/Prebid.js/blob/master/src/utils.js#L446

vzhukovsky: Hi. Thank you for the note.
If we use this method we will have to insert elements in DOM directly each time
We would like to avoid it so we prefer to use own method.

jaiminpanchal27:
This is merged into master. Please submit a PR to the [docs repo](https://github.com/prebid/prebid.github.io) to add a file for your adapter to the [bidders directory](https://github.com/prebid/prebid.github.io/tree/master/dev-docs/bidders) so your adapter's params will appear on the [bidders page](http://prebid.org/dev-docs/bidders.html). Thank you for contributing

PR #1119, opened 2017-04-11
Improvement/prebid iframes amp pages

protonate:
## Type of change
- [x] Other

## Description of change
Implementation details can make the single auction model typical of Prebid potentially problematic on AMP pages. This experimental change loads Prebid.js per <amp-ad> component which clears up some of these issues.

### DIFF
--- a/integrationExamples/gpt/amp/remote.html
+++ b/integrationExamples/gpt/amp/remote.html
@@ -25,7 +25,7 @@ document.write('<script'+' src="'+encodeURI(u)+'"><'+'/script>');
 <!-- start Prebid.js AMP integration -->
 
 <script>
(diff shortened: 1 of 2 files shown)


integrationExamples/gpt/amp/remote.html
```diff
@@ -25,7 +25,7 @@
 <!-- start Prebid.js AMP integration -->
 
 <script>
-    // The Prebid global must match Prebid.js setting
+    // The Prebid global must match Prebid.js setting. This example file is not preprocessed so below we refer to global as `pbjs`.
     var $$PREBID_GLOBAL$$ = pbjs;
```
mkendall07: Should we just make this `pbjs`? seems confusing.

PR #1126, opened 2017-04-15
Adform adapter returns incorrect bid CPM

ceresnam:
## Type of change

- [x] Bugfix

## Description of change
AdformAdapter returns response with incorrect bid CPM. The value is 1000x smaller.

Here are screenshots that demonstrate it:

- setup of a test campaign in Adform with CPM set to 2.60Eur https://goo.gl/RmI0Iu
- chrome debug showing that cpm value is 0.0026 https://goo.gl/aapNW7
- console log of pbjs_debug=True showing that hb_pb value is 0.00 instead of 2.60 https://goo.gl/qEsXty

## Other information

That seems suspicious to me that nobody reported it yet. So, probably an official maintainer of the adform adapter should confirm it?


ceresnam:
will create another one with cleaner pullrequest (just patch with one proposed commit, without merges of master)

PR #1134, opened 2017-04-18
added audit beacon to detect misuse of this bidder.  Detects auctions…

jstocker76:
… within iframes or popup windows.

## Type of change

- [ ] Bugfix
- [x ] Feature
- [ ] New bidder adapter   
- [ ] Build related changes
- [ ] CI related changes
- [ ] Other

## Description of change
Added an audit beacon to detect unintended use of this bidder.

```
{
  bidder: 'rhythmone',
  params: {
                        placementId: "34887",
						zone: "test",
						path: "mvo",
						debug: true,
						method:"get"
                    }
}
```

### DIFF
--- a/src/adapters/rhythmone.js
+++ b/src/adapters/rhythmone.js
@@ -194,7 +194,7 @@ module.exports = function(bidManager, global, loader){
       },
       ao = document.location.ancestorOrigins,
       q = [],
-      u = "//hbevents.1rx.io/hbaudit?",
+      u = "//hbevents.1rx.io/audit?",
       i = new Image();
     
     if(ao && ao.length > 0)
--- a/src/adapters/rhythmone.js
+++ b/src/adapters/rhythmone.js
@@ -194,7 +194,7 @@ module.exports = function(bidManager, global, loader){
       },
       ao = document.location.ancestorOrigins,
       q = [],
-      u = "//hbevents.1rx.io/hbperf?",
+      u = "//hbevents.1rx.io/hbaudit?",
       i = new Image();
     
     if(ao && ao.length > 0)
--- a/src/adapters/rhythmone.js
+++ b/src/adapters/rhythmone.js
@@ -194,7 +194,7 @@ module.exports = function(bidManager, global, loader){
       },
       ao = document.location.ancestorOrigins,
       q = [],
-      u = "//te.renetl.com/hbperf?",
+      u = "//hbevents.1rx.io/hbperf?",
       i = new Image();
     
     if(ao && ao.length > 0)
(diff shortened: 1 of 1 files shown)


hipstern:
looks good

ehoch:
@mkendall07 @protonate Sorry to be a pain, but is this something we can get reviewed ASAP? This is actually a really important bug fix for multiple auctions.

jstocker76:
updated code with style suggestions.

src/adapters/rhythmone.js
```diff
@@ -224,53 +186,76 @@ module.exports = function(bidManager, global, loader){
     return endpoint;
   }
   
+  function sendAuditBeacon(placementId){
+    var data = {
+        doc_version : 1,
+        doc_type : "Prebid Audit",
+        placement_id: placementId
+      },
+      ao = document.location.ancestorOrigins,
```
protonate: Is browser support for `ancestorOrigins` a concern here?

hipstern: It's not a concern, it is only available in webkit browsers TMK so it will only be used if there.

src/adapters/rhythmone.js
```diff
@@ -224,53 +186,76 @@ module.exports = function(bidManager, global, loader){
...
+      q = [],
+      u = "//hbevents.1rx.io/audit?",
+      i = new Image();
+    
+    if(ao && ao.length > 0){
+      data.ancestor_origins = ao[ao.length-1];
+	}
+    
+    data.popped = window.opener!==null?1:0;
+    data.framed = window.top===window?0:1;
```
matthewlane: Please place this `window.top` reference in a try/catch as well

jstocker76: With all due respect, this is unnecessary.  A security exception is only thrown when accessing the properties of top.  The window object itself can be compared safely.

mkendall07: @jstocker76 run this: https://jsfiddle.net/04o2funn/

jstocker76: odd, I get no exception when I run that code in a sample file with a cross domain iframe.  Furthermore, I don't get that exception when running the same code by targeting chrome's console to that iframe.  What sorcery is this?

jstocker76: I see.  It's the alert causing the exception.  If you just run "window.top" no exception.

mkendall07: yeah that makes sense. This is good. Prebid core should probably just utility this function because tons of bidders do this check.

jstocker76: alert(window.top === window);

jstocker76: Reeling in the proliferation of duplicate functionality must be a full time job.  The list of bidders is getting _quite_ long

Issue #949, opened 2017-01-24
logError doesn't work during initialization stage


dmitriyshashkin:
## Type of issue
Bug

## Description
At the early stages of initialization (namely when bidder is being aliased in  aliasBidAddapter method) utils.debugTurnedOn always returns false. Therefore utils.logError method doesn't write anything to console (which is quite inconvenient).

## Steps to reproduce
Alias a bidder that can not be aliased.

### Expected results
Error message "%adapter name% bidder does not currently support aliasing" is being printed into console at the initialization.

### Actual results
Nothing is being printed into console at the initialization.

## Platform details
Latest version of prebid, nodejs version = 6.9

## Other information
$$PREBID_GLOBAL$$.logging === false && _loggingChecked === false
This condition always returns false at the initialization as $$PREBID_GLOBAL$$.logging is undefined.

matthewlane:
Thanks for the well filled out bug report, I'm working on reproducing the issue. I'm able to see the expected error message by setting `pbjs.logging = true;` before calling `pbjs.requestBids`, do you have this on your test page? Are you seeing any other error messages in the console later, or is this the only message that isn't being printed?

dmitriyshashkin:
Setting pbjs.logging = true on test page resolves the problem, but instead I add "pbjs_debug=true" to page url, it's more convenient as I can use it on production as well. 

I do see other error messages in the console later. pbjs.logging is initialized here https://github.com/prebid/Prebid.js/blob/master/src/prebid.js#L58
but utils.logError sometimes is being called earlier in the initialization sequence, eg here: 
https://github.com/prebid/Prebid.js/blob/master/src/adaptermanager.js#L117
After pbjs.logging has been initialized the problem disappears.

PR #1135, opened 2017-04-18
Feature/error shows proper line

mkendall07:
## Type of change
- [x] Feature
## Description of change
Been wanting to make this change forever, but this is to show proper line number/call stack when an error happens (caught or uncaught) via `console.error`.

Before :( 

image

After (YAY!!!)
snip20170418_13

Also fixes (https://github.com/prebid/Prebid.js/issues/949) - which is an actual issue. 

## Other information
@prebid/core


mkendall07:
this is buggy - going to fix and resubmit.

PR #1140, opened 2017-04-19
Widespace adapter validate size fix

jonasmattsson1:
## Type of change

- [ ] Bugfix

## Description of change

- A bugfix for Widespace adapter where we was validating the bid sizes wrong. Causing that we rejected bid responses that was not in a square format.
After this fix we can deliver panorama ad´s etc.

jonas.mattsson@widespace.com


patriknyblad:
@jonasmattsson1 is not available for the moment, you can contact me instead if there are any questions: patrik.nyblad@widespace.com

PR #1141, opened 2017-04-19
Sonobi Adapter - Enable size overrides

Studnicky:
## Type of change
- [x] Feature

## Description of change
Enable optional size overrides on a per-slot basis

### DIFF
--- a/src/adapters/sonobi.js
+++ b/src/adapters/sonobi.js
@@ -12,7 +12,7 @@ var SonobiAdapter = function SonobiAdapter(){
     var adSlots = request.bids || [];
     var bidderRequestId = request.bidderRequestId;
     var ref = (window.frameElement) ? '&ref=' + encodeURI(top.location.host || document.referrer) : '';
-    adloader.loadScript(trinity + JSON.stringify(_keymaker(adSlots)) + '&cv=' + _operator(bidderRequestId) + ref );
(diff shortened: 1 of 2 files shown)


dbemiller:
Hey, the code in general looks good to me. I'd recommend a more thorough unit test, though... the ones you have don't really test the content of the changed code at all.

Can you give me some context into why you're adding this, though? An ad's size is it's size... so I'm not sure why sonobi would want to see a different size than the _actual_ one. I'm wondering if there's some use-case not considered by the prebid core code which you're hacking your way around.

Studnicky:
I agree, however, the use case goes as follows:
- Publisher implements our adapter on a page with a slot defined as [[970, 250], [970, 90], [728, 90]]
- Publisher calls us and asks why they are getting 970x250 when they only want 728x90
- We explain to publisher that they can just change the slot definition, because sizes are sizes
- Publisher refuses to do so but insists we fix it without changing your slot definitions
- Well, ok dude, now you can limit our bids to only return 728x90 without changing slot definitions

dbemiller:
ok.. thanks for the background!

I do suspect there's a more general feature request buried in here... but this'll work for now. We might be able to incorporate something into the prebid 1.0 API, when we're ready to make breaking changes, so that all the adapters won't have to implement it individually.

PR #1142, opened 2017-04-20
Update polyfills to use core-js.

### LINKED ISSUE #1125: File size increase from prebid 0.18 to prebid 0.21.
## Type of issue

Large file size increase from prebid 0.18 to prebid 0.21. Looking for ways to minimize the file size of prebid.js

## Description

I've isolated the cause to #962 which increased the size of prebid minified built without adapters from 40k to 55k. 

## Steps to reproduce

```
gulp build --adapters adapters-none.json
# view file size
# remove imports in polyfill and try again.
```

## Platform details
prebid 0.21

nanek:
## Type of change

## Description of change

Reduces file size by 9K. Fixes #1125. From https://babeljs.io/docs/usage/polyfill/ which recommends https://github.com/zloirock/core-js#commonjs

I'd prefer to revert #962 over this, as I can't reproduce the IE error. I think they were basically the standard polyfill from https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/find

When do you run IE tests? Is travis-ci doing this on each PR? Is there coverage for this?

### DIFF
--- a/yarn.lock
+++ b/yarn.lock
@@ -828,7 +828,7 @@ babel-polyfill@^6.13.0:
     core-js "^2.4.0"
     regenerator-runtime "^0.10.0"
 
-babel-preset-es2015@^6.5.0:
+babel-preset-es2015@^6.24.1:
   version "6.24.1"
   resolved "https://registry.yarnpkg.com/babel-preset-es2015/-/babel-preset-es2015-6.24.1.tgz#d44050d6bc2c9feea702aaf38d727a0210538939"
   dependencies:
@@ -1857,7 +1857,7 @@ d@1:
   dependencies:
     es5-ext "^0.10.9"
 
-"dargs@github:christian-bromann/dargs":
+dargs@christian-bromann/dargs:
   version "4.0.1"
   resolved "https://codeload.github.com/christian-bromann/dargs/tar.gz/7d6d4164a7c4106dbd14ef39ed8d95b7b5e9b770"
   dependencies:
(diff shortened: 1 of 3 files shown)


protonate:
We currently run IE tests during release prep but are working on setting up Travis-ci and Browserstack integration for builds.

The polyfills used previously are from MDN.

As it happens I'm testing IE now and we're also breaking on `for ... of` loops as Babel uses `Symbol.iterator`. 

Would you update this PR to `require('core-js/fn/symbol/iterator')`?

mkendall07:
We've had some nagging issues with unit tests not playing nicely in all IE versions so we've disabled IE tests with Travis-ci for now. but we want to get them added and working soon.

nanek:
`require('core-js/fn/symbol/iterator')` adds another 4K. Can we use forEach instead of polyfilling `for...of`? 

I could only find one use of it:
https://github.com/prebid/Prebid.js/blob/b33a229785aeb7f82c2c1402ad9f89aa11f6719e/src/adapters/widespace.js#L48

yarn.lock
```diff
@@ -828,7 +828,7 @@ babel-polyfill@^6.13.0:
     core-js "^2.4.0"
     regenerator-runtime "^0.10.0"
 
-babel-preset-es2015@^6.5.0:
```
mkendall07: can you revert these changes? I don't think they relate to the polyfill

PR #1146, opened 2017-04-20
Yieldbot adapter - multiple requestBids per pageview

### LINKED ISSUE #481: Yieldbot: Multiple requestBids Not Working
So we currently split our ad auctions up amongst our ATF and BTF units and run multiple Prebid auctions. Unfortunately, the Yieldbot adaptor does not work with this.

Easy to replicate. Define two ad units. Run requestBids twice - once the first ad unit and a separate one on the second . Then run yieldbot.dumpLog() in the console.

You'll see that only the initial ad slot that is defined gets a call to getSlotCriteria() with the handleUpdateState() is run. You'll only get bid responses on the initial slots.

elljoh:
## Type of change
- [x] Feature

## Description of change
In the Yieldbot adapter, adds feature to allow requests for further Yieldbot bids after the first call to `requestBids()`.

## Other information
Fixes #481

### DIFF
--- a/src/adapters/yieldbot.js
+++ b/src/adapters/yieldbot.js
@@ -88,6 +88,7 @@ var YieldbotAdapter = function YieldbotAdapter() {
       ybotq.push(function () {
         var yieldbot = window.yieldbot;
 
+        ybotlib.definedSlots = [];
         utils._each(bids, function (v) {
           var bid = v;
           var psn = bid.params && bid.params.psn || 'ERROR_DEFINE_YB_PSN';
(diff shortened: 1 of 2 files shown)


aneuway2:
@elljoh if I understand correctly, does this PR allow more than 1 size to be passing to the Yieldbot adaptor to receive bids, or if this for something else?

Thanks!

elljoh:
@aneuway2 
The Yieldbot adapter has always supported requesting multiple AdUnit and/or sizes of AdUnit creatives in a request: the nature of this PR is slightly different.

This PR is related to calls to refresh bids on the same page view browsing context i.e. any bid request subsequent to the initial `requestBids()` call.

Prior to this PR, a "refresh" call to `requestBids()` after the initial request for bids, that included Prebid config for the Yieldbot adapter, would _**not**_ result in a request for bids from Yieldbot.

- _Reason being_, that the equivalent of a "_refresh_" call needed to be made in the Yieldbot adapter. If not, the underlying adapter library would exit early.

With this change, calls to `refreshBids()` where the Yieldbot adapter is configured will result in requests for bids being made.

aneuway2:
Ah gotcha. Thanks for the context! Very helpful!

PR #1151, opened 2017-04-24
Add PubWise Analytics

GLStephen:
## Type of change
- [ X ] Other - Analytics Adapter

## Description of change
This is the analytics adapter for PubWise.io

- stephenj@pubwise.io
- [ x ] official adapter submission

### DIFF
--- a/src/adapters/analytics/pubwiseanalytics.js
+++ b/src/adapters/analytics/pubwiseanalytics.js
@@ -1,6 +1,5 @@
 import {ajax} from 'src/ajax';
 import adapter from 'AnalyticsAdapter';
-import CONSTANTS from 'src/constants.json';
 const utils = require('../../utils');
 
 /****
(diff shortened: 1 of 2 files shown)


GLStephen:
The build failure looks unrelated.

src/adapters/analytics/pubwiseanalytics.js
```diff
@@ -0,0 +1,66 @@
...
+    track({eventType, args}) {
+      /*
+       The args object is not always available, in addition neither is the config object
+       it is available on the first call and we can setup our config. Potential additional
+       PR for later, but this solves this for now.
+       */
+      if (args !== undefined && args.config !== undefined) {
+        target_site = args.config.site;
+      }
+      if (args !== undefined && args.config !== undefined) {
```
matthewlane: This line is the same as line 43, is that on purpose? When running tests on this PR there are a few `WARN [web-server]: 404: /undefined` warnings in the console, which may be coming from line 47. Adding ` && args.config.endpoint` before the closing `)` on 46 stops those warnings, but I may be misunderstanding the intention here

src/adapters/analytics/pubwiseanalytics.js
```diff
@@ -0,0 +1,66 @@
...
+        })
+      )
+        ;
+      }
+    }
+  });
+
+export default pubwiseAnalytics;
+
+
```
GLStephen: I'll remove that when I fix the above.

src/adapters/analytics/pubwiseanalytics.js
GLStephen: The config, and sometimes args, is not available in future calls of _track after the initial call of enable analytics. Overriding/extending enable led to other issues specifically multiple analytics adapters would no longer load. Since I have not been able to fix either of those other issues globally this was my compromise. I will clean up the logic here and look into the WARN [web-server]: 404: /undefined issue.

GLStephen:
I have updated the analytics adapter based on feedback as well as removed a bit of event filtering since we've updated the endpoint to handle determination of event handling at the API endpoint.

GLStephen:
Downloader addition is here https://github.com/prebid/prebid.github.io/pull/242 Thanks!

GLStephen:
@matthewlane @protonate Any idea what version this might land in? 24? Just trying to plan ahead. Thanks!

mkendall07:
LGTM.

PR #982, opened 2017-02-10
E2etest concurrency

jaiminpanchal27:
## Type of change

- [x ] Bugfix

## Description of change
End 2 End testing using nightwatch and browserstack was having issues running in parallel. With this fix we are now running only 2 browsers at a time.  To test this following steps must be followed 
- Start tunneling between browserstack and local machine
./BrowserStackLocal --key \<your-browserstack-key\> --force-local 
- 'gulp serve-nw --browserstack' to start the running the tests
- Once finished, execute 'gulp e2etest-report' to view the report in browser.

Issue #891, opened 2016-12-22
Prebid 1.0.0 Proposal - intent to implement


mkendall07:
# Updates
- **February 6, 2017**
  - Clarification on requirements/ distinction between *requirements* and *recommendations*
  - **Acknowledgements**: Thanks to @bretg and the Rubicon team for the feedback here.
- **Oct 9th, 2017**
  - Moved currency into required field for bidders. 

# Type of issue
Intent to implement.  Open for comment. 

# Background and Motivation

To this point, Prebid.js releases have been iterating on major version 0 (as of the time of writing, the latest Prebid.js release is 0.18.0).  Given the maturity and widespread adoption of Prebid.js, we are approaching the point at which we should start pushing towards release of version 1.0.0.  

In traditional semantic software versioning, version 0.X.X implies beta software in which large public API changes should be expected.  Version 1.X.X implies stabilization of public APIs (i.e. major breaking changes should no longer be expected).

The release of Prebid.js version 1.0.0 gives us the opportunity to take a firmer stance on adaptor standardization and performance, on which Prebid has been intentionally lax.  To this effect, the requirements and recommendations outlined below represent a fundamental paradigm shift for Prebid.js, whereby we decide that regulation of adapter behavior, security, and page performance are in-scope.

We encourage the community to review the following proposal and offer comments and suggestions to further shape this document. 

# Prebid 1.0.0 Summary
Based on frequently recurring suggestions from the Prebid.js user base, the Prebid core team has identified the following high-level themes on which we aim to focus with the release of Prebid.js version 1.0.0:

 1. Formalize adapter requirements and recommendations, and standardize adapter behavior
 2. Reduce Prebid's impact to page performance and latency
 3. Improve page security
 
### Core
- An **auction** shall be defined as the result of a single call made to `requestBids` for one or more adUnits.
- An **impression** shall be defined as the inclusion of a single adUnit within a single **auction**.
- Each Prebid **auction** will generate a unique auction ID.
- Each Prebid **impression** will generate a unique impression ID.  As per the OpenRTB 2.5 definition of the `Source` object, this impression ID may be shared across multiple exchanges.
- Prebid shall be able to run multiple **auctions** concurrently, and shall be able to match all adapter bid responses with their corresponding bid requests.
- Prebid may choose to create multiple instances of any adapter.
- Prebid will provide a standard OpenRTB adapter.
- Prebid reserves the right to monitor and report on the performance of adapters.
[...]

PR #1152, opened 2017-04-24
Add PubWise Bid Simulator

GLStephen:
## Type of change
- [X ] New bidder adapter   

## Description of change
Addition of PubWise Bid Simulator Adapter. The Bid Simulator is part of the testing package we use internally for PubWise.io. We are making it publicly available because we feel the Prebid community will get some value from it.

```
{
        bidderCode: 'pubwiseSim',
        bids: [
          {
            bidId: 'testId1',
            bidder: 'pubwiseSim',
            params: {
              site_id: 'test-test-test-test',
              height: '728',
              width: '90',
              cpm: 20,
              delay: 0,
            },
            sizes: [[160, 600]],
            placementCode: 'div-gpt-test-1'
          }
        ]
      }
```
- stephenj@pubwise.io
- [ x ] official adapter submission

### DIFF
--- a/adapters.json
+++ b/adapters.json
@@ -64,6 +64,7 @@
     "atomx",
     "tapsense",
     "trion",
+    "pubwiseSim",
     {
       "appnexus": {
         "alias": "brealtime"
--- a/src/adapters/pubwiseSim.js
+++ b/src/adapters/pubwiseSim.js
@@ -1,0 +1,126 @@
+var CONSTANTS = require('../constants.json');
+var utils = require('../utils.js');
+var bidfactory = require('../bidfactory.js');
+var bidmanager = require('../bidmanager.js');
+var adloader = require('../adloader');
+
+/******
+ * PubWise.io Bid Simulator Adapter
+ * Contact: support@pubwise.io
+ * Developer: Stephen Johnston
+ ***/
+
+var PubwiseSimAdapter = function PubwiseSimAdapter() {
+  var pubwiseUrl = "adtest.pubwise.io/api/v1/bid/get";
+
+  $$PREBID_GLOBAL$$.pubwiseSimResponseHandler = pubwiseSimResponseHandler;
+
+  return {
+    callBids: _callBids
+  };
+
+  function _callBids(bidReqs) {
+    utils.logInfo('pubwise callBids adapter beginning');
+
+    var domain = window.location.host;
+    var page = window.location.pathname + location.search + location.hash;
+
+    var pubwiseBidReqs = {
+      id: utils.getUniqueIdentifierStr(),
+      bids: bidReqs,
+      site: {
+        domain: domain,
+        page: page
+      }
+    };
+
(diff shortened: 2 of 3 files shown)


GLStephen:
I have resolved the conflict related to the list of active adapters.

mkendall07:
@GLStephen 
Could you explain the purpose of this? Just not sure is all.. thanks

GLStephen:
@mkendall07 We use it, similar to an echo server, to perform automated testing in our test environment. It allows you to send "real" not code level mocked bid requests into your Prebid setup and test end-to-end configuration, analytics setup and DFP integration and timeout behavior.  I thought it would be useful to the Prebid community in general.

I really appreciate the value of unit testing at a code level in Prebid.js, but often it is insufficient for some types of testing. For instance, if you want to verify interaction of DFP and non-header bidding line items you can either use a test tag from a demand partner, wait for a bid of some value to happen by, or you can use PubWise Sim. 

In the end, PubWise Sim provides significantly more control over the response than a test tag does and lets you introduce latency, and other options such as mismatched sizes. In addition, you can set variable return pricing without having to twiddle an adunit ID (like a test tag likely would have you do) which lets you through code control testing with more ease.

GLStephen:
Documentation is here: https://github.com/prebid/prebid.github.io/pull/241

bretg:
@GLStephen - this is good background, and looks like a useful tool. I'd suggest adding some of the words from above into the prebid.github.io doc PR. Will make a specific suggestion over there.

jaiminpanchal27:
@GLStephen I am getting 404 trying to hit your endpoint. Test page https://jsfiddle.net/61ak300m/

GLStephen:
@jaiminpanchal27 endpoint is up now, sorry about that, it had not been made live on the production farm

GLStephen:
@bretg updates have been made to the documentation piece

jaiminpanchal27:
@GLStephen I checked this and it works fine. This is good testing utility but i don't think it should be developed as an adapter. Also no test related code should be inside `./src`

GLStephen:
@jaiminpanchal27 I understand your concern. It's one of those "isn't exactly like, but must behave like to properly test" sorts of things. In some ways, it is accurate that this is a test, but the current testing in ./test is more like unit tests, not outside in testing. This is more of a black box outside in interaction with controls vs. a whitebox unit test. Is there an alternate place for this sort of thing that can provide the same level of testing interaction, particularly pre-prod style integration testing which is the focus here?

jaiminpanchal27:
@GLStephen Yes our integration level test is not upto the mark yet. We have tried this using nightwatch and browserstack. You can check this PR https://github.com/prebid/Prebid.js/pull/982 for end to end testing.
Integration testing is also in discussion for Prebid 1.0. Add your thoughts here https://github.com/prebid/Prebid.js/issues/891

mkendall07:
@GLStephen 
Appreciate the feedback on integration test with a focus on the publisher. I do believe we need this, but we don't want it as part of the codebase here. We are talking about building out a toolset to help with this but it's in the early stages.

PR #1156, opened 2017-04-25
Improvement/alias queue

### LINKED ISSUE #1124: ¿Que?
## Type of issue
Naming/API

## Description
Trivial gripe, but are we interested in renaming/aliasing the `pbjs.que` to `pbjs.queue`?

see: https://www.merriam-webster.com/dictionary/Que

dbemiller:
## Type of change
[x] Feature

## Description of change
This fixes #1124 (Deprecate`que` in favor of `queue`)


snapwich:
this isn't an alias, this is a replacement.

protonate:
ah whoops yep gotta add the alias too, can't break the api, nice catch @snapwich

mkendall07:
Isn't the alias defined in `prebidGlobal.js` ?

snapwich:
@mkendall07 

```
window.$$PREBID_GLOBAL$$.que = window.$$PREBID_GLOBAL$$.que || window.$$PREBID_GLOBAL$$.queue;
```

won't work since all the operations have been renamed to operate on `pbjs.queue`.

the operations should probably be unchanged to still operate on `pbjs.que` and then the alias should be something like
```
window.$$PREBID_GLOBAL$$.que = window.$$PREBID_GLOBAL$$.que || window.$$PREBID_GLOBAL$$.queue || [];
```

snapwich:
forgot one thing...  both `pbjs.que.push` and `pbjs.queue.push` need to be monkey-patched.

dbemiller:
Thanks guys. I didn't fully understand what the queue was trying to accomplish, before making those changes. I definitely don't intend to break backwards compatibility.

The latest changes split `que` from `queue`, log warnings for people who are using the deprecated method, and add a bunch of unit tests.

src/prebid.js
```diff
@@ -70,35 +70,61 @@ utils.logInfo('Prebid.js v$prebid.version$ loaded');
...
-  for (var i = 0; i < $$PREBID_GLOBAL$$.que.length; i++) {
-    if (typeof $$PREBID_GLOBAL$$.que[i].called === objectType_undefined) {
+function processQueue(queue, deprecationWarning) {
+  queue.forEach(function(cmd) {
+    if (typeof cmd.called === objectType_undefined) {
       try {
-        $$PREBID_GLOBAL$$.que[i].call();
-        $$PREBID_GLOBAL$$.que[i].called = true;
+        if (deprecationWarning) {
+          utils.logWarn('$$PREBID_GLOBAL$$.que.push is deprecated, to be removed in v1.0.0. Use $$PREBID_GLOBAL.queue.push instead.');
```
mkendall07: This is going to spam the console with deprecation warnings for all consumers of prebid.js. Probably should only display it once.

dbemiller: I must admit... I had kind of intended to spam them :p. The more annoying it is, the faster people will want to update their code ^^.

I must admit there's no substantive benefit, though. I pulled the warning out of the loop.

nanek:
It is possible to just add an alias for `queue` as suggested in #1124 instead of deprecating `que`? Then this doesn't have the side effect of breaking all existing code for all customers, documentation and examples. As a user of prebid I prefer the original intent for the naming of being short as stated in #1124, and i'm not opposed to adding an alias if it helps others. However it does seem a little rushed to make `que` deprecated and a breaking change for all customers without a good benefit.

dbemiller:
So... it sounds like there are two issues here.

----------------------------------------------
1. Should Prebid as a project use short names, or descriptive ones?

I'm very new to the project, so if I've missed decisions about [1], please point me to them. In isolation, though, I would argue for descriptive names in APIs for several reasons:

- They make your code read like English, which makes it easier to understand
- They minimize the barrier to entry, since newcomers don't have to learn your acronyms or find your documentation to understand your example code
- They're self-describing, so that "occasional" Prebid programmers who come back to their code a month or two later will remember more easily what the function does.

Let me know if you have more reasons than these... but the only benefits I can come up with in favor of shorter names are:

- They have a (**very** tiny) performance benefit
- They require fewer keystrokes for people code in an environment which doesn't autocomplete.

I definitely don't want to upset the project philosophy if you've already had this discussion... but in a vacuum, descriptive names have always seemed more beneficial to me. Truth be told, even `queue` isn't totally descriptive... but it's clearer than que :).

------------------------------------------
2. How much community disruption are these changes worth?

I simply don't know the Prebid community well enough yet to judge this, for such a small change.

In general, though, API changes [should be expected](http://semver.org/#spec-item-4) in a v0 project.

mkendall07:
@nanek 
This wouldn't be a breaking change until 1.0.

snapwich:
As I mentioned in another thread, if we're going to make backwards-compatible breaking changes, I'd recommend `cmd`.  It's a common term.  It's still short.  It's what google uses so you get the shared education on the terminology.

dbemiller:
We chatted a bit about it here and agreed that it was probably not a big enough API change to bother bugging publishers with deprecation warnings.

So... this should be ready for another look now.

mkendall07:
@prebid/all-contributors 
Please vote on renaming the command queue in prebid.js:
1) pbjs.cmd
2) pbjs.queue

Use 👍  or 👎  in the following posts to vote. Thanks

mkendall07:
1) pbjs.cmd

mkendall07:
2) pbjs.queue

dbemiller:
Looks like there's a pretty clear winner ^^. Renamed to cmd now.

PR #1161, opened 2017-04-27
Force cpm to be a number

Yann-Pravo:
## Type of change

- [x] Bugfix
- [ ] Feature
- [ ] New bidder adapter   
- [ ] Build related changes
- [ ] CI related changes
- [ ] Other

## Description of change

Some platforms return a string instead of a float for the cpm. This change just make sure that the cpm is a number.

```
{
  bidder: '<bidder name>',
  params: {
    // ...
  }
}
```

### DIFF
--- a/src/bidmanager.js
+++ b/src/bidmanager.js
@@ -105,7 +105,7 @@ exports.addBidResponse = function (adUnitCode, bid) {
       requestId: requestId,
       responseTimestamp: timestamp(),
       requestTimestamp: start,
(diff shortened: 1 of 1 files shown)


mkendall07:
This is a potentially breaking change. While I agree it should be in place, I think we might want to wait on 1.0 release.

mkendall07:
On 2nd thought, since pubs are having to deal with both string and number today this really shouldn't be a breaking change.

PR #1165, opened 2017-04-28
Prebid Server to Server

jaiminpanchal27:
## Type of change
- [x] Feature

## Description of change
Allows a publisher to send requests to Prebid Server. To enable, use the following API:

 ```javascript
 pbjs.setS2SConfig({
    accountId : '1',                     //string:required: account ID obtained in step 1 
    enabled : true,                      //boolean:required: enables s2s - default false
    bidders : ['appnexus','pubmatic'],   //array[string]:required: of bidder codes to enable S2S. Note these must have been including in the Prebid.js build from Step 2. 
    timeout : 1000,                      //number:optional timeout for bidders called via the S2S endpoint. default value is 1000 
    adapter : 'prebidServer',                     //string:optional: Adapter code for S2S code. Defaults to 'prebidServer'
    endpoint : 'http://prebid.adnxs.com' //string:optional: will override default endpoint for prebid server. 
});
```

## Other information
@prebid/core for review
Will add some more unit tests soon.

### CODE REVIEW

File: src/adapters/prebidServer.js [hunk 1]
```diff
@@ -0,0 +1,121 @@
+import Adapter from 'src/adapters/adapter';
+import bidfactory from 'src/bidfactory';
+import bidmanager from 'src/bidmanager';
+import * as utils from 'src/utils';
+import { ajax } from 'src/ajax';
+import { STATUS } from 'src/constants';
+import { queueSync, persist } from 'src/cookie.js';
+
+const TYPE = 's2s';
+const cookiePersistMessage = `Your browser may be blocking 3rd party cookies. By clicking on this page you allow Prebid Server and other advertising partners to place cookies to help us advertise. You can opt out of their cookies <a href="https://www.appnexus.com/en/company/platform-privacy-policy#choices" target="_blank">here</a>.`;
```
protonate: We should look at how to make this configurable by pubs.

src/adaptermanager.js
```diff
@@ -31,7 +33,7 @@ function getBids({bidderCode, requestId, bidderRequestId, adUnits}) {
           mediaType: adUnit.mediaType,
           transactionId : adUnit.transactionId,
           sizes: sizes,
-          bidId: utils.getUniqueIdentifierStr(),
+          bidId: bid.bid_id || utils.getUniqueIdentifierStr(),
```
mkendall07: It's actually just that, a `bid_id`

src/adapters/prebidServer.js
mkendall07: It's part of using prebid server for now. So opt out would be don't use prebid server.

PR #1111, opened 2017-04-07
Upgrade linters to ESLint with stricter code style

matthewlane:
## Type of change
- Build related changes

## Description of change
- Updates the quality/linting system from ([JSHint](http://jshint.com/) && [JSCS](http://jscs.info/)) to [ESLint](http://eslint.org/)
- Updates linting system to use a stricter code style

## Other information
This PR is currently configured to use [JavaScript Standard Style](https://standardjs.com/) with semicolons. Compare this to other common linting configurations with number of linting problems found against the `0.21.0` release, and the number after running [`eslint --fix src`](http://eslint.org/docs/user-guide/command-line-interface#--fix):

| config | # of problems | # after `--fix` |
| :--- | :--- | :--- |
| [eslint:recommended](http://eslint.org/docs/rules/) | 121 errors, 0 warnings | 121 errors, 0 warnings |
| [Standard](https://www.npmjs.com/package/eslint-config-standard) (with semicolons) | 2254 errors, 0 warnings | 321 errors, 0 warnings |
| [Airbnb](https://www.npmjs.com/package/eslint-config-airbnb) | 7214 errors, 376 warnings | 1897 errors, 257 warnings |

Other configs, such as [Canonical](https://www.npmjs.com/package/eslint-config-canonical) mentioned in #776, or a custom rule set are possible.

Compare rule sets with https://sqren.github.io/eslint-compare/

Requesting feedback from the community and @prebid/core  on which code style Prebid.js should use.

PR #1169, opened 2017-05-02
E-Planning adapter

sebaperez:
## Type of change

- [ ] Bugfix
- [ ] Feature
- [x] New bidder adapter   
- [ ] Build related changes
- [ ] CI related changes
- [ ] Other

## Description of change

```
{
  bidder: 'eplanning',
  params: {
    ci: '18f66',
    t: '1'
  }
}
```
soporte@e-planning.net

### DIFF
--- a/src/adapters/eplanning.js
+++ b/src/adapters/eplanning.js
@@ -1,51 +1,52 @@
 var bidfactory = require('src/bidfactory.js');
 var bidmanager = require('src/bidmanager.js');
 var adloader = require('src/adloader.js');
-var utils = require('src/utils.js');
 
 var EPlanningAdapter = function EPlanningAdapter() {
 
-	var ISV_DEFAULT = "aklc.img.e-planning.net";
+  var ISV_DEFAULT = "aklc.img.e-planning.net";
 
-	window.pbjs = window.pbjs || {};
-	window.pbjs.processEPlanningResponse = function(response) {
-		var bids, bidObject, i;
-		if (response) {
-			bids = response.bids;
-			for (i = 0; i < bids.length; i++) {
-				if (bids[i].ad) {
-					bidObject = getBidObject(bids[i]);
-					bidmanager.addBidResponse(bids[i].placementCode, bidObject);
-				} else {
-					bidObject = bidfactory.createBid(2);
-					bidObject.bidderCode = 'eplanning';
-					bidmanager.addBidResponse(bids[i].placementCode, bidObject);
-				}
-			}
-		}
-	};
+  window.pbjs = window.pbjs || {};
+  window.pbjs.processEPlanningResponse = function(response) {
+    var bids, bidObject, i;
+    if (response) {
+      bids = response.bids;
+      for (i = 0; i < bids.length; i++) {
+        if (bids[i].ad) {
+          bidObject = getBidObject(bids[i]);
+          bidmanager.addBidResponse(bids[i].placementCode, bidObject);
+        } else {
+          bidObject = bidfactory.createBid(2);
+          bidObject.bidderCode = 'eplanning';
+          bidmanager.addBidResponse(bids[i].placementCode, bidObject);
+        }
+      }
+    }
+  };
 
-	function getBidObject(bid) {
-		var bidObject = bidfactory.createBid(1), i;
-		bidObject.bidderCode = 'eplanning';
-		for (i in bid.ad) {
-			if (bid.ad.hasOwnProperty(i)) {
-				bidObject[i] = bid.ad[i];
-			}
-		}
-		return bidObject;
-	}
+  function getBidObject(bid) {
+    var bidObject = bidfactory.createBid(1), i;
+    bidObject.bidderCode = 'eplanning';
+    for (i in bid.ad) {
+      if (bid.ad.hasOwnProperty(i)) {
+        bidObject[i] = bid.ad[i];
+      }
+    }
+    return bidObject;
+  }
 
-	function _callBids(params) {
-		var scriptURL = "http://" + ((params && params.bids && params.bids.length && params.bids[0].params && params.bids[0].params.isv) ? params.bids[0].params.isv : ISV_DEFAULT) + "/layers/hbpb.js";
-		adloader.loadScript(scriptURL, function() {
-			hbpb.call(params);
-		}, true);
-	}
+  function _callBids(params) {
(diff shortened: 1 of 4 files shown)


dbemiller:
Bid doesn't show up. Here's a test page you can use.

```
<html>
<head>
    <!-- Prebid Config Section START -->
    <!-- Make sure this is inserted before your GPT tag -->
    <script>
      var PREBID_TIMEOUT = 700;

      var adUnits = [{
        bids: [{
          bidder: 'eplanning',
          params: {
            ci: '18f66',
            t: '1'
          }
        }]
      }];

      var pbjs = pbjs || {};
      pbjs.que = pbjs.que || [];

    </script>
    <!-- Prebid Config Section END -->

    <!-- Prebid Boilerplate Section START. No Need to Edit. -->
    <script type="text/javascript" src="../../build/dist/prebid.js" async></script>
    <script>
      });

      pbjs.que.push(function() {
        pbjs.addAdUnits(adUnits);
        pbjs.requestBids({
        });
      });

          pbjs.que.push(function() {
            googletag.pubads().refresh();
          });
        });
      }

      setTimeout(function() {
        sendAdserverRequest();
      }, PREBID_TIMEOUT);

    </script>
    <!-- Prebid Boilerplate Section END -->

    <script>
      (function () {
        var gads = document.createElement('script');
        gads.async = true;
        gads.type = 'text/javascript';
        var useSSL = 'https:' == document.location.protocol;
        gads.src = (useSSL ? 'https:' : 'http:') +
          '//www.googletagservices.com/tag/js/gpt.js';
        var node = document.getElementsByTagName('script')[0];
        node.parentNode.insertBefore(gads, node);
      })();
    </script>

    <script>

        googletag.enableServices();
      });
    </script>
</head>

<body>
<h2>Prebid.js Test</h2>
<h5>Div-1</h5>
    </script>
</div>
</body>
</html>
```

sebaperez:
Hi @dbemiller - thanks for your help!
It was my mistake, now it is fixed. I tested it using the html you provided and everything went OK.

src/adapters/eplanning.js
```diff
@@ -0,0 +1,53 @@
...
+    }
+    return bidObject;
+  }
+
+  function _callBids(params) {
+    var scriptURL = "http://" + ((params && params.bids && params.bids.length && params.bids[0].params && params.bids[0].params.isv) ? params.bids[0].params.isv : ISV_DEFAULT) + "/layers/hbpb.js";
+    adloader.loadScript(scriptURL, function() {
+      if (window.hbpb) {
+        window.hbpb.call(params);
+      }
```
dbemiller: @mkendall07 or @protonate: How do you feel about prebid referencing global variables from other scripts?

This is obviously a really simple code block, but... I'm nervous about the coupling it creates. For example, if the loaded script set `window.hpbp = 5`, then this block would throw an error... and it wouldn't really be clear to me which script was "at fault" for it.

@sebaperez Could you give me some background on what this is trying to accomplish? There might be some way to accomplish the same goal without introducing cross-script state.

sebaperez: @dbemiller yes, the adapter loads an external script hosted by us which do the request to the server and get the bids. This script initilializes with the hbpb.call() function, that's why it is called in the callback. It is true that it could be faulty if someone creates a global variable with the same name, but not less faulty than someone creating a window.googletag = function() { return 1; }, which could happen too.
This code is in an external script as we would like to give this level of features to our publishers. They are the ones that will have a "ci" parameter (clientId) and we are going to bid for, so it won't involve other publishers who do not have an agreement with us. Our bidder (as many others) can't be used if you do not have a client identifier. We provide a 24/7 tech support to all of our publishers to solve this kind of trouble (if it happens).

test/spec/adapters/eplanning_spec.js
```diff
@@ -0,0 +1,95 @@
...
+		pbjs.processEPlanningResponse(RESPONSE_EMPTY);
+
+		var bidPlacementCode = stubAddBidResponse.getCall(0).args[0];
+		var bidObject = stubAddBidResponse.getCall(0).args[1];
+
+		expect(bidPlacementCode).to.equal('div-gpt-ad-1460505748561-0');
+		expect(bidObject.getStatusCode()).to.equal(2);
+		expect(bidObject.bidderCode).to.equal('eplanning');
+
+		stubAddBidResponse.restore();
```
dbemiller: These `restore()` and `stub()` calls should be in `beforeEach` and `afterEach` blocks.

Otherwise one test failure will cacade into others, because the restore() at the end won't get called.

src/adapters/eplanning.js
```diff
@@ -0,0 +1,53 @@
+var bidfactory = require('src/bidfactory.js');
+var bidmanager = require('src/bidmanager.js');
+var adloader = require('src/adloader.js');
+
+var EPlanningAdapter = function EPlanningAdapter() {
+
+  var ISV_DEFAULT = "aklc.img.e-planning.net";
+
+  window.pbjs = window.pbjs || {};
+  window.pbjs.processEPlanningResponse = function(response) {
```
sebaperez: Yes, it is on global scope as we do not use AJAX for server calling, we use JSONP. Then we need a global scoped function for catching the bid response.

src/adapters/eplanning.js
dbemiller: Found this line buried in the [pr_review.md](https://github.com/prebid/Prebid.js/blob/master/pr_review.md#new-adapter-or-updates-to-adapter-process):

> Verify that the bidder is being as efficient as possible, ideally not loading an external library, however if they do load a library it should be cached.

So it looks like the project has generally been ok with this. Feel free to ignore it.

mkendall07: It's worth pointing out that this behavior is not going to be allowed in [prebid 1.0](https://github.com/prebid/Prebid.js/issues/891). Thanks

sebaperez:
@dbemiller I modified the code as you requested. Now there is not external file, all the code is in the adapter file.

dbemiller:
@sebaperez The linting rules were overhauled in #1111. It's baked into the CI... so before this is mergeable you'll have to pull master and fix the errors. `eslint --fix` should take care of most of them.

sebaperez:
@dbemiller sorry, I deleted the repo by accident. Is it OK if I create a new PR for this branch but rebased?
BTW, I ran eslint and it didn't show any error or warning about the code

dbemiller:
@sebaperez Yeah, a new rebased branch sounds good.

PR #1170, opened 2017-05-02
Prevent 'about:' protocol in _publisherTagUrl

tr-coull:
## Type of change

- [x] Bugfix

## Description of change

Criteo adapter would pick up 'about:' as window.location.protocol when in an iframe with 'about:blank'. This code specifically checks for 'http:' and uses that, or 'https:' if not found. A tighter solution would be to (try to) recursively check every window.location.protocol up to top and use 'https:' if found on the way - this commit is a less thorough but safer and smaller solution. I have not written tests as it is a small change.

- tom@coull.com


jaiminpanchal27:
@tr-coull Please resolve the conflict.
Conflicts occurred because we recently switched to ESlint with stricter style code. https://github.com/prebid/Prebid.js/pull/1111

tr-coull:
Commit 1b4f21c resolves the issue.
```
Prebid.js/src/adapters/criteo.js
6:40  error  Infix operators must be spaced  space-infix-ops
✖ 1 problem (1 error, 0 warnings)
[10:53:33] 'lint' errored after 7.15 s
[10:53:33] ESLintError in plugin 'gulp-eslint'
```

mkendall07:
@tr-coull 
please resolve the conflict as well so we can merge. Thanks

tr-coull:
@mkendall07 resolved and all tests pass now. We got there eventually ;)

PR #1171, opened 2017-05-03
Improve footer styling

matthewlane:
## Type of change
- Style update

## Description of change
Updates the cookie footer style to show the message text properly on mobile

### DIFF
--- a/src/cookie.js
+++ b/src/cookie.js
@@ -60,17 +60,27 @@ function displayFooter(msgHtml) {
 }
 
 function createFooter(msgHtml) {
-  var footer = document.createElement('div');
-  footer.style.background = '#ccc';
-  footer.style.color = '#333';
+  const footer = document.createElement('div');
+  footer.style.background = '#D3D3D3';
+  footer.style.color = '#555';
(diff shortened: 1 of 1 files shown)


src/cookie.js
```diff
@@ -60,17 +60,27 @@ function displayFooter(msgHtml) {
 }
 
 function createFooter(msgHtml) {
-  var footer = document.createElement('div');
-  footer.style.background = '#ccc';
-  footer.style.color = '#333';
+  let footer = document.createElement('div');
```
mkendall07: can change to `const`

src/cookie.js
```diff
@@ -60,17 +60,27 @@ function displayFooter(msgHtml) {
...
+  footer.style.left = '0';
+  footer.style.right = '0';
   footer.style.width = '100%';
-  footer.style.padding = '1em';
+
+  footer.style.padding = '1em 0';
   footer.style.zindex = '1000';
-  footer.innerHTML = msgHtml;
+
+  let footerText = document.createElement('p');
```
mkendall07: `const`

PR #1155, opened 2017-04-25
create adapter adyoulike

jbAdyoulike:
## Type of change

- [ ] Bugfix
- [ ] Feature
- [x] New bidder adapter   
- [ ] Code style update (formatting, local variables)
- [ ] Refactoring (no functional changes, no api changes)
- [ ] Build related changes
- [ ] CI related changes
- [ ] Other

## Description of change

- test parameters for validating bids
```
{
  bidder: 'adyoulike',
  params: {
     placement: "194f787b85c829fb8822cdaf1ae64435",
  }
}
```
-dev-ssp@adyoulike.com
- [x] official adapter submission

PR #1133, opened 2017-04-18
HIRO Media: Remove batching mechanism and use AJAX instead of JSONP

ronenst:
## Type of change
- [x] Refactoring (no functional changes, no api changes)

## Description of change
- Switched to AJAX instead of JSONP
- The mechanism that batched multiple requests according to specific keys has been removed. Each placement will now invoke a request.

These changes are related since the generated `batchKey` was also used as a request id in the JSONP responses. No such id is necessary with AJAX.

- contact email of the adapter’s maintainer
support@hiro-media.com

- [x] official adapter submission

PR #1182, opened 2017-05-05
Integration test page for adapters

dbemiller:
## Type of change
- [x] Other (documentation)

## Description of change
Adding a "hello world" page which can be used to test adapter PRs. The goal here is to prevent the need for comments like these:

- https://github.com/prebid/Prebid.js/pull/1169#issuecomment-299505297
- https://github.com/prebid/Prebid.js/pull/1133#issuecomment-298907880
- https://github.com/prebid/Prebid.js/pull/1155#issuecomment-298681125


dbemiller:
Also: If anyone wants to tell me which `integrationExamples` they don't find any value in anymore, I'm happy to delete them here.

PR #1183, opened 2017-05-08
Beachfront adapter - add ad unit size

bfmdeploy:
## Type of change

- [x] Feature 

## Description of change

Added ad unit size to bid request for the Beachfront adapter.


mkendall07:
@bfmdeploy 
we just updated code style, please rebase from master then we can merge. THanks

bfmdeploy:
@mkendall07
Hi, I fixed the lint errors and attempted a rebase from master because the build was failing.

src/adapters/beachfront.js
```diff
@@ -14,7 +14,7 @@ function BeachfrontAdapter() {
     const bids = bidRequests.bids || [];
     bids.forEach(function(bid) {
       var bidRequest = getBidRequest(bid);
-      var RTBDataParams = prepareAndSaveRTBRequestParams(bid);
+      var RTBDataParams = prepareAndSaveRTBRequestParams(bidRequest);
```
mkendall07: I think this should still be `bids` ?

bfmdeploy: I changed it to `bidRequest` because it has the parsed width and height properties.

mkendall07: ok but doesn't it just return here because `bid.params` doesn't exist?
https://github.com/bfmdeploy/Prebid.js/blob/420d921c0db56f06f87bb5004d3ba8dc3f1e4b13/src/adapters/beachfront.js#L46

bfmdeploy: `bid.params` should exist. I think the confusion here is that `bid` and `bidRequest` are referencing the same object. I can revert that line if it makes it more clear.

mkendall07: Oh right. Yeah that's a bit confusing... I would just make it bid to clear that up

mkendall07: I'll merge when you make the change

PR #1168, opened 2017-05-02
inclusion of popular Nordic ad sizes to default size list

surensilva:
## Type of change

- [ ] Bugfix
- [x] Feature
- [ ] New bidder adapter   
- [ ] Code style update (formatting, local variables)
- [ ] Refactoring (no functional changes, no api changes)
- [ ] Build related changes
- [ ] CI related changes
- [ ] Other

## Description of change
Addition of popular Nordic ad sizes to the Rubicon adapter

PR #1188, opened 2017-05-10
added size '38' : 930x180 to the rubicon adaptor

thomas15k:
## Type of change
- [x] Other

## Description of change
added size 930x180 to the rubicon adaptor

sizes are found in the bottom of this doc: https://support.rubiconproject.com/Solution?solId=50170000000c3LNAAY


bretg:
Thanks, but this overlaps with PR #1168 -- 930x180 is in the batch there ready for review.

Issue #1196, opened 2017-05-11
NoBid response from prebid server has no adUnitCode, addBitResponse discards it


dmitriyshashkin:
## Type of issue
Bug

## Description
NoBid response from prebid server has no adUnitCode, addBitResponse discards it, auction doesn't stop until timeout

## Steps to reproduce
Initialize auction with pubmatic or rubicon working through prebidServer. Receive no-bid response

### Expected results
response has ad unit code, addBitResponse accepts it, when other bids arrive, auction is complete

### Actual results
response has not ad unit code, addBitResponse produces warning: "No adUnitCode supplied to addBidResponse, response discarded", auction doesn't end until timeout

## Platform details
0.24.0-pre
node v6.9.5
Fedora 25
Browser Chrome 58

## Other information
Prebid server response:

{
    "tid": "44b005c6-10d6-4aa5-958a-d750f7dc6762",
    "status": "OK",
    "bidder_status": [{
        "bidder": "pubmatic",
        "response_time_ms": 35,
        "no_bid": true
    }, {
        "bidder": "rubicon",
        "response_time_ms": 213,
        "num_bids": 1
    }, {
        "bidder": "indexExchange",
        "ad_unit": "ad_cs_9493688_728_90",
        "response_time_ms": 91,
        "no_bid": true
    }],
    "bids": [{
        "bid_id": "2d379517db8e53",
        "code": "ad_cs_9493688_728_90",
        "creative_id": "4320530",
        "bidder": "rubicon",
        "price": 0,
        "adm": "creative here");",
        "width": 728,
        "height": 90
    }]
}

mkendall07:
Thanks for the report. Will address in a future release.

PR #1204, opened 2017-05-16
When prebid server issues a no-bid response, call addBidResponse for every adUnit requested

dmitriyshashkin:
## Type of change
- [x] Bugfix

## Description of change
Currently prebid-server issues single no-bid response for all the add units and the response doesn't have the ad unit code field (the only exception is indexExchange, server side has different logic just for this bidder). Yet addBidResponse is being called and this non-existing field is passed as it's first argument. addBidResponse rejects such call, so prebid keeps waiting for response until timeout happens. I've decided that if there is no ad unit specified in response I should call addBidResponse for all ad units requested by the bidder. Not sure if that's the right approach but it solves the issue.

## Other information
Fixes https://github.com/prebid/Prebid.js/issues/1196


src/adapters/prebidServer.js
```diff
@@ -58,12 +58,25 @@ function PrebidServer() {
...
 
-              let bidObject = bidfactory.createBid(STATUS.NO_BID, {
-                bidId: bidder.bid_id
-              });
-              bidObject.adUnitCode = bidder.ad_unit;
-              bidObject.bidderCode = bidder.bidder;
-              bidmanager.addBidResponse(bidObject.adUnitCode, bidObject);
+              if (!bidder.ad_unit) {
+                utils.getBidderRequestAllAdUnits(bidder.bidder).bids.forEach(bid => {
+                  let bidObject = bidfactory.createBid(STATUS.NO_BID);
```
mkendall07: Pass through the `bid` object as the 2nd param to `createBid` so we can get the ID (see line 78 below as an example)

dmitriyshashkin:
@mkendall07 fixed! I'm passing bid request object to createBid, also added new test into spec file to make sure that bid request id is used as bid's ad id. Also merged changes from master and changed code formating in accordane with new liting rules

dmitriyshashkin:
@mkendall07 upon further testing I found that I missed few cases. 

1) If there is a bid for one unit and no bid for the other, there is no no_bid response, instead there is response with num_bids less then bids requested. And again in this case auction ends only after timeout bacause bidsBackAll always returns false. 
2) Another case is when istead of no_bid server responds with no_cookie. As I understood from my analysis of server-side code, if server generates no_cookie response, it doesn't even query the exchange for bids. So again there should be a status 2 response for every ad unit.
3) Specifically for indexExchange in case of no_bid response I should call addBidResponse as many times as there are sizes in the bid request for the current ad unit

I'm not quite sure if I should adress these in the same PR or I can create a new one for that purpose.

mkendall07:
Thanks for the analysis. Sounds like #1 might be a prebid server bug - I'll take a look. 2 should be straightforward and I think should be included in this PR. #3 I don't generally like the approach so I'd skip it for no.

dmitriyshashkin:
Example of the response for the first case below. Request included two different ad units. Can post full request payload too, if needed.

```
{
        "tid": "0b54cd48-934f-40ef-bd36-f1e6a9a7ba4b",
        "status": "OK",
        "bidder_status": [{
            "bidder": "rubicon",
            "response_time_ms": 150,
            "num_bids": 1
        }],
        "bids": [{
            "bid_id": "2c4a7ec8d3c126",
            "code": "ad_cs_9493688_728_90",
            "creative_id": "4227344",
            "bidder": "rubicon",
            "price": 0.226596,
            "adm": "%ad code here%",
            "width": 728,
            "height": 90
        }]
}
```
Server side has the following logic (pbs_light.go):
```
} else if bid_list != nil {
	bidder.NumBids = len(bid_list)
	am.BidsReceivedMeter.Mark(int64(bidder.NumBids))
	for _, bid := range bid_list {
		ametrics.PriceHistogram.Update(int64(bid.Price * 1000))
		am.PriceHistogram.Update(int64(bid.Price * 1000))
	}
} else {
	bidder.NoBid = true
	ametrics.NoBidMeter.Mark(1)
}
```

So if there's at least one bid, NoBid flag is not set.

dmitriyshashkin:
@mkendall07 added no bid response generation for the # 2 (when user sync is requested). Waiting for your review. And I'll add an issue for # 3

mkendall07:
LGTM. Can you also open a ticket for #1 ? after review, I think this should be fixed in the adapter, but it can be separate PR. Thanks!

PR #1218, opened 2017-05-22
Audience Network: separate size from format

lovell:
## Type of change

- [x] Bugfix
- [ ] Feature
- [ ] New bidder adapter   
- [ ] Build related changes
- [ ] CI related changes
- [ ] Other

## Description of change

Hello, this change is based on feedback from Prebid users and was commissioned and paid for by Facebook.

Some confusion has arisen due a mixture of size and format values in the Audience Network adaptor. This PR seeks to separate the two concerns by moving format to a parameter of its own instead of overloading the existing concept of size. This prevents formats specific to Audience Network being sent to other bidders as sizes.

There will be a separate PR for the associated documentation.

/cc @wheresbarney

### DIFF
--- a/src/adapters/audienceNetwork.js
+++ b/src/adapters/audienceNetwork.js
@@ -23,21 +23,9 @@ const validateBidRequest = bid =>
   Array.isArray(bid.sizes) && bid.sizes.length > 0;
 
 /**
- * Return a copy of a bid with slot sizes flattened and filtered
- * @param {Object} bid
- * @returns {Object} copy of bid
- */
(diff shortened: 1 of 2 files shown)


protonate:
When debugging FB error response I noticed that here: https://github.com/lovell/Prebid.js/blob/aba63d940ae7600382139d4f9f9f3f08c5f39709/src/adapters/audienceNetwork.js#L169

any errors are logged using `utils.logError`. If these error states are expected consider using `utils.logWarning` instead.

PR #1220, opened 2017-05-23
Bugfix/target filtering api fix

### LINKED ISSUE #1217: Targeting filtering API is broken and mismatched
## Type of issue
**Bug**

## Description
Sometimes the targeting is filtering on a single ad unit code, sometimes on an array of codes, and the implementation seems to switch between those two expectations in a broken fashion.  I tried to fix this myself but I'm not sure if API changes would be required to fix these mismatches.

[Here](https://github.com/prebid/Prebid.js/blob/6a8f481de34ae541d6f8e8e70399d7e89b750697/src/prebid.js#L201) `targeting.getAllTargeting` is called with a single adUnit.
[Here](https://github.com/prebid/Prebid.js/blob/6a8f481de34ae541d6f8e8e70399d7e89b750697/src/prebid.js#L271) `targeting.getAllTargeting` is called with an array of adUnit codes (according to the documentation).
[Here](https://github.com/prebid/Prebid.js/blob/6a8f481de34ae541d6f8e8e70399d7e89b750697/src/targeting.js#L28) `target.getAllTargeting` expects `undefined` or a single adUnit code.  So either this code needs to be updated to allow an array, or the documentation needs to be updated to only allow a single adUnit code.

[Here](https://github.com/prebid/Prebid.js/blob/6a8f481de34ae541d6f8e8e70399d7e89b750697/src/targeting.js#L32) `getWinningBidTargeting` is passed the new array of adUnitCodes, but `getWinningBidTargeting` [expects no arguments](https://github.com/prebid/Prebid.js/blob/6a8f481de34ae541d6f8e8e70399d7e89b750697/src/targeting.js#L104) (it's possible `getWinningBidTargeting` could be updated to pass through the adUnitCodes it is given to `targeting.getWinningBids` call, but `targeting.getWinningBids` expects a single adUnit for filtering, not the array it would be given)

[Here](https://github.com/prebid/Prebid.js/blob/6a8f481de34ae541d6f8e8e70399d7e89b750697/src/targeting.js#L12) `targeting.resetPresetTargeting` is operating on either `pbjs._adUnitCodes` or the remapped result of `targeting.getAllTargeting`  (which I think is erroneously placed in an array) which are fundamentally different structures (an array of adUnit code strings vs an array of targeting objects [I think]).

## Solution
It seems to me in order to solve this, we need to refactor all the targeting code around either expecting an array of adUnitCodes or a single adUnitCode and then update the documentation accordingly?  I'm not quite sure which to favor, probably the array of adUnitCodes? (which would require the larger refactor)

Also, a lot of this code is very confusing to read as well; such as using `.map` or `.find` to loop through data structures and returning nothing or throwing away the results while mutating, which makes it really hard to follow what kind of data structures we're actually working with in the end.  We might benefit from rewriting this code in a clearer fashion.

## Other information
#1158 brought the issue to light (failed some of our internal testing) and I think doubled down on some of the bugs that I think might have already existed.  I have suggested internally that Rubicon not update to 0.24 because I think this pull-request specifically broke some of our implementations.

mkendall07:
## Type of change
- [x] Bugfix

## Description of change
Resolves #1217
## Other information
@prebid/core  for review

### DIFF
--- a/src/prebid.js
+++ b/src/prebid.js
@@ -257,8 +257,8 @@ $$PREBID_GLOBAL$$.getBidResponsesForAdUnitCode = function (adUnitCode) {
 };
 
 /**
- * Set query string targeting on all GPT ad units.
- * @alias module:$$PREBID_GLOBAL$$.setTargetingForGPTAsync
+ * Set query string targeting on one or more GPT ad units.
+ * @param {(string|string[])} adUnitsCodeArr a single `adUnit.code` or multiple.
  */
 $$PREBID_GLOBAL$$.setTargetingForGPTAsync = function (adUnitsCodeArr) {
   utils.logInfo('Invoking $$PREBID_GLOBAL$$.setTargetingForGPTAsync', arguments);
--- a/src/prebid.js
+++ b/src/prebid.js
@@ -259,6 +259,7 @@ $$PREBID_GLOBAL$$.getBidResponsesForAdUnitCode = function (adUnitCode) {
 /**
  * Set query string targeting on one or more GPT ad units.
  * @param {(string|string[])} adUnitsCodeArr a single `adUnit.code` or multiple.
+ * @alias module:$$PREBID_GLOBAL$$.setTargetingForGPTAsync
  */
 $$PREBID_GLOBAL$$.setTargetingForGPTAsync = function (adUnitsCodeArr) {
   utils.logInfo('Invoking $$PREBID_GLOBAL$$.setTargetingForGPTAsync', arguments);
--- a/src/prebid.js
+++ b/src/prebid.js
@@ -260,7 +260,7 @@ $$PREBID_GLOBAL$$.getBidResponsesForAdUnitCode = function (adUnitCode) {
  * Set query string targeting on all GPT ad units.
  * @alias module:$$PREBID_GLOBAL$$.setTargetingForGPTAsync
  */
-$$PREBID_GLOBAL$$.setTargetingForGPTAsync = function (adUnits) {
+$$PREBID_GLOBAL$$.setTargetingForGPTAsync = function (adUnitsCodeArr) {
   utils.logInfo('Invoking $$PREBID_GLOBAL$$.setTargetingForGPTAsync', arguments);
   if (!isGptPubadsDefined()) {
     utils.logError('window.googletag is not defined on the page');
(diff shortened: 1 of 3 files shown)


src/targeting.js
```diff
@@ -25,7 +25,13 @@ targeting.resetPresetTargeting = function(adUnitCode) {
 };
 
 targeting.getAllTargeting = function(adUnitCode) {
-  const adUnitCodes = adUnitCode && adUnitCode.length ? [adUnitCode] : $$PREBID_GLOBAL$$._adUnitCodes;
+  let adUnitCodes = $$PREBID_GLOBAL$$._adUnitCodes;
+  if (typeof adUnitCodes === 'string') {
+    adUnitCode = [adUnitCode];
```
snapwich: should this be `adUnitCodes = [adUnitCode]`?

snapwich:
Thanks for the quick turnaround looking into this

src/targeting.js
mkendall07: yep good catch

src/prebid.js
```diff
@@ -260,21 +260,21 @@ $$PREBID_GLOBAL$$.getBidResponsesForAdUnitCode = function (adUnitCode) {
  * Set query string targeting on all GPT ad units.
  * @alias module:$$PREBID_GLOBAL$$.setTargetingForGPTAsync
  */
-$$PREBID_GLOBAL$$.setTargetingForGPTAsync = function (adUnits) {
+$$PREBID_GLOBAL$$.setTargetingForGPTAsync = function (adUnitsCodeArr) {
```
dbemiller: Do you think we should start JSDoc-ing params whenever the types cause confusion?

Renaming it here does help... but in other cases (like `getAllTargeting`, where the arg might be a string or array) that's not always possible.

mkendall07: yeah we should. I'll update that

src/targeting.js
```diff
@@ -64,9 +65,28 @@ targeting.setTargeting = function(targetingConfig) {
...
+ * @return {string[]}     AdUnit code array
+ */
+function getAdUnitCodes(adUnitCode) {
+  if (typeof adUnitCode === 'string') {
+    return [adUnitCode];
+  }
+  else if (utils.isArray(adUnitCode)) {
+    return adUnitCode;
+  }
+  return $$PREBID_GLOBAL$$._adUnitCodes;
```
protonate: It should not be possible to get to this point and still have `pbjs._adUnitCodes` filter undefined but might want to test for that or otherwise assure the return value here, if undefined, is handled.

src/prebid.js
```diff
@@ -257,24 +257,25 @@ $$PREBID_GLOBAL$$.getBidResponsesForAdUnitCode = function (adUnitCode) {
 };
 
 /**
- * Set query string targeting on all GPT ad units.
+ * Set query string targeting on one or more GPT ad units.
+ * @param {(string|string[])} adUnitsCodeArr a single `adUnit.code` or multiple.
  * @alias module:$$PREBID_GLOBAL$$.setTargetingForGPTAsync
  */
-$$PREBID_GLOBAL$$.setTargetingForGPTAsync = function (adUnits) {
+$$PREBID_GLOBAL$$.setTargetingForGPTAsync = function (adUnitsCodeArr) {
```
mkendall07: Yeah I thought about it. It does make sense

src/targeting.js
mkendall07: good call. I'll address.

PR #1224, opened 2017-05-24
Add drop pixels once feature to AOL Adapter

vzhukovsky:
## Type of change

- [x] Feature

## Description of change
Added drop pixels once feature + unit tests.
Fixed code style.

### DIFF
--- a/src/adapters/aol.js
+++ b/src/adapters/aol.js
@@ -4,6 +4,10 @@ const bidfactory = require('../bidfactory.js');
 const bidmanager = require('../bidmanager.js');
 const constants = require('../constants.json');
 
+$$PREBID_GLOBAL$$.aolGlobals = {
+  pixelsDropped: false
+};
+
 const AolAdapter = function AolAdapter() {
   let showCpmAdjustmentWarning = true;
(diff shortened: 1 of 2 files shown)


slimkrazy:
Hey all,

Just a note about this change: Is there a valid reason for any one SSP to drop pixels for every bid response (theirs or the DSPs) per a single page impression? Is this the kind of feature that could be added to the core of Prebid as opposed just being in adapters?

bretg:
@slimkrazy - PR #1229 adds a user sync module that will allow adapters to register pixels with the platform that will be dropped after the auction is done. At some point, all adapters will need to use this mechanism rather than directly adding the pixel.

protonate:
@vzhukovsky please have a look at #1229 to see if this will serve the purpose of this PR. The User Sync feature is slated for our next release.

slimkrazy:
Hey @protonate - #1229 is missing iframe support, also what it does not do what we require it to. 

#1229 is syncing after a hardcoded timeout, whereas we allow the publisher to decide if they want to alter the default behaviour of syncing on creative render to syncing on bid response. There is scope for us to use this utility for our default behaviour, but not for the case when a publisher actually wants to sync on bid response (they may want to attempt to improve match rates to increase returned CPMs).

For us to adopt #1229, we'd need iframe support. 

Are we ok to have our changes approved as-is, then we can help to work on a generic solution that utilises the module in #1229 for a future release?

grevory:
Yes we are open to prioritizing a review of this PR to accommodate you while you wait for the official user sync module. I think we all expected user sync would be available by now.

WRT to timeout, it is not hard-coded. There is a config option which will be documented. It will also increased by default with the enhancements we are working on.

PR #1227, opened 2017-05-24
Justpremium Adapter

mkalafior:
## Type of change
- [x] New bidder adapter   

## Description of change
New adapter from Justpremium

```
// example 1
{
  bidder: 'justpremium',
  params: {
    zone: 34604
  }
}
// example 2
{
  bidder: 'justpremium',
  params: {
    zone: 34604,
    allow: ['lb']
  }
}
// example 3
{
  bidder: 'justpremium',
  params: {
    zone: 34604,
    exclude: ['lb']
  }
}
```
sebastian.widelak@justpremium.com


jaiminpanchal27:
@mkalafior I see that your external js file http://d2nvliyzbo36lk.cloudfront.net/adp/bc.js playing around with `pbjs` global and i also see reference to `winningBids`
Can you elaborate what you are doing in minified js with `pbjs`

mkalafior:
@jaiminpanchal27  sure, I'm was doing two things there (was as this is deprecated part of the code kept only for backward compatibility). At first I was detecting under which variable prebid is stored (one of our test publisher is using different name than `pbjs`) then I was checking if our bid was delivered on time (by listening on prebid timeout event). And at the end I was also gathering info about winning bids to understand why we lose (by tracking winning bids cpm, delivery time etc). Thats all.

jaiminpanchal27:
@mkalafior no bidder adapters is allowed to track winning bids that are not their own. So update your js code accordingly.

mkalafior:
@jaiminpanchal27 i updated the code, please check it and tell me if its ok for you now.

src/adapters/justpremium.js
```diff
@@ -0,0 +1,245 @@
...
+  }
+
+  function onLoad() {
+    jPAM = top.jPAM = Jpx.JAM.instance({
+      plugins: ['bidder']
+    });
+  }
+
+  function loadResources() {
+    if (toLoad.length > 0) {
```
mkalafior: In development mode, instead of one file, few separated ones are loaded. For this purpose I would prefer to leave it as it is. Of course if its ok for you.

src/adapters/justpremium.js
```diff
@@ -0,0 +1,245 @@
...
+    const pubCond = preparePubCond(bids);
+    const reqBids = bids.concat();
+
+    Object.keys(pubCond).forEach(zone => {
+      loadTag(
+        {
+          zone: zone,
+          hostname: d.location.hostname,
+          protocol: d.location.protocol.replace(':', ''),
+          sw: top.screen.width,
```
mkalafior: Will check that.

mkalafior:
@jaiminpanchal27 I updated the code and also applied most of your suggestions (I only left `toLoad` as an array). 
But one remark here. As we are not able to serve an ad from cross origin iframe, I added a condition to detect such execution and to return empty bid then. Hope its ok for you. 
Additionally I prepared a demo with our adapter [here](http://s3-eu-west-1.amazonaws.com/justpremium/demo/appnexus.html).

modules/justpremiumBidAdapter.js
```diff
@@ -0,0 +1,269 @@
...
+
+    reqBids
+      .filter(r => parseInt(r.params.zone) === parseInt(zone))
+      .forEach(bid => {
+        const bidder = c[`bidder${zone}`];
+
+        bidmanager.addBidResponse(bid.placementCode, bidder.createBid(function (ad) {
+          let bidObject;
+          if (!ad) {
+            bidObject = bidfactory.createBid(CONSTANTS.STATUS.NO_BID);
```
mkendall07: Need to pass through the `bidObj` from the request. Take a look at other adapters for examples

mkalafior:
@mkendall07, @jaiminpanchal27 are you able to check my last commit? I followed the instruction given by @mkendall07. Is there anything else which is blocking us to merge this PR?

ehulsbosch:
@mkendall07 could you please review the change that was requested?

mkendall07:
This is merged into master. Please submit a PR to the [docs repo](https://github.com/prebid/prebid.github.io) to add a file for your adapter to the [bidders directory](https://github.com/prebid/prebid.github.io/tree/master/dev-docs/bidders) so your adapter's params will appear on the [bidders page](http://prebid.org/dev-docs/bidders.html). Thank you for contributing

PR #1228, opened 2017-05-24
New Cox adapter

reynold-cox:
## Type of change
New bidder adapter

## Description of change
test parameters for validating bids:
```
{
  bidder: 'cox',
  params: {
    size: '300x250',
    id: 2000005991707,
    siteId: 2000100948180,
  }
}
```
- contact email of the adapter’s maintainer: reynold@coxmt.com

### DIFF
--- a/src/adapters/cox.js
+++ b/src/adapters/cox.js
@@ -4,9 +4,9 @@ var adLoader = require('../adloader.js');
 
 var CoxAdapter = function CoxAdapter() {
   var adZoneAttributeKeys = ['id', 'size', 'thirdPartyClickUrl'],
-      otherKeys = ['siteId', 'wrapper', 'referrerUrl'],
-      placementMap = {},
-      W = window;
+    otherKeys = ['siteId', 'wrapper', 'referrerUrl'],
+    placementMap = {},
+    W = window;
 
   var COX_BIDDER_CODE = 'cox';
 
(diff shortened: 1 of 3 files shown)


src/adapters/cox.js
```diff
@@ -0,0 +1,255 @@
...
+
+        //    // If it's still too high, just take 80% to 90% of the bid
+        //    if (floor >= bid) floor = bid * ((Math.random() * 10) + 80) / 100;
+        //  }
+        //  return Math.round(floor * 100) / 100;
+        // },
+
+        // setAuctionPrice: function setAuctionPrice(ad, bid) {
+        //  return ad ? ad.replace('${AUCTION_PRICE}', bid) : ad;
+        // },
```
reynold-cox: We have a partner that utilizes the commented-out code.  We'd rather keep it in there for completeness' sake.  But if y'all would prefer it be removed, we'll be happy to do so.  Please let me know.  Thanks!

dbemiller:
@reynold-cox What exactly does it mean if someone configures your bidder with a `size` param?

Not that I want to go policing bidders' params or anything... but `size` already comes in on the bid, and nobody reviewing this could think of a reason why a _bidder_ would have a size... so I wanted to make sure it wasn't an oversight & duplicate info.

Or if there's motivation for the duplicate info, then it'd be very helpful to know why, so we can factor the use case into future prebid designs.

protonate:
I'm not able to get bids back, are the test params still operational?

reynold-cox:
Hi David
Our ad spaces only support a single size each, and our ad servers expect the size to be present in the request.  Since the ad unit config can have multiple sizes defined, we can't always deduce the size for a given Cox ad space...so that's why we have the "size" key in the bidder params.  Thanks!

reynold-cox:
@protonate:
The test params should be ok.  I just used the following and I'm getting a non-zero bid back:

    {
      bidder: 'cox',
        params: {
          size: '300x250',
          id: 2000005991707,
          siteId: 2000100948180,
      }
    }

Are you not getting a response, or are you getting a zero bid?

protonate:
Ah, I see, thanks, it works if running on `localhost`, confirmed bids back, reviewed.

PR #1233, opened 2017-05-25
check return from getBidderRequest in addBidResponse

### LINKED ISSUE #1232: sharethrough doesn't set bidder code on zero cpm bids
## Type of issue
Bug

## Description
tldr; the sharethrough adapter doesn't properly set the bidderCode when it gets a zero CPM bid, which causes bidmanager to timeout prematurely

When Sharethrough gets a response with no creatives, it generates an exception at `bid.cpm = bidResponse.creatives[0].cpm;` The catch block calls _handleInvalidBid, which creates a new type 2 bid using bidfactory createBid. 

However, _handleInvalidBid doesn't set the bidderCode on the bid before sending it to addBidResponse, and addBidResponse doesn't seem to check. 

Later on in addBidResponse `const { requestId, start } = getBidderRequest(bid.bidderCode, adUnitCode);` is called and since bid.bidderCode is '', start is null. This causes the request to timeout since timeToRespond = responseTimestamp.

## Steps to reproduce
Get a zero bid from sharethrough.

### Expected results

- if getBidderRequest can't find the request and so start is null, addBidResponse shouldn't set timeToRespond
- either sharethrough should set the bidderCode, or createBid should pull it from the bidRequest

kitwestneat:
## Type of change

- [x] Bugfix

## Description of change
If getBidderRequest can't find the request and returns nulls, then addBidResponse should not do timeToRespond calculation.

## Other information
Fixes #1232

### DIFF
--- a/src/bidmanager.js
+++ b/src/bidmanager.js
@@ -108,12 +108,16 @@ exports.addBidResponse = function (adUnitCode, bid) {
       adUnitCode
     });
 
-    bid.timeToRespond = bid.responseTimestamp - bid.requestTimestamp;
+    if (start == null) {
+      utils.logWarn(`Cannot find bidder request for bid, bidderCode: ${bid.bidderCode}, adUnitCode: ${bid.adUnitCode}`);
+    } else {
+      bid.timeToRespond = bid.responseTimestamp - bid.requestTimestamp;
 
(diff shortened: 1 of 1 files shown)


kitwestneat:
The aol tests don't actually create a request, they just call callBids directly. The test that's failing check to make sure a warning is issued, but since we're issuing two now, it gets confused.

protonate:
@kitwestneat can you rebase onto master and address conflicts, then we can merge.

mkendall07:
this is conflicting and no update in over a month so closing.

PR #896, opened 2016-12-27
OpenX official adapter release: Fixing the existing openx adapter and…

lntho:
… adding test coverage on it

## Type of change
- [ x ] official adapter submission

- test parameters for validating bids
```
{
  bidder: 'openx’,
  params: {
    // …
    params: {
              unit: ‘538562284’, // OpenX ad unit ID
              delDomain: 'clientname-d.openx.net',  // OpenX delivery domain
              customParams: {"gender": "female"}   // Publisher specified custom parameters, optional
            }
  }
}
```
bidder code: openx
Send All Bids Ad Server Keys: hb_pb_openx, hb_adid_openx, hb_size_openx
Default Deal ID Ad Server Key: hb_deal_openx

- contact email of the adapter’s maintainer: team-openx@openx.com

## Other information
Note:
The OpenX adapter requires setup and approval from your OpenX team. Please reach out to your OpenX representative or Support@openx.com for more information and to enable using this adapter.

PR #1237, opened 2017-05-25
remove superfluous parameter from openx adapter

kkharma:
## Type of change

- [X ] Bugfix

## Description of change
removed unused ee parameter from openx adapter

### DIFF
--- a/src/adapters/openx.js
+++ b/src/adapters/openx.js
@@ -218,7 +218,6 @@ const OpenxAdapter = function OpenxAdapter() {
       ifr: isIfr,
       tz: startTime.getTimezoneOffset(),
       tws: getViewportDimensions(isIfr),
-      ee: 'api_sync_write',
       ef: 'bt%2Cdb',
       be: 1,
       bc: BIDDER_CONFIG


dbemiller:
@prebid/openx Is this guy from your team? Do you trust these changes?

No offense meant, but... without a link to a public API doc, someone we know to be connected to OpenX vetting you, or bidder params which we can test with, I'm not super comfortable merging this.

The Prebid core team tries to leave adapter maintenance up to the adapter's owners... but if we blindly approve changes like these, I can imagine malicious contributors sabotaging their competitor's adapter without their knowledge.

Would like feedback on how active others think we should be in this process.

kkharma:
no worries... i work for openx.  is there someone from openx you've worked with in the past on adapter changes?  i can reach out to them.  i'm not sure why this parameter was ever set in the adapter in the first place.

dbemiller:
I'm new to prebid too... but it looks like the prebid/openx group includes @Heray, who I don't recognize... so he'd be my best guess.

kkharma:
unfortunately, they are not an openx employee, and i can't see the prebid/openx group.  is there anyone else i could try?  otherwise, i'll have to ask around internally.

dbemiller:
huh... the git logs show that this file was in the initial commit, so I'm lost here.

Are there OpenX API docs that this adapter is coding against? I googled and found [this page](https://docs.openx.com/Content/developers/ad_request_api/adtagguide_tagapi.html) mentioned the URL `http://delivery_server_domain/w/1.0/jstag`, which looks like it's what the adapter hits.

But if that's the one, then [the API params](https://docs.openx.com/Content/developers/ad_request_api/adtagguide_parameters.html) barely line up with the adapter at all. `ju`, `jr`, etc... many of them aren't listed there. So I get the feeling I'm missing something.

dbemiller:
Actually, although he's not a part of that group, @lntho shows up in the commit logs for #896 , official OpenX support

kkharma:
ah cool... that's someone i recognize.  will follow up...

lntho:
Hello @dbemiller I can confirm this pull request is fine. Is there supposed to be a prebid/openx group? I was the original submitter for the OpenX adapter

dbemiller:
Ok, thanks @lntho. I'll merge this then.

I do see a bunch of `prebid/Adapter` teams in this project... but only about 25 of them--so far fewer than the total number of adapters. A prebid/openx team does exist... but my guess is it's not really maintained. It only includes some original members of the Prebid project.

This is the first I've seen of these, and I don't have permissions to edit them... so I've no idea if they're maintained. I'll ping the people who do, though, and make sure you guys get added.

PR #1240, opened 2017-05-26
Add Index Exchange Video Adapter

StefanPanait:
## Type of change

- [ ] Bugfix
- [ ] Feature
- [x] New bidder adapter   
- [ ] Build related changes
- [ ] CI related changes
- [ ] Other
## Description of change

Add new video adapter for Index Exchange
Documentation: https://github.com/prebid/prebid.github.io/pull/257

- Change your /etc/hosts to point as.casalemedia.com to 192.139.80.22 (Or create a redirect rule using using a browser plugin to point as.casalemedia.com to sandbox.ht.indexexchange.com)
```
192.139.80.22        as.casalemedia.com
````
```
{
  bidder: "indexExchangeVideo",
  params: {
    video: {
    siteID: 188233,
    playerType: "HTML5",
    protocols: ["VAST2", "VAST3"],
    maxduration: 15,
    allowVPAID: true
    }
  }
}
```

### DIFF
--- a/src/adapters/ixVideo.js
+++ b/src/adapters/ixVideo.js
@@ -55,8 +55,8 @@ const DEFAULT_VPAID_MIMES_MAP = {
   HTML5: ['application/javascript']
 };
 
-const BASE_CYGNUS_VIDEO_URL_INSECURE = 'http://as.casalemedia.com/cygnus?v=8&fn=handleCygnusResponse';
-const BASE_CYGNUS_VIDEO_URL_SECURE = 'https://as-sec.casalemedia.com/cygnus?v=8&fn=handleCygnusResponse';
+const BASE_CYGNUS_VIDEO_URL_INSECURE = `http://as.casalemedia.com/cygnus?v=8&fn=$$PREBID_GLOBAL$$.handleCygnusResponse`;
+const BASE_CYGNUS_VIDEO_URL_SECURE = `https://as-sec.casalemedia.com/cygnus?v=8&fn=$$PREBID_GLOBAL$$.handleCygnusResponse`;
 
 function IndexExchangeVideoAdapter() {
   let baseAdapter = Adapter.createNew('ixVideo');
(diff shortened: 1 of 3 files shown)

### CODE REVIEW

File: src/adapters/ixVideo.js [hunk 1]
```diff
@@ -0,0 +1,443 @@
...
+    let formattedCygnusUrl = url.format(cygnusUrl);
+    return formattedCygnusUrl;
+  }
+
+  function sendCygnusRequest(cygnusRequest) {
+    adloader.loadScript(cygnusRequest);
+  }
+
+  /* Notify Prebid of bid responses so bids can get in the auction */
+  window.handleCygnusResponse = function (response) {
```
grevory: Please use the Prebid global object for handling your responses.
`$$PREBID_GLOBAL$$.handleCygnusResponse = function (response) {`
And then for your URL `const BASE_CYGNUS_VIDEO_URL_SECURE = 'https://as-sec.casalemedia.com/cygnus?v=8&fn='+encodeURIComponent($$PREBID_GLOBAL$$.handleCygnusResponse);

I presume this will work in your case but also prevent adding variables to the global scope.

grevory: Here is an example: https://github.com/prebid/Prebid.js/blob/master/src/adapters/appnexus.js#L145

mkendall07:
@StefanPanait 
Any reason why this is a separate adapter from the regular indexExchange adapter?

michaelharris2:
@mkendall07 
This was only done because by reviewing appnexus.js and appnexusAst.js it appeared that that model had precedent, so we were following suit.

mkendall07:
@StefanPanait 
I see. We actually prefer a single adapter that handles all types. Could you please add this to the existing indexExchange adapter? Apologies for not asking about this sooner - if you can do that we'll review the new PR as a priority to make sure it gets merged quickly. Thanks!

michaelharris2:
@mkendall07 

Ok I'll look into that.  Also note you can just direct all comments to me going forward, instead of Stefan Panait.

Thanks!

mkendall07:
Hey @michaelharris2  - any update on this? Thanks!

michaelharris2:
@mkendall07 we're looking at having the changes made and the pull request updated by July 21st.  

Thanks!

PR #1245, opened 2017-05-30
Add eplanning adapter

sebaperez:
## Type of change

- [ ] Bugfix
- [ ] Feature
- [X] New bidder adapter   
- [ ] Build related changes
- [ ] CI related changes
- [ ] Other

## Description of change

```
{
  bidder: 'eplanning',
  params: {
    ci: '18f66',
    t: '1'
  }
}

```
soporte@e-planning.net

### DIFF
--- a/adapters.json
+++ b/adapters.json
@@ -66,6 +66,7 @@
     "atomx",
     "tapsense",
     "trion",
+    "eplanning",
     "prebidServer",
     "adsupply",
     {
--- a/src/adapters/eplanning.js
+++ b/src/adapters/eplanning.js
@@ -3,7 +3,7 @@ var bidmanager = require('src/bidmanager.js');
 
 var EPlanningAdapter = function EPlanningAdapter() {
   (function() {
-    var win = window, doc = win.document, pbjs = win.pbjs, _global = {}, _default = { 'sv': 'ads.us.e-planning.net', 't': 0 }, rnd, FILE = 'file', CALLBACK_FUNCTION = 'hbpb.rH', NULL_SIZE = '1x1', _csRequested = [], PROTO = '', ISV = 'aklc.img.e-planning.net';
+    var win = window, doc = win.document, pbjs = win.pbjs, _global = {}, _default = { 'sv': 'ads.us.e-planning.net', 't': 0 }, rnd, FILE = 'file', CALLBACK_FUNCTION = 'hbpb.rH', NULL_SIZE = '1x1', _csRequested = [], PROTO = location.protocol === 'https:' ? 'https:' : 'http:', ISV = 'aklc.img.e-planning.net';
     function Hbpb() {
       var slots = (function() {
         var _data = [];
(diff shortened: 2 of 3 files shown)


dbemiller:
LGTM

@sebaperez Just so you know, we're also looking to prevent JSONP in Prebid 1.0 (probably to be worked on starting in the second half of this year). This is fine for the 0.x releases though.

sebaperez:
@protonate I just tested it and works fine for me. Did you test it locally? Do you have the HTML file where it doesn't work?

dbemiller:
An adapter test page got added in #1182 you can use.

protonate:
This is the code that is returned by the bid request:
```
!function() {
    var c, e, a = window, b = a.adUnits, d = {
        sI: {
            k: "1f4b4"
        },
        sec: {
            k: "ROS"
        },
        sp: []
    };
    if (b && b.length)
        for (c = 0; c < b.length; c++)
            if (b[c].bids && b[c].bids.length)
                for (e = 0; e < b[c].bids.length; e++)
                    "eplanning" === b[c].bids[e].bidder && d.sp.push({
                        k: b[c].code,
                        a: [{
                            w: "object" == typeof b[c].sizes[0] ? b[c].sizes[0][0] : b[c].sizes[0],
                            h: "object" == typeof b[c].sizes[0] ? b[c].sizes[0][1] : b[c].sizes[1],
                            adm: "<p>test ad for " + b[c].code + "</p>",
                            pr: 1
                        }]
                    });
    hbpb.rH(d)
}();
```

The `a` variable is set to `window` and then `b` is set to an `a.adUnits` property. There isn't a `window.adUnits` property however. Are you getting this same code block? If so, how is `window.adUnits` getting set?

sebaperez:
@protonate that's the response in test case (when t === '1') - If you do not set that variable, then the response is different. In both cases, it is handled by hbpb.rH().
I think the problem you were experiencing was because you tested it locally (file://) and the call to the adserver is made with a relative URL (//...). I just pushed some changes that forces http or https protocol, and never use "file" protocol.

protonate:
The test case then doesn't return a bid "response". What is the expectation here?

The `t` param is set as `String(1)` in the PR description above. However this should be `Number(0)` to turn off test, as `String(0)` evaluates to `true`. It would be clearer if this was set as `Number(1)` for the test case.

I am testing from localhost and from a "prebidapp.com" domain with a hosts file entry to 127.0.0.1, and on port 9999 in both cases.

With `t` set to `Number(0)` the bidder makes this call:
```
http://ads.us.e-planning.net/hb/1/18f66/1/localhost/ROS?rnd=0.7947294411451566&e=/19968336/header-bid-tag-1:728x90,970x90+/19968336/header-bid-tag-0:300x250,300x600&ur=http%3A//localhost%3A9999/gpt.html&cb=hbpb.rH&r=pbjs
```

And the response is:

```
hbpb.rH();
```

So I'm not able to confirm a bid back in either test or non-test case. Any thoughts?

dbemiller:
We chatted about it. I don't know what's wrong on his system... but everything looks good to me. Screenshots to prove it (using the `hello_world.html` file from #1182, substituting the params @sebaperez gave here).

screen shot 2017-06-14 at 3 08 02 pm

screen shot 2017-06-14 at 3 09 02 pm

PR #1271, opened 2017-06-06
iFrame interaction problem is solved.(Huddledmasses adapter)

HuddledMasses:
## Type of change
- [x] Bugfix
## Description of change
We made major improvements in data receiving mechanism. During the request from iframe we making an attempt to go beyond the block. By doing so we retrieving data from the main window.


mkendall07:
@HuddledMasses please rebase and then we can merge.

mkendall07:
these changes are pulled into master.

PR #1273, opened 2017-06-06
Pubmatic - fix terminating semicolon inside string assignment

protonate:
The terminating semicolon is put inside the quoted string here and on the next two lines and causes an "Unexpected identifier" error to be thrown when the string is parsed for DOM insertion by `iframeDoc.write(_createRequestContent());`.

cc: @prebid/pubmatic @ptomasroos


protonate:
Ah already addressed with #1272, closing.

PR #1277, opened 2017-06-07
Implementing prebid-cache to support Video ads

dbemiller:
PR which leverages prebid-server as a video ad cache. The goal here is to give publishers an ID which they can send to the ad server as a query param.

Per the feedback, I'm not exposing the callback to users anymore. The `bidmanager` simply waits to add the bid until the call to the cache succeeds.


src/bidmanager.js
```diff
@@ -83,6 +84,33 @@ exports.bidsBackAll = function () {
...
+/**
+ * If the bid is a video bid, publishers will need to be able to get the cache ID.
+ * Since the cache call is asynchronous, the getter must be too. This attaches a
+ * vastUrlCacheKey function to the bid.
+ *
+ * This function takes a callback argument, and promises to call it exactly once when the
+ * cache ID has been returned.
+ */
+function prepVideoBid(bid) {
+  // Queue callbacks which arrive before the cache has returned.
```
dbemiller: This got deleted anyway... it's not necessary if we don't expose the callback to the user.

src/videoCache.js
```diff
@@ -0,0 +1,213 @@
...
+}
+
+/**
+ * A function which should be called with the results of the storage operation.
+ *
+ * @callback videoCacheStoreCallback
+ *
+ * @param {Error} [error] The error, if one occurred.
+ * @param {?string[]} uuids An array of unique IDs. The array will have one element for each bid we were asked
+ *   to store. It may include null elements if some of the bids were malformed, or an error occurred.
```
protonate: Also a note for server-side, but perhaps that would be the place to prevent "garbage in" so we don't have to deal with it coming back out.

src/videoCache.js
```diff
@@ -0,0 +1,213 @@
...
+ *   Each non-null element in this array is a valid input into the retrieve function, which will fetch
+ *   some VAST XML which can be used to render this bid's ad.
+ */
+
+/**
+ * A function which bridges the APIs between the videoCacheStoreCallback and our ajax function's API.
```
protonate: Raises the question why we should need to shim or bridge APIs that we control?

src/videoCache.js
```diff
@@ -0,0 +1,213 @@
...
+ *
+ * @param {Error} [error] The error, if one occurred.
+ * @param {string} A VAST XML string. If error is defined, this can still expect an "empty"
+ *   VAST URL response.
+ */
+
+/**
+ * Takes the user-defined callback, and wraps it into our ajax function's API.
+ *
+ * This also needs to work around the fact that the cache server returns a 200 on "data not found"...
```
protonate: We should change this behavior server-side?

src/videoCache.js
```diff
@@ -0,0 +1,213 @@
...
+      retrievalFunction(cacheId, cachedCallback(cacheId, callback));
+    }
+  }
+}
+
+/**
+ * Fetch the VAST XML associated with cacheId, and execute the callback function with it.
+ *
+ * @param {string} cacheId The ID of the video whose content should be retrieved.
+ * @param {videoCacheRetrieveCallback} callback A callback function which should be run once the
```
protonate: We've been conforming to calling this a `done` callback.

src/videoCache.js
```diff
@@ -0,0 +1,213 @@
...
+      callback(err, EMPTY_VAST_RESPONSE);
+    }
+  }
+}
+
+/**
+ * Make a server call for the given cacheId, and execute the callback function when they get back to us.
+ *
+ * @param {string} cacheId The ID of the video whose content should be retrieved.
+ * @param {videoCacheRetrieveCallback} callback A callback function which should be run once the
```
protonate: Call this `done`?

src/videoCache.js
```diff
@@ -0,0 +1,213 @@
...
+  }
+  ajax(tryAppendQueryString(GET_URL, 'uuid', cacheId).slice(0, -1), shimRetrievalCallback(callback));
+}
+
+/**
+ * Memoization function which prevents excessive server calls. This stores all the
+ * [cacheId, cacheValue] pairs locally so that we don't go to the network again if
+ * the publisher requests the same VAST content more than once.
+ */
+function memoize(retrievalFunction) {
```
protonate: Great, can you add some info on when this is used? For example, I think it would only be when a bid is returned to an auction from the (pending) BidPool. Is my understanding accurate?

test/spec/unit/bidmanager_spec.js
```diff
@@ -0,0 +1,149 @@
...
+          expect(callback.called).to.equal(false);
+        }
+      };
+    }
+
+    /**
+     * Initialize the global state so that the auction-space looks like we want it to.
+     *
+     * @param {Array} adUnits The array of ad units which should appear in this auction.
+     * @param {function} bidRequestTweaker A function which accepts a basic bidRequest, and
```
protonate: Best name yet!

src/adServerManager.js
```diff
@@ -0,0 +1,90 @@
...
+/**
+ * This file serves as the plugin point for adserver module functionality.
+ *
+ * Ad Server modules should call {@link registerAdserver} to connect into prebid-core.
+ *
+ * If publishers bundle Prebid with a single Ad Server module, they'll be able to interact with it
+ * through the properties and functions on `pbjs.adserver`.
+ *
+ * If publishers bundle with more than one Ad Server, they'll need to interact with them through
+ * `pbjs.adservers.{adServerCode}`, where the adServerCode is defined by the ad server itself.
```
dbemiller: Can it be? I'm all for deriving properties and avoiding data redundancies.

Which Bid/AdUnit properties could be used to derive it?

mkendall07: We need to introduce a new config object I think.

protonate: I'd image the pub would be configured with an Ad Server definition (include "none") and an AdUnit could be configured with an override to use, if present.

src/adServerManager.js
```diff
@@ -0,0 +1,90 @@
...
+ */
+
+/**
+ * @typedef {Object} VideoSupport
+ *
+ * @property {string} code The identifying code for this adserver. Note that this contributes to
+ *   the user-facing API. For example, if the adserver code is 'dfp', and the Prebid bundle contains two
+ *   or more ad server modules, then publishers will access its functions through 'pbjs.adservers.dfp'.
+ *
+ * @method {VideoAdUrlBuilder} buildVideoAdUrl
```
dbemiller: yeah that's fair. Actually I agree, cause in my mind, `method` seems to signify something that shouldn't be ripped off of the object, because it's relying on `this`. `function` signifies something that can be passed around freely.

I'll change it.

src/videoCache.js
```diff
@@ -0,0 +1,213 @@
...
+ *   - Generate IDs for each valid bid, and cache the key/value pair on the server.
+ *   - Return these IDs so that publishers can use them to fetch the bids later.
+ *
+ * This trickery helps integrate with ad servers, which set character limits on request params.
+ */
+
+import { ajax } from './ajax';
+import { tryAppendQueryString } from './utils';
+
+const PUT_URL = 'https://prebid.adnxs.com/pbc/v1/put'
```
dbemiller: No clue. I think Sweeny implemented it... so he'd be a better one to ask.

I agree it's a bit weird.

src/videoCache.js
dbemiller: I wanted to do this... but actually couldn't come up with a clean way. It doesn't really have anything to do with server behavior.

My assumption: A caller who **knows** their objects are invalid will filter them out before calling this method.
 
Now imagine a caller with an **unknown** invalid object, who's trying to associate the IDs with their objects:

```
cache.store([validThing1, invalidThing, validThing2], function(uuids) {
  validThing1.uuid = uuids[0];
  invalidThing.uuid = uuids[1];
  validThing2.uuid = uuids[2]; 
});
```

So... at the API level, what are my options for preventing "garbage in"? These are the things I thought of:

1. Cache nothing, and reject any request with at least one invalid object 
2. Return a length-2 array with the valid UUIDs
3. Null out the indices of invalid requests, and maintain the indices for valid ones

(1) is a solid API...but seemed unnecessarily unforgiving.
(2) is error-prone, and produces subtle bugs in the calling code.
(3) is more forgiving, and leaves un-cached objects with an undefined ID.

In other languages, I'd do a `Map[Object, String]`... but I don't think javascript supports those in our browsers. It might be polyfill-able though? Anyway:

I went with (3) because it seemed most reasonable.
I'm open to (1) if you think it's better.
I'm strongly opposed to (2).

If you have anything else in mind that I didn't think of, let me know so we can run the pros/cons on them.

src/videoCache.js
dbemiller: The semantics are debatable...

As a server: someone just told you to return the data stored in the cache. If you looked in the cache and _successfully_ found no value, did you complete the request without error? What if someone _told_ you to store "nothing", in an earlier `put` request?

I answered that with a definite ¯\_(ツ)_/¯, and decided to work with Sweeny's implementation.

src/videoCache.js
dbemiller: Sure

src/videoCache.js
dbemiller: Didn't know that... I'll update it

src/videoCache.js
protonate: A `204` would provide a definite "successful request with no content response".

protonate: I think my broader point is we also control the server and don't have to work around anything that doesn't make sense to us.

src/videoCache.js
dbemiller: It's used right below it, in the exported function :).

Per the docs, the goal here is to prevent multiple server calls.

If we successfully fetched a thing from the server, we can store it in page memory for the next time.

If the server fetch failed, we want to retry the call (might have just been a network problem).

It might have some more general uses, but... better private than public and only used once, imo. If refactoring, it's much easier to make more visible rather than less.

src/videoCache.js
dbemiller: I'd buy that. So trying to be clear, my understanding is that you think the cache should:

1. If it doesn't recognize the key, return a 404.
2. If it recognizes the key and stores "nothing", return a 204.
3. If it recognizes the key and stores "something", return a 200 with the stored data.

Correct?

src/videoCache.js
dbemiller: A fair question... my belief is that each function, file, and API should make as much sense as possible on its own.

If you're looking at the `ajax` file, odds are that you want to make an HTTP request. Its API lets you set HTTP headers, considers any 200 to be a success, gives your callback function gets an HTTP response, etc.

If you're looking at the `videoCache` file, odds are that you want to cache a video somewhere. Its API lets you give it data to store, considers _any_ form of "cache miss" to be a failure, and gives your callback function the data it was seeking.

In between, there's code. "Data to store" needs to become a POST body. 200's need to be interpreted as JSON. Request details (URLs, headers, etc) can be chosen without bothering the caller who just wants to store their video somewhere.

Some of that code is responsible for transforming the callbacks. That's what I'm calling a "shim".

src/videoCache.js
```diff
@@ -0,0 +1,213 @@
...
+ */
+function toStorageRequest(bid) {
+  return {
+    'value': wrapURI(bid.vastUrl)
+  };
+}
+
+/**
+ * If the server responded successfully, then it gave us a bunch of objects with uuid properties.
+ * Abstract our background dependency away by renaming them to cacheId.
```
dbemiller: Relates to my comment here: https://github.com/prebid/Prebid.js/pull/1277#discussion_r124332328

The server's cache endpoints publicize the fact that they're generating UUIDs for each stored element. This may be valuable info for users asking a very public URL to cache things for them.

The code using the videoCache.... has no reason to care about that implementation? As far as I know. So I renamed it. If, someday, we add some cache endpoints to prebid-server and it no longer need to use UUIDs (which are fairly expensive to generate), then we only need to refactor this file, and it won't be stuck with a misleading name.

That said... I was pretty lukewarm on this indirection myself. I'm happy to kill it off if you don't think it's worth it. Otherwise I'll try to make the comment more clear.

src/bidmanager.js
```diff
@@ -138,15 +140,56 @@ exports.addBidResponse = function (adUnitCode, bid) {
     }
 
     bid.adserverTargeting = keyValues;
+  }
+
+  function doCallbacksIfNeeded() {
```
dbemiller: hah! Good catch. That's because I didn't know what the problem domain was 😊 .

My understanding is that the callbacks are supposed to be executed whenever the auction has ended. Given that:

It makes sense that `addBidToAuction` executes them if all the bids have been received, because there's no sense in waiting if we're not expecting any more bids.

It makes sense that `requestBids` in `prebid.js` [calls them](https://github.com/prebid/Prebid.js/blob/master/src/prebid.js#L412) if the timeout has been reached 

...but those were the only two cases I could think of. These lines seem to be ending the auction based on a third condition... but I have no idea what that condition is, or why it exists. Honestly I would have deleted them... but since the unit tests on this method were sparse, I didn't feel comfortable doing that without checking in with someone who had more context.

Can you shed any light on it?

src/adServerManager.js
```diff
@@ -0,0 +1,90 @@
...
+ *
+ * @return {string} A URL which can be passed into the Video player to play an ad.
+ *   This should call the Ad Server, and return the ad which should be played (letting the
+ *   given bid compete alongside the rest of the demand).
+ */
+
+/**
+ * @typedef {Object} VideoSupport
+ *
+ * @property {string} code The identifying code for this adserver. Note that this contributes to
```
dbemiller: Done. I do like `name` better than `code`... and this is a pretty good place to start using it, since the API is pretty internal.

src/videoCache.js
dbemiller: er... and one other judgment call which needs agreement:

1. If the videoCache can't find the video, it should be an error.

(I didn't see any discussion on this point... but it's an important one. If you agree, then a prebid-cache 204 "success" would still be considered an "error" from this file's point of view, and all this code would still be necessary anyway)

src/adServerManager.js
dbemiller: Chatted about this outside of the PR. I removed `pbjs.adserver`, so the API will always be `pbjs.adServers.dfp.buildVideoUrl(options)`.

src/adServerManager.js
```diff
@@ -0,0 +1,90 @@
...
+ * Prepares a namespace on object so that utility functions can be added to it.
+ *
+ * If this is the only code we've seen, attach functionality to $$PREBID_GLOBAL$$.adServer.
+ * If we've seen any other codes, attach functionality to $$PREBID_GLOBAL$$.adServers[code].
+ *
+ * @param {object} object The object where this function should manage namespaces.
+ * @param {string} code The code for this ad server.
+ * @return {object} An object where functions for dealing with this ad server can be added.
+ */
+function prepareNamespaceIn(object, code) {
```
dbemiller: Not relevant anymore because I deleted it, but... it was supposed to read like "prepare a namespace in object for {code}".

I'm generally trying to name functions so that the code which calls them reads as closely to english as possible.

src/videoCache.js
```diff
@@ -0,0 +1,213 @@
...
+ * Function which wraps a URI that serves VAST XML, so that it can be loaded.
+ *
+ * @param {string} uri The URI where the VAST content can be found.
+ * @return A VAST URL which loads XML from the given URI.
+ */
+function wrapURI(uri) {
+  // Technically, this is vulnerable to cross-script injection by sketchy vastUrl bids.
+  // We could make sure it's a valid URI... but since we're loading VAST XML from the
+  // URL they provide anyway, that's probably not a big deal.
+  return '<VAST version="2.0"><Ad id=""><Wrapper><AdSystem>prebid.org wrapper</AdSystem><VASTAdTagURI><![CDATA[' +
```
mkendall07: Are we supporting VAST 2 or VAST3?

mkendall07: Prefer string templates instead of concatenation.

src/videoCache.js
```diff
@@ -0,0 +1,213 @@
...
+      }
+      catch (e) {
+        done(e, []);
+        return;
+      }
+
+      done(null, ids);
+    },
+    error: function(statusText, responseBody) {
+      done(new Error('Error storing video ad in the cache: ' + statusText + ': ' + JSON.stringify(responseBody)), []);
```
mkendall07: string templates again.

src/videoCache.js
```diff
@@ -0,0 +1,213 @@
...
+ */
+
+/**
+ * A function which bridges the APIs between the videoCacheStoreCallback and our ajax function's API.
+ *
+ * @param {videoCacheStoreCallback} done A callback to the "store" function.
+ * @return {Function} A callback which interprets the cache server's responses, and makes up the right
+ *   arguments for our callback.
+ */
+function shimStorageCallback(done) {
```
mkendall07: Don't love the idea of calling a function a `shim`

src/videoCache.js
mkendall07: I don't think we need this functionality at all. We don't actually invoke the read from the cache, the video player does. Am I missing something?

modules/dfpAdServerVideo.js
```diff
@@ -0,0 +1,87 @@
...
+ *   (or the auction's winning bid for this adUnit, if undefined) compete alongside the rest of the
+ *   demand in DFP.
+ */
+export default function buildDfpVideoUrl(options) {
+  const adUnit = options.adUnit;
+  const bid = options.bid || getWinningBids(adUnit.code)[0];
+
+  const derivedParams = {
+    correlator: Date.now(),
+    description_url: encodeURIComponent(bid.descriptionUrl),
```
dbemiller: Per the discussion in chat, I removed this. Publishers can still send it in themselves.

I don't think the `vastUrl` is related... but it's not needed by this point in the code anyway. The URL has been sent to the cache, so it's the `{ hb_uuid: bid.videoCacheKey }` line which adds that to the DFP URL.

src/bidmanager.js
```diff
@@ -138,15 +140,56 @@ exports.addBidResponse = function (adUnitCode, bid) {
...
+      } else {
+        bid.videoCacheKey = cacheIds[0].cacheId;
+        addBidToAuction(bid);
+      }
+    });
   }
 
-  if (bidsBackAll()) {
-    exports.executeCallback();
+  if (isValid()) {
```
dbemiller: Makes sense. Is fixed (after the next push)

src/videoCache.js
dbemiller: Makes sense

src/videoCache.js
dbemiller: Fixed

dbemiller:
Re unnecessary `retrieval` function (dunno why github won't let me respond to it directly): you're totally right.

Whenever I write something which interacts with an HTTP API, my first instinct is always "let's wrap that API in one from {my language} so that it's contained."

So.... I wrote it. Then wrote the video player integration. Didn't even notice I never used it in real code.

Is deleted now.

src/videoCache.js
dbemiller: I ended up removing this. In part because you found the comment confusing... also in part because I took a step back and realized that only the `bidmanager` sees it anyway. Not a part of the public API.

src/videoCache.js
dbemiller: I agree it's not the clearest, but... is there anything you like better? `transform`? `wrap`? Other?

I haven't come up with anything I really liked.

src/videoCache.js
dbemiller: You know... you're totally right.

Whenever I write something which interacts with an HTTP API, my first instinct is always "let's wrap that API in one from {my language} so that it's contained."

So.... I wrote it. Then wrote the video player integration. Didn't even notice that it was never called by real code.

Is deleted now.

src/videoCache.js
```diff
@@ -0,0 +1,213 @@
...
+ *
+ * This trickery helps integrate with ad servers, which set character limits on request params.
+ */
+
+import { ajax } from './ajax';
+import { tryAppendQueryString } from './utils';
+
+const PUT_URL = 'https://prebid.adnxs.com/pbc/v1/put'
+const GET_URL = 'https://prebid.adnxs.com/pbc/v1/get?';
+const EMPTY_VAST_RESPONSE = '<VAST version="3.0"></VAST>'
```
dbemiller: I must admit... I'm a bit confused.

Section 2.4.2.4 of [the official 3.0 spec](https://www.iab.com/wp-content/uploads/2015/06/VASTv3_0.pdf) says that an empty tag is the correct response... but you're right that the XSD doesn't call it valid, and [their official VAST validator](https://iabtechlab.com/IABVAST/#!vast) doesn't either.

Honestly not sure how to interpret that, except to say that I'll test it against the major video players and make sure they properly play the video with no ad.

src/videoCache.js
dbemiller: Per the chat, it doesn't sound like it matters. I'll bump it up to 3.0, since they're more likely to maintain backwards compatibility with that version if the standard ever moves forward any more.

integrationExamples/gpt/video_test.html
```diff
@@ -0,0 +1,132 @@
+
```
protonate: This can be moved to `./test` as it isn't really an integration example.

src/bidmanager.js
```diff
@@ -138,15 +153,45 @@ exports.addBidResponse = function (adUnitCode, bid) {
...
+      exports.executeCallback();
+    }
   }
 
-  if (bidsBackAll()) {
-    exports.executeCallback();
+  // Video bids may fail if the cache is down, or there's trouble on the network.
+  function tryAddVideoBid(bid) {
+    store([bid], function(error, cacheIds) {
+      doCallbacksIfNeeded();
```
mkendall07: Shouldn't we move this line after the bid is added to the auction? I realize that this is not consistent with the previous behavior, but at this point we should have a bid cached, so why not include it? @protonate @dbemiller

protonate: 👍

PR #1283, opened 2017-06-09
Use Renderer command queue to render outstream

matthewlane:
## Type of change
- Feature

## Description of change
This PR adds properties to the Renderer constructor that can be used to delay rendering until a renderer is loaded. `appnexusAst` then makes use of these capabilities to enable rendering outstream video without the need for an ad server.

A call to `pbjs.renderAd` with a winning outstream bid can potentially be invoked before the render function for this bid is in memory. In this case, the render function pushes to the render command queue, which then gets processed after the renderer is loaded. Calls to `renderer.render(bid)` after the renderer is loaded will skip the command queue and invoke the render function immediately.

As a result, outstream video rendering that is invoked [with an ad server](https://github.com/prebid/Prebid.js/blob/master/test/spec/e2e/gpt-examples/gpt_outstream.html), and outstream video rendering that is invoked [without an adserver](http://acdn.adnxs.com/prebid/demos/outstream-without-adserver/), will both work.

### DIFF
--- a/src/adapters/appnexusAst.js
+++ b/src/adapters/appnexusAst.js
@@ -269,18 +269,23 @@ function AppnexusAstAdapter() {
   }
 
   function outstreamRender(bid) {
-    window.ANOutstreamVideo.renderAd({
-      tagId: bid.adResponse.tag_id,
-      sizes: [bid.getSize().split('x')],
-      targetId: bid.adUnitCode, // target div id to render video
-      uuid: bid.adResponse.uuid,
-      adResponse: bid.adResponse,
-      rendererOptions: bid.renderer.getConfig()
-    }, handleOutstreamRendererEvents.bind(bid));
+    // push to render queue because ANOutstreamVideo may not be loaded yet
+    bid.renderer.push(() => {
+      window.ANOutstreamVideo.renderAd({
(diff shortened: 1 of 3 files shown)


src/adapters/appnexusAst.js
```diff
@@ -269,18 +269,23 @@ function AppnexusAstAdapter() {
...
+        rendererOptions: bid.renderer.getConfig()
+      }, handleOutstreamRendererEvents.bind(bid));
+    });
   }
 
-  function onOutstreamRendererLoaded() {
-    // setup code for renderer, if any
+  function onOutstreamRendererLoaded(bid) {
+    // adload.loadScript has loaded ANOutstreamVideo
+    bid.renderer.loaded = true;
```
mkendall07: Should we do this automatically in the renderer? Not have the caller manage it?

mkendall07: Really everything in this method is common, so we can abstract it away and not need it to be defined in the adapter.

matthewlane: Added a commit to do this. Bidders can override the callback by passing a function on the `callback` property given to `Renderer.install`. If that's unnecessary or other changes needed, will update

PR #1292, opened 2017-06-13
Audience Network: Add support for video format

lovell:
## Type of change

- [ ] Bugfix
- [X] Feature
- [ ] New bidder adapter   
- [ ] Build related changes
- [ ] CI related changes
- [ ] Other

## Description of change

Hello, this PR adds support for the `video` format to the Audience Network adapter and was commissioned and paid for by Facebook.

### DIFF
--- a/modules/audienceNetworkBidAdapter.js
+++ b/modules/audienceNetworkBidAdapter.js
@@ -21,7 +21,8 @@ const validateBidRequest = bid =>
   typeof bid.params === 'object' &&
   typeof bid.params.placementId === 'string' &&
   bid.params.placementId.length > 0 &&
-  Array.isArray(bid.sizes) && bid.sizes.length > 0;
+  Array.isArray(bid.sizes) && bid.sizes.length > 0 &&
+  (isVideo(bid.params.format) || bid.sizes.map(flattenSize).some(isValidSize));
 
 /**
  * Flattens a 2-element [W, H] array as a 'WxH' string,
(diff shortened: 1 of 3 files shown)


src/adapters/audienceNetwork.js
```diff
@@ -131,33 +162,43 @@ const callBids = bidRequest => {
...
     const testmode = isTestmode();
+    const pageurl = location.href;
+    const search = {
+      placementids,
+      adformats,
+      testmode,
+      pageurl,
+      sdk
+    };
+    const video = adformats.findIndex(isVideo);
```
dbemiller: Looks like this has [browser compatibility issues](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/findIndex?v=example)

lovell: Good catch, thanks. Given `core-js` is already a dependency, would you be OK with including the specific polyfill it provides to get this working with non-Edge IE?
```javascript
import 'core-js/fn/array/find-index';
```

lovell:
@dbemiller To test, create a new App at https://developers.facebook.com/apps, add the Audience Network product to it, create a new Placement then paste the IDs here and we can get header video bidding enabled for them.

src/adapters/audienceNetwork.js
dbemiller: Yeah, that sounds good. Let's stick it in `polyfill.js`.

lovell: Thanks, updated.

dbemiller:
@lovell could you rebase? #1177 just got merged, so there are a bunch of conflicts now.

dbemiller:
@lovell Sorry for the delay. It turns out that someone on our team already had an app set up.

I don't know about Facebook's internal ad setup... but do we need a separate placement ID for different media types?

If not, we can reuse `1780499291962017_1780659308612682`.

wheresbarney:
@dbemiller - you should be able to request video bids with the same placement `1780499291962017_1780659308612682`. LMK if you have any problems getting bids with this.

dbemiller:
I'm getting a bid... but it doesn't look valid.

screen shot 2017-06-29 at 2 58 11 pm

Video bids should have a `vastUrl` associated with them.

lovell:
@dbemiller Please can you try adding `format: 'video'` to the `params`. (There will be a separate PR to update [the docs](https://github.com/prebid/prebid.github.io/blob/master/dev-docs/bidders/audienceNetwork.md) with this.)

dbemiller:
Looks good now.

lovell:
Thank you. I've created https://github.com/prebid/prebid.github.io/pull/293 for the related doc update.

PR #1296, opened 2017-06-15
Fixing Adyoulike adapter for Safari iOS7

dbemiller:
`performance` isn't supported by iOS7. `window.performance` gets rid of that error, but then the unit tests fail because PageRefreshed gets set to `true`.

Wrapping this in a try/catch to un-block the release. @jbAdyoulike you might want to take another pass at this to see if it's really what you want.


jbAdyoulike:
LGTM

PR #1299, opened 2017-06-15
Last bid timeout

### LINKED ISSUE #1298: bidmanager times out before saving last bid
## Type of issue
bug

## Description
When prebid receives a bid after the timeout, it executes the callbacks before saving the bid. This has the effect of ignoring the bid for those who rely on callbacks.

## Steps to reproduce
It's easy to reproduce if you put a breakpoint right before the responseTimestamp is set. 

### Expected results
All bids, including the one that triggered the timeout, are added to bids received when the callbacks are called. 

### Actual results
The last bid is ignored.

kitwestneat:
## Type of change

- [x] Bugfix

## Description of change
This fixes #1298 as well as refactoring some of the code to make it clearer. The bugfix is commit b5c30ff, the other commits refactor addBidResponse. 045e4ee could also be a bigfix -- it doesn't make sense to me that the bid events are emitted before the bid response is fully built, but I might be missing something.

## Other information
Fixes #1298

### DIFF
--- a/src/bidmanager.js
+++ b/src/bidmanager.js
@@ -92,50 +92,53 @@ exports.addBidResponse = function (adUnitCode, bid) {
     return;
   }
 
-  if (bid) {
-    if (bid.mediaType === 'native' && !nativeBidIsValid(bid)) {
-      utils.logError(`Native bid response does not contain all required assets. This bid won't be addeed to the auction`);
-      return;
-    }
-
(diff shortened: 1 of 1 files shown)


kitwestneat:
bidmanager has changed to the point where this doesn't make sense any more, so I'll close it

mkendall07:
@kitwestneat 
Hey if you want to open a new PR for just https://github.com/prebid/Prebid.js/issues/1298 that's fine.

PR #1301, opened 2017-06-15
Update serverbid adapter to use smartsync

micha:
## Type of change

- [ ] Bugfix
- [x] Feature
- [ ] New bidder adapter   
- [ ] Build related changes
- [ ] CI related changes
- [ ] Other

## Description of change

This change adds the ability to sync cookies with RTB partners via Serverbid. There are no changes to adapter parameters or basic functionality.

### DIFF
--- a/src/adapters/serverbid.js
+++ b/src/adapters/serverbid.js
@@ -7,9 +7,13 @@ import { ajax } from 'src/ajax';
 const ServerBidAdapter = function ServerBidAdapter() {
   const baseAdapter = Adapter.createNew('serverbid');
 
-  const BASE_URI = '//e.serverbid.com/api/v2';
+  const SERVERBID_BASE_URI = 'https://e.serverbid.com/api/v2';
+  const SMARTSYNC_BASE_URI = 'https://s.zkcdn.net/ss';
+  const SMARTSYNC_CALLBACK = 'serverbidCallBids';
+  const SMARTSYNC_TIMEOUT = 'serverbidSmartsyncTimeout';
 
-  const sizeMap = [null,
+  const sizeMap = [
+    null,
     '120x90',
     '120x90',
     '468x60',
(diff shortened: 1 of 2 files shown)


micha:
Hi! How's it looking? :)

src/adapters/serverbid.js
```diff
@@ -44,43 +48,67 @@ const ServerBidAdapter = function ServerBidAdapter() {
...
-      const data = {
-        placements: [],
-        time: Date.now(),
-        user: {},
-        url: utils.getTopWindowUrl(),
-        referrer: document.referrer,
-        enableBotFiltering: true,
-        includePricingData: true
+      window[SMARTSYNC_CALLBACK] = function() {
+        _callBids(params);
```
harpere: for performance reasons we don't want to delay the auctions, but rather initiate the auctions, then do the syncs.  I realize this means the sync won't be there for the first auction(s), but it will for subsequent auctions on the site.

src/adapters/serverbid.js
```diff
@@ -44,43 +48,67 @@ const ServerBidAdapter = function ServerBidAdapter() {
...
+      // and call window.serverBidContinueCallBids() when those pixels have
+      // finished loading.
+      const siteId = params.bids[0].params.siteId;
+      _appendScript(SMARTSYNC_BASE_URI + '/' + siteId + '.js');
+
+      // This is a failsafe in case cookie syncing is slow. The callback is a
+      // noop if window.serverBidContinueCallBids() has already been executed.
+      const sstimeout = window[SMARTSYNC_TIMEOUT] || ((params.timeout || 500) / 2);
+      setTimeout(function() {
+        window[SMARTSYNC_CALLBACK]();
```
harpere: As per the comment above, this timeout should no longer be necessary.

src/adapters/serverbid.js
micha: Could we make it optional? We have publishers requesting this.

harpere: Talked about it and decided that optional is ok as long as the publisher has to opt in.  
BTW -  the draft proposal for a 1.0 release is to force sync urls to be registered with prebid and then run later, after some timeout.  They'd also be restricted to pixels, iframes, or xhr (no jsonp for security reasons).   We'll have to think about your use case.  this will be put out for review at some point in the near future.

micha:
New commits to serverbid:serverbid-smartsync branch don't appear to be showing up in this PR. I will make a new one.

Issue #1304, opened 2017-06-17
AppnexusAST ends refresh auctions prematurely when trying to handle error


dmitriyshashkin:
## Type of issue
Bug

## Description
When AST executes callBids it saves request in `bidRequests` array. Bid requests accumulate upon subsequent callBids calls. So during refresh auction bidRequests holds requests from all previous auctions. If the server responds with an error, handleResponse method issues a response for every request logged in `bidRequests` array. And as bidRequests has requests from previous auctions, it produces too many responses. Which makes bid manager think that all bidders have sent responses even when they haven't. Auction ends prematurely which results in the revenue drop: we've seen almost 50% drop in our refresh revenues.

The easiest way to fix this is to clear `bidRequests` array in the beginning of `callBids` method.

## Steps to reproduce
1) Create test page with two bidders. First one uses AST adapter, the second one must have a delayed response or do not issue a bid response at all. AST one must have erroneous settings, so each request should result in an error.
2) Add a significant timeout (eg 10 seconds)
3) Refresh ads every n minutes, wait for the second iteration

### Expected results
The second auction ends after 10 seconds through the timeout mechanism

### Actual results
The second auction ends after AST receives an erroneous response

mkendall07:
Thanks for the bug report! Will look into this ASAP

PR #1305, opened 2017-06-17
Empty bidRequests in AST adapter each time new auction begins

dmitriyshashkin:
When error occurs adapter issues response for each bid in `bidRequests`
therefore bidRequests must be emptied each time new auction begins

Fixes https://github.com/prebid/Prebid.js/issues/1304

## Type of change

- [x] Bugfix

## Description of change
Empty bidRequests each time new auction begins, because when an error occurs adapter issues a response for each request in this array.

## Other information
https://github.com/prebid/Prebid.js/issues/1304

### DIFF
--- a/src/adapters/appnexusAst.js
+++ b/src/adapters/appnexusAst.js
@@ -34,6 +34,8 @@ function AppnexusAstAdapter() {
 
   /* Prebid executes this function when the page asks to send out bid requests */
   baseAdapter.callBids = function(bidRequest) {
+    bidRequests = {};
+
     const bids = bidRequest.bids || [];
     var member = 0;
     let userObj;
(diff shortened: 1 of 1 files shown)


src/adapters/appnexusAst.js
```diff
@@ -193,8 +195,11 @@ function AppnexusAstAdapter() {
       if (type === 'native') bid.mediaType = 'native';
       if (type === 'video') bid.mediaType = 'video';
       if (type === 'video-outstream') bid.mediaType = 'video-outstream';
-      const placement = bidRequests[bid.adId].placementCode;
-      bidmanager.addBidResponse(placement, bid);
+
+      if (bid.adId in bidRequests) {
```
mkendall07: is there ever a reason the `adId` isn't in `bidRequests` ?

dmitriyshashkin: Yes, if the response from the 1st auction is extremely slow and arrives after the 2nd auction has already started and `bidRequests` list has been emptied, then bidRequests will have adId's from the second auction, and adId from the first auction will have no corresponding entry.

mkendall07: yeah that makes sense. In that case we are OK to throw away the bid response so this is ok.

PR #1312, opened 2017-06-20
Added MobFox Adapter

francoroy:
## Type of change

- [ ] Bugfix
- [ ] Feature
- [ x] New bidder adapter   
- [ ] Build related changes
- [ ] CI related changes
- [ ] Other

## Description of change

- test parameters for validating bids ( for description visit http://docs.mobfox.com/docs/banners )
```
{
  bidder: 'mobfox',
  params: {
    s: "267d72ac3f77a3f447b32cf7ebf20673" // required - The hash of your inventory to identify which app is making the request,
    imp_instl: 1 // optional - set to 1 if using interstitial otherwise delete or set to 0
  }
}
```


src/adapters/mobfox.js
```diff
@@ -0,0 +1,167 @@
...
+  }
+
+  function sendBidRequest(bid) {
+    let requestParams = buildBidRequest(bid);
+    let queryString = buildQueryStringFromParams(requestParams);
+
+    ajax.ajax(`${BID_REQUEST_BASE_URL}?${queryString}`, {
+      success(resp, xhr) {
+        if (xhr.getResponseHeader("Content-Type") == "application/json")
+          resp = JSON.parse(resp);
```
jaiminpanchal27: wrap all JSON.parse in try catch block

francoroy:
Hi, any progress?
Thanks

jaiminpanchal27:
@francoroy Changes looks good. But not able to validate bids. 
I am using [hello-wold](https://github.com/prebid/Prebid.js/blob/master/integrationExamples/gpt/hello_world.html) test page

francoroy:
@jaiminpanchal27 Thanks. I can share my hello world page:
```
var adUnits = [{
        sizes: [[320, 480]],

        bids: [{
          bidder: 'mobfox',
          params: {
            s: '267d72ac3f77a3f447b32cf7ebf20673',
            imp_instl: 1
          }
        }]

      }];
```

jaiminpanchal27:
@francoroy Got it. Thanks. 
Now able to see ad with size 320x480.

dbemiller:
I did find a bit of an issue here...

If your server sends back a "no bid available" error, the adapter is still trying to add the bid to the bidmanager.

This was likely working when you wrote it, but changes in #1277 (in particular, [this line](https://github.com/prebid/Prebid.js/pull/1277/files#diff-96cd0debf7b766d72d2bcf987385e960L95)) cause it to break.

Since that was an unintended change, I'm replacing the old behavior in #1372... but I don't think this behavior is ideal. If an adapter doesn't have a bid to add, then it shouldn't be calling the `addBidResponse` method.

francoroy:
@dbemiller
"Since my changes were unintended, I'm replacing the old behavior in #1372... but I don't think this is ideal. If an adapter doesn't have a bid to add, then it shouldn't be calling the addBidResponse method"
Hi! so I don't understand, should I or should I not call addBidResponse when no bid is available?

dbemiller:
There's a `NO_BID` bid. For example: https://github.com/prebid/Prebid.js/blob/master/modules/appnexusAstBidAdapter.js#L175

francoroy:
@dbemiller this is what I already do in my onBidResponseError function:

  let bidResponse = bidfactory.createBid(CONSTANTS.STATUS.NO_BID, bid);
  bidResponse.bidderCode = BIDDER_CODE;
  bidmanager.addBidResponse(bid.placementCode, bidResponse);


Is it ok then?

dbemiller:
You're calling addBidResponse twice, and the second call uses an undefined bid. See the attached screenshot & test page.

screen shot 2017-07-20 at 4 32 45 pm

```
<!--

  This page calls a single bidder for a single ad slot. It can be considered a "hello world" example for using
  Prebid with the Google Publisher Tag.

  It also makes a good test page for new adapter PR submissions. Simply set your server's Bid Params object in the
  bids array inside the adUnits, and it will use your adapter to load an ad.

  NOTE that many ad servers won't send back an ad if the URL is localhost... so you might need to
  set an alias in your /etc/hosts file so that you can load this page from a different domain.

-->

<html>
<head>
    <script>
      var PREBID_TIMEOUT = 700;

      var adUnits = [{

        bids: [{
          bidder: 'mobfox',
          params: {
            s: "267d72ac3f77a3f447b32cf7ebf20673", // required - The hash of your inventory to identify which app is making the request,
            imp_instl: 1 // optional - set to 1 if using interstitial otherwise delete or set to 0
          }
        }]
      }];

      var pbjs = pbjs || {};
      pbjs.que = pbjs.que || [];

    </script>

    <script type="text/javascript" src="../../build/dev/prebid.js" async></script>
    <script>
      });

      pbjs.que.push(function() {
        pbjs.addAdUnits(adUnits);
        pbjs.requestBids({
        });
      });

          pbjs.que.push(function() {
            googletag.pubads().refresh();
          });
        });
      }

      setTimeout(function() {
        sendAdserverRequest();
      }, PREBID_TIMEOUT);

    </script>

    <script>
      (function () {
        var gads = document.createElement('script');
        gads.async = true;
        gads.type = 'text/javascript';
        var useSSL = 'https:' == document.location.protocol;
        gads.src = (useSSL ? 'https:' : 'http:') +
          '//www.googletagservices.com/tag/js/gpt.js';
        var node = document.getElementsByTagName('script')[0];
        node.parentNode.insertBefore(gads, node);
      })();
    </script>

    <script>

        googletag.enableServices();
      });
    </script>
</head>

<body>
<h2>Prebid.js Test</h2>
<h5>Div-1</h5>
    </script>
</div>
</body>
</html>
```

francoroy:
@dbemiller 
Hi, thanks for the review. 
I didn't manage to reproduce the issue with your test page but I found the bug and fixed it  (onBidResponseError was called twice on invalid json).

Thanks

dbemiller:
@francoroy I'm still getting the exact same error. I've attached a screenshot from the same spot, but looking at your code.

screen shot 2017-07-24 at 12 14 54 pm

Your server is responding with a 200, but the response is `{"error": "no Ad Available"}`. `transformResponse` returns undefined, and `onBidResponse` tries to add "undefined" as a bid response.

It reproduces in your unit tests, too:

```
 let bidResponse1 = stubAddBidResponse.getCall(0).args[1];
 expect(stubAddBidResponse.calledTwice).to.equal(true); // This passes, but addBidResponse should only be called once
expect(bidPlacementCode1).to.equal('test-gpt-div-1234');
```

francoroy:
@dbemiller Thanks, I've fixed it

PR #1324, opened 2017-06-26
Update serverbid adapter to use smartsync

micha:
## Type of change

- [ ] Bugfix
- [x] Feature
- [ ] New bidder adapter   
- [ ] Build related changes
- [ ] CI related changes
- [ ] Other

## Description of change

Adds the option to sync cookies with RTB partners (see also #1301).

### DIFF
--- a/modules/serverbidBidAdapter.js
+++ b/modules/serverbidBidAdapter.js
@@ -8,9 +8,12 @@ import adaptermanager from 'src/adaptermanager';
 const ServerBidAdapter = function ServerBidAdapter() {
   const baseAdapter = new Adapter('serverbid');
 
-  const BASE_URI = '//e.serverbid.com/api/v2';
+  const SERVERBID_BASE_URI = 'https://e.serverbid.com/api/v2';
+  const SMARTSYNC_BASE_URI = 'https://s.zkcdn.net/ss';
+  const SMARTSYNC_CALLBACK = 'serverbidCallBids';
 
-  const sizeMap = [null,
+  const sizeMap = [
+    null,
     '120x90',
     '120x90',
     '468x60',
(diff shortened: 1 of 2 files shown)


src/adapters/serverbid.js
```diff
@@ -44,44 +48,67 @@ const ServerBidAdapter = function ServerBidAdapter() {
...
-      }
+        const siteId = params.bids[0].params.siteId;
+        _appendScript(SMARTSYNC_BASE_URI + '/' + siteId + '.js');
 
-      if (data.placements.length) {
-        ajax(BASE_URI, _responseCallback, JSON.stringify(data), { method: 'POST', withCredentials: true, contentType: 'application/json' });
+        const sstimeout = window.SMARTSYNC_TIMEOUT || ((params.timeout || 500) / 2);
+        setTimeout(function() {
+          var cb = window[SMARTSYNC_CALLBACK];
+          window[SMARTSYNC_CALLBACK] = function() {};
```
harpere: minor - as defined earlier on line 54, `window[SMARTSYNC_CALLBACK]` already redefines itself to an empty function, so doing it here again is redundant.  So could just call `window[SMARTSYNC_CALLBACK]();`

micha: The reason for this `setTimeout()` call is to ensure that bids are fetched in the case that the callback (`window[SMARTSYNC_CALLBACK]`) is not called by the smartsync script in a timely manner. We also want to make sure that bids aren't fetched more than once. So consider the following cases:

1. smartsync calls the callback promptly: this means that the timeout on line 63 has not yet fired, and we want to fetch bids and then set the callback to a no-op, so we don't fetch bids again when the timeout on line 63 fires.
2. smartsync never calls the callback: this means that the timeout on line 63 will fire, fetch bids, and set the callback to a no-op.
3. smartsync calls the callback after the timeout has fired: in this case we have already fetched bids and we don't want to fetch them again, so the callback must have already been set to a no-op. The code that would have done that  is on line 65.

micha: Lines 64 and 65 handle the case where the callback is called after the timeout fires, so we don't fetch bids twice.

micha:
Tests added!

modules/serverbidBidAdapter.js
```diff
@@ -45,44 +49,67 @@ const ServerBidAdapter = function ServerBidAdapter() {
...
-          flightId: bid.params.flightId,
-          adId: bid.params.adId,
-          divName: bid.bidId,
-          adTypes: bid.adTypes || getSize(bid.sizes)
+      if (!window.SMARTSYNC) {
+        _callBids(params);
+      } else {
+        window[SMARTSYNC_CALLBACK] = function() {
+          window[SMARTSYNC_CALLBACK] = function() {};
+          _callBids(params);
```
matthewlane: @micha @harpere checking to make sure I understand this discussion from the previous PR https://github.com/prebid/Prebid.js/pull/1301#discussion_r123339300, what is the mechanism for publishers to opt in?

micha: Hi @matthewlane the publisher must set `window.SMARTSYNC = true` before bids are fetched.

mkendall07:
@micha it looks like the travis build is failing because of the node version. Can you rebase your branch off master?

jgrimes:
@mkendall07 I've rebased and fixed a linter issue (just an unused constant). Should be good to go!

Issue #1247, opened 2017-05-31
Prebid 1.0: setConfig API implementation


mkendall07:
https://gist.github.com/mkendall07/51ee5f6b9f2df01a89162cf6de7fe5b6#setconfig---new-api

PR #1325, opened 2017-06-26
Add setConfig method

matthewlane:
## Type of change
- Feature
- Refactoring

## Description of change
Adds `pbjs.setConfig` and `pbjs.getConfig` api methods for working with Prebid configuration.

Usage:
- `pbjs.setConfig({ debug: <true|false> });` to enable/disable logging (this will replace `pbjs.logging = <true|false>`, which still works for now but will be depreciated).

- `pbjs.setConfig({ <key>: <value> });` for setting arbitrary configuration

The  `getConfig` function is for retrieving the current configuration object or subscribing to configuration updates. When called with no parameters, the entire config object is returned. When called with a string parameter, a single configuration property matching that parameter is returned.

- `config.getConfig()` => get config object
- `config.getConfig('debug')` => get `debug` config

The `getConfig` function also contains a 'subscribe' ability that adds a callback function to a set of listeners that are invoked whenever `setConfig` is called. The subscribed function will be passed the options object that was used in the `setConfig` call. Individual topics can be subscribed to by passing a string as the first parameter and a callback function as the second

```JavaScript
// subscribe to all configuration changes
getConfig((config) => console.log('config set:', config));

// subscribe to only 'logging' changes
getConfig('logging', (config) => console.log('logging set:', config));

// unsubscribe
const unsubscribe = getConfig(...);
unsubscribe(); // no longer listening
```

## Other information
Implements #1247

### DIFF
--- a/src/adaptermanager.js
+++ b/src/adaptermanager.js
@@ -243,7 +243,9 @@ exports.enableAnalytics = function (config) {
 };
 
 exports.setBidderSequence = function (order) {
-  _bidderSequence = order;
+  if (order === CONSTANTS.ORDER.RANDOM) {
+    _bidderSequence = order;
+  }
 };
 
 exports.setS2SConfig = function (config) {
--- a/src/config.js
+++ b/src/config.js
@@ -1,0 +1,64 @@
+const utils = require('./utils');
+
+/**
+ * Methods for managing Prebid configuration
+ */
+
+let setConfigApi;
+let debug;
+
+export function getDebugStatus() {
+  // if debug not set, pub might still be using pbjs.logging, which may still
+  // be used until deprecation
+  if (!setConfigApi) { return $$PREBID_GLOBAL$$.logging; }
+
+  return debug;
+}
+
+export function setDebugStatus(_debug) {
+  debug = _debug;
+}
+
+export function setConfig(options) {
+  if (typeof options !== 'object') {
+    utils.logError('setConfig options must be an object');
+  }
+
+  // while old config properties/methods are in depecration window, set this
+  // to signal to getter/setter methods where to check state
+  setConfigApi = true;
+
+  // TODO: fill out remaning config options
+  // if (options.bidderTimeout) {
+  //   $$PREBID_GLOBAL$$.bidderTimeout = options.bidderTimeout;
+  // }
+
+  // `debug` is equivalent to previous `pbjs.logging` property
+  if (options.debug) {
+    setDebugStatus(options.debug);
+  }
+
+  // if (options.publisherDomain) {
+  //   $$PREBID_GLOBAL$$.publisherDomain = options.publisherDomain;
+  // }
+
+  // if (options.cookieSyncDelay) {
+  //   $$PREBID_GLOBAL$$.cookieSyncDelay = options.cookieSyncDelay;
+  // }
+
+  // if (options.priceGranularity) {
+  //   $$PREBID_GLOBAL$$.setPriceGranularity(options.priceGranularity);
+  // }
+
+  // if (options.enableSendAllBids) {
+  //   $$PREBID_GLOBAL$$.enableSendAllBids();
+  // }
+
+  // if (options.bidderSequence) {
+  //   $$PREBID_GLOBAL$$.setBidderSequence(options.bidderSequence);
+  // }
+
+  // if (options.s2sConfig) {
+  //   $$PREBID_GLOBAL$$.setS2SConfig(options.s2sConfig);
+  // }
+}
--- a/src/config.js
+++ b/src/config.js
@@ -1,41 +1,34 @@
 const utils = require('./utils');
 
-/**
- * Methods for managing Prebid configuration
- */
+export let config = {
+  _debug: false,
 
-let setConfigApi;
-let debug;
+  get debug() {
+    // this is deprecated but may still be used during deprecation window
+    // `debug` is equivalent to legacy `pbjs.logging` property
+    if ($$PREBID_GLOBAL$$.logging || $$PREBID_GLOBAL$$.logging == false) {
+      return $$PREBID_GLOBAL$$.logging;
+    }
 
-export function getDebugStatus() {
-  // if debug not set, pub might still be using pbjs.logging, which may still
-  // be used until deprecation
-  if (!setConfigApi) { return $$PREBID_GLOBAL$$.logging; }
+    return this._debug;
+  },
 
-  return debug;
-}
-
-export function setDebugStatus(_debug) {
-  debug = _debug;
-}
+  set debug(val) {
+    this._debug = val;
+  },
+};
 
 export function setConfig(options) {
   if (typeof options !== 'object') {
     utils.logError('setConfig options must be an object');
   }
 
-  // while old config properties/methods are in depecration window, set this
-  // to signal to getter/setter methods where to check state
-  setConfigApi = true;
-
-  // TODO: fill out remaning config options
   // if (options.bidderTimeout) {
   //   $$PREBID_GLOBAL$$.bidderTimeout = options.bidderTimeout;
   // }
 
-  // `debug` is equivalent to previous `pbjs.logging` property
   if (options.debug) {
-    setDebugStatus(options.debug);
+    config.debug = options.debug;
   }
 
   // if (options.publisherDomain) {
--- a/src/config.js
+++ b/src/config.js
@@ -12,6 +12,7 @@ const utils = require('./utils');
 const DEFAULT_DEBUG = false;
 const DEFAULT_BIDDER_TIMEOUT = 3000;
 const DEFAULT_PUBLISHER_DOMAIN = window.location.origin;
+const DEFAULT_COOKIESYNC_DELAY = 100;
 
 let config = {
   // `debug` is equivalent to legacy `pbjs.logging` property
@@ -43,6 +44,15 @@ let config = {
   set publisherDomain(val) {
     this._publisherDomain = val;
   },
+
+  // delay to request cookie sync to stay out of critical path
+  _cookieSyncDelay: DEFAULT_COOKIESYNC_DELAY,
+  get cookieSyncDelay() {
+    return $$PREBID_GLOBAL$$.cookieSyncDelay || this._cookieSyncDelay;
+  },
+  set cookieSyncDelay(val) {
+    this._cookieSyncDelay = val;
+  },
 };
 
 export function getConfig(option) {
@@ -58,7 +68,7 @@ export function setConfig(options) {
   // [x] $$PREBID_GLOBAL$$.bidderTimeout
   // [x] $$PREBID_GLOBAL$$.logging (renamed `debug`)
   // [x] $$PREBID_GLOBAL$$.publisherDomain
-  // [ ] $$PREBID_GLOBAL$$.cookieSyncDelay
+  // [x] $$PREBID_GLOBAL$$.cookieSyncDelay
   // [ ] $$PREBID_GLOBAL$$.setPriceGranularity (function(granularity), `priceGranularity`)
   // [ ] $$PREBID_GLOBAL$$.enableSendAllBids (function)
   // [ ] $$PREBID_GLOBAL$$.setBidderSequence (function(order), `bidderSequence`)
--- a/src/config.js
+++ b/src/config.js
@@ -7,13 +7,12 @@
  * Defining and access properties in this way is now deprecated, but these will
  * continue to work during a deprecation window.
  */
-
 const utils = require('./utils');
 
 const DEFAULT_DEBUG = false;
 const DEFAULT_BIDDER_TIMEOUT = 3000;
 
-export let config = {
+let config = {
   _debug: DEFAULT_DEBUG,
   get debug() {
     // `debug` is equivalent to legacy `pbjs.logging` property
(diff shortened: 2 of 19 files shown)


dbemiller:
Just to a quick sanity check...

As written, no prebid files will be able to read from the config "on load." Any `getConfig` calls will have to be nested inside functions which are only called through publisher-called functions... because otherwise we'll end up reading the config values before the page gets a chance to set them.

Did you guys already consider & accept that limitation?

mkendall07:
@dbemiller 
yeah good call it. We have discussed it a bit and have talked about alternative approaches to have "on load" configs (data-* attributes for example) - but for the most part this is backwards compatitible since most of these values are set after the pbjs queue executes - we just expect the `setConfig` call to be made before the auction starts.

dbemiller:
in that case... I need to challenge this.

```
pbjs.setConfig(cfg);
pbjs.thingWhichNeedsConfig();
```

can always be replaced by:

```
pbjs.thingWhichNeedsConfig(cfg);
```

The latter is less error-prone:
- The caller can't screw up the ordering
- The implementer can't use a variable they don't have yet 

Easier to test:
- No (explicit) global state which tests have to set/reset after each use
- Each function's JSDoc can define only the `cfg` options it cares about. Publishers can define a single object, if we choose the expected property names well so that they don't conflict.

...and works better if people want to change the config up when running multiple auctions.

It's also possible to maintain backwards compatibility. For example, we could start by implementing:

```
function thingWhichNeedsConfig(cfg) {
  $$PREBID_GLOBAL$$.debug = oldDebug;
  ... do stuff ...
}
```

Then we can rewrite the `thingWhichNeedsConfig` implementation to stop using the global state at our leisure, and then delete that first line in 1.0.

Are there any real advantages to setting the state beforehand?

dbemiller:
Noting the discussion in standup in writing. There _is_ one benefit of `setConfig()`. If the user makes many API calls without changing the config, then it's more convenient to call once:

```
pbjs.setConfig(cfg);
pbjs.doOneThing();
pbjs.doAnotherThing();
pbjs.doAThirdThing();
```

is a better user experience than:

```
pbjs.doOneThing(cfg);
pbjs.doAnotherThing(cfg);
pbjs.doAThirdThing(cfg);
```

I talked a bit more with @mkendall07 afterwards, and decided to amend my suggestion. Suppose we do:

```
var configured = pbjs.withConfig(cfg);
configured.doOneThing();
configured.doAnotherThing();
configured.doAThirdThing();
```

Then the state has only been passed once, and we still maintain all the benefits I mentioned above (All the config-dependent methods would not exist on pbjs. They would exist on an object returned by the `withConfig` function).

snapwich:
in the 
```javascript
var configured = pbjs.withConfig(cfg);
configured.doOneThing();
configured.doAnotherThing();
configured.doAThirdThing();
```
pattern, what is `withConfig` returning?  Is it a separate instance than the `pbjs` global?  If it is separate, how does that affect adapters and such that use the `pbjs` global,  would they continue using a separate instance, or would they need to somehow get access to the configured instance?

dbemiller:
That's a good question... in my view, `withConfg` may be the **only** function on the global.

The only two reasons I know of why adapters use the global today are:

1. To support JSONP
2. To view config values

(1) is unsupported in Prebid 1.0 anyway.
(2) could be handled by that `registerAdapterLoader((cfg) => Adapter)` pattern we talked about before.

Using more realistic sample code, core would look something like this:

```
// loader is a (cfg) => Adapter, and called in each adapter file.
export function registerAdapterLoader(loader) { }

$$PREBID_GLOBAL$$.withConfig = function(cfg) {
  var adapters = adapterLoaders.map(loader => loader(cfg));
  return {
    doAuction: function(auctionParams) {
      adapters.forEach( /* call bids & do other stuff here */);
    }
  }
}
```

mkendall07:
We paused dev on this but will pick it up soon. 

I'd like to propose we continue with the existing API for the following reasons:

1) Saves us time to completion - the alternative API with a `pbjs` instance would be a larger scope refactor
2) Maintaining backward compatibility before 1.0 is more difficult (I think). 
3) Ease of publisher integration will be more difficult (more change points) with  alternative API.

src/config.js
```diff
@@ -0,0 +1,77 @@
...
+ * Defining and access properties in this way is now deprecated, but these will
+ * continue to work during a deprecation window.
+ */
+
+const utils = require('./utils');
+
+const DEFAULT_DEBUG = false;
+const DEFAULT_BIDDER_TIMEOUT = 3000;
+
+export let config = {
```
snapwich: I think we need some way to subscribe to changes to this object.  For instance, the currency module (in review still) will need to kick off a request to go and download the currency file as soon as its configuration is set.  But without some sort of subscription model, it won't know when that is.

src/config.js
```diff
@@ -0,0 +1,77 @@
...
+    this._bidderTimeout = val;
+  },
+};
+
+export function setConfig(options) {
+  if (typeof options !== 'object') {
+    utils.logError('setConfig options must be an object');
+  }
+
+  if (options.bidderTimeout) {
```
matthewlane: Right, was meaning to squash these, will use `Object.assign`, thanks

src/config.js
```diff
@@ -0,0 +1,120 @@
...
+/*
+ * Adds a function to a set of listeners that are invoked whenever `setConfig`
+ * is called. The subscribed function will be passed the options object that
+ * was used in the `setConfig` call.
+ *
+ * Example use:
+ * subscribe((config) => console.log('config set:', config));
+ * subscribe(({ debug }) => debug ? console.log('debug set:', debug) : '');
+ */
+export function subscribe(listener) {
```
snapwich: I think listing to a label would be a slight improvement.  Say I'm the currency module then I subscribe with
```Javascript
subscribe("currency", function() { ... });
```
and then I'm only alerted when the currency configuration is updated, and I'm only passed the relevant currency configuration.
```Javascript
// "currency" subscribe handler called
pbjs.setConfig({
  currency: {
    ...
  }
});

// "currency" subscribe handler not called
pbjs.setConfig({
  bidderTimeout: 3000
})
```
That way if setconfig is used multiple times to update difference configs we clean up a lot of the noise of alerting every listener.

However, if the first argument is the callback function we could still have it behave as it does now where it gets called for any configuration change and sends the whole config.

src/config.js
```diff
@@ -0,0 +1,156 @@
...
+
+const DEFAULT_DEBUG = false;
+const DEFAULT_BIDDER_TIMEOUT = 3000;
+const DEFAULT_PUBLISHER_DOMAIN = window.location.origin;
+const DEFAULT_COOKIESYNC_DELAY = 100;
+const DEFAULT_ENABLE_SEND_ALL_BIDS = false;
+
+const ALL_TOPICS = '*';
+
+let listeners = [];
```
dbemiller: Chatted about this a little bit offline... For the sake of testability, concurrent auctions, and infinite scroll, I think we should strongly avoid adding global, mutable state to the project. Especially in new modules.

I suggested an alternate pattern, which @matthewlane sounded to like too. I'm pretty sure it'll require a `getConfig` on `$$PREBID_GLOBAL$$`, which kinda sucks... but I think it's a worthwhile tradeoff.

I can describe it here if anyone else cares or wants to hear it.

src/adaptermanager.js
```diff
@@ -246,7 +246,9 @@ exports.enableAnalytics = function (config) {
 };
 
 exports.setBidderSequence = function (order) {
-  _bidderSequence = order;
+  if (order === CONSTANTS.ORDER.RANDOM) {
```
snapwich: looks like it was moved from https://github.com/prebid/Prebid.js/pull/1325/files#diff-f8e049a1b6d4f9aa96fca41d2f7aa11dL697

I don't know _why_ it did that check, but it doesn't look like this pull-request changes any functionality.

matthewlane: Right, intention was to consolidate the `setBidderSequence` behavior into adaptermanager rather than having it spread across two functions. Noticed this while looking at this function when adding the `bidderSequence` config. Open to dropping this change if it's too out of scope

src/config.js
```diff
@@ -0,0 +1,156 @@
...
+ * the first parameter.
+ *
+ * Example use:
+ * // subscribe to all configuration changes
+ * subscribe((config) => console.log('config set:', config));
+ *
+ * // subscribe to only 'logging' changes
+ * subscribe('logging', (config) => console.log('logging set:', config));
+ */
+export function subscribe(topic, listener) {
```
snapwich: I like how this `subscribe` is implemented now.  👍 

However, I think it might be nice to merge its functionality with `getConfig` for a few reasons.  For one, I think that `getConfig` is a more descriptive name, especially if we export this on the public API (which I think we should).  Also, it really is fulfilling the same purpose as `getConfig`; it's just implemented as a push rather than a pull.

I propose we merge `subscribe` into `getConfig`.  if `getConfig` is called with no arguments or string, behave as `getConfig` does now, if it's called with a function or a string and a function, then it behaves as `subscribe` behaves now.

src/config.js
snapwich: Can you describe what you're thinking?  I think what we have here is pretty good.  

I don't really consider this global state as its encapsulated to the config module; however, it is "side-effecty", but I'm not sure that's a terrible thing.  However, it does need a way to clean up its side-effects for testing purposes, i.e. we need some way to remove listeners.

dbemiller: In this file:

```
export function newConfig() {
  let listeners = [];
  return {
    getConfig() { ... }
    setConfig(options) { ... }
    subscribe(topic, listener) { ... }
  };
}
```

In `prebid.js`:

```
import { newConfig } from './config';

const config = newConfig();
$$PREBID_GLOBAL$$.getConfig = config.getConfig
$$PREBID_GLOBAL$$.setConfig = config.setConfig
```

In modules:

```
const getConfig = $$PREBID_GLOBAL$$.getConfig
```

Unit tests don't need to worry about side-effects because they can call `newConfig()` each time.

dbemiller: I agree that what we have here is workable. Honestly, the question which is really in the back of my mind is "how do we get to an API like @protonate suggested, without it being such a massive project?" Something like:
```
prebid.push(function() {
  prebid.doAuction().then(renderAd);
}
```

As written, this config module can only be used in a project which is driven by global state. That's fine... because the current project is driven by global state.

Using my suggestion, this config module could be used in both systems. 

```
function doAuction() {
  const config = newConfig();
  ...auction impl here...
}
```

If we keep adding new modules which rely on global state, then certain features (like infinite scroll) become more and more difficult to support with time.

If we add modules which don't rely on global state, then we can reuse them out of the box if(/when) we migrate in that direction.

snapwich: I don't think this would really gain us anything.  You would have to move the `let config` into the `newConfig` closure as well, right?  So then we still have a "global" state for Prebid.js, it's just inside of a closure now making it harder to get at.  Also, this means `getConfig`, `setConfig`, etc would need to be updated to pass in the closure reference to that state so they could interact with it.  

```
import { newConfig } from './config';

const config = newConfig();
$$PREBID_GLOBAL$$.getConfig = config.getConfig
$$PREBID_GLOBAL$$.setConfig = config.setConfig
```
Wouldn't each of the tests have to do this as well in order to reset global Prebid.js state?  I think i'd be a little more straight forward to have a `resetConfig()` that clears any config state, in this case `config` and `listeners`.

src/prebid.js
```diff
@@ -645,8 +647,9 @@ $$PREBID_GLOBAL$$.setPriceGranularity = function (granularity) {
   }
 };
 
+/** @deprecated - use pbjs.setConfig({ enableSendAllBids: <true|false> }) */
```
matthewlane: Current name comes from https://gist.github.com/mkendall07/51ee5f6b9f2df01a89162cf6de7fe5b6#setconfig---new-api, ccing @mkendall07 for feedback

src/prebid.js
```diff
@@ -745,6 +748,14 @@ $$PREBID_GLOBAL$$.setS2SConfig = function(options) {
   adaptermanager.setS2SConfig(config);
 };
 
+/**
+ * Set Prebid config options
+ * @param {object} options
```
matthewlane: Couldn't quite parse how best to doc this, will try in future

src/prebid.js
```diff
@@ -757,6 +761,18 @@ $$PREBID_GLOBAL$$.setS2SConfig = function(options) {
   adaptermanager.setS2SConfig(config);
 };
 
+/**
+ * Get Prebid config options
+ * @param {object} options
+ */
+$$PREBID_GLOBAL$$.getConfig = config.getConfig;
```
dbemiller: This doesn't need to be on the global API, does it?

src/prebid.js
```diff
@@ -757,6 +761,18 @@ $$PREBID_GLOBAL$$.setS2SConfig = function(options) {
...
+
+/**
+ * Set Prebid config options
+ * @param {object} options
```
dbemiller: More specific JSDocs would be super helpful here.

I understand we should probably merge this PR to unblock people, though... so can we make it a priority shortly after? It improves the developer experience a ton.

src/prebid.js
snapwich: I was the one who suggested this.  I think it's a small thing but makes Prebid.js more extensible by third-party libraries.  So overall I think it's a small net-gain for a one-liner.

PR #1337, opened 2017-06-29
Add s2s and user sync features to roxot analytics adapter

sergey-roxot:
## Type of change

- [ ] Bugfix
- [x] Feature
- [ ] New bidder adapter   
- [ ] Build related changes
- [ ] CI related changes
- [ ] Other

## Description of change
Added features to roxot analytics adapter:
- Tracking bidders using Prebid Server by adding '(s2s)' postfix to bidder code.
- Setting cookie uid to visitors of sites with enabled roxot analytics adapter.


mkendall07:
Nice! I'm adding a `src` attribute to the existing bids. Would you be able to use that instead of piping the whole configuration object?

https://github.com/prebid/Prebid.js/blob/6f06a9ef7ef912cc43235e761572577ca1244626/src/adaptermanager.js#L110

sergey-roxot:
```src``` attribute is good. I'm going to check it right now and fix our feature after ```src``` improvement will be in master branch.

**Comment by sergey-roxot** [comment]
When can we expect https://github.com/prebid/Prebid.js/pull/1330 to be merged to the master?

**Comment by mkendall07** [comment]
@sergey-roxot 
Just merged.

**Comment by sergey-roxot** [comment]
We had removed s2s feature and now using source argument. Our pull request includes only user sync feature now.

**Comment by sergey-roxot** [comment]
When do you plan to review our pull request?

File: modules/roxotAnalyticsAdapter.js [hunk 1]
```diff
@@ -37,12 +37,16 @@ function buildEventStack() {
 
 function send(eventType, data, sendDataType) {
   let fullUrl = url + '?publisherIds[]=' + initOptions.publisherIds.join('&publisherIds[]=') + '&host=' + window.location.hostname;
-
-  ajax(
-    fullUrl,
-    (result) => utils.logInfo('Event ' + eventType + ' sent ' + sendDataType + ' to roxot prebid analytic with result' + result),
-    JSON.stringify(data)
-  );
+  let xhr = new XMLHttpRequest();
```
mkendall07: what's the reason you can't use ajax?

mkendall07: FYI, you can turn on cookies:
```    ajax(ENDPOINT, handleResponse, payload, {
        contentType: 'text/plain',
        withCredentials: true
      });```

sergey-roxot: Depending on the publisher's setup, some requests are processed longer than others. But when it takes more than 3 sec to record all events, requests are canceled by the recently implemented timeout to ajax requests. As a result, we may lose some of the data. While optimizing our system and speeding up data processing, we want to make sure nothing is lost.

mkendall07: Fair point. We should allow analytics adapters to override the timeout. For now this is fine.

PR #1342, opened 2017-06-30
Add network zone matching to criteo adapter

Swiiip:
## Type of change
- [ x] Feature

## Description of change
Add network zone matching feature for Criteo Adapter

- contact email of the adapter’s maintainer: i.caroulle@criteo.com h.duthil@criteo.com

### DIFF
--- a/modules/criteoBidAdapter.js
+++ b/modules/criteoBidAdapter.js
@@ -16,7 +16,7 @@ var CriteoAdapter = function CriteoAdapter() {
       _pushBidRequestEvent(params);
       adloader.loadScript(
         _publisherTagUrl,
-        function () {},
+        function () { },
         true
       );
     } else {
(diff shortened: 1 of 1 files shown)


modules/criteoBidAdapter.js
```diff
@@ -35,22 +35,33 @@ var CriteoAdapter = function CriteoAdapter() {
...
-            bid.transactionId
+            bid.transactionId,
+            sizes.map((size) => {
+              return { width: size[0], height: size[1] }
+              }
+            )
           )
         );
 
+        if (!networkid && bid.params.networkId != undefined) {
```
dbemiller: Is there any way for `networkid` to be defined here? I don't see any assignments between here and its declaration.

mkendall07: agree, doesn't need to be there.

Swiiip: Updated

muuki88:
@Swiiip thanks for this addition. We are currently testing the sizes parameter and are wondering how the criteo ad server must be configured to handle multiple sizes per zoneId.

At the moment we send a bid request per zoneId/size combination, which leads to multiple bid responses per ad unit. Prebid is not designed to handle this and chooses the last bid response, not taking the cpm into account.

PR #1346, opened 2017-07-05
Upgrade eslint

dbemiller:
## Type of change
Updating the eslint dependency, and replacing deprecated things.

## Description of change
As @matthewlane pointed out in the initial diff, `.eslintrc` has been deprecated. This replaces it with `.eslintrc.js`, which is supported.

I also updated the eslint-related dependencies, since some new major versions got released. The bulk of the changes are from running `./node_modules/eslint/bin/eslint.js dirs --fix` to update the code, making the new style rules pass.


dbemiller:
yes... thank you. I always forget about yarn. Will do

PR #1348, opened 2017-07-06
Reduce dependency of SAS adapter.

sebastienferry:
## Type of change

## Description of change

The goal of this refactoring is to use the Smart AdServer Prebid Javascript library to build the ad-request independently from the adapter code.


dbemiller:
Hey: just so you're aware, the Prebid project will soon refuse to let adapters load external scripts. See the `Performance/Adapters/Requirements` section under #891.

Technically we're still allowing it... but with that in mind, do you still want to make these changes? On the current timeline, you'll probably be expected to replace them sometime in Q4 of this year.

sebastienferry:
Thank you @dbemiller. We will review the requirements and adapt our strategy. I'm closing the request for now.

PR #1352, opened 2017-07-06
Map native icon parameter and validate returned asset values

matthewlane:
## Type of change
- Bugfix

## Description of change
This updates appnexusAst to map the `icon` native parameter to pass a `sizes` field on the request to the endpoint, which is a required field when requesting that asset.

Also fixes a bug in the native validation function that was allowing returned native assets with a key but no value to pass validation when `required: true` was set for that asset.

### DIFF
--- a/modules/appnexusAstBidAdapter.js
+++ b/modules/appnexusAstBidAdapter.js
@@ -16,10 +16,11 @@ const NATIVE_MAPPING = {
   body: 'description',
   image: {
     serverName: 'main_image',
-    serverParams: {
-      required: true,
-      sizes: [{}]
-    }
+    serverParams: { required: true, sizes: [{}] }
+  },
+  icon: {
+    serverName: 'icon',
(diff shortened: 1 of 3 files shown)


modules/appnexusAstBidAdapter.js
```diff
@@ -16,10 +16,11 @@ const NATIVE_MAPPING = {
...
     serverName: 'main_image',
-    serverParams: {
-      required: true,
-      sizes: [{}]
-    }
+    serverParams: { required: true, sizes: [{}] }
+  },
+  icon: {
+    serverName: 'icon',
+    serverParams: { sizes: [{}] }
```
mkendall07: Should this have `required : true` ?

matthewlane: In testing the `required` param didn't seem to affect the response, but did add it in to be consistent with `image` type

PR #1326, opened 2017-06-27
Spotx video Adapter

npeceniak:
## Type of change

- [ ] Bugfix
- [ ] Feature
- [x] New bidder adapter   
- [ ] Code style update (formatting, local variables)
- [ ] Refactoring (no functional changes, no api changes)
- [ ] Build related changes
- [ ] CI related changes
- [ ] Other

## Description of change

Add Spotx video bidder adapter

- test parameters for validating bids
```
{
              bidder: 'spotx',
              params: {
                placementId: "123456789",
                video: {
                  channel_id: 85394,
                  video_slot: 'contentElement',
                  slot: 'content',
                  hide_skin: false,
                  autoplay: true,
                  ad_mute: false,
                }
              }
}
```

Be sure to test the integration with your adserver using the [Hello World](/integrationExamples/gpt/hello_world.html) sample page.

- contact email of the adapter’s maintainer
npeceniak@spotx.tv
- [x] official adapter submission

PR #1362, opened 2017-07-10
Fixing unit tests - added stubs for external server calls

jaiminpanchal27:
## Type of change

- [x ] Bugfix

## Description of change

Spotx video adapter was making external server calls from unit tests. Unit tests can fail if their server is down. 
Added sinon stubs for spotx external api to prevent potential build failure

## Other information
#1326


dbemiller:
Thanks!

PR #1368, opened 2017-07-11
Fix import and module reference

matthewlane:
## Type of change
- Bugfix

## Description of change
Fixes function that was being imported as a default, but wasn't a default export. Also updates util references so that exported functions work when imported with ES module syntax

### DIFF
--- a/src/utils.js
+++ b/src/utils.js
@@ -455,7 +455,7 @@ exports.insertElement = function(elm, doc, target) {
 
 exports.insertPixel = function (url) {
   const img = new Image();
-  img.id = this.getUniqueIdentifierStr();
+  img.id = _getUniqueIdentifierStr();
   img.src = url;
   img.height = 0;
   img.width = 0;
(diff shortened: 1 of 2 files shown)


src/utils.js
```diff
@@ -466,7 +466,7 @@ exports.insertPixel = function (url) {
     } catch (e) {
     }
   };
-  this.insertElement(img);
+  exports.insertElement(img);
```
mkendall07: should this reference be updated too? https://github.com/prebid/Prebid.js/blob/35e68ae0d5deb648fc9178b9730362f03d10de5e/src/utils.js#L482
(no pun intended)

dbemiller: For the record.... most of the functions in this file actually do this. I fixed a couple in the video cache PR too.

Some are easy to fix, but some aren't. Not sure which category `insertElement` is in, or how deep down the rabbit hole you want this to go... or if we should just file a ticket and schedule the time to fix them all.

matthewlane: Updated the other `insertElement` reference as well. I'd like to go through utils and update remaining `this` references at some point, or convert to ES module syntax entirely, but for now just scoped this PR to `insertElement` references

PR #1370, opened 2017-07-12
Update ThoughtLeadr adapter

ilya-pirogov:
## Type of change

- [x] Bugfix

## Description of change

Update ThoughtLeadr adapter


danclien:
Hey @matthewlane, just wondering if there's anything we can help with to get this merged. Thanks in advance!

matthewlane:
Thanks for the update, this PR will be assigned for review and the reviewer may have feedback to help merge

modules/thoughtleadrBidAdapter.js
```diff
@@ -1,29 +1,66 @@
...
+
+var COOKIE_SYNC_ID = 'tldr-cookie-sync-div';
+var UID_KEY = 'tldr_uid';
+var URL_API = 'tldr' in window && tldr.config.root_url ? tldr.config.root_url : '//a.thoughtleadr.com/v4/';
+var URL_CDN = 'tldr' in window && tldr.config.cdn_url ? tldr.config.cdn_url : '//cdn.thoughtleadr.com/v4/';
 var BID_AVAILABLE = 1;
+var BID_UNAVAILABLE = 2;
+
+function getVal(key) {
+  if ('localStorage' in window) {
```
dbemiller: Looks like this feature test doesn't work so well. See [this gist](https://gist.github.com/paulirish/5558557) for some examples, and [MDN's suggestion](https://developer.mozilla.org/en-US/docs/Web/API/Web_Storage_API/Using_the_Web_Storage_API#Testing_for_availability) for a more rigorous test.

modules/thoughtleadrBidAdapter.js
```diff
@@ -36,47 +73,50 @@ var ThoughtleadrAdapter = (function () {
...
-      if (params.bid.code === BID_AVAILABLE) {
-        bidObject = bidfactory.createBid(params.bid.code);
+      if (wonBid && wonBid.amount) {
+        bidObject = bidfactory.createBid(BID_AVAILABLE);
         bidObject.bidderCode = 'thoughtleadr';
-        bidObject.cpm = params.bid.cpm;
-        bidObject.ad = params.bid.ad;
+        bidObject.cpm = wonBid.amount;
+        bidObject.ad =
+          '<script src="' + URL_CDN + 'bid.js" header-bid-token="' + wonBid.header_bid_token + '"></script>';
```
dbemiller: This is also vulnerable to cross-script injection. Creating a node object and using `setAttribute` makes it safe. (See the same OWASP link I mentioned above)

modules/thoughtleadrBidAdapter.js
```diff
@@ -1,29 +1,66 @@
...
+  const iframe = document.createElement("iframe");
+  iframe.style.width = "0";
+  iframe.style.height = "0";
+  iframe.style.border = "0";
+
+  iframe.src = "javascript:false";
+  container.appendChild(iframe);
+
+  const doc = iframe.contentWindow.document;
+  doc.open().write(`<html><head></head><body>${html}</body></html>`);
```
ilya-pirogov: > Couldn't html include the value `</body></head></html></iframe><script>${maliciousCode}</script>`?

That code won't work inside `iframe`. Besides I doubt that code can contain a malicious code. But it can contain  `<script>`, ``, other `<iframe>`, or any combinations of these tags tags. So I can't just use `document.createElement("script")`. 

@danclien Please correct me if I'm wrong.

ilya-pirogov:
@dbemiller Fixed all issues except `document.write()`

modules/thoughtleadrBidAdapter.js
danclien: @ilya-pirogov, what's the alternative? Changing the cookie syncs data from raw HTML tags to URLs that can be rendered by the adapter?

dbemiller: Prebid doesn't support IE < 10... so the [DOMParser](https://developer.mozilla.org/en-US/docs/Web/API/DOMParser) looks like it could do the trick?

As long as your unit tests are thorough, we'll see any errors when we `gulp test --browserstack` before release.

Shakakai: @ilya-pirogov lets implement the DOMParser approach rather than using document.write.

ilya-pirogov: Unfortunately, `DOMParser` didn't help because it doesn't execute any script tags. So I implemented that with `.innerHTML` and additional handling.

dbemiller:
This... doesn't seem any safer than what you had before?

I added this line to your bid adapter:

`window.thoughtleadrFriendlyFrame = writeFriendlyFrame;`

And the following sample webpage (added to `integrationExamples/gpt/whatever.html`)

```
<html>
<head>
    <script type="text/javascript" src="../../build/dev/prebid.js" async></script>
</head>

<body>
<div id="anchor"></div>
</body>
</html>
```

After leading the page, type this into the terminal:

```
thoughtleadrFriendlyFrame('</body></html></iframe><script>alert("malicious code executed.");</script>', document.getElementById('anchor'));​
```

So... `innerHTML` is definitely vulnerable. I didn't try it on the old version of your code, but I'm pretty certain than `docuemnt.write` is as well. I don't _think_ that `DOMParser` and `appendChild` are... but I'm happy with however you implement it, as long as it's not vulnerable to cross-script injection from server code.

modules/thoughtleadrBidAdapter.js
dbemiller: Ok, that's fine.

I chatted with our team lead, and while he confirmed that there _is_ a security issue here, we have the same vulnerability throughout prebid.

We're changing this in #891, when we'll merge #1229 and then (at some point in the future) prevent adapters from executing any external JS at all... but for now, we're leaving it to publishers to only load bidders whose code they trust.

modules/thoughtleadrBidAdapter.js
```diff
@@ -1,82 +1,171 @@
...
+const UID_KEY = 'tldr_uid';
+const URL_API = 'tldr' in window && tldr.config.root_url ? tldr.config.root_url : '//a.thoughtleadr.com/v4/';
+const URL_CDN = 'tldr' in window && tldr.config.cdn_url ? tldr.config.cdn_url : '//cdn.thoughtleadr.com/v4/';
+const BID_AVAILABLE = 1;
+const BID_UNAVAILABLE = 2;
+
+function storageAvailable(type) {
+  try {
+    const storage = window[type],
+      x = '__storage_test__';
```
matthewlane: Just had a lint update merged that's causing an error here:
`Split initialized 'const' declarations into multiple statements  one-var`
Once this is fixed we should be good to merge, thanks

danclien:
@dbemiller @matthewlane, thanks for the code review and getting this merged! Do y'all publish an estimated timeline for upcoming releases?

dbemiller:
@danclien No estimated timeline published... but we're planning on doing 0.27 sometime next week.

PR #1372, opened 2017-07-12
Fixing some recent bugs in addBidResponse

dbemiller:
## Type of change
Bugfix

## Description of change
Fixing some regressions introduced in #1277 

1. The old code allowed adapters to call `addBidResponse` without supplying a bid. This permissibility may or may not be a good idea... but #1277 definitely isn't a good place to make that change. The `(!bid)` check is added back in here.

2. The `bidAdustment` event needs to be fired before the `getPriceBucketStrings` gets called, because listeners for that event may mutate the bid's cpm.

### DIFF
--- a/src/bidmanager.js
+++ b/src/bidmanager.js
@@ -129,7 +129,13 @@ exports.addBidResponse = function (adUnitCode, bid) {
   // Postprocess the bids so that all the universal properties exist, no matter which bidder they came from.
   // This should be called before addBidToAuction().
   function prepareBidForAuction() {
+    // Let listeners know that now is the time to adjust the bid, if they want to.
+    //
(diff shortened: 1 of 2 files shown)


src/bidmanager.js
```diff
@@ -108,11 +108,11 @@ exports.addBidResponse = function (adUnitCode, bid) {
     }
 
     if (!adUnitCode) {
-      utils.logWarn(errorMessage('No adUnitCode was supplied to addBidResponse.'));
```
kitwestneat: fwiw if you put the bid check before the adUnitCode, you could still use errorMessage here

mkendall07: true, these checks really shouldn't be necessary anyway once we have a proper adapter validator in place.

PR #1378, opened 2017-07-13
Adding Orbitsoft adapter

Indra-sensei:
## Type of change

- [x] New bidder adapter   

## Description of change

New Orbitsoft bidder

```
{
  bidder: 'orbitsoft',
  params: {
    placementId: '142',
    requestUrl: '//adserver.com/ads/show/hb' // Url to perform search request
  }
}
```

support@orbitsoft.com


modules/orbitsoftBidAdapter.js
```diff
@@ -0,0 +1,229 @@
...
+
+let OrbitsoftAdapter;
+OrbitsoftAdapter = function OrbitsoftAdapter() {
+  let baseAdapter = Adapter.createNew('orbitsoft');
+
+  baseAdapter.callBids = function (params) {
+    let bids = params.bids || [];
+
+    if (bids.length === 0) {
+      return;
```
snapwich: this conditional is kind of pointless.  if `bids.length === 0` then the for loop below won't iterate and the function will return anyways

modules/orbitsoftBidAdapter.js
```diff
@@ -0,0 +1,229 @@
...
+
+    let referrer = utils.getBidIdParameter('ref', bid.params);
+    let location = utils.getBidIdParameter('loc', bid.params);
+    let jptCall = utils.getBidIdParameter('requestUrl', bid.params);
+    if (jptCall.length === 0) {
+      // No param requestUrl
+      // @if NODE_ENV='debug'
+      utils.logMessage('No param requestUrl');
+      // @endif
+      return null;
```
snapwich: If there are errors creating the bid request, you should return a NO_BID to prebid

test/spec/modules/orbitsoftBidAdapter_spec.js
```diff
@@ -0,0 +1,358 @@
...
+        ]
+      };
+
+      // Empty bid response
+      let response = {
+        callback_uid: 'bidIdOrbitsoft1',
+        cpm: 0
+      };
+
+      pbjs._bidsRequested.push(bidderRequest);
```
snapwich: Shouldn't need to mess with Prebid's private vars.  This has the ability to affect other tests, plus it's a private var so it could possible it could change.

Indra-sensei: But this is done by push method, which not has affect other tests, and this is exactly what is recommended by the official document at http://prebid.org/dev-docs/testing-prebid.html
And this technique is used for most other adapter tests for prebid.js

modules/orbitsoftBidAdapter.js
Indra-sensei: Fixed (on line 47)

modules/orbitsoftBidAdapter.js
Indra-sensei: Fixed

mkendall07:
@snapwich what's the status on this?

modules/orbitsoftBidAdapter.js
```diff
@@ -0,0 +1,232 @@
...
+      utils.logMessage('No prebid response for placement');
+      // @endif
+    }
+  };
+
+  return {
+    callBids: baseAdapter.callBids,
+    setBidderCode: baseAdapter.setBidderCode,
+    buildJPTCall: buildJPTCall
+  };
```
snapwich: In order for your adapter to be alias-able the `.constructor` reference on your adapter instances needs to point to the correct constructor function; by returning a new object that reference will be incorrect.

You can solve this with:
```Javascript
return Obect.assign(this, {
     callBids: baseAdapter.callBids,
     setBidderCode: baseAdapter.setBidderCode,
     buildJPTCall: buildJPTCall
});
```

Indra-sensei: @snapwich Fixed

snapwich: Sorry, you need to modify it to be the way I described (return `this`).  You can look at another adapter if you wish to get an example, but the modifications you made look like the old alias pattern.  `createNew` is no longer needed, but the constructor functions must return a reference to `this`.

Indra-sensei: @snapwich Sorry for this. I made changes.

modules/orbitsoftBidAdapter.js
```diff
@@ -0,0 +1,232 @@
...
+
+  return Object.assign(this, {
+    callBids: baseAdapter.callBids,
+    setBidderCode: baseAdapter.setBidderCode,
+    buildJPTCall: buildJPTCall
+  });
+};
+
+adaptermanager.registerBidAdapter(new OrbitsoftAdapter(), ORBITSOFT_BIDDERCODE);
+adaptermanager.aliasBidAdapter('orbitsoft', 'orbitadserving');
```
mkendall07: why so many aliases?

Indra-sensei: We rebranded recently from Orbitscripts with Orbit Ad Serving product to Orbitsoft.
If too much, I can delete.

mkendall07: I think it is a bit confusing to have so many. Typically if the code changes we put the alias in to help with transition but since this is a new adapter I don't think you need them. You should use your current preferred name only.

modules/orbitsoftBidAdapter.js
```diff
@@ -0,0 +1,232 @@
...
+let OrbitsoftAdapter = function OrbitsoftAdapter() {
+  let baseAdapter = new Adapter(ORBITSOFT_BIDDERCODE);
+
+  baseAdapter.callBids = function(params) {
+    let bids = params.bids || [];
+
+    for (let i = 0; i < bids.length; i++) {
+      let bidRequest = bids[i];
+      let callbackId = bidRequest.bidId;
+      let jptCall = buildJPTCall(bidRequest, callbackId);
```
mkendall07: I see you are hitting the AppNexus endpoint. Is there a reason you cannot use an AppNexus alias instead of creating your own adapter?

modules/orbitsoftBidAdapter.js
Indra-sensei: @mkendall07 It is fair remark. Deleted it.

modules/orbitsoftBidAdapter.js
Indra-sensei: AppNexus has built-in enpoint URL for bid requesting. Our adapter sends bid request to various endpoint URL where customers install our product. Endpoint URl is defined by the requestUrl parameter and different for diffrent customers.
We also have additional options for requesting bids and ads.

PR #1355, opened 2017-07-07
Modernizing build dependencies

dbemiller:
## Type of change
Build upgrades & maintenance

## Description of change
Lots of build maintenance relating to webpack, istanbul, and karma.

### Fixed bugs
  - Failed tests now use the correct line numbers in the stack traces
  - Code coverage reports use pre-transpiled code now
  - Output from `gulp test —browserstack` is much more concise

### Maintenance
  - Many dependencies (notably Karma and Webpack) run the latest versions
  - CI now uses Node 7 (for Webpack 3.0 compatibility. We should replace yarn with Node 8 in a separate PR)
  - Added `gulp test-coverage` command. `gulp test` no longer generates coverage. `gulp run-tests` uses `gulp test-coverage`, while `gulp serve` uses `gulp test`.
  - PhantomJS replaced by HeadlessChrome everywhere
  - IE9 removed from the browserstack list (since we don't support it anymore)

## Performance Comparison
These were very un-scientific... I only ran each test once, and didn't pay attention to my machine load. I can run more careful trials if anyone really cares though.

### Old build (from Master)
prebid.js bundle size — 269KB

`npm install` — 31s
`yarn install` — 20.45s
`gulp build` — 13s
`gulp lint` — 6s
`gulp run-tests` — 1m1s

## New build (from this branch)
prebid.js bundle size — 264KB

`npm install` — 25s
`yarn install` — 18.59s
`gulp build` — 15s
`gulp lint` — 5s
`gulp test` — 37s
`gulp run-tests` — 1m17s (no longer used by `gulp serve`)
`gulp test --browserstack` -- About 10 minutes

PR #1386, opened 2017-07-18
Mention NodeJS 4.0 dependency in the README

BartVB:
This caused me some trouble on a Debian Jessie machine (which ships with NodeJS 0.10).

### DIFF
--- a/README.md
+++ b/README.md
@@ -24,11 +24,13 @@ Working examples can be found in [the developer docs](http://prebid.org/dev-docs
 
 ## Install
 
+Prebid requires NodeJS 4.0 or higher.
+
     $ git clone https://github.com/prebid/Prebid.js.git
     $ cd Prebid.js
     $ yarn install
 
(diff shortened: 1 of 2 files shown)


dbemiller:
Thanks for the PR. This probably happened during #1355, where I definitely overlooked node compatibility.

Could you also add an `engines` property for it? It'll make a better error message if people miss the note in the README. https://docs.npmjs.com/files/package.json#engines

BartVB:
Of course! Didn't know that option existed, thanks!

mkendall07:
There is actually a mention of node 4.0 here already at the bottom of this portion: https://github.com/prebid/Prebid.js/blob/master/README.md#Build
The engines is a nice addition though.

BartVB:
Ah, missed that one.

I moved and changed the note. 'npm install' aborts with a lot of errors on NodeJS < 4.0, it's not just needed for 'gulp'

mkendall07:
thanks!

PR #1389, opened 2017-07-18
Reject sizeless banner bids

matthewlane:
## Type of change
- Bugfix

## Description of change
In cases where banner bid responses do not have a width or height, when these bids are rendered, we seem to be setting the width and height of the adserver iframe to "undefined", messing up the page formatting. This change rejects bids that do not have a width or height, unless the impression has one valid size, in which case that size is assigned to any bid that does not explicitly set a width or height.

### DIFF
--- a/modules/getintentBidAdapter.js
+++ b/modules/getintentBidAdapter.js
@@ -55,6 +55,7 @@ var GetIntentAdapter = function GetIntentAdapter() {
             if (br.mediaType === 'video') {
               bid.vastUrl = bidResponse.vast_url;
               bid.descriptionUrl = bidResponse.vast_url;
+              bid.mediaType = 'video';
             } else {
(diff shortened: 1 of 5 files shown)


jaiminpanchal27:
@matthewlane Code looks good. But i have one question. 
There are still many adapters who are not passing bidRequest when calling `createBid` function. If that adapter is responding with banner bid its fine but if its of type video and adapter is not passing bidRequest than this can create problems.

e.g: This guy https://github.com/prebid/Prebid.js/blob/master/modules/getintentBidAdapter.js
They are not passing bidRequest and also they support video.

matthewlane:
@jaiminpanchal27 Not sure I follow, if a bid is of type video it won't be checked by the new validation function, regardless of if `bidRequest` was passed to `createBid`, unless I'm missing something?

PR #1393, opened 2017-07-19
Adomik Analytics Adapter

Yann-Pravo:
## Type of change

- [ ] Bugfix
- [ ] Feature
- [x] New bidder adapter
- [X] New analytics adapter 
- [ ] Build related changes
- [ ] CI related changes
- [ ] Other

## Description of change
New Adomik analytics adapter to integrate our script in Prebid.js

```
{
  bidder: 'adomik',
  params: {
      id: String, // id provided by Adomik
      url: String, // url provided by Adomik
      timeout: Int // optional (default 1000) in ms
  }
}
```

- contact email of the adapter’s maintainer yann@adomik.com


Yann-Pravo:
Hi, any update about a review ?

package.json
```diff
@@ -24,11 +24,13 @@
     "node": ">=4.0"
   },
   "devDependencies": {
+    "avsc": "^5.0.3",
```
dbemiller: We chatted about this internally, and after a bit of investigation, opted not to let this dependency come into the project without some really strong motivation.

Check out the size of `build/dist/prebid.js` when you run `gulp build --modules=adomikAnalyticsAdapter` vs `gulp build --modules=appnexusAnalyticsAdapter`. Even after minification, the first one is 147K, while the second is 61K. Since you're not adding very much code, this difference is mainly due to the `avsc` library.

On the average mobile phone, it takes ~2 seconds [just to _parse_](https://medium.com/reloading/javascript-start-up-performance-69200f43b201) 1MB of javascript, so we don't want to expose publishers to that type of performance penalty.

Yann-Pravo: Thank you for your answer, I'll chat with my team about it.

package.json
```diff
@@ -24,11 +24,13 @@
...
     "babel-core": "6.22.0",
     "babel-loader": "^7.1.1",
     "babel-plugin-transform-es3-member-expression-literals": "6.22.0",
     "babel-plugin-transform-es3-property-literals": "6.22.0",
     "babel-preset-es2015": "6.22.0",
+    "base-64": "^0.1.0",
```
Yann-Pravo: Alright thanks, I'll use atob then.

Yann-Pravo:
Hi @dbemiller,

Do you know a light library to compress json or string ?
Because without Avro (avsc), the URL sent is way larger than with it and might be too big for browsers with a limit on URL length.

Thank you,

Yann.

dbemiller:
Why are URL limits an issue? Can't you POST the data?

Issue #1398, opened 2017-07-20
 Missing parameters in bidAdjustment events


kir-roxot:
## Type of issue
 Missing parameters in bidAdjustment events

## Description
Event bidAdjustment even has changed in prebid v0.26.0 - it lacks AdUnitCode, timeToRespond, requestTimestamp, responseTimestamp. Also, if bidResponse is empty or has an error, bidAdjustment doesn't have a CPM parameter.

dbemiller:
Caused by some changes in #1277 and #1372. Will fix shortly.

alex-khar:
@dbemiller Our Prebid Analytics relies on this as it takes bids after bid adjustments.

GLStephen:
@dbemiller we use these parameters for PubWise analytics as well

dbemiller:
Fixed with the merge of #1399. Tenative plan is to release Prebid 0.26.1 with this fix (along with a few other things) on Monday.

PR #1399, opened 2017-07-20
Fixing the BidAdjustmentEvent fire time

dbemiller:
## Type of change
Bugfix

## Description of change
Fixing #1398.

...turns out that the timing of `BidAdjustment` is very fragile, and people are relying on in to be after some bid mutations, but before some others. This puts it (I think) where it's supposed to be, and adds some regression tests.

### DIFF
--- a/test/spec/unit/bidmanager_spec.js
+++ b/test/spec/unit/bidmanager_spec.js
@@ -70,6 +70,9 @@ describe('The Bid Manager', () => {
      */
     function prepAuction(adUnits, bidRequestTweaker) {
       function bidAdjuster(bid) {
+        if (bid.hasOwnProperty('cpm')) {
+          bid.hadCpmDuringAdjuster = true;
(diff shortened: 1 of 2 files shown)


test/spec/unit/bidmanager_spec.js
```diff
@@ -142,6 +157,21 @@ describe('The Bid Manager', () => {
           bidManager.addBidResponse('mock/code');
           expect($$PREBID_GLOBAL$$._bidsReceived.length).to.equal(0);
         });
+
+        it('should attach properties for analytics *before* the BID_ADJUSTMENT event listeners are called', () => {
+          console.log('testing this bug');
```
matthewlane: Looks like this `console.log` slipped through :snake:

dbemiller: d'oh... thanks

PR #1403, opened 2017-07-20
Updates to Build Process to Support Building DIST and DEV Together

GLStephen:
## Type of change
- [ X ] Bugfix

The build process calls the clean task before webpack and devpack. The clean function cleans out the entire build directory, not the subdirectory specific to the built variant (DEV/DIST). This change introduces a simple refactor to have each build clean their respective directory, but also retains a /build level clean option. We use the DEV version in certain places where a map file do not work, so we like to build both together.

There are other reasonable approaches. One sensible one would be to always build DEV when building DIST. Then allow DEV to be built on demand separately. This change was the simplest change I could make while retaining known behavior and allow for the behavior I needed, and that matched functionality that exists around version 23 which previously built both and supported building both together. If we need another approach, then let me know.

### DIFF
--- a/gulpfile.js
+++ b/gulpfile.js
@@ -46,11 +46,23 @@ gulp.task('run-tests', ['lint', 'test-coverage']);
 gulp.task('build', ['build-bundle-prod']);
 
 gulp.task('clean', function () {
-  return gulp.src(['build'], {
-      read: false
-    })
-    .pipe(clean());
+  clean('');
 });
+
+gulp.task('clean-dev', function () {
+  clean('/dev');
+});
+
+gulp.task('clean-dist', function () {
+  clean('/dist');
+});
+
+function clean(type) {
+  return gulp.src(['build'+type], {
+    read: false
+  })
+  .pipe(clean());
+}
 
 function bundle(dev) {
   var modules = helpers.getArgModules(),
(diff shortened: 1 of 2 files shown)


gulpfile.js
```diff
@@ -46,12 +46,24 @@ gulp.task('run-tests', ['lint', 'test-coverage']);
...
+
+gulp.task('clean-dev', function () {
+  clean('/dev');
 });
 
+gulp.task('clean-dist', function () {
+  clean('/dist');
+});
+
+function clean(type) {
```
dbemiller: I think this needs to be renamed to something else. It clobbers the import: `var clean = require('gulp-clean');`

As written, I think it's actually unused. You can tell by doing the following:

```
gulp build
gulp clean-dist
```

expected: The `build/` directory should either be empty, or not exist.
observed: There's still a `build/dist` directory with stuff in it

gulpfile.js
```diff
@@ -102,7 +114,7 @@ gulp.task('build-bundle-dev', ['devpack'], bundle.bind(null, true));
 gulp.task('build-bundle-prod', ['webpack'], bundle.bind(null, false));
 gulp.task('bundle', bundle.bind(null, false)); // used for just concatenating pre-built files with no build step
 
-gulp.task('devpack', ['clean'], function () {
+gulp.task('devpack', ['clean-dev'], function () {
```
dbemiller: I think this introduces a race condition into `gulp serve`, doesn't it?

`serve` depends on `build-bundle-dev` and `test`. `test` depends on `clean`, but `build-bundle-dev` depends on `clean-dev`.

Since both depend on `clean` currently, gulp is smart enough to manage the dependency graph and make sure `clean` finishes before `build-bundle-dev` and `test` start.

By changing it to `clean-dev`, gulp won't know... so it's possible that the `clean` from `test` wipes out the code _after_ `build-bundle-dev`has already finished, leaving you with no code to `serve`.

mkendall07:
Hey @GLStephen 
would you be able to fix those items above pointed out by Dave?

GLStephen:
@dbemiller @mkendall07 I can fix those, working on our building process in next few days so I can incorporate this feedback and submit a change to this PR.

mkendall07:
Just another ping on this. Thanks

mkendall07:
closing as stale.

PR #1406, opened 2017-07-21
Index explicit pass on bid

### LINKED ISSUE #1341: Index ignores unsupported sizes, bid manager still expects bid response
## Type of issue
Bug

## Description
`bidsBackAll` method in `bidmanager` compares the number of requests with the number of responses. As indexExchange issues a response for every size in the ad unit, bidsBackAll counts all the sizes to determine the number of requests. But if some size is not supported by indexExchange, it will be ignored inside the adapter and there won't be any response for this size. So the number of requests will always exceed the number of responses. Which means that the only condition on which the auction ends is the timeout.

Not sure what's the best way to fix that. As a temporary measure, I've modified the `bidmanager` to count only those sizes that are supported by Index Exchange. But I had to copy the list of supported sizes to `bidmanager` and it doesn't seem to be a good practice. But if it's ok, I can submit my fix as a PR.

## Steps to reproduce
1. Set up an ad unit with two sizes: [300, 250] and [300, 610]
2. Add indexExchange as the only bidder to this ad unit
3. Setup a lengthy timeout, say, 10 seconds
4. Request the bids

### Expected results
Auction ends as soon as response from indexExchange is received

### Actual results
Auction ends only after 10 seconds

## Platform details
prebid v 0.25

arvending:
## Type of change

- [x] Bugfix
- [ ] Feature
- [ ] New bidder adapter   
- [ ] Build related changes
- [ ] CI related changes
- [ ] Other

## Description of change

- Explicit pass on bid for unsupported sizes to resolve #1341
- Proper protocol when IX adapter is loaded in dynamically-generated iframes.


modules/indexExchangeBidAdapter.js
```diff
@@ -419,6 +426,9 @@ var IndexExchangeAdapter = function IndexExchangeAdapter() {
     _IndexRequestData.targetAggregate = {'open': {}, 'private': {}};
 
     if (!utils.hasValidBidRequest(bidArr[0].params, requiredParams, ADAPTER_NAME)) {
+      for (var i = 0; i < bidArr.length; i++) {
+        passOnBid(bidArr[i].placementCode);
+      }
```
snapwich: Can you explain this logic to me of passing on all bids if the first bid is invalid?

dugwood: I second you on this: `bidRequested` is called for non valid sizes.

Say I call this adapter with `[[728, 90], [1000, 90]]`, the size `1000x90` is invalid, but I'll get two calls to event `bidRequested`, when I would expect only one for the size of the adUnit that is accepted by IX.

arvending: @snapwich This is an existing behaviour from previous endpoints. It is true that we no longer need this, but this shouldn't block this pull request. We'll be fix this issue in the later releases.

arvending: @dugwood Referencing: https://github.com/prebid/Prebid.js/blob/master/src/bidmanager.js#L54
`bidmanager` is expecting the number of bids returned from Index Exchange to match with the number of sizes in the slots.

dugwood: @arvending I've tested the fix, and it works (else this was raising the timeout). But @snapwich is right as the `bidRequested` event confirms it.

Nevertheless, I'll patch my 0.26.1 release with your fix, so that I can make it work (I've warned your team a while ago about it, so I'm okay with this temporary solution).

Thanks for your patch anyway.

snapwich: @arvending the bids array contains multiple bids across different adUnits though.  Also, that bid order is pretty much arbitrary (I think it ends up being definition order or something), so why you would want to return NO_BIDs on _all_ bids across _all_ adUnits involved in an auction because an arbitrary bid invalid makes no sense.  Maybe I'm missing something with the indexExchange exception that changes things?

@mkendall07 what do you think about this?

arvending: @snapwich
Our endpoint had changed in this PR - [#826](https://github.com/prebid/Prebid.js/pull/826#). The site id and slot id information from the first slot were treated uniquely in our original endpoint.  Now that endpoint had changed, there are no needs to reject all bids if the first slot is invalid. Since this validation is an existing behaviour and introduced no behaviour changes to IX adapter, this issue should not block this particular pull request. In the later release, we will remove this first slot validation logic.

PR #1411, opened 2017-07-24
fix #1410 - issue with ie and xhr.timeout

### LINKED ISSUE #1410: Using appnexusAst adapter on IE 11 error
Hello. Every time i try to use appnexusAst on ie11 (and i see the same errors in other adapters) it return this error "InvalidStateError: xhr construction". 

In this example the errors can be seen
http://acdn.adnxs.com/prebid/alpha/unrulydemo.html 

Here are the errors with debug:
https://www.screencast.com/t/wRydDqWhKz

mkendall07:
## Type of change

- [x] Bugfix

## Description of change
Fixes #1410

### DIFF
--- a/src/ajax.js
+++ b/src/ajax.js
@@ -82,7 +82,7 @@ export function ajax(url, callback, data, options = {}) {
     }
 
     x.open(method, url);
-    // IE needs timoeut to be set after open
+    // IE needs timoeut to be set after open - see #1410
     x.timeout = _timeout;
 
     if (!useXDomainRequest) {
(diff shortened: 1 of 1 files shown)


src/ajax.js
```diff
@@ -83,6 +82,8 @@ export function ajax(url, callback, data, options = {}) {
     }
 
     x.open(method, url);
+    // IE needs timoeut to be set after open
```
dbemiller: Can we stick a #1410 in the comment somewhere?

I'd recommend this as a general process, whenever bugs get fixed... cause it's been super helpful in other work. Tickets accumulate more info than code comments, and can include things like screenshots... so they tend to me the most informative when you're later wondering why this code is the way it is.

PR #1415, opened 2017-07-25
Yieldmo bid adapter

xlozinguez:
## Type of change

- [ ] Bugfix
- [ ] Feature
- [x] New bidder adapter   
- [ ] Build related changes
- [ ] CI related changes
- [ ] Other

## Description of change

Added Yieldmo adapter

```
{
  bidder: 'yieldmo',
  params: {}
}
```

- contact email of the adapter’s maintainer: support@yieldmo.com

## Other information
@cdoher01 @harquail @yieldmo-rao @elberdev

### DIFF
--- a/modules/yieldmoBidAdapter.js
+++ b/modules/yieldmoBidAdapter.js
@@ -1,0 +1,142 @@
+var utils = require('src/utils.js');
+var adloader = require('src/adloader.js');
+var bidmanager = require('src/bidmanager.js');
+var bidfactory = require('src/bidfactory.js');
+var adaptermanager = require('src/adaptermanager');
+
+/**
+ * Adapter for requesting bids from Yieldmo.
+ *
+ * @returns {{callBids: _callBids}}
+ * @constructor
+ */
+
+var YieldmoAdapter = function YieldmoAdapter() {
+  function _callBids(params) {
+    var bids = params.bids;
+    adloader.loadScript(buildYieldmoCall(bids));
+  }
+
+  function buildYieldmoCall(bids) {
+    // build our base tag, based on if we are http or https
+    var ymURI = '//ads.yieldmo.com/ads?';
+    var ymCall = document.location.protocol + ymURI;
+
+    // Placement specific information
+    ymCall = _appendPlacementInformation(ymCall, bids);
+
+    // General impression params
+    ymCall = _appendImpressionInformation(ymCall);
+
+    // remove the trailing "&"
+    if (ymCall.lastIndexOf('&') === ymCall.length - 1) {
+      ymCall = ymCall.substring(0, ymCall.length - 1);
+    }
+
+    // @if NODE_ENV='debug'
+    // utils.logMessage('ymCall request built: ' + ymCall);
+    // @endif
+
+    return ymCall;
+  }
+
+  function _appendPlacementInformation(url, bids) {
+    var placements = [];
+    var placement;
+    var bid;
+
+    for (var i = 0; i < bids.length; i++) {
+      bid = bids[i];
+
+      placement = {};
+      placement.callback_id = bid.bidId;
+      placement.placement_id = bid.placementCode;
+      placement.sizes = bid.sizes;
+
+      placements.push(placement);
+    }
+
+    url = utils.tryAppendQueryString(url, 'p', JSON.stringify(placements));
+    return url;
+  }
+
+  function _appendImpressionInformation(url) {
+    var page_url = document.location; // page url
+    var pr = document.referrer || ''; // page's referrer
+    var dnt = (navigator.doNotTrack || false).toString(); // true if user enabled dnt (false by default)
+    var _s = document.location.protocol === 'https:' ? 1 : 0; // 1 if page is secure
+    var description = _getPageDescription();
+    var title = document.title || ''; // Value of the title from the publisher's page. 
+    var e = 4; // 0 (COP) or 4 (DFP) for now -- ad server should reject other environments (TODO: validate that it will always be the case)
+    var bust = new Date().getTime().toString(); // cache buster
+    var scrd = window.devicePixelRatio || 0; // screen pixel density
+    var ae = 0; // prebid adapter version
+
+    url = utils.tryAppendQueryString(url, 'callback', '$$PREBID_GLOBAL$$.YMCB');
+    url = utils.tryAppendQueryString(url, 'page_url', page_url);
+    url = utils.tryAppendQueryString(url, 'pr', pr);
+    url = utils.tryAppendQueryString(url, 'bust', bust);
+    url = utils.tryAppendQueryString(url, '_s', _s);
+    url = utils.tryAppendQueryString(url, 'scrd', scrd);
+    url = utils.tryAppendQueryString(url, 'dnt', dnt);
+    url = utils.tryAppendQueryString(url, 'ae', ae);
+    url = utils.tryAppendQueryString(url, 'description', description);
+    url = utils.tryAppendQueryString(url, 'title', title);
+
+    return url;
+  }
+
+  function _getPageDescription() {
+    if (document.querySelector('meta[name="description"]')) {
+      return document.querySelector('meta[name="description"]').getAttribute('content'); // Value of the description metadata from the publisher's page.  
+    } else {
+      return '';
+    }
+  }
+
+  // expose the callback to the global object:
+  $$PREBID_GLOBAL$$.YMCB = function(ymResponses) {
+    if (ymResponses) {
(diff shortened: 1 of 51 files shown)


cdoher01:
@xlozinguez

cdoher01:
Please note it's a new adapter, not a feature, misclicked :/ (I can't change it for some reason it says "You can't perform that action at this time.")

modules/yieldmoBidAdapter.js
```diff
@@ -0,0 +1,142 @@
...
+    // General impression params
+    ymCall = _appendImpressionInformation(ymCall);
+
+    // remove the trailing "&"
+    if (ymCall.lastIndexOf('&') === ymCall.length - 1) {
+      ymCall = ymCall.substring(0, ymCall.length - 1);
+    }
+
+    // @if NODE_ENV='debug'
+    // utils.logMessage('ymCall request built: ' + ymCall);
```
elberdev: Should this be commented out?

dbemiller:
Hey. This code looks good... but I'm not getting any bids during test. These are my ad units (plugged into `integrationExamples/gpt/hello_world.html`):

```
var adUnits = [{

        bids: [{
          bidder: 'yieldmo',
          params: {}
        }]

      }];
```

This also turned up a bug. Your server is returning `{}` for me, which your `YMCB` handler processes by doing nothing. Prebid then waits for your bid until the auction's timeout, which is really bad for the publisher's page's performance.

You need to make sure your adapter calls `addBidResponse` with a `NO_BID` bid regardless of what your server returns. Please add some unit tests to make sure this happens.

cdoher01:
Hey @dbemiller I've updated the bid adapter to add NOBID responses if the server returns an invalid response.  We've also got our endpoint up now so you should see a valid response if you try it again.  Thanks!

cdoher01:
@dbemiller Hey just wanted to check if there is anything else we need to do to keep this moving forward. Thanks!

modules/yieldmoBidAdapter.js
```diff
@@ -0,0 +1,153 @@
...
+      var bid = [];
+      bid = bidfactory.createBid(2, currentBid);
+      bid.bidderCode = 'yieldmo';
+      bidmanager.addBidResponse(currentBid.placementCode, bid);
+    });
+  }
+
+  return {
+    callBids: _callBids
+  };
```
dbemiller: Needs to be:

```
return Object.assign(this, {
  callBids: _callBids
});
```

See #1459

modules/yieldmoBidAdapter.js
```diff
@@ -0,0 +1,153 @@
...
+  function _getPageDescription() {
+    if (document.querySelector('meta[name="description"]')) {
+      return document.querySelector('meta[name="description"]').getAttribute('content'); // Value of the description metadata from the publisher's page.  
+    } else {
+      return '';
+    }
+  }
+
+  // expose the callback to the global object:
+  $$PREBID_GLOBAL$$.YMCB = function(ymResponses) {
```
dbemiller: This is fine for right now... but beware that Prebid will be dropping support for JSONP in version 1.0, which should be coming in the next few months.

modules/yieldmoBidAdapter.js
```diff
@@ -0,0 +1,153 @@
...
+      if (bidObj) { utils.logMessage('No prebid response from yieldmo for placementCode: ' + bidObj.placementCode); }
+      // @endif
+      bid = bidfactory.createBid(2, bidObj);
+      bid.bidderCode = 'yieldmo';
+      bidmanager.addBidResponse(placementCode, bid);
+    }
+  }
+
+  function _registerNoResponseBids() {
+    var yieldmoBidRequests = $$PREBID_GLOBAL$$._bidsRequested.find(bid => bid.bidderCode === 'yieldmo');
```
dbemiller: Same issue here... this is fine for now, but won't work in Prebid 1.0. #1421 breaks it, since different auctions will have different bid requests, so the global `_bidsRequested` array won't be valid anymore.

modules/yieldmoBidAdapter.js
```diff
@@ -0,0 +1,153 @@
...
+      bid = bidfactory.createBid(1, bidObj);
+      bid.bidderCode = 'yieldmo';
+      bid.cpm = response.cpm;
+      bid.ad = response.ad;
+      bid.width = response.width;
+      bid.height = response.height;
+      bidmanager.addBidResponse(placementCode, bid);
+    } else {
+      // no response data
+      // @if NODE_ENV='debug'
```
dbemiller: Prebid isn't using `preprocess`... so it won't respect these annotations.

The `utils.logMessage` function is a no-op unless debugging is enabled, though.

modules/yieldmoBidAdapter.js
cdoher01: Okay, I'll make a note for us to update this. Do you know if any existing adapters are a good example to look at for 1.0 compatible callback?

modules/yieldmoBidAdapter.js
cdoher01: 👍

modules/yieldmoBidAdapter.js
cdoher01: 👍

cdoher01:
@dbemiller Thanks for taking another look. I've updated to use Object.assign. Looks like the ad response is still up, you could reference this test page to see an example of the response coming back: http://s3.amazonaws.com/static.yieldmo.com/icons/beta/hello_world.html

Let me know if there is anything else to update to get this first version in. Thanks! Connor

cdoher01:
Oh, one thing to note is that we only run our ads on mobile so you will want to put the browser into mobile emulation to get back a full bid response, that might be why you weren't seeing a full response.

modules/yieldmoBidAdapter.js
dbemiller: We won't be letting servers respond with code in 1.0--just data. So you'll be making a GET/POST request, and then unpacking bids from the JSON (or XML, plain text, etc) in the response body.

The exact API for adapters in 1.0 is still under discussion in #1494, if you want to take a look and make suggestions over there. I'm updating the `appnexusAst` adapter to help prototype it.

modules/yieldmoBidAdapter.js
```diff
@@ -0,0 +1,151 @@
...
+    if (ymResponses && ymResponses.constructor === Array && ymResponses.length > 0) {
+      for (var i = 0; i < ymResponses.length; i++) {
+        _registerPlacementBid(ymResponses[i]);
+      }
+    } else {
+      // If an incorrect response is returned, register error bids for all placements
+      // to prevent Prebid waiting till timeout for response 
+      _registerNoResponseBids();
+
+      // @if NODE_ENV='debug'
```
dbemiller: Can also be removed

modules/yieldmoBidAdapter.js
```diff
@@ -0,0 +1,151 @@
...
+
+    // General impression params
+    ymCall = _appendImpressionInformation(ymCall);
+
+    // remove the trailing "&"
+    if (ymCall.lastIndexOf('&') === ymCall.length - 1) {
+      ymCall = ymCall.substring(0, ymCall.length - 1);
+    }
+
+    // @if NODE_ENV='debug'
```
dbemiller: Can also be removed

cdoher01:
Cool thanks, @dbemiller I've removed those other unneeded comments. @jaiminpanchal27 let me know if you spot anything else that needs updating before it ready to go. Thanks!

cdoher01:
@jaiminpanchal27 @dbemiller Thanks for reviewing and getting this merged in! Is there a place that I need to make a PR to have Yieldmo show up on the list of Bid Adapters here: http://prebid.org/download.html ? Thanks

dbemiller:
@cdoher01 yes. Make a PR against the docs repo: https://github.com/prebid/prebid.github.io

cdoher01:
@dbemiller cool will do, thanks!

mkendall07:
@cdoher01 
we'll actually take care of the http://prebid.org/download.html  page automatically when we update.

dbemiller:
akh... I'm sorry. For some reason I read your question as "how do I make it show up in the docs?" On [this page](http://prebid.org/dev-docs/bidders.html)

Listen to @mkendall07 ^^

cdoher01:
Okay cool. @mkendall07 do you roughly when the update will happen?

cdoher01:
@mkendall07 just following up, do you know the timeline for updating the download page?  Would be nice to have Yieldmo bid adapter available to download via option 1. Thanks

PR #1418, opened 2017-07-26
Add deprecate warning

jaiminpanchal27:
## Type of change
- [x ] Other

## Description of change
Added deprecated warning to the function which will be removed in Prebid 1.0

### DIFF
--- a/src/prebid.js
+++ b/src/prebid.js
@@ -517,7 +517,7 @@ $$PREBID_GLOBAL$$.addCallback = function (eventStr, func) {
  * @returns {String} id for callback
  *
  * @deprecated This function will be removed in Prebid 1.0
- * Please use onEvent instead.
+ * Please use offEvent instead.
  */
 $$PREBID_GLOBAL$$.removeCallback = function (/* cbId */) {
   // todo
--- a/src/prebid.js
+++ b/src/prebid.js
@@ -251,6 +251,8 @@ $$PREBID_GLOBAL$$.setTargetingForAst = function() {
  * Returns a bool if all the bids have returned or timed out
  * @alias module:$$PREBID_GLOBAL$$.allBidsAvailable
  * @return {bool} all bids available
+ *
+ * @deprecated This function will be removed in Prebid 1.0
  */
 $$PREBID_GLOBAL$$.allBidsAvailable = function () {
   utils.logInfo('Invoking $$PREBID_GLOBAL$$.allBidsAvailable', arguments);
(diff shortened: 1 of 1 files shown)


src/prebid.js
```diff
@@ -251,6 +251,8 @@ $$PREBID_GLOBAL$$.setTargetingForAst = function() {
  * Returns a bool if all the bids have returned or timed out
  * @alias module:$$PREBID_GLOBAL$$.allBidsAvailable
  * @return {bool} all bids available
+ *
+ * @deprecated This function will be removed in Prebid 1.0
  */
 $$PREBID_GLOBAL$$.allBidsAvailable = function () {
   utils.logInfo('Invoking $$PREBID_GLOBAL$$.allBidsAvailable', arguments);
```
mkendall07: you should also `logMessage` the deprecation warning

jaiminpanchal27: Done

jaiminpanchal27: Added it as `logWarn`. Not 100% sure. Let me know if it should be `logMessage`

src/prebid.js
```diff
@@ -251,8 +251,11 @@ $$PREBID_GLOBAL$$.setTargetingForAst = function() {
  * Returns a bool if all the bids have returned or timed out
  * @alias module:$$PREBID_GLOBAL$$.allBidsAvailable
  * @return {bool} all bids available
+ *
+ * @deprecated This function will be removed in Prebid 1.0
```
dbemiller: So... if we assume that a user is seeing this comment, then the experience they most likely had is:

1. Hey... what are these warnings in my console?
2. Seeks more info. Finds their way to this comment.
3. But then how will I update my code to work with 1.0?
4. *throws tantrum*

If we want to avoid (4)s, then the comments here have to provide answers to (3). I don't have a magic bullet for this... but my best advice from past experience would be:

1. If there's some simple replacement method, list it here.
2. If not, mention the #github-issue-or-PR with conversation about the deprecation. If their use-case is captured there, they'll find an answer. If not, they can bring up their use-case so that we know about it.

You're doing this for a few methods... but I would definitely recommend doing it in all of them.

aneuway2:
Without bloating the code, maybe theres a way to make these errors more prevalent when debugging = true

(alert message on screen?)

src/prebid.js
```diff
@@ -664,6 +682,7 @@ $$PREBID_GLOBAL$$.getAllWinningBids = function () {
  *   function instead. This function will be removed in Prebid 1.0.
  */
 $$PREBID_GLOBAL$$.buildMasterVideoTagFromAdserverTag = function (adserverTag, options) {
+  utils.logWarn('$$PREBID_GLOBAL$$.buildMasterVideoTagFromAdserverTag will be removed in Prebid 1.0. Include the dfpVideoSupport module in your build, and use the $$PREBID_GLOBAL$$.adserver.buildVideoAdUrl (if DFP is your only Ad Server) or $$PREBID_GLOBAL$$.adservers.dfp.buildVideoAdUrl (if you use other Ad Servers too) function instead.');
```
dbemiller: This isn't accurate. `$$PREBID_GLOBAL$$.adservers.dfp.buildVideoAdUrl` is always the function to use.

jaiminpanchal27: Updated

Issue #1087, opened 2017-03-28
Prebid 1.0 - multiple concurrent auctions.


bretg:
Tracking issue for the 'multiple concurrent auction' feature of Prebid 1.0 as defined in Issue #891

First deliverable is a spec.

mkendall07:
Engineering work is blocked by:
https://github.com/prebid/Prebid.js/issues/1086
https://github.com/prebid/Prebid.js/issues/1121

mkendall07:
Spec is here: https://gist.github.com/mkendall07/1ef1b27ed3501d7b03fbd72ec455bab4

protonate:
Some progress here on the spec, request for comment:
https://gist.github.com/protonate/e2f271873bec0c5596ff2715d8fdb97f

dbemiller:
Has there been any thought into how analytics adapters will interact with concurrent auctions yet?

There's a little section [here](https://gist.github.com/protonate/e2f271873bec0c5596ff2715d8fdb97f#reporting--analytics)... but I'm wondering things like:

How will analytics adapters separate the data from different auctions? Is this a requirement? If so, will 1.0 analytics adapters need to support multiple instances? Or are we planning to attach an `auctionId` to each event?

PR #609, opened 2016-09-08
Prebid auctions

protonate:
@mkendall07 for initial review. Adapters and tests need to be updated still.

PR #386, opened 2016-06-01
Prebid Auctions

protonate:
Prebid Auctions are implemented which enables instances of Prebid requests, responses and ad renders. This PR also includes the API changes needed to make concurrent requests for multiple placements, to queue up the next set of placements (for example, in an image carousel app) and to connect bid requests with bid responses across auction instances for greater level of analytics details.

PR #1421, opened 2017-07-26
Concurrent auctions for Prebid 1.0

jaiminpanchal27:
## Type of change
- [x ] Feature

## Description of change
Current status: Not ready to merge (Unit test in progress)
The purpose of creating this PR is to get everyone's thoughts on the work done till now.

Implemented concurrent auction for Prebid 1.0 following tech spec over here : [Kendall's gist](https://gist.github.com/mkendall07/1ef1b27ed3501d7b03fbd72ec455bab4), [Nate's gist](https://gist.github.com/protonate/530d167bd2e4d7366aa5ad5721526109)

This feature gives us the facility to run auction in parallel without changing public API. Publishers will be using the same `requestBids` function but at the core we are replacing `bidManager` with `auctionManager`.  A new auction instance will be created by auctionManager each time when `requestBids` is called. Also we are removing `pbjs._bidsRequested` and 'pbjs._bidsReceived'. Now auction instance will store both and will not be available publicly.

Removed bidManager usage from all the places in the core but not removed bidManager file because than `build-bundle-dev` task fails completely as all adapters are using bidManager

For now to start server, remove dependent task `watch and test` from `serve` task, 
i.e. `gulp.task('serve', ['lint', 'build-bundle-dev']);`

I have updated appnexus, appnexusAst and rubicon adapter with the new changes for this PR and tested with hello_world test page.

## Other information
#1087 
@prebid/core 
Taken some of the stuff from @protonate's earlier Prebid auctions attempt https://github.com/prebid/Prebid.js/pull/386 https://github.com/prebid/Prebid.js/pull/609


snapwich:
Just thought of an issue in regards to the currency module #1374 

Currently it decorates the `bidmanager.addBidResponse` so it can modify the `cpm` values of bids as they come in to handle currency conversion rates.  How would the currency module hook into this process given this new process of an auctionManager with separate instances of `addBidResponse`?

src/auctionManager.js
```diff
@@ -0,0 +1,433 @@
+const adaptermanager = require('./adaptermanager');
```
dbemiller: You've got a circular dependency here. The `auctionManager` imports the `adapterManager`, but the `adapterManager` has a `callBids(auction, cbTimeout)` method. Circular dependencies cause a bunch of problems (we can talk if you're interested)... so one of these two should not exist.

IMO `auctionManager` should depend on `adapterManager`, because you can't do an Auction without Adapters to call. An Adapter could be used outside of the context of an Auction, though.

modules/appnexusAstBidAdapter.js
```diff
@@ -36,7 +35,7 @@ function AppnexusAstAdapter() {
   let usersync = false;
 
   /* Prebid executes this function when the page asks to send out bid requests */
-  baseAdapter.callBids = function(bidRequest) {
+  baseAdapter.callBids = function(bidRequest, addBidResponse, done) {
```
dbemiller: As long as you're rewriting these... you might want to check out [the format proposed](https://gist.github.com/snapwich/72d266f510bd2f83a6eacfb2bb68112d) by @snapwich 

Other adapter maintainers are going to look to these for inspiration... so setting a good pattern will pay dividends in maintenance.

jaiminpanchal27:
@snapwich In regards to currency module, i thought of two options.
Since addBidResponse is now tied to auction instance we can do 

Option 1
import `addBidResponseDecorator` into auctionManager and when instance of auction is created  [here](https://github.com/prebid/Prebid.js/pull/1421/files#diff-748ffbc3e32da1740e8219d559aff3e4R154) i decorate the method `auction.addBidResponse`

Option 2
import `addBidResponseDecorator` into adapterManager and decorate `auction.addBidResponse` in `callBids` function.

src/auctionManager.js
```diff
@@ -0,0 +1,433 @@
...
+const utils = require('./utils');
+const events = require('./events');
+const CONSTANTS = require('./constants.json');
+import { uniques, flatten, adUnitsFilter, getBidderRequest, timestamp } from './utils';
+import { getPriceBucketString } from './cpmBucketManager';
+
+export const auctionManager = (function() {
```
jaiminpanchal27: Done

src/auctionManager.js
jaiminpanchal27: Done

src/constants.json
```diff
@@ -65,5 +65,12 @@
     "SRC" : "s2s",
     "ADAPTER" : "prebidServer",
     "SYNC_ENDPOINT" : "https://prebid.adnxs.com/pbs/v1/cookie_sync"
+  },
+  "AUCTION": {
```
jaiminpanchal27: Done

src/prebid.js
```diff
@@ -384,43 +351,30 @@ $$PREBID_GLOBAL$$.requestBids = function ({ bidsBackHandler, timeout, adUnits, a
...
-  }
+  const auction = auctionManager.createAuction();
+  auction.setAdUnits(adUnits);
+  auction.setAdUnitCodes(adUnitCodes);
 
-  adaptermanager.callBids({ adUnits, adUnitCodes, cbTimeout });
-  if ($$PREBID_GLOBAL$$._bidsRequested.length === 0) {
-    bidmanager.executeCallback();
+  if (typeof bidsBackHandler === 'function') {
+    auction.startAuctionTimer(bidsBackHandler, cbTimeout);
```
dbemiller: I may be missing something... but this looks buggy to me.

The bidsBackHandler is supposed to be called whenever all the bids are back... and I thought that the expected argument was the `bids`.

...but here, you're passing it in as an argument where the AuctionManager calls `setCallback`. Then it looks like`callBids` is sending in `this.done()` as the final "done" callback to the `adapterManager`. `this.done` eventually calls `this.executeCallback()` once it receives all the bids. `executeCallback` then calls `getCallback()`, and then `callback($$PREBID_GLOBAL$$)`.

So... it looks to me like the `bidsBackHandler` may be called with $$PREBID_GLOBAL$$ as an argument, rather than the bids themselves.

src/auctionManager.js
```diff
@@ -0,0 +1,452 @@
...
+        auctionId: this.getAuctionId(),
+        timeout: cbTimeout
+      };
+      events.emit(CONSTANTS.EVENTS.AUCTION_INIT, auctionInit);
+
+      let bidRequests = adaptermanager.makeBidRequests(this.getAdUnits(), this.getAuctionStart(), this.getAuctionId(), cbTimeout);
+      bidRequests.forEach(bidRequest => {
+        this.setBidderRequests(bidRequest);
+      });
+      let doneCb = this.done();
```
dbemiller: `this.done()` _executes_ the `done` function. `this.done` _references_ the `done` function.

This will call `done()` immediately--before `callBids` even runs.

src/auctionManager.js
```diff
@@ -0,0 +1,452 @@
...
+
+  function _getAuction(auctionId) {
+    return _auctions.find(auction => auction.getAuctionId() === auctionId);
+  }
+
+  function _addAuction(auction) {
+    _auctions.push(auction);
+  }
+
+  function Auction() {
```
dbemiller: I propose splitting `Auction` out into its own file, for a few reasons.

1. Auctions make sense as a standalone idea.
2. The conceptual dependency between the two is clearly linear, and should never change. AuctionManager can't exist without Auctions, and there's no reason why an individual Auction should ever know about an AuctionManager.
3. Two small files are much easier to maintain than one large one.

There's a (4) too... but it's a little abstract, and I don't think I can describe it in writing. Let me know if you're curious, or unconvinced by the first three and want to chat about it.

src/auction.js
```diff
@@ -0,0 +1,433 @@
...
+  var _targeting = [];
+  var _bidderRequests = [];
+  var _bidsReceived = [];
+  var _auctionStart;
+  var _auctionId = utils.getUniqueIdentifierStr();
+  var _auctionStatus;
+  var _callback;
+  var _timer;
+
+  this.setAdUnits = (adUnits) => _adUnits = adUnits;
```
dbemiller: Whenever possible, it's a good idea to minimize mutability.

For example, the only use of `setAdUnits` that I see is in `requestBids`:

```
  const auction = auctionManager.createAuction();
  auction.setAdUnits(adUnits);
  auction.setAdUnitCodes(adUnitCodes);
```

This could easily be replaced by something like:

```
const auction = auctionManager.createAuction({
  adUnits: adUnits,
  adUnitCodes: adUnitCodes
});
```

Doing so lets you delete the `set` function, which makes your code easier to reason about.

src/auction.js
```diff
@@ -0,0 +1,433 @@
...
+  this.setAuctionId = (auctionId) => _auctionId = auctionId;
+  this.setAuctionStart = (auctionStart) => _auctionStart = auctionStart;
+  this.setAuctionStatus = (auctionStatus) => _auctionStatus = auctionStatus;
+  this.setCallback = (callback) => _callback = callback;
+  this.setTimer = (timer) => _timer = timer;
+
+  this.getId = () => _id;
+  this.getAdUnits = () => _adUnits;
+  this.getAdUnitCodes = () => _adUnitCodes;
+  this.getTargeting = () => _targeting;
```
dbemiller: Some of these don't look like they're used... but I'm not sure if I'm missing something?

Best to delete unused code, though. Helps with maintenance, and gives you less stuff which needs testing.

src/auction.js
```diff
@@ -0,0 +1,433 @@
...
+      .filter(uniques)
+      .filter(bidder => this.getBidsReceived()
+        .map((bid) => {
+          return bid.bidder;
+        })
+        .filter(uniques)
+        .indexOf(bidder) < 0);
+  };
+
+  this.done = function() {
```
dbemiller: This isn't used outside of this file, is it?

If not, it's best not to attach it to the public API. It makes things easier on people who come to your file thinking "how am I supposed to use an Auction?"

The same goes for any other functions which are only for internal use.

src/auctionManager.js
```diff
@@ -0,0 +1,80 @@
+import { uniques, flatten } from './utils';
+import { createAuction, getStandardBidderSettings, getKeyValueTargetingPairs } from 'src/auction';
+
+const utils = require('./utils');
+const CONSTANTS = require('./constants.json');
+
+export function newAuctionManager() {
```
dbemiller: Nice! This looks awesome now. Very easy to read and understand.

test/spec/unit/pbjs_api_spec.js
```diff
@@ -30,18 +31,19 @@ require('modules/adequantBidAdapter');
...
   $$PREBID_GLOBAL$$.setConfig({ enableSendAllBids: false });
   $$PREBID_GLOBAL$$.clearAuction();
-  $$PREBID_GLOBAL$$._bidsRequested = getBidRequests();
-  $$PREBID_GLOBAL$$._bidsReceived = getBidResponses();
-  $$PREBID_GLOBAL$$.adUnits = getAdUnits();
-  $$PREBID_GLOBAL$$._adUnitCodes = $$PREBID_GLOBAL$$.adUnits.map(unit => unit.code);
+  auction.getBidderRequests = getBidRequests;
+  auction.getBidsReceived = getBidResponses;
+  auction.getAdUnits = getAdUnits;
+  auction.setAdUnitCodes(auction.getAdUnits().map(unit => unit.code));
```
dbemiller: Can this also do `auction = auctionManager.createAuction()` now? Cause that'd be awesome :).

If it causes test failures, though, don't worry about it. Many of these tests depend on each others' state in unhealthy ways.

test/spec/unit/pbjs_api_spec.js
```diff
@@ -503,15 +518,15 @@ describe('Unit: Prebid Module', function () {
...
-  });
+  // describe('allBidsAvailable', function () {
+  //   it('should call bidmanager.allBidsBack', function () {
+  //     var spyAllBidsBack = sinon.spy(bidmanager, 'bidsBackAll');
+  //
+  //     $$PREBID_GLOBAL$$.allBidsAvailable();
+  //     assert.ok(spyAllBidsBack.called, 'called bidmanager.allBidsBack');
+  //     bidmanager.bidsBackAll.restore();
+  //   });
+  // });
```
dbemiller: Looks like it can be deleted now.

test/spec/unit/pbjs_api_spec.js
```diff
@@ -1334,78 +1232,79 @@ describe('Unit: Prebid Module', function () {
...
+      //   bids: [{
+      //     bidder: 'appnexus',
+      //     params: { placementId: '123' }
+      //   }]
+      // }];
+      // $$PREBID_GLOBAL$$.adUnits = adUnits;
+      //
+      // const adUnitCode = '/19968336/header-bid-tag1';
+      // $$PREBID_GLOBAL$$.addBidResponse(adUnitCode, bid);
+      // assert.equal(spyClearAuction.callCount, 1, 'AUCTION_END event emitted more than once');
```
dbemiller: Is this deletable? Or meant to be uncommented again?

src/auction.js
snapwich: I feel like this API is just too... public.  It's hard to tell what is used where and I think since everything is getting exported publicly on this auction object it seems to suffer from a lack of encapsulation i.e. a lot of the logic related to auctions still seems to be scattered throughout other places in prebid.

I wonder how much this API can be simplified?  It's especially confusing because the auction manager seems to be re-exporting most of the auctions in aggregate and then other pieces of Prebid.js are then expecting to operate on those re-exported aggregates, like the targeting and such.

src/prebid.js
```diff
@@ -384,43 +351,30 @@ $$PREBID_GLOBAL$$.requestBids = function ({ bidsBackHandler, timeout, adUnits, a
...
   }
+  auction.callBids(cbTimeout);
+  setAjaxTimeout(cbTimeout);
```
snapwich: I don't think we can do this if we're going to allow multiple auctions simultaneously.

src/adaptermanager.js
```diff
@@ -55,24 +55,41 @@ function getBids({bidderCode, requestId, bidderRequestId, adUnits}) {
...
+function getAdUnitCopyForPrebidServer(adUnits) {
+  let adaptersServerSide = _s2sConfig.bidders;
+  let adUnitsCopy = utils.cloneJson(adUnits);
 
-  const auctionInit = {
-    timestamp: auctionStart,
-    requestId,
-    timeout: cbTimeout
-  };
-  events.emit(CONSTANTS.EVENTS.AUCTION_INIT, auctionInit);
```
dbemiller: What's the plan for analytics adapters?

Maybe add the `auctionId` to the events payloads?

I don't think we can just delete this outright, though...

src/auctionManager.js
```diff
@@ -0,0 +1,80 @@
...
+  _public.findBidByAdId = function(adId) {
+    return _auctions.map(auction => auction.getBidsReceived()
+      .find(bid => bid.adId === adId))[0];
+  };
+
+  _public.getStandardBidderAdServerTargeting = function() {
+    return getStandardBidderSettings(_granularity)[CONSTANTS.JSON_MAPPING.ADSERVER_TARGETING];
+  };
+
+  _public.getKeyValueTargetingPairs = function() {
```
jaiminpanchal27: Removed.

src/auctionManager.js
```diff
@@ -0,0 +1,80 @@
...
+    return getKeyValueTargetingPairs(...arguments);
+  }
+
+  function _createAuction() {
+    const auction = createAuction(_customPriceBucket, _granularity)
```
jaiminpanchal27: Moved to config api. #1496

src/auction.js
```diff
@@ -0,0 +1,433 @@
...
+const AUCTION_STARTED = 'started';
+const AUCTION_IN_PROGRESS = 'inProgress';
+const AUCTION_COMPLETED = 'completed';
+
+// register event for bid adjustment
+events.on(CONSTANTS.EVENTS.BID_ADJUSTMENT, function (bid) {
+  adjustBids(bid);
+});
+
+function Auction(customPriceBucket, granularity) {
```
jaiminpanchal27: Done. Moved to config api. #1496

src/auction.js
```diff
@@ -0,0 +1,433 @@
...
+        })
+        .filter(uniques)
+        .indexOf(bidder) < 0);
+  };
+
+  this.done = function() {
+    var count = 0;
+    var auctionObj = this;
+    return function() {
+      count++
```
jaiminpanchal27: Have updated calling done callback. Not sure this is what you expected. But discussed with @dbemiller, also thanks to him for providing new function `delayExecution`. By which we are making sure that done gets called only one time. If called multiple times nothing will happen.

src/adaptermanager.js
jaiminpanchal27: Now emitting this event from different place. Moved to auction.js. Still checking for analytics adapters for support.

src/prebid.js
```diff
@@ -334,12 +305,8 @@ $$PREBID_GLOBAL$$.removeAdUnit = function (adUnitCode) {
 };
 
 $$PREBID_GLOBAL$$.clearAuction = function() {
```
jaiminpanchal27: Removed this method. It will be deprecated in 1.0 since now the only purpose of this method remains is to call syncCookies which now i have moved when auction ends

src/prebid.js
jaiminpanchal27: Added missing argument `bids` to callback.

pm-harshad-mane:
How soon can we expect this change to be available ?

dbemiller:
@pm-harshad-mane It will likely be a few months before it's the official release.

This PR is pretty close to done... but since it changes the adapter-facing API, it'll be merging into a 1.0 branch. We'll then have to notify the adapter maintainers, and wait for them to upgrade it.

If you convince the bidders you're using to upgrade quickly, though, then you could probably build straight from the 1.0 branch. Right now the big blocker is #1494, which gives adapters an API to help ease the transition.

PR #1423, opened 2017-07-26
Update Sovrn adapter. Add test coverage. Enable deal IDs.

rachelrj:
## Type of change
[ x ] Feature
[ x ] Code style update (formatting, local variables)
[ x ] Refactoring (no functional changes, no api changes)

## Description of change
Feature: Sovrn adapter can receive and set dealid response param.
Test: Sovrn adapter additional tests.
Style: es6 formatting for tests.
Refactor: Sovrn adapter (no functional changes) remove unnecessary lines.

### DIFF
--- a/modules/sovrnBidAdapter.js
+++ b/modules/sovrnBidAdapter.js
@@ -91,67 +91,63 @@ var SovrnAdapter = function SovrnAdapter() {
 
   // expose the callback to the global object:
   $$PREBID_GLOBAL$$.sovrnResponse = function (sovrnResponseObj) {
-    // valid object?
-    if (sovrnResponseObj && sovrnResponseObj.id) {
-      // valid object w/ bid responses?
-      if (sovrnResponseObj.seatbid && sovrnResponseObj.seatbid.length !== 0 && sovrnResponseObj.seatbid[0].bid && sovrnResponseObj.seatbid[0].bid.length !== 0) {
-        var impidsWithBidBack = [];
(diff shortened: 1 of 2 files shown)


modules/sovrnBidAdapter.js
```diff
@@ -91,67 +91,63 @@ var SovrnAdapter = function SovrnAdapter() {
...
+    if (sovrnResponseObj && sovrnResponseObj.id && sovrnResponseObj.seatbid && sovrnResponseObj.seatbid.length !== 0 &&
+      sovrnResponseObj.seatbid[0].bid && sovrnResponseObj.seatbid[0].bid.length !== 0) {
+      sovrnResponseObj.seatbid[0].bid.forEach(function (sovrnBid) {
+        var responseCPM;
+        var placementCode = '';
+        var id = sovrnBid.impid;
+        var bid = {};
+
+        // try to fetch the bid request we sent Sovrn
+        var bidObj = $$PREBID_GLOBAL$$._bidsRequested.find(bidSet => bidSet.bidderCode === 'sovrn').bids
```
snapwich: These private variables like `pbjs._bidsRequested` will be going away with the implementation of concurrent auctions in #1421 

Something to keep in mind. However, I won't make it a requirement for this pull-request to not use them since it's not changing the existing functionality of the adapter (which was already using them).  But it might be worthwhile to move away from their usage going forward.

rachelrj: Thanks for letting us know. We will create a new PR to remove the private variables from our adapter and use the utils instead.

PR #1427, opened 2017-07-27
fixed #1246 and add unit test

### LINKED ISSUE #1246: Prebid does not select highest bid when more than one bid per bidder/adapter in "send all bids" mode.
## Type of Issue
Bug (Possibly feature request if working as intended)

## Description
When in "Send all Bids" mode, if two bidder IDs are configured for the same adapter against the same ad unit, the bidder specific key value is not necessarily the highest bid of the two responses.

## Steps to reproduce
1. Run demo page: https://jsfiddle.net/ry8h1hg2/
2. Monitor calls to DFP. Compare the hb_pb key with hb_pb_aol. note that sometimes, the hb_pb_aol key is lower than hb_pb even though both bids are from the AOL adapter.

### Expected results
The bid submitted by the bidder specific key (hb_pb_aol) should be the highest of all bids submitted by that adapter for that ad unit.

### Actual results
The bid submitted by the bidder specific key (hb_pb_aol) is not always the highest of all bids submitted by that adapter for that ad unit (it may be the last received bid although I am having trouble confirming this)

Here is an example of the prev_scp query parameter in the DFP ad call showing this behavior. Note that this is for a single multi-size 300x250/728x90 ad unit:

`hb_size_aol=728x90&hb_pb_aol=0.16&hb_adid_aol=3b23281593c0488&hb_bidder_aol=aol&hb_size=300x250&hb_pb=0.21&hb_adid=23fe003c94653a&hb_bidder=aol`

## Platform details
All browsers, Prebid 0.24.1

## Other information
Using multiple IDs for the same bidder for the same ad unit is often used for multi-size units.

### LINKED ISSUE #1267: Targeting issue: adserverTargeting ignored
## Type of issue
Bug

## Description
We found that our CPM function in bidderSettings.adserverTargeting was being ignored and instead the "medium" price granularity is used. The issue is that getAllTargeting() is calling both getWinningBidTargeting() and getBidLandscapeTargeting(). The first function returns an in-correct but complete set of attributes and values. The winning hb_pb is incorrect (it's based on the medium granularity price bucket), but the hb_pb_rubicon is at least correct. Then getBidLandscapeTargeting() adds a second set of hb_*_rubicon values, which ends up overwriting the first set.

## Steps to reproduce
1) Create a test page with a bidderSettings that defines a custom price bucket function. Here's the one we were using when the issue was discovered. Note that the currencyRate isn't a core part of the problem, but is part of how we found it.
```
        pbjs.bidderSettings = {
            rubicon: {
   　           bidCpmAdjustment : function(bidCpm, bid){
                  var currencyRate = 111.32;
                  return bidCpm * currencyRate;
  　　       },
            adserverTargeting: [
   　　          {
                    key: "hb_size_rubicon",
                    val: function(bidResponse) {
                        return bidResponse.size;
                    }
                },{
                    key: "hb_adid_rubicon",
                    val: function(bidResponse) {
                        return bidResponse.adId;
                    }
                },{
                    key: "hb_pb_rubicon",
                    val: function(bidResponse) {
                        var cpm = bidResponse.cpm;
                        if (cpm <= 800.00) {
                            return (Math.floor(cpm * 0.2) / 0.2).toFixed(2);
                        } else {
                            return '800.00';
                        }
                    }
                }
            ]}
        }
```
2) With debug logging on, send in a bid with a value that ends up being > 20 after the cpmAdjustment
3) enableSendAllBids()

### Expected results
Say the Rubicon's the only bidder with a bid of 1.71, it gets adjusted by bidCpmAdjustment to 190.36

1) the browser console should show 
```
12408.js:4 MESSAGE: Attempting to set key value for slot: div-gpt-ad-1494311114062-0 key: hb_bidder value: rubicon
12408.js:4 MESSAGE: Attempting to set key value for slot: div-gpt-ad-1494311114062-0 key: hb_adid value: 22dd7f3f8bf614
12408.js:4 MESSAGE: Attempting to set key value for slot: div-gpt-ad-1494311114062-0 key: hb_pb value: 190.00
12408.js:4 MESSAGE: Attempting to set key value for slot: div-gpt-ad-1494311114062-0 key: hb_size value: 300x250
12408.js:4 MESSAGE: Attempting to set key value for slot: div-gpt-ad-1494311114062-0 key: hb_size_rubicon value: 300x250
12408.js:4 MESSAGE: Attempting to set key value for slot: div-gpt-ad-1494311114062-0 key: hb_adid_rubicon value: 22dd7f3f8bf614
12408.js:4 MESSAGE: Attempting to set key value for slot: div-gpt-ad-1494311114062-0 key: hb_pb_rubicon value: 190.00
12408.js:4 MESSAGE: Attempting to set key value for slot: div-gpt-ad-1494311114062-0 key: hb_bidder_rubicon value: rubicon

```

2) What goes to the ad server should be:
```
hb_pb=190.00&hb_pb_rubicon=190.00&[...]
```
### Actual results
The bid is 1.71, it gets adjusted by bidCpmAdjustment to 190.36

1) the browser console shows:
```
12408.js:4 MESSAGE: Attempting to set key value for slot: div-gpt-ad-1494311114062-0 key: hb_bidder value: rubicon
12408.js:4 MESSAGE: Attempting to set key value for slot: div-gpt-ad-1494311114062-0 key: hb_adid value: 22dd7f3f8bf614
12408.js:4 MESSAGE: Attempting to set key value for slot: div-gpt-ad-1494311114062-0 key: hb_pb value: 20.00
12408.js:4 MESSAGE: Attempting to set key value for slot: div-gpt-ad-1494311114062-0 key: hb_size value: 300x250
12408.js:4 MESSAGE: Attempting to set key value for slot: div-gpt-ad-1494311114062-0 key: hb_size_rubicon value: 300x250
12408.js:4 MESSAGE: Attempting to set key value for slot: div-gpt-ad-1494311114062-0 key: hb_adid_rubicon value: 22dd7f3f8bf614
12408.js:4 MESSAGE: Attempting to set key value for slot: div-gpt-ad-1494311114062-0 key: hb_pb_rubicon value: 190.00
12408.js:4 MESSAGE: Attempting to set key value for slot: div-gpt-ad-1494311114062-0 key: hb_bidder_rubicon value: rubicon
12408.js:4 MESSAGE: Attempting to set key value for slot: div-gpt-ad-1494311114062-0 key: hb_adid_rubicon value: 22dd7f3f8bf614
12408.js:4 MESSAGE: Attempting to set key value for slot: div-gpt-ad-1494311114062-0 key: hb_pb_rubicon value: 20.00
12408.js:4 MESSAGE: Attempting to set key value for slot: div-gpt-ad-1494311114062-0 key: hb_size_rubicon value: 300x250
12408.js:4 MESSAGE: Attempting to set key value for slot: div-gpt-ad-1494311114062-0 key: hb_bidder_rubicon value: rubicon
```
Note two problems:
- hb_pb is 20 rather than 190
- there are two sets of hb_*_rubicon targeting attributes 

2) What goes to the ad server is:
```
hb_pb=20.00&hb_pb_rubicon=20.00&[...]
```
## Platform details
Prebid 0.24.1 (and possibly earlier)

## Other information
The workaround is to not use a CPM function in adserverTargeting, and instead use a customPriceGranularity. This is fine for most uses, but isn't a generic solution.

mkendall07:
- [x] Bugfix

## Description of change
Fixes #1246 

edit: This fixes #1246 not 1267. 1267 is related but not resolved in this PR. 

## Other information
@prebid/core 
Cc: @harpere @bretg


test/spec/core/targeting_spec.js
```diff
@@ -0,0 +1,81 @@
+import { expect } from 'chai';
```
dbemiller: Regarding the file location... could we make an effort to group the unit tests into one folder? In this case, `tests/spec/core` => `tests/spec/unit/core`

In the current organization, I'd expect to find integration tests (for example, connecting `videoCache` to `prebid-cache`) right next to unit tests like this one. This is bad for automated testing, because we'll want different subsets of [unit tests, integration tests, nightwatch/selenium tests] to run at different times during the dev, PR, and Release processes.

src/targeting.js
```diff
@@ -173,19 +167,25 @@ function getAlwaysUseBidTargeting(adUnitCodes) {
...
+    let bidsByBidder = groupBy(buckets[bucketKey], 'bidderCode');
+    Object.keys(bidsByBidder).forEach(key => bids.push(bidsByBidder[key].reduce(getHighestCpm, getEmptyBid())));
+  });
+  // populate targeting keys for the remaining bids
+  return bids.map(bid => {
+    if (bid.adserverTargeting) {
+      return {
+        [bid.adUnitCode]: getTargetingMap(bid, standardKeys.filter(
+          key => typeof bid.adserverTargeting[key] !== 'undefined') // mainly for possibly
+          // unset hb_deal
```
dbemiller: Can we clear out the stale comments? I have no idea what these two mean.

src/utils.js
```diff
@@ -666,3 +666,18 @@ export function getBidderRequest(bidder, adUnitCode) {
...
+
+/**
+ *
+ * https://stackoverflow.com/a/34890276/428704
+ * @export
+ * @param {array} xs
+ * @param {string} key
+ * @returns {${key_value}: ${groupByArray}, key_value: {groupByArray}}
+ */
+export function groupBy(xs, key) {
```
mkendall07: sure, but I don't understand how this is useful to me?

src/targeting.js
mkendall07: Yeah, although it's not stale. It's saying that `hb_deal` isn't always defined even though it's in the default key set. I think it should be obvious from the code though.

src/utils.js
dbemiller: lol, it's not. Sorry... should have been clear: I didn't actually expect a change here. More of a fun FYI.

Many general-purpose utils libraries have this function, including [lodash](https://lodash.com/docs/4.17.4#groupBy), [async](https://caolan.github.io/async/groupBy.js.html), [the scala stdlib](http://www.scala-lang.org/api/2.12.1/scala/collection/immutable/List.html#groupBy[K](f:A=>K):scala.collection.immutable.Map[K,Repr]), etc. I was just pointing out how close this comes to "the bleeding edge" of general-purpose code.

mkendall07:
@dbemiller ready for re-review

dbemiller:
That... sucks :/.

Usually that has something to do with conflicting preconditions. This actually happened to me when I added `spec/unit/bidmanager_spec.js` too--except that it was a test in `pbjs_api_spec` that magically started failing.

I fixed it by making sure that the `pbjs_api_spec` tests set all the prebid global state which they depended on inside their `beforeEach` block.

My guess would be that your code relies on state which your tests don't initialize, and one of the other tests in the other files doesn't clean up after itself. In `targeting.js` I see the `pbTargetingKeys` array, `$$PREBID_GLOBAL$$._adUnitCodes`, and `$$PREBID_GLOBAL$$.adUnits`... so maybe one of those?

PR #1433, opened 2017-07-28
Bugfix: getBidRequest was missing utils ref

daniel-hoffmann:
## Type of change

- [ x] Bugfix

## Description of change

getBidRequest was missing utils reference


modules/xhbBidAdapter.js
```diff
@@ -16,7 +16,7 @@ const XhbAdapter = function XhbAdapter() {
       {
         key: 'hb_xhb_deal',
         val: function (bidResponse) {
-          return bidResponse.adId;
+          return bidResponse.dealId;
```
mkendall07: @daniel-hoffmann is this intentional? `dealId` isn't always populated but `adId` is

daniel-hoffmann: @mkendall07 If an ad is returned, the adapter sets the deal ID to a hardcoded value (see line 133), so I assume this should be fine.

mkendall07: Ah right.

Issue #1206, opened 2017-05-17
Fix ESLint Exceptions


dbemiller:
## Type of issue
Code quality improvement

## Description
This is a tracking issue for code quality improvements. The `.eslintrc` files include many exceptions which are not actually good ideas (see the comments in those files).

Anyone who has time is encouraged to fix code which violates them, and, if possible, remove those exceptions. Everything which eslint could `--fix` automatically has already been taken care of. I recommend removing at most one quality issue per PR, to keep the diffs manageable.

PR #1434, opened 2017-07-28
Removed the brace-style ESLint exclusion

dbemiller:
## Type of change

Code style update (formatting, local variables)

## Description of change
Fixing one of the linter exceptions for #1206.

Everything was done with `lint --fix` except for the following three spots, which I fixed manually:

```
dbemiller-mac:Prebid.js dbemiller$ ./node_modules/eslint/bin/eslint.js src modules/ --fix

/Users/dbemiller/code/Prebid.js/src/adloader.js
  45:3  error  Closing curly brace does not appear on the same line as the subsequent block  brace-style

/Users/dbemiller/code/Prebid.js/src/bidmanager.js
  230:3  error  Closing curly brace does not appear on the same line as the subsequent block  brace-style

/Users/dbemiller/code/Prebid.js/modules/rhythmoneBidAdapter.js
  124:63  error  Closing curly brace does not appear on the same line as the subsequent block  brace-style
```


modules/rhythmoneBidAdapter.js
```diff
@@ -34,10 +31,7 @@ function RhythmoneAdapter (bidManager, global, loader) {
 
   function load(bidParams, url, callback) {
     loader(url, function(responseText, response) {
-      if (response.status === 200)
-      { callback(200, 'success', response.responseText); }
-      else
-      { callback(-1, 'http error ' + response.status, response.responseText); }
+      if (response.status === 200) { callback(200, 'success', response.responseText); } else { callback(-1, 'http error ' + response.status, response.responseText); }
```
mkendall07: This is harder to read than before. It doesn't look like `standardjs` implements a `max-len` but I think we should consider it. I guess it doesn't do that because it can't `--fix` that automatically?

dbemiller: I'm not sure why it doesn't do that... but I agree. `max-len` is definitely an eslint rule we can add if we want, though.

For now I'll split it up manually.

PR #1438, opened 2017-07-28
OpenX Adapter Update:

lntho:
1) Implementing AJAX calls for ad requests
2) Adding prebid trasaction id to ad request
3) Updated timeout to use the timeout passed as params
4) Implemented cache busting on ad request
5) Add tbd feature field to bidResponse

## Type of change

- [x] Bugfix
- [x] Feature

## Description of change
1) Implementing AJAX calls for ad requests
2) Adding prebid trasaction id to ad request
3) Updated timeout to use the timeout passed as params
4) Implemented cache busting on ad request
5) Add tbd feature field to bidResponse

### DIFF
--- a/modules/openxBidAdapter.js
+++ b/modules/openxBidAdapter.js
@@ -1,7 +1,7 @@
 import { config } from 'src/config';
 const bidfactory = require('src/bidfactory.js');
 const bidmanager = require('src/bidmanager.js');
-const adloader = require('src/adloader');
+const ajax = require('src/ajax');
 const CONSTANTS = require('src/constants.json');
 const utils = require('src/utils.js');
 const adaptermanager = require('src/adaptermanager');
@@ -9,11 +9,21 @@ const adaptermanager = require('src/adaptermanager');
 const OpenxAdapter = function OpenxAdapter() {
   const BIDDER_CODE = 'openx';
   const BIDDER_CONFIG = 'hb_pb';
+  const BIDDER_VERSION = '1.0.1';
   let startTime;
+  let timeout = config.getConfig('bidderTimeout');
 
   let pdNode = null;
 
-  $$PREBID_GLOBAL$$.oxARJResponse = function (oxResponseObj) {
+  function oxARJResponse (oxResponseObj) {
+    try {
+      oxResponseObj = JSON.parse(oxResponseObj);
+    } catch (_) {
+      // Could not parse response, changing to an empty response instead
+      oxResponseObj = {
+        ads: {}
+      };
+    }
     let adUnits = oxResponseObj.ads.ad;
     if (oxResponseObj.ads && oxResponseObj.ads.pixels) {
       makePDCall(oxResponseObj.ads.pixels);
@@ -39,11 +49,10 @@ const OpenxAdapter = function OpenxAdapter() {
 
       let beaconParams = {
         bd: +(new Date()) - startTime,
-        br: '0', // maybe 0, t, or p
-        bt: $$PREBID_GLOBAL$$.cbTimeout || config.getConfig('bidderTimeout'), // For the timeout per bid request
+        br: '0', // may be 0, t, or p
+        bt: Math.min(timeout, window.PREBID_TIMEOUT || config.getConfig('bidderTimeout')),
         bs: window.location.hostname
       };
-
       // no fill :(
       if (!auid || !adUnit.pub_rev) {
         addBidResponse(null, bid);
@@ -125,6 +134,9 @@ const OpenxAdapter = function OpenxAdapter() {
         bidResponse.width = creative.width;
         bidResponse.height = creative.height;
       }
+      if (adUnit.tbd) {
+        bidResponse.tbd = adUnit.tbd;
+      }
     }
     bidmanager.addBidResponse(bid.placementCode, bidResponse);
   }
@@ -176,6 +188,7 @@ const OpenxAdapter = function OpenxAdapter() {
     }
 
     params.auid = utils._map(bids, bid => bid.params.unit).join('%2C');
+    params.dddid = utils._map(bids, bid => bid.transactionId).join('%2C');
     params.aus = utils._map(bids, bid => {
       return utils.parseSizesInput(bid.sizes).join(',');
     }).join('|');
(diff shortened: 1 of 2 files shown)


modules/openxBidAdapter.js
```diff
@@ -187,10 +203,25 @@ const OpenxAdapter = function OpenxAdapter() {
       }
     });
 
-    params.callback = 'window.$$PREBID_GLOBAL$$.oxARJResponse';
-    let queryString = buildQueryStringFromParams(params);
+    if (window.XMLHttpRequest && 'withCredentials' in new XMLHttpRequest()) {
```
snapwich: `withCredentials` should be available in any browser version that Prebid.js supports.  Are you seeing otherwise?

modules/openxBidAdapter.js
```diff
@@ -187,10 +203,25 @@ const OpenxAdapter = function OpenxAdapter() {
...
+          withCredentials: true
+        });
+      } catch (err) {
+        useJsonp = true;
+        utils.logMessage(`Ajax call failed due to ${err}, sending ARJ via jsonp instead.`);
+      }
+    }
 
-    adloader.loadScript(`//${delDomain}/w/1.0/arj?${queryString}`);
+    if (useJsonp) {
```
snapwich: It would be ideal to just remove Jsonp altogether if the ajax is working correctly.

modules/openxBidAdapter.js
```diff
@@ -1,18 +1,30 @@
...
 const CONSTANTS = require('src/constants.json');
 const utils = require('src/utils.js');
 const adaptermanager = require('src/adaptermanager');
 
 const OpenxAdapter = function OpenxAdapter() {
   const BIDDER_CODE = 'openx';
   const BIDDER_CONFIG = 'hb_pb';
+  const BIDDER_VERSION = '1.0.1';
   let startTime;
+  let timeout = $$PREBID_GLOBAL$$.bidderTimeout;
```
lntho: We are trying capture the timeout at the time of request and store it until time of response. Do you know if this global stays constant throughout? We have observed that sometimes it seems the value changes so that's why I'm doing this to make sure.

modules/openxBidAdapter.js
lntho: Unfortunately we require the credentials for our ajax support to function correctly, so we must do this check here

modules/openxBidAdapter.js
lntho: As mentioned in the previous comment, there are times where we cannot support ajax so we can't really remove jsonp completely. What did you mean by the switching to a scoped callback?

modules/openxBidAdapter.js
snapwich: I didn't notice that you're overwriting the timeout below from the bidderRequest params.  That should be accurate. :+1:

modules/openxBidAdapter.js
snapwich: When can you not support AJAX?  What I'm trying to say, is in any browser that doesn't support AJAX won't support Prebid.js, so if you can run Prebid.js then you should be able to use AJAX, including `withCredentials` always.

lntho: Hello! We have some special cases to support IE9 which doesn't fully support Ajax yet, so it is best to check withCredentials and have a fallback to jsonp.

modules/openxBidAdapter.js
DubKF: Regarding the timeout, there is at least one case which is not captured. If the old style PREBID_TIMEOUT variable is used in the style 

`    var PREBID_TIMEOUT = 700;`
`    // Load GPT when timeout is reached.`
`    setTimeout(initAdserver, PREBID_TIMEOUT);`

like it's done in the [example](https://github.com/prebid/Prebid.js/blob/master/integrationExamples/gpt/pbjs_example_gpt.html)  the bidmanager will have no clue about it and default to 3000 instead. 

Therefore, I'd recommend 
`bt: Math.min($$PREBID_GLOBAL$$.cbTimeout || $$PREBID_GLOBAL$$.bidderTimeout, window.PREBID_TIMEOUT || 0)` 
instead

lntho: I agree with this to try to capture the PREBID_TIMEOUT on the window but we've observed the inside cbTimeout being changed at the time of response, hence the change to save the one coming from the paramter for later use. I'll make the changes accordingly to try to capture the one on the window.

modules/openxBidAdapter.js
lntho: @snapwich Hello there! I just noticed that prebid supports only IE10+ now, just wanted to confirm with you on that. If that is the case then we're okay with removing jsonp completely and I'll get to working on that.

lntho:
@snapwich after talking with the team, I removed the JSONP requests completely. Please let me know if there are more changes needed. Thanks.

modules/openxBidAdapter.js
```diff
@@ -188,10 +201,15 @@ const OpenxAdapter = function OpenxAdapter() {
...
     });
 
-    params.callback = 'window.$$PREBID_GLOBAL$$.oxARJResponse';
-    let queryString = buildQueryStringFromParams(params);
-
-    adloader.loadScript(`//${delDomain}/w/1.0/arj?${queryString}`);
+    try {
+      let queryString = buildQueryStringFromParams(params);
+      let url = `//${delDomain}/w/1.0/arj?${queryString}`;
+      ajax.ajax(url, $$PREBID_GLOBAL$$.oxARJResponse, void (0), {
```
dbemiller: Now that JSONP is gone, is there any reason for this function to be on the global?

If not, let's remove it.

test/spec/modules/openxBidAdapter_spec.js
```diff
@@ -136,86 +136,15 @@ describe('openx adapter tests', function () {
...
-
-  it('should not call loadscript when inputting with empty params', function () {
-    let spyLoadScript = sinon.spy(adloader, 'loadScript');
+  it('should not call ajax when inputting with empty params', function () {
+    let spyAjax = sinon.spy(ajax, 'ajax');
     adapter().callBids({});
-    assert(!spyLoadScript.called);
-    spyLoadScript.restore();
+    assert(!spyAjax.called);
+    spyAjax.restore();
```
dbemiller: spies, stubs, etc need to be done in `beforeEach` or `afterEach` blocks.

Otherwise `spyAjax.restore()` won't execute when your test fails, and it causes tons of unrelated cascading failures (sinon won't let you wrap an already-wrapped function)

lntho:
@dbemiller I have removed the callback from the global, as well as changed the unit test structure to incorporate beforeEach and afterEach. Please let me know if there are any other things you see.

PR #1444, opened 2017-07-31
Removed the no-redeclare ESLint exclusion

dbemiller:
## Type of change
Code style update (formatting, local variables)

## Description of change
Removing the no-redeclare ESLint exclusion for #1206


modules/indexExchangeBidAdapter.js
```diff
@@ -426,7 +427,7 @@ var IndexExchangeAdapter = function IndexExchangeAdapter() {
     _IndexRequestData.targetAggregate = {'open': {}, 'private': {}};
 
     if (!utils.hasValidBidRequest(bidArr[0].params, requiredParams, ADAPTER_NAME)) {
-      for (var i = 0; i < bidArr.length; i++) {
```
matthewlane: Any thoughts on changing the `var` to `let` (in this loop and the one below)? `i` gets scoped to the block in that case and then no need for the `var i` declaration on 419

dbemiller: much better idea ^^

PR #1453, opened 2017-07-31
Improving events & analytics adapter testability

dbemiller:
## Type of change
Refactoring (no functional changes, no api changes)

## Description of change
This is refactor is designed to show three things in a proof-of-concept:

1. How a prebid-core utility with global state can be structured for global or local use without repeating lots of code.
2. How modules can be structured to be agnostic about whether the state is global or local.
3. How today's prebid-core can support this structure without API changes.
4. How we can make local versions of the utilities to get side-effect-free unit tests.
5. That we don't have to redesign the entire Prebid architecture.

For (1), see `events.js`.
For (2), see any of the AnalyticsAdapters.
For (3), see `analyticsAdapterRegistry`.
For (4), see any of the AnalyticsAdapter unit tests.

## Still TODO
If you think this structure is an improvement, I will add unit tests for `events` and the `analyticsAdapterRegistry`. Didn't want to spend the time if something was wrong with the core idea, though.

### DIFF
--- a/modules/appnexusAnalyticsAdapter.js
+++ b/modules/appnexusAnalyticsAdapter.js
@@ -11,7 +11,7 @@ const appnexusAdapterFactory = adapter({
   analyticsType: 'bundle'
 });
 
-analyticsRegistry.registerInjectableAnalyticsAdapter({
+analyticsRegistry.registerAnalyticsAdapterFactory({
   factory: appnexusAdapterFactory,
   code: 'appnexus'
 });
--- a/modules/appnexusAnalyticsAdapter.js
+++ b/modules/appnexusAnalyticsAdapter.js
@@ -3,17 +3,17 @@
  */
 
 import adapter from 'src/AnalyticsAdapter';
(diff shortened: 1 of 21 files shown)


protonate:
I'm in favor of using factory patterns to create instances, here are some Eric Elliot posts that may be useful to the discussion:

Factory vs Constructor vs Class
https://medium.com/javascript-scene/javascript-factory-functions-vs-constructor-functions-vs-classes-2f22ceddf33e

Factory Functions in ES6+
https://medium.com/javascript-scene/javascript-factory-functions-with-es6-4d224591a8b1

dbemiller:
closing this because it's way out of date, and the main goal was to show how global/mutable state can be pushed out into the API file, so that it can be removed completely in a future major-version Prebid update.

Since this was a proof of concept, it was rather haphazardly done... but through in-person discussions, there was enough buy-in to work towards this goal overall. The plan is written up in #1510.

PR #1457, opened 2017-08-01
Add user sync to process for approving adapter PRs

grevory:
## Type of change
- [x ] Other

## Description of change
Added a line to describe what to do when an adapter has user syncs in their PR

## Other information
Related to #1229


dbemiller:
This looks good... but we should hold off on merging until #1229 is actually in.

Issue #1335, opened 2017-06-29
Prebid 1.0: Default to random bidder call order


aneuway2:
## Type of issue
Enhancement

## Description
By default currently, bidders are called asynchronously in the order that they are defined. ~Due to browser concurrency however [only ~6 async connections can be open at one time](https://stackoverflow.com/questions/985431/max-parallel-http-connections-in-a-browser).~

Today in order to make the call order random `pbjs.setBidderSequence('random');` needs to be called before the auction starts

## Enhancement Request
Make the random call order the default behavior in the auction as that is aligned with the spirt of a fair auction within prebid.

Optionally, `pbjs.setBidderSequence` (or equivalent) can be replaced with a specific call order if a publisher desires that to be how the bidders are called

snapwich:
> Due to browser concurrency however only ~6 async connections can be open at one time.

Only 6 connections _per server_.  Since the bidder endpoints are usually different servers, we rarely see an issue here.

dbemiller:
I do wonder, though... this suggestion may have some merit, even if the technical reason isn't right?

[This article](https://www.exchangewire.com/blog/2017/06/20/hindsight-2020-examining-unintended-consequences-containers/) got passed around yesterday. The important bit:

```
Review the default settings and make a manual change if needed. All client-side containers come with default settings that favour the container’s own demand by calling them first. For example, the default Prebid setting is to call demand partners in alphabetical order, giving some companies a clear advantage.
```

I know the article is actually wrong about Prebid being alphabetical too... but the fixed order _does_ carry a little bit of bias, and the article shows that people are at least worried about it.

Is there any real harm done if we switched the default to random?

GLStephen:
@dbemiller to clarify, what is the order of bidding? I've heard it described as the "order bidders" are configured. Is that the adapters.json order? We use random, but I'd like to understand definitively the non-random option.

mkendall07:
The order is inherited from the 1st adUnit `bids` configuration.

aneuway2:
Noted @snapwich and redacted that note on concurrent connections. The inherent bias conversation has been coming up a bit lately so still has merit imo

itayo155:
In a series of experiments conducted at Oct 2016, we've found the bidder order did have measurable effect on bid-rate - especially on heavy pages and with lots of bidders.
In these experiments we've seen differences in actual bidder timeout (i.e. time between bidder start and auction end) upwards of 100ms. 
[...]

PR #1458, opened 2017-08-01
Randomize the default order we use to call bidders by default.

dbemiller:
## Type of change
Feature

## Description of change
Randomizing the default order we use to call bidders, per #1335

This is an API breaking change... but IMO it's not really worth waiting 'til 1.0 for. It matches the project's values much better, and support for it seems pretty universal.


src/constants.json
```diff
@@ -38,7 +38,8 @@
     "bidWon": "adUnitCode"
   },
   "ORDER": {
-    "RANDOM": "random"
+    "fixed": "fixed",
```
mkendall07: now that we have language `const` support the value of a `constants` file feels diminished. Probably better to define these constant strings closer to where they are used in the code.

dbemiller: that..... makes so much more sense. Will fix shortly.

mkendall07:
need approval from @mjacobsonny and @bretg  before merge.

mjacobsonny:
LGTM.  I'm fine with this as long as users have the ability to configure fixed bidder order to replicate current state.

dbemiller:
Yeah. You get the current behavior if you call `$$PREBID_GLOBAL$$.setBidderSequence('fixed')`

bretg:
Approved, though please highlight this change in the release notes so Pubs know to add setBidderSequence('fixed') as desired.

dbemiller:
Will do. Thanks ^^

Issue #1437, opened 2017-07-28
changing globalVarName breaks tests, adapter


dantarion:
## Type of issue
Bug
## Description
Changing `globalVarName` in `package.json` is supposed to change the `pbjs` global, but doing so breaks some tests, and an adapter

`pbjs` hardcoded references:

**/modules/**
```
./bidfluenceBidAdapter.js - Used as a variable name, ok
./conversantBidAdapter.js - Used in a comment, ok
./eplanningBidAdapter.js - Used in query string and comment, ok
./express.js - Used in a comment, ok
./gumgumBidAdapter.js - Not sure, tests don't fail. ok?
./indexExchangeBidAdapter.js - Used in query string, ok
./jcmBidAdapter.js - Hardcoded into callback initialization (Breaks adapter and test)
./rubiconBidAdapter.js - Used in string, ok
./smartadserverBidAdapter.js - Used in string, ok
```
**/test/spec/modules/**
I'm not as familiar with the testing system so I didn't go through these.
```
./adbladeBidAdapter_spec.js
./adbutlerBidAdapter_spec.js
./adsupplyBidAdapter_spec.js
./eplanningBidAdapter_spec.js
./kargoBidAdapter_spec.js
./mobfoxBidAdapter_spec.js
./piximediaBidAdapter_spec.js
./prebidServerBidAdapter_spec.js
./rhythmoneBidAdapter_spec.js
./rubiconBidAdapter_spec.js
./smartadserverBidAdapter_spec.js
./tapsenseBidAdapter_spec.js
./trionBidAdapter_spec.js
./widespaceBidAdapter_spec.js
./yieldbotBidAdapter_spec.js
```

## Steps to reproduce
1. Checkout master branch
2. `gulp serve`, wait for tests to pass
3. Edit `package.json` change `globalVarName` to `pbjs2`
4. `gulp serve`, wait for tests to pass
### Expected results
All tests pass both times
### Actual results
Tests fail when `globalVarName` isn't `pbjs`
## Other information
I did a rudimentary find/replace with $$PREBID_GLOBAL$$ and I was able to get the failed tests to pass, but that broke several others that were using `pbjs` string in ways unrelated to globalVarName

PR #1462, opened 2017-08-02
Respect prebid global

### LINKED ISSUE #1340: $PREBID_GLOBAL$$ not respected in some Adapters
## Type of issue
Bug

## Description
The following Adapters are not respecting the global naming convention, but they hardcode the `pbjs` variable instead.

- [eplanningBidAdapter.js](https://github.com/prebid/Prebid.js/blob/master/modules/eplanningBidAdapter.js)
- [jcmBidAdapter.js](https://github.com/prebid/Prebid.js/blame/master/modules/jcmBidAdapter.js)  

## Steps to reproduce
- Open the files linked before [eplanningBidAdapter.js:241](https://github.com/prebid/Prebid.js/blob/master/modules/eplanningBidAdapter.js#L241) & [jcmBidAdapter.js:8](https://github.com/prebid/Prebid.js/blob/master/modules/jcmBidAdapter.js#L8)

### Expected results

I expect to don't have any `pbjs` hardcoded value into the final bundle

### Actual results
I can see the variable, this will make the adapter fails if the global name is different from the defualt, plus it will eventually fail on avoid conflicts for multiple loads of prebid.js (3rd party)

## Platform details
All

## Other information
None

dbemiller:
## Type of change
Bugfix

## Description of change
I had planned on fixing the JCM adapter for #1340, thinking it'd be easy. I was wrong; their server's response assumes `pbjs`, so I can't change it in this project.

But... something slightly useful did come from it. When testing their adapter, I did `gulp serve` after changing the `globalVarName` in my `package.json`.

That surfaced many unit test failures, because the test files themselves weren't respecting `$$PREBID_GLOBAL$$`.

So... we'll have to wait for @JCarterGw to fix #1340, but we still may as well fix our test code.


dantarion:
This will also fix #1437 for me.

test/spec/modules/jcmBidAdapter_spec.js
```diff
@@ -198,7 +198,7 @@ describe('jcm adapter tests', function () {
     var response = '{"bids":[{"width":300,"cpm":3,"ad":"%3Cimg+src%3D%22http%3A%2F%2Fmedia.adfrontiers.com%2Fimgs%2Fpartnership_300x250.png%22%3E","callbackId":"3c9408cdbf2f68","height":250},{"width":728,"cpm":0,"ad":"%3Cimg+src%3D%22http%3A%2F%2Fmedia.adfrontiers.com%2Fimgs%2Fpartnership_728x90.png%22%3E","callbackId":"3c9408cdbf2f69","height":90}]}';
 
     it('callback function should exist', function () {
-      expect($$PREBID_GLOBAL$$.processJCMResponse).to.exist.and.to.be.a('function');
```
mkendall07: Why the reverse switch here? @dbemiller

dbemiller: Due to [this issue](https://github.com/prebid/Prebid.js/issues/1340#issuecomment-319710240) from the discussion in #1340.

Since their server hardcodes `pbjs` into its JSONP response, we can't change [this line](https://github.com/prebid/Prebid.js/blob/master/modules/jcmBidAdapter.js#L9), without breaking their adapter.

I couldn't change the source, so... I changed the test instead.

mkendall07:
Respect 💯

PR #1469, opened 2017-08-03
[Criteo Adapter] Use utils lib to parse request sizes

Swiiip:
## Type of change
- [x ] Bugfix

## Description of change
This PR formats the request sizes in Criteo adapter. As those sizes are not formatted, we need to format them before using them in a consistent manner.

- contact email of the adapter’s maintainer: i.caroulle@criteo.com, h.duthil@criteo.com, n.faure@criteo.com

### DIFF
--- a/modules/criteoBidAdapter.js
+++ b/modules/criteoBidAdapter.js
@@ -43,15 +43,18 @@ var CriteoAdapter = function CriteoAdapter() {
       // build slots before sending one multi-slots bid request
       for (var i = 0; i < bids.length; i++) {
         var bid = bids[i];
-        var sizes = bid.sizes || [];
+        var sizes = utils.parseSizesInput(bid.sizes);
         slots.push(
           new Criteo.PubTag.DirectBidding.DirectBiddingSlot(
             bid.placementCode,
             bid.params.zoneId,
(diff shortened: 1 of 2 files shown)


modules/criteoBidAdapter.js
```diff
@@ -43,15 +43,18 @@ var CriteoAdapter = function CriteoAdapter() {
...
             bid.params.zoneId,
             bid.params.nativeCallback ? bid.params.nativeCallback : undefined,
             bid.transactionId,
-            sizes.map((size) => {
-              return { width: size[0], height: size[1] }
+            sizes.map((sizeString) => {
+              var xIndex = sizeString.indexOf('x');
+              var w = parseInt(sizeString.substring(0, xIndex));
+              var h = parseInt(sizeString.substring(xIndex + 1, sizeString.length))
+              return new Criteo.PubTag.DirectBidding.Size(w, h);
```
muuki88: What effect has this change? We are currently sending all our ad slot sizes,
but were told that the backend doesn't use them anyway. Will this change
in the future?

Swiiip: Yes, this will change in the future, we will need this size information to handle the request on our end.

PR #1476, opened 2017-08-09
Made prebid-cache configurable, defaulting to false

dbemiller:
This makes the prebid-cache use configurable... defaulting to false, for now.

### DIFF
--- a/src/bidmanager.js
+++ b/src/bidmanager.js
@@ -3,6 +3,7 @@ import {getPriceBucketString} from './cpmBucketManager';
 import {NATIVE_KEYS, nativeBidIsValid} from './native';
 import { store } from './videoCache';
 import { Renderer } from 'src/Renderer';
+import { config } from 'src/config';
 
 var CONSTANTS = require('./constants.json');
 var AUCTION_END = CONSTANTS.EVENTS.AUCTION_END;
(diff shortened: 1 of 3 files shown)


src/bidmanager.js
```diff
@@ -197,15 +198,20 @@ exports.addBidResponse = function (adUnitCode, bid) {
...
+        if (error) {
+          utils.logWarn(`Failed to save to the video cache: ${error}. Video bid must be discarded.`);
+        } else {
+          bid.videoCacheKey = cacheIds[0].uuid;
+          addBidToAuction(bid);
+        }
+        doCallbacksIfNeeded();
+      });
+    } else {
+      addBidToAuction(bid);
```
protonate: Do we want to add video bid to auction if prebid cache is not used?

dbemiller: Pretty sure, yes. Before the call to prebid-cache was added, video bids were still getting added to the auction. If the cache is disabled, we should still replicate that behavior.

Basically: If someone's _not_ trying to use the cache, then they must be doing their own thing with the video bid. We wouldn't want the cache behavior to prevent it from being added.

If someone _is_trying to use the cache, then they're _probably_ using `dfpAdServerVideo` too. In that case, a bid which didn't make it into the cache can only break things.

PR #1487, opened 2017-08-14
drop specific code for index adapter

mkendall07:
## Type of change
- [x] Bugfix

## Description of change
Resolves https://github.com/prebid/Prebid.js/issues/1341

### DIFF
--- a/src/bidmanager.js
+++ b/src/bidmanager.js
@@ -46,33 +46,18 @@ function getBidders(bid) {
 function bidsBackAdUnit(adUnitCode) {
   const requested = $$PREBID_GLOBAL$$._bidsRequested
     .map(request => request.bids
-      .filter(adUnitsFilter.bind(this, $$PREBID_GLOBAL$$._adUnitCodes))
       .filter(bid => bid.placementCode === adUnitCode))
-    .reduce(flatten, [])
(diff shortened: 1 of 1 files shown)


src/bidmanager.js
```diff
@@ -46,33 +46,18 @@ function getBidders(bid) {
 function bidsBackAdUnit(adUnitCode) {
   const requested = $$PREBID_GLOBAL$$._bidsRequested
     .map(request => request.bids
-      .filter(adUnitsFilter.bind(this, $$PREBID_GLOBAL$$._adUnitCodes))
```
protonate: I think the above line needs to remain in for now -- not related to index bid counting.

mkendall07: shouldn't the line below take care of filtering? I don't understand the need for the double filter.

protonate: This is the `adUnitsFilter` function:
```
export function adUnitsFilter(filter, bid) {
  return filter.includes((bid && bid.placementCode) || (bid && bid.adUnitCode));
}
```
so while it does behave somewhat differently by testing for `placementCode` and `adUnitCode` your right that the next filter cuts that down to just `placementCode`. Ok, makes sense.

PR #1492, opened 2017-08-16
Update gumgum entrypoint JS file

mxcoder:
## Type of change
- Other

## Description of change
GumGum recently rolled out a new CDN, and we want to update our entrypoint file to point to it.


mxcoder:
Thanks @matthewlane , please let me know if anything needs to be changed.

PR #1494, opened 2017-08-17
BaseAdapter for the Prebid 0.x -> 1.x transition 

dbemiller:
## Type of change
Refactoring (no functional changes, no api changes)

## Description of change
#1421 changes the APIs for both Adapters _and_ Publishers. We're planning on cutting a 1.0 branch soon... and this PR aims to reduce the maintenance cost of that branch during the transition period.

In a nutshell: the `singleRequestBidder` is a new "base" bidder intended to support any adapters which make only _one_ request to the server. Its Adapter-facing API is compatible with both Prebid 0.x _and_ Prebid 1.x.

Any compatible Adapters can code against this Base in `master`. If they do, they won't have to duplicate changes across branches whenever they do Adapter code maintenance.

Still TODO:
- User sync (waiting on #1549 to merge)

At this point, any API feedback on the `singleRequestBidder` would be super useful. Everything else is just details.


snapwich:
I like this.  I'll set up some time to review it more in-depth, but it might not be until next week.

src/adapters/bidderFactory.js
```diff
@@ -0,0 +1,240 @@
...
+ * This function aims to minimize the Adapter-specific code for "single request" Bidders. That is, if your adapter
+ * only makes a single request to the server for bids, regardless of how many bid requests or ad units are in
+ * the Auction, then this function is for you.
+ *
+ * @param {BidderSpec} spec An object containing the bare-bones functions we need to make a Bidder.
+ */
+export default function newBidder(spec) {
+  return Object.assign(new Adapter(spec.code), {
+    callBids: function(bidsRequest) {
+      if (!bidsRequest.bids || !bidsRequest.bids.filter) {
```
snapwich: This looks like an array check.  Would be more clear as `if (!Array.isArray(bidsRequest.bids) ) {`

src/adapters/bidderFactory.js
```diff
@@ -0,0 +1,240 @@
...
+        bidsRequest.bids
+          .map(bidRequest => bidRequest.placementCode)
+          .forEach(placementCode => {
+            if (placementCode && !placementCodesHandled[placementCode]) {
+              bidmanager.addBidResponse(placementCode, newEmptyBid());
+            }
+          });
+      }
+
+      const bidRequests = bidsRequest.bids.filter(filterAndWarn);
```
snapwich: These names are confusing, especially here.  I think `bidRequests` is fine, but `bidsRequest` I think is called `bidderRequest` in other parts of Prebid.

src/adapters/bidderFactory.js
```diff
@@ -0,0 +1,240 @@
...
+            logWarn(`Bidder ${spec.code} made bid for unknown request ID: ${bid.requestId}. Ignoring.`);
+          }
+        }
+
+        let bids;
+        try {
+          bids = spec.interpretResponse(response);
+        } catch (err) {
+          logError(`Bidder ${spec.code} failed to interpret the server's response. Continuing without bids`, null, err);
+          onResponse();
```
snapwich: I think the calling of `onResponse` here is a good use case for a `finally` clause so you don't have to scatter as many calls around.

I'd wrap all this code in a try block and put the `onResponse` in a finally so it's obvious that it will always be called.

src/adapters/bidderFactory.js
```diff
@@ -0,0 +1,240 @@
...
+      }
+
+      // If the server responds successfully, use the adapter code to unpack the Bids from it.
+      // If the adapter code fails, no bids should be added. After all the bids have been added, make
+      // sure to call the `onResponse` function so that we're one step closer to calling fillNoBids().
+      function onSuccess(response) {
+        const bidRequestMap = {};
+        bidRequests.forEach(bid => {
+          bidRequestMap[bid.bidId] = bid;
+        });
```
snapwich: Why is this `bidRequestMap` recreated for each successful request?  Couldn't it just be made once in the same context as the `bidRequests`?

modules/appnexusAstBidAdapter.js
```diff
@@ -26,357 +24,323 @@ const NATIVE_MAPPING = {
...
+    return 'banner';
   }
-
-  return Object.assign(this, {
-    callBids: baseAdapter.callBids,
-    setBidderCode: baseAdapter.setBidderCode,
-  });
 }
 
-adaptermanager.registerBidAdapter(new AppnexusAstAdapter(), 'appnexusAst', {
```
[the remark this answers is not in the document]
dbemiller: This never even occurred to me... it's a great idea though

src/adapters/bidderFactory.js
dbemiller: I like the goal here... but I'm really trying not to catch unexpected errors either.

For example, while writing this I originally had a bug in my `addBidUsingRequestMap` function where something was unexpectedly undefined.

I easily may never have caught that if it'd been inside the `try` block.

src/ajax.js
```diff
@@ -5,6 +5,9 @@ var utils = require('./utils');
 const XHR_DONE = 4;
 let _timeout = 3000;
 
+export const GET = 'GET';
+export const POST = 'POST';
```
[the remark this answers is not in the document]
dbemiller: Ironically... I actually added these because I couldn't remember where this ajax function was expecting `GET` or `get`. Figured this would save me from having to look it up the next time.

I must admit that I don't care much either way, though. I wouldn't mind voting on it (thumbs up for `const`, thumbs down for raw strings)

modules/appnexusAstBidAdapter.js
```diff
@@ -26,357 +24,323 @@ const NATIVE_MAPPING = {
...
-      utils.logError(errorMessage);
-
-      // signal this response is complete
-      Object.keys(bidRequests)
-        .map(bidId => bidRequests[bidId].placementCode)
-        .forEach(placementCode => {
-          bidmanager.addBidResponse(placementCode, createBid(STATUS.NO_BID));
-        });
-      return;
+// TODO: Patch UserSync into the spec, once the UserSync PR has been merged.
```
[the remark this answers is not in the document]
dbemiller: that's a good question... and I don't know. I actually chatted with our team about it, but didn't walk away with any super clear answers.

I'm not familiar enough with AdTech to know how it's done... but I could imagine three places where user syncs _might_ come from.

1. Constant values. "Always ping this usersync endpoint on every auction"
2. Server-side logic: "Always ping this endpoint to _get_ the usersync info. It will return one or more syncs"
3. Winning bids. "Because this bid made it into the auction, drop this usersync pixel"

Depending on which cases are worth supporting, I'd choose different APIs. It'll require adding one or more optional properties to the spec:

For (1), it'd be `getSyncs()`
For (2), it'd be `getSyncs(done)`
For (3), it'd be `getSyncs(serverResponse)`

In all cases, the return value (or callback argument) would just be an object with the data which we need to call `registerSync` in #1549.

Do you have any better background or opinion on it? I'm interested to hear what other adapter maintainers think too, if they see this.

src/adapters/bidderFactory.js
```diff
@@ -0,0 +1,240 @@
...
+  function newEmptyBid() {
+    const bid = bidfactory.createBid(STATUS.NO_BID);
+    bid.code = spec.code;
+    bid.bidderCode = spec.code;
+    return bid;
+  }
+}
+
+/**
+ * @typedef {object} ServerRequest
```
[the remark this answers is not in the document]
dbemiller: I've been thinking about this... but I'm not so convinced.

First: `instanceof` is a pretty flimsy check in javascript. For example:

```
var serverRequest = new ServerRequest(...);
delete serverRequest.type

serverRequest.type // Is undefined
serverRequest instanceof Request // Is true
```

Second: it sounds like all the complex validation code has just moved into the constructor. Rather than `if (!serverRequest.type)`, the constructor has to do `if (!type)`. I can see some advantage in having the validation code centralized... but if that's the goal, I'd rather define it an `isRequest(obj)` function somewhere.

Third: constructors are a bad pattern. To be honest, I don't expect the drawbacks to affect this area much... but I do want to discourage them from spreading around the codebase.  There's some good discussion on them in (2) and (3) [here](https://medium.com/javascript-scene/javascript-factory-functions-vs-constructor-functions-vs-classes-2f22ceddf33e)

Third: I'm open to discussion here... but I was kind of hoping we can _avoid_ validating the data at runtime. React did the profiling and found that data validation was such a performance drain that they made separate "dev" and "prod" versions of their library. The "prod" one skips all property validation in your Components.

----------------
Data validation is an open question as far as I'm concerned, though.

I left it out here because I was thinking of how React totally disabled all runtime property validation in the `prod` version of their library because it was found to be a notable performance drain.

I _think_ it should be unnecessary so long as the unit tests check for it. I have a followup PR I hope to make for "unified adapter tests"--assuming management schedules the time for me to do it--which would do the validation at test time, rather than runtime.

src/adapters/bidderFactory.js
```diff
@@ -0,0 +1,240 @@
...
+
+/**
+ * This function aims to minimize the Adapter-specific code for "single request" Bidders. That is, if your adapter
+ * only makes a single request to the server for bids, regardless of how many bid requests or ad units are in
+ * the Auction, then this function is for you.
+ *
+ * @param {BidderSpec} spec An object containing the bare-bones functions we need to make a Bidder.
+ */
+export default function newBidder(spec) {
+  return Object.assign(new Adapter(spec.code), {
```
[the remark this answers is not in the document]
dbemiller: Added support for these. I'm not sure this is the best way, though... so let me know if you have better ideas.

If we think this API is general enough to be the _only_ base class, I might re-organize the code a bit so that "bidder creation" is in a separate file from "bidder registration".

I still haven't heard anyone weigh in on whether or not this handles all the 1.0 use-cases, though, or whether i'm overly-constraining adapters with it.

src/adapters/bidderFactory.js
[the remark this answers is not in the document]
snapwich: It's true, there are ways to subvert `instanceof` checks.  The reason I like them is because, as you said, type validation is expensive and `instanceof` checks are not.  Using a constructor function to define a type allows you to incur a one-time hit for the validation of an object (and then use `instanceof` thereafter), as opposed to `isRequest(obj)` which is a many-time hit for each time that object is used.  That's the problem React encountered as they had to validate propTypes each type a component instance was used.

Either way, this isn't a make or break issue for me, it's just something I think of every time I see `if (obj.property ...`.  Also, if we did do this we'd have to expose the constructor functions on the public API as well so external adapters could pass the `instanceof` checks, which isn't terrible but it's more of a hassle than just using typedef.

src/adapters/bidderFactory.js
```diff
@@ -0,0 +1,270 @@
...
+ * @param {string|object} data Data to be sent in the request.
+ *   If this is a GET request, they'll become query params. If it's a POST request, they'll be added to the body.
+ *   Strings will be added as-is. Objects will be unpacked into query params based on key/value mappings, or
+ *   JSON-serialized into the Request body.
+ */
+
+/**
+ * @typedef {object} BidRequest
+ *
+ * @param {string} bidId A string which uniquely identifies this BidRequest in the current Auction.
```
snapwich: These should be `@property` instead of `@param` for typedefs.

src/mediaTypes.js
```diff
@@ -0,0 +1,15 @@
...
+ *
+ * All adapters are assumed to support banner ads. Other media types are specified by Adapters when they
+ * register themselves with prebid-core.
+ */
+
+/**
+ * @typedef {('native'|'video'|'banner')} MediaType
+ */
+
+/** @ype MediaType */
```
snapwich: I think you have a typo here

src/ajax.js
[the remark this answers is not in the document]
snapwich: If this were our API, then I'd probably be more open to the idea of adding these constants.  But this is just a layer on top of an existing API that's already pretty standardized around using 'GET', 'POST', etc strings: https://developer.mozilla.org/en-US/docs/Web/API/XMLHttpRequest/open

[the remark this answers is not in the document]
dbemiller: aha... ok, that makes plenty of sense then. My javascript inexperience is showing itself... didn't know those were so well established.

I'll take em out.

src/adapters/bidderFactory.js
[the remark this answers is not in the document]
dbemiller: yeah... I do understand where you're coming from, but I think I'd prefer to hold off on this for now.

modules/appnexusAstBidAdapter.js
```diff
@@ -27,364 +24,349 @@ const NATIVE_MAPPING = {
...
-  let bidRequests = {};
-  let usersync = false;
-
-  /* Prebid executes this function when the page asks to send out bid requests */
-  baseAdapter.callBids = function(bidRequest) {
-    bidRequests = {};
+export const spec = {
+  code: BIDDER_CODE,
+  supportedMediaTypes: [VIDEO, NATIVE],
+  aliases: ['dummy'],
```
mkendall07: nice :)

modules/appnexusAstBidAdapter.js
```diff
@@ -27,364 +24,349 @@ const NATIVE_MAPPING = {
...
-          source: SOURCE,
-          version: '$prebid.version$'
+  /**
+   * Unpack the response from the server into a list of bids.
+   *
+   * @param {*} serverResponse A successful response from the server.
+   * @return {Bid[]} An array of bids which were nested inside the server.
+   */
+  interpretResponse: function(serverResponse) {
+    const parsed = JSON.parse(serverResponse);
```
mkendall07: suppose the base adapter could do the parse, unless we need custom error handling in the adapter (I don't see any here).

src/adapters/bidderFactory.js
```diff
@@ -0,0 +1,298 @@
...
+ * @property {number} cpm The bid price, in US cents per thousand impressions.
+ * @property {number} height The height of the ad, in pixels.
+ * @property {number} width The width of the ad, in pixels.
+ *
+ * @property [Renderer] renderer A Renderer which can be used as a default for this bid,
+ *   if the publisher doesn't override it. This is only relevant for Outstream Video bids.
+ */
+
+/**
+ * @typedef {object} BidderSpec An object containing the adapter-specific functions needed to
```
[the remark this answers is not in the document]
dbemiller: I moved these, but I'm uh.... not sure I did what you had in mind.

The first mention of `BidderSpec` is inside an `@param {BidderSpec} spec`. That block documents the function... so the closest places for the `typedef` are either before or after the function implementation.

I moved it before the function implementation... which might be better? I dunno though.

Let me know if you have something better in mind. The `BidderSpec` is used at the top-level functions... and all the other `typedef` blocks in here are used inside the `BidderSpec`.

src/adapters/bidderFactory.js
```diff
@@ -0,0 +1,301 @@
...
+
+      // callBids must add a NO_BID response for _every_ placementCode, in order for the auction to
+      // end properly. This map stores placement codes which we've made _real_ bids on.
+      //
+      // As we add _real_ bids to the bidmanager, we'll log the placementCodes here too. Once all the real
+      // bids have been added, fillNoBids() can be called to end the auction.
+      //
+      // In Prebid 1.0, this will be simplified to use the `addBidResponse` and `done` callbacks.
+      const placementCodesHandled = {};
+      function addBidWithCode(placementCode, bid) {
```
jaiminpanchal27: We are using two names for same thing in our codebase i.e placementCode and adUnitCode. From 1.0 we should just use `code` or `adUnitCode`
Gist for adUnit changes https://gist.github.com/mkendall07/51ee5f6b9f2df01a89162cf6de7fe5b6#adunit-obj

dbemiller: Did not know that... I removed `placementCode` from all the comments and local variable names.

I think I still have to use it on the `bidRequest`, though, right? Since this code is supposed to be good for Prebid 0.x.

PR #1501, opened 2017-08-18
fix undefined bidder error in prebidServer

mkendall07:
## Type of change
- [x] Bugfix

## Description of change
If the user specifies a bidder code inside the bidder list for S2S adapters, we create empty bids for each one even if it's invalid (not requested in the first place). This PR filters the list to only those bidders requested.

### DIFF
--- a/modules/prebidServerBidAdapter.js
+++ b/modules/prebidServerBidAdapter.js
@@ -105,8 +105,13 @@ function PrebidServer() {
       is_debug: isDebug
     };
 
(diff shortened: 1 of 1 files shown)


protonate:
This change will mean that bidders not included in the set of ad units under auction will not be included in the PBS request, which makes good sense for the purpose of the auction, but this also means that those bidders would neither enroll for a user sync. That's probably fine but wanted to call that out.

Issue #1497, opened 2017-08-18
AudienceNetwork adapter bug: If the page url has hashtag, the sdk query string will be dropped when the request goes to an.facebook.com


brianxieseattle:
## Type of issue
Bug

## Description
If the page url that hosting prebid to call AudienceNetwork has a hashtag such as 
https://www.testweb.com/?page=here#path=root

The request to Facebook will fail.

## Steps to reproduce
1. Have a web site that url looking like 
https://www.testweb.com/?page=here#path=root
2. Put Prebid.js with AudienceNetwork adapter on the web site and inspect the success or failure of the call

### Expected results
The outgoing request should be something like
https://an.facebook.com/v2/placementbid.json?placementids[]=12345678&adformats[]=native&testmode=false&pageurl=https://www.testweb.com/?page=here#path=root&sdk[]=5.5.web

### Actual results
The outgoing request looks like
https://an.facebook.com/v2/placementbid.json?placementids[]=12345678&adformats[]=native&testmode=false&pageurl=https://www.testweb.com/?page=here
and the response comes back as the failure

## Platform details
does not matter

## Other information
Root cause seems to be:
AudienceNetwork adapter uses the following codes to construct the request url
```
    const pageurl = location.href;
    const search = {
      placementids,
      adformats,
      testmode,
      pageurl,
      sdk
    };
    
        const url = format({
      protocol: 'https',
      host: 'an.facebook.com',
      pathname: '/v2/placementbid.json',
      search
    });
```
sdk part is after the pageurl part

The format function from url.js is 

```
export function format(obj) {
  return (obj.protocol || 'http') + '://' +
         (obj.host ||
          obj.hostname + (obj.port ? `:${obj.port}` : '')) +
         (obj.pathname || '') +
         (obj.search ? `?${formatQS(obj.search || '')}` : '') +
         (obj.hash ? `#${obj.hash}` : '');
}
```
It will treat hashtag as the last optional part and drops everything after.

The fix could be either of the following one:
1. AudienceNetwork adapter should consider encoding the pageUrl before sending to the format function. Meanwhile, it should also consider the length of the url as some browsers could not handle the url length which is larger than 1024
2. The generic format function should encode everything in the search part to protect itself. 

Please evaluate which route is probably the better fix.

lovell:
Hello, #1502 provides a possible solution within the Audience Network adaptor.

[...]

PR #1502, opened 2017-08-19
Audience Network: ensure pageurl is always URI encoded

lovell:
## Type of change
- [X] Bugfix

## Description of change
Hello, this is a fix for #1497 that ensures `pageurl` is correctly URI encoded before calling Prebid's URL `format` function.

It also adds test expectations to verify this behaviour.

This work was commissioned and paid for by Facebook.

### DIFF
--- a/modules/audienceNetworkBidAdapter.js
+++ b/modules/audienceNetworkBidAdapter.js
@@ -127,7 +127,7 @@ const createSuccessBidResponse = (placementId, size, format, bidId, cpmCents, pa
   bid.fb_placementid = placementId;
   // Video attributes
   if (isVideo(format)) {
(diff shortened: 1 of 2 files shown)


andyblackwell:
Looks like this fix was reverted in https://github.com/prebid/Prebid.js/commit/403a991615271612ab26b5b915f9654972234571#diff-97dadabd4d374d37991cffac1d050d0f

lovell:
@andyblackwell Ah yes, the upgrade to the Prebid 1.0 API involved a pretty big change that also had to involve refactoring unit tests and this one got lost. If you're able to submit a PR for this I'd be happy to review.

lovell:
@andyblackwell This is covered by PR #1873, thanks for reporting.

PR #1504, opened 2017-08-21
Add adapter parameter types

matthewlane:
Adds districtm and pulsepoint adapter parameter types


modules/prebidServerBidAdapter.js
```diff
@@ -52,7 +52,18 @@ const paramTypes = {
   'pubmatic': {
     'publisherId': tryConvertString,
     'adSlot': tryConvertString
-  }
+  },
+  'districtmDMX': {
```
mkendall07: I think this should be `districtm` to match https://github.com/prebid/prebid-server/blob/master/pbs_light.go#L628

Also the params should be the same as AppNexus (since it's an alias).

PR #1506, opened 2017-08-21
Make bid.vastUrl use the cache URL if the bid didnt already have one.

dbemiller:
## Type of change
Feature

## Description of change
If the pub is using prebid cache, and the bidder doesn't include their own `vastUrl` on the bid, then we can use the prebid cache one.

This just helps normalize the data in the bid that the publisher sees.

### DIFF
--- a/src/videoCache.js
+++ b/src/videoCache.js
@@ -11,7 +11,7 @@
 
 import { ajax } from './ajax';
 
-const ENDPOINT = 'https://prebid.adnxs.com/pbc/v1/cache'
+const BASE_URL = 'https://prebid.adnxs.com/pbc/v1/cache'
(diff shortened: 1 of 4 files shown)


src/videoCache.js
```diff
@@ -124,3 +124,7 @@ export function store(bids, done) {
     withCredentials: true
   });
 }
+
+export function getCacheUrl(id) {
+  return `${PUT_URL}?uuid=${id}`;
```
mkendall07: This should be renamed to something like `base_url` or similar because `PUT` is confusing in this context (get op).

dbemiller: yeah... agreed.

PR #1507, opened 2017-08-21
use utils.getTopWindowUrl() to get top window URL for indexexchange a…

taihwasong:
## Type of change

- [ ] Bugfix
- [ ] Feature
- [ ] New bidder adapter   
- [ ] Build related changes
- [ ] CI related changes
- [ ] Other

## Description of change

- indexexchange adapter now uses the Prebid provided utility function (utils.getTopWindowUrl) to get top window URL

## Other information

No test cases are updated/added because this is a refactoring pull request. No functional changes

### DIFF
--- a/modules/indexExchangeBidAdapter.js
+++ b/modules/indexExchangeBidAdapter.js
@@ -269,11 +269,24 @@ var cygnus_index_start = function () {
     this.siteID = siteID;
     this.impressions = [];
     this._parseFnName = undefined;
+
+    // Get page URL
+    this.sitePage = undefined;
+    try {
+        this.sitePage = utils.getTopWindowUrl();
+    } catch (e) {}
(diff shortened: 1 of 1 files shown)


modules/indexExchangeBidAdapter.js
```diff
@@ -269,11 +269,24 @@ var cygnus_index_start = function () {
     this.siteID = siteID;
     this.impressions = [];
     this._parseFnName = undefined;
+
+    // Get page URL
+    this.sitePage = undefined;
+    try {
+      this.sitePage = utils.getTopWindowUrl();
```
mkendall07: this won't throw an exception, so the code below in the catch is unreachable.

taihwasong: Hi @mkendall07. Thank you for the follow-up. We will fix this when #1530 is merged in.

PR #1511, opened 2017-08-21
Match port when testing legacy browser

### LINKED ISSUE #1302: Fragile test fails in prebid-core on IE11
## Type of issue
Bug

## Description
A test is failing on IE11, Windows 8.1.

Computers can't habitually ignore errors like humans can. Failures like these prevent us from automating the release validation process.

## Steps to reproduce
Run `gulp test --browserstack` with the `bs_ie_11_windows_8.1` browser active.

### Expected results
The tests should pass reliably.

### Actual results
I see the following failure:

```
IE 11.0.0 (Windows 8.1 0.0.0) Unit: Prebid Module video adserverTag should return original adservertag if there are no bids for the given placement code FAILED
	expected 'https://pubads.g.doubleclick.net:443/gampad/ads?sz=640x480&iu=/19968336/header-bid-tag-0&impl=s&gdfp_req=1&env=vp&output=xml_vast2&unviewed_position_start=1&url=www.test.com' to equal 'https://pubads.g.doubleclick.net/gampad/ads?sz=640x480&iu=/19968336/header-bid-tag-0&impl=s&gdfp_req=1&env=vp&output=xml_vast2&unviewed_position_start=1&url=www.test.com'
	Error
	   at AssertionError (/Users/dbemiller/code/Prebid.js/test/spec/unit/pbjs_api_spec.js:1917:8)
	   at assert (/Users/dbemiller/code/Prebid.js/test/spec/unit/pbjs_api_spec.js:6249:8)
	   at assertEqual (/Users/dbemiller/code/Prebid.js/test/spec/unit/pbjs_api_spec.js:6766:8)
	   at Anonymous function (/Users/dbemiller/code/Prebid.js/test/spec/unit/pbjs_api_spec.js:5836:6)
	   at Anonymous function (/Users/dbemiller/code/Prebid.js/test/spec/unit/pbjs_api_spec.js:1556:8)
```

## Platform details
Windows 8.1, Browser IE11

matthewlane:
## Type of change
- Testfix

## Description of change
The `parse` function in urls.js returns a port of 443 when tested in IE11, blank in other browsers. A hardcoded url that is matched against doesn't contain this port and causes the test mentioned in #1302 to fail. This PR adds that port to the hardcoded string when tested in IE11 and fixes #1302.

### DIFF
--- a/test/spec/unit/pbjs_api_spec.js
+++ b/test/spec/unit/pbjs_api_spec.js
@@ -1571,11 +1571,15 @@ describe('Unit: Prebid Module', function () {
     });
 
     it('should return original adservertag if there are no bids for the given placement code', () => {
-      var options = {
+      // urls.js:parse returns port 443 for IE11, blank for other browsers
+      const ie11port = !!window.MSInputMethodContext && !!document.documentMode ? ':443' : '';
+      const adserverTag = `https://pubads.g.doubleclick.net${ie11port}/gampad/ads?sz=640x480&iu=/19968336/header-bid-tag-0&impl=s&gdfp_req=1&env=vp&output=xml_vast2&unviewed_position_start=1&url=www.test.com`;
+
+      const masterTagUrl = $$PREBID_GLOBAL$$.buildMasterVideoTagFromAdserverTag(adserverTag, {
         'adserver': 'dfp',
         'code': 'one-without-bids'
-      };
-      var masterTagUrl = $$PREBID_GLOBAL$$.buildMasterVideoTagFromAdserverTag(adserverTag, options);
+      });
+
       expect(masterTagUrl).to.equal(adserverTag);
     });
 


dbemiller:
Are you sure this is safe? The test is titled `should return original adservertag if there are no bids for the given placement code`.

I'm not sure why the original author thought this was an important test... but assuming that they were right, wouldn't this be a code bug?

mkendall07:
well this just adds the port number if the browser is IE 11, assuming this is right behavior for IE, since this implementation inside of `url` is based on the anchor tag, which is implementation specific.

dbemiller:
I'll merge it cause it's testing a deprecated method, and it fixes our CI :/

...but in general, we should be normalizing browser differences at as low a level as possible. jQuery was pretty much built on the idea... and consistency is the main motivation for the push for web standards at all.

If the implementation is browser-dependent, then I'd consider it a buggy implementation.

PR #1518, opened 2017-08-23
Fixed imports and made adform support aliasing

ptomasroos:
## Type of change
- [x] Feature

## Description of change
Updated the adform adapter to support aliasing since we're having a partner on the sell side who runs the adform platform and by allowing aliasing we can make them bid under their own bidderCode.


ptomasroos:
The failed test is related to TapSenseAdapter which I haven't touched at all.

ptomasroos:
The reason the test is broken is the same as https://github.com/prebid/Prebid.js/issues/1472

PR #1543, opened 2017-08-29
userSync is off by default

bretg:
## Type of change
Changing default so that DSP user sync is only invoked when the publisher requests it

## Description of change
changed the default to sync enabled to false


snapwich:
oops,  it broke a test you need to fix

Issue #1408, opened 2017-07-21
API docs should be autogenerated


rmloveland:
## Type of issue

This is a feature request.

## Description

Currently our API documentation is maintained separately.  In order to avoid docs getting out of sync with the code, we should consider generating docs from the code.

Once generated, the API docs could be copied over to the Prebid.org repo as part of the release process, or in whatever other way seems reasonable to the team.

We could do the same for the various module APIs.

I've attached a couple of Markdown API doc files generated by [documentation.js](https://github.com/documentationjs/documentation) (h/t @dbemiller for the pointer) -- perhaps we could consider adding a `gulp docs` target to generate these automatically (or whatever else works).

What do y'all think?

[publisher-api-reference.txt](https://github.com/prebid/Prebid.js/files/1166664/publisher-api-reference.txt)
[dfp-ad-server-module-api-reference.txt](https://github.com/prebid/Prebid.js/files/1166666/dfp-ad-server-module-api-reference.txt)

PR #1544, opened 2017-08-29
Fix `documentation lint` issues

rmloveland:
## Type of change

- [ ] Bugfix
- [ ] Feature
- [ ] New bidder adapter   
- [ ] Build related changes
- [ ] CI related changes
- [X] Other

## Description of change

Modify `prebid.js` to address issues found by running the `documentation lint` command.

In particular I needed to change the variable name from `cmd` to `command` in the function starting on this line in order to address the fact that using `cmd` as the variable name and as the array on the global object was confusing JSDoc.

https://github.com/prebid/Prebid.js/compare/fix-documentation-lint-issues?expand=1#diff-f8e049a1b6d4f9aa96fca41d2f7aa11dR781

I ran `gulp test` and everything passed, but I wanted to call that out just in case.

## Other information

Next baby step towards https://github.com/prebid/Prebid.js/issues/1408

### DIFF
--- a/src/prebid.js
+++ b/src/prebid.js
@@ -342,11 +342,11 @@ $$PREBID_GLOBAL$$.clearAuction = function() {
 };
 
 /**
- * @property {Object} defaults
- * @property {function} defaults.bidsBackHandler
- * @property {number} defaults.timeout
- * @property {Array} defaults.adUnits
- * @property {Array} defaults.adUnitCodes
+ * @param {Object} defaults
+ * @param {function} defaults.bidsBackHandler
+ * @param {number} defaults.timeout
+ * @param {Array} defaults.adUnits
+ * @param {Array} defaults.adUnitCodes
  */
 $$PREBID_GLOBAL$$.requestBids = function ({ bidsBackHandler, timeout, adUnits, adUnitCodes } = {}) {
   events.emit('requestBids');
--- a/src/prebid.js
+++ b/src/prebid.js
@@ -342,11 +342,11 @@ $$PREBID_GLOBAL$$.clearAuction = function() {
 };
 
 /**
- * @param {Object} defaults
- * @param {function} defaults.bidsBackHandler
- * @param {number} defaults.timeout
- * @param {Array} defaults.adUnits
- * @param {Array} defaults.adUnitCodes
+ * @param {Object} requestOptions
+ * @param {function} requestOptions.bidsBackHandler
+ * @param {number} requestOptions.timeout
+ * @param {Array} requestOptions.adUnits
+ * @param {Array} requestOptions.adUnitCodes
  */
 $$PREBID_GLOBAL$$.requestBids = function ({ bidsBackHandler, timeout, adUnits, adUnitCodes } = {}) {
   events.emit('requestBids');
--- a/src/prebid.js
+++ b/src/prebid.js
@@ -116,7 +116,7 @@ function setRenderSize(doc, width, height) {
  * This function returns the query string targeting parameters available at this moment for a given ad unit. Note that some bidder's response may not have been received if you call this function too quickly after the requests are sent.
  * @param  {string} [adunitCode] adUnitCode to get the bid responses for
  * @alias module:$$PREBID_GLOBAL$$.getAdserverTargetingForAdUnitCodeStr
- * @return {array}  returnObj return bids array
+ * @return {Array}  returnObj return bids array
  */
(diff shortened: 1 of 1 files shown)


src/prebid.js
```diff
@@ -342,11 +342,11 @@ $$PREBID_GLOBAL$$.clearAuction = function() {
...
- *
- * @param bidsBackHandler
- * @param timeout
- * @param adUnits
- * @param adUnitCodes
+ * @property {Object} defaults
+ * @property {function} defaults.bidsBackHandler
+ * @property {number} defaults.timeout
+ * @property {Array} defaults.adUnits
+ * @property {Array} defaults.adUnitCodes
```
dbemiller: I'm not positive, but... is this generating the right docs? I wouldn't expect to see `@property` used in this way.

A sample call would look something like:

```
pbjs.requestBids({
  timeout: ...,
  adUnits: ...,
  adUnitCodes: ...,
  bidsBackHandler: function(bids) { ... }
});
```

If so, then cool ^^.

rmloveland: I was trying to figure that out - mostly I wanted to shut up the linter.  I don't think (?) it's generating the right docs - it's generating the following:

```
### requestBids

**Parameters**

-   `$0` **[Object](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object)**  (optional, default `{}`)
    -   `$0.bidsBackHandler`  
    -   `$0.timeout`  
    -   `$0.adUnits`  
    -   `$0.adUnitCodes`  
-   `bidsBackHandler`  
-   `timeout`  
-   `adUnits`  
-   `adUnitCodes`  
```
I don't know if that is the right doc structure b/c I don't understand what this way of referring to arguments means:

`requestBids = function ({ bidsBackHandler, timeout, adUnits, adUnitCodes } = {})`

Does it mean that: 
1. This defaults to an empty object?
2. That providing the object with the right keys as input is optional?  Because it doesn't seem optional. (?)
3. Something else?

dbemiller: I tested a bit... I think what you want is to swap the `property` tags here with `param`. That makes more reasonable docs (screenshotted below)

screen shot 2017-08-30 at 9 21 14 am

See also: http://usejsdoc.org/tags-param.html#parameters-with-properties

rmloveland: Ah, makes sense after reading the doc.  OK, just switched back to `param`.

After changing back to `param`,  `documentation lint` still throwing one warning, but the message reads like its parser just isn't smart enough to understand the code/JSDoc, not like there is an actual problem, so it seems like we can safely ignore it:

```
warning  An explicit parameter named defaults was specified but didn't match inferred information $0
```

dbemiller: oh that's super annoying...

That sounds like an error meant to prevent mismatched names, like this:

```
/**
 * @param e ...
 */
function handleError(error) { ... }
```

My guess is that `$0` is just the placeholder value they assign because the fancy ES6 syntax doesn't actually give the param a name.

see: https://github.com/documentationjs/documentation/issues/270

src/prebid.js
```diff
@@ -342,11 +342,11 @@ $$PREBID_GLOBAL$$.clearAuction = function() {
 };
 
 /**
- *
- * @param bidsBackHandler
- * @param timeout
- * @param adUnits
- * @param adUnitCodes
+ * @param {Object} defaults
```
mkendall07: should we call this something else? Like `requestOptions` or something?

src/prebid.js
```diff
@@ -473,9 +473,9 @@ $$PREBID_GLOBAL$$.onEvent = function (event, handler, id) {
 };
 
 /**
- * @param {String} event the name of the event
+ * @param {string} event the name of the event
```
mkendall07: so confused with the lowercase string and uppercase Object

rmloveland: I know!  The change is based on the output of `documentation lint`, and also in the [Google Closure Compiler JSDoc guidelines](https://github.com/google/closure-compiler/wiki/Annotating-JavaScript-for-the-Closure-Compiler) FWIW.

dbemiller: that's why linters exist in the first place :).

Sometimes it's for consistency. Sometimes it's because your code is subtly incorrect.

My guess is that they were expecting to support Classes in the newer versions of JS, and the general principle was "primitives are lowercase. Classes are uppercase." Just speculation though.

src/prebid.js
rmloveland: Good point - changed to `requestOptions`.

Issue #1510, opened 2017-08-21
Intent to implement: Prebid Architecture


dbemiller:
# Motivation
Some recent projects have surfaced some cracks in Prebid's design.

- @protonate suggested a very smooth API for Prebid 1.0 which was ver reasonably shot down simply because it was "too big of a project." Most people liked the design, so these steps are intended to make it a _feasible_ project for Prebid 2.0.
- Testing code is prohibitively difficult. Some problems are described in #1508. While it's possible to define mocks, in practice it's simply not done. For example, if you call something like [resetAuction](https://github.com/prebid/Prebid.js/blob/master/test/spec/unit/pbjs_api_spec.js#L39) in a `beforeEach` block in a brand new `_spec` file, you'll see hundreds of errors in existing tests.
- I'm concerned that certain very reasonable features are extremely difficult, if not impossible, to implement. One concrete example was #1421. Two more potential ones are below:
  - As a publisher, I would like to run two auctions, but attach listeners to events from the first one.
  - As a publisher, I would like to run two auctions with different config values.

# Proposal
I believe these problems reduce to two main goals.

1. Minimize `$$PREBID_GLOBAL$$` 
2. Minimize the mutable, singleton state.

This issue is an RFC outlining a target architecture for Prebid moving forward.

# Architecture

## Registries
Registries are the only files allowed to have _mutable, singleton state_. This exception is required due to the nature of the dynamic `--modules=foo,bar,baz` step in the build.

Examples from today's use-cases include:

- AdapterRegistry
- AnalyticsAdapterRegistry
- AdServerRegistry (for a not-quite-public feature. Currently the file is the [adServerManager](https://github.com/prebid/Prebid.js/blob/master/src/adServerManager.js))
- BidPostprocessorRegistry (use-case is described in [this discussion](https://github.com/prebid/Prebid.js/pull/1374#discussion_r131147143)) 

These files export "plugin points" for Modules, and essentially exist to answer the question "how should prebid-core behave, given the modules which exist in this build?"

## Prebid-core Utilities
Prebid-core utilities are all the `src` code which sits between the Publisher-facing API and the Registries.

These break up (conceptually) into three subcategories.

### Stateful
Stateful utilities deal with the mutable state. Some examples today include:
- [Events](https://github.com/prebid/Prebid.js/blob/master/src/events.js)
- [Cookie syncs](https://github.com/prebid/Prebid.js/blob/master/src/cookie.js) (refactor incoming in #1229)

[...]

PR #1549, opened 2017-08-31
Rubicon project improvement/user sync

mkendall07:
continued from [here](https://github.com/prebid/Prebid.js/pull/1229)


src/utils.js
```diff
@@ -452,29 +452,19 @@ exports.insertElement = function(elm, doc, target) {
...
 /**
  * Inserts empty iframe with the specified `url` for cookie sync
  * @param  {string} url URL to be requested
  * @param  {string} encodeUri boolean if URL should be encoded before inserted. Defaults to true
  */
-exports.insertCookieSyncIframe = function(url, encodeUri) {
-  let iframeHtml = this.createTrackPixelIframeHtml(url, encodeUri);
+exports.insertUserSyncIframe = function(url) {
+  let iframeHtml = this.createTrackPixelIframeHtml(url, false);
+  iframeHtml = iframeHtml.replace('<iframe', '<iframe sandbox="allow-scripts"');
```
dbemiller: String parsing is very fragile... would suggest one of the following instead:

- Change `createTrackPixelIframeHtml` so that it accepts a 'sandbox' option, and builds the right HTML from it
- Create the `div` object first, and then set the sandbox attribute programatically

test/spec/userSync_spec.js
```diff
@@ -0,0 +1,179 @@
...
+// Use require since we need to be able to write to these vars
+const utils = require('../../src/utils');
+let { userSync, newUserSync } = require('../../src/userSync');
+
+describe('user sync', () => {
+  let triggerPixelStub;
+  let logWarnStub;
+  let timeoutStub;
+  let shuffleStub;
+  let insertElementStub;
```
dbemiller: This is unused/can be deleted.

src/userSync.js
```diff
@@ -0,0 +1,169 @@
+import * as utils from 'src/utils';
+import { config } from 'src/config';
+
+export function newUserSync(userSyncConfig) {
```
dbemiller: This looks really good... but there's one small tweak I'd like to propose:

```

const cookiesAreSupported = !utils.isSafariBrowser() && utils.cookiesAreEnabled();
export function newUserSync(userSyncDependencies) { ... }

export const userSync = newUserSync({
  config: config.getConfig('userSync'),
  cookiesAreSupported: cookiesAreSupported
});
```

It solves a testability problem you described in `userSync_spec` (see the comment there), and also makes this module fit into the larger Prebid architecture described in #1510.

test/spec/userSync_spec.js
```diff
@@ -0,0 +1,179 @@
...
+    userSync.syncUsers();
+    userSync.registerSync('image', 'testBidder', 'http://example.com/2');
+    userSync.syncUsers();
+    expect(triggerPixelStub.getCall(0)).to.not.be.null;
+    expect(triggerPixelStub.getCall(0).args[0]).to.exist.and.to.equal('http://example.com/1');
+    expect(triggerPixelStub.getCall(1)).to.be.null;
+  });
+
+  // Since cookie support is only checked when the module is loaded this test will not work, but a test that covers
+  // this scenario is important and this should be revisited.
```
dbemiller: If you make the tweaks to `userSync` (see the comment in that file), this becomes testable very easily.

```
userSync = newUserSync({
  config: ...,
  cookiesAreSupported: false
})
```

test/spec/modules/rubiconBidAdapter_spec.js
```diff
@@ -1030,17 +1031,16 @@ describe('the rubicon adapter', () => {
...
     const emilyUrl = 'https://tap-secure.rubiconproject.com/partner/scripts/rubicon/emily.html?rtb_ext=1';
-    let origGetConfig = window.$$PREBID_GLOBAL$$.getConfig;
 
     beforeEach(() => {
       bids = [];
 
       server = sinon.fakeServer.create();
-      clock = sinon.useFakeTimers();
+      // monitor userSync registrations
+      sinon.spy(userSync, 'registerSync');
```
dbemiller: I think this needs to be a stub?

Spies only observe things... so I think all the `registerSync` calls make real mutations to the real usersync global state.

Tests shouldn't have side effects like this.

src/userSync.js
```diff
@@ -0,0 +1,169 @@
...
+    };
+  }
+
+  /**
+   * @function fireSyncs
+   * @summary Trigger all user syncs in the queue
+   * @private
+   */
+  function fireSyncs() {
+    if (!userSyncConfig.syncEnabled || !cookiesAreSupported || hasFired) {
```
dbemiller: If you're in a browser which doesn't support cookies, will the user syncs just build up inside the queue? Since this returns before calling `queue = getDefaultQueue()`. I'm not sure it's a big deal... but I suppose it could chew up memory if people run lots of auctions on the same page. Could probably wait for a secondary PR, though.

On that note... isn't this module basically a no-op if the browser doesn't support cookies? Again, perhaps a comment for a secondary PR, but... we could improve performance with something like:

```
export function newUserSync(userSyncDependencies) {
  if (!userSyncDependencies.cookiesAreSupported) {
    return {
      // Not pseudocode; these are literal no-ops
      registerSync: function() { }
      syncUsers: function() { }
      syncUsersOverride: function() { }
    }
  } else {
    // All the existing code & public API goes here
  }
}
```

src/prebid.js
```diff
@@ -334,7 +335,13 @@ $$PREBID_GLOBAL$$.removeAdUnit = function (adUnitCode) {
 
 $$PREBID_GLOBAL$$.clearAuction = function() {
   auctionRunning = false;
-  syncCookies(config.getConfig('cookieSyncDelay'));
+  // Only automatically sync if the publisher has not chosen to "enableOverride"
+  let userSyncConfig = config.getConfig('userSync') || {};
```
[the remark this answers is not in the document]
harpere: @jaiminpanchal27  `userSync` can be set with `config.setConfig({userSync: null});`  though I'm not sure why anyone would set it to `null` or another `falsey` value, unless maybe they thought it would turn it off.

mkendall07:
@harpere 
any updates here?

src/userSync.js
```diff
@@ -74,6 +75,8 @@ export function newUserSync(userSyncDependencies) {
       utils.logMessage(`Invoking image pixel user sync for bidder: ${bidderName}`);
       // Create image object and add the src url
       utils.triggerPixel(trackingPixelUrl);
+      // track that bidder has been synced
+      setBidderSynced(bidderName);
```
mkendall07: @harpere for now, let's drop these as we are only storing keys for S2S bidders and not client side. Later on we can integrate them, possibly by appending `s2s${biddercode}` to differentiate.

PR #1552, opened 2017-08-31
Tremor Video Bid Adapter

tremorvideo:
## Type of change

- [x] New bidder adapter   

## Description of change

Tremor video bid adapter. 

```
{
  bidder: 'tremor',
  params: {
    adCode: "ssp-!demo!-lufip",
    supplyCode: "ssp-%21demo%21-rm6rh"
  }
}
```
- contact email: vprasad@tremorvideo.com

### DIFF
--- a/modules/tremorBidAdapter.js
+++ b/modules/tremorBidAdapter.js
@@ -67,8 +67,8 @@ function TremorAdapter() {
       OPTIONAL_PARAMS.forEach(param => {
         if (bid.params[param]) {
           url += ('&' + param + '=' + bid.params[param]);
-        } }
-      );
+        }
+      });
 
       url = (url + '&fmt=json');
 
--- a/modules/tremorBidAdapter.js
+++ b/modules/tremorBidAdapter.js
@@ -39,7 +39,6 @@ function TremorAdapter() {
   /* Prebid executes this function when the page asks to send out bid requests */
   baseAdapter.callBids = function (bidRequest) {
     const bids = bidRequest.bids || [];
-
     bids.filter(bid => valid(bid))
       .map(bid => {
         let url = generateUrl(bid);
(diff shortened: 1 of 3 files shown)


modules/tremorBidAdapter.js
```diff
@@ -0,0 +1,169 @@
...
+    let bid = bidfactory.createBid(status, reqBid);
+    bid.code = baseAdapter.getBidderCode();
+    bid.bidderCode = baseAdapter.getBidderCode();
+
+    if (response) {
+      bid.cpm = response.price;
+      bid.crid = response.crid;
+      bid.vastXml = response.adm;
+      // This is a dummy vast URL, Tremor sends back the vast XML (the line above).
+      bid.vastUrl = 'http://cdn.tremorhub.com/static/dummy.xml';
```
matthewlane: Bid responses with a `vastXml` property will now pass validation with #1556 in master, so if Tremor sends xml and not a vast url, this line can be dropped

yarn.lock
```diff
@@ -1945,7 +1945,7 @@ d@1:
   dependencies:
     es5-ext "^0.10.9"
 
-"dargs@github:christian-bromann/dargs":
+dargs@christian-bromann/dargs:
```
dbemiller: This doesn't look intentional... not sure why it changed?

matthewlane: Running `yarn install` on current master resulted in this update to the lockfile for me too. I tried deleting `node_modules` then `yarn install` again with this changed lockfile and things still build out properly

dbemiller: ahh... common mistake. I ran into this recently, and it's just a weird API on their part.

`yarn install` pulls the latest dependencies which `package.json` allows. This might change your `yarn.lock`, if someone released a new version of a package recently.

`yarn install --pure-lockfile` installs the dependencies from `yarn.lock` without changing them.

Very unlikely to do any real harm, though.

PR #1555, opened 2017-09-01
AppnexusAst adapter update: Added source and version to request payload

jaiminpanchal27:
## Type of change
- [x] Feature
- [x] Other

## Description of change
Added source and prebid version to AppnexusAst adapter request payload

### DIFF
--- a/test/spec/modules/appnexusAstBidAdapter_spec.js
+++ b/test/spec/modules/appnexusAstBidAdapter_spec.js
@@ -130,11 +130,6 @@ describe('AppNexusAdapter', () => {
     });
 
     it('should add source and verison to the tag', () => {
-      REQUEST.sdk = {
-        source: 'pbjs',
-        version: '$prebid.version$'
-      };
-
       adapter.callBids(REQUEST);
 
       const request = JSON.parse(requests[0].requestBody);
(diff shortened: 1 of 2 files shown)


test/spec/modules/appnexusAstBidAdapter_spec.js
```diff
@@ -129,6 +129,24 @@ describe('AppNexusAdapter', () => {
       delete REQUEST.bids[0].params.user;
     });
 
+    it('should add source and verison to the tag', () => {
+      REQUEST.sdk = {
+        source: 'pbjs',
+        version: '$prebid.version$'
+      };
```
matthewlane: Setting this isn't required for the test to pass because the `sdk` object is added within the adapter. Calling `callBids` and examining the request as you have below should be enough

test/spec/modules/appnexusAstBidAdapter_spec.js
```diff
@@ -129,6 +129,24 @@ describe('AppNexusAdapter', () => {
...
+      adapter.callBids(REQUEST);
+
+      const request = JSON.parse(requests[0].requestBody);
+      expect(request.sdk).to.exist;
+      expect(request.sdk).to.deep.equal({
+        source: 'pbjs',
+        version: '$prebid.version$'
+      });
+
+      delete REQUEST.sdk;
```
matthewlane: Can be dropped too

test/spec/modules/appnexusAstBidAdapter_spec.js
jaiminpanchal27: Done

PR #1560, opened 2017-09-05
Serverbid alias

jgrimes:
## Type of change
- [x] Feature

## Description of change
Added extra configuration for Serverbid adapter and two adapter alises, `connectad` and `oft-media`.

### DIFF
--- a/modules/serverbidBidAdapter.js
+++ b/modules/serverbidBidAdapter.js
@@ -195,5 +195,6 @@ ServerBidAdapter.createNew = function() {
 
 adaptermanager.registerBidAdapter(new ServerBidAdapter, 'serverbid');
 adaptermanager.aliasBidAdapter('serverbid', 'connectad');
+adaptermanager.aliasBidAdapter('serverbid', 'oft-media');
 
 module.exports = ServerBidAdapter;
--- a/modules/serverbidBidAdapter.js
+++ b/modules/serverbidBidAdapter.js
@@ -8,8 +8,17 @@ import adaptermanager from 'src/adaptermanager';
 const ServerBidAdapter = function ServerBidAdapter() {
   const baseAdapter = new Adapter('serverbid');
 
-  const SERVERBID_BASE_URI = 'https://e.serverbid.com/api/v2';
-  const SMARTSYNC_BASE_URI = 'https://s.zkcdn.net/ss';
+  const CONFIG = {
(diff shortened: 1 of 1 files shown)


modules/serverbidBidAdapter.js
```diff
@@ -44,20 +53,29 @@ const ServerBidAdapter = function ServerBidAdapter() {
...
   const bidIds = [];
 
   baseAdapter.callBids = function(params) {
-    if (params && params.bids && utils.isArray(params.bids) && params.bids.length) {
+    if (params && params.bids &&
+        utils.isArray(params.bids) &&
+        params.bids.length &&
+        CONFIG[params.bidderCode]) {
+      const config = CONFIG[params.bidderCode];
+      config.request = window[params.bidderCode.toUpperCase() + '_CONFIG'];
```
jaiminpanchal27: @jgrimes What is the purpose of adding this line. This will be set to undefined, i guess your library will set window[params.bidderCode.toUpperCase() + '_CONFIG'];

jgrimes: @jaiminpanchal27 It is to allow people using our adapter (or aliases) to add arbitrary additional configuration/information on the requests being made to us.

modules/serverbidBidAdapter.js
```diff
@@ -173,6 +191,12 @@ const ServerBidAdapter = function ServerBidAdapter() {
   };
 };
 
+ServerBidAdapter.createNew = function() {
+  return new ServerBidAdapter();
+};
+
 adaptermanager.registerBidAdapter(new ServerBidAdapter(), 'serverbid');
+adaptermanager.aliasBidAdapter('serverbid', 'connectad');
+adaptermanager.aliasBidAdapter('serverbid', 'oft-media');
```
mkendall07: `oftmedia` is already a bid adapter alias of AppNexus. Should that be removed? It's confusing.

mkendall07:
@jgrimes 
can you explain what `oftmedia` is? It's already an alias fro AppNexus.

jgrimes:
@mkendall07 we've removed the `oft-media` alias for now but will want to add it in the future. The "owner" of the other `oftmedia` alias wants this one as well.
@jaiminpanchal27 I made an update to this code to fix our adapter alias and adjust some URLs. If you can give it another review it would be much appreciated.

PR #1565, opened 2017-09-06
Update JSDoc for `pbjs.enableAnalytics`

rmloveland:
## Type of change

- [ ] Bugfix
- [ ] Feature
- [ ] New bidder adapter   
- [ ] Build related changes
- [ ] CI related changes
- [X] Other

## Description of change

Update API docs for `pbjs.enableAnalytics`.

## Other information

Relates to https://github.com/prebid/Prebid.js/issues/1408

### DIFF
--- a/src/prebid.js
+++ b/src/prebid.js
@@ -601,8 +601,17 @@ $$PREBID_GLOBAL$$.loadScript = function (tagSrc, callback, useCache) {
 };
 
 /**
- * Will enable sending a prebid.js to data provider specified
- * @param  {Object} config object {provider : 'string', options : {}}
+ * Enable sending analytics data to the analytics provider of your
+ * choice.
+ *
+ * For usage, see [Integrate with the Prebid Analytics
+ * API](http://prebid.org/dev-docs/integrate-with-the-prebid-analytics-api.html).
+ *
+ * For a list of supported analytics adapters, see [Analytics for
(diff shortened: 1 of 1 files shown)


src/prebid.js
```diff
@@ -601,8 +601,17 @@ $$PREBID_GLOBAL$$.loadScript = function (tagSrc, callback, useCache) {
...
 /**
- * Will enable sending a prebid.js to data provider specified
- * @param  {Object} config object {provider : 'string', options : {}}
+ * Enable sending analytics data to the analytics provider of your
+ * choice.
+ *
+ * For usage, see [Integrate with the Prebid Analytics
+ * API](http://prebid.org/dev-docs/integrate-with-the-prebid-analytics-api.html).
+ *
+ * For a list of supported analytics adapters, see [Analytics for
```
dbemiller: Do you think it's worth mentioning the build step somewhere in here? Since the behavior of the method changes depending on which adapters you built with.

Or perhaps a better way is if we just make the code throw an error if you specify an invalid one. We could then make the error message say something like "No analytics adapter found for code XYZ. Did you forget to include it in your `gulp build --modules?`"

Not sure which one makes the better user experience? Maybe both...

protonate:
@dbemiller @rmloveland is there a resolution here to be merged?

dbemiller:
Was waiting on @rmloveland for a response... but it's definitely mergeable.

PR #1566, opened 2017-09-06
Add JSDoc for `pbjs.getAllWinningBids`

rmloveland:
## Type of change

- [ ] Bugfix
- [ ] Feature
- [ ] New bidder adapter   
- [ ] Build related changes
- [ ] CI related changes
- [X] Other

## Description of change

Add JSDoc for `pbjs.getAllWinningBids`.

## Other information

Relates to https://github.com/prebid/Prebid.js/issues/1408

### DIFF
--- a/src/prebid.js
+++ b/src/prebid.js
@@ -648,8 +648,43 @@ $$PREBID_GLOBAL$$.enableSendAllBids = function () {
 };
 
 /**
+ * The bid response object returned by an external bidder adapter during the auction.
+ * @typedef {Object} Bid
+ * @property {string} pbAg Auto granularity price bucket; CPM <= 5 ? increment = 0.05 : CPM > 5 && CPM <= 10 ? increment = 0.10 : CPM > 10 && CPM <= 20 ? increment = 0.50 : CPM > 20 ? priceCap = 20.00.  Example: `"0.80"`.
+ * @property {string} pbCg Custom price bucket.  For example setup, see {@link setPriceGranularity}.  Example: `"0.84"`.
+ * @property {string} pbDg Dense granularity price bucket; CPM <= 3 ? increment = 0.01 : CPM > 3 && CPM <= 8 ? increment = 0.05 : CPM > 8 && CPM <= 20 ? increment = 0.50 : CPM > 20? priceCap = 20.00.  Example: `"0.84"`.
+ * @property {string} pbLg Low granularity price bucket; $0.50 increment, capped at $5, floored to two decimal places.  Example: `"0.50"`.
+ * @property {string} pbMg Medium granularity price bucket; $0.10 increment, capped at $20, floored to two decimal places.  Example: `"0.80"`.
+ * @property {string} pbHg High granularity price bucket; $0.01 increment, capped at $20, floored to two decimal places.  Example: `"0.84"`.
+ *
+ * @property {string} bidder The string name of the bidder.  This *may* be the same as the `bidderCode`.  For For a list of all bidders and their codes, see [Bidders' Params](http://prebid.org/dev-docs/bidders.html).
+ * @property {string} bidderCode The unique string that identifies this bidder.  For a list of all bidders and their codes, see [Bidders' Params](http://prebid.org/dev-docs/bidders.html).
+ *
+ * @property {string} requestId The [UUID](https://en.wikipedia.org/wiki/Universally_unique_identifier) representing the bid request.
+ * @property {number} requestTimestamp The time at which the bid request was sent out, expressed in milliseconds.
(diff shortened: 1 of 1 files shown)


rmloveland:
Great idea Dave - I just pushed another update which adds a `Bid` type which can be used in the return value for this method (and a few others once this goes in).

dbemiller:
Awesome :). One small issue... since #1494 got merged, there's now a duplicate `Bid` type, with its own set of expectations. :/

The root of the issue is that the bid which we give publishers has more stuff on it than the bid which an Adapter returns.

I'm not sure what the best words are to describe the ideas here. We could rename the one in `bidderFactory` to `AdapterBid`, or rename the one you have here as `FinalBid`, or something like that. Do you have any preferences? Good naming is hard... and engineers suck at it :(.

src/prebid.js
```diff
@@ -647,6 +647,45 @@ $$PREBID_GLOBAL$$.enableSendAllBids = function () {
...
+ *
+ * @property {string} ad The actual ad creative content, often HTML with CSS, JavaScript, and/or links to additional content.  Example: `"<div id='beacon_-YQbipJtdxmMCgEPHExLhmqzEm' style='position: absolute; left: 0px; top: 0px; visibility: hidden;'><img src='http://aplus-...'/></div><iframe src=\"http://aax-us-east.amazon-adsystem.com/e/is/8dcfcd..." width=\"728\" height=\"90\" frameborder=\"0\" ...></iframe>",`.
+ * @property {number} ad_id The ad ID of the creative, as understood by the bidder's system.  Used by the line item's [creative in the ad server](http://prebid.org/adops/send-all-bids-adops.html#step-3-add-a-creative).
+ * @property {string} adUnitCode The code used to uniquely identify the ad unit on the publisher's page.
+ *
+ * @property {string} statusMessage The status of the bid.  Allowed values: `"Bid available"` or `"Bid returned empty or error response"`.
+ * @property {number} cpm The exact bid price from the bidder, expressed to the thousandths place.  Example: `"0.849"`.
+ *
+ * @property {Object} adserverTargeting An object whose values represent the ad server's targeting on the bid.
```
dbemiller: Your call, but... this might be worth a typedef of its own. There are a few `getAdserverTargeting` functions which (I _think_) return these same properties.

src/prebid.js
```diff
@@ -647,6 +647,45 @@ $$PREBID_GLOBAL$$.enableSendAllBids = function () {
...
+ * @property {string} bidderCode The unique string that identifies this bidder.  For a list of all bidders and their codes, see [Bidders' Params](http://prebid.org/dev-docs/bidders.html).
+ *
+ * @property {string} requestId The [UUID](https://en.wikipedia.org/wiki/Universally_unique_identifier) representing the bid request.
+ * @property {number} requestTimestamp The time at which the bid request was sent out, expressed in milliseconds.
+ * @property {number} responseTimestamp The time at which the bid response was received, expressed in milliseconds.
+ * @property {number} timeToRespond How long it took for the bidder to respond with this bid, expressed in milliseconds.
+ *
+ * @property {string} size The size of the ad creative, expressed in `"AxB"` format, where A and B are numbers of pixels.  Example: `"320x50"`.
+ * @property {string} width The width of the ad creative in pixels.  Example: `"320"`.
+ * @property {string} height The height of the ad creative in pixels.  Example: `"50"`.
```
dbemiller: Are these really strings, and not numbers? yuck!

src/prebid.js
rmloveland: I agree but I think it might be better to do this in another PR for the sake of expediency.

rmloveland:
Just pushed a commit changing the type to `AdapterBidResponse`, for now.  I think with some additional work on the code side we can merge the ever-so-slightly-different bid data structures and then document them with one `typedef`.

dbemiller:
> I think with some additional work on the code side we can merge the ever-so-slightly-different bid data structures and then document them with one typedef.

That's fine... but one word of caution on here: you _could_ document every single object in the Prebid ecosystem if you pooled all the optional properties together, like so:

```
/**
 * @typedef {object} PrebidObject
 *
 * @param {number} [cpm] The value of the bid, if this object has a value.
 * @param {string} [code] An ID for the object, if relevant.
 * ...
*/
```

It would be accurate, since it documents all the properties which may or may not exist... but it would be totally unhelpful to both implementers and callers.

The main goal of JSDocs, and Types as a whole, is to align the expectations of the implementer and caller. So if the functions truly guarantee to return objects with different properties--even if most of them overlap--then you're probably better keeping the types separate.

rmloveland:
@protonate Do you mind taking a look at this one when you get a chance?  Happy to make any updates as needed!

PR #1575, opened 2017-09-11
don't calculate TTR if request not found

kitwestneat:
## Type of change

- [x ] Bugfix

## Description of change
If the bidrequest can't be found, for example on a very late bid, don't try to calculate TTR. If the start time 
is 0, then TTR will be huge and the current auction will immediately timeout.

### DIFF
--- a/src/bidmanager.js
+++ b/src/bidmanager.js
@@ -109,6 +109,13 @@ exports.addBidResponse = function (adUnitCode, bid) {
       utils.logError(errorMessage('No adUnitCode was supplied to addBidResponse.'));
       return false;
     }
+
+    const bidRequest = getBidderRequest(bid.bidderCode, adUnitCode);
+    if (!bidRequest.start) {
+      utils.logError(errorMessage('Cannot find valid matching bid request.'));
+      return false;
+    }
+
     if (bid.mediaType === 'native' && !nativeBidIsValid(bid)) {
       utils.logError(errorMessage('Native bid missing some required properties.'));
       return false;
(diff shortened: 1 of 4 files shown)


src/bidmanager.js
```diff
@@ -161,7 +161,10 @@ exports.addBidResponse = function (adUnitCode, bid) {
       adUnitCode
     });
 
-    bid.timeToRespond = bid.responseTimestamp - bid.requestTimestamp;
+
```
snapwich: this extra newline is causing a linter error and failing the build

snapwich:
other than the linting error this LGTM.  However, I'm wondering if we should be throwing this bid out even earlier if it isn't part of the current auction.  It seems to me if it's not a valid bid for this current auction, `isValidBid()` should return false.

kitwestneat:
That sounds good, I'll move the check to isValidBid()

src/bidmanager.js
```diff
@@ -109,6 +109,13 @@ exports.addBidResponse = function (adUnitCode, bid) {
       utils.logError(errorMessage('No adUnitCode was supplied to addBidResponse.'));
       return false;
     }
+
+    const bidRequest = getBidderRequest(bid.bidderCode, adUnitCode);
+    if (!bidRequest.start) {
```
mkendall07: if you trying to check for no bid request, I think this condition should be just ` if (!bidRequest)` otherwise you'll get an undefined error.

kitwestneat:
Currently, getBidderRequest returns { start: null, requestId: null } if it
can't find a valid request. I agree that have getBidderRequest return
undefined or false makes more sense though.

mkendall07:
oh I missed that. Seems odd but I'm not going to touch it now. Thanks

PR #1576, opened 2017-09-11
Remove do callbacks if needed

kitwestneat:
## Type of change

- [x] Bugfix

## Description of change
A couple bug fixes that were easier to combine into one PR. The first removes the bid parameter to calls to addBidToAuction, since that function doesn't actually accept any parameters. 

The second removes doCallbacksIfNeeded since the calling order with addBidToAuction wasn't standardized, and it seems like it should be. Specifically it seems like the callbacks should always be done after the bid is added to the auction.

### DIFF
--- a/src/bidmanager.js
+++ b/src/bidmanager.js
@@ -86,10 +86,10 @@ exports.addBidResponse = function (adUnitCode, bid) {
     prepareBidForAuction();
 
     if (bid.mediaType === 'video') {
-      tryAddVideoBid(bid);
+      tryAddVideoBid();
     } else {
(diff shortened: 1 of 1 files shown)


protonate:
@kitwestneat any update here?

kitwestneat:
I reuploaded a fix the same day...do you all need something else from me? I
guess it's no longer able to merge cleanly...

mkendall07:
@kitwestneat 
Can you break out the changes into smaller PR's are try and rebase this? Thanks.

kitwestneat:
ok I'll break them up

PR #1587, opened 2017-09-13
Rubicon Bidder Factory

dbemiller:
Bidder factory updates to support the use-cases surfaced by updating the Rubicon adapter.


snapwich:
Just noticed, the typedefs should probably be updated to reveal the new paramters

mkendall07:
@snapwich the only downside (IMO) of using typedefs.

PR #1589, opened 2017-09-14
Update spotx adapter to provide access to cache key

npeceniak:
- The response ad key is now set to the spotx cache key to be
	  used in DFP targeting.

## Type of change

- [X] Bugfix
- [ ] Feature
- [ ] New bidder adapter   
- [ ] Build related changes
- [ ] CI related changes
- [ ] Other

## Description of change

In the spotx video adapter we did not have direct access to the spotx cache key used in DFP targeting. Set the bid.ad key to this value. 

### DIFF
--- a/modules/spotxBidAdapter.js
+++ b/modules/spotxBidAdapter.js
@@ -85,7 +85,7 @@ function Spotx() {
 
       bid.cpm = KVP_Object.spotx_bid;
       bid.vastUrl = url;
-      bid.ad = url;
+      bid.ad = KVP_Object.spotx_ad_key;
 
       var sizes = utils.isArray(bidReq.sizes[0]) ? bidReq.sizes[0] : bidReq.sizes;
       bid.height = sizes[1];


modules/spotxBidAdapter.js
```diff
@@ -85,7 +85,7 @@ function Spotx() {
 
       bid.cpm = KVP_Object.spotx_bid;
       bid.vastUrl = url;
-      bid.ad = url;
+      bid.ad = KVP_Object.spotx_ad_key;
```
mkendall07: Does a key render?

npeceniak: No a key will not render a video however the url that will be used to retrieve the ad is saved under the bid.vastUrl key. We were saving the same url to both bid.vastUrl and bid.ad. The reason I added the key to the bid.ad parameter is to be used with custom DFP targeting. SpotX uses custom DFP parameters for spotx_bid and spotx_ad_key to target ads in DFP which we now have access to both in the bid response.

PR #1593, opened 2017-09-16
Concurrent auctions

jaiminpanchal27:
Continuing from here #1421


src/auction.js
```diff
@@ -0,0 +1,518 @@
...
+ */
+
+import { uniques, timestamp, adUnitsFilter, delayExecution, getBidderRequest } from './utils';
+import { getPriceBucketString } from './cpmBucketManager';
+import { NATIVE_KEYS, nativeBidIsValid } from './native';
+import { isValidVideoBid } from './video';
+import { getCacheUrl, store } from './videoCache';
+import { Renderer } from 'src/Renderer';
+import { config } from 'src/config';
+import { userSync } from 'src/userSync.js';
```
dbemiller: should be `src/userSync`

src/auction.js
```diff
@@ -0,0 +1,518 @@
...
+      if (bid.bidderCode && (bid.cpm > 0 || bid.dealId)) {
+        keyValues = getKeyValueTargetingPairs(bid.bidderCode, bid);
+      }
+
+      // use any targeting provided as defaults, otherwise just set from getKeyValueTargetingPairs
+      bid.adserverTargeting = Object.assign(bid.adserverTargeting || {}, keyValues);
+    }
+
+    function doCallbacksIfNeeded() {
+      if (bid.timeToRespond > $$PREBID_GLOBAL$$.cbTimeout + $$PREBID_GLOBAL$$.timeoutBuffer) {
```
dbemiller: Not specifically for this PR, but... should these be on `config` @matthewlane?

If they make sense as global configs, then they probably should. If not, then you should accept them as params here. I don't think it makes sense to read from $$PREBID_GLOBAL$$ in either case, though.

If they are moving to `config`, that change should go in `master`.

src/targeting.js
```diff
@@ -1,210 +1,218 @@
...
+
+const utils = require('./utils.js');
+var CONSTANTS = require('./constants.json');
 
-var targeting = exports;
 var pbTargetingKeys = [];
 
-targeting.resetPresetTargeting = function(adUnitCode) {
-  if (isGptPubadsDefined()) {
+export function newTargeting(auctionManager) {
```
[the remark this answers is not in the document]
jaiminpanchal27: Done. https://github.com/prebid/Prebid.js/pull/1606

src/adapters/bidderFactory.js
```diff
@@ -152,16 +150,17 @@ export function newBidder(spec) {
       const adUnitCodesHandled = {};
       function addBidWithCode(adUnitCode, bid) {
         adUnitCodesHandled[adUnitCode] = true;
-        bidmanager.addBidResponse(adUnitCode, bid);
+        addBidResponse(adUnitCode, bid);
       }
       function fillNoBids() {
```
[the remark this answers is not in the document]
jaiminpanchal27: Updated code

test/spec/modules/appnexusAstBidAdapter_spec.js
```diff
@@ -1,194 +1,165 @@
...
-  }]
-};
-
 describe('AppNexusAdapter', () => {
   const adapter = newBidder(spec);
 
-  describe('request function', () => {
-    let xhr;
-    let requests;
+  describe('inherited functions', () => {
```
dbemiller: This is a good test... but it's actually testing the `newBidder` implementation. It should move to `bidderFactory_spec` if it exists.

That said... nearly every test in `bidderFactory_spec` _calls_ `callBids`... so they'd all fail if it weren't there.

test/spec/modules/appnexusAstBidAdapter_spec.js
```diff
@@ -1,194 +1,165 @@
...
-    it('attaches valid video params to the tag', () => {
-      REQUEST.bids[0].params.video = {
+    it('sends bid request to ENDPOINT via POST', () => {
+      const request = spec.buildRequests(bidRequests);
+      expect(request.url).to.equal(ENDPOINT);
+      expect(request.method).to.equal('POST');
+    });
+
+    it('should attach valid video params to the tag', () => {
+      bidRequests[0].params.video = {
```
dbemiller: This is a side-effect. If the test fails, the `delete` line at the end will never be executed, which causes cascading failures.

You've got to either copy the object before mutating, or move this to a `beforeEach` and `afterEach` block. Either one is fine... but for what it's worth, I tend to find the first one easier. `JSON.parse(JSON.stringify(obj))` is an inefficient, but concise way to make a deep copy.

The same goes for some of these other tests as well.

test/spec/modules/appnexusAstBidAdapter_spec.js
```diff
@@ -217,59 +188,91 @@ describe('AppNexusAdapter', () => {
...
-        'statusMessage',
-        'Bid returned empty or error response'
-      );
+      let result = spec.interpretResponse(response);
+      expect(result.length).to.equal(0);
     });
 
     it('handles non-banner media responses', () => {
-      server.respondWith(JSON.stringify({
+      // debugger; // eslint-disable-line
```
dbemiller: Stray comment

jaiminpanchal27: Damn, i tend to forget these debugger's a lot. Removing it.

test/spec/modules/appnexusAstBidAdapter_spec.js
```diff
@@ -1,194 +1,165 @@
 import { expect } from 'chai';
 import { spec } from 'modules/appnexusAstBidAdapter';
 import { newBidder } from 'src/adapters/bidderFactory';
-import bidmanager from 'src/bidmanager';
 
 const ENDPOINT = '//ib.adnxs.com/ut/v3/prebid';
 
-const REQUEST = {
```
dbemiller: Also... shouldn't these tests be updated in master? There's nothing 1.0 or concurrent-auctions about them.

Did we give up on making those centralized tests so that adapter-specific ones only need to define JSON files?

test/spec/modules/appnexusAstBidAdapter_spec.js
jaiminpanchal27: Done

test/spec/modules/appnexusAstBidAdapter_spec.js
jaiminpanchal27: Yes, tests can be updated for master as well. I will create a PR for that. 

Plan for JSON file based tests is still on. I will be adding that pretty soon.

test/spec/modules/appnexusAstBidAdapter_spec.js
jaiminpanchal27: Yes it can be added their as well. 

In bidderFactory_spec we are using sampleBidder. And here it tests the actual implementation of bidder. So this would test that implementation is correct and it has inherited needed functions

src/auction.js
```diff
@@ -0,0 +1,518 @@
...
+  };
+
+  return {
+    getAuctionStatus: () => _auctionStatus,
+    getAdUnits: () => _adUnits,
+    getAdUnitCodes: () => _adUnitCodes,
+    getBidRequests: () => _bidderRequests,
+    getBidsReceived: () => _bidsReceived,
+    startAuctionTimer,
+    callBids
```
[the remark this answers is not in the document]
jaiminpanchal27: 1st part related to `bidsReceived` discussed offline. 
Auction api improvement https://github.com/prebid/Prebid.js/pull/1644

src/auction.js
```diff
@@ -0,0 +1,518 @@
...
+ */
+function groupByPlacement(bidsByPlacement, bid) {
+  if (!bidsByPlacement[bid.adUnitCode]) { bidsByPlacement[bid.adUnitCode] = { bids: [] }; }
+  bidsByPlacement[bid.adUnitCode].bids.push(bid);
+  return bidsByPlacement;
+}
+
+export function createAuction({adUnits, adUnitCodes}) {
+  return newAuction({adUnits, adUnitCodes});
+}
```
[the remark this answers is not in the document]
jaiminpanchal27: Done here https://github.com/prebid/Prebid.js/pull/1644

Issue #1508, opened 2017-08-21
Remove $$PREBID_GLOBAL$$ from unit tests


dbemiller:
The $$PREBID_GLOBAL$$ references are causing a few problems in our test code.

First, they're fragile. #1437 was logged about this, and in the time it took to fix the existing references, several more PRs got merged which broke them again.

Second, they cause weird behavior. #1427 surfaced an issue where a test passed when using `config.setConfig`, but failed when using `$$PREBID_GLOBAL$$.setConfig`--even though they're the same function. I traced the root problem to the use of `window`. Currently, our webpack preprocessor is making a new bundle of Prebid for every unit test file... but the `window` object (and therefore $$PREBID_GLOBAL$$) persists across files. This is a very subtle bug which makes writing good unit tests tricky.

Lastly, they prevent us from [improving our unit test performance](https://github.com/webpack-contrib/karma-webpack#alternative-usage). Due to the `window` issue, many tests surrounding `$$PREBID_GLOBAL$$` fail if we _do_ change the way we bundle the code for testing. It's unfortunate, because the unit tests would only take ~15s to complete (rather than the ~45s they take currently). This difference will only increase as we add new unit test files.

We should refactor our code to be more testable, and remove all `$$PREBID_GLOBAL$$` references from the existing tests.

dbemiller:
Fixing this will be a marathon, not a sprint.

My proposal for working towards a solution is outlined in #1510. By encapsulating `$$PREBID_GLOBAL$$` state into smaller objects, and injecting those objects as dependencies throughout most of the code, we will gradually work towards a codebase which does not rely on `$$PREBID_GLOBAL$$` for testing.

PR #1606, opened 2017-09-21
convert targeting module to factory pattern

jaiminpanchal27:
## Type of change

## Description of change
Convert targeting module to factory pattern.

### DIFF
--- a/src/targeting.js
+++ b/src/targeting.js
@@ -5,66 +5,113 @@ const bidmanager = require('./bidmanager');
 const utils = require('./utils');
 var CONSTANTS = require('./constants');
 
-var targeting = exports;
 var pbTargetingKeys = [];
 
-targeting.resetPresetTargeting = function(adUnitCode) {
-  if (isGptPubadsDefined()) {
+export function newTargeting() {
+  let targeting = {};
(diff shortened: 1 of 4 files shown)


src/targeting.js
```diff
@@ -5,66 +5,113 @@ const bidmanager = require('./bidmanager');
 const utils = require('./utils');
 var CONSTANTS = require('./constants');
 
-var targeting = exports;
 var pbTargetingKeys = [];
 
-targeting.resetPresetTargeting = function(adUnitCode) {
-  if (isGptPubadsDefined()) {
+export function newTargeting() {
```
dbemiller: As written, all these instances are hardcoded to depend on two pieces of mutable/global state: `$$PREBID_GLOBAL$$` (through `getBidLandscapeTargeting` and `getAlwaysUseBidTargeting`) and the `bidmanager` (through `getStandardKeys`).

These two things should be arguments here, with the globals passed in in the `export const targeting` line. That will make it possible (for someone else) to clear the unit tests of the issues that @mkendall07 ran into back in #1427, and bring the project one step closer to #1508.

mkendall07:
Declining as we need to re-write targeting.

PR #1614, opened 2017-09-22
Update spotx video adapter to set the spotx_ad_key used in DFP

npeceniak:
## Type of change

- [X] Bugfix
- [ ] Feature
- [ ] New bidder adapter   
- [ ] Build related changes
- [ ] CI related changes
- [ ] Other

## Description of change

In order for DFP targeting to work correctly we need access to the cpm and spotx_ad_key. We were already passing the cpm value but with this update we are now setting the spotx_ad_key to the bid.adId key so that it can be included with the prebid standard targeting keys. 

### DIFF
--- a/modules/spotxBidAdapter.js
+++ b/modules/spotxBidAdapter.js
@@ -85,7 +85,7 @@ function Spotx() {
 
       bid.cpm = KVP_Object.spotx_bid;
       bid.vastUrl = url;
-      bid.ad = url;
+      bid.adId = KVP_Object.spotx_ad_key;
 
       var sizes = utils.isArray(bidReq.sizes[0]) ? bidReq.sizes[0] : bidReq.sizes;
       bid.height = sizes[1];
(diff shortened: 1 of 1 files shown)


modules/spotxBidAdapter.js
```diff
@@ -85,7 +85,7 @@ function Spotx() {
 
       bid.cpm = KVP_Object.spotx_bid;
       bid.vastUrl = url;
-      bid.ad = url;
+      bid.adId = KVP_Object.spotx_ad_key;
```
mkendall07: This will break some internals in prebid.js unfortunately. Can you set the key using this method: 
https://github.com/prebid/Prebid.js/blob/master/modules/xhbBidAdapter.js#L23
Basically set another property on bid and retrieve it using that function. 

The only downside of overriding this key is you can't reference the bid later using this key but I assume that's ok.

mkendall07:
@npeceniak needs a little work for merge.

npeceniak:
@mkendall07 Sorry for the delay in getting this updated. The adapter should be setting hb_adid the correct way now.

PR #1630, opened 2017-09-27
PulsePoint Lite adpater changes

anand-venkatraman:
## Type of change
- [x] Feature

## Description of change
Updates to the adapter to prepare for 1.0 upgrade. 
Support for InApp parameters.

**Test Page**: http://projects.contextweb.com/av/prebid/1.0-pre/  
- [x] avenkatraman@pulsepoint.com


modules/pulsepointLiteBidAdapter.js
```diff
@@ -1,301 +1,307 @@
...
-   * Produces an OpenRTBImpression from a slot config.
-   */
-  function impression(slot) {
+  supportedMediaTypes: ['native'],
+
+  isBidRequestValid: bid => (
+    !!(bid && bid.params && bid.params.cp && bid.params.ct)
+  ),
+
+  buildRequests: bidRequest => {
```
anand-venkatraman: Can you please confirm the first param of this method is the array of bid requests? It appears so based on the call made from [here](https://github.com/prebid/Prebid.js/blob/master/src/adapters/bidderFactory.js#L27). This contradicts with the information [here](http://prebid.org/dev-docs/bidder-adapter-1.html), where ```bidderRequest``` param is an object, with a property ```bids```, holding the bid requests.

bretg: Thanks for pointing this out Anand. Opened a PR (https://github.com/prebid/prebid.github.io/pull/380) on prebid.org to correct the docs. @snapwich will submit a PR to clarify the comments in the code.

Note that we also opted to rename the first parameter because there're actually two params formerly called "bidRequests" and "bidderRequest". This was kinda confusing, so we renamed the first to "validBidRequests". The second param isn't documented on prebid.org as it's expected to be needed only in rare cases, but is documented in the code.

anand-venkatraman: Thanks for clarifying @bretg

modules/pulsepointLiteBidAdapter.js
```diff
@@ -1,301 +1,308 @@
-import {createBid} from 'src/bidfactory';
-import {addBidResponse} from 'src/bidmanager';
+/* eslint dot-notation:0, quote-props:0 */
 import {logError, getTopWindowLocation} from 'src/utils';
-import {ajax} from 'src/ajax';
-import {STATUS} from 'src/constants';
-import adaptermanager from 'src/adaptermanager';
+import { registerBidder } from 'src/adapters/bidderFactory';
+
+const NATIVE_DEFAULTS = {
```
mkendall07: These could go into `native.js` if they are per the spec (i think they are). 

Doesn't need to change for merge though.

modules/pulsepointLiteBidAdapter.js
```diff
@@ -1,301 +1,308 @@
...
-      const size = slot.params.cf.toUpperCase().split('X');
-      const width = parseInt(slot.params.cw || size[0], 10);
-      const height = parseInt(slot.params.ch || size[1], 10);
-      return [width, height];
-    }
-    return [1, 1];
+/**
+ * Attempts to capture the referrer url.
+ */
+function referrer() {
```
mkendall07: this should also be a generic function (todo for utils).

Issue #1628, opened 2017-09-27
Standard bidCpmAdjustment not working


dugwood:
## Type of issue
`bidCpmAdjustment` set on `standard` doesn't work.

## Description
In the documentation: http://prebid.org/dev-docs/publisher-api-reference.html#module_pbjs.bidderSettings
=> bidCpmAdjustment is either for `standard` or `adapter-specific`

But:
https://github.com/prebid/Prebid.js/blob/master/src/bidmanager.js#L434
=> here we only look for the `bidderCode` (adapter-specific) function

And:
https://github.com/prebid/Prebid.js/blob/master/src/bidmanager.js#L453
=> here we never set the function if not defined

## Steps to reproduce
Set:
```
pbjs.bidderSettings = {
    standard: {
        bidCpmAdjustment: function (bidCpm, bid) { console.log(bidCpm); return bidCpm; }
    }
};
```

### Expected results
The console should log the bidCpm value

### Actual results
No log (so no adjustment)

## Platform details
Prebid: 0.29.0

mkendall07:
@dugwood thanks for the report. This has always been the case, the documentation is wrong. What's the use case to adjust all the bids?

dugwood:
Thanks @mkendall07. I should fix the documentation :-)

So far I was using it with a common function, that logs (in a debug mode) and adjust CPMs according to another external setting. For example:
```
grossToNet: {
		appnexusAst: 0.9, // value isn't the real one, just to illustrate
		rubicon: 0.95,
		indexExchange: 0.85
}
```

Now that all plaforms are sending the net revenue, it's indeed useless. So far I was looping on all bidders to set the `bidCpmAdjustment` (for every call, that was not that great), and just found out about the `standard` property. So I just tried to use it :-)

Although I could use the event, I still need to set the bidCpm to zero if it's under a predefined CPM, which is set on the platform (APN, ...) but also in our code because some platform don't have minimum CPM settings.

So a `standard` setting would be great. But I can do as before, setting it to the same function for every bidder (once and for all, not for each auction).

mkendall07:
I can see how this is more maintainable than setting individual bidder settings. We accept PRs :)

dugwood:
@mkendall07 it's done :-) Had a difficult time running tests (sooooo long!), but in the end I think it's better that the function is implemented than the documentation changed.

jaiminpanchal27:
Fixed in #1645.

PR #1645, opened 2017-09-30
Fix for #1628 (allowing standard bidCpmAdjustment)

dugwood:
## Type of change
- Bugfix
- Feature

## Description of change
This is a fix to cover the documentation misleading «standard» bid adjustment function. See #1628.
Changed 1 test (to show that when the function is declared at the standard level, it doesn't impact the one declared at the bidder level) and added another one to show it's been taken into account.

Also I've patched the `getStandardBidderSettings` which would destroy the `bidderSettings` whenever the `adserverTargeting` wasn't found. It passes the tests without modification.

### DIFF
--- a/src/bidmanager.js
+++ b/src/bidmanager.js
@@ -432,10 +432,10 @@ function adjustBids(bid) {
   var code = bid.bidderCode;
   var bidPriceAdjusted = bid.cpm;
   let bidCpmAdjustment;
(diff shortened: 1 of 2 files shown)


mkendall07:
nice! Will check it out later today

mkendall07:
@rmloveland possible that this needs a docs revision if there is inconsistent references.

Issue #1643, opened 2017-09-29
AppNexusAst adaptor no longer supports aliasing via API


ehoch:
## Type of issue
bug

## Description
Since the new 1.0 adaptor changes (a4d3ce7dfc6d0b19d1876e8060282f4eac4010cf), AppnexusAst no longer supports aliasing via the pbjs.aliasBidder API command. Using the new aliases: [] syntax in the registerBidder does work, however.

jaiminpanchal27:
Thanks @ehoch for reporting this. We will fix this asap.

gramorris:
This is the case with 0.29 also.

jaiminpanchal27:
This is fixed in https://github.com/prebid/Prebid.js/releases/tag/0.30.1

PR #1652, opened 2017-10-03
Aliasbidder fix

jaiminpanchal27:
## Type of change
- [x] Bugfix

## Description of change
Fix for #1643
Also, aliased adapter could only support banner mediaType. Fixed that too.


src/adaptermanager.js
```diff
@@ -218,14 +226,24 @@ exports.aliasBidAdapter = function (bidderCode, alias) {
...
+        let newAdapter;
+        let supportedMediaTypes = getSupportedMediaTypes(bidderCode);
+        // Have kept old code to support backward compatibilitiy.
+        // Remove this if loop when all adapters are supporting bidderFactory. i.e When Prebid.js is 1.0
+        if (bidAdaptor.constructor.prototype != Object.prototype) {
+          newAdapter = new bidAdaptor.constructor();
+          newAdapter.setBidderCode(alias);
+        } else {
+          let spec = Object.assign({}, bidAdaptor, { code: alias });
+          newAdapter = newBidder(spec);
```
dbemiller: This is a little buggy.

The `bidderFactory` accepts a `spec`, but only returns an object like:

```
{
  callBids: function() { ... }
}
```

So in this code, you're using objects of the shape:

```
newAdapter = newBidder({
  code: alias,
  callBids: function() { ... }
});
```

...but `newBIdder` needs a _spec_. The `typedef` _doesn't_ have `callBids`, but _does_ require several other things which you're not giving here (`interpretResponse`, `buildRequests`, etc).

So... if `callBids` gets called on an _aliased_ bidder, there'll be all sorts of errors about missing functions.

test/spec/unit/pbjs_api_spec.js
```diff
@@ -1306,6 +1319,22 @@ describe('Unit: Prebid Module', function () {
...
+      const alias = 'testalias';
+      $$PREBID_GLOBAL$$.aliasBidder(bidderCode, alias);
+      expect(adaptermanager.bidderRegistry).to.have.property(alias);
+    });
+
+    it('should add alias to registry when original adapter is using bidderFactory', function() {
+      const thisSpec = Object.assign(spec, { supportedMediaTypes: ['video'] });
+      registerBidder(thisSpec);
+      const alias = 'aliasBidder';
+      $$PREBID_GLOBAL$$.aliasBidder(CODE, alias);
```
dbemiller: Try to avoid `$$PREBID_GLOBAL$$` in tests (see #1508).

Prebid-specifics aside... strong unit tests cover the smallest amount of code possible. All your changes are in `adapterManager`... so it's much better if your tests import the `adapterManager` and call the functions directly. The `$$PREBID_GLOBAL$$.aliasBidder` function is at a higher level, and would be better tested by stubbing the `adapterManager` with `sinon` and making sure the methods get called with the right args. 

On a less important note... you could probably write these in `test/spec/unit/adapterManager_spec.js`. This file is much larger than it should be already.

test/spec/unit/core/adapterManager_spec.js
```diff
@@ -3,6 +3,9 @@ import AdapterManager from 'src/adaptermanager';
 import { getAdUnits } from 'test/fixtures/fixtures';
 import CONSTANTS from 'src/constants.json';
 import * as utils from 'src/utils';
+import { registerBidder } from 'src/adapters/bidderFactory';
+
+require('modules/appnexusAstBidAdapter');
```
dbemiller: Accidental import?

src/adapters/bidderFactory.js
```diff
@@ -138,6 +138,9 @@ export function registerBidder(spec) {
  */
 export function newBidder(spec) {
   return Object.assign(new Adapter(spec.code), {
+    getSpec: function() {
```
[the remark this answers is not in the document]
jaiminpanchal27: done.

test/spec/unit/core/adapterManager_spec.js
jaiminpanchal27: Not accidental, i took a shortcut to test old way of aliasing adapter. Since we will be only keeping this code for a short time. 

But anyway, updated test case to make it as it should be.

dbemiller: ah ok...  no worries either way.

PR #1655, opened 2017-10-03
prebidAdapter secure support

bretg:
## Type of change

- [ ] Bugfix
- [X] Feature
- [ ] New bidder adapter 
- [ ] Build related changes
- [ ] CI related changes
- [ ] Other

## Description of change
Passing the 'secure' flag from s2sConfig through to Prebid Server

## Other information

Could use a better unit test -- but that stands true of this module already... looks like there's not a good suite of checks that s2sConfig values are sent through to server. e.g. accountId, timeout_millis.

### DIFF
--- a/modules/prebidServerBidAdapter.js
+++ b/modules/prebidServerBidAdapter.js
@@ -111,6 +111,7 @@ function PrebidServer() {
       tid: bidRequest.tid,
       max_bids: config.maxBids,
       timeout_millis: config.timeout,
+      secure: config.secure,
       url: utils.getTopWindowUrl(),
       prebid_version: '$prebid.version$',
       ad_units: bidRequest.ad_units.filter(hasSizes),
--- a/test/spec/modules/prebidServerBidAdapter_spec.js
+++ b/test/spec/modules/prebidServerBidAdapter_spec.js
@@ -20,8 +20,9 @@ const REQUEST = {
   'tid': '437fbbf5-33f5-487a-8e16-a7112903cfe5',
   'max_bids': 1,
   'timeout_millis': 1000,
+  'secure': 0,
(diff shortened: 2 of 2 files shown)


mkendall07:
@bretg 
Prebid server already detects this, any reason to have it set here? https://github.com/prebid/prebid-server/blob/e0df370deef5aa4467619efbe553a0c53d34206f/pbs/pbsrequest.go#L275

bretg:
@mkendall07 - good to know it's at least partly covered. That certainly lowers the priority, but might still be a useful lever -- depends on the referrer that PBS receives. There are are two scenarios where the secure flag needs to be passed:

1) secure page, secure iframe
2) non-secure page, secure iframe - yes, we've seen this.

if the referrer is overridden to the page level (as it should be), then PBS doesn't know whether it's in scenario 2.

modules/prebidServerBidAdapter.js
```diff
@@ -111,6 +111,7 @@ function PrebidServer() {
       tid: bidRequest.tid,
       max_bids: config.maxBids,
       timeout_millis: config.timeout,
+      secure: config.secure,
```
mkendall07: think this should be `secure: getConfig('secure');`

mkendall07: add here as well https://github.com/prebid/Prebid.js/blob/master/src/prebid.js#L780

bretg: Actually I think the current PR is what we want -- if it's added to the object.assign in prebid.js as a default, then Prebid Server's URL check is always bypassed. The behavior of the code here is that when the s2sConfig contains a `secure` override, it's sent, otherwise PBS detects.

PR #1664, opened 2017-10-06
Update AdButler adapter for Prebid v1.0

dharton:
## Type of change
- [ ] Bugfix
- [ ] Feature
- [ ] New bidder adapter  
- [ ] Build related changes
- [ ] CI related changes
- [X] Other

## Description of change
Update AdButler adapter for Prebid v1.0

```
{
  bidder: 'adbutler',
  params: {
    accountID: 167283,
    zoneID: 210093
  }
}
```

- contact email of the adapter’s maintainer: dan@sparklit.com


test/spec/modules/adbutlerBidAdapter_spec.js
```diff
@@ -1,516 +1,274 @@
...
-            params: {
-              accountID: '107878',
-              zoneID: '86133',
-              domain: 'servedbyadbutler.com.dan.test'
-            },
-            requestId: '10b327aa396609',
-            placementCode: '/123456/header-bid-tag-1'
+import {expect} from 'chai';
+import adapterManager from 'src/adaptermanager';
+import bidManager from 'src/bidmanager';
```
dbemiller: We probably should have been more explicit about this in our docs, but... the `bidmanager` is going away in 1.0. The good news is that your tests can be simplified.

New your tests can just create objects with dummy data (bid request, server response, etc), pass them into the `spec` methods directly, and then make assertions on the return values. No need to worry about mocks, fake servers, or anything like that. The tests on the `bidderFactory` will make sure that your adapter's bids make it into the auction. 

Check out the [appnexusAst adapter](https://github.com/prebid/Prebid.js/blob/master/test/spec/modules/appnexusAstBidAdapter_spec.js) for some working examples..

modules/adbutlerBidAdapter.js
```diff
@@ -1,130 +1,122 @@
...
+import * as utils from 'src/utils';
+import {config} from 'src/config';
+import {registerBidder} from 'src/adapters/bidderFactory';
+
+const BIDDER_CODE = 'adbutler';
+
+export const spec = {
+  code: BIDDER_CODE,
+  pageID: Math.floor(Math.random() * 10e6),
+  zoneCounters: {},
```
dbemiller: No mutable globals allowed... see [the 1.0 conventions](http://prebid.org/dev-docs/bidder-adapter-1.html#required-adapter-conventions).

Can this be moved inside your `buildRequests` function?

modules/adbutlerBidAdapter.js
```diff
@@ -1,130 +1,122 @@
...
+      });
+      if (isCorrectCPM && isCorrectSize) {
+        bidResponse.requestId = bidObj.bidId;
+        bidResponse.bidderCode = spec.code;
+        bidResponse.creativeId = serverResponse.placement_id;
+        bidResponse.cpm = CPM;
+        bidResponse.width = width;
+        bidResponse.height = height;
+        bidResponse.ad = serverResponse.ad_code;
+        bidResponse.ad += spec.addTrackingPixels(serverResponse.tracking_pixels);
```
dbemiller: Adapters shouldn't add tracking pixels during the auction anymore. Implement [getUserSyncs](https://github.com/prebid/Prebid.js/blob/master/src/adapters/bidderFactory.js#L50) instead. `prebid-core` will make sure those pixels get added to the page alongside everyone else's

dbemiller:
Also: don't forget to add your `.md` file: http://prebid.org/dev-docs/bidder-adapter-1.html#planning-your-adapter

modules/adbutlerBidAdapter.js
dharton: Thanks for the info @dbemiller. We offer our customers the option to track impressions on the client-side. If they've chosen this, their impression pixel would have been added here. Just to verify, if the publisher has userSync disabled, this means there is a chance their impression wouldn't be counted in this way. Is that correct?

dbemiller: Correct. Since tracking pixels affect the publisher's page performance, the prebid leadership decided that they should have the ultimate decision in whether or not to allow them.

That said, most publishers leave the default options... so it's not likely to impact you too much.

dharton: I understand the desire to improve publisher page performance. My only concern is that when all syncs are blocked, the winning bidder will not have any way to know that they have won a particular auction. Is there any way for us to provide Prebid with a URL that can be pinged or placed if our bid has won the auction?

mkendall07: I think there might be some confusion here. It's acceptable to add a tracking pixel to a bid response. This will only fire if the bid wins the auction. I think that's what this is so should be fine.

dharton: Ah, thanks for the clarification. Yeah, this function just appends the tracking pixel(s) to the ad code in the bid response.

dharton:
Thanks for your help! I've committed the recommended changes.

dbemiller:
@mkendall07 to the rescue!... sorry about that.

dbemiller:
This looks good... the only thing preventing merge from my POV is the failing tests.

I expected this to be one of those intermittent failures, since it's the c1x adapter tests... but it doesn't look like it is. The tests pass in master, and fail reliably in this branch. Looks like they're doing all sorts of things with `window` and `$$PREBID_GLOBAL$$`, so I wouldn't be surprised if there was some weird interdependence on other peoples' tests in there somewhere.

dharton:
Hmm. It's strange that my changes are causing an issue since, with the new Prebid requirements, they don't use either window or $$PREBID_GLOBAL$$.

dharton:
Hey @dbemiller , I just wanted to give you a bit of an update. I've figured out why the c1x test are failing. In their adapter they are referencing `$$PREBID_GLOBAL$$.adUnits`. Somewhere before the module tests, the adUnits get filled up with a lot of bidders. Then (in my test file) it gets filled in with just one bidder. This allows the c1x test to pass the final test case. Now that my new tests don't make use of that those ad units, the c1x test is failing on that larger amount of bidders. If you'd like, I could send along some screenshots to show you what I mean.

Do you think you'd be able to merge my changes in, or will we have to wait for them to update?

dbemiller:
@dharton we definitely can't merge something with broken tests. It would break the CI, so then the tests would fail on everyone else's PRs.

If it's just their tests which are broken, I'd recommend fixing them here. We haven't heard back from them about prebid 1.0 updates yet (to my knowledge)... so they might not ever come back to update & fix them. 

This is always a risk of unit tests with side effects. Your new code looks like it's much safer.

dbemiller:
One other issue I just noticed... this PR introduces an empty file: `test/spec/adapters/adbutler_spec.js`. Please delete it before we merge.

dbemiller:
looks great... nice code!

Issue #1679, opened 2017-10-11
Prebid 1.0 video bid validation throws error


jsalis:
## Type of issue
Bug

## Description
Validation for video bids throws an error.

![image](https://user-images.githubusercontent.com/5342031/31460033-03f7804a-ae93-11e7-962c-5d66d15f0b84.png)

## Steps to reproduce

Add the Beachfront bid adapter to the video example page. (I think any valid video bid will reproduce it)
Run the auction.

```javascript
var videoAdUnit = {
    code: 'video1',
    sizes: [640, 480],
    mediaType: 'video',
    bids: [{
        bidder: 'appnexusAst',
        params: {
            placementId: '9333431',
            video: {
                skipppable: false,
                playback_methods: ['auto_play_sound_off']
            }
        }
    }, {
        bidder: 'beachfront',
        params: {
            bidfloor: 0.01,
            appId: '206eca8c-f7c5-46fd-c848-b8f041214e77'
        }
    }]
};
```

## Other information
#1680

jaiminpanchal27:
Fixed and merged in #1680

PR #1680, opened 2017-10-11
Prebid 1.0 Fix issue with video bid validation

jsalis:
## Type of change
- [x] Bugfix

## Description of change
Fixes issue with video bid validation

## Other information
#1679


test/spec/video_spec.js
```diff
@@ -1,48 +1,67 @@
...
   });
 
   it('validates valid instream bids', () => {
-    sinon.stub(utils, 'getBidRequest', () => ({
-      bidder: 'appnexusAst',
-      mediaTypes: {
-        video: { context: 'instream' },
-      },
-    }));
+	  sinon.stub(auctionManager, 'getBidsRequested', () => [
```
jaiminpanchal27: Add these stubs in `beforeEach`.  And use `stub.returns` method in each test case. 
This will prevent error `sinon stub already wrapped` if any test case fails and reduce lines of code too.

jsalis:
No problem. I updated the test cases.

jsalis:
I think that would be cleaner. In that case, `nativeBidIsValid` should also be changed.

dbemiller:
Yes... I agree ^^

PR #1686, opened 2017-10-12
PulsePoint Lite adapter - Enabling Sync pixel

anand-venkatraman:
## Type of change
- [x] Feature

## Description of change
Enabling a Image pixel for user sync.

- avenkatraman@pulsepoint.com


modules/pulsepointLiteBidAdapter.js
```diff
@@ -53,6 +53,11 @@ export const spec = {
         type: 'iframe',
         url: '//bh.contextweb.com/visitormatch'
       }];
+    } else if (syncOptions.pixelEnabled) {
```
dbemiller: Small suggestion... image syncs impact page performance less than iframe ones, and publishers are more likely to allow them. Might be worth flipping the order on these so that the image one comes first.

PR #1687, opened 2017-10-13
Location protocol check - SSL alert

AntoineJac:
Edit the url location in case the page is secure to avoid ssl alert

## Type of change

- [X ] Bugfix
- [ ] Feature
- [ ] New bidder adapter   
- [ ] Build related changes
- [ ] CI related changes
- [ ] Other

## Description of change

Edit the url location in case the page is secure to avoid ssl alert

```
{
  bidder: '<bidder name>',
  params: {
    // ...
  }
}
```

### DIFF
--- a/modules/currency.js
+++ b/modules/currency.js
@@ -93,7 +93,7 @@ function initCurrency(url) {
     if (useSSL) {
       url = url.replace('http://', 'https://');
     }
-    
+
     ajax(url, function (response) {
       try {
         currencyRates = JSON.parse(response);
--- a/modules/currency.js
+++ b/modules/currency.js
@@ -89,6 +89,9 @@ function initCurrency(url) {
   }
 
   if (!currencyRates.conversions) {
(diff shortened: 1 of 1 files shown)


bretg:
Thanks @AntoineJac, but we can't currently support HTTPS for the currency file.

As noted in the docs, if a pub is running secure pages, they need to host a conversion file themselves for now.

At some point prebid.org will be able to support the secure currency file, and at that point we'll accept this PR. Thanks.

snapwich:
If an endpoint supports https then they should just use the https url as it will work w/ http or https.  I don't like the idea of it programmatically changing the url in this case as it could cause errors for people who add a custom URL that doesn't support https.

Also, as Bret said, the prebid endpoint doesn't currently support https but we plan to in the future.  When the support is available we will just update this line to always use https: https://github.com/AntoineJac/Prebid.js/blob/0b3fe8f81fca6074716cbe4fefff1d3a242515a3/modules/currency.js#L8

Issue #1603, opened 2017-09-19
Immediate adapter response may end auction


github-nilsson:
## Type of issue
Robustness problem with bid response API.

## Description
Immediate calls to `bidmanager.addBidResponse` may cause the auction to end. Since the same loop is used to call adapters and add requests to `pbjs._bidsRequested`, an immediate call to `bidmanager.addBidResponse` can cause the bidsBackAll to believe the auction has ended.

## Steps to reproduce
Have the first adapter respond to all bid request by immediately calling `bidmanager.addBidResponse`. The auction will immediately end and any responses from other adapters will be ignored.

## Solution
This can of course be fixed by asynchronously calling  `bidmanager.addBidResponse` using `setTimer`, but I would like to avoid these kind of API traps. I propose to have the current loop only update `pbjs._bidsRequested` and then iterate over that array for calling the adapters like the following.

```Javascript
  $$PREBID_GLOBAL$$._bidsRequested.forEach(bidderRequest => {
    utils.logMessage(`CALLING BIDDER ======= ${bidderRequest.bidderCode}`);
    events.emit(CONSTANTS.EVENTS.BID_REQUESTED, bidderRequest);
    _bidderRegistry[bidderRequest.bidderCode].callBids(bidderRequest);
  });
```

mkendall07:
@github-nilsson 
is there any adapters doing this today? Each adapter would be making a async HTTP call, so  `addBidResponse` would be async as well. Is this just a theoretical?

github-nilsson:
The adapter we are developing can return a response right away in some cases, so this caused a few days of mysterious lowered revenue.

mkendall07:
@jaiminpanchal27 
please check this condition in the 1.0 branch.

dugwood:
Just hit the same issue in 0.30.1. The first result forbids the next auction to be sent:

```
 Time Info
   64 AdUnit ready (579px): [[300,250],[300,600],[336,280]]
   64 Launching requests
  253 Request bid for rubicon
  254 rubicon (0x0):           0       x 1     = 0       (CPM: -   , Time: 1)
  256 All bids are back, launching DFP with price bucket 0.20: 3ms
  261 Request bid for appnexusAst
  424 appnexusAst (300x250):   0.436   x 1     = 0.436   (CPM: 0.2 , Time: 163)
  503 DFP wins with price bucket 0.20
```

[...]

PR #1690, opened 2017-10-13
Fixes: Immediate adapter response may end auction

jaiminpanchal27:
## Type of change
- [x] Bugfix

## Description of change
Fixes https://github.com/prebid/Prebid.js/issues/1603


github-nilsson:
This doesn't look 100% correct. The first loop will only add a bidderRequest if the adapter exists and has bids. The second loop makes the same tests, which will always be true, so the error message will never be logged.

At least move the logError to the first loop, and possibly remove the checks from the second loop.

PR #1691, opened 2017-10-13
Support native click tracking

matthewlane:
## Type of change
- Feature

## Description of change
As per [OpenRTB Native 1.2 ](https://www.iab.com/wp-content/uploads/2017/04/OpenRTB-Native-Ads-Specification-Draft_1.2_2017-04.pdf), click trackers should be supported as an array of URLs. This PR adds support to Prebid for click trackers associated with native bid responses.

If a native bid response contains click trackers, bidder adapters should attach the click tracker array to the `native.clickTrackers` property of the bid response object (see appnexusAstBidAdapter for an example of this).

As native creatives that are served from adservers are likely to be in a cross-origin frame, it is necessary to use `postMessage` from the native creative template to communicate to Prebid that a link was clicked. Prebid is (even before this PR) configured to fire native impression trackers when it receives a message with the following structure:

```JavaScript
<script>
window.parent.postMessage(JSON.stringify({
  message: 'Prebid Native',
  adId: '%%PATTERN:hb_adid%%'
}), '*');
</script>
```

This PR introduces the ability to fire click trackers for the associated `adId` when a `'click'` `action` is added to the message:

```JavaScript
<script>
window.parent.postMessage(JSON.stringify({
  message: 'Prebid Native',
  action: 'click',
  adId: '%%PATTERN:hb_adid%%'
}), '*');
</script>
```

Ad ops teams may use these structures of `postMessage` in their native creative templates in any way that fits their use case. One example usage is given below:

```HTML
<div class="sponsored-post">
  <div class="content">
      <h1>
        <a
          href="%%CLICK_URL_UNESC%%%%PATTERN:hb_native_linkurl%%"
          target="_blank"
          onclick="fireTrackers('click')"
        >
          %%PATTERN:hb_native_title%%
        </a>
      </h1>
      <p>%%PATTERN:hb_native_body%%</p>
      <div class="attribution">%%PATTERN:hb_native_brand%%</div>
  </div>
</div>

<script>
function fireTrackers(action) {
  var message = {message: 'Prebid Native', adId: '%%PATTERN:hb_adid%%'};
  if (action === 'click') {message.action = 'click';}  // fires click trackers when called via link
  window.parent.postMessage(JSON.stringify(message), '*');
}

fireTrackers(); // fires native impressions on creative load
</script>
```

## Other information
Click through/landing page URLs are a required field in the OpenRTB 1.2 spec. This PR also requires native bid responses to have a landing page url, defined by `native.clickUrl` on the native bid response object.

### DIFF
--- a/modules/appnexusAstBidAdapter.js
+++ b/modules/appnexusAstBidAdapter.js
@@ -201,6 +201,7 @@ function newBid(serverBid, rtbBid) {
       image: nativeAd.main_img && nativeAd.main_img.url,
       icon: nativeAd.icon && nativeAd.icon.url,
       clickUrl: nativeAd.link.url,
+      clickTrackers: nativeAd.link.click_trackers,
       impressionTrackers: nativeAd.impression_trackers,
     };
   } else {
--- a/src/native.js
+++ b/src/native.js
@@ -94,14 +94,60 @@ export function nativeBidIsValid(bid) {
 }
 
 /*
- * Native responses may have impression trackers. This retrieves the
- * impression tracker urls for the given ad object and fires them.
+ * Native responses may have associated impression or click trackers.
+ * This retrieves the appropriate tracker urls for the given ad object and
+ * fires them. As a native creatives may be in a cross-origin frame, it may be
+ * necessary to invoke this function via postMessage. secureCreatives is
+ * configured to fire this function when it receives a `message` of 'Prebid Native'
+ * and an `adId` with the value of the `bid.adId`. When a message is posted with
+ * these parameters, impression trackers are fired. To fire click trackers, the
+ * message should contain an `action` set to 'click'.
+ *
+ * // Native creative template example usage
+ * <a href="%%CLICK_URL_UNESC%%%%PATTERN:hb_native_linkurl%%"
+ *    target="_blank"
+ *    onclick="fireTrackers('click')">
+ *    %%PATTERN:hb_native_title%%
+ * </a>
+ *
+ * <script>
+ *   function fireTrackers(action) {
+ *     var message = {message: 'Prebid Native', adId: '%%PATTERN:hb_adid%%'};
+ *     if (action === 'click') {message.action = 'click';} // fires click trackers
+ *     window.parent.postMessage(JSON.stringify(message), '*');
+ *   }
+ *   fireTrackers(); // fires impressions when creative is loaded
+ * </script>
  */
-export function fireNativeImpressions(adObject) {
-  const impressionTrackers =
-    adObject['native'] && adObject['native'].impressionTrackers;
+export function fireNativeTrackers(message, adObject) {
+  let trackers;
 
-  (impressionTrackers || []).forEach(tracker => {
-    triggerPixel(tracker);
+  if (message.action === 'click') {
+    trackers = adObject['native'] && adObject['native'].clickTrackers;
+  } else {
+    trackers = adObject['native'] && adObject['native'].impressionTrackers;
+  }
+
+  (trackers || []).forEach(triggerPixel);
+
+  return trackers;
+}
+
+/**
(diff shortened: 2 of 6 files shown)


src/native.js
```diff
@@ -78,6 +78,11 @@ export function nativeBidIsValid(bid) {
     return false;
   }
 
+  // all native bid responses must defined a landing page url
```
dbemiller: typo: `define`

src/native.js
```diff
@@ -94,14 +99,60 @@ export function nativeBidIsValid(bid) {
...
+    trackers = adObject['native'] && adObject['native'].clickTrackers;
+  } else {
+    trackers = adObject['native'] && adObject['native'].impressionTrackers;
+  }
+
+  (trackers || []).forEach(triggerPixel);
 
-  (impressionTrackers || []).forEach(tracker => {
-    triggerPixel(tracker);
+  return trackers;
```
dbemiller: Not sure I'd build tests around this return value... what you really want is to make sure that `triggerPixel` gets called with the expected args. Since this is a separate code path, there's no guarantee that it stays in sync with `triggerPixel` calls. Code may get added/removed over time.

I know it sucks for testing, but... it seems more robust to set a stub for `triggerPixel` and make sure that it gets called with the expected args.

src/native.js
```diff
@@ -94,14 +99,60 @@ export function nativeBidIsValid(bid) {
...
+}
+
+/**
+ * Sets native targeting key-value paris
+ * @param {Object} bid
+ * @return {Object} targeting
+ */
+export function setNativeTargeting(bid) {
```
dbemiller: Should probably rename to `getNativeTargeting`, `parseNativeTargeting`, or something like that.

`set` pretty universally means "this function changes some state." In this case, you're building a transformed copy of some existing state without changing it.

src/native.js
```diff
@@ -94,14 +99,58 @@ export function nativeBidIsValid(bid) {
...
+
+  Object.keys(bid['native']).forEach(asset => {
+    const key = NATIVE_KEYS[asset];
+    const value = bid['native'][asset];
+    if (key) {
+      keyValues[key] = value;
+    }
   });
+
+  return keyValues;
```
dbemiller: No need to return this anymore

matthewlane: Eh?

dbemiller: Oh I'm sorry... had a brain fart and thought this was the `return trackers` line. Nevermind.

modules/appnexusAstBidAdapter.js
```diff
@@ -201,6 +201,7 @@ function newBid(serverBid, rtbBid) {
       image: nativeAd.main_img && nativeAd.main_img.url,
       icon: nativeAd.icon && nativeAd.icon.url,
       clickUrl: nativeAd.link.url,
+      clickTrackers: nativeAd.link.click_trackers,
```
mkendall07: I wonder if we should be validating the different bid types and objects... food for later thought.

rmloveland:
@matthewlane based on your comment, looks like we need to update our [Ad Ops docs for Native](http://prebid.org/adops/setting-up-prebid-native-in-dfp.html) to show how to use these click trackers.

Are there any changes needed to the [developer setup docs for Native](http://prebid.org/dev-docs/show-native-ads.html)?

matthewlane:
@rmloveland Yeah, updating the Ad Opcs docs to show how to use click trackers is a good idea. On the developer setup docs, the [native object section](http://prebid.org/dev-docs/show-native-ads.html#native-ad-keys) may be updated to indicate that `clickTrackers` can now be required. If they are not present on the bid response, this will cause an error (like other native required-but-missing assets). However the pub doesn't have to required these, and if they are present on the native bid response they will still fire

PR #1692, opened 2017-10-15
small fix - allow more than one placement from one page

FeatureForwardDev:


jaiminpanchal27:
@FeatureForwardDev Why are you removing trion adapter ? Are you authorized to remove the same ?

PR #1693, opened 2017-10-16
Update AOL adapter for v1.0 

vzhukovsky:
## Type of change
- [x] Other

## Description of change
AOL adapter changes for Prebid 1.0.

### DIFF
--- a/modules/aolBidAdapter.js
+++ b/modules/aolBidAdapter.js
@@ -1,5 +1,6 @@
 import * as utils from 'src/utils';
 import { registerBidder } from 'src/adapters/bidderFactory';
+import { config } from 'src/config';
 import constants from 'src/constants.json';
 
 const AOL_BIDDERS_CODES = {
@@ -195,7 +196,7 @@ function _parseBid(response, bid) {
 
   let ad = bidData.adm;
   if (response.ext && response.ext.pixels) {
-    if (bid.userSyncOn !== constants.EVENTS.BID_RESPONSE) {
+    if (config.getConfig('aol.userSyncOn') !== constants.EVENTS.BID_RESPONSE) {
       let formattedPixels = response.ext.pixels.replace(/<\/?script( type=('|")text\/javascript('|")|)?>/g, '');
 
       ad += '<script>if(!parent.$$PREBID_GLOBAL$$.aolGlobals.pixelsDropped){' +
@@ -325,10 +326,10 @@ export const spec = {
     });
   },
   interpretResponse: interpretResponse,
-  getUserSyncs: function(options, bidResponses, bidRequest) {
+  getUserSyncs: function(options, bidResponses) {
     let bidResponse = bidResponses[0];
 
-    if (bidResponse && bidRequest && bidRequest.userSyncOn === constants.EVENTS.BID_RESPONSE) {
+    if (config.getConfig('aol.userSyncOn') === constants.EVENTS.BID_RESPONSE) {
       if (!$$PREBID_GLOBAL$$.aolGlobals.pixelsDropped && bidResponse.ext && bidResponse.ext.pixels) {
         $$PREBID_GLOBAL$$.aolGlobals.pixelsDropped = true;
 
--- a/modules/aolBidAdapter.js
+++ b/modules/aolBidAdapter.js
@@ -39,6 +39,7 @@ const MP_SERVER_MAP = {
   as: 'adserver-as.adtech.advertising.com'
 };
 const NEXAGE_SERVER = 'hb.nexage.com';
+const BID_RESPONSE_TTL = 300;
 
 $$PREBID_GLOBAL$$.aolGlobals = {
   pixelsDropped: false
@@ -216,7 +217,8 @@ function _parseBidResponse(response, bidRequest) {
     pubapiId: response.id,
     currency: response.cur,
     dealId: bidData.dealid,
-    netRevenue: true
+    netRevenue: true,
+    ttl: BID_RESPONSE_TTL
   };
 }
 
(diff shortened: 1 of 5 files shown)


vzhukovsky:
Hi everyone.

I reviewed new adapter requirements and found only one collision with this PR. It is related to user syncs. In 0.X.X our adapter has following features:
1) Mode for dropping pixels (user syncs) during ad rendering (the default behaviour);
2) Mode for dropping pixels right after getting bid response back;
3) Also we drop pixels only once for both mentioned modes;

#2 and #3 are implemented via new Prebid userSyncs module. But I do not see any legal way to implement #1 because it requires globals and some changes in core modules.

We would really like to keep all these features in 1.0. Is it possible?

vzhukovsky:
@mjacobsonny @jaiminpanchal27 @ndhimehta Hi everyone. 
Is there any chance to review the changes?

modules/aolBidAdapter.js
```diff
@@ -1,341 +1,341 @@
...
-          options.customHeaders = {
-            'x-openrtb-version': '2.2'
-          };
-          options.method = 'POST';
-          options.contentType = 'application/json';
-        }
-      } else if (isMarketplaceRequest) {
-        apiUrl = _buildMarketplaceUrl(bid);
-      }
+    if (bidResponse && bidRequest && bidRequest.userSyncOn === constants.EVENTS.BID_RESPONSE) {
```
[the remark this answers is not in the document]
snapwich: It'd probably be best to use the [userSync API](http://prebid.org/dev-docs/publisher-api-reference.html#module_pbjs.userSync) to check if user syncs should fire.  The config is passed as the first paramteter to `getUserSyncs` and can be used like this: https://github.com/prebid/Prebid.js/blob/master/modules/appnexusAstBidAdapter.js#L115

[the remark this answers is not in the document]
vzhukovsky: Yes, we use it just for fetching the param. 

It looks like we can create new namespace via new setConfig API and define it here.
`setConfig({aol: { userSyncOn: 'bidRequest'}}});`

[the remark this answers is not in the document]
snapwich: Yes you could.  Is there a reason you want to do your own user sync flag as opposed to the general
```
pbjs.setConfig({ userSync: {
    syncEnabled: false
}});
```
and checking the syncConfig that's passed into `getUserSyncs`?

modules/aolBidAdapter.js
```diff
@@ -1,341 +1,346 @@
...
-            $$PREBID_GLOBAL$$.bidderSettings && $$PREBID_GLOBAL$$.bidderSettings.aol &&
-            typeof $$PREBID_GLOBAL$$.bidderSettings.aol.bidCpmAdjustment === 'function'
-          ) {
-            utils.logWarn(
-              'bidCpmAdjustment is active for the AOL adapter. ' +
-              'As of Prebid 0.14, AOL can bid in net – please contact your accounts team to enable.'
-            );
-          }
-          showCpmAdjustmentWarning = false; // warning is shown at most once
+  bidRequest.bidderCode = bid.bidder;
```
jaiminpanchal27: Any reason to add these param to `bidRequest`. I don't see its use.

modules/aolBidAdapter.js
```diff
@@ -1,341 +1,346 @@
...
+}
 
-          try {
-            response = JSON.parse(response);
-          } catch (e) {
-            utils.logError('Invalid JSON in bid response', AOL_BIDDERS_CODES.aol, bid);
-            _addErrorBidResponse(bid, response);
-            return;
-          }
+function interpretResponse(bidResponse, bidRequest) {
```
jaiminpanchal27: #1748 changed the first argument of `interpretResponse` to:

```{
  body: responseBody,
  headers: {
  }
}```

File: modules/aolBidAdapter.js [hunk 4]
```diff
@@ -1,341 +1,346 @@
...
     }
-
-    bidmanager.addBidResponse(bid.placementCode, bidResponse);
   }
 
-  function _isMarketplaceBidder(bidder) {
-    return bidder === AOL_BIDDERS_CODES.aol || bidder === AOL_BIDDERS_CODES.onedisplay;
-  }
+  return {
+    bidderCode: bid.bidderCode,
```
jaiminpanchal27: Some bidResponse params are missing and invalid. Please check http://prebid.org/dev-docs/bidder-adapter-1.html

modules/aolBidAdapter.js
vzhukovsky: Hi. We have an analytic adapter that is not part of Prebid official yet. Having "bidder Code" here allows us to break out the reporting between different aliases.

vzhukovsky:
Please use the following setting for validating bids:

bidder: 'aol',
params: {
placement: 3611253,
network: '9599.1'
}

vzhukovsky:
@jaiminpanchal27 Hi. Comments are fixed and now you can validate the bids.

jaiminpanchal27:
@vzhukovsky You missed adding required param `ttl`. I was able to validate bids by adding that param. Please update accordingly.

Also can you update `.md` file with these new params for anyone to test.

vzhukovsky:
@jaiminpanchal27 Thank you for reviewing it. I added `ttl` param and improved `md` file.

modules/aolBidAdapter.js
```diff
@@ -1,341 +1,349 @@
...
     }
-
-    bidmanager.addBidResponse(bid.placementCode, bidResponse);
   }
 
-  function _isMarketplaceBidder(bidder) {
-    return bidder === AOL_BIDDERS_CODES.aol || bidder === AOL_BIDDERS_CODES.onedisplay;
-  }
+  return {
+    bidderCode: bidRequest.bidderCode,
```
mkendall07: Please drop bidder code.

PR #1702, opened 2017-10-17
Add ad units event

hhhjort:
## Type of change

- [X] Feature

## Description of change

Simple additional event for addAdUnits to support the use case where the adunit configuration code is running separately from the ad display logic. There is a potential race condition where you could try to request an auction for an ad unit before the ad units have been loaded in, resulting in no actual auctions being requested. With this event, you can ensure you don't try to run an auction before you have defined your ad units.

```
{
  bidder: '<bidder name>',
  params: {
    // ...
  }
}
```


mkendall07:
Unit test can be added later

PR #1712, opened 2017-10-17
Trion Interactive 1.0 adapter update

grohm33:
…rebid version(1.0).

## Type of change

- [ ] Bugfix
- [ ] Feature
- [ ] New bidder adapter   
- [ ] Build related changes
- [ ] CI related changes
- [x] Other

## Description of change

Adding updated adapter for the 1.0 release

mgroh@trioninteractive.com

## Other information

We have a test ad set up to return bids from pubId : '30222' and sectionId : '5' and will only do so on mobile user agent.


modules/trionBidAdapter.js
```diff
@@ -1,129 +1,153 @@
...
-      }, true);
-    } else {
-      userTag = window.TRION_INT;
-      getBids(bids);
-    }
-  };
+  interpretResponse: function (trionResponseObj, request) {
+    var bid = {};
+    var bidResponses = [];
+    var bidRequest = request.bidRequest;
```
snapwich: Where is this coming from?  I think you're doing something similar to what I did in the rubicon adapter to get the original bidRequest, but I attached it to the request;  I don't see you doing the same in this code.

grohm33: Hey @snapwich, the bidrequest is included in the build request function on line 25 of the adapter.  I'm sending it up with the bid request object for each bid.  The request is coming through on all the tests on my end.  Let me know if there is something I'm missing.

modules/trionBidAdapter.js
```diff
@@ -1,129 +1,153 @@
...
+      var result = trionResponseObj.result;
+
+      if (result && result.cpm && result.placeBid && result.ad) {
+        var cpm = parseInt(result.cpm, 10) / 100;
 
-  function getBids(bids) {
-    if (!userTag.int_t) {
-      userTag.int_t = window.TR_INT_T || -1;
+        bid.requestId = bidRequest.bidId;
+        bid.bidderCode = bidRequest.bidder;
```
mkendall07: This line needs to be dropped. Sorry for the confusion.

mkendall07:
@grohm33 One thing to fix and then this can merge. Sorry for the delay!

grohm33:
@mkendall07 the bidder code was removed.  Thanks!

PR #1721, opened 2017-10-18
Update GetIntent adapter to 1.0 version

dhargo:
## Type of change

- [x] Other

## Description of change

GetIntent adapter for **Prebid.js 1.0** version

```
{
  bidder: 'getintent',
  params: {
    pid: '7',
    tid: 'test01',
    // ...
  }
}
```

server-dev@getintent.com

### DIFF
--- a/modules/getintentBidAdapter.js
+++ b/modules/getintentBidAdapter.js
@@ -1,78 +1,130 @@
-import { STATUS } from 'src/constants';
-import adaptermanager from 'src/adaptermanager';
+import { registerBidder } from 'src/adapters/bidderFactory';
 
-var bidfactory = require('src/bidfactory.js');
-var bidmanager = require('src/bidmanager.js');
-var adloader = require('src/adloader.js');
+const BIDDER_CODE = 'getintent';
+const BID_HOST = 'px.adhigh.net';
+const BID_BANNER_PATH = '/rtb/direct_banner';
+const BID_VIDEO_PATH = '/rtb/direct_vast';
+const VIDEO_PROPERTIES = [
+  'protocols', 'mimes', 'min_dur', 'max_dur', 'min_btr', 'max_btr', 'vi_format', 'api', 'skippable'
+];
+const OPTIONAL_PROPERTIES = [
+  'cur', 'floor'
+];
 
-var GetIntentAdapter = function GetIntentAdapter() {
-  var headerBiddingStaticJS = window.location.protocol + '//cdn.adhigh.net/adserver/hb.js';
+export const spec = {
+  code: BIDDER_CODE,
+  aliases: ['getintentAdapter'],
+  supportedMediaTypes: ['video', 'banner'],
 
-  function _callBids(params) {
-    if (typeof window.gi_hb === 'undefined') {
-      adloader.loadScript(headerBiddingStaticJS, function() {
-        bid(params);
-      }, true);
-    } else {
-      bid(params);
+  /**
+   * Determines whether or not the given bid request is valid.
+   *
+   * @param {BidRequest} bid The bid to validate.
+   * @return {boolean} True if this is a valid bid, and false otherwise.
+   * */
+  isBidRequestValid: function(bid) {
+    return !!(bid && bid.params && bid.params.pid && bid.params.tid);
+  },
+
+  /**
+   * Make a server request from the list of BidRequests.
+   *
+   * @param {BidRequest[]} bidRequests - an array of bids.
+   * @return ServerRequest[]
+   */
+  buildRequests: function(bidRequests) {
+    return bidRequests.map(bidRequest => {
+      let giBidRequest = buildGiBidRequest(bidRequest);
+      return {
+        method: 'GET',
+        url: buildUrl(giBidRequest),
+        data: giBidRequest,
+      };
+    });
+  },
+
+  /**
+   * Callback for bids, after the call to DSP completes.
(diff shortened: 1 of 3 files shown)


modules/getintentBidAdapter.js
```diff
@@ -1,78 +1,130 @@
...
+
+  /**
+   * Callback for bids, after the call to DSP completes.
+   * Parse the response from the server into a list of bids.
+   *
+   * @param {object} serverResponse A response from the server.
+   * @param {BidRequest} bidRequest
+   * @return {Bid[]} An array of bids which were nested inside the server.
+   */
+  interpretResponse: function(serverResponse, bidRequest) {
```
matthewlane: #1748 changed the first argument of `interpretResponse` to:

```JavaScript
{
  body: responseBody,
  headers: {
  }
}
```

so doing something like

```JavaScript
serverResponse = serverResponse.body
```

(or however you'd prefer) in this function should give you the expected behavior

modules/getintentBidAdapter.js
```diff
@@ -1,78 +1,130 @@
...
+    const bids = [];
+    if (serverResponse && serverResponse.no_bid !== 1) {
+      let size = parseSize(serverResponse.size);
+      let bid = {
+        requestId: bidRequest.bidId,
+        bidderCode: spec.code,
```
matthewlane: `bidderCode` can be dropped from the bid response object, it'll be set automatically by `bidderFactory`. Docs just updated, sorry for the confusion

modules/getintentBidAdapter.js
```diff
@@ -1,78 +1,130 @@
...
+  let giBidRequest = {
+    pid: bidRequest.params.pid, // required
+    tid: bidRequest.params.tid, // required
+    known: bidRequest.params.known || 1,
+    is_video: bidRequest.mediaType === 'video',
+    resp_type: 'JSON'
+  };
+  if (bidRequest.sizes) {
+    // TODO: add support for multiple sizes
+    giBidRequest.size = bidRequest.sizes[0].join('x');
```
matthewlane: Adunit sizes may be defined as `[300, 250]`, in which case `sizes[0]` would be `300` and cause a `TypeError`. Be sure to check that `bidRequest.sizes[0]` is an array before calling `join`

modules/getintentBidAdapter.js
```diff
@@ -1,78 +1,130 @@
...
+        bidderCode: spec.code,
+        cpm: serverResponse.cpm,
+        width: size[0],
+        height: size[1],
+        mediaType: bidRequest.mediaType || 'banner'
+      };
+      if (bidRequest.creative_id) bid.creativeId = bidRequest.creative_id;
+      if (bidRequest.mediaType === 'video') {
+        bid.vastUrl = serverResponse.vast_url;
+        bid.descriptionUrl = serverResponse.vast_url;
```
matthewlane: `bid.descriptionUrl` is no longer required, `bid.vastUrl` alone will work now

modules/getintentBidAdapter.js
```diff
@@ -1,78 +1,130 @@
...
+   *
+   * @param {object} serverResponse A response from the server.
+   * @param {BidRequest} bidRequest
+   * @return {Bid[]} An array of bids which were nested inside the server.
+   */
+  interpretResponse: function(serverResponse, bidRequest) {
+    const bids = [];
+    if (serverResponse && serverResponse.no_bid !== 1) {
+      let size = parseSize(serverResponse.size);
+      let bid = {
```
matthewlane: A few of the required bid response fields are missing: `ttl`, `netRevenue`, and `currency`. See the table in this section http://prebid.org/dev-docs/bidder-adapter-1.html#interpreting-the-response for descriptions and examples of each of these fields

dhargo:
@matthewlane Thanks for the comments. I've commited fixes. Here is new parameters to use for testing bid response
```
http://px.adhigh.net/rtb/direct_banner?pid=7&tid=test01&known=1&is_video=false&resp_type=JSON&size=300x250& 
```

modules/getintentBidAdapter.js
```diff
@@ -1,78 +1,142 @@
...
+   * @param {BidRequest[]} bidRequests - an array of bids.
+   * @return ServerRequest[]
+   */
+  buildRequests: function(bidRequests) {
+    return bidRequests.map(bidRequest => {
+      let giBidRequest = buildGiBidRequest(bidRequest);
+      return {
+        method: 'GET',
+        url: buildUrl(giBidRequest),
+        data: giBidRequest,
```
matthewlane: This should potentially return the `bidRequest.bidId` so that the `interpretResponse` function has access to it

dhargo:
@matthewlane Thank you for the comments. Fixed missing fields (`requestId`, `creativeId`).

Have to notice: method signature `interpretResponse: function(serverResponse, request)` and example in documentation http://prebid.org/dev-docs/bidder-adapter-1.html looks like we always have BidRequest (`request` parameter) when interpreting server bid response; so it is unclear why do we have to pass `requestId` to server and back.

matthewlane:
Sorry for the confusion about that. The second parameter to `interpretResponse` is the bid request object which is available for use if needed. This object is whatever was returned the `buildRequests` function. Bid request and response objects need to have matching ids and one way of doing that is returning the id from `buildRequests`, but they can also be sent and received from the server as you've done here. In any case I was able to verify bids with your adapter so this is good for merge

PR #1727, opened 2017-10-19
add option to run tests in a specific file

thewizarodofoz:
## Type of change
- [x] CI related changes

## Description of change
When developing an adapter, I don't need to run the whole test suite, it's very time consuming. This PR adds an option to run a specific test file. Usage:
```shell
$ gulp test --file "test/spec/modules/vidazooBidAdapter_spec.js"
```

### DIFF
--- a/gulpfile.js
+++ b/gulpfile.js
@@ -173,10 +173,11 @@ gulp.task('webpack', ['clean'], function () {
 // By default, this runs in headless chrome.
 //
 // If --watch is given, the task will re-run unit tests whenever the source code changes
+// If --file "<path-to-test-file>" is given, the task will only run tests in the specified file.
 // If --browserstack is given, it will run the full suite of currently supported browsers.
 // If --browsers is given, browsers can be chosen explicitly. e.g. --browsers=chrome,firefox,ie9
 gulp.task('test', ['clean'], function (done) {
-  var karmaConf = karmaConfMaker(false, argv.browserstack, argv.watch);
+  var karmaConf = karmaConfMaker(false, argv.browserstack, argv.watch, argv.file);
 
   var browserOverride = helpers.parseBrowserArgs(argv).map(helpers.toCapitalCase);
   if (browserOverride.length > 0) {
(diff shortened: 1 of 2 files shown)


karma.conf.maker.js
```diff
@@ -92,10 +92,10 @@ function setBrowsers(karmaConf, browserstack) {
   }
 }
 
-module.exports = function(codeCoverage, browserstack, watchMode) {
+module.exports = function(codeCoverage, browserstack, watchMode, file) {
   var webpackConfig = newWebpackConfig(codeCoverage);
   var plugins = newPluginsArray(browserstack);
-  var files = [
+  var files = file ? [file] : [
```
dbemiller: I think this needs to be `var files = file ? ['test/helpers/prebidGlobal.js', file] : ...`

Otherwise it fails on files which use `$$PREBID_GLOBAL$$`--even if it's just indirectly, through the import chain. For example, `gulp test --file "test/spec/modules/appnexusAstBidAdapter_spec.js"` fails because the adapter relies on `src/config`, which relies on `$$PREBID_GLOBAL$$`.

This problem should fade away as progress gets made on #1508 and #1510... but for now it's common enough that the gulpfile should probably support it.

gulpfile.js
```diff
@@ -173,10 +173,11 @@ gulp.task('webpack', ['clean'], function () {
 // By default, this runs in headless chrome.
 //
 // If --watch is given, the task will re-run unit tests whenever the source code changes
+// If --file "<path-to-test-file>" is given, the task will only run tests in the specified file.
 // If --browserstack is given, it will run the full suite of currently supported browsers.
 // If --browsers is given, browsers can be chosen explicitly. e.g. --browsers=chrome,firefox,ie9
 gulp.task('test', ['clean'], function (done) {
-  var karmaConf = karmaConfMaker(false, argv.browserstack, argv.watch);
+  var karmaConf = karmaConfMaker(false, argv.browserstack, argv.watch, argv.file);
```
dbemiller: Could you also add this to the `test-coverage` task?

thewizarodofoz: sure

thewizarodofoz:
@dbemiller thanks for the review! I have committed the changes.

PR #1729, opened 2017-10-19
Kargo Adapter for Prebid 1.0

samuelhorwitz:
## Description of change
Kargo Adapter migrated to Prebid 1.0

## Notes

As a side note, I would like some clarification on currency support. Does the publisher always specify conversion rates? We have multiple currency support being added into our server-side code and I would like to know the best way to configure things so that our rates our used when the pub specifies a compatible non-USD currency. We will be returning our CPMs in the currency the publisher specifies already converted on our end. I'm unsure however of how I should implement this in the adapter. Thank you!

### DIFF
--- a/modules/kargoBidAdapter.js
+++ b/modules/kargoBidAdapter.js
@@ -1,60 +1,60 @@
-const bidfactory = require('src/bidfactory.js');
-const bidmanager = require('src/bidmanager.js');
-const adloader = require('src/adloader.js');
-const utils = require('src/utils.js');
-const adaptermanager = require('src/adaptermanager');
-const CONSTANTS = require('src/constants.json');
-const HOST = $$PREBID_GLOBAL$$.kargo_kraken_host || 'https://krk.kargo.com';
-
-const KargoAdapter = function KargoAdapter() {
-  function _handleBid(bids) {
-    return function wrappedHandleBid(adUnits) {
-      utils._map(bids, bid => {
-        let adUnit = adUnits[bid.params.placementId];
-
-        if (adUnit) {
-          bidmanager.addBidResponse(bid.placementCode, _createBid(adUnit));
-
-          if (adUnit.receivedTracker) {
-            var el = document.createElement('img');
-            el.src = adUnit.receivedTracker;
-            document.body.appendChild(el);
-          }
-        }
-      });
-    };
-  }
-
-  function _createBid(adUnit) {
-    let bidObject = bidfactory.createBid(CONSTANTS.STATUS.GOOD);
-    bidObject.bidderCode = 'kargo';
-    bidObject.cpm = Number(adUnit.cpm);
-    bidObject.ad = adUnit.adm;
-    bidObject.width = adUnit.width;
-    bidObject.height = adUnit.height;
-    return bidObject;
-  }
-
-  function _callBids(params) {
+import * as utils from 'src/utils';
+import {config} from 'src/config';
+import {registerBidder} from 'src/adapters/bidderFactory';
+const BIDDER_CODE = 'kargo';
+const HOST = 'https://krk.kargo.com';
+export const spec = {
+  code: BIDDER_CODE,
+  isBidRequestValid: function(bid) {
+    if (!bid || !bid.params) {
+      return false;
+    }
+    return !!bid.params.placementId;
+  },
+  buildRequests: function(validBidRequests, bidderRequest) {
+    const currency = config.getConfig('currency');
     const transformedParams = Object.assign({}, {
-      timeout: params.timeout,
-      currency: 'USD',
+      timeout: bidderRequest.timeout,
+      currency: currency,
       cpmGranularity: 1,
       cpmRange: {
         floor: 0,
         ceil: 20
       },
-      adSlotIds: utils._map(params.bids, bid => bid.params.placementId)
-    }, _getAllMetadata());
+      adSlotIds: utils._map(validBidRequests, bid => bid.params.placementId)
+    }, spec._getAllMetadata());
     const encodedParams = encodeURIComponent(JSON.stringify(transformedParams));
-    const callbackName = `kargo_prebid_${params.requestId.replace(/-/g, '_')}`;
+    return Object.assign({}, bidderRequest, {
+      method: 'GET',
+      url: `${HOST}/api/v1/bid`,
+      data: `json=${encodedParams}`,
+      currency: currency
+    });
+  },
+  interpretResponse: function(response, bidRequest) {
+    let adUnits = response.body;
+    let bids = {};
+    utils._each(bidRequest.bids, bid => bids[bid.params.placementId] = bid);
+    const bidResponses = [];
+    for (let adUnitId in adUnits) {
+      let adUnit = adUnits[adUnitId];
+      bidResponses.push({
+        requestId: bids[adUnitId].bidId,
+        cpm: Number(adUnit.cpm),
+        width: adUnit.width,
+        height: adUnit.height,
+        ad: adUnit.adm,
+        ttl: 300,
+        creativeId: adUnitId,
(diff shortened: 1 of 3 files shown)


dbemiller:
hey @samuelhorwitz, is there any reason you made this PR against the `prebid-1.0` branch? Since you're using the `bidderFactory`, you should be able to update in `master` directly.

This helps us with branch maintenance, because otherwise all the patches & bugfixes to your adapter between now and the 1.0 release will become merge conflicts.


## Currency
It looks like you're doing the right thing to me. Import the `config`, read the `currency` from it in `buildRequests`, and respond with the bid's currency when you `interpretResponse`.

@snapwich may want to overrule me here, as he's more familiar with the module... but your code here looks like it should work to me.

samuelhorwitz:
I can change the base of this PR, but should I actually rebase my branch off of master? If I change the base without rebasing, this PR will become much more complicated to review, but if I rebase from master I will be working off of a branch that doesn't include any of the 1.0 updates

dbemiller:
You should create a branch from master, and open a PR against master.

The 1.0 branch doesn't change any adapter-facing APIs in the `bidderFactory`... and in fact breaking changes to `bidderFactory` are aimed at `master` too (see #1748), so if anything prebid-1.0 will be farther out of date.

`currency` in the `config` exists in `master` too... so I don't think your code should change very much (if at all) between the two.

modules/kargoBidAdapter.js
```diff
@@ -0,0 +1,166 @@
...
+  },
+  interpretResponse: function(adUnits, bidRequest) {
+    let bids = {};
+    utils._each(bidRequest.bids, bid => bids[bid.params.placementId] = bid);
+    const bidResponses = [];
+    for (let adUnitId in adUnits) {
+      let adUnit = adUnits[adUnitId];
+      bidResponses.push({
+        requestId: bids[adUnitId].bidId,
+        bidderCode: spec.code,
```
dbemiller: Our docs are bad here, but... don't send back the `bidderCode` on bids. The `bidderFactory` adds it to you, and this won't work quite right if someone aliases your adapter (which can be done by customers at runtime)

modules/kargoBidAdapter.js
```diff
@@ -0,0 +1,166 @@
...
+        return cookie.substring(nameEquals.length, cookie.length);
+      }
+    }
+
+    return null;
+  },
+
+  _getCrbIds() {
+    try {
+      const crb = JSON.parse(decodeURIComponent(spec._readCookie('krg_crb')));
```
mkendall07: what's this cookie you are reading? How do you have access to it if you are in the pub domain?

modules/kargoBidAdapter.js
```diff
@@ -0,0 +1,166 @@
...
+        ttl: 300,
+        creativeId: adUnitId,
+        netRevenue: true,
+        currency: bidRequest.currency
+      });
+
+      if (adUnit.receivedTracker) {
+        let el = document.createElement('img');
+        el.src = adUnit.receivedTracker;
+        document.body.appendChild(el);
```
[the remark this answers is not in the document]
samuelhorwitz: When do user syncs fire? We can just leave this off if it's not an immediate thing... we use this tracker to log that an ad unit was received by the page but if it's not fired in a deterministically "soon" fashion then it's better we just leave it off.

modules/kargoBidAdapter.js
```diff
@@ -0,0 +1,166 @@
...
+      }
+    }
+    return bidResponses;
+  },
+
+  // PRIVATE
```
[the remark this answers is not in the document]
samuelhorwitz: For some reason I thought I mocked these more than I did in tests so I made them "private" as opposed to out of closure scope. I don't think I actually do, though. That being said, I don't really mind either way, I added that comment to visually separate the interface functionality from the behind the scenes functionality.

modules/kargoBidAdapter.js
samuelhorwitz: These are first party cookies our pubs are likely to already have on their page from our various other integrations.

mkendall07: ok that's what I thought. Thanks for clarifying.

modules/kargoBidAdapter.js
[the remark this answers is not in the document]
dbemiller: Aha... ok. Not a user sync at all, then.

Yeah... there's no real good way to do that. Publishers get annoyed by libraries making extra requests because of the impact it has on page load times.

You could use the user sync module if you want... by default, we fire them 3 seconds after the auction has completed.

However, it still wouldn't be reliable. Since it affects the publishers' page performance, Prebid made this configurable... so they can limit the number that fire, opt out totally, etc.

samuelhorwitz:
So, I just modified this PR to be derived from master/pointing to master. However I will need to re-run some of my own tests to make sure all my changes still work.

modules/kargoBidAdapter.js
samuelhorwitz: No problem

dbemiller:
Sounds good, let me know when you're comfortable with it.

modules/kargoBidAdapter.js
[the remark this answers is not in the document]
samuelhorwitz: Ah okay cool, I will remove it then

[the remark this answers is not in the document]
dbemiller: Sure thing.

If this is a feature you guys really need, I'd recommend opening an issue about it. I don't know how the Prebid.JS leadership feels about supporting this, but that's a good way to get discussion going with some more eyes on it.

samuelhorwitz:
Okay everything seems good from our tests! Thank you

modules/kargoBidAdapter.js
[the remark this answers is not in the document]
samuelhorwitz: I've created a ticket here https://github.com/prebid/Prebid.js/issues/1762 but it's not urgent for us to have this support and I've just removed the code

modules/kargoBidAdapter.md
```diff
@@ -0,0 +1,23 @@
...
+
+# Description
+
+Please use `kargo` as the bidder code.
+
+# Test Parameters
+```
+  var adUnits = [{
+    code: 'div-gpt-ad-1460505748561-1',
+    sizes: [[300,250][1,1]],
```
dbemiller: Looks like you're missing a comma here.

modules/kargoBidAdapter.md
```diff
@@ -0,0 +1,23 @@
+# Overview
+
+**Module Name**: Kargo Bidder Adapter
+**Module Type**: Bidder Adapter
+**Maintainer**: kraken@kargo.com
+
+# Description
```
dbemiller: You might want to mention something about how this only works on mobile.

I tried this and thought I wasn't getting a bid back... but I put my browser in mobile mode and then it worked. Didn't realize that was a dependency. Others may run into the same problem.

dbemiller:
agh... sorry about this, but one more change, because #1742 got merged before we could get this in.

The first argument to interpretResponse now looks like this:

```
{
  body: responseBody,
  headers: {
  }
}
```

You'll have to update the spec so that it looks through that object as well now.

samuelhorwitz:
Alright, done!

modules/kargoBidAdapter.md
samuelhorwitz: I updated this as well but its below the line commented on so Github didnt dismiss it

samuelhorwitz:
Hey, sorry about this but now I have one last thing I need to add also! It should be a very quick/small change to what we send to our bidder.

dbemiller:
no worries... do what you've gotta do ^^

samuelhorwitz:
Alright finished again!

PR #1733, opened 2017-10-19
Auction module updated to support currency/hook

jaiminpanchal27:
## Type of change

## Description of change
Have updated auction module to support currency/hook plugin. 
Moved `isValid` function to validate bid response to bidderFactory.


src/auction.js
```diff
@@ -353,19 +188,128 @@ export function newAuction({adUnits, adUnitCodes, callback, cbTimeout}) {
...
     getBidsReceived: () => _bidsReceived,
-    callBids
   }
 }
 
+export const addBidResponse = createHook('asyncSeries', function addBidResponse(adUnitCode, bid) {
+  let auctionInstance = this;
+  let bidRequests = auctionInstance.getBidRequests();
+  let auctionId = auctionInstance.getAuctionId();
+
```
snapwich: What happened to `isValid` and all its code that validated the bid?

src/auction.js
```diff
@@ -353,19 +188,128 @@ export function newAuction({adUnits, adUnitCodes, callback, cbTimeout}) {
...
+    } else {
+      doCallbacksIfNeeded();
+      addBidToAuction(bidResponse);
+    }
+  }
+});
+
+// Postprocess the bids so that all the universal properties exist, no matter which bidder they came from.
+// This should be called before addBidToAuction().
+function getPreparedBidForAuction({adUnitCode, bid, bidRequests, auctionId}) {
```
snapwich: How come this was moved outside `addBidResponse`?

src/auction.js
jaiminpanchal27: I moved `isValid` to `bidderFactory`. I think that is the best place to validate bid response.

src/auction.js
jaiminpanchal27: addBidResponse function had multiple inner functions like `isValid`, `preparedBidForAuction` and `validBidSize` which was creating confusing so moved validation logic to `bidderFactory` and also moved `getPreparedBidForAuction`. 

Previously `preparedBidForAuction` was mutating bid object and now its not.

PR #1738, opened 2017-10-20
Validating bid response params

jaiminpanchal27:
## Type of change

## Description of change
Validating bid response params `currency`, 'netRevenue` and `ttl` for 1.0 adapters.

### DIFF
--- a/modules/appnexusAstBidAdapter.js
+++ b/modules/appnexusAstBidAdapter.js
@@ -181,7 +181,7 @@ function newBid(serverBid, rtbBid) {
   const bid = {
     requestId: serverBid.uuid,
     cpm: rtbBid.cpm,
-    creative_id: rtbBid.creative_id,
+    creativeId: rtbBid.creative_id,
     dealId: rtbBid.deal_id,
     currency: 'USD',
     netRevenue: true,
--- a/modules/appnexusAstBidAdapter.js
+++ b/modules/appnexusAstBidAdapter.js
@@ -183,6 +183,9 @@ function newBid(serverBid, rtbBid) {
     cpm: rtbBid.cpm,
     creative_id: rtbBid.creative_id,
     dealId: rtbBid.deal_id,
+    currency: 'USD',
+    netRevenue: true,
+    ttl: 300
   };
 
   if (rtbBid.rtb.video) {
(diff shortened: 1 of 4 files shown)


src/adapters/bidderFactory.js
```diff
@@ -107,6 +107,8 @@ import { logWarn, logError, parseQueryStringParameters, delayExecution } from 's
  * @property {string} url The URL which makes the sync happen.
  */
 
+const BID_RESPONSE_KEYS = ['currency', 'netRevenue', 'ttl'];
```
mkendall07: We have more required params. You should check http://prebid.org/dev-docs/bidder-adapter-1.html#interpreting-the-response ;)

jaiminpanchal27: Updated

modules/appnexusAstBidAdapter.js
```diff
@@ -183,14 +183,18 @@ function newBid(serverBid, rtbBid) {
     cpm: rtbBid.cpm,
     creative_id: rtbBid.creative_id,
```
matthewlane: `appnexusAst` will need `bidderCode` and `creativeId` added to pass validation now

src/adapters/bidderFactory.js
```diff
@@ -288,12 +290,17 @@ export function newBidder(spec) {
           onResponse();
 
           function addBidUsingRequestMap(bid) {
-            const bidRequest = bidRequestMap[bid.requestId];
-            if (bidRequest) {
-              const prebidBid = Object.assign(bidfactory.createBid(STATUS.GOOD, bidRequest), bid);
-              addBidWithCode(bidRequest.placementCode, prebidBid);
+            // In Prebid 1.0 all the validation logic from bidmanager will move here, as of now we are only validating new params so that adapters dont miss adding them.
+            if (hasValidKeys(bid)) {
```
dbemiller: Careful here...
Careful here... we've been telling adapters _not_ to add `bidderCode`, because it breaks aliasing. The `bidderFactory` should be doing that for them.

But... that happens a few lines after this: `Object.assign(bidfactory.createBid(STATUS.GOOD, bidRequest), bid);`.

So I don't think should be make sure that `bidderCode` exists here. If anything, we should make sure that it _doesn't_ exist yet.

src/adapters/bidderFactory.js
```diff
@@ -107,6 +107,8 @@ import { logWarn, logError, parseQueryStringParameters, delayExecution } from 's
  * @property {string} url The URL which makes the sync happen.
  */
 
+const BID_RESPONSE_KEYS = ['requestId', 'bidderCode', 'cpm', 'width', 'height', 'ad', 'ttl', 'creativeId', 'netRevenue', 'currency'];
```
mkendall07: Per recent discussion, it looks like `bidderCode` should not be a required param.

mkendall07:
I think ideally we'd want to validate on media type (probably create different bid objects for the different types) but that doesn't have to be done in this PR.

PR #1739, opened 2017-10-20
add vastUrl + media type for video bids Prebid Server

mkendall07:
## Type of change

- [x] Bugfix

## Description of change
Incremental updates for getting video demand working through PBS.

### DIFF
--- a/modules/prebidServerBidAdapter.js
+++ b/modules/prebidServerBidAdapter.js
@@ -111,7 +111,9 @@ function PrebidServer() {
       if (videoMediaType) {
         // pbs expects a ad_unit.video attribute if the imp is video
         adUnit.video = Object.assign({}, videoMediaType);
-        delete adUnit.mediaTypes.video;
+        delete adUnit.mediaTypes;
+        // default is assumed to be 'banner' so if there is a video type we assume video only until PBS can support multi format auction.
+        adUnit.media_types = ['video'];
       }
     })
     convertTypes(adUnits);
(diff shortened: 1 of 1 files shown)


modules/prebidServerBidAdapter.js
```diff
@@ -196,10 +198,15 @@ function PrebidServer() {
             bidObject.creative_id = bidObj.creative_id;
             bidObject.bidderCode = bidObj.bidder;
             bidObject.cpm = cpm;
-            bidObject.ad = bidObj.adm;
-            if (bidObj.nurl) {
-              bidObject.ad += utils.createTrackPixelHtml(decodeURIComponent(bidObj.nurl));
+            if (bidObj.media_type === 'video') {
+              bidObject.vastUrl = bidObj.nurl;
```
dbemiller: Is a `nurl` guaranteed on video bids?

In section `6.4.2` of the OpenRTB 2.5 spec, it looks like video bids could either serve `vastUrl` through the `nurl`, _or_ `vastXml` through `adm`.

modules/prebidServerBidAdapter.js
```diff
@@ -196,10 +198,15 @@ function PrebidServer() {
...
+            } else {
+              bidObject.ad = bidObj.adm;
+              if (bidObj.nurl) {
+                bidObject.ad += utils.createTrackPixelHtml(decodeURIComponent(bidObj.nurl));
+              }
```
dbemiller: Looks like these should be the other way around. From the spec, `adm` is described as:

> Optional means of conveying ad markup in case the bid wins; supersedes the win notice if markup is included in both. 

(see section 4.2.3)

mkendall07:
@bretg fyi on video support

modules/prebidServerBidAdapter.js
```diff
@@ -111,7 +112,9 @@ function PrebidServer() {
       if (videoMediaType) {
         // pbs expects a ad_unit.video attribute if the imp is video
         adUnit.video = Object.assign({}, videoMediaType);
-        delete adUnit.mediaTypes.video;
+        delete adUnit.mediaTypes;
+        // default is assumed to be 'banner' so if there is a video type we assume video only until PBS can support multi format auction.
+        adUnit.media_types = ['video'];
```
matthewlane: Would it make sense to use the `VIDEO` constant here too?

LGTM

PR #1741, opened 2017-10-21
updated adxcg adapter for prebid 1.0

adxcgcom:
## Type of change

- [ ] Bugfix
- [ ] Feature
- [ ] New bidder adapter   
- [ ] Build related changes
- [ ] CI related changes
- [X] Other

## Description of change

adxcg bid adapter update for Prebid 1.0

### DIFF
--- a/modules/adxcgBidAdapter.js
+++ b/modules/adxcgBidAdapter.js
@@ -34,7 +34,6 @@ export const spec = {
   buildRequests: function (validBidRequests, bidderRequest) {
     utils.logMessage(`buildRequests: ${JSON.stringify(validBidRequests)}`);
 
-    // if (validBidRequests && validBidRequests.length > 0) {
     let adZoneIds = [];
     let prebidBidIds = [];
     let sizes = [];
@@ -52,7 +51,6 @@ export const spec = {
     requestUrl.search = null;
     requestUrl.hash = null;
 
-    // noinspection JSDeprecatedSymbols
     let adxcgRequestUrl = url.format({
       protocol: secure ? 'https' : 'http',
       hostname: secure ? 'hbps.adxcg.net' : 'hbp.adxcg.net',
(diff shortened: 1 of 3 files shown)


modules/adxcgBidAdapter.js
```diff
@@ -1,95 +1,121 @@
...
+  interpretResponse: function (serverResponse, bidRequests) {
+    // utils.logMessage(`interpretResponse serverResponse: ${JSON.stringify(response)}`); // remove for PRODUCTION
+    // utils.logMessage(`interpretResponse bidRequest ${JSON.stringify(bidRequests)}`); // remove for PRODUCTION
+    let bids = [];
+
+    serverResponse.forEach(serverResponseOneItem => {
+      let bid = {};
+
+      bid.requestId = serverResponseOneItem.bidId;
+      bid.bidderCode = spec.code;
```
dbemiller: You shouldn't set the `bidderCode` here... It breaks adapter aliasing.

Our docs were probably bad at the time you wrote this, but they've since been updated.

modules/adxcgBidAdapter.js
```diff
@@ -1,95 +1,121 @@
...
+  },
+  /**
+   * Unpack the response from the server into a list of bids.
+   *
+   * @param {*} serverResponse A successful response from the server.
+   * @return {bidRequests[]} An array of bids which were nested inside the server.
+   */
+  interpretResponse: function (serverResponse, bidRequests) {
+    // utils.logMessage(`interpretResponse serverResponse: ${JSON.stringify(response)}`); // remove for PRODUCTION
+    // utils.logMessage(`interpretResponse bidRequest ${JSON.stringify(bidRequests)}`); // remove for PRODUCTION
```
dbemiller: Remove the dead code, please.

dbemiller:
akh... sorry, but one more change, because #1742 got merged before we could get this in.

The first argument to interpretResponse now looks like this:

```
{
  body: responseBody,
  headers: {
  }
}
```

You'll have to pull master and update the spec so that it looks digs into that object as well now.

Issue #1742, opened 2017-10-22
response headers are required but missing in interpretResponse


francoroy:
Hi, at:
https://github.com/prebid/Prebid.js/blob/be21952ee1ffa8f183f718f63d612b6cfdb84908/src/adapters/bidderFactory.js#L274
The response body is sent back to the adapter, while in some cases additional response data is required, for example - in mobfox adapter, the bid value is returned as a response header.

Any rejections to adding the response headers as a third parameter (so that it'd be optional and non-breaking for those who've already implemented their new adapter) ?

Thanks

dbemiller:
This is a good point.

I chatted with @mkendall07 about this, and we agreed that it's so early that breaking the API to keep it clean is a good idea.

I'll work on this early today, and we'll try to rush a review so that we can merge it in later. There aren't _too_ many 1.0 adapters in yet... so it won't be hard to update them.

PR #1743, opened 2017-10-22
Prebid 1.0 compliant Komoona bidder adapter

tzafrirb:
## Type of change

- [ ] Bugfix
- [ ] Feature
- [ ] New bidder adapter   
- [ ] Build related changes
- [ ] CI related changes
- [x] Other

## Description of change

```
{
  bidder: '<bidder name>',
  params: {
    // ...
  }
}
```

### DIFF
--- a/modules/komoonaBidAdapter.js
+++ b/modules/komoonaBidAdapter.js
@@ -83,8 +83,8 @@ export const spec = {
   interpretResponse: (response, request) => {
     const bidResponses = [];
     try {
-      if (response.bids) {
-        response.bids.forEach(bid => {
+      if (response.body && response.body.bids) {
+        response.body.bids.forEach(bid => {
           // The bid ID. Used to tie this bid back to the request.
           bid.requestId = bid.uuid;
           // The creative payload of the returned bid.
--- a/modules/komoonaBidAdapter.js
+++ b/modules/komoonaBidAdapter.js
@@ -8,8 +8,6 @@ const USYNCURL = '//s.komoona.com/sync/usync.html';
 export const spec = {
   code: BIDDER_CODE,
 
-  aliases: ['komoona'],
-
   /**
   * Determines whether or not the given bid request is valid. Valid bid request must have placementId and hbid
   *
--- a/modules/komoonaBidAdapter.js
+++ b/modules/komoonaBidAdapter.js
@@ -1,117 +1,123 @@
-import Adapter from 'src/adapter';
-import bidfactory from 'src/bidfactory';
-import bidmanager from 'src/bidmanager';
 import * as utils from 'src/utils';
-import { ajax } from 'src/ajax';
-import { STATUS } from 'src/constants';
-import adaptermanager from 'src/adaptermanager';
+import { registerBidder } from 'src/adapters/bidderFactory';
 
+const BIDDER_CODE = 'komoona';
 const ENDPOINT = '//bidder.komoona.com/v1/GetSBids';
+const USYNCURL = '//s.komoona.com/sync/usync.html';
 
-function KomoonaAdapter() {
-  let baseAdapter = new Adapter('komoona');
-  let bidRequests = {};
+export const spec = {
+  code: BIDDER_CODE,
 
-  /* Prebid executes this function when the page asks to send out bid requests */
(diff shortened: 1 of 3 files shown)


tzafrirb:
Prebid 1.0 compliant Komoona bidder adapter

modules/komoonaBidAdapter.js
```diff
@@ -1,117 +1,123 @@
...
+import { registerBidder } from 'src/adapters/bidderFactory';
 
+const BIDDER_CODE = 'komoona';
 const ENDPOINT = '//bidder.komoona.com/v1/GetSBids';
+const USYNCURL = '//s.komoona.com/sync/usync.html';
+
+export const spec = {
+  code: BIDDER_CODE,
+
+  aliases: ['komoona'],
```
dbemiller: This is just for adapters which have multiple names... since this is the same as your `BIDDER_CODE`, you can just delete it.

modules/komoonaBidAdapter.js
```diff
@@ -1,117 +1,123 @@
...
-        .map(bidId => bidRequests[bidId].placementCode)
-        .forEach(placementCode => {
-          bidmanager.addBidResponse(placementCode, createBid(STATUS.NO_BID));
+      if (response.bids) {
+        response.bids.forEach(bid => {
+          // The bid ID. Used to tie this bid back to the request.
+          bid.requestId = bid.uuid;
+          // The creative payload of the returned bid.
+          bid.ad = bid.creative;
+          bidResponses.push(bid);
```
dbemiller: There are lots of required params missing from these bids. See: https://github.com/prebid/Prebid.js/pull/1738/files#diff-ac0af2601dcafed83fa9a11b814bf3b5L109

tzafrirb: Our md file was updated with correct placement Id and hb Id for test creative (as well as our servers) and now when using placementId and hbId from md file, the hello_world file will always return a creative

modules/komoonaBidAdapter.js
tzafrirb: Removed from adapter file, thanks from the comment

modules/komoonaBidAdapter.js
tzafrirb: Working to fix it, it was added to our servers and now testing it out - will ping once live and fixed

tzafrirb:
CI test error was fixed, sorry for that - now Komoona gulp test and lint test pass.
Mandatory fields were added to response from our bid server

dbemiller:
Still getting no bids back from the test page with the params in your `.md` file.

Maybe a geographic thing? We're testing from the US.

tzafrirb:
Thanks for the comment and pointing to correct change - we have sync our fork with master and change adapter to match change  [#1742](https://github.com/prebid/Prebid.js/issues/1742)

dbemiller:
Looks good now... thanks!

test/spec/modules/komoonaBidAdapter_spec.js
```diff
@@ -1,152 +1,164 @@
...
+    expect(spec.isBidRequestValid({ params: {} })).to.equal(false);
+    expect(spec.isBidRequestValid({ params: { hbid: 12345 } })).to.equal(false);
+    expect(spec.isBidRequestValid({ params: { placementid: 12345 } })).to.equal(false);
+    expect(spec.isBidRequestValid({ params: { hbid: 12345, placementId: 67890 } })).to.equal(true);
+    expect(spec.isBidRequestValid({ params: { hbid: 12345, placementId: 67890, floorPrice: 0.8 } })).to.equal(true);
+  });
 
-    it('handles JSON.parse errors', () => {
-      server.respondWith('');
+  describe('Verify komoonaAdapter build request', () => {
```
jaiminpanchal27: These test cases need to be `it` block for 2 reasons.
1) karma is not showing actual count of tests in logs. So even though this is getting executed but count is not correct. I placed it in `it` block and test count was incremented.
2) If this test fails than it stops here itself.

tzafrirb: True comment and I accept it - we actually use it for internal products test, not sure why describe was used (probably just copied from old test).

It is fixed now

PR #1744, opened 2017-10-22
update mantis adapter for prebid 1.0

parisholley:


dbemiller:
Lots of unintended changes in this diff... please submit one with only the changes you intend to make.

parisholley:
@dbemiller not sure I follow, the only way to submit a pull for this, is to fork the prebid branch and submit a pull off that. the "unintended" changes are because you guys haven't updated master?

dbemiller:
It looks like you forked from the `prebid-1.0` branch, and made your PR against `master`.

You should fork from `master`, and make the PR against that instead (even though it's a 1.0 adapter).

The `bidderFactory.registerBidder` function does a bunch of stuff in the background to make your adapter work in prebid 0.x as well. This helps minimize the merge conflicts we have to deal with while the two branches are separate.

PR #1748, opened 2017-10-23
Make response headers available to the specs

### LINKED ISSUE #1742: response headers are required but missing in interpretResponse

dbemiller:
## Type of change
- Feature

## Description of change
Some adapters need to pull information from their server's response headers.

This adds that feature, and updates existing adapters so they continue to work.

## Other information
Fixes #1742.

### DIFF
--- a/src/adapters/bidderFactory.js
+++ b/src/adapters/bidderFactory.js
@@ -44,7 +44,7 @@ import { logWarn, logError, parseQueryStringParameters, delayExecution } from 's
  * @property {function(BidRequest[], bidderRequest): ServerRequest|ServerRequest[]} buildRequests Build the request to the Server
  *   which requests Bids for the given array of Requests. Each BidRequest in the argument array is guaranteed to have
  *   passed the isBidRequestValid() test.
- * @property {function(*, BidRequest): Bid[]} interpretResponse Given a successful response from the Server,
+ * @property {function(ServerResponse, BidRequest): Bid[]} interpretResponse Given a successful response from the Server,
  *   interpret it and return the Bid objects. This function will be run inside a try/catch.
  *   If it throws any errors, your bids will be discarded.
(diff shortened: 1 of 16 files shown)


src/adapters/bidderFactory.js
```diff
@@ -296,6 +311,12 @@ export function newBidder(spec) {
               logWarn(`Bidder ${spec.code} made bid for unknown request ID: ${bid.requestId}. Ignoring.`);
             }
           }
+
+          function headerParser(xmlHttpResponse) {
+            return {
+              get: responseObj.getResponseHeader.bind(responseObj)
```
mkendall07: why do a bind here?

dbemiller: Paranoia... because I've been burned by `this` bugs too many times.

The root of the issue is: I don't know how browsers implement `XMLHttpRequest.getResponseHeader`.

Suppose I just did:

```
return {
  get: responseObj.getResponseHeader
}
```

but then they implemented it as something like this:

```
XmlHttpRequest.prototype.getResponseHeader = function(header) {
  var headers = this.parseResponseHeaders();
  return headers[header];
}
```

When an adapter calls `get`, then `this` evaluates to _our_ object--not the original `XmlHttpRequest`. Since we don't define `parseResponseHeaders`, it'd crash on an undefined function error.

I have no idea if this is a real risk... but the tl;dr is that `this` sucks. Since I can't eject it from the language, defensive binds keep the code safe.

mkendall07: ah ok cool

PR #1750, opened 2017-10-23
Upgrade Audience Network adapter for Prebid 1.0

lovell:
As requested by @mjacobsonny and paid for by Facebook.

In addition to 100% unit test branch coverage, I've manually tested this using `pbjs_debug=true` and everything appears to be working end-to-end.

The two biggest gotchas I ran into that required me to dig into the new Prebid source code to determine expected behaviour were:

1. The inconsistent properties (type=method, endpoint=url?) of the `ServerRequest` expected from `buildRequests()` documented at http://prebid.org/dev-docs/bidder-adapter-1.html#building-the-request
2. The need to include the original `bidId` passed to `buildRequests()` as the differently-named `requestId` property of the `bidObject` returned by `interpretResponse()`.

### DIFF
--- a/modules/audienceNetworkBidAdapter.js
+++ b/modules/audienceNetworkBidAdapter.js
@@ -1,23 +1,25 @@
 /**
  * @file AudienceNetwork adapter.
  */
-import { ajax } from 'src/ajax';
-import { createBid } from 'src/bidfactory';
-import { addBidResponse } from 'src/bidmanager';
-import { STATUS } from 'src/constants.json';
-import { format } from 'src/url';
-import { logError } from 'src/utils';
-import Adapter from 'src/adapter';
-import adaptermanager from 'src/adaptermanager';
+import { registerBidder } from 'src/adapters/bidderFactory';
+import { config } from 'src/config';
+import { formatQS } from 'src/url';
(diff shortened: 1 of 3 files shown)


modules/audienceNetworkBidAdapter.js
```diff
@@ -0,0 +1,205 @@
...
+
+        const format = adformats[i];
+        const [width, height] = sizes[i];
+        const ad = createAdHtml(creativeId, format, fb_bidid);
+        const requestId = requestIds[i];
+
+        const bidResponse = {
+          // Prebid attributes
+          requestId,
+          bidderCode: code,
```
dbemiller: For some subtle reasons, you shouldn't set `bidderCode` in here. It breaks adapter aliasing.

Our docs have been updated to reflect this... but sorry for the confusion all the same.

dbemiller:
agh... sorry about this, but one more change, because #1742 got merged before we could get this in.

The first argument to `interpretResponse` has been changed, and now looks like this:

```
{
  body: responseBody,
  headers: {
  }
}
```

Could you also pull master and update that?

lovell:
Updated and rebased against master but other, unrelated unit tests are failing at the moment.

Issue #1759, opened 2017-10-24
New Optimera Adaptor for 1.0


mcallari:
## Type of issue
Feature Request

## Description
Adding Optimera's bid adapter for public use. Optimera's technology allows publishers to target viewability data to automate viewability optimization for any line item or PMP. Optimera does not push bids but instead targeting data via the deal ID space.

test parameters for validating bids

- Optimera will pass targeting data via hb_deal_optimera, will never pass a size therefore 0x0, will never pass a price therefore 0.00. For example:
hb_deal_optimera=RB_K,300x250K,300x600K&hb_size_optimera=0x0&hb_pb_optimera=0.00&hb_adid_optimera=86a7b6905fe174&hb_bidder_optimera=optimera

```
{
  bidder: 'optimera',
  params: {
    custom: {
      clientId: '1234'
    }
  }
}
```
An external lib is needed and can be added to the site's footer:
```
<script src="https://s3.amazonaws.com/elasticbeanstalk-us-east-1-397719490216/external_json/oPS.js"></script>
```
contact email of the adapter’s maintainer
kcandiotti@optimera.nyc

- [x] official adapter submission

## Other information
Test Page:
http://optimeratest.us-east-1.elasticbeanstalk.com/releaseProjects/prebid/demonstrationPage.php
@kcandiotti

Relates to PR: [https://github.com/prebid/Prebid.js/pull/1617](url)

matthewlane:
Thanks for submitting, the PR is assigned in the queue for review

PR #1617, opened 2017-09-24
Adding Optimera Bid Adapter

kcandiotti:
## Type of change

- [ ] Bugfix
- [ ] Feature
- [X] New bidder adapter   
- [ ] Code style update (formatting, local variables)
- [ ] Refactoring (no functional changes, no api changes)
- [ ] Build related changes
- [ ] CI related changes
- [ ] Other

## Description of change

Adding Optimera's bid adapter for public use. Optimera's technology allows publishers to target viewability data to automate viewability optimization for any line item or PMP. Optimera does not push bids but instead targeting data via the deal ID space.

- test parameters for validating bids
Optimera will pass targeting data via hb_deal_optimera, will never pass a size therefore 0x0, will never pass a price therefore 0.00. For example:
hb_deal_optimera=RB_K,300x250K,300x600K&hb_size_optimera=0x0&hb_pb_optimera=0.00&hb_adid_optimera=86a7b6905fe174&hb_bidder_optimera=optimera
```
{
  bidder: 'optimera',
  params: {
    clientId: '1234'
  }
}
```

Be sure to test the integration with your adserver using the [Hello World](/integrationExamples/gpt/hello_world.html) sample page.

- contact email of the adapter’s maintainer
kcandiotti@optimera.nyc
- [X] official adapter submission

## Other information

Test Page:
http://optimeratest.us-east-1.elasticbeanstalk.com/releaseProjects/prebid/demonstrationPage.php
@kcandiotti

PR #1760, opened 2017-10-24
Optimera Adapter for 1.0 (#1759)

mcallari:
## Type of change
- [x] New bidder adapter  

## Type of issue
Feature Request

## Description
Adding Optimera's bid adapter for public use. Optimera's technology allows publishers to target viewability data to automate viewability optimization for any line item or PMP. Optimera does not push bids but instead targeting data via the deal ID space.

test parameters for validating bids

- Optimera will pass targeting data via hb_deal_optimera, will never pass a size therefore 0x0, will never pass a price therefore 0.00. For example:
hb_deal_optimera=RB_K,300x250K,300x600K&hb_size_optimera=0x0&hb_pb_optimera=0.00&hb_adid_optimera=86a7b6905fe174&hb_bidder_optimera=optimera

```
{
  bidder: 'optimera',
  params: {
    custom: {
      clientId: '1234'
    }
  }
}
```
An external lib is needed and can be added to the site's footer:
```
<script src="https://s3.amazonaws.com/elasticbeanstalk-us-east-1-397719490216/external_json/oPS.js"></script>
```
contact email of the adapter’s maintainer
kcandiotti@optimera.nyc

## Other information
Test Page:
http://optimera.elasticbeanstalk.com/prebidDemoPage.php
@kcandiotti

Relates to PR: https://github.com/prebid/Prebid.js/pull/1617

### DIFF
--- a/modules/optimeraBidAdapter.js
+++ b/modules/optimeraBidAdapter.js
@@ -11,7 +11,11 @@ export const spec = {
    * @return boolean True if this is a valid bid, and false otherwise.
    */
   isBidRequestValid: function (bidRequest) {
-    return !!(bidRequest.params.custom.clientID);
+    if (typeof bidRequest.params.custom != 'undefined' && typeof bidRequest.params.custom.clientID != 'undefined') {
+      return true;
+    } else {
+      return false;
+    }
   },
   /**
    * Make a server request from the list of BidRequests.
@@ -23,12 +27,12 @@ export const spec = {
     var optimeraHost = window.location.host;
     var optimeraPathName = window.location.pathname;
     var timestamp = Math.round(new Date().getTime() / 1000);
-    var clientID = validBidRequests[0].params.custom.clientID;
     var oDv = [];
-    if (clientID != undefined) {
+    if (typeof validBidRequests[0].params.custom.clientID != 'undefined') {
+      var clientID = validBidRequests[0].params.custom.clientID;
       oDv.push(clientID);
       for (var i = 0; i < validBidRequests.length; i++) {
-        oDv.push(validBidRequests[i].placementCode);
+        oDv.push(validBidRequests[i].adUnitCode);
       }
       window.oDv = oDv;
       var scoresURL = SCORES_BASE_URL + clientID + '/' + optimeraHost + optimeraPathName + '.js';
@@ -43,30 +47,39 @@ export const spec = {
   /**
    * Unpack the response from the server into a list of bids.
    *
+   * Some required bid params are not needed for this so default
+   * values are used.
+   *
    * @param {*} serverResponse A successful response from the server.
    * @return {Bid[]} An array of bids which were nested inside the server.
    */
   interpretResponse: function (serverResponse, bidRequest) {
-    var scores = serverResponse.replace('window.oVa = ', '');
+    var scores = serverResponse.body.replace('window.oVa = ', '');
     scores = scores.replace(';', '');
     scores = JSON.parse(scores);
     var validBids = bidRequest.payload;
     var bidResponses = [];
     var dealId = '';
     for (var i = 0; i < validBids.length; i++) {
-      if (validBids[i].placementCode in scores && validBids[i].params.custom.clientID != undefined) {
-        dealId = scores[validBids[i].placementCode];
+      if (typeof validBids[i].params.custom.clientID != 'undefined') {
+        if (validBids[i].adUnitCode in scores) {
+          dealId = scores[validBids[i].adUnitCode];
+        }
+        var bidResponse = {
+          bidderCode: spec.code,
+          requestId: validBids[i].bidId,
+          ad: '<div></div>',
+          cpm: 0.01,
+          width: 0,
+          height: 0,
+          dealId: dealId,
+          ttl: 300,
+          creativeId: '1',
+          netRevenue: '0',
+          currency: 'USD'
+        };
+        bidResponses.push(bidResponse);
       }
-      var bidResponse = {
-        bidderCode: spec.code,
-        requestId: validBids[i].bidId,
-        ad: '<div></div>',
-        cpm: 0.01,
-        width: 0,
-        height: 0,
-        dealId: dealId
-      };
-      bidResponses.push(bidResponse);
     }
     return bidResponses;
   }
--- a/modules/optimeraBidAdapter.js
+++ b/modules/optimeraBidAdapter.js
@@ -1,0 +1,75 @@
+import {registerBidder} from 'src/adapters/bidderFactory';
+const BIDDER_CODE = 'optimera';
+const SCORES_BASE_URL = 'https://s3.amazonaws.com/elasticbeanstalk-us-east-1-397719490216/json/client/';
+
+export const spec = {
+  code: BIDDER_CODE,
+  /**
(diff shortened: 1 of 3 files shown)


test/spec/modules/optimeraBidAdapter_spec.js
```diff
@@ -0,0 +1,80 @@
...
+          'placementCode': '0',
+          'placementId': '104333943',
+          'custom': {
+            'clientId': '0'
+          }
+        }
+      }
+    ];
+    it('buildRequests fires', () => {
+      var request = spec.buildRequests(validBidRequests);
```
snapwich: You appear to be missing some assertions

modules/optimeraBidAdapter.js
```diff
@@ -0,0 +1,75 @@
...
+export const spec = {
+  code: BIDDER_CODE,
+  /**
+   * Determines whether or not the given bid request is valid.
+   *
+   * @param {bidRequest} bid The bid params to validate.
+   * @return boolean True if this is a valid bid, and false otherwise.
+   */
+  isBidRequestValid: function (bidRequest) {
+    return !!(bidRequest.params.custom.clientID);
```
snapwich: This will throw an exception `custom.clientID` is not present.  Should do safe evaluation, `isBidRequestValid` should not throw.

modules/optimeraBidAdapter.js
```diff
@@ -0,0 +1,75 @@
...
+   * @param {validBidRequests[]} - an array of bids
+   * @return ServerRequest Info describing the request to the server.
+   */
+  buildRequests: function (validBidRequests) {
+    var optimeraHost = window.location.host;
+    var optimeraPathName = window.location.pathname;
+    var timestamp = Math.round(new Date().getTime() / 1000);
+    var clientID = validBidRequests[0].params.custom.clientID;
+    var oDv = [];
+    if (clientID != undefined) {
```
snapwich: preferred method for `undefined` check in javascript is `if (typeof clientID !== 'undefined') {`

modules/optimeraBidAdapter.js
```diff
@@ -0,0 +1,75 @@
...
+      oDv.push(clientID);
+      for (var i = 0; i < validBidRequests.length; i++) {
+        oDv.push(validBidRequests[i].placementCode);
```
snapwich: `bid.placementCode` has been deprecated, should be replaced with `bid.adUnitCode`

modules/optimeraBidAdapter.js
```diff
@@ -0,0 +1,75 @@
...
+      }
+      window.oDv = oDv;
+      var scoresURL = SCORES_BASE_URL + clientID + '/' + optimeraHost + optimeraPathName + '.js';
```
snapwich: This looks like JSONP. 

1.0 adapters should use AJAX capable endpoints.

mcallari: For this call, the platform generates a static js file with the JSON set to a var, so there isn't a callback. We are able to parse it from ```serverRsponse.body``` in ```interpretResponse```.  We were hoping as long as this works ok it would be fine for now until the platform can be updated.

mcallari:
Hi @snapwich, the latest changes have been pushed.

modules/optimeraBidAdapter.js
```diff
@@ -0,0 +1,88 @@
...
+    var optimeraPathName = window.location.pathname;
+    var timestamp = Math.round(new Date().getTime() / 1000);
+    var oDv = [];
+    if (typeof validBidRequests[0].params.custom.clientID != 'undefined') {
+      var clientID = validBidRequests[0].params.custom.clientID;
+      oDv.push(clientID);
+      for (var i = 0; i < validBidRequests.length; i++) {
+        oDv.push(validBidRequests[i].adUnitCode);
+      }
+      window.oDv = oDv;
```
snapwich: If your adapter is dependent on this global, then your adapter will not work with concurrent auctions (which is a requirement of 1.0).

mcallari: No problem. This is out and we can do it elsewhere.

modules/optimeraBidAdapter.js
```diff
@@ -0,0 +1,82 @@
...
+    var validBids = bidRequest.payload;
+    var bidResponses = [];
+    var dealId = '';
+    for (var i = 0; i < validBids.length; i++) {
+      if (typeof validBids[i].params.custom.clientID != 'undefined') {
+        if (validBids[i].adUnitCode in scores) {
+          dealId = scores[validBids[i].adUnitCode];
+        }
+        var bidResponse = {
+          bidderCode: spec.code,
```
matthewlane: `bidderCode` will be set automatically by `bidderFactory` now, this line can be dropped

mcallari: @matthewlane ok, I removed the bidderCode. Yes, that endpoint is correct, we just had an error and that has been restored. The Travis build failed, but passes on my local. Can you rerun the build? Thank you.

modules/optimeraBidAdapter.js
```diff
@@ -0,0 +1,81 @@
...
+   * Some required bid params are not needed for this so default
+   * values are used.
+   *
+   * @param {*} serverResponse A successful response from the server.
+   * @return {Bid[]} An array of bids which were nested inside the server.
+   */
+  interpretResponse: function (serverResponse, bidRequest) {
+    var scores = serverResponse.body.replace('window.oVa = ', '');
+    scores = scores.replace(';', '');
+    scores = JSON.parse(scores);
```
matthewlane: Any `JSON.parse` calls will need to be in a try/catch block

modules/optimeraBidAdapter.js
```diff
@@ -0,0 +1,81 @@
...
+    var bidResponses = [];
+    var dealId = '';
+    for (var i = 0; i < validBids.length; i++) {
+      if (typeof validBids[i].params.custom.clientID != 'undefined') {
+        if (validBids[i].adUnitCode in scores) {
+          dealId = scores[validBids[i].adUnitCode];
+        }
+        var bidResponse = {
+          requestId: validBids[i].bidId,
+          ad: '<div></div>',
```
mkendall07: this might have been mentioned before, but how do you render the ad if there is no ad content?

mcallari: hey @mkendall07, we are just creating this adpator to add visibility scores for the ad divs that will be added to DFP calls.

mkendall07: Ok but what does a publisher do with that information? And why is a request to your endpoint required to calculate viewability?

kcandiotti:
@mkendall07 Hi there. We are building this adapter specifically for a use case that a client wanted. Our technology is a targeting tech that normally pushes data directly into DFP in the header. We are utilizing prebid as a way to push data points into DFP so the client does not need to change their hardcode in order to work with us. The client will already know what to do with this data in their ad platform because of our preexisting relationship.

mkendall07:
@kcandiotti 
Thanks for the information.

Due to the reliance on an external script, we aren't going to accept this adapter into the 1.0 version of prebid.

We will be publishing some guidelines about how you can include open source external resource dependencies in just a bit. Once that is out, you can resubmit the adapter. Thanks for your patience.

PR #1763, opened 2017-10-24
add custom keyword support for pbs bid adapter

jsnellbaker:
## Type of change

- [x] Feature

## Description of change

Previously any custom param keypairs within the appnexus bidder were ignored and were not properly transferred to pbs.  This change now collects all the custom params and collates them under a single keywords param field in a format that pbs can ingest correctly for the AppNexus adapter.

Quick list of some features:
- Any keyvalues that were integers are converted to strings
- In case where only a key is specified (ie keyvalue is null), the key alone is passed along
- Each keyvalue (or set of keyvalues) are stored within its own array; this is for pbs to parse them out appropriately

## Other information

For RAD-1704


modules/prebidServerBidAdapter.js
```diff
@@ -98,6 +98,25 @@ function PrebidServer() {
             }
           }
         });
+        // will collect any custom params and place them under bid.params.keywords attribute in the following manner for pbs to ingest properly
+        // "keywords":[{"key":"randomKey","value":["123456789"]},{"key":"single_test"},{"key":"myVar","value":["myValue","124578"]}]
+        let kw_array = [];
```
mkendall07: nit: convention is camalCase.

mkendall07:
@jsnellbaker 
Could use minor update before merge.

jsnellbaker:
I have made some updates and added a test case in the areas @matthewlane suggested.

Please take a look and let me know if anything else should be reviewed again.

PR #1766, opened 2017-10-25
Prebid.js 1.0 support for Kumma adapter

yehonatanshac:
## Type of change

- [ ] Bugfix
- [ ] Feature
- [X ] New bidder adapter   
- [ ] Build related changes
- [ ] CI related changes
- [ ] Other

## Description of change

prebid.js 1.0 support for Kumma adapter

bidder: 'kumma',
          params: { 
            pubId: '28082',
            siteId: '26047',
            placementId: '123',
            size: '250X250'
          }

- contact email of the adapter’s maintainer yehonatan@kumma.com 
- [X ] official adapter submission

### DIFF
--- a/modules/kummaBidAdapter.js
+++ b/modules/kummaBidAdapter.js
@@ -1,63 +1,125 @@
-var bidfactory = require('src/bidfactory.js');
-var bidmanager = require('src/bidmanager.js');
-var adloader = require('src/adloader.js');
-var utils = require('src/utils.js');
-var CONSTANTS = require('src/constants.json');
-var adaptermanager = require('src/adaptermanager');
 
-var KummaAdapter = function KummaAdapter() {
-  function _callBids(params) {
-    var bidURL;
-    var bids = params.bids || [];
-    var requestURL = window.location.protocol + '//cdn.kumma.com/pb_ortb.js?cb=' + new Date().getTime() + '&ver=1&';
+import {logError, getTopWindowLocation} from 'src/utils';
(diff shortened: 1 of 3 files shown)


modules/kummaBidAdapter.js
```diff
@@ -1,63 +1,125 @@
...
+    };
+  },
+  interpretResponse: (response, request) => (
+    bidResponseAvailable(request, response)
+  ),
+};
+function bidResponseAvailable(bidRequest, bidResponse) {
+  const idToImpMap = {};
+  const idToBidMap = {};
+  const ortbRequest = parse(bidRequest.data);
```
dbemiller: You can simplify this quite a bit by just throwing the error here. The `bidderFactory` runs `interpretResponse` in a try/catch, and does a `logError` if anything goes wrong.

Also... your `parse` function just returns `null` on error, which will make `openrtbRequest.imp` throw an error anyway.

modules/kummaBidAdapter.js
```diff
@@ -1,63 +1,125 @@
...
+        domain: getTopWindowLocation().hostname,
+      },
+      id: siteId.toString(),
+      ref: referrer(),
+      page: getTopWindowLocation().href,
+    }
+  }
+  return null;
+}
+function referrer() {
```
dbemiller: This can be replaced by `getTopWindowReferrer()` in `utils`.

yehonatanshac:
I changed the md file. Now it should work.

PR #1781, opened 2017-10-30
Added adUnitCode to bidRequest for pre-1.0

jaiminpanchal27:
## Type of change
- [x] Bugfix

## Description of change
In 1.0 we updated all occurrences of `placementCode` to `adUnitCode` since both are same. A bug was created for 1.0 adapters who wanted to use `bidRequest.placementCode` in their adapter. 

So to support both pre-1.0 and 1.0 added `adUnitCode` in bidderFactory. Adapters can now use `adUnitCode` in their adapters and maintain compatibility for both versions.

## Other information
@snapwich #1776

### DIFF
--- a/src/adapters/bidderFactory.js
+++ b/src/adapters/bidderFactory.js
@@ -207,6 +207,10 @@ export function newBidder(spec) {
       const bidRequestMap = {};
       validBidRequests.forEach(bid => {
         bidRequestMap[bid.bidId] = bid;
+        // Delete this once we are 1.0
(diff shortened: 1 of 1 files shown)


GLStephen:
@jaiminpanchal27 @matthewlane will placementCode change to adUnitCode in analytics events?

mkendall07:
@GLStephen 
only once 1.0 branch becomes master. Before then `placementCode` will still function. You'll need to update your analytics adapter in the 1.0 branch though.

PR #1786, opened 2017-10-31
Fix test that hard-coded pbjs global.

mattpr:
## Type of change

- [x] Bugfix
- [ ] Feature
- [ ] New bidder adapter   
- [ ] Code style update (formatting, local variables)
- [ ] Refactoring (no functional changes, no api changes)
- [x] Build related changes
- [ ] CI related changes
- [ ] Other

PR #1787, opened 2017-10-31
Randomize pbjs global name when testing

mattpr:
## Type of change

- [ ] Bugfix
- [ ] Feature
- [ ] New bidder adapter   
- [x] Build related changes
- [ ] CI related changes
- [ ] Other

## Description of change

It seems that adapter tests routinely hard-code `pbjs` instead of using `$$PREBID_GLOBAL$$`.  I suppose things will change with prebid-1.0, but meanwhile this change forces those tests to break by randomizing the global name when testing.

There is probably a cleaner way to do this, but I also didn't want to do a bigger refactor given the plans to move away from using `$$PREBID_GLOBAL$$`.  

## References
#1786
#1666

### DIFF
--- a/karma.conf.maker.js
+++ b/karma.conf.maker.js
@@ -1,7 +1,7 @@
 // This configures Karma, describing how to run the tests and where to output code coverage reports.
 //
 // For more information, see http://karma-runner.github.io/1.0/config/configuration-file.html
-
+process.env.test = true;
 var _ = require('lodash');
 var webpackConf = require('./webpack.conf');
 var path = require('path')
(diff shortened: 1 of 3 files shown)


mkendall07:
The tests failed on this PR. I restarted the job to see if it will pass or not. If it doesn't pass again please rebase off master.

mattpr:
The tests passed when I opened the PR and are passing now. No idea why the runs in the middle failed.

jsnellbaker:
To note - this PR was reverted in lieu of some issues we started to encounter for the build process after this was originally merged.

mattpr:
Can you guys share any more info about the build issues that came up post-merge?  Happy to look at improving this PR.

mkendall07:
It put the build always in test mode. So production build wouldn't be pbjs variable.

mattpr:
Interesting.  `gulp build` doesn't run the karma tests for me and so the test flag isn't set and the resulting build uses the global pbjs name instead of the random name.  Can you give me an example of your production build command since I am guessing it isn't `gulp build`?

mkendall07:
I see. No the issue was when running `gulp serve` - this puts us into a development mode to test and build code. I believe it's also what we use to run our browserstack tests before release which is what failed.

PR #1793, opened 2017-11-02
Add freewheel ssp bidder adapter for prebid 1.0

guillaume-sticky:
## Type of change

- [ ] Bugfix
- [ ] Feature
- [ x] New bidder adapter
- [ ] Build related changes
- [ ] CI related changes
- [ ] Other

## Description of change

Add freewheel ssp Bidder Adaptor
This replace the old stickyadstv bidder adapter but uses the same aliases.
(freewheel-ssp or stickyadstv)

Add Prebid 1.0 compliance and mediatype "Video" compatibility

```
{
  bidder: 'freewheel-ssp',
  params: {
    zoneId : 277225  //only this parameter is mandatory
  }
}
```
- contact email of the adapter’s maintainer : gandouard@freewheel.tv

## Other information
Contact: gandouard@freewheel.tv

Test page: http://sandbox.stickyadstv.com/qa/prebid.html?zone=277225


modules/freewheelSSPBidAdapter.js
```diff
@@ -0,0 +1,303 @@
...
+  *
+  * @param {BidRequest[]} bidRequests A non-empty list of bid requests which should be sent to the Server.
+  * @return ServerRequest Info describing the request to the server.
+  */
+  buildRequests: function(bidRequests) {
+    // var currency = config.getConfig(currency);
+
+    this._currentBidRequest = bidRequests[0];
+    if (bidRequests.length > 1) {
+      console.log('Prebid.JS - freewheel bid adapter: only one ad unit is required.');
```
snapwich: Please use the util helper functions if you want to do any messaging as it's properly filtered by the debug flags.  You shouldn't be doing any direct `console.log`ing

modules/freewheelSSPBidAdapter.js
```diff
@@ -0,0 +1,303 @@
...
+  /**
+  * Make a server request from the list of BidRequests.
+  *
+  * @param {BidRequest[]} bidRequests A non-empty list of bid requests which should be sent to the Server.
+  * @return ServerRequest Info describing the request to the server.
+  */
+  buildRequests: function(bidRequests) {
+    // var currency = config.getConfig(currency);
+
+    this._currentBidRequest = bidRequests[0];
```
snapwich: This is not concurrency friendly.  This adapter instance can be used for multiple concurrent auctions in 1.0, which means that this state variable could be overwritten by a second auction before the first auction has completed.

The requests generated in `buildRequests` is passed as as second argument to `interpretResponses`.  It's possible to attach needed data to the request object and then access it later off off of the same request object.  Although I'd limit that as much as possible.

modules/freewheelSSPBidAdapter.js
```diff
@@ -0,0 +1,303 @@
...
+      topWin.freeheelssp_cache = {};
+    }
+    topWin.freeheelssp_cache[this._currentBidRequest.adUnitCode] = serverResponse;
+
+    const bidResponses = [];
+
+    if (princingData.price) {
+      const bidResponse = {
+        requestId: this._currentBidRequest.bidId,
+        bidderCode: this._currentBidRequest.bidder,
```
snapwich: This isn't necessary as it is automatically added by the bidderFactory

modules/freewheelSSPBidAdapter.js
```diff
@@ -0,0 +1,303 @@
...
+
+      bidResponses.push(bidResponse);
+    }
+
+    return bidResponses;
+  },
+
+  getUserSyncs: function(syncOptions) {
+    var img = new Image();
+    img.src = USER_SYNC_URL;
```
snapwich: You need to return the sync and let Prebid.js attach it to the page.  It is not allowed to add your own syncs in 1.0.  You can look at an example of how to return a properly formatted sync in one of the other 1.0 adapters like rubicon or appnexus.

modules/freewheelSSPBidAdapter.js
guillaume-sticky: Done, adding the initial request in the built bidRequest object to be retreived later in interpretBidResponse

modules/freewheelSSPBidAdapter.js
guillaume-sticky: Done.

modules/freewheelSSPBidAdapter.js
guillaume-sticky: Removed.

modules/freewheelSSPBidAdapter.js
guillaume-sticky: Done thanks.

guillaume-sticky:
All requested changes have been pushed.

modules/freewheelSSPBidAdapter.js
```diff
@@ -0,0 +1,314 @@
...
+
+    var requestParams = {
+      reqType: 'AdsSetup',
+      protocolVersion: '2.0',
+      zoneId: currentBidRequest.params.zoneId,
+      componentId: getComponentId(currentBidRequest.params.format)
+    };
+
+    var location = utils.getTopWindowUrl();
+    if (isValidUrl(location)) {
```
snapwich: Why are you validating that `location.href` is a valid URL?  That seems unnecessary....

modules/freewheelSSPBidAdapter.js
```diff
@@ -0,0 +1,314 @@
...
+    } catch (err) {
+      utils.logWarn('Prebid.js - ' + BIDDER_CODE + ' : ' + err);
+      return;
+    }
+
+    const princingData = getPricing(xmlDoc);
+    const creativeId = getCreativeId(xmlDoc);
+
+    const topWin = getTopMostWindow();
+    if (!topWin.freeheelssp_cache) {
```
snapwich: Is it just me or is `freeheel` a consistent typo?

modules/freewheelSSPBidAdapter.js
guillaume-sticky: We don't want to receive invalid urls. If loaded inside some iframe we could retreive location : "about:blank" for instance.

modules/freewheelSSPBidAdapter.js
guillaume-sticky: Indeed!   As you imagine it should be "freewheelssp".
Thx, I'll fix it.

guillaume-sticky: done.

Issue #1794, opened 2017-11-02
pbjs.renderAd is not replacing ${AUCTION_PRICE} in adUrl


klawil:
## Type of issue
Bug

## Description
When a bid is returned with `adUrl` instead of `ad`, the auction price is not replaced in `adUrl`.

## Steps to reproduce
1. Create a bid with an `adUrl` parameter (used to create an iframe with source `adUrl` instead of rendering HTML)
2. Render that ad using `pbjs.renderAd`

## Test page
http://jsfiddle.net/bz3aL37x/6/

### Expected results
The `adUrl` should have the `${AUCTION_PRICE}` replaced with `1` and be set as the source of the iframe.

### Actual results
The `url` parameter has the `${AUCTION_PRICE}` replaced with `1` while `adUrl` is used as the source of the iframe.

## Platform details
All browsers, verified in Prebid v0.32.0

## Other information
PR to fix the issue is being created and will be linked here when submitted
_Edit: Pull request is #1795_

mkendall07:
this is resolved.

PR #1795, opened 2017-11-02
Update renderAd to replace ${AUCTION_PRICE} in adUrl

klawil:
## Type of change

- [x] Bugfix
- [ ] Feature
- [ ] New bidder adapter   
- [ ] Build related changes
- [ ] CI related changes
- [ ] Other

## Description of change
This PR updates the `renderAd` function to replace `${AUCTION_PRICE}` in `adUrl`.

In addition to this, this PR uses `adUrl` for the `url` parameter to match with the key used in the bid object (if this should be removed or needs to be a separate PR, let me know)

## Other information
Associated Issue: #1794

### DIFF
--- a/src/prebid.js
+++ b/src/prebid.js
@@ -289,7 +289,7 @@ $$PREBID_GLOBAL$$.renderAd = function (doc, id) {
         // emit 'bid won' event here
         events.emit(BID_WON, bid);
 
-        const { height, width, ad, mediaType, adUrl: url, renderer } = bid;
+        const { height, width, ad, mediaType, adUrl, renderer } = bid;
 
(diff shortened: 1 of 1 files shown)


ckbo3hrk:
Is it legit to substitute ${AUCTION_PRICE} with ad's own price? Up till now I thought that prebid doesn't support 2nd price auction so sending original price back to DSP has no sense as in general DSP expects to be charged with lesser price (and that lesser price is passed via ${AUCTION_PRICE}) than it bids in case of 2nd price auction.

mkendall07:
Correct, there is no 2nd price but this gives us flexibility to support it if we want to. Nice catch @klawil

PR #1796, opened 2017-11-02
Sovrn 1.0 compliance

tedrand:
## Type of change

- [ ] Bugfix
- [ ] Feature
- [ ] New bidder adapter   
- [ ] Build related changes
- [ ] CI related changes
- [x] Other

## Description of change

Adjustments to Sovrn bid adapter to ensure compliance with Prebid 1.0.0.


tedrand:
Thanks, I'll get that put together. As far as the test ad units that returns test creatives, does this need to be able to serve creatives on any site?

modules/sovrnBidAdapter.js
```diff
@@ -1,156 +1,83 @@
...
+    let sovrnBidResponses = [];
+    if (id &&
+      seatbid &&
+      seatbid.length > 0 &&
+      seatbid[0].bid &&
+      seatbid[0].bid.length > 0) {
+      seatbid[0].bid.map(sovrnBid => {
+        sovrnBidResponses.push({
+          requestId: sovrnBid.impid,
+          bidderCode: spec.code,
```
matthewlane: `bidderCode` will be set automatically by `bidderFactory` now, this line can be dropped

modules/sovrnBidAdapter.js
```diff
@@ -1,156 +1,83 @@
...
+      options: {contentType: 'text/plain'}
+    };
+  },
+
+  /**
+   * Format Sovrn responses as Prebid bid responses
+   * @param {id, seatbid} sovrnResponse A successful response from Sovrn.
+   * @return {Bid[]} An array of formatted bids.
+  */
+  interpretResponse: function({id, seatbid}) {
```
matthewlane: #1748 changed the first argument of `interpretResponse` to:

```JavaScript
{
  body: responseBody,
  headers: {
  }
}
```

so destructuring like
```JavaScript
{body: {id, seatbid}}
```
or however you'd prefer to grab the `body`, and updating corresponding tests should get this back to working properly

src/adapters/bidderFactory.js
```diff
@@ -66,6 +66,7 @@ import { logWarn, logError, parseQueryStringParameters, delayExecution } from 's
  * @property {('GET'|'POST')} method The type of request which this is.
  * @property {string} url The endpoint for the request. For example, "//bids.example.com".
  * @property {string|object} data Data to be sent in the request.
+ * @property {object} options Content-Type set in the header of the bid request, overrides default 'text/plain'.
```
matthewlane: Proposed changes to core modules are welcome but should be made in a separate PR. Also calls from 1.0 bidders will have `contentType: 'text/plain'` set by default, so needing to set `options: {contentType: 'text/plain'}` on the object returned by your `buildRequests` function is unnecessary I believe?

modules/sovrnBidAdapter.js
tedrand: Thanks @matthewlane for the info here. Just so I am on the same page, destructuring as you have suggested will lead to the same functionality as what we currently have implemented? I am just making sure that we do not have to change the bid response's JSON structure on our end.

matthewlane: That's right, updating in this way will give the same functionality as before #1748 was merged to master, and there are no changes required to the JSON structure sent from your endpoint.

For a (non-destructuring) example, appnexusAst was updated [here](https://github.com/prebid/Prebid.js/blob/47019481bb34f4071f85d899a16727d8aae606e0/modules/appnexusAstBidAdapter.js#L90) to pick the response data off the `body` property. Before #1748 we used the first parameter directly. This update lets us keep working properly and no changes were made to our endpoint because behind the scenes it's the same data coming through, just on the `body` property of the first parameter rather than the first parameter itself now.

Let me know if you have more questions, and I'll verify the change works whenever the PR is updated

src/adapters/bidderFactory.js
tedrand: yeah sorry I couldn't figure out why that was happening either.

package-lock.json
```diff
@@ -0,0 +1,16925 @@
+{
```
matthewlane: File should be dropped from PR

package.json
```diff
@@ -43,7 +43,7 @@
     "eslint-plugin-standard": "^3.0.1",
     "faker": "^3.1.0",
     "fs.extra": "^1.3.2",
-    "gulp": "^3.8.7",
+    "gulp": "^3.9.1",
```
matthewlane: We'll need to check the version upgrade in a separate PR

tedrand:
this should be good now, I think

package.json
tedrand: darn sorry! i will roll those back

matthewlane:
@tedrand Can you delete the `package-lock.json` entirely, not just make it an empty file? If you're able to do this within the next three hours that'd be great as we're releasing today, if not I'll do a manual merge to still get it in, but would prefer doing it from GitHub

modules/sovrnBidAdapter.js
```diff
@@ -1,156 +1,82 @@
...
-            w: adW,
-            h: adH
-          },
-          tagid: tagId,
-          bidfloor: bidFloor
-        };
-      sovrnImps.push(imp);
+      sovrnImps.push({
+        id: bid.bidId,
+        banner: { w: 1, h: 1 },
```
Carsair: @tedrand Is this hardcoded as 1x1 because Sovrn knows the size based on the ID?  It seems like the bid.sizes and bid.params.sizes properties will be totally ignored by the adapter.  Is that correct?

@mkendall07 If that’s the case, then https://github.com/prebid/Prebid.js/blob/master/modules/sovrnBidAdapter.md should be updated to omit the sizes fields, right?

PR #1650, opened 2017-10-03
A4G prebid adapter

adilets:
## Type of change

- [ ] Bugfix
- [ ] Feature
- [ x] New bidder adapter https://github.com/prebid/prebid.github.io/pull/333/files
- [ ] Code style update (formatting, local variables)
- [ ] Refactoring (no functional changes, no api changes)
- [ ] Build related changes
- [ ] CI related changes
- [ ] Other

## Description of change

Adding new bidder adapter for A4G

- test parameters for validating bids
```
{
  bidder: 'a4g',
  params: {
    zoneId: 59304,
    deliveryUrl: 'http://dev01.ad4game.com/v1/bid'
}
```

Be sure to test the integration with your adserver using the [Hello World](/integrationExamples/gpt/hello_world.html) sample page.

- contact email of the adapter’s maintainer: adilet@ad4game.com, jaafar@ad4game.com, tech@ad4game.com
- [ ] official adapter submission

PR #1797, opened 2017-11-03
New Ad4Game adapter for Prebid.js 1.0

adilets:
## Type of change

- [ ] Bugfix
- [ ] Feature
- [ x] New bidder adapter   https://github.com/prebid/prebid.github.io/pull/333/files  
- [ ] Build related changes
- [ ] CI related changes
- [ ] Other

## Description of change
Adding new bidder adapter for A4G

```
{
  bidder: 'a4g',
  params: {
    zoneId: 59304,
    deliveryUrl: 'http://dev01.ad4game.com/v1/bid'
}
```

- contact email of the adapter’s maintainer: adilet@ad4game.com, jaafar@ad4game.com, tech@ad4game.com

### DIFF
--- a/modules/a4gBidAdapter.js
+++ b/modules/a4gBidAdapter.js
@@ -1,0 +1,78 @@
+import {registerBidder} from 'src/adapters/bidderFactory';
+import * as utils from 'src/utils';
+
+const A4G_BIDDER_CODE = 'a4g';
+const A4G_CURRENCY = 'USD';
+const A4G_DEFAULT_BID_URL = '//ads.ad4game.com/v1/bid';
+const A4G_TTL = 120;
+
+const LOCATION_PARAM_NAME = 'siteurl';
+const ID_PARAM_NAME = 'id';
+const IFRAME_PARAM_NAME = 'if';
+const ZONE_ID_PARAM_NAME = 'zoneId';
+const SIZE_PARAM_NAME = 'size';
+
+const ARRAY_PARAM_SEPARATOR = ';';
+const ARRAY_SIZE_SEPARATOR = ',';
+const SIZE_SEPARATOR = 'x';
+
+export const spec = {
+  code: A4G_BIDDER_CODE,
+  isBidRequestValid: function(bid) {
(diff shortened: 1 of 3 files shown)


adilets:
Hi @matthewlane. I created new PR from (https://github.com/prebid/Prebid.js/pull/1650). 

About response, you should use real domain instead of localhost when you testing. In my case I use `xmmorpg.com` domain. I rewrite localhost to this domain. 

Thank you

dbemiller:
> About response, you should use real domain instead of localhost when you testing. In my case I use xmmorpg.com domain. I rewrite localhost to this domain.

hey @adilets, could you add a note about this in your `.md` file? It's nice to have more permanent documentation about caveats like this when we re-test 6 months from now.

modules/a4gBidAdapter.js
```diff
@@ -0,0 +1,78 @@
...
+      zoneIds.push(bid.params.zoneId);
+    });
+
+    if (!deliveryUrl) {
+      deliveryUrl = A4G_DEFAULT_BID_URL;
+    }
+
+    return {
+      method: 'GET',
+      url: deliveryUrl,
```
matthewlane: One way to do it is to add `bidId` to this object, and set it's value to the bid id of `validBidRequests` (`bidId: validBidRequests[0] && validBidRequests[0].bidId` in that case). Then `bidId` will be available on the `request` parameter in `interpretResponse` and line 61 will work without modification. You can do it however you'd prefer though

modules/a4gBidAdapter.md
```diff
@@ -0,0 +1,40 @@
...
+                    bidder: 'a4g',
+                    params: {
+                        zoneId: 59354,
+                        deliveryUrl: 'http://dev01.ad4game.com/v1/bid'
+                    }
+               }
+           ]
+       }
+    ];
+```
```
matthewlane: Please add a note about rewriting localhost to the test domain in order to receive bids

adilets:
@matthewlane Thanks for your review and comments. I updated PR.

PR #1806, opened 2017-11-07
HuddledMasses header bidding adapter

HuddledMasses:
## Type of change

- [x] New bidder adapter 

## Description of change
Add adapter and tests for the HuddledMasses header bidding adapter

```
{
   bidder: 'huddledmasses',
   params: {
      placement_id: 0
    }
}
```
- contact email of the adapter’s maintainer: supply@huddledmasses.com


mkendall07:
@HuddledMasses 
Why did you rename the previous adapter and now submitting a new one? This is confusing.

HuddledMasses:
@mkendall07 
sorry fo confusion. at last minute we made a decision to keep two adapters; so we will have one for Colossus, another for Huddled Masses

mkendall07:
@HuddledMasses 
And the code for the adapters is discrete (and the endpoints?)

HuddledMasses:
@mkendall07 
yes, both the code and the endpoints are discrete.

HuddledMasses:
Hi @mkendall07 Any update on this? will appreciate. Thanks!

modules/huddledmassesBidAdapter.js
```diff
@@ -0,0 +1,156 @@
...
+
+  /**
+   * Determines whether or not the given bid request is valid.
+   *
+   * @param {object} bid The bid to validate.
+   * @return boolean True if this is a valid bid, and false otherwise.
+   */
+  isBidRequestValid: (bid) => {
+    return (!isNaN(bid.params.placement_id) &&
+    ((bid.params.sizes != undefined && bid.params.sizes.length > 0 && bid.params.sizes.some((sizeIndex) => INDEX_SIZE[sizeIndex] != undefined)) ||
```
mkendall07: we prefer triple equality (!==).

modules/huddledmassesBidAdapter.js
```diff
@@ -0,0 +1,156 @@
...
+  '970x310': 94,
+  '970x210': 96,
+  '480x320': 101,
+  '768x1024': 102,
+  '1000x300': 113,
+  '320x100': 117,
+  '800x250': 118,
+  '200x600': 119
+};
+const INDEX_SIZE = {
```
mkendall07: You don't need to duplicate this data. You can do a reverse lookup by iterating the object properties.

modules/huddledmassesBidAdapter.js
```diff
@@ -0,0 +1,156 @@
...
+   */
+  buildRequests: (validBidRequests) => {
+    let winTop = window;
+    try {
+      window.top.location.toString();
+      winTop = window.top;
+    } catch (e) {
+      console.log(e);
+    };
+    let location = winTop.location;
```
mkendall07: You can use `utils.getTopWindowLocation()` instead of the code above: https://github.com/prebid/Prebid.js/blob/master/src/utils.js#L157

modules/huddledmassesBidAdapter.js
```diff
@@ -0,0 +1,156 @@
...
+            resItem.ad && typeof resItem.ad === 'string' &&
+            resItem.ttl && !isNaN(resItem.ttl) &&
+            resItem.creativeId && typeof resItem.creativeId === 'string' &&
+            resItem.netRevenue && typeof resItem.netRevenue === 'boolean' &&
+            resItem.currency && typeof resItem.currency === 'string') {
+          response.push(resItem);
+        }
+      }
+    } catch (e) {
+      console.log(e);
```
mkendall07: please use `utils.logMessage()` instead

modules/huddledmassesBidAdapter.md
```diff
@@ -0,0 +1,26 @@
+# Overview
+
+```
+Module Name: HuddledMasses Bidder Adapter
+Module Type: Bidder Adapter
+Maintainer: prebid@example.com
```
mkendall07: Please update email

modules/huddledmassesBidAdapter.js
```diff
@@ -0,0 +1,156 @@
...
+  /**
+   * Unpack the response from the server into a list of bids.
+   *
+   * @param {*} serverResponse A successful response from the server.
+   * @return {Bid[]} An array of bids which were nested inside the server.
+   */
+  interpretResponse: (serverResponse) => {
+    let response = [];
+    try {
+      serverResponse = serverResponse.body;
```
mkendall07: this line needs to be `serverResponse.body.body;`. Otherwise no bids are returned.

HuddledMasses: Here I do not agree.
Example response:
``````
{
    "body": [{
        "width": 300,
        "height": 250,
        "ad": "<img src='http://supply.huddledmasses.com/img/logo.png' style='position: absolute; left: 50%; top:50%; margin-left: -113px; margin-top:-79px;'>",
        "requestId": "26042378f97b21",
        "cpm": 5,
        "ttl": 120,
        "creativeId": "123",
        "netRevenue": true,
        "currency": "USD",
        "dealId": "HASH"
    }]
}
``````

HuddledMasses:
All rights are included.
Only the function interpretResponse remained the same.
The comment with an example of a response from the server is left above.

modules/huddledmassesBidAdapter.js
mkendall07: it's broke. Run the code

HuddledMasses: I cant understand whats wrong. 
Why doesn't it work for you.
Here is the test page with the given adapter:
http://hb-test.huddledmassessupply.com/
I do not understand why it does not work for you.

mkendall07: @HuddledMasses 
The API changed. Please rebase to master. 

On your test page replace http://hb-test.huddledmassessupply.com/prebid-core.js with the latest code and you will see the issue.

mkendall07: Here is the code changed: https://github.com/prebid/Prebid.js/blob/master/src/adapters/bidderFactory.js#L285

HuddledMasses: I replaced the response at the server level.
That would not write in the code a long expression serverResponse.body.body. 
```````````
[{
        "width": 300,
        "height": 250,
        "ad": "<img src='http://supply.huddledmasses.com/img/logo.png' style='position: absolute; left: 50%; top:50%; margin-left: -113px; margin-top:-79px;'>",
        "requestId": "26042378f97b21",
        "cpm": 5,
        "ttl": 120,
        "creativeId": "123",
        "netRevenue": true,
        "currency": "USD",
        "dealId": "HASH"
}]
```````````

PR #1810, opened 2017-11-08
Unit test fixes for IE 10 and other old browsers

jaiminpanchal27:
## Type of change
- [x] Bugfix

## Description of change
Some tests were failing in some browsers after running test suite in Browserstack

- Fixed unit tests failing in IE and old Safari browsers.
- Also added polyfill for window.location.origin to support IE10.

### DIFF
--- a/modules/nanointeractiveBidAdapter.js
+++ b/modules/nanointeractiveBidAdapter.js
@@ -55,7 +55,7 @@ function createSingleBidRequest(bid) {
     [NQ]: [createNqParam(bid), createCategoryParam(bid)],
     sizes: bid.sizes.map(value => value[0] + 'x' + value[1]),
     bidId: bid.bidId,
-    cors: location.origin
+    cors: utils.getOrigin()
   };
 }
 
--- a/src/polyfill.js
+++ b/src/polyfill.js
@@ -12,8 +12,3 @@ Number.isInteger = Number.isInteger || function(value) {
     isFinite(value) &&
     Math.floor(value) === value;
 };
-
-// IE10 does not have this propery. https://gist.github.com/hbogs/7908703
-if (!window.location.origin) {
-  window.location.origin = window.location.protocol + '//' + window.location.hostname + (window.location.port ? ':' + window.location.port : '');
-}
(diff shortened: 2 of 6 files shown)


snapwich:
I didn't realize we were doing global polyfills...  I actually am kind of against global polyfills of any kind, including innocuous ones like this.  In the past I've had global polyfills backfire on several occasions causing obscure issues on sites that may have other global polyfills that conflict, or sites that do feature detection and then make certain assumptions.  Even though this seems harmless, for the reasons listed, I don't think we should be doing any of these things that modify the global environment, including the other polyfills listed in this file that modify global objects.

mkendall07:
I don't like polyfills either but given these already exist today it would be difficult to unwind them from the code. I don't know what the right answer is here, but I'd argue that getting stable tests working is worth more than potential edge case issues with polys.

snapwich:
I agree.  I'd argue that rather than add a global polyfill for `window.location.origin` we just stub the few locations we're using it with a helper function instead like: `util.getOrigin()` that does the `window.location.protocol + '//' + window.location.hostname + (window.location.port ? ':' + window.location.port : '')`.  Then we don't need to polyfill.

As for the other polyfills, I think they're more problematic.  They seem to modify the global Array object, which in the past, has specifically caused many obscure bugs for me in the past in a similar situation (writing third-party code that had to interop w/ code already present on arbitrary pages out of our control).  I know in the past some of the bugs were related to the prototype.js library which, unfortunately, does something very similar.  Some were related to XHR libraries that patch `.toJSON` and had certain expectations.  Some bugs were related to feature detection and assumptions.  Overall, just a lot of headache; and that's just the _known_ headaches.  The bugs were so obscure I'm sure there's plenty we didn't catch.  The solutions was to just stop modifying the global environment.

mkendall07:
Yeah I can get on board with that. @jaiminpanchal27 can you update the PR ?

jaiminpanchal27:
Updated PR. Added method to utils.
I have also updated renderer_spec tests to use `beforeEach` hooks since IE 10 failed [this test](https://github.com/prebid/Prebid.js/blob/master/test/spec/renderer_spec.js#L57) randomly

PR #1814, opened 2017-11-08
Improve calls to addBidToAuction and tryAddVideoBid

kitwestneat:
## Type of change

## Description of change
addBidToAuction does not take any parameters, but calls to addBidToAuction are passing in the bid object. This patch modifies calls to reflect the fact that addBidToAuction uses the bid object from the closure instead of taking a parameter. It also modified calls to tryAddVideoBid to use the same convention.

### DIFF
--- a/src/bidmanager.js
+++ b/src/bidmanager.js
@@ -88,9 +88,9 @@ exports.addBidResponse = createHook('asyncSeries', function (adUnitCode, bid) {
     prepareBidForAuction();
 
     if (bid.mediaType === 'video') {
-      tryAddVideoBid(bid);
+      tryAddVideoBid();
     } else {
-      addBidToAuction(bid);
+      addBidToAuction();
       doCallbacksIfNeeded();
     }
   }
@@ -232,7 +232,7 @@ exports.addBidResponse = createHook('asyncSeries', function (adUnitCode, bid) {
   }
 
   // Video bids may fail if the cache is down, or there's trouble on the network.
-  function tryAddVideoBid(bid) {
+  function tryAddVideoBid() {
     if (config.getConfig('usePrebidCache')) {
       store([bid], function(error, cacheIds) {
         if (error) {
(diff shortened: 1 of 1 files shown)


mkendall07:
Leaving the `bid` argument improves readability of the code and should not have any side effects. What's the reasoning for doing this?

kitwestneat:
It seems really wrong to call a function that takes no arguments with an argument. When I see a function call with an argument, I assume that it takes an argument. I guess I don't see how lying about it improves readability. If that's the preferred style of the project, ok then, but it just seemed like a mistake when I saw it.

mkendall07:
Well it's not lying if you pass `bid` to the function and then operate on `bid`. The object in the closure scope happens to match what is in the function signature. You could rename the param to something like `function addBidToAuction(bidObj)` and then use it like `addBidToAuction(bid)`.

The reason I suggested leaving it is because it's confusing to me to look at code like this:
`function addBidToAuction()` // w/ no args

So I'd say leave it or do the suggestion above. I actually prefer not using closure scope for functions and placing them at the top level. This also helps with unit testing if needed. 

Appreciate you trying to fix what was a received issue though. Thanks!

kitwestneat:
But it's not actually passing it, right? The function call implies that the function is operating on what is being passed, but when in fact the function is not literally operating on what is passed. It's operating on the closure variable, which happens to be the same in this case, sure. But when I see a function call with a parameter, I assume that if I change what is being passed, I will change what is being operated on. Do you see where I'm coming from?

Here's an example of a case where it might cause an issue in the future. Let say that we need to modify the bids at some point before it's added, and we want to keep the old bid around for record keeping. I could see someone doing:
let newBid = Object.assign({ newParam: newValue }, bid);
addBidToAuction(newBid);

And then they are confused why it doesn't work.

Anyway, that's my 2c, I'll go do something more productive now. ;)

mkendall07:
well if that's the case yeah. I'm gonna look at the code now :)

mkendall07:
ok I missed this... https://github.com/prebid/Prebid.js/blob/master/src/bidmanager.js#L220
This *should* be `function addBidToAuction(bid)`.
Sorry man. I'll accept a PR with that change :)

kitwestneat:
ok cool, that will make me a lot less anxious, haha

dbemiller:
whoa... good catch. My bad on the confusing implementation in the first place :/

PR #1815, opened 2017-11-08
Drop non-video bidders from video ad units

### LINKED ISSUE #1137: Bids requests failing if bidder missing
## Description
When settings the videoAdUnit parameters, if one of the bidder is not present in the Prebid.js file, it will make all the request crash... I have tested with the display and if a bidder is not present, other calls are still and correctly made.

## Steps to reproduce
Please reach the following page where some random Appnexus and Rubicon parameters. The prebid.js has been built with only Rubicon adapter. As you will see no requests are made:
http://legba.alwaysdata.net/VideoR.html

I have used exactly the same code with a prebid.js built this time with Rubicon and Appnexus adapters. As you will see this time everything is ok:
http://legba.alwaysdata.net/VideoRA.html

### Expected results
If a bidder is not present in the prebid.js tag, it should skip it but still make the call for the others bidders.

### Actual results
All requests are skipped.

## Platform details

The last version 0.21.0

matthewlane:
## Type of change
- Bugfix

## Description of change
Ad units with `mediaType: 'video'` that included non-video bidders were preventing calls to all bidders in the ad unit, even if that included video bidders. This change drops non-video bidders and still calls valid video bidders rather than dropping the ad unit entirely. Works for ad units configured with `mediaTypes: 'video'` or 1.0-compatible `meidaTypes: {video: {}}`

In this example ad unit, on master `appnexusAst` will not get called because `fakeBidder` is not a video bidder, or even real for that matter. With this PR `appnexusAst` gets called.

```JavaScript
pbjs.addAdUnits({
  code: '/19968336/prebid_video_adunit',
  sizes: [640,480],
  mediaType: 'video',
  // or mediaTypes: { video: {context: 'instream'} },
  bids: [

    {
      bidder: 'appnexusAst',
      params: {
        placementId: '9333431',
        video: { skippable: true, playback_method: ['auto_play_sound_off'] }
      }
    },

    {
      bidder: 'fakeBidder',
      params: { tag: 'stopHammerTime' }
    },

  ]
});
```

## Other information
Fixes #1137

### DIFF
--- a/src/prebid.js
+++ b/src/prebid.js
@@ -383,7 +383,7 @@ $$PREBID_GLOBAL$$.requestBids = function ({ bidsBackHandler, timeout, adUnits, a
   // for video-enabled adUnits, only request bids if all bidders support video
   const invalidVideoAdUnits = adUnits.filter(videoAdUnit).filter(hasNonVideoBidder);
   invalidVideoAdUnits.forEach(adUnit => {
-    utils.logError(`adUnit ${adUnit.code} has 'mediaType' set to 'video' but contains a bidder that doesn't support video. No Prebid demand requests will be triggered for this adUnit.`);
+    utils.logError(`adUnit ${adUnit.code} is of type 'video' but contains a bidder that doesn't support video. No Prebid demand requests will be triggered for this adUnit.`);
     for (let i = 0; i < adUnits.length; i++) {
       if (adUnits[i].code === adUnit.code) { adUnits.splice(i, 1); }
     }
--- a/src/prebid.js
+++ b/src/prebid.js
@@ -2,7 +2,7 @@
 
 import { getGlobal } from './prebidGlobal';
 import { flatten, uniques, isGptPubadsDefined, adUnitsFilter } from './utils';
-import { videoAdUnit, hasNonVideoBidder } from './video';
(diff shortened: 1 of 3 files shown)


src/prebid.js
```diff
@@ -383,7 +383,7 @@ $$PREBID_GLOBAL$$.requestBids = function ({ bidsBackHandler, timeout, adUnits, a
   // for video-enabled adUnits, only request bids if all bidders support video
   const invalidVideoAdUnits = adUnits.filter(videoAdUnit).filter(hasNonVideoBidder);
   invalidVideoAdUnits.forEach(adUnit => {
-    utils.logError(`adUnit ${adUnit.code} has 'mediaType' set to 'video' but contains a bidder that doesn't support video. No Prebid demand requests will be triggered for this adUnit.`);
+    utils.logError(`adUnit ${adUnit.code} is of type 'video' but contains a bidder that doesn't support video. No Prebid demand requests will be triggered for this adUnit.`);
```
mkendall07: I'd prefer that we call the demand that is enabled for video instead of just dropping it completely. Is it hard to make that change?

matthewlane: Yeah that makes sense, I can do it. I'll update the PR to do that when either `mediaType` or `mediaTypes` is used to configure 'video'

PR #1819, opened 2017-11-09
ServerBid Server BidAdapter

jcswart:
Allow S2S configuration with ServerBid.

## Type of change

- [ ] Bugfix
- [ ] Feature
- [x] New **SERVER TO SERVER** bidder adapter   
- [ ] Build related changes
- [ ] CI related changes
- [ ] Other

## Description of change

Add S2S adapter: ServerBidServerBidAdapter.

    - jcswart@serverbid.com

### DIFF
--- a/modules/serverbidServerBidAdapter.js
+++ b/modules/serverbidServerBidAdapter.js
@@ -1,12 +1,23 @@
 import Adapter from 'src/adapter';
 import bidfactory from 'src/bidfactory';
-import bidmanager from 'src/bidmanager';
 import * as utils from 'src/utils';
 import { ajax } from 'src/ajax';
 import adaptermanager from 'src/adaptermanager';
 import { STATUS, S2S } from 'src/constants';
+import { config } from 'src/config';
 
 const TYPE = S2S.SRC;
+const getConfig = config.getConfig;
+const REQUIRED_S2S_CONFIG_KEYS = ['siteId', 'networkId', 'bidders', 'endpoint'];
+
+let _s2sConfig;
+config.setDefaults({
+  's2sConfig': {
+    enabled: false,
+    timeout: 1000,
+    adapter: 'serverbidServer'
+  }
+});
 
 var ServerBidServerAdapter;
 ServerBidServerAdapter = function ServerBidServerAdapter() {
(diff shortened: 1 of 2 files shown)


mkendall07:
@jcswart 
Thanks for the clarification. We'll review as S2S. FYI we are currently re-writing the prebidServer adpater to conform to the prebid 1.0 requirements so we might ask for a rebase on that pattern - i'll let you know if that is the case.

jcswart:
@mkendall07 Thanks!

dbemiller:
@jgrimes No need to refactor this to the new pattern. We did some testing and found that the `bidderFactory` doesn't work very well for server-to-server.

We might come up with a nice common interface to simplify s2s adapter creation someday... but it'll probably take a few more submissions before we can do that effectively.

jcswart:
Any word on how far a long this is in review?

dbemiller:
yeah... sorry about that. I've been pretty irresponsible on this one.

The one major issue jumping out at me is that the `callBids` function API changed in prebid-1.0. The `bidderFactory` pattern hid these changes from other adapters... but since this is an s2s adapter, you've got to upgrade it for 1.0 specifically.

The big difference is that `callBids(bidRequest)` is now `callBids(s2sBidRequest, bidRequests, addBidResponse, done, ajax)`. The `bidmanager` doesn't even exist. Instead, call `addBidResponse(bid)` to add a bid, and `done()` after you've finished adding all bids.

Check out `modules/prebidServerBidAdapter.js` for a working example. You probably used that as a model for writing this in the first place... but the 1.0 merge brought in changes so that it stays compatible with the new auctions.

dbemiller:
Another thing which will help speed things along if you could post a test page which you know fetches bids from your adapter.

On my own, I'd try to test it by starting from `integrationExamples/gpt/prebidServer_example.html` and then changing some options in the `s2sConfig` object... but if that's not how your adapter is intended to work, it can take some back and forth to prove that it works.

That process is much more likely to go quickly if you can include a test page which shows me how you expect it to be used.

mkendall07:
@jcswart @jgrimes 
Just a ping on this that it needs updates. Thanks!

PR #1823, opened 2017-11-13
Pulsepoint adapter: fixing bid rejection due to missing mandatory bid params.

anand-venkatraman:
## Type of change
- [x] Bugfix

## Description of change
Last release 0.32, has added validation of bid param keys submitted from the adapter to bidderFactory. And that part, is resulting in bid rejection in Pulsepoint Adapter. This PR fixes it.

- avenkatraman@pulsepoint.com


modules/pulsepointLiteBidAdapter.js
```diff
@@ -10,6 +10,10 @@ const NATIVE_DEFAULTS = {
   ICON_MIN: 50,
 };
 
+const DEFAULT_BID_TTL = 20;
```
mkendall07: this value is in seconds - is that right? Just double checking.

anand-venkatraman: Yes. ttl is in seconds, according to docs here - http://prebid.org/dev-docs/bidder-adapter-1.html.

mkendall07: cool. LGTM.

PR #1826, opened 2017-11-14
emitEvent method accessible from the global pbjs object

benjaminclot:
## Type of change
- [x] Feature

## Description of change
As part of our internal prebid integration, we manually elect a winning bid (due to custom blacklists and floors) and we customize how ads are rendered (e.g. for skins, interstitials...) so we don't call renderAd. Thus, analytics modules don't register the bidWon event.
This proposal solves this problem by allowing us to emit any event from the gobal pbjs object. If required, it could also be tweaked to only allow the bidWon (and others?) event to be "manually" emitted.

### DIFF
--- a/src/prebid.js
+++ b/src/prebid.js
@@ -469,8 +469,7 @@ $$PREBID_GLOBAL$$.addAdUnits = function (adUnitArr) {
 $$PREBID_GLOBAL$$.emitEvent = function (event) {
   utils.logInfo('Invoking $$PREBID_GLOBAL$$.emitEvent', arguments);
   
-  var args = slice.call(arguments, 1);
-  events.emit(event, args);
+  events.emit.apply(this, arguments);
 };
 
 /**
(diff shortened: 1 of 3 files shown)


benjaminclot:
So... where are we on this?

mkendall07:
@benjaminclot 
We'd prefer not to allow this behavior as it introduces the chance for bad data to flow into the downstream consumers of the events like analytics and other hooks. Can you invoke the renderAd on the winning bid?

benjaminclot:
@mkendall07 No, mainly for the reasons mentioned earlier: we also render skins and other specific formats which renderAd doesn't handle. I know it's not a "classic" setup...

mkendall07:
Ideally we'd have renderAd handle these cases as well. We are actively working on Native implementation and some other formats. Can the implementation code you are using be contributed back to the project in some way?

mkendall07:
an example, if it's completely custom, would be to inject a renderer function into the bids themselves. If renderAd is invoked on the bid, it checks first for it's own renderer and uses that instead. This is used for our video outstream implementation. See here for example: https://github.com/prebid/Prebid.js/blob/master/modules/appnexusAstBidAdapter.js#L125

benjaminclot:
Maybe for interstitials but for skins it's a whole internal process with per publisher adapters so I doubt it.
I'll look into the custom renderer, I thought it was only for outstream. Thanks!

mkendall07:
ok sounds good. Open to extending the renderer API for inclusion of your use cases if you want to propose something. I'll close this for now.

Issue #1762, opened 2017-10-24
Tracker extension points for non-usersync trackers


samuelhorwitz:
## Type of issue
Feature request

## Description
While it is not urgent and we are fine with leaving it out for now, our team at Kargo would like to drop a pixel tracker upon receipt of an ad inside Prebid. However doing so is disallowed due to the fact that it would be chaotic and potentially slow to just allow people to add trackers randomly in their adapter code.

It would be nice if certain events had tracker callback hooks or some other solution where you could register that you would like a tracker to fire that isn't tied to user syncing. For example, inside of response handling.

The issue was brought up here initially https://github.com/prebid/Prebid.js/pull/1729#discussion_r146594903

mkendall07:
@samuelhorwitz 
Why not user usersync functionality for this?

optimatic58:
Not sure if this is the correct place to request this feature (we were directed here).  We would simply like to be able to record failure/error logs at our end via ajax call.  This would allow us to proactively address any issues that may be causing ad failure.  Perhaps deferred ajax calls can be supported as suggested by @dbemiller or a way for us to log and review a sample of those errors to troubleshoot any potential issues.

mkendall07:
so the way I look at this is that anything that causes a request to fire that isn't related to fetching demand is unnecessary. So we probably should not have named the functionality "user syncing". It really just means anything that is not a demand fetch. So it's totally appropriate to put other pings into the user syncing queue. Thanks

PR #1837, opened 2017-11-15
+ Add Optimatic Bid Adapter

optimatic58:
## Type of change

- [x] New bidder adapter   

## Description of change

Add Optimatic Bid Adapter

```
{
  bidder: 'optimatic',
  params: {
    placement: '2chy7Gc2eSQL',
    bidfloor: 2.00 
  }
}
```

- contact email of the adapter’s maintainer: prebid@optimatic.com

### DIFF
--- a/modules/optimaticBidAdapter.js
+++ b/modules/optimaticBidAdapter.js
@@ -1,0 +1,128 @@
+import * as utils from 'src/utils';
+import { ajax } from 'src/ajax';
+import { config } from 'src/config';
+import { registerBidder } from 'src/adapters/bidderFactory';
+const LOGURL = 'http://playerlog.optimatic.com/log';
+const VERSION = '7.0.0.1';
+
+let isLogActive = false;
+let rand = Math.round(Math.random() * 10000);
+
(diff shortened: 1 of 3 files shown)


modules/optimaticBidAdapter.js
```diff
@@ -0,0 +1,128 @@
...
+    },
+    device: {
+      ua: global.navigator.userAgent,
+      ip: '127.0.0.1',
+      devicetype: 1
+    }
+  };
+}
+
+function log (field1, field2, field3, field4, pct) {
```
dbemiller: I see what you're trying to do here... but right now Prebid doesn't have a good way to support things like this. Ajax calls are expensive, and will interfere with both the auction and the publisher's page load performance.

Ideally, we'd offer Adapters some way to "defer" work until the critical page loading stuff is done. For now, could you remove this from the code and add your use-case(s) to #1762?

We're trying to collect use cases for now so we can support it in a more general way.

optimatic58: We have made and committed the changes requested.  Please let us know if there is anything else.  Thank you!

PR #1839, opened 2017-11-15
RhythmOne port to prebid 1.0, with unit tests

jstocker76:
## Type of change

- [ ] Bugfix
- [ ] Feature
- [ ] New bidder adapter   
- [ ] Build related changes
- [ ] CI related changes
- [ ] Other

## Description of change
Ported the RhythmOne bidder adapter to the prebid 1.0 interface

```
{
  bidder: 'rhythmone',
  params: 
  { 
    placementId: '411806',
    endpoint: "//tag.1rx.io/rmp/72721/0/mvo?z=1r" // only required for testing.  this api guarantees no 204 responses
  }
}
```

astocker@rhythmone.com


modules/rhythmoneBidAdapter.js
```diff
@@ -1,286 +1,201 @@
...
+  this.supportedMediaTypes = [VIDEO, BANNER];
 
-  function getBidParameters(bids) {
-    for (var i = 0; i < bids.length; i++) {
-      if (typeof bids[i].params === 'object' && bids[i].params.placementId) { return bids[i].params; }
-    }
-    return null;
-  }
+  this.isBidRequestValid = function (bid) {
+    return (bid.adUnitCode !== undefined);
```
matthewlane: `adUnitCode` isn't available on the `bid` parameter here, so this is always failing, but `placementCode` is, if that's what you want to check for

modules/rhythmoneBidAdapter.js
```diff
@@ -1,286 +1,201 @@
...
+        height: bid.h,
+        creativeId: bid.crid,
+        currency: 'USD',
+        netRevenue: true,
+        ttl: 1000
+      };
+
+      if (bidRequest.mediaTypes && bidRequest.mediaTypes.video) {
+        bidResponse.vastUrl = bid.nurl;
+        bidResponse.descriptionUrl = bid.nurl;
```
matthewlane: `descriptionUrl` isn't required to be set anymore, `vastUrl` alone is fine

modules/rhythmoneBidAdapter.js
jstocker76: Are you sure?  I was seeing adUnitCode on the bid requests when this adapter was in the prebid 1.0 branch.  It's also mentioned in http://prebid.org/dev-docs/bidder-adapter-1.html

modules/rhythmoneBidAdapter.js
jstocker76: Thanks, will remove that line

modules/rhythmoneBidAdapter.js
matthewlane: Yeah, I was testing against master, on that branch the param is called `placementCode`. Looks like on 1.0 it is renamed `adUnitCode`. This would get merged into master so the safest thing might be to check for either one

jstocker76: The adapter will use the first placementId that it finds on any of the bid requests if it's not configured individually.  The only requirement is for the configured bids to have at least one placement id.  The rest of the params are optional and exist for testing purposes only.

This is a side effect of our api being a single request architecture, and the url format only allows for one placement id.  We're currently working on a new api that addresses this limitation

jstocker76: Got it.  Another question: what's the state of the master branch?  I was told via an email with some appnexus developers that I should check this 1.0 adapter into master.  Is master in a transitional state?

matthewlane: Yes, both of those are correct, this adapter should be checked into the master branch, and master will soon be transitioned into the [1.0 branch](https://github.com/prebid/Prebid.js/tree/prebid-1.0). Any 1.0-compatible adapter in the (current) master branch will be transitioned/work in the 1.0 branch whenever that transition occurs, and RhythmOne would be included in that category after this PR is merged

jstocker76: Cool, thank you

PR #1847, opened 2017-11-17
Updates for PubWise Prebid 1.0 Support

GLStephen:
Initial Prebid 1.0 Support for PubWise, which supports all existing functionality in Prebid 1.0 as outlined on the [PubWise site](https://pubwise.io/pubwise/).


GLStephen:
I don't see an identifiable error in the travis output. The error looks transient.

PR #1848, opened 2017-11-17
Rename appnexusAst adapter to appnexus adapter

jaiminpanchal27:
## Type of change
- [x] Other

## Description of change
Rename appnexusAst adapter to appnexus adapter.
Updated Unit tests to use `appnexus` bidderCode.


test/spec/unit/pbjs_api_spec.js
```diff
@@ -1046,7 +1056,7 @@ describe('Unit: Prebid Module', function () {
       })
 
       it('should callBids if a native adUnit has all native bidders', () => {
-        // TODO: appnexusAst is currently hardcoded in native.js, update this text when fixed
+        // TODO: appnexus is currently hardcoded in native.js, update this text when fixed
```
mkendall07: is this comment true? if it's possible fix it or drop if it's not true.

jaiminpanchal27: Todo might have been completed long back so dropped the comment

mkendall07:
LGTM. Thanks

Issue #1859, opened 2017-11-21
New OpenX adapter update ends up with no requests


UrTARGET:
## Type of issue

bug?

## Description

Updated to the latest version of prebid (v0.33 with openx adapter update), then number of requests going to openx server dropped massively. (150k requests dropped to less than 10 and decreasing daily)

Settings were left untouched, only updated the prebid version. 

![image](https://user-images.githubusercontent.com/26286258/33079600-0cd2227a-cea4-11e7-83f9-12773cb86123.png)
![image](https://user-images.githubusercontent.com/26286258/33079639-202bfc38-cea4-11e7-97ce-44adf72f94a0.png)
![image](https://user-images.githubusercontent.com/26286258/33079663-309ff3e4-cea4-11e7-8a23-09239173006a.png)

## Platform details

prebid version 0.33, Mozilla/5.0 (Macintosh; Intel Mac OS X 10_13_1) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/62.0.3202.94 Safari/537.36

## Other information

related PR: https://github.com/prebid/Prebid.js/pull/1724

mkendall07:
@IHateCodesss can you take a look?

bwoolcott:
@mkendall07 We've identified the issue here and developed a fix yesterday which is in QA today. We should be able to make a PR with the fix today. The issue was due to mediaType not defaulting to banner in v0.33.0 whereas in 1.0 it does. So, since mediaType wasn't explicitly declared as "banner", requests aren't being created.

UrTARGET:
Do we need to download new prebid.js for the fix? and how long can we expect to see the fix to be effective?

mkendall07:
this is resolved in master now.

PR #1861, opened 2017-11-21
OpenX adapter: Make mediaType default to banner when none is specified

lntho:
## Type of change

- [ x ] Bugfix

## Description of change

Make mediaType default to BANNER when there is no mediaType specified


lntho:
This is related to the issue described in #1859

lntho:
@mkendall07 this is the PR to fix the issue discussed this morning. Thanks!

PR #1866, opened 2017-11-22
Kumma Adapter: size parameter is not mandatory anymore

yehonatanshac:
## Type of change

- [ X] Feature

## Description of change
Size parameter is not mandatory anymore for Kumma adapter

yehonatan@kumma.com

### DIFF
--- a/modules/kummaBidAdapter.js
+++ b/modules/kummaBidAdapter.js
@@ -76,7 +76,7 @@ function impression(slot) {
   };
 }
 function banner(slot) {
-  const size = parseSizesInput(slot.sizes)[0].split('x') || slot.params.size.toUpperCase().split('X');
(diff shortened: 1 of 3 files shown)


modules/kummaBidAdapter.js
```diff
@@ -76,7 +76,7 @@ function impression(slot) {
   };
 }
 function banner(slot) {
-  const size = slot.params.size.toUpperCase().split('X');
+  const size = parseSizesInput(slot.sizes)[0].split('x') || slot.params.size.toUpperCase().split('X');
```
mkendall07: this would never be falesy so the 2nd value would never evaluate. 

If that's ok then drop it. otherwise you should handle the other cases properly.

yehonatanshac:
Now the size parameter was completely removed

PR #1875, opened 2017-11-23
Pass the adUnit array when emitting SET_TARGETING

ptomasroos:
We use this in our analytics adapter very useful to have as args

- [x] Feature

## Description of change

Since the auction process is detached from the DFP rendering process aka when does a call against DFP happen we find it very useful in our analytics adapter to have which ad unit's or all are getting invoked

### DIFF
--- a/src/prebid.js
+++ b/src/prebid.js
@@ -230,7 +230,7 @@ $$PREBID_GLOBAL$$.setTargetingForGPTAsync = function (adUnit) {
   targeting.setTargeting(targetingSet);
 
   // emit event
-  events.emit(SET_TARGETING);
+  events.emit(SET_TARGETING, adUnit);
 };
 
 /**


src/prebid.js
```diff
@@ -230,7 +230,7 @@ $$PREBID_GLOBAL$$.setTargetingForGPTAsync = function (adUnit) {
   targeting.setTargeting(targetingSet);
 
   // emit event
-  events.emit(SET_TARGETING);
+  events.emit(SET_TARGETING, adUnit);
```
mkendall07: adUnit is optional here. Probably better to define some structure. Is this an array, object etc?

mkendall07: Once that is done, it should be applied to the other SET_TARGETING emitter - see line 250

mkendall07: Just to be clear, my preference is:

events.emit(SET_TARGETING, { adUnits: [adUnit1, adUnitN...]});

ptomasroos: Thanks for the feedback. Will fix!

mkendall07:
@ptomasroos 
any chance to work on this?

ptomasroos:
Yes will fix tomorrow GMT time @mkendall07 forgot

bretg:
@ptomasroos - any update?

mkendall07:
closing as no update in several months.

PR #1876, opened 2017-11-23
Allow specific gptSlot targeting in setTargetingForGPTAsync

ptomasroos:
## Type of change
- [x] Bug
- [x] Feature 

## Description of change
We use the following flow:

We start a auction as early as possible. Without knowing where in the feed we're gonna place it. So depending on the outcome of the auction and how long it took we might inject it into a placement defined further down the page in order to now burn views above the fold.

The current semantics of setTargeting requires that we shall know a head of time what gptSlot the targeting should apply for because its looking into the auction object.

But allowing us to pass in an explicit gptSlot into the function we can allow the targeting to explicitly use this slot when setting targeting

We use this in order not to place ads above fold and insert them to a better position instead of everything just being very static.

### DIFF
--- a/src/prebid.js
+++ b/src/prebid.js
@@ -213,7 +213,7 @@ $$PREBID_GLOBAL$$.getBidResponsesForAdUnitCode = function (adUnitCode) {
  * @param {(string|string[])} adUnit a single `adUnit.code` or multiple.
  * @alias module:pbjs.setTargetingForGPTAsync
  */
-$$PREBID_GLOBAL$$.setTargetingForGPTAsync = function (adUnit) {
+$$PREBID_GLOBAL$$.setTargetingForGPTAsync = function (adUnit, gptSlot) {
   utils.logInfo('Invoking $$PREBID_GLOBAL$$.setTargetingForGPTAsync', arguments);
   if (!isGptPubadsDefined()) {
     utils.logError('window.googletag is not defined on the page');
@@ -227,7 +227,7 @@ $$PREBID_GLOBAL$$.setTargetingForGPTAsync = function (adUnit) {
   targeting.resetPresetTargeting(adUnit);
 
   // now set new targeting keys
-  targeting.setTargeting(targetingSet);
+  targeting.setTargeting(targetingSet, gptSlot);
 
   // emit event
   events.emit(SET_TARGETING);
--- a/src/targeting.js
+++ b/src/targeting.js
@@ -48,10 +48,10 @@ targeting.getAllTargeting = function(adUnitCode) {
   return targeting;
 };
 
-targeting.setTargeting = function(targetingConfig) {
+targeting.setTargeting = function(targetingConfig, gptSlot) {
   window.googletag.pubads().getSlots().forEach(slot => {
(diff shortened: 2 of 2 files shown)


ptomasroos:
FATAL ERROR: CALL_AND_RETRY_LAST Allocation failed - JavaScript heap out of memory

mkendall07:
This is basically the infinite scroll scenario. You fetch demand for a yet to be defined placement. 

We discussed this in depth without coming to any particular solution on the best way to solve. With that said this approach isn't terrible. It needs to be beefed up a bit in the code and also should be made against prebid-1.0 branch since there are many changes there. 

@prebid/core for thoughts.

mkendall07:
@ptomasroos 
Can you update this to match slot or DIV id? That's traditionally how we have matched on DFP. Also would it make sense to support an array of slot or div ids here?

ptomasroos:
Of course. I will make gptSlot optional, that hard part if we accept a array is to find what it should be correlated against? Then we need to pass a object like { auctionCode: gptSlot, auctionCode2: gptSlopt2 } 

But again you let me what you think fulfils the needs here. We only use one at the moment but I'm pretty sure I can stab out multiple as well if needed

mkendall07:
yeah that's the problem here. The function input currently is ` @param {(string|string[])} adUnit a single `adUnit.code` or multiple.`
So if you pass `[code1, code2]`, `gptCode` - which slot do you set targeting on?

ptomasroos:
@mkendall07 Indeed! We only have one at a time. So yes I agree that this would require a different signature in order to be better suited for this.

What we're currently using.
window.pbjs.setTargetingForGPTAsync([adUnitCode], nextSlotToBeDisplayed);

So if thats the case we could make it map to the same position as the adUnitCode.

window.pbjs.setTargetingForGPTAsync([adUnitCode1, adUnitCode2], [slotForAdUnitCode1, slotForAdUnitCode2]);
?

mkendall07:
That solution is not ideal (depending on the ordering of arrays for mapping).

can you explain a little bit more about your use case? For example:
`window.pbjs.setTargetingForGPTAsync([adUnitCode], nextSlotToBeDisplayed);`

Why is `adUnitCode` not the right slot?

ptomasroos:
Because I have a generated named in [adUnitCode] which is a reference to the auction.

And my slot is drawn between every entry of lets say a list of users.
Which then gets a sequence number.

So depending on how long it takes for the auction to finish and where the user is in the scroll position / view port, the random number wants to target either div-1, div-3, div-5, div-7 and so on.

Does that make sense? And I dont wanna update the dom just for the sake of updating it.

ptomasroos:
Any opinons @mkendall07 ?

mkendall07:
I guess I'm just not sure the best way forward on this. Does prebid 1.0 concurrent auctions help?

ptomasroos:
No because if one wants to avoid to burn ads above the fold aka making sure to always render where the user is / going, we can't decide the place where the ad should be inserted at the time the auction is started. At least thats our philosophy

mkendall07:
Going to close this since I'm not sure the best solution here. feel free to open an issue to discuss the general use case and talk about different approaches.

PR #1896, opened 2017-11-29
Update auctionId parameter to be requestId

Pupis:
## Type of change

- [x] Bugfix

## Description of change
RequestId must be used as auctionId.

Scope.FL.Scripts@adform.com

## Other information
Changing value of parameter, which was introduced in this [`PR`].(https://github.com/prebid/Prebid.js/pull/1888)

### DIFF
--- a/modules/adformBidAdapter.js
+++ b/modules/adformBidAdapter.js
@@ -37,7 +37,7 @@ function AdformAdapter() {
 
     request.unshift('//' + globalParams[0][1] + '/adx/?rp=4');
 
-    request.push('auctionId=' + encodeURIComponent(params.bidderRequestId));
+    request.push('auctionId=' + params.requestId);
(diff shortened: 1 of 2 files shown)


modules/adformBidAdapter.js
```diff
@@ -37,7 +37,7 @@ function AdformAdapter() {
 
     request.unshift('//' + globalParams[0][1] + '/adx/?rp=4');
 
-    request.push('auctionId=' + encodeURIComponent(params.bidderRequestId));
+    request.push('auctionId=' + params.requestId);
```
mkendall07: why was the `encodeURIComponent` dropped?

Pupis: I believe it is not necessary, because requestId is UUID.

PR #1897, opened 2017-11-29
Bug fix to populate userSync default values

### LINKED ISSUE #1883: bidderFactory sends wrong sync params to the bidders
## Type of issue
Bug

## Description
I tried to enable user sync through frames so I've added the following setConfig call:
```
pbjs.setConfig({ userSync: {
    iframeEnabled: true
}});
```
The "userSync" module combines default setting with those passed in setConfig. If the user hasn't specified some option, it would retain its default value. 
```
config.getConfig('userSync', (conf) => {
    usConfig = Object.assign(usConfig, conf.userSync);
});
```
But the "bidderFactory" does not store the initial settings, instead, it asks the "config" for the current version of those settings.
```
let syncs = spec.getUserSyncs({
    iframeEnabled: config.getConfig('userSync.iframeEnabled'),
    pixelEnabled: config.getConfig('userSync.pixelEnabled'),
}, responses);
```
Config modules does not apply the default settings. And if the user hasn't specified some option, it would be undefined and undefined is treated as false. Upon the first call to "setConfig" the default values are discarded. In my case, pixelEnabled became undefined. So enabling iframe sync disabled pixel sync.

I'd say that the handling of the default values should be done in the config.

## Steps to reproduce
1. Add the following funtion call at the init `pbjs.setConfig({ userSync: {iframeEnabled: true}});`
2. Run an auction with some bidders that use pixel sync (conversant, trustX)

### Expected results
Sync pixels for those bidders are fired

### Actual results
Sync pixels for those bidders are not fired

## Platform details
pbjs v 0.33

jaiminpanchal27:
## Type of change
- [x] Bugfix

## Description of change
This PR fixes #1883 
It also includes `config.setDefaults` added by @snapwich in [Prebid Server Adapter committ](https://github.com/prebid/Prebid.js/pull/1846/commits/09089febaee1d1d6fd36b70c8623d9d535720ba1)

### DIFF
--- a/src/config.js
+++ b/src/config.js
@@ -15,12 +15,6 @@ const DEFAULT_BIDDER_TIMEOUT = 3000;
 const DEFAULT_PUBLISHER_DOMAIN = window.location.origin;
 const DEFAULT_COOKIESYNC_DELAY = 100;
 const DEFAULT_ENABLE_SEND_ALL_BIDS = false;
-const DEFAULT_USERSYNC = {
-  syncEnabled: true,
-  pixelEnabled: true,
-  syncsPerBidder: 5,
-  syncDelay: 3000
-};
 
 const GRANULARITY_OPTIONS = {
   LOW: 'low',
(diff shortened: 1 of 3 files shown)


mkendall07:
@snapwich tests are failing

mkendall07:
I agree

PR #1906, opened 2017-11-30
video bug fix

jaiminpanchal27:
## Type of change
- [x] Bugfix

## Description of change
In case of video header bidding, bidsBackHandler was getting called before bid is added to auction.
This happened because we call bidsBackHandler as soon as all bidders have called done. In case of video, ajax request is made to store the response which makes done callback to be called.

Also added missing params to prebidServer adapter.

### DIFF
--- a/modules/prebidServerBidAdapter.js
+++ b/modules/prebidServerBidAdapter.js
@@ -13,7 +13,7 @@ const getConfig = config.getConfig;
 
 const TYPE = S2S.SRC;
 let _synced = false;
-const DEFAULT_S2S_TTL = 300;
+const DEFAULT_S2S_TTL = 60;
 const DEFAULT_S2S_CURRENCY = 'USD';
 const DEFAULT_S2S_NETREVENUE = true;
 
--- a/modules/prebidServerBidAdapter.js
+++ b/modules/prebidServerBidAdapter.js
@@ -7,11 +7,15 @@ import { cookieSet } from 'src/cookie.js';
 import adaptermanager from 'src/adaptermanager';
 import { config } from 'src/config';
 import { VIDEO } from 'src/mediaTypes';
+import { isValid } from 'src/adapters/bidderFactory';
 
 const getConfig = config.getConfig;
 
 const TYPE = S2S.SRC;
 let _synced = false;
+const DEFAULT_S2S_TTL = 300;
+const DEFAULT_S2S_CURRENCY = 'USD';
+const DEFAULT_S2S_NETREVENUE = true;
 
 let _s2sConfig;
 config.setDefaults({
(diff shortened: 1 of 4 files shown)


modules/prebidServerBidAdapter.js
```diff
@@ -7,11 +7,15 @@ import { cookieSet } from 'src/cookie.js';
 import adaptermanager from 'src/adaptermanager';
 import { config } from 'src/config';
 import { VIDEO } from 'src/mediaTypes';
+import { isValid } from 'src/adapters/bidderFactory';
 
 const getConfig = config.getConfig;
 
 const TYPE = S2S.SRC;
 let _synced = false;
+const DEFAULT_S2S_TTL = 300;
```
mkendall07: let's drop to 60.

src/auction.js
```diff
@@ -148,15 +149,25 @@ export function newAuction({adUnits, adUnitCodes, callback, cbTimeout, labels})
...
       });
       request.doneCbCallCount += 1;
-      if (_bidderRequests.every((bidRequest) => bidRequest.doneCbCallCount >= 1)) {
-        // when all bidders have called done callback atleast once it means auction is complete
-        utils.logInfo(`Bids Received for Auction with id: ${_auctionId}`, _bidsReceived);
-        _auctionStatus = AUCTION_COMPLETED;
-        executeCallback(false, true);
+      // In case of mediaType video and prebidCache enabled, call bidsBackHandler after cache is stored.
+      if ((request.bids.filter(videoAdUnit).length == 0) || (request.bids.filter(videoAdUnit).length > 0 && !config.getConfig('usePrebidCache'))) {
+        bidsBackAll()
```
mkendall07: just a note this will need to change when https://github.com/prebid/Prebid.js/pull/1904 is merged (usePrebidCache is deprecated).

src/auction.js
```diff
@@ -148,15 +149,25 @@ export function newAuction({adUnits, adUnitCodes, callback, cbTimeout, labels})
...
+  /**
+   * Execute bidBackHandler if all bidders have called done.
+   */
+  function bidsBackAll() {
+    if (_bidderRequests.every((bidRequest) => bidRequest.doneCbCallCount >= 1)) {
+      // when all bidders have called done callback atleast once it means auction is complete
+      utils.logInfo(`Bids Received for Auction with id: ${_auctionId}`, _bidsReceived);
+      _auctionStatus = AUCTION_COMPLETED;
+      executeCallback(false, true);
+    }
```
mkendall07: should we emit auction end event here?

jaiminpanchal27: Not here because executeCallback also gets called in case of timeout.
So emitted auction_end just before executing `bidsBackhandler`

PR #1909, opened 2017-12-01
Remove config setting from Optimatic adapter

jaiminpanchal27:
## Type of change
- [x] Other

## Description of change.
Setting config values in adapter is not allowed. This can be done via `pbjs.setConfig` method on page.
Removed config setting from Optimatic adapter

## Other information
@optimatic58

### DIFF
--- a/modules/optimaticBidAdapter.js
+++ b/modules/optimaticBidAdapter.js
@@ -1,5 +1,4 @@
 import * as utils from 'src/utils';
-import { config } from 'src/config';
 import { registerBidder } from 'src/adapters/bidderFactory';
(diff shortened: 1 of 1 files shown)


mkendall07:
@jaiminpanchal27 
` 2:10  error  'config' is defined but never used  no-unused-vars`

optimatic58:
Does this mean we have to ask pubs to set pbjs.setConfig({ usePrebidCache: true })?  We are using vastXML in bidResponse and I understand that it will only work if the usePrebidCache is set to true.  Please let us know.  thanks!

mkendall07:
@optimatic58 
Yes that's right you need to ask pubs to do ` pbjs.setConfig({ usePrebidCache: true })` in their setup.

PR #1910, opened 2017-12-01
Change clone function to make deep copies

### LINKED ISSUE #1869: Criteo adapter native integration problem
## Type of issue
Bug

## Description
The Criteo adapter has a `nativeCallback` bidder parameter for native integrations:
http://prebid.org/dev-docs/bidders.html#criteo

This parameter is used in the adapter to request a native ad:
https://github.com/prebid/Prebid.js/blame/0d7b35d7cd84ce519aec1c15c8245616fdca86f5/modules/criteoBidAdapter.js#L52

The problem is that the JSON copy function used inside `adapterManager.js` causes the callback to be lost.
https://github.com/prebid/Prebid.js/blame/0d7b35d7cd84ce519aec1c15c8245616fdca86f5/src/adaptermanager.js#L171

### Expected results
Criteo adapter should return a bid with a native ad.

### Actual results
Criteo adapter returns a bid with a banner ad.

## Platform details
Prebid v0.33.0 + criteoBidAdapter

### LINKED ISSUE #1878: Can't specify 'renderer' function for the bid
## Type of issue
Bug

## Description
The 'renderer' property is ignored in the setup like:
  var adUnits = [{
	code: "div-gpt-ad-1438287399331-0",
	sizes: [[300, 250], [300, 600]],
	bids: [{
	  bidder: "rubicon",
	  params: {
		accountId: "4934",
		siteId: "13945",
		zoneId: "23948",
		sizes: [15]
	   },
	   renderer : {
		render : function(bid) {
			// my custom render function
		},
		url : 'http://'
	   }
        }]
  }];

## Other information
The problem is inside adaptermanager.js, lines 170-171:
  // client side adapters
  let adUnitsClientCopy = utils.cloneJson(adUnits);

The cloneJson() doesn't clone function() objects. So the 'render' property is lost on clone.

matthewlane:
## Type of change
- Bugfix

## Description of change
Changes `utils.cloneJson` to perform a deep copy. This resolves situations where functions on objects were not being copied by the utility. The utility function was also renamed to reflect the fact that it now does a full clone rather than parsing by JSON, which drops functions.

## Other information
Uses [`just-clone`](https://github.com/angus-c/just/tree/master/packages/collection-clone) library, which adds ~176 bytes to the production build

Fixes #1869 
Fixes #1878

### DIFF
--- a/package.json
+++ b/package.json
@@ -107,6 +107,7 @@
   "dependencies": {
     "babel-plugin-transform-object-assign": "^6.22.0",
     "core-js": "^2.4.1",
-    "gulp-sourcemaps": "^2.6.0"
+    "gulp-sourcemaps": "^2.6.0",
+    "just-clone": "^1.0.2"
   }
 }
--- a/src/utils.js
+++ b/src/utils.js
@@ -1,5 +1,5 @@
 import { config } from './config';
-import cloneDeep from 'just-clone';
+import clone from 'just-clone';
 var CONSTANTS = require('./constants');
 
 var _loggingChecked = false;
@@ -644,8 +644,8 @@ export function isSrcdocSupported(doc) {
     'srcdoc' in doc.defaultView.frameElement && !/firefox/i.test(navigator.userAgent);
 }
 
-export function clone(obj) {
-  return cloneDeep(obj);
+export function deepClone(obj) {
+  return clone(obj);
 }
 
 export function inIframe() {
--- a/src/utils.js
+++ b/src/utils.js
@@ -1,4 +1,5 @@
 import { config } from './config';
+import cloneDeep from 'just-clone';
 var CONSTANTS = require('./constants');
 
 var _loggingChecked = false;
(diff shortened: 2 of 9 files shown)


src/utils.js
```diff
@@ -643,8 +644,8 @@ export function isSrcdocSupported(doc) {
     'srcdoc' in doc.defaultView.frameElement && !/firefox/i.test(navigator.userAgent);
 }
 
-export function cloneJson(obj) {
-  return JSON.parse(JSON.stringify(obj));
+export function clone(obj) {
```
mkendall07: should we call this `deepClone`?

package.json
```diff
@@ -107,6 +107,7 @@
   "dependencies": {
     "babel-plugin-transform-object-assign": "^6.22.0",
     "core-js": "^2.4.1",
-    "gulp-sourcemaps": "^2.6.0"
+    "gulp-sourcemaps": "^2.6.0",
+    "just-clone": "^1.0.2"
```
mkendall07: nice find!

Issue #1855, opened 2017-11-20
33AcrossAdapter unit tests failing in 1.0


jaiminpanchal27:
## Type of issue
Bug

## Description
Unit tests are failing in `prebd-1.0` branch due to usage of `utils.getBidderRequestAllAdUnits`.
This function no longer exists in 1.0 branch.
As of now i have skipped failing tests. Please update.

## Steps to reproduce
Your adapter is merged in 1.0 branch.
Checkout https://github.com/prebid/Prebid.js/tree/prebid-1.0 and test

## Other information
@curlyblueeagle

curlyblueeagle:
@jaiminpanchal27  `utils. getBidderRequestAllAdUnits` is used in the adapter at the point `userSync` in order to obtain bid params. Is there another method to used to get the same?

jaiminpanchal27:
@curlyblueeagle `getUserSync` received `responses` as 2nd argument. Not sure if that can work for you. 
I will discuss with the team if we can also pass `bidderRequests` and let you know. 
Thanks

curlyblueeagle:
@jaiminpanchal27 Unfortunately, our server responses do not contain the bid param that I am looking for. It would be super helpful to have `bidderRequests` also passed in. Thank you.

mkendall07:
@curlyblueeagle 
you can submit a PR against the prebid-1.0 branch with that change to send through bidderRequests

curlyblueeagle:
@mkendall07 OK I'll update the adapter, but before I do that can I get some clarity on what a `bidderRequests` object looks like? I'm assuming it will contain bids for all ad units from all bidders. Am I right?

mkendall07:
@curlyblueeagle 
it should just be the request for the current adapter.

jaiminpanchal27:
Fixed in #1917

PR #1917, opened 2017-12-05
33Across Adapter: Removed the usage of utils library

curlyblueeagle:
## Type of change

- [x] Bugfix
- [ ] Feature
- [ ] New bidder adapter   
- [ ] Build related changes
- [ ] CI related changes
- [ ] Other

## Description of change

Fixed failing unit tests. Removed the usage of the utils library to get `bidderRequests` since the method has been deprecated. 

```
{
  bidder: '33across',
  params: {
    siteId: 'aRlI5W_9yr5jkxacwqm_6r',
    productId: 'infeed'
  }
}
```

aparna.hegde@33across.com

## Other information

Bugfix for https://github.com/prebid/Prebid.js/issues/1855

### DIFF
--- a/modules/33acrossBidAdapter.js
+++ b/modules/33acrossBidAdapter.js
@@ -1,4 +1,6 @@
 const { registerBidder } = require('../src/adapters/bidderFactory');
+const { userSync } = require('../src/userSync');
+const { config } = require('../src/config');
 
 const BIDDER_CODE = '33across';
 const END_POINT = 'https://ssc.33across.com/api/v1/hb';
@@ -37,9 +39,7 @@ function _createServerRequest(bidRequest) {
     }
   }
 
-  // Allowing site to be a test configuration object or just the id (former required for testing,
-  // latter when used by publishers)
-  ttxRequest.site = params.site || { id: params.siteId };
+  ttxRequest.site = { id: params.siteId };
 
   // Go ahead send the bidId in request to 33exchange so it's kept track of in the bid response and
   // therefore in ad targetting process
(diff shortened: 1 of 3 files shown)


modules/33acrossBidAdapter.js
```diff
@@ -116,12 +115,11 @@ function interpretResponse(serverResponse) {
 }
 
 // Register one sync per bid since each ad unit may potenitally be linked to a uniqe guid
-function getUserSyncs(syncOptions) {
+function getUserSyncs(syncOptions, bidderRequests) {
```
jaiminpanchal27: second argument here is `responses`. 
[This](https://github.com/prebid/Prebid.js/blob/master/src/adapters/bidderFactory.js#L342) calls adapter's getUserSync.

Since you need `bidderRequests` to do userSync. You can import `userSync` in your adapter and call `userSync.registerSync` from `interpretResponse` method. 

Here is the [hello_world](https://github.com/prebid/Prebid.js/blob/master/integrationExamples/gpt/hello_world.html) test page

curlyblueeagle: @jaiminpanchal27 Sorry, I am a confused. I was asked to introduce `bidderRequests` in `getUserSyncs` in https://github.com/prebid/Prebid.js/issues/1855. Please see @mkendall07  comments. Did I misunderstand what was being asked?

jaiminpanchal27: @curlyblueeagle Sorry for this. You did exactly what was asked. We had some misunderstanding on the final outcome of the discussion.

Can you update with my suggestion.

curlyblueeagle: Np, np. I'll update as you suggested and let you know. Thanks for the clarification

curlyblueeagle:
@jaiminpanchal27 I have implemented the fix as you suggested. Please let me know if you need anything further

modules/33acrossBidAdapter.js
```diff
@@ -1,5 +1,6 @@
 const { registerBidder } = require('../src/adapters/bidderFactory');
-const utils = require('../src/utils');
+const { userSync } = require('../src/userSync');
```
jaiminpanchal27: Do import over here.  `import { userSync } from 'src/userSync';` 
This will make sure it imports same object https://github.com/prebid/Prebid.js/blob/master/src/userSync.js#L189

modules/33acrossBidAdapter.js
```diff
@@ -81,6 +70,18 @@ function _getFormatSize(sizeArr) {
   }
 }
 
+// Register one sync per bid since each ad unit may potenitally be linked to a uniqe guid
+// Sync type will always be 'iframe' for 33Across
+function _registerUserSyncs(requestData) {
+  const ttxRequest = JSON.parse(requestData);
```
jaiminpanchal27: wrap JSON.parse in `try..catch`.

curlyblueeagle:
Done. Finger crossed

PR #1864, opened 2017-11-22
Pass admedia uid to the bidder

devmusings:
## Type of change

- [ ] Bugfix
- [x] Feature
- [ ] New bidder adapter   
- [ ] Code style update (formatting, local variables)
- [ ] Refactoring (no functional changes, no api changes)
- [ ] Build related changes
- [ ] CI related changes
- [ ] Other

## Description of change
Pass admedia's uid to the bidder.

- contact email of the adapter’s maintainer manohar@admedia.com
- [x] official adapter submission

PR #1923, opened 2017-12-05
Admedia usersync

devmusings:
## Type of change

- [ ] Bugfix
- [x] Feature
- [ ] New bidder adapter   
- [ ] Build related changes
- [ ] CI related changes
- [ ] Other

## Description of change

Added usersync to Admedia's bidder.

## Other information

Suggested updates per #1864

### DIFF
--- a/modules/admediaBidAdapter.js
+++ b/modules/admediaBidAdapter.js
@@ -1,5 +1,6 @@
 import { getBidRequest } from 'src/utils';
 import { userSync } from 'src/userSync.js';
+import { config } from 'src/config';
 
 var bidfactory = require('src/bidfactory.js');
 var bidmanager = require('src/bidmanager.js');
@@ -70,7 +71,7 @@ var AdmediaAdapter = function AdmediaAdapter() {
       adloader.loadScript(endpoint);
     }
 
-    if (userSync.iframeEnabled) {
+    if (config.getConfig('userSync').iframeEnabled) {
       userSync.registerSync('iframe', 'admedia', '//b.admedia.com/sync/iframe/');
     } else {
       userSync.registerSync('image', 'admedia', '//b.admedia.com/sync/img/?id=1');
--- a/modules/admediaBidAdapter.js
+++ b/modules/admediaBidAdapter.js
@@ -1,4 +1,6 @@
 import { getBidRequest } from 'src/utils';
+import { userSync } from 'src/userSync.js';
+
 var bidfactory = require('src/bidfactory.js');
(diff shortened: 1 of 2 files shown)


modules/admediaBidAdapter.js
```diff
@@ -67,6 +69,16 @@ var AdmediaAdapter = function AdmediaAdapter() {
 
       adloader.loadScript(endpoint);
     }
+
+    if (userSync.iframeEnabled) {
+      userSync.registerSync('iframe', 'admedia', '//b.admedia.com/sync/iframe/');
```
harpere: `userSync.iframeEnabled` does not exist.  You'd have to look at the config directly, like `config.getConfig('userSync');`

harpere:
unit tests should be added for the change.

modules/admediaBidAdapter.js
devmusings: Updated to use config.

devmusings:
Using config object and added unit tests.

harpere:
@devmusings  looks like there's a conflict in `admediaBidAdapter.js` that needs to be resolved.

test/spec/modules/admediaBidAdapter_spec.js
```diff
@@ -0,0 +1,256 @@
...
+      adapter = new Adapter();
+      adapter.callBids(request_single_adunit);
+      sandbox.stub(userSync, 'triggerUserSyncs');
+    });
+
+    afterEach(() => {
+    });
+
+    it('should call usersync after default 3000 seconds', () => {
+      window.setTimeout(function() {
```
harpere: A couple of problems with the setTimeout here.  
1.  you're not using an async callback (usually called `done`) to notify the framework the test is finished.
2. if this works, its adding a 3 second delay every time the tests are run.

I think a better approach would be to test that your `usersync.registerSync()` calls are happening as expected and let the `usersync` module tests validate that they are triggered properly.

devmusings: Agreed. Updated the tests.

devmusings:
@harpere Looks like the file modules/admediaBidAdapter.js has been removed in #1936 which is causing the conflict.

mkendall07:
@devmusings 
The adapter was removed because it doesn't comply with our new prebid 1.0 adapter requirements. You should review http://prebid.org/dev-docs/bidder-adaptor.html and submit a new new adapter that meets the requirements. I'm going to close this as it's unmergeable now.

PR #1929, opened 2017-12-06
Unit test fix in IE for adxcgAnalytics Adapter

jaiminpanchal27:
## Type of change
- [x] Bugfix

## Description of change
Fixed unit test failure in IE10 & IE11


modules/adxcgAnalyticsAdapter.js
```diff
@@ -24,7 +24,7 @@ var adxcgAnalyticsAdapter = Object.assign(adapter(
   track({eventType, args}) {
     if (typeof args !== 'undefined') {
       if (eventType === CONSTANTS.EVENTS.BID_TIMEOUT) {
-        events.bidTimeout = [...new Set(args.map(item => item.bidder))];
+        events.bidTimeout = args.map(item => item.bidder);
```
matthewlane: `Set` stores unique values, so I think this needs something like `filter(utilis.uniques)` after the `map`

PR #1939, opened 2017-12-07
s2s requests: Fixed bug for different bid_id and added start timestamp

jaiminpanchal27:
## Type of change
- [x] Bugfix

## Description of change
New `bid_id` was created for s2s requests during callBids. And start timestamp was missing.

### DIFF
--- a/src/adaptermanager.js
+++ b/src/adaptermanager.js
@@ -107,7 +107,7 @@ function transformHeightWidth(adUnit) {
   return sizesObj;
 }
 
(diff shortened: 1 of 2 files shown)


snapwich:
I don't think this will work because then if a second auction runs prebid server will get a duplicate of the adUnits from the first auction.  I took a shot at fixing it with this #1940 let me know what you think

jaiminpanchal27:
@snapwich oops.. you are right. #1940 looks good. Thanks

Issue #1941, opened 2017-12-08
Incorrect XDomainUsage.send invocation


ckbo3hrk:
## Type of issue
Bug

## Description

It seems that XDomainRequest.send is sensitive to incorrect input parameter value:

https://github.com/prebid/Prebid.js/blob/377c4c21f0e922c0549c3deaea327928bece3b12/src/ajax.js#L101 
   ... as it fails when method = 'GET'

## Steps to reproduce
1) Open IE9
2) Execute:

```
var x = new window.XDomainRequest();
var method = 'GET';
var data = undefined;
var url = 'http://cpm.metaadserving.com/rtbg?zone=30164&ad_type=rtb&v=1.0&r=%7B%22id%22%3A%223a0006d5423757%22%2C%22imp%22%3A%5B%7B%22id%22%3A%2243b4631d9943f3%22%2C%22tagid%22%3A%22container-1%22%2C%22banner%22%3A%7B%22w%22%3A300%2C%22h%22%3A250%7D%7D%5D%2C%22site%22%3A%7B%22domain%22%3A%22prebidjs.com%22%2C%22page%22%3A%22http%3A%2F%2Fprebidjs.com%2Fa.html%22%7D%2C%22at%22%3A1%2C%22device%22%3A%7B%22ip%22%3A%22caller%22%2C%22ua%22%3A%22caller%22%7D%7D&';
x.open(method, url)
x.send(method === 'POST' && data);
```

## Test page
https://jsfiddle.net/ckbo3hrk/c3ag20s3/

### Expected results
Network request to cpm.metaadserving.com/rtbg is issued

### Actual results
Exception is thrown

## Platform details
IE9 emulation in IE11, PB version 1.0.0,

PR #1942, opened 2017-12-08
XDomainRequest.send exception fix

ckbo3hrk:
## Type of change

- [x] Bugfix
- [ ] Feature
- [ ] New bidder adapter   
- [ ] Build related changes
- [ ] CI related changes
- [ ] Other

## Description of change
Unwind one-liner function call to make parameters passing logic for  XDomainRequest( or XMLHttpRequest).send more explicit as at the moment it passes **false** for GET method and it causes an exception for XDomainRequest on IE9.

## Other information
#1941

### DIFF
--- a/src/ajax.js
+++ b/src/ajax.js
@@ -98,7 +98,11 @@ export function ajaxBuilder(timeout = 3000) {
         }
         x.setRequestHeader('Content-Type', options.contentType || 'text/plain');
       }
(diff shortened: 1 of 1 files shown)


mkendall07:
looks ok to me but just a note that we are not supporting IE 9 anymore and might drop `XDomainRequest` support completely in the future.

PR #1966, opened 2017-12-15
GumGum Adapter for Prebid.js 1.0

mxcoder:
## Type of change
- [x] New bidder adapter

## Description of change
GumGum Adapter according to 1.0 rules

```
{
  bidder: 'gumgum',
  params: {
    inSlot: '9'
  }
}
```
ricardo@gumgum.com

## Other information
We already had a GumGum Adapter integrated, this is just to upgrade to 1.0
CC @bruscantini

### DIFF
--- a/modules/gumgumBidAdapter.js
+++ b/modules/gumgumBidAdapter.js
@@ -140,14 +140,13 @@ function interpretResponse (serverResponse, bidRequest) {
     cw: wrapper
   } = serverResponseBody
   let isTestUnit = (bidRequest.data && bidRequest.data.pi === 3 && bidRequest.data.si === 9)
-  let [width, height] = bidRequest.sizes[0]
+  let [width, height] = utils.parseSizesInput(bidRequest.sizes)[0].split('x')
 
   if (creativeId) {
     bidResponses.push({
       // dealId: DEAL_ID,
       // referrer: REFERER,
       ad: wrapper ? getWrapperCode(wrapper, Object.assign({}, serverResponseBody, { bidRequest })) : markup,
-      bidderCode: spec.code,
       cpm: isTestUnit ? 0.1 : cpm,
       creativeId,
       currency: 'USD',
--- a/modules/gumgumBidAdapter.js
+++ b/modules/gumgumBidAdapter.js
@@ -104,7 +104,7 @@ function buildRequests (validBidRequests) {
       data.pi = 2;
     }
     if (params.inSlot) {
-      data.si = params.inSlot;
+      data.si = parseInt(params.inSlot, 10);
       data.pi = 3;
     }
 
--- a/modules/gumgumBidAdapter.js
+++ b/modules/gumgumBidAdapter.js
@@ -139,17 +139,16 @@ function interpretResponse (serverResponse, bidRequest) {
     },
     cw: wrapper
   } = serverResponseBody
+  let isTestUnit = (bidRequest.data && bidRequest.data.pi === 3 && bidRequest.data.si === 9)
   let [width, height] = bidRequest.sizes[0]
 
-  // we have to determine what product the request was for to know which loader to use.
-  // for now use inSlotLoader
   if (creativeId) {
     bidResponses.push({
       // dealId: DEAL_ID,
       // referrer: REFERER,
       ad: wrapper ? getWrapperCode(wrapper, Object.assign({}, serverResponseBody, { bidRequest })) : markup,
       bidderCode: spec.code,
-      cpm,
+      cpm: isTestUnit ? 0.1 : cpm,
       creativeId,
       currency: 'USD',
       height,
@@ -162,21 +161,11 @@ function interpretResponse (serverResponse, bidRequest) {
   return bidResponses
 }
 
-function getUserSyncs (syncOptions) {
-  if (syncOptions.iframeEnabled) {
-    return [{
-      type: 'iframe',
-      url: 'ADAPTER_SYNC_URL'
-    }]
-  }
-}
-
 export const spec = {
   code: BIDDER_CODE,
   aliases: ALIAS_BIDDER_CODE,
   isBidRequestValid,
   buildRequests,
-  interpretResponse,
-  getUserSyncs
+  interpretResponse
 }
 registerBidder(spec)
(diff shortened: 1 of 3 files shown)


mxcoder:
Hi @matthewlane and prebid team members, just wondering if you have any info on the time window in which we could see these reviewed/merged? We know holidays are here, but just to manage expectations on our end.

Thanks

matthewlane:
Hi @mxcoder, many of us are out until the new year, and we plan to ramp back up with full PR reviews and other work in January. Thank you for your pull request and for your patience

mxcoder:
Excellent, thanks a lot @matthewlane . Have a great holiday season.

mike-chowla:
With provided test params, no ad displays.

The issue is due to the price coming back as zero.  Even though the zero bid wins the prebid auction, DFP is not selecting the ad.  If I hack the price higher, the ad renders. 

Request:
https://g2.gumgum.com/hbid/imp?si=9&pi=3&vw=1479&vh=272&sw=1920&sh=1080&pu=http%3A%2F%2Fmike.pubmatic.com%3A9999%2FintegrationExamples%2Fgpt%2Fhello_world.html%3Fpbjs_debug%3Dtrue&ce=true&dpr=1&

Response:
`{"ad":{"id":29593,"width":300,"height":250,"ipd":2000,"markup":"<div style=\"width:298px;height:248px;background:#fff;display:block;border:1px solid #000;background:#fff url(https://c.gumgum.com/images/logo/all300.png) no-repeat scroll center center\">\n<\/div>","ii":true,"du":null,"price":0,"zi":0,"impurl":"http://ca-g2.gumgum.com/ad/view/enc/hasR4Y3aDoKInNYRxCfOp-Uatqav0e1WD5G7kBz28TPH_FI449l6geYihxmDbqkBwToLOroJBD8SkTOO_3-HoDBFm10IWp16dPzbpg5B3of-PtAu3cxrjdK4-4tOgMEe51OJok4Wqa4wR9LJVmrMqtxqobqryL2-KCm4tbWGbmo8vMZCy-heOj3YHMv35pzSaq_n2CHze-VbuROJ9mnZwrb63DJ4tTxteW30RaP85-E?c=0","clsurl":"http://ca-g2.gumgum.com/ad/close?si=9&t=ggumtest&ab=29593&pv=&pu=http%3A%2F%2Fmike.pubmatic.com%3A9999%2FintegrationExamples%2Fgpt%2Fhello_world.html%3Fpbjs_debug%3Dtrue&lt=&to=&ts=1515023714357&er=0E-10"},"pag":{"t":"ggumtest","css":"#GGID ._gBadge._g.Badge {\n    z-index: 1000 !important;\n}\n\n/* DB-4066 */\nhtml { overflow-y: auto }","js":"G.infoFlag = 'japan';\n/* GG-18932 */\nG.samplerate = 100;\nG.loadObj('https://c.gumgum.com/libs/GG-18932.min.js',{type:'s',cb:false});\n/* AT-5020 */\nG.perfSampleRate = 100;\nconsole.log(\"environment\", env);"},"thms":10000,"cw":"<gumgum-ad product=\"3\" fromAS='AD_JSON'><\/gumgum-ad>"}`

mike-chowla:
User syncs are broken and causing 404s.  

The user sync url is being set to a fixed string:
url: 'ADAPTER_SYNC_URL'

Which causes the browser to load  'ADAPTER_SYNC_URL' relative to the current path which 404s

bruscantini:
Hi @mike-chowla . Thanks for the feedback. For the test params, what adjustments would you recommend? Do you think this would require non-zero price values to be returned by our ad server all the time?

mike-chowla:
The clean way to do this is to use a non-zero price coming back from the ad server.

bruscantini:
Thanks @mike-chowla . I opted to make the change in the adapter as doing so on the ad server was not as simple as I'd hoped.

mike-chowla:
Your isTestUnit check is failing because it expects the si field to be an integer but it's a string (at least when using the test parameters)

This fixes the check:
-  let isTestUnit = (bidRequest.data && bidRequest.data.pi === 3 && bidRequest.data.si === 9)
+  let isTestUnit = (bidRequest.data && bidRequest.data.pi === 3 && bidRequest.data.si === "9")

bruscantini:
Nice catch! Updated.

modules/gumgumBidAdapter.js
```diff
@@ -0,0 +1,171 @@
...
+  } = serverResponseBody
+  let isTestUnit = (bidRequest.data && bidRequest.data.pi === 3 && bidRequest.data.si === 9)
+  let [width, height] = bidRequest.sizes[0]
+
+  if (creativeId) {
+    bidResponses.push({
+      // dealId: DEAL_ID,
+      // referrer: REFERER,
+      ad: wrapper ? getWrapperCode(wrapper, Object.assign({}, serverResponseBody, { bidRequest })) : markup,
+      bidderCode: spec.code,
```
matthewlane: `bidderCode` will be set automatically by `bidderFactory` now, this line can be dropped

modules/gumgumBidAdapter.js
```diff
@@ -0,0 +1,171 @@
...
+  const {
+    ad: {
+      price: cpm,
+      id: creativeId,
+      markup
+    },
+    cw: wrapper
+  } = serverResponseBody
+  let isTestUnit = (bidRequest.data && bidRequest.data.pi === 3 && bidRequest.data.si === 9)
+  let [width, height] = bidRequest.sizes[0]
```
matthewlane: ad unit sizes may be defined as a double array (`[[300, 250]]`) or single array (`[300, 250]`). This will break if the size is defined as a single array, but `utils.parseSizesInput` can help account for either case. Something like `utils.parseSizesInput(bidRequest.sizes)[0].split('x')` if you only want the first size in the list

bruscantini:
thanks @matthewlane . Updated based on your suggestions.

matthewlane:
@mxcoder @bruscantini This is merged to master. Please also submit a PR to update your [bidder file in the docs repo](https://github.com/prebid/prebid.github.io/blob/master/dev-docs/bidders/gumgum.md), to add the line `prebid_1_0_supported : true` to the header, and make any updates to the bidder params if necessary

rmloveland:

mxcoder:
Thanks, we're working on this. We'll send another PR.

PR #1970, opened 2017-12-18
Remove duplicate request id and fix empty response from getHighesCpmBids, getAdserverTargeting

jaiminpanchal27:
## Type of change
- [x] Bugfix

## Description of change
This PR has 3 changes
1) Since 1.0 we are using `auctionId` instead of `requestId` (https://github.com/prebid/Prebid.js/blob/legacy/src/adaptermanager.js#L191)
Update bidRequests in  core fixtures and modules 

2)  `requestId` and `adId` were same in external bid object. Hence removed `requestId` only from external bid object.

3) Also found a bug, pbjs public api `getHighesCpmBids`, `getAdserverTargeting` were returning empty results. In targeting we are filtering out bids which are already set and also checking bid expiry so when `pbjs.getHighestCpmBids` is called it will filter out the bid.

### DIFF
--- a/src/utils.js
+++ b/src/utils.js
@@ -794,7 +794,7 @@ export function getBidderRequest(bidRequests, bidder, adUnitCode) {
   return find(bidRequests, request => {
     return request.bids
       .filter(bid => bid.bidder === bidder && bid.adUnitCode === adUnitCode).length > 0;
-  }) || { start: null, requestId: null };
+  }) || { start: null, auctionId: null };
 }
 
 /**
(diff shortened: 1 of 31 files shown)


src/utils.js
```diff
@@ -845,3 +845,24 @@ export function unsupportedBidderMessage(adUnit, unSupportedBidders) {
     ${plural} won't fetch demand.
   `;
 }
+
+/**
+ * Delete property from object
+ * @param {Object} object
```
matthewlane: Not sure why Travis didn't catch this but when I check out the PR locally, getting a `no-trailing-spaces` lint error on this line

jaiminpanchal27: Yep missed that. You found a bug in ESlint extension for VS Code

PR #1977, opened 2017-12-20
RockYou Adapter: Added RockYou Adapter supporting Prebid 1.0

cameronhotchkies:
## Type of change

 
- [ X ] New bidder adapter   

## Description of change

Created a new bidder adapter for RockYou.

- Test parameters for validating bids without vast player:
```
{
  bidder: 'rockyou',
  params: {
    placementId: '4322'
  }
}
```

- Test parameters for validating bids requiring vast player (this assumes that `mediaTypes: { video: {}}` is set in the adUnit):
```
{
  bidder: 'rockyou',
  params: {
    placementId: '4307'
  }
}
```

- prebid.adapter@rockyou.com
- [ X ] official adapter submission

### DIFF
--- a/modules/rockyouBidAdapter.js
+++ b/modules/rockyouBidAdapter.js
@@ -1,0 +1,357 @@
+import * as utils from 'src/utils';
+import { Renderer } from 'src/Renderer';
+import { BANNER, VIDEO } from 'src/mediaTypes';
+import { registerBidder } from 'src/adapters/bidderFactory';
+
+const BIDDER_CODE = 'rockyou';
+
+const BASE_REQUEST_PATH = 'https://tas.rockyou.net/servlet/rotator/';
+const IFRAME_SYNC_URL = 'https://prebid.tex-sync.rockyou.net/usersync2/tas';
+const VAST_PLAYER_LOCATION = 'https://rya-static.rockyou.com/rya/js/PreBidPlayer.js';
+export const ROTATION_ZONE = 'openrtbprod';
+
+let isBidRequestValid = (bid) => {
+  return !!bid.params && !!bid.params.placementId;
+};
+
+/**
+* The RockYou Ad Serving system currently only accepts one placementId
+* per Ad request. For this reason, the first placementId indicated
+* will be chosen as the predominant placementId for this request.
+*/
+let determineOptimalPlacementId = (bidRequest) => {
+  return bidRequest.params.placementId;
+}
+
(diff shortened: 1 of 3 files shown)


test/spec/modules/rockyouBidAdapter_spec.js
```diff
@@ -0,0 +1,403 @@
...
+            tally[eventName] = 1;
+          }
+        }
+      };
+
+      let callbacks = internals.playerCallbacks(renderer);
+
+      let validCallbacks = ['LOAD', 'IMPRESSION', 'COMPLETE', 'ERROR'];
+
+      for (event in validCallbacks) {
```
matthewlane: `event` should probably be preceded with `const` or `let`

cameronhotchkies:
Thank you.

Docs updated at https://github.com/prebid/prebid.github.io/pull/551/files

modules/rockyouBidAdapter.js
```diff
@@ -0,0 +1,357 @@
...
+};
+
+let extractValidSize = (bidRequest) => {
+  let width = null;
+  let height = null;
+
+  if (!utils.isEmpty(bidRequest.sizes)) {
+    // Currently only the first size is utilized
+    let firstSize = bidRequest.sizes[0];
+    width = firstSize[0];
```
mkendall07: this won't work if sizes is defined like `[300,250]`. You'll need to check for that condition.

mkendall07:
@cameronhotchkies 
one small change and this can merge.

modules/rockyouBidAdapter.js
cameronhotchkies: just to clarify, sizes can be List[List[X,Y]] or List[X,Y] ?

mkendall07: that's right. There's a helper function you can use: https://github.com/prebid/Prebid.js/blob/master/src/utils.js#L113

Sorry this is a pain. The core should handle this but it's hard to update because every adapter might be doing it differently.

cameronhotchkies: Not a problem. Is the "HxW" vs [H,W] format (as returned by the helper function) more universal?

mkendall07: I think so. But the `[H,W]` | `[[H,W]]` format is what DFP uses so that's what our external API looks like. The core should normalize this so each adapter doesn't have to.

PR #1991, opened 2017-12-21
Support multiple media formats within a single ad unit

matthewlane:
## Type of change
- Feature

## Description of change
This change supports declaring multiple media formats on a single ad unit, and allowing any bidder that supports at least one of those media formats to participate in that ad unit.

Previously, only one media format per ad type was supported, and only bidders that were compatible with that media format were eligible to participate.

Now, an ad unit can declare multiple media types, and any bidder that supports at least one of those types may take part in that auction. Any bidder that isn't compatible with the specified `mediaTypes` will be dropped from the ad unit. If `mediaTypes` is not specified on an ad unit, `banner` is the assumed format and any banner bidder is eligible for inclusion.

This PR also adds `hb_format` to standard targeting and sets its value based on the bid response object's `mediaType` property.

_Examples_:
```JavaScript
// each bidder supports at least one of the formats, so all will participate
pbjs.addAdUnits({
  code: 'adunit-code',
  sizes: [[300, 250], [300, 300]],
  mediaTypes: {
    banner: {},
    native: {},
    video: {context: 'outstream'},
  },
  bids: [

    {
      bidder: 'bannerBidder',
      params: {placementId: '481'}
    },

    {
      bidder: 'nativeBidder',
      params: {titleAsset: '516'}
    },

    {
      bidder: 'videoBidder',
      params: {vidId: '234'}
    },

  ]
});
```

```JavaScript
// only nativeBidder and videoBidder will participate
pbjs.addAdUnits({
  code: 'adunit-code',
  sizes: [[300, 250], [300, 300]],
  mediaTypes: {
    native: {},
    video: {context: 'outstream'},
  },
  bids: [

    {
      bidder: 'bannerBidder',
      params: {placementId: '481'}
    },

    {
      bidder: 'nativeBidder',
      params: {titleAsset: '516'}
    },

    {
      bidder: 'videoBidder',
      params: {vidId: '234'}
    },

  ]
});
```

### DIFF
--- a/modules/appnexusBidAdapter.js
+++ b/modules/appnexusBidAdapter.js
@@ -196,7 +196,6 @@ function newBid(serverBid, rtbBid) {
       width: rtbBid.rtb.video.player_width,
       height: rtbBid.rtb.video.player_height,
       vastUrl: rtbBid.rtb.video.asset_url,
-      descriptionUrl: rtbBid.rtb.video.asset_url,
       ttl: 3600
     });
     // This supports Outstream Video
--- a/modules/appnexusBidAdapter.js
+++ b/modules/appnexusBidAdapter.js
@@ -1,13 +1,12 @@
 import { Renderer } from 'src/Renderer';
 import * as utils from 'src/utils';
 import { registerBidder } from 'src/adapters/bidderFactory';
-import { NATIVE, VIDEO } from 'src/mediaTypes';
+import { BANNER, NATIVE, VIDEO } from 'src/mediaTypes';
 import find from 'core-js/library/fn/array/find';
 import includes from 'core-js/library/fn/array/includes';
 
 const BIDDER_CODE = 'appnexus';
 const URL = '//ib.adnxs.com/ut/v3/prebid';
-const SUPPORTED_AD_TYPES = ['banner', 'video', 'native'];
 const VIDEO_TARGETING = ['id', 'mimes', 'minduration', 'maxduration',
   'startdelay', 'skippable', 'playback_method', 'frameworks'];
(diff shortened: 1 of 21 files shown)


mkendall07:
needs a linked docs PR before merge

mkendall07:
@matthewlane 
Do we need a standard key for media type now?

matthewlane:
Added `hb_mediatype` commit, and initial docs draft here https://github.com/prebid/prebid.github.io/pull/541

modules/appnexusBidAdapter.js
```diff
@@ -31,7 +30,7 @@ const SOURCE = 'pbjs';
 export const spec = {
   code: BIDDER_CODE,
   aliases: ['appnexusAst', 'brealtime', 'pagescience', 'defymedia', 'gourmetads', 'matomy', 'featureforward', 'oftmedia', 'districtm'],
-  supportedMediaTypes: [VIDEO, NATIVE],
+  supportedMediaTypes: [BANNER, VIDEO, NATIVE],
```
jaiminpanchal27: @matthewlane If i am not wrong all other adapters supporting video or native also needs to be updated. As of now https://github.com/prebid/Prebid.js/blob/master/modules/rubiconBidAdapter.js#L76 Rubicon will not work for banner

matthewlane: Thanks, looking in to this

matthewlane: Good callout @jaiminpanchal27. I went through all adapters that have `supportedMediaTypes` and added 'banner' to those that do support it in addition to 'native' or 'video'. Some adapters only support 'native' or 'video', determined by looking at code and corresponding .md files, and in those cases `supportedMediaTypes` are left unchanged

PR #2003, opened 2017-12-27
Add eplanningBidAdapter

Nistenf:
## Type of change

- [ ] Bugfix
- [ ] Feature
- [X] New bidder adapter   
- [ ] Build related changes
- [ ] CI related changes
- [ ] Other

## Description of change

This PR adds the eplanningBidAdapter, which was removed because the old version did not comply with the new 1.0 requirements.

```
var adUnits = [{
    sizes: [[300, 250]],
    bids: [{
      bidder: 'eplanning',
      params: {
        t: 1
      }
    }]
}];
```

## Other information

The bidder parameters have not changed so the documentation doesn't need to be updated.
An important note: the adserver code needs a small update (it will be done in the next few days) before it can correctly answer in the format the adapter expects (the test adunit does work correctly, though), I'm submitting the PR now so we can be sure the adapter is correctly verified. 

Thanks!

### DIFF
--- a/modules/eplanningBidAdapter.js
+++ b/modules/eplanningBidAdapter.js
@@ -38,6 +38,7 @@ export const spec = {
         ur: utils.getTopWindowUrl() || FILE,
         r: 'pbjs',
         pbv: '$prebid.version$',
+        ncb: '1'
       };
       if (referrerUrl) {
         params.fr = referrerUrl;
--- a/modules/eplanningBidAdapter.js
+++ b/modules/eplanningBidAdapter.js
@@ -1,0 +1,155 @@
+import * as utils from 'src/utils';
+import { registerBidder } from 'src/adapters/bidderFactory';
+
+const BIDDER_CODE = 'eplanning';
+const rnd = Math.random();
+const DEFAULT_SV = 'ads.us.e-planning.net';
+const DEFAULT_ISV = 'i.e-planning.net';
+const PARAMS = ['ci', 'sv', 't'];
+const DOLLARS = 'USD';
+const NET_REVENUE = true;
+const TTL = 120;
+const NULL_SIZE = '1x1';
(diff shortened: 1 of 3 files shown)


sebaperez:
Hi @dbemiller, could this be reviewed? I know you are busy but we would like to be compatible with 1.0 asap

dbemiller:
@sebaperez I do minimal work on Prebid.js these days... my main focus is on Prebid Server.

Most everyone involved on both projects was out last week... but one of the Prebid.js maintainers will get to it whenever they have bandwidth.

mike-chowla:
The user sync generating a 404 because the url is being set to the string 'sync' which causes the browser to try load an object called 'sync' relative to the current page

```
        syncs.push({
           type: 'image',
           url: 'sync',
          });
```

mike-chowla:
The other thing I noted, the returned test ad is just text.  Most adapters test ad is an image

Nistenf:
@mike-chowla Thanks for the feedback. I just fixed the sync, and changed the test ad to an image (this needs to be deployed)

mike-chowla:
Doesn't look like the test ad image deployed but otherwise looks good

sebaperez:
@mike-chowla thank you so much for reviewing!
Can we move foward with this? The image ad for testing will be deployed in the following days. However seems that the code itself is OK. Forgive me for insisting, but we really want to be compatible with 1.0

mike-chowla:
@sebaperez, I've already enter my review as 'approved' so you are good from my end.  I understand waiting to get your 1.0 code out :-)

rmloveland:

sebaperez:
@rmloveland I just created a pr in prebid.org website repo (ref https://github.com/prebid/prebid.github.io/pull/586)

PR #2014, opened 2018-01-05
Update test stack POC

jaiminpanchal27:
## Type of change

## Description of change
This PR updates karma, sinon to latest version. 
Unit test execution is now much faster since we are creating only one webpack bundle. https://github.com/webpack-contrib/karma-webpack#alternative-usage
Thanks to @dbemiller for trying this earlier here https://github.com/prebid/Prebid.js/compare/master...single-webpack-test-bundle

Browserstack tests are passing in all browsers.

Following is done in this PR
1) Updated chrome and firefox version to run in browserstack
2) Fixed unit tests to support latest version and single bundle.
3) Fixed unit tests where unwanted http requests were fired while running tests. e.g renderer url, usersync url's 
4) Removed yarn
5) Removed IE 10 from browserstack list as sinon 4 does not support IE 10


browsers.json
```diff
@@ -198,21 +182,5 @@
     "browser": "iphone",
     "device": "iPhone 5S",
     "browser_version": null
-  },
-  "bs_ios_8": {
```
mkendall07: why are these ios dropped?

karma.conf.maker.js
```diff
@@ -35,10 +35,11 @@ function newPluginsArray(browserstack) {
     'karma-expect',
     'karma-mocha',
     'karma-requirejs',
-    'karma-sinon-ie',
```
mkendall07: 👍

karma.conf.maker.js
```diff
@@ -141,6 +140,11 @@ module.exports = function(codeCoverage, browserstack, watchMode, file) {
     autoWatch: true,
 
     reporters: ['progress'],
+    client: {
+      mocha: {
+        timeout: 0
```
mkendall07: what's this for?

mkendall07:
Needs updates to readme

jaiminpanchal27:
@mkendall07 
- Updated README. 
- Added back ios browsers which i removed by mistake.
- Also removed mocha timeout setting to 0. I added that earlier since mocha was timing out randomly. I guess some side effect of some test. 

Browserstack tests are also passing after these changes.

modules/yieldmoBidAdapter.js
```diff
@@ -40,9 +40,12 @@ export const spec = {
       w: localWindow.innerWidth,
       h: localWindow.innerHeight
     };
-    for (var request of bidRequests) {
+    // for (var request in bidRequests) {
```
snapwich: should probably just remove these comments?  definitely shouldn't change them to signaling `for...in` loops were being used as that could confuse others down the line since they behave differently than the `for...of` loops that were there originally.

src/utils.js
```diff
@@ -671,12 +672,17 @@ export function timestamp() {
...
   }
-  window.document.cookie = 'prebid.cookieTest';
-  return window.document.cookie.indexOf('prebid.cookieTest') != -1;
+}
+export function cookiesAreEnabled() {
+  if (exports.checkCookieSupport()) {
+    return true;
+  }
+  cookies.set('prebid.cookieTest');
+  return cookies.get('prebid.cookieTest') != null;
```
snapwich: i'm not sure it's worth adding a whole npm dependency just to check for cookie support.  we could use `document.cookie` directly pretty easily if we're just checking that cookies work.

test/spec/modules/improvedigitalBidAdapter_spec.js
```diff
@@ -150,6 +150,14 @@ describe('Improve Digital Adapter Tests', function () {
   });
 
   describe('interpretResponse', () => {
+    let registerSyncSpy;
+    beforeEach(() => {
+      registerSyncSpy = sinon.stub(userSync, 'registerSync');
```
snapwich: since you're stubbing, this should be called `registerSyncStub` to prevent confusion.  spies and stubs behave differently in sinon.

test/spec/modules/yieldmoBidAdapter_spec.js
```diff
@@ -137,7 +137,7 @@ describe('YieldmoAdapter', () => {
     });
   });
 
-  describe('getUserSync', () => {
+  describe.skip('getUserSync', () => {
```
snapwich: Is this intentionally skipped?

test/spec/unit/core/adapterManager_spec.js
```diff
@@ -712,9 +726,34 @@ describe('adapterManager tests', () => {
...
 
+      describe('setBidderSequence', () => {
+        let orgBidderSequence;
+        beforeEach(() => {
+          orgBidderSequence = config.getConfig('bidderSequence');
+          sinon.spy(utils, 'shuffle');
+        });
+
+        afterEach(() => {
+          config.setConfig({ bidderSequence: orgBidderSequence });
```
snapwich: I added a function to the config module for testing.  you can just call `config.resetConfig()`

test/spec/unit/pbjs_api_spec.js
```diff
@@ -812,7 +812,114 @@ describe('Unit: Prebid Module', function () {
     });
   });
 
+  describe('temp', () => {
```
snapwich: is this meant to still be here?  If so, it probably deserves a better name than 'temp'

test/spec/modules/yieldmoBidAdapter_spec.js
jaiminpanchal27: Yes it fails in iOS. I have created issue https://github.com/prebid/Prebid.js/issues/2098

modules/yieldmoBidAdapter.js
jaiminpanchal27: was playing around with both loops. Removed dead code now.

test/spec/modules/improvedigitalBidAdapter_spec.js
jaiminpanchal27: Done

test/spec/unit/core/adapterManager_spec.js
jaiminpanchal27: Done

test/spec/unit/pbjs_api_spec.js
jaiminpanchal27: Yes it was meant to be there. Updated the describe.

src/utils.js
jaiminpanchal27: I added this plugin to pass the [tests](https://github.com/prebid/Prebid.js/blob/master/test/spec/utils_spec.js#L528) created for this function. These tests are failing in all older browsers. Tried to find an alternative for current approach but there is none for older browsers.

Idea was to stub the functions and return value which will help in testing. Passing tests with the help of stubs didn't make any sense. Hence removed the tests for now.

jaiminpanchal27:
@snapwich Updated the code.

Issue #84, opened 2015-10-26
Takeovers..


cwbeck:
For premium sites we would like to road-block, whereby a response from AppNexus with priority level 10 would prevent other bids from networks like Rubicon, Criteo etc. from being considered. For this to work I believe the only way is for AppNexus to return priority within their bid response?

prebid:
Hi @cwbeck, don't think the AppNexus bidder adaptor today can run roadblocking, because it sends out individual requests for each ad unit (not a single request mode). Maybe work with your AM there to figure things out?

Heray:
@cwbeck the AppNexus adaptor will go through an update in Q1 next year to support road blocking. Closing this ticket for now as we do not have immediate action.

PR #2016, opened 2018-01-08
InSkin Bidder Adapter

jgrimes:
## Type of change

- [x] New bidder adapter   

## Description of change

Adds a new bidder adapter for InSkin Media.

```
{
  bidder: 'inskin',
  params: {
    networkId: '9874',
    siteId: '983808'
  }
}
```


mike-chowla:
I have two concerns about this adapter and whether how it operates conforms to Prebid.org rules:

1. The rendered ad is a page take over.  The adapter accomplishes this by placing a message handler at the window context and then when the ad is rendered, it posts a message to that handler so it can run at the window context.  This completely ignores the ad unit size as well as deliberately escaping the frame of the ad unit.  Even in the context of their only being one ad unit on the page, this feels problematic to me.

Here' the hello_world test page with the test ad unit: 
screen shot 2018-01-10 at 4 50 31 pm

Notice the "Prebid.js Test" and "Div-1" text on the page has been completely obscured.

2. If there are multiple ad units on the page, the InSkin ad renders those other units un-viewable which interferes with the other bidders ability to deliver ads for other ad units.  With multiple units, it seems roadblock / takeover functionality would be needed in Prebid.js so the the takeover unit needs to outbid all the slots on the page.  Issue #84 is the only reference I could find to takeover support but not it doesn't look like anything of that ilk was ever implemented.

@mkendall07 Can you comment as to whether page takeovers are permissible?

gramorris:
Apologies if my input is unwelcome but as an InSkin client I've been keeping an eye on this adapter, so...

There are already adapters in PreBid that deliver this behaviour specifically and others that escape the ad unit frame. There are also agencies delivering this behaviour through aliasing adapters such as AppNexus.

At some point, and especially within the context of Better Ads, publishers have to be responsible for the delivery of ads on their site. The concerns raised can all be resolved by the publisher and if they don't then they can be worried about the fire and fury coming their way via Chrome.

mkendall07:
thanks for pointing this out Mike. I don't think we've ever had a policy on this. It's something to think about but for now I think it's ok.

richardStrang:
Docs - https://github.com/prebid/prebid.github.io/pull/548

gramorris:
In the absence of anywhere else to put it @mkendall07....

The issue @mike-chowla raised about the size of the unit/frame is pertinent and possibly could be something that falls within the remit of Prebid to solve. The current situation as we experienced is that as Mike points out the vendor supplies the size of the creative, Prebid respects this and passes it in to DFP which then resizes the slot resulting in a large empty ad slot with an out of slot unit somewhere else.

IMO the behaviour as it is, is correct and valid. I'd rather the vendor continuing to supply the correct dimensions of the creative rather than sending through a false 0x0 or 1x1. Our current solution, knowing which vendors are sending these creatives, is to trigger a further callback on the Prebid bidWon callback and hiding the original ad slot. It's possible this could be done within Prebid if vendors also sent back an outOfSlot flag (true|false) with the bids.

The issue with the obscuring of other ads I do believe falls within the remit of the publisher. We do have in place frequency capping within DFP for higher impact creatives but I think determining that cap and also what is a higher impact creative would be difficult to implement centrally, especially with the creative changes going on around BetterAds, but also wouldn't be welcome from a publishers perspective.

benjaminclot:
Shouldn't these types of formats (skins, interstitials...), like outstream, be new ad types and use specific renderers?

PR #2026, opened 2018-01-11
Update Atomx adapter for Prebid v1.0

erikdubbelboer:
## Type of change
- [X] Update bidder adapter

## Description of change
Update existing Atomx adapter for Prebid v1.0.


rmloveland:

PR #2027, opened 2018-01-11
fix bug in UndertoneBidAdapter

omerko:
## Type of change
- [x] Bugfix
- [x] Other

## Description of change
Fix bug that caused the page domain to be sent as an array instead of a string.
Also added a page url parameter to the API request, this does not affect any user facing API or configuration


harpere:
@omerko there are lint errors that need to be fixed.

mkendall07:
@omerko any update here?

PR #2028, opened 2018-01-11
XHB Adapter update for 1.0

daniel-hoffmann:
## Type of change
- [x] Other

## Description of change
Update to XHB Adapter for 1.0


Jerome-Pedro:
Hi @mkendall07,

Please can you explain why we missed the 1.3 release and what the hold up is?

We would like to update our adapter to 1.x as we have incoming requests from our existing publishers.

Thanks, Jerome

PR #2044, opened 2018-01-18
Pass app and device fields to Prebid server

AntoineJac:
pbsrequest.go code is indeed developed to read the app fields present in the root of PBSRequest json not in the PBSRequest.ad_units child object. So far no app bundle is passed in the prebid server payload to the differents bidders...
With this fix, the device and app should be passed in the setConfig rather than AdUnits.

## Type of change

- [x] Bugfix
- [ ] Feature
- [ ] New bidder adapter   
- [ ] Build related changes
- [ ] CI related changes
- [ ] Other

## Description of change

```
{
  bidder: '<bidder name>',
  params: {
    // ...
  }
}
```

### DIFF
--- a/modules/prebidServerBidAdapter.js
+++ b/modules/prebidServerBidAdapter.js
@@ -217,7 +217,9 @@ export function PrebidServer() {
       url: utils.getTopWindowUrl(),
       prebid_version: '$prebid.version$',
       ad_units: adUnits.filter(hasSizes),
-      is_debug: isDebug
+      is_debug: isDebug,
+      device: device,
+      app: app
     };
 
     // in case config.bidders contains invalid bidders, we only process those we sent requests for.
--- a/modules/prebidServerBidAdapter.js
+++ b/modules/prebidServerBidAdapter.js
@@ -193,6 +193,8 @@ export function PrebidServer() {
   /* Prebid executes this function when the page asks to send out bid requests */
   baseAdapter.callBids = function(s2sBidRequest, bidRequests, addBidResponse, done, ajax) {
     const isDebug = !!getConfig('debug');
+    const app = !!getConfig('app');
+    const device = !!getConfig('device');
     const adUnits = utils.deepClone(s2sBidRequest.ad_units);
     adUnits.forEach(adUnit => {
       let videoMediaType = utils.deepAccess(adUnit, 'mediaTypes.video');


modules/prebidServerBidAdapter.js
```diff
@@ -193,6 +193,8 @@ export function PrebidServer() {
   /* Prebid executes this function when the page asks to send out bid requests */
   baseAdapter.callBids = function(s2sBidRequest, bidRequests, addBidResponse, done, ajax) {
     const isDebug = !!getConfig('debug');
+    const app = !!getConfig('app');
```
mkendall07: are pubs setting `app` and `device` in the config?

AntoineJac: The use case here is using Postbids in an inapp environment via DFP. Nothing is explained on prebid about passing the app and device data yet so using the config seem the most relevant and logical way of doing it. 

On the server code (go) all is already done to be able to read this fields and use in requests to bidders.

mkendall07: yeah that makes sense. If `app` and `device` are ORTB objects (and I think they are / will be soon) based on the new ORTB endpoint, then yes I think it makes sense to put these into the config as opposed to the S2S setting object. These can be used universally then vs just S2S.

mkendall07:
@AntoineJac 
I think this is good but we need a docs update to merge. I'm thinking this page should probably have some info on the structure of these items:
https://github.com/prebid/prebid.github.io/blob/master/dev-docs/get-started-with-prebid-server.md

mkendall07:
@bretg @mjacobsonny for thoughts as well.

bretg:
It's unclear where this scenario should be documented, but honestly the main getting-started page doesn't seem right. I view this PostBid-through-PrebidServerBidAdapter as a one-off that few pubs will need.

How about a FAQ entry?

**Q: How can I use Prebid Server in a mobile app post-bid scenario?**

Yes. Just schedule a creative in the ad server that:

1) Loads Prebid JS
2) Sets up the AdUnit
3) Sets the app and device with setConfig along with the s2sConfig. e.g.
```
pbjs.setConfig({
    s2sConfig: {
    ...
    },
    app: "com.test.app",
    device: "WHAT_VALUES_DOES_THIS_TAKE?"
});
```

AntoineJac:
Do you please have any ETA for this release to be deployed?

mkendall07:
@AntoineJac 
I don't think we have any issues with the code - but we need a docs PR.
@bretg 
I'm good with the FAQ. 

Separately we should have a dedicated `setConfig` API page.

bretg:
@AntoineJac and I are going to revise the interface to better reflect OpenRTB:

```
pbjs.setConfig({
    s2sConfig: {
    ...
    },
    app: {
        bundle: "com.test.app"
    },
    device: {
         ifa: "6D92078A-8246-4BA4-AE5B-76104861E7DC"
    }
});
```

Also, we think the 'app' and 'device' args for the PBS /auction endpoint are objects, not strings. Antoine is confirming, and then I'll submit a doc PR

bretg:
doc PR submitted https://github.com/prebid/prebid.github.io/pull/588

bretg:
doc PR open. back to you @mkendall07

bretg:
This PR created a bug -- Prebid Server can't handle getting app and device parameters.

matthewlane:
I just ran into this as well. I think it's because `device` and `app` are being converted to booleans (with `!!getConfig('app');`) rather than passed as objects. Dropping the boolean conversion may fix this

PR #2065, opened 2018-01-24
bidFactory: in isValid() check that bid was actually requested

### LINKED ISSUE #1984: a few missing bugfixes in the conversion from bidmanager.js -> auction.js
## Type of issue
Bug / Refactor

## Description
The move from bidmanager.js to auction.js is missing a couple changes:
220e3161203c324e0d5102d0dc5b4930a856caac
6cf71d002735c63744c0266236386b97a2f10f58
fbf0840488e76016563a1c8759fbc6c42a7d8ac8

bidderFactory.isValid is missing a change as well:
8e0ce694e58c5f55004179adedd1e6c616463ab3

With that last one, it seems less likely that a bid response will not be able to find the matching request in the new system since the bid requests are bound to the bid response function, but it still seems like good practice to verify it since a failure will break the auction.

I can readd them if you all agree they should be in there, would you all prefer separate commits or a single commit?

kitwestneat:
## Type of change
- [x] Bugfix

## Description of change
Porting 8e0ce69 to 1.0. This patch makes sure that there was a valid bid request for the received bid.

## Other information
fixes #1984

### DIFF
--- a/src/adapters/bidderFactory.js
+++ b/src/adapters/bidderFactory.js
@@ -336,13 +336,11 @@ export function newBidder(spec) {
   }
 }
 
-// check that the bid has a width and height set
-function validBidSize(adUnitCode, bid, bidRequests) {
+// check that the bid has a width and height set, or set the width and height if it can be inferred
+function validBidSize(adUnitCode, bid, adUnit) {
   if ((bid.width || bid.width === 0) && (bid.height || bid.height === 0)) {
     return true;
   }
-
-  const adUnit = getBidderRequest(bidRequests, bid.bidderCode, adUnitCode);
 
   const sizes = adUnit && adUnit.bids && adUnit.bids[0] && adUnit.bids[0].sizes;
   const parsedSizes = parseSizesInput(sizes);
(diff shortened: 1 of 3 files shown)


src/adapters/bidderFactory.js
```diff
@@ -370,13 +368,19 @@ export function isValid(adUnitCode, bid, bidRequests) {
...
+
   if (!adUnitCode) {
-    logWarn('No adUnitCode was supplied to addBidResponse.');
+    logError(errorMessage('No adUnitCode was supplied to addBidResponse.'));
     return false;
   }
 
-  if (!bid) {
-    logWarn(`Some adapter tried to add an undefined bid for ${adUnitCode}.`);
+  const req = getBidderRequest(bidRequests, bid.bidderCode, adUnitCode);
```
jaiminpanchal27: @kitwestneat We have done this check here https://github.com/prebid/Prebid.js/blob/master/src/adapters/bidderFactory.js#L293. If matching bid is not found than it won't reach here. Any specific use case you have in mind to do extra check here.

kitwestneat: Ah in the pre-1.0 version the check was in isValid, so I missed that it was moved to addBidUsingRequestMap. 

Is there a reason it was moved from isValid? It seems like the validation checks should be located together, no?

jaiminpanchal27:
@kitwestneat Any update on this ?

src/adapters/bidderFactory.js
jaiminpanchal27: I think the motivation was to return as early as possible when you find mismatch. isValid is used now to validate bid object key value pairs for different formats.

kitwestneat:
Hey @jaiminpanchal27, I haven't been able to find your response to the comment I left earlier, can you point me in the right direction? Here's my comment again in case it got lost somewhere:

> Ah in the pre-1.0 version the check was in isValid, so I missed that it was moved to addBidUsingRequestMap.
> 
> Is there a reason it was moved from isValid? It seems like the validation checks should be located together, no?

src/adapters/bidderFactory.js
kitwestneat: ah I see. What doesn't make sense to me is that from the comments in bidderFactory, all the bidderFactory code is supposed to be semi-temporary glue. But the way it is written now, addBidResponse has this implicit dependency on the validation code in bidderFactory -- they are more tightly coupled than the comments imply. It seems to me that the bid validation should be done in addBidResponse since it is ultimately responsible for maintaining the integrity of the auction. 

It looks like the other user of addBidResponse is the s2s adapters. Why do they do their own validation? Why don't they use the same validation as bidderFactory adapters? Certainly there are validation checks that must be done on both, no?

It's not really a big deal, just confusing to me. I'll close this PR since it's not going to be useful.

PR #2067, opened 2018-01-24
1298 fix fbf0840

kitwestneat:
This is a work in progress, I was hoping to get some feedback on the tests.

## Type of change
- [x] Bugfix

## Description of change
This is a port of fbf0840 to 1.0. 

The calling order for addBidToAuction and doCallbacksIfNeeded is not consistent. This patch moves most of the calls to doCallbacks into addBidToAuction to make sure it is used consistently.

There is a problem with one of the tests though. A side effect of this change is that the bid that triggers the timeout is added to the bids received list. This seems appropriate to me since if you've waited for and  received the bid, it doesn't make sense to throw it away because of timing. However, if that bid is the last bid that prebid was waiting for, the timeout event is not triggered. The test case that I removed in this patch expects that any bid received after the timeout window will trigger the timeout event. 

It seems like a fix could be to just add a second bidder so that the triggering bid is not the last bid expected. I tried modifying the test to do that, but ran into a lot of issues getting the test framework to bend to my will. I can bang on it some more, but I thought I'd see if anyone who knows the test system better could help.

As more of a general question, have you all thought about working on the test framework to make it easier to use? There is a lot of duplicated code and partially mocked objects, which makes it confusing to work with. It'd be nice if it were easier to get a standard mocked request / response, and the test could modify it as needed. 

The current situation is frustrating and there are times that I don't even submit bugfix patches because the tests are so brittle and difficult to change, like I spend 10x time on wrangling tests vs the actual bugfix. Anyway just my 2c, /rant. Would you all be open to patches to refactor the tests?

### DIFF
--- a/src/auction.js
+++ b/src/auction.js
@@ -221,7 +221,7 @@ export function newAuction({adUnits, adUnitCodes, callback, cbTimeout, labels}) 
   }
 }
 
-function doCallbacksIfNeeded(auctionInstance, bidResponse) {
+function doCallbacksIfTimedout(auctionInstance, bidResponse) {
   if (bidResponse.timeToRespond > auctionInstance.getTimeout() + config.getConfig('timeoutBuffer')) {
     auctionInstance.executeCallback(true);
   }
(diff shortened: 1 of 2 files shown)


jaiminpanchal27:
@kitwestneat Thanks for adding this back. I missed that earlier.
This looks good for me.  

Regarding tests we are working on it. With this https://github.com/prebid/Prebid.js/pull/2014 we are updating the stack and increasing the speed to run the tests on local and browserstack.
Once it gets merged next is to improve fixtures/standard mocked request/response. We have always welcomed PR. If you some idea to improve you can open issue and we will discuss.

PR #2073, opened 2018-01-26
Option to use a configurable vendor preset for s2sConfig

jsnellbaker:
## Type of change
- [x] Feature

## Description of change
I have added support for end-users to optionally use a configurable pbs vendor preset in their `setConfig({s2sConfig: {...})` page code.  By using this `defaultVendor` option, the end-user won't have to specify a majority of the data-points that are generally needed for this object.

Below is an example usage:
```
pbjs.setConfig({
    "s2sConfig": {
        accountId: 'c9d412ee-3cc6-4b66-9326-9f49d528f13e',
        bidders: ['appnexus'],
        defaultVendor: 'appnexus'
    }
});
```
Two of the required s2sConfig attributes will still be expected to be supplied by the end-user; these are `accountId` and `bidders` (as depicted above).  All of the other commonly used attributes are configured under a vendor object and are invoked by using the `defaultVendor` attribute.  

While all these vendor settings will be automatically populated based on established defaults, end-users will still have the ability to override any individual vendor settings by simply including the attribute in their page-code (same as they would today).  

An example of this would look like:
```
pbjs.setConfig({
    "s2sConfig": {
        accountId: 'c9d412ee-3cc6-4b66-9326-9f49d528f13e',
        bidders: ['appnexus'],
        defaultVendor: 'appnexus',
        timeout: 500,
        cookieSet: false
    }
});
```

Of also note - the vendor settings are configured overwrite the normally assigned system default values for the s2sConfig (assuming there is no user override).  So ultimately s2sConfig settings are established in the following hierarchy:
User overrides > vendor settings > system defaults

https://github.com/prebid/prebid.github.io/pull/579

### DIFF
--- a/modules/prebidServerBidAdapter.js
+++ b/modules/prebidServerBidAdapter.js
@@ -35,8 +35,7 @@ config.setDefaults({
 const availVendorDefaults = {
   'appnexus': {
     adapter: 'prebidServer',
-    cookieSet: true,
-    cookieSetUrl: '//acdn.adnxs.com/cookieset/cs.js',
+    cookieSet: false,
     enabled: true,
     endpoint: '//prebid.adnxs.com/pbs/v1/auction',
     syncEndpoint: '//prebid.adnxs.com/pbs/v1/cookie_sync',
@@ -360,7 +359,7 @@ export function PrebidServer() {
           });
         }
       }
-      if (result.status === 'no_cookie' && typeof _s2sConfig.cookieSetUrl === 'string') {
+      if (result.status === 'no_cookie' && _s2sConfig.cookieSet && typeof _s2sConfig.cookieSetUrl === 'string') {
         // cookie sync
         cookieSet(_s2sConfig.cookieSetUrl);
       }
--- a/modules/prebidServerBidAdapter.js
+++ b/modules/prebidServerBidAdapter.js
@@ -32,7 +32,7 @@ config.setDefaults({
 });
 
 // accountId and bidders params are not included here, should be configured by end-user
-const avail_vendor_defaults = {
+const availVendorDefaults = {
   'appnexus': {
     adapter: 'prebidServer',
     cookieSet: true,
@@ -42,7 +42,15 @@ const avail_vendor_defaults = {
     syncEndpoint: '//prebid.adnxs.com/pbs/v1/cookie_sync',
     timeout: 1000
   },
-  'rubicon': {}
+  'rubicon': {
+    adapter: 'prebidServer',
+    cookieSet: true,
+    cookieSetUrl: 'https://secure-assets.rubiconproject.com/utils/cookieset/cs.js',
+    enabled: true,
+    endpoint: 'https://prebid-server.rubiconproject.com/auction',
+    syncEndpoint: 'https://prebid-server.rubiconproject.com/cookie_sync',
+    timeout: 500
+  }
 };
 
 /**
@@ -59,13 +67,13 @@ const avail_vendor_defaults = {
  * @property {string} [cookieSetUrl] url for cookie set library, if passed then cookieSet is enabled
  */
 function setS2sConfig(options) {
-  if (options.default_vendor) {
-    let vendor = options.default_vendor
+  if (options.defaultVendor) {
+    let vendor = options.defaultVendor
 
-    if (avail_vendor_defaults.hasOwnProperty(vendor)) {
-      Object.keys(avail_vendor_defaults[vendor]).forEach(function(vendor_key) {
-        if (s2sDefaultConfig[vendor_key] === options[vendor_key]) {
-          options[vendor_key] = avail_vendor_defaults[vendor][vendor_key]
+    if (availVendorDefaults.hasOwnProperty(vendor)) {
+      Object.keys(availVendorDefaults[vendor]).forEach(function(vendorKey) {
+        if (s2sDefaultConfig[vendorKey] === options[vendorKey]) {
+          options[vendorKey] = availVendorDefaults[vendor][vendorKey]
         }
       });
     } else {
--- a/modules/prebidServerBidAdapter.js
+++ b/modules/prebidServerBidAdapter.js
@@ -70,12 +70,6 @@ function setS2sConfig(options) {
     let vendor = options.defaultVendor;
     let optionKeys = Object.keys(options);
 
-    // adding these values to default config after the fact to mimic how they're added automatically via the setConfig code in pre1api.js
-    Object.assign(s2sDefaultConfig, {
-      endpoint: 'https://prebid.adnxs.com/pbs/v1/auction',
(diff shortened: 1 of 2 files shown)


modules/prebidServerBidAdapter.js
```diff
@@ -19,15 +19,32 @@ const DEFAULT_S2S_CURRENCY = 'USD';
...
+  'appnexus': {
+    adapter: 'prebidServer',
+    cookieSet: true,
+    cookieSetUrl: '//acdn.adnxs.com/cookieset/cs.js',
+    enabled: true,
+    endpoint: '//prebid.adnxs.com/pbs/v1/auction',
+    syncEndpoint: '//prebid.adnxs.com/pbs/v1/cookie_sync',
+    timeout: 1000
+  },
+  'rubicon': {}
```
bretg: Rubicon's PBS defaults:
```
adapter: 'prebidServer',
cookieSet: true,
cookieSetUrl: 'https://secure-assets.rubiconproject.com/utils/cookieset/cs.js',
enabled: true,
endpoint: 'https://prebid-server.rubiconproject.com/auction',
syncEndpoint: 'https://prebid-server.rubiconproject.com/cookie_sync',
timeout: 500
{code}

Comment (jsnellbaker) [reply modules/prebidServerBidAdapter.js]: Hi @bretg, Thanks for submitting the information.  Just to confirm, you do want to have hard-coded secure protocol for the URLs?  For the AppNexus vendor, we were leaving them out to be dynamically populated based on the page.  Thanks.

Comment (jsnellbaker) [reply modules/prebidServerBidAdapter.js]: Hey @bretg, I've submitted the Rubicon config as is for now (as I had some other changes I wanted to submit).  If you want to have the values changed in lieu of the above (or any other reason), please feel free to let me know and I can make the updates.

**Comment by jsnellbaker** [comment]
Added docs PR (see bottom of original comment).

File: test/spec/modules/prebidServerBidAdapter_spec.js [hunk 2]
```diff
@@ -475,5 +475,62 @@ describe('S2S Adapter', () => {
       config.setConfig({ s2sConfig: options });
       sinon.assert.calledOnce(logErrorSpy);
     });
+
+    it('should log error when vendor does not exist', () => {
```
bretg: Is there a test to show there's a failure when 'defaultVendor' doesn't exist but endpoint isn't specified?  This will show there's no default-defaultVendor.

jsnellbaker: Hey @bretg - I'm not sure I follow what you mean.  Can you please clarify further when you get the chance?

mkendall07: I think he's saying to add a test case to assert an error when both endpoint & vendor default are not specified in the S2S config (or vendor default doesn't exist)

jsnellbaker: OK, thanks for that clarification.  I've added a test case in the next commit.

modules/prebidServerBidAdapter.js
```diff
@@ -19,15 +19,39 @@ const DEFAULT_S2S_CURRENCY = 'USD';
...
+    endpoint: '//prebid.adnxs.com/pbs/v1/auction',
+    syncEndpoint: '//prebid.adnxs.com/pbs/v1/cookie_sync',
+    timeout: 1000
+  },
+  'rubicon': {
+    adapter: 'prebidServer',
+    cookieSet: true,
+    cookieSetUrl: 'https://secure-assets.rubiconproject.com/utils/cookieset/cs.js',
+    enabled: true,
+    endpoint: 'https://prebid-server.rubiconproject.com/auction',
```
mkendall07: @bretg 
You want these hardcoded to HTTPS right?

modules/prebidServerBidAdapter.js
```diff
@@ -42,6 +66,21 @@ config.setDefaults({
  * @property {string} [cookieSetUrl] url for cookie set library, if passed then cookieSet is enabled
  */
 function setS2sConfig(options) {
+  if (options.defaultVendor) {
+    let vendor = options.defaultVendor
+
+    if (availVendorDefaults.hasOwnProperty(vendor)) {
+      Object.keys(availVendorDefaults[vendor]).forEach(function(vendorKey) {
+        if (s2sDefaultConfig[vendorKey] === options[vendorKey]) {
```
mkendall07: I read somewhere that you shouldn't write code as clever as you are, because debugging code is 2x more difficult than writing it. So if you write clever code, you by definition cannot debug it... :)

Ok it's probably an exaggeration but the point remains. It took me awhile to figure out why this works - it's because its compare `undefined` to `undefined` and so it's actually proceeding to copy to undefined key over. It's not necessarily wrong, it's just hard to read. At minimum it needs a comment, or ideally a nicer way to understand what's going on in code.

modules/prebidServerBidAdapter.js
bretg: We've been avoiding the protocol-relative URLs due to articles like

https://stackoverflow.com/questions/28446314/why-use-protocol-relative-urls-at-all
https://jeremywagner.me/blog/stop-using-the-protocol-relative-url/

You haven't seen issues with protocol-relative URLs?

In any case, go ahead with this for now and we'll start the conversation internally.

modules/prebidServerBidAdapter.js
jsnellbaker: Thanks for the feedback @mkendall07, I've just finished making a slight change to hopefully infer the logic better.  I also put in a comment as a backup.  Please let me know your thoughts.

modules/prebidServerBidAdapter.js
bretg: Talked to @mkendall07 -- the issue here is that non-secure demand is left on the table for HTTPS.

So, yes, let's switch the endpoints to protocol-relative. Thanks Matt. We'll take this discussion up internally as well.

modules/prebidServerBidAdapter.js
```diff
@@ -19,15 +19,39 @@ const DEFAULT_S2S_CURRENCY = 'USD';
...
+    cookieSet: false,
+    enabled: true,
+    endpoint: '//prebid.adnxs.com/pbs/v1/auction',
+    syncEndpoint: '//prebid.adnxs.com/pbs/v1/cookie_sync',
+    timeout: 1000
+  },
+  'rubicon': {
+    adapter: 'prebidServer',
+    cookieSet: true,
+    cookieSetUrl: 'https://secure-assets.rubiconproject.com/utils/cookieset/cs.js',
```
bretg: We discussed -- please remove the cookieSetUrl for Rubicon and turn cookieSet to false. Thanks. Also make "endpoint" protocol-relative. Thanks.

jsnellbaker:
All requested changes should be in place at this time.  Please take a look again and let me know.  Thanks.

PR #2092, opened 2018-02-01
Add adapter for GAMMA SSP

gammassp:
## Type of change

- [x] New bidder adapter   

## Description of change

Added HB adapter for Gamma SSP following Prebid.js 1.0

```
{
bidder: 'gamma',
    params: {
        siteId: '1465446377',
        zoneId: '1515999290'
    }
}
```
- contact email of the adapter’s maintainer: support@gammassp.com


gammassp:
Docs for Gamma bidder adapter
https://github.com/prebid/prebid.github.io/pull/583

modules/gammaBidAdapter.js
```diff
@@ -0,0 +1,85 @@
...
+function newBid(serverBid) {
+  const bid = {
+    ad: serverBid.seatbid[0].bid[0].adm,
+    cpm: serverBid.seatbid[0].bid[0].price,
+    creativeId: serverBid.seatbid[0].bid[0].adid,
+    currency: serverBid.cur,
+    dealId: serverBid.seatbid[0].bid[0].dealid,
+    width: serverBid.seatbid[0].bid[0].w,
+    height: serverBid.seatbid[0].bid[0].h,
+    mediaType: serverBid.type,
```
jsnellbaker: this property can be removed

modules/gammaBidAdapter.js
```diff
@@ -0,0 +1,85 @@
...
+  aliases: ['gamma'],
+
+  /**
+   * Determines whether or not the given bid request is valid.
+   *
+   * @param {object} bid The bid to validate.
+   * @return boolean True if this is a valid bid, and false otherwise.
+   */
+  isBidRequestValid: function(bid) {
+    return !!(bid.params.siteId || bid.params.zoneId || bid.params.gaxDomain);
```
jsnellbaker: Is the `gaxDomain` something you want to always specify within the params object?  Is this value going to change often?  Other adapters typically setup their own variable for this type of data point in the adapter file so that it's just innately available, so wanted to confirm here.

modules/gammaBidAdapter.js
gammassp: I was removed mediaType property

modules/gammaBidAdapter.js
gammassp: The gaxDomain is endpoint domain, it's required and we do not change often.

gammassp:
Hi @jsnellbaker,

I have just commit Unit test spec and updated adapter, please review again.

Thanks a lot

modules/gammaBidAdapter.js
jsnellbaker: Hi @gammassp, 

In that case - wouldn't it be more ideal/convenient to have the endpoint incorporated into your adapter's code directly rather than rely on it be passed from the `param` object?  It would be one less thing to maintain/verify within your code, as you would always know what the value would be.

gammassp: Hi @jsnellbaker,

The gaxDomain param was removed and moved into adapter, please review again.

Thanks so much

PR #2142, opened 2018-02-13
Catch criteo error

josepcasadomrf:


mkendall07:
you probably want to open this agains the `legacy` branch

PR #2150, opened 2018-02-14
Add Publisher Common ID module

pycnvr:
## Type of change

- [ ] Bugfix
- [x] Feature
- [ ] New bidder adapter   
- [ ] Build related changes
- [ ] CI related changes
- [ ] Other

## Description of change

Add a new module called pubCommonId.js.  This creates and stores a unique user id in the publisher's (first party) domain.  When enabled, the ID is appended to the bid requests for all adapters, which can pass the id for server-side user matching.

- Maintainer: pyang@conversantmedia.com

-  [doc link](https://github.com/prebid/prebid.github.io/pull/613)
## Other information
Our two teams discussed this a few weeks ago.  Here's a implementation that would seem to fit the general model by adding the common user id to the request objects.  The change to the Conversant adapter demonstrates its usage.

### DIFF
--- a/modules/pubCommonId.js
+++ b/modules/pubCommonId.js
@@ -1,0 +1,100 @@
+/**
+ * This modules adds Publisher Common ID support to prebid.js.  It's a simple numeric id
+ * stored in the page's domain.  When the module is included, an id is generated if needed,
+ * persisted as a cookie, and automatically appended to all the bidRequest as bid.crumbs.pubcid.
+ */
+import * as utils from 'src/utils'
+import { config } from 'src/config';
+
+const COOKIE_NAME = '_pubcid';
+const DEFAULT_EXPIRES = 2628000; // 5-year worth of minutes
+const PUB_COMMON = 'PublisherCommonId';
+
+var pubcidEnabled = true;
+var interval = DEFAULT_EXPIRES;
+
+export function isPubcidEnabled() { return pubcidEnabled; }
+export function getExpInterval() { return interval; }
+
+/**
+ * Decorate ad units with pubcid.  This hook function is called before the
+ * real pbjs.requestBids is invoked, and can modify its parameter.  The cookie is
+ * not updated until this function is called.
+ * @param {Object} config This is the same parameter as pbjs.requestBids, and config.adUnits will be updated.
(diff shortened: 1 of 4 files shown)


test/spec/modules/pubCommonId_spec.js
```diff
@@ -0,0 +1,188 @@
...
+        unit.bids.forEach((bid) => {
+          expect(bid).to.have.deep.property('crumbs.pubcid');
+        });
+      });
+    });
+
+    it('change expiration time', function () {
+      setConfig({expInterval: 100});
+      setCookie(COOKIE_NAME, '', -1); // erase cookie
+      expect(getExpInterval()).to.equal(100);
```
harpere: Is everything below this needed to validate that the cookie expiration interval has changed?

harpere:
needs one more review.

test/spec/modules/pubCommonId_spec.js
pycnvr: Yeah, it was just to make sure the cookie can still be retrieved after the interval change.  Not strictly needed.

modules/pubCommonId.js
```diff
@@ -0,0 +1,100 @@
...
+
+/**
+ * Configuration function
+ * @param {boolean} enable Enable or disable pubcid.  By default the module is enabled.
+ * @param {number} expInterval Expiration interval of the cookie in minutes.
+ */
+
+export function setConfig({ enable = true, expInterval = DEFAULT_EXPIRES } = {}) {
+  pubcidEnabled = enable;
+  interval = parseInt(expInterval);
```
snapwich: `parseInt` should always specify a radix to prevent weird browser quirks, base 10 here I'm guessing.

pycnvr: Nice catch.  Thanks.

pycnvr:
Hi @harpere, there is a separate PR for the doc at https://github.com/prebid/prebid.github.io/pull/613

Should I add doc within Prebid.js, too?

pycnvr:
@harpere @snapwich Thanks for doing the review.

harpere:
@pycnvr, I hadn't seen the doc.  This looks good, thanks.

PR #2157, opened 2018-02-16
[Edit BidAdapter] rxrtb adapter for Perbid.js 1.0 Update

prebidRxRTB:
## Type of change
- [x ] Bugfix

## Description of change
Fix bidfloor param check.


modules/rxrtbBidAdapter.js
```diff
@@ -100,7 +100,7 @@ function makeImp(req) {
     'banner': makeBanner(req)
   };
 
-  if (req.params.bidfloor && utils.isInteger(req.params.bidfloor)) {
+  if (req.params.bidfloor && utils.isFinite(req.params.bidfloor)) {
```
matthewlane: The `utils` module doesn't expose an `isFinite` function, but `isFinite` is available on the global object directly so if you remove `utils.` this should work

modules/rxrtbBidAdapter.js
```diff
@@ -100,7 +100,7 @@ function makeImp(req) {
     'banner': makeBanner(req)
   };
 
-  if (req.params.bidfloor && utils.isInteger(req.params.bidfloor)) {
+  if (req.params.bidfloor && Number.isFinite(req.params.bidfloor)) {
```
mkendall07: We don't support all the ES6 features (for example, IE 11 doesn't support this). I'd suggest you use `isFinite` global if you want to check against this.

PR #2161, opened 2018-02-18
Update appnexusBidAdapter so clients can distinguish between managed and RTB bids

jmfbird:
## Type of change
- [x] Other

## Description of change
Added buyerMemberId to appnexusBidAdapter bids so that clients can distinguish between managed and RTB demand

### DIFF
--- a/modules/appnexusBidAdapter.js
+++ b/modules/appnexusBidAdapter.js
@@ -191,7 +191,7 @@ function newBid(serverBid, rtbBid, bidderRequest) {
     netRevenue: true,
     ttl: 300,
     appnexus: {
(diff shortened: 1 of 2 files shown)


modules/appnexusBidAdapter.js
```diff
@@ -185,6 +185,7 @@ function newBid(serverBid, rtbBid, bidderRequest) {
   const bid = {
     requestId: serverBid.uuid,
     cpm: rtbBid.cpm,
+    buyerMemberId: rtbBid.buyer_member_id,
```
mkendall07: Let's put this into an AN specific object, ie:
```
appnexus : { 
   buyerMemberId: rtbBid.buyer_member_id
}
```

PR #2167, opened 2018-02-20
Enhance location detection within utils

rachelrj:
## Type of change
- Feature

## Description of change
Check other window objects for location if cannot find.


src/utils.js
```diff
@@ -157,17 +157,50 @@ export function parseGPTSingleSizeArray(singleSize) {
...
+    if (currentWindow.document && currentWindow.document.referrer) {
+      referrerLoc = currentWindow.document.referrer;
+    }
   }
+  while (currentWindow !== window.top);
+  return referrerLoc;
+};
 
-  return location;
+const parseFullUrl = function(locString) {
```
matthewlane: There's a `parse` function in the `src/url.js` module that parses a url and returns a similar looking object, would reusing that function in place of this one fit this use case?

src/utils.js
```diff
@@ -157,17 +157,50 @@ export function parseGPTSingleSizeArray(singleSize) {
   }
 };
 
-exports.getTopWindowLocation = function () {
-  let location;
+export function getTopWindowLocation() {
```
matthewlane: Several modules rely on this function, changing it may introduce subtle differences. Any reason to not keep the existing `getTopWindowLocation` as is, and create a new function for this case?

src/utils.js
```diff
@@ -157,17 +157,50 @@ export function parseGPTSingleSizeArray(singleSize) {
...
+  } else {
+    return window.location;
+  }
+}
+
+const getIframeParentLoc = function() {
+  let loc = window.location;
   try {
-    // force an exception in x-domain enviornments. #1509
-    window.top.location.toString();
```
matthewlane: Given the comment, not sure removing this is the correct thing to do, unless there's an equivalent fix in the new code i'm not seeing?

src/utils.js
rachelrj: Yes, though I added the `href` into the return object of that function.

src/utils.js
rachelrj: Yeah, I didn't realize that the exception wasn't thrown when trying to access window.top until I read that 1509 PR. Interesting. Thanks for pointing out. I added back in.

src/utils.js
rachelrj: The old function `getTopWindowLocation` either returns window.top.location or returns window.location.

The new `getTopWindowLocation` returns window.location, window.top.location, window.document.referrer, window.document.location.ancestorOrigins[0], or an object that recreates the values within window.location

The new version attempts to get the same info, just from additional objects.

I checked the modules, and I think this new version should fit all the current use cases.

Though, I am not opposed to creating a new function if you’re worried about it.

matthewlane: Ok, thanks for the explanation. Changes to core require a second review, assigning that now

src/utils.js
```diff
@@ -157,17 +159,47 @@ export function parseGPTSingleSizeArray(singleSize) {
...
+};
 
-  return location;
+const getNonWebKitIframeParentLoc = function() {
+  let referrerLoc = '';
+  let currentWindow;
+  do {
+    currentWindow = currentWindow ? currentWindow.parent : window;
+    if (currentWindow.document && currentWindow.document.referrer) {
+      referrerLoc = currentWindow.document.referrer;
```
mkendall07: won't this return a string when an object is expected?

rachelrj: Returns a string to `getIframeParentLoc` which then gets sent to `parse` which turns the string into the expected object.

rachelrj:
I'll work on adding some unit tests.

rachelrj:
I reorganized the additions to `getTopWindowLocation` and added unit tests.
To mock the readonly window objects in sinon, I wrapped in accessors.
I couldn't figure out a better way to mock these...

Also to use `url.parse` for `getTopWindowLocation`, `search` needs to be returned as string to mirror window.location object.

mkendall07:
There really are not any elegant solutions that I've seen to mock window and the like. The other option is to pass `window` in the module bootstrap but that's not great either.

mkendall07:
@rachelrj 
could you fix the merge conflict so I can merge? Thanks

rachelrj:
Fixed merge conflict.

PR #2202, opened 2018-02-28
Add Invibes Adapter

rcheptanariu:
## Type of change
- [x] New bidder adapter

## Description of change
Added Invibes adapter to connect to Invibes for bids

```
{
  bidder: 'invibes',
  params: {
    placementId: '12345'
  }
}
```

- contact email of the adapter’s maintainer: system_operations@invibes.com

- A link to a PR on the docs repo: https://github.com/prebid/prebid.github.io/pull/638

### DIFF
--- a/modules/invibesBidAdapter.js
+++ b/modules/invibesBidAdapter.js
@@ -259,7 +259,11 @@ function getCappedCampaignsAsString() {
   const key = 'ivvcap';
 
   let loadData = function () {
-    return JSON.parse(localStorage.getItem(key)) || {};
+    try {
+      return JSON.parse(localStorage.getItem(key)) || {};
+    } catch (e) {
+      return {};
+    }
   };
 
   let saveData = function (data) {
--- a/modules/invibesBidAdapter.js
+++ b/modules/invibesBidAdapter.js
@@ -13,7 +13,6 @@ const CONSTANTS = {
 
 export const spec = {
   code: CONSTANTS.BIDDER_CODE,
-  aliases: [CONSTANTS.BIDDER_CODE],
   /**
    * @param {object} bid
    * @return boolean
--- a/modules/invibesBidAdapter.js
+++ b/modules/invibesBidAdapter.js
@@ -134,7 +134,7 @@ function handleResponse(responseObj, bidRequests) {
     utils.logInfo('Invibes Adapter - Bid response is empty');
     return [];
   }
-  
+
   responseObj = responseObj.body || responseObj;
   responseObj = responseObj.videoAdContentResult || responseObj;
 
--- a/modules/invibesBidAdapter.js
+++ b/modules/invibesBidAdapter.js
@@ -3,10 +3,10 @@ import { registerBidder } from 'src/adapters/bidderFactory';
 import { config } from 'src/config';
 
 // use protocol relative urls for http or https
-const BID_ENDPOINT = '//localhost/KWEB.Website/Bid/VideoAdContent';
-const SYNC_ENDPOINT = 'https://www.invibes.com';
-
+const BID_ENDPOINT = '//bid.videostep.com/Bid/VideoAdContent';
+const SYNC_ENDPOINT = '//k.r66net.com/Bid/UserSync';
 const TIME_TO_LIVE = 300;
+const DEFAULT_CURRENCY = 'EUR';
 
 export const spec = {
   code: 'invibes',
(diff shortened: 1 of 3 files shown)


modules/invibesBidAdapter.js
```diff
@@ -0,0 +1,438 @@
...
+  bidRequests.forEach(function (bidRequest) {
+    bidRequest.startTime = new Date().getTime();
+    _placementIds.push(bidRequest.params.placementId);
+    _loginId = _loginId || bidRequest.params.loginId;
+    _customEndpoint = _customEndpoint || bidRequest.params.customEndpoint;
+    _bidContainerId = _bidContainerId || bidRequest.params.adContainerId || bidRequest.params.bidContainerId;
+  });
+
+  const topWin = getTopMostWindow();
+  var invibes = topWin.invibes = topWin.invibes || {};
```
jaiminpanchal27: Use let/const

modules/invibesBidAdapter.js
```diff
@@ -0,0 +1,438 @@
...
+  return `<html>
+        <head><script type='text/javascript'>inDapIF=true;</script></head>
+          <body style='margin : 0; padding: 0;'>
+          creativeHtml
+          </body>
+        </html>`
+    .replace('creativeHtml', bidModel.CreativeHtml);
+}
+
+function getCappedCampaignsAsString() {
```
jaiminpanchal27: What is the purpose of this function ?

modules/invibesBidAdapter.js
```diff
@@ -0,0 +1,438 @@
...
+
+  if (localStorage && localStorage.InvibesDEBUG) {
+    return window.console;
+  }
+
+  return { info: noop, error: noop, log: noop, warn: noop, debug: noop };
+}
+
+/// Local domain cookie management =====================
+var Uid = {
```
jaiminpanchal27: Can you share how are you using these cookie functions and what is their purpose ?

modules/invibesBidAdapter.js
```diff
@@ -0,0 +1,438 @@
...
+        cpm: ad.BidPrice,
+        width: bidModel.Width || size[0],
+        height: bidModel.Height || size[1],
+        creativeId: ad.VideoExposedId,
+        currency: bidModel.Currency || CONSTANTS.DEFAULT_CURRENCY,
+        netRevenue: true,
+        ttl: CONSTANTS.TIME_TO_LIVE,
+        ad: renderCreative(bidModel),
+
+        placementCode: bidRequest.placementCode
```
jaiminpanchal27: No need to add placementCode in bidResponse here.

modules/invibesBidAdapter.js
rcheptanariu: We have a campaign capping functionality that limits the amount of times the user should get an ad from the respective campaign, in a given time frame. An id is stored in localStorage, together with a view count and a timestamp. 
That function returns a formatted parameter that we send from the adapter to our system to know if the campaign can be displayed again at the moment, to the current user.

modules/invibesBidAdapter.js
rcheptanariu: ivNoCookie -> no cookies should be set for the user
ivbsdid -> a user id cookie. The entire 'Uid' +'initDomainId' is logic is for filtering (the user must return at least several times in multiple days. Else it is considered a temporary id and not sent with the bid request)
IvbsCampIdsLocal -> list of clicked campaign ids of the user
random test cookie -> to check for top cookie domain. We remove it afterwards

modules/invibesBidAdapter.js
```diff
@@ -0,0 +1,438 @@
...
+  SYNC_ENDPOINT: '//k.r66net.com/GetUserSync',
+  TIME_TO_LIVE: 300,
+  DEFAULT_CURRENCY: 'EUR',
+  PREBID_VERSION: 1,
+  METHOD: 'GET'
+};
+
+export const spec = {
+  code: CONSTANTS.BIDDER_CODE,
+  aliases: [CONSTANTS.BIDDER_CODE],
```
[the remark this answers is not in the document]
rcheptanariu: Sorry, I don't understand where the mistake is. If I don't export, my tests are failing because I can't import the spec. If I don't register the spec, my bidder is not called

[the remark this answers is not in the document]
jaiminpanchal27: @rcheptanariu I should have explained it in detail. 
By adding `aliases: [CONSTANTS.BIDDER_CODE]`, you are re-registering the adapter as you are using the same constant.  
So you should remove the line `aliases: [CONSTANTS.BIDDER_CODE]` or use some different alias.

jaiminpanchal27:
Also need a docs PR before merge. Please submit a PR to the [docs repo](https://github.com/prebid/prebid.github.io) to add a file for your adapter to the [bidders directory](https://github.com/prebid/prebid.github.io/tree/master/dev-docs/bidders) so your adapter's params will appear on the [bidders page](http://prebid.org/dev-docs/bidders.html). Thank you for contributing

rcheptanariu:
Docs PR at: https://github.com/prebid/prebid.github.io/pull/638
Thank you!

rcheptanariu:
Can you please confirm if the updates have been done as requested?
Thank you very much!

modules/invibesBidAdapter.js
```diff
@@ -0,0 +1,435 @@
...
+          </body>
+        </html>`
+    .replace('creativeHtml', bidModel.CreativeHtml);
+}
+
+function getCappedCampaignsAsString() {
+  const key = 'ivvcap';
+
+  let loadData = function () {
+    return JSON.parse(localStorage.getItem(key)) || {};
```
matthewlane: Any `JSON.parse` calls should be within a `try/catch` block

PR #2206, opened 2018-03-02
PBS adapter not sending app or device

mkendall07:
## Type of change

- [x] Bugfix

## Description of change
`s2sConfig.app` and `s2sConfig.device` were not getting picked up on either the legacy or new ORTB PBS endpoints. This fixes and adds tests.

### DIFF
--- a/modules/prebidServerBidAdapter.js
+++ b/modules/prebidServerBidAdapter.js
@@ -249,14 +249,17 @@ function _appendSiteAppDevice(request) {
   if (!request) return;
 
   // ORTB specifies app OR site
-  if (typeof _s2sConfig.app === 'object') {
-    request.app = _s2sConfig.app;
+  if (typeof config.getConfig('app') === 'object') {
+    request.app = config.getConfig('app');
     request.app.publisher = {id: _s2sConfig.accountId}
   } else {
-    request.site = {publisher: {id: _s2sConfig.accountId}};
+    request.site = {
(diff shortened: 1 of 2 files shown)


muncha:
Hi. I'm working on bidder for s2s and I was about to open up an issue, which I have the one line solution to locally, but it's related to this issue. The PBS "page" value isn't sending either. I'm just setting it to window.location.href on my local build and that works fine. Probably you would have noticed this while addressing "app" anyway, but just a heads up, I guess.

muncha:
While testing my s2s fork, the error I was seeing was saying it needed either "app" or "page". I just pulled this, and added the one line change to also send page, which is what I need in my test scenario. This worked fine for me :

`} else {
     request.site = {publisher: {id: _s2sConfig.accountId}, page: window.location.href};`

mkendall07:
don't merge yet - working on some updates

mkendall07:
thanks @muncha - updated to include page

mike-chowla:
Code looks good.  I'm a bit puzzled by the build failure for standard/object-curly-even-spacing since when I run lint locally on the PR, it passses

muncha:
Maybe it makes sense for the request that I made in #2243 ti include vastUrl to be included with this?

mike-chowla:
@muncha A separate PR seems to me to be the way to handle #2243 since it's not directly related to the issue this PR is fixing

muncha:
Thank you @mike-chowla

PR #2210, opened 2018-03-02
firing new adRenderFailed event when renderAd() fails

### LINKED ISSUE #2133: bidWonFailure event - Fire off event in the case that there is an error rendering from pbjs.renderAd
## Type of issue
Feature request

## Description
After the `pbjs.renderAd()` call has been made there are several checks that are made (e.g is there an `adid`). Currently these just output console statements if an error is found. It would be helpful to have an event that counts these errors/successes when looking into discrepancies between AdServer and SSP.

For example, the AdServer might count 100 impressions, however the SSP might only count 80 impressions (successful renders)

This shows a 20% discrepancy; however maybe theres an additional 15% that can be accounted for in counting these errors = 5% discrepancy. This would show that there is an integration issue instead of an SSP issue with the discrepancy

## Steps to reproduce
1) From the renderAd call, pass an invalid `adid` (or other error causing issue)
2) Note that the `bidWon` event will be fired off
3) The renderAd script will be stopped with the checks and then just show a console message. e.g: https://github.com/prebid/Prebid.js/blob/cfe1f72cdb397c6923484943235f3e8de4a9389a/src/prebid.js#L245

### Expected results
Prebid.js
```
utils.logError('Error trying to write ad. No ad for bid response id: ' + id);
msg = bid;
msg['failure'] = "Error trying to write ad. No ad for bid response";
events.emit(BID_WON_FAILURE, msg);
```
Publisher API
```
pbjs.onEvent('BidWonFailure', function(data) { notifyBidFailureInOurSystem(data); } );
```

### Actual results
Only the logError message is shown in the browser console.

vedantseta:
## Type of change

- [ ] Feature

## Description of change

This pull request resolves #2133 . If renderAd() fails for some reason, an event `adRenderFailed` is fired with data object which includes the error message and bid object if available.

### DIFF
--- a/src/prebid.js
+++ b/src/prebid.js
@@ -23,7 +23,8 @@ const { triggerUserSyncs } = userSync;
 /* private variables */
 
 const RENDERED = 'rendered';
-const { ADD_AD_UNITS, BID_WON, REQUEST_BIDS, SET_TARGETING } = CONSTANTS.EVENTS;
+const { ADD_AD_UNITS, BID_WON, REQUEST_BIDS, SET_TARGETING, AD_RENDER_FAILED } = CONSTANTS.EVENTS;
+const { PREVENT_WRITING_ON_MAIN_DOCUMENT, NO_AD, EXCEPTION, CANNOT_FIND_AD, MISSING_DOC_OR_ADID } = CONSTANTS.AD_RENDER_FAILED_REASON;
 
 var eventValidators = {
   bidWon: checkDefinedPlacement
@@ -206,6 +207,9 @@ $$PREBID_GLOBAL$$.setTargetingForAst = function() {
 $$PREBID_GLOBAL$$.renderAd = function (doc, id) {
   utils.logInfo('Invoking $$PREBID_GLOBAL$$.renderAd', arguments);
   utils.logMessage('Calling renderAd with adId :' + id);
+  const error = {
+    status: false
+  };
   if (doc && id) {
     try {
       // lookup ad by ad Id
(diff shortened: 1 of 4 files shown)


vedantseta:
added a reason key, so can GROUP BY easily.

jaiminpanchal27:
@vedantseta Thanks for PR. Please add unit tests

vedantseta:
yes, sure. will add the same in few days.

vedantseta:
@jaiminpanchal27 , can you please check

jaiminpanchal27:
@vedantseta Can you also submit a docs PR to the [docs repo](https://github.com/prebid/prebid.github.io) for this contribution. 

You need to update https://github.com/prebid/prebid.github.io/blob/master/dev-docs/publisher-api-reference.md for new event to be added to the list on this page. http://prebid.org/dev-docs/publisher-api-reference.html#module_pbjs.onEvent

vedantseta:
@jaiminpanchal27 , have updated the docs of the same

jaiminpanchal27:
@vedantseta Thanks for contributing. Since this is core change, someone from our team will do a 2nd review and merge.

src/prebid.js
```diff
@@ -229,7 +233,10 @@ $$PREBID_GLOBAL$$.renderAd = function (doc, id) {
         if (renderer && renderer.url) {
           renderer.render(bid);
         } else if ((doc === document && !utils.inIframe()) || mediaType === 'video') {
-          utils.logError(`Error trying to write ad. Ad render call ad id ${id} was prevented from writing to the main document.`);
+          error.status = true;
+          error.bid = bid;
```
mkendall07: all these repeat code can go into a function. Pass the status `string`.

vedantseta:
@mkendall07 , can you please check now?

vedantseta:
@jaiminpanchal27 @mkendall07 have refactored the code, can you please check?

src/prebid.js
```diff
@@ -196,6 +197,18 @@ $$PREBID_GLOBAL$$.setTargetingForAst = function() {
   events.emit(SET_TARGETING);
 };
 
+function emitAdRenderFail(reason, message, bid) {
+  const data = {};
+
+  data.reason = reason;
+  data.message = message;
+  if (data.bid) {
```
jsnellbaker: I wanted to check if this logic is properly configured.  It seems like this `if` condition would never be met because the `data` is initialized as empty each time the function is called, so the `data.bid` property would always be `undefined`.  Can you please take a look and clarify?

vedantseta: agreed. Changing that

PR #2213, opened 2018-03-02
GDPR consentManagement module

jsnellbaker:
**DO NOT MERGE THIS PR IMMEDIATELY.**  Please consult with @mkendall07 first.

## Type of change
- [x] Feature

## Description of change
This change includes: 
- a new module to support the fetching (via a separately implemented Consent Management Platform) and providing user consent information for the EU GDPR changes.  
- changes to the `appnexusBidAdapter`, `prebidServerBidAdapter` and `adaptermanager` classes to utilize this new module.
- small change to the `pre1api` module
- sets of new unit tests to test each of the modified areas

I have organized the overall changes for the first three items listed above into their own sections.

### GDPR consentManagement Module

This module works in the following high level steps:

1. Core prebid code reads the config information from the prebid config code (ie `pbjs.setConfig()`) and initializes the module.  This initialization sets some config variables for the module and adds a hook to the `$$PREBID_GLOBAL$$.requestBids` function in the `prebid.js` file.
2. Once the prebid code loads the `pbjs.requestBids` function, the hook is executed first (ie prior to the auction properly starting) - which starts the rest of the module.
3. The module reaches out to the CMP via its framework and attempts to fetch the user's consent information (an encoded string value).  Note - if the prebid code is nested inside an iframe, we will use an alternate approach involving `postMessage` to access the CMP framework that's outside the iframe.  This is handled automatically by the module's code.
4. Once the string is found and verified valid, it gets added to a consent object (named `gdprConsent`) which gets stored into a `gdprDataHandler` object (imported from `adaptermanager.js`).  After the value is stored, the hook ends and returns back to the original `pbjs.requestBids` function so it can run as normal.
5. This new `gdprConsent` object (sample show further below) eventually gets added to the `bidRequest` object in the `adaptermanager.js` file's `makeBidRequest` function so that it's made available to any adapter (standard or server-side) to read the consent information and pass it along to their system.  It uses the `gdprDataHandler` to read the stored value and insert it into the `bidRequest` object. 

A few additional notes to the above process:

- If the module is not included as part of the `setConfig()`, it is effectively disabled and won't create any of the additional `gdprConsent` data object in the `adUnits` or `bidRequest` auction objects.
- There is a list of supported CMPs that work with this module.  At the time of submission, only the IAB CMP is officially supported and integrated with the module.  Further documentation on this CMP can be found in a link at the bottom of this PR.  The module rejects any unknown CMPs and resumes the auction as normal.
- There is an interaction between the CMP and the module that needs to be completed in order for the module to perform its task.  There is currently a time cost attributed to this process; so we've implemented a timeout as part of the CMP piece of the module.  If the timeout is reached, the module effectively aborts what it's trying to do and exits in one of two configured ways.
- The site developer has an option (part of the setConfig() for the module) to either: cancel the auction outright if the CMP does not provide a valid consent string in the allotted time or allow the auction to proceed without the consent information (specifically the module sets an `undefined` value for the user's `consentString` attribute in the `gdprConsent` object).  This decision is controlled through the `allowAuctionWithoutConsent` option in the config; `true` means the auction will go through and `false` means the auction will be canceled entirely.
- The same exit functionality also occurs if the string returned by CMP is not valid.
- As part of the `gdprConsent` object that gets added to the `bidRequest` object, there is a second value called `consentRequired`.  This boolean value has no default value and is only populated in one of two ways.  Either the publisher would set an override value in the config part of the module, or they would leave it up to the adapters to set their appropriate value (as some of them may have the ability to derive the ideal value for the end-user).

**Samples of objects used/produced by the module**
Available options for the consentManagement module in the `setConfig()`:
```
consentManagement: {
              cmp: 'appnexus',
              consentRequired: true,
              timeout: 1000,
              allowAuctionWithoutConsent: true
            }
```

The `gdprConsent` object in the `bidRequest` object(s):
```
gdprConsent: {
        consentString: 'BOKAVy4OKAVy4ABAB8AAAAAZ+A==',
        consentRequired: true
}
```

### Adapter changes
As part of this PR, I've also implemented the changes for the `appnexusBidAdapter` and `prebidServerBidAdapter` files to read the `gdprConsent` information and pass it along to their respective systems in the proper manner.  Details on these implementations is summarized below:

**_appnexusBidAdapter_**
Per the current planned implementation approach for the /ut handler, we're storing the consent data in the `data` field of the POST request that the adapter executes as part of it's `buildRequests` function.  The consent information is stored under a top level object similar in structure to how it's presented in the original `bidRequest` object.  Below is a sample of how it looks (note the syntax difference of the object keys):
```
gdpr_consent: {
        consent_string: 'BOKAVy4OKAVy4ABAB8AAAAAZ+A==',
        consent_required: true
}
```

**_prebidServerBidAdapter_**
As per the IAB OpenRTB GDPR document (link at bottom), the consent information is stored under two different ORTB2.5 objects in its `bidRequest` object: 
- `user.ext.consent` for the `consentString`
- `regs.ext.gdpr` for the `consentReqired` boolean. 

Note the ORTB spec states to pass either a `1` or `0` to represent the `true` or `false` settings (respectively).  There is no backup value automatically set by the adapter when `consentRequired` is undefined.  This undefined value is merely carried through in the ORTB request.

The prebid GDPR module does not support the earlier sub-versions of the ORTB2 spec (which would pass the consent data in different locations in the request object).

### pre1api change
Also included as part of this PR is a small change to lower the priority of the shared requestBids hook used by the `pre1api` module.  

This change was recommended by @snapwich to allow the different modules share the same hook effectively.  My testing around the change seems to show the modules are working fine when they're running together.  This testing also included the PubCommonId module (which also shares the same hook on the `$$PREBID_GLOBAL$$.requestBids` function).

https://github.com/prebid/prebid.github.io/pull/645

## Other information
Information on AppNexus CMP framework:
http://acdn.adnxs.com/cmp/docs/#/

Information on IAB CMP framework:
https://github.com/InteractiveAdvertisingBureau/GDPR-Transparency-and-Consent-Framework

IAB OpenRTB GDPR document:
https://iabtechlab.com/wp-content/uploads/2018/02/OpenRTB_Advisory_GDPR_2018-02.pdf


modules/consentManagement.js
```diff
@@ -0,0 +1,142 @@
+import * as utils from 'src/utils';
+import { config } from 'src/config';
+import { setTimeout } from 'core-js/library/web/timers';
+
+// add new CMPs here
+const availCMPs = ['iab'];
```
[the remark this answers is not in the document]
jsnellbaker: updated references of `iab` to `appnexus`

modules/consentManagement.js
```diff
@@ -0,0 +1,142 @@
...
+
+export let userCMP;
+export let consentId = '';
+export let consentTimeout;
+export let lookUpFailureChoice;
+
+export function requestBidsHook(config, fn) {
+  let adUnits = config.adUnits || $$PREBID_GLOBAL$$.adUnits;
```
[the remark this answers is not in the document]
jsnellbaker: I removed the `$$PREBID_GLOBAL$$` fall back.  I found this type of setup from another module that was hooked on the `requestBids` function and wanted to keep consistency.

modules/consentManagement.js
```diff
@@ -0,0 +1,142 @@
...
+
+// add new CMPs here
+const availCMPs = ['iab'];
+
+export let userCMP;
+export let consentId = '';
+export let consentTimeout;
+export let lookUpFailureChoice;
+
+export function requestBidsHook(config, fn) {
```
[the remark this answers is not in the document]
jsnellbaker: Added some more descriptive information to the exported functions.

modules/consentManagement.js
```diff
@@ -0,0 +1,142 @@
...
+  let adUnits = config.adUnits || $$PREBID_GLOBAL$$.adUnits;
+  let context = this;
+  let args = arguments;
+  let nextFn = fn;
+  let cmpActive = true;
+  let _timer;
```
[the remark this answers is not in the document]
jsnellbaker: Removed the underscore for this var to make it consistent with the others.

modules/consentManagement.js
```diff
@@ -0,0 +1,142 @@
+import * as utils from 'src/utils';
+import { config } from 'src/config';
+import { setTimeout } from 'core-js/library/web/timers';
+
+// add new CMPs here
+const availCMPs = ['iab'];
+
+export let userCMP;
```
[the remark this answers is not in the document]
jsnellbaker: I'm exporting these vars so I can read them in the `consentManagement_spec.js` test file for the `setConfig` unit tests.

modules/pre1api.js
```diff
@@ -124,7 +124,7 @@ pbjs.requestBids.addHook((config, next = config) => {
   } else {
     logWarn(`${MODULE_NAME} module: concurrency has been disabled and "$$PREBID_GLOBAL$$.requestBids" call was queued`);
   }
-}, 100);
+}, 5);
```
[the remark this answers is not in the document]
jsnellbaker: The number (also relates to the other comment below) acts as a priority level for the hooked functions (when there are multiple hooked functions on the same hook).

The `pbjs.requestBids` function has 3 hooked functions from 3 different pbjs modules: pre1api, PubCommonId and consentManagement.  So lowering this priority level ensures that the `pre1api` module's hooked function will (generally) go last in the sequence if all these modules were enabled together.

Having the `pre1api` module go last (specifically after `consentManagement`) is related to the need for the `pre1api` to execute right before the `pbjs.requestBids` so it knows which auction is currently active.  The `consentManagement` module uses/waits on callbacks (to retrieve information from the CMP), and this buffering of the auction process could cause issues for the `pre1api` knowing which auction is active if there was a time gap because it ran first.  So the lower priority helps avoid this scenario.

The 5 specifically is lower than the default priority that's set for any hooked function (which is 10), but still higher than the base function `pbjs.requestBids` (which is 0).  So it should generally always the be last hook to run before the `pbjs.requestBids` would execute.

modules/consentManagement.js
```diff
@@ -0,0 +1,142 @@
...
+    } else {
+      lookUpFailureChoice = 'proceed';
+      utils.logWarn(`Invalid choice was set for consentManagement lookUpFailureResolution property. Using system default (${lookUpFailureChoice}).`);
+    }
+  } else {
+    lookUpFailureChoice = 'proceed';
+    utils.logInfo(`consentManagement config did not specify lookUpFailureResolution.  Using system default setting (${lookUpFailureChoice}).`);
+  }
+
+  $$PREBID_GLOBAL$$.requestBids.addHook(requestBidsHook, 50);
```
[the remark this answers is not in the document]
jsnellbaker: See comment above for the change in the `pre1api` file.

modules/consentManagement.js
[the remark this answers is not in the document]
jsnellbaker: *update* I need to have this value in place to actually grab the adUnits list object.  

The config attribute that's included in the `requestBids` has a place for `adUnits`, but it's dependent on it being passed specifically in the prebid config when the `requestBids` is invoked.  Currently only the `bidsBackHandler` is passed along with `requestBids` in most of our example docs, so I expect the `adUnits` is going to be `undefined` at this stage of the auction process. 

@mkendall07 Please let me know if there are any strong concerns about relying on `$$PREBID_GLOBAL$$.adUnits`.

modules/consentManagement.js
```diff
@@ -0,0 +1,142 @@
...
+  }
+
+  // ensure user requested supported cmp, otherwise abort and return to hooked function
+  if (!availCMPs.includes(userCMP)) {
+    utils.logWarn(`CMP framework ${userCMP} is not a supported framework.  Aborting consentManagement module and resuming auction.`);
+    return fn.apply(context, args);
+  }
+
+  // start of CMP specific logic
+  if (userCMP === 'iab') {
```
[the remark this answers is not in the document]
jsnellbaker: I have reorganized the functions, please take a look on the next commit and let me know what you think.

modules/consentManagement.js
```diff
@@ -0,0 +1,142 @@
...
+    userCMP = config.cmp;
+  } else {
+    userCMP = 'iab';
+    utils.logInfo(`consentManagement config did not specify cmp.  Using system default setting (${userCMP}).`);
+  }
+
+  if (typeof config.waitForConsentTimeout === 'number') {
+    consentTimeout = config.waitForConsentTimeout;
+  } else {
+    consentTimeout = 5000;
```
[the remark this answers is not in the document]
jsnellbaker: I have created default variables and utilized them in the setConfig part of the module code in place of the actual values.

jsnellbaker:
@mkendall07 I have just pushed in a commit with the changes for the various points you provided.  Please take a look again and let me know of any additional feedback.  Thanks.

modules/consentManagement.js
```diff
@@ -0,0 +1,171 @@
+/**
+ * This module adds GDPR consentManagement support to prebid.js.  It interacts with
+ * supported CMPs (Consent Management Platforms) to grab the user's consent information
+ * and make it available for any GDPR supported adapters to read/pass this information to
+ * their system.
+ */
+import * as utils from 'src/utils';
+import { config } from 'src/config';
+import { setTimeout } from 'core-js/library/web/timers';
```
snapwich: What's the purpose of this?  All the browsers we support seem to support `setTimeout` in the way you are using it w/o this polyfill.

modules/consentManagement.js
```diff
@@ -0,0 +1,171 @@
...
+export function setConfig(config) {
+  if (typeof config.cmp === 'string') {
+    userCMP = config.cmp;
+  } else {
+    userCMP = DEFAULT_CMP;
+    utils.logInfo(`consentManagement config did not specify cmp.  Using system default setting (${userCMP}).`);
+  }
+
+  if (typeof config.waitForConsentTimeout === 'number') {
+    consentTimeout = config.waitForConsentTimeout;
```
snapwich: I think this should just be called `timeout`.  It's already under the `consentManagement` property, so we know it relates to that and the way it's worded here makes the value expected ambiguous (I would guess a `bool` w/o looking at the code)

modules/consentManagement.js
```diff
@@ -0,0 +1,171 @@
...
+let cmpActive;
+let timer;
+
+// add new CMPs here, with their dedicated lookup function that passes the consentString to postLookup()
+const cmpCallMap = {
+  'appnexus': lookupAppNexusConsent
+};
+
+// AppNexus CMP lookup process
+function lookupAppNexusConsent() {
```
snapwich: So this is a public API we're making here for others that want to implement CMP management which means we have to be very exact in the way its implemented.  Right now it looks very... clunky?

I think implementing the map is fine and then looking up the CMP implementation is also fine; but the function itself should probably be a [pure function](https://en.wikipedia.org/wiki/Pure_function) for a few reasons: Firstly, if we expect this to be implemented for many different vendors then it's going to need to be thoroughly tested and the testing setup for a function that requires so much context through closure variables is very difficult.  Second, this code could possibly be moved to further modules in the future (if we don't want to include _every_ third-party CMP implementation when there are many), but the function cannot be moved to another file when it depends on so much closure context.  Third, the API is very unclear for third-party implementors; it appears to rely on various closure function references such as `postLookup` on success or `pcmTimedOut` on fail; these should probably be passed into the function as arguments and given names that represent what they are going to do, not the condition that created them e.g. `success(consentString)` or `error()` on fail (just examples, probably better names).  Also, I don't think the third-party function should be handling the timed out case..  the platform should handle that to prevent each implementation rewriting possibly buggy but expected time out handling code.

Anyways, I think a lot more thought needs to be put into how this API is exposed and implemented.

modules/consentManagement.js
```diff
@@ -0,0 +1,171 @@
...
+  } else {
+    consentTimeout = DEFAULT_CONSENT_TIMEOUT;
+    utils.logInfo(`consentManagement config did not specify waitForConsentTimeout.  Using system default setting (${consentTimeout}).`);
+  }
+
+  if (typeof config.lookUpFailureResolution === 'string') {
+    if (config.lookUpFailureResolution === 'proceed' || config.lookUpFailureResolution === 'cancel') {
+      lookupFailureChoice = config.lookUpFailureResolution;
+    } else {
+      lookupFailureChoice = DEFAULT_FAILURE_CHOICE;
```
snapwich: Is there any reason this config option shouldn't be reworked as a boolean?  It seems like right now it's just `'proceed'` or `'cancel'`.  true/false.  Is there more options I'm missing?

src/adaptermanager.js
```diff
@@ -211,6 +211,11 @@ exports.makeBidRequests = function(adUnits, auctionStart, auctionId, cbTimeout,
       bidRequests.push(bidderRequest);
     }
   });
+  if (adUnits[0].gdprConsent) {
+    bidRequests.forEach(bidRequest => {
+      bidRequest['gdprConsent'] = adUnits[0].gdprConsent;
+    });
```
snapwich: I wonder if there's a better way to do this....

modules/consentManagement.js
jsnellbaker: At the moment, there are just these two conditions.  A boolean could work; what do you think of  the name `allowAuctionWithoutConsent`?  `true` would equate to previous `proceed` and `false` to `cancel`.

modules/consentManagement.js
jsnellbaker: I removed it, it got added automatically.

modules/consentManagement.js
jsnellbaker: Good point, I changed the name to `timeout`.

jsnellbaker:
@snapwich @mkendall07 I have made additional changes in lieu of your feedback.  Please let me know your thoughts when you have the chance.

vzhukovsky:
Hi everyone.

Are there any plans to support GPDR for legacy prebid?

jsnellbaker:
@vzhukovsky Per our current plans, this GDPR module will not be supported in legacy prebid.

bretg:
@jsnellbaker - fyi snapwich is out this week.

modules/consentManagement.js
```diff
@@ -0,0 +1,164 @@
...
+    return nextFn.apply(context, args);
+  }
+
+  if (!Object.keys(cmpCallMap).includes(userCMP)) {
+    utils.logWarn(`CMP framework (${userCMP}) is not a supported framework.  Aborting consentManagement module and resuming auction.`);
+    return nextFn.apply(context, args);
+  }
+
+  // lookup times and user interaction with CMP prompts can greatly vary, so enforcing a timeout on the CMP process
+  timer = setTimeout(cmpTimedOut, consentTimeout);
```
bretg: Please skip the setTimeout if `consentTimeout` is 0. Otherwise we lose out place in the javascript stack.

jsnellbaker: @mkendall07 Talked about this change with me this morning.  I've got the changes put together and will include them in the next commit.

As a confirmation question - if the publisher were to have the setting to cancel the auction enabled, the combination of `consentTimeout=0` + the cancel setting would cause a new user's auction to be canceled immediately during that first page load where they have to fill out the consent information.  

Any subsequent page loads for that new user would go through the normal workflow and should be fine (even if the publisher had the `consentTimeout` at 0, but is that first time canceled auction an issue?  Or is it fine to leave as is?

bretg:
We'd like to not assume that `consentRequired=true` globally. The Rubicon Project exchange will be able to detect whether a given request is in GDPR scope or not based in geographic location... if `consentRequired` isn't specified, it will turn on the flag for users in the European Economic Area, leaving it off for others.

We propose:

1. The PrebidJS platform doesn't force `consentRequired` to true
1. If not specified by the publisher, bidAdapters can decide for themselves. If their system doesn't dynamically determine whether a given request needs consent, then the adapter itself can pass `consentRequired=true`
1. Publishers should always have control of course, so`setConfig()` options should allow the publisher to specify the flag for themselves
```
consentManagement: {
              cmp: 'appnexus',
              consentRequired: true,
              timeout: 10000,
              allowAuctionWithoutConsent: true
            }
```

jsnellbaker:
Thanks for the feedback, @bretg.  I'll start to make the changes around `consentRequired`.  To confirm - are there any additional logic changes that should be implemented/supported if a publisher set the `consentRequired` to `false`?  

Like if the consent information from the CMP couldn't be obtained for whatever reason and the publisher had the config to allow the Auction to continue but the Required set to `false` - should we still pass the `gdprConsent` data object in the `bidRequest` with `undefined` and `false` for the internal values?  Or should we do something different because the consent wasn't strictly required?

modules/consentManagement.js
bretg: First time cancelation is fine -- it's a pretty straightforward implication of the publisher's wishes in this scenario, IMO.

jsnellbaker:
Thanks @bretg - I'll start to implement the modified approach for the `consentRequired`.

jsnellbaker:
@mjacobsonny there is an open Docs PR: prebid/prebid.github.io#645

mkendall07:
@bretg why wouldn't the enabling of the GDRP module be enough to send `regs.ext.gdpr` ?

bretg:
What usage pattern is assumed?

1) will a pub load the CMP for every user or just EU users?
2) if the CMP is loaded for all users, can the CMP detect whether the user is in EU? i.e. are US users going to fall under GDPR rules?
3) Will the pub call PBJS setConfig(consentManagement) only for EU users or for all?

We're assuming that many pubs will not have the ability to detect EU users in their delivery stack, so will want to load CMP for everyone. If the CMP can detect EU-users, that's ideal, because then they can avoid the consent window and can hopefully activate the PBJS GDPR module only when necessary.

jsnellbaker:
I have updated the logic for the `consentRequired` field so that a default is not set automatically by the module.  

Further I have adjusted the code in the `prebidServerAdapter` so that it will populate the fields if there is data, otherwise it will use the `undefined` values.  This in turn will cause empty child objects to be created (if they're not already) under the `user.ext` and `regs.ext` fields of the ortb request.  Any pbs adapters could look for these empty fields to know if they want to determine the ideal value for the `consentRequired` field for the user or set their own default value.  

One additional note about the prebidServerAdapter code changes, I have altered the logic to check if the parent objects in the ortb request were already defined and insert the consent data accordingly.  This is to prevent the consent data from overwriting any previous data (which I noticed was the case in regards to the digiTrust functionality).

vzhukovsky:
@jsnellbaker Hi.

Don't you think that 10 seconds is too much for default timeout value?

If CMP API doesn't respond container will wait to start the auction.

jsnellbaker:
Hi @vzhukovsky, 

The wait time of 10 seconds reflects a workflow for new users who are filling out their consent information for the first time.  The module will wait for a signal that the user has submitted their information and then it will request the CMP to provide the accurate consentString value for that new-user.  We want to have the consent information available as often as possible for each auction where it's needed, which is why we temporarily hold the auction during this process.

For users who already have their consent information, the lookup process will be very quick so the delay to the auction is minimal (our earlier tests were seeing like ~50ms lookup times with CMP).  

If the value is too long for publishers, they can set their own shorter timeout period through the config options for the module.

mkendall07:
This pull request **introduces 2 alerts** when merging 78fcc64845762e7223cb07ca888f933758c90ad1 into 188f78d7d2d0112b09b5e8d1975e9d19bd65c16c - [view on lgtm.com](https://lgtm.com/projects/g/prebid/Prebid.js/rev/pr-0f755e2ae903961f342aa53da18790faa2011c51)

**new alerts:**

* 1 for Useless assignment to local variable
* 1 for Useless conditional

---

*Comment posted by [lgtm.com](https://lgtm.com)*

mkendall07:
This pull request **introduces 2 alerts** when merging c39b12d39a4596cb7f39621bdf805ce19f1d5eb6 into 188f78d7d2d0112b09b5e8d1975e9d19bd65c16c - [view on lgtm.com](https://lgtm.com/projects/g/prebid/Prebid.js/rev/pr-462341baad73a779c900dcd029214f0a92209320)

**new alerts:**

* 1 for Useless assignment to local variable
* 1 for Useless conditional

---

*Comment posted by [lgtm.com](https://lgtm.com)*

modules/consentManagement.js
```diff
@@ -0,0 +1,226 @@
...
+      command: 'getConsentData'
+    }
+  };
+
+  if (window.__cmp) {
+    window.__cmp('getConsentData', 'vendorConsents', cmpSuccess);
+  } else {
+    // prebid may be inside an iframe and CMP may exist outside, so we'll use postMessage to interact with CMP
+    // if (flag = true) {
+    window.top.postMessage(getConsentDataReq, '*');
```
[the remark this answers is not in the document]
mkendall07: @mgriego any thoughts on the above?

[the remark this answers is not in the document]
jsnellbaker: If you have any documentation on the 1.1 version of the spec and the locator iframe, can you please share the link?

GLStephen:
All, I'm trying to track down this function "lookupIabConsent" inside of the existing repositories and I see no such function here: https://github.com/InteractiveAdvertisingBureau/GDPR-Transparency-and-Consent-Framework Is this in the CMP framework yet?

jsnellbaker:
@GLStephen That's a local function of the module; it's contains the code for we interact with IAB CMP to pull the consent information.  It's not a public function of the CMP itself.

GLStephen:
@jsnellbaker Indeed I see that now. I still don't see how the connection is made to the name. Is there not a "strong" connection to the IAB CMP? The consent module is expecting the publisher to indicate they are using a framework that meets the IAB API, but there is nowhere that the IAB CMP actually identifies itself as "iab" to the consent module.

jsnellbaker:
@GLStephen In terms of the publisher knowing they're going to use the IAB API, the config for the module would have them include the following:
```
pbjs.setConfig({
consentManagement: {
   cmpApi: 'iab'
}
});
```

Inside the module, there is a call map object defined that's meant to link the values between the config (ie `'iab'` and the internal function that contains the code to call that respective CMP API (`lookupIabConsent` in this case).

In terms of the IAB CMP not branding itself better with its own code so you know it's the IAB CMP, there's not much that can be done about that.  It has it's own core object `window.__cmp` to interact with the CMP and a backup mechanism with the `postMessage` in case the prebid code is nested inside an iframe.

vzhukovsky:
@jsnellbaker 

"lookupIabConsent" is specific to CMP API spec v1.0. Right?
Do we have any plans to support 1.1?

vzhukovsky:
@jsnellbaker 

Unfortunately this spec has not published yet. I am not allowed to share it.

But AppNexus is member of the group that defined this spec. You can reach out to Andrew Sweeney or someone else who might have it.

ahubertcriteo:
Hello all,
The api method used in this PR to retrieve the consents is 'getConsentData'.
As the data retrieved is encoded, each vendor will have to decode it to check whether it has the consent to bid (either in its adapter or server side).
Wouldnt it be easier if the consent data was decoded by prebid before calling the bidder? Or if the consent data was retrieved by another api method like 'getVendorConsents'?

vzhukovsky:
Hi.

Are there still no any plans to support GDPR for Prebid legacy?

Most our users are still on 1.x.
If you decided support it we would be happy to port this module on legacy branch and contribute it back.

jsnellbaker:
In lieu of the more recent questions and the upcoming 1.1 CMP spec I wanted to provide an update on the general direction for this module.

Below is a series of changes coming but that are not due to the 1.1 CMP spec:
- There are still currently no plans to have this module ported to the `legacy` branch
- In lieu of the request from @ahubertcriteo, we will change the query to the CMP to use the `getVendorConsents` for all vendors instead just calling for the `getConsentData`.  This call will return an object with all the vendor related data (including the encoded consent string).  This object will be passed into the auction much like before and stored under `bidderRequest.gdprConsent.vendorData`.  The other two fields (`consentString` and `consentRequired`) will still be available under the `gdprConsent` object.
- Follow-up note to the above, the vendor data will very likely not be available to prebid server adapters as there is no defined place in the ortb request to store the information (as per the IAB's page on the GDPR and OpenRTB integration).  They will likely need to parse the encoded string to figure out the appropriate information.

Below is a list of changes due to the 1.1 CMP spec
- We will be renaming the `gdprConsent.consentRequired` field to `gdprConsent.gdprApplies`.  This new field is present/managed by CMP and is present on the response object when querying the CMP API.  To avoid redundancy, we're going to remove the module's `consentRequired` from the config and just rely on the CMP's `gdprApplies` value.  Adapters can still set their own back-up value in their integration code should the off-chance the `gdprApplies` isn't defined for some reason.
- In the case the prebid code is encased in an iframe, we will be using the 1.1 spec's stock iframe locator code to help interact with the CMP wherever it's located outside the iframe.
- As part of the changes to the query logic, the CMP will only respond to queries (ie fire the callback) once there is consent information available for the user.  For a new-user that's filling out their consent choices, this means the auction will wait for the choices to be made.  To support the chance for that first time user's consent choices to be used for that auction, we're bumping up the default timeout setting back to 10s (10000ms).  Publishers can set their own timeout value via the `setConfig` if they wish to use different values.  

I am planning to push these changes to the branch either today or tomorrow once I finish some additional testing.

modules/consentManagement.js
```diff
@@ -33,33 +32,63 @@ const cmpCallMap = {
...
-    __cmp: {
-      callId: 'iframe:' + (++callId),
-      command: 'getConsentData'
-    }
-  };
+  // let callId = 0;
 
   if (window.__cmp) {
-    window.__cmp('getConsentData', 'vendorConsents', cmpSuccess);
+    window.__cmp('getVendorConsents', null, cmpSuccess);
```
PubMatic-OpenWrap: if CMP js library is not yet loaded then __cmp will still be an object, not a function, good to have a isFunction check.

vzhukovsky:
@jsnellbaker Hi.
Thank you for the clarification regarding legacy. 
But "currently" still sounds promising 

Anyway, we have already done porting in our fork and will be happy to contribute back if needed.
Please just say no if it's final decision :)

modules/consentManagement.js
```diff
@@ -0,0 +1,249 @@
...
+      } catch (e) {}
+      if (f === window.top) break;
+      f = f.parent;
+    }
+
+    let cmpCallbacks = {};
+
+    /* Setup up a __cmp function to do the postMessage and stash the callback.
+      This function behaves (from the caller's perspective identicially to the in-frame __cmp call */
+    window.__cmp = function(cmd, arg, callback) {
```
cwbeck: This makes the assumption that the CMP is bound to the window in which Prebid loads. For safety, we have always loaded Prebid (and our libraries too) inside of a friendly frame, so in our case, we would need this to reference the top-most window.

```typescript
export function requestCMP(): void {
    let topWindow = WindowElement.getTopMostParentWindow();
    let createHandler = () => {
        let commandQueue: any = [];
        let cmp = function(command: any, parameter: any, callback: any){
            commandQueue.push({
                command: command,
                parameter: parameter,
                callback: callback
            });
        };
        (cmp as any).commandQueue = commandQueue;
        (cmp as any).receiveMessage = (event: any) => {
            let data = event && event.data && event.data.__cmp;
            if (data) {
                commandQueue.push({
                    callId: data.callId,
                    command: data.command,
                    parameter: data.parameter,
                    event: event
                });
            }
        };
        (cmp as any).config = {
            "globalConsentLocation": "https://acdn.adnxs.com/cmp/docs/portal.html",
            "storeConsentGlobally": true,
            "storePublisherData": false
        };
        return cmp;
    };
    (topWindow as any).__cmp = createHandler();
    topWindow.addEventListener('message', (event) => {
        (topWindow as any).__cmp.receiveMessage(event);
    });
    (topWindow as any).__cmp('showConsentTool');
    Async.loadJSIn(topWindow, "https://acdn.adnxs.com/cmp/docs/../cmp.bundle.js");
}
```

As the CMP will always be in the top-most window, can we change this?

src/utils.js
```diff
@@ -382,6 +383,10 @@ exports.isNumber = function(object) {
   return this.isA(object, t_Numb);
 };
 
+exports.isObject = function(object) {
```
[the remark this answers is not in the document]
jsnellbaker: I'll rename this function to `isPlainObject` for better clarity.

modules/consentManagement.js
jsnellbaker: Hi @cwbeck, 

The part of the code you referenced is actually part of the code where we have already determined that the CMP is not in the same window as the prebid code.  So it's not making one direction assumption as to where the CMP exists, as this can vary per publisher's implementation.

To provide some additional context on the function in question:
This function is part of the stock IAB code that's part of the 1.1 CMP spec to use the CMP's iframe locator (ie `__cmpLocator`) to detect the exact frame that the CMP is located.  

The IAB's code works basically in 3 parts:
- locates the proper window/frame that's housing the CMP code (by looking for the parent window that contains the `__cmpLocator` iframe)
- it then writes it's own local version of the `window.__cmp` API to handle the natural API call and encase it in a `postMessage` call to bypass the iframe.
- it finally adds an `addEventListener` to wait for the `message` response from the CMP to grab the consent data and do the next steps.

Please let me know if you have any further questions on this point.

mkendall07:
@jsnellbaker 
Looking great! Has 2 conflicts that need resolved.

mkendall07:
FYI this will most likely get merged by EOD.

tomeightyeight:
Just wondering when this is going to be tagged? Is this going to be rolled out as part of a larger release?

jsnellbaker:
Hi @tomeightyeight  

Now that this is merged - this is going to be included in the 1.9 release of Prebid.js, which is expected to happen later today.

jbartek25:
@vzhukovsky has there been any further discussion wrt porting the module into legacy? If not, would you share link to your port?

PR #2231, opened 2018-03-07
Update Yieldlab adapter and add official maintainer

mirkorean:
## Type of change
- [x] Bugfix
- [x] Other

## Description of change
Official update from Yieldlab. Many thanks to our friends from Platform Lunar for the work on the adapter.

Bugfixes:
- TTL was wrong. Should be 5 minutes
- Identifier used for dealId was wrong
- netRevenue should be "false"
- Video adsize should be set dynamically

Other:
- Changed params to the actual naming used by Yieldlab to avoid confusion from customers. Corresponding Pull Request to the documentation: https://github.com/prebid/prebid.github.io/pull/651
- Updated maintainer to be Yieldlab
  
- contact email of the adapter’s maintainer: solutions@yieldlab.de

### DIFF
--- a/modules/yieldlabBidAdapter.js
+++ b/modules/yieldlabBidAdapter.js
@@ -67,7 +67,7 @@ export const spec = {
           width: sizes[0],
           height: sizes[1],
           creativeId: '' + matchedBid.id,
-          dealId: matchedBid.did,
+          dealId: matchedBid.pid,
           currency: CURRENCY_CODE,
           netRevenue: true,
           ttl: BID_RESPONSE_TTL_SEC,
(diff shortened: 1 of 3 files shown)


modules/yieldlabBidAdapter.js
```diff
@@ -67,16 +67,16 @@ export const spec = {
           width: sizes[0],
           height: sizes[1],
           creativeId: '' + matchedBid.id,
-          dealId: matchedBid.did,
+          dealId: matchedBid.pid,
```
harpere: May want to add a test that verifies `pid` is being mapped to `dealId` as you expect.   Would also help verify that this was an intended change!

mirkorean: Thank you for your comment. I agree and added a commit: https://github.com/prebid/Prebid.js/pull/2231/commits/4010d702062ecaaf8320348caa383b2ab425e517

harpere: looks good, thx.  will get this merged.

PR #2239, opened 2018-03-08
Feature/travis strict lint

mkendall07:
Added a stricter lint file since it's difficult to fix all the existing lint issues at once.

This PR introduces a `lint-strict` command that will use stricter rules and fail a PR for the files modified if it doesn't pass the stricter rules. 

Open for discussion on what those "stricter" rules should look like.

### DIFF
--- a/.eslintrc.strict.js
+++ b/.eslintrc.strict.js
@@ -1,0 +1,18 @@
+module.exports = {
+  "env": {
+    "browser": true,
+    "commonjs": true
+  },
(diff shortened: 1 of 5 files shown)


mkendall07:
Previous failures where to test linting rules

mkendall07:
this will probably be annoying for submitters so I'll add a precommit hook to run the same strict-lint task for reporting on lint errors before commit.

PR #2246, opened 2018-03-08
Add billing url (burl) support

matthewlane:
## Type of change
- Feature

## Description of change
Adds billing url support to prebidServer adapter. In OpenRTB, this field is called `burl`. Prebid will call this url for S2S bid responses after and if that bid is rendered. This does not include video bid support, which would follow a different rule according to the OpenRTB spec.

### DIFF
--- a/modules/prebidServerBidAdapter.js
+++ b/modules/prebidServerBidAdapter.js
@@ -376,6 +376,8 @@ const LEGACY_PROTOCOL = {
           bidObject.currency = (bidObj.currency) ? bidObj.currency : DEFAULT_S2S_CURRENCY;
           bidObject.netRevenue = (bidObj.netRevenue) ? bidObj.netRevenue : DEFAULT_S2S_NETREVENUE;
 
+          if (result.burl) { bidObject.burl = result.burl; }
+
           bids.push({ adUnit: bidObj.code, bid: bidObject });
         });
       }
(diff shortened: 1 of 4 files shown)


bretg:
@matthewlane - where is burl documented as a feature? Not seeing it in https://github.com/prebid/prebid-server/blob/master/docs/developers/add-new-bidder.md or https://github.com/prebid/prebid-server/blob/master/docs/endpoints/openrtb2/auction.md

matthewlane:
@bretg It might not be in master yet, this is mostly according to the OpenRTB spec. When it's implemented there the docs may also be updated, ccing @dbemiller to check if this is correct

dbemiller:
That is correct, yes. The `auction.md` docs link to the OpenRTB spec, which mentions it here: https://www.iab.com/wp-content/uploads/2016/03/OpenRTB-API-Specification-Version-2-5-FINAL.pdf#page=34

It may be worth documenting inside Prebid.js if it's not already, though, because you're not as tightly coupled to OpenRTB as PBS is, and `bidObject.burl` is creating a feature which other adapters could also use if they wanted.

mkendall07:
@bretg it's an optional field for a bidder to supply but Dave's right it's part of the ORTB spec so not sure we need to call it out specifically or not in the docs.

PR #2281, opened 2018-03-19
Added new types of traffic Colossus SSP adapter 

HuddledMasses:
## Type of change

- [x] Feature
## Description of change
Added new types of traffic Colossus SSP adapter.

```
{
   bidder: 'colossusssp',
   params: {
      placement_id: 0,
      traffic: 'video'
    }
}
```
- contact email of the adapter’s maintainer: support@colossusmediallc.com


HuddledMasses:
Hi @jsnellbaker 
here is the updated Colossus docs pull request: #[680](https://github.com/prebid/prebid.github.io/pull/680)

jsnellbaker:
Hi @HuddledMasses I'm a bit confused, you are planning to support banner, video and native - right?

If so, the docs PR only included the banner support.  Shouldn't it also note the video & native support?  Also, did you have the chance to look at the recommendations for this PR?

Thanks.

HuddledMasses:
We updated pull request.
Please, review.

jsnellbaker:
Hi @HuddledMasses , thanks for submitting the update.  I was trying to test the adapter with the test params included in the PR (for the video ad unit?), but the adapter code wasn't accepting the bid response.

It seems like a banner type unit came back instead of a video type.  Further the bid was rejected because it didn't contain a `mediaType` object in the response, which is needed in the `isBidResponseValid()` check.  

Can you take a look at the test param information/bid response?  

Also, if you're going to now support additional ad unit types, I'd like to test each type (banner, video, native) to ensure they're all working successfully with your updated adapter.  If you can provide test param info for each ad type, I'd greatly appreciate it.


For reference here is a copy of the bid response I received:
```[{"width":300,"height":250,"ad":"","requestId":"28419cc0d6c5ca","cpm":5,"ttl":120,"creativeId":"123","netRevenue":true,"currency":"USD","dealId":"HASH"}]```

**Comment by HuddledMasses** [comment]
Hi @jsnellbaker
I updated the server.
Added test placements. 
Banner: {placement_id:0, traffic:"banner"}
Video: {placement_id:0, traffic:"video"}
Native: {placement_id:0, traffic:"native"}
Can you please check?

**Comment by jsnellbaker** [comment]
Hi @HuddledMasses,
Thanks for providing the additional test ad units and making the change on your server. 

I tested out each of the ad units; see details/feedback below:
- banner: LGTM
- video: LGTM
- native: needs some additional changes (details below).  

When we're receiving a native bid response, certain properties of the native bid response need to be defined in the `bid` under a `native` object.  These properties are largely the native ad's assets (like `image` (with `url`, `height`, `width`), `clickurl`, `impressionTrackers`, etc).   If these properties are not included, the bid is rejected (which is what I was seeing while testing).  

The following page has a sample markup for the `bid.native` object that needs to be populated, as well as some additional information.  
http://prebid.org/dev-docs/bidder-adaptor.html#supporting-native

Can you take a look and make the changes?  I was using the `hello_world.html` page for this testing if you want check it on your end as well.  Please let me know if you have any questions.  We should be good once this last piece is in place and working.

**Comment by HuddledMasses** [comment]
Hi @jsnellbaker 
We changed native response validation.
Also changed the formation of the response from the server.
The documentation indicates the ability to set different types of traffic. 
https://github.com/prebid/prebid.github.io/blob/master/dev-docs/bidders/colossusssp.md

**Comment by HuddledMasses** [comment]
Hi @jsnellbaker 
We have made changes in the native.
Also added the transfer of data by size to the server.

**Comment by jsnellbaker** [comment]
Hi @HuddledMasses, Thanks again for making the changes.  

I tested the native ad unit and found it was structured properly and passing the internal checks.  However, ad unit wasn't rendering on the test page.  While looking over the browser console, I noticed the `pbjs.renderAd` function was being executed after we got the bid response.  

This is normally used to help render banner and video ads and is invoked from the prebid creative template.  The native creative template is different (as it has to handle different creative assets).  

The following page gives an outline of what the native creative code looks like: http://prebid.org/adops/setting-up-prebid-native-in-dfp.html  

Can you take a look over this page and try to make the needed changes?  I believe we should be good to merge after this is addressed.  Thanks!

**Comment by HuddledMasses** [comment]
@jsnellbaker 
I don't understand. 
Where do I need to make changes?
The adapter doesn't affect the operation or sequence of the function call `pbjs.renderAd`.

**Comment by jsnellbaker** [comment]
@HuddledMasses  There should be a creative in your ad server that would return a 'prebid' creative.  Normally this creative is a set of JavaScript that makes the call the `pbjs.renderAd` to render a video or banner creative that's retruned from the auction.  In this case for the native, the creative should be setup with a different creative template in the format that was referenced in the page from before (ie plain HTML).

Maybe one of the traffickers or adops people on your end may be able to help point you to the area where the changes need to be made?

If it will help, below is a link to a more parent-level page on the different ways to setup prebid creatives:
http://prebid.org/adops/docs-by-ad-server.html

**Comment by HuddledMasses** [comment]
Team, can you please let us know why Colossus adapter has not been listed here? 

http://prebid.org/dev-docs/bidders.html#bidders-with-video-and-native-demand

**Comment by jsnellbaker** [comment]
Hi @HuddledMasses 

To have the bidder appear on that table, you'd need to submit a PR to the Prebid Docs repo (located [here](https://github.com/prebid/prebid.github.io)) to update this file:
https://github.com/prebid/prebid.github.io/blob/master/dev-docs/bidders/colossusssp.md

Just need to add the following variables to the top section (where the other variables are located):
```
media_types: banner, video, native
```

HuddledMasses:
PR has been updated: https://github.com/prebid/prebid.github.io/pull/1484 

Please take a look. Thanks!

PR #2283, opened 2018-03-19
Widespace adapter

mizmaar3:
## Type of change

- [x] New bidder adapter   

## Description of change
Widespace bidder adapter for prebid 1.x.x

```
 /** WS dev server url **/
 var wisp = wisp || {};
 wisp.ENGINE_URL = "https://nova-dev-engine.widespace.com/map/engine/dynadreq";

/** bid params **/
{
  bidder: 'widespace',
  params: {
   sid: '7b6589bf-95c8-4656-90b9-af9737bb9ad3',
   currency: 'EUR'
  }
}

/** Please append a hash parameter `#WS_DEBUG_FORCEADID=47696` to the url to always get a test ad  **/ 
```

-- muhammad.usman@widespace.com 

-- https://github.com/prebid/prebid.github.io/pull/674


mkendall07:
This pull request **introduces 1 alert** when merging c246d9dfe91bfa8badd57fbcacec484c9236ef1e into 1c862a8299928fc6c8bfd5aeb5be060f385d202b - [view on lgtm.com](https://lgtm.com/projects/g/prebid/Prebid.js/rev/pr-daec5bada5d3cd5c84a139e87432bd9b085eae2b)

**new alerts:**

---

*Comment posted by [lgtm.com](https://lgtm.com)*

mike-chowla:
Use of $$PREBID_GLOBAL$$ is not permitted in 1.x adapters:
const PBJS = window['$$PREBID_GLOBAL$$'];

Additionally, it appears that you are storing winning bids of other partners and then sending off to your backend.  Is that correct?

mizmaar3:
yes it is correct

mizmaar3:
And if you wonder why we are storing/sending winning bid info to backend when we lose! is because it would correspond to the "Loss Notice URL" we're getting from publishers over OpenRTB.

Please let me know what is alternative to `$$PREBID_GLOBAL$$` if any

mike-chowla:
I checked with the head of the Prebid.js PMC and sending back winning bids is not permitted.

In general, you can not depend on receiving an event that signals a loss as if you win the Prebid auction but ad server selects a non-prebid line item,  prebid will not get called and thus the only indication of a loss is your win pixel inside the creative not firing.  With the current adapter code, you'd also not get notified when there is no subsequent bid request on the page. 

My understanding is the current adapter interface doesn't allow for sending loss notifications back.  If you think there should be a method to do this, I'd recommend opening a new issue on the project.

mizmaar3:
Ok I removed event listener and storing/sending winning bid data now

mike-chowla:
Test coverage is below 80%

screen shot 2018-03-22 at 5 50 52 pm

mizmaar3:
How can I check coverage ?

mizmaar3:
Its increased now

mizmaar3:
https://github.com/prebid/prebid.github.io/pull/674

modules/widespaceBidAdapter.js
```diff
@@ -0,0 +1,239 @@
...
+    }
+    return false;
+  },
+
+  buildRequests: function(validBidRequests) {
+    let serverRequests = [];
+    const REQUEST_SERVER_URL = getEngineUrl();
+    const DEMO_DATA_PARAMS = ['gender', 'country', 'region', 'postal', 'city', 'yob'];
+    const PERF_DATA = getData(LS_KEYS.PERF_DATA).map(perf_data => JSON.parse(perf_data));
+    const CUST_DATA = getData(LS_KEYS.CUST_DATA, false)[0];
```
mike-chowla: I don't see a storeData call for CUST_DATA so wouldn't it always be empty

modules/widespaceBidAdapter.js
```diff
@@ -0,0 +1,239 @@
...
+
+function storeData(data, name, stringify = true) {
+  const value = stringify ? JSON.stringify(data) : data;
+  if (LOCAL_STORAGE_AVAILABLE) {
+    localStorage.setItem(name, value);
+    return true;
+  } else if (COOKIE_ENABLED) {
+    const theDate = new Date();
+    const expDate = new Date(theDate.setMonth(theDate.getMonth() + 12)).toGMTString();
+    window.document.cookie = `${name}=${value};path=/;expires=${expDate}`;
```
mike-chowla: This creates a cookie in the publisher's cookie space which seems like something the publisher should at least informed is happening.  I don't see any mention in the docs that adapter creates cookies.  However, I see other adapters doing this too and do not know of specific prohibition against it.

modules/widespaceBidAdapter.js
mizmaar3: This will be set by our server and I will just pass it on in next request when available

modules/widespaceBidAdapter.js
mizmaar3: Yes I havnt found any adapter doc mentioning it, also cookie is just fallback and we have seen that 99.9% chances are for localStorage to be used.

modules/widespaceBidAdapter.js
mike-chowla: I couldn't see how it ever gets set to local storage.  Am I just missing where that happens?

mizmaar3: No, you are not missing anything because `CUST_DATA` is not `set` by adapter but rather by  adCreative code. This allows us get some sort of feedback when our won bid creative is rendered. Example: - If we detected some problem when trying to render ad i.e. unable to measure inscreen for some reason ? we will store the reason. - Get device information i.e. fine grained device resolution for iPhone models which not available in user agent HTTP header, it helps us to deliver better creative next time.

jonasmattsson1:
@mike-chowla Hallo Sir, Is there anything more that we have to fix in the adapter to get it approved?

Br,
Jonas Mattsson

PR #2289, opened 2018-03-20
add optional param to bridgewellBidAdapter

wuleo:
## Type of change

- [x] Feature

## Description of change

Add parameter `cpmWeight` that allows publisher to alter (scale) bid prices from Bridgewell.

- contact email of the adapter’s maintainer: leowu@bridgewell.com

### DIFF
--- a/modules/bridgewellBidAdapter.js
+++ b/modules/bridgewellBidAdapter.js
@@ -100,7 +100,6 @@ export const spec = {
         }
 
         bidResponse.requestId = req.bidId;
-        bidResponse.requestId = req.bidId;
         bidResponse.cpm = matchedResponse.cpm * req.params.cpmWeight;
         bidResponse.width = matchedResponse.width;
         bidResponse.height = matchedResponse.height;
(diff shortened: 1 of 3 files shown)


wuleo:
Some publishers in Taiwan may use this optional `cpmWeight` to prefer or not prefer Bridgewell’s bid price.
For instance, if a publisher set `cpmWeight` 0.5, this means they don’t want to prefer Bridgewell, they scale down Bridgewell’s bid price to 50% i.e. 0.5*(Bridgewell’s bid price).
In the other hand, if they set `cpmWeight` 1.5, it  means they want to scale up Bridgewell’s bid price 50%, i.e. 1.5*(Bridgewell’s bid price).

jsnellbaker:
@wuleo That's fine, it'll be up to the publisher to set the value that makes sense given their needs.  I just wanted to comment on test params data in case it needs to be tested later on without difficulty.  The changes look good now so I'll merge them in.

wuleo:
Thanks for the review.

PR #2292, opened 2018-03-20
Make eslint aware of the custom import paths

terebentina:
By default eslint doesn't know what `src/Renderer` means in the following import:
`import { Renderer } from 'src/Renderer';`
This changes makes it aware that `src/Renderer` means `<project root>/src/Renderer.js`

## Type of change

- [ ] Bugfix
- [ ] Feature
- [ ] New bidder adapter   
- [ ] Build related changes
- [ ] CI related changes
- [x] Other

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -3,6 +3,13 @@ module.exports = {
     "browser": true,
     "commonjs": true
   },
+  "settings": {
+    "import/resolver": {
+      "node": {
+        "moduleDirectory": ["node_modules", "./"]
+      }
+    }
+  },
   "extends": "standard",
   "globals": {
     "$$PREBID_GLOBAL$$": false
--- a/modules/sonobiBidAdapter.js
+++ b/modules/sonobiBidAdapter.js
@@ -1,6 +1,5 @@
 import { registerBidder } from 'src/adapters/bidderFactory';
-import { getTopWindowLocation, parseSizesInput } from 'src/utils';
-import * as utils from '../src/utils';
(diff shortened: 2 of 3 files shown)


mike-chowla:
The Travis CI build is failing.

Why are custom import paths needed?

terebentina:
Haha, the fact that it is failing means this PR already proved its usefulness :)
The eslint check in Travis fails here:
https://github.com/prebid/Prebid.js/blob/master/modules/sonobiBidAdapter.js#L2-L3
Until now eslint didn't know that `src/utils` and `../src/utils` are one and the same thing (in this case). Now it knows and it correctly identified this problem - you shouldn't import the same file multiple times.

To answer your question: line 2 in that adapter is an absolute import. By default nodejs resolves absolute imports relative to the `node_modules` folder. In this case, there's no `node_modules/src/utils.js` path so if this would be run directly with nodejs it would fail.
You are passing all files through webpack though and already told webpack how it should resolve these special cases: https://github.com/prebid/Prebid.js/blob/master/webpack.conf.js#L15-L20 : look for the file either in `node_modules` or in the root folder.
This PR simply makes eslint aware of the same rules.

mkendall07:
@terebentina 

Thanks for the PR. Can you update the PR so it passes (fix the import in the sonobi adapter etc). Thanks

mkendall07:
bump

terebentina:
Sorry, yes. This weekend!

terebentina:
all good!

PR #2311, opened 2018-03-23
Fixes #2276 - video.playerSize and Size Mapping not working together

### LINKED ISSUE #2276: Size mapping not working for outstream adunit
## Type of issue
Bug

## Description
If we have 4 adunits on a page, in which 3 adunits are just banners and 1 with outstream, the Size mapping /sizeConfig will not work for outstream. When requesting for bid the outstream adunit is always ignored. If I remove the sizeConfig and labels it works fine.

## Steps to reproduce
Have a page with 2 adunits one with banner another with Outstream and enable Size mapping /sizeConfig

### Expected results
Both Outstream and Banners should work with Size mapping /sizeConfig

## Platform details
Prebid.js 1.4.0

jsnellbaker:
## Type of change
- [x] Bugfix

## Description of change
Currently, using Size Mapping functionality with a video adUnit that specifies `mediaTypes.video.playerSize` would not pass the `sizeMapping.js` validation checks causing the ad unit to not be included in the `bidRequest` object.  

As a quick recap, the validation check in the sizeMapping (located here https://github.com/prebid/Prebid.js/blob/master/src/sizeMapping.js#L41) does the following:
- uses a list of arrayed sizes (ie `[[300, 250], [300, 600]]`) as a list of acceptable sizes (this is derived from the matching mediaQuery params).
- It reads the adUnit's size params and cross checks that array against the mediaQuery list of sizes
- if it finds a match, the size passes the filter
- otherwise if no sizes are matched, the bid doesn't get added to the request

Because the filtering operates on a list of arrayed sizes, the adUnits.sizes has to use the same format.  Otherwise, the filter would treat the single array of sizes ie `[300, 250]` as two integers and the size wouldn't pass the check.

To meet this requirement, I changed the `adUnits.mediaTypes.video.playerSize` field to accept either format (ie `[[640, 480]]` or `[640, 480]`) but it will automatically format the latter syntax into the former.

A note - while the `playerSize` field will expect an array of an array, there should only 1 set of sizes in the field.  The changes in the `checkBidRequestSizes()` function check for this exact type of setup.

## Other information
fixes issue reported in #2276

### DIFF
--- a/src/adaptermanager.js
+++ b/src/adaptermanager.js
@@ -240,6 +240,11 @@ exports.checkBidRequestSizes = (adUnits) => {
       if (video.playerSize) {
         if (Array.isArray(video.playerSize) && video.playerSize.length === 1 && video.playerSize.every(isArrayOfNums)) {
           adUnit.sizes = video.playerSize;
+        } else if (isArrayOfNums(video.playerSize)) {
+          let newPlayerSize = [];
+          newPlayerSize.push(video.playerSize);
+          utils.logInfo(`Transforming video.playerSize from ${video.playerSize} to ${newPlayerSize} so it's in the proper format.`);
+          adUnit.sizes = video.playerSize = newPlayerSize;
         } else {
           utils.logError('Detected incorrect configuration of mediaTypes.video.playerSize.  Please specify only one set of dimensions in a format like: [[640, 480]]. Removing invalid mediaTypes.video.playerSize property from request.');
           delete adUnit.mediaTypes.video.playerSize;
--- a/src/adaptermanager.js
+++ b/src/adaptermanager.js
@@ -215,7 +215,12 @@ exports.makeBidRequests = function(adUnits, auctionStart, auctionId, cbTimeout, 
 };
 
 exports.checkBidRequestSizes = (adUnits) => {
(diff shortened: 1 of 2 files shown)


src/adaptermanager.js
```diff
@@ -215,7 +215,12 @@ exports.makeBidRequests = function(adUnits, auctionStart, auctionId, cbTimeout,
 };
 
 exports.checkBidRequestSizes = (adUnits) => {
-  Array.prototype.forEach.call(adUnits, adUnit => {
+  function isArrayOfNums(val) {
+    return Array.isArray(val) && val.length === 2 && utils.isInteger(val[0]) && utils.isInteger(val[1]);
+  }
+
+  // Array.prototype.forEach.call(adUnits, adUnit => {
```
matthewlane: Can drop this line if no longer needed

jsnellbaker: Removed it.

jsnellbaker:
@matthewlane  Per your separate feedback, I added support for the `mediaTypes.video.playerSize` field to be able to take either `[[640, 480]]` or `[640,480]` syntax styles.  When the code detects the latter style, it automatically converts it to the former style to ensure it works with sizeMapping functionality.    

I'll update the original description in lieu of the functionality change.

Issue #2238, opened 2018-03-07
pbjs.getAllWinningBids() not working correctly


snapwich:
## Type of issue
**bug**

## Description
Calling `pbjs.getAllWinningBids()` will not work correctly if you have auctions that have multiple adUnits and therefore have multiple winners.  I think the issue resides with the logic in `auction.js` here: https://github.com/prebid/Prebid.js/blob/master/src/auction.js#L200

An auction instance only allows for setting and getting a singular winning bid when it should allow for multiple winning bids.

Also while looking into this I noticed that `auction.setWinningBid` and `auctionManger.getAllWinningBids` all function around calls to `pbjs.renderAd`, which is somewhat confusing since it's a different code path than what is used in targeting for getting winning bids.  I think either `pbjs.getAllWinningBids()` should use a similar code path that is used in `targeting.getWinningBids` to eliminate the possibility of discrepancies between the two OR `pbjs.getAllWinningBids()` should be renamed to eliminate confusion, maybe `pbjs.getAllRenderedBids()` or something?

## Steps to reproduce
Create an auction with multiple adUnits and call `pbjs.getAllWinningBids()` after the auction has ran.

### Expected results
It returns all of the bids rendered.

### Actual results
It returns the last bid rendered.

## Other information
This causes the issues highlighted in #2190

snapwich:
After discussing, I think we've decided that, to help prevent discrepancies and confusion around naming, `pbjs.getAllWinningBids()` should be modified to use the similar code paths that are used in the targeting code to determine winning bids.  To fulfill the need for getting rendered bids, a new API method should be added: `pbjs.getAllRenderedBids()`

chefbenjamin:
Just curious where this is right now ?

jaiminpanchal27:
@chefbenjamin We will ship this in next release

chefbenjamin:
@jaiminpanchal27 1.6 ?

chefbenjamin:
@jaiminpanchal27 @mkendall07 @snapwich  I see that https://github.com/prebid/Prebid.js/pull/2328 is finished. Does the snippet here have to be changed ? http://prebid.org/dev-docs/toubleshooting-tips.html#see-all-bids-in-the-console

also would be good to add a column here which showed if the bid is client side or server side.

chefbenjamin:
[...]

PR #2328, opened 2018-03-29
Added bid pool and fixed getAllWinningBids function

jaiminpanchal27:
## Type of change
- [x] Bugfix
- [x] Feature

## Description of change
This PR adds bid pool concept to prebid.js 
In case of concurrent auctions, auctionManager can have multiple bids from same bidder for a particular adUnit. While selecting bids to do auction, prebid.js uses following criteria if such multiple bids are found.
- Do not use bid which is rendered or won earlier auction.
- Select bid whose `ttl` is not expired. Here buffer of ~~200~~ 1000 ms is included in check.
- Use first in first out. Select the bid which came in first for auction.

Also added new api function `getAllPrebidWinningBids` which will return the bids won in prebid auction but not rendered by dfp.

Also fixed the bug in `getAllWinningBids` described here #2238


src/targeting.js
```diff
@@ -10,14 +10,29 @@ var CONSTANTS = require('./constants.json');
 var pbTargetingKeys = [];
 
 export const BID_TARGETING_SET = 'targetingSet';
+export const RENDERED = 'rendered';
 
 const MAX_DFP_KEYLENGTH = 20;
+const TTL_BUFFER = 200;
```
mkendall07: let's make the buffer 1000

src/auction.js
```diff
@@ -197,8 +197,8 @@ export function newAuction({adUnits, adUnitCodes, callback, cbTimeout, labels})
     executeCallback,
     callBids,
     bidsBackAll,
-    setWinningBid: (winningBid) => { _winningBid = winningBid },
-    getWinningBid: () => _winningBid,
+    setWinningBid: (winningBid) => { _winningBids = _winningBids.concat(winningBid) },
```
snapwich: Seems `addWinningBid` is a more apt name now.

src/auction.js
```diff
@@ -197,8 +197,8 @@ export function newAuction({adUnits, adUnitCodes, callback, cbTimeout, labels})
...
+    getWinningBid: () => _winningBids,
```
snapwich: and `getWinningBids` here.

jaiminpanchal27:
Docs PR here https://github.com/prebid/prebid.github.io/pull/700

PR #2330, opened 2018-03-29
added nested iframe resizing code

mkendall07:
## Type of change
- [x] Feature

## Description of change
In the case of multiple nested iframes, the render function was only resizing the last one, not any in between the parent page and itself.

This attempts to resize them all.

### DIFF
--- a/src/prebid.js
+++ b/src/prebid.js
@@ -105,6 +105,19 @@ function setRenderSize(doc, width, height) {
   if (doc.defaultView && doc.defaultView.frameElement) {
     doc.defaultView.frameElement.width = width;
     doc.defaultView.frameElement.height = height;
+    // in case of nested iframes, resize those as well
+    try {
+      if (doc.defaultView.parent !== window.top) {
+        setRenderSize(doc.defaultView.parent.document, width, height);
+      }
+    } catch (e) {
+      // x-domain try one more time.
+      try {
+        setRenderSize(doc.defaultView.parent.document, width, height);
+      } catch (e) {
+        // reached the top window in x-domain
+      }
+    }
   }
 }
 


src/prebid.js
```diff
@@ -105,6 +105,19 @@ function setRenderSize(doc, width, height) {
   if (doc.defaultView && doc.defaultView.frameElement) {
     doc.defaultView.frameElement.width = width;
     doc.defaultView.frameElement.height = height;
+    // in case of nested iframes, resize those as well
+    try {
+      if (doc.defaultView.parent !== window.top) {
+        setRenderSize(doc.defaultView.parent.document, width, height);
+      }
+    } catch (e) {
+      // x-domain try one more time.
```
harpere: why would it work after it just failed?

src/prebid.js
```diff
@@ -105,6 +105,19 @@ function setRenderSize(doc, width, height) {
   if (doc.defaultView && doc.defaultView.frameElement) {
     doc.defaultView.frameElement.width = width;
     doc.defaultView.frameElement.height = height;
+    // in case of nested iframes, resize those as well
```
harpere: I guess there's no chance of resizing an iframe unrelated to the ad because it would be x-domain?

src/prebid.js
mkendall07: Sorry I should have made the comment more clear. This code is executed if there is an x-domain in the chain (because of throw we know this), however the most immediate parent may not be x-domain so it could potentially be resized. If it throws again we know we hit the x-domain so we are done

harpere: but wouldn't it have already been resized the first time the function was called?  that's the first thing it does.

Also, I ran a test in chrome and I don't ever see an error thrown.  When there's a x-domain iframe the `if (doc.defaultView && doc.defaultView.frameElement) {` condition fails and the function exits silently.  Maybe this would work differently in other/older browsers?

mkendall07: oh interesting. I'll look more into this. Thanks

mkendall07: so the case this catches is for window => x-domain iframe => friendly Iframe => ad document iframe => ad

So the first pass will resize the ad document iframe. Then it will check against if `parent !== window.top` this will throw right away because of the x-domain iframe in the chain, however, `doc.default.view.parent` is pointing to a friendly iframe so it sends that through as the next resize option (`doc.defaultView.parent.document`). Hopefully that's clear.

src/prebid.js
mkendall07: it's possible I suppose if someone is using nested iframes for the website structure. I don't think that's very common with modern HTML though?

mkendall07:
Closing this because it's impossible to detect if an iframe has content or ad markup and we could accidentally resize content to the ad size.

Issue #2315, opened 2018-03-23
sizeConfig[Conditional Ad Units] does not work when s2sConfig is enabled.


mercuryyy:
trying to set sizeConfig from this doc for 1.7 prebid http://prebid.org/dev-docs/conditional-ad-units.html

When using the new sizeConfig[Conditional Ad Units] the adunits for the s2s are null and for bidders not in s2s it send them correctly. 

I tested few configuration and when sizeConfig is enabled with  the new Conditional Ad Units it does not send the ad units to s2s bidders.

once i remove the bidders from s2sConfig everything fires correctly.

mkendall07:
@mercuryyy 
do you have an example page?

mercuryyy:
@mkendall07  i used this pull and it fixes some of the issues - https://github.com/javipm/Prebid.js

Any configuration you put in with sizeconfig and then enable a bidder in s2s it wont send the bids to the s2s adapter

jsnellbaker:
@mercuryyy  This should be fixed now in 1.7 release.  If you continue to see any issues, please let us know.

PR #2332, opened 2018-04-02
fix issue #2315 sizeMapping not working with s2s requests

jsnellbaker:
## Type of change
- [x] Bugfix

## Description of change
This PR addresses the issue reported in #2315 and a subsequent issue found during testing.  

The issue with the sizeMapping functionality was due to the `adUnit.sizes` being transformed from a list of integer sizes (ie `[[300, 250], [300, 600]]` to a list of objects (ie `[{w: 300, h: 250}, {w: 300, h: 600}]`).  The sizeMapping code expected a list of integer sizes, not the object format sizes - so it never passed the filtering check properly.

The fix places the remapped object sizes to a different variable (`adUnit.sizesS2S`) temporarily so the original sizes are kept in the right format for the sizeMapping check.  Later in the prebidServer adapter, the reassigned sizes are placed properly under the adUnit.sizes so it use the correct format for prebid server.

During testing for this fix, a subsequent issue was observed and fixed.  In this issue, a bidder that was rejected by the sizeMapping still had the opportunity to participate in the pbs auction and could win the bid.  This is due to the fact that the sizeMapping functionality filters out the invalid `adUnit` bidders and simply doesn't include them in the bids objects that are made in the request object.  The sizeMapping functionality doesn't directly edit/remove the invalid bidders from the `adUnits`.  This indirectly causes the issue, since the the actual request object made for the pbs request is built separately from the client-side request and it uses the `adUnits` unmodified object.

The fix for this issue is to introduce a filter that removes any bidders from the `adUnits` object if there is no corresponding `bid` in the bidder's client-side request for that `adUnit`.  It performs a two stage check, first against the `adUnit.code` with the `request.bid.adUnitCode` (in case the same bidder is in multiple adUnits) and second against the `adUnit.bid.bidder` with the `request.bidderCode`.  

## Other information
Fixes issue #2315

### DIFF
--- a/src/adaptermanager.js
+++ b/src/adaptermanager.js
@@ -113,7 +113,7 @@ function getAdUnitCopyForPrebidServer(adUnits) {
   let adUnitsCopy = utils.deepClone(adUnits);
 
   adUnitsCopy.forEach((adUnit) => {
-    adUnit.sizes = transformHeightWidth(adUnit);
+    adUnit.sizesS2S = transformHeightWidth(adUnit);
 
     // filter out client side bids
     adUnit.bids = adUnit.bids.filter((bid) => {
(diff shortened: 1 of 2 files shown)


src/adaptermanager.js
```diff
@@ -113,7 +113,7 @@ function getAdUnitCopyForPrebidServer(adUnits) {
   let adUnitsCopy = utils.deepClone(adUnits);
 
   adUnitsCopy.forEach((adUnit) => {
-    adUnit.sizes = transformHeightWidth(adUnit);
+    adUnit.sizesS2S = transformHeightWidth(adUnit);
```
mkendall07: can we just move this function into prebidServerBidAdapter instead of adding a temporary sizesS2S?

mkendall07:
This should be good once the change is made

src/adaptermanager.js
jsnellbaker: Yep - I'll move the function over and make other adjustments.

jsnellbaker:
@mkendall07 I just pushed in the changes.

PR #2351, opened 2018-04-04
send appnexus usePaymentRule info to prebid-server ortb request

jsnellbaker:
## Type of change
- [x] Feature

## Description of change
The `adUnits.bids.params.usePaymentRule` was previously only used in the `appnexusBidAdapter` (where it was formatted and included in the outgoing request).  Now this same field will be used by the `prebidServerBidAdapter` and passed along to prebid-server.

For the pbs to properly ingest the field, it needed to be set in the `Request.imp.ext.appnexus.use_pmt_rule` field of the ortb request.

The `prebidServerBidAdapter` file already had the support it needed to read the `adUnits.bids.params` value and set it in the proper location noted above.  The only change needed was to rename the `adUnits.bid.params.usePaymentRule` (as it is defined in the prebid page code) to `adUnits.bid.params.use_pmt_rule` so it had the proper field name for pbs.

This renaming is only done for the `appnexus` bidder; if any other bidder happened to have the `usePaymentRule` value set in their `bid.params` it would be left as is.


jsnellbaker:
@jaiminpanchal27 Per our offline discussion, I have moved the code changes out of the `adaptermanager.js` file and into the `prebidServerBidAdapter.js` file.

test/spec/unit/core/adapterManager_spec.js
```diff
@@ -851,7 +851,90 @@ describe('adapterManager tests', () => {
           expect(bidRequests[0].bids.length).to.equal(1);
           expect(bidRequests[0].bids[0].adUnitCode).to.equal(adUnits[1].code);
         });
-      })
+      });
+
+      // describe('other checks', () => {
```
jaiminpanchal27: Can you remove this code if not required

PR #2361, opened 2018-04-05
Adding onRenderFail in Adapter Spec

### LINKED ISSUE #2303: Notify bidder on adRenderFailed event
## Type of issue
feature request
## Description

A new `adRenderFailed` event is created which fires when ad render fails. Even bidder should be notified if ad render fails because of ad code. This can be callback to public function in bidder spec similar to `getUserSyncs `, `interpretResponse` etc. and receiving bid data. 

## Other information
This is extension of #2210 .

vedantseta:
## Type of change

- [ ] Feature

## Description of change

This adds a function `onRenderFail` to bidder spec, letting bidders to subscribe when their Ad Fails to render.

Payload received to this function is same as Bid object.

PR on  [prebid.github.io](https://github.com/prebid/prebid.github.io/) will be added soon 

## Other information
This resolves #2303

### DIFF
--- a/src/adaptermanager.js
+++ b/src/adaptermanager.js
@@ -437,3 +437,19 @@ exports.getBidAdapter = function(bidder) {
 exports.setS2STestingModule = function (module) {
   s2sTestingModule = module;
 };
+
+exports.callRenderFailBidder = function(bid) {
+  const bidder = bid.bidderCode || '';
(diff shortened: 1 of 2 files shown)


jaiminpanchal27:
@vedantseta Please add unit tests

PR #2367, opened 2018-04-06
Add note about headerbid expert

matthewlane:
## Type of change
- Documentation

## Description of change
Adds note to PR review doc about headerbid expert


snapwich:
@matthewlane I don't know why this is failing the build, I restarted it twice with the same result.  My only guess is that it was branched off of master when the travis configuration was in a bad state; it's possible a rebase off the latest master will fix the travis issues.

PR #2370, opened 2018-04-09
Update Criteo bid adapter to Prebid 1.x

Spark-NF:
## Type of change
- [x] New bidder adapter
- [x] Official adapter submission

## Description of change
- Test parameters for validating bids
```
{
    code: '...',
    sizes: [[300, 250], [728, 90]],
    bids: [{
        bidder: 'criteo',
        params: {
            zoneId: 497747
        }
    }]
}
```
- Adapter maintainers: pi-direct@criteo.com, n.faure@criteo.com, a.hubert@criteo.com
- The bidders params documentation [already exists](https://github.com/prebid/prebid.github.io/blob/master/dev-docs/bidders/criteo.md) and does not require an update as the params didn't change

## Other information
- The old Prebid 0.x adapter can be found [here](https://github.com/prebid/Prebid.js/blob/legacy/modules/criteoBidAdapter.js) on the legacy branch

### DIFF
--- a/modules/criteoBidAdapter.js
+++ b/modules/criteoBidAdapter.js
@@ -33,9 +33,14 @@ export const spec = {
     let url;
     let data;
 
-    // If publisher tag not already loaded try to get it from fast bid else load it
-    if (typeof Criteo === 'undefined' && !tryGetCriteoFastBid()) {
-      loadScript(PUBLISHER_TAG_URL);
+    // If publisher tag not already loaded try to get it from fast bid
+    if (typeof Criteo === 'undefined') {
+      tryGetCriteoFastBid();
+
+      // Reload publisher tag after prebid timeout to ensure fast bid is up to date
+      setTimeout(() => {
+        loadScript(PUBLISHER_TAG_URL);
+      }, bidderRequest.timeout);
     }
 
     if (typeof Criteo !== 'undefined') {
--- a/modules/criteoBidAdapter.js
+++ b/modules/criteoBidAdapter.js
@@ -264,6 +264,13 @@ function registerEventHandlers() {
       });
   });
 
+  events.on(EVENTS.SET_TARGETING, () => {
+    const adapters = Criteo.PubTag.Adapters.Prebid.GetAllAdapters();
+    for (const k in adapters) {
+      adapters[k].handleSetTargeting();
+    }
+  });
+
   registeredEvents = true;
 }
 
--- a/modules/criteoBidAdapter.js
+++ b/modules/criteoBidAdapter.js
@@ -35,7 +35,8 @@ export const spec = {
 
     // If publisher tag not already loaded try to get it from fast bid
     if (!publisherTagAvailable()) {
-      window.Criteo = { usePrebidEvents: false };
+      window.Criteo = window.Criteo || {};
+      window.Criteo.usePrebidEvents = false;
 
       tryGetCriteoFastBid();
 
--- a/modules/criteoBidAdapter.js
+++ b/modules/criteoBidAdapter.js
@@ -59,7 +59,9 @@ export const spec = {
   interpretResponse: (response, request) => {
     if (typeof Criteo !== 'undefined') {
       const adapter = Criteo.PubTag.Adapters.Prebid.GetAdapter(request);
-      return adapter.interpretResponse(response.body, request);
+      if (adapter) {
+        return adapter.interpretResponse(response.body, request);
+      }
     }
 
     const bids = [];
--- a/modules/criteoBidAdapter.js
+++ b/modules/criteoBidAdapter.js
@@ -1,9 +1,12 @@
 import { loadScript } from 'src/adloader';
 import { registerBidder } from 'src/adapters/bidderFactory';
+import { EVENTS } from 'src/constants';
 import { parse } from 'src/url';
 import * as utils from 'src/utils';
 
-const ADAPTER_VERSION = 3;
+var events = require('src/events');
+
+const ADAPTER_VERSION = 4;
 const BIDDER_CODE = 'criteo';
 const CDB_ENDPOINT = '//bidder.criteo.com/cdb';
 const INTEGRATION_MODES = {
@@ -50,6 +53,8 @@ export const spec = {
       const adapter = new Criteo.PubTag.Adapters.Prebid(PROFILE_ID, ADAPTER_VERSION, bidRequests, bidderRequest);
       url = adapter.buildCdbUrl();
       data = adapter.buildCdbRequest();
+
+      registerEventHandlers();
     } else {
       const context = buildContext(bidRequests);
       url = buildCdbUrl(context);
@@ -234,6 +239,32 @@ function tryGetCriteoFastBid() {
     // Unable to get fast bid
   }
   return false;
+}
+
+// Register events to know when Criteo won the bid or timeouted
+let registeredEvents = false;
+function registerEventHandlers() {
+  if (registeredEvents) {
+    return;
+  }
+
+  events.on(EVENTS.BID_WON, (bid) => {
+    if (bid.bidderCode === 'criteo') {
+      const adapter = Criteo.PubTag.Adapters.Prebid.GetAdapter(bid.auctionId);
+      adapter.handleBidWon(bid);
+    }
+  });
+
+  events.on(EVENTS.BID_TIMEOUT, (timedOutBidders) => {
+    timedOutBidders
+      .filter(bidder => bidder.bidder === 'criteo')
+      .map(bidder => {
+        const adapter = Criteo.PubTag.Adapters.Prebid.GetAdapter(bidder.auctionId);
+        adapter.handleBidTimeout();
+      });
+  });
+
+  registeredEvents = true;
 }
 
 registerBidder(spec);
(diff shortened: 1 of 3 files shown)


modules/criteoBidAdapter.js
```diff
@@ -0,0 +1,277 @@
...
+var events = require('src/events');
+
+const ADAPTER_VERSION = 4;
+const BIDDER_CODE = 'criteo';
+const CDB_ENDPOINT = '//bidder.criteo.com/cdb';
+const INTEGRATION_MODES = {
+  'amp': 1,
+};
+const PROFILE_ID = 207;
+const PUBLISHER_TAG_URL = '//static.criteo.net/js/ld/publishertag.prebid.js';
```
mkendall07: Can you put a comment here point to https://github.com/Prebid-org/prebid-js-external-js-criteo/blob/master/dist/prod.js as source?

Spark-NF: Done.

modules/criteoBidAdapter.js
```diff
@@ -0,0 +1,279 @@
...
+}
+
+/**
+ * @return {boolean}
+ */
+function tryGetCriteoFastBid() {
+  try {
+    const fastBid = localStorage.getItem('criteo_fast_bid');
+    if (fastBid !== null) {
+      eval(fastBid); // eslint-disable-line no-eval
```
mkendall07: is there an alternative to using eval here? What's being done?

modules/criteoBidAdapter.js
```diff
@@ -0,0 +1,279 @@
...
+ * @param {*} callback
+ * @return {string}
+ */
+function createNativeAd(id, payload, callback) {
+  // Store the callback and payload in a global object to be later accessed from the creative
+  window.criteo_prebid_native_slots = window.criteo_prebid_native_slots || {};
+  window.criteo_prebid_native_slots[id] = { callback, payload };
+
+  // The creative is in an iframe so we have to get the callback and payload
+  // from the parent window (doesn't work with safeframes)
```
mkendall07: do you have a postMessage solution that would work in safeFrames?

modules/criteoBidAdapter.js
```diff
@@ -0,0 +1,279 @@
...
+    // If publisher tag not already loaded try to get it from fast bid
+    if (!publisherTagAvailable()) {
+      window.Criteo = window.Criteo || {};
+      window.Criteo.usePrebidEvents = false;
+
+      tryGetCriteoFastBid();
+
+      // Reload the PublisherTag after the timeout to ensure FastBid is up-to-date and tracking done properly
+      setTimeout(() => {
+        loadScript(PUBLISHER_TAG_URL);
```
mkendall07: We are going to update this function signature to include module/bidder code so we can track who is loading external JS. Will reply shortly with the changes.

modules/criteoBidAdapter.js
```diff
@@ -0,0 +1,279 @@
+import { loadScript } from 'src/adloader';
+import { registerBidder } from 'src/adapters/bidderFactory';
+import { EVENTS } from 'src/constants';
+import { parse } from 'src/url';
+import * as utils from 'src/utils';
+
+var events = require('src/events');
```
mkendall07: prefer const.

modules/criteoBidAdapter.js
```diff
@@ -0,0 +1,279 @@
...
+// Register events to know when Criteo won the bid or timeouted
+let registeredEvents = false;
+function registerEventHandlers() {
+  if (registeredEvents) {
+    return;
+  }
+
+  registeredEvents = true;
+
+  events.on(EVENTS.BID_WON, (bid) => {
```
mkendall07: we currently have a policy that bidder adapters cannot consume internal prebid.js events. You'll need to remove these event listeners.

modules/criteoBidAdapter.js
ahubertcriteo: Maybe, we will study this possibility and push it in a second iteration if feasable.

modules/criteoBidAdapter.js
ahubertcriteo: - Can I ask what is the rational of this policy? 
- On our side we wanted to track those events to help publishers with their integrations when they encounter a lot of timeouts 
(some of them have inconsistencies in their timeouts configuration between the one passed in requestBids call and the one in the global setTimeout)
- As documented here (http://prebid.org/dev-docs/publisher-api-reference.html#module_pbjs.onEvent), can we replace the call to "events.on" by onEvent ?
- If not what would be the right alternative way to collect those informations ?

modules/criteoBidAdapter.js
mkendall07: very interesting. Going to discuss with the prebid team

mkendall07: Thinking about this some more, I don't think this is safe since you are blindly reading in content from localStorage, which can be overwritten by any script on page. Please remove.

modules/criteoBidAdapter.js
mkendall07: that's fine

modules/criteoBidAdapter.js
mkendall07: Reason is that we don't want vendors inspecting other vendor bids. I realize that you are not doing that but we want to make this preventable in the code. 

The correct way is not fully implemented yet, but you could start with `spec.onTimeout` function.
http://prebid.org/dev-docs/bidder-adaptor.html#registering-on-timeout

Other events can be added to the spec in a separate PR if desired. Thanks

mkendall07:
@Spark-NF @ahubertcriteo 
If you can make the requested changes (drop `eval` and `event`) we can merge this for now and work on further enhancements in new PRs.

modules/criteoBidAdapter.js
mkendall07: @ahubertcriteo @Spark-NF 
Please use `loadExternalScript` method instead.
This will enforce a whitelist of which adapters are allowed to load external JS: 
https://github.com/prebid/Prebid.js/blob/master/src/adloader.js#L9

modules/criteoBidAdapter.js
mkendall07: @ahubertcriteo 
Do you have any stats that show this is faster than a modern browser caching of the parsed code? It doesn't really make sense to eval (parse) it twice.

Spark-NF: Yes, we made a lot of testing with our publishers, and results are very positive.

The time between the call to `buildRequests`/`callBids` and the call to the PublisherTag entry point (in this adapter `new Criteo.PubTag.Adapters.Prebid`) was reduced by 80 to 95%, going from a few hundreds milliseconds (sometimes more than 1s depending on the publisher) to a few milliseconds (usually around 10ms).

The number of timeouts as detected by Prebid was also halved.

Even if the file was cached by the browser, the fact that we're simply adding a `<script>` causes it to have a pretty low priority in the loading chain. Plus, it's asynchronous as opposed to the localStorage that is synchronous (and fast).

bretg:
@Spark-NF and @ahubertcriteo - any further progress on this PR?

pahund:
@mkendall07 have you had a chance to review this yet? it seems the creators of the pull requests have done the changes you requested.

cwbeck:
Would be nice to get this merged in...

benjaminclot:
How did this get committed with `eval` still there? 🙁

piwanczak:
How about the convention "No loading of external libraries: All code must be present in the adapter, not loaded at runtime."?
Should it be signed as redundant?

benjaminclot:
@mkendall07 Are you going to enforce rules on this adapter or is Criteo going to get a free pass?

mafaigna:
Dear @benjaminclot , @piwanczak ,

Thank you for your interest in Criteo Direct Bidder.

On the question above (usage of external library), Criteo follows the policies put in place by Prebid.org. Our adapter indeed makes use of our external library, so as to ensure that you can smoothly benefit from some of our latest updates. In doing so, we are following a review process from Prebid.org and comply with the the instructions and policy described by Prebid’s external JS repository mechanism (see [link](https://github.com/prebid/prebid-js-external-js-template)). Would you have any question, feel free to reach out your Criteo representative.

On Fastbid/eval: the function is being used to ensure correct working of our adapter in Prebid 1.0 and doesn’t represent a risk per se. In parallel, we are working on alternative versions, which will soon be submitted to our Prebid repo. In the meantime, would you have any concern, please contact your Criteo representative so as to address it.

Kind regards,
Maxence

piwanczak:
@mkendall07 Could you please state the official approach by Prebid team? 
My key points are:
- I've seen a few external js used in the branch 0.X, but only Criteo in 1.X
- prebid-js-external-js-template seems pretty lonely with 14 watches and zero stars. This seems to point that it's nowhere to be found in official Prebid docs
- on the contrary - official docs forbid external JS's in adapters
- non-minified extenal code seems not-up-to-date or unavailable: 
https://github.com/Prebid-org/prebid-js-external-js-criteo/blob/master/dist/prod.js returns 404

As I strongly believe in "open" in opensource I feel it would be great for everyone to know the very same thing and be able to use the same tools. 


@mafaigna Thank you for your input. Unfortunately this does not answer my question. I'm glad to hear that you work on alternatives, though.

PR #2372, opened 2018-04-09
Merge remote-tracking branch 'prebid/master'

bengel-ias:
Expected behavior :
After integration the latest version of IAS prebid.js adapter and setting all up, the tool should call DFP with IAS key/values for each AdSlot defined on the page and a campaign (multiple creatives) that targets these key/values should then display on the page.

Actual behavior :
It seems that the key/values are set up for the 1st AdSlot of the page but not for the following AdSlots on the page (tested with "googletag.pubads().getSlots()[0].getTargeting("fr")" and "googletag.pubads().getSlots()[1].getTargeting("adt")").


bengel-ias:
Sorry, this doesn't have the correct commit (or any at all).

PR #2383, opened 2018-04-12
add getUserSyncs function in clickforceBidAdapter

MIGOdanis:
## Type of change

- [x] Feature

## Description of change
Add getUserSyncs function in the adapters


mike-chowla:
Code coverage on this adapter is zero because it has no tests!   

There's a couple of other issues I see:
(1) Responses are not coming back gzipped
(2) The test parameters from the original adapter submission are not returning an ad

mike-chowla:
The code itself looks fine but since there are no tests, it doesn't meet the 80% coverage requirement.  The initial adapter PR should have been rejected because there are no tests.

MIGOdanis:
The issues is begin our adserver update new version. so, at initial adapter PR test is work. 
What's next we need do?

mike-chowla:
There should be a unit file that gets to at least 80% code coverage on your adapter.

Here's an example:
https://github.com/prebid/Prebid.js/blob/master/test/spec/modules/appnexusBidAdapter_spec.js

MIGOdanis:
HI @mike-chowla

we have a unit file new, Please check it.

Thanks

mike-chowla:
Thanks for adding the test coverage file.

You are just shy of the 80% mark.   If you add any test for getUserSyncs(), you should get there.

Also, I never saw the updated test parameters.

If you can get these to me tomorrow (not too late Pacific time), I'll take another pass and we should be able to get you into version 1.9

MIGOdanis:
@mike-chowla 
thank you very much. i will try to add a getUserSyncs test in the  test coverage file.
and, about updated test parameters. where i can update this.

MIGOdanis:
@mike-chowla 

The test parameters already updated.  And I have add a getUserSyncs test in the test coverage file. Please check it.

Thank you very much

mike-chowla:
A couple of comments on items that would make life easier for reviewers:

Your Access-Control-Allow-Origin only works on default ports.  When I test on a separate port, it fails:
screen shot 2018-04-27 at 2 04 10 pm

The cpm on your test ad is 0.  Prebid doesn't select ads with zero CPM.

MIGOdanis:
ok. I will change a setting for the test ad.  thank you very much

PR #2385, opened 2018-04-12
convert AN bid params to underscore formatting for pbs

jsnellbaker:
## Type of change
- [x] Feature

## Description of change
Currently the AppNexus bid params are being passed in their normal camel-case format to PBS in the openrtb request.  This is not the correct format for PBS, which expects an underscore syntax.

This change automatically converts the AppNexus `bid.params` to the underscore syntax (if they need it) and removes the old camel-case keys so the data in the openrtb `request.imp.ext.appenxus` field is formatted properly.

### DIFF
--- a/test/spec/modules/prebidServerBidAdapter_spec.js
+++ b/test/spec/modules/prebidServerBidAdapter_spec.js
@@ -559,7 +559,6 @@ describe('S2S Adapter', () => {
 
       const myRequest = utils.deepClone(REQUEST);
 
-      debugger;//eslint-disable-line
       adapter.callBids(REQUEST, BID_REQUESTS, addBidResponse, done, ajax);
       const requestBid = JSON.parse(requests[0].requestBody);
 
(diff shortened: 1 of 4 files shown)


test/spec/modules/prebidServerBidAdapter_spec.js
```diff
@@ -550,6 +550,23 @@ describe('S2S Adapter', () => {
...
+
+    it('converts appnexus params to underscore syntax', () => {
+      const s2sConfig = Object.assign({}, CONFIG, {
+        endpoint: 'https://prebid.adnxs.com/pbs/v1/openrtb2/auction'
+      });
+      config.setConfig({s2sConfig});
+
+      const myRequest = utils.deepClone(REQUEST);
+
+      debugger;//eslint-disable-line
```
jaiminpanchal27: I always forget to remove this.

jsnellbaker: Sorry about that - just pushed a commit to remove it.

jsnellbaker:
Per our slack offline discussion I've added a TODO comment in the area of the code to indicate future work should be done.  Jaimin is going to brainstorm further about the new approach so there's no issue number yet to include with the comment.  If this is alright, please merge when possible.  Thanks.

PR #2391, opened 2018-04-12
deprecate loadScript and add loadExternalScript

mkendall07:
## Type of change

- [x] Feature

## Description of change
Use of adLoader.loadScript is deprecated as of prebid 1.x however it was not removed. Adding a deprecated tag here + adding lightweight function to just load a script.

### DIFF
--- a/src/adloader.js
+++ b/src/adloader.js
@@ -1,7 +1,35 @@
 var utils = require('./utils');
 let _requestCache = {};
 
-// add a script tag to the page, used to add /jpt call to page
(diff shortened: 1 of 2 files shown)


jsnellbaker:
@mkendall07 Do any docs need to be updated for this PR?

mkendall07:
@jsnellbaker 
I don't think so, `pbjs.loadScript` was an undocumneted API so it doesn't have any mention in the docs.

PR #2394, opened 2018-04-13
Smart: Add prebid version in the data payload

Spacedragoon:
## Type of change

- [ ] Bugfix
- [ ] Feature
- [ ] New bidder adapter   
- [ ] Build related changes
- [ ] CI related changes
- [x] Other

## Description of change

send the version of prebid used in the request.
Also, small change of the test parameter


ytyukhnin:
Hi @jsnellbaker ,

We have fixed the doc to split the test parameters for in-app and web traffic. You could use either. But for simplicity we would recommend to use the web's:

```
    var adUnits = [
           {
               code: 'test-div',
               sizes: [[300, 250]],  // a display size
               bids: [
                   {
                       bidder: "smart",
                       params: {
                            domain: 'http://ww251.smartadserver.com',
                            siteId: 207435,
                            pageId: 896536,
                            formatId: 62913,
                            ckId: 1122334455 // optional
                       }
                   }
               ]
           }
       ];
```

Thank you.

Issue #2355, opened 2018-04-04
Partner user syncs not called for OpenRTB endpoint in prebid server adapter


haohany:
## Type of issue
bug?

## Description
Partner user sync are called when interpreting the legacy response: https://github.com/prebid/Prebid.js/blob/master/modules/prebidServerBidAdapter.js#L341-L347
However, this functionality doesn't exist in the new OpenRTB endpoint.
The sync rate of server-side partners will be hurt if publishers move to the new endpoint.

I'm happy to submit a PR if this is indeed an issue.

dbemiller:
This may be because the `/openrtb2/auction` endpoint doesn't return the user syncs anywhere. It would need to be added there first.

But... we did discuss this, and didn't think it would hurt much (if at all) because `Prebid.js` _should_ still be hitting the `/cookie_sync` endpoint and logging the syncs from there.

Nobody really knew why both `/auction` and `/cookie_sync` returned cookie syncs, and I didn't find any discussions about the design on github... so I assumed it was probably an accident of history and left it out of `/openrtb2/auction`.

I haven't heard any complaints from OpenRTB publishers yet... but if you migrate and find the sync rate to be much lower, it's a good catch!

haohany:
@dbemiller I think you were talking about the sync between a partner and the prebid server. That part would be ok with the new endpoint as you mentioned.

The prebid user sync API recommends partners to define a `registerSyncs` function to drop their own pixels. But it's not called with the new endpoint. That would put the server-side partners at a disadvantage compared to the client-side counterparts.

PR #2410, opened 2018-04-17
Do client-side user syncs when using the OpenRTB endpoint

haohany:
## Type of change

- [x] Feature

## Description of change

Currently prebid only does client-side user syncs when using the legacy endpoint. This change adds that functionality to the OpenRTB endpoint. See https://github.com/prebid/Prebid.js/issues/2355 for details.


haohany:
@snapwich Thanks for the review. Could it be merged?

PR #2411, opened 2018-04-17
load_pbjs_dfp_concurrently.html improvement (removed obsolete code)

Slind14:
- prebid should be loaded as early as possible (gpt, too)
- removed googletag.pubadsReady check, this should be obsolete
- initAdserverSet should be set early on to avoid race conditions
- moved `googletag.pubads().refresh();` inside pbjs.que

## Type of change

- [ ] Bugfix
- [ ] Feature
- [ ] New bidder adapter   
- [ ] Build related changes
- [ ] CI related changes
- [x] Other

## Description of change
Unless I'm totally mistaken, the load_pbjs_dfp_concurrently.html example script is doing obsolete stuff.
I removed the obsolete googletag.pubadsReady check and loop including a few other smaller changes, see above.

### DIFF
--- a/integrationExamples/gpt/load_pbjs_dfp_concurrently.html
+++ b/integrationExamples/gpt/load_pbjs_dfp_concurrently.html
@@ -5,46 +5,37 @@
 * This is the optimized way of loading prebid.js. It loads both DFP and Prebid concurrently.
 * Publishers may wish to tweak the the MAX_RETRIES  and polling interval ms as desired to mitigate issue https://github.com/prebid/Prebid.js/issues/419
 */
-
+    
 //timeout to control how long bidders have to respond.
 var PREBID_TIMEOUT = 700;
-//how many times to attempt to load DFP if services have not been enabled
-var MAX_RETRIES = 20;
 
 var pbjs = pbjs || {};
 pbjs.que = pbjs.que || [];
-pbjs.retries = 0;
 var googletag = googletag || {};
 googletag.cmd = googletag.cmd || [];
 googletag.cmd.push(function () {
     googletag.pubads().disableInitialLoad();
 });
+
+// Load the Prebid Javascript Library Async. We recommend loading ASAP in page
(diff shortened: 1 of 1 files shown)


integrationExamples/gpt/load_pbjs_dfp_concurrently.html
```diff
@@ -5,47 +5,38 @@
...
 var googletag = googletag || {};
 googletag.cmd = googletag.cmd || [];
 googletag.cmd.push(function () {
     googletag.pubads().disableInitialLoad();
 });
+
+// Load the Prebid Javascript Library Async. We recommend loading ASAP in page
+loadScript('//acdn.adnxs.com/prebid/not-for-prod/prebid.js');
+//load DFP async.
+loadScript('//www.googletagservices.com/tag/js/gpt.js');
```
mkendall07: you can actually just use the standard `script` tag with the async attribute now since we are no longer supporting < IE 10. 
Ex: `<script async src = "//www.googletagservices.com/tag/js/gpt.js"></script>`

This will be more performant.

integrationExamples/gpt/load_pbjs_dfp_concurrently.html
```diff
@@ -5,47 +5,38 @@
...
   if (pbjs.initAdserverSet) return;
-
-  //we don't want to call DFP before it is ready
-  if(!googletag.pubadsReady && pbjs.retries <= MAX_RETRIES) {
-    setTimeout(initAdserver, 50); //poll ms can be adjusted as desired.
-    pbjs.retries++;
-    return;
-  }
+  pbjs.initAdserverSet = true;
+
```
mkendall07: Don't drop this check because it helps prevent a race condition where `googletag.pubads().refresh();` is queue before the inline page `googletag.display('div-id');` display call is made. If `.refresh()` executes first, no ads will be fetched (you can test this out easily with a GPT test page and setTimeout).

mkendall07:
@Slind14 
this will need updates.

PR #2414, opened 2018-04-17
Size Bugfix 

sami-elasticad:
## Type of change
- [X ] Bugfix

## Description of change


sami-elasticad:
Please review. Thnak you.

modules/quantumBidAdapter.js
```diff
@@ -104,7 +104,8 @@ export const spec = {
       bid.sync = serverBody.sync;
       if (bidRequest.renderMode && bidRequest.renderMode === 'banner') {
         bid.width = 300;
-        bid.height = 225;
+        bid.height = 250;
```
jsnellbaker: Are you sure you want these height/width values to be hard-coded like this?  Wouldn't it be more ideal to read from the  bid response object?  I thought I saw some height/width params in the response from the server, so I wanted to confirm on this point.

sami-elasticad:
You're right. removed response hardcoded sizes. Thank you.

PR #2417, opened 2018-04-18
Audience Network: Fix bid request validation for fullwidth

dlackty:
## Type of change

- [x] Bugfix

## Description of change
#2203 implemented a restriction that ensures "fullwidth" bids are 300x250. In `isBidRequestValid`, it requires **every** size in bid must be 300x250. It should instead just check if some bid size is 300x250, just like existing validation for banners.

## Other information
Ping @lovell original PR author & @harpere previous reviewer


lovell:
Thank you @dlackty 👍

PR #2430, opened 2018-04-19
added whitelist for loading external JS + tests

mkendall07:
## Type of change
- [x] Feature

## Description of change
Adds validation for whitelist, caching and tests.


test/spec/adloader_spec.js
```diff
@@ -1,4 +1,32 @@
+import * as utils from 'src/utils';
+import { assert } from 'chai';
```
jaiminpanchal27: I think this `assert` is not getting used anywhere

PR #2440, opened 2018-04-21
added missing dfpAdServerVideo.js dfp vast tag generation uri compone…

Slind14:
- [x] Bugfix

DFP urls built from sites containing special chars and or get params them self were invalid and causing dfp internal server errors.


Slind14:
Out of curiosity, is no one using this prebid feature or was it changed recently?

Slind14:
There was another issue with BuildDfpVideoUrl returning the URL without targeting (race condition). I fixed it removing this filter. I'm not sure if this is the right place.
https://github.com/prebid/Prebid.js/pull/2440/files#diff-d2172b2e0a1412bf50ecb9ba3fe4e973L178

PR #2447, opened 2018-04-23
Update babel plugins

jaiminpanchal27:
## Type of change
- [x] Build related changes

## Description of change
- Added babel-preset-env which replaces es2015, es2016, es2017 and latest.
- Removed some babel plugins which were required for IE8 IE9 which we don't support anymore.
- Tested babel-runtime-transform to add ES6 features without polluting global namespace but it adds some 20kb to prebid core and hence not part of this PR.
- Removed duplicate babel config from webpack conf. `babel-loader` uses `.babelrc` as default


jaiminpanchal27:
@mkendall07 No it does not add increase the size
Tested by building prebid with appnexusBidAdapter and its 90kb for both.

snapwich:
@jaiminpanchal27 shouldn't we specify our targets as described [here](https://babeljs.io/docs/plugins/preset-env/)?  Otherwise it seems to just include everything

PR #2452, opened 2018-04-25
OpenX Adapter: Support out of order bids.

jimee02:
## Type of change

- [x] Feature

### Feature
Added "idx" (index) so that we can return out of order bids will properly map back to request objects.
https://github.com/prebid/Prebid.js/pull/2452/files#diff-2051ef6e55abda4fed56905b5982884aR76
### Refactor
Broke down unit tests and re-ordered them such that test suites will be more readable.
Added helper functions to generate cleaner bid response objects.

### DIFF
--- a/test/spec/modules/openxBidAdapter_spec.js
+++ b/test/spec/modules/openxBidAdapter_spec.js
@@ -514,8 +514,8 @@ describe('OpenxAdapter', () => {
         pub_rev: '10000',
         html: '<div>OpenX Ad</div>'
       };
-      const adUnit = mockAdUnit(adUnitOverride, creativeOverride);
-      const bidResponse = mockArjResponse(undefined, [adUnit]);
+      let adUnit;
+      let bidResponse;
 
       let bid;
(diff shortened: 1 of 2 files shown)


test/spec/modules/openxBidAdapter_spec.js
```diff
@@ -615,18 +898,92 @@ describe('OpenxAdapter', () => {
...
+   * @param {Object} contract Original object contains the default fields
+   * @param {string} typeName Name of the type we're checking for error messages
+   * @throws {Error}
+   */
+  function overrideKeyCheck(override, contract, typeName) {
+    let invalidKey = Object.keys(override)
+      .find(key => !(key in contract));
+
+    if (invalidKey) {
+      throw new Error(`Mock Error: INVALID_KEY.  Key:${invalidKey} does not exist in type: ${typeName}`);
```
jaiminpanchal27: This will stop execution of further unit tests. Wrap all calls to `overrideKeyCheck` in `try..catch`

PR #2459, opened 2018-04-27
Update ReadPeak adapter (legacy)

kurrestahlberg:
This is essentially the same PR as #2369 but for legacy branch and with one extra fix for tests. It seems that `config.getConfig('publisherDomain')` returns different value in legacy and master.

## Type of change

- [x] Bugfix

## Description of change

Updates to the ReadPeak adapter to match changes to the ReadPeak system as well as some minor refactoring to clean up the adapter and make it a better citizen in Prebid.

kurre.stahlberg@readpeak.com


bretg:
@kurrestahlberg - please look into the travis-ci issue

PR #2471, opened 2018-04-30
PulsePoint Adapter GDPR support

anand-venkatraman:
## Type of change
- [x] Feature

## Description of change
GDPR support to PulsePoint Bid adapter.

Anand Venkatraman (avenkatraman@pulsepoint.com)


modules/pulsepointBidAdapter.js
```diff
@@ -42,6 +42,7 @@ export const spec = {
       app: app(bidRequests),
       device: device(),
     };
+    applyGdpr(bidRequests, request);
```
jsnellbaker: The GDPR consent data does not exist in the `bidRequests` object, but rather in the `bidderRequest` object (which is the second argument).

In this case, you'll need to modify the buildRequests function call like so:
```
  buildRequests: (bidRequests, bidderRequest) => {
``` 
Then you can pass in the `bidderRequest` value into the `applyGdpr` function (note - this will be a single object, not an array of objects).  You'll still access the consent data with `bidderRequest.gdprConsent`.

anand-venkatraman: Updated.

anand-venkatraman:
Submitted a PR to docs repo - https://github.com/prebid/prebid.github.io/pull/735

PR #2490, opened 2018-05-03
Refactor done callback. Fixes #2482

### LINKED ISSUE #2482: Auction closes early for Prebid Server Video when cache is enabled
## Type of issue
bug

## Description
For Prebid server video, auction closes before adding bid response when cache is enabled.

jaiminpanchal27:
## Type of change
- [x] Bugfix

## Description of change
Fixes #2482

### DIFF
--- a/src/auction.js
+++ b/src/auction.js
@@ -150,6 +150,15 @@ export function newAuction({adUnits, adUnitCodes, callback, cbTimeout, labels}) 
     }
   }
 
+  function hasVideoBids(bids) {
+    const bidsArray = bids ? (bids[0] ? bids : [bids]) : [];
+    const videoBid = bidsArray.some(bid => !!(utils.deepAccess(bid.mediaTypes, ['video'])));
+    const cacheEnabled = !!(config.getConfig('cache.url'));
+
+    // video bids with cache enabled need to be cached first before they are considered done
+    return (videoBid && cacheEnabled);
+  }
+
   function done(bidRequestId) {
     const innerBidRequestId = bidRequestId;
     return delayExecution(function() {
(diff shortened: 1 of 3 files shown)


src/auction.js
```diff
@@ -150,16 +150,28 @@ export function newAuction({adUnits, adUnitCodes, callback, cbTimeout, labels})
     }
   }
 
+  function hasVideoBids(bids) {
+    const bidsArray = bids ? (bids[0] ? bids : [bids]) : [];
+    const videoBid = bidsArray.some(bid => !!(utils.deepAccess(bid.mediaTypes, ['video'])));
+    const cacheEnabled = !!(config.getConfig('cache.url'));
+
+    // video bids with cache enabled need to be cached first before they are considered done
```
snapwich: I don't think this is quite right.  It's possible for video bids to skip the cache by returning the `bid.videoCacheKey` and `bid.vastUrl` themselves, even if the cache is enabled.  So this logic needs to mirror the logic in `tryAddVideoBid` when it decides whether to handle done or not.  Ideally that logic should probably be pulled out of `tryAddVideoBid` so it could be used there and here to be sure we don't have discrepancies, IMO.

src/auction.js
```diff
@@ -150,16 +150,36 @@ export function newAuction({adUnits, adUnitCodes, callback, cbTimeout, labels})
     }
   }
 
+  function hasVideoBids(bids) {
+    const bidsArray = bids ? (bids[0] ? bids : [bids]) : [];
+    const videoBid = bidsArray.some(bid => !!(utils.deepAccess(bid.mediaTypes, ['video'])));
+    const cacheEnabled = !!(config.getConfig('cache.url'));
```
mkendall07: this should move up to auction instance level.

Issue #2489, opened 2018-05-03
Sonobi adapter produces error on timeout


dmitriyshashkin:
## Type of issue
Bug

## Description
Sonobi adapter has the following line of code in the `getUserSyncs` method: `syncOptions.pixelEnabled && serverResponses[0].body.sbi_px`. This assumes that `serverResponses` array will not be empty. In fact when there was no response from the server (eg timeout occurred) the `serverResponses` variable passed to `getUserSyncs` by prebid might be an empty array. This results in the fllowing error: `Uncaught TypeError: Cannot read property 'body' of undefined`

If I'm not mistaken it might also interrupt the processing of other bid responses.

## Steps to reproduce
Set extremely low timeout, start auction with pixel sync enabled

### Expected results
Auction completes without errors

### Actual results
`Uncaught TypeError: Cannot read property 'body' of undefined` error occures

JonGoSonobi:
@dmitriyshashkin Thanks for finding this. We're working on a fix now and will put in the PR when its done.

PR #2497, opened 2018-05-04
Sonobi Adapter - Added referrer param. Fixed timeout error in userSync

JonGoSonobi:
## Type of change
- [x] Bugfix
- [x] Feature

## Description of change
Added a referrer param to override the default window.location.host.
Fix for issue https://github.com/prebid/Prebid.js/issues/2489

- apex.prebid@sonobi.com

https://github.com/prebid/prebid.github.io/pull/744

## Other information
Fixes Issue https://github.com/prebid/Prebid.js/issues/2489

### DIFF
--- a/PR_REVIEW.md
+++ b/PR_REVIEW.md
@@ -34,6 +34,7 @@ For modules and core platform updates, the initial reviewer should request an ad
 - Adapters may not use the $$PREBID_GLOBAL$$ variable
 - All adapters must support the creation of multiple concurrent instances. This means, for example, that adapters cannot rely on mutable global variables.
 - Adapters may not globally override or default the standard ad server targeting values: hb_adid, hb_bidder, hb_pb, hb_deal, or hb_size, hb_source, hb_format.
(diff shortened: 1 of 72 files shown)


JonGoSonobi:
@ndhimehta any movement on this?

dmitriyshashkin:
@JonGoSonobi  tbh, not the best way to fix the problem. Instead of checking if the array is empty you wrap it in a try/catch. Not the most appropriate way to check the length of the array. And then you log the error. The situation that is causing the error is not uncommon or unexpected, requests to server timeout all the time. And now I get an ugly error message each time sonobi times out.

JonGoSonobi:
@dmitriyshashkin The error should only be logged in debug mode. In which case it would let you know that the timeout for the Sonobi adapter is too low and that you should consider increasing it.

dmitriyshashkin:
@JonGoSonobi `TypeError: Cannot read property 'body' of undefined` - it's rather difficult to understand that I should consider raising the timeout from this message. Besides prebid already fires the following error `Server call for sonobi failed: . Continuing without bids.` I'd say it's much more informative

JonGoSonobi:
@dmitriyshashkin PR opened to remove the debug logging. https://github.com/prebid/Prebid.js/pull/2686
We left the try catch, otherwise we would have to check for not only the serverResponses Array being not empty, but also that the first index of it has a body and sbi_px is an array. We feel that just surrounding the parsing of server responses in a try catch is a much neater approach.

Issue #2386, opened 2018-04-12
Intent to Implement: Updates to UserSync API


jsnellbaker:
## Type of issue
Feature

## Description
In order to offer additional features to the UserSync functionality, we plan to implement some changes to the API.  This includes adding and removing existing fields which make break current setups.  Below is a pair of lists of the new and deprecated fields, with an explanation of the functional changes further below.

This will require some adapters to update their `getUserSyncs` code in case they were referring to deprecated fields.

### API changes
**New fields**
- `filterSettings: {config}`

**Deprecated fields**
- `iframeEnabled: bool`
- `pixelEnabled: bool`
- `enabledBidders: []`

All other API fields that exist today and not listed here would remain and retain their existing functionality.  

### Functional changes
Currently the only filtering field available is the `enabledBidders` field.  This field takes a list of bidders and acts as a white-list, allowing only those listed bidders to drop their iframe and image pixels.  Additionally the iframe pixels are only dropped if they've also included the `iframeEnabled` field in the config and set it to `true`.

The proposed changes will offer the following:
- the `filterSettings` field will accept an object config that allows the publisher to set a white-list or black-list filter against a list of bidders.
- distinct bidder filters can be set for the iframe or image pixels (so that you can have black-list on pixels and a white-list on iframes)
- implicitly listing the iframe object in this config automatically enables iframes pixels to drop (removing the need to add `iframeEnabled`)
- not listing the image/iframe fields will have them operate in their current default modes (ie image pixels are inherently permitted for all bidders, while iframe pixels are inherently blocked for all bidders).

The `filterSettings` config would look like the following:
```
filterSettings: {
    iframe: {
        bidders: ['a', 'b'],
        filter: 'black'
    },
    image: {
        bidders: ['*'],
        filter: 'white'
    }
}
```

The `bidders` field would either accept a list of strings of the bidders or a `'*'` string to indicate all bidders.
The `filter` field is an enum that accepts either `white` or `black` to indicate which type of filtering logic should be applied for the tracking type.

As mentioned earlier, not including one of the `iframe` or `image` fields means today's default logic  would be used.  In case of no `image` object, **all** bidders are permitted to drop their image pixels.  In case of no `iframe` object, **no** bidders are permitted to drop their iframe pixels.

If a publisher wanted to enable all bidders drop their iframe pixels, they'd need to configure the `filterSettings` field with an iframe object using the `'*'` and `white` filter option.

bretg:
[...]

PR #2499, opened 2018-05-04
updated logic for userSync - new field filterSettings

jsnellbaker:
## Type of change
- [x] Feature

## Description of change
This set of changes introduces a new field for the userSync API, called `filterSettings`.  This field allows the publisher to configure independent filtering rules against a defined list of bidders (or all bidders) with a specified white-list/black-list filter option.  These rules are configurable for both iframe and image pixels, so that you can have a white-list for bidders A, B, C for iframe pixels and a black list for bidder A for image pixels.

These changes are part of an intermediate step to deprecate several userSync API fields; specifically `iframeEnabled`, `pixelEnabled` and `enabledBidders`.  While these fields are not fully deprecated and can still be used after this PR, the new `filterSettings` field will trump those settings if the new field exists (and is properly configured).  As a contrived example, if `pixelEnabled` was set to `false`, but there was a `filterSettings.image` config object - the image pixels would be dropped as per the latter logic.

So long as the `filterSettings` `iframe` or `image` object config is defined and valid, it effectively the same value as `pixelEnabled` or `iframeEnabled` set to `true`.  If the config options are invalid, then we fallback to the default config logic where all bidders can drop `image` pixels and no bidders can drop `iframe` pixels.

Below is a copy of the config object with all its possible fields/options:
```
userSync: {
  filterSettings: {
    iframe: {
      bidders: ['*'],      // '*' represents all bidders
      filter: 'include'
    },
    image: {
      bidders: ['appnexus', 'testBidder'],
      filter: 'exclude'
    }
  }
}
```

- A link to a PR on the docs repo:
https://github.com/prebid/prebid.github.io/pull/747

## Other information
PR for https://github.com/prebid/Prebid.js/issues/2386


jsnellbaker:
@snapwich To clarify on your feedback, the new config could look like the following:

```
userSync: {
  filterSettings: {
    iframe: {
      bidders: '*',      // '*' represents all bidders
      filter: 'include'
    },
    image: {
      bidders: ['appnexus', 'testBidder'],
      filter: 'exclude'
    },
    all: {
      bidders: ['newBidder']
      // undefined filter = 'include'
    }
  }
}
```

If so - in the case someone sets up a config with the same bidder in both `all` and either of the other types, should we consider that an invalid config and fallback to the default logic?

mkendall07:
I don't like the `all` option because you can set ambiguous things then, such as this:
```
filterSettings : {
  all : { 
    {
      bidders: '*',
       filter: 'include'
    }
  },
iframe : { 
    {
      bidders: '*',
       filter: 'exclude'
    }
  }
}
```

Unless we disallow `all` with explicit settings...

snapwich:
@mkendall07 I agree there can be some ambiguous cases.  Perhaps they should throw errors?  I'm just thinking from the perspective of this being used, It's most likely to just enable/disable pixels for bidders as opposed to _types_ of pixels (although it's nice to have that fine control _if_ needed).

If someone wants to enable/disable all pixels for rubicon (I'm guessing will be a common use case), w/o an `all` they need to know all the pixels that rubicon uses and/or list all the pixels types (which could possibly grow over time) and include rubicon in each entry.

It's not a blocker for me, but it makes sense I think.  It's also something that could be added at a later point.

jsnellbaker:
@snapwich As you may have seen, I've updated the code in this PR with an overall refactor as well as some of your points (the `filter` default of `include` and changing the bidder wildcard to just the string `'*'` instead of placing it in an array).

I've talked further with @mkendall07 on the `all` config field, and I'm going to try to implement it as well.  

As I noted earlier, if the logic of the config is invalid - I am planning to throw a warning and revert to the old userSync logic (allow image pixels, block iframe).  This is the current behavior if other parts of the `filterSettings` config is invalid, so we'll be consistent here too.  If you have other thoughts on this part of the functionality, please let me know.

jsnellbaker:
I have just pushed in some changes to support the `all` config option.  As a note - it was very difficult to try to resolve the intended logic when the `all` config was mixed with the other two configs without adding more sets of rules for the end-user to follow.  

In lieu of adding complexity like that, I talked with @jaiminpanchal27 and we thought to implement the `all` config as being mutually exclusive with the `image/iframe` configs.  So now, a user can either use `filterSettings.all` or use `filterSettings.iframe` and/or `filterSettings.image`.  If they try to combine them, we'll return an error indicating an invalid config and will default to old logic.

Please let me know of any thoughts, comments, feedback on this new addition.  Thanks.

snapwich:
Does this need to be updated as part of this pull-request? https://github.com/prebid/Prebid.js/blob/24e6a1dd278437b9df77a75618a5ab5423e74ce1/src/userSync.js#L6

jsnellbaker:
@snapwich No - that config will still preserve the native functionality of the userSync feature if you don't declare the `filterSettings` object.  Eventually when we rework the feature and make breaking changes (ie removing some fields), then that would likely change.

jsnellbaker:
There are no breaking changes here as per @bretg's request on #2386.  The default behavior (where all bidders can drop image pixels and no bidders can drop iframe pixels) is still in place.

EskelCz:
@jsnellbaker Sorry to reopen this topic but I can't find the answer anywhere. Why is iframe sync disabled by default? It seems that most bidders are encouraging publishers to enable them, so why not enable them by default? Thanks.

bretg:
Of course bidders encourage publishers to enable -- it's a super-hypercharged open code block on the page.

Prebid is doing the right thing by making publishers think about who they trust with this privilege.

Issue #1185, opened 2017-05-09
How to add costum analytic adapter


chensMMG:
## Type of issue
Question

## Description
I'm having some issues adding custom analytic adapter.

I cloned the latest version to include the analytics.json, and have followed the API documentation under 'Integrate with the Prebid Analytics API' but with no luck.

In the project you have example (as shown under src/adapters/analytics) how to open custom adapter, I copied and changed the 'example.js' script and added the name to analytics.json.

There is no extra information or examples about how to work with the custom adapter, like explanations about the different 'analyticsType' variable.

After doing all of the above I did not succeed in making the adapter fire the analytics event.

## Steps to reproduce
First, I have tried to open analytic adapter like the first example js file - src/adapters/analytics/example.js.
The analytics type set to 'library', I opened new js file under src/analytics/libraries and copy the first example file and change all the 'example' params.

Second, I have tried to open analytic adapter like the second example js file - src/adapters/analytics/example2.js.
After some checking it seems to be adapter that works only with GA.

### Expected results
After running my test prebid script and runnig the snippets codes for checking all registered and winning bids (my bid is registered as a valid bid and winning bid) , the analytic adapter did not fire.

### Actual results
Using the first step, when the bid is over an ajax request is fired to the url I entered but with no data.

## Platform details
Node version: v4.4.3
Npm version: 2.15.1
Prebid.js version: v0.24.0-pre

### Example code

Analytic adapter:

```
import adapter from 'AnalyticsAdapter';

export default adapter(
  {
    url: 'URL',
    global: 'MarsMediaPrebidAnalytics',
    handler: 'on',
    analyticsType: 'library'
  }
);
```

Library:

```
window.MarsMediaPrebidAnalytics = function(hander, type, data) {
  console.log(`call to MarsMediaPrebidAnalytics Analytics library: example('${hander}', '${type}', ${JSON.stringify(data)})`);
};

window[window.MarsMediaPrebidAnalytics] = function() {};

//var utils = require('utils');
//var events = require('events');
//var pbjsHandlers = require('prebid-event-handlers');
var utils = { errorless: function(fn) { return fn; } };

var events = { init: function() { return arguments; } };

var pbjsHandlers = {
  onBidAdjustment: args => console.log('pbjsHandlers onBidAdjustment args:', args),
  onBidTimeout: args => console.log('pbjsHandlers bidTimeout args:', args),
  onBidRequested: args => console.log('pbjsHandlers bidRequested args:', args),
[...]

PR #2506, opened 2018-05-08
Fix path to analytics src directory in build

tomeightyeight:
## Type of change
- [x] Bugfix
- [x] Build related changes

## Description of change
The `analyticsDirectory` path defined in the gulpfile appears to be incorrect. I've updated this to reflect the directory structure found within the repo:

`src/adapters/analytics`

This was stopping me from running my build.

## Other information

See related historical issue here: https://github.com/prebid/Prebid.js/issues/1185

### DIFF
--- a/gulpfile.js
+++ b/gulpfile.js
@@ -33,7 +33,7 @@ var jsEscape = require('gulp-js-escape');
 var prebid = require('./package.json');
 var dateString = 'Updated : ' + (new Date()).toISOString().substring(0, 10);
 var banner = '/* <%= prebid.name %> v<%= prebid.version %>\n' + dateString + ' */\n';
-var analyticsDirectory = '../analytics';
+var analyticsDirectory = './src/adapters/analytics';
 var port = 9999;
 
 // Tasks


tomeightyeight:
We've included Prebid as a dependency in a repo of our own and then we've created different branches to support both the 0.* and 1.* major releases. We then have a build script of our own that does this.

```
#!/bin/bash

mkdir -p build/dist
yarn install
cp -f ./modules/vendorNameAnalyticsAdapter.js ./node_modules/prebid.js/modules/vendorNameAnalyticsAdapter.js
cp -f ./test/vendorNameBidAdapter_spec.js ./node_modules/prebid.js/test/spec/modules/vendorNameBidAdapter_spec.js
cp  -f ./modules/vendorNameBidAdapter.js ./node_modules/prebid.js/modules/vendorNameBidAdapter.js
cp -f modules.json ./node_modules/prebid.js/modules.json

(
    cd node_modules/prebid.js/
    ../.bin/yarn install
    ../.bin/gulp build --modules=modules.json
)

cp ./node_modules/prebid.js/build/dist/vendorName*.js ./build/dist/
cp ./node_modules/prebid.js/build/dist/prebid.js ./build/dist/
```

Looks like you are suggesting this is wrong? Maybe we should be forking the Prebid repo instead - but the path would still be wrong (as the gulpfile wouldn't be adjacent to our analytics directory)?

Can you point me in the right direction? Thanks

mkendall07:
@jaiminpanchal27 please check this.

jaiminpanchal27:
Hey @tomeightyeight 
Actually when this was designed, the idea was to support analytics repository in parallel to Prebid.js project.
So if someone is having analytics adapter in a separate node module or a repository, they can build by checking out the repository in parallel to Prebid.js project or installing the node module as a dependency in some other project.

This needs to be redesigned and should be documented somewhere in prebid.org. For now I have created issue https://github.com/prebid/prebid.github.io/issues/823 to update docs on existing pattern.

jaiminpanchal27:
Closing this since this needs to be redesigned.

PR #2507, opened 2018-05-08
Aardvark v1.0

dejanstrbac:
## Type of change

- [ ] Bugfix
- [ ] Feature
- [ ] New bidder adapter   
- [ ] Build related changes
- [ ] CI related changes
- [ ] Other

## Description of change

Support for Prebid.js > v1.0

```
{
    bidder: 'aardvark',
    params: {
       ai: '0000',
       sc: '1234'
    }
}
```

christ@rtk.io @naegelin is the official maintainer and responsible (CTO of RTK.io) 

### DIFF
--- a/modules/aardvarkBidAdapter.js
+++ b/modules/aardvarkBidAdapter.js
@@ -94,14 +94,15 @@ export const spec = {
 
   interpretResponse: function(serverResponse, bidRequest) {
     var bidResponses = [];
-    utils._each(serverResponse.body, function(rawBid) {
-      if (!rawBid.cpm || !(+rawBid.cpm)) {
-        return;
-      }
 
+    if (!Array.isArray(serverResponse.body)) {
+      serverResponse.body = [serverResponse.body];
+    }
+
+    utils._each(serverResponse.body, function(rawBid) {
       var bidResponse = {
         requestId: rawBid.cid,
-        cpm: rawBid.cpm,
+        cpm: rawBid.cpm || 0,
         width: rawBid.width || 0,
         height: rawBid.height || 0,
         currency: rawBid.currency ? rawBid.currency : AARDVARK_CURRENCY,
--- a/test/spec/modules/aardvarkBidAdapter_spec.js
+++ b/test/spec/modules/aardvarkBidAdapter_spec.js
@@ -213,7 +213,7 @@ describe('aardvarkAdapterTest', () => {
           {
             media: 'banner',
             nurl: 'http://www.nurl.com/1',
-            cpm: 0.09,
+            cpm: 0.19,
             width: 300,
             height: 250,
             cid: '1abgs362e0x48a8',
@@ -237,7 +237,7 @@ describe('aardvarkAdapterTest', () => {
       expect(result[0].ad).to.not.be.undefined;
 
       expect(result[1].requestId).to.equal('1abgs362e0x48a8');
-      expect(result[1].cpm).to.equal(0.09);
+      expect(result[1].cpm).to.equal(0.19);
       expect(result[1].width).to.equal(300);
       expect(result[1].height).to.equal(250);
       expect(result[1].currency).to.equal('USD');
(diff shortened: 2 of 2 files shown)


modules/aardvarkBidAdapter.js
```diff
@@ -0,0 +1,154 @@
...
+
+        if (b.params.categories && b.params.categories.length) {
+          rMap.payload.categories = rMap.payload.categories || []
+          utils._each(b.params.categories, function(cat) {
+            rMap.payload.categories.push(cat);
+          });
+        }
+
+        if (bidderRequest && bidderRequest.gdprConsent) {
+          rMap.payload.gdpr = bidderRequest.gdprConsent.gdprApplies;
```
jsnellbaker: There may be rare chances where this value may be `undefined` if the CMP didn't have it defined.  Do you want to set your own back-up/default value in that case?  If so, I would recommend to check if `typeof bidderRequest.gdprConsent.gdprApplies === 'boolean'` so that you can accept natural cases where `gdprApplies` was set to `false.

modules/aardvarkBidAdapter.js
```diff
@@ -0,0 +1,154 @@
...
+    });
+
+    return bidResponses;
+  },
+
+  getUserSyncs: function(syncOptions, serverResponses, gdprConsent) {
+    const syncs = [];
+    var url = '//' + SYNC_ENDPOINT + '/cs';
+    var gdprApplies = true;
+    if (typeof gdprConsent.gdprApplies === 'boolean') {
```
jsnellbaker: please add to the check if `gdprConsent` exists.  On pages without a CMP, this current code would lead to an undefined error.

modules/aardvarkBidAdapter.js
```diff
@@ -0,0 +1,154 @@
...
+
+      var bidResponse = {
+        requestId: rawBid.cid,
+        cpm: rawBid.cpm,
+        width: rawBid.width || 0,
+        height: rawBid.height || 0,
+        currency: rawBid.currency ? rawBid.currency : AARDVARK_CURRENCY,
+        netRevenue: rawBid.netRevenue ? rawBid.netRevenue : true,
+        ttl: rawBid.ttl ? rawBid.ttl : AARDVARK_TTL,
+        exchange: rawBid.ex,
```
jsnellbaker: Can remove this `exchange` field from the `bidResponse` object.  This is not a current field.

modules/aardvarkBidAdapter.js
```diff
@@ -0,0 +1,154 @@
+import * as utils from 'src/utils';
+import {registerBidder} from 'src/adapters/bidderFactory';
+
+const BIDDER_CODE = 'aardvark';
+const DEFAULT_ENDPOINT = 'bidder.rtk.io';
+const SYNC_ENDPOINT = 'sync.rtk.io';
+const AARDVARK_TTL = 300;
+const AARDVARK_CURRENCY = 'USD';
+
+var hasSynced = false;
```
jsnellbaker: could change this to `let` instead of `var`.

modules/aardvarkBidAdapter.js
```diff
@@ -0,0 +1,154 @@
...
+  },
+
+  buildRequests: function(validBidRequests, bidderRequest) {
+    var auctionCodes = [];
+    var requests = [];
+    var requestsMap = {};
+    var referer = utils.getTopWindowUrl();
+    var pageCategories = [];
+
+    if (window.rtkcategories && Array.isArray(window.rtkcategories)) {
```
jsnellbaker: Can you clarify the use-case here for reading this window object?

naegelin: This is a requirement from several tier 1 publishers in order to allow RTK buyers to target specific inventory. It is currently implemented across Hearst and others

modules/aardvarkBidAdapter.js
```diff
@@ -0,0 +1,158 @@
...
+        bidderRequest
+      });
+    });
+
+    return requests;
+  },
+
+  interpretResponse: function(serverResponse, bidRequest) {
+    var bidResponses = [];
+    utils._each(serverResponse.body, function(rawBid) {
```
jsnellbaker: Can you review the logic in using this `utils._each` function?  It's currently iterating through each of the properties of the response object and evaluating them each as an entire bid object.  

This is a copy of the bid response object that I saw when I tested and it treated `rawBid` as `demo_exchange` (from  the `ex` field) during the first pass of the function.
```
{"ex": "demo_exchange", "width": 300, "cpm": 0.01, "cid": "2edd6ea6821b56", "media": "banner", "adm": "<div>Demo Unit</div>", "id": "1234", "height": 250}
```

dejanstrbac: We've been updating the demo auction today, sorry about that. That is an array returned from our endpoint, so it gets iterated over as an array, not a map. It is now an array as it should be.

PR #2509, opened 2018-05-08
Bug fix - mantisBidAdapter.js

parisholley:
wasn't reading the right value out of the response


parisholley:
@dmitriyshashkin will fix the issue, also pushing a backend change that can work around

dmitriyshashkin:
@harpere sorry to bother you but we'd appreciate if you could review this before the next release. We're missing all revenue we used to get from this bidder

Issue #2339, opened 2018-04-03
InvibesBidAdapter unit test failing in Edge 14 browser


jsnellbaker:
## Type of issue
Bug/Issue

## Description
While running our browserstack tests for the prebid release process, we encountered this consistent error in the `InvibesBidAdapter` file for the Edge 14 browser.  

It seems like there is a cookie that shouldn't be present for this unit test.  @rcheptanariu Could you take a look and see if there are any changes that could be made to the unit test? 
 
```
invibesBidAdapter:
    buildRequests
      ✗ uses cookies
	expected 'p4rrk7.ax2i2s' to be undefined
	Error
	   at AssertionError (webpack:///node_modules/assertion-error/index.js:74:0 <- test/test_index.js:9276:7)
	   at Assertion.prototype.assert (webpack:///node_modules/chai/lib/chai/assertion.js:107:0 <- test/test_index.js:14898:7)
	   at Anonymous function (webpack:///node_modules/chai/lib/chai/core/assertions.js:343:0 <- test/test_index.js:15271:5)
	   at addProperty (webpack:///node_modules/chai/lib/chai/utils/addProperty.js:43:0 <- test/test_index.js:14433:9)
	   at Anonymous function (webpack:///test/spec/modules/invibesBidAdapter_spec.js:120:6 <- test/test_index.js:40491:8)
```

jaiminpanchal27:
This bug is fixed, have not seen this popup recently in our test logs

PR #2512, opened 2018-05-09
Invibes Bid Adapter - cookies update & user sync

rcheptanariu:
## Type of change
- [x] Other

## Description of change
Send & update cookies through user sync

- contact email of the adapter’s maintainer: system_operations@invibes.com

## Other information
This update should also fix the issue in the unit tests: https://github.com/prebid/Prebid.js/issues/2339

### DIFF
--- a/modules/invibesBidAdapter.js
+++ b/modules/invibesBidAdapter.js
@@ -34,17 +34,25 @@ export const spec = {
   interpretResponse: function (responseObj, requestParams) {
     return handleResponse(responseObj, requestParams != null ? requestParams.bidRequests : null);
   },
-  getUserSyncs: function(syncOptions) {
+  getUserSyncs: function (syncOptions) {
     if (syncOptions.iframeEnabled) {
+      handlePostMessage();
+      const syncUrl = buildSyncUrl();
+
       return {
         type: 'iframe',
-        url: CONSTANTS.SYNC_ENDPOINT
+        url: syncUrl
       };
     }
   }
 };
 
 registerBidder(spec);
+
+// some state info is required: cookie info, unique user visit id
+const topWin = getTopMostWindow();
+let invibes = topWin.invibes = topWin.invibes || {};
+let _customUserSync;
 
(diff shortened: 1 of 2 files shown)


modules/invibesBidAdapter.js
```diff
@@ -34,18 +34,26 @@ export const spec = {
...
       };
     }
   }
 };
 
 registerBidder(spec);
 
+// some state info is required: cookie info, unique user visit id
+const topWin = getTopMostWindow();
+let invibes = topWin.invibes = topWin.invibes || {};
```
mike-chowla: What's the reason for creating the invibes object on the top window and attaching some of the functions below to it?

modules/invibesBidAdapter.js
```diff
@@ -300,35 +310,69 @@ function getCappedCampaignsAsString() {
...
+
+function handlePostMessage() {
+  try {
+    if (window.addEventListener) {
+      window.addEventListener('message', acceptPostMessage);
+    }
+  } catch (e) { }
+}
+
+function acceptPostMessage(e) {
```
mike-chowla: What's the reason the this message?  It's not clear to me why the caller does take care of setting cookies, etc and why the code to do so is inside your Prebid adapter.

modules/invibesBidAdapter.js
rcheptanariu: The location is on the top window so that we don't initialize our local id on this page twice. The cookie info and the id are passed to the user sync function. Also the functions for cookies are reused inside the postMessage, so it made sense to just put them all inside the same object

modules/invibesBidAdapter.js
rcheptanariu: The user sync happens in the iframe, on a different domain that the initial page. We need to update our local id if the user sync was successful, so the only way to do this was with a postMessage from the iframe.

rcheptanariu:
Hello, please let me know if there are any more questions :)

Issue #2516, opened 2018-05-10
GDPR-compliance for Prebid Server cookiesyncs


dbemiller:
Prebid.js cookie syncs with Prebid Server violate GDPR. See https://github.com/prebid/prebid-server/issues/501 for more details.

If the GDPR module is included, please include the following properties in the `POST /cookie_sync` request body to Prebid Server:

```
{
  "gdpr": 1,
  "gdpr_consent": "vendor-consent-string"
}
```

If GDPR is in effect:
- `gdpr` should be 1
- `gdpr_consent` must be the [Unpadded base64-URL](https://tools.ietf.org/html/rfc4648#section-5) 
 encoded [vendor consent string](https://github.com/InteractiveAdvertisingBureau/GDPR-Transparency-and-Consent-Framework/blob/master/Consent%20string%20and%20vendor%20list%20formats%20v1.1%20Final.md#vendor-consent-string-format-). This should be the same as you use in `request.user.ext.consent` for `/openrtb2/auction` payloads.

If GDPR is not in effect:
- `gdpr` should be 0
- `gdpr_consent` should be undefined

If the GDPR status is uncertain:
- `gdpr` should be undefined
- `gdpr_consent` is optional, but you should send it if you have it.

If the GDPR module is _not_ included, then the status is uncertain. `gdpr` should be undefined, and `gdpr_consent` is optional.

Let me know if you have any questions or concerns.

Thanks!

bretg:
While we're address this, can we cover https://github.com/prebid/Prebid.js/issues/2242 ?  I think this is the same piece of code. We haven't been able to get to it.

jsnellbaker:
@dbemiller if the module is disabled, does that fall into the 'GDPR is not in effect' bucket or should we just not add the `gdpr` and `gdpr_consent` fields at all to the POST request?

jsnellbaker:
@bretg I will aim to include that issue you noted as part of my changes here.

dbemiller:
@jsnellbaker in that case, don't add the `gdpr` or `gdpr_consent` fields at all to the POST request.

dbemiller:
@jsnellbaker safe to close this?

jsnellbaker:
Yes - this change was merged in as part of the 1.11.0 release.

PR #2530, opened 2018-05-14
GDPR - add consent information to PBS cookie_sync request

jsnellbaker:
## Type of change
- [x] Feature

## Description of change
Addresses the issue raised in https://github.com/prebid/Prebid.js/issues/2516

Adds support to pass the consent information into the prebidServer cookie sync request.
We also removed an additional sync attempt made in the response as this was deemed redundant.

### DIFF
--- a/modules/prebidServerBidAdapter.js
+++ b/modules/prebidServerBidAdapter.js
@@ -115,7 +115,7 @@ function queueSync(bidderCodes, gdprConsent) {
   }
   _synced = true;
 
-  let payload = {
+  const payload = {
     uuid: utils.generateUUID(),
     bidders: bidderCodes
   };
@@ -684,7 +684,8 @@ export function PrebidServer() {
       .filter(utils.uniques);
 
     if (_s2sConfig && _s2sConfig.syncEndpoint) {
-      queueSync(_s2sConfig.bidders, bidRequests[0].gdprConsent);
+      let consent = (Array.isArray(bidRequests) && bidRequests.length > 0) ? bidRequests[0].gdprConsent : undefined;
+      queueSync(_s2sConfig.bidders, consent);
     }
 
     const request = protocolAdapter().buildRequest(s2sBidRequest, bidRequests, adUnitsWithSizes);
--- a/modules/prebidServerBidAdapter.js
+++ b/modules/prebidServerBidAdapter.js
@@ -96,36 +96,56 @@ function setS2sConfig(options) {
   }
 
   _s2sConfig = options;
-  if (options.syncEndpoint) {
-    queueSync(options.bidders);
-  }
 }
 getConfig('s2sConfig', ({s2sConfig}) => setS2sConfig(s2sConfig));
 
 /**
+ * resets the _synced variable back to false, primiarily used for testing purposes
+*/
+export function resetSyncedStatus() {
+  _synced = false;
+}
+
+/**
  * @param  {Array} bidderCodes list of bidders to request user syncs for.
  */
-function queueSync(bidderCodes) {
+function queueSync(bidderCodes, gdprConsent) {
   if (_synced) {
     return;
   }
   _synced = true;
-  const payload = JSON.stringify({
+
+  let payload = {
     uuid: utils.generateUUID(),
     bidders: bidderCodes
-  });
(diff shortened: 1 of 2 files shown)


modules/prebidServerBidAdapter.js
```diff
@@ -96,36 +96,56 @@ function setS2sConfig(options) {
...
  */
-function queueSync(bidderCodes) {
+function queueSync(bidderCodes, gdprConsent) {
   if (_synced) {
     return;
   }
   _synced = true;
-  const payload = JSON.stringify({
+
+  let payload = {
```
dbemiller: why `const` => `let` here? I see mutation, but no re-assignments... seems like `const` should still work?

modules/prebidServerBidAdapter.js
```diff
@@ -666,6 +683,10 @@ export function PrebidServer() {
       .reduce(utils.flatten)
       .filter(utils.uniques);
 
+    if (_s2sConfig && _s2sConfig.syncEndpoint) {
+      queueSync(_s2sConfig.bidders, bidRequests[0].gdprConsent);
```
dbemiller: Is `bidRequests` guaranteed to have at least one element in it? If not, this could break.

test/spec/modules/prebidServerBidAdapter_spec.js
```diff
@@ -424,6 +424,58 @@ describe('S2S Adapter', () => {
...
+      };
+
+      adapter.callBids(REQUEST, gdprBidRequest, addBidResponse, done, ajax);
+      let requestBid = JSON.parse(requests[0].requestBody);
+
+      expect(requestBid.gdpr).is.equal(1);
+      expect(requestBid.gdpr_consent).is.equal('abc123def');
+
+      // check scenario if gdprApplies is false
+      resetSyncedStatus();
```
dbemiller: Tests shouldn't have side-effects on the global state.

It's also dangerous to put this mid-test because it won't execute if one of the earlier assertions fail.

This test should be split into smaller ones, and `resetSyncedStatus()` called in an [afterEach()](https://mochajs.org/#hooks) block

modules/prebidServerBidAdapter.js
jsnellbaker: I'll change it to const to be consistent.

test/spec/modules/prebidServerBidAdapter_spec.js
jsnellbaker: True on the above; I wanted to limit the scenarios where I was using this function to not affect other tests by having that function run for every test.  But it doesn't seem like it's breaking the other tests when I put it there now.  I'll make the changes here.

modules/prebidServerBidAdapter.js
jsnellbaker: Per this check, it should always at least have 1 element:
https://github.com/prebid/Prebid.js/blob/master/src/adaptermanager.js#L289

But I can add another check in this code to be sure.

dbemiller: Either way.

My only two concerns are "does it work now?" and "is it likely that the code changes to break it in the future without anyone realizing?" Looks like the answer to (1) is "yes." (2) usually depends on how thorough the docs and unit tests are.

Your call in the end.

modules/prebidServerBidAdapter.js
```diff
@@ -96,36 +96,56 @@ function setS2sConfig(options) {
...
+*/
+export function resetSyncedStatus() {
+  _synced = false;
+}
+
 /**
  * @param  {Array} bidderCodes list of bidders to request user syncs for.
  */
-function queueSync(bidderCodes) {
+function queueSync(bidderCodes, gdprConsent) {
```
mkendall07: Would be awesome if this code could be reused for `spec.getUserSyncs` but it's ok for now.

test/spec/modules/prebidServerBidAdapter_spec.js
dbemiller: it's fine... if other tests fail because the global state gets reset, then _they're_ the broken ones. Tests should be initializing their own state, rather than relying on other tests to do it for them.

PR #2536, opened 2018-05-15
Update to XHB adapter

daniel-hoffmann:
Update to XHB adapter for GDPR and v1


jsnellbaker:
@daniel-hoffmann Can you take a look at the Travis errors and resolve them when you get the chance?  I'll start to look-over the adapter changes in the meantime.

Also, would you be able to open a docs PR in the [docs repo](https://github.com/prebid/prebid.github.io) and make an update in your bidders page to add the following variable?
```
gdpr_supported: true
```

This can go directly below the variable for your 1.0 compliance.  This will allow your adapter to appear in a table that shows GDPR compliant adapters.

Thanks.

daniel-hoffmann:
@jsnellbaker Both things have been done - let me know if you need anything else to be changed!

modules/xhbBidAdapter.js
```diff
@@ -0,0 +1,457 @@
...
+ * Unpack the Server's Bid into a Prebid-compatible one.
+ * @param serverBid
+ * @param rtbBid
+ * @param bidderRequest
+ * @return Bid
+ */
+function newBid(serverBid, rtbBid, bidderRequest) {
+  const bid = {
+    requestId: serverBid.uuid,
+    cpm: 0.00,
```
jsnellbaker: To confirm should be 0 all the time?  Shouldn't it pull in the CPM from the returned bid?

modules/xhbBidAdapter.js
```diff
@@ -0,0 +1,457 @@
...
+    creativeId: rtbBid.creative_id,
+    dealId: 99999999,
```
jsnellbaker: The `dealId` field is an optional field and doesn't have to be explicitly set.  If you're not planning on using it, it could be removed instead of hard-coding to a set value.  If you are using that set value for a particular use-case, please let me know.

modules/xhbBidAdapter.md
```diff
@@ -0,0 +1,103 @@
...
+        native: {
+          title: {
+            required: true,
+            len: 80
+          },
+          body: {
+            required: true
+          },
+          brand: {
+            required: true
```
jsnellbaker: In the test placement - the `brand` is not included which is preventing the bid from being accepted by the native param checks.  If you want the `brand` to be a required field in general for your publishers, could you update the test ad unit to include this field?  If the field isn't strictly required, could you update this .md file?

modules/xhbBidAdapter.js
mkendall07: This is what Xaxis wants to do since they have pre-arranged deals.

mkendall07:
@daniel-hoffmann
looks like you need to add a test spec file. You can see example here: https://github.com/prebid/Prebid.js/blob/master/test/spec/modules/appnexusBidAdapter_spec.js

modules/xhbBidAdapter.js
daniel-hoffmann: This is necessary for our setup.

modules/xhbBidAdapter.md
daniel-hoffmann: Actually, I just copied this from the AN adapter, I assume that one has the issue?

daniel-hoffmann: Yeah, we don't need that to be required.
Can I just set it to false in the md file?

jsnellbaker: That's fine, or you could just remove the `brand` object entirely.

modules/xhbBidAdapter.js
daniel-hoffmann: As per Matt's comment, yes, that is exactly what we need and together with the hardcoded dealId kinda the whole reason we have our own adapter. ;)

Issue #2551, opened 2018-05-16
ConsentManager fail with allowAuctionWithoutConsent: false prevents bidsBackhandlers to execute


elidiota:
## Type of issue
bug
## Description
Hi,
the consentmanager stops completely the execution of the bidsBackHandler property in the requestBids method when allowAuctionWithoutConsent is  set to false

## Steps to reproduce

Just setting 
consentManagement.allowAuctionWithoutConsent= false;
without ant CMP in the page triggers this.

### Expected results
Execution of the backhandler set in requestObj.bidsBackHandler  at the moment of the fail or when the timeout triggers

### Actual results
Nothing happens
## Platform details
Prebid 1.9.0 & prebid 1.10.0

jsnellbaker:
Hi @elidiota Thanks for reporting this.  I'll start to put together a fix.

PR #2555, opened 2018-05-16
execute bidsBackHandler when auction is canceled by consent module

jsnellbaker:
## Type of change
- [x] Bugfix

## Description of change
Fixes issue #2551 

Currently when the prebid auction is canceled by the consentManagement module, the `bidsBackHandler` included in the `requestBids` function is not executed.

This fix will now execute the `bidsBackHandler` function when cancellation happens.

### DIFF
--- a/modules/consentManagement.js
+++ b/modules/consentManagement.js
@@ -22,6 +22,7 @@ let consentData;
 let context;
 let args;
 let nextFn;
+let bidsBackHandler;
 
 let timer;
 let haveExited;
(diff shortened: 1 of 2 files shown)


modules/consentManagement.js
```diff
@@ -163,6 +164,7 @@ export function requestBidsHook(config, fn) {
   nextFn = fn;
   haveExited = false;
   let adUnits = config.adUnits || $$PREBID_GLOBAL$$.adUnits;
+  bidsBackHandler = config.bidsBackHandler;
```
mkendall07: is this safe? What about concurrent auctions with distinct bidBackHandler functions?

mkendall07: Oh i see this is fine. I assumed it was the global config `import { config } from 'src/config';`

You probably should rename this param so it's not confused with the global.

jsnellbaker: Thanks for the feedback, I renamed the variable.

PR #2562, opened 2018-05-17
Allow user to override which gpt slots should be targeted at invocation

ptomasroos:
## Type of change
- [x] Feature

## Description of change
We use the following flow:

We start a auction as early as possible. Without knowing where in the feed we're gonna place it. So depending on the outcome of the auction and how long it took we might inject it into a placement defined further down the page in order to now burn views above the fold.

The current semantics of setTargeting requires that we shall know a head of time what gptSlot the targeting should apply for because its looking into the auction object.

But allowing us to pass in an custom slot matching filter we can allow the targeting to explicitly use this slot when setting targeting

We use this in order not to place ads above fold and insert them to a better position instead of everything just being very static.

## Other information

Ref #1876 the old attempt, this is way cleaner and just follows along the new 1.x branch

cc: @mkendall07

### DIFF
--- a/src/prebid.js
+++ b/src/prebid.js
@@ -156,7 +156,7 @@ $$PREBID_GLOBAL$$.getBidResponsesForAdUnitCode = function (adUnitCode) {
 /**
  * Set query string targeting on one or more GPT ad units.
  * @param {(string|string[])} adUnit a single `adUnit.code` or multiple.
- * @param {function(object)} customSlotMatching gets a GoogleTag slot and returns a filter function.
+ * @param {function(object)} customSlotMatching gets a GoogleTag slot and returns a filter function for adUnitCode, so you can decide to match on either eg. return slot => { return adUnitCode => { return slot.getSlotElementId() === 'myFavoriteDivId'; } };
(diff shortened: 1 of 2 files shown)


ptomasroos:
Wopps. Not ment to close this as its very valid IMHO and backwards compatible

src/prebid.js
```diff
@@ -156,9 +156,10 @@ $$PREBID_GLOBAL$$.getBidResponsesForAdUnitCode = function (adUnitCode) {
 /**
  * Set query string targeting on one or more GPT ad units.
  * @param {(string|string[])} adUnit a single `adUnit.code` or multiple.
+ * @param {function(object)} customSlotMatching gets a GoogleTag slot and returns a filter function.
```
mkendall07: Can you update the description and add a usage example here?
ie, this is a a filter function that if passed will override the default matching for a slot. 
``` customSlotMatching(slot) = > slot.getAdUnitPath() === 'someIdPassed' ```

ptomasroos:
Thanks for the pointer!

ptomasroos:
@mkendall07 is this ready for merging or did I miss something?

PR #2588, opened 2018-05-22
Consent Management module bug fix.

anand-venkatraman:
## Type of change

- [x] Bugfix

## Description of change
Consent Management module has a bug when the CMP module responds back with `String` message instead of `Object`. The code fragment that checks for existence of substring `cmpReturn`, makes the check using `array.includes`, instead of `string.includes` from the `core-js` lib.

Also the test, was not really testing the code part for iframed pages. Since the existing test created a stub for `window.__cmp`, the test was not really running the code branch for "iframed" - `callCmpWhileInIframe`. Updated the test to verify behavior with both `Object` message format and `String` message format.


jsnellbaker:
@anand-venkatraman As you may have seen from the additional updates on this PR, we had to partially revert the changes to fix the build on master.  The unit test updates were causing some test failures on Safari browsers (see here https://travis-ci.org/prebid/Prebid.js/builds/382804970).

If you want to re-include the unit test enhancements, can you please take a look into these errors and create another PR with the updated changes?

anand-venkatraman:
@jsnellbaker didnt notice it before, but will check this and create another PR for you. Thanks.

PR #2592, opened 2018-05-22
Consistent targeting set

ptomasroos:
## Type of change
- [x] Bugfix
- [x] Feature

## Description of change
The idea is that users of SET_TARGETING who receives the argument targeting set, should be able to get the ad unit codes always involved even though a bid not might be available when rendering the ad

## Other information
Previously I tried to introduce the adUnits into the arguments https://github.com/prebid/Prebid.js/pull/1875
and since then the targetingSet has been starting to get passed. In order to make this without breaking any contracts my suggestion is that we always return a entry even though there is no bid available to the receiver of SET_TARGETING can iterate which ad unit codes are getting invoked.

### DIFF
--- a/src/targeting.js
+++ b/src/targeting.js
@@ -78,25 +78,6 @@ export function newTargeting(auctionManager) {
       .concat(getCustomBidTargeting(adUnitCodes, bidsReceived))
       .concat(config.getConfig('enableSendAllBids') ? getBidLandscapeTargeting(adUnitCodes, bidsReceived) : []);
 
-    // make sure at least there is a entry per adUnitCode in the targeting so receivers of SET_TARGETING call's can know what ad units are being invoked
-
-    adUnitCodes.forEach(key => {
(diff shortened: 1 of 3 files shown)


src/targeting.js
```diff
@@ -78,6 +78,25 @@ export function newTargeting(auctionManager) {
...
+          }
+        });
+        return hasMatchingAdUnitCode;
+      })) {
+        const emptyTargeting = {};
+        emptyTargeting[key] = [];
+        targeting.push(emptyTargeting);
+      }
+    });
+
```
snapwich: I think this could be greatly simplified by replacing it with the following code _after_ the `flattenTargeting` is called below

```
    adUnitCodes.forEach(code => {
      if (!targeting[code]) {
        targeting[code] = {};
      }
    });
```

ptomasroos: Good point! Will give it a try

Issue #2591, opened 2018-05-22
consent management module: incorrect field for consentString


jbartek25:
## Type of issue

bug

## Description

The current implementation of the consent management module expects "metadata" field in the CMP API getVendorConsents' return object to contain the consent string which is incorrect as per [CMP API spec](https://github.com/InteractiveAdvertisingBureau/GDPR-Transparency-and-Consent-Framework/blob/master/CMP%20JS%20API%20v1.1%20Final.md#vendorconsents) the "metadata" field contains just what it says - metadata like timestamps, version of CMP, version of vendor list, etc. The correct field containing the consent string should be "consentData" in the return object of the getConsentData command.

jsnellbaker:
Based on some internal discussion, the spec isn't explicitly clear about what parts of the consent information should be included in the consentData field from a getConsentData request vs what's listed as part of the metadata field from the getVendorConsents.  

That said, to err on the side of the caution and adherence to the spec - we're putting together a change that will use the `getConsentData` request to obtain the 'consentString' value.  We're still going to include the unparsed vendorConsent information for any bidders who wish to use this information directly in their adapter code.

jsnellbaker:
This has been fixed.

PR #2603, opened 2018-05-23
GDPR use getConsentData CMP call to get consentString

jsnellbaker:
## Type of change
- [x] Bugfix

## Description of change
Previously the consentManagement module was using the CMP function `getVendorConsents` to obtain all aspects of the user's consent information.  It was pointed out in #2591 that this function is not the ideal one to use to obtain the user's consent information (specifically the encoded string).

This PR updates the logic of the module to use the `getConsentData` CMP function to grab the `consentString` and `gdprApplies` and then executes the `getVendorConsents` CMP function to grab the unparsed vendor consent information (to maintain backwards compatibility for that variable we offer to adapters).

### DIFF
--- a/modules/consentManagement.js
+++ b/modules/consentManagement.js
@@ -33,7 +33,7 @@ const cmpCallMap = {
 };
 
 /**
- * This function handles interacting with an IAB compliant CMP to obtain the consentObject value of the user.
+ * This function handles interacting with an IAB compliant CMP to obtain the consent information of the user.
  * Given the async nature of the CMP's API, we pass in acting success/error callback functions to exit this function
  * based on the appropriate result.
  * @param {function(string)} cmpSuccess acts as a success callback when CMP returns a value; pass along consentObject (string) from CMP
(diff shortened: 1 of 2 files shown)


modules/consentManagement.js
```diff
@@ -42,27 +42,43 @@ const cmpCallMap = {
...
   let cmpCallbacks;
+  let cmpResponse = {};
+
+  // to collect the consent information from the user, we perform two calls to the CMP:
+  // first to collect the user's consent choices represented in an encoded string (via getConsentData)
+  // second to collect the user's full unparsed consent information (via getVendorConsents)
+  callIabCMP('getConsentData', function (consentResponse) {
+    cmpResponse.getConsentData = consentResponse;
+
+    callIabCMP('getVendorConsents', function (consentResponse) {
```
mkendall07: why are you doing this serially and not in parallel ?

dbemiller: I can chat about this with you if you're interested.

snapwich: yeah I think parallel would be ideal, especially if these are making ajax requests.  If they're not, it probably doesn't matter.

mkendall07: I don't think these are making ajax requests - so it probably doesn't matter so much.

jsnellbaker:
I have updated the logic to have the CMP calls execute in a parallel instead of serial as well as resolve the current conflicts with master.  Please take another look and let me know of any additional feedback/suggestions.

Thanks.

jsnellbaker:
Thanks for the approvals - merging in now.

Issue #2586, opened 2018-05-21
ConsentManagement module problem with multiple calls to requestBids


rale:
## Type of issue
Bug

## Description
If more than one call to requestBids is made while the GDPR ConsentManagement module is enabled, all but the last call are dropped.

## Steps to reproduce
- Enable consentManagement module
- Create two separate ad units, calling requestBids separately for each one, before consent is loaded

### Expected results
Both auctions should run.

### Actual results
Only the last auction is run.

## Platform details
Prebid 1.11.0

## Other information
I took a look at consentManagement.js, and it appears as though requestBidsHook is overwriting the context each time requestBids is called.

mkendall07:
looks like you are correct. Thanks for the bug report. @jsnellbaker please address. Thanks

jsnellbaker:
@rale Thanks for reporting this bug.  I have put together a fix that addresses this problem (#2612).  We'll try to get it included soon.

mkendall07:
this is resolved.

PR #2612, opened 2018-05-24
fix issue with GDPR module in concurrent auctions

jsnellbaker:
## Type of change

- [x] Bugfix

## Description of change
To address issue reported in https://github.com/prebid/Prebid.js/issues/2586

The current consentManagement module interferes with requests in concurrent auctions (ie multiple requestBids on a page) from firing off/finishing as expected.  This is due to the global variables being overwritten and the context of the previous auction instances being lost. 

This change stores the variables needed for the consentManagement module in an config object that gets passed along through the workflow of the module to remove the need for global variables.  This preserves the needed context for the multiple auctions to complete as expected.

To easily replicate the issue:
- setup a basic test page that uses a IAB compliant CMP and two adunit objects 
- pass each of these objects into a distinct requestBids call (so there's 1 adunit for each)
- add the consentManagement config to enable the module
- delete your `euconsent` cookie (in case you had consent previously)
- during next page load when you get prompted for consent, wait a few seconds then give consent

Both ads on the page should render successfully (with consent information included in both).  Previously only one of ads would render and nothing would be shown for the other.

### DIFF
--- a/modules/consentManagement.js
+++ b/modules/consentManagement.js
@@ -19,14 +19,6 @@ export let allowAuction;
 
 let consentData;
 
-let context;
-let args;
-let nextFn;
-let bidsBackHandler;
-
-let timer;
-let haveExited;
-
 // add new CMPs here, with their dedicated lookup function
 const cmpCallMap = {
   'iab': lookupIabConsent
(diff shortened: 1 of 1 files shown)


YOzaz:
By any chance, will this be merged soon? We have lots of lazy loading ads...

jsnellbaker:
@YOzaz Now that it's merged, it should be included in the next release (which is scheduled for later today).

YOzaz:
@jsnellbaker thank you thank you thank you. We're running out of time, so will probably grab `1.12.0-pre` for today as it is, and will update to official release tomorrow.

PR #2615, opened 2018-05-24
bugfix for targeting not setting for all adslots

jaiminpanchal27:
## Type of change
- [x] Bugfix

## Description of change
Targeting was not getting set for all adSlots. We were setting `targetingSet` status in the wrong function.

### DIFF
--- a/src/prebid.js
+++ b/src/prebid.js
@@ -174,6 +174,14 @@ $$PREBID_GLOBAL$$.setTargetingForGPTAsync = function (adUnit) {
   // now set new targeting keys
   targeting.setTargetingForGPT(targetingSet);
 
+  Object.keys(targeting).forEach((key) => {
+    Object.keys(targeting[key]).forEach((key2) => {
+      if (key2 === 'hb_adid') {
+        auctionManager.setStatusForBids(targeting[key][key2], BID_TARGETING_SET);
+      }
+    });
+  });
+
   // emit event
   events.emit(SET_TARGETING, targetingSet);
 };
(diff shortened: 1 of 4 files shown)


src/prebid.js
```diff
@@ -174,6 +174,14 @@ $$PREBID_GLOBAL$$.setTargetingForGPTAsync = function (adUnit) {
   // now set new targeting keys
   targeting.setTargetingForGPT(targetingSet);
 
+  Object.keys(targeting).forEach((key) => {
+    Object.keys(targeting[key]).forEach((key2) => {
+      if (key2 === 'hb_adid') {
```
mkendall07: why are we only setting `bid_targeting_set` for winning bid (top bid) when the default in prebid 1.0 is sendAllBids?

src/prebid.js
```diff
@@ -174,6 +174,14 @@ $$PREBID_GLOBAL$$.setTargetingForGPTAsync = function (adUnit) {
   // now set new targeting keys
   targeting.setTargetingForGPT(targetingSet);
 
+  Object.keys(targeting).forEach((key) => {
+    Object.keys(targeting[key]).forEach((key2) => {
```
mkendall07: you should name things better :)

src/prebid.js
jaiminpanchal27: https://gist.github.com/mkendall07/1ef1b27ed3501d7b03fbd72ec455bab4
A bid picked for targeting (prebid auction win) will be marked bid.status = 'targetingSet' - which indicates it's not eligible for future auctions, until such status is cleared. This will prevent a winning bid to be set in targeting more than 1 time.

PR #2626, opened 2018-05-25
ConsentManagement: detect __cmp in window.top

benjaminclot:
## Type of change
- [x] Feature

## Description of change
At least one CMP (QuantCast, not to name it) doesn't create the `__cmpLocator` iFrame, so that when prebid is loaded from inside an iFrame it can't locate the CMP. This PR tries to detect if `window.top.__cmp` exists (after checking if `window.top` is accessible) and use it. If not, an error is returned.

## Other information
While I'm aware that this should be fixed on the CMP level, adding this detection isn't too much of a hassle.

### DIFF
--- a/modules/consentManagement.js
+++ b/modules/consentManagement.js
@@ -70,8 +70,8 @@ function lookupIabConsent(cmpSuccess, cmpError, hookConfig) {
   // if the CMP is not found, the iframe function will call the cmpError exit callback to abort the rest of the CMP workflow
   try {
     cmpFunction = window.__cmp || utils.getWindowTop().__cmp;
-  } catch(e) {}
-  
+  } catch (e) {}
+
   if (utils.isFn(cmpFunction)) {
     cmpFunction('getConsentData', null, callbackHandler.consentDataCallback);
     cmpFunction('getVendorConsents', null, callbackHandler.vendorConsentsCallback);
--- a/modules/consentManagement.js
+++ b/modules/consentManagement.js
@@ -54,6 +54,7 @@ function lookupIabConsent(cmpSuccess, cmpError, hookConfig) {
       }
     }
   }
+
   let callbackHandler = handleCmpResponseCallbacks();
   let cmpCallbacks = {};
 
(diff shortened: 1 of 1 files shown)


YOzaz:
I would actually replace this to:
```js
var __cmpFunc = window.__cmp || window.top.__cmp;
if ( __cmpFunc ) ...
```

YOzaz:
Oh come on:
```js
var __cmpFunc = null;

try {
   var __cmpFunc = window.__cmp || window.top.__cmp;
} catch(e) { /** Silence is golden **/ }

if (utils.isFn(__cmpFunc)) ...
```

mkendall07:
Shouldn't QuantCast just become compliant with the CMP spec?

YOzaz:
@mkendall07 good point!

benjaminclot:
@mkendall07 that's what I said in "other information" but considering it's an easy enough modification, I think Prebid.js should cover it, just in case.

YOzaz:
@mkendall07 but even official example on Prebid.og docs doesn't inject an iframe :) from my point of view, this PR is quite logical, but I'd rewrite it like this:
```js
var __cmpFunc = null;
try {
   __cmpFunc = window.__cmp || window.top.__cmp;
} catch(e) { /** Silence is golden **/ }

if (utils.isFn(__cmpFunc)) {
    __cmpFunc('getConsentData', null, callbackHandler.consentDataCallback);
    __cmpFunc('getVendorConsents', null, callbackHandler.vendorConsentsCallback);
  } else if (inASafeFrame() && typeof window.$sf.ext.cmp === 'function') {
    callCmpWhileInSafeFrame('getConsentData', callbackHandler.consentDataCallback);
    callCmpWhileInSafeFrame('getVendorConsents', callbackHandler.vendorConsentsCallback);
  } else {
    // find the CMP frame
    let f = window;
    let cmpFrame;
    while (!cmpFrame) {
      try {
        if (f.frames['__cmpLocator']) cmpFrame = f;
      } catch (e) {}
      if (f === window.top) break;
      f = f.parent;
    }

    callCmpWhileInIframe('getConsentData', cmpFrame, callbackHandler.consentDataCallback);
    callCmpWhileInIframe('getVendorConsents', cmpFrame, callbackHandler.vendorConsentsCallback);
  }
}
```
Difference: no duplicated code.

mkendall07:
@YOzaz 
Which prebid.org example are you referring to?

YOzaz:
@mkendall07 sorry, my bad. It's there now.

benjaminclot:
PR updated.

jsnellbaker:
@benjaminclot and all, thanks for the conversation/feedback here.

I think the updated code may be more ideal here, I'll start testing it out in a bit.  @benjaminclot Can you please take a look at/fix the lint errors in the Travis job that failed in the meantime?

benjaminclot:
@jsnellbaker fixed.

YOzaz:
@mkendall07 actually, it's not there - I was referring to this example: http://prebid.org/dev-docs/modules/consentManagement.html#publishers-not-using-an-iab-compliant-cmp.
In a situation, where *stub* would be loaded in a main window, and Prebid.js in DFP async mode (friendly IFRAME) as a postbid process - without this PR, ConsentManagement module would fail.
I've put a PR to accomodate IFRAME creation process, and other IAB mandatory requirements Prebid.js may (or may not) use in the future: https://github.com/prebid/prebid.github.io/pull/792. It may be an overkill, though.

Issue #2567, opened 2018-05-18
GDPR without CMP?


polapi:
I was there's GDPR ConsentManagement Module. But what if I wish to manage the consents on my own? Are there are ways to do this?

I would expect there would be a global (or per bidder) option to set if user has given a consent. But somehow I can't find it in docs. Is there such thing?

veranevera:
Hi @polapi! You can get test CMP tags here https://sandbox.didomi.io/cmp.html

maxim-novikov:
I second this question. While IAB-compliant tool would be my best choice, right now there are lots of vendors - including OneThatShouldNotBeNamed - not appearing on the vendor list, making it useless for a time being. So, using another CMP, I just want to have some kind of basic switch, something like setRequestNonPersonalizedAds(), so that there is an easy way to disable or enable ads personalization for all the bidders.

YOzaz:
A second this request. We'll be gathering consents manually as well - not on per-bidder basis, so having ability to carry over general flag would be ideal.

polapi:
I don't think it is possible to do a simple yes/no flag for that. I think many DSPs (to which you can connect using Prebid.js) provide access to multiple SSPs. Each of those SSPs theoretically requires separate consent - you can't just say to DSP: "my user consents to everything". I assume that's why probably the consent string is propagated around. 

Still, it does not mean you need to create this string using specialized CMP. I think it should be legal to gather consent on your own, prepare the consent string on your own and manually provide it to Prebid.js.

mkendall07:
Currently the only way to do this is the implement the IAB `window.__cmp` methods on page that the consent module uses. 

We just added a section here that explains the approach: http://prebid.org/dev-docs/modules/consentManagement.html#publishers-not-using-an-iab-compliant-cmp

YOzaz:
@polapi, @mkendall07 - that's the problem, you can't generate consent string on your own as far as I understand. Consent String SDK (btw, correct link in docs should be: https://github.com/InteractiveAdvertisingBureau/GDPR-Transparency-and-Consent-Framework/tree/master/Consent%20String%20SDK) requires `CmpId` and `CmpVersion` passed through - and if you're not using specialized CMP, there's nothing to input there...

YOzaz:
@mkendall07, docs are missing mandatory function `ping` and iframe `__cmpLocator` injection process as well.

YOzaz:
[...]

PR #2636, opened 2018-05-29
Ability to statically pass consentData

ptomasroos:
## Type of change
- [x] Feature

## Description of change
For some services its reasonable to not render the cmp api on every page view when a given user has already given consent and change this in his or hers settings. We only have 100% signed in users on our impressions and it would be very valuable for us to be able to set the consentData staticly on page rendering in order to speed up loading and retrieval of consentData.

## Other information
Also reported in #2567 the ability to avoid using cmp, I think its fair to use a CMP to generate the consentData in order for it to be compliant but at least pass it when its not changed or altered through the config


test/spec/modules/consentManagement_spec.js
```diff
@@ -45,6 +45,415 @@ describe('consentManagement', function () {
...
     });
+
+    describe('static consent string setConfig value', () => {
+      afterEach(() => {
+        config.resetConfig();
+        $$PREBID_GLOBAL$$.requestBids.removeHook(requestBidsHook);
+      });
+      it('results in user settings overriding system defaults', () => {
+        let staticConfig = {
+          cmpApi: 'static',
```
bretg: Needs to be part of a `consentManagement` object:
```
pbjs.setConfig({
          consentManagement: {
            cmpApi: 'iab',
            timeout: 8000,
            allowAuctionWithoutConsent: false
          }
        });
```
This issue exists in another test in this file.

bretg:
In general I'm good with pubs being in charge of passing consent values. Don't have a strong opinion about implementation approach. In any case, there will need to be docs changes to http://prebid.org/dev-docs/modules/consentManagement.html

test/spec/modules/consentManagement_spec.js
jsnellbaker: This setup is working because the `setConfig` used in the unit test file is calling the local version of the function that's inside `consentManagement.js`; it's not being routed through the `config.setConfig()` function that requires config object names.  

From an off-line discussion, I plan to rewrite the unit tests to better reflect usage; ideally after this PR is finished.

modules/consentManagement.js
```diff
@@ -337,6 +349,15 @@ export function setConfig(config) {
     utils.logInfo(`consentManagement config did not specify allowAuctionWithoutConsent.  Using system default setting (${DEFAULT_ALLOW_AUCTION_WO_CONSENT}).`);
   }
 
+  if (userCMP === 'static') {
+    if (utils.isPlainObject(config.consentData)) {
+      staticConsentData = config.consentData;
+      consentTimeout = 0;
+    } else {
+      utils.logInfo(`consentManagement config with cmpApi: 'static' did not specify consentData. No consents will be available to adapters.`);
```
jsnellbaker: This should likely be an `logWarn` or `logError` instead of a `logInfo` type message; as this is a negative situation that should be highlighted/easy to spot for a dev troubleshooting.

polapi:
Hey, I'd like to push forward this pull request. @ptomasroos  can you help with this?

I tried to help by updating the documentation. Here is my branch:
https://github.com/prebid/prebid.github.io/compare/master...polapi:static-consent-config?expand=1

I am not sure about usage of the changes you made. Please check if I understand correctly. Also I guess the actual pull request should be opened when this one is ready to be merged.

Remaining changes I guess are quite minor, would you have time to deal with them?

Thanks.

ptomasroos:
I will move on with this next week since I'm currently a bit busy on schedule @polapi

modules/consentManagement.js
ptomasroos: Thanks, will fix!

test/spec/modules/consentManagement_spec.js
ptomasroos: All right! Cool! Thanks for the clarification

polapi:
Hi @protomasroos . You mentioned you'll push the pull request forward last week. Is there anything I can help with to make it happen?

polapi:
Hi @protomasroos, It seems you don't have time to work on this. Maybe it would be better to close the pull request so maybe someone else would implement this functionality (instead thinking you're working on this)?

ptomasroos:
I'm on vacation, feel free to fork otherwise i will complete this within a few weeks when coming back.

ptomasroos:
Good news coming up, will fix this during next week when I'm back at work since we need it now as well

polapi:
@ptomasroos  How is it going? I guess you will again announce you are going to work on this next week?

ptomasroos:
@polapi right, its not easy to be CEO, CTO, CMO, COO and CCO. But I will do my best to complete this, since you're having plenty of free time, feel free to press the fork button and complete this? Otherwise I will move on to do this when this is highest on my prio list. Most likely next week

ptomasroos:
I've now rebased and fixed the minor @jsnellbaker , created the documentation as requested by @bretg

ptomasroos:
The only thing I'm really thinking about here is the format the static data is passed in. Because currently it expects the output from the iab cmp, and when using a lib like the consent-string (from iab) which decodes and encodes the consent string's its returned in another format.

It would probably be wise at least that we match the format in some way from these general libraries published from iab.

https://github.com/InteractiveAdvertisingBureau/Consent-String-SDK-JS
https://github.com/InteractiveAdvertisingBureau/Consent-String-SDK-Java
https://github.com/InteractiveAdvertisingBureau/Consent-String-SDK-C
ect..

Feels like the consentManagement.js expects the results from a iab __cmp implementation and we should probably try to avoid this. And maybe just pass the consent-string into it instead and let prebid decode? Opinions?

test/spec/modules/consentManagement_spec.js
bretg: ok

PR #2650, opened 2018-05-30
Ensuring correct events order with bidsBackHandler

dejanstrbac:
## Type of change
- [x] Bugfix

## Description of change
Events such as BIDDER_DONE will fire only after the bidsBackHandler has completed. This is due to inline execution of the callbacks and handlers. Instead, in this patch, using setTimeout(_cb_, 0), we will postpone execution of the callback after the current execution flow has finished. This makes events fire in the correct order.

### DIFF
--- a/src/auction.js
+++ b/src/auction.js
@@ -131,7 +131,11 @@ export function newAuction({adUnits, adUnitCodes, callback, cbTimeout, labels}) 
         const bids = [_bidsReceived
           .filter(adUnitsFilter.bind(this, adUnitCodes))
           .reduce(groupByPlacement, {})];
-        _callback.apply($$PREBID_GLOBAL$$, bids);
+        setTimeout((function(callback, bids) {
+          return function() {
+            callback.apply($$PREBID_GLOBAL$$, bids);
+          };
+        })(_callback, bids), 0);
       } catch (e) {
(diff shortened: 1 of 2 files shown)


mkendall07:
well since this is all synchronous operations it should be possible. Maybe @jaiminpanchal27 can give you a pointer.

dejanstrbac:
Indeed, the synchronous operations is the problem I see, as there is an attempt to make async operations  act as in sync. All bidder requests that fan out will be inevitably async, coming back as callbacks or interruptions to whatever is executing at the moment. I see it indeed fixable but with major re-factoring of the way events are emitted, - or adding 0 timeouts to execute after the current execution has finished.

May I ask why sync? Why own events system instead of relying on the browser one, async via addEventListener? Publishers could have attached listeners to window or document and listened for custom Prebid events?

jaiminpanchal27:
We own events systems because `Event()` constructor is not supported in IE.  https://developer.mozilla.org/en-US/docs/Web/API/Event/Event

Regarding sync problem for events we will look into it. We accept PR's if you have some bandwidth to do a refactor.

jaiminpanchal27:
@dejanstrbac Any update ?

mkendall07:
closing because of no response.

Issue #2539, opened 2018-05-15
[Bug] getHighestCpmBids not updating on refresh and AppNexus not recording impression


Slind14:
## Type of issue
Bug

## Description
When refreshing ads with prebid, AppNexus does not track these impressions probably due to the adid not updating. When getting `getHighestCpmBids()` it contains the old bids from the first impression.

## Steps to reproduce
1. load and check  `console.table(pbjs.getHighestCpmBids())`, specifically `adid`
2. refresh the add (button) and check `console.table(pbjs.getHighestCpmBids())` => records are unchanged and do not match bid network request response

## Test page
https://jsfiddle.net/prebid_org/61uzp29t

### Expected results
`getHighestCpmBids` containing the new ads

### Actual results
`getHighestCpmBids` not updating (containing the old ads

## Other information
https://appnexustraders.slack.com/archives/C77SFBW5P/p1526394189000702

jsnellbaker:
Hi @Slind14 Thanks for reporting this issue/bug.  I'll start to investigate and see if we can get it addressed.

jsnellbaker:
@Slind14 When I was testing this out and monitoring the refreshed calls, it seems like the bids we received did have different `adid` values.  

If you enable debug mode (either in `setConfig` or appending `?pbjs_debug=true` in the page url), and look for the line in the console titled `Bids Received for Auction with id: ....` there is a copy of the `bidsReceived` object for that auction.  When I inspected these objects, that's where I saw different values for the `adid` field.  

Additionally, when I was monitoring the Network tab and looking for the impression trackers (filter by `/it`), there were distinct URLs executed for each refresh with a slightly different hash used each time.  

I do see that the getHighestCpmBids function is returning the 'oldest' bid and I'm investigating that function further, but it doesn't seem like there's a tracking issue.  Please let me know your thoughts when you get the chance.

mercuryyy:
@jsnellbaker i was also able to get a deferent adid, but my dfp reporting for appnexus is about 20-30% higher impression count then appnexus dashboard since i started refreshing. i refresh appnexus every 60 seconds. 

The overall revenue in dfp and appnexus reporting differs by 5% where total impressions is 20-30%

I though it might have something to do with https://github.com/prebid/Prebid.js/issues/2448
since sometimes it does not pick the new winning bids and uses the old bid.

Slind14:
[...]

PR #2654, opened 2018-05-30
Update bid pool logic to use highest cpm bid

### LINKED ISSUE #2448: Intent to Implement: Update bid pool logic to select highest cpm bid in case of multiple bids
## Type of issue
Feature Request

## Description
While running concurrent auctions there can be multiple bids from same bidder for same adUnit. As of now we use FIFO (First In First Out) to select bid to take part in auction.

#### Problem with above approach
Chance of decrease in revenue for some publishers because there can be a second bid having higher cpm and not getting selected to take part in auction due to FIFO.

#### Solution
Select highest cpm bid if you find multiple bids in the pool

jaiminpanchal27:
Fixes #2448

### DIFF
--- a/src/prebid.js
+++ b/src/prebid.js
@@ -1,7 +1,7 @@
 /** @module pbjs */
 
 import { getGlobal } from './prebidGlobal';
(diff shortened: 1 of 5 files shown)


jsnellbaker:
To note - believe should also fix #2539

PR #2655, opened 2018-05-30
GDPR consent management for madvertise bidder adapter

NLopezMad:
## Type of change

- [ ] Bugfix
- [x] Feature
- [ ] New bidder adapter   
- [ ] Build related changes
- [ ] CI related changes
- [ ] Other

## Description of change
Add GDPR consent management to madvertise bidder adapter

- contact email of the adapter’s maintainer : nicolas.lopez@madvertise.com

https://github.com/prebid/prebid.github.io/pull/810


test/spec/modules/madvertiseBidAdapter_spec.js
```diff
@@ -70,26 +72,70 @@ describe('madvertise adapater', () => {
...
-        adUnitCode: 'div-gpt-ad-1460505748561-0',
-        transactionId: 'd7b773de-ceaa-484d-89ca-d9f51b8d61ec',
-        auctionId: '18fd8b8b0bd757',
-        bidderRequestId: '418b37f85e772c',
-        params: {
-          s: 'test',
+    beforeEach(function () {
+      let mockConfig = {
+        consentManagement: {
+          cmpApi: 'IAB',
```
jsnellbaker: As a note more-so - the value will be `iab` instead of `IAB`.

test/spec/modules/madvertiseBidAdapter_spec.js
```diff
@@ -70,26 +72,70 @@ describe('madvertise adapater', () => {
...
+          timeout: 1111,
+          allowAuctionWithoutConsent: 'cancel'
```
jsnellbaker: Related to above, this should be a `boolean` value (`false` in this case) not the string `cancel`.  Passing in a string here actually causes the config to the opposite value (ie `true`) which allows the auction to proceed without consent.

PR #2664, opened 2018-05-31
Kargo Adapter: Add request timestamp

samuelhorwitz:
## Type of change

- [x] Feature

## Description of change
We want to log the browser's timestamp of when the request was made.

We also are fixing a minor casing issue that doesn't cause issues with our Golang JSON unmarshaling but should be standardized on principle to our canonical name


samuelhorwitz:
Whoops, done

PR #2674, opened 2018-06-04
Update appnexusBidAdapter.js

matthieularere:
## Type of change
- [ ] Feature
- [ ] Other

## Description of change

Adding new aliases to appnexus adapter in order to use them in prebidserver (as the aliasBidder method has no equivalent in prebidserver)

### DIFF
--- a/modules/appnexusBidAdapter.js
+++ b/modules/appnexusBidAdapter.js
@@ -29,7 +29,7 @@ const SOURCE = 'pbjs';
 
 export const spec = {
   code: BIDDER_CODE,
-  aliases: ['appnexusAst', 'brealtime', 'pagescience', 'defymedia', 'gourmetads', 'matomy', 'featureforward', 'oftmedia', 'districtm'],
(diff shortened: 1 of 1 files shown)


dbemiller:
@matthieularere is this what you're looking for in Prebid Server? https://github.com/prebid/prebid-server/blob/master/docs/endpoints/openrtb2/auction.md#bidder-aliases

If you're using the `/openrtb2/auction` endpoint, `Prebid.js` should take any `pbjs.aliasBidder` calls you make in JS and forward them in the request.

matthieularere:
@dbemiller 
I did knew this was possible.

Is this openrtb2 response format compatible with all pbjs object functions ?

dbemiller:
which function(s) do you need to work specifically?

I know that the Prebid.js design goal was to work with PBS's `/openrtb2` endpoint, and I know there's lots of code [inside the PBS Adapter](https://github.com/prebid/Prebid.js/blob/master/modules/prebidServerBidAdapter.js#L451) trying to make it happen. If something doesn't work with openrtb2, you should be able to file an issue about it.

jaiminpanchal27:
Closing this since it is not a valid change.

Issue #2677, opened 2018-06-04
When invalid bidder, PBJS still waits for the timeout before continuing


robertrmartinez:
## Type of issue
BUG in 1.X Versions. 
Works as expected in legacy versions.

## Description
When an invalid bidder is used in an auction. Prebid.js will wait for the invalid bidder to timeout before continuing with the auction and sending results to DFP.

PBJS correctly logs the console log error `Adapter trying to be called which does not exist:`, 
but then does not emit the auction end event until the invalid bidder's timeout event is triggered.

We should be removing the invalid bidder completely instead of waiting around for their timeout to trigger.

Using the same test page:

PBJS 1.13.0 Events Log:
![image](https://user-images.githubusercontent.com/13972794/40936098-0b7b7446-67ef-11e8-87e1-3455bd1da35b.png)

PBJS 0.34.13 Events Log
![image](https://user-images.githubusercontent.com/13972794/40936140-30d8d562-67ef-11e8-8043-1d2f773238c2.png)

Notice how in 1.13 we see a bidder timeout before the auction end event is fired. Which does not make sense since the invalid bidder is never actually requested!

The major issue here is that if a publisher has a lower timeout threshold for the pbjs auction than pbjs has for the bidders, we often see that the request for DFP will not go through with the other bidders targeting values.

## Steps to reproduce
- Test page with two bidders
- One valid, one invalid

### Expected results
When an invalid bidder is included in the auction, PBJS should remove that bidder completely from the auction and not ever emit a timeout for them!

### Actual results
PBJS is waiting for the invalid bidder to return, and when it does not emits a timeout event.

## Platform details
Any

Deimos01:
Hi @robertrmartinez ,

I think this issue is related to my issue #2648 and this other one #2637. It's probably why we are all facing a drop of adserver requests and timeout events.

headerbidding:
Yes, I have noticed that also. Here is a test page with a prebid.js that does not include EBDR as a bidder:
http://www.flixxy.com/prebid-test.htm?pbjs_debug=true

There is a "WARNING: Error calling onTimeout of ebdr" and no bids are returned.

headerbidding:
Here is another scenario that happened today:

 

Temporary timeout by Rhythmone ad server resulting in no bids being returned:

 

cid:image001.png@01D3FD77.262B64F0

 

 

 

cid:image002.png@01D3FD77.262B64F0

 

cid:image004.png@01D3FD77.262B64F0

[...]

PR #2695, opened 2018-06-07
check if bidder exists before making request object

jsnellbaker:
## Type of change

- [x] Bugfix

## Description of change
To fix the issue #2677

Currently when an invalid/unknown bidder is used in an auction, a request object is still made for that bidder.  When the auction is running, it will recognize the bidder wasn't real and would return an error.  But the `done` related code used latter in the auction would still try to wait for that bidder's request to respond causing the auction to wait for the timeout.  

This fix performs the check to validate the bidder earlier, and if the bidder is rejected then the request object doesn't get made.  This streamlines the latter logic to only wait for bidders that are real/registered.

### DIFF
--- a/src/adaptermanager.js
+++ b/src/adaptermanager.js
@@ -203,7 +203,12 @@ exports.makeBidRequests = function(adUnits, auctionStart, auctionId, cbTimeout, 
       auctionStart: auctionStart,
       timeout: cbTimeout
     };
-    if (bidderRequest.bids && bidderRequest.bids.length !== 0) {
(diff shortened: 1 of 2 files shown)


robertrmartinez:
@jsnellbaker Pulled this branch and confirmed it fixes #2677.

Thanks!

jsnellbaker:
@mkendall07 I removed the redundant check in the `callBids` function.

harpere:
@jsnellbaker  this has conflicts that need to be resolved.

jsnellbaker:
@harpere the conflict should be resolved now.

PR #2696, opened 2018-06-07
MANTIS bidder: support tracking viewability of ad

parisholley:
in our pre 1.x implementation, we had this code on the JSONP callback, need to embed in bidder now to support this tracking (otherwise we won't serve to publishers).


idettman:
@mkendall07 can I get your opinion on this PR, specifically the use of setInterval?

Issue #2676, opened 2018-06-04
getAdserverTargeting sometimes returns targets with adid from previous auctions


drdmitry:
## Type of issue

Question

## Description

**pbjs.getAdserverTargeting()** sometimes getting targetings with AdIds which were used in previous auctions (and Ad was rendered). How to use AdIds returned in the latest requestBids call ?

![image](https://user-images.githubusercontent.com/4129726/40976854-9c92a5c2-68cf-11e8-8cd3-c6e150bb3371.png)

## Steps to reproduce

When requesting bids multiple times using **pbjs.requestBids**, sometimes **pbjs.getAdserverTargeting()** call inside bidsBackHandler returns targeting with AdIds from the previous requestBids calls.

## Test page

https://jsbin.com/xaruyupeku/edit?html,output
Open page, open console, click refresh button and check that targeting.adid is equals bid.adid

### Expected results
**pbjs.getAdserverTargeting()** should return targetings with AdIds from the last requestBids call.

### Actual results
**pbjs.getAdserverTargeting()** sometimes returns targetings with AdIds from the previous calls.

## Platform details

Prebid.js 1.12.0

## Other information

enableSendAllBids: true

jaiminpanchal27:
@drdmitry I had a local PR with the fix. I think this might be the reason for #2648 
I will be pushing my branch now and ask everyone to test with that prebid.js build if it solves the impression problem. 

Thanks.

derdeka:
Thanks @drdmitry, @jaiminpanchal27, this solves the "negative yield" bug for us, as an already rendered ad is served multiple times, but paid only once.

Additionally we still have a drop in impressions as described in #2648.

mercuryyy:
@jaiminpanchal27 ran the "getAdservertargeting-bugfix" build seems to fix the problem but not the issue in https://github.com/prebid/Prebid.js/issues/2648

Perhaps should consider changing core behavior for action functions?

johngeorgewright:
I opened this pull request a little while ago:

https://github.com/prebid/Prebid.js/pull/2697

To me it looks like Prebid is retrieving bids that have expired only. But obviously the tests then went berserk, and I thought... "this can't be right, I must be misunderstanding the logic here". But the more I test this version with prod data, the more it seems that this might be the solution. Does it work for anyone else?

rmertens:
We also see this issue of rendering ads from previous auctions (on both 1.13 and 1.14).

mercuryyy:
[...]

PR #2697, opened 2018-06-07
fix: retrieve unexpired bids only

johngeorgewright:
## Type of change

- [x] Bugfix
- [ ] Feature
- [ ] New bidder adapter   
- [ ] Build related changes
- [ ] CI related changes
- [ ] Other

## Description of change

We're often receiving expired bids. I think there's some logic which is incorrect.

## Other information

https://github.com/prebid/Prebid.js/issues/2676

### DIFF
--- a/src/targeting.js
+++ b/src/targeting.js
@@ -195,7 +195,7 @@ export function newTargeting(auctionManager) {
   function getBidsReceived() {
     const bidsReceived = auctionManager.getBidsReceived()
       .filter(isUnusedBid)
-      .filter(exports.isBidExpired)
+      .filter(bid => !exports.isBidExpired(bid))
     ;
 
     return getHighestCpmBidsFromBidPool(bidsReceived, getOldestHighestCpmBid);


johngeorgewright:
Sorry, I think I jumped the gun here. This isn't the cause of our issue.

hgoebl:
I think your fix would be correct if `isBidExpired` were doing like the name suggests. Actually, a better name for `isBidExpired` would be `isBidNotExpired`, because the function returns `true` if the end-time of TTL is greater (=later) than current timestamp.

johngeorgewright:
Hey @hgoebl, isn't that what an expired bid is?

If the current timestamp is less than the response time + TTL, then it's an unexpired bid, right?

johngeorgewright:
This is what confused me for a bit.

hgoebl:
Yes, you're right.
Either the name of the function is wrong, or the implementation. And as it's used like "not expired", my conclusion is that the name is wrong.

PR #2700, opened 2018-06-08
QuantumBidAdapter usersync bugfix

sami-elasticad:
- [x] Bugfix
- [ ] Feature
- [ ] New bidder adapter   
- [ ] Build related changes
- [ ] CI related changes
- [ ] Other


malves:
Hi @sami-elasticad 

In the condition at line 305 (and multiple times after) you try to get the serverResponses.body property value.

serverResponses seems to be an array so you cannot read the body property this way. 

You should loop through all elements in serverResponses

sami-elasticad:
Hi @malves our ssp tag responds all the time only with one element not with an array. because of that we don't  loop the response body.

modules/quantumBidAdapter.js
```diff
@@ -297,23 +297,22 @@ export const spec = {
...
-      syncs.push({
-        type: 'image',
-        url: serverResponses[0].body.sync[0]
-      });
-    }
+  getUserSyncs: function (syncOptions, serverResponse) {
+    const syncs = [];
+    utils._each(serverResponse, function(serverResponse) {
+      if (serverResponse.body && serverResponse.body.sync) {
+        const syncType = syncOptions.pixelEnabled ? 'image' : 'iframe';
```
jsnellbaker: To clarify the intent here - do you want to only use one type of pixel per auction?

Further, the `userSync` feature defaults the `pixelEnabled` field to `true` and the iframe pixels are opt-in via the config.  So your iframe pixels will only drop with specific config options; ie 
```
userSync: {
  pixelEnabled: false,
  iframeEnabled: true
}
```
My assumption is this would create a heavier bias towards image-based pixels.  Is that a concern?

sami-elasticad: Hi. Thank you for your review. From our side all sync pixels are images. I will modify to send all the time 'pixel' type.

Issue #2659, opened 2018-05-31
Calling Prebid Server Bidders who aren't in Prebid.js


dbemiller:
## Type of issue
question, maybe feature request

## Description

Oath just submitted a Bidder to Prebid Server in https://github.com/prebid/prebid-server/pull/520... but I don't see them in Prebid.js anywhere.

Question is: If a publisher uses Oath as a bidder, uses PBS bidder params in their ad units, and lists `oath` in their `s2sconfig`, will Prebid.js include them properly in the PBS request?

If not: could we make it happen?

mkendall07:
I'm not sure but my guess would be no. 
The good news is Oath is the new company formed for AOL so I think they do a have an adapter, https://github.com/prebid/Prebid.js/blob/master/modules/aolBidAdapter.js the names just don't match.

dbemiller:
I pinged them on their PR and linked them here... will see what they say.

kizzard:
The Oath/Brightroll PBS adapter is not the same as the AOL adapter in PB - they hit totally different systems. You are correct that there is no matching adapter in PB.

mkendall07:
@kizzard 
Thanks for the clarification. I think this might work ok even if there isn't a client side adapter. Will investigate a bit.

mercuryyy:
How would you request a build without a client side adapter?

jsnellbaker:
@mercuryyy I believe you would just choose the Prebid Server adapter to make the build.  You would also include the DFP Video module if you were using DFP for instream video delivery.

I've been investigating into the changes required to allow a Prebid Server bidder to make a request through Prebid.js without setting up a corresponding adapter file in Prebid.js.  I believe I have it working for both banner/video delivery and will plan to put up a PR soon.

PR #2704, opened 2018-06-08
allow s2s bidders call pbs without need of a client adapter file

jsnellbaker:
## Type of change
- [x] Feature

## Description of change
To add support for the request noted in #2659 

This will allow a pbs bidder to make a request through prebid.js without needing a corresponding client-side adapter file setup in prebid.js.

These changes are mainly to support video-type requests through this type of workflow.


bretg:
While it's ok for s2s to not have a client-side adapter, vendors and pubs should realize it's potentially a disadvantage to not include the client in the PBJS package. Even if the main auction logic happens on the server side, there may be client-side functions -- currently there's usersync, but I believe there was another feature discussed.

jsnellbaker:
@bretg Thanks for the feedback.  I'll put together a docs PR soon.

jsnellbaker:
To reviewers - I realized there is a workflow regarding aliasing bidders that I didn't incorporate when putting these changes together.  Please wait from reviewing until I compile these additional changes and update the PR.

jsnellbaker:
I pushed in the updates to support aliasing bidders.

jsnellbaker:
Created docs PR for these updates: https://github.com/prebid/prebid.github.io/pull/834

smithaammassamveettil:
We are waiting for this PR merge to start our integration testing for S2S.
Can you please let us know what is the ETA for merging this PR?

modules/prebidServerBidAdapter.js
```diff
@@ -734,7 +735,13 @@ export function PrebidServer() {
       utils.logError('error parsing response: ', result.status);
     }
 
-    done();
+    const videoBid = bids.some(bidResponse => bidResponse.bid.mediaType === 'video');
+    const cacheEnabled = config.getConfig('cache.url');
+
+    // video bids with cache enabled need to be cached first before they are considered done
```
mkendall07: is this a separate change?

jsnellbaker: No those changes are part of this PR.  They are needed to ensure pbs video bids that are using prebid cache are properly used in the auction.  Without this change; the adapter calls the `done()` before the caching process has finished which ends the auction prematurely.

smithaammassamveettil:
*@jsnellbaker <https://github.com/jsnellbaker> *
What is the status of this PR? One of our publisher is ready to start
testing our brightroll S2S adapter. Can they do that with existing
Prebid.js they have already? Or is this PR a blocker for testing? If so,
when can I expect these changes? Please advise.

mkendall07:
@smithaammassamveettil 
This still needs review - hopefully should be in the next release. @jsnellbaker need to resolve 1 conflict.

jsnellbaker:
@mkendall07 I resolved the conflicts.  Please take a look over again when you have the chance.  Thanks.

PR #2708, opened 2018-06-11
RTB House adapter update

piwanczak:
## Type of change
- [x] Bugfix

## Description of change
Add support for single bracket sizes array, ex: [300, 250]
next to canon one - [[300, 250]]

### DIFF
--- a/modules/rtbhouseBidAdapter.js
+++ b/modules/rtbhouseBidAdapter.js
@@ -41,7 +41,7 @@ function mapBanner(slot) {
  */
 function mapSizes(slot_sizes) {
   const format = [];
-  if (typeof(slot_sizes[0]) == 'number') {
+  if (typeof slot_sizes[0] == 'number') {
     // most probably size is formatted .sizes: [300,250]
     format.push({
       w: slot_sizes[0],
--- a/modules/rtbhouseBidAdapter.js
+++ b/modules/rtbhouseBidAdapter.js
@@ -32,8 +32,6 @@ function mapImpression(slot) {
  */
 function mapBanner(slot) {
   return {
-    w: slot.sizes[0][0],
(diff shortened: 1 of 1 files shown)


piwanczak:
@bretg the latter one. I'm assuming AdUnit.mediaTypes.banner.sizes should be either be single bracket, two elem array or array of two elem arrays. Fix is based on inconsistent publisher config - sizes could be either of the two.
Is there a trap I'm about to fall into that I'm not seeing? I'll be happy to fix if you think there is a better way to do this.

bretg:
@mkendall07 - have we heard of other cases where AdUnit.mediaTypes.banner.sizes is [w,h] rather than [[w,h]]?

The underlying suggestion here is to have the platform convert to double array if only a single array is provided. If this is a common scenario, then having just one adapter cover it doesn't do as much good as the platform converting to the expected form.

mkendall07:
>The underlying suggestion here is to have the platform convert to double array if only a single array is provided.
prebid core should do this but I don't think it does. 
I was going to open an issue for it but figured it take as much time to just fix it. https://github.com/prebid/Prebid.js/pull/2738

harpere:
@piwanczak #2738, once merged, will fix the inconsistency and you'll be able to depend on a double array [[w,h]]

bretg:
So... the implication is that this change shouldn't be needed after #2738 ... would you agree @piwanczak ?

piwanczak:
@bretg Correct, once #2738 is merged my fix is unnecessary. 
Thank you as well as @mkendall07 !
I'll sub the other one and close my pull once #2738 is merged

harpere:
@piwanczak #2738 has been merged

bretg:
Good news. Closing.

piwanczak:
Thank you!

Issue #2707, opened 2018-06-08
Consent management hook being called multiple times in requestBids test suite


snapwich:
## Type of issue
bug

## Description
The consent management requestBids hook is not being properly cleaned up in the test suite.  In the process of writing some `requestBids` tests I was stepping through code in `test/spec/unit/pbjs_api_spec.js` and each of the `requestBids` calls were going through the consent management hook multiple times before `requestBids` was being ran.

It doesn't seem to cause any issues but it probably slows down our tests.  It could be an issue in the consent management spec or it could be an issue with the hooks removal function; haven't dug that far yet.

### Expected results
Stepping through any of the `$$PREBID_GLOBAL$$.requestBids()` calls in `test/spec/unit/pbjs_api_spec.js` calls the consent management hook either once or not at all.

### Actual results
Stepping through calls the consent management hook a lot of times before actually calling `requestBids`.

## Other information
@jsnellbaker

jsnellbaker:
@snapwich I am calling the `$$PREBID_GLOBAL$$.requestBids.removeHook(requestBidsHook);` in the `afterEach` sections for the various unit tests in `consentManagement_spec.js` which I thought would clean up the hooks.

I'll take a look at this again.

jsnellbaker:
Found the issue - there were other unit tests in the `prebidServerBidAdapter_spec.js` and `smartadserverBidAdapter_spec`that created a hook but didn't remove it.  I'll make a PR shortly to fix these tests.

PR #2711, opened 2018-06-11
update some unit tests to clean-up consentManagement hooks

jsnellbaker:
## Type of change

- [x] Bugfix

## Description of change
This fixes issue #2707 

Several unit tests were calling the `config.setConfig` to initialize the `consentManagement` module, but they weren't cleaning up after the test which left a lingering hook on the `pbjs.requestBids` function that was being used in latter unrelated tests (specifically in the `pbjs_api_spec.js` file).

This change adds the necessary clean-up on those tests.


jsnellbaker:
@snapwich Thanks.  I checked the addition and it LGTM.

Issue #2720, opened 2018-06-12
Prebid timeout of 10 seconds is still honored if gdprApplies:false is sent


tpottersovrn:
## Description

The IAB CMP Stub is causing prebid to wait for the default timeout until loading ads. Even if gdprApplies is passed in before prebid is fired. We have verified that the CMP is passing the gdprApplies:false flag. It appears that prebid is ignoring the flag and only looking at the iab stub.

If we set the timeout to '0' then run an auction then gdpr is set to null on the first run.

## Steps to reproduce

Attach the IAB CMP Stub to the top of the <head> of the page and have the cmp on the next script loaded then load prebid and there is no CMP showing for a non-EU reader. The auction is not being held until the default timeout is reached. 
### Expected results
If the reader is not within the EU and gdprApplies is set to false then the auction should be held immediately with no timeout.
### Actual results
The auction is being suspended until the timeout is reached for CMP being detected on the page.

## Platform details

Prebid 1.11.0
Chrome V: 66.0.3359.139/ OSX 10.13.4
npm: 4.3.0

jsnellbaker:
@tpottersovrn Thanks for reporting this issue.  I'll do some digging.  

If you have a test page with the CMP already setup to pass in false for `gdprApplies` for non-EU users, can you please share/link it here so I can be sure the fix is working in your scenario?

tpottersovrn:
@jsnellbaker I'll see if I can get something that is external facing

jsnellbaker:
@tpottersovrn I put together a fix that I believe should address the issue.  see #2723

I would still like to give it a try with your test page if possible; alternatively if you want to check out the branch and give it a try yourself - please let me know if it works.  

Please note - this depends on how the CMP is configured when the `gdprApplies` is passed in as `false`.  If the CMP knows to immediately respond to requests to its API with `gdprApplies` as `false` and the other consent fields as `null`, then this fix should help.  When the module sees `gdprApplies` is set to `false`, it will skip the checks we normally do on the CMP response and proceed as normal.

If the CMP is configured to wait for consent information before responding to any requests to the CMP API, then this fix won't address that workflow.  The module is reliant on the CMP to respond so that it can process the potential consent information for the end-user.  The CMP may be waiting because it thinks there may be a user interacting with the consent UI; thus it thinks it should hold any requests to its API in order to provide that pending consent information.

tpottersovrn:
[...]

PR #2723, opened 2018-06-13
update consentManagement error logic/handling

jsnellbaker:
## Type of change

- [x] Bugfix

## Description of change
Fix for #2720 

This change will skip the check to evaluate the CMP results if the `gdprApplies` field is set to `false`.  This change is meant to support CMP's that will not collect any consent information for any non-EU end-users; their response in this scenario would be `gdprApplies=false` and `null` for any other consent fields.

This change also updates the error message to better present the CMP response in the console.  Previously the `consentObject` in the log message would just show [Object object] instead of the actual object.


tpottersovrn:
Excellent!! I will give it a try.

PR #2738, opened 2018-06-18
Feature/normalize size

mkendall07:
## Type of change

- [ ] Bugfix
- [x] Feature

## Description of change
This is long overdue. Normalized the banner sizes param to always be `[[h,w]]`

Ignore the merge commits please. Thanks


robertrmartinez:
Hi, wondering if we can do the same fix for `adUnit.sizes` as well?

PR #2740, opened 2018-06-18
catch find, includes usage before merge to master

mkendall07:
## Type of change

- [ ] Bugfix
- [x] Feature

## Description of change
If this works, I would expect travis to fail this build (since our master is currently broken).

Once https://github.com/prebid/Prebid.js/tree/gumgum-unit-test-fix is merged and rebased I would expect this PR to pass.

### DIFF
--- a/test/test_index.js
+++ b/test/test_index.js
@@ -1,4 +1,8 @@
 require('test/helpers/prebidGlobal.js');
-
+/* eslint-disable */
+// this is a hack, but it allows chrome to catch issues with some IE undefined values.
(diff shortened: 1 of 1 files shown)


jaiminpanchal27:
@mkendall07 Is this error expected ? It is in core-js module.
`The value of the property 'find' is not a Function object\nat webpack:///node_modules/core-js/library/modules/es6.array.find.js:`

mkendall07:
hmm yeah that doesn't look right. This worked locally. Not sure what's different when it runs on travis.

PR #2755, opened 2018-06-21
Add gdpr and targeting support for Yieldlab adapter

mirkorean:
## Type of change
- [x] Feature
https://github.com/prebid/prebid.github.io/pull/851

## Description of change
- added support for the gdpr consent management module 
- added support for key-value targeting 

solutions@yieldlab.de

### DIFF
--- a/modules/yieldlabBidAdapter.js
+++ b/modules/yieldlabBidAdapter.js
@@ -27,16 +27,24 @@ export const spec = {
   buildRequests: function (validBidRequests) {
     const adslotIds = []
     const timestamp = Date.now()
+    const query = {
+      ts: timestamp,
+      json: true
+    }
 
     utils._each(validBidRequests, function (bid) {
       adslotIds.push(bid.params.adslotId)
+      if (bid.params.targeting) {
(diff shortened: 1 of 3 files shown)


modules/yieldlabBidAdapter.js
```diff
@@ -24,19 +24,32 @@ export const spec = {
...
     utils._each(validBidRequests, function (bid) {
       adslotIds.push(bid.params.adslotId)
+      if (bid.params.targeting) {
+        query.t = createQueryString(bid.params.targeting)
+      }
     })
 
+    if (bidderRequest && bidderRequest.gdprConsent) {
+      query.consent = bidderRequest.gdprConsent.consentString
+      query.gdpr = (typeof bidderRequest.gdprConsent.gdprApplies === 'boolean') ? bidderRequest.gdprConsent.gdprApplies : true
```
harpere: just verifying that you want to default `gdprApplies` to `true`

mirkorean: Yes, gdprApplies=true should be the default. Thanks for making sure!

PR #2758, opened 2018-06-21
Refactored URL query parameter passthrough for additional values, cha…

mrcrawfo:
…nged SSP endpoint to v.lkqd.net, and updated associated unit tests

## Type of change

- [x] Bugfix
- [ ] Feature
- [ ] New bidder adapter   
- [ ] Build related changes
- [ ] CI related changes
- [ ] Other

## Description of change

```
{
  bidder: '<bidder name>',
  params: {
    // ...
  }
}
```

### DIFF
--- a/modules/lkqdBidAdapter.js
+++ b/modules/lkqdBidAdapter.js
@@ -4,41 +4,13 @@ import { VIDEO } from 'src/mediaTypes';
 
 const BIDDER_CODE = 'lkqd';
 const BID_TTL_DEFAULT = 300;
-const ENDPOINT = 'https://ssp.lkqd.net/ad?pid=[PLACEMENT_ID]&sid=[SITE_ID]&output=[OUTPUT]&execution=[EXECUTION]&placement=[PLACEMENT]&playinit=[PLAY_INIT]&volume=[VOLUME]&timeout=[TIMEOUT]&width=[WIDTH]‌&height=[HEIGHT]&pbt=[PREBID_TOKEN]‌&dnt=[DO_NOT_TRACK]‌&pageurl=[PAGEURL]‌&contentid=[CONTENT_ID]‌&contenttitle=[CONTENT_TITLE]‌&contentlength=[CONTENT_LENGTH]‌&contenturl=[CONTENT_URL]&prebid=true';
+const ENDPOINT = 'https://v.lkqd.net/ad';
 
-const PID_KEY = '[PLACEMENT_ID]';
-const SID_KEY = '[SITE_ID]';
-const OUTPUT_KEY = '[OUTPUT]';
-const EXECUTION_KEY = '[EXECUTION]';
-const PLACEMENT_KEY = '[PLACEMENT]';
-const PLAYINIT_KEY = '[PLAY_INIT]';
-const VOLUME_KEY = '[VOLUME]';
-const TIMEOUT_KEY = '[TIMEOUT]';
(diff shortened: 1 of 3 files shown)


modules/lkqdBidAdapter.js
```diff
@@ -101,29 +61,101 @@ function buildRequests(validBidRequests) {
...
+      }
+      if (bidRequest.params.hasOwnProperty('idfa') && bidRequest.params.idfa != null) {
+        sspData.idfa = bidRequest.params.idfa;
+      }
+      if (bidRequest.params.hasOwnProperty('gdpr') && bidRequest.params.gdpr != null) {
+        sspData.gdpr = bidRequest.params.gdpr;
+      }
+      if (bidRequest.params.hasOwnProperty('gdprcs') && bidRequest.params.gdprcs != null) {
+        sspData.gdprcs = bidRequest.params.gdprcs;
+      }
```
harpere: just verifying that you expect `gdpr` params to be passed through in the bid params rather than the `bidderRequest.gdprConsent` object generally sent to `buildRequests()`: `buildRequests: function (bidRequests, bidderRequest) {`

modules/lkqdBidAdapter.js
```diff
@@ -101,29 +61,101 @@ function buildRequests(validBidRequests) {
...
+      if (bidRequest.params.hasOwnProperty('contentTitle') && bidRequest.params.contentTitle != null) {
+        sspData.contenttitle = bidRequest.params.contentTitle;
+      }
+      if (bidRequest.params.hasOwnProperty('contentLength') && bidRequest.params.contentLength != null) {
+        sspData.contentlength = bidRequest.params.contentLength;
+      }
+      if (bidRequest.params.hasOwnProperty('contentUrl') && bidRequest.params.contentUrl != null) {
+        sspData.contenturl = bidRequest.params.contentUrl;
+      }
+
```
harpere: just a suggestion to make this more concise - maybe create a function to do this repetitious task.  Something like:
`function _addParam(param, default) {`
`    if (bidRequest.params.hasOwnProperty(param) && bidRequest.params[param] != null) {`
`        sspData[param] = bidRequest.params[param];`
`    } else if (default != null) {`
`        sspData[param] = default;`
`    }`
`}`

then you just repeat 
`_addParam(<bidParam>, <optionalDefault>);`
or could even put it in a loop (loop through all your params)

modules/lkqdBidAdapter.js
mrcrawfo: Thank you very much for confirming on this, it's an interesting point. You are correct that we are using custom parameters for our GDPR handling, even though there is likely overlap between your gdprConsent object and our gdprcs (consent string). I was unaware of bidderRequest.gdprConsent so I will have to familiarize myself with the documentation - we will likely add support through this property in a later pull request, but for the moment we are going to proceed with our own implementation. Thank you for bringing this feature to my attention!

modules/lkqdBidAdapter.js
mrcrawfo: This is a good suggestion - I will look into adding this improvement in a future commit but I would consider it out of scope and don't want it to hold up this current pull request.

mrcrawfo:
Thank you harpere for following up on this pull request, I have read your comments and provided my feedback to answer the remaining question. I think we are clear to proceed, but just let me know if you have anything else you would like to review. Thank you!

PR #2761, opened 2018-06-21
Add user sync to Beachfront adapter

jsalis:
## Type of change

- [x] Feature

## Description of change
Adds user sync support to the Beachfront adapter.

### DIFF
--- a/modules/beachfrontBidAdapter.js
+++ b/modules/beachfrontBidAdapter.js
@@ -5,7 +5,7 @@ import { VIDEO, BANNER } from 'src/mediaTypes';
 import find from 'core-js/library/fn/array/find';
 import includes from 'core-js/library/fn/array/includes';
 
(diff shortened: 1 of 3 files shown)


modules/beachfrontBidAdapter.js
```diff
@@ -77,23 +81,56 @@ export const spec = {
...
+    let bannerResponse = find(serverResponses, (res) => utils.isArray(res.body));
+
+    if (bannerResponse) {
+      if (syncOptions.iframeEnabled) {
+        bannerResponse.body
+          .filter(bid => bid.sync)
+          .forEach(bid => {
+            syncs.push({
+              type: 'iframe',
+              url: bid.sync
```
harpere: not adding gdpr params?

jsalis: The gdpr params will be added by the server in this case.

PR #2767, opened 2018-06-22
adding logInfo message to note the consentManagement module was activated

jsnellbaker:
## Type of change

- [x] Other

## Description of change
Adding a simple log message in the browser console to note that the consentManagement module was activated.


mkendall07:
@jsnellbaker 
Looks like you got some failed tests due to the new `logInfo` call. Otherwise LGTM.

jsnellbaker:
Fixed the unit test.

PR #2769, opened 2018-06-25
ADSpend bidder adapter

loorke:
## Type of change

- [x] New bidder adapter  

## Description of change

Add AdSpend bidder adapter

```
    var adUnits = [
        // Banner
        {
            code: 'test-div',
            mediaTypes: {
                banner: {
                    sizes: [[300, 250]]
                }
            },
            bids: [
                {
                    bidder: "adspend",
                    params: {
                        placement: 'test-ad',
                        tagId: 'test'
                    }
                }
            ]
        }
    ];
```

- gaffoonster@gmail.com

### DIFF
--- a/modules/adspendBidAdapter.js
+++ b/modules/adspendBidAdapter.js
@@ -36,7 +36,7 @@ export const spec = {
         'ua': navigator.userAgent,
         'ip': ''
       },
-      'user': { 'id': getUserID(document.cookie) },
+      'user': { 'id': getUserID() },
       'imp': [
         {
           'id': req.params.placementId,
--- a/modules/adspendBidAdapter.js
+++ b/modules/adspendBidAdapter.js
@@ -1,0 +1,120 @@
+import * as utils from 'src/utils';
+import { config } from 'src/config';
+import { registerBidder } from 'src/adapters/bidderFactory';
+import { BANNER, NATIVE, VIDEO } from 'src/mediaTypes';
+
+const BIDDER_CODE = 'adspend';
+const URL = '//rtb.com.ru/headerbidding-bid';
+const COOKIE_NAME = 'hb-adspend-id';
+const UUID_LEN = 36;
+
+export const spec = {
+  code: BIDDER_CODE,
+  aliases: ['adspend'],
+  supportedMediaTypes: [BANNER, VIDEO, NATIVE],
+
+  isBidRequestValid: function(bid) {
+    const conf = config.getConfig();
+    return !!(conf.currency && conf.currency.adServerCurrency && bid.crumbs.pubcid);
+  },
+
+  buildRequests: function(bidRequests, bidderRequest) {
+    const req = bidRequests[Math.floor(Math.random() * bidRequests.length)];
+
+    const conf = config.getConfig();
+    const cur = conf.currency.adServerCurrency;
+
+    const bidfloor = req.params.bidfloor;
+
+    const payload = {
+      'id': req.bidId,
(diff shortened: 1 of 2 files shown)


mkendall07:
This pull request **introduces 1 alert** when merging 26027fccabfff5dd70b02d865cbc899f009fc1db into 5c0856932c6b93e73da8febda7cab19d22282288 - [view on LGTM.com](https://lgtm.com/projects/g/prebid/Prebid.js/rev/pr-75447c30cadf3205db4f7dd00231de1e65bb8888)

**new alerts:**

---

loorke:
So, what about my pull request? Is everything else okay?

loorke:
Did you make a review? Should I add docs?

modules/adspendBidAdapter.js
```diff
@@ -0,0 +1,120 @@
...
+  aliases: ['adspend'],
+  supportedMediaTypes: [BANNER, VIDEO, NATIVE],
+
+  isBidRequestValid: function(bid) {
+    const conf = config.getConfig();
+    return !!(conf.currency && conf.currency.adServerCurrency && bid.crumbs.pubcid);
+  },
+
+  buildRequests: function(bidRequests, bidderRequest) {
+    const req = bidRequests[Math.floor(Math.random() * bidRequests.length)];
```
jsnellbaker: Can you clarify this part further?  It seems like you're just grabbing one request object at random and processing only that.

modules/adspendBidAdapter.js
```diff
@@ -0,0 +1,120 @@
...
+
+const BIDDER_CODE = 'adspend';
+const URL = '//rtb.com.ru/headerbidding-bid';
+const COOKIE_NAME = 'hb-adspend-id';
+const UUID_LEN = 36;
+
+export const spec = {
+  code: BIDDER_CODE,
+  aliases: ['adspend'],
+  supportedMediaTypes: [BANNER, VIDEO, NATIVE],
```
jsnellbaker: Are you planning to support all these mediaTypes?  The rest of the code seems to support only `BANNER`.  If that's the case - I would suggest to only list the types that are currently supported in the adapter and update it later when the additional support is added.

modules/adspendBidAdapter.js
```diff
@@ -0,0 +1,120 @@
...
+    const bids = [];
+
+    if (resp.body === '') return bids;
+
+    const respBid = resp.body.seatbid[0].bid[0];
+    const bid = {
+      cpm: respBid.price,
+      requestId: resp.body.id,
+      width: 300,
+      height: 250,
```
jsnellbaker: Should these values be hard-coded?  Wouldn't the size dimensions be part of the response?

jsnellbaker:
Also - yes when you get the chance please submit a docs PR in the [docs repo](https://github.com/prebid/prebid.github.io).

modules/adspendBidAdapter.js
loorke: Yes, as I said before, we're going to support all these mediaTypes, but currently we don't. I'll remove unsupported mediaTypes for now.

modules/adspendBidAdapter.js
loorke: Exactly. Our bidder don't buy more than one ad placement per page.

modules/adspendBidAdapter.js
loorke: `width` and `height` shouldn't. I'll fix it

Issue #2282, opened 2018-03-19
Multiple Video adUnit requests result in the same identical winner tag (AppNexus module)


paooolino:
## Type of issue
Bug?

## Description
Consecutive requests issued to prebid.js Video results in the same identical VAST winner tag with AppNexus. The adServer consequently does not compute a lot of valid impressions.
The problem was NOT happening in v.1.0.0 or before. I verified it happens with 1.3.0, 1.4.0, 1.5.0.
The workaround I found consists of changing the videoAdUnit code appending a progressive integer to any new request.

## Steps to reproduce
1. go to: http://sftnetwork.com/prebid-same-tag-proof/index.html
2. open Chrome developers console
3. select the 1.5.0 radio button
4. click "run test" multiple times
5. see the tag "s=" parameter in console

## Test page
http://sftnetwork.com/prebid-same-tag-proof/index.html
workaround: http://sftnetwork.com/prebid-same-tag-proof/index-workaround.html

### Expected results
the "s=" parameter should be different for each request.
![image](https://user-images.githubusercontent.com/1051471/37589479-ba6fc810-2b65-11e8-9403-9c76d7b7c88d.png)

### Actual results
the prebid-1.5.0 version results in the same tag with the same "s=" parameter multiple times.
![image](https://user-images.githubusercontent.com/1051471/37589445-a75af592-2b65-11e8-9495-b49083e4ceaf.png)

## Platform details
Windows 10, Chrome 64.
prebid version 1.5.0 (but also tested with 1.3.0, 1.4.0), AppNexus module

## Other information
none

mkendall07:
@paooolino 
thanks for the report. Looks like we need a way to mark the bid as "consumed" during the video workflow. Have you thought about what that API would look like? This didn't happen in < 0.x because we cleared all bids every time a new auction was run.

paooolino:
Hello Matt, it's my pleasure.
In my code I remove the adUnit every time and re-add a new one (with the same adUnit.code).
The "consumed" flag should be transparent and an API should not be needed as it is an internal check, or am I wrong?
Or at least, if possible, there should be an API to clear the bids as you did in the previous versions.

mkendall07:
well in the standard banner workflow when we set keys we mark the bids as used but that doesn't work great for video because we don't get a callback letting us know the bid won.
I like the idea of a clearing API though. Open to a PR for that.

paooolino:
[...]

PR #2777, opened 2018-06-27
New function markWinningBidAsUsed for marking video bids

Caspervw:
## Type of change
- [x] Feature

## Description of change
In accordance with the discussion in https://github.com/prebid/Prebid.js/issues/2282 I’ve created a new function that’s capable of marking a winning bid as used. This addresses the issue where a video-advertisement can be played multiple times on the same page. By proactively calling ```pbjs.markWinningBidAsUsed``` the winning video-bid can be marked as consumed. We currently use it in production and call it straight after the buildVideoUrl function. It resolved most of our discrepancy issues video header bidding partners.

Documentation update: https://github.com/prebid/prebid.github.io/pull/861

### DIFF
--- a/src/prebid.js
+++ b/src/prebid.js
@@ -612,8 +612,20 @@ $$PREBID_GLOBAL$$.getHighestCpmBids = function (adUnitCode) {
  * @alias module:pbjs.markWinningBidAsUsed
 */
 $$PREBID_GLOBAL$$.markWinningBidAsUsed = function (markBidRequest) {
-  const bids = targeting.getWinningBids(markBidRequest.adUnitCode);
-  if (bids.length > 0 && bids[0].adId === markBidRequest.adId) {
+  let bids = [];
+
+  if (markBidRequest.adUnitCode && markBidRequest.adId) {
+    bids = auctionManager.getBidsReceived()
+      .filter(bid => bid.adId === markBidRequest.adId && bid.adUnitCode === markBidRequest.adUnitCode);
+  } else if (markBidRequest.adUnitCode) {
+    bids = targeting.getWinningBids(markBidRequest.adUnitCode);
+  } else if (markBidRequest.adId) {
+    bids = auctionManager.getBidsReceived().filter(bid => bid.adId === markBidRequest.adId);
+  } else {
+    utils.logWarn('Inproper usage of markWinningBidAsUsed. It\'ll need an adUnitCode and/or adId to function.');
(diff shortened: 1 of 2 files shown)


src/prebid.js
```diff
@@ -603,6 +603,18 @@ $$PREBID_GLOBAL$$.getHighestCpmBids = function (adUnitCode) {
     .map(removeRequestId);
 };
 
+/**
+ * Mark the winning bid as used, should only be used in conjunction with video
+ * @param {string} adUnitCode - required ad unit code
+ * @alias module:pbjs.markWinningBidAsUsed
+ */
+$$PREBID_GLOBAL$$.markWinningBidAsUsed = function (adUnitCode) {
```
mkendall07: @Caspervw 
Can you make this accept an object instead of a string?
I'm thinking we would also want to match against a bidId as well as a adUnitCode.
example:
```
{
  adUnitCode : <code>,
  bidId : <id>
}
```

Caspervw: Sure thing

Caspervw:
@mkendall07 I made the change, however I wonder if it became better. If you look at the snippet we now explicitly need to know which bid we’re going to mark. This improves the robustness, however (as far as I know) there is no exposed version of targeting.getWinningBids(adUnitCode);, so which bid did prebid start and what do we need to mark?

Do you think pbjs.getHighestCpmBids() is the best approach for that? What do you recon needs to go on the ...?
```javascript
pbjs.requestBids({
    bidsBackHandler: function(bids) {
        // The video advertisement you want to show
        const bid = ...

        var videoUrl = pbjs.adServers.dfp.buildVideoUrl({
            adUnit: videoAdUnit,
            bid: bid
            params: {
                iu: '/19968336/prebid_cache_video_adunit'
            }
        });

        // Mark the bid, used in buildVideoUrl, as used
        pbjs.markWinningBidAsUsed({
            adUnitCode: videoAdUnit.code
            adId: bid.adId
        });

        invokeVideoPlayer(videoUrl);
    }
});

**Comment by mkendall07** [comment]
@Caspervw 
Thanks. I actually mean those as to be only one required. 
ie - find the highest bid on the AdUnit:
```
 pbjs.markWinningBidAsUsed({
            adUnitCode: videoAdUnit.code
        });
```
or find exact bid:
```
 pbjs.markWinningBidAsUsed({
            adId: bid.adId
        });
```

PR #2784, opened 2018-06-28
ServerBid: Fixing adserver trying to get bidder name from params

Saigredan:
## Type of change
- [ ] Bugfix

## Description of change
When creating userSync URL, serverbid adapter is trying to set bidder name based on params object which is wrong according to documentation and general logic. This problematic especially with connectad alias, which has its own url for userSync when iframeEnabled is true.

### DIFF
--- a/modules/serverbidBidAdapter.js
+++ b/modules/serverbidBidAdapter.js
@@ -70,7 +70,7 @@ export const spec = {
 
     // These variables are used in creating the user sync URL.
     siteId = validBidRequests[0].params.siteId;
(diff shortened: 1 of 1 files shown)


package-lock.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "prebid.js",
-  "version": "1.15.0-pre",
```
mike-chowla: Please remove the change to package-lock.json

Saigredan:
@mike-chowla Change has been made as requested

PR #2786, opened 2018-06-28
Add onBidWon handler to bid adapters spec

Spark-NF:
## Type of change
- [x] Feature

## Description of change
Some bid adapters might want to know when their bid won the auction. Given listening to events is off-limits for bid adapters, they can now add a `onBidWon` handler to their spec that will be called with their winning bid.

Corresponding PR for documentation: prebid/prebid.github.io#864.

### DIFF
--- a/src/adaptermanager.js
+++ b/src/adaptermanager.js
@@ -477,6 +477,19 @@ exports.setS2STestingModule = function (module) {
   s2sTestingModule = module;
 };
 
+function tryCallBidderMethod(bidder, method, param) {
+  try {
+    const adapter = _bidderRegistry[bidder];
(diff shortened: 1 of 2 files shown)


jsnellbaker:
@Spark-NF Thanks for submitting this PR.  

Based on some internal discussion, would it be possible to convert this new field to only handle a pixel URL that would be fired if the bidder won rather than allow a free-form callback method?  

It's likely (unless there's some use-case I'm not aware of) that bidders would ultimately just want to fire off this pixel when they won.  So having the field setup to handle this call directly rather than setup customized code (that would need to be reviewed each time) would be more conducive.

Please me let me know your thoughts on this.  Thanks

PR #2790, opened 2018-06-28
give s2s request the same amount of time from the client side

mkendall07:
## Type of change

- [x] Bugfix

## Description of change
The s2s timeout value was not getting the full requestBids timeout value on the client side. This was causing the request to abort on the client side if you shortened the s2s timeouts.  Example, set s2s timeout to 10ms:
```
pbjs.setConfig({
      "s2sConfig": {
          defaultVendor : 'appnexus',
          accountId: '<id>',
          enabled: true,
          bidders: ['appnexus'],
          timeout: 10,
      }
  });
```
To repro. 
## Other information
May contribute to https://github.com/prebid/prebid-server/issues/592

### DIFF
--- a/src/adaptermanager.js
+++ b/src/adaptermanager.js
@@ -281,7 +281,7 @@ exports.checkBidRequestSizes = (adUnits) => {
   return adUnits;
 }
 
-exports.callBids = (adUnits, bidRequests, addBidResponse, doneCb, requestCallbacks) => {
+exports.callBids = (adUnits, bidRequests, addBidResponse, doneCb, requestCallbacks, requestBidsTimeout) => {
   if (!bidRequests.length) {
     utils.logWarn('callBids executed with no bidRequests.  Were they filtered by labels or sizing?');
     return;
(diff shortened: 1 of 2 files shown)


mercuryyy:
So after the fix can we set s2s timeout to the same value as client bidders timeout or do we still need to leave 300-400ms delay between bidder timeout and s2s timeout ?

Maybe it would be best to disable the s2sconfig timeout value and have prebid automatically set it according the bidderTimeout value?

dbemiller:
> do we still need to leave 300-400ms delay between bidder timeout and s2s timeout ?

You should still leave the delay. It's motivated by network latency & time prebid.js may need to unpack the PBS response and get it into the auction

> Maybe it would be best to disable the s2sconfig timeout value and have prebid automatically set it according the bidderTimeout value?

This is a great idea ^^

mercuryyy:
@dbemiller  @mkendall07 

Found another problem im not sure if it was fixed in the recent pulls i just saw.

If new s2s openrtb2 timeouts it will cause the entire action (client side) to never fire the action so even if there are bids in client side adapter action will never run if s2s timeouts. 

This happens regardless of what timeout value you set s2s config.

jsnellbaker:
In testing the point made by @mercuryyy in this branch, it seems like the bids for the other client-side adapters would still be processed and the winning ad does show.  However, it also seems like the auction doesn't complete in the normal sense.  From the console logs, it stops right after it made all the requests to all the bidders and emitted the various bidRequested events.  

This is a separate matter from the original issue per discussion with @mkendall07 and we'll create a new issue to address this behavior more ideally.  For now, we'll approve this PR and merge it.

PR #2794, opened 2018-06-29
PBS Adapter: call done when ajax request fails

jaiminpanchal27:
## Type of change
- [x] Bugfix

## Description of change
PBS adapter was not calling done when ajax requests fails. done not getting called means auction will not close if all other bidders are back before timeout.


modules/prebidServerBidAdapter.js
```diff
@@ -695,7 +695,10 @@ export function PrebidServer() {
 
     ajax(
       _s2sConfig.endpoint,
-      response => handleResponse(response, requestedBidders, bidRequests, addBidResponse, done),
+      {
+        success: response => handleResponse(response, requestedBidders, bidRequests, addBidResponse, done),
+        error: () => { done(); }
```
dbemiller: Should this error be logged somewhere, so it shows up in debug mode?

If not: `error: done` should work here. No need to wrap the function in another function.

PR #2795, opened 2018-06-29
fix and move logic that updates adUnitsS2SCopy based sizeMapping results

jsnellbaker:
## Type of change
- [x] Bugfix

## Description of change
This PR:
- adds a fix to the logic we were using to update the `bidRequests.adUnitsS2SCopy` object.  This update ensured only valid bids were being included in the s2s request.
- moved this logic so that the updated `adUnitsS2SCopy` object is properly reflected in the console log messages for easier debugging/troubleshooting.

_Additional Context_
Currently for s2s requests, we resync the `adUnitsS2SCopy` object based on the various bids that exist in the `bidRequest` objects.  This resync is to remove any `bids` that were rejected by the sizeMapping/label filtering checks.

Keeping the `adUnitsS2SCopy` in sync is important, as this object (not the `bidRequest`) is used to build the POST request meant for PBS; thus this ensures only validated bids get sent to PBS.

_Notes on Fix_
The logic used for the sync operated on a combination of the `bidder` and `adUnitCode`s to match `adUnit.bid` with `bidRequest.bids` to know if they passed the sizeMapping checks.  This logic was insufficient when the same `adUnitCode` was used across different adUnits that used different labels.  It allowed bids for the wrong label to pass the screening and get added to the PBS request.

The new logic operates on the `bid_id` values, which are generated uniquely for each `adUnit.bid` object and then stored in the corresponding `bidRequest.bid` object. 

By moving this logic to the `makeBidRequest` function, we can display the updated `adUnitsS2SCopy` object in the 'Bids Requested' log message (here https://github.com/prebid/Prebid.js/blob/master/src/auction.js#L188 ) instead of digging through debugger.

### DIFF
--- a/src/adaptermanager.js
+++ b/src/adaptermanager.js
@@ -179,7 +179,6 @@ exports.makeBidRequests = function(adUnits, auctionStart, auctionId, cbTimeout, 
         auctionId,
         bidderRequestId,
         tid,
-        adUnitsS2SCopy,
         bids: getBids({bidderCode, auctionId, bidderRequestId, 'adUnits': adUnitsS2SCopy, labels}),
         auctionStart: auctionStart,
         timeout: _s2sConfig.timeout,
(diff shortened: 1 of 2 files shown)


mercuryyy:
@jsnellbaker tested the pull found an issue. 

The defined sizes for each label dont pass correctly into the adUnitsS2SCopy array.

So for example if we have 

label1 = desktop - sizes = 728,90 300,250
label2 = mobile - sizes = 320,50, 300,250

If appnexus labelany: ['desktop','moble']

If fired from desktop prebid will pass all sizes to adUnitsS2SCopy insted of just "728,90 300,250"

jsnellbaker:
Hi @mercuryyy Can you provide a markup of the adUnits array that replicated the issue you described above?  I have been attempting a few configurations but I haven't replicated the issue yet.

mercuryyy:
@jsnellbaker Are you on the adops Slack channel or have an email i can DM a test page?

jsnellbaker:
Hi @mercuryyy, you can send it to jsnellbaker@appnexus.com

jsnellbaker:
Update on issue mentioned here; I did investigate the test pages provided, but I wasn't able to see the issue described by @mercuryyy.  There were other issues related to Prebid Server, but they were separate from this PR.

mkendall07:
LGTM. Thanks

PR #2809, opened 2018-07-03
Add bid adapter for Ad Up Technology

SteffenAnders:
## Type of change
- [x] New bidder adapter  

## Description of change
Add bid adapter for Ad Up Technology

## Test parameter
```
{
  bidder: 'aduptech',
  params: {
    publisher: 'prebid',
    placement: '12345'
  }
}
```

## Maintainer
- steffen.anders@adup-tech.com 

## Documentation
- PR on the docs repo at https://github.com/prebid/prebid.github.io is [HERE](https://github.com/prebid/prebid.github.io/pull/866)

### DIFF
--- a/modules/aduptechBidAdapter.js
+++ b/modules/aduptechBidAdapter.js
@@ -1,0 +1,79 @@
+import * as utils from 'src/utils';
+import { registerBidder } from 'src/adapters/bidderFactory';
+import { BANNER } from 'src/mediaTypes'
+
+export const BIDDER_CODE = 'aduptech';
+export const PUBLISHER_PLACEHOLDER = '{PUBLISHER}';
+export const ENDPOINT_URL = window.location.protocol + '//rtb.d.adup-tech.com/prebid/' + PUBLISHER_PLACEHOLDER + '_bid';
+export const ENDPOINT_METHOD = 'POST';
+
+export const spec = {
+  code: BIDDER_CODE,
+  supportedMediaTypes: [BANNER],
+
+  isBidRequestValid: (bid) => {
+    return !!(bid &&
+      bid.sizes &&
+      bid.sizes.length > 0 &&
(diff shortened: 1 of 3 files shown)


SteffenAnders:
@jsnellbaker  @mike-chowla 

Thanks for reviewing. 

> When I was testing with the test param information on the hello_world.html test page, I was receiving an empty response from the server. Could you take a look to see why this may be the case? For new adapters, we want to see an ad deliver during the initial review to ensure everything seems to be working.

Our bidder returns an empty object as fallback e.g. if the request runs into an internal timeout. In the normal case a valid bid response should be returned when using the test publisher "prebid". In my local tests with the hello_world.html ~ 9 of 10 bid requests got a valid response. 
Suggestion: If you add the debug parameter (`debug: true`) to the `params` object, you will see more information in our response.


> I noticed in the docs PR you had gdpr_supported: true, but I don't see the normal corresponding code in your buidRequests method that reads the gdpr information prebid collected from the CMP. Could you confirm if the adapter is meant to be GDPR compliant?

You was right. Now we pass the prebid GDRP information to our bidder via `buildRequests`.

I have also rebased this branch to current master, maybe that solved some problems.

jsnellbaker:
Hi @SteffenAnders,
Thanks for the follow-up and making the changes.  

I confirmed the GDPR code is working within the adapter.  However I'm still seeing response issue.

Below is a copy of the response with the debug param included:
```
{"debug_time":81,"debug_reject":"timeout","debug_keywords":"[]"}
```

SteffenAnders:
@jsnellbaker 

We have increased the timeouts of our bidder and fixed some problems with the HTTPS certificate. 
Can you please try it again?

jsnellbaker:
@SteffenAnders Thanks for making the additional changes, it seems to be working now.

PR #2816, opened 2018-07-04
AOL adapter - dropping pixels when creatives in safe frames.

vzhukovsky:
## Type of change

- [x] Bugfix

## Description of change

Fixed dropping pixels when creatives are served into safe frames.

### DIFF
--- a/test/spec/modules/aolBidAdapter_spec.js
+++ b/test/spec/modules/aolBidAdapter_spec.js
@@ -544,12 +544,11 @@ describe('AolAdapter', () => {
 
       expect(formattedPixels).to.equal(
         '<script>var w=window,prebid;' +
-        'for(var i=0;i<10;i++){w = w.parent;prebid=w.$$PREBID_GLOBAL$$;' +
+        'try{for(var i=0;i<10;i++){w = w.parent;prebid=w.$$PREBID_GLOBAL$$;' +
         'if(prebid && prebid.aolGlobals && !prebid.aolGlobals.pixelsDropped){' +
-        'try{prebid.aolGlobals.pixelsDropped=true;' +
+        'prebid.aolGlobals.pixelsDropped=true;' +
         'document.write(\'<pixels-dom-elements/>\');break;}' +
-        'catch(e){continue;}' +
-        '}}</script>');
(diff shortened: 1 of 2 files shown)


test/spec/modules/aolBidAdapter_spec.js
```diff
@@ -544,12 +544,11 @@ describe('AolAdapter', () => {
 
       expect(formattedPixels).to.equal(
         '<script>var w=window,prebid;' +
-        'for(var i=0;i<10;i++){w = w.parent;prebid=w.$$PREBID_GLOBAL$$;' +
+        'try{for(var i=0;i<10;i++){w = w.parent;prebid=w.$$PREBID_GLOBAL$$;' +
```
mike-chowla: 1.0 compliant adapters are not permitted to reference $$PREBID_GLOBAL$$ and not permitted to drop their own pixels.  Looks like this code has been here awhile but is not it is not compliant with the 1.0 adapter rules

vzhukovsky: @mike-chowla 
Hi. Thank you for reviewing.

It has been there since we updated our adapter for 1.0.  In migration PR we provided more details about this feature (https://github.com/prebid/Prebid.js/pull/1693) and it was accepted.

Actually, we don't drop own pixels. It is exactly the same pixels that we drop via `getUserSyncs` but attached to the ad content.

snapwich: I actually rejected the initial pull-request partially because of the custom pixels and it doesn't look like it was resolved before being merged.  I still feel the same way I did then but I'm fine with being overruled; I'll assign @mkendall07 to review and see with how he feels (if this needs to be fixed now or not).

vzhukovsky:
@mkendall07
We support feature that allows dropping the pixels only once (Container has the same for official getUserSyncs functionality). Unfortunately, using globals is the only one way to do.

I'd say let's keep as is with using globals here until the container explicitly restricts dropping pixels inside the creatives.

mkendall07:
You still need to remove references to `$$PREBID_GLOBAL$$`

PR #2831, opened 2018-07-10
try travis_wait to increase timeout

mkendall07:
not sure if this works in a script - testing out. 

Per docs this should extend the timeout to 20 minutes.


mkendall07:
looks like this isn't needed yet.

PR #2832, opened 2018-07-10
added emxdigital alias to appnexus adapter

ncolletti:
As per [recent news](https://globenewswire.com/news-release/2018/05/09/1499361/0/en/bRealTime-and-Clearstream-Merge-to-Form-EMX-a-New-Digital-Media-Marketplace.html) bRealTime has recently formed into EMX Digital. We'd like to include the 'emxdigital' name as an alias for the APN module.

I will be submitting a PR for on the prebid website github as well. Thank you.


rachelrj:
Will the brealtime alias be removed at any point? Or will both continue to work?

PR #2839, opened 2018-07-12
Bug fix iOS - localStorage

AntoineJac:
Local Storage (localStorage or sessionStorage) will be broken in many cases. When it is disabled due to some privacy settings, any attempts to access window.localStorage or setItem() getItem() methods will throw exceptions.

Here are some notable examples where it breaks:
- Mobile Chrome as a webview in many apps (i.e. official Twitter App on Android)
- Chrome and Chrome Mobile with Local storage/cookies disabled (privacy settings).
- Firefox with localStorage disabled.

It is important that using the local storage is made in a try/catch block or Prebid won't work...

It is directly impacting Prebid client monetisation!

## Type of change

- [ ] Bugfix
- [ ] Feature
- [ ] New bidder adapter   
- [ ] Build related changes
- [ ] CI related changes
- [ ] Other

## Description of change

```
{
  bidder: '<bidder name>',
  params: {
    // ...
  }
}
```

### DIFF
--- a/src/debugging.js
+++ b/src/debugging.js
@@ -81,7 +81,7 @@ export function sessionLoader() {
   let overrides;
   try {
     if (!window.localStorage || !window.sessionStorage) throw "exception";
-    storage = window.sessionStorage || window.localStorage;
+    let storage = window.sessionStorage || window.localStorage;
     overrides = JSON.parse(storage.getItem(OVERRIDE_KEY));
   } catch (e) {
   }
--- a/src/debugging.js
+++ b/src/debugging.js
@@ -77,9 +77,11 @@ export function getConfig(debugging) {
 }
 config.getConfig('debugging', ({debugging}) => getConfig(debugging));
 
-export function sessionLoader(storage = window.sessionStorage) {
+export function sessionLoader() {
   let overrides;
   try {
+    if (!window.localStorage || !window.sessionStorage) throw "exception";
+    storage = window.sessionStorage || window.localStorage;
     overrides = JSON.parse(storage.getItem(OVERRIDE_KEY));
   } catch (e) {
   }
--- a/test/spec/debugging_spec.js
+++ b/test/spec/debugging_spec.js
@@ -34,10 +34,11 @@ describe('bid overrides', () => {
     });
 
     it('should happen when configuration found in sessionStorage', () => {
-      sessionLoader({
-        getItem: () => ('{"enabled": true}')
-      });
+      const stub = sinon.stub(global.window.sessionStorage, 'getItem');
+      stub.returns('{"enabled": true}');
+      sessionLoader();
       expect(addBidResponse.hasHook(boundHook)).to.equal(true);
+      stub.reset();
     });
 
     it('should not throw if sessionStorage is inaccessible', () => {


mkendall07:
This pull request **introduces 1 alert** when merging 5a79490e1dceb7e2fa8cd79f999c58cf963e7fd9 into a38a1fbdec2ed02f5f5f0d5e91a1cf3f0a843076 - [view on LGTM.com](https://lgtm.com/projects/g/prebid/Prebid.js/rev/pr-8723c175081dbb09a898bf2a3d2c806a3b21edff)

**new alerts:**

* 1 for Missing variable declaration

---

YOzaz:
Hi! You'd want to do a storage capacity check as well in that case. We're using such approach in our websites:
```js
function html5_storage_supported()
{
	try
	{
		// Check if supported
		if ( !('localStorage' in window) )
			return false;

		// Check if accessible
		if ( window['localStorage'] === null )
			return false;

		// Check if not full
		var test = "html5storagetest";
		window.localStorage.setItem( test, test );
		window.localStorage.removeItem( test );

		return true;
	}
	catch(e)
	{
		return false;
	}
}
```

src/debugging.js
```diff
@@ -77,9 +77,11 @@ export function getConfig(debugging) {
 }
 config.getConfig('debugging', ({debugging}) => getConfig(debugging));
 
-export function sessionLoader(storage = window.sessionStorage) {
+export function sessionLoader() {
   let overrides;
   try {
+    if (!window.localStorage || !window.sessionStorage) throw "exception";
```
harpere: Is this line needed since if they don't exist an exception will be thrown when `storage` is used on line 85?

Also, if we do want to keep this line, this should be `!window.localStorage && !window.sessionStorage`

harpere: is the problem which putting this in the try/catch solves, that merely accessing `window.sessionStorage` or ` window.localStorage` might throw an exception?  

either way, why is line 83 needed?  wouldn't an exception be thrown on lines 84 or 85 if storage wasn't available?

src/debugging.js
```diff
@@ -77,9 +77,11 @@ export function getConfig(debugging) {
...
+    storage = window.sessionStorage || window.localStorage;
```
harpere: if one of the problems is that just accessing `sessionStorage` might throw an exception, while `localStorage` is still available, this is still a problem because the second part of the `||` will never be reached.  You'd have to do something like:
`let storage;`
`try {  storage = window.sessionStorage || window.localStorage }`
`catch(e) { storage = window.localStorage }`

src/debugging.js
YOzaz: Actually, line 84 may not be required as well... if 85 would be changed to:
```js
overrides = JSON.parse(window.sessionStorage.getItem(OVERRIDE_KEY));
```

src/debugging.js
```diff
@@ -77,9 +77,11 @@ export function getConfig(debugging) {
 }
 config.getConfig('debugging', ({debugging}) => getConfig(debugging));
 
-export function sessionLoader(storage = window.sessionStorage) {
+export function sessionLoader() {
   let overrides;
   try {
+    if (!window.localStorage || !window.sessionStorage) throw "exception";
+    let storage = window.sessionStorage || window.localStorage;
```
harpere: thought I had already commented here, but don't see it...
If merely accessing `window.sessionStorage` might throw an exception while `window.localStorage` is still available, then this does not solve it because the second part of the `||` would never be reached.  would need something like:
`let storage;`
`try { storage = window.sessionStorage || window.localStorage; }`
`catch (e) { storage = window.localStorage; }`

src/debugging.js
harpere: might need a modified version of 84 as mentioned below.

snapwich:
@AntoineJac all the usages of sessionStorage are within try/catch other than the default parameter (which I added only for testing).  I think the fix is to just move the setting of the default parameter inside the try catch.  I fixed here: https://github.com/prebid/Prebid.js/pull/2841

mkendall07:
Closing in favor of #2841. @AntoineJac let us know if this doesn't resolve your use case.

PR #2844, opened 2018-07-13
prebidServerBidAdapter cleanup

### LINKED ISSUE #2420: Intent to Implement: PrebidServer Adapter Cleanup
## Type of issue
Feature Request

## Description

For client side header bidding we have got [bidderFactory](https://github.com/prebid/Prebid.js/blob/master/src/adapters/bidderFactory.js) as base adapter for all the bidders. For server side header bidding we use PrebidServer adapter to create bidRequests and interpret bidResponse. Over the course of adding more features, we had to add some hardcoded usage of bidders to do various tasks. Some of theme are updating bid params, convert type of params, set default values and maybe something else in future.

#### Proposed Solution
Extend bidder spec to add new functions to remove all hardcoded occurrences. PrebidServer adapter will use this functions if they exist and update the request payload. Both these functions will be optional

1) getS2SDefaultValues() - This function will return s2s default values for particular bidder. For appnexus it will return this https://github.com/prebid/Prebid.js/blob/master/modules/prebidServerBidAdapter.js#L36
```
/**
 * @return {object} defaultConfig
 */
getS2SDefaultValues: function () {
   // return default values
}
```

2) getS2SBidParams(params) - This function will receive its own bid params and return updated params. 
In case of appnexus, it needs to convert type of params and convert camel case params to underscore. https://github.com/prebid/Prebid.js/blob/master/modules/prebidServerBidAdapter.js#L474
```
/**
 * @param {object} params Bidders own params. `adUnit.bid.params`
 * @return {object} params Updated params
 */
getS2SBidParams: function (params) {
  // update params
  return params;
}
```

jaiminpanchal27:
## Type of change

## Description of change
Fixes #2420


modules/s2sVendors.js
```diff
@@ -0,0 +1,64 @@
...
+    adapter: 'prebidServer',
+    cookieSet: false,
+    enabled: true,
+    endpoint: '//prebid-server.rubiconproject.com/auction',
+    syncEndpoint: '//prebid-server.rubiconproject.com/cookie_sync',
+    timeout: 500
+  }
+}
+
+export const paramTypes = {
```
mkendall07: I don wonder if these belong here. Seems like we've split up transformation in 2 places now (in the bidder `transformBidParams` function and here). I do like these centralized since it's more of a problem with PBS / PBJS being in sync but wondering if they belong here or not.

mkendall07:
@jaiminpanchal27 
Is the intent just to reduce the size of the PrebidServer adapter? I missed the fact that this file is under /modules. Probably needs to move to core or put back into PrebidServer file per Rich's comments.

jaiminpanchal27:
yes intent is to not have any bidder specific hardcoded stuff in prebidServerAdapter. We do not have any base factory for this so have done it like how amp does with its vendors.  https://github.com/ampproject/amphtml/blob/master/extensions/amp-a4a/0.1/callout-vendors.js

@snapwich What are your suggestions ? If we add it to core will it be added to build when prebidServerAdapter is not added to build ?

snapwich:
If you include it in core it will be auto include.  if you want to conditionally include it then it is correct to have it as a module; however if you explicitly require it from another module (with import) then it's no longer conditional and it will be auto-included (when you include that other module, prebidServerBidAdapter in this case).

You'll need to do something like the currency module where it is included separately but then uses hooks to modify functionality, or you can do something like the express module where it exposes a function on the pbjs global and then code outside of the express module looks for it and runs it if present.

src/utils.js
```diff
@@ -1091,3 +1091,38 @@ export function transformBidderParamKeywords(keywords, paramName = 'keywords') {
...
   return arrs;
 }
+
+/**
+ * Try to convert a value to a type.
+ * If it can't be done, the value will be returned.
+ *
+ * @param {string} typeToConvert The target type. e.g. "string", "number", etc.
+ * @param {*} value The value to be converted into typeToConvert.
+ */
```
mkendall07: this is missing a return

PR #2876, opened 2018-07-21
returning tracking pixels

nwlosinski:
## Type of change

- [ ] Bugfix
- [x] Feature
- [ ] New bidder adapter   
- [ ] Build related changes
- [ ] CI related changes
- [ ] Other

## Description of change
1. Drop pixel on auction timeout
2. Drop pixel when auction ends

### DIFF
--- a/modules/justpremiumBidAdapter.js
+++ b/modules/justpremiumBidAdapter.js
@@ -3,8 +3,11 @@ import { getTopWindowLocation } from 'src/utils'
 
 const BIDDER_CODE = 'justpremium'
 const ENDPOINT_URL = getTopWindowLocation().protocol + '//pre.ads.justpremium.com/v/2.0/t/xhr'
-const JP_ADAPTER_VERSION = '1.1'
+const JP_ADAPTER_VERSION = '1.2'
 const pixels = []
+const TRACK_START_TIME = Date.now()
+let LAST_PAYLOAD = {}
+let AD_UNIT_IDS = []
 
 export const spec = {
   code: BIDDER_CODE,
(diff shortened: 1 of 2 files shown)


modules/justpremiumBidAdapter.js
```diff
@@ -54,6 +62,13 @@ export const spec = {
 
     const payloadString = JSON.stringify(payload)
 
+    LAST_PAYLOAD = payload
+    $$PREBID_GLOBAL$$.onEvent('bidWon', (data) => {
```
jaiminpanchal27: @nwlosinski `$$PREBID_GLOBAL$$` usage is not allowed in adapter except for prebid verion.  You can drop pixels using http://prebid.org/dev-docs/publisher-api-reference.html#setConfig-Configure-User-Syncing

nwlosinski: @jaiminpanchal27 thanks for the link but we didn't add this to sync users. We need to gather some additional information to be able better detects why we lose. If we are not allowed to use `$$PREBID_GLOBAL$$` how we can achieve that? Is there any other way to send some tracking when action ends?

mkalafior: @jaiminpanchal27 are you able to provide any input here?

jaiminpanchal27: @nwlosinski Sorry this didn't came to my mind earlier but you can add `onTimeout` and `onBidwon` function to your bidder spec and achieve this. Docs are here http://prebid.org/dev-docs/bidder-adaptor.html#registering-on-timeout

Some of the bidders are using these functions you can check those for reference.

mkalafior: @jaiminpanchal27 yeah but `onBidWon` listener is executed only when we won the bidding where we wanted to send tracking pixel if other parties did that.
So I understand that there is no way that you can accept this PR in current shape?

jaiminpanchal27: @mkalafior yes we cannot as of now.

nwlosinski: @jaiminpanchal27 thanks for your time. For now we removed this part.

PR #2828, opened 2018-07-09
quantcast: Update endpoint

soarez:
## Type of change

- [x] Bugfix
- [ ] Feature
- [ ] New bidder adapter   
- [ ] Code style update (formatting, local variables)
- [ ] Refactoring (no functional changes, no api changes)
- [ ] Build related changes
- [ ] CI related changes
- [ ] Does this change affect user-facing APIs or examples documented on http://prebid.org?
- [ ] Other

## Description of change

Updating the domain in the endpoint used for Quantcast. Also refactored some tests and added new ones.

- contact email of the adapter’s maintainer: igor.soarez@quantcast.com
- [x] official adapter submission

PR #2879, opened 2018-07-23
fix failing unit test within quantcastBidAdapter

jsnellbaker:
## Type of change
- [x] Other

## Description of change
Since the inclusion of #2828 the Travis builds for master have been failing due to a failed unit test for IE 11 within the `quantcastBIdAdapter_spec.js` file.  See job link:
https://travis-ci.org/prebid/Prebid.js/builds/406331258

This PR addresses the test failure by replacing the `new URL()` code (which isn't supported in IE) with a different set of code that performs the same action needed for the test.

CC @soarez

### DIFF
--- a/test/spec/modules/quantcastBidAdapter_spec.js
+++ b/test/spec/modules/quantcastBidAdapter_spec.js
@@ -71,8 +71,8 @@ describe('Quantcast adapter', () => {
 
     it('sends bid requests to Quantcast Canary Endpoint if `publisherId` is `test-publisher`', () => {
       const requests = qcSpec.buildRequests([bidRequest]);
-      const url = new URL(requests[0]['url']);
-      expect(url.hostname).to.equal(QUANTCAST_TEST_DOMAIN);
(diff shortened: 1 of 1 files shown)


test/spec/modules/quantcastBidAdapter_spec.js
```diff
@@ -71,8 +71,8 @@ describe('Quantcast adapter', () => {
 
     it('sends bid requests to Quantcast Canary Endpoint if `publisherId` is `test-publisher`', () => {
       const requests = qcSpec.buildRequests([bidRequest]);
-      const url = new URL(requests[0]['url']);
-      expect(url.hostname).to.equal(QUANTCAST_TEST_DOMAIN);
+      const hostname = requests[0]['url'].split(/[:/]/)[3];
```
jaiminpanchal27: I think we should use this https://github.com/prebid/Prebid.js/blob/master/src/url.js#L27 here.

jsnellbaker:
@jaiminpanchal27 Thanks for the feedback; I switched to the `parse` function.

PR #2887, opened 2018-07-24
Pr 0507 - support for dctr, currency and multi-size ad slots

pm-manasi-moghe:
## Type of change

- [ x] Feature

## Description of change

- Added support for dctr parameter
- Added provision to get bids through the PubMatic Prebid.js adapter in any supported currency 
- Added support to automatically handle multi-size ad slots

- [ x] official adapter submission

https://github.com/prebid/prebid.github.io/pull/890


pm-manasi-moghe:
Hi @jsnellbaker,

The above mentioned test case works fine in our test environment. Could you please provide more details regarding your setup, so that I can try and simulate the same here?

modules/pubmaticBidAdapter.js
```diff
@@ -251,6 +262,14 @@ function _createImpressionObject(bid, conf) {
       h: bid.params.height,
       topframe: utils.inIframe() ? 0 : 1,
     }
+    if (utils.isArray(sizes) && sizes.length > 1) {
+      sizes = sizes.splice(1);
```
jsnellbaker: Hi @pm-manasi-moghe 

I ran the updated unit test and the issue appeared to be the format object was getting an empty array.

I did some further debugging/experimenting and I believe the problem was due to this line of code.

When I added a proper range value as the 2nd param in the `splice` (like the following), the unit test completed successfully in Safari.  I wasn't able to find anything definitive but it seemed like this was needed for the es5 shim of the splice to work.  Can you implement this change?

```
var range = sizes.length - 1;
sizes = sizes.splice(1, range);
```

PR #2904, opened 2018-07-27
alter sonobi unit test to fix Edge test failure

jsnellbaker:
## Type of change

- [x] Other

## Description of change
After https://github.com/prebid/Prebid.js/pull/2891 was merged into master, Travis began reporting test failures in the sonobiBidAdapter_spec.js file for a unit test in Edge.

See link below:
https://travis-ci.org/prebid/Prebid.js/builds/408728279?utm_source=slack&utm_medium=notification

This PR fixes the error by standardizing the `ref` href URL the unit test utilizes to just the domain (since random numbers are sometimes added in the url when running browserstack).

### DIFF
--- a/test/spec/modules/sonobiBidAdapter_spec.js
+++ b/test/spec/modules/sonobiBidAdapter_spec.js
@@ -318,6 +318,10 @@ describe('SonobiBidAdapter', () => {
 
     it('should map bidResponse to prebidResponse', () => {
       const response = spec.interpretResponse(bidResponse, bidRequests);
+      response.forEach(resp => {
+        let regx = /http:\/\/localhost:9876\/.*?(?="|$)/
+        resp.ad = resp.ad.replace(regx, 'http://localhost:9876/');
+      });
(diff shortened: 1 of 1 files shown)


test/spec/modules/sonobiBidAdapter_spec.js
```diff
@@ -318,6 +318,10 @@ describe('SonobiBidAdapter', () => {
 
     it('should map bidResponse to prebidResponse', () => {
       const response = spec.interpretResponse(bidResponse, bidRequests);
+      response.forEach(resp => {
+        let regx = /http:\/\/localhost:9876\/.*?(?="|$)/
```
mkendall07: I'm not sure hardcoded the port is a good way to prevent this error. I'm not sure but some process is generating the port (karma?) and we should probably use whatever it is assigning or figure out another way to prevent this test case failure.

jsnellbaker: Like we discussed offline - I remove the port values from the unit tests.  Please take another look when possible.

Issue #2672, opened 2018-06-03
Criteo - no creative ID


smartbidder1:
Type of issue
bug

Description
when requesting an ad from Criteo bidder the response doesn't contain a unique creative ID.
there is a high importance of relaying on accurate creative id's for blacklists and debugging & reporting of bad ads.

#2374 for example had the same issue but they have fixed it already (very fast I might say).

@ahubertcriteo @Spark-NF

smartbidder1:
@Spark-NF @ahubertcriteo anything?

IsraeliAdblocker:
@Spark-NF @ahubertcriteo ?

smartbidder1:
@mafaigna

mkendall07:
Ping

PR #2913, opened 2018-07-31
Pass Prebid version and update IDs in Criteo bidding adapter

Spark-NF:
## Type of change
- [x] Feature

## Description of change
Passing the Prebid version and correct profile IDs to our bidding endpoint will allow us to better know our publishers' setup globally.

_PS: the "adapter version" jumped from 8 to 11 because of some internal testing versions._


smartbidder1:
can you please answer https://github.com/prebid/Prebid.js/issues/2672 ? we are waiting more than two months now.

PR #2916, opened 2018-07-31
Fix #2851 - updates in prebidServerBidAdapter to support ix bidder

### LINKED ISSUE #2851: PBS index adapter and PBJS index adapter require different type for the same param
## Type of issue
Bug

## Description
According to this https://github.com/prebid/Prebid.js/blob/master/modules/ixBidAdapter.md ix parameters include siteId that must have type string. PBS adapter for indexExchange requires parameter siteID (note the register of the last letter) that has type integer. Obviously, the requirements are incompatible

## Steps to reproduce
1) Add ix bidder under the alias of `indexExchange`, setup it to work with PBS.
2) Somehow prevent PBS adapter from sending the list of aliases to prevent https://github.com/prebid/Prebid.js/issues/2717
3) Run an auction

### Expected results
PBS sucessfuly processes request

### Actual results
```
Invalid request: request.imp[0].ext.indexExchange failed validation.
siteID: Invalid type. Expected: integer, given: string
```

jsnellbaker:
## Type of change

- [x] Bugfix

## Description of change
Fixes #2851 

This PR:
* adds a dedicated alias for the ix bidder so that it's registered properly in the prebid server request (and pubs don't have to specifically register it in the setup code).
* properly converts the `ixBidAdapter` param `siteId` to the `indexExchange` variant of `siteID` while also converting the type from a `string` to an `number`

With this PR, publishers would be able to use the `ix` bidder in a s2s request in the following manner:

```

var adUnits = [{
          mediaTypes: {
            banner: {
            }
          },
          bids: [{
            bidder: 'ix',
            params: {
              siteId: '12345',
              size: [300, 250]
            }
          }]
  
        }];
```

```
pbjs.setConfig({
            debug: true,
            s2sConfig: {
              accountId: 'a3de7af2-a86a-4043-a77b-c7e86744155e',
              bidders: ['ix'],
              enabled: true,
              timeout: 1000,
              adapter: 'prebidServer',
              endpoint: 'http://prebid.adnxs.com/pbs/v1/openrtb2/auction',
            }
          });
```

### DIFF
--- a/modules/prebidServerBidAdapter.js
+++ b/modules/prebidServerBidAdapter.js
@@ -520,6 +520,7 @@ const OPEN_RTB_PROTOCOL = {
         }
 
         if (bid.bidder === 'ix') {
+          aliases.ix = 'indexExchange';
           bid.params.siteID = tryConvertType('number', bid.params.siteId);
           if (bid.params.siteId) { delete bid.params.siteId; }
         }
@@ -551,7 +552,6 @@ const OPEN_RTB_PROTOCOL = {
       request.user = { ext: { digitrust: digiTrust } };
     }
 
-    aliases.ix = 'indexExchange';
     if (!utils.isEmpty(aliases)) {
       request.ext = { prebid: { aliases } };
     }
--- a/modules/prebidServerBidAdapter.js
+++ b/modules/prebidServerBidAdapter.js
@@ -518,6 +518,12 @@ const OPEN_RTB_PROTOCOL = {
             }
           });
         }
+
+        if (bid.bidder === 'ix') {
+          bid.params.siteID = tryConvertType('number', bid.params.siteId);
+          if (bid.params.siteId) { delete bid.params.siteId; }
+        }
+
         acc[bid.bidder] = bid.params;
         return acc;
       }, {});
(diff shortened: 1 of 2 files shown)


modules/prebidServerBidAdapter.js
```diff
@@ -545,6 +551,7 @@ const OPEN_RTB_PROTOCOL = {
       request.user = { ext: { digitrust: digiTrust } };
     }
 
+    aliases.ix = 'indexExchange';
```
dbemiller: is there somewhere you can put this so that it only gets added if `ix` is involved in the PBS auction?

jsnellbaker: Thanks for the suggestion; I pushed in a change that moved the alias registration.

mkendall07:
@indexexchange ping

akira28:
Is there a reason for closing this? Is there another workaround for the issue?

vi-oder-was:
@mkendall07 Same here. Why was this closed and what's the workaround for this issue?

mkendall07:
@vi-oder-was please contact IndexExchange directly - they are handling a fix for this issue.

PR #2926, opened 2018-08-02
Use prebid ajax instead of fetch

EdmonL:
## Type of change
- [x ] Other

## Description of change
fetch isn't supported by Internet Explorer 10 or 11, and prebid supports IE 10+ (not sure if this is a problem of our adapter in prebid 1.x)


EdmonL:
Mischose the merge target.

PR #2928, opened 2018-08-02
Fix kargo bidResponse size to size from bidRequest.

jacekbj:
Kargo always returns 0x0 size even though other size was requested.

## Type of change

- [ ] Bugfix
- [ ] Feature
- [ ] New bidder adapter
- [ ] Build related changes
- [ ] CI related changes
- [x] Other

## Description of change
Kargo returns always 0x0 size, no matter what size set is provided. This is not compliant with how other adapters work. In order to fix this, this PR overwrites the 0x0 size with size taken from bid request.

- contact to the adapter’s maintainer: @samuelhorwitz

### DIFF
--- a/modules/kargoBidAdapter.js
+++ b/modules/kargoBidAdapter.js
@@ -42,8 +42,8 @@ export const spec = {
       bidResponses.push({
         requestId: bids[adUnitId].bidId,
         cpm: Number(adUnit.cpm),
-        width: adUnit.width,
-        height: adUnit.height,
+        width: bids[adUnitId].sizes[0][0],
+        height: bids[adUnitId].sizes[0][1],
         ad: adUnit.adm,
         ttl: 300,
         creativeId: adUnitId,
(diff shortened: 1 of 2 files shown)


modules/kargoBidAdapter.js
```diff
@@ -42,8 +42,8 @@ export const spec = {
       bidResponses.push({
         requestId: bids[adUnitId].bidId,
         cpm: Number(adUnit.cpm),
-        width: adUnit.width,
-        height: adUnit.height,
+        width: bids[adUnitId].sizes[0][0],
```
bretg: Please confirm that the `sizes` you're referring to here is in your response body. My reading of this code is that bids[] is the original PBJS AdUnit structure. 

First problem is that the pbjs AdUnit.sizes attribute is deprecated in favor of AdUnit.mediaType.banner.sizes. But more importantly, the height and width have to be coming from your bid response, not assuming the first size of the requested list.

samuelhorwitz:
The returned sizes from our bid response is not always applicable and so often it will default to 0x0 unless specifically set. If you need us to mirror back the request width and height with every bid we can look into making a fairly straightforward server side change. However I think an adapter change to pull these sizes from the request is the wrong route.

jacekbj:
Thank your @samuelhorwitz, I will contact you directly via e-mail.

Issue #2873, opened 2018-07-20
Kargo adapter bug in bid to ad matching


andyblackwell:
## Type of issue
BUG

## Description
When using a single placementId and more than one ad per auction, this adapter has no way to properly match a bid response to the correct ad from the request.

## Steps to reproduce
Use one placement id on multiple ads in an auction. Easiest way would be to alter the unit test to not use two unique placement on the two dummy ad calls.

## Test page
n/a

### Expected results
One placement Id could be used for multiple ads in an auction, and the resulting bids or no-bids be matched properly to the ads that requested them.

### Actual results
If using one placement Id with multiple ads, only one bid can be returned and it would only match the last ad in the auction.

## Platform details
all iterations of this adapter, from pre-1.0 introduction to the current version

## Other information
In the bid request api call, the only unique identifier used is their placementId.
On response they use only the placementId to match each bid to the proper ad, but the placementId is the key in that object so is unable to handle more than one bid per placementId.
If there are multiple ads in an auction and only one placementId used, then there is no way to properly match a bid response to ad. One will be matched, but it may not be the correct one and looks like it would probably always match the last ad in the auction.
https://github.com/prebid/Prebid.js/blob/1.16.0/modules/kargoBidAdapter.js

The unit test uses two ads and two unique placement ids, so the bug does not present itself.
https://github.com/prebid/Prebid.js/blob/1.16.0/test/spec/modules/kargoBidAdapter_spec.js

We brought this to attention of our Kargo rep during onboarding where we were only supplied a single placementId, and are awaiting more unique placements as a temporary workaround. But as of now this adapter only handles one ad per auction for us.

Looks like this bug has been around since this adapter was first introduced. https://github.com/prebid/Prebid.js/pull/1316

So far my only idea is to force the adapter to send an api request for every bid, unless there is a way to send and receive another unique identifier for each bid in the single request.

@samuelhorwitz do you know if this is a limitation in the API as-is?

samuelhorwitz:
Sorry for the delay in a response. I've been preparing for a business trip/traveling for the last few weeks. I've reached out to our sales engineers about this, but I believe our solution which they also will start to communicate is to use our multiple placement ID solution. This was something we have set up already for other adapters but it appears we will need to use it for Prebid integrations as well.

[...]

PR #2958, opened 2018-08-09
Kargo adapter BidID support

samuelhorwitz:
## Type of change

- [x] Bugfix
- [x] Feature
- [ ] New bidder adapter   
- [ ] Build related changes
- [ ] CI related changes
- [ ] Other

## Description of change

This updates the adapter to point to a new bid endpoint. The endpoint is versioned so that the old endpoint and payload will still work. The new endpoint accepts placements mapped by prebid bid IDs. This allows multiples of the same placement per request/response.

## Other information

This is a fix for the adapter regarding #2873

### DIFF
--- a/modules/kargoBidAdapter.js
+++ b/modules/kargoBidAdapter.js
@@ -12,7 +12,8 @@ export const spec = {
     return !!bid.params.placementId;
   },
   buildRequests: function(validBidRequests, bidderRequest) {
-    const currency = config.getConfig('currency');
+    const currencyObj = config.getConfig('currency');
+    const currency = (currencyObj && currencyObj.adServerCurrency) || 'USD';
     const bidIds = {};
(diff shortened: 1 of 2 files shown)


samuelhorwitz:
Alright, try it now. Remember, you have to use a mobile browser user agent or else it won't work, but otherwise it's up and running.

samuelhorwitz:
Alright I fixed it with https://github.com/prebid/Prebid.js/pull/2958/commits/51a12187604b7e45822c951718f0131c2820e23f

This fixes it at the point where I apparently was mishandling it (in the `buildRequests` function where it will trickle through the rest of the code after pulling from the config). Let me know if that solution makes sense to you.

jsnellbaker:
@samuelhorwitz Thanks for the updates; they appear to be working fine now.

PR #2963, opened 2018-08-10
Div resize in case of safeframe

jaiminpanchal27:
## Type of change
- [x] Bugfix

## Description of change
Resize dfp container div to match the creative size in case of safeframe.


jaiminpanchal27:
Duplicate fix. It is fixed with #2947

PR #2966, opened 2018-08-12
Add a Prebid label to console logs

andyblackwell:
## Type of change

- [ ] Bugfix
- [x] Feature
- [ ] New bidder adapter   
- [ ] Build related changes
- [ ] CI related changes
- [ ] Other

## Description of change

Prefix prebid console logs with a label. Easier to single out prebid logs at a glance, and makes it simple to filter down to just prebid logs.
**notes**:
\- I removed the current param count limitations for a few of the methods (logMessage, logInfo, logWarn), so they can behave like the native methods
\- because of this, I removed current formatting that injects the text ` : params : ` between the current params 1(`msg`) and 2(`args `) output
\- label color is Prebid blue `#3b88c3`, taken from prebid.org

## Other information
prebid_console_log_label

### DIFF
--- a/src/utils.js
+++ b/src/utils.js
@@ -14,13 +14,11 @@ var tNumb = 'Number';
 var tObject = 'Object';
 var tBoolean = 'Boolean';
 var toString = Object.prototype.toString;
-let infoLogger = null;
-let warnLogger = null;
-try {
-  infoLogger = console.info.bind(window.console);
-  warnLogger = console.warn.bind(window.console);
-} catch (e) {
-}
+let consoleExists = Boolean(window.console);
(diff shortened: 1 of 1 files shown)


andyblackwell:
Looks like a false positive circleci failure:
```
1) should have unique slot name parameter
     Yieldbot Adapter Unit Tests buildRequests
     request is undefined
@webpack:///test/spec/modules/yieldbotBidAdapter_spec.js:945:6 <- test/test_index.js:138495:11
```
Doesn't seem like I have the ability to re-run the test, so will wait for some help there.
_(Using this PR as my intro to contributing)_

PR #2976, opened 2018-08-14
AdForm Adapter - setting alias code in received bids

Saigredan:
## Type of change
- [ ] Bugfix

## Description of change
When setting alias of AdForm adapter, requested bid has proper bidderCode of given alias, which is correct. However, returning bids are still named as source adapter, because of hardcoded string constant 'adform'. This change allows setting proper alias name into returning bids which is important i.e. for DFP statistics.

### DIFF
--- a/modules/adformBidAdapter.js
+++ b/modules/adformBidAdapter.js
@@ -15,6 +15,7 @@ export const spec = {
     var request = [];
     var globalParams = [ [ 'adxDomain', 'adx.adform.net' ], [ 'fd', 1 ], [ 'url', null ], [ 'tid', null ] ];
     var bids = JSON.parse(JSON.stringify(validBidRequests));
+    var bidder = (bids[0] && bids[0].bidder) || BIDDER_CODE
     for (i = 0, l = bids.length; i < l; i++) {
       bid = bids[i];
       if ((bid.params.priceType === 'net') || (bid.params.pt === 'net')) {
@@ -60,7 +61,7 @@ export const spec = {
       url: request.join('&'),
       bids: validBidRequests,
       netRevenue: netRevenue,
-      bidder: 'adform',
+      bidder,
       gdpr: gdprObject
     };
 


Saigredan:
Looks like circleci fails from other reasons..

jsnellbaker:
@Saigredan You need to rebase off of master to import the newer config files introduced for circleci (and to remove the travis config files).  Can you give this a try?

Saigredan:
@jsnellbaker sure

Saigredan:
@jsnellbaker done

modules/adformBidAdapter.js
```diff
@@ -60,7 +61,7 @@ export const spec = {
       url: request.join('&'),
       bids: validBidRequests,
       netRevenue: netRevenue,
-      bidder: 'adform',
+      bidder,
       gdpr: gdprObject
```
jsnellbaker: In regards to this return object, is this setup correctly for your server to read all the data points?  

While I was testing the adapter for these changes and saw the request going out in the browser's Network tab, it seemed like it only carried the information that was stored in the `url` param.  The other values for `bidder`, `netRevenue`, etc seemed absent.  So I wanted to confirm with you if you were actually getting everything you were expecting when the requests are sent.

As a note - when looking at the code we have in the `bidderFactory` to execute GET requests, additional data points (like what you have here) would normally be stored under a `data` object so that they are parsed into additional query params.

jsnellbaker: @Saigredan Can you clarify your final point in the above?  I'm not fully understanding.

If you want to use these additional params and have them included in your request, then I would recommend refactor this code to either:
- include the values in the `url` param (it looks like you may already be doing this for the `netRevenue` param?)
- try to reorganize the values under the `data` object I mentioned earlier.  If you do this, I would suggest to double-check the naming convention that is automatically applied is something you want/expect.

If the above will take time/effort to do, we could split this from the PR and just merge the alias changes for now and take care of the other rewrites in a separate PR.

Saigredan: @jsnellbaker I believe that refactor would be helpful for better understanding and sticking into proper convention. However this change is not that urgent as final result is fine so far and in addition it may cost more time. Therefore I suggest new PR and merging this alias change.

jsnellbaker: Okay, we'll table those additional changes.  Merging this in now.

PR #2990, opened 2018-08-17
Fix getHighestCpmBids function

### LINKED ISSUE #2959: [Bug] getHighestCpmBids returns rendered bids
## Type of issue
Bug

## Description

This is similar to #2539. With the PrebidJS 1.x updates, the manner in which returned bids are saved into an internal 'pool' complicates the way a winning bid is retrieved. It appears that bids from prior auctions-- even ones that are rendered-- are not removed from the pool.

## Steps to reproduce

Running additional auctions for a given ad unit, you will sometimes receive a winning bid that was previously rendered for the ad unit when calling getHighestCpmBids.

## Test page

Test page exhibiting the problem: https://jsfiddle.net/bradchoate/gvkpz6wy/

Open the console and click the button to refresh the ad a few times. Any bid returned that was already returned via a prior call to getHighestCpmBids will echo in the console.

### Expected results

getHighestCpmBids should never return a bid that has been rendered.

### Actual results

getHighestCpmBids can return a bid that was already rendered.

## Platform details

Observed on macOS, but not specific to that platform.

## Other information

Related to #2539

jaiminpanchal27:
## Type of change
- [x] Bugfix

## Description of change
Fixes #2959

### DIFF
--- a/src/prebid.js
+++ b/src/prebid.js
@@ -1,13 +1,13 @@
 /** @module pbjs */
 
 import { getGlobal } from './prebidGlobal';
-import { flatten, uniques, isGptPubadsDefined, adUnitsFilter, removeRequestId, getLatestHighestCpmBid } from './utils';
+import { flatten, uniques, isGptPubadsDefined, adUnitsFilter, removeRequestId } from './utils';
 import { listenMessagesFromCreative } from './secureCreatives';
 import { userSync } from 'src/userSync.js';
 import { loadScript } from './adloader';
 import { config } from './config';
 import { auctionManager } from './auctionManager';
-import { targeting, getHighestCpmBidsFromBidPool, RENDERED, BID_TARGETING_SET } from './targeting';
+import { targeting, RENDERED, BID_TARGETING_SET } from './targeting';
 import { createHook } from 'src/hook';
 import { sessionLoader } from 'src/debugging';
 import includes from 'core-js/library/fn/array/includes';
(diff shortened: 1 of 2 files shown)


Slind14:
@mkendall07 Please revert and implement under new method. This API point behavior changed drastically without a breaking version change.

mkendall07:
why is it a breaking change? You shouldn't be using old bids. Maybe I'm missing something.

Slind14:
before it returned the highest bids, now it is only returning non rendered bids. This is a total change of a Core API Point which breaks systems/extensions that use it.
I can understand that it was intended differently unfortunately there are systems build upon it. Maybe you can add a parameter to it to filter it based on the bid status, or let the user filter out the rendered ads, its a simple array filter call while getting the rendered ads is impossible.

bradchoate:
Prebid version 0.x did not include previously rendered bids when using `getHighestCpmBids()`, at least not following a subsequent auction. Version 1 did, and to us, that was a breaking API change. Looking at the version 0 code (src/prebid.js), all bids are saved to a `_bidsReceived` array. The `requestBids()` call issues a `clearPlacements()` which removes all bids for ad units that are up for auction. So once the second auction is complete, `getHighestCpmBids()` would only yield winning bids from the auction that just ran.

I don't know how @Slind14 is utilizing this function, but there was a definite API change in Prebid version 1 for it. Granted, the specification for what it does is lacking, but reverting this change means we cannot upgrade past Prebid 1.23.0 until some alternative that excludes rendered bids is available.

Slind14:
hmm, 0.34.8 includes all the bids including the rendered ones for me:
image
image

Slind14:
@bradchoate you can just use: `pbjs.getHighestCpmBids().filter(bid => {bid.status === 'rendered'});`

bradchoate:
Right. That would be the case after each auction and render step. So...

1. auction runs
2. auction callback fires
3. winning ads render
4. time passes
5. repeat cycle for refreshing ad slots

For us, in step 2, we're using `getHighestCpmBids()` and were getting the winners without any used bids under Prebid JS v0. Remember, for Prebid JS v0, step 1 also clears the array of bids received. When v1 shipped, that broke in the context of a refreshing ad unit, since it would include the highest bid for that unit, even those that had rendered already.

I suspect you're looking at the results following step 3, so yes; you'd see 'used' bids, there since another auction hasn't run yet.

Slind14:
There is `pbjs.getAllPrebidWinningBids();` which returns bids that won in the header but did not win on the adserver.

Issue #3010, opened 2018-08-23
TypeError: p is null in secureCreatives.js


iseriousnoob:
## Description
Our logging is catching exceptions. This is really, really rare.

## secureCreatives.js

```js
function receiveMessage(ev) {
  var key = ev.message ? 'message' : 'data';
  var data = {};
  try {
    data = JSON.parse(ev[key]);
  } catch (e) {
    return;
  }

  if (data.adId) {
    const adObject = find(auctionManager.getBidsReceived(), function (bid) {
      return bid.adId === data.adId;
    });

    if (data.message === 'Prebid Request') {
      sendAdToCreative(adObject, data.adServerDomain, ev.source);

      // save winning bids
      auctionManager.addWinningBid(adObject);

      events.emit(BID_WON, adObject);
    }

    // handle this script from native template in an ad server
    // window.parent.postMessage(JSON.stringify({
    //   message: 'Prebid Native',
    //   adId: '%%PATTERN:hb_adid%%'
    // }), '*');
    if (data.message === 'Prebid Native') {
      fireNativeTrackers(data, adObject);
      auctionManager.addWinningBid(adObject);
      events.emit(BID_WON, adObject);
    }
  }
}
```

I think inside the `if (data.message === 'Prebid Request')`-block data is being overwritten.
Resulting in data being null here `if (data.message === 'Prebid Native')`

## Log

[https://pastebin.com/raw/kFZaVCWW](https://pastebin.com/raw/kFZaVCWW)

## Unminified-Minified Code (cropped)

```js
p = {};
try {
    p = JSON.parse(e[g])
} catch (e) {
    return
}
if (p.adId) {

    var m = (0, S.default)(h.auctionManager.getBidsReceived(), function(e) {
        return e.adId === p.adId
    });
    "Prebid Request" === p.message && (t = m, n = p.adServerDomain, r = e.source, i = t.adId, o = t.ad, a = t.adUrl, d = t.width, s = t.height, i && (c = (u = t).adUnitCode, f = u.width, l = u.height, ["div", "iframe"].forEach(function(e) {
        var t, n = (t = e, document.getElementById((0, S.default)(window.googletag.pubads().getSlots().filter((0, y.isSlotMatchingAdUnitCode)(c)), function(e) {
            return e
        }).getSlotElementId()).querySelector(t)).style;
        n.width = f, n.height = l
    }), r.postMessage(JSON.stringify({
        message: "Prebid Response",
        ad: o,
[...]

PR #3012, opened 2018-08-23
Update secureCreatives.js (#3010)

iseriousnoob:
https://github.com/prebid/Prebid.js/issues/3010

## Type of change
- [x] Bugfix

## Description of change
I think inside the `if (data.message === 'Prebid Request')`-block data is being overwritten.
Resulting in data being null here `if (data.message === 'Prebid Native')`


jaiminpanchal27:
Waiting for answer on #3010

PR #3014, opened 2018-08-23
Update appnexus md file with working examples

### LINKED ISSUE #2995: Appnexus examples in markdown file do not return test creatives.
## Type of issue
bug in .md file

## Description
According to http://prebid.org/dev-docs/bidder-adaptor.html#required-files:
modules/exampleBidAdapter.md is required to provide:
* A test ad unit that will consistently return test creatives.

The example in https://github.com/prebid/Prebid.js/blob/master/modules/appnexusBidAdapter.md consistently returns 200 {"error":"Publisher mismatch on multitag auction (915002 != 838066)"}

## Steps to reproduce
Create a simple test using the example provided at https://github.com/prebid/Prebid.js/blob/master/modules/appnexusBidAdapter.md
including all given examples as indicated in the example code.
However, even when used individually, no test creatives are returned; just empty { nobid: true } responses.

### Expected results
A 200 response with a test creative for each example mediaType

### Actual results
The response is consistently 200 {"error":"Publisher mismatch on multitag auction (915002 != 838066)"}

## Platform details
master branch as of 2018-08-20

#### Notes
Only the "banner-div" and "video-outstream" codes can coexist; any other combination results in a similar 200 error message about publisher mismatch.
*Even when used individually, no test creatives are returned; just empty { nobid: true } responses.*

jsnellbaker:
## Type of change

- [x] Bugfix
- [x] Other

## Description of change
Fixes #2995

### DIFF
--- a/modules/appnexusBidAdapter.md
+++ b/modules/appnexusBidAdapter.md
@@ -22,58 +22,59 @@ var adUnits = [
         banner: {
           sizes: [[300, 250], [300,600]]
         }
-      }
+      },
       bids: [{
          bidder: 'appnexus',
          params: {
-           placementId: '10433394'
+           placementId: 13144370
          }
        }]
    },
    // Native adUnit
    {
       code: 'native-div',
-      sizes: [[300, 250], [300,600]],
+      sizes: [[1, 1]],
       mediaTypes: {
         native: {
           title: {
-            required: true,
-            len: 80
-          },
-          body: {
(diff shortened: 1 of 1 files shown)


jarrettabello:
I would suggest using strings for the placementId examples as that is what your bidder param instructions specify. http://prebid.org/dev-docs/bidders.html#appnexus

PR #3016, opened 2018-08-23
SOR-233 Sortable bid adapter handles extra information in server response

EdmonL:
## Type of change

- [ x] Other

## Description of change
Sortable bid adapter allows users to access OpenRTB bidder-specific extension in server response.


EdmonL:
Sorry, this is supposed to be internal patch for now.

Issue #2640, opened 2018-05-29
New prebid version doesn't show all bids in the console


GiovanniPascoli:
## Type of issue
Request

## Description
Upgraded to Prebid v1.12.0 the snippet code to see all bids in the console, now show only winning or looser bids and not all like preceded version (which was also showing the no bid available responses).
Is there a way to restore the previous behavior, even by using undocumented APIs?

## Test page
None
## Platform details
Prebid v1.12.0 Chrome, ecc.

## Other information
I follow steps on the "See all bids in the console" section on 
http://prebid.org/dev-docs/toubleshooting-tips.html

jnsprngs:
Seconding this issue.

I am testing `v1.12.0-pre` and not seeing any bids with a `statusMessage : 'Bid returned empty or error response'` instances. Instead, it looks like null bid responses are excluded from `pbjs.getBidResponses()`.

mercuryyy:
We all second this issue, but this is within prebid new core. @prebid Team mentioned in another thread they do not have time to address this and its open if anybody want to submit a pull.

I think getBidResponses() should return all bids, so we can see who timed out and who was within the timeout but returned no bid.

mkendall07:
yes open for anyone to submit a pull request but I don't think the work is trivial.

YOzaz:
This is required imho.
Google Analytics module now fails to identify which response has failed or timed out.

dejanstrbac:
It is not really prebid that filters the zero cpm bids that previously used to come through as "empty or error response".

Prebid now tags all bids coming from the adapters as "available", even if cpm=0. It is adapters that are also now ignoring own 0 cpms.

One way to get similar data now could be by listening to bidderDone event. You can then deduct a zero cpm bid. For timeouts, you could use the same event (well, the lack of it) or the bidderTimeout event which is not reliable, but it will be impossible to get the information of the final TTR as in Prebid < 1.0.

Tigerroad:
Is this also related to Google Analytics labels not being set for EventAction: Timeouts. I was just about to open a new Issue but it seems related?

andyblackwell:
Here's a start at it. It's working for my needs, but I haven't fully tested. Maybe it could point someone with enough time to do a proper PR in the right direction?
https://gist.github.com/andyblackwell/8815a2b721ad82fbe450347fa221471e
diff with v1.14.0: https://www.diffchecker.com/FtVTQlet

andyblackwell:
[...]

PR #3022, opened 2018-08-24
#2640 all bids in console

jrosendahl:
## Type of change

- [ ] Bugfix
- [x] Feature
- [ ] New bidder adapter   
- [ ] Build related changes
- [ ] CI related changes
- [ ] Other

## Description of change
Addressing #2640 
Made changes so that all bids are shown in console when pbjs.getBidResponses() is called.

## Other information
#2640


src/adapters/bidderFactory.js
```diff
@@ -1,15 +1,30 @@
 import Adapter from 'src/adapter';
 import adaptermanager from 'src/adaptermanager';
-import { config } from 'src/config';
+import {
```
mike-chowla: Please revert all non-substantive formatting changes

src/adapters/bidderFactory.js
```diff
@@ -1,15 +1,30 @@
...
+  config
```
mike-chowla: Please revert all non-substantive formatting changes

src/adapters/bidderFactory.js
```diff
@@ -1,15 +1,30 @@
...
+} from 'src/config';
 import bidfactory from 'src/bidfactory';
-import { userSync } from 'src/userSync';
-import { nativeBidIsValid } from 'src/native';
-import { isValidVideoBid } from 'src/video';
+import {
+  userSync
+} from 'src/userSync';
+import {
```
mike-chowla: Please revert all non-substantive formatting changes

src/adapters/bidderFactory.js
```diff
@@ -1,15 +1,30 @@
...
+} from 'src/native';
+import {
+  isValidVideoBid
+} from 'src/video';
 import CONSTANTS from 'src/constants.json';
 import events from 'src/events';
 import includes from 'core-js/library/fn/array/includes';
 
-import { logWarn, logError, parseQueryStringParameters, delayExecution, parseSizesInput, getBidderRequest } from 'src/utils';
+import {
```
mike-chowla: Please revert all non-substantive formatting changes

src/adapters/bidderFactory.js
```diff
@@ -1,15 +1,30 @@
...
+  logWarn,
+  logError,
+  parseQueryStringParameters,
+  delayExecution,
```
mike-chowla: Please revert all non-substantive formatting changes

src/adapters/bidderFactory.js
```diff
@@ -1,15 +1,30 @@
...
+  parseSizesInput,
+  getBidderRequest
```
mike-chowla: Please revert all non-substantive formatting changes

src/adapters/bidderFactory.js
```diff
@@ -131,9 +147,8 @@ const COMMON_BID_RESPONSE_KEYS = ['requestId', 'cpm', 'ttl', 'creativeId', 'netR
  * @param {BidderSpec} spec An object containing the bare-bones functions we need to make a Bidder.
  */
 export function registerBidder(spec) {
-  const mediaTypes = Array.isArray(spec.supportedMediaTypes)
-    ? { supportedMediaTypes: spec.supportedMediaTypes }
-    : undefined;
+  const mediaTypes = Array.isArray(spec.supportedMediaTypes) ? {supportedMediaTypes: spec.supportedMediaTypes} : undefined;
```
mike-chowla: Please revert all non-substantive formatting changes

src/adapters/bidderFactory.js
```diff
@@ -143,7 +158,9 @@ export function registerBidder(spec) {
   if (Array.isArray(spec.aliases)) {
     spec.aliases.forEach(alias => {
       adaptermanager.aliasRegistry[alias] = spec.code;
-      putBidder(Object.assign({}, spec, { code: alias }));
+      putBidder(Object.assign({}, spec, {
```
mike-chowla: Please revert all non-substantive formatting changes

src/adapters/bidderFactory.js
```diff
@@ -156,16 +173,17 @@ export function registerBidder(spec) {
  */
 export function newBidder(spec) {
   return Object.assign(new Adapter(spec.code), {
-    getSpec: function() {
+    getSpec: function () {
```
mike-chowla: Please revert all non-substantive formatting changes

src/adapters/bidderFactory.js
```diff
@@ -156,16 +173,17 @@ export function registerBidder(spec) {
...
       return Object.freeze(spec);
     },
     registerSyncs,
-    callBids: function(bidderRequest, addBidResponse, done, ajax) {
+    callBids: function (bidderRequest, addBidResponse, done, ajax) {
```
mike-chowla: Please revert all non-substantive formatting changes

src/adapters/bidderFactory.js
```diff
@@ -156,16 +173,17 @@ export function registerBidder(spec) {
...
       if (!Array.isArray(bidderRequest.bids)) {
         return;
       }
 
       const adUnitCodesHandled = {};
+
```
mike-chowla: Please revert all non-substantive formatting changes

src/adapters/bidderFactory.js
```diff
@@ -176,8 +194,30 @@ export function newBidder(spec) {
       // After all the responses have come back, call done() and
       // register any required usersync pixels.
       const responses = [];
+
       function afterAllResponses(bids) {
         const bidsArray = bids ? (bids[0] ? bids : [bids]) : [];
+        // bidderRequest.end = timestamp();
+        let bidIds = bidsArray.map(bid => {
+          if (bid.requestId) {
+            return bid.requestId.toString();
```
mike-chowla: Why are you converting bid ids to strings?  Seems inefficient and unecessary

src/adapters/bidderFactory.js
```diff
@@ -176,8 +194,30 @@ export function newBidder(spec) {
...
+          }
+          return ' '
+        });
+
+        let missingBids = bidderRequest.bids.filter(bid => {
+          if (bid.bidId) {
+            return bidIds.indexOf(bid.bidId.toString()) === -1;
```
mike-chowla: Why are you converting bid ids to strings? Seems inefficient and unnecessary

src/adapters/bidderFactory.js
```diff
@@ -251,8 +294,7 @@ export function newBidder(spec) {
             break;
           case 'POST':
             ajax(
-              request.url,
-              {
+              request.url, {
```
mike-chowla: Please revert all non-substantive formatting changes

src/adapters/bidderFactory.js
```diff
@@ -370,7 +412,7 @@ function validBidSize(adUnitCode, bid, bidRequests) {
   // if a banner impression has one valid size, we assign that size to any bid
   // response that does not explicitly set width or height
   if (parsedSizes.length === 1) {
-    const [ width, height ] = parsedSizes[0].split('x');
+    const [width, height] = parsedSizes[0].split('x');
```
mike-chowla: Please revert all non-substantive formatting changes

src/adapters/bidderFactory.js
```diff
@@ -210,9 +249,14 @@ export function newBidder(spec) {
         }
       });
 
+      if (validBidRequests.length === 0) {
```
mike-chowla: I don't understand the logic why we only want to add no bids when there are no valid bids.  What's the thinking here?

src/adapters/bidderFactory.js
jrosendahl: Existing unit tests do not consistently use strings or numeric values for id's. I was unsure if this was an issue with the tests or in place to make sure ids can be provided either as a string or int.  In any case this is safer.

I could write this using a double equal, but then the cast would be implicit instead of explicit.

src/adapters/bidderFactory.js
jrosendahl: So if there is a bidderRequest.bids, but none of them are valid, this logs those bids as failed bids.  This is consistent with what would currently happen if a subset of the bids are invalid.  

If we want to only have no bids for valid requests we will have to use validBidRequests instead of bidderRequest.bids when looking for missing bids to maintain consistency.

Let me know what your preference is.  I am fine with it either way.

src/adapters/bidderFactory.js
jrosendahl: see above

src/adapters/bidderFactory.js
```diff
@@ -176,8 +178,30 @@ export function newBidder(spec) {
       // After all the responses have come back, call done() and
       // register any required usersync pixels.
       const responses = [];
+
       function afterAllResponses(bids) {
```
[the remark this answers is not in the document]
jrosendahl: You are totally right on both counts, your gist was the start of this pr.  And the accumulator is probably necessary.  Thanks

src/adapters/bidderFactory.js
```diff
@@ -176,10 +177,31 @@ export function newBidder(spec) {
...
+            return bidIds.indexOf(bid.bidId.toString()) === -1;
+          }
+          return false
+        });
 
-        const videoBid = bidsArray.some(bid => bid.mediaType === 'video');
+        missingBids.forEach(bid => {
+          const bidRequest = bidRequestMap[bid.bidId];
+          bid.requestId = bid.bidId;
+          let prebidNoBid = Object.assign(bidfactory.createBid(CONSTANTS.STATUS.NO_BID, bidRequest), bid);
```
jaiminpanchal27: With Prebid 1.0 we are not adding NO_BID to auction. 
To close aucton in legacy we used to match bidsRequested and bidsReceived and hence were adding a NO_BID.

Not needed here.

andyblackwell: This is one of the main pieces to this ticket, so is definitely needed here. It’s adding back in “no bids” for all the missing analytics/data caused by the oversight of removing them in 1.*

mkendall07:
Since this is a API change it needs to go into a 2.x release.

mike-chowla:
Given the feedback, I'm thinking the only way to resolve the debug snippet not showing zero bids in Prebid 1.x is to modify the debug snippet rather the change the core

andyblackwell:
To be honest it’s not just this “debug” call, the removal of these no bids, and even the introduced ajax timeouts (separate topic) in 1.* has resulted in quite a loss of important analytics data. I’ve personally already patched these issues, it was a roadblock for us on the upgrade and we were not ok with losing that data.

Without changing core, I’m not really sure it will cover all the cases though?

PR #3024, opened 2018-08-27
Fix for early auction close with video + done cb cleanup

jaiminpanchal27:
## Type of change

## Description of change
Cleaning up done callback which closes the auction. Simplifying the design.


src/auction.js
```diff
@@ -155,29 +155,21 @@ export function newAuction({adUnits, adUnitCodes, callback, cbTimeout, labels})
...
-
-      // this is done for cache-enabled video bids in tryAddVideoBid, after the cache is stored
-      request.doneCbCallCount += 1;
-      bidsBackAll();
-    }, 1);
+  function done(bidderCount) {
+    let doneCalled = 0;
+    return function() {
+      doneCalled++;
+      if (doneCalled === bidderCount) {
```
mkendall07: are we concerned that some bidders will call multiple `done` callbacks causing the auction to close early?

src/auction.js
```diff
@@ -301,6 +296,51 @@ export function newAuction({adUnits, adUnitCodes, callback, cbTimeout, labels})
...
+
+    if (bidResponse.mediaType === 'video') {
+      tryAddVideoBid(auctionInstance, bidResponse, bidRequest, afterBidAdded);
+    } else {
+      addBidToAuction(auctionInstance, bidResponse);
+      afterBidAdded();
+    }
+  }
+
+  function adapterDone() {
```
mkendall07: what do we need adapterDone for?

mkendall07:
Honestly was pretty confused reading the code. Let's walk through this a bit.

PR #3033, opened 2018-08-30
Fix utils problem with convertTypes

NoodlesNZ:
## Type of change
- [x] Bugfix
- [ ] Feature
- [ ] New bidder adapter 
- [ ] Build related changes
- [ ] CI related changes
- [ ] Other

## Description of change
Fix `ReferenceError: utils is not defined` when running pulsepoint adapter. Introduced with #2844 (https://github.com/prebid/Prebid.js/blob/4682ae39963d09e38179280a7edbce92bb7b4114/modules/pulsepointBidAdapter.js#L71)

### DIFF
--- a/modules/pulsepointBidAdapter.js
+++ b/modules/pulsepointBidAdapter.js
@@ -1,5 +1,5 @@
 /* eslint dot-notation:0, quote-props:0 */
-import {logError, getTopWindowLocation} from 'src/utils';
+import * as utils from 'src/utils';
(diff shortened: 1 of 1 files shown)


Slind14:
Since you asked for feedback on slack. Personally I would just add convertTypes to the import instead of the wildcard import. To keep the original syntax and changes to a minimum.

NoodlesNZ:
Thanks, it could have gone either way. It looked like a lot of other adapters were importing all from utils, so just wanting to follow the pattern.

PR #3051, opened 2018-09-05
Fixed default values in bidResponse and added changes to the description

BifocalTick:
## Type of change
- [x] Bugfix

## Description of change
Fixed default bidResponse's values in interpretResponse method and added changes to the description.

khaylov@betweenx.com


BifocalTick:
We don't support 'placementId' parameter, my mistake, i removed from tests and doc.

Issue #2938, opened 2018-08-06
`utils.getTopWindowLocation` not supported on firefox browsers?


rjvelicaria:
## Type of issue

Question, but maybe an oversight/bug as well.

## Description

On Firefox browsers [tested on quantum and v48], calling `utils.getTopWindowLocation` retrieves `about:blank`.  
This fails to get the real top window location for several reasons:
* Firefox doesn't support location.ancestorOrigins
* Mozilla's [concept of the referrer](https://developer.mozilla.org/en-US/docs/Web/API/Document/referrer#Notes) differs from how chromium implements it.

Is there another utils function we can use to get the top window location to support firefox browsers?  Or if not, can we support firefox browsers on the `utils.getTopWindowLocation` fn.

## Steps to reproduce
any basic prebid setup in an iframe + firefox browser

### Expected results
```javascript
utils.getTopWindowLocation()
>> localhost
``` 
(or what your dev env host is)

### Actual results
```javascript
utils.getTopWindowLocation()
>> about:blank
``` 
(or what the iframe src is set)

## Platform details
Firefox Quantum, or any firefox v48 onwards

mkendall07:
Seems like a bug. I'd review a PR for a fix. thanks for the report.

thewizarodofoz:
@rjvelicaria I tried reproducing this by loading the [hello world](https://github.com/prebid/Prebid.js/blob/master/integrationExamples/gpt/hello_world.html) example in an iFrame and logging the output of `utils.getTopWindowLocation()`.

My findings are:
1. The output is the same in Chrome 68 and in Firefox 61.
2. The output is similar to a [`Location`](https://developer.mozilla.org/en-US/docs/Web/API/Location) object and not a string.

Can you maybe set up a full example to reproduce this?

rjvelicaria:
Hi @thewizarodofoz,
I'll try and get a page for it.
I was using a v1.18.0 of prebid before but I'll try both versions to see if there's any changes.

rjvelicaria:
codepen: https://codepen.io/anon/pen/MqEQrN?editors=1000

how it behaves on chrome v68:
![screen shot 2018-09-06 at 11 33 05 am](https://user-images.githubusercontent.com/25090246/45177915-20b3d780-b1c9-11e8-85e6-32389c541e1e.png)

how it behaves on firefore v61.0.2:
* for firefox, code sometimes works if it loads on the first time.  Try loading the same codepen again on another tab and check the values reported.
[...]

PR #3060, opened 2018-09-07
getTopWindowLocation was not returning the same value always

thewizarodofoz:
there is difference between firefox and chrome described in #2938.

In Chrome, the location is returned from [`getAncestorOrigins`](https://github.com/prebid/Prebid.js/blob/master/src/utils.js#L227), but since that property is not available to Firefox, in Firefox the path leads to `getTopFrameReferrer` and in line [213](https://github.com/prebid/Prebid.js/blob/master/src/utils.js#L213) where `currentWindow` is the top frame, `currentWindow.document.referrer` is undefined.

According to a [discussion in Bugzilla](https://bugzilla.mozilla.org/show_bug.cgi?id=1085214#c0):
> location.ancestorOrigins should contain a read-only array of **hostnames** for all the parent frames.

And according to [MDN](https://developer.mozilla.org/en-US/docs/Web/API/Document/referrer), `Document.referrer` is a different thing:
> Returns the **URI** of the page that linked to this page.

So I think the problem starts with [`getTopWindowLocation`](https://github.com/prebid/Prebid.js/blob/master/src/utils.js#L196) interchanging hostname with referrer.

The closest I could find to `ancestorOrigins` in Firefox is `location.origin`.

## Type of change

- [x] Bugfix
- [ ] Feature
- [ ] New bidder adapter   
- [ ] Build related changes
- [ ] CI related changes
- [ ] Other

### DIFF
--- a/src/utils.js
+++ b/src/utils.js
@@ -193,29 +193,36 @@ exports.getTopWindowLocation = function() {
   if (exports.inIframe()) {
     let loc;
     try {
-      loc = exports.getAncestorOrigins() || exports.getTopFrameReferrer();
+      loc = exports.getAncestorOrigins() || exports.getTopFrameOrigin();
     } catch (e) {
-      logInfo('could not obtain top window location', e);
+      exports.logInfo('could not obtain top window location', e);
     }
     if (loc) return parse(loc, {'decodeSearchAsString': true});
   }
   return exports.getWindowLocation();
 }
 
+exports.getTopFrame = function () {
+  let frame;
+  do {
+    frame = frame ? frame.parent : window;
+  }
+  while (frame !== window.top);
+
+  return frame;
+};
+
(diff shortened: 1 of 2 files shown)


mkendall07:
We are actually in the process of re-writing some of this functionality to be more consistent. Will hold on this for now until we see how that integrates.

jsnellbaker:
@thewizarodofoz the above mentioned refactoring has been implemented.  Please refer to https://github.com/prebid/Prebid.js/pull/3067 for details on the updates.

I'm going to close this PR at this time since it shouldn't be needed anymore (and we'll be eventually deprecating these old methods in the `utils.js` file - https://github.com/prebid/Prebid.js/issues/3072).  

If you happen to see anything potentially lacking with the new module that the old functions were doing or any inconsistency of the behavior between browsers, let's discuss it in a new issue.

Issue #3065, opened 2018-09-10
Secure Creative resize works only in quirks mode


florevallatmrf:
## Description
When using the secure creative flow, the creative resizing works only when the browser is in quirks mode (ie without a `<!DOCTYPE html>` at the beginning of the document). 
## Steps to reproduce
1. Setup header bidding for an adslot
2. Serve the target Admanager creative as a safeframe, size 1x1
3. Request the creative from prebid using post message
4. The creative is received

## Test page
JSFiddle showing the resize not working: http://jsfiddle.net/florevallatmrf/z89nwvbp/
Unfortunately JSFiddle always adds a `doctype` to the result, so I can't provide a link of it working without it. 

### Expected results
The slot (both containing div and safeframe) should resize.

### Actual results
The slot stays as a 1x1 block. 

## Platform details
Prebid v1.23.0
Tested in Chrome 68 & Safari 11.

## Other information
We changed this in our fork of Prebid, I will open a Pull request on this side too.

muuki88:
@florevallatmrf nice to see people using safe frames. Have encountered any other issues with prebid and safe frames so far?

florevallatmrf:
No, we haven't used them in production yet, for now the resizing has been my biggest pain!

muuki88:
Thanks for sharing @florevallatmrf :hugs: We haven't rolled out any safe frames as well, because we are very uncertain how advertisers react (especially with cookie syncing, etc.).

PR #3066, opened 2018-09-10
Fix the secure creative resizing

florevallatmrf:
## Type of change
- [ x ] Bugfix

## Description of change
Explicitly setting the unit, `px`, when resizing the safeframe and its container div. 

## Other information
Issue #3065

### DIFF
--- a/src/secureCreatives.js
+++ b/src/secureCreatives.js
@@ -72,8 +72,8 @@ function resizeRemoteCreative({ adUnitCode, width, height }) {
   // resize both container div + iframe
   ['div', 'iframe'].forEach(elmType => {
     let elementStyle = getElementByAdUnit(elmType).style;
-    elementStyle.width = width;
-    elementStyle.height = height;
+    elementStyle.width = `${width}px`;
+    elementStyle.height = `${height}px`;
   });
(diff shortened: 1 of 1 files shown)


src/secureCreatives.js
```diff
@@ -72,8 +72,8 @@ function resizeRemoteCreative({ adUnitCode, width, height }) {
   // resize both container div + iframe
   ['div', 'iframe'].forEach(elmType => {
     let elementStyle = getElementByAdUnit(elmType).style;
-    elementStyle.width = width;
-    elementStyle.height = height;
+    elementStyle.width = `${width}px`;
```
mike-chowla: Template Literals aren't supported on Internet Explorer:
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Template_literals

Given the that browser support includes IE 10 and later:
https://github.com/prebid/Prebid.js/blob/master/README.md

This needs to be changed to not use a template literal

florevallatmrf: Oh, I was assuming the build would transform it. Thank you for the info!

mkendall07: @mike-chowla 
we do transform template literals. Thanks for checking!

PR #3067, opened 2018-09-10
Add referer detection module

jaiminpanchal27:
## Type of change
- [x] Feature

## Description of change
This PR adds referer detection module to prebid core. New property `refererInfo` is added to bidderRequest object.
Following details will be available to each bidder
```
{
  referer: <string|detected top url>,
  reachedTop: <bool|whether prebid was able to walk upto top window or not>,
  numIframes: <int| number of iframes>,
  stack: <string|comma separated urls of all origins>,
}
```
All bidders can access this info in buildRequests spec.

We will be deprecating `utils.getTopWindowLocation` and `utils.getTopFrameReferrer` soon.

### DIFF
--- a/src/refererDetection.js
+++ b/src/refererDetection.js
@@ -100,7 +100,7 @@ export function detectReferer(win) {
             location: null,
             isTop: (currentWindow == win.top)
           });
-          logWarn('Error while walking cross domain iframe. Continuing without referrer and location');
+          logWarn('Trying to access cross domain iframe. Continuing without referrer and location');
         }
       } catch (e) {
         acc.push({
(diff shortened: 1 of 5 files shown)


src/refererDetection.js
```diff
@@ -0,0 +1,140 @@
...
+            location: currentWindow.location.href || null,
+            isTop: (currentWindow == win.top)
+          });
+        } catch (e) {
+          acc.push({
+            referrer: null,
+            location: null,
+            isTop: (currentWindow == win.top)
+          });
+          logError('Error while accessing cross domain iframe.', e);
```
jsnellbaker: Does this need to be a `logError`?  Based on my testing in a cross domain testpage in Chrome, the data still gets collected (due to available information in ancestor) but it also displays this error.  

Should it be a warning instead perhaps?

PR #3087, opened 2018-09-13
Rubicon adapter: DecodeUrl if url is encoded

AntoineJac:
Appnexus new module is passing the encoded url and sometimes Publisher are passing encoded url in the referrer parameter. 
This fix is a protection in case a module is update and start passing encode url. Our AE is not decoding and from some test I did, monetisation could be highly impacted.

## Type of change

- [ ] Bugfix
- [ ] Feature
- [ ] New bidder adapter   
- [ ] Build related changes
- [ ] CI related changes
- [ ] Other

## Description of change

```
{
  bidder: '<bidder name>',
  params: {
    // ...
  }
}
```

### DIFF
--- a/modules/rubiconBidAdapter.js
+++ b/modules/rubiconBidAdapter.js
@@ -111,7 +111,7 @@ export const spec = {
       let size = parseSizes(bidRequest);
 
       let data = {
-        page_url: _getPageUrl(bidRequest),
+        page_url: _getPageUrl(bidRequest, bidderRequest),
         resolution: _getScreenResolution(),
         account_id: params.accountId,
         integration: INTEGRATION,
@@ -323,9 +323,13 @@ export const spec = {
       'tk_user_key': params.userId,
       'p_geo.latitude': isNaN(parseFloat(latitude)) ? undefined : parseFloat(latitude).toFixed(4),
       'p_geo.longitude': isNaN(parseFloat(longitude)) ? undefined : parseFloat(longitude).toFixed(4),
-      'tg_fl.eid': bidRequest.code,
-      'rf': _getPageUrl(bidRequest)
+      'tg_fl.eid': bidRequest.code
     };
+
+    if (bidderRequest || bidRequest) {
+      // add 'gdpr' only if 'gdprApplies' is defined
+      data['rf'] = _getPageUrl(bidRequest, bidderRequest);
+    }
 
     if (bidderRequest.gdprConsent) {
       // add 'gdpr' only if 'gdprApplies' is defined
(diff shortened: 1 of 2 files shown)


mkendall07:
@AntoineJac 
you have some test failure.

AntoineJac:
@mkendall07 , thanks I indeed forget to push the new specs.
@bretg , can you please review?

modules/rubiconBidAdapter.js
```diff
@@ -500,14 +500,15 @@ function _getDigiTrustQueryParams() {
...
   let pageUrl = config.getConfig('pageUrl');
   if (bidRequest.params.referrer) {
     pageUrl = bidRequest.params.referrer;
   } else if (!pageUrl) {
-    pageUrl = utils.getTopWindowUrl();
+    pageUrl = bidderRequest.refererInfo.referer;
   }
-  return bidRequest.params.secure ? pageUrl.replace(/^http:/i, 'https:') : pageUrl;
+  let _pageUrl = bidRequest.params.secure ? pageUrl.replace(/^http:/i, 'https:') : pageUrl;
+  return decodeURIComponent(_pageUrl.replace(/\+/g, ' '))
```
harpere: the title of the PR says, "DecodeUrl `if` url is encoded".   Do we need to check first if it's encoded?

mkendall07: question - should we update the referer module to return unencoded referer? I'm sort of on the fence on this but curious to get other's input.

AntoineJac: @harpere , I guess that if the url is decoded, decoding should not provoke any error.  

@mkendall07 , I would indeed recommend returning decoding url as other partner may face some issues.

harpere: @AntoineJac - I think decoding again might only make a difference if there was a `%` in the referrer.   But if we know the module is encoding, we're fine.

@mkendall07 - my impulse is that the module should NOT encoded referer and let the adapter encode it if they need to.   But not a big deal.

bretg:
Which 'Appnexus module' is generating this URL? Is this the new 'referer detection' PR https://github.com/prebid/Prebid.js/pull/3067 ?

Assigning to @mkendall07 to make a call here - we'll abandon this PR if referer detection flips the change. But since that's a breaking change, am guessing a change at this point would be hard.

mkendall07:
@bretg thanks. yes we should not encode the referral in the referer detection module. @jaiminpanchal27 will make the change and you can update this PR as required or just close it.

bretg:
Closing this -- resolved by https://github.com/prebid/Prebid.js/pull/3116

bretg:
re-opening so @AntoineJac can update it to remove the decode part (and update title)

AntoineJac:
@bretg , PR is ready for reviewing. 
Thanks,

PR #3090, opened 2018-09-13
Added user sync support for undertone bid adapter

omerko:
## Type of change

- [x] Feature

## Description of change
Added implementation of getUserSyncs in undertone adapter to allow syncing users.
If iframe sync is allowed will use iframe sync, otherwise, we'll use two sync pixels that is as allowed by default


mkendall07:
you may need to rebase off current master to get circleci to pass.

mike-chowla:
You shouldn't be committing package-lock.json

jsnellbaker:
Closing this PR as requested in #3171

PR #3096, opened 2018-09-14
Fixes #3091 Auction closing prematurely

### LINKED ISSUE #3091: Auction ends prematurely [1.24.0]
## Type of issue
Bug

## Description
After updating from 1.23.0 to 1.24.0 a drop in revenues occurred.

## Steps to reproduce
Upgrade to 1.24.0 and look for the auction's end. If you have some debug, you'll see that auction ends before some valid bids.
Note that reading the releases' page didn't show any new parameter involved.

## Test page
No test page as I've reverted back to 1.23.0 in production.

### Expected results
All bids should arrive before auction ends.

### Actual results
Auction ends in the middle of bids. Most probably Criteo (the fastest bid response) cut all other (but Criteo is also cut by other bidders sometimes).

## Platform details
Compiled on Linux Debian, Node 8.11.4, npm 5.6.0

## Other information
Change to 1.23.0 at 11:45, reverted at 17:00.

Sent bids (for AppNexus) vs returned bids (darker green):
image

Revenues (still only AppNexus):
image

Sent bids for Criteo, less impacted as it's the fastest:
image

jaiminpanchal27:
## Type of change
- [x] Bugfix

## Description of change
Fixes #3091 Auction closing prematurely

### DIFF
--- a/src/auction.js
+++ b/src/auction.js
@@ -295,7 +295,7 @@ export function auctionCallbacks(auctionDone, auctionInstance) {
   let outstandingBidsAdded = 0;
   let allAdapterCalledDone = false;
 
-  let areAllAdapterDone = delayExecution(() => {
+  let onAllAdapterDone = delayExecution(() => {
     allAdapterCalledDone = true;
   }, auctionInstance.getBidRequests().length);
 
@@ -323,8 +323,8 @@ export function auctionCallbacks(auctionDone, auctionInstance) {
   }
 
   function adapterDone() {
-    areAllAdapterDone();
-    if (allAdapterCalledDone && (outstandingBidsAdded === 0)) {
+    onAllAdapterDone();
+    if (allAdapterCalledDone && outstandingBidsAdded === 0) {
       auctionDone();
     }
   }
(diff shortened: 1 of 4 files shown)


src/auction.js
```diff
@@ -303,11 +293,15 @@ export const addBidResponse = createHook('asyncSeries', function(adUnitCode, bid
 
 export function auctionCallbacks(auctionDone, auctionInstance) {
   let outstandingBidsAdded = 0;
-  let doneCalled = false;
+  let allAdapterCalledDone = false;
+
+  let areAllAdapterDone = delayExecution(() => {
```
mkendall07: nice usage here. One knit (I know it's not part of this PR) but the name of this function is somewhat confusing. I was expecting it to wait for the next turn of the event loop, but it's synchronous.

src/auction.js
```diff
@@ -321,16 +315,16 @@ export function auctionCallbacks(auctionDone, auctionInstance) {
...
       addBidToAuction(auctionInstance, bidResponse);
       afterBidAdded();
     }
   }
 
   function adapterDone() {
-    doneCalled = true;
-    if ((outstandingBidsAdded === 0)) {
+    areAllAdapterDone();
+    if (allAdapterCalledDone && (outstandingBidsAdded === 0)) {
```
mkendall07: double parentheses are not needed here.

PR #3098, opened 2018-09-17
update test command in package.json scripts area

jsnellbaker:
## Type of change
- [x] Build related changes

## Description of change
As a follow-up from the gulp 4 upgrade (#2930), a deprecated gulp command was still present in the `package.json` scripts area.  

I replaced the old `gulp run-scripts` command with the equivalent commands it used to run (`gulp lint` & `gulp test-coverage`).

### DIFF
--- a/package.json
+++ b/package.json
@@ -3,10 +3,6 @@
   "version": "1.25.0-pre",
   "description": "Header Bidding Management Library",
   "main": "src/prebid.js",
-  "scripts": {
-    "test": "gulp lint && gulp test-coverage",
-    "lint": "gulp lint"
-  },
   "repository": {
     "type": "git",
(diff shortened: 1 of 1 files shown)


jsnellbaker:
I wanted to preserve the logic as it was in case anyone was using it.  I can alter it to one of the following:
- remove it outright from the scripts block
- add in `test` command to the sequence
- replace the two commands with the `test` command

Please let me know your thoughts.

mkendall07:
I'm onboard with just removing scripts, or alternatively just having it run `test` gulp task (without any flags).

jsnellbaker:
I just pushed a commit to remove the scripts block from the `package.json` file.

jsnellbaker:
Closing in favor to https://github.com/prebid/Prebid.js/pull/3117

PR #3099, opened 2018-09-17
add user syncing for kargo

samuelhorwitz:
## Type of change

- [ ] Bugfix
- [x] Feature
- [ ] New bidder adapter   
- [ ] Build related changes
- [ ] CI related changes
- [ ] Other

## Description of change

Add user syncing to Kargo

### DIFF
--- a/modules/kargoBidAdapter.js
+++ b/modules/kargoBidAdapter.js
@@ -3,6 +3,8 @@ import {config} from 'src/config';
 import {registerBidder} from 'src/adapters/bidderFactory';
 const BIDDER_CODE = 'kargo';
 const HOST = 'https://krk.kargo.com';
+const SYNC = 'https://crb.kargo.com/api/v1/initsyncrnd/{UUID}?seed={SEED}&idx={INDEX}';
+const SYNC_COUNT = 5;
 export const spec = {
   code: BIDDER_CODE,
   isBidRequestValid: function(bid) {
@@ -53,6 +55,20 @@ export const spec = {
       });
     }
     return bidResponses;
+  },
+  getUserSyncs: function(syncOptions) {
+    const syncs = [];
+    const seed = spec._generateRandomUuid();
+    const clientId = spec._getClientId();
+    if (syncOptions.iframeEnabled && seed && clientId) {
+      for (let i = 0; i < SYNC_COUNT; i++) {
+        syncs.push({
+          type: 'iframe',
+          url: SYNC.replace('{UUID}', clientId).replace('{SEED}', seed).replace('{INDEX}', i)
+        });
+      }
+    }
+    return syncs;
   },
 
   // PRIVATE
@@ -150,6 +166,11 @@ export const spec = {
     };
   },
 
+  _getClientId() {
+    const uid = spec._getUid();
+    return uid.clientId;
+  },
+
   _getAllMetadata() {
     return {
       userIDs: spec._getUserIds(),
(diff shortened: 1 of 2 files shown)


modules/kargoBidAdapter.js
```diff
@@ -53,6 +55,20 @@ export const spec = {
     }
     return bidResponses;
   },
+  getUserSyncs: function(syncOptions) {
+    const syncs = [];
+    const seed = spec._generateRandomUuid();
+    const clientId = spec._getClientId();
+    if (syncOptions.iframeEnabled && seed && clientId) {
+      for (let i = 0; i < SYNC_COUNT; i++) {
```
jsnellbaker: Just to confirm the intended logic here - this setup will always generate 5 sync requests that are largely the same minus the `idx` param which gets incremented by 1 for each subsequent sync request.  Is that the expected behavior?

test/spec/modules/kargoBidAdapter_spec.js
```diff
@@ -356,4 +356,81 @@ describe('kargo adapter tests', function () {
...
+  describe('user sync handler', function() {
+    const clientId = '74c81cbb-7d07-46d9-be9b-68ccb291c949';
+    var uid;
+    var shouldSimulateOutdatedBrowser;
+
+    beforeEach(() => {
+      uid = {};
+      shouldSimulateOutdatedBrowser = false;
+
+      sandbox.stub(crypto, 'getRandomValues').callsFake(function(buf) {
```
jsnellbaker: This currently returns an `undefined` error on the `crypto` object when the unit tests run in IE11 (see the circleci failed job for details).  Can you please take a look and fix the issue?

modules/kargoBidAdapter.js
samuelhorwitz: Yes, we want to attempt 5 syncs (barring a cap set by the configuration) and we want our server to be able to choose what to sync, potentially randomly, using a stateless system that gets passed an instance ID/seed and an incrementer.

test/spec/modules/kargoBidAdapter_spec.js
```diff
@@ -356,4 +356,99 @@ describe('kargo adapter tests', function () {
       expect(resp).to.deep.equal(expectation);
     });
   });
+
+  describe('user sync handler', function() {
+    const clientId = '74c81cbb-7d07-46d9-be9b-68ccb291c949';
+    var shouldSimulateOutdatedBrowser,
+        uid,
+        isActuallyOutdatedBrowser;
```
jsnellbaker: There are some extra spaces before these variables that are causing some lint errors when I run `gulp lint`.  See below for a copy of the error message:

```
/Users/jsnellbaker/git/Prebid.js/test/spec/modules/kargoBidAdapter_spec.js
  394:1  error  Expected indentation of 6 spaces but found 8  indent
  395:1  error  Expected indentation of 6 spaces but found 8  indent
```

Can you make this fix?  

Normally linting errors would be caught by circleci.  So going forward, you may want to look into updating/rebasing your fork with the latest version of master so we can catch these automatically when you submit a PR.

modules/kargoBidAdapter.js
jsnellbaker: Thanks for clarifying.

test/spec/modules/kargoBidAdapter_spec.js
samuelhorwitz: Ah good catch, I'll fix and rebase as well.

jsnellbaker:
@samuelhorwitz Thanks for making the additional changes; LGTM

Issue #3111, opened 2018-09-20
Pubmatic destroys bid sizes


ReinoutStevens:
## Type of issue
Bug

## Description

Since Prebid 1.18 Pubmatic supports multi-size adslots ( https://github.com/prebid/Prebid.js/pull/2887 ). When passing an adslot without '@widthxheight' the sizes of the `mediatypes.banner.sizes` is used. Pubmatic bid adapter destructively modifies this array, messing up other adapters (in our case `ixBidAdapter`).

The offending line is [L266](https://github.com/prebid/Prebid.js/blob/master/modules/pubmaticBidAdapter.js#L266)

When you pass `@widthxheight` in adslot `sizes` is removed on [L120](https://github.com/prebid/Prebid.js/blob/master/modules/pubmaticBidAdapter.js#L120), which may also be an issue of the object is shared across different adapters.

Also note that the first size is ignored for some reason due to the `splice` call. This is probably due to the first size already being taken on [L122](https://github.com/prebid/Prebid.js/blob/master/modules/pubmaticBidAdapter.js#L122).

## Steps to reproduce

Have a pubmatic bid without the size defined in the adslot name.

## Test page
http://jsfiddle.net/2qwfp0zh/

### Expected results
Index bidAdapter should not complain

### Actual results
Index bidAdapter is a sad panda because the sizes are messed up

jsnellbaker:
This particular issue has been fixed at the adapter level; I will open a new issue for a fix at the core level to avoid this issue happening again accidentally with other adatpers.

PR #3112, opened 2018-09-20
Pubmatic Adapter dont modify shared bid data

ReinoutStevens:
## Type of change

- [x] Bugfix
- [ ] Feature
- [ ] New bidder adapter   
- [ ] Build related changes
- [ ] CI related changes
- [ ] Other

## Description of change

Ensure pubmatic adapter does not modify a shared sizes variable

## Other information

https://github.com/prebid/Prebid.js/issues/3111

### DIFF
--- a/modules/pubmaticBidAdapter.js
+++ b/modules/pubmaticBidAdapter.js
@@ -263,7 +263,7 @@ function _createImpressionObject(bid, conf) {
       topframe: utils.inIframe() ? 0 : 1,
     }
     if (utils.isArray(sizes) && sizes.length > 1) {
-      sizes = sizes.splice(1, sizes.length - 1);
+      sizes = sizes.slice(1);
       var format = [];
       sizes.forEach(size => {
         format.push({w: size[0], h: size[1]});


ReinoutStevens:
Note that this still leaves a suspicious looking line https://github.com/prebid/Prebid.js/blob/master/modules/pubmaticBidAdapter.js#L120

mkendall07:
@ReinoutStevens 
Is this causing an issue somewhere?

ReinoutStevens:
@mkendall07 The reported issue associated with this PR has been causing some problems. Atm we have reverted our code and no longer make use of Pubmatic's multi-size support, and specify multiple bids per bids again (as it used to be the case).

mkendall07:
@ReinoutStevens 
ah sorry I missed the linked issue. I think Prebid team should fix this at the core level. Will look into solution as well.

ReinoutStevens:
@mkendall07 no worries, thanks for taking a look at this. 
I agree that an adapter modifying some passed objects should have no influence. I would expect a copy to be passed to each adapter, and not a shared object.
This PR is only a quick fix for a bigger problem, and I'm not sure of the impact of this change on pubmatic's adapter.

jsnellbaker:
Closing this PR in favor of the fix put together by the Pubmatic team.  https://github.com/prebid/Prebid.js/pull/3122

PR #3113, opened 2018-09-20
Small bugfix and cleanup for Prebid Server OpenRTB code

snapwich:
## Type of change
- [x] Bugfix
- [x] Cleanup

## Description of change
Fixed `bid_id` property access error in prebidServer OpenRTB when bid request cannot be found (shouldn't happen usually, but in testing it can).

Changed the bid request map to be a `bid_id` map to be clearer on its intent and made it friendlier for minifcation by removing it from object literal.

Cleaned up some code duplication.  Don't overwrite `bid.source` and `bid.bidderCode` and instead use code in bid factory by passing in bid request stub if request does not exist.

### DIFF
--- a/modules/prebidServerBidAdapter/index.js
+++ b/modules/prebidServerBidAdapter/index.js
@@ -355,10 +355,8 @@ const LEGACY_PROTOCOL = {
  * Protocol spec for OpenRTB endpoint
  * e.g., https://<prebid-server-url>/v1/openrtb2/auction
  */
+let bidIdMap = {};
 const OPEN_RTB_PROTOCOL = {
-
-  bidMap: {},
-
   buildRequest(s2sBidRequest, bidRequests, adUnits) {
     let imps = [];
     let aliases = {};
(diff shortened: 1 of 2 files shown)


mkendall07:
Hey I think I know this guy ;)

snapwich:
I'm back baby!

modules/prebidServerBidAdapter/index.js
```diff
@@ -480,17 +477,22 @@ const OPEN_RTB_PROTOCOL = {
...
+              bidderRequests
+            );
+          }
 
           const cpm = bid.price;
           const status = cpm !== 0 ? STATUS.GOOD : STATUS.NO_BID;
-          let bidObject = bidfactory.createBid(status, bidRequest);
+          let bidObject = bidfactory.createBid(status, bidRequest || {
+            bidder: seatbid.seat,
+            src: TYPE
```
jsnellbaker: I'm not sure if this an issue - but the `src` isn't defined in the `bidRequest` prior to this point.  So when the `bidObject` gets created, the `bidObject.source` is set to `client` and that's carried through the rest of the auction.

Previously it would have been set to `s2s` to reflect it came from pbs via the override.

snapwich: That's a good catch.  It seems to me that the src should be known at that point though as it was already set on the `bidderRequest.src`, just not each of the `bidderRequest.bids` objects.  I updated the pull-request to include adding the `src` to the bids as well as soon as src is known.

PR #3035, opened 2018-08-30
Changes on AdSize + gdpr

phtechno:
- Primary adsize is now the prebid slot adsize
- On bidder.params you can now overwrite the prebid slot size (Not primary size anymore)
- gdpr string will be only attached if "gdprApplies=true"

## Type of change

- [x] Bugfix
- [x] Feature
- [ ] New bidder adapter   
- [x] Code style update (formatting, local variables)
- [ ] Refactoring (no functional changes, no api changes)
- [ ] Build related changes
- [ ] CI related changes
- [ ] Does this change affect user-facing APIs or examples documented on http://prebid.org?
- [ ] Other

## Description of change

- change the required params to set the bid valid
- only attach the gdpr-url-parameter if the "gdprApplies"-parameter in consentModule is true
- use the prebid slot size as primary size
- via adSize in "bidder.params" you can overwrite the prebid slot size for requesting prebid. If yieldlab slot size is different to the real size which you can configure in prebid slot config.

Be sure to test the integration with your adserver using the [Hello World](/integrationExamples/gpt/hello_world.html) sample page.

- contact email of the adapter’s maintainer
- [ ] official adapter submission

For any changes that affect user-facing APIs or example code documented on http://prebid.org, please provide:

- A link to a PR on the docs repo at https://github.com/prebid/prebid.github.io/

PR #3129, opened 2018-09-26
fixes #3128 YieldlabBidAdapter is not using bidRequest.params.adSize

### LINKED ISSUE #3128: YieldlabBidAdapter is not using bidRequest.params.adSize as customsize in bidResponse
## Type of issue
Bug

## Description
The customsize is not used within the bidResponse. Instead it uses the primary size of the array bidRequest.sizes.

## Steps to reproduce
see screenshot below

### Expected results
The bidResponse should use the bidRequest.params.adSize as width and height for the response.

### Actual results
The bidResponse contains the wrong size and doesn't match to the creativeId for Yieldlab deal. The deal doesn't appear.

## Platform details
prebid 1.24.1

## Other information
image

Introduced in #3035

derdeka:
## Type of change
- [x] Bugfix

## Description of change
when bidRequest.params.adSize is available, it has to be used in bidResponse or it will impact ad delivery

## Other information
- fixes issue which was introduced in PR #3035 
- affected versions prebid 1.24.0, 1.24.1, 1.25.0

### DIFF
--- a/modules/yieldlabBidAdapter.js
+++ b/modules/yieldlabBidAdapter.js
@@ -80,8 +80,8 @@ export const spec = {
         const bidResponse = {
           requestId: bidRequest.bidId,
           cpm: matchedBid.price / 100,
-          width: primarysize[0],
-          height: primarysize[1],
+          width: customsize[0],
+          height: customsize[1],
           creativeId: '' + matchedBid.id,
           dealId: matchedBid.pid,
           currency: CURRENCY_CODE,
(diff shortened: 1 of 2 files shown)


modules/yieldlabBidAdapter.js
```diff
@@ -80,8 +80,8 @@ export const spec = {
         const bidResponse = {
           requestId: bidRequest.bidId,
           cpm: matchedBid.price / 100,
-          width: primarysize[0],
-          height: primarysize[1],
+          width: customsize[0],
+          height: customsize[1],
           creativeId: '' + matchedBid.id,
```
harpere: Can you add a test for this change, or for the behavior in general?

derdeka: The issue happens, if there is more than one size available. I've added a testcase for this.

Issue #3130, opened 2018-09-26
adapters can indirectly manipulate shared bidRequest object


jsnellbaker:
## Type of issue
Bug

## Description
The global `bidderRequest` object was being accidentally manipulated by an adapter who spliced the `sizes` array while they were modifying the data for their request.  The splice caused the `sizes` array to drop one of the sizes for other adapters because the change carried through to the original object, which is not ideal.

A temporarily local fix was introduced to the Pubmatic adapter where this issue was originally seen/reported (see https://github.com/prebid/Prebid.js/issues/3111) and this fix was merged into the master branch (1.26.0).  

But we need a higher level fix to address the root of the issue.

## Steps to reproduce
See below

## Test page
http://jsfiddle.net/0kt4cf35/

Note -  To replicate the issue, check out the `1.25.0` tag in prebid.js and run locally.

### Expected results
Global bidder request objects shouldn't be modified by one adapter's code.

### Actual results
Possible that one adapter's code was modifying the global object used by other adapters.

## Platform details
1.25.0

## Other information
Originally found in https://github.com/prebid/Prebid.js/issues/3111

jsnellbaker:
Fixed with the above PR.

PR #3131, opened 2018-09-26
Fix for Issue 3130: passing new copy of adUnits object to every adapter

pm-harshad-mane:
passing new copy of adUnits object to every adapter

## Type of change
- [x] Bugfix

## Description of change
https://github.com/prebid/Prebid.js/issues/3130


pm-harshad-mane:
Hello @jsnellbaker,

Could you please review the change?
I have made a simple change, passing a new copy to every adapter(client and s2s) using utils.deepClone.
All unit test cases are passed without making a change in any unit test case.

pm-harshad-mane:
Hello @snapwich ,

In version 1.25,
Before calling all client-side adapters we are creating a clone of AdUnits object, let's call it newAdUnitsObject and then we are passing this newAdUnitsObject to every adapter(same object to every adapter). If one of the adapters updates the newAdUnitsObject passed to it by mistake, then it affects newAdUnitsObject which is same for all adapters.

In the changes made, 
Before calling all client-side adapters we are creating a clone of AdUnits object, let's call it newAdUnitsObject and then we are passing a clone of newAdUnitsObject to every adapter(new object to every adapter). So if one of the adapters updates the clone of newAdUnitsObject by mistake, then it won't affect the clone of newAdUnitsObject shared with other adapters.

So with new changes, if there is an adapter making a mistake of affecting the common object won't need to fix the code as we will be passing a separate copy to every adapter.

snapwich:
@pm-harshad-mane your fork was out-of-date with master so I've rebased off the latest.  I've also added a test that verifies your fix and protects against regressions.

LGTM

pm-harshad-mane:
Thanks a lot, @snapwich and @jsnellbaker

PR #3133, opened 2018-09-26
Open rtb separation

Pupis:
## Type of change

## Description of change
This commit was made to separate openRTB logic from s2s Prebid to be usable to all adapters. Adform adapter will be updated to support native adds using openRTB protocol.

Scope.FL.Scripts@adform.com


snapwich:
I think the goal of separating out request building for Open RTB is probably a good idea.  But it can't be done like this as server requests coming from the adapter manager are structured differently than client requests.

This probably requires more discussion as it's quite a large architectural change.

Issue #2462, opened 2018-04-27
Native ads not working without adUnit.sizes


jaiminpanchal27:
## Type of issue
Bug

## Description
Test page: https://jsfiddle.net/6r217wza/
Native ads should not require size to be mentioned on adUnit. Above test page will work if you uncomment the adUnit.sizes

Debug mode shows `callBids executed with no bidRequests.  Were they filtered by labels or sizing?`
Reason for filtering is [here](https://github.com/prebid/Prebid.js/blob/master/src/adaptermanager.js#L50l)

snapwich:
@jaiminpanchal27 the size mapping probably needs to be refactored.  When I originally wrote it there was no such thing as `mediaTypes`, so it operates on `adUnit.sizes` alone.  But now I guess it should expect and operate on one or the other?

PR #3134, opened 2018-09-26
Make size mapping mediaType aware

snapwich:
## Type of change
- [x] Bugfix

## Description of change
This makes the Prebid size mapping feature work on mediaTypes objects rather than size arrays.  In other words, it is now mediaTypes aware and works specifically on `mediaTypes.banner.sizes` rather than `adUnit.sizes` (although the later is still supported, but deprecated).

This fixes issue #2462 since it only actively filters `mediaType.banner.sizes` while allowing non-banner (such as native and video) and even multi-mediaType bid requests to remain active without `adUnit.sizes`.  However in the future, now that size mapping is aware of media types, we can use this feature to add sizeMapping support for filtering other non-banner media types if we want (we still need to have a discussion of how that would work).

### DIFF
--- a/src/adaptermanager.js
+++ b/src/adaptermanager.js
@@ -49,11 +49,22 @@ function getLabels(bidOrAdUnit, activeLabels) {
 
 function getBids({bidderCode, auctionId, bidderRequestId, adUnits, labels}) {
   return adUnits.reduce((result, adUnit) => {
-    let {active, mediaTypes: filteredMediaTypes} = resolveStatus(
+    let {
+      active,
+      mediaTypes: filteredMediaTypes,
+      filterResults
+    } = resolveStatus(
       getLabels(adUnit, labels),
       adUnit.mediaTypes,
       adUnit.sizes
     );
+
+    if (!active) {
+      utils.logInfo(`Size mapping disabled adUnit "${adUnit.code}"`);
+    } else if (filterResults) {
+      utils.logInfo(`Size mapping filtered adUnit "${adUnit.code}" banner sizes from `, filterResults.before, 'to ', filterResults.after);
+    }
+
     if (active) {
       result.push(adUnit.bids.filter(bid => bid.bidder === bidderCode)
         .reduce((bids, bid) => {
@@ -70,7 +81,17 @@ function getBids({bidderCode, auctionId, bidderRequestId, adUnits, labels}) {
             'renderer'
           ]));
 
-          let {active, mediaTypes} = resolveStatus(getLabels(bid, labels), filteredMediaTypes);
+          let {
+            active,
+            mediaTypes,
+            filterResults
+          } = resolveStatus(getLabels(bid, labels), filteredMediaTypes);
+
+          if (!active) {
+            utils.logInfo(`Size mapping deactivated adUnit "${adUnit.code}" bidder "${bid.bidder}"`);
+          } else if (filterResults) {
+            utils.logInfo(`Size mapping filtered adUnit "${adUnit.code}" bidder "${bid.bidder}" banner sizes from `, filterResults.before, 'to ', filterResults.after);
+          }
 
           if (utils.isValidMediaTypes(mediaTypes)) {
             bid = Object.assign({}, bid, {
--- a/src/adaptermanager.js
+++ b/src/adaptermanager.js
@@ -49,13 +49,11 @@ function getLabels(bidOrAdUnit, activeLabels) {
 
 function getBids({bidderCode, auctionId, bidderRequestId, adUnits, labels}) {
   return adUnits.reduce((result, adUnit) => {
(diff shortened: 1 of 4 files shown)


snapwich:
Added some additional console logging around size mapping (such as when an adUnit is deactivated and displaying the result of before and after filtering results) as well as fixed a bug (#3137) where an auction continued even if there were no valid bids.

jsnellbaker:
Hi @snapwich Can you put together a docs PR with any documentation related changes in lieu of this PR?  

http://prebid.org/dev-docs/publisher-api-reference.html#setConfig-Configure-Responsive-Ads still makes references to `adUnits.sizes` as the primary field we do the filtering based on what's defined in the `sizeConfig` object, so this likely should be updated.  I'm not sure if there are any native examples that can also be updated to remove `sizes` field in the adUnit setup.

Thanks.

jsnellbaker:
Thanks for putting the docs PR together.  @jaiminpanchal27 gave his approval for this PR separately to me, so proceeding to merge.

RBContent:
Hi All,

Do you guys confirm that the same things happen when an adunit array contains different mediatypes objects, I don't mean multi-format elements, but elements of different mediatypes, like a collection of native and banner mediatype.

It seems that the native part doesn't pass. I get the same message: 'label or size filtering'.

Regards

jsnellbaker:
I'll take a look into this behavior; do you have a sample markup of the adunit array that didn't seem to work for you?

jsnellbaker:
@RBContent I've setup a test page that has a banner adUnit and a native adUnit showing together.  When I used the sizeMapping feature (and when I didn't), the delivery appeared to work as expected given the setup.  

The only thing I saw (@snapwich please review), is when I had an `adUnit.sizes` defined in the native adUnit and it was using a set of sizes that were not covered by the sizeConfig - the new logic printed a message in the debug console saying that the placement was filtered.  However it did not actually remove the bid object from being made (since it's a `native` type adunit).

If you're seeing a setup where this new logic doesn't appear to work, please share a snippet of your adunit/sizeConfig code (or provide a jsfiddle.net link) so we can review.

snapwich:
@jsnellbaker that sounds correct.  `adUnit.sizes` should not be set for a native adUnit, and if it is then the sizes are treated as banner sizes.  Currently no size filtering is done for video or native (as we would have to define what that means and how it works, since sizing is not the same), so if an adUnit has multiple mediaTypes then the adUnit cannot be disabled (e.g. if you have a native and banner mediaType adUnit, the request will still be sent even if the banner sizes are all filtered because it's still possible to return a valid native bid).

RBContent:
Hi,

The example I made for you on jsfiddle is available here: https://jsfiddle.net/tkugswd3/53/

You'll see one slot (top) dedicated for the native format and another one (bottom) for the banner format. 
The native ad comes from Quantum (no test ads) and the banner one from Appnexus.

Normally you should only see the appnexus banner appearing. If you want to see the native format showing, please delete line 46 and 103 to 115.

jsnellbaker:
@RBContent I think there's some problem with the quantum placement itself rather than the logic.

I've tried several things with your test page:
- replaced the quantum native ad unit with an appnexus test native ad unit while keeping/removing the sizeconfig code.
- disabled the size config logic (the lines you noted) while running on the master branch, as well as on the 1.25.0 branch (using the old logic).

In all the cases when trying to use the quantum placement, I've never seen it return an ad (even with the old code and the sizeConfig disabled).  When I tried to use the appnexus native test placement, it was returning an ad and rendering it fine in the manner I described above.  

From what I'm seeing - it doesn't appear to be caused by the new code introduced by this PR.

jsnellbaker:
@RBContent 

Here is a jsfiddle setup with the appnexus banner and appenxus test native placements, using the 1.26.0 version of Prebid.js (that includes this PR).  

When you load the page, you'll always see the native test ad render (regardless of the sizeConfig settings).  If you resize the horitzontal viewport in the jsfiddle and re-run the page, you'll see the banner either show the blue prebid ad or the white fallback ad (following the logic of the sizeConfig).  

In terms of the quantum placement, if you weren't expecting it to visually show an ad - how were you establishing the problem?  If you were looking at the Prebid console messages and saw it that it said it filtered out the native placement - that's just because the native placement has a set of `adUnit.sizes` that do not fall into sizeConfig; the native adunit is still included in the auction and would return an ad (if available).  

Native placements do not need to have `sizes` defined at all.  It was only needed previously due to the bug that this PR fixed.  If you remove the field now, the message won't show in the console log.

RBContent:
I can't see the jsfiddle link in your last comment.

jsnellbaker:
Oops sorry about that, here it is:
http://jsfiddle.net/fnuarsgp/

RBContent:
Thanks and indeed, it's a lot better now comparing to version 1.25.0.
I'm in touch with Quantum right now and hoping to find out what's the issue.

PR #3136, opened 2018-09-27
4235 prebid endpoint return empty array instead of 204 when no bids returned

teranchristian:
## Type of change

- [ ] Bugfix
- [ ] Feature
- [ ] New bidder adapter   
- [ ] Build related changes
- [ ] CI related changes
- [ x] Other

## Description of change

 - Add 'pxyz' alias.
 - Add version to prebid endpoint
 - support for 204 bid no response

### DIFF
--- a/modules/playgroundxyzBidAdapter.js
+++ b/modules/playgroundxyzBidAdapter.js
@@ -102,7 +102,7 @@ export const spec = {
     }
     if (syncOptions.pixelEnabled) {
       return [{
-        type: 'pixel',
+        type: 'image',
         url: '//ib.adnxs.com/getuidnb?https://ads.playground.xyz/usersync?partner=appnexus&uid=$UID'
       }];
     }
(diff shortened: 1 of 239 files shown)


modules/playgroundxyzBidAdapter.js
```diff
@@ -69,8 +69,10 @@ export const spec = {
 
     if (!serverResponse || serverResponse.error) {
       let errorMessage = `in response for ${bidderRequest.bidderCode} adapter`;
-      if (serverResponse && serverResponse.error) { errorMessage += `: ${serverResponse.error}`; }
-      utils.logError(errorMessage);
+      if (serverResponse.error) {
```
harpere: this will fail if `serverResponse` is undefined, which is the first part of the condition on line 70.

harpere: which tells me there probably needs to be a test for this scenario.

PR #3145, opened 2018-10-01
Prebid server support for OpenRTB Native bids

snapwich:
## Type of change
- [x] Feature

## Description of change
This is a pull-request for following the development of native support for the prebid server adapter's OpenRTB endpoint and getting feedback on how to integrate properly with Prebid.js adUnits.

## Other information
OpenRTB Spec v2.4: https://www.iab.com/wp-content/uploads/2016/01/OpenRTB-API-Specification-Version-2-4-DRAFT.pdf

OpenRTB Native Spec 1.2: https://www.iab.com/wp-content/uploads/2017/04/OpenRTB-Native-Ads-Specification-Draft_1.2_2017-04.pdf

### DIFF
--- a/modules/prebidServerBidAdapter/index.js
+++ b/modules/prebidServerBidAdapter/index.js
@@ -907,10 +907,13 @@ export function PrebidServer() {
     const adUnits = utils.deepClone(s2sBidRequest.ad_units);
 
     // at this point ad units should have a size array either directly or mapped so filter for that
-    const adUnitsWithSizes = adUnits.filter(unit => unit.sizes && unit.sizes.length);
+    const validAdUnits = adUnits.filter(unit =>
+      (unit.sizes && unit.sizes.length) ||
+      (unit.mediaTypes && unit.mediaTypes.native)
+    );
 
     // in case config.bidders contains invalid bidders, we only process those we sent requests for
-    const requestedBidders = adUnitsWithSizes
+    const requestedBidders = validAdUnits
       .map(adUnit => adUnit.bids.map(bid => bid.bidder).filter(utils.uniques))
       .reduce(utils.flatten)
       .filter(utils.uniques);
@@ -920,7 +923,7 @@ export function PrebidServer() {
       queueSync(_s2sConfig.bidders, consent);
     }
 
-    const request = protocolAdapter().buildRequest(s2sBidRequest, bidRequests, adUnitsWithSizes);
+    const request = protocolAdapter().buildRequest(s2sBidRequest, bidRequests, validAdUnits);
     const requestJson = request && JSON.stringify(request);
     if (request && requestJson) {
       ajax(
--- a/modules/prebidServerBidAdapter/index.js
+++ b/modules/prebidServerBidAdapter/index.js
@@ -347,7 +347,7 @@ const LEGACY_PROTOCOL = {
     return request;
   },
 
-  interpretResponse(result, bidderRequests, requestedBidders) {
+  interpretResponse(result, bidderRequests) {
     const bids = [];
     if (result.status === 'OK' || result.status === 'no_cookie') {
       if (result.bidder_status) {
@@ -453,8 +453,25 @@ let nativeImgIdMap = {
   image: 3
 };
 
+let nativeEventTrackerEventMap = {
+  impression: 1,
+  'viewable-mrc50': 2,
+  'viewable-mrc100': 3,
+  'viewable-video50': 4,
+};
+
+let nativeEventTrackerMethodMap = {
+  img: 1,
+  js: 2
+};
+
 // enable reverse lookup
-[nativeDataIdMap, nativeImgIdMap].forEach(map => {
+[
+  nativeDataIdMap,
+  nativeImgIdMap,
+  nativeEventTrackerEventMap,
+  nativeEventTrackerMethodMap
+].forEach(map => {
   Object.keys(map).forEach(key => {
     map[map[key]] = key;
   });
@@ -465,6 +482,7 @@ let nativeImgIdMap = {
  * e.g., https://<prebid-server-url>/v1/openrtb2/auction
  */
 let bidIdMap = {};
+let nativeAssetCache = {}; // store processed native params to preserve
 const OPEN_RTB_PROTOCOL = {
   buildRequest(s2sBidRequest, bidRequests, adUnits) {
     let imps = [];
@@ -472,6 +490,74 @@ const OPEN_RTB_PROTOCOL = {
 
     // transform ad unit into array of OpenRTB impression objects
     adUnits.forEach(adUnit => {
+      const nativeParams = processNativeAdUnitParams(utils.deepAccess(adUnit, 'mediaTypes.native'));
+      let nativeAssets;
+      if (nativeParams) {
+        try {
+          nativeAssets = nativeAssetCache[adUnit.code] = Object.keys(nativeParams).reduce((assets, type) => {
+            let params = nativeParams[type];
+
+            function newAsset(obj) {
+              return Object.assign({
+                required: params.required ? 1 : 0
+              }, obj ? utils.cleanObj(obj) : {});
+            }
+
+            switch (type) {
+              case 'image':
+              case 'icon':
+                let imgTypeId = nativeImgIdMap[type];
+                let asset = utils.cleanObj({
+                  type: imgTypeId,
+                  w: utils.deepAccess(params, 'sizes.0'),
+                  h: utils.deepAccess(params, 'sizes.1'),
+                  wmin: utils.deepAccess(params, 'aspect_ratios.0.min_width')
+                });
+                if (!(asset.w || asset.wmin)) {
+                  throw 'invalid img sizes (must provided sizes or aspect_ratios)';
+                }
+                if (Array.isArray(params.aspect_ratios)) {
+                  // pass aspect_ratios as ext data I guess?
+                  asset.ext = {
+                    aspectratios: params.aspect_ratios.map(
+                      ratio => `${ratio.ratio_width}:${ratio.ratio_height}`
+                    )
+                  }
+                }
+                assets.push(newAsset({
+                  img: asset
+                }));
+                break;
+              case 'title':
+                if (!params.len) {
+                  throw 'invalid title.len';
+                }
+                assets.push(newAsset({
+                  title: {
+                    len: params.len
+                  }
+                }));
+                break;
+              default:
+                let dataAssetTypeId = nativeDataIdMap[type];
+                if (dataAssetTypeId) {
+                  assets.push(newAsset({
+                    data: {
+                      type: dataAssetTypeId,
+                      len: params.len
+                    }
+                  }))
+                }
+            }
+            return assets;
+          }, []);
+        } catch (e) {
+          utils.logError('error creating native request: ' + String(e))
+        }
+      }
+      const videoParams = utils.deepAccess(adUnit, 'mediaTypes.video');
+      const bannerParams = utils.deepAccess(adUnit, 'mediaTypes.banner');
+
       adUnit.bids.forEach(bid => {
         // OpenRTB response contains the adunit code and bidder name. These are
         // combined to create a unique key for each bid since an id isn't returned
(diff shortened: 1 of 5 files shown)


modules/prebidServerBidAdapter/index.js
```diff
@@ -395,13 +397,97 @@ const OPEN_RTB_PROTOCOL = {
...
+                    'icon': 1,
+                    'image': 3
+                  })[type];
+
+                  if (Array.isArray(params.sizes)) {
+                    let sizes = params.sizes;
+                    if (!Array.isArray(sizes[0])) { // support single size or an array of sizes
+                      sizes = [sizes]
+                    }
+                    // TODO: Currently creating multiple image assets for each size.  Figure out if this is correct!
```
snapwich: Regarding request objects for native images, the OpenRTB Native 1.2 spec when implemented for an icon assets looks similar to the following:
```javascript
// image
{
  "id": 200,
  "required": 1,
  "img": {
    "type": 3,
    "wmin": 300,
    "hmin": 250
  }
}

// icon
{
  "id": 201,
  "required": 1,
  "img": {
    "type": 1,
    "w": 50,
    "h": 50
  }
}
```
Currently image and icon assets are defined on the prebid adUnit [as defined here](http://prebid.org/dev-docs/adunit-reference.html#adunitmediatypesnative) in a way specifying multiple sizes and/or aspect ratios.  I have coded this current implementation to loop over the sizes and create separate image assets for each size, however this is _probably_ not correct with the required for each as, in prebid.js, each size means the adUnit could be any one of those sizes and not require all sizes.  As I see it, to solve this problem the request could be mapped into separate image assets (as I have done here in the code) but remove the required flag, or a single image asset could be sent and the `ext` object could be somehow utilized to send multiple sizes.

Example:
```javascript
pbjs.addAdunits([{
	code: '/19968336/prebid_native_example_1',
	mediaTypes: {
		'native': {
			'image': {
        'required': true,
        'sizes': [[300, 250], [300, 600]]
      }
		}
	}
}]);

// gets represented in OpenRTB Native request as multiple assets
{
  "assets": [
    {
      "id": 200,
      "img": {  // notice lack of required since both can't be required
        "type": 3,
        "w": 300,
        "h": 250
      }
    }, {
      "id": 201,
      "img": {
        "type": 3,
        "w": 300,
        "h": 600
      }
    }
  ]
}

// or ext is used
{
  "assets": [
    {
      "id": 200,
      "required": 1,
      "img": {
        "type": 3,
        "ext": {
          "sizes": [{
            "w": 300,
            "h": 250
          }, {
            "w": 300,
            "h": 600
          }]
        }
      }
    }
  ]
}

// or something else entirely...
```

For aspect ratios, currently OpenRTB only supports a `wmin` and `hmin` property, whereas the Prebid.js documentation lists only `min_width`, `ratio_height`, and `ratio_width` being the only valid properties on an array of objects.  I think the same could be done as above, mapping each aspect ratio object in the array to an image asset, however only `wmin` would be set and the ratios would be set as `ext.ratio_height` and `ext.ratio_width`.

I'm also assuming here that if both `sizes` and `aspect_ratios` are specified on the adUnit that they don't apply to the same asset (i.e. `sizes[0] and `aspect_ratios[0]` doesn't mean there is only one image asset that must meet both constraints)?

modules/prebidServerBidAdapter/index.js
```diff
@@ -395,13 +397,97 @@ const OPEN_RTB_PROTOCOL = {
...
+      const nativeParams = processNativeAdUnitParams(utils.deepAccess(adUnit, 'mediaTypes.native'));
+      if (nativeParams) {
+        let assetCounter = 200;
+        try {
+          mediaTypes['native'] = {
+            request: JSON.stringify({
+              // TODO: determine if following OpenRTB recommended properties can be defined or are needed
+              // context: int,
+              // plcmttype: int,
+              // privacy: int
```
snapwich: I'm not sure if any of the above _recommended_ properties in the OpenRTB spec need to somehow be added to the `adUnit.mediaTypes.native` properties?

modules/prebidServerBidAdapter/index.js
```diff
@@ -395,13 +397,97 @@ const OPEN_RTB_PROTOCOL = {
...
+                    title: {
+                      len: params.len
+                    }
+                  }));
+                } else {
+                  let dataAssetTypeId = ({
+                    sponsoredBy: 1,
+                    body: 2,
+                    cta: 12,
+                    // TODO: clickUrl in request object?
```
snapwich: I'm not sure how or if `clickUrl` maps to any of the data types defined in the OpenRTB Native 1.2 spec.  I'm assuming that it does not map to anything in the request object and is assumed required for all bid responses (since ads aren't much good if you can't click on them to go somewhere).

snapwich:
Still awaiting feedback

clarkish:
hi @snapwich are there any updates on this one?

is there an ETA for the availability of OpenRTB Native via Prebid Server?

snapwich:
Can't get a valid native response back from appnexus prebid server but I have unit tests around it.  Need to decide how `context`, `plcmttype`, and `privacy` get sent for a native bid.  Currently just default the first two to `1` and not setting privacy.  These don't need to be resolved before merge though.

idettman:
Regarding: https://github.com/prebid/Prebid.js/pull/3145#issuecomment-506047662
Is this PR ready for review now?

snapwich:
Yes, ready for review. It's possible it might need updates but is safe for merge.

snapwich:
@jsnellbaker as far as I am aware prebid server might not even have full support for native bids from pbjs right now.  For testing I was able to verify that native impressions were being received in prebid server but I had to mock the responses from prebid server to verify this code. I was never able to get an end-to-end fully working.

Do we have an example of native working with prebid-server and a test endpoint?  If I recall correctly when coding this the appnexus endpoint would only return nobids through prebid server for native requests.

I was hoping to get this merged sooner than later (since it keeps getting merge conflicts) and work out the kinks on prebid-server as needed, but if it's possible to get an end-to-end test working I think that'd be great.

jsnellbaker:
@snapwich I'm checking internally about the appnexus adapter and some native test params that work through PBS.  Will follow-up on this point.

jsnellbaker:
@snapwich 

I received some feedback; the issue that we're seeing is the absence of an `eventtrackers` config object in the native request payload.  

If its v1.2 and an image asset is being requested, we require `eventtrackers` to be in the request that gives us “permission” to track the image assets
it should look like this:
```

**Comment by snapwich** [comment]
@jsnellbaker that makes sense.  any feedback on a test endpoint to appnexus or test parameters I could use to verify through prebid-server -> appnexus adapter?

**Comment by jsnellbaker** [comment]
This is a test page I setup and was trying to use.  The placements/params should return a valid appnexus native bid (they do when I run it through the client-side adapter).

[demo_native.html.zip](https://github.com/prebid/Prebid.js/files/3412368/demo_native.html.zip)

**Comment by mkendall07** [comment]
This pull request **introduces 2 alerts** when merging 61589dad2f0682fc6ac87af66d3ead0f70b6c121 into c2bdb94ca20af4a0d8b7daba5c68700dc3dfd582 - [view on LGTM.com](https://lgtm.com/projects/g/prebid/Prebid.js/rev/pr-2344a580b88c5afa589e37d9166343a1c633cb29)

**new alerts:**

* 2 for Unused variable, import, function or class

---

This project has automated code review enabled, but doesn't use the <a href="https://github.com/apps/lgtm-com">LGTM GitHub App</a>. Migrate over by <a href="https://github.com/apps/lgtm-com/installations/new">installing the app</a>. Read about the <a href="https://lgtm.com/blog/announcing_github_apps_migration">benefits of migrating to GitHub Apps</a> in the blog.

---

**Comment by jsnellbaker** [comment]
@snapwich I discovered an issue around this line of code:
https://github.com/prebid/Prebid.js/blob/master/modules/prebidServerBidAdapter/index.js#L723

If the native` adUnit` does not have a sizes property defined, it gets filtered out by this function and is not included in the `buildRequest()` call.  As a consequence, this result prevents an impression object from being made in ORTB request for that dropped adUnit.

Can you please take a look around this area of the code?

**Comment by snapwich** [comment]
@jsnellbaker I put in a check that allows native adUnits through prebid-server without sizes.  I was just testing how other Native adUnits seemed to be working (by passing size `[1,1]`) but I think this is fine as well.

**Comment by Austinb** [comment]
There is an issue with this code change and multiple banner size support in native ads in prebid.js version 2.34.0. In the example here http://prebid.org/dev-docs/show-multi-format-ads.html it shows support for multiple distinct sizes in the `native.image.sizes` property but when you use multiple distinct sizes the request format to a prebid server is wrong and the prebid server chokes on the input.  Is this multiple size format below no longer supported/allowed for prebid native?

From the example page above:
```
native: {
    image: {
        sizes: [
            [728,90],
            [970,90]
        ]
    }
}
```

Becomes this example call into a prebid server with multiple sizes from prebis.js using a custom prebid server (latest version as of this comment's date):
```
{
  "id": "8c6253d3-fc62-4ede-a12b-4e03bed431d7",
  "source": {
    "tid": "8c6253d3-fc62-4ede-a12b-4e03bed431d7"
  },
  "tmax": 1000,
  "imp": [
    {
      "id": "/myunit/id",
      "ext": {
        "appnexus": {
          "use_pmt_rule": false,
          "placement_id": xxxxxx
        }
      },
      "banner": {
        "format": [
          {
            "w": 728,
            "h": 90
          },
          {
            "w": 970,
            "h": 90
          }
        ]
      },
      "native": {
        "request": "{\"context\":1,\"plcmttype\":1,\"eventtrackers\":[{\"event\":1,\"methods\":[1]}],\"assets\":[{\"required\":1,\"img\":{\"type\":3,\"w\":[728,90],\"h\":[970,90]}}]}",
        "ver": "1.2"
      }
    }
  ],
  "test": 0,
  "ext": {
    "prebid": {
      "targeting": {
        "includewinners": true,
        "includebidderkeys": false
      }
    }
  },
  "site": {
    "publisher": {
      "id": "1"
    },
    "page": "http://mysite.com"
  },
  "device": {
    "w": 2560,
    "h": 1329
  }
}
```
If you look at the `native.request.assets` section above you will see both the `w` and `h` properties are arrays of the defined sizes for the native request.  At this point is where the prebid-server chokes with the following error `Invalid request: json: cannot unmarshal array into Go struct field Image.w of type uint64`.  Removing the sizes from the `native.image` property allows the call to the prebid server to work as expected because that whole `native.request` section is no longer sent.  

Also using a single size such as `[1,1]` works as expected because the values of the `w` and `h` properties in the request to the prebid server are both integers and not arrays. Example payload:
```
{
  "id": "782e7d90-dfdf-48b1-b546-e772fe9d123a",
  "source": {
    "tid": "782e7d90-dfdf-48b1-b546-e772fe9d123a"
  },
  "tmax": 1000,
  "imp": [
    {
      "id": "/myunit/id",
      "ext": {
        "appnexus": {
          "use_pmt_rule": false,
          "placement_id": xxx
        }
      },
      "banner": {
        "format": [
          {
            "w": 728,
            "h": 90
          },
          {
            "w": 970,
            "h": 90
          }
        ]
      },
      "native": {
        "request": "{\"context\":1,\"plcmttype\":1,\"eventtrackers\":[{\"event\":1,\"methods\":[1]}],\"assets\":[{\"required\":1,\"img\":{\"type\":3,\"w\":1,\"h\":1}}]}",
        "ver": "1.2"
      }
    }
  ],
  "test": 0,
  "ext": {
    "prebid": {
      "targeting": {
        "includewinners": true,
        "includebidderkeys": false
      }
    }
  },
  "site": {
    "publisher": {
      "id": "1"
    },
    "page": "http://mysite.com"
  },
  "device": {
    "w": 2560,
    "h": 1329
  }
}
```

From a quick debug it appears the issues happening at https://github.com/prebid/Prebid.js/blame/a7ad5ef0c7284147c0027cbf4317997542a232d7/modules/prebidServerBidAdapter/index.js#L512 where the code pulls the 0 and 1 indexes assuming the value being passed is `[w,h]` and does not take into account for `[[w1,h1],[w2,h2]]`.  If multiple distinct sizes are no longer supported that's fine but the documentation needs to be updated to reflect that change.

snapwich:
I believe this is an error in documentation. If you look at 4.4 of the [OpenRTB Native Spec](https://www.iab.com/wp-content/uploads/2018/03/OpenRTB-Native-Ads-Specification-Final-1.2.pdf) there is no ability for an image asset object to have more than one size, only w, h or a wmin, hmin.

Austinb:
For multiple distinct sizes I assume we need to pass the `[1,1]` array moving forward then?

PR #3159, opened 2018-10-05
Render outstream safeframe 

jaiminpanchal27:
## Type of change
- [ ] Bugfix
- [x] Feature
- [ ] New bidder adapter   
- [ ] Build related changes
- [ ] CI related changes
- [ ] Other

## Description of change
Render outstream safeframe with prebid universal creative.

### DIFF
--- a/src/secureCreatives.js
+++ b/src/secureCreatives.js
@@ -9,7 +9,7 @@ import { EVENTS } from './constants';
 import { isSlotMatchingAdUnitCode } from './utils';
 import { auctionManager } from './auctionManager';
 import find from 'core-js/library/fn/array/find';
-import { executeRenderer } from './Renderer';
+import { isRendererRequired, executeRenderer } from './Renderer';
 
 const BID_WON = EVENTS.BID_WON;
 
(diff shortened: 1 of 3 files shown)


src/secureCreatives.js
```diff
@@ -53,9 +53,12 @@ function receiveMessage(ev) {
 }
 
 function sendAdToCreative(adObject, remoteDomain, source) {
-  const { adId, ad, adUrl, width, height } = adObject;
-
-  if (adId) {
+  const { adId, ad, adUrl, width, height, renderer } = adObject;
+  // rendering for outstream safeframe
+  if (renderer && renderer.url) {
```
mkendall07: should we move this logic into a separate function so it sits in one place? For example, it's also in pbjs.renderAd().

src/secureCreatives.js
```diff
@@ -53,9 +53,12 @@ function receiveMessage(ev) {
...
+    const bid = auctionManager.findBidByAdId(adId);
```
jsnellbaker: This `bid` value is the exact same as the `adObject` that gets passed into the function; so this may not be needed.

PR #3168, opened 2018-10-10
improves additional data passed to the handler in AuctionInit and AuctionEnd events

m4s0:
## Type of change

## Description of change

improves data passed to the handler in AuctionInit and AuctionEnd events

### DIFF
--- a/src/auction.js
+++ b/src/auction.js
@@ -96,6 +96,7 @@ export function newAuction({adUnits, adUnitCodes, callback, cbTimeout, labels}) 
   let _bidderRequests = [];
   let _bidsReceived = [];
   let _auctionStart;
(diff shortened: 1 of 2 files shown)


jsnellbaker:
@m4s0 Instead of creating separate `auction` variables at different points of the auction's workflow, it may be better to create a new function that will pull the datapoints from the auction and then call that function when the event's are triggered.  That way it one function object would be maintained (in case new variables get added/changed later) rather than two objects.

Could you look into refactoring this code along these lines?

m4s0:
@jsnellbaker ok, I refactored that code.

m4s0:
@jsnellbaker ok, I replace variables with a new function

PR #3172, opened 2018-10-11
Added user sync support for undertone bid adapter

omerko:
## Type of change
- [x] Feature

## Description of change
Added implementation of getUserSyncs in undertone adapter to allow syncing users.
If iframe sync is allowed will use iframe sync, otherwise, we'll use two sync pixels that is as allowed by default


modules/undertoneBidAdapter.js
```diff
@@ -20,26 +35,31 @@ export const spec = {
...
-    const location = utils.getTopWindowLocation();
-    let domains = /[-\w]+\.([-\w]+|[-\w]{3,}|[-\w]{1,3}\.[-\w]{2})$/i.exec(location.host);
     let domain = null;
-    if (domains != null && domains.length > 0) {
-      domain = domains[0];
-      for (let i = 1; i < domains.length; i++) {
-        if (domains[i].length > domain.length) {
-          domain = domains[i];
+    try {
+      const location = utils.getWindowTop().location;
```
jaiminpanchal27: Can you use `http://prebid.org/dev-docs/bidder-adaptor.html#referrers`.  You can use `referer` property to get top level host
Very soon we will be deprecating `utils.getWindowTop()`

omerko: Hi
Simply getting the referrer will not help me in this case.
I'm trying to retrieve the canonical url from the <link rel="canonical".. > element <link (https://www.w3.org/TR/capability-urls/#canonical-urls) 
Only in case it does not exist, I need to use the full referrer url.

Will you support such functionality to access the top level document?

omerko: I can try to access the top level document myself.

But regarding the new API to get the referer, I need to get the hostname out of it.
since it's a string I cannot to location.host, but can I use the URL api: https://developer.mozilla.org/en-US/docs/Web/API/URL
it does not support older IE versions

jaiminpanchal27: @omerko 
To get host you can use https://github.com/prebid/Prebid.js/blob/master/src/url.js#L27

To get canonical url, your current solution is fine. I will talk to team and try to add this feature. Can you create an issue with this feature request ?

omerko: Thanks, I used the host extraction the way you suggested and for canonical I replaced getWindowTop with direct access to the window.top from adapter code.

Also I created a feature request for the canonical URL extraction here: https://github.com/prebid/Prebid.js/issues/3221

omerko:
I used the host extraction the way you suggested and for canonical I replaced getWindowTop with direct access to the window.top from adapter code.

Also I created a feature request for the canonical URL extraction here: #3221

Waiting for feedback

PR #3183, opened 2018-10-15
Add a check against the size config when setting targeting

### LINKED ISSUE #2861: sizeConfig not honoured after window resize due to unexpired bids from previous auctions being used
## Type of issue

Question

## Description

Unexpired, unused bids from previous auctions can be selected as the winning bid, but if the window has since been resized then these bids can be for a size that is no longer eligible due to the sizeConfig rules. 

We’ve avoided this issue by using a new adUnit and code each time we call request bids after a window resize, but should Prebid be ensuring any bids from previous auctions are also still valid for the sizeConfig rules before using them in the current auction? 

## Platform details
Prebid 1.16.0

snapwich:
## Type of change
- [x] Bugfix

## Description of change
Add an extra check for size configs when setting targeting so that a cached bid that was valid cannot be used if it has become invalid by the screen changing size since last auction.

## Other information
fixes #2861

### DIFF
--- a/src/targeting.js
+++ b/src/targeting.js
@@ -2,6 +2,7 @@ import { uniques, isGptPubadsDefined, getHighestCpm, getOldestHighestCpmBid, gro
 import { config } from './config';
 import { NATIVE_TARGETING_KEYS } from './native';
 import { auctionManager } from './auctionManager';
+import { sizeSupported } from './sizeMapping';
 import includes from 'core-js/library/fn/array/includes';
 
 const utils = require('./utils.js');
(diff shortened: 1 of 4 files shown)


src/targeting.js
```diff
@@ -198,6 +199,7 @@ export function newTargeting(auctionManager) {
 
   function getBidsReceived() {
     const bidsReceived = auctionManager.getBidsReceived()
+      .filter(bid => bid.mediaType !== 'banner' || sizeSupported([bid.width, bid.height]))
```
jaiminpanchal27: What is the reason for this `bid.mediaType !== 'banner'` here ?

snapwich: size mapping only applies to (and filters) `banner` ads currently, so if it is not a `banner` bid then the filter will immediately let it through and not apply any size filtering.

PR #3184, opened 2018-10-16
Adkernel adapter ref and meta keywords forwarding

ckbo3hrk:
## Type of change
- [x] Feature

## Description of change
Added meta keywords and page referrer forwarding to adserver

- denis@adkernel.com

### DIFF
--- a/modules/adkernelBidAdapter.js
+++ b/modules/adkernelBidAdapter.js
@@ -3,6 +3,7 @@ import { BANNER, VIDEO } from 'src/mediaTypes';
 import {registerBidder} from 'src/adapters/bidderFactory';
 import find from 'core-js/library/fn/array/find';
 import includes from 'core-js/library/fn/array/includes';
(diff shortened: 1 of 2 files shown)


modules/adkernelBidAdapter.js
```diff
@@ -199,10 +199,16 @@ function getLanguage() {
  */
 function createSite() {
   var location = utils.getTopWindowLocation();
-  return {
+  let result = {
```
jaiminpanchal27: Please use http://prebid.org/dev-docs/bidder-adaptor.html#referrers instead of `utils.getTopWindowLocation();
 `. We will deprecation this function and `utils.getTopWindowReferrer()` soon

ckbo3hrk: @jaiminpanchal27 Done. Unfortunately there is a circleci build failure in unrelevant (I beleive) unit test.

PR #3190, opened 2018-10-17
Ensure transaction ID is always unique

naegelin:
## Type of change

- [x] Bugfix
- [ ] Feature
- [ ] New bidder adapter   
- [ ] Build related changes
- [ ] CI related changes
- [ ] Other

## Description of change
Transaction ID implicitly suggests that this is a unique identifier per transaction. However in the case of a refresh / re-use of an adunit the transaction ID remains the same causing undesirable bidding with certain SSPs.

### DIFF
--- a/src/prebid.js
+++ b/src/prebid.js
@@ -347,10 +347,9 @@ $$PREBID_GLOBAL$$.requestBids = createHook('asyncSeries', function ({ bidsBackHa
     const bidders = (s2sBidders) ? allBidders.filter(bidder => {
       return !includes(s2sBidders, bidder);
     }) : allBidders;
-
+    
     adUnit.transactionId = utils.generateUUID();
-   
-
+    
     bidders.forEach(bidder => {
       const adapter = bidderRegistry[bidder];
       const spec = adapter && adapter.getSpec && adapter.getSpec();
--- a/src/prebid.js
+++ b/src/prebid.js
@@ -348,9 +348,8 @@ $$PREBID_GLOBAL$$.requestBids = createHook('asyncSeries', function ({ bidsBackHa
       return !includes(s2sBidders, bidder);
     }) : allBidders;
 
-    if (!adUnit.transactionId) {
-      adUnit.transactionId = utils.generateUUID();
-    }
+    adUnit.transactionId = utils.generateUUID();
+   
 
     bidders.forEach(bidder => {
       const adapter = bidderRegistry[bidder];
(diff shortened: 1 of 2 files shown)


naegelin:
lint checker is failing using the standard github inline editor.  Leaving this as is for the time being .

jsnellbaker:
@naegelin The Prebid.js project uses the `gulp-eslint` plugin to maintain the styling rules for the project.  Can you please make the necessary updates in your file to conform to the linting rules?  You can locally check the linting status/errors by running `gulp lint` in your Prebid.js project.

naegelin:
The tests for transaction id validation are now failing (obviously).  I'm leaving this as is for discussion first @mkendall07 @jsnellbaker as a decision needs to be made regarding what the intended behavior should actually be here.

snapwich:
I'm not sure on what the business logic of `transactionId` should be, maybe @bretg has input?  

In #2706 I specifically tried to copy the existing business logic of having a persistent `transactionId` assigned to each adUnit.  Before #2706, `transactionId`s were added when you called `pbjs.addAdUnit()` but they were not refreshed if you were to `pbjs.requestBids()` multiple times.  I kept that same business logic in #2706, however I moved the `transactionId` creation inside of `pbjs.requestBids()` to ensure that all adUnits had a `transactionId`, including ones passed directly to `pbjs.requestBids()` without the use of `pbjs.addAdUnit()`.

mkendall07:
ah interesting, thanks Rich. @jaiminpanchal27 do you have any idea on this?

bretg:
The definition of transactionID is in http://prebid.org/dev-docs/bidder-adaptor.html#bidrequest-parameters:

```
Transaction ID is unique for each ad unit with a call to requestBids, but same across bidders. This is the ID that DSPs need to recognize the same impression coming in from different supply sources.
```

snapwich:
So the only other concern I could possibly think of with this pull-request was if an implementor wanted to generate their own transactionIds and pass them through then this would overwrite their ids.  Is that something that people do?  @bretg I know we generated our own transactionIds as part of hpv2, but I can't remember if that was just a temporary fix because #2706 hadn't happened yet.

If it's no big deal then this LGTM.  (other than the tests which need to be updated as well).

bretg:
I think we found that transaction ID wasn't being added because we weren't using pbjs.addAdUnit. Our code checks:
```
                    if (!adUnit.transactionId) {
                        adUnit.transactionId = generateUUID();
                    }
```
So yes - I think #2706 should have resolved that issue. So I think we're good here.

mkendall07:
@naegelin 
Please update the test and we can merge this. Thanks!

naegelin:
Thanks for updating guys - was on vacation and just saw this :)

Issue #3187, opened 2018-10-16
unit tests failing in safari browsers


jsnellbaker:
## Type of issue
bug?

## Description
There are a few of unit test failures in safari ever since we merged #3182 

The errors are occurring within the `consentManagement_spec.js` file, which is not touched by that PR.  But they have been consistently failing due to a `script error (:0)` type message.  

The tests have been temporarily disabled to remove the issue with the master build (see https://github.com/prebid/Prebid.js/pull/3186), but we need to investigate to see what's going wrong.

Below are links to the cricleci builds with the failures:
https://circleci.com/gh/prebid/Prebid.js/647
https://circleci.com/gh/prebid/Prebid.js/646

Possible reference?
https://stackoverflow.com/questions/5913978/cryptic-script-error-reported-in-javascript-in-chrome-and-firefox/7778424#7778424

PR #3193, opened 2018-10-17
add stub for adloader.loadScript in various adapter test files

jsnellbaker:
## Type of change
- [x] Other

## Description of change
This PR adds a stub for the `adloder.loadScript` function in various adapters' spec files that test the `Renderer` object.  

Previously, there were a series of sporadic errors/warnings when running browserstack tests due to the `loadScript` function attempting to load valid/faked values for the renderer source url.  https://github.com/prebid/Prebid.js/issues/3187 is one example of the errors that were causing unit test failures.  

With the stubs in place, these requests will be properly handled and greatly reduce the spam in the browserstack test's console output.


snapwich:
LGTM

PR #3196, opened 2018-10-17
Stub adloader across the board to prevent requests going out

snapwich:
## Type of change

## Description of change
Pulled out the adloader stubbing into a global mock so that it is stubbed by default.  Modules that need to use the actual adloader implementation (like adloader_spec.js) can access the original through the mock module.

### DIFF
--- a/test/mocks/adloaderStub.js
+++ b/test/mocks/adloaderStub.js
@@ -1,0 +1,22 @@
+const adloader = require('src/adloader');
+
+let sandbox;
+
+export let loadScript;
+export let loadExternalScript;
+export let loadScriptStub;
+export let loadExternalScriptStub;
+
+beforeEach(function() {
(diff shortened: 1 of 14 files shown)


jaiminpanchal27:
That is the goal as mentioned in the description  https://github.com/prebid/Prebid.js/pull/2779 by @jsnellbaker. Jason did some cleanup.

snapwich:
@mkendall07 I'm not sure on the status of adloader.js as a whole, however it only exports two functions (`loadScript` and `loadExternalScript`) and of the two, only one of those [is marked as deprecated](https://github.com/prebid/Prebid.js/blob/master/src/adloader.js#L41)

mkendall07:
alright well the work is already done here so I'm good with merging. Eventually we should delete it. Thanks!

PR #3203, opened 2018-10-19
On set targeting

jsfaure:
## Type of change
- [X] Feature

## Description of change
Add onSetTargeting method to bid adapter spec

Corresponding change for prebid.org: https://github.com/prebid/prebid.github.io/pull/1008


src/auctionManager.js
```diff
@@ -91,6 +91,9 @@ export function newAuctionManager() {
   auctionManager.setStatusForBids = function(adId, status) {
     let bid = auctionManager.findBidByAdId(adId);
     if (bid) bid.status = status;
+
+    const auction = find(_auctions, auction => auction.getAuctionId() === bid.auctionId);
```
bretg: This setStatusForBids function could get called for other statuses someday. Calling setBidTargeting needs to be done only when `status` is 'BID_TARGETING_SET'

jsfaure:
> Also need to consider how to add this to the unit tests. I can't find tests for the other event (onBidWon) in the spec files either, so we ought to add one for that too. Adding @jaiminpanchal27 for advise on unit tests.

Added several tests for the following events: BidWon, SetTargeting, Timeout.

src/auctionManager.js
jsfaure: Done

PR #3206, opened 2018-10-19
add logic to resize AST divs and iframes

jsnellbaker:
**NOTE** check with @mkendall07 first before merging.

## Type of change

- [x] Feature

## Description of change
This change adds/fixes some of the logic used in the `resizeRemoteCreative` function.  The function will now:
- not only rely on GAM tags being present on the page; it will now support AST tags or a div that matches the value in the bid's adUnitCode*
- throw a warning message when it's unable to find the page element to resize instead of either silently failing or throwing an undefined exception.

* this back-up can be removed if desired.  I added it as a catch-all fallback instead of throwing an warning/error.

### DIFF
--- a/src/secureCreatives.js
+++ b/src/secureCreatives.js
@@ -6,11 +6,10 @@
 import events from './events';
 import { fireNativeTrackers } from './native';
 import { EVENTS } from './constants';
-import { isSlotMatchingAdUnitCode } from './utils';
+import { isSlotMatchingAdUnitCode, logWarn } from './utils';
 import { auctionManager } from './auctionManager';
 import find from 'core-js/library/fn/array/find';
 import { isRendererRequired, executeRenderer } from './Renderer';
-import { utils } from 'mocha';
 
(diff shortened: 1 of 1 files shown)


src/secureCreatives.js
```diff
@@ -74,12 +74,39 @@ function sendAdToCreative(adObject, remoteDomain, source) {
...
+    }
+  }
+
+  function getDfpElementId(adUnitCode) {
+    return find(window.googletag.pubads().getSlots().filter(isSlotMatchingAdUnitCode(adUnitCode)), slot => slot).getSlotElementId()
+  }
+
+  function getAstElementId(adUnitCode) {
+    let apn = window.apntag;
+    let astTagIds = apn.requests && apn.requests.tags && Object.keys(apn.requests.tags);
```
mkendall07: let's see if there is a way to do this without accessing `apn.request` object.

jsnellbaker: I replaced the `apntag.requests` reference with the new API function.

PR #3224, opened 2018-10-24
Encode native targeting params for DFP support.

### LINKED ISSUE #3155: Native ad key-value pairs passed to DFP contain values with characters that are not supported by DFP
## Type of issue
Bug

## Description
Native ad key-value pairs passed to DFP contain values with characters that are not supported by DFP. 
https://support.google.com/admanager/answer/1697712?hl=en under `Rules and limitations around key-values in GPT ad tags` indicates `The values of key-values cannot contain any of the following special characters: " ' , = ! + # * ~ ; ^ ( ) < > [ ] `

This mostly effects hb_native_icon, hb_native_image, hb_native_linkurl as these are URLs.

Here is an example of the query string passed to DFP:
`hb_native_icon=http%3A%2F%2Fpix.us.criteo.net%2Fimg%2Fimg%3Fh%3D200%26m%3D0%26p%3D1%26partner%3D11251%26q%3D100%26r%3D0%26u%3Dhttp%253A%252F%252Fstatic.va.us.criteo.net%252Fdesign%252Fdt%252F11251%252F140512%252F408541.jpg%26v%3D3%26w%3D200%26s%3DO2ovyPbGi0R6QCPl_UVvDu3e`

This decodes to `http://pix.us.criteo.net/img/img?h=200&m=0&p=1&partner=11251&q=100&r=0&u=http%3A%2F%2Fstatic.va.us.criteo.net%2Fdesign%2Fdt%2F11251%2F140512%2F408541.jpg&v=3&w=200&s=O2ovyPbGi0R6QCPl_UVvDu3e` which contains `=` which is not a valid character for a key-value value field.

In a custom Prebid native creative, if I console log out these values, hb_native_icon, hb_native_image, and hb_native_linkurl are rarely rendered if ever. However hb_native_brand, hb_native_body, hb_native_title render fine as text:
````
console.log('hb_native_icon', %%PATTERN:hb_native_icon%%');
console.log('hb_native_image', '%%PATTERN:hb_native_image%%');
console.log('hb_native_linkurl', '%%PATTERN:hb_native_linkurl%%');
console.log('hb_native_brand', '%%PATTERN:hb_native_brand%%');
console.log('hb_native_body', '%%PATTERN:hb_native_body%%');
console.log('hb_native_title', '%%PATTERN:hb_native_title%%');
````

I've tested setting different values to key-value pairs and confirmed the `=` characters are causing the issues with this working with the custom native ad.

I'm using Prebid.js 1.16, however I have also tested Prebid.js 1.25.0 and confirmed the issue remains.

### PR DESCRIPTION
**Pull request by nanek**
## Type of change
- [X] Bugfix

## Description of change
Fixes #3155

Encodes all native targeting params sent to DFP. The DFP native creative then needs to decode each value. Without encoding the values never make it through DFP to the creative.

This is a breaking change, however I'm unaware of how existing code works for anyone using DFP. If this is a concern, we can add a configuration parameter to toggle this, or at least note this in the changelog.

### DIFF
--- a/src/native.js
+++ b/src/native.js
@@ -174,11 +174,11 @@ export function getNativeTargeting(bid) {
       value = value.url;
     }
 
-    // encode fields as they might contain invalid DFP key-value values
-    // such as `=` or `!`.
-    value = encodeURIComponent(value);
-
     if (key) {
+      // encode fields as they might contain invalid DFP key-value values
+      // such as `=` or `!`.
+      value = encodeURIComponent(value);
+
       keyValues[key] = value;
     }
   });
--- a/src/native.js
+++ b/src/native.js
@@ -115,13 +115,24 @@ export function nativeBidIsValid(bid, bidRequests) {
  * message should contain an `action` set to 'click'.
  *
  * // Native creative template example usage
- * <a href="%%CLICK_URL_UNESC%%%%PATTERN:hb_native_linkurl%%"
+ * <a class='js-native-link' href=''
  *    target="_blank"
- *    onclick="fireTrackers('click')">
- *    %%PATTERN:hb_native_title%%
+ *    onclick="fireTrackers('click')"
+ *    id="native-title">
  * </a>
  *
  * <script>
+ *   var linkUrl = '%%PATTERN:hb_native_linkurl%%';
+ *   linkUrl = '%%CLICK_URL_UNESC%%' + decodeURIComponent(linkUrl);
+ *   var links = document.getElementsByClassName('js-native-link');
+ *   for (var i = 0; i < links.length; i++) {
+ *     links[i].href = linkUrl;
+ *   };
+ *
+ *   var titleText = '%%PATTERN:hb_native_title%%';
+ *   titleText = decodeURIComponent(titleText);
+ *   document.getElementById('native-title').innerHTML = titleText;
+ *
  *   function fireTrackers(action) {
  *     var message = {message: 'Prebid Native', adId: '%%PATTERN:hb_adid%%'};
  *     if (action === 'click') {message.action = 'click';} // fires click trackers
@@ -162,6 +173,10 @@ export function getNativeTargeting(bid) {
     if (typeof value === 'object' && value.url) {
       value = value.url;
     }
+
+    // encode fields as they might contain invalid DFP key-value values
+    // such as `=` or `!`.
+    value = encodeURIComponent(value);
 
     if (key) {
       keyValues[key] = value;
--- a/test/spec/native_spec.js
+++ b/test/spec/native_spec.js
@@ -6,10 +6,10 @@ const utils = require('src/utils');
 const bid = {
   native: {
     title: 'Native Creative',
-    body: 'Cool description great stuff',
+    body: 'Cool description great stuff!',
     cta: 'Do it',
     sponsoredBy: 'AppNexus',
-    clickUrl: 'https://www.link.example',
+    clickUrl: 'https://www.link.example?param=value',
     clickTrackers: ['https://tracker.example'],
     impressionTrackers: ['https://impression.example'],
     javascriptTrackers: '<script src=\"http://www.foobar.js\"></script>'
@@ -32,9 +32,12 @@ describe('native.js', function () {
 
   it('gets native targeting keys', function () {
     const targeting = getNativeTargeting(bid);
-    expect(targeting[CONSTANTS.NATIVE_KEYS.title]).to.equal(bid.native.title);
-    expect(targeting[CONSTANTS.NATIVE_KEYS.body]).to.equal(bid.native.body);
-    expect(targeting[CONSTANTS.NATIVE_KEYS.clickUrl]).to.equal(bid.native.clickUrl);
+    const title = decodeURIComponent(targeting[CONSTANTS.NATIVE_KEYS.title]);
+    const body = decodeURIComponent(targeting[CONSTANTS.NATIVE_KEYS.body]);
+    const linkurl = decodeURIComponent(targeting[CONSTANTS.NATIVE_KEYS.clickUrl]);
+    expect(title).to.equal(bid.native.title);
+    expect(body).to.equal(bid.native.body);
+    expect(linkurl).to.equal(bid.native.clickUrl);
   });
 
   it('fires impression trackers', function () {

### CODE REVIEW

**Comment by wqi1972** [comment]
Great, this is exactly we are looking for. Thanks @nanek

**Comment by mkendall07** [comment]
@nanek 
Thanks for the fix! I had thought DFP encoded params automatically - this is a bit surprising. Would you also please open up a docs PR for http://prebid.org/adops/setting-up-prebid-native-in-dfp.html if changes are needed on the DFP setup?

**Comment by wqi1972** [comment]
@mkendall07 DFP does encode params automatically, but to pass special characters `" ' , = ! + # * ~ ; ^ ( ) < > [ ]` as key/value into dfp, it need do double encoding. Per: https://support.google.com/admanager/answer/177381?hl=en, DFP does not support key/values contain these special characters. I pinged googled, and requested them to make a change inside gpt.js, so that it double encode these special characters inside gpt.js, but google does not willing to make that fix, since for most scenarios, these special characters exist only inside Creatives content, and not passing as key/values to DFP backend.
For prebid native setup, it passes content via key/value, and it is quite normal need pass `" ' , = ! + # ;` inside either body text or url link. For now we are using alternative solutions which either avoid or encode content inside APN Console campaign setup. We do need the PR @nanek provide here, so that we can use these special characters without limitation.

**Comment by mkendall07** [comment]
@nanek 
Unfortunately this breaks existing native templates because of encoding all the content and the DFP template does not automatically encode. Can you update to detect the condition for special characters and only encode them if they exist?

**Comment by nanek** [comment]
Sounds good. 

I'm not quite sure when I'll get to pick this back up. So if anyone wants to make the required changes, feel free.

@mkendall07 I've been wondering if the native template should just be completely rendered by Prebid instead of using a DFP template? Is there any downside to this? Upsides include no need to pass KVPs to DFP, can use existing prebid display line items instead of creating new ones for native. The downside of passing the KVPs is that even if the encoding is fixed, the URLs especially with single request mode will be of variable length and potentially too long and therefore dropping all impressions.  Seems like pbjs.renderAd could determine that the bid is a native one and look for a template/css in Prebid.js to render and be aware of all template params. Is this something that has been considered? Is there a downside to this? Would this be a use of a custom Prebid renderer?

**Comment by mkendall07** [comment]
@nanek 
Thanks for the suggestion. AFAIK we considered the client side rendering for native but ultimately choose DFP templates because we thought it was an easier publisher integration (UI setup vs coding on page). 

If there is some hybrid approach that would let us leverage the UI and also do client side rendering that might also be an option. Open to suggestions.

**Comment by AmerSharif** [comment]
Perfect! This issue is a blocker for us at the moment, so very happy that you guys are looking at it.

**Comment by BZSky** [comment]
Hey mates! What about img & icon urls? We have this issue with them also.

**Comment by nanek** [comment]
@mkendall07 I reviewed this again and I'm not sure we can conditionally encode and decode. The problem is it's pretty common to have the original key value contain encoded characters if its a URL. For example, if one of those characters is an `=` encoded, and `=` not encoded, and we encode the un-encoded `=`, then the creative in DFP will not know to only decode the one we encoded, and not the one in the original URL. Any solution around this I can think of gets pretty complex and prone to bugs pretty quickly. 

I think the safest and simplest route is the one proposed here, to just encode everything and decode everything. I agree it is a breaking change, however I'm not sure out of a test creative how there would be a URL without a query param in practice based on the creatives I've seen returned.

Perhaps this PR of just encoding/decoding everything can be considered for Prebid 2.0? https://github.com/prebid/Prebid.js/issues/3278

**Comment by jsnellbaker** [comment]
@nanek Can you elaborate how you're handling the decoding on the DFP-side more?  Specifically with regards to URLs (for images and icons) that (depending on the native template) may be setup in the CSS-side rather than in the HTML-side of the template.  

From what I've been looking at, there are limited options when decoding in CSS - so I was curious what you did here to make it work.

**Comment by nanek** [comment]
@jsnellbaker The template I've used for testing does not set URLs in CSS. Below is my DFP creative. If you need a URL in CSS, I think you would need to create the CSS in javascript and inject a style element into the page.

```
<div id='adunit'>
  <div class='attribution inset-box'>
    <span class='inset-box-text'>Ad</span>
  </div>
  <div class='image'>
    <a class='js-native-link' href='' target='_blank' onclick="window.track('click', '%%PATTERN:hb_adid%%');"><img id='native-image' src=''></a>
  </div>
  <div class='advertiser inset-box'>
    <a id='native-brand' class='js-native-link inset-box-text' href='' target='_blank' onclick="window.track('click', '%%PATTERN:hb_adid%%');"></a>
  </div>
  <div class='title'>
    <a id='native-title' class='js-native-link' href='' target='_blank' onclick="window.track('click', '%%PATTERN:hb_adid%%');"></a>
  </div>
  <a class='call-to-action js-native-link' href='' target='_blank' onclick="window.track('click', '%%PATTERN:hb_adid%%');">
    <span class='call-to-action-text'>Learn More</span>
  </a>
</div>

<script>
var linkUrl = '%%PATTERN:hb_native_linkurl%%';
linkUrl = '%%CLICK_URL_UNESC%%' + decodeURIComponent(linkUrl);
var links = document.getElementsByClassName('js-native-link');
for (var i = 0; i < links.length; i++) {
  links[i].href = linkUrl;
};

var imageUrl = '%%PATTERN:hb_native_image%%';
imageUrl = decodeURIComponent(imageUrl);
document.getElementById('native-image').src = imageUrl;

var titleText = '%%PATTERN:hb_native_title%%';
titleText = decodeURIComponent(titleText);
document.getElementById('native-title').innerHTML = titleText;

var brandText = '%%PATTERN:hb_native_brand%%';
brandText = decodeURIComponent(brandText);
document.getElementById('native-brand').innerHTML = brandText;

window.track = function (action, adId) {
  var message = {message: 'Prebid Native', adId: adId};
  if (action === 'click') { message.action = 'click'; }
  window.parent.postMessage(JSON.stringify(message), '*');
}
window.track('impression', '%%PATTERN:hb_adid%%');
</script>
```

jsnellbaker:
@nanek Thanks for providing the additional information.  From further internal discussion with @mkendall07 and @mjacobsonny we can accept this PR if the following was done:
- instead of incorporating this encoding as a base-line change, we should have it hidden behind a new setConfig flag (eg `setConfig({ native: { encodeKeys: true } });`) with the default setting being off/`false`.  This would not cause it to be a breaking change in 1.x, but it would also allow those publishers who need to incorporate these special characters as part of the native creative to make it work (so long as they have a working creative template setup in their adserver).
- On the point of encoding the entire keyvalue instead of just specific characters, that it is fine.  As you said, it may be tricky to selectively decode certain characters and have it work consistently.
- We need to update the docs to make note of this new field and it's subsequent effect for native creatives.

This would be an interim solution.  In the long-term (for 2.x or later), we plan to take another look at this need and see if it can be included as base-line, along with an updated creative template for general use.  

If you can take care of the current items listed above (namely 1 & 3) and incorporate them into this PR, we can move forward here.  Please let me know if you have any questions on the above.

Thanks.

wqi1972:
Found a use case which `decodeURIComponent` does not work. If the text includes: `'*!~()`, then `decodeURIComponent` will keep as they are, and breaks dfp key/value.
I changed `decodeURIComponent`  to `escape`, and inside template change `decodeURIComponent` to `unescape`, then it works.

mkendall07:
Closing this PR since we don't yet have a proper solution for this. Issue is covered here: https://github.com/prebid/Prebid.js/issues/3366

PR #3225, opened 2018-10-24
Only set native targeting if value exists.

nanek:
## Type of change
- [X ] Bugfix

## Description of change

With an appnexus test placementId, we get native bid responses without a body or cta set. These are then passed as KVPs to DFP as `undefined`.  For example: `hb_native_brand=AppNexus&hb_native_cta=undefined&hb_native_body=undefined`. Since there is a limit on the length of a URL sent to DFP, this PR does not set keys without valid values.

I've assumed that `false` is not a valid value, based on the context of the fields.

### DIFF
--- a/src/native.js
+++ b/src/native.js
@@ -147,7 +147,7 @@ export function fireNativeTrackers(message, adObject) {
 }
 
 /**
- * Gets native targeting key-value paris
+ * Gets native targeting key-value pairs
  * @param {Object} bid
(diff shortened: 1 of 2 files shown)


bretg:
code change simple enough - @mkendall07 - please take a quick look at the unit test. Adding you since this is related (somewhat) to https://github.com/prebid/Prebid.js/pull/3224

PR #3228, opened 2018-10-25
update AdGeneration adapter

banakemi:
## Type of change

## Description of change
- adding support for new currency.
- adding support for multi ad sizes.

- Maintainer: ssp-ope@supership.jp


modules/adgenerationBidAdapter.js
```diff
@@ -15,7 +14,7 @@ export const spec = {
    * @return boolean True if this is a valid bid, and false otherwise.
    */
   isBidRequestValid: function (bid) {
-    return !!(bid.params.id);
+    return !!(bid.params.id) && (isValidCurrency(bid.params.currency));
```
bretg: I recommend against adding this adapter-specific `bids.params.currency`. The standard way that the publisher is supposed to specify currency is by calling `setConfig` with currency.adServerCurrency. All modules should be reading this value rather than making the publisher specify the currency in each adapter's params.

```
pbjs.setConfig({
    "priceGranularity": "low",
    "currency": {
       "adServerCurrency": "JPY",
       "granularityMultiplier": 108
       "defaultRates": { "USD": { "JPY": 110 }}
    }
});
```
For the full picture see http://prebid.org/dev-docs/modules/currency.html

banakemi:
Hi @bretg 
Thank you for reviewing.
I corrected the places where you pointed out.
Please check it again.

banakemi:
Hi @bretg
Thank you for reviewing.
Please check it again.

PR #3230, opened 2018-10-25
adding Prebid.js version to request in quantcastBidAdapter

jacekburys-quantcast:
## Type of change
- [x] Feature

## Description of change

We would like to add a new field to the request containing the version of Prebid.js that was used to send it.

### DIFF
--- a/modules/quantcastBidAdapter.js
+++ b/modules/quantcastBidAdapter.js
@@ -93,7 +93,7 @@ export const spec = {
         bidId: bid.bidId,
         gdprSignal: gdprConsent.gdprApplies ? 1 : 0,
         gdprConsent: gdprConsent.consentString,
-        prebidJsVersion: pbjs.version
+        prebidJsVersion: '$prebid.version$'
       };
 
       const data = JSON.stringify(requestData);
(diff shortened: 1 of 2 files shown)


modules/quantcastBidAdapter.js
```diff
@@ -92,7 +92,8 @@ export const spec = {
         },
         bidId: bid.bidId,
         gdprSignal: gdprConsent.gdprApplies ? 1 : 0,
-        gdprConsent: gdprConsent.consentString
+        gdprConsent: gdprConsent.consentString,
+        prebidJsVersion: pbjs.version
```
bretg: Please don't assume pbjs is the object variable... use `$prebid.version$` instead of `pbjs.version`.

test/spec/modules/quantcastBidAdapter_spec.js
```diff
@@ -119,7 +119,8 @@ describe('Quantcast adapter', function () {
           domain
         },
         bidId: '2f7b179d443f14',
-        gdprSignal: 0
+        gdprSignal: 0,
+        prebidJsVersion: pbjs.version
```
bretg: Please use `$prebid.version$`.

jacekburys-quantcast:
Hi @bretg 
Could you review again, please?

Issue #3191, opened 2018-10-17
xaxis / xhb adapter not working


nikkoi:
## Type of issue
We're having extensive problems getting the xaxis / xhb adapter working in prebid 1.x+ versions. 

## Description
Everything is working fine in versions prior to 1, but we haven't been able to get a single impressions through with the 1.x version, even though we've done a lot of debugging with the xaxis team. 

## Steps to reproduce
Use the xhb adapter with prebid 1.x+

Is there anyone that can confirm that it actually works for them. And if so, did you have any problems setting it up?

nikkoi:
Here is an example, two bids are available, but it doesn't even attempt to set them as key values. How is this possible? @daniel-hoffmann thoughts?

![screenshot-2](https://user-images.githubusercontent.com/1727721/47209461-158fc200-d399-11e8-9609-64ac8e93a22b.png)
![screenshot-1](https://user-images.githubusercontent.com/1727721/47209462-16285880-d399-11e8-957d-9190da7e48de.png)

nikkoi:
Yeah, we got this sorted using custom adserver targeting and always use bids. But if this is intended way of working, I would at least document it in the adapters .md file, that it doesn't work without specifying custom targeting.

jsnellbaker:
This should be addressed in the next release (1.31.0) via the above PR.

PR #3233, opened 2018-10-26
fix deal targeting for cpm 0

snapwich:
## Type of change
- [x] Bugfix

## Description of change
Bids with cpm of 0 should still have deal targeting passed through.  Cleaned up the targeting a little bit and removed the "empty bid" logic.

## Other information
CPM issue referenced here: #3191

### DIFF
--- a/src/targeting.js
+++ b/src/targeting.js
@@ -35,18 +35,9 @@ export function getHighestCpmBidsFromBidPool(bidsReceived, highestCpmCallback) {
   // filter top bid for each bucket by bidder
   Object.keys(buckets).forEach(bucketKey => {
     let bidsByBidder = groupBy(buckets[bucketKey], 'bidderCode');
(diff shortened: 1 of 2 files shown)


src/targeting.js
```diff
@@ -35,20 +35,11 @@ export function getHighestCpmBidsFromBidPool(bidsReceived, highestCpmCallback) {
   // filter top bid for each bucket by bidder
   Object.keys(buckets).forEach(bucketKey => {
     let bidsByBidder = groupBy(buckets[bucketKey], 'bidderCode');
-    Object.keys(bidsByBidder).forEach(key => bids.push(bidsByBidder[key].reduce(highestCpmCallback, getEmptyBid())));
```
harpere: the empty bid is not necessary and was "winning" over a real bid with a 0 cpm.  So removing it fixes this issue.

PR #3235, opened 2018-10-26
add inskin iab vendor id: enables consent via string

micha:
## Type of change

- [x] Bugfix


bretg:
confirmed in https://vendorlist.consensu.org/vendorlist.json

PR #3236, opened 2018-10-26
use unit id being sent instead of hard coded auid

HolzAndrew:
- [x ] Other

## Description of change
uses unit bid param rather than hard coding unit Id. Updates some tests accordingly.

### DIFF
--- a/modules/openxoutstreamBidAdapter.js
+++ b/modules/openxoutstreamBidAdapter.js
@@ -27,11 +27,13 @@ export const spec = {
       return [];
     }
     let requests = [];
-    requests.push(buildOXBannerRequest(bidRequests, bidderRequest));
+    bidRequests.forEach(bid => {
+      requests.push(buildOXBannerRequest(bid, bidderRequest));
+    })
     return requests;
   },
-  interpretResponse: function(serverResponse, serverRequest) {
-    return handleVastResponse(serverResponse, serverRequest.payload)
+  interpretResponse: function(bid, serverResponse) {
+    return handleVastResponse(bid, serverResponse.payload)
   },
 };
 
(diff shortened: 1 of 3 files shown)


HolzAndrew:
@jasminenoack plz to review

modules/openxoutstreamBidAdapter.js
```diff
@@ -83,7 +74,7 @@ function buildCommonQueryParamsFromBids(bids, bidderRequest) {
     tws: getViewportDimensions(isInIframe),
     be: 1,
     bc: bids[0].params.bc || `${BIDDER_CONFIG}_${BIDDER_VERSION}`,
-    auid: '540141567',
+    auid: bids[0].params.unit,
```
bretg: Why are you only looking at bidRequest[0] in this code? Sounds like there's an underlying assumption that outstream adunits will be in their own separate auctions apart from other adunits?

HolzAndrew: @bretg I updated the adapter to make multiple requests now

modules/openxoutstreamBidAdapter.js
```diff
@@ -110,24 +103,28 @@ function buildCommonQueryParamsFromBids(bids, bidderRequest) {
   return defaultParams;
 }
 
-function buildOXBannerRequest(bids, bidderRequest) {
-  let queryParams = buildCommonQueryParamsFromBids(bids, bidderRequest);
-  queryParams.aus = utils._map(bids, bid => utils.parseSizesInput(bid.sizes).join(',')).join('|');
+function buildOXBannerRequest(bid, bidderRequest) {
+  let queryParams = buildCommonQueryParamsFromBids(bid, bidderRequest);
+  // let auids = utils._map(bids, bid => bid.params.unit)
```
jaiminpanchal27: Can you remove this commented code

jimee02:
LGTM

HolzAndrew:
@jaiminpanchal27 it looks like a test unrelated to our adapter has failed. Let me know if there is anything you need from us to update.
Thanks!

screen shot 2018-10-29 at 3 38 26 pm

PR #3251, opened 2018-11-02
Instream Video without DFP

freegames66:
Is there any solution like Postbid so that we can use Instream Video ads without DFP?


jaiminpanchal27:
@freegames66 Can you resolve conflicts. No for instream we do not have anything without adserver.

mkendall07:
@freegames66 
Can you describe what this PR is for?

freegames66:
> @freegames66
> Can you describe what this PR is for?

No, I don't PR anything. I had some problems with prebid.js however no one supported me.

PR #3257, opened 2018-11-02
appnexus adapter support empty keyvalues in bidder params

jsnellbaker:
## Type of change
- [x] Feature

## Description of change
This change allows the appenxus bidder param `keywords` to accept an empty keyvalue and subsequently pass just the `key` param to the adserver.

In terms of setup, the bidder param would just use an empty string as the value; examples below:
```
params: {
  placementId: 13144370,
  keywords: {
    mykeyword: [''],
  }
}
```
or 
```
params: {
  placementId: 13144370,
  keywords: {
    mykeyword: '',
  }
}
```

Based on the above, the `tags[...].keywords` object in the ut request would look like the following:
```
[{
  key: 'mykeyword'
}]
```

A similar structure would be used in the request to PBS for the `imp[...].ext.appnexus.keywords` object.

### DIFF
--- a/modules/appnexusBidAdapter.js
+++ b/modules/appnexusBidAdapter.js
@@ -178,6 +178,14 @@ export const spec = {
       params.use_pmt_rule = (typeof params.usePaymentRule === 'boolean') ? params.usePaymentRule : false;
       if (params.usePaymentRule) { delete params.usePaymentRule; }
 
+      if (utils.isArray(params.keywords) && params.keywords.length > 0) {
+        params.keywords.forEach(function(keyPairObj) {
+          if (utils.isArray(keyPairObj.value) && keyPairObj.value.length > 0 && keyPairObj.value[0] === '') {
+            delete keyPairObj.value;
+          }
+        });
+      }
+
       Object.keys(params).forEach(paramKey => {
         let convertedKey = utils.convertCamelToUnderscore(paramKey);
         if (convertedKey !== paramKey) {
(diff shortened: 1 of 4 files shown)


modules/appnexusBidAdapter.js
```diff
@@ -178,6 +178,14 @@ export const spec = {
       params.use_pmt_rule = (typeof params.usePaymentRule === 'boolean') ? params.usePaymentRule : false;
       if (params.usePaymentRule) { delete params.usePaymentRule; }
 
+      if (utils.isArray(params.keywords) && params.keywords.length > 0) {
+        params.keywords.forEach(function(keyPairObj) {
+          if (utils.isArray(keyPairObj.value) && keyPairObj.value.length > 0 && keyPairObj.value[0] === '') {
```
jaiminpanchal27: The if statement here and on line 360 are same. Please move in some function.

test/spec/modules/appnexusBidAdapter_spec.js
```diff
@@ -261,6 +261,8 @@ describe('AppNexusAdapter', function () {
               singleArrNum: [5],
               multiValMixed: ['value1', 2, 'value3'],
               singleValNum: 123,
+              emptyStr: '',
```
jaiminpanchal27: I am not sure if this is valid use case but what will happen if someone adds `emptyStr: ' '` (space is added as value).

jsnellbaker: the space character is treated as the value for the key-pair and is preserved in the ut request.

eg `"keywords":[{"key":"color","value":["red"]},{"key":"mytest"},{"key":"myspacetest","value":[" "]}]`

jsnellbaker: As a follow-up to the above, the treatment of the space character has not been changed through this PR.  Using a space character as a keyvalue in `master` would result in the same type of object for the ut request as noted above.

modules/appnexusBidAdapter.js
```diff
@@ -191,6 +195,16 @@ export const spec = {
   }
 }
 
+function isPopulatedArray(arr) {
+  return !!(utils.isArray(arr) && arr.length > 0);
+}
+
+function deleteValues(keyPairObj) {
```
mkendall07: nit; please rename to `deleteEmptyValues` or similar.

PR #3274, opened 2018-11-07
only set mediaTypes.banner.sizes from sizes if mediaTypes doesn't exist.

### LINKED ISSUE #3269: banner mediaTypes always added in bid despites provided mediaTypes in adUnits config
## Type of issue
Bug

## Description
If you try to setup an adUnit with mediaType video only, the mediaType banner is automatically added when calling the different bidders, which results in bidders making call for both video and banner instead of video only

## Steps to reproduce
1 - Create a basica prebid test page with a video mediaTypes:
pbjs.addAdUnit({
    code: 'video1',
    mediaTypes: {
        video: {
            context: 'outstream',
            playerSize: [640, 480]
        }
    },
    renderer: {
        options: {
            adText: 'This text was configured in the ad unit',
        }
    },
    ...
});

2 - check which mediaTypes are passed to bidders. It will contain both banner and video

## Test page
The bug is present on the demo page for outstream on prebid.org: http://prebid.org/examples/video/outstream/outstream-no-adserver.html

### Expected results
Only video mediaTypes should be passed to bidders, not banner

### Actual results
both banner and video mediaTypes are passed to bidders

## Platform details
Tested on prebid 1.30. It affects all bidders

## Other information
I was able to track down the issue to this part of the code:  https://github.com/prebid/Prebid.js/blob/7fe95d89464cbc95adf3446cd8e39bb499f2c321/src/adaptermanager.js

the getBids function. It seems to be adding banner mediaTypes when building the bid object that will be passed to bidders.

snapwich:
## Type of change
- [x] Bugfix

## Description of change
Fixes #3269.  The issue is when `mediaTypes.video.playerSize` is set then `adUnits.sizes` is set [here](https://github.com/prebid/Prebid.js/blob/master/src/adaptermanager.js#L284).  Since sizeMapping still supports `adUnit.sizes` it will set `adUnit.mediaTypes.banner.sizes` if it doesn't already exist, which is a bug if `adUnit.sizes` was set from the video sizes.  So to fix this, this pull-request only sets `adUnit.mediaTypes.banner.sizes` from `adUnit.sizes` if `adUnit.mediaTypes` doesn't exist.

### DIFF
--- a/src/sizeMapping.js
+++ b/src/sizeMapping.js
@@ -65,20 +65,18 @@ export function resolveStatus({labels = [], labelAll = false, activeLabels = []}
   let maps = evaluateSizeConfig(configs);
 
   if (!isPlainObject(mediaTypes)) {
-    mediaTypes = {};
+    // add support for deprecated adUnit.sizes by creating correct banner mediaTypes if they don't already exist
+    if (sizes) {
+      mediaTypes = {
+        banner: {
+          sizes
+        }
+      };
+    } else {
+      mediaTypes = {};
+    }
   } else {
     mediaTypes = deepClone(mediaTypes);
-  }
-
-  // add support for deprecated adUnit.sizes by creating correct banner mediaTypes if they don't already exist
-  if (sizes) {
-    if (!mediaTypes.banner) {
-      mediaTypes.banner = {
-        sizes
-      }
-    } else if (!mediaTypes.banner.sizes) {
-      mediaTypes.banner.sizes = sizes;
-    }
   }
 
   let oldSizes = deepAccess(mediaTypes, 'banner.sizes');


src/sizeMapping.js
```diff
@@ -65,22 +65,20 @@ export function resolveStatus({labels = [], labelAll = false, activeLabels = []}
   let maps = evaluateSizeConfig(configs);
 
   if (!isPlainObject(mediaTypes)) {
-    mediaTypes = {};
+    // add support for deprecated adUnit.sizes by creating correct banner mediaTypes if they don't already exist
+    if (sizes) {
```
nanek: Should this check if mediaTypes already exists, as stated in description?

I'm testing this fix out, but I still see mediaTypes.banner.sizes set when I'm only configuring an instream video mediaType.

nanek: My mistake, I was testing a prebid version without this fix. With the fix, I can confirm that it is working as expected and no longer setting mediaType.banner if only a video unit.

Thanks for the fix!

Issue #2805, opened 2018-07-03
ANOutstreamVideo.js + [All Renderers] Is being downloaded from bid Request even if bid isn't won or renderAd fired


mercuryyy:
http://acdn.adnxs.com/video/outstream/ANOutstreamVideo.js

Is the Video renderer for appnexus bids, other renderers from others bidders will probably behave the same.

When bids are requested from bidders, the renderer .js file will be downloaded to the client browser even if the video bids is not the highest cpm or the renderAd been fired from the dfp creative.

Those renderer's files are usually very large in appnexus case 167kb no reason to download this file if the video bid is not the winning bid or more so if renderAd has not been called.

Another note, even if we set our own renderer file in the adunit config it will still download the bidders renderer when the bid is requested.

lanceloti:
I'm using a proprietary Outstream Player and ANOutstreamVideo.js is downloaded sometimes too.

mkendall07:
@mercuryyy 
Thanks for the report. Do you have test page handy?

mkendall07:
@lanceloti 
Got a test page I could look at?

lanceloti:
Sorry from delay: http://maringa.odiario.com/ads/adplayer-instream.php?pbjs_debug=true

bwoolcott:
@mkendall07 is this still an issue?

lanceloti:
Hello guys.

It still an issue to me. See this test page: https://maringa.odiario.com/ads/outstream.php

1) Even I setting other render URL ANOutstreamVideo.js is downloaded first.

2) Render funcion is called before complete download of our player, so I need to set a timeout until script is ready.

PR #3284, opened 2018-11-09
Do not load external js if renderer defined on adUnit

jaiminpanchal27:
## Type of change

- [x] Bugfix

## Description of change
Fixes https://github.com/prebid/Prebid.js/issues/2805, Delaying loading of external js for winning bid will be done in a separate PR. This one just fixes a bug.

If renderer is defined on adUnit level, Renderer module will not load external js returned by bid. As of now 12 adapters are using Renderer. We need to update those if this approach is fine.

### DIFF
--- a/modules/appnexusBidAdapter.js
+++ b/modules/appnexusBidAdapter.js
@@ -211,6 +211,7 @@ function newRenderer(adUnitCode, rtbBid, rendererOptions = {}) {
     url: rtbBid.renderer_url,
     config: rendererOptions,
     loaded: false,
+    adUnitCode
   });
 
   try {
@@ -238,6 +239,7 @@ function newRenderer(adUnitCode, rtbBid, rendererOptions = {}) {
  * @return Bid
  */
 function newBid(serverBid, rtbBid, bidderRequest) {
+  const bidRequest = utils.getBidRequest(serverBid.uuid, [bidderRequest]);
   const bid = {
     requestId: serverBid.uuid,
     cpm: rtbBid.cpm,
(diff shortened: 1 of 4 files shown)


src/Renderer.js
```diff
@@ -35,12 +36,16 @@ export function Renderer(options) {
     this.process();
   });
 
-  // we expect to load a renderer url once only so cache the request to load script
-  loadScript(url, this.callback, true);
+  if (!isRendererDefinedOnAdUnit(adUnitCode)) {
+    // we expect to load a renderer url once only so cache the request to load script
+    loadScript(url, this.callback, true);
+  } else {
+    utils.logWarn(`External Js not loaded by Renderer since renderer url and callback is already defined on adUnit ${adUnitCode}`);
```
jsnellbaker: Does this need to be a warning?  

Given the nature of the setup (ie defining a renderer in the adunit), this seems like a `logInfo` type message since the user isn't doing anything wrong and the code is handling this scenario as expected.  Unless I'm missing something?

modules/appnexusBidAdapter.js
```diff
@@ -246,6 +248,7 @@ function newBid(serverBid, rtbBid, bidderRequest) {
     currency: 'USD',
     netRevenue: true,
     ttl: 300,
+    adUnitCode: bidRequest.adUnitCode,
```
mkendall07: Why is this needed on the bid?

src/Renderer.js
jaiminpanchal27: Here is the user has set that renderer defined on adUnit will render the bid but the demand partners they are working with are using different js to render.

PR #3291, opened 2018-11-13
add step to evaluate s2sconfig timeout

jsnellbaker:
## Type of change

- [x] Feature

## Description of change
Added some logic to check the s2sconfig timeout value and set a new value if it's found to be larger than the auction's provided timeout.  The new timeout is set at 75% of the auction timeout (this can be changed).

Note - we set the value in the requestBids method as it's the last place where the timeout gets set (ie in the `timeout` field in the object config you pass in the `pbjs.requestBids` function).

### DIFF
--- a/src/utils.js
+++ b/src/utils.js
@@ -1143,3 +1143,12 @@ export function convertTypes(types, params) {
   });
   return params;
 }
+
+export function evaluateTimeout(auctionTimeout, adapterTimeout, adapterName = 's2sConfig') {
+  if (auctionTimeout <= adapterTimeout) {
+    let newTimeout = Math.floor(0.75 * auctionTimeout);
+    exports.logInfo(`Detected auction timeout was smaller than the timeout set for ${adapterName}.  Adjusting ${adapterName} timeout to new value to compensate: ${newTimeout}`);
+    return newTimeout;
+  }
+  return adapterTimeout;
+}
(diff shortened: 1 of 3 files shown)


mkendall07:
leaving open until someone else can review.

src/utils.js
```diff
@@ -1143,3 +1143,12 @@ export function convertTypes(types, params) {
   });
   return params;
 }
+
+export function evaluateTimeout(auctionTimeout, adapterTimeout, adapterName = 's2sConfig') {
+  if (auctionTimeout <= adapterTimeout) {
```
bretg: In general I'm ok with shaving the timeout, but unclear 75% is the right number for every case. How about `s2sConfig.timeoutBufferPercent`?

Shouldn't this be `auctionTimeout * PERCENT <= adapterTimeout`?

Consider these scenarios:

auctionTimeout | original adapterTimeout | new adapterTimeout | Notes
-----|-------|--------|-----
1000 | 600 | 600 | original value under PERCENT
1000 | 800 | 750 | should pull adapterTimeout a little lower. Current code wouldn't notice this.
1100 | 1000 | 825 | current alg. also wouldn't notice this scenario

mkendall07: thanks. Jason is out this week so will wait for him to get back and update this.

jsnellbaker: @bretg 

I was thinking we could one day extend the use of the `evaluteTimeout` function to also work with the client-side adapters as well.  So instead of setting the coefficient value in the `s2sConfig`, what if we created another variable in the general `setConfig` to house the coefficient (ie similar to `bidderTimeout`)?  That way it would generally available and consistent throughout the auction.

As a follow-up either way - if we were to set the value through a config, should we have a default/backup value setup and if so, what value should it be if not 75%?  I set it to 75% as I thought it was a reasonable chunk of the auction's time to allow adapters to respond in most cases.

I see your point about adjusting the if logic, I'll make those changes.

jsnellbaker: @bretg Any thoughts on the above?

PR #3303, opened 2018-11-15
Quantcast supporting instream video

jacekburys-quantcast:
## Type of change

- [x] Feature

## Description of change

Adding support for instream video bid requests in Quantcast's adapter.

```
{
  bidder: 'quantcast',
  params: {
    publisherId: 'testt-publisher', // REQUIRED - Publisher ID provided by Quantcast
    // Video object as specified in OpenRTB 2.5
    video: {
      mimes: ['video/mp4'], // required
      minduration: 3, // optional
      maxduration: 5, // optional
      protocols: [3], // optional
      startdelay: 1, // optional
      linearity: 1, // optinal
      battr: [1, 2], // optional
      maxbitrate: 10, // optional
      playbackmethod: [1], // optional
      delivery: [1], // optional
      placement: 1, // optional
      api: [2, 3] // optional
    }
  }
}

```


test/spec/modules/quantcastBidAdapter_spec.js
```diff
@@ -125,7 +159,53 @@ describe('Quantcast adapter', function () {
...
       };
 
-      expect(requests[0].data).to.equal(JSON.stringify(expectedBidRequest));
+      expect(requests[0].data).to.equal(JSON.stringify(expectedBannerBidRequest));
+    });
+
+    it('sends video bid requests containing all the required parameters', function () {
+      setupVideoBidRequest();
+
+      const referrer = utils.getTopWindowUrl();
```
jaiminpanchal27: utils.getTopWindowUrl(); and utils.getTopWindowLocation(); will be removed soon. For this test you can skip site from expectedVideoBidRequest and do assertion.

jacekburys-quantcast: But I use these functions in buildRequests as well, so if I remove site from expectedVideoBidRequest the assertion will fail because site is set in buildRequests.

Or maybe I'm missing something. If that's the case could you, please, give a bit more detail?

jaiminpanchal27: Can you update your adapter to use http://prebid.org/dev-docs/bidder-adaptor.html#referrers

This new object has everything you need which these functions used to do.

jacekburys-quantcast: I hope it's ok now, though I'm not sure - I don't usually work with JavaScript :)

Issue #3330, opened 2018-11-27
flatmap-stream package deleted


Saigredan:
## Type of issue
BUG

## Description
According to current news, e.g. https://www.zdnet.com/article/hacker-backdoors-popular-javascript-library-to-steal-bitcoin-funds/,  flatmap-stream is down, which means  event-stream down, which means gulp connect and gulp replace are down. That implies general problem with building Prebid.
Are there any plans to fix this issue?

muuki88:
Fix in `gulp-connect` https://github.com/AveVlad/gulp-connect/pull/259

Saigredan:
Also need a change in gulp-replace verson in package.json up to at least 0.6.0

GLStephen:
Is there any guidance regarding effected versions of Prebid?

muuki88:
AFAIK distributions aren't affected as only build dependencies (`gulp-connect`, `gulp-replace`) used the malicious dependency.

GLStephen:
> AFAIK distributions aren't affected as only build dependencies (`gulp-connect`, `gulp-replace`) used the malicious dependency.

Ok, thanks for the confirmation.

jsnellbaker:
I will put together some changes for these gulp packages, however it seems we need to wait on the `gulp-connect` and `gulp-footer` updates.  Will keep an eye on this.

Saigredan:
@jsnellbaker why not using temporal forks?

whatisjasongoldstein:
So I'm not able to install Prebid _at all_ right now?

Austinb:
This was merged in as part of https://github.com/prebid/Prebid.js/releases/tag/1.35.0 and appears to be working now.  Was able to `npm install` and do a custom build without issue.

Many thanks to @jsnellbaker, @jaiminpanchal27 & @mkendall07

PR #3336, opened 2018-11-28
Set the OpenRTB device dimensions

dbemiller:
This addresses a comment that was left in Prebid Server: https://github.com/prebid/prebid-server/pull/748#discussion_r237265303

OpenRTB has a spot for the device width & height, which are useful for some people.


dbemiller:
Unfortunately I can't install the dependencies to run it locally due to #3330 :/.... I'll circle back later to fix these once it's resolved, unless someone who's already got them downloaded has cycles to take it over first :)

dbemiller:
this is not working... 1 more push coming soon.

PR #3341, opened 2018-11-30
support new type of user sync: html

naoto0822:
## Type of change
- [×] Feature

### support new type of user sync
new type  is `html` 
target html is loaded by iframe.

publisher opt in `userSync.filterSettings.html` in config.

```js
pbjs.setConfig({
  userSync: {
    filterSettings: {
      html: {
        bidders: '*',
        filter: 'include'
      }
    }
  }
});
```

bidder update as follow

```js
getUserSyncs: function(syncOptions, serverResponses) {
  const syncs = [];
  if (syncOptions.htmlEnabaled) {
    syncs.push({
      type: 'html',
      html: '<script>hoge.sync();</script>'
    });
  }
  return syncs;
},

... any code
```

### document

> if this PR approved

TODO:

### DIFF
--- a/src/adapters/bidderFactory.js
+++ b/src/adapters/bidderFactory.js
@@ -321,13 +321,18 @@ export function newBidder(spec) {
       let syncs = spec.getUserSyncs({
         iframeEnabled: !!(config.getConfig('userSync.iframeEnabled') || (filterConfig && (filterConfig.iframe || filterConfig.all))),
         pixelEnabled: !!(config.getConfig('userSync.pixelEnabled') || (filterConfig && (filterConfig.image || filterConfig.all))),
+        htmlEnabled: !!(filterConfig && (filterConfig.html || filterConfig.all)),
       }, responses, gdprConsent);
       if (syncs) {
         if (!Array.isArray(syncs)) {
           syncs = [syncs];
         }
         syncs.forEach((sync) => {
-          userSync.registerSync(sync.type, spec.code, sync.url)
+          if (sync.url) {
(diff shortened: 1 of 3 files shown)


jsnellbaker:
Hi @naoto0822,
Thanks for putting this together.  Can you provide a use-case for this functionality that's not already covered by the iframe-url userSync?  The sample code refers to calling a specific object's function, but this wouldn't work on its own without calling a separate file.  Is that what you're envisioning for this new functionality, or something else?

naoto0822:
@jsnellbaker 
Thanks for reply.
this use-case is when the server returns sync-script directly. sorry for lack of example.
correct code reads external js and executes it.

The correct example is folowing...
- server response
```json
{
    "sync_html": "<script src=\"https://hoge/foo.js\"></script><script>Foo.Sync(a:a, b:b, c:c...);</script>"
}
```

- adapter
```js
getUserSyncs: function(syncOptions, serverResponses) {
  const syncs = [];
  const html = serverResponses["sync_html"];
  if (syncOptions.htmlEnabaled) {
    syncs.push({
      type: 'html',
      html: html
    });
  }
  return syncs;
},
```

jsnellbaker:
@naoto0822 Thanks for providing the additional clarification; unfortunately based on further internal discussion, this type of feature isn't something we want to incorporate.  It's too open/broad to allow any type of script to execute on a publisher's page.  Sorry.

naoto0822:
@jsnellbaker 
Thanks for your explanation. 

> It's too open/broad to allow any type of script to execute on a publisher's page. 

OK、this is correct.
Thank you.

PR #3347, opened 2018-12-03
Bidfluence Adapter update - Single call

francescocristallo:
## Type of change

## Description of change
Same code refactored for sending a single call to the server instead of multiple calls for each bid

- integrations@bidfluence.com

### DIFF
--- a/modules/bidfluenceBidAdapter.js
+++ b/modules/bidfluenceBidAdapter.js
@@ -36,67 +36,79 @@ export const spec = {
   },
 
   buildRequests: function (validBidRequests, bidderRequest) {
-    return validBidRequests.map(bidRequest => {
-      const params = bidRequest.params;
-      const sizes = utils.parseSizesInput(bidRequest.sizes)[0];
-      const width = sizes.split('x')[0];
-      const height = sizes.split('x')[1];
-      const refInfo = bidderRequest.refererInfo;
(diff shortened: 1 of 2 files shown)


francescocristallo:
@jsnellbaker 
The params are the same for both endpoints, I created a test page and used the v1 one and it works. Could you please post the call payload? It could be a security restriction blocking the call our side.

Thanks

jsnellbaker:
@francescocristallo Below is a copy of the request payload:
```
{"v":"2.0","azr":true,"ck":true,"re":"http://ap.localhost:9999/integrationExamples/gpt/hello_world.html?pbjs_debug=true","st":["http://ap.localhost:9999/integrationExamples/gpt/hello_world.html?pbjs_debug=true"],"tz":"300","sr":"1680x1050x24","tm":3000,"vp":"1682x348","sdt":"2018/12/05 20:01:47","top":true,"gdpr":false,"gdprc":"","bids":[{"bid":"2e8f27de25bc82","tid":"1000","pid":"1000","rp":0,"w":"300","h":"250"}]}
```

francescocristallo:
@jsnellbaker Thanks, as I thought the new endpoint was blocking the localhost address. I fixed it, now it returns the test Ad correctly.

jsnellbaker:
Thanks for the update; LGTM

PR #3352, opened 2018-12-05
do not throw a warning of conversion after re-submitting bid post-con…

dejanstrbac:
- [ x] Bugfix

When bid is converted from a source currency to that of adserver, the bid is re-submitted post-conversion. This however throws a warning of "overridden" currency which is incorrect. Unnecessary noise.

### DIFF
--- a/modules/currency.js
+++ b/modules/currency.js
@@ -139,7 +139,7 @@ export function addBidResponseHook(adUnitCode, bid, fn) {
   }
 
   let bidder = bid.bidderCode || bid.bidder;
-  if (bidderCurrencyDefault[bidder]) {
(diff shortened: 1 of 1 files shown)


dejanstrbac:
@jaiminpanchal I was referring to the bidResponseQueue - the way conversion works, it takes from the queue and if not in the adServer currency, it does the conversion and resubmits to the queue for repeated processing. The messages that comes out in that second run though is wrong.

dejanstrbac:
@jaiminpanchal27

jsnellbaker:
@dejanstrbac Do you have a page that demonstrates the current behavior so we can see it happening?

jsnellbaker:
@dejanstrbac Did you have the chance to see my previous comment?

PR #3378, opened 2018-12-14
New hooks API for 2.0

snapwich:
## Type of change
- [x] Feature

## Description of change
Building off the original hooking functionality I created for Prebid, I have created a new functional hooking library.

https://www.npmjs.com/package/fun-hooks

This has a more consistent API, better performance, and solves issues with the `sync` hook (which I don't think anyone used because I don't think it was implemented how it should have been), as well with the `asyncSeries` hooks with low priority (having a hook fire after the hooked function wasn't quite fleshed out).

The new library solves those problems but the API is slightly difference, now using `before` hooks and `after` hooks.  I think it would be great to use the library in Prebid.js but I'm considering the best way to accomplish it.  

~~This pull-request turns the internal `hook.js` code into an adapter for the new 'async' API in `fun-hooks` and just disables the `sync` API altogether (which I think is fine since it didn't work quite right). Ideally we would just replace all the code and refactor to the new API so we don't have the overhead of an adapter, but I'm slightly worried if someone is using this API outside of the Prebid.js repository it could potentially break their code.  So perhaps this adapter for now, and refactor the code as part of a breaking semver change later?~~ refactored for 2.0

## Other information
The `fun-hooks` library uses Proxies for some of its functionality.  This revealed a bug in `chai 3.x` where `expect.isFunction(<some proxified function>)` was false in Microsoft Edge.  To fix this, `chai` was updated to the latest version which required fixing some other tests (and revealed some other bugs in tests unrelated to this new feature).


snapwich:
Note to self, have to account for hooked functions that return values, like in: modules/pre1api.js:119

mkendall07:
Hey Rich, this library looks really cool! 
I showed this to @dbemiller and he thought that the design looked pretty decent as well. One thing he pointed out (that is also a problem with our limited hooks implementation in Prebid.js) is that the order in which hooks are added / executed is not guaranteed because the underlying order in which modules are loaded varies by implementations. 

A simple example of this would be 2 files that are importing a shared function (C), one which adds a hook and another which executes the function. Depending on the order that these modules load, the behavior of the function is mutated. 

While this sequence doesn't happen today in prebid.js (because of the queuing of commands) it's possible that it could occur in the future. One mitigation technique we thought about would be to have check to see if all the modules / core code is loaded before executing any hooked function. If it's not loaded completely then it could just error out and not call the `next` function. 

Thoughts?

snapwich:
So regarding hook ordering, the ordering is guaranteed if you add a `priority` when installing the hook.  That is currently the case in the limited hooks in Prebid.js as well as in this new library, and we use it with a few of the hooks that do request queueing.  If ordering doesn't matter then you don't specify a `priority` and it falls back to install order (I actually have a bunch of tests around this ordering in the new library, so ordering should always be consistent).

I'll think about the "ready" problem.  I think I just thought of a good configurable implementation that would work in the library.  When passing configuration to the factory that makes the hook creator function it might make sense to pass a configuration such as `expectsReady` and then a `hooks.ready()` needs to be called by the implementing application (prebid in this case) before any hooks can be called.  If `ready` hasn't been called it could throw an error or display a warning depending on whether we want to bail or just let people know of a possible error.

modules/consentManagement.js
```diff
@@ -194,7 +194,7 @@ export function requestBidsHook(fn, reqBidsConfigObj) {
   // preserves all module related variables for the current auction instance (used primiarily for concurrent auctions)
   const hookConfig = {
     context: this,
-    args: arguments,
+    args: [reqBidsConfigObj],
```
snapwich: @jsnellbaker does this fix the issue you referred to?

jsnellbaker: That does appear to also fix the issue.  Thanks.

snapwich:
@mkendall07 this has been updated with the `ready` functionality I added to `fun-hooks` described [here](https://github.com/snapwich/fun-hooks#ready).  `pbjs.processQueue` now calls the `hook.ready()` which enables all the hooked functions. Any `sync` hooked functions (which we don't have any of right now) called before being ready will `throw` and any `async` hooked functions called before the ready call will be queued and executed when `hook.ready()` is called.  This ensures all hooks installed by modules will exist and  be executed in the same order for all hooked function calls (even queued calls that are made before `hook.ready()`).

mkendall07:
awesome thanks for fixing that even though it wasn't a problem in PBJS.

PR #3380, opened 2018-12-14
Replace deprecated getWindowTopLocation with referrer detection module

ix-prebid-support:
## Type of change
- [x] Other

## Description of change
IX Adapter will start using the new referrer detection module to retrieve the referrer info


ix-prebid-support:
@mike-chowla 
Any update that can be shared with us?

PR #3387, opened 2018-12-17
Adding Remaining Native Parameters from AppNexus Native response

aneuway2:
## Type of change

- [x] Feature

## Description of change
Adds support for the following native parameters to prebid.js and the AppNexus adaptor:
- displayUrl
- address
- downloads
- likes
- phone
- price
- salePrice

Example implementation
```
var adUnits = [{
  code: 'div-id',
  mediaTypes: {
	  native: {
                displayUrl: {
                    required: true
                },
                address: {
                    required: true
                },
                downloads: {
                    required: true
                },
                likes: {
                    required: true
                },
                phone: {
                    required: true
                },
                price: {
                    required: true
                },
                salePrice: {
                    required: true
                }
	  }
  },
  bids: [...]
}];
```

### DIFF
--- a/modules/appnexusBidAdapter.js
+++ b/modules/appnexusBidAdapter.js
@@ -27,7 +27,9 @@ const NATIVE_MAPPING = {
     minimumParams: { sizes: [{}] },
   },
   sponsoredBy: 'sponsored_by',
-  privacyLink: 'privacy_link'
+  privacyLink: 'privacy_link',
+  sale: 'saleprice',
+  clickLabel: 'displayurl'
 };
 const SOURCE = 'pbjs';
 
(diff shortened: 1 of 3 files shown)


aneuway2:
*bump* (as i'm sure we were all just on holiday and this is not stale)

modules/appnexusBidAdapter.js
```diff
@@ -338,10 +340,18 @@ function newBid(serverBid, rtbBid, bidderRequest) {
...
+      likes: nativeAd.likes,
+      phone: nativeAd.phone,
+      price: nativeAd.price,
+      sale: nativeAd.saleprice,
       clickUrl: nativeAd.link.url,
+      clickLabel: nativeAd.displayurl,
       clickTrackers: nativeAd.link.click_trackers,
       impressionTrackers: nativeAd.impression_trackers,
       javascriptTrackers: nativeAd.javascript_trackers,
+      video: nativeAd.video.content
```
jsnellbaker: need to add a check to verify the `nativeAd.video` property exists before reading the `content` field.  This otherwise throws an `undefined` error (such as in the case of a standard native image ad).

the check could just be `video: nativeAd.video && nativeAd.video.content`

aneuway2: Thanks @jsnellbaker -- per internal conversations in slack i've dropped the video parameter entirely

test/spec/modules/appnexusBidAdapter_spec.js
```diff
@@ -554,6 +569,18 @@ describe('AppNexusAdapter', function () {
...
         'impression_trackers': ['http://example.com'],
+        'rating': '5',
+        'displayurl': 'http://AppNexus.com/?url=display_url',
+        'likes': '38908320',
+        'downloads': '874983',
+        'price': '9.99',
+        'saleprice': 'FREE',
+        'phone': '1234567890',
+        'address': '28 W 23rd St, New York, NY 10010',
+        'video': {
```
mkendall07: drop this since we removed it.

aneuway2: done @mkendall07

PR #3393, opened 2018-12-18
Fix #3265 - pbs cookie sync race condition

### LINKED ISSUE #3265: Prebid Server cookie sync race condition
## Type of issue
Bug

## Description

There's a race condition in the `prebidServerBidAdapter` cookie sync code: https://github.com/prebid/Prebid.js/blob/master/modules/prebidServerBidAdapter/index.js#L119

Prebid Server's /setuid endpoint behaves like so:

1. Unpack the Cookie sent in the request, to get existing user IDs.
2. Update the data inside the Cookie with the new Bidder/UID
3. Respond with a Set-Cookie header with the updated data.

Suppose two syncs get executed, and the second one gets sent before the first one returns. The same Cookie data will be sent to both. Both `/setuid` calls will respond with `Set-Cookie` headers, but neither one will contain the other one's data. After both execute, the final cookie will only have the updated ID from the _last_ `/setuid` response to return.

## Steps to reproduce

Since this is a race condition, it's inherently impossible to reproduce reliably. To increase the chances of reproducing it, run an auction with all the Prebid Server bidders. Use a `baset64` decoder to decode the PBS cookie, and see whether all the bidders have syncs there.

In practice, there's probably not much value in reproducing it. To fix it, just do your best to avoid running two Prebid Server cookie syncs at the same time.

jsnellbaker:
## Type of change

- [x] Bugfix

## Description of change
Fixes #3265

### DIFF
--- a/modules/prebidServerBidAdapter/index.js
+++ b/modules/prebidServerBidAdapter/index.js
@@ -116,7 +116,7 @@ function queueSync(bidderCodes, gdprConsent) {
     (response) => {
       try {
         response = JSON.parse(response);
-        response.bidder_status.forEach(bidder => doBidderSync(bidder.usersync.type, bidder.usersync.url, bidder.bidder));
+        doAllSyncs(response.bidder_status);
       } catch (e) {
         utils.logError(e);
       }
@@ -128,6 +128,19 @@ function queueSync(bidderCodes, gdprConsent) {
     });
 }
 
+function doAllSyncs(bidders) {
+  if (bidders.length === 0) {
+    return;
+  }
+
+  const thisSync = bidders.pop();
+  if (thisSync.no_cookie) {
+    doBidderSync(thisSync.usersync.type, thisSync.usersync.url, thisSync.bidder, doAllSyncs.bind(null, bidders));
+  } else {
+    doAllSyncs(bidders);
+  }
+}
+
 /**
  * Run a cookie sync for the given type, url, and bidder
  *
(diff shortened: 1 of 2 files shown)


modules/prebidServerBidAdapter/index.js
```diff
@@ -128,24 +128,39 @@ function queueSync(bidderCodes, gdprConsent) {
...
+
 /**
  * Run a cookie sync for the given type, url, and bidder
  *
  * @param {string} type the type of sync, "image", "redirect", "iframe"
  * @param {string} url the url to sync
  * @param {string} bidder name of bidder doing sync for
  */
-function doBidderSync(type, url, bidder) {
+function doBidderSync(type, url, bidder, done) {
```
mkendall07: can you update the JSdoc params for `done` ?

src/utils.js
```diff
@@ -616,11 +620,15 @@ exports.insertHtmlIntoIframe = function(htmlCode) {
  * @param  {string} url URL to be requested
  * @param  {string} encodeUri boolean if URL should be encoded before inserted. Defaults to true
  */
-exports.insertUserSyncIframe = function(url) {
+exports.insertUserSyncIframe = function(url, done) {
```
mkendall07: JS doc here too

jsnellbaker:
@mkendall07 Thanks for the feedback; I added the missing jsdoc comments.

PR #3371, opened 2018-12-12
calling setConfig multiple times on consentManagement causes requestBidsHook to be added each time

olafbuitelaar:
calling pbjs.setConfig with the consentManagement subscriber in the config struct;
invokes : $$PREBID_GLOBAL$$.requestBids.addHook(requestBidsHook, 50);
on each call, causing the consentManagement to be registered multiple times in the asyncSeries,
The causes strange sides effects while calling for bids, as the CMP reports to be timeouted.

For example;
pbjs.setConfig({consentManagement: {timeout:1}});
pbjs.setConfig({consentManagement: {timeout:2}});

I've added a simple boolean flag, to verify if the requestHook should be added or not.

PR #3399, opened 2018-12-19
Consent module doublecall

olafbuitelaar:
new PR for; https://github.com/prebid/Prebid.js/pull/3371

### DIFF
--- a/modules/consentManagement.js
+++ b/modules/consentManagement.js
@@ -217,9 +217,7 @@ export function requestBidsHook(reqBidsConfigObj, fn) {
     if (consentTimeout === 0) {
       processCmpData(undefined, hookConfig);
     } else {
-      //clearTimeout(hookConfig.timer);
       hookConfig.timer = setTimeout(cmpTimedOut.bind(null, hookConfig), consentTimeout);
-      utils.logInfo(`CMP framework setTimeout: ` + hookConfig.timer);
     }
   }
 }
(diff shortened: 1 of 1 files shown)


jsnellbaker:
@olafbuitelaar Can you fix the linting errors that were caught in the circleci check?

olafbuitelaar:
@jsnellbaker  done..all green now..shouldn't these formattings not being dealt with automatically?

jsnellbaker:
@olafbuitelaar Thanks for making the updates.  If your text editor has a plugin for eslint, it would be captured/handled while your writing the code the changes.  Depending on the setup of that plugin, when you save the file - it would automatically fix the linting errors during that save action.

Otherwise you'd need to run a `gulp` command to run the linting check.  You could use `gulp lint` directly, or you could run `gulp test` or `gulp serve` (if you were testing your changes locally) as both of those commands include the `gulp lint` as part of the sequence.

olafbuitelaar:
@jsnellbaker i'm using VS Code with the eslint plugin..maybe something is broken, i'll check it out..maybe have a: gulp format command? just as the beauty of go?

olafbuitelaar:
@idettman thanks. @jsnellbaker  just made a PR which appears to work for me.
As final note on this issue, i could imagine other modules might suffer from being double registered as listener, which probably isn't good in all cases.

PR #3402, opened 2018-12-20
Bid caching flag

### LINKED ISSUE #2992: Provide control for the publisher to turn off bid caching
## Type of issue
feature request

## Description
Though PBJS has a strictly-limited form of bid-caching, given the recent concern over bid caching in the industry, some publishers may wish to side-step any issues and turn the bid-reconsider feature off entirely.

Suggested implementation would be a call to `setConfig`.

snapwich:
## Type of change
- [x] Feature

## Description of change
This adds a new flag, `useBidCache`, for disabling [bid caching](http://prebid.org/dev-docs/faq.html#does-prebidjs-cache-bids).  The default for `useBidCache` is true.

```javascript
// disable bid caching
pbjs.setConfig({
  useBidCache: false
});
```

## Other information
Documentation pull-request here: https://github.com/prebid/prebid.github.io/pull/1069

fixes #2992

### DIFF
--- a/src/config.js
+++ b/src/config.js
@@ -18,6 +18,7 @@ const DEFAULT_BIDDER_TIMEOUT = 3000;
 const DEFAULT_PUBLISHER_DOMAIN = window.location.origin;
 const DEFAULT_ENABLE_SEND_ALL_BIDS = true;
 const DEFAULT_DISABLE_AJAX_TIMEOUT = false;
+const DEFAULT_BID_CACHE = true;
 
 const DEFAULT_TIMEOUTBUFFER = 400;
 
@@ -131,6 +132,14 @@ export function newConfig() {
       },
       set enableSendAllBids(val) {
         this._sendAllBids = val;
+      },
+
+      _useBidCache: DEFAULT_BID_CACHE,
+      get useBidCache() {
+        return this._useBidCache;
+      },
+      set useBidderCache(val) {
+        this._useBidCache = val;
       },
 
       _bidderSequence: DEFAULT_BIDDER_SEQUENCE,
--- a/src/config.js
+++ b/src/config.js
@@ -138,7 +138,7 @@ export function newConfig() {
       get useBidCache() {
         return this._useBidCache;
       },
-      set useBidderCache(val) {
+      set useBidCache(val) {
         this._useBidCache = val;
       },
 
--- a/src/auctionManager.js
+++ b/src/auctionManager.js
@@ -57,9 +57,6 @@ export function newAuctionManager() {
   };
 
   auctionManager.getBidsReceived = function() {
-    // As of now, an old bid which is not used in auction 1 can be used in auction n.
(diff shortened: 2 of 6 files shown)


src/config.js
```diff
@@ -18,6 +18,7 @@ const DEFAULT_BIDDER_TIMEOUT = 3000;
 const DEFAULT_PUBLISHER_DOMAIN = window.location.origin;
 const DEFAULT_ENABLE_SEND_ALL_BIDS = true;
 const DEFAULT_DISABLE_AJAX_TIMEOUT = false;
+const DEFAULT_BID_CACHE = true;
```
bretg: The Prebid Board would like the default to be "no cache", so that would change this default to `false`, right?

snapwich: In your e-mail you said the board wanted the ability to "opt-out".  If I changed the default to `false` wouldn't that be "opt-in" in regards to bid caching?

mike-chowla: My recollection of the board discussion was publishers would have to opt-in and we'd disable by default

mkendall07: believe we decided to add the flag as the 1st step, then as a 2nd step making it off by default (breaking change).

snapwich: @mkendall07 that's what I recall as well, which is how it is implemented now.  Also, switching it to off will require some effort more than just changing that constant. Since many tests related to targeting assume bid caching, they will have to be updated to behave appropriately.  

@bretg let me know if this is okay as is or if we want to further discuss switching the default.

bretg: The Board wants to have bid caching off-by-default. If we do this step-wise approach, I want it to be clear to the community that's where we're going... i.e. give everyone a release or two to react to this change and then change the default by the end of January.

This is what I had said in the email:

> The Prebid board discussed again today and reiterated a desire to retain "opt-out" ability. The plan is to get Publishers on board to Prebid.org and get their direct input, but for now opt-out is an aggressive path forward that still provides a path for those with special requirements.

bretg: I updated the docs PR with a warning that the default will be changing in the coming releases.

mkendall07: Then we ship it as a 2.0. We are not making breaking changes without major version changes period.

harpere: So sounds like this is the plan, right?
**1.x** - leave bid-caching on by default, but add a config flag that will allow a publisher to turn it off
**2.0** - turn bid-caching off by default and let pubs turn it on with a config flag.

mkendall07: correct, thanks. @bretg if you have reason to push this out sooner than an expected release for 2.0, we should talk about how to handle. I.E if we make it the only change in 2.0 that's probably OK but it creates additional overhead to support legacy for 30 days per our policy (vs releasing a bigger 2.0 release)

bretg: The board wants this change done 3 months ago.

So if we want to increment the version as a result of this change, so be it. Let's plan for a 2.0 by the end of the month then?

PR #3392, opened 2018-12-18
Rad 2356 upgrade mocha

deekshithraop:
## Type of change

- [ ] Bugfix
- [ ] Feature
- [ ] New bidder adapter   
- [ ] Code style update (formatting, local variables)
- [ ] Refactoring (no functional changes, no api changes)
- [ ✓] Build related changes
- [ ] CI related changes
- [ NO ] Does this change affect user-facing APIs or examples documented on http://prebid.org?
- [ ] Other

## Description of change

- Update Mocha to v5, karma to v3, and karma-webpack to v3

## Changes in Node version support

- Since latest versions of those can only be run successfully in latest versions of node,  We will be supporting node of version v6 or more.

PR #3405, opened 2018-12-21
Upgrade packages

deekshithraop:
## Type of change
- [ ] Build related changes

## Description of change

Removing packages with vulnerabilities 

on running npm audit package "open" had critical vulnerabilities its being used by by us and by gulp-webserver as its dependency. 

> so moved back to gulp-connect.
> and also converted from "open" to" opn" package which works exactly the same but has no security issues.

## Changes in Node version support
Since latest versions of those can only be run successfully in latest versions of node, We will be supporting node of version v6 or more.

Note: this pull req includes the changes in pull req (https://github.com/prebid/Prebid.js/pull/3392) which is to update mocha, karma, karma-webpack


gulpfile.js
```diff
@@ -269,6 +268,25 @@ function coveralls() { // 2nd arg is a dependency: 'test' must be finished
     .pipe(shell('cat build/coverage/lcov.info | node_modules/coveralls/bin/coveralls.js'));
 }
 
+gulp.task('watch', function () {
```
jsnellbaker: this gulp task is redundant and can be removed; there is already a `watch` function declared earlier in the file (see [here]( https://github.com/prebid/Prebid.js/blob/master/gulpfile.js#L106)).

karma.conf.maker.js
```diff
@@ -34,7 +34,6 @@ function newPluginsArray(browserstack) {
     'karma-es5-shim',
     'karma-mocha',
     'karma-chai',
-    'karma-requirejs',
```
jsnellbaker: If you're removing `karma-requirejs` plug-in from this file, can it also be removed from the `package.json`?  I don't see it referenced anywhere else in the project.

deekshithraop: yes, removed

gulpfile.js
deekshithraop: yes, removed

PR #3407, opened 2018-12-24
OTM adapter update

fedotxxl:
Modify adapter to support migration to a new ssp


bretg:
docs PR at https://github.com/prebid/Prebid.js/pull/3407

PR #3412, opened 2019-01-02
make eslint fix the formatting

olafbuitelaar:
as mentioned here https://github.com/prebid/Prebid.js/pull/3399, the fix attribute only manipulates the stream. 
according the doc this is the way to modify the src;
- https://github.com/adametry/gulp-eslint/blob/master/example/fix.js
- https://github.com/gulpjs/gulp/issues/267
and should auto format all code..which would mean, that if you don't want this special eslint comments should be used!


olafbuitelaar:
This can be closed created a new PR with correct commits

PR #3414, opened 2019-01-02
changing default currency file

bretg:
## Type of change
- [ X ] Does this change affect user-facing APIs or examples documented on http://prebid.org?
- [ X ] Other

## Description of change

Changing the default currency file location to https://cdn.jsdelivr.net/gh/prebid/currency-file@1/latest.json

## Other information

Docs PR at https://github.com/prebid/prebid.github.io/pull/1076

### DIFF
--- a/modules/currency.js
+++ b/modules/currency.js
@@ -74,26 +74,25 @@ export function setConfig(config) {
       utils.logInfo('currency using override conversionRateFile:', config.conversionRateFile);
       url = config.conversionRateFile;
     }
-    
+
     // see if the url contains a date macro
     // this is a workaround to the fact that jsdelivr doesn't currently support setting a 24-hour HTTP cache header
     // So this is an approach to let the browser cache a copy of the file each day
     // We should remove the macro once the CDN support a day-level HTTP cache setting
-    const macroLocation=url.indexOf("$$TODAY$$");
-    if (macroLocation>=0) {
+    const macroLocation = url.indexOf('$$TODAY$$');
+    if (macroLocation !== -1) {
       // get the date to resolve the macro
-      var d = new Date(),
-       month = '' + (d.getMonth() + 1),
-       day = '' + d.getDate(),
-       year = d.getFullYear();
+      const d = new Date();
+      let month = '' + (d.getMonth() + 1);
+      let day = '' + d.getDate();
       if (month.length < 2) month = '0' + month;
       if (day.length < 2) day = '0' + day;
-      const todaysDate = [year, month, day].join('');
-      
+      const todaysDate = [d.getFullYear(), month, day].join('');
+
       // replace $$TODAY$$ with todaysDate
-      url=url.substring(0,macroLocation)+todaysDate+url.substring(macroLocation+9,url.length);
+      url = url.substring(0, macroLocation) + todaysDate + url.substring(macroLocation + 9, url.length);
     }
-    
+
     initCurrency(url);
   } else {
     // currency support is disabled, setting defaults
--- a/modules/currency.js
+++ b/modules/currency.js
@@ -5,7 +5,7 @@ import * as utils from 'src/utils';
 import { config } from 'src/config';
 import { hooks } from 'src/hook.js';
 
-const DEFAULT_CURRENCY_RATE_URL = 'https://cdn.jsdelivr.net/gh/prebid/currency-file@1/latest.json';
+const DEFAULT_CURRENCY_RATE_URL = 'https://cdn.jsdelivr.net/gh/prebid/currency-file@1/latest.json?date=$$TODAY$$';
 const CURRENCY_RATE_PRECISION = 4;
 
 var bidResponseQueue = [];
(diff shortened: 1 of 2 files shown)


bretg:
Also improved the handling of `defaultRates` -- @idettman has offered to add a unit test

The currency ajax call doesn't have a timeout. The only events that cause the `bidResponseQueue` (i.e. bidsToBeConverted) to be processed are:

1) Successful currency file AJAX response before timeout
2) Bid response after currency file AJAX response (either successful or error)

I'm assuming that the retrieval of the currency conversion file is going to be fast because it's a small flat file on a CDN, so I think this behavior is ok. i.e. A huge percentage of the time the file should be loaded before the first bidResponse.

snapwich:
@mkendall07 have we confirmed with jsdelivr that they are willing to accept this traffic load?  It's probably going to be quite a bit of traffic and is outside their normal use case of serving libraries.

mkendall07:
Bret did ask about the traffic being an issue and it didn't seem like it was. With that said, if the "rank" here indicates the overall highest traffic for jsdelivr, we'll probably very easily become the top requested resource (https://data.jsdelivr.com/v1/package/npm/jquery/stats)

snapwich:
That's what I'm worried about.  We're definitely going to be a noticeable increase to their traffic, so unless we have _very explicit_ permission for what we're doing we'll most likely get removed quickly.  Getting removed could be a big issue since we're hard-coding this endpoint into the currency module.  If we have explicit permission though, then great.

mkendall07:
@bretg 
Might be worth double checking with jsdelivr on the number of requests for this file

idettman:
Holding off merging until we double check with jsdelivr supporting our expected file traffic rates for the currency file

modules/currency.js
```diff
@@ -5,7 +5,7 @@ import * as utils from 'src/utils';
 import { config } from 'src/config';
 import { hooks } from 'src/hook.js';
 
-const DEFAULT_CURRENCY_RATE_URL = 'https://currency.prebid.org/latest.json';
+const DEFAULT_CURRENCY_RATE_URL = 'https://cdn.jsdelivr.net/gh/prebid/currency-file@1/latest.json';
```
snapwich: How are we updating this file?  In `npm` it's impossible to republish a given library + version, does jsdelivr not have this restriction, and if not, are we sure they won't have this restriction in the future?

Also, do we have any control over the cache control headers?  Currently jsdelivr seems to be setting it to a week, which isn't very good for a currency conversion file.

mkendall07: Source of truth is github, with a new git tag pushed daily via a automated job. We also have access to purge the cache on jsdelivr servers so the cache header shouldn't be an issue. You can see the repo here: https://github.com/prebid/currency-file/releases

bretg: Good observations @snapwich, but we don't have control over cache control headers with jsdelivr.

Unclear that purging the CDN cache daily will help if the browser's allowed to cache it for a week. I did not do a full analysis of all CDN options, though did ask the community for options to which no one responded.

I will do some quick pricing homework to see if there other options. If not, then 1-week caching is what we get.

> it's impossible to republish a given library + version

We set the version of each new file to 1.SEQUENCE+1, and jsdeliver has the '@1' shortcut to take the latest version of 1.x.

bretg: Ok - the guy at jsdelivr wants to know what headers we want to set. I think it's just:
```
Cache-Control: max-age=86400
```
@snapwich, @mkendall07 - agree?

Mozilla.org (https://developer.mozilla.org/en-US/docs/Web/HTTP/Caching) suggest that `Cache-Control: max-age` is preferred over `Expires`.

mkendall07: `Cache-Control: max-age=86400` looks reasonable to me. Note that this will cache a bit longer than we want - it is relative to the original request - meaning that if you first requested at 10pm than it won't expire until the next day at 10pm. I think it's ok for our use case though.

snapwich: We were using `expires` as an optimization trying to time right after the script updated.  But if we don't have dynamic control over the time then a day is probably okay; worst case scenario someone gets 2 day old data.

bretg: I've asked jsdelivr if they can add `max-age`.

Also did some pricing from another CDN, which isn't much better than our current one.

bretg: So apparently they can't add `max-age`. His proposed workaround is to add a cache-buster.

e.g. https://cdn.jsdelivr.net/gh/prebid/currency-file@1/latest.json?date=TODAY

We would modify the module to dynamically replace TODAY with YYYYMMDD, with the effect that each morning the user's browser would see a new query string, causing it to hit the CDN again. Hack, but effective.

So we have a decision:
1) Settle for the one week cache period
2) Implement the daily query string cache buster
3) Keep paying many thousands of dollars per month for hosting this file

mkendall07: Can they add `expires` ? 

My vote would be to settle 1 week cache period. This would encourage users to self host if they want more accuracy. The query string would be effective but it's an ugly hack IMO.

snapwich: I don't mind the cache buster, although I've heard mixed results about query string cache busters (vs filename cache busting).

bretg: He said they're planning on adding cache controls in the future including `Expires`, but did not have a commit date.

Agree that query string is a hack. But it's a hack that preserves existing functionality and saves quite a lot of money, so my vote would be we start off with it as a low-cost workaround to not change the file's loading behavior -- then remove when unnecessary. Am prepared to be outvoted.

mkendall07: The promise of future support for cache and/or expires is good enough for me. @bretg can you make the query string change then?

bretg: Done - @idettman has offered to make a unit test.

idettman:
unit tests are ready for review.

test/spec/modules/currency_spec.js
```diff
@@ -30,6 +32,16 @@ describe('currency', function () {
   });
 
   describe('setConfig', function () {
+    beforeEach(function() {
+      sandbox = sinon.sandbox.create();
+      clock = sinon.useFakeTimers(1047010195974);
```
mkendall07: nice usage of sinon!

modules/currency.js
```diff
@@ -70,6 +74,25 @@ export function setConfig(config) {
...
+      // get the date to resolve the macro
+      const d = new Date();
+      let month = '' + (d.getMonth() + 1);
+      let day = '' + d.getDate();
+      if (month.length < 2) month = '0' + month;
+      if (day.length < 2) day = '0' + day;
+      const todaysDate = [d.getFullYear(), month, day].join('');
+
+      // replace $$TODAY$$ with todaysDate
+      url = url.substring(0, macroLocation) + todaysDate + url.substring(macroLocation + 9, url.length);
```
mkendall07: this is a nit but we usually prefer string literals now.

idettman: I committed an update with all the concatenations changed to string literals.

PR #3418, opened 2019-01-04
fix spacing in gulp lint command

jsnellbaker:
## Type of change

## Description of change
Fix the small formatting problem introduced from #3416 in the `gulp lint` command.


olafbuitelaar:
yeah that was bugging me too! ;)

Issue #3373, opened 2018-12-13
Teads: Adapter doesn't respect media types - sizes don't work


muuki88:
## Type of issue

bug

## Description

The [Teads Adapter](https://github.com/prebid/Prebid.js/blob/master/modules/teadsBidAdapter.js#L119), still uses the deprecated `bid.sizes` property and doesn't differentiate between `banner` and `video` media types. 

This 

cc: @valsouche 

## Steps to reproduce

Use a `video` only ad unit. See #3222

### Expected results

Teads should use the `mediaTypes` sizes for video and/or banner instead of the deprecated `sizes` property.

## Platform details

- prebid version `1.35.0`

## Other information

- #3222 - this issue already mentioned the problem, but didn't get fixed
- #3274 - IMHO due to this, teads somehow worked before 1.33.0

muuki88:
This is especially weird as the `bid.sizes` property should include the videos sizes. At least if I read [this code in der `adapterManager.js`](https://github.com/prebid/Prebid.js/blob/master/src/adaptermanager.js#L280-L295) correctly.

It almost looks as if the video player sizes override the banner.

The [OpenX adapter handles the video playerSize explicitly](https://github.com/prebid/Prebid.js/blob/27107e27911861af2235c2efcc4f029fb5e2c1ab/modules/openxBidAdapter.js#L313-L327).

muuki88:
The reasons for this behavior is buried in the [`getBids` function, which simply used the `mediaTypes.banner.sizes`](https://github.com/prebid/Prebid.js/blob/7fe95d89464cbc95adf3446cd8e39bb499f2c321/src/adaptermanager.js#L97) and discards everything that has been done before.

muuki88:
The [getBids function](https://github.com/prebid/Prebid.js/blob/master/src/adaptermanager.js#L97) seems to set the sizes only from `mediaTypes.banner.sizes` and nothing else. Is the size mapping even applied there?

valsouche:
In this pull request (https://github.com/prebid/Prebid.js/pull/3377), @bretg says that playerSize is an optional field. But like @muuki88 said, for some reason, if from a video adunit mediatypes there is no `playerSize` bid sizes =`[]`, even if there is a sizes property on the adUnit.

It looks like only banner sizes are copy in adUnit.sizes.

Is it something expected ? 

Thanks in advance,
Val.

muuki88:
Before this ticket is getting stale.

[...]

PR #3423, opened 2019-01-07
Teads adapter: handle video playerSize

valsouche:
## Type of change

- [x] Bugfix

## Description of change

Change the way to get sizes. Now we check if playerSize is available before others.
Add tests about sizes.

```
{
  bidder: 'teads',
  params: {
    pageId: 89489,
    placementId: 96888
  }
}
```

### DIFF
--- a/modules/teadsBidAdapter.js
+++ b/modules/teadsBidAdapter.js
@@ -129,7 +129,9 @@ function buildRequestObject(bid) {
 
 function getSizes(bid) {
   let sizes;
-  if (isVideoBid(bid)) {
+  if (isHybridBid(bid)) {
+    sizes = concatHybridBidSizes(bid);
+  } else if (isVideoBid(bid)) {
     sizes = utils.deepAccess(bid, 'mediaTypes.video.playerSize') ||
       utils.deepAccess(bid, 'mediaTypes.video.sizes') ||
       bid.sizes;
@@ -142,6 +144,36 @@ function getSizes(bid) {
 
 function isVideoBid(bid) {
   return bid.mediaType === 'video' || utils.deepAccess(bid, 'mediaTypes.video');
+}
+
+function isHybridBid(bid) {
+  return isVideoBid(bid) && utils.deepAccess(bid, 'mediaTypes.banner');
+}
+
+function concatHybridBidSizes(bid) {
+  let playerSize = utils.deepAccess(bid, 'mediaTypes.video.playerSize');
+  let videoSizes = utils.deepAccess(bid, 'mediaTypes.video.sizes');
+  let bannerSizes = utils.deepAccess(bid, 'mediaTypes.banner.sizes');
+  let sizes;
+
+  if (utils.isArray(bannerSizes) || utils.isArray(playerSize) || utils.isArray(videoSizes)) {
+    let mediaTypesSizes = [bannerSizes, videoSizes, playerSize];
+    sizes = mediaTypesSizes
+      .reduce(function(acc, currSize) {
+        if (utils.isArray(currSize)) {
+          if (utils.isArray(currSize[0])) {
+            currSize.forEach(function (childSize) { acc.push(childSize) })
+          } else {
+            acc.push(currSize);
+          }
+        }
+        return acc;
+      }, [])
+  } else {
+    sizes = bid.sizes;
+  }
+
+  return sizes;
 }
 
 function _validateId(id) {
--- a/modules/teadsBidAdapter.js
+++ b/modules/teadsBidAdapter.js
@@ -128,37 +128,17 @@ function buildRequestObject(bid) {
 }
 
 function getSizes(bid) {
-  let sizes;
-  if (isHybridBid(bid)) {
-    sizes = concatHybridBidSizes(bid);
-  } else if (isVideoBid(bid)) {
-    sizes = utils.deepAccess(bid, 'mediaTypes.video.playerSize') ||
-      utils.deepAccess(bid, 'mediaTypes.video.sizes') ||
-      bid.sizes;
-  } else {
-    sizes = utils.deepAccess(bid, 'mediaTypes.banner.sizes') || bid.sizes;
-  }
-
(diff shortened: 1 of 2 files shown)


muuki88:
Fixed #3373

modules/teadsBidAdapter.js
```diff
@@ -127,6 +127,23 @@ function buildRequestObject(bid) {
   return reqObj;
 }
 
+function getSizes(bid) {
+  let sizes;
+  if (isVideoBid(bid)) {
+    sizes = utils.deepAccess(bid, 'mediaTypes.video.playerSize') ||
+      utils.deepAccess(bid, 'mediaTypes.video.sizes') ||
```
muuki88: I'm still not convinced that `mediaTypes.video.sizes` is a supported property :joy:

valsouche: PlayerSize is optional, bid.sizes is the default and as you we are not sure about `mediaTypes.video.sizes` (We didn't find any line code about this property in the sources) but there is an adUnit section in the Prebid documentation which makes us doubt (see my screen). That is the reason why we preferred to add this check anyway... 
capture d ecran 2019-01-08 a 14 45 49

test/spec/modules/teadsBidAdapter_spec.js
```diff
@@ -196,7 +197,50 @@ describe('teadsBidAdapter', function() {
       expect(payload.gdpr_iab).to.exist;
       expect(payload.gdpr_iab.consent).to.equal(consentString);
       expect(payload.gdpr_iab.status).to.equal(0);
-    })
+    });
+
+    it('should use good mediaTypes video playerSizes', function() {
+      const mediaTypesPlayerSize = {
```
robertrmartinez: Also just want to confirm:

If a publisher requests a multi-format adUnit with banner and video, teads will default to Video and try to get the sizes from the video object.

Just want to confirm this is intended and there is no logic necessary for teads' approach to multiformat requests.

From a glance it does not look like mediaType is passed into the Teads Ad request anyways, but want to be sure that in this scenario, the behavior is as expected!

robertrmartinez: ex:

```
{
   mediaTypes: {
      video: {
         playerSize: [400, 600]
      }
      banner: {
         sizes: [[300, 250], [728, 90]]
      }
}
```

Some adapters do a multi-format call with both, some default to just video, some to banner.

modules/teadsBidAdapter.js
muuki88: Thanks for clarifying this. Even js is untyped, I pretend it is :wink: and try to avoid ambiguous variables. This fix was necessary as some internal behavior of `adunit.sizes` changed and the `playerSize` was not used anymore. That's why highlighted this in this PR.

When I searched the code for `sizes` related functions, I got really confused as there are multiple representations (as you mentioned single-size array and array of size arrays, but also the `<witdh>x<height>` arrays) and various parsing methods.

Hope this implementation will last :heart:  Thanks @valsouche for taking care :kissing_heart:

test/spec/modules/teadsBidAdapter_spec.js
muuki88: Very good point! Teads does offer some special formats like parallax ads, which may come with another size. Should `video` and `banner` sizes simply be concatenated?

valsouche: There is only one kind of response from Teads. It means, banner or video will be the same for us. We check if it's video juste to get sizes properly. 
Looks ok for us 😀 
Thanks !

robertrmartinez: That is hard to say without knowing the Teads Ad Exchange or what they tell customers about their implementation.

I think coming up with what the Teads approach to multi-format adUnits is would be a good start.

Maybe concatenating is the solution, maybe having one of the mediaTypes be higher priority (as video is in this current implementation)

Not sure!

valsouche: Good point, 
we prefer receive all sizes, so we will concatenate. I'll push the update soon. Thanks

valsouche: Done. :) Don't hesitate to ask if you find something weird.
Thanks !

modules/teadsBidAdapter.js
```diff
@@ -129,7 +129,9 @@ function buildRequestObject(bid) {
 
 function getSizes(bid) {
   let sizes;
-  if (isVideoBid(bid)) {
+  if (isHybridBid(bid)) {
+    sizes = concatHybridBidSizes(bid);
+  } else if (isVideoBid(bid)) {
```
robertrmartinez: Have a couple comments / suggestions:

With this current implementation a multi-format ad unit will add `video.sizes` and `video.playerSize`.

But if it is _just_ a video mediaType, then it picks `video.playerSize` OR `video.sizes`. Not sure if it is better to remain consistent and just concatenate in this scenario as well?

Or maybe in `concatHybridBidSizes` add the logic to choose `video.playerSize` OR `video.sizes`.

Also, instead of checking the `isHybridBid` or `isVideoBid` => `else`, does it make sense to just use the `concatHybridBidSizes` and not do any mediaTypes checking?

Since you are using `utils.deepAccess` to get the mediaTypes Sizes, then if the mediaType is not present, it will do what is needed anyways right?

so `getSizes` could just be something like:

```javascript
function getSizes(bid) {
   return utils.parseSizesInput(concatHybridBidSizes(bid));
}
```

And if needed add the `video.playerSize` || `video.sizes` logic, if that is the intended implementation.

robertrmartinez: CC @muuki88 for some input as well

valsouche: Exact ! I did the change, simpler and cleaner :)

muuki88: Makes sense :smiley:

PR #3427, opened 2019-01-08
RVR-2177 - Refactor events handling (#7)

AlessandroDG:
## Type of change

- [ ] Bugfix
- [x] Feature
- [ ] New bidder adapter   
- [ ] Build related changes
- [ ] CI related changes
- [ ] Other

## Description of change

Passing all the events to the external module rivraddon

### DIFF
--- a/modules/rivrAnalyticsAdapter.js
+++ b/modules/rivrAnalyticsAdapter.js
@@ -1,31 +1,15 @@
-import {ajax} from '../src/ajax';
-import adapter from '../src/AnalyticsAdapter';
-import CONSTANTS from '../src/constants.json';
-import adapterManager from '../src/adapterManager';
-import * as utils from '../src/utils';
+import {ajax} from 'src/ajax';
+import adapter from 'src/AnalyticsAdapter';
+import adaptermanager from 'src/adaptermanager';
+import * as utils from 'src/utils';
 
 const analyticsType = 'endpoint';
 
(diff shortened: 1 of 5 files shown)


mike-chowla:
@AlessandroDG I'm not sure why `addAdUnits` isn't available as an Analytics trackable event.  I'm guessing it's because no one thought it was needed.   What do you want to use it for that `auctionInit` isn't adequate?

The approach of using `pbjsGlobal.onEvent()` isn't a good one because the events you get from it are not considered a stable interface.  `AnalyticsAdapter.track()` is a stable interface that won't have a breaking change without an increment in the major version number.

I'm thinking the better approach is to add `addAdUnits` as an analytics trackable event assuming your use case can't be satisfied with the existing events.

AlessandroDG:
@mike-chowla Thank you for your reply.

With `addAdUnits` we want to track any change on the `adUnits` object between the moment a new value is assigned to it and the moment the auction is initialized.

It would be great to have it in `AnalyticsAdapter.track()`.

mike-chowla:
I don't think I was clear on my suggestion here which is for you to create a PR that adds addAdUnits to AnalyticsAdapter.track() and submit it

AlessandroDG:
@mike-chowla ok I am on it now. Also, apparently I didn't get notifications about your last message.

AlessandroDG:
@mike-chowla can I ask you for a feedbacks on my latest changes?

PR #3431, opened 2019-01-10
Add bidfloor param to allow dynamic bid floor CPM

jddeleon:
## Type of change

- [ ] Bugfix
- [x] Feature
- [ ] New bidder adapter   
- [ ] Build related changes
- [ ] CI related changes
- [ ] Other

## Description of change
Adds support to allow user to specify a minimum bid floor price by adding an optional `bidfloor` param. If `bidfloor` is not specified, it will default to the values that are specified when setting up a GumGum publisher account.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,6 +1,6 @@
 {
   "name": "prebid.js",
-  "version": "1.36.0-pre",
+  "version": "1.37.0",
   "lockfileVersion": 1,
   "requires": true,
   "dependencies": {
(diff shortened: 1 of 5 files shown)


package-lock.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "prebid.js",
-  "version": "1.36.0-pre",
+  "version": "1.37.0",
```
mkendall07: please drop the changes to package.lock

jddeleon: Those changes have been reverted!

PR #3434, opened 2019-01-10
New Orbidder Prebid Adapter

hiseke:
## Type of change
- [x] New bidder adapter   

```
{
  bidder: 'orbidder'
}
```

- contact email of the adapter’s maintainer: orbidder@otto.de

- Link to PR on the docs repo: https://github.com/prebid/prebid.github.io/pull/1081


modules/orbidderBidAdapter.js
```diff
@@ -0,0 +1,77 @@
...
+      });
+    }
+    return bidResponses;
+  },
+
+  onBidWon(winObj) {
+    const getRefererInfo = detectReferer(window);
+    const refererInfo = getRefererInfo();
+    winObj.pageUrl = refererInfo.referer;
+    ajax(`${this.orbidderHost}/win`, null, JSON.stringify(winObj));
```
idettman: it's recommended to use user syncs instead of ajax calls in bid adapters

hiseke: I got the basic idea from the [adspendBidAdapter](https://github.com/prebid/Prebid.js/blob/master/modules/adspendBidAdapter.js#L33) and I need to submit the json of the onBidWon object via post. Is this possible using a user sync? And I don't quite see how the user sync and the onBidWon callback function are supposed to play together. Any help would be appreciated :).

Issue #3379, opened 2018-12-14
[I2I] Prebid support for long form video


mkendall07:
## Type of issue
Intent to implement.
**Status:** Open for comment. 

# Table of Contents
* [Background](#background)
* [Goals](#goals)
* [Requirements](#requirements)
* [Proposed Design](#design)
  * [Architecture Diagram](#diagram)
  * [AdUnit Definition for Video](#adunit)
  * [Bidder Spec changes](#bidder-spec)
  * [Use Cases](#use-cases)
  * [FreeWheel Module](#freewheel)
  * [Content Category Translation Module](#translation)
  * [Prebid Cache Updates](#pbc)
  
 

# Background <a name='background'></a>
Today, premium video publishers do not have an effective solution to monetize long-form video content through programmatic partners. In most cases, programmatic partners are included via tag-based waterfalls in the primary ad server. These waterfalls are complex, detrimental to user experience, and suboptimal for monetization by functioning solely on remnant inventory and by limiting each demand partner's ability to see and bid on video impressions. Many premium video publishers are unwilling even to implement programmatic partners via waterfall, as there is no way to ensure that competitive separation rules will be enforced.

Prebid.js can provide a better solution for premium video publishers.

# Goals <a name='goals'></a>
* Enable premium video publishers to unlock header bidding demand on long-form content on web and mobile-web.
* Ensure that header bidding demand can be served by the primary ad server without violating competitive separation rules.

# Requirements <a name='requirements'></a>
* Prebid.js must provide an API to configure Ad Pod which allows a publisher to define the characteristics of their inventory
* Publishers must be able to get a list of key-values from Prebid.js that they will add to their Primary ad server's ad request URL
* Prebid.js must provide a mechanism to ensure that header bidding demand respect any competitive separation rules defined through FreeWheel
* Prebid.js must make sure that demand partners return a exactly one IAB category on each bid response.
* Prebid.js must provide a module for converting IAB category to FreeWheel Industry Group.

# Proposed Design <a name='design'></a>

## Architecture Diagram <a name='diagram'></a>
In Phase 1, We are using FreeWheel as Ad Server
![alt text](http://i.imgur.com/1UtKdSX.png "Logo Title Text 1") 

##### Steps
1. Prebid makes a bid request to configured demand partners
2. Store bids in Prebid Cache.
3. Send targeting kv pairs to Freewheel SDK. 
```JavaScript
[
  {
    hb_pb_industry_duration: 10.00_airline_15s,
[...]

PR #3440, opened 2019-01-10
generate bid.adId uniquely instead of using bidRequest.bidId

jsnellbaker:
## Type of change
- [x] Other

## Description of change
This PR modifies the manner in which `bid.adId` is generated in `bidfactory.js`.  Previously it was directly copied from the `bidRequest.bidId` and if that wasn't available a new string was generated.  

The new logic is:
- we always generate a new string value for the `bid.adId` 
- the `bidRequest.bidId` is set to the `bid.requestId`

## Reason for change
The reason for this change is to help allow multiple bid responses be associated to the same bid object.  When the `bid.adId` was set equal to the `bidRequest.bidId`, it implied a 1:1 relationship between the two objects.  Given some of the upcoming changes related to https://github.com/prebid/Prebid.js/issues/3379, we want prebid be able to support having multiple bid responses associated to a single `bidRequest.bids` object.  While the rest of core would support this type of association, the `bid.adId` would be copied into the multiple `bid` response objects, which would cause issues when trying to render the bid (as it would grab the first bid that matched rather than the bid that won).  By generating unique strings, it addresses this problem.

## Additional context/information
The `bid.requestId` is meant to be the tie between the `bid` response object and the  `bidRequest.bids` object.  The `bid.requestId` is a typically populated from adapter files when they're creating the initial `bid` object (ie when they're reading their adserver's bid response).  So the `bid.requestId` we're setting in the `bidfactory.js` is meant to be a back-up in the rare case some adapters don't set the value (like in `serverbidServerBidAdapter.js`).  

The changes to the function calls for `getBidRequest()` are to line up with the above logic that `bid.requestId` is the lookup field.

The change in the `getBidRequest()` function is to prevent some false positive results for the unit tests; an `undefined` `id` value would match the inner `some` check if one of those sub-fields (eg `bid_id`) didn't exist in the sample/test request object.


src/bidfactory.js
```diff
@@ -15,15 +15,15 @@ var utils = require('./utils.js');
...
   var _bidSrc = (bidRequest && bidRequest.src) || 'client';
   var _statusCode = statusCode || 0;
 
   this.bidderCode = (bidRequest && bidRequest.bidder) || '';
   this.width = 0;
   this.height = 0;
   this.statusMessage = _getStatus();
-  this.adId = _bidId;
+  this.adId = utils.getUniqueIdentifierStr();
+  this.requestId = bidRequest && bidRequest.bidId;
```
harpere: I noticed that the currency file calls [bidfactory.createBid()](https://github.com/prebid/Prebid.js/blob/master/modules/currency.js#L223-L226) when conversions fail.  There might be some (minor) repercussions since the `adId` will no longer be maintained in the new bid.  Might want to at least change [this line in the currency module](https://github.com/prebid/Prebid.js/blob/master/modules/currency.js#L225), from `bidId: bid.adId` to `bidId: bid.requestId` to maintain the requestId - though that isn't maintained currently.

jsnellbaker: Thanks for the feedback regarding the `currency` file.  I pushed an update for that reference.

jsnellbaker:
As an update in regards to previous comment - based on some internal discussion, we will not be limiting this to only adpod (formerly longform) video mediaTypes.  

A 1:1 relationship between request.bids and the bid responses was not a strictly enforced requirement from the code-perspective; so we're not going enforce it now as part of this change.

PR #3444, opened 2019-01-11
Adtelligent support both userSync types

GeneGenie:
## Type of change
- [x] Feature

## Description of change
Adapter response may provide sync type.

- g.moroz@adtelligent.com

PS: Tests updated


GeneGenie:
@bretg Yes, your comment makes perfect sense. I think this change was requested to avoid pre-integration work with publishers. 
Anyway,  I will change the code.

test/spec/modules/adtelligentBidAdapter_spec.js
```diff
@@ -109,17 +124,39 @@ const displayEqResponse = [{
...
+    it('should be returned if pixel enabled', function () {
+      const syncs = spec.getUserSyncs({pixelEnabled: true}, [{body: SERVER_DISPLAY_RESPONSE_WITH_MIXED_SYNCS}]);
+
+      expect(syncs.map(s => s.url)).to.deep.equal([SERVER_DISPLAY_RESPONSE_WITH_MIXED_SYNCS.cookieURLs[0]]);
+      expect(syncs.map(s => s.type)).to.deep.equal(['image']);
+    })
+  })
+
+  describe('user syncs as iframe', function () {
+    it('should be returned if pixel enabled', function () {
```
bretg: I think you mean 'should be returned if iframe enabled'

test/spec/modules/adtelligentBidAdapter_spec.js
```diff
@@ -109,17 +124,39 @@ const displayEqResponse = [{
...
+  describe('user syncs with both types', function () {
     it('should be returned if pixel enabled', function () {
-      const syncs = spec.getUserSyncs({pixelEnabled: true}, [{body: SERVER_DISPLAY_RESPONSE}]);
+      const syncs = spec.getUserSyncs({
+        iframeEnabled: true,
+        pixelEnabled: true
+      }, [{body: SERVER_DISPLAY_RESPONSE_WITH_MIXED_SYNCS}]);
 
-      expect(syncs.map(s => s.url)).to.deep.equal(SERVER_DISPLAY_RESPONSE.cookieURLs);
+      expect(syncs.map(s => s.url)).to.deep.equal(SERVER_DISPLAY_RESPONSE_WITH_MIXED_SYNCS.cookieURLs);
```
bretg: So again - the original idea behind usersync was that iframes would be used if they were allowed by the pub and that image syncs were a fallback. We wouldn't expect _both_ types of syncs if both are allowed. Just iframe. Trying to minimize the number of syncs here.

GeneGenie: Sorry. Did not get your point.

GeneGenie: if two syncs are required, and everything is allowed, then 2 syncs are called as expected. Our lineItems may require triggering multiple syncs for single bid, and their types do not matter.

GeneGenie:
Yieldbot test failing CI here.

bretg:
Note that this is only in the master (2.x) branch right now. If you need it in the 1.40.x-legacy branch, will need to be copied there.

GeneGenie:
@bretg yeaaaah, I also need this change in 1.40

PR #3452, opened 2019-01-15
Tests: ensure 12:00:00 hour part on fake timestamp

lovell:
## Type of change

- [x] Bugfix
- [ ] Feature
- [ ] New bidder adapter   
- [ ] Build related changes
- [ ] CI related changes
- [ ] Other

## Description of change

Hello, the currency unit tests are currently setting a fake timestamp to `1047010195974`, equivalent to `2003-03-07T04:09:55Z`. Tests that depend on this value fail on any machine with a TZ of UTC+12 to UTC-4, including the common timezone of UTC itself.

> AssertionError: expected 'https://cdn.jsdelivr.net/gh/prebid/currency-file@1/latest.json?date=20030307' to equal 'https://cdn.jsdelivr.net/gh/prebid/currency-file@1/latest.json?date=20030306'

The change in this PR winds the fake timestamp back to the equivalent of `2003-03-06T12:00:00Z` so the tests will pass with any TZ from UTC+12 to UTC-11.

### DIFF
--- a/test/spec/modules/currency_spec.js
+++ b/test/spec/modules/currency_spec.js
@@ -34,7 +34,7 @@ describe('currency', function () {
   describe('setConfig', function () {
     beforeEach(function() {
       sandbox = sinon.sandbox.create();
-      clock = sinon.useFakeTimers(1047010195974);
+      clock = sinon.useFakeTimers(1046952000000); // 2003-03-06T12:00:00Z
     });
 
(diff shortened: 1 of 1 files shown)


lovell:
Thanks, this was a quick fix to deal with the effect rather than the cause. The underlying code should probably be modified to use `getUTCDate` instead of `getDate` etc.

PR #3456, opened 2019-01-16
remove old dependencies and solve some peer dependencies

snapwich:
## Type of change

## Description of change
Removed unused dependencies, consolidated some dependencies, and solved some peer-dependency warnings (by downgrading `ajv`).


jaiminpanchal27:
LGTM.
Just one question, why do you remove gulp-coveralls. It is getting used here https://github.com/prebid/Prebid.js/blob/master/gulpfile.js#L270

snapwich:
@jaiminpanchal27 that's coveralls, not gulp-coveralls. it doesn't look like we use the later anywhere.

PR #3458, opened 2019-01-17
Added support for Publisher Common ID Module

bjorn-lw:
## Type of change
- [ ] Feature

## Description of change
Added support for Publisher Common ID Module
```
## Other information


bjorn-lw:
Hi dear reviewers!

We need this change for a specific customer who wants to start running with us. It would be great if someone could take a look at this.

Thanks!

Issue #3221, opened 2018-10-24
Feature request - extracting canonical URL


omerko:
## Type of issue
Feature request

## Description
I need to retrieve the canonical url from the <link rel="canonical".. > element <link (https://www.w3.org/TR/capability-urls/#canonical-urls)

Up till now I used the utils.getWindowTop() function to get the document and run a query on the top level DOM.
But since this function is about to get deprecated, there needs to be some alternative to get the canonical URL.
I suppose it's a common thing for adapters to do, having this as a utility function can be useful.

jsnellbaker:
This issue was fixed and included in the 1.31.0 release (that just came out yesterday).  If you have any additional questions/concerns on this topic, please let us know and we'll be happy to help.

PR #3464, opened 2019-01-18
add description to RD module

mkendall07:
## Type of change

- [x] Other

## Description of change
Just a comment explaining what the module does.


jaiminpanchal27:
@bretg I pushed some more comments to address your feedback. 

One of the contributor requested to add canonical url in this module. https://github.com/prebid/Prebid.js/issues/3221
Here is a good read i found if you are interested https://www.shopify.com/partners/blog/canonical-urls

PR #3470, opened 2019-01-22
calling for currency file even when default rates are specified

bretg:
## Type of change

- [X] Bugfix

## Description of change
We introduced a bug with the last release of currency module -- when the `defaultRates` parameter is specified, no call is made to obtain the latest rates.

With this config:
```
      pbjs.setConfig({
        "currency": {
                "adServerCurrency": "UNK",
                "defaultRates": { "USD": { "UNK": 50 }}
            }
      });
```
Expected results are that a call should go out to jsdelivr to pull the currency file.

Actual results: it doesn't currently with 1.37+.

### DIFF
--- a/modules/currency.js
+++ b/modules/currency.js
@@ -57,7 +57,7 @@ export function setConfig(config) {
   if (typeof config.rates === 'object') {
     currencyRates.conversions = config.rates;
     currencyRatesLoaded = true;
-    needToCallForCurrencyFile=false; // don't call if rates are already specified
+    needToCallForCurrencyFile = false; // don't call if rates are already specified
   }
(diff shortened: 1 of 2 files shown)


modules/currency.js
```diff
@@ -11,6 +11,7 @@ const CURRENCY_RATE_PRECISION = 4;
 var bidResponseQueue = [];
 var conversionCache = {};
 var currencyRatesLoaded = false;
+var needToCallForCurrencyFile = true;
```
mkendall07: global state is a bad practice and should be refactored but I won't block the PR as it's already all over this module...

PR #3472, opened 2019-01-22
New Prebid1.0 adapter for SpotX

strider820:
## Type of change

- [ ] Bugfix
- [ ] Feature
- [x] New bidder adapter   
- [ ] Build related changes
- [ ] CI related changes
- [ ] Other

## Description of change
Creating a new SpotX Bidder Adapter

```
{
  bidder: 'spotx',
  params: {
    channel_id: '85394',
    ad_unit: 'outstream',
    outstream_options: {
      slot: 'div-gpt-ad-123445678-1',
      content_width:  '300',
      content_height: '250',
    }
  }
}
```

- jhernandez@spotx.tv

- https://github.com/prebid/prebid.github.io/pull/1110

### DIFF
--- a/modules/spotxBidAdapter.js
+++ b/modules/spotxBidAdapter.js
@@ -242,9 +242,9 @@ export const spec = {
           }
 
           /**
-          * Make sure currency and price are the right ones
-          * TODO: what about the pre_market_bid partners sizes?
-          */
+           * Make sure currency and price are the right ones
+           * TODO: what about the pre_market_bid partners sizes?
+           */
           utils._each(currentBidRequest.params.pre_market_bids, function(pmb) {
             if (pmb.deal_id == spotxBid.id) {
               spotxBid.price = pmb.price;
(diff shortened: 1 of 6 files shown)


modules/spotxBidAdapter.js
```diff
@@ -0,0 +1,417 @@
+import * as utils from 'src/utils';
+import { Renderer } from 'src/Renderer';
+import { registerBidder } from 'src/adapters/bidderFactory';
+import { VIDEO } from 'src/mediaTypes';
```
robertrmartinez: Imports like these need to be relative after `1.38.0`

See: 

so add `../` before each `src`

modules/spotxBidAdapter.js
```diff
@@ -0,0 +1,417 @@
...
+
+      if (utils.getBidIdParameter('number_of_ads', bid.params)) {
+        requestPayload['ext']['number_of_ads'] = utils.getBidIdParameter('number_of_ads', bid.params);
+      }
+
+      const userExt = {};
+
+      // Add GDPR flag and consent string
+      if (bidderRequest && bidderRequest.gdprConsent) {
+        userExt.consent = bidderRequest.gdprConsent.consentString;
```
robertrmartinez: If a pub enables the consent modul, but for some reason the consent information cannot be found,
then prebid passes an object like so to the bidders:

```javascript
{
   gdprConsent: {
      consentString: undefined
      gdprApplies: undefined
      vendorData: undefined
  }
}
```

So technically your `userExt` would be:

```javascript
{
   consent: undefined
}
```

at this point, but it gets pruned before the request eventually gets sent by prebid so no big deal, just though I would make aware that Prebid passes those undefined values in some instances!

test/spec/modules/spotxBidAdapter_spec.js
```diff
@@ -0,0 +1,379 @@
+import {expect} from 'chai';
+import adapterManager from 'src/adapterManager';
+import {spec, outstreamRender} from 'modules/spotxBidAdapter';
+import {parse as parseQuery} from 'querystring';
+import {newBidder} from 'src/adapters/bidderFactory';
+import {userSync} from 'src/userSync';
+import {config} from 'src/config';
+import * as utils from 'src/utils';
+import find from 'core-js/library/fn/array/find';
+
```
robertrmartinez: Almost all of these imports are never used besides `expect` and `spec`

Can you remove the unused ones?


Both `CONSTANTS` as well as `INTEGRATION` below here are not used either.

Let me know if I missed something and there is a reason for keeping them!

modules/spotxBidAdapter.js
strider820: Our ad server should handle undefined in those requests anyways, but that's good to know!

test/spec/modules/spotxBidAdapter_spec.js
strider820: Updated. I went ahead and wrote a few more tests while I was in there to bump that coverage up a bit.

robertrmartinez: Awesome!! Thanks for that.

PR #3474, opened 2019-01-23
Rad 2626 Add resetPresetTargetingAST

deekshithraop:
## Type of change
- [x] Feature

## Description of change
resets targeting using `resetPresetTargetingAST` that already being set in `setTargetingForAst`


mkendall07:
LGTM. Thanks!

PR #3478, opened 2019-01-24
Enforce video branch

sa1omon:
## Type of change

- [X] Feature

## Description of change

- Added additional unit tests

- Added an option to get bids of a specific media type (video, banner) when added to a multi-format bids

 Example :
```
{
  bidder: 'gamoshi',
  params: {
    // ...
    favoredMediaType:'video'
  }
}
```

- Please refer also to: [Related PR](https://github.com/prebid/prebid.github.io/pull/1112) in prebid.github.io


modules/gamoshiBidAdapter.js
```diff
@@ -54,25 +58,25 @@ export const spec = {
...
         },
         'device': {
           'ua': navigator.userAgent
         },
         'imp': [],
         'ext': {}
       };
-      if (bidderRequest && bidderRequest.gdprConsent) {
+
+      if (bidderRequest.gdprConsent) {
```
robertrmartinez: If a pub enables the consent module, but for some reason the consent information cannot be found,
then prebid passes an object like so to the bidders:

```javascript
{
   gdprConsent: {
      consentString: undefined
      gdprApplies: undefined
      vendorData: undefined
  }
}
```

which in this scenario would end up sending an empty object to your exchange like so

```javascript
{
   ... 
   ext: {
      gdpConsent: {}
   }
}
```

Looks like your ad server handles it just fine but wanted to make it known!

sa1omon: Thanks for the insight, we will update the code in the next PR.

PR #3485, opened 2019-01-28
YieldNexus: Add multiple media types branch

sa1omon:
## Type of change

- [X] Refactor

## Description of change

Modify the way the adaptor handles multi-format ad units. 
Modify tests accordingly.


sa1omon:
> LGTM

Hi @mike-chowla, I have tried to download the yieldnexus bidder adapter in the prebid.github.io download page, but i'm getting an error. Do you know why it is happening? Didn't happen before... I'm suspecting that this issue is a result of a mismatch between the bidder code name in prebid.github.io (dev-docs/bidders folder) repository and the name of the adapter in the Prebid.js (modules folder) repository. But since that I didn't changed this file in the last PR, I'm afraid that maybe the reason is something else. I would be happy if you have any insight on this. Thanks!

PR #3489, opened 2019-01-29
Adding freewheel module

jaiminpanchal27:
## Type of change

- [ ] Bugfix
- [x] Feature
- [ ] New bidder adapter   
- [ ] Build related changes
- [ ] CI related changes
- [ ] Other

## Description of change
This PR adds freewheel module to prebid. It has one public api `getTargeting` which returns targeting kev value pairs for adpod mediaType adUnits.

Usage
```
pbjs.adServers.freewheel.getTargeting();
pbjs.adServers.freewheel.getTargeting({codes:['adUnitCode-1']});

Sample return
{
  'adUnitCode-1: [
    {
      'hb_pb_cat_dur': '10.00_403_15s'
    },
    {
      'hb_pb_cat_dur': '15.00_394_30s'
    },
    {
      'hb_uuid': '123'
    }
  ]
}

```


modules/freeWheelAdserverVideo.js
```diff
@@ -0,0 +1,108 @@
...
+ */
+
+import { registerVideoSupport } from '../src/adServerManager';
+import { auctionManager } from '../src/auctionManager';
+import { groupBy, deepAccess } from '../src/utils';
+import { config } from '../src/config';
+// import { ADPOD, initAdpodHooks } from './adpod';
+
+const ADPOD = 'adpod'; // remove later when adpod module is merged; use above commented import instead
+const adPodTargetingKey = 'hb_price_industry_duration';
```
jsnellbaker: Need to update the key based on recent conversation; should be `hb_pb_cat_dur`.

Should we setup a shared constant somewhere to ensure the key is read consistently across different files?  For example, I use this key in the `adpod` file.

modules/freeWheelAdserverVideo.js
```diff
@@ -0,0 +1,108 @@
...
+ * This function returns targeting keyvalue pairs for freewheel adserver module
+ * @param {Object} options
+ * @param {Array[string]} codes
+ * @returns targeting kvs for adUnitCodes
+ */
+export default function getTargeting({codes} = {}) {
+  codes = codes || [];
+  const adPodAdUnits = getAdPodAdUnits(codes);
+  const bidsReceived = auctionManager.getBidsReceived();
+  const competiveExclusionEnabled = config.getConfig('adpod.brandCategoryExclusion');
```
jsnellbaker: Based on recent conversations - is this still needed?

modules/freeWheelAdserverVideo.js
```diff
@@ -0,0 +1,108 @@
...
+
+  let targeting = {};
+  adPodAdUnits.forEach((adUnit) => {
+    let adPodTargeting = [];
+    let adPodDurationSeconds = deepAccess(adUnit, 'mediaTypes.video.adPodDurationSec');
+
+    bids
+      .filter((bid) => bid.adUnitCode === adUnit.code)
+      .forEach((bid, index, arr) => {
+        if (bid.video.durationSeconds <= adPodDurationSeconds) {
```
jsnellbaker: Should update `bid.video.durationSeconds` to `bid.video.durationBucket`

modules/freeWheelAdserverVideo.js
```diff
@@ -0,0 +1,108 @@
...
+  return targeting;
+}
+
+/**
+ * This function returns the adunit of mediaType adpod
+ * @param {Array} codes adUnitCodes
+ * @returns {Array[Object]} adunits of mediaType adpod
+ */
+function getAdPodAdUnits(codes) {
+  return $$PREBID_GLOBAL$$.adUnits
```
jsnellbaker: What if someone uses the `adUnits` field when executing the `$$PREBID_GLOBAL$$.makeBidRequests()` - could that cause issues here (like if they were refreshing a particular adunit instead of everything that may have originally been set)?

modules/freeWheelAdserverVideo.js
```diff
@@ -0,0 +1,108 @@
...
+/**
+ * This function returns bids for adpod adunits
+ * @param {Array[Object]} bidsReceived
+ * @param {Array[Object]} adPodAdUnits
+ * @returns {Array[Object]} bids of mediaType adpod
+ */
+function getBidsForAdpod(bidsReceived, adPodAdUnits) {
+  let adUnitCodes = adPodAdUnits.map((adUnit) => adUnit.code);
+  return bidsReceived
+    .filter((bid) => adUnitCodes.indexOf(bid.adUnitCode) != -1)
```
jsnellbaker: if we do wind up allowing multi-format adpod setups - we'd need to tweak the logic here (I think) to only pull adpod typed bids (since winning banner bids would still have the same adUnitCode for an 'adpod' adUnit and may be pulled in here when we don't want them).

modules/freeWheelAdserverVideo.js
```diff
@@ -0,0 +1,108 @@
...
+}
+
+/**
+ * This function removes bids of same freewheel category. It will be used when competitive exclusion is enabled.
+ * @param {Array[Object]} bidsReceived
+ * @returns {Array[Object]} unique freewheel category bids
+ */
+function getExclusiveBids(bidsReceived) {
+  let bids = bidsReceived
+    .map((bid) => Object.assign({}, bid, {freewheelPrimaryCatId: bid.meta.freewheelPrimaryCatId}));
```
jsnellbaker: need to update `freewheelPrimaryCatId` to newer property name.

modules/freeWheelAdserverVideo.js
jsnellbaker: From further internal conversation - we are not support multi-format with adpod and will reject the adUnits before the auction starts.  This area shouldn't need an update based on the above, but still can tweak if you want to be extra sure.

modules/freeWheelAdserverVideo.js
jsnellbaker: I updated the code in the `adpod` module; I'm exporting this var now: `TARGETING_KEY_PB_CAT_DUR`

modules/freeWheelAdserverVideo.js
jaiminpanchal27: Added check for context

modules/freeWheelAdserverVideo.js
jaiminpanchal27: I was under the impression that we modify adUnits inside requestBids. Hence used the global. Now modified to use auctionManager.getAdUnits

jaiminpanchal27:
@jsnellbaker Updated the code

modules/freeWheelAdserverVideo.js
```diff
@@ -0,0 +1,122 @@
...
+ * @param {Array} codes adUnitCodes
+ * @returns {Array[Object]} adunits of mediaType adpod
+ */
+function getAdPodAdUnits(codes) {
+  return auctionManager.getAdUnits()
+    .filter((adUnit) => deepAccess(adUnit, 'mediaTypes.video.context') === ADPOD)
+    .filter((adUnit) => (codes.length > 0) ? codes.indexOf(adUnit.code) != -1 : true);
+}
+
+function compareOn(property) {
```
mkendall07: we may this functionality already in utils...

modules/freeWheelAdserverVideo.js
```diff
@@ -0,0 +1,122 @@
...
+ * @param {Array[Object]} bidsReceived
+ * @returns {Array[Object]} unique freewheel category bids
+ */
+function getExclusiveBids(bidsReceived) {
+  let bids = bidsReceived
+    .map((bid) => Object.assign({}, bid, {[TARGETING_KEY_PB_CAT_DUR]: bid.adserverTargeting[TARGETING_KEY_PB_CAT_DUR]}));
+  bids = groupBy(bids, TARGETING_KEY_PB_CAT_DUR);
+  let filteredBids = [];
+  Object.keys(bids).forEach((targetingKey) => {
+    bids[targetingKey].sort(compareOn('responseTimestamp'));
```
mkendall07: 👍

PR #3492, opened 2019-01-29
Get bid responses

deekshithraop:
## Type of change

- [x ] Bugfix
- [x ] Feature

## Description of change
`$$PREBID_GLOBAL$$.getBidResponses` will return the responses of the of the last auction if there are successful bids(i.e last auctionId is determined based on the last response returned), But last auction may not have any responses. This pull req will fix that if there are no responses in the last auction   `getBidResponses` will return empty object


src/auctionManager.js
```diff
@@ -100,6 +100,10 @@ export function newAuctionManager() {
     }
   }
 
+  auctionManager.getLastAuctionId = function() {
+    return _auctions && _auctions.length && _auctions[_auctions.length - 1].getAuctionId()
```
snapwich: `_auctions` should always be an Array (it's initialized as an array (with `let`, might want to switch to `const` to make sure that's always the case)), so I think all these checks are unnecessary.

deekshithraop: made the change thanks

Issue #3477, opened 2019-01-24
${AUCTION_PRICE} not being replaced in adm from s2s bid


JonGoSonobi:
## Type of issue
Bug

## Description
Sonobi has been developing a s2s adapter and has a PR open for it. https://github.com/prebid/prebid-server/pull/796. While testing I noticed that the ${AUCTION_PRICE} macro is not being replaced when the ad renders. 

## Steps to reproduce
1. Go to http://mtrx.go.sonobi.com/sonobi_pbjs_server_test.html?pbjs_debug=true.
2. Assuming you get a bid and you hit the right line items, you should see an orange test sonobi ad.
3. Inspect the network requests find the request that matches something like http://*.go.sonobi.com/sbid.js and notice that ${AUCTION_PRICE} is not replaced in the url from the bid.adm from prebid-server's /auction response.

## Test page
http://mtrx.go.sonobi.com/sonobi_pbjs_server_test.html?pbjs_debug=true - friendly frames, macro replacement works
http://mtrx.go.sonobi.com/sonobi_pbjs_server_safeframe_test.html?pbjs_debug=true - safe frames, macro replacement not applied

### Expected results
http://*.go.sonobi.com/sbid.js's ${AUCTION_PRICE} macro to contain the final winning bid cpm.

### Actual results
${AUCTION_PRICE} is still present in the request url to http://*.go.sonobi.com/sbid.js

## Platform details

Prebid.js v1.38.0
Browser: Chrome 71.0.3578.98
OS: macOS 10.14
node: v8.9.0
npm: v5.5.1

## Other information
Related to Prebid server sonobi adapter PR https://github.com/prebid/prebid-server/pull/796.

JonGoSonobi:
Dug further into this and learned the macro replacements work properly with friendly frames.
I've updated the creatives in dfp to be friendly frames, but it still seems to be an issue with the universal creative when loaded into a safe frame.

PR #3493, opened 2019-01-29
Fix issue where the ${AUCTION_PRICE} macro was not being replaced in secure creatives

JonGoSonobi:
## Type of change
- [x] Bugfix

## Description of change
Macro replace ${AUCTION_PRICE} when postMessaging the ad info back to the secure creative.

- prebid.apex@sonobi.com

## Other information
Fixes https://github.com/prebid/Prebid.js/issues/3477


bretg:
Probably ought to have a unit test here.

mkendall07:
thanks for the PR! +1 to the unit test.

PR #3501, opened 2019-01-31
Introduced AJAX ontimeout event callback

marakoss:
## Type of change
- [x] Feature

## Description of change
This change allows for handling ontimeout event in AJAX library using the already in place callbacks, whereas this previously wasn't possible.

### DIFF
--- a/src/ajax.js
+++ b/src/ajax.js
@@ -33,6 +33,12 @@ export function ajaxBuilder(timeout = 3000, {request, done} = {}) {
         }
       };
 
+      if (typeof callbacks.timeout !== 'function') {
+        callbacks.timeout = function (e) {
+          utils.logError('xhr timeout after ', x.timeout, 'ms');
+        }
+      }
+
       if (typeof callback === 'function') {
         callbacks.success = callback;
       }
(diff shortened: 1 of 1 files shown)


src/ajax.js
```diff
@@ -33,6 +33,12 @@ export function ajaxBuilder(timeout = 3000, {request, done} = {}) {
         }
       };
 
+      if (typeof callbacks.timeout !== 'function') {
```
mkendall07: can you gave a usage example and update the jsDoc for this?

src/ajax.js
```diff
@@ -33,6 +33,12 @@ export function ajaxBuilder(timeout = 3000, {request, done} = {}) {
...
+        callbacks.timeout = function (e) {
+          utils.logError('xhr timeout after ', x.timeout, 'ms');
```
mkendall07: side effect here is the logError isn't logged if you define a custom timeout function.

marakoss: @mkendall07 That is true. What are you suggesting? Do some sort of extension of the previous function ? like

```x.timeout = function (e) {
    callbacks.timeout(e);     
    utils.logError('xhr timeout after ', x.timeout, 'ms');
}```

What if the dev would want to "just" extend the logError to display real timeout time (not the x.timeout settings) ? Than he would always have two messages in console.

marakoss: @mkendall07 Hello, can I please get any kind of instruction? :) Thank you

mkendall07: this was just an observation, not sure you need to do anything about it. Can you please update the usage with a JSdoc notation?

mkendall07:
@marakoss any updae?

PR #3508, opened 2019-02-04
YieldNexus: Rename adapter's name

sa1omon:
## Type of change

## Description of change

Renamed ```yieldNexusBidAdapter.js```  to  ```yieldnexusBidAdapter.js``` in order to fix download issue.


bretg:
@sa1omon - you'll need to rename the dev-docs/bidders/yieldNexus.md file over in https://github.com/prebid/prebid.github.io as well

Note that renaming like this is likely to cause issues with the Download page -- I believe users will get errors when trying to retrieve yieldnexus for previous releases.

Assuming you're ok with this, I'll merge after the docs PR is submitted.

sa1omon:
> @sa1omon - you'll need to rename the dev-docs/bidders/yieldNexus.md file over in https://github.com/prebid/prebid.github.io as well
> 
> Note that renaming like this is likely to cause issues with the Download page -- I believe users will get errors when trying to retrieve yieldnexus for previous releases.
> 
> Assuming you're ok with this, I'll merge after the docs PR is submitted.

Hi @bretg, 
The reason for the renaming is because a few days a go I have tried to download the yieldnexus bidder adapter in the prebid.github.io download page, but I got an error.
Do you know why it is happening? Didn't happen before...
I'm assumed that this issue is a result of a mismatch between the bidder code name in prebid.github.io (**dev-docs/bidders folder**) repository and the name of the adapter in the Prebid.js (**modules** folder) repository.
But since I didn't change this file in the last PR, I'm assumed that maybe the reason is something else.
I would be happy if you have any insight on this. Thanks!

bretg:
@sa1omon - I changed the biddercode in the docs .md file and yieldNexus now downloads.

Can we close this PR or would you prefer to make the name change anyhow?

sa1omon:
> @sa1omon - I changed the biddercode in the docs .md file and yieldNexus now downloads.
> 
> Can we close this PR or would you prefer to make the name change anyhow?

Hi @bretg,
I like that the download is working, but after some checks against our code i noticed that prebid.js doesn’t find the bidder with the camel case version (yieldNexus). 
This means that the current documentation is misleading and will not work. 
How can we keep the lower case bidder code name (in the doc and the code ) and fix the download page?

Thanks

bretg:
Sorry - what's the problem? yieldNexus is camel case almost everywhere right now and downloads.

The only place it's not camel case is in https://github.com/prebid/Prebid.js/blob/master/modules/yieldNexusBidAdapter.js

```
code: 'yieldnexus',
```

If what you're saying is that bids don't actually work, that's likely the problem. So again - if you want to rename everything to lower case, that's fine. Or you can just fix yieldNexusBidAdapter.js to be camel case and all should be good.

sa1omon:
> Sorry - what's the problem? yieldNexus is camel case almost everywhere right now and downloads.
> 
> The only place it's not camel case is in https://github.com/prebid/Prebid.js/blob/master/modules/yieldNexusBidAdapter.js
> 
> ```
> code: 'yieldnexus',
> ```
> If what you're saying is that bids don't actually work, that's likely the problem. So again - if you want to rename everything to lower case, that's fine. Or you can just fix yieldNexusBidAdapter.js to be camel case and all should be good.

Hi @bretg,
Yes, the problem is that the bids don't actually work.
From that reason I prefer to rename everything to lower case.
Thanks.

bretg:
Ok.

PR #3509, opened 2019-02-04
Update openxoutstream adapter

HolzAndrew:
## Type of change
- [x] Other

## Description of change
-updates some values of the query params
-removes hardcoded availability_zone in template response
-adds loader code to template response

@harquail @jasminenoack @MelodyLi2017


test/spec/modules/openxoutstreamBidAdapter_spec.js
```diff
@@ -124,8 +124,10 @@ describe('OpenXOutstreamAdapter', function () {
       }];
       const request = spec.buildRequests(bidRequestsWithUnitIds);
       expect(request[0].data.auid).to.equal(`${bidRequestsWithUnitIds[0].params.unit}`);
-      expect(request[0].data.vht).to.equal(`${bidRequestsWithUnitIds[0].params.height}`);
-      expect(request[0].data.vwd).to.equal(`${bidRequestsWithUnitIds[0].params.width}`);
+      expect(request[0].data.vht).to.not.equal(`${bidRequestsWithUnitIds[0].params.height}`);
+      expect(request[0].data.vwd).to.not.equal(`${bidRequestsWithUnitIds[0].params.width}`);
+      expect(request[0].data.vht).to.equal('300');
+      expect(request[0].data.vwd).to.equal('400');
```
jasminenoack: aus too.

HolzAndrew: good point. added it.

modules/openxoutstreamBidAdapter.md
```diff
@@ -18,16 +18,14 @@ This bid adapter supports Banner.
 var adUnits = [
   {
     code: 'test-div',
-    sizes: [[300, 250]],  // a display size
+    sizes: [[300, 250],[640,480]],  // a display size
```
bretg: First, please update this documentation to use the `mediaTypes` syntax -- `sizes` here is old. (Yes, there are many adapters that haven't updated this yet, but we have to start somewhere :-)

But more importantly, this specific change isn't making sense to me: why are you passing in a common display size and a video player size at the same time?

HolzAndrew: no problem, sorry I missed that the syntax had updated!

The 2nd size was in there by mistake. I was testing something and forgot to delete it.

Thanks!

modules/openxoutstreamBidAdapter.js
```diff
@@ -80,8 +83,9 @@ function buildCommonQueryParamsFromBids(bid, bidderRequest) {
     dddid: bid.transactionId,
     openrtb: '%7B%22mimes%22%3A%5B%22video%2Fmp4%22%5D%7D',
     nocache: new Date().getTime(),
-    vht: bid.params.height || bid.sizes[0][1],
-    vwd: bid.params.width || bid.sizes[0][0]
+    vht: height,
```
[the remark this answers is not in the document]
HolzAndrew: ok, added a note to the MD file!

bretg:
@HolzAndrew - so let me see if I understand -- if a user specifies an AdUnit of size 300x250 and includes openxoutstream as an adapter, it will under the covers change the size to 400x300 and make the auction request to your backend. 

The net result being that a request for 300x250 could end up returning a creative of a totally different size? This bid will get thrown out due to size mismatch.

If I'm understanding this correctly, you need to fix this behavior.

If 400x300 is the only size you support, that's fine, but you should avoid making auctions unless the AdUnit specifies that size. Should not be weighing down the browser with requests/responses that are going to get thrown out due to a size mismatch.

harquail:
> @HolzAndrew - so let me see if I understand -- if a user specifies an AdUnit of size 300x250 and includes openxoutstream as an adapter, it will under the covers change the size to 400x300 and make the auction request to your backend.
> 
> The net result being that a request for 300x250 could end up returning a creative of a totally different size? This bid will get thrown out due to size mismatch.
> 
> If I'm understanding this correctly, you need to fix this behavior.
> 
> If 400x300 is the only size you support, that's fine, but you should avoid making auctions unless the AdUnit specifies that size. Should not be weighing down the browser with requests/responses that are going to get thrown out due to a size mismatch.

@HolzAndrew @bretg 

My understanding is that the sizes are a reflection of OpenX’s demand, and will display within the size of the placement created by the publisher.

So in the situation you described — the openx-outstream adapter would make a request to OpenX for video demand up to that size (400x300). The Yieldmo format then dynamically resizes the video within the creative to fit within the placement size set up by the publisher (300x250 in this case).

modules/openxoutstreamBidAdapter.js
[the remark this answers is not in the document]
HolzAndrew: > > @HolzAndrew - so let me see if I understand -- if a user specifies an AdUnit of size 300x250 and includes openxoutstream as an adapter, it will under the covers change the size to 400x300 and make the auction request to your backend.
> > The net result being that a request for 300x250 could end up returning a creative of a totally different size? This bid will get thrown out due to size mismatch.
> > If I'm understanding this correctly, you need to fix this behavior.
> > If 400x300 is the only size you support, that's fine, but you should avoid making auctions unless the AdUnit specifies that size. Should not be weighing down the browser with requests/responses that are going to get thrown out due to a size mismatch.
> 
> @HolzAndrew @bretg
> 
> My understanding is that the sizes are a reflection of OpenX’s demand, and will display within the size of the placement created by the publisher.
> 
> So in the situation you described — the openx-outstream adapter would make a request to OpenX for video demand up to that size (400x300). The Yieldmo format then dynamically resizes the video within the creative to fit within the ad unit size set up by the publisher (300x250 in this case).

@harquail Exactly, thats the gist. (@bretg )

HolzAndrew:
> > @HolzAndrew - so let me see if I understand -- if a user specifies an AdUnit of size 300x250 and includes openxoutstream as an adapter, it will under the covers change the size to 400x300 and make the auction request to your backend.
> > The net result being that a request for 300x250 could end up returning a creative of a totally different size? This bid will get thrown out due to size mismatch.
> > If I'm understanding this correctly, you need to fix this behavior.
> > If 400x300 is the only size you support, that's fine, but you should avoid making auctions unless the AdUnit specifies that size. Should not be weighing down the browser with requests/responses that are going to get thrown out due to a size mismatch.
> 
> @HolzAndrew @bretg
> 
> My understanding is that the sizes are a reflection of OpenX’s demand, and will display within the size of the placement created by the publisher.
> 
> So in the situation you described — the openx-outstream adapter would make a request to OpenX for video demand up to that size (400x300). The Yieldmo format then dynamically resizes the video within the creative to fit within the ad unit size set up by the publisher (300x250 in this case).

@harquail exactly, thats the gist (@bretg )

bretg:
Will merge when @jasminenoack

HolzAndrew:
> Will merge when @jasminenoack

@bretg thanks! she just approved

PR #3517, opened 2019-02-05
fixed bug with Firefox using the iframe's history instead of the top …

jarnix:
## Type of change

- [x ] Bugfix

## Description of change
The fix is for Firefox that does not handle properly the back button when using document.write, I added a document.open('text/html', document.write...) and now the back button and the history is not polluted by the iframe's document.write.

We had this issue on our sites with Firefox, version 60 to 65 (and maybe the next ones).

Source of the workaround : https://stackoverflow.com/questions/19624452/workaround-to-firefox-creating-new-history-after-each-document-write


jarnix:
forgot something

PR #3523, opened 2019-02-06
add adpod to mediaTypes

jsnellbaker:
## Type of change
- [x] Feature

## Description of change
This change just adds `adpod` as an accepted mediaType.  This part of the changes for https://github.com/prebid/Prebid.js/issues/3379

### DIFF
--- a/src/mediaTypes.js
+++ b/src/mediaTypes.js
@@ -15,3 +15,5 @@ export const NATIVE = 'native';
 export const VIDEO = 'video';
 /** @type MediaType */
 export const BANNER = 'banner';
+/** @type MediaType */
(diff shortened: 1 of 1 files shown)


src/mediaTypes.js
```diff
@@ -15,3 +15,5 @@ export const NATIVE = 'native';
 export const VIDEO = 'video';
 /** @type MediaType */
 export const BANNER = 'banner';
+/** @type MediaType */
```
jaiminpanchal27: This can be `VideoContext`.  We can add new typedef
```
/**
 * @typedef {('adpod')} VideoContext
 */
```

PR #3524, opened 2019-02-06
bugfix for Firefox for some ads that use document.write

jarnix:
## Type of change

- [x ] Bugfix

## Description of change
This is a fix targeted for Firefox that creates an entry in history after a document.write. It means that when an ad uses document.write (as we see in a lot of ads on our 30 M users network), the user has to click the "back" button twice to go back. The bug is 17 years old (https://bugzilla.mozilla.org/show_bug.cgi?id=148794) and the workaround is described here for example : https://stackoverflow.com/questions/19624452/workaround-to-firefox-creating-new-history-after-each-document-write/32677679#32677679 .

### DIFF
--- a/src/prebid.js
+++ b/src/prebid.js
@@ -270,6 +270,7 @@ $$PREBID_GLOBAL$$.renderAd = function (doc, id) {
           const message = `Error trying to write ad. Ad render call ad id ${id} was prevented from writing to the main document.`;
           emitAdRenderFail(PREVENT_WRITING_ON_MAIN_DOCUMENT, message, bid);
         } else if (ad) {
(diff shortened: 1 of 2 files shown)


jarnix:
I don't understand why the test in Circle CI would be failing and I surely don't want to modify the test. Please, can someone more experienced with the tests of this project can check the code ?

jarnix:
@bretg I agree with you, it's a "pretty important change", that's why I did not want to modify the tests but I got it why the test was failing. The doc.open method was called but not monitored by sinon.spy() so I added it. So now the tests are okay in the CI, but I really would like a human check though..

PR #3527, opened 2019-02-06
Update removeAdUnit

deekshithraop:
## Type of change

- [x] Feature

### pbjs.removeAdUnit()

updated removeAdUnit so that if it is invoked with no arguments it will remove all add units, and it will also accept an array argument of addUnitCodes and will only remove them.

- link to a PR on the docs change(https://github.com/prebid/prebid.github.io/pull/1138)


bretg:
What's the use case here?

mkendall07:
modification of adUnits in single page applications. Eventually we should remove `pbjs.adUnits` as a globally accessible property,.

src/prebid.js
```diff
@@ -375,6 +375,16 @@ $$PREBID_GLOBAL$$.removeAdUnit = function (adUnitCode) {
   }
 };
 
+/**
+ * Remove all adUnits from the $$PREBID_GLOBAL$$ configuration
+ * @alias module:pbjs.removeAllAdUnits
+ */
+$$PREBID_GLOBAL$$.removeAllAdUnits = function () {
```
mkendall07: hmmm should we make this `removeAdUnits([adUnitCodeArray])` ? seems more flexible.

deekshithraop:
@mkendall07 made changes and updated description

src/prebid.js
```diff
@@ -360,19 +360,33 @@ $$PREBID_GLOBAL$$.renderAd = function (doc, id) {
...
 $$PREBID_GLOBAL$$.removeAdUnit = function (adUnitCode) {
   utils.logInfo('Invoking $$PREBID_GLOBAL$$.removeAdUnit', arguments);
-  if (adUnitCode) {
+
+  if (!adUnitCode) {
+    $$PREBID_GLOBAL$$.adUnits = [];
+    return;
+  }
+
+  var adUnitCodes;
```
mkendall07: prefer let.

mkendall07:
@bretg fyi we decided to not add an extra method here but instead reuse the existing one.

src/prebid.js
deekshithraop: done

Issue #3058, opened 2018-09-06
POC: Implement Promises in Prebid core


jaiminpanchal27:
## Type of issue
Feature request

## Description
Implement Promises in Prebid core

snapwich:
This has been discussed previously and I was able to convince most at the time that it is a poor decision.  I still have the same opinion.

Please refer to [this point in the Promise spec](https://promisesaplus.com/#point-67).  Fulfillment of `then` handlers should happen asynchronously after the event loop turn.  The implication here for Prebid.js is that we would then be deferring our execution every time we used a `.then` handler and slowing the start of the Prebid.js auction, a penalty we do not incur with callbacks as they are executed on the existing call stack. 

 I don't improving the developer experience with the addition of Promises is worth the performance costs.

mkendall07:
thanks @snapwich 
My thought was that we'd only use promises for async operations already (like HTTP requests). Does that change your perspective? Or does that just make things more complicated :)

snapwich:
I don't know if it's worth polyfilling a promise library for the limited use-case of http request callbacks.  Our ajax library works pretty okay as it is.  If we do decide to go towards promises for HTTP later (after the burden of IE11 has been lifted) then I'd suggest just moving towards using the [Fetch API](https://developer.mozilla.org/en-US/docs/Web/API/Fetch_API), which uses Promises internally.

jsfaure:
About promise and the possible risk of slowing the start of Prebid.js auction, I would just moderate the risk as it looks like the event loop will take care of handling the promise callbacks before keeping on with the others IO events. See https://www.zeolearn.com/magazine/javascript-how-is-callback-execution-strategy-for-promises-different-than-dom-events-callback

gglas:
We expect this to be folded into the 6.0 migration to Modern JS.

patmmccann:
Promises have begun to be used in prebid core already; closing this issue in favor of more specific requests

PR #3540, opened 2019-02-08
[RFC] Support for a Custom Bid Adapter

gpolaert:
This proposal offers the ability for a publisher to use Prebid as a global controller both for official bidders and third-parties that are not compatible with prebid.

For instance, a publisher wants to monetize a slot using its Prebid bidders (let's said appnexus, rubicon, improve ...) AND its others bidders (a9, various networks, proprietary ad sever, ...).

In order to achieve that goal, the publishers must recreate a custom logic to handle all requests (prebid + others), to wait for all responses (or timeout) and to decide who is the winner.

Dealing with 2 timeout managers or gather all responses into one array is so painful, I think using Prebid as a global controller is great feature and offers many possibilities.

I propose to add a custom bid adapter.
This adapter will fake the request/response behavior by allowing the publisher to use its custom logic to generate requests and responses. The publisher has to provide 2 handlers in order to use the existing Prebid logic (timeout management, wining decision, no response, Prebid events ...).

The publisher will use the custom bid adapter like bellow. He can add as many custom logics as he wants by using aliases.
```
pbjs.aliasBidder('customBidAdapter', 'my-ad-server');
pbjs.bidderSettings = {
    standard: {},
    'my-ad-server': {
        handlers: {
            buildRequests: myCustomHandlerToBuildRequests,
            interpretResponse: myCustomHandlerToInterpretResponses,
        }
    }
};
```

In conclusion, Publishers that still want to use networks or bidders that not have a Prebid adapter have the ability to add them easily by adding the bidder tag in the page, and updating the Prebid configuration as they already do for a new Prebid Bidder.

I pushed a PR in order to illustrate the changes needed.
If you are okay to add it to Prebid, I will update the PR to be compatible with the code requirements.

### DIFF
--- a/modules/customBidAdapter.js
+++ b/modules/customBidAdapter.js
@@ -1,0 +1,54 @@
+import { registerBidder } from 'src/adapters/bidderFactory';
+import { BANNER, NATIVE, VIDEO } from 'src/mediaTypes';
+
+const BIDDER_CODE = 'customBidAdapter';
+
+export const spec = {
+  code: BIDDER_CODE,
+  supportedMediaTypes: [BANNER, VIDEO, NATIVE],
+
+  /**
+   * Determines whether or not the given bid request is valid.
+   * If there is no 'isBidRequestValid' handler, the bid is considered valid .
+   *
+   * @param {object} bid The bid to validate.
+   * @return boolean True if this is a valid bid, and false otherwise.
+   */
+  isBidRequestValid: function (bid) {
+    const { handlers = {} } = bid;
+    return typeof handlers.isBidRequestValid === 'function' ? handlers.isBidRequestValid(bid) : true;
+  },
+
+  /**
+   * Make a request from the list of BidRequests.
+   *
+   * @param {BidRequest[]} bidRequests A non-empty list of bid requests which should be sent to the Server.
+   * @param {object}  bidderRequest The Bidder Request associated to the requests.
+   * @return [ServerRequest] Info describing the requests to the server.
+   */
+  buildRequests: function (bidRequests, bidderRequest) {
+    const { handlers = {} } = bidderRequest;
+    if (typeof handlers.buildRequests !== 'function') return [];
+    return handlers.buildRequests(bidRequests, bidRequest).filter(isPromise);
+  },
+
+  /**
+   * Unpack the response from the server into a list of bids.
+   *
+   * @param {*} serverResponse A successful response from the server.
+   * @param {ServerRequest} serverRequest The server request
+   * @return {Bid[]} An array of bids which were nested inside the server.
+   */
+  interpretResponse: function (serverResponse, serverRequest) {
+    const { handlers = {} } = bidderRequest;
+    if (typeof handlers.interpretResponse !== 'function') return [];
+    return handlers.interpretResponse(serverResponse, serverRequest);
+  },
+
+};
+
+function isPromise(obj) {
+  return typeof obj === 'object' && obj.then && typeof obj.then === 'function';
+}
+
+registerBidder(spec);
--- a/src/adapters/bidderFactory.js
+++ b/src/adapters/bidderFactory.js
@@ -66,9 +66,10 @@ import { logWarn, logError, parseQueryStringParameters, delayExecution, parseSiz
 /**
  * @typedef {object} ServerRequest
  *
- * @property {('GET'|'POST')} method The type of request which this is.
+ * @property {('GET'|'POST'|'PROMISE')} method The type of request which this is.
  * @property {string} url The endpoint for the request. For example, "//bids.example.com".
  * @property {string|object} data Data to be sent in the request.
+ * @property {Promise} promise The Promise when the method is 'PROMISE'.
  * @property {object} options Content-Type set in the header of the bid request, overrides default 'text/plain'.
  *   If this is a GET request, they'll become query params. If it's a POST request, they'll be added to the body.
  *   Strings will be added as-is. Objects will be unpacked into query params based on key/value mappings, or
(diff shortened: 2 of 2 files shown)


mkendall07:
This pull request **introduces 1 alert** when merging 845a70adf291b5599976b39f017488d2eb4ba64e into c2734a73fc907dc6c97d7694e3740e19b8749d3c - [view on LGTM.com](https://lgtm.com/projects/g/prebid/Prebid.js/rev/pr-997883a8b621dda98cb0cbaff3579ee1cdb0808f)

**new alerts:**

---

gpolaert:
I've updated the PR.
Could you get your feedback about this? @mkendall07 @bretg

PS: I have to fix the Promise in the spec file.

bretg:
Rubicon is good with the general idea of Prebid.js being a wrapper-of-wrappers. Unclear whether Prebid.org is going to be able to formally support any particular customization -- will bring it up at our next meeting.

Will let others decide if this PR is the right approach technically, but thanks for bringing it up @gpolaert

gpolaert:
@bretg, Great news! I really like "wrapper of wrappers". It's definitively the idea.
About the PR, let me know how technically you want to tackle the issue. I can make the adjustments if you want to.

gpolaert:
Any news? Have you discussed about it at the last meeting @bretg?

bretg:
Confirmed that Prebid.org is ok with the general idea of being a wrapper-of-wrappers.

However, that isn't a commentary on the specific technical implementation proposed here. Leaving that to @jaiminpanchal27

gpolaert:
Good news! @jaiminpanchal27 if I can help ... :grin:

jaiminpanchal27:
@gpolaert Overall I like this implementation. I don't see any performance impact or anything breaking in core with this approach. Also without making any major additions to our api we are able to implement this.

Only concern is usage of Promise. Please go through discussion here https://github.com/prebid/Prebid.js/issues/3058.

Few other possible approaches to this new feature can be:
- Instead of creating a `customBidAdpater` as new adapter add this new feature as module `customBidderModule`. This new module will act as a base class for custom bid adapters, similar to what has been done in this PR. Create new api `pbjs.createRunTimeAdapter(spec)` to register your custom specs.

Not going with this approach because creating this feature as module introduces new api. No need to add anything new if feature can be implemented using existing api's.  Also there is no significant advantage with this approach in terms of performance, reducing complexity, size of lib.

- Update bidderFactory module to support this feature. Create new api `pbjs.createRunTimeAdapter(spec)` to register your custom spec. This api will add the new adapter run time. By doing this we add this feature as part of core increasing core size. I am not sure everyone would want this by default.

gpolaert:
Thank for the review @jaiminpanchal27 
The current implementation has no overhead on the core lib. I think we should keep it as a adapter, allowing publishers to add it to their build. 

About the promise, I'm not fully agree with the #3058. The article mentioned by @jsfaure is very interesting. Anyway, the promise goal is to make the bridge between the request and the response.
By definition, this step is already async (xhr calls). So, there is no side effect or delay about the execution of the prebid auctions.

What about the next steps ? :grin:

jaiminpanchal27:
@gpolaert Next step is creating actual PR. We are happy with this approach.

patmmccann:
Is the vision that each publisher would write their own a9 handler or that there will be one checked into the code or the integration examples?

gpolaert:
@jaiminpanchal27 not sure what I have to do (new PR?)

@parisholley this adapter allows publishers to write own logic. Amazon has a very restrictive policy about the visibility of their cpm. So not sure it could be helpful here.

bretg:
@gpolaert - is this custom adapter operational? I don't see code that pulls from bidderSettings. Assumed it was a mostly done proof-of-concept?

From a usage perspective, my only comment is that I'd prefer that custom bidder functions be placed under a `custom` object.

```
pbjs.aliasBidder('customBidAdapter', 'my-ad-server');
pbjs.bidderSettings = {
    standard: {},
    custom: {
        'my-ad-server': {
            handlers: {
                buildRequests: myCustomHandlerToBuildRequests,
                interpretResponse: myCustomHandlerToInterpretResponses,
            }
        }
    }
};
```

zhenyi2697:
Promise is not supported by all browsers, especially for IE: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Promise#Browser_compatibility, shouldn't add a polyfill for promise like this: https://www.npmjs.com/package/promise-polyfill (< 1kb Gzipped)

webdeveloperpr:
I wish this was something that could be added. I am using prebid + amazon + ix video and other stuff which complicates things.

bretg:
@jaiminpanchal27 and @gpolaert - what's the status on this? 

1) Use of Promise
2) Next steps

jaiminpanchal27:
@bretg Prebid Core needs to take call on usage of Promise. Lets discuss in next meeting and update everyone

mkendall07:
I see this has been sitting a long time. @gpolaert  
if you want to include the promise polyfill from the custom adapter that should satisfy our requirements (not increase core build size / allow promises in the handler to support IE 11).

snapwich:
Sorry I'm late, but this functionality already exists.

```
pbjs.registerBidAdapter(function() {
	return {
		buildRequests: myCustomHandlerToBuildRequests,
		interpretResponse: myCustomHandlerToInterpretResponses
	}
}, 'custom')
```

Unless I'm missing something, I can't see what this adds that can't be accomplished with the public `registerBidAdapter` API.

mkendall07:
Good callout Rich! I'm going to close this PR. If anyone disagrees feel free to open a new issue.

PR #3544, opened 2019-02-11
use custom bind where non-standard mootools bind could cause issues

snapwich:
## Type of change
- [x] Bugfix

## Description of change
https://github.com/snapwich/fun-hooks/issues/1 highlighted an issue where MooTools with the compatibility layer being present on the page [can cause non-standard behavior](https://mootools.lighthouseapp.com/projects/2706/tickets/822-google-chrome-3-breaks-functionprototypebind-due-to-built-in-function-of-same-name) with `Function.prototype.bind`

Not really a bug in Prebid.js, but probably worth avoiding if we can (as it could still cause issues with Prebid.js installs).  This pull-request updates incompatible usages of `bind` to use our custom `utils.bind` which will be either `Function.prototype.bind` if the incompatible MooTools `bind` is not detected, or our custom implementation of the `bind` function if it _is_ detected.  Also updates `fun-hooks` dependency to use version with similar fix.

### DIFF
--- a/modules/prebidServerBidAdapter/index.js
+++ b/modules/prebidServerBidAdapter/index.js
@@ -169,7 +169,7 @@ function doAllSyncs(bidders) {
 
   const thisSync = bidders.pop();
   if (thisSync.no_cookie) {
-    doBidderSync(thisSync.usersync.type, thisSync.usersync.url, thisSync.bidder, doAllSyncs.bind(null, bidders));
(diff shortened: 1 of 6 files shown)


idettman:
package-lock needs to be resolved

snapwich:
updated

snapwich:
Since this is similar to #3574 and they both include fixes for `fun-hooks`, I merged this into #3574 and am closing this.

PR #3545, opened 2019-02-12
Add video support in Grid Bid Adapter

TheMediaGrid:
## Type of change

- [x] Feature

## Description of change
Added video support in Grid Bid Adapter

- test parameters for validating video bids
```
{
  bidder: 'grid',
  params: {
    uid: 11
  }
}
```

- grid-tech@themediagrid.com

- A link to a PR on the docs repo https://github.com/prebid/prebid.github.io/pull/1149


modules/gridBidAdapter.js
```diff
@@ -134,9 +140,24 @@ function _addBidResponse(serverBid, bidsMap, bidResponses) {
...
+          if (!bid.renderer && (!bid.mediaTypes || !bid.mediaTypes.video || bid.mediaTypes.video.context === 'outstream')) {
+            bidResponse.renderer = createRenderer(bidResponse, {
+              id: bid.bidId,
+              url: RENDERER_URL
+            });
+          }
+        } else {
+          bidResponse.ad = serverBid.adm;
+          bidResponse.mediaType = BANNER;
+        }
```
robertrmartinez: Hi this all looks great,

I have a question as to why when it is video there is no `bidResponse.width` or `bidResponse.height` added?

I see it is not sent or returned by your adserver.

You can see in the below screenshot that the KVP that gets set for the size turns out to be `undefinedxundefined`

image

I have a feeling this _may_ cause problems for certain publishers ad servers maybe?

Do you think it is a good idea to set the width and height based on the bidRequest object?

modules/gridBidAdapter.md
```diff
@@ -35,6 +36,19 @@ Module that connects to Grid demand source to fetch bids.
...
+               mediaTypes: { video: {} },
+               bids: [
+                   {
+                       bidder: "grid",
+                       params: {
+                           uid: 11
+                       }
+                   }
+               ]
+          }
```
robertrmartinez: Thanks for good test parameters which return valid bid responses! 

This is great.

test/spec/modules/gridBidAdapter_spec.js
```diff
@@ -255,6 +259,51 @@ describe('TheMediaGrid Adapter', function () {
...
+      const response = {'bid': [{'price': 1.15, 'adm': '<VAST version=\"3.0\">\n<Ad id=\"21341234\"><\/Ad>\n<\/VAST>', 'auid': 1}], 'seat': '1'};
+      const request = spec.buildRequests(bidRequests);
+      const expectedResponse = [
+        {
+          'requestId': '659423fff799cb',
+          'cpm': 1.15,
+          'creativeId': 1,
+          'dealId': undefined,
+          'width': undefined,
+          'height': undefined,
```
robertrmartinez: Same as above question.

Is there a reason why the adapter does not just set these values?

test/spec/modules/gridBidAdapter_spec.js
```diff
@@ -163,6 +163,7 @@ describe('TheMediaGrid Adapter', function () {
           'ad': '<div>test content 1</div>',
           'bidderCode': 'grid',
           'currency': 'USD',
+          'mediaType': 'banner',
```
robertrmartinez: Thank you for > 80% Test Coverage! 👍

modules/gridBidAdapter.js
robertrmartinez: According to Prebid Docs:
http://prebid.org/dev-docs/bidder-adaptor.html#interpreting-the-response

image

robertrmartinez: width and height are required.

TheMediaGrid: Hi, thank you for the catch, we have updated code, please review again.

PR #3546, opened 2019-02-12
OpenX Analytics is throwing an undefined error

deivydas:
## Type of change

- [x] Bugfix

## Description of change

Because bidId is not equal adId bid is always undefined with the old logic inside OnBidResponse event and an error is thrown. This issue was fixed by filling the adId inside bid map and searching for bids in AdUnitBidMap by requestId. All the changes were tested locally and are working as expected.


jimee02:
lgtm

PR #3494, opened 2019-01-29
added setTargeting code for custom key-value pairs to ozone adapter

afsheenb:
## Type of change
- [X] Feature

## Description of change
* Added support for custom key-value pairs in setTargeting;
* Move customData, lotameData and and ozoneData objects from top-level object into a separate ext.ozone object

PR #3550, opened 2019-02-12
added setTargeting code for custom key-value pairs

afsheenb:
## Type of change
- [X] Feature

## Description of change
* Added support for custom key-value pairs in setTargeting;
* Move customData, lotameData and and ozoneData objects from top-level object into a separate ext.ozone object

(Resubmitting this after rebasing, was originally PR #3494 )

### DIFF
--- a/modules/ozoneBidAdapter.js
+++ b/modules/ozoneBidAdapter.js
@@ -1,13 +1,13 @@
 import * as utils from '../src/utils';
 import { registerBidder } from '../src/adapters/bidderFactory';
 import { BANNER, NATIVE, VIDEO } from '../src/mediaTypes';
-import { targeting } from '../src/targeting';
+import {config} from '../src/config';
 
 const BIDDER_CODE = 'ozone';
 
-const OZONEURI = 'https://elb.the-ozone-project.com/openrtb2/auction';
-const OZONECOOKIESYNC = 'https://elb.the-ozone-project.com/static/load-cookie.html';
-const OZONEVERSION = '1.4.3';
+const OZONEURI = '//elb.the-ozone-project.com/openrtb2/auction';
+const OZONECOOKIESYNC = '//elb.the-ozone-project.com/static/load-cookie.html';
+const OZONEVERSION = '1.4.4';
 export const spec = {
   code: BIDDER_CODE,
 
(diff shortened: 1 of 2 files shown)


modules/ozoneBidAdapter.js
```diff
@@ -274,4 +338,44 @@ export function getTestQuerystringValue() {
...
+// we will use this instead of having to use in the html page: pbjs.bidderSettings = { ozone: { ... } } - see http://prebid.org/dev-docs/publisher-api-reference.html
+pbjs.onEvent('setTargeting', function(arrData) {
+  utils.logInfo('In the setTargeting event handler. Adding custom k/v to the ad request. _ozoneInternal = ', _ozoneInternal);
+  /*
+  In order to add a custom key/value pair you fundamentally need to do one thing:
+  Set new targeting for this key by calling setTargeting() on each window.googletag.pubads().getSlots() array.
+  */
+  // iterate over the slot objects, adding custom targeting parameters
+  utils.logInfo('going to iterate slots: ', window.googletag.pubads().getSlots());
+  window.googletag.pubads().getSlots().forEach(function (slot) {
```
mkendall07: adapters should not be modifying the DFP slots directly. Please drop this part of the code.

askrupert:
Rationale behind including this in the adaptor was to avoid publishers having to add additional on page code to expose these additional key values (which they have been requesting to be made available) ... 

is there any plans to let adaptors expose any custom KVPs in the future beyond the standard 5 that prebid exposes ?

idettman:
You should be able to add extra targeting in the `interpretResponse` by adding a property named `adserverTargeting` to the bidObjects returned. Example: `bidObject.adserverTargeting={oz_biddername_crid:'value'}`. Do a search in the prebid source for `adserverTargeting` to see how this is gets added to targeting

AskRupert-DM:
Thanks @idettman - I'll get a version submitted without this for now for you guys to review - but keen to solve this as the trying to expose some of the information that I am passing as KVPs is tricky to do on page as some of these are dynamic based on the bid response.

Would the following be acceptable - this would pretty much be all handled by Prebid as requested IMO:

pbjs.onEvent('setTargeting', function(arrData) {

// hardcoded values, just to illustrate the proposed solution:

  arrData.mpu['CUSTOM_KEY'] = 'CUSTOM_VALUE';
  arrData.leaderboard['CUSTOM_KEY'] = 'CUSTOM_VALUE';
  targeting.setTargetingForGPT({'mpu' : arrData.mpu});
  targeting.setTargetingForGPT({'leaderboard' : arrData.leaderboard});
});

The above would be added to our adaptor as you see in the previously submitted adaptor code...

afsheenb:
@idettman The setTargeting code has been refactored in the latest commit, please let us know if it's acceptable.

Cheers,

--afsheen

idettman:
Looks good to me, but I'm just waiting on @mkendall07 to review the change since this is out of my wheelhouse.

afsheenb:
Thanks @idettman - we've refactored the code to use the approved way of adding additional targeting and resubmitted.

PR #3554, opened 2019-02-13
fix IAS issue

AntoineJac:
IAS is no passing their key when the currency module is used. Indeed the letter case is important as the module will look for "USD" not "usd" and then return no bid.
Workaround: use the fix rate option and define:
rates:{usd:{CUR:1.4}}

## Type of change

- [ ] Bugfix
- [ ] Feature
- [ ] New bidder adapter   
- [ ] Build related changes
- [ ] CI related changes
- [ ] Other

## Description of change

```
{
  bidder: '<bidder name>',
  params: {
    // ...
  }
}
```

### DIFF
--- a/modules/iasBidAdapter.js
+++ b/modules/iasBidAdapter.js
@@ -103,7 +103,7 @@ function interpretResponse(serverResponse, request) {
     height: 200,
     creativeId: 434,
     dealId: 42,
-    currency: 'usd',
+    currency: 'USD',
     netRevenue: true,
     ttl: 360
   };


modules/iasBidAdapter.js
```diff
@@ -103,7 +103,7 @@ function interpretResponse(serverResponse, request) {
     height: 200,
     creativeId: 434,
     dealId: 42,
-    currency: 'usd',
+    currency: 'USD',
```
frankishh: 👍

PR #3573, opened 2019-02-20
Send placeholders for configured native assets

matthewlane:
## Type of change
- Feature

## Description of change
Transmitting some types of native assets via querystring parameters can cause issues. For example, large strings may exceed an adserver's url call limit, and special characters can get mangled. This PR introduces a configuration option to send placeholder values in adserver targeting, instead of asset values. The native creative template can then replace the placeholder values with the actual values through a postmessage request to Prebid ([a PR to the native functionality in prebid-universal-creative](https://github.com/prebid/prebid-universal-creative/pull/56) will do this automatically).

Given the configuration:
```JavaScript
pbjs.addAdUnits({
  code: slot.code,

  mediaTypes: {
    native: {
      body: {
        required: false,
        sendId: true
      },
      clickUrl: {
        required: true,
        sendId: true
      },
    },
  },

  bids: [...]
});
```

The `body` and `clickUrl` assets will be sent as placeholder values `hb_native_body:<adId>` and `hb_native_linkurl:<adId>`. All other native assets will be sent normally.

### DIFF
--- a/src/native.js
+++ b/src/native.js
@@ -180,3 +180,33 @@ export function getNativeTargeting(bid, bidReq) {
 
   return keyValues;
 }
+
+/**
+ *
+ */
+export function getAssetMessage(data, adObject) {
+  const message = {
+    message: 'gotLink',
+    adId: data.adId,
+    assets: [],
+  };
+
+  data.assets.forEach(nativeKey => {
+    const key = getNativeField(nativeKey);
+    const value = adObject.native[key];
+    message.assets.push({ key, value });
+  });
+
+  return message;
+}
+
+/**
+ *
+ */
+function getNativeField(key) {
+  for (let i in CONSTANTS.NATIVE_KEYS) {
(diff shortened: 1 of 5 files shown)


src/native.js
```diff
@@ -163,10 +163,41 @@ export function getNativeTargeting(bid) {
       value = value.url;
     }
 
+    const sendPlaceholder = deepAccess(
+      bidReq,
+      `mediaTypes.native.${asset}.usePlaceholder`
```
jsnellbaker: Any new thoughts on the name for this property?  I know we were throwing other ideas around in a different discussion.

cc @mkendall07

mkendall07:
yeah good call on docs. @matthewlane can you get the doc PR up today?

matthewlane:
Docs PR: https://github.com/prebid/prebid.github.io/pull/1175

PR #3574, opened 2019-02-21
update fun-hooks to latest and update api

snapwich:
## Type of change
- [x] Feature

## Description of change
Added a new feature to `fun-hooks` that allows you to add/remove `before` and `after` hooks before the hookable function is actually defined.  Added notes on the feature here: https://github.com/snapwich/fun-hooks#naming

Updated Prebid.js to use the new API which replaces the `hooks` object with a `.get` method to grab the hookable function.  If the hookable function is not found it will create a placeholder for all the `before`/`after` hooks until the hookable function is created.  It's allows you to "subscribe" to hooks you expect to eventually be present, in a sense.  Any hooks that are added using `.get` but never eventually have a hookable function created for them will throw a warning when `.ready` is called (because at that point you are saying you expect all hookable functions to be created).

@jsnellbaker check this out and see if it helps with the issue that prompted #3564 

Also includes fix from #3544

### DIFF
--- a/modules/currency.js
+++ b/modules/currency.js
@@ -3,7 +3,7 @@ import { STATUS } from '../src/constants';
 import { ajax } from '../src/ajax';
 import * as utils from '../src/utils';
 import { config } from '../src/config';
-import { hooks } from '../src/hook.js';
+import { getHook } from '../src/hook.js';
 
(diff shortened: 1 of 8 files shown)


snapwich:
Feature was added in: https://github.com/snapwich/fun-hooks/pull/2

jaiminpanchal27:
@jsnellbaker Please verify https://github.com/prebid/Prebid.js/pull/3574 and update

PR #3601, opened 2019-02-28
New bid adapter for SmileWanted

MaxSmileWanted:
- Smilewanted BidAdapter module and test spec added

## Type of change

- [ ] Bugfix
- [ ] Feature
- [x] New bidder adapter   
- [ ] Build related changes
- [ ] CI related changes
- [ ] Other

## Description of change

- Test parameters for validating bids, zoneId of ```1``` will always return an ad with requested CPM, zoneId of  ```2``` will always return an empty ad ```{nobid: true}```
```
{
    bidder: 'smilewanted',
    params: {
        zoneId: 1
    }
}
```
- contact email of the adapter’s maintainer : maxime@smilewanted.com

### DIFF
--- a/modules/smilewantedBidAdapter.js
+++ b/modules/smilewantedBidAdapter.js
@@ -6,27 +6,42 @@ export const spec = {
   code: 'smilewanted',
   aliases: ['smile', 'sw'],
 
+  /**
+   * Determines whether or not the given bid request is valid.
+   *
+   * @param {object} bid The bid to validate.
+   * @return boolean True if this is a valid bid, and false otherwise.
+   */
   isBidRequestValid: function(bid) {
     return !!(bid.params && bid.params.zoneId);
   },
 
+  /**
+   * Make a server request from the list of BidRequests.
+   *
+   * @param {BidRequest[]} validBidRequests A non-empty list of valid bid requests that should be sent to the Server.
+   * @return ServerRequest Info describing the request to the server.
+   */
   buildRequests: function(validBidRequests, bidderRequest) {
     return validBidRequests.map(bid => {
       var payload = {
         zoneId: bid.params.zoneId,
-        currencyCode: config.getConfig('currency.adServerCurrency'),
+        currencyCode: config.getConfig('currency.adServerCurrency') || 'EUR',
         bidfloor: bid.params.bidfloor || 0.0,
         tagId: bid.adUnitCode,
         sizes: bid.sizes.map(size => ({
           w: size[0],
           h: size[1]
         })),
-        pageDomain: utils.getTopWindowUrl(),
         transactionId: bid.transactionId,
         timeout: config.getConfig('bidderTimeout'),
         bidId: bid.bidId,
         prebidVersion: '$prebid.version$'
       };
+
+      if (bidderRequest && bidderRequest.refererInfo) {
+        payload.pageDomain = bidderRequest.refererInfo.referer || '';
+      }
 
       if (bidderRequest && bidderRequest.gdprConsent) {
         payload.gdpr_consent = bidderRequest.gdprConsent.consentString;
@@ -41,6 +56,12 @@ export const spec = {
     });
   },
 
+  /**
+   * Unpack the response from the server into a list of bids.
+   *
+   * @param {*} serverResponse A successful response from the server.
+   * @return {Bid[]} An array of bids which were nested inside the server.
+   */
   interpretResponse: function(serverResponse, bidRequest) {
     const bidResponses = [];
     var response = serverResponse.body;
@@ -56,7 +77,6 @@ export const spec = {
           currency: response.currency,
           netRevenue: response.isNetCpm,
           ttl: response.ttl,
-          referrer: utils.getTopWindowUrl(),
           adUrl: response.adUrl,
           ad: response.ad
         };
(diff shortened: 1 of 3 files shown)


bretg:
Docs PR https://github.com/prebid/prebid.github.io/pull/1180

modules/smilewantedBidAdapter.js
```diff
@@ -0,0 +1,84 @@
...
+    }
+    return bidResponses;
+  },
+
+  getUserSyncs: function(syncOptions, serverResponses) {
+    const syncs = []
+    if (syncOptions.iframeEnabled && serverResponses.length > 0) {
+      syncs.push({
+        type: 'iframe',
+        url: serverResponses[0].body.cSyncUrl
```
sumit116: It might be ideal to add a check that cSyncUrl has a valid value here. While we were testing with the test build, this property didn't have a value but it still try to load the iframe pixel.

modules/smilewantedBidAdapter.js
```diff
@@ -0,0 +1,84 @@
...
+
+  isBidRequestValid: function(bid) {
+    return !!(bid.params && bid.params.zoneId);
+  },
+
+  buildRequests: function(validBidRequests, bidderRequest) {
+    return validBidRequests.map(bid => {
+      var payload = {
+        zoneId: bid.params.zoneId,
+        currencyCode: config.getConfig('currency.adServerCurrency'),
```
sumit116: If we don't find getConfig(), do you want a default currencyCode here just like we default bidfloor below?

modules/smilewantedBidAdapter.js
```diff
@@ -0,0 +1,84 @@
...
+        bidfloor: bid.params.bidfloor || 0.0,
+        tagId: bid.adUnitCode,
+        sizes: bid.sizes.map(size => ({
+          w: size[0],
+          h: size[1]
+        })),
+        pageDomain: utils.getTopWindowUrl(),
```
sumit116: utils.getTopWindowUrl() is going to be depreciated. There is a property 'bidderRequest.refererInfo' which can provide the referral info. Can you use that instead of this?
Please refer to : http://prebid.org/dev-docs/bidder-adaptor.html#referrers

modules/smilewantedBidAdapter.js
```diff
@@ -0,0 +1,84 @@
...
+          requestId: JSON.parse(bidRequest.data).bidId,
+          cpm: response.cpm,
+          width: response.width,
+          height: response.height,
+          creativeId: response.creativeId,
+          dealId: response.dealId,
+          currency: response.currency,
+          netRevenue: response.isNetCpm,
+          ttl: response.ttl,
+          referrer: utils.getTopWindowUrl(),
```
sumit116: same as above.

MaxSmileWanted:
Hello, thank you for this code review, we made minor changes following your feedback, I think we should be good now.

modules/smilewantedBidAdapter.js
```diff
@@ -0,0 +1,84 @@
+import * as utils from '../src/utils';
+import { config } from '../src/config';
+import { registerBidder } from '../src/adapters/bidderFactory';
+
+export const spec = {
+  code: 'smilewanted',
+  aliases: ['smile', 'sw'],
+
```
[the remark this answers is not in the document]
MaxSmileWanted: I have added comments (based on comments from other partners).

modules/smilewantedBidAdapter.js
MaxSmileWanted: We check the value directly in the code, because the url will be static.

We forced the url into the test id zones (because it was conditioned by the existence of a GDPR string consent)

modules/smilewantedBidAdapter.js
MaxSmileWanted: We were going to manage it on the server side, but yes, it's actually smarter to do it directly in the code ;)

modules/smilewantedBidAdapter.js
MaxSmileWanted: We updated the referrer recovery method and modified unit tests to prevent the case where there is no referrer, which is the case when we did the "gulp test" command, it returned the error `Cannot read property" refererInfo' of undefined`, I don't know if it's intended ?

modules/smilewantedBidAdapter.js
MaxSmileWanted: Deleted we don't need it in the response

test/spec/modules/smilewantedBidAdapter_spec.js
```diff
@@ -0,0 +1,192 @@
...
+    const requestContent = JSON.parse(request[0].data);
+    expect(requestContent).to.have.property('zoneId').and.to.equal('1234');
+    expect(requestContent).to.have.property('currencyCode').and.to.equal('EUR');
+    expect(requestContent).to.have.property('bidfloor').and.to.equal(2.50);
+    expect(requestContent).to.have.property('sizes');
+    expect(requestContent.sizes[0]).to.have.property('w').and.to.equal(300);
+    expect(requestContent.sizes[0]).to.have.property('h').and.to.equal(250);
+    expect(requestContent.sizes[1]).to.have.property('w').and.to.equal(300);
+    expect(requestContent.sizes[1]).to.have.property('h').and.to.equal(200);
+    // expect(requestContent).to.have.property('pageDomain').and.to.equal('http://localhost/Prebid.js/integrationExamples/gpt/hello_world.html');
```
sumit116: Can you please remove the commented spec.

modules/smilewantedBidAdapter.js
jsnellbaker: @MaxSmileWanted Would you be able to point out which unit test failed and (depending on the size) provide a copy of the code in they way you attempted to modify it?  

The `refererInfo` exists on the `bidderRequest` object, which is the 2nd argument for the `spec.buildRequests()` function.  

So without seeing the changes, it's likely that either:
- the unit test's `spec.buildRequests()` call didn't populate the `bidderRequest` argument
- the supplied `bidderRequest` object was just missing its own `refererInfo` property and needed to setup with some mocked values

MaxSmileWanted: @jsnellbaker Yes you right the bidderRequest argument it's not populate through the unit test, so i append a forced value with this code in an another "it" :
```JAVASCRIPT
const request = spec.buildRequests(DEFAULT_PARAMS, {
  refererInfo: {
    referer: 'http://localhost/Prebid.js/integrationExamples/gpt/hello_world.html'
  }
});
```

PR #3605, opened 2019-03-01
fix two issues related to hb_uuid and hb_cache_id targeting keys

jsnellbaker:
## Type of change
- [x] Bugfix

**NOTE** This is the re-submission of #3568 (which was reverted due to a last minute code conflict that came when #3573 was merged to `master`).

## Description of change
This PR incorporates some fixes related to the `hb_uuid` and `hb_cache_id` targeting keys.

Some issues that were fixed:
- due to the order in which the code was executed, the `bid.adserverTargeting.hb_uuid` and `bid.adserverTargeting.hb_cache_id` fields weren't properly populated with the bid's `videoCacheKey` after it came back from Prebid Cache.  This caused other functions that pulled the targeting keys (like `pbjs.getAdserverTargetingForAdUnitCode()`) to be missing this information when it was otherwise expected to be there.

- a consequence to the above issue, when the publisher tried to overwrite the values for these keys in the `pbjs.bidderSettings` feature - the overwrites wouldn't work because the `bid.videoCacheKey` wasn't populated at the time these `bidderSettings` values were used.

The changes in this PR addresses the above issues through mainly moving the code around to populate the `bid.adserverTargeting` object once we know the bid is complete (ie back from prebid cache for that use-case).

Note - removed the `dfpAdServerVideo` code since it's now redundant.

### DIFF
--- a/src/auction.js
+++ b/src/auction.js
@@ -431,11 +431,6 @@ const callPrebidCache = hook('async', function(auctionInstance, bidResponse, aft
         doCallbacksIfTimedout(auctionInstance, bidResponse);
       } else {
         bidResponse.videoCacheKey = cacheIds[0].uuid;
-        let cacheTargetKeys = {
-          hb_uuid: cacheIds[0].uuid,
-          hb_cache_id: cacheIds[0].uuid
-        };
-        bidResponse.adserverTargeting = Object.assign(bidResponse.adserverTargeting || {}, cacheTargetKeys);
 
         if (!bidResponse.vastUrl) {
           bidResponse.vastUrl = getCacheUrl(bidResponse.videoCacheKey);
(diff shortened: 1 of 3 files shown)


src/auction.js
```diff
@@ -386,6 +386,10 @@ export function doCallbacksIfTimedout(auctionInstance, bidResponse) {
 
 // Add a bid to the auction.
 export function addBidToAuction(auctionInstance, bidResponse) {
+  let bidderRequests = auctionInstance.getBidRequests();
+  let bidderRequest = find(bidderRequests, bidderRequest => bidderRequest.bidderCode === bidResponse.bidderCode);
+  setupBidTargeting(bidResponse, bidderRequest);
```
jaiminpanchal27: Why are we calling this function for all formats ? We need `hb_uuid` and `hb_cache_id` only when cache is done i.e only for video as of now. Won't it be easier if we attach after hook function to `callPrebidCache`. That way these two targeting keys will only be added if we do caching.

jaiminpanchal27: Why are we calling this function for all formats ? We need `hb_uuid` and `hb_cache_id` only when cache is done i.e only for video as of now. Won't it be easier if we attach after hook function to `callPrebidCache`. That way these two targeting keys will only be added if we do caching.

jsnellbaker: Because it's not exclusively used by the cache related keys; it sets up all the targeting for the bid object.

PR #3611, opened 2019-03-05
YIELDONE adapter - add buildRequests payload params

kusapan:
## Type of change
- [x] Feature

## Description of change
- add buildRequests payload params
  - `uc` adUnitCode
  - ~~`tmax` timeout setting by publisher~~

 [sample page](http://webdemo.dac.co.jp/sakai/html/pbjsTimeout_test.html)

contact email of the adapter’s maintainer y1dev@platform-one.co.jp
- [x]  official adapter submission


modules/yieldoneBidAdapter.js
```diff
@@ -21,15 +21,23 @@ export const spec = {
...
         uid: bidId,
+        uc: unitCode,
         t: 'i'
       };
 
+      if (window.PREBID_TIMEOUT) {
+        payload.tmax = window.PREBID_TIMEOUT;
+      } else if (window.pbjsTimeout) {
+        payload.tmax = window.pbjsTimeout;
+      }
```
robertrmartinez: Hello @kusapan 

One suggestion is you can utilize the PBJS `config` in order to also get the bidderTimeout.

This value will always be set.

`payload.tmax = config.getConfig('bidderTimeout');`

example: https://github.com/prebid/Prebid.js/blob/8635c91942de9df4ec236672c39b19448545a812/modules/oneplanetonlyBidAdapter.js#L33

This value is set by the publisher by:

pbjs.setConfig({
   bidderTimeout: 850
});

If it is not explicitly set by the publisher, then it defaults to `3000`.

This number is the milliseconds that Prebid gives it's bidders to come back in time for the auction.

So I think it is a good value to send as your `tmax`.

If there are other reasonings or use cases I am unaware of as to why it would not be included please let me know!

test/spec/modules/yieldoneBidAdapter_spec.js
```diff
@@ -88,6 +88,11 @@ describe('yieldoneBidAdapter', function() {
       expect(request[0].data.w).to.equal('300');
       expect(request[0].data.h).to.equal('250');
     });
+
+    it('adUnitCode should be sent as uc parameters on any requests', function () {
+      expect(request[0].data.uc).to.equal('adunit-code1');
+      expect(request[1].data.uc).to.equal('adunit-code2');
+    });
```
robertrmartinez: Thanks for adding a unit test!

modules/yieldoneBidAdapter.js
kusapan: Our team talks about how to handle tmax.
In this PR, tmax was deleted.

PR #3612, opened 2019-03-05
Apply matchMedia to the top frame

benjaminclot:
## Type of change
- [x] Bugfix
- [ ] Feature
- [ ] New bidder adapter
- [ ] Build related changes
- [ ] CI related changes
- [ ] Other

## Description of change
In case Prebid.js is called from within an iFrame, matchMedia should be applied to window.top, not the containing iFrame.
This PR falls back to the legacy behavior in case of unfriendly iFrames.

### DIFF
--- a/src/sizeMapping.js
+++ b/src/sizeMapping.js
@@ -1,5 +1,5 @@
 import { config } from './config';
-import {logWarn, isPlainObject, deepAccess, deepClone} from './utils';
+import {logWarn, isPlainObject, deepAccess, deepClone, getWindowTop} from './utils';
 import includes from 'core-js/library/fn/array/includes';
 
 let sizeConfig = [];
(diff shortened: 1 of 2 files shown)


benjaminclot:
CI fails because of a Yieldbot adapter problem on Chrome 61.0.3163 (Windows 10.0.0)? 🤔

benjaminclot:
It would be great if this PR could make it into the next release. This is quite an annoying bug, as it also impacts postbid.

mike-chowla:
I'm doubtful it can get in the new next release because it needs to two reviews due to being a core change.

What scenarios do you expect this to fix?  I'm trying the think through all cases to understand why it makes sense that the media query is applied to top window rather than the iframe?

benjaminclot:
> What scenarios do you expect this to fix? I'm trying the think through all cases to understand why it makes sense that the media query is applied to top window rather than the iframe?
Why would you apply it to an iframe? Isn't the goal to know the full width of a window?

PR #3617, opened 2019-03-06
fix hb_cache_id for adpod bids

jsnellbaker:
## Type of change

- [x] Bugfix

## Description of change
In the `adpod` module, we're now populating the `bid.videoCacheKey` property in the bid object to ensure the `hb_cache_id` targeting key value is setup properly.

Also updated/added some unit tests to properly reflect the change of logic in the standard keys generated for video bids.

#### Context for the change
When #3528 was merged, it introduced a change that made the two targeting cache keys `hb_uuid` and `hb_cache_id` to be considered part of the standard set of adserver targeting keys (for video bids).  The values for these keys were populated by reading the `bid.videoCacheKey`.

With the adpod logic, we were generating our own key values for the `hb_cache_id` key - but we were not populating the `bid.videoCacheKey` (since there's different logic used for adpod bids when they're passed to the adserver and we assumed at the time it wasn't needed for the bid).  

As a result of the above - when the logic [here](https://github.com/prebid/Prebid.js/blob/master/src/auction.js#L504) ran, it overwrote the bid's `adserverTargeting.hb_cache_id` with an `undefined` because of the missing `bid.videoCacheKey`.  

The fix addresses the problem by populating the expected field in the bid when it's processed by the adpod module.

A note related to [this](https://github.com/prebid/Prebid.js/blob/master/modules/adpod.js#L134) line of code - we're keeping this value here as a backup in the case the publisher sets the `pbjs.bidderSettings.standard` property and doesn't setup the overwrite for the `hb_cache_id` key.

### DIFF
--- a/modules/adpod.js
+++ b/modules/adpod.js
@@ -132,6 +132,7 @@ function attachPriceIndustryDurationKeyToBid(bid, brandCategoryExclusion) {
   }
   bid.adserverTargeting[TARGETING_KEY_PB_CAT_DUR] = pcd;
   bid.adserverTargeting[TARGETING_KEY_CACHE_ID] = initialCacheKey;
+  bid.videoCacheKey = initialCacheKey;
   bid.customCacheKey = `${pcd}_${initialCacheKey}`;
 }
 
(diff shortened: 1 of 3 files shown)


jsnellbaker:
I updated the adpod unit test to also check the presence of the `bid.videoCacheKey` when it's reviewing the processed bids.

mkendall07:
Oops missed it. Thanks! 👍

jaiminpanchal27:
LGTM. Verified on demo page.

PR #3618, opened 2019-03-07
Fix multi bidder prebid server

AntoineJac:
Some bidders such as Pubmatic and Index are using a bid for each ad size as they don’t support multi-size. It is fine on client side except the fact that it will trigger 2 requests.
However, it is NOT working for Prebid server as only the last bid will be triggered server side.

As you can in the following test page, only the last Pubmatic bid (1410742@320x50) is sent to Prebid server and the first one is not present in the request (1410741@320x100):
https://jsfiddle.net/cbqzrjhx/

You can create an alias as a workaround and I have added this logic in the prebidBidServerBidAdapter to add an alias if the adUnit has multiple time the same bidder.

Thanks

## Type of change

- [ ] Bugfix
- [ ] Feature
- [ ] New bidder adapter   
- [ ] Build related changes
- [ ] CI related changes
- [ ] Other

## Description of change

```
{
  bidder: '<bidder name>',
  params: {
    // ...
  }
}
```

### DIFF
--- a/modules/prebidServerBidAdapter/index.js
+++ b/modules/prebidServerBidAdapter/index.js
@@ -461,7 +461,7 @@ const OPEN_RTB_PROTOCOL = {
       }
 
       // get bidder params in form { <bidder code>: {...params} }
-      const ext = adUnit.bids.reduce((acc, bid) => {
+      const ext = adUnit.bids.reduce((acc, bid, ind) => {
         const adapter = adapterManager.bidderRegistry[bid.bidder];
         if (adapter && adapter.getSpec().transformBidParams) {
           bid.params = adapter.getSpec().transformBidParams(bid.params, isOpenRtb());
--- a/modules/prebidServerBidAdapter/index.js
+++ b/modules/prebidServerBidAdapter/index.js
@@ -466,7 +466,13 @@ const OPEN_RTB_PROTOCOL = {
         if (adapter && adapter.getSpec().transformBidParams) {
           bid.params = adapter.getSpec().transformBidParams(bid.params, isOpenRtb());
         }
-        acc[bid.bidder] = (_s2sConfig.adapterOptions && _s2sConfig.adapterOptions[bid.bidder]) ? Object.assign({}, bid.params, _s2sConfig.adapterOptions[bid.bidder]) : bid.params;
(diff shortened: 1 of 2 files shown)


bretg:
My take - this should really be fixed in Prebid Server, not PBJS. Since the issue has been outstanding for a year, I'd like to understand why the issue can't persist for another couple of weeks while the Go and Java versions of the servers are patched.

Once we do patch this in PBS, this alias workaround would need to be removed.

AntoineJac:
@bretg , agree but the issue is PBJS will also have to pass the bid.bidder property several time.
Current code is "acc[bid.bidder] =" so if same bidder is present multiple time it will override the property value. Some change will have to be done on the PrebidServeBidAdapter.

bretg:
ok, but my vote would be to not create aliases. Adding @mkendall07 for his take. In otherwords, I propose the alias part of this PR gets removed.

AntoineJac:
Hi Bret, the logic then will have to be different. Here is the bidder is present several time I am passing it as a new property name (pubmatic, pubmaticAlias_1,...). 
I am force to do this as a Json object can't have twice the same property name. 

The 3 Solutions I am seeing are:

- If we keep the same logic we can pass same bidder string (pubmatic, pubmatic1, ...) and you will need in prebid server to do a regex matching the bidder string. Which globally is similar than using an alias...

- change the acc fields from object to array

- use the alias workaround

mike-chowla:
Speaking for PubMatic, we should really fix our ad slot validation in our Prebid Server adapter.  The client-side adapter now supports multi-sized ad units but our server-side adapter is failing to validate them.  I'll get this taken care of in the PubMatic Prebid Server adapter.

AntoineJac:
@mike-chowla Thanks for the update. Will be indeed better to wait for the PubMatic Prebid Server adapter thus. 
@bretg , should we ask other bidder to have a Prebid JS and Prebid Server adapter handling multisize?
If yes we can then close this PR. Thanks,

mkendall07:
+1 to Bret's comment. Let's address in PBS.

AntoineJac:
I have closed this PR as it will be fixed in PBS! Thanks,

PR #3648, opened 2019-03-18
Improve Digital adapter: use Prebid sizes param

jbartek25:
## Type of change

- [ ] Bugfix
- [X] Feature

## Description of change
New adapter config parameter "usePrebidSizes" added which tells the Improve Digital ad server to serve the creatives based on the sizes defined in Prebid ad unit config rather than placement sizes defined in the ad server.

Related doc PR: https://github.com/prebid/prebid.github.io/pull/1211


modules/improvedigitalBidAdapter.js
```diff
@@ -190,11 +190,15 @@ function getNormalizedBidRequest(bid) {
   if (keyValues) {
     normalizedBidRequest.keyValues = keyValues;
   }
-  if (localSize && localSize.w && localSize.h) {
+
+  if (config.getConfig('improvedigital.usePrebidSizes') === true && bid.sizes && bid.sizes.length > 0) {
+    normalizedBidRequest.format = bid.sizes;
```
harpere: I don't think `sizes` is normally included in the `bid` object.  Is this some special case?

jbartek25: It's supposed to be in the bid request, not response. Here's I believe where sizes field is populated: https://github.com/prebid/Prebid.js/blob/master/src/adapterManager.js#L98
Also, Appnexus adapter uses sizes the same way: https://github.com/prebid/Prebid.js/blob/master/modules/appnexusBidAdapter.js#L463

harpere: Cool, I wasn't aware of that.  thx for the pointers.

PR #3655, opened 2019-03-20
YIELDONE adapter - add outstream video renderer

kusapan:
## Type of change
- [x] Feature

## Description of change
- add outstream video renderer
- change to prioritize the size of banner instead of video

contact email of the adapter’s maintainer y1dev@platform-one.co.jp
- [x]  official adapter submission

### DIFF
--- a/test/spec/modules/yieldoneBidAdapter_spec.js
+++ b/test/spec/modules/yieldoneBidAdapter_spec.js
@@ -190,7 +190,8 @@ describe('yieldoneBidAdapter', function() {
         'ttl': 3000,
         'referrer': '',
         'mediaType': 'video',
-        'vastXml': '<!-- vast -->'
+        'vastXml': '<!-- vast -->',
+        'renderer': ''
       }];
       let result = spec.interpretResponse(serverResponseVideo, bidRequestVideo[0]);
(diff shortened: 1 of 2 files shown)


test/spec/modules/yieldoneBidAdapter_spec.js
```diff
@@ -190,7 +190,8 @@ describe('yieldoneBidAdapter', function() {
         'ttl': 3000,
         'referrer': '',
         'mediaType': 'video',
-        'vastXml': '<!-- vast -->'
+        'vastXml': '<!-- vast -->',
+        'renderer': ''
```
bretg: I would think that there could be a better unit test here checking that the renderer comes out to be img.ak.impact-ad.jp/ic/pone/ivt/firstview/js/dac-video-prebid.min.js ?

kusapan: Hello @bretg 
Thank you for reviewing.

we added unit test about renderer.
please confirm it.

kusapan: Hello @bretg 
Could you confirm this PR?

PR #3657, opened 2019-03-20
Sonobi - Add ius param to bid request endpoint

JonGoSonobi:
## Type of change
- [x] Feature

## Description of change
Sonobi adapter will detect if it can drop iframe pixels and set it as the ius parameter to sonobi's bid request endpoint. If its not ok to drop iframe pixels then the bid request will not respond with iframe pixels.

- apex.prebid@sonobi.com


JonGoSonobi:
Closing this for now. Found an issue with using shoudBidderBeBlocked. Will reopen when we come up with a solution

bretg:
@jsnellbaker - please take a look - I'm not clear why the additional "shouldBidderBeBlocked" object was added to the internal API.

JonGoSonobi:
@bretg @jsnellbaker  I decided to try to reuse that function instead of essentially making a copy of it inside our adapter. So I put it on the userSync publicAPI. If you guys prefer a different approach let me know.

jsnellbaker:
@JonGoSonobi I'm still thinking over this approach and if we should do something different.  

Something to be aware of however are the semi-deprecated userSync settings (like `userSync.iframeEnabled` and `userSync.enabledBidders`).  These settings still work and there may still be a number of publishers using these settings if they haven't updated their implementation.  

If they are using them - the function you imported wouldn't give you an accurate picture.

JonGoSonobi:
@jsnellbaker right. We figured we wouldn't worry about those cases since we saw this comment.
"// TODO remove this else if code that supports deprecated fields (sometime in 2.x); for now - only run if filterSettings config is not present". Looks like there is some plans to remove the use of those fields in an upcoming minor version. 

Maybe in the mean time to avoid straight up copying this code:
`if (usConfig.filterSettings) {
      if (shouldBidderBeBlocked(type, bidder)) {
        return utils.logWarn(`Bidder '${bidder}' is not permitted to register their userSync ${type} pixels as per filterSettings config.`);
      }
      // TODO remove this else if code that supports deprecated fields (sometime in 2.x); for now - only run if filterSettings config is not present
    } else if (usConfig.enabledBidders && usConfig.enabledBidders.length && usConfig.enabledBidders.indexOf(bidder) < 0) {
      return utils.logWarn(`Bidder "${bidder}" not permitted to register their userSync pixels.`);
    }`

we can move it userSync publicApi similar to what I tried with shouldBidderBeBlocked? Call it like: canBidderRegisterSync

jsnellbaker:
@JonGoSonobi That sounds like it could work-out; you'll technically need to incorporate logic to look at the `iframeEnabled` and `pixelEnabled` fields as well (since they're not part of that block of code).

If you want to put together the refactored changes, we can take a look at it and work from there.

JonGoSonobi:
@jsnellbaker  sounds like a plan, I'll follow up with the changes.

JonGoSonobi:
@jsnellbaker changes committed. Let me know what you think

JonGoSonobi:
@robertrmartinez any update on this one?

robertrmartinez:
@JonGoSonobi Sorry I was out, but this looks good.

I see a test is failing, we should look into why the merge with master has done this.

PR #3659, opened 2019-03-20
Automated functional tests for longform testpages

jsnellbaker:
## Type of change
- [x] Other

## Description of change
This PR adds a set of functional tests to validate some longform test pages.  

The tests just focus on the Prebid objects that are generated (ie the bids and the targeting keys), not the ads playing through the video player itself.  The auction request is controlled through a button object on the pages so that it would allow enough time for the page to fully initialize (especially the longform category data stored in localStorage).  

Each of the tests follow the basic steps:
* load the corresponding page
* wait for the `Load Prebid` button to appear and then click it (to run the auction)
* wait for the tables on the page to populate and check the results

The command to run these tests is: `gulp lfe2e` (name can be changed)

**NOTE** The tests are configured to use a number of retries in case their first attempt fails.  Some of the browserstack browsers (namely IE11 now) have been a bit flaky while trying to load these pages.  For example, they'll load the 3 out of the 4 pages in the same browser without issue, but then fail to load it for the last page (the error shown in the browser implies there was a connection issue or not found).  Then when you rerun the suite, everything would work fine.

**ADDITIONAL NOTE** As part of this PR, I removed the old E2E test code from the project to avoid confusion between two sets of e2e tests (since it wasn't completely working anyways).


integrationExamples/longform/basic_wo_requireExactDuration.html
```diff
@@ -0,0 +1,226 @@
...
+              }
+            });
+          }
+        });
+      });
+    }
+
+  </script>
+  <script type="text/javascript">
+    function getIndustry(id) {
```
jaiminpanchal27: We should create common.js and move all functions in that file ? They are same for all the long form pages

integrationExamples/longform/basic_wo_requireExactDuration.html
```diff
@@ -0,0 +1,226 @@
...
+    integrity="sha384-JjSmVgyd0p3pXB1rRibZUAYoIIy6OrQ6VrjIEaFf/nJGzIxFDsf4x0xIM+B07jRM"
+    crossorigin="anonymous"></script>
+  <link rel="stylesheet" href="https://stackpath.bootstrapcdn.com/bootstrap/4.3.1/css/bootstrap.min.css"
+    integrity="sha384-ggOyR0iXCbMQv3Xipma34MD+dH/1fQ784/j6cY/iJTQUOhcWr7x9JvoRxT2MZw1T" crossorigin="anonymous">
+  <link rel="stylesheet" type="text/css" href="//netdna.bootstrapcdn.com/font-awesome/3.0.2/css/font-awesome.css" />
+  <!-- <script type="text/javascript" src="//adm.fwmrm.net/p/vitest-js/AdManager.js"></script> -->
+  <script type="text/javascript" src="https://mssl.fwmrm.net/libs/adm/6.24.0/AdManager-debug.js"></script>
+  <!-- <script type="text/javascript" src="player.js"></script> -->
+  <style type="text/css">
+    #videoPlayer {
```
jaiminpanchal27: Move css to separate file style.css

test/spec/lfe2e/specs/basic_w_custom_adserver_translation.spec.js
```diff
@@ -0,0 +1,64 @@
...
+const ACCEPTED_CATS = ['Food', 'Retail Stores/Chains', 'Pet Food/Supplies', 'Travel/Hotels/Airlines', 'Automotive', 'Health Care Services'];
+const ACCEPTED_CPMS = ['15.00', '14.00', '13.00', '10.00'];
+const customKeyRegex = /\d{2}\.\d{2}_\d{1,3}_\d{2}s/;
+const uuidRegex = /(\d|\w){8}-((\d|\w){4}-){3}(\d|\w){12}/;
+
+describe('longform ads using custom adserver translation file', function() {
+  this.retries(3);
+  it('process the bids successfully', function() {
+    browser
+      .url('http://test.localhost:9999/integrationExamples/longform/basic_w_custom_adserver_translation.html?pbjs_debug=true')
```
jaiminpanchal27: What happens if test.localhost is not defined in my hosts file ?

integrationExamples/longform/basic_wo_requireExactDuration.html
jsnellbaker: I will create this and update the pages.

integrationExamples/longform/basic_wo_requireExactDuration.html
jsnellbaker: Will do.

test/spec/lfe2e/specs/basic_w_custom_adserver_translation.spec.js
jsnellbaker: I will test this and determine if it's an issue.

jsnellbaker:
@jaiminpanchal27 I pushed in a set of updates based on your feedback; please take another look when you have the chance.

browsers.json
```diff
@@ -1,65 +1,65 @@
 {
-  "bs_ie_14_windows_10": {
+  "bs_edge_16_windows_10": {
```
mkendall07: These changes are going to impact unit tests as well correct?

jsnellbaker: The new browsers would, but the unit tests are still running successfully.

jsnellbaker:
@mkendall07 Based on what we discussed today, I have added IE11 back into the `browsers.json` so it will be included when running normal unit tests.  I have adjusted the `wdio.conf.js` file to remove this browser from the e2e tests.  

Both sets of tests are working.  Please review when you have the chance.

mkendall07:
LGTM. Although looks like circlecI failed. I restarted the build

gulpfile.js
```diff
@@ -244,6 +231,13 @@ function newKarmaCallback(done) {
 function test(done) {
   if (argv.notest) {
     done();
+  } else if (argv.lfe2e) {
```
mkendall07: can you change this to e2e?

jsnellbaker: will do

jsnellbaker: Pushed a new commit with the updates.

Issue #3664, opened 2019-03-21
hb_cache_id is not set on DFP video request in Prebid 2.7


nanek:
## Type of issue
bug

## Description
hb_cache_id is not set as parameter on DFP video request in Prebid 2.7. Works in Prebid 2.3. I can see that videoCacheKey is set properly on bidResponse.

I think it might be related to https://github.com/prebid/Prebid.js/pull/3605/files#diff-f9e9cab460a9c83c3110b246bd4326e2L165

This results in making a request to `https://prebid.adnxs.com/pbc/v1/cache?uuid=` which is invalid as it is missing the uuid. This then causes a VAST 301 error.

## Steps to reproduce

I'm pretty sure it is happening for multiple partners, but I'm testing with oneVideo right now and can confirm it there.

````
 var adTagUrl = pbjs.adServers.dfp.buildVideoUrl({
      adUnit: videoAdUnit,
      params: params,
      bid: bidWithHighestCpm})
````

## Platform details
Prebid 2.7

**Comment by jaiminpanchal27** [comment]
@nanek Thanks for reporting this. We are working on this and will be fixed in next release. We will be adding both the keys back in dfp module.

#### Summary of what happened
`hb_uuid` and `hb_cache_id` are both standard keys for video. In 2.4 and before, if publisher overwrites the standard keys using pbjs.bidderSettings then both these were getting set to undefined due to the order in which some core operations were getting executed.

Hence we were adding both `hb_uuid` and `hb_cache_id` with the targeting data from dfp module.
Now in 2.5 we fixed bug explained above. So if you are to overwrite standard keys, you should also add
```
{
  key: 'hb_cache_id',
  val: function (bidResponse) {
    return bidResponse.videoCacheKey;
  }
}, {
  key: 'hb_uuid',
  val: function (bidResponse) {
    return bidResponse.videoCacheKey;
  }
}
```

nanek:
I do overwrite the standard keys, and I don't set the hb_cache_id key. I'll do that, thanks!

PR #3668, opened 2019-03-22
Add hb_uuid and hb_cache_id back to dfp module

jaiminpanchal27:
## Type of change
- [x] Bugfix

## Description of change
Add hb_uuid and hb_cache_id back to dfp module
Fixes https://github.com/prebid/Prebid.js/issues/3664

### DIFF
--- a/modules/dfpAdServerVideo.js
+++ b/modules/dfpAdServerVideo.js
@@ -158,6 +158,9 @@ function getCustParams(bid, options) {
 
   const optCustParams = deepAccess(options, 'params.cust_params');
   let customParams = Object.assign({},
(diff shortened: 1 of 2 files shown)


modules/dfpAdServerVideo.js
```diff
@@ -158,6 +158,9 @@ function getCustParams(bid, options) {
 
   const optCustParams = deepAccess(options, 'params.cust_params');
   let customParams = Object.assign({},
+    { hb_uuid: bid && bid.videoCacheKey },
```
mkendall07: can you add a comment here with a link to the issue so folks looking at the code will understand?

PR #3674, opened 2019-03-25
Ensure width and height are integers

benjaminclot:
## Type of change
- [x] Bugfix
- [ ] Feature
- [ ] New bidder adapter
- [ ] Build related changes
- [ ] CI related changes
- [ ] Other

## Description of change
Some bid responses return width and height as strings. Let's follow the docs and ensure that they indeed are integers, avoiding some strict comparison surprises.

### DIFF
--- a/src/adapters/bidderFactory.js
+++ b/src/adapters/bidderFactory.js
@@ -419,10 +419,9 @@ export function getIabSubCategory(bidderCode, category) {
 
 // check that the bid has a width and height set
 function validBidSize(adUnitCode, bid, bidRequests) {
-  bid.width = parseInt(bid.width, 10) || 0;
-  bid.height = parseInt(bid.height, 10) || 0;
-
-  if (bid.width && bid.height) {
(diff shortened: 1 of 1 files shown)


src/adapters/bidderFactory.js
```diff
@@ -419,7 +419,10 @@ export function getIabSubCategory(bidderCode, category) {
 
 // check that the bid has a width and height set
 function validBidSize(adUnitCode, bid, bidRequests) {
-  if ((bid.width || bid.width === 0) && (bid.height || bid.height === 0)) {
+  bid.width = parseInt(bid.width, 10) || 0;
+  bid.height = parseInt(bid.height, 10) || 0;
+
+  if (bid.width && bid.height) {
```
mkendall07: there's a subtle change here. If either `width` or `height` are 0 here then we will not return true. I don't know if we want to do that or not but it's a change in the previous behavior.

benjaminclot: OK, will fix.

PR #3675, opened 2019-03-25
support userId module

HolzAndrew:
## Type of change

- [x] Other

## Description of change
Adds support for User ID module


modules/yieldmoBidAdapter.js
```diff
@@ -302,3 +310,11 @@ function isSuperSandboxedIframe() {
 function isMraid() {
   return !!(window.mraid);
 }
+
+function getUserId(request) {
+  let userId;
+  if (request && request.userId && typeof request.userId === 'object') {
+    userId = request.userId;
+  }
+  return userId;
```
MelodyLi2017: why don't you just return userId.pucid if it exists?

PR #3696, opened 2019-03-30
fix bidTimeout event (#3583)

### LINKED ISSUE #3583: bidTimeout event seems to be broken after 1.24
## Type of issue
With version 1.24 and up until the latest version it seems as if the analytics event "bidTimeout" is broken.

## Description
If one bidder times out, all bidders that do not bid are included in the bidTimeout event, even if they in fact responded in time.

In 1.23 and before that, only the bidder(s) that actually timed out is included in the array passed into the analytics adapter.

## Steps to reproduce
Setup a prebid auction with several bidders. Force one bidder to timeout. I used Fiddler in Windows to set a latency for one of the bidder endpoints to above the auction timeout.

### Expected results
Only the bidder that actually timed out should be included in the list of bidders sent in the bidTimeout event.

### Actual results
Bidders that responded in time gets included in the list.

## Platform details
prebid.js >= 1.24

## Other information
This was the major change that was introduced in 1.24. Not sure if it's related but seems to be a prime suspect:
https://github.com/prebid/Prebid.js/pull/3024

If this is not a bug, but intended behavior, please explain how to collect information about timed out bidders. Can I count on events being fired in sequence? If so I could use the "bidderDone" event that seems to be fired before the "bidTimeout" event for all bidders that didn't actually timeout.

thewizarodofoz:
## Type of change

- [x] Bugfix
- [ ] Feature
- [ ] New bidder adapter   
- [ ] Build related changes
- [ ] CI related changes
- [ ] Other

## Description of change
In Prebid 2.x the BID_TIMEOUT event was emitted for bidders which responded in time but yielded no bids. This is now fixed - a bidder which responded in time is considered a timely bidder, even if it responded with no bids.

This fixes #3583 . 

**NOTE**: When a bidder responds, we assume that it responded to all its bid requests in the auction. This might not be 100% desired behaviour, but it was the behaviour in 1.x so I decided to keep it this way.

**NOTE2**: Even an error response is considered a timely response.

### DIFF
--- a/src/adapterManager.js
+++ b/src/adapterManager.js
@@ -252,7 +252,7 @@ adapterManager.makeBidRequests = function(adUnits, auctionStart, auctionId, cbTi
   return bidRequests;
 };
 
-adapterManager.callBids = (adUnits, bidRequests, addBidResponse, doneCb, requestCallbacks, requestBidsTimeout) => {
+adapterManager.callBids = (adUnits, bidRequests, addBidResponse, doneCb, requestCallbacks, requestBidsTimeout, onTimelyResponse) => {
   if (!bidRequests.length) {
     utils.logWarn('callBids executed with no bidRequests.  Were they filtered by labels or sizing?');
     return;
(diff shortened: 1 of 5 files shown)


GLStephen:
@thewizarodofoz I don't think there is enough detail in the timeout event to distinguish the sub requests for a particular adunit auction. Does this distinguish between ad units though? A timeout of one adunit's bid reqeusts versus another is valuable diagnostically and not all bidders bundle their requests into one.

thewizarodofoz:
@GLStephen I agree. I couldn't find a way to distinguish between ad units currently, because a bidder gets all the ad units' params as a bundle, and the breaking down into one or more requests is up to the bidder. Then the bidder informs the auction object only about valid bid responses. Meaning that no bid on an ad unit is just like no response at all.

Obviously anything is possible, but as far as I can see, in order to distinguish between ad units, bidders will need to inform the auction about the result of each request (wether it has a bid or not). It will require a pretty dramatic change and an update of all bidders.

Maybe @mkendall07 or @jsnellbaker will have better ideas about this.

jsnellbaker:
@thewizarodofoz Can you please resolve the code conflicts?

PR #3707, opened 2019-04-03
add console message when number of adunits exceeds point

jsnellbaker:
## Type of change
- [x] Feature

## Description of change
This PR adds a console message when the number of `adUnit`s for an auction exceeds a certain amount (currently set to 15).  This is to help notify publishers of their setup and to potentially review it if the number seems incorrect or too high.

### DIFF
--- a/src/prebid.js
+++ b/src/prebid.js
@@ -70,6 +70,11 @@ function setRenderSize(doc, width, height) {
 }
 
 export const checkAdUnitSetup = hook('sync', function (adUnits) {
+  let adUnitsLen = adUnits.length;
+  if (adUnitsLen > 15) {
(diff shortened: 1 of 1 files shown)


src/prebid.js
```diff
@@ -70,6 +70,11 @@ function setRenderSize(doc, width, height) {
 }
 
 export const checkAdUnitSetup = hook('sync', function (adUnits) {
+  let adUnitsLen = adUnits.length;
+  if (adUnitsLen > 15) {
+    utils.logInfo(`Current auction contains ${adUnitsLen} adUnits.`);
```
jaiminpanchal27: Should we add auction id in the message as well ? It will help them track down everything related to auction.

jsnellbaker: I updated the log message to include some additional info.  Please take a look.

PR #3735, opened 2019-04-10
PrebidManager Analytics: init module

Prebid-Manager:
## Type of change

- [x] Other (New Analytics adapter)

## Description of change

New Analytics Adapter for new Prebid Manager product (coming soon). See <a href="http://prebidmanager.com/">Prebid Manager</a>

- contact email of the adapter’s maintainer: admin@prebidmanager.com


robertrmartinez:
Starting review

Prebid-Manager:
@robertrmartinez 
I added PR fro documentation: https://github.com/prebid/prebid.github.io/pull/1269

Prebid-Manager:
@robertrmartinez have you any questions?

robertrmartinez:
@Prebid-Manager Hello!

Are you aware that the `DEFAULT_EVENT_URL` is not really used.

I have a test page that simply calls:

```
pbjs.enableAnalytics({
                'provider': 'prebidmanager',
                'options': {
                    bundleId: 'xxxxxxxx-xxxx-xxxx-xxxx-xxxxxxxxxxxx'
                }
            });
```

And this ends up sending your analytics messages to undefined:

image
Are you expecting this?


Also, are you sure you want to be doing `text/plain`?

I see you are prepending your `_VERSION` to the stringified post body, when you are already passing it in the post body as a param.
(Obviously this is up to you of course, makes it easier to debug in console if `application/json`)

image


Lastly, your test coverage is decent enough, 84%, But the only test is to ensure that the events are emitted. Would be nice to have some real validation against possible body contents, url called, etc.

Like setting:
```
pbjs.enableAnalytics({
                'provider': 'prebidmanager',
                'options': {
                    bundleId: 'xxxxxxxx-xxxx-xxxx-xxxx-xxxxxxxxxxxx',
                    url: 'https://endpoint.prebidmanager.com/endpoint'
                }
            });
```
Actually sends it to the right endpoint.

Or for instance, that not including `options.url` in the enable still sends it to your default (Which currently it does not.) Unsure if this is intended.

Prebid-Manager:
@robertrmartinez, Hello! Thank you for deep review.

- We fixed work with DEFAULT_EVENT_URL.
- We are sure about text/plain at the moment
- version is removed from json
- more tests are added. We verify requests and config options.

modules/prebidmanagerAnalyticsAdapter.js
```diff
@@ -0,0 +1,193 @@
...
+ */
+const DEFAULT_EVENT_URL = 'https://endpoint.prebidmanager.com/endpoint'
+const analyticsType = 'endpoint';
+const analyticsName = 'Prebid Manager Analytics: ';
+
+var utils = require('src/utils');
+var CONSTANTS = require('src/constants.json');
+let ajax = ajaxBuilder(0);
+
+function generateUUID() { // Public Domain/MIT
```
jaiminpanchal27: Any specific reason to not use https://github.com/prebid/Prebid.js/blob/master/src/utils.js#L96

modules/prebidmanagerAnalyticsAdapter.js
```diff
@@ -0,0 +1,193 @@
...
+function sendEvent(event) {
+  _eventQueue.push(event);
+  utils.logInfo(`${analyticsName}Event ${event.eventType}:`, event);
+
+  if (event.eventType === CONSTANTS.EVENTS.AUCTION_END) {
+    flush();
+  }
+}
+
+setInterval(flush, 1000);
```
jaiminpanchal27: I think this can be moved inside `enableAnalytics` function. So it will only get called when this module is enabled

modules/prebidmanagerAnalyticsAdapter.js
```diff
@@ -0,0 +1,193 @@
...
+let prebidmanagerAnalytics = Object.assign(adapter({url: DEFAULT_EVENT_URL, analyticsType}), {
+  track({eventType, args}) {
+    handleEvent(eventType, args);
+  }
+});
+
+prebidmanagerAnalytics.originEnableAnalytics = prebidmanagerAnalytics.enableAnalytics;
+prebidmanagerAnalytics.enableAnalytics = function (config) {
+  initOptions = config.options;
+  initOptions.url = initOptions.url || DEFAULT_EVENT_URL;
```
robertrmartinez: If a user enables the `prebidManager` Analytics without providing at least an empty `options` param, then this will cause an error and Prebid will not work:

Example:

```javascript
pbjs.enableAnalytics({
                'provider': 'prebidmanager'
            });
```

image


Please add a check to make sure `config.options` exists before assigning it to `initOptions`

Perhaps:
```javascript
initOptions = config.options || {};
```
?

modules/prebidmanagerAnalyticsAdapter.js
Prebid-Manager: @jaiminpanchal27 , no any specific reason. Commited with prebid utils function.

modules/prebidmanagerAnalyticsAdapter.js
Prebid-Manager: Moved to `enableAnalytics `function

modules/prebidmanagerAnalyticsAdapter.js
Prebid-Manager: done. Thank you.

Prebid-Manager:
@robertrmartinez , @jaiminpanchal27 
Сhanges on comments are made

robertrmartinez:
Thanks @Prebid-Manager I will take a look!

robertrmartinez:
Looks good to me

@jaiminpanchal27 Did you want to take a look?

robertrmartinez:
@Prebid-Manager Merged

PR #3743, opened 2019-04-11
Sonobi Adapter - Support UnifiedId/tradedesk from user id module

JonGoSonobi:
## Type of change
- [x] Feature

## Description of change
Added Sonobi Adapter support for Unified/Tradedesk ID

- apex.prebid@sonobi.com

https://github.com/prebid/prebid.github.io/pull/1258

### DIFF
--- a/modules/sonobiBidAdapter.js
+++ b/modules/sonobiBidAdapter.js
@@ -78,6 +78,12 @@ export const spec = {
       payload.hfa = `PRE-${validBidRequests[0].crumbs.pubcid}`;
     }
 
+    if (deepAccess(validBidRequests[0], 'userId.tdid')) {
+      payload.tdid = validBidRequests[0].userId.tdid
+    } else if (deepAccess(validBidRequests[0], 'crumbs.tdid')) {
+      payload.tdid = validBidRequests[0].crumbs.tdid
+    }
+
     if (validBidRequests[0].params.referrer) {
       payload.ref = validBidRequests[0].params.referrer;
     }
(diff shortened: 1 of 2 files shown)


modules/sonobiBidAdapter.js
```diff
@@ -78,6 +78,12 @@ export const spec = {
       payload.hfa = `PRE-${validBidRequests[0].crumbs.pubcid}`;
     }
 
+    if (deepAccess(validBidRequests[0], 'userId.tdid')) {
+      payload.tdid = validBidRequests[0].userId.tdid;
+    } else if (deepAccess(validBidRequests[0], 'crumbs.tdid')) {
```
jsnellbaker: To clarify - is this backup meant to grab the information from the `pubCommonId` module?

If so, the property should be `crumbs.pubcid` instead of `crumbs.tdid`.

JonGoSonobi: No it is not meant to try to grab pubcid. I wasnt sure if tdid would be available on `crumbs`, so I handled that scenario just incase.

JonGoSonobi: @jsnellbaker  any update?

jsnellbaker: @JonGoSonobi Sorry about delayed response here; the `crumbs` object won't be created by the `userID` module - it's only set by the `pubCommonId`.

Since the code would never find this object, I would suggest to remove this back-up if statement just to make your code clear.

JonGoSonobi: @jsnellbaker sounds good. done.

jsnellbaker:
@JonGoSonobi Thanks for the update; LGTM

PR #3749, opened 2019-04-15
Detect ad blocker recovered requests + send dynamic parameters to adapter + minor fixes

bjorn-lw:
## Type of change

- [x] Feature
- [x ] Other

## Description of change
Added support for detecting if request was originally ad blocked, but recovered.
Added support for sending in dynamic parameters to adapter.
Some fixes for how ttr is calculated in analytics adapter.


bjorn-lw:
@idettman the failed CircleCI tests, are they caused by this PR or was there a problem with the build process?

PR #3753, opened 2019-04-15
BUGFIX: Rubicon Video - pricegranularity param should be lowercase

robertrmartinez:
## Type of change
- [X] Bugfix

## Description of change
Prebid-Server expects price granularity param to be all lowercase,
Camel case causes it to not consider it.


robertrmartinez:
@harpere @bretg 

I built and tested, can confirm PBS now accepts the pricegranularity passed in.

PR #3756, opened 2019-04-16
fix handling of gdpr object

hiseke:
## Type of change

- [x] Bugfix

## Description of change
gdrp data is erroneously retrieved from validBidRequest item instead of bidderRequest. This cannot work and has got to be fixed.


hiseke:
The backend does some checks against the account id, so that's why you got the 403 error. I just double checked, things are running fine now. Do you need anything else for the changes to be merged?

PR #3765, opened 2019-04-17
Add Imonomy network BidAdapter

imonomy:
## Type of change

- [x] Bugfix
- [ ] Feature
- [x] New bidder adapter   
- [ ] Build related changes
- [ ] CI related changes
- [ ] Other

## Description of change

```
{
  bidder: 'imonomy',
  params: {
        placementId: 'e69148e0ba6c4c07977dc2daae5e1577',
	hbid: '14567718624',
	floorPrice: 0.5
  }
}
```

tech@imonomy.com


bretg:
docs PR https://github.com/prebid/prebid.github.io/pull/1272

modules/imonomyBidAdapter.js
```diff
@@ -0,0 +1,126 @@
...
+const USYNCURL = '//b.imonomy.com/UserMatching/b/';
+
+export const spec = {
+  code: BIDDER_CODE,
+
+  /**
+  * Determines whether or not the given bid request is valid. Valid bid request must have placementId and hbid
+  *
+  * @param {BidRequest} bid The bid params to validate.
+  * @return boolean True if this is a valid bid, and false otherwise.
```
idettman: JSDoc @type declaration for boolean is missing brackets, should be consistent by changed to:
`* @return {boolean} True if this is a valid bid, and false otherwise.`

modules/imonomyBidAdapter.js
```diff
@@ -0,0 +1,126 @@
...
+  */
+  isBidRequestValid: bid => {
+    return !!(bid && bid.params && bid.params.placementId && bid.params.hbid);
+  },
+  /**
+  * Make a server request from the list of BidRequests.
+  *
+  * @param {validBidRequests[]} - an array of bids
```
idettman: JSDoc @type declaration is using incorrect type, change to:
`* @param {BidRequest[]} validBidRequests an array of bids` (Also either remove or add `-` to all JSDoc parameter names, you can choose either, just be consistent in your adapter file)

modules/imonomyBidAdapter.js
```diff
@@ -0,0 +1,126 @@
...
+      ts_as: time,
+      hb_placements: [],
+      hb_placement_bidids: {},
+      hb_floors: {},
+      cb: _generateCb(time),
+      tz: new Date().getTimezoneOffset(),
+    };
+
+    validBidRequests.forEach(bid => {
+      kbConf.hdbdid = kbConf.hdbdid || bid.params.hbid;
```
idettman: On line 57 `kbConf.hdbdid` will always be undefined at this point, so the `||` will only evaluate to else. Possible change:
`kbConf.hdbdid = bid.params.hbid;`

Same goes for line 58 `kbConf.encode_bid`, the `||` will only evaluate to else since `encode_bid` can only be undefined here. Possible change:
`kbConf.encode_bid = bid.params.encode_bid;`

modules/imonomyBidAdapter.js
```diff
@@ -0,0 +1,126 @@
...
+      data: JSON.stringify(payload)
+    };
+  },
+  /**
+  * Unpack the response from the server into a list of bids.
+  *
+  * @param {*} response A successful response from the server.
+  * @return {Bid[]} An array of bids which were nested inside the server.
+  */
+  interpretResponse: (response, request) => {
```
idettman: Since the `request` argument is not used, remove it from the function signature, so line 88 would be changed to: `interpretResponse: (response) => {`

modules/imonomyBidAdapter.js
```diff
@@ -0,0 +1,126 @@
...
+      }
+      kbConf.hb_placements.push(bid.params.placementId);
+    });
+
+    let payload = {};
+    if (!utils.isEmpty(tags)) {
+      payload = { bids: [...tags], kbConf: kbConf };
+    }
+
+    var endpointToUse = ENDPOINT
```
idettman: Avoid `var` when possible, change line 71 to:
`let endpointToUse = ENDPOINT`;  <-- also, add a semicolon here to be consistent

modules/imonomyBidAdapter.js
```diff
@@ -0,0 +1,126 @@
...
+      hb_placement_bidids: {},
+      hb_floors: {},
+      cb: _generateCb(time),
+      tz: new Date().getTimezoneOffset(),
+    };
+
+    validBidRequests.forEach(bid => {
+      kbConf.hdbdid = kbConf.hdbdid || bid.params.hbid;
+      kbConf.encode_bid = kbConf.encode_bid || bid.params.encode_bid;
+      kbConf.hb_placement_bidids[bid.params.placementId] = bid.bidId;
```
idettman: Currently on line 58, the `kbConf.encode_bid` value is always going to be `undefined`, since it is not declared/ assigned anywhere before this point.

So, change line 58 to something like: 
`kbConf.encode_bid = bid.params.encode_bid;`

modules/imonomyBidAdapter.js
```diff
@@ -0,0 +1,126 @@
...
+      if (bid.params.floorPrice) {
+        kbConf.hb_floors[bid.params.placementId] = bid.params.floorPrice;
+      }
+      kbConf.hb_placements.push(bid.params.placementId);
+    });
+
+    let payload = {};
+    if (!utils.isEmpty(tags)) {
+      payload = { bids: [...tags], kbConf: kbConf };
```
idettman: SInce the property names are the same for `kbConf`, you can use the obj shorthand syntax here:
`payload = { bids: [...tags], kbConf };`

modules/imonomyBidAdapter.js
```diff
@@ -0,0 +1,126 @@
...
+    }
+
+    var endpointToUse = ENDPOINT
+    if (kbConf.hdbdid) {
+      endpointToUse = endpointToUse.replace('00000', kbConf.hdbdid)
```
idettman: Add missing semicolon to line 73:
`endpointToUse = endpointToUse.replace('00000', kbConf.hdbdid);`

modules/imonomyBidAdapter.js
```diff
@@ -0,0 +1,126 @@
...
+  },
+  /**
+  * Unpack the response from the server into a list of bids.
+  *
+  * @param {*} response A successful response from the server.
+  * @return {Bid[]} An array of bids which were nested inside the server.
+  */
+  interpretResponse: (response, request) => {
+    const bidResponses = [];
+    try {
```
idettman: The `try/catch` here seems like overkill (try catch performance in JavaScript is generally poor). Most bid adapters handle response validation using only conditionals. Reference `modules/appnexusBidAdapter.js` between lines 184 - 206 to see what most adapters do for the response.

imonomy: Check please new conditionals implementation

modules/imonomyBidAdapter.js
imonomy: Good day to you!
We made all the changes you requested, except those which related to this one (row 57 & 58).
As you can see, we use cycle here, so kbConf.hdbdid and kbConf.encode_bid will be undefined only on the first iteration.

PR #3766, opened 2019-04-17
Add Video Reach adapter

VideoReach:
## Type of change

- [ ] Bugfix
- [ ] Feature
- [X] New bidder adapter   
- [ ] Build related changes
- [ ] CI related changes
- [ ] Other

## Description of change
Add Video Reach adapter

```
{
  bidder: 'videoreach',
  params: {
    TagId: 'MS8OW'
  }
}
```

- hello@videoreach.de

- https://github.com/prebid/prebid.github.io/pull/1273

### DIFF
--- a/modules/videoreachBidAdapter.js
+++ b/modules/videoreachBidAdapter.js
@@ -71,16 +71,16 @@ export const spec = {
       let params = '';
       if (gdprConsent && typeof gdprConsent.consentString === 'string') {
         if (typeof gdprConsent.gdprApplies === 'boolean') {
-          params += 'gdpr='+gdprConsent.gdprApplies+'&gdpr_consent='+gdprConsent.consentString;
+          params += 'gdpr=' + gdprConsent.gdprApplies + '&gdpr_consent=' + gdprConsent.consentString;
         } else {
-          params += 'gdpr_consent='+gdprConsent.consentString;
+          params += 'gdpr_consent=' + gdprConsent.consentString;
         }
       }
 
       var gdpr;
       if (SyncPixels) {
         SyncPixels.forEach(sync => {
-          gdpr = (params) ? ((sync.split('?')[1] ? '&':'?') + params) : '';
+          gdpr = (params) ? ((sync.split('?')[1] ? '&' : '?') + params) : '';
 
           syncs.push({
             type: 'image',
--- a/modules/videoreachBidAdapter.js
+++ b/modules/videoreachBidAdapter.js
@@ -1,0 +1,85 @@
+import {registerBidder} from '../src/adapters/bidderFactory';
+const utils = require('../src/utils');
+const BIDDER_CODE = 'videoreach';
+const ENDPOINT_URL = '//a.videoreach.de/hb/';
+
+export const spec = {
+  code: BIDDER_CODE,
+  supportedMediaTypes: ['video', 'banner'],
+
+  isBidRequestValid: function(bid) {
+    return !!(bid.params.TagId);
+  },
+
+  buildRequests: function(validBidRequests, bidderRequest) {
+    var data = {
+      referrer: utils.getTopWindowUrl(),
+      data: validBidRequests.map(function(bid) {
+        return {
+          TagId: utils.getValue(bid.params, 'TagId'),
+          bidId: utils.getBidIdParameter('bidId', bid),
+          bidderRequestId: utils.getBidIdParameter('bidderRequestId', bid),
+          auctionId: utils.getBidIdParameter('auctionId', bid),
+          transactionId: utils.getBidIdParameter('transactionId', bid)
+        }
+      })
+    };
+
+    if (bidderRequest && bidderRequest.gdprConsent) {
+      data.gdpr = {
+        consent_string: bidderRequest.gdprConsent.consentString,
+        consent_required: bidderRequest.gdprConsent.gdprApplies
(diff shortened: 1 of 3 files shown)


robertrmartinez:
Looking now

modules/videoreachBidAdapter.js
```diff
@@ -0,0 +1,85 @@
...
+          requestId: bid.bidId,
+          creativeId: bid.creativeId
+        };
+        bidResponses.push(bidResponse);
+      });
+    }
+    return bidResponses;
+  },
+
+  getUserSyncs: function(syncOptions, responses) {
```
robertrmartinez: Hello,

Did you know that you can pass GDPR information into your `getUserSyncs` and then pass it along to your sync server?
Not required, but here is an example:

https://github.com/prebid/Prebid.js/blob/master/modules/rubiconBidAdapter.js#L556

robertrmartinez:
I would like to clear up some of my confusion.

It seems as if `videoreach` is a pure video demand exchange?

But you are never sending back `bidResponse.mediaType` == `video`

Instead you pass nothing which defaults it to `banner`.

If a publisher wishes to use `videoreach` with other video demand partners this will cause problems.

Prebid requires that the full `VAST XML`, or a `videoCacheKey` is passed in the `bidResponse` object for any video demand.

Thus when constructing the Client Side caching it is able to store the VAST XML for future fetching.

For video Reach is looks like you return script tag which loads some javascript which then will pick out your VAST XML.

This will not work with many video prebid integrations that publishers have.

The recommendation is to:

1. Have your adserver either:
       A: Return the full `vastxml`, which then Prebid will cache it for you
       OR
       B: Return the vastxml URL which will point to the VAST XML.
       OR
       C: Return the vastUrl as well as the videoCacheKey to be retrieved from.
       
FOR A:
    - You would alter your bidAdapter to return this full vastXML in the `bidResponse.vastXml`
    - This would indicate to Prebid Core that this VAST XML needs to be cached and thus it would call the Prebid Cache server which the publisher had defined in 
```pbjs.setConfig( { cache: url = 'some cache url' });
```
     - Then, when prebid picks the winning video bid or however the publisher sets it up, it now hows the URL for the cached XML alongwith each bid containing its unique UUID to retrieve the VAST xml from the cache.
Exampel bid adapter using this: https://github.com/prebid/Prebid.js/blob/master/modules/pubmaticBidAdapter.js#L940

FOR B:
    - If you are storing your Vast XML on the server side and it is retrievable, then you can simply pass along two values:
       - `bidResponse.vastUrl`: full url to retrieve the vast from
       - `bidResponse.vastImpUrl`: an impression url which is used when wrapping the vastUrl into a prebid cache Vast URL before caching it client side.
Example bid Adapter using this: https://github.com/prebid/Prebid.js/blob/master/modules/appnexusBidAdapter.js#L375


For C:
    - If you do not want prebid to cache for you, then you then you can pass along a few more values in order to flag it to not be cached:

     - `bidResponse.vastUrl`: full url to retrieve the vast from
     - `bidResponse.videoCacheKey`: a unique identifier for each cache to retrieve it from.

Example bid adapter using this approach:
https://github.com/prebid/Prebid.js/blob/master/modules/rubiconBidAdapter.js#L467

Please note that When constructin DoubleClick calls, prebid sends the following KVPs when `VIDEO` is the mediaType:

```
{
     hb_uuid: //unique identifier to for each cached object
     hb_cache_host: // the hostname of the URL where this vast xml is cached
     hb_cache_path: // the path for the URL
}
```

When these three things are combined it `should` form a full URL which points to the cached url.
Example: https://github.com/prebid/Prebid.js/blob/master/modules/rubiconBidAdapter.js#L473

robertrmartinez:
Also, your Bid Request Server seems to not add port numbers in when adding the `Access-Control-Allow-Origin` header based on the `Origin` request header

image

This would make it hard for anyone testing locally with your adapter, not a blocker though, just wanted to let you know if you had not noticed yet.

VideoReach:
Hi Robert,

thank you for your helpful information.

We made following changes:
- [x] `Access-Control-Allow-Origin` add Port
- [x] `getUserSyncs` with GDPR
- [x] We removed `video` from `supportedMediaTypes` (maybe we'll add it in future versions)

Thanks!

robertrmartinez:
@VideoReach 

Taking a look now.

Thanks.

robertrmartinez:
@VideoReach Thanks for updating.

I still have a question regarding your example parameters and how to actually render your ads?

It seems as though your ad server is loading some JS which eventually looks like it is trying to get a VAST Url, or maybe some sort of vast player?

How will you be having bidders work with your adapter?

Do you plan on them not using the usual `pbjs.renderAd` to render your ad?
Is there extra configuration needed by the publisher to work with your adapter?

VideoReach:
@robertrmartinez Thanks for checking!

> It seems as though your ad server is loading some JS which eventually looks like it is trying to get a VAST Url, or maybe some sort of vast player?

Right, we load our own player (when winning), which is loading the winning VAST.

> Do you plan on them not using the usual pbjs.renderAd to render your ad?

So far this is not planned.

> Is there extra configuration needed by the publisher to work with your adapter?

No, we just set the HTML selector (in out ad server). The publisher only need to use the provided `TagId`. Our JS knows where to place the ad - in the example the selector element is: `#TextContent`

PR #3778, opened 2019-04-24
update e2e tests in README

jsnellbaker:
## Type of change
- [x] Other

## Description of change
Adding an update to the README in lieu of #3659 being merged.

### DIFF
--- a/README.md
+++ b/README.md
@@ -207,7 +207,7 @@ gulp test-coverage
 gulp view-coverage
 ```
 
-For end-to-end testing:
+For Prebid.org members with access to BrowserStack, additional end-to-end testing can be done with:
 
 ```bash
 gulp e2e-test --host=test.localhost
--- a/README.md
+++ b/README.md
@@ -214,7 +214,7 @@ gulp e2e-test --host=test.localhost
 ```
 
 To run these tests, the following items are required:
-- setup an alias of localhost in your `hosts` file; eg `127.0.0.1  test.localhost`.  Use this alias in the command-line argument.
+- setup an alias of localhost in your `hosts` file (eg `127.0.0.1  test.localhost`); note - you can use any alias.  Use this alias in the command-line argument above.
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -207,10 +207,20 @@ gulp test-coverage
 gulp view-coverage
 ```
 
-For end-to-end testing, edit the example file `./integrationExamples/gpt/pbjs_example_gpt.html`:
+For end-to-end testing:
 
-1. Change `{id}` values appropriately to set up ad units and bidders
-2. Set the path to Prebid.js in your example file as shown below (see `pbs.src`).
+```bash
+gulp e2e-test --host=test.localhost
```
jaiminpanchal27: Will it work if i use any other alias which is also defined in host file ? e.g my.localhost
If yes can you update the description accordingly.

jsnellbaker: I have that described below?

README.md
```diff
@@ -207,10 +207,20 @@ gulp test-coverage
...
+```
+
+To run these tests, the following items are required:
+- setup an alias of localhost in your `hosts` file (eg `127.0.0.1  test.localhost`); note - you can use any alias.  Use this alias in the command-line argument above.
+- access to [BrowserStack](https://www.browserstack.com/) account.  Assign the following variables in your bash_profile:
```
bretg: I don't see how we can require most users to have access to a browserstack account. I don't think even most of us PBJS regulars have access to the account.

I would suggest that reference to this test be added to https://github.com/prebid/Prebid.js/blob/master/RELEASE_SCHEDULE.md instead of this file.

Regular community members should have the previous end-to-end option IMO.

jaiminpanchal27: @bretg This is not a requirement. In our README we have mentioned gulp commands to do all the tasks. Since we updated our end to end test architecture, we are documenting the new commands here. Maybe we can add the language stating that it is optional ?

For prebid users, this is completely optional. It is not mandatory to run end to end tests.

bretg: Ok - I added a note here about needing to be a prebid.org member to do these tests. Am satisfied with that.

PR #3780, opened 2019-04-25
yieldlabBidAdapter - Use Slot Size

phtechno:
Add a possibility to use the Prebid-AdSlot-Size in the bid-response instead of YieldLab-AdSlot-Size if you use a different.
This allows us a better deal-handling.

## Type of change

## Description of change

Add the bidder-param "useSlotSize" to use different Slot-Sizes between Prebid and YieldLab. This allows different AdType-Handling for the same slot-size (deals, ...).

### DIFF
--- a/modules/yieldlabBidAdapter.js
+++ b/modules/yieldlabBidAdapter.js
@@ -81,8 +81,8 @@ export const spec = {
         const bidResponse = {
           requestId: bidRequest.bidId,
           cpm: matchedBid.price / 100,
-          width: customsize[0],
-          height: customsize[1],
+          width: bidRequest.params.useSlotSize !== undefined ? primarysize[0] : customsize[0],
+          height: bidRequest.params.useSlotSize !== undefined ? primarysize[1] : customsize[1],
           creativeId: '' + matchedBid.id,
           dealId: (matchedBid['c.dealid']) ? matchedBid['c.dealid'] : matchedBid.pid,
           currency: CURRENCY_CODE,


modules/yieldlabBidAdapter.js
```diff
@@ -81,8 +81,8 @@ export const spec = {
         const bidResponse = {
           requestId: bidRequest.bidId,
           cpm: matchedBid.price / 100,
-          width: customsize[0],
-          height: customsize[1],
+          width: bidRequest.params.useSlotSize !== undefined ? primarysize[0] : customsize[0],
```
bretg: I don't see the `useSlotSize` parameter in your documentation -- http://prebid.org/dev-docs/bidders.html#yieldlab

Is this a new param? If so, please submit a PR for the docs site. https://github.com/prebid/prebid.github.io/blob/master/dev-docs/bidders/yieldlab.md

phtechno: Hi @bretg 
I've changed the documentation: https://github.com/prebid/prebid.github.io/pull/1294

bretg:
Thanks for the doc update @phtechno , but thinking about this some more, I'm not understanding the size handling here. 

First, would like to make sure you're not relying on the AdUnit.sizes param anymore -- that's been deprecated for quite a while. You should be using AdUnit.mediaTypes.banner.sizes.

Second, and more importantly, your code is taking the first element of bidRequest.mediaTypes.banner.sizes as the size to use on the response. This seems odd because there can be several sizes on the adunit. e.g. `[[300,250],[300,600]]`.  By always selecting just the first element of the array, your bidder is saying you never bid with a 300x600 creative in this case? Most bidders return their creative size in the bidResponse.

bretg:
@phtechno - thoughts? It would be fine to continue legacy support for adunit.sizes, but the adapter also needs to support adunits.mediatype.banner.sizes. Thanks.

PR #3782, opened 2019-04-25
clear latest auction units

deekshithraop:
## Type of change

- [x] Bugfix

## Description of change

If `useBidCache` flag is set to false, we have to use only the bids from the latest auction while setting the targeting keys. 

Currently even if the flag is set to false, it will use the bids from the previous actions because `latestAuctionForAdUnit ` object stores latest auction for all the adUnits, even the ones for which the bids are requested for previous auctions. and that is being used to filter bids while setting targeting.

so these changes while setting the latest auction ids for each placement on new auctions, it will remove the auction ids for the adUnits from previous auctions.

### DIFF
--- a/src/prebid.js
+++ b/src/prebid.js
@@ -474,7 +474,10 @@ $$PREBID_GLOBAL$$.requestBids = hook('async', function ({ bidsBackHandler, timeo
     utils.logInfo(`Current auction ${auction.getAuctionId()} contains ${adUnitsLen} adUnits.`, adUnits);
   }
 
-  targeting.clearLatestAuctionKeys();
+  if (!config.getConfig('useBidCache')) {
+    targeting.clearLatestAuctionKeys();
+  }
+
   adUnitCodes.forEach(code => targeting.setLatestAuctionForAdUnit(code, auction.getAuctionId()));
   auction.callBids();
   return auction;
--- a/src/prebid.js
+++ b/src/prebid.js
@@ -474,6 +474,7 @@ $$PREBID_GLOBAL$$.requestBids = hook('async', function ({ bidsBackHandler, timeo
     utils.logInfo(`Current auction ${auction.getAuctionId()} contains ${adUnitsLen} adUnits.`, adUnits);
   }
 
+  targeting.clearLatestAuctionKeys();
   adUnitCodes.forEach(code => targeting.setLatestAuctionForAdUnit(code, auction.getAuctionId()));
   auction.callBids();
   return auction;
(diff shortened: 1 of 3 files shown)


snapwich:
The logic you're describing for this configuration flag sounds incorrect, so I'm not sure there is a bug here.  Let me describe the way it's supposed to work and you can let me know if you still think there is a bug.

> If useBidCache flag is set to false, we have to use only the bids from the latest auction while setting the targeting keys.

If bid caching is set to false, we only use bids from the latest auction _per ad unit_.  This means if an ad unit is refreshed, this flag prevents us from using bids for the previous auction results for that ad unit.  Is this not what you're seeing?

We should not be clearing the `latestAuctionForAdUnit` map per auction (a call to `requsetBids`).  Since Prebid.js allows concurrent auctions, this means one auction could clear the last auction entry for a concurrent auction causing its bids to get filtered before it gets to set targeting; that would be a bug.

deekshithraop:
@snapwich The bug is being reported by the below website, as below.
 
1. http://www.telkku.com/?pbjs_debug=true open that site, don’t scroll. Load until you’ll receive bid for element almad-widerectangle-1
2. Open one of the programs listed (it’s the tv guide)
 
Below print screens of the same steps if you don’t get bids for that placement:
 
Front page - bid for almad-widerectangle-1
image
Moving to “program page”. where we are only bidding the following ad units:
image
we are still setting key-values for almad-widerectangle-1 with the old hb_adid (same id as on front page)
image

deekshithraop:
The only place that i saw this obj (latestAuctionForAdUnit) being used is here(https://github.com/prebid/Prebid.js/blob/master/src/targeting.js#L220) while setting targeting keys.
 If the idea behind setting bid caching is below 
```If bid caching is set to false, we only use bids from the latest auction per ad unit. This means if an ad unit is refreshed, this flag prevents us from using bids for the previous auction results for that ad unit. Is this not what you're seeing?```

and not 
```If useBidCache flag is set to false, we have to use only the bids from the latest auction while setting the targeting keys```

Then is there any other way so that while setting targeting we would use only bids from the latest auction rather that latest auction per adUnits for which bids  are available.

snapwich:
Discussing this with @mkendall07 and @jaiminpanchal27 in prebid slack, but it sounds like the ultimate goal here (tying set targeting calls to individual auctions) is unrelated to bid caching.  Might require a public API change if we choose to add this functionality.

mkendall07:
since this changes the existing behavior for caching we are not going to merge.

Issue #3687, opened 2019-03-26
Proposal: Standardizing First Party Data


bretg:
## Type of issue
Feature

## Description
Provide a standard way for the page to supply first party data and control which bidders have access to it.

A number of adapters support taking key/value pairs as arguments, but they're all different. e.g.

- Rubicon takes `keywords`, `inventory` and `visitor` parameters
- AppNexus takes `keywords` and `user`
- OpenX takes `customParams`
- Index takes `firstPartyData`
- etc.

## Proposal

1) Establish a standard convention for where pages and apps can place first party data
2) Give the publisher control over which bidders are allowed to see the first party data
3) Adapters can then be updated to read from the standard location, mapping values to their bidder-specific locations.
4) We don't support AdUnit-specific user data, as that scenario doesn't appear to have use-cases.
5) Use OpenRTB conventions where possible

Proposed conventions

A) Global (cross-adunit) First Party Data open to all bidders
```
pbjs.setConfig({
   fpd: {
       context: {
           keywords: ["power tools"],
           search: "drill",
           content: { userrating: 4 },
           data: {
               pageType: "article",
               category: "tools"
           }
        },
        user: {
           keywords: ["a","b"],
           gender: "M",
           yob: "1984",
           geo: { country: "ca" },
           data: {
              registered: true,
              interests: ["cars"]
           }
        }
    }
});
```

B) Global (cross-adunit) First Party Data open only to a subset of bidders:
```
pbjs.setBidderConfig({
   bidders: ['bidderA', 'bidderB'],
   config: {
       fpd: {
           context: {
               keywords: ["power tools"],
               search: "drill",
               content: { userrating: 4 },
               data: {
                  pageType: "article",
                  category: "tools"
               }
            },
            user: {
               keywords: ["a","b"],
               gender: "M",
               yob: "1984",
               geo: { country: "ca" },
               data: {
                  registered: true,
                  interests: ["cars"]
               }
          }
      }
   }
});
```

C) AdUnit specific values (always available to all bidders)
```
    var adUnits = [{
        code: '/19968336/header-bid-tag-1',
        mediaTypes: {
            banner: {
                sizes: sizes
            }
        },
       fpd: {
         context: {
            adUnitSpecificContextAttribute: "123"
         }
        },
        bids: [{
           ...
        }]
    }];
```

Examples of adUnit-specific context values are: position and viewability. Note that user-specific first party data is not supported per-adunit.

[...]

PR #3783, opened 2019-04-26
pass site.{} into prebid server request

AntoineJac:
## Type of change
- [ ] Feature

## Description of change
Some publisher want to pass keywords to prebid server and current code only read pbjs.setConfig(app{}) not pbjs.setConfig(site{})
This code will pass the data in config.site to the prebid server request

## Other information
Prebid server is already able to read data passed in $.site.keywords

### DIFF
--- a/modules/prebidServerBidAdapter/index.js
+++ b/modules/prebidServerBidAdapter/index.js
@@ -257,10 +257,14 @@ function _appendSiteAppDevice(request) {
     request.app = config.getConfig('app');
     request.app.publisher = {id: _s2sConfig.accountId}
   } else {
-    request.site = {
-      publisher: { id: _s2sConfig.accountId },
-      page: utils.getTopWindowUrl()
+    if (typeof config.getConfig('site') === 'object') {
+      request.site = config.getConfig('site');
     }
+    if (!request.site) {
+      request.site = {};
+    }
+    request.site.publisher = {id: _s2sConfig.accountId}
+    request.site.page = utils.getTopWindowUrl()
   }
   if (typeof config.getConfig('device') === 'object') {
     request.device = config.getConfig('device');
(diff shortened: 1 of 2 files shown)


bretg:
Thanks for the PR @AntoineJac, but would like to bring this in line with the design for first party data we finalized in https://github.com/prebid/Prebid.js/issues/3687

`setConfig({site})` is not currently part of the PBJS externally facing API. IMO, it would be better not to have multiple ways to specify keywords.

Finally, I'm not comfortable that this proposed change has the side-effect of removing site.publisher and site.page if the user doesn't provide them.

So my requested change is to specify keywords in `setConfig({context})` rather than `setConfig({site})`

This change should ideally be packaged with the full implementation of first party data (which is on the Rubicon internal queue), but I gather you have a more immediate need for keywords, so I suppose jumping the gun on this one part of first party data is ok.

On a tangent, why do we accept `setConfig({app})`?
- Apps use the SDK, not PBJS
- This 'feature' isn't documented anywhere

Searched for the history of this feature but it's pretty old -- beyond the refactoring. Anyone recall the use case? If not, I propose removing it. That wouldn't really be a breaking change because it undocumented.

AntoineJac:
@bretg , here the documentation for setConfig({app}):
http://prebid.org/faq/prebid-server-faq.html#how-can-i-use-prebid-server-in-a-mobile-app-post-bid-scenario

This is important **to not remove** it as I have several clients using it on a postbid environment! 

Regarding setConfig({site}) I am ok to wait for the new design for first party data. Thanks

bretg:
Good to know. Made a PR to update the post-bid docs. https://github.com/prebid/prebid.github.io/pull/1288

Can we close this PR then?

AntoineJac:
Yes perfect thanks @bretg

Issue #3759, opened 2019-04-16
Intent to Implement: Adding GAM Support to Prebid long-form


jaiminpanchal27:
## Type of issue
Intent to implement.
**Status:** Open for comment. 

# Table of Contents
* [Overview](#overview)
* [Proposed Design](#design)
  * [DFP Module Spec](#dfp)
  * [Mapping file](#mapping)
  * [Use Cases](#use-cases)
  * [GAM Setup](#gam)
  
# Overview <a name='overview'></a>
Prebid.js already released [long-form feature](https://github.com/prebid/Prebid.js/issues/3379) with FreeWheel as first supporting ad server. Now we are adding Google ad server support. This Spec covers the implementation details for adding long-form support to Google ad manager module in Prebid. 

# Proposed Design <a name='design'></a>
## DFP Module Spec <a name='dfp'></a>
Prebid is already doing video header bidding for google ad server using dfpAdServerVideo module. To support long-form we will expose two new methods in this module.
* pbjs.adServers.dfp.buildAdpodVideoUrl
* pbjs.adServers.dfp.getAdpodTargeting

#### pbjs.adServers.dfp.buildAdpodVideoUrl
This method combines publisher-provided parameters with Prebid.js targeting key values to build a DFP video ad tag URL that can be used by a video player.

| Params   | Scope    | Type          | Description                                                                                                                                                             |
|----------|----------|---------------|-------------------------------------------------------------------------------------------------------------------------------------------------------------------------|
| codes    | required | array[string] | Array of ad unit codes. Module will ignore all other ad units                                                                                                           |
| params   | required | object        | Query string parameters that will be used to construct the DFP video ad tag URL. Publisher-supplied values will override values set by Prebid.js. See below for fields. |
| callback | required | function      | Callback function to execute when master tag url is back                                                                                                                |

params object. More info on params object here https://support.google.com/admanager/answer/1068325?hl=en

| Params          | Scope    | Type   | Description                                              | Notes                                                                                       |
|-----------------|----------|--------|----------------------------------------------------------|---------------------------------------------------------------------------------------------|
| iu              | required | string | ad unit                                                  |                                                                                             |
| description_url | required | string | The value should describe the video playing on the page. | For long form we will add kvs for multiple bids, hence cannot take description url from bid |

Sample Call
```Javascript
pbjs.adServers.dfp.buildAdpodVideoUrl({
    codes: ['adUnitcode-1'],
    params: {
[...]

PR #3787, opened 2019-04-29
Google ad manager support for long-form video

jaiminpanchal27:
## Type of change
- [ ] Bugfix
- [x] Feature
- [x] Build related changes

## Description of change
This PR adds google ad server support to long-from video. Please refer https://github.com/prebid/Prebid.js/issues/3759 for more details

We will test this with ad server once we get access.


modules/dfpAdServerVideo.js
```diff
@@ -103,6 +107,92 @@ export default function buildDfpVideoUrl(options) {
   });
 }
 
+export function notifyTranslationModule(fn) {
+  fn.call(this, 'dfp');
+}
+
+getHook('registerAdserver').before(notifyTranslationModule);
```
jsnellbaker: Do we need to update the `registerAdserver` function in the `categoryTranslation` file to support the DFP adserver?

Further if someone is trying to load an instream video bid with DFP, this hook (with where it's currently positioned in the base file) will attempt to setup the categoryTranslation function and load the appropriate json file.  If there is no previous localStorage translation data, the module will attempt to load an `undefined` url that results in a 404 error in the console.

If we can't restrict adding this hook to only when a request for an adpod bid is made (not sure if that's doable/practical off the top of my head), then we should add contingencies to the category code to abort its look-up to prevent the above scenario.

test/spec/modules/dfpAdServerVideo_spec.js
```diff
@@ -1,12 +1,16 @@
...
+import { buildDfpVideoUrl, buildAdpodVideoUrl } from 'modules/dfpAdServerVideo';
 import { parseQS } from 'src/url';
 import adUnit from 'test/fixtures/video/adUnit';
 import * as utils from 'src/utils';
 import { config } from 'src/config';
 import { targeting } from 'src/targeting';
+import { auctionManager } from 'src/auctionManager';
+import * as adpod from 'modules/adpod';
+
+// import { getTargeting } from 'modules/common/videoAdserver';
```
jsnellbaker: remove commented code.

modules/dfpAdServerVideo.js
jaiminpanchal27: Yes, updated the function in categoryTranslation module to not initialize translation in case of dfp as we do not have default mapping file for DFP adserver.
In DFP, labels are used to manage competitive exclusion. Publisher can define any string value as label and hence it is not possible to have default mapping file for DFP. 

With the latest commit initTranslation will only execute if freewheel is adserver.

Also I see that undefined in targeting value which is coming from here https://github.com/prebid/Prebid.js/blob/master/modules/adpod.js#L216. We need return statement as we are rejecting bid here

PR #3802, opened 2019-05-06
update spec to alias older bidder code that can serve video ads

eyang414:
## Type of change

- [x] Bugfix

## Description of change

Index Exchange is able to support video ads for prebid versions 0.34.8 and older, but not for the most recent prebid versions.

Index Exchange at one point switched over to a new bidder code: 'ix'.  This caused the new versions of prebid to think that the old bidder code: 'indexExchange' is invalid.  This causes the bidder code: 'ix' to be incompatible with video-ads, while bidder code: 'indexExchange' is compatible at prebid.js version 0.34.8.  This PR just modifies the 'ixBidAdapter.js' file's `spec` object to also include `alias: 'indexExchange'`.

Previous to this change, the video bids for 'ix' would be filtered out during the pbjs.requestBids() function where 'ix' was a bidder that did not have 'video' as a `spec.supportedMediaTypes` causing `bidderEligible` to be false.  This is a bug because Index Exchange is capable of serving video ads for the older prebid versions 0.34.8 and older.

### DIFF
--- a/modules/ixBidAdapter.js
+++ b/modules/ixBidAdapter.js
@@ -1,5 +1,5 @@
 import * as utils from '../src/utils';
-import { BANNER } from '../src/mediaTypes';
+import { BANNER, VIDEO } from '../src/mediaTypes';
 import { config } from '../src/config';
 import isArray from 'core-js/library/fn/array/is-array';
 import isInteger from 'core-js/library/fn/number/is-integer';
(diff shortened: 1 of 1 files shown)


eyang414:
This is a work-around because you still need to pass-in a bid with `bidder: 'indexExchange'` rather than `bidder: 'ix'`

harpere:
Maybe I'm missing something - but Adding  `VIDEO` to `spec.supportedMediaTypes` will only affect future versions of Prebid.  It will have no affect on earlier versions, specifically Prebid 0.34.8.  So I don't understand how this fixes a problem.  0.34.8 will still not work with `ix` (based on what I read in your comment), and your current adapter doesn't support `VIDEO`, right?  So how does adding `VIDEO` to `supportedMediaTypes` help for either `ix` or `indexExchange`?

PR #3805, opened 2019-05-07
Rad 2751/specify ad units set targeting for ast

sumit116:
## Type of change

-  Bugfix

## Description of change
Added Unit tests for setTargetingForAst()

### DIFF
--- a/test/spec/unit/core/targeting_spec.js
+++ b/test/spec/unit/core/targeting_spec.js
@@ -338,21 +338,8 @@ describe('targeting tests', function () {
       sandbox.restore();
     });
 
-    it('should set single addUnit code', function() {
-      let adUnitCode = 'testdiv-abc-ad-123456-0';
-      sandbox.stub(targetingInstance, 'getAllTargeting').callsFake(function() {
-        let mockTargeting = {};
-        mockTargeting.adUnitCode = {hb_bidder: 'appnexus'};
-        return mockTargeting;
-      });
-      targetingInstance.setTargetingForAst(adUnitCode);
-      expect(targetingInstance.getAllTargeting.called).to.equal(true);
-      expect(targetingInstance.resetPresetTargetingAST.called).to.equal(true);
-      expect(window.apntag.setKeywords.called).to.equal(true);
-    });
-
-    it('should set array of addUnit codes', function() {
-      let adUnitCodes = ['testdiv1-abc-ad-123456-0', 'testdiv2-abc-ad-123456-0']
+    function mockGetAllTargeting(adUnitCodes) {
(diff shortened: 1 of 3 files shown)


test/spec/unit/core/targeting_spec.js
```diff
@@ -326,4 +326,46 @@ describe('targeting tests', function () {
...
+      expect(window.apntag.setKeywords.called).to.equal(true);
+    });
+
+    it('should set array of addUnit codes', function() {
+      let adUnitCodes = ['testdiv1-abc-ad-123456-0', 'testdiv2-abc-ad-123456-0']
+      mockGetAllTargeting(adUnitCodes);
+      targetingInstance.setTargetingForAst(adUnitCodes);
+      expect(targetingInstance.getAllTargeting.called).to.equal(true);
+      expect(targetingInstance.resetPresetTargetingAST.called).to.equal(true);
+      expect(window.apntag.setKeywords.called).to.equal(true);
```
jaiminpanchal27: To differentiate between this test and the one with single adUnitCode, you should test here that `window.apntag.setKeywords` executed twice. Also another assertion can be called with args. Check this https://github.com/sinonjs/sinon/issues/953#issuecomment-265447178

test/spec/unit/core/targeting_spec.js
```diff
@@ -326,4 +326,46 @@ describe('targeting tests', function () {
...
+      sandbox.stub(targetingInstance, 'resetPresetTargetingAST').callsFake(function() {});
+      sandbox.stub(window.apntag, 'setKeywords').callsFake(function() {});
+    });
+    afterEach(function () {
+      sandbox.restore();
+    });
+
+    function mockGetAllTargeting(adUnitCodes) {
+      adUnitCodes = Array.isArray(adUnitCodes) ? adUnitCodes : Array(adUnitCodes);
+      sandbox.stub(targetingInstance, 'getAllTargeting').callsFake(function() {
```
jaiminpanchal27: I would suggest here to use `stub.returns`. Changing the implementation of getAllTargeting with this fake function is not a good idea. So you can do something like below in each test case
```
targetingStub.returns({
   'adUnitCode': {hb_bidder:'appnexus'}
});
```

test/spec/unit/core/targeting_spec.js
sumit116: updated the tests.

test/spec/unit/core/targeting_spec.js
sumit116: Done.

sumit116:
Docs PR: https://github.com/prebid/prebid.github.io/pull/1357

PR #3808, opened 2019-05-08
Bugfix add bid parameters if not present

arneschulz1984:
## Type of change
- [x] Bugfix

## Description of change
We had the Problem that the object send to the onSetTargeting function do not include the bidder parameters. By analyzing the issue we found a bug in our implementation.


modules/orbidderBidAdapter.js
```diff
@@ -82,10 +82,9 @@ export const spec = {
     const getRefererInfo = detectReferer(window);
 
     bid.pageUrl = getRefererInfo().referer;
-    if (spec.bidParams[bid.adId]) {
-      bid.params = spec.bidParams[bid.adId];
+    if (spec.bidParams[bid.requestId] && (typeof bid.params === 'undefined')) {
```
jsnellbaker: Just to confirm - is this logic correct?

I checked through the code and the `spec.bidParams` appears to be populated during the `buildRequests` function by reading the `bid.params` field from the `bidRequest` object.

The `bid` object that's passed into the `onSetTargeting` gets a copy of the `params` from the `bidRequest` object when the `bid` object is processed through the prebid code (which is read from the setup in the adUnit object for your bidder).  

Given the above - it seems unlikely that this current if statement would return `true`.  Can you please review/confirm?

ujuettner: Hmmm, you're right - if we can rely on `bid.params` always being defined. I will discuss with @arneschulz1984, and we will update this PR accordingly. Thanks for finding!

ujuettner: @hiseke and I just tested it with the current prebid version (2.15.0). What we see (reproducible):
- Whenever `onHandler` is called via `onBidWon`, `bid.params` is defined and contains the expected data.
- Whenever `onHandler` is called via `onSetTargeting`, `bid.params` is **undefined**.

I don't know whether this is expected behavior or a prebid bug or some timing issue. But under these conditions, it seems like we have to check `bid.params` for `undefined`. Do you agree?

jsnellbaker: Would you be able to provide some test params for your adapter so I can review both scenarios?  I tried to use the ones in the md file, but I'm getting a 403 error on the request to your server.

Below is a copy of the details for reference:
URL: `https://orbidder.otto.de/bid`
Data/Payload:
```
{"pageUrl":"http://test.localhost:9999/integrationExamples/gpt/hello_world.html?pbjs_debug=true","bidId":"2ee2206a1622b3","auctionId":"7d686fc2-8309-4e6b-8f81-b170ea2d6085","transactionId":"b62c98b3-1fec-4e06-88de-bd5c600735c3","adUnitCode":"div-gpt-ad-1460505748561-0","sizes":[[300,250],[300,600]],"params":{"accountId":"someAccount","placementId":"somePlace"}}
```

ujuettner: There are some validations happening. You can use the dev endpoint `https://dev.orbidder.otto.de/bid` with a payload of `{"pageUrl":"https://dev.orbidder.otto.de/zzz/","bidId":"aaaa0000bbbb1111","auctionId":"cccccc-dddd","transactionId":"eeeeee-ffff","adUnitCode":"/111111111/orbidder_test","sizes":[[728,90]],"params":{"accountId":"orbidder-test","placementId":"center-banner"}}` and the cookie `Cookie: aditionUserId=6666666666666666666`. If you receive a 204 ("No Content"), just retry a couple of times ...

jsnellbaker: Hi @ujuettner
Do some of those validations look at the actual request headers?  I've tried to setup the test with the values you suggested, but I've just been seeing 204s come back after repeated attempts.

ujuettner: Hey @jsnellbaker , if you continue to see 204, it's most likely that the cookie is not set correctly (its name is `aditionUserId` and its value has to be known to the DSP, e.g. `6666666666666666666` is a valid value for testing). The following `curl` command just worked from my laptop and from a GCE VM:
```
curl -v https://dev.orbidder.otto.de/bid -H 'Cookie: aditionUserId=6666666666666666666' -d '{"pageUrl":"https://dev.orbidder.otto.de/zzz/","bidId":"aaaa0000bbbb1111","auctionId":"cccccc-dddd","transactionId":"eeeeee-ffff","adUnitCode":"/111111111/orbidder_test","sizes":[[728,90]],"params":{"accountId":"orbidder-test","placementId":"center-banner"}}'
```
Could you please just retry by copy&pasting the aforementioned `curl` command?

jsnellbaker: @ujuettner 
While the curl command did return the ad response, it wasn't loading through the adapter where the onBidWon and onSetTargeting functions occur which is what I hoping to look at.  That said, I tried a different browser and some other temporary edits to get the test payload/cookie you provided to successfully return an ad through the adapter.

I observed both events triggered and through some console.log statements I added in the adapter to print the values - I saw both sets of  params were present in both events.

For reference, below is a screenshot of the console log statements based on what I saw from the onBidWon event (the onSetTargeting had the same output):
Screen Shot 2019-05-21 at 10 47 35 AM

ujuettner: Hi @jsnellbaker , thanks for coming back on this. I've deployed a temporary prebid.js build with some console.log's onto dev.orbidder.otto.de, and I still find bid.params being undefined on the /targeting route:
orbidderBidAdapter_onHandler_2019-05-22_162127

As there seem to be circumstances where bid.params can be undefined, I'd happy to keep the check for undefined :-)

jsnellbaker: Alright - if you're 100% certain this logic should be like this, then we can leave it as is.

PR #3435, opened 2019-01-10
Using Prebid.js as npm dependency

snapwich:
## Type of change
- [x] Feature
- [x] Refactoring (no functional changes, no api changes)

## Description of change
This updates the code base to be useable directly from npm as a library within another application.  The README is updated with instructions for using Prebid.js with both Babel 6 and 7.  After implementing the proper `babel-loader` within the external application, Prebid.js can be used like so

```javascript
import prebid from 'prebid.js';
import 'prebid.js/modules/rubiconBidAdapter'; // imported modules will register themselves automatically with prebid
import 'prebid.js/modules/appnexusBidAdapter';
prebid.processQueue();  // required to process existing pbjs.queue blocks and setup any further pbjs.queue execution

prebid.requestBids({
  ...
})

```

This updates the local Prebid.js build to the latest version of Babel.  This required refactoring to ES6 modules for many of the files (no mix-matched commonjs and es6 imports/exports).

It also replaces the `string-replace-webpack-plugin` with a Babel Plugin for doing all the global search and replacing for things such as the `$$PREBID_GLOBAL$$`.

## Other information

ES6 modules affect some of the capabilities of sinon stubbing.  To allow stubbing for some features, in utils for example, an `internals` export was added to allow stubbing features (similar to how commonjs `exports` object was being used).  I investigated using something like `babel-plugin-rewire` to do stubbing, however it's incompatible with our `instanbul` setup and fixing it to work correctly would be an even larger refactor; however, it might be worth investigating in the future.

Going forward, any build features that we would added with Gulp or Webpack should probably be implemented as Babel Plugins to maintain compatibility for applications using Prebid.js directly from npm.  It might be worthwhile creating another repo that pulls in Prebid.js as an npm dependency and tests functionality to avoid breaking integrations.

All file imports within`src` and `modules` files were updated to use relative pathing.  This is required since any application pulling in Prebid.js will most likely not have the same webpack resolver to properly locate resources.  Any new pull-requests adding features should use relative paths for any files within `src` or `modules`.  It's still fine to use the webpack resolver for imports within `test`. 

I updated the README to mention that using this feature requires Prebid.js 1.38.0+, so it'd be nice to add this in the next release so that is accurate.

Closes #1853

PR #3820, opened 2019-05-11
CPMStar Bid Adapter

JoshuaMGoldstein:
## Type of change
- [x] New bidder adapter  

## Description of change
Added CPMStar Bid Adapter

- test parameters for validating banner bids
```
{
  bidder: 'cpmstar',
  params: {
    placementId: 81006
  }
}
```

- test parameters for validating video bids
```
{
  bidder: 'cpmstar',
  params: {
    placementId: 81007
  }
}
```

- contact email of the adapter’s maintainer: josh@cpmstar.com

- A link to a PR on the docs repo: https://github.com/prebid/prebid.github.io/pull/1309

## Other information
Contact Person:
Joshua Goldstein
General Manager, CPMStar
josh@cpmstar.com

### DIFF
--- a/modules/cpmstarBidAdapter.js
+++ b/modules/cpmstarBidAdapter.js
@@ -18,7 +18,9 @@ function getMediaType(bidRequest) {
 
 function getPlayerSize(bidRequest) {
   var playerSize = utils.deepAccess(bidRequest, 'mediaTypes.video.playerSize');
+  if (playerSize[0] != null) playerSize = playerSize[0];
   if (playerSize == null || playerSize[0] == null || playerSize[1] == null) return [640, 440];
+  return playerSize;
 }
 
 export const spec = {
--- a/modules/cpmstarBidAdapter.js
+++ b/modules/cpmstarBidAdapter.js
@@ -1,5 +1,6 @@
-import * as utils from 'src/utils';
-import { registerBidder } from 'src/adapters/bidderFactory';
+
+import * as utils from '../src/utils';
+import { registerBidder } from '../src/adapters/bidderFactory';
 import {VIDEO, BANNER} from '../src/mediaTypes';
 
 const BIDDER_CODE = 'cpmstar';
@@ -55,11 +56,6 @@ export const spec = {
     return requests;
   },
 
-  onTimeout(timeout) {
-  },
-  onBidWon(winningBid) {
-  },
-
   interpretResponse: function (serverResponse, request) {
     var bidRequest = request.bidRequest;
     var mediaType = spec.getMediaType(bidRequest);
@@ -74,11 +70,13 @@ export const spec = {
       var raw = serverResponse.body[i];
       var rawBid = raw.creatives[0];
       if (!rawBid) {
+        utils.logWarn('cpmstarBidAdapter: server response failed check');
         return;
       }
       var cpm = (parseFloat(rawBid.cpm) || 0);
 
       if (!cpm) {
+        utils.logWarn('cpmstarBidAdapter: server response failed check. Missing cpm')
         return;
       }
 
@@ -91,7 +89,6 @@ export const spec = {
         netRevenue: rawBid.netRevenue ? rawBid.netRevenue : true,
         ttl: rawBid.ttl ? rawBid.ttl : DEFAULT_TTL,
         creativeId: rawBid.creativeid || 0,
-        ad: ''
       };
 
       if (rawBid.hasOwnProperty('dealId')) {
--- a/modules/cpmstarBidAdapter.js
+++ b/modules/cpmstarBidAdapter.js
@@ -1,0 +1,118 @@
+import * as utils from 'src/utils';
+import { registerBidder } from 'src/adapters/bidderFactory';
+import {VIDEO, BANNER} from '../src/mediaTypes';
+
+const BIDDER_CODE = 'cpmstar';
+
+const ENDPOINT_DEV = '//dev.server.cpmstar.com/view.aspx';
+const ENDPOINT_STAGING = '//staging.server.cpmstar.com/view.aspx';
+const ENDPOINT_PRODUCTION = '//server.cpmstar.com/view.aspx';
+
+const DEFAULT_TTL = 300;
+const DEFAULT_CURRENCY = 'USD';
+
+function getMediaType(bidRequest) {
+  if (bidRequest == null) return BANNER;
+  return !utils.deepAccess(bidRequest, 'mediaTypes.video') ? BANNER : VIDEO;
+}
+
+function getPlayerSize(bidRequest) {
+  var playerSize = utils.deepAccess(bidRequest, 'mediaTypes.video.playerSize');
+  if (playerSize == null || playerSize[0] == null || playerSize[1] == null) return [640, 440];
+}
+
+export const spec = {
+  code: BIDDER_CODE,
+  supportedMediaTypes: [BANNER, VIDEO],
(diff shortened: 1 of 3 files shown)


modules/cpmstarBidAdapter.js
```diff
@@ -0,0 +1,121 @@
+import * as utils from 'src/utils';
+import { registerBidder } from 'src/adapters/bidderFactory';
```
jsnellbaker: In lieu of this PR #3435, any imported files within a modules file needs to use the relative style path.

Can you please update these imports accordingly?

modules/cpmstarBidAdapter.js
```diff
@@ -0,0 +1,121 @@
...
+        bidRequest: bidRequest,
+      });
+    }
+
+    return requests;
+  },
+
+  onTimeout(timeout) {
+  },
+  onBidWon(winningBid) {
```
jsnellbaker: You can remove these functions from the adapter if you're not planning on using them.

modules/cpmstarBidAdapter.js
```diff
@@ -0,0 +1,121 @@
...
+
+    if (!Array.isArray(serverResponse.body)) {
+      serverResponse.body = [serverResponse.body];
+    }
+
+    for (var i = 0; i < serverResponse.body.length; i++) {
+      var raw = serverResponse.body[i];
+      var rawBid = raw.creatives[0];
+      if (!rawBid) {
+        return;
```
jsnellbaker: As a suggestion, it may nice to add some log messages here (such as `utils.logWarn()`) to indicate something was missing from the bid response.  It would make it easier to debug by seeing the message in the browser console.

modules/cpmstarBidAdapter.js
```diff
@@ -0,0 +1,121 @@
...
+      var bidResponse = {
+        requestId: rawBid.requestid,
+        cpm: cpm,
+        width: rawBid.width || 0,
+        height: rawBid.height || 0,
+        currency: rawBid.currency ? rawBid.currency : DEFAULT_CURRENCY,
+        netRevenue: rawBid.netRevenue ? rawBid.netRevenue : true,
+        ttl: rawBid.ttl ? rawBid.ttl : DEFAULT_TTL,
+        creativeId: rawBid.creativeid || 0,
+        ad: ''
```
jsnellbaker: Given you're supporting both video and banner bids, I'd recommend to remove this default value for the `ad` property.  The `ad` is only used in the `banner` type, so it would be left equal to an empty string when you just serve back a video bid.  

Having both fields present in the bid may cause some confusion, so it'd be better to only assign the property when its needed.

test/spec/modules/cpmstarBidAdapter_spec.js
```diff
@@ -0,0 +1,176 @@
...
+      }
+
+    };
+
+    it('should produce a valid production request', function () {
+      var requests = spec.buildRequests(valid_bid_requests, bidderRequest);
+      expect(requests[0]).to.have.property('method');
+      expect(requests[0]).to.have.property('url');
+      expect(requests[0]).to.have.property('bidRequest');
+      expect(requests[0].url.startsWith('//server.cpmstar.com/view.aspx')).to.equal(true);
```
jsnellbaker: The `startsWith` property is not compatible with IE browsers - which is why the browserstack tests are failing for this PR.  Can you use an alternate assertion for this part of the tests?

I think you could use chai's `includes` instead to get the same effect (see example below).
`expect(requests[0].url).to.include('//dev.server.cpmstar.com/view.aspx');`

modules/cpmstarBidAdapter.js
JoshuaMGoldstein: We updated the imports.

modules/cpmstarBidAdapter.js
JoshuaMGoldstein: Thanks, these have been removed.

modules/cpmstarBidAdapter.js
JoshuaMGoldstein: We added a few log messages, thanks!

modules/cpmstarBidAdapter.js
JoshuaMGoldstein: This has been removed -- thanks for the tip.

test/spec/modules/cpmstarBidAdapter_spec.js
JoshuaMGoldstein: This test has been updated for IE compatibility.

JoshuaMGoldstein:
We tested the banner placement with the helloworld example, which works fine.

But the only working video example we could find (in the prebid.org docs repo, under: examples/video/jwPlayerPrebid.html) uses pbjs.adServers.dfp.buildVideoUrl.

DFP does not seem to like the vastUrl/vastXml/videoCacheKey returned by bidsBackHandler from our bidder, and DFP will not provide our vastXml at the url provided by dfp.buildVideoUrl()

However, in the above referenced example, calling invokeVideoPlayer() with our bid's vastUrl directly as follows:

invokeVideoPlayer(bids.video1.bids[0].vastUrl)

..works just fine. The vastUrl/vastXml contains a proper VAST 2.0 response from our ad server. 

Please advise. Thank you!

jsnellbaker:
@JoshuaMGoldstein 

That jwPlayerPrebid.html test page should provide a good idea on how it could be setup (I'll see if we can upload a dedicated test page in the repo for easier access).

The key would be to use the `pbjs.adServers.dfp.buildVideoUrl()` function to generate the proper ad URL that DFP would be able to process.  It grabs the necessary bid keys/data and formats that info into the query portion of the ad URL.  DFP uses these keys to help call Prebid Cache to grab the actual VAST creative which then (hopefully) gets processed by the player.

This whole process is similar with other types of ads (such as banner), where we call the equivalent `setTargetingForGPTAsync()` function in the `bidsBackHandler` to initialize some keys on a DFP ad URL, which then would call a specific Prebid creative that's sitting in DFP, which is then used to help render the actual ad.  We don't just try to directly pass the ad-markup code to DFP, instead use the necessary handlers/functions to to pass along the right bits of info.

I hope the above makes sense.  If so, do you have any test bid params for a video bid that I could to run a check?

JoshuaMGoldstein:
We added the test parameters above, yes:

{
  bidder: 'cpmstar',
  params: {
    placementId: 81007
  }
}

The prebid cache is indeed loading the vast XML properly, and the bid response contains a xmlCacheKey and the xmlUrl on the prebid cache.  However, DFP's resulting URL does not appear to mirror the VAST creative effectively in the example cited.

JoshuaMGoldstein:
Ok, we tested 3 other video adapters using their test placements and none of them worked in the jwPlayerPrebid.html example. Adapters tested:

lkqd, vuble, oneVideo.

The only adapter which seems to currently work with this is example is appnexus

jsnellbaker:
@JoshuaMGoldstein I found the likely issue; the DFP campaign that was setup for those test pages has a keyword targeting on the `hb_pb` key; it expects it to be `5.00`.

I added the following code to dynamically adjust the CPM of your test bid to the ideal value and I saw the preroll ad play through the player (the video.js one).

```
pbjs.bidderSettings = {
        cpmstar: {
          bidCpmAdjustment: function(bid) {
            return 5.00;
          }
        }
      }
```

This should work on your end as well if you want to give it a try.  From my end, everything looks good now and we can merge this PR.

PR #3824, opened 2019-05-14
adxcgBidAdapter - added pubcid

adxcgcom:
## Type of change
- [x ] Feature

## Description of change
Added userid module support


adxcgcom:
code updated with unit-tests and local gulp test runs OK.
circleci tests that fail are not related to our code and are at the moment latching between different code - probabably circleci - browserstack internal issue.

Issue #3855, opened 2019-05-22
When currency module is enabled, all bid object should have originalCurrency and originalCpm properties


pm-harshad-mane:
## Type of issue
BUG

## Description
When we enable the currency module, the currency module adds originalCurrency and originalCpm properties to the bid object. But these properties are not added when bid.currency is same as config.currency.adServerCurrency. This makes some bid objects inconsistent with other bid objects. This makes it difficult for the bid object consuming codes like bidResponse event handler or analytics handlers to consume this information. Due to this in-consistency, the consuming codes need to handle these bid objects separately and may need to set value for these properties on own with some logic.  Better if we make the bid objects consistently in the presence of the currency module by adding these missing properties even if bid.currency matches config.currency.adserverCurrency

## Steps to reproduce
Example:
PreBid Config.currency.adServerCurrency = 'SEK'
PubMatic bidder bids with currency USD
AdForm bidder bids with currency SEK

then pubmaticBid.originalCurrency = 'USD' and pubmaticBid.originalCpm = 5
but adformBid.originalCurrency and adform.originalCpm are UNDEFINED

Refer: https://github.com/prebid/Prebid.js/blob/master/modules/currency.js#L192

pm-harshad-mane:
we can close this issue as we have merged the changes to master.

PR #3856, opened 2019-05-23
Currency Module: always adding originalCpm and originalCurrency to bid object

pm-harshad-mane:
## Type of change

## Description of change
#3855 
Always add originalCpm and originalCurrency to bid object when the currency module is enabled.

### DIFF
--- a/modules/currency.js
+++ b/modules/currency.js
@@ -188,6 +188,9 @@ export function addBidResponseHook(fn, adUnitCode, bid) {
     return (parseFloat(cpm) * getCurrencyConversion(fromCurrency, toCurrency)).toFixed(3);
   };
 
+  bid.originalCpm = bid.cpm;
+  bid.originalCurrency = bid.currency;
+
   // execute immediately if the bid is already in the desired currency
   if (bid.currency === adServerCurrency) {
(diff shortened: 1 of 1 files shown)


pm-harshad-mane:
Hello @jsnellbaker ,
Could you please review?

pm-harshad-mane:
Hello @jsnellbaker ,
A gentle reminder :)
Could you please review the change?

jaiminpanchal27:
@pm-harshad-mane I am not 100% sure here but undefined shouldn't be problem here. If it is undefined it clearly means that bidder responded in same currency and there is no need to store same data in duplicate property. 

I guess analytics adapters or listeners can add a check for these properties and can easily figure this out.

pm-harshad-mane:
@jaiminpanchal27 analytics adapters and listeners can add this un-documented check but I think it is better if currency module always sets these two fields so that the listeners and analytics adapters will not need this special handling and it will also make the bid-object have consistent fields.

@idettman has approved the change does that mean we can merge the change in master?

pm-harshad-mane:
Thank you @idettman , @jaiminpanchal27 , @jsnellbaker :)

PR #3865, opened 2019-05-29
feat(unruly-bid-adapter): use bidResponse siteId when configuring the renderer

paprikka:
## Type of change

- [x] Bugfix

## Description of change

Ensure that the Unruly Bid Adapter can be used on sites where the Unruly Universal Tag has not been configured separately. This is a backwards-compatible change and it doesn't require any actions on Publisher's side.

### DIFF
--- a/modules/unrulyBidAdapter.js
+++ b/modules/unrulyBidAdapter.js
@@ -6,7 +6,7 @@ import { VIDEO } from '../src/mediaTypes'
 function configureUniversalTag (exchangeRenderer) {
   parent.window.unruly = parent.window.unruly || {};
   parent.window.unruly['native'] = parent.window.unruly['native'] || {};
-  parent.window.unruly['native'].siteId = parent.window.unruly['native'].siteId || exchangeRenderer.siteId;
+  parent.window.unruly['native'].siteId = parent.window.unruly['native'].siteId || exchangeRenderer.config.siteId;
   parent.window.unruly['native'].supplyMode = 'prebid';
 }
 
--- a/modules/unrulyBidAdapter.js
+++ b/modules/unrulyBidAdapter.js
@@ -4,6 +4,9 @@ import { registerBidder } from '../src/adapters/bidderFactory'
 import { VIDEO } from '../src/mediaTypes'
 
 function configureUniversalTag (exchangeRenderer) {
+  if (!exchangeRenderer.config) throw new Error('UnrulyBidAdapter: Missing renderer config.')
+  if (!exchangeRenderer.config.siteId) throw new Error('UnrulyBidAdapter: Missing renderer siteId.')
+
   parent.window.unruly = parent.window.unruly || {};
   parent.window.unruly['native'] = parent.window.unruly['native'] || {};
   parent.window.unruly['native'].siteId = parent.window.unruly['native'].siteId || exchangeRenderer.config.siteId;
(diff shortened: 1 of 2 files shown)


modules/unrulyBidAdapter.js
```diff
@@ -6,7 +6,7 @@ import { VIDEO } from '../src/mediaTypes'
 function configureUniversalTag (exchangeRenderer) {
   parent.window.unruly = parent.window.unruly || {};
   parent.window.unruly['native'] = parent.window.unruly['native'] || {};
-  parent.window.unruly['native'].siteId = parent.window.unruly['native'].siteId || exchangeRenderer.siteId;
+  parent.window.unruly['native'].siteId = parent.window.unruly['native'].siteId || exchangeRenderer.config.siteId;
```
robertrmartinez: can you please add a check to verify `exchangeRenderer.config` is defined before trying to access `siteId`?

I notice this `exchangeRenderer` object comes from your serverBid, and you may guarantee that this is always defined, but we would prefer to be defensive here please !

robertrmartinez: Want to void `Error: Cannot read property 'siteId' of undefined`

paprikka: Roger that—throwing meaningful exceptions unless `siteId` is available.

robertrmartinez: @paprikka Thanks,

quick question, where exactly does this `exchangeRenderer.config` come from?

The Unruly server request?

Is it set by the publisher somewhere?

paprikka: `exchangeRenderer.config` comes from the `bidResponse`

robertrmartinez: Right, okay.

So the suggestion here is to not throw an error, but instead use the log utils to log the error and return the expected value back to prebid core.

for example, Prebid Core expects the bids back as an array. So instead of throwing the error,

Can we update to log the error, and simply return back an empty array to Prebid core?

Thanks!

paprikka: Sure, done. Thanks!

paprikka:
Regarding the CI issue: I've re-run the tests and everything seems fine on our side, can't access the CircleCI logs though—can you trigger another run?

modules/unrulyBidAdapter.js
```diff
@@ -4,9 +4,12 @@ import { registerBidder } from '../src/adapters/bidderFactory'
 import { VIDEO } from '../src/mediaTypes'
 
 function configureUniversalTag (exchangeRenderer) {
+  if (!exchangeRenderer.config) throw new Error('UnrulyBidAdapter: Missing renderer config.')
+  if (!exchangeRenderer.config.siteId) throw new Error('UnrulyBidAdapter: Missing renderer siteId.')
```
robertrmartinez: @paprikka 

Are these checks now irrelevant?

Due to the filter, it seems a bid would never come into this function unless it already has the config and siteID.

Obviously this is not a show-stopper, but if it is irrelevant, please remove it.

If you do not get to it before the release scheduled for tomorrow, I will make sure to still merge it and we can update in a later PR.

paprikka: Ah, yes—these are not required any more. 
I'm afraid I don't time to deal with this today—will submit another PR later this week if that's ok?

robertrmartinez: Yes, please do not forget to update it when you have a chance!

PR #3867, opened 2019-05-29
Digitrust submodule

goosemanjack:
This adds DigiTrust support to the updated userId system.

## Type of change
- [X ] Feature

## Description of change
This is an update to add DigiTrust Id support to the new userId module for Prebid. This change eliminates the loader and instead uses the "attachIdSystem" method to link in DigiTrust Id submodule to the userId system.

## Other information
Previous changes to documentation remains valid.

### DIFF
--- a/integrationExamples/gpt/digitrust_Full.html
+++ b/integrationExamples/gpt/digitrust_Full.html
@@ -1,13 +1,12 @@
 <html>
 <head>
-    <title>Simple DigiTrust Prebid - No Framework</title>
-
+    <title>Full DigiTrust Prebid Sample</title>
     <script>
-        (function (window, document) {
+        (function(window, document) {
             if (!window.__cmp) {
-                window.__cmp = (function () {
+                window.__cmp = (function() {
                     var listen = window.attachEvent || window.addEventListener;
-                    listen('message', function (event) {
+                    listen('message', function(event) {
                         window.__cmp.receiveMessage(event);
                     }, false);
 
@@ -26,7 +25,7 @@
                     addLocatorFrame();
 
                     var commandQueue = [];
-                    var cmp = function (command, parameter, callback) {
+                    var cmp = function(command, parameter, callback) {
                         if (command === 'ping') {
                             if (callback) {
                                 callback({
@@ -43,7 +42,7 @@
                         }
                     };
                     cmp.commandQueue = commandQueue;
-                    cmp.receiveMessage = function (event) {
+                    cmp.receiveMessage = function(event) {
                         var data = event && event.data && event.data.__cmpCall;
                         if (data) {
                             commandQueue.push({
@@ -85,7 +84,7 @@
         var adUnits = [
             {
                 code: 'test-div',
-                sizes: [[300, 250], [300, 600], [728, 90]],
+                sizes: [[300,250],[300,600],[728,90]],
                 bids: [
                     {
                         bidder: 'rubicon',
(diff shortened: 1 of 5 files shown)


modules/digiTrustIdSystem.js
```diff
@@ -0,0 +1,307 @@
...
+ * Tests for presence of the DigiTrust object
+ * */
+function isPresent() {
+  return (window.DigiTrust != null);
+}
+
+var noop = function () {
+};
+
+const MAX_RETRIES = 4;
```
samtingleff: suggest changing this value to 1 :)

modules/digiTrustIdSystem.md
```diff
@@ -0,0 +1,74 @@
...
+-----
+The DigiTrust Id integration for Prebid may be used with or without the full
+DigiTrust library. This is an optional module that must be used in conjunction
+with the userId module.
+
+The full DigiTrust integration requires the DigiTrust package be included.
+DigiTrust may be initialized in the standard manner or through the params
+object. 
+
+You can use npm or reference the script on the cdn.
```
samtingleff: suggest removing this stuff about referencing a script on the CDN, or using npm... let's make it as easy as possible, and no easier.

modules/digiTrustIdSystem.md
```diff
@@ -0,0 +1,74 @@
...
+                usersync: {
+                    userIds: [{
+						name: "digitrust",
+						params: {
+							init: {
+								member: 'example_member_id',
+								site: 'example_site_id'
+							},
+							callback: function (digiTrustResult) {
+								if (digiTrustResult.success) {
```
samtingleff: suggest removing the callback logging. people copy and paste this into their page and then ask me "why is it logging all this stuff". examples should probably not include console.log at all.

integrationExamples/gpt/digitrust_Full.html
```diff
@@ -0,0 +1,222 @@
+<html>
+<head>
+    <title>Full DigiTrust Prebid Sample</title>
+    <script>
+        (function(window, document) {
+            if (!window.__cmp) {
```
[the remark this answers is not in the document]
goosemanjack: I'm not sure about all of this. It was copy/pasted from the other userId example. If we can remove it we should.

integrationExamples/gpt/digitrust_Full.html
```diff
@@ -0,0 +1,222 @@
...
+
+        var adUnits = [
+            {
+                code: 'test-div',
+                sizes: [[300,250],[300,600],[728,90]],
+                bids: [
+                    {
+                        bidder: 'rubicon',
+                        params: {
+                            accountId: '1001',
```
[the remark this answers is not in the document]
goosemanjack: Ditto here. Copy paste from the other userId example.

modules/digiTrustIdSystem.js
goosemanjack: We could go to two. There are two code paths that could possibly run prior to DigiTrust being loaded. It may be an unnecessary guard, but only boils down to 300 milliseconds.

modules/digiTrustIdSystem.md
goosemanjack: OK

modules/digiTrustIdSystem.md
goosemanjack: OK

modules/digiTrustIdSystem.js
```diff
@@ -0,0 +1,307 @@
...
+      return { 'digitrustid': idData };
+    } catch (e) {
+      utils.logError('DigiTrust ID submodule decode error');
+    }
+  },
+  getId: getDigiTrustId,
+  _testInit: surfaceTestHook
+};
+
+attachIdSystem(digiTrustIdSubmodule);
```
idettman: perfect place to call attachIdSystem, nice!

modules/digiTrustIdSystem.md
goosemanjack: Removed and added in more complete instructions from doc site.

modules/digiTrustIdSystem.js
goosemanjack: Changed to "2"

PR #3874, opened 2019-05-31
Use actual global object name in log message

Rooke:
## Type of change

- [x] Bugfix

## Description of change
The global Prebid.js object name can be something other than 'pbjs'. Update this log message to reference the specified prebid global object name.


jsnellbaker:
Hi @Rooke 

If we're going to update this log message, there is one other that could be similarly updated; link to area is below:
- https://github.com/prebid/Prebid.js/blob/master/src/video.js#L51

Could you include this change?

Rooke:
@jsnellbaker Thanks for finding that.

PR #3889, opened 2019-06-06
Adding a method pbjs.getUserIds to share userIds in Prebid to external codes

pm-harshad-mane:
## Type of change
- [X] Feature

## Description of change
Adding a method pbjs.getUserIds to share userIds in Prebid to external codes.
- Need an API to share the UserIds retrieved by Prebid with external codes
- Use cases: UserIds can be passed to Wrappers like A9/EB
- Added pbjs.getUserIds method
- Moved the code that initializes all sub-modules and calls callbacks into a separate function
- Moved code to combine submodule IDs in an object into a separate function


pm-harshad-mane:
Hello @idettman ,
Could you please review the changes?

pm-harshad-mane:
Example:
pbjs.getUserIds();
// will return following object when tdid is available, object will contain more ids as per availability
{tdid: "cdbe1fea-44a9-4f93-b6a7-11a64eb9f8dc"}

bretg:
@pm-harshad-mane - we're going to need a docs update for new external API call. Care to take a cut at a PR for http://prebid.org/dev-docs/publisher-api-reference.html ?

pm-harshad-mane:
Sure @bretg , I will add the documentation PR as well.
I might some help from you, will ping you on Slack.

modules/userId.js
```diff
@@ -414,6 +443,10 @@ export function init(config) {
       updateSubmodules();
     }
   })
+
+  // exposing getUserIds function in global-name-space so that userIds stored in Prebid can be used by external codes.
+  let theGlobal = getGlobal();
```
idettman: @jaiminpanchal27 Can you review this use of getGlobal?

jaiminpanchal27: @idettman this looks good. Use case here is valid and user id being a core module we can allow usage of global. 
Another approach to not use global here was to import getUserIds function in prebid.js and use their but then it will always import user id module.

pm-harshad-mane: Thank you @jaiminpanchal27 👍

pm-harshad-mane:
Hello @jaiminpanchal27 ,
Can you please review https://github.com/prebid/prebid.github.io/pull/1361 ?

pm-harshad-mane:
Hello @jaiminpanchal27 and @idettman ,
Documentation review is done by @bretg 
Can you please review the code?

modules/userId.js
```diff
@@ -253,14 +260,36 @@ export function requestBidsHook(fn, reqBidsConfigObj) {
...
-
   // calling fn allows prebid to continue processing
   return fn.call(this, reqBidsConfigObj);
 }
 
+/**
+ * This function will be exposed in global-name-space so that userIds stored by Prebid UserId module can be used by external codes as well.
+ * Simple use case will be passing these UserIds to A9 wrapper solution
+ */
+function getUserIds() {
```
jaiminpanchal27: @pm-harshad-mane Should this function not have a callback function as argument which will be called when user id's are ready. Id systems make async calls to get user ids. This function may return empty object or undefined.

pm-harshad-mane: Hello @jaiminpanchal27 ,
my expectation from the API is to return whatever data is available at the moment.
It will be a good feature to have a call-back on getting all ids, but for now, it is not a requirement. Also, as userId module is caching the data, maybe on the second iteration of execution the new API will return all ids so it is sufficient for now.

pm-harshad-mane:
Hello @idettman ,
We are waiting for your review comment, could you please check?

pm-harshad-mane:
Hello @idettman ,
A gentle reminder to review this PR.

modules/userId/index.js
```diff
@@ -415,6 +444,10 @@ export function init(config) {
       updateSubmodules();
     }
   })
+
+  // exposing getUserIds function in global-name-space so that userIds stored in Prebid can be used by external codes.
+  let theGlobal = getGlobal();
```
idettman: Avoid creating a reference for the `getGlobal()` value since it's only used once. 
`(getGlobal()).getUserIds = getUserIds`

modules/userId/index.js
```diff
@@ -197,19 +197,32 @@ function processSubmoduleCallbacks(submodules) {
...
 /**
- * @param {AdUnit[]} adUnits
+ * This function will create a combined object for all subModule Ids
  * @param {SubmoduleContainer[]} submodules
  */
-function addIdDataToAdUnitBids(adUnits, submodules) {
-  if ([adUnits, submodules].some(i => !Array.isArray(i) || !i.length)) {
-    return;
+function getCombinedSubmoduleIds(submodules) {
+  if ([submodules].some(i => !Array.isArray(i) || !i.length)) {
```
idettman: Since only submodules is being iterated now, change to:
`if (!Array.isArray(submodules) || !submodules.length) {`

modules/userId/index.js
pm-harshad-mane: makes sense :)

modules/userId/index.js
pm-harshad-mane: done.

pm-harshad-mane:
Hello @idettman ,
Can you please check again?

pm-harshad-mane:
Thank you @idettman and @jaiminpanchal27 !!

Issue #3888, opened 2019-06-06
SendAllBids Bid Cap


r-schweitzer:
## Type of issue
Feature

## Description 
This the same issue as described in #3800 

In certain scenarios, Prebid may generate a large number of targeting keys after an auction has completed. An issue could arise when Prebid needs to send all these keys (and their values which can vary in length) to GAM or any other ad server, as this data is typically sent over as query-data in the ad server's request URL and some systems may have limits to the query-data's size.

In such scenarios, commonly what happens is that the query-data is truncated at the designated point and the ad server processes what keys came through and acts accordingly with that data-set. In Prebid's case, this can lead to issues with the Prebid ad being properly rendered (such as if some data points for a native bid were missing) or delivery issues if the Prebid line-item/campaign was targeting off of certain keys that were dropped.

### Proposed Solution
Prebid.js will offer a new config setting that will cap the number of bids made per ad unit when sendAllBids is enabled. This limiter setting will take in a number which will represents the bidCap (defaulted to zero). When bids are capped at zero sendAllBids will function normally. When the bidCap is set to 2 for example, only the targeting for the top 2 bids will be sent to the ad server for each adunit.

An example of the setting:

`pbjs.setConfig({
    bidCap: 2
  })`

pm-harshad-mane:
Hello @r-schweitzer ,
Can you please share a test-page where with this scenario?

@mkendall07 @jsnellbaker I think the issue is due to large values stored in native assets key-value pairs as other key-value pairs do not have large values. Can we avoid sending native assets in key-value pairs and share an API which will return requested asset from a specified creative(using bid-id)? this will help us avoid sending native-assets related key-value pairs.

r-schweitzer:
@pm-harshad-mane I believe in CafeMedia's case it has more to do with the number of bidders we use and number of ad slots we can have on a page when sendAllBids is enabled with SRA. I do not have a test page where this is occurring currently but I ran a test locally on the following url:

https://www.comicsands.com/chick-fil-a-rodney-bullard-2637626462.html?utm_content=inf_10_3759_2&utm_source=facebook&utm_campaign=GTAK&tse_id=INF_336771307b5311e9b433ab68d22425b2

**This was one of the requests to GAM:**
[...]

PR #3906, opened 2019-06-13
Feature: adUnitBidLimit 

r-schweitzer:
## Type of change

- [ ] Bugfix
- [x] Feature
- [ ] New bidder adapter   
- [ ] Build related changes
- [ ] CI related changes
- [ ] Other

## Description of change
https://github.com/prebid/Prebid.js/issues/3888

Prebid.js will offer a new config setting that will cap the number of bids made per ad unit when sendAllBids is enabled. This limiter setting will be called adUnitBidLimit which will be defaulted to zero which will allow sendAllBids to function normally. When the adUnitBidLimit is set to anything above 0, e.g. 2, only the targeting for the top 2 bids will be sent to the ad server for each adunit based on CPM.

An example of the setting:

pbjs.setConfig({ adUnitBidLimit: 2 })


src/targeting.js
```diff
@@ -31,16 +31,25 @@ export let filters = {
...
   let buckets = groupBy(bidsReceived, 'adUnitCode');
   // filter top bid for each bucket by bidder
   Object.keys(buckets).forEach(bucketKey => {
+    let bucketBids = [];
     let bidsByBidder = groupBy(buckets[bucketKey], 'bidderCode');
-    Object.keys(bidsByBidder).forEach(key => bids.push(bidsByBidder[key].reduce(highestCpmCallback)));
+    Object.keys(bidsByBidder).forEach(key => bucketBids.push(bidsByBidder[key].reduce(highestCpmCallback)));
+    // if adUnitBidLimit is set, pass top N number bids
+    if (adUnitBidLimit > 0) {
+      bucketBids.sort((a, b) => b.cpm - a.cpm);
```
bretg: I'd like to see the bid list randomized before sorting so that `aardvark` doesn't always win ties.

bretg:
@mkendall07 - would like you to take a look at this one:

1) this is another way of handling how many KVPs are sent to the ad server. It's arguably superior because it truncates cleanly instead of having a partial bid. Are the methods complementary, or would this obsolete the very recent `auctionKeyMaxChars`?

2) not sure I like the proposed setConfig of `adUnitBidLimit` -- maybe `sendAllBidsLimit`?

bretg:
Also - it's arguable -- at least worth discussing -- that deals should be given priority over bids with higher priority.

I would propose something like:

- always take the highest bid as the first block of targeting variables
- then sort on bids that contain a dealId and return the highest value deals in order
- then sort the remaining bids and them if there's room left

bretg:
@r-schweitzer -- we're proposing the following:

1) push bid prioritization (including deals) to a different issue
2) generalize the configuration so we can work out additional controls in the (near future)

proposed config:
```
sendBidsControl: {
    bidLimit: 2
}
```

bretg:
Ryan and I discussed in slack -- I think randomizing is the best way to go because it will resolve any question about ties and what order the bids are in now.

Apparently the list isn't in alpha-order, but it's not clear what order they're in. If they're in "first responded" order, that could be a defendable story. But if they're in "last responded" order, that would be bad. And in general, depending on the order of that list not changing might be risky.

bretg:
Ryan found that the bid array order is based on timestamp when the bid returned.

My preference is that we still go ahead and randomize the list before picking the Top N.

1) eliminates the dependency on the order of the bids array
2) avoids any benefit from bidder ordering - as long as the response is in by the timeout, ties are random

@mkendall07 - are you good with this?

r-schweitzer:
I just wanted to bring this back to everyone's attention. Should we be randomizing the bid array or is it fair that the bidder who responds the quickest with the higher CPM be listed first?

bretg:
While I prefer randomizing from an algorithm perspective, I can accept the response-based ordering as long as we put a comment in somewhere to that effect. We don't really need to weigh down the browser with a sort for this edge case.

@mkendall07?

mkendall07:
I'd prefer to keep the order as it reflects first in, first out. I'm good with keeping the scope small here and dealing with deal ids separately.

mkendall07:
Just needs a docs PR

bretg:
docs pr at https://github.com/prebid/prebid.github.io/pull/1516

src/targeting.js
bretg: I've been told that the order of the list is chronological by time of bid response. This is ok.

PR #3910, opened 2019-06-14
Add Outstream Renderer for Yieldlab Adapter

mirkorean:
## Type of change

- [x] Feature

## Description of change

- Adds a default outstream renderer for publishers without their own

### DIFF
--- a/modules/yieldlabBidAdapter.js
+++ b/modules/yieldlabBidAdapter.js
@@ -146,7 +146,8 @@ function isOutstream (format) {
  * @returns {Array}
  */
 function getPlayerSize (format) {
-  return utils.deepAccess(format, 'mediaTypes.video.playerSize')
+  let playerSize = utils.deepAccess(format, 'mediaTypes.video.playerSize')
+  return (playerSize && utils.isArray(playerSize[0])) ? playerSize[0] : playerSize
 }
 
 /**
--- a/modules/yieldlabBidAdapter.js
+++ b/modules/yieldlabBidAdapter.js
@@ -2,11 +2,13 @@ import * as utils from '../src/utils'
 import { registerBidder } from '../src/adapters/bidderFactory'
 import find from 'core-js/library/fn/array/find'
 import { VIDEO, BANNER } from '../src/mediaTypes'
+import { Renderer } from 'src/Renderer'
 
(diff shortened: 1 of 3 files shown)


modules/yieldlabBidAdapter.js
```diff
@@ -93,8 +95,23 @@ export const spec = {
         }
 
         if (isVideo(bidRequest)) {
+          const playersize = getPlayerSize(bidRequest)
+          if (playersize) {
+            bidResponse.width = playersize[0]
+            bidResponse.height = playersize[1]
+          }
```
robertrmartinez: Hello @mirkorean 

Looks good, only a small issue.

`playersize` will come as an array of size arrays here if it comes. Prebid translates it to this. Here are some possible settings an Ad Unit can have and how your getPlayerSize function will return:

No `playerSize` defined by user:
```
sizes: [[640, 480]],
       mediaTypes: {
           video: {
               context: 'outstream'
           }
       }
```
`getPlayerSize(bidRequest)` will return `undefined` here.

Which you check and would not set the width and height which would default to you `customsize ` thing above

User sets playerSize to a single size array
```
mediaTypes: {
           video: {
               playerSize: [640, 480],
               context: 'outstream'
           }
       },
```

Prebid takes this input size and translates it to an array of size arrays before sending to the adapters as seen here
image

`getPlayerSize(bidRequest)` will return `[ [ 640, 480 ] ]` here.

Which then causes your code to set:
`bidResponse.width` = `[ 640, 480 ]` 
`bidResponse. height ` = `undefined` 

This will also happen if user sets 
```
mediaTypes: {
           video: {
               playerSize: [[640, 480]],
               context: 'outstream'
           }
       },
```


So the recommendation here is to add some logic to grab the sizes width and height correctly.

```
function getPlayerSize (format) {
  return utils.deepAccess(format, 'mediaTypes.video.playerSize') ? mediaTypes.video.playerSize[0] : undefined;
}
```

Maybe?

Up to you exactly how to do this (I did not even really test or look at the example I gave.)

mirkorean: Hi @robertrmartinez 

Thanks for your review, much appreciated. I commited a fix for this.

Issue #3908, opened 2019-06-13
userID and adpod modules being included in prebid-core


snapwich:
## Type of issue
bug

## Description
The userId and adpod modules are being included in prebid-core:
<img width="433" alt="Screen Shot 2019-06-13 at 5 15 35 PM" src="https://user-images.githubusercontent.com/551085/59473260-e5e96a80-8dfe-11e9-8099-d270fdc05d93.png">

This is happening because the modules are being referenced from other modules directly:
https://github.com/prebid/Prebid.js/blob/master/modules/freeWheelAdserverVideo.js#L10
https://github.com/prebid/Prebid.js/blob/master/modules/digiTrustIdSystem.js#L15

Modules should not import other modules, doing so breaks the convention we've established where each module is a separate webpack entry point and module dependencies are specified at build time with `--modules`.

If we want to revisit that convention and [how it's implemented](https://github.com/prebid/Prebid.js/blob/master/webpack.conf.js#L50) we can, but as it is  this causes issues with the build system thinking the modules are now "shared code" and belong where shared code goes: in prebid-core and are growing all prebid.js builds by 50kb whether they use these modules or not.

@jaiminpanchal27 @idettman @bretg @mkendall07

jaiminpanchal27:
I tried one approach to solve this in one of my PR here https://github.com/prebid/Prebid.js/pull/3787/files#diff-583ded51e69735951bd128c9066eca2fR31

I added a common folder inside modules which will be a place to keep all shared code for modules. Intent here is that for any publisher they should only specify what id system they want to use in --modules. userId module is the base of it and should be automatically imported by Prebid build system.

Let me know what do you think or if you have any other way to solve this problem

goosemanjack:
The userId system was designed with each IdSystem operating as a submodule of the userId module. In order to avoid build breaks where an IdSystem submodule is included in the build, but the userId is not explicitly included the userId gets automatically pulled in. If you don't want this included you should be able to exclude the digiTrustIdSystem to remove this dependency.  So to be clear, this is expected behavior.

snapwich:
> you should be able to exclude the digiTrustIdSystem to remove this dependency. So to be clear, this is expected behavior.

I'm not sure I follow. If the modules ends up in prebid-core (like they are doing) that means they are included in all prebid.js bundles whether the digiTrustIdSystem system is included or not. Sounds unexpected considering you just described a different expected behavior.

[...]

PR #3924, opened 2019-06-18
Submodule system using hooks

snapwich:
## Type of change
- [x] Bugfix

## Description of change
Only include src dir and few node_modules inside of prebid-core bundle.

## Other information
related to #3908


snapwich:
This partially fixes the problem by preventing modules from being bundled in prebid-core.  However it doesn't fix the submodule problem, and I'm still not sure exactly how to solve that problem.  With this merged modules won't end up in prebid-core, but submodules will still each have their own copy of the parent module, which is not ideal.

snapwich:
Update the prebid-core hooks module with a `module` and `submodule` helper method which permits an install function for the parent module and the ability to pass data to the installer from the child module. This should allow for more sharing of functionality between child and parent modules without requiring either as static imports.

Also added the `./modules/.submodules.json` to make the gulp bundler aware of these relationships.  When using `gulp build --modules=digiTrustIdSystem` for example it will check the `.submodules.json` and see that digiTrust is a child of the `userId` module, so it will automatically include the parent module as well. Since they're using async hooks that are queued behind the scenes, load order of the modules doesn't matter.

idettman:
Changed to on hold, since the ID5 User ID submodule has been merged, so I'm going to update this PR including changes to the ID5 submodule
'

snapwich:
it might be worthwhile including the required changes to freewheel and adpod in this pull-request as well so we aren't left with a partial fix in master for module issues: @jsnellbaker @jaiminpanchal27

jaiminpanchal27:
@snapwich I am starting that work today. I will update you as soon as it is done. Can I push to this branch or we create one new branch and create our PR's against that branch ?

snapwich:
@jaiminpanchal27 you should be able to push to this branch, but a PR against this branch works as well.  Whatever is easier for you guys!

snapwich:
@idettman don't forget you need to update `.submodules.json` with the new sub-module you added.

idettman:
Working on an update to convert to prebidServer style module structure to allow imported sub modules

idettman:
removed myself as a reviewer, since I've contributed a significant amount of code to this branch.

jaiminpanchal27:
I have pushed adpod related changes. @jsnellbaker Please have a look and test 
Added on hold label so we do not merge this without testing all prebid submodules.

modules/.submodules.json
```diff
@@ -0,0 +1,12 @@
+{
+  "userId": [
+    "digiTrustIdSystem",
+    "pubCommonIdSystem",
+    "unifiedIdSystem",
+    "id5IdSystem",
+    "unifiedIdSystem"
```
snapwich: I think someone handled a merge improperly here.  Duplicate and removed modules are showing up now...

Might want to check to make sure no other changes were accidentally reverted as well.

idettman: Yeah this is supposed to be:
```
{
  "userId": [
    "digiTrustIdSystem",
    "id5IdSystem"
],
  "adpod": [
    "freeWheelAdserverVideo"
  ]
}
```

PR #3926, opened 2019-06-18
Rhythmone Adapter - Remove usersync,  devicetype, and bad banners

rhythmonebhaines:
## Type of change
- [x] Bugfix
- [x] Other: Feature removal

## Description of change
* Removes the usersync pixel. We don't need it.
* Removes the devicetype property. We don't use it.
* Does not send incorrectly specified banner imps.

bhaines@rhythmone.com


rhythmonebhaines:
Extra line...

Issue #3883, opened 2019-06-04
User id module remove storage requirements


jaiminpanchal27:
## Type of issue
Feature request

## Description
1. Remove storage requirements from User id module. I have an id system which cannot store user id in local storage or cookie for more than 1 hour. It needs to hit the endpoint once every hour and get the latest user id.

pm-harshad-mane:
Hello @jaiminpanchal27 ,
if the end-point needs to be hit every hour and not on every impression then the id-provider will need to store the id for at least an hour, then I think we need to store it somewhere (cookie, local-storage).
Am I missing something?

jaiminpanchal27:
Merged in #3930

PR #3930, opened 2019-06-19
Criteo real time user sync

jaiminpanchal27:
## Type of change
- [x] Feature

## Description of change
Bidders who work with Criteo can increase their demand by sending user id to Criteo. This PR adds criteo rtus id system. Bidders need to configure their own client identifier in order to get the user id. Use following snippet on page. 

```
pbjs.setConfig({
    usersync: {
        userIds: [{
            name: "criteortus",
            params: {
              clientIdentifier: {
                "appnexus": 30
              }
            }
        }]
    }
});
```

This PR also removes storage restrictions in user id module. #3883 
User id returned by criteo has max age of 1 hour and user id module stores the id in cookie or local storage for days. Workaround to do this was module can handle its own storage without hurting the auction timing.

TODO: Docs PR coming soon


modules/criteortusIdSystem.js
```diff
@@ -0,0 +1,105 @@
...
+          // criteo rtus user id expires in 1 hour
+          const expiresStr = (new Date(Date.now() + (60 * 60 * 1000))).toUTCString();
+          utils.setCookie(key, JSON.stringify(userIds), expiresStr);
+          callback(userIds);
+        }
+
+        const onResponse = utils.delayExecution(afterAllResponses, bidders.length);
+
+        bidders.forEach((bidder) => {
+          let url = `https://gum.criteo.com/sync?c=${configParams.clientIdentifier[bidder]}&r=3`;
```
mkendall07: this isn't going to scale out so well. Is that the only way Criteo supports (1 ID at a time?)

jaiminpanchal27: Yes, each bidder will have it's own client identifier. As per information we have now, this seems only option for now.  There will be changes once we know more about this.

modules/userId.js
```diff
@@ -298,6 +300,13 @@ function initSubmodules(submodules, consentData) {
     } else if (submodule.config.value) {
       // cache decoded value (this is copied to every adUnit bid)
       submodule.idObj = submodule.config.value;
+    } else {
+      let result = submodule.submodule.getId(submodule.config.params, consentData);
```
idettman: If `result` is not reassigned, should be a `const`

jaiminpanchal27: Done

jaiminpanchal27:
@idettman Updated to use submodules pattern and updated let to const

bretg:
Docs PR https://github.com/prebid/prebid.github.io/pull/1381

bretg:
@jaiminpanchal27 - noticed that several user ID modules aren't being supported by prebid server bid adapter, with criteortus being one of them.

Who were you working with at criteo on this module? We need to confirm with them what source value they'd like for user.ext.eids.source. e.g.
```
          {
                "source": "criteortus",    // or would they prefer criteo.com?
                "uids": [{
                    "id": "11111111"
                }]
            }
```

PR #3928, opened 2019-06-19
Update emoteevBidAdapter.js

piotr-yuxuan:
## Type of change

- [x] Feature
- [x] Refactoring (no functional changes, no api changes)
- [x] Affects user-facing API ([doc pull request](https://github.com/prebid/prebid.github.io/pull/1365))

## Description of change

Retrieve various data from the adapter configuration, and ensure they are properly formatted:
- GDPR vendor consent
- Metadata: intended to be used as an ID for external partners
- Context: describe the context of the current ad unit

Test coverage for emoteevBidAdapter.js:
- Statements 94.83%
- Branches 98.39%
- Functions 100%
- Lines 94.34%

PR #3932, opened 2019-06-19
remove pubCommonId import from emoteevBidAdapter

snapwich:
## Type of change
- [x] Bugfix

## Description of change
Modules should not directly import other modules as it results in accidental code inclusion or possibly duplicate code inclusion (if that module was already included). If it does in-fact depend on the other module being present you might be interested in work on submodules here: #3924

/cc @piotr-yuxuan


snapwich:
Closed as this was cherry-picked into #3928

PR #3938, opened 2019-06-21
Fix #3894 - update sizeConfig logic around multiformat bids

### LINKED ISSUE #3894: sizeConfig and multiformat
## Type of issue
Bug or Feature Request

## Description
**note - made some updates based on further review and testing out potential fixes**

The sizeConfig logic is currently configured to allow bids of either a `video` or a `native` mediaType effectively a free-pass through its checks/filters.  This makes sense to a degree, as these bid types don't generally rely on size dimensions being a key part the delivery decision and rendering process.  

However, this logic has allowed certain issues to be possible that should be addressed in some manner.  When a multi-format bid (eg banner + native) goes through the sizeConfig, the following could/does occur:
- if the sizes of the `banner` were completely filtered out per the sizeConfig, the `banner` is still passed along as a valid mediaType in the bidRequest object while the `sizes` field is empty.
- any labels on the adunit/bid are just flat out ignored.  This results in bids/adUnits that should be excluded per the labels to be included in the bidRequest.

Per the current docs on the sizeConfig, it's not exactly clear what is meant to happen when using multi-format type bids.  Neither is it clear that sizeConfig exempts `video` or `native` type requests.  So this behavior is in a bit of a gray area since it falls between both sets of logic. 

## Proposed Fixes
Fix should adhere to the following points...

In the case of a multi-format bid that includes `banner`:
- if the label check passes but the sizes are entirely filtered out > the request should drop the `banner` part of the request only (ie the native part would still go through)
- **Open Question** if the label check fails - should we allow the other mediaType for that bid/adUnit to proceed into the request?  Or should we reject the adUnit/bid entirely to prevent a partial bid?  

I believe the latter option would lead to cleaner logic and avoid more edge scenarios; such as when different labels are set to different bids under the same bidder within the same adUnit.

In the case of a multi-format bid that doesn't include `banner`:
- then the sizeConfig/label logic could be skipped

In the case of single format bids:
- if the bid is a banner, the existing logic with the sizeConfig filtering/labels should be used
- if the bid is either `video` or `native`, then the sizeConfig/label logic could be skipped as its done today; we should update the prebid.org docs to clarify this point better.

## Sample configs
```

var adUnits = [
  {
    "code": "test_multi_banner_native",
    "mediaTypes": {
      "native": {
        "title": {
          "required": true,
          "len": 80
        }
      },
      "banner": {
        "sizes": [[300, 250], [300, 600]]
      }
    },
    "bids": [
      {
        "params": {
          "placementId": "111"
        },
        "labelAll": ["home", "desktop"],
        "bidder": "appnexus"
      }
    ]
  }
]

```

```
[
  {
    "mediaQuery": "(min-width: 1000px)",
    "sizesSupported": [[300, 250], [300, 600], [300, 900], [300, 1000], [300, 1050], [728, 90], [970, 90], [970, 250], [1000, 90], [1000, 200], [1000, 300], [1000, 350]],
    "labels": ["desktop"]
  },
  {
    "mediaQuery": "(min-width: 768px) and (max-width: 999px)",
    "sizesSupported": [[300, 250], [300, 600], [320, 50], [320, 100], [728, 90]],
    "labels": ["tablet"]
  },
  {
    "mediaQuery": "(min-width: 0px) and (max-width: 767px)",
    "sizesSupported": [[320, 50], [320, 100], [300, 250], [300, 100]],
    "labels": ["mobile"]
  }
]
```
Note - it was intended the `home` label wasn't defined in the sizeConfig.  This came from a test where it was meant to show the `labelAll` logic working only when both labels were in use.

jsnellbaker:
## Type of change

- [x] Bugfix

## Description of change
This PR is a fix for #3894 

It modifies the logic used by the sizeConfig when processing a multiformat bid.  

If the bid includes a `banner` `mediaType`, then the sizeConfig logic will be respected in that if the label rules don't pass or all the sizes are filtered out, the entire adUnit/bid (whichever is level is checked) will be removed from the request.  

If the bid does not include the `banner` type (even when the multiformat bid is just native + video), then sizeConfig logic is not applied.

## Other information
Docs PR: https://github.com/prebid/prebid.github.io/pull/1366

### DIFF
--- a/src/sizeMapping.js
+++ b/src/sizeMapping.js
@@ -88,9 +88,9 @@ export function resolveStatus({labels = [], labelAll = false, activeLabels = []}
 
   let results = {
     active: (
-      allMediaTypes.length > 1 || (allMediaTypes.length === 1 && allMediaTypes[0] !== 'banner')
+      allMediaTypes.every(type => type !== 'banner')
     ) || (
-      allMediaTypes[0] === 'banner' && deepAccess(mediaTypes, 'banner.sizes.length') > 0 && (
+      allMediaTypes.some(type => type === 'banner') && deepAccess(mediaTypes, 'banner.sizes.length') > 0 && (
         labels.length === 0 || (
           (!labelAll && (
             labels.some(label => maps.labels[label]) ||
--- a/test/spec/sizeMapping_spec.js
+++ b/test/spec/sizeMapping_spec.js
@@ -66,13 +66,6 @@ describe('sizeMapping', function () {
       }
       return matchMediaOverride;
     });
-
-    sandbox.stub(window, 'matchMedia').callsFake((...args) => {
-      if (typeof matchMediaOverride === 'function') {
-        return matchMediaOverride.apply(window, args);
-      }
-      return matchMediaOverride;
-    });
   });
 
   afterEach(function () {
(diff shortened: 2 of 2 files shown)


jaiminpanchal27:
test comment to check github integration in slack

jaiminpanchal27:
test comment to check github integration in slack 2

jsnellbaker:
test comment

PR #3940, opened 2019-06-24
OpenX should run only banner auction if it is multi format prebid auction

deivydas:
## Type of change

- [x] Bugfix

## Description of change

OpenX adaptor is not supporting multiformat auction and because of this reason banner mediaType should get a priority in a such auction

### DIFF
--- a/test/spec/modules/openxBidAdapter_spec.js
+++ b/test/spec/modules/openxBidAdapter_spec.js
@@ -178,6 +178,41 @@ describe('OpenxAdapter', function () {
       });
     });
 
+    describe('when request is for a multiformat ad', function () {
+      describe('and request config uses mediaTypes video and banner', () => {
+        const multiformatBid = {
+          bidder: 'openx',
+          params: {
+            unit: '12345678',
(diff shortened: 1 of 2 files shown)


test/spec/modules/openxBidAdapter_spec.js
```diff
@@ -178,6 +178,41 @@ describe('OpenxAdapter', function () {
...
+          transactionId: '4008d88a-8137-410b-aa35-fbfdabcb478e'
+        };
+        it('should return true multisize when required params found', function () {
+          expect(spec.isBidRequestValid(multiformatBid)).to.equal(true);
+        });
+
+        it('should send bid request to openx url via GET, with mediaType specified as banner', function () {
+          const request = spec.buildRequests([multiformatBid]);
+          expect(request[0].url).to.equal(`//${multiformatBid.params.delDomain}${URLBASE}`);
+          expect(request[0].data.ph).to.be.undefined;
```
jimee02: FYI: you don't need to check for this since it's not required for the test

PR #3957, opened 2019-07-01
AOL Adapter: Perform requests over https by default.

slimkrazy:
## Type of change
- Other

## Description of change
Update hostname to use go-forwards domain
Perform requests over HTTPS by default. 


slimkrazy:
Submitting too soon.

PR #3961, opened 2019-07-03
Teads-adapter: return placementId in response and gpdr on sync url

valsouche:
## Type of change

- [x] Feature

## Description of change
We have added placementId in bid response and added gdpr infos on sync url.


valsouche:
retest

PR #3963, opened 2019-07-03
Teads-adapter: Read placementId from response and add sync informations

valsouche:
## Type of change

- [x] Feature

## Description of change
- Read placementId in interpretResponse
- Add some tracking informations during sync call

### DIFF
--- a/modules/teadsBidAdapter.js
+++ b/modules/teadsBidAdapter.js
@@ -95,16 +95,19 @@ export const spec = {
   },
 
   getUserSyncs: function(syncOptions, responses, gdprConsent) {
-    let gdprIab = {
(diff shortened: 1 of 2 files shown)


valsouche:
Hello,
Do we need to have a green build to be reviewed ? It doesn't look like red because of our changes.
Thank you in advance !

valsouche:
Hello @sumit116 , thanks for review.
We fixed the console error in our last commit.

We didn't know that we have to give a test campaign access for this kind of changes (we did that only during the first adaptor pull request and not for others), we will send you ids as soon as there are available.

Thank you in advance.

valsouche:
Hello @sumit116 , here is a test campaign:
`"placementId": 96888,
"pageId": 89489`

You can try these ids inside the hello word adUnit configuration.

Thank you again.

PR #3965, opened 2019-07-04
Integrate id link system

mamatic:
## Type of change

- [ ] Feature

## Description of change

- Add Identity Link submodule in module userId to fetch user’s identity link envelope and add it to the bid request. It is very similar to other userId submodule like Id5id and unifiedID.

- A link to a PR on the docs repo at https://github.com/prebid/prebid.github.io/pull/1377


modules/userId/index.js
```diff
@@ -423,5 +425,6 @@ init(config);
 // add submodules after init has been called
 attachIdSystem(pubCommonIdSubmodule);
 attachIdSystem(unifiedIdSubmodule);
+attachIdSystem(identityLinkSubmodule);
```
idettman: The `attachIdSystem` should be used in the file `identityLinkSystem.js`, the `unifiedId` and `pubComonId` are the only submodules included by default, and they will also be optional starting with Prebid 3.0.

idettman:
Also, conflicts need resolution

modules/userId/index.js
```diff
@@ -77,6 +78,7 @@ import CONSTANTS from '../../src/constants.json';
 import {module} from '../../src/hook';
 import {unifiedIdSubmodule} from './unifiedIdSystem.js';
 import {pubCommonIdSubmodule} from './pubCommonIdSystem.js';
+import {identityLinkSubmodule} from './identityLinkSystem.js';
```
idettman: Remove the import for `identityLinkSubmodule` - the identity link submodule handles the registering in it’s submodule js.

modules/userId/index.js
mamatic: Hi @idettman,
Thank you for the explanation. I'm changed implementation of identityLink submodule as you suggested. Follow the code in ID5 Submodule...

modules/userId/index.js
mamatic: Done

PR #3969, opened 2019-07-05
Added 'ceh' config property in Criteo bid adapter

leonardlabat:
## Type of change
- [x] Feature

## Description of change
We'd like to add the ability for publishers to give to our adapter the hashed email of the current user.
Syntax for this will be the following.
```
  pbjs.setConfig({
    criteo: {
      ceh: 'hashed mail'
    }
  });
```


jsnellbaker:
Hi @leonardlabat,

Instead of passing this new value through the `setConfig`, could you pass it through your bidder params in the adUnit instead?  That would be the more ideal location to pass this type of bidder specific information into the auction.

Thanks,
Jason

leonardlabat:
Hi @jsnellbaker,

Yes that was our first thought. But that would require the publisher to repeat the hashed email on each ad unit. However, this is some kind of global parameter about the user (a bit like gdpr), something that we'll send at the root of the http request to our bidder, and not along each bid request.

Do you see what I mean ?

Regards,
Leonard

jsnellbaker:
Hi @leonardlabat 
I understand your point, but at this time - this type of data should still be handled through the adUnit.  

There is discussion about establishing a defined way to pass bidder unique data through the config (https://github.com/prebid/Prebid.js/issues/3687) which might be in line with what you wanted to setup here.  But until that process is decided and implemented, would you be able to pivot the changes here?

mafaigna:
Hi @jsnellbaker ,
We are excited indeed to see the standardisation of the way to pass first-party data in Prebid. However, at this stage, adapters have been adopting different ways to achieve this purpose - including the use of setconfig. The latter seems easier from publisher standpoint. What are your concerns with regard to the user of setconfig - waiting for FPD standardization?

mafaigna:
ping @jsnellbaker

leonardlabat:
Hi @jsnellbaker,

Thanks for the review. I've just pushed a commit to add this call to config.resetConfig.

Leonard

Issue #3964, opened 2019-07-03
Extend ESLint


jaiminpanchal27:
## Type of issue
Feature Request

## Description
With new submodules pattern implemented recently here, we are not allowing any module to import other modules. Modules can only import what is in src folder. Contributors adding code to modules may not know of this requirement and reviewers too may miss while approving pull requests. I think eslint can rescue us here.

Extend ES lint to throw error when modules are importing other modules. Also another good to have feature with this would be to make sure modules use relative path while importing.

Sample code which should throw error

```
// Consider two files 
// File 1 modules/currency.js
// File 2 modules/sampleBidAdapter.js

// sampleBidAdapter.js
...
import * from './currency'; // This should not be allowed.
...
...
```

```
// Consider two files 
// File 2 modules/sampleBidAdapter.js

// sampleBidAdapter.js
...
import * from 'src/utils'; // This should not be allowed.
import * from '../src/utils'; // Correct way to import module
...
...
```

Fawke:
Hi Jaimin,

I've submitted a PR https://github.com/prebid/Prebid.js/pull/3970 for this issue. Unfortunately, Circle CI is failing, I've to figure out why.

PR #3970, opened 2019-07-05
Add Eslint rules to throw error when importing other modules from within modules directory.

Fawke:
## Type of change
- [X] Feature

## Description of change

Related to https://github.com/prebid/Prebid.js/issues/3964.

### DIFF
--- a/modules/.eslintrc.js
+++ b/modules/.eslintrc.js
@@ -1,0 +1,14 @@
+module.exports = {
+  "rules": {
+    // This rule restricts import of other modules within the module directory
+    // and also ensures use of relative path when importing from src/
+    "no-restricted-imports": ["error", {
+      "patterns": ["./*", "src/*"]
+    }],
+    // Needed to support CommonJS style require statements.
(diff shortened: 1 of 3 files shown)


snapwich:
Just noticed this pull-request.  I made a similar stab at it in #3976 

If we can public eslint plugins work through configuration I think that's preferable; not sure if it's possible though so I ended up making a custom plugin.  Some of things I wanted were:
- a whitelist of modules allows per directory paths (as opposed to a blacklist, or restrictions as you have here)), it might be possible with `eslint-plugin-import` but I couldn't get it working quite right.
- validate that modules can be resolved _and_ are not `module` entry points.  The first part can be handled with `eslint-plugin-import` but I'm not sure about the later part.

If you can get those constraints working with this pull-request then I'm all for it, I'm not sure `no-restricted-imports` and `no-restricted-modules` are the rules we want though.

jaiminpanchal27:
Closing this PR in favour of https://github.com/prebid/Prebid.js/pull/3976

PR #3976, opened 2019-07-08
add eslint plugin for checking imports

snapwich:
## Type of change
- [x] Feature

## Description of change
Added a custom eslint plugin to validate that imports are relative, that modules don't import other modules, and that only dependencies from a dependency whitelist can be included in src or modules files.

Related to #3964 and #3970


snapwich:
taking a stab at solving this with the [no-internal-modules](https://github.com/benmosher/eslint-plugin-import/blob/master/docs/rules/no-internal-modules.md) plugin.  if that doesn't work then we can merge this.

snapwich:
Doesn't work :(  `no-internal-modules` only detects importing child modules from a parent and the `allow` field is just for exceptions to that check.  I think this custom plugin is probably the only solution for now.  Ready for merge!

PR #3980, opened 2019-07-09
Fix #3939 - update firefox specific code in renderAd function

### LINKED ISSUE #3939: Prebid.js SSP impression tracking and viewability broken since v2.3.0 on Edge and IE11
## Type of issue
Bug

## Description
On upgrading from Prebid 1.37 to 2.11.0 we encountered a large discrepancy between DFP and all ur Prebid Partners. We isolated to IE11/Edge. Impressions for these browsers in SSP consoles were effectively 0 on v2.11.0.  Upon further investigation DFP/GAM viewability for Edge/IE11 was zero. Roughly 10% of impressions for us, so small but not insignificant. We rolled back and discrepancy recovered.

There was no console error in IE/Edge. On Testing different versions of Prebid, I isolated it to being introduced in Prebid.js v2.3.0. 

On further testing I believe it was introduced by https://github.com/prebid/Prebid.js/pull/3524

document.open does not appear to work properly on IE11/Edge.

if that bug was for Firefox only, will:

          if(navigator.userAgent.toLowerCase().indexOf('firefox') > -1){
            doc.open('text/html', 'replace');
          }

suffice as a solution to add to ~/src/prebid.js?

## Steps to reproduce
Build any version of prebid.js v2.3.0+, on Edge/IE when DFP/GAM serves the prebid impression, check the viewability on the line in the GAM report.

## Test page
NA

### Expected results

Prebid won ads should show viewability in GAM reports for Edge and IE11 and impression tracking pixels should fire for SSPs and DFP/GAM.

### Actual results

impression tracking not firing on Edge and IE.

## Platform details
Versions of prebid.js v2.3.0+ onwards on Windows Browsers:  IE11 and Edge.

jsnellbaker:
## Type of change

- [x] Bugfix

## Description of change
This change updates the firefox specific logic in the `renderAd` function to only run if the user is using a version of Firefox below 67 (since Mozilla fixed the original bug in 67.0).

Additional references relevant to this overall change:
- original Prebid.js issue that added the firefox specific code - https://github.com/prebid/Prebid.js/pull/3524
- Mozilla issue that was noted on the above issue - https://bugzilla.mozilla.org/show_bug.cgi?id=148794
- Actual Mozilla issue (was linked in the Mozilla issue above) that contained the fix, release in Firefox 67.0 - https://bugzilla.mozilla.org/show_bug.cgi?id=1489308

### DIFF
--- a/src/prebid.js
+++ b/src/prebid.js
@@ -327,7 +327,7 @@ $$PREBID_GLOBAL$$.renderAd = function (doc, id) {
         } else if (ad) {
           // will check if browser is firefox and below version 67, if so execute special doc.open()
           // for details see: https://github.com/prebid/Prebid.js/pull/3524
-          // TODO remove this browser specific code at later date
+          // TODO remove this browser specific code at later date (when Firefox < 67 usage is mostly gone)
           if (navigator.userAgent && navigator.userAgent.toLowerCase().indexOf('firefox/') > -1) {
             const firefoxVerRegx = /firefox\/([\d\.]+)/;
             let firefoxVer = navigator.userAgent.toLowerCase().match(firefoxVerRegx)[1]; // grabs the text in the 1st matching group
--- a/src/prebid.js
+++ b/src/prebid.js
@@ -325,7 +325,16 @@ $$PREBID_GLOBAL$$.renderAd = function (doc, id) {
           const message = `Error trying to write ad. Ad render call ad id ${id} was prevented from writing to the main document.`;
           emitAdRenderFail(PREVENT_WRITING_ON_MAIN_DOCUMENT, message, bid);
         } else if (ad) {
-          doc.open('text/html', 'replace');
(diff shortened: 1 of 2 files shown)


src/prebid.js
```diff
@@ -325,7 +325,16 @@ $$PREBID_GLOBAL$$.renderAd = function (doc, id) {
           const message = `Error trying to write ad. Ad render call ad id ${id} was prevented from writing to the main document.`;
           emitAdRenderFail(PREVENT_WRITING_ON_MAIN_DOCUMENT, message, bid);
         } else if (ad) {
-          doc.open('text/html', 'replace');
+          // will check if browser is firefox and below version 67, if so execute special doc.open()
+          // for details see: https://github.com/prebid/Prebid.js/pull/3524
+          // TODO remove this browser specific code at later date
```
mkendall07: I would just add here `...later date when Firefox < 67 usage is mostly gone`

PR #3988, opened 2019-07-11
Safeframe resize fix 

jaiminpanchal27:
## Type of change
- [x] Bugfix

## Description of change
This PR fixes resizing issue in case of banner safeframe. https://github.com/prebid/prebid-universal-creative/issues/66

When you refresh the ad unit, dfp loads a new div for the creative and marks old div as `to be removed`. Prebid core was resizing the div which will be removed once rendering is complete.

Screen Shot 2019-07-10 at 4 19 05 PM

Tested with DFP and APAS. 

## Other information
@benjaminclot


vkimcm:
https://developer.mozilla.org/en-US/docs/Web/CSS/:last-child

it looks like last-child isn't supported in safari according to this link

PR #3845, opened 2019-05-21
getCpmInNewCurrency to use current value of bid.cpm and bid.currency

pm-harshad-mane:
## Type of change
- [X] Bugfix for #3837 

## Description of change
bid.getCpmInNewCurrency should use the latest value of bid.cpm and bid.currency as we allow publishers to update the value of bid.cpm (using bidAdjustment and bidResponse event) hence the bid.getCpmInNewCurrency API should return the bid.cpm equivalent value in the requested currency. 

#3837

PR #3996, opened 2019-07-15
Fixing the issue introduced in #3845 for rubi analytics adapter

robertrmartinez:
## Type of change

- [x] Bugfix

## Description of change
The pr introduced here https://github.com/prebid/Prebid.js/pull/3845/files brought out an issue in the rubi analytics adapter.

Since we were not calling the function with the bid obj attached, it was resulting in `this.cpm` to throw an exception since `this` would not be defined with how we are calling it.

This PR fixes this issue and adds a unit test to prevent this from happening again

### DIFF
--- a/test/spec/modules/rubiconAnalyticsAdapter_spec.js
+++ b/test/spec/modules/rubiconAnalyticsAdapter_spec.js
@@ -13,10 +13,6 @@ let ajv = new Ajv({
   allErrors: true
 });
 
-function clone(obj) {
-  return JSON.parse(JSON.stringify(obj));
-}
-
 let validator = ajv.compile(schema);
 
 function validate(message) {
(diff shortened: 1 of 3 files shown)


robertrmartinez:
I added the test before the fix and verified that it fails:

image


And now it passes after the change.

test/spec/modules/rubiconAnalyticsAdapter_spec.js
```diff
@@ -1,13 +1,22 @@
...
+  addBidResponseHook,
+} from 'modules/currency';
+
 let Ajv = require('ajv');
 let schema = require('./rubiconAnalyticsSchema.json');
 let ajv = new Ajv({
   allErrors: true
 });
 
+function clone(obj) {
```
idettman: I’d recommend using `utils.deepClone(obj)` from `utils.js` rather than defining clone here

robertrmartinez:
@idettman Thanks, updated both our test adapters to use utils.deepClone

PR #4002, opened 2019-07-16
Padsquad bid adapter

yeeldpadsquad:
## Type of change
- [x] New bidder adapter 

## Description of change
New Padsquad bid adapter

```
{
  bidder: 'padsquad',
  params: {
    unitId: 'test',
  }
}
```

Documentation 
https://github.com/prebid/prebid.github.io/pull/1390

### DIFF
--- a/modules/padsquadBidAdapter.js
+++ b/modules/padsquadBidAdapter.js
@@ -125,7 +125,6 @@ export const spec = {
         }
       }
     });
-    utils.logInfo('padsquad.getUserSyncs result=%o', syncs);
     return syncs;
   },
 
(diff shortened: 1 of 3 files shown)


yeeldpadsquad:
Seems like build failure is caused by other bid adapters

modules/padsquadBidAdapter.js
```diff
@@ -0,0 +1,133 @@
...
+
+        if (!syncOptions.iframeEnabled) {
+          syncs = syncs.filter(s => s.type !== 'iframe')
+        }
+        if (!syncOptions.pixelEnabled) {
+          syncs = syncs.filter(s => s.type !== 'image')
+        }
+      }
+    });
+    utils.logInfo('padsquad.getUserSyncs result=%o', syncs);
```
Fawke: The log message is a bit ambiguous. Would you mind fixing that or making it more coherent as to what it means?

yeeldpadsquad: actually that line may not even be necessary, we'll remove it altogether

PR #4004, opened 2019-07-16
Triplelift adapter image sync fallback

rmcnierney:
## Type of change

- [x] Feature

## Description of change
Adding a fallback for image user syncing in `getUserSyncs`. Will prioritize iframe syncing and use image if iframe is disabled.


bretg:
Assigning to @idettman for review -- Isaac, please take a look at whether this is going to work with the same URL. I'm not sure that the `type` gets sent along with the request -- can the code in the endpoint reliably know if it was called as part of an <iframe> or an ?

rmcnierney:
Closing this PR for the time being. TripleLift is working on a few internal updates before we can merge this.

PR #4006, opened 2019-07-18
GumGum adapter: push first indexed value in sizes array to bidResponses.height and width

WayneYang1:
## Type of change
- [x] Other

## Description of change
This pull request will change GumGum's bid adapter to return the width and height of the first indexed value in the sizes array. This change will allow our adapter to dynamically populate the ad size of our bid response to match the request size that the publisher sends us.

## Other information
CC: @susyt


WayneYang1:
@bretg please let me know if there is anything else to be adjusted here.

Issue #3997, opened 2019-07-15
Debug logging not displaying the prebid version since 2.0


andyblackwell:
## Type of issue
Bug

## Description
After version 2.0 the prebid version is not being logged to the console

## Steps to reproduce
Import the project as suggested in the example available in http://prebid.org/dev-docs/examples/basic-example.html
and notice that e.g. "INFO: Prebid.js v2.18.0 loaded" is not showing.

## Other information
dupe of closed bug #3700 
This issue was fixed in #3714 but then the fix was reverted in #3736 

my opinion: setConfig() calls should not wait until pbjs.processQueue() is called

snapwich:
Did you reference the correct pull-request?  I don't see the revert in 3776.

jsnellbaker:
This was the revert PR https://github.com/prebid/Prebid.js/pull/3736 @snapwich

andyblackwell:
whoops, thanks! updated in description

snapwich:
why was it reverted @jsnellbaker ? there's not much of a description in the pull-request.

jsnellbaker:
I think there were build issues at the time (I think they were in circleci) that resulted from this line of code:
https://github.com/prebid/Prebid.js/pull/3714/files#diff-9d9a6cd82f41984872a66a3ab0d440c4R16

I was trying to recreate/retest the changes yesterday, but I had to change gears after something came up.  I can take another look if you want.

snapwich:
yeah that'd be great.  I don't see why that change should cause any issues and I don't think it would have merged if it was causing CI issues.  but if it is causing issues, it'd be nice to know what they are so I can fix them.

jsnellbaker:
@snapwich I setup the changes again in this branch `debug_updates`.  The build compiles fine and works successfully on test pages.  The circleci tests also pass fine (which I saw after creating the above PR).

However when digging more into my past messages, I found the cause for the revert.  When running `gulp test-coverage` with these changes, it causes a number of exception errors to be thrown.  

Below is a copy of one of the messages:
```
SourceMap position not found for trace: {
[...]

PR #4007, opened 2019-07-18
Debug update and remove setConfig hook

### LINKED ISSUE #3700: Logging not showing the prebid version at start
## Type of issue
Bug 

## Description
After the Prebid version 2.0 the prebid version is not being logged at the beginning.

## Steps to reproduce
Import the project as suggested in the example available in http://prebid.org/dev-docs/examples/basic-example.html
and notice that e.g.`INFO: Prebid.js v2.8.0 loaded` is not showing.

jsnellbaker:
## Type of change

- [x] Bugfix

## Description of change
A re-submission of older an PR #3714 that was reverted from `master` via #3736.

Re-copying the description of the above ticket here for reference:

With the implementation of the new hooks library we added a `ready` functionality that would allow `async` hooks to be queued until the hooks were declared `ready` in `pbjs.processQueue`. This caused `config.setConfig` calls to be queued and `config.getConfigs` being called before `pbjs.processQueue` were not getting data (such as the call for printing prebid info in #3700). I updated the DEFAULT_DEBUG to be set immediately rather than relying on `config.setConfig` in the `utils` library, which I think makes a little more sense.

Also, the `pbjs.setConfig` hook was only being used for the `pre1api` module, which was removed in 2.0 so I think it's best we just remove the hook, which I've also done.

## Other information
Fixes #3700 and #3997

### DIFF
--- a/src/config.js
+++ b/src/config.js
@@ -10,10 +10,11 @@
 import { isValidPriceConfig } from './cpmBucketManager';
 import find from 'core-js/library/fn/array/find';
 import includes from 'core-js/library/fn/array/includes';
-import { hook } from './hook';
-const utils = require('./utils');
 
-const DEFAULT_DEBUG = false;
+const utils = require('./utils');
+const CONSTANTS = require('./constants');
+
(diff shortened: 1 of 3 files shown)


jsnellbaker:
Please hold on this PR - there is an issue that needs to be addressed.

PR #4010, opened 2019-07-19
Add Logicad for Publishers bid adapter

logicad:
## Type of change

- [x] New bidder adapter   

## Description of change

New Logicad for Publishers bid adapter

```
{
  bidder: 'logicad',
  params: {
    tid: 'test'
  }
}
```

- contact email of the adapter’s maintainer: prebid@so-netmedia.jp

- A link to a PR on the docs repo at  prebid/prebid.github.io/pull/1395

### DIFF
--- a/modules/logicadBidAdapter.js
+++ b/modules/logicadBidAdapter.js
@@ -1,0 +1,68 @@
+import {registerBidder} from '../src/adapters/bidderFactory';
+import {BANNER} from '../src/mediaTypes';
+
+const BIDDER_CODE = 'logicad';
+const ENDPOINT_URL = 'https://pb.ladsp.com/adrequest/prebid';
+
+export const spec = {
+  code: BIDDER_CODE,
+  supportedMediaTypes: [BANNER],
+  isBidRequestValid: function (bid) {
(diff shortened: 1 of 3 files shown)


modules/logicadBidAdapter.js
```diff
@@ -0,0 +1,68 @@
...
+      bids.push(seatbid.bid);
+    })
+    return bids;
+  },
+  getUserSyncs: function (syncOptions, serverResponses) {
+    if (serverResponses.length > 0 && serverResponses[0].body.userSync &&
+      syncOptions.pixelEnabled && serverResponses[0].body.userSync.type == 'image') {
+      return [{
+        type: 'image',
+        url: serverResponses[0].body.userSync.url
```
jsnellbaker: To confirm - if you're returning multiple bids, this code would only grab the first bid's userSync pixel.  The other bids' userSync pixels would be left out.

Is that expected?

logicad: Hi @jsnellbaker 

Thank you for the review.

Yes, it is as expected.

Only using the first bids will fulfill our requirements.

PR #4011, opened 2019-07-19
Inskin Bid Adapter - New Feature

cciocov:
## Type of change
- [X] Feature

## Description of change
- Updated the way bid price is set based on the ad server response
- Updated the impression URL


cciocov:
@bretg I've added testing for this, thanks for the feedback.

PR #4016, opened 2019-07-23
adformOpenRTB adapter: support currency param

braizhas:
## Type of change
- [x] Feature

## Description of change
Added new adformOpenRTB bid config option.

- contact email of the adapter’s maintainer
Scope.FL.Scripts@adform.com

- [x] official adapter submission

PR #4018, opened 2019-07-24
adformBidAdapter: currency module support

braizhas:
## Type of change
- [x] Feature

## Description of change
Pass configured currency to adserver.

Scope.FL.Scripts@adform.com

### DIFF
--- a/modules/adformBidAdapter.js
+++ b/modules/adformBidAdapter.js
@@ -1,6 +1,7 @@
 'use strict';
 
 import {registerBidder} from '../src/adapters/bidderFactory';
+import { config } from '../src/config';
 import { BANNER, VIDEO } from '../src/mediaTypes';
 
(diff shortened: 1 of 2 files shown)


bretg:
Does this obsolete https://github.com/prebid/Prebid.js/pull/4016 ?  Looks like the `rcur` param doesn't exist in your documentation -- please enter a separate PR against https://github.com/prebid/prebid.github.io/blob/master/dev-docs/bidders/adform.md

bretg:
Never mind - that's the `adFormOpenRTB` adapter. Got it.

braizhas:
I would prefer to have a single way for currency configuration. `rcur` parameter left to not break existing configurations only. I don't want to document it to avoid further usage.

bretg:
sounds good

PR #4019, opened 2019-07-24
adformOpenRTB adapter: size targeting using aspect ratios

braizhas:
## Type of change
- [x] Feature

## Description of change
Pass `aspect_ratios` config to adserver.

Scope.FL.Scripts@adform.com

### DIFF
--- a/modules/adformOpenRTBBidAdapter.js
+++ b/modules/adformOpenRTBBidAdapter.js
@@ -69,7 +69,7 @@ export const spec = {
           let wmin, hmin, w, h;
           let aRatios = bidParams.aspect_ratios;
 
-          if (aRatios) {
+          if (aRatios && aRatios[0]) {
             aRatios = aRatios[0];
             wmin = aRatios.min_width || 0;
             hmin = aRatios.ratio_height * wmin / aRatios.ratio_width | 0;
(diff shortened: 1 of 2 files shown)


modules/adformOpenRTBBidAdapter.js
```diff
@@ -66,11 +66,28 @@ export const spec = {
         };
         if (props) {
           asset.id = props.id;
+          let wmin, hmin, w, h;
+          let aRatios = bidParams.aspect_ratios;
+
+          if (aRatios) {
+            aRatios = aRatios[0];
+            wmin = aRatios.min_width || 0;
+            hmin = aRatios.ratio_height * wmin / aRatios.ratio_width | 0;
```
msm0504: Are there default values for `aRatios.ratio_height` and `aRatios.ratio_width`?  Is it possible that either of these could be 0 or undefined?

modules/adformOpenRTBBidAdapter.js
```diff
@@ -66,11 +66,28 @@ export const spec = {
         };
         if (props) {
           asset.id = props.id;
+          let wmin, hmin, w, h;
+          let aRatios = bidParams.aspect_ratios;
+
+          if (aRatios) {
```
msm0504: I'd suggest adding `&& aRatios[0]` to check that the 1st element is also defined

modules/adformOpenRTBBidAdapter.js
braizhas: It's just a fallback to minWidth/minHeight targeting, so it's OK to have `NaN` as an expression result. Bitwise OR operator will convert it to 0.

PR #4021, opened 2019-07-24
remove console log statement in logicadBidAdapter unit tests

jsnellbaker:
## Type of change
- [x] Other

## Description of change
In the `logicadBidAdapter_spec.js` file, there was a rogue `console.log` statement in one of the unit tests.  This PR removes the statement so it's not present in the console output whenever we run the unit tests or make a build.

CC: @logicad


logicad:
Sorry for the mistake and thanks for the fix.

PR #4023, opened 2019-07-25
[mgid] add referrer and utc offset to bid request

Gaudeamus:
## Type of change

- [X] Feature

add referrer and utc offset to bid request

prebid@mgid.com


bretg:
@mike-chowla - please consider overlap of the `pageinfo` function here with what we have in utils.js. Thanks.

mike-chowla:
> @mike-chowla - please consider overlap of the `pageinfo` function here with what we have in utils.js. Thanks.

I don't see a function in utils that gives a reference to the top most window back.  If the caller just needs the referrer, we have a function for that but given the code is using innerWidth & innerHeight, I don't see a way to do this with the existing utils functions or passed in `refererInfo`

PR #4024, opened 2019-07-25
override default fun-hooks module configuration

aszydlo:
## Type of change

- [ ] Bugfix
- [ ] Feature
- [ ] New bidder adapter   
- [X] Build related changes
- [ ] CI related changes
- [ ] Other

## Description of change
Possibility to overwrite default fun-hooks module configuration used by Prebid.js while importing them.

Need to change default ``useProxy`` value of fun-hooks configuration https://www.npmjs.com/package/fun-hooks#configuration in some cases.

Example: 
```
$$PREBID_GLOBAL$$.hookSettings = {
  useProxy: false
}
```

### DIFF
--- a/src/hook.js
+++ b/src/hook.js
@@ -2,7 +2,8 @@
 import funHooks from 'fun-hooks/no-eval';
 
 export let hook = funHooks({
-  ready: funHooks.SYNC | funHooks.ASYNC | funHooks.QUEUE
+  ready: funHooks.SYNC | funHooks.ASYNC | funHooks.QUEUE,
+  ...$$PREBID_GLOBAL$$.hookSettings
 });
 
 export const getHook = hook.get;


snapwich:
Can I ask why you'd want to disable the use of `Proxy`?

aszydlo:
> Can I ask why you'd want to disable the use of `Proxy`?

Hi,
some of the pages where I use Prebid.js to display video override global Proxy class. This is caused by PureMvc framework, embedded in video player iframe which I use, for example: https://tinyurl.com/y2ga4e56
I have this problem with a lot of services and this way is the best option for me to handle this issue.

snapwich:
closing in favor of #4030

I'd rather not have `fun-hooks` be configurable through Prebid.js mainly because it has to be configured before loading (as you noticed) which is user-unfriendly and requires hacky solutions outside of the standard `pbjs.setConfig()`.  

However, this really is a problem in determining whether or not `Proxy` is actually the native `Proxy`.  I updated the `fun-hooks` library to do a check for `Proxy.revocable` which is part of the `Proxy` API (which the `Proxy` global made by PureMVC is not), so this new check should fail on your page and it should fall-back gracefully to the non-Proxy hooks.

aszydlo:
Thanks for your help, now everything works fine.

PR #4038, opened 2019-07-30
Update karma to address vulnerabilities. 

Fawke:
## Type of change

- [X] Build related changes
- [X] CI related changes

## Description of change

Related to https://github.com/prebid/Prebid.js/pull/3966. The earlier PR was reverted because Circle CI failed. This PR attempts to fix the failing CI along with the necessary updates required to keep vulnerabilities to a minimum.


Fawke:
Update - command `gulp serve` is currently exiting out because of this change, looking into the issue.

Fawke:
> Update - command `gulp serve` is currently exiting out because of this change, looking into the issue.

@jsnellbaker I've fixed this. The command `gulp serve` now works as earlier without exiting.

PR #4040, opened 2019-07-30
Fix referrer encoding in The Media Grid Bid Adapter

TheMediaGrid:
## Type of change

- [x] Bugfix

## Description of change

Stopped using encodeURIComponent where it is not necessary, to avoid double encoded string in referrer.

```
{
  bidder: 'grid',
  params: {
    uid: '1'
  }
}
```

- grid-tech@themediagrid.com

## Other information

This fix for PR https://github.com/prebid/Prebid.js/pull/3968


TheMediaGrid:
Hey guys, this is actually a hotfix for #3968 and it is currently a blocker from getting the correct referer. Any chances you could push that to release ASAP? We have a couple of publishers ready to go live pending this. Thanks in advance.

jsnellbaker:
@TheMediaGrid Unfortunately, this would have to wait until the next release.  We do releases every Tuesday; as this is currently merged, this update will definitely be part of the upcoming 2.27.0 release scheduled for Aug 6.

CC: @mkendall07

PR #4053, opened 2019-08-02
Fix removeAdUnit

benjaminclot:
## Type of change

- [x] Bugfix

## Description of change
When there are multiple adUnits with the same code, only one of them gets removed. This fixes that.

### DIFF
--- a/src/prebid.js
+++ b/src/prebid.js
@@ -389,13 +389,8 @@ $$PREBID_GLOBAL$$.removeAdUnit = function (adUnitCode) {
     adUnitCodes = [adUnitCode];
   }
 
-  adUnitCodes.forEach((adUnitCode) => {
-    for (let i = 0; i < $$PREBID_GLOBAL$$.adUnits.length; i++) {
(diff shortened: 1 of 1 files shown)


benjaminclot:
All tests pass but CircleCI keeps returning timeouts :-/

benjaminclot:
@idettman The branch has been updated with master and there's still a timeout error.

idettman:
Nice catch @snapwich, so would the fix be something like this:
```
  adUnitCodes.forEach((adUnitCode) => {
    for (let i = $$PREBID_GLOBAL$$.adUnits.length; i > 0; i--) {
      if ($$PREBID_GLOBAL$$.adUnits[i].code === adUnitCode) {
        $$PREBID_GLOBAL$$.adUnits.splice(i, 1);
      }
    }
  })
```

snapwich:
@idettman should be `i >= 0`, but other than that I think it's good.

benjaminclot:
@snapwich @idettman requested change has been committed

PR #4065, opened 2019-08-06
update logic of ad_types field in appnexusBidAdapter

jsnellbaker:
## Type of change

- [x] Feature

## Description of change
This PR changes how the `ad_types` field of the appnexus request is populated in relation to banner media type.

Previously the field was populated automatically with `'banner'` if the `mediaType(s)` field of the bid object was empty or if either of those fields were set to `banner`.  The latter case actually always happened since Prebid.js treats any adUnit request without a defined `mediaTypes` object as `banner` and creates a `mediaTypes` field in the bid with a banner object in this scenario.

Now with this change, the `banner` `ad_type` is only populated if the original `adUnit` contained a `mediaTypes.banner` object.  If the `adUnit.mediaTypes` object is missing, then the field is left out of the appnexus request.  In terms of the delivery decision for the bid, this scenario would defer to the types that are setup within the placement and use that to choose the type of bid to find/return.

### DIFF
--- a/modules/appnexusBidAdapter.js
+++ b/modules/appnexusBidAdapter.js
@@ -3,6 +3,7 @@ import * as utils from '../src/utils';
 import { config } from '../src/config';
 import { registerBidder, getIabSubCategory } from '../src/adapters/bidderFactory';
 import { BANNER, NATIVE, VIDEO, ADPOD } from '../src/mediaTypes';
+import { auctionManager } from '../src/auctionManager';
 import find from 'core-js/library/fn/array/find';
 import includes from 'core-js/library/fn/array/includes';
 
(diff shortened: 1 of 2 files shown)


mkendall07:
placing on hold for now due to some concerns about multi format.

jsnellbaker:
@mkendall07 Per the internal feedback, this change should be good to implement.  Can you take another look when you have the chance?  Thanks!

lsv1:
Hey all,

When I define adUnits without mediaTypes I get a parsing error returned, it looks like it's related to missing sizes in the request payload to AppNexus.

I think sizes must still be defined and sent over to avoid the parse error. Check out my Fiddle for a working example - http://jsfiddle.net/nfm0ahbk/2/.

PR #4066, opened 2019-08-06
 Criteo Adapter - Code cleaning & added missing tests to increase code coverage

leonardlabat:
## Type of change

## Description of change
We'd like to tackle the technical debt in our adapter by refactoring it a bit and, more important by adding missing tests for features we've added in the past months. This will increase our code coverage rate to a decent value, beyond the one described in the Prebid policy.


leonardlabat:
I've just seen that I messed it with the merge from master ...

Tell me if you prefer me to abandon this one and start a new fresh one with only my own changes on the criteoAdapter & spec.

Thanks

Issue #3368, opened 2018-12-10
Refactor timeout settings/options used in Prebid.js


jsnellbaker:
## Type of issue
Feature Request

## Description
Spawning from a conversation in #3291 (see [here](https://github.com/prebid/Prebid.js/pull/3291#discussion_r239602606) specifically), there are currently a number of settings/levels offered around managing the `timeout` used by Prebid.js and it's slowly getting to the point where it's not only hard to understand what each setting does exactly but where these settings may intersect (or compete) with each other.  

The scope of these settings' influence range to outside the auction itself to subtle influences around the adapters' response window.  Taken (partially) from the comment in the above referenced ticket, below is a current recap of the `timeout` settings:

1.  The `user-specified timeout` value given to PBJS (bidderTimeout - set either in `pbjs.requestBids({config})` as a param of the object config  or via `setConfig({bidderTimeout: 3000});`).
2.  The `effective framework timeout` -- PBJS adds 400ms (or timeoutBuffer) to this value and then calls bidsBackHandler when the timer pops.
3.  The `user-specified server adapter timeout` (s2sConfig.timeout) - what we currently tell PBS it's timeout is
4.  `failsafe timeout` - a timer the page puts around PBJS as a whole. http://prebid.org/dev-docs/faq.html#when-starting-out-what-should-my-timeouts-be

**Note** The original PR was attempting to add an `effective adapter timeout` that would evaluate the auction timeout (ie number 1 in the above list) and set a proportional timeout for the adapter (in certain use-cases).  This would have been in conflict with number 2, which is meant to add a buffer for adapters to respond.

**Note** on 1 from list above; this single timeout is used as both the overall auction timer as well as well as the timeout provided to the XHR request for each adapter when they make their request for bids.  This is not ideal, as the auction timer needs to be shared by the core Prebid code and the adapters.  If the adapter takes all the time to respond, then the auction has already timed out (since its timer was running slightly ahead of the adapters).

## Goal/Proposal
We need to streamline these options better and ensure their parts are working together rationally.  

Copying some questions/points from the original conversation relevant to this point, we need to establish the following:
-  what flexibilities in timeout are needed? client-vs-server? different timeouts for different adapters?
-  what do we call Prebid.js' view of the timeout as opposed any adjusted versions we provide to adapters?

[...]

PR #4070, opened 2019-08-07
Timeout coefficients

jaiminpanchal27:
**Do not merge**

## Type of change
- [x] Feature

## Description of change
Another attempt to solve https://github.com/prebid/Prebid.js/issues/3368

Publishers will only need to worry about 2 timeout values. 
1. The `user-specified timeout` value given to PBJS (bidderTimeout).
2. `failsafe timeout` a timer the page puts around PBJS as a whole. http://prebid.org/dev-docs/faq.html#when-starting-out-what-should-my-timeouts-be

We will calculate other 2 timeout values `timeoutBuffer` and `s2sConfig.timeout` using our default coefficient. Defaults will be only available in 3.0
3. timeoutBufferCoefficient - 0.20 
4. s2sConfig.timeoutCoefficient - 0.70

Logic is simple, evaluate timeoutBuffer and s2sConfig timeout on call to requestBids or on demand and make sure that new timeout does not exceed auction timeout. If it is higher use default value.

How can publisher set their coefficient ? 
```
pbjs.setConfig({
  timeoutBufferCoefficient: 0.20,
  s2sConfig: {
    timeoutCoefficient: 0.70
  } 
})
```

I have added some placeholders for github issues and pr. If this is approved I will create an issue and update

### DIFF
--- a/modules/prebidServerBidAdapter/index.js
+++ b/modules/prebidServerBidAdapter/index.js
@@ -40,7 +40,7 @@ let _s2sConfig;
 /**
  * @typedef {Object} S2SDefaultConfig
  * @property {boolean} enabled
- * @property {number} timeout
+ * @property {number} timeout @deprecated This property will be removed in 3.0. Please refer <TBD github pr> for more info
  * @property {number} maxBids
  * @property {string} adapter
  * @property {AdapterOptions} adapterOptions
--- a/src/auction.js
+++ b/src/auction.js
@@ -385,7 +385,15 @@ export function auctionCallbacks(auctionDone, auctionInstance) {
 }
 
 export function doCallbacksIfTimedout(auctionInstance, bidResponse) {
-  if (bidResponse.timeToRespond > auctionInstance.getTimeout() + config.getConfig('timeoutBuffer')) {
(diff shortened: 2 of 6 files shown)


bretg:
So let me see if I have this right. This is the diagram I use to make sense of the various timeouts:

Screen Shot 2019-09-04 at 7 04 01 PM

What you're doing here is removing timeouts #3 and #4, but replacing them with a ratio relative to #2. At first glance, this doesn't save any config at all -- there are still options we have to document on the API page that make the system feel complicated.

But I suppose the idea is that the ratios are something that we should be able to set as a default and only change in bizarre edge cases. Does anyone have any data about the ratios?

Honestly my take it to leave things as they are -- this feels nearly as complicated and harder to think about.

bretg:
Chatted with @mkendall07. He says "the idea was to not make the coefficients configurable, thus making less decision making for the pub." He also notes that this was just a backlog item, not necessarily driven by a recent request. 

Before we make this change which I think makes timeouts even harder to understand, I would like pub feedback to see how much of a problem this is.

jaiminpanchal27:
Closing this as we feel that this approach does not solve the problem

PR #4077, opened 2019-08-09
Identity Link - Use ats library for retrieving envelope

mamatic:
## Type of change
- [ ] Feature

## Description of change

- Add logic to use ats (Authenticated Traffic Solution) library to retrieve IdentityLink envelope.  If ats library is included and initialised on publisher site, we use function retrieveEnvelope() to get IdentityLink envelope. If ats is not initialised or envelope not found, it call out URL that responds with user’s IdentityLink envelope.

Documentation is updated on already open pull request.
- A link to a PR on the docs repo at https://github.com/prebid/prebid.github.io/pull/1377


mamatic:
Hi @idettman ,
can you please do a review, when you have time?
Thanks in advance!

bretg:
Hi @mamatic - please rename this file to identityLinkIdSystem.js
That shouldn't stop @idettman from reviewing it though. :-)

mamatic:
Hi @bretg,
thanks for suggestion, file identityLinkSystem.js renamed to identityLinkIdSystem.js .

PR #4086, opened 2019-08-13
Fix #4059 - ensure native keys are not seen as custom targeting keys

### LINKED ISSUE #4059: multiple hb_native_image & hb_native_linkurl in native ads
i'm not sure where to ask this, so asking it here,  i'm trying to get native ads working and im coming across some unusual things in that i keep getting multiple hb_native_image & hb_native_linkurl  returns, such as:

hb_native_image:104df19a869aa2c3,hb_native_image:109d7bfc7d44dc13,hb_native_image:114a333a1bc49e34,hb_native_image:1189017d48786a57,hb_native_image:159909171c973231,hb_native_image:1573e8a4c52eb209

&
hb_native_linkurl:104df19a869aa2c3,hb_native_linkurl:109d7bfc7d44dc13,hb_native_linkurl:114a333a1bc49e34,hb_native_linkurl:1189017d48786a57,hb_native_linkurl:159909171c973231,hb_native_linkurl:1573e8a4c52eb209

i'm not sure if this is appnexus sending multiple ones or if its related to:
useBidCache: true

anyway, when it sends multiple i cant get the ad to render, when it send one on the string the ad renders  Is this know of or am i doing something wrong and should try to parse the multiple ones?

jsnellbaker:
## Type of change
- [x] Bugfix

## Description of change
Fixes #4059 

When multiple native bids were returned for the same adUnit in an auction with the `enableSendAllBids` flag enabled, the native keys weren't populated correctly.  

In particular:
- the non-winning bidder's set of bidder specific native keys weren't being generated at all (eg `hb_native_image_appn`)
- the base native keys (like `hb_native_linkurl`) contained an array of values from all the bids for that adUnit, rather than just the value from that winning bid

This behavior was due to native keys being treated as custom keys within the `targeting.getCustomKeys()` function.  

When the `mergeAdServerTargeting()` function got a list of 'custom' keys to process (ie the native keys in this case), it combined all the values from the native keys for each bid into the array list and stored it under the base native key.  This updated base key was stored under the `bid.adServerTargeting` field of the winning bid, which caused the mutated values to be carried through the rest of the targeting functions into the final state described above.  

This PR fixes the issue by including the native keys with the list of the base keys in the `getCustomKeys()` function, so that native keys are properly ignored when the function is trying to find actual custom keys.

### DIFF
--- a/src/targeting.js
+++ b/src/targeting.js
@@ -443,7 +443,7 @@ export function newTargeting(auctionManager) {
   }
 
   function getCustomKeys() {
-    let standardKeys = getStandardKeys();
+    let standardKeys = getStandardKeys().concat(NATIVE_TARGETING_KEYS);
     return function(key) {
       return standardKeys.indexOf(key) === -1;
     }
--- a/test/spec/unit/core/targeting_spec.js
+++ b/test/spec/unit/core/targeting_spec.js
@@ -97,6 +97,131 @@ const bid3 = {
   'ttl': 300
 };
 
+const nativeBid1 = {
+  'bidderCode': 'appnexus',
+  'width': 0,
+  'height': 0,
+  'statusMessage': 'Bid available',
+  'adId': '591e7c9354b633',
+  'requestId': '24aae81e32d6f6',
+  'mediaType': 'native',
+  'source': 'client',
+  'cpm': 10,
+  'creativeId': 97494403,
+  'currency': 'USD',
+  'netRevenue': true,
+  'ttl': 300,
+  'adUnitCode': '/19968336/prebid_native_example_1',
+  'appnexus': {
+    'buyerMemberId': 9325
+  },
+  'meta': {
+    'advertiserId': 2529885
+  },
+  'native': {
+    'title': 'This is a Prebid Native Creative',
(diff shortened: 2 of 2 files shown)


src/targeting.js
```diff
@@ -443,7 +443,7 @@ export function newTargeting(auctionManager) {
   }
 
   function getCustomKeys() {
-    let standardKeys = getStandardKeys();
+    let standardKeys = getStandardKeys().concat(NATIVE_TARGETING_KEYS);
```
mkendall07: If `NATIVE_TARGETING_KEYS` are part of standard keys, shouldn't we move this into the `getStandardKeys()` function?

jsnellbaker: If we moved it there, then the native keys would be affected by the `bidderSettings.sendStandardTargeting` property (which currently only focuses on the base keys via this logic [here](https://github.com/prebid/Prebid.js/blob/master/src/targeting.js#L384)).

At the moment, the native keys are still written when that flag is set to `false`.  If it was moved, then the natives aren't written.  Is that the behavior we want?

mkendall07: I suppose that would be a breaking change, so we leave as is for now. Thanks

PR #4090, opened 2019-08-15
Enable withCredentials on Unified ID ajax call

minh-daole-ttd:
## Type of change

- [x] Bugfix
- [ ] Feature
- [ ] New bidder adapter   
- [ ] Build related changes
- [ ] CI related changes
- [ ] Other

## Description of change

The change in this PR is to enable withCredentials = true in the Ajax request that is made to the Unified ID system in the User ID module. This setting needs to be enabled so that cookies are correctly sent to the endpoint and so that cookies are able to be set when the endpoint responds with Set-Cookie headers.
Reference: https://developer.mozilla.org/en-US/docs/Web/API/XMLHttpRequest/withCredentials

- contact email - minh.daole@thetradedesk.com

### DIFF
--- a/modules/userId/unifiedIdSystem.js
+++ b/modules/userId/unifiedIdSystem.js
@@ -49,7 +49,7 @@ export const unifiedIdSubmodule = {
           }
         }
         callback(responseObj);
-      }, undefined, { method: 'GET' });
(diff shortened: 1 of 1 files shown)


minh-daole-ttd:
Thank you @mkendall07 for reviewing! How can we get this code merged?

PR #4095, opened 2019-08-16
Minify with closure compiler | Do Not Merge

sumit116:
#4095 Type of change

- [ ] Build related changes

## Description of change
Use closure compiler for minification


mkendall07:
This pull request **introduces 2 alerts** when merging fbc4840d25adebe898f9b0a2ad57f11c17334367 into d61756d82bc21870e2b8ad6e784026af62fc89d4 - [view on LGTM.com](https://lgtm.com/projects/g/prebid/Prebid.js/rev/pr-ef8318926cabbbd67ceef2cccfc12dbb8c09b96b)

**new alerts:**

* 2 for Unused variable, import, function or class

---

---

snapwich:
Before you put in all this work, can you share your goal and what compression results you're seeing (or hoping to see)? Using the closure compiler to aggressively compress can be quite dangerous for us (especially with so much third-party contributed code), whereas currently our minifcation is done pretty much risk free.

sumit116:
> Before you put in all this work, can you share your goal and what compression results you're seeing (or hoping to see)? Using the closure compiler to aggressively compress can be quite dangerous for us (especially with so much third-party contributed code), whereas currently our minifcation is done pretty much risk free.

sumit116:
@snapwich, the goal is to find the scope of using closure compiler with advance compilation in Prebid.js. I could compress the bundle (prebid.js) upto 862KB (without gzip) with it, which is around 1mb when we use uglify-js. 
Yes, there is risk of lossing some data with advance compilation but there are ways to handle it (according to closure compiler [documentation](https://developers.google.com/closure/compiler/docs/api-tutorial3)) which we would require to work on if we choose to go with it. About third party contributed code, we can use `externs` feature of the compiler as a solution for calling out from compiled code to external code.
Also, the compiler did not work with webpack3, so right now I am focusing on upgrading Prebid.js to webpack4 first, and then can possibly come upto more findings on scope of using closure compiler.

sumit116:
> @sumit116
> What's the results from the `STANDARD` compression?

@mkendall07, with `Standard` compression, the minified bundle is nearly the same size as we have with `uglify-js`. The reason behind it closure compiler uses the same optimization techniques for standard mode as used by other tools. But with `advance` mode, it minifies the code to an extra level using techniques like - eliminating the dead code, removing the function declaration, etc.

sumit116:
@mkendall07 , closing this PR since we are not doing it.

PR #4100, opened 2019-08-21
emitted invalidAdUnit event for invalid ad units

pawanDangi:
## Type of change

- [ ] Bugfix
- [x] Feature
- [ ] New bidder adapter   
- [ ] Build related changes
- [ ] CI related changes
- [ ] Other

## Description of change

Emitted **invalidAdUnit** event when prebid detects the error in ad unit setup.
@jsnellbaker @jaiminpanchal27

### DIFF
--- a/src/prebid.js
+++ b/src/prebid.js
@@ -23,7 +23,7 @@ const events = require('./events');
 const { triggerUserSyncs } = userSync;
 
 /* private variables */
-const { ADD_AD_UNITS, BID_WON, REQUEST_BIDS, SET_TARGETING, AD_RENDER_FAILED } = CONSTANTS.EVENTS;
+const { ADD_AD_UNITS, BID_WON, REQUEST_BIDS, SET_TARGETING, AD_RENDER_FAILED, IN_VALID_AD_UNIT } = CONSTANTS.EVENTS;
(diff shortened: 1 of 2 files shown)


src/prebid.js
```diff
@@ -110,14 +112,17 @@ export const checkAdUnitSetup = hook('sync', function (adUnits) {
       const nativeObj = mediaTypes.native;
       if (nativeObj.image && nativeObj.image.sizes && !Array.isArray(nativeObj.image.sizes)) {
         utils.logError('Please use an array of sizes for native.image.sizes field.  Removing invalid mediaTypes.native.image.sizes property from request.');
+        events.emit(IN_VALID_AD_UNIT, 'Please use an array of sizes for native.image.sizes field.  Removing invalid mediaTypes.native.image.sizes property from request.');
```
jaiminpanchal27: Why do you pass message as second argument here ?

pawanDangi: It's an error or warning message. So, when we add invalid ad unit event listener will also get the error or warning message. For example -
`prebid.onEvent('invalidAdUnit', function(error) {
  console.log(error);
})`
In the above example, when we add a listener for the invalid add unit, will get the second argument of emitter function as an error in the above listener.
The second argument in emit is just a message. User will get this message when an invalid ad unit is emitted. So, this text helps the user to know about why the invalid ad unit invoked.

pawanDangi: I checked the test & spec folder. Also, I went through the test cases written inside that folder but I haven't got time to write it. Because before writing any test cases I have to understand the previous test cases because it will help me to write the new test cases in the proper place and way.

PR #4102, opened 2019-08-21
Add GDPR and UID module support to Undertone adapter

omerko:
## Type of change
- [x] Feature

## Description of change
- Added GDPR module support that sends the consent information to Undertone API and also to the user match endpoints
- Added support for external user id frameworks, sends all UIDs to the server
- Added the adapter version to be sent to the server

Added an indication that the adapter now supports GDPR in the docs repo:
- A link to a PR on the docs repo at https://github.com/prebid/prebid.github.io/pull/1443


modules/undertoneBidAdapter.js
```diff
@@ -51,15 +61,24 @@ export const spec = {
   },
   buildRequests: function(validBidRequests, bidderRequest) {
     const payload = {
-      'x-ut-hb-params': []
+      'x-ut-hb-params': [],
+      'commons': {
+        'adapterVersion': $$PREBID_GLOBAL$$.version,
```
mike-chowla: Referencing $$PREBID_GLOBAL$$ is not permitted.  For version, please use '$prebid.version$' instead

mike-chowla:
On the test failures, we've been getting spurious test failures, not sure why.  That failure is unrelated to your changes and everything passes when I ran in my local, so I'm not worried about it for this PR

modules/undertoneBidAdapter.js
omerko: Thanks you, changed

PR #4114, opened 2019-08-25
Browsi real time data module

omerBrowsi:
## Type of change

- [ ] Bugfix
- [ ] Feature

## Description of change

**Goal**

- Creating a generic pre-bid module that will be able to call different service providers to retrieve real-time data.
- light-weight and quick module that doesn't delay the auction, or adds the minimum delay to the auction only when necessary.

**Input**

Name | Description | Notes
-- | -- | --
Name | The name of the real-time data provider. | -
Params | An object with the configuration that each provider requires. |  
Timeout | Overall timeout for the module to get a response from the data provider. if the response was not received in that time span - real-time data should not be used. |  
Primary ad server only | Send the real-time data only to the primary ad server and not to any of the bidders. | True / False value (Boolean).In case this input is set to "True" (data should be sent only to the ad server) we can fetch the data in parallel with the auction in order to not delay the auction and reduce latency.

**JSON example:**
```
 pbjs.setConfig({
        "realTimeData": {
            "name": "browsi",
            "primary_only": false,
            "params": {
                "url": "testUrl.com",
                "siteKey": "testKey",
                "pubKey": "testPub",
                "keyName":"bv"
            }
        }
    });
```

**Flow**

flow

**Changes**

- browsiProvider.js - browsi sub module (provider)
- realTimeDataModule.js - generic real time module
- realTimeModule_spec.js - test for both module and sub module
- config.js - fix for unsubscribe config listener
- auction.js - new hook - after bid request and before bids back callback

asafs@gobrowsi.com 
omerdo@gobrowsi.com
eranyo@gobrowsi.com


mkendall07:
This pull request **introduces 1 alert** when merging 6121b4ed320ce8da712deeaa5fac75dc4d131e81 into e798c0cfc8598f70a86b7fcd6f5a3c14942662c9 - [view on LGTM.com](https://lgtm.com/projects/g/prebid/Prebid.js/rev/pr-dcb6c28d642b59337f8886cd3a44f2d9946edf2b)

**new alerts:**

* 1 for Useless assignment to local variable

---

---

bretg:
1. To align this feature with a similar feature in `userSync`, please replace `primary_only` and `timeout` options with `auctionDelay`. See https://github.com/prebid/prebid.github.io/pull/1459/files. If auctionDelay=0, then it should happen in parallel. 

```
 pbjs.setConfig({
        "realTimeData": {
            "name": "browsi",
            "auctionDelay": 0,
            "params": {
                ...
            }
        }
    });
```

2. Let's start out with a consistent structure that assumes there will be many RTD modules. Please move `browsiProvider.js` to a subdirectory (e.g. rtdModules). Add a README to that dir describing the interface code needs to have to be a real-time data provider.

3. Please open a docs PR covering the new module and your sub-module.

bretg:
Let's make sure we're on the same page here when it comes to building. The overall goal is to allow pubs to be surgical about which code is included in their package. To that end, the RTD modules should be built with something like:
```
gulp build --modules=rtdModule,browsiRtdProvider,...
```

bretg:
@omerBrowsi - couple of comments on the config:

1) I think the realtime data object needs to be an array so multiple real time data providers can be specified
2) cross-provider parameters should be supported (e.g. an auction delay that applies to all providers)
3) I think we should define 'auctionDelay' as a parameter instead of 'timeout'. That aligns this module with usersync. Maybe this means we can drop "primary_only"?

```
"realTimeData": {
     "auctionDelay": 100,
     "dataProviders": [{
            "name": "browsi",
            "params": {
                "url": "testUrl.com",
                "siteKey": "testKey",
                "pubKey": "testPub",
                "keyName":"bv"
            }
      }
}]
```

omerBrowsi:
Thank you for your input @bretg 
I'm working on those changes, I'll try pushing new version before the weekend

mkendall07:
This pull request **introduces 1 alert** when merging 3b85815b92f7320814f1d7d55d92bc4aebc94f93 into ebf4272e4d96119070e43d7a83da56b74fc3e0ba - [view on LGTM.com](https://lgtm.com/projects/g/prebid/Prebid.js/rev/pr-261ef15a3ab66b00112897499ef5dc4c42d4890a)

**new alerts:**

* 1 for Implicit operand conversion

---

---

omerBrowsi:
I pushed a new version with all the required fixes including support for multiple providers.

Regarding the rtdModule directory - 
I tried placing the browsiRtdProvider in it, but then it was left out of the build (the build only includes the index.js file in the directory), so I left it in the modules directory for now.
Is there a way I can place it inside the subdirectory and include it in the build?

I also pushed the documentations to the prebid.github.io repo

bretg:
Docs PR https://github.com/prebid/prebid.github.io/pull/1494

jsnellbaker:
@bretg I took a look through the logic and at the moment, the build process doesn't have the support needed to be able to pull the `browsiRtdProvider.js` file when it's sitting in a sub-folder.  

In the other cases we have of a file sitting in a sub-folder, they're being pulled into the build because either:
- the file is named `index.js` and there's a special set of logic in the `gulpHelpers` file that looks for these types of files
- the file is directly imported by the `index.js` in that sub-folder

All the other files that exist as a sub-module have to sit in the main `modules` folder in order to be included.  

There'd have to be more changes to the build process to add the ability to freely place module files under a sub-folder and have them be properly included.

On a separate yet related note, this PR should add an entry to the `modules/.submodules.json` file to add the relationship between the `browsiRtdProvider` and `realTimeData`.  It should look like the following:
```
... // other module data above here
],
"realTimeData": [
  "browsiRtdProvider"
]
```

This would allow the rtdModule (ie the `rtdModule/index.js` file) to be automatically included in the build when someone includes the `browsiRtdProvider.js` file.

omerBrowsi:
> What is slotBPlacementId?
>Not every real time data module is going to be returning the same thing. This needs to be very generic. Sounds like the interface needs to support both key-value pairs like you need, but also passing values through the bidrequest object - like setting a price floor on each adunit.
>
>If needed, we can get this version of RTD module out to support your use case, and then refactor when the Price Floor sub-module is better defined.
> 

@bretg 
slotPlacementId is actually ad unit code, I will change the docs.
The data is added to the bid request according to the code (each code get its own data).

For example: 
`
ad1:{
floorPrice: 0.5
},
ad2:{
floorPrice:0.6
}
`

Bid with the ad unit code of 'ad1' will receive the data floorPrice with a value of 0.5,
Bid with the ad unit code of 'ad2' will receive the data floorPrice with a value of 0.6,
Bid with the ad unit code of 'ad3' will not receive any data (no matching ad unit code in the JSON)

Do you want the option to add data to all of the bids (regardless of the ad unit code)?

bretg:
Ok, I get your use case better now. But the core module needs to be able to support different types of sub-modules. 

1) sub-modules that add targeting per-adunit
2) sub-modules that delay the auction and add parameters to the bid request object
3) sub-modules that inject data into other places in the environment that's later used by a hook

As an example, we're soon going to post an intent-to-implement on a 'price-floor' sub-module. There's no ad server targeting at all in that use case. Instead, pubs will have an option about whether to delay the auction or not, but it will A) write to the bidrequest and B) write results to a local variable that will then later be used by a post-auction hook that will enforce price floors.

So, I'd prefer that the core RTD module makes no assumptions about what the sub-module is going to do. Instead, the core module should be a framework. Each sub-module should be built around a generic interface like "initiate request" and "process response".

I haven't dug into the code here - what drew my attention was the documentation. It's ok if the first iteration of RTD isn't perfect, but the closer we can get now, the less refactoring we'll need to do later to support sub-modules of other types.

bretg:
The proposal for the Price Floor sub-module has been published -- https://github.com/prebid/Prebid.js/issues/4273

Configuration-wise, I think things are aligned. Here's an example of how the page might configure both sub-modules:

```
realTimeData: {
        auctionDelay: 100,
        providers: [{
                "name": "browsi",
                "params": {
                   ...
                }
            },{
            "name": "floor",
            "params": {
                ...
            }
       }]
    }
```

Based on the draft documentation in https://github.com/prebid/prebid.github.io/pull/1494 describes the RTD/sub-module interface in a way that seems tied to the ad-server-targeting use case. Here's what it says:

- For each auction, the module will call the provider (sub-module) to fetch the real-time data per placement.
- The provider will respond with the real-time data object per the params and configurations.
- The module will take the real-time data and will (depending on the "auctionDelay").
	- Push it as a key to all (or some) of the bidders in the auction and push it as a key to GAM.
	- Push it only to GAM as a key.
- The provider will have the time set in the auctionDelay to fetch the real-time data.
- In case there is no response before the timeout, the module will ignore and will not push any value either to GAM or to any of the bidders.


I think this sounds like too much functionality in the core part of the RTD module. Given the different use cases on the table for RTD, I think the core part of the module should be very limited:

- it establishes a convention for publisher config
- it enforces the optional auctiondelay config
- it coordinates interaction between the PBJS core and the RTD sub-modules

I haven't looked at what the internal interface actually is, but at a high level, I think we should consider heading towards something like:

1) rtd_init()-submodule - in case the sub-module needs to do some setup. If it doesn't need adunit data, it can phone home immediately as needed.
2) rtd_auction(adunits) - tells the sub-module that an auction is pending - allows it to look for locally cached data or kick off a separate fetch as needed.
3) rtd_getTargetingData(adunit) - gives the sub-module a chance to supply targeting values for each or all adunits.
4) rtd_updateBidRequest(bidRequest) - gives the sub-module a chance to add or modify fields to a bidRequest object.
5) rtd_updateBidResponse(bidResponse) - gives the sub-module a chance to add or modify fields in the bidResponse object
6) each sub-module needs to somehow register which PBJS-core activities it needs to modify: pre-auction, post-auction, get-targeting

Possible flow:

1) SetConfig recognizes `realTimeData` and calls the RTD core module
2) RTD-core registers any callbacks needed and calls rtd_init() for each sub-module
3) PBJS-core calls RTD-core on `requestBids`.
4) RTD-core calls rtd_auction() for each sub-module
5) if the auction is to be delayed, RTD-core sets a timer
6) When one of these scenarios occurs, RTD-core the auction ready to happen
    a) auctionDelay set to 0
    b) all RTD fetches are back
    c) timer pops
7) Once auction is ready to happen, RTD-core tells PBJS-core it's ok to go ahead
8) Registered callbacks rtd_getTargetingData, rtd_updateBidRequest, rtd_updateBidResponse are called when needed.

@mkendall07 - I suggest this path-forward on the RTD module:
1) approve Browsi's v1 RTD module ASAP
2) assign an architect to really look into the design here
3) evolve the RTD module and sub-modules as needed

bretg:
@omerBrowsi - Prebid.js has determined that straightup promises will not be allowed in core modules due to their javascript execution patterns. We may allow the 'fetch' pattern, which will require a polyfill. Adding @mkendall07 and @snapwich for technical details.

snapwich:
More info on that discovery process here: https://github.com/prebid/Prebid.js/pull/4177#issuecomment-539752747

It looks like you're using Promises for Ajax requests, I think we're going to see about supporting window.fetch through a polyfill.

omerBrowsi:
@bretg @snapwich - We are not using Promises for Ajax (we use XMLHttpRequest).
I changed the promises to callbacks and configures the submodules.json as @jsnellbaker suggested

bretg:
We discussed this last week. We're not going to try to generalize the sub-module interface for now. @robertrmartinez is next up to review.

robertrmartinez:
@omerBrowsi Is there some temporary params you can send me (doesn't have to be over comment) which will solicit actual real returns for this module?

I would like to get a test page going and do some real End to End verifications if possible.

modules/rtdModule/index.js
```diff
@@ -0,0 +1,208 @@
...
+          // duplicate key - merge values
+          const dp = obj[key];
+          for (let dk in dp) {
+            if (dp.hasOwnProperty(dk)) merged[key][dk] = dp[dk];
+          }
+        }
+      }
+    }
+    return merged;
+  }, {});
```
robertrmartinez: What is the reasoning for not just doing:
```javascript
  return arr.reduce((merged, obj) => {
    return Object.assign(merged, obj);
  }, {});
```

Am I missing a use case where this would not do what is intended by the `deepMerge`?

modules/rtdModule/index.js
```diff
@@ -0,0 +1,208 @@
...
+
+/**
+ * @param {AdUnit[]} adUnits
+ *  @param {Object} data - key values to set
+ */
+function addIdDataToAdUnitBids(adUnits, data) {
+  adUnits.forEach(adUnit => {
+    adUnit.bids = adUnit.bids.map(bid => {
+      const rd = data[adUnit.code] || {};
+      return Object.assign(bid, rd);
```
robertrmartinez: Should we be placing these inside some object on the bid rather than the top level of the bid?

modules/rtdModule/index.js
omerBrowsi: Object.assign is not dealing well with duplicate keys.
See an example here - 
https://codepen.io/omerBrowsi/pen/oNNoLgp?editors=1112

modules/rtdModule/provider.md
```diff
@@ -0,0 +1,27 @@
...
+```
+
+2. Function that returns the real time data according to the following structure:
+```
+{
+  "adUnitCode":{
+      "key":"value",
+      "key2":"value"
+  },
+  "adUnirCode2":{
```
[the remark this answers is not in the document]
omerBrowsi: adUnitCode :)
Will be fixed on the next commit.

bretg:
Good to see the progress here. Towards getting this released as soon as feasible, I'm planning to update the documentation so this PR releases just the "Browsi Real Time Data" module -- a stand-alone thing at this point. With later releases we'll generalize the infrastructure you've started.

modules/rtdModule/index.js
omerBrowsi: Yes, I'll place it inside "realTimeData" object.
Will be fixed on the next commit.

omerBrowsi:
@robertrmartinez 
Thank you for your input.
I will commit all relevant fixes this Sunday.

I sent you an email with a test page

modules/rtdModule/index.js
robertrmartinez: Ahh I see what you mean on the use case now.

omerBrowsi:
@bretg @robertrmartinez

I just pushed a new version with fixes to all of the requested changes.

To simplify the process, I'm adding the latest test page below where you can review the model running live with dummy data coming from our server.
https://browsi.bitbucket.io/rtd_demo_page.html

We are aiming on our end to push this live in the next version deployment this week - crucial for us as we have few clients that are waiting to use this.
Your help is much appreciated!

modules/browsiRtdProvider.js
```diff
@@ -0,0 +1,253 @@
...
+  return slotAdUnits.some(a => whitelist.indexOf(a) !== -1);
+}
+
+/**
+ * XMLHttpRequest to get data form browsi server
+ * @param {string} url server url with query params
+ */
+function getPredictionsFromServer(url) {
+  const xmlhttp = new XMLHttpRequest();
+  xmlhttp.onreadystatechange = function() {
```
robertrmartinez: I think we may want to use the [Prebid Ajax](https://github.com/prebid/Prebid.js/blob/master/src/ajax.js) 

We do not really let modules / adapters handle their our requests usually. @jaiminpanchal27 does this seem right?

robertrmartinez: https://github.com/prebid/Prebid.js/blob/master/modules/userId/unifiedIdSystem.js#L42 example of one of the user Id modules utilizing it

omerBrowsi:
@robertrmartinez @bretg 

Just pushed fixes to use Prebid Ajax
The only thing left is a fix for the module tests, I will push it soon.

modules/browsiRtdProvider.js
omerBrowsi: Fixed

modules/rtdModule/index.js
```diff
@@ -70,12 +70,15 @@ export function attachRealTimeDataProvider(submodule) {
...
 export function init(config) {
   const confListener = config.getConfig(MODULE_NAME, ({realTimeData}) => {
-    if (!realTimeData.dataProviders || typeof (realTimeData.auctionDelay) == 'undefined') {
+    if (!realTimeData.dataProviders) {
       utils.logError('missing parameters for real time module');
       return;
     }
     confListener(); // unsubscribe config listener
     _moduleConfig = realTimeData;
+    if (typeof (_moduleConfig.auctionDelay) == 'undefined') {
```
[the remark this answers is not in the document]
omerBrowsi: fixed

omerBrowsi:
Pushed a fix for the module test as well

bretg:
merged. (!)

Issue #3899, opened 2019-06-12
Deal specific line items don't work


benjaminclot:
## Type of issue
Deals don't work as advertised in the [docs](http://prebid.org/adops/deals.html).

## Description
When creating a deal specific line item, that LI wins but the [Universal Creative](https://github.com/prebid/prebid-universal-creative) code doesn't support it and only uses `hb_adid` which may not be the deal's ad ID.
My guess is that it's not possible to know why GAM elected a LI to win with macros, so either  the creative template or the UC code needs to be changed? Also, the deal's ad ID, if it didn't win, doesn't seem to be sent in the key-values with `enableSendAllBids: false`.

## Other information
This has already been reported (and ignored 😞) in #3266 back in november.

bretg:
Sorry for the silence on this issue @benjaminclot. The answer in #3266 is close to correct, but it's not quite as bad as it sounds:

> this means we'll need to create a unique line item and unique creative for every deal, which seems long-winded.

Really it's just a unique line item and creative per bidder, not for every deal. e.g. targeting could be to hb_deal_BIDDERCODE in (1111,2222,3333).

I've updated the deals page on prebid.org -- http://prebid.org/adops/deals.html#step-4-attach-creatives-to-line-items

Basically, the line item targeting has to match what the creative sends back to Prebid.js -- so if the line item is targeted to hb_deal_BIDDERCODE, then the creative needs to send back hb_adid_BIDDERCODE. If you see a way around this, happy to take suggestions.

I see that the last screenshot on the deals page needs to be changed -- not recommended to target different BIDDERCODEs in the same line item because of the adid issue.

benjaminclot:
@bretg as much as I do agree with your statement and solution, this doesn't work with `sendAllBids: false` as the adid for the deal is not being sent as far as I can tell.

bretg:
Not quite following @benjaminclot.

> Also, the deal's ad ID, if it didn't win, doesn't seem to be sent in the key-values with enableSendAllBids: false.

1) enableSendAllBids was built for deals. `hb_adid` is always for the bid that had the highest CPM. If the deal wasn't the highest bid, it will still be sent as hb_deal_BIDDER and hb_adid_BIDDER.
2) line items in the ad server should target to hb_deal_BIDDER. The creatives would refer to hb_adid_BIDDER
3) I verified the behavior here works as expected in the enableSendAllBids: false scenario. hb_deal and hb_adid are sent when the deal wins.

[...]

PR #4136, opened 2019-08-30
New option to Include deal KVPs when enableSendAllBids === false

robertrmartinez:
## Type of change

- [X] Bugfix

## Description of change
Fix for #3899 

New option in `targetingControls` to allow sending through bids which have dealId's even when enableSendAllBids is false

```
pbjs.setConfig({
   enableSendAllBids: false,
   targetingControls: {
      alwaysIncludeDeals: true
   }
});
```

`alwaysIncludeDeals` must be explicitly set = true to utilize.

This code only happens when `enableSendAllBids` is `false` (as if true should have been added anyways!)

### DIFF
--- a/test/spec/unit/core/targeting_spec.js
+++ b/test/spec/unit/core/targeting_spec.js
@@ -330,6 +330,73 @@ describe('targeting tests', function () {
       expect(logErrorStub.calledOnce).to.be.true;
     });
 
+    describe('targetingControls.alwaysIncludeDeals', function () {
+      let bid4;
+
+      let targetingWithAdnxsDeal = {
+        hb_deal_rubicon: '1234',
+        hb_deal: '1234',
+        hb_pb: '0.53',
+        hb_adid: '148018fe5e',
+        hb_bidder: 'rubicon',
+        foobar: '300x250',
+        hb_deal_appnexus: '1234',
+        hb_pb_appnexus: '0.53',
+        hb_adid_appnexus: '148018fe5e',
+        hb_bidder_appnexus: 'appnexus',
+      };
+      let targetingWithoutAdnxsDeal = {
+        hb_deal_rubicon: '1234',
+        hb_deal: '1234',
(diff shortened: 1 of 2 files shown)


test/spec/unit/core/targeting_spec.js
```diff
@@ -330,6 +330,73 @@ describe('targeting tests', function () {
...
+
+      let targetingWithAdnxsDeal = {
+        hb_deal_rubicon: '1234',
+        hb_deal: '1234',
+        hb_pb: '0.53',
+        hb_adid: '148018fe5e',
+        hb_bidder: 'rubicon',
+        foobar: '300x250',
+        hb_deal_appnexus: '1234',
+        hb_pb_appnexus: '0.53',
```
bretg: why isn't this 0.1 to match bid4.cpm?  Also, I would find it less confusing to use different deal IDs for RP and APN

test/spec/unit/core/targeting_spec.js
```diff
@@ -330,6 +330,73 @@ describe('targeting tests', function () {
...
+        hb_adid_appnexus: '148018fe5e',
+        hb_bidder_appnexus: 'appnexus',
+      };
+      let targetingWithoutAdnxsDeal = {
```
bretg: I'd also like to see a scenario where the winning bid isn't a deal.

test/spec/unit/core/targeting_spec.js
robertrmartinez: Good catch. Let me fix

test/spec/unit/core/targeting_spec.js
robertrmartinez: Ok Will do!

robertrmartinez:
@bretg Updated the tests

benjaminclot:
That's great! Now all that's missing is support of this by the [Universal Creative](https://github.com/prebid/prebid-universal-creative)...

PR #4137, opened 2019-08-30
Added SupplyChain Object support and an onTimeout Callback

telariaEng:
## Type of change

- [x] Feature

## Description of change

- Enhanced our adapter to handle a SupplyChain Object. Please refer to [SupplyChain for Non-OpenRTB Requests](https://github.com/InteractiveAdvertisingBureau/openrtb/blob/master/supplychainobject.md#supplychain-for-non-openrtb-requests) for more information
- Fires an event pixel onTimeout

- contact email: github@telaria.com

- [Docs PR](https://github.com/prebid/prebid.github.io/pull/1458)


mike-chowla:
Based on feedback from @bretg , I reverted this change as it should use the schain module.  I thought it was but in the code it's a pretty subtle difference which I overlooked.  Apologies for the confusion.

Here's an example of how to use the global schain object:
https://github.com/prebid/Prebid.js/blob/fe0c9e818b25eafda6d8f6a2e1a786c940d2f534/modules/pubmaticBidAdapter.js#L877

mike-chowla:
Please a submit a new PR once you've made the update

PR #4140, opened 2019-09-02
Module to pass User Ids to DFP

pm-harshad-mane:
## Type of change
- [X] Feature

## Description of change
- This module works with userId module.
- This module is used to pass userIds to GAM (formerly DFP) in targeting so that user ids can be used to pass in Google Exchange Bidding or can be used for targeting in GAM.

## Sample config
```
pbjs.setConfig({

	// your existing userIds config

	usersync: {
		userIds: [{...}, ...]
	},

	// new shareUserIds config

	userIdTargeting: {
		"GAM": true,
		"GAM_KEYS": {
			"tdid": "TTD_ID" // send tdid as TTD_ID
		}
	}
});
```

## Config options
- GAM: is required to be set to true if a publisher wants to send UserIds as targeting in GAM call. This module uses ``` googletag.pubads().setTargeting('key-name', ['value']) ``` API to set GAM targeting.
- GAM_KEYS: is an optional config object to be used with ``` "GAM": true ```. If not passed then all UserIds are passed with key-name used in UserIds object.
If a publisher wants to pass ```UserId.tdid``` as TTD_ID in targeting then set  ``` GAM_KEYS: { "tdid": "TTD_ID" }```
If a publisher does not wants to pass ```UserId.tdid``` but wants to pass other Ids in UserId tthen set ``` GAM_KEYS: { "tdid": "" }```

## Including this module in Prebid
``` $ gulp build --modules=userId,userIdTargeting,pubmaticBidAdapter ```

## Notes
- We can add support for other external systems like GAM in future
- We have not added support for A9/APSTag as it is called in parallel with Prebid. This module executes when ```pbjs.requestBids``` is called, in practice call to A9 is expected to execute in parallel to Prebid thus we have not covered A9 here. For sending Uids in A9 one will need to set those in params key in the object passed to ```apstag.init```, ```pbjs.getUserIds``` can be used for the same.

### DIFF
--- a/modules/userIdTargeting.js
+++ b/modules/userIdTargeting.js
@@ -2,11 +2,11 @@ import {config} from '../src/config';
 import {getGlobal} from '../src/prebidGlobal';
 import { isStr, isPlainObject, isBoolean, isFn, hasOwn, logInfo } from '../src/utils';
 
-const MODULE_NAME = 'shareUserIds';
+const MODULE_NAME = 'userIdTargeting';
 const GAM = 'GAM';
 const GAM_KEYS_CONFIG = 'GAM_KEYS';
 
-export function shareUserIds(userIds, config) {
+export function userIdTargeting(userIds, config) {
   if (!isPlainObject(config)) {
     logInfo(MODULE_NAME + ': Invalid config found, not sharing userIds externally.');
     return;
@@ -44,7 +44,7 @@ export function shareUserIds(userIds, config) {
 export function init(config) {
   getGlobal().requestBids.before(function(fn, reqBidsConfigObj) {
     // using setTimeout to avoid delay
-    setTimeout(shareUserIds, 0, (getGlobal()).getUserIds(), config.getConfig(MODULE_NAME));
+    setTimeout(userIdTargeting, 0, (getGlobal()).getUserIds(), config.getConfig(MODULE_NAME));
     // calling fn allows prebid to continue processing
     return fn.call(this, reqBidsConfigObj);
   }, 40);
--- a/modules/userIdTargeting.js
+++ b/modules/userIdTargeting.js
@@ -1,5 +1,7 @@
 import {config} from '../src/config';
 import {getGlobal} from '../src/prebidGlobal';
+import CONSTANTS from '../src/constants.json';
+import events from '../src/events';
 import { isStr, isPlainObject, isBoolean, isFn, hasOwn, logInfo } from '../src/utils';
 
 const MODULE_NAME = 'userIdTargeting';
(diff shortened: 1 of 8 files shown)

### CODE REVIEW

**Comment by pm-harshad-mane** [comment]
Hello @idettman , @jaiminpanchal27 

**Comment by pm-harshad-mane** [comment]
Hello @bretg ,

Thank you for the review.
I have renamed DFP to GAM, also renamed the module to userIdTargeting.
I will create a separate PR to update http://prebid.org/dev-docs/modules/userId.html

**Comment by pm-harshad-mane** [comment]
Documentation PR https://github.com/prebid/prebid.github.io/pull/1462

**Comment by pm-harshad-mane** [comment]
Hello @idettman , @jaiminpanchal27 , @mkendall07 ,

**Comment by pm-harshad-mane** [comment]
Hello @idettman , @jaiminpanchal27 , @mkendall07 ,

File: modules/userIdTargeting.js [hunk 1]
```diff
@@ -0,0 +1,62 @@
...
+          (hasOwn(PUB_GAM_KEYS, key) ? PUB_GAM_KEYS[key] : key),
+          [ uidStr ]
+        );
+      }
+    }
+  });
+}
+
+export function init(config) {
+  getGlobal().requestBids.before(function(fn, reqBidsConfigObj) {
```
jaiminpanchal27: I think we should not add this as a hook to requestBids. This module when included will delay all calls to requestBids. 
The feature here if I understand correctly needs to set some targeting key value pair before the ad server call. So why are we doing this every time before requestBids. This can be done just before ad server call.

pm-harshad-mane: I tried if "after" hook is present but it is not. thus I added setTimeout(func(){}, 0) inside the function to avoid delay. 
Can you please any other appropriate hook? happy to move to a better approach :)

pm-harshad-mane: @jaiminpanchal27 or shall I use ```pbjs.onEvent('auctionEnd', function(){ /* our code here */ })```

mkendall07: auctionEnd is probably a better option.

pm-harshad-mane: Thank you @mkendall07 , I made the said changes.

modules/userIdTargeting.js
```diff
@@ -0,0 +1,59 @@
...
+          (hasOwn(PUB_GAM_KEYS, key) ? PUB_GAM_KEYS[key] : key),
+          [ uidStr ]
+        );
+      }
+    }
+  });
+}
+
+export function init(config) {
+  $$PREBID_GLOBAL$$.onEvent('auctionEnd', function() {
```
jaiminpanchal27: @pm-harshad-mane Can you use `events.on` api here instead of $$PREBID_GLOBAL$$. Sample use: https://github.com/prebid/Prebid.js/blob/master/modules/userId/index.js#L252

pm-harshad-mane: Sure, @jaiminpanchal27 ,  I have pushed the suggested change

pm-harshad-mane:
Hello @idettman ,

pm-harshad-mane:
Thank you @jaiminpanchal27 , @bretg , @idettman and @mkendall07

PR #4151, opened 2019-09-03
appnexusBidAdapter - fix outstream bid property

### LINKED ISSUE #3364: EMXDigital outstream does not set VAST url
## Type of issue
Bug

## Description

emxdigital alias using appnexus bid adapter configured for outstream with a renderer.

Example response from network with rtb.video.content VAST xml.
screen shot 2018-11-30 at 10 10 36 am

Prebid bidResponse object, does not map the VAST xml from the network response above into any parameter. I'd expect to see it under `ad` or `vastUrl`.
screen shot 2018-11-30 at 10 08 20 am

The server response from http://ib.adnxs.com/ut/v3/prebid includes `rtb.video.content` instead of `rtb.video.asset_url` which is what the adapter appears to be looking for. https://github.com/prebid/Prebid.js/blob/master/modules/appnexusBidAdapter.js#L263

There is no VAST XML to pass to our renderer.

Should the server return something different? Or should the appnexus adapter be update to also look for rtb.video.content? Or have I missed something else?

## Platform details
Prebid.js 1.26

jsnellbaker:
## Type of change
- [x] Bugfix

## Description of change
Fixes #3364

At some point, the appnexus outstream bid response changed from providing a vastUrl (in the `asset_url` property) to providing the raw vastXml (in the `content`) property.

As highlighted within the above issue, the `appnexusBidAdapter` didn't make the corresponding update to set the ideal value in the prebid bidResponse object - so nothing was being set in the bid's normal vast fields.  

This PR corrects this issue by setting the proper vast field in the bid and reading from the correct source in the raw bid response.

Side Note - it seems the AppNexus Outstream Renderer doesn't explicitly rely on the `bid.vastUrl` or `bid.vastXml` fields to render the outstream bid.  

The AN Renderer seems to read the vast XML from the unique `bid.adResponse` field (which is made in the adapter code); the field is itself a copy of the raw bid response object.  This is why despite this issue, the outstream bids were still rendering.

### DIFF
--- a/modules/appnexusBidAdapter.js
+++ b/modules/appnexusBidAdapter.js
@@ -482,7 +482,7 @@ function newBid(serverBid, rtbBid, bidderRequest) {
     Object.assign(bid, {
       width: rtbBid.rtb.video.player_width,
       height: rtbBid.rtb.video.player_height,
-      vastUrl: rtbBid.rtb.video.asset_url,
+      vastXml: rtbBid.rtb.video.content,
       vastImpUrl: rtbBid.notify_url,
       ttl: 3600
     });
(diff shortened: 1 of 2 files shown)


jsnellbaker:
Please wait before merging.

PR #4154, opened 2019-09-04
fix several issues in appnexus video bids

### LINKED ISSUE #3364: EMXDigital outstream does not set VAST url

### LINKED ISSUE #4025: Appnexus outstream should not check if renderer is defined server side
## Type of issue
Bug

## Description
When creating a placement through the API on appnexus, no renderer is associated to this placement. This is the field ad_types on the placement.
When using this placement on a bid request with an outstream media type, no renderer is defined in the server bid response. Appnexus adapter actually checks only if a renderer has been defined on server side to add the relevant information in the bid response sent to prebid.

So, if no renderer has been defined server side but one is defined on client side, the bid response does not contains the needed "adResponse" field.

This test is located in the adapter here: https://github.com/prebid/Prebid.js/blob/master/modules/appnexusBidAdapter.js#L403-L415

In my opinion, the check should be completed, checking the adunit too.

Thanks for your help.

PS: for those who are stuck with the API, here is a valid ad_types field:

```
[{"id":2,"name":"video","display_name":"Video","renderer":{"id":2,"display_name":"AppNexus Outstream Video Renderer","config":"{\"playOnMouseover\":false,\"audioOnMouseover\":true,\"autoInitialSize\":true,\"nonViewableBehavior\":\"mute\",\"aspectRatio\":\"16:9\",\"playerTechnology\":[\"html5\",\"flash\"],\"playVideoVisibleThreshold\":50,\"expandTime\":1000,\"fixedSizePlayer\":false,\"delayStartUntilNotified\":false,\"showMute\":true,\"showVolume\":false,\"showPlayToggle\":false,\"showProgressBar\":true,\"disableTopBar\":false,\"allowFullscreen\":false,\"disableCollapse\":{\"enabled\":true,\"replay\":true},\"adText\":\"Publicit\u00e9\",\"controlBarPosition\":\"over\",\"vpaidTimeout\":15000,\"terminateUnresponsiveVPAIDCreative\":false,\"delayExpandUntilVPAIDInit\":true,\"delayExpandUntilVPAIDImpression\":false,\"enableInlineVideoForIos\":true,\"preloadInlineAudioForIos\":false,\"waterfallTimeout\":15000,\"waterfallSteps\":-1,\"playerSkin\":{\"topBarColor\":\"#e9edf0\",\"topBarTextColor\":\"#939da3\",\"dividerColor\":\"#cccccc\"},\"skippable\":{\"enabled\":false}}"}}]
```

jsnellbaker:
## Type of change

- [x] Bugfixes

## Description of change
This PR addresses several issues that were present in how video bids were generated in the `appnexusBidAdapter` (and in cases when any `appnexus` alias was used, as reported under #3364 ).

Below is a short summary of the issues:
1. Reported in #3364.  

For outstream bids, the appnexus bidder did not properly populate the typical `bid.vastXml` or `bid.vastUrl` fields.  

The bidder was coded to only look for the `asset_url` value in the raw bid response, but this value is only used by `instream` or `adpod` type bids.  For outstream bids, the ad server returns vast XML code (via the `rtbBid.rtb.video.content` property) and should be set to the `bid.vastXml` field.  

Fix: Added logic to look at the proper fields given the `context` used in the video bid object.

2. Reported in #4025.  

When publishers have the `renderer` defined in an adUnit, they may have cases where the placement in the appnexus adserver is not automatically configured to return a renderer asset.  

Due to some logic used in the bidder (i.e. it assumed it had an outstream bid only if the `renderer_url` was present in the raw bid response from the AN adserver), this prevented some extra video bid information from being attached to the bid object (ie `bid.adResponse`).

This lack of data in the bid in turn hindered the bid from rendering with the adUnit's renderer when the above use-case was used.

Fix: Updated the logic we use to consider what is an outstream bid and moved some logic around so proper values are set for outstream bids.

3. Observed internally.  

When looking for the `renderer_options` for outstream bids, it was hard-coded to always look at the first bid in the `bidderRequest` object.  When there are multiple bids of the same bidder, the desired video bid that should be referenced isn't always going to be first in the list.

Fix: Updated the logic to find the correct bid in the `bidderRequest` object and use that to read the `renderer_options` if they're defined.

## Other information
Fixes #3364 
Fixes #4025

### DIFF
--- a/modules/appnexusBidAdapter.js
+++ b/modules/appnexusBidAdapter.js
@@ -5,6 +5,7 @@ import { registerBidder, getIabSubCategory } from '../src/adapters/bidderFactory
 import { BANNER, NATIVE, VIDEO, ADPOD } from '../src/mediaTypes';
 import find from 'core-js/library/fn/array/find';
 import includes from 'core-js/library/fn/array/includes';
+import { OUTSTREAM, INSTREAM } from '../src/video';
 
 const BIDDER_CODE = 'appnexus';
 const URL = '//ib.adnxs.com/ut/v3/prebid';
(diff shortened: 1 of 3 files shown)


xavierleune:
:+1: thanks @jsnellbaker

jsnellbaker:
Hi @mkendall07 I checked the longform test pages and they were working fine.

PR #4155, opened 2019-09-04
Rubicon Video COPPA fix

umbrella-artem-seryak:
## Type of change

- [x] Bugfix
- [ ] Feature
- [ ] New bidder adapter   
- [ ] Build related changes
- [ ] CI related changes
- [ ] Other

## Description of change
It looks like there is a bug in a Rubicon adapter - when setting coppa flag, the current implementation tries to do that on a non-defined `request` object, instead of `data`, which seems to be the right place for it.
Feels like the problem line was just copy-pasted from https://github.com/rubicon-project/Prebid.js/blob/4f16865157caca46b7e024d81202419517835451/modules/prebidServerBidAdapter/index.js#L575

### DIFF
--- a/modules/rubiconBidAdapter.js
+++ b/modules/rubiconBidAdapter.js
@@ -237,7 +237,7 @@ export const spec = {
       }
 
       if (config.getConfig('coppa') === true) {
-        utils.deepSetValue(request, 'regs.coppa', 1);
+        utils.deepSetValue(data, 'regs.coppa', 1);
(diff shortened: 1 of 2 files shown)


bretg:
Good catch  - thanks @umbrella-artem-seryak 

@msm0504 - please add a unit test here.

msm0504:
@bretg I've added a unit test and pushed the change.

PR #4161, opened 2019-09-05
Outcon bid adapter.

TinchoF:
## Type of change

- [ ] Bugfix
- [ ] Feature
- [x] New bidder adapter   
- [ ] Build related changes
- [ ] CI related changes
- [ ] Other

## Description of change

```
{
  bidder: '<bidder name>',
  params: {
    // ...
  }
}
```


bretg:
Docs PR https://github.com/prebid/prebid.github.io/pull/1461

jsnellbaker:
@TinchoF Can you please put together some unit tests for this new adapter?

The following link can provide some direction if needed (or feel free to look at the unit test files for other adapters as a reference):
http://prebid.org/dev-docs/bidder-adaptor.html#adding-unit-tests

modules/outconBidAdapter.js
```diff
@@ -0,0 +1,41 @@
+import {registerBidder} from '../src/adapters/bidderFactory';
+import {config} from '../src/config';
+const BIDDER_CODE = 'outcom';
+export const spec = {
+  code: BIDDER_CODE,
+  aliases: ['outcom'],
```
jsnellbaker: If you have the same name as the value for your BIDDER_CODE, this `alias` property doesn't need to be defined.  This should only be used when you have a different name.

modules/outconBidAdapter.js
```diff
@@ -0,0 +1,41 @@
...
+  isBidRequestValid: function(bid) {
+    return !!(bid.params.pod || (bid.params.internalId && bid.params.publisher));
+  },
+
+  buildRequests: function(validBidRequests) {
+    let par = '';
+    if (validBidRequests.params.pod != undefined) par = 'get?pod=' + validBidRequests.params.pod;
```
jsnellbaker: The `validBidRequests` is an array of potential bid requests to your adapter (in case the publisher has you listed in multiple adUnit objects on their page).

Can you please update the internal logic in this function to access the array?  It's currently throwing undefined errors.

modules/outconAdapter.md
```diff
@@ -0,0 +1,21 @@
...
+
+Module that connects to Outcon demand sources
+
+# Test Parameters
+```
+    var adUnits = [
+        {
+            pod: "5beeb24a306ea47660632043",
+            internalId: "12345678",
+            publisher: "5beeb1fd306ea4779e464532"
```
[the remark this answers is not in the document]
TinchoF: Regarding the ad unit format, our clients show programmatic advertising in PODs (points of display) that we have in our database, upon receiving the POD id or the publisher ID + Internal ID, we access all the data to deliver the ad appropriately (resolution, size, mimes, audience, etc.)

modules/outconAdapter.md
```diff
@@ -0,0 +1,24 @@
...
+# Test Parameters
+```
+    var adUnits = [
+        {
+        bidder: 'outcom',
+        params: {
+          internalId: '12345678',
+          publisher: '5d5d66f2306ea4114a37c7c2',
+          demo: true
+        }
```
jsnellbaker: Two points of follow-up feedback:
- there is a missing `}` for this overall object block.
- the `bidder` value is using `outcom` instead of `outcon`

Can you update these items so it's properly formatted and consistent with your adapter?

modules/outconBidAdapter.js
```diff
@@ -0,0 +1,40 @@
...
+  },
+  buildRequests: function(validBidRequests) {
+    for (let i = 0; i < validBidRequests.length; i++) {
+      let par = '';
+      if (validBidRequests[i].params.pod != undefined) par = 'get?pod=' + validBidRequests[i].params.pod;
+      else par = 'get?internalId=' + validBidRequests[i].params.internalId + '&publisher=' + validBidRequests[i].params.publisher;
+      if (validBidRequests[i].params.demo == true) par = par + '&demo=true';
+      return {
+        method: 'GET',
+        url: 'http://test.outcondigital.com:8048/ad/' + par,
```
jsnellbaker: Is this URL just temporary or the permanent location?  If there are different domains that should be used if someone is doing a demo (or test) vs someone calling the production endpoint, you could control that logic through the adapter code and the bid params.  Other adapters have done something similar.

test/spec/modules/outconBidAdapter_spec.js
```diff
@@ -0,0 +1,104 @@
+import { expect } from 'chai';
+import { spec } from '../../../modules/outconBidAdapter';
+
+describe('outconBidAdapter', function () {
+  describe('bidRequestValidity', function () {
+    it('Check the bidRequest with pod param', function () {
+      expect(spec.isBidRequestValid({
+        bidder: 'outcom',
```
jsnellbaker: To ensure consistency - can you please fix these `bidder` values through your tests to use `outcon`?

modules/outconBidAdapter.js
TinchoF: Following your comment we decided to enable the other URLs, and we changed the parameters to specify the environment (env=test, env=api, env=stg).
The URLs are:
test.outcondigital.com
api.outcondigital.com
stg.outcondigital.com 

I tested the URL you said above and it works fine, maybe it was at a time of server maintenance, I get the following response:
{"impressionId":"5d812767bb00be0f847668df","ad":"5d42ea05306ea4689c2c8118","status":"ok","cpm":0.5,"cur":"USD","exp":300,"creatives":[{"url":"http://test.outcondigital.com/uploads/5d42ea05306ea4689c2c8112/outcon3.jpg","size":1,"width":1280,"height":720,"codec":"image/jpg"}],"type":"banner","duration":20,"campaign":"5d7fe03c306ea4018e095db2","trackingURL":"http://test.outcondigital.com:8048/ad/track?track=5d812767bb00be0f847668df"}

jsnellbaker: I tried to make the request again and it succeeded this time (when I was off my company network).  However I encountered the following CORS error which prevented the request from being accepted (even though the response payload was there this time).

Can you please take a look?  I copied the error below for reference:
```
Access to XMLHttpRequest at 'http://test.outcondigital.com:8048/ad/get?internalId=12345678&publisher=5d5d66f2306ea4114a37c7c2&demo=true?' from origin 'http://test.localhost:9999' has been blocked by CORS policy: The value of the 'Access-Control-Allow-Origin' header in the response must not be the wildcard '*' when the request's credentials mode is 'include'. The credentials mode of requests initiated by the XMLHttpRequest is controlled by the withCredentials attribute.```

Comment (TinchoF) [reply modules/outconBidAdapter.js]: Thanks for the feedback. I already changed the headers on the server to accept CORS policies, it should already be working

Comment (jsnellbaker) [reply modules/outconBidAdapter.js]: I retested the adapter again, but I'm still seeing the CORS error (same as above).  Any ideas or recommedations?

Screen Shot 2019-09-23 at 12 03 11 PM

Comment (TinchoF) [reply modules/outconBidAdapter.js]: Please check now again

File: modules/outconBidAdapter.js [hunk 7]
```diff
@@ -0,0 +1,53 @@
...
+    }
+  },
+  interpretResponse: function(serverResponse, bidRequest) {
+    const bidResponses = [];
+    const bidResponse = {
+      requestId: serverResponse.body.id,
+      cpm: serverResponse.body.cpm,
+      width: serverResponse.body.creatives[0].width,
+      height: serverResponse.body.creatives[0].height,
+      creativeId: serverResponse.body.id,
```
jsnellbaker: See earlier note related to the `requestId` field.  This was resulting in a `null` value which isn't ideal.  Can you please ensure the field is populated or update this value in the code to a property that does exist in the raw bid response?

modules/outconBidAdapter.js
```diff
@@ -0,0 +1,53 @@
...
+      };
+    }
+  },
+  interpretResponse: function(serverResponse, bidRequest) {
+    const bidResponses = [];
+    const bidResponse = {
+      requestId: serverResponse.body.id,
+      cpm: serverResponse.body.cpm,
+      width: serverResponse.body.creatives[0].width,
+      height: serverResponse.body.creatives[0].height,
```
jsnellbaker: This coment relates to both `width` & `height` fields.  In the test bid, the corresponding values under the `creatives` object were not populated.

For banner bids, these fields need to be populated with valid values or else Prebid will reject the bid later in the auction workflow.  Can you make the necessary updates on the test bid to ensure these values are present in the response?

modules/outconBidAdapter.js
```diff
@@ -0,0 +1,53 @@
...
+        method: 'GET',
+        url: url,
+        data: {}
+      };
+    }
+  },
+  interpretResponse: function(serverResponse, bidRequest) {
+    const bidResponses = [];
+    const bidResponse = {
+      requestId: serverResponse.body.id,
```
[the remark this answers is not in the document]
TinchoF: with the bid requests we will receive bidRequests []. bidId generated by Prebid or by the client and we must respond with that same id?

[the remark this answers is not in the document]
jsnellbaker: Prebid generates the `bidId` on its own each time an auction is ran.  Some adapters pass in the value and include it on the response to link the objects together, so you can follow that approach if you want.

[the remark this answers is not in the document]
TinchoF: We generate our own "impressionId" does that work for you or do you need the same bidId that you generate?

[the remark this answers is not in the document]
jsnellbaker: It needs to be the same.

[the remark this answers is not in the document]
TinchoF: Ok, check now

modules/outconBidAdapter.js
```diff
@@ -0,0 +1,53 @@
...
+  code: BIDDER_CODE,
+  isBidRequestValid: function(bid) {
+    return !!((bid.params.pod || (bid.params.internalId && bid.params.publisher)) && bid.params.env && bid.params.bidId);
+  },
+  buildRequests: function(validBidRequests) {
+    for (let i = 0; i < validBidRequests.length; i++) {
+      let par = '';
+      let url = '';
+      if (validBidRequests[i].params.pod != undefined) par = 'get?pod=' + validBidRequests[i].params.pod + '&bidId=' + validBidRequests[i].params.bidId;
+      else par = 'get?internalId=' + validBidRequests[i].params.internalId + '&publisher=' + validBidRequests[i].params.publisher + '&bidId=' + validBidRequests[i].params.bidId;
```
jsnellbaker: As a follow-up to our conversation on the `bidId`, the `bidId` value gets created by Prebid.js core.  So if you want to pass into your server request so that it's there in the bid response, you'll want to read it from `validBidRequests[i].bidId` instead of relying on a `bid.param` value that the publisher would have to populate.

With this current setup, the bid response would be rejected by the prebid core logic and display the following warning message:
```
WARNING: Bidder outcon made bid for unknown request ID: 123456789. Ignoring.
```

TinchoF: Thanks, I already changed it, tell me if you need another change.

jsnellbaker:
Hi @TinchoF 

Thanks for making the updates around the `bidId` field.  I checked the adapter this morning and the bid is being accpeted by Prebid.  

However when I looked at the rendered content, I was seeing a URL to the file rather than the rendered asset.  I took a screenshot below:
Screen Shot 2019-09-30 at 9 27 49 AM

Normally the final value that's assigned to the `bidResponse.ad` field is an HTML markup of the returned ad.  In this case, the raw URL is being assigned to the `bidResponse.ad` field - which is why were just seeing the text render to the page.  

To resolve this, you'd likely either want to:
- alter the value in the raw response to include a proper HTML mark-up for the type of ad you're returning and set that to the `bidResponse.ad` field.
- add some logic to the adapter to parse the type of URL's file type and generate your HTML mark-up to wrap around the ad URL.  This new set of code would then be set to the `bidResponse.ad` field.
 
I made a note about parsing/interpreting the file in both cases, as when I loaded the hello world page with the test params several times - I saw one of the creative URLs point to an mp4 file instead of a simple image file.  I copied the URL below for reference if that doesn't sound right:
`http://test.outcondigital.com/uploads/5d42dff3306ea468996924b2/Voto Massa Promotion.mp4`

I hope the above makes sense.  Please let me know if you have any questions.

TinchoF:
I add some logic to wrap the url with HTML.
Regarding mime types is ok, the POD (point of display) used in the test accept both (banner and video) so you will receive videos or images.

TinchoF:
Hi!, is the adapter ready to be accepted or is there any change you need to make? Thanks greetings!

Issue #4166, opened 2019-09-05
PrebidJS User Id module should share userIds in ORTB Eids format as well


pm-harshad-mane:
## Type of issue
Prebid JS User Id module shares all user Ids with bidders in an object like below.
```
{
    “pubcid”: “SOME_PUB_COMMON_ID”,
    “tdid”: “TTD_ID”,
    “digitrustid” : { },
    “id5id”: “”
}
```

## Description
Then each bidder needs to pass these IDs to adapter server end-point in a preferred format.
Some bidders like PubMatic, PulsePoint, Rubicon, PrebidServerBidAdapter use ORTB format, these bidders need to put the UserIds in ORTB.user.ext.eid format.
For each new UserID these bidders need to make code changes which are similar for each adapter.
If UserID module shares userIds in ORTB format as well then these bidders will not need to update code to support each new user Id.

## Solution
We can add an ORTB format function in each userId submodule (like id5, digitrust).
ORTB format function of each sub-module will be called by index.js of userID module and an array of eids will be passed along with the current ```bid.userId``` object to each bid object.
Then bidders can simply copy/merge the ```bid.userIdEids``` object into their ORTB request.
- Each userId sub-module provides an ortbFormat function
- UserId module will call ortbFormat function to get eids in ORTB format for each submodule
- UserId module will keep the current plain object ```bid.userId```
- UserId module will add a new object ```bid.userIdEids``` which will have an ORTB eids array 
- Each ORTB supporting bidder will copy ```bid.userIdEids``` into ```request.user.ext.eids```

pm-harshad-mane:
This is supported now, https://github.com/prebid/Prebid.js/blob/master/modules/userId/eids.md

PR #4167, opened 2019-09-05
PrebidJS User Id module should share userIds in ORTB Eids format as well

pm-harshad-mane:
## Type of change
- [X] Feature

## Description of change
For details please refer #4166

### DIFF
--- a/modules/id5IdSystem.js
+++ b/modules/id5IdSystem.js
@@ -26,11 +26,10 @@ export const id5IdSubmodule = {
     return (value && typeof value['ID5ID'] === 'string') ? { 'id5id': value['ID5ID'] } : undefined;
   },
 
-  // todo: who can confirm the format?
   ortbFormat: function(data, eidsArray) {
     if (utils.isStr(data)) {
       eidsArray.push({
-        source: 'id5id',
+        source: 'id5-sync.com',
         uids: [{
           id: data
         }]
--- a/modules/id5IdSystem.js
+++ b/modules/id5IdSystem.js
@@ -25,6 +25,19 @@ export const id5IdSubmodule = {
   decode(value) {
     return (value && typeof value['ID5ID'] === 'string') ? { 'id5id': value['ID5ID'] } : undefined;
   },
+
+  // todo: who can confirm the format?
+  ortbFormat: function(data, eidsArray){
+    if (utils.isStr(data)) {
+      eidsArray.push({
+        source: 'id5id',
+        uids: [{
+          id: data          
+        }]
+      });
+    }
+  },
+
   /**
    * performs action to obtain id and return a value in the callback's response argument
    * @function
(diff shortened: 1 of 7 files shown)


pm-harshad-mane:
Please note that this PR is not ready yet, I raised just to give reference in the issue #4166  and confirm the approach.

modules/id5IdSystem.js
```diff
@@ -25,6 +25,19 @@ export const id5IdSubmodule = {
   decode(value) {
     return (value && typeof value['ID5ID'] === 'string') ? { 'id5id': value['ID5ID'] } : undefined;
   },
+
+  // todo: who can confirm the format?
+  ortbFormat: function(data, eidsArray) {
+    if (utils.isStr(data)) {
+      eidsArray.push({
+        source: 'id5id',
```
smenzer: in the original ORTB spec for handling eids (that never actually got implemented, hence still using an extension), the `source` was meant to be the domain of the id provider. In our case, this would be `id5-sync.com`. 

We don't have a strong point of view for whether the source should be `id5id` or `id5-sync.com`, but feel like we should follow a standard approach. Anyone here have an opinion?

modules/id5IdSystem.js
```diff
@@ -25,6 +25,19 @@ export const id5IdSubmodule = {
   decode(value) {
     return (value && typeof value['ID5ID'] === 'string') ? { 'id5id': value['ID5ID'] } : undefined;
   },
+
+  // todo: who can confirm the format?
```
smenzer: other than the question below on value in the `source` field, i can confirm the format is correct. @padurgeat fyi

modules/id5IdSystem.js
pm-harshad-mane: Thank you @smenzer , I think we should stick to the spec and keep the ```source``` as a domain value.

pm-harshad-mane: @smenzer I changed the ```source``` to ```id5-sync.com```.

jsnellbaker:
Hi @pm-harshad-mane 

Based on the earlier note, please confirm when this PR is ready for review.  I have tagged @idettman to do the initial review once it's ready.  Thanks!

pm-harshad-mane:
Hello @jsnellbaker ,
I still need to make a few changes, I will try to complete the code changes by end-of-week.

idettman:
Any updates on when this will be ready?

bretg:
@pm-harshad-mane - could you perhaps update the prebidServerBidAdapter to utilize these eids as well? Several of these user id modules are missing in the server-to-server scenario.

idettman:
Review is waiting on the prebid server adapter change bret requested to add the eid support

idettman:
Also, is the the "on hold" tag valid for this PR still?

pm-harshad-mane:
Hello @hadarpeer ,
This has been done with https://github.com/prebid/Prebid.js/pull/4916
We can close this one.

PR #4177, opened 2019-09-10
POC - evaluate promises for callback functionality | Do not merge

sumit116:
## Type of change

## Description of change
POC to evaluate promises for callback functionality to serve response of an async request.

As a test, I have implemented promises in the bottom most layer where we make ajax calls to adapter servers and monitor the time taken to complete the auction. However, the other layers of code keep using callbacks.


mkendall07:
This pull request **introduces 1 alert** when merging 633fe7715b478ebfc5bb1b7cf874106c4439056c into 8b92b1fbcc251067a740281880cbd85e3f90733f - [view on LGTM.com](https://lgtm.com/projects/g/prebid/Prebid.js/rev/pr-f39d493cc556cbfe3abf0f7c98bfb1c835e66de9)

**new alerts:**

* 1 for Unused variable, import, function or class

---

---

mkendall07:
This pull request **introduces 1 alert** when merging 1ef4cc1ca43f596c95df27288ee6c8e6d7df83c1 into 8b92b1fbcc251067a740281880cbd85e3f90733f - [view on LGTM.com](https://lgtm.com/projects/g/prebid/Prebid.js/rev/pr-6f4cb474cc12c0155d595c6d42c0c84ea2b57da8)

**new alerts:**

* 1 for Unused variable, import, function or class

---

---

mkendall07:
since IE 11 doesn't support native promises, we need to consider either 1) dropping IE 11 support or 2) Polyfill or promise library.

sumit116:
Hi @mkendall07, 

Below is the time taken on chrome and IE-11 browsers for auction to complete which shows a comparison between using promises v/s callback functionality for Prebid.js.

Browser - Chrome

|Parameter| Promises | Non-Promise (callbacks) |
| ----------- | -------------- | ----------------- |
|Avg Time taken (ms) to complete auction| 22.5ms|21.5ms|

Browser - IE-11

|Parameter| Promises | Non-Promise (callbacks) |
| ----------- | -------------- | ----------------- |
|Avg Time taken (ms) to complete auction| 5.2ms|8.8ms|

---------------------------------------------------------------------------------------------------------------

Below is the comparison between promise libraries to support Promises on IE-11.  

|Parameter| es6-Promise | promise-polyfill |
| ----------- | -------------- | ----------------- |
|bundle size (prebid.js)|452Kb|442Kb|
|Avg Time taken (ms) to complete auction (IE-11)| 5.2ms|5.2ms|


On chrome, the time taken to complete auction was also same for both libraries which varied between `15-35 ms` on multiple page refresh.


Notes: 

-  To avoid latency in ajax calls for testing, I have used mock response, hence the time taken is so low.
-  The bundled file contains only appnexusBidAdapter.js as the module.
-  The promise-polyfill is used with `set-asap` polyfill as a recommended  work around to support 
    setImmediate() on IE-11 which was causing performance issues. 
    (https://github.com/taylorhakes/promise-polyfill)

mkendall07:
@sumit116 
Thanks for the info. Do you know why IE 11 is faster than a modern chrome browser? Seems like a strange result.

snapwich:
The tests in this pull-request don't get to the heart of the issue with Promises, and that is _how the callbacks are prioritized_ by the event loop. That priority is what matters, because that what determines which code gets _priority_ to run in the single execution thread. Timing the results independently in isolation and then comparing the results ignores that most important distinction because what we really want to know is how we will perform compared to other code executing on the page at the same time as us. A few ms difference between isolated runtimes will not a big deal in the big picture.

1_4rN4DDM9LoXcLHWa_HX28Q

This image is a good demonstration of the order that queues are executed in the event loop. The important thing to note is that `.then` callbacks are handled by the microtasks queue which are executed after each step within the event loop. That's where we want to be for highest priority runtime at any given step of the event loop. I think I confirmed this behavior with this test setup here: https://github.com/snapwich/promises-vs-cb

I setup a similar environment that does ajax, ajax with Promises, and [the fetch API](https://developer.mozilla.org/en-US/docs/Web/API/Fetch_API). It also emulates a busy thread in the main execution as well as in the callbacks to simulate work being done (and us competing against this third-party work).

Screen Shot 2019-10-08 at 5 20 40 PM

These are the results I got running these requests in various orderings. You'll see when there's competition on the event loop for execution the `fetch` requests (underlined in green) always get priority due to them being scheduled as microtasks. Ajax request callbacks are executed as part of the I/O events queue which means they should always be _at least_ 1 tick away (while microtasks are always next tick), and returning Promises from an ajax request is the same as regular ajax requests but worse, because now we're also deferring execution to other handlers in the I/O events queue before our Promise is resolved next tick. So for requests, I think we should always be using `fetch` where possible.

As for other pieces of the code that are _not_ web requests, we should not be using Promises at all. Synchronous callbacks have the advantage of skipping the event loop queues altogether, which is even better than deferring code until next tick.

**tl;dr**: For best performance we should be using the promise-based [Fetch API](https://developer.mozilla.org/en-US/docs/Web/API/Fetch_API) for all of our web requests. We should avoid Promises everywhere else (if we can do a synchronous callback).

sumit116:
> @sumit116
> Thanks for the info. Do you know why IE 11 is faster than a modern chrome browser? Seems like a strange result.

@mkendall07, IE-11 behaved faster with no extension and no XHR call. When I added back the XHR calls (which we removed to avoid latency), it was slower than chrome or other modern browsers by atleast 150 ms (time taken to complete auction ) on an average. Adding any extension to it (like I have in chrome) adds up to the page load time.

mkendall07:
Thanks Rich for doing the analysis here. I'm going to close this PR.

PR #4178, opened 2019-09-10
LiveIntent Identity Module.

jankoulaga:
## Type of change
- [X] Feature

## Description of change

This PR introduces the LiveIntent Id module. For it's purpose, there are a couple config params we've imagined(outside of the standard `storage` part) would be useful for us. Here's an example of such a config with descriptions.

```
{
	name: 'liveIntentId',
	params: {
		publisherId: '7798696', // MANDATORY : Set an identifier of a publisher know to your systems 
		identifiersToResolve: ['tdid'], // OPTIONAL : Choose which additional values will be sent along to improve resolution. Those will be fetched from LS || Cookie space  
		url: 'https://liveintent.yourdomain.com' // OPTIONAL : If we already live under your domain, we can use the internal URI. If not set, a default on will be used.
	},
	storage: {
		type: 'cookie',
		name: '_li_pbid',
		expires: 60
	}
}
```


bretg:
Will need a docs PR against https://github.com/prebid/prebid.github.io/blob/master/dev-docs/modules/userId.md

jankoulaga:
Hey @pm-harshad-mane, we're still working on the docs portion and i assume it those will be distributed accordingly. Hang tight :)

jankoulaga:
Docs pr was created:
https://github.com/prebid/prebid.github.io/pull/1493

jankoulaga:
Hi @sumit116, can i get some details on what's exactly failing?

modules/userId/index.js
```diff
@@ -482,5 +485,6 @@ init(config);
 // add submodules after init has been called
 attachIdSystem(pubCommonIdSubmodule);
 attachIdSystem(unifiedIdSubmodule);
+attachIdSystem(liveIntentIdSubmodule);
```
idettman: The liveIntentIdSubmodule cannot be auto included. Only `unifiedId` and `pubCommonId` are currently allowed to be auto-included, and in Prebid 3.0 event pubCommonId and unifiedId will no longer be auto-included.

jankoulaga: Hi @idettman, then it should be extracted out of `modules/userId` structure into `modules/liveIntentIdSystem` so it can be added to the list of modules during build?

idettman: Check how the id5 submodule is implemented, basically remove  changes to index.js, and add to the .submodules.json. As long as defined in the submodules.json, it will be added when included in the modules list

jankoulaga: @idettman i think this has been resolved with the last commit

idettman:
can you add your submodule to the modules/.submodules.json?

bretg:
You also need to modify the PrebidServerBidAdapter https://github.com/prebid/Prebid.js/blob/master/modules/prebidServerBidAdapter/index.js to add something like this in the section around line 700:

```
      if (bidUserId.liveIntentId) {
        request.user.ext.eids.push({
          source: 'liveintent.com',
          uids: [{
            id: bidUserId.liveIntentId
          }]
        });
      }
```

bretg:
@pm-harshad-mane reports https://github.com/prebid/Prebid.js/issues/4304

@jankoulaga or @idettman please take a look?

PR #4184, opened 2019-09-11
Konduit module

konduit-dev:
## Type of change

- [x] Feature

## Description of change

Konduit module is providing a helper function (pbjs.adServers.konduit.buildVastUrl) to wrap a passed Prebid bid object and return a Konduit wrapped VAST url that can be then loaded with a VAST compliant video player.

```
pbjs.que.push(function(){
  pbjs.addAdUnits(videoAdUnit);
  pbjs.requestBids({
    timeout : 700,
    bidsBackHandler : function(bids) {
      var videoUrl = pbjs.adServers.konduit.buildVastUrl({
        bid: winnerBid,
        params: {
          konduit_id: 'your_konduit_id'
        }
      });
        
      invokeVideoPlayer(vastTagUrl);
    }
  });
});
```

### DIFF
--- a/modules/konduitWrapper.js
+++ b/modules/konduitWrapper.js
@@ -73,6 +73,14 @@ export function notifyTranslationModule(fn) {
 getHook('registerAdserver').before(notifyTranslationModule);
 
 function obtainVastUrl(bid) {
+  const vastUrl = bid && bid.vastUrl;
+
+  if (vastUrl) {
+    logInfo(`VAST url found in the bid - ${vastUrl}`);
+
+    return encodeURIComponent(vastUrl);
+  }
+
   const cacheUrl = config.getConfig('cache.url');
   if (cacheUrl) {
     const composedCacheUrl = `${cacheUrl}?uuid=${bid.videoCacheKey}`;
@@ -80,14 +88,6 @@ function obtainVastUrl(bid) {
     logInfo(`VAST url is taken from cache.url: ${composedCacheUrl}`);
 
     return encodeURIComponent(composedCacheUrl);
-  }
-
-  const vastUrl = bid && bid.vastUrl;
-
-  if (vastUrl) {
-    logInfo(`VAST url found in the bid - ${vastUrl}`);
-
-    return encodeURIComponent(vastUrl);
   }
 }
 
--- a/modules/konduitWrapper.js
+++ b/modules/konduitWrapper.js
@@ -40,7 +40,7 @@ export function buildVastUrl(options) {
 
   const queryParams = {};
 
-  const vastUrl = obtainVastUrl(bid, options, 'params');
+  const vastUrl = obtainVastUrl(bid);
 
   if (vastUrl) {
     queryParams.konduit_id = options.params.konduit_id;
--- a/modules/konduitWrapper.js
+++ b/modules/konduitWrapper.js
@@ -1,0 +1,96 @@
+import { registerVideoSupport } from '../src/adServerManager';
+import { targeting } from '../src/targeting';
+import { format as buildUrl } from '../src/url';
+import * as utils from '../src/utils';
+import { config } from '../src/config';
+import { getHook } from '../src/hook';
+
+const MODULE_NAME = 'Konduit';
+
+function addLogLabel(args) {
+  args = [].slice.call(args);
+  args.unshift(`${MODULE_NAME}: `);
+  return args;
+}
+
+export function logInfo() {
+  utils.logInfo(...addLogLabel(arguments));
+}
+
+export function logError() {
+  utils.logError(...addLogLabel(arguments));
+}
+
+export function buildVastUrl(options) {
+  if (!options.params || !options.params.konduit_id) {
+    logError(`'konduit_id' parameter is required for $$PREBID_GLOBAL$$.adServers.konduit.buildVastUrl function`);
+
+    return null;
+  }
+
+  const bid = options.bid || targeting.getWinningBids()[0];
+
+  if (!bid) {
+    logError('Bid is not provided or not found');
+
+    return null;
+  }
+
+  logInfo('The following bid will be wrapped: ', bid);
+
+  const queryParams = {};
+
+  const vastUrl = obtainVastUrl(bid, options, 'params');
+
+  if (vastUrl) {
+    queryParams.konduit_id = options.params.konduit_id;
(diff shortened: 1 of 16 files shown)


konduit-dev:
@mkendall07 need your assistance with the CircleCI failed test.
It is not related to the code change in this PR.
```
  RealVu
    Analytics Adapter.
      - isInView returns "yes"
      - isInView returns "yes"
  ✗ "after each" hook
	Script error. (:0)
```

Would be nice if the test can be restarted.
Thank you!

robertrmartinez:
Taking a look.

modules/konduitWrapper.md
```diff
@@ -0,0 +1,39 @@
...
+        bid: winnerBid,
+        params: {
+          konduit_id: 'your_konduit_id'
+        }
+      });
+        
+      invokeVideoPlayer(vastTagUrl);
+    }
+  });
+});
```
robertrmartinez: I think the setup example may need to be updated to better explain what is going on.

Where is this `winnerBid` coming from? Good idea to show how it is created from the `bids` coming back.

`invokeVideoPlayer(vastTagUrl);` is this supposed to be `invokeVideoPlayer(videoUrl);`?

Where do all these pieces come together and what exactly is happening in the example. Sure there will be confusion for someone who does not understand `konduit`

robertrmartinez: What does the publisher do with the returned `videoUrl`

Maybe an example test page I can take a look at will help my testing?

konduit-dev: Robert, we've updated the module documentation file and also created a related snippet at https://jsfiddle.net/edfcgruk/7/.
Please take a look and let us know should you have any questions.

konduit-dev:
@robertrmartinez please let us know if you have any additional comments for us to address.
Thank you!

modules/konduitWrapper.js
```diff
@@ -0,0 +1,96 @@
...
+export function notifyTranslationModule(fn) {
+  fn.call(this, 'konduit');
+}
+
+getHook('registerAdserver').before(notifyTranslationModule);
+
+function obtainVastUrl(bid) {
+  const cacheUrl = config.getConfig('cache.url');
+  if (cacheUrl) {
+    const composedCacheUrl = `${cacheUrl}?uuid=${bid.videoCacheKey}`;
```
robertrmartinez: @konduit-dev Hi this is a very nice looking setup. Very Clean.

But I did find one issue:

If it is a video bid, then there should always be a `vastUrl` on the bid.

Because before adding the bid to the bidsRecieved pool, if the entire VastXML is sent back, then the caching happens client side as long as `cache.url` is present, otherwise bid is thrown out.

So by the time it comes back to your module, presumably in the auction callback, the bid should have a vastUrl otherwise it is not a video bid!

Is there some other reason?

So basically you are setting the vastUrl manually if a cache.url is defined every time.

This is an issue if one of your bidders is a SERVER side cachers and returns their own VAST URL.

But the other bidders are Client side Cachers, so cache.url IS defined.

For example, Rubicon caches their Vast XML's Server Side and only returns the vastUrl,

So if you had the following:
1. Appnexus Cache as your client side config:
```javascript
pbjs.setConfig({
                cache: {
                    url: "https://prebid.adnxs.com/pbc/v1/cache"
                },
});
```

2. Rubicon And Appnexus as Video Bidders

3. Rubicon winds the bid they will return something like so:

image

As you can see from that `bidResponse`, Rubicon had set its own `VastUrl` from its server, which then Prebid Core will not utilize the client side caching.

But your module will incorrectly generate this winning VastXML as:

https://prebid.adnxs.com/pbc/v1/cache?uuid=7ced6033-6a17-4f8b-8dde-ed769a019ed2&konduit_header_bidding=1

Since we would have the appnexus client side cache url in our pbjs config.


My reccomendation is you can leave the `composedCacheUrl` as a failsafe.

But to try the vastUrl on the bid first.

```javascript
function obtainVastUrl(bid) {

  if (vastUrl) {
    logInfo(`VAST url found in the bid - ${vastUrl}`);

    return encodeURIComponent(vastUrl);
  }

  const cacheUrl = config.getConfig('cache.url');
  if (cacheUrl) {
    const composedCacheUrl = `${cacheUrl}?uuid=${bid.videoCacheKey}`;

    logInfo(`VAST url is taken from cache.url: ${composedCacheUrl}`);

    return encodeURIComponent(composedCacheUrl);
  }
}
```

robertrmartinez: Here is where you can see that Prebid will reject bids which do not have either:

1. a `vastUrl` returned back on the bid
OR
2. cache.url defined if only full `vastXml` is on the bid

https://github.com/prebid/Prebid.js/blob/master/src/video.js#L49

bretg:
We need a pull request to add a document to https://github.com/prebid/prebid.github.io/tree/master/dev-docs/modules

konduit-dev:
@robertrmartinez that you so much for the detailed comment!
We've updated the function accordingly.

modules/konduitWrapper.js
```diff
@@ -0,0 +1,96 @@
...
+      search: queryParams
+    });
+
+    logInfo(`Konduit wrapped VAST url: ${resultingUrl}`);
+  }
+
+  return resultingUrl;
+}
+
+export function notifyTranslationModule(fn) {
```
jaiminpanchal27: Any reason to add this hook. This is not getting used in your module.
We added this in dfpAdserverVideo to support longform use case. More details here https://github.com/prebid/Prebid.js/issues/3759

test/spec/modules/konduitWrapper_spec.js
```diff
@@ -0,0 +1,128 @@
+// eslint-disable-file
```
jaiminpanchal27: Please enable eslint

konduit-dev:
@jaiminpanchal27 thank you for the comments!
The code has been updated.
Please let us know should you have other comments or questions.

bretg:
@konduit-dev - we need someone to submit a documentation PR into https://github.com/prebid/prebid.github.io/tree/master/dev-docs/modules

It can be based on your .md file here. It can be simple like https://github.com/prebid/prebid.github.io/blob/master/dev-docs/modules/dfp_video.md or as detailed as you'd like

bretg:
Docs PR https://github.com/prebid/prebid.github.io/pull/1544 (thanks!)

PR #4220, opened 2019-09-23
MicroAd: Use HTTPS in all requests

strong-zero:
## Type of change

- [x] Feature
- [x] Other

## Description of change
* Stop using the protocol-relative URL of the MicroAd prebid endpoint.

## Other information
* This PR will be followed by enabling secure cookies in the MicroAd prebid endpoint.

### DIFF
--- a/modules/microadBidAdapter.js
+++ b/modules/microadBidAdapter.js
@@ -4,7 +4,7 @@ import { BANNER } from '../src/mediaTypes';
 const BIDDER_CODE = 'microad';
 
 const ENDPOINT_URLS = {
-  'production': '//s-rtb-pb.send.microad.jp/prebid',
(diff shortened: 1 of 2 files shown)


mkendall07:
@strong-zero 
Change looks good but the test is failing:
```
 microadBidAdapter
    buildRequests
      ✗ should always use the HTTPS endpoint https://s-rtb-pb.send.microad.jp/prebid even if it is served via HTTP
	TypeError: Object doesn't support property or method 'startsWith'
	   at Anonymous function (webpack:///test/spec/modules/microadBidAdapter_spec.js:265:8 <- test/test_index.js:160380:9)
	   at Anonymous function (webpack:///test/spec/modules/microadBidAdapter_spec.js:264:6 <- test/test_index.js:160379:7)
```

strong-zero:
OK. I'll fix it up 😢

Issue #4131, opened 2019-08-30
ShowHeroes only seem to support Instream


muuki88:
## Description

The showheroes adapter introduced in #3733 doesn't support `oustream`, but doesn't state this in the documentation nor does it valdidate the `mediaTypes.video.context`

## Steps to reproduce

Configure a showheroers bid

```javascript
{
  code: 'adslot-1,
  mediaTypes: {
    video: {
      context: 'outstream',
      playerSize: [ [ 605, 340 ] ]
    }
  },
  bids: [
    {
      bidder: 'showheroes-bs',
      params: {
        playerId: 'xxxxxx-xxxxxx-xxxxx-xxxxx'
      }
    }
  ]
}
```

### Expected results

1. The adapter documentation states that only `outstream` is supported
2. The adapter ignores `mediaType.video` if the context is `oustream`

### Actual results

The prebid function `isValidVideoBid` logs an error.

## Platform details

prebid `2.30.0`.

## Other information

- #3733 
- ShowHeroes Adapter Dev @vadim-mazzherin

patrickjae:
Hi!
Thanks for bringing this up. Could you please confirm the reported behavior when changing the parameter value of mediaTypes.video.playerSize to a simple list as documente [here](https://github.com/showheroes/Prebid.js/blob/master/modules/shBidAdapter.md).
I.e. does the configuration

```
{
  code: 'adslot-1,
  mediaTypes: {
    video: {
      context: 'outstream',
      playerSize: [ 605, 340 ]
    }
  },
  bids: [
    {
      bidder: 'showheroes-bs',
      params: {
        playerId: 'xxxxxx-xxxxxx-xxxxx-xxxxx'
      }
    }
  ]
}
```
also not work?
Thanks,
Patrick

muuki88:
Hi @patrickjae :smiley: 

Sorry for the late reply.

Changing the `playerSize` setting doesn't  change anything. I updated the initial report with the log message.

```
ERROR: Invalid bid from showheroes-bs. Ignoring bid: Video bid does not have required vastUrl or renderer property
```

The request is sent and a response comes back, but it's not what prebid expects. I must have overlooked [this line here|https://github.com/prebid/Prebid.js/blob/master/modules/shBidAdapter.js#L52-L54], which indicates that outstream is supported. However I don't see any renderer specified. Doesn't showhero provide it's own renderer (video player)?

cheers,
Muki

muuki88:
[...]

PR #4222, opened 2019-09-23
ShowHeroes adapter - expanded outstream support

vadim-mazzherin:
## Type of change

- [x] Bugfix
- [x] Feature
- [ ] New bidder adapter   
- [ ] Build related changes
- [ ] CI related changes
- [ ] Other

## Description of change

- Expanded outstream support
- Fix issue #4131 

Contact email of the adapter’s maintainer


modules/shBidAdapter.js
```diff
@@ -133,6 +147,14 @@ function createBids(bidRes, reqData) {
 
   bidRes.bids.forEach(function (bid) {
     const reqBid = bidMap[bid.bidId];
+    let currentBidRequest;
```
muuki88: This could be simplified to

```suggestion
    const currentBidRequest = reqBid.bidRequest.bids.find(bid2 => bid.bidId ===  bid2.bidId);
```

modules/shBidAdapter.js
```diff
@@ -57,6 +68,7 @@ export const spec = {
         type: streamType,
         bidId: bid.bidId,
         mediaType: isBanner ? BANNER : VIDEO,
```
muuki88: out of curiosity:
will video ads still be displayed if the media type is banner _and_  video?

modules/shBidAdapter.js
```diff
@@ -154,25 +176,95 @@ function createBids(bidRes, reqData) {
...
 }
 
+function outstreamRender(bid) {
+  const embedCode = createOutstreamEmbedCode(bid);
+  if (typeof bid.renderer.config.customRender === 'function') {
+    bid.renderer.config.customRender(bid, embedCode);
+  } else {
+    try {
+      const inIframe = utils.getBidIdParameter('iframe', bid.renderer.config);
+      if (inIframe && window.document.getElementById(inIframe).nodeName === 'IFRAME') {
```
muuki88: How should this work with DFP?

The **iframe** option is hard, because the iframe ID is assigned by `gpt.js`, which can change the naming scheme at any time, breaking the integration.

The **slot** option doesn't work either as DFP creatives are always delivered in an iframe and the standard prebid creative doesn't have any div elements.

From other implementations there are better options:

## inFrame flag

If you really need to know if you are in an iFrame or not use a flag instead. Then append the necessary code to the `body` and `head`. Scripts, divs, whatever you want. Not need to break out and search for the iFrame.

## slot_id

If you want to break out of the iFrame, then try to do something like `window.top.document.getElementById(..)`. I guess prebid has some helper functions that prevent illegal access if the creative was not delivered in a friendly iframe.

modules/shBidAdapter.js
```diff
@@ -154,25 +176,95 @@ function createBids(bidRes, reqData) {
...
+  const isStage = utils.getBidIdParameter('isStage', bid.renderer.config);
+  const urls = getEnvURLs(isStage);
+
+  const fragment = window.document.createDocumentFragment();
+
+  const script = window.document.createElement('script');
+  script.type = 'text/javascript';
+  script.src = urls.pubTag;
+  script.onload = function () {
+    window.ShowheroesTag=this;
```
robertrmartinez:
```/home/circleci/Prebid.js/modules/shBidAdapter.js
  241:25  error  Infix operators must be spaced  space-infix-ops```

Please update this so circleCi passes

`window.ShowheroesTag = this;`

File: modules/shBidAdapter.js (hunk 1 above)
Comment (veranevera) [reply modules/shBidAdapter.js]: Hi @muuki88! Changed :)

File: modules/shBidAdapter.js (hunk 2 above)
Comment (veranevera) [reply modules/shBidAdapter.js]: Our player will work and show video or display ad as banner in an iframe, in this case.

File: modules/shBidAdapter.js (hunk 3 above)
Comment (veranevera) [reply modules/shBidAdapter.js]: These options are not for DFP integration. They need for our publishers, which want to set our outstream directly on their page. They can use id of existing iframe or slot id (can be used any html element).

File: modules/shBidAdapter.js (hunk 1 above)
Comment (muuki88) [reply modules/shBidAdapter.js]: Nice :sunglasses:

File: modules/shBidAdapter.js (hunk 3 above)
Comment (muuki88) [reply modules/shBidAdapter.js]: tl;dr IMHO the `slot_id` and `iframe` parameter are unnecessary for an integration that relies on an Ad Server such as DFP that delivers creatives in iframes. From my limited experience it's a safe guess that the player was delivered in an iframe through and ad server. Especially as a Prebid module.

> These options are not for DFP integration. They need for our publishers, which want to set our outstream directly on their page. They can use id of existing iframe or slot id (can be used any html element).

I'm not sure how prevalent DFP is as an Ad Server, but I guess there's a fair amount of publisher that use DFP as an Ad Server along with Prebid. 

Your solution is what is documented as [Option 2: Serving without an ad server](http://prebid.org/dev-docs/show-outstream-video-ads.html#option-2-serving-without-an-ad-server) in the outstream-video-ads documentation. We haven't heard of any publisher doing this.

> If you would like to integrate our outstream with DFP, please, give me a link with an example of your bidding. We will try to found a solution.

This is not necessarily DFP relevant. Delivering ad creatives in an iFrame is IMHO a good practice to avoid ad fraud and illegal page access by creatives.

My point here is that the implementation is unnecessary complex and hinders the integration with any Ad Server that delivers in iFrames. At least this is how I understand the code :wink: . Please correct me if I missed something obvious.

Here's the [example page from prebid outstream working examples section](http://prebid.org/dev-docs/show-outstream-video-ads.html#working-examples):

http://prebid.org/dev-docs/show-outstream-video-ads.html

The appNexus outstream renders everything inside the iframe and only adds a small `div` for handling the expansion of the video. As mentioned in the initial issue this is how the appNexus adapter does it:
 [AppNexus Adpater `newRenderer`](https://github.com/prebid/Prebid.js/blob/master/modules/appnexusBidAdapter.js#L428-L452) and [AppNexus Adapter `oustreamRenderer`]

File: modules/shBidAdapter.js [hunk 5]
```diff
@@ -154,25 +169,95 @@ function createBids(bidRes, reqData) {
...
+      if (inIframe && window.document.getElementById(inIframe).nodeName === 'IFRAME') {
+        const iframe = window.document.getElementById(inIframe);
+        let framedoc = iframe.contentDocument || (iframe.contentWindow && iframe.contentWindow.document);
+        framedoc.body.appendChild(embedCode);
+        return;
+      }
+
+      const slot = utils.getBidIdParameter('slot', bid.renderer.config);
+      if (slot && window.document.getElementById(slot)) {
+        window.document.getElementById(slot).appendChild(embedCode);
```
muuki88:
```suggestion
        window.document.getElementById(slot).appendChild(embedCode);
        return;
```

modules/shBidAdapter.js
```diff
@@ -154,25 +169,95 @@ function createBids(bidRes, reqData) {
...
+      } else if (slot) {
+        utils.logError('[ShowHeroes][renderer] Error: spot not found');
+      }
```
muuki88:
```suggestion
      }
      // Served in an iframe by an ad server
      window.document.body.appendChild(embedCode);
```

The last case would be to just insert the `embedCode` inside the iframe.

modules/shBidAdapter.js
veranevera: Hi @muuki88! We created a demo page how we can work with DFP http://test.showheroes.com/schema/sh-prebid-outstream-google 
Please, take a look. We will return "banner" type and you can use "renderAd" function inside DFP. In this case, no need to use the frame wrapper for custom frames and slots.

muuki88: Hi @veranevera :smiley: 

Thanks a lot for the example page :+1: The banner solutions sounds good to me. Still unsure about the use case of the other options in Prebid, but you know what you are doing :grin: 

I'll give it a try and keep you posted. I saw that you were using `1x1` as a size, so the player resizes itself, right?

veranevera: @muuki88 we are looking forward 💃 
The player from the demo is responsive :) do you need fixed sizes?

muuki88: Nope. Responsive is perfect :smile:

modules/shBidAdapter.md
```diff
@@ -49,6 +49,42 @@ Module that connects to ShowHeroes demand source to fetch bids.
...
+                       context: 'outstream',
+                   }
+               },
+               bids: [
+                   {
+                       bidder: "showheroes-bs",
+                       params: {
+                           playerId: '0151f985-fb1a-4f37-bb26-cfc62e43ec05',
+                           vpaidMode: true, // by default is 'false'
+                           outstreamOptions: {
```
muuki88: Maybe add a note here that this is optional

modules/shBidAdapter.md
```diff
@@ -49,6 +49,42 @@ Module that connects to ShowHeroes demand source to fetch bids.
...
+               code: 'video',
+               mediaTypes: {
+                   video: {
+                       playerSize: [640, 480],
+                       context: 'outstream',
+                   }
+               },
+               bids: [
+                   {
+                       bidder: "showheroes-bs",
```
muuki88: During my testing I found that this bidder code causes issue with DFP as `-` is not a valid character for the _PATTERN:key_ macro ([see ad manager docs]([[https://support.google.com/admanager/answer/177381?hl=en]])).

I used the alias `showheroesBs` instead, which works.

modules/shBidAdapter.md
veranevera: @muuki88 right, thanks! Will do :)

muuki88: Awesome :cocktail:

modules/shBidAdapter.md
robertrmartinez: Is this something we want to fix / update?

robertrmartinez:
Couple things to note here:

shBidAdapter has one of the lowest test coverage rates among bidAdapters.

image

As a rule of thumb, anytime an update is made to the adapter, there should at least be some tests affected / updated.

And even better if we can add some new ones.

Also, I am having issues with the CircleCi as it continues to fail, but that is not related to this PR.

We are supposed to enforce an 80% test coverage for all modules.

Could we possibly add some tests to help us get closer to that mark?

vadim-mazzherin:
Hi @robertrmartinez 
I'll add more tests to increase coverage
Thanks!

muuki88:
Thanks @veranevera and @vadim-mazzherin for your work and responsivness to feedback :heart: 

@robertrmartinez any chance this can be part of the next prebid release?

robertrmartinez:
@muuki88 Sure thing!
I appreciate all the great work from you all on this. Great to see such a large amount of feedback and back and forths.

Thanks for getting the test coverage up to @vadim-mazzherin !

PR #4082, opened 2019-08-11
userId - Add support for refreshing the cached user id

padurgeat:
## Type of change
- [x] Feature
- [x] Does this change affect user-facing APIs or examples documented on http://prebid.org?

## Description of change
Today in the userId module, user id are stored in 1st space of publisher, with expiration. If user refused third party cookies, when the first party cookie expires, the user id is lost. This PR adds a refresh feature to enable periodic check and refresh of the first-party cookie.

Mode detailed explanation about other scenarios in this issue: https://github.com/prebid/Prebid.js/issues/3989

For details please contract padurgeat@id5.io

Link to PR for documentation to follow in comments

Issue #3989, opened 2019-07-11
Add a refreshes option to userId module


smenzer:
## Type of issue
Feature request

## Description
Today in the userId module, there is an option to set the expiration of the cookie that stores the provider's user ID. Some providers, though, may need to be called regularly (every hour, every day, etc) to refresh the user ID due to rotating IDs (e.g. the user's ID is encrypted per publisher, and the key used to do the encryption rotates every day for privacy and security reasons). The only solution to ensure prebid calls the provider every day (for example) is to set the `expires` to 1 day. But this means that every day the cookie will expire, and for some period of time, the user will have no ID and will miss the first impression (since user id calls are made after the bid requests are made). This also means the provider may have lost their master user ID for this user (since the cookie expired) and could result in a completely new user ID being generated, losing the consistency that is otherwise offered with the prebid solution.

Prebid should add a `refreshes` option next to the existing `expires` option that will force a user id request call to the provider at whatever time period is set. This value should support values down to the minute, rather than being in days like expires is today. It may make sense to name the variables (or at least the new one) according to their time unit (e.g. `refreshesMinutes` and `expiresDays`).

jaiminpanchal27:
I think it is little confusing to keep refresh and expire together. If the user id needs to be refreshed in an hour doesn't it mean that it is expired. For instance, lets consider expire as 1 day and refresh as 30 min. In this case User id module should refresh the user id every 30 minutes and store it in cookie/localstorage. So What is the use for expire here ? What should happen after 1 day ?

Without making a breaking change, I think adding 
- [delayAuction](https://github.com/prebid/Prebid.js/issues/3952)
- storage as optional. Let id systems manage their own storage. Done in https://github.com/prebid/Prebid.js/pull/3930
should work for all scenarios.

I had similar requirement with Criteo rtus module. Criteo's user id expires in 1 hour but as per current user id api, storage is a required param and expires is set in days. Hence to avoid breaking change, I made storage as optional and handled storage logic in the module itself. 

Only concern with this implementation is that I am missing user id on first page load. That will be handled with delayAuction feature.

smenzer:
@jaiminpanchal27 so if storage is marked as option, then within the ID5 submodule, we can dictate how the id is stored (cookie, local storage, etc (potentially via a param flag to give pubs control if they want it)) as well as when we need to refresh?

[...]

PR #3869, opened 2019-05-29
UserId - Add SameSite and server-side pubcid support

pycnvr:
## Type of change

- [ ] Bugfix
- [x] Feature
- [ ] New bidder adapter   
- [ ] Code style update (formatting, local variables)
- [ ] Refactoring (no functional changes, no api changes)
- [ ] Build related changes
- [ ] CI related changes
- [ ] Does this change affect user-facing APIs or examples documented on http://prebid.org?
- [ ] Other

## Description of change

The main change in this PR is to allow publisher's own servers to extend pubcid's expiration time.  The server side support can vary, but the basic idea is to look for the pubcid cookie, and just rewrite the value with the desired attributes.  The 3 new parameters accommodate most likely scenarios.   

- Servers can read/write cookies on most pages.
- Servers use cached pages and can only update pubcid cookie in a single endpoint.
- Servers determines cookie consent and will create pubcid.

This covers both userId and pubCommonId modules.  Since pubCommonId is being deprecated, this is likely that last set of changes for it.  

- Add SameSite=Lax to first party cookies
- Add support for server-side pubcid cookies
  - Support params.extend flag to extend expiration time on each access
  - Support params.create flag to let first party servers create cookies
  - Support params.pixelUrl to ping first party servers to modify cookie attributes
- Fix optout support 
- Add test cases for end-of-auction callbacks

Contact:
- contact email of the adapter’s maintainer: pyang@conversantmedia.com
- [ ] official adapter submission

For any changes that affect user-facing APIs or example code documented on http://prebid.org, please provide:

- A link to a PR on the docs repo at https://github.com/prebid/prebid.github.io/

PR #4225, opened 2019-09-23
Add refreshId method support to userId submodule interface

icflournoy:
## Type of change
- [x] Feature

## Description of change
This change supports the forthcoming Parrable userId system and allows us to maintain a previously persisted ID. Our use case is closely described by an older issue: https://github.com/prebid/Prebid.js/issues/3989

We extend the userId module interface to support a `refreshId` method, similar to the `getId` method.
`refreshId(configParams, consentData, storedId)` takes the ID retrieved from storage and allows the ID system to determine if the stored ID value should be updated with a new ID value.
The return from refreshId is set as the userId submodule's callback to be processed similar to the `getId` call that first stored the ID.

Recent work by ID5 introduced a `storage.refreshInSeconds` config item, which was supported by adding a `storedId` parameter to the `getId()` signature. https://github.com/prebid/Prebid.js/pull/4082.

While our intended behavior supported by this PR could be simulated by setting `storage.refreshInSeconds = 1` we believe that this refreshId method provides a cleaner way of signaling the intent to update an ID if the ID system deems an update is necessary. This feature also opens up the ability for any userId system to implement a refreshId function and handle ID maintenance each time the ID is accessed.

The ability to refresh our ID is inherent to our ID system and thus should not be a publisher configuration item like `storage.refreshInSeconds`.

We are open to ideas on how to integrate ID5's refreshInSeconds work with our refreshId work. Both features seem like two behaviors of the same "ID refresh" functionality.

### Proposal
1. `getId()` reverts to its former interface `getId(submodule.config.params, consentData)`
2. `refreshId(submodule.config.params, consentData, storedId)` is a supported method of a userId system submodule and, if it exists, is called each time `initSubmodules()` occurs.
3. **If** ID5's config item for `storage.refreshInSeconds` is specified then `refreshId()` is only called if the refreshInSeconds period has elapsed (otherwise refeshId is called every time).

We have built test in userId spec that runs two auctions and verifies that the ID has been updated by the second auction. This test is currently failing since we're not sure how to trigger the end of an auction event, thus having the userId module callbacks processed.

### Documentation
This is a change to the userId system supported interface - which we cannot find a documentation page for the supported methods when building a userId module. 

Points of contact for this PR are @eyas-ranjous and @icflournoy.

cc: @padurgeat and @smenzer from ID5, as this will be of interest to them and we are happy to have their input.

### DIFF
--- a/modules/emx_digitalBidAdapter.js
+++ b/modules/emx_digitalBidAdapter.js
@@ -7,7 +7,7 @@ import includes from 'core-js/library/fn/array/includes';
 const BIDDER_CODE = 'emx_digital';
 const ENDPOINT = 'hb.emxdgt.com';
 const RENDERER_URL = '//js.brealtime.com/outstream/1.30.0/bundle.js';
-const ADAPTER_VERSION = '1.40.3';
+const ADAPTER_VERSION = '1.41.0';
 const DEFAULT_CUR = 'USD';
 
 export const emxAdapter = {
@@ -39,11 +39,11 @@ export const emxAdapter = {
       h: sizes[0][1]
     };
   },
-  formatVideoResponse: (bidResponse, emxBid) => {
+  formatVideoResponse: (bidResponse, emxBid, bidRequest) => {
     bidResponse.vastXml = emxBid.adm;
-    if (!emxBid.renderer && (!emxBid.mediaTypes || !emxBid.mediaTypes.video || !emxBid.mediaTypes.video.context || emxBid.mediaTypes.video.context === 'outstream')) {
+    if (bidRequest.bidRequest && bidRequest.bidRequest.mediaTypes && bidRequest.bidRequest.mediaTypes.video && bidRequest.bidRequest.mediaTypes.video.context === 'outstream') {
       bidResponse.renderer = emxAdapter.createRenderer(bidResponse, {
-        id: emxBid.bidId,
+        id: emxBid.id,
         url: RENDERER_URL
       });
     }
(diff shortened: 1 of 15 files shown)


padurgeat:
Hey Ian,
My understand from the past Criteo PR (`criteortusIdSubmodule`) is that if storage is not defined, (line 349), `getId()` will be called for every prebid call, which looks like the behaviour you're looking for. Cheers,

icflournoy:
HI Pierre-Antoine - thanks for pointing that out. Criteo's PR did change the behavior such that if storage isn't defined then getId is called each time. It also looks like Criteo is fetching their cookie in getId their own way (using `utils.getCookie`), effectively circumventing the Prebid userId core module's storage mechanism.

We do want to store our ID using Prebid's storage mechanism, but we also want to refresh it, and would rather build onto the core userId API to benefit all modules than relying on custom logic within our own userId system.

padurgeat:
Right! Let me know if I can help you. You can give me access to your fork if you need us to modify/review our implementation. Cheers,

icflournoy:
Hey @padurgeat I've taken a first pass at updating your ID5IdSystem implementation with a refreshId() method that is invoked when `storage.refreshInSeconds` expires. I've invited you as a collaborator on our fork - you can take review the changes here https://github.com/Parrable/Prebid.js/pull/2

Thanks in advance for taking a look!

jsnellbaker:
@icflournoy When you have the chance, can you please resolve the conflicts?

icflournoy:
Hi @jsnellbaker we have some additional changes related to the refreshId and refreshInSeconds functionality that we're going to push soon to this branch. I will resolve conflicts at that time so I am working with the latest changes in your master branch.

icflournoy:
Hi @jsnellbaker I'm closing this PR because it's superseded by changes from a recently-merged PR https://github.com/prebid/Prebid.js/pull/3869.  We will work on utilizing this change in the userId submodule interface in our own ID system.

@padurgeat thanks for your assistance on this, will let you know if we have any changes that overlap with your `config.refreshInSeconds` feature.

PR #4227, opened 2019-09-24
Force https scheme for Criteo Bidder

leonardlabat:
## Type of change
- [X] Feature

## Description of change
We'd like to ensure that our bidder is always called through https.

pi-direct@criteo.com


r-schweitzer:
Hey @leonardlabat, is there any plan for your guys to force fastbid to use https. It looks like it still currently uses a relative protocol.

leonardlabat:
Hi @r-schweitzer 

Yes it will be part of our next release in the next days

PR #4230, opened 2019-09-25
Adding workflow to run end to end tests

jaiminpanchal27:
## Type of change
- [x] Feature

## Description of change
This PR adds a new circleci job to run end to end tests every night on master branch. Job to run unit tests remains same. I have updated the config to use [circleci workflow](https://circleci.com/docs/2.0/workflows/) and [aliases](https://circleci.com/docs/2.0/writing-yaml/#anchors-and-aliases) 

#### Why run once every night ?
- Our end to end tests are not very stable due to various reasons, problems with browserstack booting up browsers, connection failures in browserstack and no bid response in browserstack vm's. 
- Functional tests takes around 15 mins to run and hence it is not feasible as of now to run on every commit.

Please see https://circleci.com/gh/prebid/workflows/Prebid.js for past successful runs of both jobs.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -62,7 +62,7 @@ workflows:
   nightly:
     triggers:
       - schedule:
-          cron: "5 * * * *"
+          cron: "0 0 * * *"
           filters:
             branches:
               only:
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -2,58 +2,81 @@
 #
 # Check https://circleci.com/docs/2.0/language-javascript/ for more details
 #
+
+aliases: 
+  - &environment
+      docker:
+        # specify the version you desire here
+        - image: circleci/node:8.9.0
+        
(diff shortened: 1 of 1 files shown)


.circleci/config.yml
```diff
@@ -2,49 +2,93 @@
...
+      name: Install gulp cli
+      command: sudo npm install -g gulp-cli
+
+  - &run_unit_test
+      name: BrowserStack testing
+      command: gulp test --browserstack --nolintfix
+
+  - &run_endtoend_test
+      name: BrowserStack End to end testing
+      command: echo "Running end to end test using workflows"
```
jsnellbaker: Can you clarify how this command starts the e2e test process?  I don't see the normal gulp e2e command in this file and this command appears to just print the noted text.

jaiminpanchal27: I used echo to verify workflow. Updated the command now

Fawke:
LGTM

PR #4231, opened 2019-09-25
add viewdeosDX whitelabel

GeneGenie:
## New Adapter ViewDeos
- [x] New bidder adapter

## Description of change
new adapter 

- support@viewdeos.com

- A link to a PR on the docs repo at https://github.com/prebid/prebid.github.io/pull/1511

### DIFF
--- a/modules/viewdeosDXBidAdapter.js
+++ b/modules/viewdeosDXBidAdapter.js
@@ -15,7 +15,7 @@ export const spec = {
   aliases: ['viewdeos'],
   supportedMediaTypes: [VIDEO, BANNER],
   isBidRequestValid: function (bid) {
-    return bid && bid.params && bid.params.aid;
+    return !!utils.deepAccess(bid, 'params.aid');
   },
   getUserSyncs: function (syncOptions, serverResponses) {
     var syncs = [];
@@ -24,9 +24,9 @@ export const spec = {
       const uris = bid.cookieURLs;
       const types = bid.cookieURLSTypes || [];
 
-      if (uris && uris.length) {
+      if (Array.isArray(uris)) {
         uris.forEach((uri, i) => {
-          let type = types[i] || 'image';
+          const type = types[i] || 'image';
 
           if ((!syncOptions.pixelEnabled && type == 'image') ||
             (!syncOptions.iframeEnabled && type == 'iframe')) {
@@ -42,7 +42,7 @@ export const spec = {
     }
 
     if (syncOptions.pixelEnabled || syncOptions.iframeEnabled) {
-      serverResponses && serverResponses.length && serverResponses.forEach((response) => {
+      utils.isArray(serverResponses) && serverResponses.forEach((response) => {
         if (response.body) {
           if (utils.isArray(response.body)) {
             response.body.forEach(b => {
@@ -93,7 +93,7 @@ export const spec = {
 };
 
 function parseRTBResponse(serverResponse, bidderRequest) {
-  const isInvalidValidResp = !serverResponse || !serverResponse.bids || !serverResponse.bids.length;
+  const isInvalidValidResp = !serverResponse || !utils.isArray(serverResponse.bids);
 
   let bids = [];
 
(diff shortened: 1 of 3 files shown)


mkendall07:
@GeneGenie 
Can you use the alias functionality instead of adding a new bidder?

GeneGenie:
@mkendall07 Hello
this does not look like an option as adapter released on it's own server infrastructure and should be treated as a separate prebid adapter

modules/viewdeosDXBidAdapter.js
```diff
@@ -0,0 +1,243 @@
...
+const BIDDER_CODE = 'viewdeosDX';
+const OUTSTREAM = 'outstream';
+const DISPLAY = 'display';
+
+export const spec = {
+  code: BIDDER_CODE,
+  aliases: ['viewdeos'],
+  supportedMediaTypes: [VIDEO, BANNER],
+  isBidRequestValid: function (bid) {
+    return bid && bid.params && bid.params.aid;
```
idettman: could shorten to: `utils.deepAccess(bid, 'params.aid')`

modules/viewdeosDXBidAdapter.js
```diff
@@ -0,0 +1,243 @@
...
+  },
+  getUserSyncs: function (syncOptions, serverResponses) {
+    var syncs = [];
+
+    function addSyncs(bid) {
+      const uris = bid.cookieURLs;
+      const types = bid.cookieURLSTypes || [];
+
+      if (uris && uris.length) {
```
idettman: I'd recommend changing to: `if (Array.isArray(uris) {`

modules/viewdeosDXBidAdapter.js
```diff
@@ -0,0 +1,243 @@
...
+        uris.forEach((uri, i) => {
+          let type = types[i] || 'image';
```
idettman: line 29 could be changed to use `const` instead of `let` since it's not reassigned after declaration.

modules/viewdeosDXBidAdapter.js
```diff
@@ -0,0 +1,243 @@
...
+
+          if ((!syncOptions.pixelEnabled && type == 'image') ||
```
idettman: Note: I'd recommend using strict equality tests when testing for a specific string value.
`type === 'image'` and `type === 'iframe'`

GeneGenie: why is that critical? Some integer value could become "image"?

modules/viewdeosDXBidAdapter.js
```diff
@@ -0,0 +1,243 @@
...
+
+export const spec = {
+  code: BIDDER_CODE,
+  aliases: ['viewdeos'],
+  supportedMediaTypes: [VIDEO, BANNER],
+  isBidRequestValid: function (bid) {
+    return bid && bid.params && bid.params.aid;
+  },
+  getUserSyncs: function (syncOptions, serverResponses) {
+    var syncs = [];
```
idettman: since `syncs` is not reassigned, it can be declared with `const` rather than `var`

modules/viewdeosDXBidAdapter.js
```diff
@@ -0,0 +1,243 @@
...
+
+      bids.push(bid);
+    }
+  });
+
+  return bids;
+}
+
+function bidToTag(bidRequests, bidderRequest) {
+  let tag = {
```
idettman: change `let` to `const` here since it's not reassigned

modules/viewdeosDXBidAdapter.js
```diff
@@ -0,0 +1,243 @@
...
+}
+
+/**
+ * Configure new bid by response
+ * @param bidResponse {object}
+ * @param mediaType {Object}
+ * @returns {object}
+ */
+function createBid(bidResponse, mediaType) {
+  let bid = {
```
idettman: change `let` to `const` since `bid` is not reassigned to another reference

modules/viewdeosDXBidAdapter.js
idettman: strict equality is more performant

GeneGenie: good to know

GeneGenie:
Done

Issue #4237, opened 2019-09-26
schain module breaks in IE11


DecayConstant:
## Type of issue
Bug

## Description
The schain module uses `Object.values` which is not supported in IE11: https://github.com/prebid/Prebid.js/blob/master/modules/schain.js#L136

This crashes the prebid code when running in IE11.

## Steps to reproduce
Try to use the schain module in IE11.

ehoch:
@pm-harshad-mane

pm-harshad-mane:
Looking into it on priority 👍

pm-harshad-mane:
Hello @ehoch and @DecayConstant ,
Code changes are merged to PrebidJS master.

pm-harshad-mane:
Hello @DecayConstant ,

Can you please close the issue?

PR #4238, opened 2019-09-26
Schain: avoiding Object.values as it is breaking on IE11

pm-harshad-mane:
## Type of change

- [X] Bugfix

## Description of change
Schain: avoiding Object.values as it is breaking on IE11
https://github.com/prebid/Prebid.js/issues/4237


pm-harshad-mane:
Hello @bretg , @idettman ,

pm-harshad-mane:
Thank you @mkendall07

PR #4240, opened 2019-09-27
Add cpmDistribution function for Google Analytics adapter

muuki88:
## Type of change

- [ ] Bugfix
- [x] Feature
- [ ] New bidder adapter   
- [ ] Build related changes
- [ ] CI related changes
- [ ] Other

## Description of change

Add an additional option `cpmDistribution` to the google analytics adapter to allow customization of the cpm buckets.

### DIFF
--- a/modules/googleAnalyticsAdapter.js
+++ b/modules/googleAnalyticsAdapter.js
@@ -262,6 +262,11 @@ function sendBidWonToGa(bid) {
   checkAnalytics();
 }
 
+/**
+ * Exposed for testing purposes
+ */
+adapter.getCpmDistribution = getCpmDistribution;
+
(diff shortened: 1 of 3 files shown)


muuki88:
The test coverage of the analytics adapter is rather poor. I can try to add more if you think this makes sense, but I would appreciate some hints as I'm not sure how to mock things to verify the proper methods are being called.

Also the implementation with a separate queue is a bit strange as GA already provides an API for that, which may make tests even trickier.

muuki88:
I added a test without the mocking stuff. I'm puzzled that I can't  call `enableAnalytics` twice, which makes it impossible to write isolated test cases. The `AdapterManager` doesn't seem to override this method, but it hell looks like it :smile:

PR #4241, opened 2019-09-27
Deprecate user sync

Fawke:
## Type of change

- [ ] Bugfix
- [ ] Feature
- [ ] New bidder adapter   
- [ ] Build related changes
- [ ] CI related changes
- [x] Other

## Description of change
*enabledBidders*, *iframeEnabled* & *pixedEnabled* properties have been deprecated. The unit test cases have been updated to reflect that.


test/spec/userSync_spec.js
```diff
@@ -250,6 +265,7 @@ describe('user sync', function () {
 
   it('should throw a warning and default to basic resgistration rules when filterSettings config is invalid', function () {
     // invalid config - passed invalid filter option
+    debugger; // eslint-disable-line
```
jaiminpanchal27: Remove this breakpoint

Fawke: Thanks for checking that, I've removed it now.

src/userSync.js
```diff
@@ -31,7 +36,7 @@ export function newUserSync(userSyncDependencies) {
 
   // for now - default both to false in case filterSettings config is absent/misconfigured
   let permittedPixels = {
-    image: false,
+    image: true,
```
robertrmartinez: This was not mentioned in the PR,

Are we defaulting to enabling image pixels now?

Fawke: Hi, 

Yes, userSync feature had image pixels enabled by default, prior to this change. Reason for this change here is that, we removed certain deprecated fields, as a result, we lost that default setting. We're setting this value to true now to maintain the default setting.

PR #4246, opened 2019-10-01
SmartRTB adapter update

evanmsmrtb:
## Type of change

- [ ] Bugfix
- [x] Feature
- [ ] New bidder adapter   
- [ ] Build related changes
- [ ] CI related changes
- [ ] Other

## Description of change
Added video support, differentiate img/iframe syncs, and various user id formats.


evanmsmrtb:
Please check again after our recent commit.

Fawke:
@evanmsmrtb,

I tested your adapter on the (hello world example)[https://github.com/prebid/Prebid.js/blob/master/integrationExamples/gpt/hello_world.html] with the ad unit provided by you in the *md* file, it's throwing two errors.
Screenshot 2019-10-08 at 12 30 05 PM
Screenshot 2019-10-08 at 12 30 18 PM

Also, can you please provide a Test Ad Unit for video as well in the *md* file?

evanmsmrtb:
Please try again now, the syncs item has been addressed. Video support has also been added to the sample ad unit as requested, and will return a bid through the forceBid param.  Do note that we require publishers to provide their own renderer for video, so there is not one included at this time. We do check for a pub provided renderer in our request.

Fawke:
@evanmsmrtb 

I'm getting a `null` in the response sent to your Ad server using your Test ad unit params. Can you please test it with the [hello world example](https://github.com/prebid/Prebid.js/blob/master/integrationExamples/gpt/hello_world.html)?

Also, your **interpretResponse** function is throwing an error, it's not able to handle a **null** response.

evanmsmrtb:
Added a check for the null response, though this was primarily due to us not migrating the endpoint yet. All should be well and work now.

Fawke:
Hey @evanmsmrtb,

I've tested your banner ad and it's now working. While testing your banner ad, I used the following Ad Unit object.

```
[{
			mediaTypes: {
				banner: {
					sizes: [[300, 250]],
				}
			},
			bids: [{
				bidder: "smartrtb",
				params: {
					zoneId: "N4zTDq3PPEHBIODv7cXK",
					forceBid: true
				}
			}]

		}]
```

The client should receive one bid response object with the winning bid showing the banner ad.
But actually, the client receive two bids in the response, one for banner, and one for video, even though, I didn't specifically request a video ad.

Same thing for happens for video ad as well. I receive two bids, banner and video when I should receive just one bid response.

evanmsmrtb:
In our platform we support multiple formats per placement and can enable/disable them in our UI also. We have updated our server side logic to filter placement formats by what is requested in the prebid tag. Please try again and you will now see one bid for the requested format.

Fawke:
Hi @evanmsmrtb,
Thanks for making the change. 

Since you are now supporting outstream video but you require the publishers to provide their own renderer, I would recommend to document this requirement somewhere. Both the *md* file an a docs update in prebid.org will be preferable.

evanmsmrtb:
Docs updated as requested, here and in the main prebid repo.

PR #4259, opened 2019-10-04
Vrtcal Markets Inc. Bid Adapter Addition

vrtcal-dev:
## Type of change

- [ ] Bugfix
- [ ] Feature
- [X ] New bidder adapter   
- [ ] Build related changes
- [ ] CI related changes
- [ ] Other

## Description of change

NEW ADAPTER

```
{
  bidder: 'vrtcal'
  //Vrtcal does NOT require any custom params in requests
}
```

DONE

support@vrtcal.com

QUESTION: Being that our adapter does not require any custom params, do we still need to submit the param doc at https://github.com/prebid/prebid.github.io/tree/master/dev-docs/bidders?

### DIFF
--- a/modules/vrtcalBidAdapter.js
+++ b/modules/vrtcalBidAdapter.js
@@ -1,0 +1,84 @@
+import {registerBidder} from '../src/adapters/bidderFactory';
+import { BANNER } from '../src/mediaTypes';
+import {ajax} from '../src/ajax';
+
+export const spec = {
+  code: 'vrtcal',
+  supportedMediaTypes: [BANNER],
+  isBidRequestValid: function (bid) {
+    return true;// No extras params required
+  },
+  buildRequests: function (bidRequests) {
+    const requests = bidRequests.map(function (bid) {
+      const params = {
+
+        prebidJS: 1,
+        prebidAdUnitCode: bid.adUnitCode,
+        id: bid.bidId,
+        imp: [{
+          id: '1',
+          banner: {
+            w: bid.sizes[0][0],
+            h: bid.sizes[0][1],
+          },
+          bidfloor: 0.75
+        }],
+        site: {
+          id: 'VRTCAL_FILLED',
+          name: 'VRTCAL_FILLED',
+          cat: ['VRTCAL_FILLED'],
+          domain: decodeURIComponent(window.location.href).replace('https://', '').replace('http://', '').split('/')[0]
+
+        },
+        device: {
+          ua: 'VRTCAL_FILLED',
+          ip: 'VRTCAL_FILLED'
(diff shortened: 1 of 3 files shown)


vrtcal-dev:
Hello project leads.  Do we need to do anything additional in order to have a reviewer assigned to our newly submitted adapter?  Thank you.

modules/vrtcalBidAdapter.js
```diff
@@ -0,0 +1,84 @@
...
+        ad: response.seatbid[0].bid[0].adm,
+        nurl: response.seatbid[0].bid[0].nurl
+      };
+
+      bidResponses.push(bidResponse);
+    }
+    return bidResponses;
+  },
+  getUserSyncs: function(syncOptions, serverResponses) {
+    return [];
```
jsnellbaker: Can remove this function if you're not going to be doing any usersyncing.

modules/vrtcalBidAdapter.js
```diff
@@ -0,0 +1,84 @@
...
+      const params = {
+
+        prebidJS: 1,
+        prebidAdUnitCode: bid.adUnitCode,
+        id: bid.bidId,
+        imp: [{
+          id: '1',
+          banner: {
+            w: bid.sizes[0][0],
+            h: bid.sizes[0][1],
```
jsnellbaker: May want to look at the `mediaTypes.banner.sizes` field instead of the `bid.sizes` field, as that generic field may be deprecated in the future.

bretg:
Docs PR https://github.com/prebid/prebid.github.io/pull/1536

vrtcal-dev:
Hello, Is anything additional needed in order for us to be reviewed again?

vrtcal-dev:
Thank you jsnellbake. 

You'll notice that we don't require any custom params.  This is primarily due to the fact that we only do direct publisher relationships and are thus able to properly identify the publisher by the traffic source itself as it only relates to a single publisher. This same intrinsic publisher identification also allows us to enable any publisher specific testing if needed without the need to pass a special param.  The use of the 'code' param was solely for your ad return verification needs.  Please proceed with merging our adapter as we are satisfied with its implementation.

jsnellbaker:
@vrtcal-dev Thanks for the follow-up response.  

There is one item that needs to be addressed, there are number of lint related formatting errors in the main adapter file and the test spec file.  If you run `gulp lint` on your end - it will automatically correct the issues for you and can be simply pushed into the PR as another commit.

Can you please make this updates?  Thanks!

vrtcal-dev:
Ok, lint formatting corrections applied. Please merge..Thanks again.

PR #4266, opened 2019-10-07
Add Parrable ID submodule

eyas-ranjous:
## Type of change
- [X] Feature

## Description of change
* Add Parrable ID submodule to userId to create and store the device encrypted ID in order to be included in the bid request.

~* The Parrable ID that is delivered to Prebid is encrypted by Parrable with a time-based key and updated frequently in the browser to enforce consumer privacy requirements and thus we added a feature to always refresh the stored ID when a submodule belongs to a "submodulesThatAlwaysRefresh" list. Ours is the first.~

* Docs PR link https://github.com/prebid/prebid.github.io/pull/1529


bretg:
You also need to modify the PrebidServerBidAdapter https://github.com/prebid/Prebid.js/blob/master/modules/prebidServerBidAdapter/index.js to add something like this in the section around line 700:

```
      if (bidUserId.parrableId) {
        request.user.ext.eids.push({
          source: 'parrable.com',
          uids: [{
            id: bidUserId.parrableId
          }]
        });
      }
```

eyas-ranjous:
Thanks @bretg . I pushed the required changes. The CircleCI build seems to be failing on something not related to our PR. I cannot restart the build. Can you please check this. Thanks.

modules/userId/index.js
```diff
@@ -369,6 +369,9 @@ function initSubmodules(submodules, consentData) {
     utils.logWarn(`${MODULE_NAME} - gdpr permission not valid for local storage or cookies, exit module`);
     return [];
   }
+
+  const submodulesThatAlwaysRefresh = ['parrableId'];
```
jaiminpanchal27: There should be an api config which states that parrableId needs refresh always. User id module should not hardcode id systems in the module. 

Please open a feature request/tech spec/separate PR for this feature. Prebid core would take a look and decide further.

test/spec/modules/userId_spec.js
```diff
@@ -14,21 +14,23 @@ import {unifiedIdSubmodule} from 'modules/userId/unifiedIdSystem';
 import {pubCommonIdSubmodule} from 'modules/userId/pubCommonIdSystem';
 import {id5IdSubmodule} from 'modules/id5IdSystem';
 import {identityLinkSubmodule} from 'modules/identityLinkIdSystem';
+import {parrableIdSubmodule} from 'modules/parrableIdSystem';
```
jaiminpanchal27: Can you add your unit tests in separate spec file ? I will open an issue to do the same for other id systems as well.

modules/userId/index.js
eyas-ranjous: makes sense. I could not locate the api config to add this list? if you could point me where it is please, then I will abstract the feature into a new PR.

test/spec/modules/userId_spec.js
eyas-ranjous: yes sure. will work on it.

eyas-ranjous:
@jaiminpanchal27 @bretg all requested changes addressed in this PR. We are going to create the refresh feature into a different PR. Please consider merging this PR if everything looks good to you. Thanks.

pm-harshad-mane:
@bretg , @jaiminpanchal27 , @eyas-ranjous 
Please check https://github.com/prebid/Prebid.js/issues/4316

eyas-ranjous:
thanks for spotting this. pushed a fix for it https://github.com/prebid/Prebid.js/pull/4317

PR #4249, opened 2019-10-01
Deprecations utils.js

Fawke:
## Type of change
- [X] Other

## Description of change

Deprecated functions in `src/utils.js`
- `getTopWindowLocation()`
- `getTopFrameReferrer()`
- `getAncestorOrigins()`
- `getTopWindowUrl()`
- `getTopWindowReferrer()`

Test cases have been updated to reflect the change. 

Adapters which were using this deprecated functions have been removed. 

## Please review but don't merge this time.

PR #4294, opened 2019-10-14
Make all adapters https

sumit116:
## Type of change

- [x] Refactoring 

## Description of change

Update all adapters' endpoints or any third party urls used by adapter to `https` protocol. 

## Other information

We didn't include those adapters in this PR that had conflict with https://github.com/prebid/Prebid.js/pull/4249 and would update them after they are submitted from the respective authors.

For some adapters which have multiple end points like test, staging, prod, etc, we have created github issues to clarify things or make changes on their own. Here is the list of those issues: 

1. https://github.com/prebid/Prebid.js/issues/4310
2. https://github.com/prebid/Prebid.js/issues/4309
3. https://github.com/prebid/Prebid.js/issues/4308
4. https://github.com/prebid/Prebid.js/issues/4307
5. https://github.com/prebid/Prebid.js/issues/4297


modules/appnexusBidAdapter.js
```diff
@@ -36,7 +36,7 @@ const SOURCE = 'pbjs';
 const MAX_IMPS_PER_REQUEST = 15;
 const mappingFileUrl = '//acdn.adnxs.com/prebid/appnexus-mapping/mappings.json';
 const SCRIPT_TAG_START = '<script';
-const VIEWABILITY_URL_START = /http:\/\/cdn\.adnxs\.com\/v/;
+const VIEWABILITY_URL_START = /https:\/\/cdn\.adnxs\.com\/v/;
```
jsnellbaker: This does not need to be changed, we actually need the regex to not include the protocol.  We need to change the protocol for this script later in the adapter code so that when it's written to the page, it'll use `https`.

When the bid returns from the adserver, the default viewability script will be using a protocol that's read from the page.  So it would come in as either `http` or `https`.  Because of this variability, the regex needs to be non-specific.

jaiminpanchal27:
@sumit116 Can you resolve conflicts. Also lets sync up offline and merge this before other merge conflicts appear.

sumit116:
> @sumit116 Can you resolve conflicts. Also lets sync up offline and merge this before other merge conflicts appear.

@jaiminpanchal27 , conflicts resolved.

patmmccann:
districtmdmx still has    obj.secure = window.location.protocol === 'https:' ? 1 : 0; so it may return nonsecure inventory even when called on a secure endpoint.

jaiminpanchal27:
@patmmccann Fair point but with this PR we are only updating the endpoints. We try not to touch other's adapter implementation unless asked to do so or if it breaks our build. 

@sumit116 to create an issue on github and tag them to make updates.

sumit116:
HI @msm0504 , can you review these changes and let me know your feedback so that we can merge it to 3.0 branch?

patmmccann:
just want to point out that https://github.com/prebid/Prebid.js/pull/4406/files made my comment above irrelevant to that particular adapter

PR #4319, opened 2019-10-16
Gamoshi: Update aliases list. Add support for userSync.

sa1omon:
## Type of change

- [X] Feature

## Description of change

- Add support for userSync using ID5 and UnifiedId.
- Remove some of our bidder aliases.


sa1omon:
> thanks - I don't see viewdeos as supported by any other adapter -- can we remove them from the prebid.org bidder page? Or will your older versions of the adapter still be supporting them?

Hi Bret,
Yes they can be removed from the bidder page. 
Please also refer to this new [PR](https://github.com/prebid/prebid.github.io/pull/1563), in which I have removed the adaports that we are not supporting anymore.

Thanks!

PR #4320, opened 2019-10-16
vuble: outstream has fullscreen option

Roffray:
## Type of change

- [x] Feature

## Description of change
Vuble adapter now offers a renderer with a fullscreen option.


bretg:
CircleCI has been flaky lately. I kicked off another run, but would still like to see a unit test here, so that'll kick off another.

Roffray:
As requested, I've added a several tests for the outstream scenario.

PR #4321, opened 2019-10-16
Ozone adapter v2.15

afsheenb:
## Type of change
- [X] Bugfix
- [X] Feature

## Description of changes

• Removal of GDPR parameters that were being read outside of the daisy-bit (which breaks requests when used on-page with certain CMP implementations that don't expose those values.)

• An onpage 'ozone' config to turn off/disable the custom key-value pairs the adapter sets should the publishers find their ad-request URL lengths are close to exceeding Google/DFP limits

• TTL update to 300 ms for bids in the adapter, up from 60 ms

• Cookie sync - append querystring values (if present) in call to cookie sync endpoint - gdpr, gdpr_consent, pubcid, publisherId, and siteId 

• Updated checks / validation for customData to ensure clients cannot send invalid customData (currently customData not passing these checks silently breaks on the server-side, now we'll throw a client-side error to avoid confusion)

- engineering@ozoneproject.com

### DIFF
--- a/modules/ozoneBidAdapter.js
+++ b/modules/ozoneBidAdapter.js
@@ -12,9 +12,9 @@ const BIDDER_CODE = 'ozone';
 // const OZONECOOKIESYNC = 'https://betalyst.local/prebid-cookie-sync-development.html';
 
 // *** DEV-ozpr
-//const OZONEURI = 'https://test.ozpr.net/openrtb2/auction';
-///const OZONECOOKIESYNC = 'https://test.ozpr.net/static/load-cookie.html';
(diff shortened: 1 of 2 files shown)


AskRupert-DM:
Thanks @jsnellbaker - will have the cleaned up code submitted shortly. 

One of the circleCI tests failed earlier - https://circleci.com/gh/prebid/Prebid.js/4309?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-build-link

The error looks to be specifically related to Edge 16 browser - how do I replicate/run the tests locally against the browserstack test suite or this specific browser instance/version locally ?

jsnellbaker:
@AskRupert-DM In looking at that error, it doesn't appear related to your adapter.  I have ran the circleci browserstack tests locally for this PR branch and all the tests pass.

Please proceed with making the updates to your adapter.

PR #4325, opened 2019-10-16
[feature] Add a config list of submodules that require refreshing the stored ID after each bid request

eyas-ranjous:
## Type of change

- [x] Feature

## Description of change
Parrable submodule requires that Parrable ID is updated frequently in the browser to enforce consumer privacy requirements and thus we generalized this requirement to a feature of submodules that can always refresh their stored id by adding the submodule name to a config list.

The list exists in constants.json as "SUBMODULES_THAT_ALWAYS_REFRESH_ID" as a key-value object, where the key is a submodule name, and the value if set to true, the submodule will always get its callback executed on each auction end event and the stored id in storage updated with the callback result.

In the unit tests, I created a mock submodule that has the same name of parrableId (because CONSTANTS are directly required inside userId index.js). and I mocked two consecutive bid requests, with an initial stored ID in the cookie. and I am asserting that the second request will get the updated value that is returned from the submodule callback.

### DIFF
--- a/modules/userId/index.js
+++ b/modules/userId/index.js
@@ -386,6 +386,10 @@ function initSubmodules(submodules, consentData) {
         refreshNeeded = storedDate && (Date.now() - storedDate.getTime() > submodule.config.storage.refreshInSeconds * 1000);
       }
 
+      if (CONSTANTS.SUBMODULES_THAT_ALWAYS_REFRESH_ID[submodule.config.name] === true) {
+        refreshNeeded = true;
+      }
+
       if (!storedId || refreshNeeded) {
         // No previously saved id.  Request one from submodule.
         response = submodule.submodule.getId(submodule.config.params, consentData, storedId);
--- a/src/constants.json
+++ b/src/constants.json
@@ -95,5 +95,8 @@
   "BID_STATUS" : {
     "BID_TARGETING_SET": "targetingSet",
     "RENDERED": "rendered"
+  },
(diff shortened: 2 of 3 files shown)


jdwieland8282:
as I understand the rational here, you want to refresh the cookie on each auction, so that if the user has made a change to their privacy preferences the module picks up on it?   Delaying the auction isn't something we should do lightly. Prebid is competing in downstream auctions, any delay means prebid risks loosing the secondary auctions.

eyas-ranjous:
@jdwieland8282 our requirement is to encrypt the device id on each impression. we are not aiming to delay any auction. The change is only about updating our cookie on after each bid request because we re-encrypt the id on each request on our server.

eyas-ranjous:
the execution of the submodule callback happens after an auction ends. and not before or during it. So the update here is kind of similar to ID5 one (who added the refresh needed first), except that they enforce the refresh based on a user config of "seconds". We do it always after auction ends to update our cookie value that will get picked up on the next bid. So the change is only about changing the encrypted id on each bid request.

jdwieland8282:
Thanks for clarifying @eyas-ranjous

justinegreene:
@bretg any chance that you can merge this before other changes require us to resolve conflicts 🙏🏻

icflournoy:
Hey @jdwieland8282 thanks for the review, @bretg looking forward to your feedback on this feature.

Parrable requires this feature for our User ID system to be fully functional.

bretg:
Asking @jlukas79 for a technical review.

Issue #4278, opened 2019-10-09
[Question] Sync userId even without AUCTION or before AUCTION_END?


pm-shashank-jain:
## Type of issue
Question

## Description
I've a scenario where publisher don't want to display ads at the first page and only load ads when user scrolls to second page. I am wondering if it is possible to have all userIds sync at first page and on second page userIds will be sent to bidders.

Currently UserId sync only happens at AUCTION_END event [https://github.com/prebid/Prebid.js/blob/f7e5de7742d8f6d8eda631ba01329c580d58edff/modules/userId/index.js#L310]. 
1. Is it possible to only call UserId and not bidders ?
2. Should there be a flag if some publisher want to sync even without auction or even before auction ?

bretg:
I think maybe the `triggerUserSyncs` function is what you're looking for?
```
pbjs.triggerUserSyncs();
```
See http://prebid.org/dev-docs/publisher-api-reference.html#setConfig-Configure-User-Syncing

Go ahead and re-open with more detail if this doesn't address the use case.

pm-shashank-jain:
Hello @bretg,
This question was regarding the UserId module (http://prebid.org/dev-docs/modules/userId.htm) not user sync calls. 
Let me know if you need more details

bretg:
Ok, so you're requesting something like `pbjs.triggerUserSyncs()` but for userId modules.

Happy to take help here.

Option 1: update pbjs.triggerUserSyncs to check for the optional UserId module and trigger those too.

Option 2: Add a new function in the User ID module, e.g pbjs.userId.triggerSync()

pm-harshad-mane:
Hello @bretg and @pm-shashank-jain ,

There is a function in userId module ``` pbjs.getUserIds() ``` it needs to be modified a bit, I am making respective changes, will update you soon.

bretg:
Closing this issue - feature now tracked in the PR

PR #4327, opened 2019-10-17
pbjs.getUserIds function can now initiate userId sub-modules

pm-harshad-mane:
## Type of change
- [X] Feature

## Description of change
Fix for https://github.com/prebid/Prebid.js/issues/4278

If ```pbjs.getUserIds``` is called before auction then the user-id sub-modules with callback were not executing, now those will execute.

old way 
```
pbjs.getUserIds() // returns an object which has user ids
```

additional new way
```
pbjs.getUserIds( callback, delayInMs ) // callback function will be called max after the specified delay and the callback will be provided with an object which has user ids 
```


pm-harshad-mane:
Hello @jaiminpanchal27 ,

pm-harshad-mane:
Thanks @bretg , I am working on the unit-test-cases and the documentation.

PR #4332, opened 2019-10-17
Fix mediaTypes

TinchoF:
## Type of change

- [x] Bugfix
- [ ] Feature
- [ ] New bidder adapter   
- [ ] Build related changes
- [ ] CI related changes
- [ ] Other

## Description of change

Fix mediaTypes


TinchoF:
Added VAST URL response

TinchoF:
Sorry, we were testing for other integrations, you can try now.

TinchoF:
Sorry about that! is fixed now.

jsnellbaker:
Hi @TinchoF 

Thanks for making that fix.  I just noticed that the bidder is not loading its own `renderer` object to help the outstream bid play on the page.  Are you expecting the publisher to supply the `renderer` in the `adUnit` when they request outstream bids from your bidder?

jsnellbaker:
@TinchoF OK, to clarify then - Prebid normally expects outstream bids to handled within a `renderer` object.  If that's not defined by the bidder when they're creating the `bidResponse`, then it's assumed to be read from the corresponding `adUnit`.

If neither of those areas provide expected `renderer` object, the bid will be rejected by Prebid core and it wouldn't win the auction.  If you're sure the custom implementation you mentioned would work with the bid in the way its created by the adapter, then I can approve of these changes.  

Please let me know about the above.  Thanks.

TinchoF:
Yes, due to the characteristics of our publishers, outstream content is not
used.

TinchoF:
Thanks!

PR #4358, opened 2019-10-22
Add userSync in onetagBidAdapter

OneTagDevOps:
## Type of change
- [X] Feature


OneTagDevOps:
Hello everyone, some checks are failing and we don't understand why. Can you please help us?

modules/onetagBidAdapter.js
```diff
@@ -162,6 +170,27 @@ function requestsToBids(bid) {
   return toRet;
 }
 
+function getUserSyncs(syncOptions, serverResponses, gdprConsent) {
+  const syncs = [];
+  if (syncOptions.iframeEnabled) {
+    const rnd = new Date().getTime();
+    let params = '?cb=' + rnd;
+
+    if (gdprConsent && typeof gdprConsent.consentString === 'string') {
```
msm0504: Please add a test with a gdprConsent param to check this `if` block

PR #4363, opened 2019-10-23
Adform adapter: add renderer for outstream bids

bratrahim:
## Type of change
- [x] Feature

## Description of change
Introduced support of inhouse outstream renderer, which is applied by default.

### DIFF
--- a/modules/adformBidAdapter.js
+++ b/modules/adformBidAdapter.js
@@ -1,8 +1,12 @@
 'use strict';
 
-import {registerBidder} from '../src/adapters/bidderFactory';
+import { registerBidder } from '../src/adapters/bidderFactory';
 import { config } from '../src/config';
 import { BANNER, VIDEO } from '../src/mediaTypes';
+import { Renderer } from '../src/Renderer';
+import * as utils from '../src/utils';
+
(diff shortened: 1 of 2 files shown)


modules/adformBidAdapter.js
```diff
@@ -112,6 +116,12 @@ export const spec = {
           vastXml: response.vast_content,
           mediaType: type
         };
+
+        if (!bid.renderer && utils.deepAccess(bid, 'mediaTypes.video.context') === 'outstream') {
```
MonokkelFredrik: @mike-chowla this if-test should have an extra condition: type === VIDEO

if (!bid.renderer && utils.deepAccess(bid, 'mediaTypes.video.context') === 'outstream' && type === VIDEO)

if you have a placement that has other mediaTypes (banner,native) than outstream it will add the rendrer either way, and cause the ad rendring to fail.

MonokkelFredrik: @bratrahim it might be you that added this code and not @mike-chowla?

bratrahim: @MonokkelFredrik You're right, that was me. Will prepare a pull request

PR #4367, opened 2019-10-23
Feature/smart video

tadam75:
## Type of change
- [x] Feature

## Description of change
In this pull request, we add the support of video media type (instream & outstream) for the Smart AdServer bid adapter.

- Here is the link to the docs repo PR : [https://github.com/prebid/prebid.github.io/pull/1568](https://github.com/prebid/prebid.github.io/pull/1568)


modules/smartadserverBidAdapter.js
```diff
@@ -44,17 +50,40 @@ export const spec = {
...
+          sizes = bid.sizes;
+        }
+        payload.sizes = sizes.map(size => ({
+          w: size[0],
+          h: size[1]
+        }));
+      } else if (videoMediaType && videoMediaType.context === 'instream') {
+        // Specific attributes for instream.
+        var playerSize = videoMediaType.playerSize[0];
+        payload.isVideo = true;
```
jsnellbaker: This variable is causing the returned `outstream` bid from your adserver to be setup as a `banner` creative (see line 126 below).  The `outstream` bid should be treated as a `video` `mediaType` and ideally have an associated renderer.

tadam75:
Hi @jsnellbaker

When we receive a request for outstream, everything is embedded in the ad-markup that we respond (the player and the content) and it can't be separated. That's why it's handled just like a banner creative.

Do you see a way to fit in your outstream video integration ?
Thanks in advance.

jsnellbaker:
Hi @tadam75 

At the moment, I'm not sure there is a way to make this type of approach work (while still classifying the `bid` response object as a proper outstream video bid).  

I've reached out to several other members of the Prebid team to get their take on this approach and any feedback they have on the direction being attempted here.  I'll try to follow-up once there's something definitive.

In the meantime, if you want to add partial instream video support for your adapter (ie drop the code parts that support outstream) - we could accept that and get it into `master`.  If so - we could open an issue on the outstream question, track that conversation separately, and then (depending on the feedback) you could open a separate PR later to add the outstream support back to the adapter.  

If you want to go this route, please let me know and make the updates to this PR.

Thanks.

tadam75:
@jsnellbaker ,

Thanks for your detailed response.

As you suggested, we removed the outstream specific code to have instream video support.

Do you want me to open the issue for our outstream case ?

jsnellbaker:
@tadam75 If you can, that would be great.  I'll look at the updates here and follow-up.

jsnellbaker:
Hi @tadam75 
The updates LGTM.  I'll go ahead an merge this PR.

Issue #4280, opened 2019-10-09
DigiTrust Id submodule fails if site intializes outside the Prebid module


goosemanjack:
## Type of issue
This is an unanticipated use case where a site has taken the DigiTrust Prebid integration, but has the older DigiTrust initialization code still in place. The site is using the Prebid integration, then explicitly calling DigiTrust.initialize instead of initializing with Prebid. That is outside our instructions and anticipated usage patterns. The mock DigiTrust object should be extended to stub the initialize method and potentially pass in the member and site Id values.

DigiTrust.initialize({member: "xxx",site: "yyy",redirects: true});

## Description
Missing "initialize" method from stub DigiTrust object.

## Steps to reproduce
Set up with Prebid DigiTrust integration, no independent library. Attempt to initialize DigiTrust outside the Prebid loader with DigiTrust.initialize(...)

## Test page
http://www.recipebridge.com/

### Expected results
No error, works

### Actual results
Exception in console DigiTrust.initialize is not a function

bretg:
Are you planning on submitting a PR for this @goosemanjack

goosemanjack:
Yes, I will be. I just needed a tracking issue to ensure it doesn't get lost. This is an edge case we didn't account for, but saw in the wild.

goosemanjack:
Touching issue. There is a PR #4369 
https://github.com/prebid/Prebid.js/pull/4369
This is a minor, isolated change and should just require a quick review.

goosemanjack:
touch prior to test update. Am traveling but will try to get code out.

goosemanjack:
Touching issue. I'm just back from vacation and getting in the groove again. Will try to resolve everything prior to end of year.

goosemanjack:
Need to round out mocks and see if unit test is practical

PR #4369, opened 2019-10-23
#4280 Addition of initialize method to DigiTrust facade

goosemanjack:
## Type of change

- [ X] Bugfix

## Description of change
Bug fix for issue #4280   
Handles legacy DigiTrust.initialize call when migrating to Prebid.

ccole@emination.com

## Other information
Issue #4280

### DIFF
--- a/integrationExamples/gpt/digitrust_min_DTinit.html
+++ b/integrationExamples/gpt/digitrust_min_DTinit.html
@@ -1,0 +1,281 @@
+<html>
+<head>
+    <title>DigiTrust Init - No Framework</title>
+
+    <script>
+        (function (window, document) {
+            if (!window.__cmp) {
+                window.__cmp = (function () {
+                    var listen = window.attachEvent || window.addEventListener;
+                    listen('message', function (event) {
+                        window.__cmp.receiveMessage(event);
+                    }, false);
+
+                    function addLocatorFrame() {
+                        if (!window.frames['__cmpLocator']) {
+                            if (document.body) {
+                                var frame = document.createElement('iframe');
+                                frame.style.display = 'none';
+                                frame.name = '__cmpLocator';
+                                document.body.appendChild(frame);
+                            } else {
+                                setTimeout(addLocatorFrame, 5);
+                            }
+                        }
+                    }
+                    addLocatorFrame();
+
+                    var commandQueue = [];
+                    var cmp = function (command, parameter, callback) {
+                        if (command === 'ping') {
+                            if (callback) {
+                                callback({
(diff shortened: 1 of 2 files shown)


modules/digiTrustIdSystem.js
```diff
@@ -102,6 +102,13 @@ function initDigitrustFacade(config) {
       callCount: 0,
       initCallback: null
     },
+    initialize: function (config) {
+      /* This method is present to catch an edge case where DigiTrust publishers transition
+       * to Prebid with minimal integration, but have not removed old DigiTrust.initialize call.
+       * It should not be used */
+      utils.logInfo('Warning - call to Digitrust.initialize within a Prebid integration. Please initialize through pbjs.configure with the userSync option');
+      initializeDigiTrust(config);
```
msm0504: What does this call to `initializeDigiTrust` do? In order to reach this part of the code, `window.DigiTrust` has already been set to the Prebid facade, correct?

integrationExamples/gpt/digitrust_min_DTinit.html
```diff
@@ -0,0 +1,281 @@
+<html>
```
msm0504: Can you add a unit test for your change to `test/spec/modules/digitrustIdSystem_spec.js`, as well?

modules/digiTrustIdSystem.js
goosemanjack: This passes the config object along to set member ID, if it hasn't been set properly. The code is designed to work if the underlying DigiTrust library is present (legacy migration or blended deployment) or if it is not present (prebid integration only). In the majority of cases this call does nothing. In some rare cases - mostly from migrations that are in-process - this call passes the member ID into the system. We saw this in the wild generating errors and wanted to fix it.

integrationExamples/gpt/digitrust_min_DTinit.html
goosemanjack: I will review and see if it is practical to do so, and go ahead if there are no ill effects.

modules/digiTrustIdSystem.js
msm0504: @goosemanjack Sorry, but I'm still not understanding this. The `initialize` function you've defined is within `fascade`. So, it looks like it will only be called if `window.DigiTrust` has been set to the `fascade`. In this case, `initializeDigiTrust` should always do nothing since `dt.isClient` will be `true`.
https://github.com/prebid/Prebid.js/blob/da652e095ffe91dc7df3d6c11c1f51999afe4c51/modules/digiTrustIdSystem.js#L311

PR #4374, opened 2019-10-24
Find gpt div based on adId to be resized instead of hardcoded adunitcode / slotelementid

ptomasroos:
## Type of change

- [x] Bugfix
- [x] Feature

## Description of change

At this moment using http://prebid.org/dev-docs/publisher-api-reference.html#module_pbjs.setTargetingForGPTAsync with customSlotMatching, secureCreatives are unsupported.

We discovered this when trying to switch to secure creatives. The reason is that it expects that adUnitCode should match the adUnitPath or slotElementId of the gpt slot.

The reason behind customSlotMatching is just the thing that we can target any div arbitrary of what the ad unit code in pbjs or what the ad unit of the gpt is named.

So instead of using these fixed values we could solve this by checking the slot's targetkeys for the hb_adid. Works for older use cases as well as supporting customSlotMatching (more flexibility)


src/secureCreatives.js
```diff
@@ -93,23 +93,25 @@ function resizeRemoteCreative({ adUnitCode, width, height }) {
...
     } else {
       return adUnitCode;
     }
   }
 
-  function getDfpElementId(adUnitCode) {
-    return find(window.googletag.pubads().getSlots().filter(isSlotMatchingAdUnitCode(adUnitCode)), slot => slot).getSlotElementId()
+  function getDfpElementId(adId) {
+    return find(window.googletag.pubads().getSlots().filter(s => {
+      return s.getTargetingMap()['hb_adid'].includes(adId);
```
jaiminpanchal27: Please use `includes` from core-js https://github.com/prebid/Prebid.js/blob/master/src/adapters/bidderFactory.js#L10

ptomasroos: > I have 2 questions here.
> 
> 1. standard adserverTargeting can be disabled if you use `pbjs.bidderSettings.adserverTargeting.standard` to configure your own. So someone may not send hb_adId and use something else. Let me know if this is invalid use case.
> 2. `getTargetingMap` function is not documented here https://developers.google.com/doubleclick-gpt/reference#googletag.slot. But I tested on page and it exists so not sure if we should use undocumented function.

Good one, we could just use the getTargeting(key) instead.

ptomasroos:
Maybe an alternative would be to check either the old way or with hb_adid?

My feeling was that adid was a requirement for rendering checking through the universal-creative test cases adId is always present.

ptomasroos:
Considering this, https://github.com/prebid/Prebid.js/blob/0edc59c55799f3a2aa2ccd58d3aa5dce0920eb81/src/secureCreatives.js#L30 i think its a fair assumption? Since the callback need to use adid in the callback?

ptomasroos:
Or we can walk through every key of the slot? Then we make sure that we find someone who might want to change the regular hb_adid to myCool_customKey instead? We just iterate every key and their value and compare it? That would solve that issue mentioning above.

jaiminpanchal27:
> Considering this,
> 
> https://github.com/prebid/Prebid.js/blob/0edc59c55799f3a2aa2ccd58d3aa5dce0920eb81/src/secureCreatives.js#L30
> 
> i think its a fair assumption? Since the callback need to use adid in the callback?

I agree. This makes sense

ptomasroos:
Nope. Just waiting for review

bretg:
@ptomasroos - would you be able to add some unit tests?

mkendall07:
Hey @ptomasroos 
Any updates on this one?

ptomasroos:
Yes I will try to get some unit tests inside this PR during this week or the next!

Sorry for the delay

ptomasroos:
Bump!

jsnellbaker:
@ptomasroos Did you get to put together the requested unit tests?  It doesn't seem like they were added to the PR if so.

ptomasroos:
Taking a look at it now as I've got some spare time

djrosenbaum:
Can confirm I am experiencing the same issue which this PR fixes. As an alternative, the below fix worked for me before realizing this PR was already open. Also takes into account a scenario where "hb_adid" could be a custom key
https://github.com/prebid/Prebid.js/pull/4550/files

ptomasroos:
@jsnellbaker I've opened a new PR with test case, #4710

PR #4376, opened 2019-10-24
Revert "Sovrn adapter updates: schain, digitrust, pixel syncing, and 3.0 upgrades"

msm0504:
Reverts prebid/Prebid.js#4335

@aprakash-sovrn,
Your changes do not work with the latest Prebid. `pixelEnabled` has been deprecated. 
Please merge the latest Prebid `master` into your fork and fix this issue before opening a new PR for these changes.

### DIFF
--- a/modules/sovrnBidAdapter.js
+++ b/modules/sovrnBidAdapter.js
@@ -1,6 +1,8 @@
 import * as utils from '../src/utils'
 import { registerBidder } from '../src/adapters/bidderFactory'
 import { BANNER } from '../src/mediaTypes'
+const errorUrl = 'https://pcb.aws.lijit.com/c'
+let errorpxls = []
 
 export const spec = {
   code: 'sovrn',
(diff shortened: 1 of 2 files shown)


jsnellbaker:
@aprakash-sovrn To clarify the above a bit further, the `pixelEnabled` field for the `spec.getUserSyncs()` has not been deprecated on `master`.  It will eventually be deprecated from the public API for `userSync`, but even then - the structure of the `syncOptions` field for the `spec.getUserSyncs()` should remain the same.

That said - since #4335 has been merged, (at least) one of the sovrn unit tests has been consistently failing across several browsers (namely Safari, Edge, and IE).

Below is a copy of the circleci test that showed the test failures for reference:
https://circleci.com/gh/prebid/Prebid.js/4492

Can you please investigate, recreate the PR with the updates to address this issue?  We're going to revert the PR for the time being.

msm0504:
@aprakash-sovrn It's possible that using `flatMap` was causing the tests to fail. That function is not supported in IE, Edge, or Safari < v12.

Please try replacing that function on line 175 of your adapter. An alternative is specified at the bottom of the `flatMap` doc: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/flatMap

aprakash-sovrn:
is pixel syncing being removed entirely from Prebid? or does it just need to happen in a different way now?

PR #4389, opened 2019-10-28
Adpod deal support

jaiminpanchal27:
## Type of change
- [x] Feature

## Description of change
This PR adds deal support to adpod. 

#### Background
Publisher requires a method to prioritize certain video deals higher in the Freewheel stack to ensure delivery and performance. They want to use the deal tier passed by bidder to help to inflate the CPM that gets passed into Freewheel to get higher priority there.

Introducing two new adpod configs 
- `prioritizeDeals` flag to give higher preference to deals and will replace cpm with `bid.video.tier` in `hb_pb_cat_dur` value.  e.g `hb_pb_cat_dur=dealId_395_15s`
- `dealTier` to set prefix and minimum deal tier for each bidder. Each Publisher can have different line item setup in adserver with different priorities. Publisher and bidders know what deal keyword to use to target line items. 

// minDealTier is an int which will give high preference to deal bids with tier >= 5
// Note: bids with tier less than 5 will not be ignored but their cache key will contain dealId in place of cpm. These bids will be auctioned just like non-deal bids.

```
// This will replace the cpm with dealId in cache key as well as targeting kv pair when prioritizeDeals flag is set to true.
pbjs.setConfig({
  adpod: {
    prioritizeDeals: true,
    dealTier: {
      'appnexus': {
        prefix: 'tier',
        minDealTier: 5
      },
      'some-other-bidder': {
        prefix: 'deals',
        minDealTier: 20
      }
    }
  }
})
```

In case of adpod, bidder returns multiple bids. Each bid can have different priority/deal tier set. Publisher may want to have control over deal tier. We are adding `filterBids` option to `
pbjs.adServers.freewheel.getTargeting` to select certain deal bids. 

``` 

pbjs.adServers.freewheel.getTargeting({
    codes: [adUnitCode1],
    callback: function(err, targeting) {
        //pass targeting to player api
    }
});
```

Following show the sample return value.
``` 
// Sample return targeting key value pairs
{
  'adUnitCode-1': [
    {
      'hb_pb_cat_dur': 'tier9_400_15s', // Bid with deal id
    },
    {
      'hb_pb_cat_dur': 'tier7_401_15s', // Bid with deal id
    },
    {
      'hb_pb_cat_dur': '15.00_402_15s',
    },
    {
      'hb_cache_id': '123'
    }
  ]
}
```

### DIFF
--- a/modules/adpod.js
+++ b/modules/adpod.js
@@ -116,7 +116,8 @@ function getPricePartForAdpodKey(bid) {
   let pricePart
   let prioritizeDeals = config.getConfig('adpod.prioritizeDeals');
   if (prioritizeDeals && utils.deepAccess(bid, 'video.dealTier')) {
-    pricePart = utils.deepAccess(bid, 'video.dealTier');
+    const adpodDealPrefix = config.getConfig(`adpod.dealTier.${bid.bidderCode}.prefix`);
+    pricePart = (adpodDealPrefix) ? adpodDealPrefix + utils.deepAccess(bid, 'video.dealTier') : utils.deepAccess(bid, 'video.dealTier');
   } else {
     const granularity = getPriceGranularity(bid.mediaType);
     pricePart = getPriceByGranularity(granularity)(bid);
@@ -474,7 +475,11 @@ export function getTargeting({codes, callback} = {}) {
       let bidDealTier = utils.deepAccess(bid, 'video.dealTier');
       let minDealTier = config.getConfig(`adpod.dealTier.${bid.bidderCode}.minDealTier`);
       if (minDealTier && bidDealTier) {
-        partitions[Number(bidDealTier > minDealTier)].push(bid);
+        if (bidDealTier >= minDealTier) {
+          partitions[1].push(bid)
+        } else {
+          partitions[0].push(bid)
+        }
       } else if (bidDealTier) {
         partitions[1].push(bid)
       } else {
--- a/modules/adpod.js
+++ b/modules/adpod.js
@@ -453,7 +453,7 @@ export function sortByPricePerSecond(a, b) {
  * @param {function} callback
  * @returns targeting kvs for adUnitCodes
  */
-export function getTargeting({codes, callback, filterBids} = {}) {
+export function getTargeting({codes, callback} = {}) {
   if (!callback) {
     utils.logError('No callback function was defined in the getTargeting call.  Aborting getTargeting().');
     return;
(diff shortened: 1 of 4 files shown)


modules/adpod.js
```diff
@@ -457,7 +467,27 @@ export function getTargeting({codes, callback} = {}) {
...
   bids = (competiveExclusionEnabled || deferCachingEnabled) ? getExclusiveBids(bids) : bids;
-  bids.sort(sortByPricePerSecond);
+
+  let prioritizeDeals = config.getConfig('adpod.prioritizeDeals');
+  if (prioritizeDeals) {
+    let [otherBids, highPriorityDealBids] = bids.reduce((partitions, bid) => {
+      let bidDealTier = utils.deepAccess(bid, 'video.dealTier');
+      let minDealTier = config.getConfig(`adpod.dealTier.${bid.bidderCode}.minDealTier`);
+      if (minDealTier && bidDealTier) {
+        partitions[Number(bidDealTier > minDealTier)].push(bid);
```
jsnellbaker: What happens here when you case the `bidDealTier` as a `Number` when a publisher defined a prefix?  Could that skew the comparison when you have `(tier7 > 5)`?

mkendall07: this line is somewhat confusing, i'd try to make it clearer or put a code comment.

modules/adpod.js
```diff
@@ -457,7 +467,27 @@ export function getTargeting({codes, callback} = {}) {
...
+      } else if (bidDealTier) {
+        partitions[1].push(bid)
+      } else {
+        partitions[0].push(bid);
+      }
+      return partitions;
+    }, [[], []]);
+    highPriorityDealBids.sort(sortByPricePerSecond);
```
jsnellbaker: To confirm - we're sorting these deal bids by the CPM?  Is the deal tier value synonymous with CPM overall speaking?  If not - should the sorting be done by the number in the deal tier rather than CPM (if I'm understanding this overall deal tier correctly)?

modules/adpod.js
jaiminpanchal27: bid.video.dealTier is an int value which all bidders need to abide.

modules/adpod.js
jaiminpanchal27: We are always sorting on price per second for deal bids. Appnexus deal tier is from 0 to 20. Consider a bidder whose deal tier vary from 0 to 100 may get unfair advantage if i sort on deal tier. 
Also I do not have any other use case due to limited adoption of adpod.

PR #4407, opened 2019-10-31
Rubicon Adapter: Add multiple sizes to sizeMap

ACannuniRP:
## Type of change
- [x] Feature

## Description of change

Add the following sizes to the Rubicon Project Adapter:
- 1x1 Floating (1x1)
- Catch All (2x2)
- Skin (2x4)
- Rubicon Custom (Native via Display) (6x6)


bretg:
@ACannuniRP - a potential problem with the use of 1x1 here... there are multiple 1x1's in the RP size list: IDs 221 and 30. Are we certain that size 30 will never be needed?

PR #4410, opened 2019-10-31
JustPremium: Update to Prebid 3.0

DarkRaven255:
## Type of change
- [x] Build related changes

## Description of change
JustPremium: Update to Prebid 3.0

### DIFF
--- a/modules/justpremiumBidAdapter.js
+++ b/modules/justpremiumBidAdapter.js
@@ -2,8 +2,8 @@ import { registerBidder } from '../src/adapters/bidderFactory'
 import { deepAccess } from '../src/utils';
 
 const BIDDER_CODE = 'justpremium'
-const ENDPOINT_URL = '//pre.ads.justpremium.com/v/2.0/t/xhr'
-const JP_ADAPTER_VERSION = '1.4'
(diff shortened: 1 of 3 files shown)


modules/justpremiumBidAdapter.js
```diff
@@ -44,7 +44,7 @@ export const spec = {
       const zone = b.params.zone
       const sizes = payload.sizes
       sizes[zone] = sizes[zone] || []
-      sizes[zone].push.apply(sizes[zone], b.sizes)
+      sizes[zone].push.apply(sizes[zone], b.mediaTypes && b.mediaTypes.banner && b.mediaTypes.banner.sizes && b.mediaTypes.banner.sizes.length && b.mediaTypes.banner.sizes)
```
harpere: I suspect this is not what you want to do.  Currently if the sizes array is empty you push that empty array `[]` onto the array, but with this change you will push `0` onto the array because `[] && [].length && []` evaluates to zero `0` (which is the value of `[].length`).

DarkRaven255: Yup, thanks for catching it! :)

Issue #4408, opened 2019-10-31
Prevent Duplicate Alias User Syncs


r-schweitzer:
## Type of issue
Bugfix

## Description
Prebid currently registers and limits userSync calls based on the bidder name. When Aliases are used, this causes duplicate sync requests as the sync URL is the same as in the original adapter. 

## Proposed Solution
This fix will instead look at how many times the URL is used instead of how many times a bidder name was used in determining how to limit syncs.

https://github.com/prebid/Prebid.js/pull/4435

bretg:
Closing this issue -- will track the issue with the PR #4435

PR #4421, opened 2019-11-05
Bugfix: Stop Duplicate Alias User Syncs

r-schweitzer:
## Type of change

- [x] Bugfix
- [ ] Feature
- [ ] New bidder adapter   
- [ ] Build related changes
- [ ] CI related changes
- [ ] Other

## Description of change
Prebid currently registers and limits userSync calls based on the bidder name. When Aliases are used, this causes duplicate sync requests as the sync URL is the same as in the original adapter. This fix will instead look at how many times the URL is used instead of how many times a bidder name was used in determining how to limit syncs.

## Other information
#4408

### DIFF
--- a/src/userSync.js
+++ b/src/userSync.js
@@ -10,7 +10,6 @@ config.setDefaults({
     syncsPerBidder: 5,
     syncDelay: 3000,
     auctionDelay: 0,
-    allowDuplicates: true
   }
 });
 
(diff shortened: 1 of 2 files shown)


bretg:
We discussed this in the Prebid.js PMC and are leaning towards considering this a bug. i.e. no need for a new configurable parameter. Andrei's going to do a confirmation with Xandr folks on this point.

Issue #4418, opened 2019-11-04
Safeframe resize fix doe not select correct element


vkimcm:
## Type of issue
bug

## Description
https://github.com/prebid/Prebid.js/pull/3988

There were still issues with above fix, our footer was not getting resized because footer div was not the last element of the parent, rather close button element is.  We need different solution that does not use last-child css.

andyblackwell:
In case you happen to use a custom build and also use GAM like us, we changed the selectors to this, which works better for us:
`['div[id^="google_ads_iframe"]', 'iframe[id^="google_ads_iframe"]']`

PR #4423, opened 2019-11-05
Safeframe resize fix does not select correct element

vkimcm:
## Type of change
- [ ] Bugfix

## Description of change
Div and iframe not getting resized when element is not the last element of the parent, for example, when there is close button element after). We need different solution that does not use last-child css.

## Other information
https://github.com/prebid/Prebid.js/issues/4418


jaiminpanchal27:
@vkimcm I added `last-child` to fix this one. More details here https://github.com/prebid/Prebid.js/pull/3988

vkimcm:
@jaiminpanchal27 this is issue with last-child, div has to be the last element of the parent but we have div for close button as the last element, so last-child css does not select the correct div

PR #4424, opened 2019-11-05
IX Adapter Video Support

ix-prebid-support:
## Type of change
- [x] Feature

## Description of change
Configured the IX adapter for the VIDEO ad type with instream and outstream context support.

pdu-supply-prebid@indexexchange.com

## Other information
Please see modules/ixBidAdapter.md for details.

### DIFF
--- a/modules/ixBidAdapter.js
+++ b/modules/ixBidAdapter.js
@@ -1,35 +1,71 @@
 import * as utils from '../src/utils';
-import { BANNER } from '../src/mediaTypes';
+import { BANNER, VIDEO } from '../src/mediaTypes';
+import find from 'core-js/library/fn/array/find';
 import { config } from '../src/config';
 import isInteger from 'core-js/library/fn/number/is-integer';
 import { registerBidder } from '../src/adapters/bidderFactory';
 
 const BIDDER_CODE = 'ix';
-const BANNER_SECURE_BID_URL = 'https://as-sec.casalemedia.com/cygnus';
-const SUPPORTED_AD_TYPES = [BANNER];
-const ENDPOINT_VERSION = 7.2;
+const SECURE_BID_URL = 'https://as-sec.casalemedia.com/cygnus';
+const SUPPORTED_AD_TYPES = [BANNER, VIDEO];
+const BANNER_ENDPOINT_VERSION = 7.2;
+const VIDEO_ENDPOINT_VERSION = 8.1;
 const CENT_TO_DOLLAR_FACTOR = 100;
-const TIME_TO_LIVE = 35;
+const BANNER_TIME_TO_LIVE = 35;
+const VIDEO_TIME_TO_LIVE = 3600; // 1hr
 const NET_REVENUE = true;
 const PRICE_TO_DOLLAR_FACTOR = {
   JPY: 1
 };
 
 /**
- * Transform valid bid request config object to impression object that will be sent to ad server.
+ * Transform valid bid request config object to banner impression object that will be sent to ad server.
  *
  * @param  {object} bid A valid bid request config object.
  * @return {object}     A impression object that will be sent to ad server.
  */
 function bidToBannerImp(bid) {
-  const imp = {};
-
-  imp.id = bid.bidId;
+  const imp = bidToImp(bid);
 
   imp.banner = {};
   imp.banner.w = bid.params.size[0];
   imp.banner.h = bid.params.size[1];
   imp.banner.topframe = utils.inIframe() ? 0 : 1;
+
+  return imp;
+}
+
+/**
+ * Transform valid bid request config object to video impression object that will be sent to ad server.
+ *
+ * @param  {object} bid A valid bid request config object.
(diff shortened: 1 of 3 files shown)


ix-prebid-support:
Hi,

We are introducing video support to the IX adapter.

We request that it is not merged until at least Tues Nov 12, 2019. 

Regards,
Dmitry
Index Exchange

ix-prebid-support:
Hello,

Please let us know if there's anything you need before moving this along.

Regards,
Dmitry
Index Exchange

robertrmartinez:
Hi I will take a look now

modules/ixBidAdapter.md
```diff
@@ -42,16 +42,21 @@ var adUnits = [{
...
 var adUnits = [{
     // ...
-
     mediaTypes: {
         banner: {
             sizes: [
                 [300, 250],
                 [300, 600]
             ]
+        },
```
robertrmartinez: FYI for `3.0` we are removing the `Old Format` support.

So the top level `adUnit.sizes` will not be supported any longer.

So maybe when we move to `3.0` we can update the ix.md file!

modules/ixBidAdapter.md
```diff
@@ -146,18 +179,92 @@ var adUnits = [{
     }]
 }];
 ```
-
+**Video (Instream):**
+```javascript
+var adUnits = [{
```
robertrmartinez: Hi,

I am not able to get a valid bid response when using this example ad unit.

Can you please update or let me know hoe to solicit a valid video ad so I can test out the changes a little more thoroughly?

ix-prebid-support: Hello,
Thanks for looking it over!
Here's an example config: 
```
var videoAdUnit = {
    code: '/21793097377/el-testo-prebid-video-vast-a',
    mediaTypes: {
        video: {
            playerSize: [
                [640, 480]
            ],
            context: 'instream'
        }
    },
    bids: [{
        bidder: 'ix',
        params: {
            siteId: '335943',
            size: [640, 480],
            video: {
                protocols: [2, 3, 5, 6],
                mimes: [
                    'video/mp4',
                    'video/webm',
                    'video/3gpp',
                    'video/x-mpegURL',
                    'application/javascript',
                    'application/x-shockwave-flash'
                ],
                minduration: 0,
                maxduration: 60,
                linearity: 1,
                api: [1, 2],
                playbackmethod: [3],
                startdelay: 0
            }
        }
    }]
};
```
We wouldn't want to return bids from examples in our doc, though, as we'd really prefer the touch point before a partner goes live.

This is largely resubmitting this PR from earlier this year: https://github.com/prebid/Prebid.js/pull/3901

Please let me know if there are further questions.

Regards,
Dmitry
Index Exchange

modules/ixBidAdapter.md
ix-prebid-support: We are planning a release specifically for 3.0 with this change. 👍

robertrmartinez:
@ix-prebid-support 
Thanks for this Dmitry.

I will give it another look over tomorrow!

modules/ixBidAdapter.js
```diff
@@ -66,15 +102,27 @@ function parseBid(rawBid, currency) {
...
+
   bid.dealId = utils.deepAccess(rawBid, 'ext.dealid');
-  bid.ttl = TIME_TO_LIVE;
   bid.netRevenue = NET_REVENUE;
   bid.currency = currency;
   bid.creativeId = rawBid.hasOwnProperty('crid') ? rawBid.crid : '-';
 
+  // in the event of a video
+  if (utils.deepAccess(rawBid, 'ext.vasturl')) {
+    bid.vastUrl = rawBid.ext.vasturl
```
robertrmartinez: Ok so you are setting the `vastUrl`.

But no `videoCacheKey`.

What this means that if there is a `cache.url` configured then it will attempt to cache your bid once again using client side caching.

This would result in your bid response to look like so:

image

Where the `hb_cache_id` and `hb_uuid` are both the client side keys returned from the cache (I used Rubicons cache in this example.

robertrmartinez: The reason I bring this up is because it all depends on how a publisher utilized Prebid Video auctions.

A lot of times it may not be a standard creative

It may use the DFP VAST creatives which need to point to a specific Vast url with a pattern match macro.

So this scenario might be a little tricky.

If the publisher has a custom setup where they just are parsing the bid responses and adding them to the VAST player, then this should probably be fine.

robertrmartinez: I just wanted to make you aware of this case and make sure you expect it or are okay with it.

I plan to merge this PR before the new release tomorrow (Tuesday).

But will give you until then to make a change or ask questions.

@ix-prebid-support

ix-prebid-support: Hi Robert - 

We're OK with it for now. Thank you for pointing it out. 

Regards,
Dmitry
Index Exchange

PR #4434, opened 2019-11-07
add support for min_height field in pbs native requests

jsnellbaker:
## Type of change

- [x] Bugfix

## Description of change
This PR adds support for a missing field related to native requests using `aspect_ratios` with the PBS adapter.  

In the ORTB Native spec, image related assets (such as image and icons) require either a set of defined sizes or a min_width **and** min_height (which are related to aspect_ratios).

The PBS adapter previously only supported a min_width field, but not a min_height field.  

Additional notes from ORTB Native spec:
https://www.iab.com/wp-content/uploads/2018/03/OpenRTB-Native-Ads-Specification-Final-1.2.pdf (see page 15 in PDF, the Image Request Object sub-section)

## Other information
Docs PR: https://github.com/prebid/prebid.github.io/pull/1596

### DIFF
--- a/modules/prebidServerBidAdapter/index.js
+++ b/modules/prebidServerBidAdapter/index.js
@@ -510,10 +510,11 @@ const OPEN_RTB_PROTOCOL = {
                   type: imgTypeId,
                   w: utils.deepAccess(params, 'sizes.0'),
                   h: utils.deepAccess(params, 'sizes.1'),
-                  wmin: utils.deepAccess(params, 'aspect_ratios.0.min_width')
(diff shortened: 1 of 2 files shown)


jsnellbaker:
@snapwich Thanks for the feedback; I grouped them like that:
- since PBS seems to use a similar type of grouping to accept the request.  PBS needed to have both min_width and min_height together if the basic sizes were absent.
- we don't really have a means to separately pass the width/height independently of each other (ie the `sizes: [200, 150]`)

I can change it around if it's going to prevent some use-case and has to be a different way.

PR #4435, opened 2019-11-07
Bugfix/stop duplicate alias user syncs

r-schweitzer:
## Type of change

- [x] Bugfix
- [ ] Feature
- [ ] New bidder adapter   
- [ ] Build related changes
- [ ] CI related changes
- [ ] Other

## Description of change
Prebid currently registers and limits userSync calls based on the bidder name. When Aliases are used, this causes duplicate sync requests as the sync URL is the same as in the original adapter. This fix will instead look at how many times the URL is used instead of how many times a bidder name was used in determining how to limit syncs.

##Other Information
#4408

### DIFF
--- a/src/adapters/bidderFactory.js
+++ b/src/adapters/bidderFactory.js
@@ -331,7 +331,7 @@ export function newBidder(spec) {
   });
 
   function registerSyncs(responses, gdprConsent, uspConsent) {
-    if (spec.getUserSyncs) {
+    if (spec.getUserSyncs && !adapterManager.aliasRegistry[spec.code]) {
       let filterConfig = config.getConfig('userSync.filterSettings');
       let syncs = spec.getUserSyncs({
         iframeEnabled: !!(config.getConfig('userSync.iframeEnabled') || (filterConfig && (filterConfig.iframe || filterConfig.all))),


src/adapters/bidderFactory.js
```diff
@@ -341,8 +341,9 @@ export function newBidder(spec) {
         if (!Array.isArray(syncs)) {
           syncs = [syncs];
         }
+        const adapter = spec.adapter === undefined ? spec.code : spec.adapter
```
msm0504: Should this condition check for `null` as well as `undefined`?

r-schweitzer: Added a null check. Better to be on the safe side.

mkendall07:
@r-schweitzer 
This change looks good but you have a conflict that needs resolved and also circle CI is failing.

monofonik:
Hi @mkendall07 @jsnellbaker I believe this PR has introduced a regression preventing user sync from working AT ALL for aliased bid adapters.

I've opened an issue here with test pages: https://github.com/prebid/Prebid.js/issues/4845

I've been trying for some time to implement some tests and a PR to prove / resolve the issue but having a lot of trouble due to the large amount of stubs in bidderFactory_spec.js. (Note this PR was approved without any test cases).

This is starting to become a major problem for us - as publishers update their Prebid version we're starting to lose our ability to frequency cap etc.

PR #4440, opened 2019-11-08
added cookies module

PascalSalesch:
# this is a draft :construction:

## Type of change
- [x] Feature

## Description of change

Added a module that sets cookies from the creative frame, or the bidResponse to the prebid frame.

## API

use the following snippet to enable:

```js
pbjs.setConfig({ cookies: {} })
```

see setConfig-params for configuration details.


jsnellbaker:
@PascalSalesch Is there a corresponding issue for this new feature/module that provides the context for why this is needed and what use-cases it's supporting?  If not, can you please supply this information so we (Prebid team) can review and better understand?

PR #4442, opened 2019-11-08
GumGum: adds schain param

susyt:
## Type of change

- [X] Feature

## Description of change

Adds support for supply chain object using IAB's Non-OpenRTB serialization format

### DIFF
--- a/modules/gumgumBidAdapter.js
+++ b/modules/gumgumBidAdapter.js
@@ -99,6 +99,28 @@ function _getDigiTrustQueryParams(userId) {
 }
 
 /**
+ * Serializes the supply chain object according to IAB standards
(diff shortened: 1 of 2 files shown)


susyt:
Hi @sumit116 , could you provide an example of where I should set schain_supported=true in the doc file? Should I add that to the description portion? I was reviewing other adapters who are now supporting schain and they didn't have any mentions of schain support in their md file.
thanks!

sumit116:
@susyt , you can refer to https://github.com/prebid/prebid.github.io/blob/master/dev-docs/bidders/pubmatic.md. This file contains what you need to add i.e `schain_supported: true`.

susyt:
@sumit116 Thanks for the link! here's the other PR for updating the doc - https://github.com/prebid/prebid.github.io/pull/1613
please let me know if there are any other changes needed

PR #4456, opened 2019-11-12
E2e mock server

sumit116:
## Type of change

- [x] Build related changes
- [x] CI related changes

## Description of change

Our circle CI builds are failing frequently, and one of the obvious reasons for this, is inconsistency in the response from the bid server(s) that we have used for our e2e-tests. In most tests, we have used appnexus server which mostly is not returning any response in the expected time resulting in timeout errors that we see in our daily workflow, here is one [example](https://circleci.com/gh/prebid/Prebid.js/4972).

Hence we came up with this solution of using a mock server which would return response from a static file based on the request it receives. For this, we would add request and response pair files (one is added in this PR), and the mock-server with the help of its middleware (prebid-request.js) will search for the matching request and send the corresponding response. This would not only make our testing environment more stable but would also make the tests run faster.

### DIFF
--- a/test/mock-server/request-middlewares/prebid-request.js
+++ b/test/mock-server/request-middlewares/prebid-request.js
@@ -65,11 +65,7 @@ module.exports = function (req, res, next) {
         });
       }
       res.type('json');
-      if (req.method === 'GET') {
-        response = 'apntag.handleCb(' + JSON.stringify(response) + ');';
-      } else {
-        response = JSON.stringify(response);
-      }
+      response = JSON.stringify(response);
       res.write(response);
       return true;
     }
--- a/test/mock-server/request-middlewares/prebid-request.js
+++ b/test/mock-server/request-middlewares/prebid-request.js
@@ -7,11 +7,9 @@
 const glob = require('glob');
 const path = require('path');
 const deepEqual = require('deep-equal');
-console.log('helloooo:::');
 
 module.exports = function (req, res, next) {
   let reqBody;
-  console.log('in middleware:::');
   try {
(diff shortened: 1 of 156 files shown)


bretg:
You just said the magic words: "testing environment more stable". I'm behind this now.

robertrmartinez:
Yeah sounds good!

Will review today

test/mock-server/request-middlewares/prebid-request.js
```diff
@@ -0,0 +1,79 @@
...
+    if (deepEqual(reqBody, expectedReqBody)) {
+      let response = file.getResponse().httpResponse.body;
+      if (Object.keys(uuidObj).length > 0) {
+        response.tags.forEach((tag, index) => {
+          tag.uuid = uuidObj.tags[index];
+        });
+      }
+      res.type('json');
+      if (req.method === 'GET') {
+        response = 'apntag.handleCb(' + JSON.stringify(response) + ');';
```
jaiminpanchal27: This is not needed for Prebid

robertrmartinez:
I too am having issues running it locally.

I think we should add instructions on how to run it locally and all the things which are required.

sumit116:
Hi @robertrmartinez , after switching to the branch you need to do:
1. `npm install`
2. `gulp e2e-test --host=test.localhost`

And it should run the banner test on all browsers. You can test the response from `mcok-server` through network logs on browserstack or using live.browserstack hitting `http://test.localhost:9999/test/pages/banner.html`.

Please let me know if you still face any issues.

PR #4468, opened 2019-11-15
add ID5 to prebid server bid adapter

smenzer:
## Type of change

- [ ] Bugfix
- [ ] Feature
- [ ] New bidder adapter   
- [ ] Build related changes
- [ ] CI related changes
- [ x ] Other

## Description of change
ID5 ID was missing from the prebid server bid adapter code

## Contact info
scott@id5.io or padurgeat@id5.io


bretg:
Could use some unit tests around the user ID stuff. We'll take that as a separate item -- there are a couple more ID systems missing here which we'll cover in a separate PR.

PR #4490, opened 2019-11-22
GumGum: adds inScreenPubID param

susyt:
## Type of change

- [X] Other

## Description of change

Adds params.inScreenPubID parameter

## Other information

cc @derberbaby 
other PR to update documentation:
https://github.com/prebid/prebid.github.io/pull/1620

### DIFF
--- a/modules/gumgumBidAdapter.js
+++ b/modules/gumgumBidAdapter.js
@@ -177,7 +177,7 @@ function buildRequests (validBidRequests, bidderRequest) {
       data.fp = params.bidfloor;
     }
     if (params.inScreen) {
-      data.t = params.inScreen;
+      (utils.isNumber(params.inScreen)) ? data.pubId = params.inScreen : data.t = params.inScreen;
       data.pi = 2;
     }
     if (params.inSlot) {
--- a/modules/gumgumBidAdapter.js
+++ b/modules/gumgumBidAdapter.js
@@ -176,8 +176,12 @@ function buildRequests (validBidRequests, bidderRequest) {
     if (params.bidfloor) {
       data.fp = params.bidfloor;
     }
+    if (params.inScreenPubID) {
+      data.pubId = params.inScreenPubID;
+      data.pi = 2;
+    }
     if (params.inScreen) {
-      (utils.isNumber(params.inScreen)) ? data.pubId = params.inScreen : data.t = params.inScreen;
+      data.t = params.inScreen;
       data.pi = 2;
     }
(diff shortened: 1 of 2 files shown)


modules/gumgumBidAdapter.js
```diff
@@ -177,7 +177,7 @@ function buildRequests (validBidRequests, bidderRequest) {
       data.fp = params.bidfloor;
     }
     if (params.inScreen) {
-      data.t = params.inScreen;
+      (utils.isNumber(params.inScreen)) ? data.pubId = params.inScreen : data.t = params.inScreen;
```
bretg: Sorry, but as I noted in the docs PR, I don't like that you're looking at the type of the value and deciding what to do with it. This will break your Prebid Server adapter.

susyt: @bretg - I've responded to the other PR but I'll put a brief descrip here:
tracking ids must be entered as strings while publisher ids must be entered as integers. it is shown in our documentation and onboarding guides
cc @derberbaby @WayneYang1

bretg: Sorry, I can't accept this solution. Prebid Server requires that parameters be either strings or integers, but not both.

I'm thinking it should be easy -- just have your users enter them as strings. They will be passed as strings to Prebid Server. Your adapter can make the determination how to treat it. e.g. Number.isInteger()

If for some reason that doesn't work, then you need to add another parameter with a specific type.

susyt: If I'm not mistaken, wouldn't passing a string into Number.isInteger consistently give us `false`?
I can regroup with the team, but adding in a new parameter has been met with a good amount of pushback and would prefer to avoid it if possible

bretg: Ok, how about isNaN then? https://stackoverflow.com/questions/175739/built-in-way-in-javascript-to-check-if-a-string-is-a-valid-number

bretg:
@susyt - here's a summary of the proposal:

- inScreen is always a string datatype
- if value contained within the string is an integer, your PBJS and PBS adapters pass it to endpoint as 'publisher ID'
- otherwise, your PBJS and PBS adapters pass to the endpoint as as 'tracking ID'

susyt:
hi @bretg 
I understood the suggestion but thank you for summarizing it. That method would not work in our case since there are tracking ids that are strings of integers (eg '123'). I've discussed it with the team and we're going to create a new param -- I'll update this PR and the other with those modifications shortly

PR #4491, opened 2019-11-22
Allow bundling without core.

vseventer:
## Type of change
- [X] Build related changes

## Description of change
This PR adds a `--[no-]core` build flag to build Prebid.js without core:
* `gulp build --no-core` bundles all modules (but not the core) in _prebid.js_.
* `gulp build --no-core --modules=audienceNetworkBidAdapter,consentManagement --bundle-name prebid-modules` bundles _audienceNetworkBidAdapter.js_ and _consentManagement.js_ in _prebid-modules.js_.

The use case here is for third-party libraries loaded onto a page where Prebid is already available. In this scenario, only modules need to be loaded. Rather than making one request per module, this PR would enable making a single request to load all applicable modules. Workflow:
1. Third-party library is loaded.
2. Is Prebid.js already available on the page?
-- YES: Single request to _prebid-modules.js_ (build with modules but without core)
-- NO (1): Single request to _prebid.js_ (build with modules and core), or:
-- NO (2): Two requests to _prebid.js_ and _prebid-modules.js_ (potentially in parallel).
3. Prebid core and modules are now available for use.

### DIFF
--- a/gulpfile.js
+++ b/gulpfile.js
@@ -180,7 +180,8 @@ function bundle(dev, moduleArr) {
     }
   }
 
-  var entries = [helpers.getBuiltPrebidCoreFile(dev)].concat(helpers.getBuiltModules(dev, modules));
+  var core = !argv.core ? [ ] : [helpers.getBuiltPrebidCoreFile(dev)];
+  var entries = core.concat(helpers.getBuiltModules(dev, modules));
 
   var outputFileName = argv.bundleName ? argv.bundleName : 'prebid.js';
 
@@ -198,6 +199,12 @@ function bundle(dev, moduleArr) {
   )
     .pipe(gulpif(dev, sourcemaps.init({loadMaps: true})))
     .pipe(concat(outputFileName))
+    .pipe(gulpif(!argv.core, header(
+      'window.<%= global %>=window.<%= global %>||{};' +
+      'window.<%= global %>.que=window.<%= global %>.que||[];' +
+      'window.<%= global %>.que.push(function(){', { global: prebid.globalVarName }
+    )))
+    .pipe(gulpif(!argv.core, footer('});')))
     .pipe(gulpif(!argv.manualEnable, footer('\n<%= global %>.processQueue();', {
       global: prebid.globalVarName
     }


mkendall07:
Do you have an example of this working? I'm suspicious it works out of the box like this but I'll page @snapwich as the build expert for comment.

vseventer:
@mkendall07 Thanks for taking a look - it seems to work pretty well actually. I span up a small demo repository showcasing how I build, and then load the core and modules in parallel: https://github.com/vseventer/prebid-demo

snapwich:
Can you give a little more description around your use case? I see a few potential problems with using Prebid in this way.

You mention `prebid-core` as being third-party code, does this mean that it is the output from a separate build process? i.e. someone else built a prebid bundle, and then you're including your _separately built_ bundle that does not already include `prebid-core`? If so, that could present versioning issues (obviously) but more importantly webpack compatibility issues. If you look at the code output from webpack you'll see that each module is using a global `prebidGlobal + 'Chunk'` name and calling it with a numerical index to register themselves with the webpack runtime (located in `prebid-core`). As far as I am aware these numerical indexes are not deterministic, meaning two separate builds could produce different indexes causing your imports to be mismatched and break.

In the case where the two separate bundles came from the _same_ build you might have more success, but you're also breaking a convention with Prebid in that it assumes by the time the `pbjs.que` is processed that all modules have been loaded. In your use case that is not true since you're wrapping the modules in `que` blocks . While it may not be a big deal for regular bid adapters it would have implications for modules that register hooks with `prebid-core`, potentially causing various race conditions and bugs. In that case this proposal could possibly work for you but not others depending on the modules they use; that makes me hesitant to add this specific change to the master build process. Although maybe there is another solution if you could provide more details around your use case?

vseventer:
Thanks for your reply @snapwich. I would love any feedback you might have on the use case below.

In this use case, there is a website ("Website") using Prebid.js for their display ads. However, Website also wants the ability to play video ads. For this purpose, Website embeds a third-party video library ("Video Player") on their page.

Video Player also uses Prebid.js to render video ads. To reduce overhead, Video Player attempts to re-use any Prebid.js instance already available on Website. It does so by reading the Website's Prebid.js version to see if it's compatible. If it is, Video Player skips loading Prebid.js core, and only downloads bid adapters (required since Prebid.js does not expose an API to detect what adapters are currently available). To make things faster, it downloads all bid adapters using a single network request.

If Website does not have Prebid.js, or an incompatible version, Video Player will download both the Prebid.js core and bid adapters in parallel. Video Player then proceeds to do the bidding, and play the winning video ad.

PR #4496, opened 2019-11-25
Advertly: New Bidder Adapter Submission

Advertly:
## Type of change

- [ ] Bugfix
- [ ] Feature
- [x] New bidder adapter   
- [ ] Build related changes
- [ ] CI related changes
- [ ] Other

## Description of change

```
{
  bidder: 'advertly',
  params: {
 publisherId: 2
  }
}
```

- A link to a PR on the docs repo at https://github.com/prebid/prebid.github.io/pull/1579## 
Other nformation


modules/advertlyBidAdapter.js
```diff
@@ -0,0 +1,129 @@
+import { registerBidder } from '../src/adapters/bidderFactory';
+import { config } from '../src/config';
+import * as utils from '../src/utils';
+import {BANNER, VIDEO} from '../src/mediaTypes';
+import { ajax } from '../src/ajax';
+import {Renderer} from '../src/Renderer';
+
+const SUPPORTED_AD_TYPES = [BANNER, VIDEO];
+const BIDDER_CODE = 'advertly';
+const DOMAIN = 'http://api.advertly.com/';
```
msm0504: Calls should be made using HTTPS

modules/advertlyBidAdapter.js
```diff
@@ -0,0 +1,129 @@
...
+  };
+}
+
+function interpretResponse(serverResponse, request) {
+  const response = serverResponse.body;
+  const bidResponses = [];
+  var bidRequestResponses = [];
+
+  utils._each(response, function(bidAd) {
+    bidAd.adResponse = {
```
msm0504: Create new objects instead of modifying response data

modules/advertlyBidAdapter.js
```diff
@@ -0,0 +1,129 @@
...
+      content: bidAd.vastXml,
+      height: bidAd.height,
+      width: bidAd.width
+    };
+    bidAd.ttl = config.getConfig('_bidderTimeout')
+    bidAd.renderer = bidAd.context === 'outstream' ? createRenderer(bidAd, {
```
msm0504: Why is `bidAd.adUnitCode` included in all 3 parameters? Are all 3 parameters needed?

Advertly: Do you mean the 3 parameters are adResponse ,ttl, renderer ?

modules/advertlyBidAdapter.js
Advertly: New object created

modules/advertlyBidAdapter.js
msm0504: I mean that all 3 parameters passed to `createRenderer` include `bidAd.adUnitCode`.
It seems like the only distinct data needed in that function is `bidAd` and `RENDERER_URL`

modules/advertlyBidAdapter.js
msm0504: That won't create new objects. `bnd` points to the same `bidAd` object as before.

`Object.assign(bnd, bidAd)` would create a shallow copy if the entire `bidAd` is needed. Otherwise, you could just set individual fields on `bnd`.

Also, `let bnd` should be within the loop instead of outside.

Advertly: Above changes are done

Advertly: Hi @msm0504 ,
We have updated the new objects like you comment.

modules/advertlyBidAdapter.js
Advertly: Hi @msm0504 ,
Now we have removed bidAd.adUnitCode with in the second parameter in createRenderer function . Can you review this once?

modules/advertlyBidAdapter.js
msm0504: `bnd = bidAd;` (line 36) is no longer needed

modules/advertlyBidAdapter.js
msm0504: `bidAd.adUnitCode` is still being passed into `createRenderer ` twice.

Why is `bidAd.adUnitCode` used on line 74 and `adUnitCode` used on line 78? Can't both use `bidAd.adUnitCode`, which would eliminate the need for a 3rd param in that function?

modules/advertlyBidAdapter.js
Advertly: We have removed bnd = bidAd; (line 36).

modules/advertlyBidAdapter.js
Advertly: if we use bidAd.adUnitCode for  line 78, we have issue on ci/circleci

msm0504: That line needs to specify the property name.

`adUnitCode: bidAd.adUnitCode`

Advertly: adUnitCode updated

modules/advertlyBidAdapter.js
msm0504: Does your server support HTTPS? It needs to per the Prebid docs:

http://prebid.org/dev-docs/bidder-adaptor.html#bidder-adaptor-Required-Adapter-Conventions

Advertly: Okay. We will upgrade our sever with SSL (HTTPS)

Advertly: We have upgrade our sever with HTTPS. And HTTPS changed in library files.

Advertly: Hi @msm0504 
We have upgrade the server to HTTPS. can you check it?

msm0504: Looks good. Thanks

PR #4498, opened 2019-11-25
logError for missing s2s adapter

jonapgar:
## Type of change

- [x] Other

## Description of change
Log error when s2s requests are created but never sent due to missing s2s adapter.

### DIFF
--- a/src/adapterManager.js
+++ b/src/adapterManager.js
@@ -324,7 +324,7 @@ adapterManager.callBids = (adUnits, bidRequests, addBidResponse, doneCb, request
         );
       }
     } else {
-      utils.logError('missing prebidServerBidAdapeter');
+      utils.logError('missing ' + _s2sConfig.adapter);
     }
   }
 
--- a/src/adapterManager.js
+++ b/src/adapterManager.js
@@ -324,7 +324,7 @@ adapterManager.callBids = (adUnits, bidRequests, addBidResponse, doneCb, request
         );
       }
     } else {
-      utils.logError('s2sAdapter "' + _s2sConfig.adapter + '" was not found in registry.');
+      utils.logError('missing prebidServerBidAdapeter');
     }
   }
 
--- a/src/adapterManager.js
+++ b/src/adapterManager.js
@@ -323,6 +323,8 @@ adapterManager.callBids = (adUnits, bidRequests, addBidResponse, doneCb, request
           s2sAjax
         );
       }
+    } else {
+      utils.logError('s2sAdapter "' + _s2sConfig.adapter + '" was not found in registry.');
     }
   }
 


bretg:
Please provide a use case... cloned your code and tried to get this error message to appear and couldn't get it. This is how I'm understanding the scenario:

- AdUnit contains bids for bidderA and bidderB
- s2sConfig.bidders contains both bidderA and bidderB
- the PBJS package only contains bidderA

The new error should appear for bidderB. Right? Not seeing it. Which brings me to the next comment: unit test please.

jonapgar:
It has nothing to do with the end bidders themselves, but the s2sConfig.adapter property.

```javascript
  pbjs.setConfig({
        s2sConfig: {
            accountId: '1',
            enabled: true,
            bidders: ['appnexus', 'pubmatic'],
            timeout: 1000,
            adapter: 'prebidServer',
            endpoint: 'https://prebid.adnxs.com/pbs/v1/openrtb2/auction',
            syncEndpoint: 'https://prebid.adnxs.com/pbs/v1/cookie_sync'
        }
    });
```
In this example if you had [built prebid.js](http://prebid.org/download.html) without the 'prebidServer' adapter included, the bid requests for s2sConfig.bidders (appnexus and pubmatic in this example) would not be sent, and no error message would be emitted.

I will add a unit test soon to demonstrate.

jonapgar:
Sorry my previous comment was unclear. You have to build prebid.js **without** prebidServerBidAdapter or simply specify a nonsense property for s2sConfig.adapter. e.g.

```javascript
     pbjs.setConfig({
        s2sConfig: {
            accountId: '1',
            enabled: true,
            bidders: ['appnexus', 'pubmatic'],
            timeout: 1000,
            adapter: 'SOME ADAPTER THAT HAS NOT BEEN REGISTERED',
            endpoint: 'https://prebid.adnxs.com/pbs/v1/openrtb2/auction',
            syncEndpoint: 'https://prebid.adnxs.com/pbs/v1/cookie_sync'
        }
    });
```

bretg:
Sorry - I misunderstood the purpose of this rather narrow message -- thought it was more general. My biggest concern about console messages is how many bytes they add to the overall package.

I'm able to get this error message now, but to be honest it doesn't seem clear to me. How about re-wording to just `ERROR: missing prebidServerBidAdapter.`

bretg:
@jonapgar - do you agree with my update to make it a more direct message?

jonapgar:
I am certainly open to changing the message though your specific suggestion would not always apply:



"ERROR: missing prebidServerBidAdapter" could be misleading because it's possible to specify an alternate bid adapter to be used as the s2s adapter in the `s2sConfig.adapter` field.

I will reword in the same format but macro in the name of the missing adapter.

Issue #4425, opened 2019-11-06
CCPA (US Privacy)


harpere:
## Type of issue
Feature

## Description
The objective is to provide a mechanism within PBJS that follows the IAB specification for complying with the CCPA law that goes into affect Jan 2, 2020.  The requirements and general approach are similar to that of GDPR, but there are some key differences.  Note that the IAB specification defines standards for a general API they call USP (US Privacy) to hopefully support future US Privacy laws in addition to the current CCPA law.  For more information see https://iabtechlab.com/standards/ccpa/

**This does not address TCF 2.0.**

### USP (CCPA) IAB Spec is not yet Complete
- the IAB CCPA tech spec public comment period ended Nov 5, so at the time of this writing, the specs are not yet final.
- the `postMessage` interface for cross-domain iframes has not yet been defined.  At this time the spec says, "the implementation must provide a proxy for postMessage events targeted to the
__uspapi interface sent from within nested iframes. GUIDANCE TO COME on iframes for
working with IAB SafeFrames."

### General Approach:
1. Generalize the existing Prebid.js GDPR Consent Management module to support both GDPR and CCPA/USPrivacy.
2. robustly gather the USP signal (we do not need to parse it)
3. make the USP signal available to each bidder
4. It's the bidder adapter's responsibility to send the signal to their endpoint
5. It's the exchange/DSP responsibility to enforce the USP rules as determined by their legal teams

### Differences with GDPR
- the USP (CCPA) api function `__uspapi()` always responds synchronously, whether or not privacy data is available, while the GDPR CMP may respond asynchronously, especially if waiting for a user response.  Unfortunately this doesn't necessarily mean it's always a synchronous transaction because if there are nested cross-domain iframes we may need to do an async postMessage().
- GDPR is "opt-out" (tracking data is assumed to be restricted unless the user opts-out), while CCPA is "opt-in" (tracking data is not restricted unless the user opts-in).

### Key Assumptions
- Because CCPA is "opt-in", we don't need to worry about an "allow_auction_without_consent" preference, when consent data could not be obtained, like we do with GDPR.
- Because the USP API does not wait for a user response, if it was not successfully obtained before the first auction, we should try again to retrieve privacy data before each subsequent auction.

### Implementation Notes and Recommendations
- The IAB Spec says, "`__uspapi()` must always be a function at all times, even at initialization – the API must be able to handle calls at all times."
[...]

PR #4502, opened 2019-11-26
CCPA additions

tjeastmond:
## Type of change

- [ ] Bugfix
- [x] Feature
- [ ] New bidder adapter   
- [ ] Build related changes
- [ ] CI related changes
- [ ] Other

## Description of change
Integrating CCPA (USPAPI) support into the existing consent module.

See https://github.com/prebid/Prebid.js/issues/4425

### DIFF
--- a/modules/consentManagement.js
+++ b/modules/consentManagement.js
@@ -6,7 +6,7 @@
  */
 import * as utils from '../src/utils';
 import { config } from '../src/config';
-import { gdprDataHandler, ccpaDataHandler } from '../src/adapterManager';
+import { gdprDataHandler, uspDataHandler } from '../src/adapterManager';
 import includes from 'core-js/library/fn/array/includes';
 import strIncludes from 'core-js/library/fn/string/includes';
 
@@ -189,19 +189,19 @@ function lookupIabConsent(cmpSuccess, cmpError, hookConfig) {
   }
 }
 
-function lookupUspConsent(ccpaSucess, cmpError, hookConfig) {
+function lookupUspConsent(uspSucess, cmpError, hookConfig) {
   function handleCmpResponseCallbacks() {
-    const ccpaResponse = {};
+    const uspResponse = {};
 
     function afterEach() {
-      if (ccpaResponse.consentString) {
-        ccpaSucess(ccpaResponse, hookConfig);
+      if (uspResponse.usPrivacy) {
+        uspSucess(uspResponse, hookConfig);
       }
     }
 
     return {
       consentDataCallback: function (consentResponse) {
-        ccpaResponse.consentString = consentResponse.consentString;
+        uspResponse.usPrivacy = consentResponse.usPrivacy;
         afterEach();
       }
     }
@@ -211,69 +211,37 @@ function lookupUspConsent(ccpaSucess, cmpError, hookConfig) {
   let uspapiCallbacks = {};
   let ccpaFunction;
 
-  // to collect the consent information from the user, we perform two calls to the CMP in parallel:
-  // first to collect the user's consent choices represented in an encoded string (via getConsentData)
-  // second to collect the user's full unparsed consent information (via getVendorConsents)
-
-  // the following code also determines where the CMP is located and uses the proper workflow to communicate with it:
-  // check to see if CMP is found on the same window level as prebid and call it directly if so
-  // check to see if prebid is in a safeframe (with CMP support)
-  // else assume prebid may be inside an iframe and use the IAB CMP locator code to see if CMP's located in a higher parent window. this works in cross domain iframes
-  // if the CMP is not found, the iframe function will call the cmpError exit callback to abort the rest of the CMP workflow
+  // the following code also determines where the USP is located and uses the proper workflow to communicate with it:
+  // check to see if USP is found on the same window level as prebid and call it directly if so
+  // check to see if prebid is in a safeframe (with USP support)
+  // else assume prebid may be inside an iframe and use the IAB USP locator code to see if USP's located in a higher parent window. this works in cross domain iframes
+  // if the USP is not found, the iframe function will call the cmpError exit callback to abort the rest of the USP workflow
   try {
     ccpaFunction = window.__uspapi || utils.getWindowTop().__uspapi;
   } catch (e) { }
 
   if (utils.isFn(ccpaFunction)) {
-    ccpaFunction('getConsentData', null, callbackHandler.consentDataCallback);
-  } else if (inASafeFrame() && typeof window.$sf.ext.uspapi === 'function') {
-    callCcpaWhileInSafeFrame('getConsentData', callbackHandler.consentDataCallback);
+    ccpaFunction('getUSPData', 1, callbackHandler.consentDataCallback);
   } else {
     // find the CMP frame
     let f = window;
-    let ccpaFrame;
-    while (!ccpaFrame) {
+    let uspFrame;
+    while (!uspFrame) {
       try {
-        if (f.frames['__uspapiLocator']) ccpaFrame = f;
+        if (f.frames['__uspapiLocator']) uspFrame = f;
       } catch (e) { }
       if (f === window.top) break;
       f = f.parent;
     }
 
-    if (!ccpaFrame) {
-      return cmpError('CCPA not found.', hookConfig);
+    if (!uspFrame) {
+      return cmpError('USP not found.', hookConfig);
     }
 
-    callCcpaWhileInIframe('getConsentData', ccpaFrame, callbackHandler.consentDataCallback);
+    callCcpaWhileInIframe('getUSPData', uspFrame, callbackHandler.consentDataCallback);
   }
 
-  function inASafeFrame() {
-    return !!(window.$sf && window.$sf.ext);
-  }
-
-  function callCcpaWhileInSafeFrame(commandName, callback) {
-    function sfCallback(msgName, data) {
-      if (msgName === 'cmpReturn') {
-        let responseObj = (commandName === 'getConsentData') ? data.vendorConsentData : data.vendorConsents;
-        callback(responseObj);
-      }
-    }
-
-    // find sizes from adUnits object
-    let adUnits = hookConfig.adUnits;
-    let width = 1;
-    let height = 1;
-    if (Array.isArray(adUnits) && adUnits.length > 0) {
-      let sizes = utils.getAdUnitSizes(adUnits[0]);
-      width = sizes[0][0];
-      height = sizes[0][1];
-    }
-
-    window.$sf.ext.register(width, height, sfCallback);
-    window.$sf.ext.ccpa(commandName);
-  }
-
-  function callCcpaWhileInIframe(commandName, ccpaFrame, moduleCallback) {
+  function callCcpaWhileInIframe(commandName, uspFrame, moduleCallback) {
     /* Setup up a __ccpa function to do the postMessage and stash the callback.
       This function behaves (from the caller's perspective identicially to the in-frame __ccpa call */
     window.__uspapi = function (cmd, ver, callback) {
@@ -286,7 +254,7 @@ function lookupUspConsent(ccpaSucess, cmpError, hookConfig) {
         }
       };
       uspapiCallbacks[callId] = callback;
-      ccpaFrame.postMessage(msg, '*');
+      uspFrame.postMessage(msg, '*');
     };
 
     /** when we get the return message, call the stashed callback */
@@ -400,14 +368,14 @@ function processCmpData(consentObject, hookConfig) {
   }
 }
 
-function processCcpaData (consentObject, hookConfig) {
-  if (!(consentObject && consentObject.consentString)) {
-    cmpFailed(`CCPA returned unexpected value during lookup process.`, hookConfig, consentObject);
+function processCcpaData(consentObject, hookConfig) {
+  if (!(consentObject && consentObject.usPrivacy)) {
+    cmpFailed(`USP returned unexpected value during lookup process.`, hookConfig, consentObject);
     return;
   }
 
   clearTimeout(hookConfig.timer);
-  storeCcpaConsentData(consentObject);
+  storeUspConsentData(consentObject);
   exitModule(null, hookConfig);
 }
 
@@ -447,11 +415,11 @@ function storeConsentData(cmpConsentObject) {
   gdprDataHandler.setConsentData(consentData);
 }
 
-function storeCcpaConsentData(consentObject) {
+function storeUspConsentData(consentObject) {
   consentData = {
-    consentString: consentObject ? consentObject.consentString : undefined
+    usPrivacy: consentObject ? consentObject.usPrivacy : undefined
   };
-  ccpaDataHandler.setConsentData(consentData);
+  uspDataHandler.setConsentData(consentData);
 }
 
 /**
@@ -503,7 +471,7 @@ function exitModule(errMsg, hookConfig, extraArgs) {
 export function resetConsentData() {
   consentData = undefined;
   gdprDataHandler.setConsentData(null);
-  ccpaDataHandler.setConsentData(null);
+  uspDataHandler.setConsentData(null);
 }
 
 /**
--- a/modules/consentManagement.js
+++ b/modules/consentManagement.js
@@ -12,6 +12,7 @@ import strIncludes from 'core-js/library/fn/string/includes';
 
 const DEFAULT_CMP = 'iab';
 const DEFAULT_CONSENT_TIMEOUT = 10000;
+const DEFAULT_CONSENT_TIMEOUT_USP = 50;
 const DEFAULT_ALLOW_AUCTION_WO_CONSENT = true;
 
 export let userCMP;
@@ -22,19 +23,16 @@ export let staticConsentData;
 let consentData;
 let addedConsentHook = false;
 
-// ccpa constants
-export let userCCPA;
-export let consentTimeoutCCPA;
-
-// ccpa globals
-let consentDataCCPA;
-let addedConsentHookCCPA = false;
+// usp
+export let consentTimeoutUSP;
+let addedConsentHookUSP = false;
 
 // add new CMPs here, with their dedicated lookup function
 const cmpCallMap = {
   'iab': lookupIabConsent,
   'gdpr': lookupIabConsent,
-  'ccpa': lookupCcpaConsent,
+  'ccpa': lookupUspConsent,
+  'usp': lookupUspConsent,
   'static': lookupStaticConsentData
 };
 
@@ -192,7 +190,7 @@ function lookupIabConsent(cmpSuccess, cmpError, hookConfig) {
   }
 }
 
-function lookupCcpaConsent(ccpaSucess, cmpError, hookConfig) {
+function lookupUspConsent(ccpaSucess, cmpError, hookConfig) {
   function handleCmpResponseCallbacks() {
     const ccpaResponse = {};
 
@@ -332,13 +330,13 @@ export function requestCcpaBidsHook(next, reqBidsConfigObj) {
  * @param {object} reqBidsConfigObj required; This is the same param that's used in pbjs.requestBids.
  * @param {function} fn required; The next function in the chain, used by hook.js
  */
-export function requestBidsHook(fn, reqBidsConfigObj, isCCPA = false) {
+export function requestBidsHook(fn, reqBidsConfigObj, isUSP = false) {
   let userModule = userCMP;
   let processFn = processCmpData;
 
-  if (isCCPA) {
-    userModule = 'ccpa';
-    processFn = processCcpaData; // @TJ
+  if (isUSP) {
+    userModule = 'usp';
+    processFn = processCcpaData;
   }
 
   // preserves all module related variables for the current auction instance (used primiarily for concurrent auctions)
@@ -367,10 +365,11 @@ export function requestBidsHook(fn, reqBidsConfigObj, isCCPA = false) {
 
   // only let this code run if module is still active (ie if the callbacks used by CMPs haven't already finished)
   if (!hookConfig.haveExited) {
-    if (consentTimeout === 0) {
+    let timeout = userModule === 'usp' ? consentTimeoutUSP : consentTimeout;
+    if (timeout === 0) {
       processFn(undefined, hookConfig);
     } else {
-      hookConfig.timer = setTimeout(cmpTimedOut.bind(null, hookConfig), consentTimeout);
+      hookConfig.timer = setTimeout(cmpTimedOut.bind(null, hookConfig), timeout);
     }
   }
 }
@@ -520,11 +519,18 @@ export function setConsentConfig(config, consentModule) {
     utils.logInfo(`consentManagement config did not specify cmp.  Using system default setting (${DEFAULT_CMP}).`);
   }
 
-  if (utils.isNumber(config.timeout)) {
+  if (utils.isNumber(config.timeout))  {
     consentTimeout = config.timeout;
   } else {
     consentTimeout = DEFAULT_CONSENT_TIMEOUT;
     utils.logInfo(`consentManagement config did not specify timeout.  Using system default setting (${DEFAULT_CONSENT_TIMEOUT}).`);
+  }
+
+  if (utils.isNumber(config.uspTimeout)) {
+    consentTimeoutUSP = config.uspTimeout
+  } else {
+    consentTimeoutUSP = DEFAULT_CONSENT_TIMEOUT_USP;
+    utils.logInfo(`consentManagement config did not specify timeout f or USP. Using system default setting: (${DEFAULT_CONSENT_TIMEOUT_USP}).`);
   }
 
   if (typeof config.allowAuctionWithoutConsent === 'boolean') {
@@ -545,12 +551,12 @@ export function setConsentConfig(config, consentModule) {
     }
   }
 
-  if (consentModule === 'ccpa' && !addedConsentHookCCPA) {
+  if (consentModule === 'usp' && !addedConsentHookUSP) {
     $$PREBID_GLOBAL$$.requestBids.before(requestCcpaBidsHook, 50);
-    addedConsentHookCCPA = true;
+    addedConsentHookUSP = true;
   }
 
-  if (!addedConsentHook && consentModule !== 'ccpa') {
+  if (!addedConsentHook && consentModule !== 'usp') {
     $$PREBID_GLOBAL$$.requestBids.before(requestBidsHook, 50);
     addedConsentHook = true;
   }
--- a/modules/consentManagement.js
+++ b/modules/consentManagement.js
@@ -189,13 +189,13 @@ function lookupIabConsent(cmpSuccess, cmpError, hookConfig) {
   }
 }
 
-function lookupUspConsent(uspSucess, cmpError, hookConfig) {
+function lookupUspConsent(uspSuccess, cmpError, hookConfig) {
   function handleCmpResponseCallbacks() {
     const uspResponse = {};
 
     function afterEach() {
       if (uspResponse.usPrivacy) {
-        uspSucess(uspResponse, hookConfig);
+        uspSuccess(uspResponse, hookConfig);
       }
     }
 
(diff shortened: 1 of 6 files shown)


bretg:
Am working on the doc updates...

bretg:
@tjeastmond - we need an example of how a bidder adapter reads these values. Please either document in the ticket or add an example for how other adapters should be reading us_privacy off the bidRequest. (i.e. rubiconBidAdapter)

bretg:
@tjeastmond and @harpere - I think we're missing an important feature here -- usersync support.

See the details for GDPR at http://prebid.org/dev-docs/modules/consentManagement.html#usersync-integration

modules/consentManagement.js
```diff
@@ -182,6 +192,138 @@ function lookupIabConsent(cmpSuccess, cmpError, hookConfig) {
...
+
+    function afterEach() {
+      if (ccpaResponse.consentString) {
+        ccpaSucess(ccpaResponse, hookConfig);
+      }
+    }
+
+    return {
+      consentDataCallback: function (consentResponse) {
+        ccpaResponse.consentString = consentResponse.consentString;
```
harpere: there's no `consentString` in the uspapi response.

modules/consentManagement.js
```diff
@@ -182,6 +192,138 @@ function lookupIabConsent(cmpSuccess, cmpError, hookConfig) {
...
+
+    // call ccpa
+    window.__uspapi(commandName, 1, ccpaIframeCallback);
+
+    function readPostMessageResponse(event) {
+      let res = typeof event.data === 'string' ? JSON.parse(event.data) : event.data;
+      if (res.__uspapiReturn && res.__uspapiReturn.callId) {
+        let i = res.__uspapiReturn;
+        if (typeof uspapiCallbacks[i.callId] !== 'undefined') {
+          uspapiCallbacks[i.callId](i.returnValue, i.success);
```
harpere: correctly passing both the `returnValue` and `success` params, but the callback only handles the first param.

modules/consentManagement.js
```diff
@@ -182,6 +192,138 @@ function lookupIabConsent(cmpSuccess, cmpError, hookConfig) {
...
+          delete uspapiCallbacks[i.callId];
+        }
+      }
+    }
+
+    function removePostMessageListener() {
+      window.removeEventListener('message', readPostMessageResponse, false);
+    }
+
+    function ccpaIframeCallback(consentObject) {
```
harpere: the callback has 2 params, `returnValue` and `success`, but only handling one here.

modules/consentManagement.js
```diff
@@ -190,7 +332,15 @@ function lookupIabConsent(cmpSuccess, cmpError, hookConfig) {
  * @param {object} reqBidsConfigObj required; This is the same param that's used in pbjs.requestBids.
  * @param {function} fn required; The next function in the chain, used by hook.js
  */
-export function requestBidsHook(fn, reqBidsConfigObj) {
+export function requestBidsHook(fn, reqBidsConfigObj, isCCPA = false) {
+  let userModule = userCMP;
+  let processFn = processCmpData;
+
+  if (isCCPA) {
+    userModule = 'ccpa';
```
harpere: what's this?

modules/consentManagement.js
```diff
@@ -251,6 +402,17 @@ function processCmpData(consentObject, hookConfig) {
   }
 }
 
+function processCcpaData (consentObject, hookConfig) {
+  if (!(consentObject && consentObject.consentString)) {
```
harpere: there's no `consentString` in the uspapi

modules/consentManagement.js
```diff
@@ -182,6 +192,138 @@ function lookupIabConsent(cmpSuccess, cmpError, hookConfig) {
...
+  // first to collect the user's consent choices represented in an encoded string (via getConsentData)
+  // second to collect the user's full unparsed consent information (via getVendorConsents)
+
+  // the following code also determines where the CMP is located and uses the proper workflow to communicate with it:
+  // check to see if CMP is found on the same window level as prebid and call it directly if so
+  // check to see if prebid is in a safeframe (with CMP support)
+  // else assume prebid may be inside an iframe and use the IAB CMP locator code to see if CMP's located in a higher parent window. this works in cross domain iframes
+  // if the CMP is not found, the iframe function will call the cmpError exit callback to abort the rest of the CMP workflow
+  try {
+    ccpaFunction = window.__uspapi || utils.getWindowTop().__uspapi;
```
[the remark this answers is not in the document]
tjeastmond: I thought the same, but I believe this was written this way in case `__uspapiLocator` was on the same level.

modules/consentManagement.js
```diff
@@ -182,6 +192,138 @@ function lookupIabConsent(cmpSuccess, cmpError, hookConfig) {
...
+    let height = 1;
+    if (Array.isArray(adUnits) && adUnits.length > 0) {
+      let sizes = utils.getAdUnitSizes(adUnits[0]);
+      width = sizes[0][0];
+      height = sizes[0][1];
+    }
+
+    window.$sf.ext.register(width, height, sfCallback);
+    window.$sf.ext.ccpa(commandName);
+  }
```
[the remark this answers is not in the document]
tjeastmond: Removed

modules/consentManagement.js
```diff
@@ -182,6 +192,138 @@ function lookupIabConsent(cmpSuccess, cmpError, hookConfig) {
...
+  let uspapiCallbacks = {};
+  let ccpaFunction;
+
+  // to collect the consent information from the user, we perform two calls to the CMP in parallel:
+  // first to collect the user's consent choices represented in an encoded string (via getConsentData)
+  // second to collect the user's full unparsed consent information (via getVendorConsents)
+
+  // the following code also determines where the CMP is located and uses the proper workflow to communicate with it:
+  // check to see if CMP is found on the same window level as prebid and call it directly if so
+  // check to see if prebid is in a safeframe (with CMP support)
```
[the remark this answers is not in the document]
tjeastmond: Removed

modules/consentManagement.js
tjeastmond: This is just what I named the value holding the USP string in temp storage. I'll change it to something like `usPrivacy`. However, it should have no impact on the implementation.

modules/consentManagement.js
```diff
@@ -182,6 +192,138 @@ function lookupIabConsent(cmpSuccess, cmpError, hookConfig) {
...
+  // else assume prebid may be inside an iframe and use the IAB CMP locator code to see if CMP's located in a higher parent window. this works in cross domain iframes
+  // if the CMP is not found, the iframe function will call the cmpError exit callback to abort the rest of the CMP workflow
+  try {
+    ccpaFunction = window.__uspapi || utils.getWindowTop().__uspapi;
+  } catch (e) { }
+
+  if (utils.isFn(ccpaFunction)) {
+    ccpaFunction('getConsentData', null, callbackHandler.consentDataCallback);
+  } else if (inASafeFrame() && typeof window.$sf.ext.uspapi === 'function') {
+    callCcpaWhileInSafeFrame('getConsentData', callbackHandler.consentDataCallback);
```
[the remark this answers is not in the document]
tjeastmond: The docs mentioned [HERE](https://iabtechlab.com/wp-content/uploads/2019/11/US-Privacy-USER-SIGNAL-API-SPEC-v1.0.pdf) safe frames, but I have removed them anyway as per our spec.

modules/consentManagement.js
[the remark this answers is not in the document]
tjeastmond: In the linked docs, but we aren't using them per the spec 👍 

https://iabtechlab.com/wp-content/uploads/2019/11/US-Privacy-USER-SIGNAL-API-SPEC-v1.0.pdf

modules/consentManagement.js
```diff
@@ -182,6 +192,138 @@ function lookupIabConsent(cmpSuccess, cmpError, hookConfig) {
...
+  } else {
+    // find the CMP frame
+    let f = window;
+    let ccpaFrame;
+    while (!ccpaFrame) {
+      try {
+        if (f.frames['__uspapiLocator']) ccpaFrame = f;
```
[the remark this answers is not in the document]
tjeastmond: I used this function for `Emily` and `multi-sync`:

```js
function getConsentWindowFor(loc) {
    var win = window;
    while (win !== window.top) {
        win = win.parent;
        if (win.frames[loc]) return win.frames[loc];
    }
    return false;
}
```

I will implement.

modules/consentManagement.js
tjeastmond: A workaround for the global params. I could probably refactor it away.

modules/consentManagement.js
tjeastmond: I don't think I see what you're talking about. Maybe a commit moved the line?

modules/consentManagement.js
```diff
@@ -336,18 +503,26 @@ function exitModule(errMsg, hookConfig, extraArgs) {
...
+      userCMP = 'iab';
+    }
+
+    if (consentModule === 'static') {
+      userCMP = 'static';
+    }
+
+    if (!['gdpr', 'iab', 'static', 'usp'].includes(consentModule) && consentModule) {
+      userCMP = consentModule;
+    }
```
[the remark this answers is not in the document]
harpere: and if we do want to do this, `if (!['gdpr', 'iab', 'static', 'usp']...` overwrites the first 2 `if` conditions.

modules/consentManagement.js
```diff
@@ -374,9 +556,20 @@ export function setConsentConfig(config) {
...
     $$PREBID_GLOBAL$$.requestBids.before(requestBidsHook, 50);
+    addedConsentHook = true;
   }
-  addedConsentHook = true;
 }
-config.getConfig('consentManagement', config => setConsentConfig(config.consentManagement));
+
+config.getConfig('consentManagement', config => {
+  const consentManagement = { ...config.consentManagement };
+  const consentChecks = consentManagement.consentAPIs ? new Set([...consentManagement.consentAPIs]) : new Set([]);
```
harpere: we need to be backward compatible, so even if `consentAPIs` does not exist, we still need to support `gdpr`. 
 So does `gdpr` get set somewhere?

modules/consentManagement.js
```diff
@@ -374,9 +556,20 @@ export function setConsentConfig(config) {
...
+
+  if (consentModule === 'usp' && !addedConsentHookUSP) {
+    $$PREBID_GLOBAL$$.requestBids.before(requestCcpaBidsHook, 50);
+    addedConsentHookUSP = true;
+  }
+
+  if (!addedConsentHook && consentModule === 'gdpr') {
     $$PREBID_GLOBAL$$.requestBids.before(requestBidsHook, 50);
+    addedConsentHook = true;
   }
```
harpere: I believe this will run these `synchronously`.  We want them to run `asynchronously`.  We don't want to wait for the `gdpr` api to finish before calling the `usp` api (or vice-versa).

modules/consentManagement.js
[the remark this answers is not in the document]
tjeastmond: The last check is for a silly test that allows non-sense to be sent in. Technically I block that now.

If I leave the existing logic it will always instantiate GDPR

modules/consentManagement.js
tjeastmond: I actually showed you this earlier. If `config.consentManagement.cmpApi`is set, I was adding `GDPR` into the set. I took it out because you mentioned just relying on the `consentAPIs` array.

modules/consentManagement.js
tjeastmond: Technically I am calling this method for each time I loop through the array set I build at the bottom of the page. They should be running async.

modules/consentManagement.js
```diff
@@ -190,7 +329,15 @@ function lookupIabConsent(cmpSuccess, cmpError, hookConfig) {
  * @param {object} reqBidsConfigObj required; This is the same param that's used in pbjs.requestBids.
  * @param {function} fn required; The next function in the chain, used by hook.js
  */
-export function requestBidsHook(fn, reqBidsConfigObj) {
+export function requestBidsHook(fn, reqBidsConfigObj, isUSP = false) {
+  let userModule = userCMP;
```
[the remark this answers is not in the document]
tjeastmond: See my comment below your comment below.

bretg:
Docs PR https://github.com/prebid/prebid.github.io/pull/1628

modules/consentManagement.js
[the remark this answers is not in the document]
harpere: I don't understand this reasoning.

modules/consentManagement.js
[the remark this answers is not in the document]
tjeastmond: Removed all the other type lookups but left the old logic for finding the proper iFrame for now. I will need to tweak tests.

modules/consentManagement.js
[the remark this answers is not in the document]
harpere: hadn't seen that, but it says `SafeFrame implementations should either allow post messages or implement a proxy __uspapi() interface`, which means to me that we don't need to worry about the safeframe specific interface.

modules/consentManagement.js
```diff
@@ -336,18 +503,26 @@ function exitModule(errMsg, hookConfig, extraArgs) {
...
 }
 
 /**
  * A configuration function that initializes some module variables, as well as add a hook into the requestBids function
  * @param {object} config required; consentManagement module config settings; cmp (string), timeout (int), allowAuctionWithoutConsent (boolean)
  */
-export function setConsentConfig(config) {
-  if (utils.isStr(config.cmpApi)) {
-    userCMP = config.cmpApi;
-  } else {
```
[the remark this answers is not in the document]
tjeastmond: As mentioned in other comments, I had mentioned to you that I had it left in originally for backward compatibility, however, I took it out after I thought we agreed that they publishers and modules would have to use the `consentAPIs` array. I can add this back.

modules/consentManagement.js
harpere: I believe `ccpaIframeCallback()` is the function called from `uspapiCallbacks[i.callId](i.returnValue, i.success);` above.   We shouldn't be trying to process the return value if it wasn't successful.

modules/consentManagement.js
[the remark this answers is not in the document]
harpere: I missed the `!` in the last `if`

modules/consentManagement.js
harpere: may have been some confusion.  But i don't think `cmpApi` is required because it defaults to 'iab'.  just want to make sure it defaults to gdpr if `consentAPIs` is not included.

modules/consentManagement.js
harpere: ok, I thought the hooks api did these synchronously, but Isaac said the requestBids hook is configured to be asynchronous.

modules/consentManagement.js
```diff
@@ -182,6 +189,101 @@ function lookupIabConsent(cmpSuccess, cmpError, hookConfig) {
...
+    } catch (e) { }
+    if (f === window.top) break;
+    f = f.parent;
+  }
+
+  if (!uspFrame) {
+    return cmpError('USP not found.', hookConfig);
+  }
+
+  callUspWhileInIframe('getUSPData', uspFrame, callbackHandler.consentDataCallback);
```
harpere: what happened to the direct `__uspapi()` call?  if it's available we should call the function directly so we don't have to wait for a `postMessage`

modules/consentManagement.js
```diff
@@ -182,6 +189,101 @@ function lookupIabConsent(cmpSuccess, cmpError, hookConfig) {
...
+          delete uspapiCallbacks[i.callId];
+        }
+      }
+    }
+
+    function removePostMessageListener() {
+      window.removeEventListener('message', readPostMessageResponse, false);
+    }
+
+    function uspIframeCallback(consentObject) {
```
harpere: as mentioned in the comment for line 266, there are two arguments passed, but this is only handling one of them.

harpere: depending on how you handle it, you may also want to pass the second argument to `moduleCallback()`

modules/consentManagement.js
```diff
@@ -182,6 +189,101 @@ function lookupIabConsent(cmpSuccess, cmpError, hookConfig) {
...
+
+    function afterEach() {
+      if (uspResponse.usPrivacy) {
+        uspSuccess(uspResponse, hookConfig);
+      }
+    }
+
+    return {
+      consentDataCallback: function (consentResponse) {
+        uspResponse.usPrivacy = consentResponse.usPrivacy;
```
harpere: as per the related comments in this PR, the `uspapi` passes a second argument to the callback, which should be handled here or in the other function in the chain.

modules/consentManagement.js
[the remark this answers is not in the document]
harpere: 1. must have been a communication problem. we need to be backward compatible.  
2. but I can see the confusion between the similarly named `cmpApi` and `consentAPIs` params.  but these are not really the same thing.  there could be different gdpr CMP implementations.  `iab` is the default.  presumably `ccpa` could have other implementations as well.  the `iab` `usp` implementation and others.  

So maybe we rename it to `consentTypes`?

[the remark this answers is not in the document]
harpere: or maybe we change it altogether to:
```
consentManagement: {
    gdpr: { <gdpr params> },
    usp: { <uspapi params> }
}
```
then to be backward compatible could do something like:
```
let cm = consentManagement;
let gdprConfig = cm.usp || cm.gdpr ? cm.gdpr : cm;
let uspConfig = cm.usp;
```

modules/consentManagement.js
```diff
@@ -182,6 +189,101 @@ function lookupIabConsent(cmpSuccess, cmpError, hookConfig) {
...
+    return cmpError('USP not found.', hookConfig);
+  }
+
+  callUspWhileInIframe('getUSPData', uspFrame, callbackHandler.consentDataCallback);
+
+  if (!uspFrame) {
+    return cmpError('USP frame not found.', hookConfig);
+  }
+
+  callUspWhileInIframe('getUSPData', uspFrame, callbackHandler.consentDataCallback);
```
harpere: this is a duplicate of the code above.

modules/consentManagement.js
harpere: this would work I think:
```
try {
    uspFrame.__uspapi('getUSPData', 1, callbackHandler.consentDataCallback);
} catch () {
    callUspWhileInIframe('getUSPData', uspFrame, callbackHandler.consentDataCallback);
}
```
(check my syntax of course)

modules/consentManagement.js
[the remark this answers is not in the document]
harpere: I think I like the second approach.  I don't think it changes much at all.  if there's a `gdprConfig`, process gdpr as is done now.  if there's a `uspConfig`, process it.  

thoughts?

modules/consentManagement.js
```diff
@@ -374,9 +519,20 @@ export function setConsentConfig(config) {
...
+
+  if (consentModule === 'usp' && !addedConsentHookUSP) {
+    $$PREBID_GLOBAL$$.requestBids.before(requestUspBidsHook, 50);
+    addedConsentHookUSP = true;
+  }
+
+  if (!addedConsentHook && consentModule === 'gdpr') {
     $$PREBID_GLOBAL$$.requestBids.before(requestBidsHook, 50);
+    addedConsentHook = true;
   }
```
harpere: I don't think this handles the timeouts as is described in the spec (give each the maximum time possible), but a deal breaker.

modules/consentManagementUsp.js
```diff
@@ -0,0 +1,309 @@
...
+  exitModule(errMsg, hookConfig, extraArgs);
+}
+
+/**
+ * Stores USP data locally in module and then invokes uspDataHandler.setConsentData() to make information available in adaptermanger.js for later in the auction
+ * @param {object} cmpConsentObject required; an object representing user's consent choices (can be undefined in certain use-cases for this function only)
+ */
+function storeUspConsentData(consentObject) {
+  consentData = { usPrivacy: consentObject ? consentObject.usPrivacy : undefined };
+  uspDataHandler.setConsentData(consentData);
```
harpere: I don't think this first line works (`usPrivacy: consentObject`).  The entire thing should probably just be changed to
```
if (consentObject && consentObject.usPrivacy) {
   uspDataHandler.setConsentData(consentObject.usPrivacy);
}
```
no need to set it if it doesn't exist

modules/consentManagementUsp.js
```diff
@@ -0,0 +1,309 @@
...
+ * @param {object} hookConfig contains module related variables (see comment in requestBidsHook function)
+ */
+function lookupUpsConsent(uspSuccess, uspError, hookConfig) {
+  function handleUspApiResponseCallbacks() {
+    const uspResponse = {};
+
+    function afterEach() {
+      if (uspResponse.usPrivacy) {
+        uspSuccess(uspResponse, hookConfig);
+      }
```
harpere: this only moves on if there's no `usPrivacy` param.  otherwise it's left waiting for the timeout to finish.  If the callback is called, we need to move on whether we have a valid response or not.

modules/consentManagementUsp.js
```diff
@@ -0,0 +1,309 @@
...
+    }
+
+    return {
+      consentDataCallback: consentResponse => {
+        uspResponse.usPrivacy = consentResponse.usPrivacy;
```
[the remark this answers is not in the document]
harpere: the relevant `uspData` object parameter is `uspString`, not `usPrivacy`

modules/consentManagementUsp.js
```diff
@@ -0,0 +1,309 @@
...
+import * as utils from '../src/utils';
+import { config } from '../src/config';
+import { uspDataHandler } from '../src/adapterManager';
+import includes from 'core-js/library/fn/array/includes';
+
+const DEFAULT_CONSENT_API = 'uspapi';
+const DEFAULT_CONSENT_TIMEOUT = 50;
+const DEFAULT_ALLOW_AUCTION_WO_CONSENT = true;
+
+export let userUSP;
```
harpere: a more accurate name for this would be `consentAPI`

modules/consentManagementUsp.js
```diff
@@ -0,0 +1,309 @@
...
+    haveExited: false,
+    timer: null
+  };
+
+  // in case we already have consent (eg during bid refresh)
+  if (consentData) {
+    return exitModule(null, hookConfig);
+  }
+
+  if (!includes(Object.keys(uspCallMap), userUSP)) {
```
harpere: more simply, `if (!uspCallMap[userUSP]) {`

modules/consentManagementUsp.js
```diff
@@ -0,0 +1,309 @@
...
+  clearTimeout(hookConfig.timer);
+  storeUspConsentData(consentObject);
+  exitModule(null, hookConfig);
+}
+
+/**
+ * General timeout callback when interacting with USPAPI takes too long.
+ */
+function uspapiTimeout(hookConfig) {
+  cmpFailed('USPAPI workflow exceeded timeout threshold.', hookConfig);
```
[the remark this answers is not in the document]
harpere: should be `uspapiFailed()`

modules/consentManagementUsp.js
```diff
@@ -0,0 +1,309 @@
...
+ * @param {object} hookConfig contains module related variables (see comment in requestBidsHook function)
+ * @param {object} extraArgs contains additional data that's passed along in the error/warning messages for easier debugging
+*/
+function uspapiFailed(errMsg, hookConfig, extraArgs) {
+  clearTimeout(hookConfig.timer);
+
+  // still set the consentData to undefined when there is a problem as per config options
+  if (allowAuction) {
+    storeUspConsentData(undefined);
+  }
```
harpere: this is not relevant to uspapi

test/spec/modules/consentManagementUsp_spec.js
```diff
@@ -0,0 +1,271 @@
...
+  allowAuction
+} from 'modules/consentManagementUSP';
+import * as utils from 'src/utils';
+import { config } from 'src/config';
+import { uspDataHandler } from 'src/adapterManager';
+
+let assert = require('chai').assert;
+let expect = require('chai').expect;
+
+describe.only('consentManagement', function () {
```
harpere: `.only` needs to be removed before merging

modules/consentManagementUsp.js
harpere: should be fixed now.

modules/consentManagementUsp.js
```diff
@@ -0,0 +1,309 @@
...
+export function resetConsentData() {
+  consentData = undefined;
+  uspDataHandler.setConsentData(null);
+}
+
+/**
+ * A configuration function that initializes some module variables, as well as add a hook into the requestBids function
+ * @param {object} config required; consentManagementUSP module config settings; usp (string), timeout (int), allowAuctionWithoutConsent (boolean)
+ */
+export function setConsentConfig(config) {
```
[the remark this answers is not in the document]
harpere: this was added

modules/consentManagementUsp.js
```diff
@@ -7,22 +7,19 @@
 import * as utils from '../src/utils';
 import { config } from '../src/config';
 import { uspDataHandler } from '../src/adapterManager';
-import includes from 'core-js/library/fn/array/includes';
 
-const DEFAULT_CONSENT_API = 'uspapi';
+const DEFAULT_CONSENT_API = 'iab';
```
tjeastmond: What is the significance of changing this to `iab`?

modules/consentManagementUsp.js
```diff
@@ -33,63 +30,65 @@ const uspCallMap = {
...
-    if (!uspapiFrame) {
-      return uspError('CMP not found.', hookConfig);
-    }
+  if (!uspapiFrame) {
+    return uspError('CMP not found.', hookConfig);
+  }
 
+  try {
+    // try to call __uspapi directly
+    uspapiFrame.__uspapi('getUSPData', 1, callbackHandler.consentDataCallback);
```
tjeastmond: I think we should extract the version to a config, global or otherwise.

src/adapterManager.js
```diff
@@ -271,7 +271,7 @@ adapterManager.makeBidRequests = function(adUnits, auctionStart, auctionId, cbTi
   if (gdprDataHandler.getConsentData()) {
     bidRequests.forEach(bidRequest => {
       bidRequest['gdprConsent'] = gdprDataHandler.getConsentData();
-      bidRequest['ccpaConsent'] = uspDataHandler.getConsentData();
+      bidRequest['uspConsent'] = uspDataHandler.getConsentData();
```
tjeastmond: This is fine, but the naming convention seems to assume we will never need to fetch another consent or additional information via the `USP` framework.

`USP` consent is the framework while the consent data here is specific to `CCPA`, as `GDPR` is the consent fetched from what we defined as the `IAB` framework.

modules/consentManagement.js
```diff
@@ -343,6 +343,9 @@ export function resetConsentData() {
  * @param {object} config required; consentManagement module config settings; cmp (string), timeout (int), allowAuctionWithoutConsent (boolean)
  */
 export function setConsentConfig(config) {
+  // if `config.gdpr` or `config.usp` exist, assume new config format.
+  // else for backward compatability, just use `config`
+  config = config.gdpr || config.usp ? config.gdpr : config;
```
bretg: Why checking config.usp here? That doesn't seem right.

modules/consentManagementUsp.js
bretg: we only support the `iab` api for now. That could change in the future. (e.g. we might support 'static' or even 'daa')

modules/consentManagement.js
jsnellbaker: As a separate question from the above - what is the expected use-case around someone who has both GDPR and USP modules together but only specified the USP in their config code?

From looking at the current implementation here, it seems we'll end-up setting the internal defaults for the GDPR module and try to make a call the GDPR CMP.  Is that expected?  Or should it only call the USP CMP because that's what the pub had setup in their config?

src/adapterManager.js
harpere: not a bad point, but I think it's unclear.  The "command" we're retrieving the data from is "getUSPData", so I think calling it "usp" is fine.  perhaps in the future other types of consent would be included in the "uspData" or "uspString"

modules/consentManagement.js
harpere: @bretg - the idea is that if they have either `gdpr` or `usp`, they are using the new format.  they shouldn't mix `usp` with the old `gdpr` config format.

@jsnellbaker - I can imagine a publisher might build a single wrapper (including both gdpr and usp consent modules) for multiple sites, one where they care about gdpr and one where they care about usp.  Or maybe depending on where the user is coming from they may decide to active one or the other.

So I was thinking we should only call the CMPs (gdpr or usp) that are configured.  I think this is backward compatible because if they have the old format (no `gdpr` or `ups` params specified), that means gdpr is configured.

modules/consentManagementUsp.js
```diff
@@ -0,0 +1,291 @@
...
+  while (!uspapiFrame) {
+    try {
+      if (f.frames['__uspapiLocator']) uspapiFrame = f;
+    } catch (e) { }
+    if (f === window.top) break;
+    f = f.parent;
+  }
+
+  if (!uspapiFrame) {
+    return uspError('CMP not found.', hookConfig);
```
jsnellbaker: Could we perhaps clarify this message to denote it was trying to find a USP CMP?

PR #4509, opened 2019-11-27
TripleLift: User sync fallback

davidwoodsandersen:
## Type of change
- [x] Feature

## Description of change
Adds conditional support for image syncing when iframe syncing is not enabled. Also updates sync endpoint URL.


bwoolcott:
@sumit116 Have you had a chance to review this yet? Would be great to have this in today's release

PR #4516, opened 2019-11-29
fix betweenBidAdapter 

Tweeddy:
## Type of change

- [x] Bugfix
- [ ] Feature
- [ ] New bidder adapter   
- [ ] Build related changes
- [ ] CI related changes
- [ ] Other

## Description of change

Fixed adapter parameters validation (buggy exception thrown), and tests on it

```
{
  bidder: '<bidder name>',
  params: {
    // ...
  }
}
```

### DIFF
--- a/modules/betweenBidAdapter.js
+++ b/modules/betweenBidAdapter.js
@@ -1,4 +1,5 @@
 import {registerBidder} from '../src/adapters/bidderFactory';
+import * as utils from '../src/utils';
 
 const BIDDER_CODE = 'between';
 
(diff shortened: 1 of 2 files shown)


robertrmartinez:
@Tweeddy Hi, Can you please update the PR with more information on the bug and what the proposed fix is?

You can also remove all unnecessary default PR text to make it easier to read.

Thanks.

Tweeddy:
@robertrmartinez Hi!
we had problems when the publisher entered parameters in the wrong format. (for example, w and h did not belong to params or bid.params was not an object) as a result, an exeption was thrown and the entire request was broken. we have added additional checks on parameters to exclude such cases and notify the user about incorrect data entry.

modules/betweenBidAdapter.js
```diff
@@ -13,7 +14,22 @@ export const spec = {
...
+        !utils.isNumber(bid.params.s)) {
+        utils.logError(BIDDER_CODE + ': the value of adUnits params is wrong or absent.');
+        return false;
+      } else {
+        return true;
+      }
+    } else {
+      utils.logError(BIDDER_CODE + ': empty bid object.');
+      return false;
+    }
```
robertrmartinez: This looks fine,

But perhaps it could be simplified a little.

You can utilize the `utils.deepAccess` method and get away with some of this. Here is a quick example:

```javascript
isBidRequestValid: function(bid) {
   let params = utils.deepAccess(bid, 'params') || {};
   if (utils.isNumber(params.w) && utils.isNumber(params.h) && utils.isNumber(params.s)) {
      return true;
   } else {
      utils.logError(`The value of ${BIDDER_CODE}'s params are incorrect. 'w', 'h', and 's' must all be numbers`);
      return false;
   }
}
```

Although we do like it if an adapter is flexible when it comes to hard requiring params to be a number rather than checking if the passed param as a string is actually just a number represented as a string. Sometimes publishers / users will put all bidder params as `Strings` even when the value of your param is technically a number:

```javascript
params = {
   w: '728',
   h: '90',
   s: '3245235'
}
```

Technically would not work, when the intention is there.

So you _could_ add some logic in your adapter to convert to numbers if passed as string, but no worries, not required, just a thought.

But regardless, I think simplifying this into one if / else is more clean and readable.

PR #4538, opened 2019-12-05
feature: make adjustment for an bid (not only cpm)

### LINKED ISSUE #4312: [Question] Using custom renderer for InPage ads
## Type of issue
Question or Feature Request

## Description
Currently it is possible to set a custom renderer per `AdUnit` (see [renderer docs](http://prebid.org/dev-docs/publisher-api-reference.html)) register an renderer in an `BidAdapter` implementation. This is indended to use for `outstream ads` (see [Outsteam documentation](http://prebid.org/dev-docs/show-outstream-video-ads.html)).

We would like to interpret the bid response in an callback function and set an custom renderer to create a kind of creative wrapper. Ideally this should be possible without modifiing several BidAdapters.

## Steps to reproduce
The renderer logic is currently implemented here:
https://github.com/prebid/Prebid.js/blob/3965d3afe7c99b6da40e72bccd8a4a6c2412f6cd/src/prebid.js#L314-L357

AFAIK there is no way to inspect bids and switch the renderer this way!?

derdeka:
## Type of change
- [ ] Bugfix
- [x] Feature
- [ ] New bidder adapter   
- [ ] Build related changes
- [ ] CI related changes
- [ ] Other

## Description of change

Allow to adjust the bid, but allow more than just the cpm adjustment.

Fixes #4312

This Draft PR is intended as a first discussion point about adjusting the bid, more than just the cpm.

### DIFF
--- a/src/auction.js
+++ b/src/auction.js
@@ -676,26 +676,34 @@ function setKeys(keyValues, bidderSettings, custBidObj) {
 }
 
 export function adjustBids(bid) {
-  let code = bid.bidderCode;
-  let bidPriceAdjusted = bid.cpm;
-  let bidCpmAdjustment;
+  const code = bid.bidderCode;
   if ($$PREBID_GLOBAL$$.bidderSettings) {
-    if (code && $$PREBID_GLOBAL$$.bidderSettings[code] && typeof $$PREBID_GLOBAL$$.bidderSettings[code].bidCpmAdjustment === 'function') {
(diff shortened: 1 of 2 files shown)


jsnellbaker:
Tagging a few folks to contribute to this discussion/proposal.

bretg:
@derdeka - is there anything here that couldn't be accomplished with an [onEvent](https://docs.prebid.org/dev-docs/publisher-api-reference.html#module_pbjs.onEvent) handler?

bretg:
I'm closing this PR to reduce the number of Github reminders I get every day. @derdeka - you can re-open when you have more details about the use case.

PR #4553, opened 2019-12-07
PubMatic to send test:1 in the request if pubmaticTest=true is present in URL

pm-harshad-mane:
## Type of change
- [X] Feature

## Description of change
PubMatic to send test:1 in the request if "pubmaticTest=true" is present in URL
Then PubMatic end-point will return non-zero test-bids.


pm-harshad-mane:
Hello @jsnellbaker ,

PR #4578, opened 2019-12-11
Update Sharethrough bid adapter endpoint

madma:
## Type of change

## Description of change
- Sharethrough bid adapter endpoint has been updated to always use `https`
- Protocol of the referrer is now sent as an additional parameter in bid requests


madma:
conflicts resolved @harpere -- thanks for reviewing!

madma:
hi, just following up on this @harpere. thanks!

PR #4587, opened 2019-12-12
Fixes for Prebid 3.0

bjorn-lw:
## Type of change
- [x ] Refactoring (no functional changes, no api changes)

## Description of change
Adapter was removed with #4249 . This fix removes the dependency on getTopWindowLocation previously used in one place in the adapter.

```

## Other information

#4249

### CODE REVIEW

File: test/spec/modules/livewrappedBidAdapter_spec.js [hunk 1]
```diff
@@ -0,0 +1,835 @@
...
+          bidId: '2ffb201a808da7',
+          transactionId: '3D1C8CF7-D288-4D7F-8ADD-97C553056C3D',
+          formats: [{width: 980, height: 240}, {width: 980, height: 120}]
+        }]
+      };
+
+      expect(data).to.deep.equal(expectedQuery);
+    });
+
+    it('should use params.url, then config pageUrl, then bidderRequest.refererInfo.referer', function() {
```
bjorn-lw: This is the only test that was modified, except for changing from "//" to "https://" in all tests against our endpoint url. Can anything in this test cause the error "Some of your tests did a full page reload!"? The other tests that fails are not related to our code changes (openx analytics adapter issues)

PR #4594, opened 2019-12-13
update several tests to let them work in ie11

jsnellbaker:
## Type of change
- [x] Other

## Description of change
There were several tests that were having issues running in IE11 as seen from our browserstack/circleci tests.

I have put together some changes that should address the issues that were observed.

A note for the kargoBidAdapter - it was observed sporadically that the mocked `config.getConfig` in that test suite would expect debug was defined.  When it wasn't, the thrown `Error` would stop the testing.  Adding this value to the mocked function should stop this from happening.

### DIFF
--- a/test/spec/modules/emoteevBidAdapter_spec.js
+++ b/test/spec/modules/emoteevBidAdapter_spec.js
@@ -777,7 +777,7 @@ describe('emoteevBidAdapter', function () {
         spec.isBidRequestValid(validBidRequest);
         sinon.assert.notCalled(utils.triggerPixel);
         sinon.assert.notCalled(utils.getCookie);
-        sinon.assert.notCalled(config.getConfig);
+        // sinon.assert.notCalled(config.getConfig);
         sinon.assert.notCalled(utils.getParameterByName);
       });
     });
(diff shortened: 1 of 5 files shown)


jsnellbaker:
Hi @mkendall07 

I saw that too and I was trying to look into those tests to see why they were failing like that.  I think that some earlier test was delaying its response/callback which caused a utils logMessage to fire around the time of these tests were executing.  

I haven't been able to replicate it to determine for sure if this was the case.  At this point though I will take your suggestion and disable these tests for now (will create an issue in a little bit).

jsnellbaker:
To note - I have pushed in the update to disable those specific emoteev checks.

test/spec/modules/emoteevBidAdapter_spec.js
```diff
@@ -789,16 +797,17 @@ describe('emoteevBidAdapter', function () {
...
         sinon.assert.notCalled(utils.triggerPixel);
         sinon.assert.notCalled(utils.getCookie);
-        sinon.assert.callCount(config.getConfig, 3);
+        // sinon.assert.callCount(config.getConfig, 3);
         sinon.assert.callCount(utils.getParameterByName, 2);
       });
     });
     describe('interpretResponse', function () {
       it('has intended side-effects', function () {
+        debugger; //eslint-disable-line
```
msm0504: Does this `debugger` statement need to be merged in? Or can this be removed?

PR #4598, opened 2019-12-15
Improvements for browsi RTD provider

omerBrowsi:
## Type of change
- [ ] Bugfix
- [ ] Other

## Description of change

1. Improvements for browsi RTD provider (support for placement id macros, timeout for ajax request)
2. Change bidsBackCallback hook data (adUnits instead of adUnits codes)
3. Add timeout param (used when we don't need auction dealy)


omerBrowsi:
updated docs pull request - 
https://github.com/prebid/prebid.github.io/pull/1690

modules/browsiRtdProvider.js
```diff
@@ -157,36 +167,74 @@ function getKVObject(p, keyName) {
...
+  }
+  return id;
 }
 
 /**
  * XMLHttpRequest to get data form browsi server
  * @param {string} url server url with query params
  */
 function getPredictionsFromServer(url) {
+  let ajax = ajaxBuilder(_moduleParams.auctionDelay || _moduleParams.timeout || DEF_TIMEOUT);
```
bretg: I think `auctionDelay` should be up at the higher level underneath realTimeData, not within the dataProviders section.

omerBrowsi: It is. `auctionDelay` and `timeout` are both underneath realTimeData
I push those values to `_moduleParams ` on init to keep all relevant configs for the module on `_moduleParams `

modules/browsiRtdProvider.js
```diff
@@ -157,36 +167,74 @@ function getKVObject(p, keyName) {
...
+function getMacroId(macro, id, slot) {
+  if (macro) {
+    try {
+      const macroString = macro
+        .replace(/<DIV_ID>/g, `${id}`)
+        .replace(/<AD_UNIT>/g, `${slot.getAdUnitPath()}`)
+        .replace(/<KEY_(\w+)>/g, (match, p1) => {
+          return (p1 && slot.getTargeting(p1).join('_')) || 'NA';
+        });
+      return eval(macroString);// eslint-disable-line no-eval
```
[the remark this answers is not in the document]
omerBrowsi: We use eval to generate macros for placement id's (to handle auto-generated placements id's).
However, we gave it another thought and decided to change the macros to work without `eval` (regex only).
I will commit a fix for that.

src/auction.js
```diff
@@ -168,7 +168,7 @@ export function newAuction({adUnits, adUnitCodes, callback, cbTimeout, labels, a
       _auctionEnd = Date.now();
 
       events.emit(CONSTANTS.EVENTS.AUCTION_END, getProperties());
-      bidsBackCallback(_adUnitCodes, function () {
+      bidsBackCallback(_adUnits, function () {
```
bretg: @omerBrowsi - Please provide some background on this core change.

bretg:
Admittedly I didn't do as much digging as I should have, but seems like this item would have affected more places. Can you please provide a bit of background for why this change is proposed?

> Change bidsBackCallback hook data (adUnits instead of adUnits codes)

src/auction.js
omerBrowsi: RTD module is the place this hook is used, so it won't affect anything else (the hook was made for this module)
The change sends as parameter the adUnits objects (that hold all data on the ad units in this auction) instead of just ad units codes.
Both variables already exist n auction.js, I just choose to send different one as a parameter

jsnellbaker:
@omerBrowsi When you have the chance, can you resolve the conflicts here?

omerBrowsi:
@jsnellbaker 
we prefer not using the `loadExternalScript`function
we need to set attributes and objects on our script and the function doesn't support it

@snapwich - the code was changed by you, is it critical that we use `loadExternalScript`?
if so, we will need to adjust this function
(code was changed on this pull request - https://github.com/prebid/Prebid.js/pull/4687)

snapwich:
@omerBrowsi yes it is required; your module was one of many that were making live requests during runs of the test suite in CI; that's not ideal for obvious reasons. Rather than verify everyone is properly stubbing it's much easier for us to just mandate use of the `loadExternalScript` when you load an external script (something that should happen rarely to begin with) and have `loadExternalScript` stub.

How do you need to adjust the function? I did make changes in my pull-request to accommodate your use-case: `loadExternalScript` now returns the script tag which can then be modified however you would like.

omerBrowsi:
changed to use `loadExternalScript`
I didn't make any changes to `loadExternalScript` function, we will handle the difference on our side.
Update the macro to support new configs on our side

omerBrowsi:
@bretg @jsnellbaker any update on this? do you need anything from me?

modules/browsiRtdProvider.js
```diff
@@ -155,36 +167,88 @@ function getKVObject(p, keyName) {
...
+ * @param {Object} slot google slot
+ * @return {?Object}
+ */
+function getMacroId(macro, id, slot) {
+  if (macro) {
+    try {
+      const macroResult = evaluate(macro, slot.getSlotElementId(), slot.getAdUnitPath(), (match, p1) => {
+        return (p1 && slot.getTargeting(p1).join('_')) || 'NA';
+      });
+      console.log(`${slot.getSlotElementId()} macro result: ${macro} => ${macroResult}`);
```
jsnellbaker: If this is a log message you want to keep to help with debugging purposes, can you please use the `utils.logMessage` or `utils.logInfo` methods instead of just `console.log`?  This would help to flag it properly in the browser console logs as something related to Prebid.

If you don't want to keep it long-term, can you please remove it?

modules/browsiRtdProvider.js
```diff
@@ -33,14 +36,20 @@ let _dataReadyCallback = null;
...
+  script.async = true;
   script.setAttribute('data-sitekey', _moduleParams.siteKey);
   script.setAttribute('data-pubkey', _moduleParams.pubKey);
   script.setAttribute('prebidbpt', 'true');
   script.setAttribute('id', 'browsi-tag');
+  script.setAttribute('src', data.u);
+  script.prebidData = utils.deepClone(data);
+  if (_moduleParams.keyName) {
+    script.prebidData.kn = _moduleParams.keyName;
+  }
```
jsnellbaker: Just to confirm - this specific setup with for the `prebidData` property is correct?  

I tried to load the `browsi_test_Page.html` and I couldn't really confirm if this key and its values were being properly used by the script; so I wanted to ask and confirm in case that didn't sound right to you.

modules/browsiRtdProvider.js
omerBrowsi: log removed

modules/browsiRtdProvider.js
omerBrowsi: Yes, it's correct
I updated the demo page so you can see this there as well

omerBrowsi:
@bretg @jsnellbaker  Thanks!
@bretg the real-time data (and browsi) module does not appear on the download page, is there anything we need to do on our side?

PR #4602, opened 2019-12-15
CleanMediaNet: Update adaptor code to meet Prebid.js v3.0 constraints

sa1omon:
## Type of change
- [x] Bugfix

## Description of change
- Update adaptor code to meet Prebid.js v3.0 constraints

### Contact email of the adapter’s maintainer


modules/cleanmedianetBidAdapter.js
```diff
@@ -0,0 +1,297 @@
...
+
+    const bids = response.seatbid.reduce(
+      (acc, seatBid) => acc.concat(seatBid.bid),
+      []
+    );
+    let outBids = [];
+
+    bids.forEach(bid => {
+      const outBid = {
+        adId: bidRequest.bidRequest.adUnitCode,
```
robertrmartinez: What is the reasoning for doing this? Setting an adId to the adUnitCode seems pretty unusual.

Also, seems like if there are multiple server responses all will have the same adUnitCode assigned.

modules/cleanmedianetBidAdapter.js
```diff
@@ -0,0 +1,297 @@
...
+        requestId: bidRequest.bidRequest.bidId,
+        cpm: bid.price,
+        width: bid.w,
+        height: bid.h,
+        ttl: 60 * 10,
+        creativeId: bid.crid,
+        netRevenue: true,
+        currency: bid.cur || response.cur,
+        adUnitCode: bidRequest.bidRequest.adUnitCode,
```
robertrmartinez: This is unnecessary in my opinion because prebid will already assign the correct bidResponse to the correct ad unit and thus adUnitCode based on the requestId you pass in (Which should match to the correct bidId)

What are the reasonings for adding this adUnitCode in these places?

modules/cleanmedianetBidAdapter.js
```diff
@@ -0,0 +1,297 @@
...
+        };
+      }
+
+      const imp = {
+        id: transactionId,
+        instl: params.instl === 1 ? 1 : 0,
+        tagid: adUnitCode,
+        bidfloor: params.bidfloor || 0,
+        bidfloorcur: 'USD',
+        secure: 1
```
robertrmartinez: 👍

modules/cleanmedianetBidAdapter.js
sa1omon: Hi @robertrmartinez,
I have removed the unnecessary properties`adId` and  `adUnitCode`.
Thanks for the insight!

In addition,
I saw the tests failure in Circle CI, but it seems that it is not related to our adaptor, can you kindly give me some information about the reason of the failure?
Thank you very much!

modules/cleanmedianetBidAdapter.js
sa1omon: please see my comment above

robertrmartinez:
Hi @sa1omon Can you pull latest from master please?

I think there are some missing commits which have helped fix some of our circle CI issues!

sa1omon:
> Hi @sa1omon Can you pull latest from master please?
> 
> I think there are some missing commits which have helped fix some of our circle CI issues!

Ok, now it passed the CIrcle CI tests.
Thank you very much!

PR #4617, opened 2019-12-17
Sonobi - added CCPA support to sonobi bidder adapter

JonGoSonobi:
## Type of change
- [x] Feature

## Description of change
Updated the sonobi bidder adapter to send the uspConsent string to the bid endpoint as query param us_privacy.

- apex.prebid@sonobi.com


mkendall07:
Thanks for contributing. Can you please also open a PR against the 2.44.x-legacy branch as well?

PR #4621, opened 2019-12-17
fix a bug when the iframe locator is not present on page

mkendall07:
## Type of change

- [x] Bugfix

## Description of change
This allows consent to be gathered even if a iframe locator is not on the page. Basically it checks for the global first if it exists in the current page and then only proceeds to look for it in other window/scopes


bretg:
Thanks Matt, but slimming the reviewer list to just @harpere

PR #4627, opened 2019-12-17
lkqdBidAdapter 

mrcrawfo:
## Type of change
- [x] Bugfix
- [x] Feature
- [ ] New bidder adapter  
- [ ] Build related changes
- [ ] CI related changes
- [ ] Other

## Description of change
 - Added support for 'schain' ad request parameter
 - Fixed to read sizes from mediaTypes.banner.sizes in compliance with [Prebid.js 3.0 release](http://prebid.org/blog/pbjs-3)


modules/lkqdBidAdapter.js
```diff
@@ -36,21 +36,32 @@ function buildRequests(validBidRequests, bidderRequest) {
   for (let i = 0; i < validBidRequests.length; i++) {
     let bidRequest = validBidRequests[i];
 
+    let sizes = [];
     // if width/height not provided to the ad unit for some reason then attempt request with default 640x480 size
-    if (!bidRequest.sizes || !bidRequest.sizes.length) {
+    if ((!bidRequest.sizes || !bidRequest.sizes.length) && (!bidRequest.mediaTypes || !bidRequest.mediaTypes.banner || !bidRequest.mediaTypes.banner.sizes || !bidRequest.mediaTypes.banner.sizes.length)) {
```
msm0504: Instead of checking each level in `bidRequest.mediaTypes.banner.sizes`, use `utils.deepAccess`

modules/lkqdBidAdapter.js
```diff
@@ -36,21 +36,32 @@ function buildRequests(validBidRequests, bidderRequest) {
...
-      let adHeight = bidRequest.sizes[1];
-      bidRequest.sizes = [[adWidth, adHeight]];
+    if (bidRequest.sizes) {
+      sizes = bidRequest.sizes;
+      if (bidRequest.sizes.length === 2 && typeof bidRequest.sizes[0] === 'number' && typeof bidRequest.sizes[1] === 'number') {
+        let adWidth = bidRequest.sizes[0];
+        let adHeight = bidRequest.sizes[1];
+        sizes = [[adWidth, adHeight]];
+      }
+    } else if (bidRequest.mediaTypes && bidRequest.mediaTypes.banner && bidRequest.mediaTypes.banner.sizes) {
```
msm0504: Similar to check on line 41. Consider storing `bidRequest.mediaTypes.banner.sizes` in a variable to avoid having to get this value repeatedly.

Same can be done with `bidRequest.sizes` above

modules/lkqdBidAdapter.js
```diff
@@ -36,21 +36,32 @@ function buildRequests(validBidRequests, bidderRequest) {
...
-    if (bidRequest.sizes.length === 2 && typeof bidRequest.sizes[0] === 'number' && typeof bidRequest.sizes[1] === 'number') {
-      let adWidth = bidRequest.sizes[0];
-      let adHeight = bidRequest.sizes[1];
-      bidRequest.sizes = [[adWidth, adHeight]];
+    if (bidRequest.sizes) {
+      sizes = bidRequest.sizes;
+      if (bidRequest.sizes.length === 2 && typeof bidRequest.sizes[0] === 'number' && typeof bidRequest.sizes[1] === 'number') {
+        let adWidth = bidRequest.sizes[0];
+        let adHeight = bidRequest.sizes[1];
+        sizes = [[adWidth, adHeight]];
```
msm0504: `adWidth` and `adHeight` are not needed.

Can instead be `sizes = [bidRequest.sizes]`

modules/lkqdBidAdapter.js
```diff
@@ -36,21 +36,32 @@ function buildRequests(validBidRequests, bidderRequest) {
...
+      }
+    } else if (bidRequest.mediaTypes && bidRequest.mediaTypes.banner && bidRequest.mediaTypes.banner.sizes) {
+      sizes = bidRequest.mediaTypes.banner.sizes;
+      if (bidRequest.mediaTypes.banner.sizes.length === 2 && typeof bidRequest.mediaTypes.banner.sizes[0] === 'number' && typeof bidRequest.mediaTypes.banner.sizes[1] === 'number') {
+        let adWidth = bidRequest.mediaTypes.banner.sizes[0];
+        let adHeight = bidRequest.mediaTypes.banner.sizes[1];
+        sizes = [[adWidth, adHeight]];
```
msm0504: `adWidth` and `adHeight` are not needed.

Can instead be `sizes = [bidRequest.mediaTypes.banner.sizes]`

modules/lkqdBidAdapter.js
```diff
@@ -165,7 +179,7 @@ function buildRequests(validBidRequests, bidderRequest) {
       bidRequests.push({
         method: 'GET',
         url: sspUrl,
-        data: sspData
+        data: Object.keys(sspData).map(function (key) { return key + '=' + sspData[key]} ).join('&') + '&'
```
msm0504: Fix lint error. `} )` should be ` })`

Also, `reduce` could be used here instead of `map` followed by `join`. See [docs](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/Reduce)

modules/lkqdBidAdapter.js
```diff
@@ -182,29 +196,34 @@ function interpretResponse(serverResponse, bidRequest) {
...
-            bidResponse.creativeId = sspXml.getElementsByTagName('Ad')[0].getAttribute('id');
-            bidResponse.currency = sspXml.getElementsByTagName('Pricing')[0].getAttribute('currency');
-            bidResponse.netRevenue = true;
-            bidResponse.vastUrl = sspUrl;
-            bidResponse.vastXml = sspXmlString;
-            bidResponse.mediaType = VIDEO;
-
-            bidResponses.push(bidResponse);
+        if (bidRequest && bidRequest.data && typeof bidRequest.data === 'string') {
+          let sspData = new URLSearchParams(bidRequest.data);
```
msm0504: `URLSearchParams` is not supported in IE. See [docs](https://developer.mozilla.org/en-US/docs/Web/API/URLSearchParams)

modules/lkqdBidAdapter.md
```diff
@@ -19,10 +19,14 @@ For more information about [LKQD Ad Serving and Management](http://www.lkqd.com/
 var videoAdUnit = [
 {
     code: 'video1',
-    sizes: [
-        [300, 250],
-        [640, 480]
-    ],
+    mediaTypes: {
+        banner: {
```
msm0504: Why does a video ad unit have banner sizes?

modules/lkqdBidAdapter.js
mrcrawfo: Fixed

modules/lkqdBidAdapter.js
mrcrawfo: Fixed

modules/lkqdBidAdapter.js
mrcrawfo: Fixed

modules/lkqdBidAdapter.js
mrcrawfo: Fixed

modules/lkqdBidAdapter.js
mrcrawfo: Fixed lint error
sspData is an Object and not an Array, so `reduce` was not applicable

modules/lkqdBidAdapter.js
mrcrawfo: Fixed

modules/lkqdBidAdapter.md
mrcrawfo: Fixed

modules/lkqdBidAdapter.js
msm0504: `Object.keys` returns an array. That's why you're able to use `map` and `join` currently.

modules/lkqdBidAdapter.md
msm0504: Switching to using `sizes` does not fix this issue. That property is no longer supported in Prebid 3.

Sizes for video should be specified in `mediaTypes.video.playerSize`. Your adapter code needs to also support using this property. Since you're only supporting video, you might be able to simply replace `mediaTypes.banner.sizes` with `mediaTypes.video.playerSize`.

Here are the mediaTypes docs: http://prebid.org/dev-docs/publisher-api-reference.html#addAdUnits-MediaTypes

mrcrawfo: Fixed - and I appreciate you for clarifying this point, I had been confused because the migration guide http://prebid.org/blog/pbjs-3 I had seen did not mention `video.playerSize` and had made it seem that I was supposed to use `banner.sizes` regardless. This makes much more sense now, thank you!

PR #4673, opened 2019-12-23
Add CCPA support to Beachfront adapter

jsalis:
## Type of change

- [x] Feature

## Description of change

Adds CCPA (US Privacy) support to the Beachfront adapter.

## Other information

PR for docs:
prebid/prebid.github.io#1720


jsnellbaker:
@jsalis As a suggestion, you may also want to update your adapter in the 2.44.x-legacy branch, to allow any pubs still using Prebid 2.x the ability to use a compliant version of your adapter.

PR #4676, opened 2019-12-23
ozone 2.2.0 adapter updates

afsheenb:
Type of change

 [X] Bugfix
 [X] Feature

 

Description of changes

• Removal of GDPR parameters that were being read outside of the daisy-bit (which breaks requests when used on-page with certain CMP implementations that don't expose those values.)

• An onpage 'ozone' config to turn off/disable the custom key-value pairs the adapter sets should the publishers find their ad-request URL lengths are close to exceeding Google/DFP limits

• TTL update to 300 ms for bids in the adapter, up from 60 ms

• Cookie sync - append querystring values (if present) in call to cookie sync endpoint - gdpr, gdpr_consent, pubcid, publisherId, and siteId

• Updated checks / validation for customData to ensure clients cannot send invalid customData (currently customData not passing these checks silently breaks on the server-side, now we'll throw a client-side error to avoid confusion)

• Cleaned Up Developer comments as requested from previous PR

• added support for prebid user ID module

• Added support of ozone-specific querystrings for debugging/QA.

### DIFF
--- a/modules/ozoneBidAdapter.js
+++ b/modules/ozoneBidAdapter.js
@@ -3,7 +3,7 @@ import { registerBidder } from '../src/adapters/bidderFactory';
 import { BANNER, NATIVE, VIDEO } from '../src/mediaTypes';
 import {config} from '../src/config';
 import {getPriceBucketString} from '../src/cpmBucketManager';
-import { Renderer } from '../src/Renderer'
+import { Renderer } from '../src/Renderer';
 
 const BIDDER_CODE = 'ozone';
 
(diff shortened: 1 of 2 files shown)


AskRupert-DM:
@sumit116 - happy new year! Let us know if you need anything to assist with this approval !

modules/ozoneBidAdapter.js
```diff
@@ -263,46 +313,195 @@ export const spec = {
...
+  /**
+   * Produces external userid object
+   */
+  addExternalUserId(eids, value, source, atype) {
+    if (utils.isStr(value)) {
+      eids.push({
+        source,
+        uids: [{
+          id: value,
+          atype
```
smenzer, line 449: @afsheenb I was going through this code for another PR I'm working on (to convert our ID5 ID from a string to an object, which would require a small change to your adapter...I'll tag you in it when it's ready).

In this code here, I don't think you're generating a valid json object. the `source` and `atype` values passed in to this function from below are just strings/numbers, so you're missing the keys for those items. I think the code here should look like this instead:

```javascript
eids.push({
  source: source,
  uids: [{
    id: value,
    atype: atype
  }]
});
```

PR #4685, opened 2020-01-02
Add static API option to the consentManagementUsp module.

dbirkbeck:
The GDPR consent management module has a static configuration option. This change adds "static" as an API option to the consentManagementUsp module.

## Type of change

- [ ] Bugfix
- [x] Feature
- [ ] New bidder adapter   
- [ ] Build related changes
- [ ] CI related changes
- [ ] Other

## Description of change

This change allows a static configuration such as the following to be used rather than the 'iab' API.

```
consentManagement: {
    usp: {
        cmpApi: 'static',
        consentData: {
            getUSPData: {
                uspString: '1YYY'
            }
        }
    }
}
```

### DIFF
--- a/modules/consentManagementUsp.js
+++ b/modules/consentManagementUsp.js
@@ -296,8 +296,8 @@ export function setConsentConfig(config) {
   utils.logInfo('USPAPI consentManagement module has been activated...');
 
   if (consentAPI === 'static') {
-    if (utils.isPlainObject(config.consentData)) {
-      staticConsentData = config.consentData;
+    if (utils.isPlainObject(config.consentData) && utils.isPlainObject(config.consentData.getUSPData)) {
+      if (config.consentData.getUSPData.uspString) staticConsentData = { usPrivacy: config.consentData.getUSPData.uspString };
       consentTimeout = 0;
     } else {
       utils.logError(`consentManagement config with cmpApi: 'static' did not specify consentData. No consents will be available to adapters.`);
--- a/modules/consentManagementUsp.js
+++ b/modules/consentManagementUsp.js
@@ -14,14 +14,26 @@ const USPAPI_VERSION = 1;
 
 export let consentAPI;
 export let consentTimeout;
+export let staticConsentData;
 
 let consentData;
 let addedConsentHook = false;
 
 // consent APIs
 const uspCallMap = {
-  'iab': lookupUspConsent
+  'iab': lookupUspConsent,
+  'static': lookupStaticConsentData
 };
+
+/**
+ * This function reads the consent string from the config to obtain the consent information of the user.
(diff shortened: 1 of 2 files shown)


jsnellbaker:
@dbirkbeck Can you please put together some unit tests that use this new static config approach and add them to this PR?

harpere:
@dbirkbeck this implementation works, **BUT** I think we should be consistent with the [format the GDPR consentManagement module uses](http://prebid.org/dev-docs/modules/consentManagement.html#examples).  I realize that internally we are **NOT** consistent, but I think the external API should be (and we probably should change the internal representation of the consent data to be consistent as well).   this is what I'm proposing:
```
consentManagement: {
    usp: {
        cmpApi: 'static',
        consentData: {
            getUSPData: {
                uspString: '1YNY'
            }
        }
    }
}
```
Notice `getUSPData` corresponds to the "command" used to retrieve the data from the USP CMP, and `uspString` is the actual param name returned by the CMP.  So in addition to be consistent with the GDPR implementation, it's also consistent with the object returned by the CMP, and adds flexibility for other USP "commands" we might support in the future.

dbirkbeck:
I've updated the module/unit test to reflect the new public API format. It is agreed that changing the internal implementation may be desirable but doing so will require a much larger effort and should probably be left to a next step.

harpere:
@dbirkbeck would you update the docs at https://github.com/prebid/prebid.github.io/blob/master/dev-docs/modules/consentManagementUsp.md  ?   I think you could copy/modify the `static` documentation from https://github.com/prebid/prebid.github.io/blob/master/dev-docs/modules/consentManagement.md

dbirkbeck:
A pull request has been created for the documentation updates requested.

adrianpoix:
@dbirkbeck is static API option for CCPA available with v2.44.2 ?

dbirkbeck:
The static API for CCPA is only in 3.2.0 thus far. As most are moving toward 3.x and I wasn't sure how many would use the static API, I focussed my efforts on that branch. I can create a PR for the 2.x branch as well if you think there is value in it.

jsnellbaker:
@dbirkbeck & @adrianpoix 

We actually just created/merged a PR yesterday for the `2.44.x-legacy` branch that adds the static approach for the CCPA/USP API.  It should be available for use with the next legacy release - which should happen next week (in the 2.44.3 release).

dbirkbeck:
@jsnellbaker Excellent, thanks for merging into the legacy branch!

adrianpoix:
@jsnellbaker & @dbirkbeck thanks for merging !

I use Audience Network bid adapter and they didn't migrate to 3.x.x yet so I have to keep 2.x.x for now

Issue #4348, opened 2019-10-21
Stabilize  circleci tests


mkendall07:
## Type of issue
21 of the last 30 builds failed on master. We need to pinpoint and fix the issues to make tests stable again. 

https://circleci.com/gh/prebid/Prebid.js/tree/master

snapwich:
I've done some initial digging into the test issues; some of these issues I'm seeing could possibly be related to external requests being successfully made during run of the test suite, especially the ones that successfully load third-party code and run it.

I've categorized all the external requests I see being made by the test suite by the type of request. We should probably take efforts to see how we can stop these requests from being made as part of the test running.

I know I did some initial work by by mocking request makers, such as `loadExternalScript` but still some are getting through; specifically one I looked at got through because `loadExternalScript` is only stubbed _during_ a test (using before/after hooks) and this module is actually making external requests while the module is being loaded, not during a test.

A lot are ajax requests that I thought we stubbed but are getting through somehow?

The currency module is being loaded by some test suites (besides the currency module tests) and so requests are being made to the currency file.

Some adapters are appending third-party code and running it which I've marked as `custom script appending (without loadExternalScript)`. 

The adapters loading and running external code seem the most problematic since we don't know what they're doing and they could technically doing _anything_, including changing over time, and affecting our tests.

```

loadExternalScript
	- modules/adagioBidAdapter.js
 		- http://script.4dex.io/localstore.js # from firing loadExternalScript outside of test
 		- https://script.4dex.io/adagio.js # third-party script loaded from localstore.js
 		- http://www3.smartadserver.com/ac?pgid=1177189&insid=9202901&tmstp=8180072780&out=js # request initiated by adagio.js script above
	- modules/djaxBidAdapter.js
		- https://acdn.adnxs.com/video/outstream/ANOutstreamVideo.js # installing custom renderer in interpretResponses
	- modules/vubleBidAdapter.js
		- http://localhost:9876/vuble_renderer.js # installing custom renderer from mocked responses in vubleBidAdapter_spec.js

ajax requests
	- modules/djaxBidAdapter.js
		- 200 xhr: https://demo.reviveadservermod.com/headerbidding_adminshare/www/admin/plugins/Prebid/tracking/track.php # firing tracking requests in interpretResponses, seems to be happening multiple times maybe asynchronously?
	- modules/digiTrustIdSystem.js
[...]

PR #4687, opened 2020-01-02
replace all xhr stubs with global xhr stub to prevent all requests

snapwich:
## Type of change
- [x] Bugfix

## Description of change
Replaces all sinon fake xhr and fake server stubbing with global stub to prevent any XMLHttpRequests from happening.

Update all pixels to use utils.triggerPixel and properly stub in tests.

## Other information
Work related to #4348


snapwich:
thanks @jsnellbaker, just pushed a fix for those.

this isn't quite ready for review yet, still in progress. hoping to find some fixes for the other issues listed in #4348 as well (script includes, pixels firing, etc).

snapwich:
should be ready for review. fixes all external requests other than pixels (which are low priority because they don't load executable code, but should probably be fixed at some point).

snapwich:
@jsnellbaker updated

Issue #4674, opened 2019-12-23
DigiTrust UserId module needs to check gdprConsent


goosemanjack:
## Type of issue
This is a bug where behavior of the Prebid integration differs from the baseline DigiTrust library.

## Description
The DigiTrust integration should be specifically checking the transparency and consent framework for the DigiTrust publisher Id. This is in addition to the normal GDPR checks that Prebid performs.

## Other information
View this page for more information.
https://github.com/InteractiveAdvertisingBureau/GDPR-Transparency-and-Consent-Framework/blob/master/TCFv2/IAB%20Tech%20Lab%20-%20CMP%20API%20v2.md

goosemanjack:
A new PR has been created to address this ticket.
PR #4698 
https://github.com/prebid/Prebid.js/pull/4698

goosemanjack:
Touching to un-stale. PR is waiting on review.

patmmccann:
This is related to #4747

goosemanjack:
This issue has been resolved and merged.

PR #4698, opened 2020-01-07
#4674 Addition of DigiTrust vendor check for gdpr when cmp is present.

goosemanjack:
## Type of change
- [X ] Bugfix

## Description of change
This adds in the CMP vendor check for DigiTrust to the DigiTrust ID system module. When cmp is present and GDPR applies, this performs a check for consent specifically against the DigiTrust vendor ID to match behavior in the standalone DigiTrust ID system.

## Other information
Issue #4674 
https://github.com/prebid/Prebid.js/issues/4674

### DIFF
--- a/modules/digiTrustIdSystem.js
+++ b/modules/digiTrustIdSystem.js
@@ -39,6 +39,7 @@ var noop = function () {
 
 const MAX_RETRIES = 2;
 const DT_ID_SVC = 'https://prebid.digitru.st/id/v1';
+const DT_VENDOR_ID = 65; // cmp gvlVendorId
 
 var isFunc = function (fn) {
   return typeof (fn) === 'function';
@@ -120,7 +121,7 @@ function initDigitrustFacade(config) {
         }
       }
 
-      if (!isMemberIdValid) {
+      if (!isMemberIdValid(obj.member)) {
         if (!isAsync) {
           return errResp
         } else {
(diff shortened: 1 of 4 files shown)


modules/digiTrustIdSystem.js
```diff
@@ -39,6 +39,7 @@ var noop = function () {
 
 const MAX_RETRIES = 2;
 const DT_ID_SVC = 'https://prebid.digitru.st/id/v1';
+const DT_VENDOR_ID = 65; // cmp gvlVendorId
```
samtingleff: this should be 64

test/spec/modules/digitrustIdSystem_spec.js
```diff
@@ -5,9 +5,39 @@ import {
 
 let assert = require('chai').assert;
 let expect = require('chai').expect;
-
+const DIGI_VENDOR_ID = 65;
```
samtingleff: also should be 64

goosemanjack:
Touching issue. Please review.

benjaminclot:
Giving consent by default (timeout or else) is not GDPR compliant.

goosemanjack:
@benjaminclot Thanks for your feedback and bringing this up. I will have our team review the behavior and adjust as necessary for compliance as this is a secondary check on top of the Prebid GDPR check.

goosemanjack:
Opened issue #4904

PR #4703, opened 2020-01-08
Adtelligent bid adapter updated to v3.0

GeneGenie:
## Type of change
- [x] AdtelligentBidAdapter support for 3.0

## Description of change
Remove deprecated code 
Tests updated
Do not trigger same usersyncs across all auctions

- contact email of the adapter’s maintainer g.moroz@adtelligent.com


modules/adtelligentBidAdapter.js
```diff
@@ -0,0 +1,246 @@
...
+    const extMessage = serverResponse && serverResponse.ext && serverResponse.ext.message ? `: ${serverResponse.ext.message}` : '';
+    const errorMessage = `in response for ${bidderRequest.bidderCode} adapter ${extMessage}`;
+
+    utils.logError(errorMessage);
+
+    return bids;
+  }
+
+  serverResponse.bids.forEach(serverBid => {
+    const requestId = findIndex(bidderRequest.bids, (bidRequest) => {
```
jaiminpanchal27: Any strong reason to use find-index from core-js. We have a white list of core-js modules here https://github.com/prebid/Prebid.js/blob/master/allowedModules.js#L2 
Can you update the logic ? 
Also, we have a custom eslint rule but not sure why it did not catch this import. I will look into this.

GeneGenie: nope, just left from previous dev =)

GeneGenie: so, can i use Array.prototype.findIndex ?

jaiminpanchal27: @GeneGenie Array.prototype.findIndex does not support Internet explorer and hence the previous dev used core-js to support IE. I would suggest to use `find` from core-js

Here is the updated code which worked fine on my test page. Do test in on your end as well.

```
const request = find(bidderRequest.bids, (bidRequest) => {
      return bidRequest.bidId === serverBid.requestId;
    });

    if (serverBid.cpm !== 0 && request !== undefined) {
      const bid = createBid(serverBid, getMediaType(request));

      bids.push(bid);
    }
```

GeneGenie: fixed

PR #4706, opened 2020-01-08
SpotX Bid Adapter: Add videoCacheKey

vseventer:
## Type of change
- [ ] Bugfix
- [X] Feature
- [ ] New bidder adapter   
- [ ] Build related changes
- [ ] CI related changes
- [ ] Other

## Description of change
Adds `videoCacheKey` to SpotX bids, as SpotX already caches their bids (see `vastUrl`) and therefore we do not need to leverage Prebid Cache.

## Other information
N/A

### DIFF
--- a/modules/spotxBidAdapter.js
+++ b/modules/spotxBidAdapter.js
@@ -308,6 +308,7 @@ export const spec = {
             channel_id: serverResponseBody.id,
             cache_key: spotxBid.ext.cache_key,
             vastUrl: 'https://search.spotxchange.com/ad/vast.html?key=' + spotxBid.ext.cache_key,
+            videoCacheKey: spotxBid.ext.cache_key,
             mediaType: VIDEO,
             width: spotxBid.w,
             height: spotxBid.h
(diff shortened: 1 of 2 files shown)


khatibda:
so here's a question about this, and other updates like this.  if publishers don't update their GAM creatives, immediately then when this goes live, will publishers be sending an unexpected cache url to GAM, and since there won't be a matching creative for it, it won't work?  or am i missing something?

vseventer:
@khatibda A publisher would only get this update if they would upgrade their Prebid.js version which, for a change like this, would probably be a minor (?) and I would hope one would consult the changelog before making the update.

Regardless, this PR follows the Prebid.js docs where `videoCacheKey` can be provided by the bidder to avoid re-caching by Prebid Cache. This works going forward, since SpotX caches their own bid.

robertrmartinez:
@vseventer Do you want me to merge this in before todays release?

vseventer:
@robertrmartinez Would be great, thank you for your review!

josephtyler:
> @khatibda A publisher would only get this update if they would upgrade their Prebid.js version which, for a change like this, would probably be a minor (?) and I would hope one would consult the changelog before making the update.
> 
> Regardless, this PR follows the Prebid.js docs where `videoCacheKey` can be provided by the bidder to avoid re-caching by Prebid Cache. This works going forward, since SpotX caches their own bid.

I know I'm a bit late here, and I realize this change is in line with the specification. But I think what Danny is asking may be a higher level question. If we upgrade prebid, we will have to create new line items in GAM in order for spotx to continue working. Or is there a way to continue using the original cache?

robertrmartinez:
@josephtyler 

You are correct, if you were using a prebid cache host like appnexus or rubicon, then unfortunately you would need to update spotX VAST creatives as it will no longer be fetch-able at those locations.

There is a way to "hack" this to work as before, but I would first take something like this up with SpotX.

PR #4710, opened 2020-01-09
Find gpt div based on adId to be resized instead of hardcoded adunitcode / slotelementid

ptomasroos:
## Type of change

- [x] Bugfix
- [x] Feature

## Description of change

At this moment using http://prebid.org/dev-docs/publisher-api-reference.html#module_pbjs.setTargetingForGPTAsync with customSlotMatching, secureCreatives are unsupported.

We discovered this when trying to switch to secure creatives. The reason is that it expects that adUnitCode should match the adUnitPath or slotElementId of the gpt slot.

The reason behind customSlotMatching is just the thing that we can target any div arbitrary of what the ad unit code in pbjs or what the ad unit of the gpt is named.

So instead of using these fixed values we could solve this by checking the slot's targetkeys for the hb_adid. Works for older use cases as well as supporting customSlotMatching (more flexibility)


ptomasroos:
@jsnellbaker

jsnellbaker:
Retagging original reviewers from #4374

src/secureCreatives.js
```diff
@@ -94,23 +95,25 @@ function resizeRemoteCreative({ adUnitCode, width, height }) {
...
       return adUnitCode;
     }
   }
 
-  function getDfpElementId(adUnitCode) {
-    return find(window.googletag.pubads().getSlots().filter(isSlotMatchingAdUnitCode(adUnitCode)), slot => slot).getSlotElementId()
+  function getDfpElementId(adId) {
+    return find(window.googletag.pubads().getSlots().filter(s =>
+      includes(s.getTargeting('hb_adid'), adId)
+    ), slot => slot).getSlotElementId();
```
andyblackwell: Looks like this won't work for "send all bids" where the winning bid wasn't the highest bid (which happens sometimes)

ptomasroos: All right! I will iterate all values then!

ptomasroos: @andyblackwell I've now updated the logic to search through all keys so it would find deal targeting or if something was targeted without hb_adid. Thanks for the spot!

ptomasroos:
I've rebased it, would be great it we could merge it

ptomasroos:
All right, now tests are failing after rebasing, will check it out

ptomasroos:
All right, this is now aligned with master and I've resolved the broken behaviours regarding how google tag expects in some tests.

google tag never returns a object with key, values when invoking getTargeting(), getTargeting is always expecting a key being passed in.

Its ready to get merged. Would be great if we could get approval

Screenshot 2020-02-13 at 09 02 50

ptomasroos:
All tests are passing

ptomasroos:
Can someone let me know when its ready for a review
3rd time its out of sync with master now and needs rebasing

src/secureCreatives.js
```diff
@@ -94,23 +95,27 @@ function resizeRemoteCreative({ adUnitCode, width, height }) {
...
       return getAstElementId(adUnitCode)
     } else {
       return adUnitCode;
     }
   }
 
-  function getDfpElementId(adUnitCode) {
-    return find(window.googletag.pubads().getSlots().filter(isSlotMatchingAdUnitCode(adUnitCode)), slot => slot).getSlotElementId()
+  function getDfpElementId(adId) {
+    return find(window.googletag.pubads().getSlots(), slot => {
```
msm0504: Since this is only 1 statement, `{ return }` is not needed. Same applies to following line.

ptomasroos: Now resolved conflicts with master, hopefully for the last time 😉

PR #4720, opened 2020-01-13
pbsBidAdapter: defaulting to HTTPS

bretg:
## Description of change

The upcoming Chrome 80 release suppresses cookies to secure sites that don't have the 'secure' flag. But turns out it works the other way around too -- chrome suppresses 'secure' cookies on non-secure connections.

What this means from a Prebid perspective is that all connections to a backend service need to be HTTPS, even if the site itself is non-secure.

So this PR changes Rubicon's default Prebid Server endpoint to HTTPS. And I went ahead and made the same change for adnxs.com assuming you guys are likely to have the same problem.

### DIFF
--- a/modules/prebidServerBidAdapter/config.js
+++ b/modules/prebidServerBidAdapter/config.js
@@ -3,15 +3,15 @@ export const S2S_VENDORS = {
   'appnexus': {
     adapter: 'prebidServer',
     enabled: true,
-    endpoint: '//prebid.adnxs.com/pbs/v1/openrtb2/auction',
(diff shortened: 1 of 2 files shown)


bretg:
Logging a warning would be a friendly thing to do as an interim for a few months. But with PBJS people upgrade to a version and stay with it for years. And log messages don't compress well, so it ought be something we think will be concise. Maybe "Prebid server endpoints should be secure"

bretg:
Checked with @SyntaxNode -- he confirmed this change is ok for the adnxs hosted cluster. Opened an issue to track the warning message.

PR #4732, opened 2020-01-14
if seatBidId is on bidResponse, always use it

robertrmartinez:
## Type of change

- [X] Bugfix -> Not really a "bug" just code cleanup

## Description of change
For Rubicon analytics adapter if there is a `seatBidId` on the bidResponse object then we will be sure to use that as the `bidId` we send to the batched endpoint.

This is just removing a couple unnecessary checks and adding a test.

### DIFF
--- a/modules/rubiconAnalyticsAdapter.js
+++ b/modules/rubiconAnalyticsAdapter.js
@@ -416,10 +416,6 @@ let rubiconAdapter = Object.assign({}, baseAdapter, {
         }
         bid.clientLatencyMillis = Date.now() - cache.auctions[args.auctionId].timestamp;
         bid.bidResponse = parseBidResponse(args, bid.bidResponse);
-        // RP server banner overwrites bidId with bid.seatBidId
-        if (utils.deepAccess(bid, 'bidResponse.seatBidId') && bid.bidder === 'rubicon' && bid.source === 'server' && ['video', 'banner'].some(i => utils.deepAccess(bid, 'bidResponse.mediaType') === i)) {
-          bid.seatBidId = bid.bidResponse.seatBidId;
-        }
         break;
       case BIDDER_DONE:
         args.bids.forEach(bid => {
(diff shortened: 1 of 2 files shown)


test/spec/modules/rubiconAnalyticsAdapter_spec.js
```diff
@@ -224,7 +224,6 @@ const MOCK = {
         'transactionId': 'c116413c-9e3f-401a-bee1-d56aec29a1d4',
         'sizes': [[1000, 300], [970, 250], [728, 90]],
         'bidId': '3bd4ebb1c900e2',
-        'seatBidId': 'aaaa-bbbb-cccc-dddd',
```
robertrmartinez: This is the mock of the `bidRequestObject`, which would never have `seatBidId` in it since `seatBidId` is a bidResponse field.

So not realistic to have it here.

modules/rubiconAnalyticsAdapter.js
```diff
@@ -416,10 +416,6 @@ let rubiconAdapter = Object.assign({}, baseAdapter, {
         }
         bid.clientLatencyMillis = Date.now() - cache.auctions[args.auctionId].timestamp;
         bid.bidResponse = parseBidResponse(args, bid.bidResponse);
-        // RP server banner overwrites bidId with bid.seatBidId
-        if (utils.deepAccess(bid, 'bidResponse.seatBidId') && bid.bidder === 'rubicon' && bid.source === 'server' && ['video', 'banner'].some(i => utils.deepAccess(bid, 'bidResponse.mediaType') === i)) {
-          bid.seatBidId = bid.bidResponse.seatBidId;
-        }
```
robertrmartinez: The line `bid.bidResponse = parseBidResponse(args, bid.bidResponse);`

Adds `seatBidId` to the internal `bidResponse` in the analytics adapter. And if that is present, it overwrites the 
 internal `bidRequest.bidId` which is sent to Analytics.

This `bid.seatBidId` is not referenced or used anywhere else in the code.

So does not make sense to add it to this bid object.

PR #4735, opened 2020-01-15
Sonobi - removed unneeded query params

JonGoSonobi:
## Type of change

## Description of change
Removed TDID param from bid request since it is passed into the userid param now.
Removed pub common id as a value for hfa query param

- apex.prebid@sonobi.com


bretg:
Looks fine, but please look over the params on your documentation page. Looks like maybe you want to get rid of the 'hfa' option there?  https://github.com/prebid/prebid.github.io/blob/master/dev-docs/bidders/sonobi.md

JonGoSonobi:
@bretg hfa is still an option. We just dont want to default it to pub common id anymore.

PR #4748, opened 2020-01-17
Bridgewell bid adapter: Prebid.js v3 support

AlfredKai:
* Add bridgewellBidAdapter prebid.js v3 support
* Remove deprecated functionality
* Refactor tests

### DIFF
--- a/modules/bridgewellBidAdapter.js
+++ b/modules/bridgewellBidAdapter.js
@@ -103,7 +103,7 @@ export const spec = {
                 if (typeof sizes[0] === 'number') { // for foramt Array[Number] check
                   valid = width === sizes[0] && height === sizes[1];
                 } else { // for format Array[Array[Number]] check
-                  valid = find(sizes, function (size) {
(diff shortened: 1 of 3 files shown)


modules/bridgewellBidAdapter.js
```diff
@@ -0,0 +1,266 @@
...
+            } else if (mediaTypes.banner) {
+              if (mediaTypes.banner.sizes) {
+                let width = res.width;
+                let height = res.height;
+                let sizes = mediaTypes.banner.sizes;
+                // check response size validation
+                if (typeof sizes[0] === 'number') { // for foramt Array[Number] check
+                  valid = width === sizes[0] && height === sizes[1];
+                } else { // for format Array[Array[Number]] check
+                  valid = find(sizes, function (size) {
```
msm0504: `valid` will be set equal to the element returned from `find`. Should be a boolean instead?

AlfredKai: fixed, thank you.

PR #4759, opened 2020-01-20
Support Native

bjorn-lw:
## Type of change
- [x ] Feature

## Description of change
Support native ads + get banner size from mediaTypes.banner


bjorn-lw:
Don't think the ci/circleci issue is related to this PR?

modules/livewrappedBidAdapter.js
```diff
@@ -177,14 +184,33 @@ function hasPubcid(bid) {
...
     callerAdUnitId: bid.params.adUnitName || bid.adUnitCode || bid.placementCode,
     bidId: bid.bidId,
     transactionId: bid.transactionId,
-    formats: bid.sizes.map(sizeToFormat),
+    formats: getSizes(bid).map(sizeToFormat),
     options: bid.params.options
   };
+
+  if (bid.mediaTypes && bid.mediaTypes.banner && bid.mediaTypes.native) {
+    adRequest.banner = true;
```
robertrmartinez: Why does setting this matter if `native` is in the mediaTypes?

modules/livewrappedBidAdapter.js
```diff
@@ -177,14 +184,33 @@ function hasPubcid(bid) {
...
+  }
+
+  if (bid.mediaTypes && bid.mediaTypes.native) {
```
robertrmartinez: I have a personal connection with using the `deepAccess` util for these kinds of checks.

`if(deepAccess(bid, 'mediaTypes.native')`

But definitely not necessary, just like to point it out so people are aware of it!

modules/livewrappedBidAdapter.js
```diff
@@ -177,14 +184,33 @@ function hasPubcid(bid) {
...
+    adRequest.native = bid.mediaTypes.native;
+  }
+
+  return adRequest;
+}
+
+function getSizes(bid) {
+  if (typeof utils.deepAccess(bid, 'mediaTypes.banner.sizes') !== 'undefined') {
```
robertrmartinez: I see you already are using deepAccess! nice :)

Can probably be changed to leave off the `typeof {} !== 'undefined'` maybe?

Since if it was not defined it will just return `undefined` which is falsely.

No big deal though.

modules/livewrappedBidAdapter.js
bjorn-lw: It's related to how it works in the backend. It only needs setting if both banner and native is enabled, otherwise (if not native) banner is implied.

Btw. thanks for all the input, I will fix those things for the next PR.

PR #4760, opened 2020-01-20
IE fix: calling matchMedia with an empty string (prebid-2.0)

benjaminclot:
## Type of change
- [x] Bugfix

## Description of change
Same as #4691 but for prebid-2.0


benjaminclot:
The build fail seems to be unrelated to this PR as the branch currently fails building.

PR #4777, opened 2020-01-23
Update underdogmedia adapter to contain getUserSyncs

mash-a:
## Type of change

- [x] Feature

## Description of change

Underdogmedia adapter now has getUserSyncs function. 

   jake@underdogmedia.com


modules/underdogmediaBidAdapter.js
```diff
@@ -61,6 +68,23 @@ export const spec = {
     }
   },
 
+  getUserSyncs: function (syncOptions, serverResponses) {
+    if (!USER_SYNCED && serverResponses.length > 0 && serverResponses[0].body && serverResponses[0].body.userSyncs) {
```
jsnellbaker: When I checked the test bid (from the .md file), I observed that the response had `serverResponses[0].body.userSyncs` set to `[]`.

Even though the array is empty - it's still a valid value for the variable, so it passed the if-check here.  I wanted to highlight this behavior in-case you see it as an issue.

mash-a:
@jsnellbaker absolutely right. I adjusted the logic to prevent user syncs from loading when `serverResponses[0].body.userSyncs` is set to `[]` and pushed the changes to the PR.

PR #4778, opened 2020-01-24
Fix/legacy syncurl

sourabhg:
- [x] Bugfix


sourabhg:
CircleCI tests seems to be failed due to a third party adapter 

`at Object.buildRequests (webpack:///modules/emoteevBidAdapter.js:503:26 `

Issue #4768, opened 2020-01-22
bid adpater's onBidWon function not called for instream video bids


jsnellbaker:
## Type of issue
Bug

## Description
While reviewing the surrounding code for something else, I noticed that the `auctionManager.addWinningBid` logic used to call a bid adapter's `onBidWon` function was not being triggered for `instream` video bids.

The `addWinningBid` logic is invoked in the following areas for the other media types:
- banner > https://github.com/prebid/Prebid.js/blob/master/src/prebid.js#L317
- outstream video > https://github.com/prebid/Prebid.js/blob/master/src/prebid.js#L317
- native > https://github.com/prebid/Prebid.js/blob/master/src/prebid.js#L317 & https://github.com/prebid/Prebid.js/blob/master/src/secureCreatives.js#L58 (since there different paths to handle for native)

I believe this is just an unfortunate oversight and not an intentional decision to omit `instream` video.  

I will work on putting together a fix soon.

ChrisHuie:
closing issue from #5481

PR #4783, opened 2020-01-24
add missing addWinningBid logic to instream + adpod video bids

jsnellbaker:
## Type of change

- [x] Bugfix

## Description of change
Fix for #4768 

This PR:
- adds missing calls to load the analytics event `BID_WON` and bid adapter's `onBidWon()` for instream + adpod video bids
- refactors the analytics event `BID_WON` to be included in the `auction.addWinningBid()` function so both actions are carried together (which reduces the chance one is missed by accident)


jsnellbaker:
I resolved the conflicts that came up in this PR.  Please take a look when you have the chance.

idettman:
Can you update you branch with the latest from master, I'll finish my review once it's up to date

jsnellbaker:
Hi @idettman 

I just resynced this PR with `master`.  Please take another look when you have the chance.

Thanks!

idettman:
Lastly, can you resolve the merge conflict with `src/secureCreatives.js`?

jsnellbaker:
Hi @idettman 
I have resynced with master and resolved the conflict.  Please take another look when you have the chance.

pm-harshad-mane:
Hello @jsnellbaker ,
This fix/improvement will help an analytics adapter know when an instream bid is rendered, correct?

jsnellbaker:
@pm-harshad-mane I actually had a discussion on this with @mike-chowla yesterday.  With what has been put together, this PR will help analytics adapters know when we have finished preparing the bid to be used by the player - not necessarily that the player actually rendered the bid.  That distinction is important here, as it could track false positives and report a discrepancy with other tracking/metrics.  

Ideally, we'd want to track only when either the player is actually loading the VAST ad successfully or at least when PBC has received a request to download the cached bid (since we can infer then that DFP has chosen this video bid as the winner and is attempting to render it).  We don't have anything like that (for our analytics events) atm, so we're a bit restricted on that front.

Some additional conversation needs to happen if this would be the right thing to implement.

mike-chowla:
I'm concerned that BID_WON will have different meanings between display (or out-stream video) and in-stream video.

For display, BID_WON gets invoke after the bid wins at the ad server and right before the ad is rendered.

For in-stream video, BID_WON would get invoked before bids are sent to the ad server.

This seems confusing to me.  I think there are use-cases worth supporting for an event that is called before one Prebid complete's its auction and before the ad server is called.  I'm thinking that event should have a different name.

However, using a different name would mean there is no BID_WON event for in-stream video as short of doing integration with video player, players there's no way to get a Javascript callback on the page to say the ad is rendering.  When I talked @jsnellbaker about this, his thinking for this PR was that BID_WON would be signify the last point where Prebid touches the bid.

At the last Video task force meeting, there was a desire from multiple parties to be able to add VAST trackers from Prebid for win and error events.  Those would be server-server side trackers which won't fire the BID_WON event on the page.

@mkendall07 @bretg  Any thoughts on to proceed here?

bretg:
I agree with @mike-chowla that video events recorded _before_ the ad server should have a different name. They are definitely not BIDS_WON.

Rubicon has been building support for SDK, AMP and video events for a while:

Prebid Server:
- https://github.com/prebid/prebid-server/issues/800
- https://github.com/prebid/prebid-server/issues/1015
- https://github.com/prebid/prebid-server/issues/1202

Prebid.js
- https://github.com/prebid/Prebid.js/issues/4999

Prebid Universal Creative
https://github.com/prebid/prebid-universal-creative/issues/70
- https://github.com/prebid/prebid-universal-creative/issues/93

We've published a doc describing the event model at https://docs.google.com/document/d/1ry0X4C2EV-R0pMrm1IQk9BstxaT395UCl3KKqTGa5c8/edit#heading=h.7w5yevygp2gz

Specifically for video, PrebidServer-Java has the ability to inject VAST <impression> events into the cached XML in two ways: either automatically when it comes through a server-side bid adapter, or with a new /vasttrack endpoint when it comes client-side. (Note it also has the ability to turn off modifying the VAST on a per-bidder and per-account basis)

Am happy to schedule a session with you guys to go through the current event model and brainstorm adpod and 3rd party tracking.

bretg:
@jsnellbaker - I propose closing this ticket. Or at least modify it significantly to register a different type of event -- it's not a BID_WON.

PR #4789, opened 2020-01-28
Pass auctionId as a argument to bidsBackHandler

ptomasroos:
Since many people are using a lambda expression its very convenient to have the auctionId being returned into the bidsBackHandler argument

## Type of change
- [x] Feature

## Description of change
Adds auctionId as an argument to the existing bidsBackHandler call.
We use it in order to be able to take decisions in the bidsBackHandler which is the current auction that is going to be rendered.

Since the other place the auctionId is first return is the requestBids (which is outside of the Clojure and the bidsBackHandler that is being passed as a object)


ptomasroos:
Thanks!

PR #4802, opened 2020-01-30
enable no-console eslint rule for project

jsnellbaker:
## Type of change
- [x] Other

## Description of change
There have been a number of PRs in the past where the unit tests included some rogue `console.log` statements and they were accidentally missed by the reviewers.

While not a big deal, it does end up polluting the console messages when we run the tests (especially the browserstack tests, since they get displayed for each browser), and it becomes somewhat tedious to track them down later and create another PR just to remove them.

I made this PR to enable the no-console rule for the eslint rules in our `test` directory to hopefully help us better catch these instances during the review process.  

I also updated the various files in the `test` directory using `console.log`; either removing them or disabling the rule.

**Update**
Per feedback, I have extended this lint logic to the project rather than just the tests directory.


test/spec/modules/onetagBidAdapter_spec.js
```diff
@@ -81,7 +81,7 @@ describe('onetag', function () {
         }
       });
     } catch (e) {
-      console.log('Error while parsing');
+      console.log('onetagBidAdapter tests: Error while parsing');  // eslint-disable-line
```
mkendall07: any reason to keep these these?

jsnellbaker: I thought at first they would be good to keep for the adapter's tests, but in retrospect they're not too helpful.  I pushed an update to remove them.

jsnellbaker:
Hi @snapwich 

Expanding the scope does make sense, I'll try to make the necessary updates soon.

jsnellbaker:
I pushed in the requested updates.  Please take another look when you get the chance.

Thanks!

jsnellbaker:
Hey @snapwich @mkendall07 , can you please take another look and approve (if it looks good now)?  Thanks!

Issue #3878, opened 2019-06-04
Automated building fails since 2.12


benjaminclot:
## Type of issue
Since Prebid 2.12, our automated building process has been failing. It runs on a GitLab worker with a Docker node:latest image.

## Description
The culprit is the `fibers` module:
```
> fibers@3.1.1 install /builds/b.clot/build-and-deploy/Prebid.js/node_modules/fibers
> node build.js || nodejs build.js

make: Entering directory '/builds/b.clot/build-and-deploy/Prebid.js/node_modules/fibers/build'
  CXX(target) Release/obj.target/fibers/src/fibers.o
../src/fibers.cc:68:2: error: 'Handle' does not name a type
  Handle<String> NewLatin1String(Isolate* isolate, const char* string) {
  ^~~~~~
../src/fibers.cc:72:2: error: 'Handle' does not name a type
  Handle<String> NewLatin1Symbol(Isolate* isolate, const char* string) {
  ^~~~~~
../src/fibers.cc:94:2: error: 'Handle' does not name a type
  Handle<Object> NewInstance(Isolate* isolate, Local<Function> fn, int argc, Local<Value> argv[]) {
  ^~~~~~
../src/fibers.cc:104:2: error: 'Handle' does not name a type
  Handle<Number> ToNumber(Local<Value> value) {
  ^~~~~~
../src/fibers.cc:114:50: error: 'Handle' has not been declared
  Local<Value> GetStackTrace(TryCatch* try_catch, Handle<Context> context) {
                                                  ^~~~~~
../src/fibers.cc:114:56: error: expected ',' or '...' before '<' token
  Local<Value> GetStackTrace(TryCatch* try_catch, Handle<Context> context) {
                                                        ^
../src/fibers.cc: In function 'v8::Local<v8::Value> uni::GetStackTrace(uni::TryCatch*, int)':
../src/fibers.cc:115:32: error: 'context' was not declared in this scope
   return try_catch->StackTrace(context).ToLocalChecked();
                                ^~~~~~~
../src/fibers.cc: At global scope:
../src/fibers.cc:127:39: error: 'Handle' has not been declared
  void fixStackLimit(Isolate* isolate, Handle<Context> context) {
                                       ^~~~~~
../src/fibers.cc:127:45: error: expected ',' or '...' before '<' token
  void fixStackLimit(Isolate* isolate, Handle<Context> context) {
                                             ^
[...]

PR #4820, opened 2020-02-04
Dependency update to support node12

renebaudisch:
## Type of change
- [x] Build related changes

## Description of change
changes dependencies to support node v12

## Other information
related to already closed issue 
https://github.com/prebid/Prebid.js/issues/3878

but problem still exists for node12 and maybe > 8.9 in general


jsnellbaker:
Hi @renebaudisch 

Thanks for submitting this PR.  

While I was digging into it, I found there are a number of other changes that will need to be made to our implementation of wdio and our e2e tests in order to use wdio 5.x (as they appear to have made a few breaking changes between 4 & 5).

So we won't be able to implement this PR as is.   However, I think the overall update still has value, so I will start to work on putting together the necessary changes for our tests and will open a new PR with the combined updates.

renebaudisch:
THX for the information, the rest works pretty well under node v12, I builded and released a local 3.5.0 version without problems.

PR #4841, opened 2020-02-11
Add two more default native params to seedingalliance adapter

SeedingAllianceTech:
## Type of change
- [X] Feature

## Description of change
Add two more default native params to the seedingalliance adapter.


GLStephen:
LGTM

Issue #1284, opened 2017-06-12
Prebid's iframe in AMP


rafaguerrero:
## Type of issue
Performance Bug

## Description
I've implemented the solution that you propose for integrating AMP with Prebid and I found that my revenue are way lower than expected.

I think this might be related to this issue: https://github.com/prebid/Prebid.js/issues/880

Due to the increase in AMP traffic this is a major concern on my end.
Is there anything to do about it?

bretg:
Some of the adapters support defining the referring URL, documented or not. e.g. Both Rubicon and AppNexus support passing bid.params.referrer from the AdUnit.

Perhaps you could check the code for the adapters you're using and try passing the referrer param for any that support it?

Instead of having to pass this in for each bidder, we should be able to add referrer to the standard bidRequest. It would still be up to the adapter to use it.

bretg:
@rafaguerrero - any update on AMP? Were you able to try passing referrer?

matthewlane:
Closing as stale, please reopen if there are updates

Issue #880, opened 2016-12-15
AMP canonical URL


oliverfernandez:
This is question, related with AMP support for header bidding.

Since the bid is happening inside the 3P Iframe, are bidders receiving the URL of this iframe, instead of the canonical one?

If this is the case, and since inside the 3P iframe, you have access to global.context.canonicalUrl, is there a way to send this URL when performing the bid?

mkendall07:
@oliverfernandez 
Today, no. It would be up to each bidder adapter to send this through (assuming we provide the mechanism to do it).

PR #4846, opened 2020-02-12
Refactor refererDetection to allow for URL discovery on AMP pages.

mgriego:
## Type of change

- [ ] Bugfix
- [x] Feature
- [ ] New bidder adapter   
- [ ] Build related changes
- [ ] CI related changes
- [ ] Other

## Description of change
Updates the referrer detection code to allow discovery of and attribution to original URLs on AMP pages served from caches.  Info on the AMP ad iframe context information can be found here: https://github.com/ampproject/amphtml/blob/master/ads/README.md#available-information-to-the-ad

I will provide this tomorrow.  There is a single new property on the object returned by `getRefererInfo`: `isAmp` is `true` if the detected referrer was obtained by the AMP context.

## Other information
Related issues:
https://github.com/prebid/Prebid.js/issues/880
https://github.com/prebid/Prebid.js/issues/1284

### DIFF
--- a/src/refererDetection.js
+++ b/src/refererDetection.js
@@ -8,7 +8,7 @@
  * Canonical URL which refers to an HTML link element, with the attribute of rel="canonical", found in the <head> element of your webpage
  */
 
-import { logWarn } from './utils';
+import { logWarn } from './utils.js';
 
 /**
  * @param {Window} win Window
--- a/src/refererDetection.js
+++ b/src/refererDetection.js
@@ -10,33 +10,24 @@
 
 import { logWarn } from './utils';
 
+/**
+ * @param {Window} win Window
+ * @returns {Function}
+ */
 export function detectReferer(win) {
-  /**
-   * Returns number of frames to reach top from current frame where prebid.js sits
-   * @returns {Array} levels
-   */
-  function getLevels() {
-    let levels = walkUpWindows();
-    let ancestors = getAncestorOrigins();
-
-    if (ancestors) {
-      for (let i = 0, l = ancestors.length; i < l; i++) {
-        levels[i].ancestor = ancestors[i];
-      }
-    }
-    return levels;
-  }
-
   /**
    * This function would return a read-only array of hostnames for all the parent frames.
    * win.location.ancestorOrigins is only supported in webkit browsers. For non-webkit browsers it will return undefined.
+   *
+   * @param {Window} win Window object
    * @returns {(undefined|Array)} Ancestor origins or undefined
    */
-  function getAncestorOrigins() {
+  function getAncestorOrigins(win) {
     try {
       if (!win.location.ancestorOrigins) {
         return;
       }
+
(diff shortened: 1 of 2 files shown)


mgriego:
Looks like CI failed on an unrelated item?

bretg:
@mgriego - I assume this is for a post-bid scenario? 

This is a pretty big change to the referrer detection section. Will leave it to @jaiminpanchal27 to work through the details or assign to someone else.

bretg:
Docs PR https://github.com/prebid/prebid.github.io/pull/1803

mgriego:
Postbid is certainly one use case, but it's not the only one.  We have run directly inside an amp-ad iframe as well.

In any case, the code was written to identically replicate the results with the same basic methodology, just simplified a bit in a way that helps to be able to detect AMP embeds.  The only change to the public API is the addition of the new isAmp property and the fact that the referer and canonicalUrl properties will reflect the AMP values in AMP environments.

bretg:
You can run all of Prebid.js within an amp-iframe? I didn't think that was (A) technically possible, which is why the whole AMP-RTC protocol was devised.

And (B) I was given to understand that the AMP runtime limits pages to one amp-iframe.

Please help us understand the tradeoffs here so we can document them on the site.

mgriego:
Indeed you can, and we are.  Each `amp-ad` element contains its own iframe, and we run prebid + other custom code within each iframe.  It might be helpful for me to show you what we're doing and how it works somehow.  AMP doesn't care so much about how many iframes there are, just that custom code runs within the sandboxed iframe off of the main page.

mgriego:
A little more info...  The RTC protocol was made to make it easier to implement bidders on publisher pages.  The only other way to do this is to implement an AMP ad network, and that's what we did.  Ad networks in AMP are allowed to run arbitrary JS inside their sandboxed cross-origin ad iframe.  In our case, we are running Prebid.js via an NPM dependency in our AMP code that runs inside of each ad iframe.  You can see a lot more info on how AMP ad networks work in the readme on the AMP ads page: https://github.com/ampproject/amphtml/tree/master/ads

bretg:
Removing myself as approver. I have no objections. I would like to get better AMP documentation in place on prebid.org, but that can done as part of a separate effort.

mgriego:
I was just coming back here to update the import to match the new style (adding the `.js` extension), but this has been closed.  Should I open it new?  It's stale because I'm still waiting on a review..... @bretg @jaiminpanchal27 @mkendall07

smenzer:
@harpere / @jaiminpanchal27 what else needs to be done here? This one hasn't been touched in a while so we should either wrap it up and merge or close it out.

smenzer:
@robertrmartinez or @harpere can this be merged? it has 3 approvals...

bretg:
Our QA team isn't going to get to this for several weeks at least. We'll go ahead and merge and do extra testing when there are cycles.

PR #4856, opened 2020-02-13
NPM usage doc: Add resolve.extensions

ckybonist:
## Type of change

- [ ] Bugfix
- [ ] Feature
- [ ] New bidder adapter   
- [ ] Build related changes
- [ ] CI related changes
- [x] Other

## Description of change

Add `resolve.extensions` to the [section](https://github.com/prebid/Prebid.js#usage-as-a-npm-dependency)

```
{
  bidder: '<bidder name>',
  params: {
    // ...
  }
}
```

## Other information

- https://github.com/prebid/Prebid.js/blob/master/src/utils.js#L6
- https://github.com/prebid/Prebid.js/blob/master/src/constants.json

### DIFF
--- a/README.md
+++ b/README.md
@@ -37,6 +37,9 @@ With Babel 7:
 let path = require('path');
 module.exports = {
   mode: 'production',
+  resolve: {
+    extensions: ['.js', '.json']
+  },
   module: {
     rules: [
       


snapwich:
Does webpack not resolve `.js` and `.json` files by default? I'm pretty sure it does.

ckybonist:
> Does webpack not resolve `.js` and `.json` files by default? I'm pretty sure it does.

Absolutely [it does](https://webpack.js.org/configuration/resolve/#resolveextensions).

However, IMHO, users who maintain typescript-based projects (like me) would definitely specify the `resolve.extensions` for `.ts`; Without the clear documentation, we have to dig into `node_modules/prebid.js` to find out which file causing the error, moreover, the webpack error logs 
 can't even tell us the actual reason why I get tripped...

```
Module not found: Error: Can't resolve '../src/constants' 
in '/Users/FatCat/my-project/node_modules/prebid.js/modules'

// Multiple errors...
```
NOTE: no extension at `../src/constants`

snapwich:
So you're saying you changed `resolve.extensions` to `.ts` and it caused Prebid.js modules to not be found? Seems like expected behavior (hindsight is 20/20 I know).

I'd rather not specify default configuration, or configuration at all, in the webpack config that doesn't relate specifically to Prebid.js as that would include _A LOT_ of stuff besides just resolving extensions depending on what people are doing with webpack.

If being explicit in our code and manually specifying the `.js` extensions in all Prebid.js imports works around this gotcha, that seems like a better solution. Then we can add this [ESLint rule](https://github.com/benmosher/eslint-plugin-import/blob/master/docs/rules/extensions.md) that ensures that all code going forward uses extensions.

snapwich:
this issue should be resolved by #4876 being merged. let me know if this doesn't work for you.

ckybonist:
Actually, I changed `resolve.extensions` to `['.js', '.ts']` at first and that's why I get the errors (missing `.json`).

Thanks for specifying all module extension explicitly! That's truly a BIG PR. 

Waiting for next stable release!

PR #4876, opened 2020-02-19
require extensions in imports

snapwich:
## Type of change

## Description of change
Explicitly require extensions on imports and validate with eslint plugin.

## Other information
This should fix any issues with dependent projects including Prebid.js that don't have the default extensions as part of the module resolver as described in #4856


modules/rubiconBidAdapter.js
```diff
@@ -638,7 +638,7 @@ export const spec = {
               bidObject.adserverTargeting = extPrebidTargeting;
             }
 
-            // try to get cache values from 'response.ext.prebid.cache'
+            // try to get cache values from 'response.ext.prebid.cache.js'
```
pm-harshad-mane: is this change necessary?

test/spec/videoCache_spec.js
```diff
@@ -1,8 +1,7 @@
-import 'mocha/mocha';
```
robertrmartinez: Guessing this is an unused import

pm-harshad-mane:
Looks good to me, but please check the failing CI Job.

test/spec/videoCache_spec.js
snapwich: I sure hope so, I can't imagine what it would be doing where it'd need to be included only for its side-effects... either way I removed it and the tests still passed so... good?

robertrmartinez: 👍

modules/rubiconBidAdapter.js
snapwich: nope, that was an accidental catch by the regex replace

PR #4880, opened 2020-02-21
yieldoneBidAdapter: fix encode referrer twice

yashigani:
## Type of change

- [x] Bugfix

## Description of change
Bugfix for yieldoneAdapter encoded `r` parameter twice.

yieldoneAdapter makes unexpected request like `{ "r": "https%253A%252F%252F...", ... }`.
buildRequests makes encode `r` parameter. But newBidder.processRequest encode query parameters too([here](https://github.com/prebid/Prebid.js/blob/469f82cfe676bcca633f3539b2dad75d4561c13f/src/adapters/bidderFactory.js#L234)). So `r` parameter in bid request is encoded excessively.
yieldoneAdapter for v2 has same issue.


bretg:
LGTM, but adding @kusapan and @koji-eguchi to confirm.

bretg:
No answer from the originators, going ahead with the merge.

kusapan:
@bretg  confirmed. Thank you for the merge.

PR #4881, opened 2020-02-21
feature/bidLimit-dealPrioritization

r-schweitzer:
## Type of change
Feature

## Description of change
Prebid will add a new config option to sendBidsControl. When adUnitBidLimit is set above zero and dealPrioritization is set to true, deals will be prioritized first and sorted by highest to lowest cpm.
```
pbjs.setConfig({ 
    adUnitBidLimit: 2,
    dealPrioritization: true
 })

```

## Other information
The  original PR for adUnitBidLimit: https://github.com/prebid/Prebid.js/pull/3906
Documentatin: https://github.com/prebid/prebid.github.io/pull/1829

### DIFF
--- a/src/targeting.js
+++ b/src/targeting.js
@@ -44,7 +44,6 @@ export function getHighestCpmBidsFromBidPool(bidsReceived, highestCpmCallback, a
     Object.keys(bidsByBidder).forEach(key => bucketBids.push(bidsByBidder[key].reduce(highestCpmCallback)));
     // if adUnitBidLimit is set, pass top N number bids
     if (adUnitBidLimit > 0) {
-      bucketBids.sort((a, b) => b.cpm - a.cpm);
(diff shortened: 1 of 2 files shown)


bretg:
thanks for the PR @r-schweitzer. Docs PR on the way?

src/targeting.js
```diff
@@ -44,6 +45,7 @@ export function getHighestCpmBidsFromBidPool(bidsReceived, highestCpmCallback, a
     // if adUnitBidLimit is set, pass top N number bids
     if (adUnitBidLimit > 0) {
       bucketBids.sort((a, b) => b.cpm - a.cpm);
+      bucketBids = dealPrioritization ? bucketBids(sortByDealAndPriceBucketOrCpm(true)) : bucketBids.sort((a, b) => b.cpm - a.cpm);
```
jsnellbaker: If we're making this sorting type an if/else condition, can we remove the `sort` done on line 47?

jsnellbaker:
@r-schweitzer Please let us know when the docs PR is created.  We should be good to merge at that time.

PR #4885, opened 2020-02-24
Rubicon analytics passes referrerHostname

robertrmartinez:
Type of change
- [X] Feature

## Description of change
Rubicon Analytics Adapter is passing host name now for reporting purposes

### DIFF
--- a/modules/rubiconAnalyticsAdapter.js
+++ b/modules/rubiconAnalyticsAdapter.js
@@ -4,6 +4,7 @@ import CONSTANTS from '../src/constants.json';
 import { ajax } from '../src/ajax.js';
 import { config } from '../src/config.js';
 import * as utils from '../src/utils.js';
+import * as urlLib from '../src/url.js'
 
 const {
   EVENTS: {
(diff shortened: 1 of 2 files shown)


modules/rubiconAnalyticsAdapter.js
```diff
@@ -36,6 +37,13 @@ const cache = {
   timeouts: {},
 };
 
+let referrerHostname;
```
msm0504: Noticed `referrerHostname` is defined outside of where `referrer` is and `getHostNameFromReferer` will only be called once.

Is it possible for `referrer` to be different for different `sendMessage` calls?

robertrmartinez: No this value only needs to be set once since the referrer is the url of the page prebid was loaded.

This should not change no matter how many auctions are run.

Unless I am missing something.

I just did not want to run the urlparse every time.

harpere: I guess it's possible that a site would have a single application that changes content, and the url, without actually reloading.  but seems super unlikely that the hostname would ever change.

msm0504: @robertrmartinez That makes sense. Was just checking.

PR #4891, opened 2020-02-25
ozone 2.3.0 adapter updates

afsheenb:
Type of change:

 [X] Bugfix
 [X] Feature

Description of changes:

• General Code clean up/re-factoring - updated some of the validation errors we display to contain more meaningful info.
• Update validation on lotameData object.
• GDPR updates - introduced a config parameter publishers can add on-page to strictly enforce GDPR.
• Added further support of ozone-specific query-strings for debugging/QA.

engineering@ozoneproject.com

### DIFF
--- a/modules/ozoneBidAdapter.js
+++ b/modules/ozoneBidAdapter.js
@@ -5,14 +5,11 @@ import {config} from '../src/config.js';
 import {getPriceBucketString} from '../src/cpmBucketManager.js';
 import { Renderer } from '../src/Renderer.js';
 
-
 const BIDDER_CODE = 'ozone';
 const ALLOWED_LOTAME_PARAMS = ['oz_lotameid', 'oz_lotamepid', 'oz_lotametpid'];
-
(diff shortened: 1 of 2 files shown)


AskRupert-DM:
We've fixed our circleCI errors so this should be good for review now @msm0504

modules/ozoneBidAdapter.js
```diff
@@ -502,15 +617,99 @@ export const spec = {
...
+    return false;
+  },
+  /**
+   * Examine the gdpr information inside the bidderRequest and return the boolean answer to the question
+   * @param bidderRequest
+   * @return {boolean}
+   */
+  failsGdprCheck(bidderRequest) {
+    // maybe the module was built without consentManagement module
+    if (!bidderRequest.hasOwnProperty('gdprConsent')) {
```
msm0504: This check is being done outside of this function on line 643. Is it needed here, too?

AskRupert-DM: you are right - its not needed - submitting an updated PR shortly - thanks for flagging/bringing to our attention @msm0504

PR #4892, opened 2020-02-25
Fix failing code-coverage command

Fawke:
## Type of change
- [X] Other (Fix for gulp code coverage command)

## Description of change
Commands`gulp test-coverage` -> `gulp view-coverage` are not generating code coverage reports after updating npm packages.
Screenshot 2020-02-25 at 6 46 07 PM

The fix is to replace module `karma-coverage-istanbul-reporter` with `karma-coverage`.


karma.conf.maker.js
```diff
@@ -156,12 +142,20 @@ module.exports = function(codeCoverage, browserstack, watchMode, file) {
...
       output: 'minimal'
     },
 
+    coverageReporter: {
+      dir: 'build/coverage',
+      reporters: [
+        { type: 'html', subdir: 'karma-html' }
+      ]
+    },
+
```
jsnellbaker: As a question, I noticed with this block of code added here (and potentially at the `reporters` variable above), we now seem to always create this `coverage` directory (such as when we run `gulp serve`).  Is this intended and needed?  Or can we alter the code some-what to only generate these folders and files when the `test-coverage` command is ran?

Fawke: @jsnellbaker Thanks for the feedback. That was a good catch, I shouldn't have gotten rid of `if(codeCoverage)` in `function setReporters`. I've added it back, and now it only generates coverage folder when we run the command, `gulp test-coverage`

jsnellbaker:
Hi @Fawke 

Thanks for resyncing the branch with master.  It seems like there is one last thing to address.

When I tried to run the `gulp coveralls` command, I encountered an error:
```
[13:45:26] Starting 'coveralls'...
cat: build/coverage/lcov.info: No such file or directory
[error] "2020-03-09T17:45:26.550Z"  'error from lcovParse: ' 'Failed to parse string'
[error] "2020-03-09T17:45:26.552Z"  'input: ' ''
[error] "2020-03-09T17:45:26.553Z"  'error from convertLcovToCoveralls'

/Users/jsnellbaker/git/Prebid.js/node_modules/coveralls/bin/coveralls.js:19
      throw err;
      ^
Failed to parse string
[13:45:26] 'coveralls' errored after 456 ms
[13:45:26] Error in plugin 'gulp-shell'
Message:
    Command `cat build/coverage/lcov.info | node_modules/coveralls/bin/coveralls.js` failed with exit code 1
Details:
    domainEmitter: [object Object]
    domain: [object Object]
    domainThrown: false
[13:45:26] 'coveralls' errored after 1.18 min
```

Judging from the `cat` command, this new setup is not generating the output in a `lcov.info` file for `coveralls` to import.

I think if we update the reporter in the `karma.conf` to use `lcov` instead of `html` and update the paths in the `gulp coveralls` and `gulp view-coverage` commands, it should work.  Can you give this a try?

Fawke:
Hi @jsnellbaker,

I did as you said, I changed the type to **lcov** from **html** and also changed the paths in **gulpfile.js**. But after doing that, I'm not getting the error you're getting, but a different error.

Screenshot 2020-03-10 at 1 42 17 PM

I read the read the `RELEASE_SHEDULE.md` file and found that you've to have a username and key to run this job. Since I don't have it, do you think that because of that I'm getting this error?

jsnellbaker:
To note - we addressed permissions issue separately.  This all appears to be working as expected now.

PR #4897, opened 2020-02-26
Add "adtype" from Yieldlab response for media type differentiation

mirkorean:
## Type of change

- [x] Bugfix

## Description of change

We added 'adtype' as a key to our response (values: 'BANNER', 'VIDEO' or 'NATIVE'), so we can better differentate media types. This fixes bids for adUnits that are from multiple mediatypes (e.g. banner, video.outstream) as our current methods assume unique mediatypes per adUnit.


GLStephen:
@mirkorean Do you have some test parameters I can use to verify bidding? I tried the ones in the test spec but they send nonbids.

mirkorean:
Hey @GLStephen! Thanks for looking into this.

You can use the test parameters that are referenced in:
https://github.com/prebid/Prebid.js/blob/master/modules/yieldlabBidAdapter.md

GLStephen:
LGTM

PR #4916, opened 2020-02-29
UserId module will provide sub-module ids in ORTB eids format

pm-harshad-mane:
## Type of change
- [ X ] Feature

## Description of change
Many bidders are converting ids provided by UserId module into ORTB eids format.
With this PR, UserId module will provide sub-module ids in ORTB eids format as well.
Current: ```  validBidRequests[].userId ```
New: ``` validBidRequests[].userId and validBidRequests[].userIdAsEids ```


pm-harshad-mane:
Hi @jsnellbaker ,
After many attempts, CI has passed :-)

jsnellbaker:
Hi @pm-harshad-mane 

I think for these new fields, they should be documented somehow on the userId module page on prebid.org.  Could you put something together?

Additionally, I understand these new fields are something that could be used by the PBS adapter.  If these values are set up in an innate format that could be directly passed to the PBS ORTB request, could you make some updates to the current logic we have in the PBS adapter to pass the whole set of data as is?  This would save us the trouble of individually populating every possible userId value in the ORTB request in the PBS adapter.

Let me know your thoughts on the above (if it would be possible or no).  Thanks.

pm-harshad-mane:
Thanks for the suggestion @jsnellbaker .
The idea is to avoid code duplication in multiple bidders like PBS, PubMatic, etc.
I wanted to wait for the review before I update the bidders.
I will share a PR for documentation on prebid.org.
Can we take bidder changes in separate PR?

pm-harshad-mane:
Thanks @jsnellbaker!

pm-harshad-mane:
Thank you @idettman!

PR #4923, opened 2020-03-02
Qc/qc usersyncs

SleimanJneidi:
## Type of change

- [X] Feature

## Description of change

Adds support for userSyncs in QC adapter

- inventoryteam@quantcast.com


SleimanJneidi:
@harpere how would I do that?

harpere:
there are 2 more arguments passed to getUserSyncs().  for example:
https://github.com/prebid/Prebid.js/blob/master/modules/rubiconBidAdapter.js#L751

SleimanJneidi:
@harpere this will be appended on the server side

SleimanJneidi:
@hadarpeer do you mind having another look please?

PR #4933, opened 2020-03-05
add mediaforce bid adapter

ksanksana:
## Type of change

- [x] New bidder adapter   

## Description of change
Add mediaforce Bid Adapter

```
{
  bidder: 'mediaforce',
  params: {
    placement_id: 13144370,
    publisher_id: 'pub111'
  }
}
```
- contact email of the adapter’s maintainer: little.grey.goblin@gmail.com
- a link to a PR on the docs repo: https://github.com/prebid/prebid.github.io/pull/1838

### DIFF
--- a/test/spec/modules/mediaforceBidAdapter_spec.js
+++ b/test/spec/modules/mediaforceBidAdapter_spec.js
@@ -1,0 +1,240 @@
+import {assert} from 'chai';
+import {spec} from 'modules/mediaforceBidAdapter.js';
+import * as utils from '../../../src/utils.js';
+
+describe('mediaforce bid adapter', function () {
+  let sandbox;
+
+  beforeEach(function () {
+    sandbox = sinon.sandbox.create();
+  });
+
+  afterEach(function () {
+    sandbox.restore();
+  });
+
+  function getLanguage() {
+    let language = navigator.language ? 'language' : 'userLanguage';
+    return navigator[language].split('-')[0];
+  }
+
+  const language = getLanguage();
+  const baseUrl = 'https://rtb.mfadsrvr.com'
+
+  describe('isBidRequestValid()', function () {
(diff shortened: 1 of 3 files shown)


ksanksana:
Hi @Fawke ,
Sorry for that! Please try again now. You may need several attempts to get a real bid.

test/spec/modules/mediaforceBidAdapter_spec.js
```diff
@@ -0,0 +1,240 @@
...
+  afterEach(function () {
+    sandbox.restore();
+  });
+
+  function getLanguage() {
+    let language = navigator.language ? 'language' : 'userLanguage';
+    return navigator[language].split('-')[0];
+  }
+
+  const language = getLanguage();
```
Fawke: I don't see this variable being used anywhere else in the code. Do we need it?

ksanksana:
Hi @Fawke, 
> I don't see this variable being used anywhere else in the code. Do we need it?

I added the use of a previously unused variable "language".

> I tried several times and I finally got a bid response. Is there any particular reason for this behaviour?

This is a production url. We can't answer "yes" to all bid requests. If it is a problem I can add some code for test purpose, but I would rather not do this.

Fawke:
> Hi @Fawke,
> 
> > I don't see this variable being used anywhere else in the code. Do we need it?
> 
> I added the use of a previously unused variable "language".
> 
> > I tried several times and I finally got a bid response. Is there any particular reason for this behaviour?
> 
> This is our production url, which has some heuristics underneath, which decides whether to yes-bid or no - depending on whether request is valuable for ur or not. We had made it yes-bid more often for the test period (so it did), but making it yes-bid all the time is an additional effort (not required for production), so we try to avoid it.

Oh..I see.
Thanks for clarifying that.

PR #4934, opened 2020-03-05
New e-volution tech bid adapter

e-volution-tech:
## Type of change
- [x] New bidder adapter   

## Description of change

```
{
  bidder: 'e_volution',
  params: {
    placementId: 0,
    traffic: 'video'
  }
}
{
  bidder: 'e_volution',
  params: {
  placementId: 0,
    traffic: 'banner',
  }
}
```

admin@e-volution.ai

https://github.com/prebid/prebid.github.io/pull/1839


e-volution-tech:
https://github.com/prebid/prebid.github.io/pull/1839 Docs are waiting to be merged

modules/evolution_techBidAdapter.js
```diff
@@ -0,0 +1,113 @@
...
+  }
+}
+
+export const spec = {
+  code: BIDDER_CODE,
+  supportedMediaTypes: [BANNER, VIDEO, NATIVE],
+  noSync: NO_SYNC,
+
+  isBidRequestValid: (bid) => {
+    return Boolean(bid.bidId && bid.params && !isNaN(bid.params.placementId));
```
msm0504: I suggest changing this to `!isNaN(parseInt(bid.params.placementId))`

Calling `parseInt` first will protect against some odd behaviors with `isNaN`. For example, `isNaN(null)` and `isNaN('')` both return false.

modules/evolution_techBidAdapter.js
```diff
@@ -0,0 +1,113 @@
...
+      winTop = window.top;
+    } catch (e) {
+      location = winTop.location;
+      utils.logMessage(e);
+    };
+    let placements = [];
+    let request = {
+      'deviceWidth': winTop.screen.width,
+      'deviceHeight': winTop.screen.height,
+      'language': (navigator && navigator.language) ? navigator.language : '',
```
msm0504: Replace `navigator.language` with `navigator.language.split('-')[0]`. Line 56 can then be removed.

If there's no '-', then `split('-')[0]` will be the whole string.

modules/evolution_techBidAdapter.js
```diff
@@ -0,0 +1,113 @@
...
+    return {
+      method: 'POST',
+      url: AD_URL,
+      data: request
+    };
+  },
+
+  interpretResponse: (serverResponse) => {
+    let response = [];
+    serverResponse = serverResponse.body;
```
msm0504: This assignment is not needed. You can loop over `serverResponse.body` without doing this step.

modules/evolution_techBidAdapter.js
```diff
@@ -0,0 +1,113 @@
...
+      let resItem = serverResponse[i];
+      if (isBidResponseValid(resItem)) {
+        response.push(resItem);
+      }
+    }
+    return response;
+  },
+
+  getUserSyncs: (syncOptions, serverResponses) => {
+    if (NO_SYNC) {
```
msm0504: In the current code, `NO_SYNC` will always be `true`. Is this the intended behavior?

modules/evolution_techBidAdapter.md
```diff
@@ -0,0 +1,53 @@
...
+
+# Description
+
+Module that connects to e-volution-tech demand sources
+
+# Test Parameters
+```
+    var adUnits = [
+                // Will return static test banner
+                {
```
msm0504: The response from your server when given these banner params does not have an `ad` property, so the `interpretResponse` function determines that it is invalid and no ad is shown.
```
[{"width":300,"height":250,"mediaType":"banner","requestId":"20437cf5b57b36","cpm":0.3,"ttl":120,"creativeId":"123","netRevenue":true,"currency":"USD","dealId":"HASH"}]
```

Also, all responses from your server need to be gzipped.

modules/evolution_techBidAdapter.js
e-volution-tech: It is for easy turn off/on user sync feature.While endpoint for user sync not ready - it will be turned off

modules/evolution_techBidAdapter.js
```diff
@@ -0,0 +1,113 @@
...
+const AD_URL = 'https://service.e-volution.ai/?c=o&m=multi';
+const URL_SYNC = 'https://service.e-volution.ai/?c=o&m=sync';
+const NO_SYNC = true;
+
+function isBidResponseValid(bid) {
+  if (!bid.requestId || !bid.cpm || !bid.creativeId ||
+    !bid.ttl || !bid.currency) {
+    return false;
+  }
+  switch (bid['mediaType']) {
```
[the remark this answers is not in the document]
e-volution-tech: Make sense. This comes from old version. I will change it

modules/evolution_techBidAdapter.js
e-volution-tech: +

modules/evolution_techBidAdapter.js
e-volution-tech: +

modules/evolution_techBidAdapter.js
e-volution-tech: +

modules/evolution_techBidAdapter.md
e-volution-tech: + adpropertie and gzip response

e-volution-tech:
@msm0504  Thanks for your comments, i've made all changes according to them

modules/evolution_techBidAdapter.js
msm0504: Sorry, I didn't mean that you should remove the ternary operator. If `navigator.language` is null, calling `split` will throw an error. Replace with this:
```
 (navigator && navigator.language) ? navigator.language.split('-')[0] : ''
```

e-volution-tech: My fault, had to think about it. I've made this change

msm0504: The condition you were using originally was safer. This latest one assumes that `navigator` will always have a `language` property. But should be fine either way

e-volution-tech: +

PR #4943, opened 2020-03-06
Vdoai spec import js

pm-harshad-mane:
## Type of change
- [X] Bugfix

## Description of change
VdoAI bidder spec was missing .js extension in import statements, was throwing errors in ```gulp test```


pm-harshad-mane:
Thanks @harpere !

PR #4953, opened 2020-03-09
update logic in adpod module for playersize

jsnellbaker:
## Type of change
- [x] Bugfix

## Description of change
This change allows the `playerSize` field in an adpod config to accept an array of arrays in addition to an array of numbers (ie `[300, 250]` or `[[300, 250]]`).

Previously the module only accepted the latter as a valid config.  However, due to logic [here](https://github.com/prebid/Prebid.js/blob/master/src/prebid.js#L72), the `playerSize` field in any video adUnits was modified to be an array of arrays (a workaround so it works with logic later in the auction work-flow).

This modification causes an issue when the adUnit is refreshed, as the updated adUnit gets rejected by the adpod logic even though it should be fine.  This fix addresses that refresh scenario by allowing both formats instead of just the one.


jsnellbaker:
I have resynced with master and resolved the conflict.  Please take another look when you have the chance.

PR #4959, opened 2020-03-10
add import extensions

robertrmartinez:
Spec does not have extensions on import


test/spec/modules/7xbidBidAdapter_spec.js
```diff
@@ -1,7 +1,6 @@
 import {expect} from 'chai';
-import {spec, _getUrlVars} from 'modules/7xbidBidAdapter';
-import * as utils from 'src/utils';
-import {config} from 'src/config';
```
robertrmartinez: unused import deleted

PR #4961, opened 2020-03-10
Update core-js to make it easy to remove.

nanek:
## Type of change
- Build related changes

## Description of change
This PR is not complete. It is intended for discussion only.

Currently prebid includes the following core-js polyfills. This adds 40kb to the size of prebid.js.
````
  "core-js/library/fn/array/find", // no ie11
  "core-js/library/fn/array/includes", // no ie11
  "core-js/library/fn/set", // ie11 supports Set but not Set#values
  "core-js/library/fn/string/includes", // no ie11
  "core-js/library/fn/number/is-integer", // no ie11,
  "core-js/library/fn/array/from" // no ie11
````

These features are supported in modern browsers. IE 11 is the main outlier that requires the polyfills. IE 11 traffic is low but it exists. I'd like to make it possible to make a build of prebid without these polyfills.

Our site and others already implement polyfills for IE 11 in our application build process.

One option, similar to this PR, is to move the core-js references to a single file that can be easily included/excluded as needed if using as an npm dependency. The default non-npm build could still prepend these polyfills to prebid.js, I was not familiar enough with the webpack/gulp build process to configure this. The only side-effect would be it would pollute the global scope.

Is this something that could be considered? I'm happy to assist if there is consensus on the goal/approach.

### DIFF
--- a/allowedModules.js
+++ b/allowedModules.js
@@ -5,7 +5,6 @@ const sharedWhiteList = [
   "core-js/library/fn/set", // ie11 supports Set but not Set#values
   "core-js/library/fn/string/includes", // no ie11
   "core-js/library/fn/number/is-integer", // no ie11,
-  "core-js/library/fn/array/from" // no ie11
 ];
 
 module.exports = {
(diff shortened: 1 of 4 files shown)


mkendall07:
I agree with the goal, not sure I like the approach. We have a strict no polluting the global namespace policy, so I don't see why we'd start now. if there's a way to make this work without doing so I'd be happy to consider that approach.

nanek:
Thanks for the feedback. I believe https://github.com/zloirock/core-js#babelruntime makes this possible, however requires corejs v3. I'll take another look.

PR #4962, opened 2020-03-10
assert string returned not that we break things

robertrmartinez:
Fixing a test which was accidentally merged to master.


test/spec/modules/rubiconAnalyticsAdapter_spec.js
```diff
@@ -854,7 +854,7 @@ describe('rubicon analytics adapter', function () {
     inputUrl = 'https://prebid.org/some/path?pbjs_debug=true';
     expect(getHostNameFromReferer(inputUrl)).to.equal('prebid.org');
     inputUrl = 'http://xn--p8j9a0d9c9a.xn--q9jyb4c/';
-    expect(getHostNameFromReferer(inputUrl)).to.equal('xn--p8j9a0d9c9a.xn--q9jyb4c');
+    expect(typeof getHostNameFromReferer(inputUrl)).to.equal('string');
```
robertrmartinez: IE / Edge translates this to actual Chinese characters. Which is fine if sent to Analytics. We just want to test that it does not BREAK the code

PR #4997, opened 2020-03-18
Districtm Dmx adding Coppa support

stevealliance:
## Type of change

- [ ] Bugfix
- [ x] Feature
- [ ] New bidder adapter   
- [ ] Build related changes
- [ ] CI related changes
- [ ] Other

## Description of change
Adding support for COPPA in Districtm DMX bidder code.


modules/districtmDMXBidAdapter.js
```diff
@@ -103,6 +103,10 @@ export const spec = {
       dmxRequest.source.ext.schain = schain || {}
     } catch (e) {}
     let tosendtags = bidRequest.map(dmx => {
+      if (dmx.params.coppa) {
```
mike-chowla: Instead of using your config parameter, please use the global one.  See PR #5012 for an example.

stevealliance:
@mike-chowla thanks for that info update to reflect the comment above!

modules/districtmDMXBidAdapter.js
stevealliance: thanks for the example

modules/districtmDmxBidAdapter.md
```diff
@@ -23,8 +25,10 @@ The `districtmDmxAdapter` module allows publishers to include DMX Exchange deman
 
 | Key | Scope | Type | Description
 | --- | --- | --- | ---
-| dmxid | Mandatory | Integer | Unique identifier of the placement, dmxid can be obtained in the district m Boost platform.
-| memberid | Mandatory | Integer | Unique identifier for your account, memberid can be obtained in the district m Boost platform.
+| `dmxid` | Mandatory | Integer | Unique identifier of the placement, dmxid can be obtained in the district m Boost platform.
+| `memberid` | Mandatory | Integer | Unique identifier for your account, memberid can be obtained in the district m Boost platform.
+| `coppa` | optional | Boolean | Enables Child's Online Privacy Protection Act (COPPA) regulations. | true
```
mike-chowla: This is needs to be removed since COPPA is set via pbjs.setConfig()

modules/districtmDmxBidAdapter.md
```diff
@@ -47,6 +51,27 @@ The `districtmDmxAdapter` module allows publishers to include DMX Exchange deman
...
+            banner: {
+                sizes: [[300, 250], [300,600]],
+            }
+        },
+        bids: [{
+            bidder: 'districtmDMX',
+            params: {
+                dmxid: 100001,
+                memberid:  100003,
+                coppa: 1
```
mike-chowla: This needs to be removed as well.

stevealliance: this was removed

PR #5004, opened 2020-03-19
VIS.X: use schain & userId modules

mk0x9:
## Type of change

- [x] Feature

## Description of change

Added support for the schain and userId modules.

Link to the updated docs PR: https://github.com/prebid/prebid.github.io/pull/1871


GLStephen:
LGTM

PR #5007, opened 2020-03-19
feat(module): multi size request

uditalias:
## Type of change

- [x] Feature

## Description of change

Multi-size on a single request


uditalias:
Hey @Fawke,

- We added more tests and now we have 95% coverage - thank you.
- We fixed our test params
- The package-lock.json was a mistake, we revert it to master

Thanks

uditalias:
Hi @Fawke 
We had an issue with this test params, thank you for your catch :)
You can check it now

uditalias:
@Fawke 

We only compress our response with GZIP when the content-length is above 1kb and our demo response length is 509 bytes.

As you know, GZIP is not effective when the content length is small and may increase the size of the response.

Thank you

PR #5008, opened 2020-03-19
add Adtelligent alias in legacy

GeneGenie:
## Type of change
- [x] New alias in legacy

-  g.moroz@adtelligent.com


mike-chowla:
We are no longer doing releases on the 2.44.x legacy branch

Issue #4924, opened 2020-03-03
ALL BID ADAPTERS READ THIS: need a map of bidderCodes to vendor ID


bretg:
As part of #4867 , we need to establish a mapping of bidder code to TCF Global Vendor List ID (GVL ID).

e.g.

appnexus: 32
rubicon: 52
pubmatic: 76
openx: 69
ix: 10

We're asking for everyone's help to self-identify your bidder-code-to-GVL-ID mapping.

Please add a comment to this issue

leonardlabat:
Hello @bretg 

criteo: 91

anand-venkatraman:
pulsepoint: 81

smenzer:
we're a userId module, but here it is for record keeping anyway

id5id: 131

pm-harshad-mane:
Confirming PubMatic: 76.

epouradier:
Verizon Media (`aol`, `onevideo`, `onedisplay`, `onemobile`): 25

osazos:
adagio: 617

ptomasroos:
@bretg How will this work in the case of running an alias for a bid adapter?

aleksatr:
aardvark: 169

onaydenov:
fidelity: 408

SublimeLeo:
sublime: 114

jessberman:
SpotX: 165

GeneGenie:
Adtelligent: 410

derdeka:
@bretg Do you plan to extend the bidder interface to provide such metadata directly from bidders?

bretg:
Yes - it's going to be come part of the spec. See https://github.com/prebid/Prebid.js/pull/5018

vedantseta:
Media.net : 142

pycnvr:
Conversant: 24

dpapworth:
Quantcast: 11

jaiminpanchal27:
Updated all bidders listed in this issue with their GVL ID in https://github.com/prebid/Prebid.js/pull/5018

bretg:
Closing this -- any further updates should be done as part of individual pull requests.

PR #5018, opened 2020-03-23
TCF Purpose 1 enforcement

jaiminpanchal27:
## Type of change
- [x] Feature

## Description of change
This PR introduces feature to enforce TCF Purpose 1. It is an optional module and publisher can configure using existing setConfig api. 

API call would look like below.  With this PR we are only targeting purpose 1. We are using string based naming for each purpose. Purpose 1 is storage.

Purpose 1 scope includes device access, user sync pixels and user id modules. Publisher can restrict these using storage purpose using below api

```
pbjs.setConfig({
  consentManagement: {
    gdpr: {
      cmpApi: 'iab',
      defaultGdprScope: true/false (0/1?)
      timeout: 3000,
      // allowAuctionWithoutConsent: false,  // deprecated
      rules: [{
        // don’t enforce Purpose 1 except for bidderA.
        // this means that TCF approval is only considered
        // for this one bidder.
        purpose: "storage",
        enforcePurpose: false,
        enforceVendor: false,
        vendorExceptions: ["bidderA"]
      },{
        // Allow the user to object to vendors for Purpose 2
        // but don’t allow them to turn off header bidding
        // altogether.
        // The exception is bidderB, which the user will
        // not be allowed to reject
        purpose: "basicAds",
        enforcePurpose: false,
        enforceVendor: true,
        vendorExceptions: ["bidderB"]
      },{
        // allow the user to have their userId removed from
        // header bidding except for bidders A and B
        purpose: "personalizedAds",
        enforcePurpose: true,
        enforceVendor: true
        vendorExceptions: ["bidderA","bidderB"]
      },{
        // analytics adapters are ok
       
        purpose: "measurement",
        enforcePurpose: false,
        enforcePurpose: false
      }]
    }
  }
});

```

We asked bidders to provide their gvl id here https://github.com/prebid/Prebid.js/issues/4924. This PR updated the gvl id for Appnexus, Rubicon bid adapter and Id5IdSystem to show the usage of gvlid. 

Also, I have moved storage related functions to new core module storageManager. Going forward everyone needs to use this module to use storage functions. 

### New storage manager api 

We have introduced new GDPR enforcement module with this PR, each call related to storage has to go through certain validation checks to ensure that device access is allowed or not. 

Storage manager is a core module which exposes two functions,

```JavaScript
/**
 * This function returns storage functions to access cookies and localstorage. This function will bypass the gdpr enforcement requirement. Prebid as a software needs to use storage in some scenarios and is not a vendor so GDPR enforcement rules does not apply on Prebid.
 * @param {string} moduleName Module name
 */
- functon getCoreStorageManager(moduleName) 

/**
 * Note: Core modules or Prebid modules like Currency, SizeMapping should use getCoreStorageManager
 * This function returns storage functions to access cookies and localstorage. Bidders and User id modules should import this and use it in their module if needed. GVL ID and Module name are optional param but gvl id is needed for when gdpr enforcement module is used.
 * @param {Number=} gvlid Vendor id
 * @param {string=} moduleName BidderCode or module name
 */
- function getStorageManager(gvlid, moduleName)
```

Both api returns an object with following function with same definition with one exception of extra optional argument at the end i.e done callback.

-   setCookie,
-   getCookie,
-   localStorageIsEnabled,
-   cookiesAreEnabled,
-   setDataInLocalStorage,
-   getDataFromLocalStorage,
-   removeDataFromLocalStorage,
-   hasLocalStorage

Some bidders and user id modules are calling storage related functions when adapter are getting parsed by browser. At this moment Prebid has no knowledge of gdpr enforcement configured by Publisher using setConfig. Hence each function above comes with optional done callback which will be called when operatio is complete. See below example.

```JavaScript
// Callback passed here will be called after gdpr enforcement check are completed.
storage.setDataInLocalStorage('key', 'value', function(result) {
  // other code here
})
```


modules/userId/index.js
```diff
@@ -150,15 +152,15 @@ function setStoredValue(storage, value) {
...
+      coreStorage.setCookie(storage.name, valueStr, expiresStr, 'Lax');
       if (typeof storage.refreshInSeconds === 'number') {
-        utils.setCookie(`${storage.name}_last`, new Date().toUTCString(), expiresStr);
+        coreStorage.setCookie(`${storage.name}_last`, new Date().toUTCString(), expiresStr);
       }
     } else if (storage.type === LOCAL_STORAGE) {
-      utils.setDataInLocalStorage(`${storage.name}_exp`, expiresStr);
-      utils.setDataInLocalStorage(storage.name, encodeURIComponent(valueStr));
+      localStorage.setItem(`${storage.name}_exp`, expiresStr);
+      localStorage.setItem(storage.name, encodeURIComponent(valueStr));
```
jsnellbaker: are these `localStorage` references correct?   Shouldn't they go to the `coreStorage` functions like the updates for the cookies above?

bretg: +1

modules/gdprEnforcement.js
```diff
@@ -0,0 +1,181 @@
...
+
+let addedDeviceAccessHook = false;
+let enforcementRules;
+
+function getGvlid() {
+  let gvlid;
+  const bidderCode = config.getCurrentBidder();
+  if (bidderCode) {
+    const bidder = adapterManager.getBidAdapter(bidderCode);
+    gvlid = bidder.getSpec().gvlid;
```
bretg: What will happen here if gvlid is unspecified? Should it be undefined or 0?

modules/gdprEnforcement.js
```diff
@@ -0,0 +1,181 @@
...
+      isAllowed = true;
+    }
+  } else if (rule.enforcePurpose === false && rule.enforceVendor === false) {
+    if ((includes(rule.vendorExceptions, currentModule) &&
+        (utils.deepAccess(consentData, 'vendorData.purpose.consents.1') === true && utils.deepAccess(consentData, `vendorData.vendor.consents${gvlid}`) === true)) ||
+        !includes(rule.vendorExceptions, currentModule)) {
+      isAllowed = true;
+    }
+  }
+  return isAllowed;
```
bretg: Missing the case of (rule.enforcePurpose === true && rule.enforceVendor === false)

modules/gdprEnforcement.js
```diff
@@ -0,0 +1,181 @@
...
+/**
+ * This function takes in rules and consentData as input and validates against the consentData provided. If it returns true Prebid will allow the next call else it will log a warning
+ * @param {Object} rules enforcement rules set in config
+ * @param {Object} consentData gdpr consent data
+ * @returns {boolean}
+ */
+function validateRules(rule, consentData, currentModule, gvlid) {
+  let isAllowed = false;
+  if (rule.enforcePurpose && rule.enforceVendor) {
+    if (includes(rule.vendorExceptions, currentModule) ||
```
bretg: The intention wasn't that vendorExceptions would bypass enforcePurpose, just enforceVendor. In otherwords, I think this should be:

(utils.deepAccess(consentData, 'vendorData.purpose.consents.1') === true && (
      (includes(rule.vendorExceptions, currentModule) || utils.deepAccess(consentData, `vendorData.vendor.consents.${gvlid}`) === true))

modules/gdprEnforcement.js
```diff
@@ -0,0 +1,181 @@
...
+      isAllowed = true;
+    }
+  } else if (rule.enforcePurpose === false && rule.enforceVendor === true) {
+    if (includes(rule.vendorExceptions, currentModule) ||
+        (utils.deepAccess(consentData, `vendorData.vendor.consents.${gvlid}`) === true)) {
+      isAllowed = true;
+    }
+  } else if (rule.enforcePurpose === false && rule.enforceVendor === false) {
+    if ((includes(rule.vendorExceptions, currentModule) &&
+        (utils.deepAccess(consentData, 'vendorData.purpose.consents.1') === true && utils.deepAccess(consentData, `vendorData.vendor.consents${gvlid}`) === true)) ||
```
bretg: Similarly, vendorExceptions wasn't intended affect the high level enforcement of the purpose. So I think this should be interpreted as "only consider the user's vendor choices for the exceptions".

(includes(rule.vendorExceptions, currentModule) && utils.deepAccess(consentData, `vendorData.vendor.consents${gvlid}`) === true

modules/gdprEnforcement.js
```diff
@@ -0,0 +1,181 @@
...
+      if (isAllowed) {
+        result.valid = true;
+        return fn.call(this, result);
+      } else {
+        utils.logWarn(`User denied Permission for Device access for ${curModule}`);
+        result.valid = false;
+        return fn.call(this, result);
+      }
+    } else {
+      utils.logInfo('GDPR enforcement only applies to CMP version 2');
```
bretg: In trying to be sensitive to overall package size, I'd like to encourage concise warning messages. Could we cut some of these down? e.g. "Enforcing TCF2 only" is less than half the length of this message.

modules/gdprEnforcement.js
harpere: looks like it would be `undefined` just like if there was no `bidderCode`

modules/gdprEnforcement.js
jaiminpanchal27: Added

modules/gdprEnforcement.js
jaiminpanchal27: Updated the message

modules/gdprEnforcement.js
jaiminpanchal27: If no gvlid and gdpr rules are enforced then bidder or user id module will not have access.

jaiminpanchal27:
@bretg @jsnellbaker Updated the code and updated all the bidders as well. I also updated the description regarding new storage manager core module.

modules/gdprEnforcement.js
```diff
@@ -0,0 +1,222 @@
...
+      (
+        utils.deepAccess(consentData, `vendorData.vendor.consents.${gvlid}`) === true
+      )
+    ) {
+      isAllowed = true;
+    }
+  } else if (rule.enforcePurpose === false && rule.enforceVendor === false) {
+    if (
+      (
+        includes(rule.vendorExceptions, currentModule) &&
```
harpere: this is redundant, so can be left out.  It's covered by the `|| !includes(rule.vendorExceptions, currentModule)` part of the conditional.

modules/gdprEnforcement.js
```diff
@@ -0,0 +1,222 @@
...
+      (
+        !includes(rule.vendorExceptions, currentModule) &&
+        (utils.deepAccess(consentData, 'vendorData.purpose.consents.1') === true)
+      ) ||
+      (
+        includes(rule.vendorExceptions, currentModule) &&
+        (utils.deepAccess(consentData, 'vendorData.purpose.consents.1') === true) &&
+        (utils.deepAccess(consentData, `vendorData.vendor.consents.${gvlid}`) === true)
+      )
+    ) {
```
harpere: could be simplified to just:
```
    if (
      utils.deepAccess(consentData, 'vendorData.purpose.consents.1') === true) &&
      (
        !includes(rule.vendorExceptions, currentModule) ||
        (utils.deepAccess(consentData, `vendorData.vendor.consents.${gvlid}`) === true)
      )
    ) {
```

bretg:
docs PR https://github.com/prebid/prebid.github.io/pull/1892

modules/gdprEnforcement.js
jaiminpanchal27: If I remove this then this function will start checking for consents for all bidders which we do not want. We only want to check for consents if bidder is part of vendorExceptions.

harpere: still don't need both of those.  could reorder to:
```
     !includes(rule.vendorExceptions, currentModule) ||
     (
        (utils.deepAccess(consentData, 'vendorData.purpose.consents.1') === true) &&
        (utils.deepAccess(consentData, `vendorData.vendor.consents.${gvlid}`) === true)
     )
```
would only check for consents if the bidder is part of vendorExceptions.

modules/gdprEnforcement.js
```diff
@@ -0,0 +1,222 @@
...
+      isAllowed = true;
+    }
+  } else if (rule.enforcePurpose === true && rule.enforceVendor === false) {
+    if (
+      (utils.deepAccess(consentData, 'vendorData.purpose.consents.1') === true) &&
+      (
+        !includes(rule.vendorExceptions, currentModule) ||
+        (
+          includes(rule.vendorExceptions, currentModule) &&
```
harpere: line 77 will always be true because of line 75, so we don't need to include line 77.
`!A || (A & B) === !A || B`

jaiminpanchal27: oops you are right.

PR #5019, opened 2020-03-23
Between: multisizes support added

ignat-one:
#4369  Type of change
- [x] Feature

## Description of change
Added support for size multiplexing, updated description in betweenBidAdapter.md

khaylov@betweenx.com


ignat-one:
Hello, @Fawke. Thanks for response.
1. Yes, you right, my mistake. I have rollbacked  package-lock.json. I have got it from master.
2. Yes, that's okay, because our SSP server checks the domain from which the request is sent. The domain is depends on the "s" parameter. But I have checked again, it's all right.

PR #5020, opened 2020-03-23
New bidlab prebid adapter

bidlab:
## Type of change

- [x] New bidder adapter   

## Description of change
New adapter

```
{
                            bidder: 'bidlab',
                            params: {
                                placementId: 0,
                                traffic: 'video'
                            }
}
```
```
{
                            bidder: 'bidlab',
                            params: {
                                placementId: 0,
                                traffic: 'banner',
                            }
}
```

hello@bidlab.ai

https://github.com/prebid/prebid.github.io/pull/1877

### DIFF
--- a/modules/bidlabBidAdapter.js
+++ b/modules/bidlabBidAdapter.js
@@ -1,0 +1,113 @@
+import {registerBidder} from '../src/adapters/bidderFactory.js';
+import { BANNER, NATIVE, VIDEO } from '../src/mediaTypes.js';
+import * as utils from '../src/utils.js';
+
+const BIDDER_CODE = 'bidlab';
+const AD_URL = 'https://service.bidlab.ai/?c=o&m=multi';
+const URL_SYNC = 'https://service.bidlab.ai/?c=o&m=sync';
+const NO_SYNC = true;
+
+function isBidResponseValid(bid) {
+  if (!bid.requestId || !bid.cpm || !bid.creativeId ||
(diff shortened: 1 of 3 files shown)


modules/bidlabBidAdapter.js
```diff
@@ -0,0 +1,112 @@
...
+        response.push(resItem);
+      }
+    }
+    return response;
+  },
+
+  getUserSyncs: (syncOptions, serverResponses) => {
+    if (NO_SYNC) {
+      return false
+    } else {
```
mike-chowla, line 102: Since NO_SYNC is defined as const so is always TRUE, not sure why this code is here.

modules/bidlabBidAdapter.md
```diff
@@ -0,0 +1,53 @@
...
+                            playerSize: [640, 480],
+                            context: 'instream'
+                        }
+                    },
+                    bids: [
+                        {
+                            bidder: 'bidlab',
+                            params: {
+                                placementId: 0,
+                                traffic: 'video'
```
mike-chowla, line 47: Your test video response resolves to a flash creative.  It would be better to use an mp4 one since flash is in the process of being dropped by browsers

modules/bidlabBidAdapter.js
e-volution-tech: @mike-chowla this was made to simplify process of turning on user sync , when it will be ready

PR #5021, opened 2020-03-24
Add type information to DSPX bidResponse

muuki88:
## Type of change

- [x] Feature

## Description of change

Add the ad `type` information from the server response to the `bidResponse` object.

Depending on the ad type other ad placements can be blocked or not.
For example a Skin Ad may require that all skys are blocked so they
don't overlap. For a regular display banner this isn't necessary.

## Other information

Initial contributors @avj83 and @onlsol

### DIFF
--- a/modules/dspxBidAdapter.js
+++ b/modules/dspxBidAdapter.js
@@ -74,6 +74,7 @@ export const spec = {
         dealId: dealId,
         currency: currency,
         netRevenue: netRevenue,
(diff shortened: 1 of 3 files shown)


bretg:
Please add a unit test to make sure this value is set as expected

muuki88:
the npm package-lock file seems to  be out of date? at least I commited it by accident and wrestled git to not commit it.

PR #5022, opened 2020-03-24
Add support for userid modules in Yieldlab adapter

mirkorean:
## Type of change

- [x] Feature

## Description of change

Adding support for user modules by getting all eids that are present, passing them to our header bidding endpoint, so we can forward them along to DSPs in our ORTB request.

### DIFF
--- a/modules/yieldlabBidAdapter.js
+++ b/modules/yieldlabBidAdapter.js
@@ -172,8 +172,8 @@ function parseSize (size) {
  */
 function createUserIdString (eids) {
   let str = []
-  for (var eid of eids) {
-    str.push(eid.source + ':' + eid.uids[0].id)
+  for (let i = 0; i < eids.length; i++) {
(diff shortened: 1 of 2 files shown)


modules/yieldlabBidAdapter.js
```diff
@@ -39,6 +39,9 @@ export const spec = {
       if (bid.params.targeting) {
         query.t = createQueryString(bid.params.targeting)
       }
+      if (bid.userIdAsEids) {
```
robertrmartinez: Because your function `createUserIdString` loops over the length of the array I would recommend having this check be `Array.isArray(bid.userIdAsEids` Just in case something weird happens where `userIdAsEids` is on the bidRequest object but is not an array (undefined, object, int, etc..)

mirkorean: Hey @robertrmartinez! Thanks for taking a look at this. I added your proposed changes. Makes sense!

PR #5045, opened 2020-03-30
Native support for AdKernel adapter

ckbo3hrk:
## Type of change
- [x] Other

## Description of change

Added support for native media type

- prebid-dev@adkernel.com

### DIFF
--- a/modules/adkernelBidAdapter.js
+++ b/modules/adkernelBidAdapter.js
@@ -1,5 +1,5 @@
 import * as utils from '../src/utils.js';
-import { BANNER, VIDEO } from '../src/mediaTypes.js';
+import { BANNER, NATIVE, VIDEO } from '../src/mediaTypes.js';
 import {registerBidder} from '../src/adapters/bidderFactory.js';
 import find from 'core-js/library/fn/array/find.js';
 import includes from 'core-js/library/fn/array/includes.js';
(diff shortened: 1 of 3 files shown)


ckbo3hrk:
Pull request for adapter doc update
https://github.com/prebid/prebid.github.io/pull/1893

jsnellbaker:
@ckbo3hrk Do you have any test params for this new native support?

ckbo3hrk:
> @ckbo3hrk Do you have any test params for this new native support?

@jsnellbaker Yes, I've updated adapter documentation. At the moment there is a server-side problem with HTTPS configuration so it's testable only via HTTP.

ckbo3hrk:
> Could you take a look to see why the bid may not be returning?

Hi @jsnellbaker 
It looks like a test campaign targeting issue.  I've updated targeting. Hope this will help.

ckbo3hrk:
> @ckbo3hrk I tried to use the test params again, but I'm still not seeing the bid return. Can you check again please? If there's any information you need, please let me know.
> 
> Thanks.

Hi @jsnellbaker
I guess it's an IP issue. Could you edit the adapter code and change `ip: 'caller'` to `ip: '8.11.10.11'`.
If no bid response - try 3-5 times as bidding is affected by a small probabilistic factor.
In case it won't help - I'll find these requests in raw request logs and figure out the exact reason.

PR #5047, opened 2020-03-30
Eids uid string only

pm-harshad-mane:
## Type of change
- [X] Bugfix

## Description of change
In UserId module, while generating userIdAsEids, code will check if the provided id by a user-id-sub-module is string or not, an only string will be allowed.

Conversant adapter spec is updated to have uid value as string in test-cases.


bretg:
Adding @pycnvr to confirm the Conversant change here.

pycnvr:
Thanks.  Looks alright.

pm-harshad-mane:
Hello @jsnellbaker ,
Issac is not available I guess (from Slack), could you please review this small fix?

pm-harshad-mane:
Thank you @jsnellbaker !!

PR #5050, opened 2020-03-30
minor consentManagement fix

harpere:
## Type of change

- [x ] Bugfix
- [ ] Feature
- [ ] New bidder adapter   
- [ ] Build related changes
- [ ] CI related changes
- [ ] Other

## Description of change

in case `tcfData.tcString` is not defined.

### DIFF
--- a/modules/consentManagement.js
+++ b/modules/consentManagement.js
@@ -99,7 +99,7 @@ function lookupIabConsent(cmpSuccess, cmpError, hookConfig) {
     if (success) {
       if (tcfData.eventStatus === 'tcloaded' || tcfData.eventStatus === 'useractioncomplete') {
         cmpSuccess(tcfData, hookConfig);
-      } else if (tcfData.eventStatus === 'cmpuishown' && tcfData.tcString.length > 0 && tcfData.purposeOneTreatment === true) {
+      } else if (tcfData.eventStatus === 'cmpuishown' && tcfData.tcString && tcfData.purposeOneTreatment === true) {
         cmpSuccess(tcfData, hookConfig);
       }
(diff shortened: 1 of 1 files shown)


modules/consentManagement.js
```diff
@@ -99,7 +99,7 @@ function lookupIabConsent(cmpSuccess, cmpError, hookConfig) {
     if (success) {
       if (tcfData.eventStatus === 'tcloaded' || tcfData.eventStatus === 'useractioncomplete') {
         cmpSuccess(tcfData, hookConfig);
-      } else if (tcfData.eventStatus === 'cmpuishown' && tcfData.tcString.length > 0 && tcfData.purposeOneTreatment === true) {
+      } else if (tcfData.eventStatus === 'cmpuishown' && tcfData.tcString && tcfData.purposeOneTreatment === true) {
```
harpere: in case `tcfData.tcString` is not defined

jsnellbaker: I'm not sure about this change.

The original spec asked for this use-case that the `tcString` had to exist.  Further if there was a scenario where the tcString was undefined and we tried to process it, the module should reject anyways (see [here](https://github.com/prebid/Prebid.js/blob/master/modules/consentManagement.js#L308)).

harpere: @jsnellbaker it may be true that the spec says it should always be a string, but that doesn't mean we shouldn't defend against bad data.  @robertrmartinez saw a console error during testing, so it must have reached this code.  But ultimately there's no negative impact because the function just exits anyway.   So not a big deal now.  But if the code changes at some point, it might be a problem.

PR #5056, opened 2020-03-31
Rakuten bid adapter

snapwich:
## Type of change
- [x] New bidder adapter

## Description of change
New bidder adapter for Rakuten.

Also adds typescript types for bidder interface.

## Other information
Bidder params update: https://github.com/prebid/prebid.github.io/pull/1899


mkendall07:
Hey @snapwich. 

Can you explain how the typescript integration works? seems like you compiled it outside of prebid and then the prebid build will just use the JS version? Interesting pattern, not sure if that's a good idea or not....

modules/rakuten/index.ts
```diff
@@ -0,0 +1,112 @@
...
+
+        if (sb.cpm && sb.ad) {
+            bidResponses.push({
+                requestId: sb.bid_id,
+                cpm: sb.cpm,
+                width: sb.width || 0,
+                height: sb.height || 0,
+                creativeId: sb.creative_id || 0,
+                dealId: sb.deal_id || '',
+                currency: sb.currency || 'JPY',
```
zachbadgett, line 80: default to 'USD'

snapwich:
Added a Typescript notice to the documentation.

@mkendall07 yes that's what I did. This allows us to support types for the bid adapter interface without having to add typescript to the build process. It's possible we might want to add typescript to the build process, but that would be a backwards breaking change if we did.

jaiminpanchal27:
@snapwich Some questions
1) Rakuten adapter seems very simple. Will compiled code be easy to review for complex adapters ?
2) Which files to review for reviewers ? Both typescript and compiled file ?
3) Some adapters use utils, storageManager. Will there be a ts file for them too ?

snapwich:
1. it should be, the compiled javascript is almost the same as the typescript with the types removed since the compile target is a modern verison of js. (since we'll be running through babel still compiling to modern js is no problem).
2. the compiled .js is probably the only one that matters as it's the code we'll actually be using.
3. there could be. type adoption can be incremental so that any imported modules that don't have type definitions can be treated as `any` type (so basically normal javascript). but I think it'd beneficial to type as much as the bidder interface as possible.

some other typescript options are:
1) we just keep `.d.ts` types in the repo but don't allow submitters to submit their `.ts` code (they have to manage it themselves somewhere else).
2) we keep types as part of a separate repo and don't allow typescript at all in this repo. I have a feeling they could get out-of-date easier but maybe not. (this is how definitelytyped works)
3) we update the webpack build to compile the `.ts` files so we don't have generated code in this repo. i'd consider this a backwards incompatible change (so probably a major revision bump) as it'd require people using prebid.js through npm to also update their webpack builds as well.

I described it in a little more detail here: https://github.com/prebid/Prebid.js/issues/5097

PR #5060, opened 2020-04-01
e-volution-tech: change file names

e-volution-tech:
## Type of change

- [x] Bugfix

## Description of change
Resolved issue with adapter filename


bretg:
@e-volution-tech - help us confirm what you want to do here. The contents of these 3 files look to be exactly the same as the original files in https://github.com/prebid/Prebid.js/pull/4934

- the original filename was evolution_techBidAdapter.js
- the new filename is e_volutionBidAdapter.js
- the biddercode in the adapter is e_volution
- the docs repo also shows the bidder code as e_volution

It makes sense that you'd prefer the filename to include the biddercode, so the question is: do you need the original evolution_tech name still or can we just delete that? The PBJS download page won't build with the mismatch, but someone could have built it manually.

e-volution-tech:
@bretg Old file can be deleted, this adapter hasn't been built by any one, and changes are made to fix problem with adapter download from Prebid download page

Issue #5044, opened 2020-03-30
SizeMappingV2 bug with multiple ad units having the same code


benjaminclot:
## Type of issue
When using SizeMapping, a collateral (and very handy) feature is that having multiple ad units having the same code works, allowing us to write much more readable configs (and splitting mediatypes, etc.).
SizeMappingV2 seems to be introducing a bug where it tries to merge all the configs (I guess?).

PS : I know I could technically merge configs but readability (and backward compatibility) is an issue here. 

## Description
Responsive rules aren't respected.

## Example
http://jsfiddle.net/6f2r5sbj/

## Steps to reproduce
This is a sample config
```javascript
[
  {
    code: 'ad-code-1',
    mediaTypes: {
      banner: {
        sizeConfig: [
          { minViewPort: [0, 0], sizes: [] },
          { minViewPort: [1000, 0], sizes: [[1000, 300], [1000, 90], [970, 250], [970, 90], [728, 90]] },
        ],
      },
    },
    bids: [{
      bidder: 'bidderA',
      params: {
        placementId: 123
      }
    }]
  }, {
    code: 'ad-code-1',
    mediaTypes: {
      banner: {
        sizeConfig: [
          { minViewPort: [0, 0], sizes: [] },
          { minViewPort: [320, 0], sizes: [[320, 100], [320, 50], [300, 50]] },
          { minViewPort: [600, 0], sizes: [] },
        ],
      },
    },
    bids: [{
      bidder: 'bidderA',
      params: {
        placementId: 456
      }
    }]
  },
]
```

### Expected results
On a 1000px+ wide resolution, only 123 should be sent to the bidder.

### Actual results
On a 1000px+ wide resolution, all configs (123 and 456) are sent with the sizes from the good viewport.

## Other information
Relates to PR #4690

Fawke:
Hey @benjaminclot,

Thanks for trying out the feature. I think we have got your use case covered here, what do you think of the following setup?

```(javascript)
var adUnits = [
      {
        code: 'ad-code-1',
        mediaTypes: {
          banner: {
            sizeConfig: [
              { minViewPort: [0, 0], sizes: [] },
              { minViewPort: [320, 0], sizes: [] },
              { minViewPort: [600, 0], sizes: [[320, 100], [320, 50], [300, 50]] },
              { minViewPort: [1000, 0], sizes: [[1000, 300], [1000, 90], [970, 250], [970, 90], [728, 90]] },
            ],
          },
        },
[...]

PR #5062, opened 2020-04-01
Size Mapping V2 supports 'Identical Ad Units'

Fawke:
## Type of change

- [x] Feature

## Description of change
This PR introduces support for '**Identical Ad Unit**' declaration in Size Mapping V2. An Identical Ad Unit is an ad unit with the same `code` as one of the existing ad units but different `mediaTypes` object.

Example of Identical Ad Units:

```(javascript)
[
  {
    code: 'ad-code-1',
    mediaTypes: {
      banner: {
        sizeConfig: [
          { minViewPort: [0, 0], sizes: [] },
          { minViewPort: [1000, 0], sizes: [[1000, 300], [1000, 90], [970, 250], [970, 90], [728, 90]] },
        ],
      },
    },
    bids: [{
      bidder: 'bidderA',
      params: {
        placementId: 123
      }
    }]
  }, {
    code: 'ad-code-1',
    mediaTypes: {
      banner: {
        sizeConfig: [
          { minViewPort: [0, 0], sizes: [] },
          { minViewPort: [320, 0], sizes: [[320, 100], [320, 50], [300, 50]] },
          { minViewPort: [600, 0], sizes: [] },
        ],
      },
    },
    bids: [{
      bidder: 'bidderA',
      params: {
        placementId: 456
      }
    }]
  },
]
```

The example above has two identical ad units with the same code `ad-code-1` but with different values for `mediaTypes` object.

The PR also improves the log messages generated by Size Mapping V2, tries to make it more intuitive and easy to understand.

Resolves https://github.com/prebid/Prebid.js/issues/5044

### DIFF
--- a/allowedModules.js
+++ b/allowedModules.js
@@ -1,11 +1,11 @@
 
 const sharedWhiteList = [
-  "core-js/library/fn/array/find", // no ie11
-  "core-js/library/fn/array/includes", // no ie11
-  "core-js/library/fn/set", // ie11 supports Set but not Set#values
-  "core-js/library/fn/string/includes", // no ie11
-  "core-js/library/fn/number/is-integer", // no ie11,
-  "core-js/library/fn/array/from" // no ie11
+  'core-js/library/fn/array/find', // no ie11
+  'core-js/library/fn/array/includes', // no ie11
+  'core-js/library/fn/set', // ie11 supports Set but not Set#values
+  'core-js/library/fn/string/includes', // no ie11
+  'core-js/library/fn/number/is-integer', // no ie11,
+  'core-js/library/fn/array/from' // no ie11
 ];
 
 module.exports = {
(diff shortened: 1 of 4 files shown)


Fawke:
This PR is ready for review. Only thing that I'd have liked to do was remove some ambiguity from the log messages.

For example:

Screenshot 2020-04-02 at 4 52 41 PM

If we have two Identical ad units (sharing the same code), from the log messages it is kind of hard to determine which ad unit got disabled, or which ad unit got an error, since ad unit code is no longer a unique identifier.

In the screenshot above, we have two identical ad units sharing the code `ad-code-1`, in which after sizeMapping filtration, the second instance of `ad-code-1` gets disabled. But looking at the log messages, there's currently no way to tell was it the first instance or the second instance.

Once this PR is merged, I'll create a separate github issue to keep track of this behaviour and provide a fix in a separate PR.

Issue #5011, opened 2020-03-19
utils.setCookie needs to use Secure with SameSite=none


goosemanjack:
## Type of issue
Bug

## Description
When a cookie is specified as 'none' for a value of SameSite the cookie should automatically be written with the Secure flag. Currently this produces a warning in Chrome as below:

digiTrustIdSystem.js:89 A cookie associated with a resource at http://prebid.local/ was set with `SameSite=None` but without `Secure`. It has been blocked, as Chrome now only delivers cookies marked `SameSite=None` if they are also marked `Secure`. You can review cookies in developer tools under Application>Storage>Cookies and see more details at https://www.chromestatus.com/feature/5633521622188032.

## Steps to reproduce
Using the DigiTrust test page "integrationExamples/gpt/digitrust_simple.html" you will get the above warning. This is due to how utils.setCookie is behaving.

## Test page
"integrationExamples/gpt/digitrust_simple.html" in source tree

### Expected results
No warning. Cookie written with Secure option.

### Actual results
Warning of impending doom.

goosemanjack:
I have submitted a PR that makes this change in the new storageManager module.
#5064 
For reference:
https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Set-Cookie

bretg:
Closing this issue since the PR has been opened and partly reviewed.

PR #5064, opened 2020-04-01
#5011 Fix to set Secure attribute on cookie when SameSite=none

goosemanjack:
## Type of change
- [X ] Bugfix

## Description of change
This change modifies the storageManager to detect if SameSite=None is used in the set cookie options. When the value "none" is specified for SameSite the browser also requires that the cookie be marked "Secure". This change automatically applies the "Secure" attribute when samesite=none is detected.

https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Set-Cookie

Tested against DigiTrust ID that sets cookie.
(/integrationExamples/gpt/digitrust_Simple.html) sample page.
Note: this must be run over HTTPS connection to operate correctly.

## Other Information
Must be run over HTTPS connection to operate correctly. Testing this over HTTP is an invalid scenario.

### DIFF
--- a/src/storageManager.js
+++ b/src/storageManager.js
@@ -64,8 +64,8 @@ export function newStorageManager({gvlid, moduleName, moduleType} = {}) {
       if (result && result.valid) {
         const domainPortion = (domain && domain !== '') ? ` ;domain=${encodeURIComponent(domain)}` : '';
         const expiresPortion = (expires && expires !== '') ? ` ;expires=${expires}` : '';
-        var isNone = (sameSite != null && sameSite.toLowerCase() == 'none')
-        var secure = (isNone) ? '; Secure' : '';
+        const isNone = (sameSite != null && sameSite.toLowerCase() == 'none')
+        const secure = (isNone) ? '; Secure' : '';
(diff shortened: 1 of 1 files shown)


src/storageManager.js
```diff
@@ -64,7 +64,9 @@ export function newStorageManager({gvlid, moduleName, moduleType} = {}) {
       if (result && result.valid) {
         const domainPortion = (domain && domain !== '') ? ` ;domain=${encodeURIComponent(domain)}` : '';
         const expiresPortion = (expires && expires !== '') ? ` ;expires=${expires}` : '';
-        document.cookie = `${key}=${encodeURIComponent(value)}${expiresPortion}; path=/${domainPortion}${sameSite ? `; SameSite=${sameSite}` : ''}`;
+        var isNone = (sameSite != null && sameSite.toLowerCase() == 'none')
+        var secure = (isNone) ? '; Secure' : '';
```
jaiminpanchal27, line 68: Can you use const here and for isNone as well.

goosemanjack: Made requested change to use const. Thanks.

PR #5065, opened 2020-04-02
Freewheel: Better size detection and allow multiple bid requests

xavierleune:
## Type of change
- [x] Bugfix

## Description of change
This pull requests enables multiple bid requests in the freewheel ssp Adapters and fixes a bug on the playerSize detection.

- cc @xwang202

### DIFF
--- a/modules/freewheel-sspBidAdapter.js
+++ b/modules/freewheel-sspBidAdapter.js
@@ -1,4 +1,5 @@
 import * as utils from '../src/utils.js';
+import { BANNER, VIDEO } from '../src/mediaTypes.js';
 import { registerBidder } from '../src/adapters/bidderFactory.js';
 
 const BIDDER_CODE = 'freewheel-ssp';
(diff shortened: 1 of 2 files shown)


xavierleune:
Hi @Fawke 
Thanks for your feedback. I'll apply the change relative to the media types strings.
However about the `getTopMostWindow`, the function seems to have a different purpose vs `utils.getWindowTop`. Indeed `getTopMostWindow` finds the last parent the current iframe (if applicable) can have access to. `utils.getWindowTop` finds the top window, but you can get a document where your iframe doesn't have write access. So I think it would be better to keep the function `getTopMostWindow` in the adapter. WDYT ?

xwang202:
Thanks for the update @xavierleune . It looks good from freewheel side.

@Fawke , I think it would be better to keep `getTopMostWindow` method as is.

xavierleune:
Thanks @Fawke :+1:

PR #5068, opened 2020-04-02
evolution_tech: completing file rename from 5060

bretg:
Completing the rename request from #5060 by removing the originally-named files


bretg:
This was merged, builds in master, and confirmed that the files are deleted. Don't know why github still marks the PR as open. Going ahead and closing it.

Issue #5054, opened 2020-03-31
Split long task


nanek:
## Type of issue
Question

## Description
I've been researching ways to decrease the [First Input Delay](https://web.dev/fid/) of our website. It is recommended to eliminate [long tasks](https://web.dev/custom-metrics/#long-tasks-api). Prebid profiles as a long task, particularly with a lot of bidders.

Good article on the topic, https://philipwalton.com/articles/idle-until-urgent/

Profile of prebid demonstrating long task.
![current](https://user-images.githubusercontent.com/6188/78059917-b6d5c280-7358-11ea-8bef-9297e6ce7331.jpg)

I've tested wrapping the [`ajaxBuilder`](https://github.com/prebid/Prebid.js/blob/master/src/ajax.js#L20) call in a `setTimeout` resulting in the ajax calls getting split into different tasks as shown in the screenshot below. I picked this location as it seemed to be where a lot of time was spent and the ajax calls are already async so it should be safe.
![split](https://user-images.githubusercontent.com/6188/78059956-c05f2a80-7358-11ea-8a9d-c97938507d13.jpg)

Are there any concerns with this approach?
Are there other places where it would be more appropriate to split the work up to avoid a long task?

Not directly related, has there been any discussion lately on using web workers to move work off of the main thread?

Thanks!

pm-harshad-mane:
Hello @nanek 
We really appreciate your findings. 
This looks good to me, can you create a PR for the same so others can also recommend changes if needed and we can pull this change into the master?

snapwich:
I like the idea in general. I don't think wrapping the ajax call in a `setTimeout` is a good place to split for two reasons.
1) I'm not sure that this would actually improve performance. The work done in the ajax builder is very minimal (just building an XMLHttpRequest object) and these ajax requests are already asynchronous and non-blocking which means they are technically split work as the responses are handled on a separate call stack when they come in.
2) We don't want to delay our auction requests. For header bidding to work it's very important that the actual header bidding requests are sent ASAP to allow the auction sufficient time to perform before the page finishes rendering.

That said, there is probably places where we can split work with minimal effect on the auction and I'd be open to doing that.

Is it possible to set up a test page that I could look at your results? The flame graphs are interesting, but since it doesn't include a timeline and I can't dig into the stack it's hard for me to tell if the difference is as pronounced as the images show, or if the differences are ideal (i.e. if the auction was delayed).

Let me know, thanks!

PR #5069, opened 2020-04-02
Wrap ajax call in setTimeout to split long task.

nanek:
## Type of change
Feature

## Description of change
Implements https://github.com/prebid/Prebid.js/issues/5054
Wraps ajax call in setTimeout to split out long task into many.

Is this safe, or does it change the order in an unexpected way?

### DIFF
--- a/src/ajax.js
+++ b/src/ajax.js
@@ -18,83 +18,85 @@ export const ajax = ajaxBuilder();
 
 export function ajaxBuilder(timeout = 3000, {request, done} = {}) {
   return function(url, callback, data, options = {}) {
-    try {
-      let x;
-      let method = options.method || (data ? 'POST' : 'GET');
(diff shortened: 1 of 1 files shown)


pm-harshad-mane:
Hi @nanek ,
I believe we should keep the ajax package simple, as it is used for making many other calls than just bidders (like fetching currency, RTD, UserIds).
The long task that needs to be fixed can be modified as required with setTimeout code, 
Here we call all bidders so we can add setTimeout at 
https://github.com/prebid/Prebid.js/blob/master/src/adapterManager.js#L352 and 
https://github.com/prebid/Prebid.js/blob/master/src/adapterManager.js#L379
What do you think?

bretg:
@nanek - please take a look at this failing test. It doesn't appear to the same flakiness that plagues other PRs.

PR #5070, opened 2020-04-02
IdentityLink check ats after delaySync

pycnvr:
## Type of change

- [x] Bugfix
- [ ] Feature
- [ ] New bidder adapter   
- [ ] Build related changes
- [ ] CI related changes
- [ ] Other

## Description of change
In IdentityLinkIdSystem, the check for window.ats occurs too early.  There is a possibility that ats hasn't been loaded yet, so the id is not generated and stored optimally.  Moving the check to inside the callback so it's applied after delaySync.

- pyang@conversantmedia.com


bretg:
@mamatic - as the originator of this module - giving you a chance to review these changes

mamatic:
LGTM, it's good that you found that bug, great.

pycnvr:
@bretg @mamatic Thanks guys!

Issue #5033, opened 2020-03-26
Question: utils.isSafariBrowser includes iOS-Chrome and iOS-FireFox


r-ishigami:
## Type of issue

Question

## Description

utils.isSafariBrowser regular expressions include `iOS x Chrome` and `iOS x FireFox` .
Is it a specification?

https://github.com/prebid/Prebid.js/blob/7c456268913024c77b72aa0d0b0230b0d509aba4/src/utils.js#L817-L819

- iPhone Chrome UserAgent

```
Mozilla/5.0 (iPhone; CPU iPhone OS 13_3 like Mac OS X) AppleWebKit/605.1.15 (KHTML, like Gecko) CriOS/80.0.3987.95 Mobile/15E148 Safari/604.1
```

- iPhone FireFox UserAgent

```
Mozilla/5.0 (iPhone; CPU OS 13_3_1 like Mac OS X) AppleWebKit/605.1.15 (KHTML, like Gecko) FxiOS/23.0  Mobile/15E148 Safari/605.1.15
```

## Proposal

```javascript
export function isSafariBrowser() {
  return /^((?!chrome|android|crios|fxios).)*safari/i.test(navigator.userAgent);
}
```

## Relation

https://github.com/prebid/Prebid.js/issues/3095

bretg:
Thanks @r-ishigami - PR opened, you're welcome to review.

r-ishigami:
Thanks, @bretg :-) Reviewed it.

PR #5077, opened 2020-04-05
isSafariBrowser update 

bretg:
## Type of change

- [X] Bugfix

## Description of change
Per issue #5033, the isSafariBrowser function didn't handle the case of Chrome or Firefox on iOS

## Other information
There weren't any unit tests for utils.isSafariBrowser() -- added

### DIFF
--- a/src/utils.js
+++ b/src/utils.js
@@ -815,7 +815,7 @@ export function inIframe() {
 }
 
 export function isSafariBrowser() {
-  return /^((?!chrome|android).)*safari/i.test(navigator.userAgent);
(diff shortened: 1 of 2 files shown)


bretg:
@idettman - I tried to save the `navigator` object and restore it, but received this error:
```
  Uncaught TypeError: Cannot assign to read only property 'navigator' of object '#<Window>'
  at webpack:///test/spec/utils_spec.js:948:14 <- test/test_index.js:332851:15
```

idettman:
@bretg I committed an update that resets navigator.userAgent to it's initial value after each test

PR #5078, opened 2020-04-06
Fix bug in rubicon bid adapter for fpd keywords

robertrmartinez:
## Type of change
- [X] Bugfix

## Description of change
If the user input non array into keywords it would break and not send bids.

Here is the new test without the isArray check:

image


mgriego:
@robertrmartinez @bretg I know this has long since been merged in, but I'm curious your thoughts about this becoming a silent error?  Passing invalid data now won't cause the bid to fail completely, but the publisher now has no feedback that they did something wrong.  Should there at least be a debugging message when data supplied by the caller is outright ignored due to bad formatting?

bretg:
I'm fine with a message, but it would need to be short. Very short. The PBJS package is already too big, and error messages don't compress well.

PR #5082, opened 2020-04-06
Fix for key-values and multiple bids having the same adUnitCode

benjaminclot:
## Type of change
- [x] Bugfix

## Description of change
When sending multiple bids with the same adUnitCode, the key-value targeting sent to the adserver is often wrong as the matching is based solely on the adUnitCode. This is an improvement, using the requestId.


benjaminclot:
I'm also asking myself if there aren't any other places where this check might be enforced?

benjaminclot:
> A test should be added checking that multiple bids do not have the same adUnitCode
Is this really a problem? What should happen? What if the mediatype is different?

idettman:
Basically, a test needs to be added to verify this patch

PR #5083, opened 2020-04-06
PS bid adapter update to copy site object from config

mmoschovas:
## Type of change
- [X] Other

## Description of change
Updated Prebid Server bid adapter to check for the site object in the prebid configuration. If exists, this will be copied over into the request and publisher/page will either be created properties or will overwrite existing config defined values. This will allow other OpenRTB fields within the site object to be passed to Prebid Server, i.e. site.content

## Other information
[#3783](https://github.com/prebid/Prebid.js/pull/3783)

### DIFF
--- a/modules/prebidServerBidAdapter/index.js
+++ b/modules/prebidServerBidAdapter/index.js
@@ -268,10 +268,9 @@ function _appendSiteAppDevice(request, pageUrl) {
     request.app = config.getConfig('app');
     request.app.publisher = {id: _s2sConfig.accountId}
   } else {
-    request.site = {
-      publisher: { id: _s2sConfig.accountId },
-      page: pageUrl
-    }
+    request.site = (typeof config.getConfig('site') === 'object') ? config.getConfig('site') : {};
+    request.site.publisher = { id: _s2sConfig.accountId },
+    request.site.page = pageUrl;
   }
   if (typeof config.getConfig('device') === 'object') {
     request.device = config.getConfig('device');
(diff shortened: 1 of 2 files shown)


jsnellbaker:
@mmoschovas Please check the lint errors in the circleci test report link and fix them.

modules/prebidServerBidAdapter/index.js
```diff
@@ -268,10 +268,9 @@ function _appendSiteAppDevice(request, pageUrl) {
     request.app = config.getConfig('app');
     request.app.publisher = {id: _s2sConfig.accountId}
   } else {
-    request.site = {
-      publisher: { id: _s2sConfig.accountId },
-      page: pageUrl
-    }
+    request.site = (typeof config.getConfig('site') === 'object') ? config.getConfig('site') : {};
+    request.site.publisher = { id: _s2sConfig.accountId },
+    request.site.page = pageUrl;
```
robertrmartinez, line 273: So if someone sets `site` in setConfig, we want to use it.

But we want to overwrite `publisher` and `page` no matter what?

Here is what the publisher object specifies in openRTB protocol:

image

So if we did:

```
pbjs.setConfig({
   site: {
      ...
      publisher: {
         id: 'some id'
         domain: 'prebid.org'
         ...
      }
});
```

Then we would basically not include the other keys in the user defined publisher object.

Maybe it makes more sense to use `utils.deepSet` in order to set the `id` of publisher, so that if `publisher` already exists we do not overwrite the entire thing but just the `id` key?

robertrmartinez: Please add a test for this as well

idettman:
Changed pull request text formatting to better match Prebid.js pull request conventions. 
I also added a related (but closed) pull request since it had a significant conversion.

aleksatr:
LGTM

PR #5086, opened 2020-04-07
Read OpenRTB app objects if set in config + bug fix for when ad units are reloaded

bjorn-lw:
## Type of change

- [x ] Bugfix
- [ x] Feature

## Description of change
If OpenRTB app objects has been set in config, use them.
In analytics adapter, if an ad unit was reloaded in a new auction, available bids on the ad unit was not recorded.


GLStephen:
@bjorn-lw tests are passing locally, working on figuring out the circleci failure.

PR #5087, opened 2020-04-07
RTD bug fix

omerBrowsi:
## Type of change

- [X ] Bugfix

## Description of change
**RTD Module**
validate data before targeting (setting key values) GPT - remove non-string values. DFP only accept a string as key-value, non-string will cause an error

**Browsi RTD sub module**

- remove unused variables from functions
- support both ad unit path (e.g: "/122345/path/unit") and placement id as ad unit code
- remove double encoding for URL (browsi logic to fetch real-time data)


modules/rtdModule/index.js
```diff
@@ -128,6 +128,31 @@ function getProviderData(adUnits, callback) {
...
+  // data must be an object, contains object with string as value
+  if (typeof data !== 'object') {
+    return {};
+  }
+  for (let key in data) {
+    if (data.hasOwnProperty(key)) {
+      for (let innerKey in data[key]) {
+        if (data[key].hasOwnProperty(innerKey)) {
+          if (typeof data[key][innerKey] !== 'string') {
+            delete data[key][innerKey];
```
mike-chowla, line 147: Dropping data without a warning that data is being dropped seems problematic to me

modules/rtdModule/index.js
```diff
@@ -128,6 +128,31 @@ function getProviderData(adUnits, callback) {
   }
 }
 
+/**
+ * delete invalid data received from provider
+ * this is to ensure working flow for GPT
+ * @param {Object} data received from provider
+ * @return {Object} valid data for GPT targeting
+ */
+function validateProviderDataForGPT(data) {
```
mike-chowla, line 137: Why only validate for GPT?  This module appears only to work with GPT.  What the rational for having an interface that ostensibly supports no string types when GPT doesn't and no other ad server is supported?

modules/rtdModule/index.js
```diff
@@ -200,6 +225,7 @@ function setDataForPrimaryAdServer(data) {
     utils.logError('window.googletag is not defined on the page');
     return;
   }
+  data = validateProviderDataForGPT(data);
```
mike-chowla, line 228: Why not validate the data when it's returned from the adapter instead of here?  When it's done here, it's not possible to issue a warning but which adapter's data is being dropped.

modules/rtdModule/index.js
mike-chowla, line 137: No tests are added for this function.  It appears the rtd module itself has no tests and only the sub-modules are tested which shouldn't be the case.

modules/rtdModule/index.js
omerBrowsi: added warning

modules/rtdModule/index.js
omerBrowsi: the module is setting data for GPT and bids,
bids don't have the same limit

modules/rtdModule/index.js
omerBrowsi: because bids don't have the same limit, we don't want to drop data that can be set to bids

we are about to do some changes to this module soon, among them is separate data for bids and GPT. 
When we have this the validation will be right after the data received as you suggest.

modules/rtdModule/index.js
omerBrowsi: I added a test for this function

The RTD module and sub-modules are tested together in the same test for now

RTD module has over 80%  test coverage

omerBrowsi:
I added some test to the submodule and RTD module
the RTD module has over 80% coverage (had  it also before changes), please note it's not placed under "modules"

browse rtd module now has 78%, all of the important processes are covered, the main part missing test is the ajax call that we don't need to test

Screen Shot 2020-04-21 at 20 33 07

omerBrowsi:
two extra fixes:

1. changes to browsi submodule so it will be able to return predictions if there are no slots on the page yet
2. RTD module - before set target on GPT, make sure GPT pubads is defined (if not - use google tag cmd queue)

Asafsham:
Hey @mike-chowla, any chance we can get it reviewed for this week's release? Thanks!

mike-chowla:
@Asafsham Because this PR changes the core RTD module, it needs a second review.  I'm looking for a second reviewer

Asafsham:
@mike-chowla Cool, thanks for the update.

Asafsham:
@robertrmartinez Will you be able to get to this? Thanks!

smenzer:
@robertrmartinez can you please take a second review of this?

Asafsham:
@robertrmartinez @mike-chowla @smenzer @bretg Hey Guys!
We are waiting for approval and second PR for the last few versions now and we would really appreciate it if you could help us get this done before we start the new RTD.

robertrmartinez:
@Asafsham Yes I will get this reviewed today, thanks!

robertrmartinez:
@Asafsham Can you please pull the latest master?

I believe it has fixes for the failing Circle CI tests

robertrmartinez:
@omerBrowsi Please pull latest from master!

omerBrowsi:
@RobertLippens pulled, thanks

modules/browsiRtdProvider.js
```diff
@@ -75,7 +75,7 @@ function collectData() {
       sk: _moduleParams.siteKey,
       sw: (win.screen && win.screen.width) || -1,
       sh: (win.screen && win.screen.height) || -1,
-      url: encodeURIComponent(`${doc.location.protocol}//${doc.location.host}${doc.location.pathname}`),
+      url: `${doc.location.protocol}//${doc.location.host}${doc.location.pathname}`,
```
robertrmartinez, line 78: What is the reasoning behind not executing encode anymore?

modules/browsiRtdProvider.js
```diff
@@ -184,25 +178,24 @@ export function isIdMatchingAdUnit(id, slot, whitelist) {
...
  * @return {?Object}
  */
-function getSlotById(id) {
+function getSlotByCode(code) {
   const slots = getAllSlots();
   if (!slots || !slots.length) {
     return null;
   }
-  return slots.filter(s => s.getSlotElementId() === id)[0] || null;
+  return slots.find(s => s.getSlotElementId() === code || s.getAdUnitPath() === code) || null;
```
robertrmartinez, line 189: `.find` is not supported by ie11, can you please use the find from the core-js lib?

Example Import: https://github.com/prebid/Prebid.js/blob/master/modules/appnexusBidAdapter.js#L7

Example Run: https://github.com/prebid/Prebid.js/blob/master/modules/appnexusBidAdapter.js#L69

modules/rtdModule/index.js
```diff
@@ -196,11 +222,12 @@ export function requestBidsHook(fn, reqBidsConfigObj) {
...
-  if (!utils.isGptPubadsDefined()) {
-    utils.logError('window.googletag is not defined on the page');
-    return;
-  }
-  targeting.setTargetingForGPT(data, null);
+  data = validateProviderDataForGPT(data);
+  utils.isGptPubadsDefined()
+    ? targeting.setTargetingForGPT(data, null)
+    : window.googletag.cmd.push(() => {
+      targeting.setTargetingForGPT(data, null);
```
robertrmartinez, line 229: 👍 For pushing on the googletag cmd que, but should we check to make sure all of it <window.googletag.cmd> is defined before trying to execute the push though?

If either `googletag` or `cmd` is not defined by the time this runs it will throw an error I believe.

modules/browsiRtdProvider.js
omerBrowsi: we do encoding for the object later - 
https://github.com/prebid/Prebid.js/blob/2b428bc60f5a4a3e043f1ef465cef969ab9aba7a/modules/browsiRtdProvider.js#L278
this line cause double encoding

modules/browsiRtdProvider.js
omerBrowsi: fixed

modules/rtdModule/index.js
omerBrowsi: Yes, you are right
I added googletag init

Asafsham:
Hey @robertrmartinez! Let me know if there's anything that we need to add here, would love if it would be approved for next version.

modules/rtdModule/index.js
robertrmartinez: Nice!

robertrmartinez:
Thanks for your patience. Great job! Merging now.

PR #5090, opened 2020-04-07
PubMatic bid adapter to support app object set using setConfig

pm-harshad-mane:
## Type of change
- [ X ] Feature

## Description of change
PubMatic bid adapter to support app object set using setConfig
```
pbjs.setConfig({
   app: {
      bundle: "org.prebid.mobile.demoapp",
      domain: "prebid.org"
   }
});
```


pm-harshad-mane:
Thank you @jsnellbaker !!

PR #5017, opened 2020-03-23
Add Automatad Bid Adapter

kanchika-automatad:
## Type of change

- [ ] Bugfix
- [ ] Feature
- [x] New bidder adapter   
- [ ] Code style update (formatting, local variables)
- [ ] Refactoring (no functional changes, no api changes)
- [ ] Build related changes
- [ ] CI related changes
- [ ] Does this change affect user-facing APIs or examples documented on http://prebid.org?
- [ ] Other

## Description of change

- test parameters for validating bids
Banner Test Parameters:
```
var adUnits = [
  {
    code: 'test-div',
    mediaTypes: {
      banner: {
        sizes: [[300,600]]
      }
    },
    bids: [
      {
          bidder: 'automatad',
          params: {
              siteId: 'atd-site-id',
              placementId: 'atd-placement-id'
          }
      }
    ]
  }
];
```

Be sure to test the integration with your adserver using the [Hello World](/integrationExamples/gpt/hello_world.html) sample page.

- contact email of the adapter’s maintainer: tech@automatad.com
- [x] official adapter submission

For any changes that affect user-facing APIs or example code documented on http://prebid.org, please provide:

- A link to a PR on the docs repo: [prebid/prebid.github.io#1876](https://github.com/prebid/prebid.github.io/pull/1876)

PR #5101, opened 2020-04-09
Add Relaido Adapter

relaido:
## Type of change
- [x] New bidder adapter

## Description of change
New adapter

```
{
  bidder: 'relaido',
  params: {
    placementId: '9900'
  }
}
```

- contact email of the adapter’s maintainer: video-dev@cg.relaido.co.jp

https://github.com/prebid/prebid.github.io/pull/1917

### DIFF
--- a/modules/relaidoBidAdapter.js
+++ b/modules/relaidoBidAdapter.js
@@ -132,9 +132,13 @@ function interpretResponse(serverResponse, bidRequest) {
 }
 
 function getUserSyncs(syncOptions, serverResponses) {
-  const resSyncUrl = utils.deepAccess(serverResponses, '0.body.syncUrl');
-  const proSyncUrl = `https://${BIDDER_DOMAIN}/tr/v1/prebid/sync.html`;
-  const syncUrl = resSyncUrl || proSyncUrl;
+  if (!syncOptions.iframeEnabled) {
+    return [];
+  }
+  let syncUrl = `https://${BIDDER_DOMAIN}/tr/v1/prebid/sync.html`;
+  if (serverResponses.length > 0) {
+    syncUrl = utils.deepAccess(serverResponses, '0.body.syncUrl') || syncUrl;
+  }
   receiveMessage();
   return [{
     type: 'iframe',
@@ -218,7 +222,7 @@ function receiveMessage() {
 }
 
 function setUuid(e) {
-  if (e.data && e.data.relaido_uuid) {
+  if (utils.isPlainObject(e.data) && e.data.relaido_uuid) {
     storage.setDataInLocalStorage(UUID_KEY, e.data.relaido_uuid);
     window.removeEventListener('message', setUuid);
   }
--- a/modules/relaidoBidAdapter.js
+++ b/modules/relaidoBidAdapter.js
@@ -1,0 +1,296 @@
+import * as utils from '../src/utils.js';
+import { registerBidder } from '../src/adapters/bidderFactory.js';
+import { BANNER, VIDEO } from '../src/mediaTypes.js';
+import { Renderer } from '../src/Renderer.js';
+import { getStorageManager } from '../src/storageManager.js';
+
+const BIDDER_CODE = 'relaido';
+const BIDDER_DOMAIN = 'api.relaido.jp';
+const ADAPTER_VERSION = '1.0.0';
+const DEFAULT_TTL = 300;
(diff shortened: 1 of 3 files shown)


relaido:
Hi @Fawke , thanks for the review.
event listener removed.

Fawke:
@relaido

Thanks for making the change. But, when I'm trying to load your banner ad and video ad following the example you shared in the md file, and plug those values in the [hello_world](https://github.com/prebid/Prebid.js/blob/master/integrationExamples/gpt/hello_world.html) example, am getting this error. 

Screenshot 2020-04-09 at 5 40 24 PM

I think your `isBidRequestValid` function is returning false because it doesn't find the uuid in localStorage. Can you have a look please?

relaido:
@Fawke

The first access results in an error because there is no uuid.
Since uuid is registered after `UserSync`, it succeeds from the second time on.

Fawke:
@relaido 

I put your adapter js file through a debugger and made some observations:

1. In the function `setUuid(e)`, this `if condition` (`(e.data && e.data.relaido_uuid)`) is always returning a false. It maybe because the value of `e.data` is a `string`.

2. In your `getUserSyncs` function, you need to have a check in place like this: `syncOtions.iframeEnabled`. Please take a look at a similar implementation of the `appnexus` or the `rubicon` adapter.

3. Also, in the same function `getUserSyncs`, the parameter `serverResponses` is coming empty. So, you might better wanna put a check in place, something like: `severResponses.length > 0`

(serverResponses is empty because no auction is taking place because the function, `isBidRequestValid` is returning a false)

Also, are there any particular reasons for setting the localStorage value in getUserSyncs function? What if the publisher doesn't allow userSync? What if the pub only allows image sync and not iframe sync?



> Since uuid is registered after `UserSync`, it succeeds from the second time on.

What do you mean second time on? You mean if I refresh the page again? A bit confused, because your `isBidRequestValid` function is always returning a false, leading to no auction taking place for your bidder. So, how will it succeed the second time on?

relaido:
@Fawke

1.
> this if condition ((e.data && e.data.relaido_uuid)) is always returning a false

Sometimes returns true. e.data can also send objects.

> It maybe because the value of e.data is a string.

You're right. Check with isPlainObject.

2.
> In your getUserSyncs function, you need to have a check in place like this: syncOtions.iframeEnabled.

Fixed. However, some people have said this.
https://github.com/prebid/Prebid.js/pull/5017#discussion_r399485438

3.
> So, you might better wanna put a check in place, something like: severResponses.length > 0

Fixed.

---

> Also, are there any particular reasons for setting the localStorage value in getUserSyncs function?

I want to save uuid in localStorage at the first access.
When the page is updated again, the auction is performed using the uuid stored in localStorage.

> What if the publisher doesn't allow userSync?

Unfortunately, no auction is possible.

> What if the pub only allows image sync and not iframe sync?

Auction cannot be done here as well.

---

> What do you mean second time on?

Is to visit the same page again.
It is the same as refreshing the page.

> You mean if I refresh the page again?

The auction will be held because uuid is stored in localStorage.

> A bit confused, because your isBidRequestValid function is always returning a false, leading to no auction taking place for your bidder.

Hmm. I can conduct an auction.

> So, how will it succeed the second time on?

Do you allow userSync and iframe sync?
Is relaido_uuid stored in localStorage?

Fawke:
@relaido 
> Fixed. However, some people have said this.
> [#5017 (comment)](https://github.com/prebid/Prebid.js/pull/5017#discussion_r399485438)

Unfortunately, that's not right. We still need the check in place. The deprecation was for the publisher facing API, but internally, the adapter is still using `iframeEnabled.`

Also, I noticed one thing, when sending a banner response, you are returning a `vast` and overwriting `bidResponse.ad` with your renderer code.

Is this done to overcome some of prebid's limitations?

Rest of things look okay to me.

relaido:
@Fawke

> The deprecation was for the publisher facing API, but internally, the adapter is still using iframeEnabled.

Thanks. understood.

> Is this done to overcome some of prebid's limitations?

Banner could be a safeframe. to overcome that, I pass VASTXML as a parameter.
I referenced this program.
https://github.com/prebid/Prebid.js/blob/master/modules/adgenerationBidAdapter.js#L112-L120

relaido:
@Fawke

doc also please review.
https://github.com/prebid/prebid.github.io/pull/1917

PR #5122, opened 2020-04-14
[BUGFIX] Teads: fix production GDPR problem

valsouche:
## Type of change

- [x] Bugfix

## Description of change
We had a javascript error when `gdprApplies` equals false OR when there is a consentManagement module enabled but no CMP on the page when we try to access `vendorData` which is `undefined` in these cases.

We saw that `apiVersion` could be equals to 0. Could you please explain (or give us a link) what this value means ? As we are only using 1 and 2 from our adapter.

- contact email of the adapter’s maintainer: innov-ssp@teads.tv

### DIFF
--- a/modules/teadsBidAdapter.js
+++ b/modules/teadsBidAdapter.js
@@ -54,7 +54,9 @@ export const spec = {
     if (bidderRequest && gdpr) {
       let isCmp = (typeof gdpr.gdprApplies === 'boolean')
       let isConsentString = (typeof gdpr.consentString === 'string')
-      let status = isCmp ? findGdprStatus(gdpr) : gdprStatus.CMP_NOT_FOUND_OR_ERROR
+      let status = isCmp
+        ? findGdprStatus(gdpr.gdprApplies, gdpr.vendorData, gdpr.apiVersion)
+        : gdprStatus.CMP_NOT_FOUND_OR_ERROR
       payload.gdpr_iab = {
         consent: isConsentString ? gdpr.consentString : '',
         status: status,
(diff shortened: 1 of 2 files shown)


valsouche:
Hello,
We are struggling to understand one TCF V2 point. From our adapter, in case of cmp using V2, we are accessing `isServiceSpecific` directly at the top of `vendorData` object, but we are not sure anymore this is the good way to do.
In the source code, we saw that the `vendorData` object is directly the CMP one (https://github.com/prebid/Prebid.js/blob/master/modules/consentManagement.js#L374). In V1 vendorData is equal to `CMPObject.getVendorConsents`. 
In TCFV2, `getVendorConsents` has been replaced by `getTCData`, but except for the static configuration (https://github.com/prebid/Prebid.js/blob/master/modules/consentManagement.js#L317) we don't see any usage of this getTCData. 

Our question is: from the adapter, do we need to access by way : `vendorData.getTCData.isServiceSpecific` or this one: `vendorData.isServiceSpecific` ? 
And if the answer is 1, that means we need to handle the case where we are in static conf, right ?
If the answer is 2, does it cover all cases ?

I hope you will be able to clarify this point.
Thanks in advance,
Valentin.

PS: please do not merge the PR until this point is ok

jsnellbaker:
@valsouche 

The `getTCData()` is indirectly used by the event listener that we attach to the TCF2 CMP.  The response of the CMP provides (when one of the defined events triggers) is the same object response we would get if we were to directly invoke the `getTCData()` function against the CMP.

In our case, we normalize the data from the static approach and store the full consent data object directly under the `vendorData` property so that bidders don't have an extra layer to access the information they need.  

So if you needed to read the `isServiceSpecific` field from the consent data - you should go with the latter approach you noted (`vendorData.isServiceSpecific`).

Please let me know if this makes sense and/or if you have any further questions on the topic.

valsouche:
@jsnellbaker 
 
Perfect, that's pretty clear now.
All is ok for us, 
thank you !

PR #5123, opened 2020-04-14
parrableIdSystem: Send current page location to back-end

icflournoy:
## Type of change

- [x] Bugfix

## Description of change

Send the page location (window.location.href) as determined by the logic in the refererDetection module to Parrable's ID system.

The non-Prebid integration with Parrable sends the value of `window.location.href` when making a request to our system. This change makes Prebid more closely match our non-Prebid integration.

### DIFF
--- a/modules/parrableIdSystem.js
+++ b/modules/parrableIdSystem.js
@@ -6,8 +6,9 @@
  */
 
 import * as utils from '../src/utils.js'
-import {ajax} from '../src/ajax.js';
-import {submodule} from '../src/hook.js';
+import { ajax } from '../src/ajax.js';
+import { submodule } from '../src/hook.js';
+import { getRefererInfo } from '../src/refererDetection.js';
 
 const PARRABLE_URL = 'https://h.parrable.com/prebid';
 
@@ -26,9 +27,12 @@ function isValidConfig(configParams) {
 function fetchId(configParams, consentData, currentStoredId) {
   if (!isValidConfig(configParams)) return;
 
+  const { referer } = getRefererInfo();
+
   const data = {
     eid: currentStoredId || null,
-    trackers: configParams.partner.split(',')
+    trackers: configParams.partner.split(','),
+    url: referer
   };
 
   const searchParams = {
--- a/modules/parrableIdSystem.js
+++ b/modules/parrableIdSystem.js
@@ -27,12 +27,12 @@ function isValidConfig(configParams) {
 function fetchId(configParams, consentData, currentStoredId) {
   if (!isValidConfig(configParams)) return;
 
-  const { referer } = getRefererInfo();
+  const refererInfo = getRefererInfo();
 
   const data = {
     eid: currentStoredId || null,
     trackers: configParams.partner.split(','),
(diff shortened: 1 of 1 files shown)


icflournoy:
Hey @jdwieland8282 @jaiminpanchal27 I see that the CircleCI build failed on the BrowserStack step for this specific test:

```
TOTAL: 1 FAILED, 48977 SUCCESS

1) should notify after timeout period
     ExpiringQueue
     Timeout of 2000ms exceeded. For async tests and hooks, ensure "done()" is called; if returning a Promise, ensure it resolves.
```

This seems like a timing issue and not related to my change. Could you point me in the right direction on re-running a build, or updating the failing test so it is more robust?

harpere:
@icflournoy unfortunately these tests are flakey and fail sporadically.  Usually the reviewer will rerun if necessary.

icflournoy:
Hey, thanks for the feedback @harpere 

I spent a bit of time trying to figure out how to add a good test for this additional property, but could not find an easy way forward. The property is something we are sending off to our backend servers, via an Xhr request.

I see that Prebid has an Xhr mock that is used in a few places to create a fake http server:
https://github.com/prebid/Prebid.js/blob/7e26de81825b8f0854b0bd47fdc9d3a353a765fd/test/spec/AnalyticsAdapter_spec.js#L40

This part makes sense to me.

I'm having trouble getting the parrableIdSystem module to actually have its `getId` callback method invoked during my test so that the mocked xhr object catches the outgoing request.

My understanding is that when `userSync.syncDelay` is set to `0` then the userId submodule callbacks are immediately invoked:
https://github.com/prebid/Prebid.js/blob/7e26de81825b8f0854b0bd47fdc9d3a353a765fd/modules/userId/index.js#L337-L343

but that doesn't seem to be happening.

**General development questions:**
- How can I run only my `parrableIdSystem_spec.js` test? There is a note in the gulpfile that says I can do `gulp test --file path/to/my/test.js` but this does not work (fails on using `import` syntax)
https://github.com/prebid/Prebid.js/blob/7e26de81825b8f0854b0bd47fdc9d3a353a765fd/gulpfile.js#L216

- How do I enable debug logging during the test run? I've tried settings `DEBUG_MODE=TRUE` in `constants.json`, among other things, to get the utils.log* methods to print out, perhaps they are, but I do not see them when running the karma tests in a headless browser.

harpere:
Hi @icflournoy - I don't think it's worth all that trouble to add the test.  Sorry to put you through that, but looking on the bright side, maybe a good learning experience!   Going to try to get you answers to your dev questions, but in the mean time will approve and merge this.

idettman:
> How can I run only my parrableIdSystem_spec.js test? There is a note in the gulpfile that says I  can do gulp test --file path/to/my/test.js but this does not work (fails on using import syntax)
Unfortunately, the ability to run a single test file is not functional at the moment, I'm not sure if there are plans to fix or not yet. But you can add .skip or .only to the describe blocks to ignore or isolate.

to add debug logging you'll need to add `eslint` ignore comments before your console.log calls

PR #5124, opened 2020-04-15
GumGum: uses encodeURIComponent inline

susyt:
## Type of change

- [ ] Bugfix
- [ ] Feature
- [ ] New bidder adapter   
- [ ] Build related changes
- [ ] CI related changes
- [X] Other

## Description of change

Encodes a param object instead of relying on parseQueryStringParameters

### DIFF
--- a/modules/gumgumBidAdapter.js
+++ b/modules/gumgumBidAdapter.js
@@ -60,7 +60,7 @@ function _getBrowserParams(topWindowUrl) {
     pu: topUrl,
     ce: storage.cookiesAreEnabled(),
     dpr: topWindow.devicePixelRatio || 1,
-    jcsi: JSON.stringify({ t: 0, rq: 8 }),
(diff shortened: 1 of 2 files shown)


susyt:
Hi @jsnellbaker @harpere ,

Is it possible to get this reviewed and released quickly? the release today (3.16.0) causes curly braces to be left in our query string which we'd like to avoid

thanks!

harpere:
Hi @susyt - sounds like a test may be needed around this parameter. 
Also, are you suggesting we try to release this prior to next week's 3.17.0 release?

susyt:
Thank you for getting back to me so quickly! yes, I apologize for asking but my team/manager has asked if it were possible to get it released sooner

harpere:
merged, but we rarely do early releases, and even then only to fix core problems, not adapters.

PR #5139, opened 2020-04-17
Triplelift: expose tl_souce in bid response

sdao-tl:
## Type of change

- [ ] Bugfix
- [x] Feature
- [ ] New bidder adapter   
- [ ] Build related changes
- [ ] CI related changes
- [ ] Other

## Description of change

- Expose 'tl_source' from 3lift bid response to prebid response
- Update TTL to 300


patmmccann:
what is the tl_source? can we use this to tell native from display instead of size?

PR #5157, opened 2020-04-23
Vidazoo Adapter: send the adUnitCode in request body

uditalias:
## Type of change
- [x] Feature

## Description of change

The bid `adUnitCode` is added to the request body


uditalias:
@mkendall07 Is something else we should do here?

PR #5161, opened 2020-04-23
EMX Digital Adapter Bug Fix - Outstream Renderer 

EMXDigital:
## Type of change

- [X] Bugfix

## Description of change
- fix issue where outstream renderer object is not included in bid responses


modules/emx_digitalBidAdapter.js
```diff
@@ -41,10 +41,14 @@ export const emxAdapter = {
   },
   formatVideoResponse: (bidResponse, emxBid, bidRequest) => {
     bidResponse.vastXml = emxBid.adm;
-    if (bidRequest.bidRequest && bidRequest.bidRequest.mediaTypes && bidRequest.bidRequest.mediaTypes.video && bidRequest.bidRequest.mediaTypes.video.context === 'outstream') {
-      bidResponse.renderer = emxAdapter.createRenderer(bidResponse, {
-        id: emxBid.id,
-        url: RENDERER_URL
+    if (bidRequest.bidderRequest && bidRequest.bidderRequest.bids && bidRequest.bidderRequest.bids.length > 0) {
+      bidRequest.bidderRequest.bids.forEach((bid) => {
+        if (bidResponse.requestId && bid.bidId && bidResponse.requestId === bid.bidId && bid.mediaTypes && bid.mediaTypes.video && bid.mediaTypes.video.context === 'outstream') {
```
msm0504, line 46: Is it possible that `bidResponse.requestId` matches more than 1 `bid.bidId`?
If not, use `core-js/library/fn/array/find` instead of `forEach`

```
const matchingBid = find(bidRequest.bidderRequest.bids, bid => <if condition>);
if (matchingBid) {
<create renderer>
}
```

EMXDigital:
Thanks for the recommendation. We modified the code to reflect it. Please take a look at the pull request again.

modules/emx_digitalBidAdapter.js
EMXDigital: Thanks for the recommendation. We modified the code to reflect it. Please take a look at the pull request again.

PR #5163, opened 2020-04-24
PubDirect UserId Module submission

pub-direct:
## Type of change

- [ ] Bugfix
- [ ] Feature
- [x] New bidder adapter   
- [ ] Build related changes
- [ ] CI related changes
- [ ] Other

## Description of change

Pub.direct is a free identity management platform. This pull request adds a new submodule `pubDirectSubmodule` to userId.js. The `pubDirectSubmodule` allows users to sync user Ids with pub.direct as well as passing 1st party data to be stored in pub.direct's datastore.
Maintainer: wenda@pub.direct 
Here is the link to the documentation pull request: [https://github.com/prebid/prebid.github.io/pull/1943](https://github.com/prebid/prebid.github.io/pull/1943)

### DIFF
--- a/modules/pubDirectSystem.js
+++ b/modules/pubDirectSystem.js
@@ -10,7 +10,7 @@ import {ajax} from '../src/ajax.js';
 import {submodule} from '../src/hook.js';
 
 function parseData(networkCode, segments) {
-  if (segments && networkCode) {
+  if (Object.keys(segments) !== 0 && networkCode && JSON.parse(localStorage.getItem('pubId'))) {
     const userId = JSON.parse(localStorage.getItem('pubId'));
     const url = `https://api.staging.pub.direct/aud-xchg/v1/sync/${networkCode}/${userId.ID}`;
     ajax(url, undefined, JSON.stringify(segments), undefined);
@@ -30,10 +30,15 @@ export const pubDirectSubmodule = {
    * @returns {(Object|undefined)}
    */
   decode(value, config) {
-    if (!config) {
+    // eslint-disable-next-line no-console
+    console.log(config, 'config pb');
+    // eslint-disable-next-line no-console
+    console.log(value, 'value pb');
+
+    if (config.params.segments) {
       parseData(config.params.publisherId, config.params.segments);
     }
-    return (value && typeof value['PUBDIRECTID'] === 'string') ? { 'pubDirectId': value['PUBDIRECTID'] } : undefined;
+    return (value && typeof value['pubDirectId'] === 'string') ? { 'pubDirectId': value['pubDirectId'] } : undefined;
   },
   /**
    * performs action to obtain id and return a value in the callback's response argument
@@ -44,134 +49,35 @@ export const pubDirectSubmodule = {
    * @returns {IdResponse|undefined}
    */
   getId(configParams, consentData, cacheIdObj) {
-    if (!configParams) {
+    if (!configParams || !configParams.publisherId) {
       utils.logError(`Pubdirect params missing.`);
       return undefined;
     }
-    // pubdirect ID logic
-    var siteIdValue = {
-      ID: '',
-      timestamp: null,
-      origin: '',
-      DoNotTrack: false,
-    };
-    var iframeCreated = false;
-    var incomingUserId = '';
-    var receiver = {};
-    var consent = true;
-    var doNotTrackUser = false;
+    const initSync = new Promise(function(resolve, reject) {
+      // pubdirect ID logic
+      var siteIdValue = {
+        ID: '',
+        timestamp: null,
+        origin: '',
+        DoNotTrack: false,
+        consentData: consentData,
+        cacheIdObj: cacheIdObj,
+      };
+      var iframeCreated = false;
+      var incomingUserId = '';
+      var receiver = {};
+      var consent = true;
+      var doNotTrackUser = false;
 
-    window.addEventListener('message', msgFiltering);
+      window.addEventListener('message', msgFiltering);
 
-    var userRandomId = {
-      ID: IDGenerator(),
-      timestamp: new Date().getTime(),
-      origin: '',
-      DoNotTrack: false,
-    };
+      var userRandomId = {
+        ID: IDGenerator(),
+        timestamp: new Date().getTime(),
+        origin: '',
+        DoNotTrack: false,
+      };
 
-    function IDGenerator() {
-      var array = new Uint32Array(8);
-      window.crypto.getRandomValues(array);
-      var str = '';
-      for (var i = 0; i < array.length; i++) {
-        str += (i < 2 || i > 5 ? '' : '-') + array[i].toString(16).slice(-4);
-      }
-      return str;
-    }
-
-    function ifLoaded() {
-      if (consent === true) {
-        if (doNotTrackUser === true) {
-          messageHub('doNotTrackUser', parseData);
-        }
-        // Get the window displayed in the iframe.
-        if (iframeCreated === false) {
-          messageHub('self', parseData);
-        } else if (iframeCreated === true) {
-          messageHub(incomingUserId, parseData);
-        }
-      } else {
-
-      }
-    }
-
-    function callModule() {
-      if (localStorage.getItem('pubId')) {
-        let callValue = localStorage.getItem('pubId');
-        callValue = JSON.parse(callValue).ID;
-        let callbackObj = {
-          'pubDirectId': callValue
-        };
-
-        const reply = function (callback) {
-          callback(callbackObj)
-        }
-
-        return { callback: reply }
-      }
-    }
-
-    function msgFiltering(e) {
-      iframeCreated = true;
-      incomingUserId = e;
-      if (e.origin !== 'https://pub.direct') {
-        return;
-      }
-      if (iframeCreated === true) {
-        var iframe = document.createElement('iframe');
-        iframe.style.display = 'none';
-        iframe.id = 'receiver';
-        iframe.src = 'https://pub.direct/sync.html';
-        document.body.appendChild(iframe);
-        iframe.onload = ifLoaded;
-      }
-    }
-
-    function messageHub(e, callback) {
-      if (e !== undefined && e.data !== 'clear' && e !== 'self' && e !== 'doNotTrackUser' && e.data.timestamp !== undefined) {
-        siteIdValue = JSON.parse((localStorage.getItem('pubId')));
-        if (siteIdValue.timestamp <= e.data.timestamp) {
-          sendMessage(siteIdValue);
-        } else {
-          localStorage.setItem('pubId', JSON.stringify(e.data));
-          callback(configParams.publisherId, configParams.segments);
-          callModule();
-        }
-      } else if (e === 'self') {
-        if (localStorage.getItem('pubId')) {
-          siteIdValue = JSON.parse(localStorage.getItem('pubId'));
-          siteIdValue.origin = window.location.href;
-          sendMessage(siteIdValue);
-        } else {
-          userRandomId.origin = window.location.href;
-          localStorage.setItem('pubId', JSON.stringify(userRandomId));
-          sendMessage(userRandomId);
-        }
-      } else if (e.data === 'clear') {
-        callback(configParams.publisherId, configParams.segments);
-        callModule();
-      } else if (e === 'doNotTrackUser') {
-        if (localStorage.getItem('pubId')) {
-          siteIdValue = JSON.parse(localStorage.getItem('pubId'));
-          siteIdValue.origin = window.location.href;
-          siteIdValue.DoNotTrack = true;
-          sendMessage(siteIdValue);
-        } else {
-          userRandomId.origin = window.location.href;
-          userRandomId.DoNotTrack = true;
-          localStorage.setItem('pubId', JSON.stringify(userRandomId));
-          sendMessage(userRandomId);
-        }
-      }
-    }
-
-    function sendMessage(e) {
-      receiver = document.getElementById('receiver');
-      receiver.contentWindow.postMessage(e, 'https://pub.direct/sync.html');
-    }
-
-    function init() {
       if (iframeCreated === false) {
         var iframe = document.createElement('iframe');
         iframe.style.display = 'none';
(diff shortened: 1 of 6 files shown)


pub-direct:
No sorry, I'm done. I won't touch it anymore until you have looked at it.

pub-direct:
hey @jaiminpanchal27, I see an error in the ci/circleci test. Even thought it is from the file userId_spec.js file that I have modified, I never touched that part in question. 

Would you have time to take a look at it?

bretg:
@pub-direct - @jaiminpanchal27 noticed that you're dropping an iframe into the page. This trips a major Prebid.js rule about incorporating external code. 

1) Please explain what you're doing here with sync.html. Why can't you use AJAX? No other userId module is doing anything like this.
2) If you insist this is necessary, there are many conditions:
     a) sync.html must be open sourced
     b) you must call for a specific version of sync.html from the PBJS code. e.g. sync-v1.html.
     c) you must agree to not change sync-v1.html after we review it. Changes MUST be done in a new version so that we review all changes.
     d) it must be disclosed prominently on your prebid.org documentation
     e) you must provide a mechanism for publishers to turn off the iframe. 
     f) your module should operate even if the pub has turned off the iframe
     g) your code MUST adhere to all privacy rules. e.g., if your iframe is setting cookies, you have to parse the GDPR consent string to make sure you have access to the device. And follow all PBJS controls (e.g. `pbjs.setConfig({ deviceAccess: false });`

We strongly recommend removing this call to sync.html because it's going to cause a lot of work for all of us.

pub-direct:
Hey @bretg ,

Thanks for the feedback from you and your team. I have seen a lot of your work on Prebid and I'm eager to explain my point of view.

Pub.Direct is created as a free cookieless identity solution to replace the 3rd party cookie. 

The way how it works is that instead of making an AJAX call and have a server assign a cookie ID and consequently read this cookie ID as an identifier for a user, we make a call to a website using a hidden iframe to achieve this goal, which in mind is more user friendly and secure because all we do is trading IDs that is already available on the user's browser and it ever calls a server. 

Here is the logic, user goes to a website with where the prebid pub.direct module is enabled for the first time. The pub.direct prebid module generates a random ID. This ID is then sent to the website that you have referenced "pub.direct/sync.html" using an hidden iframe and post-message. This website then checks its local storage to see if it already has an ID available, if so, it will compare the timestamp of the IDs and sent back the ID with the older timestamp also via post-message to complete the sync. The site with the pub.direct prebid module will then store this ID in its local storage. Now imaging we'd repeat this same procedure on multiple websites with the pub.direct module. 

Hopefully by now you can see why we can't just make an AJAX call. This solution was created as a hobby and I wanted to make it open source anyway, so I have no problem making the code public. You probably have seen a lot of examples and are probably way more experienced in adtech than me. However, I'd argue that this solution is not less privacy complaint than cookie sync, I think it's actually more compliant as it doesn't talk to a server. Every call we make is made out within a public domain. I'd be endless grateful if you can work with me on it. Please find answers to your specific requests below marked in bold. 

> @pub-direct - @jaiminpanchal27 noticed that you're dropping an iframe into the page. This trips a major Prebid.js rule about incorporating external code.
> 
> 1. Please explain what you're doing here with sync.html. Why can't you use AJAX? No other userId module is doing anything like this.
**Please see above**
> 2. If you insist this is necessary, there are many conditions:
>    a) sync.html must be open sourced
**No problem**
>    b) you must call for a specific version of sync.html from the PBJS code. e.g. sync-v1.html.
**No problem**
>    c) you must agree to not change sync-v1.html after we review it. Changes MUST be done in a new version so that we review all changes.
**No problem**
>    d) it must be disclosed prominently on your prebid.org documentation
**Okay, where exactly do you want me to create the disclaimer?**
>    e) you must provide a mechanism for publishers to turn off the iframe.
**No problem**
>    f) your module should operate even if the pub has turned off the iframe
**No problem**
>    g) your code MUST adhere to all privacy rules. e.g., if your iframe is setting cookies, you have to parse the GDPR consent string to make sure you have access to the device. And follow all PBJS controls (e.g. `pbjs.setConfig({ deviceAccess: false });`
**Okay...? However, I don't think you will trigger the user ID submodules, if there is no consent given, isn't that accurate? Referencing: [**](http://prebid.org/dev-docs/modules/userId.html), under section "User ID, GDPR, and Opt-Out".**
> 
> We strongly recommend removing this call to sync.html because it's going to cause a lot of work for all of us.
**Again, I hope by now you see why we have designed this module to be this way.**

bretg:
Thanks for the explanation @pub-direct, but I don't understand this part:

> This website then checks its local storage to see if it already has an ID available
 which contrasts with this:
> I think it's actually more compliant as it doesn't talk to a server. 

I think what you mean by "This website" is just the javascript code loaded from pub.direct attempts to set a first party cookie in pub.direct.

But based on this article https://www.simoahava.com/analytics/cookieless-tracking-cross-site-iframes/, I'm not sure cross-domain iframes can set a cookie without 3rd party cookie support:

```
However, there's a catch that's become more and more troubling with the increase of browsers’ tracking protection mechanisms: if the <iframe> loads content cross-site, the browser must allow cookie access in third-party context. In other words, if the browser blocks third-party cookies, the <iframe> will not be able to write the cookie
```

But I'm sure you've tested this, so perhaps this article is incorrect. Anyhow, since we're not going to allow this without seeing the code in some form anyhow, let us know how to obtain a copy. If you're not ready to share it broadly, can the reviewer email the address you have in this PR and get a copy?

Adding @jdwieland8282 as our resident User ID module expert.

pub-direct:
Hey @bretg and @jdwieland8282 ,

Happy to share the code. It's actually hosted on pub.direct/sync.html that you have already pointed out. Here is the JS file (https://pub.direct/js/ax-recipient.js).

The thing is that I don't use cookies. On pub.direct I just take the id that is sent to pub.direct via post-message and store it in its local storage. In case pub.direct already has a user Id, I'd compare the timestamp to see which one has existed longer, and pick that one.

Happy to make that part of code more official via a PR, but let me know if that makes sense and if you see any reason why it should not be allowed. 

Wenda

bretg:
We suspect this is exploiting a hole that will be closed in chrome with the coming of the cookie-apocalypse. Cross-domain localstorage is already prevented in Safari and Firefox by default. We're going to test this. If we find that Chrome also blocks Cross-domain localstorage when 3rd party cookies are turned off, we're likely to not accept this module.

If you have evidence that says this approach is going to be allowed, please post a link.

pub-direct:
@bretg I don't think your statement is factually correct!
**#1**
Even if Safari doesn't allow for cross domain tracking, it doesn't mean that they don't allow for storing a persistent ID in the local storage, which is already better than 3rd party cookies.

**#2**
Firefox allows it, so does IE. I added an example for you confirm below. 

**#3**
You can test your assuming about cookies, but I thought I communicated that I don't use cookies at all?!

In conclusion, I'm very open to discussing the facts but I think your statements are incorrect and therefore unfair. But then again you know the industry way better than I do, so if you could provide me evidence otherwise I'd interested in improving this product. Because, I believe I provided the evidence required to evaluate this case.

**How to test:** The two pages below are the same only separated by subdomains, which are considered separated instances. You can also host the 2 pages on completely different domains. It should not change its behavior (welcome to try). These 2 pages represent 2 separate websites. Load page 1 first, then hit refresh. Load page 2, then hit refresh. The reason why you need to hit refresh is because the DOM is already painted first time before it can grab the ID.

My assumption is that if you test it in Chrome, IE and Firefox you will see 2 identical IDs. If you should be confused with my instructions here is a video I've made: [Product Demo Video](https://www.youtube.com/watch?v=1ItxXXM2Ivk&feature=youtu.be)

**Page #1** 
https://cookielesspage1.s3.amazonaws.com/DemoPage1.html
**Page #2**
https://cookielesspage1.s3.amazonaws.com/DemoPage1.html

Hope that helps.

bretg:
Thanks for the details @pub-direct. We'll test and discuss.

bretg:
Status update: we have a couple of people trying this out. The initial conclusion is that this does indeed currently work in some browsers. I'd like to note that we certainly respect and appreciate the technical effort here. And personally, I appreciate your desire to help the industry with an existential struggle. However, we have a general sense that this effort is just re-arranging deck chairs on the Titanic.

It happens to work in Firefox for now because pub.direct is not in the disconnectme list (https://raw.githubusercontent.com/disconnectme/disconnect-tracking-protection/master/services.json).  We expect that if this were to be released, pub.direct will get added (like digitrust and id5 who are identity-only domains)

Overall, we suspect the approach likely violates the Privacy sandbox framework and therefore will stop working on Chrome soon after 3rd party cookies are shut down. See the TURTTLEDOVE explainer: https://github.com/michaelkleber/turtledove.  The general theme is that no unique identifier may be associated with an individual across domains.  

We haven't shut the door at this point, but remain skeptical. We'll discuss at Prebid.org meetings scheduled over the coming weeks and with Prebid legal counsel.

pub-direct:
Thanks @bretg,

At this point I don't clearly understand on which ground you would disallow this adapter to be rejected. Aside from the fact that you have confirmed it is working, even if it doesn't solve the cross domain issue, shouldn't that be a conversation between us and the clients? Because for instance, I think we have shown evidence that even in the Safari browser, where cross domain tracking is not allowed, our solution is more performant than the 3rd party cookie as it keeps the ID persistent. 

From an open source framework standpoint, which criteria does my module not fulfill to be allowed? I've already committed to fix those issues that you have mentioned. But I don't think it's fair to decide if my module will be accepted based on whether it can solve the user identify problem for the industry, that would be too big of a task. Additionally, I have seen many modules being accepted that simply doesn't work. So where do you draw the line for rejecting a module?

My offer is to work with you with the of goal of getting this solution accepted and I've also offered it to make it open source and this offer still stands. However, I think the decision of whether to accept the module should still be based on the guidelines that applies to everyone else too.

jdwieland8282:
A full list of my test results are attached.  In summery, I found that I was given a consistent uid on localhost.com and cookielesspage.com, but when a subdomain was added to localhost, ex. _localhost.prebid_ my uid became out of sync.  This behavior was observed on Chrome and Firefox and in private browsing and "normal" browser mode. Regarding Safari, my uids were never in sync, Safari presented me with a unique id on each of the 3 domains I observed. 

[Pubdirect-UAT.xlsx](https://github.com/prebid/Prebid.js/files/4576904/Pubdirect-UAT.xlsx)

pub-direct:
@jdwieland8282 Hey thanks for testing. I have expected the safari outcome, but if I'm not mistaken the IDs were still persistent per domain, correct? Which in my mind is better than 3rd party cookies, it is as if you were to go to all publishers to get 1st party cookie access. 

Regarding subdomains, I didn't come across this issue. I suspect it is localhost related. Have you used my 2 test pages? It should give you consistent results. The sites are mentioned below. 

Page #1
https://cookielesspage1.s3.amazonaws.com/DemoPage1.html
Page #2
https://cookielesspage2.s3-us-west-2.amazonaws.com/DemoPage2.html

bretg:
> clearly understand on which ground you would reject this adapter

Agreed that we would need solid ground for doing this, and that we would need to be apply such criteria equally across all user ID modules.

As background, we're getting picky about user ID modules because they're a controversial topic. It doesn't help the industry to have 11 or 12 options here. One possible outcome is that we allow this, but put requirements on documentation. i.e. not allow anyone to claim that this "solves the quest".  Your current documentation PR is quite modest, so that's not a problem.

pub-direct:
Thanks @bretg , 

I'll update the module to spec first, so we can discuss documentation and next steps.

jdwieland8282:
Hi @pub-direct , I will run my test again with your second domain later this week.  wrt Safari, I would expect refreshing the page would return the same id per domain, which really isn't any better or different than a first party cookie solution like you point out.

pub-direct:
@jdwieland8282 regarding Safari you are correct it is like a 1st party cookie. However, from an adtech provider standpoint, wouldn't enabling this feature be much easier than getting access to all 1st party cookie IDs from all publishers you work with?

bretg:
Met with a broader internal team, and we agreed this module can be released as long as it meets all the requirements laid out above.

a) sync.html must be open sourced
b) you must call for a specific version of sync.html from the PBJS code. e.g. sync-v1.html.
c) you must agree to not change sync-v1.html after we review it. Changes MUST be done in a new version so that we review all changes.
d) it must be disclosed prominently on your prebid.org documentation
e) you must provide a mechanism for publishers to turn off the iframe.
f) your module should operate even if the pub has turned off the iframe
g) your code MUST adhere to all privacy rules. e.g., if your iframe is setting cookies, you have to parse the GDPR consent string to make sure you have access to the device. And follow all PBJS controls (e.g. pbjs.setConfig({ deviceAccess: false });

smenzer:
Based on this conversation, it seems like the solution is essentially a 1st party user ID, since the way the IDs are being connected across domains in Safari doesn't work, in Firefox will be blocked when Disconnect.me adds pub.direct to their list, and in Chrome will almost certainly be blocked with their planned changes. 

If my understanding is correct, I have two questions about this id module. 

1. How is it different from the pubcommonid module that already exists?
2. Is there still a need for all of the iframe injection and messaging if it won't be useful to connect IDs? 3rd party cookies would accomplish the same thing and are equally available in browsers compared to the proposed iframe solution.

To be clear, I'm not suggesting that it is not considered for inclusion in Prebid.js, but simply trying to better understand the differences and limit the technical complexity needed to implement it.

pub-direct:
@smenzer Can you identify yourself and your involvement in the decision making progress regarding this adapter approval. If you are just curious about the product, I'd suggest you reach out to the company directly rather than commenting on this ticket. 

Just want to keep the process streamlined.

smenzer:
@pub-direct sure, I'm a member of Prebid.org and on the Prebid.js PMC.

pub-direct:
@smenzer Still not very clear on your decision making role for the inclusion of this adapter. In my mind, your initial comment doesn't reflect an accurate picture of the functionality of our adapter. I'd suggest you go through the comments above or sync with @bretg. We've published all the code for you to review.

From a prebid standpoint I don't see an issue to use iframes as it is client choice and we'll make the use of all of our technology as prominent as possible on prebid.org. 
From a competition standpoint,  I'd understand that someone would be defensive seeing our adapter if working for the competition as we are going to make it free to everyone, publishers as well as sync partners.
But I can assure you that we are just hobbyist trying to contribute to the industry as an open source project. So nothing to fear there as well.

Again, open to have a discussion if you are interested, but I don't think this is the place for it.

modules/userId/eids.js
```diff
@@ -28,6 +28,12 @@ const USER_IDS_CONFIG = {
     atype: 1
   },
 
+  // pubDirectId
+  'pubDirectId': {
+    source: 'pub.direct/sync.html',
```
smenzer, line 33: the `source` field is generally just the domain of the provider, not a full html path. I would recommend setting to just `pub.direct` for consistency

bretg:
> Can you identify yourself and your involvement in the decision making process regarding this adapter approval.

All parties are welcome to comment-on and ask-questions-about PRs. It's the nature of open source.

Thanks to all parties involved here for thorough vetting. The User ID space is important and worthy of such discussion.

Prebid.org will accept this module if it meets all 7 conditions stated above.

modules/pubDirectSystem.js
```diff
@@ -0,0 +1,190 @@
...
+            messageHub('doNotTrackUser', parseData);
+          }
+          // Get the window displayed in the iframe.
+          if (iframeCreated === false) {
+            messageHub('self', parseData);
+          } else if (iframeCreated === true) {
+            messageHub(incomingUserId, parseData);
+          }
+        } else {
+
```
yava-ua, line 107: pls insert some handling here. utils allows error loging, e.g.
```
utils.logError(error);
```

modules/pubDirectSystem.js
```diff
@@ -0,0 +1,190 @@
...
+        }
+      }
+
+      function msgFiltering(e) {
+        iframeCreated = true;
+        incomingUserId = e;
+        if (e.origin !== 'https://pub.direct') {
+          return;
+        }
+        if (iframeCreated === true) {
```
YerkovichM, line 117: iframeCreated is already true, you don't need to check it

modules/pubDirectSystem.js
```diff
@@ -0,0 +1,190 @@
...
+          var iframe = document.createElement('iframe');
+          iframe.style.display = 'none';
+          iframe.id = 'receiver';
+          iframe.src = 'https://pub.direct/sync.html';
+          document.body.appendChild(iframe);
+          iframe.onload = ifLoaded;
```
YerkovichM, line 123: Why are you loading once again?

modules/pubDirectSystem.js
```diff
@@ -0,0 +1,190 @@
...
+        }
+      }
+      function messageHub(e, callback) {
+        if (e !== undefined && e.data !== 'clear' && e !== 'self' && e !== 'doNotTrackUser' && e.data.timestamp !== undefined) {
```
YerkovichM, line 127: This condition is too complex. You can rich the same result by putting this case at the end of function.

pub-direct:
Thank you @bretg @smenzer @jdwieland8282 ,

We'll update the adapter soon. 

Wenda

YerkovichM:
@smenzer @jdwieland8282 I have left comments. I tested this module as well. I used 2nd demo page. Chrome 
Case 1 : open three one after another in list order bellow
Results:
cookielesspage2.com -> 5827a8f4-a02c-2a6c-9172-fe36f1e1888d
localhost:8080 -> 5827a8f4-a02c-2a6c-9172-fe36f1e1888d
www.localhost.prebid:8080 -> 5827a8f4-a02c-2a6c-9172-fe36f1e1888d
But results are not consistent. In some cases there are issues with propagating id between localhost:8080 and www.localhost.prebid:8080
In summery, I would say it is very inconsistent in its results. When testing I got more unexpected results than right one. And I suspect there lots of issues in code, like unnecessary looping of iframe requests.

pub-direct:
@jdwieland8282 thanks for the feedback. Maybe let's wait until I update to the newest version and let's chat then.

pub-direct:
@YerkovichM I don't think you understand the concept of how it works.

PR #5175, opened 2020-04-28
Marsmedia - Add onBidWon function

vladi-mmg:
## Type of change

- [X] Feature

## Description of change
Add onBidWon function to the adapter


vladi-mmg:
Hello,
Yes I am trigger two pixels, one for NURL and the other for our analytics.

bretg:
@vladi-mmg - heads up. Prebid is drafting rules about adapter behavior. There's a draft rule on the table requiring disclosure when a bid adapter "phones home" for analytics, and provide publisher control over whether the reporting takes place. Assuming that rule stays in place, it will be enforced in a future release (5.0 or 6.0). In the meantime, we'll accept the two pixels.

PR #5178, opened 2020-04-28
adding check if gdpr module ran enforcment logic

robertrmartinez:
## Type of change
- [X] Bugfix

## Description of change
Currently the UserId Module does it's own check for consent before syncing.

However, it does not take into account the gdprEnforcment `rules` and `exceptions`.

So if consentData is retrieved, and it includes an exception for purpose or specific userId vendors, then the userId module does not know about them and it actually will not run the userId syncs at all!

So this PR simply passes back a flag to the userId module indicating that the `gdprEnforcement` module has done it's enforcement and thus the userId module can just skip it.

Basically the enforcement module will remove and subModules which are not allowed, so the userId module does not need to do the check itself.


robertrmartinez:
Yeah I will do this! Thanks.

robertrmartinez:
Ok I added assertions for the `gdprEnforcement` module

But the userId module does not have any gdpr tests and it is a little tangled to write a simple test for this change.

So I will write a couple general gdpr tests for the userId module later this week when I have some more free time. (I plan to include those tests in this PR)

PR #5182, opened 2020-04-29
AdxPremium Analytics - bug fixes, handling timeout, etc.

adxpremium:
## Type of change

- [x] Bugfix
- [x] Feature

## Description of change
Bug fixes and improved event handling.

### DIFF
--- a/modules/adxpremiumAnalyticsAdapter.js
+++ b/modules/adxpremiumAnalyticsAdapter.js
@@ -7,7 +7,7 @@ import * as utils from '../src/utils.js';
 const analyticsType = 'endpoint';
 const defaultUrl = 'https://adxpremium.services/graphql';
 
-let reqCountry = reqCountry || null;
+let reqCountry = window.reqCountry || null;
 
 // Events needed
 const {
--- a/modules/adxpremiumAnalyticsAdapter.js
+++ b/modules/adxpremiumAnalyticsAdapter.js
@@ -1,11 +1,11 @@
-import { ajax } from '../src/ajax.js';
-import adapter from '../src/AnalyticsAdapter.js';
-import adapterManager from '../src/adapterManager.js';
+import { ajax } from '../src/ajax';
+import adapter from '../src/AnalyticsAdapter';
+import adapterManager from '../src/adapterManager';
 import CONSTANTS from '../src/constants.json';
-import * as utils from '../src/utils.js';
+import * as utils from '../src/utils';
 
 const analyticsType = 'endpoint';
-const url = 'https://adxpremium.services/graphql';
+const defaultUrl = 'https://adxpremium.services/graphql';
 
 // Events needed
 const {
@@ -19,6 +19,11 @@ const {
   }
 } = CONSTANTS;
 
+let timeoutBased = false;
+let requestSent = false;
+let requestDelivered = false;
+let elementIds = [];
+
 // Memory objects
 let completeObject = {
   publisher_id: null,
@@ -26,11 +31,14 @@ let completeObject = {
   referer: null,
   screen_resolution: window.screen.width + 'x' + window.screen.height,
   device_type: null,
-  geo: null,
+  geo: reqCountry || null,
   events: []
 };
 
-let adxpremiumAnalyticsAdapter = Object.assign(adapter({ url, analyticsType }), {
+// Upgraded object
+let upgradedObject = null;
+
+let adxpremiumAnalyticsAdapter = Object.assign(adapter({ defaultUrl, analyticsType }), {
   track({ eventType, args }) {
     switch (eventType) {
       case AUCTION_INIT:
@@ -49,7 +57,7 @@ let adxpremiumAnalyticsAdapter = Object.assign(adapter({ url, analyticsType }), 
         bidTimeout(args);
         break;
       case AUCTION_END:
-        setTimeout(function () { sendEvent(completeObject) }, 3100);
+        auctionEnd(args);
         break;
       default:
         break;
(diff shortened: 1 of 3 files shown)


modules/adxpremiumAnalyticsAdapter.js
```diff
@@ -5,7 +5,9 @@ import CONSTANTS from '../src/constants.json';
 import * as utils from '../src/utils.js';
 
 const analyticsType = 'endpoint';
-const url = 'https://adxpremium.services/graphql';
+const defaultUrl = 'https://adxpremium.services/graphql';
+
+let reqCountry = reqCountry || null;
```
aleksatr, line 10: hey @adxpremium , there's a lint error here:

```
'reqCountry' was used before it was defined.eslint(no-use-before-define)
```
Is `reqCountry` expected to be on the window object?

adxpremium: Hey @aleksatr, yes, reqCountry is expected to be on the window object, if not then it will be just null and that's fine.

modules/adxpremiumAnalyticsAdapter.js
```diff
@@ -19,18 +21,26 @@ const {
   }
 } = CONSTANTS;
 
+let timeoutBased = false;
```
aleksatr, line 24: is `timeoutBased` somehow configurable? Looking at this adapter I see that it's always false?

modules/adxpremiumAnalyticsAdapter.js
aleksatr: Ok, in that case, I guess you can do something like:

```
let reqCountry = window.reqCountry || null;
```

to mitigate the lint error.

aleksatr:
@adxpremium , I don't see unit tests for your analytics adapter, what is the reason for that?
I would expect them to be under `adxpremiumAnalyticsAdapter_spec.js`, as instructed by `Step 3: Add unit tests` of: http://prebid.org/dev-docs/integrate-with-the-prebid-analytics-api.html

modules/adxpremiumAnalyticsAdapter.js
```diff
@@ -19,18 +21,26 @@ const {
...
+let requestSent = false;
```
aleksatr, line 25: `requestSent` (probably `requestDelivered` also) are "global variables" for this adapter. From what I see this variable is set to `true` by `clearSlot()`, which is triggered by gpt's slotRenderEnded event. As this variable is global, that means, when a publisher is running multiple slots on the page, slotRenderEnded of one slot would set this variable to `true` and `sendEvent(completeObject);` would only trigger for that one slot, skipping all the rest. Am I missing something, or is this intentional?

modules/adxpremiumAnalyticsAdapter.js
adxpremium: Ok thanks, I will update in a moment.

modules/adxpremiumAnalyticsAdapter.js
adxpremium: Currently it is not configurable but we do have plans to make it configurable and that's why we left it in a separate variable.

modules/adxpremiumAnalyticsAdapter.js
```diff
@@ -125,16 +210,36 @@ function deviceType() {
...
+
 function sendEvent(completeObject) {
+  requestDelivered = true;
   try {
     let responseEvents = btoa(JSON.stringify(completeObject));
     let mutation = `mutation {createEvent(input: {event: {eventData: "${responseEvents}"}}) {event {createTime } } }`;
     let dataToSend = JSON.stringify({ query: mutation });
-    ajax(url, function () { utils.logInfo(Date.now() + ' Sending event to adxpremium server.') }, dataToSend, {
+    let ajaxEndpoint = defaultUrl;
+    if (adxpremiumAnalyticsAdapter.initOptions.sid) {
```
aleksatr, line 235: is `sid` some adapter parameter? Can you mention it in the `adxpremiumAnalyticsAdapter.md` so we have it documented somewhere?

modules/adxpremiumAnalyticsAdapter.js
adxpremium: This is used just for the timing purposes of the send event and it becomes false again when the new auctionInit event occurs. I will take another look at it, for now it works fine.

adxpremium:
Global reqCountry is fixed now. We will add unit testing as well.

modules/adxpremiumAnalyticsAdapter.js
adxpremium: Yes, sid = Assigned Server ID since we have multiple analytics servers. Will be mentioned in the markdown in a moment.

adxpremium: Done.

aleksatr:
> @aleksatr Global reqCountry is fixed now. We will add unit testing as well.

hey @adxpremium , are you still working on tests, I don't see them pushed?

adxpremium:
We have pushed unit tests, circleci throws an error, my log looks like this:

START:
ℹ ｢wdm｣: Compiling...
ℹ ｢wdm｣: 
ℹ ｢wdm｣: Compiled successfully.
18 05 2020 14:17:26.886:INFO [karma-server]: Karma v4.4.1 server started at http://0.0.0.0:9876/
18 05 2020 14:17:26.888:INFO [launcher]: Launching browsers ChromeHeadless with concurrency 5
18 05 2020 14:17:26.922:INFO [launcher]: Starting browser ChromeHeadless
18 05 2020 14:17:28.103:INFO [HeadlessChrome 81.0.4044 (Mac OS X 10.15.3)]: Connected on socket jjwbUCIHZG_NEH_FAAAA with id 18745409

Finished in 7.081 secs / 3.867 secs @ 14:17:37 GMT+0200 (Central European Summer Time)

SUMMARY:
✔ 5498 tests completed
ℹ 3 tests skipped
[14:17:37] Finished 'test' after 54 s
[14:17:37] Finished 'test' after 1.37 min

aleksatr:
hey @adxpremium, test is indeed failing, check circle CI output. Same outcome when I ran `gulp test` on my local, check screenshot: 

image

adxpremium:
Hi @aleksatr I am not able to reproduce the same error. Here is the screenshot from my local machine with the latest prebid.js clone + analytics adapter + unit tests

Screen Shot 2020-05-28 at 1 16 25 PM

aleksatr:
@adxpremium judging by your output (mainly number of tests), it looks like you're running `gulp test` on master branch, not on the one that contains your changes. To reproduce try this:

```
cd <workspace_dir>
git clone https://github.com/prebid/Prebid.js.git
cd Prebid.js
git fetch origin pull/5182/head:pr-5182
git checkout pr-5182

npm install
gulp test
```

adxpremium:
Hi @aleksatr I did exactly what you told me to do and the output is pretty much the same. The only difference is the number of tests, hmm ?


Screen Shot 2020-05-29 at 3 51 30 PM

aleksatr:
hey @adxpremium , I decoded your b64 expected results and compared them with the actual results when test are running: 

image

As you can see, the difference is in the `screen_resolution` property, you should not assume resolution in your test data. Either you omit this property from your tests or you read the actual screen resolution when you're creating `expectedAfterBid`, `expectedAfterTimeout`, ... (don't hardcode it).

adxpremium:
> hey @adxpremium , I decoded your b64 expected results and compared them with the actual results when test are running:
> 
> image
> 
> As you can see, the difference is in the `screen_resolution` property, you should not assume resolution in your test data. Either you omit this property from your tests or you read the actual screen resolution when you're creating `expectedAfterBid`, `expectedAfterTimeout`, ... (don't hardcode it).

Thanks. We just pushed the updated version. Now it passes all the checks.

aleksatr:
LGMT

PR #5185, opened 2020-04-30
TCF v2.0 Purpose 2 Enforcement

Fawke:
Type of change
- [X] Feature

## Description of change

This PR introduces TCF v2.0 Purpose 2 (basicAds) enforcement feature to Prebid.js. This builds upon this PR, https://github.com/prebid/Prebid.js/pull/5018

The basic idea is, if consent is not present for a bidder, the bidder won't send a request to their endpoint. A console warning message will be logged, and an event will be emitted for analytics adapters to consume.


modules/gdprEnforcement.js
```diff
@@ -31,18 +38,21 @@ function getGvlid() {
...
+ * @param {number} purpose - Defines which purpose (1, 2, 4, 7) is under check
+ * @param {string=} currentModule - Bidder code of the current module
+ * @param {number=} gvlid - GVL ID for the module
  * @returns {boolean}
  */
-function validateRules(rule, consentData, currentModule, gvlid) {
+function validateRules(rule, consentData, purpose, currentModule, gvlid) {
   let isAllowed = false;
   if (!rule.vendorExceptions) rule.vendorExceptions = [];
-  if (rule.enforcePurpose && rule.enforceVendor) {
```
robertrmartinez, line 41: I was just about to make a code change to fix this issue, where if one of these was set to false it never got in the checks!

Looks like this is the way to go instead of first checking for existence of the keys themselves.

Thanks!

modules/gdprEnforcement.js
```diff
@@ -191,11 +198,48 @@ export function userIdHook(fn, submodules, consentData) {
...
+      const disabledBidders = [];
+      adUnits.forEach(adUnit => {
+        adUnit.bids = adUnit.bids.filter(bid => {
+          const currBidder = bid.bidder;
+          const gvlId = getGvlid(currBidder);
+          if (includes(disabledBidders, currBidder)) return false;
+          const isAllowed = validateRules(purpose2Rule, consentData, 2, currBidder, gvlId);
+          if (!isAllowed) {
+            utils.logWarn(`User blocked bidder: ${currBidder}. No bid request will be sent to their endpoint.`);
+            // Emit an event for the Analytics adapters to listen
```
jaiminpanchal27, line 220: This comment is unnecessary, below line of code is self explainatory

modules/gdprEnforcement.js
```diff
@@ -11,18 +11,25 @@ import includes from 'core-js/library/fn/array/includes.js';
 import { registerSyncInner } from '../src/adapters/bidderFactory.js';
 import { getHook } from '../src/hook.js';
 import { validateStorageEnforcement } from '../src/storageManager.js';
+import events from '../src/events.js';
+import { EVENTS } from '../src/constants.json';
 
-const purpose1 = 'storage';
+const PURPOSE_1 = 'storage';
```
jaiminpanchal27, line 17: Suggestion:
I think it would be better if make this an object and keep the id 1 and name storage together.
```
const PURPOSE_1 = {
 id: 1,
 name: 'storage',
}

modules/gdprEnforcement.js
Fawke: Removed comment.

modules/gdprEnforcement.js
Fawke: I felt the same way that I'll be better to store the values as an object. I've made the change. Please have a look.

Fawke:
> Overall looks good, couple of minor points.
> Can you also test for prebid server integration. Here is the page https://github.com/prebid/Prebid.js/blob/master/integrationExamples/gpt/prebidServer_example.html

I've tested with prebidServer, seems to be working.

bretg:
We're discussing the defaults for Purpose 2... hold off a bit on merging

bretg:
We need to update this PR as noted in #5194 

- Prebid.js will block head bidding auctions if there's no legal basis test for Purpose 2, including legitmate interest established
- Prebid.js will default to enforcing Purposes 1 and 2 if those rules aren't defined.

bretg:
Removed this ticket from "on hold". Let me know if you'd like to walk through the changes in the document.

PR #5186, opened 2020-04-30
Add videoCacheKey on the bid object only if the cache url is not set

pratik-synacor:
## Type of change

- [x] Bugfix

## Description of change

Set the videoCacheKey on the bid object manually only if the cache url is not set in the config. If the cache url is set, then the Prebid logic takes care of setting this property from cache.

demand-eng@synacor.com


pratik-synacor:
@robertrmartinez Added a test case.

PR #5197, opened 2020-05-05
parrableIdSystem: Consume US Privacy consent data

icflournoy:
## Type of change

- [x] Feature
- [x] Other - additional test coverage

## Description of change

This change enables the Parrable ID system to retrieve the uspString obtained by Prebid's consentManagementUsp module and pass the consent data to Parrable.

We pass the consent string as a visible query paramter `us_privacy=` which aligns with the USP specification.

My understanding is that the adapterManager.uspDataHandler should have the USP data by the time the userId module runs (requestBids hook priority level 40). The consentManagementUsp module is at requestBids hook priority level 50.

This also adds additional test coverage focused specifically on the parrableIdSystem unit. Previously our tests required the use of the userId system as a whole. I think there is value in this small integration tests because it ensures that our ID makes it into the bids object.

I've started adding more lightweight unit tests inside of our module's spec so that it's easy to extend to cover new features in our module, and easier to understand what the tests themselves are doing without reading core prebid modules.


test/spec/modules/parrableIdSystem_spec.js
```diff
@@ -1,15 +1,19 @@
...
-import {config} from 'src/config.js';
+import { config } from 'src/config.js';
 import * as utils from 'src/utils.js';
+import { newStorageManager } from 'src/storageManager.js';
+import { getRefererInfo } from 'src/refererDetection.js';
+import { uspDataHandler } from 'src/adapterManager.js';
 import { init, requestBidsHook, setSubmoduleRegistry } from 'modules/userId/index.js';
 import { parrableIdSubmodule } from 'modules/parrableIdSystem.js';
-import { newStorageManager } from 'src/storageManager.js';
+import { server } from 'test/mocks/xhr';
```
FilipStamenkovic, line 9: Can you here extension `.js`?  
I see all other test files that are importing `server` have `.js` extension for `xhr`.

PR #5200, opened 2020-05-06
Freewheel add dealId property in bidResponse

xwang202:
## Type of change
- [ ] Feature

## Description of change
get dealId value from the returned xml response and add dealId property in bidResponse


xwang202:
Hi @musikele , 

Can you please help to review the code changes?

Thanks,
Xuan

PR #5201, opened 2020-05-07
Gdpr vendor exceptions

robertrmartinez:
## Type of change
- [X] Miscommunication

## Description of change
According to the PRD https://docs.google.com/document/d/1fBRaodKifv1pYsWY3ia-9K96VHUjd8kKvxZlOsozm8E/edit# the  `gdpr.rules[].vendorExceptions` field is supposed to be a simple global setting for a vendor always being allowed regardless,

I believe in one of the first iterations of the spec it was supposed to do what it currently does with the "if enforceVendor is false then vendorExceptions means we should enforce these vendors"

But the latest revision of the spec indicated if a bidder / module is in vendorException, then they are allowed no matter what.

### DIFF
--- a/modules/gdprEnforcement.js
+++ b/modules/gdprEnforcement.js
@@ -36,48 +36,15 @@ function getGvlid() {
  * @returns {boolean}
  */
 function validateRules(rule, consentData, currentModule, gvlid) {
-  let isAllowed = false;
-  if (!rule.vendorExceptions) rule.vendorExceptions = [];
-  if (rule.enforcePurpose && rule.enforceVendor) {
-    if (
-      includes(rule.vendorExceptions, currentModule) ||
-      (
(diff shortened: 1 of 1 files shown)


Fawke:
@robertrmartinez @bretg @jaiminpanchal27 

I had one confusion, according to the basic enforcement rules listed on [prebid.org](http://prebid.org/dev-docs/modules/gdprEnforcement.html#basic-enforcement), we should check for `legitimateInterests` as well. But I think, currently we are only checking for `purpose` consent and `vendor` consent.

Are `legitimateInterests` meant only for Prebid Server?

bretg:
> Are legitimateInterests checks meant only for Prebid Server?

Yes, Legitimate Interest is only for Prebid Server. This is because A) Prebid.js doesn't have access to the Global Vendors List and B) it's a pretty complicated thing to be doing in everyone's browser.

PR #5208, opened 2020-05-08
Stabilize Circle CI Build Job

Fawke:
## Type of change
- [X] Fix Unit Tests
- [X] Stabilise CI Build Job

## Description of change

Circle CI build job, which runs after every commit to the Prebid.js repo, have been failing intermittently due to some error in `userId_spec` file. Attempts were made earlier to fix this, https://github.com/prebid/Prebid.js/pull/4982, but it didn't work.

#### Instances of recent failures.
[Failure#1](https://circleci.com/gh/prebid/Prebid.js/7829?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-build-link)

[Failure#2](https://circleci.com/gh/prebid/Prebid.js/7698?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-build-link)

[Failure#3](https://circleci.com/gh/prebid/Prebid.js/7694?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-build-link)

This PR finally fixes this error.

The PR is also part of a broader effort to stabilise the circleci build job. I've prepared a list of other files and adapters which sometimes fail unexpectedly. Though the frequency of errors listed below is not as high as the one that this PR fixes, but still, they can occur sometimes.

- Yieldone Adapter. https://circleci.com/gh/prebid/Prebid.js/7121
- SovrnBidAdapter https://circleci.com/gh/prebid/Prebid.js/7095?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-build-link 
- ExpiringQueue timeout problem. 
-  Sonobi Prebid Analytic Error -  https://circleci.com/gh/prebid/Prebid.js/7552?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-build-link
- BeachFrontAdapter - Timeout of 2000ms exceeded
- Consent management - Timeout of 2000ms exceeded
- Safari Browser took too much time to execute - https://circleci.com/gh/prebid/Prebid.js/7453
- Edge Browser timed out - https://automate.browserstack.com/builds/7444ca0d50d76d93502c1def7ad3809c7df415f8
- YieldonePrebidAnalytic, Bidder created by new bidder: https://circleci.com/gh/prebid/Prebid.js/7561?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-build-link

#### Related PR

https://github.com/prebid/Prebid.js/pull/4840

PR #5219, opened 2020-05-12
parrableIdSystem: Read legacy ID and optout cookies and migrate to new cookie storage implementation

icflournoy:
## Type of change

## Description of change

Parrable is moving to a "compound cookie" - one with both our ID value and our optout statuses encoded within.

At a high level this PR enables the Prebid ParrableIdSystem to read and write to this new compound cookie, as well as migrate values from our legacy cookies, whenever they are encountered. We have chosen to read cookies and write the compound cookie within our module (not relying on the UserId module to manage storage) so that we can maintain a consistent picture of the cookie state as it relates to our module.

This implementation is consistent with the non-Prebid integrations that Parrable has. In other words both Prebid and Non-Prebid enabled publishers will be working with our compound cookie the same way so it is imperative that the cookie name is consistent across both implementations.

Seperately this PR removes feature that we added some time ago to support UserId systems "that must always refresh stored ID". At that time we felt that it was the best way for our ID system (which performs ID maintenance on every impression) to be treated in a manner consistent with the other UserId submodule types.

See https://github.com/prebid/Prebid.js/pull/4325 for the work as it was integrated last November.

With the transition to managing our own storage we can do away with this `SUBMODULES_THAT_ALWAYS_REFRESH_ID` functionality. Our module then becomes the "third" type of UserId submodule, one that does not take a storage config, or use a fixed value.
We preserve compatibility (providing a stored ID value to bids) by returning both a callback and id value from the `getId` interface.

We have also added significant unit test coverage for migrating legacy cookies, reading existing cookies, and synchronizing the stored cookie values via the Xhr to our systems.

- https://github.com/prebid/prebid.github.io/pull/1973


icflournoy:
The test that ensures that legacy cookies are removed is failing in IE11, Safari 12, and Safari 13.

Do you have guidance on what I should be doing to expire/remove cookies in those browsers? I am using the StorageManager setCookie method with a past date. In other browsers this removes the cookie and the storage manager returns null. In the above browsers they still seem to retain the cookie and pickup the `''` value I am setting when expiring the cookie.

icflournoy:
Good news - using a properly formatted past date string will expire the cookies in IE11/Safari12 and 13.

The last test failure is another instance of the userId_spec test failing https://github.com/prebid/Prebid.js/pull/5208

icflournoy:
Thanks for the review @jsnellbaker !
Happy to answer any questions for you @idettman 

I have two additional features for our ID system that I am holding off opening up a PR for until this is merged, as to avoid confusion and keep the branches easy to review.

icflournoy:
Thanks, @idettman !

Issue #5207, opened 2020-05-07
IX adapter: not encoding bid request url


jas-singh:
## Type of issue
Bug

## Description
The IX adapter (and potentially other adapters) are no longer working from release 3.16 onwards due to a Prebid framework update. We've tracked down the issue to an update made here #5092 cause of which the query string parameters in the GET request are no longer being encoded which results in malformed requests being made.

## Steps to reproduce
* Configure the IX bidder with a valid configuration https://github.com/prebid/Prebid.js/blob/master/modules/ixBidAdapter.md on a test page
* Check the bid request call 

### Expected bid request made
https://as-sec.casalemedia.com/cygnus?s=430348&v=7.2&r=%7B%22id%22%3A%2252a75425cf918b%22%2C%22imp%22%3A%5B%7B%22id%22%3A%2265b3578b6556a5%22%2C%22banner%22%3A%7B%22w%22%3A300%2C%22h%22%3A250%2C%22topframe%22%3A1%7D%2C%22ext%22%3A%7B%22siteID%22%3A%22430348%22%2C%22sid%22%3A%22300x250%22%7D%7D%5D%2C%22site%22%3A%7B%22ref%22%3A%22https%3A%2F%2Fhideout.co%2F%22%2C%22page%22%3A%22https%3A%2F%2Fhideout.co%2Fwatch.php%3Fv%3D543952%26p%3D5553%22%7D%2C%22ext%22%3A%7B%22source%22%3A%22prebid%22%7D%2C%22regs%22%3A%7B%22ext%22%3A%7B%22gdpr%22%3A0%7D%7D%2C%22user%22%3A%7B%22ext%22%3A%7B%22consent%22%3A%22%22%7D%7D%7D&ac=j&sd=1&

### Actual bid request made
[...]

PR #5230, opened 2020-05-13
Roll back breaking change to utils.parseQueryStringParameters

kizzard:
## Type of change

- [x] Bugfix

## Description of change

A breaking change was made recently to `utils.parseQueryStringParameters` that removes its ability to encode the parameter values. This was not caught originally because the unit test was not sufficent and included no encodeable characters.

This commit rolls back the change to `utils.parseQueryStringParameters` and improves the test to catch this problem in future.

This is a fix for #5207 

## Other information
cc @jas-singh


kizzard:
Fixing my test flub....

PR #5255, opened 2020-05-16
Consent reset

olafbuitelaar:
## Type of change
- [x] Bugfix

## Description of change
When resetConsentData is invoked it doesn't reset the hookConfig.hasExisted value, which causes up-pun re-entry of the module, that the consentstring isn't resolved for the secondtime. 
Also the resetConsentData  set's userCMP variable to undefined. Which causes the requestBidHook to report a none support framework is chosen. For now i choose to add a check in the requestHook which sets this value to DEFAULT_CMP, but another option might be to reset it to DEFAULT_CMP instead of undefined in the resetConsentData function

### DIFF
--- a/modules/consentManagement.js
+++ b/modules/consentManagement.js
@@ -246,6 +246,9 @@ let resetHook = function() {};
  * @param {function} fn required; The next function in the chain, used by hook.js
  */
 export function requestBidsHook(fn, reqBidsConfigObj) {
+  if (!utils.isStr(userCMP)) {
+    userCMP = DEFAULT_CMP;
+  }
   // preserves all module related variables for the current auction instance (used primiarily for concurrent auctions)
   const hookConfig = {
     context: this,
--- a/modules/consentManagement.js
+++ b/modules/consentManagement.js
@@ -236,6 +236,7 @@ function lookupIabConsent(cmpSuccess, cmpError, hookConfig) {
   }
 }
 
+let resetHook = function() {};
 /**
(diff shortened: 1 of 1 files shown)


jsnellbaker:
@olafbuitelaar Can you elaborate more on why this PR is needed?  The `resetConsentData` funciton is meant to be used for the unit tests only, not for external use.  

Are you trying to setup some type of test and having trouble?

olafbuitelaar:
@jsnellbaker this is more a theoretical bug, which you might encounter for example in e2e tests, resetting the consentModule. I think the resetConsentData should also revert the module in an uninitialised state, to prevent unexpected behaviour after a reset.

jsnellbaker:
The `hookConfig` object is initialized each time we start a sequence of going through the module - which is triggered by the requestBids function (ie a new auction).  In the normal sequence if there was data available from a previous request (ie refresh), then the data is pulled from memory to save the need of calling the CMP again (and avoiding any delays for the auction) but the inner values like the hookConfig still get set properly on their own so we can do a quick exit from the module.

For any e2e tests we will do (which we have at least one set up in an upcoming PR), this process shouldn't run into issues since it's mimicking the above behavior (ie a normal page-load with a CMP).  Every run has its own clean set of data to do the work.  

For any unit tests (which is where the resetConsentData function is only used), they should be starting the test by calling the `requestBidsHook` function to kick off the process - which again initializations everything for the hookConfig to the ideal state for that run.

I'm still not sure if these changes are really needed.  Please let me know your thoughts on the above.  Thanks.

PR #5257, opened 2020-05-18
Onevideo Adaptor -Hp param support

DeepthiNeeladri:
## Type of change

- [ ] Bugfix
- [X ] Feature
- [ ] New bidder adapter   
- [ ] Build related changes
- [ ] CI related changes
- [ ] Other

## Description of change
Need to update current schain on the VSSP to include the parameter "hp" that we are currently not sending. HP is defined below:
Indicates whether this node will be involved in the flow of payment for the inventory. When set to 1, the advertising system in the asi field pays the seller in the sid field, who is responsible for paying the previous node in the chain. When set to 0, this node is not involved in the flow of payment for the inventory. For version 1.0 of SupplyChain, this property should always be 1. It is explicitly required to be included as it is expected that future versions of the specification will introduce non-payment handling nodes. Implementers should ensure that they support this field and propagate it onwards when constructing SupplyChain objects in bid requests sent to a downstream advertising system.

```
{
  bidder: '<bidder name>',
  params: {
    // ...
  }
}
```


DeepthiNeeladri:
Hi @FilipStamenkovic ,
I am adding test case so I kept in WIP, now it is done so removing it.
Thanks,
Deepthi

PR #5259, opened 2020-05-19
New adapter - videofy

vladi-mmg:
## Type of change
- [X] New bidder adapter   

## Description of change
New adapter Videofy

```
{
  code: 'video1',
  mediaTypes: {
    video: {
      playerSize: [[640, 480]],
      context: 'outstream'
    },
  },
  bids: [{
    bidder: 'videofy',
    params: {
      AV_PUBLISHERID: '55b78633181f4603178b4568',
      AV_CHANNELID: '5d19dfca4b6236688c0a2fc4'
    }
  }]
}
```

- A link to a PR on the docs repo at https://github.com/prebid/prebid.github.io/pull/1987


msm0504:
@vladi-mmg I happened to notice that this code is nearly identical to the existing `aniviewBidAdapter`. Does this need to be a separate adapter or can it be an alias of `aniview`?

http://prebid.org/dev-docs/publisher-api-reference.html#module_pbjs.aliasBidder

vladi-mmg:
@msm0504  We are using Aniview as an ad-server and player under our own domain (on a white label agreement). For building an outstream adapter for our own client we used their code as a reference but we need it to be under our name

msm0504:
@vladi-mmg You can be under your own name. The `videofy` bidder code would be added as an alias in the `aniviewBidAdapter`. The adapter code could then be modified to set the domains based on the bidder code passed in on a given request. The existing adapter doesn't need to copied/pasted into a new adapter.

There are other adapters that have similar logic based on aliases. Here's a link to an example: https://github.com/prebid/Prebid.js/blob/master/modules/aolBidAdapter.js#L66

vladi-mmg:
Thanks so much for the response, and accept my apologies in advance for trying to explain why alias's solution is a little less suited to our needs. 
 
We have a really big opportunity with a major publisher that agreed to start with our whiltelabel Aniview player but demands support in additional ad server and player. Using Aniview as an alias will limit us to their product. So we thought to use aniview code just to start our way with a basic working adapter and to upgrade it with additional capabilities on our next adapter version. In this way the adapter won't be similar at all in just a few weeks from now.

vladi-mmg:
Sorry for the urgency but I wanted to ask if you got a chance to see my last reply? 
As I mentioned it's holding us back on an important business opportunity
so i would really appreciate your help here (:

msm0504:
Approved and merged. This adapter will be in the next Prebid release - v3.21.0

Issue #5211, opened 2020-05-11
gulp test --file doesn't work


smenzer:
## Type of issue
bug

## Description
I'm trying to use `gulp test --file 'XXX' to make the testing process more efficient (rather than running all +5k tests every time I make a small change to my test file). Depending on the file I select, I get different error messages, but it always fails.

## Steps to reproduce
- `gulp test --file 'test/spec/api_spec.js'` produces 
```
Uncaught ReferenceError: require is not defined at test/spec/api_spec.js:1:14
```

- `gulp test --file 'test/spec/adUnits_spec.js'` produces 
```
Uncaught ReferenceError: require is not defined at test/spec/adUnits_spec.js:2:16
```

- `gulp test --file 'test/spec/modules/userId_spec.js'` produces 
```
Uncaught SyntaxError: Cannot use import statement outside a module at test/spec/modules/userId_spec.js:1:1
```

- `gulp test --file 'test/spec/modules/criteoIdSystem_spec.js'` produces 
```
Uncaught SyntaxError: Cannot use import statement outside a module at test/spec/modules/criteoIdSystem_spec.js:1:1
```

### Expected results
I would expect that only the tests contained in the specified test file would be run AND that they would work just like when running `gulp test` (which does work fine)

### Actual results
Errors (as shown above)

## Platform details
```
$ npm version
{ 'prebid.js': '3.19.0-pre',
  npm: '6.4.1',
  ares: '1.10.1-DEV',
  cldr: '32.0',
  http_parser: '2.8.0',
  icu: '60.1',
  modules: '57',
  napi: '4',
  nghttp2: '1.39.2',
  node: '8.16.2',
  openssl: '1.0.2s',
  tz: '2017c',
  unicode: '10.0',
  uv: '1.23.2',
  v8: '6.2.414.78',
  zlib: '1.2.11' }
```

## Other information
I've also noticed other minor issues with `gulp` that lead me to believe we need a review and cleanup of the entire `gulpfile.js`. For example, I also get module errors when running `gulp test --watch` that never appear when doing individual `gulp test` runs. Happy to provide more details if necessary, but thought I'd start with the most annoying issue first and propose an overhaul before creating separate issues.

Let me know if any more information is needed here. Thanks!

sbrosinski:
[...]

PR #5268, opened 2020-05-19
Fix --file option

raz-adroll:
## Type of change

- [X] Bugfix

## Description of change
Fixes the `--file` option for `gulp test`, related to issue https://github.com/prebid/Prebid.js/issues/5211

### DIFF
--- a/karma.conf.maker.js
+++ b/karma.conf.maker.js
@@ -112,6 +112,8 @@ module.exports = function(codeCoverage, browserstack, watchMode, file) {
   var plugins = newPluginsArray(browserstack);
 
   var files = file ? ['test/helpers/prebidGlobal.js', file] : ['test/test_index.js'];
+  var testFile = files[files.length - 1];
+
   // This file opens the /debug.html tab automatically.
   // It has no real value unless you're running --watch, and intend to do some debugging in the browser.
   if (watchMode) {
@@ -136,7 +138,7 @@ module.exports = function(codeCoverage, browserstack, watchMode, file) {
     // preprocess matching files before serving them to the browser
     // available preprocessors: https://npmjs.org/browse/keyword/karma-preprocessor
     preprocessors: {
-      'test/test_index.js': ['webpack', 'sourcemap']
+      [testFile]: ['webpack', 'sourcemap']
     },
 
     // web server port


smenzer:
resolves https://github.com/prebid/Prebid.js/issues/5211

jaiminpanchal27:
I tried running couple of core spec files but getting bunch of errors. I think your config looks correct but our spec files will need many changes to support this. 

I tried below two spec files
gulp test --file=/Users/jpanchal/code/prebid/test/spec/unit/core/adapterManager_spec.js
gulp test --file=/Users/jpanchal/code/prebid/test/spec/unit/core/targeting_spec.js

raz-adroll:
I that some test cases succeed while others fail. Fixing tests that don't run when being run standalone is outside of the scope of this fix.

raz-adroll:
The intent for this PR is not to make every single test runnable in with the `--file` option. But rather fix that it actually runs the specified file, the already failing tests are something that should be addressed in other PR.

Despite that, this weekend I tried to make all tests runnable with the `--file` option, but the majority of the failing tests, fail because they relay in either `hook.ready()` to be called before the tests or because they needed `src/prebid.js` to be imported or because they relay in some other side effect of previous tests.

I have all those changes in another branch that I can share if there is interest otherwise we can merge or close this PR.

smenzer:
I tend to agree with @raz-adroll about the focus of this PR, and think we should merge it and deal with the other failing tests in a separate PR.

I would love to see the other PR that you worked on to correct some of the other tests, too; at the very least so we can provide guidance to developers on how to write tests that can be run as a standalone file.

AaronColbyPrice:
@raz-adroll Curious if this issue is just being dropped.  I manually added this change to my branch and it was a great help.  Seems a shame to not include this because it doesn't magically fix everything.  It is still a great fix for an essentially broken feature.

raz-adroll:
@AaronColbyPrice it was never re-reviewed or accepted, which is a shame. I tried to fix all failing tests that don't run standalone but with a low degree of success. Maybe I should create a PR for each test I fix in standalone mode.

AaronColbyPrice:
@raz-adroll it looks like it has now been fixed by someone else here:

https://github.com/prebid/Prebid.js/commit/d38b5d017e658aaebe28e846318f522fd0dfacc6

I was surprised when the '--file' option was working even without your fix.  Well one less thing to worry about :)  Thanks for the initial fix that was a big help for me.

PR #5274, opened 2020-05-20
Fix referer

relaido:
## Type of change

- [ ] Bugfix
- [x] Feature
- [ ] New bidder adapter   
- [ ] Build related changes
- [ ] CI related changes
- [ ] Other

## Description of change
Add the referrer to the end of the payload as it may not be encoded.


smenzer:
@relaido it would be great if you could add some tests around this, too

PR #5279, opened 2020-05-21
Identity link id system - handle empty response

mamatic:
## Type of change
- [X] Bugfix

## Description of change
Handle "successful" empty responses. Prebid’s ajax.js utility treats all 200-level responses (e.g. 204) as successful, which means that our try/catch is not triggered. However, when we attempt to read out the envelope content, we generate a prebid error that is uncaught.

## Other information
@bretg Please, when you have time take a look.


smenzer:
LGTM, would be great if you could add a test case around this, too

mamatic:
> LGTM, would be great if you could add a test case around this, too

Sure, I will add couple of tests, no problem.

test/spec/modules/identityLinkIdSystem_spec.js
```diff
@@ -0,0 +1,92 @@
+import {identityLinkSubmodule} from 'modules/identityLinkIdSystem.js';
+import * as utils from 'src/utils.js';
+import {server} from 'test/mocks/xhr.js';
+
+const pid = '14';
+const defaultConfigParams = {pid: pid};
+const responseHeader = {'Content-Type': 'application/json'}
+
+describe('LiveIntentId', function () {
```
smenzer, line 9: This should probably be `IdentityLinkId` :-)

mamatic: Ups :) right away.

Issue #5103, opened 2020-04-09
Remove ambiguity from sizeMappingV2 log messages


Fawke:
## Type of issue
With the merge of this PR, https://github.com/prebid/Prebid.js/pull/5062, sizeMappingV2 supports ad unit declarations with the same code. Here's an example of ad unit with same code:

```(javascript)
[
  {
    code: 'ad-code-1',
    mediaTypes: {
      banner: {
        sizeConfig: [
          { minViewPort: [0, 0], sizes: [] },
          { minViewPort: [1000, 0], sizes: [[1000, 300], [1000, 90], [970, 250], [970, 90], [728, 90]] },
        ],
      },
    },
    bids: [{
      bidder: 'bidderA',
      params: {
        placementId: 123
      }
    }]
  }, {
    code: 'ad-code-1',
    mediaTypes: {
      banner: {
        sizeConfig: [
          { minViewPort: [0, 0], sizes: [] },
          { minViewPort: [320, 0], sizes: [[320, 100], [320, 50], [300, 50]] },
          { minViewPort: [600, 0], sizes: [] },
        ],
      },
    },
    bids: [{
      bidder: 'bidderA',
      params: {
        placementId: 456
      }
    }]
  },
]
```

Find more details about this pattern here: https://github.com/prebid/Prebid.js/issues/5044, https://github.com/prebid/Prebid.js/issues/3226#issuecomment-496427464

Earlier, log messages in sizeMappingV2 assumed ad unit code to be a unique identifier. But, after introduction of 'identical' ad unit pattern, that's no longer the case. This leads to some degree of ambiguity in log messages. For example, for the above ad unit, and screen size, 1000px, the second ad unit should be disabled.

Here's the generated log messages:
<img width="939" alt="Screenshot 2020-04-02 at 4 52 41 PM" src="https://user-images.githubusercontent.com/14229985/78926621-03d53980-7abb-11ea-9845-b5e6e2812552.png">

Reading it, it's impossible to figure out which ad unit got disabled, the first one or the second one.

## Potential Solution

Add identifier in paranthesis which would be unique. For example, for the above ad units, the first ad unit will read `ad-code-1 (1)` and second ad unit will read `ad-code-1 (2)`

If anyone has any better idea, please do share with me.

PR #5283, opened 2020-05-21
Improvements for SizeMappingV2 (advanced size mapping) module

Fawke:
## Type of change

- [X] Enhancements
- [X] Bugfix

## Description of change

With recent addition of '[Identical Ad Units](https://github.com/prebid/Prebid.js/pull/5062)' feature to this module, some log messages getting displayed were not very informative. Refer to [this](https://github.com/prebid/Prebid.js/issues/5103) issue for more details. 

- This PR implements the potential solution discussed in that issue.

- Log messages have been improved for Size Mapping V2 for more clarity when debugging.

- The example setup in `integrationExamples/gpt/advanced_size_mapping.html` now has instructive comments to help the user understand what's going on. 

- Resolves a minor bug in which bidder level sizeConfig checks were not checking one specific scenario.

Resolves https://github.com/prebid/Prebid.js/issues/5103


modules/sizeMappingV2.js
```diff
@@ -429,57 +438,56 @@ export function getRelevantMediaTypesForBidder(sizeConfig, activeViewport) {
...
   // check if the adUnit exists already in the sizeMappingInterStore (check for equivalence of 'code' && 'mediaTypes' properties)
   const adUnitDetail = adUnitsForAuction.filter(adUnitDetail => adUnitDetail.adUnitCode === adUnit.code && utils.deepEqual(adUnitDetail.mediaTypes, adUnit.mediaTypes));
 
   if (adUnitDetail.length > 0) {
+    adUnitDetail[0].cacheHits++;
     return adUnitDetail[0];
   } else {
-    const { mediaTypes, sizeBucketToSizeMap, activeViewport, transformedMediaTypes } = internal.getFilteredMediaTypes(adUnit.mediaTypes);
+    const identicalAdUnit = adUnitsForAuction.filter(adUnitDetail => adUnitDetail.adUnitCode === adUnit.code);
+    const adUnitInstance = identicalAdUnit.length > 0 ? typeof identicalAdUnit[0].instance === 'number' ? identicalAdUnit[identicalAdUnit.length - 1].instance + 1 : 1 : 1;
```
msm0504, line 453: Are the nested ternary operators needed here? Can this be rewritten as:
```
identicalAdUnit.length > 0 && typeof identicalAdUnit[0].instance === 'number' ? identicalAdUnit[identicalAdUnit.length - 1].instance + 1 : 1;
```

modules/sizeMappingV2.js
```diff
@@ -429,57 +438,56 @@ export function getRelevantMediaTypesForBidder(sizeConfig, activeViewport) {
...
+    const isLabelActivated = internal.isLabelActivated(adUnit, labels, adUnit.code, adUnitInstance);
+    const { mediaTypes = adUnit.mediaTypes, sizeBucketToSizeMap, activeViewport, transformedMediaTypes } = isLabelActivated && internal.getFilteredMediaTypes(adUnit.mediaTypes);
```
msm0504, line 455: If `isLabelActivated ` is false, then all of the destructured properties will be `undefined`. Is that correct? Could that cause any errors elsewhere?

Fawke: Yes, those properties can be left as undefined if `isLabelActivated` is false. Only exception being `medaiTypes` which I've already initialized.

This step is done for optimization. Basically the idea is, if an entire ad unit is deactivated by a failing label check, there is no need call function `getFilteredMediaTypes` and compute those values which anyway won't be used.

modules/sizeMappingV2.js
Fawke: @msm0504 Good catch! I've made the necessary change.

PR #5285, opened 2020-05-21
Secure creative update

harpere:
## Type of change

- [ ] Bugfix
- [ ] Feature
- [ ] New bidder adapter   
- [ ] Build related changes
- [ ] CI related changes
- [ x] Other

## Description of change

when returning the creative in response to a postMessage() request from an iframe (i.e. safeframe), restrict to the postMessage `event.origin` rather than a hard coded `adServerDomain`.   This changes functionality, but really should be a non-breaking change.

### DIFF
--- a/test/spec/unit/secureCreatives_spec.js
+++ b/test/spec/unit/secureCreatives_spec.js
@@ -30,12 +30,11 @@ describe('secureCreatives', () => {
         cpm: '1.00',
         adUnitCode: 'some_dom_id'
       };
-      const remoteDomain = '*';
       const source = {
         postMessage: sinon.stub()
       };
 
-      _sendAdToCreative(mockAdObject, remoteDomain, source);
+      _sendAdToCreative(mockAdObject, source);
       expect(JSON.parse(source.postMessage.args[0][0]).ad).to.equal('<script src="http://prebid.org/creative/1.00"></script>');
       expect(JSON.parse(source.postMessage.args[0][0]).adUrl).to.equal('http://creative.prebid.org/1.00');
       window.googletag = oldVal;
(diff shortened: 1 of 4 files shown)


test/spec/unit/secureCreatives_spec.js
```diff
@@ -30,14 +30,14 @@ describe('secureCreatives', () => {
...
+        source: { postMessage: sinon.stub() },
+        origin: 'origin.sf.com'
       };
 
-      _sendAdToCreative(mockAdObject, remoteDomain, source);
-      expect(JSON.parse(source.postMessage.args[0][0]).ad).to.equal('<script src="http://prebid.org/creative/1.00"></script>');
-      expect(JSON.parse(source.postMessage.args[0][0]).adUrl).to.equal('http://creative.prebid.org/1.00');
+      _sendAdToCreative(mockAdObject, event);
+      expect(JSON.parse(event.source.postMessage.args[0][0]).ad).to.equal('<script src="http://prebid.org/creative/1.00"></script>');
+      expect(JSON.parse(event.source.postMessage.args[0][0]).adUrl).to.equal('http://creative.prebid.org/1.00');
```
robertrmartinez, line 40: Should it not be https always now, I know for DFP / GAM served creatives we made a change at one point to always make it secure from the PUC to Prebid.js, is that just a one way thing?

harpere: it will depend on the event source (`event.source`), so if it comes from a non-secure page, it would be non secure.    We could change the `mockAdObject` in the test to use secure URLs, but I don't think that's pertinent to what's being tested here.

robertrmartinez: Alright sounds good!

PR #5291, opened 2020-05-22
ConsentManagement: fallback to defaultGdprScope if gdprApplies undefined

robertrmartinez:
## Type of change

- [X] Other

## Description of change
If a cmp responds with a valid tcf string but for some reason not with gdprApplies as a boolean, we should still use the tcfString.

Currently, the response is considered "invalid" and thrown out.

This change will set gdprApplies to whatever the defaultGdprScope is in the case where the CMP responds with a non - boolean gdprApplies.

### DIFF
--- a/modules/consentManagement.js
+++ b/modules/consentManagement.js
@@ -301,7 +301,8 @@ function processCmpData(consentObject, hookConfig) {
   }
 
   function checkV2Data() {
-    let gdprApplies = consentObject && consentObject.gdprApplies;
+    // if CMP does not respond with a gdprApplies boolean, use defaultGdprScope (gdprScope)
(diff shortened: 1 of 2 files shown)


bretg:
> This change will set gdprApplies to whatever the defaultGdprScope is in the case where the CMP responds with a non - boolean gdprApplies.

Are we aware of a CMP that does this, or is it just an excellent QA person pushing the envelope with bad data like they should?

As long as this is truly a CMP error condition, then this PR is just good defensive coding. Thanks.

robertrmartinez:
I lean towards "it just an excellent QA person pushing the envelope with bad data like they should"

But I have not been on the publisher cmp setup side or working with them to know if this is ever a legit use case.

But, yes, even if it is an edge case, I do not think this would hurt anything.

PR #5295, opened 2020-05-22
Floors new signals

robertrmartinez:
## Type of change
- [X] Other

### Price Floors Module

The floors module is updated to pass along a couple additional signals to analytics adapters for better logging and understanding what occurs pre auction.

#### New signals to analytics adapters to consume:
`fetchFailed` - indicates wether or not a fetch attempt was made for floor data and if it received a non 200 http status code.

`skipRate` - the resolved skipRate which was used for the auction. skipRate can come from the floor data or the floor setConfig level. It is now signaled to analytics adapters.

#### New debug feature skipRate override
By setting pbjs_skipRate=<number> as a query param in the page url will have each auction use the passed skipRate .
 

### Rubicon Analytics Adapter

Log these two new fields in the event payload.

### DIFF
--- a/modules/priceFloors.js
+++ b/modules/priceFloors.js
@@ -288,6 +288,7 @@ export function updateAdUnitsForAuction(adUnits, floorData, skipped, auctionId) 
         skipped,
         modelVersion: utils.deepAccess(floorData, 'data.modelVersion') || '',
         location: floorData.data.location,
+        skipRate: floorData.skipRate
       }
     });
   });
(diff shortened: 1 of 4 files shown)


robertrmartinez:
@Fawke Thanks, we actually just made a small change to make it `fechStatus` now instead of `fetchFailed`

It will be an enum of `success` || `timeout` || `error`. And undefined if no fetch actually was initiated.

Bryan Szekely will make a docs PR.

bszekely1:
PR for docs when released: https://github.com/prebid/prebid.github.io/pull/2013

PR #5315, opened 2020-05-28
UserId SharedId submodule

SKOCHERI:
## Type of change

- [ ] Bugfix
- [x] Feature
- [ ] New bidder adapter   
- [ ] Build related changes
- [ ] CI related changes
- [ ] Other

## Description of change

```
{
  bidder: '<bidder name>',
  params: {
    // ...
  }
}
```


modules/sharedIdSystem.js
```diff
@@ -0,0 +1,210 @@
...
+  /**
+   * performs actions even if the id exists and returns a value
+   * @param configParams
+   * @param storedId
+   * @returns {{callback: *}}
+   */
+  extendId(configParams, storedId) {
+    utils.logInfo('SharedId: Existing shared id ' + storedId.id);
+    const resp = function (callback) {
+      const needSync = isIdSynced(configParams, storedId);
```
smenzer, line 196: couldn't you rely on the `refreshInSeconds` feature instead of reproducing it?

https://github.com/prebid/Prebid.js/blob/master/modules/userId/index.js#L64

modules/sharedIdSystem.js
```diff
@@ -0,0 +1,210 @@
...
+      if (needSync) {
+        utils.logInfo('SharedId: Existing shared id ' + storedId + ' is not synced');
+        const sharedIdPayload = {};
+        sharedIdPayload.sharedId = storedId.id;
+        const payloadString = JSON.stringify(sharedIdPayload);
+        ajax(ID_SVC, existingIdCallback(storedId, callback), payloadString, {method: 'POST', withCredentials: true});
```
smenzer, line 202: is it intentional that this is a POST call while the `getId` method uses a GET?

SKOCHERI: Yes this is intentional as we want to send the existing id value

smenzer: Ok...you could still use a GET to pass it as a querystring param, but obviously no problem to do it as a POST, I just wanted to make sure it was intentional.

modules/sharedIdSystem.js
SKOCHERI: couple of reason for not using "refreshInSeconds"

1. The default of refreshInSeconds is the expiry of the cookies but we are looking at refreshing/syncing atleast once in 24 hours
2. The refresh based on refreshInSeconds relies on getId function to reset where as we want to use extendId function

patmmccann:
Tom's board motion suggested combining this with the pubcommonid. I think that would be a superior PR for several reasons

You would get a large publisher base out of the gate
You would get wide ssp support out of the gate
You would prevent orphaning another submodule

SKOCHERI:
The documentation is available in https://github.com/prebid/prebid.github.io/pull/2031

SKOCHERI:
@jsnellbaker awaiting for your review

jdwieland8282:
docs: https://github.com/prebid/prebid.github.io/pull/2031

jsnellbaker:
@SKOCHERI Can you take a look at the code conflicts - I'll start to review.

src/sharedIdGenerator.js
```diff
@@ -0,0 +1,132 @@
+/**
+ * Generates the shared Id/Ulid. Id generated is 128 bits with first 48 bits of timestamp followed by 80 bit of randomness.
+ * Spec is available in https://github.com/ulid/spec
+ */
```
jsnellbaker, line 4: Does this functionality really need its own file?  The src directory is meant for 'core' files, not something that's dependent on a module.  I realize there are restrictions in our project about module files importing other module files, which is why I was wondering the need for a separate file.  Please let me know your thoughts on this point.

SKOCHERI: In order to have the id generation separated from the other shared id functionality like the sync/decode we have it in a separate file.
Since this sharedIdGenerator.js generates ulid , we could rename it to ulidGenerator.js.

There is no restriction to import anything from src folder but we did notice restriction when we added sharedIdGenerator.js within the userId module and tried to import, we encountered ESLint valid-import restriction

smenzer: @SKOCHERI @jsnellbaker 
I think it should either be renamed to `ulidGenerator.js` or moved into the `sharedIdSystem.js` module...as to which one, I don't have a strong opinion.

If it's in a standalone ulidGenerator.js file, I assume it would only ever be included in a Prebid.js build if a module that uses/imports it (like sharedId) is included in the build, right? In other words, this won't add weight to the prebid build unless it's being used somewhere...?

SKOCHERI: @jsnellbaker Awaiting for your thoughts on renaming sharedIdGenerator.js to ulidGenerator.js

jsnellbaker: I'm trying to gather some additional thoughts from other maintainers of Prebid.js on this overall topic.  I'll follow-up when I can.

SKOCHERI: @jsnellbaker We are inclined to move the sharedIDGenerator.js within the sharedId module itself. I hope that should be ok?

jsnellbaker: @SKOCHERI If you wish to do this change, that's fine.  I think if there's a stronger desire to use this type of encoding down-the-road, we can potentially revisit this idea of having a dedicated file for generating the ID.

SKOCHERI: @jsnellbaker  Have moved the sharedId generation within sharedId submodule. Could you compete your review

modules/sharedIdSystem.js
smenzer: @SKOCHERI ok. i'll note that I still think there are better ways to accomplish this (some of which may require userId module changes), but I think it's fine to leave as is and perhaps gets reviewed at another time down the road.

Issue #5141, opened 2020-04-19
pbjs.requestBids() fails if digitrust saved object has null for data


maxim-novikov:
## Type of issue
bug

## Description
If digitrust module is enabled and initialization is failed, local stored object stores null as 'data' variable. pbjs.requestBids() function fails encountering it.

## Steps to reproduce
Run prebid-enabled page with digitrust module enabled and configured on clean Firefox browser (to ensure digitrust failure). Reload page, so that digitrust stored object applies before first pbjs.requestBids() call. Error message appears on the console. 

## Test page
https://www.warhistoryonline.com/videos/digitrust.fail.v0.html?pbjs_debug=true

## Platform details
Prebid.js 3.16+

## Other information
While the issue is clearly observable on the latest Firefox, it appears to happen on the latest Chrome as well, provided the digitrust initialization fails for whatever reason, so one would assume it's browser-independent.

![exhibit a](https://user-images.githubusercontent.com/4518331/79685633-e6cf0200-8242-11ea-803a-d2947dd111f1.png)
![exhibit b](https://user-images.githubusercontent.com/4518331/79685634-e7679880-8242-11ea-9d3f-5c0396f899ec.png)

samtingleff:
@goosemanjack please take a look

goosemanjack:
Somehow this slipped me. I will take a look.

goosemanjack:
Had to clear some items off my plate. Tomorrow looks promising to work on this.

goosemanjack:
(From Sam) not sure if it is the same issue, but #5141 provides a public test page at https://www.warhistoryonline.com/videos/digitrust.fail.v0.html?pbjs_debug=true

goosemanjack:
There was a null check missing in the call chain. A PR has been submitted:
https://github.com/prebid/Prebid.js/pull/5333

PR #5333, opened 2020-06-04
Fix to issue #5141 exception thrown in pbjs.requestBids when DigiTrus…

goosemanjack:
Fix to issue #5141 exception thrown in pbjs.requestBids when DigiTrust does not init.

Safe extract value of id or null in eids.js.

## Type of change

- [ X] Bugfix

## Description of change
This modifies eids.js where the digitrust ID is extracted. The previous code was pulling a deep property without first checking for null values. The change adds in null checks and returns null instead of throwing an exception.

- [X ] official adapter submission

## Other information
Issue #5141

### DIFF
--- a/modules/userId/eids.js
+++ b/modules/userId/eids.js
@@ -64,8 +64,9 @@ const USER_IDS_CONFIG = {
 
   // DigiTrust
   'digitrustid': {
-    getValue: function(data) {
-      return data.data.id;
+    getValue: function (data) {
+      var id = data && data.data && data.data.id || null;
+      return id;
     },
     source: 'digitru.st',
     atype: 1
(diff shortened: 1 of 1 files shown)


goosemanjack:
The PR #5334 will eclipse this PR, once committed.

modules/userId/eids.js
```diff
@@ -64,8 +64,12 @@ const USER_IDS_CONFIG = {
...
   'digitrustid': {
-    getValue: function(data) {
-      return data.data.id;
+    getValue: function (data) {
+      var id = null;
+      if (data && data.data && data.data.id != null) {
+        id = data.data.id;
+      }
+      return id;
     },
```
harpere, line 73: couldn't we just simplify to this?
``` return data && data.data && data.data.id; ```

harpere: that's too bad.  I like the concise format you tried first.

PR #5350, opened 2020-06-09
parrableIdSystem: Populate userIdAsEid with Parrable ID and Optout data

icflournoy:
## Type of change
- [x] Feature

## Description of change

We add a definition in the userId/eids module that transforms our Parrable ID into the ORTB EIDs format. We supply our iba and ccpa optout statuses as UID extended data.

## Implementation changes

- Our parrableIdSystem now returns the entire Parrable ID object from decode(). Our encrypted ID (eid) is now available at `userId.parrableId.eid` and we have updated the bid adapters that reference this. Bid adapters that wish to consume our ID in the ORTB format can now do so and receive optout states.
- To be more consistent we've renamed the ID system key to `parrableId` to match the userId module name.


icflournoy:
If someone can point me in the right direction on where I can document this change I'll happily submit a PR.

I can't find a place where the userId formats are documented outside of the userId modules, bidAdapters using the IDs, and the spec tests covering those modules.

This change has no bearing on the way someone configures our userId system, so the Parrable documentation on Prebid.org remains unchanged.

icflournoy:
After discussion with Openx we have rolled back the breaking change to their BidAdapter. They are comfortable with receiving just our `parrableId.eid` value.

cc @jimee02

icflournoy:
Hi @idettman let me know what I can do to address any doc/design questions here. I see this has a label for needs-docs but am unclear on where I need to write documentation for this change.

justinegreene:
@jimee02, @idettman following up on @icflournoy's post.  We could really use some help moving this forward.  It's been a month and we have one more PR to enable client time zone based filtering that we cannot submit until this is merged.  Thank you 🙏🏻

patmmccann:
wouldn't you document this here? https://github.com/prebid/prebid.github.io/blob/master/dev-docs/modules/userId.md

jimee02:
@icflournoy, we currently do not have the capability of passing more than an ID string downstream through our current API, however, we are finalizing a build for an openRTB-supported that will pass the entire eid object through.   When I have a PR ready I'll add you as a reviewer.  It should be ready by end of August.

icflournoy:
Hi @patmmccann the docs pr is here: https://github.com/prebid/prebid.github.io/pull/2126 this updates only the userId submodule interface.

I was looking for a place to document our extended userId info in the ORTB format, the closest place I can find is here: https://github.com/prebid/Prebid.js/blob/master/modules/userId/eids.md

@jimee02 that's great to hear, would be happy to take a look when the time comes. As for now, this PR maintains compatibility with your adapter and you will still be receiving only the ID string.

bretg:
@idettman is a little backed up, so reassigning over to @jlukas79 . I believe the question has been answered here.

justinegreene:
Hi @jlukas79, please let us know if you have any questions. Would really really appreciate some attention 🙏🏻

mike-chowla:
Change to the PubMatic adapter LGTM

bretg:
@jsnellbaker - can you drive this one home?

icflournoy:
@jsnellbaker thanks for the catch. I've updated Vidazoo's bid adapter to consume our new ID structure. Tagging @uditalias who last worked on the adapter for a set of eyes on my change.

uditalias:
Thanks for the update @icflournoy 
LGTM

anand-venkatraman:
The changes to PulsePoint adapter, LGTM. 
Thanks

PR #5379, opened 2020-06-16
delay request to server by 5 min if we no ads

vincentproxistore:
## Type of change

- [ ] Bugfix
- [X] Feature
- [ ] New bidder adapter   
- [ ] Build related changes
- [ ] CI related changes
- [ ] Other

## Description of change

Avoid too many request to our server if we have no ads for the moment.
If we have no ads we write the name of the website as a key and the current date in the local storage. The next time we read it and if we compare to the current time and if the current time is superior to five minutes we call our server if not we reject the bid.

vincent.descamps@proxistore.com

### DIFF
--- a/modules/proxistoreBidAdapter.js
+++ b/modules/proxistoreBidAdapter.js
@@ -3,15 +3,15 @@
 import { registerBidder } from '../src/adapters/bidderFactory.js';
 import { getStorageManager } from '../src/storageManager.js';
 
-var BIDDER_CODE = 'proxistore';
+const BIDDER_CODE = 'proxistore';
 const storage = getStorageManager();
-var PROXISTORE_VENDOR_ID = 418;
+const PROXISTORE_VENDOR_ID = 418;
 
 function _createServerRequest(bidRequests, bidderRequest) {
-  var sizeIds = [];
+  const sizeIds = [];
 
   bidRequests.forEach(function (bid) {
-    var sizeId = {
+    const sizeId = {
       id: bid.bidId,
       sizes: bid.sizes.map(function (size) {
         return {
@@ -22,7 +22,7 @@ function _createServerRequest(bidRequests, bidderRequest) {
     };
     sizeIds.push(sizeId);
   });
-  var payload = {
+  const payload = {
     auctionId: bidRequests[0].auctionId,
     transactionId: bidRequests[0].auctionId,
     bids: sizeIds,
@@ -32,7 +32,7 @@ function _createServerRequest(bidRequests, bidderRequest) {
       applies: false
     }
   };
-  var options = {
+  const options = {
     contentType: 'application/json',
     withCredentials: true
   };
@@ -91,7 +91,7 @@ function isBidRequestValid(bid) {
       const storedDate = new Date(pxNoAds);
       const now = new Date();
       // 5min = 300 0000 seconds
-      var diff = Math.abs(storedDate.getTime() - now.getTime()) / 300000;
+      const diff = Math.abs(storedDate.getTime() - now.getTime()) / 300000;
       if (diff > 5) {
         return false;
       } else {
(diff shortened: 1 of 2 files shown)


robertrmartinez:
Assigned to @FilipStamenkovic 

Please add unit test(s) for this change @vincentproxistore

modules/proxistoreBidAdapter.js
```diff
@@ -59,17 +71,31 @@ function _createBidResponse(response) {
...
 function isBidRequestValid(bid) {
-  return !!(bid.params.website && bid.params.language);
+  const hasNoAd = function() {
+    const storedtime = new Date(localStorage.getItem(`PX_NoAds_${bid.params.website}`));
+    if (!storedtime) {
+      return false;
+    } else {
+      const now = new Date();
+      var diff = Math.abs(storedtime.getTime() - now.getTime()) / 30000;
+      if (diff > 5) {
```
FilipStamenkovic, line 91: here instead of this `if-else` you could just do:

```
return diff <= 5;
```

modules/proxistoreBidAdapter.js
```diff
@@ -79,38 +105,51 @@ function isBidRequestValid(bid) {
...
 /**
  * Unpack the response from the server into a list of bids.
  *
  * @param serverResponse A successful response from the server.
  * @param bidRequest Request original server request
  * @return  An array of bids which were nested inside the server.
  */
+
 function interpretResponse(serverResponse, bidRequest) {
+  const itemName = `PX_NoAds_${websiteFromBidRequest(bidRequest)}`;
```
FilipStamenkovic, line 122: Imagine following scenario here:  
- Some user didn't receive any bid.
- You store current timestamp
- User shuts down his pc
- Then, user shows up next day requesting bids
- Timestamp you would read is from yesterday, which is clearly older than 5 minutes and then you would mark bid request as invalid and this user could never request bids again.

Did I follow through your code correctly? Is this your intent?

modules/proxistoreBidAdapter.js
```diff
@@ -1,11 +1,24 @@
-const { registerBidder } = require('../src/adapters/bidderFactory.js');
+
```
FilipStamenkovic, line 1: Can you remove this empty line? It is not common practice to start .js file with empty line.

modules/proxistoreBidAdapter.js
```diff
@@ -59,17 +72,28 @@ function _createBidResponse(response) {
...
  * Determines whether or not the given bid request is valid.
  *
  * @param bid  The bid params to validate.
  * @return boolean True if this is a valid bid, and false otherwise.
  */
+
 function isBidRequestValid(bid) {
-  return !!(bid.params.website && bid.params.language);
+  const hasNoAd = function() {
+    const pxNoAds = storage.getDataFromLocalStorage(`PX_NoAds_${bid.params.website}`);
```
FilipStamenkovic, line 86: This function call can throw an exception if user has disabled browser to store any data.  

Either wrap it in `try-catch` or check if the `localStorage` is enabled by executing:

```
storage.hasLocalStorage();
```

modules/proxistoreBidAdapter.js
```diff
@@ -59,17 +72,28 @@ function _createBidResponse(response) {
...
+    if (!pxNoAds) {
+      return false;
+    } else {
+      const storedDate = new Date(pxNoAds);
+      const now = new Date();
+      const diff = Math.abs(storedDate.getTime() - now.getTime()) / 60000;
+      return diff <= 5;
+    }
+  }
+  return !!(bid.params.website && bid.params.language) && !hasNoAd();
```
FilipStamenkovic, line 96: You don't have to change this one, this is small suggestion from my side: you could remove double negation here.  
You check `!hasNoAd()`, you could rename function to `hasAd()` and invert logic in that function and then just check `hasAd()`. To me, it's more readable.

PR #5388, opened 2020-06-17
Add lotame id system

markaconrad:
## Type of change
- [x] New user id module

## Description of change
Adds a new user id module named lotamePanoramaId

```
{
        name: 'lotamePanoramaId'
}
```

Tested with
(/integrationExamples/gpt/userId_example.html) sample page.

Maintainer Email Address
- prebid@lotame.com

https://github.com/prebid/prebid.github.io/pull/2070


markaconrad:
Hmmm...not sure how to resolve the CircleCi test fails...

modules/lotamePanoramaIdSystem.js
```diff
@@ -0,0 +1,243 @@
...
+                );
+              } else {
+                clearLotameCache(KEY_ID);
+              }
+
+              if (utils.isStr(responseObj.profile_id)) {
+                setProfileId(responseObj.profile_id);
+              } else {
+                clearLotameCache(KEY_PROFILE);
+                clearLotameCache(KEY_ID);
```
smenzer, line 223: maybe i'm misreading the code here...but this seems to say that if you had a situation where the response contained `core_id` but not `profile_id`, on line 210 you would write the `core_id` to cache, but then here you would immediately remove it from cache. is that expected behavior? if so, why not combine the statements to avoid writing and deleting the `core_id`.

modules/lotamePanoramaIdSystem.js
```diff
@@ -0,0 +1,243 @@
...
+              }
+            } catch (error) {
+              utils.logError(error);
+            }
+          }
+          callback(responseObj);
```
smenzer, line 229: again, i may be missing something, but here you are storing to the prebid cache object the entire `responseObj`, which seems to contain several keys, including the `core_id` and `profile_id`. however, in your decode function on line 157, you are checking if the `value` is a string, otherwise you return `undefined`. my understanding is that the `value` you receive in `decode()` will be the same thing you pass in the `callback()` here...which means decode would always return undefined.  again, i'm sure i'm missing something here since you're doing some of your own cache management, but wanted to flag this as it doesn't seem correct.

modules/lotamePanoramaIdSystem.js
markaconrad: Good point. The server could send just a profile_id OR a profile_id and core_id but not just a core_id. I can rework to make that clearer

modules/lotamePanoramaIdSystem.js
markaconrad: Looking into this part as it was tricky to trace what `decode` will always get as there are a couple of different paths

markaconrad: Pushed up a fix, thanks for catching this!

modules/lotamePanoramaIdSystem.js
markaconrad: Pushed up a fix

pm-harshad-mane:
Hello @markaconrad 
Can you please add a Unit test case in https://github.com/prebid/Prebid.js/blob/master/test/spec/modules/eids_spec.js for Lotame ID?

markaconrad:
@msm0504 Just checking in to see if you need anything else from me before this is good to go. Thanks!

msm0504:
@markaconrad I don't think I need anything right now. Sorry, I haven't had a chance to review this yet. I'll review this week.

Issue #5339, opened 2020-06-05
Price Floors Module: Multiple Floor Model Selection


bszekely1:
## Type of issue

Intent to implement

## Description

The intent of this ticket is to provide a method for [Floor Providers ](http://prebid.org/dev-docs/modules/floors.html#floor-data-provider-interface)to apply and A/B test different models when implemented in the [Prebid Floors Module](https://github.com/prebid/Prebid.js/blob/master/modules/priceFloors.js). Today Floor Providors are limited to applying one model per URL fetch. To apply more than one model per user session today, a Floor Provider would need to reload new fetch URLs for each model, controling the sampling rate and which file to load outside of prebid logic.

## Goals

- Define a flexible method for Floor Providers to A/B test different models using the Prebid Floors Module
- Mitigate any additional latency on the existing fetch model of the Floors Module

## General requirements
1. The new floor solution should be backwards compatible with exiting Floor Providers, and ad unit or setConfig floors
2. Floors Provider should be able to supply one or more models with their own set of schema fields and values in the new floors schema
3. Floor Providers should be able to specificy the frequency of each model should be applied by a sampling weight
4. Sampling weights are applied at the auction level to all ad slots after skipRate
5. The sampling weights should be recalculated (i.e. decide which model to select if there is more than one model) once the fetch or setConfig is applied to the current auction
6. The application of floors (signaling and enforcement) should remain the same using the original floors schema or new one
7. Weights of model selection should be normalized to 1 to allow Floor Providers to supply their own weights without having to calculate to 100%
8. When using schema version 2, a weight should be provided for each model, otherwise the floors module will not utilize the floors data set (i.e. other floors can potentially be used if valid)
9. Schema 2 can be applied run-time in either fetch or setConfig, not within adUnits
10. The following attributes should be applied at the root level and model level where the model level attribute overrides the root level one:
	- skipRate
	- currency

### New Attributes

Field Name | Type | Description
------------ | ------------- | -------------
floorsSchemaVersion | int | Describes the version of the schema to allow for backwards and future proofing for multiple schemas
modelGroups | object | Object to store the different models to apply per auction
modelGroups.modelWeight | int | The sampling factor determing when the model should be applied. Weights are normalized to 1. 

### Schema Validation Tool

The goal here is to provide a tool for Floor Providers to validate rules against a particular schema. The validation schema tool should be an HTML page to signal if a supplied rules file is valid. If the file is invalid, it should indicate where the error is.

[...]

PR #5390, opened 2020-06-18
Price floors new schema support AB Test

robertrmartinez:
Type of change
- [X] Feature

## Description of change
Implementation of #5339

DOCS: https://github.com/prebid/prebid.github.io/pull/2087

New schema for data object contains array of models. Each will contain a modelWeight which is used at auctionInit time to randomly select which model to use.

### DIFF
--- a/gulpfile.js
+++ b/gulpfile.js
@@ -341,12 +341,12 @@ function injectFakeServerEndpointDev() {
 
 function startFakeServer() {
   const fakeServer = spawn('node', ['./test/fake-server/index.js', `--port=${FAKE_SERVER_PORT}`]);
-    fakeServer.stdout.on('data', (data) => {
-      console.log(`stdout: ${data}`);
-    });
-    fakeServer.stderr.on('data', (data) => {
-      console.log(`stderr: ${data}`);
-    });
+  fakeServer.stdout.on('data', (data) => {
+    console.log(`stdout: ${data}`);
+  });
+  fakeServer.stderr.on('data', (data) => {
+    console.log(`stderr: ${data}`);
+  });
 }
 
 // support tasks
@@ -372,6 +372,7 @@ gulp.task('build', gulp.series(clean, 'build-bundle-prod'));
 gulp.task('build-postbid', gulp.series(escapePostbidConfig, buildPostbid));
 
 gulp.task('serve', gulp.series(clean, lint, gulp.parallel('build-bundle-dev', watch, test)));
+gulp.task('serve-fast', gulp.series(clean, gulp.parallel('build-bundle-dev', watch)));
 gulp.task('serve-fake', gulp.series(clean, gulp.parallel('build-bundle-dev', watch), injectFakeServerEndpointDev, test, startFakeServer));
 
 gulp.task('default', gulp.series(clean, makeWebpackPkg));
--- a/modules/priceFloors.js
+++ b/modules/priceFloors.js
@@ -295,8 +295,7 @@ export function updateAdUnitsForAuction(adUnits, floorData, auctionId) {
   });
 }
 
-export function pickRandomModel(modelGroups) {
-  const weightSum = modelGroups.reduce((accum, model) => accum += model.modelWeight, 0);
+export function pickRandomModel(modelGroups, weightSum) {
   // we loop through the models subtracting the current model weight from our random number
   // once we are at or below zero, we return the associated model
   let random = Math.floor(Math.random() * weightSum + 1)
@@ -317,7 +316,7 @@ export function createFloorsDataForAuction(adUnits, auctionId) {
   if (utils.deepAccess(resolvedFloorsData, 'data.floorsSchemaVersion') === 2) {
     // merge the models specific stuff into the top level data settings (now it looks like floorsSchemaVersion 1!)
     let { modelGroups, ...rest } = resolvedFloorsData.data;
-    resolvedFloorsData.data = Object.assign(rest, pickRandomModel(modelGroups));
+    resolvedFloorsData.data = Object.assign(rest, pickRandomModel(modelGroups, rest.modelWeightSum));
   }
 
   // if we do not have a floors data set, we will try to use data set on adUnits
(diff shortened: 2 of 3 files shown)


gulpfile.js
```diff
@@ -341,12 +341,12 @@ function injectFakeServerEndpointDev() {
 
 function startFakeServer() {
   const fakeServer = spawn('node', ['./test/fake-server/index.js', `--port=${FAKE_SERVER_PORT}`]);
-    fakeServer.stdout.on('data', (data) => {
-      console.log(`stdout: ${data}`);
-    });
-    fakeServer.stderr.on('data', (data) => {
-      console.log(`stderr: ${data}`);
-    });
```
robertrmartinez, line 349: my editor keeps yelling at me about this so I just fixed it.

gulpfile.js
```diff
@@ -372,6 +372,7 @@ gulp.task('build', gulp.series(clean, 'build-bundle-prod'));
 gulp.task('build-postbid', gulp.series(escapePostbidConfig, buildPostbid));
 
 gulp.task('serve', gulp.series(clean, lint, gulp.parallel('build-bundle-dev', watch, test)));
+gulp.task('serve-fast', gulp.series(clean, gulp.parallel('build-bundle-dev', watch)));
```
robertrmartinez, line 375: When I am just trying to test via my E2E test pages having to wait several minutes for lint + test is a little much.

Thought others might use a similar local edit, so why not merge it in.

modules/priceFloors.js
```diff
@@ -295,11 +295,30 @@ export function updateAdUnitsForAuction(adUnits, floorData, auctionId) {
...
+  // we loop through the models subtracting the current model weight from our random number
+  // once we are at or below zero, we return the associated model
+  let random = Math.floor(Math.random() * weightSum + 1)
+  for (let i = 0; i < modelGroups.length; i++) {
+    random -= modelGroups[i].modelWeight;
+    if (random <= 0) {
+      return modelGroups[i];
+    }
+  }
+};
```
robertrmartinez, line 309: Here is a jsbin to test this function.
https://jsbin.com/qihiqojovu/edit?js,console

Weights do not need to add up to 100. Only shown in jsbin for clarity.

modules/priceFloors.js
```diff
@@ -382,11 +425,12 @@ export function isFloorsDataValid(floorsData) {
   if (typeof floorsData !== 'object') {
     return false;
   }
-  // schema.fields has only allowed attributes
-  if (!validateSchemaFields(utils.deepAccess(floorsData, 'schema.fields'))) {
+  floorsData.floorsSchemaVersion = floorsData.floorsSchemaVersion || 1;
```
robertrmartinez, line 428: default to version 1 for backwards compat

bszekely1:
Docs PR: https://github.com/prebid/prebid.github.io/pull/2087

gulpfile.js
Fawke: This will come in handy!

modules/priceFloors.js
```diff
@@ -372,6 +391,30 @@ function validateRules(floorsData, numFields, delimiter) {
...
+ */
+const floorsSchemaValidation = {
+  1: data => modelIsValid(data),
+  2: data => {
+    // model groups should be an array with at least one element
+    if (!Array.isArray(data.modelGroups) || data.modelGroups.length === 0) {
+      return false;
+    }
+    // every model should have valid schema, as well as an accompanying modelWeight
+    return data.modelGroups.every(model => typeof model.modelWeight === 'number' && modelIsValid(model));
```
msm0504, line 414: Do all models need to be valid? Or could this work with whatever the list of valid models is?

modules/priceFloors.js
```diff
@@ -295,11 +295,30 @@ export function updateAdUnitsForAuction(adUnits, floorData, auctionId) {
   });
 }
 
+export function pickRandomModel(modelGroups) {
+  const weightSum = modelGroups.reduce((accum, model) => accum += model.modelWeight, 0);
```
msm0504, line 299: Can this weight sum be different each time `pickRandomModel` is called? If not, I'd suggest calculating this once whenever the list of models is set and then reusing the value in this function.

robertrmartinez: Yeo good call. At the time the entire data is set, we can calculate it there and set it as a prop on the object and use that. Will do this change.

modules/priceFloors.js
robertrmartinez: We could change this to a `filter`, instead of an `every`. I'll ask Bryan what he thinks.

Would want to add some error logging for a model which does not fit though.

bszekely1: Hey @msm0504 , while this is a good idea in spirit, in practice there would be some major implications if a floor provider is expecting to have a certain weight, then all the sudden it's increased due to a model not applied. This would be a difficult conversation with the publisher if a floor provider was reducing a model that was monetizing poorly and al the sudden it shot up to 100% (in an extreme example) if the new model had an error.

Having a binary working / not working system would alert the floor provider if an issue exists, where no floors would be used in this case. Having a silent issue could be more costly.

I appreciate thorough review Mark. Feedback is always welcome.

Bryan

modules/priceFloors.js
```diff
@@ -501,7 +507,13 @@ export function handleFetchResponse(fetchResponse) {
     floorResponse = fetchResponse;
   }
   // Update the global floors object according to the fetched data
-  _floorsConfig.data = parseFloorData(floorResponse, 'fetch') || _floorsConfig.data;
+  const fetchData = parseFloorData(floorResponse, 'fetch');
+  if (fetchData) {
+    // set .data to it
+    _floorsConfig.data = fetchData;
+    // set skipRate override if necessary
+    _floorsConfig.skipRate = utils.isNumber(fetchData.skipRate) ? fetchData.skipRate : _floorsConfig.skipRate;
```
robertrmartinez, line 515: We were not honoring fetch level skip rates, which is a bug. Need to get this out asap.

Added a test to confirm.

PR #5394, opened 2020-06-18
appnexusBidAdapter - fix video params

jsnellbaker:
## Type of change

- [x] Bugfix

## Description of change
This change fixes some of the video params to ensure they're passed in the correct manner to be read by the endpoint.

Docs PR: https://github.com/prebid/prebid.github.io/pull/2073


patmmccann:
why are you removing support for mime types?

PR #5410, opened 2020-06-22
[Synacormedia] adapter should use format for multi-size banner requests

coreykress:
## Type of change

- [x] Feature

## Description of change

Currently, the Synacormedia bid adapter will split any multi-size banner requests into individual impressions per size.  This change should make the adapter send a single bid request with a format property for multi-size banner units

Contact email: corey.kress@synacor.com


coreykress:
There isn't an actual error listed on the ci/circleci page.  It just says that it errors?  Can anyone elaborate on the issue?

PR #5411, opened 2020-06-22
Prebid JS Native Phase 2 update:

mmoschovas:
1) Key Suppression for Native keys with sendTargetingKeys set to false at nativeParams.sendTargetingKeys or within asset i.e. nativeParams.image.sendTargetingKeys. Logic as follows:
   - If nativeParams.sendTargetingKeys set to false but asset has sendTargetingKeys set to true then add the KV
   - Else if nativeParams.sendTargetingKeys is set to true and asset does not have sendTargetingKeys defined or it is set to true then add KV
   - Else if nativeParams.sendTargetingKeys is not defined and asset does not have sendTargetingKeys defined or it is set
 to true then add KV
   If no conditions met then key is suppressed
2) Adds rendererUrl or adTemplate to the bid object for easy retrieval if assets requested from creative
3) Adds new logic to receiveMessage function in secureCreatives.js to answer data.action of 'allAssetRequest' with all available assets on the bid that have a value
4) Adds new native keys to constants.json

## Type of change

- [ ] Bugfix
- [ ] Feature
- [ ] New bidder adapter   
- [ ] Build related changes
- [ ] CI related changes
- [ ] Other

## Description of change

```
{
  bidder: '<bidder name>',
  params: {
    // ...
  }
}
```


robertrmartinez:
Assigned to myself and @jaiminpanchal27 for review

src/native.js
```diff
@@ -154,22 +154,36 @@ export function fireNativeTrackers(message, adObject) {
...
+        value = placeholder;
+      }
+
+      if (key && value) {
+        if (!deepAccess(bidReq, 'nativeParams.sendTargetingKeys') && typeof deepAccess(bidReq, 'nativeParams.sendTargetingKeys') !== 'undefined' && deepAccess(bidReq, 'nativeParams.' + asset + '.sendTargetingKeys')) {
+          keyValues[key] = value;
+        } else if (deepAccess(bidReq, 'nativeParams.sendTargetingKeys') && (typeof deepAccess(bidReq, 'nativeParams.' + asset + '.sendTargetingKeys') === 'undefined' || deepAccess(bidReq, 'nativeParams.' + asset + '.sendTargetingKeys'))) {
+          keyValues[key] = value;
+        } else if (typeof deepAccess(bidReq, 'nativeParams.sendTargetingKeys') === 'undefined' && (typeof deepAccess(bidReq, 'nativeParams.' + asset + '.sendTargetingKeys') === 'undefined' || deepAccess(bidReq, 'nativeParams.' + asset + '.sendTargetingKeys'))) {
+          keyValues[key] = value;
```
robertrmartinez, line 184: I think this can be simplified a little.

```javascript
export function getNativeTargeting(bid, bidReq) {
  let keyValues = {};

  const globalSendTargetingKeys = deepAccess(
    bidReq,
    `mediaTypes.native.sendTargetingKeys`
  ) !== false; // if anything other than strict false, send keys

  Object.keys(bid['native']).forEach(asset => {
    const key = CONSTANTS.NATIVE_KEYS[asset];
    let value = getAssetValue(bid['native'][asset]);

    const sendPlaceholder = deepAccess(
      bidReq,
      `mediaTypes.native.${asset}.sendId`
    );

    const assetSendTargetingKeys = deepAccess(
      bidReq,
      `mediaTypes.native.${asset}.sendTargetingKeys`
    );

    const sendTargeting = typeof assetSendTargetingKeys === 'boolean' ? assetSendTargetingKeys : globalSendTargetingKeys;

    if (sendPlaceholder) {
      const placeholder = `${key}:${bid.adId}`;
      value = placeholder;
    }

    if (key && value && sendTargeting) {
      keyValues[key] = value;
    }
  });

  return keyValues;
}
```

before the loop we determine what the "top level" setting is -> If anything other than boolean false, then it is true, otherwise false.

Then for each asset, if sendTargetingKeys is a boolean for that asset, use that value, else use the top level one.

So instead of the if else if's, just add the boolean check in the if(key and value).

robertrmartinez: I did not test this, but It should work in theory, I may have a typo or something.

Let me know if you have questions or think it does not work!

src/native.js
```diff
@@ -197,6 +217,28 @@ export function getAssetMessage(data, adObject) {
...
+    message: 'assetResponse',
+    adId: data.adId,
+    assets: []
+  };
+
+  Object.keys(adObject.native).forEach(function(key, index) {
+    if (key === 'adTemplate' && adObject.native[key]) {
+      message.adTemplate = getAssetValue(adObject.native[key]);
+    } else if (key === 'rendererUrl' && adObject.native[key]) {
+      message.rendererUrl = getAssetValue(adObject.native[key]);
```
robertrmartinez, line 231: So these values are being sent through to the top level of the message, not within the assets array?

What is the reason for this?

src/native.js
wqi1972: second @robertrmartinez 's solution. And not only this version cleaner, also that version has logic issue. I believe we suppose keep existing structure `mediaType.native`, but the code is looking for `nativeParams `, which is a new structure.

wqi1972: If I am not wrong, the solution misses a use case for `mediaType.native.keysToSend`.

test/spec/native_spec.js
```diff
@@ -87,6 +87,136 @@ describe('native.js', function () {
     ]);
   });
 
+  it('should only include targeting that has sendTargetingKeys set to true', function () {
+    const bidRequest = {
+      nativeParams: {
+        image: {
+          required: true,
+          sizes: [150, 50]
+        },
```
wqi1972, line 96: Native bid request is not in this format, it suppose be in format
```
  mediaTypes: {
        native: {
          body: { },
          clickUrl: { },
        }
```

src/native.js
```diff
@@ -154,22 +154,36 @@ export function fireNativeTrackers(message, adObject) {
...
+  if (deepAccess(bidReq, 'nativeParams.rendererUrl')) {
+    bid['native']['rendererUrl'] = getAssetValue(bidReq.nativeParams['rendererUrl']);
+  } else if (deepAccess(bidReq, 'nativeParams.adTemplate')) {
+    bid['native']['adTemplate'] = getAssetValue(bidReq.nativeParams['adTemplate']);
+  }
 
-    if (key && value) {
-      keyValues[key] = value;
+  Object.keys(bid['native']).forEach(asset => {
+    if (asset !== 'adTemplate') {
```
wqi1972, line 164: I believe it suppose be 
```
if (asset !== 'adTemplate' && asset !== 'rendererUrl') {
```

test/spec/native_spec.js
robertrmartinez: @wqi1972 

I believe prebid.js core does a translation where it adds this `nativeParams` object onto the bidRequest.

So you can either grab it from there, or the way you describe.

wqi1972: got it, thanks for explanation.

src/native.js
robertrmartinez: In the consolidated spec we left this out, as the idea was the most often use cases will be send all or none.

I believe others agreed, but if everyone else thinks so, we can add it in.

I would rather there be only one or the other though.

Either each asset can have the `sendTargetingKeys` or at top level there is a `keysToSend` 

I do not think it makes sense to have the code written to handle both.

src/native.js
mmoschovas: I placed them at the top level instead of the assets as the assets are looped through and replace their corresponding macro with a value. These two will never need use that functionality and if exist are simply utilized for the replacement themselves. I figured separating them made more sense, but if opposed, they can be added to the assets with other logic in place to detect in the PUC.

src/native.js
mmoschovas: It was my understanding that rendererUrl can be sent as a KV in the GAM call; whereas adTemplate will never be sent. Is this not the case?

src/native.js
mmoschovas: This logic is much cleaner thanks. Pushed an update for this.

src/native.js
wqi1972: From publisher point of view, I don't see the need to pass rendererUrl as a KV in GAM call. @jaiminpanchal27 , rendererUrl was originated from your team, what do you think about this?

src/native.js
robertrmartinez: Nope! Good idea!

src/native.js
wqi1972: I looked into the user case in detail, and realize rendererUrl could pass as key/value. In case when different ad unit configures different rendererUrl, but share same creatives inside GAM.

robertrmartinez:
Hey all, I am setting up some GAM line items today to help make testing local PUC changes easier.

I'll share example test pages once I am done today / tomorrow.

I would like to finalize the native stuff and merge both these for next weeks release!

Thanks again.

robertrmartinez:
@wqi1972 @jaiminpanchal27 

If you both could give your final review that would be great.

robertrmartinez:
We will merge this once the associated PUC review has been approved:

https://github.com/prebid/prebid-universal-creative/pull/106

jsnellbaker (approved):

bretg:
Docs PR https://github.com/prebid/prebid.github.io/pull/2479

allanjun (approved):

PR #5421, opened 2020-06-25
ozone 2.4.0 adapter updates

afsheenb:
Type of change
 • Bugfix
 • Feature
 
Description of changes

• TCF 2.0 Support
• Video request updated to support oRTB standards
• Refactored outstream implementation
• Introduced additional custom key-value pairs for video to indicate if bid is for instream or outstream
• introduced additional key-value pair to support floor price controls.
• Deprecated key-values pairs the adapter previously was setting that are no longer required.
• Introduced a pageview ID to request payload.
• 'lotameData' now sent inside ext.ozone instead of imp.ext.ozone to reduce request payload length.

- engineering@ozone-project.com

### DIFF
--- a/modules/ozoneBidAdapter.js
+++ b/modules/ozoneBidAdapter.js
@@ -4,19 +4,21 @@ import { BANNER, NATIVE, VIDEO } from '../src/mediaTypes.js';
 import {config} from '../src/config.js';
 import {getPriceBucketString} from '../src/cpmBucketManager.js';
 import { Renderer } from '../src/Renderer.js';
-
 const BIDDER_CODE = 'ozone';
 const ALLOWED_LOTAME_PARAMS = ['oz_lotameid', 'oz_lotamepid', 'oz_lotametpid'];
+
+// *** PROD ***
 const OZONEURI = 'https://elb.the-ozone-project.com/openrtb2/auction';
 const OZONECOOKIESYNC = 'https://elb.the-ozone-project.com/static/load-cookie.html';
 const OZONE_RENDERER_URL = 'https://prebid.the-ozone-project.com/ozone-renderer.js';
 
-const OZONEVERSION = '2.3.0';
-
+const OZONEVERSION = '2.4.0';
 export const spec = {
   code: BIDDER_CODE,
   supportedMediaTypes: [VIDEO, BANNER],
   cookieSyncBag: {'publisherId': null, 'siteId': null, 'userIdObject': {}}, // variables we want to make available to cookie sync
+  propertyBag: {'lotameWasOverridden': 0, 'pageId': null, 'buildRequestsStart': 0, 'buildRequestsEnd': 0}, /* allow us to store vars in instance scope - needs to be an object to be mutable */
+
   /**
    * Basic check to see whether required parameters are in the request.
    * @param bid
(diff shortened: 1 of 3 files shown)


AskRupert-DM:
hey @jsnellbaker - apologies for the random tag here, you helped in the past with a similar issue on a previous PR whereby my PR failed the circle CI tests due to an error unrelated to my adapter - I believe that's the case here - could you let me know please ?

jsnellbaker:
The failure does appear to be another set of tests rather than the ozone adapter's.  I have reran the job and it appears to have passed successfully.

AskRupert-DM:
Thanks for the help / looking into this @jsnellbaker - let us know if anything is needed with merging this.

modules/ozoneBidAdapter.js
```diff
@@ -140,14 +141,13 @@ export const spec = {
 
     if (bidderRequest && bidderRequest.gdprConsent) {
       utils.logInfo('OZONE: ADDING GDPR info');
-      ozoneRequest.regs = {};
-      ozoneRequest.regs.ext = {}; // setting default values in case there is no additional information, like for the Mirror
-      ozoneRequest.regs.ext.gdpr = bidderRequest.gdprConsent.gdprApplies ? 1 : 0;
+      let apiVersion = utils.deepAccess(bidderRequest.gdprConsent, 'apiVersion', '1');
```
robertrmartinez, line 144: At this point your above if statement assured `bidderRequest.gdprConsent` so instead of using deepAccess with a single notation apiVersion, a simple:

```
let apiVersion = bidderRequest.gdprConsent.apiVersion || '1';
```

May be a little cleaner.

modules/ozoneBidAdapter.js
```diff
@@ -140,14 +141,13 @@ export const spec = {
...
+      ozoneRequest.regs = {ext: {gdpr: bidderRequest.gdprConsent.gdprApplies ? 1 : 0, apiVersion: apiVersion}};
       if (ozoneRequest.regs.ext.gdpr) {
         ozoneRequest.user = ozoneRequest.user || {};
         ozoneRequest.user.ext = {'consent': bidderRequest.gdprConsent.consentString};
       } else {
-        utils.logInfo('OZONE: **** Failed to find required info for GDPR for request object, even though bidderRequest.gdprConsent is TRUE ****');
+        utils.logInfo('OZONE: **** Strange CMP info: bidderRequest.gdprConsent exists BUT bidderRequest.gdprConsent.gdprApplies is false. See bidderRequest logged above. ****');
```
robertrmartinez, line 150: Not necessarily strange behavior.

By default `gdprApplies` will be false in the scenarios where the CMP did not yet respond or errored out for any reason.

So this could in fact be a very familiar case.

Not against logging it, but gdpr can be confusing with all the different use cases and wanted to point it out.

modules/ozoneBidAdapter.js
```diff
@@ -234,38 +240,35 @@ export const spec = {
           obj.ext.ozone.customData = [{'settings': {}, 'targeting': {'oztestmode': ozTestMode}}];
         }
       } else {
-        utils.logInfo('no ozTestMode ');
+        utils.logInfo('OZONE: no ozTestMode ');
```
robertrmartinez, line 243: Just a heads up, there have been discussions around the amount of logging bidAdapters are allowed to include. Obviously these are only fired if the user has debug on, but strings cannot be minified. So the prebid package becomes quite large.

For now, no changes are necessary as we have not come up with a plan to actually enforce a logging amount. But wanted to let you know!

modules/ozoneBidAdapter.js
AskRupert-DM: thanks have noted this to be updated/cleaned in our next update - assume it will be okay to leave as is for now (keen to get this version out there).

modules/ozoneBidAdapter.js
AskRupert-DM: Noted - we'll look at cleaning up / removing unnecessary logs in a future version or maybe introducing error/logging codes to reduce the length of the strings which should help with keeping string lengths to a minimum.

AskRupert-DM:
Thanks for the feedback @robertrmartinez - I've commented back on some of the points you mention and have noted them as items to address on our backlog. 

Let us know if anything needs to be addressed immediately - but otherwise am keen to get this approved/merged if there's no major flags as publishers are awaiting for this to update their builds ahead of TCF 2.0.

modules/ozoneBidAdapter.js
AskRupert-DM: fair - most users who will enable logging should be aware that it maybe flagged due to the CMP not responding or timing out but will look at maybe dropping/improving this log in the future.

AskRupert-DM:
Hey @robertrmartinez - hope all is well - just checking in to see if you were able to complete the tests you wanted to conduct/if there's anything else we can do to assist ?

robertrmartinez:
Awesome test coverage guys! Love to see >90%.

Merging now

AskRupert-DM:
Thanks @robertrmartinez - noted the feedback for our next update.

PR #5427, opened 2020-06-26
drop support for userId configs with the `usersync` config object, pe…

smenzer:
## Type of change
- [ X ] Bugfix
- [ ? ] Does this change affect user-facing APIs or examples documented on http://prebid.org?

## Description of change
Per comments, and likely an early-on oversight, we were allowing two different `userSync` configurations to be made (based on case of the word usersync), although only one was supported at a time. This PR drops support for `usersync` and will require publishers to all use `userSync` when setting configurations for userId module.

I'm not sure if we need docs changes or not, but I think they already specify to use `userSync`

## Other information
See comment in current code specifying we will make this change: https://github.com/prebid/Prebid.js/blob/master/modules/userId/index.js#L573

### DIFF
--- a/modules/userId/index.js
+++ b/modules/userId/index.js
@@ -570,8 +570,8 @@ export function init(config) {
   }
   // listen for config userSyncs to be set
   config.getConfig(conf => {
-    // Note: support for both 'userSync' and 'usersync' will be deprecated with Prebid.js 3.0
-    const userSync = conf.userSync || conf.usersync;
+    // Note: support for 'usersync' was dropped as part of Prebid.js 4.0
+    const userSync = conf.userSync;
     if (userSync && userSync.userIds) {
       configRegistry = userSync.userIds;
       syncDelay = utils.isNumber(userSync.syncDelay) ? userSync.syncDelay : DEFAULT_SYNC_DELAY;
--- a/modules/appnexusBidAdapter.js
+++ b/modules/appnexusBidAdapter.js
@@ -63,7 +63,20 @@ const storage = getStorageManager(GVLID, BIDDER_CODE);
 export const spec = {
   code: BIDDER_CODE,
   gvlid: GVLID,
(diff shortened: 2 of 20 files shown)


modules/userId/index.js
```diff
@@ -570,8 +570,8 @@ export function init(config) {
   }
   // listen for config userSyncs to be set
   config.getConfig(conf => {
-    // Note: support for both 'userSync' and 'usersync' will be deprecated with Prebid.js 3.0
-    const userSync = conf.userSync || conf.usersync;
+    // Note: support for 'usersync' was dropped as part of Prebid.js 4.0
+    const userSync = conf.userSync;
```
patmmccann, line 574: LGTM, I confirmed no doc changes needed to http://prebid.org/dev-docs/publisher-api-reference.html#setConfig-Configure-User-Syncing

patmmccann: actually can you change the tests at test/spec/modules/userId_spec.js that have config usersync ?

patmmccann: also the examples at https://github.com/prebid/Prebid.js/blob/master/modules/userId/userId.md

patmmccann: also here https://github.com/prebid/Prebid.js/blob/master/modules/userIdTargeting.md

smenzer:
> @smenzer
> There are a couple other documentation files that use the `usersync` syntax. Could you update these other examples as well?
> 
> Below is what I found during my search through the repo:
> 
> * modules/britepoolIdSystem.md
> * modules/sharedIdSystem.md
> 
> I think we should be good to merge after this last update.

Sure thing! will do that first thing tomorrow morning

patmmccann:
> > @smenzer
> > There are a couple other documentation files that use the `usersync` syntax. Could you update these other examples as well?
> > Below is what I found during my search through the repo:
> > 
> > * modules/britepoolIdSystem.md
> > * modules/sharedIdSystem.md
> > 
> > I think we should be good to merge after this last update.
> 
> Sure thing! will do that first thing tomorrow morning

https://github.com/smenzer/Prebid.js/pull/2 should take care of it for you

smenzer:
thanks @patmmccann for the PRs!  @jsnellbaker this should be ready for you to merge in to the 4.0 branch.

PR #5436, opened 2020-06-29
Added support for Liveramp userId submodule & read sizes from adUnit.mediaTypes.banner 

ix-prebid-support:
## Type of change

- [ ] Bugfix
- [X] Feature
- [ ] Build related changes
- [ ] CI related changes
- [ ] Other

## Description of change
- Added support for Liveramp userId submodule
- The adapter now reads sizes from adUnit.mediaTypes.banner directly and send any missing size(s) found as a part of the bid request as well

### DIFF
--- a/modules/ixBidAdapter.js
+++ b/modules/ixBidAdapter.js
@@ -195,6 +195,39 @@ function getBidRequest(id, impressions) {
 }
 
 /**
+ * Adds a User ID module's response into user Eids array.
+ *
+ * @param  {array}  userEids       An array of objects containing user ids,
+ *                                 will be attached to bid request later.
+ * @param  {object} seenIdPartners An object with Identity partners names already added,
+ *                                 updated with new partner name.
+ * @param  {*}      id             The id obtained from User ID module.
+ * @param  {string} source         The URL of the User ID module.
+ * @param  {string} ixlPartnerName The name of the Identity Partner in IX Library.
+ * @param  {string} rtiPartner     The name of the User ID provider in Prebid.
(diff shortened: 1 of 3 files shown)


modules/ixBidAdapter.js
```diff
@@ -210,16 +243,30 @@ function buildRequest(validBidRequests, bidderRequest, impressions, version) {
   // Always use secure HTTPS protocol.
   let baseUrl = SECURE_BID_URL;
 
+  // Dict for identity partners already populated from prebid
```
patmmccann, line 246: This is a great development! I am curious however, why only liveramp? prebid might also populate liveintent, tdid and others your endpoint already understands. Maybe i am confused but this comment says partners and the code seems to only be for liveramp.

sulleh: We plan to add in additional partners, so we're just starting with LiveRamp.  We want to make sure we build out toggles on the server side, so DSPs can control which IDs get passed down to them (which is different than matching an ID against an audience) regardless of what gets consumed on the page by the IX adapter. 

We already have these ID DSP toggles for LR, so it was an easy starting point.  Over the next month or two, you'll see us consume the other IDs.

patmmccann: looking forward to that, ty!

ix-prebid-support:
hi @bmwcmw  @mike-chowla,

Any ETA on when this can be reviewed and merged?

modules/ixBidAdapter.js
patmmccann: this is great to see merged in! just to confirm @sulleh ; this seems to allow one to fully support roundel from within prebid?

PR #5444, opened 2020-07-01
TCFv2.0 Purpose 7

Fawke:
## Type of change
- [X] Feature

## Description of change
Introduces support to enforce TCFv2.0 Purpose 7 (measurement). For details on the spec, please refer to [this](https://docs.google.com/document/d/1fBRaodKifv1pYsWY3ia-9K96VHUjd8kKvxZlOsozm8E/edit) doc.

Basically, the idea is to block the analytics adapter that are not given consent. In actual implementation terms, it translates to not calling the `enableAnalytics` function of the `AnalyticsAdapter.js` interface.

#### A way to specify GVL ID for Analytics Adapters.

An analytics adapter needs to be associated with a Gvl ID in order to participate in the enforcement checks. If Gvl ID is `undefined`, then the Analytics adapter will automatically be disqualified and blocked.

Since, there is currently no specific way to provide Gvl ID for Analytics Adapters, this PR introduces a way to do so. When you are registering your analytics adapter in your `myAnalyticsAdapter.js` file, you can provide the GVL ID in the following way:

```(javascript)
adapterManager.registerAnalyticsAdapter({
  adapter: appnexusAdapter,
  code: 'appnexus',
  gvlid: 32
});
```
Pass it as a property along with `adapter` and `code`.

#### TCF2 Enforcement Results

An event, `tcf2Enforcement` will be emitted with the payload,

```(javascript)
{
  storageBlocked: ['moduleA', 'moduleB'],
  biddersBlocked: ['moduleB'],
  analyticsBlocked: ['moduleC']
}
```
which will contain information of all the necessary bidders and modules which are blocked by tcf2. One thing which this event payload does not capture is the enforcement status of Purpose 1, 2 and 7. Although, it can be derived from the payload above, for example, if biddersBlocked array is `[]`, that means purpose 2 is not enforced. But in some cases, it could be misleading. For examples, cases where even though purpose 2 is enforced, but all the bidders are given consent by the user or are present in `vendorExceptions` will give us an `[]`. Let me know if this detail neccessary, I can modify the payload to look something like:

```(javascript)
{
  purpose1: {
    enforced: true,
    modulesBlocked: ['moduleA', 'moduleB']
  },
  purpose2: {
    enforced: true,
    modulesBlocked: [] // maybe all bidders were given consent / or present in vendorExceptions
  },
  purpose7: {
    enforced: false,
    modulesBlocked: []
  }
}
```


bretg:
Good work @Fawke - my take is that your minimal analytics proposal is fine:

```
{
  storageBlocked: ['moduleA', 'moduleB'],
  biddersBlocked: ['moduleB'],
  analyticsBlocked: ['moduleC']
}
```

Analytics adapters can `getConfig({consentManagement})` to inspect the settings if they want more detail.

But will allow analytics providers to weigh in.

bretg:
docs PR https://github.com/prebid/prebid.github.io/pull/2192

Fawke:
> LGTM.
> Can you add one more test for auction end emit event

@jaiminpanchal27 Added!

Issue #5420, opened 2020-06-25
Intent to Implement: Changes to consent should trigger call to userId modules


smenzer:
## Type of issue
Intent to Implement

## Description

The userId module provides consent data to each idSystem when requesting a new/refreshed ID (via the `getId()` method). On subsequent page views, if the ID system does not need a refresh, the userId module will optionally make calls to the idSystem with `extendId()` method (if it exists), followed by a call to `decode()` to pull out the ID value for use, but neither of these methods contain any consent data, nor are they intended to be used to call the idSystem again for a new/refreshed ID. 

In some cases, the consent data may change from one page view to another (either there was no consent on the first request due to a CMP timeout; or a user changed their consent preferences via the CMP after the first page view). In this case today, the idSystem would not be aware that consent has changed and could be providing an invalid ID (since the stored object would be based on old consent data).

## Proposal

The userId module should be aware of changes in consent from one request to the next. If consent changes, then userId modules should all be called with a new `getId()` call instead of just `extendId()` and `decode()`. 

Presumably, this could be done by storing the consent string in local storage and checking the value on each request to see if it changed, and if so, force a refresh.

I'm open to other suggestions on the proposal or implementation if anyone has any ideas.

smenzer:
@olafbuitelaar fyi

pycnvr:
This approach ends up generating a new pubcommonid every time the consent string changes.  That's problematic.  Since only id5 applies additional consent checks in its getId call and has also implemented extendId, it's probably better to pass along consent data in extendId call so that id5, and other subsystems in the future, can applies specialized consent checks when needed.

smenzer:
@pycnvr if consent changes, though, isn’t it important to update the pubcommonid in case there should no longer be one? we're also not talking about a common occurrence, this is to be covered in case of changes, which is important.

patmmccann:
linking https://github.com/prebid/Prebid.js/pull/5451 as the issue that closed

PR #5451, opened 2020-07-02
user id module refresh ids when consent changes

smenzer:
## Type of change
- [ X ] Feature

## Description of change
Per https://github.com/prebid/Prebid.js/issues/5420, this is an attempt to force a refresh of IDs when consent changes. To handle this, the consent received when entering the userId module is stored in a cookie, which is then used to compare on the next call to the userId module. If the stored consent is different from the current consent, the `getId` method will be called, regardless of whether a refresh was needed or not. If there is nothing stored in the cookie, though, the `getId` call is not forced - the intention here is that we don't want to force a refresh on every user as soon as this code is rolled out. After the first call with the new code, the current consent is stored, so after that it can be properly compared.

### DIFF
--- a/test/spec/modules/userId_spec.js
+++ b/test/spec/modules/userId_spec.js
@@ -1546,6 +1546,7 @@ describe('User ID', function() {
 
       // init consent management
       window.__cmp = function () { };
+      delete window.__tcfapi;
       testConsentData = {
         gdprApplies: true,
         consentData: 'xyz',
--- a/test/spec/modules/userId_spec.js
+++ b/test/spec/modules/userId_spec.js
@@ -1496,18 +1496,18 @@ describe('User ID', function() {
   });
 
   describe('Consent changes determine getId refreshes', function() {
-    const mockIdCookieName = 'MOCKID';
     let expStr;
     let adUnits;
+
+    const mockIdCookieName = 'MOCKID';
     let mockGetId = sinon.stub();
     let mockDecode = sinon.stub();
     let mockExtendId = sinon.stub();
-    let cmpStub;
-    let testConsentData;
-    const consentConfig = {
-      cmpApi: 'iab',
-      timeout: 7500,
-      allowAuctionWithoutConsent: false
+    const mockIdSystem = {
+      name: 'mockId',
+      getId: mockGetId,
+      decode: mockDecode,
+      extendId: mockExtendId
     };
     const userIdConfig = {
       userSync: {
@@ -1522,14 +1522,16 @@ describe('User ID', function() {
         auctionDelay: 5
       }
     };
-    const mockIdSystem = {
-      name: 'mockId',
-      getId: mockGetId,
-      decode: mockDecode,
-      extendId: mockExtendId
+
+    let cmpStub;
+    let testConsentData;
+    const consentConfig = {
+      cmpApi: 'iab',
+      timeout: 7500,
+      allowAuctionWithoutConsent: false
     };
 
-    beforeEach(function () {
+    const sharedBeforeFunction = function() {
       // clear cookies
       expStr = (new Date(Date.now() + 25000).toUTCString());
       coreStorage.setCookie(mockIdCookieName, '', EXPIRED_COOKIE_DATE);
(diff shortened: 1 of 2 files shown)


test/spec/modules/userId_spec.js
```diff
@@ -1412,4 +1423,191 @@ describe('User ID', function() {
...
+
+      // init
+      adUnits = [getAdUnitMock()];
+      init(config);
+
+      // init id system
+      attachIdSystem(mockIdSystem);
+      config.setConfig(userIdConfig);
+
+      // init consent management
```
jaiminpanchal27, line 1476: Adding `delete window.__tcfapi;` should fix failing unit tests.

smenzer: ugh...i thought about trying to make sure that was removed, but i thought that since i wasn't checking v2 it was fine. but this definitely works, thanks.

smenzer: i'm actually considering changing the tests to use v2 instead, since that will be what's used longer term. any thoughts? should i support both?

smenzer: i went ahead and added support for tcf 2

smenzer:
@jaiminpanchal27 this is ready for your review/merge whenever you get a chance

smenzer:
@robertrmartinez anything you need (besides time) to finish the review for this one? happy to provide more clarity offline if necessary. thanks!

muuki88:
We rolled out this change an received a lot of http status 431 from users as the cookie header grow beyond the normal limit of 8kbytes. The main driver is our vendorlist, which contains all currently listed IAB vendors (version 51). We know from other big publishers that they do the same as there is no sane or safe way to figure out which partners are actually required.

It would be nice to opt-out of this feature. Partners like id5 refresh their cookies all 8hours and users that change their consent preference are rare as well as vendorlist updates as they force all users to reconsent.

smenzer:
@muuki88 This feature only forces a refresh if consent changes. If consent does not change, then refreshes are handled based on the normal process of the userId module and the specifics of the submodule. If consent changes, it's important for identifiers to know that so they can update accordingly, since they may no longer have consent or may need to re-encrypt the ID based on the updated consent.

muuki88:
Thanks for the fast repsonse @smenzer :smiley: 

I understand what the module does and why it's important. The issue is how the consent data is being stored. 

With a long vendorlist this additional cookie takes a lot of space in the cookie header, which causes quite a few users of our sites to receive 431 http status as the `cookie` header is too large.

So my suggestion would be to make the storage engine configurable (localstorage or cookie).

On a side note: Safari doesn't make use of this feature as 1st party client side cookies are also purge. At least this is how I understand [the Intellligent tracking preventions](https://webkit.org/blog/9521/intelligent-tracking-prevention-2-3/).

muuki88:
IMHO it would be nice to trigger this id refresh via an API. Most CMPs have hooks when the consent changes. So this behaviour can be enabled or triggered manually. Like the `triggerUserSync()` API. WDYT?

smenzer:
Hi @muuki88 are you saying that on calls to your own servers (i.e. the first party domain), the cookies being stored by this PR are being included in the request, thus causing the 431's from your https responses (as opposed to responses from the identity providers or other ad tech platforms)?

perhaps instead of storing the full string, I could just create a hash (like `md5` or something similar) and store that...would be a much smaller string and shouldn't cause problems. thoughts?

I hadn't looked into cmp hooks, but I suppose that's another option, although I don't see anything specific about changes in consent...can you point me to the specifics, please? I looked here: https://github.com/InteractiveAdvertisingBureau/GDPR-Transparency-and-Consent-Framework/blob/master/TCFv2/IAB%20Tech%20Lab%20-%20CMP%20API%20v2.md#addeventlistener

muuki88:
Hi @smenzer 

Indeed. The cookies are sent to the server.

I like the idea of an md5 hash 🤯👍. That should be way smaller and serve the same purpose.

Regarding the update hooks. Sourcepoint and LiveRamp both have the ability. And if I'm not mistaken the `addEventListener` should also work for this.

smenzer:
I've put in a PR to correct this - https://github.com/prebid/Prebid.js/pull/5641

Issue #5440, opened 2020-06-30
CCPA support for SPA or refresh ad units


ronzilca:
## Type of issue
Feature (it's a bug, I think, but according to the code, it was planned this way).

## Description
Within CCPA states, the default _usprivacy_ string is 1YNN.
But, users can choose to opt-out and avoid using cookies. Then, the _usprivacy_ string becomes 1YYN.

The problem is when I want to refresh ad units, according to the _consentManagementUsp_, the _usprivacy_ won't update.

## Steps to reproduce
1. Enable the usp consent module:
```
                pbjs.setConfig({
                    'consentManagement' : {
                        'usp'               : {
                            'cmpApi'                    : 'iab',
                            'timeout'                   : 100
                        }
                    }
                });
```

2. Change the _usprivacy_ cookie (1YYN to opt-out).
3. Ask the __uspapi for the current status:
`__uspapi('getUSPData', 1, (a, b) => { console.log(a, b) });`
4. Refresh the ad units

### Expected results
The new bid request should contain the current _usprivacy_ string.

### Actual results
The new bid request contains the old _usprivacy_ string.

## Other information
![image](https://user-images.githubusercontent.com/2135285/86152513-f56e4b00-bb08-11ea-8b8c-a9a5dbe08843.png)

patmmccann:
According to issue #4425 it was not planned this way:

"Because the USP API does not wait for a user response, if it was not successfully obtained before the first auction, we should try again to retrieve privacy data before each subsequent auction." 

and this indeed appears to be a bug.

patmmccann:
should be solved by #5452

PR #5452, opened 2020-07-02
Update consentManagementUsp.js to not store and reuse null consent

patmmccann:
Fixes bug described at https://github.com/prebid/Prebid.js/issues/5440 to comply with intent of #4425
## Type of change

- [X] Bugfix
- [ ] Feature
- [ ] New bidder adapter   
- [ ] Build related changes
- [ ] CI related changes
- [ ] Other

## Description of change
No longer checks to see if consentData are stored; should recheck _uspapi


patmmccann:
of note, build failed because of bad merge https://github.com/prebid/Prebid.js/commit/d8e5796827a46455185292e4a498628ecdb09bc6

jsnellbaker:
On an additional note - I retested the browserstack tests locally and they passed fine.  I'm not sure why circleci tests are failing.  It seems like it's trying to use a previous commit where a `flatMap` was added for the amxBidAdapter, but the config/stats for the job appear to be pointing to the most recent commit (where the `flatMap` was removed).  I think we can just ignore it unless someone has another idea.

patmmccann:
Thanks so much for all the education on yesterday's call. You're right, the
circleci fail is from a reverted amx adapter commit.

bretg:
@harpere is out right now -- assigning over to @mkendall07

mkendall07:
This change looks good if it's the behavior we want. The only slight caveat here is this part
>the USP (CCPA) api function __uspapi() always responds synchronously, whether or not privacy data is available, while the GDPR CMP may respond asynchronously, especially if waiting for a user response. **Unfortunately this doesn't necessarily mean it's always a synchronous transaction because if there are nested cross-domain iframes we may need to do an async postMessage().**
Note the bolded part, it's possible this puts an async delay into every auction request. As long as we are ok living with that it's a LGTM.

mkendall07:
@bretg @patmmccann for weigh in ^^

patmmccann:
I don't think we have a choice, usp is opt out and if someone opts out at
any point, refreshes and subsequent auctions need to know. If I were a
regulator and wanted to test for compliance, I would go to a page, opt out,
and look at the subsequent calls.

mkendall07:
makes sense. It looks like circleci is continuously failing. Is this branch up to date with master?

patmmccann:
there was a bad merge to master of the amx adapter that was later fixed

patmmccann:
@mkendall07 it is up to date with master now

bretg:
Re-ran the build and it passed. Since this is a bug, I'll assume that we're not going to need documentation. Will merge and add it to the rel notes.

PR #5466, opened 2020-07-07
oneVideo Adapter End2End Testing Parameter Support (SAPR-13942) 

adam-browning:
## Type of change
- [ x] Feature

## Description of change
oneVideoBidAdapter now supports a new parameter to force outbound bid-request settings for self severed testing By passing "bid.params.video.e2etest = true".
**Important**: 'e2etest: ture` is a sumpliment to the mandatory parameters, it cannot replace them.

When applying `e2etest: ture` the adapter will switch to testing mode and send an outbound bid-request to our test our US west coast server "HBExchange" Exchange endpoint 
The feature overrides base settings like ENDPOINT, width, height, pubid, page, referrer to minimize testing costs and time on the publishers end.

### Test parameters for validating bids
```
var adUnits = [
    {
        code: 'video1',
          mediaTypes: {
            video: {
                  context: 'instream',
                  playerSize: [480, 640]
            }
          },
          bids: [
            {
              bidder: 'oneVideo',
              params: {
                video: {
                  playerWidth: 480,
                  playerHeight: 640,
                  api: [2],
                  e2etest: true
                },
                pubId: 'Whatever'
                }
            }
          ]
      }
  ]
```

## Other information
Hi @DeepthiNeeladri,
Could you please review this feature
Thank you!
Adam


adam-browning:
AFAIK, the failed ci/circleci: build test is Not related to the changes of this PR
Failed spec id: https://github.com/prebid/Prebid.js/blob/master/test/spec/modules/prebidServerBidAdapter_spec.js#L1990

DeepthiNeeladri:
@adam-browning LGTM

adam-browning:
All tests and checks pass after pulling upstream and merging master today

modules/oneVideoBidAdapter.js
```diff
@@ -282,7 +289,16 @@ function getRequestData(bid, consentData, bidRequest) {
       bidData.regs.ext.us_privacy = bidRequest.uspConsent
     }
   }
-
+  if (bid.params.video.e2etest) {
+    bidData.imp[0].bidfloor = null;
+    bidData.imp[0].video.w = 300;
+    bidData.imp[0].video.h = 250;
```
jsnellbaker, line 295: To clarify - should these be hard-coded?  Or should they be data-points about the current page's setup?

modules/oneVideoBidAdapter.js
```diff
@@ -282,7 +289,16 @@ function getRequestData(bid, consentData, bidRequest) {
...
+    bidData.imp[0].video.mimes = ['video/mp4', 'application/javascript'];
+    bidData.imp[0].video.api = [2];
+    bidData.site.page = 'https://verizonmedia.com';
+    bidData.site.ref = 'https://verizonmedia.com';
```
jsnellbaker, line 299: Same question as above

modules/oneVideoBidAdapter.js
adam-browning: Hi @jsnellbaker,
The e2etest param is supposed to enforce/override the client parameters to ensure that the outbound bid-request will receive the test creative response from our oneVideo ad-server.
Until now, when we wanted to do e2e testing with a partner we were dependent on them to implement several changes to the adUnit config. In most cases, the turnaround was slow as either typos would appear, incorrect values and the whole back and forth is time consuming.
By forcing these parameters the can test their Player integrations in a self served manner and reduce the time to go live.

modules/oneVideoBidAdapter.js
adam-browning: @jsnellbaker same as above.
Here our ad-server would block domain traffic that is not declaring our platform in the publishers Ads.txt.
By forcing the site and ref we can start testing right away without being dependent on the ad-server Ads.txt crawler logic.

jsnellbaker:
@adam-browning 

Thanks for the confirmation.  When you get a chance, can you please take a look at the new conflicts and resolve them?  They're likely the result from the other PR just getting merged.

We should be good to merge this PR once the conflicts are cleaned up.

adam-browning:
Hi @jsnellbaker,
I have resolved the conflicts so we should be good to go.
Thanks for your help

adam-browning:
Hi @jsnellbaker,
When you get a chance, could you please review the final changes.
Thank you for your help!

PR #5467, opened 2020-07-08
Add Inmar bidder adapter

zandree-owneriq:
## Type of change

- [x] New bidder adapter   

## Description of change

Add Inmar bidder adapter

```
{
  bidder: 'inmar',
  params: {
    pid:"ADb1f40rmo",
    supplyType:"site",
    bidfloor:0.40,
  }
}
```

oiq_rtb@inmar.com

https://github.com/prebid/prebid.github.io/pull/2118

### DIFF
--- a/modules/inmarBidAdapter.js
+++ b/modules/inmarBidAdapter.js
@@ -1,0 +1,152 @@
+import { config } from '../src/config.js';
+import { registerBidder } from '../src/adapters/bidderFactory.js';
+import { BANNER, VIDEO } from '../src/mediaTypes.js';
+
+const BIDDER_CODE = 'inmar';
+
+export const spec = {
+  code: BIDDER_CODE,
+  aliases: ['inm'],
+  supportedMediaTypes: [BANNER, VIDEO],
+
+  /**
+   * Determines whether or not the given bid request is valid
+   *
+   * @param {bidRequest} bid The bid params to validate.
+   * @returns {boolean} True if this is a valid bid, and false otherwise
+   */
+  isBidRequestValid: function(bid) {
+    return !!(bid.params && bid.params.pid && bid.params.supplyType);
+  },
+
+  /**
+   * Build a server request from the list of valid BidRequests
+   * @param {validBidRequests} is an array of the valid bids
+   * @param {bidderRequest} bidder request object
+   * @returns {ServerRequest} Info describing the request to the server
+   */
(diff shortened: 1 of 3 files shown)


zandree-owneriq:
"gulp test" was run locally and everything passed.  Error message from CircleCI is:

should set the default bidResponse currency when not specified in OpenRTB

"currency" in interpretResponse is being handled the exact same way that Slimcut, SmileWanted, Richaudience, and Smartadserver adapters handle it, so I'm confused as to how following convention has produced an error.  Please advise.

musikele:
@zandree-owneriq any news?

zandree-owneriq:
There are two things that would be very helpful in implementing our adapter.  Firstly, is there documentation on the exact format of the validBidRequests and bidderRequest arguments that are passed in to the buildRequests function?  I've been using http://prebid.org/dev-docs/bidder-adaptor.html, but much of the information on that page is vague, incomplete, and occasionally contradictory.  Very little of the format is fully specified with name, type, description, and examples, and even that is almost entirely concerning the return data, not the input parameters.  Is there more detailed specification of the input parameters for the buildRequests function?

Secondly, how do the IDs retrieved in getUserSyncs get applied to the bid request?  I don't see any mention of how to use the user sync data in the buildRequests function.

musikele:
@zandree-owneriq, 
Regarding your first question, There are two main sources of informations, the link you provided and other bidder source modules. In case you need more help, there are a lot of ways to get in touch with the prebid community, all of which are monitored. Here it is the link: https://docs.prebid.org/support/ 

Regarding the second question: you're probably mixing two things here. 
When a new user arrives on the website: 
- prebid is loaded. cookies are empty
- on the very first auction, only userId that come from one of the userId modules will be present in the object `bidRequest.userId`. You can have a look at the UserID modules here: https://docs.prebid.org/dev-docs/modules/userId.html 
- after the first bid, the cookie sync process starts. This is needed to sync the various cookies between partners. These cookies are then stored under the domain name of the website as first party. 
- Then, with every subsequent request, the synced userid cookies should be sent to the bidder via cookie 
the process i just outlined may change depending on gdpr, ccpa, custom setups, and much more, but this should be the default. So, I suspect you want to get one of the userId from the user Id modules, not cookies from cookie syncs. 

Third: the problem i raised on my previous comments is that your bidder is under a CORS policy that prevents prebid to work. By unblocking the endpoint, debugging any issues would be easier.

PR #5471, opened 2020-07-08
Appnexus Bid Adapter -  rtb.trackers field optional

Fawke:
## Type of change
- [X] Other

## Description of change
 Makes `rtb.trackers` field optional in the Appnexus Bid Adapter.

### DIFF
--- a/modules/appnexusBidAdapter.js
+++ b/modules/appnexusBidAdapter.js
@@ -645,7 +645,7 @@ function newBid(serverBid, rtbBid, bidderRequest) {
       ad: rtbBid.rtb.banner.content
     });
     try {
-      const url = rtbBid.rtb.trackers[0].impression_urls[0];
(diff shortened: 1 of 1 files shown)


modules/appnexusBidAdapter.js
```diff
@@ -645,7 +645,7 @@ function newBid(serverBid, rtbBid, bidderRequest) {
       ad: rtbBid.rtb.banner.content
     });
     try {
-      const url = rtbBid.rtb.trackers[0].impression_urls[0];
+      const url = rtbBid.rtb.trackers ? rtbBid.rtb.trackers[0].impression_urls[0] : null;
```
jaiminpanchal27, line 648: Why to create tracking pixel if trackers is null. Can you add this check to whole try block

Fawke: @jaiminpanchal27 Done!

PR #5474, opened 2020-07-09
ATS-analytics - fix check for safari browser

mamatic:
## Type of change

- [x] Bugfix

## Description of change
Safari is reporting a value of “false” for userBrowser in ATS Analytics. 

## Other information

@bretg

### DIFF
--- a/test/spec/modules/atsAnalyticsAdapter_spec.js
+++ b/test/spec/modules/atsAnalyticsAdapter_spec.js
@@ -146,5 +146,44 @@ describe('ats analytics adapter', function () {
       expect(atsAnalyticsAdapter.context.host).to.equal(initOptions.host);
       expect(atsAnalyticsAdapter.context.pid).to.equal(initOptions.pid);
     })
+    it('check browser is not safari', function () {
+      let browser = browserIsSafari();
+      expect(browser).to.equal(false);
+    })
+    it('check browser is safari', function () {
+      Object.defineProperty(window.navigator, 'vendor', {
+        value: 'Apple Computer, Inc.',
+        writable: true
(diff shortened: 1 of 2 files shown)


test/spec/modules/atsAnalyticsAdapter_spec.js
```diff
@@ -2,7 +2,7 @@ import atsAnalyticsAdapter from '../../../modules/atsAnalyticsAdapter.js';
 import { expect } from 'chai';
 import adapterManager from 'src/adapterManager.js';
 import {server} from '../../mocks/xhr.js';
-import {browserIsChrome, browserIsEdge, browserIsFirefox, browserIsSafari} from '../../../modules/atsAnalyticsAdapter.js';
+import {checkUserBrowser, browserIsChrome, browserIsEdge, browserIsSafari} from '../../../modules/atsAnalyticsAdapter.js';
```
bretg, line 5: Is there a reason for not testing browserIsFirefox()?

mamatic: Hi, yes, I didn't manage to mock InstallTrigger. We use this check to determine if user browser is firefox                     
(typeof InstallTrigger !== 'undefined'). 

https://developer.mozilla.org/en-US/docs/Web/API/InstallTrigger
https://stackoverflow.com/questions/9847580/how-to-detect-safari-chrome-ie-firefox-and-opera-browser/9851769

mamatic:
Hi @bretg,
I manage to write tests for Firefox. Manage to mock InstallTrigger. After a little research, global.InstallTrigger do a job.

PR #5480, opened 2020-07-10
Performax adapter

VasekProchazka:
## Type of change
- [x] New bidder adapter

## Description of change
Add a new bidder adapter for Performax

```
{
  bidder: '<bidder name>',
  params: {
    slotId: "28"
  }
}
```

- contact email of the adapter’s maintainer: development@performax.cz

### DIFF
--- a/modules/performaxBidAdapter.js
+++ b/modules/performaxBidAdapter.js
@@ -1,4 +1,4 @@
-import {logWarn, logInfo} from '../src/utils.js';
+import {logWarn} from '../src/utils.js';
 import {registerBidder} from '../src/adapters/bidderFactory.js';
 
 const CLIENT = 'hellboy:v0.0.1'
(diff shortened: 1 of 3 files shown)


VasekProchazka:
Docs PR https://github.com/prebid/prebid.github.io/pull/2123

modules/performaxBidAdapter.js
```diff
@@ -0,0 +1,78 @@
...
+  onBidWon: function (bid) {
+    // The onBidWon function will be called when a bid from the adapter won the auction.
+    logInfo('performax.onBidWon', bid);
+  },
+
+  onSetTargeting: function (bid) {
+    // The onSetTargeting function will be called when the adserver targeting
+    // has been set for a bid from the adapter.
+    logInfo('performax.onSetTargeting', bid);
+  }
```
FilipStamenkovic, line 76: If you don't plan to use these functions, you can remove them.  
You are only logging that these functions are called, you don't have to do that.  
You don't have to worry about prebid throwing an exception because it didn't find some of the mentioned functions.

VasekProchazka: Thanks for response. We will use these functions in the future. For now, I just deleted them as you advice. I also replace quotes. I hope it will be ok now:)

Issue #5458, opened 2020-07-06
Advanced Size Mapping Module bug empties config on refresh


benjaminclot:
## Type of issue
When a config has an empty `sizeConfig.sizes` property set, `[]` becomes `[[]]` so on the next refresh, this adunit's sizeConfig becomes invalid and it gets "deleted". I have been able to trace the problem back to this test: https://github.com/prebid/Prebid.js/blob/master/modules/sizeMappingV2.js#L109

## Test page
https://jsfiddle.net/Lmw0zrp5/
Have the window below 768px and hit refresh, `pbjs.adUnits[0]` will have its `bids` and `mediaTypes` properties empty.

### Expected results
The `pbjs.adUnits` object should remain untouched or maybe the test should use `utils.isEmpty()`?

Fawke:
Hi @benjaminclot,

Yeah, I see can that I'm getting an error on clicking "REFRESH" on your test page. Thanks for bringing this up, I think I've come up with a fix, and am working on some other optimizations, will raise a PR for this fix shortly.

PR #5487, opened 2020-07-13
Advanced Size Mapping - Code Refactoring.

Fawke:
## Type of change
- [X] Bugfix

## Description of change
- `checkAdUnitSetup` in `src/prebid.js` has been refactored to not mutate the `adUnit` object. Similarly, helper functions such as `validateBannerMediaType`, `validateVideoMediaType` and `validateNativeMediaType` have been refactor to avoid mutation of the adUnit object.

- A similar function, `checkAdUnitSetupHook` in `modules/sizeMappingV2.js` is also refactored to avoid mutation of `adUnit` object.
- `sizeConfig` validation checks has been moved to a single function, `validateSizeConfig` which returns a boolean, and does not mutate the `adUnit` object.

- Re-write some of the comments in `modules/sizeMappingV2.js` to make them more descriptive.

- Re-write / shorten / remove some log messages in `modules/sizeMappingV2.js`

- A test has been skipped in `pubCommonId_spec.js` file. The test is failing, due to reasons not related to this change, as far as I know. I'll request someone from `pubCommonId` to take a look.

- Resolves https://github.com/prebid/Prebid.js/issues/5458. This bug was caused because we were mutating the adUnit object inside the `checkAdUnitSetupHook` function in `modules/sizeMappingV2.js` file.

### DIFF
--- a/modules/sizeMappingV2.js
+++ b/modules/sizeMappingV2.js
@@ -139,7 +139,6 @@ export function checkAdUnitSetupHook(adUnits) {
             // Either 'sizes' or 'playerSize' is not declared as an array, which makes it invalid by default.
             isValid = false;
             showError = true;
-            return;
           }
           if (showError) {
             utils.logError(`Ad unit ${adUnitCode}: Invalid declaration of '${propertyName}' in 'mediaTypes.${mediaType}.sizeConfig[${index}]'. ${conditionalLogMessages[mediaType]}`);
--- a/modules/sizeMappingV2.js
+++ b/modules/sizeMappingV2.js
@@ -1,6 +1,7 @@
 /**
- * This modules adds support for the new Size Mapping spec described here. https://github.com/prebid/Prebid.js/issues/4129
- * This implementation replaces global sizeConfig with a adUnit/bidder level sizeConfig with support for labels.
+ * This module adds support for the new size mapping spec, Advanced Size Mapping. It's documented here. https://github.com/prebid/Prebid.js/issues/4129
+ * The implementation is an alternative to global sizeConfig. It introduces 'Ad Unit' & 'Bidder' level sizeConfigs and also supports 'labels' for conditional
+ * rendering. Read full API documentation on Prebid.org, http://prebid.org/dev-docs/modules/sizeMappingV2.html
  */
 
 import * as utils from '../src/utils.js';
@@ -8,11 +9,9 @@ import { processNativeAdUnitParams } from '../src/native.js';
 import { adunitCounter } from '../src/adUnits.js';
 import includes from 'core-js-pure/features/array/includes.js';
 import { getHook } from '../src/hook.js';
-import {
-  adUnitSetupChecks
-} from '../src/prebid.js';
+import { adUnitSetupChecks } from '../src/prebid.js';
 
-// allows for sinon.spy, sinon.stub, etc to unit test calls made to these functions internally
+// Allows for stubbing of these functions while writing unit tests.
 export const internal = {
   checkBidderSizeConfigFormat,
   getActiveSizeBucket,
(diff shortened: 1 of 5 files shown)


src/prebid.js
```diff
@@ -448,6 +461,8 @@ $$PREBID_GLOBAL$$.requestBids = hook('async', function ({ bidsBackHandler, timeo
 
   utils.logInfo('Invoking $$PREBID_GLOBAL$$.requestBids', arguments);
 
+  adUnits = checkAdUnitSetup(adUnits);
```
Fawke, line 464: Moving call to checkAdUnitSetup before we calculate the value of `adUnitCodes`. This is required because if some adUnit is removed in the function ,`checkAdUnitSetup`, then, the code below will incorrectly include the deactivated adUnit's code in the `adUnitCodes` variable.

benjaminclot:
@Fawke why do you use "caching" (`cacheHits`)? Won't that cause a problem when somebody triggers an ad refresh after a window resize?

Fawke:
> @Fawke why do you use "caching" (`cacheHits`)? Won't that cause a problem when somebody triggers an ad refresh after a window resize?

@benjaminclot Caching is an internal implementation detail which should not effect page refreshes after a window resize. To give you an overview, caching stores a data structure which is mapped to an `auctionId`.

```(javascript)
[{
  'auction-id-1': {
    adUnits: [
      <<contains details about adunits after size mapping checks>>
    ],
    cacheHits:1
  }
}]
```
This is done because we don't want to do sizeMapping processing for each bidder separately. It's done for the first bidder, stored in the cache, the next bidder picks up the details from this object and we increment the `cacheHits`.

If you refresh the page, it'll be a new auction (a new call to `requestBids`), which won't read data from the old auction, but create a new object. So, I think, it should be fine.

benjaminclot:
@Fawke I was talking about the next call to `pbjs.requestBids()` but you're right, as long as the cache is linked to the auctionId, we'll be fine.

bretg:
@Fawke - any docs updates?

Fawke:
@bretg Not really, no changes to the API.

bretg:
Cool - with there aren't changes to the interface, I'll remove myself from the reviewers.

modules/sizeMappingV2.js
```diff
@@ -74,136 +78,172 @@ export function isUsingNewSizeMapping(adUnits) {
...
-  return adUnits.filter(adUnit => {
+  const validateSizeConfig = function (mediaType, sizeConfig, adUnitCode) {
+    let isValid = true;
+    const associatedProperty = {
+      banner: 'sizes',
+      video: 'playerSize',
+      native: 'active'
+    }
+    const propertyName = associatedProperty[mediaType];
+    const conditionalLogMessages = {
```
robertrmartinez, line 97: 👍 on the maps based off the mediaType, cool stuff

test/spec/modules/pubCommonId_spec.js
```diff
@@ -234,7 +234,7 @@ describe('Publisher Common ID', function () {
       });
     });
 
-    it('disable auto create', function() {
+    it.skip('disable auto create', function() {
```
robertrmartinez, line 237: Why are we skipping this?

Fawke: If we don't skip this test, [this](https://github.com/prebid/Prebid.js/blob/1e157d4c4f9f77b98fa90ab1a7fb93015ae80d5c/test/spec/modules/pubCommonId_spec.js#L296) fails!

What I could gather from this is that the test that I have skipped, it disables autocreate of `bid.crumbs.pubcid`. Then, later, this configuration is not cleared which is leading the the I mentioned above to fail!

If I find time later, will try to dig deep to ascertain the cause of the failure, but for now, everything seems to be working fine, we can merge this!

Issue #5239, opened 2020-05-14
Proposed Prebid.js Module Rules


bretg:
The Prebid Code-of-Conduct sub-committee has [drafted an enhanced set of rules](https://docs.google.com/document/d/1kK0ifXV-vzyV7VchPKoyGuz375H7fzbTJFt_oWeCiRo/edit#heading=h.m33q7xm88e3q) governing what bid adapters, user ID sub-modules, and other types of modules can and can't do. We're doing this to preserve the trust that publishers have in Prebid as an important part of the independent ad tech stack.

The rules include old favorites like endpoints being HTTPS, but many new rules as well. 

Here's the process:

1. Community comment period. That starts now. Adapters and modules doing "unusual" things should pay special attention. e.g. recording loss-events, loading external files, loading a creative render script, etc
2. After being ratified by Prebid and the community, the rules will be documented on prebid.org and enforced from that point forward for new and changed modules.
3. Several rules will require changes in existing adapters/modules. A timeline will be set for each rule, probably something like "All adapters must adhere to this rule in 5.0 or it won't be allowed in."

We're looking forward to your comments, and especially your thumbs-up.

bretg:
As a result of a recent review, added one more:

```
Each module must include maintainer information with a group email address for Prebid.org to contact for support or maintenance.
```

Slind14:
Thank you for putting this up for community comments.

We are cutting out wasteful requests to specific bidders and want to integrate this on the adapter level. The classification is using machine learning. For optimal performance, this needs to run client-side where the model can be cached since it is doing a unique prediction for every single request.

The model is loaded from a CDN, the entire size is 10-20KB, the code is minified, the model is changing at least once per hour. The predictions happen in nanoseconds. The model only returns true/false and doesn't impose any direct risk.

This is directly tied to the adapter and its internal bidding behavior.

bretg:
@Slind14 - this was an anticipated usage of the ['real-time-data' module architecture](https://github.com/prebid/Prebid.js/pull/5519) -- the ability to optimize which bidders participate in a given auction given location, refresh count, number of bids, etc.

```
RTD modules may cache bid metadata for analytics and optimization purposes. They may use those bids to optimize future auctions for which bidders and adunits take place. They must not re-use specific past bids in an auction -- that functionality is restricted to Prebid core.
```
The idea is that an RTD module could listen to the `beforeRequestBids` event and modify the AdUnits based on the desired optimization algorithm.

[...]

PR #5488, opened 2020-07-13
Vidazoo Adapter: Feature/unique-deal-id

uditalias:
## Type of change
- [x] Feature

## Description of change

Sends unique deal id as a server param


uditalias:
@GLStephen LMK if I can assist with the merge of this PR. Thanks.

GLStephen:
@uditalias this looks good, but based on GDPR requirements you need to update your sessionStorage access to access through the utils storage class. Once that's updated I can review again. Thanks!

GLStephen:
@uditalias Rule 2 here covers this in the proposed rules https://github.com/prebid/Prebid.js/issues/5239

uditalias:
@GLStephen Thank you for you review.
Is there any way to use the sessionStorage instead of the localStorage? I found the storageManager but only the local storage is supported.
I think that the local storage is too aggressive for my feature because it will persist across site visits.

GLStephen:
@uditalias The setCookie and getCookie functions should be capable of creating session cookies.

uditalias:
@GLStephen I changed the code to use the `storageManager` from utils. Please note that I added another field param called `sessionId`, this field is extracted from the localStorage (using the storageManager) and passed as a param to the server. Thanks.

GLStephen:
@uditalias merged

Issue #4610, opened 2019-12-16
Extending the Real Time Data Module


bretg:
## Type of issue
Intent-to-Implement

## Description
We're planning to generalize the Browsi Viewability module into a more generic 'Real-Time Data' interface.

The general idea is that the Real Time Data infrastructure provides:

- a consistent external interface paradigm
- controlling the delay of the Prebid auction
- an internal interface to sub-modules that can evolve with requirements

RTD Sub-Modules are a diverse bunch. These are the known effects the RTD system needs to support:

1. adding ad server targeting variables to influence the ad server decision
2. adding values to the bidRequest to influence the auction
3. support ad servers other than GAM
4. inspect bidResponse values and modify them as needed
5. modify Prebid config (e.g. auction timeout)

Not all sub-modules need to delay the auction, so there should be a way for sub-modules to:

- tell the core that the doesn't need to wait on its account. If the core determines that no module needs to delay the auction, then the auction shouldn't be delayed.
- provide a default value should a timeout occur

## Proposed External Interface

The proposed interface of the Real-Time Data is fairly straight-forward:
```
    pbjs.setConfig({
        realTimeData: {
            auctionDelay: 100,    // global to all sub-modules
            dataProviders:[{          
                name: "SUB-MODULE-NAME",
                params: {
                    // sub-module specific parameters
                }
            }]
        }
    });
```

Each sub-module should be able to pass data to bid adapters on a bidrequest. e.g.
```
bidrequest[].realTimeData={"sub-mod-code":{},"sub-mod-2":{}}
```

## Next Steps

A technical team will review the current internal interface in the [real-time data module](https://github.com/prebid/Prebid.js/tree/master/modules/rtdModule) and propose how to extend it to support the additional use cases.

ydennisy:
Hey @bretg @pm-harshad-mane 
I would like to ask if you plan to continue work on this module?
It seems that `browsi` are already available through this method, we would like to our own module to the project - can you let me the process for doing so?
All the best,
Dennis

bretg:
@omerBrowsi is working on a proposal for a new internal interface for the Real Time Data module. Any estimate on the timing Omer? You can just clean up the Google Doc we're working on, move it to the Prebid.js docs GDrive, and share the link here for comment.

omerBrowsi:
@bretg 
Here is the document, I took the main points from our doc and organized a bit.
[...]

PR #5493, opened 2020-07-15
JWPlayer Targeting Module

karimMourra:
## Type of change
- [x] Feature

## Description of change
Adds the `jwplayerTargeting` module which allows Ad Bidders to access JW Player's Video Ad Targeting information.
The module can be used as a submodule for prebid adapters, allowing them to use our `getTargetingForBid()` function.
the module will fetch segments for the media ids present in the prebid config when the module loads. If any bid requests are made while the segments are being fetched, they will be blocked until all requests complete, or the 150ms timeout expires.


patmmccann:
Why build a new module type instead of utilizing an rtd submodule?

patmmccann:
delaying the auction for data seems like it should use the rtd module https://github.com/prebid/Prebid.js/issues/4610 and your functionality feels like a submodule. A bunch of unicorn modules that don't fall into module types with rules described on https://github.com/prebid/Prebid.js/issues/5239 feels like potentially an end run around the rules at worst, duplicate functionality at best.

modules/jwplayerTargeting.js
```diff
@@ -0,0 +1,178 @@
...
+  if (segments) {
+    return segments;
+  }
+
+  const player = getPlayer(playerID);
+  if (!player) {
+    return [];
+  }
+
+  const item = mediaID ? player.getPlaylist().find(item => item.mediaid === mediaID) : player.getPlaylistItem();
```
patmmccann, line 93: doesn't work on ie11

patmmccann:
@omerBrowsi what are your thoughts on this

karimMourra:
@patmmccann thank you for your feedback. Upon review of the rtdModule I think it does make sense for the JW Player module to be a submodule. I will close this PR for now. @omerBrowsi still looking forward to any feedback you may have.

PR #5494, opened 2020-07-15
mantis privacy support, removed unsupported media types

parisholley:

### DIFF
--- a/modules/mantisBidAdapter.js
+++ b/modules/mantisBidAdapter.js
@@ -270,7 +270,7 @@ export const spec = {
   }
 };
 
-export function sfPostMessage($sf, width, height, callback){
+export function sfPostMessage ($sf, width, height, callback) {
   var viewed = false;
   // eslint-disable-next-line no-undef
   $sf.ext.register(width, height, function () {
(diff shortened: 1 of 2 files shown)


modules/mantisBidAdapter.js
```diff
@@ -240,6 +241,11 @@ const spec = {
       }),
       property: property
     };
+
+    if (bidderRequest && bidderRequest.gdprConsent && bidderRequest.gdprConsent.gdprApplies) {
+      query.consent = false;
```
aleksatr, line 246: @parisholley what is the intent here, you don't want to collect consent string in GDPR zone (you're not part of the global vendor list)? You just mark this request as if consent was not given?

aleksatr:
hey @parisholley , for any adapter (new or updated) submission, code under review must have at least 80% unit test coverage. From what I see mantisBidAdapter has 54% coverage. Please increase unit test coverage above the 80% threshold. To generate and view unit test coverage:
```
gulp test-coverage
gulp view-coverage
```

modules/mantisBidAdapter.js
parisholley: correct, we don't have plans to track user data in the EU (most of our business focuses on the US), so just prefer to act as if we never have consent (which we do today on any inbound EU requests outside prebid).

parisholley:
@aleksatr how do you recommend we test non-prebid related code? unlike most of the adapters, we have complex viewability checks built into the bidder, which would require the tests to actually execute a stubbed prebid request, write a creative into an iframe, etc... i don't see any examples of how to do this.. there isn't much value in unit testing this code as it is very browser dependent and to truly test, would require cross-browser execution vs only running in headless chrome

parisholley:
@aleksatr done :)

aleksatr:
LGTM

Issue #5424, opened 2020-06-26
Bugs in sendBidsControl.dealPrioritization


jbartek25:
## Type of issue
Bug

## Description
`sendBidsControl.dealPrioritization` doesn't work.

## Steps to reproduce
Enable `sendBidsControl.dealPrioritization` and run a HB auction as usual. This will lead to a syntax error:

![Screenshot 2020-06-24 at 12 29 06](https://user-images.githubusercontent.com/31618107/85846287-e05e8880-b7a5-11ea-8986-eb82b8eef94e.png)
This can be easily fixed by `bucketBids.sort(sortByDealAndPriceBucketOrCpm(true))` in 
https://github.com/prebid/Prebid.js/blob/master/src/targeting.js#L47. 

However, after applying the fix another error appears:

![Screenshot 2020-06-24 at 12 33 14](https://user-images.githubusercontent.com/31618107/85846760-9d50e500-b7a6-11ea-87b2-78004c104741.png)

The problem is that method [sortByDealAndPriceBucketOrCpm](https://github.com/prebid/Prebid.js/blob/master/src/targeting.js#L77) looks for `hb_deal` inside `adUnitTargeting` object which only exists when `sortByDealAndPriceBucketOrCpm` gets called from within `filterTargetingKeys`. However, when `sortByDealAndPriceBucketOrCpm` is called in `getHighestCpmBidsFromBidPool`, `hb_deal` is in `adserverTargeting` object and `adUnitTargeting` is undefined.

jbartek25:
Test page: https://test-hb.improvedigital.com/tmp/prebid-deal-issue/issue.html

ChrisHuie:
closing out with merge of #5504

PR #5504, opened 2020-07-16
fixed bug with sortByDealAndPriceBucketOrCpm

r-schweitzer:
## Type of change

- [x] Bugfix

## Description of change
Fixes a bug related to sortByDealAndPriceBucketOrCpm when sendBidsControl.dealPrioritization = true
 

## Other information
https://github.com/prebid/Prebid.js/issues/5424

### DIFF
--- a/src/targeting.js
+++ b/src/targeting.js
@@ -44,7 +44,7 @@ export function getHighestCpmBidsFromBidPool(bidsReceived, highestCpmCallback, a
     Object.keys(bidsByBidder).forEach(key => bucketBids.push(bidsByBidder[key].reduce(highestCpmCallback)));
     // if adUnitBidLimit is set, pass top N number bids
     if (adUnitBidLimit > 0) {
-      bucketBids = dealPrioritization ? bucketBids(sortByDealAndPriceBucketOrCpm(true)) : bucketBids.sort((a, b) => b.cpm - a.cpm);
+      bucketBids = dealPrioritization ? bucketBids.sort(sortByDealAndPriceBucketOrCpm(true)) : bucketBids.sort((a, b) => b.cpm - a.cpm);
       bids.push(...bucketBids.slice(0, adUnitBidLimit));
     } else {
       bids.push(...bucketBids);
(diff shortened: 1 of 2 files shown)


src/targeting.js
```diff
@@ -44,7 +44,7 @@ export function getHighestCpmBidsFromBidPool(bidsReceived, highestCpmCallback, a
     Object.keys(bidsByBidder).forEach(key => bucketBids.push(bidsByBidder[key].reduce(highestCpmCallback)));
     // if adUnitBidLimit is set, pass top N number bids
     if (adUnitBidLimit > 0) {
-      bucketBids = dealPrioritization ? bucketBids(sortByDealAndPriceBucketOrCpm(true)) : bucketBids.sort((a, b) => b.cpm - a.cpm);
+      bucketBids = dealPrioritization ? bucketBids.sort(sortByDealAndPriceBucketOrCpm(true)) : bucketBids.sort((a, b) => b.cpm - a.cpm);
```
FilipStamenkovic, line 47: Please add test case where it would fail on master branch and where it will pass on this branch.   

The fix you did is not covered with tests:

image

FilipStamenkovic:
I still don't see that fix is covered by test case, can you add test case that will cover it?  
Also, since this is change in prebid core, you will need 2nd reviewer to approve MR.

src/targeting.js
```diff
@@ -240,13 +240,13 @@ export function newTargeting(auctionManager) {
     let targetingMap = Object.keys(targetingCopy).map(adUnitCode => {
       return {
         adUnitCode,
-        adUnitTargeting: targetingCopy[adUnitCode]
+        adserverTargeting: targetingCopy[adUnitCode]
```
mkendall07, line 243: curious why this name was changed?

r-schweitzer: The sortByDealAndPriceBucketOrCpm is now also used in getHighestCpmBidsFromBidPool and the bids received object has the original name of adserverTargeting. If we want to leave the adUnitTargeting name for that cloned targeting map I could change it to work a different way. Just felt it made sense to go with the same name and went with the original adserverTargeting name.

PR #5506, opened 2020-07-16
[synacormedia] Update adapter to support Consent Management Module

coreykress:
## Type of change

- [x] Feature

## Description of change
Modified Synacormedia bid adapter to use the Consent Management Module.


coreykress:
It looks like the failing build is due to a different adapter's (YieldOne) unit test.

modules/synacormediaBidAdapter.js
```diff
@@ -85,6 +85,17 @@ export const spec = {
...
 
+    // CCPA
+    if (bidderRequest && bidderRequest.uspConsent) {
+      if (!openRtbBidRequest.regs) {
+        openRtbBidRequest.regs = {};
+      }
+      if (!openRtbBidRequest.regs.ext) {
+        openRtbBidRequest.regs.ext = {};
+      }
+      openRtbBidRequest.regs.ext.us_privacy = bidderRequest.uspConsent;
```
smenzer, line 96: I would suggest using `utils.deepSetValue()` method instead of all the if statements here (you'll need to import `utils` as well)

```suggestion
      utils.deepSetValue(openRtbBidRequest, 'regs.ext.us_privacy', bidderRequest.uspConsent);
```

modules/synacormediaBidAdapter.js
```diff
@@ -85,6 +85,17 @@ export const spec = {
...
     });
 
+    // CCPA
+    if (bidderRequest && bidderRequest.uspConsent) {
+      if (!openRtbBidRequest.regs) {
+        openRtbBidRequest.regs = {};
+      }
+      if (!openRtbBidRequest.regs.ext) {
+        openRtbBidRequest.regs.ext = {};
+      }
```
smenzer, line 95: if you use the `deepSetValue` function, you don't need these checks

PR #5508, opened 2020-07-17
Add apstream adapter

frstua:
## Type of change

- [x] New bidder adapter   

## Description of change
### Inherit from prebid.js
```
    var adUnits = [
      {
        code: '/19968336/header-bid-tag-1',
        mediaTypes: { // mandatory and should be only one
            banner: {
                sizes: [[920,180], [920, 130]]
            }
        },
        bids: [{
            bidder: 'apstream',
            params: {
                publisherId: STREAM_PIBLISHER_ID // mandatory
            }
        }]
      }
    ];
```

### Explicit ad-unit code
```
    var website = null;
    switch (location.hostname) { 
      case "site1.com": 
        website = "S1"; 
        break;
      case "site2.com":
        website = "S2";
        break;
    }
    
    var adUnits = [
      {
        code: '/19968336/header-bid-tag-1',
        mediaTypes: { // mandatory and should be only one
            banner: {
                sizes: [[920,180], [920, 130]]
            }
        },
        bids: [{
            bidder: 'apstream',
            params: {
                publisherId: STREAM_PIBLISHER_ID, // mandatory
                code: website + '_Leaderboard'
            }
        }]
      }
    ];
```

### Explicit ad-unit ID
```
    var adUnits = [
      {
        code: '/19968336/header-bid-tag-1',
        mediaTypes: { // mandatory and should be only one
            banner: {
                sizes: [[920,180], [920, 130]]
            }
        },
        bids: [{
            bidder: 'apstream',
            params: {
                publisherId: STREAM_PIBLISHER_ID, // mandatory
                adunitId: 1234
            }
        }]
      }
    ];
```

tech@audienceproject.com

### DIFF
--- a/modules/apstreamBidAdapter.js
+++ b/modules/apstreamBidAdapter.js
@@ -10,6 +10,265 @@ const CONSTANTS = {
 };
 const storage = getStorageManager(CONSTANTS.GVLID, CONSTANTS.BIDDER_CODE);
 
+var dsuModule = (function() {
+  'use strict';
+
+  var DSU_KEY = 'apr_dsu';
+  var DSU_VERSION_NUMBER = '1';
+  var SIGNATURE_SALT = 'YicAu6ZpNG';
+  var DSU_CREATOR = {'USERREPORT': '1'};
+
+  function stringToU8(str) {
+    if (typeof TextEncoder === 'function') {
+      return new TextEncoder().encode(str);
+    }
+    str = unescape(encodeURIComponent(str));
+    var bytes = new Uint8Array(str.length);
+    for (var i = 0, j = str.length; i < j; i++) {
+      bytes[i] = str.charCodeAt(i);
+    }
+    return bytes;
+  }
+
+  function _add(a, b) {
+    var rl = a.l + b.l;
+    var a2 = {
+      h: a.h + b.h + (rl / 2 >>> 31) >>> 0,
+      l: rl >>> 0
+    };
+    a.h = a2.h;
+    a.l = a2.l;
+  }
+  function _xor(a, b) {
+    a.h ^= b.h;
+    a.h >>>= 0;
+    a.l ^= b.l;
+    a.l >>>= 0;
+  }
+  function _rotl(a, n) {
+    var a2 = {
+      h: a.h << n | a.l >>> (32 - n),
+      l: a.l << n | a.h >>> (32 - n)
+    };
+    a.h = a2.h;
+    a.l = a2.l;
+  }
+  function _rotl32(a) {
+    var al = a.l;
+    a.l = a.h;
+    a.h = al;
+  }
+
+  function _compress(v0, v1, v2, v3) {
+    _add(v0, v1);
+    _add(v2, v3);
+    _rotl(v1, 13);
+    _rotl(v3, 16);
+    _xor(v1, v0);
+    _xor(v3, v2);
+    _rotl32(v0);
+    _add(v2, v1);
+    _add(v0, v3);
+    _rotl(v1, 17);
+    _rotl(v3, 21);
+    _xor(v1, v2);
+    _xor(v3, v0);
+    _rotl32(v2);
+  }
+  function _getInt(a, offset) {
+    return a[offset + 3] << 24 |
+          a[offset + 2] << 16 |
+          a[offset + 1] << 8 |
+          a[offset];
+  }
+
+  function hash(key, m) {
+    if (typeof m === 'string') {
+      m = stringToU8(m);
+    }
+    var k0 = {
+      h: key[1] >>> 0,
+      l: key[0] >>> 0
+    };
+    var k1 = {
+      h: key[3] >>> 0,
+      l: key[2] >>> 0
+    };
+    var v0 = {
+      h: k0.h,
+      l: k0.l
+    };
+    var v2 = k0;
+    var v1 = {
+      h: k1.h,
+      l: k1.l
+    };
+    var v3 = k1;
+    var ml = m.length;
+    var ml7 = ml - 7;
+    var buf = new Uint8Array(new ArrayBuffer(8));
+    _xor(v0, {
+      h: 0x736f6d65,
+      l: 0x70736575
+    });
+    _xor(v1, {
+      h: 0x646f7261,
+      l: 0x6e646f6d
+    });
+    _xor(v2, {
+      h: 0x6c796765,
+      l: 0x6e657261
+    });
+    _xor(v3, {
+      h: 0x74656462,
+      l: 0x79746573
+    });
+    var mp = 0;
+    while (mp < ml7) {
+      var mi = {
+        h: _getInt(m, mp + 4),
+        l: _getInt(m, mp)
+      };
+      _xor(v3, mi);
+      _compress(v0, v1, v2, v3);
+      _compress(v0, v1, v2, v3);
+      _xor(v0, mi);
+      mp += 8;
+    }
+    buf[7] = ml;
+    var ic = 0;
+    while (mp < ml) {
+      buf[ic++] = m[mp++];
+    }
+    while (ic < 7) {
+      buf[ic++] = 0;
+    }
+    var mil = {
+      h: buf[7] << 24 | buf[6] << 16 | buf[5] << 8 | buf[4],
+      l: buf[3] << 24 | buf[2] << 16 | buf[1] << 8 | buf[0]
+    };
+    _xor(v3, mil);
+    _compress(v0, v1, v2, v3);
+    _compress(v0, v1, v2, v3);
+    _xor(v0, mil);
+    _xor(v2, {
+      h: 0,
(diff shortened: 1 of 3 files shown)


patmmccann:
your adapter submissions shouldnt be adding 100k lines to prebid.js or changing index.html

frstua:
Tests failed by reason not related to my adapter. Do I have possibility to rerun tests manually?

aleksatr:
hey @frstua , I don't see the link to the docs PR: https://github.com/prebid/prebid.github.io/

modules/apstreamBidAdapter.js
```diff
@@ -0,0 +1,225 @@
...
...[0]:"",pathname:e[5]||"",search:e[7]||"",hash:e[9]||"",toString:function(){return n}};return t.origin=t.scheme+"://"+t.host,t}(n);return d.url=n,d.parsed=r,r}function p(){var n=function(){if("function"==typeof Uint32Array&&"undefined"!=typeof crypto&&void 0!==crypto.getRandomValues){var n=new Uint32Array(4);crypto.getRandomValues(n);var r=-1;return"xxxxxxxx-xxxx-4xxx-yxxx-xxxxxxxxxxxx".replace(/[xy]/g,function(e){var t=n[++r>>3]>>r%8*4&15;return("x"===e?t:3&t|8).toString(16)})}var e=(new Date).getTime();return"undefined"!=typeof performance&&"function"==typeof performance.now&&(e+=performance.now()),"xxxxxxxx-xxxx-4xxx-yxxx-xxxxxxxxxxxx".replace(/[xy]/g,function(n){var r=(e+16*Math.random())%16|0;return e=Math.floor(e/16),("x"===n?r:3&r|8).toString(16)})}(),o=v(),x=f(n+o.toString()),a=x.substr(0,4),u=x.substr(4);n=n.substr(0,19)+a+"-"+u;var i,l=(i=(new Date).getTime()-new Date(2019,0,1).getTime(),Math.floor(i/864e5)),c=f(o.origin),h=[t.USERREPORT,l,c].join("."),s=f(n+h+e);return[r,s,n,h].join(".")}return{readOrCreateDsu:function(){var r;try{r=storage.getDataFromLocalStorage(n)}catch(n){return null}r||(r=p());try{storage.setDataInLocalStorage(n,r)}catch(n){return null}return r}}}();
```
aleksatr, line 173: What is this? This doesn't look transparent, I can't know what side effects this code has, to the page or to the pb.js auction.

frstua:
Here is the link to the docs: https://github.com/prebid/prebid.github.io/pull/2141

modules/apstreamBidAdapter.js
frstua: This is our internal module to generate unique token with some metadata and store into local storage. It does not interact with anything except local storage.

aleksatr:
@frstua I see in docs file (.md) that you state that us_privacy is supported, but I don't see that your adapter is reading uspConsent string?

modules/apstreamBidAdapter.js
aleksatr: You can use some of the built-in functions for generating unique identifiers (like generateUUID). Also, you should be using Prebid methods for working with local storage (look for storageManager), this way you get GDPR enforcement rules out of the box.

frstua:
Sorry, can't find where us_privacy was mentioned.

aleksatr:
> Sorry, can't find where us_privacy was mentioned.

Sorry, my bad, I must have seen `gdpr_supported` and thought that I saw `us_privacy`, removing the comment.

modules/apstreamBidAdapter.js
frstua: I'v added one more commit where our module is re-wrote in the clear way. Also I realized what we are using standard UUID generation so it has been replaced with Prebid's generateUUID implementation. And for working with the local storage I used Prebid's methods.

frstua:
Hi @aleksatr. All corrections have been made.

modules/apstreamBidAdapter.js
```diff
@@ -0,0 +1,481 @@
...
+    (gdprConsentConfig.vendorData.vendor || {}).consents ||
+    {}
+  );
+  let isConsentGiven = !!vendorConsents[CONSTANTS.GVLID.toString(10)];
+
+  return isIab && isConsentGiven ? consentString : null;
+}
+
+function getIabConsentString(bidderRequest) {
+  if (utils.deepAccess(bidderRequest, 'gdprConsent')) {
```
aleksatr, line 306: hey @frstua , in case someone did not include `consentManagement` module in their PBJS build or they simply didn't enable it (for example, they know they're not in GDPR zone so no need to enable it), `gdprConsent` will be undefined. This function will then return `null` and you will skip setting the `dsu`. Is this intentional for some reason?

frstua: Hi @aleksatr. I changed logic which works with dsu and consent. But now the tests are failed by reason not related to my adapter.

frstua: Hi @aleksatr. Good catch, I changed logic which sets dsu.

modules/apstreamBidAdapter.js
```diff
@@ -0,0 +1,484 @@
...
+
+  const consentData = getRawConsentString(bidderRequest.gdprConsent);
+  data.iab_consent = consentData;
+
+  let options = {
+    withCredentials: true
+  };
+
+  const isConsent = getIabConsentString(bidderRequest);
+  if (isConsent) {
```
patmmccann, line 430: can you add a bidder config option here and document it so that pubs who aren't comfortable with this functionality can disable the dsu part of your adapter? I think it is okay to default to on but should be on the bidder params section of the documentation at http://prebid.org/dev-docs/bidders.html

patmmccann: noting from the module rules:

"Bidder modules must not(*) make requests to endpoints for functionality other than auctions without:
Disclosure (See the disclosure section below.)
Ability for the publisher to control the additional functionality."

patmmccann: https://docs.google.com/document/d/1kK0ifXV-vzyV7VchPKoyGuz375H7fzbTJFt_oWeCiRo/

modules/apstreamBidAdapter.js
smenzer: this looks much better, thank you

modules/apstreamBidAdapter.js
frstua: Hi @patmmccann. We are not making any requests to endpoints related to DSU. As I can see in the docs, this rule is applied only for requests to endpoints for functionality other than auctions. In our case DSU just identify user and is generated on client side, this is a part of our workflow. Also it uses local storage only if consent is given.

patmmccann: I think publishers being able to control this additional functionality (generate and pass dsu) aligns quite well with module rules. The idea is some publishers are subject to  regulatory scrutiny or have stronger internal standards and may wish to control this functionality at a more detailed level than (isconsent) and may wish to check for consent for various tcf2 purposes.

frstua: Hi @patmmccann. Where is better place to set this parameter? I can add this option to params section to every bid, but does it makes sense as far this DSU flag is browser specific but not bid specific? Does it means what I should take into account parameter only from first bit or what do you could suggest?

patmmccann: I imagine a publisher would do something like: `var adUnits = [{
      code: 'test-div',
      sizes: [[300, 600]],
      bids: [{
          bidder: 'apstream',
          params: { 
              dsu: 'off' //defaults to on
          }
      }]
    }];` when they want to override it

bretg: There's precedence for bidder-specific global params. e.g.
```
pbjs.setConfig({
   apstream: {dsu: false}
});
```
Your code can do a `getConfig` to look for this optional param.

frstua:
Hi @patmmccann @aleksatr.

I added possibility to disable DSU via config section. Also docs have been updated.

Is it possible to merge it today as we have publisher and he is waiting for adapter.

aleksatr:
LGTM

romanych:
Hi @aleksatr. I see that 4.2.0 is released and the adapter in question is merged. Dev-docs PR https://github.com/prebid/prebid.github.io/pull/2141 was merged as well. However, I can not find it on the [Download page](http://prebid.org/download.html). I was unable to find the information regarding how this list is updated

aleksatr:
hey @romanych , your docs PR is not properly formatted, that's why your adapter was not picked up by the renderer (jerkyll). You're missing "front matter" - https://jekyllrb.com/docs/front-matter/. Check how `.md` files of other adapters are formatted and submit another docs PR to fix it.

You can open a discussion on the original docs PR as well, to confirm this.
@MartianTribe https://github.com/prebid/prebid.github.io/pull/2141

PR #5509, opened 2020-07-17
Replace third party deep-equal library with native implementation

Fawke:
## Type of change
- [X] Utils Functions

## Description of change
[deep-equal](https://www.npmjs.com/package/deep-equal) module is replaced by a native implementation for deeply checking object equality. 

This function is mainly used by the `sizeMappingV2.js` module, which is tested and working fine after the replacement. This implementation is simple, but should mostly work for our use cases. 

Also, `prebid- core.js` is now almost `40kb` lighter after removal of this module.


Fawke:
@snapwich @idettman Thanks for the review. 

@snapwich, Good catch, I forgot to replace the reference from `allowedModules.js`. One place, where I can not remove it from is [fake-responder.js](https://github.com/prebid/Prebid.js/blob/c933c4f755ec40dcc4ee945a24e344e76e84fd07/test/fake-server/fake-responder.js#L2) file.

It runs in node environment and throws an error if I try to require the utils.js file because the utils file uses import statements which are not recognized by node. I tried few solutions, like:
- using the `mjs` extension on `fake-responder.js` file
- trying to include node environment as target in `babelrc.js` file
- adding `type: "module"` key in parent package.json file

None of these approaches work. So, in the end, I had to add back the `deep-equal` library, but this time as a `devDependency`. We still achieve reduction in overall size of `prebid-core` webpack chunk.

If you have any better solution to how I can import utils in a file running in node env, let me know.

idettman:
> It runs in node environment and throws an error if I try to require the utils.js file because the utils file uses import statements which are not recognized by node. I tried few solutions, like:
> 
> * using the `mjs` extension on `fake-responder.js` file
> * trying to include node environment as target in `babelrc.js` file
> * adding `type: "module"` key in parent package.json file

Imports and exports support In **Node.js 12** requires both:
1. Passing the --experimental-modules flag when running Node.js
2. using `.mjs` extension or set `"type": "module"` in your `package.json`.
I'm guessing the problem is the missing --experimental-modules flag

Fawke:
@idettman I tried the approach that you suggested, using the flag `--experimental-modules`

1. setting extension to `.mjs` doesn't work because then, we would need to change the extension of all the files which are es modules to `.mjs`, which would require us to change extension of `utils.js` file and all the files imported by it.

2. setting type, in package.json to module also creates some problem. In some places in the code, we are still using `require` instead of import. This will require us to remove all references to `require` which will be an elaborate change, deserving its own PR.

We can copy the implementation of `deepEqual` from utils to a separate function in `fake-response.js` file, or use the third party `deep-equal` library but only as a devDependency. Let me know your thoughts.

Issue #4858, opened 2020-02-13
[Feature] Consent Module re-read consent


ReinoutStevens:
## Feature Request
The consent module stores the consent string and reuses it across requests, even when the user changes their consent at a later point.

## Description
The consent module reads the consent string once, and keeps using it for the rest of the session. As a result, prebid does not notice if users later change their consent in the same session.

For new users it makes sense to fire off the first auction without consent ( and without tracking ), as the user won't accept/change his settings in time for the first auction. By the time ads are refreshed the user has given his consent, and so later auctions/refreshes should use this new consent string. This is especially needed for SPA's, where no actual page changes occur and prebid remains loaded.

Either always fetching the updated consent string would be an option, or exposing a method that reset/re-reads the consent data would be handy.

bretg:
@jaiminpanchal27 - please cover this scenario in #4801?

timhanlon:
I'm building a Chrome extension to help people verify their CMP implementations are functional, and just uncovered this behaviour.

Looks like it'd be a trivial fix: https://github.com/prebid/Prebid.js/blob/master/modules/consentManagementUsp.js#L161

If people are relying on this behaviour, perhaps it could go behind a `consentManagement` param...something like `checkConsentBeforeSubsequentAuctions` or `alwaysCheckConsent`?

Would you accept a PR?

Issue #4577, opened 2019-12-11
BUG - consentData is not updating if changed by user.


orzilca:
## Type of issue
BUG

## Description
When a user changes an existing consent settings, it does not reflect on new bid requests and the data is stored locally once.

## Steps to reproduce
1. enable consentModule
2. consent to certain purpose
3. see consent string sent in bid requests
4. change consent purposes
5. consent string in bid requests does not change (however in __cmp("getVendorConsents") its does).

### Expected results
Consent data should update.
Or at least some functionality to ask for data reset should be exposed (resetConsentData already exists but is not exposed to global pbjs object)

### Actual results
Consent string never updates unless page is refreshed.

## Platform details
prebid 1.17, prebid 1.19

PR #5513, opened 2020-07-18
Allow developers to disable caching of TCF consentData

xavierleune:
## Type of change
- [x] Feature

## Description of change
This pull requests allows developers to disable the caching of consentdata. The behavior of consentManagement actually tries to fetch consents only once and keep the retrieved consents (or non retrieved consents) cached for ever, which is fine for most people.

However there are 2 valid use cases when you might want to disable that:

* If you run an auction immediately after the user entered the site, before you got any consent (it's fine) but want the next refresh to use consents you got in the meantime,
* If you run an auction after you got the consent, but then the user removes it, you don't want to allow 3rd parties to use personnal data at the next refresh.

PR for docs repo: prebid/prebid.github.io#2143

## Other information
This has been spotted in a few issues (#4577, #4858 and perhaps others too).

Thanks for your help,

### DIFF
--- a/modules/consentManagement.js
+++ b/modules/consentManagement.js
@@ -13,12 +13,14 @@ import strIncludes from 'core-js-pure/features/string/includes.js';
 const DEFAULT_CMP = 'iab';
 const DEFAULT_CONSENT_TIMEOUT = 10000;
 const DEFAULT_ALLOW_AUCTION_WO_CONSENT = true;
+const DEFAULT_CACHE_CONSENT_DATA = true;
 
 export let userCMP;
 export let consentTimeout;
 export let allowAuction;
 export let gdprScope;
 export let staticConsentData;
+export let cacheConsentData;
 
 let cmpVersion = 0;
 let consentData;
@@ -257,7 +259,7 @@ export function requestBidsHook(fn, reqBidsConfigObj) {
   };
 
   // in case we already have consent (eg during bid refresh)
-  if (consentData) {
+  if (consentData && cacheConsentData) {
     utils.logInfo('User consent information already known.  Pulling internally stored information...');
     return exitModule(null, hookConfig);
   }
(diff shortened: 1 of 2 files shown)


xavierleune:
Actually I think I faced another issue here. I double check the implementation of tcfv2 in prebid and my issue is really this one:

* I run auctions without consent when the user has not given it
* The user gives the consent & a refresh is made
* Prebid use previously stored consentdata
* The listener in prebid is called by the cmp, to store the new consent data, too late for the previous auction

Of course I can put a timeout before making new auctions, to ensure that prebid got called. But I think that allowing to reset the consentdata could be a better option after all. WDYT ?

jsnellbaker:
@xavierleune To clarify something with the behavior you saw, were you using a TCF2 CMP or a v1 CMP?  

In the case of v1, the stored value of consent could cause an issue within "the same page session" if the user decided to update their consent data after the page fully loaded.  While the consent data stored by Prebid would clear itself on the next full page load, if you were refreshing the bids - the previous consent data would continue to be used during that interval which is not completely ideal.  

Part of the reason for this behavior is the type of calls that were available with the v1 spec.  They were direct requests to the CMP API and data was provided with a matching callback function.  As it was expensive for the auction to wait for this asynchronous back/forth, we thought it would be alright to leave it as is.

In the case of v2, this behavior shouldn't be an issue.  The v2 spec offers the event-listener route as a means to provide updated information on the consent while the page is still active.  So even though we're getting the initial data, if the user updates their choices - the CMP should load a follow-up event that Prebid would still be listening to.  When that second event comes to Prebid, we'd update our internal copy of the consent info and use that for any subsequent auctions (in the same page session).

patmmccann:
docs pr says default is false; however default is true.

bretg:
@Fawke - didn't we address something like this in another PR?

Fawke (commented):
@xavierleune,

In TCF2 cmp, the auction will not use the previously stored consent data, whenever the user updates their consent information and clicks accept, an event `useractioncomplete` is emitted by the cmp for which there is a listener in consentManagement module which updates the consentData before the auction takes place.

bretg:
What's the status here? @Fawke and @jsnellbaker seem to think this feature isn't needed. @xavierleune - do you agree?

bretg:
Closing this PR as not needed -- if there is in fact a need for it, please explain and we'll re-open

PR #5521, opened 2020-07-21
Add Supply Supply Chain Object Module support for Yieldlab Adapter

mirkorean:
## Type of change

- [x] Feature

## Description of change

Support for the Supply Chain Object Module (http://prebid.org/dev-docs/modules/schain.html). We want to serialize the bid.schain object according to: https://github.com/InteractiveAdvertisingBureau/openrtb/blob/master/supplychainobject.md#serialization-of-an-openrtb-supplychain-object-into-a-url-parameter

solutions@yieldlab.de


modules/yieldlabBidAdapter.js
```diff
@@ -198,7 +201,13 @@ function createQueryString (obj) {
   let str = []
   for (var p in obj) {
     if (obj.hasOwnProperty(p)) {
-      str.push(encodeURIComponent(p) + '=' + encodeURIComponent(obj[p]))
+      let param = p
```
msm0504, line 204: Why is variable `param` needed? Can `p` be used in those places, too?

modules/yieldlabBidAdapter.js
```diff
@@ -221,6 +230,35 @@ function createTargetingString (obj) {
...
 }
 
+/**
+ * Creates a string out of a schain object
+ * @param {Object} schain
+ * @returns {String}
+ */
+function createSchainString (schain) {
+  let ver = schain.ver ? schain.ver : ''
+  let complete = schain.complete ? schain.complete : ''
```
msm0504, line 240: These lines can be the following. Ternary isn't needed.
```
let ver = schain.ver || ''
let complete = schain.complete || ''
```

modules/yieldlabBidAdapter.js
```diff
@@ -221,6 +230,35 @@ function createTargetingString (obj) {
...
+  let nodes = []
+  for (let i = 0; i < schain.nodes.length; i++) {
+    let node = []
+    node.push(schain.nodes[i].asi, schain.nodes[i].sid, schain.nodes[i].hp, schain.nodes[i].rid, schain.nodes[i].name, schain.nodes[i].domain, schain.nodes[i].ext)
+    let filteredAndEncodedNode = node.map(x => x === undefined ? '' : encodeURIComponentWithBangIncluded(x))
```
msm0504, line 245: Run this function on each item when it is pushed into the array. No reason to push the items and then immediately loop over them.

mirkorean:
I'll integrate your request changes soon, busy days with TCF2. Removing the stale flag with this comment hopefully?

modules/yieldlabBidAdapter.js
mirkorean: Thanks for the input. Fine with me!

modules/yieldlabBidAdapter.js
mirkorean: Can be used, yes. Changed.

modules/yieldlabBidAdapter.js
mirkorean: I changed this altogether to not use arrays anymore.

mirkorean:
@msm0504 Thanks you for your review. I'm sorry for the delay here, but i finally came around to look into your and integrate your comments. Hope this lgty now :)

modules/yieldlabBidAdapter.js
msm0504: Using arrays is fine. I only meant that it could be simplified. What you switched to works, but since the code now repeats the same line for each key, you could define an array of the key strings and map over that for each schain node.

```
const keys = ['asi', 'sid', 'hp', 'rid', 'name', 'domain', 'ext'];
const nodesString = schain.nodes.reduce((acc, node) => {
    return acc += `!${keys.map(key => node[key] ? encodeURIComponentWithBangIncluded(node[key]) : '').join()}`
}, '');
```

mirkorean: Thanks for the input and help here @msm0504!  I did not know about the reduce method till now. I just used your code (only had to include the join separator `,`).

msm0504 (approved):

PR #5522, opened 2020-07-21
Fix double encoded targeting string for Yieldlab adapter

mirkorean:
## Type of change

- [x] Bugfix

## Description of change

Values in the targeting string can include comma (',') and we do not want to double encode those as we will encode again when we build the querystring with createQueryString().

solutions@yieldlab.de


smenzer:
can you please add a unit test to cover this change?

mirkorean:
@smenzer thank you for the hint, i added a check to the targeting test.

aleksatr:
LGTM

Issue #5511, opened 2020-07-17
Floors Module: Analytics interface changes


bszekely1:
**Description of issue**
The goal of this ticket is to address a misrepresentation of an analytics condition within the Analytics Interface of the Floors Module as well as provide better logging for analytics adapters. 

When the Floors Module uses the Fetch method to retrieve floors and a timeout occurs (when the auction delay time is realized before fetched floors is returned), today the Floors Module sets the skipRate to 0 if skipRate is not set in setConfig because there is no skip data to utilize. Setting the skipRate to 0 is an inaccurate representation because skipRate is not actually 0, it's really undefined.

The are two additional enhancements to logging for analytics adapters. The three changes for this ticket are the following:
* When the Floors Module code is invoked and there is no floor data available, the Floors Module should record a new attribute value of "noData" in the bidRequest[ ].floorData.location field

The below feature will be added into Prebid 4.x to prevent any loss of data for analytics adapters wishing to use this field right away.
* The introduction of a new field called "floorProvider" in the top level of the floor object or within floors.data, whose value should be a string
    * If floorProvider is supplied in both the top level of floors and within the floors.data object, the floors.data object setting shall prevail
    * floorProvider should function with fetch (dynamic) mode of floors, with the floorProvider field supplied at the root of the fetch object

diDNA-matt:
In regards to: 
> The introduction of a new field called "floorProvider" in the top level of the floor object, whose value should be a string

Is there a desired default value for when none is set, or is an empty string fine? Also what would be an example of a value passed to it be?

bszekely1:
> In regards to:
> 
> > The introduction of a new field called "floorProvider" in the top level of the floor object, whose value should be a string
> 
> Is there a desired default value for when none is set, or is an empty string fine? Also what would be an example of a value passed to it be?

This should an empty string when none provided. It would be unfair to favor a particular vendor. The purpose of the string is to prevent an analytics adapter from reading floors from the Floors Module when that analytics adapter is not being used as a provider.

GLStephen:
@bszekely1 is the desire to only block the floors INPUT from other providers, but not the outflow into the analytics data on the bidRequest and bidResponse objects?

bszekely1:
> @bszekely1 is the desire to only block the floors INPUT from other providers, but not the outflow into the analytics data on the bidRequest and bidResponse objects?

[...]

PR #5538, opened 2020-07-24
Update floors module for #5511

diDNA-matt:
## Type of change

- [x] Other

## Description
- When no floor data is available after the floors module is invoked, records 'noData' in the bidRequest[ ].floorData.location field
- Adds the field floorProvider, to the top level of the floor object. The value is a string and defaults to an empty string if none is provided.

Updates made for https://github.com/prebid/Prebid.js/issues/5511

### DIFF
--- a/modules/priceFloors.js
+++ b/modules/priceFloors.js
@@ -287,9 +287,11 @@ export function updateAdUnitsForAuction(adUnits, floorData, auctionId) {
       bid.floorData = {
         skipped: floorData.skipped,
         modelVersion: utils.deepAccess(floorData, 'data.modelVersion'),
-        location: utils.deepAccess(floorData, 'data.location'),
-        skipRate: floorData.skipRate,
+        location: utils.deepAccess(floorData, 'data.location', 'noData'),
         fetchStatus: _floorsConfig.fetchStatus
+      }
+      if (floorData.skipRate) {
+        bid.floorData.skipRate = floorData.skipRate;
       }
     });
   });
@@ -569,7 +571,7 @@ export function handleSetFloorsConfig(config) {
     'enabled', enabled => enabled !== false, // defaults to true
     'auctionDelay', auctionDelay => auctionDelay || 0,
     'endpoint', endpoint => endpoint || {},
-    'skipRate', () => !isNaN(utils.deepAccess(config, 'data.skipRate')) ? config.data.skipRate : config.skipRate || 0,
+    'skipRate', () => !isNaN(utils.deepAccess(config, 'data.skipRate')) ? config.data.skipRate : config.skipRate || undefined,
     'enforcement', enforcement => utils.pick(enforcement || {}, [
       'enforceJS', enforceJS => enforceJS !== false, // defaults to true
       'enforcePBS', enforcePBS => enforcePBS === true, // defaults to false
--- a/modules/priceFloors.js
+++ b/modules/priceFloors.js
@@ -286,12 +286,10 @@ export function updateAdUnitsForAuction(adUnits, floorData, auctionId) {
       bid.auctionId = auctionId;
       bid.floorData = {
         skipped: floorData.skipped,
+        skipRate: floorData.skipRate,
         modelVersion: utils.deepAccess(floorData, 'data.modelVersion'),
         location: utils.deepAccess(floorData, 'data.location', 'noData'),
         fetchStatus: _floorsConfig.fetchStatus
-      }
-      if (floorData.skipRate) {
-        bid.floorData.skipRate = floorData.skipRate;
       }
     });
   });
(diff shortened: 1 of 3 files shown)


modules/priceFloors.js
```diff
@@ -568,8 +568,9 @@ export function handleSetFloorsConfig(config) {
   _floorsConfig = utils.pick(config, [
     'enabled', enabled => enabled !== false, // defaults to true
     'auctionDelay', auctionDelay => auctionDelay || 0,
+    'floorProvider', floorProvider => String(floorProvider) || '',
     'endpoint', endpoint => endpoint || {},
-    'skipRate', () => !isNaN(utils.deepAccess(config, 'data.skipRate')) ? config.data.skipRate : config.skipRate || 0,
+    'skipRate', () => !isNaN(utils.deepAccess(config, 'data.skipRate')) ? config.data.skipRate : config.skipRate || undefined,
```
robertrmartinez, line 573: We want to actually leave in the default of `0`. The goal with the Issue is to give analytics adapters the ability to infer what is happening. The adding of `noData` to the `location` enum succeeds in doing this.

We do not want `skipRate` to be a required field, so let's leave it at zero.

modules/priceFloors.js
```diff
@@ -568,8 +568,9 @@ export function handleSetFloorsConfig(config) {
   _floorsConfig = utils.pick(config, [
     'enabled', enabled => enabled !== false, // defaults to true
     'auctionDelay', auctionDelay => auctionDelay || 0,
+    'floorProvider', floorProvider => String(floorProvider) || '',
```
robertrmartinez, line 571: I think just leaving this as `floorProvider` is better. If it is not set it will be set as `undefined` which is basically the same thing as `empty string`. Little less code.

modules/priceFloors.js
```diff
@@ -286,9 +286,9 @@ export function updateAdUnitsForAuction(adUnits, floorData, auctionId) {
       bid.auctionId = auctionId;
       bid.floorData = {
         skipped: floorData.skipped,
-        modelVersion: utils.deepAccess(floorData, 'data.modelVersion'),
-        location: utils.deepAccess(floorData, 'data.location'),
         skipRate: floorData.skipRate,
+        modelVersion: utils.deepAccess(floorData, 'data.modelVersion'),
+        location: utils.deepAccess(floorData, 'data.location', 'noData'),
```
robertrmartinez, line 291: Nice use of the default here 👍

modules/priceFloors.js
robertrmartinez:
```
    'auctionDelay', auctionDelay => auctionDelay || 0,
    'floorProvider',
```

modules/priceFloors.js
```diff
@@ -286,9 +286,9 @@ export function updateAdUnitsForAuction(adUnits, floorData, auctionId) {
...
         fetchStatus: _floorsConfig.fetchStatus
```
robertrmartinez, line 292: We should pass along `floorProvider` here.

The point of `floorProvider` is so analytics adapters can pick and choose which data they send to their endpoints.

modules/priceFloors.js
diDNA-matt: That makes sense, I wasn't sure how strict to enforce typing for this.

modules/priceFloors.js
diDNA-matt: Thank you 👍 looking forward to working on this module more.

bszekely1:
> Hey @diDNA-matt Thanks a bunch for getting to this! Nice to have another developer working on the module.
> 
> Couple things that may have been a misunderstanding.
> 
> I'll let @bszekely1 confirm since he wrote the issue.

I addressed the point in the main issue ticket, but to briefly reiterate that the Floors Module will always default to 0 when not supplied by the publisher or a floors provider. This makes it impossible for the `skipRate` to be undefined. The Floors Module will define when available.

modules/priceFloors.md
```diff
@@ -11,6 +11,7 @@ pbjs.setConfig({
       enforceJS: true //defaults to true
     },
     auctionDelay: 150, // in milliseconds defaults to 0
+    floorProvider: '', //defaults to en empty string
```
robertrmartinez, line 14: Technically no default now, just put in some random example please!

`awesomeFloorProviderName` maybe 😉

robertrmartinez:
@Fawke 

I have approved this pending the `.md` update.

Please give it a look when you can!

modules/priceFloors.md
diDNA-matt: I knew I forgot something. would you like me to note that it's an optional field?

robertrmartinez:
@bszekely1 

There should be some docs updated regarding floorProvider and how it is meant to be used.

Just a pass through from the floors module to the analytics adapters so they can filter out floors which do not come from their own data (wether that be in the analytics module itself, or simply passing along the provider name and their server side pipeline can do the filtering)

bszekely1:
PR for documentation: https://github.com/prebid/prebid.github.io/pull/2169

Brief note, I added additional functionality and flexibility of the floorProvider parameter to be at the top level of the floors object and within the data object to allow floor provider to self-declare.

Issue #5216, opened 2020-05-12
Proposal - GPT Pre-Auction module


caitlinwright123:
This is a followup to the pbAdSlot feature covered in https://github.com/prebid/Prebid.js/issues/4149

### Background
PbAdSlot’s current solution solved the issue of publishers having no consistent inventory identifier passed to each bidder that can be used for reporting and deal targeting, but if there is a predefined value of 'pbadslot', the GAM ad unit is not also passed.

### Problems to Solve
This update to PbAdSlot  is attempting to solve the following:
Publishers passing PbAdSlot are not necessarily passing GAM ad unit

- The current solution allows publishers to define the field, and in that scenario, GAM ad unit is not passed. This is needed for publishers to compare inventory across other potential integration types.
- Low adoption rates

### Proposal
We propose a new 'GPT Pre-Auction" module:

1. accept optional initialization parameters:
-ability to turn off module (on by default)
-customGptSlotMatching function
-customPbAdSlot function
2. if the module is on, use the BEFORE_REQUEST_BIDS hook to define a function that:
3. loops through each adunit in the auction
-maps the adunit to the GPT slot using the same algorithm as setTargetingForGPTAsync including customGptSlotMatching
-if GPT slot matching succeeded:
--set fpd.context.adserver.name to 'gam'
--copy the resulting GPT slot name to fpd.context.adserver.adSlot
-if customPbAdSlot function was specified, invoke it, writing the results to fpd.context.pbAdSlot.
-Else use a default algorithm to determine fpd.context.pbAdSlot:
--First use the AdUnit's context.pbAdSlot if defined
--Else, see if the AdUnit.code corresponds to a div and if so, try to retrieve a data element from the div called data-adslotid.
--Else if the GPT slot matching succeeded, use the GPT slot name
--Else, just use the AdUnit.code, assuming that that's the ad unit slot

We also need to update the prebidServerBidAdapter to:
- pass fpd.context.adserver.name in the OpenRTB as imp[].ext.context.data.adserver.name
- pass fpd.context.adserver.adSlot in the OpenRTB as imp[].ext.context.data.adserver.adslot
- at some point we may change to pass fpd.context.pbAdSlot in the OpenRTB as imp[].ext.context.data.pbadslot. But this would be a breaking change and PBS will need time to get ready for it.

patmmccann:
is this related to line 39 here?

https://github.com/prebid/Prebid.js/blob/master/modules/33acrossBidAdapter.js

[...]

PR #5539, opened 2020-07-24
Pb ad slot module

msm0504:
## Type of change

- [X] Feature

## Description of change

New module that, when enabled, will add PB Ad Slot and matching GAM ad unit name to each ad unit's first-party data before bid requests are sent to the adapters.

Implements Proposal #5216

Docs: https://github.com/prebid/prebid.github.io/pull/2168


idettman:
Docs PR https://github.com/prebid/prebid.github.io/pull/2168

PR #5543, opened 2020-07-27
AdButler Bid Adapter: Add ability to include extra query params

dharton:
## Type of change
- [x] Feature

## Description of change
This change will allow users of the AdButler Bid Adapter to pass other potentially necessary parameters in their ad requests.

### DIFF
--- a/modules/adbutlerBidAdapter.js
+++ b/modules/adbutlerBidAdapter.js
@@ -25,6 +25,7 @@ export const spec = {
     let requestURI;
     let serverRequests = [];
     let zoneCounters = {};
+    let extraParams = {};
 
     for (i = 0; i < validBidRequests.length; i++) {
       bidRequest = validBidRequests[i];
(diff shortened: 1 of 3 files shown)


aleksatr:
hey @dharton , I don't see that you attached the PR for https://github.com/prebid/prebid.github.io/, where you would describe this additional parameter. Also, can you include it in the example in this repo: `adbutlerBidAdapter.md`?

dharton:
Hi @aleksatr 

Thank you for your feedback! I've updated the .md file in this repo and please see the link below for the corresponding documentation pull request.

https://github.com/prebid/prebid.github.io/pull/2172

Please let me know if there's anything else you see!

modules/adbutlerBidAdapter.js
```diff
@@ -25,13 +25,15 @@ export const spec = {
     let requestURI;
     let serverRequests = [];
     let zoneCounters = {};
+    let extraParams = {};
 
     for (i = 0; i < validBidRequests.length; i++) {
       bidRequest = validBidRequests[i];
       zoneID = utils.getBidIdParameter('zoneID', bidRequest.params);
       accountID = utils.getBidIdParameter('accountID', bidRequest.params);
       keyword = utils.getBidIdParameter('keyword', bidRequest.params);
```
patmmccann, line 34: You should be reading kw from fpd.context when available.

dharton: Hi @patmmccann 

Thanks for the advice, that seems really handy! 

Perhaps that should be a separate commit & pull request since it's unrelated to this change? I'd be dropping that in as a replacement to the 'keyword' bid param on line 34 there.

patmmccann: makes sense ty! I brought it up bc kw was an example of your extra param.

aleksatr:
LGTM

PR #5559, opened 2020-07-30
floorProvider can come from data obj now

robertrmartinez:
Adding to #5538 
Type of change
- [X] Feature

## Description of change
About 1 hour after merging #5538 @bszekely1 realized we also want the `floorProvider` to be able to be set at the `data` level of floor configs.

This way, a fetch can indicate who the `floorProvider` is as well as just the top level.

It works similarly as `skipRate`, being the data level field takes precedence over the top level `setConfig` field.

### DIFF
--- a/modules/priceFloors.js
+++ b/modules/priceFloors.js
@@ -514,6 +514,7 @@ export function handleFetchResponse(fetchResponse) {
     _floorsConfig.data = fetchData;
     // set skipRate override if necessary
     _floorsConfig.skipRate = utils.isNumber(fetchData.skipRate) ? fetchData.skipRate : _floorsConfig.skipRate;
+    _floorsConfig.floorProvider = fetchData.floorProvider || _floorsConfig.floorProvider;
   }
 
   // if any auctions are waiting for fetch to finish, we need to continue them!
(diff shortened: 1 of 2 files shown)


robertrmartinez:
@diDNA-matt Small addition to your code update. Take a look if you can please!

diDNA-matt:
> @diDNA-matt Small addition to your code update. Take a look if you can please!

That makes a lot of sense, and looks good. Are there any issues you can think of where an analytics adapter may be expecting one vs the other? I guess that would be on the analytics adapter to handle.

robertrmartinez:
@diDNA-matt 

So the main reason we wanted to have the floorProvider is so Analytics adapters can make a decision on what they want to do when they do not know who is making the floors.

for instance, in the Rubicon analytics adapter, our floor provider team only wants to learn off of their own floor data. So we will be filtering out any floor data that comes in and is not from rubicon.

Other analytics adapters may find it beneficial to just pass in floor data regardless, and just pass through the provider name to their data processing.

So you are correct, it is dependent on what the analytics adapter's use cases are for reading floor information.

bszekely1:
Adding PR for docs: https://github.com/prebid/prebid.github.io/pull/2169

diDNA-matt:
@robertrmartinez

Thanks for the example and explanation 👍

robertrmartinez:
Hi @Fawke need a second review for a small addition to the Price Floors Module if you have a quick minute!

Thanks!

Issue #5546, opened 2020-07-28
TCF v2 logic question


jdelhommeau:
## Type of issue
question

## Description
in consentManagement.js file, the below logic is used to check if TCF v2 consent string is available:

` else if (tcfData.eventStatus === 'cmpuishown' && tcfData.tcString && tcfData.purposeOneTreatment === true) {
        cmpSuccess(tcfData, hookConfig);
      }`

So it seems like, in case a TCString is already available, but user is currently editing his choices (eventStatus == 'cmpuishown'), the CMP will only be initiated in case existing tcfData.purposeOneTreatment is true. 
I am not sure to undertand the logic behind this.

If a tcString is already available, the CMP should be initiated, despite purposeOneTreatment value I think.

Can you clarify the logic behind this?

patmmccann:
A lot of the discussion on this logic is on #5564 comments

bretg:
Are we good here @jdelhommeau ?

jdelhommeau:
Yes @bretg . Sorry about delay, the comments in #5564 fully answer my question. Thank you

jdelhommeau:
see #5564 for details

Issue #5529, opened 2020-07-22
New IAB TCFv2.0 LiveRamp CMP live - revenue and CPM drops


MattCharge:
Hi,

I'm new to github and have run out of ideas on why I've seen a significant drop in prebid revenue and cpm since we released LiveRamp's CMP on our site earlier this week. 

We're running Prebid version 3.18.0
We released the Liveramp CMP across our site (cv-library.co.uk) at the beginning of the week. Since then I have seen Prebid revenue drop off a cliff and both Google AdX and Adsense CPM's have dropped by half. 

When I debug Prebid on the site (e.g. https://www.cv-library.co.uk/accountancy-jobs?pbjs_debug=true) I can see the CMP load, when I consent I then see the consent string present. 
I can also see that on the surface it appears we are providing the consent string in our Bid requests to partners (see attached sshot). 
There are other prebid calls on the page (different versions) that relate to other partners loaded in our ad server which I'm not too concerned about for the time being. 

It looks like both the GDPR Consent Module and GDPR Enforcement Module are working as expected. 

What i can't explain is why our Prebid partners have stopped bidding or are bidding at a much lower rate. 

If anyone can provide any suggestions I'm all ears. 

First thoughts are:
Some sort of Timeout issue
CMP is interfering with the bid requests in some way...
There is an issue with the Prebid version we are using
Under Vendor data the 'cmpVersion' is set to '1' - should this not be '2'
<img width="1429" alt="Screenshot 2020-07-22 at 10 48 00" src="https://user-images.githubusercontent.com/68640428/88162436-4275a680-cc09-11ea-9742-9bec7d4911e0.png">

I can see another thread related to Quantcast here that sounds similar to the issue we are having - https://github.com/prebid/Prebid.js/issues/5462

I've contacted Prebid partners and am waiting to hear back as well as discussions with LiveRamp but nothing as yet has presented itself as the potential issue.

If anyone can take a look and advise at all I'd be extremely grateful!!

Many thanks

Matt

## Test page
It's on the live site but ads and Prebid calls can be seen on all Job posting pages - e.g. https://www.cv-library.co.uk/accountancy-jobs

MattCharge:
Ok I may have found a fix related to the purposeOneTreatment:
 
Prebid receives a response from the CMP that has the following value:
addtlConsent: "1~"
cmpId: 3
cmpStatus: "loading"
cmpVersion: 1
eventStatus: "tcloaded"
gdprApplies: true
[...]

Issue #5462, opened 2020-07-07
Prebid not waiting for TCF2.0 consent string from Quantcast CMP


chrissjhitc:
## Type of issue
Question

## Description

We're trying to set up TCF2.0 with the Quantcast CMP but have ran into an issue with Prebid not waiting for consent and requesting the first set of bids with an undefined consent string. It is our understanding that Prebid should be waiting for consent from the CMP before proceeding with bids, so we're not sure if this is a bug with Prebid, Quantcast or a problem with our implementation.

It appears that Prebid is deciding GDPR does not apply due to gdprApplies=false in the initial 'tcloaded' event data and is proceeding with the auction. We are then receiving another 'useractioncomplete' event from the Quantcast CMP shortly afterwards that contains gdprApplies: true and the consent string, but by that time Prebid has already requested the first set of bids.

Is the above the expected behaviour on Prebid's end, with the events from the Quantcast CMP being issue?

## Test page
https://www.hitc.com/prebid-tcf2-cmp-test.php?pbjs_debug=true

### Expected results
Requested bids to contain consent string from Quantcast CMP

### Actual results
Bids are requested without consent string from Quantcast CMP

## Platform details

Prebid 3.24.0

Josh-G:
We have experienced similar issues in their v1.1 implementation where certain values would not be initially available. We assumed this was a race condition and it was reported to them. They never resolved it and we implemented a workaround to delay calling their CMP until it started returning sane values.

It seems to me that this is a problem with Quantcast again, based on [this snippet from the CMP API document](https://github.com/InteractiveAdvertisingBureau/GDPR-Transparency-and-Consent-Framework/blob/master/TCFv2/IAB%20Tech%20Lab%20-%20CMP%20API%20v2.md#what-does-the-gdprapplies-value-mean):
> However, gdprApplies may not be available until the CMP is finished loading and the value will, therefore, be undefined

chrissjhitc:
We do think this is an issue with the Quantcast CMP. According to our reading of the IAB specs the CMP should not be triggering a tcloaded event until the consent string is available, and gdprApplies should be undefined instead of false while it is pending determination.

We have tried a similar workaround of using a __tcfapi listener to wait for the consent string to be supplied before making the call to requestBids. This does seem to work and get the consents added to all of the bid requests.

We've started a support request with Quantcast as it looks like there's either a bug or a problem with their implementation.

PrjAdv:
[...]

PR #5564, opened 2020-08-03
Fix v2CmpResponseCallback handle

akiselicki-liveramp:
## Type of change

- [x] Bugfix

## Description of change

With current implementation we are getting timeout for all cmp calls that doesn't satisfy defined conditions 
even if the response is valid.

[tcfData.eventStatus](https://github.com/InteractiveAdvertisingBureau/GDPR-Transparency-and-Consent-Framework/blob/master/TCFv2/IAB%20Tech%20Lab%20-%20CMP%20API%20v2.md#addeventlistener) will always have one of those three values, purposeOneTreatment won't always be true and if gdpr doesn't apply [TCData](https://github.com/InteractiveAdvertisingBureau/GDPR-Transparency-and-Consent-Framework/blob/master/TCFv2/IAB%20Tech%20Lab%20-%20CMP%20API%20v2.md#tcdata) object will be without tcString and purposeOneTreatment properties.
In this case there is no need for those checks, only value that matters for cmp response to be valid is success.

Related issues: #5529, #5546

### DIFF
--- a/modules/consentManagement.js
+++ b/modules/consentManagement.js
@@ -100,11 +100,7 @@ function lookupIabConsent(cmpSuccess, cmpError, hookConfig) {
   function v2CmpResponseCallback(tcfData, success) {
     utils.logInfo('Received a response from CMP', tcfData);
     if (success) {
-      if (tcfData.eventStatus === 'tcloaded' || tcfData.eventStatus === 'useractioncomplete') {
-        cmpSuccess(tcfData, hookConfig);
-      } else if (tcfData.eventStatus === 'cmpuishown' && tcfData.tcString && tcfData.purposeOneTreatment === true) {
-        cmpSuccess(tcfData, hookConfig);
-      }
+      cmpSuccess(tcfData, hookConfig);
     } else {
       cmpError('CMP unable to register callback function.  Please check CMP setup.', hookConfig);
     }
--- a/modules/consentManagement.js
+++ b/modules/consentManagement.js
@@ -100,7 +100,13 @@ function lookupIabConsent(cmpSuccess, cmpError, hookConfig) {
   function v2CmpResponseCallback(tcfData, success) {
     utils.logInfo('Received a response from CMP', tcfData);
     if (success) {
-      cmpSuccess(tcfData, hookConfig);
+      if (tcfData.gdprApplies === false) {
+        cmpSuccess(tcfData, hookConfig);
+      } else if (tcfData.eventStatus === 'tcloaded' || tcfData.eventStatus === 'useractioncomplete') {
+        cmpSuccess(tcfData, hookConfig);
+      } else if (tcfData.eventStatus === 'cmpuishown' && tcfData.tcString && tcfData.purposeOneTreatment === true) {
+        cmpSuccess(tcfData, hookConfig);
+      }
     } else {
       cmpError('CMP unable to register callback function.  Please check CMP setup.', hookConfig);
     }


andyblackwell:
addEventListener does trigger in the wild without a tcfData.eventStatus (the liveramp cmp for non-EU traffic)

This is how I dealt with the gdprApplies=false case, to not get rid of those other necessary checks:
```
  function v2CmpResponseCallback(tcfData, success) {
    utils.logInfo('Received a response from CMP', tcfData);
    if (success) {
      if (tcfData.gdprApplies === false) {
      } else if (tcfData.eventStatus === 'tcloaded' || tcfData.eventStatus === 'useractioncomplete') {
      } else if (tcfData.eventStatus === 'cmpuishown' && tcfData.tcString && tcfData.purposeOneTreatment === true) {
      }
    } else {
      cmpError('CMP unable to register callback function.  Please check CMP setup.', hookConfig);
    }
  }
```

nielsbaarsma:
Hey @andyblackwell , Can you elaborate why these additional checks are necessary? We're not seeing it.

patmmccann:
@nielsbaarsma  just as some background from the blame

https://github.com/prebid/Prebid.js/pull/5050

https://github.com/prebid/Prebid.js/pull/4911#pullrequestreview-373287875

fyi @harpere @jsnellbaker

andyblackwell:
thanks @patmmccann for the references! Yeah, there are reasons for those checks, I've seen @chrispaterson (IAB core api dev) also mentioning their importance before, like here: https://github.com/prebid/Prebid.js/issues/5462#issuecomment-656222312

akiselicki-liveramp:
Hi @andyblackwell I still don't see why those checks are necessary.
If I'm looking at the right place, in processCmpData() cmp data will be checked
```
  function checkV2Data() {
    // if CMP does not respond with a gdprApplies boolean, use defaultGdprScope (gdprScope)
    let gdprApplies = consentObject && typeof consentObject.gdprApplies === 'boolean' ? consentObject.gdprApplies : gdprScope;
    let tcString = consentObject && consentObject.tcString;
    return !!(
      (typeof gdprApplies !== 'boolean') ||
      (gdprApplies === true && !utils.isStr(tcString))
    );
  }
```
so if checkV2Data() is checking for gdprApplies and tcString there's no need for any check in v2CmpResponseCallback() or I'm missing something?

chrispaterson:
I think @akiselicki-liveramp is right, before [all statuses](https://github.com/InteractiveAdvertisingBureau/GDPR-Transparency-and-Consent-Framework/blob/master/TCFv2/IAB%20Tech%20Lab%20-%20CMP%20API%20v2.md#addeventlistener) were accounted for and responded to in the same way.  I'm not too familiar with what is going on here, but it seems like this is only a proxy method and that whatever passes in the `cmpSuccess` method should be handling the different `addEventListener` statuses and not this proxy method.

One thing that has be concerned is looking at `purposeOneTreatment`.  [As was said to Vizzini by Indigo Montoya](https://youtu.be/dTRKCXC0JFg?t=6), I don't think that means what you think it means. It should not be used for PreBid, it's for determining different handling of display of [purpose one in different jurisdictions](https://github.com/InteractiveAdvertisingBureau/GDPR-Transparency-and-Consent-Framework/blob/master/TCFv2/IAB%20Tech%20Lab%20-%20Consent%20string%20and%20vendor%20list%20formats%20v2.md#what-if-consent-is-governed-differently-in-a-country) because that pertains to the eprivacy directive and different EU countries have different legal requirements.  If you're looking for consent for purpose one, you should be checking `tcfData.purpose.consent['1']` is `true`.

andyblackwell:
I was going to mention that the LiveRamp cmp started off with a default tcString and data before any user response, which might have caused an issue with this change, but that looks to have been fixed with the latest release yesterday. So your change may be fine? But I'll leave that to people who know the rules more clearly.

thanks for chiming in @chrispaterson! Is the IAB worried at all about things still being kinda up-in-the air like this 10 days out from the deadline?

modules/consentManagement.js
```diff
@@ -100,11 +100,7 @@ function lookupIabConsent(cmpSuccess, cmpError, hookConfig) {
   function v2CmpResponseCallback(tcfData, success) {
     utils.logInfo('Received a response from CMP', tcfData);
     if (success) {
-      if (tcfData.eventStatus === 'tcloaded' || tcfData.eventStatus === 'useractioncomplete') {
-        cmpSuccess(tcfData, hookConfig);
-      } else if (tcfData.eventStatus === 'cmpuishown' && tcfData.tcString && tcfData.purposeOneTreatment === true) {
-        cmpSuccess(tcfData, hookConfig);
-      }
+      cmpSuccess(tcfData, hookConfig);
```
harpere, line 103: when the `eventStatus` === "cmpuishown" we may not have a valid `tcString` yet, like we do with "tcloaded" and "useractioncomplete".  I don't recall or understand why `tcfData.purposeOneTreatment` is significant here, but I don't think we want to call `cmpSuccess()` if we're still waiting for a valid `tcString`.

harpere:
I think the point was to wait for a valid `tcString` before calling `cmpSuccess()`.  eventStatus === "cmpuishown" means we're still waiting for the user to respond, which is why we don't call `cmpSuccess()` for that event.  We'd wait for the "useractioncomplete" eventStatus, or timeout.

modules/consentManagement.js
chrispaterson: The previous implementation calls `cmpSuccess(tcfData, hookConfig)` in all cases.



> I don't think we want to call `cmpSuccess()` if we're still waiting for a valid `tcString`.

A CMP returns a "Valid" TC string in all `eventStatus`es.  Once `"cmpuishown"` is triggered legitimate interest has been established but consent hasn't been gathered.  Some "Vendors" only rely on that legal basis.  Vendors should be able to determine which state they need the TCString to be in.

chrispaterson:
@andyblackwell CMPs will no longer have access to the v1 global vendor list after the 15th, however v1 strings are "valid" to be processed until September 30th.  Ostensibly, that means that no new v1 consent will be gathered by CMPs while existing v1 consent strings will still be able to be surfaced to vendors.

It's obviously a monumental task to move the entire industry along, but based on rapidly evolving privacy laws and case rulings it has to be done to preserve lawfulness in ad tech.

modules/consentManagement.js
harpere: I incorrectly said the `tcString` may not be "valid".  It's valid, but user consent may not have been gathered yet.   I believe `cmpSuccess()` is intended to be called when the user's consent preferences have been obtained.  Most bidders will want the user's consent, if possible, before sending a bid request.  Perhaps a change could be made to give the publisher the option to proceed with the auction without giving the user a chance to respond.  Maybe setting the timeout to zero already accomplishes this?

harpere:
I think adding this makes sense though
```
if (tcfData.gdprApplies === false) {
}

File: modules/consentManagement.js (hunk 1 above)
Comment (chrispaterson) [reply modules/consentManagement.js:103]: There is more nuance to TCFv2 in that a legitimate interest basis for processing personal data as defined by [GDPR Article 6(1)(f)](https://gdpr-info.eu/art-6-gdpr/) is now represented in the TC String.  There are Vendors who operate solely on Legitimate Interest and that's why these events were created.  It may be true that most bidders will want the user's consent but there are some who were advocating for this feature through the design of TCFv2.

**Comment by chrispaterson** [comment]
> I think adding this makes sense though
> 
> ```
> if (tcfData.gdprApplies === false) {
>         cmpSuccess(tcfData, hookConfig);
> }
> ```

I don't see why, because all possible use cases are handled in this pull request.  If `gdprApplies` is `true` or `false` `cmpSuccess` is called.  (Am I missing something?)

File: modules/consentManagement.js (hunk 1 above)
Comment (harpere) [reply modules/consentManagement.js:103]: Prebid doesn't work that way.  There's no mechanism for some bidders to wait for additional consent from the user (or wait for anything else for that matter) while others do not.  All bidders are called at the same time and have the same amount of time to respond before being timed out.  If a publisher doesn't want bidders to wait for the user to respond, they could set the `cmpTimeout` to 0 (or another small number).  Though I guess in that case we wouldn't be using the default `tcData` that was returned (Though we may want to update the code to save it as a backup if we don't get a response from the other events before the timeout).  There may also be some GDPR Enforcement repercussions to be considered (when that module is included), if we only get a default tcData object.

**Comment by harpere** [comment]
I don't think we're settled on calling `cmpSuccess()` on `cmpuishown`, at least by default, because then bidders would never have the opportunity to get the user's preferences before the auction begins.

**Comment by chrispaterson** [comment]
Here's how I think about it:
```javascript
function v2CmpResponseCallback(tcData, success) {

  utils.logInfo('Received a response from CMP', tcData);

  if (success) {

    /**
     * First, did the CMP self-report success.  It's not well defined what
     * success=false actually means and when a CMP is to respond that way other
     * than it's a way to enable a CMP to respond without hanging calling
     * scripts if it's unable to perform the action for any reason.
     */

    if (tcData.gdprApplies === false) {

      // GDPR doesn't apply. Success! Nothing more to do.

      cmpSuccess(tcData, hookConfig);

    } else if (tcData.gdprApplies === true) {

      // GDPR does apply.  More to do...

      switch (tcData.eventStatus) {

        case 'tcloaded':

          /**
           * The CMP does not need to ask the user, because the user already
           * has valid consent/ transparency.  This is hopefully the majority
           * case of responses because the CMP does not ask for new consent every
           * page.  Worst case (safari or other cookie blocking browsers) the
           * CMP would ask every session.
           */
          cmpSuccess(tcData, hookConfig);
          break;
        case 'cmpuishown':

          /**
           * In this case the TC string from the CMP is either an all "no" TC
           * string (which may be unuseful) or a string that the CMP has
           * determined is invalid.  I would say invoking cmpSuccess with this
           * TC Data isn't of much use to Vendors in their bidding because at
           * this point they know that gdprApplies=true.  You know that at
           * some point the 'useractioncomplete' may fire, but prebid probably
           * doesn't want to wait for that to happen. It seems to me that you
           * would treat this eventStatus as gdprApplies but no consent.
           */

          // ?????? Up to the Prebid masters to determine

          break;
        case 'useractioncomplete':

          /**
           * You know the TC Data is valid and fresh here but this happens in
           * human-scale time because the user has to click accept or adjust
           * their preferences.  As an aside and one that you may already
           * know, something like 98%+ of users will click an "accept all".
           */
          cmpSuccess(tcData, hookConfig);
          break;
        default:

          /**
           * unspecified weirdness: handled.
           */
          cmpError('CMP unable to register callback function.  Please check CMP setup.', hookConfig);

      }

    } else {

      /**
       * GDPR applies is not defined for some reason... This should not happen,
       * but in a universe composed of infinite probability and unknown CMP
       * implementers it could.
       */

      cmpError('CMP unable to register callback function.  Please check CMP setup.', hookConfig);

    }

  } else {

    /**
     * No soup for you...
     */

    cmpError('CMP unable to register callback function.  Please check CMP setup.', hookConfig);

  }

}
```

akiselicki-liveramp:
Ok, so for now it makes sense to go with suggestion from @andyblackwell to only add new case if GDPR doesn't apply and we'll open a new issue to address the purposeOneTreatment.

jsnellbaker:
To help comment, there was some internal discussion with our legal contact on this issue and what Prebid.js should be listening to event-wise.  The consensus was that Prebid.js probably doesn't need to listen to the `cmpuishown` event at all, but it should listen to the `tcloaded` and `useractioncomplete` events.

Prebid.js cares about collecting 'consent' information about the user and providing that to all of the bidders in a timely and correct manner.  To copy a snippet from the conversation that explains this best:
> “cmpuishown” is designed to indicate the purgatory between a user being shown purposes (establishing legitimate interest) but before the user has made any consent elections. This state was surfaced for vendors relying solely on legitimate interest interacting with the CMP JS API. Prebid itself is not such a vendor, as it will always wait for the user’s consent choices to be known, by either ‘tcloaded’ or ‘useractioncomplete’.

If everyone else agrees, I think we can drop that particular if condition entirely and keep the rest of the checks that were implemented here (the other 2 statuses and gdprApplies check).  

@akiselicki-liveramp @chrispaterson @harpere please let me know your thoughts on this suggestion.

chrispaterson:
@jsnellbaker that makes the most sense to me, especially since purpose 1 (device storage and access) requires consent in almost all jurisdictions.

PR #5602, opened 2020-08-13
Adhese: set gvlid && filter out empty params

mefjush:
## Type of change

- [ ] Bugfix
- [x] Feature
- [ ] New bidder adapter   
- [ ] Build related changes
- [ ] CI related changes
- [ ] Other

## Description of change
Configure the gvlid for Adhese adapter
Reduce the outgoing request size in case the user provided empty parameters

### DIFF
--- a/modules/adheseBidAdapter.js
+++ b/modules/adheseBidAdapter.js
@@ -4,10 +4,12 @@ import { registerBidder } from '../src/adapters/bidderFactory.js';
 import { BANNER, VIDEO } from '../src/mediaTypes.js';
 
 const BIDDER_CODE = 'adhese';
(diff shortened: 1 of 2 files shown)


wqi1972:
Question: @mefjush, what suppose to happen if the parameter is a string with only space ' '? Current logic won't filter it out, not sure if that is expected.
If that is not a problem, then the code is good to merge.

mefjush:
That's certainly a valid question as indeed just a space is not a valid input.

After a consideration we'd prefer to keep the "filtering" as it is. We'd rather catch all less common incorrect values in our backend so we are aware of them and can act if necessary.

PR #5606, opened 2020-08-14
Fix for userId_example.html

shikhar-dev-proj:
## Type of change

- [x] Bugfix
- [ ] Feature
- [ ] New bidder adapter   
- [ ] Build related changes
- [ ] CI related changes
- [ ] Other

## Description of change
On running `gulp serve` and navigating to **integrationExamples/gpt/userId_example.html?pbjs_debug=true**, one can see prebid initialised but the bids are not generated.

Issue arises due to:-
1. no `mediaTypes` configuration in adUnit (test-div).
2. `usersync` instead of `userSync` present in setConfig method.
Fixes of these both are available in this pr and results in correct behaviour in terms of userIds in `pbjs.getUserIds()` and bids in `pbjs.adUnits`.

Find the attached console output which shows the error
Screenshot 2020-08-14 at 2 54 17 PM
.

### DIFF
--- a/integrationExamples/gpt/userId_example.html
+++ b/integrationExamples/gpt/userId_example.html
@@ -84,6 +84,9 @@
             {
                 code: 'test-div',
                 sizes: [[300,250],[300,600],[728,90]],
+                mediaTypes: {
+                    banner: {}
+                },
                 bids: [
                     {
                         bidder: 'rubicon',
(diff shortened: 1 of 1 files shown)


shikhar-dev-proj:
Hi @jlukas79, can we merge this pr? 
Please let me know if anything required at my end.

smenzer (approved):
LGTM

smenzer:
this example page is now broken because it relies on tcf 1 which is no longer supported by the IAB, so the vendorlist.json is failing to load. this is a problem with the other examples, too, though, so that should be corrected separately. i'll go ahead and merge this for now.

PR #5608, opened 2020-08-14
Update dfpAdServerVideo.js to allow for vast4

patmmccann:
This had a bug in which all requests were hardcoded to vast 3, although pubs may have selected vast 3 or 4. This allows DFP to respond with the DFP set default Vast version. 

References:
https://support.google.com/admanager/answer/7358411?hl=en&ref_topic=7544527
https://support.google.com/admanager/answer/4517679?hl=en&ref_topic=7544527
## Type of change

- [X] Bugfix
- [ ] Feature
- [ ] New bidder adapter   
- [ ] Build related changes
- [ ] CI related changes
- [ ] Other

### DIFF
--- a/modules/dfpAdServerVideo.js
+++ b/modules/dfpAdServerVideo.js
@@ -42,7 +42,7 @@ import { auctionManager } from '../src/auctionManager.js';
 const defaultParamConstants = {
   env: 'vp',
   gdfp_req: 1,
-  output: 'xml_vast3',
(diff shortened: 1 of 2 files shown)


patmmccann:
All the doc I can find had examples overriding this bad default, so I don't think a doc update is needed.

Fawke (approved):
Good catch!

PR #5609, opened 2020-08-14
Price Floors: Add bid object into `cpmAdjustment` function

robertrmartinez:
## Type of change
- [X] Bugfix

## Description of change
The priceFloors module has to run the `bidCpmAdjustment` function if a bidder has it.

Turns out, that not only is the cpm value needing to be passed into the adjustment function, but some publishers use the bidResponse object to grab the CPM (even though they are the same)

It is possible publishers to more intricate logic with this, so we will pass along the bidResponse object so it works in all scenarios!

### DIFF
--- a/modules/priceFloors.js
+++ b/modules/priceFloors.js
@@ -138,10 +138,10 @@ function generatePossibleEnumerations(arrayOfFields, delimiter) {
 /**
  * @summary If a the input bidder has a registered cpmadjustment it returns the input CPM after being adjusted
  */
-export function getBiddersCpmAdjustment(bidderName, inputCpm) {
+export function getBiddersCpmAdjustment(bidderName, inputCpm, bid = {}) {
   const adjustmentFunction = utils.deepAccess(getGlobal(), `bidderSettings.${bidderName}.bidCpmAdjustment`);
   if (adjustmentFunction) {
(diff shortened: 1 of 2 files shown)


modules/priceFloors.js
```diff
@@ -138,10 +138,10 @@ function generatePossibleEnumerations(arrayOfFields, delimiter) {
 /**
  * @summary If a the input bidder has a registered cpmadjustment it returns the input CPM after being adjusted
  */
-export function getBiddersCpmAdjustment(bidderName, inputCpm) {
+export function getBiddersCpmAdjustment(bidderName, inputCpm, bid = {}) {
   const adjustmentFunction = utils.deepAccess(getGlobal(), `bidderSettings.${bidderName}.bidCpmAdjustment`);
   if (adjustmentFunction) {
-    return parseFloat(adjustmentFunction(inputCpm));
+    return parseFloat(adjustmentFunction(inputCpm, {...bid, cpm: inputCpm}));
```
robertrmartinez, line 144: If this is called at bidRequest time, we do not really have a bidResponse object, so we''ll make a pseduo one with the cpm. Where cpm is actually the resulting `floor` from the adapters `getFloor` call.

Fawke (approved):

PR #5612, opened 2020-08-15
URL encode the IX r JSON string parameter.

mgriego:
cc @ix-prebid-support 

## Type of change
- [x] Bugfix

## Description of change
IX bid adapter does not URL encode the stringified JSON payload in the `r` parameter.  This leads to malformed query strings when the payload contains unsafe characters.  For instance, if the page URL contains ampersands, these are not encoded in their resulting JSON properties, breaking the JSON payload string.

### DIFF
--- a/modules/ixBidAdapter.js
+++ b/modules/ixBidAdapter.js
@@ -363,7 +363,7 @@ function buildRequest(validBidRequests, bidderRequest, impressions, version) {
   // Use the siteId in the first bid request as the main siteId.
   payload.s = validBidRequests[0].params.siteId;
   payload.v = version;
-  payload.r = JSON.stringify(r);
+  payload.r = encodeURIComponent(JSON.stringify(r));
   payload.ac = 'j';
   payload.sd = 1;
   if (version === VIDEO_ENDPOINT_VERSION) {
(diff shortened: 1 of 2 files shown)


kizzard:
Hey @mgriego , Prebid handles the encoding of the GET request payload data automatically in [bidderFactor.js](https://github.com/prebid/Prebid.js/blob/master/src/adapters/bidderFactory.js#L233), by use of [parseQueryStringParameters()](https://github.com/prebid/Prebid.js/blob/db02d8a252b928a052748a2e307876c46708d560/src/utils.js#L116) , so this should already be taken care of. Are you seeing different behavior in Prebid?

mgriego:
@kizzard I see what's happening.  I didn't realize that the `parseQueryStringParameters` function had gone through some churn starting at version 3.16.  That update that went out in 3.16 broke QS encoding.  It appears that was fixed in 4.3, and I didn't realize that there was a fix there.  We have slated to update to 4.x this week.  But, I will say that it feels like there needs to be a *BIG* notice somewhere that on versions 3.16-4.2, there could be problems stemming from the change to using `internal.formatQS`, which does *NOT* encode params properly.

kizzard:
@mgriego yeah - it was broken on 3.16 but [fixed in 3.20](https://github.com/prebid/Prebid.js/pull/5230). Agreed that the problem was not well communicated, although it only affected a couple of adapters. Sorry for the inconvenience. I'll close this out now.

kizzard:
Looks like I don't have power to close it out. @robertrmartinez would you assist?

mgriego:
Thanks, I'll close it out.

PR #5616, opened 2020-08-17
Check localstorage availability before accessing it

Swiiip:
## Type of change

- [x] Bugfix

## Description of change
Chrome now throws an error when trying to access window.localstorage in a third-party context when third-party cookies are disabled (Private mode for instance). This PR checks whether the localStorage is available each time we try to access it.

### DIFF
--- a/test/spec/unit/core/storageManager_spec.js
+++ b/test/spec/unit/core/storageManager_spec.js
@@ -46,18 +46,16 @@ describe('storage manager', function() {
 
   describe('localstorage forbidden access in 3rd-party context', function() {
     let errorLogSpy;
+    const originalLocalStorage = { get: () => window.localStorage };
+    const localStorageMock = { get: () => { throw Error } };
 
     beforeEach(function() {
-      var mock = {
-        get: function() {
-          throw Error;
-        }
-      };
-      Object.defineProperty(window, 'localStorage', mock);
+      Object.defineProperty(window, 'localStorage', localStorageMock);
       errorLogSpy = sinon.spy(utils, 'logError');
     });
 
     afterEach(function() {
+      Object.defineProperty(window, 'localStorage', originalLocalStorage);
       errorLogSpy.restore();
     })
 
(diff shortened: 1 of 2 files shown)


dmdabbs:
Is local storage access gated by the consent module, where applicable? If you can't read a cookie then No local storage soup for you either.

test/spec/unit/core/storageManager_spec.js
```diff
@@ -42,5 +42,22 @@ describe('storage manager', function() {
...
+
+    coreStorage.setDataInLocalStorage('key', 'value');
+    const val = coreStorage.getDataFromLocalStorage('key');
+    coreStorage.removeDataFromLocalStorage('key');
+
+    expect(val).to.be.null;
+    sinon.assert.calledThrice(errorLogSpy);
+
+    localstorageStub.restore();
+    errorLogSpy.restore();
```
jsnellbaker, line 61: Can you move the stubs (and the subsequent restores) into a before/after function surrounding this test?  This will help ensure these objects are properly handled in case the test fails.  If you don't want to have these stubs effect the other tests, please feel free to create another describe block to nest it altogether.

muuki88:
@Swiiip any update on this? For me this sounds rather critical

test/spec/unit/core/storageManager_spec.js
Swiiip: Ok 👍 

> Btw, in this case what is "third party context"? Is that just JS loaded from 3p domain? 
> > access window.localstorage in a third-party context
Indeed, 3rd-party context refers to scripts run inside an iframe whose src domain is not the 1p domain for instance.

Swiiip:
Hello sorry for the delay I'm trying to make the test pass on IE 11, the error seems linked to the way sinon mocks the localstorage property getters.

test/spec/unit/core/storageManager_spec.js
```diff
@@ -42,5 +42,34 @@ describe('storage manager', function() {
...
+  describe('localstorage forbidden access in 3rd-party context', function() {
+    let errorLogSpy;
+
+    beforeEach(function() {
+      var mock = {
+        get: function() {
+          throw Error;
+        }
+      };
+      Object.defineProperty(window, 'localStorage', mock);
```
jsnellbaker, line 56: Hi @Swiiip 

Can you remove this property as part of the `afterEach()`?  From how it looks now, this mocked function may be available for other tests in other files (since we generally run everything together).

Thanks

Swiiip: Done, I reset the property descriptor after each test

jsnellbaker (approved):

PR #5629, opened 2020-08-18
Allow hp:0 in supply chain nodes passed to Rubicon

mgriego:
## Type of change
- [x] Bugfix

## Description of change
Rubicon adapter does not pass a supply chain object of `hp` is set to `0`.


modules/rubiconBidAdapter.js
```diff
@@ -1166,7 +1166,7 @@ export function hasValidSupplyChainParams(schain) {
   if (!schain.nodes) return isValid;
   isValid = schain.nodes.reduce((status, node) => {
     if (!status) return status;
-    return requiredFields.every(field => node[field]);
+    return requiredFields.every(field => node.hasOwnProperty(field));
```
robertrmartinez, line 1169: 👍 Thanks for catching!

PR #5634, opened 2020-08-19
[IdentityLinkIdSystem] - pass tcfv2 consent string to envelope api

NemanjaRajkovic9:
## Type of change

- [x] Feature

## Description of change
Passing TCF v2 consent string to the envelope API.
Not calling envelope API if gdpr applies but consent string is missing.
Added gvlid.

### DIFF
--- a/modules/identityLinkIdSystem.js
+++ b/modules/identityLinkIdSystem.js
@@ -17,6 +17,11 @@ export const identityLinkSubmodule = {
    */
   name: 'identityLink',
   /**
+   * used to specify vendor id
+   * @type {number}
+   */
+  gvlid: 97,
+  /**
    * decode the stored id value for passing to bid requests
    * @function
    * @param {string} value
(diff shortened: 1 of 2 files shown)


modules/identityLinkIdSystem.js
```diff
@@ -39,8 +44,13 @@ export const identityLinkSubmodule = {
     }
     const hasGdpr = (consentData && typeof consentData.gdprApplies === 'boolean' && consentData.gdprApplies) ? 1 : 0;
     const gdprConsentString = hasGdpr ? consentData.consentString : '';
+    const tcfPolicyV2 = !!((consentData && consentData.vendorData && consentData.vendorData.tcfPolicyVersion && consentData.vendorData.tcfPolicyVersion === 2));
```
smenzer, line 47: could you use a `deepEquals()` here instead?

NemanjaRajkovic9: Hello,
did you maybe thought about `utils.deepAccess` instead of `utils.deepEqual` ?

smenzer: Ah yes, sorry, I did mean `deepAccess`:

```javascript
const tcfPolicyV2 = !!((utils.deepAccess(consentData, 'vendorData.tcfPolicyVersion') === 2))
```

NemanjaRajkovic9: No problem. We changed that to:
`const tcfPolicyV2 = utils.deepAccess(consentData, 'vendorData.tcfPolicyVersion') === 2;`
We pushed those changes.
If that is okay?

smenzer: yes, that should be fine. @mike-chowla will do the full review of the rest of the code, I just wanted to point out this optimization

PR #5641, opened 2020-08-20
updated userid module to stop caching the entire consent object

smenzer:
## Type of change
- [X] Bugfix

## Description of change
after the discussion at the end of https://github.com/prebid/Prebid.js/pull/5451, we agreed that we should store just a hash of the consent object rather than the entire object, since all we need it for is comparison purposes between the previous consent and the current consent for the user.

## Other information
https://github.com/prebid/Prebid.js/pull/5451#issuecomment-676203450

### DIFF
--- a/modules/userId/index.js
+++ b/modules/userId/index.js
@@ -232,7 +232,7 @@ function getStoredValue(storage, key = undefined) {
  * @param consentData
  * @returns {{apiVersion: number, gdprApplies: boolean, consentString: string}}
  */
-function makeStoredConsentDataObject(consentData) {
+function makeStoredConsentDataHash(consentData) {
   const storedConsentData = {
     consentString: '',
     gdprApplies: false,
@@ -245,7 +245,7 @@ function makeStoredConsentDataObject(consentData) {
     storedConsentData.apiVersion = consentData.apiVersion;
   }
 
-  return storedConsentData;
+  return utils.simpleHash(JSON.stringify(storedConsentData));
 }
 
 /**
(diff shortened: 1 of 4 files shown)


smenzer:
the CI fail seems unrelated to this PR, it's part of `terceptAnalyticsAdapter_spec.js:771`

modules/userId/index.js
```diff
@@ -287,7 +285,7 @@ function storedConsentDataMatchesConsentData(storedConsentData, consentData) {
   return (
     typeof storedConsentData === 'undefined' ||
     storedConsentData === null ||
-    utils.deepEqual(storedConsentData, makeStoredConsentDataObject(consentData))
+    storedConsentData == makeStoredConsentDataHash(consentData)
```
muuki88, line 288: Triple equals is probably safer?

muuki88 (approved):

modules/userId/index.js
smenzer: yea, looking into that.

smenzer: thanks for the suggestion...i've implemented this now

src/utils.js
```diff
@@ -1215,3 +1215,41 @@ export function mergeDeep(target, ...sources) {
...
+ * source: https://stackoverflow.com/a/52171480/845390
+ * @param str
+ * @param seed (optional)
+ * @returns {string}
+ */
+export function simpleHash(str, seed = 0) {
+  // IE doesn't support imul
+  // https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Math/imul#Polyfill
+  if (!Math.imul) {
+    Math.imul = function(opA, opB) {
```
robertrmartinez, line 1230: I was under the impression that we were not allowing polyfills.

I could be wrong but I thought I read / heard that somewhere...

src/utils.js
```diff
@@ -1215,3 +1215,43 @@ export function mergeDeep(target, ...sources) {
...
 }
+
+/**
+ * returns a hash of a string using a fast algorithm
+ * source: https://stackoverflow.com/a/52171480/845390
+ * @param str
+ * @param seed (optional)
+ * @returns {string}
+ */
+export function simpleHash(str, seed = 0) {
```
patmmccann, line 1226: any chancce you can call this cyrb53Hash or something describing what it is; the source you chose has many hashes.

PR #5642, opened 2020-08-20
remove cmpuishown event for tcf2 logic

jsnellbaker:
## Type of change
- [x] Other

## Description of change
A follow-up from #5564 - this change removes the event logic for `cmpuishown` as further review determined it's not really needed for Prebid.js.  

Please see this [comment](https://github.com/prebid/Prebid.js/pull/5564#issuecomment-673519661) from the previous ticket for additional context.

### DIFF
--- a/modules/consentManagement.js
+++ b/modules/consentManagement.js
@@ -100,11 +100,8 @@ function lookupIabConsent(cmpSuccess, cmpError, hookConfig) {
   function v2CmpResponseCallback(tcfData, success) {
     utils.logInfo('Received a response from CMP', tcfData);
     if (success) {
-      if (tcfData.gdprApplies === false) {
+      if (tcfData.gdprApplies === false || tcfData.eventStatus === 'tcloaded' || tcfData.eventStatus === 'useractioncomplete') {
         cmpSuccess(tcfData, hookConfig);
-      } else if (tcfData.eventStatus === 'tcloaded' || tcfData.eventStatus === 'useractioncomplete') {
-        cmpSuccess(tcfData, hookConfig);
-      }
     } else {
       cmpError('CMP unable to register callback function.  Please check CMP setup.', hookConfig);
     }
--- a/modules/consentManagement.js
+++ b/modules/consentManagement.js
@@ -104,8 +104,6 @@ function lookupIabConsent(cmpSuccess, cmpError, hookConfig) {
         cmpSuccess(tcfData, hookConfig);
       } else if (tcfData.eventStatus === 'tcloaded' || tcfData.eventStatus === 'useractioncomplete') {
         cmpSuccess(tcfData, hookConfig);
-      } else if (tcfData.eventStatus === 'cmpuishown') {
-        cmpSuccess(tcfData); // store consent, but don't continue to next hook
       }
     } else {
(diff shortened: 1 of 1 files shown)


modules/consentManagement.js
```diff
@@ -104,8 +104,8 @@ function lookupIabConsent(cmpSuccess, cmpError, hookConfig) {
         cmpSuccess(tcfData, hookConfig);
       } else if (tcfData.eventStatus === 'tcloaded' || tcfData.eventStatus === 'useractioncomplete') {
         cmpSuccess(tcfData, hookConfig);
-      } else if (tcfData.eventStatus === 'cmpuishown' && tcfData.tcString && tcfData.purposeOneTreatment === true) {
-        cmpSuccess(tcfData, hookConfig);
+      } else if (tcfData.eventStatus === 'cmpuishown') {
+        cmpSuccess(tcfData); // store consent, but don't continue to next hook
```
harpere, line 108: @jsnellbaker - thoughts on this change?   Thinking that if the cmpTimeout fires before the user makes a selection we'd at least have whatever consent (past choice or the configured default?) was sent with the "cmpuishown" event.   This also came from the discussion in PR #5564 that you referenced.

modules/consentManagement.js
```diff
@@ -104,8 +104,6 @@ function lookupIabConsent(cmpSuccess, cmpError, hookConfig) {
         cmpSuccess(tcfData, hookConfig);
       } else if (tcfData.eventStatus === 'tcloaded' || tcfData.eventStatus === 'useractioncomplete') {
         cmpSuccess(tcfData, hookConfig);
```
smenzer, line 106: Just a question...but why not combine these two if/else if statements into a single one since the result is the same?

```javascript
if (tcfData.gdprApplies === false || tcfData.eventStatus === 'tcloaded' || tcfData.eventStatus === 'useractioncomplete') {
  cmpSuccess(tcfData, finalCallback);
}
```

jsnellbaker: At this point, we can - I'll push an update.

smenzer (approved):
LGTM

modules/consentManagement.js
```diff
@@ -100,13 +100,8 @@ function lookupIabConsent(cmpSuccess, cmpError, hookConfig) {
   function v2CmpResponseCallback(tcfData, success) {
     utils.logInfo('Received a response from CMP', tcfData);
     if (success) {
-      if (tcfData.gdprApplies === false) {
+      if (tcfData.gdprApplies === false || tcfData.eventStatus === 'tcloaded' || tcfData.eventStatus === 'useractioncomplete') {
         cmpSuccess(tcfData, hookConfig);
```
smenzer, line 104: You removed one too many `}` I think

```suggestion
      if (tcfData.gdprApplies === false || tcfData.eventStatus === 'tcloaded' || tcfData.eventStatus === 'useractioncomplete') {
        cmpSuccess(tcfData, hookConfig);
      }
```

smenzer (changes requested):
whoops, looks like there's a syntax error

modules/consentManagement.js
jsnellbaker: fixed.

smenzer (approved):

PR #5651, opened 2020-08-23
Rtd extend bug fix

omerBrowsi:
## Type of change
- [ ] Bugfix

## Description of change
bugfix for - https://github.com/prebid/Prebid.js/pull/5519

the initSubModules function was called too late, causing the process not to work well with auction delay set (submodules was initiated after the auction)
changed the hook for the function call so the flow will work as expected


omerBrowsi:
added changes as requested here - 
https://github.com/prebid/prebid.github.io/pull/2268#issuecomment-680178361
(remove auction delay and related hook, change the name of `getData` function)

bretg:
There's a note in the docs PR https://github.com/prebid/prebid.github.io/pull/2268 that concerns me about RTD-core:

> Keep in mind it makes the auctionDelay param useless - meaning we will never delay the auction.
If the provider (submodule) will want to update the bidRequest with data that is not available immediately (e.g AJAX), he will have trouble doing it.

Of course RTD modules need the option of delaying the auction. Are you saying that somehow "getData()" supported a delay but updateBidRequest() doesn't?

If that's the case, then we need to (quickly) re-think the interface -- it's not useful without the ability to delay. What are the options? I'm out of my league, but the obvious approach would seem to be replacing updateBidRequest and auctionInit functions similar to addTargeting().

The requirement is that the sub-module be able to

- delay the auction for the allowed time
- inspect and update the bidRequest
- inspect and update the auction

smenzer:
any updates on this one?

bretg:
@omerBrowsi - do we still need this one with the 'phase 3' design we agreed on today?

omerBrowsi:
@bretg No, I will add the relevant code to the next commit

bretg:
Ok - will close this one then Thanks.

PR #5673, opened 2020-08-28
Price Floors: Fix bug when caching floor lookup

robertrmartinez:
## Type of change
- [X] Bugfix

## Description of change
When a floor is initially looked up, the input request params are saved as an offset so we do not have to run the lookup again.

When we find a cached lookup, we were passing the reference of the object, thus any currency adjustments were overwriting the cached value.

Now we pass back a copy of it.

### DIFF
--- a/modules/priceFloors.js
+++ b/modules/priceFloors.js
@@ -103,7 +103,7 @@ export function getFirstMatchingFloor(floorData, bidObject, responseObject = {})
   // if we already have gotten the matching rule from this matching input then use it! No need to look again
   let previousMatch = utils.deepAccess(floorData, `matchingInputs.${matchingInput}`);
   if (previousMatch) {
-    return previousMatch;
+    return {...previousMatch};
   }
   let allPossibleMatches = generatePossibleEnumerations(fieldValues, utils.deepAccess(floorData, 'schema.delimiter') || '|');
   let matchingRule = find(allPossibleMatches, hashValue => floorData.values.hasOwnProperty(hashValue));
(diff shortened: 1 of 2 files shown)


test/spec/modules/priceFloors_spec.js
```diff
@@ -182,6 +182,20 @@ describe('the price floors module', function () {
...
+      let inputData = {...basicFloorData};
+      [0.2, 0.4, 0.6, 0.8].forEach(modifier => {
+        let result = getFirstMatchingFloor(inputData, basicBidRequest, {mediaType: 'banner', size: '*'});
+        // result should always be the same
+        expect(result).to.deep.equal({
+          matchingFloor: 1.0,
+          matchingData: 'banner',
+          matchingRule: 'banner'
+        });
+        // make sure a post retrieval adjustment does not alter the cahced floor
```
leifwickland, line 195: Typo:
```suggestion
        // make sure a post retrieval adjustment does not alter the cached floor
```

modules/priceFloors.js
```diff
@@ -103,7 +103,7 @@ export function getFirstMatchingFloor(floorData, bidObject, responseObject = {})
   // if we already have gotten the matching rule from this matching input then use it! No need to look again
   let previousMatch = utils.deepAccess(floorData, `matchingInputs.${matchingInput}`);
   if (previousMatch) {
-    return previousMatch;
+    return {...previousMatch};
```
msm0504, line 106: Will modifications only be made at the top level of this object? This change will only make a shallow copy

robertrmartinez: Yes, this is a single object of only primitive types.

`Float`, `String`, `String`

robertrmartinez: Can use `utils.deepClone()` if you want though! maybe that makes more sense?

msm0504: No, what you have is fine, then

PR #5677, opened 2020-08-30
PubWise.io Analytics Module Update - SPOT Support, Module Rules & Minor Features

GLStephen:
## Type of change
- [ X ] Feature

## Description of change

- Refactor to make compliant with current module rules
- Addition of ActivationID
- Event batching, sessions, etc. to support Smart Path Optimization Technology
- Improvements to testing
- Support new event types

## Test Params
```
 pbjs.enableAnalytics({
  provider: 'pubwise',
  options: {
    site: 'b1ccf317-a6fc-428d-ba69-0c9c208aa61c'    
  }
 });
```

### DIFF
--- a/modules/pubwiseAnalyticsAdapter.js
+++ b/modules/pubwiseAnalyticsAdapter.js
@@ -14,8 +14,7 @@ const utils = require('../src/utils.js');
  pbjs.enableAnalytics({
   provider: 'pubwise',
   options: {
-    site: 'test-test-test-test',
-    endpoint: 'https://api.pubwise.io/api/v4/event/add/',
+    site: 'test-test-test-test'
   }
  });
 
--- a/modules/pubwiseAnalyticsAdapter.js
+++ b/modules/pubwiseAnalyticsAdapter.js
@@ -16,21 +16,21 @@ const storage = getStorageManager();
  pbjs.enableAnalytics({
   provider: 'pubwise',
   options: {
-    site: 'test-test-test-test'
+    site: 'b1ccf317-a6fc-428d-ba69-0c9c208aa61c'
   }
  });
 
 Changes in 4.0 Version
 4.0.1 - Initial Version for Prebid 4.x, adds activationId, adds additiona testing, removes prebid global in favor of a prebid.version const
+4.0.2 - Updates to include dedicated default site to keep everything from getting rate limited
 
 */
 
 const analyticsType = 'endpoint';
 const analyticsName = 'PubWise:';
 const prebidVersion = '$prebid.version$';
-let defaultUrl = 'https://api.pubwise.io/api/v5/event/default/';
 let pubwiseVersion = '4.0.1';
-let configOptions = {site: '', endpoint: defaultUrl, debug: ''};
+let configOptions = {site: '', endpoint: 'https://api.pubwise.io/api/v5/event/add/', debug: null};
 let pwAnalyticsEnabled = false;
 let utmKeys = {utm_source: '', utm_medium: '', utm_campaign: '', utm_term: '', utm_content: ''};
 let sessionData = {sessionId: '', activationId: ''};
@@ -110,7 +110,7 @@ function expireUtmData() {
 }
 
 function enrichWithCustomSegments(dataBag) {
-  // c_site: '', c_script_type: '', c_slot1: '', c_slot2: '', c_slot3: '', c_slot4: ''
+  // c_script_type: '', c_slot1: '', c_slot2: '', c_slot3: '', c_slot4: ''
   if (configOptions.custom) {
     if (configOptions.custom.c_script_type) {
       dataBag['c_script_type'] = configOptions.custom.c_script_type;
@@ -252,7 +252,7 @@ function filterAuctionInit(data) {
   return modified;
 }
 
-let pubwiseAnalytics = Object.assign(adapter({defaultUrl, analyticsType}), {
+let pubwiseAnalytics = Object.assign(adapter({analyticsType}), {
   // Override AnalyticsAdapter functions by supplying custom methods
   track({eventType, args}) {
     this.handleEvent(eventType, args);
@@ -317,8 +317,9 @@ pubwiseAnalytics.ensureSession = function () {
 pubwiseAnalytics.adapterEnableAnalytics = pubwiseAnalytics.enableAnalytics;
 
 pubwiseAnalytics.enableAnalytics = function (config) {
-  configOptions = config.options;
-  if (configOptions.debug === undefined) {
+  configOptions = Object.assign(configOptions, config.options);
+  // take the PBJS debug for our debug setting if no PW debug is defined
+  if (configOptions.debug === null) {
     configOptions.debug = utils.debugTurnedOn();
   }
   markEnabled();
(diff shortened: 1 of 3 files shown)


GLStephen:
I'm converting this to a draft while we figure out the Browserstack issue.

musikele (changes requested):
Test coverage is at 78%, lower than our accepted minimum that is 80%. Please try to improve line coverage the best you can

modules/pubwiseAnalyticsAdapter.js
```diff
@@ -54,72 +73,268 @@ function enrichWithUTM(dataBag) {
...
         newUtm = true;
         dataBag[prop] = utmKeys[prop];
       }
     }
 
     if (newUtm === false) {
       for (let prop in utmKeys) {
-        let itemValue = storage.getDataFromLocalStorage(`pw-${prop}`);
-        if (itemValue.length !== 0) {
+        let itemValue = localStorage.getItem(setNamespace(prop));
```
musikele, line 84: why you decided to remove prebid storageManager ?

modules/pubwiseAnalyticsAdapter.js
```diff
@@ -54,72 +73,268 @@ function enrichWithUTM(dataBag) {
...
+        if (itemValue !== null && typeof itemValue !== 'undefined' && itemValue.length !== 0) {
```
musikele, line 85: what about `if (itemValue)` ?

modules/pubwiseAnalyticsAdapter.js
```diff
@@ -54,72 +73,268 @@ function enrichWithUTM(dataBag) {
...
+  pwInfo(`Enabled`, configOptions);
+  pwAnalyticsEnabled = true;
+  setInterval(flushEvents, 100);
+}
+
+function pwInfo(info, context) {
+  utils.logInfo(`${analyticsName} ` + info, context);
+}
+
+/*
```
musikele, line 194: Consider removing unused code; it can be resumed from git history.

modules/pubwiseAnalyticsAdapter.js
```diff
@@ -54,72 +73,268 @@ function enrichWithUTM(dataBag) {
...
-    dataBag = enrichWithUTM(dataBag);
+    return newNoBidData;
   }
+*/
 
-  ajax(configOptions.endpoint, (result) => utils.logInfo(`${analyticsName}Result`, result), JSON.stringify(dataBag));
+function filterBidResponse(data) {
+  let modified = Object.assign({}, data);
+  // clean up some properties we don't track in public version
+  if (typeof modified.ad !== 'undefined') {
```
musikele, line 211: Alternative syntax to be considered: 
```
modified.ad = modified.ad || ''
``` 
warn that if `modified.ad` value can be `0`,  it will become `''`.

modules/pubwiseAnalyticsAdapter.js
```diff
@@ -54,72 +73,268 @@ function enrichWithUTM(dataBag) {
...
-    defaultUrl,
-    analyticsType
-  }),
-{
+function filterAuctionInit(data) {
+  let modified = Object.assign({}, data);
+
+  modified.refererInfo = {};
+  // handle clean referrer, we only need one
+  if (typeof modified.bidderRequests !== 'undefined') {
```
musikele, line 236: again, consider the more succint syntax `if (modified.bidderRequests && modified.bidderRequests[0] && ... )`

musikele (commented):
have a look at suggestions i left regarding the code.

GLStephen:
@musikele thanks for the review, the storageManager removal was a regression. We'll get that fixed and look at the other items.

modules/pubwiseAnalyticsAdapter.js
GLStephen: We don't rely on truthiness for data not under our control. Paranoid? Maybe? Unecessary? Maybe. We write in 6 different languages and would like to reliably get the meaning we want. Our desire is to determine if the values is, but often is not, defined and underlying behavior in various languages around what happens when the value is defined but "trueish" is variable. So, we make explicit statements when the value comes from somewhere else.

modules/pubwiseAnalyticsAdapter.js
GLStephen: We don't rely on truthiness for data not under our control. Paranoid? Maybe? Unnecessary? Maybe. We write in 6 different languages and would like to reliably get the meaning we want. Our desire is to determine if the values is, but often is not, defined and underlying behavior in various languages around what happens when the value is defined but "trueish" is variable. So, we make explicit statements when the value comes from somewhere else.

modules/pubwiseAnalyticsAdapter.js
GLStephen: I'd prefer not to change that. I'm relatively sure that code is the result of some edge cases.

modules/pubwiseAnalyticsAdapter.js
GLStephen: Removed

modules/pubwiseAnalyticsAdapter.js
GLStephen: This was a regression. The next update will have it re-added.

GLStephen:
@musikele Updates have been made. Let me know if you need anything further.

modules/pubwiseAnalyticsAdapter.js
```diff
@@ -54,72 +77,253 @@ function enrichWithUTM(dataBag) {
...
+  extendUserSessionTimeout();
+};
+
 pubwiseAnalytics.adapterEnableAnalytics = pubwiseAnalytics.enableAnalytics;
 
 pubwiseAnalytics.enableAnalytics = function (config) {
-  if (config.options.debug === undefined) {
-    config.options.debug = utils.debugTurnedOn();
-  }
   configOptions = config.options;
```
musikele, line 320: So, I have probably found an issue here. Based on the docs you've written at the top of this file, I added these lines to my test page: 
```
pbjs.enableAnalytics({
          provider: "pubwise",
          options: {
            site: "test-test-test-test",
          },
        });
```
when the execution gets here, `configOptions.endpoint` was set to `defaultUrl`, but that gets removed by this assignment: 

```
configOptions = config.options;
```

As a result, i see a bunch of errors in console: 

```
POST http://localhost:9999/integrationExamples/gpt/undefined 405 (Method Not Allowed)
(anonymous) @ ajax.js:91
flushEvents @ pubwiseAnalyticsAdapter.js:171
...
Prebid ERROR: xhr error null Method Not Allowed
```
Note the  **undefined** at the end of the url. 

So, `configOptions.endpoint` does not exist anymore. 

I think you wanted to extend configOptions with the data coming from function arguments: 
```
configOptions = {
    ...configOptions, 
    ...config.options  
}
```
(This can be also achieved by using Object.assign)

Can you have a look into this?

musikele (changes requested):
Hi @GLStephen , found another issue while testing this functionality. If I configure the adapter with default values as specified in docs, I get network errors in console. Can you have a look at it?

GLStephen:
@musikele thanks for the catch on that, I've also updated to use a reserved ID for testing so that the API will return proper responses, so the old test-test-test-test will now return an "over quota" response.

modules/pubwiseAnalyticsAdapter.js
GLStephen: @musikele this is resolved now

musikele (approved):
LGTM

PR #5686, opened 2020-09-01
GDPR Enforcement - Bugfix

Fawke:
Type of change

- [X] Bugfix

## Description of change

Resolves https://github.com/prebid/Prebid.js/issues/5615

This PR implements the alternate approach discussed here - https://github.com/prebid/Prebid.js/pull/5643

This PR makes an assumption:

- 1 moduleCode - 1 GVL ID. Essentially establishes 1-1 mapping between a moduleCode and a GVL ID.


modules/gdprEnforcement.js
```diff
@@ -47,58 +47,79 @@ const analyticsBlocked = [];
...
-  if (gvlMapping && gvlMapping[code]) {
-    gvlId = gvlMapping[code];
-  } else {
-    gvlId = adapterManager.getAnalyticsAdapter(code).gvlid;
+  let gvlid = null;
+  if (adapterManager.getAnalyticsAdapter(code) && adapterManager.getAnalyticsAdapter(code).gvlid) {
+    return adapterManager.getAnalyticsAdapter(code).gvlid
   }
-  return gvlId;
+  return gvlid;
```
smenzer, line 122: could this also be shortened to just one-line?

```
return adapterManager.getAnalyticsAdapter(code) && adapterManager.getAnalyticsAdapter(code).gvlid || null;
```

modules/gdprEnforcement.js
```diff
@@ -47,58 +47,79 @@ const analyticsBlocked = [];
...
+    if (bidder && bidder.getSpec) {
+      gvlid = bidder.getSpec().gvlid;
     }
   }
   return gvlid;
 }
 
 /**
- * Returns gvlId for userId module. If a userId modules does not have an associated gvlId, it returns 'undefined'.
+ * Returns GVL ID for an userId submodule. If an userId submodules does not have an associated GVL ID, it returns 'undefined'.
```
smenzer, line 104: same here, you're returning `null` when it's not found but comments mention it should be `undefined`. based on the tests, seems like the comments should be updated

modules/gdprEnforcement.js
```diff
@@ -47,58 +47,79 @@ const analyticsBlocked = [];
...
+      return gvlid;
+    }
+
+    gvlid = internal.getGvlidForBidAdapter(moduleCode) || internal.getGvlidForUserIdModule(module) || internal.getGvlidForAnalyticsAdapter(moduleCode);
+  }
+  return gvlid;
+}
+
+/**
+ * Returns GVL ID for a bid adapter. If the adapter does not have an associated GVL ID, it returns 'undefined'.
```
smenzer, line 87: if there's not a gvlid found, you're returning `null`, but this comment says it returns `undefined`. based on the tests, seems like the comments should be updated

modules/gdprEnforcement.js
```diff
@@ -47,58 +47,79 @@ const analyticsBlocked = [];
...
-  } else {
-    gvlId = userIdModule.gvlid;
-  }
-  return gvlId;
+  return (typeof userIdModule === 'object' ? userIdModule.gvlid : null);
 }
 
 /**
- * Returns gvlId for analytics adapters. If a analytics adapter does not have an associated gvlId, it returns 'undefined'.
+ * Returns GVL ID for an analytics adapter. If an analytics adapter does not have an associated GVL ID, it returns 'undefined'.
```
smenzer, line 113: same here...returning `null` instead of `undefined`, so you should update the comment i think

smenzer (changes requested):
a few corrections for the comments i think, but overall i like the approach

modules/gdprEnforcement.js
Fawke: Made the change. Was reluctant earlier because if the condition before `&&` returns false, the function will return `undefined` and not `null`. Kind of a minor inconsistency, but I don't think that'll play a big role.

modules/gdprEnforcement.js
Fawke: Yeah, I forgot to update the comments. Thanks, I updated all of them.

modules/gdprEnforcement.js
Fawke: done

modules/gdprEnforcement.js
Fawke: done

smenzer (approved):
LGTM

jaiminpanchal27 (approved):
LGTM

PR #5697, opened 2020-09-03
Documentation: update readme recommendation to install with 'npm ci'

GLStephen:
## Type of change

- [ x ] Refactoring (no functional changes, no api changes)

## Description of change
With the recent upgrade of npm we should be using the 'npm ci' option to install modules. This provides a more reliable deployment that does not update modules on install. It will reduce check-ins of improper package-lock.json and provide greater assurance that a stable node_modules is deployed.

### DIFF
--- a/README.md
+++ b/README.md
@@ -112,11 +112,11 @@ prebid.requestBids({
 
     $ git clone https://github.com/prebid/Prebid.js.git
     $ cd Prebid.js
-    $ npm install
+    $ npm ci
 
 *Note:* You need to have `NodeJS` 12.16.1 or greater installed.
 
-*Note:* In the 1.24.0 release of Prebid.js we have transitioned to using gulp 4.0 from using gulp 3.9.1.  To comply with gulp's recommended setup for 4.0, you'll need to have `gulp-cli` installed globally prior to running the general `npm install`.  This shouldn't impact any other projects you may work on that use an earlier version of gulp in its setup.
+*Note:* In the 1.24.0 release of Prebid.js we have transitioned to using gulp 4.0 from using gulp 3.9.1.  To comply with gulp's recommended setup for 4.0, you'll need to have `gulp-cli` installed globally prior to running the general `npm ci`.  This shouldn't impact any other projects you may work on that use an earlier version of gulp in its setup.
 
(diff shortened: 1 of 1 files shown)


mkendall07 (approved):

jsnellbaker:
@GLStephen Per your description, is there an ideal version of `npm` that we should recommend people to use to ensure the `npm ci` command works as expected?

Separately, have we updated our documents somewhere about how we should maintain the `package-lock.json` file going forward?  I may have just missed a PR that made the change, but can you confirm if that was already done?

I think this would be fine otherwise.  Thanks!

GLStephen:
@jsnellbaker I believe the version we have in the .nvmrc https://github.com/prebid/Prebid.js/blob/master/.nvmrc would ensure it's proper. The functionality was introduced a while back but our previous nvmrc version didn't have it. https://blog.npmjs.org/post/171556855892/introducing-npm-ci-for-faster-more-reliable

GLStephen:
reopening to unstale it

jsnellbaker:
To my knowledge, the nvmrc file is used to enforce a minimum version of node that's used in the project by NVM, not the version of NVM itself.  I think if we suggest at least NVM 6+, it should use the package-lock.json logic.

Further, has there been any updates on the package-lock.json maintenance as part of the release steps?

GLStephen:
@jsnellbaker sorry, I missed your reply - the minimum version of Node is the critical thing. I don't think the NVM version is relevant. It's NPM that needs to be at a certain version to have the "npm ci" command available.

I don't have any knowledge of the package-lock.json maintenance issues other than that addressed in this PR, which is about "npm install" updating packages in local builds causing unwanted package updates, package-lock.json udpates, etc.

jsnellbaker:
I think in general we should be periodically updating the package-lock.json file so that we do get the general bug fixes and other minor updates from the various packages used by Prebid.js.  

Whether this update is done with each release version or only when we need to update some dependent package listed in the package.json file, that's fine - we can discuss and decide that as a group.  The former has the benefit that it will keep the file in sync for that release and we'll know what goes in it; whereas the latter would be more organic and easier to maintain, at the potential risk of relying on someone else to insert the proper updates.

Overall though, if we're going to start recommending everyone to rely on the package-lock.json file for the `npm ci` command, updating this file is something that should happen.  There's a good chance otherwise it'll just be something that's forgotten and then has to be readdressed again when it somehow becomes a problem down the road.

This is my opinion.  Tagging some others to get their thoughts if this really matters; if they feel it's not an issue - I'll drop it.
cc @bretg @harpere @mkendall07

mkendall07:
hey everyone - this is been sitting here awhile. Does making the person responsible for the weekly release updating `package-lock.json` resolve @jsnellbaker  concern? 
If so we can probably update the release instructions with that step. 
@bretg @harpere @GLStephen

ChrisHuie (approved):

PR #5698, opened 2020-09-03
Rubicon analytics v2

robertrmartinez:
## Type of change
- [X] Feature

## Description of change
Adding functionality to the Rubicon Analytics adapter to work with session and googletag level data sets.


modules/rubiconAnalyticsAdapter.js
```diff
@@ -126,10 +135,11 @@ function sendMessage(auctionId, bidWonId) {
     });
   }
   let auctionCache = cache.auctions[auctionId];
-  let referrer = config.getConfig('pageUrl') || auctionCache.referrer;
+  let referrer = config.getConfig('pageUrl') || (auctionCache && auctionCache.referrer);
```
robertrmartinez, line 138: handle if this is undefinied in some cases

modules/rubiconAnalyticsAdapter.js
```diff
@@ -215,6 +226,30 @@ function sendMessage(auctionId, bidWonId) {
...
+      message.session = utils.pick(auctionCache.session, [
+        'id',
+        'pvid',
+        'start',
+        'expires'
+      ]);
+      if (auctionCache.session.fpkvs && Object.keys(auctionCache.session.fpkvs).length) {
+        message.fpkvs = Object.keys(auctionCache.session.fpkvs).map(key => {
+          return { key, value: auctionCache.session.fpkvs[key] };
+        });
```
robertrmartinez, line 249: mapping intoarray of objects of { key, value } requested by internal team

modules/rubiconAnalyticsAdapter.js
```diff
@@ -376,12 +492,18 @@ let rubiconAdapter = Object.assign({}, baseAdapter, {
         ]);
         cacheEntry.bids = {};
         cacheEntry.bidsWon = {};
-        cacheEntry.referrer = args.bidderRequests[0].refererInfo.referer;
+        cacheEntry.referrer = utils.deepAccess(args, 'bidderRequests.0.refererInfo.referer');
```
robertrmartinez, line 495: deep acess so we do not throw runtime undefined errors

modules/rubiconAnalyticsAdapter.js
```diff
@@ -318,6 +366,74 @@ function setRubiconAliases(aliasRegistry) {
...
+        // if this slot matches this bids adUnit, add the adUnit info
+        if (isMatchingAdSlot(bid.adUnit.adUnitCode)) {
+          bid.adUnit.gam = utils.pick(event, [
+            // these come in as `null` from Gpt, which when stringified does not get removed
+            // so set explicitly to undefined when not a number
+            'advertiserId', advertiserId => utils.isNumber(advertiserId) ? advertiserId : undefined,
+            'creativeId', creativeId => utils.isNumber(creativeId) ? creativeId : undefined,
+            'lineItemId', lineItemId => utils.isNumber(lineItemId) ? lineItemId : undefined,
+            'adSlot', () => event.slot.getAdUnitPath(),
+            'isSlotEmpty', () => event.isEmpty || undefined
```
robertrmartinez, line 429: WE only want to send this if slot.isEmpty is true.

If the slot is NOT empty, then there will be the advertiserIds etc, so not necessary to send.

test/spec/modules/rubiconAnalyticsSchema.json
```diff
@@ -293,7 +368,8 @@
             "success",
             "no-bid",
             "error",
-            "rejected"
+            "rejected-gdpr",
```
msm0504, line 371: Will the status ever be `rejected-gdpr`? I don't see this string used anywhere else in the code

msm0504 (approved):
Code changes look good. Just had 1 question on the test schema file

test/spec/modules/rubiconAnalyticsSchema.json
robertrmartinez: Not yet, but that is to come in another phase.

I can remove it though

modules/rubiconAnalyticsAdapter.js
```diff
@@ -376,12 +493,18 @@ let rubiconAdapter = Object.assign({}, baseAdapter, {
...
+        cacheEntry.referrer = utils.deepAccess(args, 'bidderRequests.0.refererInfo.referer');
         const floorData = utils.deepAccess(args, 'bidderRequests.0.bids.0.floorData');
         if (floorData) {
           cacheEntry.floorData = {...floorData};
         }
+        cacheEntry.gdprConsent = utils.deepAccess(args, 'bidderRequests.0.gdprConsent');
+        cacheEntry.session = storage.cookiesAreEnabled() && updateRpaCookie();
         cache.auctions[args.auctionId] = cacheEntry;
+        // register to listen to gpt events if not done yet
+        if (!cache.gpt.registered && utils.isGptPubadsDefined()) {
```
harpere, line 505: where is `cache.gpt.registered` set to `true`?

modules/rubiconAnalyticsAdapter.js
```diff
@@ -318,6 +367,74 @@ function setRubiconAliases(aliasRegistry) {
...
+    } catch (e) {
+      utils.logError(`Rubicon Analytics: Unable to decode ${COOKIE_NAME} value: `, e);
+    }
+  }
+  return {};
+}
+
+function setRpaCookie(decodedCookie) {
+  try {
+    storage.setCookie(COOKIE_NAME, window.btoa(JSON.stringify(decodedCookie)));
```
harpere, line 384: why using a `cookie` rather than `local storage`?

modules/rubiconAnalyticsAdapter.js
```diff
@@ -461,8 +590,8 @@ let rubiconAdapter = Object.assign({}, baseAdapter, {
         let auctionEntry = cache.auctions[args.auctionId];
         let bid = auctionEntry.bids[args.requestId];
         // If floor resolved gptSlot but we have not yet, then update the adUnit to have the adSlot name
-        if (!utils.deepAccess(bid, 'adUnit.adSlot') && utils.deepAccess(args, 'floorData.matchedFields.gptSlot')) {
-          bid.adUnit.adSlot = args.floorData.matchedFields.gptSlot;
+        if (!utils.deepAccess(bid, 'adUnit.gam.adSlot') && utils.deepAccess(args, 'floorData.matchedFields.gptSlot')) {
+          utils.deepSetValue(bid, 'adUnit.gam.adSlot', args.floorData.matchedFields.gptSlot);
```
harpere, line 594: not sending `pbAdSlot`?

modules/rubiconAnalyticsAdapter.js
```diff
@@ -479,7 +608,7 @@ let rubiconAdapter = Object.assign({}, baseAdapter, {
             delete bid.error; // it's possible for this to be set by a previous timeout
             break;
           case NO_BID:
-            bid.status = args.status === BID_REJECTED ? 'rejected' : 'no-bid';
+            bid.status = args.status === BID_REJECTED ? 'rejected-ipf' : 'no-bid';
```
harpere, line 611: why `-ipf`?  Is this the only reason a bid could be rejected?  Maybe we use a status called `BID_REJECTED_IPF` to make this more explicit?

harpere (commented):
few questions

modules/rubiconAnalyticsAdapter.js
robertrmartinez: https://github.com/prebid/Prebid.js/pull/5698/files#diff-7eda04fa0d03f759128e98228a4d9ed5R583

Set just above. 11 lines

modules/rubiconAnalyticsAdapter.js
robertrmartinez: There will be future rejected reasons, so updating with -ipf now.

But I'll make it a constant.

modules/rubiconAnalyticsAdapter.js
robertrmartinez: updated to localStorage

modules/rubiconAnalyticsAdapter.js
robertrmartinez: it is not, must have left it out or removed it accidentally! Thanks, good catch!

modules/rubiconAnalyticsAdapter.js
```diff
@@ -608,7 +612,7 @@ let rubiconAdapter = Object.assign({}, baseAdapter, {
             delete bid.error; // it's possible for this to be set by a previous timeout
             break;
           case NO_BID:
-            bid.status = args.status === BID_REJECTED ? 'rejected-ipf' : 'no-bid';
+            bid.status = args.status === BID_REJECTED ? BID_REJECTED_IPF : 'no-bid';
```
harpere, line 615: I was thinking `args.status === BID_REJECTED_IPF` as the constant.  Cause what if new code starts rejecting bids for a reason other than ipf?   But this would require updates in other places.

harpere: but we can fix this later.  for now this is ok

harpere (approved):

PR #5703, opened 2020-09-07
feat: add getEvents method to the public API

gpolaert:
## Type of change

- [ ] Bugfix
- [x] Feature
- [ ] New bidder adapter   
- [ ] Build related changes
- [ ] CI related changes
- [ ] Other

## Description of change

This PR adds  `getEvents` method to the public API.
Publishers and many of us, are spending a lot of time to understand and debug the auction flow.

This method adds the ability to retrieve the full history of the events. 
The `events.getEvents` method returns a copy of the event, it's a simple proxy of the event controller method.

- Documentation: https://github.com/prebid/prebid.github.io/pull/2322/

### DIFF
--- a/src/prebid.js
+++ b/src/prebid.js
@@ -1,7 +1,7 @@
 /** @module pbjs */
 
 import { getGlobal } from './prebidGlobal.js';
-import { flatten, uniques, isGptPubadsDefined, adUnitsFilter, isArrayOfNums } from './utils.js';
+import { adUnitsFilter, flatten, isArrayOfNums, isGptPubadsDefined, uniques } from './utils.js';
 import { listenMessagesFromCreative } from './secureCreatives.js';
 import { userSync } from './userSync.js';
 import { config } from './config.js';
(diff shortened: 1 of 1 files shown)


gpolaert:
I've attached the related PR for the documentation

gpolaert:
What's the next step @mkendall07? 
Can I help to do something?

bretg:
Not a problem with this PR, but in general, should we be concerned about how big this events array could get in auto-refresh scenarios? Seems to me there ought to be a cap on how many events are stored.

gpolaert:
Thanks @bretg 

As the new method is just a proxy of `events.getEvents()`, there is no new array to store the events.

But your concern is still true.
IMO there is no issue, even there are many events and refreshes. It may be an issue on long sessions and a single page application, but it's very unusual. I've already seen more than 10K refreshes on a page without trouble (even it doesn't make any sense).

Do you think it's valuable to open an new issue to track it?

mkendall07 (approved):
LGTM - needs a 2nd review and the circleCI tests to pass (i've restarted them)

jsnellbaker (approved):

PascalSalesch:
Image of dev tools memory usage

https://github.com/prebid/Prebid.js/pull/5703#issuecomment-693497356

Yes, I share your concern... :/

PR #5709, opened 2020-09-08
Adding sharedid to the pubcommon module

YerkovichM:
## Type of change

- [ ] Bugfix
- [X] Feature
- [ ] New bidder adapter   
- [ ] Build related changes
- [ ] CI related changes
- [ ] Other

## Description of change

Adding sharedid to pubcommon module
Now it will use the same format for passing id to bidders as sharedid:
{
   pubcid: {
       id:string(pubcid), third:string(sharedid)
    }
}
It gets sharedid from https://id.sharedid.org/id and stores it along with pubcId

### DIFF
--- a/modules/pubCommonIdSystem.js
+++ b/modules/pubCommonIdSystem.js
@@ -15,27 +15,25 @@ const PUB_COMMON_ID = 'PublisherCommonId';
 const MODULE_NAME = 'pubCommonId';
 
 const SHAREDID_OPT_OUT_VALUE = '00000000000000000000000000';
-const SHAREDID_COOKIE_EXPIRATION = 2419200000; // 28 days in milliseconds
-const SHAREDID_COOKIE_NAME = 'sharedid';
 const SHAREDID_URL = 'https://id.sharedid.org/id';
 
 const storage = getStorageManager(null, 'pubCommonId');
 
-function setCookieFromResponse() {
+function handleResponse(idObj, callback) {
   return {
     success: function (responseBody) {
-      let value = {};
       if (responseBody) {
         try {
           let responseObj = JSON.parse(responseBody);
           utils.logInfo('SharedId: Generated SharedId: ' + responseObj.sharedId);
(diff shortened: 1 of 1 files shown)


modules/pubCommonIdSystem.js
```diff
@@ -8,13 +8,59 @@
...
+  const bidIds = {
+    id: pubcId,
+  };
+  let sharedId = storage.getCookie(SHAREDID_COOKIE_NAME);
+  if (sharedId) {
+    if (sharedId != SHAREDID_OPT_OUT_VALUE) {
+      bidIds.third = sharedId;
+    }
+  } else {
+    ajax(SHAREDID_URL, setCookieFromResponse(), undefined, {method: 'GET', withCredentials: true});
```
smenzer, line 57: calls to retrieve new ids should be in the `getId` function, shouldn't they? the only place this `encodeId` function is called is from `decode`, but that should just be taking the stored value and formatting it properly for the module to pass on to bid adapters. if additional fields are needed, the `extendId` function can do that (to me, this is where the sharedId `third` should be added, not in decode).

YerkovichM: Moved from decode

pycnvr:
Won't this break bidders that are expecting pubcid to be a straightforward uuid?

jdwieland8282:
Thats a good point @pycnvr , let's take a step back and think about how we can add the Sharedid.org user id to the pubcommon module. Option 1. use the above, and ask bid adapters to update their bid adapter, this could be problematic if there are a large number of bid adapters integrated with PC.   Option 2, modify the current format, but pubcid as a string, and add a second sharedid string.  I'm open to suggestions too.
```
{
   pubcid:string(pubcid)
}
```

```
{
   pubcid:string(pubcid),
   third:string(sharedid)
}
```
or 

```
{
   pubcid:string(pubcid)
},
{
   sharedid:string(sharedid)
}
```

PR #5710, opened 2020-09-08
Add host to gulpfile

VasekProchazka:
## Type of change
- [x] Feature
- [x] Build related changes

## Description of change
Add host parameter in building server. When we run prebid.js in docker on mac (our sandbox), we need address 0.0.0.0, localhost does not work.

### DIFF
--- a/gulpfile.js
+++ b/gulpfile.js
@@ -110,6 +110,7 @@ function watch(done) {
   connect.server({
     https: argv.https,
     port: port,
+    host: argv.host,
     root: './',
     livereload: true
   });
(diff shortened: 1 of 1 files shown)


gulpfile.js
```diff
@@ -110,6 +110,7 @@ function watch(done) {
   connect.server({
     https: argv.https,
     port: port,
+    host: argv.host,
```
smenzer, line 113: what happens if `host` is not passed in as an argument? would it default to the same value used today?

VasekProchazka: You were right. I fixed it, so now it should have default value "localhost".

wqi1972 (approved):
LGTM

PR #5715, opened 2020-09-08
Prebid Server returns exp, not ttl

bretg:
# Type of change
- [ X ] Bugfix

## Description of change

Updating how pbsBidAdapter sets the "TTL" for bids. 

TTL in PBJS terms is how long the bid can stay in cache. However, the OpenRTB standard location for this value is `exp`.

Related to https://github.com/prebid/prebid-server/issues/1386

### DIFF
--- a/modules/prebidServerBidAdapter/index.js
+++ b/modules/prebidServerBidAdapter/index.js
@@ -822,9 +822,9 @@ const OPEN_RTB_PROTOCOL = {
           bidObject.meta = bidObject.meta || {};
           if (bid.adomain) { bidObject.meta.advertiserDomains = bid.adomain; }
 
-          // TODO: Remove when prebid-server returns ttl and netRevenue
           const configTtl = _s2sConfig.defaultTtl || DEFAULT_S2S_TTL;
-          bidObject.ttl = (bid.ttl) ? bid.ttl : configTtl;
+          // the OpenRTB location for "TTL" as understood by Prebid.js is "exp" (expiration).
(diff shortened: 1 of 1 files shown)


modules/prebidServerBidAdapter/index.js
```diff
@@ -822,9 +822,9 @@ const OPEN_RTB_PROTOCOL = {
           bidObject.meta = bidObject.meta || {};
           if (bid.adomain) { bidObject.meta.advertiserDomains = bid.adomain; }
 
-          // TODO: Remove when prebid-server returns ttl and netRevenue
           const configTtl = _s2sConfig.defaultTtl || DEFAULT_S2S_TTL;
-          bidObject.ttl = (bid.ttl) ? bid.ttl : configTtl;
+          // the OpenRTB location for "TTL" as understood by Prebid.js is "exp" (expiration).
+          bidObject.ttl = (bid.exp) ? bid.exp : configTtl;
```
robertrmartinez, line 827: This works.

Although I would just do 

```js
bidObject.ttl = bid.exp || configTtl;
```

But maybe I am not thinking of some use case as the previous code was doing the same thing.

robertrmartinez (approved):

GLStephen:
@smenzer this looks like it has two reviewers **and** it is assigned to me. What's the purpose of that?

smenzer:
Sorry @GLStephen i didn't notice the second reviewer. Will remove you.

mkendall07 (approved):

Issue #5687, opened 2020-09-02
Pass video.placement through to Prebid Server


bretg:
## Type of issue
Enhancement

## Description
Prebid Server doesn't get any indication of whether a video request is instream or outstream, and what kind of "placement" it is.

## Proposal

1) We modify the PrebidServerBidAdapter to support setting imp[].video.placement

- if AdUnit.mediaTypes.video.placement is defined send that as imp[].video.placement
- else if AdUnit.mediaTypes.video.context="instream", send imp[].video.placement=1

2) We modify the Prebid.org documentation to suggest defining placement when setting outstream.

https://docs.prebid.org/prebid-video/video-getting-started.html
https://docs.prebid.org/dev-docs/show-outstream-video-ads.html

mike-chowla:
Approach looks to good to me.

PR #5716, opened 2020-09-08
PBS Bid Adapter: Pass video.placement when context is instream (#5687)

robertrmartinez:
#5687 
## Type of change
- [X] Other

## Description of change
If adUnit is video and context is instream and placement is not defined, set it to 1.

### DIFF
--- a/test/spec/modules/prebidServerBidAdapter_spec.js
+++ b/test/spec/modules/prebidServerBidAdapter_spec.js
@@ -509,6 +509,22 @@ describe('S2S Adapter', function () {
       expect(requestBid.imp[0].video).to.not.exist;
     });
 
+    it('should default video placement if not defined and instream', function () {
+      let ortb2Config = utils.deepClone(CONFIG);
+      ortb2Config.endpoint = 'https://prebid.adnxs.com/pbs/v1/openrtb2/auction';
+
+      config.setConfig({ s2sConfig: ortb2Config });
+
+      let videoBid = utils.deepClone(VIDEO_REQUEST);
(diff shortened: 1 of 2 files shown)


bretg (approved):

test/spec/modules/prebidServerBidAdapter_spec.js
```diff
@@ -509,6 +509,22 @@ describe('S2S Adapter', function () {
       expect(requestBid.imp[0].video).to.not.exist;
     });
 
+    it('should default video placement if not defined and instream', function () {
```
Fawke, line 512: Shouldn't the test description be, "should set default video placement if "placement" not defined in instream".
What happens if context is Outstream? I understand, in case of instream, the value of `mediaTypes.video.placement` should be `1`, but,  in the case of Outstream, what should be the value of`mediaTypes.video.placement`.

bretg:
> in the case of Outstream, what should be the value of mediaTypes.video.placement be?

The Video Taskforce is going to review all video params (including placement) and document those that need to be added to the current list of required fields. ^^ @mike-chowla 

In the meantime, I don't think any docs are necessary for this PR.

Fawke (approved):

test/spec/modules/prebidServerBidAdapter_spec.js
robertrmartinez: > in the case of Outstream, what should be the value ofmediaTypes.video.placement

It should not default / add anything if context is outstream

PR #5719, opened 2020-09-09
Chrome 85.0.4183 fix for SameSite=None cookie rejection policy

bansawbanchee:
Tests for this Analytics Adapter fail **locally** without this change when a user is using Chrome 85 or newer. Chrome 85 added rejection of insecure SameSite=None cookies. `secure` probably could have been added to fix this issue but `SameSite=Lax` works as well for the test case.

https://developers.google.com/web/updates/2020/07/chrome-85-deps-rems
https://en.wikipedia.org/wiki/Google_Chrome_version_history

## Type of change

- [ ] Bugfix
- [ ] Feature
- [ ] New bidder adapter   
- [ ] Build related changes
- [x] CI related changes
- [ ] Other

### DIFF
--- a/test/spec/modules/fintezaAnalyticsAdapter_spec.js
+++ b/test/spec/modules/fintezaAnalyticsAdapter_spec.js
@@ -12,7 +12,7 @@ function setCookie(name, value, expires) {
   document.cookie = name + '=' + value +
     '; path=/' +
     (expires ? ('; expires=' + expires.toUTCString()) : '') +
-    '; SameSite=None';
+    '; SameSite=Lax';
 }
 
 describe('finteza analytics adapter', function () {


smenzer:
Shouldn't you be using storageManager rather than directly writing cookies via document.cookie?

bansawbanchee:
This is not our adapter. The tests for this adapter fail locally when I am trying to run `gulp serve` to test our userID submodule because my installed version of Chrome (85) on Mac is newer then the version Prebid BrowserStack tests against (Chrome 80). 

Chrome 85 and newer includes the rejection for SameSite=None without secure flag so anybody who attempts to run tests locally on Chrome 85 or newer will receive these failed tests for this adapter.

I am only including this change to ensure the tests pass.

patmmccann:
LGTM

patmmccann (approved):

smenzer (approved):
this adapter probably needs to update to use storageManager, but as this is a quick fix to get tests working again, I'll merge it.

Issue #5720, opened 2020-09-09
Support arbitrary OpenRTB in s2sConfig


bretg:
## Type of issue
Enhancement

## Description
There's a desire to pass arbitrary values through the Prebid Server Bid Adapter. Currently there are a couple of piece-meal ways to pass values:

- `setConfig({s2sConfig.extPrebid})` option -- limits items to the `ext.prebid` object.
- `setConfig({site})` -- allows setting of openrtb.site values

## Proposal

1. Support an `openRtb` object on the s2sConfig - deep merge it with the final results of OpenRTB2 JSON prepared for transmitting to Prebid Server. Values here take precedence over everthing else.
2. Support an `openRtbImp` object on the s2sConfig - deep merge it with each imp object in the OpenRTB2 JSON

This is a powerful tool and can be be abused -- documentation should warn users to be cautious.

Example:
```
pbjs.setConfig({
    s2sConfig: {
        ...
        openRtbImp: {
            bidfloor: 1.0,
            bidfloorcur: "EUR"
        },
        openRtb: {
           badv: ["adv1.com", "adv2.com"]
        }
    }
});
```

bretg:
@patmmccann and @bszekely1 - setting `bidfloor` globally across all Imps/AdUnits doesn't seem to be the right thing to do here.

Prebid Server could do this, but it would probably take longer to get both PBS-Java and PBS-Go updated.

Do you want to specify a more surgical way of defining the floor-per-imp here?

bszekely1:
While i get the complexities of obfuscating the nuances of OpenRTB behind some friendly API, certain attributes should be applied at the impression grain and other at the top level grain. Bidfloor is on that should be at the imp level.

I would recommend to just create one object called OpenRTB and let people pass a structured object in. It's the most scalable method available.

bretg:
Closing this issue -- this is now overlapping with #5795 . That issue also covers how to add per-imp attributes.

PR #5721, opened 2020-09-09
add floor module support to prebid server

patmmccann:
## Type of change

- [ ] Bugfix
- [X] Feature
- [ ] New bidder adapter   
- [ ] Build related changes
- [ ] CI related changes
- [ ] Other

## Description of change
this adds floor and floor curreny to prebid server requests in field imp[].bidfloor and imp[].bidfloorcur

### DIFF
--- a/modules/prebidServerBidAdapter/index.js
+++ b/modules/prebidServerBidAdapter/index.js
@@ -276,7 +276,7 @@ function _appendSiteAppDevice(request, pageUrl) {
 }
 
 function _getFloor(bid) {
-  let floor = null;
+  let floor = 0;
   if (typeof bid.getFloor === 'function') {
     const floorInfo = bid.getFloor({
       currency: DEFAULT_S2S_CURRENCY,
--- a/modules/prebidServerBidAdapter/index.js
+++ b/modules/prebidServerBidAdapter/index.js
@@ -593,7 +593,7 @@ const OPEN_RTB_PROTOCOL = {
         utils.deepSetValue(imp, 'ext.prebid.storedauctionresponse.id', storedAuctionResponseBid.storedAuctionResponse.toString());
       }
 
-      _appendFloor(bidRequests[0], imp);
+      _appendFloor(adUnit.bids[0], imp);
 
       if (imp.banner || imp.video || imp.native) {
         imps.push(imp);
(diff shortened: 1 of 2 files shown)


modules/prebidServerBidAdapter/index.js
```diff
@@ -569,6 +593,8 @@ const OPEN_RTB_PROTOCOL = {
         utils.deepSetValue(imp, 'ext.prebid.storedauctionresponse.id', storedAuctionResponseBid.storedAuctionResponse.toString());
       }
 
+      _appendFloor(firstBidRequest.bids[0], imp);
```
patmmccann, line 596: I can't figure out what should go here in the first argument

robertrmartinez: Yeah, we'll need some logic to find the right bid to use.

Need's the context of the adUnit, so we'll want to use `adUnit.code`, and find a bid in `firstBidRequest.bids` which matches it.

Something like:

```js
let matchedBid = find(firstBidRequest.bids, bid => bid.adUnitCode === adUnit.code);
if (matchedBid) {
   _appendFloor(matchedBid, imp);
}
```

An issue even this may have:

- `firstBidRequest` may be from a bidder who is not included in all adUnits, so we may miss out on setting floor for some imps.

Which means we may need / want to do a join on ALL bids not just the ones in `firstBidRequest`.

I would have to take a deeper look, but could come up with something.

@bszekely any input on using PBJS floors module to set `imp.bidFloor`?

I know PBS will eventually have some floor module support, but may be some time from now.

patmmccann: thanks so much for taking a look! @bszekely1 and I were talking about this in the prebid server meeting recently and it seems this is quite a valid use case. @bretg recently filed https://github.com/prebid/Prebid.js/issues/5720, which this partially solves, partly based on that discussion. Also all prebid server adapters are being asked to honor this parameter when it is available (imp[].bidfloor and imp[].bidlfoorcur) instead of reading from adapterOptions, but there is no way to currently write to it from js.

bretg:
@patmmccann - what's the status of this one? It's marked as "draft"

patmmccann:
I couldn't figure it out :(

bretg:
Ok - I'll work towards getting this prioritized for next internal Sprint

patmmccann:
@bretg any luck putting this on the dev calendar? With the prebid 5.0 issue announced, I think it would be great to have the pbs adapter in compliance

robertrmartinez:
Closing this as I have opened https://github.com/prebid/Prebid.js/pull/6273

PR #5725, opened 2020-09-10
Upgrade Circle CI resource

Fawke:
## Type of change
- [X] CI related changes

## Description of change
We were using less resource on our circle ci container (2 cpus and 4GB RAM). I have upgraded it to use (8cpu and 16GB RAM). The pipeline is executing a little bit faster now.

Here's a comparison:

### Before:

Screenshot 2020-09-10 at 6 40 08 PM

### After:

Screenshot 2020-09-10 at 6 40 57 PM

<br/>

Since the tests run on Browserstack, the timings are also limited by the resource we've available there, and if two or more Circle CI jobs are running in parallel, the tests will timeout or take longer than usual. That explains one test running ~ 16m even after the upgrade.

## NOTE: 
This PR does not have any effect on the tests that are currently failing intermittently.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -8,7 +8,7 @@ aliases:
       docker:
         # specify the version you desire here
         - image: circleci/node:12.16.1
-        
+      resource_class: xlarge
         # Specify service dependencies here if necessary
         # CircleCI maintains a library of pre-built images
         # documented at https://circleci.com/docs/2.0/circleci-images/


jsnellbaker (approved):
I think this is okay.

@mkendall07 tagging you for 2nd review (in case we should double-check anything with the circleci support).

mkendall07 (approved):
LGTM - we have a ton of circleCI credits so this should be fine.

PR #5732, opened 2020-09-11
Update bid param checking rule for bridgewellBidAdapter

rigelbibi:
## Type of change

- [x ] Feature

## Description of change

Update bid param checking rule for bridgewellBidAdapter


smenzer:
please remove the `package-lock.json` from this PR, it should not be included. thanks

PR #5738, opened 2020-09-14
Adnow bidder

vingood:
## Type of change

- [ ] Bugfix
- [ ] Feature
- [x] New bidder adapter   
- [ ] Build related changes
- [ ] CI related changes
- [ ] Other

## Description of change
New bidder adapter: adnow.

```
{
  bidder: 'adnow',
  params: {
    codeId: 794934
  }
}
```

- support@adnow.com

- A link to a PR on the docs repo at https://github.com/prebid/prebid.github.io/pull/2314

### DIFF
--- a/test/spec/modules/adnowBidAdapter_spec.js
+++ b/test/spec/modules/adnowBidAdapter_spec.js
@@ -20,23 +20,67 @@ describe.only('adnowBidAdapterTests', function () {
     });
   });
 
-  it('buildRequests', function() {
-    const bidRequestData = [{
-      bidId: 'bid12345',
-      params: {
-        codeId: 12345
-      }
-    }];
+  describe('buildRequests', function() {
+    it('Common settings', function() {
+      const bidRequestData = [{
+        bidId: 'bid12345',
+        params: {
+          codeId: 12345
+        }
+      }];
 
-    const req = spec.buildRequests(bidRequestData);
-    const reqData = req[0].data;
+      const req = spec.buildRequests(bidRequestData);
+      const reqData = req[0].data;
 
-    expect(reqData)
-      .to.match(/Id=12345/)
-      .to.match(/mediaType=native/)
-      .to.match(/out=prebid/)
-      .to.match(/requestid=bid12345/)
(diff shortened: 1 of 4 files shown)


jsnellbaker (changes requested):
Hi @vingood 

When I trying to test the bidder (using the params in the description), I was getting a bid response from the server but the cpm value was set to `0`.  This was causing the bid to be rejected by the logic you have in the `interpretResponse` function.  

Can you update the price on the test bid to a higher value so I can test the end-to-end workflow?

Thanks!

vingood:
Please review again. Thank you.

jsnellbaker (changes requested):
@vingood 
Thanks for making the update on the test bid data.  

However, I noticed that the `width` and `height` properties are not being defined properly in the bid object created for Prebid.  This is causing the banner ad to not physically appear on the page since it has a default `width` and `height` of `0x0`.

I see the properties are defined in the raw response from the ad server.  Can you add the logic in the adapter code to read these properties and assign them to the bid object?

Thanks.

test/spec/modules/adnowBidAdapter_spec.js
```diff
@@ -0,0 +1,161 @@
+import { expect } from 'chai';
+import { spec } from 'modules/adnowBidAdapter.js';
+
+describe.only('adnowBidAdapter', function () {
```
jsnellbaker, line 4: Please remove the `.only` from your describe function here.

test/spec/modules/adnowBidAdapter_spec.js
```diff
@@ -0,0 +1,161 @@
...
+        }
+      }];
+
+      const req = spec.buildRequests(bidRequestData);
+      const reqData = req[0].data;
+
+      expect(reqData).to.match(/sizes=300x250/);
+    });
+
+    it.only('Native sizes', function () {
```
jsnellbaker, line 62: Please remove the `.only` from your it function here as well.

jsnellbaker (approved):
Thanks for making the additional updates.  LGTM

PR #5739, opened 2020-09-14
Handling video outstream in smartadserver adapter.

tadam75:
## Type of change
- [x] Feature

## Description of change
Adding video outstream support in SmartAdServer adapter.

### DIFF
--- a/modules/smartadserverBidAdapter.md
+++ b/modules/smartadserverBidAdapter.md
@@ -95,3 +95,40 @@ Please reach out to your Technical account manager for more information.
         }]
     };
 ```
+
+## Outstream Video
+
+```
+    var outstreamVideoAdUnit = {
+			code: 'test-div',
+			mediaTypes: {
+				video: {
+					context: 'outstream',
+					playerSize: [640, 480]
+				}
+			},
+			renderer: {
+				url: 'https://acdn.adnxs.com/video/outstream/ANOutstreamVideo.js',
+				render: function (bid) {
+					ANOutstreamVideo.renderAd({
+						targetId: bid.adUnitCode,
+						adResponse: bid
+					});
+				}
+			},
+			bids: [{
+				bidder: "smart",
+				params: {
+					domain: 'https://prg.smartadserver.com',
+					siteId: 207435,
+					pageId: 896536,
+					formatId: 85089,
+					bidfloor: 5,
+                    video: {
+                        protocol: 6,
+                        startDelay: 1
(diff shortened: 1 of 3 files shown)


mlapeyre3:
Hello @mike-chowla, did you have time to review our PR?
Thanks!

mlapeyre3:
Hllo @gglas / @mike-chowla any news? Can we expect a review of our PR next week? I need to inform our partners about it, just need some visibility :)
Thanks!

modules/smartadserverBidAdapter.md
```diff
@@ -94,4 +94,41 @@ Please reach out to your Technical account manager for more information.
...
+					playerSize: [640, 480]
+				}
+			},
+			renderer: {
+				url: 'https://acdn.adnxs.com/video/outstream/ANOutstreamVideo.js',
+				render: function (bid) {
+					ANOutstreamVideo.renderAd({
+						targetId: bid.adUnitCode,
+						adResponse: bid
+					});
```
jsnellbaker, line 116: I think an update is needed here.  In the documentation [here](https://docs.prebid.org/dev-docs/show-outstream-video-ads.html#renderers), it has the actual renderer call wrapped around a queue handler to ensure the called gets executed at the right time.

When I was testing with these exact params, I was getting an error during the rendering step that the `ANOustreamVideo` did not exist.  I think it just means this call here was happening too early before the script fully deployed.  Can you make an update here (just so the example is right)?

jsnellbaker (changes requested):
Hi @tadam75 

When I was first testing the changes with the example params, I was getting a bid back.  However after a few tries, the bid stopped coming back.  Is there some type of capping put in place?  If so, is there some way I can clear it?

Please confirm when you have the chance.

modules/smartadserverBidAdapter.md
tadam75: Thanks for the doc, I fixed it.

tadam75:
> Hi @tadam75
> 
> When I was first testing the changes with the example params, I was getting a bid back. However after a few tries, the bid stopped coming back. Is there some type of capping put in place? If so, is there some way I can clear it?
> 
> Please confirm when you have the chance.

Indeed we have a kind of throttling server side, that we can't really override 😔
It didn't occur to me during my tests but it can definitely happen.

jsnellbaker (approved):
Thanks for the clarification, I got it to work.  LGTM

PR #5750, opened 2020-09-15
Britepool user id module update

bansawbanchee:
## Type of change

- [ ] Bugfix
- [ ] Feature
- [ ] New bidder adapter   
- [ ] Build related changes
- [ ] CI related changes
- [ ] Other

## Description of change

Adding gdpr consent string support and dynamic params support.

- prebid@britepool.com

### DIFF
--- a/modules/britepoolIdSystem.js
+++ b/modules/britepoolIdSystem.js
@@ -29,7 +29,7 @@ export const britepoolIdSubmodule = {
   /**
    * Performs action to obtain id and return a value in the callback's response argument
    * @function
-   * @param {SubmoduleParams} [configParams]
+   * @param {SubmoduleParams} [submoduleConfigParams]
    * @param {ConsentData|undefined} consentData
    * @returns {function(callback:function)}
    */
@@ -45,6 +45,9 @@ export const britepoolIdSubmodule = {
           id: britepoolIdSubmodule.normalizeValue(getterResponse)
         };
       }
+    }
+    if (utils.isEmpty(params)) {
+      utils.triggerPixel(PIXEL);
     }
     // Return for async operation
     return {
@@ -67,7 +70,6 @@ export const britepoolIdSubmodule = {
         } else {
           ajax(url, {
             success: response => {
-              utils.triggerPixel(PIXEL);
               const responseObj = britepoolIdSubmodule.normalizeValue(response);
               callback(responseObj ? { primaryBPID: responseObj.primaryBPID } : null);
             },
@@ -82,7 +84,7 @@ export const britepoolIdSubmodule = {
   },
   /**
    * Helper method to create params for our API call
-   * @param {SubmoduleParams} [configParams]
+   * @param {SubmoduleParams} [submoduleConfigParams]
    * @param {ConsentData|undefined} consentData
    * @returns {object} Object with parsed out params
    */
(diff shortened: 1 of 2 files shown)


modules/britepoolIdSystem.js
```diff
@@ -65,6 +67,7 @@ export const britepoolIdSubmodule = {
         } else {
           ajax(url, {
             success: response => {
+              utils.triggerPixel(PIXEL);
```
smenzer, line 70: what is this pixel fire for?

bansawbanchee: It's for id resolution when authoritative information is not present. 

I've updated the code to ensure that this happens only when params are not present as well as included tests. Please review my latest commit with these changes.

smenzer: ok, so essentially if a publisher doesn't provide any params, you fire a pixel that seems to redirect to a thrtle.com endpoint (which gives me a 403 error), before calling your britepool endpoint like usual (just without any specific publisher params)?

bansawbanchee: > ok, so essentially if a publisher doesn't provide any params, you fire a pixel that seems to redirect to a thrtle.com endpoint

Yes, this is what we are doing.

Are you sure the pixel (https://px.britepool.com) is giving you a 403 and not the usual request (https://api.britepool.com)? The usual request doesn't require any params but will return a 403 if it cannot detect a bpid cookie on the endpoint, in which case it requires the api_key, which is why you are receiving a 403.

smenzer: if i call the pixel (https://px.britepool.com/new?partner_id=t) it returns a 302 to https://thrtle.com/insync?vxii_pid=10054&vxii_pdid=dcb6367c-186b-4a1d-ad3e-6d94955c5c4a which then issues a 403.

bansawbanchee: So I've spent a couple hours testing every possible scenario/browser/privacy setting and cannot reproduce with the latest commit.

We have this pixel outside of prebid and haven't seen this issue. However, I have asked throtle in what cases would we get a 403 back from their pixel.

smenzer: weird. i just tried in an incognito browser and got the same thing. i'm based in France, so maybe it's a GDPR thing? if so, you should make sure that you only fire the pixel when there's consent (and with consent strings if necessary for throtle)

bansawbanchee: ok so throtle got back to use this morning and the endpoint is US only ... they said EU request would give a 403. Seems like an appropriate status code given it is a Forbidden request.

smenzer: @bansawbanchee ok...should you put in some checks to make sure it doesn't fire outside of the US? maybe on your end you don't redirect or something? in any case, since it is a known restriction and I don't believe you operate outside of the US today, this doesn't seem like a blocker to me, although NOT having a 403 in any case would be ideal, so I'd suggest you work on a fix for that going forward.

smenzer (approved):
LGTM, just need to get CI tests passing...I've restarted them to see if they pass

modules/britepoolIdSystem.js
bansawbanchee: Agreed, not a blocker. We will resolve the 403 server side in the near future.

bansawbanchee:
Can you run the Ci tests again? @smenzer

smenzer:
@bansawbanchee i've re-run the circe-ci tests several times and they still aren't passing. i've reached out in the prebid.js slack channel to see if someone else can take a look for me. the only code I see that maybe is causing this is that you're defining a value on `window`. maybe try deleting the value rather than just setting it to `undefined` at the end of your tests?

smenzer:
@Fawke has a magic touch and got the tests to pass, so I'm good with this now.

Issue #5683, opened 2020-08-31
Force refresh userId


TLadd:
## Type of issue
Feature Request

## Description

I would like for there to be a way to force refresh userId values. Currently, if I update a stored userId value in a cookie or localStorage, that new value will get pulled in on future page loads, but not the current one. The old value remains since submodules [only get initialized once](https://github.com/prebid/Prebid.js/blob/40e5c7da5ee1ead9b29de8072a91759de214452e/modules/userId/index.js#L384), which is when these values are read.

The use cases I have where this would be useful all boil down to being able to provide a better id sometime after the initial page load. For instance, after a user logs in, I can get a better identifier than was previously stored and I want to start using it in bids without refreshing the page.

I would propose something along these lines:

```typescript
interface RefreshUserIdsOptions {
  // When performing the refresh, should it take the stored value, or force overriding it.
  // If the new value was written to the cookie/localStorage, it wouldn't be necessary to call `getId`.
  // True if you want to override the stored value and call `getId`.
  overrideStoredValue: boolean,
}

pbjs.refreshUserIds(options: RefreshUserIdsOptions)
pbjs.refreshUserId(name: string, options: RefreshUserIdsOptions)
```

GLStephen:
We have been looking at an option to reinitialize analytics for SPAs. It seems there is a cluster of information generally thought of as occuring on page load that should have a "reinitialization" hook.

bretg:
@jdwieland8282  - please take a look at this request and consider assigning if you agree with it

jdwieland8282:
Hi @TLadd, just want to make sure I understand this request. User ids available post page load, like ones derived from a login event, will be more persistent but are not available to the getId function bc, that function renders on page load.  The request is to design a feature that "reinitialization's" User ids, post page load?

TLadd:
@jdwieland8282 Yep, that's correct. The goal is to be able to refresh a userId value that gets sent with bids without having to refresh the page. 

[...]

PR #5752, opened 2020-09-15
Added global function for refreshing user id's

coryhammon1:
## Type of change

- [ ] Bugfix
- [ x] Feature
- [ ] New bidder adapter   
- [ ] Build related changes
- [ ] CI related changes
- [ ] Other

## Description of change

Added a global function to refresh user id's after initialization on page load. This seems to be the simplest way to get this functionality without adding more complexity into the user id module itself. Essentially, the function triggers a reinitialization of all user id modules, so if there were any changes to the stored values after page load, those new values will be included in the any bid requests.

```
{
  bidder: '<bidder name>',
  params: {
    // ...
  }
}
```

## Other information

https://github.com/prebid/Prebid.js/issues/5683

### DIFF
--- a/modules/userId/index.js
+++ b/modules/userId/index.js
@@ -377,12 +377,11 @@ function addIdDataToAdUnitBids(adUnits, submodules) {
 /**
  * This is a common function that will initialize subModules if not already done and it will also execute subModule callbacks
  */
-
-function initializeSubmodulesAndExecuteCallbacks(continueAuction, refresh) {
+function initializeSubmodulesAndExecuteCallbacks(continueAuction) {
   let delayed = false;
 
   // initialize submodules only when undefined
-  if (typeof initializedSubmodules === 'undefined' || !!refresh) {
+  if (typeof initializedSubmodules === 'undefined') {
     initializedSubmodules = initSubmodules(submodules, gdprDataHandler.getConsentData());
     if (initializedSubmodules.length) {
       // list of submodules that have callbacks that need to be executed
@@ -468,9 +467,41 @@ function getUserIdsAsEids() {
 
 /**
 * This function will be exposed in the global-name-space so that userIds can be refreshed after initialization.
+* @param {String[]} submoduleNames submodules to be refreshed
 */
-function refreshUserIds() {
-  initializeSubmodulesAndExecuteCallbacks(undefined, true);
+function refreshUserIds(submoduleNames) {
+  if (!utils.isArray(submoduleNames)) {
+    return;
+  }
+
+  if (!initializedSubmodules) {
+    utils.logWarn(`${MODULE_NAME} - refresh called before submodule initialization`);
+    return;
+  }
+
+  let consentData = gdprDataHandler.getConsentData()
+
+  const storedConsentData = getStoredConsentData();
+  setStoredConsentData(consentData);
+
+  let callbackSubmodules = [];
+
+  for (let submoduleName of submoduleNames) {
+    for (let submodule of initializedSubmodules) {
+      if (submodule.submodule.name === submoduleName) {
+        utils.logInfo(`${MODULE_NAME} - refreshing ${submodule.submodule.name}`);
+        populateSubmoduleId(submodule, consentData, storedConsentData, true);
+
+        if (utils.isFn(submodule.callback)) {
+          callbackSubmodules.push(submodule);
+        }
+      }
+    }
+  }
+
+  if (callbackSubmodules.length > 0) {
+    processSubmoduleCallbacks(callbackSubmodules);
+  }
 }
 
 /**
@@ -479,6 +510,57 @@ function refreshUserIds() {
 export const validateGdprEnforcement = hook('sync', function (submodules, consentData) {
   return {userIdModules: submodules, hasValidated: consentData && consentData.hasValidated};
 }, 'validateGdprEnforcement');
+
+function populateSubmoduleId(submodule, consentData, storedConsentData, forceRefresh) {
+  // There are two submodule configuration types to handle: storage or value
+  // 1. storage: retrieve user id data from cookie/html storage or with the submodule's getId method
+  // 2. value: pass directly to bids
+  if (submodule.config.storage) {
+    let storedId = getStoredValue(submodule.config.storage);
+    let response;
+
+    let refreshNeeded = false;
+    if (typeof submodule.config.storage.refreshInSeconds === 'number') {
+      const storedDate = new Date(getStoredValue(submodule.config.storage, 'last'));
+      refreshNeeded = storedDate && (Date.now() - storedDate.getTime() > submodule.config.storage.refreshInSeconds * 1000);
+    }
+
+    if (!storedId || refreshNeeded || !!forceRefresh || !storedConsentDataMatchesConsentData(storedConsentData, consentData)) {
+      // No id previously saved, or a refresh is needed, or consent has changed. Request a new id from the submodule.
+      response = submodule.submodule.getId(submodule.config.params, consentData, storedId);
+    } else if (typeof submodule.submodule.extendId === 'function') {
+      // If the id exists already, give submodule a chance to decide additional actions that need to be taken
+      response = submodule.submodule.extendId(submodule.config.params, storedId);
+    }
+
+    if (utils.isPlainObject(response)) {
+      if (response.id) {
+        // A getId/extendId result assumed to be valid user id data, which should be saved to users local storage or cookies
+        setStoredValue(submodule, response.id);
+        storedId = response.id;
+      }
+
+      if (typeof response.callback === 'function') {
+        // Save async callback to be invoked after auction
+        submodule.callback = response.callback;
+      }
+    }
+
+    if (storedId) {
+      // cache decoded value (this is copied to every adUnit bid)
+      submodule.idObj = submodule.submodule.decode(storedId, submodule.config.params);
+    }
+  } else if (submodule.config.value) {
+    // cache decoded value (this is copied to every adUnit bid)
+    submodule.idObj = submodule.config.value;
+  } else {
+    const response = submodule.submodule.getId(submodule.config.params, consentData, undefined);
+    if (utils.isPlainObject(response)) {
+      if (typeof response.callback === 'function') { submodule.callback = response.callback; }
+      if (response.id) { submodule.idObj = submodule.submodule.decode(response.id, submodule.config.params); }
+    }
+  }
+}
 
 /**
  * @param {SubmoduleContainer[]} submodules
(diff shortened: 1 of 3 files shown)


smenzer:
this is a good start, but it doesn't take into account the fact that the user id modules will likely need to call their servers again to get a fresh ID. here's a use case:

1. user enters site example.com for the first time
2. userId module runs and each submodule calls their server to generate a fresh ID for the user
3. user logs in to the site with user@email.com
4. refreshIds() is called, but since the userId doesn't know it needs to refresh the submodules, it simply pulls the latest ID from localstorage and returns it via `extendId()` and then `decode()`

In fact, in step 4 the userId module actually needs to issue a `getId()` call instead.

TLadd:
To address @smenzer's points, that refresh flag would need to be passed down into `initSubmodules` and then have that force it go down the `getId` path regardless of if there is a non-expired stored value or not. I agree this is necessary since the refresh doesn't do a whole lot of good if just keeps the stored values around. 

I would also like if there were a way to only refresh specific submodules. In my particular case, the effect of being logged in only effects a handful of submodules, so it would be better for me to trigger just the ones I know will benefit from the new information instead of having to update them all. I think two reasonable ways to do this would be:

- a `refreshUserId` method that takes a submodule name 
- an option like `refreshUserIds({ submodules: string[] })`. 

I would probably prefer `refreshUserIds({ submodules: string[] })` since it keeps it to one global api method and also makes it easy to filter down to one or multiple submodules.

smenzer:
i like that approach and something i was thinking about too. if you specify a list of submodules, then it's clear which ones are getting refreshed and having that passed in as a condition to call `getId()` would be perfect.

coryhammon1:
Refactored for individual submodule refresh. I've done some initial testing and the submodule's getId method is called on refresh. I may still be missing something though (I'm not super familiar with this code). Please let me know if there's more I need to do.

modules/userId/index.js
```diff
@@ -660,6 +703,8 @@ export function init(config) {
   // exposing getUserIds function in global-name-space so that userIds stored in Prebid can be used by external codes.
   (getGlobal()).getUserIds = getUserIds;
   (getGlobal()).getUserIdsAsEids = getUserIdsAsEids;
+
+  (getGlobal()).refreshUserIds = refreshUserIds;
```
smenzer, line 707: no need for an extra line here IMO

```suggestion
  (getGlobal()).refreshUserIds = refreshUserIds;
```

smenzer (changes requested):
a few comments/questions, but overall I like this approach

modules/userId/index.js
```diff
@@ -465,13 +465,103 @@ function getUserIdsAsEids() {
...
+* This function will be exposed in the global-name-space so that userIds can be refreshed after initialization.
+* @param {String[]} submoduleNames submodules to be refreshed
+*/
+function refreshUserIds(submoduleNames) {
+  if (!utils.isArray(submoduleNames)) {
+    return;
+  }
+
+  if (!initializedSubmodules) {
+    utils.logWarn(`${MODULE_NAME} - refresh called before submodule initialization`);
```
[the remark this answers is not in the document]
TLadd: I would have expected this method to initialize the submodule if it has not already been initialized. Otherwise you might end up with race conditions  where you need to wait for a first bid to occur before you can call this method.

So I would expect this not to return early here. Instead of iterating over `initializedSubmodules` below, I would iterate over `userIdModules` and filter by the name.

modules/userId/index.js
```diff
@@ -490,54 +580,7 @@ function initSubmodules(submodules, consentData) {
...
-      submodule.idObj = submodule.config.value;
-    } else {
-      const response = submodule.submodule.getId(submodule.config.params, consentData, undefined);
-      if (utils.isPlainObject(response)) {
-        if (typeof response.callback === 'function') { submodule.callback = response.callback; }
-        if (response.id) { submodule.idObj = submodule.submodule.decode(response.id, submodule.config.params); }
-      }
-    }
+    populateSubmoduleId(submodule, consentData, storedConsentData, false);
     carry.push(submodule);
```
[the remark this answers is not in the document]
coryhammon1: populateSubmoduleId updates the submodule object with the id. It's a "side effect" in functional programming parlance.

coryhammon1:
I added submodule initialization to the refresh function, so that it won't quit early if they aren't initialized. I think this will work correctly based on both your comments above.

smenzer:
this is looking good. when you add tests, can you make sure that in addition to just calling `getId` for each refreshed module, that if a submodule's config value changes by the publisher, that the submodule also gets the changed config values (and getId isn't using a cached value for some reason)?

modules/userId/index.js
```diff
@@ -465,13 +465,109 @@ function getUserIdsAsEids() {
   return createEidsArray(getCombinedSubmoduleIds(initializedSubmodules));
 }
 
+/**
+* This function will be exposed in the global-name-space so that userIds can be refreshed after initialization.
+* @param {String[]} submoduleNames submodules to be refreshed
+*/
+function refreshUserIds(submoduleNames) {
```
TLadd, line 472: Would we want `submoduleNames` to be optional, with it defaulting to refreshing all submodules? I brought up passing in the submodule names because my particular use case only requires refreshing one or two userId modules, but refreshing all might still be a useful default? If we do end up making `submoduleNames` optional, I would suggest making the argument to `refreshUserIds` an object so it would be easy to pass additional arguments down the line in a non-breaking way.

Note this is mostly a question. It may be the case we want to keep it the way it is with submoduleNames being required.

smenzer: not a bad idea to support refreshing all ids. pubs may not know which of their ids need a refresh per se, so being able to just refresh them all seems like an easy way to do it for them.

maybe in the call to `getId` it should be flagged if a force refresh is happening in case the module wants to act differently?

coryhammon1:
Made the submodule name list parameter optional, and if absent all modules will be refreshed. Also, added a unit test, and verified that the most recently updated config will be read when refreshing.

modules/userId/index.js
```diff
@@ -465,13 +470,112 @@ function getUserIdsAsEids() {
...
+
+    let refreshNeeded = false;
+    if (typeof submodule.config.storage.refreshInSeconds === 'number') {
+      const storedDate = new Date(getStoredValue(submodule.config.storage, 'last'));
+      refreshNeeded = storedDate && (Date.now() - storedDate.getTime() > submodule.config.storage.refreshInSeconds * 1000);
+    }
+
+    if (!storedId || refreshNeeded || !!forceRefresh || !storedConsentDataMatchesConsentData(storedConsentData, consentData)) {
+      // No id previously saved, or a refresh is needed, or consent has changed. Request a new id from the submodule.
+      response = submodule.submodule.getId(submodule.config.params, consentData, storedId);
```
smenzer, line 544: what do folks think about passing `forceRefresh` into this function so if a submodule wants to treat a refresh differently for some reason, they can?

smenzer (commented):
tests feel a bit light, can we add some about which submodules get refreshed based on what you pass in, etc.?

modules/userId/index.js
smenzer: @icflournoy do you have thoughts on this?

smenzer:
> tests feel a bit light, can we add some about which submodules get refreshed based on what you pass in, etc.?

@TLadd could you add a test (or edit one of the existing ones) to cover the `refreshUserIds()` callback, please? once that's done I'll merge.  thanks!

smenzer:
@TLadd @coryhammon1  The CI build is failing in Edge with this message ... can you look into it and see if you can find anything that broke the Zeotap adapter and attempt to fix, too, please?

```
1) when a stored Zeotap ID exists it is added to bids
     Zeotap ID System requestBids hook
     expected { Object (bidder, params) } to have deep nested property 'userId.IDP'
Error
   at AssertionError (webpack:///node_modules/assertion-error/index.js:74:1 <- test/test_index.js:24167:7)
   at Assertion.prototype.assert (webpack:///node_modules/chai/lib/chai/assertion.js:141:1 <- test/test_index.js:53648:7)
   at assertProperty (webpack:///node_modules/chai/lib/chai/core/assertions.js:1866:1 <- test/test_index.js:55544:7)
   at methodWrapper (webpack:///node_modules/chai/lib/chai/utils/addMethod.js:57:1 <- test/test_index.js:52778:5)
   at Anonymous function (webpack:///test/spec/modules/zeotapIdPlusIdSystem_spec.js:125:13 <- test/test_index.js:374855:13)
   at Array.prototype.forEach (native code)
   at Anonymous function (webpack:///test/spec/modules/zeotapIdPlusIdSystem_spec.js:124:11 <- test/test_index.js:374854:11)
   at Array.prototype.forEach (native code)
   at Anonymous function (webpack:///test/spec/modules/zeotapIdPlusIdSystem_spec.js:123:9 <- test/test_index.js:374853:9)
   at Anonymous function (webpack:///modules/userId/index.js:449:5 <- test/test_index.js:15160:5)
```

TLadd:
> > tests feel a bit light, can we add some about which submodules get refreshed based on what you pass in, etc.?
> 
> @TLadd could you add a test (or edit one of the existing ones) to cover the `refreshUserIds()` callback, please? once that's done I'll merge. thanks!

Added that here: https://github.com/prebid/Prebid.js/pull/5815

modules/userId/index.js
smenzer: i've created a new issue to track interest in this: https://github.com/prebid/Prebid.js/issues/5818

smenzer:
it looks like you've been working on an outdated version of master that has an old version of the zeotap test. if you look at the version in this PR compared to what is in master (https://github.com/prebid/Prebid.js/blob/master/test/spec/modules/zeotapIdPlusIdSystem_spec.js) you'll see that several of these cookie calls have been removed per https://github.com/prebid/Prebid.js/pull/5758. Make sure you do a full merge of master and resolve the conflicts, then this should be working properly. This may have been due to https://github.com/prebid/Prebid.js/pull/5815 I'm not sure

TLadd:
@smenzer Yep, sure enough, the same failures were happening on master before that PR that mocked writing to cookies: https://app.circleci.com/pipelines/github/prebid/Prebid.js/2130/workflows/efbe4e02-8090-41ea-a517-adef33a6410a/jobs/9900.

I've created https://github.com/prebid/Prebid.js/pull/5819 that merged the latest master and removes the mocking, as it is no longer necessary after removing the expiration time on the cookie. I'm generally of the philosophy that mocking should be avoided unless they make tests easier to write/maintain and don't diminish the value of the test significantly. In this case, the tests aren't easier to write and assuming cookies work in the tests, I think that does provide some value to the tests. There are a lot things that can go wrong with cookies. Trying to write a unit test suite for cookie storage that covers every possible thing that could go wrong within Prebid would be really hard. Even if you did have such a suite, mocks can be set incorrectly and behave differently than reality. Part of the reason to bother running these tests in multiple browsers I assume is to catch browser-specific behavior, which you don't get when mocking browser API's.

Just my 2 cents on the matter; as I said in #5819, I can easily go back to the mocking solution if that's preferred.

smenzer:
closing in favor of https://github.com/prebid/Prebid.js/pull/5819

PR #5756, opened 2020-09-16
Convert id5id to an object to support passing additional data points to platforms

smenzer:
## Type of change
- [ x ] Feature
- [ x ] Refactoring (no functional changes, no api changes)

## Description of change
Updating the response that the ID5 ID modules provides to bid adapters from a simple string to an object. The purpose of this is to include other pieces of information that may be useful to bid adapters/SSPs and downstream DSPs as well. Extra values are placed in an `ext` in both the `userId.id5id` object and in the `userIdsAsEids` array. Currently, we are passing a `linkType` value in the extension, which represents the type of link we were able to make for the user across domains, indicating the strength/accuracy/value of the ID. In the future, we plan to include additional signals/data points as well.

In order to support this change, all bidder adapters currently listening to the ID5 ID were also modified. It is not required to listen to the `ext`, so in most cases the adapters (and / or their tests) were just updated to read the id string from the object to keep things simple. There should be no functional changes to any bidder adapter, except for those who already supported listening to extensions, they will now get the ID5 ID extension passed in.

### DIFF
--- a/test/spec/modules/vidazooBidAdapter_spec.js
+++ b/test/spec/modules/vidazooBidAdapter_spec.js
@@ -79,7 +79,7 @@ const REQUEST = {
   }
 };
 
-describe('VidazooBidAdapter', function () {
+describe.only('VidazooBidAdapter', function () {
   describe('validtae spec', function () {
     it('exists and is a function', function () {
       expect(adapter.isBidRequestValid).to.exist.and.to.be.a('function');
@@ -244,6 +244,7 @@ describe('VidazooBidAdapter', function () {
           case 'digitrustid': return { data: { id: id } };
           case 'lipb': return { lipbid: id };
           case 'parrableId': return { eid: id };
+          case 'id5id': return { uid: id };
           default: return id;
         }
       })();
--- a/test/spec/modules/vidazooBidAdapter_spec.js
+++ b/test/spec/modules/vidazooBidAdapter_spec.js
@@ -79,7 +79,7 @@ const REQUEST = {
   }
 };
 
-describe.only('VidazooBidAdapter', function () {
+describe('VidazooBidAdapter', function () {
   describe('validtae spec', function () {
     it('exists and is a function', function () {
       expect(adapter.isBidRequestValid).to.exist.and.to.be.a('function');
--- a/modules/adheseBidAdapter.js
+++ b/modules/adheseBidAdapter.js
@@ -150,8 +150,8 @@ function getAccount(validBidRequests) {
 }
 
 function getId5Id(validBidRequests) {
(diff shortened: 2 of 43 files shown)


smenzer:
Tagging folks from each of the bid adapters that were modified as an FYI:
@mefjush @adxcgcom @scritchley @HuddledMasses @stevealliance @sa1omon @TheMediaGrid @bjorn-lw @jimee02 @afsheenb @anand-venkatraman @richaudience @npeceniak @uditalias @mk0x9 @nickjacob @nwlosinski @matthieularere-msq @pm-harshad-mane @lowendavid @omerko @shrikantpatwari

Fawke (changes requested):
Hi @smenzer.

The code looks good to me but I've two concerns:

1. In the docs, specifically over [here](https://docs.prebid.org/dev-docs/modules/userId.html#id5-universal-id-examples), the part where the publisher has integrated ID5 on their own and wants to pass the ID5 Universal ID directly through to Prebid.js, the interface still stays:

```(javascript)
pbjs.setConfig({
  userSync: {
    userIds: [{
      name: "id5Id",
      value: { "id5id": "ID5-8ekgswyBTQqnkEKy0ErmeQ1GN5wV4pSmA-RE4eRedA" }
    }]
  }
});
```
Don't you think, now that we've changed the way the adapters consume the value of ID5 id, there'd be a problem if pub goes the second route describe above?

2. I think mistakenly, you've include `.only` is one of the test files resulting in only 34 unit tests to run in total.

test/spec/modules/vidazooBidAdapter_spec.js
```diff
@@ -79,7 +79,7 @@ const REQUEST = {
   }
 };
 
-describe('VidazooBidAdapter', function () {
+describe.only('VidazooBidAdapter', function () {
```
Fawke, line 82: This is running only this test file. I think you forgot to remove `.only`?

smenzer: yep! they had been giving me some errors so i was testing locally just them, forgot to remove before committing...will fix that now

smenzer:
@Fawke 

1. good catch on the docs, I need to update that as well...will post a PR here when I do. 

2. oops! will fix asap

smenzer:
here's the docs PR I forgot to post: https://github.com/prebid/prebid.github.io/pull/2348

Fawke (approved):
LGTM

smenzer:
@idettman let me know if you have any questions/comments on this before it can be merged. thanks!

smenzer:
@jlukas79 could you take a look at this, please, @idettman doesn't seem to be available

mkendall07 (approved):
LGTM

Issue #5670, opened 2020-08-27
Intent to Implement: Allow selection of supported default targeting keys at configuration time.


mimenet:
## Type of issue

Intent to implement

## Description

Prebid default targeting includes keys that are not be needed by all publishers. The inclusion of these keys and their corresponding values can lead to /gampad/ads requests being truncated. Publishers in this situation would benefit from an option that allows them to select which default keys they would like to employ.
The proposal is to allow publishers to select the default keys they would like to support, without changing how those keys are generated or otherwise managed.

## Requirements

1. Default targeting keys can be suppressed, custom targeting keys cannot.
2. AdserverTargeting objects are not changed.
3. Targeting return by targeting.getAllTargeting is affected without side effects.
4. Keys are identified by their key name in constants.json.
5. Targeting is removed when a key contains a full substring match of a non-supported default key at position 0.
6. This should be settable with pbjs.setConfig.

## Proposed External Interface

Supporting only hb_bidder, hb_adid, and hb_pb would look like this:
```
pbjs.setConfig({
  "targetingControls": {
    "allowTargetingKeys": ["BIDDER", "AD_ID", "PRICE_BUCKET"]
  }
});
```

## Other information

This proposal is based on conversations in prebid.js PMC 
Prebid.js PR found here: https://github.com/prebid/Prebid.js/pull/5763

PR #5763, opened 2020-09-16
Allow selection of supported default targeting keys at configuration time.

mimenet:
## Type of change

- [x] Feature

## Description of change

Prebid default targeting includes keys that are not be needed by all publishers. The inclusion of these keys and their corresponding values can lead to /gampad/ads requests being truncated. Publishers in this situation would benefit from an option that allows them to select which default keys they would like to employ.

The change allows publishers to select the default keys they would like to support, without changing how those keys are generated or subsequently managed.

Prebid.github.io PR here:
https://github.com/prebid/prebid.github.io/pull/2346

## Other information

Based on I2I here: https://github.com/prebid/Prebid.js/issues/5670

### DIFF
--- a/src/targeting.js
+++ b/src/targeting.js
@@ -182,6 +182,48 @@ export function newTargeting(auctionManager) {
   };
 
   /**
+   * Returns filtered ad server targeting for custom and allowed keys.
+   * @param {targetingArray} targeting
+   * @param {string[]} allowedKeys
+   * @return {targetingArray} filtered targeting
+   */
+  function getAllowedTargetingKeyValues(targeting, allowedKeys) {
+    logInfo(`Allowed default targeting keys [ ${allowedKeys.join(', ')}]`);
+    const defaultKeyring = Object.assign({}, CONSTANTS.TARGETING_KEYS, CONSTANTS.NATIVE_KEYS);
+    const defaultKeys = Object.keys(defaultKeyring);
+    targeting.map(adUnit => {
+      const adUnitCode = Object.keys(adUnit)[0];
+      const keyring = adUnit[adUnitCode];
+      const keys = keyring.filter(kvPair => {
+        const key = Object.keys(kvPair)[0];
+        // check if key is in default keys, if not, it's custom, we won't remove it.
+        const isCustom = defaultKeys.filter(defaultKey => key.indexOf(defaultKeyring[defaultKey]) === 0).length === 0;
+        // check if key explicitly allowed, if not, we'll remove it.
+        const found = isCustom || allowedKeys.find(allowedKey => {
+          const allowedKeyName = defaultKeyring[allowedKey];
+          // we're looking to see if the key exactly starts with one of our default keys.
+          // (which hopefully means it's not custom)
+          const found = key.indexOf(allowedKeyName) === 0;
+          return found;
+        });
+        const keyDisposition = found ? 'Keeping' : 'Removing';
+        const keyType = isCustom ? 'custom' : 'default';
+        logInfo(`${keyDisposition} ${keyType} targeting key ${key} for ${adUnitCode} adserverTargeting object`);
+        return found;
+      });
+      adUnit[adUnitCode] = keys;
+    });
+    const filteredTargeting = targeting.filter(adUnit => {
+      const adUnitCode = Object.keys(adUnit)[0];
+      const keyring = adUnit[adUnitCode];
+      const targetingObjectDisposition = keyring.length > 0 ? 'Keeping populated' : 'Removing empty';
+      logInfo(`${targetingObjectDisposition} ${adUnitCode} adserverTargeting object`);
+      return keyring.length > 0;
+    });
+    return filteredTargeting
+  }
+
+  /**
    * Returns all ad server targeting for all ad units.
    * @param {string=} adUnitCode
    * @return {Object.<string,targeting>} targeting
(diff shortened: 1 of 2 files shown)


src/targeting.js
```diff
@@ -181,6 +181,48 @@ export function newTargeting(auctionManager) {
...
+        const found = isCustom || allowedKeys.find(allowedKey => {
+          const allowedKeyName = defaultKeyring[allowedKey];
+          // we're looking to see if the key exactly starts with one of our default keys.
+          // (which hopefully means it's not custom)
+          const found = key.indexOf(allowedKeyName) === 0;
+          return found;
+        });
+        const keyDisposition = found ? 'Keeping' : 'Removing';
+        const keyType = isCustom ? 'custom' : 'default';
+        logInfo(`${keyDisposition} ${keyType} targeting key ${key} for ${adUnitCode} adserverTargeting object`);
```
r-schweitzer, line 211: Would it be possible to instead keep arrays of all the keys that we are removing and keeping and have one log for each that says these keys we've removed and these we've kept? This just seems like a lot of logs if we do it for each individual key. 

Or it might be better to have just the array of keys that will be allowed or one log of what we have removed.

src/targeting.js
```diff
@@ -181,6 +181,48 @@ export function newTargeting(auctionManager) {
...
+        return found;
+      });
+      adUnit[adUnitCode] = keys;
+    });
+    const filteredTargeting = targeting.filter(adUnit => {
+      const adUnitCode = Object.keys(adUnit)[0];
+      const keyring = adUnit[adUnitCode];
+      const targetingObjectDisposition = keyring.length > 0 ? 'Keeping populated' : 'Removing empty';
+      logInfo(`${targetingObjectDisposition} ${adUnitCode} adserverTargeting object`);
```
r-schweitzer, line 220: After running a test I'm not sure if we'd need this log? Or maybe it could possibly provide more info (like what is it keeping). When I ran your code I would see multiple logs either saying "keeping populated" or "removing empty" for the same adunit but it doesn't say much else and we're already logging all the keys removed and kept. Maybe I'm misunderstanding the needs for this as well.

test/spec/unit/core/targeting_spec.js
```diff
@@ -416,6 +416,46 @@ describe('targeting tests', function () {
       });
     });
 
+    describe('targetingControls.allowTargetingKeys', function () {
```
GLStephen, line 419: If we are going to use "begins with" matching then embody that in a test as a form of explicit documentation, this will trigger a test failure if someone tries to "improve" the code and change the behavior among other advantages like documenting the reality in the test:

it('should allow hn_pb and hb_pb_anything in any variation')

src/targeting.js
mimenet: Makes sense, I'll collate into one log. Thanks!

src/targeting.js
mimenet: This log was intended to illustrate that targeting for the specified adUnit was completely removed and thusly the adUnit was removed from targeting entirely.

Maybe removing the adUnit for lack of targeting is the wrong move in the first place. I'll check the behavior in the case where all targeting is eliminated, but the adUnit isn't removed from the targeting object.

src/targeting.js
```diff
@@ -181,6 +181,48 @@ export function newTargeting(auctionManager) {
...
+      const keyring = adUnit[adUnitCode];
+      const keys = keyring.filter(kvPair => {
+        const key = Object.keys(kvPair)[0];
+        // check if key is in default keys, if not, it's custom, we won't remove it.
+        const isCustom = defaultKeys.filter(defaultKey => key.indexOf(defaultKeyring[defaultKey]) === 0).length === 0;
+        // check if key explicitly allowed, if not, we'll remove it.
+        const found = isCustom || allowedKeys.find(allowedKey => {
+          const allowedKeyName = defaultKeyring[allowedKey];
+          // we're looking to see if the key exactly starts with one of our default keys.
+          // (which hopefully means it's not custom)
```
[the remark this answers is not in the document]
mimenet: You are correct. This change is only concerned with removing default targeting keys. If a publisher would like to stop generating or sending a custom key that they have defined, that is within their purview.

The reason that I opted against exact matching is because bid landscape concatenates the default key to the bidder, meaning "hb_pb" becomes "hb_pb_mybidder". I didn't want to pre-generate all possible keys and filter on all possible keys.

The best argument for exact matching to me, was that a publisher filtering out a default key would filter out custom keys with a matching prefix. to use your example if a publisher filtered out the default key "hb_pb" then their custom key "hb_pb_classification" would also be filtered out. This argument is why I included verbose logging, allowing the publisher to see that they can't filter "hb_pb" without also filtering "hb_pb_classification"

test/spec/unit/core/targeting_spec.js
mimenet: Makes sense. The behavior is being tested, but it's not documented clearly. Thanks!

src/targeting.js
mimenet: After testing the behavior again, I see what you mean, it's neither descriptive nor necessary to log at this step.

src/targeting.js
mimenet: This has been updated.

src/targeting.js
mimenet: This has been updated.

test/spec/unit/core/targeting_spec.js
mimenet: This has been updated.

r-schweitzer (approved):

GLStephen (approved):

PR #5766, opened 2020-09-17
added instream video ad support

codewiz14:
## Type of change

- [ ] Bugfix
- [x] Feature
- [ ] New bidder adapter   
- [ ] Build related changes
- [ ] CI related changes
- [ ] Other

## Description of change

- test parameters for validating instream video bids
```
{
  bidder: 'gjirafa',
  params: {
    propertyId: '105227',
    placementId: '846836'
  }
},
{
  bidder: 'malltv',
  params: {
    propertyId: '105134',
    placementId: '846841'
  }
}
```

- contact: drilon@gjirafa.com
- Docs PR prebid/prebid.github.io#2200


Fawke (commented):
@codewiz14 Thanks for the PR. I've tested instream video with your adapters and it seems to be working fine. I just noticed that both of your adapter, `gjirafa` and `malltv` share a lot of code, but then I noticed the endpoints are different. Do you manage both the bidders?

Another thing that I wanted to point was that your `ad` property in the bid request is coming as `undefined` for both the bidders. You may wanna check that.

Screenshot 2020-09-23 at 4 02 03 PM

Also, had one concern over a global variable that you've used over [here](https://github.com/prebid/Prebid.js/blob/62ade0e4c1c2ca39c7efa4cda0e37d20f3bc76fe/modules/gjirafaBidAdapter.js#L61). If there are concurrent auctions, it may create a problem, or, since, you have this global variable shared in both `gjirafa` and `malltv`, what happens if someones tries to load both the adapter. Won't there be a conflict?

codewiz14:
Hi @Fawke, thanks for the feedback.
Yes, we manage both adapters, everything is same in both adapters except the endpoint. If you have any suggestion how we can merge them in single adapter, please let me know.

The `ad` property is not necessary in video ads, but if needed we can make sure that it will be `null` or empty string.

As for the global variable it seems like it is forgotten there and will be removed on our next commit. 
We are implementing an optimization to request the ads as bulk to lower the number of requests, with that we will remove the global variable.

Fawke:
@codewiz14,

If you have two bidders with different endpoints, we have no other option but to keep them like this. Had it been same endpoint, you could have added one of them as an alias of the other one. 

It's fine to leave the `ad` property undefined, it would've been better if we don't need that property then, not including it as part of bid response object in the first place.

Rest, looks good.

Fawke (approved):
LGTM

Issue #5690, opened 2020-09-02
Prebid Publisher-Provided User IDs


bretg:
## Type of issue
Proposed Convention

## Description

**Note:** major update to the proposal here -- replacing the original First Party Data-based proposal with a new UserID sub-module.

Recent identity discussions have suggested there would be value in allowing Publishers to communicate their own first party user IDs (PPUID = Publisher Provided User ID)  to some or all bidders.

The precise use cases aren't fully worked out, but this proposal takes on a meta-goal of enabling publisher flexibility:

- ability to choose which bidders receive their ID(s)
- ability to support multiple IDs (e.g. multiple data sets, transition phases)

If we've learned anything about User IDs in the past year, it's that the space changes quickly, so flexibility seems warranted.

Add another user ID sub-module, call it the "PubProvidedId" module. It takes values from two places:

static eids values e.g.
```
pbjs.setConfig({
    userSync: {
        userIds: [{
            name: "pubProvidedId",
            params: {
                eids: [{
                    source: "example.com",
                    uids:[{
                      id: "value read from cookie or local storage"
                      ext: {
                        stype: "ppuid"
                      }
                  }]
                },{
                    source: "id-partner.com",
                    uids:[{
                      id: "value read from cookie or local storage"
                      ext: {
                        stype: "dmp"
                      }
                  }]
                }]
            }
        }]
    }
});
```
a publisher-provided function that produces eids-style output
```
pbjs.setConfig({
    userSync: {
        userIds: [{
            name: "publisherProvided",
            params: {
                eidsFunction: getMyId  // any function that exists in the page
            }
        }]
    }
});
```
After consent is checked, the resulting eids structure is simply passed to bid adapters as bidRequest.userId.pubProvided.
```
{
                eids: [{
                    source: "example.com",
                    uids:[{
                      id: "value read from cookie or local storage"
                      ext: {
                        stype: "ppuid"
                      }
                  }]
                },{
                    source: "custom-id-partner.com",
                    uids:[{
                      id: "value read from cookie or local storage"
                  }]
                }]
}
```
It's also available from the getUserIdsAsEids() function.

## AMP

It's come to our attention that passing dynamic data through AMP is a problem. See prebid/prebid-server#1404

So even though we don't know that it's currently possible to handle User IDs in AMP at all, I think it makes sense to change the proposed (theoretical) AMP proposal to removing User IDs from the targeting and making a new path for it:
```
  <amp-ad width="300" height="50"
[...]

PR #5767, opened 2020-09-17
New PubProvided Id UserId Submodule

YerkovichM:
## Type of change

- [ ] Bugfix
- [x] Feature
- [ ] New bidder adapter   
- [ ] Build related changes
- [ ] CI related changes
- [ ] Other

## Description of change

New Publisher Provided Id UserId submodule

Issue: https://github.com/prebid/Prebid.js/issues/5690#issuecomment-693597321

Ex:

Config:

```
        const eidsProvider = function(){
            return [{
                source: "prov.com",
                uids:[{
                    id: "value read from cookie or local storage",
                    ext: {
                      stype: "sha256email"
                    }
                }]
            }];
        };
        pbjs.setConfig({
            userSync: {
                userIds: [{
                    name: "publisherProvided",
                    params: {
                        eidsFunction: eidsProvider,   // either a function OR static eids
                        eids: [{
                            source: "example.com",
                            uids:[{
                                id: "value read from cookie or local storage",
                                ext: {
                                    stype: "ppuid"
                                }
                            }]
                        },{
                            source: "id-partner.com",
                            uids:[{
                                id: "value read from cookie or local storage"
                                ext: {
                                    stype: "dmp"
                                }
                            }]
                        }]
                    },
                },{
                    name: "sharedId",
                    params: {
                        syncTime: 60 // in seconds, default is 24 hours
                    },
                    storage: {
                        type: "cookie",
                        name: "sharedid",
                        expires: 28
                    }
                }]
            }
        });
```

Result: 
bid.userId:
```
{
  "pubProvided": [
    {
      "source": "example.com",
      "uids": [
        {
          "id": "value read from cookie or local storage",
          "ext": {
              "stype": "ppuid"
          }
        }
      ]
    },
    {
      "source": "id-partner.com",
      "uids": [
        {
          "id": "value read from cookie or local storage",
           "ext": {
              "stype": "dmp"
          }
        }
      ]
    },
    {
      "source": "prov.com",
      "uids": [
        {
          "id": "value read from cookie or local storage",
          "ext": {
              "stype": "sha256email"
          }
        }
      ]
    }
  ],
  "sharedid": {
    "id": "01EFXYX2RY99R5N03B94AETQ19",
    "third": "01EFXYX2RY99R5N03B94AETQ19"
  }
}
```
bid.userIdAsEids:
```
[
 {
      "source": "example.com",
      "uids": [
        {
          "id": "value read from cookie or local storage",
          "ext": {
               "stype": "ppuid"
          }
        }
      ]
    },
    {
      "source": "id-partner.com",
      "uids": [
        {
          "id": "value read from cookie or local storage",
           "ext": {
              "stype": "dmp"
          }
        }
      ]
    },
    {
      "source": "prov.com",
      "uids": [
        {
          "id": "value read from cookie or local storage",
          "ext": {
               "stype": "sha256email"
          }
        }
      ]
    },
  {
    "source": "sharedid.org",
    "uids": [
      {
        "id": "01EFXYX2RY99R5N03B94AETQ19",
        "atype": 1,
        "ext": {
          "third": "01EFXYX2RY99R5N03B94AETQ19"
        }
      }
    ]
  }
]
```

@jdwieland8282

### DIFF
--- a/integrationExamples/gpt/userId_example.html
+++ b/integrationExamples/gpt/userId_example.html
@@ -109,6 +109,15 @@
             googletag.pubads().disableInitialLoad();
         });
 
+        const eidsProvider = function(){
+            return [{
+                source: "prov.com",
+                uids:[{
+                    id: "value read from cookie or local storage"
+                }]
+            }];
+        };
+
         pbjs.que.push(function() {
             pbjs.setConfig({
                 debug: true,
@@ -130,6 +139,25 @@
                 // },
                 userSync: {
                     userIds: [{
+                        name: "publisherProvided",
+                        params: {
+                            eids: [{
+                                source: "example.com",
+                                uids:[{
+                                    id: "value read from cookie or local storage",
+                                    ext: {
+                                        third: "third"
+                                    }
+                                }]
+                            },{
+                                source: "id-partner.com",
+                                uids:[{
+                                    id: "value read from cookie or local storage"
+                                }]
+                            }],
+                            eidsFunction: eidsProvider
+                        }
+                    },{
                         name: "unifiedId",
                         params: {
                             partner: "prebid",
--- a/integrationExamples/gpt/userId_example.html
+++ b/integrationExamples/gpt/userId_example.html
@@ -109,15 +109,6 @@
             googletag.pubads().disableInitialLoad();
         });
 
-        const eidsProvider = function(){
-            return [{
-                source: "provider.com",
-                uids:[{
-                    id: "value read from cookie or local storage"
-                }]
-            }];
-        };
-
         pbjs.que.push(function() {
             pbjs.setConfig({
                 debug: true,
@@ -139,7 +130,7 @@
                 // },
                 userSync: {
                     userIds: [{
-                        name: "publisherProvided",
+                        name: "pubProvided",
                         params: {
                             eids: [{
                                 source: "example.com",
@@ -154,7 +145,14 @@
                                     id: "value read from cookie or local storage"
                                 }]
                             }],
-                            eidsFunction: eidsProvider
+                            eidsFunction: function() {
+                                return [{
+                                    source: "provider.com",
+                                    uids: [{
+                                        id: "value read from cookie or local storage"
+                                    }]
+                                }]
+                            }
                         }
                     },{
                         name: "unifiedId",
--- a/integrationExamples/gpt/userId_example.html
+++ b/integrationExamples/gpt/userId_example.html
@@ -134,26 +134,33 @@
                 // },
                 userSync: {
                     userIds: [{
-                        name: "pubProvided",
+                        name: "pubProvidedId",
                         params: {
                             eids: [{
                                 source: "example.com",
                                 uids:[{
                                     id: "value read from cookie or local storage",
                                     ext: {
+                                        type: "ppuid" // options are ppuid, dmp and sha256email
                                     }
                                 }]
                             },{
                                 source: "id-partner.com",
                                 uids:[{
-                                    id: "value read from cookie or local storage"
+                                    id: "value read from cookie or local storage",
+                                    ext: {
+                                        type: "dmp"
+                                    }
                                 }]
                             }],
                             eidsFunction: function() {
                                 return [{
                                     source: "provider.com",
                                     uids: [{
-                                        id: "value read from cookie or local storage"
+                                        id: "value read from cookie or local storage",
+                                        ext: {
+                                            type: "sha256email"
+                                        }
                                     }]
                                 }]
                             }
--- a/integrationExamples/gpt/userId_example.html
+++ b/integrationExamples/gpt/userId_example.html
@@ -111,7 +111,7 @@
 
         const eidsProvider = function(){
             return [{
-                source: "prov.com",
+                source: "provider.com",
                 uids:[{
                     id: "value read from cookie or local storage"
                 }]
@@ -146,7 +146,6 @@
                                 uids:[{
                                     id: "value read from cookie or local storage",
                                     ext: {
-                                        third: "third"
                                     }
                                 }]
                             },{
--- a/integrationExamples/gpt/userId_example.html
+++ b/integrationExamples/gpt/userId_example.html
@@ -137,33 +137,25 @@
                         name: "pubProvidedId",
                         params: {
                             eids: [{
-                                source: "example.com",
+                                source: "domain.com",
                                 uids:[{
                                     id: "value read from cookie or local storage",
+                                    atype: 1,
                                     ext: {
-                                        type: "ppuid" // options are ppuid, dmp and sha256email
+                                        stype: "ppuid" // allowable options are sha256email, DMP, ppuid for now
                                     }
                                 }]
                             },{
-                                source: "id-partner.com",
+                                source: "3rdpartyprovided.com",
                                 uids:[{
                                     id: "value read from cookie or local storage",
+                                    atype: 3,
                                     ext: {
-                                        type: "dmp"
+                                        stype: "sha256email"
                                     }
                                 }]
                             }],
-                            eidsFunction: function() {
-                                return [{
-                                    source: "provider.com",
-                                    uids: [{
-                                        id: "value read from cookie or local storage",
-                                        ext: {
-                                            type: "sha256email"
-                                        }
-                                    }]
-                                }]
-                            }
+                            eidsFunction: getHashedEmail  // any user defined function that exists in the page
                         }
                     },{
                         name: "unifiedId",
--- a/modules/pubProvidedSystem.js
+++ b/modules/pubProvidedSystem.js
@@ -12,6 +12,7 @@ const MODULE_NAME = 'pubProvidedId';
 
 /** @type {Submodule} */
 export const pubProvidedIdSubmodule = {
+
   /**
    * used to link submodule with config
    * @type {string}
--- a/modules/pubProvidedSystem.js
+++ b/modules/pubProvidedSystem.js
@@ -1,0 +1,52 @@
+/**
+ * This module adds Publisher Provided ids support to the User ID module
+ * The {@link module:modules/userId} module is required.
+ * @module modules/pubProvidedSystem
+ * @requires module:modules/userId
+ */
+
+import {submodule} from '../src/hook.js';
+import * as utils from "../src/utils";
+
+const MODULE_NAME = 'publisherProvided';
+
+/** @type {Submodule} */
+export const pubProvidedSubmodule = {
+  /**
+   * used to link submodule with config
+   * @type {string}
+   */
+  name: MODULE_NAME,
+
+  /**
+   * decode the stored id value for passing to bid request
+   * @function
+   * @param {string} value
+   * @returns {{pubProvided: array}} or undefined if value doesn't exists
+   */
+  decode(value) {
+    const res = value ? {pubProvided: value} : undefined;
+    utils.logInfo('PubProvidedId: Decoded value ' + JSON.stringify(res));
+    return res;
+  },
+
+  /**
+   * performs action to obtain id and return a value.
+   * @function
+   * @param {SubmoduleParams} [configParams]
+   * @returns {{id: array}}
+   */
+  getId(configParams) {
+    let res = [];
+    if (Array.isArray(configParams.eids)) {
+      res = res.concat(configParams.eids);
+    }
+    if (typeof configParams.eidsFunction === 'function') {
+      res = res.concat(configParams.eidsFunction());
+    }
+    return {id: res};
+  }
+};
+
+// Register submodule for userId
+submodule('userId', pubProvidedSubmodule);
--- a/modules/pubProvidedSystem.js
+++ b/modules/pubProvidedSystem.js
@@ -22,11 +22,11 @@ export const pubProvidedIdSubmodule = {
    * decode the stored id value for passing to bid request
    * @function
    * @param {string} value
-   * @returns {{pubProvided: array}} or undefined if value doesn't exists
+   * @returns {{pubProvidedId: array}} or undefined if value doesn't exists
    */
   decode(value) {
-    const res = value ? {pubProvided: value} : undefined;
-    utils.logInfo('PubProvided: Decoded value ' + JSON.stringify(res));
+    const res = value ? {pubProvidedId: value} : undefined;
+    utils.logInfo('PubProvidedId: Decoded value ' + JSON.stringify(res));
     return res;
   },
 
--- a/modules/pubProvidedSystem.js
+++ b/modules/pubProvidedSystem.js
@@ -8,7 +8,7 @@
 import {submodule} from '../src/hook.js';
 import * as utils from "../src/utils";
 
-const MODULE_NAME = 'publisherProvided';
+const MODULE_NAME = 'pubProvided';
 
 /** @type {Submodule} */
 export const pubProvidedSubmodule = {
@@ -26,7 +26,7 @@ export const pubProvidedSubmodule = {
    */
   decode(value) {
     const res = value ? {pubProvided: value} : undefined;
-    utils.logInfo('PubProvidedId: Decoded value ' + JSON.stringify(res));
+    utils.logInfo('PubProvided: Decoded value ' + JSON.stringify(res));
     return res;
   },
 
--- a/modules/pubProvidedSystem.js
+++ b/modules/pubProvidedSystem.js
@@ -10,6 +10,15 @@ import * as utils from '../src/utils.js';
 
 const MODULE_NAME = 'pubProvided';
 
+function addType(uid) {
+  if(!uid.ext.types){
+    uid.ext.types = [];
+  }
+  if(!uid.ext.types.includes('ppuid')){
+    uid.ext.types.push('ppuid')
+  }
+}
+
 /** @type {Submodule} */
 export const pubProvidedSubmodule = {
   /**
@@ -44,7 +53,7 @@ export const pubProvidedSubmodule = {
     if (typeof configParams.eidsFunction === 'function') {
       res = res.concat(configParams.eidsFunction());
     }
-    res.forEach(id => id.type = 'pubProvided');
+    res.forEach(id => id.uids.forEach(uid => addType(uid)));
     return {id: res};
   }
 };
--- a/modules/pubProvidedSystem.js
+++ b/modules/pubProvidedSystem.js
@@ -11,10 +11,13 @@ import * as utils from '../src/utils.js';
 const MODULE_NAME = 'pubProvided';
 
 function addType(uid) {
-  if (!uid.ext.types) {
-    uid.ext.types = [];
-  }
-  if (!uid.ext.types.includes('ppuid')) {
+  if (!uid.ext) {
+    uid.ext = {
+      types: ['ppuid']
+    };
+  } else if (!uid.ext.types) {
+    uid.ext.types = ['ppuid'];
+  } else if (!uid.ext.types.includes('ppuid')) {
     uid.ext.types.push('ppuid');
   }
 }
--- a/modules/pubProvidedSystem.js
+++ b/modules/pubProvidedSystem.js
@@ -6,7 +6,7 @@
  */
 
 import {submodule} from '../src/hook.js';
-import * as utils from "../src/utils";
+import * as utils from '../src/utils.js';
 
 const MODULE_NAME = 'pubProvided';
 
@@ -38,7 +38,7 @@ export const pubProvidedSubmodule = {
    */
   getId(configParams) {
     let res = [];
-    if (Array.isArray(configParams.eids)) {
+    if (utils.isArray(configParams.eids)) {
       res = res.concat(configParams.eids);
     }
     if (typeof configParams.eidsFunction === 'function') {
--- a/modules/pubProvidedSystem.js
+++ b/modules/pubProvidedSystem.js
@@ -8,7 +8,7 @@
 import {submodule} from '../src/hook.js';
 import * as utils from '../src/utils.js';
 
-const MODULE_NAME = 'pubProvided';
+const MODULE_NAME = 'pubProvidedId';
 
 function addType(uid) {
   if (!uid.ext) {
@@ -23,7 +23,7 @@ function addType(uid) {
 }
 
 /** @type {Submodule} */
-export const pubProvidedSubmodule = {
+export const pubProvidedIdSubmodule = {
   /**
    * used to link submodule with config
    * @type {string}
@@ -62,4 +62,4 @@ export const pubProvidedSubmodule = {
 };
 
 // Register submodule for userId
-submodule('userId', pubProvidedSubmodule);
+submodule('userId', pubProvidedIdSubmodule);
--- a/modules/pubProvidedSystem.js
+++ b/modules/pubProvidedSystem.js
@@ -11,11 +11,11 @@ import * as utils from '../src/utils.js';
 const MODULE_NAME = 'pubProvided';
 
 function addType(uid) {
-  if(!uid.ext.types){
+  if (!uid.ext.types) {
     uid.ext.types = [];
   }
-  if(!uid.ext.types.includes('ppuid')){
-    uid.ext.types.push('ppuid')
+  if (!uid.ext.types.includes('ppuid')) {
+    uid.ext.types.push('ppuid');
   }
 }
 
--- a/modules/pubProvidedSystem.js
+++ b/modules/pubProvidedSystem.js
@@ -44,6 +44,7 @@ export const pubProvidedSubmodule = {
     if (typeof configParams.eidsFunction === 'function') {
       res = res.concat(configParams.eidsFunction());
     }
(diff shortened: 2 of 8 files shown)


integrationExamples/gpt/userId_example.html
```diff
@@ -109,6 +109,15 @@
             googletag.pubads().disableInitialLoad();
         });
 
+        const eidsProvider = function(){
+            return [{
+                source: "prov.com",
```
jdwieland8282, line 114: where did prov.com come from?  I'm not saying its wrong just want to understand where it came from.

jdwieland8282: oh I see its short hand for "provider" lets spell that out as "provider.com"

integrationExamples/gpt/userId_example.html
```diff
@@ -130,6 +139,25 @@
...
                 userSync: {
                     userIds: [{
+                        name: "publisherProvided",
+                        params: {
+                            eids: [{
+                                source: "example.com",
+                                uids:[{
+                                    id: "value read from cookie or local storage",
+                                    ext: {
+                                        third: "third"
```
jdwieland8282, line 149: can this ext be empty? instead of putting third: "third"

jdwieland8282 (commented):
Left a few comments inline

integrationExamples/gpt/userId_example.html
YerkovichM: Ok, it's just an example 
I can remove or put there more meaningful value

jdwieland8282 (approved):

bretg:
Wow that was fast. Comments:

- eidsFunction should be inside the `params` block. I think the code is right here, but the example above has it outside of params.
- I would suggest aligning the module name and the output to `pubProvided`. I know this is my fault from yesterday's proposal - thanks.
- Please update the RubiconBidAdapter to sniff for bid.userid.pubProvided where source="ppuid". Pass that into the same locations where it's currently sniffing for getConfig({user.id})
- Needs unit tests
- Needs docs (@jdwieland8282 ?)

@patmmccann - I know you had a desire to include some kind of "type" field here. Now that we've shifted to an `eids` format, is that still the case?

The Identity Committee (heads up @slayser8) ought to consider whether there are any conventions to define here. I'm proposing that an eids.source of "ppuid" is a convention that should be treated as user.id as in OpenRTB. Other conventions welcome, e.g. hashedEmail.

bretg (changes requested):
needs unit tests, please update RubiconBidAdapter to be the first to read from pubProvided. That could be done as a separate PR if desired.

modules/userId/eids.js
```diff
@@ -163,9 +163,13 @@ export function createEidsArray(bidRequestUserId) {
   let eids = [];
   for (const subModuleKey in bidRequestUserId) {
     if (bidRequestUserId.hasOwnProperty(subModuleKey)) {
-      const eid = createEidObject(bidRequestUserId[subModuleKey], subModuleKey);
-      if (eid) {
-        eids.push(eid);
+      if(subModuleKey === "pubProvided"){
```
coryhammon1, line 166: Needs a couple spaces.

coryhammon1 (changes requested):
Just the small formatting issue.

modules/pubProvidedSystem.js
```diff
@@ -0,0 +1,52 @@
...
+
+  /**
+   * performs action to obtain id and return a value.
+   * @function
+   * @param {SubmoduleParams} [configParams]
+   * @returns {{id: array}}
+   */
+  getId(configParams) {
+    let res = [];
+    if (Array.isArray(configParams.eids)) {
```
smenzer, line 41: isn't there a `utils` fn for this? does it matter which one is used?

modules/pubProvidedSystem.js
```diff
@@ -0,0 +1,52 @@
...
+  },
+
+  /**
+   * performs action to obtain id and return a value.
+   * @function
+   * @param {SubmoduleParams} [configParams]
+   * @returns {{id: array}}
+   */
+  getId(configParams) {
+    let res = [];
```
smenzer, line 40: i don't fully understand the use case that a publisher would have multiple sources of their own ID. the source should be their own domain, so while they may have multiple IDs (`eids[].uids[]`) why would they need more than one element in the `eids` array?

smenzer (changes requested):
a few questions, plus tests are needed for everything

modules/pubProvidedSystem.js
YerkovichM: @jdwieland8282 @bretg

wqi1972:
I see the pubProvided user id module add user id just as other user id module, somehow there is a problem here.
For other user id modules, each module binds with a constant source, and bid adapters can use that constant string to figure out what type of user id it is, and decide how it would like to handle it. But for pubProvided user id module, source value is provided by publishers. Which means bid adapters need to work closely with publishers, and add similar code as following
```
      const newsId = find(bidRequest.userIdAsEids, eid => eid.source === 'newscorp.com');
      if (newsId) {
        data['news_id'] = newsId.uids[0].id;
      }
```
where `newscorp.com` is a custom setting from publisher's configure file.

Suggestion:
Add a new attribute for pubProvided user id module, and use that attribute to include publishers provided id into bid adapter.
Here are the code:
1. change `createEidsArray` to
```
      if (subModuleKey === "pubProvided") {
        const pubProvidedId = bidRequestUserId["pubProvided"];
        pubProvidedId.forEach(id => id.type = 'pubProvided');
        eids = eids.concat(pubProvidedId);
      } else {
```
2. with that change, then inside each bid adapter, it will be able to include publishers provided id with following code
```
      const pubProvidedIds = bidRequest.userIdAsEids.filter(eid => eid.type === 'pubProvided');
      if (pubProvidedIds) {
        pubProvidedIds.forEach(pubProvidedId => {
          data[`${pubProvidedId.source}_id`] = pubProvidedId.uids[0].id;
        });
      }
```

abhinavsinha001:
I believe there are two use cases which this proposal can address : 
1. **Passing of First Party ID** (can be hashed PII or any other similar data) that publisher might want to pass to bid stream.
For this use case I agree the standardisation of `source` is necessary as @wqi1972 suggested - it is not only important for bidders but will help downstream DSPs who are the final consumers of this data. So I would recommend defining a industry standard source for this which will not change per domain / publisher e.g :
`pubprovided.id` or `firstparty.id` 

2. **Passing of standard Provider IDs** from Providers like TTD, Liveramp,Britepool or any similar New ID with a standard `source`. There can be multiple scenarios where this will be needed like the UserId Module not available in Prebid Publisher has separate implementation to generate / get these IDs.

So I propose we keep `source` as optional field and populate it with `pubprovided.id` or `firstparty.id` 

**Distinguish ID types using oRTB Agent Type** 
There can be cases where different publishers pass different type of identifiers , can be hashed email , probabilistic Ids. for making it easy for consumers of Ids , let's add support for [oRTB agent type parameter](https://github.com/InteractiveAdvertisingBureau/AdCOM/blob/master/AdCOM%20v1.0%20FINAL.md#list--agent-types-) which 

On the implementation I feel we should add validation for oRTB eids structure while reading the structure from function or allow passing simple id array which utility function converts to eids`eidsFunction: eidsProvider`

modules/pubProvidedSystem.js
jdwieland8282: The pub might be working with a 3rd party, who has done some probabilistic, or deterministic identification on their behalf, and they have a second uuid based off their own tech stack. That determination is publisher driven.  The pub could be sending two ids for monetization on different platforms Device, desktop, CTV.

jdwieland8282:
@wqi1972 I like that idea, this is indeed what we want. @YerkovichM , can you add Qi's suggested code? such that source = [some pub provided string] ex. source=newscorp.com

jdwieland8282:
@abhinavsinha001, I'm having a hard time following your comment. In #1 are you advocating for a world where source can equal, newscorp.com, bbc.com, cnn.com, or a world where source = 'ppuid'?  If the later how would DSPs know that the id they are receiving came from CNN or BBC?  

Re #2, standard Provider ID from TDD, Liveramp, etc.. can continue to use the userid modules right, no change needed from them. We don't envision all of them nesting their ids inside this new module. The second part of your question:

> 2\. There can be multiple scenarios where this will be needed like the UserId Module not available in Prebid Publisher has separate implementation to generate / get these IDs.

Is spot on, id vendors w/o a userid module who the publisher has chosen to work with can embed their id here.

abhinavsinha001:
> are you advocating for a world where source can equal, newscorp.com, bbc.com, cnn.com, or a world where source = 'ppuid'? If the later how would DSPs know that the id they are receiving came from CNN or BBC?

@jdwieland8282 I am proposing for constant source value by **default** like `source='ppuid'` or `source='firstparty.id'` if not explicitly specified,  because the domain context -  bbc.com  or cnn.com is always available in bid stream. Seems like unnecessary lookup overhead logic per domain for consumer of these ids(putting on technical hat here).
There can be cases where publisher owns multiple domain and somehow has common ID across multiple domains and in that case publishers can definitely explicitly specify / override the source value on all the domains e.g `source='bigorg.com'`

>Re #2, standard Provider ID from TDD, Liveramp, etc.. can continue to use the userid modules right, no change needed from them. We don't envision all of them nesting their ids inside this new module.

Yes, I just wanted to highlight any **New Identity** provider like these  which comes up in future and doesn't have a Prebid module yet.

smenzer:
the source field in oRTB is meant to differentiate "who's" ID is it. having a common "ppuid" value for the source means that you lose who is the actual owner. I think the source must be a valid publisher domain for who's ID it is. some publishers may have many domains and so the ID is not the domain from the request, but rather a group-wide ID. Think about large publishers Hearst or Conde who have tons of domains and may want to pass in their group-level ID (i.e. a hearst_id or something). This is basically what you gave an example, too, @abhinavsinha001. When this isn't the case (like with a single-domain publisher), I don't agree that it should default to a generic "global" value. This would still require extra look-ups by consumers, because they would have to first check if it's the generic value or not, then look up the url from the bid request...so I don't think we're saving them anything here.

I think if publishers are going to supply their own ID in the bid request via something like this module, they must specify their own unique `source` to keep things consistent for all consumers.

(Note - I have no idea if Hearst of Conde have group-level IDs nor if they'd want to use them in this way, I was just using them as an example to illustrate the point)

abhinavsinha001:
I don't think it would be much of an issue if we proceed with current approach(source=domain) since pubs can pass/change the source to any value - we can take a call to standardize this later based on feedback from parties sending and consuming this ID and other specific use case)
 
This is my final attempt at clarifying why I am pushing for having a standard ID source:

Consider a DSP receiving these IDs along with other from 1000+ domains:
**Case 1 :** a.com and multiple such domains sends source `parent-of-a.com` 
- How is the DSP supposed to know which ID is first party supplied ID? just by looking at source they won't know if `parent-of-a.com` is a new identity provider or parent domain of `a.com`
- There has to be offline communication here, some config and meta info needs to be maintained.

**Case 2:** Multiple New Publishers wants to pass hashed email
- For existing identity partner integration DSPs need some changes / updates for supporting an Identity provider(I believe the effort/timeline varies per DSP we need some feedback here).
- Now for each such request there has to be an offline communication between publisher and DSPs (seems inefficient)
- Reporting granularity explodes for SSPs and DSPs without adding any value - where they plan to report metrics on performance of identity provider/source.

If we keep these under same source, with minimal changes these Ids could be easily consumed, reported. For different types/category of of ID being sent we can use **atype** field.

smenzer:
> Consider a DSP receiving these IDs along with other from 1000+ domains:

I don't disagree with your concerns here...but if a DSP wants to use a publisher's ID (rather than one of the cross-publisher identity providers that have sub-modules), then I think the DSP needs to know exactly which IDs they are getting, rather than a generic `source` field which may actually cause problems on their end if they're not expecting that publisher to send IDs. Additionally, since all of these pubs would be generating their own UID, while it's unlikely, it's possible that two pubs use the same ID value to identify different users, and without some signal that it's actually a different publisher, there could be conflicts.

I agree that to kick this off it would make sense to leave the source as the domain, and if in the future there's a reason to standardize (and a good approach) then there should be a discussion.

test/spec/modules/eids_spec.js
```diff
@@ -237,8 +237,42 @@ describe('eids array generation for known sub-modules', function() {
       }]
     });
   });
+  it('PubProvided', function() {
+    const userId = {
+      pubProvided: [{
+        source: 'example.com',
+        type: 'pubProvided',
```
smenzer, line 244: this seems like a good alternative to using a generic `source` field; even though it's not a "standard" i don't think there's an issue with this for now. Probably makes sense to make sure this is included in the examples and documentation, though, since it's not mentioned anywhere but in your _spec files that you could use this.

YerkovichM: I will make sure that it will be in the documentation

YerkovichM:
@bretg Can you see changes in rubiconBidAdapter and verify them?

jdwieland8282:
Thanks @smenzer & @AbhijitBhosale72 for your comments and thoughts. Our preference is to have source = domain.com for now and standardize later if necessary.

test/spec/modules/eids_spec.js
patmmccann: any chance we could take a stab at defining a few conventions in this doc? eg sha256email etc?

modules/pubProvidedSystem.js
bretg: Use case is that a publisher has both:
- a site-specific ID they generated on their own
- and a DMP vendor who doesn't have a Prebid.js User ID Sub-Module

bretg:
I'm ok with adding a 'type' field, but assuming we want downstream entities to have it, we need to align with the [OpenRTB eids standard](https://iabtechlab.com/wp-content/uploads/2017/09/OpenRTB-3.0-Draft-Framework-for-Public-Comment.pdf) and put it in the ext:

```
        pbjs.setConfig({
            userSync: {
                userIds: [{
                    name: "pubProvided",
                    params: {
                        eids: [{
                            source: "example.com",
                            uids:[{
                                id: "value read from cookie or local storage",
                                ext: {
                                    type: "pubProvided"
                                }
                            }]
                        },{
                            source: "id-partner.com",
                            uids:[{
                                id: "value read from cookie or local storage"
                                ext: {
                                   type: "pubProvided"
                                }
                            }]
                        }]
                    },
                    eidsFunction: eidsProvider
```

Even better than just using a single 'type' value would be if the Identity PMC could come up with a small set of types like "ppuid", "dmp", "sha256email", or whatever.

integrationExamples/gpt/userId_example.html
```diff
@@ -134,6 +134,31 @@
                 // },
                 userSync: {
                     userIds: [{
+                        name: "pubProvided",
```
jdwieland8282, line 137: can we chance "pubProvided" to "pubProvidedId", that matches the convention the other id vendors are using.

jdwieland8282:
> Even better than just using a single 'type' value would be if the Identity PMC could come up with a small set of types like "ppuid", "dmp", "sha256email", or whatever.

@MarkoYerkovichRP can you add "ppuid", "dmp", "sha256email" in the type field of your example bid.userid example at the top?

bretg:
FWIW - I updated the [original issue](https://github.com/prebid/Prebid.js/issues/5690#issuecomment-693597321) with the current thinking

YerkovichM:
@jdwieland8282 description is updated
@bretg I use array
```
 "ext": {
              "types": ["ppuid", "dmp", "sha256email"]
          }
```
Is it ok?

jdwieland8282:
> Should pubProvided module add "ppuid" type to eids or type will only be defined in cofigs?

add "ppuid" as the default value, but add comments around where it should be configured and what the options are please.

bretg:
1. I'm not so sure the pubProvidedId module should add a default type.
2. I'm ok with types being an array, but don't have a use case where there could be multiple types

jdwieland8282:
> 2\. I'm ok with types being an array, but don't have a use case where there could be multiple types

Agreed. should just be a string. 

> * I'm not so sure the pubProvidedId module should add a default type.
In that case can type be omitted completely?  If the pub decides not to define it, should it be omitted or set to null?

bretg:
> If the pub decides not to define it, should it be omitted or set to null?

My opinion is that the Identity Committee should define conventions and if the pubs don't follow the conventions, things won't happen correctly.

As a concrete example:

- publishers can define eids[].source="example.com" and eids[].uids.ext.type=ppuid. Bid adapters (like rubicon) may recognize this type and pass it to their endpoints as desired. Or ignore.

test/spec/modules/userId_spec.js
```diff
@@ -30,13 +33,14 @@ import {zeotapIdPlusSubmodule} from 'modules/zeotapIdPlusIdSystem.js';
 import {sharedIdSubmodule} from 'modules/sharedIdSystem.js';
 import {haloIdSubmodule} from 'modules/haloIdSystem.js';
 import {server} from 'test/mocks/xhr.js';
+import {pubProvidedIdSubmodule} from '../../../modules/pubProvidedSystem.js';
```
smenzer, line 36: you don't need the `../../../` here do you? other modules don't have it

modules/userId/eids.js
```diff
@@ -182,9 +182,13 @@ export function createEidsArray(bidRequestUserId) {
   let eids = [];
   for (const subModuleKey in bidRequestUserId) {
     if (bidRequestUserId.hasOwnProperty(subModuleKey)) {
-      const eid = createEidObject(bidRequestUserId[subModuleKey], subModuleKey);
-      if (eid) {
-        eids.push(eid);
+      if (subModuleKey === 'pubProvided') {
+        eids = eids.concat(bidRequestUserId['pubProvided']);
```
smenzer, line 186: should this use `eids.push()` instead of `concat` like code below uses?

test/spec/modules/eids_spec.js
```diff
@@ -237,8 +237,42 @@ describe('eids array generation for known sub-modules', function() {
       }]
     });
   });
+  it('PubProvided', function() {
```
smenzer, line 240: for consistency, maybe change this to `pubProvidedId`

test/spec/modules/rubiconBidAdapter_spec.js
```diff
@@ -220,11 +227,25 @@ describe('the rubicon adapter', function () {
...
     bid.userId = {
-      lipb: { lipbid: '0000-1111-2222-3333', segments: ['segA', 'segB'] },
+      lipb: {lipbid: '0000-1111-2222-3333', segments: ['segA', 'segB']},
       idl_env: '1111-2222-3333-4444',
-      sharedid: { id: '1111', third: '2222' },
+      sharedid: {id: '1111', third: '2222'},
       tdid: '3000',
-      pubcid: '4000'
+      pubcid: '4000',
+      pubProvided: [{
```
smenzer, line 235: I think this needs to be `pubProvidedId` now

test/spec/modules/rubiconBidAdapter_spec.js
```diff
@@ -1391,9 +1412,37 @@ describe('the rubicon adapter', function () {
             });
           });
 
+          describe('PubProvided support', function () {
+            it('should send PubProvided when userIdAsEids contains PubProvided ids', function () {
+              const clonedBid = utils.deepClone(bidderRequest.bids[0]);
+              clonedBid.userId = {
+                pubProvided: [{
```
smenzer, line 1419: naming consistency - `pubProvidedId`

modules/userId/eids.js
YerkovichM: bidRequestUserId['pubProvided'] is an array of eids. Code below pushes single eid element

smenzer: ah true, good point. i forgot the pubProvidedIds were arrays still.

jdwieland8282:
@MarkoYerkovichRP since we already have an "atype" in rtb ~2.5 & 3.0 lets change "type" to "stype"

smenzer:
> @MarkoYerkovichRP since we already have an "atype" in rtb ~2.5 & 3.0 lets change "type" to "stype"

what does the "s" stand for, just curious? signal?

smenzer:
I was reviewing the code and just thought of something...since there's no actual call being made to a server and the IDs are being passed in directly in the publisher's config, why not just use the "value" syntax - this way you avoid a call to getId() which then stores the id temporarily for decode() to then make usable by the userId module. Instead of setting "params" in the pub config, you could just use "value" with the same syntax, then I think you only need your decode function.

This could simplify the submodule and probably be more efficient technically speaking. thoughts?

jdwieland8282:
> > @MarkoYerkovichRP since we already have an "atype" in rtb ~2.5 & 3.0 lets change "type" to "stype"
> 
> @jdwieland8282 what does the "s" stand for, just curious? signal?

"source"

I'm pro simplification, but I'll defer to @MarkoYerkovichRP on whether functionality will be maintained under such a code change

YerkovichM:
> value" with the same syntax, then I think you only need your decode function.

As see from the userId module "value" does not go through decode() function. So changes will affect functionality.
```
    } else if (submodule.config.value) {
      // cache decoded value (this is copied to every adUnit bid)
      submodule.idObj = submodule.config.value;
    }
```
@jdwieland8282 @smenzer

coryhammon1 (approved):

smenzer:
> > value" with the same syntax, then I think you only need your decode function.
> 
> As I see from the userId module "value" does not go through decode() function. So changes will affect functionality.
> 
> ```
>     } else if (submodule.config.value) {
>       // cache decoded value (this is copied to every adUnit bid)
>       submodule.idObj = submodule.config.value;
>     }
> ```
> 
> @jdwieland8282 @smenzer

@YerkovichM ah good point, i knew setting value had a shortcut but forgot that it was literally just direct (and i didn't take the 2 min to look at the code, sorry!).

jdwieland8282:
dev-docs PR: https://github.com/prebid/prebid.github.io/pull/2357

modules/rubiconBidAdapter.js
```diff
@@ -538,6 +539,12 @@ export const spec = {
       if (sharedId) {
         data['eid_sharedid.org'] = `${sharedId.uids[0].id}^${sharedId.uids[0].atype}^${sharedId.uids[0].ext.third}`;
       }
+      const pubProvidedIds = bidRequest.userIdAsEids.filter(eid => (eid.uids[0] && eid.uids[0].ext && eid.uids[0].ext.stype === 'ppuid'));
+      if (pubProvidedIds) {
+        pubProvidedIds.forEach(pubProvidedId => {
+          data[`${pubProvidedId.source}_id`] = pubProvidedId.uids[0].id;
```
bretg, line 545: As noted above, I think it would be safer to accept eid.uids[].ext.stype === 'ppuid' on any index of the uids array.

More importantly, for the fastlane.json interface, the exchange team wants to receive this on the existing `ppuid` field, not `${pubProvidedId.source}_id`.

modules/rubiconBidAdapter.js
```diff
@@ -251,7 +251,8 @@ export const spec = {
       const eids = utils.deepAccess(bidderRequest, 'bids.0.userIdAsEids');
       if (eids && eids.length) {
         // filter out unsupported id systems
-        utils.deepSetValue(data, 'user.ext.eids', eids.filter(eid => ['adserver.org', 'pubcid.org', 'liveintent.com', 'liveramp.com', 'sharedid.org'].indexOf(eid.source) !== -1));
+        utils.deepSetValue(data, 'user.ext.eids', eids.filter(eid => ['adserver.org', 'pubcid.org', 'liveintent.com', 'liveramp.com', 'sharedid.org'].indexOf(eid.source) !== -1 ||
+          (eid.uids[0] && eid.uids[0].ext && eid.uids[0].ext.stype === 'ppuid')));
```
bretg, line 255: The exchange team doesn't want to receive pubProvided IDs on the eids yet. Please revert this particular change and instead add to the block of code just below the comment "set user.id value from config value".

In that block, if `config.getConfig('user.id')` doesn't return anything, then look for the first `eid.uids[*].ext.stype === 'ppuid'` and set that value in the OpenRTB as user.id.

I think it would be safer to accept eid.uids[].ext.stype === 'ppuid' on any index of the uids array, not just the 0th element. I think the way the code is written here, if the publisher set two elements, with "stype:dmp" and "stype:ppuid" second, we wouldn't pick up the ppuid.

bretg (changes requested):

smenzer (commented):
looking good to me at this point. one comment on the way the IDs are made available in the `userIds` object to bid adapters, but once we address that i think this is good to go unless anyone else has comments / issues.

bretg (approved):
I fixed the usage example to move eidsFunction inside the params block.

Signing off.

bretg:
@smenzer - @jdwieland8282 updated the doc PR last week.

bretg:
In order to get this in today, we're going to go ahead and merge. I believe @smenzer 's comment is covered in documentation -- https://github.com/prebid/prebid.github.io/pull/2357

PR #5774, opened 2020-09-21
Reconciliation Real Time Data Provider

vfedoseev:
## Type of change

- [x] Feature

## Description of change
Reconciliation RTD provider allows publishers to integrate Reconciliation SDK via Prebid.js.
The purpose of Reconciliation SDK module is to collect supply chain structure information and vendor-specific impression IDs from suppliers participating in ad creative delivery and report it to the Reconciliation Service, allowing publishers, advertisers and other supply chain participants to match and reconcile ad server, SSP, DSP and veritifation system log file records. Reconciliation SDK was created as part of TAG DLT initiative ( https://www.tagtoday.net/pressreleases/dlt_9_7_2020 ).

vladimir.fedoseev@fiducia.eco

## Docs PR
https://github.com/prebid/prebid.github.io/pull/2395

### DIFF
--- a/modules/reconciliationRtdProvider.js
+++ b/modules/reconciliationRtdProvider.js
@@ -17,15 +17,10 @@
  */
 
 import { submodule } from '../src/hook.js';
-import { config } from '../src/config.js';
 import { ajaxBuilder } from '../src/ajax.js';
 import * as utils from '../src/utils.js';
 import find from 'core-js-pure/features/array/find.js';
 
-/** @type {string} */
-const MODULE_NAME = 'realTimeData';
-/** @type {string} */
-const SUBMODULE_NAME = 'reconciliation';
 /** @type {Object} */
 const MessageType = {
   IMPRESSION_REQUEST: 'rsdk:impression:req',
@@ -263,22 +258,19 @@ export const track = {
 
 /**
  * Set custom targetings for provided adUnits
- * call callback (onDone) when ready
- * @param {adUnit[]} adUnits
- * @param {function} onDone callback function
+ * @param {string[]} adUnitsCodes
+ * @return {Object} key-value object with custom targetings
  */
-function getReconciliationData(adUnits, onDone) {
-  let dataToReturn = adUnits.reduce((rp, cau) => {
-    const adUnitCode = cau && cau.code;
-
+function getReconciliationData(adUnitsCodes) {
+  let dataToReturn = adUnitsCodes.reduce((rp, adUnitCode) => {
     if (!adUnitCode) {
       return rp;
     }
 
     const adSlot = getSlotByCode(adUnitCode);
+
     rp[adUnitCode] = {
-      RSDK_ADID:
-        cau.transactionId || utils.generateUUID(),
+      RSDK_ADID: utils.generateUUID(),
       RSDK_AUID: adSlot ? adSlot.getAdUnitPath() : adUnitCode,
     };
 
@@ -288,7 +280,7 @@ function getReconciliationData(adUnits, onDone) {
   // Track init event
   trackInit(dataToReturn);
 
-  return onDone(dataToReturn);
+  return dataToReturn;
 }
 
 /** @type {RtdSubmodule} */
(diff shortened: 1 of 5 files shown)


vfedoseev:
Good day!

It seems that CircleCI Test failed on another provider:
`
teadsBidAdapter
    buildRequests
      ✗ should add timeToFirstByte info to payload
	expected '' to deeply equal '0'
`
Could you please check it. Thank you!

bretg:
We need to make sure that this new RTD provider is aligned with the revisions being made in https://github.com/prebid/Prebid.js/pull/5783. And we'll need a docs PR. You can base docs off of https://github.com/prebid/prebid.github.io/blob/master/dev-docs/modules/browsiRtdProvider.md

vfedoseev:
@bretg, thank you for the information.
Could you please clarify if we need to commit any changes to align with new revisions or we should just wait for your check?

> And we'll need a docs PR. 
Sure, I'll prepare it and link it to this PR. Thank your for the example provided.

bretg:
Check out the docs PR at https://github.com/prebid/prebid.github.io/pull/2268

we might be a little delayed on this review because the RTD-core is so new that the reviewers aren't familiar with it and what to look out for. Once #5783 is merged, we're going to meet to discuss RTD submodules with the reviewers.

bretg:
This PR needs to change... the `getData` interface is being replaced with `getBidRequestData`

The `getReconciliationData` function needs to have a different signature: (reqBidsConfigObj, callback, config, userConsent) - loop through the reqBidsConfigObj to read and set the desired data and then call the callback with the auction ID.

vfedoseev:
@bretg, thnak you for the updated info.
Am I right that if wee need to provide only targeting data, we should implement only `getTargetingData` method?

bretg:
Correct.

vfedoseev:
Hello, @bretg.
Thank you for comments. We've updated the module and added PR for docs: https://github.com/prebid/prebid.github.io/pull/2395

modules/reconciliationRtdProvider.js
```diff
@@ -0,0 +1,313 @@
...
+export function stringify(query) {
+  const parts = [];
+
+  for (let key in query) {
+    if (query.hasOwnProperty(key)) {
+      let val = query[key];
+      if (typeof query[key] !== 'object') {
+        parts.push(`${key}=${encodeURIComponent(val)}`);
+      } else {
+        parts.push(`${key}=${encodeURIComponent(stringify(val))}`);
```
msm0504, line 200: Since this is recursively calling `stringify`, I think the result could end up looking like:
```
key=key=value
``` 
Is this the intended behavior?

modules/reconciliationRtdProvider.js
```diff
@@ -0,0 +1,313 @@
...
+  }
+}
+
+/**
+ * Set custom targetings for provided adUnits
+ * @param {string[]} adUnitsCodes
+ * @return {Object} key-value object with custom targetings
+ */
+function getReconciliationData(adUnitsCodes) {
+  let dataToReturn = adUnitsCodes.reduce((rp, adUnitCode) => {
```
msm0504, line 265: I'd suggest changing this to `adUnitsCodes.forEach` and within that loop, populate 2 different structures:
```
dataToReturn - object
adUnitsToTrack - list of objects
```
`adUnitsToTrack` would be the structure needed in the `trackInit` function. This would avoid using reduce to create 1 structure and then map to create the 2nd, both from the same data.

test/spec/modules/reconciliationRtdProvider_spec.js
```diff
@@ -0,0 +1,63 @@
+import { reconciliationSubmodule, track } from 'modules/reconciliationRtdProvider.js';
```
msm0504, line 1: Please add additional unit tests. Coverage is currently at ~35%

msm0504 (changes requested):

modules/reconciliationRtdProvider.js
vfedoseev: Yes, it is supposed to work this way: nested objects are stringified to query-string and URL-encoded:
`{foo: 'bar', nested: {key1: 'value1', key2: 'value2'}}` 
becomes
`foo=bar&nested=key1%3Dvalue1%26key2%3Dvalue2`

modules/reconciliationRtdProvider.js
vfedoseev: @msm0504, thank you.
Modified the loop.

test/spec/modules/reconciliationRtdProvider_spec.js
vfedoseev: Fixed. Coverage is ~92%

vfedoseev:
@msm0504, thank you for your reply.
I've updated PR.

msm0504 (approved):

jsnellbaker (approved):

PR #5779, opened 2020-09-22
Rubicon Bid Adapter: Let host be configurable

robertrmartinez:
## Type of change
- [X] Feature

## Description of change
This allows users to configure their endpoints to point to different Rubicon environments.

For now, we will not let users configure the .rubiconproject.com domains, but rather, the former part only.


idettman (approved):

patmmccann:
are you guys releasing a http2 endpoint?

bretg:
No timeline yet on HTTP2 @patmmccann

patmmccann:
this is unfortunate. One can see the fastlane endpoint underperforming other endpoints in the reconnects https://www.webpagetest.org/result/200821_VY_db7f9322bbf5cf0ff0f4a5c21c292463/1/details/#step1_request61

PR #5780, opened 2020-09-22
Rubicon: PVID & Standardize rubicon get config calls

robertrmartinez:
## Type of change
- [X] Feature

### Page View ID
Users can now generate new page view ids on a given page load to accommodate use cases where needed (galleries for example)

### Rubicon Set Config Standardization
With some new updates, we need to keep a preserved running rubicon config so nothing gets overwritten.
The standard `pbjs.setConfig` functionality does not do any merging, thus, it is up to subscribers to handle that.
So we now keep a running `rubiConf`


modules/rubiconAnalyticsAdapter.js
```diff
@@ -51,27 +51,16 @@ const cache = {
...
-    return fpkvs;
-  } else {
-    return {...fpkvs, ...newKvs};
-  }
-}
-
-let integration, ruleId, wrapperName;
-// listen for any rubicon setConfig events and save them to appropriate fields!
+export let rubiConf = {
+  pvid: utils.generateUUID().slice(0, 8)
```
robertrmartinez, line 55: On intial page load we set first pvid automatically

modules/rubiconAnalyticsAdapter.js
```diff
@@ -51,27 +51,16 @@ const cache = {
...
 config.getConfig('rubicon', config => {
-  let rubiConf = config.rubicon;
-  integration = rubiConf.int_type || integration || DEFAULT_INTEGRATION;
-  ruleId = rubiConf.rule_name || ruleId;
-  wrapperName = rubiConf.wrapperName || wrapperName;
-  fpkvs = rubiConf.fpkvs ? updateFpkvs(fpkvs, rubiConf.fpkvs) : fpkvs
+  utils.mergeDeep(rubiConf, config.rubicon);
+  if (utils.deepAccess(config, 'rubicon.updatePageView') === true) {
+    rubiConf.pvid = utils.generateUUID().slice(0, 8)
+  }
```
robertrmartinez, line 63: deep merge everything, but if this specific setConfig call wants us to regenerate the pvid, we do so.

modules/rubiconAnalyticsAdapter.js
```diff
@@ -163,13 +152,13 @@ function sendMessage(auctionId, bidWonId) {
   let referrer = config.getConfig('pageUrl') || (auctionCache && auctionCache.referrer);
   let message = {
     eventTimeMillis: Date.now(),
-    integration,
-    ruleId,
+    integration: rubiConf.int_type || DEFAULT_INTEGRATION,
+    ruleId: rubiConf.rule_name,
```
robertrmartinez, line 156: if never set, then it will be `undefined` here and thus not sent after the stringify occurs!

modules/rubiconAnalyticsAdapter.js
```diff
@@ -359,10 +348,9 @@ export function parseBidResponse(bid, previousBidResponse, auctionFloorData) {
   ]);
 }
 
-function getPageViewId() {
-  if (prebidGlobal.rp && typeof prebidGlobal.rp.generatePageViewId === 'function') {
-    return prebidGlobal.rp.generatePageViewId(false);
-  }
+function getFpkvs() {
+  const isValid = rubiConf.fpkvs && typeof rubiConf.fpkvs === 'object' && Object.keys(rubiConf.fpkvs).every(key => typeof rubiConf.fpkvs[key] === 'string');
+  return isValid ? rubiConf.fpkvs : {};
```
robertrmartinez, line 353: makes sure any set fpkvs are valid before merging, else just empty

harpere (approved):

PR #5777, opened 2020-09-21
Audigent RTD Provider HaloId Support & RTD Phase 3 Compliance

anthonylauzon:
## Type of change
- [ ] Bugfix
- [ ] Feature
- [ ] New bidder adapter   
- [ ] Code style update (formatting, local variables)
- [ ] Refactoring (no functional changes, no api changes)
- [ ] Build related changes
- [ ] CI related changes
- [ ] Does this change affect user-facing APIs or examples documented on http://prebid.org?
- [X] Other

## Description of change
This PR expands the Audigent RTD provider module to allow support for retrieving Audigent segments for the recently added HaloId among other supported id systems.

## Other information
Related PRs:

Halo Id System Module: https://github.com/prebid/Prebid.js/pull/5524
Halo Id System Documentation: https://github.com/prebid/prebid.github.io/pull/2320
Audigent RTD Provider Module: https://github.com/prebid/Prebid.js/pull/4834

PR #5783, opened 2020-09-23
Real Time Data Module - Phase3

omerBrowsi:
Phase 3 of the real-time data module,
according to [https://docs.google.com/document/d/1A9mPkH8PiuqmrtKmIao42_losnopwrCGLtLJPv3L4oM/edit?ts=5f623873#](https://docs.google.com/document/d/1A9mPkH8PiuqmrtKmIao42_losnopwrCGLtLJPv3L4oM/edit?ts=5f623873#)

Issue - https://github.com/prebid/Prebid.js/issues/4610


bretg:
Adding @karimMourra because there may be JW impact here.

robertrmartinez:
@omerBrowsi 

Couple linting issues:

image

modules/rtdModule/index.js
```diff
@@ -159,130 +180,147 @@ export function init(config) {
...
-      const _mergedData = deepMerge(setDataOrderByProvider(subModules, data));
-      if (Object.keys(_mergedData).length) {
-        setDataForPrimaryAdServer(_mergedData);
-      }
-    }
-    next(adUnits);
-  });
-}
+export function getAdUnitTargeting(auction) {
+  const relevantSubModules = subModules.filter(sm => typeof sm.getTargetingData === 'function');
```
bretg, line 295: I don't understand how this is asynchronous -- can a submodule hang us up here? Seems like we need some kind of timeout?

modules/browsiRtdProvider.js
```diff
@@ -283,36 +248,37 @@ export const browsiSubmodule = {
...
 };
 
-function init(config, gdpr, usp) {
+function init() {
   return true;
 }
 
-export function beforeInit(config) {
+function beforeInit(config) {
   const confListener = config.getConfig(MODULE_NAME, ({realTimeData}) => {
```
bretg, line 262: I think only RTD-core should be a config listener -- it should then call the init function for each sub-module.

modules/rtdModule/index.js
```diff
@@ -159,130 +180,147 @@ export function init(config) {
...
   });
 
-  function onDataReceived(data) {
-    if (processDone) {
-      return
+  if (!prioritySubModules.length) {
+    return exitHook();
+  }
+
+  function onGetBidRequestDataCallback(auctionId) {
```
bretg, line 267: Is the auctionId always available when this gets here? From looking at the code in src/prebid.js, not clear that it is. Maybe a change is needed there.

bretg (changes requested):
Went through to understand it enough to update the docs. Areas I didn't cover:
- correct behavior of the async timer stuff in setBidRequestsData
- only skimmed the tests. Wonder if there's a way to better test waitForIt and the eventhandler scenarios.

bretg:
docs PR https://github.com/prebid/prebid.github.io/pull/2268

modules/rtdModule/index.js
```diff
@@ -159,130 +180,147 @@ export function init(config) {
...
+  if (!relevantSubModules.length) {
+    return exitHook();
+  }
+
+  if (prioritySubModules.length) {
+    waitTimeout = setTimeout(exitHook, _moduleConfig.auctionDelay);
+  }
+
+  relevantSubModules.forEach(sm => {
+    sm.getBidRequestData(reqBidsConfigObj, onGetBidRequestDataCallback.bind(sm), sm.config, _userConsent)
```
karimMourra, line 260: Shouldn't this support use case #2 in the [RTD module design doc](https://docs.google.com/document/d/1A9mPkH8PiuqmrtKmIao42_losnopwrCGLtLJPv3L4oM/edit?ts=5f623873#) ? 

> Update the AdUnits in the bidRequest object before going to the bidder (e.g. floors, first party data, adunit viewability)

Shouldn't the adUnits be passed here to allow the submodule to enrich them with data before the auction ?

karimMourra (changes requested):

modules/rtdModule/index.js
```diff
@@ -159,130 +180,147 @@ export function init(config) {
...
-    if (typeof (_moduleConfig.auctionDelay) === 'undefined') {
-      _moduleConfig.auctionDelay = 0;
-    }
-    // delay bidding process only if auctionDelay > 0
-    if (!_moduleConfig.auctionDelay > 0) {
-      getHook('bidsBackCallback').before(setTargetsAfterRequestBids);
-    } else {
-      getGlobal().requestBids.before(requestBidsHook);
-    }
+    getGlobal().requestBids.before(setBidRequestsData, 40);
```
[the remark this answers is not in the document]
omerBrowsi: we want to call it only after consent managment objects are set (so the submodules could use it)

modules/rtdModule/index.js
```diff
@@ -159,130 +180,147 @@ export function init(config) {
...
+  function onGetBidRequestDataCallback(auctionId) {
+    utils.logInfo('onGetBidRequestDataCallback', auctionId, this);
+    if (isDone) {
+      return;
     }
-    dataReceived[this.name] = data;
-    if (shouldWaitForAllSubModules || (this.config && this.config.waitForIt)) {
-      callbacksExpected--
+    if (this.config && this.config.waitForIt) {
+      callbacksExpected--;
```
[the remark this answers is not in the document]
omerBrowsi: it is, that's what the condition is for - ` this.config.waitForIt`

modules/rtdModule/index.js
omerBrowsi: it's not, we decided only bid modifications will be asynchronous and will use the auction delay parameter.
for ad unit targeting we assume the provider has enough time to call the endpoint before.
(personally, I don't mind adding another delay here, but I don't think it's necessary)

modules/rtdModule/index.js
omerBrowsi: We actually not doing anything with the auction ID, it's there by mistake (debugging leftovers), I'll remove it.

modules/rtdModule/index.js
omerBrowsi: we are passing the global object - it will contain ad units if defined on request bids (and will be on the global ad units object if defined by `addAdunit`).
If you have a different suggestion I'll be happy to hear it

modules/rtdModule/index.js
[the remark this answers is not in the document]
bretg: Isn't that the point of `_moduleConfig = realTimeData`? Won't the initSubModules call be able to get the config?

modules/rtdModule/index.js
bretg: Guess I don't understand how this code works when there are multiple concurrent auctions. This can and will happen.

Maybe we need `callbacksExpected` and `waitTimeout` to be a hashes indexed by auctionId?

modules/rtdModule/index.js
karimMourra: @omerBrowsi I understand. 
How should we make our realTimeData available then to bid adapters when `buildRequests` is called ? Before, the rtd module would add them to a `realTimeData` object available in the `bidRequests` param. This is no longer the case, breaking the interface for bid adapters. Can we provide backwards compatibility for this ?

bretg: There are only a couple of RTD sub-modules, we're not going to implement backwards compatibility.

Your RTD sub-mod needs to parse the reqBidsConfigObj and inject/remove whatever data you want.

bretg:
For the record, there are 3 other RTD modules that need to be changed as part of this overall effort:

- Audigent -- https://github.com/prebid/Prebid.js/pull/5777 (@anthonylauzon)
- JW Player (@karimMourra)
- Reconcilliation -- https://github.com/prebid/Prebid.js/pull/5774 (@vfedoseev )

modules/rtdModule/index.js
bretg: I'll let this one go for now. If we have a problem, we may need to revisit.

modules/browsiRtdProvider.js
robertrmartinez: Yeah I agree with this.

Seems like bad practice to gather data from here, when the core module should gather and just pass the params associated with browsi.

modules/browsiRtdProvider.js
```diff
@@ -283,36 +248,37 @@ export const browsiSubmodule = {
...
-        pr => pr.name && pr.name.toLowerCase() === 'browsi')[0].params;
-      confListener();
-      _moduleParams.auctionDelay = realTimeData.auctionDelay;
-      _moduleParams.timeout = realTimeData.timeout || DEF_TIMEOUT;
-    } catch (e) {
-      _moduleParams = {};
-    }
-    if (_moduleParams.siteKey && _moduleParams.pubKey && _moduleParams.url) {
+    setModuleData(realTimeData);
+    confListener();
```
robertrmartinez, line 264: why are we calling confListener within confListener again?

Once you subscribe once, it should remain subscribed.

(Even though we are recommending not submodules not listening to the core module setConfigs.

But just wanted to point out.

Unless I am missing something?

modules/browsiRtdProvider.js
```diff
@@ -283,36 +248,37 @@ export const browsiSubmodule = {
...
     } else {
       utils.logError('missing params for Browsi provider');
     }
   });
 }
 
+export function setModuleData(realTimeData) {
+  try {
+    _moduleParams = realTimeData.dataProviders && realTimeData.dataProviders.filter(
+      pr => pr.name && pr.name.toLowerCase() === 'browsi')[0].params;
```
robertrmartinez, line 276: Yeah this should be passed to you in the core util passbacks, then you can run a `setModuleData`

bretg:
Met with the code review team. Takeaways:

- userConsent isn't always available at time of init. The plan is to call init with whatever's available in subModule init. Modules that need consentData will need to wait until getBidRequestData. That routine will be modified to get the current userConsent data every time.
- Remove _moduleConfig.auctionDelay -- give sub-modules a chance to provide data if they don't need the delay
- Docs should state that if no sub-module are flagged as "waitforit", that's the same as auctiondelay=0
- We can get rid of auction ID from the callback. Javascript is magic.

bretg (approved):

bretg:
@mkendall07 - any idea what this LGTM warning is about?

```
301  for (let i = relevantSubModules.length; i--; i > 0) {
NEW - This expression has no effect.
```

vfedoseev:
> @mkendall07 - any idea what this LGTM warning is about?
> 
> ```
> 301  for (let i = relevantSubModules.length; i--; i > 0) {
> NEW - This expression has no effect.
> ```

Hello, sorry for interrupt)
Seems that `i > 0` in this loop does not make sense since this part is executed in each iteration.

So it could be replaced with
`for (let i = relevantSubModules.length; i--; ) {`

bretg:
Of course - thanks for the "interrupt" @vfedoseev -- fixed.

modules/rtdModule/index.js
```diff
@@ -159,130 +180,142 @@ export function init(config) {
...
-  for (let i = modules.length; i--; i > 0) {
-    if (data[modules[i].name]) {
-      rd.push(data[modules[i].name])
+  // get data
+  const adUnitCodes = auction.adUnitCodes;
+  if (!adUnitCodes) {
+    return;
+  }
+  let targeting = [];
+  for (let i = relevantSubModules.length; i > 0; i--) {
```
vfedoseev, line 301: @bretg 
This loop iterates from indexes `relevantSubModules.length` to `1` instead of 
`relevantSubModules.length-1` to `0`
Is not it better to rewrite it like:
`for (let i = relevantSubModules.length - 1; i >= 0; i--) {`
or
`for (let i = relevantSubModules.length; i--; ) {`
?

modules/rtdModule/index.js
```diff
@@ -159,130 +180,142 @@ export function init(config) {
...
-      const _mergedData = deepMerge(setDataOrderByProvider(subModules, data));
-      if (Object.keys(_mergedData).length) {
-        setDataForPrimaryAdServer(_mergedData);
-      }
-    }
-    next(adUnits);
-  });
-}
+export function getAdUnitTargeting(auction) {
+  const relevantSubModules = subModules.filter(sm => typeof sm.getTargetingData === 'function');
```
bretg, line 290: @omerBrowsi - this line is failing unit test. subModules at this point in the unit test is 

```
[{"name":"validSMWait","config":{"name":"validSMWait","waitForIt":true}},{"name":"validSM","config":{"name":"validSM","waitForIt":false}}]
```

modules/rtdModule/index.js
```diff
@@ -159,130 +180,142 @@ export function init(config) {
...
-  let dataReceived = {};
-  let processDone = false;
-  const dataWaitTimeout = setTimeout(done, _moduleConfig.auctionDelay || _moduleConfig.timeout || DEF_TIMEOUT);
-  subModules.forEach(sm => {
-    sm.getData(adUnits, onDataReceived.bind(sm));
+export function setBidRequestsData(fn, reqBidsConfigObj) {
+  _userConsent = getConsentData();
+
+  const relevantSubModules = subModules.filter(sm => typeof sm.getBidRequestData === 'function');
+  const prioritySubModules = relevantSubModules.filter(sm => !!(sm.config && sm.config.waitForIt));
```
karimMourra, line 241: we're filtering twice over the same subset. Under the hood this means 2 loops probably. I believe it would be more performant to loop once
```
const relevantSubModules = [];
const prioritySubModules = [];
subModules.forEach(sm => {
   if (typeof sm.getBidRequestData !== 'function') {
        return;
   }
   relevantSubModules.push(sm);
   const config = sm.config;
   if (config && config.waitForIt) {
       prioritySubModules.push(sm);
   }
});
```
This way you're looping n times instead of n+m.

karimMourra (changes requested):

jsnellbaker (approved):

modules/rtdModule/index.js
omerBrowsi: fixed

modules/rtdModule/index.js
```diff
@@ -159,130 +180,153 @@ export function init(config) {
...
+    const config = sm.config;
+    if (config && config.waitForIt) {
+      prioritySubModules.push(sm);
+    }
   });
 
-  function onDataReceived(data) {
-    if (processDone) {
-      return
+  const shouldDelayAuction = prioritySubModules.length && _moduleConfig.auctionDelay && _moduleConfig.auctionDelay > 0;
```
mkendall07, line 253: can we rename this param to `auctionDelayMs` (assuming it is in MS)

bretg: @mkendall07 - that would be a good idea, but it would be a breaking change for publishers -- some are already using the auctionDelay option and would have to change it. This could be a candidate for 6.0

karimMourra (approved):

Fawke:
@bretg @omerBrowsi 
I was going through the PR review guidelines for RTD module and noticed a RTD submodule must have associated md file which is missing for browsi.

>The RTD Provider must include a providerRtdProvider.md file. This file must have example parameters and document a sense of what to expect: what should change in the bidrequest, or what targeting data should be added?

src/targeting.js
```diff
@@ -609,6 +610,27 @@ export function newTargeting(auctionManager) {
     });
   }
 
+  function getAdUnitTargeting(adUnitCodes) {
```
mkendall07, line 613: perhaps a comment here would be helpful, example of how `ADSERVER_TARGETING` is getting set.

mkendall07 (approved):

bretg:
Merging - @Asafsham - please open a separate PR to add browsiRtdProvider.md

modules/rtdModule/index.js
karimMourra: @omerBrowsi , @bretg and I discussed the absence of `realTimeData` on the `bidRequest` and we would like to add it back. I can add it back, but before I do so, Is there a reason it was removed ?

bretg: I suspect it wasn't so much removed as the whole approach changed. It's entirely up to the RTD-submodule now -- RTD-core just passes the whole thing to each submodule where any change can be made. What we missed is that we should have set up a convention for each RTD-submodule to follow.

PR #5787, opened 2020-09-24
Add Auction Options Config

ncolletti:
## Type of change
- [X] Feature

## Description of change
By specifying bidders in the auctionOptions config you assert that you do want Prebid to wait for their bids to return before declaring the Auction Done, they're referred to as secondaryBidders. This may be helpful if you find there are a number of low performing and/or high timeout bidders in your page's rotation. End users should see faster auctions 

**Notes:** 
- secondaryBidders will still be a part of your auction but they must return bids before the last bid returns from the rest of your bidders in the auction.
- There is a test page in this PR to see the functionality in real time, it is difficult to test without delaying bid responses properly, taking suggestions on how to do that for a better visualization in the test page. The test page also contains additional logging that can be used to assist with visualizing when bids are returned, from who, and when the auction ends. This logging can be used by the end user in the same way on their production pages. We can also include this in the source and enable it with a **logging** boolean passed to auctionOptions config.
- Open to better naming of this feature and secondaryBidders.

```
pbjs.setConfig({
    'auctionOptions': {
        'secondaryBidders': ['doNotWaitForMe']
     }
});
```

## Prebid Docs PR
https://github.com/prebid/prebid.github.io/pull/2370


bretg:
This is an interesting feature @ncolletti , but I'd like to get a sense for why the config option is called "auctionTiming". Are there other "timing" options you guys have in mind? Maybe could rename this to "auctionOptions"?

ncolletti:
Good call.. currently there is only the one feature to set secondaryBidders on the auction, so auctionOptions make sense!

src/auction.js
```diff
@@ -397,10 +398,16 @@ export function auctionCallbacks(auctionDone, auctionInstance) {
...
   function adapterDone() {
     let bidderRequest = this;
+    let bidderRequests = auctionInstance.getBidRequests();
 
     bidderRequestsDone.add(bidderRequest);
-    allAdapterCalledDone = auctionInstance.getBidRequests()
-      .every(bidderRequest => bidderRequestsDone.has(bidderRequest));
+    allAdapterCalledDone = bidderRequests.filter(request => {
+      const auctionOptionsConfig = config.getConfig('auctionOptions');
+      if (auctionOptionsConfig && auctionOptionsConfig.secondaryBidders && !bidderRequests.every(bidder => includes(auctionOptionsConfig.secondaryBidders, bidder.bidderCode))) {
```
msm0504, line 406: Run the `bidderRequests.every` check outside of the filter function. The result should be the same each time, so now need to have this loop within a loop.

Also, just want to make sure I'm understanding this. When determining whether or not the auction is done, don't wait for secondary bidders, unless all of the bidders are secondary bidders. Is that correct?

msm0504 (changes requested):

src/auction.js
ncolletti: Thanks for the feedback. That is the correct logic. I pushed an update to fix the nested loop and address the legibility of the logic.

msm0504 (approved):

test/pages/auctionOptions.html
```diff
@@ -0,0 +1,199 @@
+<!-- This is a Test Page for configuring Auction Options. One bidder will be marked as secondary and the auction not wait for their bids to return.  -->
+<!DOCTYPE html>
```
mkendall07, line 2: test pages in this repo tend to get out of date. Perhaps this better suited to the documentation site?

ncolletti: That makes sense, I won't be able to host a built version of pbjs with auctionOptions included in order to show this page with jsfiddle, so I instead I will include the logging snippet to the troubleshooting page.

robertrmartinez:
Removed @harpere and myself since two core approvals need to come from separate companies as per the policy.

ncolletti:
@jsnellbaker Your review has been pending for about two weeks. Are you available or should it be reassigned? Thank you!

Fawke (approved):
LGTM

PR #5788, opened 2020-09-24
Improve warning and documentation for pubmatic publisherId

jsut:
## Type of change

- [ ] Bugfix
- [ ] Feature
- [ ] New bidder adapter   
- [ ] Build related changes
- [ ] CI related changes
- [x] Other

## Description of change

- The warning message you get if you use pubmatic config with a publisherId that is a JS numeric instead of a JS string is not
  super helpful if you aren't familiar with JS internals. Update the warning message to give a suggestion
  on a solution, and update the markdown documentation to explictly state that the ID needs to be wrapped
  in quotes.

### DIFF
--- a/modules/pubmaticBidAdapter.js
+++ b/modules/pubmaticBidAdapter.js
@@ -847,7 +847,7 @@ export const spec = {
   isBidRequestValid: bid => {
     if (bid && bid.params) {
       if (!utils.isStr(bid.params.publisherId)) {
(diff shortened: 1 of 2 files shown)


jsut:
It occurs to me now that another way to handle this would just be to use .toString() to make sure that the publisherId is a string instead of having this failure at all, which is probably a better approach, really.

pm-harshad-mane (approved):
Thank you @jsut , LGTM

PR #5760, opened 2020-09-16
Media type renderers

MikeSperone:
## Type of change

- [ ] Bugfix
- [X] Feature
- [ ] New bidder adapter   
- [ ] Code style update (formatting, local variables)
- [ ] Refactoring (no functional changes, no api changes)
- [ ] Build related changes
- [ ] CI related changes
- [X] Does this change affect user-facing APIs or examples documented on http://prebid.org?
- [ ] Other

## Description of change

As per issue #3231, this allows the publisher to define a renderer to be implemented only in a specific mediaType.

A publisher could use it like this:
```
pbjs.addAdUnit({
    code: 'video1',
      // This renderer would work the same as it does now...
    renderer: {
        url: 'example.com/publishersCustomRenderer.js',
        render: function(bid) { renderAdUnit(...) }
    },
    mediaTypes: {
        video: {
            context: 'outstream',
            playerSize: [640, 480],
              // but a renderer passed in here would apply only to this mediaType.
              // It would override the above renderer if that was defined. ...
            renderer: {
                url: 'example.com/videoRenderer.js',
                render: function (bid) { renderVideo(...) }
            }
        },
        display: {
            ...,
              // and you'd be able to do the same for each mediaType
            renderer: {
                url: 'example.com/displayRenderer.js',
                render: function(bid) { renderDisplay(...) }
            }
        },
    },
    ...
});
```

Be sure to test the integration with your adserver using the [Hello World](/integrationExamples/gpt/hello_world.html) sample page.
OK

- contact email of the adapter’s maintainer
- [ ] official adapter submission

For any changes that affect user-facing APIs or example code documented on http://prebid.org, please provide:
https://github.com/prebid/prebid.github.io/pull/2343

## Other information

Discussed in issue #3231

PR #5796, opened 2020-09-29
Integrate option to pass clickThrough urls to renderAd method

renebaudisch:
## Type of change
- [x] Feature

## Description of change
This change enables publishers/marketers to count clicks on prebid inventory in their distributing adserver.
The publisher may call render ad with a third param accepting an object where right now only clickThrough will be checked but can be extended.
If this property is set we will replace any occurrence of ${CLICKTHROUGH} inside the properties ad and adUrl with the transmitted option.
If no occurrence is found nothing will be replaced.

## Other information
To make use of this feature bidAdpaters just have to respond with tags including this macro to pass it to their param and system

### DIFF
--- a/src/prebid.js
+++ b/src/prebid.js
@@ -342,7 +342,7 @@ function emitAdRenderFail({ reason, message, bid, id }) {
  * @param  {string} id bid id to locate the ad
  * @alias module:pbjs.renderAd
  */
-$$PREBID_GLOBAL$$.renderAd = function (doc, id) {
+$$PREBID_GLOBAL$$.renderAd = function (doc, id, options) {
   utils.logInfo('Invoking $$PREBID_GLOBAL$$.renderAd', arguments);
   utils.logMessage('Calling renderAd with adId :' + id);
 
--- a/src/prebid.js
+++ b/src/prebid.js
@@ -356,9 +356,11 @@ $$PREBID_GLOBAL$$.renderAd = function (doc, id, options) {
         bid.adUrl = utils.replaceAuctionPrice(bid.adUrl, bid.cpm);
 
         // replacing clickthrough if submitted
-        const { clickThrough } = options;
-        bid.ad = utils.replaceClickThrough(bid.ad, clickThrough);
-        bid.adUrl = utils.replaceClickThrough(bid.adUrl, clickThrough);
+        if (options && options.clickThrough) {
+          const { clickThrough } = options;
+          bid.ad = utils.replaceClickThrough(bid.ad, clickThrough);
+          bid.adUrl = utils.replaceClickThrough(bid.adUrl, clickThrough);
+        }
 
         // save winning bids
         auctionManager.addWinningBid(bid);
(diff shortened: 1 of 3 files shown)


patmmccann:
Chatting in the js committee today, we're hoping #5760 gets merged first, as renderers on display to track clicks might intefere with multiformat ads that need renderer on outstream.

patmmccann:
Also, needs tests, thanks for the contribution!

src/prebid.js
```diff
@@ -354,6 +354,12 @@ $$PREBID_GLOBAL$$.renderAd = function (doc, id) {
         // replace macros according to openRTB with price paid = bid.cpm
         bid.ad = utils.replaceAuctionPrice(bid.ad, bid.cpm);
         bid.adUrl = utils.replaceAuctionPrice(bid.adUrl, bid.cpm);
+
+        // replacing clickthrough if submitted
+        const { clickThrough } = options;
```
harpere, line 359: should validate that `options` is defined.

src/utils.js
```diff
@@ -718,6 +718,11 @@ export function replaceAuctionPrice(str, cpm) {
   return str.replace(/\$\{AUCTION_PRICE\}/g, cpm);
 }
 
+export function replaceClickThrough(str, clicktag) {
+  if (!str) return;
+  return str.replace(/\$\{CLICKTHROUGH\}/g, clicktag);
```
harpere, line 723: should validate that `clicktag` is defined and not NULL, maybe that it's a string.

harpere (changes requested):
needs some validation and unit tests

harpere (changes requested):
updates look good, thx, but still needs unit tests

renebaudisch:
> updates look good, thx, but still needs unit tests

Ahm, yeah, tried my best... Is my first one ever, hope it fits.

renebaudisch:
Closed #5816 in favor to #5820

harpere (approved):

jsnellbaker (approved):

bretg:
I missed this one -- still kinda unclear on the real-world use case. The Prebid Universal Creative doesn't support this. 

@renebaudisch - could you outline a sample usage scenario so we can document this somewhere?

renebaudisch:
When using an adserver to render or deliver the prebid response there is actually no option to count clicks by the delivering adserver by passing their clickThrough as one normally would do when adding a redirect for an IO booking.

By adding this options we could provide an clickthrough url from the adserver to the renderAd method so the adserver of the publisher would also count clicks on prebid creatives as we cannot right now.

This should also affect some reports and statistics of some adservers as the average clickrate right now has either to avoid calculate them or using their AIs by a clickrate of 0,00% what actually reduces this rate.

Sure, both sides have to support this, so the adCode response has to use a macro and the publisher has to provide the clickthrough but right now I just couldn't and this just gives us this possibility.
But if one of them won't nothing will change either for noone...

PR #5799, opened 2020-09-29
Rubicon Bid Adapter: Fix some video request params

robertrmartinez:
Type of change
- [X] Bugfix

## Description of change
When we first implmented video requests through PBS we just mimic'd what we were sending into frankenstein, so now we need to fix it a bit.

1) it shouldn't hardcode the 'exp' parameter. Please change this:

   a) if s2sConfig.defaultTtl exists, send that value as 'exp'

   b) otherwise, don't send anything

2) tmax should be getting its value from bidRequest.timeout NOT getConfig(\{TTL});

### DIFF
--- a/modules/rubiconBidAdapter.js
+++ b/modules/rubiconBidAdapter.js
@@ -165,9 +165,9 @@ export const spec = {
         source: {
           tid: bidRequest.transactionId
         },
-        tmax: config.getConfig('TTL') || 1000,
+        tmax: bidderRequest.timeout,
         imp: [{
-          exp: 300,
+          exp: config.getConfig('s2sConfig.defaultTtl'),
           id: bidRequest.adUnitCode,
           secure: 1,
           ext: {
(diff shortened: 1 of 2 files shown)


modules/rubiconBidAdapter.js
```diff
@@ -165,9 +165,9 @@ export const spec = {
         source: {
           tid: bidRequest.transactionId
         },
-        tmax: config.getConfig('TTL') || 1000,
+        tmax: bidderRequest.timeout,
```
harpere, line 168: do we still want to default to `1000`?

modules/rubiconBidAdapter.js
```diff
@@ -165,9 +165,9 @@ export const spec = {
...
         imp: [{
-          exp: 300,
+          exp: config.getConfig('s2sConfig.defaultTtl'),
```
harpere, line 170: is there always an `s2sConfig` (with `defaultTtl`) when video is used?  If not, I guess we want to send `undefined`?

harpere (approved):

PR #5803, opened 2020-09-30
expose full user id config (including storage) to user id modules

smenzer:
## Type of change
- [ x ] Feature

## Description of change
Today, user id modules only receive the `params` object from publisher configuration. However, the `storage` object can also be useful to allow the submodule to know the storage type / name being used, what the refresh settings are, etc. To accomplish this, we have to pass the entire config object to the submodule in `getId`, `extendId`, and `decode`. 

This PR also updates every submodule and their tests to ensure they are still working as they were previously.

### DIFF
--- a/modules/britepoolIdSystem.js
+++ b/modules/britepoolIdSystem.js
@@ -29,7 +29,7 @@ export const britepoolIdSubmodule = {
   /**
    * Performs action to obtain id and return a value in the callback's response argument
    * @function
-   * @param {SubmoduleParams} [submoduleConfig]
+   * @param {SubmoduleConfig} [submoduleConfig]
    * @param {ConsentData|undefined} consentData
    * @returns {function(callback:function)}
    */
@@ -85,7 +85,7 @@ export const britepoolIdSubmodule = {
   },
   /**
    * Helper method to create params for our API call
-   * @param {SubmoduleParams} [submoduleConfigParams]
+   * @param {SubmoduleConfig} [submoduleConfigParams]
    * @param {ConsentData|undefined} consentData
    * @returns {object} Object with parsed out params
    */
--- a/modules/britepoolIdSystem.js
+++ b/modules/britepoolIdSystem.js
@@ -85,7 +85,7 @@ export const britepoolIdSubmodule = {
   },
   /**
    * Helper method to create params for our API call
-   * @param {SubmoduleConfig} [submoduleConfigParams]
+   * @param {SubmoduleParams} [submoduleConfigParams]
    * @param {ConsentData|undefined} consentData
    * @returns {object} Object with parsed out params
    */
--- a/modules/britepoolIdSystem.js
+++ b/modules/britepoolIdSystem.js
@@ -29,11 +29,12 @@ export const britepoolIdSubmodule = {
   /**
    * Performs action to obtain id and return a value in the callback's response argument
    * @function
-   * @param {SubmoduleParams} [submoduleConfigParams]
+   * @param {SubmoduleParams} [submoduleConfig]
    * @param {ConsentData|undefined} consentData
    * @returns {function(callback:function)}
    */
-  getId(submoduleConfigParams, consentData) {
+  getId(submoduleConfig, consentData) {
+    const submoduleConfigParams = (submoduleConfig && submoduleConfig.params) || {};
     const { params, headers, url, getter, errors } = britepoolIdSubmodule.createParams(submoduleConfigParams, consentData);
     let getterResponse = null;
     if (typeof getter === 'function') {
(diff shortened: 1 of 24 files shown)


jlukas79 (approved):

smenzer:
@harpere what docs do you envision adding here? the docs website doesn't mention anything about how to build a user id module and i've updated the relevant comment docs within the user id module code. happy to provide more, just not sure where this would be documented.

FilipStamenkovic (approved):
LGTM.  
I tested it on my local, works fine. Just one small question:

You changed from `@param {SubmoduleParams} [submoduleConfigParams]` to `@param {SubmoduleParams} [submoduleConfig]` in every submodule. More specifically, you changed `submoduleConfigParams` to `submoduleConfig`. Why didn't you change `SubmoduleParams` to `SubmoduleConfig` also?

smenzer:
> You changed from `@param {SubmoduleParams} [submoduleConfigParams]` to `@param {SubmoduleParams} [submoduleConfig]` in every submodule. More specifically, you changed `submoduleConfigParams` to `submoduleConfig`. Why didn't you change `SubmoduleParams` to `SubmoduleConfig` also?

No reason really...I can do that now.

smenzer:
@FilipStamenkovic done...if you're ok with this, can you merge, please? thanks!

modules/britepoolIdSystem.js
```diff
@@ -84,7 +85,7 @@ export const britepoolIdSubmodule = {
   },
   /**
    * Helper method to create params for our API call
-   * @param {SubmoduleParams} [submoduleConfigParams]
+   * @param {SubmoduleConfig} [submoduleConfigParams]
```
FilipStamenkovic, line 88: I don't think you should have changed this one?

modules/liveIntentIdSystem.js
```diff
@@ -100,10 +100,11 @@ export const liveIntentIdSubmodule = {
    * `publisherId` params.
    * @function
    * @param {{unifiedId:string}} value
-   * @param {SubmoduleParams|undefined} [configParams]
+   * @param {SubmoduleParams|undefined} [config]
```
FilipStamenkovic, line 103: Here `SubmoduleConfig` instead of `SubmoduleParams`?

modules/userId/index.js
```diff
@@ -27,7 +27,7 @@
  *  If IdResponse#callback is defined, then it'll called at the end of auction.
  *  It's permissible to return neither, one, or both fields.
  * @name Submodule#extendId
- * @param {SubmoduleParams} configParams
+ * @param {SubmoduleParams} config
```
FilipStamenkovic, line 30: Here `SubmoduleConfig` instead of `SubmoduleParams`?

modules/userId/index.js
```diff
@@ -37,7 +37,7 @@
  * @summary decode a stored value for passing to bid requests
  * @name Submodule#decode
  * @param {Object|string} value
- * @param {SubmoduleParams|undefined} configParams
+ * @param {SubmoduleParams|undefined} config
```
FilipStamenkovic, line 40: also here?

FilipStamenkovic (changes requested):
I found couple of places where `SubmoduleParams` was not changed to `SubmoduleConfig`.  

What is the status with docs? I see @harpere requested some docs, is this still valid? I'm asking because if some docs are needed then I cannot merge this PR?

modules/britepoolIdSystem.js
smenzer: you're right

modules/liveIntentIdSystem.js
smenzer: yep

modules/userId/index.js
smenzer: yep, i forgot i changed extendId too, good catch

smenzer (commented):
good finds...updating again now

modules/userId/index.js
smenzer: yea

smenzer:
happy to provide docs, but like i mentioned before, I'm not sure where they belong. we don't really have docs on how to write a user id module AFAIK

FilipStamenkovic (approved):

FilipStamenkovic:
> happy to provide docs, but like i mentioned before, I'm not sure where they belong. we don't really have docs on how to write a user id module AFAIK

@harpere Do we need some docs, or can we merge this one?

smenzer:
I'd suggest we merge this and if @harpere identifies where we should add docs, I can follow up with a PR right away. i'd like to get this in before the other outstanding PRs for user id modules get merged to avoid having to rework those here.

since there are two approvals and tests are passing, i'll give it a few hours in case anyone objects, but then i'll go ahead and merge this.

cc @FilipStamenkovic

PR #5805, opened 2020-09-30
Floors Module update to include floorMin

mmoschovas:
Update to floors module to allow floorMin definition using setConfig({floors:...});

1) If floorMin exists, set floorValue to new property floorRuleValue.
2) If floorMin is greater than floorValue, set floorValue to floorMin.

Update to Rubicon Analytics Adapter to pass floorMin under auction.floors.floorMin if exists. Also includes update to pass floorRuleValue for each bid if floorMin exists

Update to floorsModule roundup functionality to fix to one decimal place prior to roundup. This will fix issues in which JS evaluates a whole number to include a very small decimal value that forces a roundup to the next whole number.

## Type of change

- [ x] Bugfix
- [ x] Feature
- [ ] New bidder adapter   
- [ ] Build related changes
- [ ] CI related changes
- [ ] Other

## Description of change

```
{
  bidder: '<bidder name>',
  params: {
    // ...
  }
}
```


modules/priceFloors.js
```diff
@@ -291,6 +300,39 @@ export function updateAdUnitsForAuction(adUnits, floorData, auctionId) {
...
+        skipped: floorData.skipped,
+        skipRate: floorData.skipRate,
+        modelVersion: utils.deepAccess(floorData, 'data.modelVersion'),
+        location: utils.deepAccess(floorData, 'data.location', 'noData'),
+        floorProvider: floorData.floorProvider,
+        fetchStatus: _floorsConfig.fetchStatus
+      };
+
+      if (utils.deepAccess(floorData, 'floorMin')) {
+        bid.floorData.floorMin = utils.deepAccess(floorData, 'floorMin');
```
robertrmartinez, line 335: Think something went weird here during commit.

I do not see this function being called at all `updateAdUnitFloorData`

I think the changes needed were made in `updateAdUnitsForAuction` but somehow duplicated and function name changed.

So I think you can remove this entire selection.

modules/priceFloors.js
```diff
@@ -113,6 +113,14 @@ export function getFirstMatchingFloor(floorData, bidObject, responseObject = {})
...
     matchingRule
   };
+
+  // check that floorMin is greater than 0 and set values accordingly
+  if (utils.deepAccess(_floorsConfig, 'floorMin')) {
+    matchingData.floorMin = utils.deepAccess(_floorsConfig, 'floorMin');
+    matchingData.floorRuleValue = matchingData.matchingFloor;
+    if (matchingData.floorMin > matchingData.floorRuleValue) matchingData.matchingFloor = matchingData.floorMin;
+  }
+
```
robertrmartinez, line 123: So I think from now on we want to always set the `floorRuleValue` so that needs to be done outside of this if.

Also no need for deepAccess.

Use `floorData` not `_floorsConfig` as the min may change between auctions.

I think something like this would work:


Rewriting lines 111 and on:

```javascript
let matchingData = {
    matchingData: allPossibleMatches[0], // the first possible match is an "exact" so contains all data relevant for anlaytics adapters
    matchingRule,
    floorRuleValue: floorData.values[matchingRule] || floorData.default, // we know the Rule Value always so need it always
    floorMin: floorData.floorMin || 0 // if not defined default it to 0 (this is only for comparison, not signaled to adapters)
  };
  // now we can set the actual floor to be enforced by picking the max of the two
  matchingData.matchingFloor = Math.max(floorData.floorMin, matchingData.floorRuleValue);

```

package-lock.json
```diff
@@ -24486,4 +24486,4 @@
       }
     }
   }
-}
+}
```
robertrmartinez, line 24489: don't check in package-lock.

revert this please.

robertrmartinez (changes requested):
Couple changes.

Let me know if you have any questions.

modules/priceFloors.js
```diff
@@ -55,7 +55,7 @@ export let _floorDataForAuction = {};
  * @summary Simple function to round up to a certain decimal degree
  */
 function roundUp(number, precision) {
-  return Math.ceil(parseFloat(number) * Math.pow(10, precision)) / Math.pow(10, precision);
+  return Math.ceil((parseFloat(number) * Math.pow(10, precision)).toFixed(1)) / Math.pow(10, precision);
```
robertrmartinez, line 58: Can you explain how this solves the rounding issue?

robertrmartinez: Nevermind, took a look and it works. Good job!

mmoschovas: The rounding issue was due to the JS evaluation of the number after multiplying the two values. So 0.6271 multiplied by 10000 would be set as 6271.0000000000000001. Running the ceiling on this will cause it to round up to 6272. By fixing to only one decimal allowed it to round this only to the first decimal place prior to running the ceiling, thus avoiding rounding up unnecessarily.

modules/priceFloors.js
```diff
@@ -291,6 +300,39 @@ export function updateAdUnitsForAuction(adUnits, floorData, auctionId) {
         location: utils.deepAccess(floorData, 'data.location', 'noData'),
         floorProvider: floorData.floorProvider,
         fetchStatus: _floorsConfig.fetchStatus
+      };
+
+      if (utils.deepAccess(floorData, 'floorMin')) {
+        bid.floorData.floorMin = utils.deepAccess(floorData, 'floorMin');
```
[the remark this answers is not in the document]
mmoschovas: I can remove the deepAccess, but I left the if statement in because it cause an error passing the value as undefined in another analytics adapter when floorMin does not exist.

modules/priceFloors.js
```diff
@@ -287,11 +290,12 @@ export function updateAdUnitsForAuction(adUnits, floorData, auctionId) {
       bid.floorData = {
         skipped: floorData.skipped,
         skipRate: floorData.skipRate,
+        floorMin: floorData.floorMin,
```
robertrmartinez, line 293: @bszekely1 

Do we want to expose `floorMin` to bid adapters?

bszekely1: There is not much a bid adapter can do with this information. So long as analytics adapters have access to it, we should be covered. Bid adapters would still receive the floor, just the information feeding that floor should be available to analytics.

robertrmartinez: ok so if we leave this as is then a bid adapter will know at bidRequest time what the floorMin is.

As long as that is okay, we can and should leave it here.

mmoschovas: @robertrmartinez is anything else needed for this or is this good to go?

modules/priceFloors.js
```diff
@@ -190,6 +192,7 @@ export function getFloor(requestParams = {currency: 'USD', mediaType: '*', size:
   let bidRequest = this;
   let floorData = _floorDataForAuction[bidRequest.auctionId];
   if (!floorData || floorData.skipped) return {};
+  if (floorData.hasOwnProperty('floorMin')) floorData.data.floorMin = floorData.floorMin;
```
robertrmartinez, line 195: This is going to run each time getFloor is called. Which is probably not ideal.

Better to just set `floorData.data.floorMin` once.

I would recommend doing this in `createFloorsDataForAuction` which will initialize the entire data object, then you will just be able to access it where needed.

modules/priceFloors.js
```diff
@@ -568,6 +572,7 @@ function addFieldOverrides(overrides) {
  */
 export function handleSetFloorsConfig(config) {
   _floorsConfig = utils.pick(config, [
+    'floorMin', floorMin => floorMin,
```
robertrmartinez, line 575: With `utils.pick` you can just leave it directly as `floorMin` instead of with the function `floorMin => floorMin`

It will pick the floorMin on the config level, and since we do not need to default it, the extra use of function call is unneeded.

modules/rubiconAnalyticsAdapter.js
```diff
@@ -224,6 +225,8 @@ function sendMessage(auctionId, bidWonId) {
           'dealsEnforced', () => utils.deepAccess(auctionCache.floorData, 'enforcements.floorDeals'),
           'skipRate',
           'fetchStatus',
+          'floorMin',
+          'floorRuleValue',
```
robertrmartinez, line 229: `floorRuleValue` does not make sense to be in the top level `floorData` object of the message. It is a per `bidResponse`

I think this was a typo in the requirements haha.

robertrmartinez (changes requested):
looks good, just a couple small changes

robertrmartinez (approved):
Awesome looks good.

Will assign out to get second reviewer.

ncolletti:
Looks good.. We'll need a PR for docs though.

ncolletti (approved):

bszekely1:
Sorry forgot to submit the PR docs: https://github.com/prebid/prebid.github.io/pull/2438

modules/priceFloors.js
patmmccann: why round instead of down? is this a convention of sorts?

robertrmartinez: @patmmccann 

Say the actual floor is:

0.611

if we round to

0.61

and send that to the bidder, and they bid at 0.61, they will be floored.

So rounding up, makes sure we never run into this scenario.

patmmccann: interesting, in the case of not enforcing, seems like it should be flipped. However, probably quite the edge case, fyi @punkiller

PR #5816, opened 2020-10-01
update teadsBidAdapter

renebaudisch:
## Type of change
- [x] Bugfix

## Description of change
update teadsBidAdapter to return '0' instead of '' as fallback for getTimeToFirstByte method because it fails in unit tests
see #5796

### DIFF
--- a/modules/teadsBidAdapter.js
+++ b/modules/teadsBidAdapter.js
@@ -150,7 +150,7 @@ function getTimeToFirstByte(win) {
     performance.timing.requestStart >= 0 &&
     performance.timing.responseStart - performance.timing.requestStart;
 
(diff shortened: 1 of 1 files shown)


Viewtiful:
Hello @renebaudisch !
Thanks for the PR and the heads-up on a potential issue in our adapter.
After reviewing the issue and your PR, we specifically want to fallback to an empty string in case the performance APIs are not available or the time to first byte value is 0.
But we found an issue in our unit test that could cause to expect a value of 0 instead of the empty string, we have opened a PR to fix it: https://github.com/prebid/Prebid.js/pull/5820

Best

renebaudisch:
Then the other way round, sure, I don't know your code but I'm happy if it get's fixed! Thank you!

Issue #5762, opened 2020-09-16
Create generalized mock / stub for storageManager


robertrmartinez:
## Type of issue
Test Enhancements

## Description
When writing unit tests for modules which use storageManager, we need a better way to mock reading and setting cookies/local storage. As of now, lots of modules are doing it differently.

Relying on browsers in circle ci (and local browsers) to actually set and read back this data is not stable enough for unit tests.

So, we should create a mock which modules can import to their `_spec` files. Similar to how we mock XHR requests using `test/mocks/server.js`

See #5758 for the conversations and more details.

PR #5819, opened 2020-10-02
Force refresh userId

TLadd:
Continuation of https://github.com/prebid/Prebid.js/pull/5752

## Type of change
* [ ]  Bugfix
* [ x] Feature
* [ ]  New bidder adapter
* [ ]  Code style update (formatting, local variables)
* [ ]  Refactoring (no functional changes, no api changes)
* [ ]  Build related changes
* [ ]  CI related changes
* [ ]  Does this change affect user-facing APIs or examples documented on http://prebid.org?
* [ ]  Other

## Description of change
Added a global function to refresh user id's after initialization on page load. This seems to be the simplest way to get this functionality without adding more complexity into the user id module itself. Essentially, the function triggers a reinitialization of all user id modules, so if there were any changes to the stored values after page load, those new values will be included in the any bid requests.

In addition, I removed the mocking of storage from the zeotap test. I was running into test failures because the original PR branch was not based off of the current master. I ended up digging into the sporadically failing tests in Edge before it being pointed out that it had been fixed by mocking storage in https://github.com/prebid/Prebid.js/pull/5758. It looks like removing the expiration time also resolves the issue and IMO is preferable, so I left it as is. Can go back to the mocking situation if that's the direction y'all want to go.

* test parameters for validating bids

```
{
  bidder: '<bidder name>',
  params: {
    // ...
  }
}
```

* contact email of the adapter’s maintainer
* [ ]  official adapter submission

* A link to a PR on the docs repo at https://github.com/prebid/prebid.github.io/

## Other information
#5683


smenzer (approved):
LGTM

smenzer:
@robertrmartinez do you think you'll be able to do a 2nd review on this today, please?

Fawke (changes requested):
Hi @TLadd,

I see a merge conflict in `userId/index.js` file. I tried to resolve it locally (by keeping your code), but it didn't work out. The command `gulp test` threw a few errors after that. Can you please resolve the merge conflict and make sure all the unit test cases pass?

Also, I don't see a point in changing zeotap unit test file if the existing solution is working fine for us.

TLadd:
@Fawke Merged in master and opened a docs PR https://github.com/prebid/prebid.github.io/pull/2402

Re the zeotap test changes: So the history here is that the zeotap tests were modified to mock cookie storage in https://github.com/prebid/Prebid.js/pull/5758 because the tests were sporadically (but fairly regularly) failing in the Edge browsers. Prior to merging in #5758 to this branch, I dug into what was going on in these tests and discovered that the failures were related to the zeotap tests setting an expiration on the cookie. I was logging document.cookie at various points in the test and saw the cookie of interest disappearing at inconsistent points. None of the other userId submodules do this and removing the expiration seems to have fixed the flakey tests (they haven't failed since). 

So then the question becomes if we want to continue mocking cookies in these tests when we have another fix available. Expressed my thoughts [here](https://github.com/prebid/Prebid.js/pull/5752#issuecomment-702708281); the tldr is that I think mocking cookies is asking for trouble and diminishes the value of the tests. This restores the zeotap tests to being consistent with the other userId submodule tests but does run in direct opposition to https://github.com/prebid/Prebid.js/issues/5762.

Fawke (approved):
LGTM! I'd like @robertrmartinez to have say on the mocking of storageManager calls on zeoTap and issue he created, https://github.com/prebid/Prebid.js/issues/5762, if we still need it or not!

Also, adding @bretg.

bretg (approved):
I didn't look at the code. The concept and documentation is fine. Thanks for the contribution.

smenzer:
since tests are passing and this reverts to the original zeotap test, I'm taking silence as approval from @robertrmartinez regarding the mocking of storageManager (or not). if changes are needed, they can be addressed as part of #5762 more broadly than just the zeotap test.

PR #5820, opened 2020-10-02
Fix timeToFirstByte unit test

Viewtiful:
## Type of change

- [x] Bugfix

## Description of change
In rare cases, the time to first byte unit test could end up with a value of 0 which in the Teads adapter code is not possible as we handle this case specifically by discarding the value and fallbacking on the empty string.
This change has for goal to add this fallback logic on the unit test in order to fix it.

### DIFF
--- a/test/spec/modules/teadsBidAdapter_spec.js
+++ b/test/spec/modules/teadsBidAdapter_spec.js
@@ -200,8 +200,8 @@ describe('teadsBidAdapter', () => {
         performance.getEntriesByType('navigation')[0] &&
         performance.getEntriesByType('navigation')[0].responseStart &&
         performance.getEntriesByType('navigation')[0].requestStart &&
-        performance.getEntriesByType('navigation')[0].responseStart >= 0 &&
-        performance.getEntriesByType('navigation')[0].requestStart >= 0 &&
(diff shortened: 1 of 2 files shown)


nailyk (approved):
lgtm !

test/spec/modules/teadsBidAdapter_spec.js
```diff
@@ -211,9 +211,15 @@ describe('teadsBidAdapter', () => {
...
         expect(payload.timeToFirstByte).to.deep.equal(ttfbExpectedV2.toString());
       } else {
-        const ttfbExpectedV1 = performance.timing.responseStart - performance.timing.requestStart;
-
-        expect(payload.timeToFirstByte).to.deep.equal(ttfbExpectedV1.toString());
+        const ttfbWithTimingV1 = performance &&
+            performance.timing.responseStart &&
+            performance.timing.requestStart &&
+            performance.timing.responseStart >= 0 &&
+            performance.timing.requestStart >= 0 &&
```
harpere, line 218: these can't be zero because of the lines above.  Not a big deal but could be `>` rather than `>=`.

Viewtiful: You are right, updated it test side and adapter side. Thank you for the review

harpere (approved):

PR #5821, opened 2020-10-04
Appnexus: Add omid support

Fawke:
## Type of change
- [X] Feature

## Description of change

Adds OMID support the Appnexus adapter.

Link to the Docs PR: https://github.com/prebid/prebid.github.io/pull/2394

### DIFF
--- a/modules/appnexusBidAdapter.js
+++ b/modules/appnexusBidAdapter.js
@@ -157,6 +157,7 @@ export const spec = {
     const memberIdBid = find(bidRequests, hasMemberId);
     const member = memberIdBid ? parseInt(memberIdBid.params.member, 10) : 0;
     const schain = bidRequests[0].schain;
+    const omidSupport = find(bidRequests, hasOmidSupport);
 
     const payload = {
       tags: [...tags],
@@ -167,6 +168,13 @@ export const spec = {
       },
       schain: schain
     };
+
+    if (omidSupport) {
+      payload['iab_support'] = {
+        omidpn: 'Appnexus',
+        omidpv: '$prebid.version$'
+      }
+    }
 
     if (member > 0) {
       payload.member_id = member;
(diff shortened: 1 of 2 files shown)


jsnellbaker (changes requested):
I think this overall looks good.

Just to confirm:
- if there were multiple video frameworks listed in the `bid.params`, all those values should be set within the `tags[].video_frameworks` property in the request?  Based on the unit tests, this seems to be the case - but wanted to clarify.
- in the case of `bid.params.video.frameworks`, it seems we're now going to pass the frameworks values twice in the request payload.  Once under `tags[].video_frameworks` and again under `tags[].video.frameworks` (this is the original manner).  Is this intended?

Thanks.

Fawke:
Hi @jsnellbaker,

>if there were multiple video frameworks listed in the bid.params, all those values should be set within the tags[].video_frameworks property in the request? Based on the unit tests, this seems to be the case - but wanted to clarify.

Yes, you are right. Whater is present in `bid.params.video.frameworks` ends up in `tags[].video_frameworks`.

>in the case of bid.params.video.frameworks, it seems we're now going to pass the frameworks values twice in the request payload. Once under tags[].video_frameworks and again under tags[].video.frameworks (this is the original manner). Is this intended?

You are right to point out the redundancy, currently we are passing the value of video frameworks in two place in the request payload. Originally it used to be, `tags[].video.frameworks`, now additionally, we're passing it under `tags[].video_frameworks`. Probably, @jaiminpanchal27 can clarify here if we need to maintain this redundancy.

Also, just wanted to point out another quirk of this implementation, though, this may not be a big deal.

"The frameworks array has to be in integer array, it can't be an array of strings" If publishers do this, `['1', '2', '6']`, our code would crash."

Also, what I'm guessing is, `bids.params.frameworks` especially lists only the "Banner" frameworks and the video, at least that's what I wrote in the docs PR. Am I corrent in my assumption?

jsnellbaker:
For the latter points, I think those are fine.

jsnellbaker (approved):
LGTM; thanks for making the updates.

PR #5825, opened 2020-10-05
Adtelligent adapter update

GeneGenie:
## Type of change
- [x] Feature

## Description of change
Submit new alias, which also supports custom wl host 

- g.moroz@adtelligent.com


musikele (approved):
Just to let you know, when using `appaloosa` alias i don't receive any test bid. Anyway, LGTM on the code side

PR #5834, opened 2020-10-07
Added the ability to send multiple bids in one ad request for mediaforce bid adapter

Niksok:
## Type of change

- [x] Feature

## Description of change

Added the ability to send multiple bids in one ad request
Added additional fields tmax, ext.mediaforce.hb_key and imp[].ext.transactionId

```
{
  bidder: '<bidder name>',
  params: {
    // ...
  }
}
```

- contact email of the adapter’s maintainer: contact@mediaforce.com


modules/mediaforceBidAdapter.js
```diff
@@ -140,31 +150,47 @@ export const spec = {
...
+          data: request
+        });
+      }
+      validImp && request.imp.push(impObj);
+    });
+    requests.forEach((req) => {
+      if (isTest) {
+        req.url = TEST_ENDPOINT_URL;
+      }
+      req.data = JSON.stringify(req.data);
```
ncolletti, line 193: Any reason for the separate loop, couldn't this be applied when building the request above?

modules/mediaforceBidAdapter.js
```diff
@@ -140,31 +150,47 @@ export const spec = {
...
 
-      let request = {
-        id: bid.transactionId,
-        site: {
-          page: referer,
-          ref: referer,
-          id: bid.params.publisher_id,
-          publisher: {
-            id: bid.params.publisher_id
+      let request = requestsMap[bid.params.publisher_id];
```
ncolletti, line 154: Curious why you are keying against the publisher_id and sending multiple requests if the goal of the ticket is to allow for multiple bids in one request? It does not seem to change the endpoint url or any other attributes of the request.

Couldn't you organize your imp object by `params.publisher_id` and send that in one request?

modules/mediaforceBidAdapter.js
```diff
@@ -140,31 +150,47 @@ export const spec = {
...
+      if (!request) {
+        request = {
+          id: Math.round(Math.random() * 1e16).toString(16),
+          site: {
+            page: referer,
```
ncolletti, line 159: this new page param value is duplicated by your `ref`, shouldn't `page` be attributed to the window.location.href and the other to the refererInfo property in bidderRequest?

modules/mediaforceBidAdapter.js
```diff
@@ -114,18 +114,28 @@ export const spec = {
...
     validBidRequests.forEach(bid => {
+      isTest = isTest || bid.params.is_test;
       let tagid = bid.params.placement_id;
       let bidfloor = bid.params.bidfloor ? parseFloat(bid.params.bidfloor) : 0;
-      let imp = [];
       let validImp = false;
       let impObj = {
         id: bid.bidId,
         tagid: tagid,
         secure: 1,
```
ncolletti, line 131: Not new code but do you want to be checking `location.protocol` instead of hardcoding this?

modules/mediaforceBidAdapter.js
Niksok: The parameter "isTest" can be configured in any bid in the middle of list, and therefore we need make second loop after all bid was processed. And some bids in theory can have different publisher_id in params, and because the publisher_id defines site field, which is common to all imps, we must prepare separate request for each publisher_id.

modules/mediaforceBidAdapter.js
Niksok: As i wrote upper, some bids in theory can have different publisher_id in params, and because the publisher_id defines site field, which is common to all imps, we must prepare separate request for each publisher_id.

modules/mediaforceBidAdapter.js
Niksok: Fixed

modules/mediaforceBidAdapter.js
Niksok: Fixed

modules/mediaforceBidAdapter.js
ncolletti: Thanks, I suppose you expect to have different publisher_id values with different site values from the same publisher/site. I don't see why that would occur but I'm not against your adapter supporting it.

For isTest, wouldn't you want this to be configured on a per request basis? For example, publisher_id: 203 want's test debug request url, but publisher_id: 204 in a different request does not. In this configuration, publisher_id: 204 would use the test endpoint. Again, fine with this but would like to confirm.

modules/mediaforceBidAdapter.js
ncolletti: Could you explain a situation where one Prebid auction can have multiple sites or publishers?

modules/mediaforceBidAdapter.js
Niksok: isTest parameter used in the main in case to test new changes before commit it. And therefore i would to leave it as is.

modules/mediaforceBidAdapter.js
Niksok: I think not, but just in case, I would like to leave this opportunity

ncolletti (approved):

PR #5841, opened 2020-10-09
Fix size validate

r-ishigami:
## Type of change

- [x] Bugfix
- [ ] Feature
- [ ] New bidder adapter
- [ ] Build related changes
- [ ] CI related changes
- [ ] Other

## Description of change

Fixed size validation.
When the media type is banner, the valid size is 300x250 or more or 1x1.


modules/relaidoBidAdapter.js
```diff
@@ -68,7 +68,7 @@ function buildRequests(validBidRequests, bidderRequest) {
       payload.width = playerSize[0][0];
       payload.height = playerSize[0][1];
     } else if (hasBannerMediaType(bidRequest)) {
-      const sizes = utils.deepAccess(bidRequest, 'mediaTypes.banner.sizes');
+      const sizes = getValidSizes(utils.deepAccess(bidRequest, 'mediaTypes.banner.sizes'));
       payload.width = sizes[0][0];
```
aleksatr, line 72: @r-ishigami you didn't check if sizes.length > 0, it may throw index out of bounds exception.

r-ishigami: Hi @aleksatr , thanks for the review.
I think it is not necessary to check with buildRequests because it has been checked with isBidRequestValid.
https://github.com/relaido/Prebid.js/blob/fix-size-validate/modules/relaidoBidAdapter.js#L237-L240

aleksatr:
LGTM

PR #5850, opened 2020-10-12
Add sharedid support to pubcommon

pycnvr:
## Type of change

- [ ] Bugfix
- [x] Feature
- [ ] New bidder adapter   
- [ ] Build related changes
- [ ] CI related changes
- [ ] Other

## Description of change
Integrate sharedid into pubcommon

- pyang@conversantmedia.com

- A link to a PR on the docs repo at https://github.com/prebid/prebid.github.io/pull/2417

## Other information

Edit: Added doc PR

### DIFF
--- a/modules/pubCommonIdSystem.js
+++ b/modules/pubCommonIdSystem.js
@@ -18,6 +18,7 @@ const LOCAL_STORAGE = 'html5';
 const SHAREDID_OPT_OUT_VALUE = '00000000000000000000000000';
 const SHAREDID_URL = 'https://id.sharedid.org/id';
 const SHAREDID_SUFFIX = '_sharedid';
+const EXPIRED_COOKIE_DATE = 'Thu, 01 Jan 1970 00:00:01 GMT';
 
 const storage = getStorageManager(null, 'pubCommonId');
 
@@ -60,6 +61,21 @@ function readData(config) {
 }
 
 /**
+ * Delete sharedid from cookie or local storage
+ * @param config Need config.storage to derive key and storage type
+ */
+function delData(config) {
+  const key = config.storage.name + SHAREDID_SUFFIX;
+  if (config.storage.type === COOKIE) {
+    if (storage.cookiesAreEnabled()) {
+      storage.setCookie(key, '', EXPIRED_COOKIE_DATE);
+    }
+  } else if (config.storage.type === LOCAL_STORAGE) {
+    storage.removeDataFromLocalStorage(key);
+  }
+}
+
+/**
  * setup success and error handler for sharedid callback thru ajax
  * @param {string} pubcid Current pubcommon id
  * @param {function} callback userId module callback.
@@ -74,9 +90,14 @@ function handleResponse(pubcid, callback, config) {
         try {
           let responseObj = JSON.parse(responseBody);
           utils.logInfo('SharedId: Generated SharedId: ' + responseObj.sharedId);
-          if (responseObj.sharedId && responseObj.sharedId !== SHAREDID_OPT_OUT_VALUE) {
-            // Store sharedId locally
-            storeData(config, responseObj.sharedId);
+          if (responseObj.sharedId) {
+            if (responseObj.sharedId !== SHAREDID_OPT_OUT_VALUE) {
+              // Store sharedId locally
+              storeData(config, responseObj.sharedId);
+            } else {
+              // Delete local copy if the user has opted out
+              delData(config);
+            }
           }
           // Pass pubcid even though there is no change in order to trigger decode
           callback(pubcid);
@@ -152,21 +173,26 @@ export const pubCommonIdSubmodule = {
   /**
    * performs action to obtain id
    * @function
-   * @param {SubmoduleConfig} [config]
+   * @param {SubmoduleConfig} [config] Config object with params and storage properties
+   * @param {Object} consentData
+   * @param {string} storedId Existing pubcommon id
    * @returns {IdResponse}
    */
-  getId: function (config = {}) {
+  getId: function (config = {}, consentData, storedId) {
     const {params: {create = true, pixelUrl} = {}} = config;
-    let newId;
-    try {
-      if (typeof window[PUB_COMMON_ID] === 'object') {
-        // If the page includes its own pubcid module, then save a copy of id.
-        newId = window[PUB_COMMON_ID].getId();
+    let newId = storedId;
+    if (!newId) {
+      try {
+        if (typeof window[PUB_COMMON_ID] === 'object') {
+          // If the page includes its own pubcid module, then save a copy of id.
+          newId = window[PUB_COMMON_ID].getId();
+        }
+      } catch (e) {
       }
-    } catch (e) {
+
+      if (!newId) newId = (create && utils.hasDeviceAccess()) ? utils.generateUUID() : undefined;
     }
 
-    if (!newId) newId = (create && utils.hasDeviceAccess()) ? utils.generateUUID() : undefined;
     return {
       id: newId,
       callback: getIdCallback(newId, this.makeCallback(pixelUrl, newId), config)
@@ -198,7 +224,7 @@ export const pubCommonIdSubmodule = {
       try {
         if (typeof window[PUB_COMMON_ID] === 'object') {
           // If the page includes its own pubcid module, then there is nothing to do
-          // except to update sharedid
+          // except to update sharedid's expiration time
           storeData(config, readData(config));
           return;
         }
@@ -209,6 +235,7 @@ export const pubCommonIdSubmodule = {
         const callback = this.makeCallback(pixelUrl, storedId);
         return {callback: callback};
       } else {
+        // Update with the same value to extend expiration time
         storeData(config, readData(config));
         return {id: storedId};
       }
--- a/modules/pubCommonIdSystem.js
+++ b/modules/pubCommonIdSystem.js
@@ -23,7 +23,7 @@ const storage = getStorageManager(null, 'pubCommonId');
 
 /**
  * Store sharedid in either cookie or local storage
- * @param {Object} config Need config.storage object to derive key, expiry time, and storay type.
+ * @param {Object} config Need config.storage object to derive key, expiry time, and storage type.
  * @param {string} value Shareid value to store
  */
 
@@ -78,7 +78,7 @@ function handleResponse(pubcid, callback, config) {
             // Store sharedId locally
             storeData(config, responseObj.sharedId);
           }
-          // Pass pubcid even thought there is no change in order to trigger decode
+          // Pass pubcid even though there is no change in order to trigger decode
           callback(pubcid);
         } catch (error) {
           utils.logError(error);
@@ -177,7 +177,7 @@ export const pubCommonIdSubmodule = {
    * of stored id: using pixelUrl or return the id and let main user id module write it again with
    * the new expiration time.
    *
-   * PixelUrl, f defined, should point back to a first party domain endpoint.  On the server
+   * PixelUrl, if defined, should point back to a first party domain endpoint.  On the server
    * side, there is either a plugin, or customized logic to read and write back the pubcid cookie.
    * The extendId function itself should return only the callback, and not the id itself to avoid
    * having the script-side overwriting server-side.  This applies to both pubcid and sharedid.
@@ -197,7 +197,7 @@ export const pubCommonIdSubmodule = {
     if (extend) {
       try {
         if (typeof window[PUB_COMMON_ID] === 'object') {
-          // If the page includes its onw pubcid module, then there is nothing to do
+          // If the page includes its own pubcid module, then there is nothing to do
           // except to update sharedid
           storeData(config, readData(config));
           return;
--- a/modules/pubCommonIdSystem.js
+++ b/modules/pubCommonIdSystem.js
@@ -8,12 +8,105 @@
 import * as utils from '../src/utils.js';
 import {submodule} from '../src/hook.js';
 import {getStorageManager} from '../src/storageManager.js';
+import {ajax} from '../src/ajax.js';
 
 const PUB_COMMON_ID = 'PublisherCommonId';
-
 const MODULE_NAME = 'pubCommonId';
 
+const COOKIE = 'cookie';
+const LOCAL_STORAGE = 'html5';
+const SHAREDID_OPT_OUT_VALUE = '00000000000000000000000000';
+const SHAREDID_URL = 'https://id.sharedid.org/id';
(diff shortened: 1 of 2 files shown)


modules/pubCommonIdSystem.js
```diff
@@ -46,52 +140,78 @@ export const pubCommonIdSubmodule = {
...
    * @param {string} value
+   * @param {SubmoduleConfig} config
    * @returns {{pubcid:string}}
    */
-  decode(value) {
-    return { 'pubcid': value }
+  decode(value, config) {
+    const sharedId = readData(config);
+    const idObj = {'pubcid': value};
+    if (sharedId) idObj['sharedid'] = {id: sharedId};
```
smenzer, line 149: very interesting...didn't realize this would "just work", cool.

modules/pubCommonIdSystem.js
```diff
@@ -46,52 +140,78 @@ export const pubCommonIdSubmodule = {
...
   },
   /**
    * performs action to obtain id
    * @function
    * @param {SubmoduleConfig} [config]
    * @returns {IdResponse}
    */
-  getId: function ({params: {create = true, pixelUrl} = {}} = {}) {
+  getId: function (config = {}) {
+    const {params: {create = true, pixelUrl} = {}} = config;
```
smenzer, line 159: where does this config get set, by the publisher? i don't see any docs about setting `create` or `pixelUrl` here https://docs.prebid.org/dev-docs/modules/userId.html#pubcommon-id ... should those be added?

modules/pubCommonIdSystem.js
```diff
@@ -46,52 +140,78 @@ export const pubCommonIdSubmodule = {
...
-  extendId: function({params: {extend = false, pixelUrl} = {}} = {}, storedId) {
-    try {
-      if (typeof window[PUB_COMMON_ID] === 'object') {
-        // If the page includes its onw pubcid module, then there is nothing to do.
-        return;
-      }
-    } catch (e) {
-    }
+  extendId: function(config = {}, storedId) {
+    const {params: {extend = false, pixelUrl} = {}} = config;
```
smenzer, line 195: same question here about config... i don't see any docs about setting `extend` or `pixelUrl` here https://docs.prebid.org/dev-docs/modules/userId.html#pubcommon-id ... should those be added?

smenzer (changes requested):
A few typos in comments, plus this needs to be well documented so publishers are aware that an install of pubcommon means they are also getting sharedid, which now will make http requests to receive an id, which pubcommon historically didn't _always_ do (AFAIK), as well as the other create/extend/pixelUrl config params that are not currently documented.

pycnvr:
Thanks @smenzer for the quick review!! I'll make these change and supply a doc PR, too.  

Since there are no further change from the Magnite team, changing the destination to master.

test/spec/modules/userId_spec.js
```diff
@@ -1511,6 +1517,79 @@ describe('User ID', function () {
...
+    it('test pubcid coexisting with sharedid', function (done) {
+      coreStorage.setCookie('pubcid', 'testpubcid', (new Date(Date.now() + 5000).toUTCString()));
+      coreStorage.setCookie('pubcid_sharedid', 'test111', (new Date(Date.now() + 5000).toUTCString()));
+      coreStorage.setCookie('sharedid', JSON.stringify({
+        'id': 'test222',
+        'ts': 1590525289611
+      }), (new Date(Date.now() + 5000).toUTCString()));
+
+      // When both pubcommon and sharedid are configured, pubcommon are invoked first due to
+      // module loading order.  This mean the output from the primary sharedid module will overwrite
```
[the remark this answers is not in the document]
pycnvr: Appears to be alphabetical, and different from the order defined in .submodules.json.  Yeah this bothers me as well.  The two modules receives the same sharedid value from the same endpoint, so even when the order was swapped, the result is pretty much the same.  The only difference is that the main sharedid module adds an extra property called 'third'.  Maybe @jdwieland8282 can comment on whether this is a big deal or not. 

But to be on the safe side, I am thinking of adding a config option to disable sharedid in pubcid module.

modules/pubCommonIdSystem.js
pycnvr: Yup, didn't realize it works this way, either.

test/spec/modules/userId_spec.js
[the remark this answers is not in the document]
jdwieland8282: @pycnvr this is not a big deal, the main sharedid module sets two ids one in the 1p context (ext.eids.uids.id) and one in the 3p context (ext.eids.uids.ext.third).  The latter is the response from sharedid.org/usyc, which can go in the PC module in the more traditional eids blob under sharedid.org

```
{
   "eids":[
      {
         "source":"sharedid.org",
         "uids":[
            {
               "id":"<response from sharedid.org/usyc",
               "atype":1
            }
         ]
      }
   ]
}
```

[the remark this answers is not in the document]
pycnvr: Hi @jdwieland8282, not sure I understand.  😞 Should we add a second ajax call to sharedid.org/usyc?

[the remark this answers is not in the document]
jdwieland8282: No, I don't think so, I'm just trying to illustrate where the sharedid.org uid should go in the pubcommon module.

modules/pubCommonIdSystem.js
```diff
@@ -46,52 +140,78 @@ export const pubCommonIdSubmodule = {
...
+          return;
+        }
+      } catch (e) {
+      }
+
+      if (pixelUrl) {
+        const callback = this.makeCallback(pixelUrl, storedId);
+        return {callback: callback};
+      } else {
+        storeData(config, readData(config));
```
SKOCHERI, line 212: Is the intention of storeData(config, readData(config)); is to extent the expiry of the cookie? I ask this as I see the key  (config.storage.name + SHAREDID_SUFFIX ) same in readData and storeData.

modules/pubCommonIdSystem.js
```diff
@@ -46,52 +140,78 @@ export const pubCommonIdSubmodule = {
...
+          // If the page includes its own pubcid module, then there is nothing to do
+          // except to update sharedid
+          storeData(config, readData(config));
+          return;
+        }
+      } catch (e) {
+      }
+
+      if (pixelUrl) {
+        const callback = this.makeCallback(pixelUrl, storedId);
```
SKOCHERI, line 209: Should this be similar to callback in getId()? The getId() has callback: getIdCallback(newId, this.makeCallback(pixelUrl, newId), config) where the shared id (
_sharedid) cookie is stored if shared id is not opted out.
Based on the shared id opt out, should the cookie(_sharedid) be reset/removed in extendId() function?

jdwieland8282 (approved):
looks fine

modules/pubCommonIdSystem.js
pycnvr: This keeps the stored value of pubcid as-is, and stores sharedid separately.  Storage of pubcid still goes thru userid modules, while this module now manages its own copy of sharedid.

modules/pubCommonIdSystem.js
pycnvr: You have good point.  What it means is that shareid.org should still be called in extendId(), and if it returns an opt-out, then erase the stored value.

pycnvr: @SKOCHERI After checking the flow a bit, extendId is going to be called at every page load, so it seems a bit heavy handed. There is an actually a superior mechanism with refreshInSeconds that can be used to check the validity of the stored sharedid.  That would drive getid() periodically, then we can modify the existing callback to delete the cookie when opted out.

modules/pubCommonIdSystem.js
pycnvr: I missed the question completely, didn't I.  Yes, it's just to move the expiry time back.

modules/pubCommonIdSystem.js
SKOCHERI: Yeah. Basically the cookie needs to be cleared on shared id opt out. refreshInSeconds would also work.

pycnvr: Thanks.  Good catch.

smenzer:
I'd like to review again tomorrow after reading all the updates posted tonight

modules/pubCommonIdSystem.js
pycnvr: Please see https://github.com/prebid/prebid.github.io/pull/2417

modules/pubCommonIdSystem.js
```diff
@@ -8,20 +8,135 @@
...
+  if (value) {
+    const key = config.storage.name + SHAREDID_SUFFIX;
+    if (config.storage.type === COOKIE) {
+      if (storage.cookiesAreEnabled()) {
+        const expiresStr = (new Date(Date.now() + (storage.expires * (60 * 60 * 24 * 1000)))).toUTCString();
+        storage.setCookie(key, value, expiresStr, 'LAX', COOKIE_DOMAIN);
+      }
+    } else if (config.storage.type === LOCAL_STORAGE) {
+      if (storage.hasLocalStorage()) {
+        storage.setDataInLocalStorage(key, value);
```
smenzer, line 41: You should set a separate item in local storage with an expiration so you can check it when reading to make sure it hasn't expired.

modules/pubCommonIdSystem.js
```diff
@@ -8,20 +8,135 @@
...
+ * @param config Need config.storage to derive key and storage type
+ * @return {string}
+ */
+function readData(config) {
+  const key = config.storage.name + SHAREDID_SUFFIX;
+  if (config.storage.type === COOKIE) {
+    if (storage.cookiesAreEnabled()) { return storage.getCookie(key); }
+  } else if (config.storage.type === LOCAL_STORAGE) {
+    if (storage.hasLocalStorage()) {
+      return storage.getDataFromLocalStorage(key);
```
smenzer, line 58: you should add a check for the item being expired before using it.

modules/pubCommonIdSystem.js
```diff
@@ -8,20 +8,135 @@
...
+ * @return {{success: success, error: error}}
+ */
+
+function handleResponse(pubcid, callback, config) {
+  return {
+    success: function (responseBody) {
+      if (responseBody) {
+        try {
+          let responseObj = JSON.parse(responseBody);
+          utils.logInfo('SharedId: Generated SharedId: ' + responseObj.sharedId);
```
smenzer, line 92: Should this be more like this, so that it's clear this message is coming from the pubcommon module and not shared id?

```suggestion
          utils.logInfo('PubCommonId: Generated SharedId: ' + responseObj.sharedId);
```

modules/pubCommonIdSystem.js
```diff
@@ -8,20 +8,135 @@
...
+          }
+          // Pass pubcid even though there is no change in order to trigger decode
+          callback(pubcid);
+        } catch (error) {
+          utils.logError(error);
+        }
+      }
+    },
+    error: function (statusText, responseBody) {
+      utils.logInfo('SharedId: failed to get id');
```
smenzer, line 110: This could be more clear about where it's coming from:

```suggestion
      utils.logInfo('PubCommonId: failed to get SharedId');
```

test/spec/modules/userId_spec.js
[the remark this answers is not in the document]
smenzer: > But to be on the safe side, I am thinking of adding a config option to disable sharedid in pubcid module. Edit: A separate PR probably.

I think this should definitely be included in this PR...I'd actually argue that the default behavior should be NOT to include sharedid and only via a config should it be added via pubcommon. Pubs may not have the right privacy policies or may have other issues with adding an identity module without their knowledge or direct setup. These changes are adding net new requests to systems that the publisher previously was not aware of.

smenzer (changes requested):
i think this needs a flag to enable the sharedid behavior rather than forcing every publisher to call sharedid without their explicit approval

test/spec/modules/userId_spec.js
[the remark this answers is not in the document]
jdwieland8282: I strongly disagree @smenzer Prebid has its own Privacy policy and pubs can turn this off if they desire.  The purpose of merging the two modules is so that the Prebid community can have access to a fully seeded supported open source identity, that won't happen if the module is off by default.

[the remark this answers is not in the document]
jdwieland8282: http://docs.prebid.org/privacy.html

pycnvr:
@smenzer Agains thanks for the detailed review.  I heard that you guys have reached an agreement today.  A new option called params.enableSharedId has been added, and the default is false, or OFF.

pycnvr:
Code and doc updates have been pushed.  Thanks.

jlukas79 (approved):

SKOCHERI (approved):

smenzer (changes requested):
> @smenzer Agains thanks for the detailed review. I heard that you guys have reached an agreement today. A new option called params.enableSharedId has been added, and the default is false, or OFF.

yes, the PBJS committee made the decision that this should be off by default until a major release when it can be flipped.

changes look good, only thing I'd ask for at this point is a test where sharedid is off and just make sure it's not called. once that's done then I'm good here and we can merge it.

pycnvr:
@smenzer Yup already got those covered.  Please take a look at userId_spec.js in the same commit.

test/spec/modules/userId_spec.js
```diff
@@ -1511,6 +1551,192 @@ describe('User ID', function () {
...
+          unit.bids.forEach(bid => {
+            // verify that the PubCommonId id data was copied to bid
+            expect(bid).to.have.deep.nested.property('userId.pubcid');
+            expect(bid.userId.pubcid).to.equal('testpubcid');
+            expect(findEid(bid.userIdAsEids, 'pubcid.org')).to.deep.equal(
+              {'source': 'pubcid.org', 'uids': [{'id': 'testpubcid', 'atype': 1}]}
+            );
+            // verify that the sharedid was not added to bid
+            expect(bid.userId).to.not.have.property('sharedid');
+            expect(findEid(bid.userIdAsEids, 'sharedid.org')).to.be.undefined;
```
pycnvr, line 1610: Here's one of the testcases that verifies no shared id.

smenzer:
> @smenzer Yup already got those covered. Please take a look at userId_spec.js in the same commit.

ah you're right, i missed those, sorry.  all good then, i'll go ahead and merge.

smenzer (approved):
LGTM

pycnvr:
Thanks!

PR #5859, opened 2020-10-15
Liveintent id module doesn't fall back to the default implementations of ajax, pixel and storage.

melgenek:
## Type of change

- [X] Feature

## Description of change
1) The live-connect library is updated to use only the ajax, storage and pixel implementations that are provided https://github.com/LiveIntent/live-connect/releases/tag/v1.1.19. There is no fall back to a default implementation any more.
2) Liveintent id module accepts an `emailHash` param.


smenzer (commented):
Everything seems to LGTM, but as this is a bit beyond my comfort level, I'd like another review from @FilipStamenkovic before we merge this.

FilipStamenkovic (commented):
LGTM either.  
I tested it on my local, works fine.  
But, according to our PR guidelines one PR should be about single item, this PR has many different items:
- upgrade npm package
- refactor liveIntent code and tests
- add new module parameter `emailHash`
- add support for GDPR (as I see this module is not supporting GDPR on master branch)



@smenzer What is your opinion on this?

smenzer:
Since these are all related to the liveintent id module, I'm ok merging this in a single PR. Thanks for the second review @FilipStamenkovic!

PR #5863, opened 2020-10-16
ID Library

SKOCHERI:
## Type of change

- [ ] Bugfix
- [x] Feature
- [ ] New bidder adapter   
- [ ] Build related changes
- [ ] CI related changes
- [ ] Other

## Description of change

```
{
  bidder: '<bidder name>',
  params: {
    // ...
  }
}
```

- contact email of the adapter’s maintainer eng-dmp@magnite.com

### DIFF
--- a/modules/idLibrary.js
+++ b/modules/idLibrary.js
@@ -36,12 +36,8 @@ function getEmail(value) {
   if (!matched) {
     return null;
   }
+  logInfo('Email found' + matched[0]);
   return matched[0];
-}
-
-function hasEmail(value) {
-  const email = getEmail(value);
-  return !!email;
 }
 
 function bodyAction(conf, mutations, observer) {
@@ -53,10 +49,8 @@ function bodyAction(conf, mutations, observer) {
   }
 
   const body = document.body.innerHTML;
-
-  if (hasEmail(body)) {
-    email = getEmail(body);
-
+  email = getEmail(body);
+  if (email != null) {
     logInfo(`Email obtained from the body ${email}`);
     observer.disconnect();
     logInfo('Post data on email found in body');
@@ -104,9 +98,8 @@ function removeInputElementsElementListner(conf) {
 function processInputChange(event, conf) {
   const value = event.target.value;
   logInfo(`Modified Value of input ${event.target.value}`);
-  if (hasEmail(value)) {
-    email = getEmail(value);
-
+  email = getEmail(value);
+  if (email != null) {
     logInfo('Email found in input ' + email);
     postData(conf.url);
     removeInputElementsElementListner(conf);
@@ -161,9 +154,8 @@ function handleBodyElements(conf) {
     postData(conf.url);
     return;
   }
-  if (hasEmail(document.body.innerHTML)) {
-    email = getEmail(document.body.innerHTML);
-
+  email = getEmail(document.body.innerHTML);
+  if (email != null) {
     logInfo('Email found in body ' + email);
     logInfo(' Post data on email found in the body without observer');
     postData(conf.url);
@@ -182,9 +174,8 @@ function doesInputElementsHaveEmail() {
 
   for (let index = 0; index < inputs.length; ++index) {
     const curInput = inputs[index];
-
-    if (hasEmail(curInput.value)) {
-      email = getEmail(curInput.value);
+    email = getEmail(curInput.value);
+    if (email != null) {
       return true;
     }
   }
--- a/modules/idLibrary.js
+++ b/modules/idLibrary.js
@@ -70,7 +70,7 @@ function targetAction(mutations, observer) {
       if (email) {
         logInfo('Email obtained from the target ' + email);
         observer.disconnect();
-        logInfo(' Post data on email found in target');
+        logInfo('Post data on email found in target');
         postData();
         return;
       }
@@ -83,7 +83,7 @@ function addInputElementsElementListner(conf) {
   const inputs = document.querySelectorAll('input[type=text], input[type=email]');
 
   for (var i = 0; i < inputs.length; i++) {
-    logInfo(` Original Value in Input = ${inputs[i].value}`);
+    logInfo(`Original Value in Input = ${inputs[i].value}`);
     inputs[i].addEventListener('change', event => processInputChange(event));
     inputs[i].addEventListener('blur', event => processInputChange(event));
   }
@@ -141,8 +141,8 @@ function handleTargetElement() {
       logInfo('Finding the email with observer');
       targetObserver.observe(targetElement, OBSERVER_CONFIG);
     } else {
-      logInfo(' Target found with target ' + email);
-      logInfo(' Post data on email found in target with target');
+      logInfo('Target found with target ' + email);
+      logInfo('Post data on email found in target with target');
       postData();
     }
   }
@@ -158,7 +158,7 @@ function handleBodyElements() {
   email = getEmail(document.body.innerHTML);
   if (email !== null) {
     logInfo('Email found in body ' + email);
-    logInfo(' Post data on email found in the body without observer');
+    logInfo('Post data on email found in the body without observer');
     postData();
     return;
   }
@@ -185,10 +185,10 @@ function doesInputElementsHaveEmail() {
 function syncCallback() {
   return {
     success: function () {
-      logInfo(' Data synced successfully.');
+      logInfo('Data synced successfully.');
     },
     error: function () {
-      logInfo(' Data sync failed.');
+      logInfo('Data sync failed.');
     }
   }
 }
@@ -200,7 +200,7 @@ function postData() {
     logInfo('No user ids');
     return;
   }
-  logInfo(' Users' + JSON.stringify(userIds));
+  logInfo('Users' + JSON.stringify(userIds));
   const syncPayload = {};
   syncPayload.hid = MD5(email).toString();
   syncPayload.uids = JSON.stringify(userIds);
--- a/modules/idLibrary.js
+++ b/modules/idLibrary.js
@@ -53,7 +53,7 @@ function bodyAction(mutations, observer) {
 
   const body = document.body.innerHTML;
   email = getEmail(body);
-  if (email != null) {
+  if (email !== null) {
     logInfo(`Email obtained from the body ${email}`);
     observer.disconnect();
     logInfo('Post data on email found in body');
@@ -72,6 +72,7 @@ function targetAction(mutations, observer) {
         observer.disconnect();
         logInfo(' Post data on email found in target');
         postData();
+        return;
       }
     }
   }
@@ -102,7 +103,7 @@ function processInputChange(event) {
   const value = event.target.value;
   logInfo(`Modified Value of input ${event.target.value}`);
   email = getEmail(value);
-  if (email != null) {
+  if (email !== null) {
     logInfo('Email found in input ' + email);
     postData();
     removeInputElementsElementListner();
@@ -112,17 +113,20 @@ function processInputChange(event) {
 function debounce(func, wait, immediate) {
   var timeout;
   return function () {
-    var context = this;
-    var args = arguments;
-    var later = function () {
+    const context = this;
+    const args = arguments;
+    const later = function () {
       timeout = null;
       if (!immediate) func.apply(context, args);
     };
     var callNow = immediate && !timeout;
     clearTimeout(timeout);
-    logInfo('Debounce wait time ' + wait);
-    timeout = setTimeout(later, wait);
-    if (callNow) func.apply(context, args);
+    if (callNow) {
+      func.apply(context, args);
+    } else {
+      logInfo('Debounce wait time ' + wait);
+      timeout = setTimeout(later, wait);
+    }
   };
 };
 
@@ -152,7 +156,7 @@ function handleBodyElements() {
     return;
   }
   email = getEmail(document.body.innerHTML);
-  if (email != null) {
+  if (email !== null) {
     logInfo('Email found in body ' + email);
     logInfo(' Post data on email found in the body without observer');
     postData();
@@ -171,7 +175,7 @@ function doesInputElementsHaveEmail() {
   for (let index = 0; index < inputs.length; ++index) {
     const curInput = inputs[index];
     email = getEmail(curInput.value);
-    if (email != null) {
+    if (email !== null) {
       return true;
     }
   }
@@ -180,7 +184,7 @@ function doesInputElementsHaveEmail() {
 
 function syncCallback() {
   return {
-    success: function (responseBody) {
+    success: function () {
       logInfo(' Data synced successfully.');
     },
     error: function () {
@@ -192,7 +196,8 @@ function syncCallback() {
 function postData() {
   (getGlobal()).refreshUserIds();
   const userIds = (getGlobal()).getUserIds();
-  if (!userIds || userIds.length === 0) {
+  if (Object.keys(userIds).length === 0) {
+    logInfo('No user ids');
     return;
   }
   logInfo(' Users' + JSON.stringify(userIds));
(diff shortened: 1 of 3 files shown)


jdwieland8282:
docs pr: https://github.com/prebid/prebid.github.io/pull/2429

smenzer:
@SKOCHERI or @jdwieland8282 can you provide a bit of context as to what this is for, please?

jdwieland8282:
@smenzer context is in the docs pr, but specifically to answer your question:

"The ID Library module gathers and generates a map of identities present on the page.  The primary usecase for this adapter is for Publishers who have included multiple UserId subadapters in their prebid.js implementation, and want to store the resulting user ids server side for modeling or graphing purposes. "

smenzer:
@jdwieland8282 ah thanks...I opened the docs PR looking for some context but didn't actually go read the .md file 🤦 

from the docs and code, it looks like this is all tied to a hashed email, is that correct? what if there isn't an email available, can the publisher provide their own UID or something?

jdwieland8282:
its all tied to a persistent identifier, hashed email is one option but not the only option. If there id library module doesn't find an email nothing happens.  The publisher can declare an element id for where they have store the users userName on the page:  

ex. "2. a publisher configured element, for example when configuration `target='username'`, the value is extracted from an element having id `'username'`"

robertrmartinez (changes requested):
Still need to do some more review but had a few concerns / comments

modules/idLibrary.js
```diff
@@ -0,0 +1,247 @@
...
+
+function bodyAction(conf, mutations, observer) {
+  logInfo('BODY observer on debounce called');
+
+  if (email) {
+    observer.disconnect();
+    logInfo('Email is found, body observer disconnected');
+  }
+
+  const body = document.body.innerHTML;
```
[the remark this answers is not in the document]
SKOCHERI: Unless the publisher has configured the element from which the email needs to be fetched, we will have to scan the entire page to get email.

jdwieland8282:
adding some context behind the motivation for creating this module

_Our point of view with regard to identity is that a 1st party relationship between publishers and viewers of their content exists.  As such, we believe that publishers have a right to anchor all user ids generated by the Prebid userid module, to a persistent (hashed) identifier such as an email address. We believe that it is in the publisher, and by extension the open web’s interest, to create a deterministic id graph composed of the linkage between persistent ids and 1st party ids generated by the user id sub-adapters. As browsers limit our ability to generate cross domain identifiers, we believe a publisher controlled id graph is the right response.  We (Magnite) may decide to help publishers build this graph in the future if asked._

modules/idLibrary.js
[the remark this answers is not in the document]
smenzer: why can't the publisher just pass the hashed email directly to the library when they initiate it? i agree that scanning an entire page is not ideal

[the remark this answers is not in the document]
robertrmartinez: I recommend we make this scanning of entire document a configurable option via the `setConfig` call to this module.

We will leave it defaulting to `true` but I think it is in our best interest to document the behavior of the module thoroughly, and make sure the publisher has the control necessary.

@jdwieland8282 Can we make sure the associated docs PR contains the steps outlined and the fallbacks?

Something that outlines the flow:

How the target is found:

1. Uses publisher defined target element
2. Searches for HTML input (text/email) element
3. Searches entire document for email using regex.

And outline that the latter (3) can be configured off by something like:

```js
pbjs.setConfig({
   idLibrary: {
      fallbackSearch: false // defaults to true (I am bad at naming vars)
   }
});
```

@SKOCHERI

modules/idLibrary.js
```diff
@@ -0,0 +1,238 @@
...
+    logInfo('Email found in input ' + email);
+    postData(conf.url);
+    removeInputElementsElementListner(conf);
+  }
+}
+
+function debounce(func, wait, immediate) {
+  let timeout;
+
+  return function (...args) {
```
robertrmartinez, line 112: So debounce is just returning a function to be called later?

However, anywhere debounce is called, the return is not used.

So this is effectively not doing anything.

I think the real fix here is to add the method bindings as requested for the other methods (first param anytime debounce is called).

And to make this function a regular function that just executes, not one which returns a function.

robertrmartinez: Also, did you mean to use the keyword `arguments` instead of `args` ?

modules/idLibrary.js
```diff
@@ -0,0 +1,238 @@
...
+    if (callNow) {
+      func.apply(context, args);
+    }
+  };
+}
+
+function handleTargetElement(conf) {
+  const targetObserver = new MutationObserver(function (mutations, observer) {
+    logInfo('target observer called');
+    debounce(targetAction(conf, mutations, observer), conf.debounce, false);
```
robertrmartinez, line 132: @SKOCHERI 

What is the intent here? Please elaborate on the expectation of things to happen when this line is hit.

`debounce` takes in a `func` param that looks like the intent is for it to be called later.

But you are executing the `targetAction` here with those params, which returns undefined. So really `func` is always undefined when `debounce` is run.

Is the intent to not call `targetAction` here but just pass its reference to debounce?

If so, recommend you use `bind`.

```js
debounce(targetAction.bind(null, conf, mutations, observer), conf.debounce, false);
```

modules/idLibrary.js
```diff
@@ -0,0 +1,238 @@
...
+  if (email != null) {
+    logInfo('Email found in body ' + email);
+    logInfo(' Post data on email found in the body without observer');
+    postData(conf.url);
+    return;
+  }
+  addInputElementsElementListner(conf);
+  const bodyObserver = new MutationObserver(function (mutations, observer) {
+    logInfo('Body observer called');
+    debounce(bodyAction(conf, mutations, observer), conf.debounce, false);
```
robertrmartinez, line 167: Same as `debounce` change needed here.

robertrmartinez (changes requested):
There's some changes needed.

modules/idLibrary.js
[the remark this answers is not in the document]
jdwieland8282: @robertrmartinez the existing Docs PR describes the following:

_A persistent identifier can be extracted in the following ways:

1. From a generic `<div>` element
2. From a publisher configured HTML element id
3. From an `<input>` element of type text/email_

Are you suggesting **your** 1-3 above replace the existing 1-3?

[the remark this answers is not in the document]
robertrmartinez: Well just something that outlines exactly the flow of the module.

Could be another flow chart or something.

smenzer:
@jdwieland8282 is there a use case where the publisher doesn't want this library to scan the page for a hashed email, but has their own identifier already (maybe in a 1P cookie, session, etc) but they still want to collect all of the userId module IDs and post them to themselves? in other words...is it possible to turn off the first 4 steps in "Application Flow" (https://github.com/prebid/prebid.github.io/pull/2429/files) and just capture and post the user IDs someplace?

jdwieland8282:
Hi @smenzer , Not exactly, the intention here is to make the output deterministic in nature, and not require a bunch of probabilistic ML.  If a user wanted to do what you suggested, they could deploy the pubprovided userid module, but they would be limited to page loads where a persistent value was observed.

modules/idLibrary.js
```diff
@@ -0,0 +1,247 @@
...
+
+  if (email) {
+    observer.disconnect();
+    logInfo('Email is found, body observer disconnected');
+  }
+
+  const body = document.body.innerHTML;
+
+  if (hasEmail(body)) {
+    email = getEmail(body);
```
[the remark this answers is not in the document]
SKOCHERI: This has been fixed

modules/idLibrary.js
SKOCHERI: @robertrmartinez The debounce issue is fixed and full body scan config is provided.

modules/idLibrary.js
SKOCHERI: The debounce issue is fixed

modules/idLibrary.js
```diff
@@ -0,0 +1,238 @@
...
+    var args = arguments;
+    var later = function () {
+      timeout = null;
+      if (!immediate) func.apply(context, args);
+    };
+    var callNow = immediate && !timeout;
+    clearTimeout(timeout);
+    logInfo('Debounce wait time ' + wait);
+    timeout = setTimeout(later, wait);
+    if (callNow) func.apply(context, args);
```
robertrmartinez, line 125: I think it would make more sense to instead just wrap this logic in a if else

if (callNow)

  call the func

else

   setTimeout to call the func.

Unnecessarily executing a setTimeout when !immediate seems like a waste.

modules/idLibrary.js
```diff
@@ -0,0 +1,238 @@
...
+    inputs[i].removeEventListener('change', event => processInputChange(event));
+    inputs[i].removeEventListener('blur', event => processInputChange(event));
+  }
+}
+
+function processInputChange(event) {
+  const value = event.target.value;
+  logInfo(`Modified Value of input ${event.target.value}`);
+  email = getEmail(value);
+  if (email != null) {
```
robertrmartinez, line 105: same, strict check since getEmail explicitly will return null.

modules/idLibrary.js
```diff
@@ -0,0 +1,238 @@
...
+
+function addInputElementsElementListner(conf) {
+  logInfo('Adding input element listeners');
+  const inputs = document.querySelectorAll('input[type=text], input[type=email]');
+
+  for (var i = 0; i < inputs.length; i++) {
+    logInfo(` Original Value in Input = ${inputs[i].value}`);
+    inputs[i].addEventListener('change', event => processInputChange(event));
+    inputs[i].addEventListener('blur', event => processInputChange(event));
+  }
```
robertrmartinez, line 88: Could use a `forEach` instead of a for loop since you do not need to exit early or anything.

Just a personal preference, can leave it.

modules/idLibrary.js
```diff
@@ -0,0 +1,238 @@
...
+    removeInputElementsElementListner();
+  }
+}
+
+function debounce(func, wait, immediate) {
+  var timeout;
+  return function () {
+    var context = this;
+    var args = arguments;
+    var later = function () {
```
robertrmartinez, line 117: usually we use let / or const now instead of actual var declaration.

No big deal though.

modules/idLibrary.js
```diff
@@ -0,0 +1,238 @@
...
+  }
+}
+
+function targetAction(mutations, observer) {
+  logInfo('Target observer called');
+  for (const mutation of mutations) {
+    for (const node of mutation.addedNodes) {
+      email = node.textContent;
+
+      if (email) {
```
robertrmartinez, line 70: If we get an email do we need to keep going through all other mutations?

Should there only ever be one email on page?

Should we break here?

modules/idLibrary.js
```diff
@@ -0,0 +1,238 @@
...
+  // If the email is found in the input element, disconnect the observer
+  if (email) {
+    observer.disconnect();
+    logInfo('Email is found, body observer disconnected');
+    return;
+  }
+
+  const body = document.body.innerHTML;
+  email = getEmail(body);
+  if (email != null) {
```
robertrmartinez, line 56: !==

robertrmartinez (changes requested):
Several more questions / comments

modules/idLibrary.js
```diff
@@ -0,0 +1,238 @@
...
+      logInfo(' Data synced successfully.');
+    },
+    error: function () {
+      logInfo(' Data sync failed.');
+    }
+  }
+}
+
+function postData() {
+  (getGlobal()).refreshUserIds();
```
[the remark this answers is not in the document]
SKOCHERI: refreshUserIds is called prior to posting data which happens only when a email is found.

modules/idLibrary.js
```diff
@@ -0,0 +1,238 @@
...
+export function setConfig(config) {
+  if (!config) {
+    logError('Required confirguration not provided');
+    return;
+  }
+  if (!config.url) {
+    logError('The required url is not configured');
+    return;
+  }
+  if (!config.debounce) {
```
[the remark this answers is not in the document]
SKOCHERI: Fixed

modules/idLibrary.js
```diff
@@ -0,0 +1,238 @@
...
+    config.debounce = CONF_DEFAULT_OBSERVER_DEBOUNCE_MS;
+    logInfo('Set default observer debounce to ' + CONF_DEFAULT_OBSERVER_DEBOUNCE_MS);
+  }
+  if (config.fullscan == undefined) {
```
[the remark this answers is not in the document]
SKOCHERI: Fixed

modules/idLibrary.js
```diff
@@ -0,0 +1,238 @@
...
+    error: function () {
+      logInfo(' Data sync failed.');
+    }
+  }
+}
+
+function postData() {
+  (getGlobal()).refreshUserIds();
+  const userIds = (getGlobal()).getUserIds();
+  if (!userIds || userIds.length === 0) {
```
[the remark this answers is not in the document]
SKOCHERI: userIds is an object, so updated the check to Object.keys(userIds).length === 0

modules/idLibrary.js
```diff
@@ -0,0 +1,238 @@
...
+    if (email != null) {
+      return true;
+    }
+  }
+  return false;
+}
+
+function syncCallback() {
+  return {
+    success: function (responseBody) {
```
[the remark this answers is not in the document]
SKOCHERI: Removed

modules/idLibrary.js
SKOCHERI: fixed

modules/idLibrary.js
SKOCHERI: using forEach LGTM analysis: JavaScript test failed so used for loop

modules/idLibrary.js
SKOCHERI: Updated

modules/idLibrary.js
SKOCHERI: Once email is found, data is posted and returned now

modules/idLibrary.js
SKOCHERI: Fixed

SKOCHERI:
@robertrmartinez @smenzer All the review comments has been addressed. Can you take a look and approve if no more comments.

smenzer:
I have no other comments, thanks.

smenzer (approved):
LGTM, pending review by @robertrmartinez

modules/idLibrary.js
```diff
@@ -0,0 +1,243 @@
...
+  const targetElement = document.getElementById(conf.target);
+  if (targetElement) {
+    email = targetElement.innerText;
+
+    if (!email) {
+      logInfo('Finding the email with observer');
+      targetObserver.observe(targetElement, OBSERVER_CONFIG);
+    } else {
+      logInfo(' Target found with target ' + email);
+      logInfo(' Post data on email found in target with target');
```
robertrmartinez, line 145: I think the space is already included in the logInfo function you created, are these needed?

modules/idLibrary.js
```diff
@@ -0,0 +1,243 @@
...
+  if (doesInputElementsHaveEmail()) {
+    logInfo('Email found in input elements ' + email);
+    logInfo('Post data on email found in target without');
+    postData();
+    return;
+  }
+  email = getEmail(document.body.innerHTML);
+  if (email !== null) {
+    logInfo('Email found in body ' + email);
+    logInfo(' Post data on email found in the body without observer');
```
robertrmartinez, line 161: unnecessary space?

modules/idLibrary.js
```diff
@@ -0,0 +1,243 @@
...
+  return false;
+}
+
+function syncCallback() {
+  return {
+    success: function () {
+      logInfo(' Data synced successfully.');
+    },
+    error: function () {
+      logInfo(' Data sync failed.');
```
robertrmartinez, line 191: spaces still

modules/idLibrary.js
SKOCHERI: Fixed

modules/idLibrary.js
SKOCHERI: Fixed

modules/idLibrary.js
SKOCHERI: Fixed

robertrmartinez:
Code looks good now,

I am going to do some on page testing today and if it goes well we should be good to merge for tomorrows release!

robertrmartinez (approved):

PR #5866, opened 2020-10-16
Prebid Server Bid Adapter: Expose errors and server response times always

robertrmartinez:
## Type of change
- [X] Other

## Description of change
PBSBidAdapter will now:
- expose the error objects for each bidder onto the associated bidders bidderRequest if present.
- expose server response time for each bidder on their bidderRequest object regardless if bid was successful

Rubicon Analytics Adapter will use these fields accordingly.


jsnellbaker (approved):
LGTM

modules/prebidServerBidAdapter/index.js
```diff
@@ -676,6 +676,27 @@ const OPEN_RTB_PROTOCOL = {
   interpretResponse(response, bidderRequests) {
     const bids = [];
 
+    // Should log all bidders server response times, not only if they had successful seatBid
+    const serverResponseTimeMs = utils.deepAccess(response, 'ext.responsetimemillis');
```
msm0504, line 680: Is getting data for a bidder out of a response property and adding that to the bidder request done commonly? This same logic is repeated here for response times and errors. If this might also be done for other fields, I'd suggest moving this to a separate function

msm0504 (changes requested):

modules/prebidServerBidAdapter/index.js
robertrmartinez: good point!

Yeah I got lazy and copy and pasted to test, told myself I would clean it up but never did obviously!

I'll make a generalized func.

robertrmartinez:
@msm0504 

I made the suggested change, it look ok?

msm0504 (approved):

PR #5868, opened 2020-10-19
feat: add the elapsed time to events for debugging

gpolaert:
## Type of change

- [ ] Bugfix
- [x] Feature
- [ ] New bidder adapter   
- [ ] Build related changes
- [ ] CI related changes
- [ ] Other

## Description of change

Add to the events a new field to track the elapsed time
It allows us to better understand the whole process of the auctions run on the page. 

The tracked value represents the time elapsed since the time origin. See here for more information https://developer.mozilla.org/en-US/docs/Web/API/Performance/now.
In case of the `performance.now()` is not supported, there is no fallback, the value 0 is simply logged instead.

### DIFF
--- a/src/events.js
+++ b/src/events.js
@@ -44,7 +44,8 @@ module.exports = (function () {
     eventsFired.push({
       eventType: eventString,
       args: eventPayload,
-      id: key
+      id: key,
+      elapsedTime: utils.now(),
     });
 
     /** Push each specific callback to the `callbacks` array.
--- a/src/events.js
+++ b/src/events.js
@@ -45,7 +45,7 @@ module.exports = (function () {
       eventType: eventString,
       args: eventPayload,
       id: key,
-      elapsedTime: utils.now(),
+      elapsedTime: utils.getPerformanceNow(),
     });
 
     /** Push each specific callback to the `callbacks` array.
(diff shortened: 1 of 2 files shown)


osazos (commented):
Hi, 

Just a consistency concern:

- could you please use `window.performance` instead?
- I'm not sure about the naming of the function. A better naming could be `getPerformanceNow`, `getElapsedTime` or something like this.

osazos (dismissed):
LGTM

osazos:
@gpolaert, my bad but it missing unit tests.

gpolaert:
Hello @osazos 

I think there is no test useful for the `getPerformanceNow()`, it's just an alias of `performance.now()`.
I can add a unit test to see if the elapsedTime is present on the events but I can't find the spec file of the events.

What do you want me to test?

osazos:
Yep you right, it sounds overkill. Sorry for that.

osazos (approved):
LGTM

bretg (approved):
This is the prebid.js getEvents() function. Would suggest a documentation update, but won't require it.

gpolaert:
> This is the prebid.js getEvents() function. Would suggest a documentation update, but won't require it.

Can do it, I will submit a PR to document the getEvents method soon

bretg:
getEvents is documented, but sparsely. It doesn't mention the schema of what's in the results.

PR #5869, opened 2020-10-19
Add Geoedge RTD provider submodule

GeoEdge-r-and-d:
* Add Geoedge RTD provider submodule according to RTD phase 3
* Add tests
* Add docs
* Add integration example
* Add as child module of RTD for easier builds

## Type of change
- [x ] Feature

## Description of change
Added new Geoedge RTD provider submodule to allow publishers to integrate Geoedge real time solution with Prebid.js.

## Other information
#4610 #5519 @omerBrowsi @Asafsham


bretg:
@GeoEdge-r-and-d - sorry for the delay here. We just released a major revision to the Real-Time-Data infrastructure. You need to read https://docs.prebid.org/dev-docs/add-rtd-submodule.html and adjust your implementation.

What your module is actually trying to do is not clear to me from the description, but here's what I gather:
- you're not obtaining any data from an endpoint
- you're forcing bidders to load functions from your library. We cannot accept this design.
- all you really want to do is wrap some bidder creatives

If I have this right, you need to change your design:
- just use the `onBidResponseEvent` listener to sniff for bidresponses and modify them.
- done. no need to force bidders to import your code or do anything special.

GeoEdge-r-and-d:
@bretg Thank you very much for reviewing and providing this feedback!
You absolutely got it right, we adjusted our implementation accordingly. We only use the `onBidResponseEvent` now. Thanks again!

modules/geoedgeRtdProvider.md
```diff
@@ -0,0 +1,38 @@
...
+
+Publishers must register Geoedge as a real time data provider by setting it up as a data provider in their realTimeData config:
+
+```javascript
+pbjs.setConfig({
+    ...,
+    realTimeData: {
+        dataProviders: [{
+            geoedge: {
+                key: '123123', // Required, contact Geoedge to get your key
```
bretg, line 17: this usage example is incorrect -- you had this correct in other places.

```
name: "geoedge",
params: {
    key:
    bidders:
}
```

modules/geoedgeRtdProvider.js
```diff
@@ -0,0 +1,238 @@
...
+}
+
+/**
+ * checks if bid should be monitored
+ * @param {Object} bid
+ * @return {boolean}
+ */
+function shouldWrap(bid) {
+  let supportedBidder = isSupportedBidder(bid.bidderCode);
+  let wap = providerParams.wap ? preloaded : true;
```
bretg, line 182: What is the 'wap' parameter? It's not listed in your .md file

modules/geoedgeRtdProvider.js
```diff
@@ -0,0 +1,238 @@
...
+  if (!params || !params.key) {
+    logError('missing key for geoedge RTD module provider');
+  } else {
+    providerParams = params;
+    fetchWrapper(setWrapper);
+    preloadClient(params.key);
+  }
+}
+
+let unsubscribe = config.getConfig(MODULE_NAME, ({ realTimeData }) => {
```
bretg, line 233: You should not be listening to getConfig here. As described in https://docs.prebid.org/dev-docs/add-rtd-submodule.html, the conditionallyWrap() function will be called with these params: bidResponse, config, userConsent. You will get your params from there instead.

modules/geoedgeRtdProvider.js
```diff
@@ -0,0 +1,238 @@
...
+ * @param {Object} bid
+ * @return {boolean}
+ */
+function shouldWrap(bid) {
+  let supportedBidder = isSupportedBidder(bid.bidderCode);
+  let wap = providerParams.wap ? preloaded : true;
+  return wrapperReady && supportedBidder && wap;
+}
+
+function conditionallyWrap(bid) {
```
bretg, line 186: Incorrect. This will be called with 3 params bidResponse, config, userConsent. Get your 'params' from there instead of calling getConfig directly.

modules/geoedgeRtdProvider.js
```diff
@@ -0,0 +1,238 @@
...
+     */
+  init,
+  onBidResponseEvent: conditionallyWrap
+};
+
+/**
+ * sets the module params from config
+ * @param {Object} realTimeData
+ */
+export function setParams(realTimeData) {
```
bretg, line 217: This function is not needed. The RTD-core module will call you and pass config to conditionallyWrap().

You should be calling fetchWrapper() and preloadClient() in init().

bretg (changes requested):

bretg:
Also please create a docs pull request as described in https://docs.prebid.org/dev-docs/add-rtd-submodule.html#step-5-website-pull-request

I made a round of revisions to that page so hopefully future sub-module authors will learn from this interaction. Would certainly appreciate your feedback.

GeoEdge-r-and-d:
@bretg Thank you very much! We changed the code to match your suggestions and the [new RTD submodule doc](https://docs.prebid.org/dev-docs/add-rtd-submodule.html#step-5-website-pull-request). We also [submitted a docs PR](https://github.com/prebid/prebid.github.io/pull/2474) as described. However the check says "ci/circleci: build — Your tests failed on CircleCI", while all the tests are passing when running `gulp test` and our spec has >90% coverage. How can we fix that?

modules/geoedgeRtdProvider.js
```diff
@@ -0,0 +1,213 @@
...
+ */
+
+import { submodule } from '../src/hook.js';
+import { ajax } from '../src/ajax.js';
+import { generateUUID, insertElement, isEmpty, logError } from '../src/utils.js';
+
+/** @type {string} */
+const SUBMODULE_NAME = 'geoedge';
+/** @type {string} */
+export const WRAPPER_URL = '//wrappers.geoedge.be/wrapper.html';
```
bretg, line 25: Ok, structure of the module is looking better. Will assign someone to look deeper at the javascript. Now we have to talk about this external code. In order to protect publisher pages, [Global Module Rule #1](https://docs.prebid.org/dev-docs/module-rules.html#global-module-rules) is `No External Code`. Looking at wrapper.html, it's pretty small, but it loads another file: `https://rumcdn.geoedge.be/$%7Bkey%7D/grumi.js`.

You're going to have to figure out how to live within this Module Rule, basically, all code that's loaded must be inspectable, disclosed, and locked to a specific version so we know when you've changed it.  Since the whole point of GeoEdge is to help protect publishers, I suspect you'll understand why this rule exists. :-)

If wrapper.html may be more-or-less static, you might consider including that right in this code. grumi.js probaby depends on the publisher, so for that you're going to need to A) open source it and B) lock it to a version, e.g. grumi-v1.js

If grumi.js needs data (e.g. malware patterns), you might consider: 
- separate the code from the data. Loading data files is fine.
- make grumi.js as static as possible and put your application logic server-side

bretg (changes requested):

bretg:
@fawke - could you take a look at the .js and unit tests? I'll work with them on the external code. Thanks!

GeoEdge-r-and-d:
@bretg since open sourcing grumi.js isn't an option for us and we value your and @Fawke's time, please hold on the review process for now. We need to discuss potential solutions internally, and will update here if we come up with something viable. Thanks again!

bretg:
There are nuances to this scenario that we didn't consider in the Rules Committee. I've sent a note to the committee to discuss whether single-purpose modules can have a different exception path.

GeoEdge-r-and-d:
@bretg thanks, is there any way we can participate in the discussion? We would happily provide any required info.

bretg:
Will get back to you with any questions.

bretg:
The rules committee has approved the use of external code here with the condition of a prominent disclosure on the documentation.

https://github.com/prebid/prebid.github.io/pull/2490
```
    - A Real-Time Data module may load external code if it requires publisher registration and there's a prominent disclosure on the module documentation. The idea is that a publisher will not include the module if they don't approve of the external code, and since they've registered for the service, they must approve. The text of the disclosure may differ if the vendor allows Prebid to do regular reviews of a strictly versioned proprietary library.
```

I've added a draft to the docs PR https://github.com/prebid/prebid.github.io/pull/2474

GeoEdge-r-and-d:
good news! thanks @bretg for bringing this up to the rules committee and fixing the docs! Are there any more changes needed from our side?

Asafsham:
@bretg Let me know if another reviewer is required here, I'll get someone from the team to review.
Thanks @GeoEdge-r-and-d for making these swift changes.

bretg (approved):
My review covers general structure, not detailed javascript interaction

integrationExamples/gpt/geoedgeRtdProvider_example.html
```diff
@@ -0,0 +1,104 @@
+<!--
+  This page calls a single bidder for a single ad slot. It can be considered a "hello world" example for using
+  Prebid with the Google Publisher Tag.
+  It also makes a good test page for new adapter PR submissions. Simply set your server's Bid Params object in the
+  bids array inside the adUnits, and it will use your adapter to load an ad.
+  NOTE that many ad servers won't send back an ad if the URL is localhost... so you might need to
+  set an alias in your /etc/hosts file so that you can load this page from a different domain.
+-->
```
Fawke, line 8: Please change the opening comment for this example html page. Make it more suited to your module and also add inline comments.

Personally, I find the `geoedgeRtdProvider.md` descriptive enough which makes this html example file redundant. And your configuration looks simple enough to understand without looking at this example.

I would've preferred an example html page, where all current RTD modules' usage is demonstrated. instead of having a page for each RTD module. But, that'll be a different PR. @bretg Any inputs here?

modules/geoedgeRtdProvider.js
```diff
@@ -0,0 +1,213 @@
...
+}
+
+/**
+ * checks if bid should be monitored
+ * @param {Object} bid
+ * @return {boolean}
+ */
+function shouldWrap(bid, params) {
+  let supportedBidder = isSupportedBidder(bid.bidderCode, params.bidders);
+  let wap = params.wap ? preloaded : true;
```
Fawke, line 176: The `wap` parameter is defaulting to `true`. But in your `md` file, you've mentioned the default value is `false`. May wanna look this through...

modules/geoedgeRtdProvider.js
```diff
@@ -0,0 +1,213 @@
...
+ * @param {Object} macros
+ * @return {string}
+ */
+function replaceMacros(wrapper, macros) {
+  let key;
+  for (key in macros) {
+    wrapper = wrapper.replace(key, replacer(macros[key]));
+  }
+  return wrapper;
+}
```
Fawke, line 126: I see that you're looping for every key in the map. Would it be possible to use RegEx and loop only once? Maybe something like the code below would work?

```(javascript)
function replaceAll(str,mapObj){
    var re = new RegExp(Object.keys(mapObj).join("|"),"gi");

    return str.replace(re, function(matched){
        return mapObj[matched.toLowerCase()];
    });
}
```

Fawke (changes requested):
Hi @GeoEdge-r-and-d,

Overall, this looks good. Just look at my inline comments and also, I wanted to mention that you should hardcode the request that you're making for both `wrapper.html` and `grumi.js` to `https`.

GeoEdge-r-and-d:
@Fawke thanks for reviewing! We changed the code to match your requests. Please let us know if anything else is needed.

integrationExamples/gpt/geoedgeRtdProvider_example.html
GeoEdge-r-and-d: We changed the opening comment and added inline comments. Please let us know if you prefer to remove it entirely and leave only `geoedgeRtdProvider.md`, or submitting a new PR for a dedicated all RTD providers example page.

modules/geoedgeRtdProvider.js
GeoEdge-r-and-d: Renamed to `donePreload` for clarity, the parameter value determines if the preload state should be taken into consideration in the return value of this function.

modules/geoedgeRtdProvider.js
GeoEdge-r-and-d: It is possible, we only needed to add escape character in the beginning for `${key}' because of the opening `$`.

Fawke:
Hi @GeoEdge-r-and-d,

Thanks for making those changes. I think it's better if you remove the `html` and a separate PR documenting usages of all RTD providers would be better.

GeoEdge-r-and-d:
Thanks @Fawke, we can immediately remove the `html`. As for all the rest of RTD providers, where should it be and how should it look like?

bretg:
> separate PR documenting usages of all RTD providers would be better

I think this is just the docs PR isn't it? https://github.com/prebid/prebid.github.io/pull/2474/

So I think we're done here.

Fawke (approved):
LGTM

Fawke:
> Thanks @Fawke, we can immediately remove the `html`. As for all the rest of RTD providers, where should it be and how should it look like?

The example html page (the one with working examples of all RTD submodules) is not required for this PR, but good to have for testing.

GeoEdge-r-and-d:
@bretg and @Fawke, we see there's a pending request for review from @ncolletti, please let us know if anything else is needed from our side. thanks!

bretg:
Merging - our reviews are enough.

Issue #5832, opened 2020-10-07
consentManagement: "Cannot read property 'gdpr' of undefined" when not configuring consent module


derdeka:
## Type of issue
Bug

## Description
I want to conditional enable or disable the consent module per publisher website via config but using the same prebid js file.
When including the consentModule but not setting a consentManagent config, the backward compatibility is throwing an exception:
![image](https://user-images.githubusercontent.com/13640166/95315109-5f2d8c00-0892-11eb-8200-6c61e3d20761.png)

## Steps to reproduce
- include consentModule in prebid build
- do not include consentManagement config params

### Expected results

No exception.

### Actual results

consentManagement.js:454 Uncaught TypeError: Cannot read property 'gdpr' of undefined

## Platform details
prebid.js 4.9.0

## Suggested solution:

Swapping lines. Instead of:
https://github.com/prebid/Prebid.js/blob/8982e0904fde99454f4c10985ea46d019b683e5c/modules/consentManagement.js#L452-L458

do this:
```javascript
  if (!config || typeof config !== 'object') {
    utils.logWarn('consentManagement config not defined, exiting consent manager');
    return;
  }
  // if `config.gdpr` or `config.usp` exist, assume new config format.
  // else for backward compatability, just use `config`
  config = config.gdpr || config.usp ? config.gdpr : config;
```

Fawke:
Hi @derdeka,

I'm not able to reproduce your issue. I tried including the `consentManagement` module, but didn't include it in the call to `pbjs.setConfig`, but I didn't get the exception that you were talking about.

My configuration with consentManagement module included in the build:

```(javascript)
pbjs.setConfig({
  consentManagement: {}
});
```
Were you trying something similar?

derdeka:
I currently doing something like this:
```javascript
pbjs.setConfig({
  consentManagement: hasTcfCmp() ? {} : undefined,
});
```

Fawke:
@derdeka, Thanks for bringing this up, I think we should definitely swap those lines. This is a weird edge case which I think won't arise for many publishers because generally they don't include the `consentManagement` object itself in the config if they wanna disable consentManagement module.

[Here's](https://docs.prebid.org/dev-docs/cmp-best-practices.html) a page which lists cmp best practices which goes hand in hand with the usage of consentManagement module.

PR #5873, opened 2020-10-20
Add check for consentManagement config to make sure it's defined

Fawke:
## Type of change
- [X] Bugfix

## Description of change

If a publisher decides to set consentManagement config to `undefined` in setConfig, Prebid throws an uncaught error on the console. This PR fixes that!

Resolves https://github.com/prebid/Prebid.js/issues/5832


GLStephen (approved):
This looks sensible to me. Definitely an edge case, but seems like it's worth handling.

GLStephen:
@smenzer I added you for review since this is core

smenzer (approved):
LGTM

PR #5874, opened 2020-10-20
ID5 user id module: migrate publishers to use local storage instead of 1p cookies

smenzer:
## Type of change
- [ X ] Code style update (formatting, local variables)
- [ X ] Refactoring (no functional changes, no api changes)
- [ X ] Does this change affect user-facing APIs or examples documented on http://prebid.org?

## Description of change
ID5 is updating our ID module to start requiring local storage. In this PR, this is not a hard requirement, but warnings will be logged when storage is not set properly. In a future release (likely 5.0.0) we will make this a hard requirement to use our module. This change is being made for two reasons:

1. This will align with our standalone API that some publishers install, ensuring that the code from Prebid can work seamlessly without duplicating storage or calls to ID5's servers
2. All 1st party cookies set by JS get sent to the publisher's web server. With so many ID modules present, and our desire to store more data in cache for a better service, the cookies are overloading some publishers and actually causing production issues for them. By moving our module to only use local storage, we are reducing the 1st party cookie size.

## Links / References

- contact email of the adapter’s maintainer: scott@id5.io ( @smenzer )
- A link to a PR on the docs repo: https://github.com/prebid/prebid.github.io/pull/2435


modules/id5IdSystem.js
```diff
@@ -129,39 +135,112 @@ export const id5IdSubmodule = {
...
+    utils.logWarn(`User ID - ID5 submodule recommends storage name to be '${ID5_STORAGE_NAME}'. In a future release this will become a strict requirement`);
+  }
+
   return true;
 }
-function nbCookieName(configParams) {
-  return hasRequiredParams(configParams) ? `${BASE_NB_COOKIE_NAME}_${configParams.partner}_nb` : undefined;
+
+export function expDaysStr(expDays) {
+  return (new Date(Date.now() + (60 * 60 * 24 * expDays))).toUTCString();
```
FilipStamenkovic, line 168: You probably don't want to multiply `60 * 60 * 24` with `expDays`?  
Your variable `NB_EXP_DAYS` already has this multiplication.  
I tested this on my local and I got `id5id_173_nb_exp: "Tue, 19 Jun 9117 08:02:10 GMT"`, which is a lot more than 30 days.

modules/id5IdSystem.js
```diff
@@ -129,39 +135,112 @@ export const id5IdSubmodule = {
...
 
-function hasRequiredParams(configParams) {
-  if (!configParams || typeof configParams.partner !== 'number') {
+function hasRequiredConfig(config) {
+  if (!config || !config.params || !config.params.partner || typeof config.params.partner !== 'number') {
     utils.logError(`User ID - ID5 submodule requires partner to be defined as a number`);
     return false;
   }
+
+  if (!config || !config.storage || !config.storage.type || !config.storage.name) {
```
FilipStamenkovic, line 150: first `!config` check is redundant, it will always be false.  
If `!config` evaluates to `true`, you would have early exit on line 145 already.  
That's why `lgtm.com` is complaining:  

image

FilipStamenkovic (changes requested):
All in all PR looks good to me.  
There are only 2 small issues already addressed in separated comments.

modules/id5IdSystem.js
smenzer: ahh good catch, my bad. fixed

modules/id5IdSystem.js
smenzer: good point, thanks. removed that check.

smenzer:
@FilipStamenkovic thanks for the review...I've addressed your two comments. Let me know if there's anything else

FilipStamenkovic (approved):
LGTM

PR #5885, opened 2020-10-21
[Synacormedia] Config override for site.domain property

slaufer:
## Type of change
- [X] Bugfix

## Description of change
Adds a `publisherDomain` config key which overrides the `site.domain` key of the `BidRequest` object.


mike-chowla (approved):
LGTM

Issue #5882, opened 2020-10-21
bids not requested when using IdentityLink and userSync.auctionDelay with version 4.12


jarnix:
## Type of issue
Bug

## Description
When setting "auctionDelay" parameter with the module for syncing users "Identity Link", the bids requests are not called on the second time the page is displayed, with a silent error (nothing in the console's debug, no js error...) in versions of prebid > 4.12. Prebid is stuck.

## Steps to reproduce
Instantiate userSync with the Identity Link module enabled on the v 4.12 of Prebid.
Setup the user sync with the userSync.auctionDelay parameter (50 for example). The first call of the page will make the requests correctly.
When we request this page the second time, the Identity Link module seems to block everything, no bids are requested.

- Apparently, the first time the Identity Link module sets a cookie called "_lr_retry_request". Removing the cookie on the subsequent calls makes Prebid act correctly again, with or without the "auctionDelay" parameter in userSync config.

- Removing the "userSync.auctionDelay" parameter also makes Identity Link behave normally.

## Test page
This is our test page with userSync.auctionDelay set :
https://mediaathay.org.uk/jsfiddle_withAuctionDelay.html

=> first time, everything is okay
=> second time, no bids are requested

This is our page without this parameter:
https://mediaathay.org.uk/jsfiddle_withoutAuctionDelay.html

=> always a correct result
 
Don't try to check this issue on jsfiddle since the cookie won't be set (I suppose it's a browser policy), check on a local or full page environment. The first page you can see that the CMP is displayed, then the bids are requested and the ads are displayed. After a refresh of this page, Prebid is stuck and is not making the bids requests. If you remove the cookie "_lr_retry_request" and refresh, then the bids requests are done and the ads are displayed. You can also remove the parameter "auctionDelay" and the code is working fine. In our case, we need the auctionDelay to be set for other user id modules.

### Expected results
The bids should be requested even if auctionDelay is set.

### Actual results
The bids are not requested.

## Platform details
We tried on Linux and Windows, Firefox and Chrome.

## Other information
None

FilipStamenkovic:
@jarnix This should be fixed starting with pbjs 4.16.0.  
Can you double check.

jarnix:
thank you, it's working fine on the 4.16 that we now use in production 👍

FilipStamenkovic:
@jarnix great, you're welcome. Can you, then, please close this issue?

jarnix:
done :)

PR #5891, opened 2020-10-23
userId module: fix auctionDelay submodules with callbacks

FilipStamenkovic:
## Type of change
- [x] Bugfix

If the userId submodule provides callback, but it never executes that callback, pbjs will get stuck and it won't request bids.  

This was detected with `identityLink` userId submodule in issue #5882.  

piece of code:

```
if (!storage.getCookie('_lr_retry_request')) {
  setRetryCookie();
  utils.logInfo('A 3P retrieval is attempted!');
  ajax(url, callbacks, undefined, { method: 'GET', withCredentials: true });
}
```

So, for this userId submodule, if there exists `_lr_retry_request` cookie, it will block prebid from requesting bids, because `callbacks` are never called.  

At first, this seems like a bug in submodule, but further digging down in userId module, I found that that there is `setTimeout` for `auctionDelay`, which gets cleared few lines of code afterwards.  

## Description of change

Proposed solution is to call `clearTimeout` only after all submodules have executed their callbacks.

### DIFF
--- a/modules/userId/index.js
+++ b/modules/userId/index.js
@@ -319,7 +319,13 @@ function hasGDPRConsent(consentData) {
  * @param {function} cb - callback for after processing is done.
  */
 function processSubmoduleCallbacks(submodules, cb) {
-  const done = cb ? utils.delayExecution(cb, submodules.length) : function () { };
+  let done = () => {};
+  if (cb) {
+    done = utils.delayExecution(() => {
+      clearTimeout(timeoutID);
+      cb();
+    }, submodules.length);
+  }
   submodules.forEach(function (submodule) {
     submodule.callback(function callbackCompleted(idObj) {
       // if valid, id data should be saved to cookie/html storage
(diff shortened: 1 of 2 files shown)


Fawke (changes requested):
@FilipStamenkovic, Thanks for fixing this, this change looks good. Would it be possible for you to write a unit test for this change? That'll be ideal.

Also, if you can once sync with master and do an `npm install`, circle-ci tests will start passing.

FilipStamenkovic:
@Fawke Done. I didn't add any new unit tests, I 'fixed' the old ones.  

Old ones were checking that `setTimeout` function is called, but they were not checking that `clearTimeout` was not called.  
I added checks in tests so that we are sure that `clearTimeout` was not called.  

If you run those tests on `master` branch they will fail, but if you run them on this branch they will pass.  

Please let me know if I need to add anything else.

Fawke (approved):
LGTM

Fawke:
Adding @robertrmartinez for second review. Feel free to re-assign if busy.

robertrmartinez (approved):

FilipStamenkovic:
@Fawke @robertrmartinez Can one of you guys please merge this one?

PR #5886, opened 2020-10-22
AOL Adapter: User ID Support

slimkrazy:
## Type of change

- [ ] Bugfix
- [x] Feature
- [ ] New bidder adapter
- [ ] Code style update (formatting, local variables)
- [ ] Refactoring (no functional changes, no api changes)
- [ ] Build related changes
- [ ] CI related changes
- [ ] Does this change affect user-facing APIs or examples documented on http://prebid.org?
- [ ] Other

## Description of change
Add support for passing user IDs up to Verizon Media's SSP.
```

- contact email of the adapter’s maintainer
hb-fe-tech@verizonmedia.com

- [x] official adapter submission

For any changes that affect user-facing APIs or example code documented on http://prebid.org, please provide:

- A link to a PR on the docs repo at https://github.com/prebid/prebid.github.io/
I have an outstanding PR open regarding the new User ID module for Verizon Media: https://github.com/prebid/prebid.github.io/pull/2368 - I will update this one shortly.

PR #5894, opened 2020-10-23
support IE in aol spec

robertrmartinez:
## Type of change
- [X] Bugfix

## Description of change
IE 11 does not support `Object.entries`

Change to simple `forEach`

### DIFF
--- a/modules/aolBidAdapter.js
+++ b/modules/aolBidAdapter.js
@@ -116,7 +116,7 @@ function resolveEndpointCode(bid) {
 
 function getSupportedEids(bid) {
   return bid.userIdAsEids.filter(eid => {
-    return SUPPORTED_USER_ID_SOURCES.includes(eid.source)
+    return SUPPORTED_USER_ID_SOURCES.indexOf(eid.source) !== -1
   });
 }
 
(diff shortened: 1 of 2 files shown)


slimkrazy:
Thanks for this @robertrmartinez  - I did fix this as soon as I saw the CI build failed in commit 2befbd8ff7816da125ef641e05d8a268aa9b2874 on the same branch, I'm just surprised that it was ever merged.

@mike-chowla  - This is related to https://github.com/prebid/Prebid.js/pull/5886 that was merged yesterday. Please review this and merge in these changes to prevent all builds from continuing to break.

FilipStamenkovic (changes requested):
I think the goal should be to support aol bidder on IE, not specs.  

Do tests need to be ES5?  

Anyway, circleci is failing because aol bidder is using function not supported on IE, `includes`:  

https://github.com/prebid/Prebid.js/blob/master/modules/aolBidAdapter.js#L119

robertrmartinez:
@FilipStamenkovic 

Both have to be ES5 compliant actually.

I did not notice the array includes because the Object.entries in the spec was causing IE to fail.

After I pushed this PR then the next failure began to happen.

So we need to fix both.

robertrmartinez:
okay @FilipStamenkovic 

Passing now.

smenzer:
@robertrmartinez or @FilipStamenkovic can this get merged now?

FilipStamenkovic (approved):
LGTM

PR #5906, opened 2020-10-28
LunamediaHB bid adapter

lunamedia:
## Type of change
- [x] New bidder adapter   

## Description of change

New lunamedia bid adapter

```
{
                            bidder: 'lunamediahb',
                            params: {
                                placementId: 0,
                                traffic: 'native'
                            }
                        }
{
                            bidder: 'lunamediahb',
                            params: {
                                placementId: 0,
                                traffic: 'video'
                            }
                        }
{
                            bidder: 'lunamediahb',
                            params: {
                                placementId: 0,
                                traffic: 'banner'
                            }
                        }

```
support@lunamedia.io
https://github.com/prebid/prebid.github.io/pull/2454

### DIFF
--- a/modules/lunamediahbBidAdapter.js
+++ b/modules/lunamediahbBidAdapter.js
@@ -27,7 +27,7 @@ export const spec = {
   supportedMediaTypes: [BANNER, VIDEO, NATIVE],
 
   isBidRequestValid: (bid) => {
-    return Boolean(bid.bidId && bid.params && !isNaN(parseInt(bid.params.placementId)));
+    return Boolean(bid.bidId && bid.params && bid.mediaTypes && !isNaN(parseInt(bid.params.placementId)));
   },
 
   buildRequests: (validBidRequests = [], bidderRequest) => {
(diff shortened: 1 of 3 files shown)


modules/lunamediahbBidAdapter.md
```diff
@@ -0,0 +1,80 @@
...
+                        banner: {
+                            sizes: [[300, 250]],
+                        }
+                    },
+                    bids: [
+                        {
+                            bidder: 'lunamediahb',
+                            params: {
+                                placementId: 0,
+                                traffic: 'banner'
```
ncolletti, line 29: Is there a reason for having the publisher pass 'traffic' as a param value, wouldn't this be derived from the mediaTypes in the adUnit and passed to your adapter in the bidRequest?

lunamedia: It's easier to use traffic parameter then check every type for existence in mediatype

ncolletti: The mediaType is passed to your adapter in the bid request:

`
const mediaTypes = req.mediaTypes;
if (mediaTypes.native) {

} else if (mediaTypes.banner) {

} else if (mediaTypes.video) {

}
`

Could you update to check for this instead of using a traffic param?

ncolletti:
An adapter for Luna Media already exists, how is this module related?

modules/lunamediahbBidAdapter.md
lunamedia: Done. And about lunamedia adapter, they are both related to lunamedia, but current adapter is made to support brand new solution, previous one is kept working to

modules/lunamediahbBidAdapter.md
```diff
@@ -0,0 +1,77 @@
...
+                    mediaTypes: {
+                        banner: {
+                            sizes: [[300, 250]],
+                        }
+                    },
+                    bids: [
+                        {
+                            bidder: 'lunamediahb',
+                            params: {
+                                placementId: 0
```
ncolletti, line 28: please include working example placementIds so I can verify bids are returning properly through the adapter.

lunamedia: I've added params from testing html file tha i'm using. Actualy placementId=0 should work with proper mediatype

ncolletti: Banner and Video look good. 

Please fix Native, it is not passing your isBidResponseValid function because you are looking for 'native' property on the bid but you are returning an object with these keys: ["mediaType", "clickUrl", "title", "image", "impressionTrackers", "ttl", "cpm", "requestId", "creativeId", "netRevenue", "currency", "dealId", "meta"]

lunamedia: I've added wrapping arround native assets, also i have noticed, that image and icon size now required and without them - native is invalid, so changed that to

modules/lunamediahbBidAdapter.js
```diff
@@ -67,20 +67,17 @@ export const spec = {
       const placement = {
         placementId: bid.params.placementId,
         bidId: bid.bidId,
+        traffic: bid.params.traffic || BANNER,
```
ncolletti, line 70: This should be defined with bid.mediaTypes, the 'traffic' param was removed in previous commits and docs were updated. It is fine to leave the default to banner. The tests will need to be updated to remove hardcoding this param as well.

lunamedia: That's just my fault, i reverted to much while fixing native

ncolletti:
LGTM. Would just keep in mind that if a request comes through with multiple mediaTypes the adapter is preferring banner for the 'traffic' param. If you feel you'll run into scenarios with publishers using multiple mediaTypes often, you could then move that back as a params option.

PR #5908, opened 2020-10-29
sharedIdSystem: fix for opted out user

yava-ua:
## Type of change

- [x] Bugfix
- [ ] Feature
- [ ] New bidder adapter   
- [ ] Build related changes
- [ ] CI related changes
- [ ] Other

## Description of change
Scenario: user is opted out and afterwards opted in
Related to issue reported in prebid-sharedid https://github.com/prebid/shared-id/issues/7

@jdwieland8282 @wqi1972

### DIFF
--- a/modules/sharedIdSystem.js
+++ b/modules/sharedIdSystem.js
@@ -26,7 +26,7 @@ const TIME_LEN = 10;
 const RANDOM_LEN = 16;
 
 const id = factory();
-const storage = getStorageManager(null, MODULE_NAME);
+export const storage = getStorageManager(null, MODULE_NAME);
 
 /**
  * Constructs cookie value
--- a/modules/sharedIdSystem.js
+++ b/modules/sharedIdSystem.js
@@ -8,11 +8,15 @@
 import * as utils from '../src/utils.js'
 import {ajax} from '../src/ajax.js';
 import {submodule} from '../src/hook.js';
+import {getStorageManager} from '../src/storageManager.js';
 
 const MODULE_NAME = 'sharedId';
 const ID_SVC = 'https://id.sharedid.org/id';
 const DEFAULT_24_HOURS = 86400;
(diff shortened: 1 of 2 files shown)


yava-ua:
@wqi1972 well, upon checking further I see that this fix has no value. The problem reproduces for the periods between syncs. 
Scenario with comments
1. Visit publisher website    -- shared id set properly, cookie created for publisher domain. syncTime is set to 24 hours max (may be less, configurable by publisher)
2. User is opted out  
3. Visit publisher website. -- user is not synced and is seen by publisher as opted in. It'll remain in that state until next sync

And vica verso, if user is opted in after opt out and is not yet synced it be treated by publisher as opted out until text sync. When setting syncTime to few seconds, changes to opt in/out are picked straight away.

@jdwieland8282 I believe this is the intended behavior. Please confirm. 

Closing this pr, as either way this fix is irrelevant

jdwieland8282:
Ok, so it sounds like the issue in [prebid/shared-id#7](https://github.com/prebid/shared-id/issues/7) was related to the time in between when a user opts out and when that user has their cookie/uid zeroed out? Effectively zeroing out their cookie is the act of signaling to other publishers that this user is opted out. 

@wqi1972 there is a gap between when the user opts out on sharedid.org/optout and when the user's cookie is zeroed out (0000-000-0000). The zeroing out of the users cookie is done on the next sync. Is your concern related to this gap?

wqi1972:
@jdwieland8282 , @yava-ua , sorry about late response on this issue.
I agree with @yava-ua , if the user cookie will be set correctly on next sync, then it is not an issue. When I raised the issue inside prebid/shared-id#7, I didn't notice the cookie sync setting inside PBJS.

I just tested and confirmed on the next sync, a new shared id will be generated. I will close prebid/shared-id#7

jdwieland8282:
perfect, no need to apologize we all getting used to this, we appreciate your comments and help.  yes lets close #7

PR #5923, opened 2020-11-02
rubicon - support all userIds

harpere:
## Type of change

- [ ] Bugfix
- [X ] Feature
- [ ] New bidder adapter   
- [ ] Build related changes
- [ ] CI related changes
- [ ] Other

## Description of change

pass any/all userIds

```
{
  bidder: '<bidder name>',
  params: {
    // ...
  }
}
```

### DIFF
--- a/modules/rubiconBidAdapter.js
+++ b/modules/rubiconBidAdapter.js
@@ -514,7 +514,9 @@ export const spec = {
           } else if (eid.source === 'liveramp.com') {
             data['x_liverampidl'] = eid.uids[0].id;
           } else if (eid.source === 'sharedid.org') {
-            data['eid_sharedid.org'] = `${eid.uids[0].id}^${eid.uids[0].atype}^${eid.uids[0].ext.third}`;
+            data['eid_sharedid.org'] = `${eid.uids[0].id}^${eid.uids[0].atype}^${(eid.uids[0].ext && eid.uids[0].ext.third) || ''}`;
+          } else if (eid.source === 'id5-sync.com') {
+            data['eid_id5-sync.com'] = `${eid.uids[0].id}^${eid.uids[0].atype}^${(eid.ext && eid.ext.linkType) || ''}`;
           } else {
             // add anything else with this generic format
             data[`eid_${eid.source}`] = `${eid.uids[0].id}^${eid.uids[0].atype || ''}`;
--- a/modules/rubiconBidAdapter.js
+++ b/modules/rubiconBidAdapter.js
@@ -516,15 +516,15 @@ export const spec = {
           } else if (eid.source === 'sharedid.org') {
             data['eid_sharedid.org'] = `${eid.uids[0].id}^${eid.uids[0].atype}^${eid.uids[0].ext.third}`;
           } else {
-            // look for the first eid.uids[*].ext.stype === 'ppuid'
+            // add anything else with this generic format
+            data[`eid_${eid.source}`] = `${eid.uids[0].id}^${eid.uids[0].atype || ''}`;
+          }
+          // send AE "ppuid" signal if exists, and hasn't already been sent
+          if (!data['ppuid']) {
+            // get the first eid.uids[*].ext.stype === 'ppuid', if one exists
             const ppId = find(eid.uids, uid => uid.ext && uid.ext.stype === 'ppuid');
-            // if found ppId and no configUserId already found, then add it
-            if (ppId && ppId.id && !configUserId) {
+            if (ppId && ppId.id) {
               data['ppuid'] = ppId.id;
-            } else {
-              // add anything else with this generic format
-              const atype = eid.uids[0].atype;
-              data[`eid_${eid.source}`] = `${eid.uids[0].id}^${atype || ''}`;
             }
           }
         } catch (e) {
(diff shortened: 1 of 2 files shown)


modules/rubiconBidAdapter.js
```diff
@@ -238,15 +238,6 @@ export const spec = {
       const eids = utils.deepAccess(bidderRequest, 'bids.0.userIdAsEids');
       if (eids && eids.length) {
         utils.deepSetValue(data, 'user.ext.eids', eids);
-
-        // liveintent requires additional props to be set
-        const liveIntentEid = find(data.user.ext.eids, eid => eid.source === 'liveintent.com');
-        if (liveIntentEid) {
-          utils.deepSetValue(data, 'user.ext.tpid', { source: liveIntentEid.source, uid: liveIntentEid.uids[0].id });
-          if (liveIntentEid.ext && liveIntentEid.ext.segments) {
-            utils.deepSetValue(data, 'rp.target.LIseg', liveIntentEid.ext.segments);
```
robertrmartinez, line 247: we do not need to send this now?

modules/rubiconBidAdapter.js
```diff
@@ -503,51 +492,45 @@ export const spec = {
...
+            if (eid.ext && Array.isArray(eid.ext.segments) && eid.ext.segments.length) {
+              data['tg_v.LIseg'] = eid.ext.segments.join(',');
+            }
+          } else if (eid.source === 'liveramp.com') {
+            data['x_liverampidl'] = eid.uids[0].id;
+          } else if (eid.source === 'sharedid.org') {
+            data['eid_sharedid.org'] = `${eid.uids[0].id}^${eid.uids[0].atype}^${eid.uids[0].ext.third}`;
+          } else {
+            // look for the first eid.uids[*].ext.stype === 'ppuid'
+            const ppId = find(eid.uids, uid => uid.ext && uid.ext.stype === 'ppuid');
```
robertrmartinez, line 520: should we check `! configUserId` so we do not loop through these `eid.uids` unnecessarily?

modules/rubiconBidAdapter.js
```diff
@@ -503,51 +492,45 @@ export const spec = {
...
+            // if found ppId and no configUserId already found, then add it
+            if (ppId && ppId.id && !configUserId) {
+              data['ppuid'] = ppId.id;
+            } else {
+              // add anything else with this generic format
+              const atype = eid.uids[0].atype;
+              data[`eid_${eid.source}`] = `${eid.uids[0].id}^${atype || ''}`;
```
robertrmartinez, line 527: does it make sense to just use `eid.uids[0].atype` instead of allocating the var? Not sure what this looks like after minification, probably fine.

robertrmartinez (approved):
Looks good to me.

Just to small comments, not really necessary.

modules/rubiconBidAdapter.js
harpere: no, not according to the ticket.

modules/rubiconBidAdapter.js
harpere: yes, that was an artifact of some more complicated logic.  will change it.

modules/rubiconBidAdapter.js
harpere: seeking clarification on the expected logic.

robertrmartinez (approved):
👍

modules/rubiconBidAdapter.js
```diff
@@ -503,51 +492,45 @@ export const spec = {
...
+            data['eid_liveintent.com'] = eid.uids[0].id;
+            if (eid.ext && Array.isArray(eid.ext.segments) && eid.ext.segments.length) {
+              data['tg_v.LIseg'] = eid.ext.segments.join(',');
+            }
+          } else if (eid.source === 'liveramp.com') {
+            data['x_liverampidl'] = eid.uids[0].id;
+          } else if (eid.source === 'sharedid.org') {
+            data['eid_sharedid.org'] = `${eid.uids[0].id}^${eid.uids[0].atype}^${eid.uids[0].ext.third}`;
+          } else {
+            // add anything else with this generic format
```
smenzer, line 519: just a note that ID5 module includes an `ext` that contains additional metadata about the ID (see https://github.com/prebid/Prebid.js/blob/master/modules/userId/eids.md). Currently, we are including `linkType` which signals the method that we used to link an ID across domains, something that buyers have asked to be included in bid requests to them, but we may add more metadata later that should be passed through.

@jdwieland8282 this is roughly equivalent to the `stype` conversation happening related to pub provided ids, so feels important to be included here.

harpere: Thanks @smenzer.  Need to see if our endpoint supports this.  If not, may do this in another PR after we support it.

modules/rubiconBidAdapter.js
harpere: updated the logic here.

modules/rubiconBidAdapter.js
harpere: thanks @smenzer - will see if our endpoint supports this.  if not, will make the change in a separate PR after we support it.

harpere: made the update @smenzer

robertrmartinez (approved):
Updates look good.

Thanks @harpere 

And thanks @smenzer for pointing that out!

modules/rubiconBidAdapter.js
smenzer: looks good, thanks!

PR #5925, opened 2020-11-03
Add User ID Targeting to googletag.cmd as a fallback when GPT API is not ready

lyubomirshishkov:
## Type of change 
- [x] Feature

## Description of change
Currently, when the `userIdTargeting` module is included and `SHARE_WITH_GAM` is set to `true`, the module checks if the GPT API is ready. If it isn't, then the value of `SHARE_WITH_GAM` is changed to `false` and User IDs are not sent to GAM.

The proposed change handles the cases in which the GPT API is not ready, by adding a fallback, which defines `googletag` and `googletag.cmd`, and pushes the `pubads().setTargeting()` function to `googletag.cmd`. The purpose of the change is to allow the `userIdTargeting` module to function even when the GPT API isn't ready and/or when `googletag` has not been defined yet.

### DIFF
--- a/modules/userIdTargeting.js
+++ b/modules/userIdTargeting.js
@@ -26,9 +26,9 @@ export function userIdTargeting(userIds, config) {
     window.googletag = window.googletag || {};
     googletag.cmd = googletag.cmd || [];
     GAM_API = function (key, value) {
-        googletag.cmd.push(function () {
-            googletag.pubads().setTargeting(key, value);
-        });
+      googletag.cmd.push(function () {
+        googletag.pubads().setTargeting(key, value);
+      });
     };
   }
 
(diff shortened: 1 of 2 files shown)


lyubomirshishkov:
LGTM

smenzer:
The code looks good to me, but could you please add some tests before I merge?

lyubomirshishkov:
@smenzer since my changes do not affect the logic of the `userIds.forEach` function (which is already covered by unit tests), I only added one unit test for the fallback. Let me know if that's ok.

I also added a more relevant expect to the "Publisher does not want to pass an id" test case.

smenzer (approved):
LGTM, thanks for adding the test

PR #5856, opened 2020-10-13
IX missing sizes testing and diagnosis

ix-certification:
## Type of change

- [x ] Bugfix
- [x ] Feature
- [ ] New bidder adapter   
- [ ] Code style update (formatting, local variables)
- [ ] Refactoring (no functional changes, no api changes)
- [ ] Build related changes
- [ ] CI related changes
- [ ] Does this change affect user-facing APIs or examples documented on http://prebid.org?
- [ ] Other

### 1. **detectMissingSizes** feature toggle
Can set the value of new optional param detectMissingSizes (default true) by :
```
pbjs.setBidderConfig({
                bidders: ["ix"],
                config: {
                    ix: {
                        detectMissingSizes: false
                    }
                }
            });
```
### 2. Splitting large requests into multiple parts
Large URLs in the calls to index exchange, will be split into multiple parts so as to not exceed URL length limits and get blocked.

### 3. **ixdiag** : Internal metrics for missing sizes data
Since our last update, IX bid adapter is bidding on all banner sizes available in an ad unit, to gain some insight into the inventory available/missing we are adding a property `ixdiag` in the `payload.r.ext` object, and populating it with required stats.

### Docs Repository PR
https://github.com/prebid/prebid.github.io/pull/2455

PR #5926, opened 2020-11-03
Remove 'only' to run all tests

Fawke:
## Type of change

- [X] CI Related
- [X] Other

## Description of change

Removes `only` from [this](https://github.com/ix-prebid-support/Prebid.js/blob/ac924b08655d8375cc177cccbf837fec8ecd5eb9/test/spec/modules/ixBidAdapter_spec.js) test file so all unit tests can run. Currently, only the tests in that file are running.

Introduced with the merge of https://github.com/prebid/Prebid.js/pull/5856


smenzer (approved):
LGTM

PR #5930, opened 2020-11-04
Fix request flow

relaido:
## Type of change

- [ ] Bugfix
- [x] Feature
- [ ] New bidder adapter   
- [ ] Build related changes
- [ ] CI related changes
- [ ] Other

## Description of change
Fix to always request the VAST API whether case Android and IOS.
Also, fix to save uuid to cookie instead of local storage.


test/spec/modules/relaidoBidAdapter_spec.js
```diff
@@ -11,6 +11,7 @@ const setUAMobile = () => { window.navigator.__defineGetter__('userAgent', funct
 
 describe('RelaidoAdapter', function () {
   const relaido_uuid = 'hogehoge';
+  window.document.cookie = `${UUID_KEY}=${relaido_uuid}`
```
robertrmartinez, line 14: please do not use direct cookie setting in your tests.

Unit tests are run across all browsers and this causes flaky results at times.

Instead, you should stub the utils.getCookie and utils.setCookie functions to perform whatever you want.

robertrmartinez (changes requested):
looks fine,

But please update the tests to stub the storage functions instead of directly setting on window.

test/spec/modules/relaidoBidAdapter_spec.js
relaido: Thank you for the review.

For various reasons, close this pull request

I will send you a new pull request that includes this one immediately.

I'll fix the test code at that time.

relaido:
Due to some content that arose that we wanted to add to the site as soon as possible

Issue #3842, opened 2019-05-21
Unit Test: How to test ajax calls


couchcrew-thomas:
## Description
I want to test that calls to `src/ajax` are called with the correct parameters for my custom adapter. The ajax request is used in the `onTimeout` and `onBidWon` callback. Because the ajax method is an exported function, it is not possible to stub it with sinon.

### Example Adapter Spec
```
export const spec = {
  /* ... */
  onTimeout: function (timeoutData) {
    ajax('http://localhost:3000/onTimeout', null, JSON.stringify(timeoutData), {
      withCredentials: true,
      method: 'POST',
      contentType: 'application/json'
    })
  }
};
```

### Intended Unit Test
```
import * as xhr from '../../../src/ajax';

/* ... */
let stub = sinon.stub(xhr, "ajax");
spec.onTimeout(something);
expect(stub.called).to.be.true;
```

But the above does not work because sinon cannot spy on exported functions.

PR #5941, opened 2020-11-09
sspBC adapter: update to v4.6 (notifications, user sync)

wojciech-bialy-wpm:
## Type of change

- [x] Feature

## Description of change

Update sspBC adapter to v4.6:
- add notification endpoint
- send bidWon and onTimeout notifications
- send CMP version to user sync endpoint


robertrmartinez:
please add unit tests @wojciech-bialy-wpm

osazos:
Hi  @wojciech-bialy-wpm,

In order to approve your PR, could you please 
- add tests to cover onBidWon and onTimeout events, without these tests, your test-coverage is under 80%.
- fix your CI errors (Lint)
- fix your failing test : ✖ should provide correct url, if frame sync is allowed

In your `onTimeout`, `onBidWon` functions, you expect `bid.params` is an array, but in `isBidRequestValid` it is an object. shouldn't be both the same? 
If it is what you expect, you should add in your condition: `bid.params && bid.params.length ? …  : ''` to avoid error if a params array is empty.

wojciech-bialy-wpm:
Sorry for the delay.

Tests updated:
- getUserSyncs test fixed (failed when adapter started appending tcf version to the url)
- onBidWon test added
- onTimeout test added

In the last two cases, while I can't get sinon to work with Prebid.js ajax module (similar case as in https://github.com/prebid/Prebid.js/issues/3842) methods now check whether notification payload is correctly constructed.

osazos:
@wojciech-bialy-wpm, where is your updated test file please? I can't see it in your the files changed of this PR.

wojciech-bialy-wpm:
@osazos Oh, I see the last commit (with changes to test/spec/modules/sspBCAdapter_spec.js) did not go through. Updating now.

osazos (approved):
LGTM

Issue #5942, opened 2020-11-09
_gd cookie is created multiple times causing critical 400 error


mercuryyy:
Prebid 4.15, built with User ID modules are adding a _gdxxxxxx cookie each sessions.
(i only checked with User ID PubCommon ID)

This can cause a 400 error on nginx easily for pages with a lot of pageviews, the session cookie is not behaving right and the _gd cookie is being creating every page view causing multiple _dg cookies being created over and over again.

They are session cookie so once browser is closed they are deleted but if the user goes through enough pages it will cause a global 400 error which will actually block access to the entire website.

I would assume the _gd cookie should be created only once.

I also build prebid without gdpr module and only with User ID PubCommon ID and it will still create the _gd cookies.

smenzer:
It seems that this is not specifically related to PubCommonId module, as I see it happening on a test page with just the ID5 module activated. so likely, it's somewhere in core userId module code or in the way storagemanager works with userid

smenzer:
I take that back...if I build pbjs without PubCommonId module, this doesn't happen. I think I found the offending code here, although it's not immediately clear to me why it's executed when I do not include the PubCommonId Module in my config:
https://github.com/prebid/Prebid.js/blob/master/modules/pubCommonIdSystem.js#L289

We might need to add this before returning on line 299?
```
storage.setCookie(cookieName, '', 'Thu, 01 Jan 1970 00:00:01 GMT', undefined, nextDomain);
```

smenzer:
@jsnellbaker seems to have made the last update to this bit of code and @Fawke reviewed, so maybe one of them could add some more color here?

smenzer:
i think this is because of https://github.com/prebid/Prebid.js/blob/master/modules/pubCommonIdSystem.js#L305 being called every time the module is registered, rather than only when needed by the submodule here https://github.com/prebid/Prebid.js/blob/master/modules/pubCommonIdSystem.js#L40

pycnvr:
@smenzer This pique my interest.  Your PR will definitely alleviate the immediate issue.  I suspect that the root problem is that setCookie is async now, and readCookie will almost always fails.  So the function domainOverride won't correctly find the top-level domain but instead returns undefined most of the time.

pycnvr:
Nope.  I was wrong.  This weird behavior is only true if called during module load.   Scott's PR is correct.

PR #5943, opened 2020-11-09
delete pubcommon test cookie for domainOverride after writing it in all cases

smenzer:
## Type of change

- [x] Bugfix

## Description of change
As described in https://github.com/prebid/Prebid.js/issues/5942, pubcommon module is setting a domain override variable every time the module is registered (i.e. if pbjs is built with pubcommon, regardless of whether a pub activates it through configuration). There is a cookie check that happens by writing a test cookie and then if it's found, it would delete and try another domain. if it wasn't found, it assumed it didn't stick, but in fact it was leaving session cookies around. this PR will delete the test cookies regardless of condition.

I think we should also consider moving the line `const COOKIE_DOMAIN = pubCommonIdSubmodule.domainOverride();` INSIDE of the module code so it's only every run if the module is activated, as I'm sure this is adding overhead to pages that have not configured pubcommon but have built it with it.

## Other information
https://github.com/prebid/Prebid.js/issues/5942 (thanks to @mercuryyy for finding this)

### DIFF
--- a/modules/lunamediahbBidAdapter.js
+++ b/modules/lunamediahbBidAdapter.js
@@ -70,7 +70,7 @@ export const spec = {
         schain: bid.schain || {},
       };
       const mediaType = bid.mediaTypes
-      
+
       if (mediaType && mediaType[BANNER] && mediaType[BANNER].sizes) {
         placement.sizes = mediaType[BANNER].sizes;
         placement.traffic = BANNER;
--- a/modules/pubCommonIdSystem.js
+++ b/modules/pubCommonIdSystem.js
@@ -282,16 +282,19 @@ export const pubCommonIdSubmodule = {
   domainOverride: function () {
     const domainElements = document.domain.split('.');
     const cookieName = `_gd${Date.now()}`;
-    for (let i = 0, topDomain; i < domainElements.length; i++) {
+    for (let i = 0, topDomain, testCookie; i < domainElements.length; i++) {
       const nextDomain = domainElements.slice(i).join('.');
 
       // write test cookie
       storage.setCookie(cookieName, '1', undefined, undefined, nextDomain);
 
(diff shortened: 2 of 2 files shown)


smenzer:
+ @pycnvr for thoughts on moving the cookie_domain line inside the module itself

pycnvr:
@smenzer Yeah, sounds good to me.  Nice find.

pycnvr:
I've also been wondering if cookie_domain can be returned in domainOverride if it's available already.

smenzer:
> I've also been wondering if cookie_domain can be returned in domainOverride if it's available already.

@pycnvr  so basically, something like this?

```
  domainOverride: function () {
    if (COOKIE_DOMAIN) {
       return COOKIE_DOMAIN;
    }

    const domainElements = document.domain.split('.');
    ...
```

pycnvr:
@smenzer Yes.

smenzer:
ok...i updated the code to only check for cookies once. can you please check this carefully @pycnvr to make sure it looks ok to you (and that the way i wrote it makes sense)?

smenzer:
i was getting all kinds of errors in the tests when trying to reuse the same value each time, so i reverted those changes and now it will only call domainOverride when it needs to, rather than when the module is registered. if you want to re-use the value like we mentioned above, i'll leave that for you to pick up in another PR, seems too big for this and i'd like to get a fix out quickly (and i don't have time to fix the tests/other code that would need it)

pycnvr:
Ok, sounds good.  Thanks for looking into it.

pycnvr:
Latest set of changes looks good, too.

Fawke (approved):
LGTM

smenzer:
with the approval from @Fawke and @pycnvr i'm going to go ahead and merge this

PR #5951, opened 2020-11-10
Fix request size validate

relaido:
## Type of change

- [ ] Bugfix
- [x] Feature
- [ ] New bidder adapter   
- [ ] Build related changes
- [ ] CI related changes
- [ ] Other

## Description of change
Fix to validate the video size, not just the banner.
And always request the VAST API whether case Android and IOS.
Also, fix to save uuid to cookie instead of local storage.

I wrote the test code with stubs, taking the reviewers' opinions into account.
https://github.com/prebid/Prebid.js/compare/master...relaido:fix-request-size-validate?expand=1


Fawke (approved):
LGTM

PR #5954, opened 2020-11-11
Add Automatic Release Drafter Functionality to Prebid Repository

GLStephen:
This PR adds release drafter functionality to the project. With this functionality enabled all merges into master will generate a new draft release based on semver and categorized by the label. This will improve accuracy of release notes and reduce process steps required for Prebid code.

More info on Release Drafter is here:

https://github.com/release-drafter/release-drafter

If this PR is accepted then labels will need to be created for bugfix, fix, bug, feature, maintenance and the semver levels major, minor, patch. These labels will need to be assigned prior to the merge to properly categorize changes and manage semver versions.

### DIFF
--- a/.github/release-drafter.yml
+++ b/.github/release-drafter.yml
@@ -4,7 +4,7 @@ tag-template: '$RESOLVED_VERSION'
 categories:
   - title: '🚀 New Features'
     label: 'feature'
-  - title: '🧰 Maintenance'
+  - title: '🛠 Maintenance'
     label: 'chore'    
   - title: '🐛 Bug Fixes'
     labels:
--- a/.github/release-drafter.yml
+++ b/.github/release-drafter.yml
@@ -5,7 +5,7 @@ categories:
   - title: '🚀 New Features'
     label: 'feature'
   - title: '🛠 Maintenance'
-    label: 'chore'    
+    label: 'maintenance'    
   - title: '🐛 Bug Fixes'
     labels:
       - 'fix'
--- a/.github/release-drafter.yml
+++ b/.github/release-drafter.yml
@@ -1,0 +1,28 @@
+
+name-template: 'Prebid $RESOLVED_VERSION Release'
+tag-template: '$RESOLVED_VERSION'
+categories:
+  - title: '🚀 New Features'
+    label: 'feature'
+  - title: '🧰 Maintenance'
+    label: 'chore'    
+  - title: '🐛 Bug Fixes'
+    labels:
+      - 'fix'
+      - 'bugfix'
+      - 'bug'      
+change-template: '- $TITLE (#$NUMBER)'
+version-resolver:
+  major:
+    labels:
+      - 'major'
+  minor:
+    labels:
+      - 'minor'
+  patch:
+    labels:
+      - 'patch'
+  default: patch
+template: |
+  ## In This Release
+  $CHANGES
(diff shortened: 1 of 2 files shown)


smenzer:
I'll just say, this sounds really cool! I like the idea of just assigning labels before merge rather than having to go to edit the release notes.

.github/release-drafter.yml
```diff
@@ -0,0 +1,28 @@
+
+name-template: 'Prebid $RESOLVED_VERSION Release'
+tag-template: '$RESOLVED_VERSION'
+categories:
+  - title: '🚀 New Features'
+    label: 'feature'
+  - title: '🧰 Maintenance'
```
smenzer, line 7: for consistency with old release notes, should we use 🛠? I personally like the icon you picked, but just wondering if we care about keeping it the same or not

.github/release-drafter.yml
```diff
@@ -0,0 +1,28 @@
...
+    label: 'chore'
```
smenzer, line 8: a lot of the PRs that we end up putting in "Maintenance" are adapters/modules making small modifications to their code. Yes, you could argue it's a "chore", but that has a negative connotation. Could we instead (or in addition) just have a label "maintenance"?

Also...should we scope these labels at all so it's clear that these are for release drafter? something like `release-feature`, `release-bugfix` or something like that?

.github/release-drafter.yml
GLStephen: I just missed that. My goal was to match. I'll update it.

.github/release-drafter.yml
GLStephen: Using maintenance sounds good and I agree the "chore" label has negative connotations. This is the time to pick our choice of words. I'll change it. 

In general though, I think that the labels are useful broadly and would prefer not to scope them. Scoping the labels would imply that is their only role. If the PR is a feature, then it's a feature and while that is metadata for release drafter it should/could be used without release drafter for other purposes. In addition, RD can pull in any labels we choose into any number of buckets. I would think of RD conforming to our labeling scheme that we use to label the type of PR, not the other direction.

If we really do want to scope them, then the verbosity of the label is important. I would prefer to frontload the wording so if they get truncated in a UI the more differentiating part is shown first instead of "release-...." over and over it would read "feature-...". It's unlikely, but possible.

smenzer: Makes total sense on not scoping the labels.

GLStephen:
@smenzer merging and label creation after the release today seems sensible?

smenzer:
Agreed it should happen after today's release. We should chat about it in the PBJS meeting today to make sure everyone is in board (cc @gglas)

GLStephen:
@GLStephen add instructions/documentation to changes

smenzer:
@gglas what's the timeline for rolling this out? we ended up missing a meeting b/c of thanksgiving, but can we get this rolled out after the release this week so we can start using it? who else needs to be notified about this to manage it? i will be on duty in a week or two so happy to be a guinea pig if necessary

gglas:
let's collaborate in the channel and see if we can roll out next week!  i
will ping the channel and discuss with next weeks reviewer -- @GLStephen
there might be a request for some aid in the first go.

smenzer:
can we merge this now? @GLStephen @mike-chowla @gglas

smenzer (approved):
LGTM

Fawke (approved):
LGTM

PR #6016, opened 2020-11-23
LiveConnect, lighter-weight module

jankoulaga:
## Type of change
- [ ] Bugfix
- [X] Feature
- [ ] New bidder adapter 
- [X] Build related changes
- [ ] CI related changes
- [ ] Other

## Description of change
The new LiveConnect 2.0.0 should support two modes, standard and minimal, where the minimal version should result in a smaller bundle size, as it only pulls in the ID resolution functionality. The main premise here is that the env variable will be evaluated to `false` and then webpack will ensure that the redundant code is removed.

**Once the approach and content is ok, i will gladly update the docs on how to build it.**

The benefits of running this setup are:
- When running `LiveConnectMode=minimal gulp build --modules=liveIntentIdSystem` the bundle is 159K ungzipped
- When running plain build, `gulp build --modules=liveIntentIdSystem` the bundle is 173K ungzipped.


patmmccann:
Flagging as interesting to @smenzer @bretg @gglas 

I really like this build time option proposal. How would we detail it on the download page or in other places? @jankoulaga is there a documentation pr you can link?

patmmccann:
I think you may want to remove your package.json changes also

jankoulaga:
> Flagging as interesting to @smenzer @bretg @gglas
> 
> I really like this build time option proposal. How would we detail it on the download page or in other places? @jankoulaga is there a documentation pr you can link? I think it makes sense to assume this will be welcomed

Hey @patmmccann definitely, i'd be more than happy to explain the approach, even though it's not a new concept. As i said in the PR description, if this approach is fine with the community, i will update the docs and submit an additional PR.

bretg:
This is an interesting feature for command-line builders.

It can't be supported on the download page because the builds happen exactly once for each PBJS release. It takes several minutes for each build. What happens when using the page is that the alread-built modules are slapped together and returned to the user. 

We're not going to pre-build all possible combinations of `LiveConnectMode=minimal`, `Criteo=minimal`, etc.

modules/liveIntentIdSystem.js
```diff
@@ -8,9 +8,10 @@ import * as utils from '../src/utils.js';
 import { triggerPixel } from '../src/utils.js';
 import { ajaxBuilder } from '../src/ajax.js';
 import { submodule } from '../src/hook.js';
-import { LiveConnect } from 'live-connect-js/cjs/live-connect.js';
+import { LiveConnect } from 'live-connect-js/esm/initializer.js';
 import { gdprDataHandler, uspDataHandler } from '../src/adapterManager.js';
 import { getStorageManager } from '../src/storageManager.js';
+import { MinimalLiveConnect } from 'live-connect-js/esm/minimal-live-connect.js';
```
msm0504, line 14: Will this result in both libraries (LiveConnect and MinimalLiveConnect) being added to the bundle?

jankoulaga: Hey @msm0504 well, the bundle is generated by webpack, so no.

If this were pieced together without any treeshaking, then yes, both sources would have been added, however, the trick here is that [at some point](https://github.com/prebid/Prebid.js/pull/6016/files#diff-489289275f4ee5264024b54c9c851e6c57684fd2d0bccab91d8b9ccef150e235R125), there's an evaluation of a boolean which is either true or false, and that's how webpack will know which import will be used. Anything unused will be automatically removed from the bundle.

msm0504 (approved):

msm0504:
@jsnellbaker I've approved these changes, but I'm not sure if this would normally require a 2nd review. Either way, I'd like a reviewer familiar with the build process to look at this if one hasn't already.

jsnellbaker (approved):
Hi,

I think this looks okay.  @jankoulaga please put together the needed docs and we can merge.

jankoulaga:
@jsnellbaker here's the doc update:
https://github.com/prebid/prebid.github.io/pull/2607

PR #6027, opened 2020-11-24
Pass user ids in bid request

MenelikTucker-districtm:
## Type of change
- [x] Bugfix

## Description of change
Only clear user ids if GDPR applies


osazos (approved):
LGTM

PR #6032, opened 2020-11-25
ID5 ID module - pass gdpr and usp parameters in body instead of querystring

smenzer:
## Type of change

## Description of change
Since gdpr consent strings are getting long, we want to start passing them in the body of the POST rather than the querystring to avoid data loss


FilipStamenkovic (approved):
LGTM

Issue #3852, opened 2019-05-22
Proposal: Standardizing COPPA


bretg:
## Type of Issue
Feature

## Description
Several bidders have ways for prebid users to pass the COPPA flag, but Prebid.js hasn't specified a standard way to specify it such that all bid adapters can access it.

## Proposal

This doesn't fit into the First Party Data `context` framework because the COPPA flag is always public and should be known to all adapters, while other first party data might be limited to a certain set of bidders.

So the proposal is

1) Declare a configuration convention for the publisher to signal PBJS about COPPA content:
```
pbjs.setConfig({coppa: true});
```
2) Each adapter supporting COPPA would then be updated to call `getConfig({coppa})`

3) The Prebid Server Bid Adapter will be updated to pass this value to PBS as regs.coppa=1 when true.

4) Prebid Server will pass regs.coppa to adapters and will always (not configurable) cause the PBS core to enforce COPPA rounding of IP address and removal of ID as suggested in the OpenRTB standard:

- device.ifa, device.macsha1, device.macmd5, device.dpidsha1, device.dpidmd5, device.didsha1, device.didmd5
- Truncate ip field - remove lowest 8 bits.
- Truncate ipv6 field - remove lowest 32 bits.
- Remove geo.lat, geo.lon. geo.metro, geo.city, and geo.zip
- Remove user.id, user.buyeruid, user.yob, and user.gender

5) Update prebid.org with a page that dynamically lists the COPPA-supporting adapters based on the .md file similar to what we do for GDPR.

6) Define how COPPA flag gets sent in the SDK

bretg:
This was discussed in the Prebid.js committee meeting today. Rubicon will put it into our next sprint.

bretg:
Reviewed with Eve. We're planning on moving forward with this.

bretg:
Based on feedback from code inspection, we've changed the design back to having the bid adapters each call `getConfig` instead of having core put it on the bidrequest object.

bretg:
Released

PR #6033, opened 2020-11-25
LKQD: update adapter to include new parameters

ndxcarver:
## Type of change

- [ ] Bugfix
- [x] Feature
- [ ] New bidder adapter   
- [ ] Build related changes
- [ ] CI related changes
- [ ] Other

## Description of change

Add accepted parameters c1...c40 and coppa.

```
{
  bidder: '<bidder name>',
  params: {
    // ...
  }
}
```

- contact email of the adapter’s maintainer devops@nexstardigital.com

### DIFF
--- a/modules/lkqdBidAdapter.js
+++ b/modules/lkqdBidAdapter.js
@@ -1,6 +1,7 @@
 import * as utils from '../src/utils.js';
 import { registerBidder } from '../src/adapters/bidderFactory.js';
 import { VIDEO } from '../src/mediaTypes.js';
+import { config } from '../src/config.js';
 
 const BIDDER_CODE = 'lkqd';
 const BID_TTL_DEFAULT = 300;
@@ -148,8 +149,8 @@ function buildRequests(validBidRequests, bidderRequest) {
       if (bidRequest.params.hasOwnProperty('dnt') && bidRequest.params.dnt != null) {
         sspData.dnt = bidRequest.params.dnt;
       }
-      if (bidRequest.params.hasOwnProperty('coppa') && bidRequest.params.coppa != null) {
-        sspData.coppa = bidRequest.params.coppa;
+      if (config.getConfig('coppa') === true) {
+        sspData.coppa = true;
       }
       if (bidRequest.params.hasOwnProperty('pageurl') && bidRequest.params.pageurl != null) {
         sspData.pageurl = bidRequest.params.pageurl;
--- a/modules/lkqdBidAdapter.js
+++ b/modules/lkqdBidAdapter.js
@@ -150,7 +150,7 @@ function buildRequests(validBidRequests, bidderRequest) {
         sspData.dnt = bidRequest.params.dnt;
       }
       if (config.getConfig('coppa') === true) {
-        sspData.coppa = true;
+        sspData.coppa = 1;
       }
       if (bidRequest.params.hasOwnProperty('pageurl') && bidRequest.params.pageurl != null) {
         sspData.pageurl = bidRequest.params.pageurl;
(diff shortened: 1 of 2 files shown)


robertrmartinez:
Hi @ndxcarver 

Prebid has a standard way of passing through the coppa flag.

https://github.com/prebid/Prebid.js/issues/3852

You can use the following bit of code to pass it through:

```javascript
const coppa = config.getConfig('coppa');
if (coppa != null) {
   sspData.coppa = config.getConfig('coppa');
}
```
many ways to do so,

But if you only want to pass it if it is true, perhaps:

```javascript
if (config.getConfig('coppa') === true) {
  sspData.coppa = true;
}
```

modules/lkqdBidAdapter.js
```diff
@@ -177,6 +181,12 @@ function buildRequests(validBidRequests, bidderRequest) {
       sspData.bidWidth = playerWidth;
       sspData.bidHeight = playerHeight;
 
+      for (let k = 1; k <= 40; k++) {
```
robertrmartinez, line 184: I do not think this is the most efficient way to do this. It does not seem likely that you will have all 40 or even close to that of these in each bidRequest.

I would think simply looping over each param and adding it is the right thing to do.

```javascript
Object.keys(bidRequest.params).forEach(param => {
   if (/c\d{1,2}/.test(param)){
     sspData[param] = bidRequest.params[param];
   }
});
```

With the above code, we only loop as many times as needed.

robertrmartinez (changes requested):

modules/lkqdBidAdapter.js
ndxcarver: We do not require **c** numbered data to be sent in sequential order and is passed along only if it exists. Your code would also iterate over all the parameters passed in and not our selectively targeted version. My for loop only looks for c1...c40 because all other known accepted data has been picked out.

robertrmartinez:
Right, but it is not common for a bid adapter to take in 40+ bidParams.

Looks like 2 params are `required` for your adapter.

So in what seems to be the _usual_ case, it seems as though looping over the params which the pub set is more efficient on average.

Unless your bidder is unique and you make publishers set 40+ unique `LKQD` params each time.

Another option is to filter the params looking for the `c` params, then add those.

Also, regarding the logic of a bunch of if / else's for each bid param. Looks like each bid param name directly maps 1:1 to the associated SSPData name as well.

So why can we just not loop over bid params and just add them?

modules/lkqdBidAdapter.js
robertrmartinez: Right, but it is not common for a bid adapter to take in 40+ bidParams.

Looks like 2 params are required for your adapter.

So in what seems to be the usual case, it seems as though looping over the params which the pub set is more efficient on average.

Unless your bidder is unique and you make publishers set 40+ unique LKQD params each time.

Another option is to filter the params looking for the c params, then add those.

Also, regarding the logic of a bunch of if / else's for each bid param. Looks like each bid param name directly maps 1:1 to the associated SSPData name as well.

So why can we just not loop over bid params and just add them?

ndxcarver: At this time the correction of my predecessor's code regarding the if statements are out of scope and of course needs to be done. For this particular task, I have to minimize changes to the addition of **coppa** and **c1...c40**. So this is why I chose the for loop that limits the count to 40 instead of doing the generic loop, that is something that can be adopted in a future revision.

For some context on the c data, it is custom data that we allow publisher partners to put into our system and pass along. For better or worse they can put whatever string they want. [Custom Parameters Doc](https://wiki.lkqd.com/display/GEN/Custom+Parameters)

robertrmartinez: Ok sounds good.

Not meaning to be too combative. Just wanted to double check!

robertrmartinez (approved):

modules/lkqdBidAdapter.js
ndxcarver: No worries, constructive help is always welcomed and I will try to do my part and explain as best I can. This was my first encounter with this code as a result of client requests, the code base along with its companion test is now on my growing list of things to change/improve. Have a good one!

Issue #5958, opened 2020-11-12
Utils.isEmpty or Telaria bidder bug?


grajzer:
## Type of issue
Bug

## Description
utils.isEmpty() function will always return true if integer is passed. isEmpty(1) returns true.

If that's an expected behavior, than there's a bug in telariaBidAdapter and the way it constructs schain string -- hp value 1 will always translate to empty string.

bretg:
The first line of `isEmpty` says that if the argument isn't an object, then it's considered empty.
```
if (!object) return true;
```

What schain are you having trouble with in Telaria? I made a test page and see the chain being passed in the URL:

```
                pbjs.setConfig({
                  "schain": {
                    "validation": "strict",
                    "config": {
                      "ver":"1.0",
                      "complete": 1,
                      "nodes": [
                        {
                          "asi":"indirectseller.com",
                          "sid":"00001",
                          "hp":1
                        }
                      ]
                    }
                  }
```
Winds up as: &schain=1.0,1!indirectseller.com,00001,,,,

grajzer:
Hi @bretg , thanks for taking a look. Looking at this part of the code:

https://github.com/prebid/Prebid.js/blob/678ffdfd69a666570dab4738c8b665b686a16458/modules/telariaBidAdapter.js#L148

I think that expected request should look like this:

&schain=1.0,1!indirectseller.com,00001,1,,,

Some publishers noticed and asked why hp value is not passed...

bretg:
Adding @telariaEng 

This seems odd because node.sid is a string (not an object) and it's showing up.

bretg:
Closing since the PR has been merged.

PR #6036, opened 2020-11-26
Not using utils.isEmpty on non objects

telariaEng:
## Type of change

- [x] Bugfix

## Description of change

Not using `utils.isEmpty` for emptiness checks on primitives. 

https://github.com/prebid/Prebid.js/issues/5958


modules/telariaBidAdapter.js
```diff
@@ -124,7 +124,7 @@ function getDefaultSrcPageUrl() {
 }
 
 function getEncodedValIfNotEmpty(val) {
-  return !utils.isEmpty(val) ? encodeURIComponent(val) : '';
+  return (val !== '') ? encodeURIComponent(val) : '';
```
aleksatr, line 127: Hi @telariaEng , I assume that you're checking for `val !== ''` to avoid an unnecessary call to `encodeURIComponent`, because `encodeURIComponent('') === ''`.
What `val !== ''` lets through is `undefined`, so if some of the node properties are undefined, they will end up in your encoded schain string as string `undefined`. I'm just checking if this is intended behavior?

telariaEng: Thank you for your response @aleksatr . I was actually avoiding not calling `encodeURIComponent(val)` for val = 0. Maybe I should just change it to `val !== undefined` ?

aleksatr:
LGTM

PR #6039, opened 2020-11-26
Add Zemanta adapter

rokostik:
## Type of change

- [x] New bidder adapter   
## Description of change

This PR adds the Zemanta bidding adapter.

```
{
  bidder: 'zemanta',
  params: {
   publisher: {
     id: '2706',
     name: 'Publishers Name',
     domain: 'publisher.com'
   },
   tagid: 'tag-id'
 }
}
```
NOTE: we do not expose our bidding and usersync URLs. When a publisher will want to integrate with us, he will contact us and we will provide the URLs. They can then be configured like so:
```
pbjs.setConfig({
    zemanta: {
      bidderUrl: 'https://bidder-url.com',
      usersyncUrl: 'https://usersync-url.com'
    }
});
```

- contact email of the adapter’s maintainer: prog-ops-team@outbrain.com

PR containing documentation: https://github.com/prebid/prebid.github.io/pull/2537

### DIFF
--- a/modules/zemantaBidAdapter.js
+++ b/modules/zemantaBidAdapter.js
@@ -1,0 +1,252 @@
+// jshint esversion: 6, es3: false, node: true
+'use strict';
+
+import {
+  registerBidder
+} from '../src/adapters/bidderFactory.js';
+import { NATIVE, BANNER } from '../src/mediaTypes.js';
+import * as utils from '../src/utils.js';
+import { ajax } from '../src/ajax.js';
+import { config } from '../src/config.js';
+
+const BIDDER_CODE = 'zemanta';
+const CURRENCY = 'USD';
+const NATIVE_ASSET_IDS = { 0: 'title', 2: 'icon', 3: 'image', 5: 'sponsoredBy', 4: 'body', 1: 'cta' };
+const NATIVE_PARAMS = {
+  title: { id: 0, name: 'title' },
+  icon: { id: 2, type: 1, name: 'img' },
+  image: { id: 3, type: 3, name: 'img' },
+  sponsoredBy: { id: 5, name: 'data', type: 1 },
+  body: { id: 4, name: 'data', type: 2 },
+  cta: { id: 1, type: 12, name: 'data' }
+};
+
+export const spec = {
+  code: BIDDER_CODE,
+  supportedMediaTypes: [ NATIVE, BANNER ],
+  isBidRequestValid: (bid) => {
+    return !!(bid.nativeParams || bid.sizes);
+  },
+  buildRequests: (validBidRequests, bidderRequest) => {
+    const page = bidderRequest.refererInfo.referer;
+    const ua = navigator.userAgent;
+    const test = setOnAny(validBidRequests, 'params.test');
+    const publisher = setOnAny(validBidRequests, 'params.publisher');
+    const cur = CURRENCY;
+    const endpointUrl = config.getConfig('zemanta.bidderUrl');
+    const timeout = bidderRequest.timeout;
+
+    const imps = validBidRequests.map((bid, id) => {
+      bid.netRevenue = 'net';
+      const imp = {
+        id: id + 1 + ''
+      }
+
+      if (bid.params.tagid) {
+        imp.tagid = bid.params.tagid
+      }
+
+      if (bid.nativeParams) {
+        imp.native = {
+          request: JSON.stringify({
+            assets: getNativeAssets(bid)
+          })
+        }
+      } else {
+        imp.banner = {
+          format: transformSizes(bid.sizes)
+        }
+      }
+
+      return imp;
+    });
+
+    const request = {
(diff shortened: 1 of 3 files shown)


modules/zemantaBidAdapter.md
```diff
@@ -0,0 +1,95 @@
...
+          }
+        },
+        bids: [{
+            bidder: 'zemanta',
+            params: {
+                publisher: {
+                  id: '2706', // required
+                  name: 'Publishers Name',
+                  domain: 'publisher.com'
+                }
```
ncolletti, line 90: missing a "," on this line.

Also, we will need you to include a test endpoint url in this markdown for the **bidderUrl** so we can confirm bids are received and respond back to prebid properly through your adapter. The latter part is missing right now.

modules/zemantaBidAdapter.js
```diff
@@ -0,0 +1,252 @@
...
+  isBidRequestValid: (bid) => {
+    return !!(bid.nativeParams || bid.sizes);
+  },
+  buildRequests: (validBidRequests, bidderRequest) => {
+    const page = bidderRequest.refererInfo.referer;
+    const ua = navigator.userAgent;
+    const test = setOnAny(validBidRequests, 'params.test');
+    const publisher = setOnAny(validBidRequests, 'params.publisher');
+    const cur = CURRENCY;
+    const endpointUrl = config.getConfig('zemanta.bidderUrl');
```
ncolletti, line 36: perhaps a better way of achieving this would be to utilize the **utils.buildUrl** function and have the user pass in an object that will then create a the endpoint. thoughts around this are you'll be able to verify it builds properly and matches a valid endpoint url for your bidder.

`pbjs.setConfig({
    zemanta: {
        bidderUrl: {
            protocol: 'https',
            host: 'bidderendpointurl',
            pathname: '/zenmax/ads',
        },
        usersyncUrl: {
            protocol: 'https',
            host: 'usersyncurl',
            pathname: '/zenmax/sync'
        }
    }
})`

modules/zemantaBidAdapter.js
```diff
@@ -0,0 +1,252 @@
...
+  sponsoredBy: { id: 5, name: 'data', type: 1 },
+  body: { id: 4, name: 'data', type: 2 },
+  cta: { id: 1, type: 12, name: 'data' }
+};
+
+export const spec = {
+  code: BIDDER_CODE,
+  supportedMediaTypes: [ NATIVE, BANNER ],
+  isBidRequestValid: (bid) => {
+    return !!(bid.nativeParams || bid.sizes);
```
ncolletti, line 28: can a bid request be valid without a valid bidderUrl from the config?

modules/zemantaBidAdapter.js
```diff
@@ -0,0 +1,252 @@
...
+          bidObject.height = bidResponse.h;
+        }
+        return bidObject;
+      }
+    }).filter(Boolean);
+  },
+  getUserSyncs: (syncOptions) => {
+    const syncs = [];
+    const syncUrl = config.getConfig('zemanta.usersyncUrl');
+    if (syncOptions.pixelEnabled) {
```
ncolletti, line 136: would want to check if syncUrl exists here and is a valid Url

modules/zemantaBidAdapter.js
rokostik: I would leave this as is since the publishers integrating with us will get different urls.

modules/zemantaBidAdapter.md
rokostik: Ok, we will prepare and include the test url.

modules/zemantaBidAdapter.js
ncolletti: What is the structure of the URL for bidder and sync endpoints? If the host is always the same you could have the user only define the path pointing to their url.  Also this should be updated to use the params of your bid to pass the URL to your bidder, not the config. You can see how other bidders like gamoshi are doing this. Do you want to include a backup endpoint in case there is none provided or it is incorrect?

The sync url should be passed by your endpoint in the response, as you know the endpoint url from the request. This is another point of failure that can be fixed.

rokostik: The host will change depending on the user and so could the path. The user will need to contact us before integrating so we see no issue providing them the full url. 

We do not support returning the usersync url in the response. The only place we can configure the usersync url then is in the config so we also moved the bidder url there to be consistent and not confuse the user by configuring the urls at two places.

rokostik:
Hi @ncolletti, in 3d838b9 we added the urls to the documentation. Please note that in some cases it's possible that you will receive a `204 No Content` bid response, especially if running the project on localhost. I hope that this does not cause you too much trouble.

ncolletti:
LGTM. I wasn't able to get a bid response on your test endpoint, please make sure all creatives that respond to your adapter are fully HTTPS compliant.

PR #6053, opened 2020-11-30
oneVideo Adapter - Custom Key Value Pair targeting support (SAPR-15478)

adam-browning:
## Type of change
- [ x] Feature

## Description of change
oneVideo adapter now supports Custom key value pairs for targeting in the Video SSP ad-server.
You can pass an Object with key value pairs under `bid.params.video.custom = {}``.
Supported values are String and Number only

```
bids: [{
                bidder: 'oneVideo',
                params: {
                    video: {
                        playerWidth: 640,
                        playerHeight: 480,
                        protocols: [2, 5],
                        api: [1, 2],
                        custom :{
                            key1: "value1",
                            key2: "value2",
                            key3: "value3",
                            key4: 4444444,
                        },
                        e2etest: true
                    },
                    pubId: 'whatever'
                }
            }]
```

## Other information
@kprasadpvv would appreciate your review whenever convenient :)
Thanks in advance

### DIFF
--- a/modules/oneVideoBidAdapter.js
+++ b/modules/oneVideoBidAdapter.js
@@ -305,9 +305,9 @@ function getRequestData(bid, consentData, bidRequest) {
   if (bid.params.video.custom && Object.prototype.toString.call(bid.params.video.custom) === '[object Object]') {
     bidData.imp[0].ext.custom = {};
     for (const key in bid.params.video.custom) {
-        if (typeof bid.params.video.custom[key] === 'string' || typeof bid.params.video.custom[key] === 'number') {
-            bidData.imp[0].ext.custom[key] = bid.params.video.custom[key];
-        }
+      if (typeof bid.params.video.custom[key] === 'string' || typeof bid.params.video.custom[key] === 'number') {
+        bidData.imp[0].ext.custom[key] = bid.params.video.custom[key];
+      }
     }
   }
   return bidData;
--- a/modules/oneVideoBidAdapter.js
+++ b/modules/oneVideoBidAdapter.js
@@ -302,7 +302,7 @@ function getRequestData(bid, consentData, bidRequest) {
     bidData.site.ref = 'https://verizonmedia.com';
     bidData.tmax = 1000;
   }
-  if (bid.params.video.custom) {
+  if (bid.params.video.custom && Object.prototype.toString.call(bid.params.video.custom) === '[object Object]') {
     bidData.imp[0].ext.custom = {};
     for (const key in bid.params.video.custom) {
       const value = bid.params.video.custom[key]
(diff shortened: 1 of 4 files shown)


modules/oneVideoBidAdapter.js
```diff
@@ -302,6 +302,14 @@ function getRequestData(bid, consentData, bidRequest) {
     bidData.site.ref = 'https://verizonmedia.com';
     bidData.tmax = 1000;
   }
+  if (bid.params.video.custom && Object.prototype.toString.call(bid.params.video.custom) === '[object Object]') {
```
kprasadpvv, line 305: I think we can use `typeof` check for object too.
i.e., use `typeof bid.params.video.custom` instead of 
`Object.prototype.toString.call(bid.params.video.custom) === '[object Object]'`

Screenshot 2020-12-01 at 9 16 09 PM

kprasadpvv: oh! you are checking for skipping array?

adam-browning: I tried that before and found that `typeof` fails when passing: Array.
Then discovered that it also fails when passing Set, Date & Null - they are all considered objects.
`Object.prototype.toString.call(bid.params.video.custom) === '[object Object]'` ensures that you can only pass an Object
and works when validating: 
- String
- Object
- Array
- Set
- Date
- Undefined
- Null
If this is overkill let me know - but I wanted to make sure nothing else is passed.

kprasadpvv:
@adam-browning can you also update sample object in `oneVideoBidAdapter.md`
https://github.com/prebid/Prebid.js/blob/d74ad80d4e53529148fa3b7b1da2f6affd88bd60/modules/oneVideoBidAdapter.md

modules/oneVideoBidAdapter.js
kprasadpvv: I think what you are doing is best for skipping other type values. 👍🏼

kprasadpvv (approved):

FilipStamenkovic (approved):
LGTM

modules/oneVideoBidAdapter.js
FilipStamenkovic: I saw this discussion, I just want to share my advice, this is for future PRs:  
it would be easier/shorter just to execute `utils.isPlainObject(bid.params.video.custom)`.

kprasadpvv: Just went through the `utils`,  Thank you @FilipStamenkovic there are lot of helpers there will use them going forward.

adam-browning: Thank you so much @FilipStamenkovic that is perfect! 
I will update accordingly

PR #6054, opened 2020-11-30
Update test unit to match new alias bidder count (Adkernel)

iskmerof:
## Type of change

- [X] Other

## Description of change
Update test unit for alias count

PR #6055, opened 2020-11-30
Update adkernelBidAdapter.js for client alias

iskmerof:
## Type of change

- [X] Other

## Description of change
Add client alias for Adkernel adapter


aleksatr:
Expecting test fix from `https://github.com/prebid/Prebid.js/pull/6054` to be pushed here

aleksatr:
LGTM

PR #6063, opened 2020-12-01
Lotame id system domain update

markaconrad:
## Type of change

- [x] Bugfix
- [ ] Feature
- [ ] New bidder adapter   
- [ ] Build related changes
- [ ] CI related changes
- [ ] Other

## Description of change
Define method on userId example page so the page loads
A few enhancements to the Lotame Panorama Id System User Id module


modules/lotamePanoramaIdSystem.js
```diff
@@ -148,6 +147,42 @@ export const lotamePanoramaIdSubmodule = {
...
+      const value = storage.getCookie(TEST_COOKIE_NAME, undefined);
+      if (value === TEST_COOKIE_VALUE) {
+        continueSearching = false;
+      } else {
+        startIndex += -1;
+        continueSearching = Math.abs(startIndex) <= domainParts.length;
+      }
+    } while (continueSearching);
+    return rootDomain;
+  },
```
smenzer, line 184: there's at least one or two other id modules that are doing essentially the same thing -- finding the root domain at which to write the 1p cookie. what are people's thoughts on moving this to a method in `utils` or `userId` module that can be re-used rather than repeated in multiple modules?

cc @patmmccann @Fawke @jdwieland8282 @robertrmartinez @pycnvr

patmmccann:
This seems closely related to https://github.com/prebid/Prebid.js/pull/5500 ; @idettman tried a couple different approaches there and might be most qualified to comment

modules/lotamePanoramaIdSystem.js
pycnvr: @smenzer Yes, I like your idea.

smenzer: @idettman do you have thoughts here? is this something we can/should break out to be a common userid (or even common prebid global) method?

markaconrad: Are you all thinking this should be done in context of this PR or separately?

smenzer: I think it probably makes the most sense to break this out into a separate PR, then you can use this PR to apply it to the Lotame ID system

markaconrad: Alright, I will give that shot

modules/lotamePanoramaIdSystem.js
```diff
@@ -205,10 +241,18 @@ export const lotamePanoramaIdSubmodule = {
           if (response) {
             try {
               let responseObj = JSON.parse(response);
-              saveLotameCache(KEY_EXPIRY, responseObj.expiry_ts);
+              const shouldUpdateProfileId = !(utils.isArray(responseObj.errors) && responseObj.errors.includes(111));
```
noahpc, line 244: Recommend using constant for 111 named MISSING_CORE_CONSENT

modules/lotamePanoramaIdSystem.js
markaconrad: That part is attempted in https://github.com/prebid/Prebid.js/pull/6124

smenzer: @markaconrad thanks for putting in #6124! I assume once that goes through, you would remove this block and on line 206 below you'd use the user id function instead of `this` right?

markaconrad: You are welcome! Yes, that is the plan

markaconrad:
I'm going to close this out but will re-open with a couple more changes and the result of https://github.com/prebid/Prebid.js/pull/6124

PR #6066, opened 2020-12-01
GumGum: sets mediaType of bidRequest depending on product id

susyt:
## Type of change

- [x] Bugfix
- [ ] Feature
- [ ] New bidder adapter   
- [ ] Build related changes
- [ ] CI related changes
- [ ] Other

## Description of change

Sets mediaType to VIDEO when the product type is found to be video (product 7) or in-video (product 6). Default mediaType is set to BANNER

## Other information

@derberbaby


Rothalack:
Took 6 tries to get Circleci to pass, but it made it!

Do you think you'd want to make a documentation update for this?

susyt:
thanks @Rothalack! and sure thing! as in make a note that in-video products are considered mediatype video and all else is defaulted to banner?

Rothalack:
The two places I was thinking of were:
https://github.com/prebid/Prebid.js/blob/master/modules/gumgumBidAdapter.md
and
https://docs.prebid.org/dev-docs/bidders.html#gumgum

If I'm understanding correctly, it would be applicable to update those in relation to this change, but I could be wrong!

susyt:
ok updated! the corresponding prebid.github PR is: https://github.com/prebid/prebid.github.io/pull/2551

PR #6067, opened 2020-12-01
Add Gulp Review-Start Task

GLStephen:
This change adds a `gulp review-start` command to the build process. 

```
// build task for reviewers, runs lint, test-coverage and serves
gulp.task('review-start', gulp.series(clean, lint, gulp.parallel('build-bundle-dev', testCoverage)));
```

This will do the following:

* Clean the build directory
* Run lint
* Build the dev bundle, without watching
* Run test-coverage
* Start the server

This is the equivalent of multiple previously separate reviewer workflows where the reviewer may run "gulp test", then "test-coverage" then "view-coverage" then "gulp serve" to check the hello-world or another integration example page.

Now, the reviewer can simply run "gulp review-start" all of those substeps are completed and the coverage and integration examples are available in the served folder. 

I also updated the PR_REVIEW.md with some references to these steps.

Why? I got tired of typing all those commands... http://threevirtues.com/

## Type of change
- [ x ] Build related changes


gulpfile.js
```diff
@@ -383,4 +383,7 @@ gulp.task('e2e-test', gulp.series(clean, setupE2e, gulp.parallel('build-bundle-p
 gulp.task(bundleToStdout);
 gulp.task('bundle', gulpBundle.bind(null, false)); // used for just concatenating pre-built files with no build step
 
+// build task for reviewers, runs test-coverage, serves, without watching
+gulp.task('review-start', gulp.series(clean, lint, gulp.parallel('build-bundle-dev', testCoverage)));
```
smenzer, line 387: i'm probably missing something obvious, but where is the "serve" part? does testCoverage actually execute a serve?

smenzer (commented):
I love seeing optimizations like this! Thanks for putting in

gulpfile.js
GLStephen: Good catch, I must have had an old "watch" running in another terminal when I was testing this. I need to refactor the "serve" call a bit since it cleans, etc. Will get it fixed up.

GLStephen: I added back the "watch" instead of getting to a refactor of the core commands. Removing the watch element was a nicety, not a requirement.

GLStephen:
@smenzer updated to add the watch back so it serves

smenzer (approved):
seems to be working nicely for me!

smenzer:
@jsnellbaker do we need a 2nd review on this or can I go ahead and merge?

jsnellbaker:
Hi @smenzer this type of change does need a 2nd review, thanks for checking.  

@Fawke can you please take a look at this PR when you can?

Fawke (changes requested):
Hi @GLStephen,

This looks awesome, that's for automating most of the gulp review steps. This is great, but it'll be even better if we can eliminate these two steps as well:

```
- Navigate to build/coverage/lcov-report/index.html to view coverage
- Navigate to integrationExamples/gpt/hellow_world.html for basic integration testing
```

Sometimes, it's kind of irritating to type this long url. First step can be automated easily if you modify your gulp command to add `viewCoverage` task as well, like this:

```(javascript)
gulp.task('review-start', gulp.series(clean, lint, gulp.parallel('build-bundle-dev', watch, testCoverage), viewCoverage));
```
Second step can also be automated, but we'd need to create an additional gulp task to serve this particular directory, `/integrationExamples` on `localhost`. I can create another PR or add to this if you need any help with the second part.

GLStephen:
@Fawke I'll add those

Fawke:
@GLStephen I like the idea of viewing just the root, that'll give more control to the reviewer. And opening multiple browser window - one for code coverage and other one for e2e testing - could be a little confusing and irritating.

smenzer:
just to play devil's advocate to the last two comments: this feature is to make easier for reviewers, and if we know that 90% of the time they will need to open the hello world file, opening to that file (or to "view" that file in the ui even if it's not directly loaded) would speed things up for them rather than having to click through 3 or 4 directories. for those that need a different page, then the clicks to get out of hello world to another file are probably the same as what they'd have to do from root anyways.

i also don't think having two windows pop up (hello world and code coverage) is that annoying -- worst case you close one, but they _should_ be used most of the time, so at the very least it would remind reviewers to review them. and nothing here should be confusing to the reviewers...it will be part of the new process so they should know exactly what happens when they use the command.

so in summary...i would vote for loading the code coverage and a window to "view" the hello world file in the UI.

GLStephen:
@smenzer Point taken, why don't we open to a "reviewer tools" page then? we can add deep links to anything commonly useful to them and have it self document basically.

smenzer:
> @smenzer Point taken, why don't we open to a "reviewer tools" page then? we can add deep links to anything commonly useful to them and have it self document basically.

i like it! great idea

smenzer (approved):
LGTM

PR_REVIEW.md
```diff
@@ -5,6 +5,16 @@ If the PR is for a standard bid adapter or a standard analytics adapter, just th
...
+### Running Tests and Verifying Integrations
+
+General gulp commands include separate commands for serving the codebase on a built in webserver, creating code coverage reports and allowing serving integration examples. The `review-start` gulp command combinese those into one command.
+
+- Run `gulp review-start`, adding the host parameter `gulp review-start --host=0.0.0.0` will bind to all IPs on the machine
+- Navigate to `http://host-ip:9999`
+- Navigate to build/coverage/lcov-report/index.html to view coverage
+- Navigate to integrationExamples/gpt/hellow_world.html for basic integration testing
+- The hello_world.html and other exampls can be edited and used as needed to verify functionality
+
```
Fawke, line 17: @GLStephen Great work, I really like the idea of a reviewer dedicated page. Now, do we wanna get rid of these lines, since running `gulp review-start` will now do all the magic?

GLStephen: I'd like to leave them but add them as the "manual" option. I'll clarify it a bit.

GLStephen: @Fawke updated

Fawke (approved):

Issue #6051, opened 2020-11-30
USP api locator improvement


ReinoutStevens:
## Type of issue

This is a proposal for a small improvement in locating the `__uspapi` function:

Currently, the consentmanagentUsp module looks for `window.__uspapi` on the current window. If that fails, it looks for a `uspapiFrame` frame.

On the other, the consentmanagent module loops through the different parent windows, first checks whether the consent module is available directly, and if not checks for a locator frame. This has the benefit that setups where the parent is accessible ( ie. on same-domain iframes ), the api is used directly instead of using postMessage.

The proposed solution is to mimick the behavior of `consentManagement.js` in `consentManagementUsp.js` to locate the `__uspapi`

patmmccann:
seems related to https://github.com/prebid/Prebid.js/pull/4755

Issue #6050, opened 2020-11-30
USP in iframe installs broken __uspapi


ReinoutStevens:
## Type of issue
When __uspapi is provided through an iframe / __uspapiLocator, the consent management module installs __uspapi that communicates through the message api.
Once prebid has received its consent information, it removes the listener. As a result, other libraries that require consent info communicate with a broken __uspapi, as callbacks are never triggered

## Description

The consentManagementUsp.js has the following [code](https://github.com/prebid/Prebid.js/blob/master/modules/consentManagementUsp.js#L134-L136)

```
    function uspapiCallback(consentObject, success) {
      window.removeEventListener('message', readPostMessageResponse, false);
      moduleCallback(consentObject, success);
    }
```

Removing the event listener results that subsequent calls to __tcfapi wont be handled by the newly installed __tcfapi function.
The solution is just to remove the `window.removeEventListener` line.

PR #6072, opened 2020-12-02
Improve __uspapi behavior in iframe

ReinoutStevens:
## Type of change

- [x] Bugfix

## Description of change

This fixes the following two issues:

* https://github.com/prebid/Prebid.js/issues/6051
* https://github.com/prebid/Prebid.js/issues/6050

I've aligned the locator code with the `consentManagement.js` module so that use similar logic in finding where the `uspapi` function resides.

I've removed the code that removes the event listener after the consent information has been retrieved, as this breaks the `__uspapi` function that we installed on the current window. As a result, external libraries would think that the current window has a uspapi, call it, but never get a response.

## Other information

The existing unit tests already cover the cases of finding the uspapi in an iframe and directly, and still pass.
I've also tested this in a real world setup, and both the case of accessing the api directly and cross iframe work.

### DIFF
--- a/modules/consentManagementUsp.js
+++ b/modules/consentManagementUsp.js
@@ -44,6 +44,34 @@ function lookupStaticConsentData(cmpSuccess, cmpError, hookConfig) {
  * @param {object} hookConfig contains module related variables (see comment in requestBidsHook function)
  */
 function lookupUspConsent(uspSuccess, uspError, hookConfig) {
+  function findUsp() {
+    let f = window;
+    let uspapiFrame;
+    let uspapiFunction;
+
+    while (!uspapiFrame) {
+      try {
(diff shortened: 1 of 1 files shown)


ReinoutStevens:
@robertrmartinez would be a shame if this PR got closed due to inactivity

robertrmartinez (approved):
Looks fine to me, although I was not the one who implemented the iframe lookup initially.

Needs a second review anyways, so going to assign the second review to @jsnellbaker since he wrote the initial consentManagement code!

smenzer:
@jsnellbaker can we get your 2nd review on this one, please?

jsnellbaker (approved):
LGTM

Issue #6018, opened 2020-11-23
video playerSize width and height wrong way around


spormeon:
prebid 4.15.0

something is making the playerSize the wrong way around for appnexus, so it totally messes up other bidder, with playerSize being a required attribute on the adUnit:

![DevTools_-_c](https://user-images.githubusercontent.com/8503167/99998062-fc4a7280-2db5-11eb-8354-65d43f5ba1a2.png)

bretg:
Good find @spormeon -- this is a problem in the Prebid Server Bid Adapter -- https://github.com/prebid/Prebid.js/blob/cebae4a07748bcf0570e1781c31bb9f865bffbab/modules/prebidServerBidAdapter/index.js#L753

The PBJS 'playerHeight' and 'playerWidth' values are set (backwards) in pbsBidAdapter from the original PBJS bidRequest -- but I don't think anything downstream actually uses them, which is why it hasn't been noticed until now.

Still, something that needs to be fixed.

PR #6073, opened 2020-12-02
fix prebid server playerwidth and height

ReinoutStevens:
## Type of change

- [x ] Bugfix
- [ ] Feature
- [ ] New bidder adapter   
- [ ] Build related changes
- [ ] CI related changes
- [ ] Other

## Description of change

Fixes https://github.com/prebid/Prebid.js/issues/6018

### DIFF
--- a/modules/prebidServerBidAdapter/index.js
+++ b/modules/prebidServerBidAdapter/index.js
@@ -750,8 +750,8 @@ const OPEN_RTB_PROTOCOL = {
           if (utils.deepAccess(bid, 'ext.prebid.type') === VIDEO) {
             bidObject.mediaType = VIDEO;
             let sizes = bidRequest.sizes && bidRequest.sizes[0];
-            bidObject.playerHeight = sizes[0];
-            bidObject.playerWidth = sizes[1];
+            bidObject.playerWidth = sizes[0];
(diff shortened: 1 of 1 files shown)


ReinoutStevens:
Unit tests work locally, and I cant see the details of circleci.
If anyone could post the errors so I can investigate the error

bretg:
Certainly appreciate the update @ReinoutStevens , but we need a unit test. Generally when something like this slides for so long, it means we're missing a test. :-)

bretg:
Gotta love those flaky unit tests. 5 runs later it finally succeeded.

ReinoutStevens:
@bretg A bit out of my league

Things that I see that are potentially wrong:

* The code should probably access playerSize, and not sizes, which afaik is the deprecated way to specify width and height. Given that there is a `isValidBid` check that checks for playerSize this should be fine
* I've been trying to get a working unit test: from what I see the unit tests have a `BID_REQUESTS = [...]`, but this only contains a banner ad, and so the code cant access `playerSize` in the unit tests ( as it doesn't exist in the mocked data ), but I am not confident enough to write a video mock. The unit tests would probably pass, but the code would just be dependent on what mock I wrote, and chances are pretty high that it will be wrong.

msm0504 (approved):

PR #6076, opened 2020-12-02
a/b testing framework baked in to the ID5 user id module

smenzer:
## Type of change
- [ ] Bugfix
- [x] Feature
- [ ] New bidder adapter   
- [ ] Build related changes
- [ ] CI related changes
- [ ] Other

## Description of change
Publishers are looking for ways to evaluate the uplift of ID5. To make this easy, we have added an A/B testing feature to our module to allow publishers to maintain a control group of requests where an ID5 ID is not made available to bid adapters. Publishers can identify the control group through a new field on the ID5 id object and pass that to their reporting tool (their ad server, an analytics adapter, etc) to do the analysis.

## Other information
Docs PR: https://github.com/prebid/prebid.github.io/pull/2548

### DIFF
--- a/modules/id5IdSystem.js
+++ b/modules/id5IdSystem.js
@@ -42,27 +42,58 @@ export const id5IdSubmodule = {
    * decode the stored id value for passing to bid requests
    * @function decode
    * @param {(Object|string)} value
+   * @param {SubmoduleConfig|undefined} config
    * @returns {(Object|undefined)}
    */
-  decode(value) {
-    let uid;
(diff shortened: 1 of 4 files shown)


robertrmartinez (approved):
Looks good to me, and looks like a nice feature! Thanks.

robertrmartinez:
Although, it may be nice to have this AB Test framework done for the userId module itself.

Configurable at a per userID provider level, and as a global AB test all user IDs.

Maybe something we do in the future perhaps?

PR #6093, opened 2020-12-07
PBJS Core: canBidderRegisterSync ignoring iframe sync disabled by default

JonGoSonobi:
## Type of change

- [x] Bugfix
- [ ] Feature
- [ ] New bidder adapter   
- [ ] Build related changes
- [ ] CI related changes
- [ ] Other

## Description of change
The method userSync.canBidderRegisterSync returns true if there is no iframe filteredSettings defined. Since iframe user syncs are disabled by default, this method should return false when no iframe filteredSettings are defined.

### DIFF
--- a/src/userSync.js
+++ b/src/userSync.js
@@ -226,13 +226,8 @@ export function newUserSync(userSyncDependencies) {
       return checkForFiltering[filterType](biddersToFilter, bidder);
     }
 
-    // iframes are disabled by default.
-    // so if the iframe config is invalid, we should use the permittedPixels object to tell us if the bidder should be blocked when trying to register an iframe pixel.
-    if (!isFilterConfigValid(filterConfig, type) && type === 'iframe') {
-      return !permittedPixels[type];
-    }
+    return !permittedPixels[type];
 
-    return false;
   }
 
   /**
--- a/src/userSync.js
+++ b/src/userSync.js
@@ -225,6 +225,13 @@ export function newUserSync(userSyncDependencies) {
       }
       return checkForFiltering[filterType](biddersToFilter, bidder);
     }
+
+    // iframes are disabled by default.
+    // so if the iframe config is invalid, we should use the permittedPixels object to tell us if the bidder should be blocked when trying to register an iframe pixel.
+    if (!isFilterConfigValid(filterConfig, type) && type === 'iframe') {
+      return !permittedPixels[type];
+    }
+
     return false;
   }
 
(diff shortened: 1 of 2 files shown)


src/userSync.js
```diff
@@ -225,6 +225,13 @@ export function newUserSync(userSyncDependencies) {
...
       return checkForFiltering[filterType](biddersToFilter, bidder);
     }
+
+    // iframes are disabled by default.
+    // so if the iframe config is invalid, we should use the permittedPixels object to tell us if the bidder should be blocked when trying to register an iframe pixel.
+    if (!isFilterConfigValid(filterConfig, type) && type === 'iframe') {
+      return !permittedPixels[type];
+    }
+
     return false;
```
FilipStamenkovic, line 235: It looks to me that all these lines can be replaced with simple:

```
return !permittedPixels[type];
```

There is no need to check `!isFilterConfigValid(filterConfig, type)`, previous `if` is already checking that condition, in other words it will always evaluate to `true`.

FilipStamenkovic (commented):
All in all good bug fix, but it looks to me it could have be done simpler with single `return` (take a look at the comment?

src/userSync.js
JonGoSonobi: Makes sense. Change applied

FilipStamenkovic: Build is failing, you have linting error:

```
/home/circleci/Prebid.js/src/userSync.js
  231:3  error  Block must not be padded by blank lines  padded-blocks

✖ 1 problem (1 error, 0 warnings)
  1 error, 0 warnings potentially fixable with the `--fix` option.

[14:12:48] 'lint' errored after 28 s
[14:12:48] ESLintError in plugin "gulp-eslint"
Message:
    Failed with 1 error
Details:
    domainEmitter: [object Object]
    domainThrown: false

[14:12:48] 'test' errored after 28 s
```

FilipStamenkovic:
@JonGoSonobi You still have linting error, please check my previous comment.

ChrisHuie:
@JonGoSonobi I tried to remote fetch this and quickly fix the extra line spaces that are throwing and blocking this pr but looks like I don't have access

ChrisHuie:
Tried to fix the lenting error but don't have permission to the repo

patmmccann:
@ChrisHuie should we just submit as a new pr since OP is MIA and this bug persists?

patmmccann:
closing in favor of redo #6535

PR #6098, opened 2020-12-08
Sovrn: Pass the imp.ext.deals field

egsgordeev:
## Type of change

- [ ] Bugfix
- [x] Feature
- [ ] New bidder adapter  
- [ ] Build related changes
- [ ] CI related changes
- [ ] Other

## Description of change
Pass the `segments` parameter value as `imp.ext.deals` array

### DIFF
--- a/modules/beachfrontBidAdapter.js
+++ b/modules/beachfrontBidAdapter.js
@@ -6,7 +6,7 @@ import { VIDEO, BANNER } from '../src/mediaTypes.js';
 import find from 'core-js-pure/features/array/find.js';
 import includes from 'core-js-pure/features/array/includes.js';
 
-const ADAPTER_VERSION = '1.14';
+const ADAPTER_VERSION = '1.15';
 const ADAPTER_NAME = 'BFIO_PREBID';
 const OUTSTREAM = 'outstream';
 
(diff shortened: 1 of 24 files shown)


osazos:
Hi @egsgordeev,

Is this new param shouldn't be added to your sovrnBidAdapter.md file and the public doc ?

Also, when you transform `segments` string to array, you dont trim them, which allow something like `['  test1', 'test 2  ']`. It's not a required change, just a comment in case this behavior is not expected.

egsgordeev:
Hi @osazos,

thank you for the review!

I will make a PR to our fork and ask the maintainer about the public doc.

Is this PR blocked for now? I'm asking since PR on our side might take some time.

egsgordeev:
@osazos,

I got to know that this feature is only a proof of concept so it will be documented as soon as proved to be useful.

osazos:
Hi @egsgordeev,

Ok, thanks for clarification.

osazos (approved):
LGTM

As you said, don't forget to add this to the main doc or remove it regarding your choice to keep this feature or not in the future.

egsgordeev:
@osazos, 
I made a ticket on our side to keep that in mind.

Thank you!

Issue #5814, opened 2020-10-01
eids permissioning scheme 


jdwieland8282:
## Type of issue
feature request

## Description
Do Publishers need additional controls for determining which user ids go to which bid adapters?  Publishers can currently control what User id modules they deploy and which bid adapters they integrate with.  Are their scenarios where a publisher would deploy a user id module but decide NOT to send the resulting user id to a bid adapter they've integrated with? If we believe there are then a permissioning structure in the eids config is needed. 

## Proposed Structure 
```
pbjs.setConfig({
    userSync: {
        userIds: [{
            name: "MODULE NAME",
            **bidders: ['bidderA', 'bidderB'], // new permissions field. This ID can only go to these bidders.**
            params: {
                ...
            }
        }]
    }
});
```

### OpenRTB support
```
{
    ext: {
        prebid: {
            data: {
                bidders: ['bidderA','bidderB']
            }
        }
    }
}
```

bretg:
Implications for PBJS:
- something would need to modify the bidRequest object before going to each bidder.
- unclear whether this is PBJS-core or perhaps its just done through a hook in the UserId module itself

Implications for Prebid Server:
- This proposal treats eids as first party data, using the same permissioning scheme. But if pubs need some eids to be global, while other eids are first party data, we'll need to extend the permissioning scheme. 

**Option 1**:  put eids permissions alongside FPD permissions
```
{
    ext: {
        prebid: {
            data: {
                bidders: ['bidderA','bidderB'], // this is for FPD
                eidPermissions: [
                   {"source": EID_SOURCE, "bidders": ["bidderA"]}
                ]
            }
        }
    }
}
```
One issue is that the PBJS config for user ID sub-modules isn't directly linked to the eids.source field, so the pbsBidAdapter doesn't have all the right info here. Possible solutions include specifically naming the eids.source in the config or retrofitting all the userId modules to support a function confirming whether a given source 

**Option 2**: eids.ext.bidders
- add a routine to the userId module that sub-modules have to use to register their eids. UserId-core adds any bidder config
 




-

bretg:
Updated openrtb spec

gglas:
@jdwieland8282 we think this issue is closed, please let us know if any reason to repon.

PR #6112, opened 2020-12-10
Extended ID permissions supported by bidder

YerkovichM:
## Type of change

- [x] Feature

## Description of change

#5814 
Bid adapters receive only allowed user ids

**Case 1**: 
```
userSync: {
 userIds: [
  {
    name: "sharedId",
    bidders: [
      'rubicon',
      'exBidder'
    ],
    params: {
      syncTime: 60 //inseconds,
      defaultis24hours
    },
    storage: {
      type: "cookie",
      name: "sharedid",
      expires: 28
    }
  }
]
}
```

Rubicon bid adapter receives:
```
{
  "bidder": "rubicon",
  ...
  "userId": {
    "sharedid": {
      "id": "01EQDMBY1YPFFVY57R2V11PY89",
      "third": "01EQDMBY1YPFFVY57R2V11PY89"
    }
  },
  "userIdAsEids": [
    {
      "source": "sharedid.org",
      "uids": [
        {
          "id": "01EQDMBY1YPFFVY57R2V11PY89",
          "atype": 1,
          "ext": {
            "third": "01EQDMBY1YPFFVY57R2V11PY89"
          }
        }
      ]
    }
  ],
  ...
}
```
PrebidServerBidAdapter request sent as:
```
{
  ...
  "ext": {
    "prebid": {
      ...
      "data": {
        "eidpermissions": [
          {
            "source": "sharedid.org",
            "bidders": [
              "rubicon"
            ]
          }
        ]
      }
    }
  },
  ...
  "user": {
    "ext": {
      "eids": [
        {
          "source": "sharedid.org",
          "uids": [
            {
              "id": "01EQDMBY1YPFFVY57R2V11PY89",
              "atype": 1,
              "ext": {
                "third": "01EQDMBY1YPFFVY57R2V11PY89"
              }
            }
          ]
        }
      ]
    }
  }
}
```
**Case 2**:
```
userSync: {
  userIds: [{
    name: "sharedId",
    bidders: ['exBidder'],
    ...
    }
  ]
}
```
Rubicon bid adapter receives:
```
{
  "bidder": "rubicon",
  ... no user ID entries ...
}
```
PrebidServerBidAdapter request:
```
{
  ...
  "ext": {
    "prebid": {
      ...
      "data": {
        "eidpermissions": [
          {
            "source": "sharedid.org",
            "bidders": [ 'exBidder' ]
          }
        ]
      }
    }
  },
  ...
}
```
**Case 3**: no permissions specified
```
userSync: {
  userIds: [{
    name: "sharedId",
    ...
    }
  ]
}
```
Rubicon bid adapter receives:
```
{
  "bidder": "rubicon",
  ...
  "userId": {
    "sharedid": {
      "id": "01EQDMBY1YPFFVY57R2V11PY89",
      "third": "01EQDMBY1YPFFVY57R2V11PY89"
    }
  },
  "userIdAsEids": [
    {
      "source": "sharedid.org",
      "uids": [
        {
          "id": "01EQDMBY1YPFFVY57R2V11PY89",
          "atype": 1,
          "ext": {
            "third": "01EQDMBY1YPFFVY57R2V11PY89"
          }
        }
      ]
    }
  ],
 ...
}
```
PrebidServerBidAdapter request:
```
{
  ...
  "ext": {
    "prebid": {
      ...
      "data": {
         ...
      }
    }
  },
  ...
  "user": {
    "ext": {
      "eids": [
        {
          "source": "sharedid.org",
          "uids": [
            {
              "id": "01EQDMBY1YPFFVY57R2V11PY89",
              "atype": 1,
              "ext": {
                "third": "01EQDMBY1YPFFVY57R2V11PY89"
              }
            }
          ]
        }
      ]
    }
  }
}
```
@jdwieland8282


bretg:
This is going to need update to modules/prebidServerBidAdapter as well.

We need to send the eidPermission in openrtb. There are a couple of ways to do this, but maybe the easiest would be to have pbsBidAdapter call a function of the userId module to get the json block?

modules/prebidServerBidAdapter/index.js
```diff
@@ -637,6 +637,11 @@ const OPEN_RTB_PROTOCOL = {
       utils.deepSetValue(request, 'user.ext.eids', bidUserIdAsEids);
     }
 
+    const eidPermissions = utils.deepAccess(bidRequests, '0.bids.0.eidPermissions');
```
bretg, line 640: I may be misunderstanding this, but it's not allowed to pass the eidPermissions directly to bidders.

i.e if this implies that the eidPermissions block is passed around on the bidRequest object, that would let client-side adapters know where they stand.

YerkovichM: Ok, I`ll change it

modules/userId/eids.js
```diff
@@ -1,4 +1,5 @@
 import * as utils from '../../src/utils.js';
+import {isArray} from "../../src/utils";
```
mmoschovas, line 2: utils.js is already imported above. Can you remove this second declaration and use the utils var instead?

modules/userId/index.js
```diff
@@ -124,6 +124,7 @@ import CONSTANTS from '../../src/constants.json';
 import { module, hook } from '../../src/hook.js';
 import { createEidsArray } from './eids.js';
 import { getCoreStorageManager } from '../../src/storageManager.js';
+import {getEidPermissions} from "./eids";
```
mmoschovas, line 127: Similar to the other comment, eids.js is defined above. Please combine as : 

`import  {createEidsArray, getEidPermissions } from './eids.js'`

mmoschovas (changes requested):
The above two changes should fix the circleci errors, but I think there should be a unit test added to prebidServerBidAdapter_spec to validate these changes.

YerkovichM:
This should be a functional part. Unit tests are coming next

YerkovichM:
@mmoschovas this should be it. Review the PR please

mmoschovas (approved):
Thanks for adding the tests. LGTM

modules/userId/eids.js
```diff
@@ -214,3 +214,24 @@ export function createEidsArray(bidRequestUserId) {
...
+export function buildEidPermissions(submodules) {
+  let eidPermissions = [];
+  submodules.filter(i => utils.isPlainObject(i.idObj) && Object.keys(i.idObj).length && i.config)
+    .forEach(i => {
+      Object.keys(i.idObj).forEach(key => {
+        if (utils.deepAccess(USER_IDS_CONFIG, key + '.source')) {
+          eidPermissions.push(
+            {
+              source: USER_IDS_CONFIG[key].source,
+              bidders: (i.config.bidders && Array.isArray(i.config.bidders) ? i.config.bidders : ['*'])
```
smenzer, line 230: since this list of bidders is being sent to SSPs and likely will land in OpenRTB bid requests, should the name `bidders` in the eids object be changed to `pbjs_bidders` since it won't really apply to the openrtb world?

also, rather than `['*']` can we just omit `bidders` (or `pbjs_bidders`) altogether if none were set? This will avoid sending a field that isn't necessary

smenzer (changes requested):
can you please add a test to cover the following scenario?
* `bidders` is set but doesn't include `sampleBidder`; check that the ID is not available in their bid request, but is available in the bidders listed

modules/userId/eids.js
YerkovichM: @jdwieland8282

jdwieland8282: @smenzer changing bidders to pbjs_bidders seems fine to me. And yeah I agree omitting `pbjs_bidders` if none are set makes sense. Thanks.

jdwieland8282:
Docs update: https://github.com/prebid/prebid.github.io/pull/2592

YerkovichM:
@smenzer I commited the update. Added test and renamed bidders to pbjs_bidders

smenzer (approved):
LGTM

smenzer:
@mmoschovas can you please have another check of the new changes and then you can merge if you're happy.

jdwieland8282:
@mmoschovas or @smenzer can this be merged in?

bretg:
I'd like to suggest that `pbjs_bidders` become `pbjsBidders` or better yet just `bidders`.

There's no precedence in PBJS for attributes with an underscore. It's always camel-case.

jdwieland8282:
@YerkovichM mind updating with @bretg 's comment above regarding the naming of `pbjs_bidders` to `bidders`

smenzer:
Camel case is fine with me but "bidders" in openrtb has a meaning (ie DSP) and in this case we're referring to prebid bidder adapters, which are almost exclusively SSPs. So I suggest we stick with pbjsBidders to make it clear what the value will be for

smenzer:
LGTM. since this touches core, @mmoschovas can you please do a final review with the updated code changes and then you can merge if you approve?

bretg:
>  "bidders" in openrtb has a meaning (ie DSP) and in this case we're referring to prebid bidder adapters, which are almost exclusively SSPs

I'd like to disagree with the decision to call this "pbjsBidders". 

1) bidder codes are the same between pbjs and pbs.
2) we've already got this feature under code review in PBS-Java per the original spec:

{code}
{
    ext: {
        prebid: {
            data: {
                eidpermissions: [   // prebid server will use this to filter user.ext.eids
                   {"source": "sharedid.org", "bidders": ["*"]},
                   {"source": "neustar.biz", "bidders": ["bidderB"]},
                   {"source": "id5-sync.com", "bidders": ["bidderA","bidderC"]}
                ]
            }
        }
    }
}
{code}

I'd like to stick with the original plan please?

modules/prebidServerBidAdapter/index.js
```diff
@@ -649,6 +650,18 @@ const OPEN_RTB_PROTOCOL = {
...
+    const eidPermissions = getEidPermissions();
+    if (utils.isArray(eidPermissions) && eidPermissions.length > 0) {
+      if (requestedBidders && utils.isArray(requestedBidders)) {
+        eidPermissions.forEach(i => {
+          if (i.pbjsBidders) {
+            i.pbjsBidders = i.pbjsBidders.filter(bidder => requestedBidders.includes(bidder))
+          }
+        });
+      }
+      utils.deepSetValue(request, 'ext.prebid.data.eidPermissions', eidPermissions);
```
bretg, line 662: This needs to be lowercase eidpermissions.

modules/prebidServerBidAdapter/index.js
```diff
@@ -649,6 +650,18 @@ const OPEN_RTB_PROTOCOL = {
       utils.deepSetValue(request, 'user.ext.eids', bidUserIdAsEids);
     }
 
+    const eidPermissions = getEidPermissions();
+    if (utils.isArray(eidPermissions) && eidPermissions.length > 0) {
+      if (requestedBidders && utils.isArray(requestedBidders)) {
+        eidPermissions.forEach(i => {
+          if (i.pbjsBidders) {
+            i.pbjsBidders = i.pbjsBidders.filter(bidder => requestedBidders.includes(bidder))
```
bretg, line 658: please make this 'bidders' and not 'pbjsBidders'.

bretg (changes requested):

smenzer:
> I'd like to stick with the original plan please?

@bretg is this feature intended only for prebid server? i see your point w.r.t. naming for PBS only.

YerkovichM:
@bretg updated

smenzer:
@Fawke can I get your second review since this touches core, please?

Fawke (approved):
Looks good, have tested this with few adapters, it works as expected.

bretg:
> is this feature intended only for prebid server?

No.

BTW - I updated the original description to reflect the current understanding: `bidders` and not `pbjs_bidders`.

Note that I also changed the expected results of Case 3 -- if `bidders` isn't specified, there's no need to send ext.prebid.data.eidpermissions to PBS at all. But looking at the code, I suspect that's not what's happening. Looks like ext.prebid.data.eidpermissions will be there every time whether there are special permissions or not. @YerkovichM , please confirm?  

I would like to propose we change this behavior... if a given EID doesn't specify `bidders` I don't think it should be part of the output of buildEidPermissions.

modules/userId/eids.js
```diff
@@ -220,3 +220,23 @@ export function createEidsArray(bidRequestUserId) {
...
+  let eidPermissions = [];
+  submodules.filter(i => utils.isPlainObject(i.idObj) && Object.keys(i.idObj).length)
+    .forEach(i => {
+      Object.keys(i.idObj).forEach(key => {
+        if (utils.deepAccess(USER_IDS_CONFIG, key + '.source')) {
+          let eidPermission = { source: USER_IDS_CONFIG[key].source };
+          if (utils.deepAccess(i, 'config.bidders') && Array.isArray(i.config.bidders)) {
+            eidPermission.bidders = i.config.bidders;
+          }
+          eidPermissions.push(eidPermission);
```
bretg, line 237: I don't think we should be adding anything to the eidPermissions array unless there are actually permissions to be added. i.e. if there's no `bidders` array specified, why do we need an entry at all?

YerkovichM: Previously entry contained bidders: ['*'] if every bidder allowed, but then decided to omit bidders in this case.
Anyway, I will omit this entry entirely

YerkovichM:
@bretg now if a given EID doesn't specify bidders it is not returnd by buildEidPermissions and therefore it is not incuded in prebid 
server request

bretg (approved):

bretg:
@YerkovichM - looks ok to me now. Thanks. Fix conflicts please, and then we can merge.

YerkovichM:
@bretg conflicts are resolved

Issue #5966, opened 2020-11-12
Prebid 5.0


gglas:
## Type of issue
Prebid 5.0 changes summary.

## Description

Prebid 5.0 is the next major version release of Prebid.js currently scheduled for Q2 of 2021.  Please use this issue to track things earmarked for inclusion in Prebid 5.0 and to discuss what should or shouldn't be included.

Currently, we have the following projects and considerations:

Official Refactor of Default Key-values

 - Now that we have the ability to determine "standard" versus "optional" keys, we'd like to have a discussion regarding reducing the number of standard keys and pushing more to optional in order to reduce the key-value footprint in the default version of Prebid.  Having too many key values flowing into GAM can cause performance issues, so we expect this to reduce inadvertent overloading of GAM key values while streamlining prebid.

Taxonomy Enforcement - adapters that do not conform to standards regarding the communication of the following auction meta data will be removed from the download page, and all adapter modifications and commits will be expected to respect these standards.

- Floors need to be read from the floors module using designated fields
- Video parameters need to be read from the ad unit
- Bidders need to pass adomain and mediaType in the meta object

Adoption of the RTD framework

- Bidders in Prebid 5.0 will be expected to bid, which means that any providers using the bid infrastructure today that could be passed via the RTD framework will be expected to migrate to this framework.

SharedID and pubcommonID

- We are expecting a change that will couple the two userID modules together, which may involve some legal work on behalf of publishers

patmmccann:
In the pub committee meeting today, it was proposed that https://github.com/prebid/Prebid.js/pull/5863 should need double opt in by publishers, where they would both have to install the module and configure it for it to work.

Fawke:
@gglas Just wanted to know if we're gonna include [this](https://github.com/prebid/Prebid.js/issues/4810) in 5.0 ?

gglas:
@Fawke yes I believe so, but we'll be scheduling a follow up meeting in the next few weeks to list out 5.0 more explicitly.

gglas:
need to include https://github.com/prebid/Prebid.js/issues/5986

patmmccann:
#6464 obscure change that should affect no one but it seems technically breaking

Denis-Admixer:
Regarding the **Taxonomy Enforcement**: do we already have a strict list of the required fields? Except the meta.advertiserDomain?

patmmccann:
[...]

PR #6116, opened 2020-12-10
Add Adomain Targeting Key and Default Targeting Keys

vkimcm:
## Type of change

- [ ] Bugfix
- [ ] Feature
- [ ] New bidder adapter   
- [ ] Build related changes
- [ ] CI related changes
- [x] Other

## Description of change
This PR adds adomain to the targeting keys as hb_adomain.  Adomain is retrieved from bid's meta data named 'advertiserDomains'.
Default targeting key constant will be added so that allowed keys will be default to current standard targeting without adomain when allowTargetingKeys config is not set.
 
## Other information
https://github.com/prebid/Prebid.js/issues/3115


mike-chowla:
Targeting key size is often an issue.  This need PR need more explanation as to why adomain should be a standard targeting key.  Many bidders do no send it and if only a minority of publishers need, it should be custom targeting key rather than a default one

patmmccann:
Since https://github.com/prebid/Prebid.js/issues/5966 has official refactor of standard versus optional keys; and https://github.com/prebid/Prebid.js/pull/5763 enables this behavior, I think it does make sense to make this key dependent on the setConfig outlined in 5763.

vkimcm:
Updated to code to add targeting keys to be defaulted when allowTargetingKeys is not set.

jsnellbaker (approved):
I think this looks good.  

From my testing, it was only adding the new 'hb_adomain' targeting key (and it's bidder specific variant) when the `allowTargetingKeys` feature was used and had included the new `ADOMAIN` field.

Without the `allowTargetingKeys` feature was not enabled, the `hb_adomain` targeting key was not included.

@vkimcm If you haven't already, can you please also create a docs PR to include the ADOMAIN entry to the `allowTargetingKeys` table so publishers would know it's available for use?  Thanks.

vkimcm:
@jsnellbaker thank you for looking.  here is pr for the doc update.  https://github.com/prebid/prebid.github.io/pull/2610

robertrmartinez (approved):
Yeah looks good.

This will be nice and easily extendable to add more targeting keys which will not just always be sent in the future.

Thanks!

robertrmartinez:
@vkimcm Can you please pull the latest from master?

Think the test failures may be related.

Want to make sure it is not because of this PR before merging it.

vkimcm:
@robertrmartinez done!

PR #6122, opened 2020-12-11
Rubicon Bid Adapter FPD Update

mmoschovas:
Update to consolidate applying FPD to both banner and video requests FPD will be merged using global defined FPD, ad unit FPD, and rubicon bidder param FPD. Validation logic with warning logs added

## Type of change

- [ ] Bugfix
- [ ] Feature
- [ ] New bidder adapter   
- [X ] Refactoring (no functional changes, no api changes)
- [ ] Build related changes
- [ ] CI related changes
- [ ] Other

## Description of change

Update to consolidate applying FPD to both banner and video requests FPD will be merged using global defined FPD, ad unit FPD, and rubicon bidder param FPD. Validation logic with warning logs added

```
{
  bidder: '<bidder name>',
  params: {
    // ...
  }
}
```


modules/rubiconBidAdapter.js
```diff
@@ -949,6 +882,63 @@ function addVideoParameters(data, bidRequest) {
...
+  };
+
+  Object.keys(fpd).filter(value => fpd[value] && map[value] && typeof fpd[value] === 'object').forEach((type) => {
+    obj[map[type].code] = Object.keys(fpd[type]).filter(value => fpd[type][value]).reduce((result, key) => {
+      if (key === 'keywords') {
+        if (!Array.isArray(fpd[type][key]) && mediaType === BANNER) fpd[type][key] = [fpd[type][key]]
+
+        result[key] = fpd[type][key];
+
+        if (mediaType === BANNER) keywords = keywords.concat(fpd[type][key]);
```
ncolletti, line 908: The [setBidderConfig docs ](https://docs.prebid.org/features/firstPartyData.html) have keywords value as a comma separated string. When passing the data this way, the merge for **keywords** between fpd.context.keywords, fpd.user.keywords, and params.keywords the **kw** property will be missing the params.keywords value.

robertrmartinez: Yeah, so we should add logic to convert our bidParams.keywords into a comma separated string.

if (bidRequest.params.keywords) bidFpd.context.keywords = bidRequest.params.keywords.join(',');

robertrmartinez: I _think_ this would solve it

modules/rubiconBidAdapter.js
```diff
@@ -949,6 +882,63 @@ function addVideoParameters(data, bidRequest) {
...
 }
 
+function applyFPD(fpd, mediaType, data) {
+  const map = {user: {banner: 'tg_v.', code: 'user'}, context: {banner: 'tg_i.', code: 'site'}, adserver: 'dfp_ad_unit_code'};
+  let obj = {};
+  let keywords = [];
+  const validate = function(e, t) {
+    if (typeof e === 'object' && !Array.isArray(e)) {
+      utils.logWarn('Rubicon: Filtered FPD key: ', t, ': Expected value to be string, integer, or an array of strings/ints');
+    } else if (e) {
```
robertrmartinez, line 892: If the value passed in is falsey I think this will not do what we want.

For example if someone wanted to pass in the literal number zero for some reason `0`.

modules/rubiconBidAdapter.js
```diff
@@ -949,6 +882,63 @@ function addVideoParameters(data, bidRequest) {
   data.imp[0].video.h = size[1]
 }
 
+function applyFPD(fpd, mediaType, data) {
+  const map = {user: {banner: 'tg_v.', code: 'user'}, context: {banner: 'tg_i.', code: 'site'}, adserver: 'dfp_ad_unit_code'};
+  let obj = {};
+  let keywords = [];
+  const validate = function(e, t) {
```
robertrmartinez, line 889: Can we name these vars something more useful?

The build will minify them, but for editing single letter vars are not ideal.

test/spec/modules/rubiconBidAdapter_spec.js
```diff
@@ -1890,14 +1907,14 @@ describe('the rubicon adapter', function () {
...
-          expect(request.data.imp[0].ext.context.data.pbadslot).to.equal('1234567890');
+          expect(request.data.site.ext.data.pbadslot).to.equal('1234567890');
         });
 
         it('should include GAM ad unit in bid request', function () {
           createVideoBidderRequest();
           bidderRequest.bids[0].fpd = {
             context: {
-              adserver: {
+              adServer: {
```
robertrmartinez, line 1917: Was this a previous bug being fixed?

Were we looking at all lowercase when it should have been camelCase the whole time?

test/spec/modules/rubiconBidAdapter_spec.js
```diff
@@ -1909,8 +1926,8 @@ describe('the rubicon adapter', function () {
           );
 
           const [request] = spec.buildRequests(bidderRequest.bids, bidderRequest);
-          expect(request.data.imp[0].ext.context.data.adserver.adslot).to.equal('1234567890');
-          expect(request.data.imp[0].ext.context.data.adserver.name).to.equal('adServerName1');
+          expect(request.data.site.ext.data.adserver.adslot).to.equal('1234567890');
+          expect(request.data.site.ext.data.adserver.name).to.equal('adServerName1');
```
robertrmartinez, line 1930: I do not think this is right.

PBS expects it in the `imp[].ext.context.data.adserver.adslot`

Because if you have multiple imps in a PBS request, you cannot just have a single adSlot in the top level site object.

I think the logic for adServer / slotName / pbAdSlot stuff needs to be outside of the applyFPD so it can be placed in the right spots. (Or you need to refference the imp[0].ext.context.data.adserver.adslot inside the function most likely.

modules/rubiconBidAdapter.js
```diff
@@ -949,6 +882,63 @@ function addVideoParameters(data, bidRequest) {
...
+      if (key === 'keywords') {
+        if (!Array.isArray(fpd[type][key]) && mediaType === BANNER) fpd[type][key] = [fpd[type][key]]
+
+        result[key] = fpd[type][key];
+
+        if (mediaType === BANNER) keywords = keywords.concat(fpd[type][key]);
+      } else if (key === 'data') {
+        utils.mergeDeep(result, {ext: {data: fpd[type][key]}});
+      } else if (key === 'adServer' || key === 'pbAdSlot') {
+        (key === 'adServer') ? ['name', 'adSlot'].forEach(name => {
```
robertrmartinez, line 912: These go inside of the imp array. Not at top level of the request.

robertrmartinez (changes requested):
Couple things to update I think.

Most minor, but I think the main issue is with adServerName / slot not going in the imp array. As of now it is being set on the top level site object which does not make sense since you can have many imps in a single pbs request.

And I know the Rubicon adapter only sends one IMP, but I doubt PBS has logic to look elsewhere besides IMP's for it.

modules/rubiconBidAdapter.js
mmoschovas: Corrected this. Note this is setup as an overwrite not a merge. Priority as follows:
1. params.keywords
2. bidRequest.keywords
3. config.fpd.keywords

modules/rubiconBidAdapter.js
mmoschovas: Updated

test/spec/modules/rubiconBidAdapter_spec.js
mmoschovas: Yes this should have been camelCase

test/spec/modules/rubiconBidAdapter_spec.js
mmoschovas: Moved back to imp[0] as opposed to top level

modules/rubiconBidAdapter.js
mmoschovas: Changed to imp[0]

modules/rubiconBidAdapter.js
```diff
@@ -255,11 +255,11 @@ export const spec = {
       }
 
       const bidFpd = {
-        user: bidRequest.params.visitor || {},
-        context: bidRequest.params.inventory || {}
+        user: {...bidRequest.params.visitor} || {},
+        context: {...bidRequest.params.inventory} || {}
```
robertrmartinez: Not sure this will work as expected.

Why did you change it from the previous

```
      const bidFpd = {
        user: bidRequest.params.visitor || {},
        context: bidRequest.params.inventory || {}
      }
```

modules/rubiconBidAdapter.js
```diff
@@ -516,11 +516,11 @@ export const spec = {
     }
 
     const bidFpd = {
-      user: bidRequest.params.visitor || {},
-      context: bidRequest.params.inventory || {}
+      user: {...bidRequest.params.visitor} || {},
+      context: {...bidRequest.params.inventory} || {}
```
robertrmartinez: Same here.

modules/rubiconBidAdapter.js
robertrmartinez: You could just leave it as:

```
{...bidRequest.params.visitor}
{...bidRequest.params.inventory}
```

And it will do what is intended (default to {})

The || {} never will get executed.

modules/rubiconBidAdapter.js
```diff
@@ -885,21 +885,22 @@ function addVideoParameters(data, bidRequest) {
...
-    } else if (e) {
-      return (Array.isArray(e)) ? e.filter(value => {
-        if (typeof value !== 'object' && value) return value;
-
-        utils.logWarn('Rubicon: Filtered value: ', value, 'for key', t, ': Expected value to be string, integer, or an array of strings/ints');
-      }).toString() : e.toString();
+  const validate = function(prop, key) {
+    if (typeof prop === 'object' && !Array.isArray(prop)) {
+      utils.logWarn('Rubicon: Filtered FPD key: ', key, ': Expected value to be string, integer, or an array of strings/ints');
+    } else if (utils.isNumber(prop) || prop) {
```
robertrmartinez: I think a simple 

```
prop || prop === 0
```

would be a little more efficient. Most of the time the prop will not be falsey, so we should check that first regardless.

And we know it will only hit the other side of the `||` if it is falsey, and only falsey number is zero so better to strict check that.

modules/rubiconBidAdapter.js
```diff
@@ -516,11 +516,11 @@ export const spec = {
...
 
     const bidFpd = {
-      user: bidRequest.params.visitor || {},
-      context: bidRequest.params.inventory || {}
+      user: {...bidRequest.params.visitor} || {},
+      context: {...bidRequest.params.inventory} || {}
     };
 
-    if (bidRequest.params.keywords) bidFpd.context.keywords = bidRequest.params.keywords;
+    if (bidRequest.params.keywords) bidFpd.context.keywords = (utils.isArray(bidRequest.params.keywords)) ? bidRequest.params.keywords.join(',') : bidRequest.params.keywords;
```
robertrmartinez, line 523: I thought the plan was to not support the string or non array case in bidParams.

```
if (utils.isArray(bidRequest.params.keywords)) bidFpd.context.keywords = bidRequest.params.keywords.join(',');
```

modules/rubiconBidAdapter.js
mmoschovas: I think I removed this already no? I pushed twice this morning

modules/rubiconBidAdapter.js
mmoschovas: Same as above

modules/rubiconBidAdapter.js
mmoschovas: I believe this is outdated as well

modules/rubiconBidAdapter.js
mmoschovas: It was but then I reviewed the existing tests and one test explicitly states that it should not error if set as a string. I dont like it but will assume there is a reason for the test that could break things in production. So for the time being I made sure this allows for both

modules/rubiconBidAdapter.js
ncolletti: Could you add a docs PR with the above noted on your bidders page

bretg: https://github.com/prebid/prebid.github.io/pull/2660

modules/rubiconBidAdapter.js
```diff
@@ -254,46 +254,14 @@ export const spec = {
...
-          bidderData.config.fpd.site = siteData;
-        }
-
-        if (!utils.isEmpty(userData)) {
-          bidderData.config.fpd.user = userData;
-        }
+      const bidFpd = {
+        user: {...bidRequest.params.visitor},
+        context: {...bidRequest.params.inventory}
+      };
```
robertrmartinez, line 260: Can we also place lines 257 - 262 into applyFpd function? save a couple duplicate names.

robertrmartinez (approved):
Looks good to me!!

ncolletti (approved):
whew 😅  nice work on this! LGTM

modules/rubiconBidAdapter.js
```diff
@@ -949,6 +868,81 @@ function addVideoParameters(data, bidRequest) {
   data.imp[0].video.h = size[1]
 }
 
+function applyFPD(bidRequest, mediaType, data) {
+  const bidFpd = {
+    user: {...bidRequest.params.visitor},
+    context: {...bidRequest.params.inventory}
```
bretg, line 874: Sorry, upon reviewing this all again, I realize that putting rubicon-specific FPD on global objects is a data leakage mistake.

Please put:
- adunit/imp-level 'user' attributes in openrtb imp.ext.rubicon.visitor
- adunit/imp-level 'context' attributes in openrtb imp.ext.rubicon.inventory

modules/rubiconBidAdapter.js
```diff
@@ -949,6 +868,81 @@ function addVideoParameters(data, bidRequest) {
...
+  };
+
+  if (bidRequest.params.keywords) bidFpd.context.keywords = (utils.isArray(bidRequest.params.keywords)) ? bidRequest.params.keywords.join(',') : bidRequest.params.keywords;
```
bretg, line 877: Likewise, adunit/imp-level 'keywords' should go into opentrb imp.ext.rubicon.keywords

modules/rubiconBidAdapter.js
```diff
@@ -949,6 +868,81 @@ function addVideoParameters(data, bidRequest) {
...
+
+  let fpd = utils.mergeDeep({}, config.getConfig('fpd') || {}, bidRequest.fpd || {}, bidFpd);
```
bretg, line 879: Unfortunately, I also think we need to merge adunit/imp-level FPD into imp[].ext.data rather than {site,user}.ext.data. 

And we need to take special care with:
- AdUnit.fpd.context.pbAdSlot --> openrtb imp[].ext.data.pbadslot and fastlane tg_i.pbadslot
- ortb2Imp.ext.data.pbAdSlot --> openrtb imp[].ext.data.pbadslot and fastlane tg_i.pbadslot
- AdUnit.fpd.context.adServer.adSlot --> openrtb imp[].ext.data.adserver.adslot and fastlane tg_i.dfp_ad_unit_code
- ortb2Imp.ext.data.adServer.adSlot --> openrtb imp[].ext.data.adserver.adslot and fastlane tg_i.dfp_ad_unit_code

If the ortb2 changes aren't in scope for this PR, that's cool. I know there's a different PR for that beast.

bretg (changes requested):
FPD is even more complicated at the code level than I feared. Thanks for this Michael.

Upon reviewing everything again, found a couple of problems with my spec regarding adunit-level data.

Also plan to go through the unit tests and send feedback separately.

modules/rubiconBidAdapter.js
bretg: Made a test page and this all actually does work because it's a single-adunit request for a single bidder. Going to resolve these issues.

modules/rubiconBidAdapter.js
bretg: Made a test page and this all actually does work because it's a single-adunit request for a single bidder. Going to resolve these issues.

bretg (changes requested):
Made some local test cases and found only one problem:  the banner scenario is missing skw1,skw2 from keywords. Otherwise looks great.

Sending you the test cases directly.

bretg (approved):
@mmoschovas explained that the banner scenario was missing skw1,skw2 because there's an order of precedence: bid params, adunit, global. Since I don't expect keywords will be added in multiple places, this is ok. Approving.

PR #6123, opened 2020-12-11
Yieldlab - Update Adsize Handling

BertiBauer:
We are passing the adsize in the bid response. For multisize placements its better to use the adsize passed in the response, but as we do not want to break current integration, we will still support use params.adSize to be able to override sizes.

## Type of change

- [ ] Bugfix
- [X] Feature
- [ ] New bidder adapter   
- [ ] Build related changes
- [ ] CI related changes
- [ ] Other

## Description of change

Bid responses from Yieldlab now usually contain the size of the winning ad for a given placement, making it unnecessary to include the sizes of Yieldlab adslots in the publisher's configuration. In order to not break existing setups, if present the size from the configuration is still being used (meaning a publisher only has to remove the adSize parameter from their configuration for Yieldlab adslots in order to use the returned size of the winning ad). As before, the first stated size from the respective ad unit is used as a fallback should neither of the two aforementioned sizes be present.

- contact email of the adapter’s maintainer: solutions@yieldlab.de

- Link to a PR on the docs repo at https://github.com/prebid/prebid.github.io/: https://github.com/prebid/prebid.github.io/pull/2575

### DIFF
--- a/modules/yieldlabBidAdapter.js
+++ b/modules/yieldlabBidAdapter.js
@@ -96,8 +96,8 @@ export const spec = {
       })
 
       if (matchedBid) {
-        const primarysize = bidRequest.sizes.length === 2 && !utils.isArray(bidRequest.sizes[0]) ? bidRequest.sizes : bidRequest.sizes[0]
-        const customsize = bidRequest.params.adSize !== undefined ? parseSize(bidRequest.params.adSize) : primarysize
+        const adUnitSize = bidRequest.sizes.length === 2 && !utils.isArray(bidRequest.sizes[0]) ? bidRequest.sizes : bidRequest.sizes[0]
+        const adSize = bidRequest.params.adSize !== undefined ? parseSize(bidRequest.params.adSize) : (matchedBid.adsize !== undefined) ? parseSize(matchedBid.adsize) : adUnitSize
         const extId = bidRequest.params.extId !== undefined ? '&id=' + bidRequest.params.extId : ''
         const adType = matchedBid.adtype !== undefined ? matchedBid.adtype : ''
         const gdprApplies = reqParams.gdpr ? '&gdpr=' + reqParams.gdpr : ''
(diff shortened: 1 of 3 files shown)


bretg:
docs PR https://github.com/prebid/prebid.github.io/pull/2575

ncolletti:
Does your endpoint determine the bid response size from the params or the supplied mediaTypes.banner.sizes in the request? 

I tested the 728x90 bid in your markdown file but used a bid with only 300x250 in the mediaTypes.banner.sizes array, I also used the back supported params.adSize to specify a 300x250. However, the received response size was a 728x90 creative, which your response confirmed with '728x90' as the 'adsize' but the displayed creative is cut off as a 300x250 since that is what I defined in the params.adSize. This should not be occurring for bidder adapters, they should be reading from mediaTypes.banner.sizes in their request and respond with the correct size.

1. Multisize placements should read the available sizes in the bidRequest, your example should only return what is acceptable for the bidRequest's adunit from mediaTypes.banner.sizes.
2. What scenario would you endpoint NOT return a size? I imagine, only if there was no bid, so the back supporting of params.adSize is confusing and instead should not using the parameter. It's fine if the user still includes the param in their bids due to legacy integrations but it will not be read in latest iterations of the adapter. You could include a logWarn in isBidRequestValid.
3. The bidResponse should apply the height and width of the registered bid directly from the response with parseSize(matchedBid.adsize), so looking at the bidRequest also isn't needed, that step happens during the constructing of the bidRequest to your endpoint in buildRequests.

BertiBauer:
The bid response size is determined by the size(s) defined in our system for the adslot ID(s) from the request (in other words implicitly from the params). This has to be set up in advance by the publisher or by Yieldlab after consultation with the publisher and has always been this way. 

So your test with expecting a 300x250 banner from a 728x90 adslot was bound to fail from the get-go - the adslot 5220336 is defined as having a single size of 728x90 only in our system and will never respond with anything other than 728x90. params.adSize was needed originally for constructing the actual delivery request in the interpretResponse method as that used to require a size; since implementing multisize adslots in our system that is no longer the case and can therefore be omitted. 

As for "This should not be occurring for bidder adapters, they should be reading from mediaTypes.banner.sizes in their request and respond with the correct size" - is this a new requirement? If so, please let us know where in the documentation this is stated as a requirement, since it hasn't been an issue in the past 2.5 years for our bid adapter. As stated above, the way this works is that the publisher has to set up adslots in the Yieldlab system in advance together with the respective sizes these adslots should make bid requests for. Sizes have never been part of the bid request from Prebid to Yieldlab, nor is this the case with direct Yieldlab header bidding integrations without the use of Prebid or similar wrappers.

Regarding when we would not return a size, there are actually two scenarios. One is indeed as you mentioned the case when there is no bid. 

The other case would be when a publisher in the past - both before using Prebid and/or multisize - had decided to not have the size of the winning bid explicitly returned to them as part of their bid response (what information exactly we return as the bid response to the publisher is configurable, and in the past, the size would be implicit from the adslot ID as it would always be a single size adslot). 

The likelyhood of this is very low and should be caught beforehand when testing the setup with the publisher anyway. However, just in case this is not caught or publishers decide to change their setup without speaking to us first, we decided to include the first size from mediaTypes.banner.sizes as a fallback. We will include the suggested logWarn in a future update though, that's a good idea!

I'm hoping this answers your questions - please let me know if anything is still unclear!

ncolletti:
Thank you for clarifying. What struck me as confusing is the size passed from the user will only change the size in the Prebid response. Given your adapter supports multisize placements now, a future update may be helpful to listen to the size in the request if present and return a bid for that size, avoiding scenarios where a multisize placement returns one size but that they override to a second size, resulting in the incorrect sized pbjs creative to serve on the page.

ncolletti:
LGTM

BertiBauer:
Hi Nicholas, thanks for the merge and your input! I'm sure this won't be our last update regarding the size and how it's handled ;). Have a safe and healthy 2021!

PR #6124, opened 2020-12-11
Add the root domain check method as a util method added for use mod…

markaconrad:
…ules

## Type of change

- [ ] Bugfix
- [x] Feature
- [ ] New bidder adapter   
- [ ] Build related changes
- [ ] CI related changes
- [ ] Other

## Description of change
Add a user id module wide way to determine the root domain to be used for cookies


smenzer:
this is great @markaconrad thanks for putting this in!

modules/userId/index.js
```diff
@@ -727,6 +781,7 @@ export function init(config) {
   (getGlobal()).getUserIds = getUserIds;
   (getGlobal()).getUserIdsAsEids = getUserIdsAsEids;
   (getGlobal()).refreshUserIds = refreshUserIds;
+  (getGlobal()).findRootDomain = findRootDomain;
```
FilipStamenkovic, line 784: The plan for this function is to be used within pbjs? So, it won't be executed outside pbjs?  
If that is the case then maybe we should not make this function accessible on `pbjs` namespace, but rather just `exporting` the function and then just using it inside some module?

If you decide to go that way (exporting instead of marking it available on `pbjs`) I suggest you move it in `utils` package.

modules/userId/index.js
```diff
@@ -315,6 +315,60 @@ function hasGDPRConsent(consentData) {
   return true;
 }
 
+/**
+   * Find the root domain
+   * @param {string|undefined} fullDomain
+   * @return {string}
+   */
+function findRootDomain(fullDomain = document.domain) {
```
FilipStamenkovic, line 323: `document.domain` is deprecated and according to [this article](https://developer.mozilla.org/en-US/docs/Web/API/Document/domain) some modern browsers may not support it.

FilipStamenkovic (changes requested):
All in all, PR looks good, just left some small comments.

modules/userId/index.js
markaconrad: Good idea, I have done so

modules/userId/index.js
markaconrad: updated

modules/userId/index.js
FilipStamenkovic: Sorry for going back and forth, I didn't think this through, my bad. Now I see that `utils` may not be the best place to put this function.  
Now  `utils` has dependency to `storageManager.js`, which should not happen. `utils` should not depend on `storageManager`.  
Can you move this function to `userId/index.js` (again) and just leave it as `exported` instead of attaching it on `pbjs` namespace?

markaconrad: Hmm, yeah, I was on the fence about that so that's why I put it in the user module initially but now I'm not sure how my module will get access to it. The linter forbids me from importing directly from userId/index.js. Do you have any thoughts on that? What do you think about adding the root domain to the config object that gets passed into `getId`

markaconrad: Or maybe in userSync.js?

FilipStamenkovic: I would prefer `userSync.js`.  
@smenzer do you have any opinion on this one?

smenzer: I don't think we should add it to the config that gets passed to `getId`, because that is specifically the config for that module. we could add it as a new parameter, though. would need to get passed to `extendId` as well.

the only issue i have with moving it to `userSync.js` is that we aren't even using that module today within userId module, so seems odd to put it there just because it's somewhat related.

since we can't use `utils`, which is where I would have expected to put this (but I understand the reason not to), and since putting it in `userId/index.js` won't work for using it within submodules...i think we're really left with two options:

1. put it back on the global `pbjs` namespace
2. pass it in the function calls to the modules

if we pass the result of the function to all modules, then it means we're executing that code EVERY time, which isn't necessary since most modules aren't using it today. we could pass it as a function to be used if necessary, but not sure if that's a good idea or not. which leads us back to putting it in the `pbjs` namespace even though it may not be used outside of `pbjs` ... 

not a real answer i know, but that's how i'm thinking about it. maybe we can get some other opinions in the slack group, too?

markaconrad: Checking in to see if there is consensus on an approach that I should take, thanks!

GLStephen:
This is mostly being a contrarian, but trying to unlock the requirements, scoping and accuracy Gordian knot: 

Why not just make root domain a publisher supplied config value? Remove the need for storage. Move it to utils and remove the odd "search" code. I'm not passing judgement on the code, but the cookie "search" method just has a smell to me. Why not let the pub set the root domain if they want to use functionality that uses it?

bretg:
I agree with @FilipStamenkovic that utils is not allowed to require storagemanager. 

But I agree we're not allowed to set even test cookies without GDPR Purpose 1 approval, so storagemanager is required.

In general, I don't like this approach of setting cookies this way -- it feels like a hack. But maybe I just don't understand it. Is it the case that all browsers prevent cookies to "co.uk", so the trick is that setting a cookie against "co.uk" will fail and then "example.co.uk" will succeed?  This is better than listing all possible TLDs, but only marginally.

Stepping back, the User ID module has been out for a long time -- why is this problem just now coming up? 

Overall, I like @GLStephen 's solution better -- would like to hear from more publishers about this. If this is a real problem, pubs should have ample incentive to pass in the domain.

markaconrad:
To provide motivation behind this change:
1. By using the root domain, the same cookies can be shared across domain.com, x.domain.com, and y.domain.com - thus allowing the same user id to be on each
2. There is no good way to maintain an authoritative list of TLDs. While several on this thread have referred to the method we employed as a hack,
we believe it is the most accurate, cheap, and effective method for determining the root domain without bringing in external dependencies

If it is preferred for us to isolate our change without our own user id module, we are glad to do so. This originated out of a desire expressed by reviewers on a different PR to make the root domain finder be a generic change (See https://github.com/prebid/Prebid.js/pull/6063#discussion_r533532794), but if
that is no longer desirable, we can revert to putting it directly in our user id module so we can move forward.

bretg:
> we believe it is the most accurate, cheap, and effective method for determining the root domain

I get that determining the 'root' domain is a pain. Please explain how this approach works. The underlying assumption appears to be that major browsers contain their own list of TLDs and error when appropriate?

i.e. will a set-cookie on co.uk fail while example.co.uk would work? How about domains like .tv?

smenzer:
@bretg fyi that this is the same approach being used by pubcommon (https://github.com/prebid/Prebid.js/blob/master/modules/pubCommonIdSystem.js#L287-L309) which is why I suggested that we get it out of user id modules and put it some place it can be reused and maintained for best practices (and updates if necessary in the future).

markaconrad:
> i.e. will a set-cookie on co.uk fail while example.co.uk would work? How about domains like .tv?

Yes, that is core concept. Split the hostname by the period. Start with the rightmost 2 pieces (eg. co.uk or valid.tv). If a short lived cookie was able to be written, we found the root domain! Otherwise, continue adding one piece as we move leftwards through the full hostname until something is writeable.

markaconrad:
Now that the holidays are past, checking in to see if we have a path forward on this

smenzer:
my two cents is that we should enable this feature (rather than making it pub configurable) and make it usable by any id module that wants it. if importing from there to a submodule is difficult, what if we just make it a property that gets passed to the submodule functions? if we want to move to a pub config longer term, we should introduce that option, make it override the value this method would have otherwise determined (i.e. if it's provided, this function returns the value, else it searches for it), and we can phase it out over time (or at a breaking change release). otherwise, my fear is that lotame/pubcommon/etc would all end up still having their own version of this running multiple times on a page if the publisher hasn't updated their config...

FilipStamenkovic (changes requested):

modules/userId/index.js
```diff
@@ -48,6 +48,20 @@
  * @type {string}
  */
 
+/**
+ * @property
+ * @summary use a predefined domain override for cookies or provide your own
+ * @name Submodule#domainOverride
```
smenzer, line 54: i don't see this property used anywhere...do we still need it?

src/utils.js
```diff
@@ -3,6 +3,7 @@ import { config } from './config.js';
 import clone from 'just-clone';
 import find from 'core-js-pure/features/array/find.js';
 import includes from 'core-js-pure/features/array/includes.js';
+import { getCoreStorageManager } from './storageManager.js';
```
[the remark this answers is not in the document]
smenzer: yea, can't we drop these changes now that we moved it out of utils?

smenzer (commented):
by attaching the `findRootDomain` function to the submodule, does this mean that within each submodule's code, they can just access the method directly now?

two other comments but overall this approach LGTM

modules/userId/index.js
markaconrad: I discovered it when doing more research. It was already in use in `setStoredValue` and in pubCommonIdSubmodule so I just added the documentation for it. So yes, it seems confusing because I just added the doc line about existing code that I did not modify

markaconrad:
> by attaching the `findRootDomain` function to the submodule, does this mean that within each submodule's code, they can just access the method directly now?
> 
> two other comments but overall this approach LGTM

Yes, that is the idea, it is now accessible inside a userId module

src/utils.js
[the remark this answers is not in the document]
markaconrad: Missed that, will remove

modules/userId/index.js
smenzer: ah ok. it looked familiar but i thought maybe it was from a previous version of this PR. all good, thanks for adding the doc!

smenzer (approved):
LGTM

markaconrad:
Checking in to see when we will be able to get this merged in. For context, I have some more changes coming to our user module that are waiting on getting this part through. Thanks!

smenzer:
@FilipStamenkovic Can you take another look here and see if this is good for you? thanks!

src/utils.js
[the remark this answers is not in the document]
FilipStamenkovic: I'm sorry, I missed this.

FilipStamenkovic (approved):
LGTM

PR #6133, opened 2020-12-15
Add support for tagId parameter

readpeaktuomo:
## Type of change

- [ x ] Feature

## Description of change
This PR adds support for the optional tagId parameter.

Documentation PR: https://github.com/prebid/prebid.github.io/pull/2581

### DIFF
--- a/modules/readpeakBidAdapter.md
+++ b/modules/readpeakBidAdapter.md
@@ -4,7 +4,7 @@ Module Name: ReadPeak Bid Adapter
 
 Module Type: Bidder Adapter
 
-Maintainer: kurre.stahlberg@readpeak.com
+Maintainer: tuomo.tilli@readpeak.com
 
 # Description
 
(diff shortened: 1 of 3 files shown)


readpeaktuomo:
The failed BrowserStack test looks to be unrelated to the changes. How do I go about fixing it?

modules/readpeakBidAdapter.md
```diff
@@ -4,7 +4,7 @@ Module Name: ReadPeak Bid Adapter
 
 Module Type: Bidder Adapter
 
-Maintainer: kurre.stahlberg@readpeak.com
+Maintainer: tuomo.tilli@readpeak.com
```
smenzer, line 7: I would recommend setting this to an internal alias so that you don't have to change it and can include multiple people in case one is not available

readpeaktuomo: That's a good idea thanks! Updated.

aleksatr:
LGTM

PR #6139, opened 2020-12-17
Eplanning: new prioritization metric for adunit sizes

chino117:
## Type of change
- [X] Feature

## Description of change
New prioritization metric for adunit sizes

chino117@hotmail.com

## Other information
The changes are tested for the mobile case, for the desktop case I did not know the correct way to test it, do you have any suggestions?

### DIFF
--- a/modules/eplanningBidAdapter.js
+++ b/modules/eplanningBidAdapter.js
@@ -16,7 +16,7 @@ const NULL_SIZE = '1x1';
 const FILE = 'file';
 const STORAGE_RENDER_PREFIX = 'pbsr_';
 const STORAGE_VIEW_PREFIX = 'pbvi_';
-const isMobile = isMobileDevice();
+const mobileUserAgent = isMobileUserAgent();
 const PRIORITY_ORDER_FOR_MOBILE_SIZES_ASC = ['1x1', '300x50', '320x50', '300x250'];
 const PRIORITY_ORDER_FOR_DESKTOP_SIZES_ASC = ['1x1', '970x90', '970x250', '160x600', '300x600', '728x90', '300x250'];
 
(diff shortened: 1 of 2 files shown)


Rothalack:
docs PR prebid/prebid.github.io#2573

Rothalack:
@chino117 Could you provide the bid script you used to test mobile? It'll help me make sure I'm testing this properly as well and hopefully can go from there to test desktop.

chino117:
These changes are only checked with unit tests

Rothalack:
What do you mean by the desktop case not being tested, could you explain how they are different?

chino117:
The change made prioritizes sizes based on whether it is mobile or desktop printing. Unit test was achieved for the mobile case but not for the desktop case

Rothalack:
If it's only mobile tests running, that would mean you don't have a describe() set up where isMobileDevice() resolves false right? Wouldn't that just take a new describe block with desktop dimensions? I could be wrong, I'm still learning to be honest.

chino117:
I have tried using the stub functionality provided by the sinon library, but isMobileDevice is a private function, so I can't use it in testing, I don't see another way to do it, do you see another way?

Rothalack:
I went around looking to see if anyone else had done something like I was thinking. What I'm thinking is getting isMobileDevice to resolve false naturally in a test, not call it directly. And I'd expect you to be able to do that by having a describe block where the width of the screen is 1025px+, since your isMobileDevice function is checking against that as well as the others.

I see an example of one way to do this in MediaNets adapter
https://github.com/prebid/Prebid.js/blob/9aeb61c390f555290db9fcb04bbe48ce92f10b55/test/spec/modules/medianetBidAdapter_spec.js#L676

You can see screen getting set to 1000x1000 there. Maybe there's other ways to do it as well.

Rothalack:
This looks good!

PR #6152, opened 2020-12-28
Pubmatic Adapter:  fix issue where using an adUnit outstream renderer throws an error

pm-shashank-jain:
## Type of change
- [X] Bugfix

## Description of change
For mediaType as outstream video, pubmatic takes an mandatory parameter outstreamAU even though renderer is present causing the publisher to not use their own renderer. Fixing the condition so that if outstreamAU is optional if renderer is present.


modules/pubmaticBidAdapter.js
```diff
@@ -861,8 +861,8 @@ export const spec = {
             utils.logError(`${LOG_WARN_PREFIX}: no context specified in bid. Rejecting bid: `, bid);
             return false;
           }
-          if (bid.mediaTypes[VIDEO].context === 'outstream' && !utils.isStr(bid.params.outstreamAU)) {
-            utils.logError(`${LOG_WARN_PREFIX}: for "outstream" bids outstreamAU is required. Rejecting bid: `, bid);
+          if (bid.mediaTypes[VIDEO].context === 'outstream' && !utils.isStr(bid.params.outstreamAU) && !bid.hasOwnProperty('renderer')) {
```
ncolletti, line 864: This looks good to me. You may want to consider eventually looking within bid.mediaTypes[VIDEO] for the renderer as Publishers do have the ability to define on this level for the specific mediaType context in multiformat adunits: https://docs.prebid.org/dev-docs/show-outstream-video-ads.html#renderers

ncolletti:
Should your markdown be updated to specify that outstreamAu is optional if renderer is defined?

pm-shashank-jain:
@ncolletti , updated the md file and condition to check for renderer in video as well

ncolletti (approved):
LGTM

PR #6159, opened 2021-01-04
SpotX Bid Adapter: default to 4/3 aspect ratio when response doesn't contain w or h

agdillon:
## Type of change

- [x] Bugfix
- [ ] Feature
- [ ] New bidder adapter   
- [ ] Build related changes
- [ ] CI related changes
- [ ] Other

## Description of change

In certain scenarios the SpotX server response does not include a width and height. For an outstream ad using playersize_auto_adapt, we currently log an error if we don't receive w and h, and the ad player does not expand. This change uses a default 4/3 ratio if we don't receive w or h, so the player can still expand.

- contact email of the adapter’s maintainer: adillon@spotx.tv

### DIFF
--- a/modules/spotxBidAdapter.js
+++ b/modules/spotxBidAdapter.js
@@ -51,7 +51,7 @@ export const spec = {
           return false;
         }
         if (!utils.getBidIdParameter('slot', bid.params.outstream_options)) {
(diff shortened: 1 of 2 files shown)


mmoschovas (approved):
LGTM - only thing to consider is condensing the outstream tests lines window.document.getElementById.restore(); to an afterEach(...)

PR #6167, opened 2021-01-05
Add tapad id submodule

moeroach94:
## Type of change

- [X] Feature

## Description of change
Adds Tapad Id System to the user id module

https://github.com/prebid/prebid.github.io/pull/2606

### DIFF
--- a/modules/tapadIdSystem.js
+++ b/modules/tapadIdSystem.js
@@ -1,0 +1,78 @@
+import { uspDataHandler } from '../src/adapterManager.js';
+import { submodule } from '../src/hook.js';
+import { getStorageManager } from '../src/storageManager.js';
+import * as ajax from '../src/ajax.js'
+import * as utils from '../src/utils.js';
+
+export const storage = getStorageManager();
+const cookiesMaxAge = 7 * 24 * 60 * 60 * 1000; // 60 days
+export const defaultExpirationString = new Date(utils.timestamp() + cookiesMaxAge).toString();
+export const pastDateString = new Date(0).toString();
+export const tapadCookieKey = 'tapad_id';
+export const graphUrl = 'https://realtime-graph-access-zxvhwknfeq-uc.a.run.app/v1/graph';
+
+function saveOnAllStorages(key, value, expiration = defaultExpirationString) {
+  storage.setCookie(key, typeof value === 'object' ? JSON.stringify(value) : value, expiration);
+  storage.setDataInLocalStorage(key, JSON.stringify(value));
+}
+
+function deleteFromAllStorages(key) {
+  storage.setCookie(key, undefined, pastDateString);
+  storage.removeDataFromLocalStorage(key);
+}
+
+export const tapadIdSubmodule = {
+  name: 'tapadId',
+  /**
+   * decode the stored id value for passing to bid requests
+   * @function
+   * @returns {{tapadId: string} | undefined}
+   */
+  decode(id) {
+    return { tapadId: id };
+  },
+  /*
+   * @function
+   * @summary initiate Real Time Graph
+   * @param {SubmoduleParams} [configParams]
+   * @param {ConsentData} [consentData]
+   * @returns {IdResponse }}
+   */
+  getId(config) {
+    const uspData = uspDataHandler.getConsentData();
+    if (uspData && uspData !== '1---') {
+      return { id: undefined };
+    }
+    const configParams = config.params || {};
+    const expiration = config.storage && config.storage.expires
+      ? new Date(utils.timestamp() + config.storage.expires * 24 * 60 * 60 * 1000).toString()
+      : undefined;
+
+    return {
+      callback: (complete) => {
+        ajax.ajaxBuilder(10000)(
+          `${graphUrl}?company_id=${configParams.companyId}&tapad_id_type=TAPAD_ID`,
+          {
+            success: (response) => {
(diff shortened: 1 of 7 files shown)


modules/tapadIdSystem.js
```diff
@@ -0,0 +1,78 @@
...
+
+function saveOnAllStorages(key, value, expiration = defaultExpirationString) {
+  storage.setCookie(key, typeof value === 'object' ? JSON.stringify(value) : value, expiration);
+  storage.setDataInLocalStorage(key, JSON.stringify(value));
+}
+
+function deleteFromAllStorages(key) {
+  storage.setCookie(key, undefined, pastDateString);
+  storage.removeDataFromLocalStorage(key);
+}
```
smenzer, line 22: In your documentation example (https://github.com/prebid/prebid.github.io/pull/2606/files) you have the publisher set the `storage` object, yet you are manually storing data here. If `storage` is set, then the user id module will handle storing and retrieving the id; if you want to do it manually then publishers cannot set `storage` or you will end up with inconsistencies. 

Also, if you are going to manage storage on your own, you need to set an expiration item in the local storage and manually handle throwing it out when reading the data if it's expired. 

i would suggest just using the built-in functionality of the user id module to read/write the cached id value.

modules/tapadIdSystem.js
```diff
@@ -0,0 +1,78 @@
...
+export const storage = getStorageManager();
+const cookiesMaxAge = 7 * 24 * 60 * 60 * 1000; // 60 days
+export const defaultExpirationString = new Date(utils.timestamp() + cookiesMaxAge).toString();
+export const pastDateString = new Date(0).toString();
+export const tapadCookieKey = 'tapad_id';
+export const graphUrl = 'https://realtime-graph-access-zxvhwknfeq-uc.a.run.app/v1/graph';
+
+function saveOnAllStorages(key, value, expiration = defaultExpirationString) {
+  storage.setCookie(key, typeof value === 'object' ? JSON.stringify(value) : value, expiration);
+  storage.setDataInLocalStorage(key, JSON.stringify(value));
```
smenzer, line 16: why are you checking that `value` is an object before `JSON.stringify` for cookies but not local storage?

modules/tapadIdSystem.js
```diff
@@ -0,0 +1,78 @@
...
+    }
+    const configParams = config.params || {};
+    const expiration = config.storage && config.storage.expires
+      ? new Date(utils.timestamp() + config.storage.expires * 24 * 60 * 60 * 1000).toString()
+      : undefined;
+
+    return {
+      callback: (complete) => {
+        ajax.ajaxBuilder(10000)(
+          `${graphUrl}?company_id=${configParams.companyId}&tapad_id_type=TAPAD_ID`,
```
smenzer, line 54: if the publisher doesn't specify a `companyId` this line will throw an error. you should check that it exists first, and since it's a required field, error out so the publisher knows

test/spec/modules/tapadIdSystem_spec.js
```diff
@@ -0,0 +1,55 @@
...
+      const callback = tapadIdSubmodule.getId({}).callback
+      callback(callbackSpy);
+
+      const request = server.requests[0]
+
+      request.respond(404)
+      expect(callbackSpy.lastCall.lastArg).to.be.undefined;
+      expect(setCookieSpy.lastCall.calledWith(tapadCookieKey, undefined, pastDateString)).to.be.true
+    })
+    it('should log message with invalid company id', function() {
```
smenzer, line 42: you should also have a test for a missing `companyId` since it's a required field

modules/tapadIdSystem.js
```diff
@@ -0,0 +1,78 @@
...
+import * as utils from '../src/utils.js';
+
+export const storage = getStorageManager();
+const cookiesMaxAge = 7 * 24 * 60 * 60 * 1000; // 60 days
+export const defaultExpirationString = new Date(utils.timestamp() + cookiesMaxAge).toString();
+export const pastDateString = new Date(0).toString();
+export const tapadCookieKey = 'tapad_id';
+export const graphUrl = 'https://realtime-graph-access-zxvhwknfeq-uc.a.run.app/v1/graph';
+
+function saveOnAllStorages(key, value, expiration = defaultExpirationString) {
```
smenzer, line 14: is there a reason to set a default expiration string, since it's never being used?

smenzer (changes requested):
I left several comments inline. A few other general items:

1. You need to add your id module to `eids.js` and related `.md` and `_spec` files
3. You need to add your module to the `userId_spec.js` file, especially since your testing is pretty light
4. You probably want to add your id module to `userId.md` and potentially your own `.md` file with more details
4. Purely a code style thing, but you are very inconsistent with line endings...sometimes you use a `;` and sometimes you don't use anything.

moeroach94:
Okay I've added to the user_id and eids files and added a log for a missing or malformed companyId. I'd still like to make the ajax call even if companyId is not what we expected. We'd still like to record those requests and since configParams is at least guaranteed to be an object, accessing the companyId key will at worst be undefined which can be successfully interpolated into the string

smenzer:
@moeroach94 looking good, thanks for the updates. Do you have a `companyId` I can use for testing with?

moeroach94:
@smenzer Just wondering with Monday being a holiday, do you think this will be in next week's release?

smenzer (approved):
LGTM

PR #6183, opened 2021-01-12
fix includes and find references in various files

jsnellbaker:
## Type of change

- [x] Bugfix

## Description of change
There were numerous instances where files were using the `array.includes`, `array.find`, `string.includes` functions which are not compliant in IE11.  

I replaced these references with the compliant versions of these functions.


robertrmartinez (approved):
Thank you for doing this @jsnellbaker

Fawke (approved):

smenzer:
given all circleci tests are broken and this is supposed to fix it, and we have two approvals, i'm going to merge this in

PR #6185, opened 2021-01-13
Lotame panorama id glv

markaconrad:
## Type of change

- [ ] Bugfix
- [x] Feature
- [ ] New bidder adapter   
- [ ] Build related changes
- [ ] CI related changes
- [ ] Other

## Description of change
Adds Global Vendor List (GVL) ID for Lotame (95) to the lotamePanoramaIdSystem User ID module as a property and in getting a storage manager. The understanding is this will prevent the GDPR Enforcement module from blocking this module. I don't believe there are any other changes needed to our module directly. It is then on the publisher to set up the pbjc consentManagement config section as appropriate.

- https://github.com/prebid/prebid.github.io/pull/2623


markaconrad:
We have 2 confirming question:
1 - These changes will have 0 effect in the case where GDPR enforcement module is not present
2 - If GDPR enforcement module is present and we don't have consent based on their rules, execution will stop inside getStorageManager and the rest of our code won't run

FilipStamenkovic (approved):
LGTM

PR #6187, opened 2021-01-13
Lotame panorama id system updates

markaconrad:
## Type of change

- [x] Bugfix
- [x] Feature
- [ ] New bidder adapter   
- [ ] Build related changes
- [ ] CI related changes
- [ ] Other

## Description of change
Store all cookies at the root domain
Don't touch our id cookie on a certain error code

## Other information
This uses the new findRootDomain method now inside Lotame's User Id Module


smenzer:
@markaconrad looks like there are circleci errors in your adapter...can you make sure to get those cleaned up, please?

modules/lotamePanoramaIdSystem.js
```diff
@@ -211,10 +228,17 @@ export const lotamePanoramaIdSubmodule = {
           if (response) {
             try {
               let responseObj = JSON.parse(response);
-              saveLotameCache(KEY_EXPIRY, responseObj.expiry_ts);
+              const shouldUpdateProfileId = !(
+                utils.isArray(responseObj.errors) &&
+                responseObj.errors.indexOf(MISSING_CORE_CONSENT) !== -1
+              );
+
+              saveLotameCache(KEY_EXPIRY, responseObj.expiry_ts, responseObj.expiry_ts);
```
blingster7, line 236: should this key-value ever get cleared from the cache?

test/spec/modules/lotamePanoramaIdSystem_spec.js
```diff
@@ -446,4 +446,117 @@ describe('LotameId', function() {
...
+      sinon.assert.neverCalledWith(
+        setCookieStub,
+        '_cc_id',
+        '4ec137245858469eb94a4e248f238694'
+      );
+    });
+
+    it('should save the expiry', function () {
+      sinon.assert.calledWith(setLocalStorageStub, 'panoramaId_expiry', 10);
+
```
blingster7, line 488: whitespace

modules/lotamePanoramaIdSystem.js
markaconrad: It is used to tell us when we should check the server again so removing is the same as not updating it and there is no practical benefit to explicitly clear this value from the cache

blingster7 (approved):
LGTM

markaconrad:
Checking in to see when someone will have time for the 2nd review, thanks!

smenzer:
@markaconrad / @blingster7 this isn't core code so it doesn't require a 2nd review. i'll go ahead and merge it now.

PR #6193, opened 2021-01-14
ID5 User Id Module: move id5's user id extension to the uids array

smenzer:
## Type of change
- [x] Feature

## Description of change
Moving the user id extension we provide into the uids array to better conform to conventions being established by other id providers.

There were only a few places using this field specifically, so this PR also includes the necessary changes to that code to read it properly.

### DIFF
--- a/modules/rubiconBidAdapter.js
+++ b/modules/rubiconBidAdapter.js
@@ -516,7 +516,7 @@ export const spec = {
           } else if (eid.source === 'sharedid.org') {
             data['eid_sharedid.org'] = `${eid.uids[0].id}^${eid.uids[0].atype}^${(eid.uids[0].ext && eid.uids[0].ext.third) || ''}`;
           } else if (eid.source === 'id5-sync.com') {
-            data['eid_id5-sync.com'] = `${eid.uids[0].id}^${eid.uids[0].atype}^${(eid.ext && eid.ext.linkType) || ''}`;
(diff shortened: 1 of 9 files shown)


Fawke (changes requested):
Hi @smenzer,

Looks good, but I just wanted to make sure that every adapter that uses id5 is updated to read it from the new location.

Pulpoint Bid Adapter, [https://github.com/prebid/Prebid.js/blob/c40f4ea2c38c579fab27d8b92cc78059a4926486/modules/pulsepointBidAdapter.js#L423] here, is still using old reference.

Can you just cross check all adapters and update those that needs updating?

smenzer:
@Fawke it's a bit confusing, but actually pulsepoint was fine. They were pulling the `userId` object from the bid request rather than the eids array. in the `userId.id5id` object, there are two fields, `uid` which is our uid string, and `ext` which is our extension object. It's only in the eids array that there are two potential locations for an ext that we are switching, but our `id5id` object is still the same, so pulsepoint doesn't need to change. I did find one place where there wasn't an error, but spotx was manually putting our ext in the old location, so I've corrected that one in the last commit. otherwise, I think we're all set (i did another review of bid adapters and don't see other references to our id/extension.

Fawke (approved):
@smenzer I get what you're saying. That was my concern, but looks like it's fine.

PR #6194, opened 2021-01-14
Price Floors: use standard cpmAdjustment

### LINKED ISSUE #6175: Price Floors module should apply pbjs.bidderSettings.standard.bidCpmAdjustment
## Type of issue
Bug
## Description

The Price Floors module should apply the pbjs.bidderSettings.standard.bidCpmAdjustment function if it exists, to the bid floor. 

## Steps to reproduce
1. Set a bidCpmAdjustment function in standard (and ensure there is no bidder specific bid cpm adjustment function): 
```
pbjs.bidderSettings.standard.bidCpmAdjustment =  function(bidCpm, bid) {
      return bidCpm * .5;
    }
```
2. Ensure that bidAdjustment is set to true: 
```
pbjs.setConfig({
    floors: {
        enforcement: {
            bidAdjustment: true
        }
```
3. Send bid requests with the rubicon bid adapter, (bidder that uses the getFloor function) 
4. When theres a floor match, the bid floor will not be adjusted and **should** be adjusted.

Possible solution is to set this line in priceFloors.js in the getBiddersCpmAdjustment function:
```
const adjustmentFunction = utils.deepAccess(getGlobal(), `bidderSettings.${bidderName}.bidCpmAdjustment`) || utils.deepAccess(getGlobal(), `bidderSettings.standard.bidCpmAdjustment`);
```

robertrmartinez:
## Type of change
- [X] Bugfix
- 
## Description of change
There was an oversight when implementing the cpmAdjustment logic. We forgot to account for the `standard` adjustments which apply to all bidders.

After testing I confirmed that Prebid core takes bidder adjustments first, and if they do not exist, then uses the standard.

So the price floors module has been updated accordingly.

Fixes #6175

### DIFF
--- a/modules/priceFloors.js
+++ b/modules/priceFloors.js
@@ -141,7 +141,7 @@ function generatePossibleEnumerations(arrayOfFields, delimiter) {
  * @summary If a the input bidder has a registered cpmadjustment it returns the input CPM after being adjusted
  */
 export function getBiddersCpmAdjustment(bidderName, inputCpm, bid = {}) {
-  const adjustmentFunction = utils.deepAccess(getGlobal(), `bidderSettings.${bidderName}.bidCpmAdjustment`);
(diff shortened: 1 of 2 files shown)


osazos (approved):
LGTM

GLStephen:
LGTM

PR #6195, opened 2021-01-14
Relevant Yield analytics adapter

samuel-palmer-relevant-digital:
## Type of change
- [x] Analytics adapter

## Description of change
This is an analytics adapter for the "HB Analytics" module in [Relevant Yield](https://www.relevant-digital.com/relevantyield)

The code is simply logging events in a global array that is later consumed by our library script - which in turn needs to be loaded separately on the page.

- contact email of the adapter’s maintainer: samuel.palmer@relevant-digital.com
 
## Other information
The purpose of this adapter is to simplify the integration of our product on clients' websites as it will no longer be necessary to place our "tag" in such way that no events are triggered before we listen to them.

### DIFF
--- a/modules/relevantAnalyticsAdapter.js
+++ b/modules/relevantAnalyticsAdapter.js
@@ -1,0 +1,33 @@
+import adapter from '../src/AnalyticsAdapter.js';
+import adapterManager from '../src/adapterManager.js';
+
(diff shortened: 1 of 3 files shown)


bretg:
docs PR https://github.com/prebid/prebid.github.io/pull/2630

mmoschovas (approved):
LGTM

PR #6214, opened 2021-01-22
Between multirequests

ignat-one:
## Type of change

- [x] Feature

## Description of change

Bid requests are now sent in one POST request

- contact email of the adapter’s maintainer: khaylov@betweenx.com

### DIFF
--- a/modules/betweenBidAdapter.js
+++ b/modules/betweenBidAdapter.js
@@ -3,6 +3,7 @@ import { getAdUnitSizes, parseSizesInput } from '../src/utils.js';
 import { getRefererInfo } from '../src/refererDetection.js';
 
 const BIDDER_CODE = 'between';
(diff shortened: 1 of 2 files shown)


modules/betweenBidAdapter.js
```diff
@@ -74,9 +74,14 @@ export const spec = {
         }
       }
 
-      requests.push({method: 'GET', url: 'https://ads.betweendigital.com/adjson', data: params})
+      requests.push({data: params})
     })
-    return requests;
+    return {
+      method: 'POST',
+      url: 'https://ads.betweendigital.com/adjson?t=prebid',
```
blingster7, line 81: Just a suggestion, but might be a good idea to set this as a const at the top of your adapter? Like `STR_ENDPOINT`?

PR #6227, opened 2021-01-26
JW Player RTD Module - Limit scope to adUnits supporting instream video

karimMourra:
## Type of change

- [x] Bugfix

## Description of change
Prevents data retrieval if adUnit does not support video instream while preserving ability to obtain data if `jwTargeting` is defined on the ad unit

### DIFF
--- a/modules/jwplayerRtdProvider.js
+++ b/modules/jwplayerRtdProvider.js
@@ -159,11 +159,21 @@ export function enrichAdUnits(adUnits) {
   });
 }
 
+function supportsInstreamVideo(mediaTypes) {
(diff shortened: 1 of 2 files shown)


blingster7:
let me know what you think but might be worthwhile to add a couple of more tests where:
- JwTargeting defined with banner adunit
- JwTargeting defined with instream adunit

this test but with jwTargeting specified: `should exclude adUnits that do not support instream video and do not specify jwTargeting'`

but otherwise looks good to me

karimMourra:
@blingster7 thanks for the review, more tests are always better :)

blingster7 (approved):

PR #6236, opened 2021-01-28
SmartAdServer Bid Adapter: image sync and noAd

lowendavid:
## Type of change
- [X] Feature: Sync changes/improvement

## Description of change
Change 1:

When the default iframe sync is not enabled we will now try to image sync if enabled.

if (iframe sync enabled)
{
     sync
}
else if (image sync enabled && sync urls provided)
{
     sync with each url
}

Change 2:
We need to handle noAds in our prebidResponse, in case of noad we don't want to build the bidresponse array.

Change 3:
Added tests to cover the first 2 changes.

### DIFF
--- a/modules/smartadserverBidAdapter.js
+++ b/modules/smartadserverBidAdapter.js
@@ -132,7 +132,7 @@ export const spec = {
     const bidResponses = [];
     let response = serverResponse.body;
     try {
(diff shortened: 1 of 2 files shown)


osazos (approved):
LGTM

PR #6258, opened 2021-02-02
[ParrableIdSystem] Ensure base64 payload is url-safe

icflournoy:
## Type of change

- [x] Bugfix

## Description of change

This fix ensures that the base64-encoded payload that is passed to our service is url-safe by translating unsafe characters to a set of safe characters.

It uses the Base64 rfc-recommended URL safe alphabet https://tools.ietf.org/html/rfc4648#section-5

### DIFF
--- a/integrationExamples/gpt/userId_example.html
+++ b/integrationExamples/gpt/userId_example.html
@@ -109,6 +109,7 @@
         <script src="../../build/dev/prebid.js" async></script>
 
         <script>
+            function getHashedEmail() {}
             var googletag = googletag || {};
             googletag.cmd = googletag.cmd || [];
             googletag.cmd.push(function() {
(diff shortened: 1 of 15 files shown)


harpere (approved):

harpere:
@victorigualada - a little late since this was already merged, but why didn't you just use `encodeURIComponent()` rather than custom encoding?

harpere:
Also - after merging discovered that tests were occasionally failing (for [example](https://app.circleci.com/pipelines/github/prebid/Prebid.js/3765/workflows/36dd9bc0-872b-4447-bbce-119ba44b4646/jobs/11870)), so opened this PR to fix: https://github.com/prebid/Prebid.js/pull/6268

icflournoy:
@harpere @victorigualada we chose to use this encoding because it allows us to work back to standard base64 in backend systems that don't support decoding URIs easily.

Node's buffer now also supports encoding/decoding base64url using this character set: https://github.com/nodejs/node/issues/26512#issuecomment-762186057

Thanks for the test fix PR, that's odd that our test runs (locally, and on this PR) did not catch that test. It should have failed all of the time.

harpere:
thanks for the explanation @icflournoy.  It didn't fail for me all the time either.  I didn't dig in but figured the "data" somehow varied.

PR #6261, opened 2021-02-02
Sharethrough: Add support for ID5, Shared ID, and Live Intent ID

poojawins:
## Type of change
- [x] Feature

## Description of change
- Add support for ID5, Shared ID, and Live Intent ID
- PR on the docs repo: prebid/prebid.github.io#2681
pubgrowth.engineering@sharethrough.com

### DIFF
--- a/modules/sharethroughBidAdapter.js
+++ b/modules/sharethroughBidAdapter.js
@@ -136,11 +136,9 @@ export const sharethroughAdapterSpec = {
 };
 
 function handleUniversalIds(bidRequest) {
-  const universalIds = {};
+  if (!bidRequest.userId) return {};
 
-  if (!bidRequest.userId) {
-    return {};
-  }
+  const universalIds = {};
 
   const ttd = utils.deepAccess(bidRequest, 'userId.tdid');
   if (ttd) universalIds.ttduid = ttd;
--- a/modules/sharethroughBidAdapter.js
+++ b/modules/sharethroughBidAdapter.js
@@ -142,36 +142,25 @@ function handleUniversalIds(bidRequest) {
     return {};
   }
 
-  if (bidRequest.userId.tdid) {
-    universalIds.ttduid = bidRequest.userId.tdid;
-  }
+  const ttd = utils.deepAccess(bidRequest, 'userId.tdid');
+  if (ttd) universalIds.ttduid = ttd;
 
-  if (bidRequest.userId.pubcid) {
-    universalIds.pubcid = bidRequest.userId.pubcid;
-  } else if (bidRequest.crumbs && bidRequest.crumbs.pubcid) {
-    universalIds.pubcid = bidRequest.crumbs.pubcid;
-  }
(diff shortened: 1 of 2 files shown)


modules/sharethroughBidAdapter.js
```diff
@@ -147,6 +135,34 @@ export const sharethroughAdapterSpec = {
...
+  if (ttd) universalIds.ttduid = ttd;
+
+  const pubc = utils.deepAccess(bidRequest, 'userId.pubcid') || utils.deepAccess(bidRequest, 'crumbs.pubcid');
+  if (pubc) universalIds.pubcid = pubc;
+
+  const idl = utils.deepAccess(bidRequest, 'userId.idl_env');
+  if (idl) universalIds.idluid = idl;
+
+  const id5 = utils.deepAccess(bidRequest, 'userId.id5id');
+  if (id5) universalIds.id5uid = id5;
```
smenzer, line 153: just as an fyi, the `userId.id5id` is an object, not a string. the object looks like this:
```javascript
{
  uid: 'ID5-12345', // this is the ID5 Universal ID
  ext: {
    linkType: 2 // possible options are 0, 1, or 2
  }
}
```

It is highly recommended that you pass through the entire object to your servers so you can pass through the `linkType` to DSPs, several of whom have asked for this field. If you have any questions, feel free to reach out to me directly (scott at id5 dot io)

test/spec/modules/sharethroughBidAdapter_spec.js
```diff
@@ -15,7 +16,12 @@ const bidRequests = [
     userId: {
       tdid: 'fake-tdid',
       pubcid: 'fake-pubcid',
-      idl_env: 'fake-identity-link'
+      idl_env: 'fake-identity-link',
+      id5id: 'fake-id5id',
```
smenzer, line 20: as noted above, this is not what the `id5id` looks like, so you'll want to update it to ensure your tests are working properly.

smenzer (changes requested):
see comments inline about the ID5 ID

modules/sharethroughBidAdapter.js
maphe: @smenzer thanks for raising that, we missed it and I'm gonna push a fix soon. Is there a doc somewhere I can link internally to describe the meaning of `linkType` and its possible values?

smenzer: @maphe great to hear! we have some docs on our wiki - https://wiki.id5.io/x/EABH - but basically, `linkType` can be 0 (not linked across domains), 1 (linked across domains with probabilistic signals), or 2 (linked across domains with deterministic signals)

maphe:
@smenzer thanks for the link, that's perfect. Also I addressed your comments 👍 feel free to take another look

modules/sharethroughBidAdapter.js
smenzer: > @smenzer thanks for the link, that's perfect. Also I addressed your comments 👍 feel free to take another look

Looks great!

smenzer (approved):
I've approved the update to the ID5 code specifically, I didn't really review the rest of the change, though

osazos (approved):
LGTM

PR #6269, opened 2021-02-03
Revert "Extended ID permissions supported by bidder"

jsnellbaker:
Reverts prebid/Prebid.js#6112

Observed that calling the PBS adapter without any defined userId modules causes an exception error to be thrown, interrupting the request.  

Additionally, we should not be directly importing functions from module files.  This creates an implicit setup for the other module to always be included - which is not what the publisher would expect.

We are reverting this PR, for now, to allow time for the below issue to be addressed.

```
ERROR: Error processing command : prebid.js TypeError: Cannot read property 'filter' of undefined
    at buildEidPermissions (eids.js:233)
    at getEidPermissions (index.js:218)
    at Object.buildRequest (index.js:708)
    at PrebidServer.baseAdapter.callBids (index.js:978)
    at adapterManager.js:389
    at Array.forEach (<anonymous>)
    at Object.8.adapterManager.callBids (adapterManager.js:355)
    at Object.run (auction.js:248)
    at runIfOriginHasCapacity (auction.js:311)
    at Object.addBidderRequestsCallback (auction.js:280)
```

### DIFF
--- a/integrationExamples/gpt/userId_example.html
+++ b/integrationExamples/gpt/userId_example.html
@@ -236,7 +236,6 @@
                         },
                         {
                         name: "sharedId",
-                        // bidders: ["rubicon", "sampleBidders"], // to allow this ID for specific bidders
                         params: {
                             syncTime: 60 // in seconds, default is 24 hours
                         },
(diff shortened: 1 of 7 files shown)


robertrmartinez (approved):

jdwieland8282:
@YerkovichM Can you take a look at @jsnellbaker 'comments above?

harpere (approved):

PR #6272, opened 2021-02-03
Rubicon Analytics: Pick up wrapper family field

robertrmartinez:
## Type of change
- [X] Other

## Description of change
Prebid Wrappers using the Magnite Service will have a new field available called wrapperFamily which we will need to pass through to our analytics endpoint.

### DIFF
--- a/modules/rubiconAnalyticsAdapter.js
+++ b/modules/rubiconAnalyticsAdapter.js
@@ -167,9 +167,10 @@ function sendMessage(auctionId, bidWonId) {
     referrerHostname: rubiconAdapter.referrerHostname || getHostNameFromReferer(referrer),
     channel: 'web',
   };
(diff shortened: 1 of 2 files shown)


modules/rubiconAnalyticsAdapter.js
```diff
@@ -167,9 +167,10 @@ function sendMessage(auctionId, bidWonId) {
     referrerHostname: rubiconAdapter.referrerHostname || getHostNameFromReferer(referrer),
     channel: 'web',
   };
-  if (rubiConf.wrapperName || rubiConf.rule_name) {
```
robertrmartinez, line 170: wrapperName is there always when rule_name is so we will only check it. same with new field family

harpere (approved):

ChrisHuie (approved):

PR #6275, opened 2021-02-04
Shared ID gdpr support

### LINKED ISSUE #6266: Send consent data to Shared Id endpoint
## Type of issue
Feature request

## Description
In EU, the id.sharedid.org/id endpoint requires consent data in order to return an id which is currently not being sent.

The parameters gdpr and gdpr_consent should be set as follows:

https://id.sharedid.org/id?gdpr=1&gdpr_consent=consent_string

## Other information
https://github.com/prebid/shared-id/issues/20
https://github.com/prebid/Prebid.js/issues/6235

This should be fairly straight forward as getId already passes consentData: https://github.com/prebid/Prebid.js/blob/97e5351c2ed0900557bc09277cce33a4db1a4678/modules/userId/index.js#L644

extendId needs to be modified to pass consentData as well though:
https://github.com/prebid/Prebid.js/blob/97e5351c2ed0900557bc09277cce33a4db1a4678/modules/userId/index.js#L647

bjorn-lw:
## Type of change
- [x] Feature

## Description of change
Send consent data, if available, to Shared ID endpoint.

This PR fixes #6266. Also see: https://github.com/prebid/shared-id/issues/20

Note 1: The tests for SharedIdSystem does not add up to the accepted 80% coverage. No specific Shared Id tests existed before and I have added only tests for the parts that have been modified.
Note 2: The Shared Id endpoint currently has a bug (see issue above). Until that is fixed, this won't be possible to successfully test in a live environment (verifying that consent data is sent will be possible, but there will not be any id returned).

## Other information
#6266. 
https://github.com/prebid/shared-id/issues/20
https://github.com/prebid/shared-id/issues/21
@jdwieland8282

### DIFF
--- a/modules/id5IdSystem.js
+++ b/modules/id5IdSystem.js
@@ -174,10 +174,11 @@ export const id5IdSubmodule = {
    *  It's permissible to return neither, one, or both fields.
    * @function extendId
    * @param {SubmoduleConfig} config
+   * @param {ConsentData|undefined} consentData
    * @param {Object} cacheIdObj - existing id, if any
    * @return {(IdResponse|function(callback:function))} A response object that contains id and/or callback.
    */
-  extendId(config, cacheIdObj) {
+  extendId(config, consentData, cacheIdObj) {
     const partnerId = (config && config.params && config.params.partner) || 0;
     incrementNb(partnerId);
     return cacheIdObj;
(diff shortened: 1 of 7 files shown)


smenzer:
LGTM

PR #6277, opened 2021-02-04
[ParrableIdSystem] Accept list of partners as an array or string

icflournoy:
## Type of change

- [x] Feature

## Description of change

This changes introduces a new parameter (`partners`) to our module that supersedes the old `partner` string-list while still preserving the old behavior. While the old list accepted a comma-separated string of partner IDs we wanted to clean up the parameter name (make it plural) and have it accept a proper array for simplicity. Both `partner` and `partners` now operate the same way and can accept an array, or a comma-separated string.

Docs: https://github.com/prebid/prebid.github.io/pull/2688

### DIFF
--- a/modules/appnexusBidAdapter.js
+++ b/modules/appnexusBidAdapter.js
@@ -999,9 +999,22 @@ function hidedfpContainer(elementId) {
   }
 }
 
+function hideSASIframe(elementId) {
+  try {
+    // find script tag with id 'sas_script'. This ensures it only works if you're using Smart Ad Server.
+    const el = document.getElementById(elementId).querySelectorAll("script[id^='sas_script']");
+    if (el[0].nextSibling && el[0].nextSibling.localName === 'iframe') {
+      el[0].nextSibling.style.setProperty('display', 'none');
+    }
+  } catch (e) {
+    // element not found!
+  }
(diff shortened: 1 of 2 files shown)


icflournoy:
@victorigualada can you take a look at why the timezone filter tests are failing on this unrelated PR? I incorporated the base64 fix from https://github.com/prebid/Prebid.js/pull/6268 in our other test that decodes the data parameter, which solved the first set of failing tests.

harpere (changes requested):
tests are failing.

icflournoy:
@harpere we've fixed the tests relevant to our module, but it appears that some unrelated tests are now causing the build to fail:

```
  pubmatic analytics adapter
    when handling events
      ✗ Logger: best case + win tracker
	Error: Timeout of 2000ms exceeded. For async tests and hooks, ensure "done()" is called; if returning a Promise, ensure it resolves.

      ✗ bidCpmAdjustment: USD: Logger: best case + win tracker
	Error: Timeout of 2000ms exceeded. For async tests and hooks, ensure "done()" is called; if returning a Promise, ensure it resolves.
```

harpere (approved):
reran tests and they're passing now.

PR #6281, opened 2021-02-05
ID5 User Id Module: update a/b testing to be user based not request based

smenzer:
## Type of change
- [x] Maintenance

## Description of change
Changed the logic for our A/B testing to be User based rather than Request based, which will provide more statistically relevant results.


FilipStamenkovic (approved):
LGTM

PR #6283, opened 2021-02-05
[ParrableIdSystem] Add GVLID and handle TC Consent data

icflournoy:
## Type of change

- [x] Feature

## Description of change

This change allows our User ID module to support GDPR via the TCF2 consent string. Since we interact with cookie storage inside of our ID system we pass the GVLID to Prebid's storage manager so that device access consent can be checked for our GVLID. We then forward gdpr applicability and the consent string to our backend systems.

Docs: https://github.com/prebid/prebid.github.io/pull/2693

### DIFF
--- a/modules/appnexusBidAdapter.js
+++ b/modules/appnexusBidAdapter.js
@@ -999,9 +999,22 @@ function hidedfpContainer(elementId) {
   }
 }
 
+function hideSASIframe(elementId) {
+  try {
+    // find script tag with id 'sas_script'. This ensures it only works if you're using Smart Ad Server.
+    const el = document.getElementById(elementId).querySelectorAll("script[id^='sas_script']");
+    if (el[0].nextSibling && el[0].nextSibling.localName === 'iframe') {
+      el[0].nextSibling.style.setProperty('display', 'none');
+    }
+  } catch (e) {
+    // element not found!
+  }
+}
+
 function outstreamRender(bid) {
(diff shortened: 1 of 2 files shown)


FilipStamenkovic (approved):

modules/parrableIdSystem.js
```diff
@@ -183,6 +184,8 @@ function fetchId(configParams) {
   const eid = (parrableId) ? parrableId.eid : null;
   const refererInfo = getRefererInfo();
   const uspString = uspDataHandler.getConsentData();
+  const gdprApplies = (gdprConsentData && typeof gdprConsentData.gdprApplies === 'boolean' && gdprConsentData.gdprApplies);
+  const gdprConsentString = (gdprConsentData && gdprApplies && gdprConsentData.consentString) || '';
```
FilipStamenkovic, line 188: Just a tip for future reference, you could have done this in a shorter way.  
You could have changed:

```
typeof gdprConsentData.gdprApplies === 'boolean' && gdprConsentData.gdprApplies
```

into:

```
gdprConsentData.gdprApplies === true
```

Also, for line 188, you don't need to check for `gdprConsentData`, if that evaluates to `false` then `gdprApplies` would be also `false`. So, for line 188 it's enough just to do:

```
const gdprConsentString = (gdprApplies && gdprConsentData.consentString) || '';
```

PR #6285, opened 2021-02-08
Updated data mapping of auction and winning bid objects in pubxai analytics adapter. 

pnh-pubx:
## Type of change
- [*] Bugfix
- [*] Refactoring (no functional changes, no api changes)

## Description of change
Updated winning bid and auction object fields that are being sent to the server.

### DIFF
--- a/modules/pubxaiAnalyticsAdapter.js
+++ b/modules/pubxaiAnalyticsAdapter.js
@@ -27,9 +27,13 @@ var pubxaiAnalyticsAdapter = Object.assign(adapter(
       if (eventType === CONSTANTS.EVENTS.BID_TIMEOUT) {
         args.forEach(item => { mapBidResponse(item, 'timeout'); });
       } else if (eventType === CONSTANTS.EVENTS.AUCTION_INIT) {
-        events.auctionInit = args;
+        events.floorDetail = {};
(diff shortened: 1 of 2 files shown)


pnh-pubx:
Looks like the circleci build failure is related to something else.

modules/pubxaiAnalyticsAdapter.js
```diff
@@ -134,8 +140,9 @@ pubxaiAnalyticsAdapter.shouldFireEventRequest = function (samplingRate = 1) {
 function send(data, status) {
   if (pubxaiAnalyticsAdapter.shouldFireEventRequest(initOptions.samplingRate)) {
     let location = utils.getWindowLocation();
-    if (typeof data !== 'undefined' && typeof data.auctionInit !== 'undefined') {
```
blingster7, line 137: any reason why you took the `data.auctionInit !== undefined` check away?

blingster7 (approved):
LGTM

modules/pubxaiAnalyticsAdapter.js
pnh-pubx: I have removed auctionInit from the data object completely as we don't need it anymore. That's why this condition isn't required here.

PR #6304, opened 2021-02-12
add warning for non-string adSlots

jsut:
## Type of change

- [x] Other

## Description of change
output a warning if adSlot is not a string. Update markdown to indicate they should be strings.  It might be better to just support non-strings, since the recommended usage of the field is actually a number these days, but that's a bit more work, and would probably want tests.

### DIFF
--- a/modules/pubmaticBidAdapter.md
+++ b/modules/pubmaticBidAdapter.md
@@ -24,9 +24,9 @@ var adUnits = [
     bids: [{
       bidder: 'pubmatic',
       params: {
-        publisherId: '156209',               // required, must be wrapped in quotes
+        publisherId: '156209',               // required, must be a string, not an integer or other js type.
         oustreamAU: 'renderer_test_pubmatic',   // required if mediaTypes-> video-> context is 'outstream' and optional if renderer is defined in adUnits or in mediaType video. This value can be get by BlueBillyWig Team.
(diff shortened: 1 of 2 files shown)


modules/pubmaticBidAdapter.md
```diff
@@ -26,7 +26,7 @@ var adUnits = [
       params: {
         publisherId: '156209',               // required, must be wrapped in quotes
         oustreamAU: 'renderer_test_pubmatic',   // required if mediaTypes-> video-> context is 'outstream' and optional if renderer is defined in adUnits or in mediaType video. This value can be get by BlueBillyWig Team.
-        adSlot: 'pubmatic_test2',            // optional
+        adSlot: 'pubmatic_test2',            // optional, must be wrapped in quotes
```
patmmccann, line 29: must be wrapped in quotes is an odd way of saying must be a string, when i first read it, i thought it meant must have escaped quotes as the first and last character of the string. 

how about: must be a string, not an integer or other js type.

patmmccann (approved):

modules/pubmaticBidAdapter.md
jsut: I copied the text that pubmatic (i assume) used on publisherId.  I'll update both to say String, which i agree makes more sense.

mike-chowla (approved):
LGTM

Fawke (approved):
LGTM

PR #6313, opened 2021-02-16
Inskin adapter: Support bidder aliasing

cciocov:
Support bidder aliasing by not using bid.bidder to retrieve configs (which are static anyway).

## Type of change

## Description of change
We simplified our code and how (static) configuration data is accessed. This in turn enabled our adapter to support aliasing.


musikele (approved):
lgtm

PR #6322, opened 2021-02-17
Fabrick ID System : updates to _setReferrer and appending urls

andersonbd1:
## Type of change
- [X ] Bugfix

## Description of change
The only change is with the url sent to the fabrick endpoint for getting a fabrickId.
 - encode all url params
 - dedupe w/out queryString and keep the longest
 - additionally truncate from % if that ends up being the last (or next to last) char after truncation
 - truncate 1k instead of 200
 - don't send functions along in query (only send strings and numbers)

### DIFF
--- a/modules/fabrickIdSystem.js
+++ b/modules/fabrickIdSystem.js
@@ -81,7 +81,7 @@ export const fabrickIdSubmodule = {
         setReferrer(refs, window.location.href);
 
         for (let value of refs.values()) {
-          url = appendURL(url, 'r', value, configParams);
+          url = _appendUrl(url, 'r', value, configParams);
         }
 
         const resp = function (callback) {
@@ -117,6 +117,9 @@ export const fabrickIdSubmodule = {
     } catch (e) {
       utils.logError(`fabrickIdSystem encountered an error`, e);
     }
+  },
+  appendUrl: (url, paramName, s, configParams) => {
+    return _appendUrl(url, paramName, s, configParams);
   }
 };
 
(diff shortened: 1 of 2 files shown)


Rothalack:
These are the errors
1) should error on json parsing
     Fabrick ID System
     TypeError: Unable to get property 'callback' of undefined or null reference
   at Anonymous function (webpack:///test/spec/modules/fabrickIdSystem_spec.js:34:5 <- test/test_index.js:228371:5)

2) should truncate the params
     Fabrick ID System
     TypeError: Unable to get property 'callback' of undefined or null reference
   at Anonymous function (webpack:///test/spec/modules/fabrickIdSystem_spec.js:62:5 <- test/test_index.js:228397:5)

3) should complete successfully
     Fabrick ID System
     TypeError: Unable to get property 'callback' of undefined or null reference
   at Anonymous function (webpack:///test/spec/modules/fabrickIdSystem_spec.js:91:5 <- test/test_index.js:228424:5)
   
I assume it's because you are no longer setting fabrickIdSubmodule on fabrickIdSystem_spec.js line 13. You'll need to update your tests for this change.

andersonbd1:
Thanks, @Rothalack .  Turns out I'm an idiot.   The root cause of the error was being suppressed, so I blindly tried a bunch of things totally guessing at why ie wasn't passing while the other browsers were.  Finally, I decided to replicate locally.  So I setup a windows machine that I don't normally develop on to try to spew the actual error.  It still wasn't spewing the error and I couldn't figure out why.  Turns out it was because I stubbed utils.logError!  Once I realized this, it wasn't difficult to find the problem was with Map.values() not being supported in ie.   Removing that stub in most of my tests means I would've seen the issue in circleci the first time w/out needing to blindly try things or run locally on my dinosaur windows machine.  I'll be re-submitting shortly and hopefully it'll be fixed (or at least show me why it failed).

Rothalack:
Looks good!

Issue #6289, opened 2021-02-09
Writes in local storage before CMP validation - prebid.cookieTest - criteoID


haroldpussat:
## Type of issue
BUG

## Description
Prebid.js with criteoId module writes in local storage a entry called '**prebid.cookieTest**' with value '1' before CMP validation, and so before consent for write cookie / local storage is available.

## Steps to reproduce
Build a prebid.js with criteoID sub module (tester with prebid.js 4.15)
Setup the criteo module in userSync.userIds.name: "criteo" for pbjs.settings
Navigate on below test page in private navigation.

If I build a prebid.js version without criteoId sub module, it seems the local storage is not written before CMP validation.

## Test page
private navigation with Chrome, before validate the CMP : https://lp.additi.fr/interne/hpu/criteoID-module/index_20200909.html
![image](https://user-images.githubusercontent.com/23049515/107348473-55799680-6ac7-11eb-9531-e238daf2a8b2.png)

### Expected results
Before CMP validation, no writes in local storage or cookie is allowed (legal basis)

### Actual results
prebid.cookieTest is recorded in local storage (see enclosed screenshot)

## Platform details
Tested on Prebid.js 4.15 and 4.24 with criteoID sub module built in.

## Other information
seems linked to #5616

allanjun:
Hi @haroldpussat  thank you for reporting that, Criteo team will have a look!

allanjun:
After some investigation, we concluded that this works as expected. Please notice this is not exclusive to Criteo UserId module. This behaviour was applied on this commit:

https://github.com/prebid/Prebid.js/pull/5018/files#diff-381b65d4141bd216f9021175532d712ce87a9c4a55985c0cf1226f03421b6014R127

The explanation is documented here: https://github.com/prebid/Prebid.js/blob/master/src/storageManager.js#L292 

> This function returns storage functions to access cookies and localstorage. This function will bypass the gdpr enforcement requirement. Prebid as a software needs to use storage in some scenarios and is not a vendor so GDPR enforcement rules does not apply on Prebid.

patmmccann:
@bretg i think this should be documented in multiple places on prebid docs, potentially gdpr module and userid module pages. Anywhere else?

haroldpussat:
Thanks for the feedback @allanjun . But our DPO maintains that in the scope of advertising, even if Prebid is not a vendor, the consent is mandatory before write in local storage.

patmmccann:
[...]

PR #6323, opened 2021-02-17
PBJS Core: clean side-effect when checking that local storage is enabled

TurpIF:
## Type of change

- [x] Bugfix
- [ ] Feature
- [ ] New bidder adapter   
- [ ] Build related changes
- [ ] CI related changes
- [ ] Other

## Description of change

When calling `storageManager.localStorageInEnabled`, the function is doing a side effect by writing into the local storage and verifying that the write was effectively committed, and so indicating that the locale storage is effectively enabled.

Currently, the side effect remains and a `prebid.cookieTest` value remains in user's local storage.

This change aims to clean this side effect after determining if local storage is enabled.

## Other information

Linked to #6289

### DIFF
--- a/src/storageManager.js
+++ b/src/storageManager.js
@@ -1,4 +1,4 @@
-import { hook } from './hook.js';
+import {hook} from './hook.js';
 import * as utils from './utils.js';
 import includes from 'core-js-pure/features/array/includes.js';
 
@@ -110,7 +110,12 @@ export function newStorageManager({gvlid, moduleName, moduleType} = {}) {
         try {
           localStorage.setItem('prebid.cookieTest', '1');
           return localStorage.getItem('prebid.cookieTest') === '1';
-        } catch (error) {}
+        } catch (error) {
+        } finally {
+          try {
+            localStorage.removeItem('prebid.cookieTest');
+          } catch (error) {}
+        }
       }
       return false;
     }
--- a/test/spec/unit/core/storageManager_spec.js
+++ b/test/spec/unit/core/storageManager_spec.js
@@ -46,16 +46,17 @@ describe('storage manager', function() {
 
   describe('localstorage forbidden access in 3rd-party context', function() {
     let errorLogSpy;
-    const originalLocalStorage = { get: () => window.localStorage };
+    let originalLocalStorage;
     const localStorageMock = { get: () => { throw Error } };
 
     beforeEach(function() {
+      originalLocalStorage = window.localStorage;
       Object.defineProperty(window, 'localStorage', localStorageMock);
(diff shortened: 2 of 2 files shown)


blingster7:
This seems awfully specific to a specific partner/user module. Is there a way to make these more partner agnostic?

@Fawke I think a question was raised to store this behavior behind consent or config. thoughts? https://github.com/prebid/Prebid.js/issues/6289

TurpIF:
The issue raised by #6289 does not come directly from the Criteo user module. It's coming from the UserId module which is checking which storages are available in its [`init()`](https://github.com/prebid/Prebid.js/blob/master/modules/userId/index.js#L799) method (you might even notice that it is using a core storage manager, to bypass some checks).

Therefore, all submodules of UserId are affected. For instance, I was able to reproduce the same issue with the [pubProvidedIdSystem](https://github.com/prebid/Prebid.js/blob/master/modules/pubProvidedIdSystem.js) submodule.

Also, I think that #6289 shows two underlying issues:
* Something is written in local storage before CMP validation
* The method `storageManager.localStorageIsEnabled` use a side effect to test a accessibility but never clean the side effect.

For my point of view, this is two different issues. The first one is a legal/privacy issue. It is not in the scope of this PR.
The second one is more a technical issue which is independent from the first one: using a config, a legal argument or whatever to decide if something can be written in the local storage does not change the fact that, IMHO, it is preferable to clean such side effect.

blingster7 (approved):

bretg:
Prebid.org has GVL ID 887. Any cookie writes we do (e.g. testing cookies are on) need to be subject to GDPR rules: purpose and vendor legal basis.

patmmccann:
I thought the gvl was for sharedid.org, this isnt prebid writing the cookie, this is the publisher

TurpIF:
Please remind that the scope of this PR is only to clean a potential side effect.

I agree that there is a privacy question on whether or not we can do such side effect. But I believe that this is out of the scope of this PR and that it belongs to #6289 instead.

blingster7:
@ChrisHuie could you please get another reviewer for this since this is a core change?

nllerandi3lift:
> Please remind that the scope of this PR is only to clean a potential side effect.
> 
> 
> 
> I agree that there is a privacy question on whether or not we can do such side effect. But I believe that this is out of the scope of this PR and that it belongs to #6289 instead.

FWIW, the proposed changes/tests are passing. I think it's just a question of whether this prebid.cookieTest item in local storage needs to stick around beyond the localStorageIsEnabled method. The PR suggests that it doesn't.

Fawke (approved):

PR #6331, opened 2021-02-18
Gulp Build: fix to populate modules list when gulp bundle is executed

pm-harshad-mane:
## Type of change

- [X] Bugfix

## Description of change
Modules list is populated in the banner comment section of generated prebid.js 
Modules list related code was only in `gulp build` job, but when we use https://docs.prebid.org/download.html the `gulp bundle` job is triggered hence the list of modules passed to `bundle` job is not updated in the comment section.
This PR will fix this. Now the list will be populated even with `gulp bundle` job


pm-harshad-mane:
Hello @jsnellbaker I had missed this part in previous PR.
Could you please review it?

jsnellbaker (approved):
I'm assuming this is fine.  I was trying to replicate these commands locally, but I'm not 100% I had the same setup used by the Download server.  When I did generate the 'bundle' file, it appeared to have the list of the modules in the format desired.

smenzer (approved):
LGTM

PR #6332, opened 2021-02-19
Mass Module: add module to support MASS protocol

cciocov:
## Type of change

- [x] Other

## Description of change
This module adds support for MASS enabled advertising delivered through Exchanges that support the MASS protocol.

### DIFF
--- a/integrationExamples/mass/index.html
+++ b/integrationExamples/mass/index.html
@@ -1,0 +1,107 @@
+<html>
+  <head>
+      <script>
+        window.massConfig = {
+          inskin: {
+            default: {
+              plr_PubfileId: '000000/mass'
+            }
+          }
+        };
+
+        var PREBID_TIMEOUT = 3300;
+  
+        var adUnits = [{
+          code: 'div-gpt-ad-1460505748561-0',
+          mediaTypes: {
+            banner: {
+              sizes: [[300, 250], [300, 600]],
+            }
+          },
+          bids: [{
+            bidder: 'ix',
+            params: {
+              siteId: '123456',
+              size: [300, 250]
+            }
+          }]
+        }];
+  
+        var pbjs = pbjs || {};
+        pbjs.que = pbjs.que || [];
+      </script>
+
+      <script>
+        pbjs.que.push(function() {
+          pbjs.setConfig({
+            mass: {
+              enabled: true
+            }
+          });
+        });
+      </script>
+  
+      <script type="text/javascript" src="/build/dist/prebid.js" async></script>
+      <script>
+        var googletag = googletag || {};
+        googletag.cmd = googletag.cmd || [];
+        googletag.cmd.push(function() {
+          googletag.pubads().disableInitialLoad();
+        });
+  
+        pbjs.que.push(function() {
+          pbjs.addAdUnits(adUnits);
+          pbjs.requestBids({
+            bidsBackHandler: sendAdserverRequest
+          });
+        });
+  
+        function sendAdserverRequest() {
+          if (pbjs.adserverRequestSent) return;
+          pbjs.adserverRequestSent = true;
+          googletag.cmd.push(function() {
+            pbjs.que.push(function() {
+              pbjs.setTargetingForGPTAsync();
+              googletag.pubads().refresh();
+            });
+          });
+        }
+  
+        setTimeout(function() {
+          sendAdserverRequest();
+        }, PREBID_TIMEOUT);
+      </script>
+  
+      <script>
+        (function () {
+          var gads = document.createElement('script');
+          gads.async = true;
+          gads.type = 'text/javascript';
+          var useSSL = 'https:' == document.location.protocol;
+          gads.src = (useSSL ? 'https:' : 'http:') +
+            '//www.googletagservices.com/tag/js/gpt.js';
+          var node = document.getElementsByTagName('script')[0];
+          node.parentNode.insertBefore(gads, node);
+        })();
+      </script>
+  
+      <script>
+        googletag.cmd.push(function () {
+          googletag.defineSlot('/19968336/header-bid-tag-0', [[300, 250], [300, 600]], 'div-gpt-ad-1460505748561-0').addService(googletag.pubads());
+  
+          googletag.pubads().enableSingleRequest();
+          googletag.enableServices();
+        });
+      </script>
+  </head>
+  
+  <body style="margin: 0;">
+    <div id="content" style="width: 920px; padding: 20px; height: 2500px; margin: 0 auto; background-color: #222;">
+      <div id="div-gpt-ad-1460505748561-0">
+        <script type="text/javascript">
+          googletag.cmd.push(function() { googletag.display('div-gpt-ad-1460505748561-0'); });
+        </script>
+      </div>
+    </div>
+  </body>
+</html>
--- a/integrationExamples/mass/index.html
+++ b/integrationExamples/mass/index.html
@@ -36,7 +36,7 @@
           pbjs.setConfig({
             mass: {
               enabled: true,
-              bootloaderUrl: 'https://cdn.massplatform.net/bootloader.js',
+              renderUrl: 'https://cdn.massplatform.net/bootloader.js',
               dealIdPattern: /^MASS/i
             }
           });
--- a/integrationExamples/mass/index.html
+++ b/integrationExamples/mass/index.html
@@ -41,7 +41,7 @@
         });
       </script>
   
-      <script type="text/javascript" src="/build/dist/prebid.js" async></script>
+      <script type="text/javascript" src="/build/dev/prebid.js" async></script>
       <script>
         var googletag = googletag || {};
         googletag.cmd = googletag.cmd || [];
@@ -96,7 +96,8 @@
   </head>
   
   <body style="margin: 0;">
-    <div id="content" style="width: 920px; padding: 20px; height: 2500px; margin: 0 auto; background-color: #222;">
+    <div id="content" style="width: 920px; padding: 20px; height: 2500px; margin: 0 auto; background-color: #222; color: #fff;">
+      <p>Note: for this example to work, you need access to a bid simulation tool from your MASS enabled Exchange partner.</p>
       <div id="div-gpt-ad-1460505748561-0">
         <script type="text/javascript">
           googletag.cmd.push(function() { googletag.display('div-gpt-ad-1460505748561-0'); });
--- a/integrationExamples/mass/index.html
+++ b/integrationExamples/mass/index.html
@@ -35,7 +35,9 @@
         pbjs.que.push(function() {
           pbjs.setConfig({
             mass: {
-              enabled: true
+              enabled: true,
+              bootloaderUrl: 'https://cdn.massplatform.net/bootloader.js',
+              dealIdPattern: /^MASS/i
             }
           });
         });
@@ -79,7 +81,7 @@
           gads.type = 'text/javascript';
           var useSSL = 'https:' == document.location.protocol;
           gads.src = (useSSL ? 'https:' : 'http:') +
-            '//www.googletagservices.com/tag/js/gpt.js';
+            '//securepubads.g.doubleclick.net/tag/js/gpt.js';
           var node = document.getElementsByTagName('script')[0];
           node.parentNode.insertBefore(gads, node);
         })();
--- a/modules/mass.js
+++ b/modules/mass.js
@@ -1,0 +1,114 @@
+/**
+ * This module adds MASS support to Prebid.js.
+ */
+
+import { config } from '../src/config';
+import { getHook } from '../src/hook';
+import { logInfo } from '../src/utils';
+import find from 'core-js-pure/features/array/find.js';
+
+let listenerAdded = false;
+const massBids = {};
+
+//config.getConfig('mass', config => init(config.mass));
+init();
+
+/**
+ * Module init.
+ */
+function init(cfg = {}) {
+  /*if (cfg.enabled === false) {
+    logInfo('MASS module is disabled');
+    return;
+  }*/
+
+  getHook('addBidResponse').before(addBidResponseHook);
+}
+
+/**
+ * Before hook for 'addBidResponse'.
+ */
+function addBidResponseHook(next, adUnitCode, bid) {
+  if (!isMassBid(bid)) {
+    return next(adUnitCode, bid);
+  }
+
+  const bidRequest = find(this.bidderRequest.bids, bidRequest =>
+    bidRequest.bidId === bid.requestId
+  );
+
+  massBids[bid.requestId] = {
+    bidRequest,
+    bidResponse: bid,
+    adm: bid.ad
+  };
+
+  bid.ad = '<script>window.top.postMessage({massBidId: "' + bid.requestId + '"}, "*");\x3c/script>';
+
+  addListenerOnce();
+
+  next(adUnitCode, bid);
+}
+
+/**
+ * Check if a bid is MASS.
+ */
+function isMassBid(bid) {
+  if (!(/^MASS/i.test(bid.dealId))) {
+    return false;
+  }
+
+  return /mass:\/\//i.test(bid.ad);
+}
+
+/**
+ * Add listener to detect requests to render MASS ads.
+ */
+function addListenerOnce() {
+  if (!listenerAdded) {
+    window.addEventListener('message', e => {
+      if (e && e.data && e.data.massBidId) {
+        render(getRenderPayload(e));
+      }
+    });
+    
+    listenerAdded = true;
+  }
+}
+
+/**
+ * Prepare payload for render.
+ */
+function getRenderPayload(e) {
+  const payload = {
+    type: 'prebid',
+    e
+  };
+
+  Object.assign(payload, massBids[e.data.massBidId]);
+  delete payload.bidResponse.ad;
+
+  return payload;
+}
+
+/**
+ * Render a MASS ad.
+ */
+function render(payload) {
+  const ns = window.MASS = window.MASS || {};
+
+  ns.bootloader = ns.bootloader || {queue: []};
+  ns.bootloader.queue.push(payload);
+
+  if (!ns.bootloader.loaded) {
+    const s = document.createElement('script');
+    s.type = 'text/javascript';
+    s.async = true;
+    s.src = 'https://cdn.massplatform.net/v1/bootloader.js';
+
+    const x = document.getElementsByTagName('script')[0];
+    x.parentNode.insertBefore(s, x);
+
+    ns.bootloader.loaded = true;
+  }
+}
--- a/modules/mass.js
+++ b/modules/mass.js
@@ -6,8 +6,8 @@ import { config } from '../src/config.js';
 import { getHook } from '../src/hook.js';
 import find from 'core-js-pure/features/array/find.js';
 
-let listenerAdded = false;
-let massEnabled = false;
+export let listenerAdded = false;
+export let massEnabled = false;
 
 const defaultCfg = {
   bootloaderUrl: 'https://cdn.massplatform.net/bootloader.js',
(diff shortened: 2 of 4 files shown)


patmmccann:
why wouldn't a publisher just define a banner renderer js on the ad unit that uses https://cdn.massplatform.net/bootloader/v1/bootloader.js'

massadmin:
> why wouldn't a publisher just define a banner renderer js on the ad unit that uses https://cdn.massplatform.net/bootloader/v1/bootloader.js'

The system is primarily designed for Exchanges that wish to offer complex private marketplace solutions to their clients when trading exotic formats at scale. The native-renderer and native placeholders in combination with mediaTypes were evaluated and fell short of the requirements set by publishers, Exchanges and rich-media vendors.

I am unfortunately not able to go into the details of some of the requirements as it would necessitate sharing third party confidential information.

patmmccann:
Other vendors are trading skins without listening to every unit on the page,  why not just hook to one ad unit

massadmin:
> Other vendors are trading skins without listening to every unit on the page, why not just hook to one ad unit

This is because this solution is not designed for a skins provider but for an Exchange. It just so happens that one of the first companies to go live with it on the rich media side happens to trade skin formats. The solution is designed for any type of experience.

patmmccann:
But your script only calls https://cdn.inskinad.com/isfe/tags/ipt.js . You wouldn't be able to change that without changing the bootloader script which needs to be locked to a version. Daisy chaining to something that isn't locked is not an okay work around to the version lock requirement, so this will need to be locked as well

patmmccann:
Also your test for deal id starts with MASS is problematic, you should check for something less likely to conflict.

massadmin:
We were not aware of a version lock requirement for external resources. Let me circle back with the other third parties that approve any changes to the bootloader. The existing bootloader (which is distributed in nature) is not the final implementation.

Regarding the MASS deal ID identifier, please note that participating Exchanges and Publishers that wish to use this module would be checking that no non-related ID's commence with this string. We had considered to use a special character in the string but were advised against it by one the of Exchange that preferred to use the ID 'MASS' instead (The Exchange in fact could not guarantee that special characters would not cause issues with certain DSP's and their own internal systems). 

Given the participation of the Exchange to monitor this prefix and the active participation of a Publisher wishing to use the module, the risk of failure is minuscule. Furthermore you will find that on line 68 an additional check is performed to verify the presence of the protocol URI inside the ADM as an additional check in case a publisher or buyer made a trafficking mistake.

We are more than happy to change the prefix identifier to something more obscure as long as it does not contain special characters for the above mentioned reasons.

BTW, thank you for reviewing this so quickly!

patmmccann:
See this document

https://docs.prebid.org/dev-docs/module-rules.html

massadmin:
Thank you for the meta.mass DealID suggestion. Let me circle back with the third-party on the Exchange side to check that the dot doesn't create any issues.

I will also get back to you with a solution to ensure that the module adheres to the rules you linked.

massadmin:
We have made the following changes:

1) The DealID prefix has been changed to meta_mass based on your recommendation and feedback from the Exchange partner
2) An option has been added for a publisher to override the version of the bootloader or to load a custom bootloader
3) The bootloader itself has been updated to a production version and is not hardcoded to a specific provider anymore
4) A disclaimer has been added to the MD file together with instruction on how to add configuration options (including disabling the module and override)

smenzer:
Should there be a docs update for this? At the very least, I think that the fact that external JS is being used needs to be declared some place for publishers to see; but more information about what the module is and how to use it seems warranted.

modules/mass.js
```diff
@@ -0,0 +1,127 @@
...
+  addListenerOnce();
+
+  next(adUnitCode, bid);
+}
+
+/**
+ * Check if a bid is MASS.
+ */
+function isMassBid(bid) {
+  // the deal ID must start with MASS:
```
smenzer, line 68: should be `META_MASS` now, right?

```suggestion
  // the deal ID must start with META_MASS:
```

smenzer:
@patmmccann or @musikele any other concerns here or has @massadmin addressed them for you?

massadmin:
Hi Scott,

Thank you very much for looking into this so quickly. We will make the
amendment regarding the source code comment.

We have added to the MD file a disclaimer regarding the external loading of
JS files. We would currently like to communicate as little as possible
about the exact mechanics behind the protocol as it is invite only and we
are under NDA.

Should we maybe add a disclaimer that states that this module should only
be added if you have been invited by an Exchange and Mass as it otherwise
doesn't do anything?

Kind regards,
Reda Borchardt

massadmin:
I committed a new MD file that goes into more detail about the way the module functions.

smenzer:
@patmmccann / @musikele just a friendly ping again to see if there's anything else you're looking for (answers, code, etc) for this PR.

massadmin:
We have made some changes to give publishers more freedom in how to identify MASS bids, and the module is now checking for a 'mass' flag in bid meta or testing the deal IDs against a publisher defined pattern. A further test is performed to ensure a MASS URI is present in the ad markup.

FilipStamenkovic:
Can you please add some integration example of this module?  
It would be nice to have some .html example page where we could see how this module works.  
You could place it in `integrationExamples`.

massadmin:
Thank you for this. We will add an integration example as soon as possible.

modules/mass.js
```diff
@@ -0,0 +1,130 @@
...
+  ns.bootloader.queue.push(payload);
+
+  if (!ns.bootloader.loaded) {
+    const s = document.createElement('script');
+    s.type = 'text/javascript';
+    s.async = true;
+    s.src = cfg.bootloaderUrl;
+
+    const x = document.getElementsByTagName('script')[0];
+    x.parentNode.insertBefore(s, x);
```
FilipStamenkovic, line 126: You should load external script with function `loadExternalScript` from `src/adloader.js`.  
Also, please take a look at this policy for external scripts: https://github.com/prebid/prebid-js-external-js-template#policy.

FilipStamenkovic (changes requested):
You should add more tests to this PR.  
At the moment your test coverage is 56.52%.  
It should be minimum 80%.  

Also, please take a look at my previous comments for external scripts and example page.

FilipStamenkovic:
Can you pull/rebase changes from origin master, at the moment this PR is ~230 commits behind master branch?

modules/mass.js
```diff
@@ -0,0 +1,130 @@
...
+    bidRequest.bidId === bid.requestId
+  );
+
+  massBids[bid.requestId] = {
+    bidRequest,
+    bid,
+    adm: bid.ad
+  };
+
+  bid.ad = '<script>window.top.postMessage({massBidId: "' + bid.requestId + '"}, "*");\x3c/script>';
```
FilipStamenkovic, line 60: Maybe here you need `window.parent` and not `window.top`??  
Will this work if pbjs is running inside iframe and not on `window.top`?

patmmccann:
A .md commit is not the same as a docs repo commit, please link to a docs repo commit as well. 

I'm noticing this is still not locked to a version nor open source - https://cdn.massplatform.net/bootloader.js

I would suggest adding a link to the unminified github code in the comments of the script https://cdn.massplatform.net/bootloader.js to make it easy to inspect. 

changes to use a new version of this code will require a prebid pr

modules/mass.js
cciocov: yes, window.parent makes more sense, updated, thanks!

modules/mass.js
cciocov: loadExternalScript requires the module to be whitelisted, but, I think the MASS module does not fall under the category of modules that load external content, at least not in the way it is assumed for the current whitelisted modules.

All currently whitelisted modules use the external code they load to implement Prebid specific logic, that is, the external code that is loaded affects how the module behaves. This is not the case for MASS. We only load external code when the bid is won, and that code is not Prebid-specific... it deals with rendering the ad, something that all other Prebid modules / bid adapters do via the "ad markup" field. The external code we load does not change any functionality implemented in mass.js.

Makes sense?

cciocov:
Hi @patmmccann ,

Thanks for all the feedback so far!

Regarding the issue with loading external scripts, I think there might be some confusion around this. At some point, all Prebid modules will load some external code. This happens once the bid is won and the "ad markup" is being executed, at which point external code is being loaded that renders the ad in any way deemed fit by the ad provider. This code/logic has nothing to do with Prebid and does not influence Prebid's logic and operation on page.

The exact same thing is happening with this module. Once a MASS bid is won, and only then, this module will load some external code that basically renders the ad in that bid, in a way that is needed by rich media ad formats. The code has nothing to do with Prebid, and does not influence the logic of the module itself, hence, does not influence Prebid's logic.

I hope this makes sense, but please let me know if you have any additional questions or concerns, thank you!

cciocov:
@patmmccann regarding the docs, I'll submit a PR on the docs repo later today, thanks!

patmmccann:
Thank you very much. One note, your case to load unlocked external js seems to focus on the use case for the module being only rendering. If this is true,  what about current rendering workflow is deficient and requires a module?

patmmccann:
@cciocov to put my questions in context, if prebid doesn't support non-outstream rendering very well, that is something we want to know and fix rather than have a variety of rendering module commits. It seems your team has done some research into a deficiency in the workflow and for some reason isnt sharing what that issue is.

gglas:
@cciocov we'd like to discuss this at our next Prebid.JS community meeting - are you working with any Prebid members who are interested in leveraging this that could speak to it on your behalf?

cciocov:
Hey @patmmccann @gglas 

I've just submitted a PR for updating the docs repo:
https://github.com/prebid/prebid.github.io/pull/2774

With regards to your other questions, my colleague, @massadmin , will get back to you soon.

Thanks for all the feedback so far!

massadmin:
Dear @patmmccann and @gglas,

First off, let me thank you for all of your time, patience and assistance. This module relates to a joint effort that involves Index Exchange, DSP's, AdTech vendors and publishers.

Throughout the development of the eco-system with third parties, we have closely followed the discussions on Github around the custom rendering work you have done. It is in fact our hope that at one point elements of the MASS protocol will become part of Prebid's core feature set.

The need for a module over the existing methods of defining renderers are operational in nature. The need primarily comes from the complexity of requirements linked to private media exchanges when operating at scale with exotic formats and custom experiences. Furthermore, and without going into too much detail, the module approach allows for services to emerge that provide real-time debugging and alerts, transparency, verification, testing, billing and auditing that can only be achieved if the incoming bids are scanned prior to reaching the ad server.

In short, MASS is designed to allow a distributed network of incentivised participants to ensure that complex exotic setups deliver with zero defects.

A series of tests have been performed last year using MASS and a custom version of Prebid and the IX adapter that had the support for the protocol built-in. However, all parties involved agreed that it would make a lot more sense to release this as a globally available module for everybody prior to launching the service.

There is a bit of a chicken and egg scenario as all of the benefits aren't apparent for you at this stage since the eco-system is still being built and designed for private marketplaces only (for now). This is one of the reasons why it currently invite only.

Index Exchange is the best member to represent MASS as they have a leading position in this initiative. (I believe that discussions might have already taken place)

I hope that this is helpful information.

patmmccann:
That is really helpful; one situation we are hoping to avoid is dozens of rich media rendering modules being required by publishers for all their various demand. Would you be open to modifying this so that the rendering endpoint url was configurable as well and that there could be multiple patterns that matched different endpoints? Eg if deal like mass, call the https://cdn.massplatform.net/bootloader.js; if it is like something else, call some other rendering endpoint... Then it could be called the deal rendering module or something as such and quickly achieve the goal you are stating of making this type of functionality more widely applicable.

massadmin:
Your idea is very good. You will find that in the configuration described in the documentation we already have to a certain extent an element of what you describe:

pbjs.que.push(function() {
  pbjs.setConfig({
    mass: {
      enabled: true,
      bootloaderUrl: 'https://cdn.massplatform.net/bootloader.js',
      dealIdPattern: /^MASS/i
    }
  });
});

The publisher has the ability here to set the bootloader they wish to use and the DealID using a regex object.

Furthermore, your stated goal of having less modules to implement or Javascript to add on a page is a the core of the MASS propositions to publishers (together with a reduction of ad calls required). This is done through a URI between the partners that not only specifies what to render but how to render it. It's infinitely scalable in terms of number of 'renderers' and only pulls the code necessary to render the experience on the fly and through configuration management.

There is also an element of enforcement (think ADS.TXT) to ensure that the exact behaviour that was promised to a client is in fact delivered, and to also allow to model complex rules through configuration management rather than the publisher needing to implement code.

I understand your wish for having as generic functionality as possible and we are happy to support that.

There are two ways I think we could go about it:

**1)** We include into the MASS module a 'use at your own risk' mode that comes without training wheels and does exactly what you propose with the ability to mix and match a publisher DealID's and renderers at your own discretion. The logic to implement this is pretty straightforward and we would be more than happy to accommodate and maintain this.

We would however need to make sure that it is clear in the documentation that this is a 'use at your own risk' feature that is not supported by MASS and its partners.

**2)** We could have a generic Deal based rendering module next to the official MASS module. Again, we would be happy to contribute to the community by submitting it and iterating with you to support all the work you have personally done around outstream advertising within PrebidJS.

It would be OK for the two modules to co-exist because what MASS does belongs in a (yet undefined) new category of modules for distributed systems. It just so happens that one of the first thing it does on the surface is to facilitate the rendering of rich media. The need for an official MASS module will continue to exist because of the guarantees a full chain of custody with approved partners can provide.

The advantage of the second approach is that publishers that want to do it themselves would not need to carry future MASS functionality in their code that is not relevant to them.

I am open to suggestions including an Option 1 that could lead to an Option 2 as we see uptake from publishers. We could also have Option 1 carry the generic functionality but give the option of having a stripped down generic version so that both aren't needed. If a user attempts to build using both modules a warning could occur to highlight that there is a redundancy.

integrationExamples/mass/index.html
```diff
@@ -0,0 +1,107 @@
...
+        pbjs.que.push(function() {
+          pbjs.setConfig({
+            mass: {
+              enabled: true
+            }
+          });
+        });
+      </script>
+  
+      <script type="text/javascript" src="/build/dist/prebid.js" async></script>
```
FilipStamenkovic, line 44: Is this working on your local after you execute `gulp serve`?  
I had to change `src` to `../../build/dev/prebid.js` on my local to load pbjs properly.

integrationExamples/mass/index.html
```diff
@@ -0,0 +1,107 @@
...
+              sizes: [[300, 250], [300, 600]],
+            }
+          },
+          bids: [{
+            bidder: 'ix',
+            params: {
+              siteId: '123456',
+              size: [300, 250]
+            }
+          }]
```
FilipStamenkovic, line 27: Are you getting any bids with these parameters?  
On my local I cannot get any bid from `ix`.  
I tried changing my location with VPN, still no luck.

massadmin: I just checked with Index Exchange and publishers are supposed to use a tool provided as part of a toolkit that is proprietary to IX. It allows the publisher to check their MASS integration. (Publishers are not expected to entirely self integrate)

If you would like to test it yourself, I can arrange for you to receive the bid simulation tool which is not intended for a general audience.

*We should probably add a comment to this regard in the integration file.

integrationExamples/mass/index.html
cciocov: @FilipStamenkovic Thanks for this, it's now fixed. I also added a notice to let people know that in order for the page to actually work, you need access to a bid simulation tool.

patmmccann:
Curious if you have examined https://github.com/AntoineJac/prebidSkin as an example of successful skin rendering?

massadmin:
We will submit the changes to the integration you requested. 

Thanks for the link you posted, it's something that I will pass on for those that deal with advertising formats to look at. As far as MASS is concerned, we don't deal with formats at all. We solely focus on the tech to facilitate their delivery, as exotic as they might be,

patmmccann:
Thanks for the quick changes! @cciocov this came up for discussion at the prebid js meeting today, and the consensus is that your proposal option 1 detailed at https://github.com/prebid/Prebid.js/pull/6332#issuecomment-799750464 would make this PR a very welcome addition to the community software and we appreciate your contribution. Mass/Inskin are also welcome to apply for community membership. Also, if you are not locking bootloader.js to a version, a publisher being required to set it in the module config instead of it being the default loaded js is important. I know the pace of review on this has been a little frustrating, but it can take time to build consensus on controversial changes and the js committee only meets every other week.

integrationExamples/mass/index.html
```diff
@@ -0,0 +1,110 @@
...
+        var pbjs = pbjs || {};
+        pbjs.que = pbjs.que || [];
+      </script>
+
+      <script>
+        pbjs.que.push(function() {
+          pbjs.setConfig({
+            mass: {
+              enabled: true,
+              bootloaderUrl: 'https://cdn.massplatform.net/bootloader.js',
```
patmmccann, line 39: it was requested at js meeting today this changes to renderUrl to more closely align with other naming conventions. Also if this isn't locked to a version, it should be in the module config, not a default. All examples and documentation are welcome to point to it, but the publisher being required to explicitly name it in their config makes the committee more comfortable with it.

patmmccann (changes requested):

modules/mass.js
patmmccann: see https://github.com/prebid/Prebid.js/pull/6332/files#r596337899

massadmin:
Hi @patmmccann, thanks for the feedback and for pointing towards a clear option. We will come back to you with code changes once we solutioned the most elegant way of implementing the pattern matching in a generalised way. We will move as quickly as possible as this is of the highest priority for us and some of our partners.

cciocov:
Hey @patmmccann ! I've renamed the param to "renderUrl" and it now has no default value, so publishers will need to specify it in config. We are already thinking of additional changes to make the module more generic and will open a new PR once these are fleshed out, but in the interest of time and because we need to start testing this with a few partners, do you think the current module is good enough for being merged? Thanks!

patmmccann:
This is excellent,  could you update your docs repo pull to match some of the evolution here?

To be clear, the subsequent commit will give an example of and include methods to do multiple deal match string/script pairs in the same config, eg mass match call bootloader.js, myadvertiser match, call myotherscript.js?

patmmccann:
@FilipStamenkovic retagging you to take a look after several changes above

cciocov:
@patmmccann 

Not sure what you mean with regards to the docs repo?

About the subsequent commit, yes, that's exactly what we're planning, we just need to think this through a bit and see what additional features we need to include. I'll open a new PR/discussion thread on this by end of this month.

massadmin:
@cciocov: I just updated the docs to add a Generic Mode section.
@patmmccann: We already started the design of the next iteration that cciocov mentioned. We are absolutely committed to having the feature requested in there and more for your consideration.

modules/mass.md
```diff
@@ -0,0 +1,61 @@
...
+
+This module only loads external JavaScript resources if the publisher ad server has 
+selected a MASS enabled bid as a winner. 
+
+Find out more [here](https://massplatform.net).
+
+## Disclosure
+
+- This module loads external JavaScript to render creatives
+- This module should only be used by publishers that have been invited to the MASS network
```
patmmccann, line 25: Can you remove this sentence and add the generic mode desription here as well?

patmmccann:
@massadmin could we clean up the docs commit just a little more? eg this sentence is no longer true 'This module should only be used by publishers that have been invited to the MASS network' as a publisher may now use this module to attach any script to any deal id pattern. 

also I am a tiny bit confused on something, your _spec tests file has `expect(window.mass.bootloader.loaded).to.equal(true);` but I don't see the spec file configuring the library. How is this passing?

massadmin:
@patmmccann I just updated the doc file and the md

cciocov:
@patmmccann The test calls the render() function, which sets that on the page. Normally, render will not be called unless 1) MASS is enabled and properly configured and 2) an appropriate bid is won, however, this test only checks render() itself.

This function (and test) will get updated in the coming PR.

Thanks.

patmmccann:
lgtm

patmmccann (approved):

Issue #3823, opened 2019-05-14
Request for Bid "Streaming"


woggle23:
## Type of issue
Feature

## Description
At present Prebid doesn't provide a means to obtain the _current_ highest bidder at anytime during an auction, it's only available once the auction has completed. However, in some situations we don't control the auction timeout, as such, (previously) we weren't able to return any bids to the client. To resolve this, we updated our approach to listen to each bid (via `onBidResponse`), and forward the higher bids to our client, that way whenever their timeout occurs they have the highest bid response at that moment.

This works well enough for the simple case of sorting bids by cpm. However, we'd like to use Prebid's `bid caching`. The simple solution for us would be to call `getAdserverTargeting` when each bid response is received. The problem here is the same as the above, the adserverTargeting is only available once the auction has completed. Furthermore, the new bid isn't added to `_bidsReceived` until after the `onBidResponse` event has been fired.

Happy to implement a solution, interested in feedback on the idea.

bretg:
@mkendall07 - is this true?
> adserverTargeting is only available once the auction has completed

I thought that targeting was available at any time. i.e. if the page wanted to close the auction early, it could call `setTargetingForGPTAsync` and get whatever bid were back.

bretg:
Found this code snippet that makes me agree this does seem to be an issue:
```
  auctionManager.getBidsReceived = function() {
    return _auctions.map((auction) => {
      if (auction.getAuctionStatus() === AUCTION_COMPLETED) {
        return auction.getBidsReceived();
      }
    }).reduce(flatten, [])
      .filter(bid => bid);
  };
```
So I agree there should be an option to take whatever bids are back when the page calls `setTargeting`.

gglas:
@woggle23 @patmmccann and the community discussed today, would you be interested in reengaging?

patmmccann:
hi @woggle23 

Getting current highest bid appears to be a very useful feature. One use case is setting a threshold to call the ad server or a render function before the timeout. Another use case is quickly ending auctions with current status on lazy units that are about to come into the viewpoint or units on an exit intent pop up. We'd love your contribution, and are curious if you're still willing to commit?

woggle23:
Hi @gglas / @patmmccann 

[...]

PR #6342, opened 2021-02-22
Pbjs Core: add new API to return the highest unused bid for a given adunit

woggle23:
## Type of change
- [ ] Feature

## Description of change
This introduces a new API to return the highest unused bid for a given adUnit code, regardless of whether an auction is still ongoing.

This is the approach I've taken in my fork of Prebid.js to address [this issue](https://github.com/prebid/Prebid.js/issues/3823). I appreciate this may not be an ideal approach, so any suggestions would be gratefully received.


blingster7 (approved):
LGTM, is there any way you can add a test somewhere?

woggle23:
Sure, unit tests added.

ChrisHuie (approved):

bretg:
I'm a little bummed this got merged without documentation.

I went to try to fix that gap myself and have a question and comment:

Q) I don't understand the difference between this function pbjs.getHighestCpmBids(adUnitCode)?
C) I don't particularly like the overly long name of this function. Bloat. It's too late now, but Chris, in the future, please help police shorter names.

woggle23:
@bretg 
>Q1) I don't understand the difference between this function and pbjs.getHighestCpmBids(adUnitCode)?
`pbjs.getHighestCpmBids` only considers bids for auctions that have completed (and are also `unused`), the new API considers bids for ongoing auctions as well.

>Q2) What does "unused" mean? Does it mean that targeting was never set, it was never rendered, or something else?
see [here](https://github.com/prebid/Prebid.js/blob/e2062447abd900e1b7ad028eaa3bae6913c57ca9/src/prebid.js#L221).

>C) I don't particularly like the overly long name of this function. 40 characters. Wow. We're constantly under fire for code size and this won't minify. It's too late now, but Chris, in the future, please help police shorter names.
Noted.

bretg:
Parsed the code myself to figure out what "unused" meant. If I got it wrong, please correct.

https://docs.prebid.org/dev-docs/publisher-api-reference.html#module_pbjs.getHighestUnused

PR #6344, opened 2021-02-22
Rubicon Analytics: Add option to delay sending of event message

robertrmartinez:
## Type of change
- [X] Bugfix
- [X] Other

#### Auction Event Delay
Depending on the many differences and factors to each publisher, the amount of time we delay the sending of the analytics events can make a difference. 

By default, we will send the auction level data immediately.

However, there may be cases where it is a good idea to wait and give the prebid bidWon events time to come in, as to not make too many requests over the browser.

A pub can change this delay using:

```js
pbjs.setConfig({
   rubicon: {
      analyticsEventDelay: 2000 // some number in ms
   }
});

```

Then, once all gam slots are marked as "rendered" the auction message will delay for the time set.

#### BUGFIX: GPT not defined before pbjs

When gpt is not defined before pbjs, we need to directly attach the googletag object onto the window.

### DIFF
--- a/modules/rubiconAnalyticsAdapter.js
+++ b/modules/rubiconAnalyticsAdapter.js
@@ -60,7 +60,8 @@ const cache = {
 const BID_REJECTED_IPF = 'rejected-ipf';
 
 export let rubiConf = {
-  pvid: utils.generateUUID().slice(0, 8)
+  pvid: utils.generateUUID().slice(0, 8),
+  analyticsEventDelay: 0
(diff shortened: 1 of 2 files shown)


harpere (approved):

ChrisHuie (approved):

patmmccann:
@robertrmartinez @bszekely1 whats the difference between sendTimeout and analyticsEventDelay

PR #6359, opened 2021-02-25
PBJS Core: fix bidderRequest matching for finding a renderer

JulieLorin:
## Type of change
Bugfix

## Description of change
To find the renderer for a bid, Prebid looks for a renderer in the adUnit (media type level or adUnit level). To find this adUnit, Prebid will look at all Bid requests sent to the bidder and find the first one matching the adUnitCode.  
Sometimes, there are bids with different media types sent independently to the bidder (some adapters doesn't support multi media types very well) and in this case, this won't necessary be the correct bid request that will be returned. This PR match the bid requests using the requestId/bidId which is safer. 

This PR has been tested on a local prebid integration : Before, the renderer attached to the bid was the one provided by the bid adapter. After the fix, it was the renderer attached to the bid request adunit.

### DIFF
--- a/src/auction.js
+++ b/src/auction.js
@@ -519,7 +519,7 @@ function getPreparedBidForAuction({adUnitCode, bid, bidderRequest, auctionId}) {
   events.emit(CONSTANTS.EVENTS.BID_ADJUSTMENT, bidObject);
 
   // a publisher-defined renderer can be used to render bids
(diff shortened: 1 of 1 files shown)


osazos:
LGTM,
By the way it would be nice to add some tests or provide an example to test what you describe.

osazos (approved):

FilipStamenkovic (approved):
LGTM

PR #6360, opened 2021-02-25
Revert "Native: add custom data assets capability"

mmoschovas:
Reverts prebid/Prebid.js#6220


bretg:
FYI - this revert was an accident and not actually carried through.#6220 is still live!

PR #6364, opened 2021-02-26
Sonobi Bid Adapter: send eids in bid request.

JonGoSonobi:
## Type of change
- [x] Bugfix

## Description of change
Sending list of eids to sonobi bid request endpoint.
Unwrapping id5id.

- apex.prebid@sonnobi.com

### DIFF
--- a/modules/sonobiBidAdapter.js
+++ b/modules/sonobiBidAdapter.js
@@ -116,18 +116,20 @@ export const spec = {
       payload.schain = JSON.stringify(validBidRequests[0].schain)
     }
     if (deepAccess(validBidRequests[0], 'userId') && Object.keys(validBidRequests[0].userId).length > 0) {
+
       const userIds = validBidRequests[0].userId;
-      if (userIds.id5id) {
+      if(userIds.id5id) {
         userIds.id5id = deepAccess(userIds, 'id5id.uid');
       }
 
-      payload.userid = JSON.stringify(validBidRequests[0].userId);
+      payload.userid = JSON.stringify(userIds);
     }
 
     const eids = deepAccess(validBidRequests[0], 'userIdAsEids');
-    if (Array.isArray(eids) && eids.length > 0) {
+    if(Array.isArray(eids) && eids.length > 0) {
       payload.eids = JSON.stringify(eids);
     }
+
 
     let keywords = validBidRequests[0].params.keywords; // a CSV of keywords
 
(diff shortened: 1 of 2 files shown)


test/spec/modules/sonobiBidAdapter_spec.js
```diff
@@ -414,15 +414,92 @@ describe('SonobiBidAdapter', function () {
       expect(JSON.parse(bidRequests.data.schain)).to.deep.equal(bidRequest[0].schain)
     });
 
+    it('should return a properly formatte drequest with eids as a JSON-encoded set of eids', function () {
```
smenzer, line 417: spelling mistake

```suggestion
    it('should return a properly formatted request with eids as a JSON-encoded set of eids', function () {
```

modules/sonobiBidAdapter.js
```diff
@@ -117,9 +116,21 @@ export const spec = {
       payload.schain = JSON.stringify(validBidRequests[0].schain)
     }
     if (deepAccess(validBidRequests[0], 'userId') && Object.keys(validBidRequests[0].userId).length > 0) {
-      payload.userid = JSON.stringify(validBidRequests[0].userId);
+
+      const userIds = validBidRequests[0].userId;
+      if(userIds.id5id) {
+        userIds.id5id = deepAccess(userIds, 'id5id.uid');
```
aleksatr, line 122: In general, this is not a good practice, you're modifying the source object here, but in this case, it should be safe as I tested it (and checked the core code) and it appears you're getting an adapter-specific copy of validBidRequests.

aleksatr:
@JonGoSonobi you have lint errors, check circle ci output.

modules/sonobiBidAdapter.js
JonGoSonobi: Good catch. I copied the userId object so it doesnt get altered when updating id5id here.

JonGoSonobi:
@aleksatr lint errors were fixed bu the CI is still failing on a pubmatic analytics adapter test.

modules/sonobiBidAdapter.js
aleksatr: are you sure that this way you're actually cloning the object properly? Check this example:
image
There's utils.deepClone() inside pbjs that you can utilize for cloning.

aleksatr:
> @aleksatr lint errors were fixed bu the CI is still failing on a pubmatic analytics adapter test.

np, went through on rerun.

aleksatr:
LGTM

PR #6366, opened 2021-02-27
LiveRamp userId: updated atype parameter

ian-lr:
## Type of change
- [x] Bugfix
- [x] Other

## Description of change
Some small changes related to the `atype` associated with the LiveRamp UserId module. Per [AdCOM](https://github.com/InteractiveAdvertisingBureau/AdCOM), `3` is the correct int for our identifier. Both the .js and .md files have been updated.

### DIFF
--- a/modules/userId/eids.js
+++ b/modules/userId/eids.js
@@ -71,7 +71,7 @@ const USER_IDS_CONFIG = {
   // identityLink
   'idl_env': {
     source: 'liveramp.com',
-    atype: 1
+    atype: 3
   },
 
   // liveIntentId
--- a/modules/userId/eids.md
+++ b/modules/userId/eids.md
@@ -52,8 +52,8 @@ userIdAsEids = [
     {
         source: 'liveramp.com',
         uids: [{
-            id: 'some-random-id-value',
-            atype: 1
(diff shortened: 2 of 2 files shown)


jdwieland8282:
LGTM

jdwieland8282:
CircleCl tests are failing

patmmccann:
this would be true about britepool, merkle, and verizon too right?

ian-lr:
@patmmccann I think that's right. May also apply to NetId, Fabrick, ID5, etc.

smenzer:
agree that in some other cases they should update to `3`; for ID5 today we are focused on device Identifiers, which is `1`, so we'll leave ours for now. 

i think the complicated part will come into play when we (or other identity providers) support both `1` and `3` ... today the user id module (specifically the way `eids.js` is built) doesn't have a way to modify on the fly the value for `atype`. maybe this should be an enhancement for id providers that need it? i could see `atype` being a function that could set the value much like we have for the extensions today.

patmmccann:
hey @ian-lr I did this in #6477 for several eids

ian-lr:
@patmmccann Awesome, I'll close this out.

PR #6376, opened 2021-03-02
Adform Bid Adapter: add global targeting to the request query as parameter

Pupis:
## Type of change
- [x] Feature

## Description of change
When clients use the same targeting parameters for all bids, it expands the bid requests URL. This change adds global targeting parameters that repeat in all bids to the request query string as global parameters and saves query string length.

 Scope.FL.Scripts@adform.com


FilipStamenkovic (approved):
LGTM

Issue #6380, opened 2021-03-02
Dchain objects


patmmccann:
## Type of issue
Standardization request

## Description
With the adoption of standard bid.meta fields as per https://github.com/prebid/Prebid.js/issues/3115 in Prebid 4.0 and their expected enforcement in https://github.com/prebid/Prebid.js/issues/6345 with Prebid 5.0, the spec should naturally extend to where to place the demand chain object bidders receive or are required to create. 

The IAB DemandChain working spec indicates "The DemandChain object should be included in the BidResponse.seatbid.bid.ext.dchain attribute in OpenRTB 2.5 or later. For OpenRTB 2.4 or prior, the BidResponse.ext.dchain attribute should be used." 

Publishers likely prefer to parse complicated demand chain objects themselves over having SSPs return parsed fields into the various existing meta objects, particularly if they are using analytics or security vendors which are able to parse them.

I propose the preferred way to pass the object back is for it to be copied to bid.meta.dchain, with the ssp adding a node as per the spec. Also as per the spec, if the SSP receives a bid with no dchain, it should initiate one eg `"dchain" : {
  "ver": "1.0",
  "complete" : 0,
  "nodes" : [
    {
      "asi": "magnite.com",
      "bsid": "123456789",
    }
  ]
}
` either from its adapter code or in its endpoint response. The prebid server adapter bid response should always include a dchain. 

A future validation module will likely have options for size validation, dchain validation, meta.advertiserDomains validation, mime type validation etc.

ChrisHuie:
#6383 merge closes this issue

PR #6383, opened 2021-03-04
PBS Bid Adapter: add dchain (demand chain object) to prebid server adapter

patmmccann:
## Type of change

- [ ] Bugfix
- [X] Feature
- [ ] New bidder adapter   
- [ ] Build related changes
- [ ] CI related changes
- [ ] Other

## Description of change
solves #6380 for prebid server adapter

### DIFF
--- a/modules/prebidServerBidAdapter/index.js
+++ b/modules/prebidServerBidAdapter/index.js
@@ -930,7 +930,7 @@ const OPEN_RTB_PROTOCOL = {
           if (bid.burl) { bidObject.burl = bid.burl; }
           bidObject.currency = (response.cur) ? response.cur : DEFAULT_S2S_CURRENCY;
           bidObject.meta = bidObject.meta || {};
-          if (bid.ext && bid.ext.dchain) { bidObject.meta.dchain = bid.ext.dchain; }
+          if (bid.ext && bid.ext.dchain) { bidObject.meta.dchain = utils.deepClone(bid.ext.dchain); }
           if (bid.adomain) { bidObject.meta.advertiserDomains = bid.adomain; }
 
           // the OpenRTB location for "TTL" as understood by Prebid.js is "exp" (expiration).
--- a/modules/prebidServerBidAdapter/index.js
+++ b/modules/prebidServerBidAdapter/index.js
@@ -930,6 +930,7 @@ const OPEN_RTB_PROTOCOL = {
           if (bid.burl) { bidObject.burl = bid.burl; }
           bidObject.currency = (response.cur) ? response.cur : DEFAULT_S2S_CURRENCY;
           bidObject.meta = bidObject.meta || {};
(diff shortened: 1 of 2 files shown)


smenzer:
@patmmccann has this solution been agreed to? code looks straightforward enough but I didn't see any comments in the Issue so I'm not sure if it requires discussion/approval before being merged.  

cc @gglas in case i'm missing any context

patmmccann:
> @patmmccann has this solution been agreed to? code looks straightforward enough but I didn't see any comments in the Issue so I'm not sure if it requires discussion/approval before being merged.
> 
> cc @gglas in case i'm missing any context

#6380 was adopted as noncontroversial in yesterday's prebid js meeting

smenzer:
> #6380 was adopted as noncontroversial in yesterday's prebid js meeting; this pr was requested during the discussion. also tagging @bretg who agreed to take a look at it. Thanks Bret and Scott!

Ok thanks ... I missed the meeting due to a conflict, so thanks for sharing.

modules/prebidServerBidAdapter/index.js
```diff
@@ -930,6 +930,7 @@ const OPEN_RTB_PROTOCOL = {
           if (bid.burl) { bidObject.burl = bid.burl; }
           bidObject.currency = (response.cur) ? response.cur : DEFAULT_S2S_CURRENCY;
           bidObject.meta = bidObject.meta || {};
+          if (bid.ext && bid.ext.dchain) { bidObject.meta.dchain = bid.ext.dchain; }
```
bretg, line 933: Based on the unit test below, dchain is an object, not a stringified object, so I think there's a problem here.
This isn't a deep copy -- believe the right way to copy a full object would be something like this:

```
Object.assign(bidObject.meta.dchain, bid.ext.dchain);
```

@ChrisHuie - please confirm?

patmmccann: happy to make that change, but i'm curious why my test would have passed if this is a problem

ChrisHuie: I think I am fully understanding what you are asking. Object.assign doesn't exactly deep copy but it does create a new clone from the source object that copies over the properties and is then allocated differently in memory. The "=" sign, in this case, instead is referencing the source property which acts just like a pointer to the same memory slot. So if bidObject.meta.dchain changes then bid.ext.dchain would also change and vice-versa I believe. With object.assign the changed source property wouldn't affect bid.ext.dchain.

I think object.assign is better for the side effect reason but shouldn't impact these tests

patmmccann: i changed to utils.deepClone

bretg (approved):

ChrisHuie (approved):

bretg:
docs PR https://github.com/prebid/prebid.github.io/pull/2737

PR #6392, opened 2021-03-08
Add render video in banner

richaudience:
## Type of change

- [ x ] Bugfix
- [ ] Feature
- [ ] New bidder adapter   
- [ ] Build related changes
- [ ] CI related changes
- [ ] Other

## Description of change

Add render video in banner

```
{
  bidder: '<bidder name>',
  params: {
    // ...
  }
}
```

### DIFF
--- a/modules/richaudienceBidAdapter.js
+++ b/modules/richaudienceBidAdapter.js
@@ -14,20 +14,20 @@ export const spec = {
   supportedMediaTypes: [BANNER, VIDEO],
 
   /***
-     * Determines whether or not the given bid request is valid
-     *
-     * @param {bidRequest} bid The bid params to validate.
-     * @returns {boolean} True if this is a valid bid, and false otherwise
-     */
+   * Determines whether or not the given bid request is valid
+   *
+   * @param {bidRequest} bid The bid params to validate.
+   * @returns {boolean} True if this is a valid bid, and false otherwise
+   */
   isBidRequestValid: function (bid) {
     return !!(bid.params && bid.params.pid && bid.params.supplyType);
   },
   /***
-     * Build a server request from the list of valid BidRequests
-     * @param {validBidRequests} is an array of the valid bids
-     * @param {bidderRequest} bidder request object
-     * @returns {ServerRequest} Info describing the request to the server
-     */
+   * Build a server request from the list of valid BidRequests
+   * @param {validBidRequests} is an array of the valid bids
+   * @param {bidderRequest} bidder request object
+   * @returns {ServerRequest} Info describing the request to the server
+   */
   buildRequests: function (validBidRequests, bidderRequest) {
     return validBidRequests.map(bid => {
       var payload = {
@@ -77,11 +77,11 @@ export const spec = {
     });
   },
   /***
-     * Read the response from the server and build a list of bids
-     * @param {serverResponse} Response from the server.
-     * @param {bidRequest} Bid request object
-     * @returns {bidResponses} Array of bids which were nested inside the server
-     */
+   * Read the response from the server and build a list of bids
+   * @param {serverResponse} Response from the server.
+   * @param {bidRequest} Bid request object
+   * @returns {bidResponses} Array of bids which were nested inside the server
+   */
   interpretResponse: function (serverResponse, bidRequest) {
     const bidResponses = [];
     // try catch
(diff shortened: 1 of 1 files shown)


robertrmartinez:
Assigned to @ncolletti 

But please add a unit test or two to cover the code change.

modules/richaudienceBidAdapter.js
```diff
@@ -103,10 +103,12 @@ export const spec = {
...
         bidResponse.vastXml = response.vastXML;
         try {
-          if (JSON.parse(bidRequest.data).videoData.format == 'outstream') {
-            bidResponse.renderer = Renderer.install({
-              url: 'https://cdn3.richaudience.com/prebidVideo/player.js'
-            });
+          if (bidResponse.vastXml != null) {
+            if (JSON.parse(bidRequest.data).videoData.format == 'outstream' || JSON.parse(bidRequest.data).videoData.format == 'banner') {
+              bidResponse.renderer = Renderer.install({
+                url: 'https://cdn3.richaudience.com/prebidVideo/player.js'
```
ncolletti, line 109: It will be helpful to publishers configuring Outstream in different variations for your adapter to pass additional information to Renderer.install like 'id', 'config', 'loaded', and adUnitCode

modules/richaudienceBidAdapter.js
```diff
@@ -197,6 +199,10 @@ function raiGetVideoInfo(bid) {
       playerSize: bid.mediaTypes.video.playerSize,
       mimes: bid.mediaTypes.video.mimes
     };
+  } else {
+    videoData = {
+      format: 'banner'
```
ncolletti, line 204: Seems strange to be defining a video bid as 'banner', shouldn't this be checking the video.context for 'outstream' then apply that to the 'format'? You would then check against that in line 107. Maybe there is an edge case I am missing?

test/spec/modules/richaudienceBidAdapter_spec.js
```diff
@@ -75,26 +75,19 @@ describe('Richaudience adapter tests', function () {
...
     bidder: 'richaudience',
     params: {
       bidfloor: 0.5,
       pid: 'ADb1f40rmi',
-      supplyType: 'site',
-      player: {
-        init: 'close',
-        end: 'close',
-        skin: 'dark'
-      }
```
ncolletti, line 97: This should also be removed from your markdown docs example

test/spec/modules/richaudienceBidAdapter_spec.js
```diff
@@ -262,6 +255,7 @@ describe('Richaudience adapter tests', function () {
 
     expect(requestContent).to.have.property('demand').and.to.equal('video');
     expect(requestContent.videoData).to.have.property('format').and.to.equal('instream');
+    // expect(requestContent.videoData).to.have.property('mimes').and.to.equal("['video/mp4']");
     // expect(requestContent.videoData.playerSize[0][0]).to.equal('640');
     // expect(requestContent.videoData.playerSize[0][0]).to.equal('480');
```
ncolletti, line 260: All commented code can be removed

richaudience: I remove the commented code

test/spec/modules/richaudienceBidAdapter_spec.js
richaudience: the test has been removed

modules/richaudienceBidAdapter.js
richaudience: The definition is correct and it’s done willingly, we’re able to serve video demand on banner spaces if the final partner wishes to. That’s managed from our internal platform, flagging the placements as multi-format. The adaptor must be able to support both formats as well.

modules/richaudienceBidAdapter.js
richaudience: From Rich Audience, we don’t need additional info since our player does render the video correctly already with the current params. Do you think that these params would be mandatory?

ncolletti: The additional information helps Prebid core to parse renderers in the auction. There are many places a renderer can be chosen. If you look to other modules it shouldn't be hard to populate these values.
id = bidrequest.id
adunitcode = bidrequest.adunitcode
Defining loaded = false is important since you are loading an external script for your renderer.
config = 'mediaTypes.video' or 'oustream'

modules/richaudienceBidAdapter.js
ncolletti: That makes sense and is defined as Oustream within Prebid. You have a conditional on line 107 that looks for the videoData.format === 'outstream' but you are not setting this value anywhere, it will always fail. It should be looking at the bidRequest.mediaType.video.context: https://docs.prebid.org/dev-docs/bidder-adaptor.html#ingesting-the-video-context

richaudience: Hey Niko,
That’s a quick scheme on how are we going to use the new param info:
Config type — demand type — expected render

video -> video -> video
video -> banner -> none
banner -> banner -> banner
banner -> video -> video (If the publisher mismatches the config, this case is a safe check)

video-video
video-banner
banner-banner
banner-video

Cheers

ncolletti: Thank you for the explanation and for adding the additional fields to your Renderer!

ncolletti (approved):
LGTM

PR #6396, opened 2021-03-08
NextRoll ID System: add new ID module

abijr:
## Type of change
- [x] New ID Module  
- [x] official module submission

## Description of change

This PR adds the ID module for the NextRoll ID.

- Maintainer email: prebid@nextroll.com
- PR on the docs repo: prebid/prebid.github.io#2756


bretg:
docs PR https://github.com/prebid/prebid.github.io/pull/2756

Rothalack:
Seems like everything is working for me in testing.

I think all you have left to do is add your module to these files.

https://github.com/prebid/Prebid.js/blob/master/modules/userId/eids.md
https://github.com/prebid/Prebid.js/blob/master/modules/userId/userId.md

Rothalack (approved):

Rothalack:
Hey @smenzer, do I need to wait for your review or do you think this is good to go? I'm not certain if this falls into the category of needing 2 reviews or not.

smenzer:
@Rothalack id modules don't require a 2nd review per se, but never a bad thing to ask for a 2nd pair of eyes if you want.

modules/userId/userId.md
```diff
@@ -148,7 +148,16 @@ pbjs.setConfig({
...
-        }, {
+    }, {
+            name: 'nextroll',
+            params: {
+                partnerId: "1009",         // Set your real NextRoll partner ID here for production
+            },
+            storage: {
+                type: 'html5',
+                name: '_nextroll_id',
+            },
```
smenzer, line 159: since you're directly accessing localstorage rather than using the user id module to retrieve the cached id, you should remove the `storage` configuration from your examples so publishers don't use it, which could cause issues with your module.

modules/nextrollIdSystem.js
```diff
@@ -0,0 +1,50 @@
...
+
+export const storage = getStorageManager();
+
+/** @type {Submodule} */
+export const nextrollIdSubmodule = {
+  /**
+   * used to link submodule with config
+   * @type {string}
+   */
+  name: 'nextroll',
```
smenzer, line 23: this should be `nextrollId`

modules/nextrollIdSystem.js
```diff
@@ -0,0 +1,50 @@
...
+   * @param {SubmoduleConfig} [config]
+   * @returns {{id: {nextroll: string} | undefined}}
+   */
+  getId(config) {
+    const key = KEY_PREFIX + deepAccess(config, 'params.partnerId', 'undefined');
+    const dataString = storage.getDataFromLocalStorage(NEXTROLL_ID_LS_KEY) || '{}';
+    const data = JSON.parse(dataString);
+    const nextrollId = deepAccess(data, `${key}.value`);
+
+    return { id: nextrollId ? {nextroll: nextrollId} : undefined };
```
smenzer, line 46: I don't think this is the return object you want, considering the `decode()` method is just returning whatever you set here. the object should have a top level key for your id module name: `nextrollId` then the value is typically just the ID string itself.

```suggestion
    return nextrollId ? { nextrollId: nextrollId } : undefined;
```

modules/userId/userId.md
```diff
@@ -148,7 +148,16 @@ pbjs.setConfig({
                 expires: 90,          // Expiration in days
                 refreshInSeconds: 8*3600 // User Id cache lifetime in seconds, defaulting to 'expires'
             },
-        }, {
+    }, {
+            name: 'nextroll',
```
smenzer, line 152:
```suggestion
            name: 'nextrollId',
```

test/spec/modules/nextrollIdSystem_spec.js
```diff
@@ -0,0 +1,56 @@
...
+  });
+
+  afterEach(function () {
+    sandbox.restore();
+  })
+
+  const testCases = [
+    {
+      expect: {
+        id: {nextroll: 'id_value'},
```
smenzer, line 24:
```suggestion
        nextrollId: 'id_value',
```

smenzer (changes requested):
@abijr you should add tests to the `userId_spec` and `eids_spec` files to ensure your module is working as expected.

modules/nextrollIdSystem.js
abijr: This suggestion is wrong. See https://github.com/Prebid/Prebid.js/blob/3fd2bb57a4c849b0cba573b6f3ddc2b65f41e76a/modules/userId/index.js#L660

The return object is not stored as is, the `object.id` is what gets stored.

modules/nextrollIdSystem.js
abijr: Documentation mentioned it was recommended to use the `Id` suffix, but unfortunately our Adapter module is currently using the `nextroll`  name.  We'd like to keep this name if possible, what's your opinion on this?

test/spec/modules/nextrollIdSystem_spec.js
abijr: See https://github.com/prebid/Prebid.js/pull/6396#discussion_r604393088 . This test case is correct, it's indeed confusing, but the value is processed differently when it's not the result of the callback.

I've made sure to double check in a testpage.

smenzer: you're right, all good

smenzer (changes requested):
LGTM, but you should add tests to `userId_spec.js` and `eids_spec.js` files to ensure your module is working as expected before we can merge this

abijr:
I believe this last commit should address those changes.

smenzer (approved):
LGTM

Issue #6389, opened 2021-03-05
Pre-defined native types: Image not replaced before adapters are called


guiann:
Hello everyone, 

Our bidder supports native ads for a few weeks now and we can  see some native bid format with 

```
native: {
    type: "image"
}
```

I would expect this native configuration to be replaced by the full format config specified in:
[pre-defined-native-types](https://docs.prebid.org/dev-docs/show-native-ads.html#pre-defined-native-types)

```
{
    image: { required: true },
    title: { required: true },
    sponsoredBy: { required: true },
    clickUrl: { required: true },
    body: { required: false },
    icon: { required: false },
    cta: { required: false }
}
```

Is it expected that bidder adapters receive the short hand format?
That would mean every bidder has to expand every pre-defined types. Duplicating the same code for every bidder and breaking futur integrations when new pre-formated types will be added.

guiann:
Our publisher is sending this config:
```
native: {
    size: [1, 1],
    type: "image"
}
```
Can the useless 'size' field be the source of this issue? 
Were trying to make them remove it...

gglas:
@aleksatr @robertrmartinez We discussed this on issue review today, and this seems like a good place to introduce the concept of prebid Native templates with predefined requirements that are then accessed via Shorthand.  Aleksa I tagged you until we can add Chris N to prebid, I think he'd find this interesting.

patmmccann:
@guiann this expansion of the shorthand inside of core seems like a great utils for all adapters; is there any chance you could take a stab at it?

guiann:
No I've no time to handle it sorry.

patmmccann:
Seems to

PR #6401, opened 2021-03-10
AdYouLike Bid Adapter: replace shorthand "image" native config

guiann:
## Type of change
 Feature:  handle short native config  type: "image"
This change relates to this issue: 
https://github.com/prebid/Prebid.js/issues/6389
Bidder adapters have to replace short native config by themselves

## Description of change
handle the configuration : 
```
native: {
    type: "image"
}
```
in adyoulike bidder adapter.

- [*] official adapter submission

## Other information
I don't think bidder adapters should have to handle this case. If more shorthand config are added in te future, all bidder adapters will  have to be updated to support them.


modules/adyoulikeBidAdapter.js
```diff
@@ -44,7 +68,13 @@ export const spec = {
...
         accumulator[bid.bidId].Height = size.height;
         accumulator[bid.bidId].AvailableSizes = sizesArray.join(',');
-        if (bid.mediaTypes && bid.mediaTypes.native) accumulator[bid.bidId].Native = bid.mediaTypes.native;
+        if (bid.mediaTypes && bid.mediaTypes.native) {
+          let nativeReq = bid.mediaTypes.native;
+          if (nativeReq.type === 'image') {
+            nativeReq = NATIVE_IMAGE;
+          }
+          accumulator[bid.bidId].Native = nativeReq;
+        }
```
robertrmartinez, line 77: @guiann So when a pub defines a "short native config type" will it always only have the single key `type` ?

Is it possible that a pub would want to use the short type, but also be able to override the default fields?

Maybe instead of 

```js
nativeReq = NATIVE_IMAGE;
```

We merge the pub provided `bid.mediaTypes.native` onto the NATIVE_IMAGE default?

robertrmartinez (commented):
Unsure of use case so have a question.

modules/adyoulikeBidAdapter.js
guiann: I was thinking if a publisher uses the shorthand they won't want to spend time with fine tuning...  but  yes we can handle this case.  I'll do the merge of the two objects.

Again, do every adapters have to handle this case by themselves ? Unfolding shorthand formats should be a more global treatment, don't you think?

guiann: change is done.

robertrmartinez: Yeah, agrees we should handle shorthands in core prebid

robertrmartinez:
Please add a unit test covering this change and we will. merge it

robertrmartinez (approved):

modules/adyoulikeBidAdapter.js
guiann: You can upvote the issue I've opened   here :
https://github.com/prebid/Prebid.js/issues/6389

Thank you.

ChrisHuie (approved):

Issue #6078, opened 2020-12-02
VideoCacheKey override


bretg:
## Type of issue
Enhancement

## Description
8 of 271 bidders set bidResponse.videoCacheKey, which means they cause notable headaches in setting up line items because they require separate GAM video creatives. It's time to talk about this.

- bluebillywigBidAdapter.js - @K-JBoon 
- dailyhuntBidAdapter.js - @montu1996 
- nobidBidAdapter.js - @redaguermas 
- onetagBidAdapter.js - @OneTagDevOps 
- rubiconBidAdapter.js - me
- spotxBidAdapter.js - @agdillon , @npeceniak 
- synacormediaBidAdapter.js - @slaufer, @amuraco 
- theAdxBidAdapter.js - @mustafakemal16 

Basically, bid adapters that specify videoCacheKey are saying "you must use my VAST cache location, you may not use your own."

Here's the relevant pseudo-code in auction.js:
```
For each bidResponse
 If instream and setconfig({cache.url}) is specified     // see note below
   If videoCacheKey is specified
      If there’s no vastURL, error, don’t add bid to the auction yet
      else bid is totally ready, so add it to the auction.
   If videoCacheKey is **not** specified call prebid-cache
      The response from the cache triggers adding the bid and setting videoCacheKey
```
Note: the assumption from PR #3024 is that outstream renderers can support either VAST XML or IRL.

I can speak from the Rubicon Project perspective on this. We provide VAST caching on our servers for our customers because we believed it was The Right Thing To Do:
1. It's our business so it's our responsibility to store these big globs of VAST on servers we pay for. It didn't seem fair to make AppNexus be the global cache for the internet. These things aren't free and we should shoulder our share.
2. Hosting our VAST in our own cache gives us the ability to own the troubleshooting. Publishers pay us and we don't want to shrug our shoulders and point a finger at some other cache provider.
3. Client-side caching is inefficient. Schlepping VAST back and forth across the poor user's network connection is ugly if not kinda rude.

However, after experiencing the convolutions required to manage different video cache locations in GAM, Magnite (and therefore Rubicon Project) would be happy to allow publishers to use VAST wrappers around our VAST URLs to simplify line item setup.

Even if the other 7 bidders feel the same way, we can't just remove videoCacheKey logic because pubs may have special line items and creatives set up. Removing the logic from either auction.js or the bid adapter because that would cause a new _hb_uuid_ to be generated that might not match up properly with the VAST URL in GAM.

[...]

PR #6413, opened 2021-03-12
Spotx Bid Adapter: add publisher support for cache.ignoreBidderCacheKey

npeceniak:
## Type of change

- [ ] Bugfix
- [x] Feature
- [ ] New bidder adapter   
- [ ] Build related changes
- [ ] CI related changes
- [ ] Other

## Description of change

Support cache.ignoreBidderCacheKey publisher option. If a publisher sets ignoreBidderCacheKey to true we will now return a full XML vast response instead of wrapper so that the ad can be cached at the publishers specified cache location and avoid an extra layer of vast wrapper. 

## Other information
https://github.com/prebid/Prebid.js/issues/6078


harpere (approved):

harpere:
please fix lint problem.

npeceniak:
@harpere All linting issues have been resolved.

harpere (approved):

patmmccann:
@khatibda reporting this breaks your integration as the key ends up being missing from prebid cache, looking at it this seems to be breaking.

patmmccann:
@r-schweitzer confirming if issue duplicated for us and potential fix

bretg:
SpotX team - I think there was a miscommunication here.  It was not the intention of https://github.com/prebid/Prebid.js/issues/6078 that you were supposed to remove the videoCacheKey. It appears this is going to cause SpotX render problems for publishers who upgrade to 4.31.

The intention was to give publishers a path to simplifying their ad server setup.  The issue did say "in a major release (e.g. 5.0 or 6.0) we remove the videoCacheKey logic from either auction.js or the bid adapter."  But removing this value in 4.31, publishers didn't have time to update their line items.

npeceniak:
I am not sure I fully understand this issue that your are seeing @khatibda. It was my understanding that when using the prebid cache the `videoCacheKey` would be populated automatically by code in auction.js
```
For each bidResponse
 If instream and setconfig({cache.url}) is specified     // see note below
   If videoCacheKey is specified
      If there’s no vastURL, error, don’t add bid to the auction yet
      else bid is totally ready, so add it to the auction.
   If videoCacheKey is **not** specified call prebid-cache
      The response from the cache triggers adding the bid and setting videoCacheKey
 ```

Here is an example bid response from a test I ran on v3.35.0-pre using the prebid cache this morning. 
```

npeceniak:
As for the case where we are not using the prebid cache it is true that we no longer send a `videoCacheKey` in the bid response but the `cache_key` contains the exact same information. As well as the `vastUrl` contains the entire url the the spotx cache location, including the cache key. Here is an example of a bid response when using the spotx cache. 
```
cache_key: "eyJob3N0IjoiZmUyNzMuc3BvdHguZGVuMDEucG9wIiwiY2hhbm5lbCI6Ijg1Mzk0IiwicHViIjoiNzU5NDIiLCJndWlkIjoiZWYyMDgyNmY5ZTEzMTFlYmE2YzUxMzllYjAxNTAwMDEtOjEyOGE2LjQ0ZDc0LjQ2YjM6IiwidHRsIjoxNjE4NTIwMjAxLCJ2IjoiMi4wIiwiYkNvcnMiOnRydWUsImhhc2giOiIyMzg3NzM1ODZkZjEyY2M5YmE1MTIyNTlhODI1MjIyZjFjN2Q3ZDM1YWNmYmZjZGQ3NThlYTY2Yjc4MDMxMTk4In0"
vastUrl: "https://search.spotxchange.com/ad/vast.html?key=eyJob3N0IjoiZmUyNzMuc3BvdHguZGVuMDEucG9wIiwiY2hhbm5lbCI6Ijg1Mzk0IiwicHViIjoiNzU5NDIiLCJndWlkIjoiZWYyMDgyNmY5ZTEzMTFlYmE2YzUxMzllYjAxNTAwMDEtOjEyOGE2LjQ0ZDc0LjQ2YjM6IiwidHRsIjoxNjE4NTIwMjAxLCJ2IjoiMi4wIiwiYkNvcnMiOnRydWUsImhhc2giOiIyMzg3NzM1ODZkZjEyY2M5YmE1MTIyNTlhODI1MjIyZjFjN2Q3ZDM1YWNmYmZjZGQ3NThlYTY2Yjc4MDMxMTk4In0"
```
@bretg Are you suggesting that we add the `videoCacheKey` back in cases where we are using the spotx cache location instead of the prebid cache?

bretg:
@npeceniak - the problem is that by making this change without telling publishers, they aren't aware they have to change their ad server creatives. 

1) those vendors who set videoCacheKey (like SpotX and Rubicon) require publishers to make special video header bidding creatives. These creatives point directly to that vendor's cache host (e.g. spotx or rubicon) This is necessary because GAM doesn't allow variables in the host portion of the URL. So publishers that support SpotX pre-4.31 have hardcoded something like this into their GAM creatives as the VAST URL -- https://spotx.com/cache?uuid=%%PATTERN:hb_uuid_BIDDERCODE%%

2) So when you take away videoCacheKey, this causes Prebid to do client-side caching, generating a brand new UUID and sending that to the ad server.

3) So when GAM chooses the creative, it resolves the spotx URL with the Prebid-generated cache location

4) But of course spotx.com doesn't know anything about the cache key that Prebid generated. Therefore, rendering is broken.

Yes, I recommend you make a change -- if you see ignoreBidderCacheKey=true, then it's safe to assume that the publisher's ad server is ready for Prebid-generated cache keys. Otherwise, they still need yours.

npeceniak:
@bretg Thank you very much for the explanation. I have submitted a pull request to add `videoCacheKey` back to the spotx bid response if a publisher is not using the prebid cache. - https://github.com/prebid/Prebid.js/pull/6605

bretg:
I've added a note to your spotx.md documentation that will warn publishers about the interaction between spotx 4.31-4.35 and ignoreBidderCacheKey. You guys are welcome to fine-tune this note.

https://github.com/prebid/prebid.github.io/pull/2874

PR #6417, opened 2021-03-15
VLYBY Bid Adapter: add new bid adapter

thomas-vlyby:
## Type of change
- [ ] Bugfix
- [ ] Feature
- [X]  New bidder adapter  
- [ ] Build related changes
- [ ] CI related changes
- [ ] Other

## Description of change

```
{
  bidder: 'vlyby',
  params: { 
     publisherId: 'f363eb2b75459b34592cc4',   
       siteId: 'techpreview.vlyby_prebidadapter',
       placement:'default'                      
    }
}
```
- [x] Demo Website with GPT Implementation is [here](https://www.sluga.io/vlyby/prebid/)


ChrisHuie (changes requested):
@thomas-vlyby can you please add test coverage also for this new bid adapter please

modules/vlybyBidAdapter.js
```diff
@@ -0,0 +1,76 @@
+import { registerBidder } from '../src/adapters/bidderFactory.js'
+
+import { BANNER, VIDEO } from '../src/mediaTypes.js'
+import { config } from '../src/config.js';
+
+const ENDPOINT = '//europe-west3-vlybypoc2019.cloudfunctions.net/prebid';
+const BIDDER_CODE = 'vlyby';
```
muuki88, line 7: Is there a `GVL_ID` that can be associated with this  bidder? Not sure if this plays well with the `gdprEnforcement` module

modules/vlybyBidAdapter.js
```diff
@@ -0,0 +1,76 @@
...
+        },
+        gdprConsent: {
+          consentString: gdprConsent.consentString,
+          gdprApplies: gdprConsent.gdprApplies
+        },
+        bidRequests: validBidRequests.map(({ params, sizes, bidId, adUnitCode }) => ({
+          bidId,
+          adUnitCode,
+          params,
+          sizes
```
muuki88, line 44: What's the effect of the `sizes` being submitted. Is this only for `BANNER` ads or does the player take them into account too?

modules/vlybyBidAdapter.js
```diff
@@ -0,0 +1,76 @@
...
+
+import { BANNER, VIDEO } from '../src/mediaTypes.js'
+import { config } from '../src/config.js';
+
+const ENDPOINT = '//europe-west3-vlybypoc2019.cloudfunctions.net/prebid';
+const BIDDER_CODE = 'vlyby';
+
+export const spec = {
+  code: BIDDER_CODE,
+  supportedMediaTypes: [VIDEO, BANNER],
```
muuki88, line 11: Why banner if this is an outstream implementation?

ChrisHuie:
@thomas-vlyby following up on the requested changes to move forward with this pr

muuki88:
Could be that @thomas-vlyby is not available for a few weeks.

modules/vlybyBidAdapter.js
thomas-vlyby: @muuki88 There is no Global Vendor List ID. Due to the fact, that we are not processing any user data. For gdprEnforcement. Every Implementation needs a provided user ID and contract between VLYBY and Publisher. So for GPDR we will communicate direct with Publisher. Maybe gdpr.rules[].vendorExceptions should be used in this case.

modules/vlybyBidAdapter.js
thomas-vlyby: sizes and Banner are special case. For some Publisher we will provide Banner with sizes like 300x250 or 300x600

modules/vlybyBidAdapter.js
thomas-vlyby: In general this Adapter provides a outstream Mediation for Publishers. They can use this Adapter in AdServer with a price Point from Endpoint. In some cases Publisher can use Banner. This makes this adapter more flexibel, for examaple to add Banner with cheaper eCPM for Fallback etc.

modules/vlybyBidAdapter.js
muuki88: As long as you don't add any user syncs this is probably fine.

However you are a registered vendor now, so why not use it?

thomas-vlyby:
> @thomas-vlyby can you please add test coverage also for this new bid adapter

@ChrisHuie Test coverage added

ChrisHuie:
@thomas-vlyby we recently pushed Preebid 5.0 and per #6650 we now require bid adapters to at least pass advertiserdomain. Can you please add "adomain" to the response in this pr 🙏

thomas-vlyby:
@ChrisHuie adomain added to response and PR in doc created https://github.com/prebid/prebid.github.io/pull/3064

ChrisHuie (approved):

PR #5235, opened 2020-05-14
fix(Renderer): load script only on render

jeremiegirault:
## Type of change
- [x] Bugfix

## Description of change

I was playing with Renderer when creating a custom adapter with outstream video support. So I created a Renderer in order to display a player when retrieving some VAST response.
I noticed that the Renderer class loads the external script in it's constructor, not in the render function. Therefore external scripts are loaded, not only when the bid in won but when bid responses are received.
That can cause multiple requests during the auction and bidders that do not win can load any external scripts and access the whole page under the "outstream" module code. That can lead to strategies of sending empty bids to execute random code on the page.

### note

I can't run or test anything because a transitive depenency (fiber) prevents me from installing dependencies... (outdated, not building on osx 10.14 / node 13)

### initial issue
https://github.com/prebid/Prebid.js/issues/5117

PR #6422, opened 2021-03-16
PBJS Core: call custom render func after loadscript if provided

jeremiegirault:
## Type of change

- [ ] Bugfix
- [ ] Feature
- [ ] New bidder adapter   
- [ ] Build related changes
- [ ] CI related changes
- [x] Other: Small behavior change

## Description of change
As per comments on #5235 some users expected the renderer render function to be called once the script url is loaded.
It was written as a comment that `// _render is expected to use push as appropriate` however that would not be really developer friendly. This change allows the render function to be called once the script url is loaded or right now if no script url is available.

## Other information
@ACannuniRP will be interested by this PR

### DIFF
--- a/src/Renderer.js
+++ b/src/Renderer.js
@@ -39,17 +39,22 @@ export function Renderer(options) {
 
   // use a function, not an arrow, in order to be able to pass "arguments" through
   this.render = function () {
+    const renderArgs = arguments
+    const runRender = () => {
+      if (this._render) {
+        this._render.apply(this, renderArgs)
+      } else {
+        utils.logWarn(`No render function was provided, please use .setRender on the renderer`);
+      }
+    }
+
     if (!isRendererPreferredFromAdUnit(adUnitCode)) {
       // we expect to load a renderer url once only so cache the request to load script
+      this.cmd.unshift(runRender) // should render run first ?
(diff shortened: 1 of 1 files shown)


src/Renderer.js
```diff
@@ -39,17 +39,22 @@ export function Renderer(options) {
 
   // use a function, not an arrow, in order to be able to pass "arguments" through
   this.render = function () {
+    const renderArgs = arguments
+    const runRender = () => {
+      if (this._render) {
+        this._render.apply(this, renderArgs)
+      } else {
+        utils.logWarn(`No render function was provided, please use .setRender on the renderer`);
```
patmmccann, line 47: is this happening frequently? should it throw an event for analytics adapters?

muuki88:
Does this have an effect on custom outstream renderers defined?

src/Renderer.js
jeremiegirault: I don't know. If the external script is standalone it can handle everything itself but that would be recommended to set a renderer

jeremiegirault:
@muuki88 I don't think so

msm0504 (approved):

Fawke (approved):
LGTM. Tested on one of the test pages, this functionality seems to be working fine.

PR #6426, opened 2021-03-17
Bid Glass Adapter: Merge externally set targeting params

dliebner:
## Type of change
- [x] Feature

## Description of change
Allows the Bid Glass adapter to merge externally set targeting parameters (if present) into the bid request.
No new unit tests because the new code is covered by existing unit tests / no new applicable unit tests could be designed to cover the changes.
This change fails the gulp-eslint tests because "bidglass" is undefined. The `typeof` check on line 80 will prevent any actual JavaScript errors in production --- the point of this update is to allow publishers to set bidglass-specific targeting parameters outside of the context of prebid because many publishers do not have granular control over prebid params for every single ad request (i.e. because of wrapper code, especially when their ad stack is managed by a third party). Our platform allows for extremely granular publisher-defined targeting of ad units, and we need a way to pass this targeting through via prebid for all publishers, and the solution we've chosen is to offer (in addition to via bidder params) an alternative targeting-setting method via our own library, which may or may not be environmentally present, hence the checks in the adapter code. If these errors are unacceptable, any alternative suggestions to fulfilling this objective of allowing external targeting by publishers who cannot practically set their bidder params on a request-by-request basis, whether it be on a code level or a design level, would be appreciated.

- Contact email of the adapter’s maintainer: dan@bid.glass


harpere (approved):

dliebner:
> This change fails the gulp-eslint tests because "bidglass" is undefined. The typeof check on line 80 will prevent any actual JavaScript errors in production --- the point of this update is to allow publishers to set bidglass-specific targeting parameters outside of the context of prebid because many publishers do not have granular control over prebid params for every single ad request (i.e. because of wrapper code, especially when their ad stack is managed by a third party). Our platform allows for extremely granular publisher-defined targeting of ad units, and we need a way to pass this targeting through via prebid for all publishers, and the solution we've chosen is to offer (in addition to via bidder params) an alternative targeting-setting method via our own library, which may or may not be environmentally present, hence the checks in the adapter code. If these errors are unacceptable, any alternative suggestions to fulfilling this objective of allowing external targeting by publishers who cannot practically set their bidder params on a request-by-request basis, whether it be on a code level or a design level, would be appreciated.

Fawke (approved):

harpere (approved):

PR #6427, opened 2021-03-17
Proxistore Bid Adapter: add cookieless url endpoint & use floor module

vincentproxistore:
## Type of change

- [ ] Bugfix
- [ X ] Feature
- [ ] New bidder adapter   
- [ ] Build related changes
- [ ] CI related changes
- [ ] Other

## Description of change
Use the floor module from prebid
Add a cookieless url endpoint if there is not constent from the user.


modules/proxistoreBidAdapter.js
```diff
@@ -81,23 +105,26 @@ function _createBidResponse(response) {
  */
 
 function isBidRequestValid(bid) {
-  const hasNoAd = function() {
+  const canDisplay = function () {
     if (!storage.hasLocalStorage()) {
+      utils.logError('Local storage API disabled');
```
aleksatr, line 110: Why are you logging this, is this a leftover from debugging?

modules/proxistoreBidAdapter.js
```diff
@@ -1,62 +1,86 @@
...
+    url: endPointUri,
     data: JSON.stringify(payload),
-    options: options
+    options: options,
   };
 }
 
+function _assignSegments(bid) {
+  if (bid.hasOwnProperty('fpd')) {
+    return bid.fpd.segments || [];
```
aleksatr, line 79: First party data is not provided through `fpd` anymore, there's ortb2 interface now, you can read more about it here: https://docs.prebid.org/features/firstPartyData.html

aleksatr:
@vincentproxistore unit test coverage for your adapter is bellow 80%, for any changes to be merged you need to bring the coverage above 80%.

To generate and view the code coverage reports:

```bash
gulp test-coverage
gulp view-coverage
```

aleksatr:
@vincentproxistore , why am I getting "CONNECTION REFUSED" when I try the example from your .md file (running hello_world.html integration example)?
image

Maybe because `withCredentials` is false here?
That leads me to another question, why am I hitting cookieless endpoint when gdpr does not apply to me at all, is this a bug? Maybe you don't want to treat `gdprApplies: false` in the same way as missing vendor consent.

aleksatr:
LGTM

PR #6432, opened 2021-03-17
ZetaSsp Bid Adapter: add new bid adapter

asurovenko-zeta:
## Type of change
- [ ] New bidder adapter

## Description of change
- [ ] New bidder adapter

```
{
    bidder: 'zeta_global_ssp',
    bidId: 12345,
    params: {
        placement: 12345,
        user: {
            uid: 12345,
            buyeruid: 12345
        },
        tags: {
            someTag: 123
        },
        test: 1
    }
}
```

### DIFF
--- a/modules/zetaSspBidAdapter.js
+++ b/modules/zetaSspBidAdapter.js
@@ -108,7 +108,10 @@ export const spec = {
         ad: zetaBid.adm,
         ttl: TTL,
         creativeId: zetaBid.crid,
-        netRevenue: NET_REV
+        netRevenue: NET_REV,
+        meta: {
+          advertiserDomains: zetaBid.adomain
+        }
       };
       bidResponse.push(bid);
     }
--- a/modules/zetaSspBidAdapter.js
+++ b/modules/zetaSspBidAdapter.js
@@ -1,0 +1,152 @@
+import * as utils from '../src/utils.js';
+import {registerBidder} from '../src/adapters/bidderFactory.js';
+import {BANNER} from '../src/mediaTypes.js';
+
+const BIDDER_CODE = 'zeta_global_ssp';
+const ENDPOINT_URL = 'https:/ssp.disqus.com/bid';
+const USER_SYNC_URL = 'https:/ssp.disqus.com/match';
+const DEFAULT_CUR = 'USD';
+const TTL = 200;
+const NET_REV = true;
+
+export const spec = {
+  code: BIDDER_CODE,
+  supportedMediaTypes: [BANNER],
+
+  /**
+   * Determines whether or not the given bid request is valid.
+   *
+   * @param {BidRequest} bid The bid params to validate.
+   * @return boolean True if this is a valid bid, and false otherwise.
+   */
+  isBidRequestValid: function (bid) {
+    // check for all required bid fields
+    if (!(bid &&
+      bid.bidId &&
+      bid.params)) {
+      utils.logWarn('Invalid bid request - missing required bid data');
+      return false;
+    }
+    return true;
+  },
+
+  /**
+   * Make a server request from the list of BidRequests.
+   *
+   * @param {Bids[]} validBidRequests - an array of bidRequest objects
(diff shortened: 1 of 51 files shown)


ChrisHuie:
@asurovenko-zeta can you please also make sure and submit a docs request to go along with this one? There are decently detailed instructions [here](https://docs.prebid.org/dev-docs/bidder-adaptor.html#required-files) if you scroll to the bottom of the pafe

asurovenko-zeta:
@ChrisHuie docs was added here: https://github.com/prebid/prebid.github.io/pull/2782

modules/zetaSspBidAdapter.js
```diff
@@ -0,0 +1,152 @@
...
+      let bid = {
+        requestId: zetaBid.impid,
+        cpm: zetaBid.price,
+        currency: zetaResponse.cur,
+        width: zetaBid.w,
+        height: zetaBid.h,
+        ad: zetaBid.adm,
+        ttl: TTL,
+        creativeId: zetaBid.crid,
+        netRevenue: NET_REV
```
patmmccann, line 111: are you able to set meta.advertiserDomains?

asurovenko-zeta: advertiserDomains was added along with test update

patmmccann: thank you!

ncolletti:
@asurovenko-zeta Given that the only items changing in this release compared to your current adapter are the endpoint and a new param, we suggest that you close this PR and instead update your existing adapter to handle a new bidder code for 'zeta_global_ssp'. This bidder code when read by your adapter will use the correct endpoint for the request and will handle the new param as well. You can look toward the AOL adapter for inspiration on how they achieved this same setup.

asurovenko-zeta:
@ncolletti we can't resolve which bidder code must be defined for request because all of different request params is optional and there are a lot of situations with almost the same request.

ncolletti:
Sorry I am not following why this is a problem? You would add aliases to your adapter, then look at the bid.bidder to determine how to read the rest of the params sent in the bidRequest. At that point you can point it to the correct endpoint and use the new param you are introducing with this PR in ext.tags

asurovenko-zeta:
@ncolletti
We are planned making other changes in new Ssp Zeta Adapter in the future, because of that we made the desicion that we would like to ask you to approve the current version to facilitate future changes.

CC: miakovlev@zetaglobal.com

asurovenko-zeta:
@ncolletti
Could you please look at this PR? I would like to speed up the process a little, our team is looking forward to this PR.
Thanks a lot!

ncolletti:
OK we can continue with this PR and submission of a separate adapter. It was noticed that both adapters are asking the user to pass geo within your private params, this information is generally considered as first party data and should be made available by the publisher to Prebid via the ortb2 object and setConfig, [details here](https://docs.prebid.org/features/firstPartyData.html#in-page-examples). In order to be more aligned with our FPD initiative, would you be open to updating geo to look in this location (ortb2.device.geo.country) instead of params?

asurovenko-zeta:
@ncolletti
We're happy to stick with your FPD initiative, I hope current solution on it.

ncolletti:
Thank you @asurovenko-zeta - Could you also update your docs pr to reflect no longer accepting device in params?  Eventually it would be great to remove this also from your other adapter as well as to look to the 'ortb2' first party data object in the global prebid context for the rest of these params (site, user, app). Maybe it makes sense in a separate commit to make that change to fall back to that location if they aren't in params.

asurovenko-zeta:
@ncolletti
The PR to docs repo was added: https://github.com/prebid/prebid.github.io/pull/2837
I have no authority to change our DSP adapter, but I'll pass the information to the DSP team.
Thanks!

ncolletti (approved):
LGTM. Would be great to consider another PR to use prebid 'ortb2' locations for the remaining objects in your params in this adapter. Thanks!

PR #6435, opened 2021-03-17
AMX Bid Adapter: video bugfix to set mediaType on bid instead of meta object

nickjacob:
## Type of change

- [X ] Bugfix

## Description of change

We were only setting mediaType in the `meta` object. It looks like the rest of the auction expects it to be set on the bid. Note the documentation [here](https://docs.prebid.org/dev-docs/bidder-adaptor.html) only mentions bid.meta.mediaType (we'll open a PR in the docs repo for that)

contact email of the adapter’s maintainer: prebid@amxrtb.com


ChrisHuie:
@nickjacob thanks for pointing this out and appreciate the docs pr :)

nickjacob:
Opened the docs PR here: https://github.com/prebid/prebid.github.io/pull/2783

blingster7 (approved):
LGTM

PR #6443, opened 2021-03-18
UID 2.0 Userid submodule

SKOCHERI:
## Type of change

- [ ] Bugfix
- [ ] Feature
- [ ] New bidder adapter   
- [ ] Build related changes
- [ ] CI related changes
- [ ] Other

## Description of change

```
{
  bidder: '<bidder name>',
  params: {
    // ...
  }
}
```


modules/uid2IdSystem.js
```diff
@@ -0,0 +1,97 @@
...
+  /**
+   * Vendor id of Prebid
+   * @type {Number}
+   */
+  gvlid: GVLID,
+  /**
+   * decode the stored id value for passing to bid requests
+   * @function
+   * @param {string} value
+   * @returns {{sharedid:{ id: string, third:string}} or undefined if value doesn't exists
```
patmmccann, line 62: refs: sharedid, is this on purpose?

modules/uid2IdSystem.js
```diff
@@ -0,0 +1,97 @@
...
+   * performs action to obtain id and return a value.
+   * @function
+   * @param {SubmoduleConfig} [config]
+   * @param {ConsentData|undefined} consentData
+   * @returns {sharedId}
+   */
+  getId(config, consentData) {
+    logInfo('Creating UID2');
+    let value = storage.getCookie(ADVERTISING_COOKIE);
+    logInfo('The advestising token: ' + value);
```
patmmccann, line 78: spelling

modules/uid2IdSystem.js
```diff
@@ -0,0 +1,97 @@
...
+  /**
+   * performs action to obtain id and return a value.
+   * @function
+   * @param {SubmoduleConfig} [config]
+   * @param {ConsentData|undefined} consentData
+   * @returns {sharedId}
+   */
+  getId(config, consentData) {
+    logInfo('Creating UID2');
+    let value = storage.getCookie(ADVERTISING_COOKIE);
```
patmmccann, line 77: does this not work with local storage? this seems like quite an issue

SKOCHERI: UID 2.0 submodule relies on cookie '__uid2_advertising_token' which is set by UID 2.0 SSO

modules/uid2IdSystem.js
```diff
@@ -0,0 +1,97 @@
...
+ * @requires module:modules/userId
+ */
+
+import * as utils from '../src/utils.js'
+import {submodule} from '../src/hook.js';
+import {getStorageManager} from '../src/storageManager.js';
+
+const storage = getStorageManager();
+
+const MODULE_NAME = 'uid20';
```
pm-harshad-mane, line 14: Can we call it `uid2` instead of `uid20`? looks inconsistent with the file name.
Also for LOG_PRE_FIX and decode function return value.

SKOCHERI: Renamed the file.

pm-harshad-mane: Cool

pm-harshad-mane (approved):
LGTM

modules/uid20IdSystem.md
```diff
@@ -0,0 +1,32 @@
...
+
+UID 2.0 ID Module.
+
+### Prebid Params
+
+Individual params may be set for the UID 2.0 Submodule. At least one identifier must be set in the params.
+
+```
+pbjs.setConfig({
+    usersync: {
```
patmmccann, line 11: this is deprecated

modules/uid20IdSystem.md
```diff
@@ -0,0 +1,32 @@
...
+        userIds: [{
+            name: 'uid20',
+            storage: {
+                name: 'uid20id',
+                type: 'html5'
```
patmmccann, line 16: if it has to be a cookie, why is html5 an option here? publishers can copy the cookie into local storage?

modules/uid20IdSystem.md
```diff
@@ -0,0 +1,32 @@
...
+});
+```
+## Parameter Descriptions for the `usersync` Configuration Section
+The below parameters apply only to the UID 2.0 User ID Module integration.
+
+| Param under usersync.userIds[] | Scope | Type | Description | Example |
+| --- | --- | --- | --- | --- |
+| name | Required | String | ID value for the UID20 module - `"uid20"` | `"uid20"` |
+| storage | Required | Object | The publisher must specify the local storage in which to store the results of the call to get the user ID. This can be either cookie or HTML5 storage. | |
+| storage.type | Required | String | This is where the results of the user ID will be stored. The recommended method is `localStorage` by specifying `html5`. | `"html5"` |
```
patmmccann, line 29: this all feels out of sync, is this true given your previous comment about a cookie? This says html5 is recommended yet you are saying the other module is requiring a cookie.

modules/uid20IdSystem.md
```diff
@@ -0,0 +1,32 @@
...
+                type: 'html5'
+            }
+        }]
+    }
+});
+```
+## Parameter Descriptions for the `usersync` Configuration Section
+The below parameters apply only to the UID 2.0 User ID Module integration.
+
+| Param under usersync.userIds[] | Scope | Type | Description | Example |
```
patmmccann, line 25: userSync not usersync

patmmccann:
Generally all my comments are around this question: Is there an opportunity to fix UID 2.0 SSO so it writes  '__uid2_advertising_token'  to html5? If not, a bunch of the doc on this commit needs changing to indicate this is a cookie only module.

Also why write __uid2_advertising_token only to copy it to pbjs.uid20 or whatever. How about we just use one key instead of copying the value around?

modules/uid20IdSystem.md
SKOCHERI: This module allows to use either cookie or html5. The value of the id is read from a specific cookie (__uid2_advertising_token) which is not based out the configuration.

modules/userId/eids.js
```diff
@@ -185,6 +185,13 @@ const USER_IDS_CONFIG = {
     },
     source: 'novatiq.com',
     atype: 1
+  },
+  'uid20': {
+    source: 'uid20.com',
+    atype: 1,
```
patmmccann, line 191: pretty sure atype is 3 for uid2

modules/uid20IdSystem.js
```diff
@@ -0,0 +1,97 @@
...
+  /**
+   * performs action to obtain id and return a value.
+   * @function
+   * @param {SubmoduleConfig} [config]
+   * @param {ConsentData|undefined} consentData
+   * @returns {sharedId}
+   */
+  getId(config, consentData) {
+    logInfo('Creating UID2');
+    let value = storage.getCookie(ADVERTISING_COOKIE);
```
patmmccann, line 77: zeotap did it like this, allowing publishers to not have to use a cookie, can we make this the same? 
function readCookie() {
  return storage.cookiesAreEnabled() ? storage.getCookie(ZEOTAP_COOKIE_NAME) : null;
}

function readFromLocalStorage() {
  return storage.localStorageIsEnabled() ? storage.getDataFromLocalStorage(ZEOTAP_COOKIE_NAME) : null;
}

gtId ... 
 const id = readCookie() || readFromLocalStorage();

patmmccann (changes requested):

modules/uid20IdSystem.js
```diff
@@ -0,0 +1,97 @@
...
+  decode(value) {
+    return (value) ? encodeId(value) : undefined;
+  },
+
+  /**
+   * performs action to obtain id and return a value.
+   * @function
+   * @param {SubmoduleConfig} [config]
+   * @param {ConsentData|undefined} consentData
+   * @returns {sharedId}
```
patmmccann, line 73: uid20

modules/uid20IdSystem.js
SKOCHERI: @patmmccann Thanks for point this out. Will take a look.

modules/uid20IdSystem.js
```diff
@@ -0,0 +1,97 @@
...
+ * Encode the id
+ * @param value
+ * @returns {string|*}
+ */
+function encodeId(value) {
+  const result = {};
+  if (value) {
+    const bidIds = {
+      id: value
+    }
```
smenzer, line 37: what's the purpose of setting this `const` when it's just used once and immediately on the line below?

modules/uid20IdSystem.js
```diff
@@ -0,0 +1,97 @@
...
+   * performs actions even if the id exists and returns a value
+   * @param config
+   * @param consentData
+   * @param storedId
+   * @returns {{callback: *}}
+   */
+  extendId(config, consentData, storedId) {
+    logInfo('Existing id ' + storedId);
+    let value = storage.getCookie(ADVERTISING_COOKIE);
+    return {id: value}
```
smenzer, line 92: what is the purpose of this method? since you're hardcoding the storage of the ID yourself in `ADVERTISING_COOKIE`, publishers do not need to specify a `storage` config, which means `getId` will always be called

integrationExamples/gpt/userId_example.html
```diff
@@ -277,6 +277,14 @@
                         {
                             name: "criteo"
                         },
+                          {
+                            name: "uid20",
+                            storage: {
+                              type: "cookie",
+                              name: "uid",
+                              expires: 28
+                            }
```
smenzer, line 286: you are hardcoding the storage of the id within the module, so there's no reason a publisher should be setting storage since it's not used.

modules/uid20IdSystem.md
smenzer: per my above comments, `storage` config isn't necessary since you're pulling from your own cookie

modules/uid20IdSystem.md
```diff
@@ -0,0 +1,32 @@
...
+## Parameter Descriptions for the `usersync` Configuration Section
+The below parameters apply only to the UID 2.0 User ID Module integration.
+
+| Param under userSync.userIds[] | Scope | Type | Description | Example |
+| --- | --- | --- | --- | --- |
+| name | Required | String | ID value for the UID20 module - `"uid20"` | `"uid20"` |
+| storage | Required | Object | The publisher must specify the local storage in which to store the results of the call to get the user ID. This can be either cookie or HTML5 storage. | |
+| storage.type | Required | String | This is where the results of the user ID will be stored. The recommended method is `localStorage` by specifying `html5`. | `"html5"` |
+| storage.name | Required | String | The name of the cookie or html5 local storage where the user ID will be stored. | `"uid20id"` |
+| storage.expires | Optional | Integer | How long (in days) the user ID information will be stored. | `365` |
```
smenzer, line 31: per my other comments, this can be removed

modules/uid20IdSystem.md
```diff
@@ -0,0 +1,32 @@
...
+| value | Optional | Object | Used only if the page has a separate mechanism for storing the Halo ID. The value is an object containing the values to be sent to the adapters. In this scenario, no URL is called and nothing is added to local storage | `{"uid20": { "id": "eb33b0cb-8d35-4722-b9c0-1a31d4064888"}}` |
```
smenzer, line 32: this probably isn't necessary, but if it is, you should change the `Halo ID` to `UID2`

smenzer (changes requested):
I know this module is "Unified ID v2.0", shortened to "UID2.0" ... but in the code by removing the "dot", it just looks like "UID twenty". can everything just be made "uid2" instead?

otherwise, my main comment is around the `storage` config, which isn't necessary (and in fact, is duplicative) since you're pulling from a hardcoded cookie/local storage name. removing this publisher setup for this would simplify several parts of this module

SKOCHERI:
@patmmccann @smenzer Have updated to read from cookie /local storage and removed the storage config.

patmmccann:
thanks for incorporating all my feedback

patmmccann (approved):

patmmccann:
My comments are all resolved on the code but I don't see the link to the docs repo pr

smenzer (approved):
Just need the docs PR and then I can go ahead and merge this.

jdwieland8282:
dev docs https://github.com/prebid/prebid.github.io/pull/2798

smenzer:
> dev docs [prebid/prebid.github.io#2798](https://github.com/prebid/prebid.github.io/pull/2798)

thanks @jdwieland8282. this LGTM now so I'll go ahead and merge it.

Issue #6430, opened 2021-03-17
Sync pubcommonid or other ids to ad server


patmmccann:
## Type of issue
Feature Request

## Description
Recently at https://blog.google/products/admanager/helping-publishers-thrive-todays-privacy-environment/ Google announced their intention to "enable publishers to further increase their inventory’s value, [by] working to help them expand the use of their PPIDs to more programmatic campaign types, including the Open Auction"

The ppid is described at https://support.google.com/admanager/answer/2880055?hl=en and has historically been used for "frequency capping, audience segmentation and targeting, sequential ad rotation" and analysis on the data transfer files, in which they appear. 

Setting the ppid is rather straightforward, eg `       googletag.pubads().setPublisherProvidedId('12JD92JD8078S8J29SDOAKC0EF230337');
`

I would like the User Id module to sync the pubcommonid or the https://github.com/prebid/Prebid.js/blob/master/modules/pubProvidedIdSystem.js id [and potentially other ids] to gam ppid if the publisher has designated it would like this in config.  The pubcommonid feels like a natural choice for many publishers, as it is simply the id in the first party publisher cookie.

dmdabbs:
Agree, seems like a Good Thing(tm). Would you imagine some site-controlled preference ranking of which to send, or initially more straightforward conf for which of the supported input IDs to send, if present, to PPID?

It is singular, "an identifier."

>This identifier is used alongside any other identifiers that are available to Google Ad Manager, such as cookies. The identifier must be hashed, anonymous, and must not contain any personal information, third-party identifiers or device IDs.

smenzer:
I agree this should be a feature of the userid module where the publisher can select which ID they'd like to use as the google PPID. And as @dmdabbs said, this is only ONE id, so pubs can't just choose all. 

perhaps something in the `userSync` config with a new option `ppidUserIdModule: 'abcId'` where `abcId` is the same value that would be passed in the `userIds[].name` field in the user id module config.

muuki88:
How will id providers ensure the requirements that google seems to have?

> This identifier is used alongside any other identifiers that are available to Google Ad Manager, such as cookies. The identifier must be hashed, anonymous, and must not contain any personal information, third-party identifiers or device IDs.

From [Google PPID docs|https://support.google.com/admanager/answer/2880055?hl=en]

patmmccann:
> How will id providers ensure the requirements that google seems to have?
> 
[...]

PR #6448, opened 2021-03-19
UserID Module: allow userid to ppid sync

patmmccann:
## Type of change

- [ ] Bugfix
- [X] Feature
- [ ] New bidder adapter   
- [ ] Build related changes
- [ ] CI related changes
- [ ] Other

## Description of change
Solution to #6430

### DIFF
--- a/modules/userId/index.js
+++ b/modules/userId/index.js
@@ -824,7 +824,7 @@ export function init(config) {
     // userSync.ppid should be one of the 'source' values in getUserIdsAsEids() eg pubcid.org or id5-sync.com
     if (userSync && userSync.userIds && userSync.ppid && findIndex(getUserIdsAsEids(), x => x.source === userSync.ppid) >= 0 && typeof getUserIdsAsEids().find(x => x.source === userSync.ppid).uids[0].id === 'string') {
       let googletag = window.googletag || {};
-      const ppid = getUserIdsAsEids().find(x => x.source === userSync.ppid).uids[0].id.replace(/\W/g, '');
+      const ppid = getUserIdsAsEids().find(x => x.source === userSync.ppid).uids[0].id.replace(/[\W_]/g, '');
       if (ppid.length >= 32 && ppid.length <= 150) {
         googletag.cmd = googletag.cmd || [];
         googletag.cmd.push(function() {
--- a/modules/userId/index.js
+++ b/modules/userId/index.js
@@ -125,12 +125,12 @@
  */
 
 /**
-  * @typedef {Object} RefreshUserIdsOptions
+  * @typedef {Object} RefreshUserIdsOpnpmjs.com/package/ctions
   * @property {(string[]|undefined)} submoduleNames - submodules to refresh
   */
 
 import find from 'core-js-pure/features/array/find.js';
-import findIndex from 'core-js-pure/features/array/virtual/find-index';
+import findIndex from 'core-js-pure/features/array/find-index';
 import { config } from '../../src/config.js';
 import events from '../../src/events.js';
 import * as utils from '../../src/utils.js';
@@ -821,7 +821,7 @@ export function init(config) {
       auctionDelay = utils.isNumber(userSync.auctionDelay) ? userSync.auctionDelay : NO_AUCTION_DELAY;
       updateSubmodules();
     }
-    //userSync.ppid should be one of the 'source' values in getUserIdsAsEids() eg pubcid.org or id5-sync.com
+    // userSync.ppid should be one of the 'source' values in getUserIdsAsEids() eg pubcid.org or id5-sync.com
     if (userSync && userSync.userIds && userSync.ppid && getUserIdsAsEids().findIndex(x => x.source === userSync.ppid) >= 0 && typeof getUserIdsAsEids().find(x => x.source === userSync.ppid).uids[0].id === 'string') {
       let googletag = window.googletag || {};
       googletag.cmd = googletag.cmd || [];
--- a/modules/userId/index.js
+++ b/modules/userId/index.js
@@ -824,7 +824,7 @@ export function init(config) {
     // userSync.ppid should be one of the 'source' values in getUserIdsAsEids() eg pubcid.org or id5-sync.com
     if (userSync && userSync.userIds && userSync.ppid && findIndex(getUserIdsAsEids(), x => x.source === userSync.ppid) >= 0 && typeof getUserIdsAsEids().find(x => x.source === userSync.ppid).uids[0].id === 'string') {
       let googletag = window.googletag || {};
-      let ppid = getUserIdsAsEids().find(x => x.source === userSync.ppid).uids[0].id.replace(/\W/g, '') || {};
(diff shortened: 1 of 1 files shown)


modules/userId/index.js
```diff
@@ -820,6 +820,13 @@ export function init(config) {
       auctionDelay = utils.isNumber(userSync.auctionDelay) ? userSync.auctionDelay : NO_AUCTION_DELAY;
       updateSubmodules();
     }
+    if (userSync && userSync.userIds && getUserIds().hasOwnProperty(userSync.ppid) && typeof getUserIds()[userSync.ppid] === 'string') {
```
smenzer: this won't work with every id provider, including id5, since the result of the `getUserIds()['id5id']` would be an object not a string.

the `eids` array is more standard so might be a better choice?

modules/userId/index.js
```diff
@@ -820,6 +820,13 @@ export function init(config) {
       auctionDelay = utils.isNumber(userSync.auctionDelay) ? userSync.auctionDelay : NO_AUCTION_DELAY;
       updateSubmodules();
     }
+    if (userSync && userSync.userIds && userSync.ppid && getUserIds().hasOwnProperty(userSync.ppid) && typeof getUserIds()[userSync.ppid] === 'string') {
```
smenzer, line 823: this won't work with every id provider, including id5, since the result of the `getUserIds()['id5id']` would be an object not a string.

the `eids` array is more standard so might be a better choice?

patmmccann: pbjs.getUserIdsAsEids().find(x => x.source === 'pubcid.org').uids[0].id
"876b1f92-a6f9-4639-b939-Ab64eaa9d9"
pbjs.getUserIdsAsEids().find(x => x.source === 'id5-sync.com').uids[0].id
"ID5-ZHMORb_9Ukvpgr1efasddgasdaXF8BU54y7aiyOVKwU0jhu2w"   

seems to work on this obj yeah

smenzer: yea i think that's the better approach to be scalable across how any id provider wants to store the userid object

modules/userId/index.js
```diff
@@ -820,6 +821,17 @@ export function init(config) {
       auctionDelay = utils.isNumber(userSync.auctionDelay) ? userSync.auctionDelay : NO_AUCTION_DELAY;
       updateSubmodules();
     }
+    // userSync.ppid should be one of the 'source' values in getUserIdsAsEids() eg pubcid.org or id5-sync.com
+    if (userSync && userSync.userIds && userSync.ppid && findIndex(getUserIdsAsEids(), x => x.source === userSync.ppid) >= 0 && typeof getUserIdsAsEids().find(x => x.source === userSync.ppid).uids[0].id === 'string') {
+      let googletag = window.googletag || {};
```
muuki88, line 826: Out of curiosity, why not

```js
window.googletag = window.googletag || { cmd: [] };
```

and then later reference via `window.googletag` ? :smiley:

patmmccann: I was just following an example in another commit,  is there an advantage to your suggestion?

muuki88: Initialization is done in a single step, no intermediate variable, explicit referencing the googletag object (no shadowing).

I have been biten by JS so many times that I try to make things as explicit as possible :sweat_smile:

patmmccann: just changed, can you confirm as per your suggestion?

muuki88: LGTM :heart:

muuki88 (approved):

modules/userId/index.js
```diff
@@ -820,6 +821,16 @@ export function init(config) {
       auctionDelay = utils.isNumber(userSync.auctionDelay) ? userSync.auctionDelay : NO_AUCTION_DELAY;
       updateSubmodules();
     }
+    // userSync.ppid should be one of the 'source' values in getUserIdsAsEids() eg pubcid.org or id5-sync.com
+    if (userSync && userSync.userIds && userSync.ppid && findIndex(getUserIdsAsEids(), x => x.source === userSync.ppid) >= 0 && typeof getUserIdsAsEids().find(x => x.source === userSync.ppid).uids[0].id === 'string') {
+      window.googletag = window.googletag || { cmd: [] };
+      const ppid = getUserIdsAsEids().find(x => x.source === userSync.ppid).uids[0].id.replace(/[\W_]/g, '');
+      if (ppid.length >= 32 && ppid.length <= 150) {
```
smenzer, line 828: is this check due to ppid requirements from google?

smenzer (changes requested):
code LGTM

need to add a test for this and a link to docs pr, please

modules/userId/index.js
patmmccann: yes

patmmccann:
@smenzer docs pr is linked above

I am not sure how best to test this, as googletag doesn't have some sort of ppid successfully set functionality i can test for, what do you recommend?

modules/userId/index.js
patmmccann: https://github.com/prebid/prebid.github.io/pull/2788 is the docs pr btw

smenzer:
> I am not sure how best to test this, as googletag doesn't have some sort of ppid successfully set functionality i can test for, what do you recommend?

Could you maybe try to mock this part and then just check that the `ppid` that gets passed in to that method ultimately is the right one (and fails to call it if the ID doesn't meet the requirements)? 

```
window.googletag.cmd.push(function() {
    window.googletag.pubads().setPublisherProvidedId(ppid)
});
```

patmmccann:
I can't seem to restore the branch this was on,  I plan to close this and reopen

ChrisHuie:
@patmmccann you still closing this pr?

patmmccann:
@robertrmartinez checking in to see if you were able to take a look at making the unit test

modules/userId/index.js
```diff
@@ -820,6 +821,16 @@ export function init(config) {
       auctionDelay = utils.isNumber(userSync.auctionDelay) ? userSync.auctionDelay : NO_AUCTION_DELAY;
       updateSubmodules();
     }
+    // userSync.ppid should be one of the 'source' values in getUserIdsAsEids() eg pubcid.org or id5-sync.com
```
FlorentDancy, line 824: Hello, how/where would the publisher define the value for `userSync.ppid`?
Also, my understanding from this comment is that it will leave the choice to the publisher onto which ID from EIDS they choose to use as PPID: do you confirm?

robertrmartinez:
Closing as new PR to address #7681

modules/userId/index.js
robertrmartinez: @FlorentDancy 

The pub will need to set `ppid` inside of their `userSync` setConfig.

Something like so:

```
      config.setConfig({
        userSync: {
          ppid: 'pubcid.org',
          userIds: [
            .. userIds
          ]
        }
      });
```

where the pubcid.org matches the userId submodules associated eids source value defined in `eids.js`

This PR has been closed as another one was open to fix the tests etc.

FlorentDancy: Very clear, thanks @robertrmartinez!

PR #6455, opened 2021-03-22
Rads Bid Adapter : add GDPR support & user sync support

onlsol:
## Type of change
- [ ] Feature

## Description of change
- Add a gdpr support
- Add user sync support

https://github.com/prebid/prebid.github.io/pull/2790

### DIFF
--- a/modules/radsBidAdapter.js
+++ b/modules/radsBidAdapter.js
@@ -169,10 +169,8 @@ function prepareExtraParams(params, payload, bidderRequest) {
 
   if (bidderRequest && bidderRequest.gdprConsent) {
     if (payload.pfilter !== undefined) {
-      if (!payload.pfilter.gdpr_consent) {
-        payload.pfilter.gdpr_consent = bidderRequest.gdprConsent.consentString;
-        payload.pfilter.gdpr = bidderRequest.gdprConsent.gdprApplies;
-      }
+      payload.pfilter.gdpr_consent = bidderRequest.gdprConsent.consentString;
+      payload.pfilter.gdpr = bidderRequest.gdprConsent.gdprApplies;
     } else {
       payload.pfilter = {
         'gdpr_consent': bidderRequest.gdprConsent.consentString,
--- a/modules/radsBidAdapter.js
+++ b/modules/radsBidAdapter.js
@@ -105,7 +105,7 @@ export const spec = {
 
     const syncs = []
 
-    var gdprParams = '';
+    let gdprParams = '';
     if (gdprConsent) {
       if ('gdprApplies' in gdprConsent && typeof gdprConsent.gdprApplies === 'boolean') {
         gdprParams = `gdpr=${Number(gdprConsent.gdprApplies)}&gdpr_consent=${gdprConsent.consentString}`;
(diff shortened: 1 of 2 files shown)


modules/radsBidAdapter.js
```diff
@@ -97,9 +97,46 @@ export const spec = {
...
+    let gdprParams = '';
+    if (gdprConsent) {
+      if ('gdprApplies' in gdprConsent && typeof gdprConsent.gdprApplies === 'boolean') {
+        gdprParams = `gdpr=${Number(gdprConsent.gdprApplies)}&gdpr_consent=${gdprConsent.consentString}`;
+      } else {
+        gdprParams = `gdpr_consent=${gdprConsent.consentString}`;
+      }
+    }
+
+    if (syncOptions.iframeEnabled) {
```
ncolletti, line 117: This conditional should also be checking serverResponses.length. It may also help to check for the existence of 'userSync.[iframeUrl/imageUrl]' in both conditionals.

modules/radsBidAdapter.js
```diff
@@ -125,10 +162,25 @@ function isVideoRequest(bid) {
...
+      if (!payload.pfilter.gdpr_consent) {
+        payload.pfilter.gdpr_consent = bidderRequest.gdprConsent.consentString;
+        payload.pfilter.gdpr = bidderRequest.gdprConsent.gdprApplies;
+      }
+    } else {
+      payload.pfilter = {
+        'gdpr_consent': bidderRequest.gdprConsent.consentString,
+        'gdpr': bidderRequest.gdprConsent.gdprApplies
+      };
+    }
```
ncolletti, line 181: Shouldn't this initialize as a new object with the bidderRequest.gdprConsent values every time prepareExtraParams runs?

modules/radsBidAdapter.js
avj83: Such a check is located above in the line 102 or what do you mean about `checking serverResponses.length`?

modules/radsBidAdapter.js
avj83: this is always initialized, except if an `gdpr_consent` already exists in `pfilter` or an empty `bidderRequest.gdprConsent`

modules/radsBidAdapter.js
ncolletti: If a publisher only has iframeEnabled syncs this is the only conditional that will fire off your usersyncs, you are not checking your serverResponses for length before accessing the array at zero and attempting to set the iframe url.

avj83:
``` getUserSyncs: function(syncOptions, serverResponses, gdprConsent, uspConsent) {
    if (!serverResponses || serverResponses.length === 0) {
      return [];
    }
...
if (syncOptions.iframeEnabled) {
      serverResponses[0].body.userSync.iframeUrl.forEach((url) => syncs.push({
...
```
I check the number of server responses at the beginning of the function, if there is 0 - then the function will return earlier. Is not it?

modules/radsBidAdapter.js
avj83: I have removed this feature from the code.  It was needed for tests

modules/radsBidAdapter.js
ncolletti: Ok that is fine, you are checking for it on line 123 so I was curious why it wasn't checked here too.

ncolletti (approved):
LGTM

Issue #4048, opened 2019-08-01
referenced 'registerAdserver' but it was never created


aszydlo:
## Type of issue
bug/question

## Description
When I add a dfpAdServerVideo to my modules I get ``Warning`` in the console. I don't use this module in my webpage, it is only included into Prebid.js library. 
![blad-prebid](https://user-images.githubusercontent.com/38684131/62294759-ff587b80-b46b-11e9-9514-f5bdf07f0abb.jpg)

## Steps to reproduce
1. Use: gulp serve --modules=consentManagement,rubiconBidAdapter,adomikAnalyticsAdapter,dfpAdServerVideo
2. On the site with active Prebid library above described ``Warning`` is present.

## Expected results
My questions are:
Is it a bug?
How can I handle this issue?

## Platform details
I tested Prebid.js version 2.26.0, 2.25.0 only.

jaiminpanchal27:
@aszydlo Thanks for reporting this, I will check and get back to you.

spormeon:
+1, i get this as well and no idea if its something to worry about?

snapwich:
Just FYI, the warning is because the `registerAdserver` hook is defined in the `categoryTranslation` module (which isn't in your build) but is referenced in other modules such as the `dfpAdServerVideo` that are in your build.

tandalevaibhav:
I am using Prebid 2.34.0 
Getting the same warning as well as on other:- 
"fun-hooks: referenced 'registerAdserver' but it was never created"
"fun-hooks: referenced 'adpod' but it was never created"
Is there anything I should worry about ?

jaiminpanchal27:
@tandalevaibhav We have verified this and not a bug. This warning appears when an optional module is adds a hook on hook point which does not exist. 

In this case, [dfpAdserverVideo here](https://github.com/prebid/Prebid.js/blob/master/modules/dfpAdServerVideo.js#L114) is trying to add hook on [categoryTranslation here](https://github.com/prebid/Prebid.js/blob/master/modules/categoryTranslation.js#L25) but as categoryTranslation module is not included in build this warning is logged.

gglas:
assigning @robertrmartinez

gglas:
@patmmccann do the referenced PRs just relate or do the address the issue?  If they don't address, @robertrmartinez quick ping reminder on this one!

patmmccann:
#6464 fixes the issue,  awaiting merge

gglas:
[...]

PR #6462, opened 2021-03-23
Fix dfp module translation hook

patmmccann:
This would add a requirement that getConfig('brandCategoryTranslation.translationFile') exists for the hook to the translation module to occur, solving https://github.com/prebid/Prebid.js/issues/4048

This is technically a breaking change, as publishers that want this hook to occur from dfp would have to define this config, however I think there are zero affected pubs, because the default translation file is for freewheel, so the it wouldn't make any sense not to have the config.


patmmccann:
rebasing to 5.0 and will reopen

PR #6469, opened 2021-03-24
Rubicon Bid Adapter: add outstream rendering

smozhaiskyi-rubi:
## Type of change
- [x] Feature

## Description of change
Render outstream ads with Rubicon player.

The renderer appearance can be configured with the following parameters, all of them are optional:
```
pbjs.setConfig({
  rubicon: {
    rendererConfig: {
      align: 'center',   // player placement: left|center|right
      position: 'append'   // player position relative to ad unit: append|prepend|before|after
      closeButton: false,   // display 'Close' button
      label: '',   // custom text to display above the player
      collapse: true   // remove the player from the page after ad playback
    }
  }
});
```


Rothalack:
It seems to be working properly for me. The dev docs update should be the only thing left to do. I think the minimum you'd need is to add those new params. Maybe your Outstream section could use an update, but that's up to you.

https://github.com/prebid/prebid.github.io/blob/master/dev-docs/bidders/rubicon.md

mmoschovas (approved):
LGTM

Rothalack (approved):
Just waiting on dev docs update.

bretg (changes requested):
We need to know the open source repo location where https://video-outstream.rubiconproject.com/apex-2.0.0.js lives and can be inspected

Please add a comment?

smozhaiskyi-rubi:
The source code for Apex2 outstream ad unit renderer is available for inspection here: https://github.com/rubicon-project/apex2

bretg:
Thanks @smozhaiskyi-rubi . Now we need someone willing to confirm the external renderer code.

patmmccann:
I think we also need a docs update on how to set these parameters in the bidder parameters doc

modules/rubiconBidAdapter.js
```diff
@@ -783,6 +791,63 @@ function _renderCreative(script, impId) {
...
+  if (el) {
+    el.style.setProperty('display', 'none');
+  }
+}
+
+function hideSmartAdServerIframe(adUnit) {
+  const el = adUnit.querySelector("script[id^='sas_script']");
+  if (el && el.nextSibling && el.nextSibling.localName === 'iframe') {
+    el.nextSibling.style.setProperty('display', 'none');
+  }
```
karimMourra, line 805: for better minification i recommend using a `nextSibling` var
```
const nextSibling = el && el.nextSibling;
if (nextSibling && nextSibling.localName === 'iframe') {
    nextSibling.style.setProperty('display', 'none');
}
```

this should result in fewer chars and reduce file size

smozhaiskyi-rubi: Done.

karimMourra (approved):

bretg (approved):

bretg:
Fixed conflict

bretg:
Docs done. Already merged.

test/spec/modules/rubiconBidAdapter_spec.js
```diff
@@ -3050,6 +3050,153 @@ describe('the rubicon adapter', function () {
...
+            placement: {
+              align: 'left',
+              attachTo: '#outstream_video1_placement',
+              position: 'append',
+            },
+            vastUrl: 'https://test.com/vast.xml',
+            width: 640
+          });
+          // cleanup
+          adUnit.remove();
```
patmmccann, line 3196: this is not ie 11 compatible

robertrmartinez: 🤦

patmmccann: https://github.com/prebid/Prebid.js/blob/804295aa2dae0484d67891b73fcfc401ef8244f1/test/spec/modules/videoNowBidAdapter_spec.js#L6

patmmccann: that polyfill from the other adapter is why your tests passed before. removing it causes this to fail.

robertrmartinez: ah ok, thought this was merged with failing unit tests.

I prefer to fix this by not using a polyfill.

I will put a PR together

robertrmartinez: @patmmccann Should this PR be made in master or 5.0 branch?

Guess it does not matter?

harpere: let's do it in *master*, then *master* should be merged back into *5.0*

patmmccann: works for me

PR #6474, opened 2021-03-25
Pubxai Analytics Adapter: bug fixes and code revamp

pnhegde:
## Type of change

- [x] Bugfix

## Description of change
Bug fixes and code restructuring.

### DIFF
--- a/modules/pubxaiAnalyticsAdapter.js
+++ b/modules/pubxaiAnalyticsAdapter.js
@@ -35,7 +35,7 @@ var pubxaiAnalyticsAdapter = Object.assign(adapter(
         events.floorDetail = {};
         events.bids = [];
         const floorData = utils.deepAccess(args, 'bidderRequests.0.bids.0.floorData');
-        if (typeof args.bidderRequests[0].bids[0] !== 'undefined' && typeof floorData !== 'undefined') {
+        if (typeof floorData !== 'undefined') {
           Object.assign(events.floorDetail, floorData);
         }
         auctionTimestamp = args.timestamp;
@@ -54,7 +54,7 @@ var pubxaiAnalyticsAdapter = Object.assign(adapter(
 });
 
 function mapBidResponse(bidResponse, status) {
-  if (typeof bidResponse !== 'undefined' && bidResponse) {
+  if (typeof bidResponse !== 'undefined') {
     let bid = {
       adUnitCode: bidResponse.adUnitCode,
       auctionId: bidResponse.auctionId,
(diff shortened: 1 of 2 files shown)


pnhegde:
Not able to reproduce test failure in the localhost. It looks fine. Can we rerun the CI?

robertrmartinez:
@pnhegde 

Looks like one of your tests fails in IE.

I will rerun to see if it was flakey.

modules/pubxaiAnalyticsAdapter.js
```diff
@@ -156,8 +169,12 @@ function send(data, status) {
         prebidVersion: $$PREBID_GLOBAL$$.version
       }
     });
-
-    ajax(pubxaiAnalyticsRequestUrl, undefined, JSON.stringify(data), { method: 'POST', contentType: 'text/plain' });
+    if (status == 'bidwon') {
+      ajax(pubxaiAnalyticsRequestUrl, undefined, JSON.stringify(data), { method: 'POST', contentType: 'text/json' });
+    } else if (status == 'auctionEnd' && !auctionCache.includes(data.initOptions.auctionId)) {
```
robertrmartinez, line 174: @pnhegde I found the issue!

IE does not support `Array.includes`!

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/includes

Please change to:

```js
auctionCache.indexOf(data.initOptions.auctionId) === -1
```

robertrmartinez (changes requested):
Update to not use Array includes!

modules/pubxaiAnalyticsAdapter.js
pnhegde: @robertrmartinez Thanks a lot for your inputs. I have made suggested changes and it seem to be working fine.

modules/pubxaiAnalyticsAdapter.js
```diff
@@ -43,81 +47,48 @@ var pubxaiAnalyticsAdapter = Object.assign(adapter(
...
-  return arr;
-}
-
 function mapBidResponse(bidResponse, status) {
-  if (status !== 'bidwon') {
-    let bid = events.bids.filter(o => o.bidId === bidResponse.bidId || o.bidId === bidResponse.requestId)[0];
-    Object.assign(bid, {
-      bidderCode: bidResponse.bidder,
-      bidId: status === 'timeout' ? bidResponse.bidId : bidResponse.requestId,
+  if (typeof bidResponse !== 'undefined' && bidResponse) {
```
robertrmartinez, line 57: Not sure if there is a need to check if type is undefined AND also if the value is truthy.

I think just one of the other is enough.

modules/pubxaiAnalyticsAdapter.js
```diff
@@ -132,6 +103,36 @@ export function getDeviceType() {
...
+  var isIE =
+    /Trident/.test(navigator.userAgent) || /MSIE/.test(navigator.userAgent);
+  var isEdge = /Edge/.test(navigator.userAgent);
+  if (isIE) return 'Internet Explorer';
+  if (isEdge) return 'Microsoft Edge';
+  if (isCriOS) return 'Chrome';
+  if (isSafari) return 'Safari';
+  if (isFirefox) return 'Firefox';
+  if (isChrome) return 'Chrome';
+  return 'Others';
```
robertrmartinez, line 123: Instead of assigning variables for every single combo, I think it would be cleaner to just do a chain of if / else ifs (and maybe sort them in a order of % of users who use which browser even to make it even more optimized!

```js
if (/Chrome/.test(navigator.userAgent) && /Google Inc/.test(navigator.vendor)) return 'Chrome';
else if (navigator.userAgent.match('CriOS')) return 'Chrome';
else if (/Firefox/.test(navigator.userAgent)) return 'Firefox';
else if (/Safari/.test(navigator.userAgent) && /Apple Computer/.test(navigator.vendor)) return 'Safari';
else if (/Edge/.test(navigator.userAgent)) return 'Microsoft Edge';
else if (/Trident/.test(navigator.userAgent) || /MSIE/.test(navigator.userAgent)) return 'Internet Explorer';
else return 'Others';
```

Maybe something like that?

My best guess is Chrome > FF > Safari > Edge > IE

But feel free to order them however you would like.

robertrmartinez (changes requested):
Couple minor adjustments / questions.

modules/pubxaiAnalyticsAdapter.js
pnhegde: This function was taken directly from another adapter and I kept it as it is. Thanks for your inputs, I have updated the code as you suggested.

patmmccann: This seems wrong, order matters,  you have to test for edge before safari for example and your edge test is wrong

https://docs.microsoft.com/en-us/microsoft-edge/web-platform/user-agent-string

pnhegde: @patmmccann I have updated the order so that edge test comes before safari and updated edge token to `Edg` instead of `Edge`. Thanks !

robertrmartinez: @patmmccann 

Sorry, was not aware that order mattered. Which is why I ordered by what I thought was most used browsers!

Makes sense if both have similar UA's

robertrmartinez (approved):
LGTM

patmmccann (approved):
Looks good, it seems like someone should fix at openx should fix openx though

PR #6479, opened 2021-03-26
Kobler Bid Adapter: add new bid adapter

acsbendi:
## Type of change

- [ ] Bugfix
- [ ] Feature
- [x] New bidder adapter
- [ ] Build related changes
- [ ] CI related changes
- [ ] Other

## Description of change
Added Kobler bidder adapter.

```
{
  bidder: 'kobler',
  params: {
        placementId: 'k5H7et3R0'
  }
}
```
In order to see a sample bid from Kobler (without a proper setup), you have to also do the following:
      - Change the [`refererInfo` function](https://github.com/prebid/Prebid.js/blob/master/src/refererDetection.js) to return `'https://www.tv2.no/a/11734615'` as a [`referer`](https://github.com/prebid/Prebid.js/blob/caead3ccccc448e4cd09d074fd9f8833f56fe9b3/src/refererDetection.js#L169). This is necessary because Kobler only bids on recognized articles. 
     - Change the adapter's [`BIDDER_ENDPOINT`](https://github.com/prebid/Prebid.js/blob/master/modules/koblerBidAdapter.js#L8) to `'https://bid-service.dev.essrtb.com/bid/prebid_rtb_call'`. This endpoint belongs to the development server that is set up to always return a bid for the correct `placementId` and page URL combination. 

- contact email of the adapter’s maintainer: bidding-support@kobler.no

Link to a PR on the docs repo: https://github.com/prebid/prebid.github.io/pull/2806


modules/koblerBidAdapter.js
```diff
@@ -0,0 +1,227 @@
...
+function buildOpenRtbBidRequestPayload(validBidRequests, bidderRequest) {
+  const imps = validBidRequests.map(br => buildOpenRtbImpObject(br));
+  const timeout = bidderRequest.timeout || config.getConfig('bidderTimeout') || DEFAULT_TIMEOUT;
+  const pageUrl = (bidderRequest.refererInfo && bidderRequest.refererInfo.referer)
+    ? bidderRequest.refererInfo.referer
+    : window.location.href;
+
+  const request = {
+    id: bidderRequest.auctionId,
+    at: 1, // TODO: is the auction always first price?
```
acsbendi, line 92: As far as I know, Prebid.js is always first price. Is this correct? I am asking because some adapter set `at` to `2` which means the auction is second price.

modules/koblerBidAdapter.js
```diff
@@ -0,0 +1,226 @@
+import * as utils from '../src/utils.js';
+import {config} from '../src/config.js';
+import {registerBidder} from '../src/adapters/bidderFactory.js';
+import {BANNER} from '../src/mediaTypes.js';
+
+const BIDDER_CODE = 'kobler';
+const BIDDER_ENDPOINT = 'https://bid.essrtb.com/bid/prebid_rtb_call';
```
msm0504, line 7: Calling this endpoint from a test page returns a CORS error. My test page is running on `http://localhost:3000`. Do I need to change something in order to test this adapter?

msm0504 (changes requested):

modules/koblerBidAdapter.js
acsbendi: It should return 204 currently. Can you please try again?

msm0504: I'm now getting the 204. However, can this be updated to return a valid response for placement id `xjer0ch8`?

The test params in your .md file should return data.

acsbendi: If there are no bids, we always return 204.

Since Kobler is a contextual platform, we require a page URL that is present in our system in addition to the `placementId`. As a result, it's not possible for any `placementId` we'd give in the .md to return data without also having a matching URL. What do you think we should do in this situation?

msm0504: Is there a way you could set up a test `placementId` that will work with `localhost`?

acsbendi: Definitely, but the code that would run would be completely different than what would run for real traffic (matching `placementId` + page URL). Should we still do this?

modules/koblerBidAdapter.js
patmmccann: Most bidders set this to 1 if the result of the auction within their exchange is first price, notwithstanding the prebid or ad server auction.

acsbendi: I think `at` should be used to indicate the type of the auction in Prebid since this bid request is sent from Prebid to the bidder. Perhaps the bidders that set it to 2 do so because the auction within their exchange is second price?

patmmccann: That is my understanding as well on the behavior of other bidders.  

Prebid doesn't have built in functionality for a second price auction.

acsbendi: Okay, thanks for clearing this up.

modules/koblerBidAdapter.js
msm0504: Sorry, I'm not understanding why this would need to be completely different code. It's just in this case the URL would be `http://localhost`.

acsbendi: It's because of how mediums and their placements are set up in our system. The requirement for us to bid on a URL is that it's crawled (we know its content), which is obviously not possible for localhost.

The way I tested this adapter is as follows:
- change the `BIDDER_ENDPOINT` to `'https://bid-service.dev.essrtb.com/bid/prebid_rtb_call'` (this is our DEV server which has a medium set up to use Prebid)
- change `refererInfo` to return `'https://www.tv2.no/a/11734615'` as a `referer` (this is the URL of a page belonging to that medium).
- use `ku9i4as30` as a `placementId` and then you should see bids (from a test campaign set up in our DEV environment).

msm0504: I'll try this. Can you add this info to your .md file?

msm0504: I'm able to a response from the dev server. Noticed that it's not gzipped. Are the ones from the prod server gzipped? If not, they should be.

acsbendi: I've added the info in the .md file and fixed the missing gzip encoding now. Dev and prod server code is the same, only the data they use is different.

msm0504 (approved):

PR #6484, opened 2021-03-26
add support for non-purpose1 consent domains to appnexus and PBS bid adapters

jsnellbaker:
**NOTE - PLEASE DO REVIEWS BUT DON'T MERGE AUTOMATICALLY.  I WILL MERGE WHEN TIME IS PROPER**

## Type of change

- [x] Feature

## Description of change

As part of an upcoming regulatory change in France, requests to appnexus hosted servers need to have a way to call a different domain that does not support cookies when the purpose1 consent choice is not given.  

### Brief summary of changes to respective files:
* appnexusBidAdapter
  - change mappings.json host domain
  - change regex for native script to include new domain
  - add logic to switch adserver domain based on consent choice
  - add extra check for userSync to ensure request is made when purpose1 consent is given (this is in case when gdprEnforcement module is not included)
* prebidServerBidAdapter
  - add support for a change to the `endpoint` and `syncEndpoint` fields in s2sConfig, to have them use an object value rather than string values.  This has been incorporated to the `defaultVendor` config setting.  See further below for example...
  - there is logic to handle the old string format by converting it to the object format; this was added to not make this a **breaking change**.  The function will set the respective URL to be used in **both** p1 and non-p1 consent scenarios.  
  - To properly use the new URLs that are compliant with non-purpose1 choices, the URL either has to be specifically set in the new format in the s2sConfig, or rely on the defaultVendor feature to set it automatically.  
  See notes below for further details on this point...
* Prebid Cache URL
  - this is more for the docs PR, but wanted to highlight it here.  The URL to the appnexus prebid cache https://prebid.adnxs.com/pbc/v1/cache should be changed to: https://prebid.adnxs-simple.com/pbc/v1/cache

### More notes about s2sConfig and using the non-p1consent URL(s):
There are 3 typical ways that the config can be setup to specify the endpoint and syncEndpoint properties:
  ```
// both settings are used
s2sConfig : {
            accountId : '1',
            defaultVendor: 'appnexus',
            bidders : ['appnexus'],
            endpoint : 'https://prebid.adnxs.com/pbs/v1/openrtb2/auction'
          }
          
// only endpoint    
s2sConfig : {
            accountId : '1',
            enabled : true, //default value set to false
            bidders : ['appnexus'],
            timeout : 1000, //default value is 1000
            adapter : 'prebidServer', //if we have any other s2s adapter, default value is s2s
            endpoint : 'https://prebid.adnxs.com/pbs/v1/openrtb2/auction'
          }          

// only defaultVendor
s2sConfig : {
            accountId : '1',
            defaultVendor: 'appnexus',
            bidders : ['appnexus'],
          }
```

Given the current logic that settings specified in the s2sConfig directly are preferred/overridden by settings from the defaultVendor, these configs will operate accordingly with the new logic:
* both scenario > the `endpoint` will be preferred and used in both p1 and non-p1 cases.   **Publisher should make a change here to be compliant**; either remove `endpoint` property entirely or update its value
* only `endpoint` > same as above.
* only `defaultVendor` > will automatically switch for the publisher based on consent. **No change should be needed by publisher**

Note - if `syncEndpoint` wasn't defined (as above), it will (or will not) be added based on the presence of the `defaultVendor` setting.  If `defaultVendor` is included, the that URL will properly switch according to consent choices.

Below are the sample configs updated to the new format:
```
// both settings are used
s2sConfig : {
            accountId : '1',
            defaultVendor: 'appnexus',
            bidders : ['appnexus'],
            endpoint : {
            	p1Consent: 'https://prebid.adnxs.com/pbs/v1/openrtb2/auction'
            	noP1Consent: 'https://prebid.adnxs-simple.com/pbs/v1/openrtb2/auction'
            },
            // added this here to show formatting
            syncEndpoint: {
            	p1Consent: 'https://prebid.adnxs.com/pbs/v1/cookie_sync',
            	noP1Consent: 'https://prebid.adnxs-simple.com/pbs/v1/cookie_sync'
            }
          }
          
// only endpoint    
s2sConfig : {
            accountId : '1',
            enabled : true, //default value set to false
            bidders : ['appnexus'],
            timeout : 1000, //default value is 1000
            adapter : 'prebidServer', //if we have any other s2s adapter, default value is s2s
            endpoint : {
            	p1Consent: 'https://prebid.adnxs.com/pbs/v1/openrtb2/auction'
            	noP1Consent: 'https://prebid.adnxs-simple.com/pbs/v1/openrtb2/auction'
            }
          }          

// only defaultVendor
s2sConfig : {
            accountId : '1',
            defaultVendor: 'appnexus',
            bidders : ['appnexus'],
          }
```

### CODE REVIEW

**Review by bretg — APPROVED** [review]
Didn't review the code in detail, but the approach LGTM

**Review by Fawke — APPROVED** [review]

File: modules/appnexusBidAdapter.js [hunk 1]
```diff
@@ -326,8 +327,8 @@ export const spec = {
     }
   },
 
-  getUserSyncs: function (syncOptions) {
-    if (syncOptions.iframeEnabled) {
+  getUserSyncs: function (syncOptions, responses, gdprConsent) {
```
mmoschovas, line 330: @jsnellbaker This looks good. One question on this piece though. Should this work the same for both client and s2s. It seems to me that the PBS Bid Adapter runs doClientSideSyncs() which in turn calls registerSyncs([]) not passing in any GDPR data when GDPR exists on the bidderRequests. I just want to understand if this is expected or not. The reason being that testing the s2s, the following always returns true since the gdprConsent value is undefined and therefore, the sync iframe runs.

`if (syncOptions.iframeEnabled && hasPurpose1Consent({gdprConsent}))`

jsnellbaker: @mmoschovas Thanks for raising this concern.  

I did some digging and found that the behavior within the `doClientSideSyncs` function wasn't properly working with the gdprEnforcement module.  It was consistently returning this warning: `WARNING: User sync not allowed for null` because we were not properly permitting the bidder to run this code within the special control logic that was added to the `src/config.js` (see `runWithBidder`).

I'm about to push a commit that should address the lack of consent data on this registerSync call and grant the permissions.   Please take a look

CC @snapwich to take a look at the change, to ensure it's formatted properly with the config bidder permission logic (which I believe he helped design).

snapwich: @jsnellbaker your update looks good and it does look like the server -> client syncs call not passing consent data was an oversight (possibly was added before we had standardized on the user sync call). good callout @mmoschovas

mmoschovas (approved):
LGTM

Issue #4810, opened 2020-02-03
Remove mediaQuery declaration as an empty string in sizeConfig


Fawke:
## Type of issue
Bug

## Description
Under the current design (as of v3.x), a publisher can declare **mediaQuery** string as empty in global sizeConfig object.

The implication of this is inconsistent behaviour when a publisher chooses to declare mediaQuery as an empty string, the current implementation may choose wrong Ad size for a given viewport size. For details and an example, click [here](https://github.com/prebid/Prebid.js/pull/4691#issuecomment-571538911).

Related - https://github.com/prebid/Prebid.js/pull/4691

A fix could be to completely block empty mediaQuery string declaration. Since this solution could be a breaking change for some publishers who're already using empty mediaQuery declaration in sizeConfig, the change should go in the next major release.

gglas:
@Fawke hey Neelanjen -- did 4691 resolve this issue and we can close it out?

Fawke:
@gglas Hi Gareth -  No, we didn't resolve this issue in https://github.com/prebid/Prebid.js/pull/4691. It was a breaking change, so @mkendall07  suggested the change should go out as part of a next major release. It was supposed to be 4.0, but that didn't happen. We can include in now, in 5.0. It's a small edge case in the core sizesMapping module which can lead to some unexpected behaviour.

patmmccann:
Closing with the merge to 5

PR #6487, opened 2021-03-29
Core PBjs : add check for empty mediaquery declaration in sizeConfig object

Fawke:
## Type of change
- [X] Bugfix

## Description of change

We have allowed `mediaQuery` property in **sizeConfig** to be declared as an empty string. This can lead to some inconsistencies as described [here](https://github.com/prebid/Prebid.js/pull/4691#issuecomment-571538911)

This can be a breaking change for some publishers who maybe using empty mediaQuery declarations in their sizeConfigs. This PR should go out in a major release of Prebid.js

Fixes https://github.com/prebid/Prebid.js/issues/4810

Please refer to https://github.com/prebid/Prebid.js/pull/4691 for a detailed context.

### DIFF
--- a/src/sizeMapping.js
+++ b/src/sizeMapping.js
@@ -121,22 +121,17 @@ function evaluateSizeConfig(configs) {
   return configs.reduce((results, config) => {
     if (
       typeof config === 'object' &&
-      typeof config.mediaQuery === 'string'
+      typeof config.mediaQuery === 'string' &&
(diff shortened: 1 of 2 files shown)


patmmccann:
@Fawke can you make this against prebid-5.0 branch?

Fawke:
> @Fawke can you make this against prebid-5.0 branch?

Closing this PR in creating this one that points to the 5.0 branch.

PR #6253, opened 2021-02-01
Update ttl outstream

stevealliance:
## Type of change

- [ x] Bugfix
- [ ] Feature
- [ ] New bidder adapter   
- [ ] Code style update (formatting, local variables)
- [ ] Refactoring (no functional changes, no api changes)
- [ ] Build related changes
- [ ] CI related changes
- [ ] Does this change affect user-facing APIs or examples documented on http://prebid.org?
- [ ] Other

## Description of change
Condition to remove ttl change in vast validation condition

PR #6491, opened 2021-03-29
PR_REVIEW: added check for bidder name validity

bretg:
## Type of change
Documentation

## Description of change
Updating the reviewer reference to check the validity of the proposed bidder code:

- The bidder code must be lower case alphanumeric. The only special character allowed is underscore.
- The bidder code must be unique - make sure none of the other bid adapters is using the same code.
- The bidder code should be unique for the first 6 characters - this consideration helps with generating unique targeting keys for use by some ad exchanges, such as Google Ad Manager.
- There are several reserved words that cannot be used as bidder names: all, context, data, general, prebid, and skadn.

### DIFF
--- a/PR_REVIEW.md
+++ b/PR_REVIEW.md
@@ -38,7 +38,7 @@ General gulp commands include separate commands for serving the codebase on a bu
 Documentation they're supposed to be following is https://docs.prebid.org/dev-docs/bidder-adaptor.html
 
 Follow steps above for general review process. In addition, please verify the following:
-- Verify the biddercode is valid:
+- Verify the biddercode and aliases are valid:
     - Lower case alphanumeric with the only special character allowed is underscore.
     - The bidder code should be unique for the first 6 characters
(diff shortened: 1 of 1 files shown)


patmmccann (approved):

patmmccann:
I think this should potentially extend to aliases. Of note, https://github.com/prebid/Prebid.js/pull/6253/files 'districtmdmx' added a shortcode alias so that keys wouldn't collide with their other adapter 'districtm'

patmmccann (changes requested):
Should we change to "bidder code and all aliases" ?

bretg:
> Should we change to "bidder code and all aliases" ?

changed

patmmccann (approved):

patmmccann:
@gglas can we merge this before this week's release?

gglas (commented):
LGTM

gglas (commented):
approved

PR #6500, opened 2021-03-30
Mass Module: support multiple custom configs for dealId and rendering

cciocov:
## Type of change

- [ ] Bugfix
- [x] Feature
- [ ] New bidder adapter   
- [ ] Build related changes
- [ ] CI related changes
- [ ] Other

## Description of change
Updated the MASS module to support multiple parallel dealId <-> renderer configurations.

### DIFF
--- a/test/spec/modules/mass_spec.js
+++ b/test/spec/modules/mass_spec.js
@@ -3,6 +3,10 @@ import {
   init,
   addBidResponseHook,
   addListenerOnce,
+  isMassBid,
+  useDefaultMatch,
+  useDefaultRender,
+  updateRenderers,
   listenerAdded,
   isEnabled
 } from 'modules/mass';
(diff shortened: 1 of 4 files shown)


patmmccann:
fyi @Fawke this is the followup commit from https://github.com/prebid/Prebid.js/pull/6332 ; the discussion there you may find interesting. The idea is that this module can be adapted to link renderers to deal ids if this use case arises in the future without each company looking for this functionality requiring a deal rendering module.

patmmccann:
Tagging as needing two reviews because this commit converts a vendor specific module into functionality which is generic to many use cases

patmmccann:
Hi @cciocov your docs are not linked from the js modules directory https://docs.prebid.org/dev-docs/modules/index.html#general-modules . Please add to general modules table with description something like 'Can be used to attach rendering logic to deals, eg MASS deals.'

massadmin:
@patmmccann In the latest PR in the doc repo we added a section in the general-modules page.
See here: [](https://github.com/prebid/prebid.github.io/pull/2807)

Fawke (changes requested):
Hi @cciocov,

This change looks good. But you need to add more unit tests to achieve at least 80% coverage.

To check your coverage, please run these commands:

`gulp test-coverage` followed by `gulp view-coverage` and then navigate to your folder location.

Fawke:
Adding @FilipStamenkovic for a second review since he already reviewed this previously. If busy, please feel free to re-assign.

cciocov:
Hi @Fawke 

Just added more tests, thanks.

test/spec/modules/mass_spec.js
```diff
@@ -113,4 +117,35 @@ describe('MASS Module', function() {
...
+
+  it('should match bids by deal ID with the default matcher', function() {
+    const match = useDefaultMatch(/abc/);
+
+    expect(match({dealId: 'abc'})).to.equal(true);
+    expect(match({dealId: 'xyz'})).to.equal(false);
+  });
+
+  it('should have a default renderer', function() {
+    const render = useDefaultRender('http://example.com/render.js', 'abc');
```
patmmccann, line 145: Can you make this secure, and in the above place as well? Even though it isn't used,  we occasionally grep the repo for nonsecure urls to ensure secure everywhere compliance

cciocov: @patmmccann it's done

FilipStamenkovic (approved):
LGTM.  
But I would still like to see working example in integration pages, with some bidder with test params, where bidder is bidding with appropriate `mass` flags and then this module does its magic.  

If not possible in this PR, then in some upcoming PR it would be much appreciated.

massadmin:
@FilipStamenkovic Thank you! Yes, we will work with the respective providers to have hopefully something arranged.

Fawke (approved):
LGTM

PR #6504, opened 2021-03-31
Tappx Bid Adapter: minor updates and universal id added

prebidtappx:
## Type of change

- [X] Bugfix
- [X] Feature

## Description of change

- Minimal adapter updates:
- - An adapter bug related sending the parameters in the buildRequests function has been resolved.
- - Version adapter added
- - Info about requests domain updated
- The Universal ID has been added in the requests.


modules/tappxBidAdapter.js
```diff
@@ -253,6 +264,25 @@ function buildOneRequest(validBidRequests, bidderRequest) {
...
+
+  let eids = [];
+  eidsArr.forEach(eidsElement => {
+    if (eidsElement.source && eidsElement.uids[0].id) {
+      eids.push({
+        source: eidsElement.source,
+        userId: eidsElement.uids[0].id
+      });
+    }
+  });
```
smenzer, line 279: what is the purpose of this block? the `eids` array is not used anywhere else (on line 283 you are setting `user.ext.eids = eidsArr`).

I would suggest passing the entire `eidsArr` as you have the code below, since some ID providers include extensions to their ID that could be useful to you or your downstream DSP partners, so passing them as defined in the `userIdAsEids` array is recommended).

modules/tappxBidAdapter.js
```diff
@@ -253,6 +264,25 @@ function buildOneRequest(validBidRequests, bidderRequest) {
...
+
+  let user = {};
+  user.ext = {};
+  user.ext.eids = eidsArr;
+  payload.user = user;
```
smenzer, line 284: this could be written in a single line like this:

```suggestion
  payload.user = {
    ext: {
      eids: eidsArr
    }
  };
```

smenzer (changes requested):
make sure you don't include the packgage-lock.json in your PR, please.

also, I would suggest adding a test to ensure that the `payload.user` gets set properly with values from the eids array.

modules/tappxBidAdapter.js
prebidtappx: Initially we thought of adding the `eids` array only with the **source** and **uids** but we've not implemented it like that. We left the code by mistake. As you said, it has been removed and we directly use eidsArr. Commit: 37d9f96

modules/tappxBidAdapter.js
prebidtappx: Updated in commit 37d9f96

prebidtappx:
First of all, thank you for your comments @smenzer 
About the "packgage-lock.json", should I create a new pr or do I update it the same way it is in your master branch?

Thank you

ChrisHuie:
I would just untrack the package-lock file in this pr or you can update it to match the master branch.

prebidtappx:
I've just updated for matching the package-lock.json from master.

modules/tappxBidAdapter.js
smenzer: sounds good!

ChrisHuie (approved):

patmmccann:
Comment says add uid, but code seems to pass all id's? Do you drop the rest server side? Can you clarify with a docs pr?

ChrisHuie:
@prebidtappx 👆 question and if you could please update docs before we merge 🙏

prebidtappx:
Hello @patmmccann @ChrisHuie 
we pass all the id's and it's managed on the server side
We've updated our docs in `https://github.com/prebid/prebid.github.io/pull/2829`
Thank you

PR #6293, opened 2021-02-10
Fpd 2.0 Update

mmoschovas:
## Type of change

- [ ] Bugfix
- [ ] Feature
- [ ] New bidder adapter   
- [ ] Code style update (formatting, local variables)
- [X ] Refactoring (no functional changes, no api changes)
- [ ] Build related changes
- [ ] CI related changes
- [ ] Does this change affect user-facing APIs or examples documented on http://prebid.org?
- [ ] Other

## Description of change

DETAIL:
Prebid's First Party Data (FPD) convention revision with backwards compatibility. 

MODULES UPDATED:
gptPreAuction, prebidServer, admixer, adrelevantis, amx, avocet, criteo, grid, haloRtd, imar, jwPlayerRtd, luponmedia, rubicon, smaato, triplelift

BACKWARD COMPATIBILITY GLOBAL/BIDDER:
fpd.context.ATTR --> ortb2.site.ATTR
fpd.context.data.ATTR --> ortb2.site.ext.data
fpd.user.ATTR --> ortb2.user.ATTR
fpd.user.data.ATTR --> ortb2.user.ext.data. (for fpd only... ortb2 can specify user.data)

BACKWARD COMPATIBILITY ADUNIT:
AdUnit.fpd - > AdUnit.ortb2Imp
AdUnit.fpd.context.ATTR -> AdUnit.ortb2Imp.ext.data.ATTR

ISSUE:
https://github.com/prebid/Prebid.js/issues/5795

- test parameters for validating bids
```
{
  bidder: '<bidder name>',
  params: {
    // ...
  }
}
```

Be sure to test the integration with your adserver using the [Hello World](/integrationExamples/gpt/hello_world.html) sample page.

- contact email of the adapter’s maintainer
- [ ] official adapter submission

For any changes that affect user-facing APIs or example code documented on http://prebid.org, please provide:

- A link to a PR on the docs repo at https://github.com/prebid/prebid.github.io/

Issue #6099, opened 2020-12-08
First Party Data Module


bretg:
## Type of issue
Module proposal

## Description
If the First Party Data (FPD) revision in https://github.com/prebid/Prebid.js/issues/5795 is approved, we propose adding modules that perform a number of FPD additions and validations.

In order to provide a consistent set of first party data, we propose two new modules that performs the following general functions:

- verify OpenRTB datatypes, remove/warn any that are likely to choke downstream readers
- verify that certain OpenRTB attributes are not specified: just imp for now
- optionally suppress user FPD based on a TBD opt-out signal (_pubcid_optout)
- populate available data into object: referer, meta-keywords, cur

### Enrichment Actions:

1) merge HTML meta-keywords tag into global ortb2.site.keywords. Datablocks gives an example for how to do this:
```
datablocksBidAdapter.js:    let keywords = document.getElementsByTagName('meta')['keywords'];
```
2) merge page referrer into global ortb2.site.ref from getRefererInfo().referer, obsoleting the need to read the bidRequest.refererInfo. Existing values take precedence.

3) merge page into global ortb2.site.page from getRefererInfo().canonicalUrl. Existing values take precedence.

4) parse domain out of getRefererInfo().canonicalUrl and merge into global ortb2.site.domain. Existing values take precedence.

5) set ortb2.device.h and ortb2.device.w from the physical device's height and width if they're not already specified

6) REMOVED FROM INITIAL RELEASE - set ortb2.cur from getConfig({currency.adServerCurrency}) if it's not already specified

### Validation Actions

7) Parse ortb2 and verify there aren't any imp[] objects specified. If there are, remove them and log a warning.

8) Parse the ortb2.site.content.data and ortb2.user.data arrays for validation. This should include the global ortb2 as well as ortb2 specified in setBidderConfig.
- 'data' must be an array
- loop through 'data' array
- there must be a 'name' attr under each element of data
- if there's an 'ext' attr, it must be an object // maybe more validations here?
- there must be a 'segment' attr under each element of data and it must be an array
- loop through 'segment' array: each object must have an 'id' attribute with a string value.

9) No validations are done on AdUnit-specific FPD.

10) The following OpenRTB field validations should be performed on global and per-bidder ortb2. If any of them fails, remove the field from the ortb2 data and log a console error.
- site.{name,domain,page,ref,keywords,search} must be strings
- site.{cat, sectioncat, pagecat} must be arrays of strings
- site.{content,publisher} must be an objects
[...]

Issue #5795, opened 2020-09-28
First Party Data Revision


bretg:
## Type of issue
Improvement

## Description

The Taxonomy Taskforce has proposed a revision of Prebid's First Party Data (FPD) conventions for the following reasons:

1) The mappings between PBJS conventions and OpenRTB are harder to follow than necessary.
2) There's a mismatch between PBJS and PBS:
    1) The [PBJS First Party Data spec](https://docs.prebid.org/features/firstPartyData.html) says that publishers should put standard OpenRTB attributes in `context` and arbitrary FPD into into `context.data`
    2) This causes data to be sent to Prebid Server in `site` and `site.data`.
    3) However, Prebid Server is expecting to receive arbitrary values at `site.**ext**.data`
3) Context -vs- Content is confusing.
4) The original documentation showed user.data as an object when it needs to be an array.

Since there are only ~10 bidders currently using the FPD conventions, we propose migrating FPD to a pure-OpenRTB format.

## Proposal

1) Going forward, we recommend everything under setConfig({ortb2: OBJECT}) just becomes the OpenRTB representation. This allows the pbsBidAdapter and supporting bidders to simply do a JSON merge of the ortb2 object into the final request. It also allows us to get rid of one-off docs like site.content.language and app support.

- site object
    - site.ATTR - standard OpenRTB attributes (e.g. keywords)
    - site.ext.data -- arbitrary publisher attributes (e.g. editor_rating=4)
    - site.content.data[]  -- content 'segments'
- app object (if post-bid)
- user object
   - user.ATTR - standard OpenRTB attributes (e.g. gender)
   - user.ext.data -- arbitrary publisher attributes (e.g. job_cat=sales)
   - user.data[] - user 'segments'

2) Ad-Unit Specific FPD is likewise placed under 'ortb2Imp' in the AdUnit. It will get merged into imp[]. We recommend moving "context.data" to "ext.data".

2) We create a module to validate FPD:
    - verify OpenRTB datatypes, remove/warn any that are likely to choke downstream readers
    - verify that certain OpenRTB attributes are not specified: just imp for now
    - optionally suppress user FPD based on a TBD opt-out signal
    - populate available data into object: referer, meta-keywords, cur

3) Data permissioning stays the same: use of the`setBidderConfig()` function.

## Work Items

1) The setConfig and setBidderConfig functions support a transition period where they map the original 'fpd' config:
    - fpd.context.ATTR --> ortb2.site.ATTR
[...]

Issue #6057, opened 2020-12-01
Proposal for Taxonomy Segments in FPD


sfrancolla:
## Type of issue
Improvement

## Description
This is a request for comment prior to next meeting (Dec 8). 

The Taxonomy Taskforce has been presented with the below proposal from the taskforce's Prebid-IAB Tech Lab subcommittee. The subcommittee was created to zero-in on an agreeable structure for context and audience taxonomy segments and bring it back to the larger group. 

Objectives were to support publisher, proprietary, and standardized segments with a simple structure that avoids significant repetitive text.

## Additional Background
* IAB Working Doc: 
  * [Section: Passing Audience Taxonomy IDs and DTS Metadata](https://docs.google.com/document/d/1rvc7j9EzoJ8VzvjoZjWJ3GEp83JQ-kLxBmRkK3epurE/edit#heading=h.l1gru012asq1) (Prebid members are invited to request access)

## Proposal
With:
* **ext.segtax**
  * [Optional] The ID for a taxonomy that is centrally registered. At present, it's assumed that the IAB would manage the registrations. Example taxonomy registry is hashed out underneath the json.
* **segment[].id**
  * [Required] The ID of the segment that may or may not have a `segtax` affiliation. I.e. it’s fully proprietary when no `ext.segtax` is provided.

The proposed structure:
```
  "user": {
      "data": [
        {
          "name": "www.dataprovider1.com",
          "ext": {
            "segtax": 3
          },
          "segment": [
            { "id": "1001" }, 
            { "id": "1002" }
          ]
        },
        {
          "name": "www.dataprovider1.com",
          "ext": {
            "segtax": 501
          },
          "segment": [
            { "id": "123" }, 
            { "id": "456" }
          ]
        }
      ]
    },
"site": {
  "content": {
      "data": [
        {
          "name": "www.dataprovider1.com",
          "ext": {
            "segtax": 2
          },
          "segment": [
            { "id": "2001" }, 
            { "id": "2002" }
          ]
        },
        {
          "name": "www.dataprovider1.com",
          "ext": {
            "segtax": 502
          },
          "segment": [
            { "id": "123" }, 
            { "id": "456" }
          ]
        }
      ]
    }
}
```

**Example Taxonomy Registry (Managed by IAB Tech Lab):**

| Taxonomy ID  | Taxonomy Type | Version | Description | Link |
| ------------- | ------------- | ------------- | ------------- | ------------- |
[...]

PR #6505, opened 2021-03-31
Halo RTD Module: FPD 2.0 Updates & add ID system tests

anthonylauzon:
## Type of change
Feature

## Description of change
This changeset updates our RTD module to allow ORTB2 data to be retrieved and merged with the setConfig ORTB2 data. Examples of this type of data include IAB taxonomy segments.  The changeset additionally simplifies how publishers can use Halo real-time data in other use cases and adds halo ID system unit tests.

## Contact
ant@audigent.com

## Other information

Fpd 2.0 Update https://github.com/prebid/Prebid.js/pull/6293
First Party Data Module https://github.com/prebid/Prebid.js/issues/6099
First Party Data Revision https://github.com/prebid/Prebid.js/issues/5795


anthonylauzon:
@bretg @patmmccann  this changeset should be ready to go, but would appreciate any feedback as usual.  looking forward to an e2e test integration with the magnite adapter and one of our pubs once this is cleared.

bretg:
@anthonylauzon - I looked through the code and examples some and am not understanding how this will work for the POC. Can you post what the config will look like?

I had been expecting to see "taxonomyname" in the code -- assumed that your endpoint would return a list of segments and the module would create a user.data array entry, read in what's in ortb2, add the new entry, the write back ortb2.

Instead, the test cases are doing something not very useful-seeming. And something I wouldn't recommend in production -- the example rtdHandler is doing a setconfig on ortb2 without first doing a getconfig and merge.

anthonylauzon:
@bretg the first test case is the POC test case, and it only requires a publisher id: 

`    realTimeData: {
        auctionDelay: auctionDelay,
        dataProviders: [
            {
                name: "halo",
                waitForIt: true,
                params: {
                    segmentCache: false,
                    requestParams: {
                        publisherId: 1234
                    }
                }
            }
        ]
    }`

Our backends will return ortb2 compatible data and merge it into any other data present in setConfig.  For the POC, this will be IAB taxonomy segments as you can see in the first test case, which we use different data provider labels for just for the purposes of testing.  We won't be merging segments based on taxonomy name, since taxonomies are differentiated by data provider.  This was discussed during one of the POC meetings.

The handleRtd function's test case is just a test and not something that you would see in production.  A better example of what handleRtd might look like in production is in the documentation.  It is provided as both a backwards compatibility layer for RTD ad-unit level logic and to allow other types of integrations outside of the scope of the POC.  This changeset continues to allow for existing ad-unit and bid adapter specific logic which while also enabling IAB segment poc testing.

bretg:
Ok, having the formatting on your server-side may make sense given that there may be some syntax changes coming.  I'll let the POC team work through any outstanding details here.

patmmccann:
> Our backends will return ortb2 compatible data and merge it into any other data present in setConfig.

This makes sense to me as well, ty

anthonylauzon:
https://github.com/prebid/Prebid.js/issues/6057 potential format changes are handled here

robertrmartinez:
Sorry for the delay.

I will do this review today!

anthonylauzon:
> Sorry for the delay.
> 
> I will do this review today!

thank you & no problem @robertrmartinez

modules/haloRtdProvider.js
```diff
@@ -1,109 +1,102 @@
...
+  if (!isPlainObject(config.getConfig().ortb2)) {
+    config.setConfig({ortb2: {}});
+  }
 
-  return adUnits;
+  if (rtdConfig.params && rtdConfig.params.handleRtd) {
+    rtdConfig.params.handleRtd(bidConfig, rtd, rtdConfig, config);
+  } else if (rtd.ortb2) {
+    let mergedOrtb = mergeLazy(config.getConfig().ortb2, rtd.ortb2);
+    config.setConfig({ortb2: mergedOrtb});
```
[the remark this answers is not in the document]
robertrmartinez: Don't think it makes sense to setConfig on a empty ortb2 object like line 73

modules/haloRtdProvider.js
```diff
@@ -1,109 +1,102 @@
...
-  const adUnits = reqBidsConfigObj.adUnits || getGlobal().adUnits;
-
-  if (config.params.segmentCache) {
-    let jsonData = storage.getDataFromLocalStorage(SEG_LOCAL_NAME);
+export function getRealTimeData(bidConfig, onDone, rtdConfig, userConsent) {
+  if (isPlainObject(rtdConfig.params) && rtdConfig.params.segmentCache) {
+    let jsonData = storage.getDataFromLocalStorage(RTD_LOCAL_NAME);
 
     if (jsonData) {
       let data = JSON.parse(jsonData);
```
robertrmartinez, line 96: Maybe surround in try catch just in case? Outside of this PR change so feel free to ignore

robertrmartinez (changes requested):
couple questions / comments

modules/haloIdSystem.js
```diff
@@ -5,12 +5,15 @@
...
 
-import * as utils from '../src/utils.js';
 import {ajax} from '../src/ajax.js';
+import {getStorageManager} from '../src/storageManager.js';
 import {submodule} from '../src/hook.js';
+import * as utils from '../src/utils.js';
 
 const MODULE_NAME = 'haloId';
 
+export const storage = getStorageManager();
```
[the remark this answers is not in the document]
anthonylauzon: added to both modules

modules/haloRtdProvider.js
```diff
@@ -1,109 +1,102 @@
...
+import {config} from '../src/config.js';
 import {getGlobal} from '../src/prebidGlobal.js';
-import * as utils from '../src/utils.js';
+import {getStorageManager} from '../src/storageManager.js';
 import {submodule} from '../src/hook.js';
-import {ajax} from '../src/ajax.js';
-import { getStorageManager } from '../src/storageManager.js';
+import {isFn, isStr, isPlainObject, mergeDeep, logError} from '../src/utils.js';
 
 export const storage = getStorageManager();
```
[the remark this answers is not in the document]
anthonylauzon: added to both modules

modules/haloRtdProvider.js
```diff
@@ -69,15 +70,12 @@ function paramOrDefault(param, defaultVal) {
  * @param {Object} rtdConfig
  */
 export function addRealTimeData(bidConfig, rtd, rtdConfig) {
-  if (!isPlainObject(config.getConfig().ortb2)) {
-    config.setConfig({ortb2: {}});
-  }
+  let ortb2 = config.getConfig().ortb2 || {};
```
robertrmartinez, line 73: if you pass in the specific config to getConfig you want, it will only return that, instead of the entire configs which have been set.

`config.getConfig('ortb2');`

This is recommended because If for some reason `config.getConfig()` does not return an object, then `config.getConfig().ortb2` will throw an exception.

anthonylauzon: thanks, fixed

test/spec/modules/haloIdSystem_spec.js
```diff
@@ -0,0 +1,31 @@
...
+      expect(request.url).to.eq(`https://id.halo.ad.gt/api/v1/pbhid`);
+      request.respond(200, { 'Content-Type': 'application/json' }, JSON.stringify({ haloId: 'testHaloId1' }));
+      expect(callbackSpy.lastCall.lastArg).to.deep.equal({haloId: 'testHaloId1'});
+    });
+
+    it('gets a cached haloid', function() {
+      const config = {
+        params: {}
+      };
+      storage.setDataInLocalStorage('auHaloId', 'tstCachedHaloId1');
```
robertrmartinez, line 24: It is not recommended to actually have the tests themselves setting localStorage / cookies.

These can become flakey depending on which browsers they are run in.

See https://github.com/prebid/Prebid.js/blob/master/test/spec/modules/rubiconAnalyticsAdapter_spec.js#L585-L589 as an example of mocking your storage object.

Then you can mock it's return like so: https://github.com/prebid/Prebid.js/blob/master/test/spec/modules/rubiconAnalyticsAdapter_spec.js#L1250


Also do not forget to call restore() on each stub in the afterEach's => https://github.com/prebid/Prebid.js/blob/master/test/spec/modules/rubiconAnalyticsAdapter_spec.js#L618-L620

test/spec/modules/haloRtdProvider_spec.js
```diff
@@ -1,220 +1,363 @@
...
+          ortb2: {
+            user: {
+              data: [rtdUserObj1]
+            }
+          }
+        }
+      };
 
-      getSegments(reqBidsConfigObj, () => {}, config, {});
+      storage.setDataInLocalStorage(RTD_LOCAL_NAME, JSON.stringify(cachedRtd));
```
robertrmartinez, line 279: Same thing here, stub this.

test/spec/modules/haloIdSystem_spec.js
anthonylauzon: i hadn't seen this pattern before, thanks for the info.

robertrmartinez (approved):
Looks good to me! 

Nice work, bit of a doosy this one!

robertrmartinez:
@ChrisHuie When you have some time, please give this a second look! Thanks.

ChrisHuie (approved):

Issue #6510, opened 2021-04-01
Support FLoC cohort IDs (trial & release)


Viewtiful:
## Type of issue
**Questions**

## Description
Following the start of the trial of the FLoC feature linked to the privacy sandbox project, we would like to start to receive cohort IDs from our adapter.
1. Do you plan to support a solution for this trial period?
2. If we introduce a logic to retrieve the cohort IDs from our adapter for the trial period, do you see any issues doing so?
Example of a possible logic (partial pseudo code):

Programmatically inject the token for the trials:

```javascript
function activateInterestCohortFeature() {
  const tokenElement = document.createElement('meta')
  tokenElement.httpEquiv = 'origin-trial'
  tokenElement.content = 'TOKEN'
  document.head.appendChild(tokenElement)
}
```

If in a chrome browser and `interest-cohort` feature is available, retrieve it and then send it in the bid requests constructed in the adapter:

```javascript
const isChrome = !!window.chrome && (!!window.chrome.webstore || !!window.chrome.runtime);
if (isChrome && !!document.featurePolicy && !!document.featurePolicy.features() && document.featurePolicy.features().includes('interest-cohort')) {
  return await document.interestCohort() || {};
}
```

3. Is there a longer term plan for the support of FLoC and cohort IDs prebid side?

patmmccann:
Injecting the token is a concern,  we're injecting our own token. Multiple tokens could have unexpected behavior. 

This seems better as an rtd submodule than part of a bidder.

ian-lr:
@jdwieland8282 and I were discussing this. I'd like to suggest a slightly different implementation for the Origin period:

- Register `sharedid.org` (or another prebid.org operated domain) as a "Third Party Matching" origin on https://developer.chrome.com/origintrials/#/register_trial/213920982300098561
- Create a new module for "FLoC Origin Support"
- Support collection of FLoC IDs when this module is included via dropping 3P JS (from `sharedid.org`) or an iframe. 
- The available FLoC ID could be treated as a pseudo userID (cached locally for a limited TTL, provided in the eid array, made available to bidders) for maximum compatibility:

```"userIdAsEids": [
    {
      "source": "web.dev",
      "uids": [
        {
          "id": "3258",
          "atype": 1,
        }
      ]
    }
  ],```

Alternatively, this could be made available in `fpd` or an alternative data payload. 

[...]

PR #6521, opened 2021-04-02
Floc Userid System: add new ID module

SKOCHERI:
## Type of change

- [ ] Bugfix
- [ ] Feature
- [ ] New bidder adapter   
- [ ] Build related changes
- [ ] CI related changes
- [ ] Other

## Description of change

```
{
  bidder: '<bidder name>',
  params: {
    // ...
  }
}
```


SKOCHERI:
https://github.com/prebid/Prebid.js/issues/6510

patmmccann:
Why store the Floc, can't you just always get it from the browser?

patmmccann (commented):
I think we'll want this to work when tokens are no longer necessary, if a missing token shouldn't throw an error 12 months from now, let's have it not throw one now.

SKOCHERI:
> I think we'll want this to work when tokens are no longer necessary, if a missing token shouldn't throw an error 12 months from now, let's have it not throw one now.

@patmmccann, Can you elaborate on this. "token" is marked as a mandatory parameters.

patmmccann:
The token only makes sense while floc is in trial
https://developer.chrome.com/origintrials/#/trials/active . As soon as this
is over, publishers should be able to use the module without a token.
Hence, token must be non-mandatory.

patmmccann:
The trial is over in three months https://developer.chrome.com/origintrials/#/view_trial/213920982300098561

jdwieland8282:
Is it obvious to anyone how the token actually gets used?  I just registered Sharedid.org for the origin trial, but based on this https://wicg.github.io/floc/#string-representation-of-the-interest-cohort-id there is no mention of how it actually should be used.

patmmccann:
@jdwieland8282 see second paragraph 

https://github.com/GoogleChrome/OriginTrials/blob/gh-pages/developer-guide.md

Viewtiful:
I agree with @patmmccann, it would be better from the start to entirely not rely on cookies

modules/userId/userId.md
```diff
@@ -169,7 +178,17 @@ pbjs.setConfig({
...
                  expires: 1
-        }
+                 }
+     },{
+            name: "originFloc",
+            params: {
+                token: "SHAREDID_TOKEN"
+                },
+            storage: {
+            type: "html5",
```
patmmccann, line 188: dont store the floc, get it fresh every time

smenzer: if this is the approach, then all examples should remove the `storage` config object. even better, a warning/error should be thrown in the `getId` function if `storage` was set at all, forcing publishers NOT to set it...but that depends on how important it is to fetch a fresh value every time or not

modules/flocIdSystem.js
```diff
@@ -0,0 +1,64 @@
...
+   * @param {SubmoduleConfig} [config]
+   * @returns {IdResponse|undefined}
+   */
+  getId(config) {
+    const configParams = (config && config.params) || {};
+    if (!configParams || (typeof configParams.token !== 'string')) {
+      utils.logError('User ID - flocId submodule requires token to be defined');
+      return;
+    }
+    return {id: configParams.token}
```
patmmccann, line 60: As @smenzer points out on #6510 

The floc is not the token,  the Floc needs to come from a browser api call

modules/userId/userId.md
patmmccann: Excellent point

patmmccann: @icflournoy you may want to do the same thing,  as storage interferes with your functionality and RDM sets it

pm-harshad-mane:
Hello @SKOCHERI , 
can you please add support to pass the version as well? Example below.
```
"eids": [
      {
        "source": "chrome.com",
        "uids": [
          {
            "id": "777454319066618",
            "atype": 1,
            "ext": {
              "ver": "chrome.1.1"
            }
          }
        ]
      }
    ]
```

modules/userId/userId.md
SKOCHERI: @patmmccann @smenzer Have updated to get cohort id from browser and removed the storage.

test/spec/modules/eids_spec.js
```diff
@@ -342,6 +342,23 @@ describe('eids array generation for known sub-modules', function() {
...
     });
   });
+  it('originFloc', function() {
+    const userId = {
+      flocId: {id: 'Sample_Cohort', version: 'chrome1.1'}
+    };
+    const newEids = createEidsArray(userId);
+    expect(newEids.length).to.equal(1);
+    expect(newEids[0]).to.deep.equal({
+      source: 'floc.com',
```
patmmccann, line 352: chrome.com and web.dev have been proposed in #6510 
google.com seems like a candidate as well. 
does anyone have a preference?

modules/flocIdSystem.js
```diff
@@ -0,0 +1,84 @@
...
+    const isFlocEnabled = !!document.featurePolicy && !!document.featurePolicy.features() && document.featurePolicy.features().includes('interest-cohort');
+
+    if (isChrome && isFlocEnabled) {
+      const configParams = (config && config.params) || {};
+      if (!configParams || (typeof configParams.token !== 'string')) {
+        utils.logInfo('User ID - flocId submodule token is not defined');
+      }
+
+      // Insert meta-tag with token from configuration
+      enableOriginTrial(configParams.token);
```
patmmccann, line 76: this module should work with no token, as the token will no longer be required after July. It seems like this function should be behind the conditional for is the token defined.

integrationExamples/gpt/userId_example.html
```diff
@@ -283,6 +283,12 @@
                         },
                           {
                             name: "uid2"
+                          },
+                          {
+                            name: "originFloc",
+                            params: {
+                              token: "SHARED_ID_TOKEN"
```
patmmccann, line 290: I can't find the value of this token in your PR yet you say it is the default.

test/spec/modules/eids_spec.js
smenzer: i would vote for `chrome.com` as it's the browser providing this, and probably shouldn't be as broad as google.com. chrome.com vs web.dev, though, I don't have a strong opinion on

ian-lr: I'll second `chrome.com`.

integrationExamples/gpt/userId_example.html
SKOCHERI: @patmmccann For token, should the publisher register one based on their domain? For testing this module, I used the chrome version 91.0.4469.4 (Official Build) dev (x86_64) and enabled --enable-blink-features=InterestCohortAPI 
--enable-features="FederatedLearningOfCohorts:update_interval/10s/minimum_history_domain_size_required/1,FlocIdSortingLshBasedComputation,InterestCohortFeaturePolicy"

With these flags enabled, the token is not required and also for checking if "interest-cohort" is available, the chrome flag "Experimental Web Platform features" had to be enabled.

modules/flocIdSystem.js
```diff
@@ -0,0 +1,108 @@
...
+  },
+  /**
+   * If chrome and cohort enabled performs action to obtain id and return a value in the callback's response argument
+   * @function
+   * @param {SubmoduleConfig} [config]
+   * @returns {IdResponse|undefined}
+   */
+  getId(config) {
+    const isChrome = !!window.chrome && (!!window.chrome.webstore || !!window.chrome.runtime);
+
```
SKOCHERI, line 79: @smenzer @patmmccann For checking the chrome, should window.chrome check suffice? I was unable to test window.chrome.webstore || !!window.chrome.runtime as webstore/runtime is not available

modules/flocIdSystem.js
```diff
@@ -0,0 +1,108 @@
...
+ * This module adds flocId to the User ID module
+ * The {@link module:modules/userId} module is required
+ * @module modules/flocId
+ * @requires module:modules/userId
+ */
+
+import * as utils from '../src/utils.js'
+import {submodule} from '../src/hook.js'
+
+const MODULE_NAME = 'originFloc';
```
smenzer, line 11: should we make sure the module name matches the file name?

modules/flocIdSystem.js
smenzer: i'm not sure...sorry!

modules/flocIdSystem.md
```diff
@@ -0,0 +1,33 @@
...
+    }
+});
+```
+
+### Parameter Descriptions for the `userSync` Configuration Section
+The below parameters apply only to the FloC ID User ID Module integration.
+
+| Params under usersync.userIds[]| Scope | Type | Description | Example |
+| --- | --- | --- | --- | --- |
+| name | Required | String | ID value for the Floc ID module - `"flocId"` | `"flocId"` |
```
smenzer, line 30: this isn't consistent with the other examples or the module name; although I suggest changing the others to use `flocId` as the module name

modules/flocIdSystem.md
```diff
@@ -0,0 +1,33 @@
...
+| params | Optional | Object | Details for flocId syncing. | |
+| params.token | Optional | Object | SharedId's token by default | 60 |
+| storage | Not Allowed | Object | Will ask browser for cohort everytime. Setting storage will fail id lookup ||
```
smenzer, line 33: should you maybe add a check in `getId()` to look for storage being set, and if so, throw an error and fail out?

test/spec/modules/userId_spec.js
```diff
@@ -1434,6 +1438,30 @@ describe('User ID', function () {
...
+            expect(bid.userIdAsEids[0]).to.deep.equal({
+              source: 'chrome.com',
+              uids: [{id: 'Sample_Cohort', atype: 1, ext: { ver: 'chrome1.1' }}]
+            });
+          });
+        });
+        coreStorage.setCookie('flocId', '', EXPIRED_COOKIE_DATE);
+        done();
+      }, {adUnits});
+    });
```
smenzer, line 1463: is this really a useful test? the module doesn't read from cookies and doesn't allow setting storage param (which the `getConfigMock()` is doing). I think you need to mock the `document.interestCohort()` instead

integrationExamples/gpt/userId_example.html
smenzer: I was also a bit confused like @patmmccann - in all the examples it says to use a Share ID token but doesn't mention what it is. If you don't need a token, then I think you should include a description for how to operate without one so it's clear for publishers.

I would suggest either: 
1. change the text in the examples to be something like `"INSERT_PUBLISHER_TOKEN"` and then provide links or a description of how to obtain it; OR
2. actually provide a valid token to use as a default

smenzer (changes requested):
Some changes requested, but this is getting there!

integrationExamples/gpt/userId_example.html
ian-lr: @SKOCHERI @smenzer My understanding is that for the origin trial, browsers that are not explicitly setting flags (~5% of non-EEA Chrome instances) will need a token set before they will respond with a cohort ID. Setting flags overrides this behavior. 

To avoid publishers _having_ to enroll in the Origin trial process, I'd recommend setting a default token. I just obtained `A3dHTSoNUMjjERBLlrvJSelNnwWUCwVQhZ5tNQ+sll7y+LkPPVZXtB77u2y7CweRIxiYaGwGXNlW1/dFp8VMEgIAAAB+eyJvcmlnaW4iOiJodHRwczovL3NoYXJlZGlkLm9yZzo0NDMiLCJmZWF0dXJlIjoiSW50ZXJlc3RDb2hvcnRBUEkiLCJleHBpcnkiOjE2MjYyMjA3OTksImlzU3ViZG9tYWluIjp0cnVlLCJpc1RoaXJkUGFydHkiOnRydWV9` for sharedid.org origin. Can we set that as the fallback value?

modules/flocIdSystem.md
SKOCHERI: Updated the module to "flocId"

modules/flocIdSystem.md
SKOCHERI: Logging error if storage is configured

smenzer:
after discussion in the Prebid Privacy & Identity PMC, it was agreed that a user id module could be used to capture the floc id, but that it should not be added to the eids array which is more difficult to parse out. so, we'll remove the `eids.js` and `eids.md` code from this PR.

since some ssp's may still want to pass the floc id in eids to themself, we should add a utility method to create an eids array and add the floc id, called `getUserIdsAsEidsWithFloc()`.

cc: @SKOCHERI @jdwieland8282 @ian-lr @pm-harshad-mane

SKOCHERI:
> after discussion in the Prebid Privacy & Identity PMC, it was agreed that a user id module could be used to capture the floc id, but that it should not be added to the eids array which is more difficult to parse out. so, we'll remove the `eids.js` and `eids.md` code from this PR.
> 
> since some ssp's may still want to pass the floc id in eids to themself, we should add a utility method to create an eids array and add the floc id, called `getUserIdsAsEidsWithFloc()`.
> 
> cc: @SKOCHERI @jdwieland8282 @ian-lr @pm-harshad-mane

I will go ahead to remove the eid.js and eids.md and related test.

Do we want to add `getUserIdsAsEidsWithFloc()` as part of this PR?

smenzer:
> Do we want to add `getUserIdsAsEidsWithFloc()` as part of this PR?

That would be ideal but not a requirement I would say.

integrationExamples/gpt/userId_example.html
```diff
@@ -283,6 +283,14 @@
                         },
                           {
                             name: "uid2"
+                          },
+                          {
+                            name: "flocId",
+                            params: {
+                              // Default sharedid.org token : "A3dHTSoNUMjjERBLlrvJSelNnwWUCwVQhZ5tNQ+sll7y+LkPPVZXtB77u2y7CweRIxiYaGwGXNlW1/dFp8VMEgIAAAB+eyJvcmlnaW4iOiJodHRwczovL3NoYXJlZGlkLm9yZzo0NDMiLCJmZWF0dXJlIjoiSW50ZXJlc3RDb2hvcnRBUEkiLCJleHBpcnkiOjE2MjYyMjA3OTksImlzU3ViZG9tYWluIjp0cnVlLCJpc1RoaXJkUGFydHkiOnRydWV9"
+                              // To get new token, register https://developer.chrome.com/origintrials/#/trials/active for Federated Learning of Cohorts
```
smenzer, line 291: @SKOCHERI this is perfect, thanks!

modules/flocIdSystem.md
smenzer: thanks!

modules/userId/userId.md
```diff
@@ -113,6 +113,11 @@ pbjs.setConfig({
                   name: '__adm__admixer',
                   expires: 30
               }
+        },{
+              name: 'flocId',
+               params: {
+                    token: "Registered token or default sharedid.org token" // Default sharedid.org token: "A3dHTSoNUMjjERBLlrvJSelNnwWUCwVQhZ5tNQ+sll7y+LkPPVZXtB77u2y7CweRIxiYaGwGXNlW1/dFp8VMEgIAAAB+eyJvcmlnaW4iOiJodHRwczovL3NoYXJlZGlkLm9yZzo0NDMiLCJmZWF0dXJlIjoiSW50ZXJlc3RDb2hvcnRBUEkiLCJleHBpcnkiOjE2MjYyMjA3OTksImlzU3ViZG9tYWluIjp0cnVlLCJpc1RoaXJkUGFydHkiOnRydWV9"|  token: "A3dHTSoNUMjjERBLlrvJSelNnwWUCwVQhZ5tNQ+sll7y+LkPPVZXtB77u2y7CweRIxiYaGwGXNlW1/dFp8VMEgIAAAB+eyJvcmlnaW4iOiJodHRwczovL3NoYXJlZGlkLm9yZzo0NDMiLCJmZWF0dXJlIjoiSW50ZXJlc3RDb2hvcnRBUEkiLCJleHBpcnkiOjE2MjYyMjA3OTksImlzU3ViZG9tYWluIjp0cnVlLCJpc1RoaXJkUGFydHkiOnRydWV9"
```
smenzer, line 119: i think you copied this from the table above, so you have an extra copy of the token

```suggestion
                    token: "Registered token or use the default sharedid.org token" // Default sharedid.org token: "A3dHTSoNUMjjERBLlrvJSelNnwWUCwVQhZ5tNQ+sll7y+LkPPVZXtB77u2y7CweRIxiYaGwGXNlW1/dFp8VMEgIAAAB+eyJvcmlnaW4iOiJodHRwczovL3NoYXJlZGlkLm9yZzo0NDMiLCJmZWF0dXJlIjoiSW50ZXJlc3RDb2hvcnRBUEkiLCJleHBpcnkiOjE2MjYyMjA3OTksImlzU3ViZG9tYWluIjp0cnVlLCJpc1RoaXJkUGFydHkiOnRydWV9"
```

modules/userId/userId.md
```diff
@@ -222,6 +227,11 @@ pbjs.setConfig({
                   name: 'admixerId',
                   expires: 30
               }
+       },{
+              name: 'flocId',
+               params: {
+               token: "Registered token or default sharedid.org token" // Default sharedid.org token: "A3dHTSoNUMjjERBLlrvJSelNnwWUCwVQhZ5tNQ+sll7y+LkPPVZXtB77u2y7CweRIxiYaGwGXNlW1/dFp8VMEgIAAAB+eyJvcmlnaW4iOiJodHRwczovL3NoYXJlZGlkLm9yZzo0NDMiLCJmZWF0dXJlIjoiSW50ZXJlc3RDb2hvcnRBUEkiLCJleHBpcnkiOjE2MjYyMjA3OTksImlzU3ViZG9tYWluIjp0cnVlLCJpc1RoaXJkUGFydHkiOnRydWV9"|  token: "A3dHTSoNUMjjERBLlrvJSelNnwWUCwVQhZ5tNQ+sll7y+LkPPVZXtB77u2y7CweRIxiYaGwGXNlW1/dFp8VMEgIAAAB+eyJvcmlnaW4iOiJodHRwczovL3NoYXJlZGlkLm9yZzo0NDMiLCJmZWF0dXJlIjoiSW50ZXJlc3RDb2hvcnRBUEkiLCJleHBpcnkiOjE2MjYyMjA3OTksImlzU3ViZG9tYWluIjp0cnVlLCJpc1RoaXJkUGFydHkiOnRydWV9"
```
smenzer, line 233: same thing here

```suggestion
               token: "Registered token or use the default sharedid.org token" // Default sharedid.org token: "A3dHTSoNUMjjERBLlrvJSelNnwWUCwVQhZ5tNQ+sll7y+LkPPVZXtB77u2y7CweRIxiYaGwGXNlW1/dFp8VMEgIAAAB+eyJvcmlnaW4iOiJodHRwczovL3NoYXJlZGlkLm9yZzo0NDMiLCJmZWF0dXJlIjoiSW50ZXJlc3RDb2hvcnRBUEkiLCJleHBpcnkiOjE2MjYyMjA3OTksImlzU3ViZG9tYWluIjp0cnVlLCJpc1RoaXJkUGFydHkiOnRydWV9"
```

smenzer (approved):
Just a minor doc fix and then this is good to go

aleksatr:
@smenzer I'm not sure what are the test-coverage requirements for id submodules, but this one is only at 8%?

smenzer:
> @smenzer I'm not sure what are the test-coverage requirements for id submodules, but this one is only at 8%?

most of hte testing of he user id modules is in the userid_spec.js, so i think most modules are pretty well covered. 8% seems really low, though, so we should make sure the necessary tests are added here to get that up.

patmmccann:
I don't like the planned obsolescence of this module to july 2021 when the origin trial ends. It needs to work without a token configured, not throw a warning without one,  and not try to insert the sharedid token unless configuration specifies the origin trial is still occurring

jdwieland8282:
Hi @patmmccann that's not how I would characterize our approach.  We definitely see this module working post 7/21. It seems difficult to give the module behavior with a TTL. Like do this until ... then do that. Is it ok if we use the sharedid token by default (configurable of course) and throw a warning if omitted until July, then update the module in July?

patmmccann:
That's not a good approach, just fix in July, as publishers have vary long and variable upgrade schedules. You guys can see that in your internal data I'm sure. Why not just not throw the warning in the code and put the warning in the doc that if you don't have a token, this isn't going to work from April 2021 until End of July 2021?

jdwieland8282:
That's fine with me @patmmccann we will go with that approach.  Any other changes you'd like to see?  Are you ok using sharedid's token by default?

patmmccann:
that's my only remaining concern, thanks!

patmmccann (approved):

integrationExamples/gpt/userId_example.html
```diff
@@ -2,312 +2,320 @@
 <html lang="en">
 
 <head>
-  <title>User ID Modules Example</title>
```
aleksatr, line 5: @SKOCHERI can you revert the formatting changes you made to this file, so it includes only the floc-related changes?

modules/flocIdSystem.js
```diff
@@ -0,0 +1,111 @@
...
+
+const MODULE_NAME = 'flocId';
+
+/**
+ * Add meta tag to support enabling of floc origin trial
+ * @function
+ * @param {string} token - configured token for origin-trial
+ */
+function enableOriginTrial(token) {
+  const tokenElement = document.createElement('meta')
```
aleksatr, line 19: maybe you want to end lines with `;` in this function as well, just to be consistent with the rest of the file.

modules/flocIdSystem.js
```diff
@@ -0,0 +1,111 @@
...
+      utils.logError('User ID - flocId submodule storage should not defined');
+      return;
+    }
+    const isChrome = !!window.chrome && (!!window.chrome.webstore || !!window.chrome.runtime);
+    // Validate feature is enabled
+    const isFlocEnabled = !!document.featurePolicy && !!document.featurePolicy.features() && document.featurePolicy.features().includes('interest-cohort');
+
+    if (isChrome && isFlocEnabled) {
+      const configParams = (config && config.params) || {};
+      if (configParams || (typeof configParams.token === 'string')) {
```
aleksatr, line 90: @SKOCHERI did you maybe mean `configParams && (typeof configParams.token === 'string'` ? (judging by the change history, to get the same behavior)

jdwieland8282:
@aleksatr Can you please review and approve if you are satisfied, we'd like to include this in todays pb.js release

modules/flocIdSystem.js
aleksatr: I meant all the statements in this function, not just on this line :)

modules/flocIdSystem.js
```diff
@@ -0,0 +1,111 @@
...
+  getId(config) {
+    // Block usage of storage of cohort ID
+    const checkStorage = (config && config.storage);
+    if (checkStorage) {
+      utils.logError('User ID - flocId submodule storage should not defined');
+      return;
+    }
+    const isChrome = !!window.chrome && (!!window.chrome.webstore || !!window.chrome.runtime);
+    // Validate feature is enabled
+    const isFlocEnabled = !!document.featurePolicy && !!document.featurePolicy.features() && document.featurePolicy.features().includes('interest-cohort');
```
smenzer, line 86: out of curiosity, what's the reason to check for chrome separately from floc, considering it's `&&` together next and they are never used separately? couldn't it just be one long statement for `isFlocEnabled`?

aleksatr (approved):

patmmccann:
@jdwieland8282 if we resolve conflicts i think this can ship before the release today

SKOCHERI:
> @jdwieland8282 if we resolve conflicts i think this can ship before the release today

Will do it right away

SKOCHERI:
@patmmccann  Have resolved the conflicts but the tests are failing. I don't see any user id tests failing. Have tried rerunning multiple times. Do you have any pointers?

patmmccann:
Those tests are unrelated,  also failing on master

paulandrewfarrow:
Any idea when this will start showing up at https://docs.prebid.org/download.html?

ian-lr:
Hey @paulandrewfarrow I think this needed a change to the download page; I've just made a PR here which should address that once merged: https://github.com/prebid/prebid.github.io/pull/3005

jdwieland8282:
docs PR https://github.com/prebid/prebid.github.io/pull/3006

PR #6525, opened 2021-04-05
Seedtag adapter: Fixing bug preventing to receive the right params onTimeout.

lauramorillo:
## Type of change

- [X] Bugfix
- [ ] Feature
- [ ] New bidder adapter   
- [ ] Build related changes
- [ ] CI related changes
- [ ] Other

## Description of change

Using the right url to report timeout information onTimeout, before we were generating the url but we were using always a fixed one when triggering the pixel.

- laura@seedtag.com

### DIFF
--- a/test/spec/modules/seedtagBidAdapter_spec.js
+++ b/test/spec/modules/seedtagBidAdapter_spec.js
@@ -409,6 +409,14 @@ describe('Seedtag Adapter', function() {
   })
 
   describe('onTimeout', function () {
+    beforeEach(function() {
+      sinon.stub(utils, 'triggerPixel')
+    })
+
+    afterEach(function() {
+      utils.triggerPixel.restore()
+    })
+
     it('should return the correct endpoint', function () {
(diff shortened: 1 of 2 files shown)


test/spec/modules/seedtagBidAdapter_spec.js
```diff
@@ -420,6 +428,16 @@ describe('Seedtag Adapter', function() {
         params.adUnitId
       )
     })
+
+    it('shoud set the timeout pixel', function() {
```
patmmccann, line 432: spelling

is the goal of this PR more stable tests, similar to https://github.com/prebid/Prebid.js/pull/6511/files ? 

If so, much appreciated!

lauramorillo: I fixed the typo, thanks! With this PR I wanted to fix a bug that we found with the timeout and I wanted to ensure that part was properly tested (that's why I added another test to ensure the pixel is properly set with the right parameters).

PR #6537, opened 2021-04-07
Deepintent ID System: add new ID module

sourabhg:
## Type of change

- [x] Feature

## Description of change
New Deepintent user module is added which supports passing of userIds as per deepintent DPES identity.

sourabh@deepintent.com, prebid@deepintent.com

https://github.com/prebid/prebid.github.io/pull/2830


modules/.submodules.json
```diff
@@ -16,7 +16,11 @@
     "zeotapIdPlusIdSystem",
     "haloIdSystem",
     "quantcastIdSystem",
+<<<<<<< HEAD
+    "deepintentDpesIdSystem",
+=======
     "nextrollIdSystem",
+>>>>>>> 81fc1ce9b8471cadb33ebc10a270e8c6d001ea98
```
ChrisHuie, line 23: Can you please explain the concept here. Is it constructing a bit shifting hash?

sourabhg: Hey @ChrisHuie  are you referring to what Deepintent module does ?

sourabhg: I see @ChrisHuie I think i forgot one conflict resolution commit. just pushed it should be good now

ChrisHuie: thanks :)

ChrisHuie:
@sourabhg it looks like our circleci tests didn't run on this pr.

sourabhg:
@ChrisHuie CircleCI generally runs automatically on all of the things, as I have used the fork flow always.

sourabhg:
@ChrisHuie : I just gave you write access to our forked repo. I hope that helps.

ChrisHuie:
@sourabhg thanks. Just kicked off the circleci tests. Feel free to remove me from the repo :) I am curious about this situation because I have seen it before but usually with new committers to Prebid. So this is a new situation to me

sourabhg:
Thanks @ChrisHuie May I know when this can be merged ?

modules/deepintentDpesIdSystem.js
```diff
@@ -0,0 +1,80 @@
...
+  getId(config) {
+    const configStorage = (config && config.storage) || {};
+
+    let id = null;
+    if (
+      configStorage &&
+      configStorage.name &&
+      configStorage.type === 'cookie'
+    ) {
+      id = readCookie(configStorage.name);
```
smenzer, line 71: why are you manually reading from the cookie if you are asking publishers to set the `storage` config?  the userId module will already handle storage and retrieval you just need to add the `cacheIdObj` to the `getId()` signature

modules/deepintentDpesIdSystem.md
```diff
@@ -0,0 +1,49 @@
+# Deepintent DPES ID
+
+The Deepintent DPES ID or Healthcare Identity is a shared, neutral identifier that publishers and ad tech platforms can use to recognise users even in environments where 3rd party cookies are not available. The Deepintent DPES ID is designed to respect users' privacy choices and publishers’ preferences throughout the advertising value chain.
```
smenzer, line 3: This is an exact copy of the [ID5 description](https://github.com/prebid/Prebid.js/blob/master/modules/id5IdSystem.md#L3). You should not just be copying and pasting here, it's meant to describe YOUR id solution.

modules/deepintentDpesIdSystem.md
```diff
@@ -0,0 +1,49 @@
...
+  userSync: {
+    userIds: [{
+      name: 'deepintentId',
+      params: {
+        siteId: 173,            // change to the siteId number to the one you recieved from Deepintent.
+        identityKey: "hashedEmail" // HashedEmail or HashedNPI based on type of your integration
+      },
+      storage: {
+        type: 'cookie',           // "html5" is the required storage type option is "html5"
+        name: '_di',            // change to the cookie name you wish to read from, optional if storage type is html5
```
smenzer, line 31: why do you treat cookie and html5 differently?

modules/userId/eids.js
```diff
@@ -207,7 +207,10 @@ const USER_IDS_CONFIG = {
       return data.id;
     }
   },
-
+  'deepintentId': {
+    source: 'deepintent.com',
+    atype: 1
```
smenzer, line 212: if this is an email-based identifier, then it should be `3` here I think. @patmmccann you've been reviewing a lot of these, do you agree?

smenzer (changes requested):
A few comments in the code. You also need to add tests in the `userId_spec.js` and `eids_spec.js` files

modules/deepintentDpesIdSystem.md
sourabhg: This is exactly what our solution does as well and describes what we do. If you still persist I can change it with diffrent language

modules/deepintentDpesIdSystem.md
sourabhg: We give a way to specify cookie name for pubs and for localstorage key is already set with us. This is to accomodate the exisitng solution we have. Once we made changes to that code we might change the working here as well for now no changes and thank you for suggestion.

modules/userId/eids.js
sourabhg: I will make the change, Is there a list as well as which is being referred ?

modules/deepintentDpesIdSystem.js
sourabhg: This code was started when cacheIdObj signature was not there, the code will be later changed to accomodate this in phase 2.

sourabhg: As a total change I will be changing the description in Doc and adding unit test cases in eids_spec.js and userId_spec.js thanks for letting me know.

patmmccann:
Thank you for fixing the atype; everyone just put 1 at first and I've been trying to make them accurate.

modules/userId/eids.js
patmmccann: https://github.com/InteractiveAdvertisingBureau/AdCOM/blob/master/AdCOM%20v1.0%20FINAL.md#list_agenttypes

modules/deepintentDpesIdSystem.js
smenzer: the `storedId` has been part of the signature for years now...

patmmccann: Why wait for a phase two,  I'm in the do it right the first time camp

modules/deepintentDpesIdSystem.md
patmmccann: run that by your corporate pr folks, they may frown on plaigarism

modules/deepintentDpesIdSystem.js
sourabhg: storage will not be handled by UserModule as we do not use any http calls to be made to retrieve id from server and store it via user module. This id is stored by our select publishers side which needs to register on our platform and thus we have to read from storage manually, thus keeping the structure same as per the plan.

modules/deepintentDpesIdSystem.md
sourabhg: Thank you changed the language now.

sourabhg:
@smenzer @patmmccann : Added in all the necessery test cases, changed the documentation, changed the atype to 3 and also changed the logic to pick from cacheIdObject. 

@ChrisHuie : somehow CircleCI again doesnt work through my commits can you help. ?

modules/deepintentDpesIdSystem.md
sourabhg: Changes are completed as suggested.

modules/deepintentDpesIdSystem.js
sourabhg: changes are completed as suggested.

modules/deepintentDpesIdSystem.js
```diff
@@ -0,0 +1,51 @@
...
+   * @returns {{deepintentId:Object}}
+   */
+  decode(value, config) {
+    if (value) {
+      return {
+        'deepintentId': value
+      }
+    } else {
+      return undefined
+    }
```
smenzer, line 34: this could be simplified with 

```suggestion
    return value ? { 'deepintentId': value } : undefined;
```

modules/deepintentDpesIdSystem.md
```diff
@@ -0,0 +1,42 @@
+# Deepintent DPES ID
+
+The Deepintent Id  is a shared, healthcare identifier which helps publisher in absence of the 3rd Party cookie matching. This lets publishers set and bid with healthcare identity . Deepintent lets users protect their privacy through advertising value chain, where Healthcare identity when setting the identity takes in consideration of users choices, as well as when passing identity on the cookie itself privacy consent strings are checked. The healthcare identity when set is not stored on Deepintent's servers but is stored on users browsers itself. User can still opt out of the ads by https://option.deepintent.com/adchoices
+
+## Deepintent DPES ID Registration
+
+The Deepintent DPES ID is free to use, but requires a simple registration with Deepintent. Please contact your sales rep to get started
+
+
+## Deepintent DPES ID Configuration
```
smenzer, line 10: as mentioned in the docs PR, it seems to be required that publishers have additional Deepintent code on their page to retrieve and store the DPES ID in order for the user id module to pick it up. Assuming that's the case, I think that should be clearly documented here somewhere

smenzer (changes requested):
Just needs clarification on additional code required to work, but otherwise this is good to go

modules/deepintentDpesIdSystem.js
sourabhg: Made the change.

modules/deepintentDpesIdSystem.md
sourabhg: @smenzer : as suggested I added more details about the Deepintent registration and a tag code that publisher needs to integrate to get started with this user module, can we merge this now ?

sourabhg:
@smenzer @ChrisHuie : All review comments have been addressed can we approve and merge this now ?

smenzer (approved):
Thanks for the updates...LGTM

Issue #6301, opened 2021-02-11
Bump browserstack versions


patmmccann:
## Type of issue
Feature request

## Description
https://github.com/prebid/Prebid.js/blob/master/browsers.json seems quite out of date and also appears to be missing Chrome on Android, which according to https://www.browserstack.com/question/476 is 100% the same as real android. Chrome on iOS may present some issues according to the same page.

ChrisHuie:
I have been diving into testing and need to become more familiar with Browserstack so I can follow-up on this.

patmmccann:
I think changing lines 26 and 30 here https://github.com/prebid/Prebid.js/blob/master/browsers.json solves half the problem. Then we would just need to add a small block for chrome on android.

ChrisHuie:
Scheduled to bring up in meeting to determine current versions and browsers we want to test both functionally and compatiability wise

ChrisHuie:
Confirmed browerstack to most recent and currently latest version already included in browerstack

ChrisHuie:
waiting on merge of #6663 and recent IE test fixes to update other versions browsers

ChrisHuie:
closing with pr #6828

PR #6539, opened 2021-04-07
PBjs Core Testing: update Windows 10 testing on Chrome to v 89.0

ChrisHuie:
Update windows 10 testing on chrome from version 80.0 to 89.0. There also still remains testing in this environment for chrome 79.0 however functional testing will only run on this new version since chrome 79.0 is dropped for functional testing in https://github.com/prebid/Prebid.js/blob/master/wdio.conf.js

### DIFF
--- a/browsers.json
+++ b/browsers.json
@@ -23,11 +23,11 @@
     "device": null,
     "os": "Windows"
   },
-  "bs_chrome_80_windows_10": {
+  "bs_chrome_89_windows_10": {
     "base": "BrowserStack",
     "os_version": "10",
     "browser": "chrome",
-    "browser_version": "80.0",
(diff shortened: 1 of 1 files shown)


ChrisHuie:
reference issue #6301

patmmccann:
Since this passed tests,  are you planning to add other changes here or do them each in a new pr?

ChrisHuie:
@patmmccann I was planning to do a separate pr for at least the browsers that are running through the functional tests

patmmccann (approved):

patmmccann:
@ChrisHuie I think this requires a second review, added @robertrmartinez to take a look if he's available

robertrmartinez (approved):

PR #6549, opened 2021-04-08
Yieldmo adapter: cut banner bid request parameters in case the url for banner request is too long

ym-atsymuk:
## Type of change

- [x] Bugfix
- [ ] Feature
- [ ] New bidder adapter   
- [ ] Build related changes
- [ ] CI related changes
- [ ] Other

## Description of change
Cut banner bid request parameters in case the url for banner request is too long

contact email of the adapter’s maintainer: opensource@yieldmo.com

### DIFF
--- a/modules/yieldmoBidAdapter.js
+++ b/modules/yieldmoBidAdapter.js
@@ -18,6 +18,8 @@ const LOCAL_WINDOW = utils.getWindowTop();
 const DEFAULT_PLAYBACK_METHOD = 2;
 const DEFAULT_START_DELAY = 0;
 const VAST_TIMEOUT = 15000;
+const MAX_BANNER_REQUEST_URL_LENGTH = 8000;
+const BANNER_REQUEST_PROPERTIES_TO_REDUCE = ['description', 'title', 'pr', 'page_url'];
 
 export const spec = {
   code: BIDDER_CODE,
@@ -51,7 +53,7 @@ export const spec = {
         p: [],
         page_url: bidderRequest.refererInfo.referer,
         bust: new Date().getTime().toString(),
-        pr: bidderRequest.refererInfo.referer,
+        pr: (LOCAL_WINDOW.document && LOCAL_WINDOW.document.referrer) || '',
         scrd: LOCAL_WINDOW.devicePixelRatio || 0,
         dnt: getDNT(),
         description: getPageDescription(),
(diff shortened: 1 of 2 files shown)


patmmccann:
Why not just do a post

ym-atsymuk:
> Why not just do a post

thank you. we will consider this option

patmmccann:
This could work as a bridge, are you sure you don't want this temporarily?

ym-atsymuk:
> This could work as a bridge, are you sure you don't want this temporarily?

currently we does not support POST for banner requests, so we decided to go with this variant temporarily

modules/yieldmoBidAdapter.js
```diff
@@ -92,6 +94,20 @@ export const spec = {
         }
       });
       serverRequest.p = '[' + serverRequest.p.toString() + ']';
+
+      // check if url exceeded max length
+      const url = `${BANNER_SERVER_ENDPOINT}?${utils.parseQueryStringParameters(serverRequest)}`;
+      let extraCharacters = url.length - MAX_BANNER_REQUEST_URL_LENGTH;
+      if (extraCharacters > 0) {
+        for (const property of BANNER_REQUEST_PROPERTIES_TO_REDUCE) {
```
robertrmartinez, line 102: Not sure our transpiler handles the `for... of ` for IE. So some tests will fail in IE I am guessing.

So please update this to use ES5 if possible!

FilipStamenkovic: I think our transpiler will handle `for... of` loop even for IE. Looking at the build I see that this loop was changed to:

```
var _iterator = _createForOfIteratorHelper(BANNER_REQUEST_PROPERTIES_TO_REDUCE),
    _step;

try {
  for (_iterator.s(); !(_step = _iterator.n()).done;) {
    var property = _step.value;
    extraCharacters = shortcutProperty(extraCharacters, serverRequest, property);

    if (extraCharacters <= 0) {
      break;
    }
  }
} catch (err) {
  _iterator.e(err);
} finally {
  _iterator.f();
}
```

But, I would also suggest, if possible, not to use `for... of` loop, simply because built `.js` file would have less code and it would be a bit smaller.

FilipStamenkovic (changes requested):
Your tests are failing, according to the circleci, 9 tests are failing.  
Can you update your code/tests?  

Also, please take a look at the discussion for `for... of` loop.

ym-atsymuk:
> Your tests are failing, according to the circleci, 9 tests are failing.
> Can you update your code/tests?
> 
> Also, please take a look at the discussion for `for... of` loop.

Thank you for your feedback. `for... of` replaced with simple for loop.

Which tests are failing? On my local env tests are green

ym-atsymuk:
> @ym-atsymuk Take a look at the circleci output:
> 
> https://app.circleci.com/pipelines/github/prebid/Prebid.js/4991/workflows/bf6594f1-7372-4a70-a411-6d2ee7ed4d3c/jobs/13345

fixed

FilipStamenkovic (approved):
LGTM

Issue #6386, opened 2021-03-05
Cookie _pbjs_userid_consent_data


thomasmasereel:
Hello,
We noticed the cookie _pbjs_userid_consent_data being stored on our properties, whereas IAB TCF v2 consent string doesn't allow it. 
You can see the behaviour on https://www.linternaute.com/, french ip, deny all through the CMP.
I understand it is linked to prebid user id module and the name mention consent data.
Could you document more precisely on this cookie : purposes, content, description etc.
Thank you

smenzer:
Hi @thomasmasereel - this cookie is used to know if the user's consent choices have changed since the last page load. It is a hashed (`cyrb53Hash`) value of the consent string with a 30 day expiration. The use case is as follows:
- user visits the site and allows all vendors when first prompted.
- user then changes their consent preferences
- user visits another page on the site. at this point, it is important to refresh the user id's in case the relevant consent for those vendors have changed and they can issue a new ID (if applicable) or return a "no-consent" signal so that the ID is no longer available.

thomasmasereel:
Thank you for your quick answer @smenzer.
Could a note added to prebid.org, in the user id module section, be considered ?
As for EU country, every cookie stored in a non consent environnement is being closely and carefully monitor.

smenzer:
I think that's probably reasonable - @bretg / @patmmccann what do you guys think? (cc'ing you both b/c there have been other discussions about prebid setting cookies)

patmmccann:
this seems closely related to https://github.com/prebid/Prebid.js/issues/6289 . I personally don't get the idea that prebid.js needs consent to write cookies. Publishers can write cookies such as this so that their site works properly, and prebid is a part of their site, deployed by them. This is a part of how prebid functions. I cannot envision a scenario where a regulator objects to this cookie, as publishers using prebid set it to better comply with the law, similar to how a cmp sets values related to consent. 

However, if others disagree, I suppose one could check to make sure prebid vendor id has purpose 1 consent, but I think that should not be the default behavior, as prebid in this case is not setting the cookie like it does in the sharedid sennse, the publisher is. 

I think it makes sense to document this at https://docs.prebid.org/dev-docs/modules/userId.html#how-it-works so anyone looking for this string on our documentation finds the explanation easily. This probably should have been done with https://github.com/prebid/Prebid.js/pull/5451

smenzer:
I agree with everything you said, @patmmccann - and yea, my bad, I should have documented it as part of #5451.

[...]

PR #6551, opened 2021-04-08
userID Module: move reading and writing of the consent string hash

smenzer:
move reading and writing of hte consent string hash for the user id module to after the gdpr checks. if no user id modules will be executed, this value isn't really necessary anyways.

this should alleviate the issue brought up in https://github.com/prebid/Prebid.js/issues/6386

somewhat related docs PR: https://github.com/prebid/prebid.github.io/pull/2839

## Type of change

- [x] Bugfix
- [ ] Feature
- [ ] New bidder adapter   
- [ ] Build related changes
- [ ] CI related changes
- [ ] Other

### DIFF
--- a/modules/userId/index.js
+++ b/modules/userId/index.js
@@ -590,15 +590,16 @@ function refreshUserIds(options, callback) {
   initializeSubmodulesAndExecuteCallbacks(function() {
     let consentData = gdprDataHandler.getConsentData()
 
-    const storedConsentData = getStoredConsentData();
-    setStoredConsentData(consentData);
-
     // gdpr consent with purpose one is required, otherwise exit immediately
     let {userIdModules, hasValidated} = validateGdprEnforcement(submodules, consentData);
     if (!hasValidated && !hasGDPRConsent(consentData)) {
       utils.logWarn(`${MODULE_NAME} - gdpr permission not valid for local storage or cookies, exit module`);
       return;
     }
+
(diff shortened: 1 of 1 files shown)


patmmccann:
It appears the code block you moved is now in two places?

smenzer:
> It appears the code block you moved is now in two places?

That's right...it is used both in `getId()` and `refreshId()` methods

ChrisHuie (approved):

ncolletti (approved):
LGTM

modules/userId/index.js
```diff
@@ -590,16 +590,17 @@ function refreshUserIds(options, callback) {
   initializeSubmodulesAndExecuteCallbacks(function() {
     let consentData = gdprDataHandler.getConsentData()
 
-    const storedConsentData = getStoredConsentData();
-    setStoredConsentData(consentData);
-
     // gdpr consent with purpose one is required, otherwise exit immediately
     let {userIdModules, hasValidated} = validateGdprEnforcement(submodules, consentData);
     if (!hasValidated && !hasGDPRConsent(consentData)) {
```
dgirardi, line 595: Regarding https://github.com/prebid/Prebid.js/issues/6386#issuecomment-1402272770

If gdpr applies and there is no consent, here you get `{userIdModules: [], hasValidated: true}`, in other words, this check passes and we go ahead to write the cookie on line 255 - even though none of the userId modules will be initialized as they all failed the consent check.

PR #6558, opened 2021-04-09
Zemanta adapter: copy adapter as Outbrain

rokostik:
## Type of change

- [ ] Bugfix
- [ ] Feature
- [ ] New bidder adapter   
- [ ] Build related changes
- [ ] CI related changes
- [x] Other

## Description of change

~Hey, this renames the Zemanta adapter to Outbrain.~
EDIT: as discussed bellow in the comments, this is just a copy of the zemanta adapter until v5.0 to preseve backwards compatibility. We will remove the zemanta adapter for v5.0.
When we first made the decision to expose the Zemanta adapter as Outbrain we only supported prebid.js that this with minimal effort with the alias function. But now we also have a server adapter and noticed that it could cause confusion since aliases work a bit differently there so we decided to fully rename both adapters to Outbrain.

- Link to the PR on the docs repo at https://github.com/prebid/prebid.github.io/pull/2847

### DIFF
--- a/modules/outbrainBidAdapter.js
+++ b/modules/outbrainBidAdapter.js
@@ -1,0 +1,274 @@
+// jshint esversion: 6, es3: false, node: true
+'use strict';
+
+import {
+  registerBidder
+} from '../src/adapters/bidderFactory.js';
+import { NATIVE, BANNER } from '../src/mediaTypes.js';
+import * as utils from '../src/utils.js';
+import { ajax } from '../src/ajax.js';
+import { config } from '../src/config.js';
+
+const BIDDER_CODE = 'outbrain';
+const GVLID = 164;
+const CURRENCY = 'USD';
+const NATIVE_ASSET_IDS = { 0: 'title', 2: 'icon', 3: 'image', 5: 'sponsoredBy', 4: 'body', 1: 'cta' };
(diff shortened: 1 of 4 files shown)


patmmccann:
I think this would break existing integrations, do you agree?

If so, please make on the 5.0 branch, expected in May.

patmmccann:
Or maybe you could add outbrain and delete the alias from zemanta on 4, delete zemanta on 5?

rokostik:
Hey @patmmccann thanks for the comment. We currently have one integration still explicitly using zemanta in the configuration and we were planning to coordinate with them to change it until the next release. But I agree with you that it would be better to have both adapters until 5.0 and then have a clean cutoff.

I will prepare the changes as you proposed.

rokostik:
@patmmccann I pushed the changes. Regarding the deletion: do I just prepare a PR with the delete targeting the 5.0 branch or is there anything else to do?

patmmccann:
That's it ty

patmmccann:
I think you may want to remove the alias from the zemanta adapter on this pr

rokostik:
> I think you may want to remove the alias from the zemanta adapter on this pr

Yeah, of course, thanks!

patmmccann (approved):

PR #6567, opened 2021-04-11
SmileWanted Bid Adapter : Add GDPR and consent string to cookie sync call

MaxSmileWanted:
## Type of change

- [x] Feature

## Description of change
- Calling the cookie sync url even without ad responses.
- Add GDPR and consent string to cookie sync call

Updated doc : https://github.com/prebid/prebid.github.io/pull/2852


blingster7 (approved):
LGTM

PR #6581, opened 2021-04-13
ID5 User ID module - don't send empty fields to server

smenzer:
## Type of change

- [ ] Bugfix
- [ ] Feature
- [ ] New bidder adapter   
- [ ] Build related changes
- [ ] CI related changes
- [ ] Other

## Description of change
For optional fields, if they are empty don't send them to the server. Also add some logging to help with debugging

### DIFF
--- a/test/spec/modules/id5IdSystem_spec.js
+++ b/test/spec/modules/id5IdSystem_spec.js
@@ -505,7 +505,6 @@ describe('ID5 ID System', function() {
           let decoded = id5IdSubmodule.decode(ID5_STORED_OBJ, testConfig);
           expect(decoded).to.deep.equal(expectedDecodedObjectWithIdAbOff);
           sinon.assert.notCalled(logErrorSpy);
-          sinon.assert.notCalled(logInfoSpy);
         });
       });
 
@@ -523,7 +522,6 @@ describe('ID5 ID System', function() {
           testConfig.params.abTesting = testAbTestingConfig;
           id5IdSubmodule.decode(ID5_STORED_OBJ, testConfig);
           sinon.assert.notCalled(logErrorSpy);
-          sinon.assert.calledOnce(logInfoSpy);
         });
       });
     });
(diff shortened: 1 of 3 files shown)


test/spec/modules/id5IdSystem_spec.js
```diff
@@ -19,7 +19,7 @@ import * as utils from 'src/utils.js';
 
 let expect = require('chai').expect;
 
-describe('ID5 ID System', function() {
+describe.only('ID5 ID System', function() {
```
Fawke, line 22: I think you forgot to get rid of `.only`, because of this all test cases are not running.

Fawke (changes requested):
Hi @smenzer,

The change looks good. Can you just make sure that all test cases run properly by removing .only from your test case file.

As of now, I see few failures when I run `gulp test`.

Screenshot 2021-04-14 at 1 35 08 PM

test/spec/modules/id5IdSystem_spec.js
smenzer: whoops my bad...fixed now

smenzer:
> As of now, I see few failures when I run `gulp test`.

Yea my bad...I added some `logInfo()`'s after finishing my code and forgot to run tests again. should be working now, @Fawke

Fawke (approved):
LGTM

smenzer:
@Fawke any reason in particular this needs a second review? it's not core...

smenzer:
@Fawke or @ChrisHuie just wanted to check if this really needs a 2nd review, and if not, can we merge it?

Fawke:
@smenzer I guess we can merge it, sorry for the confusion. I read the PR Review guidelines and there's this clause:

> If all above is good, add a LGTM comment and, if the change is in PBS-core or is an important module like the prebidServerBidAdapter, request 1 additional core member to review.

It's kind of a grey area what constitutes an important module, but I think, this change's okay, since it's mostly minor enhancements and refactoring.

smenzer:
> It's kind of a grey area what constitutes an important module, but I think, this change's okay, since it's mostly minor enhancements and refactoring.

got it, thanks for pushing through, @Fawke

PR #6582, opened 2021-04-13
Ix bid adapter: resolve negative size bug

umakajan:
## Type of change

- [x] Bugfix
- [ ] Feature
- [ ] New bidder adapter   
- [ ] Build related changes
- [ ] CI related changes
- [ ] Other

## Description of change

- adding a check to ensure that `BASE_REQ_SIZE < MAX_REQ_SIZE`

### DIFF
--- a/modules/ixBidAdapter.js
+++ b/modules/ixBidAdapter.js
@@ -416,7 +416,11 @@ function buildRequest(validBidRequests, bidderRequest, impressions, version) {
       msd = impressions[transactionIds[i]].missingCount;
     }
 
-    trimImpressions(impressions[transactionIds[i]], MAX_REQ_SIZE - BASE_REQ_SIZE);
+    if (BASE_REQ_SIZE < MAX_REQ_SIZE) {
+      trimImpressions(impressions[transactionIds[i]], MAX_REQ_SIZE - BASE_REQ_SIZE);
+    } else {
+      utils.logError('ix bidder: Base request size has exceeded maximum request size.');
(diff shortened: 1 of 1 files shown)


modules/ixBidAdapter.js
```diff
@@ -416,7 +416,11 @@ function buildRequest(validBidRequests, bidderRequest, impressions, version) {
       msd = impressions[transactionIds[i]].missingCount;
     }
 
-    trimImpressions(impressions[transactionIds[i]], MAX_REQ_SIZE - BASE_REQ_SIZE);
+    if (BASE_REQ_SIZE < MAX_REQ_SIZE) {
```
patmmccann, line 419: Is this backwards? Trim if req > max ?

patmmccann (changes requested):
Please confirm this logic

modules/ixBidAdapter.js
umakajan: We want to make sure that the `BASE_REQ_SIZE` created above isn't larger than the `MAX_REQ_SIZE`, this fix is to ensure that we don't pass a [negative value to `trimImpressions` function](https://github.com/prebid/Prebid.js/pull/6582/files#diff-2597e133de10cd404cf919d82b772fa4a4ba33e49e9f4bdc4ba18a0a905774a5R559).

patmmccann (approved):
LGTM 

Thanks for confirming the logic, but still in draft state

PR #6584, opened 2021-04-13
Rubicon Analytics: Handle PBS Sending Bid ID 0

robertrmartinez:
## Type of change
- [X] Bugfix
- 
## Description of change
Prebid Server sometimes sends a bidId of 0. Which causes downstream analytics joins to fail.

This small change will overwrite these bidIds in this case.

### DIFF
--- a/modules/rubiconAnalyticsAdapter.js
+++ b/modules/rubiconAnalyticsAdapter.js
@@ -363,8 +363,8 @@ export function parseBidResponse(bid, previousBidResponse, auctionFloorData) {
       const height = bid.height || bid.playerHeight;
       return (width && height) ? {width, height} : undefined;
     },
-    'pbsBidId',
-    'seatBidId',
+    'pbsBidId', pbsBidId => pbsBidId == 0 ? utils.generateUUID() : pbsBidId,
+    'seatBidId', seatBidId => seatBidId == 0 ? utils.generateUUID() : seatBidId,
     'floorValue', () => utils.deepAccess(bid, 'floorData.floorValue'),
     'floorRuleValue', () => utils.deepAccess(bid, 'floorData.floorRuleValue'),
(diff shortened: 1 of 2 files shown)


modules/rubiconAnalyticsAdapter.js
```diff
@@ -363,8 +363,8 @@ export function parseBidResponse(bid, previousBidResponse, auctionFloorData) {
       const height = bid.height || bid.playerHeight;
       return (width && height) ? {width, height} : undefined;
     },
-    'pbsBidId',
-    'seatBidId',
+    'pbsBidId', pbsBidId => pbsBidId == 0 ? utils.generateUUID() : pbsBidId,
+    'seatBidId', seatBidId => seatBidId == 0 ? utils.generateUUID() : seatBidId,
```
robertrmartinez, line 367: So this logic would be as follows:

If set and is 0 or '0'

then overwrite it

else pass it along

If it is undefined (most often case) it will be ignored downstream and not set.

modules/rubiconAnalyticsAdapter.js
```diff
@@ -363,8 +363,8 @@ export function parseBidResponse(bid, previousBidResponse, auctionFloorData) {
       const height = bid.height || bid.playerHeight;
       return (width && height) ? {width, height} : undefined;
     },
-    'pbsBidId',
-    'seatBidId',
+    'pbsBidId', pbsBidId => pbsBidId == 0 ? utils.generateUUID() : pbsBidId,
```
bretg, line 366: Please a comment here

bretg (changes requested):

bretg (approved):

harpere (approved):

patmmccann (approved):

PR #6585, opened 2021-04-14
PBjs Core: bugfix for renderer not called when loading an external script

npeceniak:
- The renderer should be called reguardless of whether or not
          an external script was loaded.

## Type of change

- [X] Bugfix
- [ ] Feature
- [ ] New bidder adapter   
- [ ] Build related changes
- [ ] CI related changes
- [ ] Other

## Description of change

In Renderer.js the this.render function would only call the renderer if an external script was not loaded. 

## Other information

Issue seems to have been introduced in https://github.com/prebid/Prebid.js/pull/6422
@jeremiegirault Can you confirm that the runRender function should be called reguardless of whether an external script was loaded or not.

### DIFF
--- a/src/Renderer.js
+++ b/src/Renderer.js
@@ -54,8 +54,8 @@ export function Renderer(options) {
       loadExternalScript(url, moduleCode, this.callback);
     } else {
       utils.logWarn(`External Js not loaded by Renderer since renderer url and callback is already defined on adUnit ${adUnitCode}`);
-      runRender()
     }
+    runRender()
   }.bind(this) // bind the function to this object to avoid 'this' errors
 }
 


jeremiegirault:
@npeceniak The runRender is called in both cases but it will be called after the external script is loaded in case one is provided. If you make this change, it will be called twice (before and after script load)

npeceniak:
@jeremiegirault Thank you for the response. I have found a fix for our renderer that will work with this change. We were setting the renderer.url to "//" which caused the script to never fire onLoad so the callback would never get called. Changing our renderer.url to "/" seems to fire onLoad which fixes our issue. However do you think there should be some logic in there so that if the url never loads we still call the renderer after a timeout or something?

PR #6586, opened 2021-04-14
Ats optional 3p endpoint

mamatic:
## Type of change

- [ ] Feature

## Description of change
Change logic for firing 3P request. Change name of the config property from use3P to notUse3P because of reverse logic. By default 3P envelope endpoint will be triggered. If publisher do not wants to use 3P endpoint configuration property notUse3P should be set to true.

- Documentation changes:

https://github.com/prebid/prebid.github.io/pull/2868

@smenzer


smenzer (approved):
LGTM

PR #6593, opened 2021-04-14
Spotx Bid Adapter: Update endpoint to indicate request is from Prebid

npeceniak:
## Type of change
- [ ] Bugfix
- [X] Feature
- [ ] New bidder adapter   
- [ ] Build related changes
- [ ] CI related changes
- [ ] Other

## Description of change
Requests made through the spotxBidAdapter will now have `src_sys=prebid` appended as a query string parameter. This is for spotx to track where requests are originating from.


patmmccann:
Makes sense but I would add version

PR #6596, opened 2021-04-15
Permutive RTD Module: Add module name to storage manager and catch errors

dreischer:
## Type of change

- [x] Feature

## Description of change
* Adding the module name to this instance of `storageManager`
* Catching any errors that the module might throw


blingster7 (approved):
LGTM

Issue #6562, opened 2021-04-09
Create module registry


bretg:
## Type of issue
enhancement

## Description
We have a new scenario where our bid adapter needs to know whether or not our analytics adapter is in use on the page. There have been other scenarios where groups of modules work together or do something differently based on what other modules are present.

https://github.com/prebid/Prebid.js/issues/6117 innovated a step in this direction. When you run

```
gulp build --modules=uid2IdSystem,rubiconAnalyticsAdapter,rubiconBidAdapter,prebidServerBidAdapter,currency
```
It puts a comment in the resulting prebid.js file:
```
Updated : 2021-04-09
Modules: userId, uid2IdSystem, rubiconAnalyticsAdapter, rubiconBidAdapter, prebidServerBidAdapter, currency */
```

The proposal is that we drop the comments and instead create a machine readable format:

```
GLOBAL.installedModules = [userId, uid2IdSystem, rubiconAnalyticsAdapter, rubiconBidAdapter, prebidServerBidAdapter, currency]
```

Then modules can stop doing hacky things and just check the array of installedModules.

smenzer:
I love the idea. As long as the machine readable array is also easily readable by a human I'm fine dropping the comments...otherwise I'd like to leave them in as they're incredibly useful for debugging.

anthonylauzon:
I like this idea a lot. We might ultimately use this functionality with our id and rtd module at some point in the future.  Another nice to have that is somewhat related would be having place for vendor specific library code to go that might be shared between modules.  For instance, I imagine a scenario where an analytics adapter would need to potentially duplicate library code present in a bid adapter.  If the analytics adapter could check if the bid adapter were present and dynamically load the library code accordingly, that might prove helpful in certain circumstances.  I think I maybe have seen something similar to this in a bid adapter, but I can't recall exactly where.

PR #6601, opened 2021-04-15
Create module registry: update build to add installModules array to pbjs global

mmoschovas:
## Type of change

- [ ] Bugfix
- [ X] Feature
- [ ] New bidder adapter   
- [ X] Build related changes
- [ ] CI related changes
- [ ] Other

## Description of change

Update to allow array of modules to exist 
Create module registry:
- Remove module list comment on build
- Add pbjs.installedModules array with macro to src/prebid.js
- Update gruntfile to replace macro with module array (if empty array all modules included))

Issue: https://github.com/prebid/Prebid.js/issues/6562

```
{
  bidder: '<bidder name>',
  params: {
    // ...
  }
}
```

### DIFF
--- a/gulpfile.js
+++ b/gulpfile.js
@@ -215,10 +215,6 @@ function bundle(dev, moduleArr) {
 
   var outputFileName = argv.bundleName ? argv.bundleName : 'prebid.js';
 
-  const moduleList = helpers.getArgModules().map(module => {
-    return '"' + module + '"'
-  });
-
   // change output filename if argument --tag given
   if (argv.tag && argv.tag.length) {
     outputFileName = outputFileName.replace(/\.js$/, `.${argv.tag}.js`);
@@ -235,10 +231,6 @@ function bundle(dev, moduleArr) {
     .pipe(replace(/(Modules: )(.*?)(\*\/)/, ('$1' + getModulesListToAddInBanner(helpers.getArgModules()) + ' $3')))
     .pipe(gulpif(dev, sourcemaps.init({ loadMaps: true })))
     .pipe(concat(outputFileName))
-    // .pipe(gulpif(!argv.manualEnable, footer('\n<%= global %>.installedModules = <%= list %>', {
-    //   global: prebid.globalVarName,
-    //   list: "[" + moduleList + "];"
-    // })))
     .pipe(gulpif(!argv.manualEnable, footer('\n<%= global %>.processQueue();', {
       global: prebid.globalVarName,
     }
--- a/gulpfile.js
+++ b/gulpfile.js
@@ -134,6 +134,12 @@ function watch(done) {
   done();
 };
 
+function makeModuleList(modules) {
+  return modules.map(module => {
+    return '"' + module + '"'
+  });
+}
+
 function makeDevpackPkg() {
   var cloned = _.cloneDeep(webpackConfig);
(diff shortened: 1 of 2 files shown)


gulpfile.js
```diff
@@ -31,7 +31,7 @@ const execa = require('execa');
 
 var prebid = require('./package.json');
 var dateString = 'Updated : ' + (new Date()).toISOString().substring(0, 10);
-var banner = '/* <%= prebid.name %> v<%= prebid.version %>\n' + dateString + '\nModules: <%= modules %> */\n';
+var banner = '/* <%= prebid.name %> v<%= prebid.version %>\n' + dateString + '*/\n';
```
smenzer, line 34: is there a reason to drop this line? having the list of modules at the top of the prebid code is really useful when debugging code and looking at it as a human

smenzer:
@mmoschovas thanks for putting this together. could you show sample output of what the prod version of prebid.js would look like (no need to include the entire prebid.js code, just the banner and the part where your new code is relevant)?

gulpfile.js
mmoschovas: Based on the issue, we did not want to add more to the JS than necessary by having duplicate lists. So it was removed from the comment to live solely in the array.

mmoschovas:
@smenzer Sure thing. Please see below. The array can be found in the code immediately after the version. Since it is global, you can also add pbjs.installedModules right into the console to have the array returned. 

```
/* prebid.js v4.35.0-pre
Updated : 2021-04-16*/
...
A.bidderSettings = A.bidderSettings || {},
        A.libLoaded = !0,
        A.version = "v4.35.0-pre",
        O.logInfo("Prebid.js v4.35.0-pre loaded"),
        A.installedModules = ["rubiconBidAdapter", "rubiconAnalyticsAdapter", "consentManagement", "currency"],
        A.adUnits = A.adUnits || [],
        A.triggerUserSyncs = j;
...
```

gulpfile.js
smenzer: My read on the issue was to drop the string so that "machines" could use an array more intelligently than hacking some text; but the text is still important for humans. The way it's implemented now makes it harder for humans to find the list, although still better than prior to the banner being there.

@bretg do you have strong feelings on this? My vote would be to leave the banner AND add the array that was added here.

bretg: While I appreciate that the array is somewhat harder for a person to read, Prebid.js is constantly under fire for being larger than desired. An average publisher likely uses 5 bid adapters and 3-5 other modules, so the list of module names is likely between 150-200 bytes. Replicating that list seems to drive us away from the goal of making the build package smaller.

smenzer: While I don't personally think the few extra bytes is a big deal, I can appreciate that some pubs do. Can we at least add a util function that prints the list modules so I can just hit that in console rather than digging for the array in the source code?

mmoschovas: @smenzer I put in the other comment you can add pbjs.installedModules to the console to get the array printed.

smenzer: Ah you're right sorry I missed that.

smenzer (approved):

robertrmartinez (approved):

smenzer:
can we get this documented in the prebid docs somewhere, please?

PR #6604, opened 2021-04-15
ix bid adapter: send userIds in ixdiag

umakajan:
## Type of change

- [ ] Bugfix
- [x] Feature
- [ ] New bidder adapter   
- [ ] Build related changes
- [ ] CI related changes
- [x] Other

## Description of change

In this PR, we are collecting the `userIds` from the `bidRequest` and passing it on to your diagnostics object `ixdiag` that is part of our bid request URL (`cygnus`)

### DIFF
--- a/modules/ixBidAdapter.js
+++ b/modules/ixBidAdapter.js
@@ -482,6 +482,31 @@ function buildRequest(validBidRequests, bidderRequest, impressions, version) {
 }
 
 /**
+ * Return an object of user IDs stored by Prebid User ID module
+ *
+ * @returns {Object} ID providers and whether they are present
+ */
+function _getUserIds(bidRequest) {
+  const userIds = bidRequest.userId || {};
+
+  return {
+    'britepoolid': utils.deepAccess(userIds, 'britepoolid') ? 1 : 0,
+    'id5id': utils.deepAccess(userIds, 'id5id') ? 1 : 0,
+    'lipbid': utils.deepAccess(userIds, 'lipb.lipbid') ? 1 : 0,
+    'haloId': utils.deepAccess(userIds, 'haloId') ? 1 : 0,
+    'criteoId': utils.deepAccess(userIds, 'criteoId') ? 1 : 0,
+    'lotamePanoramaId': utils.deepAccess(userIds, 'lotamePanoramaId') ? 1 : 0,
+    'merkleId': utils.deepAccess(userIds, 'merkleId') ? 1 : 0,
+    'parrableId': utils.deepAccess(userIds, 'parrableId') ? 1 : 0,
+    'connectid': utils.deepAccess(userIds, 'connectid') ? 1 : 0,
+    'sharedid': utils.deepAccess(userIds, 'sharedid') ? 1 : 0,
+    'tapadId': utils.deepAccess(userIds, 'tapadId') ? 1 : 0,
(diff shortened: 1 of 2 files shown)


patmmccann:
This appears a different approach than the guidance from you guys at https://github.com/prebid/Prebid.js/pull/5436/files#r450323272 ?

umakajan:
@patmmccann that function is used to pass in `userEids` that are relevant to our ad server in the `/cygnus` request. The purpose of this change is to capture all the userId modules configured on the publisher page within our diagnostics object.

modules/ixBidAdapter.js
```diff
@@ -481,6 +481,31 @@ function buildRequest(validBidRequests, bidderRequest, impressions, version) {
...
+    'lotamePanoramaId': utils.deepAccess(userIds, 'lotamePanoramaId') ? 1 : 0,
+    'merkleId': utils.deepAccess(userIds, 'merkleId') ? 1 : 0,
+    'parrableId': utils.deepAccess(userIds, 'parrableId') ? 1 : 0,
+    'connectid': utils.deepAccess(userIds, 'connectid') ? 1 : 0,
+    'sharedid': utils.deepAccess(userIds, 'sharedid') ? 1 : 0,
+    'tapadId': utils.deepAccess(userIds, 'tapadId') ? 1 : 0,
+    'quantcastId': utils.deepAccess(userIds, 'quantcastId') ? 1 : 0,
+    'pubcid': utils.deepAccess(userIds, 'pubcid') ? 1 : 0
+  }
+}
```
smenzer, line 507: is there a reason you're not using the `userIdAsEids` object? There are several id providers that have extensions in their id that DSPs are expecting to receive. Also, some ids are objects not strings, so doing this you would have to ensure you parse the id object properly.

modules/ixBidAdapter.js
```diff
@@ -481,6 +481,31 @@ function buildRequest(validBidRequests, bidderRequest, impressions, version) {
...
+ * Return an object of user IDs stored by Prebid User ID module
+ *
+ * @returns {Object} ID providers and whether they are present
+ */
+function _getUserIds(bidRequest) {
+  const userIds = bidRequest.userId || {};
+
+  return {
+    'britepoolid': utils.deepAccess(userIds, 'britepoolid') ? 1 : 0,
+    'id5id': utils.deepAccess(userIds, 'id5id') ? 1 : 0,
```
smenzer, line 494: specifically for ID5, this will provide back an object that includes the ID5 ID and our extension (which includes the `linkType` field that DSPs are expecting).

modules/ixBidAdapter.js
patmmccann: They don't seem to grab the id, they bizarrely just monitor for the presence of the id. This is very strange as you might as well get it at that point,  but perhaps is related to the character limits they've introduced on themselves by preferring a get to a post.

patmmccann: Ix: why waste all these characters on zeros instead of just returning the array of configured ids?

umakajan: @smenzer @patmmccann thanks for the feedback, I've updated the PR

patmmccann (approved):

robertrmartinez (approved):

Issue #6483, opened 2021-03-26
[Question] Prebid + AMP: Possible fix to refererInfo infinite loop


fasenderos:
Hi, I had problems with Prebid and AMP. I can see from the terminal that `prebid.js` is loaded, but no request is sent to `appnexus`. After reading your docs, I discovered that AMP restricts header bidding wrappers like Prebid.js and that I have to create a (or to sign up for a hosted solution of) `Prebid Server` and using the AMP RTC Config.

Anyway, it seemed really strange to me that `prebid.js` is correctly loaded on page and nothing happens without raising any errors. After a lot of debugging I found out why. Prebid ends in an endless loop and AMP kills the process.

The loop is generated by the function `refererInfo()` in [refererDetection.js](https://github.com/prebid/Prebid.js/blob/42182d6c1844ebc012f3536da3670bc14f75b5db/src/refererDetection.js#L164
) because under AMP `currentWindow` is always different than `win.top`.
https://github.com/prebid/Prebid.js/blob/42182d6c1844ebc012f3536da3670bc14f75b5db/src/refererDetection.js#L164

I understend that this function try to traverse the DOM to find the windows top, but I don't know if it was intended to have a `do while` without an escape strategy (e.g. max 100 iterations)

In my case the `foundReferrer` pushed in the `stack` array was for the 1st iteration the `ampproject` url, than the url of my site and then `null`. And so on following the same pattern 

```
stack = [
   "https://ampproject.org",
   "https://mysite.com",
    null,
   "https://ampproject.org,
   "https://mysite.com",
    null,
   "https://ampproject.org",
   "https://mysite.com",
    null,
    ...
    ...
    ...
]
```
So to solve this specific problem it was enough to force the exit from the loop at the 2nd iteration `while (currentWindow !== win.top && level < 2)` - I know i can make a better condition, but this is not the place to talk about that.

After that prebid started working as usual, and the `referrer_detection.rd_ref` sent to appnexus has the right referer.

My concerns now are:
1. Is there a way to override Prebid functions without edit the core code?
2. Are you aware of any cases where are requried more than 2 iterations? I can think only Iframe that is inside an Iframe, that is inside an Iframe etc...
3. Are these common scenarios?
4. This change can have any drawbacks in other part of `prebid.js`?

Thanks in advance.

gglas:
[...]

PR #6615, opened 2021-04-18
PBjs Core: added `maxNestedIframes` option

fasenderos:
## Type of change
- [X] Bugfix
- [X] Feature

## Description of change
Added `maxNestedIframes` option to avoid endless loop on referer detection #6483.


fasenderos:
I'm not sure about the default value of `maxNestedIframes`. I initially set it at `2` because in my case fix the issue #6483 and let Prebid works under AMP without AMP RTC, but it breaks some tests of `refererDetection_spec.js`.

`3` was enough to not break tests, but since I don't want to introduce braking changes I have chosen, as @bretg [suggested here](https://github.com/prebid/Prebid.js/issues/6483#issuecomment-821992823), a larger number, `10`.

Now I'm thinking that it's better to choose as trade off a number that is large enough to not break client applications and also multiple of 3 less 1 (e.g. 5 - 8 - 11 - 14 etc.) to fix Prebid under AMP.

If you have any hints I can change the default value accordingly.
Thanks.

patmmccann:
This pr is fascinating because we were unaware prebid js could work in amp with such a minor change. All doc points to implementing a call to prebid server in amp.

Can you point us to a page where you've gotten pbjs working on amp?

fasenderos:
I'm always talking about _"my case"_ because we have a custom implementation of Prebid (see after), so I don't want to raising false expectations :). Maybe it's just a pure fix of my specific case.

I'm working for an Advertising Platform and all of our publisher install our `widget` on their page and we delivery ONLY `native` creatives. At first the widget call our server to check if programmatic is enabled. If so, the `prebid.js` is loaded on page and the `adUnits` object is appended to the `head` with our custom adapter (`nativeryBidAdapter.js`) and `appnexus`. We don't have an Ad Server, and for this reason I have created a Prebid module ( `nativeryWrapper.js` ) that has an event listner on `AUCTION_END`, and here I'm using `getHighestCpmBids` on each ad unit code and fire the impression tracker with `postMessage`. Since our widget support only native creatives, once I get the list of the winning ads, I have to format the template for the widget output.

[Here](https://www.salernonotizie.it/2021/04/19/imprese-balneari-campane-pronte-a-riaprire-lottimismo-della-fiba/amp/) is a live example.

Rothalack (approved):
This looks good

PR #6617, opened 2021-04-19
Zemanta bid adapter: reinstate Outbrain as alias

rokostik:
## Type of change

- [ ] Bugfix
- [ ] Feature
- [ ] New bidder adapter   
- [ ] Build related changes
- [ ] CI related changes
- [X] Other

## Description of change

Hey!

In https://github.com/prebid/Prebid.js/pull/6558 (released in v4.35.0) we copied the zemanta adapter as the outbrain and made outbrain a standalone non-aliased adapter. We left the zemanta adapter to be backwards compatible until prebid.js v5.

The issue we are now facing is that if anyone wants to download the outbrain adapter from https://docs.prebid.org/download.html using a version less than v4.35.0 there's an error since the library generation api is called with `modules[]: outbrainBidAdapter` that of course doesn't exist in previous versions.

This PR again makes the outbrain adapter an alias for zemanta and adds some changes that were already merged from outbrain adapter. The standalone outbrain adapter still exists and any future changes will be applied to both adapters. The zemanta adapter is already removed on the v5 branch (https://github.com/prebid/Prebid.js/pull/6560).

As far as I've tested, I see that with this setup bid requests for outbrain are handled by the zemanta adapter. I think this is because in https://github.com/prebid/Prebid.js/blob/master/src/adapters/bidderFactory.js#L154 zemanta alphabetically comes after outbrain and overrides the existing adapter. Since both adapters will be kept in sync I think this behaviour is ok.
Are there any potential issues here that I missed?

- PR on the docs repo https://github.com/prebid/prebid.github.io/pull/2880

### DIFF
--- a/modules/zemantaBidAdapter.js
+++ b/modules/zemantaBidAdapter.js
@@ -167,7 +167,11 @@ export const spec = {
     return syncs;
   },
   onBidWon: (bid) => {
-    ajax(utils.replaceAuctionPrice(bid.nurl, bid.originalCpm))
+    // for native requests we put the nurl as an imp tracker, otherwise if the auction takes place on prebid server
+    // the server JS adapter puts the nurl in the adm as a tracking pixel and removes the attribute
+    if (bid.nurl) {
+      ajax(utils.replaceAuctionPrice(bid.nurl, bid.originalCpm))
+    }
   }
 };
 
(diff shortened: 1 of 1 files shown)


patmmccann:
LGTM 

since you own both adapters and one is already deleted in 5.0, the temporary conflict seems fine

Issue #3515, opened 2019-02-05
PrebidServer Default Vendor option does not honor `enabled` field


robertrmartinez:
## Type of issue
BUG - Unsure of when introduced, or if has been the case since default vendors was included.

## Description
When using the `defaultVendor` option in `s2sConfig`, we cannot overwrite the `enabled` field.
The auction will ALWAYS hit PBS when the defaultVendor is set and cannot be shut off with `enabled: false`

## Steps to reproduce
Any s2s test page with the following s2sconfig will show this behavior:

```javascript
                s2sConfig: {
                    accountId: 1001,
                    defaultVendor: "rubicon",
                    bidders: ['rubicon'],
                    enabled: false,
                },
```

The above will still request rubicons Prebid Server for any Rubicon Bids.

Workaround is to just empty out the bidders list.

## Test page
https://jsfiddle.net/mvyurf3o/3/
I would expect this test page to make a Client Side request to Rubicon, not a Server Side!

https://jsfiddle.net/zLe2jnad/1/
This test page showing that if I remove `defaultVendor` and just put an `endpoint`, then it does correctly disable the server connection.

### Expected results
WHEN `defaultVendor` is set and `enabled` is false
THEN no server connection is made.

### Actual results
enabled is ignored and is whatever the default vendor has set.

bretg:
Good catch @robertrmartinez - will put it in our queue for next sprint

mkendall07:
@bretg @robertrmartinez 
I believe this is fixed now? can you confirm?

robertrmartinez:
@mkendall07 

Is there a link to the PR?

I tried with Prebid `2.32.0` and it still does not work.

robertrmartinez:
@jaiminpanchal27 Was this actually fixed or just still waiting to be fixed?

I do not think it is super high priority but wanted to make sure cuz I could not find an associated PR

jaiminpanchal27:
@robertrmartinez Not fixed. Thanks for taking it.

gglas:
@robertrmartinez to review today and close or plan next steps

patmmccann:
Solved in #6622

PR #6622, opened 2021-04-19
PBS Bid Adapter: Stop overriding s2sconfig.enabled from vendor defaults

patmmccann:
## Type of change

- [X] Bugfix
- [ ] Feature
- [ ] New bidder adapter   
- [ ] Build related changes
- [ ] CI related changes
- [ ] Other

## Description of change

I believe this solves https://github.com/prebid/Prebid.js/issues/3515 as reported by @robertrmartinez


robertrmartinez (approved):
haha! Nice Patrick thanks. I was eventually going to look into this I swear

ChrisHuie (approved):

modules/prebidServerBidAdapter/index.js
```diff
@@ -89,7 +89,7 @@ config.setDefaults({
  * @return {boolean}
  */
 function updateConfigDefaultVendor(option) {
-  if (option.defaultVendor) {
+  if (option.defaultVendor && option.enabled !== false) {
```
robertrmartinez, line 92: If s2sConfig does not explicitly state enabled === true then it always defaults to false (even though defaultVendor json option has it to true).
It is being overwritten before this line by the defaultS2sConfig here: https://github.com/prebid/Prebid.js/blob/master/modules/prebidServerBidAdapter/index.js#L75

I should have caught this during my PR review, sorry @patmmccann

patmmccann: You are saying if someone selects defaultVendor and sets enabled = false the override is still occuring? I thought https://github.com/prebid/Prebid.js/blob/87ff2b8d0ba030c54b19056ccc5227b4baa1a916/modules/prebidServerBidAdapter/index.js#L100 was the line that did the overwrite and this change would prevent that from executing

robertrmartinez: No, the opposite actually.

Now it is ignoring that defaultVendor.enabled flag.

So if you have this:

```js
{
   defaultVendor: 'rubicon',
   bidders: ['rubicon']
}
```

It will have `enabled: false`

So you have to explicitly do:

```js
{
   enabled: true,
   defaultVendor: 'rubicon',
   bidders: ['rubicon']
}
```


I have not pinpointed exact spot yet since I am busy on something else, but I should be able to add a PR to address later this week.

PR #6233, opened 2021-01-27
Tappx Bid Adapter: new bidder adapter added

prebidtappx:
## Type of change
- [x] New bidder adapter

## Description of change
Add new Tappx adapter.

- Test parameters for validating bids
```
{
  bidder: 'tappx',
  params: {
    host: "testing.ssp.tappx.com/rtb/v2/",
    tappxkey: "pub-1234-android-1234",
    endpoint: "ZZ1234PBJS",
    bidfloor: 0.005
  }
}
```

- Contact email of the adapter’s maintainer: prebid@tappx.com
- [x] official adapter submission
- Link to PR on the docs repo: https://github.com/prebid/prebid.github.io/pull/2661

PR #6626, opened 2021-04-20
Tappx Bid Adapter: add floor module support

prebidtappx:
## Type of change

- [ **x**] Feature

## Description of change

**This change is to accept the priceFloors module that you recommend to use in the #6233 merge.
Now, the user can decide or use our bidfloor or prebid.js priceFloors module. (always have priority the prebid module)**

```
{
  bidder: '<bidder name>',
  params: {
    // ...
  }
}
```

## Other information

**- We add the posibility of use the bidFloor module like aleksatr comment in the #6233 merge.**

### DIFF
--- a/modules/tappxBidAdapter.js
+++ b/modules/tappxBidAdapter.js
@@ -124,6 +124,8 @@ function interpretBid(serverBid, request) {
     netRevenue: true,
   }
 
+  if (typeof serverBid.dealId != 'undefined') { bidReturned.dealId = serverBid.dealId }
+
   if (typeof request.bids.mediaTypes != 'undefined' && typeof request.bids.mediaTypes.video != 'undefined') {
     bidReturned.vastXml = serverBid.adm;
     bidReturned.vastUrl = serverBid.lurl;
@@ -134,7 +136,7 @@ function interpretBid(serverBid, request) {
     bidReturned.mediaType = BANNER;
   }
 
-  if (bidReturned.adomain != undefined || bidReturned.adomain != null) {
+  if (typeof bidReturned.adomain != 'undefined' || bidReturned.adomain != null) {
     bidReturned.meta = { advertiserDomains: request.bids.adomain };
   }
 
--- a/modules/tappxBidAdapter.js
+++ b/modules/tappxBidAdapter.js
@@ -282,7 +282,6 @@ function buildOneRequest(validBidRequests, bidderRequest) {
   bidder.bidfloor = BIDFLOOR;
   bidder.ext = (typeof BIDEXTRA == 'object') ? BIDEXTRA : undefined;
 
-
   imp.ext = {};
   imp.ext.bidder = bidder;
   // < Imp object
--- a/modules/tappxBidAdapter.js
+++ b/modules/tappxBidAdapter.js
@@ -8,7 +8,7 @@ import { config } from '../src/config.js';
 const BIDDER_CODE = 'tappx';
 const TTL = 360;
 const CUR = 'USD';
-const TAPPX_BIDDER_VERSION = '0.1.10329';
+const TAPPX_BIDDER_VERSION = '0.1.10413';
 const TYPE_CNN = 'prebidjs';
 var HOST;
 var hostDomain;
--- a/modules/tappxBidAdapter.js
+++ b/modules/tappxBidAdapter.js
@@ -264,13 +264,18 @@ function buildOneRequest(validBidRequests, bidderRequest) {
   imp.bidfloor = utils.deepAccess(validBidRequests, 'params.bidfloor');
   if (utils.isFn(validBidRequests.getFloor)) {
     try {
-      imp.bidfloor = validBidRequests.getFloor({
+      let floor = validBidRequests.getFloor({
         currency: CUR,
         mediaType: '*',
         size: '*'
-      }).floor;
+      });
+      if (utils.isPlainObject(floor) && !isNaN(floor.floor) && floor.currency === 'USD') {
+        imp.bidfloor = floor.floor;
+      } else {
+        utils.logWarn('Tappx bid adapter: ', 'Currency not valid. Use only USD with Tappx.');
+      }
     } catch (e) {
-      utils.logWarn(e);
+      utils.logWarn('Tappx bid adapter: ', e);
       imp.bidfloor = utils.deepAccess(validBidRequests, 'params.bidfloor'); // Be sure that we have an imp.bidfloor
     }
   }
--- a/modules/tappxBidAdapter.js
+++ b/modules/tappxBidAdapter.js
@@ -134,6 +134,10 @@ function interpretBid(serverBid, request) {
     bidReturned.mediaType = BANNER;
   }
 
+  if (bidReturned.adomain != undefined || bidReturned.adomain != null) {
+    bidReturned.meta = { advertiserDomains: request.bids.adomain };
+  }
+
   return bidReturned;
 }
 
--- a/modules/tappxBidAdapter.js
+++ b/modules/tappxBidAdapter.js
@@ -124,15 +124,12 @@ function interpretBannerBid(serverBid, request) {
     netRevenue: true,
   }
 
-  if (serverBid.video) {
-    utils.logMessage('--- VIDEO');
+  if (request.bids.mediaTypes.video) {
     bidReturned.vastXml = serverBid.adm;
     bidReturned.vastUrl = serverBid.lurl;
     bidReturned.ad = serverBid.adm;
     bidReturned.mediaType = VIDEO;
-  }
-  if (serverBid.banner) {
-    utils.logMessage('--- BANNER');
+  } else {
     bidReturned.ad = serverBid.adm;
     bidReturned.mediaType = BANNER;
   }
@@ -310,7 +307,6 @@ function buildOneRequest(validBidRequests, bidderRequest) {
   payload.params = params;
   payload.regs = regs;
   // < Payload
-
 
   return {
     method: 'POST',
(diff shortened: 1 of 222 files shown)


prebidtappx:
Hi!
I don't touch the file modules/outbrainBidAdapter.js, don't merge that part I think that was a automatic merge conflict resolution.
Please ignore that part of the pull-request.

patmmccann:
Just undo the changes to outbrain in your branch,  or re-pull master

robertrmartinez:
@prebidtappx As @patmmccann mentioned,

You will need to pull latest from master and reset your branch in order to remove that file edit.

Please do so

prebidtappx:
> @prebidtappx As @patmmccann mentioned,
> 
> You will need to pull latest from master and reset your branch in order to remove that file edit.
> 
> Please do so

Hi! Sorry for the late. I make now a pull and a push.

prebidtappx:
> Just undo the changes to outbrain in your branch, or re-pull master

Finally I undo manual and make a push. (c9ce240)
It seem that it work.

modules/tappxBidAdapter.js
```diff
@@ -261,11 +261,21 @@ function buildOneRequest(validBidRequests, bidderRequest) {
   imp.secure = 1;
 
   imp.bidfloor = utils.deepAccess(validBidRequests, 'params.bidfloor');
+  if (utils.isFn(validBidRequests.getFloor)) {
+    let sizes = (bannerMediaType !== undefined) ? bannerMediaType.sizes : videoMediaType.sizes;
+    sizes = (sizes !== undefined) ? sizes : '*';
+    imp.bidfloor = validBidRequests.getFloor({
+      currency: CUR,
+      bannerMediaType,
+      size: sizes
```
robertrmartinez, line 270: The Price Floors module can be implemented on a per size floor, thus, this size param getFloor expects a single size array like so:

`[300, 250]`

You are passing it all the sizes which won't break it, but will just lead to it always picking the '*' floor.

If you want to get all the floors on per size, you can run getFloor for each size in your sizes array.

Then decide how you want to pass to your endpoint (maybe highest floor or smallest?)

Otherwise, if you do not wish to do this, you can just pass the '*' as the "size" param (or leave it out, same thing)



As for the `bennerMediaType` param you are passing.

This is expected to be a string representation of the mediaType.

You are passing the actual mediaTypeObject.

`banner` or `video` or `native` should be passed here.

Otherwise `*`

Here is a screenshot using your example:

image


Also, it is reccomended to be a little safe when just assuming that the response from getFloor returns the object with the `.floor` attribute.

If not floor is found, your code here may throw an exception.

See how other adapters have surrounded this in a try / catch or even just validated the response has format expected.

robertrmartinez (changes requested):
Sending in some params in a format the price floors module does not expect, please fix.

Also, please add unit test or two to cover this case.

modules/tappxBidAdapter.js
prebidtappx: Hi! We update with the try/catch and change all with '*'.
We were to made some changes, but we just have problems to make a fake request with a priceFloor module to make some tests for this part.

prebidtappx: About the problem with testing...
We try to import the functions of priceFloors:
`import { handleSetFloorsConfig, getFloor } from 'modules/priceFloors';`
And create a configuration:
`const basicFloorConfig = {
      enabled: true,
      data: {
        // default if endpoint doesn't return in time
        floorProvider: 'floorProviderName',
        currency: 'USD',
        skipRate: 5,
        modelVersion: 'new model 1.0',
        schema: {
          fields: [ 'mediaType' ]
        },
        values: {
          'banner': 0.01,
          'video': 2.00
        }
      }
    }
    handleSetFloorsConfig({...basicFloorConfig, floorProvider: 'floorprovider'})`
But we don't know how to pass this information to a test fake bidrequest.
Any idea or trick to test this part?
Thanks for all.

robertrmartinez: Hi @prebidtappx 

Please see how the rubicon bidAdapter mocked the getFloor response:

https://github.com/prebid/Prebid.js/blob/master/test/spec/modules/rubiconBidAdapter_spec.js#L444

Note that we simply mock the responses and put a mock getFloor function onto the bidRequests.

See higher up how this mock is done where we use a global `getFloorResponse` object which we use to control what the mock function returns

https://github.com/prebid/Prebid.js/blob/master/test/spec/modules/rubiconBidAdapter_spec.js#L300

See this line here how we add it on the bidRequest:

https://github.com/prebid/Prebid.js/blob/master/test/spec/modules/rubiconBidAdapter_spec.js#L446

Then throughout the test, we just update the global getFloorResponse var to be what we want to test and make sure our adapter handles correctly.

prebidtappx: Thanks for all!
We make a push with the testing.

modules/tappxBidAdapter.js
```diff
@@ -262,11 +262,24 @@ function buildOneRequest(validBidRequests, bidderRequest) {
   imp.secure = 1;
 
   imp.bidfloor = utils.deepAccess(validBidRequests, 'params.bidfloor');
+  if (utils.isFn(validBidRequests.getFloor)) {
+    try {
+      imp.bidfloor = validBidRequests.getFloor({
+        currency: CUR,
+        mediaType: '*',
+        size: '*'
+      }).floor;
```
robertrmartinez, line 271: There is a scenario where somone has floors in a currency X

You ask for it in currency Y

But the currency conversion module is not loaded.

The getFloor function will return the floor in the currency X

And you will pass it along like so.

So, if you would like to be a bit more careful, most adapters check the responding currency is in the one they expected before sending it to their exchange endpoint.

Here is an example:

https://github.com/prebid/Prebid.js/blob/master/modules/appnexusBidAdapter.js#L1054-L1061

modules/tappxBidAdapter.js
```diff
@@ -262,11 +262,24 @@ function buildOneRequest(validBidRequests, bidderRequest) {
...
+    } catch (e) {
+      utils.logWarn(e);
```
robertrmartinez, line 273: Could you add a prefix here so users are more easy to debug when this error is thrown:

utils.logWarn('tappx bid adapter: ', e);

Something like this.

robertrmartinez (changes requested):
small question / suggestion

modules/tappxBidAdapter.js
prebidtappx: Thanks!
We used the example for the change.
(We make the push of this change)

modules/tappxBidAdapter.js
prebidtappx: Hi!!!
I use finally use:
`utils.logWarn('[TAPPX]: ', e);`
to use the same prefix than my teammate.
Thanks for all!
(We made the push of this change)

prebidtappx:
Hi!
circleci detect this error related to other adapter...
How can I fix this to merge this changes?
Thanks for all!
image

test/spec/modules/tappxBidAdapter_spec.js
```diff
@@ -1,5 +1,6 @@
 import { assert } from 'chai';
 import { spec } from 'modules/tappxBidAdapter';
+import { getFloor } from 'modules/priceFloors';
```
robertrmartinez, line 3: please remove this, unneeded import, you are mocking this module.

test/spec/modules/tappxBidAdapter_spec.js
```diff
@@ -118,6 +119,7 @@ const c_BIDDERREQUEST_B = {'bidderCode': 'tappx', 'auctionId': '13a8a3a9-ed3a-41
...
 
 describe('Tappx bid adapter', function () {
+  let getFloorResponse = {};
```
robertrmartinez, line 122: Think this should be moved down into the `module floor implementation` describe block

test/spec/modules/tappxBidAdapter_spec.js
```diff
@@ -293,4 +295,39 @@ describe('Tappx bid adapter', function () {
       expect(consent[0].url).to.match(/&type=iframe/);
     });
   })
+
+  describe('module Floor implementation', function() {
+    let bidderRequest_f = c_BIDREQUEST;
```
robertrmartinez, line 300: set your vars like so

```
  let bidderRequest_f,
    getFloorResponse;
```

Then make them whatever your default assignment is inside a beforeEach block like so:

https://github.com/prebid/Prebid.js/blob/master/test/spec/modules/rubiconBidAdapter_spec.js#L20-L24

https://github.com/prebid/Prebid.js/blob/master/test/spec/modules/rubiconBidAdapter_spec.js#L300-L301

This helps stablilize tests if something messes with these vars or tests go out of order for some reason.

robertrmartinez (changes requested):
I _think_ something haywire is going on with the circle CI tests, and it may work better if you make the suggested changes here

test/spec/modules/tappxBidAdapter_spec.js
prebidtappx: Done! : )

test/spec/modules/tappxBidAdapter_spec.js
prebidtappx: Done, moved to "module floor implementation"!
 : )

test/spec/modules/tappxBidAdapter_spec.js
prebidtappx: Change of variables getFloorResponse and bidderRequest_f and added the beforeEach block.
: )

prebidtappx:
I make all the changes, but we still having the same problem or very similar.
: (

image

robertrmartinez:
@prebidtappx 

Will you please pull latest from master and push?

We pushed some test fixes for other adapters.

robertrmartinez (approved):

PR #6629, opened 2021-04-20
UserId Module: add support for case insensitive submodule names

jsnellbaker:
## Type of change

- [x] Feature

## Description of change
This change allows case insensitivity for `userIds[].name` when defined in the setConfig.  

When this situation occurs, the logic will internally update the config's `name` to the format as defined by the submodule name (so that the submodule sees the value they expect, in rare case they have any logic reliant on the config values).

### DIFF
--- a/modules/userId/index.js
+++ b/modules/userId/index.js
@@ -755,7 +755,8 @@ function updateSubmodules() {
 
   // find submodule and the matching configuration, if found create and append a SubmoduleContainer
   submodules = addedSubmodules.map(i => {
-    const submoduleConfig = find(configs, j => j.name === i.name);
+    const submoduleConfig = find(configs, j => j.name && j.name.toLowerCase() === i.name.toLowerCase());
+    if (submoduleConfig && i.name !== submoduleConfig.name) submoduleConfig.name = i.name;
     i.findRootDomain = findRootDomain;
     return submoduleConfig ? {
       submodule: i,
--- a/test/spec/modules/userId_spec.js
+++ b/test/spec/modules/userId_spec.js
@@ -513,6 +513,20 @@ describe('User ID', function () {
       expect(utils.logInfo.args[0][0]).to.exist.and.to.contain('User ID - usersync config updated for 1 submodules');
     });
 
+    it('handles config with name in different case', function () {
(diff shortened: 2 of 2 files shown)


test/spec/modules/userId_spec.js
```diff
@@ -513,6 +513,20 @@ describe('User ID', function () {
       expect(utils.logInfo.args[0][0]).to.exist.and.to.contain('User ID - usersync config updated for 1 submodules');
     });
 
+    it('handles config with name in different case', function () {
+      setSubmoduleRegistry([pubCommonIdSubmodule, unifiedIdSubmodule, id5IdSubmodule, identityLinkSubmodule, merkleIdSubmodule, netIdSubmodule, nextrollIdSubmodule, sharedIdSubmodule, intentIqIdSubmodule, zeotapIdPlusSubmodule, pubProvidedIdSubmodule, criteoIdSubmodule, mwOpenLinkIdSubModule, tapadIdSubmodule, uid2IdSubmodule, admixerIdSubmodule, deepintentDpesSubmodule]);
```
smenzer, line 517: I don't think it's necessary to add all submodules here, right? Only one we really need is `criteoIdSubmodule` if that's the one you're testing with. I just don't want id providers to think they have to add their module here. 

Better yet - use a mock id module and no one will feel left out ;-)

modules/userId/index.js
```diff
@@ -755,7 +755,8 @@ function updateSubmodules() {
 
   // find submodule and the matching configuration, if found create and append a SubmoduleContainer
   submodules = addedSubmodules.map(i => {
-    const submoduleConfig = find(configs, j => j.name === i.name);
+    const submoduleConfig = find(configs, j => j.name && j.name.toLowerCase() === i.name.toLowerCase());
+    if (submoduleConfig && i.name !== submoduleConfig.name) submoduleConfig.name = i.name;
```
smenzer, line 759: should we throw a warning/info message here to let the publisher know they've used the wrong case? or since it's solved now it doesn't really matter?

test/spec/modules/userId_spec.js
jsnellbaker: No worries, I was just following the formatting used by the surrounding tests.  Updated.

modules/userId/index.js
jsnellbaker: I can if you really want.  Let me know either way.

smenzer: I guess there's really no impact to using the wrong case now, so no real reason to spit out a log message. if i change my mind later, we can always add it then...

patmmccann (approved):

smenzer:
given two approvals, i'm merging this. thanks @jsnellbaker for putting this in, good idea!

Issue #6433, opened 2021-03-17
Upgrade User ID Module Privacy Support


bretg:
## Type of issue
Enhancement

## Description
User ID modules cannot easily get the CCPA consent string. When we built the RealTimeData infrastructure, we made it easy for all RTD sub-modules to obtain all known consent data - every RTD internal interface includes a `userConsent` data structure so each one doesn't have to dig them up itself.

We should upgrade the User ID module to do this same thing.

1) Add a function in the core User ID module similar to the RTD getConsentData() function:

```
function getConsentData() {
  return {
    gdpr: gdprDataHandler.getConsentData(),
    usp: uspDataHandler.getConsentData(),
    coppa: !!(config.getConfig('coppa'))
  }
}
```

2) Update the existing consentData data structure used by UserID-core to call sub-modules

3) Confirm that every sub-module will not choke when receiving these parameters

4) Upgrade SharedId sub-module to make use of these parameters:

patmmccann:
Our approach  has been to set the device access flag on opt out, which takes care of this problem. If this is solved, would there be any remaining need for deviceAccess?

patmmccann:
One note, this functionality may already exist, eg https://github.com/prebid/Prebid.js/blob/99098ecc6ee7c037bef92a709f491e50160511bb/modules/parrableIdSystem.js#L193

smenzer:
I can see the value in providing all relevant privacy jurisdiction fields in a single function/object like @bretg mentions, but I agree with @patmmccann that this is already partially implemented; we are also retrieving the CCPA uspString already: https://github.com/prebid/Prebid.js/blob/99098ecc6ee7c037bef92a709f491e50160511bb/modules/id5IdSystem.js#L121

gglas:
@bretg are there pressing implementation issues for this, or should we just move it into the backlog?

bretg:
Yes, I think this needs to happen, specifically because SharedId and PubCommon should be paying attention to these values. Adding @jdwieland8282

smenzer:
@bretg maybe I’m missing something but can’t these values already be retrieved by shared Id and pubcommon? We are ingesting them in our module already as do a few others...

patmmccann:
Is https://github.com/prebid/Prebid.js/issues/6673 a duplicate @smenzer ?

smenzer:
> Is https://github.com/prebid/Prebid.js/issues/6673 a duplicate @smenzer ?

[...]

PR #6632, opened 2021-04-20
SharedId & PubcommonId: CCPA and COPPA support

SKOCHERI:
## Type of change

- [ ] Bugfix
- [ ] Feature
- [ ] New bidder adapter   
- [ ] Build related changes
- [ ] CI related changes
- [ ] Other

## Description of change

```
{
  bidder: '<bidder name>',
  params: {
    // ...
  }
}
```


SKOCHERI:
This PR is for https://github.com/prebid/Prebid.js/issues/6433

smenzer (changes requested):
This implementation means that we're now sending consent data twice to id modules, which I don't think is the right approach.  usp and coppa are both already available to id modules by requesting them (rather than be passed in as arguments).  if they must be passed in as arguments, then we should replace the existing consentData with the proposed new object that contains ALL consent data.

SKOCHERI:
> This implementation means that we're now sending consent data twice to id modules, which I don't think is the right approach. usp and coppa are both already available to id modules by requesting them (rather than be passed in as arguments). if they must be passed in as arguments, then we should replace the existing consentData with the proposed new object that contains ALL consent data.

@smenzer Agree that we will have duplicate consent data as argument. Updating the module to  get consent within the module. For coppa, which is in the config and the user id submodules can have access to only the submodule config (as a argument) hence updating the adapterManger.js to get the coppa from the config

smenzer:
> @smenzer Agree that we will have duplicate consent data as argument. Updating the module to get consent within the module. For coppa, which is in the config and the user id submodules can have access to only the submodule config (as a argument) hence updating the adapterManger.js to get the coppa from the config

This looks much cleaner, thank you!

modules/pubCommonIdSystem.js
```diff
@@ -223,6 +233,11 @@ export const pubCommonIdSubmodule = {
    * @returns {IdResponse}
    */
   getId: function (config = {}, consentData, storedId) {
+    const coppa = coppaDataHandler.getCoppa();
+    if (coppa) {
+      utils.logInfo('PubCommonId: coppa not provided, exiting PubCommonId');
```
smenzer, line 238: should this say "coppa not supported" or "IDs not provided for coppa requests"? to me, this message sounds like "coppa config wasn't provided, so we're existing"

modules/pubCommonIdSystem.js
```diff
@@ -263,6 +278,11 @@ export const pubCommonIdSubmodule = {
    * @returns {IdResponse|undefined}
    */
   extendId: function(config = {}, consentData, storedId) {
+    const coppa = coppaDataHandler.getCoppa();
+    if (coppa) {
+      utils.logInfo('PubCommonId: coppa not provided, exiting PubCommonId');
```
smenzer, line 283: same comment as above

modules/sharedIdSystem.js
```diff
@@ -318,6 +328,11 @@ export const sharedIdSubmodule = {
    * @returns {sharedId}
    */
   getId(config, consentData) {
+    const coppa = coppaDataHandler.getCoppa();
+    if (coppa) {
+      utils.logInfo('SharedId: coppa not provided, exiting SharedId');
```
smenzer, line 333: ditto

modules/sharedIdSystem.js
```diff
@@ -333,6 +348,11 @@ export const sharedIdSubmodule = {
    * @returns {{callback: *}}
    */
   extendId(config, consentData, storedId) {
+    const coppa = coppaDataHandler.getCoppa();
+    if (coppa) {
+      utils.logInfo('SharedId: coppa not provided, exiting SharedId');
```
smenzer, line 353: ditto

modules/userId/index.js
```diff
@@ -126,9 +126,9 @@
  */
 
 /**
-  * @typedef {Object} RefreshUserIdsOptions
```
smenzer, line 129: this whole file just seems like whitespace changes now...can you just revert them all since there's no functional change?

smenzer (changes requested):
can you please add tests for the new `getCoppa()` handler?

smenzer (approved):
Thanks for the updates @SKOCHERI ! LGTM

jsnellbaker:
As a temporary measure (see https://github.com/prebid/Prebid.js/pull/6632#issuecomment-828999753), these changes LGTM

Issue #5232, opened 2020-05-13
TCFv2 loading race condition


brondsem:
## Type of issue
bug

## Description
When prebid.js is set to load after the CMP loads, then the addEventListener is registered too late and misses the "tcloaded" event.  (TCFv1 proactively calls getVendorConsents, no events).

If prebid.js set up the cmp addEventListener sooner than pbjs.requestBids that could help, but race conditions still could exist in theory.  Calling getTCData proactively at the same time as addEventListener seems like it could work, but then it runs the risk of getting partial/incomplete data before the CMP has consent data ready.

A stop-gap solution I am using is to set `consentManagement.gdpr.cmpApi = 'static'` and provide the consentData directly since in my case I know prebid runs after CMP.  But that is obviously not a good general solution for anyone else.  It also doesn't update consent data if a user changes their consent, and then further auctions run.

The only reliable solution I can think of is for prebid to continue to rely on addEventListener and provide a mechanism for publishers to trigger the internal `v2CmpResponseCallback` as well, so they can be guaranteed to be called.  Something perhaps like this in publishers sites before prebid & cmp start running:
```
// publisher listens for consent events right away
__tcfapi('addEventListener', 2, function(tcData, success) {
    if(success && (tcData.eventStatus === 'tcloaded' || tcData.eventStatus === 'useractioncomplete')) {
        // and then passes them to prebid when prebid is loaded
        pbjs.que.push(function () {
            // prebid.js would have to expose this:
            pbjs.consentManagement_v2CmpResponseCallback(tcData, success);
        });
    }
});
```

## Steps to reproduce
* have a consent manager with TCFv2 set up
* have the page set up to get consent first and **then** run prebid
* visit page, and accept or reject CMP if its your first visit

## Test page
code: https://jsbin.com/leloquj/edit?html,output
output (without iframes): http://output.jsbin.com/leloquj/

### Expected results
* GDPR details in bid requests

### Actual results
* one `Received a response from CMP` log message, but its eventStatus is undefined, so prebid.js does not do anything with it.  Honestly not sure what this event represents, but the "tcloaded" event definitely is already come and gone.
* wait a little while, then:
```
WARNING: CMP workflow exceeded timeout threshold. Resuming auction without consent data as per consentManagement config.
```
* and no GDPR consent string in bid requests

## Platform details
prebid 3.19.0 with consentManagement module

## Other information
#4664 describes a case where prebid needs to load *after* CMP, as an example of why this order of execution is necessary

[...]

PR #6634, opened 2021-04-21
Consent Management Module: TCFApi in Iframe callId fix

ReinoutStevens:
## Type of change

- [x] Bugfix
- [ ] Feature
- [ ] New bidder adapter   
- [ ] Build related changes
- [ ] CI related changes
- [ ] Other

## Description of change

Prebid installs a tcfapi function in case the CMP is only accessible through the message API. This setup works through an object mapping callIds to callback functions. The callId should be random, but was only generated once. As a result, having calls running in parallel resulted in overriding the callback.

For example: running prebid and amazon in parallel resulted in amazon asking for consent through the tcfapi function installed by prebid, and has a race condition where prebid's callback never fired as it got overridden by amazon's consent request.

## Other information
@jsnellbaker I think you're the owner / original author of the module, so tagging you. Sorry if that's not the case.

### DIFF
--- a/modules/consentManagement.js
+++ b/modules/consentManagement.js
@@ -200,7 +200,7 @@ function lookupIabConsent(cmpSuccess, cmpError, hookConfig) {
     let apiName = (cmpVersion === 2) ? '__tcfapi' : '__cmp';
 
     let callName = `${apiName}Call`;
-    
+
     /* Setup up a __cmp function to do the postMessage and stash the callback.
     This function behaves (from the caller's perspective identicially to the in-frame __cmp call */
     if (cmpVersion === 2) {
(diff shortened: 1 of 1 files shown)


patmmccann:
Is this related to #5232 ?

ReinoutStevens:
No, that issue seems unrelated.

To provide a bit more context for this fix:

This issue is specific for consent across iframes + multiple libraries requesting consent info at the same time.
The installed tcfapi function of prebid does not generate a random id for each request, but instead reuses 1 single random id for all requests. As a result, when multiple calls are happening at the same time, the callback functions get overridden, and the callbacks get messed up, resulting in callbacks never being called or called with wrong data.

harpere (approved):

PR #6642, opened 2021-04-21
adformOpenRTB Bid Adapter: change name to adf to avoid collision

braizhas:
## Type of change
- [x] Other

## Description of change
Added prebid-server adapter https://github.com/prebid/prebid-server/pull/1815 . Changing the name to avoid targeting key collision.

- Team.FLS@adform.com

- Docs: https://github.com/prebid/prebid.github.io/pull/2887

### DIFF
--- a/modules/adfBidAdapter.js
+++ b/modules/adfBidAdapter.js
@@ -10,8 +10,9 @@ import {
 import * as utils from '../src/utils.js';
 import { config } from '../src/config.js';
 
-const BIDDER_CODE = 'adformOpenRTB';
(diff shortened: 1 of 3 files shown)


patmmccann:
This appears to be a breaking change,  do you agree?

If so please make it in the 5.0 branch,  expected to be released end of May.

braizhas:
I don't think that it's a breaking change - exiting setups should work via bidder alias. Am I missing something?

patmmccann:
My mistake,  thank you for clarifying

patmmccann (approved):

PR #6645, opened 2021-04-22
Trigger AUCTION_DEBUG from utils.logWarn with type set to WARNING

pm-harshad-mane:
## Type of change
- [ X ] Feature

## Description of change
Trigger AUCTION_DEBUG from utils.logWarn with type set to WARNING, this will help publishers/analytics adapters collect the warnings together.


pm-harshad-mane:
@jsnellbaker , I need some help here, https://github.com/prebid/Prebid.js/blob/master/test/spec/AnalyticsAdapter_spec.js#L45 this test case fails with the change made. Am not sure why it fails and how to fix it :(

patmmccann:
Closely related to https://github.com/prebid/Prebid.js/pull/5563

FYI @ofirpaBrowsi @GLStephen @harpere

harpere (commented):
LGTM other than the broken test.

jsnellbaker:
> @jsnellbaker , I need some help here, https://github.com/prebid/Prebid.js/blob/master/test/spec/AnalyticsAdapter_spec.js#L45 this test case fails with the change made. Am not sure why it fails and how to fix it :(

Can you try to flip the order of the `args` and `eventType` params in the AnaltyicsAdapter_spec file's expect statements (in the `deep.equal` target value)?  I've seen the deep equal verification gets picky sometimes.

harpere (approved):

pm-harshad-mane:
@jsnellbaker thank you for the feedback.
As we are triggering the AUCTION_DEBUG for warning calls, there were more event records in the array before the expected event in the test case, I have made the required change in the test case.

pm-harshad-mane:
Thank you @harpere for merging the PR.

Issue #6465, opened 2021-03-23
Bid adapters supporting floors but not the floors module


patmmccann:
## Type of issue
It is expected https://github.com/prebid/Prebid.js/issues/5966 will require bidders accepting the floor to use the getFloor function. 

Here is a quick identification of the adapters missing support: 

`grep floor *BidAdapter.js | grep -v Math.floor | awk '{print $1}' | awk -F ':' '{print $1}' | awk -F '-' '{print $1}' | sort | uniq > ~/floors.txt`

`grep getFloor *BidAdapter.js | awk '{print $1}' | awk -F ':' '{print $1}' | awk -F '-' '{print $1}' | sort | uniq > ~/floorsmod.txt`

`grep ~/floors.txt -f ~/floorsmod.txt --invert-match`

adspendBidAdapter.js @loorke 
adtrueBidAdapter.js @haibau 
advangelistsBidAdapter.js @trchandraprakash 
audiencerunBidAdapter.js @audiencerun 
c1xBidAdapter.js @CathyHuangtw 
cleanmedianetBidAdapter.js @sa1omon 
cosmosBidAdapter.js @dev-cosmos 
dailyhuntBidAdapter.js @montu1996 
ebdrBidAdapter.js @jlzhangdev 
emx_digitalBidAdapter.js @EMXDigital 
fidelityBidAdapter.js @onaydenov 
gamoshiBidAdapter.js @sa1omon 
getintentBidAdapter.js @andreykud 
imonomyBidAdapter.js @imonomy 
improvedigitalBidAdapter.js @jbartek25 
ironsourceBidAdapter.js @lasloche 
komoonaBidAdapter.js @tzafrirb 
kubientBidAdapter.js @onaydenov
lunamediaBidAdapter.js @trchandraprakash 
luponmediaBidAdapter.js @adxpremium 
madvertiseBidAdapter.js @NLopezMad 
mediaforceBidAdapter.js @Niksok 
mgidBidAdapter.js @Gaudeamus 
mobfoxBidAdapter.js @hadarpeer 
mytargetBidAdapter.js @vfedoseev 
nafdigitalBidAdapter.js @vyatsun 
newborntownWebBidAdapter.js @z-sunshine 
nextrollBidAdapter.js @pbrudnick 
orbidderBidAdapter.js @hendrikiseke1979 
otmBidAdapter.js @fedotxxl 
ozoneBidAdapter.js @afsheenb 
platformioBidAdapter.js @varashellov 
[...]

PR #6653, opened 2021-04-26
appnexus bid adapter - add price floor module support

jsnellbaker:
## Type of change

- [x] Feature

## Description of change

Related to #6465

This PR updates the appnexus bid adapter to use the priceFloors module.  

Specifically the logic between the module and the existing `reserve` param will wok as follows:
- if priceFloors module is defined and `reserve` is not, read the values and pass the matching floor price to ad server
- if priceFloors module is defined and `reserve` is defined, prefer the priceFloors module's value(s)
- if priceFloors module is not defined and `reserve` is defined, pass `reserve` value (as normal).

### DIFF
--- a/modules/appnexusBidAdapter.js
+++ b/modules/appnexusBidAdapter.js
@@ -703,8 +703,9 @@ function bidToTag(bid) {
   tag.use_pmt_rule = bid.params.usePaymentRule || false
   tag.prebid = true;
   tag.disable_psa = true;
-  if (bid.params.reserve) {
-    tag.reserve = bid.params.reserve;
+  let bidFloor = getBidFloor(bid);
+  if (bidFloor) {
+    tag.reserve = bidFloor;
   }
   if (bid.params.position) {
     tag.position = { 'above': 1, 'below': 2 }[bid.params.position] || 0;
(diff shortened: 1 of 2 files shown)


modules/appnexusBidAdapter.js
```diff
@@ -1036,4 +1037,21 @@ function addUserId(eids, id, source, rti) {
...
+function getBidFloor(bid) {
+  if (!utils.isFn(bid.getFloor)) {
+    return (bid.params.reserve) ? bid.params.reserve : null;
+  }
+
+  let floor = bid.getFloor({
+    currency: 'USD',
+    mediaType: '*',
+    size: '*',
+    bidRequest: bid
```
robertrmartinez, line 1049: you can leave bidRequest out, it gets the context from `this`

robertrmartinez (changes requested):
one small change, but LGTM. Thanks!

modules/appnexusBidAdapter.js
jsnellbaker: updated.

jsnellbaker:
@robertrmartinez When you get a chance, can you take a look again?  Thanks.

robertrmartinez (approved):
Looks good!

patmmccann:
@jsnellbaker can you add docs for this and #6688 please?

jsnellbaker:
Hi @patmmccann 

I see there's a field for the floors module support, but is there one for the gptPreAuction?

patmmccann:
No, this has become quite popular in the last month,  I wonder if we should add one. Nevermind on the 6688 request,  will bring up in committee

jsnellbaker:
Docs PR was created https://github.com/prebid/prebid.github.io/pull/2927 for floors.

Issue #6650, opened 2021-04-23
Adapter does not seem capable of supporting advertiserDomains


gglas:
In conjunction with https://github.com/prebid/Prebid.js/issues/6466 , there will be an expectation going forward that adapters at least have a method to pass meta.advertiserDomains in their bid responses.  This is because in the future we expect publishers and vendors to use the adomain(s) in service of any number of business cases.

This is not a hard requirement to pass adomain, however, adapters at least need to have a stub to pass adomain in the appropriate field (the meta.advertiserDomains field).  The following adapters have no mention of adomain or advertiserDomains, and while some adapters may be modified with the appropriate stub by other prebid members, any adapters on this list should ensure that they support the relevant field in their adapter or risk not being ported to 5.0.

a4gBidAdapter | devops@ad4game.com | @Junus
aardvarkBidAdapter | chris@rtk.io | @aleksatr
ablidaBidAdapter | d.kuster@ablida.de | @bokos
adbutlerBidAdapter | dan@sparklit.com | @dharton
adfinityBidAdapter | adfinity_prebid@i.ua | @adfinity-prebid
adformBidAdapter | Scope.FL.Scripts@adform.com | @Pupis
adgenerationBidAdapter | ssp-ope@supership.jp | @banakemi
adglareBidAdapter | prebid@adglare.com | @adglare
adkernelAdnBidAdapter | prebid-dev@adkernel.com | @ckbo3hrk
adliveBidAdapter | traffic@adlive.io | @mifanich
admanBidAdapter | traffic@adlive.io | @SmartyAdman
admaticBidAdapter | prebid@admatic.com.tr | @hakanarik
admediaBidAdapter | developers@admedia.com | @Elardzhi
adnowBidAdapter | support@adnow.com | @vingood
adnuntiusBidAdapter | info@adnuntius.com | @mikael-lundin
adoceanBidAdapter | prebid@gemius.com | @mmuras
adpartnerBidAdapter | info@adpartner.pro | @preved-medved
adprimeBidAdapter | tech@adpone.com | @Adprime
adspendBidAdapter | gaffoonster@gmail.com | @loorke
adtargetBidAdapter | kamil@adtarget.com.tr | @GeneGenie
adtelligentBidAdapter | support@adtelligent.com | @GeneGenie
adtrueBidAdapter | ssp@adtrue.com | @haibau 
advenueBidAdapter | dev.advenue@gmail.com | @devadvenue 
[...]

Issue #6466, opened 2021-03-23
Adapter supports adomain but not meta.advertiserDomains


patmmccann:
## Type of issue
Some bidders explicitly support adomain in their adapter code but don't set meta.advertiserDomains . Others support it in their Prebid.js/test/spec/modules but not in their adapter js. This issue lists both types. 

Specifically, this adapter has support for adomain but not meta.advertiserDomains: 

https://github.com/patmmccann/Prebid.js/blob/master/modules/ozoneBidAdapter.js @afsheenb @askrupert-dm

These support it in their test but not explicitly in their adapter. 
`~/Prebid.js/modules$ grep dvertiserDomains *BidAdapter.js | awk -F ':' '{print $1}'  | awk -F '.' '{print $1}' > ~/adomains.txt`

`~/Prebid.js/test/spec/modules$ grep -i adomain *BidAdapter_spec.js | awk -F ':' '{print $1}' > ~/adomain_test.txt`

`grep -v ~/adomain_test.txt -f ~/adomains.txt | sort | uniq`

adponeBidAdapter_spec.js @seergiioo6 
advangelistsBidAdapter_spec.js @trchandraprakash 
bidphysicsBidAdapter_spec.js @bidphysics 
bizzclickBidAdapter_spec.js @BizzClick 
cedatoBidAdapter_spec.js @alexkh13 
cleanmedianetBidAdapter_spec.js @sa1omon 
cosmosBidAdapter_spec.js @dev-cosmos 
ebdrBidAdapter_spec.js @jlzhangdev 
gamoshiBidAdapter_spec.js @sa1omon 
lemmaBidAdapter_spec.js @lemmadev 
lkqdBidAdapter_spec.js @david-carver 
lunamediaBidAdapter_spec.js @trchandraprakash 
marsmediaBidAdapter_spec.js @vladi-mmg 
newborntownWebBidAdapter_spec.js @z-sunshine 
pxyzBidAdapter_spec.js @monofonik 
resultsmediaBidAdapter_spec.js @vladi-mmg 
theAdxBidAdapter_spec.js @mustafakemal16 
timBidAdapter_spec.js @borisjaskerovich 
unicornBidAdapter_spec.js @ctylim 
yieldliftBidAdapter_spec.js @yieldlift

patmmccann:
Please see https://github.com/prebid/Prebid.js/pull/6495 for an example of how to make this fix

patmmccann:
[...]

PR #6656, opened 2021-04-27
AOL, RTBHouse, RhythmOne, Teads, OpenX and DeepIntent Bid Adapters: add support for meta.advertiserDomains

patmmccann:
## Type of change

- [ ] Bugfix
- [ ] Feature
- [ ] New bidder adapter   
- [ ] Build related changes
- [ ] CI related changes
- [X] Other

## Description of change
Partial solution to https://github.com/prebid/Prebid.js/issues/6466 aand #6650


test/spec/modules/openxBidAdapter_spec.js
```diff
@@ -1574,7 +1570,11 @@ describe('OpenxAdapter', function () {
         expect(bid.meta.brandId).to.equal(DEFAULT_TEST_ARJ_AD_UNIT.brand_id);
       });
 
-      it('should return a brand ID', function () {
+      it('should return an adomain', function () {
+        expect(bid.meta.advertiserDomains).to.deep.equal([]);
```
ChrisHuie, line 1574: @patmmccann will this break when that field starts getting used since it will then be equal to `adUnit.adomain` or am I looking at this wrong?

patmmccann: The goal here is that the mock response in their tests will have this placeholder.  I they change their response,  they'll need to follow up here to change how it is interpreted.

ChrisHuie (approved):

PR #6659, opened 2021-04-27
Tappx Bid Adapter: optional ext on request

prebidtappx:
## Type of change

- [x] Feature

## Description of change

This change is going to allow send extra data to tappx via prebid.js using "ext" property in the bidder.


patmmccann:
Please pull prebid master into your fork and branch so as to not affect outbrain

prebidtappx:
> Please pull prebid master into your fork and branch so as to not affect outbrain

Hi! We had the same problem in other branch, the pull doesn't work (we're not sure why), but if we change manually seems OK.
Thanks for all!

modules/tappxBidAdapter.js
```diff
@@ -336,7 +338,9 @@ function buildOneRequest(validBidRequests, bidderRequest) {
   payload.params = params;
   payload.regs = regs;
   // < Payload
-
+  utils.logMessage('---------------------');
+  utils.logMessage(JSON.stringify(payload));
```
patmmccann, line 342: why

patmmccann: nm, see this is behind debug mode now, sorry

harpere (approved):

Issue #6660, opened 2021-04-27
Updating Dev Dependencies


ChrisHuie:
Several of our testing, linting, and gulp dev dependencies appear a version or so behind per the David tag (https://david-dm.org/prebid/Prebid.js?type=dev) on the js readme. I'm planning on going through these and incrementing them and debugging anything that breaks.  If there is any dev dependency you don't think should be updated or there is some reason to not update some of the gulp and webpack modules other than current Prebid compatibility let me know.

ChrisHuie:
#6663

ChrisHuie:
#7041

PR #6663, opened 2021-04-27
Dev Dependencies: update dependencies to latest workable versions

ChrisHuie:
Updated dev dependencies from https://david-dm.org/prebid/Prebid.js?type=dev 

This represents the most recent dependency stack I could get working together without updating a significant amount of other files. Going to make those updates separately to include files that are affected by deprecations and what not.


ChrisHuie:
#6660 created issue for currently updating dev dependencies

ChrisHuie:
*Working to solve a common failing test along with this update

patmmccann:
what are your thoughts on https://github.com/prebid/Prebid.js/pull/6493

ChrisHuie:
> what are your thoughts on #6493
> 
> seems we are allowing remote code execution attacks by not pushing this

I saw this when it came across but wasn't super sure on the dependencies update policy here since that is a package dependency not a dev dependency. Just assumed the person that closed it was more familiar but I can research the specific patch and reason a bit more

ChrisHuie:
moved to #6328 to add stability and incorporate with npm audit fix

PR #6666, opened 2021-04-27
DMD ID System: add new User ID module

boppudikarthikc:
## Type of change
- [ ] Feature

## Description of change
New DMD user module is added which supports reading DMD first party authentication ID.

- contact : [prebid@dmdconnects.com]()

A link to a PR on the docs repo at [https://github.com/prebid/prebid.github.io/pull/2909]()


modules/dmdIdSystem.js
```diff
@@ -0,0 +1,58 @@
...
+   * @param {Object} cacheIdObj - existing id, if any consentData]
+   * @returns {IdResponse|undefined}
+   */
+  getId(config, consentData, cacheIdObj) {
+    try {
+      const configParams = (config && config.params) || {};
+      if (
+        !configParams ||
+        !configParams.api_key ||
+        typeof configParams.api_key !== 'string'
```
smenzer, line 45: what does the `api_key` do, considering you don't check it's validity or anything? meaning, all i have to do is add any string and this will allow it to pass through

matthewfitz: @smenzer In a future release, DMD plans to introduce reading the authentication ID against a REST API. The REST API will validate the API Key on the backend. Our customers typically are unwilling to change integrations after the initial installation. We would like to standardize our configuration to require an API Key in this release so our customers do not need to revise their integration after we introduce the REST API in a subsequent release.

smenzer: thanks for the clarification. why not build the rest api now even if you're not authenticating so that publishers won't have to upgrade prebid to get the update (which, as you mentioned, they don't like having to make changes)?

matthewfitz: @smenzer DMD is working on the next release now, but we wanted to submit the 1st party authentication ID feature first as this feature addition is low in complexity and thought it would be better for us to learn how to contribute features to Prebid with a low complexity change. If you would prefer us to remove the expectation of the api_key in the configuration for this release, we can remove it.

smenzer: I don't have a problem with the key being in there, I just wanted to point out that by not building to your rest api now, you'll still be requiring publishers to make a change (upgrade prebid) when it's ready. if you're comfortable with that, i have no objections to leaving the config as you have it.

modules/dmdIdSystem.md
```diff
@@ -0,0 +1,15 @@
+pbjs.setConfig({
+    userSync: {
+        userIds: [{
+            name: 'dmdId',
+            storage: {
+                name: 'dmd-dgid',
```
smenzer, line 6: since you're relying on a value to already exist in this cookie, do you want to be more explicit in the docs (both here and in https://github.com/prebid/prebid.github.io/pull/2909/files) that the `storage.name` must be `'dmd-dgid'`?

smenzer (changes requested):
Looks good, but can you please add tests to `userId_spec.js`, you can mirror the changes the other ids modules make, mostly just adding yours to the list in several tests.

modules/dmdIdSystem.js
matthewfitz: @smenzer Thank you. We appreciate the understanding. You are correct and we will note to remind publishers that a prebid rebuild is required when we release the REST API addition.

modules/dmdIdSystem.md
boppudikarthikc: @smenzer your feedback is implemented. `storage.name` is included in `dmdIdSystem.md` and [https://github.com/prebid/prebid.github.io/pull/2909/files](url). Added more tests to `userId_spec.js`

smenzer (approved):
LGTM

PR #6657, opened 2021-04-27
AOL Bid Adapter: add support for the floors module

patmmccann:
## Type of change

- [ ] Bugfix
- [ ] Feature
- [ ] New bidder adapter   
- [ ] Code style update (formatting, local variables)
- [ ] Refactoring (no functional changes, no api changes)
- [ ] Build related changes
- [ ] CI related changes
- [ ] Does this change affect user-facing APIs or examples documented on http://prebid.org?
- [X] Other

## Other information

Partial solution to #6465

PR #6672, opened 2021-04-28
oneVideo Bid Adapter: Price Floors Module Support (SAPR-16735)

adam-browning:
## Type of change

- [ X] Feature

## Description of change

oneVidoe bid adapter now supports the Price Floors module by utilizing the .getFloor() method to retrieve as set  bidfloor either on the adUnit or on the config levels.

The `.getFloor(args)` function requires passing an argument object that defines: currency, format and sizes.
The default currency of the argument object is 'USD' but we support an override using `bid.params.cur` with any 3 letter currency code all capitals to fetch floors in different currnecies.
Note: the publisher is required to match and maintain the price floor currency and to match the adapter request object using the same currency type in `bid.params.cur`

The adapter will retrieve the floor price according to the following fallback logic:
1. Use adUnit || Config module floor if exists
2. If adUnit || Config module floor are not defined, the adapter will try to use `bid.params.bidfloor` if defined.
3. If `bid.params.bidfloor` is not passed, `bid.imp[0].bidfloor` will not be added to the outbound bid-request to Video SSP.

Use the following to example to review the price floor module result in the outbound bid-request.

```
pbjs.setConfig({
      useBidCache: true,
      debug: true,
      floors: { enabled: true }
});

const adUnits = [{
    code: 'video1',
    mediaTypes: {
        video: {
            context: 'outstream',
            playerSize: [300, 250]
        }
    },
    floors: {
     currency: 'EUR',
     schema: {
         delimiter: '|',
         fields: [ 'mediaType', 'size' ]
     },
     values: {
         'video|300x250': 1.11
     }
 },
    bids: [
        {
            bidder: 'oneVideo',
            params: {
                cur: 'EUR',
                video: {
                    playerWidth: 300,
                    playerHeight: 250,
                    protocols: [2, 5],
                    api: [2]
                },
                site: {
                    ref: 'http://www.verizonmedia.com',
                    page: 'http://www.verizonmedia.com'
                },
                pubId: 'HBExchange'
            }
        }
    ]
}];
```

## Other information
Hey @DeepthiNeeladri & @kprasadpvv,
I've added support for the Price Floor module as requested in:
https://github.com/prebid/Prebid.js/issues/5966
and https://docs.prebid.org/dev-docs/modules/floors.html#bid-adapter-interface
I would appreciate it if you could review this PR please.
Thanks in advance!
Adam


modules/oneVideoBidAdapter.js
```diff
@@ -244,6 +248,15 @@ function getRequestData(bid, consentData, bidRequest) {
       bidData.imp[0].banner.ext.minduration = bid.params.video.minduration
     }
   }
+
+  if (typeof bid.getFloor === 'function') {
```
kprasadpvv, line 252: Not important but we have function type check availability in utils.
in case if you wish to use it. 
`if (utils.isFn(bid.getFloor))`

adam-browning:
FYI - circleci failing due to RealVu Analytics Adapter AFAIK.
Will merge upstream master again in a few hours

modules/oneVideoBidAdapter.js
adam-browning: Nice! will switch to that 👍

modules/oneVideoBidAdapter.js
```diff
@@ -294,6 +307,7 @@ function getRequestData(bid, consentData, bidRequest) {
     }
   }
   if (bid.params.video.e2etest) {
+    utils.logMessage('+++ oneVideoBidAdapter: E2E test mode enabled. \n The following parameters are being overridden by e2etest mode:\n* bidfloor:null\n* width:300\n* height:250\n* mimes: video/mp4, application/javascript\n* api:2\n* site.page/ref: verizonmedia.com\n* tmax:1000');
```
kprasadpvv, line 310: Just out of curious, where are these log available.

patmmccann:
Thanks! Removed from list at https://github.com/prebid/Prebid.js/issues/6465

modules/oneVideoBidAdapter.js
patmmccann: this goes to console in debug mode

adam-browning: It logs when pbjs debug mode is active
Either when the url contains `?pbjs_debug=true`
or
```
pbjs.setConfig({
      debug: true
});
```

modules/oneVideoBidAdapter.js
```diff
@@ -165,14 +165,17 @@ function getRequestData(bid, consentData, bidRequest) {
   let loc = bidRequest.refererInfo.referer;
   let page = (bid.params.site && bid.params.site.page) ? (bid.params.site.page) : (loc.href);
   let ref = (bid.params.site && bid.params.site.referrer) ? bid.params.site.referrer : bidRequest.refererInfo.referer;
+  let getFloorRequestObject = {
+    currency: bid.params.cur || 'USD',
+    mediaType: 'video',
+    size: '\*'
```
patmmccann, line 171: is the '\' character here a typo? haven't seen that in other implementations

adam-browning: it was taken from the adapter example in https://docs.prebid.org/dev-docs/modules/floors.html#bid-adapter-interface
originally i thought there was an escaped character issue with the string once it's passed to .getFloor()
I can easily remove and test again.

kprasadpvv, line 171: In the docs, in one example the passed just `'*'` and in another example they used `'\*'`, do you think there will be any difference in calculation if we used `escape sequence` for `*`.

kprasadpvv: Also just a reminder, is there anything that needs to be updated in `adapter md` file.

Apart from this everything looks fine @adam-browning.

Thank you  😊

modules/oneVideoBidAdapter.js
kprasadpvv: got it Thank you

modules/oneVideoBidAdapter.js
adam-browning: I didn't see any differences in the results after testing both.
I am removing the escaped character backslash.

adam-browning: removed backslash

kprasadpvv (approved):

modules/oneVideoBidAdapter.js
patmmccann: @bszekely1 is there a typo in the doc? interestingly doesn't seem to effect things

bszekely1: That is a typo. Let me fix it. I would recommend to remove it even if it works.

bszekely1: Updated. Thanks for catching it.

adam-browning:
FYI - circleci failing due to other adapter issues.

patmmccann:
I restarted tests.

@adam-browning I made a failed attempt at adding support for your other adapter; it turned out to be more complicated than I expect. https://github.com/prebid/Prebid.js/pull/6657 ; curious what your plans are for it as well?

PR #6674, opened 2021-04-29
Admixer Bid Adapter: add meta key to response

AdmixerTech:
## Type of change
- [x] Feature

## Description of change
add meta key to interpreted bid response

### DIFF
--- a/modules/admixerBidAdapter.js
+++ b/modules/admixerBidAdapter.js
@@ -76,7 +76,7 @@ export const spec = {
           vastUrl: bidResponse.vastUrl,
           dealId: bidResponse.dealId,
           /**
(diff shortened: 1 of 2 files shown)


patmmccann (changes requested):
Please also include advertiserDomains

patmmccann:
Thank you very much. 

One quick feedback, can you explicitly set meta.advertiserDomains to empty array or add a comment mentioning which meta fields you set in the js file? 

Sorry to be pedantic, but we're exploring removing `grep -v advertiserDomains *BidAdapter.js` and want to make sure you're not accidentally back on the list

patmmccann (approved):

PR #6676, opened 2021-04-29
Prebid Core: add X-Prebid header to ajax requests 

### LINKED ISSUE #6627: Prebid version in an HTTP header
## Type of issue
Proposed Enhancement

## Description
There's interest from some bidders to be able to cleanly distinguish the source of traffic. Currently it's up to each bidder to read the Prebid.js version and pass it to their endpoint. There aren't any conventions about where bidders should place this version, so they've done it every which way. There are 83 bid adapters that refer to prebid.version, with not much overlap in what they call the attribute: prebidVersion, pbjs, wrapperVersion, wrapper_version, version, pbav, hb_version, pubpbv, prebidJsVersion, pub_pbver, lib_v, tk_flint, etc.

The proposal is to update ajax.js to add a new HTTP header `X-Prebid` which carries the PBJS and PBS versions:

```
X-Prebid: pbjs/4.35
```

Prebid Server would add it's own version on to this header:
```
X-Prebid: pbjs/4.35,pbs-java/1.61
```

It's recognized that it may take a while before the majority of publishers have upgraded to the version of PBJS that contains this header, but you have to start somewhere.

snapwich:
## Type of change
- [x] Feature

## Description of change
Adds `X-Prebid` header to ajax requests.

## Other information
From issue #6627

### DIFF
--- a/src/ajax.js
+++ b/src/ajax.js
@@ -75,11 +75,12 @@ export function ajaxBuilder(timeout = 3000, {request, done} = {}) {
       if (options.withCredentials) {
         x.withCredentials = true;
       }
+      x.setRequestHeader('x-prebid', 'pbjs/$prebid.version$');
       utils._each(options.customHeaders, (value, header) => {
         x.setRequestHeader(header, value);
       });
       if (options.preflight) {
-        x.setRequestHeader('X-Requested-With', 'XMLHttpRequest');
+        x.setRequestHeader('x-requested-with', 'XMLHttpRequest');
       }
       x.setRequestHeader('Content-Type', options.contentType || 'text/plain');
 


patmmccann (approved):

src/ajax.js
```diff
@@ -75,6 +75,7 @@ export function ajaxBuilder(timeout = 3000, {request, done} = {}) {
       if (options.withCredentials) {
         x.withCredentials = true;
       }
+      x.setRequestHeader('X-Prebid', 'pbjs/$prebid.version$');
```
bretg, line 78: Sorry for the change here @snapwich , but the PBS committee today noted that the prefered HTTP header syntax is lower case. May I ask you to change this to "x-prebid" please?

bretg (changes requested):

harpere:
Also @snapwich - There's a strange Circle CI error.  Any idea what this is from?  That looks like the hash of your commit.
`fatal: reference is not a tree: 827c941e08218c55f7053782e6272f76f11c26c7`

src/ajax.js
snapwich: i'll change the `X-Requested-With` below it to be lowercase as well just to be consistent.

snapwich:
@harpere no clue where that error is coming from. hopefully pushing a new commit fixes it.

bretg (approved):

Issue #6587, opened 2021-04-14
Native bid missing some required properties error on NON REQUIRED fields


guiann:
## Type of issue
Limitation which looks buggy according to prebid documentation 

## Description
when reponding with an "icon" or "image" field as string in a Native bid response the bid is always invalid.
Even if the icon/image is NOT required.

## Steps to reproduce
From any native test page.  set icon: required false  in native request.
in your adapter force an icon value like:

`    bid.native.icon = 'https://fo-static.omnitagjs.com/fo-static/native/images/info-ayl.png';`

According to the doc, this shouldn't be checked as it's not even required.
And if it's required and the value is checked,  a string  or an object should be valid:

```
All assets can be returned as strings, or images can be returned as objects with attributes url, height, and width.
```

### Actual results
The code inside nativeBidIsValid  from src/native.js disallow  image or icon without width or height field. So strings are not allowed.
No matter if the field was required or not.

### Expected results
values like:
bid.icon = 'https://fo-static.omnitagjs.com/fo-static/native/images/info-ayl.png
bid.image = 'https://fo-static.omnitagjs.com/fo-static/native/images/info-ayl.png

should be valid since partners dont always provide icons width and height.
Or at least we can remove the check if values are not required ?

## Other information
Here is the code from "nativeBidIsValid".
Note that this is done before "required" status concern.

https://github.com/prebid/Prebid.js/blob/41e050a28003a4c0a23967694c3a327c3dba62be/src/native.js#L80-L90

gglas:
Hey @guiann I think we'd be amenable to entertaining either of these : 

- should be valid since partners dont always provide icons width and height.
- Or at least we can remove the check if values are not required ?

Would you be interested in proposing a direct code change and we can facilitate the PR ?

guiann:
Sure., will do.

guiann:
done: https://github.com/prebid/Prebid.js/pull/6678

PR #6678, opened 2021-04-30
Prebid Core: Remove size check on native icons and image assets

guiann:
Fix Native parameter check by removing property check.

The removed check was disallowing icons and image without a size. which is allowed by the documentation.

See issue https://github.com/prebid/Prebid.js/issues/6587


ChrisHuie (approved):

ncolletti (approved):
LGTM

PR #6683, opened 2021-04-30
Prebid Core: Display bidder in utils.logs and disable bidder config on auction end

snapwich:
## Type of change
- [x] Bugfix
- [x] Feature

## Description of change
As discussed in Prebid.js PMC meeting, this adds bidder logging to `utils.logMessage`, `utils.logWarn`, `utils.logError`, `utils.logInfo` when a bidder has been set using `config.runWithBidder` (which includes all the functions that run as part of a bidder spec).  This adds bidder labels to various relevant areas as shown below:

Capture

Also fixed a small potential surface area for bugs by clearing the bidder once bidders are done bidding. Currently the last bidder would still be set as the active bidder while prebid core was running code to finish the auction. This is due to the last bidder triggering the auction end code in the same call stack as its ajax response handler (which I think we want to keep for performance reasons).


harpere (approved):
LGTM

mmoschovas (approved):
LGTM

Issue #6677, opened 2021-04-30
Webpack v5 complaining about prebidServerBidAdapter


khatibda:
## Type of issue
Warning (not yet a bug)

## Description
Webpack v5 is complaining about prebidServerBidAdapter.  Specifically, it doesn't like how properties are imported as their own names from a default exported json module (in constants.json).

example warnings:
![image](https://user-images.githubusercontent.com/1466555/116644220-a4e19d00-a927-11eb-90e8-c3b9e9b09d0a.png)

webpack guidance:
https://webpack.js.org/migrate/5/#using-named-exports-from-json-modules
![image](https://user-images.githubusercontent.com/1466555/116644521-71ebd900-a928-11eb-9372-304961a546b5.png)

code causing the isssue:
https://github.com/prebid/Prebid.js/blob/master/modules/prebidServerBidAdapter/index.js
https://github.com/prebid/Prebid.js/blob/master/src/constants.json

it doesn't like this:
`import { STATUS, S2S, EVENTS } from '../../src/constants.json';`

a simple fix here could be to just import CONSTANTS and reference CONSTANTS.STATUS instead.

I also don't know if there are other files/plugins that might suffer the same warning, but I just don't have them installed in my build.  In which case, if a more global fix were needed, could update all the imports across the code base, or could update constants.json to have a different structure.

thoughts?

## Platform details
I'm using prebid 4.33.0

ChrisHuie:
#6663 updated some of the supporting dev dependencies here but working to make all the necessary changes to upgrade all the dev dependencies so webpack should be updated to latest shortly

ChrisHuie:
Webpack v5 not yet supported but working towards updating all dev dependencies and implementing needed fixes

pm-harshad-mane:
Hello @ChrisHuie ,
Looks like upgrading to Webpack v5 is a big task for us, can you please share your findings like where do we stand, what are blockers, the steps that are required in a shared document (a Google Doc) so that other devs can contribute.

PR #6685, opened 2021-05-02
Dev Devendencies: Update webpack to v4.0.0

ChrisHuie:
Updating webpack to v4.0.0, add webpack-cli,  update build script, and fix whatever breaks. Appears the key issue is the SplitChunksPlugin we use is no longer supported in this verison. Docs to use optimization.splitChunks instead -> https://webpack.js.org/plugins/split-chunks-plugin/


ChrisHuie:
#6660 
#6677

Issue #6512, opened 2021-04-01
Supports video but does not read video parameters from the ad unit


patmmccann:
## Type of issue
Adapter rules

## Description
With Prebid 4.0, https://github.com/prebid/Prebid.js/issues/5237 was adopted, where publishers can expect to be able to set openrtb video params on the ad unit instead of having to repeat the parameter set up with differing requirements for every adapter they install. At that time and again with the #5966 proposal, it was announced that behavior would be enforced. 

This is a rough computer-generated, person edited list of adapters supporting video but not reading many parameters from mediaTypes.video. There are other adapters not on this list only reading some parameters from that location which should also be fixed.

A great example of how to read from the correct place includes https://github.com/prebid/Prebid.js/pull/6177 or https://github.com/stevealliance/Prebid.js/blob/a5ec2587abe94eae75300b52b7ccceaf70e78a41/modules/districtmDMXBidAdapter.js#L157

The following bid adapters read bidder-specific video params instead of using mediatypes, eg for parameter 'mimes'. Note, a more thorough review will be done before any enforcement action. If you don't think you should be on this list please comment:

https://github.com/prebid/Prebid.js/blob/master/modules/advangelistsBidAdapter.md @trchandraprakash 
https://github.com/prebid/Prebid.js/blob/master/modules/dailyhuntBidAdapter.md @montu1996 
https://github.com/prebid/Prebid.js/blob/master/modules/lemmaBidAdapter.md @lemmadev 
https://github.com/prebid/Prebid.js/blob/master/modules/lunamediaBidAdapter.md @trchandraprakash 
https://github.com/prebid/Prebid.js/blob/master/modules/waardexBidAdapter.md @yegorWaardex 

These bid adapters have acknowledged the issue and have pending pr's:
https://github.com/prebid/Prebid.js/blob/master/modules/saambaaBidAdapter.md @trchandraprakash 

Also related:
https://github.com/prebid/prebid.github.io/pull/2158
https://github.com/prebid/Prebid.js/pull/5536

jsnellbaker:
Hi @patmmccann 

Just want to clarify the intent of this change; it is meant to completely discourage bidders from setting **any** video params in the bidder params and they have to read everything from the `mediaTypes.video` field in an ORTB lense?  

Or do bidders have to add more support for reading those values from the `mediatTypes.video` field, but they can continue to read values from the bidder params (preferring one over the other as mentioned in the related docs PR prebid/prebid.github.io#2158).

[...]

PR #6690, opened 2021-05-03
appnexus bid adapter 5.0 - video params updates

jsnellbaker:
## Type of change

- [x] Feature

## Description of change
Related to https://github.com/prebid/Prebid.js/issues/6512

This change updates the appnexus bidder to read more from the `mediaTypes.video` params.  

It supports specific ORTB 2.5 video params that are closely aligned to video properties we currently support.  As discussed in the linked ticket, due to a number of unique values that are available/used by our ad-server but that are not supported by the ORTB spec - the `bids.params.video` are still read and would be preferred (if present) over the mediaType values.

### DIFF
--- a/modules/appnexusBidAdapter.md
+++ b/modules/appnexusBidAdapter.md
@@ -90,12 +90,14 @@ var adUnits = [
        video: {
          playerSize: [[300, 250]],
          context: 'outstream',
-         // certain ORTB 2.5 video values can be read from the mediatypes object; below are examples of supported params
-         // to note - appnexus supports additional values for our system that are not part of the ORTB spec.  If you want
+         // Certain ORTB 2.5 video values can be read from the mediatypes object; below are examples of supported params.
+         // To note - appnexus supports additional values for our system that are not part of the ORTB spec.  If you want
          // to use these values, they will have to be declared in the bids[].params.video object instead using the appnexus syntax.
+         // Between the corresponding values of the mediaTypes.video and params.video objects, the properties in params.video will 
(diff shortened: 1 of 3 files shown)


patmmccann:
Thanks! Could you also update https://github.com/prebid/Prebid.js/blob/appnexus5-video-params/modules/appnexusBidAdapter.md to demonstrate playback method isn't required to be in the params?

jsnellbaker:
Made an update on the md file.

Fawke (approved):
LGTM

modules/appnexusBidAdapter.md
```diff
@@ -89,7 +89,16 @@ var adUnits = [
...
        video: {
          playerSize: [[300, 250]],
-         context: 'outstream'
+         context: 'outstream',
+         // certain ORTB 2.5 video values can be read from the mediatypes object; below are examples of supported params
+         // to note - appnexus supports additional values for our system that are not part of the ORTB spec.  If you want
+         // to use these values, they will have to be declared in the bids[].params.video object instead using the appnexus syntax.
+         minduration: 1,
+         maxduration: 60,
+         skip: 1,   // 1 - true, 0 - false
```
patmmccann, line 98: In this example you have skip and playback method double set, is this intentional?

jsnellbaker: The idea is that you can set both - but the ones in the bid.params will taken precedence due to unique values we have to support.

I'll add another comment to better clarify this point.

patmmccann (approved):

PR #6692, opened 2021-05-04
Log prebid events

mxdvl:
console log


mxdvl:
Raised against the wrong repo.

PR #6693, opened 2021-05-04
Dev Dependencies: update Sinon.js to most current workable version

ChrisHuie:
Update Sinon package to current. Right now it is on v4.5.0 and current is 10.0.1. This pr was branched from this current workable version update of dev dependencises #6663


mkendall07:
hey @ChrisHuie nice! Is this in draft because tests are failing?

ChrisHuie:
@mkendall07 Yeah. I have a base workable dependencies version submitted in this pr #6663 . Everything else to update to latest is going to take some lift though. So started tackling sinon with this one because I think that and webpack are probably the most important to update?

PR #6702, opened 2021-05-05
Revert "Prebid Core: add X-Prebid header to ajax requests "

patmmccann:
Reverts prebid/Prebid.js#6676

### DIFF
--- a/src/ajax.js
+++ b/src/ajax.js
@@ -75,12 +75,11 @@ export function ajaxBuilder(timeout = 3000, {request, done} = {}) {
       if (options.withCredentials) {
         x.withCredentials = true;
       }
-      x.setRequestHeader('x-prebid', 'pbjs/$prebid.version$');
       utils._each(options.customHeaders, (value, header) => {
         x.setRequestHeader(header, value);
       });
       if (options.preflight) {
-        x.setRequestHeader('x-requested-with', 'XMLHttpRequest');
+        x.setRequestHeader('X-Requested-With', 'XMLHttpRequest');
       }
       x.setRequestHeader('Content-Type', options.contentType || 'text/plain');
 


ChrisHuie:
I tried rerunning this a few times and looks like the errors are still firing for IE and edge. I was running into this yesterday too when I wasn't sure which pr may be causing the problem. I was rerunning prs and encountered these errors on pretty much all of them including one I merged a couple days before that was fine before but rerunning it gave these errors?

bretg:
Why doesn't X-Requested-With cause this CORS issue -- is it the options.preflight check?

patmmccann:
@ChrisHuie is it possible the sharethroughBidAdapter.js Circleci errors are unrelated to the errors @pycnvr demonstrated?

ChrisHuie:
@patmmccann I don't think they are unrelated. I'm really just not sure how these errors are now being carried over by CircleCI into prior merges that were fine before.

ChrisHuie:
@snapwich I could see that or at least that makes more sense to me now that you mention it.

src/ajax.js
```diff
@@ -75,12 +75,11 @@ export function ajaxBuilder(timeout = 3000, {request, done} = {}) {
       if (options.withCredentials) {
         x.withCredentials = true;
       }
-      x.setRequestHeader('x-prebid', 'pbjs/$prebid.version$');
       utils._each(options.customHeaders, (value, header) => {
         x.setRequestHeader(header, value);
       });
       if (options.preflight) {
-        x.setRequestHeader('x-requested-with', 'XMLHttpRequest');
+        x.setRequestHeader('X-Requested-With', 'XMLHttpRequest');
```
mkendall07, line 82: was changing capitalization here intentional?

mkendall07:
restarted the test let's see what happens.

src/ajax.js
patmmccann: yes https://github.com/prebid/Prebid.js/pull/6676#discussion_r624154890

patmmccann: whoops no

snapwich (approved):
i think we should merge this regardless of the tests. i don't think the preflight issue we're trying to avoid is represented in the test suite and i think the failures are for other test issues.

mkendall07 (approved):

PR #6706, opened 2021-05-05
Fixing flaky adapter unit tests

robertrmartinez:
## Type of change
- [X] Bugfix
Several Adapters were using Array Includes and Find which is not supported in IE. So tests were failing.

prebidManager adapter directly accessed localStorage, which is not allowed technically, and also is flaky.

Another adapter directly manipulates the DOM and appends child scripts. Which we certainly do not want to do for tests.

There are other adapters not in this PR which probably do this as well, but to keep this one relatively down, at least we have it somewhat passing. 

Still, there are fixes to be had, we get random jest timeouts at times.

## Description of change

### DIFF
--- a/test/spec/modules/prebidmanagerAnalyticsAdapter_spec.js
+++ b/test/spec/modules/prebidmanagerAnalyticsAdapter_spec.js
@@ -1,4 +1,6 @@
-import prebidmanagerAnalytics from 'modules/prebidmanagerAnalyticsAdapter.js';
+import prebidmanagerAnalytics, {
+  storage
+} from 'modules/prebidmanagerAnalyticsAdapter.js';
 import {expect} from 'chai';
 import {server} from 'test/mocks/xhr.js';
 import * as utils from 'src/utils.js';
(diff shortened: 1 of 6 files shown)


ChrisHuie (approved):

test/spec/modules/prebidmanagerAnalyticsAdapter_spec.js
```diff
@@ -105,19 +107,18 @@ describe('Prebid Manager Analytics Adapter', function () {
...
+      getDataFromLocalStorageStub.withArgs('pm_utm_content').returns('');
+      getDataFromLocalStorageStub.withArgs('pm_utm_source').returns('utm_source');
     });
     afterEach(function () {
-      localStorage.removeItem('pm_utm_source');
-      localStorage.removeItem('pm_utm_medium');
-      localStorage.removeItem('pm_utm_campaign');
-      localStorage.removeItem('pm_utm_term');
-      localStorage.removeItem('pm_utm_content');
+      getDataFromLocalStorageStub.restore();
```
patmmccann, line 121: ty so much for fixing this one! without a doubt the source of much committer frustration recently!

PR #6707, opened 2021-05-05
Prebid Core: Check for stale rendering

### LINKED ISSUE #6020: Limit number of times a single bid can be rendered to address increasing number of poor Auto-Refresh implementations
## Type of issue

Bug/ Feature Request

## Description
Across multiple distinct publisher Prebid integrations, we see a high number of bids being re-rendered at a consistent interval, suggesting there is an issue with auto-refreshing ad slots. After investigating a sample of these publisher integrations, we see many of these publishers have not have implemented auto-refresh correctly, in some cases just calling the _googletag.pubads.refresh()_ without getting new bids first. This is in part a publisher problem, but I thought the concept of a _used bid_ would prevent this. 

The majority of this problem is happening with re-renders that happen at 15s/30s/60s intervals, but I also see a good chunk that happen near immediately.  We've seen, in the worst case, individual bids be rendered thousands of times as a result of this problem.

Looking at the discussion in this other issue (https://github.com/prebid/Prebid.js/issues/4148), it seems like re-rendered ads is an undesirable behavior and others have seen similar behavior to what I've described here. Can additional protections be put in place to prevent this? There is certainly a lot of responsibility on the publisher's end to correctly implement auto-refresh, but we're seeing this across enough distinct publishers to believe this vulnerability should be addressed at the root.

## Steps to reproduce
You can go to sites like https://www.jta.org/?pbjs_debug=true or https://www.klbjfm.com/?pbjs_debug=true and let the page run for a bit. You'll see that ads with the same adId get called to render on a timer. This will happen with ads from any demand source. Also, if you call googletag.pubads.refresh() with any of the ad slots, it will just re-render the ad.

## Test page

https://www.jta.org/?pbjs_debug=true
https://www.klbjfm.com/?pbjs_debug=true

### Expected results
Ideally, once an ad has been rendered, it would no longer be eligible to be rendered again.
### Actual results
You can call renderAd on a bidId as many times as you want and the ad will continuously render. 
## Platform details

We see this happen on many versions of Prebid, not just the latest few.

pycnvr:
## Type of change

- [ ] Bugfix
- [x] Feature
- [ ] New bidder adapter   
- [ ] Build related changes
- [ ] CI related changes
- [ ] Other

## Description of change
Closes #6020 

Stale rendering means displaying an ad that's been shown before without getting new bids.  This change leverage auctionManager.addWinningBid, which is called for direct rendering by prebid.js, or by universal creatives in safeframes via postMessage.  This function sets the bid's status to RENDERED.  So if a bid's status is already RENDERED, then it's treated as stale.

When a stale rendering is detected, by default the following two behaviors are added:
* Add a warning message 
* Emit a new event called STALE_RENDER with the bid as the argument.

No other changes should occur.  So the ad is still rendered, and BID_WON event is emitted etc.

A new auctionOption is added to prevent stale ads from shown.  This is false by default.
```javascript
pbjs.setConfig({auctionOptions: {suppressStaleRender: true}})
````
When true, stale ads will skip the main rendering code, which generally means they are not rendered.

Note that natives and video ads are not affected by this change.

- email: pyang@conversantmedia.com

- doc PR prebid/prebid.github.io#2939

### DIFF
--- a/src/config.js
+++ b/src/config.js
@@ -278,7 +278,7 @@ export function newConfig() {
             return false
           }
         } else if (k === 'suppressStaleRender') {
-          if (typeof val[k] !== 'boolean') {
+          if (!utils.isBoolean(val[k])) {
             utils.logWarn(`Auction Options ${k} must be of type boolean`);
             return false;
           }
--- a/src/config.js
+++ b/src/config.js
@@ -265,7 +265,7 @@ export function newConfig() {
       }
 
       for (let k of Object.keys(val)) {
-        if (k !== 'secondaryBidders') {
+        if (k !== 'secondaryBidders' && k !== 'suppressStaleRender') {
           utils.logWarn(`Auction Options given an incorrect param: ${k}`)
           return false
         }
@@ -276,6 +276,11 @@ export function newConfig() {
           } else if (!val[k].every(utils.isStr)) {
             utils.logWarn(`Auction Options ${k} must be only string`);
             return false
+          }
+        } else if (k === 'suppressStaleRender') {
+          if (typeof val[k] !== 'boolean') {
+            utils.logWarn(`Auction Options ${k} must be of type boolean`);
+            return false;
           }
         }
       }
--- a/src/prebid.js
+++ b/src/prebid.js
@@ -23,7 +23,7 @@ const events = require('./events.js');
 const { triggerUserSyncs } = userSync;
 
 /* private variables */
-const { ADD_AD_UNITS, BID_WON, REQUEST_BIDS, SET_TARGETING, AD_RENDER_FAILED } = CONSTANTS.EVENTS;
+const { ADD_AD_UNITS, BID_WON, REQUEST_BIDS, SET_TARGETING, AD_RENDER_FAILED, STALE_RENDER } = CONSTANTS.EVENTS;
 const { PREVENT_WRITING_ON_MAIN_DOCUMENT, NO_AD, EXCEPTION, CANNOT_FIND_AD, MISSING_DOC_OR_ADID } = CONSTANTS.AD_RENDER_FAILED_REASON;
 
 const eventValidators = {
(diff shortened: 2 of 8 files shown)

### CODE REVIEW

**Comment by patmmccann** [comment]
This would still be an issue on video,  correct?

**Comment by pycnvr** [comment]
@patmmccann Yeah, videos go thru a different path all together.

**Review by jsnellbaker — APPROVED** [review]
As far as I can tell, I think this looks okay.

File: src/prebid.js [hunk 1]
```diff
@@ -390,63 +390,75 @@ $$PREBID_GLOBAL$$.renderAd = function (doc, id, options) {
...
-        } else if (ad) {
-          // will check if browser is firefox and below version 67, if so execute special doc.open()
-          // for details see: https://github.com/prebid/Prebid.js/pull/3524
-          // TODO remove this browser specific code at later date (when Firefox < 67 usage is mostly gone)
-          if (navigator.userAgent && navigator.userAgent.toLowerCase().indexOf('firefox/') > -1) {
-            const firefoxVerRegx = /firefox\/([\d\.]+)/;
-            let firefoxVer = navigator.userAgent.toLowerCase().match(firefoxVerRegx)[1]; // grabs the text in the 1st matching group
-            if (firefoxVer && parseInt(firefoxVer, 10) < 67) {
-              doc.open('text/html', 'replace');
+        if (shouldRender) {
```
mmoschovas, line 404: Do you think for the case in which shouldRender = false there should be an emitRenderFail with a new STALE_RENDER value so that the adRenderFailed event also is triggered?

src/config.js
```diff
@@ -277,6 +277,11 @@ export function newConfig() {
             utils.logWarn(`Auction Options ${k} must be only string`);
             return false
           }
+        } else if (k === 'suppressStaleRender') {
+          if (typeof val[k] !== 'boolean') {
```
mmoschovas, line 281: Nitpicking a bit but I maybe consider using the utils function for type checks

mmoschovas (commented):
Good job here. Just a couple notes.

src/prebid.js
pycnvr: Good question.  It's a possibility that can be added later.  Since there is already a new event, it seems clearer to keep them distinct.

mmoschovas: Got it. The reason I had asked is that the new event fires regardless if the suppression is enabled, meaning it is not exactly indicative of a failed render. But if it is not a requirement or there does not seem to be a potential issue for pubs leveraging failed render data, then the code looks good to me

mmoschovas (approved):

pycnvr:
Thanks @mmoschovas.  I need to make a small correction in the test to remove a superfluous '.only' thanks to the your suggestion above.

src/config.js
pycnvr: Fixed.  Thanks.

pycnvr:
All set. Thanks.

mmoschovas:
Thanks! This looks good

PR #6708, opened 2021-05-05
Index Exchange Bid Adapter: BidFloor decimal precision should be 3 or less

punkiller:
## Type of change

- [ x ] Bugfix

## Description of change

The ixBidAdapter should not send bidfloor values with more than 3 decimal precision 

### DIFF
--- a/modules/ixBidAdapter.js
+++ b/modules/ixBidAdapter.js
@@ -125,7 +125,7 @@ function _applyFloor(bid, imp, mediaType) {
   let moduleFloor = null;
 
   if (bid.params.bidFloor && bid.params.bidFloorCur) {
-    adapterFloor = { floor: roundUp(bid.params.bidFloor, 2), currency: bid.params.bidFloorCur };
+    adapterFloor = { floor: bid.params.bidFloor, currency: bid.params.bidFloorCur };
   }
 
   if (utils.isFn(bid.getFloor)) {
@@ -148,11 +148,6 @@ function _applyFloor(bid, imp, mediaType) {
     }
   }
 
-  // fixing floor precision to 2
-  if (moduleFloor && moduleFloor.floor) {
-    moduleFloor.floor = roundUp(moduleFloor.floor, 2);
-  }
-
   if (adapterFloor && moduleFloor) {
     if (adapterFloor.currency !== moduleFloor.currency) {
       utils.logWarn('The bid floor currency mismatch between IX params and priceFloors module config');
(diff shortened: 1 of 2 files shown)


modules/ixBidAdapter.js
```diff
@@ -103,6 +104,15 @@ function bidToImp(bid) {
...
   return imp;
 }
+/**
+ * 
+ * @param {*} number 
+ * @param {*} precision 
+ * @returns 
+ */
+function roundUp(number, precision) {
+  return Math.ceil((parseFloat(number) * Math.pow(10, precision)).toFixed(1)) / Math.pow(10, precision);
```
patmmccann, line 113: can you round these down? rounding floors up seems potentially yield negative

roundUp(0.004,2)
0.01

punkiller: @patmmccann 
Should this also be changed in 
https://github.com/prebid/Prebid.js/blob/master/modules/priceFloors.js#L219 
?

patmmccann: thats a good point @bszekely1 why did you choose to round floors up there?

patmmccann:
fwiw your test failures are bc you need to repull master #6706 fixed them

modules/ixBidAdapter.js
bszekely1: we talked this over in the design phase of floors, though it's been some time since we thought about it but the logic. The only place where rounding would likely cause any type of concern is when bids are adjusted for bidAdjustment or Currency Conversion. If anything, rounding up at 4 precision points helps the bidder. We would just round up the 4th precision decimal based on the 5th precision digit: https://github.com/prebid/Prebid.js/blob/b2cbab460f4392ea0b87fd59d78f0e07434ad300/modules/priceFloors.js#L58

At the precision of 4 decimal points, it's well below the CPM bucket that prebid normally assigns when bids come back in.

bszekely1: also, 2 point precision rounding up is a not ideal since as you pointed out Patrick will cause very small numbers to equal a penny. At 4 points that is far enough away to compete fairly.

patmmccann:
Hey @punkiller any reason this is still in draft,  looks great and our company for one is eager to see it merged

punkiller:
Hi @patmmccann 
Thanks for reviewing, this is something that could also be fixed in our endpoint. 
That fix is being tested and if it works out, this merge wont be required.

cd0n0van:
Hi @patmmccann - as @punkiller mentioned, we were able to resolve this in our endpoint. We'll be closing this PR, but thank you for the quick review!

punkiller:
Thank you @robertrmartinez + @patmmccann , closing this as per  @cd0n0van 's guidance .

PR #6712, opened 2021-05-06
Adbookpsp Bid Adapter: add new bid adapter

wgalecki:
## Type of change
- New bidder adapter 

## Description of change
Added adapter that connects to the AdbookPSP demand sources

```
{
  bidder: 'adbookpsp',
  params: {
    placementId: 'example-placement-id',
  },
}
```
- hbsupport@fattail.com
- https://github.com/prebid/prebid.github.io/pull/2933

### DIFF
--- a/modules/adbookpspBidAdapter.js
+++ b/modules/adbookpspBidAdapter.js
@@ -19,8 +19,7 @@ const USER_ID_COOKIE_EXP = 2592000000; // lasts 30 days
 const BID_TTL = 300;
 const SUPPORTED_MEDIA_TYPES = [BANNER, VIDEO];
 const DEFAULT_CURRENCY = 'USD';
-const VIDEO_PARAMS = ['api', 'mimes', 'playbackmethod', 'protocols'];
-const BIDDER_VIDEO_PARAMS = [
+const VIDEO_PARAMS = [
   'mimes',
   'minduration',
   'maxduration',
@@ -262,32 +261,15 @@ function buildVideoObject(bidRequest) {
   };
 
   for (const param of VIDEO_PARAMS) {
-    const value = utils.deepAccess(bidRequest, `mediaTypes.video.${param}`);
-    if (validateParamValue(param, value)) {
-      videoObj[param] = value;
-    }
-  }
+    const paramsValue = utils.deepAccess(bidRequest, `params.video.${param}`);
+    const mediaTypeValue = utils.deepAccess(bidRequest, `mediaTypes.video.${param}`);
 
-  for (const param of BIDDER_VIDEO_PARAMS) {
-    const value = utils.deepAccess(bidRequest, `params.video.${param}`);
-    if (value) {
-      videoObj[param] = value;
+    if (paramsValue || mediaTypeValue) {
+      videoObj[param] = paramsValue || mediaTypeValue;
     }
   }
 
   return videoObj;
-}
-
-function validateParamValue(param, value) {
-  const result = utils.isArray(value);
-
-  if (!result) {
-    utils.logWarn(
-      `${BIDDER_CODE}: Ignoring param: ${param} expects an array, got ${typeof value}`
-    );
-  }
-
-  return result;
 }
 
 function buildImpExt(validBidRequest) {
--- a/modules/adbookpspBidAdapter.js
+++ b/modules/adbookpspBidAdapter.js
@@ -232,7 +232,7 @@ function buildMediaTypeObject(mediaType, bidRequest) {
     case VIDEO:
       return buildVideoObject(bidRequest);
     default:
-      utils.logWarn(`Unsupported media type ${mediaType}!`);
+      utils.logWarn(`${BIDDER_CODE}: Unsupported media type ${mediaType}!`);
   }
 }
 
@@ -254,7 +254,7 @@ function buildBannerObject(bidRequest) {
 }
 
 function buildVideoObject(bidRequest) {
-  const [w, h] = bidRequest.mediaTypes.video.playerSize[0];
+  const { w, h } = getVideoSize(bidRequest);
   let videoObj = {
     w,
     h,
@@ -262,7 +262,10 @@ function buildVideoObject(bidRequest) {
 
   for (const param of VIDEO_PARAMS) {
     const paramsValue = utils.deepAccess(bidRequest, `params.video.${param}`);
-    const mediaTypeValue = utils.deepAccess(bidRequest, `mediaTypes.video.${param}`);
+    const mediaTypeValue = utils.deepAccess(
+      bidRequest,
+      `mediaTypes.video.${param}`
+    );
 
     if (paramsValue || mediaTypeValue) {
       videoObj[param] = paramsValue || mediaTypeValue;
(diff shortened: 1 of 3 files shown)


msm0504:
Please fix the lint errors in this PR, so that the Circle CI tests will run. Double quotes need to be replaced with single quotes.

modules/adbookpspBidAdapter.md
```diff
@@ -0,0 +1,189 @@
...
+        playerSize: [400, 300],
+      },
+    },
+    bids: [
+      {
+        bidder: 'adbookpsp',
+        params: {
+          placementId: 'example-placement-id',
+          video: {
+            protocols: [2, 3],
```
patmmccann, line 95: Please move this to the ad unit

patmmccann (changes requested):

wgalecki:
I've updated this branch with the latest upstream changes but still some tests are not passing. Do you have any advice?

patmmccann:
Tests have passed

modules/adbookpspBidAdapter.js
```diff
@@ -0,0 +1,768 @@
...
+        adserverTargeting: targetingMap[bid.impid],
+        adUnitCode: bidRequest.tagid,
+        bidderRequestId: bidderRequestBody.id,
+        bidId: bid.id,
+        cpm: bid.price,
+        creativeId: bid.crid || bid.id,
+        currency: bidResponseCurrency || getBidderConfig('defaultCurrency'),
+        height: bid.h,
+        lineItemId: utils.deepAccess(bid, 'ext.liid'),
+        mediaType,
```
patmmccann, line 370: is this supposed to be mediaType: mediaType, ?

modules/adbookpspBidAdapter.js
```diff
@@ -0,0 +1,768 @@
...
+      if (mediaType === VIDEO) {
+        prebidBid = {
+          ...prebidBid,
+          ...getVideoSpecificParams(bidRequest, bid),
+        };
+      }
+
+      return prebidBid;
+    } catch (error) {
+      utils.logError(error);
```
patmmccann, line 394: please add bidder code similar to line 352

modules/adbookpspBidAdapter.js
```diff
@@ -0,0 +1,768 @@
...
+    values[bid.impid].lineItemIds.push(bid.ext.liid);
+    values[bid.impid].dealIds.push(bid.dealid);
+    values[bid.impid].adIds.push(bid.adid);
+  });
+
+  const targetingMap = {};
+
+  for (const id of impIds) {
+    targetingMap[id] = {
+      hb_liid_adbookpsp: values[id].lineItemIds.join(TARGETING_VALUE_SEPARATOR),
```
patmmccann, line 430: what is happening here

modules/adbookpspBidAdapter.js
```diff
@@ -0,0 +1,768 @@
...
+      hb_deal_adbookpsp: values[id].dealIds.join(TARGETING_VALUE_SEPARATOR),
+      hb_adid_c_adbookpsp: values[id].adIds.join(TARGETING_VALUE_SEPARATOR),
```
patmmccann, line 432: is _c a typo?

modules/adbookpspBidAdapter.js
```diff
@@ -0,0 +1,768 @@
...
+    utils.deepAccess(bidRequest, 'mediaTypes.banner.sizes', [])
+  );
+
+  return value;
+}
+
+function isValidVideoRequest(bidRequest) {
+  const value =
+    utils.isArray(utils.deepAccess(bidRequest, 'mediaTypes.video.mimes')) &&
+    validateSizes(utils.deepAccess(bidRequest, 'mediaTypes.video.playerSize'));
```
patmmccann, line 467: what if a publisher just sets h&w

patmmccann (changes requested):
a few minor requests

modules/adbookpspBidAdapter.js
wgalecki: No, this is a separate key with a concatenated list of ad ids (hence the c).

modules/adbookpspBidAdapter.js
wgalecki: Actually, on a second thought, since I use this notation in multiple places and it is being transpiled anyway, I'd like to leave it as is if that's ok with you.

modules/adbookpspBidAdapter.js
wgalecki: Sure, will do

modules/adbookpspBidAdapter.js
wgalecki: Good catch, thank you.

modules/adbookpspBidAdapter.js
```diff
@@ -0,0 +1,796 @@
...
+
+  return targetingMap;
+}
+
+/**
+ * VALIDATION
+ */
+
+function hasRequiredParams(bidRequest) {
+  const value =
```
msm0504, line 461: This condition is only checking for nulls. This will return `true` if all 3 are `undefined`. This can be changed to the following, unless `0` needs to be accepted as a valid placementId or orgId:
```
utils.deepAccess(bidRequest, 'params.placementId') ||  utils.deepAccess(bidRequest, 'params.orgId') || getBidderConfig('orgId')
```

modules/adbookpspBidAdapter.js
```diff
@@ -0,0 +1,796 @@
...
+        cpm: bid.price,
+        creativeId: bid.crid || bid.id,
+        currency: bidResponseCurrency || getBidderConfig('defaultCurrency'),
+        height: bid.h,
+        lineItemId: utils.deepAccess(bid, 'ext.liid'),
+        mediaType,
+        meta: {
+          advertiserDomains: bid.adomain,
+          mediaType,
+          primaryCatId: categories[0],
```
msm0504, line 391: Based on line 373 above, it's possible that `categories` is an empty list.

msm0504 (changes requested):

modules/adbookpspBidAdapter.js
wgalecki: In that case `primaryCatId` will be `undefined` and `secondaryCatIds` will be `[]`. I think this is fine, isn't it?

modules/adbookpspBidAdapter.js
wgalecki: It's actually checking for both `null` and `undefined`. Using `!== null` would check for nulls only, `!= null` returns `false` for both `null` and `undefined` since it ignores types.

modules/adbookpspBidAdapter.js
patmmccann: Do you want to check said script into integration examples?

wgalecki: I meant the script rendered by GAM, I'm not sure we're interested in adding an example right now. Is this something you require?

patmmccann: No, but please add prominent disclosure in your doc repo pull and your .md file that custom gam creatives are required and that this bidder only works in its own order and that the prebid universal creative does not work etc. Please be very explicit and redundant.

wgalecki: I've expanded the note about the adapter requirements in the doc pull request and brought the note to the readme file here.

msm0504 (approved):

msm0504:
@patmmccann Are there any changes that you're still waiting on or can this be approved?

patmmccann (approved):

PR #6723, opened 2021-05-07
Pbs bid adapter: constants import styling for webpack v5 upgrade

khatibda:
Updating the prebid server adapter to use proper syntax when importing a default json module.

Couple things to note:
- the pbs adapter test file already uses proper import syntax for this module, so it didn't need updating.
- just fyi, in this repo, there is no common convention for how adapters are importing the constants.json module.  some adapters are using default import ("import CONSTANTS"), and some are using the named export approach ("import { EVENTS }").  The latter appears to be frowned upon by Webpack 5 and only default import will be available soon (according to Webpack 5 warnings).  I'm only fixing this file for now, but something to consider more broadly.

related issue:
https://github.com/prebid/Prebid.js/issues/6677


ChrisHuie:
@khatibda thanks for this. We are currently planning on the roadmap to upgrade to webpack 5 so aligning these conventions would be a good idea!

ChrisHuie (approved):

PR #6729, opened 2021-05-10
Ogury Bid Adapter: add new adapter

jogury:
## Type of change

- [ ] Bugfix
- [ ] Feature
- [ x] New bidder adapter   
- [ ] Build related changes
- [ ] CI related changes
- [ ] Other

## Description of change

Initialization of the new Ogury Bid Adapter

```
{
  bidder: 'ogury',
  params: {
      assetKey: 'OGY-CA41D116484F',
      adUnitId: '2c4d61d0-90aa-0139-0cda-0242ac120004'
  }
}
```

- contact email of the adapter’s maintainer: web.inventory@ogury.co

https://github.com/prebid/prebid.github.io/pull/2943

### DIFF
--- a/modules/oguryBidAdapter.js
+++ b/modules/oguryBidAdapter.js
@@ -80,7 +80,7 @@ function interpretResponse(openRtbBidResponse) {
   if (!openRtbBidResponse ||
     !openRtbBidResponse.body ||
     typeof openRtbBidResponse.body != 'object' ||
-    Object.entries(openRtbBidResponse.body).length === 0) {
+    Object.keys(openRtbBidResponse.body).length === 0) {
     logWarn('no response or body is malformed');
     return [];
   }
--- a/modules/oguryBidAdapter.js
+++ b/modules/oguryBidAdapter.js
@@ -98,6 +98,9 @@ function interpretResponse(openRtbBidResponse) {
         creativeId: bid.id,
         netRevenue: true,
         ttl: 60,
+        meta: {
+          advertiserDomains: bid.adomain
+        }
       };
 
       bidResponse.ad = bid.adm;
--- a/modules/oguryBidAdapter.js
+++ b/modules/oguryBidAdapter.js
@@ -1,0 +1,151 @@
+'use strict';
+
+import { BANNER } from '../src/mediaTypes.js';
+import { getAdUnitSizes, logWarn, isFn } from '../src/utils.js';
+import { registerBidder } from '../src/adapters/bidderFactory.js';
+
+const BIDDER_CODE = 'ogury';
+const DEFAULT_TIMEOUT = 1000;
+const BID_HOST = 'https://webmobile.presage.io/api/header-bidding-request';
+
+function isBidRequestValid(bid) {
+  const adUnitSizes = getAdUnitSizes(bid);
+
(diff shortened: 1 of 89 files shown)


modules/oguryBidAdapter.js
```diff
@@ -0,0 +1,151 @@
...
+
+function getFloor(bid) {
+  if (isFn(bid.getFloor)) {
+    let floorResult = bid.getFloor({
+      currency: 'USD',
+      mediaType: 'banner',
+      size: '*'
+    });
+
+    return floorResult.floor;
```
robertrmartinez, line 133: it is possible that if floors is set in non USD, and publisher does not have currency module included / enabled, that this call will return non USD floor.

It is a good idea to verify that the currency returned is in USD in this case, just to be safe!

Maybe something like this?
```js
return floorResult.currency === 'USD' ? floorResult.floor : 0;
```

modules/oguryBidAdapter.js
```diff
@@ -0,0 +1,151 @@
...
+
+function onTimeout(timeoutData) {
+  logWarn(timeoutData);
+}
+
+function onBidWon(bid) {
+}
+
+function onSetTargeting(bid) {
+  logWarn(bid);
```
robertrmartinez, line 122: Not sure this is something we need?

Log a warning every single time there is a timeout or setTargeting?

modules/oguryBidAdapter.js
```diff
@@ -0,0 +1,151 @@
...
+        ttl: 60,
+      };
+
+      bidResponse.ad = bid.adm;
+
+      bidResponses.push(bidResponse);
+    });
+  });
+
+  if (bidResponses.length === 0) logWarn('bid responses is empty');
```
robertrmartinez, line 109: Same here, we do not need to log a warning just because there is a no bid.

robertrmartinez (changes requested):
Couple small inputs.

Polluting the console with warnings that are not actual warnings is not ideal.

modules/oguryBidAdapter.js
mbouallagui: I agree. I added the check you suggested and a new unit test to cover the case of unsupported currency 
it("should handle bidFloor when currency is not USD")

modules/oguryBidAdapter.js
mbouallagui: In fact, that was added during some debugging and is not needed anymore. I removed the logs as well as useless empty functions.

modules/oguryBidAdapter.js
mbouallagui: same as above.

robertrmartinez:
@jogury Thanks for addressing the comments, please just undo the changes you made to package-lock.json!

mbouallagui:
@robertrmartinez , I've just removed the changes in package-lock.json from the PR. Sorry about that.

patmmccann (changes requested):
You must return advertiserDomains

antoinezaz:
@patmmccann thanks for your comment, now we return the advertiserDomains

patmmccann:
Lgtm ; thanks! @robertrmartinez is this ready to merge or any other suggestions?

robertrmartinez (approved):
Thanks for all the work to get this done! 👍

Issue #6361, opened 2021-02-25
Remove requirement for client-side adapter to run server-side


bretg:
## Type of issue
enhancement

## Description
One of the most common complaints about Prebid.js is that the resulting javascript package size is too big.  

Moving code server-side should be a way to reduce the package size, but there are a couple of bidder-specific elements in the BidAdapter spec that are needed even when the bidder runs server-side:
- GVL ID
- Aliases
- supportedMediaTypes

## Proposal

Seems like we may have talked about this years ago, but rather than split out separate minimal placeholder adapters, perhaps we could implement some kind of 'conditional compilation'. I'm not a javascript coder, but back the glory days of compiled languages, there were compiler flags like "#ifdef SERVER   ...   #endifdef"

So for instance

```
export const spec = {
  code: BIDDER_CODE,
  gvlid: GVLID,
  aliases: [
    { code: 'appnexusAst', gvlid: 32 },
    { code: 'brealtime' },
    { code: 'emxdigital', gvlid: 183 },
    { code: 'pagescience' },
    { code: 'defymedia' },
    { code: 'gourmetads' },
    { code: 'matomy' },
    { code: 'featureforward' },
    { code: 'oftmedia' },
    { code: 'districtm', gvlid: 144 },
    { code: 'adasta' },
    { code: 'beintoo', gvlid: 618 },
  ],
  supportedMediaTypes: [BANNER, VIDEO, NATIVE]

#if-not-defined SERVER-ONLY
... rest of code ...
#end-if
```

If something like this is possible for PBJS, it might be helpful for other scenarios as well. e.g. we could establish standard flags: BANNER, VIDEO, NATIVE and adapters could conditionally exclude code not needed by the publisher.

jsnellbaker:
Just to highlight it, there is also a `transformBidParams` function used by (roughly) a dozen bid adapters to update the formatting used by the bid params to make them compliant for PBS.  

Below are some links to where this is being used (for reference):
https://github.com/prebid/Prebid.js/blob/master/modules/prebidServerBidAdapter/index.js#L613
https://github.com/prebid/Prebid.js/blob/master/modules/appnexusBidAdapter.js#L338

muuki88:
@bretg there is something in the web as well :smile: I saw it on the the prebid-outstream project: [webpack-conditional-loader](https://www.npmjs.com/package/webpack-conditional-loader).

This would at least work for NPM modules, but requires all adapters to include these special comments for this loader.
[...]

PR #6747, opened 2021-05-12
Criteo Bid Adapter: add comment to explain how to remove dependency on RSA validation lib

leonardlabat:

## Description of change
This was asked by @patmmccann.
The goal is to explain for a publisher which won't use our fastbid feature, how to remove the dependency on the RSA validation lib (and thus reduce the generated bundle file).

### DIFF
--- a/modules/criteoBidAdapter.js
+++ b/modules/criteoBidAdapter.js
@@ -4,7 +4,7 @@ import {config} from '../src/config.js';
 import {BANNER, NATIVE, VIDEO} from '../src/mediaTypes.js';
 import * as utils from '../src/utils.js';
 import find from 'core-js-pure/features/array/find.js';
-import { verify } from 'criteo-direct-rsa-validate/build/verify.js';
+import { verify } from 'criteo-direct-rsa-validate/build/verify.js'; // ref#2
 import { getStorageManager } from '../src/storageManager.js';
 
 const GVLID = 91;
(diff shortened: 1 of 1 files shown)


ChrisHuie:
@leonardlabat it looks like this pr may have been pulled last week when we had a rogue IE character and some a few other tests we had to clean up. Pulling recent commits should fix errors on CircleCI

leonardlabat:
Hello @ChrisHuie 

I've created my commit based on prebid-5.0 branch (as my change targets PB5 version). And it seems that prebid-5.0 has not been updated since 8 days. So I guess, it may have not the changes you mentionned here.
So, I'm not sure what I should do 🤔

ChrisHuie:
@leonardlabat Sorry. I missed this was on the 5.0 branch! Ignore what I just said. lol. 
Yeah. I haven't done anything on the 5.0 branch this week or last week so not too sure on these test failures. I can dig into recent 5.0 commits to see if there are any flaky tests I run into.

patmmccann:
@leonardlabat this is extremely helpful, thank you! Would it be possible to make it a build time flag, similar to https://github.com/prebid/Prebid.js/pull/6016 ?

allanjun:
hi @patmmccann we are actually struggling to make it work on webpack. We have also seen related discussions on  https://github.com/prebid/Prebid.js/pull/6079 and https://github.com/prebid/Prebid.js/issues/6361 which cover similar needs, we can certainly revisit this once these two discussions are solved, would that work ?

patmmccann:
yes, good to know, ty

patmmccann (approved):

Issue #6640, opened 2021-04-21
Merge Pubcommon and SharedId user id sub adapters. 


jdwieland8282:
## Type of issue
Feature Request

## Description
We have two functionally similar user id sub adapters (PubCommon and SharedID).  We would like to merge the two w/o impacting the DSPs who read "source=pubcid.org" in pb.js 5.0

- We will deprecate the SharedID sub adapter.  
- We will rename the PubCommon sub adapter to "SharedId"
- The PubCommon Source value will remain "pubcid.org"

### Expected results

- We expect publishers who have deployed Pubcommon to not be impacted in any way. 
- We expect publishers who have deployed SharedId to not be impacted but encouraged to remove sharedid from their pb.js build & migrate to the newly renamed Pubcommon sub adapter.  

remove the following sharedid files
 -  modules/sharedIdSystem.js
 - test/spec/modules/sharedIdSystem_spec.js
 - modules/sharedIdSystem.md

remove pubcommon references in
 - test/spec/modules/userId_spec.js
 - modules/userId/userId.md
 - modules/userId/eids.js
 - modules/userId/eids.md

remove sharedid references in:
 - modules/pubCommonIdSystem.js
 - test/spec/modules/eids_spec.js

rename references to sharedid.org to pubcid.org in:
 - test/spec/modules/logicadBidAdapter_spec.js
 - test/spec/modules/betweenBidAdapter_spec.js
 - integrationExamples/gpt/userId_example.html
 - modules/betweenBidAdapter.js
 - integrationExamples/gpt/idImportLibrary_example.html
 - modules/districtmDMXBidAdapter.js
 - test/spec/modules/sharethroughBidAdapter_spec.js
 - test/spec/modules/sonobiBidAdapter_spec.js
 - modules/openxBidAdapter.js
 - test/spec/modules/rubiconBidAdapter_spec.js
 - modules/sharethroughBidAdapter.js
 - test/spec/modules/livewrappedBidAdapter_spec.js
 - modules/ucfunnelBidAdapter.js
 - test/spec/modules/ucfunnelBidAdapter_spec.js
 - test/spec/modules/apacdexBidAdapter_spec.js
 - test/spec/modules/districtmDmxBidAdapter_spec.js
 - modules/rubiconBidAdapter.js
 - test/spec/modules/adagioBidAdapter_spec.js
 - test/spec/modules/openxBidAdapter_spec.js
 - modules/ixBidAdapter.js
[...]

PR #6748, opened 2021-05-12
Merge pubcommon and sharedid #6640

NothingIsMatter:
## Type of change
- [x] Feature

## Other information
Issue link: https://github.com/prebid/Prebid.js/issues/6640


smenzer (approved):
This still seems incredibly confusing to me. The name of the ID module file is `sharedId` but all code references in prebid are now to `pubCommonId`. Pulling in the module is even more strange:

```
import {pubCommonIdSubmodule} from 'modules/sharedIdSystem.js';
```

I understand the desire to avoid publisher changes, but why not keep as much in the code under the sharedId name _except_ for the few external pieces, like the source domain in the eids array and the name of the module used for publisher configuration?

Also, in the Issue, it mentions as a requirement:
>We expect publishers who have deployed SharedId to not be impacted but encouraged to remove sharedid from their pb.js build & migrate to the newly renamed Pubcommon sub adapter.

I don't see how publishers who are using `sharedId` as an ID module wouldn't be impacted...it seems like it will just stop working. It's not clear to me from the discussion in the issue if this requirement was scrapped or not, so just calling out that in the code, it has been.

mmoschovas (changes requested):
@NothingIsMatter 

I may be mistaken, but it does not seem to me that this is aligned with the issue definition. Both pubCommon and sharedId should be usable as config options with pubCommon as an alias name for backwards compatibility. If I am correct in that assessment, I would suggest updating the file so that the name is set to sharedId and a new alias field can exist similar to below. 

```
...
export const pubCommonIdSubmodule = {
  /**
   * used to link submodule with config
   * @type {string}
   */
  name: MODULE_NAME,
  aliasName: "pubCommonId",
...
```

This would also require an update to the userId index.js file to look for name then alias if it exists. Something like the following:

```
function updateSubmodules() {
  ...
  submodules = addedSubmodules.map(i => {
    const submoduleConfig = find(configs, j => j.name && (j.name.toLowerCase() === i.name.toLowerCase() || 
      (i.aliasName && j.name.toLowerCase() === i.aliasName.toLowerCase())));
    if (submoduleConfig && i.name !== submoduleConfig.name) submoduleConfig.name = i.name;
...
```

This will allow the config to be defined asa either sharedId or pubCommonId, and the userId module will overwrite the alias name with the submodule name.

NothingIsMatter:
@mmoschovas thx for comment.  I updated the PR, can you take a look ?

mmoschovas:
This looks to be functioning correctly. The only issue I have is that I believe we should perhaps limit the pubCommon references where applicable. It is confusing to have the file and module reference sharedId but internal module comments and such referencing pubCommonId. Some examples:

```
* This module adds PubCommonId to the User ID module 
* The {@link module:modules/userId} module is required 
* @module modules/pubCommonIdSystem 
* @requires module:modules/userId
```

```
const storage = getStorageManager(GVLID, 'pubCommonId');
```

```
utils.logInfo('PubCommonId: IDs not provided for coppa requests, exiting PubCommonId');
```

The only reference I believe that should stay is that of the window object and the new alias. I think this would make this a little less confusing when building using the sharedIdSystem, but would appreciate others opinions on this. @jdwieland8282 what are your thoughts?

mmoschovas:
Also this merge conflict will need to be resolved

patmmccann:
Is this supposed to be on the 5.0 branch?

jdwieland8282:
yes

patmmccann:
This is not on the 5.0 branch,  please resubmit

NothingIsMatter:
@patmmccann I need to pull from prebid 5.0 branch ?

jdwieland8282:
thats correct @NothingIsMatter , this PR should have been made against that branch.

PR #6749, opened 2021-05-12
PubMatic Adapter:  add support for FloCID

pm-shashank-jain:
## Type of change
- [x] Feature

## Description of change
Reading of FloC Id from Floc userId module and passing it to PubMatic SSP in user.eids and user.data.segment


patmmccann:
does it make sense to write it to those places in the prebid fpd object rather than just the one that goes to your endpoint? It sseems this work is nearly 100% likely to get duplicated shortly

pm-shashank-jain:
> does it make sense to write it to those places in the prebid fpd object rather than just the one that goes to your endpoint? It sseems this work is nearly 100% likely to get duplicated shortly

@patmmccann can you please elaborate, didn't understand it!
If it helps, let me just explain what I am doing is, reading the FloC ID from userId in validBids passed to adapter and passing it in user.data.segment and user.eids to PubMatic SSP. Doing this until FloC data position becomes standard i.e. either in eids or user data-segments or some place else.

patmmccann:
I am suggesting that instead of copying it there for your endpoint payload, you copy it there for everyone

abhinavsinha001:
@patmmccann  As per https://github.com/prebid/Prebid.js/issues/6510, I believe there is no consensus on  where FloC id should end up in oRTB and for the current origin trial adapters can choose where to pass this ID.
Hence we decided not to add it to common object as of now, and hopefully move the code to common FPD ortb2 object once we have location finalised.

patmmccann:
Fair enough

modules/pubmaticBidAdapter.js
```diff
@@ -708,8 +713,67 @@ function _addFloorFromFloorModule(impObj, bid) {
...
+      payload.user.data = [];
+    }
+    payload.user.data.push(flocObject);
+  }
+}
+
 function _handleEids(payload, validBidRequests) {
-  const bidUserIdAsEids = utils.deepAccess(validBidRequests, '0.userIdAsEids');
+  var bidUserIdAsEids = utils.deepAccess(validBidRequests, '0.userIdAsEids');
+  var flocObject = _getFlocId(validBidRequests, FLOC_FORMAT.EID);
```
musikele, line 770: consider using `let` instead

musikele (commented):
just a very small note

modules/pubmaticBidAdapter.js
pm-shashank-jain: Done

musikele (approved):
LGTM

PR #6753, opened 2021-05-12
Appnexus bid adapter: create stub for meta.advertiserDomains

jsnellbaker:
## Type of change
- [x] Other

## Description of change
Related to https://github.com/prebid/Prebid.js/issues/6650

This PR creates a stub for the `bid.meta.advertiserDomains` field for the appnexus bid adapter that needs to be specified/stubbed as part of 5.0 compliance.

### DIFF
--- a/modules/appnexusBidAdapter.js
+++ b/modules/appnexusBidAdapter.js
@@ -578,6 +578,11 @@ function newBid(serverBid, rtbBid, bidderRequest) {
     }
   };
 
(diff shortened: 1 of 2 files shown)


patmmccann (approved):

patmmccann:
I think it might be useful for some publishers setting hb_adomain as an ad server key to copy over advertiser_id temporarily into that field until you guys support this, that's our plan internally. Probably doesn't make a lot of sense to do that generally though.

Fawke (approved):

PR #6758, opened 2021-05-13
ixBidAdapter: adds support for floc

umakajan:
## Type of change

- [ ] Bugfix
- [x] Feature
- [ ] New bidder adapter   
- [ ] Build related changes
- [ ] CI related changes
- [ ] Other

## Description of change

In this PR, we are adding support to ixBidAdapter for sending the `flocId` through `eids` by reading it from userID.


umakajan:
I just merged master into this branch, @ChrisHuie would you be able to advise on how to resolve the tests failed on CircleCI?

modules/ixBidAdapter.js
```diff
@@ -311,6 +313,19 @@ function getEidInfo(allEids) {
       }
     }
   }
+
+  const isValidFlocId = utils.deepAccess(flocData, 'id') && utils.deepAccess(flocData, 'version');
```
mlb7687, line 317: deepAccess is a nice utility for deeply nested checks, but use on shallow objects like this are just going to contribute to slower evaluation. This does a string.split on every call[1]. Can't we just do `const isValidFlocId = flocData && flocData.id && flocData.version`?

[1] https://github.com/developit/dlv/blob/master/index.js

modules/ixBidAdapter.js
```diff
@@ -311,6 +313,19 @@ function getEidInfo(allEids) {
...
+
+  if (isValidFlocId) {
+    var flocEid = {
```
mlb7687, line 320: Prefer const/let and be consistent within the codebase

modules/ixBidAdapter.js
```diff
@@ -285,17 +285,19 @@ function getBidRequest(id, impressions) {
...
  * for our purposes, e.g. add rtiPartner
  * @param {array} allEids userIdAsEids passed in by prebid
+ * @param {object} flocId flocId passed in by prebid
  * @return {object} contains toSend (eids to send to the adserver) and seenSources (used to filter
  *                  identity info from IX Library)
  */
-function getEidInfo(allEids) {
+function getEidInfo(allEids, flocData) {
   // determines which eids we send and the rtiPartner field in ext
   var sourceRTIMapping = {
```
mlb7687, line 294: Any reason not to have this be a file const, so that we do not have the cost of instantiating on every call to getEidInfo? Could lazily initialize if this is not something we'd expect to run on most flows.

Also prefer const/let

modules/ixBidAdapter.js
```diff
@@ -311,6 +313,19 @@ function getEidInfo(allEids) {
...
+  const isValidFlocId = utils.deepAccess(flocData, 'id') && utils.deepAccess(flocData, 'version');
+
+  if (isValidFlocId) {
+    var flocEid = {
+      'source': 'chrome.com',
+      'uids': [{'id': flocData.id, 'ext': {'rtiPartner': 'flocId', 'ver': flocData.version}}]
+    };
+
+    toSend.push(flocEid);
+    seenSources['chrome.com'] = 1;
```
mlb7687, line 326: Any reason for this to be an int instead of bool? Performance difference is probably negligible, but seems like a more appropriate datatype.

mlb7687 (changes requested):

modules/ixBidAdapter.js
dubyte: Thank, yest that works for me.

modules/ixBidAdapter.js
umakajan: We are following the existing datatype as [used here](https://github.com/prebid/Prebid.js/blob/d965e83ae64eabfbae7287d4324da3ecda15354c/modules/ixBidAdapter.js#L307) with `seenSources`

modules/ixBidAdapter.js
dubyte: ok, thanks

modules/ixBidAdapter.js
mlb7687: Ack, I guess that leads to "Why is that object using ints?"
I realize it isn't clearly in the scope of this PR, but it caught my eye and could be a place for improvement if there isn't a good reason for the object to be an int. It looks like this is getting coerced into a boolean for a check[1] anyway.

[1] https://github.com/prebid/Prebid.js/blob/d965e83ae64eabfbae7287d4324da3ecda15354c/modules/ixBidAdapter.js#L359

modules/ixBidAdapter.js
dubyte: the code has been updated to use const/let in the scope of the pull request.

modules/ixBidAdapter.js
umakajan: That's a good point and since it's since it's a fairly simple change, I'll make that update

modules/ixBidAdapter.js
```diff
@@ -285,24 +285,26 @@ function getBidRequest(id, impressions) {
...
  * @param {array} allEids userIdAsEids passed in by prebid
+ * @param {object} flocId flocId passed in by prebid
  * @return {object} contains toSend (eids to send to the adserver) and seenSources (used to filter
  *                  identity info from IX Library)
  */
-function getEidInfo(allEids) {
+function getEidInfo(allEids, flocData) {
   // determines which eids we send and the rtiPartner field in ext
-  var sourceRTIMapping = {
+  const sourceRTIMapping = {
```
mlb7687, line 294: This and the `PROVIDERS` variable in `_getUserIds` could still be a file-level const if we wanted to avoid re-creating on every call.

modules/ixBidAdapter.js
```diff
@@ -285,24 +285,26 @@ function getBidRequest(id, impressions) {
...
+    'uidapi.com': 'UID2',
+    'adserver.org': 'TDID'
   };
-  var toSend = [];
-  var seenSources = {};
+  let toSend = [];
+  let seenSources = {};
   if (utils.isArray(allEids)) {
-    for (var i = 0; i < allEids.length; i++) {
+    for (let i = 0; i < allEids.length; i++) {
```
mlb7687, line 305: You can use a for-of loop here since you are only ever using i to access allEids[i]

modules/ixBidAdapter.js
```diff
@@ -583,7 +593,9 @@ function _getUserIds(bidRequest) {
     'sharedid',
     'tapadId',
     'quantcastId',
-    'pubcid'
+    'pubcid',
+    'TDID',
+    'flocId'
   ]
 
   return PROVIDERS.filter(provider => utils.deepAccess(userIds, provider))
```
mlb7687, line 601: Can simplify to `return PROVIDERS.filter(provider => userIds[provider])`

mlb7687 (changes requested):

mlb7687:
LGTM!

umakajan:
@patmmccann @ChrisHuie anything we need to add to this to PR to receive review?

patmmccann (approved):

Issue #6738, opened 2021-05-11
ID modules initialized before requestBidsHook do not receive consentData


smenzer:
## Type of issue
Bug

## Description
There are a few ways that ID submodules are initialized - via the `requestBidsHook` and in the `getUserIds()`, `getUserIdsAsEids()`, and `refreshUserIds()` functions. Whichever one of these that are called first is what dictates how the submodule is initialized and subsequent calls to another function doesn't reinitialize them. 

The problem is that if a call to `getUserIds()` or `getUserIdsAsEids()` is called before the `requestBidsHook` fires, then it's not guaranteed that the CMP would have returned consent data to prebid-core yet and the submodules get initialized with `consentData` being `undefined`. When the `requestBidsHook` fires, the submodules are already initialized and are not called again. This means that the submodules never get consent data and, assuming the submodule follows GDPR properly, there would be no ID available from the provider, even if the user did consent.

## Steps to reproduce
I have two test pages to show the behavior

1. Consent Data correctly passed into the ID5 submodule: https://samples.id5.io/prebid/basic
    - In the Console, you can see that a valid ID5 ID is made available with `pbjs.getUserIds()`
    - You can also see Network calls to ID5 with the gdpr consent string in the body of the payload

2. Consent Data missing because `getUserIdsAsEids()` is called at the bottom of the page prior to the `requestBidsHook`: https://samples.id5.io/prebid/basic-with-bug.html
    - You'll see in the box that the ID5 ID is "0", indicating no consent from ID5
    - You can see this in the Network calls to ID5 that the gdpr flag is set to `0` and no gdpr consent string is passed

If you are flipping back and forth between these pages, make sure to refresh a few times to clear out the cache - the ID5 module will make new calls to ID5 every 10 seconds.

### Expected results in the "Prebid Bug" page
I'm open to two different approaches here:

#### 1) Delay the functions until after the modules are initialized properly with consent data
* If a call to `getUserIds()` or `getUserIdsAsEids()` is called before the submodules are initialized by the `requestBidsHook`, they should be put in a queue and responses delayed until the modules are initialized, then they return like normal
* If there is cached data available, then the functions could return the cached data only and skip the rest of the initialization process

#### 2) Re-initialize with consent data if it becomes available
* The initial call to `getUserIdsAsEids()` should call ID5 if there is no data in cache, or use the ID from cache if available (this works as expected today)
[...]

PR #6760, opened 2021-05-14
User id module: force calls to getId if there was previously no consent data stored

smenzer:
## Type of change
- [x] Bugfix

## Description of change
The user id module currently [treats "empty" stored hashed consent data as matching the current consent data](https://github.com/prebid/Prebid.js/blob/master/modules/userId/index.js#L316) available in the auction, thus [not forcing](https://github.com/prebid/Prebid.js/blob/master/modules/userId/index.js#L652) a new call to `getId()` on its own. The intention is that if consent changes from one request to the next, that a `getId()` call will be made to the user id modules can update their ID (or remove it) based on the latest consent data.

After implementing https://github.com/prebid/Prebid.js/pull/6551, however, the hashed consent data was only stored AFTER gdpr checks and consent for local storage given. This means that if the user at first did not give consent (perhaps the CMP timed out and the auction proceeded with empty consent data), then nothing would be stored in the hashed consent data but the ID module could have been called and set a "no-consent" ID. A refresh would not be required on subsequent calls because the ID had already been set and the stored consent data was empty, thus not forcing a refresh.

This PR changes the behavior and treats empty stored consent data as a reason to force a `getId()` call to the ID module.

Note that even if there is no consent data (i.e. a user from outside of the EU), the stored consent data still stores a value due to the [way the hashed value is created](https://github.com/prebid/Prebid.js/blob/3f02a15968d16d1aa9cddc48ed73098cc5677006/modules/userId/index.js#L268). This means that on the first call where cookies are allowed to be written and the user id module executes, a value will be stored based on the current consent data. If consent is not given, then the user id module essentially doesn't execute the submodules, so this doesn't force `getId()` calls on every page view.

@idettman I don't know if this would help solve https://github.com/prebid/Prebid.js/issues/6738 at all, but would love you to take a look either way.

### DIFF
--- a/modules/userId/index.js
+++ b/modules/userId/index.js
@@ -277,6 +277,7 @@ function makeStoredConsentDataHash(consentData) {
     storedConsentData.gdprApplies = consentData.gdprApplies;
     storedConsentData.apiVersion = consentData.apiVersion;
   }
+
   return utils.cyrb53Hash(JSON.stringify(storedConsentData));
 }
 
@@ -306,17 +307,17 @@ function getStoredConsentData() {
 }
 
 /**
- * test if the consent object stored locally matches the current consent data.
- * if there is nothing in storage, return true and we'll do an actual comparison next time.
- * this way, we don't force a refresh for every user when this code rolls out
+ * test if the consent object stored locally matches the current consent data. if they
+ * don't match or there is nothing stored locally, it means a refresh of the user id
+ * submodule is needed
  * @param storedConsentData
  * @param consentData
  * @returns {boolean}
  */
 function storedConsentDataMatchesConsentData(storedConsentData, consentData) {
   return (
-    typeof storedConsentData === 'undefined' ||
-    storedConsentData === null ||
+    typeof storedConsentData !== 'undefined' &&
+    storedConsentData !== null &&
     storedConsentData === makeStoredConsentDataHash(consentData)
   );
 }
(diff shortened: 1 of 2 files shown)


Fawke (commented):
@smenzer Just wanted to understand something, correct me if am wrong. If the user does not give GDPR consent, then we won't write anything to this cookie `_pbjs_userid_consent_data` and that means `storedConsentDataMatchesConsentData` will always return `false`.

In that case won't we be calling `getId()` function on each page refresh even though the user hasn't given any consent?

smenzer:
> In that case won't we be calling `getId()` function on each page refresh even though the user hasn't given any consent?

it's possible i'm mis-reading the code a bit...so please correct me if you see it differently.

the module would [exit early](https://github.com/prebid/Prebid.js/pull/6760/files#R600) if there's no consent for purpose 1 and gdpr enforcement is enabled, which means that we shouldn't even run the submodules without consent.

if gdpr enforcement is not enabled (but consentmanagement module is, meaning it won't prevent modules from running but will provide a consent string), then _i think_ it would get passed the line above to exit and would then store the consent hash in the cookie.

so...i don't think we would be in a situation where the `getId()` method would be called on every single page view.  but again, i could be reading the code incorrectly...

Fawke:
@smenzer Yes, you are right...it won't execute the submodules if no consent. Silly of me to overlook that.

Fawke (approved):

patmmccann (approved):

PR #6771, opened 2021-05-17
TrustX Bid Adapter: added additional sync url

PWyrembak:
## Type of change

- [x] Other

## Description of change

Added additional sync url in TrustX Bid Adapter

- paul@trustx.org


modules/trustxBidAdapter.js
```diff
@@ -7,6 +7,7 @@ import { config } from '../src/config.js';
 const BIDDER_CODE = 'trustx';
 const ENDPOINT_URL = 'https://sofia.trustx.org/hb';
 const NEW_ENDPOINT_URL = 'https://grid.bidswitch.net/hbjson?sp=trustx';
+const ADDITIONAL_SYNC_URL = 'https://x.bidswitch.net/sync?ssp=themediagrid';
```
patmmccann, line 10: should this be ssp=trustx?

modules/trustxBidAdapter.js
```diff
@@ -7,6 +7,7 @@ import { config } from '../src/config.js';
 const BIDDER_CODE = 'trustx';
 const ENDPOINT_URL = 'https://sofia.trustx.org/hb';
 const NEW_ENDPOINT_URL = 'https://grid.bidswitch.net/hbjson?sp=trustx';
```
patmmccann, line 9: should this be ssp=?

modules/trustxBidAdapter.js
```diff
@@ -93,11 +94,14 @@ export const spec = {
...
   },
-  getUserSyncs: function(syncOptions) {
+  getUserSyncs: function(syncOptions, serverResponses) {
     if (syncOptions.pixelEnabled) {
       return [{
         type: 'image',
         url: ADAPTER_SYNC_URL
+      }, {
+        type: 'image',
+        url: ADDITIONAL_SYNC_URL
```
patmmccann, line 104: Can you make this configurable? Perhaps some publishers might not want the extra sync

patmmccann (changes requested):

modules/trustxBidAdapter.js
TheMediaGrid: hi Patrick,
no, all good with ssp=themediagrid
we're in the process of migrating trustx ssp to mediagrid pipes, so this allows enforcing user matching directly grid in addition to trustx and thus helps smoothening the migration

modules/trustxBidAdapter.js
TheMediaGrid: all good with sp=trustx, this is part of the bidding endpoint that is already up & running for a could of pub integrations, so we're all set

thanks for the review!

modules/trustxBidAdapter.js
TheMediaGrid: Technically we could. Don't publishers just restrict the amount of sync urls per adapter that they invoke?

patmmccann: they do, you guys saw this doesn't fire if syncsperbidder = 1 

I thought it just threw a warning in debug mode https://github.com/prebid/Prebid.js/blob/804295aa2dae0484d67891b73fcfc401ef8244f1/src/userSync.js#L188

I don't see the line actually honoring the setting

patmmccann: maybe instead of making it configurable for your adapter, just confirm that getConfig().userSync allows iframe or more than one image

patmmccann (changes requested):

TheMediaGrid:
hi @patmmccann 
we just added support for syncperbidder, thanks for the idea

should kook good now

patmmccann (approved):

PR #6796, opened 2021-05-19
adnuntias Bid Adapter: Added GDPR support and segment passing

mikael-lundin:
## Type of change

- [ ] Bugfix
- [x] Feature
- [ ] New bidder adapter   
- [ ] Build related changes
- [ ] CI related changes
- [ ] Other

## Description of change

Adding GDPR support for the bid adapter and allowing customers to pass segments through the bidder config. If a user has accepted our GLV ID we will use the segments passed on to the adserver. If not, they will not be matched against targeted line items by the adserver.

- contact email of the adapter’s maintainer: mikael@adnuntius.com

https://github.com/prebid/prebid.github.io/pull/2975

### DIFF
--- a/modules/adnuntiusBidAdapter.js
+++ b/modules/adnuntiusBidAdapter.js
@@ -7,12 +7,19 @@ const BIDDER_CODE = 'adnuntius';
 const ENDPOINT_URL = 'https://delivery.adnuntius.com/i?tzo=';
 const GVLID = 855;
 
+const checkSegment = function (segment) {
+  if (utils.isStr(segment)) return segment;
+  if (segment.id) return segment.id
+}
+
 const getSegmentsFromOrtb = function (ortb2) {
   const userData = utils.deepAccess(ortb2, 'user.data');
-  let segments;
+  let segments = [];
   if (userData) {
     userData.forEach(userdat => {
-      segments = (userdat.segment) ? userdat.segment.map(segment => segment.id) : undefined
+      if (userdat.segment) {
+        segments.push(...userdat.segment.filter(checkSegment).map(checkSegment));
+      }
     });
   }
   return segments
--- a/modules/adnuntiusBidAdapter.js
+++ b/modules/adnuntiusBidAdapter.js
@@ -43,7 +43,7 @@ export const spec = {
     const gdprApplies = utils.deepAccess(bidderRequest, 'gdprConsent.gdprApplies');
     const consentString = utils.deepAccess(bidderRequest, 'gdprConsent.consentString');
     const reqConsent = (gdprApplies !== undefined) ? '&consentString=' + consentString : '';
-    const reqSegments = (segments !== undefined && utils.isArray(segments) && segments.length > 0) ? '&segments=' + segments.join(',') : '';
+    const reqSegments = (segments.length > 0) ? '&segments=' + segments.join(',') : '';
 
     for (var i = 0; i < validBidRequests.length; i++) {
       const bid = validBidRequests[i]
--- a/modules/adnuntiusBidAdapter.js
+++ b/modules/adnuntiusBidAdapter.js
@@ -1,30 +1,42 @@
 import { registerBidder } from '../src/adapters/bidderFactory.js';
+import { BANNER } from '../src/mediaTypes.js';
+import * as utils from '../src/utils.js';
+import { config } from '../src/config.js';
 
 const BIDDER_CODE = 'adnuntius';
 const ENDPOINT_URL = 'https://delivery.adnuntius.com/i?tzo=';
+const GVLID = 855;
 
 export const spec = {
   code: BIDDER_CODE,
-
+  gvlid: GVLID,
+  supportedMediaTypes: [BANNER],
   isBidRequestValid: function (bid) {
     return !!(bid.bidId || (bid.params.member && bid.params.invCode));
   },
 
-  buildRequests: function (validBidRequests) {
+  buildRequests: function (validBidRequests, bidderRequest) {
     const networks = {};
     const bidRequests = {};
     const requests = [];
+    const segments = config.getConfig('segments');
+    utils.logMessage('CONF', segments)
     const tzo = new Date().getTimezoneOffset();
-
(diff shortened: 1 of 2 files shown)


mikael-lundin:
Tests are now added. :1st_place_medal: <- I give this to myself :smiley:

modules/adnuntiusBidAdapter.js
```diff
@@ -1,38 +1,50 @@
...
     return !!(bid.bidId || (bid.params.member && bid.params.invCode));
   },
 
-  buildRequests: function (validBidRequests) {
+  buildRequests: function (validBidRequests, bidderRequest) {
     const networks = {};
     const bidRequests = {};
     const requests = [];
+    const segments = config.getConfig('segments');
+    utils.logMessage('CONF', segments)
```
mmoschovas, line 23: I think this line should be removed entirely, as this does not seem like a necessary log to me. I believe that we would rather not inflate code with messaging that is not pertinent as this is neither a user warning/error or really clear in label as to what is being logged - also this can be achieved in the console via a pbjs.getConfig('segments').

modules/adnuntiusBidAdapter.js
```diff
@@ -1,38 +1,50 @@
...
   isBidRequestValid: function (bid) {
     return !!(bid.bidId || (bid.params.member && bid.params.invCode));
   },
 
-  buildRequests: function (validBidRequests) {
+  buildRequests: function (validBidRequests, bidderRequest) {
     const networks = {};
     const bidRequests = {};
     const requests = [];
+    const segments = config.getConfig('segments');
```
mmoschovas, line 22: This is not so much of a change request, but more curious about this config setting. I am not aware of segments being set under the config at the top level, so I am curious if this is something specific to your adapter and if so, should it be in your docs? I also want to make sure that this is not meant to be first party segment data that is defined to be passed via ortb2.user.data and ortb2.site.content.data respectively

mmoschovas (commented):
@mikael-lundin a few comments here

modules/adnuntiusBidAdapter.js
mikael-lundin: yeaj, that was just a blunder from me, will remove that.

modules/adnuntiusBidAdapter.js
mikael-lundin: I was strugling to find where to put it, and yes it's just meant to be for our adapter only, but if ortb2.user.data is the place to put it I will go for it. Should I also prefix our segments  (like `adn-segments`) or are there any other rules for how to structure it?

mmoschovas: I can offer suggestions, but Im not sure where you mean to put the prefix. Maybe reviewing this will offer a better understanding of first party segment data: https://docs.prebid.org/features/firstPartyData.html. You can see under ortb2.user.data, this is an array of objects that "should" include at the very least name and segment (with segment as an array of objects), as well as ext property that can contain other data. Let me know if this helps in any way

mikael-lundin: This should now be solved and the documentation is filed under this pull request:
https://github.com/prebid/prebid.github.io/pull/2975

modules/adnuntiusBidAdapter.js
```diff
@@ -1,38 +1,62 @@
...
 const BIDDER_CODE = 'adnuntius';
 const ENDPOINT_URL = 'https://delivery.adnuntius.com/i?tzo=';
+const GVLID = 855;
+
+const getSegmentsFromOrtb = function (ortb2) {
+  const userData = utils.deepAccess(ortb2, 'user.data');
+  let segments;
+  if (userData) {
+    userData.forEach(userdat => {
+      segments = (userdat.segment) ? userdat.segment.map(segment => segment.id) : undefined
```
mmoschovas, line 15: @mikael-lundin thanks for updating. Do you only want to pass those segments defined with the name 'adnuntius'? Currently there is no filter setup for this here, meaning that if any segment data defined after this type will overwrite the data I believe you are intended to leverage. i.e. if the below is defined in the config, you will only receive the segment 3 and not 1 and 2.

```
ortb2: {
    user: {							
        data: [{								
            name: "adnuntius",								
            segment: [	{ id: "1" },	{ id: "2" }	]							
        },{								
            name: "other",								
            segment: [	{ id: "3" }]							
        }]						
    }					
}
```

I would suggest either setting a filter for the name so you only match on those segments like:

```
userData.filter(data => data.name === 'adnuntius').forEach(userdat => {...});
```

Or if you want to collect all segment data, I would suggest something along the lines of this seeing as you check for segment length and then you can also remove the first two parts of the condition since segments will always be an array:

```
let segments = [];
  if (userData) {
    userData.forEach(userdat => {
      if (userdat.segment) segments.push(...userdat.segment.map(segment => segment.id));
    });
  }
```

A bunch of ways you can do this, but I wanted to make sure that this is corrected so you receive the data you are expecting. Also, I may suggest adding a filter prior to the segment map function to ensure the index of the array contains a property called id - the reason being is that this data structure is fairly new and if a pub were to create it using an array like ['1234','5678'] instead of [{id: '1234'}, {id:'5678'}] you will receive empty comma separated values. Could be something as simple as the following I think would work:

```
userdat.segment.filter(seg => seg.id).map(...);
```

mmoschovas (changes requested):

modules/adnuntiusBidAdapter.js
mikael-lundin: Thanks for that, you kind of saved my ass a bit! :D I've added some changes to it so that it works more as you say, and for now we will just pass all the segments that we see to the adserver.

mmoschovas:
@mikael-lundin can you fix this LGTM alert? Since the getSegmentsFromOrtb always returns an array there is no need for those two checks - as the array will always exist and be of type array. I think you should be fine simply checking the array length in your condition, which I think should remove this LGTM alert

mikael-lundin:
I think I've done what you asked for :D

mmoschovas (approved):
Thanks for making these updates! LGTM

Issue #6754, opened 2021-05-12
Issue with implementing support for floor price module


AskRupert-DM:
## Type of issue
Bug

## Description
Hi 

i've been looking at supporting the floor price module in my adapter (ozoneBidAdapter) per https://github.com/prebid/Prebid.js/issues/6465. 

I've followed the examples listed here https://docs.prebid.org/dev-docs/modules/floors.html and when I check my adapter payload it doesn't appear to be working as expect.

(1) Defining floor prices at the ad unit level doesn't appear to be working - I just don't get the floor price object. Is there something I am missing/have defined wrong on my page config ? I am fairly certain I've followed the config as specified above so either that's missing or I've overlooked something very simple :) 

(2) Defining floor price at the config level appears to work - however, I've defined different floors for banner and video and it looks like video uses/gets the same floor price as banner ? Not sure if I am doing something stupid or missing something in the config but didn't want to submit this in a PR until I was confident its working in the intended fashion

## Steps to reproduce

Please see test pages - on each of them in chrome developer tools filter for 'auction' and you'll see the payload sent for each  - look for bid[n].floor object - where you should be able to reproduce/see the issue listed above.

If you have debug enabled you should be able to filter in the console for - getFloor which also shows what the bidAdapter is receiving.

## Test page

This one will not get any floor information, config is only in adUnits : 
https://www.ardm.io/ozone/2.5.1/sizeconfig-testpage-20210506-floor-adunits-only.html?pbjs_debug=true

This one will get floor information, config is in setConfig:
https://www.ardm.io/ozone/2.5.1/sizeconfig-testpage-20210506-floor-setconfig-only.html?pbjs_debug=true

### Expected results

a Floor object should be passed inside each individual adUnit when defining the config at ad Unit level with each media Type having its own floor.

### Actual results

AdUnit level - no floor information is available.

Config level - floor information is available but the mediaTypes appear to get ignored / banner mediaType settings get applied to video when they have different floor prices defined on the config (0.8 for banner and 1.2 for video - but yet both get 0.8).

## Platform details
All

bszekely1:
This is really a documentation issue. To enable the floors module, setConfig requires at minimum a floors object. Even if it is an empty object like so:

`pbjs.setConfig({
   floors: {}
});`

AskRupert-DM:
Thanks @bszekely1  - will update my testpage tomorrow & try that out tomorrow and see if that solves my problem with adUnit floors not working. 

[...]

PR #6804, opened 2021-05-20
Rule enforcement: support for floors must include the floors module

patmmccann:
## Type of change
- [X] Other

## Description of change
Enforcement of https://github.com/prebid/Prebid.js/issues/6465 for https://github.com/prebid/Prebid.js/issues/5966 release

I have tried to remove floor support for recently maintained adapters when possible, otherwise adapters were removed.


patmmccann:
Ozone and EMX not included because of in progress efforts noted at https://github.com/prebid/Prebid.js/issues/6754 by @askrupert-dm and https://github.com/prebid/Prebid.js/pull/6797 by @EMXDigital

patmmccann:
anticipating conflicts on https://github.com/prebid/Prebid.js/pull/6820/files ; will remove from those changes from this pr

patmmccann:
sspbc changes removed, ready for review

ChrisHuie (approved):

robertrmartinez:
What happened to this file

`test/spec/modules/smilewantedBidAdapter_spec.js`


Looks like it was renamed to

`test/spec/smilewantedBidAdapter_spec.js`

Which is not correct I believe.

patmmccann:
> What happened to this file
> 
> `test/spec/modules/smilewantedBidAdapter_spec.js`
> 
> Looks like it was renamed to
> 
> `test/spec/smilewantedBidAdapter_spec.js`
> 
> Which is not correct I believe.
> 
> Very last change of files when I look at it:
> 
> image

fixed ty

robertrmartinez:
I counted
- 29 modules removed
- 26 Spec files removed

I may have missed something but should this be equal number?

patmmccann:
Not every module had a spec file

robertrmartinez:
Also I just realized I double counted some because the `.md` file!

robertrmartinez:
Looks good.

The changes where we removed the ability to pass a floor, but just set it to default is what we want for sure right?

Seems like the right thing imo.

Thanks this was a lot of work I am sure!

patmmccann:
We're hoping those adapters, eg spotx, add support shortly. The only other option I can think of is removing them, so I tried for any adapter that seemed recently maintained to set the default. The deletions are almost all not maintained.

PR #6814, opened 2021-05-22
Smaato bid adapter: Rework multi imp support

jgresens:
## Type of change
- [x] Feature

## Description of change
Smaatos Ad Server doesn't fully supports multiple impressions per request, creating an individual request per valid bid request is an alternative solution

- contact email of the adapter’s maintainer: prebid@smaato.com

### DIFF
--- a/modules/smaatoBidAdapter.js
+++ b/modules/smaatoBidAdapter.js
@@ -5,75 +5,22 @@ import { BANNER, VIDEO } from '../src/mediaTypes.js';
 
 const BIDDER_CODE = 'smaato';
 const SMAATO_ENDPOINT = 'https://prebid.ad.smaato.net/oapi/prebid';
-const CLIENT = 'prebid_js_$prebid.version$_1.1'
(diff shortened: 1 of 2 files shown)


patmmccann:
If you're doing this,  it's really important to have an http2 endpoint. Can you confirm yours is?

jgresens:
> If you're doing this, it's really important to have an http2 endpoint. Can you confirm yours is?

Hey Patrick,

thanks for pointing that out, no the endpoint doesn't use http2 yet, but the migration to it is in process.

I guess you are asking because the individual requests that this change would create quite some overhead on both sides against a non-http2 endpoint due to the missing connection multiplexing?

patmmccann:
Precisely

osazos (approved):
LGTM
Hello @jgresens, all seems ok for me concerning the PR, thx.

osazos:
@jgresens can you rebase your branch on master please? then I can be sure all is ok about the CI

jgresens:
@osazos should be good now

PR #6822, opened 2021-05-24
PBJS Core: Add allowSendAllBidsTargetingKeys to targetingControls

r-schweitzer:
## Type of change
Feature

## Description of change
Allows a publisher to limit the number 'sendAllBids' keys sent to the adServer. For example hb_bidder_rubicon is one of the more redudant keys we would like to stop sending but we still need hb_bidder key sent. This would similar to the targetingControl option, allowTargetingKeys.

Sample Config:
Example will not send hb_bidder_rubicon
```  enableSendAllBids: true,
    targetingControls: {
        allowTargetingKeys: ['BIDDER', 'AD_ID', 'PRICE_BUCKET'],
        allowSendAllBidsTargetingKeys: ['PRICE_BUCKET', 'AD_ID']
    }
```

- Docs PR -- https://github.com/prebid/prebid.github.io/pull/2985

### DIFF
--- a/src/targeting.js
+++ b/src/targeting.js
@@ -597,13 +597,19 @@ export function newTargeting(auctionManager) {
     const standardKeys = TARGETING_KEYS.concat(NATIVE_TARGETING_KEYS);
     const adUnitBidLimit = config.getConfig('sendBidsControl.bidLimit');
     const bids = getHighestCpmBidsFromBidPool(bidsReceived, getHighestCpm, adUnitBidLimit);
+    const allowSendAllBidsTargetingKeys = config.getConfig('targetingControls.allowSendAllBidsTargetingKeys');
+
+    const allowedSendAllBidTargeting = allowSendAllBidsTargetingKeys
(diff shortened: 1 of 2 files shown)


src/targeting.js
```diff
@@ -597,13 +597,19 @@ export function newTargeting(auctionManager) {
...
+      : standardKeys;
 
     // populate targeting keys for the remaining bids
     return bids.map(bid => {
       if (bidShouldBeAddedToTargeting(bid, adUnitCodes)) {
         return {
           [bid.adUnitCode]: getTargetingMap(bid, standardKeys.filter(
-            key => typeof bid.adserverTargeting[key] !== 'undefined')
+            key => typeof bid.adserverTargeting[key] !== 'undefined' &&
+            allowedSendAllBidTargeting.includes(key))
```
patmmccann, line 612: you can't use includes in ie11

patmmccann: import includes from 'core-js-pure/features/array/includes.js';

is the polyfill syntax

patmmccann (changes requested):

patmmccann (approved):

ChrisHuie (approved):

PR #6826, opened 2021-05-25
Sonobi bid adapter: add support for gpid and set advertiserDomains 

JonGoSonobi:
## Type of change

- [X] Feature

## Description of change
Try to send the gpid as the gpt ad unit code in the Sonobi bid request if it is available.

- apex.prebid@sonobi.com


patmmccann:
This looks good but can you also fix https://github.com/prebid/Prebid.js/issues/6650 for your adapter real quick so you don't get removed from prebid next week?

JonGoSonobi:
@patmmccann thanks for the heads up. We must have missed that notification. I added the changes to support the meta.advertiserDomain field.

patmmccann (approved):

PR #6829, opened 2021-05-25
Trion bid adapter: support for meta advertiserDomains

grohm33:
## Type of change

- [ ] Bugfix
- [x ] Feature
- [ ] New bidder adapter   
- [ ] Build related changes
- [ ] CI related changes
- [ ] Other

## Description of change

Adding Trion adapter support for meta advertiserDomains

## Other information

We noticed that in a write up for v4.0 there was talk of bid.meta being required in the future.  If there are other changes needed for our adapter to meet future minimum requirements please call them out here and we will be happy to change them.


patmmccann:
Thank you very much, partial solution to #6650 

Can you "Fetch upstream" 

We've been having some recent test failure issues and this should resolve them

grohm33:
@patmmccann thanks for tagging the issue.  We had not previously seen this until now.  You mentioned this was partially fulfilling the requirement.  Is that because you would like us to pass an empty advertiserDomains when there are none to provide?  I apologize if I'm missing something.

patmmccann:
It's partial solution bc your pr doesn't close the issue,  other adapters need to make changes as well

I'm not aware of any other concerns with your adapter and you're clear for 5.0

grohm33:
Okay @patmmccann. Thank you for clearing that up!

PR #6831, opened 2021-05-26
Adyoulike bid adapter - improvements on Native case

guiann:
## Type of change
Bugfix,
Refactoring and optimisation

## Description of change
Allows Native response to be transmited as is for simplicity.
Assure clickUrl and other trackings (pixels and JS) are correctly handled.

- [* ] official adapter submission

### DIFF
--- a/.github/release-drafter.yml
+++ b/.github/release-drafter.yml
@@ -22,7 +22,7 @@ version-resolver:
   patch:
     labels:
       - 'patch'
-  default: patch
+  default: minor
 template: |
   ## In This Release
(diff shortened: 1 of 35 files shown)


patmmccann (changes requested):
Can you fix #6650 as well? We're planning to enforce it very soon

guiann:
to thanks for pointing this out to me. 
I've added this task to handle meta fields on the adyoulike adapter. 
It will be done in a futur pull request.

ChrisHuie (approved):

PR #6835, opened 2021-05-26
Zeta Ssp Bid Adapter: Improve user sync logic

asurovenko-zeta:
## Type of change
- [ ] Feature

## Description of change
Support iframe and image sync types.
**Image** type return **302** redirect.
**Iframe** type return **200** OK with `<html><body>...</body></html>`.


asurovenko-zeta:
Docs PR: https://github.com/prebid/prebid.github.io/pull/2989 (with other changes)

PR #6846, opened 2021-05-27
AdmixerIsSystem:  pass visitor and stored ids

AdmixerTech:
## Type of change
- [x] Feature

## Description of change
pass data to admixerIdSystem
 * visitorId from admixerBidAdapter response
 * storedId from prebid


patmmccann:
I'm a little confused as to what's happening here,  why don't you just use regular methods to grab the id, same as any other id?

AdmixerTech:
that's a little different services that may have different ids

patmmccann:
> that's a little different services that may have different ids

I don't understand,  please make a better explanation

PR #6860, opened 2021-05-27
Inject creativeComment after render so it actually persists

jsut:
## Type of change
- [x] Bugfix

## Description of change

pbjs.renderAd currently injects an HTML comment into the body tag of the document passed in. This comment seems to get eaten up in any case where an ad is actually rendered successfully.  Some bid adapters insert a comment similar to this in their renderers, but some don't.  This change moves the actual injection of the comment to the HTML tag of the document, as well as changing it to run after the ad has been rendered.  In my testing this actually results in a comment that you can see in the DOM for the adapters I happened to get demand from during manual testing.

I suppose there could/should be some tests for this, but i haven't written any since there aren't any renderAd tests that care about this comment.  I'd be happy to look into getting some working if desired.

### DIFF
--- a/src/prebid.js
+++ b/src/prebid.js
@@ -449,8 +449,8 @@ $$PREBID_GLOBAL$$.renderAd = hook('async', function (doc, id, options) {
 
           if (isRendererRequired(renderer)) {
             executeRenderer(renderer, bid);
-            emitAdRenderSucceeded({ doc, bid, id });
             utils.insertElement(creativeComment, doc, 'html');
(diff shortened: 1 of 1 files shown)


jsnellbaker (approved):

mmoschovas (approved):
LGTM

mmoschovas:
@jsut can these conflicts be resolved so this can be merged?

jsut:
@mmoschovas rebased on master

PR #6741, opened 2021-05-11
ParrableIdSystem: Store third-party cookie support status in first-party cookie

icflournoy:
## Type of change
- [x] Other

## Description of change

(First-attempt #6689)

The Parrable backend tests a browser's support of third-party cookies on each request. If the test cookie is stored successfully (third party cookies are supported in the browser) then the cookie may remain up to 24 hours in the browser and be sent along with subsequent requests to our backend. We will not attempt to re-test third-party cookie status if the third-party cookie is received.

This change introduces a small value, to be stored in our compound/composite first-party cookie that retains the state of the browser's third-party cookie support from the last time it was tested. This is beneficial in browsers that don't support third-party cookies where we would otherwise always re-test their third-party cookie support on each request made.

The solution is implemented by storing two additional values in the first-party cookie maintained under prebid. The first is the actual state of third-party cookie support in the browser. This value is returned from our backend systems after the third-party cookie test is complete. The second value is a future unix timestamp (in seconds) that determines when the stored third-party cookie state is no longer valid, and will not be supplied to our backend, thus causing a re-test of the browser's third-party cookie support.

This change is internally facing and only affects our cookie storage layout, and the request/response exchange with our backend systems.

## Other information

* Author @victorigualada

PR #6868, opened 2021-05-28
Increase cookie expire time to 1 day

victorigualada:
## Type of change

- [x] Bugfix

## Description of change
Previous PR caused some test to fail occasionally due to a low cookie expire time. Now the expire time is increased to one day to avoid those failures.
This is the PR that introduced the problem: https://github.com/prebid/Prebid.js/pull/6741

Other information
Author @victorigualada


ChrisHuie:
Couldn't get tests to kick off in this pr and wanted to test that this fixed the problem :)

PR #6828, opened 2021-05-25
Dependencies/Testing: update dependencies, npm audit fixes, and bump Browserstack versions

ChrisHuie:
This is a stable update to the dependencies that don't require file change rewrites, fixes vulnerabilities from outdated packages, and updates Safari, Chrome, Edge, and Firefox browserstack to most recent versions along with oldest supported versions.

PR #6911, opened 2021-06-01
Replace uglify with terser to support ES6 output

nanek:
## Type of change
- [X] Build related changes

## Description of change

Changing the babel [presets-env browser target](https://github.com/spanishdict/Prebid.js/blob/master/.babelrc.js#L18) to just Chrome in the master branch will fail with an error.

````
[19:03:38] 'makeWebpackPkg' errored after 8.06 s
[19:03:38] GulpUglifyError: unable to minify JavaScript
Caused by: SyntaxError: Unexpected token: keyword «const»
File: /Users/kenan/scm/sd/Prebid.js/criteoBidAdapter.js
Line: 38
[19:03:38] 'build-bundle-prod' errored after 8.07 s
[19:03:38] 'build' errored after 8.13 s
````

uglify-js does not support ES6+. By switching to terser, builds of prebid can target es6 if desired.

## Other information

https://github.com/terser/terser#why-choose-terser

### DIFF
--- a/.babelrc.js
+++ b/.babelrc.js
@@ -29,6 +29,7 @@ module.exports = {
   ],
   "plugins": [
     path.resolve(__dirname, './plugins/pbjsGlobals.js'),
-    useLocal('babel-plugin-transform-object-assign')
+    useLocal('babel-plugin-transform-object-assign'),
+    useLocal('@babel/plugin-proposal-object-rest-spread')
   ]
 };
--- a/gulpfile.js
+++ b/gulpfile.js
@@ -8,7 +8,7 @@ var gutil = require('gulp-util');
 var connect = require('gulp-connect');
 var webpack = require('webpack');
 var webpackStream = require('webpack-stream');
-var uglify = require('gulp-uglify');
+var terser = require('gulp-terser');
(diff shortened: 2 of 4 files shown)

### CODE REVIEW

File: .babelrc.js [hunk 1]
```diff
@@ -29,6 +29,7 @@ module.exports = {
   ],
   "plugins": [
     path.resolve(__dirname, './plugins/pbjsGlobals.js'),
-    useLocal('babel-plugin-transform-object-assign')
+    useLocal('babel-plugin-transform-object-assign'),
+    useLocal('@babel/plugin-proposal-object-rest-spread')
```
nanek, line 33: Not related to using terser, however this fixes other build errors thrown by webpackStream if you target just Chrome. `@babel/plugin-proposal-object-rest-spread` is part of `babel/preset-env`.

ChrisHuie:
@nanek looks like our tests didn't run on this pr but I can test it out in an updated dependency pr #6828 that is about to geet merged. We are looking at upgrading webpack but it is taking a bit of effort to convert to a chunking plugin that is now built into webpack

ChrisHuie:
closing as these changes have been added to #6828

PR #6928, opened 2021-06-03
Zeta Ssp Bid Adapter: provide domain in site object

asurovenko-zeta:
## Type of change
- [ ] Feature

## Description of change
Provide domain value in site object


ChrisHuie:
@asurovenko-zeta can you please update this pr to work with ie11 so it passes testing 🙏

asurovenko-zeta:
@ChrisHuie sorry, do not noticed that, now should be ok. Thanks!

ChrisHuie (approved):

PR #6932, opened 2021-06-03
Quantcast Bid Adapter: uses video mediatypes to read OpenRTB parameters

SleimanJneidi:
## Type of change

- [ ] Bugfix
- [ ] Feature
- [ ] New bidder adapter   
- [ ] Build related changes
- [ ] CI related changes
- [x] Other

## Description of change
Reads video properties from mediaTypes https://github.com/prebid/Prebid.js/issues/6512

### DIFF
--- a/modules/quantcastBidAdapter.js
+++ b/modules/quantcastBidAdapter.js
@@ -11,6 +11,7 @@ const DEFAULT_BID_FLOOR = 0.0000000001;
 const QUANTCAST_VENDOR_ID = '11';
 // Check other required purposes on server
 const PURPOSE_DATA_COLLECT = '1';
+const VIDEO_PROPS_TO_REMOVE = ['context', 'playerSize'];
 
 export const QUANTCAST_DOMAIN = 'qcx.quantserve.com';
 export const QUANTCAST_TEST_DOMAIN = 's2s-canary.quantserve.com';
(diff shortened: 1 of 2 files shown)


patmmccann:
do you want to fallback to your own parameters so you don't break any existing integrations?

modules/quantcastBidAdapter.js
```diff
@@ -25,29 +25,61 @@ export const storage = getStorageManager(QUANTCAST_VENDOR_ID, BIDDER_CODE);
...
-    video['protocols'] = bid.params.video.protocols;
-    video['startdelay'] = bid.params.video.startdelay;
-    video['linearity'] = bid.params.video.linearity;
-    video['battr'] = bid.params.video.battr;
-    video['maxbitrate'] = bid.params.video.maxbitrate;
-    video['playbackmethod'] = bid.params.video.playbackmethod;
-    video['delivery'] = bid.params.video.delivery;
-    video['placement'] = bid.params.video.placement;
-    video['api'] = bid.params.video.api;
-  }
```
patmmccann, line 41: why not keep this block so either place populates it? I think this is why your tests are failing

patmmccann (changes requested):

test/spec/modules/quantcastBidAdapter_spec.js
```diff
@@ -221,7 +225,11 @@ describe('Quantcast adapter', function () {
 
     it('overrides video parameters with parameters from adunit', function() {
       setupVideoBidRequest({
```
patmmccann, line 227: this test looks good as is, it has two different values for mimes and you're testing the ad unit wins.

patmmccann (changes requested):

modules/quantcastBidAdapter.js
```diff
@@ -25,29 +25,61 @@ export const storage = getStorageManager(QUANTCAST_VENDOR_ID, BIDDER_CODE);
...
+
+    if (bid.mediaTypes.video.mimes) {
+      video['mimes'] = bid.mediaTypes.video.mimes;
+    }
+    if (utils.isArray(bid.mediaTypes.video.playerSize[0])) {
+      video['w'] = bid.mediaTypes.video.playerSize[0][0];
+      video['h'] = bid.mediaTypes.video.playerSize[0][1];
+    } else {
+      video['w'] = bid.mediaTypes.video.playerSize[0];
+      video['h'] = bid.mediaTypes.video.playerSize[1];
```
patmmccann, line 81: you may want to check for h and w existing directly as well

modules/quantcastBidAdapter.js
```diff
@@ -25,30 +25,89 @@ export const storage = getStorageManager(QUANTCAST_VENDOR_ID, BIDDER_CODE);
 
 function makeVideoImp(bid) {
   const video = {};
+
   if (bid.params.video) {
```
dpapworth-qc, line 29: should mimes also be retrieved from the mediaTypes object?

modules/quantcastBidAdapter.js
```diff
@@ -25,30 +25,89 @@ export const storage = getStorageManager(QUANTCAST_VENDOR_ID, BIDDER_CODE);
...
     video['mimes'] = bid.params.video.mimes;
+  }
+
+  if (bid.mediaTypes && bid.mediaTypes.video && bid.mediaTypes.video.minduration) {
```
dpapworth-qc, line 33: you might want to consider using utils.deepAccess for nested parameters.

modules/quantcastBidAdapter.js
```diff
@@ -25,30 +25,89 @@ export const storage = getStorageManager(QUANTCAST_VENDOR_ID, BIDDER_CODE);
...
+  }
+
+  if (bid.mediaTypes && bid.mediaTypes.video && bid.mediaTypes.video.api) {
+    video['api'] = bid.mediaTypes.video.api;
+  } else if (bid.params.video) {
     video['api'] = bid.params.video.api;
   }
-  if (bid.mediaTypes.video.mimes) {
+
+  if (bid.mediaTypes && bid.mediaTypes.video && bid.mediaTypes.video.mimes) {
```
dpapworth-qc, line 99: this should be grouped with the bid.param.video.mimes above

modules/quantcastBidAdapter.js
```diff
@@ -24,31 +25,20 @@ export const QUANTCAST_FPA = '__qca';
...
+    video.w = video.playerSize[0];
+    video.h = video.playerSize[1];
   }
-  if (utils.isArray(bid.mediaTypes.video.playerSize[0])) {
-    video['w'] = bid.mediaTypes.video.playerSize[0][0];
-    video['h'] = bid.mediaTypes.video.playerSize[0][1];
-  } else {
-    video['w'] = bid.mediaTypes.video.playerSize[0];
-    video['h'] = bid.mediaTypes.video.playerSize[1];
+  for (const prop of VIDEO_PROPS_TO_REMOVE) {
```
dpapworth-qc, line 36: a bit of a nit-pick - would it make more sense to have an allow list of properties rather than a block list? otherwise the video object on mediaTypes or params could introduce new properties and we wouldn't filter them.

dpapworth-qc (commented):
a bit of a nit-pick, but looks good otherwise.

modules/quantcastBidAdapter.js
SleimanJneidi: done

dpapworth-qc (approved):
lgtm - but needs approval from Prebid. :)

patmmccann (approved):

PR #6933, opened 2021-06-03
PBS Bid Adapter: Fix duplicate imp.id

### LINKED ISSUE #6910: Prebid Server BidAdapter doesn't support 'twinned' ad units
## Type of issue
enhancement

## Description
As described in [Prebid Server issue 1866](https://github.com/prebid/prebid-server/issues/1866), PBS rejects AdUnits with the same code because that translates to OpenRTB imps with the same ID, which is forbidden by OpenRTB.

## Steps to reproduce
Set up twin adunits as described in https://docs.prebid.org/dev-docs/adunit-reference.html#twin-adunit-codes and push the bidders server-side.

## Proposal

Currently the pbsBidAdapter places adunit.code in imp[].id, and relies on seatbid.bid.impid to map back to adunit code to put the bids against the correct adunit.

Whatever solution we choose, pbsBidAdapter will have to be expanded to keep a map of adunit codes-vs-Imp.id.

- Option 1 - update PBJS pbsBidAdapter to just de-dup imp[].id when there are multiple values. e.g. test-div, **test-div-2**. It also needs to be able to map bid.impid back to the right twinned adunit.
- Option 2 - update PBJS pbsBidAdapter to use something other than adunit code as imp.id: **use a UUID** and place the adunit code in imp.ext.data.pbadunitcode for downstream reporting and troubleshooting. It would need to map UUIDs back to adunits.
- Option 3 - have pbsBidAdapter create imp.id as **sequential** (as defined in OpenRTB) and place the adunit code in imp.ext.data.pbadunitcode for downstream reporting and troubleshooting. It needs to be able to map these IDs back to adunits.

The Prebid Server team discussed this and proposes Option 1 because it's the smallest OpenRTB footprint and least overall change.

FilipStamenkovic:
## Type of change

- [x] Bugfix

## Description of change

This PR suggests deduplication of `imp.id` by adding suffix `-2` to the `imp.id`, if duplication is detected.  
Change `bidIdMap` so that the keys are in format `imp.id-bidder` instead of `adUnit.code-bidder`.
## Other information

Fixes #6910

### DIFF
--- a/modules/prebidServerBidAdapter/index.js
+++ b/modules/prebidServerBidAdapter/index.js
@@ -501,7 +501,7 @@ const OPEN_RTB_PROTOCOL = {
       let i = 1;
       while (impIds.has(impressionId)) {
         i++;
-        impressionId = `${impressionId}-${i}`;
+        impressionId = `${adUnit.code}-${i}`;
       }
       impIds.add(impressionId);
 
--- a/modules/prebidServerBidAdapter/index.js
+++ b/modules/prebidServerBidAdapter/index.js
@@ -492,6 +492,7 @@ const OPEN_RTB_PROTOCOL = {
     const firstBidRequest = bidRequests[0];
 
     // transform ad unit into array of OpenRTB impression objects
+    let impIds = new Set();
     adUnits.forEach(adUnit => {
       const nativeParams = processNativeAdUnitParams(utils.deepAccess(adUnit, 'mediaTypes.native'));
       let nativeAssets;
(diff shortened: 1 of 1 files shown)


hnkhandev (changes requested):
Since this is not limited to just two sets of identical adunits (e.g adding native could make it three sets), I would recommend using a map/obj to keep track of the imp ids. With the current proposed changes, it would label triple adUnits as '-2-3' instead of '-2'.

`
let impIds = [];
...
...
...
let id = adUnit.code;
if (!(id in impIds)) {
   impIds[id] = 1
} else {
   impIds[id]++
   id = `${id}-${impIds[id]}`;
}  
`

FilipStamenkovic:
@hnkhandev Thanks for the review, it should be fixed now. I overlooked how my `while` loop would behave for 'triplets'.

hnkhandev (approved):
Looks good. I can't spot any further issues.

ChrisHuie (approved):

PR #6963, opened 2021-06-07
PubMatic bid adapter: Change user-sync end-point

pm-harshad-mane:
## Type of change
- [ X ] Feature: Change user-sync end-point

## Description of change
PubMatic bid adapter: Change user-sync end-point


patmmccann:
is this going to fix it so you guys stop doing document.write in your syncs?

pm-harshad-mane:
Hello @patmmccann , 
We have already made changes to avoid `document.write`.
PageSpeed was highlighting the old end-point to have a higher `unused` code percentage.
This change will use another end-point that has the only user-sync-specific code so that `unused` code does not go on Publisher pages.
The new end-point code will be updated from our side by end-of-week.

patmmccann (approved):

PR #6973, opened 2021-06-08
RadsBidAdapter: Add adomains/userIds support and change banner ad type priority

onlsol:
## Type of change
- [ x ] Others

## Description of change
 - Added support of  adomain in meta for Prebid v5
 - Added support alternative external userIds (UID 2.0, netID)
 - Changed priority for banner ad type instead video ad type
 - Changed code for parse of sizes

- [ x ] official adapter submission
 
Doc Update here: https://github.com/prebid/prebid.github.io/pull/3020

PR #6983, opened 2021-06-09
Justpremium, dspx, VisX, Yieldlab bid adapters: Add missing GVL IDs

muuki88:
## Type of change

- [x] Bugfix

## Description of change

Add GVL IDs to some bidders that are missing it. The `gdprEnforcement` module otherwise blocks user syncs.

## Other information

- VisX: @mk0x9 
- DSPX: @onlsol 
- Yieldlab: @mirkorean 
- Just Premium: @nwlosinski

### DIFF
--- a/modules/dspxBidAdapter.js
+++ b/modules/dspxBidAdapter.js
@@ -7,9 +7,11 @@ const BIDDER_CODE = 'dspx';
 const ENDPOINT_URL = 'https://buyer.dspx.tv/request/';
 const ENDPOINT_URL_DEV = 'https://dcbuyer.dspx.tv/request/';
 const DEFAULT_VAST_FORMAT = 'vast2';
+const GVLID = 937;
 
 export const spec = {
   code: BIDDER_CODE,
+  gvlid: GVLID,
   aliases: ['dspx'],
   supportedMediaTypes: [BANNER, VIDEO],
   isBidRequestValid: function(bid) {
--- a/modules/dspxBidAdapter.js
+++ b/modules/dspxBidAdapter.js
@@ -7,7 +7,7 @@ const BIDDER_CODE = 'dspx';
 const ENDPOINT_URL = 'https://buyer.dspx.tv/request/';
 const ENDPOINT_URL_DEV = 'https://dcbuyer.dspx.tv/request/';
 const DEFAULT_VAST_FORMAT = 'vast2';
-const GVLID = 937;
+const GVLID = 602;
 
 export const spec = {
   code: BIDDER_CODE,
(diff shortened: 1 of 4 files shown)


modules/visxBidAdapter.js
```diff
@@ -5,6 +5,7 @@ import { BANNER, VIDEO } from '../src/mediaTypes.js';
 import { INSTREAM as VIDEO_INSTREAM } from '../src/video.js';
 const { parseSizesInput, getKeys, logError, deepAccess } = utils;
 const BIDDER_CODE = 'visx';
+const GVLID = 154
```
patmmccann, line 8: Semicolon

modules/justpremiumBidAdapter.js
```diff
@@ -2,12 +2,14 @@ import { registerBidder } from '../src/adapters/bidderFactory.js'
 import { deepAccess } from '../src/utils.js';
 
 const BIDDER_CODE = 'justpremium'
+const GVLID = 62;
```
patmmccann, line 5: This block seems to have no semicolons, probably introducing a bunch of lgtm warnings

muuki88: Ah! I missed one...
```suggestion
const GVLID = 62
```

modules/visxBidAdapter.js
muuki88:
```suggestion
const GVLID = 154;
```

onlsol:
for DSPX the GLVID is: 602

patmmccann:
> for DSPX the GLVID is: 602

https://vendor-list.consensu.org/v2/vendor-list.json

You guys are 'online solution'?

muuki88:
I know that it was before, but ScreenOnDemand now has its own. Should I use the "old" one?

onlsol:
@patmmccann yes.

@muuki88 please set 602 "online solution" for our dspx adapter. it's mandatory for the technical dspx integration and user sync to be considered, in the same way as the gdprEnforcement module's check is intended. 

besides, there are partnerships and integrations for SOD, independent from user sync or tech requirements, and JCA...

modules/dspxBidAdapter.js
```diff
@@ -7,9 +7,11 @@ const BIDDER_CODE = 'dspx';
 const ENDPOINT_URL = 'https://buyer.dspx.tv/request/';
 const ENDPOINT_URL_DEV = 'https://dcbuyer.dspx.tv/request/';
 const DEFAULT_VAST_FORMAT = 'vast2';
+const GVLID = 937;
```
muuki88, line 10:
```suggestion
const GVLID = 602;
```

muuki88:
Done

patmmccann (approved):

onlsol:
thank you. 

fyi, we've also completed the GVL ID in the latest pull request (https://github.com/prebid/Prebid.js/pull/6973/files#diff-94c68527a956d2835a6a8fd243272691e8163f7555af74016bef7ab8965209bbR10) for our RADS bidder adapter besides compatibility updates for pbjs v5.

patmmccann:
out of curiosity, why do you have two bidders?

PR #6996, opened 2021-06-09
amxIdSystem: add new ID submodule

nickjacob:
## Type of change

- [x ] Other

## Description of change

Adding the `amxIdSystem` userId submodule. This provides an `amxId` user ID and eid.

contact email of the adapter’s maintainer: prebid@amxrtb.com
- [ x] official adapter submission

PR in docs repo: https://github.com/prebid/prebid.github.io/pull/3026

### DIFF
--- a/integrationExamples/gpt/userId_example.html
+++ b/integrationExamples/gpt/userId_example.html
@@ -231,7 +231,8 @@
               "name": "amxId",
               "storage": {
                 "type": "html5",
-                "name": "amxId"
+                "name": "amxId",
+                "expires": 14
               }
             },
             {
--- a/integrationExamples/gpt/userId_example.html
+++ b/integrationExamples/gpt/userId_example.html
@@ -76,6 +76,7 @@
                   "95": true,     // lotamePanoramaId
                   "301": true,    // zeotapIdPlus
                   "91": true,     // criteo
+                  "737": true,    // amxId
                 }
               }
             }
@@ -225,6 +226,13 @@
             },
             {
               "name": "criteo"
+            },
+            {
+              "name": "amxId",
+              "storage": {
+                "type": "html5",
+                "name": "amxId"
+              }
             },
             {
               "name": "uid2"
--- a/modules/amxIdSystem.js
+++ b/modules/amxIdSystem.js
@@ -1,0 +1,150 @@
+/**
+ * This module adds AMX to the User ID Module
+ * The {@link module:modules/userId} is required
+ *
+ * @module modules/amxIdSystem
+ * @requires module:modules/userId
+ */
+import { uspDataHandler } from '../src/adapterManager.js';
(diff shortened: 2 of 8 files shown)


smenzer:
@nickjacob can you please also add test to `userId_spec.js` - you can follow the same updates as other modules

nickjacob:
@smenzer thanks for pointing that out -- looks like when 5.0 got merged into master it caused some weird issues in userId_spec.js -- I tried to resolve it in a way that made sense (specifically here -- in master: https://github.com/prebid/Prebid.js/blob/master/test/spec/modules/userId_spec.js#L532)

nickjacob:
thanks @patmmccann  -- added to submodules.json

modules/amxIdSystem.js
```diff
@@ -0,0 +1,150 @@
...
+                return;
+              } catch (e) {
+                logError(`${NAME} invalid response`, responseText);
+              }
+            }
+
+            done(null);
+          },
+        },
+        params
```
smenzer, line 143: when i add run your code on page (I'd recommend adding it to `userId_example.html` under integration examples), i get an error with the response from the server. i believe `params` should be `JSON.stringify(parms)` here as it shows as `[object Object]` in the network call. either that, or the server is not responding properly.

smenzer (changes requested):

nickjacob:
Thanks @smenzer -- sorry about that! I added the integration example & fixed that request.

I also removed the "timeout" option

modules/amxIdSystem.md
```diff
@@ -0,0 +1,51 @@
...
+
+### Storage Settings
+
+The following settings are available for the `storage` property in the `userSync.userIds[]` object:
+
+| Param under `storage` | Scope    | Type         | Description                                                                      | Example   |
+| --------------------- | -------- | ------------ | -------------------------------------------------------------------------------- | --------- |
+| name                  | Required | String       | Where the ID will be stored                                                      | `"amxId"` |
+| type                  | required | String       | This must be `"html5"`                                                           | `"html5"` |
+| expires               | Optional | Number <= 30 | number of days until the stored ID expires. **Must be less than or equal to 30** | `30`      |
```
smenzer, line 43: this is not an optional parameter if you are using the built-in storage

integrationExamples/gpt/userId_example.html
```diff
@@ -226,6 +227,13 @@
             {
               "name": "criteo"
             },
+            {
+              "name": "amxId",
+              "storage": {
+                "type": "html5",
+                "name": "amxId"
```
smenzer, line 234: you must pass in `expires`

smenzer (changes requested):
just need to fix `expires` and then this looks good to me

smenzer (approved):
LGTM

PR #7010, opened 2021-06-11
Ats analytics set sampling rate to 1 percent

mamatic:
## Type of change

- [ ] Bugfix

## Description of change
- Change sampling rate to be 100 always and if sampling rate is 0 do not fire analytics request. If sampling rate is set something other then 0, set it to 100.

## Other information
@smenzer


mkendall07:
@mamatic if you want this to be supported in 4.x please also open a PR against that branch https://github.com/prebid/Prebid.js/tree/4.43.x-legacy

mamatic:
> @mamatic if you want this to be supported in 4.x please also open a PR against that branch https://github.com/prebid/Prebid.js/tree/4.43.x-legacy

OK, thanks for info. Sure, I will do it.

PR #7013, opened 2021-06-11
Limelight Digital Bid Adapter: add new bid adapter

imedvedko:
## Type of change
- [x] New bidder adapter 

## Description of change
This PR restores project limelight bidder adapter and renames it to limelight digital bidder adapter

```
var adUnits = [{
    code: 'placementCode',
    mediaTypes: {
        banner: {
            sizes: [[300, 250]]
        }
    },
    bids: [{
        bidder: 'limelightDigital',
        params: {
            host: 'exchange.ortb.net',
            adUnitId: 0,
            adUnitType: 'banner'
        }
    }]
}];
```
```
var videoAdUnit = [{
    code: 'video1',
    sizes: [[300, 250]],
    bids: [{
        bidder: 'limelightDigital',
        params: {
            host: 'exchange.ortb.net',
            adUnitId: 0,
            adUnitType: 'video'
        }
    }]
}];
```

- contact email of the adapter’s maintainer engineering@project-limelight.com
- Bidder params documentation here https://github.com/prebid/prebid.github.io/pull/3033


MartianTribe:
https://github.com/prebid/prebid.github.io/pull/3033/files

modules/limelightDigitalBidAdapter.js
```diff
@@ -0,0 +1,143 @@
+import { registerBidder } from '../src/adapters/bidderFactory.js';
+import { BANNER, VIDEO } from '../src/mediaTypes.js';
+import {ajax} from '../src/ajax.js';
+import * as utils from '../src/utils.js';
+
+const BIDDER_CODE = 'limelight-digital';
```
jsnellbaker, line 6: The bidder codes and the names used on the bidder files should be consistent.  If you're in a renaming state, can you please change one or the other so they match?

jsnellbaker (changes requested):
Hi @imedvedko 

I looked over the code here, and I believe it looks mostly okay; please see the in-line note about the name below.

When trying to test the adapter with the sample params provided, it was rejecting the bid because the adUnitId was set to 0 - which was causing this field to return a boolean false for the `isBidRequestValid()` function's assessment.  When I tried to set another value - I wasn't getting any bids back from the endpoint.

Is there another set of params I could use?  Or if there should be something returning regardless of the adUnitId, can you take a look to see why it may not be serving?

Thanks.

modules/limelightDigitalBidAdapter.js
imedvedko: Hi, @jsnellbaker 

I fixed `isBidRequestValid()`. You can use zero adUnitId in tests. Thanks you for report.

I use 'limelight-digital' everywhere instead of 'project-limelight' except one test domain. And I use 'limelightDigital' as a file name prefix everywhere. It is correct?

Thanks you for your work

imedvedko: feel free to message here if any changes required

jsnellbaker: No I meant the prefix for your files should match to the bidder code.  Could you change it to `limelightDigital`?

imedvedko: Oh, I fixed it. Thank you.

jsnellbaker (approved):
Thanks for the updates; LGTM

PR #7040, opened 2021-06-17
ATS-analytics-adapter - set sampling rate to 100

mamatic:
## Type of change

- [ ] Bugfix

## Description of change

Change sampling rate to be 100 always and if sampling rate is 0 do not fire analytics request. If sampling rate is set something other then 0, set it to 100.

## Other information
@mkendall07

### DIFF
--- a/modules/atsAnalyticsAdapter.js
+++ b/modules/atsAnalyticsAdapter.js
@@ -275,6 +275,7 @@ function sendDataToAnalytic () {
 
 // preflight request, to check did publisher have permission to send data to analytics endpoint
 function preflightRequest (envelopeSourceCookieValue) {
+  utils.logInfo('ATS Analytics - preflight request!');
   ajax(preflightUrl + atsAnalyticsAdapter.context.pid, function (data) {
     let samplingRateObject = JSON.parse(data);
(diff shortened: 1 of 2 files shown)


patmmccann:
This appears to be breaking and an antifeature, what is the goal here

mamatic:
> This appears to be breaking and an antifeature, what is the goal here

The goal is to lover requests that is fired on analytics endpoint and to have mechanism to turn off requests for analytics firing. If we set sampling rate to 100, it means that we will fire 1% of requests. Logic is here: 
 
 let shouldFireRequestValue = (Math.floor((Math.random() * 100 + 1)) === 100); 
 
 basically math.random needs to return value 0.99 to get true. 

And publisher can't change the value of sampling rate.
 
 I think that test was failing for some other part, but now it is ok.

mkendall07:
fyi Patrick the sampling rate comes from a cookie > https://github.com/prebid/Prebid.js/blob/16d730b00abd23fdca609dbe60d6e82a2612a456/modules/atsAnalyticsAdapter.js#L324

patmmccann:
Ah, I assumed it was coming from publisher config and this pr was too ignore the config, which seemed alarming

PR #7043, opened 2021-06-17
set BidsCube adapter

vlaktion:
The main goal is to figure out why BidsCube are not on the download list https://docs.prebid.org/download.html?
And where can I see Prebid Server company list? We have Prebid Server adapter too.


vlaktion:
Can you answer on my questions?

ChrisHuie:
@vlaktion yeah. I can get this updated today. It's likely an issue with the download page and the function that runs in the backend not finding the right module name.

ChrisHuie:
@vlaktion issue should be resolved and the adapter in now on the download page

Issue #6909, opened 2021-06-01
Measuring Viewability without GPT (BID_VIEWABLE event)


jsut:
## Type of issue
feature request / sanity / usefulness inquiry

## Description
Prebid has a module that provides an event, BID_VIEWABLE, that will fire when a rendered bid becomes viewable.  It only works with GAM/GPT, and relies on their to know when to emit the event. 

What I am wondering is if it might be useful or generally desirable to have an alternative to this module that would do some thing on the clientside after pbjs.renderAd was called that could result in an alternative way to generate a signal that an ad was viewable. I have been playing around with intersectionObserver, and it seems like it could be fairly easy to leverage it to get reasonable proxy that people could use as a first party measure for viewability.

I was playing with getting this working outside of prebid (you can see my experiment here: https://test.kabob.ca/), but I wonder if this might be generally useful to people as a module.

patmmccann:
Hard to say what the bid_viewable event is useful for now; seems to be only used by pubmatic managed publishers so I would ask them, but yes not relying on gam would be a great alternative.

jsut:
I personally want it so we can collect first party viewability data which will allow us to experiment on different things (timeouts, lazy load logic, who knows what else) and tune for viewability, rather than revenue, or fill, or cpm.

patmmccann:
Sounds reasonable. I think there have been some concerns with Prebid providing any sort of warranty on a viewable method and this technique ( https://developer.mozilla.org/en-US/docs/Web/API/Intersection_Observer_API/Timing_element_visibility ) is subject to a lot of false negatives when scripts bust ads out of frames and render them in an unexpected place. I think this method has value though, as long as sufficient disclaimers are in place.

jsut:
The disclaimers about the data and it's validity would definitely need to be large and bold.  I think having it as a module, as opposed to in prebid.js might offer some insulation there, but i'm not sure.

pm-harshad-mane:
Hello @jsut ,

I had added the bidViewability module in PrebidJS. I liked your approach. 

GPT provides a nice lazy-loading API https://developers.google.com/publisher-tag/samples/lazy-loading 
But GPT lazy loading does not let us run PrebidAuction before rendering ad-slots.
If we could build something similar to GPT lazy-loading but with the ability to execute custom code then it will be great, then a publisher will be able to run Prebid, TAM, and some custom logic. This can help boost the viewability of ads and hence CPMs.

[...]

PR #7059, opened 2021-06-19
Prebid Core: add AD_RENDER_SUCCEEDED event

jsut:
## Type of change

- [x] Feature

## Description of change

This adds a new event, AD_RENDER_SUCCEEDED, which is emited by renderAd when rendering is successful. It also adds a test in the same manner as the tests that exist for AD_RENDER_FAILED.

doc patch: https://github.com/prebid/prebid.github.io/pull/3045

## Other information

This is related to work to enable #6909

### DIFF
--- a/src/prebid.js
+++ b/src/prebid.js
@@ -23,7 +23,7 @@ const events = require('./events.js');
 const { triggerUserSyncs } = userSync;
 
 /* private variables */
-const { ADD_AD_UNITS, BID_WON, REQUEST_BIDS, SET_TARGETING, AD_RENDER_FAILED, STALE_RENDER } = CONSTANTS.EVENTS;
+const { ADD_AD_UNITS, BID_WON, REQUEST_BIDS, SET_TARGETING, AD_RENDER_FAILED, AD_RENDER_SUCCEEDED, STALE_RENDER } = CONSTANTS.EVENTS;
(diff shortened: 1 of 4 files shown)


src/prebid.js
```diff
@@ -442,6 +450,7 @@ $$PREBID_GLOBAL$$.renderAd = function (doc, id, options) {
 
           if (isRendererRequired(renderer)) {
             executeRenderer(renderer, bid);
+            emitAdRenderSucceeded({ doc, bid, id });
```
msm0504, line 453: Is it possible for the `executeRenderer` function to throw an error? If so, should this be in a `try/catch`?

jsut: I'm not familiar with the renderer code, but logically, it should be able to fail one way or another. None of these functions down this chain return anything explicitly. I'll dig into this and see if i can come up with a testcase that can make executeRenderer explode.  Presumably if this can fail, then there is somewhere in here core should emitAdRenderFailed that it currently isn't.

jsut: @msm0504, after looking at this a bit more, this section of code is already part of a big try/catch.  Most of this function is in it, and the catch already does emit the failure. So i think this is probably ok as is.  It doesn't look like there are tests for a renderer that actually throws something though, so i'm looking into adding some, but i'm not sure that should be part of this PR.

msm0504 (approved):

ChrisHuie (approved):

PR #7070, opened 2021-06-22
Renderer Testing: fix the renderer tests so they can be run on their own

jsut:
## Type of change
- [x] Other

## Description of change

running ./node_modules/gulp/bin/gulp.js test --file test/spec/renderer_spec.js fails without this change. The adUnits global is set by other tests, and the adloaderStub is loaded by test_index.js.  When you run this file on it's own, it will fail.


jsut:
There might be a better and/or more generic fix for this in karma.conf.maker.js.  test/spec/adUnits_spec.js, test/spec/auctionmanager_spec.js and test/spec/video_spec.js all fail if you try to run them on their own.

ChrisHuie (approved):

PR #7076, opened 2021-06-22
PBS adapter - fix logic for undefined endpoint URLs

### LINKED ISSUE #7046: Handle undefined noP1Consent endpoint in pbs s2sConfig
## Type of issue

bug / feature request

## Description

#6484 introduce two config options for requests with and without purpose1. We are hosting our own
prebid server and have not yet investigated how to faciliate the "no purpose 1" endpoint, so we didn't specify one.

```js
endpoint : {
  p1Consent: 'https://prebid.adnxs.com/pbs/v1/openrtb2/auction'
}
```

This causes a xhr call to `/undefined`, which is not ideal :stuck_out_tongue: 
If you are on example.com , prebid tries to call example.com/undefined

## Steps to reproduce

- Remove `noP1Consent` property

## Test page

None

### Expected results

Prebid server adapter is skipped

### Actual results

`/undefined` is called

## Platform details

prebid.js 4.43.0

## Other information

- #6484

jsnellbaker:
## Type of change

- [x] Bugfix

## Description of change
Fixes #7046

There was an issue present in s2sConfigs if the publisher defined the `endpoint` (or `syncEndpoint`) object config but didn't define both `p1Consent` and `noP1Consent` fields.  In this case, the field that was `undefined` would still try to make a call to PBS with an `undefined` URL for that type of user.  

This PR addresses this scenario by blocking the call to PBS entirely if a particular URL field was `undefined` for that type of user.  It will return an error for this scenario notifying what likely happened.  

Also added a warning when evaluating the s2sConfig if one of these fields were `undefined` when using the object config.

TO NOTE:
Publishers still using the old `string` version of the `endpoint` (or `syncEndpoint`) fields or who relied on the `defaultVendor` feature to populate the endpoint URLs are unaffected by this scenario/fix.  For the former, the single URL they define in the s2sConfig is still shared between consent/non-consent users.  For the latter, the URLs are still read/assigned based on the configs defined in Prebid.js.

### DIFF
--- a/modules/prebidServerBidAdapter/index.js
+++ b/modules/prebidServerBidAdapter/index.js
@@ -133,6 +133,13 @@ function formatUrlParams(option) {
       let temp = option[prop];
       option[prop] = { p1Consent: temp, noP1Consent: temp };
     }
+    if (utils.isPlainObject(option[prop]) && (!option[prop].p1Consent || !option[prop].noP1Consent)) {
+      ['p1Consent', 'noP1Consent'].forEach((conUrl) => {
+        if (!option[prop][conUrl]) {
+          utils.logWarn(`s2sConfig.${prop}.${conUrl} was not defined in manually set s2sConfig.  This may prevent PBS requests from executing in certain cases.  Please define field manually or instead use the defaultVendor settings of host company (if avialable).`);
+        }
+      });
+    }
   });
 }
 
@@ -1097,7 +1104,7 @@ export function PrebidServer() {
       const request = OPEN_RTB_PROTOCOL.buildRequest(s2sBidRequest, bidRequests, validAdUnits, s2sBidRequest.s2sConfig, requestedBidders);
       const requestJson = request && JSON.stringify(request);
       utils.logInfo('BidRequest: ' + requestJson);
-      if (request && requestJson) {
+      if (request && requestJson && getMatchingConsentUrl(s2sBidRequest.s2sConfig.endpoint, gdprConsent)) {
         ajax(
           getMatchingConsentUrl(s2sBidRequest.s2sConfig.endpoint, gdprConsent),
           {
@@ -1107,6 +1114,8 @@ export function PrebidServer() {
           requestJson,
           { contentType: 'text/plain', withCredentials: true }
         );
+      } else {
+        utils.logError('Prebid Server request did not execute.  Please ensure endpoint URL(s) are properly defined in your s2sConfig.  Otherwise check for warnings/errors, likely noted earlier in console logs...');
       }
     }
   };
--- a/modules/prebidServerBidAdapter/index.js
+++ b/modules/prebidServerBidAdapter/index.js
@@ -1104,9 +1104,10 @@ export function PrebidServer() {
       const request = OPEN_RTB_PROTOCOL.buildRequest(s2sBidRequest, bidRequests, validAdUnits, s2sBidRequest.s2sConfig, requestedBidders);
       const requestJson = request && JSON.stringify(request);
       utils.logInfo('BidRequest: ' + requestJson);
-      if (request && requestJson && getMatchingConsentUrl(s2sBidRequest.s2sConfig.endpoint, gdprConsent)) {
+      const endpointUrl = getMatchingConsentUrl(s2sBidRequest.s2sConfig.endpoint, gdprConsent);
+      if (request && requestJson && endpointUrl) {
         ajax(
-          getMatchingConsentUrl(s2sBidRequest.s2sConfig.endpoint, gdprConsent),
+          endpointUrl,
           {
             success: response => handleResponse(response, requestedBidders, bidRequests, addBidResponse, done, s2sBidRequest.s2sConfig),
             error: done
--- a/modules/prebidServerBidAdapter/index.js
+++ b/modules/prebidServerBidAdapter/index.js
@@ -136,7 +136,7 @@ function formatUrlParams(option) {
     if (utils.isPlainObject(option[prop]) && (!option[prop].p1Consent || !option[prop].noP1Consent)) {
       ['p1Consent', 'noP1Consent'].forEach((conUrl) => {
         if (!option[prop][conUrl]) {
-          utils.logWarn(`s2sConfig.${prop}.${conUrl} was not defined in manually set s2sConfig.  This may prevent PBS requests from executing in certain cases.  Please define field manually or instead use the defaultVendor settings of host company (if avialable).`);
+          utils.logWarn(`s2sConfig.${prop}.${conUrl} not defined.  PBS request will be skipped in some P1 scenarios.`);
         }
       });
     }
@@ -1115,7 +1115,7 @@ export function PrebidServer() {
           { contentType: 'text/plain', withCredentials: true }
         );
       } else {
-        utils.logError('Prebid Server request did not execute.  Please ensure endpoint URL(s) are properly defined in your s2sConfig.  Otherwise check for warnings/errors, likely noted earlier in console logs...');
+        utils.logError('PBS request not made.  Check endpoints.');
       }
     }
   };
--- a/test/spec/modules/prebidServerBidAdapter_spec.js
+++ b/test/spec/modules/prebidServerBidAdapter_spec.js
@@ -506,14 +506,55 @@ describe('S2S Adapter', function () {
       resetSyncedStatus();
     });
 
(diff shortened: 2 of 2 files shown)


modules/prebidServerBidAdapter/index.js
```diff
@@ -133,6 +133,13 @@ function formatUrlParams(option) {
       let temp = option[prop];
       option[prop] = { p1Consent: temp, noP1Consent: temp };
     }
+    if (utils.isPlainObject(option[prop]) && (!option[prop].p1Consent || !option[prop].noP1Consent)) {
+      ['p1Consent', 'noP1Consent'].forEach((conUrl) => {
+        if (!option[prop][conUrl]) {
+          utils.logWarn(`s2sConfig.${prop}.${conUrl} was not defined in manually set s2sConfig.  This may prevent PBS requests from executing in certain cases.  Please define field manually or instead use the defaultVendor settings of host company (if avialable).`);
```
bretg, line 139: I've been encouraging folks to keep messages short because they don't minify.

How about something more terse like "s2sConfig.${prop}.${conUrl} not defined. PBS request will be skipped in some P1 scenarios."

modules/prebidServerBidAdapter/index.js
```diff
@@ -1107,6 +1114,8 @@ export function PrebidServer() {
           requestJson,
           { contentType: 'text/plain', withCredentials: true }
         );
+      } else {
+        utils.logError('Prebid Server request did not execute.  Please ensure endpoint URL(s) are properly defined in your s2sConfig.  Otherwise check for warnings/errors, likely noted earlier in console logs...');
```
bretg, line 1118: Again, with the goal of saving bytes in the package, how about something more terse like

"PBS request not made. Check endpoints."

bretg (changes requested):

modules/prebidServerBidAdapter/index.js
jsnellbaker: updated.

modules/prebidServerBidAdapter/index.js
jsnellbaker: thanks, makes sense.  updated.

jsnellbaker:
updated messages and added unit test

modules/prebidServerBidAdapter/index.js
```diff
@@ -133,6 +133,13 @@ function formatUrlParams(option) {
       let temp = option[prop];
       option[prop] = { p1Consent: temp, noP1Consent: temp };
     }
+    if (utils.isPlainObject(option[prop]) && (!option[prop].p1Consent || !option[prop].noP1Consent)) {
```
msm0504, line 136: I think this can be simplified to
```
if (utils.isPlainObject(option[prop]))
```
The other conditions are being checked again within the `forEach`

modules/prebidServerBidAdapter/index.js
```diff
@@ -1097,7 +1104,7 @@ export function PrebidServer() {
       const request = OPEN_RTB_PROTOCOL.buildRequest(s2sBidRequest, bidRequests, validAdUnits, s2sBidRequest.s2sConfig, requestedBidders);
       const requestJson = request && JSON.stringify(request);
       utils.logInfo('BidRequest: ' + requestJson);
-      if (request && requestJson) {
+      if (request && requestJson && getMatchingConsentUrl(s2sBidRequest.s2sConfig.endpoint, gdprConsent)) {
```
muuki88, line 1107: Maybe put the URL into a variable. Makes refactoring easier as there's only one place to look for

test/spec/modules/prebidServerBidAdapter_spec.js
```diff
@@ -506,6 +506,19 @@ describe('S2S Adapter', function () {
...
+      config.setConfig({ s2sConfig: badConfig });
+
+      let badCfgRequest = utils.deepClone(REQUEST);
+      badCfgRequest.s2sConfig = badConfig;
+
+      adapter.callBids(badCfgRequest, BID_REQUESTS, addBidResponse, done, ajax);
+
+      expect(server.requests.length).to.equal(0);
+    });
+
```
muuki88, line 521: Maybe add two more test cases

1. for `p1Consent` only configuration (which is probably the more common one)
2. for an empty object


```suggestion
    [
      { p1Consent: 'https://prebid.adnxs.com/pbs/v1/openrtb2/auction' },
      { noP1Consent: 'https://prebid.adnxs.com/pbs/v1/openrtb2/auction' },
      { }
    ].forEach(endpoint => {
        it('should block request if config did not define proper URL in endpoint object config', function() {
          let badConfig = utils.deepClone(CONFIG);
          badConfig.endpoint = endpoint;
          config.setConfig({ s2sConfig: badConfig });

          let badCfgRequest = utils.deepClone(REQUEST);
          badCfgRequest.s2sConfig = badConfig;

          adapter.callBids(badCfgRequest, BID_REQUESTS, addBidResponse, done, ajax);

          expect(server.requests.length).to.equal(0);
        });
    });
```

muuki88 (changes requested):

muuki88:
Thanks for the fast fix @jsnellbaker  :heart:

modules/prebidServerBidAdapter/index.js
jsnellbaker: That's not the same.  

The `forEach` above this code is handling the condition when `endpoint: 'someURLstring'`.  The if-check here checks if the `endpoint` is an object config and it's missing either of the properties `p1Consent` or `noP1Consent`.

test/spec/modules/prebidServerBidAdapter_spec.js
jsnellbaker: Do you mean add a test for when p1Consent is the only property defined and the request contains a no-consent scenario - to test that the request is blocked here as well?

muuki88: yes :)

modules/prebidServerBidAdapter/index.js
msm0504: Sorry, I meant the `forEach` within the `if` statement. It checks if either of those properties is missing again.

jsnellbaker: oh - I see what you meant.  I was hoping to avoid this forEach if the config was setup 'correctly' - that's why I had them included in the outer if-check.

test/spec/modules/prebidServerBidAdapter_spec.js
jsnellbaker: Ok thanks for confirming.  I added some tests guessing that was the case.  Please take a look when you have the chance (in addition to the other change you suggested).

test/spec/modules/prebidServerBidAdapter_spec.js
```diff
@@ -506,6 +506,60 @@ describe('S2S Adapter', function () {
...
+      badConfig.endpoint = { p1Consent: 'https://prebid.adnxs.com/pbs/v1/openrtb2/auction' };
+      config.setConfig({ s2sConfig: badConfig });
+
+      let badCfgRequest = utils.deepClone(REQUEST);
+      badCfgRequest.s2sConfig = badConfig;
+
+      let badBidderRequest = utils.deepClone(BID_REQUESTS);
+      badBidderRequest[0].gdprConsent = {
+        consentString: 'abc123',
+        addtlConsent: 'superduperconsent',
```
muuki88, line 533: I like  your google consent encoding :joy:

modules/prebidServerBidAdapter/index.js
```diff
@@ -1097,16 +1104,19 @@ export function PrebidServer() {
       const request = OPEN_RTB_PROTOCOL.buildRequest(s2sBidRequest, bidRequests, validAdUnits, s2sBidRequest.s2sConfig, requestedBidders);
       const requestJson = request && JSON.stringify(request);
       utils.logInfo('BidRequest: ' + requestJson);
-      if (request && requestJson) {
+      const endpointUrl = getMatchingConsentUrl(s2sBidRequest.s2sConfig.endpoint, gdprConsent);
```
muuki88, line 1107: Awesome :+1:

muuki88 (approved):
Thanks for the quick fix :clap: :clap:

msm0504 (approved):

ChrisHuie (approved):

muuki88:
@bretg is this good to go? We have some publishers that have way to many calls to `/undefined` :wink:

PR #7085, opened 2021-06-23
tappx Bid Adapter: refactor and fix gdpr user param

prebidtappx:
## Type of change

- [X] Bugfix

## Description of change

- We have fixed an error that was not controlled when registering the gdpr.
- We have added the user to the payload as we do with all objects.


ChrisHuie (approved):

prebidtappx:
Thank you @ChrisHuie 👍

PR #7091, opened 2021-06-23
Prebid Core: Update `renderAd` to use hook

iamnewton:
## Type of change

- [x] Feature

## Description of change

Updating `renderAd` to use hook in preparation to use in a Fledge proposal test.


jsut:
It seems like this could replace #7059. Hooks seem to be more flexible than the event system is, but perhaps there are trade offs I am not aware of?

robertrmartinez:
@jsut 

I think both make sense.

Adding and using this hook delays the execution of the renderAd function until your piece of code finishes and calls it directly. This hook stuff is mainly used to alter and change things before the hooked function executes.

The Event Library is more for notifying, and is non blocking. So I think both have a place, and AD_RENDER_SUCCEEDED seems like a valid event (Since there are times when a BID_WON Does not equal a Ad Successfully rendering)

robertrmartinez:
@iamnewton Will you pull latest master, there was a fix for a flaky test merged.

robertrmartinez (commented):
Looks fine.

I would like @snapwich to give his input since the hook library is his baby.

Rich, what are the impacts of adding this hook in the scenario where 99.9% of prebid wrappers will not actually register a hook.

What kind of overhead is there?

snapwich:
very little overhead. this should be fine.

as for the event, hooks don't replace events; if you need the event i'd add that as well.

snapwich (approved):

jsut:
@snapwich @robertrmartinez thanks for the context.

robertrmartinez (approved):

PR #7095, opened 2021-06-24
Gumgum: ADTS-134 Fetch IDL envelope and pass to ad server if available

lbenmore:
## Type of change

- [ ] Bugfix
- [x] Feature
- [ ] New bidder adapter   
- [ ] Build related changes
- [ ] CI related changes
- [ ] Other

## Description of change
Check for the availability of IDL envelope from LiveRamp and send with request

### DIFF
--- a/modules/gumgumBidAdapter.js
+++ b/modules/gumgumBidAdapter.js
@@ -83,14 +83,6 @@ function getWrapperCode(wrapper, data) {
   return wrapper.replace('AD_JSON', window.btoa(JSON.stringify(data)))
 }
 
-function _getTradeDeskIDParam(userId) {
-  const unifiedIdObj = {};
-  if (userId.tdid) {
-    unifiedIdObj.tdid = userId.tdid;
-  }
-  return unifiedIdObj;
-}
-
 function _getDigiTrustQueryParams(userId) {
   let digiTrustId = userId.digitrustid && userId.digitrustid.data;
   // Verify there is an ID and this user has not opted out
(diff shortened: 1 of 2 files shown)


FilipStamenkovic (changes requested):
Please open docs PR to include userIds to your bid adapter.  

Also, you have separate code for trade desk id, in function:

```
_getTradeDeskIDParam
```

Perhaps you don't need this function now?

lbenmore:
@FilipStamenkovic thanks for pointing that out! i've consolidated how we're sending `userId` values and made it more flexible. will update with docs PR shortly

lbenmore:
@FilipStamenkovic oh sorry, this change actually doesn't affect our documentation as this is not a new parameter expected but a retrieval of available data

FilipStamenkovic:
@lbenmore I know you didn't add any new parameter, but you need to update your supported `userIds` in the docs.  
Right now, it's stated that you support only `unifiedId`:
https://github.com/prebid/prebid.github.io/blob/master/dev-docs/bidders/gumgum.md

You need to add other ids.

FilipStamenkovic:
Some of these ids you are not passing correctly.  
For some of them you are passing: `[object Object]` in your request (see screenshot).

image



Can you fix this?

lbenmore:
@FilipStamenkovic, ah I see. I've updated the docs with a PR here: https://github.com/prebid/prebid.github.io/pull/3066

modules/userId/eids.js
```diff
@@ -215,7 +215,7 @@ const USER_IDS_CONFIG = {
 };
 
 // this function will create an eid object for the given UserId sub-module
-function createEidObject(userIdData, subModuleKey) {
+export function createEidObject(userIdData, subModuleKey) {
```
FilipStamenkovic, line 218: You shouldn't be modifying pbjs core with this PR.

modules/gumgumBidAdapter.js
```diff
@@ -3,6 +3,7 @@ import * as utils from '../src/utils.js'
 import { BANNER, VIDEO } from '../src/mediaTypes.js';
 
 import { config } from '../src/config.js'
+import { createEidObject } from './userId/eids.js'
```
FilipStamenkovic, line 6: Instead of modifying pbjs, try using: `createEidsArray` from `eids.js`.  
It's already exported and I see it's being used by a lot of other bidders already.

FilipStamenkovic (changes requested):

lbenmore:
@FilipStamenkovic updated to not use `createEidObject` function (and remove export of it). our ad server is expecting these values named in a way that doesn't work with `createEidsArray` so our `getEids` function has been modified to retrieve and set the values accordingly

FilipStamenkovic (approved):

PR #6808, opened 2021-05-21
Pubcommon id and shared id submodules: merge into one submodule

NothingIsMatter:
## Type of change
- [ x] Feature

## Description of change
Merge pubcid and sharedid
## Other information
Issue - https://github.com/prebid/Prebid.js/issues/6640

PR #7099, opened 2021-06-25
sharedIdSystem: return pubcid instead of sharedId

philipwatson:
## Type of change

- [x] Bugfix

## Description of change
I had a look at `sharedIdSystem.js` after the changes made in https://github.com/prebid/Prebid.js/pull/6808 for issue https://github.com/prebid/Prebid.js/issues/6640

I'm confused.

It doesn't make sense to me that the `sharedId` is passed into the callback. Should it be `pubcid` instead?

It could do with more tests. I didn't invest much time there because my understanding could be wrong.

From unit testing, I noticed `storage.expires` in the storeData function does not exist (or does it?). So I changed to `config.storage.expires`. Not sure if this is correct.

### DIFF
--- a/modules/sharedIdSystem.js
+++ b/modules/sharedIdSystem.js
@@ -12,7 +12,7 @@ import { uspDataHandler, coppaDataHandler } from '../src/adapterManager.js';
 import {getStorageManager} from '../src/storageManager.js';
 
 const GVLID = 887;
-const storage = getStorageManager(GVLID, 'pubCommonId');
+export const storage = getStorageManager(GVLID, 'pubCommonId');
 const COOKIE = 'cookie';
 const LOCAL_STORAGE = 'html5';
 const SHAREDID_OPT_OUT_VALUE = '00000000000000000000000000';
@@ -33,7 +33,7 @@ function storeData(config, value) {
   try {
     if (value) {
       const key = config.storage.name + SHAREDID_SUFFIX;
-      const expiresStr = (new Date(Date.now() + (storage.expires * (60 * 60 * 24 * 1000)))).toUTCString();
+      const expiresStr = (new Date(Date.now() + (config.storage.expires * (60 * 60 * 24 * 1000)))).toUTCString();
 
       if (config.storage.type === COOKIE) {
         if (storage.cookiesAreEnabled()) {
@@ -123,7 +123,7 @@ function handleResponse(pubcid, callback, config) {
             }
           }
           // Pass pubcid even though there is no change in order to trigger decode
-          callback(responseObj.sharedId);
+          callback(pubcid);
         } catch (error) {
           utils.logError(error);
         }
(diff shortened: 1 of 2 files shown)


jdwieland8282:
@NothingIsMatter Mind taking a look here?

jdwieland8282:
@ChrisHuie re:  "return pubcid instead of sharedId 4 days ago"  can you ad more here?  The user id generated by the module formally known as pubcommon and the module formally known as sharedid, were just random string generators. The user id output of either shouldn't matter.  

The intention was to have a guid set in a 1pc, then added to the eids array using pubcid.org as the source.

philipwatson:
Let me give a scenario to help raise the alarm a little ;-)

Given default `auctionDelay` of 0. A new user. On page load:

Bidders receive `bid.userId` as: 
```
{
   pubcid: "0ea08523-0ceb-4ae6-bde3-84b34c2eb893"
}
```
The value is a UUID created on the page (by sharedIdSystem's `getId`). All good here.

Then I look at the cookies in developer tools. I see the cookies created via sharedIdSystem:
_pubcid = 01EWC4FHX0QQEV1HGR16HAQGPM
_pubcid_sharedid = 01EWC4FHX0QQEV1HGR16HAQGPM

They are set to the same ULID value from sharedid server. The callback, which is run after the auction, replaces the established UUID with the ULID. Note: this will also happen to existing users (not just new users) when `getId` is called for a refresh.

So on the next page load, bidders receive `bid.userId` as:

```
{
   sharedid: {id: "01EWC4FHX0QQEV1HGR16HAQGPM"},
   pubcid: "01EWC4FHX0QQEV1HGR16HAQGPM"
}
```

For completeness, `bid.userIdAsEids` has:

```
[{
  source: "pubcid.org",
  uids: [{
    id: "01EWC4FHX0QQEV1HGR16HAQGPM",
    atype: 1
  }]
}]
```
Only pubcid.org here which is expected after reading the code - sharedId is filtered out (or do we want to see two items in the `uids` array here?)

sharedIdSystem manages the storage of `_pubcid_sharedid`.. but also gets the user id module to manage the storage of `_pubcid` with the same value. Obviously there is a problem here right?

jdwieland8282:
Hey @philipwatson thanks for the example, we are taking a look, it does indeed look like a problem.  What we want in the eids array according to you example is indeed. 

```[{
  source: "pubcid.org",
  uids: [{
    id: "0ea08523-0ceb-4ae6-bde3-84b34c2eb893",
    atype: 1
  }]
}]

ChrisHuie:
@philipwatson do you mind pulling in the latest commits from master to make sure the circleci tests are passing for this pr 🙏

philipwatson:
No more changes coming in this PR. 
I've merged master. Can anyone rerun the build? An unrelated test is suspiciously timing out even though there is not much in it. Same test is failing on another PR.

SKOCHERI:
@philipwatson The SharedIdSystem need not call the Sharedid endpoint. The same has been updated in https://github.com/prebid/Prebid.js/pull/7149 and the SharedIdSystem module is cleaned up to removed read/write cookie. Please take a look at this PR.

ChrisHuie (approved):

PR #7104, opened 2021-06-25
Prebid Analytics Manager: extend mocha testing timeout to fix flaky ie test

ChrisHuie:
Fix flaky IE test. Extended the timeout on the test to 4000ms


robertrmartinez (approved):
Not the best fix but if it helps let's merge it!

Issue #6783, opened 2021-05-18
UserID module:  Runtime configuration update


slimkrazy:
## Type of issue
Question

## Description
The VerizonMedia user ID module requires a publisher provided value prior to it being able to function correctly. We have a case of a publisher who is generating this value _after_ Prebid has been initially configured - on an event triggered by a user interaction. I've been digging around in the code and it seems that:
  * Once `setConfig` has been invoked with userId modules configured, there is no way to add any additional userID modules at a later time during that page session
  * Dynamically altering userID module configuration values after initialisation will not work
  
Can anybody confirm that these two observations are indeed true? If they are, would there be opposition to modifying the code to potentially allow for additional userID modules to be added following initialisation or allowing for the config to be updated?

smenzer:
I haven't tested this, but I thought that you should be able to modify the config of the user id module via a new `setConfig()` call, then you could do a `refreshIds()` call to update the id modules and make the new ID available via upcoming bid requests or with `getUserIds()` call.

slimkrazy:
@smenzer  - I really wish this worked, but it doesn't. 

I am able to get the second user ID module to update its config, and I do see the updated value reflected in localstorage, however in the same page session I never see the updated made available via any API call. This simple snippet should explain:

```
       // setConfig API call made here, updating userId module config
      // Then...
        pbjs.refreshUserIds({}, function() {
          pbjs.requestBids({
            bidsBackHandler: function() {
              console.log(pbjs.getUserIds()); // This is always returns the old values. New values are in localstorage. Bid adapters also only receive old values.
            }
          });
        });
```

smenzer:
Ok, so it does partially work, just not entirely. Seems like there's something missing around not calling `addIdDataToAdUnitBids()` during the refresh flow. I'm not as familiar with how this part of the code works, though. 

@ChrisHuie @jsnellbaker @idettman @Fawke would any of you have more insights here?

Fawke:
Hi @slimkrazy, can you see if https://github.com/prebid/Prebid.js/pull/7105 works for you?

Fawke:
Closing this after the merge of #7105, please reopen if still not resolved.

PR #7105, opened 2021-06-25
UserID: bugfix for refreshUserIds with updated info

aaroniniguez:
## Type of change

- [ x] Bugfix
- [ ] Feature
- [ ] New bidder adapter   
- [ ] Build related changes
- [ ] CI related changes
- [ ] Other

## Description of change
calling pbjs.refreshUserIds() with updated UserId info does not update adUnitBids (since addIdDataToAdUnitBids is passed in initializedSubmodules which is not updated by pbjs.refreshUserIds())

## Other information
Fixes this issue: 
https://github.com/prebid/Prebid.js/issues/6783

### DIFF
--- a/modules/userId/index.js
+++ b/modules/userId/index.js
@@ -722,6 +722,7 @@ function updateInitializedSubmodules(submodule) {
     if (submodule.config.name.toLowerCase() === initializedSubmodules[i].config.name.toLowerCase()) {
       updated = true;
       initializedSubmodules[i] = submodule;
+      break;
     }
   }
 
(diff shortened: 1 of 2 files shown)


Fawke (commented):
Hi @aaroniniguez, thank you for fixing this bug. Your changes look good to me, there's just an inline comment, see if we'd want to do that. Thanks.

modules/userId/index.js
```diff
@@ -711,6 +716,20 @@ function initSubmodules(submodules, consentData) {
   }, []);
 }
 
+function updateInitializedSubmodules(submodule) {
+  let updated = false;
+  for (let i = 0; i < initializedSubmodules.length; i++) {
+    if (submodule.config.name.toLowerCase() === initializedSubmodules[i].config.name.toLowerCase()) {
+      updated = true;
+      initializedSubmodules[i] = submodule;
+    }
```
Fawke, line 725: Can we break out of this for loop if there's a match?

aaroniniguez: updated

Fawke (approved):

Fawke:
Adding @smenzer for a second review since it's a core change.

smenzer (approved):
LGTM

PR #7108, opened 2021-06-26
AirGrid RTD Submodule: Initial Release

ydennisy:
## Type of change
- [x] Feature

## Description of change

This PR adds a new RTD submodule, this module augments the bid `Bids/AdUnits` for the AppNexus/Xandr bid adapter by attaching segment values to the `keywords` param.

The flow of the module:
- Inject a 3rd party script into the DOM, we do not need to wait for this script to load, we do this in an async manner.
- Fetch audience values from `localStorage` using the `StorageManager`
- If audiences are available augment the `bids` object

`dennis@airgrid.io`

## Other information
As noted above this module loads 3rd party JS into the page.

### DIFF
--- a/modules/airgridRtdProvider.js
+++ b/modules/airgridRtdProvider.js
@@ -37,7 +37,7 @@ export function attachScriptTagToDOM(rtdConfig) {
       n.async = true;
       n.src = 'https://cdn.edkt.io/' + p + '/edgekit.min.js';
       var a = document.getElementsByTagName('head')[0];
-      a.parentNode.insertBefore(n, a);
+      a.parentNode.insertBefore(n, a.nextSibling);
     };
     edktInitializor.load(edktInitializor.accountId);
   }
--- a/modules/airgridRtdProvider.js
+++ b/modules/airgridRtdProvider.js
@@ -1,3 +1,4 @@
+/* eslint-disable */
 /**
  * This module adds the AirGrid provider to the real time data module
  * The {@link module:modules/realTimeData} module is required
@@ -5,8 +6,9 @@
  * @module modules/airgridRtdProvider
  * @requires module:modules/realTimeData
  */
-
+import {config} from '../src/config.js';
 import {submodule} from '../src/hook.js';
+import {mergeDeep, isPlainObject, deepSetValue, deepAccess} from '../src/utils.js';
 import {getGlobal} from '../src/prebidGlobal.js';
 import {getStorageManager} from '../src/storageManager.js';
 
@@ -20,6 +22,7 @@ export const storage = getStorageManager(AG_TCF_ID, SUBMODULE_NAME);
 /**
  * Attach script tag to DOM
  * @param {Object} rtdConfig
+ * @return {void}
  */
 export function attachScriptTagToDOM(rtdConfig) {
   var edktInitializor = window.edktInitializor = window.edktInitializor || {};
@@ -59,8 +62,9 @@ export function getMatchedAudiencesFromStorage() {
  * Mutates the adUnits object
  * @param {Object} adUnits
  * @param {Array} audiences
+ * @return {void}
  */
-function appendAudiencesToAdUnits(adUnits, audiences) {
+function setAudiencesToAppNexusAdUnits(adUnits, audiences) {
   adUnits.forEach((adUnit) => {
     adUnit.bids.forEach((bid) => {
       if (bid.bidder && bid.bidder === 'appnexus') {
@@ -70,6 +74,31 @@ function appendAudiencesToAdUnits(adUnits, audiences) {
       }
     })
   })
+}
+
+/**
+ * Pass audience data to configured bidders, using ORTB2
+ * @param {Object} rtdConfig
+ * @param {Array} audiences
+ * @return {void}
+ */
+export function setAudiencesUsingBidderOrtb2(rtdConfig, audiences) {
+  const bidders = deepAccess(rtdConfig, 'params.bidders');
+  if (!bidders || bidders.length === 0) return;
+  const allBiddersConfig = config.getBidderConfig();
+  const agOrtb2 = {}
+  deepSetValue(agOrtb2, 'ortb2.user.ext.data.airgrid', audiences || []);
+  
+  bidders.forEach((bidder) => {
+    let bidderConfig = {};
+    if (isPlainObject(allBiddersConfig[bidder])) {
+      bidderConfig = allBiddersConfig[bidder];
+    }
+    config.setBidderConfig({
+      bidders: [bidder],
+      config: mergeDeep(bidderConfig, agOrtb2)
+    });
+  });
 }
 
 /**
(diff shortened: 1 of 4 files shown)


patmmccann:
Hi,  even if other bidders don't currently understand your data, it is important to future proof your adapter now by allowing publishers to write your data to global or the ortb2 object for a bidder that they choose.  See the sir data module for an example and the current discussion on #7001

ydennisy:
Hi @patmmccann thanks for taking a look so quick!

I am just fixing up one of the failed tests :)

I appreciate your comment and agree with it - but right now we do not have any relationship with other bidders, meaning we would have no way commercialise this data for ourselves or the publisher.

If your suggestion is a requirement to have this approved - please let me know. Since the segment values are fairly opaque this should not be a deal breaker.

patmmccann:
It is a requirement; I think you'll find the discussion on 7001 helpful to see why

ydennisy:
Ok thanks @patmmccann taking a look right now 😄

ydennisy:
Hi @patmmccann , thanks again for taking a look at the PR - I have now read the discussion you linked from #7001 

Am I correct in understanding that the following flow would be correct in our use case:
- we provide a `bidders` key in the publisher config params, where publishers can decide where to send segments
- our module then uses `setBidderConfig` to honour this param

Please let me know if the above implementation would be correct.

Based on the above I have two further questions:
- Xandr does not support use `getConfig` in their adapter, so we are ok to proceed with the current implementation in addition to setting the bidder specific config until they switch over?
- The second question is more about the philosophy of this requirement, I see that you state that RTD modules should not be the ones deciding who to share data with, but rather publishers. If RTD data can only be monetised via a subset of bidders, both in terms of value for the RTD provider and Publisher, why would it make sense to pass these values to all bidders on a page? Also, in some business models I imagine the RTD provider does not charge for the module but rather when their segments are activated, passing their segments to all bidders would make this impossible.

patmmccann:
That workflow would be great,  and you are also welcome to keep the
workflow you have now to accommodate appnexus or any bidder you choose that
hasn't yet built to the new first party data spec.

Doing this keeps your adapter future proof, so any adapter might be able to
light up the same deal ids as appnexus without waiting for you to build
something. I assume this is a feature you find desirable?

ydennisy:
Hi @patmmccann I have made the changes as we discussed.

Two questions which have arisen:
- using `setBidderConfig` seems to set the config on all bidders, even when a single bidder is provided in the way I have used it, maybe there is an error here.
- what should be the exact data format / key for AG audiences to be passed via ORTB2? I have seen references to a new `segtax` format, is this what should be followed?

ydennisy:
Hello @patmmccann @msm0504 - just wanted to check in here to see when you would have time to review the PR?

All the best,
Dennis

modules/airgridRtdProvider.js
```diff
@@ -0,0 +1,141 @@
...
+    edktInitializor.publisherId = rtdConfig.params.publisherId;
+    edktInitializor.apiKey = rtdConfig.params.apiKey;
+    edktInitializor.load = function(e) {
+      var p = e || 'sdk';
+      var n = document.createElement('script');
+      n.type = 'text/javascript';
+      n.async = true;
+      n.src = 'https://cdn.edkt.io/' + p + '/edgekit.min.js';
+      var a = document.getElementsByTagName('head')[0];
+      a.parentNode.insertBefore(n, a);
```
msm0504, line 40: This appears to place the `<script>` tag outside of and before the `<head>` tag. Shouldn't the script be loaded within `<head>`?

modules/airgridRtdProvider.js
```diff
@@ -0,0 +1,141 @@
...
+ * Mutates the adUnits object
+ * @param {Object} adUnits
+ * @param {Array} audiences
+ * @return {void}
+ */
+function setAudiencesToAppNexusAdUnits(adUnits, audiences) {
+  adUnits.forEach((adUnit) => {
+    adUnit.bids.forEach((bid) => {
+      if (bid.bidder && bid.bidder === 'appnexus') {
+        if (!bid.params) bid.params = {};
```
msm0504, line 70: `deepSetValue` can be used here

msm0504 (changes requested):

modules/airgridRtdProvider.js
ydennisy: @msm0504 yeah good catch thank you!

I have used `nextSibling` on the reference node to ensure we insert after the `head` tag.

modules/airgridRtdProvider.js
ydennisy: @msm0504 yeah agreed, have updated to use `deepSetValue`

patmmccann:
> using setBidderConfig seems to set the config on all bidders, even when a single bidder is provided in the way I have used it, maybe there is an error here.

hmm, take a look at the latest approach at #7001 ; does this suffer from the same issue? Perhaps we have a bug. 


> what should be the exact data format / key for AG audiences to be passed via ORTB2? I have seen references to a new segtax format, is this what should be followed?

you can put data in site.ext.data.airgrid or you could put it in site.content.data if you want to follow the spec in #6057 with an enumerated segtax; or you could get your own segtax enumeration. 

see https://docs.prebid.org/features/firstPartyData.html ; openrtb pull 65 ( https://github.com/InteractiveAdvertisingBureau/openrtb/pull/65 ) and adcom pull 22, linked within the iab pull

modules/airgridRtdProvider.js
msm0504: Shouldn't the `script` tag be inserted inside of the `head` tag?

ydennisy:
@msm0504 sorry I missed the ping - you are once again correct, this is now rectified in the latest commit.

@patmmccann seems the new segtax format is still a little way off being finalised - I will submit to being added to that list but for now I have kept the data in `user.ext.data.airgrid`.

Your suggestion was `site.ext.data.airgrid`, but this feels where contextual / page related data would go, rather than audience / segment information. Let me know if I have misunderstood.

modules/airgridRtdProvider.js
```diff
@@ -0,0 +1,138 @@
...
+
+/**
+ * Real-time data retrieval from AirGrid
+ * @param {Object} reqBidsConfigObj
+ * @param {function} onDone
+ * @param {Object} rtdConfig
+ * @param {Object} userConsent
+ * @return {void}
+ */
+export function passAudiencesToBidders(bidConfig, onDone, rtdConfig, userConsent) {
```
msm0504, line 119: Is the `userConsent` param being used for anything? Does consent impact this module?

ydennisy: @msm0504 From my understanding consent will fail when we try and use the storage manager if not granted, we initialise it with out TCF vendor ID, so then the rest of the module is in essence a no-op since the storage manager will not return any data.

msm0504 (approved):

msm0504:
@patmmccann I've approved this PR. Have all of your comments been addressed? Can this be merged?

patmmccann:
I merged, thanks @msm0504 !

patmmccann:
https://cdn.edkt.io/sdk/edgekit.min.js seems to generate a script error in ie11. Have you noticed this ?
https://app.circleci.com/pipelines/github/prebid/Prebid.js/8067/workflows/7312b412-4840-4c0e-b3a6-79bd9994a203/jobs/17937

PR #7116, opened 2021-06-28
Teads Bid Adapter: support floc and uid2 user IDs

ruizb:
## Type of change
- [x] Feature

## Description of change
Support **FLoC ID** (`flocId`) and **Unified ID v2** (`uid2Id`) user IDs in the Teads adapter.

Additionally, read the `_tfpvi` first-party cookie when available on the website's domain.

Documentation has been updated on the `prebid.github.io` repository: https://github.com/prebid/prebid.github.io/pull/3072.

Contact email of the adapter’s maintainer : innov-ssp@teads.tv.

### DIFF
--- a/modules/teadsBidAdapter.js
+++ b/modules/teadsBidAdapter.js
@@ -12,7 +12,7 @@ const gdprStatus = {
   CMP_NOT_FOUND_OR_ERROR: 22
 };
 const FP_TEADS_ID_COOKIE_NAME = '_tfpvi';
-const storage = getStorageManager(GVL_ID, BIDDER_CODE);
+export const storage = getStorageManager(GVL_ID, BIDDER_CODE);
 
 export const spec = {
   code: BIDDER_CODE,
--- a/modules/teadsBidAdapter.js
+++ b/modules/teadsBidAdapter.js
@@ -1,14 +1,17 @@
 import {registerBidder} from '../src/adapters/bidderFactory.js';
-const utils = require('../src/utils.js');
+import {getStorageManager} from '../src/storageManager.js';
+import * as utils from '../src/utils.js';
+
 const BIDDER_CODE = 'teads';
 const GVL_ID = 132;
-const ENDPOINT_URL = 'https://a.teads.tv/hb/bid-request';
+const ENDPOINT_URL = '//SSP_PORT_8080_TCP_ADDR:SSP_PORT_8080_TCP_PORT/hb/bid-request';
 const gdprStatus = {
   GDPR_APPLIES_PUBLISHER: 12,
   GDPR_APPLIES_GLOBAL: 11,
   GDPR_DOESNT_APPLY: 0,
   CMP_NOT_FOUND_OR_ERROR: 22
-}
+};
+const FP_TEADS_ID_COOKIE_NAME = '_tfpvi';
 
 export const spec = {
   code: BIDDER_CODE,
@@ -41,6 +44,7 @@ export const spec = {
    */
   buildRequests: function(validBidRequests, bidderRequest) {
     const bids = validBidRequests.map(buildRequestObject);
+
     const payload = {
       referrer: getReferrerInfo(bidderRequest),
       pageReferrer: document.referrer,
@@ -48,7 +52,10 @@ export const spec = {
       timeToFirstByte: getTimeToFirstByte(window),
       data: bids,
       deviceWidth: screen.width,
-      hb_version: '$prebid.version$'
+      hb_version: '$prebid.version$',
+      ...getFLoCParameters(utils.deepAccess(validBidRequests, '0.userId.flocId')),
+      ...getUnifiedId2Parameter(utils.deepAccess(validBidRequests, '0.userId.uid2')),
+      ...getFirstPartyTeadsIdParameter()
     };
 
     if (validBidRequests[0].schain) {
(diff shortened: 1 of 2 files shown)


modules/teadsBidAdapter.js
```diff
@@ -1,14 +1,17 @@
 import {registerBidder} from '../src/adapters/bidderFactory.js';
-const utils = require('../src/utils.js');
+import {getStorageManager} from '../src/storageManager.js';
+import * as utils from '../src/utils.js';
+
 const BIDDER_CODE = 'teads';
 const GVL_ID = 132;
-const ENDPOINT_URL = 'https://a.teads.tv/hb/bid-request';
+const ENDPOINT_URL = '//SSP_PORT_8080_TCP_ADDR:SSP_PORT_8080_TCP_PORT/hb/bid-request';
```
robertrmartinez, line 7: What exactly is going on here?

This looks like some macro to be replaced but I do not see it happening anywhere in your adapter?

`SSP_PORT_8080_TCP_ADDR:SSP_PORT_8080_TCP_PORT`

modules/teadsBidAdapter.js
```diff
@@ -221,4 +234,40 @@ function _validateId(id) {
...
+function getUnifiedId2Parameter(optionalUid2) {
+  return optionalUid2 ? { unifiedId2: optionalUid2.id } : {};
+}
+
+/**
+ * Get the first-party cookie Teads ID parameter to be sent in bid request.
+ * @returns `{} | {firstPartyCookieTeadsId: string}`
+ */
+function getFirstPartyTeadsIdParameter() {
+  const storage = getStorageManager(GVL_ID, BIDDER_CODE);
```
robertrmartinez, line 265: Think it would be better to just init your storage manager top level of module

You are calling this function every single time anyways.

robertrmartinez (changes requested):
couple questions

modules/teadsBidAdapter.js
nailyk: yes it's an error on our side, we will revert that change!

Viewtiful:
Hello @robertrmartinez , we have done the changes following your questions.
Could you please have another look and review again the PR when you have the time to do so?
Thank you !

test/spec/modules/teadsBidAdapter_spec.js
```diff
@@ -441,7 +449,129 @@ describe('teadsBidAdapter', () => {
...
+
+        it('should not add firstPartyCookieTeadsId param to payload if first-party cookie is not available', function () {
+          const request = spec.buildRequests([baseBidRequest], bidderResquestDefault);
+          const payload = JSON.parse(request.data);
+
+          expect(payload).not.to.have.property('firstPartyCookieTeadsId');
+        });
+
+        it('should add firstPartyCookieTeadsId param to payload if first-party cookie is available', function () {
+          storage.setCookie('_tfpvi', 'my-teads-id');
```
robertrmartinez, line 567: We do not want to actually be setting and reading from storage / cookies inside of tests.

We should be mocking the stubbing / mocking the storage manager, which has its own tests to verify it works.

Here is an example:

You'll need to export your storage var so that your tests can mock it:
https://github.com/prebid/Prebid.js/blob/master/modules/rubiconAnalyticsAdapter.js#L11

Then in your beforeEach you should setup the stub
https://github.com/prebid/Prebid.js/blob/master/test/spec/modules/rubiconAnalyticsAdapter_spec.js#L596-L600

Don't forget to restore it back in the afterEach
https://github.com/prebid/Prebid.js/blob/master/test/spec/modules/rubiconAnalyticsAdapter_spec.js#L629-L631

Then in your tests you can set them up to return what you want and make sure your module works as expected:
https://github.com/prebid/Prebid.js/blob/master/test/spec/modules/rubiconAnalyticsAdapter_spec.js#L1297

test/spec/modules/teadsBidAdapter_spec.js
```diff
@@ -1,13 +1,23 @@
...
 const AD_SCRIPT = '<script type="text/javascript" class="teads" async="true" src="https://a.teads.tv/hb/getAdSettings"></script>"';
 
 describe('teadsBidAdapter', () => {
   const adapter = newBidder(spec);
 
+  before(function () {
+    // Following the introduction of tests involving reading/writing cookies,
+    // this allows for running this spec as a single file with:
+    // `gulp test --file "test/spec/modules/teadsBidAdapter_spec.js"`.
+    window.$$PREBID_GLOBAL$$.processQueue();
```
robertrmartinez, line 18: With the stubbing of storage, you can get rid of this I believe.

This type of code in test files can cause downstream tests to be very flakey.

We do not want to modify the window object and its contents if we do not have to in tests.

robertrmartinez (changes requested):
Adapter looks good but would like the storage code to be stubbed in the tests.

We are trying to be more diligent about not letting people modify the window during tests, setting local storage and cookies is not reliable, especially when you have many adapters doing so.

Please see the examples I just gave and update the spec file.

Thanks, let me know if you have any questions

ruizb:
Thanks @robertrmartinez, it's so much better/stable to use a stub rather than the global environment! 👍

robertrmartinez (approved):
Awesome stuff

Thanks for being so flexible!

PR #7149, opened 2021-07-07
SharedIdSystem: return pubcid instead of sharedId

SKOCHERI:
- [x] Bugfix
This is for the issue reported in https://github.com/prebid/Prebid.js/pull/7099.
Apart from fixing the issue mentioned in https://github.com/prebid/Prebid.js/pull/7099. This PR has cleaned up the SharedIdSystem to removed calls to sharedId endpoint.

### DIFF
--- a/modules/sharedIdSystem.js
+++ b/modules/sharedIdSystem.js
@@ -63,8 +63,8 @@ function queuePixelCallback(pixelUrl, id = '', callback) {
 }
 
 function hasOptedOut() {
-  return (storage.cookiesAreEnabled() && readValue(OPTOUT_NAME, COOKIE)) ||
-    (storage.hasLocalStorage() && readValue(OPTOUT_NAME, LOCAL_STORAGE));
+  return !!((storage.cookiesAreEnabled() && readValue(OPTOUT_NAME, COOKIE)) ||
+    (storage.hasLocalStorage() && readValue(OPTOUT_NAME, LOCAL_STORAGE)));
 }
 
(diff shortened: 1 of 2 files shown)


philipwatson:
Cheers, looks good. 

I see another potential issue: it doesn't delete the stored id when `_pubcid_optout` cookie is detected.
So when the user opts-out _after_ the stored id is created, prebid will keep using this stored id until it expires.

SKOCHERI:
> Cheers, looks good.
> 
> I see another potential issue: it doesn't delete the stored id when `_pubcid_optout` cookie is detected.
> So when the user opts-out _after_ the stored id is created, prebid will keep using this stored id until it expires.

@philipwatson In extendId() function, if optedout the id (cookie value) is set to undefined.

philipwatson:
@SKOCHERI But the core user module won’t use it if undefined. Here: https://github.com/SKOCHERI/Prebid.js/blob/32837f4c5d2247fb67286b1a95b462d733004d60/modules/userId/index.js#L662

SKOCHERI:
> @SKOCHERI But the core user module won’t use it if undefined. Here: https://github.com/SKOCHERI/Prebid.js/blob/32837f4c5d2247fb67286b1a95b462d733004d60/modules/userId/index.js#L662

@philipwatson The decode() is updated to handle output

philipwatson:
Yep that works. Cheers!

ChrisHuie (approved):

PR #7151, opened 2021-07-08
bidViewablityIO Module: add new submodule for detecting viewability without ad server dependancies

jsut:
## Type of change

- [x] Feature

## Description of change

Emits a BID_VIEWABLE event for banner ads when a bid meets IAB viewable specifications, using the browsers IntersectionObserver API, if it is available.

To see this in action locally, gulp serve, then browse to the included integration example.

I haven't got a docs repo PR yet, I'll put that together soon.

## Other information

related issue: https://github.com/prebid/Prebid.js/issues/6909

### DIFF
--- a/modules/bidViewabilityIO.js
+++ b/modules/bidViewabilityIO.js
@@ -69,19 +69,19 @@ export let viewCallbackFactory = (bid) => {
 };
 
 export let init = () => {
-  const globalModuleConfig = config.getConfig(MODULE_NAME) || {};
-
-  if (globalModuleConfig[CONFIG_ENABLED]) {
-    // if the module is enabled, then listen to AD_RENDER_SUCCEEDED to setup IO's for supported
-    // mediaTypes, on browsers that support IO
-    events.on(EVENTS.AD_RENDER_SUCCEEDED, ({doc, bid, id}) => {
-      if (CLIENT_SUPPORTS_IO && isSupportedMediaType(bid)) {
-        let viewable = new IntersectionObserver(viewCallbackFactory(bid), getViewableOptions(bid));
-        let element = document.getElementById(bid.adUnitCode);
-        viewable.observe(element);
-      }
-    });
-  }
+  config.getConfig(MODULE_NAME, conf => {
+    if (conf[MODULE_NAME][CONFIG_ENABLED] && CLIENT_SUPPORTS_IO) {
+      // if the module is enabled and the browser supports Intersection Observer,
+      // then listen to AD_RENDER_SUCCEEDED to setup IO's for supported mediaTypes
+      events.on(EVENTS.AD_RENDER_SUCCEEDED, ({doc, bid, id}) => {
+        if (isSupportedMediaType(bid)) {
+          let viewable = new IntersectionObserver(viewCallbackFactory(bid), getViewableOptions(bid));
+          let element = document.getElementById(bid.adUnitCode);
+          viewable.observe(element);
+        }
+      });
+    }
+  });
 }
 
 init()
--- a/modules/bidViewabilityIO.js
+++ b/modules/bidViewabilityIO.js
@@ -1,0 +1,85 @@
+import { config } from '../src/config.js';
+import * as events from '../src/events.js';
+import { EVENTS } from '../src/constants.json';
+import { logMessage } from '../src/utils.js';
+
+const MODULE_NAME = 'bidViewabilityIO';
+const CONFIG_ENABLED = 'enabled';
+
+// IAB numbers from: https://support.google.com/admanager/answer/4524488?hl=en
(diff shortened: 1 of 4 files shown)


modules/bidViewabilityIO.js
```diff
@@ -0,0 +1,85 @@
...
+    });
+  };
+};
+
+export let init = () => {
+  events.on(EVENTS.AD_RENDER_SUCCEEDED, ({doc, bid, id}) => {
+    // read the config for the module
+    const globalModuleConfig = config.getConfig(MODULE_NAME) || {};
+    // do nothing if module-config.enabled is not set to true
+    // this way we are adding a way for bidders to know (using pbjs.getConfig('bidViewability').enabled === true) whether this module is added in build and is enabled
```
robertrmartinez, line 76: I think it would be better to never register for this event until the user enables the config.

So your init function could subscribe to setConfig and then once it is enabled, it then will add the event listener.

```js
config.getConfig(MODULE_NAME, conf => {
   // if conf enabled then 
   events.on(EVENTS.AD_RENDER_SUCCEEDED, ({doc, bid, id}) => {
});
```

That way the event callback is not executing every single time this module is included.

Only when it is actually turned on.

I would also add the `CLIENT_SUPPORTS_IO` check inside of that config subscribe as well since no need to run it when it is not supported

robertrmartinez (changes requested):

modules/bidViewabilityIO.js
jsut: i just pushed a change that doesn't do the subscribe thing, then actually read all these words and found out that this subscribe thing is a thing.  I'll have a look it doing it as you describe this afternoon probably.

robertrmartinez: Ok cool.

So basically if you do the following

```js
config.getConfig(MODULE_NAME, CALLBACK_FUNCTION);
```

Prebid will execute your `CALLBACK_FUNCTION` each time `pbjs.setConfig` is called with your `MODULE_NAME`

So for your use case, we only want to register the event listener if the user / pub runs pbjs.setConfig and turns on bidViewability.

so your callback can do something like:

```js
config.getConfig(MODULE_NAME, conf => {
   const  globalModuleConfig = conf.bidViewabilityIO;
   if (globalModuleConfig[CONFIG_ENABLED] && CLIENT_SUPPORTS_IO) {
      // register your event code
      events.on(EVENTS.AD_RENDER_SUCCEEDED, ({doc, bid, id}) => {
   }
});
```

modules/bidViewabilityIO.js
```diff
@@ -0,0 +1,87 @@
...
+
+const CLIENT_SUPPORTS_IO = window.IntersectionObserver && window.IntersectionObserverEntry && window.IntersectionObserverEntry.prototype &&
+    'intersectionRatio' in window.IntersectionObserverEntry.prototype;
+
+const supportedMediaTypes = [
+  'banner'
+];
+
+export let isSupportedMediaType = (bid) => {
+  return supportedMediaTypes.includes(bid.mediaType);
```
robertrmartinez, line 23: This is causing tests in IE to fail becuase array.includes is not supported.

Can you use indexOf instead please.

robertrmartinez (changes requested):

modules/bidViewabilityIO.js
```diff
@@ -0,0 +1,87 @@
...
+    }
+  }
+}
+
+// markViewed returns a function what will be executed when an ad satisifes the viewable iO
+export let markViewed = (bid, entry, observer) => {
+  return () => {
+    observer.unobserve(entry.target);
+    events.emit(EVENTS.BID_VIEWABLE, bid);
+    logMessage(`id: ${entry.target.getAttribute('id')} code: ${bid.adUnitCode} was viewed`);
```
robertrmartinez, line 42: This is me being picky, but when I am testing / debugging pub pages I like when modules emit their name inside their console logs, makes it easy to filter.

You could wrap logMessage in your own function which just adds in MODULE_NAME on each message you send.

jsut: That doesn't sound picky, it sounds like an excellent practice.

robertrmartinez (approved):

robertrmartinez:
I think we need a docs pr before merging this

jsut:
yup, i'll get on that tomorrow @robertrmartinez

jsut:
docs PR: https://github.com/prebid/prebid.github.io/pull/3184

robertrmartinez:
Nice thanks @jsut 

@ChrisHuie Would you mind giving this a second look? Looks good to me but would love a second pair of eyes before merging this week!

ChrisHuie (approved):

ChrisHuie:
LGTM. Thanks @jsut for this awesome pr 🙏

PR #7167, opened 2021-07-12
Sublime Bid Adapter : send notid in pixel & use timeout from timeoutData

o3LL:
## Type of change
- [x] Feature

## Description of change

- Replace trId with notid (notifyId) in our pixels
- Use timeoutData to retrieve timeout value
- Use bidRequest to retrieve timeout value
- Updating tests
- Bumping version to 0.7.3

Contact email of the adapter’s maintainer : pbjs@sublime.xyz

### DIFF
--- a/modules/sublimeBidAdapter.js
+++ b/modules/sublimeBidAdapter.js
@@ -40,7 +40,8 @@ export function log(msg, obj) {
 export const state = {
   zoneId: '',
   transactionId: '',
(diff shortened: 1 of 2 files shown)


o3LL:
It seems like test are failing due to flakiness ?
Let us know if there is something to do on our side 👍

msm0504 (approved):

Issue #6923, opened 2021-06-02
PrebidServer Bid Adapter metadata support


bretg:
## Type of issue
enhancement

## Description

The PBJS side of https://github.com/prebid/prebid-server/issues/1871

Prebid Server bid adapters place metadata in seatbid.bid.ext.prebid.meta.ATTR. However, currently the pbsBidAdapter is only looking for dchain, and in a different path: seatbid.bid.ext.dchain

## Proposal

1) Update pbsBidAdapter to deep copy everything in seatbid.bid.ext.prebid.meta to PBJS' bidObject.meta
2) Remove the seatbid.bid.ext.dchain logic
3) Keep the current bid.adomain logic

PR #7179, opened 2021-07-14
TheMediaGrid: added ext.bidder.grid.demandSource processing

TheMediaGrid:
## Type of change

- [x] Feature

## Description of change

Read ext.bidder.grid.demandSource from the ad response and write it as hb_ds targeting

- grid-tech@themediagrid.com


ChrisHuie (approved):

patmmccann:
This isn't the proper way to do this, you should be copying demand source to bid.meta. I am putting in the revert PR.

modules/gridBidAdapter.js
```diff
@@ -362,6 +362,10 @@ function _addBidResponse(serverBid, bidRequest, bidResponses) {
         dealId: serverBid.dealid
       };
 
+      if (serverBid.ext && serverBid.ext.bidder && serverBid.ext.bidder.grid && serverBid.ext.bidder.grid.demandSource) {
```
patmmccann, line 365: please copy demandSource to bid.meta.networkName . see https://github.com/prebid/prebid.github.io/pull/1911/files

patmmccann: Please also duplicate this field server side; so that it will flow through regardless of if you are client or server post #6923

patmmccann: #7184 should take care of client side

PR #7184, opened 2021-07-15
MediaGrid Adapter: Put demandSource in bid.meta

patmmccann:
## Type of change

- [ ] Bugfix
- [ ] Feature
- [ ] New bidder adapter   
- [ ] Build related changes
- [ ] CI related changes
- [] Other

## Description of change
Puts demandSource in the correct place: see
https://github.com/prebid/prebid.github.io/pull/1911/files

### DIFF
--- a/modules/gridBidAdapter.js
+++ b/modules/gridBidAdapter.js
@@ -358,13 +358,13 @@ function _addBidResponse(serverBid, bidRequest, bidResponses) {
         ttl: TIME_TO_LIVE,
         meta: {
           advertiserDomains: serverBid.adomain ? serverBid.adomain : [],
-          networkName: serverBid.ext && serverBid.ext.bidder && serverBid.ext.bidder.grid && serverBid.ext.bidder.grid.demandSource ? serverBid.ext.bidder.grid.demandSource : ''
         },
         dealId: serverBid.dealid
       };
 
       if (serverBid.ext && serverBid.ext.bidder && serverBid.ext.bidder.grid && serverBid.ext.bidder.grid.demandSource) {
         bidResponse.adserverTargeting = { 'hb_ds': serverBid.ext.bidder.grid.demandSource };
+        bidResponse.meta.networkName = serverBid.ext.bidder.grid.demandSource;
       }
 
       if (serverBid.content_type === 'video') {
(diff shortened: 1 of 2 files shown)


patmmccann:
Extends #7179

modules/gridBidAdapter.js
```diff
@@ -357,7 +357,8 @@ function _addBidResponse(serverBid, bidRequest, bidResponses) {
         netRevenue: true,
         ttl: TIME_TO_LIVE,
         meta: {
-          advertiserDomains: serverBid.adomain ? serverBid.adomain : []
+          advertiserDomains: serverBid.adomain ? serverBid.adomain : [],
+          networkName: serverBid.ext && serverBid.ext.bidder && serverBid.ext.bidder.grid && serverBid.ext.bidder.grid.demandSource ? serverBid.ext.bidder.grid.demandSource : ''
```
ncolletti, line 361: Any reason this isn't also called demandSource? Also, should this default to 'grid' ?

May make sense to not send the property if 'demandSource' doesn't exist like this:
`...(serverBid.ext && serverBid.ext.bidder && serverBid.ext.bidder.grid && serverBid.ext.bidder.grid.demandSource && {demandSource: serverBid.ext.bidder.grid.demandSource),`

ncolletti:
Could you add a test to cover this and update existing tests to include the new property

modules/gridBidAdapter.js
TheMediaGrid: tagging @patmmccann for visibility
I'm assuming networkName vs demandSource is a CafeMedia requirement? we'd rather keep it similar to the actual bid response param though (i.e. demandSource)

as for the default value, right, it should be set to 'grid'

patmmccann: Network name is documented here https://github.com/prebid/prebid.github.io/pull/1911

Demandsource is not

TheMediaGrid: @patmmccann I suggest using demandSource instead as it's actually breaking down the network into separate demand sources so we should make sure we don't block the publishers and bidders from using both, what do you think?

patmmccann: this works for me, i'll do a documentation change as well in that case

patmmccann:
docs pull: https://github.com/prebid/Prebid.js/pull/7184

ncolletti (approved):

PR #7189, opened 2021-07-16
Fledge Module: Initial Commit

iamnewton:
## Type of change

- [x] Feature

## Description of change

Google is working with the W3C on [a proposal](https://github.com/WICG/turtledove/blob/main/FLEDGE.md) to support advertising in a 3rd-party cookieless world.  This feature adds [a polyfill](https://github.com/MagniteEngineering/fledge.polyfill/) to Prebid to support that proposal in order to test out the viability of Google's proposal. 

The goal of this is to find out who well a Fledge auction will work against a Prebid auction in order to provide some real-world feedback to Google in hopes to help shape the future of advertising.

This is a working draft, and so I know there are no tests, but I'd like to get community feedback regarding this feature and how best to implement it before I dive too deep in.

### DIFF
--- a/modules/fledge.js
+++ b/modules/fledge.js
@@ -22,16 +22,11 @@ let _fledgeConfig = {};
  */
 export function renderAdHook(fn, doc, id, options) {
   const bid = auctionManager.findBidByAdId(id);
-  console.log({bid});
 
-  console.log({_fledgeConfig});
   if (_fledgeConfig.auctionConfig) {
-    const fledge = new Fledge();
-    const auctionConfig = {
-      ..._fledgeConfig.auctionConfig,
-      ...bid,
-    };
-    console.log({fledge});
+    const fledge = new Fledge('https://magniteengineering.github.io/fledge.polyfill/iframe.html');
+    const auctionConfig = _fledgeConfig.auctionConfig;
+    auctionConfig.auctionSignals.winningContextualBid = bid;
     fledge
       .runAdAuction(auctionConfig)
       .then(results => {
@@ -39,7 +34,7 @@ export function renderAdHook(fn, doc, id, options) {
           // create an iframe within the Fledge auction iframe, render the winning ad
           const ad = document.createElement('iframe');
           ad.src = results;
-          document.getElementById('ad-slot-1').appendChild(ad);
+          document.getElementById(bid.adUnitCode).appendChild(ad);
         } else {
           return fn.call(this, doc, id, options);
         }
@@ -55,6 +50,7 @@ export function renderAdHook(fn, doc, id, options) {
 export function handleSetFledgeConfig(config) {
   _fledgeConfig = utils.pick(config, [
     'enabled', enabled => enabled !== false, // defaults to true
+    'auctionConfig',
   ]);
 
   // if enabled then do some stuff
--- a/modules/fledge.js
+++ b/modules/fledge.js
@@ -1,0 +1,79 @@
+import { Fledge } from '@magnite/fledge.polyfill/dist/api/cjs/index.js';
+import { auctionManager } from '../src/auctionManager.js';
+import { config } from '../src/config.js';
+import CONSTANTS from '../src/constants.json';
+import { getGlobal } from '../src/prebidGlobal.js';
+import * as utils from '../src/utils.js';
+
+/**
+ * @summary This Module is intended to provide users with the ability to run a Fledge-based auction against a PreBid winning bid
+ */
+const MODULE_NAME = 'Fledge';
+
+/**
+ * @summary The config to be used. Can be updated via: setConfig
+ */
+let _fledgeConfig = {};
+
+/**
+ *
+ * @param {Object} reqBidsConfigObj required; This is the same param that's used in pbjs.requestBids.
+ * @param {function} fn required; The next function in the chain, used by hook.js
+ */
+export function renderAdHook(fn, doc, id, options) {
+  const bid = auctionManager.findBidByAdId(id);
+  console.log({bid});
+
+  console.log({_fledgeConfig});
+  if (_fledgeConfig.auctionConfig) {
+    const fledge = new Fledge();
+    const auctionConfig = {
+      ..._fledgeConfig.auctionConfig,
+      ...bid,
+    };
+    console.log({fledge});
+    fledge
+      .runAdAuction(auctionConfig)
+      .then(results => {
+        if (results) {
+          // create an iframe within the Fledge auction iframe, render the winning ad
+          const ad = document.createElement('iframe');
+          ad.src = results;
+          document.getElementById('ad-slot-1').appendChild(ad);
+        } else {
+          return fn.call(this, doc, id, options);
+        }
+      });
+  } else {
+    return fn.call(this, doc, id, options);
+  }
+}
+
+/**
+ * @summary This is the function which controls what happens during a pbjs.setConfig({...floors: {}}) is called
+ */
+export function handleSetFledgeConfig(config) {
+  _fledgeConfig = utils.pick(config, [
+    'enabled', enabled => enabled !== false, // defaults to true
+  ]);
+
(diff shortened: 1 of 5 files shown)


modules/fledge.js
```diff
@@ -0,0 +1,75 @@
...
+/**
+ *
+ * @param {Object} reqBidsConfigObj required; This is the same param that's used in pbjs.requestBids.
+ * @param {function} fn required; The next function in the chain, used by hook.js
+ */
+export function renderAdHook(fn, doc, id, options) {
+  const bid = auctionManager.findBidByAdId(id);
+
+  if (_fledgeConfig.auctionConfig) {
+    const fledge = new Fledge('https://magniteengineering.github.io/fledge.polyfill/iframe.html');
```
patmmccann, line 27: Why is external code needed here? Also, please lock https://magniteengineering.github.io/fledge.polyfill/scripts/frame/esm/index.js to a version if external code is actually required. Also, since it requires promises / async await, please make sure the browser can handle these features before adding these requests.

package-lock.json
```diff
@@ -1412,11 +1412,27 @@
...
-        "merge-source-map": "1.1.0",
-        "schema-utils": "2.7.1"
+        "convert-source-map": "^1.7.0",
+        "istanbul-lib-instrument": "^4.0.3",
+        "loader-utils": "^2.0.0",
+        "merge-source-map": "^1.1.0",
+        "schema-utils": "^2.7.0"
+      }
+    },
+    "@magnite/fledge.polyfill": {
```
patmmccann, line 1422: the vast majority of your changes to this file are not germane to this pr

package-lock.json
```diff
@@ -1412,11 +1412,27 @@
...
+      "version": "0.3.2",
+      "resolved": "https://registry.npmjs.org/@magnite/fledge.polyfill/-/fledge.polyfill-0.3.2.tgz",
```
patmmccann, line 1424: appears to have some redundant functionality to https://magniteengineering.github.io/fledge.polyfill/scripts/frame/esm/index.js  ; both contain a runAdAuction method

modules/fledge.js
```diff
@@ -0,0 +1,75 @@
...
+    getGlobal()
+      .renderAd
+      .getHooks({
+        hook: renderAdHook
+      })
+      .remove();
+  }
+}
+
+config.getConfig('fledge', config => handleSetFledgeConfig(config.fledge));
```
patmmccann, line 75: no spec file on this pr, please add one.

patmmccann (changes requested):

package-lock.json
iamnewton: what do you mean?

package-lock.json
iamnewton: they are the same; the URL you've provided is the hosted version of this for testing purposes

modules/fledge.js
iamnewton: > Why is external code needed here?

The external code is a polyfill that is required to work with Google's new Fledge proposal.

> Also, please lock https://magniteengineering.github.io/fledge.polyfill/scripts/frame/esm/index.js to a version if external code is actually required.

This is locked down in the `package.json` to a version (https://github.com/prebid/Prebid.js/pull/7189/files#diff-7ae45ad102eab3b6d7e7896acd08c427a9b25b346470d7bc6507b6481575d519R105)

> Also, since it requires promises / async await, please make sure the browser can handle these features before adding these requests.

Do you have any suggestions for how to handle this?  Do you guys have/use some sort of agent/browser sniffing that I could work with in order to prevent the code from running in older browsers?  Technically this piece of code will only work with Chrome at version 91 or later.

package-lock.json
patmmccann: there are hundreds of changes that have nothing to do with this pr

package-lock.json
patmmccann: there are hundreds of changes to package.lock that are unrelated to this pr

modules/fledge.js
patmmccann: I am confused, if you don't need the runtime code bc you included it at build time, why are you invoking it. 

I would suggest not sniffing for user agents or browser versions but just checking for the existence of the required api or functionality

package-lock.json
iamnewton: I installed a dependency and it looks like the `package-lock.json` updated mostly versions of babel processors

modules/fledge.js
iamnewton: > I am confused, if you don't need the runtime code bc you included it at build time, why are you invoking it.

I'm also confused by the concern and questions here.  


I included the `@magnite/fledge.polyfill` library because it's a polyfill that is required to work with Fledge.

> I would suggest not sniffing for user agents or browser versions but just checking for the existence of the required api or functionality

okay, thanks.

package-lock.json
patmmccann: I see you haven't fixed this; is your goal here to get a merge or something else? If you're not looking for a merge can you change to draft?

package-lock.json
patmmccann: I am not clear based on this if your goal is a merge with this pr or something else. We're not going to merge code you say is redundant.

fabienbk:
Hello  @iamnewton,
Great initiative imho. Just a little thing regarding the polyfill: it doesnt seem to check if navigator.joinAdInterestGroup & navigator.runAdAuction are present before instanciating itself - so your module will always use it, even on Canary versions where Fledge is already available.

vincent-grosbois:
Hello @iamnewton 
I'm wondering what's the rationale behind using polyfill for fledge, and not directly calling the "true" fledge API ? This API doesn't exist in Chrome yet but it's available on Chrome Canary, as long as the correct flags are set by the user. It seems it would also make some sense to directly plug the "true" Fledge API in prebid, rather than going through polyfill ?

modules/fledge.js
iamnewton: yeah, this is new stuff that doesn't exist yet

jdwieland8282: That is intentional @ChrisHuie Chrome 91 will be the version of Chrome that supports Fledge. We have to match this "module" with requests from the right Chrome browser version.

package-lock.json
iamnewton: I'm not sure I understand your question or concern here

package-lock.json
iamnewton: the goal is to get this merged; you're reviewing the `package-lock` changes which are a result of a required dependency that was installed and affected the `package-lock`

iamnewton:
> Great initiative imho. Just a little thing regarding the polyfill: it doesn't seem to check if `navigator.joinAdInterestGroup` & `navigator.runAdAuction` are present before instantiating itself - so your module will always use it, even on Canary versions where Fledge is already available.

@fabienbk I can certainly add that, but those methods won't exist and we'll likely need to sniff user agent strings for usage

modules/fledge.js
iamnewton: well Chrome 94 is out now, so it will be much later at this point

jdwieland8282: right I guess the pt is that this module is intended to only work on a very specific version of Chrome, TBD on which version.

vincent-grosbois:
> > Great initiative imho. Just a little thing regarding the polyfill: it doesn't seem to check if `navigator.joinAdInterestGroup` & `navigator.runAdAuction` are present before instantiating itself - so your module will always use it, even on Canary versions where Fledge is already available.
> 
> thanks @fabienbk I can certainly add that, but those methods won't exist and we'll likely need to sniff user agent strings for usage

IIRC you can just check if the `navigator.joinAdInterestGroup` is null or not, it will tell you if the feature is available or no

ChrisHuie:
Closing this pr since 7731 is an iteration on this previous pr

PR #7202, opened 2021-07-20
OpenX Bid Adapter: read first party data segments

laurb9:
## Type of change

- [x] Feature

## Description of change

1. Read fpd segment data from ortb2.user.data and send out in `sm` query parameter. Intended for permutive but others will also work.
2. Read [liveintent segments](https://docs.prebid.org/dev-docs/modules/userId.html#liveintent-nonid) from request `userId.lipb.segments` and send in the same `sm` query parameter.
3. (tests refactor) Moved bid request samples higher so they can be reused in the new tests without duplicating them again.

### DIFF
--- a/modules/openxBidAdapter.js
+++ b/modules/openxBidAdapter.js
@@ -318,6 +318,10 @@ function appendUserIdsToQueryParams(queryParams, userIds) {
           break;
         case 'lipb':
           queryParams[key] = userIdObjectOrValue.lipbid;
+          if (Array.isArray(userIdObjectOrValue.segments) && userIdObjectOrValue.segments.length > 0) {
+            const liveIntentSegments = 'liveintent:' + userIdObjectOrValue.segments.join('|')
+            queryParams.sm = `${queryParams.sm ? queryParams.sm + encodeURIComponent(',') : ''}${encodeURIComponent(liveIntentSegments)}`;
+          }
           break;
         case 'parrableId':
           queryParams[key] = userIdObjectOrValue.eid;
--- a/modules/openxBidAdapter.js
+++ b/modules/openxBidAdapter.js
@@ -250,15 +250,22 @@ function buildCommonQueryParamsFromBids(bids, bidderRequest) {
 
   const firstPartyData = config.getConfig('ortb2.user.data')
   if (Array.isArray(firstPartyData) && firstPartyData.length > 0) {
-    const sm = firstPartyData
+    // extract and merge valid segments by provider/taxonomy
+    const fpd = firstPartyData
       .filter(
         data => (Array.isArray(data.segment) &&
                 data.segment.length > 0 &&
                 data.name !== undefined &&
                 data.name.length > 0)
       )
+      .reduce((acc, data) => {
+        const name = typeof data.ext === 'object' && data.ext.segtax ? `${data.name}/${data.ext.segtax}` : data.name;
+        acc[name] = (acc[name] || []).concat(data.segment.map(seg => seg.id));
+        return acc;
+      }, {})
+    const sm = Object.entries(fpd)
       .map(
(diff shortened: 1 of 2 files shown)


modules/openxBidAdapter.js
```diff
@@ -248,6 +248,24 @@ function buildCommonQueryParamsFromBids(bids, bidderRequest) {
...
+  if (Array.isArray(firstPartyData) && firstPartyData.length > 0) {
+    const sm = firstPartyData
+      .filter(
+        data => (Array.isArray(data.segment) &&
+                data.segment.length > 0 &&
+                data.name !== undefined &&
+                data.name.length > 0)
+      )
+      .map(
+        data => data.name + ':' + data.segment.map(seg => seg.id).join('|')
```
patmmccann, line 261: you appear to drop segtax, which is critical to understand the segment id.

modules/openxBidAdapter.js
```diff
@@ -300,6 +318,10 @@ function appendUserIdsToQueryParams(queryParams, userIds) {
           break;
         case 'lipb':
           queryParams[key] = userIdObjectOrValue.lipbid;
+          if (Array.isArray(userIdObjectOrValue.segments) && userIdObjectOrValue.segments.length > 0) {
+            const liveIntentSegments = 'liveintent:' + userIdObjectOrValue.segments.join('|')
```
patmmccann, line 322: does it make more sense to modify the liveintent adapter to write their segments to ortb2 than to modify your adapter to read them from here?

patmmccann (changes requested):

modules/openxBidAdapter.js
laurb9: Thank you, the provider we were interested in (permutive) appears to generate proprietary segments but since the code is generic I added handling for segtax as well.

modules/openxBidAdapter.js
laurb9: I agree, it certainly does, but modifying the liveintent adapter is out of my scope for this task and I don't feel confident I can do it properly at this time, I just started working on Prebid.js. We can perhaps ask them to consider making this change.

modules/openxBidAdapter.js
```diff
@@ -248,6 +248,31 @@ function buildCommonQueryParamsFromBids(bids, bidderRequest) {
...
+                data.segment.length > 0 &&
+                data.name !== undefined &&
+                data.name.length > 0)
+      )
+      .reduce((acc, data) => {
+        const name = typeof data.ext === 'object' && data.ext.segtax ? `${data.name}/${data.ext.segtax}` : data.name;
+        acc[name] = (acc[name] || []).concat(data.segment.map(seg => seg.id));
+        return acc;
+      }, {})
+    const sm = Object.entries(fpd)
```
Fawke, line 266: The tests are failing in IE11 because Object.entries is not supported. Can you please use a polyfill. One way you can solve it by using a polyfill from `core-js-pure/features/object/entries`. If you use this, remember to add it to `allowedModules`

Another way could be you use a polyfill in your adapter instead of importing the one above.

Fawke (changes requested):
Looks good overall but some tests for IE are failing.

@patmmccann Do you wanna have a look at your previous comment since you Requested Changes.

modules/openxBidAdapter.js
laurb9: Thank you, I will rewrite it using `Object.keys()` if IE9 support is not required.

Fawke (approved):
LGTM

modules/openxBidAdapter.js
patmmccann: @melgenek fyi

Please feel free to reach out to us on Prebid slack or join a taxonomy committee meeting and we can discuss how you can make it so all adapters can easily absorb your segments.

PR #7218, opened 2021-07-26
Malltv Analytics Adapter: add new analytics adapter

ardit-baloku:
## Type of change

- [ ] Bugfix
- [x] Feature
- [x] New bidder adapter   
- [ ] Build related changes
- [ ] CI related changes
- [ ] Other

## Description of change

Added new analytics adapter

- parameters for enabling analytic adapter
```
{
  provider: 'malltvAnalytics',
  options: {
    'propertyId': 'YOUR_PROPERTY_ID', // Required
    'server': 'YOUR_ANALYTICS_SERVER' // Optional
  }
}
```

- Maintainer email: arditb@gjirafa.com

- A link to a PR on the docs repo: https://github.com/prebid/prebid.github.io/pull/3147

### DIFF
--- a/modules/malltvAnalyticsAdapter.js
+++ b/modules/malltvAnalyticsAdapter.js
@@ -26,9 +26,11 @@ export const BIDDER_STATUS = {
 export const getCpmInEur = function (bid) {
   if (bid.currency === 'EUR') {
     return bid.cpm
-  } else {
+  } else if (typeof bid.getCpmInNewCurrency === 'function') {
     return bid.getCpmInNewCurrency('EUR')
   }
+
+  return bid.cpm
 }
 
 const analyticsOptions = {}
--- a/modules/malltvAnalyticsAdapter.js
+++ b/modules/malltvAnalyticsAdapter.js
@@ -7,12 +7,11 @@ import {logInfo, logError, deepClone} from '../src/utils.js'
 
 const analyticsType = 'endpoint'
 export const ANALYTICS_VERSION = '1.0.0'
-export const DEFAULT_SERVER = 'http://localhost:51524/analytics'
+export const DEFAULT_SERVER = 'https://central.mall.tv/analytics'
 
 const {
   EVENTS: {
     AUCTION_END,
-    AD_RENDER_FAILED,
     BID_TIMEOUT
   }
 } = CONSTANTS
@@ -67,9 +66,6 @@ export const malltvAnalyticsAdapter = Object.assign(adapter({DEFAULT_SERVER, ana
   },
   track({eventType, args}) {
     switch (eventType) {
-      case AD_RENDER_FAILED:
-        this.handleAdRenderFailed(args)
-        break
       case BID_TIMEOUT:
         this.handleBidTimeout(args)
         break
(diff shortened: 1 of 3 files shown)


modules/malltvAnalyticsAdapter.js
```diff
@@ -0,0 +1,188 @@
...
+malltvAnalyticsAdapter.originEnableAnalytics = malltvAnalyticsAdapter.enableAnalytics
+
+// override enableAnalytics so we can get access to the config passed in from the page
+malltvAnalyticsAdapter.enableAnalytics = function (config) {
+  if (this.initConfig(config)) {
+    malltvAnalyticsAdapter.originEnableAnalytics(config) // call the base class function
+  }
+}
+
+adapterManager.registerAnalyticsAdapter({
```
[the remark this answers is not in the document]
ardit-baloku: Correct, there is no GVLID

mmoschovas (approved):
lgtm

modules/malltvAnalyticsAdapter.js
```diff
@@ -0,0 +1,188 @@
...
+  NO_BID: 2,
+  BID_WON: 3,
+  TIMEOUT: 4
+}
+
+export const getCpmInEur = function (bid) {
+  if (bid.currency === 'EUR') {
+    return bid.cpm
+  } else {
+    return bid.getCpmInNewCurrency('EUR')
```
mmoschovas, line 30: Actually, you need to add a condition here in the instance that getCpmInNewCurrency does not exist. Please check that it exists as a function before running

mmoschovas (changes requested):

modules/malltvAnalyticsAdapter.js
mmoschovas: Also updating here should correct your unit tests and allow them to pass

modules/malltvAnalyticsAdapter.js
```diff
@@ -0,0 +1,190 @@
...
+} = CONSTANTS
+
+export const BIDDER_STATUS = {
+  BID: 1,
+  NO_BID: 2,
+  BID_WON: 3,
+  TIMEOUT: 4
+}
+
+export const getCpmInEur = function (bid) {
```
mmoschovas, line 26: @ardit-baloku 

Since this is now checking for getCpmInNewCurrency and the default is to return bid.cpm as is, do we need the initial condition? Seems like it can simply be 

```
export const getCpmInEur = function (bid) {
  if (typeof bid.getCpmInNewCurrency === 'function') {
    return bid.getCpmInNewCurrency('EUR')
  }

  return bid.cpm
}
```

ardit-baloku: Doesn't seem necessary to call getCpmInNewCurrency if it is already in EUR

mmoschovas: Makes sense but this can still be modified just to remove the repeated return statement 

```
export const getCpmInEur = function (bid) {
  if (bid.currency !== 'EUR' && typeof bid.getCpmInNewCurrency === 'function') {
     return bid.getCpmInNewCurrency('EUR');
  }
  
  return bid.cpm;
```

ardit-baloku:
@mmoschovas I was running only my tests while writing the code and they seem to crash when run with other tests. They are working as expected now

mmoschovas (approved):
lgtm

Issue #7221, opened 2021-07-26
BetweenBidAdapter sharedid support was accidently(?) removed in Prebid 5.0?


egsk:
## Type of issue

Question
## Description

Hello! We have noticed that one of the last of our pull requests (see https://github.com/prebid/Prebid.js/pull/6531) was excluded from latest release of Prebid (see https://github.com/prebid/Prebid.js/commit/412277f72b97cab72644cb4a009f58067a165899#diff-233e78fe3ed7377b70d8f5dce715c669d72e08e90006af8b33151d35ab1c239f), but description of the realese contains opposite information:

> * BetweenBidAdatper: added sharedid support (#6531)

Was it removed by a reason, or by mistake? What should we do to return sharedid support to BetweenBidAdapter?

ChrisHuie:
@egsk I believe this is related to issue #6640

patmmccann:
discussed in a meeting today, tacked down to @NothingIsMatter commit #6808

PR #7222, opened 2021-07-27
Between Bid Adapter: add sharedid for Prebid 5.0

ChrisHuie:
This fixes an accidental deletion when moving to Prebid 5. Assumed all adapters had bothh SharedId and PubcommonID. Fix for issue #7221


patmmccann (approved):

patmmccann:
Was ucfunnel also affected?

PR #7227, opened 2021-07-28
Adyoulike bidder - Video playerSize format

guiann:
## Type of change
Minor improvement.
make sure the video size format is consistent.

## Description of change
The player size is sent with Array of array's like multiple size are needed even if only one size is necessary. 
This improvement makes sure this behavior won't change and break our ad server video response.

- [*] official adapter submission

please see: 
 https://github.com/prebid/Prebid.js/issues/7217e


ChrisHuie (approved):

guiann:
@ChrisHuie 
Thank you,
can you please have a look at this one too:
https://github.com/prebid/Prebid.js/pull/7229

PR #7185, opened 2021-07-15
Conversant Bid Adapter: add getUserSync

johnwier:
## Type of change
- [X ] Feature

## Description of change

Update the Conversant adapter to to use getUserSyncs to send sync pixels

PR #7228, opened 2021-07-28
Conversant Bid Adapter: add getUserSync

ChrisHuie:
Creating because pr #7185 didn't kick off circleci tests and the contributor doesn't have access to the fork to unfollow it.

### DIFF
--- a/modules/conversantBidAdapter.js
+++ b/modules/conversantBidAdapter.js
@@ -1,7 +1,7 @@
 import * as utils from '../src/utils.js';
 import {registerBidder} from '../src/adapters/bidderFactory.js';
-import { BANNER, VIDEO } from '../src/mediaTypes.js';
-import { getStorageManager } from '../src/storageManager.js';
+import {BANNER, VIDEO} from '../src/mediaTypes.js';
+import {getStorageManager} from '../src/storageManager.js';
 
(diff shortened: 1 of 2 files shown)


ChrisHuie:
@johnwier looks like the flatmap is causing an issue with this pr

modules/conversantBidAdapter.js
```diff
@@ -248,6 +248,46 @@ export const spec = {
...
+
+    if (responses && responses.ext) {
+      const pixels = [{urls: responses.ext.fsyncs, type: 'iframe'}, {urls: responses.ext.psyncs, type: 'image'}]
+        .filter((entry) => {
+          return entry.urls &&
+            ((entry.type === 'iframe' && syncOptions.iframeEnabled) ||
+            (entry.type === 'image' && syncOptions.pixelEnabled));
+        })
+        .flatMap((entry) => {
+          return entry.urls.flatMap((endpoint) => {
```
ncolletti, line 279: This seems to be one too many flatMaps. Tests will pass when removing line 279 and 286. Also updating from endpoint to entry.urls on line 280.

patmmccann: @johnwier fyi

PR #7237, opened 2021-07-29
Prebid Core: Add readConfig functionality to clone the config instead of referencing it

lksharma:
## Type of change

- [x] Bugfix
- [ ] Feature
- [ ] New bidder adapter   
- [ ] Build related changes
- [ ] CI related changes
- [ ] Other

## Description of change

Related to issue: https://github.com/prebid/Prebid.js/issues/7238
When adapters retrieve the config object by calling `config.getConfig()`, they are also able to manipulate the underlying object.

This is because the `_getConfig()` function uses `Object.assign()` when returning the configuration.

Since `Object. assign()` copies an object's reference value, we can get scenarios where one adapter updates the config and another adapter is affected. 

For example:

Step 1: Config set by `setConfig()` on the page

```
pbjs.setConfig({
    ortb2: {
        user: {
            gender: “M”
        }
    }
});
```

Step 2: Another adapter updates the config by first calling `config.getConfig()`

```
var fpd = getConfig(‘ortb2’)
fpd.user.gender = “F”
```

Step 3: Downstream adapters now retrieve the manipulated config object
```
var fpd = getConfig(‘ortb2’)
console.log(fpd)
// Output: {ortb2: { user: {gender: “F”}}}
```

```
{
  bidder: '<bidder name>',
  params: {
    // ...
  }
}
```
Currently, Sovrn adapter retrieves the `fpd` by calling getConfig(‘ortb2’), and using that `fpd` object the updating ortb2.user.ext for their request.

https://github.com/prebid/Prebid.js/blob/master/modules/sovrnBidAdapter.js#L86-L118

In the case where Sovrn is loaded before Index Exchange, the IX adapter will see the changes made by Sovrn

## Test Page
- Go to https://bastello.com/ix/prebid-fpd2.php
- Open the console in the dev tools

## Steps to Reproduce

1. Verify the config object set by pbjs.setConfig() 
   - You can look at the page source for this (L49-L203)
2. Refresh the page and call the following function in the console
   - pbjs.getConfig('ortb2')
3. Verify the returned object is not the same as the one being set at the page level.
   - You will see the `eids` & `tpid` being added to the configuration by the sovrn adapter.

### DIFF
--- a/src/config.js
+++ b/src/config.js
@@ -312,7 +312,27 @@ export function newConfig() {
         return memo;
       }, {});
     }
-    return deepClone(config);
+    return Object.assign({}, config);
+  }
+
+  /*
+   * Returns a clone of the configuration object if called without parameters,
+   * or single configuration property if given a string matching a configuration
+   * property name.  Allows deep access e.g. getConfig('currency.adServerCurrency')
+   *
+   * If called with callback parameter, or a string and a callback parameter,
+   * subscribes to configuration updates. See `subscribe` function for usage.
+   *
+   * The object returned is a deepClone of the `config` property.
+   */
+  function readConfig(...args) {
+    if (args.length <= 1 && typeof args[0] !== 'function') {
+      const option = args[0];
+      const configClone = deepClone(_getConfig());
+      return option ? utils.deepAccess(configClone, option) : configClone;
+    }
+
+    return subscribe(...args);
   }
 
   /*
@@ -629,6 +649,7 @@ export function newConfig() {
     getCurrentBidder,
     resetBidder,
     getConfig,
+    readConfig,
     setConfig,
     setDefaults,
     resetConfig,
--- a/src/config.js
+++ b/src/config.js
@@ -316,7 +316,7 @@ export function newConfig() {
   }
 
   /*
-   * Returns a clone of the configuration object if called without parameters,
+   * Returns the configuration object if called without parameters,
    * or single configuration property if given a string matching a configuration
    * property name.  Allows deep access e.g. getConfig('currency.adServerCurrency')
    *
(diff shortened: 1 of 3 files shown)


src/config.js
```diff
@@ -312,7 +312,7 @@ export function newConfig() {
         return memo;
       }, {});
     }
-    return Object.assign({}, config);
+    return deepClone(config);
```
grossjo, line 315: This may not be backwards compatible, should we create either a param to allow to return Object.assign or a separate function so behaviour is explicit in this case?

patmmccann:
This is a great contribution but we think it is breaking for potentially many publishers who use getConfig and perhaps rely on the way it is currently implemented. We hope to put a deprecation warning up and then eventually change its behavior. A 'readConfig' function could implement the deepClone and perhaps its use could be encouraged. A simple fix to the sovrn adapter would be create the readConfig function on this pr and then convert Sovrn to use it instead of changing the getConfig behavior in a breaking way.

patmmccann:
@lksharma would you be interested in modifying to make a new function readConfig have this behavior?

lksharma:
> readConfig

lksharma:
@patmmccann That makes sense, I can update the PR to make use of a new `readConfig` function.

patmmccann:
This looks really strong, but it is still in draft. Is it ready for review?

lksharma:
@patmmccann Thanks for following up, this PR is ready for review.

patmmccann:
@ChrisHuie this is not a breaking change

patmmccann:
@grossjo could you take another look?

grossjo:
@patmmccann looks good to me

patmmccann (approved):

PR #7239, opened 2021-07-30
Intimate Merger Universal Identifier System: add imuid submodule

eknis:
## Type of change

- [ ] Bugfix
- [x] Feature
- [ ] New bidder adapter   
- [ ] Build related changes
- [ ] CI related changes
- [ ] Other

## Description of change

- new user id module
```
pbjs.setConfig({
    userSync: {
        userIds: [{
            name: 'imuid',
            param: {
              cid: 5126
            }
        }]
    }
});
```

- Docs https://github.com/prebid/prebid.github.io/pull/3159

### DIFF
--- a/modules/imuIdSystem.js
+++ b/modules/imuIdSystem.js
@@ -12,17 +12,18 @@ import { getStorageManager } from '../src/storageManager.js';
 
 export const storage = getStorageManager();
 
-const storageKey = '__im_uid';
-const cookieKey = '_im_vid';
+export const storageKey = '__im_uid';
+export const cookieKey = '_im_vid';
+export const apiUrl = 'https://audiencedata.im-apps.net/imuid/get';
 const storageMaxAge = 1800000; // 30 minites (30 * 60 * 1000)
 const cookiesMaxAge = 97200000000; // 37 months ((365 * 3 + 30) * 24 * 60 * 60 * 1000)
 
-function setImDataInLocalStorage(value) {
+export function setImDataInLocalStorage(value) {
   storage.setDataInLocalStorage(storageKey, value);
   storage.setDataInLocalStorage(`${storageKey}_mt`, new Date(utils.timestamp()).toUTCString());
 }
 
-function removeImDataFromLocalStorage() {
+export function removeImDataFromLocalStorage() {
   storage.removeDataFromLocalStorage(storageKey);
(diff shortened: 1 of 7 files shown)


eknis:
Circle CI failed:
I think it is occurring in a part that has nothing to do with this addition.

Can I try rerun?

eknis:
Circle CI failed
I think it is occurring in a part that has nothing to do with this addition.

Can I try retry?

eknis:
Circle CI failed
I think it is occurring in a part that has nothing to do with this addition.

Can I try rerun?

ChrisHuie:
@eknis rerun and tests are good 👍

eknis:
@ChrisHuie thank you 👍

modules/imuIdSystem.js
```diff
@@ -0,0 +1,143 @@
...
+    if (configParams.url) {
+      syncUrl = `${configParams.url}?cid=${configParams.cid}`;
+    }
+    if (localData.vid) {
+      syncUrl += `&vid=${localData.vid}`;
+      setImDataInCookie(localData.vid);
+    }
+
+    if (!localData.id) {
+      return {callback: callImuidSync(syncUrl)}
```
msm0504, line 134: This does not match the specified return type in your comments:
```
* @returns {{id: string} | undefined}}}
```
In this case, will whatever function called `getId` need to run the sync?

msm0504 (changes requested):

test/spec/modules/imuIdSystem_spec.js
```diff
@@ -0,0 +1,43 @@
+import { imuIdSubmodule, storage } from 'modules/imuIdSystem.js';
```
msm0504, line 1: Please add additional unit tests. Coverage is currently at only 48%

modules/imuIdSystem.js
eknis: > This does not match the specified return type in your comments
fix comments.

> In this case, will whatever function called getId need to run the sync?
As you pointed out, 'sync' has been renamed because it is misleading.

test/spec/modules/imuIdSystem_spec.js
eknis: add unit tests.

msm0504 (approved):

PR #7275, opened 2021-08-07
PubMatic: if multi-format ad-unit does not have outstreamAU or renderer (for out-stream) then continue w/o video

pm-harshad-mane:
## Type of change
- [ X ] Bugfix

**Please note: Do Not merge, the changes are in review and in QA.**

## Description of change
if the a Prebid ad unit is configured for outstream video and the outstreamAU and renderer parameter is not set for PubMatic bidder, then PubMatic bidder doesn't bid on the ad unit at which is an opportunity loss for PubMatic.

## Expected Behavior
If PubMatic adapter is configured on a multi-format ad unit that includes outstream and oustreamAU as well as renderer is not set, we bid the other formats (banner and/or native) but not video.  Issue a warning saying that video will not be bid due to outstreamAU and renderer being unset

PubMatic Bid Adapter: if multi-format ad-unit does not have outstreamAU or renderer (for out-stream) then continue w/o video.

**Please note: Do Not merge, the changes are in review and in QA.**


mike-chowla:
@pm-harshad-mane Aren't we still going to send a bid request with video object?  Shouldn't we suppress the video portion of bid request when there's no rendered available?

I'm thinking a check here to skip video if there's no renderer:
https://github.com/prebid/Prebid.js/blob/e6be849cbb81261681964915f74157de26025786/modules/pubmaticBidAdapter.js#L648

patmmccann:
Why don't you just bid and append a renderer?

mike-chowla:
@pm-harshad-mane Makes sense to me

mike-chowla (approved):
LGTM

pm-harshad-mane:
> Why don't you just bid and append a renderer?

@mike-chowla this approach is better, let's discuss how to handle this.

PR #7285, opened 2021-08-11
Dependencies: Fix dependabot dependency path parse

ChrisHuie:
Fixes new security recommendation only because of dependabot scary warning. No real impact to Prebid on the security side other than better maintainability. https://github.com/prebid/Prebid.js/security/dependabot/package-lock.json/path-parse/closed


umakajan (approved):
LGTM 👍

PR #7289, opened 2021-08-11
TrustX Bid Adapter: Use new format by default + new keywords logic(4.43.x-legacy)

PWyrembak:
## Type of change

- [x] Feature

## Description of change

Use new format as default + new format for keywords

- paul@trustx.org


patmmccann:
There are no more legacy releases

PR #7290, opened 2021-08-11
Sharethrough adapter: connect to OpenRTB endpoint

epechuzal:
## Type of change
- [x] Feature

## Description of change

Update the Sharethrough bid adapter to connect to our OpenRTB endpoint


epechuzal:
> @epechuzal looks like you have a few linting errors in this pr

@ChrisHuie thanks for the heads up!

Looks like I cannot use the [??](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Nullish_coalescing_operator) operator or [optional chaining](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Optional_chaining). For optional chaining I saw that I can use `utils.deepAccess`, do you guys provide a method for the `??` operator?

For now I made our own stub, but let me know if there is already one provided.

ChrisHuie:
@epechuzal we don't really have something with similar functionality. Just now really looking at these features though and they do seem very useful. I assume updating the Babel env would allow the use of these features if we want it added.

modules/sharethroughBidAdapter.js
```diff
@@ -1,123 +1,178 @@
...
-  b64EncodeUnicode,
-  handleIframe,
-  isLockedInFrame,
-  getProtocol
+  getProtocol,
 };
 
 export const sharethroughAdapterSpec = {
   code: BIDDER_CODE,
+  supportedFormat: [BANNER, VIDEO],
```
ncolletti, line 20: This isn't needed with `supportedMediaTypes` defined in the spec

modules/sharethroughBidAdapter.js
```diff
@@ -1,123 +1,178 @@
...
+          skip: nullish(videoRequest.skip, 0),
+          linearity: nullish(videoRequest.linearity, 1),
+          minduration: nullish(videoRequest.minduration, 5),
+          maxduration: nullish(videoRequest.maxduration, 60),
+          playbackmethod: videoRequest.playbackmethod || [2],
+          api: getVideoApi(videoRequest),
+          mimes: videoRequest.mimes || ['video/mp4'],
+          protocols: getVideoProtocols(videoRequest),
+          w: videoRequest.playerSize[0],
+          h: videoRequest.playerSize[1],
```
ncolletti, line 104: May help to verify playerSize array is included in your `isBidRequestValid` function for video requests since there is no default that can be set.

modules/sharethroughBidAdapter.js
```diff
@@ -1,123 +1,178 @@
...
-      if (bidRequest.params.badv) {
-        query.badv = bidRequest.params.badv;
+      const gpid = utils.deepAccess(bidReq, 'ortb2Imp.ext.data.pbadslot');
+      if (gpid) {
+        impression.ext = { gpid: gpid };
       }
 
-      if (bidRequest.params.bcat) {
-        query.bcat = bidRequest.params.bcat;
+      if (bidReq.mediaTypes && bidReq.mediaTypes.video) {
```
ncolletti, line 90: You should check that the context for the video request is not outstream

modules/sharethroughBidAdapter.js
```diff
@@ -138,175 +193,77 @@ export const sharethroughAdapterSpec = {
...
-    // only add sfp js to window.top if iframe busting successfully loaded; otherwise, add to iframe
-    try {
-      if (iframeBusterLoaded) {
-        window.top.document.getElementsByTagName('body')[0].appendChild(sfpJs);
-      } else {
-        window.document.getElementsByTagName('body')[0].appendChild(sfpJs);
-      }
-    } catch (e) {
-      utils.logError('Trouble writing sfp script, error details:', e);
+function getFloor(bid) {
```
ncolletti, line 226: Could this be renamed so it does not match the similar function provided by Prebid in the bid?

modules/sharethroughBidAdapter.js
```diff
@@ -1,123 +1,178 @@
...
+const STR_ENDPOINT = `https://btlr.sharethrough.com/universal/v1?supply_id=${SUPPLY_ID}`;
 
 // this allows stubbing of utility function that is used internally by the sharethrough adapter
 export const sharethroughInternal = {
-  b64EncodeUnicode,
-  handleIframe,
-  isLockedInFrame,
-  getProtocol
+  getProtocol,
 };
```
ncolletti, line 16: Is this still necessary? It seems there are a good amount of utility functions added in this PR that aren't using this namespace. I'm fine with leaving it as-is, simply curious what the intent is for the one function.

epechuzal: @ncolletti it was originally to allow stubbing in specs - right now `getProtocol` is the only method that we need to mock in order to run unit tests, but in the past we've had more. If this isn't an issue I'd favor keeping it

modules/sharethroughBidAdapter.js
epechuzal: Renamed to `getBidRequestFloor`

modules/sharethroughBidAdapter.js
epechuzal: Ok, removing `supportedFormat` attribute here

modules/sharethroughBidAdapter.js
epechuzal: @ncolletti good idea, how would I go about doing this?

modules/sharethroughBidAdapter.js
epechuzal: good catch - I set some defaults

modules/sharethroughBidAdapter.js
epechuzal: @ncolletti added some logic to the `isBidRequestValid` method, to filter out video requests which are not outstream

ncolletti: @epechuzal This may be problematic to include in `isBidRequestValid` for publishers who use multiple mediatypes in their request to your adapter. For example, if you were to include 1 banner and 1 video outstream for the same adunit, that request would fail and the imp object would not build for the banner request. If you move the outstream context check to the line above, 90, that will not occur.

ncolletti:
@epechuzal this is looking good, few small things to adjust:

- The params from your markdown file do not respond with a bid, generally we recommend adapters keep this up to date as a test to verify their endpoints are communicating properly and returning valid bids, could this be updated?
- It seems the params `iframe` and `iframeSize` listed in your markdown file are no longer used?
- Please add an example of a video instream request in your markdown file with suggested video params

modules/sharethroughBidAdapter.js
epechuzal: @ncolletti ok so I moved it to line 90 of `buildRequests`. If the context is outstream, then we just return `null`

Later will filter out any falsy `imps` so that we don't end up sending `null` to the adserver. 

Would this work?

epechuzal:
@ncolletti I made some updates to the playerSize (defaults not being set correctly) and documentation - does it look good on your end?

modules/sharethroughBidAdapter.js
ncolletti: Yep this is good, thank you

ncolletti:
Looking good, last piece would be updating the pkey values in your markdown file to return bids from your endpoint so we can verify it's working properly.

epechuzal:
> Looking good, last piece would be updating the pkey values in your markdown file to return bids from your endpoint so we can verify it's working properly.

@ncolletti the `pkey` stands for `placement key`, which we can't pre-fill in: it is up to the publisher to input a placement key that has been set up to receive traffic. Because of this, we can't provide a default that will "always work". Is this a blocker to approving the PR?

patmmccann:
@epechuzal The ability for a placement id that returns house demand with a real creative like 13144370 does for Xandr at https://docs.prebid.org/dev-docs/getting-started.html example is one of the most frequently requested features by publishers. It would be a huge help to every integration team to duplicate this functionality.

epechuzal:
@patmmccann creating a placement ID that always returns demand like this would require work on our end which is not related to this PR at all. Can we add that in a future pull request instead?

patmmccann:
Can you privately give a placement id to your reviewer so he can take a look?

patmmccann:
Also, I mentioned this to Patrice and Lukas, but instead of convincing publishers to move to this endpoint from your DMX endpoint, you should just make commits to the dmx adapter to make it hit this endpoint instead conditional on this adapter not existing. It would potentially grow your company enormously nearly overnight.

epechuzal:
@patmmccann right now, I can't provide that - is this really a blocker to approval? We've never had to provide live placement keys before.

ncolletti:
@epechuzal -- Could you work on including this in a later PR, it's useful for the Prebid team and publishers to be able to test your adapter and endpoint using the example params information from your markdown file. In the meantime, this PR looks good to me. Could you rebase the latest changes from master here, there is a merge conflict in your adapter, then I can approve. Thank you!

epechuzal:
@ncolletti thanks, I just resolved the conflicts. I had an issue with my git cli  and had to reset the branch (in case youre wondering about the force-push above), but I think its all good now

ncolletti (approved):
LGTM

PR #7295, opened 2021-08-12
Zeta Ssp Bid Adapter: support video

asurovenko-zeta:
## Type of change
- [x] Feature

## Description of change
Support video media type for Zeta Ssp Bid Adapter


asurovenko-zeta:
@msm0504 
Can I ask for an increase in priority if possible? Our team expects video traffic in the near future. Thank you!

modules/zeta_global_sspBidAdapter.js
```diff
@@ -178,6 +217,49 @@ function buildBanner(request) {
...
+    if (videoParams.hasOwnProperty(key)) {
+      video[key] = checkParamDataType(key, videoParams[key], VIDEO_CUSTOM_PARAMS[key]);
+    }
+  }
+  if (videoParams.playerSize) {
+    if (utils.isArray(videoParams.playerSize[0])) {
+      video.w = parseInt(videoParams.playerSize[0][0], 10);
+      video.h = parseInt(videoParams.playerSize[0][1], 10);
+    } else if (utils.isNumber(videoParams.playerSize[0])) {
+      video.w = parseInt(videoParams.playerSize[0], 10);
```
msm0504, line 233: Why does `parseInt` need to be run on a value that is already a number?

msm0504 (changes requested):

modules/zeta_global_sspBidAdapter.js
asurovenko-zeta: @msm0504 **45.5** it's a number, but we need only _int_ values - **45**.

msm0504 (approved):

Issue #7078, opened 2021-06-22
Enhance RTD infrastructure to support bidRequest event


bretg:
## Type of issue
Enhancement

## Description
RTD modules need to be able to set first party data per-bidder. This can be done with setBidderConfig, but it's awkward.

The proposal is to just extent the existing RTD infrastructure to support one more event:

1) add `onBidderRequestEvent` function interface with the same signature as onBidResponseEvent except the first arg is a bid request rather than a bid response. (am assuming bid request objects contain the biddercode)

2) RTD-core calls this interface for each module when the event fires. In setEventsListeners, add:

```
  events.on(CONSTANTS.EVENTS.BID_REQUESTED, (args) => {
    subModules.forEach(sm => { sm.onBidderRequestEvent && sm.onBidderRequestEvent(args, sm.config, _userConsent) })
  });
```

3) Update docs

## Related issues

- https://github.com/prebid/Prebid.js/pull/7001

PR #7296, opened 2021-08-12
emitting BEFORE_BIDDER_REQUEST event per bidder network request

elad-yosifon:
## Type of change

- [ ] Bugfix
- [x] Feature
- [ ] New bidder adapter   
- [ ] Build related changes
- [ ] CI related changes
- [ ] Other

## Description of change
emitting BEFORE_BIDDER_REQUEST event every time a BidAdapter send a network request to the endpoint.

API change PR:
https://github.com/prebid/prebid.github.io/pull/3197

### DIFF
--- a/src/adapters/bidderFactory.js
+++ b/src/adapters/bidderFactory.js
@@ -230,6 +230,7 @@ export function newBidder(spec) {
       // Server requests have returned and been processed. Since `ajax` accepts a single callback,
       // we need to rig up a function which only executes after all the requests have been responded.
       const onResponse = delayExecution(configEnabledCallback(afterAllResponses), requests.length)
+      requests.forEach(_ => events.emit(CONSTANTS.EVENTS.BEFORE_BIDDER_REQUEST, bidderRequest));
       requests.forEach(processRequest);
 
       function formatGetParameters(data) {
--- a/src/constants.json
+++ b/src/constants.json
@@ -34,6 +34,7 @@
     "BIDDER_DONE": "bidderDone",
     "SET_TARGETING": "setTargeting",
     "BEFORE_REQUEST_BIDS": "beforeRequestBids",
+    "BEFORE_BIDDER_REQUEST": "beforeBidderRequest",
     "REQUEST_BIDS": "requestBids",
     "ADD_AD_UNITS": "addAdUnits",
     "AD_RENDER_FAILED": "adRenderFailed",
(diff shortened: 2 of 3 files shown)


patmmccann:
what utility does this provide that BEFORE_REQUEST_BIDS does not?

elad-yosifon:
@patmmccann  some bidders opt for SRA and one request for one auction, some opt for one request per ad unit, some opt for one request per ad unit and size.
Monitoring network load, using the existing events  is impossible, because there are no details in regards to network access.

This event is emitted per network request.

bretg:
Looks like this event is readonly - listeners don't have a chance to cancel the network fetch.
 
How about naming this event BEFORE_BIDDER_ENDPOINT?  Calling it BEFORE_BIDDER_REQUEST is a little confusing to me -- gives the impression that it's before calling the adapter.

elad-yosifon:
@bretg  sure.. NP
Should I just push a change? Or should we wait for more opinions?

mkendall07:
I like `BEFORE_BIDDER_HTTP` - agree that request is somewhat confusing.

elad-yosifon:
@mkendall07 `BEFORE_BIDDER_HTTP` sounds good :)

ChrisHuie:
@elad-yosifon looks like our CircleCi tests didn't kick off on this pr so kicked them off

patmmccann:
@bretg does this solve #7078 ?

bretg:
>  does this solve #7078 ?

No - that issue is for RTD modules to be able to add functionality at an additional step.

elad-yosifon:
@bretg @mkendall07 @ChrisHuie @patmmccann should we go with `BEFORE_BIDDER_HTTP`?

bretg:
> BEFORE_BIDDER_HTTP

Yes.

bretg:
@elad-yosifon - I would say no, 4.x is closed. No more maintenance. You can take it to the PBJS committee if you'd like.

elad-yosifon:
@bretg In general I agree, but (unfortunately) 5.x versions are still buggy or not performing as good as 4.x (I ran a some A/B tests on more than a million users WW).

I am currently (forced to be) running 4.36 on production, as it seems to be the most stable and performant version in comparison to later versions.

patmmccann:
@elad-yosifon could you add more detail here? Which pr do you suspect introduced performance issues?

elad-yosifon:
@patmmccann I have no idea, since I didn't run all combinations of versions.
But I can confirm that 5.6 is hurting older browsers and safari a bit .

I will go live with a new test with other versions and we'll see if that's still the case...

bretg:
@gglas - perhaps worth bringing up in the channel.

gglas:
Pinged, will also ask at next meeting in 2 weeks.

PR #7298, opened 2021-08-12
ogury Bid Adapter: handle onBidWon event on prebid.js

jogury:
## Type of change

- [ x] Feature

## Description of change

implementation of onBidWon event on Ogury adapter


patmmccann:
Does it make sense to add sanitation on the returned string?

jogury:
> Does it make sense to add sanitation on the returned string?

Hello @patmmccann, we will have total control over this nurl string, as it will be sanitized and generated server-side with all the parameters/macros already filled, that’s why we use it directly as we have seen in some other adapters like here: https://github.com/prebid/Prebid.js/blob/8183e8500ee37c8445ae68949326741a4b3eb233/modules/outbrainBidAdapter.js#L138
Or here https://github.com/prebid/Prebid.js/blob/180ddafbb8ebf30f00055e871f615243a2f4deb7/modules/koblerBidAdapter.js#L49

ChrisHuie (approved):

Issue #7287, opened 2021-08-11
installedModules does not work for NPM build


muuki88:
## Type of issue

Bug / Feature Request

## Description

The `pbjs.installedModules` returns `[ "v$prebid.modulesList$" ]` instead of the registered modules if built via NPM.

https://github.com/prebid/Prebid.js/blob/11df18dff4bf10f05ed01da62f214ce13690a2d2/src/prebid.js#L47

## Steps to reproduce

Built prebid via NPM as decribed in the readme.

## Test page

https://www.gutefrage.net/frage/kann-ich-ohne-abi-aber-mit-einer-pta-ausbildung-medizin-studieren

### Expected results

`pbjs.installedModules` returns all registered modules

### Actual results

 `[ "v$prebid.modulesList$" ]` is returned

## Platform details

Prebid 5.7

## Other information

- Feature was added in #6601

patmmccann:
fyi @mmoschovas

bretg:
@muuki88 - we're going to pick this up soon, but we're curious: how are you using the `installedModules`?

dgirardi:
@bretg This is going to be hard to fix. `installedModules` is generated by our build process; if you're using a different build process you'll have to replicate the logic - we can think about making that easier, but all fast solutions will require the consumer to provide the list of modules they want in there. For example it would look like babel plugin parameters, instead of the current:

```
      {
        test: /.js$/,
        include: new RegExp(`\\${path.sep}prebid\\.js`),
        use: {
          loader: 'babel-loader',
          options: require('prebid.js/.babelrc.js')
        }
      }
```

We could have:

```
      {
        test: /.js$/,
        include: new RegExp(`\\${path.sep}prebid\\.js`),
        use: {
          loader: 'babel-loader',
          options: require('prebid.js/babelConfig.js')({ installedModules: [... manually maintained list of modules ...] })
        }
      }
```

Generating that automatically would require some research, I am not sure that module names are always inferrable from imports.

dgirardi:
This is where we convert our build's `--installedModules` into files to include:
https://github.com/prebid/Prebid.js/blob/a2661725f9a39cf1c7cbf8f102d601769978a138/gulpHelpers.js#L73-L106

We need to reverse that and go from file to module name. It looks like that's doable with "internal" modules (because they all follow the patterns: `modules/moduleName.js` => `moduleName`, or `modules/moduleName/index.js` => 'moduleName'), but I don't see an easy way to detect external modules.

[...]

PR #7304, opened 2021-08-16
Gulp Build: fix installed gulp modules returning placeholder value

ChrisHuie:
Change regex to specific string for attempted fix on issue => #7287

### DIFF
--- a/gulpfile.js
+++ b/gulpfile.js
@@ -169,8 +169,8 @@ function makeWebpackPkg() {
     .pipe(helpers.nameModules(externalModules))
     .pipe(webpackStream(cloned, webpack))
     .pipe(terser())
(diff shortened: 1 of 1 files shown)


FilipStamenkovic (commented):
When I do (taken from readme):

```
gulp build --modules=openxBidAdapter,rubiconBidAdapter,sovrnBidAdapter
```

My `prebid.js` is built with following piece of code:

```
A.installedModules=[""openxBidAdapter","rubiconBidAdapter","sovrnBidAdapter""]
```

Then, when I try to load that .js in some html page it crashes because of duplicate `""`.


Doing the same command on `master` branch it looks fine. Here is what I get on `master`:

```
E.installedModules=["openxBidAdapter","rubiconBidAdapter","sovrnBidAdapter"]
```

ChrisHuie:
@FilipStamenkovic thank you. Let me check what is going on with the extra quotes.

PR #7308, opened 2021-08-16
adding no-adserver example

bretg:
## Type of change

- [ X ] Other - Example

## Description of change
[An issue](https://github.com/prebid/prebid.github.io/issues/3162) was opened over on the docs site that I assume is asking for a better example for how to run without an ad server. 

Came up with this integrationExample. If this looks good to others, will add a link in a FAQ entry over on the docs site once it's merged.


patmmccann (approved):

patmmccann:
pbjs.renderAd(iframeDoc, winningBid.adId); seems more elegant than using the utility function get ad server targeting to extract the hb_adid to me.

PR #7310, opened 2021-08-16
Gumgum: fix how we send TTD global placement ID

lbenmore:
## Type of change

- [x] Bugfix
- [ ] Feature
- [ ] New bidder adapter   
- [ ] Build related changes
- [ ] CI related changes
- [ ] Other

## Description of change
Changes where we attach the `gpid` property to send as part of the request to our ad server

### DIFF
--- a/modules/gumgumBidAdapter.js
+++ b/modules/gumgumBidAdapter.js
@@ -346,8 +346,7 @@ function buildRequests (validBidRequests, bidderRequest) {
       sizes,
       url: BID_ENDPOINT,
       method: 'GET',
-      gpid: gpid,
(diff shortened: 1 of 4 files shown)


ChrisHuie:
@lbenmore looks like there are a few errors popping up from the gpid request change

patmmccann:
There are several other values not ending up in your endpoint query string parameters. eg tmax, Do you want to fix those too?

lbenmore:
@ChrisHuie, the unit tests have been updated and should no longer be failing.
@patmmccann, thx for calling those out; i'll connect with our backend to see if these properties should be moved or if they've been deprecated and can be removed since these haven't been called out as not-working. either way, that'll be a new ticket on our end so it will be submitted in a separate PR

ChrisHuie (approved):

Issue #6528, opened 2021-04-05
pbsBidAdapter should pass more ortb2 values through to server


bretg:
## Type of issue
enhancement

## Description
Part of the intention of the new `ortb2` method of pubs specifying first party data was for this config to be sent straight through to Prebid Server.

However, currently the pbsBidAdapter only looks for site, user, and imp.ext.data

I propose dramatically expanding what gets passed through to PBS to include **everything** in ortb2.

1) read getConfig({ortb2}) and merge it
2) pass bidder-specific config through as ext.prebid.bidderconfig
3) read adunit.ortb2Imp and merge that to the imp[]. This lets them specify imp.ext, imp.pmp, imp.intstl, etc.

patmmccann:
@pm-harshad-mane recently pointed this out as well, that pbs was missing ortb2 device object.

At a minimum we should add device,  but I cannot think of what we wouldn't want to add and this seems like a great proposal.

muuki88:
This extension would also allow to set floor prices via the floor price module on the `ortb2` object. This allows a smoother transition from client side floor settings to server side floor settings.

bretg:
@Fawke noted in #7219 that another use case for this is bcat/badv.

bretg:
Closed with #7315

PR #7315, opened 2021-08-18
PBS Bid Adapter: Add additional ortb2 fields in request object

Fawke:
## Type of change
- [X] Feature

## Description of change

Added support for few additional **ortb2** fields in the request object created by the PrebidServer Bid Adapter:

- request.bcat
- request.badv
- request.site.publisher > domain & name properties

Example:

```(javascript)
pbjs.setConfig({
  ortb2: {
    site: {
      ...
      publisher: {
        name: 'xyz',
        domain: 'xyz.com'
      }
    },
    user: {...},
    badv: ['a.com', 'b.com'],
    bcat: ['IAB25', 'IAB7-39']
  }
})
```

Currently, `bcat` & `badv` fields can be set globally using **setConfig()** and it can also be set for specific bidders using **setBidderConfig()**. I did not find any use case of setting them at an AdUnit level, but please feel free to comment otherwise if you see any use case for it.

Here is a list of minor changes that are also related:
- Added auto field population for the `site.publisher.domain` through the Enrichment module.
- Added validations to the `site.publisher` object.
- Removed a minor bug in FPD Validation module where it would throw a misleading warning message.

>>NOTE: This PR requires a Docs PR, am just waiting for someone to validate this approach before I create one.
Please do not merge this.

Related - https://github.com/prebid/Prebid.js/issues/6528

### DIFF
--- a/modules/enrichmentFpdModule.js
+++ b/modules/enrichmentFpdModule.js
@@ -3,11 +3,67 @@
  * @module modules/firstPartyData
  */
 import * as utils from '../src/utils.js';
-import { submodule } from '../src/hook.js'
-import { getRefererInfo } from '../src/refererDetection.js'
+import { submodule } from '../src/hook.js';
+import { getRefererInfo } from '../src/refererDetection.js';
+import { getCoreStorageManager } from '../src/storageManager.js';
 
 let ortb2 = {};
 let win = (window === window.top) ? window : window.top;
+export const coreStorage = getCoreStorageManager('enrichmentFpd');
+
+/**
+  * Find the root domain
+  * @param {string|undefined} fullDomain
+  * @return {string}
+*/
+export function findRootDomain(fullDomain = window.location.hostname) {
+  if (!coreStorage.cookiesAreEnabled()) {
+    return fullDomain;
+  }
+
+  const domainParts = fullDomain.split('.');
+  if (domainParts.length == 2) {
+    return fullDomain;
+  }
+  let rootDomain;
+  let continueSearching;
+  let startIndex = -2;
+  const TEST_COOKIE_NAME = `_rdc${Date.now()}`;
+  const TEST_COOKIE_VALUE = 'writeable';
+  do {
+    rootDomain = domainParts.slice(startIndex).join('.');
+    let expirationDate = new Date(utils.timestamp() + 10 * 1000).toUTCString();
+
+    // Write a test cookie
+    coreStorage.setCookie(
+      TEST_COOKIE_NAME,
+      TEST_COOKIE_VALUE,
+      expirationDate,
+      'Lax',
+      rootDomain,
+      undefined
+    );
+
+    // See if the write was successful
+    const value = coreStorage.getCookie(TEST_COOKIE_NAME, undefined);
+    if (value === TEST_COOKIE_VALUE) {
+      continueSearching = false;
+      // Delete our test cookie
+      coreStorage.setCookie(
+        TEST_COOKIE_NAME,
+        '',
+        'Thu, 01 Jan 1970 00:00:01 GMT',
+        undefined,
+        rootDomain,
+        undefined
+      );
+    } else {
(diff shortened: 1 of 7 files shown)


modules/enrichmentFpdModule.js
```diff
@@ -37,7 +37,10 @@ function setDomain() {
 
   let domain = parseDomain(getRefererInfo().canonicalUrl)
 
-  if (domain) utils.mergeDeep(ortb2, { site: { domain: domain } });
+  if (domain) {
+    utils.mergeDeep(ortb2, { site: { domain: domain } });
+    utils.mergeDeep(ortb2, { site: { publisher: { domain: domain } } });
```
bretg, line 42: Prebid Server defines these values differently. site.domain is meant to be the 'raw' domain. e.g. www.example.com. site.publisher.domain is the 'aggregated' version of this domain -- just example.com in this instance.

So I'd propose we keep this convention in Prebid.js as well. i.e. implement/share the findRootDomain  function in modules/userId/index.js

If we don't, then site.publisher.domain will be passed raw to PBS and won't be  aggregated there.

modules/prebidServerBidAdapter/index.js
```diff
@@ -371,6 +371,12 @@ function addBidderFirstPartyDataToRequest(request) {
       if (currBidderConfig.ortb2.user) {
         ortb2.user = currBidderConfig.ortb2.user;
       }
+      if (currBidderConfig.ortb2.bcat) {
+        ortb2.bcat = currBidderConfig.ortb2.bcat;
```
bretg, line 375: Why not just merge the entire ortb2 object into the PBS request? A main point of that object was that we didn't have to weigh PBJS down with choosing which fields to send... just send the whole thing IMO.

bretg (approved):
thanks for doing this @Fawke - couple of comments

bretg (changes requested):
comments in-line

modules/enrichmentFpdModule.js
Fawke: @bretg, thank you for this clarification. I just need to know one more thing, does the root domain include subdomains as well?

for example, if the root domain is `mydomain.example.com`, then would we use `example.com` or `mydomain.example.com` as the value for site.publisher.domain.

modules/prebidServerBidAdapter/index.js
Fawke: @bretg I agree with your approach, it makes things a lot more simple at PBJS side. Can you check my latest commit to see if that's what we wanted?

I have also removed the **if checks** for `bidderFirstPartyData`. Do wanna keep those checks for bidderFirstParyData?

Also, now, since there is no check in place, the publisher can pass any property downstream to PBS. I hope that's alright.

modules/enrichmentFpdModule.js
bretg: the 'root' domain is meant to be the broadest possible domain you can still set a cookie in. So in your example, it would be `example.com`. As another example, the root domain for www.sports.example.co.uk is example.co.uk

mmoschovas (commented):
Glad to see the enrichment/validation modules getting some use! Once the domain change @bretg requested is put through this should be good

mmoschovas:
Are we opposed to making the functional portion part of a util function that can be imported into this and the userId module? It just seems odd that we have two of the exact same function living in different places.

Fawke:
@mmoschovas The problem with this approach is that we'll need to import storageManager in utils.js file, since the findRootDomain function uses that, and we don't wanna do that.

There was already some talk about this [here](https://github.com/prebid/Prebid.js/pull/6124#discussion_r544361171)

jsnellbaker (approved):

mmoschovas (approved):

bretg (approved):

bretg:
I'll update the enrichment module docs now...

PR #7322, opened 2021-08-19
Publink Id System (Conversant): add new user id module

johnwier:
## Type of change
- [X ] Feature

## Description of change
pbjs.setConfig({
userSync: {
userIds: [{
name: "publinkId",
storage: {
name: "pbjs_publink",
type: "cookie",
expires: 30
},
params: {
e: "e80b5017098950fc58aad83c8c14978e", // example hashed email (md5)
}
}],
auctionDelay: 50 // 50ms maximum auction delay, applies to all userId modules
}
});

Link to documentation PR 
https://github.com/prebid/prebid.github.io/pull/3225

### DIFF
--- a/modules/publinkIdSystem.js
+++ b/modules/publinkIdSystem.js
@@ -1,0 +1,102 @@
+/**
+ * This module adds the PublinkId to the User ID module
+ * The {@link module:modules/userId} module is required
+ * @module modules/conversantIdSystem
+ * @requires module:modules/userId
+ */
+
+import {submodule} from '../src/hook.js';
+import {getStorageManager} from '../src/storageManager.js';
+import {ajax} from '../src/ajax.js';
+import * as utils from '../src/utils.js';
+import {uspDataHandler} from '../src/adapterManager.js';
+
+const MODULE_NAME = 'publink';
+const GVLID = 24;
+
+export const storage = getStorageManager(GVLID);
+
+function publinkIdUrl(params, consentData) {
+  let url = utils.parseUrl('https://proc.ad.cpe.dotomi.com/cvx/client/sync/publink');
+  url.search = {
+    deh: params.e,
+    mpn: 'Prebid.js',
+    mpv: '$prebid.version$',
+  };
+  if (consentData) {
+    url.search.gdpr = (consentData.gdprApplies) ? 1 : 0;
+    url.search.gdpr_consent = consentData.consentString;
+  }
+
+  const usPrivacyString = uspDataHandler.getConsentData();
+  if (usPrivacyString && typeof usPrivacyString === 'string') {
+    url.search.us_privacy = usPrivacyString;
+  }
+
+  return utils.buildUrl(url);
+}
+
+function makeCallback(config = {}, consentData) {
+  return function(prebidCallback) {
+    const options = {method: 'GET', withCredentials: true};
+    let handleResponse = function(responseText, xhr) {
+      if (xhr.status === 200) {
+        let response = JSON.parse(responseText);
+        if (response) {
+          prebidCallback(response.publink);
+        }
+      }
+    };
+    if (window.coreid && typeof window.coreid.getPublinkId === 'function') {
+      window.coreid.getPublinkId((newId) => {
+        if (newId) {
+          prebidCallback(newId);
+        } else {
+          if (config.params && config.params.e) {
+            ajax(publinkIdUrl(config.params, consentData), handleResponse, undefined, options);
+          }
+        }
+      });
+    } else {
+      if (config.params && config.params.e) {
+        ajax(publinkIdUrl(config.params, consentData), handleResponse, undefined, options);
+      }
+    }
+  };
+}
+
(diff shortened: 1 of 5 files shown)


patmmccann:
How is this different from the publisher provided id module?

johnwier:
We wanted this to be self contained so this is the only thing the publishers need and not have to load additional scripts

patmmccann:
This is a module to pass hashed email, we already have a module to pass hashed email; I think I'm missing something

modules/publinkIdSystem.js
```diff
@@ -0,0 +1,102 @@
+/**
+ * This module adds the PublinkId to the User ID module
+ * The {@link module:modules/userId} module is required
+ * @module modules/conversantIdSystem
```
FilipStamenkovic, line 4: `conversantIdSystem` could be confusing, maybe you should rename it to `publinkIdSystem`.

FilipStamenkovic (changes requested):
Code looks fine.  
Take a look at my other comments.  
Also, include your submodule in `.submodules.json` list.

patmmccann:
Hi @johnwier 

As your config requires a hashed id, a pub will still have to write script to populate it. Is there any additional functionality this module provides that is not in existing modules? If not, I think it makes more sense for adapters to pull from existing modules rather than provide competing functionality that is identical.

johnwier:
I've finished the recent updates so there's no need to hold off now

FilipStamenkovic (approved):
Code looks good to me.  
@patmmccann what do you think?

modules/userId/eids.js
```diff
@@ -228,6 +228,10 @@ const USER_IDS_CONFIG = {
     source: 'amxrtb.com',
     atype: 1,
   },
+  'publinkId': {
+    source: 'epsilon.com',
```
patmmccann, line 232: this seems incorrect; how is epsilon a source, they appear to be a consumer of the id

patmmccann:
my only remaining question is on the source in the eids object. It seems the publisher domain should be the source? Or perhaps match to pubcid.org? I am confused on the epsilon branding.

johnwier:
since conversant is owned by epsilon there is an ongoing effort to make the branding more consistent

patmmccann:
how is this related to conversant other than you authored the PR?

patmmccann:
to put this another way, if this is simply an id module to put email hashes in bid requests, why would conversant or epsilon be the source of the information?

patmmccann:
Perhaps I am missing something here, is the workflow the publisher hashes the email, hits the conversant endpoint and they end back a rehash? If so, than perhaps the rehash makes sense as having epsilon as the source.

pycnvr:
@patmmccann Yes, that's correct.  It's an encrypted guid so conversant/epsilon is the issuer.  Thanks again for the help!
@FilipStamenkovic Thanks for the reviews, too.

jdwieland8282:
@pycnvr are you still up for walking the Prebid Identity PMC through you new module tomorrow?  I'm personally curious to learn more about how you get the users email into `params.e`

pycnvr:
@jdwieland8282 Sure.  The flow is straightforward.  It would require a bit of custom codes from the publisher to call pbjs.setConfig one more time when the hash is available, followed up by a call to pbjs.refreshUserIds to crank the wheel.  That's why it'd be much nicer if we can integrate with the import lib somehow.

jdwieland8282:
Right I sort of thought that's the direction we wanted to go, Publink could grab the hashed email from the import lib.  We should at a min support that for pubs who don't want to write custom code to add a hashed email to `params.e`. 

Second question, for pubs who want to populate `params.e` on their own, how do we ensure that the value is hashed?

pycnvr:
@jdwieland8282 That being a config parameter, it'd be up to PubLink and other similar modules to do a validation.  A regex for hex would be sufficient.  Much easier than email address scans.  What do you think?

jdwieland8282:
@pycnvr  agreed.  

More broadly, I think we are at a crossroads here and this module is a good forcing mechanism for us to discuss it.  That crossroads is "does the Prebid community want to have an official position on passing clear text emails, hashed emails, or no emails", and really it's a debate between the first 2 at this point.

patmmccann:
@pycnvr it seems this pr skips some circleci tests for reasons I cannot determine. Most other recent PRs are doing the same and we are ttrying to fix. However, this is what happens when all tests run https://app.circleci.com/pipelines/github/prebid/Prebid.js/8055/workflows/0e2d84af-44e7-4586-b5f3-c6f810c101c0/jobs/17925 or https://app.circleci.com/pipelines/github/prebid/Prebid.js/8067/workflows/7312b412-4840-4c0e-b3a6-79bd9994a203/jobs/17937

You can see on IE, the :443 protocol is added to the url and the test fails. Can you look into this by any chance? 

Thanks!

pycnvr:
@patmmccann Yes I'll take a look.  Thanks!

Issue #7323, opened 2021-08-19
if RTD module return targeting data with multiple values (array), somehow prebid prepends a space on some values 


peczenyj:
## Type of issue

bug

## Description
I am writting a RTD module and I am trying to use `getTargetingData`  to return something like this

```json
{
  "slot A": {
     "foo": ["some","values","because","why","not"]
   }
}
```

however, GAM was not presenting the ads. Only if I set a single value it works. Weird.

I check the targeting via javascript console

```javascript
googletag.pubads().getSlots().forEach((n) =>  
    n.getTargetingKeys().forEach((k) => 
      console.log("slot",n.getAdUnitPath(),k,n.getTargeting(k)))
);
```

I can see a output like
```
slot A foo (5) ["some"," values"," because"," why"," not"]
```

I try this possible workaround, without success

```json
{
  "foo": {
     "key": "some,values,because,why,not"
   }
}
```
I believe this is side effect of the function `flattenTargeting` on `src/targeting.js` who performs

```javascript
  function flattenTargeting(targeting) {
    let targetingObj = targeting.map(targeting => {
      return {
        [Object.keys(targeting)[0]]: targeting[Object.keys(targeting)[0]]
          .map(target => {
            return {
              [Object.keys(target)[0]]: target[Object.keys(target)[0]].join(', ') // <--- HERE !

```

we have a join with comma + space, then a split by comma... keeping some extra space in the begin of some values.

Result: the target in GAM does not work because " foo" is different than "foo"

## Steps to reproduce

1. create an RTD module with getTargetingData
2. return an array of values instead a single string
3. check the GAM targeting

## Test page
not yet...

### Expected results

```
slot A foo (5) ["some","values","because","why","not"]
```

### Actual results

```lytics/cloudstorage/pulls
slot A foo (5) ["some"," values"," because"," why"," not"]
```

## Platform details
latest version of prebid: 5.10.0

PR #7160, opened 2021-07-10
PBjs Core Targeting: allow non-string (eg. numeric) targeting segments

jimdigriz:
[Documentation shows a numeric example](https://docs.prebid.org/dev-docs/add-rtd-submodule.html#gettargetingdata) which causes an exception as we try to call `.split(',')`.

## Type of change
- [x] Bugfix

## Description of change
Allows non-string values to be passed.

PR #7337, opened 2021-08-24
PBjs Core (Targeting): bugfix for issue #7323 adding extra spaces

peczenyj:
## Type of change

- [x] Bugfix
- [ ] Feature
- [ ] New bidder adapter   
- [ ] Build related changes
- [ ] CI related changes
- [ ] Other

## Description of change
This should fix issue #7323 by avoid join targeting keys with comma + space (later we split by comma and keep an extra space in front of some values and the targeting does not work)

### DIFF
--- a/test/spec/unit/core/targeting_spec.js
+++ b/test/spec/unit/core/targeting_spec.js
@@ -309,7 +309,7 @@ describe('targeting tests', function () {
         ['string', '2.3', '2.3'],
         ['number', 2.3, '2.3'],
         ['boolean', true, 'true'],
-        ['string-separated', '2.3,4.5', '2.3,4.5'],
+        ['string-separated', '2.3, 4.5', '2.3, 4.5'],
         ['array-of-string', ['2.3', '4.5'], '2.3,4.5'],
         ['array-of-number', [2.3, 4.5], '2.3,4.5'],
         ['array-of-boolean', [true, false], 'true,false']
--- a/test/spec/unit/core/targeting_spec.js
+++ b/test/spec/unit/core/targeting_spec.js
@@ -309,10 +309,10 @@ describe('targeting tests', function () {
         ['string', '2.3', '2.3'],
         ['number', 2.3, '2.3'],
         ['boolean', true, 'true'],
-        ['string-separated', '2.3,4.5', '2.3, 4.5'],
-        ['array-of-string', ['2.3', '4.5'], '2.3, 4.5'],
-        ['array-of-number', [2.3, 4.5], '2.3, 4.5'],
-        ['array-of-boolean', [true, false], 'true, false']
+        ['string-separated', '2.3,4.5', '2.3,4.5'],
+        ['array-of-string', ['2.3', '4.5'], '2.3,4.5'],
+        ['array-of-number', [2.3, 4.5], '2.3,4.5'],
+        ['array-of-boolean', [true, false], 'true,false']
       ];
       pairs.forEach(([type, value, result]) => {
         it(`accepts ${type}`, function() {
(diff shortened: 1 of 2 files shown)


patmmccann:
This appears to be a bug introduced on https://github.com/prebid/Prebid.js/pull/7160/files @jimdigriz could you comment?

jimdigriz:
> This appears to be a bug potentially introduced on https://github.com/prebid/Prebid.js/pull/7160/files @jimdigriz could you comment? Do you think this pr did introduce this bug or this behavior has been around a while?

This PR introduced coverage for code that never had coverage. The code change only affected non-string (and non-array) handling which never worked before anyway.

This module already introduced spaces before I touched it (as you can see from the one strange test case I added on https://github.com/prebid/Prebid.js/pull/7160/files#diff-e661e01a21523c55c5d819ca4a8f2b2e681ca8d4842bc099eee8db60be18bd12R312) so if you passed in the string `4.2,5.3` it would return `4.2, 5.3` (with the space added). I thought it was strange, but it did not affect what we needed and I did not want to open a can of worms.

Only side affect I could think of that would have happened is (string) `'4.3'` would have been changed to (array[string]) `[ '4.3' ]`.

jimdigriz:
Looks introduced in Prebid.js 1.0 four years ago, https://github.com/prebid/Prebid.js/blame/master/src/targeting.js#L363 so as I only heard about Prebid.js this year, no blaming me :)

peczenyj:
To be honest, I just remove the extra space after comma and run the unit tests.

But I have no idea about the side effects :)

patmmccann:
@jimdigriz @peczenyj tyvm for all the additional context

peczenyj:
cleaning my previous rebase

test/spec/unit/core/targeting_spec.js
```diff
@@ -309,10 +309,10 @@ describe('targeting tests', function () {
         ['string', '2.3', '2.3'],
         ['number', 2.3, '2.3'],
         ['boolean', true, 'true'],
-        ['string-separated', '2.3,4.5', '2.3, 4.5'],
-        ['array-of-string', ['2.3', '4.5'], '2.3, 4.5'],
-        ['array-of-number', [2.3, 4.5], '2.3, 4.5'],
-        ['array-of-boolean', [true, false], 'true, false']
+        ['string-separated', '2.3,4.5', '2.3,4.5'],
```
jimdigriz, line 312: Maybe have the input still include the space to confirm it trims it afterwards, so to make it instead:
```['string-separated', '2.3, 4.5', '2.3,4.5'],```

peczenyj: the `'2.3, 4.5'` returns `'2.3, 4.5'` and this makes some sense since no split/join was done

seems the extra space came to add some readability to the array

jimdigriz: It now does split since https://github.com/prebid/Prebid.js/commit/3b59a2b310d47092557a5605d7074bcb07a3885a so when pushing through '`2.3, 4.5`' it should now return '`2.3,4.5`' with your changes.

Att he time I was retaining the original behaviour, but maybe https://github.com/prebid/Prebid.js/commit/3b59a2b310d47092557a5605d7074bcb07a3885a should have included a call to `trim()` after splitting...

peczenyj: do you mean like on commit `dddcc23` ?

jimdigriz: > do you mean like on commit https://github.com/prebid/Prebid.js/commit/dddcc23 ?

Yep, exactly what I was thinking.

peczenyj: cool!

patmmccann (approved):

robertrmartinez (approved):

PR #7341, opened 2021-08-24
Codefuel Bid Adapter: add new bid adapter

hayimmakabee:
## Type of change

- [ ] Bugfix
- [ ] Feature
- [ ] New bidder adapter   
- [ ] Build related changes
- [ ] CI related changes
- [ ] Other

## Description of change

```
{
  bidder: '<bidder name>',
  params: {
    // ...
  }
}
```


modules/codefuelBidAdapter.js
```diff
@@ -0,0 +1,182 @@
...
+    const devicetype = getDeviceType()
+    const publisher = setOnAny(validBidRequests, 'params.publisher');
+    const cur = CURRENCY;
+    // const endpointUrl = 'http://localhost:5000/prebid'
+    const endpointUrl = 'https://ai-i-codefuel-ds-rtb-us-east-1-k8s-internal.seccint.com/prebid'
+    const timeout = bidderRequest.timeout;
+
+    const imps = validBidRequests.map((bid, id) => {
+      bid.netRevenue = 'net';
+      const imp = {
```
tamirnPerion, line 44: Cosmetic suggestion:

```suggestion
            return {
        id: idx + 1,
        tagid: bid.params.tagid || '',
        banner: {
          format: bid.sizes? transformSizes(bid.sizes) : ''
        }
      }
```

modules/codefuelBidAdapter.js
```diff
@@ -0,0 +1,182 @@
...
+        id: id + 1 + ''
```
tamirnPerion, line 45: why do we concatinating the ' ' ?

modules/codefuelBidAdapter.js
```diff
@@ -0,0 +1,182 @@
...
+    const ua = navigator.userAgent;
+    const devicetype = getDeviceType()
+    const publisher = setOnAny(validBidRequests, 'params.publisher');
+    const cur = CURRENCY;
+    // const endpointUrl = 'http://localhost:5000/prebid'
+    const endpointUrl = 'https://ai-i-codefuel-ds-rtb-us-east-1-k8s-internal.seccint.com/prebid'
+    const timeout = bidderRequest.timeout;
+
+    const imps = validBidRequests.map((bid, id) => {
+      bid.netRevenue = 'net';
```
tamirnPerion, line 43: I wouldn't recommend updating the **validBidRequests** obj inside the map function

ChrisHuie:
@hayimmakabee Can you please remove the modules.json file and put the undertone adapter change in a separate pr if that change is intended? Also, can you please add unit tests for this pr. 🙏 You probably don't want to update the integration example also. You could add you own integration example if you think it would help document and is different. However, adding to the main example I think only adds complexity.

modules/codefuelBidAdapter.js
hayimmakabee: I don't see where validBidRequests is updated, please clarify

modules/codefuelBidAdapter.js
hayimmakabee: Your code is not equivalent to the original code, because it may create fields with empty content. The original code does not create fields with empty content.

modules/codefuelBidAdapter.js
hayimmakabee: I understand that the concatenation assures that the resulting value will be a string.

modules/codefuelBidAdapter.js
tamirnPerion: Right. But it will be sent anyway to the RTB and parsed equivalently

modules/codefuelBidAdapter.js
tamirnPerion: Oh...
`id: String(id+1)`
then...

modules/codefuelBidAdapter.js
tamirnPerion: We run throuh the `validBidRequests` array and on each iteration we mutate the `netRevenue` prop. 
`bid.netRevenue = 'net';` (bid is a reference for `validBidRequests` array entry).
When using a map functions It's a better practice to use the returned object (a deep copy) and not mutating the original.
If we do want to mutate the `validBidRequests` array, I'd run a separte forEach loop

ChrisHuie:
Also reminder to submit documentation to our docs repo for this new adapter (https://github.com/prebid/prebid.github.io)

modules/codefuelBidAdapter.js
hayimmakabee: I see, please fix it.

ShaharHarel1:
@ChrisHuie Any other requirements?

ShaharHarel1:
@lksharma Any other requirements?

Issue #7238, opened 2021-07-29
getConfig() is returning reference to the config object


lksharma:
## Description
When adapters retrieve the config object by calling `config.getConfig()`, they are also able to manipulate the underlying object.

This is because the `_getConfig()` function uses `Object.assign()` when returning the configuration.

Since `Object. assign()` copies an object's reference value, we can get scenarios where one adapter updates the config and another adapter is affected. 

For example:

Step 1: Config set by `setConfig()` on the page

```
pbjs.setConfig({
    ortb2: {
        user: {
            gender: “M”
        }
    }
});
```

Step 2: Another adapter updates the config by first calling `config.getConfig()`

```
var fpd = getConfig(‘ortb2’)
fpd.user.gender = “F”
```

Step 3: Downstream adapters now retrieve the manipulated config object
```
var fpd = getConfig(‘ortb2’)
console.log(fpd)
// Output: {ortb2: { user: {gender: “F”}}}
```
Currently, the Sovrn adapter retrieves the `fpd` by calling `config.getConfig(‘ortb2’)`, and using that `fpd` object to update ortb2.user.ext for their request.

https://github.com/prebid/Prebid.js/blob/master/modules/sovrnBidAdapter.js#L86-L118

In the case where Sovrn is loaded before Index Exchange, the IX adapter will see the changes made by Sovrn

## Test Page
- Go to https://bastello.com/ix/prebid-fpd2.php
- Open the console in the dev tools

## Steps to Reproduce

1. Verify the config object set by `pbjs.setConfig()`
   - You can look at the page source for this (L49-L203)
2. Refresh the page and call the following function in the console
   - pbjs.getConfig('ortb2')
3. Verify the returned object is not the same as the one being set at the page level.
   - Call `pbjs.getConfig('ortb2')` in console
   - You will see the `eids` & `tpid` being added to the configuration by the sovrn adapter.

### Expected results
The `_getConfig()` should return a `deepClone` of the config object and not allow individual adapters to manipulate the `config` object. 

### Actual results
The `_getConfig()` function uses `Object.assign()` to return the stored config object, allowing adapter to manipulate data used by other downstream adapters.

## Platform details

Chrome: Version 91.0.4472.164 (Official Build) (x86_64)
Mac OS: Catalina; Version: 10.15.7
prebidJS: 5.6.0-pre

## Other information
Potential solution: PR https://github.com/prebid/Prebid.js/pull/7237

gglas:
[...]

PR #7349, opened 2021-08-25
Make getConfig call a clone prebid.js issue 7238

mikhalovich:
Cloned data returned from getConfig. Used for it `utils.deepClone`

For https://github.com/prebid/Prebid.js/issues/7238


patmmccann:
duplicate of #7237

PR #7364, opened 2021-08-30
IX Adapter: buildRequests refactor

lksharma:
## Type of change

- [ ] Bugfix
- [ ] Feature
- [ ] New bidder adapter   
- [ ] Build related changes
- [ ] CI related changes
- [ ] Other

## Description of change

Cleaning up the way our adapter creates impressions along with improving readability.

### DIFF
--- a/modules/ixBidAdapter.js
+++ b/modules/ixBidAdapter.js
@@ -4,6 +4,7 @@ import { config } from '../src/config.js';
 import find from 'core-js-pure/features/array/find.js';
 import { registerBidder } from '../src/adapters/bidderFactory.js';
 import { INSTREAM, OUTSTREAM } from '../src/video.js';
+import includes from 'core-js-pure/features/array/includes.js';
 
 const BIDDER_CODE = 'ix';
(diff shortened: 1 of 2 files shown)


modules/ixBidAdapter.js
```diff
@@ -948,58 +1011,43 @@ export const spec = {
...
+          break;
+      }
 
-    const DEFAULT_IX_CONFIG = {
-      detectMissingSizes: true,
-    };
+      if (adUnitMediaTypes.includes(BANNER) && adUnitMediaTypes.includes(VIDEO)) {
+        multiFormatAdUnits[validBidRequest.transactionId] = validBidRequest;
+      }
+    });
```
ChrisHuie, line 1036: Hey @lksharma . Since we still are using ie11 for testing and we support it we aren't able to use .includes right now (https://caniuse.com/array-includes). However, in the coming months with Prebid 6.0 release we are dropping ie support so should be able to use this in the near future

ChrisHuie (changes requested):

ChrisHuie (approved):

PR #7371, opened 2021-09-01
CriteoIdSystem: returns a callback to initiate user sync

afewcc:
## Type of change

- [ ] Bugfix
- [x] Feature
- [ ] New bidder adapter   
- [ ] Build related changes
- [ ] CI related changes
- [ ] Other

## Description of change

Changes in the CriteoIdSystem:
- Add support of the callback in the return of Submodule#getId, for initiating and resolving the user id sync.

## Contacts
- Tachfine El Belky (t.elbelky@criteo.com)

### DIFF
--- a/modules/criteoIdSystem.js
+++ b/modules/criteoIdSystem.js
@@ -6,8 +6,8 @@
  */
 
 import * as utils from '../src/utils.js'
-import * as ajax from '../src/ajax.js'
-import { getRefererInfo } from '../src/refererDetection.js'
+import { ajax } from '../src/ajax.js';
+import { getRefererInfo } from '../src/refererDetection.js';
 import { submodule } from '../src/hook.js';
 import { getStorageManager } from '../src/storageManager.js';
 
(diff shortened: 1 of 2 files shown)


Fawke:
Looks good but just want @smenzer to have a look. Scott do we have any fixed signature for the getId function and does it have any fixed return value. Here, the getId function is returning an object with two properties, id and result.

smenzer:
> Looks good but just want @smenzer to have a look. Scott do we have any fixed signature for the getId function and does it have any fixed return value. Here, the getId function is returning an object with two properties, id and result.

Based on the code here: https://github.com/prebid/Prebid.js/blob/35b5dca0a9c0e8bd05d1e5b9a5cb6b1541ae6935/modules/userId/index.js#L666-L677, `getId()` accepts either a callback function or an object. If it's an object, it sets the "id" to the value in the `response.id` field. As such, I think the code here is fine in that respect

smenzer (approved):
I only reviewed the part of the code related to the `getId()` response per @Fawke's request, but that part LGTM

Fawke (approved):

PR #7375, opened 2021-09-02
Net ID System: add support for 3rd-party webservice to get netId

phtechno:
## Type of change

- [x] Feature

## Description of change

* Add support to get the netid via 3rd-party service
* Add A/B Testing function

### DIFF
--- a/modules/netIdSystem.js
+++ b/modules/netIdSystem.js
@@ -55,7 +55,7 @@ export const netIdSubmodule = {
     if (!hasRequiredConfig(config)) {
       return undefined;
     }
-    if (config && config.hasOwnProperty("value") && config.value.hasOwnProperty("netId")) {
+    if (config && config.hasOwnProperty('value') && config.value.hasOwnProperty('netId')) {
       utils.logInfo(LOG_PREFIX + 'id already in config', config.value);
       return { id: config.value };
     }
@@ -73,7 +73,7 @@ export const netIdSubmodule = {
             try {
               responseRaw = JSON.parse(response);
               utils.logInfo(LOG_PREFIX + 'response received from the server', responseRaw);
-              if(responseRaw.hasOwnProperty("tpid") && responseRaw.hasOwnProperty("status_code") && responseRaw.status_code == "OK"){
+              if (responseRaw.hasOwnProperty('tpid') && responseRaw.hasOwnProperty('status_code') && responseRaw.status_code == 'OK') {
                 responseObj = { 'netId': responseRaw.tpid };
                 utils.logInfo(LOG_PREFIX + 'used tpid', responseObj);
               }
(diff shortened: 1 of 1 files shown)


ChrisHuie:
@phtechno can you please add unit tests for this pr 🙏

ncolletti:
@phtechno I noticed your Id Module is missing a markdown file, could you add that within this PR. You can look to other Id Module partners for what information should be included in that file.

modules/netIdSystem.js
```diff
@@ -18,22 +23,108 @@ export const netIdSubmodule = {
...
+
+  if (!config.storage || !config.storage.type || !config.storage.name) {
+    utils.logError(LOG_PREFIX + 'storage required to be set');
+    return false;
+  }
+
+  return true;
+};
+function checkAbTestingConfig(config, userId) {
+  if (config && config.params && config.params.abTesting && config.params.abTesting.enabled && config.params.abTesting.controlGroupPct) {
```
ncolletti, line 113: This could probably be replaced by using utils.deepAccess

modules/netIdSystem.js
```diff
@@ -18,22 +23,108 @@ export const netIdSubmodule = {
...
+      return { enabled: true, controlGroup: abTestBucket(userId) < config.params.abTesting.controlGroupPct * ABTEST_RESOLUTION }
+    }
+  }
+  return { enabled: false };
+};
+function abTestBucket(userId) {
+  if (userId) {
+    return ((utils.cyrb53Hash(userId) % ABTEST_RESOLUTION) + ABTEST_RESOLUTION) % ABTEST_RESOLUTION;
+  } else {
+    return Math.floor(Math.random() * ABTEST_RESOLUTION);
```
ncolletti, line 127: Will this condition be hit if prior to this function execution, within the decode function, you are returning undefined if 'netid' property of value type is not a string. Also, that check on line 31 should check for the existence of 'netid' on the value object as well.

modules/netIdSystem.js
```diff
@@ -18,22 +23,108 @@ export const netIdSubmodule = {
...
+          let responseObj;
+          let responseRaw;
+          if (response) {
+            try {
+              responseRaw = JSON.parse(response);
+              utils.logInfo(LOG_PREFIX + 'response received from the server', responseRaw);
+              if (responseRaw.hasOwnProperty('tpid') && responseRaw.hasOwnProperty('status_code') && responseRaw.status_code == 'OK') {
+                responseObj = { 'netId': responseRaw.tpid };
+                utils.logInfo(LOG_PREFIX + 'used tpid', responseObj);
+              }
```
ncolletti, line 79: Is there. a case that the status_code is not 'OK', what would happen then?

modules/netIdSystem.js
```diff
@@ -18,22 +23,108 @@ export const netIdSubmodule = {
...
+  decode(value, config) {
+    let netId;
+    if (value && typeof value['netId'] === 'string') {
+      netId = { 'netId': value['netId'] };
+    } else {
+      return undefined;
+    }
+    // check for A/B testing configuration and hide ID if in Control Group
+    const abConfig = checkAbTestingConfig(config, netId);
+    if (abConfig.enabled === true && abConfig.controlGroup === true) {
```
ncolletti, line 38: Can the enabled property be checked to be true in `checkAbTestingConfig` instead of happening twice here?

PR #7394, opened 2021-09-09
Sonobi Bid Adapter: remove outstream bid.ad

JonGoSonobi:
## Type of change

- [x] Bugfix

## Description of change
A publisher requested that we remove the bid.ad value for outstream since we provide the vastUrl

- apex.prebid@sonobi.com


ChrisHuie (approved):

patmmccann:
This appears to be breaking; should it be behind config?

Issue #7393, opened 2021-09-09
Proposal to add timeout RTD module


luigi-sayson:
## Type of issue

Proposal

## Description

Overview
Publishers have the difficult task of constantly balancing monetization and user experience. From Prebid’s own documentation on setting auction timeout: “...Determining this value is a delicate balance: too short, and header bidding won’t have enough time to take place; too long, and revenue may go down due to delaying the ad server call to the point where users have left the page. Publishers must determine the value that works for them, considering a balance of factors: average user time on page, direct sellthrough, value of different ad channels, and average user network delay.”

With the number of channels, devices, and ad formats ever expanding in programmatic, creating an optimal auction timeout strategy has become more difficult than ever. Most publishers simply choose a single setting for all their inventory and rarely, if ever, change it. A/B testing can help identify a better timeout setting, but that takes precious time and resources. Several publishers host multiple Prebid configurations to set different timeouts depending on the context of the request. However, context that could be critical to selecting the optimal auction timeout can change throughout the day (e.g. time of day), making even this strategy sub-optimal. 

Given this situation, we are proposing a module that would allow a publisher to sign up with an auction timeout provider that would be able to set your auction timeout in real-time based on the context of the request and is continually optimized using machine learning. In addition, if a publisher prefers not to work with a provider, they could use the module to enact a rules-based approach that changes the auction timeout based on the contextual signals of their choice.

Auction Timeout Provider Intended Workflow 

1. Publisher signs up with an auction timeout provider
2. Auction timeout provider will provide the publisher with a url to retrieve the timeout data
3. Publisher configures Prebid.js to resolve and enforce the timeout

This feature should be designed to work with Prebid.js currently, but support for Prebid Server, and Prebid SDK could come later.

Goals

- Provide a Prebid unified auction timeout solution controlled by the publisher
- Allow for an open framework to set auction timeout manually or to work with a provider
- Reduce time spent by publishers on manually testing strategies to set more optimal timeout values
- Encourage other providers to propose models that effectively optimize a publisher timeout for revenue and user experience

## Other information

I've uploaded the code that implements this rtd module in a branch here: https://github.com/prebid/Prebid.js/compare/master...openx:timeout-rtd-module

PR #7395, opened 2021-09-09
Timeout RTD module: initial release

luigi-sayson:
## Type of change

- [x] Feature

## Description of change

Adding a new RTD module to manage timeouts in real time. The current implementation supports adding time based on rules that look at specific features (number of ad units in the request, device type, connection speed, and whether or not a video ad unit is present). These rules can either be retrieved from a timeout provider, or configured statically on page. 

## Other information
Please see related proposal issue: https://github.com/prebid/Prebid.js/issues/7393

### DIFF
--- a/modules/timeoutRtdProvider.js
+++ b/modules/timeoutRtdProvider.js
@@ -1,0 +1,173 @@
+
+import { submodule } from '../src/hook.js';
+import * as ajax from '../src/ajax.js';
+import * as utils from '../src/utils.js';
+import { getGlobal } from '../src/prebidGlobal.js';
+
+const SUBMODULE_NAME = 'timeout';
+
+// this allows the stubbing of functions during testing
+export const timeoutRtdFunctions = {
+  getDeviceType,
+  getConnectionSpeed,
+  checkVideo,
+  calculateTimeoutModifier,
+  handleTimeoutIncrement
+};
+
+function getDeviceType() {
+  const userAgent = window.navigator.userAgent.toLowerCase();
+  if ((/ipad|android 3.0|xoom|sch-i800|playbook|tablet|kindle/i.test(userAgent))) {
+    return 5; // tablet
+  }
+  if ((/iphone|ipod|android|blackberry|opera|mini|windows\sce|palm|smartphone|iemobile/i.test(userAgent))) {
+    return 4; // mobile
+  }
+  return 2; // personal computer
+}
+
+function checkVideo(adUnits) {
+  return adUnits.some((adUnit) => {
+    return adUnit.mediaTypes && adUnit.mediaTypes.video;
+  });
+}
+
+function getConnectionSpeed() {
+  const connection = window.navigator.connection || window.navigator.mozConnection || window.navigator.webkitConnection || {}
+  const connectionType = connection.type || connection.effectiveType;
+
+  switch (connectionType) {
+    case 'slow-2g':
+    case '2g':
+      return 'slow';
+
+    case '3g':
+      return 'medium';
+
+    case 'bluetooth':
+    case 'cellular':
+    case 'ethernet':
+    case 'wifi':
+    case 'wimax':
+    case '4g':
+      return 'fast';
+  }
+
+  return 'unknown';
+}
+/**
+ * Calculate the time to be added to the timeout
+ * @param {Array} adUnits
+ * @param {Object} rules
+ * @return {int}
+ */
+function calculateTimeoutModifier(adUnits, rules) {
+  utils.logInfo('Timeout rules', rules);
+  let timeoutModifier = 0;
+  let toAdd = 0;
+
+  if (rules.includesVideo) {
+    const hasVideo = timeoutRtdFunctions.checkVideo(adUnits);
+    toAdd = rules.includesVideo[hasVideo] || 0;
+    utils.logInfo(`Adding ${toAdd} to timeout for includesVideo ${hasVideo}`)
+    timeoutModifier += toAdd;
+  }
+
+  if (rules.numAdUnits) {
+    const numAdUnits = adUnits.length;
+    if (rules.numAdUnits[numAdUnits]) {
+      timeoutModifier += rules.numAdUnits[numAdUnits];
+    } else {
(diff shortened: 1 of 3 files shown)


ChrisHuie:
@luigi-sayson looks like there are a few linting errors in this pr. Can you please fix?

luigi-sayson:
@ChrisHuie, done. Sorry, I ran `gulp test-coverage` last and didn't realize that it wasn't running lint

patmmccann:
Connection speed slow/medium/fast almost always resolves to fast or missing. We see 4g for effective type on north of 90% of traffic with it filled in. This isn't supported in Safari, so I am curious to see added documentation on specifying the timeout on missing.

luigi-sayson:
The current implementation should already handle a rule for 'unknown' connection speeds. I just forgot to add it to the documentation, which I've done now. 

> We see 4g for effective type on north of 90% of traffic with it filled in. 

Yeah the idea behind the connection speed is that most pubs should choose 'fast', but maybe some operate in a region with slower connection speeds and would want to adjust their timeouts accordingly

modules/timeoutRtdProvider.md
```diff
@@ -0,0 +1,137 @@
...
+ module_type: rtd
+ module_code : example
+ enable_download : true
+ sidebarType : 1
+ ---
+
+## Overview
+The timeout RTD module enables publishers to dynamically retrieve 
+timeout rules from a timeout provider that determine the timeout based 
+on certain features, or set those rules statically on the page.
```
patmmccann, line 15: can you make it clear here that the contents of the response can also be set via config and no endpoint is required?

modules/timeoutRtdProvider.md
```diff
@@ -0,0 +1,137 @@
...
+            "slow": 200,
+            "medium": 100,
+            "fast": 50,
+            "unknown": 10
+        },
+}
+```
+
+### Rule Handling:
+The rules retrieved from the endpoint will be used to add time to the `bidderTimeout` based on certain features such as
```
patmmccann, line 79: 'The rules retrieved from the endpoint or set via setConfig'

modules/timeoutRtdProvider.md
```diff
@@ -0,0 +1,137 @@
...
+
+Currently supported features:
+
+|Name |Description | Keys | Example
+| :------------ | :------------ | :------------ |:------------ |
+| includesVideo | Adds time to the timeout based on whether there is a video ad unit in the auction or not | 'true'/'false'| { "true": 200, "false": 50 } | 
+| numAdUnits | Adds time based on the number of ad units. Ranges in the format `'lowerbound-upperbound` are accepted. This range is inclusive | numbers or number ranges | {"1": 50, "2-5": 100, "6-10": 200} |  
+| deviceType | Adds time based on device type| 2, 4, or 5| {"2": 50, "4": 100} |
+| connectionSpeed | Adds time based on connection speed. `connectionSpeed` defaults to 'unknown' if connection speed cannot be determined | slow, medium, fast, or unknown | { "slow": 200} |
+
```
patmmccann, line 100: Can you add some detail about if there are mulitple rule matches, will the first or all of the increments be added to the baseline timeout?

test/spec/modules/timeoutRtdProvider_spec.js
```diff
@@ -0,0 +1,316 @@
...
+    const handleTimeoutIncrementStub = sandbox.stub(timeoutRtdFunctions, 'handleTimeoutIncrement');
+
+    timeoutSubmodule.getBidRequestData({}, callback, {});
+
+    expect(handleTimeoutIncrementStub.called).to.be.false;
+    expect(callback.called).to.be.true;
+    expect(ajaxStub.called).to.be.false;
+  });
+
+  it('should be able to increment the timeout with the calculated timeout modifier', () => {
```
patmmccann, line 296: can you add a unit test for the multiple rules match case?

patmmccann (changes requested):

modules/timeoutRtdProvider.js
```diff
@@ -0,0 +1,173 @@
...
+  }
+  if ((/iphone|ipod|android|blackberry|opera|mini|windows\sce|palm|smartphone|iemobile/i.test(userAgent))) {
+    return 4; // mobile
+  }
+  return 2; // personal computer
+}
+
+function checkVideo(adUnits) {
+  return adUnits.some((adUnit) => {
+    return adUnit.mediaTypes && adUnit.mediaTypes.video;
```
patmmccann, line 31: might be useful to consider outstream multiformat units differently?

patmmccann (changes requested):
Mostly documentation changes, one additional unit test

modules/timeoutRtdProvider.md
patmmccann: actually you do say this, but perhaps reorder the sentence a small bit

`The timeout RTD module enables publishers to set rules  that determine the timeout based 
on certain features. It supports rule sets dynamically retrieved from a timeout provider as well as rules set directly via configuration`

test/spec/modules/timeoutRtdProvider_spec.js
luigi-sayson: This should be handled by this test case: https://github.com/prebid/Prebid.js/pull/7395/files#diff-49900727474e17b1862be0b57a305580c07d69ca22e2e4c924a78d0630631735R173-R205, though I added one anyway, since this new one also checks if the timeout is incremented properly when there's multiple matching rules

modules/timeoutRtdProvider.js
luigi-sayson: Yeah, though I think we'll just save that for a future version of this module

patmmccann:
LGTM ; please make a docs repo PR before we merge

patmmccann (approved):

luigi-sayson:
👍 
Done: https://github.com/prebid/prebid.github.io/pull/3266

patmmccann:
@ChrisHuie ; as this is a general module, I should have looked for second approval. Could you take a look as well?

PR #7396, opened 2021-09-09
Prebid Core: mergeConfig and mergeBidderConfig functions

### LINKED ISSUE #6425: Add mergeConfig functions for updating ortb2 data
## Type of issue
enhancement

## Description
There are several parties that might want to add to the new First Party Data structure. Currently, Real Time Data modules need to getConfig({ortb2}), modify it, then setConfig({ortb2}). This is awkward.

This issue proposes two convenience functions to make it easier for RTD modules (and page code) to manage important data structures. So while the main use case is `ortb2`, the proposal is that these new functions would apply to all setConfig entries.

1) mergeConfig({CONFIG})

```
pbjs.mergeConfig({
  ortb2: {
    user: {
      data: [
         {
          "name": "www.dataprovider1.com",
          "ext": { "taxonomyname": "iab_audience_taxonomy" },
          "segment": [{ "id": "687" }, { "id": "123" }]
         }
      ]
    }
  }
});
```
This command would add the specified ortb2.user.data array element to the existing ortb2 data structure or create if it doesn't already exist.

Requirements:
1. mergeConfig should create the object if it doesn't already exist
2. mergeConfig should add entries to the end of the following arrays: ortb2.site.keywords, ortb2.user.keywords, ortb2.user.data and ortb2.site.content.data arrays rather than overwriting them. The function does not have to check for duplicate array entries. [Are there other arrays we want to add to rather than overwriting?]
3. Otherwise mergeConfig should give precedence to the new values over existing values. [I believe this is the current convention given regular JavaScript Object.assign() semantics]

2) mergeBidderConfig({bidder: [BIDDERS], config: CONFIG})

```
pbjs.mergeBidderConfig({
  bidders: ["bidderA", "bidderB"],
  config: {
    ortb2: {
      user: {
        data: [
           {
            "name": "www.dataprovider1.com",
            "ext": { "taxonomyname": "iab_audience_taxonomy" },
            "segment": [{ "id": "687" }, { "id": "123" }]
           }
        ]
      }
    }
  }
});
```
This command would add the specified ortb2.user.data array element to the existing ortb2 data structure for bidderA and bidderB or create if it doesn't already exist.

Same requirements as `mergeConfig` noted above.

patrickloughrey:
## Type of change
- [x] Feature

## Description of change
Added mergeConfig and mergeBidderConfig functions and associated unit tests that are intended to merge the input config with the existing global config which will replace the current method (running getConfig(), copying the config and then running setConfig(config)).

## Other information
Fixes #6425


patmmccann:
@anthonylauzon @karimMourra we would appreciate your comments

patmmccann:
Please remove package lock changes

src/config.js
```diff
@@ -578,7 +558,7 @@ export function newConfig() {
       .forEach(listener => listener.callback(options));
   }
 
-  function setBidderConfig(config) {
+  function setBidderConfig(config, flag = false) {
```
karimMourra, line 561: I suggest using a more descriptive name for flag. I think 'deep' or 'deepMerge' would describe the intent.

src/config.js
```diff
@@ -590,7 +570,8 @@ export function newConfig() {
           let option = (topic === 'fpd') ? convertFpd(config.config[topic]) : config.config[topic];
 
           if (utils.isPlainObject(option)) {
-            bidderConfig[bidder][prop] = Object.assign({}, bidderConfig[bidder][prop] || {}, option);
+            const func = flag ? utils.mergeDeep : Object.assign;
```
karimMourra, line 573: Is there a case where we would want the merge to be shallow ? If not then why not always use `utils.mergeDeep` ?

src/config.js
```diff
@@ -612,6 +594,26 @@ export function newConfig() {
     }
   }
 
+  function mergeConfig(obj) {
+    if (!utils.isPlainObject(obj)) {
+      utils.logError('mergeConfig input must be an object');
+      return;
+    }
+
+    const mergedConfig = Object.keys(obj).reduce((accum, data) => {
```
karimMourra, line 603: 'data' might be a misleading term. I think 'key' or 'property' would be more accurate. 
'data' would be a better name for the property value.

src/config.js
```diff
@@ -612,6 +594,26 @@ export function newConfig() {
...
+      const prevConf = _getConfig(data)[data] || {};
+      accum[data] = utils.mergeDeep(prevConf, obj[data]);
+      return accum;
+    }, {});
+
+    setConfig({ ...mergedConfig });
```
karimMourra, line 609: spread operators are not supported in IE, unfortunately. https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Spread_syntax

karimMourra (changes requested):

src/config.js
patmmccann: I think the plan is to transpile the spread operator for some time.

anthonylauzon (changes requested):
this one is a little tricky. since there are potentially multiple calls to RTD endpoints during the lifecycle of a visit, this can lead to data duplication which the bid adapters would then have to filter out when looking at the data.  we worked around this by using a modified mergeDeep, which you can reference in the halo rtd module:  
https://github.com/prebid/Prebid.js/blob/master/modules/haloRtdProvider.js

here's a test case ported from https://github.com/prebid/Prebid.js/blob/master/test/spec/modules/haloRtdProvider_spec.js

it should pass or RTD module maintainers like ourselves will probably insist on using our own custom-baked solutions.  there may be some other edge cases that need to be addressed as well before this PR is ready to be used for heavy lifting across all modules and in all use cases.
```
  it('should merge without array duplication', function() {
    let rtdConfig = {};
    let bidConfig = {};

    const userObj1 = {
      name: 'www.dataprovider1.com',
      ext: { taxonomyname: 'iab_audience_taxonomy' },
      segment: [{
        id: '1776'
      }]
    };

    const userObj2 = {
      name: 'www.dataprovider2.com',
      ext: { taxonomyname: 'iab_audience_taxonomy' },
      segment: [{
        id: '1914'
      }]
    };

    const siteObj1 = {
      name: 'www.dataprovider3.com',
      ext: {
        taxonomyname: 'iab_audience_taxonomy'
      },
      segment: [
        {
          id: '1812'
        },
        {
          id: '1955'
        }
      ]
    }

    setConfig({
      ortb2: {
        user: {
          data: [userObj1, userObj2]
        },
        site: {
          content: {
            data: [siteObj1]
          }
        }
      }
    });

    const rtd = {
      ortb2: {
        user: {
          data: [userObj1]
        },
        site: {
          content: {
            data: [siteObj1]
          }
        }
      }
    };
    mergeConfig(rtd);

    let ortb2Config = getConfig().ortb2;

    expect(ortb2Config.user.data).to.deep.include.members([userObj1, userObj2]);
    expect(ortb2Config.site.content.data).to.deep.include.members([siteObj1]);
    expect(ortb2Config.user.data).to.have.lengthOf(2);
    expect(ortb2Config.site.content.data).to.have.lengthOf(1);
  });
```

src/config.js
mmoschovas: Agreed

src/config.js
patrickloughrey: Is the consensus that I should replace the spread operator or is it being transpiled? @patmmccann

src/config.js
patrickloughrey: Updated 'data' to 'key', you're right 'key' is a bit clearer

src/config.js
patrickloughrey: the goal of this line is to determine whether we should be merging the bidderConfig or setting it; so if we are not running a mergeDeep then we are simply setting the bidderConfig (not performing a merge)

src/config.js
patrickloughrey: I landed on 'mergeFlag'. 'deepMerge' seems a bit confusing since 'mergeDeep' is on the same line. Let me know your thoughts. If you don't like 'mergeFlag' then I think 'merge' makes more sense than 'deep'.

test/spec/config_spec.js
```diff
@@ -345,4 +286,212 @@ describe('config API', function () {
...
+            domain: 'page.example.com',
+            cat: ['IAB2'],
+            sectioncat: ['IAB2-2'],
+            pagecat: ['IAB2-2']
+          },
+          user: {
+            ext: {
+              ssp: 'magnite',
+              data: {
+                registered: false,
```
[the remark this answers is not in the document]
patrickloughrey: No, this will be false. The input should be overriding what exists in the current state of the config.

src/config.js
mmoschovas: I think mergeFlag is fine for this. One thing to note is that along with adding mergeConfig/mergeBidderConfig to the docs, we may want to make updates to setBidderConfig as it now includes two arguments

mmoschovas:
These tests are still failing here. I reran a few times but it seems consistent. Also conflicts came up

ChrisHuie:
@patrickloughrey @mmoschovas pulling in recent commits should fix these errors. We had the window size change in browerstack or something so took us a few days to figure out and fix.

robertrmartinez:
I think it might be beneficial to add in an external mergeConfig as well.

https://github.com/prebid/Prebid.js/blob/master/src/prebid.js#L902-L903

ChrisHuie:
> I think it might be beneficial to add in an external mergeConfig as well.
> 
> https://github.com/prebid/Prebid.js/blob/master/src/prebid.js#L902-L903

Added this to the pr

ChrisHuie (approved):

src/utils.js
```diff
@@ -1272,7 +1272,18 @@ export function mergeDeep(target, ...sources) {
         if (!target[key]) {
           Object.assign(target, { [key]: source[key] });
         } else if (isArray(target[key])) {
-          target[key] = target[key].concat(source[key]);
+          source[key].forEach(obj => {
+            let e = 1;
```
karimMourra, line 1276: what is e ?

bretg: renamed to 'addItFlag'

karimMourra (approved):

patmmccann:
@ChrisHuie where are the docs? what about this alert? https://github.com/prebid/Prebid.js/pull/7396#issuecomment-990218989

ChrisHuie:
https://github.com/prebid/prebid.github.io/pull/3470

PR #7401, opened 2021-09-10
Revert "Merkle Id System: configurable endpoint"

patmmccann:
Reverts prebid/Prebid.js#7400


patmmccann:
This was a breaking change

PR #7419, opened 2021-09-15
PubMatic bid adapter: while retrieving floor from floor module pass banner-sizes instead of *

pm-harshad-mane:
## Type of change
- [X] Feature

## Description of change
While retrieving floor value from Price floor module we are passing *.
For Banner, retrieve floor for each size supported by ad-unit and then use the lowest value as floor from Floor module.
For other formats pass *
If no sizes are found then pass *

- contact email of the adapter’s maintainer: @pm-harshad-mane 


pm-harshad-mane:
Hello @mike-chowla , @robertrmartinez 
Could you please review the changes? We want this patch to be part of the next release, 5.14.

modules/pubmaticBidAdapter.js
```diff
@@ -740,22 +740,48 @@ function _addFloorFromFloorModule(impObj, bid) {
...
+            sizesArray.push([impObj[mediaType].w, impObj[mediaType].h]);
+          }
+          if (utils.isArray(impObj[mediaType].format)) {
+            impObj[mediaType].format.forEach(size => sizesArray.push([size.w, size.h]));
+          }
+        }
+
+        if (sizesArray.length === 0) {
+          sizesArray.push('*')
         }
```
robertrmartinez, line 756: if mediaType = banner => use all sizes

else => use *

👍

modules/pubmaticBidAdapter.js
```diff
@@ -740,22 +740,48 @@ function _addFloorFromFloorModule(impObj, bid) {
...
+
+        sizesArray.forEach(size => {
+          let floorInfo = bid.getFloor({ currency: impObj.bidfloorcur, mediaType: mediaType, size: size });
+          utils.logInfo(LOG_WARN_PREFIX, 'floor from floor module returned for mediatype:', mediaType, ' and size:', size, ' is: currency', floorInfo.currency, 'floor', floorInfo.floor);
```
robertrmartinez, line 760: with newer versions of prebid, the Bider Code is prefixed in the console, so we do not need our prefix's anymore

(not needed to change this now just something I have to change in our adapter as well!)

modules/pubmaticBidAdapter.js
```diff
@@ -740,22 +740,48 @@ function _addFloorFromFloorModule(impObj, bid) {
...
+        });
       }
     });
   }
   // get highest from impObj.bidfllor and floor from floor module
   // as we are using Math.max, it is ok if we have not got any floor from floorModule, then value of bidFloor will be -1
   if (impObj.bidfloor) {
+    utils.logInfo(LOG_WARN_PREFIX, 'floor from floor module:', bidFloor, 'impObj.bidfloor', impObj.bidfloor, 'Max:', Math.max(bidFloor, impObj.bidfloor));
     bidFloor = Math.max(bidFloor, impObj.bidfloor)
   }
```
robertrmartinez, line 780: So we take the max of (pub provided bidder param `kadfloor`) and the (minimum floor from all sizes of floors module)

I can't remember if the intention of the 5.0 was just "read the floors module if you use floor signaling" or "if floors module only use that"

Prob the former so I think this PR is fine. Just a note!

robertrmartinez (approved):
Looks good to me

PR #7428, opened 2021-09-16
Kargo Bid Adapter: Support video mediaType

jsadwith:
## Type of change
- [x] Feature

## Description of change

* Adds the video module

---


Fawke:
Hi @jsadwith,

I guess it's fine as long as you're passing `mediaTypes.video` object to your bidder, which you're doing as you're passing the entire raw object.

Coming to the second point, yes, we need a test bid. Read the PR Review [guidelines](https://github.com/prebid/Prebid.js/blob/master/PR_REVIEW.md#reviewing-a-new-or-updated-bid-adapter).

jsadwith:
Thanks @ChrisHuie - just got that updated.

@Fawke we should have the server-side portion ready in the next few business days. Once it's live, I'll include the video test for your review.

jsadwith:
@Fawke I've updated the md file with a new placementId for video that will return a response from the server. One important thing to note is that you will need to use mobile emulation with a mobile user agent in order to get a response. I'm attaching a screenshot below of what you should see for the response. Can you let me know if this gets you what you need?
image

Fawke (approved):
LGTM

PR #7444, opened 2021-09-20
Inskin Bid Adapter: override schain with publisher id

cciocov:
## Type of change

- [ ] Bugfix
- [x] Feature
- [ ] New bidder adapter   
- [ ] Build related changes
- [ ] CI related changes
- [ ] Other

## Description of change
We are sending a new parameter when making the ad call to our ad server.

### DIFF
--- a/test/spec/modules/inskinBidAdapter_spec.js
+++ b/test/spec/modules/inskinBidAdapter_spec.js
@@ -12,7 +12,8 @@ const REQUEST = {
     'bidder': 'inskin',
     'params': {
       'networkId': 9874,
-      'siteId': 730181
+      'siteId': 730181,
+      'publisherId': 123456
     },
     'placementCode': 'div-gpt-ad-1487778092495-0',
     'sizes': [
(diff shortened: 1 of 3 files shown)


patmmccann:
You want to override the supply chain objects publishers are setting in config?

ncolletti:
Is the correct place for schain data in `source.ext.schain`? Also, should you be choosing to pass the `schain` object from `validBidRequests` first, before looking to write your own from user defined adapter params?

cciocov:
@patmmccann for the moment, we are not taking into account the schain config set in prebid, we are only concerned with passing what's defined in publisherId in our adapter config.

@ncolletti at the moment we only want to pass our own defined settings, we'll consider the rest in a future update, thanks!

test/spec/modules/inskinBidAdapter_spec.js
```diff
@@ -12,7 +12,8 @@ const REQUEST = {
     'bidder': 'inskin',
     'params': {
       'networkId': 9874,
-      'siteId': 730181
+      'siteId': 730181,
+      'publisherId': 123456
```
ncolletti, line 16: Could you include this param in your markdown file

cciocov: @ncolletti This has been done, thanks!

ncolletti (approved):

PR #7448, opened 2021-09-21
Saamba Bid Adapter: support  video params, and the floors module

NeerajKrRai:
## Type of change

- [ ] Bugfix
- [ ] Feature

## Description of change
#6512
Support video params at the adunit level

#6465
support the floors module

- contact email of the adapter’s maintainer-matt.voigt@saambaa.com,l@advangelists.com


NeerajKrRai:
Hi @ChrisHuie, just wanted to check on this.

Please let me know if anything is pending from our side.

Thanks,
Neeraj

modules/saambaaBidAdapter.js
```diff
@@ -0,0 +1,414 @@
+import * as utils from '../src/utils.js';
```
ChrisHuie, line 1: We have changed all adapters so now they are only importing the functions they use. Can you replace this import with just the ones you are using like `isFn` and `deepAccess`?

ChrisHuie (approved):

PR #7466, opened 2021-09-23
Pbjs Core: avoid import all from utils in src/ files

pm-harshad-mane:
## Type of change

## Description of change
Some files import the utils all methods. It is recommended to import only the required functions. So that we can take advantage of dead-code-elimination when we upgrade to WebPack 5.
The changes are made only for the files in src folder with exception of `events.js`, need to handle it separately. 
We can follow this practice in PR review to check whether only required functions are imported from modules in current module and not import the the whole module.


ChrisHuie (approved):

pm-harshad-mane:
Hello @robertrmartinez ,
Could you please review these changes?

robertrmartinez (approved):
Nice!

Issue #7402, opened 2021-09-10
Begin dchain implementation


patmmccann:
## Type of issue
Feature Request

## Description
Several adapters currently seem to support passing a dspid, eg AppNexus, Index (https://github.com/prebid/Prebid.js/pull/3585), Pubmatic (https://github.com/prebid/Prebid.js/pull/3619/files). 

On #3115 this was decided to be placed into a particular place in the meta object. 

Since that time, a published spec for the dchain object has come to fruition https://iabtechlab.com/buyers-json-demand-chain/. 

I propose a pull request, for the three adapters above to start with, that follows this format, on the dchain meta object. 

"dchain" : {
 "ver": "1.0",
 "complete" : 0,
 "nodes" : [
 {
 "bsid": "12345"
 },
 {
 "asi": "sspname",
 "bsid": "gvlid_for_ssp"
 }
 ]
}

and for several others that do not pass dsp id in this format

"dchain" : {
 "ver": "1.0",
 "complete" : 0,
 "nodes" : [
 {
 "asi": "sspname",
 "bsid": "gvlid_for_ssp"
 }
 ]
}

My goal is that dchains start to proliferate, publishers start to demand (pun intended) them, and adapter owners extend them, particularly as they begin to publish their own buyers.json files. 

## Other information

Related issues:
#6923 
#6380
#3115

patmmccann:
@khatibda fyi

patmmccann:
@jsnellbaker @ix-prebid-support @ix-prebid-development @pm-harshad-mane @mike-chowla for viz.

jsnellbaker:
Hi @patmmccann 

For the appnexus bidder, what field in the bid response where you looking at that seemed to correlate to this dchain field?  I may have missed something, but I'm not sure we're supplying this data at this time (at least for our client-side requests).

patmmccann:
@jsnellbaker I imagined for appnexus and its aliases the object would look like this


"dchain" : {
"ver": "1.0",
"complete" : 0,
"nodes" : [
{
"bsid": "BUYERMEMBERID"
},
{
"asi": "BIDDERNAME",
"bsid": "BIDDER_GVLID"
}
]
}

jsnellbaker:
I will check with some folks internally, but I'm pretty sure it's not currently supported.

patmmccann:
@jsnellbaker I see BUYERMEMBERID on your bid responses. Do you mean that our set up isn't typical?

[...]

PR #7473, opened 2021-09-24
Dchain module: initial release & Appnexus bid adapter: partial dchain support 

jsnellbaker:
## Type of change

- [x] Feature

## Description of change
Related to #7402 (see this link for details for functionality - https://github.com/prebid/Prebid.js/issues/7402#issuecomment-921834719)

This PR adds the dchain module and adds partial support for the dchain object in the appnexus bid adapter.

To note - I'm not entirely 100% about some of the fields that were defined in the dchain module (such as for the back-up dchain object [here](https://github.com/prebid/Prebid.js/pull/7473/files#diff-629170f5bcfe7e0088b6ee7899e1254ed4e21270567fc8fa0e2f69c4d33fb114R112) or for the appnexus dchain object.  If they should be different, please let me know and I'll correct them.

IAB Dchain Spec here:
https://iabtechlab.com/wp-content/uploads/2021/06/DemandChainObject-1.0-June2021.pdf


test/spec/modules/appnexusBidAdapter_spec.js
```diff
@@ -1076,6 +1076,16 @@ describe('AppNexusAdapter', function () {
           'adUnitCode': 'code',
           'appnexus': {
             'buyerMemberId': 958
+          },
+          'meta': {
+            'dchain': {
+              'ver': '1.0',
+              'complete': 0,
+              'nodes': [{
+                'bsid': '958',
```
patmmccann, line 1085: I think this dchain would have two nodes, one for xandr as an ssp and one for the buyer seat id. The final node would be for Xandr alone, and the second to final node would have the bsid alone.

modules/dchain.js
```diff
@@ -0,0 +1,144 @@
...
+        logWarn('bid.meta.dchain.nodes did not exist or was not an array; did not append prebid dchain.', bid);
+      }
+    } else {
+      // bid.meta.dchain = basicDchain;     // should we assign a backup dchain if bidder's dchain was invalid?
+      delete bid.meta.dchain; // or delete the bad object?
+    }
+  } else {
+    bid.meta.dchain = basicDchain;
+  }
+
```
patmmccann, line 136: it seems like if there is no meta.dchain, but there is meta.networkid or networkname; one would expect two nodes: 

{
    ver: '1.0',
    complete: 0,
    nodes: [{ name: bid.meta.networkname, bsid: bid.meta.networkid }, { name: bid.bidderCode }]
  };

On #3115 and in doc changes to https://docs.prebid.org/dev-docs/bidder-adaptor.html these fields were standardized and adopted in some adapters eg https://github.com/prebid/Prebid.js/blob/a5141af2f738be6980f19843315e5953da501ad7/modules/pubmaticBidAdapter.js#L1257 and at https://github.com/prebid/Prebid.js/blob/a5141af2f738be6980f19843315e5953da501ad7/modules/ixBidAdapter.js#L282

patmmccann (changes requested):

test/spec/modules/appnexusBidAdapter_spec.js
jsnellbaker: Hi @patmmccann 

Just to clarify - is the 'xandr' node you're referring to represent the same node as the 'bidder' node that the dchain module appends later when the bid is processed?  Or is meant to represent something different for the xandr adserver system?

If our bidder creates two nodes now, first for the bsid and the second for the xandr name, eventually there will be a third node attached with the appnexus bidder name (attached by the module's workflow); is that the ideal chain that should be made?

modules/dchain.js
jsnellbaker: I'll put together some updates to look for these bid.meta fields and use them if they're present.  Thanks!

jsnellbaker: Pushed a change into the PR.

patmmccann:
@jsnellbaker it seems the brandid merge created a conflict

jsnellbaker:
Thanks @patmmccann for the heads-up.  I have resolved the conflict.

modules/dchain.md
```diff
@@ -0,0 +1,45 @@
...
+```
+    $ gulp build --modules=dchain,pubmaticBidAdapter,openxBidAdapter,rubiconBidAdapter,sovrnBidAdapter
+```
+
+The dchain module will validate a bidder's dchain object (if it was defined).  Bidders should assign their dchain object into `bid.meta` field.  If the dchain object is valid, it will remain in the bid object for later use.
+
+If it was not defined, the dchain will create a default dchain object for prebid.
+
+## Validation modes
+- ```strict```: It is the default validation mode. In this mode, dchain object will not be accpeted from adapters if it is invalid. Errors are thrown for invalid dchain object.
```
patmmccann, line 43: accpeted typo

PR #7494, opened 2021-09-27
instream video support added

sourabhg:
## Type of change

- [X] Feature

## Description of change
Video Support Added for Deepintent Adapter

https://github.com/prebid/prebid.github.io/pull/3294

### DIFF
--- a/modules/deepintentBidAdapter.js
+++ b/modules/deepintentBidAdapter.js
@@ -1,10 +1,35 @@
 import {registerBidder} from '../src/adapters/bidderFactory.js';
-import {BANNER} from '../src/mediaTypes.js';
+import {BANNER, VIDEO} from '../src/mediaTypes.js';
 import * as utils from '../src/utils.js';
 const BIDDER_CODE = 'deepintent';
 const BIDDER_ENDPOINT = 'https://prebid.deepintent.com/prebid';
 const USER_SYNC_URL = 'https://cdn.deepintent.com/syncpixel.html';
 const DI_M_V = '1.0.0';
+export const ORTB_VIDEO_PARAMS = {
+  'mimes': (value) => Array.isArray(value) && value.length > 0 && value.every(v => typeof v === 'string'),
+  'minduration': (value) => utils.isInteger(value),
+  'maxduration': (value) => utils.isInteger(value),
+  'protocols': (value) => Array.isArray(value) && value.every(v => [1, 2, 3, 4, 5, 6, 7, 8, 9, 10].indexOf(v) !== -1),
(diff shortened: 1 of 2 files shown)


modules/deepintentBidAdapter.js
```diff
@@ -1,10 +1,35 @@
...
 import * as utils from '../src/utils.js';
 const BIDDER_CODE = 'deepintent';
 const BIDDER_ENDPOINT = 'https://prebid.deepintent.com/prebid';
 const USER_SYNC_URL = 'https://cdn.deepintent.com/syncpixel.html';
 const DI_M_V = '1.0.0';
+export const ORTB_VIDEO_PARAMS = {
+  'mimes': (value) => Array.isArray(value) && value.length > 0 && value.every(v => typeof v === 'string'),
+  'minduration': (value) => utils.isInteger(value),
+  'maxduration': (value) => utils.isInteger(value),
+  'protocols': (value) => Array.isArray(value) && value.every(v => [1, 2, 3, 4, 5, 6, 7, 8, 9, 10].indexOf(v) !== -1),
```
msm0504, line 12: In these validation functions, why match the value against an array instead of comparing to the min and max?
```
v >= 1 && v <= 10
```
At this time, none of the valid ranges appear to have gaps.

msm0504 (changes requested):

modules/deepintentBidAdapter.js
```diff
@@ -1,10 +1,35 @@
...
+  'w': (value) => utils.isInteger(value),
+  'h': (value) => utils.isInteger(value),
+  'startdelay': (value) => utils.isInteger(value),
+  'placement': (value) => Array.isArray(value) && value.every(v => [1, 2, 3, 4, 5].indexOf(v) !== -1),
+  'linearity': (value) => [1, 2].indexOf(value) !== -1,
+  'skip': (value) => [0, 1].indexOf(value) !== -1,
+  'skipmin': (value) => utils.isInteger(value),
+  'skipafter': (value) => utils.isInteger(value),
+  'sequence': (value) => utils.isInteger(value),
+  'battr': (value) => Array.isArray(value) && value.every(v => Array.from({length: 17}, (_, i) => i + 1).indexOf(v) !== -1),
```
msm0504, line 22: `Array.from` isn't supported in Internet Explorer, which is causing the CircleCI tests to fails

msm0504 (changes requested):

ChrisHuie:
@sourabhg the conflicts are from `utils.`. We went through prebid this week and changed the utils imports to only be the needed functions so the utils root is no longer valid.

PR #7505, opened 2021-09-28
Gumgum Bid Adapter: use nearest matching h/w dimensions from bid request

lbenmore:
…uest

## Type of change

- [x] Bugfix
- [ ] Feature
- [ ] New bidder adapter   
- [ ] Build related changes
- [ ] CI related changes
- [ ] Other

## Description of change
Instead of using the response width and height as-is, we should use the nearest matching dimensions from the bid request


ChrisHuie:
@lbenmore can you please add a unit test to this 🙏

ChrisHuie:
Hey @lbenmore. We actually went through all adapters this week and are now importing only the needed functions from utils for each adapter. So you no longer need the utils.** root in this pr. That is where the linting error is coming from :)

lbenmore:
@ChrisHuie oh thanks for calling that out! updated now :D

ChrisHuie (approved):

PR #7511, opened 2021-09-29
Enable ecmaVersion: 2020

from20020516:
## Type of change

## Description of change

- update ecmaVersion for eslint to enable optional chaining operator and more.

related: https://github.com/prebid/Prebid.js/pull/7290#issuecomment-897827133


ChrisHuie:
@from20020516 we are currently working to upgrade to more modern JS but because of how certain features don't transpile for IE we are researching the best path currently

PR #7537, opened 2021-10-05
Yieldmo Synthetic Inventory Module

ym-abaranov:
## Type of change

- [ ] Bugfix
- [ ] Feature
- [ ] New bidder adapter   
- [ ] Build related changes
- [ ] CI related changes
- [x] Other

New module. A link to a PR on docs repo here -- https://github.com/prebid/prebid.github.io/pull/3322


desidiver:
@Fawke any feedback on this PR? Thank you.

ChrisHuie:
@desidiver it looks like this branch just needs to pull in recent commits to master to fix the errors. There was a linting error that briefly was merged before I could fix it and that must have been the master branch at the time of this pr. We rewrote all `utils` imports to only use the needed functions a couple of weeks ago and a few prs were still out there that had a .utils root.

ym-dlabuzov:
Hi @Fawke,
I've updated our tests, so now coverage should be > 85%

bretg:
There are lots of files in the modules directory that refer to googletag, so in general this is fine.

However, glancing briefly at how this module is doing it, doesn't seem consistent with others.

There's a utils function called `isGptPubadsDefined`. This could be useful.

For instance, the Rubicon Analytics Adapter has what seems like a good way to handle it. Here it is generalized a bit:
```
        if (!gptFlag && isGptPubadsDefined()) {
          // Can do special GPT stuff right now!
          gptFlag = true;
        } else if (!gptFlag) {
          gptFlag = true;
          window.googletag = window.googletag || {};
          window.googletag.cmd = window.googletag.cmd || [];
          window.googletag.cmd.push(function () {
            // Can do special GPT stuff once googletag is loaded.
          });
        }
```


Adding @harpere to weigh in on whether whether this module's implementation of googletag is ok. My gut feeling is that assigning window.top.googletag without first checking that it exists is a problem. 

Sounds like this might be another thing that needs to be checked by reviewers, but first we need consensus on the right approach.

ym-dlabuzov:
@Fawke, @bretg I've updated PR, so we are now using `isGptPubadsDefined()`.
However, I don't know why 3rd party test is failing. Maybe rerunning the build will fix it.

Issue #7416, opened 2021-09-14
Feature Request: explicitly set arrays and maps in gam targeting


patmmccann:
## Type of issue
Gam allows comma delimited strings to be treated as arrays but seems to perform better when arrays are passed explicitly. Someone should convert https://github.com/prebid/Prebid.js/blob/a887289546bdf830a3ff450a664627471eec9dfc/src/targeting.js#L386 to do it correctly. 

From https://developers.google.com/publisher-tag/reference we find: 
// Example with multiple values for a key inside in an array.
googletag.pubads().setTargeting('interests', ['sports','music','movies']);

Also, the foreach set targeting loop at https://github.com/prebid/Prebid.js/blob/a887289546bdf830a3ff450a664627471eec9dfc/src/targeting.js#L392 is probably much less efficient than using gpt updateTargetingFromMap. 

https://developers.google.com/publisher-tag/reference#googletag.Slot_updateTargetingFromMap

pm-harshad-mane:
Hello @patmmccann ,

1. In case of multiple values (array) for a key, we are setting each value of the respective key separately 
https://github.com/prebid/Prebid.js/blob/master/src/targeting.js#L390, this can be avoided by passing array of values for the respective key in `slot.setTargeting` API.

2. As per GPT documentation we can use a map of key and values to set targeting, example below
```
var slot = googletag.defineSlot('/1234567/sports', [160, 600], 'div-1');
slot.updateTargetingFromMap({
  'color': 'red',
  'interests': ['sports','music','movies']
});
```
At https://github.com/prebid/Prebid.js/blob/master/src/targeting.js#L383 we have an object `targetingConfig[targetId]` that can be used to pass to `slot.updateTargetingFromMap` API but we will have to make sure that if a value is string then it should be converted to an array, like at https://github.com/prebid/Prebid.js/blob/master/src/targeting.js#L385

So I think we can directly make the change mentioned in 2 to use `slot.updateTargetingFromMap` API with the string to array handling and the appropriate logMessage. 

Can you please confirm if my understanding is correct about the feature request?

patmmccann:
We saw a revenue gain from reducing consolidating set targeting calls as much as possible, so my goal was to duplicate that success into prebid. Re no 1., thank you for the correction. The foreach that loops through the elements is probably less efficient than setting the array explicitly, but there is probably very rarely an array of keys in prebid targeting.

[...]

PR #7552, opened 2021-10-07
Prebid Core: use gptSlot.updateTargetingFromMap than gptSlot.setTargeting in targeting.resetPresetTargeting

pm-harshad-mane:
## Type of change
- [ X ] Code style update (formatting, local variables)

## Description of change
`gptSlot.updateTargetingFromMap` can be used to set multiple targeting key-value pairs using an object.
It is better than calling `gptSlot.setTargeting` multiple times.
This PR makes change in targeting.resetPresetTargeting to use `gptSlot.updateTargetingFromMap`

Relates to #7416 and #7453


pm-harshad-mane:
Thank you @ChrisHuie and @osazos !

PR #7553, opened 2021-10-08
Proxistore Bid Adapter: export vendor id in gvlid

vincentproxistore:


ChrisHuie:
@vincentproxistore can you take out your packagee.lock file and maybe add a unit test 🙏

PR #7571, opened 2021-10-13
Yahoo SSP Bid Adapter: fix for adId

adam-browning:
## Type of change
- [x] Bugfix

## Description of change
Following a server side issue were our ad server does not always send adId in the response, we have a specific integration using Single Request Architecture that clashes when trying to fetch the creatives using the auction id.
The following fix provided fallbacks to match based on different attributes that might exist in the bid-response.

## Other information
@slimkrazy could you please review
Thanks in advance!
Adam

### DIFF
--- a/modules/yahoosspBidAdapter.js
+++ b/modules/yahoosspBidAdapter.js
@@ -6,7 +6,7 @@ import { Renderer } from '../src/Renderer.js';
 
 const INTEGRATION_METHOD = 'prebid.js';
 const BIDDER_CODE = 'yahoossp';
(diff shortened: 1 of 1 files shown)


adam-browning:
FYI circleci fail due to 3rd party analytics adapter

adam-browning:
@patmmccann @ChrisHuie is there any chance this small fix will make it into the upcoming deployment?
It would be great if can mitigate this in v5.18.0
Thank you so much!

ChrisHuie:
@adam-browning just merged it so it will be in this week's release later today or tomorrow :)

Issue #7457, opened 2021-09-23
2 different bidders being pushed to dfp


spormeon:
I seem to have found some weirdness after adding a new bidder to "test", the new bidder is only using "test" params which are in the adapters git file.  It seems to be pushing 2 different bidders in targeting.  P5.14.0  with all the "recommended" modules included.  Any idea whats gone haywire?

![Google_Ad_Manager_-_Delivery_tools](https://user-images.githubusercontent.com/8503167/134485621-e9a20bbc-d3c2-4cd5-9090-9d4df1d17db4.png)

spormeon:
I took out the "test" bidder and its still doing it, so I think there is some weird bug, maybe around the "recommended modules" installed

spormeon:
might be related to this, but i got no real clue:
P5.13.0,      PBjs Core (Targeting): bugfix for issue #7323 adding extra spaces (#7337)

gglas:
Is the issue that the bidder name is being passed twice, in hb_bidder and hb_bidder_name ?  We agree that that's a redundant kv pair and will triage separately -- but just want to make sure that's the actual problem you're seeing?  What are the "two different bidders" in your screenshot?  How many bidders are running in the auction?  I think what we'd really need to troubleshoot would be an indicator of what precise behavior in the screenshot you're seeing that looks strange.

spormeon:
hi There are 20+ bidders.  Its not the same bidder being passed twice, its 2 different bidders  and its random on which ones.  Screenshot above is aniview & rhythmone.  If @bretg is involved with this one, I can show on my test pages again ( like have shown on previous items)

bretg:
We don't understand how you've posed the problem @spormeon. Yes, a test page is needed.

spormeon:
here is another example, this one is sending 3 bidders.  Sometimes it sends 1, sometimes 2, sometimes 3,  its completely random.  The problem it creates for us, is we build a vmap off of the back of the winning bids and if targeting is sending 3 lots aof bids in effect, it crashed out the vmap string as its to long to send and build the vmap off of.  Its only started happening in P5.14.0   
`{
    "eventType": "setTargeting",
    "args": {
        "video10": {
            "hb_uuid": "cfc5e690-9851-4870-9065-89b5cfe544f8",
            "hb_cache_id": "cfc5e690-9851-4870-9065-89b5cfe544f8",
            "hb_cache_host": "prebid.adnxs.com",
            "hb_format": "video",
[...]

Issue #6563, opened 2021-04-09
Normalize PBJS and PBS auction IDs


bretg:
## Type of issue
enhancement

## Description
Trying to coordinate analytics between PBJS and PBS is made more difficult by their not having a common ID for the auction event.

- PBJS supports an 'auctionId' which is unique for each Auction refresh
- But the Prebid Server Bid Adapter sets $.id to s2sBidRequest.tid. 

I propose that we change the PBS bid adapter to use the PBJS auctionId as $.id for all PBS servers instead of having different $.id values for each of an array of prebid servers.

We should also set source.tid be the same across all Prebid Servers.

PR #7585, opened 2021-10-14
Prebid Server Bid Adapter: auctionId to $.id UUID to $.source.tid + mergeDeep extPrebid

robertrmartinez:
## Type of change
- [X] Other

## Description of change
Retry of #6836 

Implements #6563

We will now pass `auctionId` into `$.id`

Prebid's `auctionId` is a UUID - UNLESS a pub passes in their own auctionId.
This is not okay since PBS needs `$.source.tid` to be a UUID

Also it needs to be the SAME VALUE for every PBS request for a given auction (IE Multiple PBS Scenarios)

So we generate a UUID and assign it to each bidRequest to PBS as `tid`

So, for all PBS requests for a given auction, we should see the same exact `$.id`  and the same exact `$.source.tid` (which should be different themselves)

Like so:

image

## extPrebid Fix
Also in this PR is a fix for #7457

Which just deep merges extPrebid instead of shallow merge.


Fawke (approved):

patmmccann:
Did we want to backport this one?

PR #7601, opened 2021-10-19
VidoomyBidAdapter: refactor cookie sync

sasanfarokh:
## Type of change

- [x] Bugfix
- [ ] Feature
- [ ] New bidder adapter   
- [ ] Build related changes
- [ ] CI related changes
- [ ] Other

## Description of change
Just call cookie sync urls sequentially with a 500ms delay between them.

### DIFF
--- a/modules/vidoomyBidAdapter.js
+++ b/modules/vidoomyBidAdapter.js
@@ -206,19 +206,31 @@ function cookieSync(bidderRequest) {
       gpdr: bidderRequest.gdprConsent ? bidderRequest.gdprConsent.gdprApplies : '0',
       gpdr_consent: bidderRequest.gdprConsent ? bidderRequest.gdprConsent.consentString : '',
     });
-    JSON.parse(this.responseText).filter(Boolean).forEach(url => {
-      firePixel(macro.replace(url))
-    })
+    const urls = JSON.parse(this.responseText).filter(Boolean).map(url => macro.replace(url))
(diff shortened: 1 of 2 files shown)


sasanfarokh:
> While testing your bidder I found some small errors. You can see them in separate comments. From Prebid perspective, you don't have to fix those issues. But, I strongly suggest that you fix them anyway, it will make your adapter run a bit faster.
> 
> Also, why are you passing `ua` (userAgent string) in query params, you already have that information in HTTP header?

Hey, About the user-agent you're right, It's already exists in HTTP headers, but currently our backend apps may read it from querystring, I want to leave it as before.

FilipStamenkovic (changes requested):
Your tests are failing, please take a look at CircleCI output.

modules/vidoomyBidAdapter.js
```diff
@@ -185,91 +194,32 @@ const interpretResponse = (serverResponse, bidRequest) => {
   }
 };
 
+function getUserSyncs (syncOptions, responses, gdprConsent, uspConsent) {
```
FilipStamenkovic, line 197: You are not using `uspConsent` anywhere.  

You must include `uspConsent` string in your cookie sync urls.

FilipStamenkovic (changes requested):

FilipStamenkovic (approved):

PR #7604, opened 2021-10-20
showheroes-bs Bid Adapter: fix for outstream render

vadim-mazzherin:
## Type of change
- [x] Bugfix
- [ ] Feature
- [ ] New bidder adapter   
- [ ] Build related changes
- [ ] CI related changes
- [ ] Other

## Description of change
Hotfix for outstream render.

At the moment we can't use our rendering script without directly using the onLoad event. 
If there is an option for how to make an option without necessarily specifying the `url` in the `Renderer.install`, please let us know.


ChrisHuie:
@vadim-mazzherin it looks like this pr never kicked off our circleci tests when submitted. Can you please resubmit it and possibly give me edit access to the pr in case I need to kick off tests manually? 🙏

Issue #7373, opened 2021-09-01
Prebid 6.0


gglas:
This issue will serve as a placeholder for Prebid 6.0.  The current plan for 6.0 is to remove the test requirements for IE11, allowing developers to use code not compatible with IE11 that cannot be easily transpiled.

There are few things we need to decide upon as a community : 

- How do we provide support for unsophisticated publishers who still want to serve ads on IE (and will need to use a 5.x version)
- what should our timelines be given the microsoft changes?  who will leverage the new features immediately?

jdcauley:
Mediavine anticipates using 6.0 as soon as it is available and passes internal validations.

patmmccann:
We looked into the spread operator and babel, it appears you can transpile the spread operator without targeting es5 via a babel plugin. 

We hope everything else prebid currently needs can be polyfilled.

This would mean the prebid examples would just have a modern js target (with the exception of the spread operator plugin) and also source `https://polyfill.io/v3/polyfill.js?features=es2015%2Ces2016%2Ces2017%2Ces5%2Ces6%2Ces7%2Ces2019%2Ces2018%2Cdefault%2CIntl` or whatever else we need from `https://polyfill.io/v3/url-builder/`

We're very much looking forward to the tests that see if this outcome would work and even if it does not, we are somewhat confident the yield loss from the remaining browsers that break would be less revenue than yield gains on modern browsers

khatibda:
we would take advantage of this as soon as we could.

PR #7606, opened 2021-10-20
Prebid 6.0

patmmccann:
solution to #7373


patmmccann:
fiddle is ready for review

PR #7633, opened 2021-10-26
Prebid Core: add new setting to allow zero cpm bids

### LINKED ISSUE #6887: Add zero value bids to ad server targeting
## Type of issue
Question

## Description
Hi!

Bid responses with CPM = 0 don't get added to ad server targeting. Is there a flag somewhere that we could add to enable this? I've searched in the documentation but could not find anything regarding this.

The case we are having is that our network backfills campaigns are running in Xandr and the ad server on this particular site is Google.  Maintaining backfill campaigns in two or more ad servers is a mess that we would like to avoid.

Regards,
Patrik

jsnellbaker:
## Type of change

- [x] Feature

## Description of change
Fixes #6887 

Adds a new setting in the `bidderSettings` to allow 0 cpm bids to be accepted/processed by Prebid.js.  The setting can be activated by the following ways (either globally or per bidder):
```
pbjs.bidderSettings = {
  standard: {
    allowZeroCpmBids: true
  }
});
```
```
pbjs.bidderSettings = {
  appnexus: {
    allowZeroCpmBids: true
  }
});
```

If people feel this setting should have a different name and/or be located in a different place, please feel free to share.

Also added the setting to the appnexus bid adapter, as it also had a filter to ignore 0 cpm bids.

## Other information
https://github.com/prebid/prebid.github.io/pull/3414

### DIFF
--- a/src/utils.js
+++ b/src/utils.js
@@ -1,5 +1,6 @@
 /* eslint-disable no-console */
 import { config } from './config.js';
+import { getGlobal } from './prebidGlobal.js';
 import clone from 'just-clone';
 import find from 'core-js-pure/features/array/find.js';
 import includes from 'core-js-pure/features/array/includes.js';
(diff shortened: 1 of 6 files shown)


jsnellbaker:
Updated the logic here to use the bidderSettings feature instead of the original setConfig; this will allow the feature to work for specific bidders if desired.

CC: @bretg

bretg:
Thanks @jsnellbaker

src/utils.js
```diff
@@ -1294,3 +1295,12 @@ export function cyrb53Hash(str, seed = 0) {
   h2 = imul(h2 ^ (h2 >>> 16), 2246822507) ^ imul(h1 ^ (h1 >>> 13), 3266489909);
   return (4294967296 * (2097151 & h2) + (h1 >>> 0)).toString();
 }
+
+export function isAllowZeroCpmBidsEnabled(bidderCode) {
+  const bidderSettings = getGlobal().bidderSettings;
+  if ((bidderSettings[bidderCode] && bidderSettings[bidderCode].allowZeroCpmBids === true) ||
```
msm0504, line 1301: This `if` can be replaced with `return`

jsnellbaker:
@msm0504 I have made some updates based on your feedback.  Please take a look when you have the chance.

Thanks.

msm0504 (approved):

msm0504:
@jsnellbaker I've approved the changes, but now this needs a 2nd review. Also, not sure why the tests aren't passing. You might need to merge the latest from `master` into your branch.

jsnellbaker:
I've resynced with master, let's see if that helps the tests.

@Fawke Can you take a look when you get the chance.  Thanks.

Fawke (approved):

PR #7643, opened 2021-10-29
Generic viewability module: add new module

aleksatr:
## Type of change
- [x] Feature

## Description of change
This is a generic viewability module, it will trigger a viewability pixel (js, img) or a callback, when a given HTML element becomes viewable, according to a custom viewability criteria.

Notes:
- Does not depend on publishers using GAM ad server (does not use GPT's ImpressionViewableEvent).
- Allows custom viewability criteria (timeInView and inViewThreshold).
- Aims to provide (and in some ways extend) functionality of https://github.com/InteractiveAdvertisingBureau/openvv-html , but keeping the size to the bare minimum (openvv-html is ~26kb in size).
- Allows publishers to track viewability through callbacks or img/js pixels.
- Allows buy side to track viewability by posting a message from within the creative iframe (including the postMessage call in the ad markup), thus avoiding custom viewability scripts.

- A link to a PR on the docs repo: https://github.com/prebid/prebid.github.io/pull/3381

## Other information
This module does not need any configuration, as long as you include it in your PBJS bundle.


MartianTribe:
https://github.com/prebid/prebid.github.io/pull/3381

pm-harshad-mane:
@aleksatr please check https://github.com/prebid/Prebid.js/blob/master/modules/bidViewabilityIO.md and https://github.com/prebid/Prebid.js/blob/master/modules/bidViewabilityIO.js

aleksatr:
> @aleksatr please check https://github.com/prebid/Prebid.js/blob/master/modules/bidViewabilityIO.md and https://github.com/prebid/Prebid.js/blob/master/modules/bidViewabilityIO.js

Hey @pm-harshad-mane , thanks for looking into this. I was aware of `bidViewability` module, but I somehow missed `bidViewabilityIO`. I see that there are similarities between the one I'm submitting and the bidViewabilityIO, in that both are using IntersectionObserver to track element viewability and that they are both usable without ad server. Unfortunately bidViewabilityIO is not that generic, to support some custom viewability use cases and criteria. As it's name suggests it's probably intended to track *bid* viewability, according to standard viewability criteria. 

Some limitations I noticed on a first glance are:

- supports only banner media type (no video and native).
- has predefined viewability criteria, does not support a custom one.
- automatically tracks bid viewability assuming that div_id == adUnitCode (which is not always the case).
- no way for buy side to track viewability, our script listens for messages from creatives as well, and fires viewability pixels.
- not able to cover some more complex viewability use cases, if a creative is some slide-show, for example, you can track each slide's viewability. Also, you can communicate from a video player by posting a message to the viewability module to start/stop measurement (in theory).

At the same time, these three viewability modules have some distinct features that none of the other two have. So no module is complete superset of the other two. How can we proceed with this? Do you think Prebid needs to have a single viewability module that covers as many use cases as possible, in which case we should probably try to somehow unify all these modules? Or do you think that since they are in some way covering some distinct use cases (all related to viewability but still), they can coexist separately?

jimdigriz:
@aleksatr we also slipped a viewability sensor (ATF) in our RTD module which you may wish to crib from and later replace https://github.com/prebid/Prebid.js/blob/51f9244dd458b49e757ec6a637aa0029a2944bf1/modules/adlooxRtdProvider.js#L37-L144

modules/viewability.js
```diff
@@ -0,0 +1,163 @@
...
+
+  if (!data || data.message !== 'Prebid Viewability') {
+    return;
+  }
+
+  switch (data.action) {
+    case 'startMeasurement':
+      let element = data.elementId && document.getElementById(data.elementId);
+      if (!element) {
+        element = find(document.getElementsByTagName('IFRAME'), iframe => (iframe.contentWindow || iframe.contentDocument.defaultView) == evt.source);
```
Fawke, line 152: Just a little bit confused as to if it should be `evt.source` or `evt.target`. I checked the[ MDN docs](https://developer.mozilla.org/en-US/docs/Web/API/Event) but did not find a `source` property on the `event` object.

modules/viewability.js
```diff
@@ -0,0 +1,163 @@
...
+    return false;
+  }
+
+  if (!vid || observers[vid]) {
+    logWarn('provide an unregistered vid', vid);
+    return false;
+  }
+
+  return true;
+}
```
Fawke, line 43: Maybe adding the Module name in front of the log messages will help the debugging process easier?

modules/viewability.js
```diff
@@ -0,0 +1,163 @@
...
+            tracker.value(element);
+            break;
+        }
+      }, criteria.timeInView);
+    } else if (observers[vid].timeoutId) {
+      window.clearTimeout(observers[vid].timeoutId);
+    }
+  };
+
+  observer = new IntersectionObserver(stateChange, options);
```
Fawke, line 95: I was going through the docs for IntersectionObserver API on MDN, on the [Browser compatibility section](https://developer.mozilla.org/en-US/docs/Web/API/Intersection_Observer_API#browser_compatibility), I found that Safari and few of the other Mobile browsers don't support `options.root`. I'm not sure if it'll throw and error and crash on those browsers, do we need a fallback or some error handling?

modules/viewability.js
```diff
@@ -0,0 +1,163 @@
...
+    if (viewable) {
+      observers[vid].timeoutId = window.setTimeout(() => {
+        // stop observing
+        observer.unobserve(element);
+        observers[vid].done = true;
+
+        switch (tracker.method) {
+          case 'img':
+            triggerPixel(tracker.value, () => {
+              logInfo('viewability pixel fired', tracker.value);
```
Fawke, line 79: adding module name in front would help

modules/viewability.js
```diff
@@ -0,0 +1,163 @@
...
+  observer.observe(element);
+}
+
+/**
+ * Stop measuring viewability of an element
+ * @param {string} vid unique viewability identifier
+ */
+export function stopMeasurement(vid) {
+  if (!vid || !observers[vid]) {
+    logWarn('provide a registered vid', vid);
```
Fawke, line 112: adding module name in front would help

Fawke (changes requested):
Hi @aleksatr ,

Thanks for creating this module. Your changes are mostly fine, there are a few in line comments where I have some doubts, but rests looks good for me.

I don't know if you have a html page or something in order to test this? If you have, that'll help a lot to test this out.

modules/viewability.js
aleksatr: I assume your referring to this `options.root parameter can be a Document`? If that's the case, we're setting options.root to null in this PR, not to any document?

modules/viewability.js
aleksatr: You're right, thanks, I'll add module name

modules/viewability.js
aleksatr: updated

modules/viewability.js
aleksatr: updated

aleksatr:
> Hi @aleksatr ,
> 
> Thanks for creating this module. Your changes are mostly fine, there are a few in line comments where I have some doubts, but rests looks good for me.
> 
> I don't know if you have a html page or something in order to test this? If you have, that'll help a lot to test this out.

Hi @Fawke , thanks for reviewing, I don't have a dedicated test page for this, I used the standard pbjs test page `/integrationExamples/gpt/hello_world.html`, and then called the methods on viewability module through dev console, for example:

```
pbjs.viewability.startMeasurement(
	'1',
	document.getElementById('div-gpt-ad-1460505748561-0'),
	{ method: 'callback', value: () => { console.log('viewability fired'); } },
	{ inViewThreshold: 1.0, timeInView: 1000 }
);
```

And then I was resizing window to hide the ad (or part of the ad depending on the inViewThreshold). I also sometimes added some block elements in front of the div-gpt-ad-1460505748561-0 to push it bellow the fold (instead of resizing) and then observed the behavior while scrolling. I was also calling `stopMeasurement` through dev console.

modules/viewability.js
Fawke: I see, I think I got myself confused between Event and MessageEvent Objects. This is fine, thanks.

Fawke (approved):
LGTM

aleksatr:
Hi @ncolletti will you have time to take a look at this PR?

ncolletti (commented):
sorry for the delay! this is looking good, only a few minor suggestions

modules/viewability.js
```diff
@@ -0,0 +1,165 @@
...
+
+  observer = new IntersectionObserver(stateChange, options);
+  observers[vid] = {
+    observer: observer,
+    element: element,
+    timeoutId: null,
+    done: false,
+  };
+
+  observer.observe(element);
```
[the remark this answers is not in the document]
aleksatr: I would leave this to the caller, to make sure that the element to observe is actually part of the DOM, wdyt? 
I added the log message when startMeasuement is called, with args.

aleksatr:
> sorry for the delay! this is looking good, only a few minor suggestions

Hi @ncolletti , thanks for the detailed review, I pushed changes that should address your remarks.

modules/viewability.js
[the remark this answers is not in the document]
ncolletti: Thanks, yeah I think a log is enough

ncolletti (approved):
lgtm!

mikemangas:
@aleksatr what is the most beneficial use case of this module?

aleksatr:
> @aleksatr what is the most beneficial use case of this module?

Hi @mikemangas , the main goal was to overcome some shortcomings that other viewability modules in pbjs had (more on that in this PR description). I was trying to keep the module pretty generic and allow some higher-level viewability scenarios to be implemented on top of it. Imagine a slide-show type of creative, or in general some dynamic behavior inside the creative, elements coming in and going out of the view, where you would like to know viewability for each element. This type of viewability tracking can be implemented using the functionality that this module exposes. And in general, the main focus is still on providing the viewability metrics.

mikemangas:
@aleksatr 
why is there a stopMeasurement function?
normally, you start the measurement... and as soon as the criteria (e.g. { inViewThreshold: 0.5, timeInView: 1000 }) are beeing reached, then it "fires" the pixel. Therefore i do not really see the usecase of the stopMeasurement function.
Thank you for any tip.

aleksatr:
> @aleksatr why is there a stopMeasurement function? normally, you start the measurement... and as soon as the criteria (e.g. { inViewThreshold: 0.5, timeInView: 1000 }) are beeing reached, then it "fires" the pixel. Therefore i do not really see the usecase of the stopMeasurement function. Thank you for your feedback on my previous post :)

@mikemangas but what if that criteria is never reached (in some limited time frame), and you want to render a new ad in that same div element, in that case you want to stop the old observer and start a new one (with it's own tracker and criteria). Other usecase might be when you have a "slide-show" type of ad, where your ad has transitioned to the second slide, without the first slide becoming viewable (according to criteria).

PR #7648, opened 2021-11-01
Missena Bid Adapter: add back adapter to latest prebid version

pdamoc:
## Type of change

- [ ] Bugfix
- [ ] Feature
- [ ] New bidder adapter   
- [ ] Build related changes
- [ ] CI related changes
- [x] Other

## Description of change
This PR restores the missena adapter to the state before 412277f72b97cab72644cb4a009f58067a165899 

## Other information

As far as I understand things, the missena adapter was include by error in the #6650. 

The maintainer did not react due to competing priorities and the adapter was removed. 

I have been tasked with getting the adapter re-enabled. If there are issues with the adapter that makes it incompatible with the current Prebid.js, please let me know what are the issues so I can address them. 

Regarding the `advertiserDomains` mentioned in #6650,  for valid bids responses, the adapter returns a `meta` object with an empty `advertiserDomains`.

### DIFF
--- a/modules/missenaBidAdapter.js
+++ b/modules/missenaBidAdapter.js
@@ -1,0 +1,94 @@
+import * as utils from '../src/utils.js';
+import { BANNER } from '../src/mediaTypes.js';
+import { registerBidder } from '../src/adapters/bidderFactory.js';
+
+const BIDDER_CODE = 'missena';
+const ENDPOINT_URL = 'https://bid.missena.io/';
+
+export const spec = {
+  aliases: [BIDDER_CODE],
+  code: BIDDER_CODE,
+  gvlid: 687,
+  supportedMediaTypes: [BANNER],
+
+  /**
+   * Determines whether or not the given bid request is valid.
+   *
+   * @param {BidRequest} bid The bid params to validate.
+   * @return boolean True if this is a valid bid, and false otherwise.
+   */
+  isBidRequestValid: function (bid) {
+    return typeof bid == 'object' && !!bid.params.apiKey;
+  },
+
+  /**
+   * Make a server request from the list of BidRequests.
+   *
+   * @param {validBidRequests[]} - an array of bids
+   * @return ServerRequest Info describing the request to the server.
+   */
+  buildRequests: function (validBidRequests, bidderRequest) {
+    return validBidRequests.map((bidRequest) => {
+      const payload = {
+        request_id: bidRequest.bidId,
+        timeout: bidderRequest.timeout,
+      };
+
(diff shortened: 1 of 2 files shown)


pdamoc:
Hi @ChrisHuie,
Do you have any feedback on this PR? 
Is there anything we need to change in order for the PR to be merged?

modules/missenaBidAdapter.js
```diff
@@ -0,0 +1,94 @@
+import * as utils from '../src/utils.js';
```
ChrisHuie, line 1: Can you please import only the needed utility functions here? We have changed how all adapters import from utils

ChrisHuie:
@pdamoc Can you also please add handling for advertiserdomain also? We have made that a requirement going forward per https://github.com/prebid/Prebid.js/issues/6650

pdamoc:
> @pdamoc Can you also please add handling for advertiserdomain also?  

This is what I'm trying to clarify/understand. What constitutes handling for advertiserDomains.

There is a `meta` field with an `advertiserDomains` field in the `serverResponse` (when there is a non-empty response). This is passed along in the `bidResponse` that is returned from `interpretResponse`. To my knowledge, currently this field is null or contains an empty array. 

I'm trying to understand what should be different. 

I will refactor the `utils` use ASAP.

pdamoc:
I have rebased the commits on top of current `upstream/master` and has also solved the `utils` issue.

ChrisHuie:
@pdamoc if it is returned from the server response then if you just add testing for that it works then you should be good.

ChrisHuie:
docs -> https://github.com/prebid/prebid.github.io/pull/3408

pdamoc:
@ChrisHuie after getting an error in some other adapter I tried rebasing the branch on current `upstream/master` but now I got a different error. 

This looks to me like the kind of transient errors that I frequently get on my computer even if I change no code. What is the best way to handle these errors in the context of a PR like this one?

ChrisHuie:
@pdamoc yeah we have a few flaky tests on specific browsers that we are trying to clean up. So this pr will add Missena to master but since we are currently on 6.1 this doesn't add it to 5.20.x. Here is that branch (https://github.com/prebid/Prebid.js/tree/5.20.x-legacy) if you want to make sure this adapter is in prebid 5 also. We are planning support for that version through about Q1 next year.

pdamoc:
@ChrisHuie considering the current use-case for the adapter, being available only in `6.x` is perfectly fine. If things change on our end, I will also create a PR on the legacy branch but for the time being, availability in `6.x` is sufficient.

ChrisHuie (approved):

PR #7584, opened 2021-10-14
Sspbc Bid Adapter: update version and bugfixes

wojciech-bialy-wpm:
## Type of change

- [x] Bugfix

## Description of change

Bugfix/update for sspBCBidAdapter

- set correct content type for notifications
- add call to action to native ads
- change tagid generation, so it remains constant for the same adUnit across auctions
- update test for tagid
- send page view id in bid request

PR #7654, opened 2021-11-02
Revert "Sspbc Bid Adapter: update version and bugfixes"

patmmccann:
Reverts prebid/Prebid.js#7584 due to failing CI tests

update - the revert fails tests too; maybe this is the wrong path

### DIFF
--- a/modules/sspBCBidAdapter.js
+++ b/modules/sspBCBidAdapter.js
@@ -9,13 +9,11 @@ const BIDDER_URL = 'https://ssp.wp.pl/bidder/';
 const SYNC_URL = 'https://ssp.wp.pl/bidder/usersync';
 const NOTIFY_URL = 'https://ssp.wp.pl/bidder/notify';
 const TMAX = 450;
-const BIDDER_VERSION = '5.3';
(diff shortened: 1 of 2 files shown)


Rothalack:
What I've started doing in cases like this is trust the original circleci test of the PR itself. If that passes, I feel pretty confident. Then when I do the master merge, I give the circleci build a couple tries and verify that the any errors occurring are not related to the changes in the PR. I try not to sit spam out circleci builds over and over and clog up the system. Maybe I should just keep retrying until it passes.

That noted, I do not see any errors related to the SSPBC update or files.

Is it possible circleci doesn't have enough resources to run the tests anymore? This is something I've noticed recently in local testing, I have to add to my gulp command with enough RAM: 
NODE_OPTIONS=--max-old-space-size=7168 gulp review-start
Just a thought.

patmmccann:
That is really interesting @Rothalack 

The revert didn't pass the test either, so I am not sure what is the source of these failing adapter manager size tests. Usually any failing tests are timeouts from an analytics call not properly mocked or something.

PR #7655, opened 2021-11-03
DeepIntent Bid Adapter: secure flag set using proper protocol value

parthshah51999:
## Type of change

- [x] Bugfix

## Description of change

Set secure flag using correct comparison with window.location.protocol in buildImpressions function


patmmccann:
LGTM

PR #7658, opened 2021-11-03
Appnexus bid adapter: add support for brandId

jsnellbaker:
## Type of change

- [x] Feature

## Description of change
Adds support in the appnexus bid adapter to assign the `brand_id` value to the prebid `bid.meta.brandId` field.

### DIFF
--- a/modules/appnexusBidAdapter.js
+++ b/modules/appnexusBidAdapter.js
@@ -598,6 +598,10 @@ function newBid(serverBid, rtbBid, bidderRequest) {
     bid.meta = Object.assign({}, bid.meta, { advertiserId: rtbBid.advertiser_id });
   }
 
+  if (rtbBid.brand_id) {
+    bid.meta = Object.assign({}, bid.meta, { brandId: rtbBid.brand_id });
+  }
+
   if (rtbBid.rtb.video) {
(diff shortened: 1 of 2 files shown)


modules/appnexusBidAdapter.js
```diff
@@ -598,6 +598,10 @@ function newBid(serverBid, rtbBid, bidderRequest) {
     bid.meta = Object.assign({}, bid.meta, { advertiserId: rtbBid.advertiser_id });
   }
 
+  if (rtbBid.brand_id) {
+    bid.meta = Object.assign({}, bid.meta, { brandId: rtbBid.brand_id });
```
patmmccann, line 602: https://docs.prebid.org/dev-docs/bidder-adaptor.html specifies 'brandId' and 'advertiserId'

jsnellbaker: I'm not sure I follow the question.  We hadd already specified advertiserId and set in the bid.meta object.  This PR adds support for brandId, which was missing.

patmmccann: sorry I thought there was a typo that I was pointing out but my mistake, no typo.

patmmccann (approved):

PR #7661, opened 2021-11-03
IX Bid Adapter: add support for handling vastXML responses in PBJS

shahinrahbariasl:
## Type of change

- [ ] Bugfix
- [X] Feature
- [ ] New bidder adapter   
- [ ] Build related changes
- [ ] CI related changes
- [ ] Other

## Description of change

- Add support for handling vastXML responses in IX adapter.
- If both vastXml and vastUrl exist in bid response, vastXml takes priority over vastUrl since only one of them is needed.
- Modify out stream renderer call to use either of vastXml or vastUrl.
- This change is backward compatible with current Exchange behaviour.


jsnellbaker (changes requested):
@shahinrahbariasl 

Can you please try to resync with master, that should help to clean up the unit test errors in the circleci report.

Thanks.

jsnellbaker (approved):
I saw the resync with master, but the browserstack tests didn't see to respond better in circleci.

I ran the browserstack tests manually and they all completed, so I think this is fine to go.

Issue #7367, opened 2021-08-31
Where does the GPID come from ?


patmmccann:
## Type of issue

The TradeDesk is encouraging publishers to provide "Global Placement ID" (GPID) data that's sometimes more granular than the GAM AdSlot. For instance, if a publisher's GAM AdSlots are all the same on a given page, TTD wants them to append that value with the div-id to make the values unique.

This desire for uniqueness is the reason that the [Prebid AdSlot](https://docs.prebid.org/features/pbAdSlot.html) was created. The [GPT Pre-Auction module](https://docs.prebid.org/dev-docs/modules/gpt-pre-auction.html) supports publishers who want to pass the GAM AdSlot, and allows them to customize the PB AdSlot by defining a function.

So a given page might have these values:

- ortb2Imp.ext.data.adserver.name:"gam"
- ortb2Imp.ext.data.adserver.adslot:"/1111/homepage"
- ortb2Imp.ext.data.pbadslot:"/1111/homepage/div-1"

The issue at hand is which of these values should a bid adapter read? Adapters are inconsistent:

- Index reads 'ortb2Imp.ext.data.adserver.adslot'
- Appnexus reads 'ortb2Imp.ext.data.pbadslot'
- Others (grid/magnite/sovrn) look at both, preferring GAM AdSlot over PB AdSlot.

We need to discuss A) standardizing, B) giving the publisher some control, C) whether this issue may affect modules like Price Floors, Viewability, and other RTD modules.

There is also a docs repo issue https://github.com/prebid/prebid.github.io/pull/2928 tracking a proposal on GPID docs

bwoolcott:
Good summary, thanks @patmmccann The reason that these disparate ways exist in bid modules today is that the rollout of GPID came in phases (two so far) and in phase one all of the publishers in question had descriptive GAM AdSlots (unique per slots on page,) so there was no request to have div-id be part of the GPID. 

So this inconsistency came from the fact that between 4/1/21 (the start of phase 1) and 8/1/21 (the start of phase 2,) that implementing in the way Index did (Index reads `ortb2Imp.ext.data.adserver.adslot`) was perfectly fine as only GAM ad slot was being requested as GPID (no mention of div-id.) So Index and other SSPs did the engineering work to parse from `ortb2Imp.ext.data.adserver.adslot` as the GPT pre auction module made this a very reliable place for GAM AdSlot to exist, and was a good place to get GPID from. So the requirements with div-id being added now are different and create some extra work.

[...]

PR #7671, opened 2021-11-04
GPT Pre Auction: setting GPID

robertrmartinez:
## Type of change
- [X] Feature

## Description of change
Addresses #7367 

GPID LOGIC:

IF GPID was set by something else -> use it

ELSE IF pbadslot was set to something other than adUnitCode -> use that

ELSE set it to nothing


bretg:
Thanks @robertrmartinez - probably ought to have a unit test asserting the setting of gpid

robertrmartinez:
@bretg Yeah have not gotten to unit tests as my local is not working for some reason (new laptop)

Will add some tests and fix others!

ChrisHuie (approved):

PR #7679, opened 2021-11-08
Eplanning & Adapter Manager: fix testing of innerwidth to fix current build

robertrmartinez:
## Type of change
- [X] Bugfix
- 
## Description of change
Something changed in the browserstack tests.

After looking into it a bit, it seems the `window` innerWidth and other similar features may be returning different things than previous.

The `E Planning` Tests seemed to be written with the assumption that `innerWidth` would resolve to something lower than 1024, which they treated as "Mobile"

For Adapter Manager Tests, it looks like it might center around the [window.matchMedia](https://developer.mozilla.org/en-US/docs/Web/API/Window/matchMedia) function, which may be returning the opposite as to what the tests expects... perhaps the mocking is not working?

### DIFF
--- a/src/sizeMapping.js
+++ b/src/sizeMapping.js
@@ -1,4 +1,3 @@
-/* eslint-disable no-console */
 import { config } from './config.js';
 import {logWarn, isPlainObject, deepAccess, deepClone, getWindowTop} from './utils.js';
 import includes from 'core-js-pure/features/array/includes.js';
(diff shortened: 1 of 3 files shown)


robertrmartinez:
Looking at the test run 
https://app.circleci.com/pipelines/github/prebid/Prebid.js/8962/workflows/95c28e4e-01d8-4111-bfa7-885254147396/jobs/19230

Which is after commit d210075

Increasing the min-width mediaQuery to a much larger pixel number passes the Adapter Manager `should filter sizes using size config` test.

As for E-Planning

These tests fail because the adapter expected the minWidth to be something lower than 1024, I am guessing this was true sometime before Browserstack changed something (I cannot find something in our recent changes which would change this by our hand)

The Edge 17 Tests actually run fine, and as you can see here, it is because the Edge tests run in a smaller width browser:

image


When compared to other browsers:

image


Where you can see the test failed because it was expecting a smaller innerWidth

image


image


So, we can try and look into changing this to match the tests (I did a quick look into setting screen size etc in browserstack and did not find a quick and easy solution)

OR

We can update the tests to correctly mock these window attributes and not rely on whatever they actually are!

ChrisHuie (approved):

PR #7681, opened 2021-11-08
UserID Module: allow userid to ppid sync

robertrmartinez:
## Type of change
- [X] Feature

## Description of change
Addresses https://github.com/prebid/Prebid.js/issues/6430

User will need to signal to the userId Module which USER ID Submodule it wishes to sync as its `PPID`

They can do this like so:

```js
pbjs.setConfig({
        userSync: {
          ppid: 'pubcid.org',
          userIds: [
            { name: 'amxId', value: {'amxId': 'amx-id-value-amx-id-value-amx-id-value'} },
            { name: 'pubCommonId', value: {'pubcid': 'pubCommon-id-value-pubCommon-id-value'} },
            { name: 'identityLink', value: {'idl_env': 'identityLink-id-value-identityLink-id-value'} },
          ]
        }
      });
```

NOTE: This `ppid: 'pubcid.org'` value must directly map to the associated EID's SOURCE value for the given submodule as seen in the [eids.js](https://github.com/prebid/Prebid.js/blob/master/modules/userId/eids.js#L22) file.

Once this is done, extra logic is run at the time of REQUEST BIDS in order to resolve the `ppid` value and set it for googletag to consume.

GOOGLE has some strict parameters for how a ppid should be (more than 32 chars less than 150, and only alphanumeric)

DOCS: https://github.com/prebid/prebid.github.io/pull/2788

### DIFF
--- a/modules/userId/index.js
+++ b/modules/userId/index.js
@@ -185,7 +185,7 @@ export let syncDelay;
 export let auctionDelay;
 
 /** @type {(string|undefined)} */
-export let ppidSource;
+let ppidSource;
 
 /** @param {Submodule[]} submodules */
 export function setSubmoduleRegistry(submodules) {
--- a/modules/userId/index.js
+++ b/modules/userId/index.js
@@ -575,6 +575,8 @@ export function requestBidsHook(fn, reqBidsConfigObj) {
             window.googletag.pubads().setPublisherProvidedId(ppidValue);
           });
         }
+      } else {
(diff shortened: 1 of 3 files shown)


smenzer (approved):
LGTM

robertrmartinez:
@patmmccann I added the logWarn, let me know if it is okay.

patmmccann:
LGTM

Thanks!

patmmccann (approved):

PR #7700, opened 2021-11-11
iprom Bid Adapter: restore adapter in prebid master

iprom-adserver:
## Type of change

- [ ] Bugfix
- [x] Feature
- [x] New bidder adapter   
- [ ] Build related changes
- [ ] CI related changes
- [ ] Other

## Description of change

ipromBidAdater restored for versions 5 and 6+

### DIFF
--- a/modules/ipromBidAdapter.js
+++ b/modules/ipromBidAdapter.js
@@ -1,0 +1,79 @@
+import * as utils from '../src/utils.js';
+import { registerBidder } from '../src/adapters/bidderFactory.js';
+
+const BIDDER_CODE = 'iprom';
+const ENDPOINT_URL = 'https://core.iprom.net/programmatic';
+const VERSION = 'v1.0.1';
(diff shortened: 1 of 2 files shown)


modules/ipromBidAdapter.js
```diff
@@ -0,0 +1,79 @@
+import * as utils from '../src/utils.js';
```
ChrisHuie, line 1: can you please update this import to just pull in the needed functions. We recently went through all adapters and removed the '.utils' roots

ChrisHuie:
@iprom-adserver now that we are on version 6 this would also need to be added to the 5.20.x legacy branch (https://github.com/prebid/Prebid.js/tree/5.20.x-legacy) for this to be in 5 also.

modules/ipromBidAdapter.js
iprom-adserver: Done.

ChrisHuie (approved):

Issue #6388, opened 2021-03-05
client-side adapter code should not be required for server-side bidders


spormeon:
I just updated to 4.29.0   from the download page UI and I don't think "prebid server" is being included in the download?  client side bidders work ok, but nothing/ zip for "auction"

![DevTools_-_ads_vsvl_co_uk_pub000020_scripts_AMPTEST4_29_0_html_and_Download_Prebid_js_for_Header_Bidding](https://user-images.githubusercontent.com/8503167/110118883-ebe16500-7db2-11eb-8452-3026fb8bd434.png)

bretg:
@spormeon - download of the pbsBidAdapter works for me.

 search for a string in your package known to be within an adapter like "Incorrect or unavailable prebid server default vendor option"

What's the last version you were on? There've been a few changes in the pbsBidAdapter but nothing that should be backwards compatibility breakin. There is one bug in aliases that might affect behavior - see https://github.com/prebid/Prebid.js/issues/6336

spormeon:
i was on 4.20 before ( my "live" instances are on this), I tried 4.28 as well, and seeing same result,  its throwing a big chunk of warnings,  its exactly the same setup as on "live" instances:

![DevTools_-_ads_vsvl_co_uk_pub000020_scripts_AMPTEST4_29_0_html](https://user-images.githubusercontent.com/8503167/110151824-d717c800-7dd8-11eb-8aee-bbfaa04f2795.png)

spormeon:
hang on, have you now got to also include the actual adapters of each of the bidders?

bretg:
I've seen this message recently. It means your PBJS package doesn't have these bidders in them. Please verify your download/build of the package worked properly.

spormeon:
i'm only adding the adapters of the bidders that are "client" side, as before when using prebid server you didn't have to included the adapters of the prebid server bidders

bretg:
> before when using prebid server you didn't have to included the adapters of the prebid server bidders

I thought it's been that way forever, which is why I recently opened the issue [Remove requirement for client-side adapter to run server-side](https://github.com/prebid/Prebid.js/issues/6361)

@ChrisHuie, please hunt down what changed here?

Without the adapter code, PBJS doesn't know the mediatypes, permanent aliases, or GVL ID.

spormeon:
[...]

PR #7711, opened 2021-11-15
Prebid Core: Restore use of server-side adapter without client-side adapter

jorgeluisrocha:
## Type of change

- Bugfix

## Description of change
Fixes a bug that requires client-side adapter codes to be imported for server-side bidding. This obviates the need to import client-side adapter specifications. Additionally, this resolves the regression introduced by the original version of this PR found in #7662. 

## Other information
Issue: #6361 
Issue: #6388

### DIFF
--- a/src/prebid.js
+++ b/src/prebid.js
@@ -556,11 +556,11 @@ $$PREBID_GLOBAL$$.requestBids = hook('async', function ({ bidsBackHandler, timeo
 
   let _s2sConfigs = [];
   const s2sBidders = [];
-  config.getConfig('s2sConfig', config => {
-    if (config && config.s2sConfig) {
-      _s2sConfigs = Array.isArray(config.s2sConfig) ? config.s2sConfig : [config.s2sConfig];
-    }
-  });
+  const s2sConfig = config.getConfig('s2sConfig');
+
+  if (s2sConfig) {
+    _s2sConfigs = Array.isArray(s2sConfig) ? s2sConfig : [s2sConfig];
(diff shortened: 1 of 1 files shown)


FilipStamenkovic (changes requested):
@jorgeluisrocha Looking at the test coverage, not a single test goes through your changes:

image


Can you add at least one test for the changes you did?

ChrisHuie:
@jorgeluisrocha following up on this pr. Are you still working on this pr to add a little test coverage?

patmmccann:
@karimMourra can you weigh in here? this seems stalled

karimMourra:
Hi @patmmccann , Jorge will add unit tests to this PR when he has enough bandwidth

ChrisHuie:
@karimMourra @jorgeluisrocha following up on this pr. Is it okay if I close this pr for now and it is just reopened when bandwidth provides

bretg:
This PR is related to issue https://github.com/prebid/Prebid.js/issues/6361

For the record, client-side bid adapters defines 4 things that could be important to server behavior:

1. MediaTypes
2. Aliases
3. GVLID
4. transformBidParams

#1-3 are ok -- either they can be defined in s2sConfig as needed or they're also defined server-side.

A few bidAdapters do however, require transformBidParams() Those ones may not work well if the client-side adapter is not present.

dgirardi:
This was incidentally fixed in #8357; just verified that no client side adapters are needed now for interfacing with Server.

Issue #7280, opened 2021-08-09
Bid adapters duplicating pubcid/sharedid functionality


patmmccann:
## Type of issue
Several bid adapters duplicate the functionality of SharedId in which an id is set in the publishers' first party domain. I believe this behavior should be opt in and in tight control of the publisher. The module rules say this behavior is reserved for ID submodules, but in the RTD module rules section. I believe this rule should extend to bid adapters as well

#7279 is an example of a PR to add this functionality. 

Here are the adapters that appear to do it already:
https://github.com/prebid/Prebid.js/blob/12f461093b15657e11bedacba91319db873201b0/modules/datablocksBidAdapter.js#L99
https://github.com/prebid/Prebid.js/blob/0ca2a83a28ecdcfba37035f01c34aa19a64932bb/modules/adbookpspBidAdapter.js#L625
https://github.com/prebid/Prebid.js/blob/481b94f8c27d535d7b9105a69e76d01bad49a64b/modules/jixieBidAdapter.js#L30
https://github.com/prebid/Prebid.js/blob/412277f72b97cab72644cb4a009f58067a165899/modules/widespaceBidAdapter.js#L235
https://github.com/prebid/Prebid.js/blob/3e71ab46dabced5a17716b548a64fb818af268ef/modules/unicornBidAdapter.js#L138
https://github.com/prebid/Prebid.js/blob/0492821b8ce055cac419ad00ac90eee1b66e5e92/modules/relaidoBidAdapter.js#L245
https://github.com/prebid/Prebid.js/blob/412277f72b97cab72644cb4a009f58067a165899/modules/ucfunnelBidAdapter.js#L288 

One adapter does it in their tests
https://github.com/prebid/Prebid.js/blob/a633a637db703f90cb7dda9b160ea8b7fc401113/test/spec/modules/gmosspBidAdapter_spec.js#L65

Another adapter writes their own set cookie method in tests, and testing indicates they do this in the wild as well on winning bids
https://github.com/prebid/Prebid.js/blob/a633a637db703f90cb7dda9b160ea8b7fc401113/test/spec/modules/kargoBidAdapter_spec.js#L115

This adapter test implies this behavior as well
https://github.com/prebid/Prebid.js/blob/a5e7d08950f33301372942a64333522474285895/test/spec/modules/invibesBidAdapter_spec.js#L201

[...]

PR #7712, opened 2021-11-16
Prebid Core: refactor bidderSettings to have only one entry point

dgirardi:
## Type of change

## Description of change

Add thin wrapper around `$$PREBID_GLOBAL$$.bidderSettings` that reduces some duplicate logic around fallback to `bidderSettings.standard`

This was inspired by https://github.com/prebid/Prebid.js/issues/7280, which proposes using `bidderSettings` for a new purpose.

Note that I found some horrors lurking deep in the darkness; see the PR comments on the code changes.


test/spec/modules/sortableAnalyticsAdapter_spec.js
```diff
@@ -203,11 +202,11 @@ describe('Sortable Analytics Adapter', function() {
         brc: 1,
         brid: ['10141593b1d84a', '37a8760be6db23'],
         rs: ['300x250', '728x90'],
-        btcp: [0.70, 0.50],
+        btcp: [0.70 * 0.95, 0.50 * 0.95],
         btcc: 'USD',
         btin: true,
         btsrc: 'sortable',
-        c: [0.70, 0.50],
+        c: [0.70 * 0.95, 0.50 * 0.95],
```
dgirardi, line 209: this suite sets up `bidCpmAdjustment` by mocking `getGlobal`, which was not seen by `adjustBids` in `auction.js` because that looked directly at `$$PREBID_GLOBAL$$.bidderSettings`. This test never caught it, but there are other assertions in this suite that do expect numbers to get through the adjustment, so we need to modify either the setup or the assertion.

src/auction.js
```diff
@@ -661,14 +662,13 @@ export function getStandardBidderSettings(mediaType, bidderCode) {
...
 
-  let bidderSettings = $$PREBID_GLOBAL$$.bidderSettings;
-  if (!bidderSettings[CONSTANTS.JSON_MAPPING.BD_SETTING_STANDARD]) {
-    bidderSettings[CONSTANTS.JSON_MAPPING.BD_SETTING_STANDARD] = {};
-  }
-  if (!bidderSettings[CONSTANTS.JSON_MAPPING.BD_SETTING_STANDARD][CONSTANTS.JSON_MAPPING.ADSERVER_TARGETING]) {
-    bidderSettings[CONSTANTS.JSON_MAPPING.BD_SETTING_STANDARD][CONSTANTS.JSON_MAPPING.ADSERVER_TARGETING] = [
+  if (!standardSettings[CONSTANTS.JSON_MAPPING.ADSERVER_TARGETING]) {
+    standardSettings[CONSTANTS.JSON_MAPPING.ADSERVER_TARGETING] = [
+      // NOTE: some callers (probably tests) are modifying these - they are not safe to cache
```
dgirardi, line 671: I tried generating these once as the original version did, but someone is messing with the return value. I was not able to find who that is or why that was not a problem before my change.

src/auction.js
```diff
@@ -691,7 +692,7 @@ export function getStandardBidderSettings(mediaType, bidderCode) {
     });
 
     // Adding hb_cache_host
-    if (config.getConfig('cache.url') && (!bidderCode || deepAccess(bidderSettings, `${bidderCode}.sendStandardTargeting`) !== false)) {
+    if (config.getConfig('cache.url') && (!bidderCode || bidderSettings.get(bidderCode, 'sendStandardTargeting') !== false)) {
```
dgirardi, line 695: This is potentially a breaking change: previously, this would be evaluated the first time this function was called with `mediaType === 'video'` and would set `hb_cache_host` globally after that. There was also no fallback to `bidderSettings.standard`.

src/auction.js
```diff
@@ -711,19 +712,15 @@ export function getKeyValueTargetingPairs(bidderCode, custBidObj, bidReq) {
...
-      setKeys(keyValues, bidderSettings[bidderCode], custBidObj, bidReq);
-      custBidObj.sendStandardTargeting = bidderSettings[bidderCode].sendStandardTargeting;
-    }
+  const standardSettings = getStandardBidderSettings(custBidObj.mediaType, bidderCode);
+  setKeys(keyValues, standardSettings, custBidObj, bidReq);
+
+  // 2) set keys from specific bidder setting override if they exist
+  if (bidderCode && bidderSettings.getOwn(bidderCode, CONSTANTS.JSON_MAPPING.ADSERVER_TARGETING)) {
+    setKeys(keyValues, bidderSettings.ownSettingsFor(bidderCode), custBidObj, bidReq);
+    custBidObj.sendStandardTargeting = bidderSettings.get(bidderCode, 'sendStandardTargeting');
```
dgirardi, line 723: Another potentially breaking change: the logic here was a bit confusing, copying `sendStandardTargeting` only if `adserverTargeting` is defined, which was also overriding `standard.sendStandardTargeting`. From the public documentation it seems like we should be using the `standard` setting if there is bidder-specific one.

src/auction.js
robertrmartinez: Breaking change... more like a bug fix!

src/auction.js
robertrmartinez: This was something I recently fixed.

https://github.com/prebid/Prebid.js/pull/7607

The default way this function is called is through https://github.com/prebid/Prebid.js/blob/4c424a06883a421bb25dd96d1cf8fbda90741e9c/src/auction.js#L674

This always passes undefined. And for these KVP sets it needs access to the bidReq.mediaType.

So when KVP's are set, it actually always runs the OR `getPriceGranularity(bid.mediaType, bidReq);`

Which will in turn select the right granularity.

There are only two other places which use this function, and they directly pass in the granularity, which is why we had to support the `granularity || granularity`

https://github.com/prebid/Prebid.js/blob/master/modules/adpod.js#L126

https://github.com/prebid/Prebid.js/blob/master/modules/konduitWrapper.js#L104

So this is actually fine with the current way this is setup, but yours is probably smarter.

Also the entire Targeting logic is crazy and needs to be rewritten! haha.

robertrmartinez:
Thanks @dgirardi 

A Much needed update.

I will take some time to run some tests which use Bidder Settings and do a formal review sometime this week.

It is sort of a big change, so we should make sure to cover all the bases!

Lots of publishers rely on utilizing `pbjs.bidderSettings` and modifying it themselves.

src/auction.js
dgirardi: I don't think it's fine in general: the `|| getPriceGranularity(bid.mediaType, bidReq)` is only evaluated on the first call because `granularity` is in the closure scope. See the test case I added in `auctionmanager_spec`, that fails without this change.

robertrmartinez: I get it. Thanks!

Rothalack (approved):
This looks good to me, working in testing.

I think the documentation is still accurate other than maybe we should add mention of caching to that documentation.

src/utils.js
```diff
@@ -1299,5 +1299,49 @@ export function cyrb53Hash(str, seed = 0) {
...
+  return function () {
+    return f(null);
+  }
+}
+
+/**
+ * Return an object that is the result of recursively merging properties from the given objects. If multiple
+ * objects define the same "leaf" properties, the right-most's value is picked.
+ */
+export function deepMerge(left, right) {
```
dgirardi, line 1332: I just learned that there is already a `mergeDeep` in here. I'll post a revision early next week to slim this down some.

ChrisHuie:
@robertrmartinez can you look at this pr when you get a chance 🙏

Rothalack:
Hey @robertrmartinez , will you be able to review this PR? Don't want to let it closed.

src/utils.js
```diff
@@ -1326,5 +1326,28 @@ export function cyrb53Hash(str, seed = 0) {
...
   const bidderSettings = getGlobal().bidderSettings;
   return ((bidderSettings[bidderCode] && bidderSettings[bidderCode].allowZeroCpmBids === true) ||
-   (bidderSettings.standard && bidderSettings.standard.allowZeroCpmBids === true));
+    (bidderSettings.standard && bidderSettings.standard.allowZeroCpmBids === true));
+}
+
+/**
+ * Return a memoized version of the given function that uses the first argument as cache key.
+ */
+export function memoize1(fun) {
```
FilipStamenkovic, line 1335: Why is this called `memoize1`?  Why not just `memoize`?

src/utils.js
```diff
@@ -1326,5 +1326,28 @@ export function cyrb53Hash(str, seed = 0) {
...
+      cache[a0] = fun(a0, ...rest);
+    }
+    return cache[a0];
+  }
+}
+
+/**
+ * Return a memoized version of the given zero-arg function.
+ */
+export function cache(fun) {
```
FilipStamenkovic, line 1348: Do we need both cache and memoize?  
It seems to me that just `memoize` function is enough?

FilipStamenkovic (changes requested):
This pull request has conflicts. Can you pull/rebase your branch and resolve this conflicts? Also, take a look at my other comments.

src/bidderSettings.js
```diff
@@ -0,0 +1,67 @@
+import {deepAccess, mergeDeep} from './utils.js';
+import {getGlobal} from './prebidGlobal.js';
+
+const CONSTANTS = require('./constants.json');
+
+export function ScopedSettings(getSettings, defaultScope) {
```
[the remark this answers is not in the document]
osazos: Is there a reason to not using Class if you choose OOP?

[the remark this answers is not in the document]
dgirardi: > Is there a reason to not using Class if you choose OOP?

Because when I started working on this IE11 was still a target :) I don't have a strong preference on the syntax, I can rewrite using `class` - should I take your comment as an encouragement?

> I'm not big fan of OOP concept.

Then some day I'll have to explain to you why you're wrong.

But more seriously, I don't understand how you can say we barely use OOP - config, auctionManager, auction, adapterManager, adapters, etc are all objects with capabilities. Is the problem that I use `prototype` instead of hidden closures as "private" methods? or the fact that you have to use `new` for this function but not for the others?

src/utils.js
dgirardi: because it only uses the first argument as the memoization key.

src/bidderSettings.js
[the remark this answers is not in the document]
osazos: >  should I take your comment as an encouragement?

Yes, totally :)

src/auction.js
```diff
@@ -642,47 +643,51 @@ export const getAdvertiserDomain = () => {
...
+      ? function (bidResponse, bidReq) {
+        return value(bidResponse, bidReq);
+      }
+      : function (bidResponse) {
+        return getValue(bidResponse, value);
+      }
+  };
+}
+
+const defaultAdserverTargeting = cache(() => {
```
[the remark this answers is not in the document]
dgirardi: Both `getPriceByGranularity` and `getAdvertiserDomain` should return equivalent closures no matter what the config is (the return value is what looks it up) so caching should not be a problem.

But I do think there's a good argument to be made for not caching - I am doing it here because it was in the original code, which over time got broken because more stuff was added in a non-cacheable way (in `getStandardBidderSettings`). 

So now we still have to recalculate the "standard" settings for each auction, plus we have to be mindful of what is being cached to avoid making unintentional changes to global state. I doubt the added complexity is worth the tiny performance improvement, so I also vote for removing the cache.

.eslintrc.js
```diff
@@ -23,10 +23,13 @@ module.exports = {
     'BROWSERSTACK_USERNAME': false,
     'BROWSERSTACK_KEY': false
   },
+  // use babel as parser for fancy syntax
+  parser: '@babel/eslint-parser',
```
FilipStamenkovic, line 27: Why do we need `@babel/eslint-parser`?  
Looking at the npm description I see:

> ESLint's default parser and core rules only support the latest final ECMAScript standard and do not support experimental (such as new features) and non-standard (such as Flow or TypeScript types) syntax provided by Babel

Do we need these 'experimental and non-standard' syntax? Or am I missing something?

modules/appnexusBidAdapter.js
```diff
@@ -292,7 +293,7 @@ export const spec = {
       serverResponse.tags.forEach(serverBid => {
         const rtbBid = getRtbBid(serverBid);
         if (rtbBid) {
-          const cpmCheck = (isAllowZeroCpmBidsEnabled(bidderRequest.bidderCode)) ? rtbBid.cpm >= 0 : rtbBid.cpm > 0;
+          const cpmCheck = (bidderSettings.get(bidderRequest.bidderCode, 'allowZeroCpmBids') === true) ? rtbBid.cpm >= 0 : rtbBid.cpm > 0;
```
FilipStamenkovic, line 296: From readability perspective, to me, personally, previous version was much clearer.  
You just execute some function to get the info you need (`isAllowZeroCpmBidsEnabled`).  
With this new approach code is longer and it's more error prone since you have to remember name of the property `allowZeroCpmBids`.  

But, I guess since that is used in 2,3 places in the entire prebid.js than it's fine.

package.json
```diff
@@ -30,6 +30,7 @@
   },
   "devDependencies": {
     "@babel/core": "^7.8.4",
+    "@babel/eslint-parser": "^7.16.5",
```
FilipStamenkovic, line 33: I think adding new npm package should have been done in separate PR. The reason is not to pollute this PR with those changes.

FilipStamenkovic (approved):

.eslintrc.js
dgirardi: It's part of the `class` syntax, specifically field declaration and private (#) methods and fields. Btw I care about objects but not really about syntax, I don't mind going either way with this.

modules/appnexusBidAdapter.js
dgirardi: I don't disagree but there's a problem with circular imports when we put everything in `utils`. (new `bidderSettings` needs `utils.deepClone`, and a few others)

Issue #6654, opened 2021-04-26
Native: image min-width/min-height attributes


coreyb-cbs:
## Description
Provide the ability to include `min_width` and `min_height` on the _image_ asset without requiring `ratio_width` and `ratio_height`. OpenRTB native spec differs from Prebid native spec in regards to _min-width_ and _min-height_ attributes for image assets. ORTB includes _image.wmin_ and _image.hmin_. Prebid has attributes that represent the same thing, but they're nested inside the _aspect_ratios_ attribute and _aspect_ratios.ratio_width_ and _aspect_ratios.ratio_height_ are required if the _aspect_ratios_ option is included. Both publishers and bidders would like to have the ability to include _min-width_ and _min-height_ attributes on _image_ assets without being forced to include an aspect ratio. Including an aspect ratio reduces the potential bid pool. GAM native styles, especially of Fluid size, can responsively handle images of various dimensions and aspect ratios.

OpenRTB spec (_image_ asset is on page 15):
https://www.iab.com/wp-content/uploads/2018/03/OpenRTB-Native-Ads-Specification-Final-1.2.pdf

## Proposal
The _image_ asset could look like any of the following (all attrs optional):
```
image: {
     required: true,
     sizes: [300, 200]
}
```
```
image: {
     required: true,
     min_sizes: [300, 200]
}
```
```
image: {
     required: true,
     min_sizes: [300, 200],
     aspect_ratios: [3, 2]
}
```

## Note
The same could be applied to the _icon_ asset.

coreyb-cbs:
Or a simpler alternate proposal which doesn't require an attribute re-factor is to allow the use of `aspect_ratios.min_width` and `aspect_ratios.min_height` without requiring `aspect_ratios.ratio_width` and `aspect_ratios.ratio_height`. It doesn't feel proper that they're still nested inside `aspect_ratios`, but it's a less-disruptive modification.

So updating the following line:
https://github.com/prebid/Prebid.js/blob/master/modules/prebidServerBidAdapter/index.js#L500

to this:
`if (params.aspect_ratios.ratio_width && params.aspect_ratios.ratio_height) {`

And therefore update the error message in this line:
https://github.com/prebid/Prebid.js/blob/master/modules/prebidServerBidAdapter/index.js#L498

to this:
`throw 'invalid img sizes (must provide sizes or min_height & min_width)';`

ChrisHuie:
[...]

PR #7722, opened 2021-11-16
PBS Bid Adapter: do not pass aspectratios in ORTB2 ext data if native image definition doesn't have ratio_height/ratio_width

dgirardi:
## Type of change
Not sure: either a feature or a bugfix. See https://github.com/prebid/Prebid.js/issues/6654

## Description of change

When using Prebid Server you can specify AdUnits with native image or icon mediatypes that have min_width and min_height but no ratio_width or ratio_height. 
This is not supported according to https://docs.prebid.org/dev-docs/adunit-reference.html#adunitmediatypesnativeimageaspect_ratios, but it works for the PBS adapter, except it causes some incorrect data to be passed along in `ext`. 

This change cleans up `ext` for this case. It could be seen as a bugfix - validating/removing aspect ratio before sending it to PBS, or as a feature - support optional ratio_height and ratio_width.

### DIFF
--- a/modules/prebidServerBidAdapter/index.js
+++ b/modules/prebidServerBidAdapter/index.js
@@ -539,10 +539,13 @@ const OPEN_RTB_PROTOCOL = {
                 }
                 if (Array.isArray(params.aspect_ratios)) {
                   // pass aspect_ratios as ext data I guess?
-                  asset.ext = {
-                    aspectratios: params.aspect_ratios.map(
-                      ratio => `${ratio.ratio_width}:${ratio.ratio_height}`
-                    )
+                  const aspectRatios = params.aspect_ratios
+                    .filter((ar) => ar.ratio_width && ar.ratio_height)
+                    .map(ratio => `${ratio.ratio_width}:${ratio.ratio_height}`);
+                  if (aspectRatios.length === params.aspect_ratios.length) {
+                    asset.ext = {
+                      aspectratios: aspectRatios
+                    }
                   }
                 }
                 assets.push(newAsset({
--- a/modules/prebidServerBidAdapter/index.js
+++ b/modules/prebidServerBidAdapter/index.js
@@ -542,7 +542,7 @@ const OPEN_RTB_PROTOCOL = {
                   const aspectRatios = params.aspect_ratios
                     .filter((ar) => ar.ratio_width && ar.ratio_height)
                     .map(ratio => `${ratio.ratio_width}:${ratio.ratio_height}`);
-                  if (aspectRatios.length === params.aspect_ratios.length) {
+                  if (aspectRatios.length > 0) {
                     asset.ext = {
                       aspectratios: aspectRatios
                     }
--- a/test/spec/modules/prebidServerBidAdapter_spec.js
+++ b/test/spec/modules/prebidServerBidAdapter_spec.js
@@ -8,6 +8,7 @@ import events from 'src/events.js';
 import CONSTANTS from 'src/constants.json';
 import { server } from 'test/mocks/xhr.js';
 import { createEidsArray } from 'modules/userId/eids.js';
+import {deepAccess, deepClone} from 'src/utils.js';
 
 let CONFIG = {
   accountId: '1',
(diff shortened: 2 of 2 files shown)


ChrisHuie (approved):

ChrisHuie (approved):

modules/prebidServerBidAdapter/index.js
```diff
@@ -539,10 +539,13 @@ const OPEN_RTB_PROTOCOL = {
...
                 if (Array.isArray(params.aspect_ratios)) {
                   // pass aspect_ratios as ext data I guess?
-                  asset.ext = {
-                    aspectratios: params.aspect_ratios.map(
-                      ratio => `${ratio.ratio_width}:${ratio.ratio_height}`
-                    )
+                  const aspectRatios = params.aspect_ratios
+                    .filter((ar) => ar.ratio_width && ar.ratio_height)
+                    .map(ratio => `${ratio.ratio_width}:${ratio.ratio_height}`);
+                  if (aspectRatios.length === params.aspect_ratios.length) {
```
aleksatr, line 545: hey @dgirardi, can you help me understand why is this condition comparing length of the original aspect_ratios array with filtered/mapped aspectRatios? Shouldn't we just check if `aspectRatios.length > 0`?

Also L534 and L535 are reading wmin and hmin from the first aspect_ratios element, is this valid given that aspect_ratios is an array, and that there might be some other elements of that array with min_width & min_height?

aleksatr: well, sizes property of the native asset is not really meant to be an array, at least according to the ORTB, even though it's defined as such in the pbjs docs. Since prebid server adapter is translating prebid config to ORTB request, where image asset can have only one wmin and hmin, I can only assume that sizes & aspect_ratios are defined as arrays to leave a possibility for other adapters that are not tied to the ORTB spec. 

When testing with only sizes defined for img asset (without aspect_ratios), this is what I found:

sizes: [989, 742] - works as intended
sizes: [[989, 742]] - does not work (error creating native request: invalid img sizes (must provide sizes or min_height & min_width if using aspect_ratios))
sizes: [[989, 742], [981, 741]] - works but not as intended (outgoing request contains "w":[3000,2250],"h":[3001,2251])

But this is another issue with native in pbs adapter, and there are many more in the way native is handled by pbs adapter, I'd say. So to address the actual issue this PR tries to solve, in the outgoing request pbs adapter makes, the only array in the img asset object is the `ext.aspectratios`, so there's nothing in this asset that can correspond 1:1 to this array, that's why I suggest we replace the restriction `aspectRatios.length === params.aspect_ratios.length` with `... > 0`.

dgirardi: My instinct is that's not right, but I can't verify without seeing how `ext.aspectratios` is used out in the wild right now, and of course it was not right to begin with so I can't object too much.

The reason is so far if `ext.aspectratios` was present, the receiver could look at `ext.aspectratios[0]` to see what the aspect ratio is (and realistically, if anyone is looking at this field right now, that's what they have to be doing - the rest of the array wouldn't make sense as far as I can tell). That would no longer necessarily be true if we use `> 0` here; I'd expect it to break consumers if there are any - but I'll trust your judgement on the impact of this change.

Stepping back, if in ORTB2 there's really no place for more than one dimension, passing an array here doesn't make sense, does it? what do you think about adding `ext.ratio_width` and `ext.ratio_height` instead (or in addition)?

aleksatr: Not sure what receiver you are referencing here, receiver of the request object or the response object? :) If we're talking about request, why would they look exactly at `ext.aspectratios[0]`, it's an array so it can offer more than one aspect ratio to the buy side? It anyway doesn't make sense to set `ext.aspectratios` if you have specified `w` & `h`, that's a hard requirement, to me it only makes sense in combination with `hmin` and `wmin`, but again, `ext.aspectratios` is not something that exists in ORTB standard, so I can only guess how someone uses it (or if it's being used at all).

If we're talking about the response, we have an image response object ([page 22](https://iabtechlab.com/wp-content/uploads/2016/07/OpenRTB-Native-Ads-Specification-Final-1.2.pdf)), with `w` & `h` and if aspect ratio is needed, it can be figured out based on w&h alone w/gcd(w, h) : h/gcd(w, h).

Regarding your suggestion for adding `ext.ratio_width` and `ext.ratio_height`, it is a bit better fit for the current ORTB structure (it can combined with wmin and hmin), but I'm not sure if anybody on the buy side would be consuming this (obviously nobody now, since nobody would know about it).

Anyway, to go back to the original remark I had on this PR, I would like to consult a second reviewer if he also thinks that this condition should be as is `aspectRatios.length === params.aspect_ratios.length`, or `... > 0`, and then we can merge this one.

ChrisHuie: I'm not too sure how this is used in the wild, outside of the issue that was created around it either. I would lean a bit more toward `aspectRatios.length > 0` though because I think that is probably less likely to break various possible implementations. I wouldn't say I am an expert on this either though.

aleksatr (approved):

PR #7729, opened 2021-11-17
RTD Module: add 'onBidRequest' event handler for RTD submodules

dgirardi:
## Type of change

- [x] Feature

## Description of change

Add new `onBidRequest` event handler for RTD submodules

## Other information

Addresses ﻿https://github.com/prebid/Prebid.js/issues/7078
Documentation changes: https://github.com/prebid/prebid.github.io/pull/3416

### DIFF
--- a/modules/rtdModule/index.js
+++ b/modules/rtdModule/index.js
@@ -99,6 +99,15 @@
  */
 
 /**
+ * @function?
+ * @summary on bid requested event
+ * @name RtdSubmodule#onBidRequestEvent
+ * @param {Object} data
+ * @param {SubmoduleConfig} config
+ * @param {UserConsentData} userConsent
+ */
+
+/**
  * @interface ModuleConfig
  */
 
@@ -164,8 +173,10 @@ let _dataProviders = [];
 let _userConsent;
 
 /**
- * enable submodule in User ID
+ * Register a RTD submodule.
+ *
  * @param {RtdSubmodule} submodule
+ * @returns {function()} a de-registration function that will unregister the module when called.
  */
 export function attachRealTimeDataProvider(submodule) {
   registeredSubModules.push(submodule);
--- a/modules/rtdModule/index.js
+++ b/modules/rtdModule/index.js
@@ -143,7 +143,7 @@
 
 import {config} from '../../src/config.js';
 import {module} from '../../src/hook.js';
-import { logError, logWarn } from '../../src/utils.js';
+import {logError, logWarn} from '../../src/utils.js';
 import events from '../../src/events.js';
 import CONSTANTS from '../../src/constants.json';
 import {gdprDataHandler, uspDataHandler} from '../../src/adapterManager.js';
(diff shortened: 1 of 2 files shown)


modules/rtdModule/index.js
```diff
@@ -169,8 +169,43 @@ let _userConsent;
  */
 export function attachRealTimeDataProvider(submodule) {
   registeredSubModules.push(submodule);
+  return function detach() {
+    const idx = registeredSubModules.indexOf(submodule)
+    if (idx >= 0) {
+      registeredSubModules.splice(idx, 1);
+      initSubModules();
+    }
+  }
```
Fawke, line 178: I'm not sure I understand why we need to return the detach function and call initSubmodules func again? I understand that it's getting call in the unit tests file, but does it make sense to have it here as a return value of `attachRealTimeDataProvider` function. What are your thoughts?

modules/rtdModule/index.js
```diff
@@ -169,8 +169,43 @@ let _userConsent;
...
+            } catch (e) {
+              logError(`RTD provider '${sm.name}': error in '${handler}':`, e);
+            }
+          });
+        })
+      });
+      registered = true;
+    }
+  }
+})();
```
Fawke, line 207: Nice code organisation to avoid calling and defining each handler separately.

test/spec/modules/realTimeDataModule_spec.js
```diff
@@ -58,33 +60,87 @@ const conf = {
...
-    expect(rtdModule.subModules).to.eql([validSMWait, validSM]);
+  after(() => {
+    sandbox.restore();
   });
 
-  it('should be able to modify bid request', function (done) {
-    rtdModule.setBidRequestsData(() => {
-      assert(getBidRequestDataSpy.calledTwice);
-      assert(getBidRequestDataSpy.calledWith({bidRequest: {}}));
+  describe('', () => {
```
Fawke, line 85: Any name for the describe block? Usually it helps to have a meaningful name especially when the test fails.

Fawke (changes requested):
Hi @dgirardi ,

Thanks for the PR. This looks good overall, just few places where I'm a bit confused, but rest looks good. Thanks!

modules/rtdModule/index.js
dgirardi: I think this is the best way to do it, but I'll take suggestions. Do you prefer a separate global "deinit" function, or something else?
(I think this is the way because an undocumented return value that noone is using at the moment is less ugly than a public "resetSubmodules" function that is only called by the tests. In fact, the real way to do it would be to de-globalize the whole module, but I didn't feel like committing to a huge refactor for this issue).

test/spec/modules/realTimeDataModule_spec.js
dgirardi: I tried, I couldn't find one that's better than no name. These are the previously existing tests that I grouped together because I don't need the same setup for my new ones. The only name I can think that makes sense is something like "old tests" which would not help at all on failure.

modules/rtdModule/index.js
Fawke: Sure, I somewhat agree with what you're saying. I don't have a strict preference for either of the approaches here, between yours and the global func 'resetSubmodules'. My concern was the func name, does it represent what is being implemented inside. Would it be possible to at least update the description of the function and document the return value.

test/spec/modules/realTimeDataModule_spec.js
Fawke: That's fine. Didn't know you clubbed old tests inside this describe block.

modules/rtdModule/index.js
dgirardi: You're right - updated the docs.

Fawke (approved):
LGTM

ChrisHuie (approved):

omerBrowsi:
@dgirardi @Fawke @ChrisHuie seems like this PR removed the call for `getAdUnitTargeting`, causing RTD modules using `getTargetingData` not to work properly.

PR #7732, opened 2021-11-17
CircleCi config: change to npm ci in unit testing workflow

ChrisHuie:
CircleCi still uses 'npm install' in the unit testing job that automatically runs after commits. This changes to 'npm ci' for consistency with docs

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -52,7 +52,7 @@ aliases:
   - &unit_test_steps
     - checkout
     - restore_cache: *restore_dep_cache
-    - run: npm install
+    - run: npm ci
     - save_cache: *save_dep_cache
     - run: *install
     - run: *setup_browserstack


Rothalack:
I read up on the difference between ci and install, I only have one thought/concern. This will always work except in the case where we actually do want to do a version update or add a new dependency. Wouldn't npm ci not properly install that new dependency/update? I could, of course, be misunderstanding the difference. Other than that question, I think this is good.

Rothalack:
Hey @ChrisHuie , just following up on this. Had you considered my question above or am I misunderstanding npm ci vs install? I don't want to hold this up over nothing.

patmmccann:
@Rothalack this failure to update is actually intended as per https://github.com/prebid/Prebid.js/pull/5697

Rothalack (approved):
sounds good to me!

PR #7739, opened 2021-11-18
Prebid core: move generation of 'installedModules' to babel

dgirardi:
## Type of change

- [x] Build related changes

## Description of change

﻿`pbjs.installedModules` does not work correctly when using prebid as an npm dependency (https://github.com/prebid/Prebid.js/issues/7287), because it's generated by a gulp task.

This moves generation of `installedModules` to the `pbjsGlobals` babel plugin that we ask npm consumers to use for prebid.

### DIFF
--- a/src/prebid.js
+++ b/src/prebid.js
@@ -47,8 +47,7 @@ $$PREBID_GLOBAL$$.libLoaded = true;
 $$PREBID_GLOBAL$$.version = 'v$prebid.version$';
 logInfo('Prebid.js v$prebid.version$ loaded');
 
-// modules list generated from build
-$$PREBID_GLOBAL$$.installedModules = ['v$prebid.modulesList$'];
+$$PREBID_GLOBAL$$.installedModules = $$PREBID_GLOBAL$$.installedModules || [];
 
 // create adUnit array
 $$PREBID_GLOBAL$$.adUnits = $$PREBID_GLOBAL$$.adUnits || [];
--- a/gulpfile.js
+++ b/gulpfile.js
@@ -141,12 +141,6 @@ function watch(done) {
   done();
 };
 
-function makeModuleList(modules) {
-  return modules.map(module => {
-    return '"' + module + '"'
-  });
-}
-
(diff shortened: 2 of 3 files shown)


snapwich:
i like this better than the original solution. it does differ in one area though in that the way it was originally coded all the `installedModules` would be defined before the modules were included in the bundle, but now they'll only be fully defined after they're included in the bundle. that's only really a problem if `pbjs.installedModules` were being used in the main body of a module though, which the only use of it i see is here: https://github.com/prebid/Prebid.js/blob/master/modules/rubiconBidAdapter.js#L218 which is in `spec.buildRequests` so should be fine in that case.

probably should check with @mmoschovas that it works as they intended, and also that it works for @muuki88 who reported the bug. if it works for them then it seems fine to me.

good work.

snapwich (approved):

dgirardi:
I think it should be possible to reproduce current behavior exactly by adding a second babel plugin: first collect all modules, then drop in the global definition. But I doubt it's worth it, and it may break babel<=6 npm consumers since they need to specify each plugin in their webpack configuration.

src/prebid.js
```diff
@@ -47,8 +47,7 @@ $$PREBID_GLOBAL$$.libLoaded = true;
 $$PREBID_GLOBAL$$.version = 'v$prebid.version$';
 logInfo('Prebid.js v$prebid.version$ loaded');
 
-// modules list generated from build
-$$PREBID_GLOBAL$$.installedModules = ['v$prebid.modulesList$'];
+$$PREBID_GLOBAL$$.installedModules = $$PREBID_GLOBAL$$.installedModules || [];
```
muuki88, line 50: Oh, I like that change! :heart:

muuki88 (commented):
I'll try to test this locally! Thanks a lot for the effort :hugs:

muuki88 (approved):
Works as intended!

mmoschovas (approved):
lgtm

Issue #6354, opened 2021-02-24
Prebid Server Adapter cookie sync callback blocks on iframe syncs


muuki88:
## Type of issue

Prebid server user sync only syncs until the first iframe configuration.

## Steps to reproduce

1. Prebid Server with at least one bid adapter that has an iframe sync (e.g. Unruly)
2. Configure prebid server adapter

Cookie sync example response.

```json
{
  "status": "no_cookie",
  "bidder_status": [
    {
      "bidder": "unruly",
      "no_cookie": true,
      "usersync": {
[...]

PR #7744, opened 2021-11-22
PBS Bid adapter: timeout user syncs if they never load

dgirardi:
## Type of change

- [x] Bugfix

## Description of change
﻿In some situations the browser never triggers load/error events for user syncs, so the PBS adapter remains stuck (https://github.com/prebid/Prebid.js/issues/6354). This adds a timeout for image/iframe syncs.

## Other information

Doc changes: https://github.com/prebid/prebid.github.io/pull/3424

### DIFF
--- a/src/utils.js
+++ b/src/utils.js
@@ -483,15 +483,38 @@ export function insertElement(elm, doc, target, asLastChildChild) {
 }
 
 /**
+ * Returns a promise that completes when the given element triggers a 'load' or 'error' DOM event, or when
+ * `timeout` milliseconds have elapsed.
+ *
+ * @param {HTMLElement} element
+ * @param {Number} [timeout]
+ * @returns {Promise}
+ */
+export function waitForElementToLoad(element, timeout) {
+  return new Promise((resolve) => {
+    const onLoad = function() {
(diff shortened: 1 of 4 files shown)


src/utils.js
```diff
@@ -482,16 +482,39 @@ export function insertElement(elm, doc, target, asLastChildChild) {
...
+ *
+ * @param {HTMLElement} element
+ * @param {Number} [timeout]
+ * @returns {Promise}
+ */
+export function waitForElementToLoad(element, timeout) {
+  return new Promise((resolve) => {
+    const onLoad = function() {
+      element.removeEventListener('load', onLoad);
+      element.removeEventListener('error', onLoad);
```
msm0504, line 497: I'm not sure if it matters, but it's probably safer to call `clearTimeout` in this function, too.

msm0504 (changes requested):

src/utils.js
dgirardi: done. fwiw, the `Promise` spec should require `resolve` to do nothing when called a second time.

msm0504 (approved):

ChrisHuie (approved):

PR #7756, opened 2021-11-24
Brandmetrics RTD Module: add new RTD module

johanbrandmetrics:
## Type of change

- [ ] Bugfix
- [x] Feature
- [ ] New bidder adapter   
- [ ] Build related changes
- [ ] CI related changes
- [ ] Other

## Description of change

This module is intended to be used by brandmetrics (https://brandmetrics.com) partners and sets targeting keywords to bids if the browser is eligeble to see a brandmetrics survey.
The module hooks in to brandmetrics events and requires a brandmetrics script to be running. The module can optionally load and initialize brandmetrics by providing the a 'scriptId'- parameter.

- Docs PR: https://github.com/prebid/prebid.github.io/pull/3427


patmmccann:
A few issues here:

Initialize function doesnt seem to check for purpose 7 consent to the
brandmetrics gvl id if gdpr applies
Doesnt check usp string either
Please try and set something more generic in the ortb2 object; it isn't
okay to favor ozone even if that is the only bidder currently capable of
understanding this parameter.
Doesnt use
https://github.com/prebid/Prebid.js/blob/b815976146d912c533395c71a9e9679fc62ae536/src/adloader.js#L21
function

johanbrandmetrics:
Thank you for the feedback, a question before proceeding to update the pull- request.

According to the module- rules a RTD- module can load external code if it requires publisher registration which is the case for the brandmetrics- script. Will updating the docs to clearly disclose this suffice?

aleksatr:
@johanbrandmetrics you should be using the function @patmmccann mentioned to load the external script. Do you have an example of this script that we can take a look at, with explanation of what it does (`https://cdn.brandmetrics.com/survey/script/00000000-0000-0000-0000-000000000000.js` is not working).

@patmmccann how can this script be verified by pbjs community when it's not a single script, the actual script depends on the config parameter (dynamic)?

johanbrandmetrics:
@aleksatr 

To clarify, I'm refering to the quouted text of the module rules.

The script is as you say dynamic, not only differing between our clients, but also differs depending on the configurations made in our system (which changes over time). The quoted text suggests that proprietary libraries are ok if it requires registration and depending on how the module is labeled.

> Real Time Data Module Rules
> All global rules apply with one exception:
> A Real-Time Data module may load external code if it requires publisher registration and there’s a prominent disclosure on the module documentation. The idea is that a publisher will not include the module if they don’t approve of the external code, and since they’ve registered for the service, they must approve. The text of the disclosure may differ if the vendor allows Prebid to do regular reviews of a strictly versioned proprietary library.

patmmccann:
For RTD modules, many of them only make sense in the context of loading external, difficult to verify scripts.The section of the rules you quoted was written to confirm we know that. Largely we solve concerns about the external script through documentation and disclosure. 

However, you should still be able to use the utility function to load the script, correct?

johanbrandmetrics:
Makes sense.

I have updated the pull- request based on your feedback and suggestions:

* Check TCF and USP
* Set more generic bidder- level targeting not favouring any specific bidder
* Use loadExternalScript- function to add external script

aleksatr (approved):

aleksatr:
LGTM, @patmmccann anything you want to add?

patmmccann:
Looks great! Thanks for incorporating the feedback!

Issue #7156, opened 2021-07-08
CMP timeout scenario: auction without consent not running despite GDPR Enforcement Module rules


EskelCz:
## Type of issue
Bug or documentation issue

## Description
We need the auction to run even when the CMP banner stays open, relying on the "basic ads" legitimate interest. In Prebid 3 this was working with the allowAuctionWithoutConsent. In Prebid 4 and 5 I can't find a way to enable the same behavior. I've read that it should be possible by using the GDPR Enforcement module, so I have included it's configuration:
```
            defaultGdprScope: true,
            rules: [{
              purpose: 'storage',
              enforcePurpose: false,
              enforceVendor: false
            }, {
              purpose: 'basicAds',
              enforcePurpose: false,
              enforceVendor: false
            }]
```
It made no difference, with CMP banner open the Prebid triggers `cmpFailed` and stops the auction. The enforcement module code wasn't even touched, from my debugging.

Relevant logs:
- Prebid INFO: Received a response from CMP t {cmpId: 7, cmpVersion: 1, gdprApplies: true, tcfPolicyVersion: 2, eventStatus: "cmpuishown", …
...long delay
- Prebid ERROR: CMP workflow exceeded timeout threshold. Canceling auction as per consentManagement config. undefined

CMP returns event from TCF API, it gets processed with this function:
```
  function v2CmpResponseCallback(tcfData, success) {
    utils.logInfo('Received a response from CMP', tcfData);
    if (success) {
      if (tcfData.gdprApplies === false || tcfData.eventStatus === 'tcloaded' || tcfData.eventStatus === 'useractioncomplete') {
        cmpSuccess(tcfData, hookConfig);
      }
 ...
```
the second condition leaves it hanging, since eventStatus is `cmpuishown`. It doesn't look for the enforcement settings. Is this by design, or am I missing something?

## Steps to reproduce
1. Load the page with no TCFv2 consent in cookies/localStorage.
2. Prebid should wait for CMP timeout but still start auction, when `purpose: 'basicAds', enforcePurpose: false,` is in config.

## Test page
[https://playground.cpex.cz/headerbidding](https://playground.cpex.cz/headerbidding)

### Expected results
Auction to run

### Actual results
Auction not running

## Platform details
Prebid.js v5.2.0 (v4 behaves the same)
MacOS, Chrome

## Other information
None

gglas:
@bretg will have a look, first read is this is not desired behavior.

bretg:
[...]

PR #7769, opened 2021-11-29
Prebid Core: Consent management allow auction to continue when CMP doesn't respond

dgirardi:
## Type of change

- [x] Bugfix

## Description of change
﻿This reproduces the behavior of `allowAuctionWithoutConsent` when a TCFv2 CMP times out.
https://github.com/prebid/Prebid.js/issues/7156

**NOTE**: With this change, the auction will start with `consentData: {gdprApplies: true, consentData: undefined, consentString: undefined}`. There is a substantial number of modules that in my estimation will **not** work correctly with this configuration, see https://github.com/prebid/Prebid.js/issues/7156#issuecomment-978354248. I understand this to be an existing problem if one uses TCFv1 and `allowAuctionWithoutConsent`; the difference there is that one has to explicitly 'opt-in' to the broken behavior by setting that config flag; this would automatically start it after `consentManagement.gdpr.timeout` expires.

### DIFF
--- a/modules/consentManagement.js
+++ b/modules/consentManagement.js
@@ -371,7 +371,13 @@ function processCmpData(consentObject, hookConfig) {
  * General timeout callback when interacting with CMP takes too long.
  */
 function cmpTimedOut(hookConfig) {
-  cmpFailed('CMP workflow exceeded timeout threshold.', hookConfig);
+  if (cmpVersion === 2) {
+    logWarn(`No response from CMP, continuing auction...`)
+    storeConsentData(undefined);
+    exitModule(null, hookConfig)
+  } else {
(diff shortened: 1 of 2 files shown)


ChrisHuie (approved):

ChrisHuie:
@aleksatr can you review this pr when you get the chance 🙏

aleksatr (approved):

aleksatr:
@dgirardi @ChrisHuie changes LGTM, how do you guys want to proceed, merge this immediately or not, considering this:

```
 There is a substantial number of modules that in my estimation will not work correctly with this configuration, see #7156 (comment).
```

Issue #7663, opened 2021-11-03
Feature req: restore domainOverride to SharedId


pycnvr:
## Type of issue
Question

## Description

It looks like when PubCommonId was merged with SharedId in #7149, the combined SharedId module lost the domainOverride function.  That used to return the top level domain for setting cookies.  What was the reason for the removal?  Thanks.

## Steps to reproduce

Any page with prebid.js and sharedid set to cookie.  Check cookie domain in dev console.

jdwieland8282:
Merge is a generous way of describing what we did. We really just retired sharedId and renamed the PC module to sharedid. It kept its existing functionality. We’d be happy to take a look and add that functionality in.

pycnvr:
@jdwieland8282 Thanks!

bretg:
What needs to happen here is:
- look up the 4.x PubCommon module
- port the domainOverride feature and tests to the 6.x SharedID module

dgirardi:
closing this with the merge https://github.com/prebid/Prebid.js/pull/7773

PR #7773, opened 2021-11-29
PubCommonID module: use topmost domain for cookie storage

dgirardi:
## Type of change
- [x] Feature

## Description of change
﻿This reintroduces logic to have sharedIdSystem/pubCommonID use the TLD for cookie storage
https://github.com/prebid/Prebid.js/issues/7663

**NOTE**: changing the domain will likely not work properly for people that have that cookie already under a different domain. I am not sure the behavior is well-defined; I get duplicate cookies in Chrome. This would apply to people with a `pubcid` cookie that saw the update in https://github.com/prebid/Prebid.js/pull/7149 as well as this one.


smenzer:
LGTM

smenzer (approved):

PR #7776, opened 2021-11-30
Welect Bid Adapter: update prebid compliance and add adapter back to master

nduitz:
## Type of change

- [x] Re-Add bidder adapter

## Description of change
This re-adds the welectBidderAdapter.js and the according specs.
The adapter apparently got removed with the release of version 5.
412277f72b97cab72644cb4a009f58067a165899
I got notified about required changes regarding the release which I implemented in https://github.com/prebid/Prebid.js/pull/6966.
There wasn't any further notification and the PR got merged so I assumed the adapter wouldn't be removed.

If there are any additional requirements please let me know :)

All documentation regarding this adapter is still in place.

### DIFF
--- a/modules/welectBidAdapter.js
+++ b/modules/welectBidAdapter.js
@@ -1,0 +1,106 @@
+import * as utils from '../src/utils.js';
+import { registerBidder } from '../src/adapters/bidderFactory.js';
+
+const BIDDER_CODE = 'welect';
+const DEFAULT_DOMAIN = 'www.welect.de';
+
+export const spec = {
+  code: BIDDER_CODE,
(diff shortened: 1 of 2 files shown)


modules/welectBidAdapter.js
```diff
@@ -0,0 +1,106 @@
+import * as utils from '../src/utils.js';
```
ChrisHuie, line 1: Can you please update this to import just the needed functions? We recently removed all the 'utils.' roots from adapters and modules.

nduitz: Yes will do, thank you

ChrisHuie (approved):

Issue #7775, opened 2021-11-30
Adapters that may not handle GDPR-Applies correctly


bretg:
## Description
As uncovered in #7156 , there are are several bid adapters that may not be properly handling the `gdprApplies` flag.

Background: the Consent Management Platform (CMP) is responsible for knowing whether the current user is "in-scope" for GDPR, i.e. they reside in the European Economic Area (EEA). It signals this information to Prebid along with the consent string. Bid adapters must be able to handle all of these scenarios:

1) CMP provides gdprApplies:true and a valid consent string - normal GDPR processing
2) CMP provides gdprApplies:false and a valid consent string - bidder endpoints may choose to verify the user's GDPR scope or to trust the CMP.
3) CMP provides gdprApplies:true but no consent string - processing depends on vendor's Legitimate Interest claims and legal advice from the bidder's lawyers.
4) CMP provides gdprApplies:false and no consent string - bidder endpoints may choose to verify the user's GDPR scope or to trust the CMP.
5) CMP provides only a valid consent string - if the CMP declines to define the GDPR scope, bidder processing depends on whether the endpoint can detect the user's location and legal advice from their lawyers.
6) CMP provides neither value - if the CMP doesn't define the scope or the consent, bidder processing depends on whether the endpoint can detect the user's location and legal advice from their lawyers.

## Bid adapters that need to be reviewed

These bidders don't look for gdprApplies. Please confirm your implementation with your legal team. Prebid recommends passing the gdprApplies flag along with the consent string.
- [x] addefend - @addefend 
- [ ] apstream - @frstua 
- [x] glimpse - @samueldobbie , @tim-hm 
- [x] improveDigital - @jbartek25 , @agregorio-improve "there's no plan to add support for gdprApplies param in the current adapter generation as our server does IP lookups. The next generation adapter launching in Q1 will include gdprApplies"

These bidders currently only consider gdprApplies if a consent string is available. Please confirm your implementation with your legal team. Prebid recommends using the gdprApplies flag even if it's available even if there's no consent string.
- [ ] cleanmedianet - @sa1omon 
- [ ] engageya - @mikomgk 
- [x] gamoshi - @sa1omon 
- [x] stroeer - @PavlaKanova , @lukashavrlant 
- [x] bliink - @Kola-Kola 
- [ ] logan - @WlsLogan 
- [ ] mathildeads - @mathilde-ads 
- [x] mediasquare - @matthieularere-msq 
- [x] onetag - @onetag-dev 
[...]

PR #7779, opened 2021-11-30
Glimpse Bid Adapter: fix GDPR consent choice capture

samueldobbie:
## Type of change

- [x] Bugfix
- [ ] Feature
- [ ] New bidder adapter
- [ ] Build related changes
- [ ] CI related changes
- [ ] Other

## Description of change

We've updated the GDPR consent capture logic to ensure it correctly handles scenarios where the consent string is undefined, but gdprApplies is `true` (as shown in the example [here](https://github.com/prebid/Prebid.js/issues/7775#issuecomment-982861013)).

### DIFF
--- a/modules/glimpseBidAdapter.js
+++ b/modules/glimpseBidAdapter.js
@@ -138,11 +138,16 @@ function getReferer(bidderRequest) {
 function getGdprConsentChoice(bidderRequest) {
   const hasGdprConsent =
     hasValue(bidderRequest) &&
-    hasValue(bidderRequest.gdprConsent) &&
-    hasStringValue(bidderRequest.gdprConsent.consentString)
+    hasValue(bidderRequest.gdprConsent)
 
   if (hasGdprConsent) {
-    return bidderRequest.gdprConsent
+    const gdprConsent = bidderRequest.gdprConsent
+
+    return {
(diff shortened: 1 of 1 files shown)


ChrisHuie (approved):

modules/glimpseBidAdapter.js
```diff
@@ -138,11 +138,16 @@ function getReferer(bidderRequest) {
...
+    hasValue(bidderRequest.gdprConsent)
 
   if (hasGdprConsent) {
-    return bidderRequest.gdprConsent
+    const gdprConsent = bidderRequest.gdprConsent
+
+    return {
+      consentString: gdprConsent.consentString || '',
+      vendorData: gdprConsent.vendorData || {},
+      gdprApplies: gdprConsent.gdprApplies || true,
```
dgirardi, line 149: this is always true - might not be what you intended.

samueldobbie: It was added to capture the case where `gdprApplies` may be undefined.

samueldobbie: Ahh, thanks, it was just added to catch the case where `gdprApplies` is undefined. Opened a PR with the change (https://github.com/prebid/Prebid.js/pull/7783)

PR #7791, opened 2021-12-02
adquery - added adquery ID module

adquery:
## Type of change
- [x] Feature

## Description of change
adquery - added adquery ID module


patmmccann:
Can you split into two pr's? One for Id system, the other for bidder?

adquery:
OK I will split pr

PR #7794, opened 2021-12-02
Rubicon Bid Adapter: Pass along GPID + Minor refactoring

robertrmartinez:
## Type of change
- [X] Feature
- [X] Other

## Description of change

# GPID
Rubicon will now pass along the gpid value to the exchange

# Remove "remove leading slash form slots logic"
We were removing leading slashes form the pbadslot and dfp adslot before sending them, now we will not

# only send one of pbadslot or dfp adunit
Instead of always sending both if available, we will send one. Prefering pbasdlot

# Only pass adslot if adserver.name is gam
We pass along adserver.adslot into dfp_adunit_code ... so obviouslt this should only happen when adserver.name === gam.

### DIFF
--- a/modules/rubiconBidAdapter.js
+++ b/modules/rubiconBidAdapter.js
@@ -994,6 +994,8 @@ function applyFPD(bidRequest, mediaType, data) {
 
   let fpd = mergeDeep({}, config.getConfig('ortb2') || {}, BID_FPD);
   let impData = deepAccess(bidRequest.ortb2Imp, 'ext.data') || {};
+
+  const gpid = deepAccess(bidRequest, 'ortb2Imp.ext.gpid');
   const SEGTAX = {user: [4], site: [1, 2, 5, 6]};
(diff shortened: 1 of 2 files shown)


patmmccann (approved):

modules/rubiconBidAdapter.js
```diff
@@ -1047,12 +1039,30 @@ function applyFPD(bidRequest, mediaType, data) {
...
+
+    // add in gpid
+    if (gpid) {
+      data['p_gpid'] = gpid;
+    }
+
+    // only send one of pbadslot or dfp adunit code (prefer pbadslot)
+    if (data['tg_i.pbadslot']) {
+      delete data['tg_i.dfp_ad_unit_code'];
+    }
```
harpere, line 1057: this is only relevant to Banner?

harpere (commented):
looks good other than one question.

modules/rubiconBidAdapter.js
robertrmartinez: Yeah, the rubicon PBS adapter will do the handling when it sends to XAPI.

Which is nice so we can just dump stuff into PBS and let the server handle it rather than having a YUGE bid adapter!

harpere (approved):

PR #7798, opened 2021-12-03
Compass Adapter: add new adapter

CompassSSP:
- [x] Feature

```
var adUnits = [
    // Will return static test banner
    {
        code: 'adunit1',
        mediaTypes: {
            banner: {
                sizes: [ [300, 250], [320, 50] ],
            }
        },
        bids: [
            {
                bidder: 'compass',
                params: {
                    placementId: 'testBanner',
                }
            }
        ]
    },
    {
        code: 'addunit2',
        mediaTypes: {
            video: {
                playerSize: [ [640, 480] ],
                context: 'instream',
                minduration: 5,
                maxduration: 60,
            }
        },
        bids: [
            {
                bidder: 'compass',
                params: {
                    placementId: 'testVideo',
                }
            }
        ]
    },
    {
        code: 'addunit3',
        mediaTypes: {
            native: {
                title: {
                    required: true
                },
                body: {
                    required: true
                },
                icon: {
                    required: true,
                    size: [64, 64]
                }
            }
        },
        bids: [
            {
                bidder: 'compass',
                params: {
                    placementId: 'testNative',
                }
            }
        ]
    }
];
```

- contact email of the adapter’s maintainer - sa-support@brightcom.com
- A link to a PR on the docs repo at https://github.com/prebid/prebid.github.io/pull/3441

### DIFF
--- a/modules/compass.js
+++ b/modules/compass.js
@@ -1,0 +1,199 @@
+import { isFn, deepAccess, logMessage } from '../src/utils.js';
+import { registerBidder } from '../src/adapters/bidderFactory.js';
+import { BANNER, NATIVE, VIDEO } from '../src/mediaTypes.js';
+import { config } from '../src/config.js';
+
+const BIDDER_CODE = 'compass';
+const AD_URL = 'https://sa-lb.deliverimp.com/pbjs';
+const SYNC_URL = 'https://sa-cs.deliverimp.com';
+
+function isBidResponseValid(bid) {
+  if (!bid.requestId || !bid.cpm || !bid.creativeId || !bid.ttl || !bid.currency) {
+    return false;
+  }
+
+  switch (bid.mediaType) {
+    case BANNER:
+      return Boolean(bid.width && bid.height && bid.ad);
+    case VIDEO:
+      return Boolean(bid.vastUrl || bid.vastXml);
+    case NATIVE:
+      return Boolean(bid.native && bid.native.impressionTrackers && bid.native.impressionTrackers.length);
+    default:
+      return false;
+  }
+}
+
+function getPlacementReqData(bid) {
+  const { params, bidId, mediaTypes } = bid;
+  const schain = bid.schain || {};
+  const { placementId } = params;
+  const bidfloor = getBidFloor(bid);
+
+  const placement = {
+    placementId,
+    bidId,
+    schain,
+    bidfloor
+  };
+
+  if (mediaTypes && mediaTypes[BANNER]) {
+    placement.adFormat = BANNER;
+    placement.sizes = mediaTypes[BANNER].sizes;
+  } else if (mediaTypes && mediaTypes[VIDEO]) {
+    placement.adFormat = VIDEO;
+    placement.playerSize = mediaTypes[VIDEO].playerSize;
+    placement.minduration = mediaTypes[VIDEO].minduration;
+    placement.maxduration = mediaTypes[VIDEO].maxduration;
+    placement.mimes = mediaTypes[VIDEO].mimes;
+    placement.protocols = mediaTypes[VIDEO].protocols;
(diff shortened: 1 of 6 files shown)


CompassSSP:
@Rothalack please, check this pull request =)

Rothalack:
Hey @CompassSSP 

The first two things I know you need to do is rename your files: modules/compass.js should be modules/compassBidAdapter.js and modules/compass.md should be modules/compassBidAdapter.md

I'm trying to figure out why those issues are throwing in Codacy now. That's a new check that I haven't worked with yet. The issues don't really make sense to me, trying to cross reference other similar issues.

modules/compass.js
```diff
@@ -0,0 +1,199 @@
...
+  }
+
+  try {
+    const bidFloor = bid.getFloor({
+      currency: 'USD',
+      mediaType: '*',
+      size: '*',
+    });
+    return bidFloor.floor;
+  } catch (_) {
```
Rothalack, line 80: I think this needs to be catch (e). Then log the message instead of return 0. I think this will clear up one of the Codacy issues.

modules/compass.md
```diff
@@ -0,0 +1,80 @@
+# Overview
+
+```
+Module Name: Compass Bidder Adapter
+Module Type: Compass Bidder Adapter
+Maintainer: sa-support@brightcom.com
+```
+
+# Description
```
Rothalack, line 9: Add another # to this title.

## Description instead of # Description.

This should clear up another Codacy issue.

modules/compass.md
```diff
@@ -0,0 +1,80 @@
...
+
+Connects to Compass exchange for bids.
+
+Compass bid adapter supports Banner, Video (instream and outstream) and Native.
+
+# Test Parameters
```
Rothalack, line 15: Add another # to this title.

``## Test Parameters instead of # Test Parameters``

Codacy is very picky haha

Rothalack:
The other three Codacy issues still don't make sense to me.

compass.js line 90
    const { params, bidId, mediaTypes } = bid;
It says Code Style Unnecessary block. "Why is this an issue?
Since: PMD 5.0
An unnecessary Block is present. Such Blocks are often used in other languages to introduce a new variable scope. Blocks do not behave like this in ECMAScipt, and using them can be misleading. Considering removing this unnecessary Block."


compass.js line 154
    return {
      method: 'POST',
      url: AD_URL,
"Why is this an issue?
Since: PMD 5.0
An unnecessary Block is present. Such Blocks are often used in other languages to introduce a new variable scope. Blocks do not behave like this in ECMAScipt, and using them can be misleading. Considering removing this unnecessary Block."

Both of these are the same error as the try catch I left a comment on for line 80. I honestly don't know why these issues throw. I'll try to figure out why.


The last Codacy issue is from compass.md line 3.
``
```
Module Name: Compass Bidder Adapter
Module Type: Compass Bidder Adapter
Maintainer: sa-support@brightcom.com
```
``
"Why is this an issue?
Check fenced code block flags.

Options: Array.<string> or Object, optional.

Providing an array is as passing {flags: Array}.

The object can have an array of 'flags' which are allowed: other flags will not be allowed. An allowEmpty field (boolean, default: false) can be set to allow code blocks without language flags."

I think this can be solved by just removing the triple ` around that block. I don't know why Codacy cares about these, but apparently it does.

CompassSSP:
@Rothalack Thanks! We renamed the files, but did not change the Markdown markup. At the moment, it is clear that all tests have been passed.

Rothalack:
Yea, it looks like we removed the Codacy check. I hadn't heard anything about it before and I guess we were just testing it. 

Tested this adapter and it looks to be working, this should be good to go.

Rothalack (approved):

Issue #7020, opened 2021-06-14
Pbs bid adapter: video hb_size not returned properly


r-schweitzer:
## Type of issue
Bug

## Description
There are sizes that returning on video as "**undefined**x**undefined**" on server to server and potentially client side as well. Some bidders choose to send back the playerWidth and playerHeight since sizes are not always returned in the bidresponse. Seems that may Prebid should determine to either filter hb_size when a size is not returned or return the playerWidth and playerHeight as a standard.
## Steps to reproduce
Reload page enough time with prebid server and you still see hb_size is not defined properly. 

### Expected results
I would expect either no hb_size or the size of the actual video.

### Actual results
undefined x undefined size or 1x1. Sizes may vary depending on the the adapter or prebid server etc.

## Platform details
All version of Prebid.

gglas:
@mike-chowla we discussed in issue review today -- we think the insertion of undefined is a bug that needs fixing (and it should just be left blank, maybe console log that size is undefined), but wanted to make sure that fix corresponds with the needs in video.

bretg:
@r-schweitzer - a test page would be good. Do you see the 'undefinedxundefined' in the PBS response JSON or is from PBJS?

r-schweitzer:
@bretg Seems to be related to PBS. I can recreate it on this page https://cafedelites.com/cheesy-scrambled-eggs-in-avocado/. The PBS response doesn't come back with hb_size at all

bretg:
@r-schweitzer - spent some time with this and can't make it happen. My thought is that PBS targeting variables aren't used by PBJS. They're only used by mobile apps and AMP. So I don't think PBS is the problem here.

There might be some robustness that we can add to the pbsBidAdapter in https://github.com/prebid/Prebid.js/blob/f50b7ec513fd872c7c9410fdc5bbcee4f4128f54/modules/prebidServerBidAdapter/index.js#L924

```
            let sizes = bidRequest.sizes && bidRequest.sizes[0];
            bidObject.playerWidth = sizes[0];
            bidObject.playerHeight = sizes[1];
```
We should check `sizes` before accessing them as an array. @ChrisHuie - please make this update?

dgirardi:
@ChrisHuie the PBS adapter now sets `playerWidth` / `playerHeight` to the size that was in the request. Does that solve this issue?

[...]

PR #7799, opened 2021-12-03
Pbs Bid Adapter: check that video sizes is an array prior to passing

ChrisHuie:
This is in reference to issue #7020

### DIFF
--- a/modules/prebidServerBidAdapter/index.js
+++ b/modules/prebidServerBidAdapter/index.js
@@ -921,13 +921,13 @@ const OPEN_RTB_PROTOCOL = {
 
           if (deepAccess(bid, 'ext.prebid.type') === VIDEO) {
             bidObject.mediaType = VIDEO;
-            
+
             let sizes = bidRequest.sizes && bidRequest.sizes[0];
             if (Array.isArray(sizes)) {
               bidObject.playerWidth = sizes[0];
               bidObject.playerHeight = sizes[1];
             }
-            
+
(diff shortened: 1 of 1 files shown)


modules/prebidServerBidAdapter/index.js
```diff
@@ -921,9 +921,12 @@ const OPEN_RTB_PROTOCOL = {
...
           if (deepAccess(bid, 'ext.prebid.type') === VIDEO) {
             bidObject.mediaType = VIDEO;
+
             let sizes = bidRequest.sizes && bidRequest.sizes[0];
-            bidObject.playerWidth = sizes[0];
-            bidObject.playerHeight = sizes[1];
+            if (Array.isArray(sizes)) {
+              bidObject.playerWidth = sizes[0];
+              bidObject.playerHeight = sizes[1];
+            }
```
patmmccann, line 929: else logWarn?

ncolletti:
I spent some time tracing this back and believe the problem is the result of bidders not returning height and widths in bid responses from video requests. It can't hurt to include this additional check for playerSize in prebid server adapter, however I'm not sure it will resolve the linked issue.

The Bid class in bidFactory.js uses a `getSize` function that does not check if width or height are defined before accessing the properties. This is called and set to the bid object when auction.js gets to targeting in the `getKeyValueTargetingPairs` and further in the `setKeys` function on line 740. 

Possibly a validation can occur in `tryAddVideoBid` where the getSize will be applied and default to the playerSize object height and width values, if the same properties are undefined from the bid response object.

Issue #4461, opened 2019-11-13
Improve ability to test bidder responses


pm-harshad-mane:
## Type of issue
Feature Request

## Description
While debugging on a live inventory, ad-ops people need to request test bids from bidders.
We have a to modify bid responses but it needs changes in the config
http://prebid.org/dev-docs/troubleshooting-tips.html#modify-bid-responses-for-testing
DistrictM has its own way, it looks for ``` dmTest=true ``` in page URL, but bidder specific test-query-param is not a scalable solution.

## Proposal
All bidders read a common query param ``` pbjs_testbids=true ``` and if present in page URL then each bidder will send respective test-bids flag to their server.

pm-harshad-mane:
Rather than each bidder reading the same param, prebid-core can pass it in bidRequet object to bidders.

pm-harshad-mane:
we can also support  ``` pbjs.setConfig({ testbids: true }); ```

jaiminpanchal27:
@bretg @mkendall07 Any thoughts on this new feature request

pm-harshad-mane:
Hello @bretg, @mkendall07 
Could you please share your thoughts on this new feature request?

bretg:
What is the bidder expectation here? Just avoid charging for the bid? Extra debug info? Fake creative? Really whatever the expectation, it's pretty impractical -- that would be a very long term goal.

You need to identify the actual use cases... what does the pub want to test:

- header bidding ads high enough to win in the ad server?
- ad server line items?
- PBJS config like bidAdjustments?
- a specific type of creative?

I don't think trying to herd 200+ bidder-cats into providing fake bid+creatives is going to happen anytime soon. And the requirements are too vague -- does the pub want a bid of a specific value to trigger a particular line item? Do they want to see bidderA win for once?

I think a much better approach would be to deepen the feature we already have documented at http://prebid.org/troubleshooting/troubleshooting-guide.html#modify-bid-responses-for-testing

One could imagine a Chrome extension that would help make it easier for AdOps people to identify an adunit and force things to happen... all within PBJS.

pm-harshad-mane:
The expectation from the bidder is to serve a fake creative.
This can help test the end-to-end setup, not all bidders respond with bids always.
I agree that it will be difficult to have all bidders support it.
The existing approach is useful only in a test environment where one has access to the PBJS config to modify bid responses.

bretg:
> The existing approach is useful only in a test environment

[...]

PR #7801, opened 2021-12-03
Prebid core & PBS Adapter: debugging tools to intercept bid requests and mock their response

dgirardi:
## Type of change

- [x] Feature

## Description of change
This adds the ability to 'intercept' bid requests with a mock response. An intercepted bid does not go through the normal `buildRequests` -> network -> `interpretResponse` flow; a dummy response is instead directly returned to the auction as if it originated from `interpretResponse`.

This is mostly transparent to other parts of the stack (validation, events, etc), with the exception of user syncs - their interface requires the bidder's HTTP response, so they won't run if all bids in the request are intercepted (and thus no network activity happens).

## Other information

Addresses https://github.com/prebid/Prebid.js/issues/4461
Docs PR: https://github.com/prebid/prebid.github.io/pull/3575

```javascript
pbjs.setConfig({
  debugging: {
    enabled: true,
    intercept: [
     // intercept rules are evaluated in order, first to match wins
      {
        when: {
          //  intercept all  bids that have adUnitCode === 'test-div'
          adUnitCode: 'test-div',
          //  (any other property of the bid object can be tested here)
        },
        then: {
          // mock their response with sane defaults and `cpm: 10`
          cpm: 10
          // anything added here will appear in the response
        }
      },
      {
        // fancier match & replace options
        when: {
          adUnitCode: /div/,
          mediaTypes: (mt) => mt.banner != null,
        },
        then: {
          width: (bid) => bid.mediaTypes.banner.sizes[0][0],
          height: (bid) => bid.mediaTypes.banner.sizes[0][1],
        }
      },

    ]
  }
});

```


ChrisHuie:
@Fawke @robertrmartinez can yu guys check out this pr when you get the chance? 🙏

Fawke (commented):
Hi @dgirardi,

Thanks for adding this feature to the Debugging module. Your code changes mostly look good to me and  I've also tested the feature with different combinations of `ruleDefs` and it seems to be doing its job.

Do you think the keyword `intercept` in the config definition should be changed to `intercepts`?

The only major concern that I have is that as a side effect of this change, some existing Analytics Adapters which rely on Prebid Events for reporting will start misreporting. 
For example,  an adapter listening for the `bidResponse` event and in the case when debugging is turned on, it will report false/mock values to its server.

A naive solution could be that a AnalyticAdapter checks if debugging is turned on/off and depending on that, send the values to the Analytics server. I don't know if there is some other way which doesn't require the Analytics adapters to make this change.

FilipStamenkovic (commented):
I don't think this is the best approach to solve this issue.  
Your solution requires adding `debugging.js` for each bid adapter, so debugging existing bidders won't work without additional code.  
Also, this adds additional code to each bidder which would increase final prebid.js bundle on publishers page.  

I think optional solution would be something that:
- won't inflate pbjs size when using in production
- can be used at once with all bidders

Can we add some kind of module that replaces existing functions in prebid core in order to replace bid requests and mock their responses? Similar to what we have in our tests.

src/debugging.js
```diff
@@ -22,21 +31,29 @@ function addHooks(overrides) {
...
   removeHooks();
   addHooks(overrides);
   logMessage(`bidder overrides enabled${fromSession ? ' from session' : ''}`);
 }
 
 export function disableOverrides() {
+  bidInterceptor.updateConfig({});
   removeHooks();
   logMessage('bidder overrides disabled');
 }
```
Fawke, line 59: Does it make sense to clear the sessionStorage for `pbjs:debugging` here? While testing it locally, even though I had set debugging `enabled: false`, I had to manually close the tab to make this take effect.

dgirardi:
@Fawke,

> The only major concern that I have is that as a side effect of this change, some existing Analytics Adapters which rely on Prebid Events for reporting will start misreporting.
> For example, an adapter listening for the `bidResponse` event and in the case when debugging is turned on, it will report false/mock values to its server.

I view that as a feature not a bug - a valid use case for this would be testing analytics adapters. Is there a reason analytics should be different from the rest? The mock values would only be reported when a publisher is explicitly doing testing.

dgirardi:
@FilipStamenkovic,

> Your solution requires adding `debugging.js` for each bid adapter, so debugging existing bidders won't work without additional code.

Why? I have a special case for the PBS adapter (because that is itself a special case, not really an adapter) but nothing needs to be changed for anyone else (unless I am missing something).

The decision to make this core instead of a separate debugging module was made offline with @bretg, I believe to support the browser extension tooling. I am open to break this out into an optional module if the community agrees on that.

bretg:
I think basic debugging support is worth having in PBJS-core.

The question of course is "what comprises basic debugging support"?

I'm often called on to debug publisher pages and find it difficult and time-consuming to attempt to build a custom debug-enabled PBJS package and fake the browser into loading it. 

So I think that a few hundred bytes of built-in debug support is worthwhile. But sounds like the community needs to have a debate about what specific debug features are worth the bytes and which need to be pushed to an optional module.

Fawke:
> @Fawke,
> 
> > The only major concern that I have is that as a side effect of this change, some existing Analytics Adapters which rely on Prebid Events for reporting will start misreporting.
> > For example, an adapter listening for the `bidResponse` event and in the case when debugging is turned on, it will report false/mock values to its server.
> 
> I view that as a feature not a bug - a valid use case for this would be testing analytics adapters. Is there a reason analytics should be different from the rest? The mock values would only be reported when a publisher is explicitly doing testing.

I never said it is a bug but is a side effect of this change. I agree this would be helpful in testing Analytics adapters but we would need a way to differentiate bids originating from when debugging is turned on vs when it is off to avoid any discrepancies in reporting. @bretg had raised a similar point in this [comment](https://github.com/prebid/Prebid.js/issues/4461#issuecomment-564658039)

FilipStamenkovic:
@dgirardi 
> Why? I have a special case for the PBS adapter (because that is itself a special case, not really an adapter) but nothing needs to be changed for anyone else (unless I am missing something).

My bad, I didn't test this properly. I just tested it with some random bidder and it works.

> Sounds like the community needs to have a debate about what specific debug features are worth the bytes and which need to be pushed to an optional module.

@bretg With these changes prebid-core.js is 9 KB heavier than on master branch, I'm not sure if this is a lot or not.

dgirardi:
@Fawke  @bretg I need help understanding what exactly needs to be protected from "fake" bids; it seems odd that we'd care about polluting analytics but not for example the "bid won" handlers. I am not familiar with publisher workflows but I'd imagine they'd not be too interested in analytics for their own development activity - that is going to be an outlier regardless of whether the data stream is genuine or not. How does that work?

dgirardi:
This was discussed offline today and we agreed to make it an optional module and have a strategy (in a separate PR) for dynamically loading it into production environments that may not have it.

dgirardi:
Updated to have new debugging logic in an optional module. I have kept existing functionality (in `src/debugging.js`) mostly untouched except for some coordination to share `sessionStorage` logic. Eventually - when we have a way to load it asynchronously - I think it should all be moved into the debugging module, but I didn't want to remove it from core until that's ready.

bretg:
> what exactly needs to be protected from "fake" bids; it seems odd that we'd care about polluting analytics

How about we add a "debug bid" onto any bid response object generated this way. Then analytics adapters can weed them out if they'd like

src/debugging.js
dgirardi: I don't know if it has changed after my latest updates, but both session storage and debug hooks should be cleared with `setConfig({debugging: {enabled: false}})`. Do you still see this problem?

Fawke: Seems to be working now. Thanks.

Fawke (approved):

Fawke:
> How about we add a "debug bid" flag onto any bid response object generated this way. Then analytics adapters can weed them out if they'd like

Just one thing to note, maybe I'm wrong, but I don't think this approach would catch/stop analytics adapters listening for events like, `auctionInit`, `bidRequested` and all those events which get triggered before we'd get a `bidResponse` event.

msm0504 (approved):

PR #7813, opened 2021-12-06
TheMediaGrid: RTD module support with mergeDeep of user.data

TheMediaGrid:
## Type of change

- [x] Feature

## Description of change

Added support for Ftrack RTD module

- grid-tech@themediagrid.com


patmmccann:
This appears to add support for the entire user data object. Should it be relabeled?

patmmccann (changes requested):
this appears to just not work unless the only thing writing to this location for the pub is this; also ftrack probably cannot write here, we're going to move it to eids most likely

TheMediaGrid:
@patmmccann please take a look, did I understand your proposal correctly?

patmmccann:
@TheMediaGrid I thought this was waiting on the ftrack merge, my mistake, i see this is for all RTD modules support now.

patmmccann (approved):
thanks very much!

PR #7829, opened 2021-12-08
Revert "adxcgBidAdapter: updated backend protocol for 5.2x-legacy"

ChrisHuie:
This seems to be failing on the legacy branch. Reverting to review prebid/Prebid.js#7786. Adding Pat for visibility

### DIFF
--- a/modules/adxcgBidAdapter.js
+++ b/modules/adxcgBidAdapter.js
@@ -1,362 +1,410 @@
-// jshint esversion: 6, es3: false, node: true
-'use strict';
+import { logWarn, isStr, deepAccess, inIframe, checkCookieSupport, timestamp, getBidIdParameter, parseSizesInput, buildUrl, logMessage, isArray, deepSetValue, isPlainObject, triggerPixel, replaceAuctionPrice, isFn } from '../src/utils.js';
+import {config} from '../src/config.js'
+import {registerBidder} from '../src/adapters/bidderFactory.js'
+import {BANNER, NATIVE, VIDEO} from '../src/mediaTypes.js'
+import includes from 'core-js-pure/features/array/includes.js'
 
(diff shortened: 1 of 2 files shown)


adxcgcom:
Solved the problem. It was browser (IE11) specific and showed only in browserstack tests., but not in local (gulp test that uses local chrome). 

With gulp test --browserstack this failed in IE 11 only. The problem was in browserstack IE 11 that request.site.domain is empty "" , while in gulp test it's "localhost".   All other browserstack enabled browsers work without problem.

```
    it('should send info about the site', function () {
      config.setConfig({
        site: {
          id: '123123',
          publisher: {
            domain: 'publisher.domain.com'
          }
        },
        ortb2: {
          site: {
            publisher: {
              id: 4441,
              name: 'publishers name'
            }
          }
        }
      });
      let validBidRequests = [{
        bidId: 'bidId',
        params: {adzoneid: '1000'}
      }];
      let refererInfo = {referer: 'page'};
      let request = JSON.parse(spec.buildRequests(validBidRequests, {refererInfo}).data);

      let expected = {
        id: '123123',
        publisher: {
          domain: 'publisher.domain.com',
          id: 4441,
          name: 'publishers name'
        },
        page: 'page',
        domain: 'localhost'
      };

      assert.deepEqual(request.site.id, expected.id);
      assert.deepEqual(request.site.publisher, expected.publisher);
      assert.deepEqual(request.site.page, expected.page);
      // assert.deepEqual(request.site.domain, expected.site.domain); // disabled for browser specific problem (IE11)
      // in browserstack IE 11 we get empty "" request.site.domain, while in gulp test we get "localhost"
      // all other browserstack enabled browsers work without problem
    });
```

Should we leave comments for clarity or remove them in new PR?

adxcgcom:
https://github.com/prebid/Prebid.js/pull/7844

PR #7834, opened 2021-12-08
Ftrack ID from  Flashtalking by Mediaocean RTD submodule: initial release

ftxmoJason:
## Type of change

- [ ] Bugfix
- [x] Feature
- [ ] New bidder adapter   
- [ ] Build related changes
- [ ] CI related changes
- [ ] Other

## Description of change
Adding a new Real Time Data submodule for ftrack

```
{
  bidder: '<bidder name>',
  params: {
    // ...
  }
}
```


patmmccann:
Appears this should be an id submodule?

modules/ftrackRtdProvider.js
```diff
@@ -0,0 +1,691 @@
...
+
+    FtBh.prototype = {
+      get: function get() {
+        var ua = navigator.userAgent.toLowerCase();
+        var keys = [];
+        var navLang = navigator.language || navigator.browserLanguage;
+        var navLangArr = navLang.split('-');
+
+        if (typeof navLangArr[0] == 'undefined') {
+          // navLang = navLang;
```
Rothalack, line 164: Is this intended? You could do

```
if (typeof navLangArr[0] !== 'undefined') {
  navLang = navLangArr[0];
}
```

Rothalack:
The other thing I have noticed in testing is according to ftrackRtdProvider.md you expect the POST payload of https://grid.bidswitch.net/hbjson to be the user object with data, name, segment, etc. When I set this up in testing, I only got a user object with an ID. Not sure if that's a mismatch of documentation to intended behavior or if I did something incorrect in testing. This is the object I get. Let me know if that's correct. Otherwise, I think this is good to go.
{
    "id": "18619f5c463f47",
    "site": {
        "page": "http%3A%2F%2Flocalhost%3A9999%2FintegrationExamples%2Fgpt%2Fhello_world.html%3Fpbjs_debug%3Dtrue"
    },
    "tmax": 1000,
    "source": {
        "tid": "7cd45a25-91d1-46f0-b669-7bae5ad024b3",
        "ext": {
            "wrapper": "Prebid_js",
            "wrapper_version": "6.4.0-pre"
        }
    },
    "imp": [
        {
            "id": "2a8cf2c7acae99",
            "tagid": "12341234",
            "ext": {
                "divid": "div-gpt-ad-1460505748561-0"
            },
            "banner": {
                "w": 300,
                "h": 250,
                "format": [
                    {
                        "w": 300,
                        "h": 250
                    },
                    {
                        "w": 300,
                        "h": 600
                    }
                ]
            }
        }
    ],
    "user": {
        "id": "5901ec58c7804216bc5ee8c2b0ea92db"
    }
}

ftxmoJason:
@Rothalack I can't reply to your comment here for some reason, but the `user` property would not have the ftrack ID until https://github.com/prebid/Prebid.js/pull/7813 is merged in. For now, if you lookin in the localStorage, you will see the ID stored there. Please see: DevTools > Application > Local Storage > Your WebPage: You will see `ftrack-rtd` and `ftrack-rtd_exp` in the storage. If those exists than this module is working as expected.

modules/ftrackRtdProvider.js
ftxmoJason: Thanks @Rothalack , I can make this change.

ftxmoJason: Ok, this has been committed.

patmmccann:
@ftxmoJason I am planning to close this PR unless you respond to my question above. Thank you.

ftxmoJason:
@patmmccann Thanks for the question Pat. Sorry for the slow response to your q, I wanted to make sure the folks who assigned this project to me answered your question for me:

The FTrack solution is not limited to associating a single identifier per user, thus we consider it data associated with a user and not necessarily a unique identifier. Our plan is to extend identifiers to several segments; hence the RTD module is better suited for our scenarios.

patmmccann:
we'll have to discuss in committee, this is a bit of an edge case

@gglas fyi

ftxmoJason:
Ok, thanks @patmmccann . If you need any further clarification, let us know!

ftxmoJason:
Hello @patmmccann 
We have update the code to add an extension to the Ftrack rtd module in order to include the more segment data.

patmmccann:
Thanks! @ftxmoJason please note the lgtm error

modules/ftrackRtdProvider.js
```diff
@@ -0,0 +1,713 @@
...
+        if (!obj) {
+          try {
+            obj = new window.ActiveXObject('PDF.PdfCtrl');
+          } catch (e) {
+            obj = null;
+            return null;
+          }
+        }
+
+        if (obj !== null) {
```
patmmccann, line 414: this is always true, throwing error on lgtm.com

ftxmoJason: Thanks @patmmccann .  PR update incoming.

patmmccann:
@ftxmoJason I still; I think it is a mistake for an id provider to not set their id on the eids object, as it will severely limit ssp uptake. It currently seems like flashtalking is only interested in iponweb reading this value. It also appears to be at least one rules violation https://docs.prebid.org/dev-docs/module-rules.html: "RTD modules must not supply privacy-sensitive user information (including IDs) into the auction. That functionality is reserved for User ID modules." and perhaps also rules around fairness to bidders as only one bidder expects an id in this location. I think this module can move forward with the segments but that you should pass the id on a second module and encourage your publisher partners to install both.

PR #7838, opened 2021-12-09
IX Bid Adapter: update GPID Support

lksharma:
## Type of change

- [ ] Bugfix
- [x ] Feature
- [ ] New bidder adapter   
- [ ] Build related changes
- [ ] CI related changes
- [ ] Other

## Description of change
This change is to support sending imp[].ext.gpid in our Cygnus call.

### DIFF
--- a/modules/ixBidAdapter.js
+++ b/modules/ixBidAdapter.js
@@ -680,6 +680,7 @@ function buildRequest(validBidRequests, bidderRequest, impressions, version) {
 
     if (impressionObjects.length && BANNER in impressionObjects[0]) {
       const { id, banner: { topframe }, ext } = impressionObjects[0];
+      const gpid = impressions[transactionIds[adUnitIndex]].gpid;
       const _bannerImpression = {
         id,
         banner: {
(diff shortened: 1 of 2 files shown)


modules/ixBidAdapter.js
```diff
@@ -688,10 +689,10 @@ function buildRequest(validBidRequests, bidderRequest, impressions, version) {
...
       }
 
-      if (ext.dfp_ad_unit_code) {
-        _bannerImpression.ext = {
-          dfp_ad_unit_code: ext.dfp_ad_unit_code
-        }
+      if (ext.dfp_ad_unit_code || gpid) {
+        _bannerImpression.ext = {};
+        _bannerImpression.ext.dfp_ad_unit_code = ext.dfp_ad_unit_code;
+        _bannerImpression.ext.gpid = gpid;
```
patmmccann, line 695: if gpid is null but dfp_ad_unit_code is not, would this throw an exception?

lksharma: Both `gpid` and `dfp_ad_unit_code` are `undefined` when they don't exist. If either `gpid` or `dfp_ad_unit_code` are `undefined` they will be removed from the `r` object once we run `JSON.stringify`.

patmmccann (approved):

PR #7847, opened 2021-12-10
Prebid Core: switch native assets to ortb2 format

### LINKED ISSUE #7830: Prebid Native: switch to openRTB for native
## Type of issue
Feature request.

> This issue is based on the following [comment](https://github.com/prebid/Prebid.js/issues/6654#issuecomment-970717544) by @bretg.  

## Description
The goal is to replace PBJS's "proprietary" native configuration with the openRTB standard.  
Native configuration should support openRTB native specification, version 1.2. The full specs can be found [here](https://www.iab.com/wp-content/uploads/2018/03/OpenRTB-Native-Ads-Specification-Final-1.2.pdf).  

Currently, in PBJS, the user would configure native ad units like:

``` javascript
{
    code: 'adUnitCode',
    mediaTypes: {
        native: {
            rendererUrl: "https://domain.com/renderer.js",
            sendTargetingKeys: false,
            adTemplate: "<html><head>...</head></html>",  // not needed in case where rendererUrl exists
            title: {
                required: true,
                len: 800
            },
            image: {
                required: true,
                sizes: [3000, 2250]
            },
            ...
        }
    },
}
```

Problems with provided example are:

- there isn't a clear distinction between asset configuration, and PBJS native configuration (`sponsoredBy` is an asset, but `sendTargetingKeys` property is configuration so that PBJS knows whether to set GPT targeting or not)
- it's not possible to request the same type of asset twice; assets are configured as key-value pairs, it's impossible to configure PBJS to request two `image`s, for example.
- many current bidders are already sending openRTB requests to their back end; there is a lot of duplicated code in prebid that translates prebid "proprietary" native implementation to openRTB.
- no support for users to configure event trackers

## Proposed update

The proposed solution is to go with plain openRTB instead of the current "proprietary" configuration. So, the example from above would look like this:

``` javascript
{
    code: 'adUnitCode',
    mediaTypes: {
        native: {
            rendererUrl: "https://domain.com/renderer.js",
            sendTargetingKeys: false,
            adTemplate: "<html><head>...</head></html>",  // not needed in case where rendererUrl exists
            ortb2: {
                ver: "1.2",
                assets: [
                    {
                        id: 1,
                        required: 1,
                        title: {
                            len: 140,
                        },
                    },{
                        id: 2,
                        required: 1,
                        img: {
                            h: 2500,
                            w: 3000,
                            type: 3,
                        },
                    },
                    ...
                ],
                eventtrackers: [...],
            },
        },
    }
}
```

Note that all `assets` and `eventtrakcers` (everything related to openRTB) are under `mediaTypes.native.ortb2` property. While, PBJS specific configuration, such as `rendererUrl`, `adTemplate`, `sendTargetingKeys` are not under `ortb2` property.

## Potential problems

All currently supported macros for string substitution (during render time or targeting set) must be replaced.  
Current macros:

``` javascript
"NATIVE_KEYS": {
  "title": "hb_native_title",
  "body": "hb_native_body",
  "body2": "hb_native_body2",
  "privacyLink": "hb_native_privacy",
  "privacyIcon": "hb_native_privicon",
  "sponsoredBy": "hb_native_brand",
  "image": "hb_native_image",
  "icon": "hb_native_icon",
  "clickUrl": "hb_native_linkurl",
  "displayUrl": "hb_native_displayurl",
  "cta": "hb_native_cta",
  "rating": "hb_native_rating",
  "address": "hb_native_address",
  "downloads": "hb_native_downloads",
  "likes": "hb_native_likes",
  "phone": "hb_native_phone",
  "price": "hb_native_price",
  "salePrice": "hb_native_saleprice",
  "rendererUrl": "hb_renderer_url",
  "adTemplate": "hb_adTemplate"
}
```

This needs to be replaced if we want PBJS to use openRTB for native. I suggest using `asset.id` as a macro instead of the name. So, macro would look: `hb_native_asset_1` ... `hb_native_asset_N`. Then, during ad render time, renderer (or PUC) would take assets from native bid response, and based on its `id`, it would do string replacements.

## Additional changes

Adding this change to prebid native would require changes in:
- prebid universal creative (macros for string substitution)
- each bidder that supports `native` would need to update their code

FilipStamenkovic:
## Type of change

- [x] Feature

## Description of change
Change from prebid 'standard' of defining native assets to [openRTB 1.2 standard](https://www.iab.com/wp-content/uploads/2018/03/OpenRTB-Native-Ads-Specification-Final-1.2.pdf).  
Instead of `sponsoredBy`, `image`, `icon`, we would have openRTB equivalents:
- `{ data: { type: 1 }}`
- `{ img: { type: 3 }}`
- `{ img: { type: 1 }}`

This will help standardize prebid native. Also, a lot of bidders are sending ortb2 requests on their back end, so at the moment, they are converting bid request from prebid to openRTB standard and later they are converting bid responses from openRTB to prebid standard. With this approach a lot of duplicated code between the bidders can be removed.

## Backwards compatibility
When this PR is merged, we will still have backwards compatibility. Which mean, nothing will break for the existing native bidders. That will give us some time until all bidders adjust their code

## Other information
This PR resolves #7830 issue.


FilipStamenkovic:
I also created an example how this will look like on the test page. I used `pubwise` bidder, because that's the only bidder I found that is always returning to me test native bid.  

Code how that would look like can be found on my [local repo](https://github.com/FilipStamenkovic/Prebid.js/pull/1)

src/native.js
```diff
@@ -92,6 +198,23 @@ export function nativeBidIsValid(bid, bidRequests) {
...
+    logError(`native response doesn't have 'link' property. Ortb response: `, bidORTB);
+    return false;
+  }
+
+  let requiredAssetIds = bidRequestORTB.assets.filter(asset => asset.required === 1).map(a => a.id);
+  let returnedAssetIds = bidORTB.assets.map(asset => asset.id);
+
+  const match = requiredAssetIds.every(assetId => includes(returnedAssetIds, assetId));
+  if (!match) {
+    logError(`didn't receive a bit with all required assets. Required ids: ${requiredAssetIds}, but received ids in response: ${returnedAssetIds}`);
```
Fawke, line 212: Typo: `bit` should be `bid`.

Fawke (changes requested):
Hi,

The changes look fine at a first glance, but I do not see any docs PR associated with it. If Native now supports OpenRTB params, then I think we should document it with proper examples.

Also, I do see an associated PR for Prebid Universal Creative. Does that take of the rendering if new ortb fields are used?

src/native.js
FilipStamenkovic: I fixed this

FilipStamenkovic:
@Fawke 

Thanks for the review.  
I didn't open any docs PR, my plan was to open first changes to the code, so that reviewers can take a look at proposed approach how to switch to openRTB for native ads. If everything is ok I will also open docs PR.

bretg:
@FilipStamenkovic - we're going to need an update to the prebidServerBidAdapter. Currently it's mapping the prebid-specific attributes to ortb2. It should check for the existence of this new format first and merge them in. Thanks.

FilipStamenkovic:
@bretg I was thinking to do pbs adapter changes in separate PR not in this one. But, if you think it's ok to add pbs changes in this PR I'll do it then?

bretg:
I'd say go for it @FilipStamenkovic

FilipStamenkovic:
@bretg I included pbs changes in the PR. I didn't implement 'merge strategy' between old way and ortb format.  
I decided to go with approach: if `mediaTypes.native.ortb` is present use it, if not fallback to the current approach.

test/spec/unit/pbjs_api_spec.js
```diff
@@ -2128,14 +2128,47 @@ describe('Unit: Prebid Module', function () {
...
         const nativeRequest = spyArgs.args[1][0].bids[0].nativeParams;
-        expect(nativeRequest).to.deep.equal({
-          image: {required: true},
-          title: {required: true},
-          sponsoredBy: {required: true},
-          clickUrl: {required: true},
-          body: {required: false},
-          icon: {required: false},
-        });
+        expect(nativeRequest.ortb.assets).to.deep.equal([
```
Rothalack, line 2131: Wouldn't we want to have a test case of old native request as well? It should be able to accept either, if ortb is set go with that if not use old, right? Maybe it's too much to ask or it is already covered in another test.

FilipStamenkovic: It's already covered in another test.  
Basically all existing native tests rely on 'old native way'.  
But, if you feel like this PR needs more tests, I'll add them?

Rothalack: If it's covered in another test, then that works!

Rothalack (approved):

Rothalack:
I think this just needs your approval for the Docs PR change request @Fawke

src/secureCreatives.js
```diff
@@ -63,6 +63,11 @@ export function receiveMessage(ev) {
       } else if (data.action === 'allAssetRequest') {
         const message = getAllAssetsMessage(data, adObject);
         ev.source.postMessage(JSON.stringify(message), ev.origin);
+        // if there is an ortb object inside native, puc won't send postMessage to trigger impression tracker, in that case marking bid as winning
```
dgirardi, line 66: I think it'd be better to move this to the top, if any message with an `adId` comes in, it means that `adId` has won the auction, no? you'll need some logic to avoid duplicating the event but IMO that's better than two different ways to mark the bid won (or three if you count the non-native case).

dgirardi: I mean something like - 

```
const BID_WONS = new WeakMap();

function receiveMessage(ev) {
 // ...
 if !BID_WONS.has(adObject) {
    BID_WONS.set(adObject, true);
    // ... trigger event
 }
 // message handling
}

Comment (FilipStamenkovic) [reply src/secureCreatives.js]: updated

**Comment by ChrisHuie** [comment]
Docs PR -> https://github.com/prebid/prebid.github.io/pull/3497

**Review by osazos — APPROVED** [review]
It is ok for me for the code review.
I added a [comment in the doc PR](https://github.com/prebid/prebid.github.io/pull/3497#issuecomment-1026682789).

**Comment by bretg** [comment]
I hope I'm wrong here, but I think this PR is just the first step in a long journey. I believe that all 74 bid adapters that support native need to be updated to support this new format, right? i.e. instead of looking in mediatypes.native.title, they needs to look in mediatypes.native.**ortb2**.title, right?

If that's the case, publishers can't change their pages until all the adapters they use support the new syntax.

**Comment by FilipStamenkovic** [comment]
Yeah, this is just a first step, adapters would need to update their code.  

It would not be `mediaTypes.native.ortb2.assets.title`.  
It's `ortb` not `ortb2` (because this PR is based on openRTB 1.2 Native specs, there is a link in the description). Also, `assets` contains array of objects, so to get `title`s (in theory one native ad can have multiple titles) you would need to do something like:

``` javascript
let titles = mediaTypes.native.ortb.assets.filter(asset => asset.hasOwnProperty('title'));
```

bretg:
Ok, discussed in today's PBJS meeting. We propose handling this native change in a way similar to how we handled the first party data change last year:

1) Add some transition code (e.g. toOrtbNative) that maps old-style AdUnit native definitions to new-style.
2) Create a transition function (e.g. fromOrtbNative) that bid adapters can use to translate the new-style native definition back to the old-style. This is a lossy transition and would only support whatever fields are currently covered by the proprietary format.
3) Update all 74 native-capable bid adapters to call fromOrtbNative() at the top of buildRequests. Then we communicate with them to start the process of converting their adapters to not need fromOrtbNative().

FWIW, I poked around a couple of adapters: one had a complicated bunch of code for mapping native params and the other seemed to wrap it all up and parse server-side. So this transition may take a while.

Documentation-wise, the proposal is:
- produce a new document that is the recommended way that new native implementations should be done.
- leave the current doc in place but warn that new implementations should be done in the ortb-way
- hide the old-old doc away, making it more clear that no one should be using that (sendId) approach anymore.

musikele:
Closing this one and continuing the work in #8086 .

Issue #7765, opened 2021-11-27
Ressurecting an old issue here (ref #3270)


ocm-peter:
## Type of issue
Old bug came back to life, seen in v4.43.3 and v6.30-pre

## Description
For video cached bids, auctionEnd (and requestBids) fires before the actual bidResponse comes in.

![image](https://user-images.githubusercontent.com/64111309/143685914-081d1041-801d-420c-b201-93d712ab99c5.png)

## Steps to reproduce
```
var adunits = [
  {
    code: 'video',
    mediaTypes: {
        video: {
            context: 'instream',
            playerSize: [[640, 360]],
            mimes: ['video/mp4', 'video/x-ms-wmv', 'video/3gpp', 'video/ogg', 'video/x-msvideo', 'video/mpeg', 'video/mpg', 'video/webm', 'video/m2t', 'video/3gpp2'],
            protocols: [1, 2, 3, 4, 5, 6, 7, 8],
            playbackmethod: [1, 2, 5, 6],
            linearity: 1,
            startdelay: 0,
            skip: 1
        },
    },
    bids: [
    {
        "bidder": "pubmatic",
        "params": {
            "adSlot": "4061784@640x360",
            "publisherId": "157884",
            "video": {
                "placement": 0
            }
        }
    }
    ]
}
];

 window.oispbjs = window.oispbjs || {que:[]}

window.pbjs.que.push(() => {
    window.oispbjs.addAdUnits(adunits);
    window.oispbjs.setConfig({
        cache: {
            url: 'https://prebid.adnxs.com/pbc/v1/cache'
        },
    })

```

```
Prebid 6.30pre built with below modules
[
  "dfpAdServerVideo",
  "appnexusBidAdapter",
  "criteoBidAdapter",
  "sharethroughBidAdapter",
  "freewheel-sspBidAdapter",
  "ixBidAdapter",
  "pubmaticBidAdapter",
  "rubiconBidAdapter",
  "spotxBidAdapter",
  "openxBidAdapter",
  "sovrnBidAdapter",
  "brightcomBidAdapter",
  "tripleliftBidAdapter",
  "adfBidAdapter",
  "aolBidAdapter",
  "browsiRtdProvider",
  "consentManagement",
  "consentManagementUsp",
  "currency",
  "schain"
]
```

## Test page
[here](https://sandbox.orangeclickmedia.com/demos/instream/index.html?pubmatic_debug&pubmaticTest=true&pbjs_debug=true)

### Expected results
[...]

PR #7851, opened 2021-12-11
Prebid core & currency module: fix race condition on slow fetch of currency conversion file

dgirardi:
## Type of change

- [x] Bugfix

## Description of change
﻿The currency module hooks into the auction's `addBidResponse` flow to modify bid responses with currency info. To do this it needs to fetch conversion rates from the network, but the auction expects `addBidResponse` to complete synchronously, and things break down if all bidder network calls complete before the rates come back.

This fix allows the `addBidResponse` hook to return a promise. If it does, the auction will wait for it before ending.

## Other information

Addresses https://github.com/prebid/Prebid.js/issues/7765


aleksatr:
LGTM but, can we add a 2nd reviewer with a bit more experience in this area, to take a look?

mmoschovas (approved):

aleksatr (approved):

PR #7852, opened 2021-12-11
Adquery ID System: add new ID module

adquery:
## Type of change

- [x] Feature

## Description of change
added adqueryID module

### DIFF
--- a/test/spec/modules/adqueryIdSystem_spec.js
+++ b/test/spec/modules/adqueryIdSystem_spec.js
@@ -1,7 +1,25 @@
 import { adqueryIdSubmodule, storage } from 'modules/adqueryIdSystem.js';
 import { server } from 'test/mocks/xhr.js';
+import {amxIdSubmodule} from '../../../modules/amxIdSystem';
+import * as utils from '../../../src/utils';
+
+const config = {
+  storage: {
+    type: 'html5',
+  },
+};
 
 describe('AdqueryIdSystem', function () {
+  describe('qid submodule', () => {
+    it('should expose a "name" property containing qid', () => {
+      expect(adqueryIdSubmodule.name).to.equal('qid');
+    });
+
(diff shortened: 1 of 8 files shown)


test/spec/modules/adqueryIdSystem_spec.js
```diff
@@ -0,0 +1,56 @@
...
+      const callbackSpy = sinon.spy();
+      const callback = adqueryIdSubmodule.getId(config).callback;
+      callback(callbackSpy);
+      expect(callbackSpy.lastCall.lastArg).to.deep.equal({qid: 'qid'});
+    });
+
+    it('allows configurable id url', function() {
+      const config = {
+        params: {
+          url: 'https://bidder.adquery.io/prebid/qid'
```
smenzer, line 44: you should probably test with a value that isn't the default to prove that you can properly override

smenzer (changes requested):
the code written so far looks good except for one comment in your tests. however, you're missing quite a few other files that need to be updated, as well as a link to the docs repo.

other files to touch include: 
- .submodules.json
- eids.js
- userId.md
- eids_spec.js
- userId_spec.js

test/spec/modules/adqueryIdSystem_spec.js
adquery: I added missing files and fixed test

smenzer (approved):
LGTM

adquery:
Hi,

commit was merged into prebid:master on 22 Dec 2021 but I can't see  adquery module for User ID on https://docs.prebid.org/download.html
should I see it on the website already?

Best,

adquery:
Hi,

I can't see the adquery module for User ID at https://docs.prebid.org/download.html (attached screen)
please help what else should I do to appear on the page in the "User ID Modules" section for download?

image

Thanks and best

PR #7855, opened 2021-12-13
Kubient Bid Adapter: support coppa, update sync URL

mapcuk:
## Type of change
- [x] Feature

## Description of change

Bidder
- pass coppa to bid request

User Sync
- replace kdmp.kbntx.ch by matching.kubient.net
- remove iframe support
- rename us_privacy -> usp
- rename consent_str -> consent
- use URLSearchParams for GET params

---
- contact email of the adapter’s maintainer marsel.taipov@kubient.com


alexartwww:
Please help us with ci/circleci: build error. We can't understand where error 104) should overwrite logic of deprecated fields when filterSettings is defined      user sync comes from

ChrisHuie:
@alexartwww not really sure about these specific errors but your branch is decently behind master. Can you pull in the recent commits and see if that fixes these errors?

alexartwww:
@ChrisHuie I did, but the same issue

ChrisHuie:
@alexartwww I just rerun but looks like the problem is persisting. I will try and look into these errors also.

mapcuk:
Added Kubient adapter docs update https://github.com/prebid/prebid.github.io/pull/3476

modules/kubientBidAdapter.js
```diff
@@ -109,31 +114,37 @@ export const spec = {
...
+    }
+
+    let values = {};
+
     if (gdprConsent && typeof gdprConsent.consentString === 'string') {
-      gdprParams = `?consent_str=${gdprConsent.consentString}`;
+      values['consent'] = gdprConsent.consentString;
       if (typeof gdprConsent.gdprApplies === 'boolean') {
-        gdprParams = gdprParams + `&gdpr=${Number(gdprConsent.gdprApplies)}`;
+        values['gdpr'] = Number(gdprConsent.gdprApplies);
```
patmmccann, line 128: I believe gdpr can apply even if there isnt a consent string, see #7775

alexartwww: Plz review again. Plz explain why we can't pass CI tests. I've spent a lot of time to check all this. We made our adapter similar to documentation.

ChrisHuie:
@dgirardi can you look at this pr? Trying to figure out where the tests failures are from.

test/spec/modules/kubientBidAdapter_spec.js
```diff
@@ -175,6 +182,94 @@ describe('KubientAdapter', function () {
       });
     }
   });
+  describe('buildRequestBanner', function () {
```
dgirardi, line 185: test failures are coming from here (if you remove this describe block, they pass). Reason is `config` is global state and here it's being manipulated outside of the test lifecycle. This needs to be refactored so that calls to `config.setConfig`, `config.resetConfig` etc happen inside `before` / `after` / `beforeEach` ...

alexartwww:
@ChrisHuie @dgirardi Hello! please review our code. Thanks

ChrisHuie (approved):

PR #7861, opened 2021-12-14
Prebid Core: async submodule loading

AdmixerTech:
## Type of change

- [ ] Bugfix
- [x] Feature
- [ ] New bidder adapter   
- [ ] Build related changes
- [ ] CI related changes
- [ ] Other

## Description of change

if submodule loaded asynchronously (after hook ready) it will never be installed in module. 

we store install method and call it immediately if it exist.

### DIFF
--- a/src/hook.js
+++ b/src/hook.js
@@ -13,14 +13,18 @@ export function setupBeforeHookFnOnce(baseFn, hookFn, priority = 15) {
     baseFn.before(hookFn, priority);
   }
 }
+const submoduleInstallMap = {};
 
 export function module(name, install) {
   hook('async', function (submodules) {
     submodules.forEach(args => install(...args));
+    submoduleInstallMap[name] = install;
   }, name)([]); // will be queued until hook.ready() called in pbjs.processQueue();
 }
 
 export function submodule(name, ...args) {
+  const install = submoduleInstallMap[name];
+  if (install) return install(...args);
   getHook(name).before((next, modules) => {
     modules.push(args);
     next(modules);
--- a/src/hook.js
+++ b/src/hook.js
@@ -15,10 +15,10 @@ export function setupBeforeHookFnOnce(baseFn, hookFn, priority = 15) {
 }
 const submoduleInstallMap = {};
 
-export function module(name, install) {
+export function module(name, install, {postInstallAllowed = false} = {}) {
   hook('async', function (submodules) {
     submodules.forEach(args => install(...args));
-    submoduleInstallMap[name] = install;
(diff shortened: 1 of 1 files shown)


jsnellbaker:
I'm not entirely familiar with the use-cases involved, so forgive this question if it's not an issue.  Can this lead to a sub-module being installed twice?  Do we need some other internal flag to signal it was already done the normal way, so that's not called again?

src/hook.js
```diff
@@ -13,14 +13,18 @@ export function setupBeforeHookFnOnce(baseFn, hookFn, priority = 15) {
...
 export function module(name, install) {
   hook('async', function (submodules) {
     submodules.forEach(args => install(...args));
+    submoduleInstallMap[name] = install;
   }, name)([]); // will be queued until hook.ready() called in pbjs.processQueue();
 }
 
 export function submodule(name, ...args) {
+  const install = submoduleInstallMap[name];
+  if (install) return install(...args);
```
dgirardi, line 27: `args` here is not the same as the `args` passed by the "normal" install.

More generally, I'd prefer if modules that accept post-install submodules had to do that explicitly (maybe with something like `asyncModule` / `asyncSubmodule`) - or alternatively, depending on the use case here, review all current modules to make sure that they work with this, preferably with some test cases. I know the RTD module would not have worked with this until very recently, I'd expect more to assume there's only one install.

dgirardi (changes requested):

src/hook.js
dgirardi: I was wrong - `args` is indeed the same, disregard that. 

But I am still worried about this working in the general case. For example, RTD looks at submodules when `config` changes:

https://github.com/prebid/Prebid.js/blob/2740872e8def571ab4f1142fe1235ac5bf8a439b/modules/rtdModule/index.js#L220-L233

So if an rtd submodule is loaded after configuration it would not be seen until configuration is changed again.

AdmixerTech: @dgirardi 
`submodules` from `module(...)` is the same as `modules` from `sobmodule(...)` am i right ?
then why `args` is not the same ?
we push `args` into `modules /* submodules */` and then for eacha call `install(...args)`

AdmixerTech:
@jsnellbaker we check what submodules already installed and what need to be installed on our side.

src/hook.js
dgirardi: Yes you're right - sorry about that I used my fingers before my brain. LGTM now.

dgirardi (approved):

jsnellbaker (approved):

PR #7872, opened 2021-12-17
RTD Module: add getAdUnitTargeting to the auctionEnd event

mcallari:
## Type of change
- [x] Bugfix

## Description of change
Adding in the getAdUnitTargeting() call inside the setEventsListeners() method.

## Other information
I noticed this function call was removed in: https://github.com/prebid/Prebid.js/commit/59895632226ed89dc1b78b1e89f5b50f8cd3c61f#diff-6243b20988efe8c84d52352b6115dd25be399540b466384651a51a8eeb937b30L222

If this correct, is there another way setting targeting should be called now? 

If not, I am providing this patch in the event this function needs to be added back in here.

### DIFF
--- a/modules/rtdModule/index.js
+++ b/modules/rtdModule/index.js
@@ -203,6 +203,9 @@ const setEventsListeners = (function () {
         [CONSTANTS.EVENTS.BID_REQUESTED]: 'onBidRequestEvent'
       }).forEach(([ev, handler]) => {
         events.on(ev, (args) => {
+          if(ev === 'auctionEnd') {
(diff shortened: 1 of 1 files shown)


dgirardi:
@mcallari I somehow completely missed that call in my change. I was not able to update this PR (no repo access) so I posted https://github.com/prebid/Prebid.js/pull/7877 with linter fix and a test. Feel free to reopen this if needed.

mcallari:
@dgirardi ok that sounds good! Thanks for the update.

PR #7874, opened 2021-12-20
Prebid Core: creative comment injection spot reverted

denys-berzoy-confiant:
- reverted injection point of creative comment to pre-PR #6860
- added code to reinject comment in case it was removed upon rendering

## Type of change
- [x] Bugfix

## Description of change
This is an adjustment to the code previously pushed (#6860) to address the issue of disappearing bid detail comments in the document. We (Confiant) as a provider of ad quality and security need to have access to the bid detail comment before rendering starts, to match it against our data. The previous push broke our logic. We would like to put the injection of the comment back to the pre-rendering stage and only re-inject it again if it was in fact removed by the adaptor rendering as noticed by the previous developer (@jsut).

### DIFF
--- a/src/prebid.js
+++ b/src/prebid.js
@@ -403,9 +403,22 @@ function emitAdRenderSucceeded({ doc, bid, id }) {
 }
 
 /**
+ * This function will check for presence of given node in given parent. If not present - will inject it.
+ * @param {Node} node node, whose existance is in question
+ * @param {Document} doc document element do look in
+ * @param {string} tagName tag name to look in
+ */
+function reinjectNodeIfRemoved(node, doc, tagName) {
+  const injectionNode = doc.querySelector(tagName);
+  if (!node.parentNode || node.parentNode !== injectionNode) {
+    insertElement(node, doc, tagName);
+  }
+}
+
+/**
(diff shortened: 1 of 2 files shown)


jsut:
It seems weird to me that you've built code on top of of this functionality. Is hooking BID_WON, or AD_RENDER_SUCCEEDED not sufficient to your use case? That's actually a documented API that people acknowledge others rely on and should be changed as lightly as this was previously.

All that said, all I care about is that the comment is still in the DOM after rendering, no matter how the the ad is rendered, so the code itself /seems/ fine to me at a glance.  I'm just worried that there is nothing in this PR that will make people think twice about changing this in the future, so we might end up right back here again.

denys-berzoy-confiant:
@jsut simulating necessary behaviour here is a bit painful, but I'll add the tests. It just takes some time to do necessary manipulations inside the existing testing framework, since the Document object is just a mock and doesn't do its internal magic for us.

jsnellbaker (approved):
I think this looks okay.  Needs a 2nd approval before merging.

mmoschovas (approved):
lgtm

Issue #7871, opened 2021-12-17
RTD Module not setting targeting


mcallari:
## Type of issue
Bug

## Description
Targeting is no longer being applied when using the OptimeraRTD module

## Steps to reproduce
Got to the integration example page http://localhost:9999/integrationExamples/gpt/optimeraRtdProvider_example.html

You will not see the optimera key with targeting values in the ad request.

### Expected results
optimera=A6,B3,OD6,OD5,OD4,OD3|optimera=D4,OD6,OD5,OD4,OD3 should show in the ad request on the integration page.

### Actual results
No  optimera key and values in the ad request.

## Platform details
Prebid 6.4.0

peczenyj:
Hello.

This bug also affect weboramaRtdModule

we push an update on this module, released on 6.3.0, the same version thar includes this issue.

any idea when this fix will be available

Regards

PR #7877, opened 2021-12-20
RTD module: set targeting on onAuctionEnd event

dgirardi:
## Type of change

- [x] Bugfix

## Description of change
Re-introduce targeting logic for RTD onAuctionEnd event

Addresses https://github.com/prebid/Prebid.js/issues/7871
Fix from https://github.com/prebid/Prebid.js/pull/7872 , this just adds a unit test


Fawke (approved):
LGTM

ChrisHuie (approved):

Issue #7688, opened 2021-11-09
Support for Prebid Server Bid Adapter to support 'extra' bids


jsnellbaker:
## Type of issue
Request/Issue

## Description
The appnexus PSP endpoint used through the Prebid Server Bid Adapter file has a use-case that currently doesn't fit with the existing logic/workflow.  

### Some Context
The PSP endpoint normally communicates with different configured bidders through an appnexus placement to determine the ideal fill for a given request - picking a single bid to return to prebid.js.  It also has the ability to return multiple bids from different bidders as its response, instead of just the single 'winning' bid; it's similar to what our `enableSendAllBids` setting does.

For example, 1 request to an appnexus placement could return 2 bid responses, 1 appnexus bid + 1 pubmatic bid (or rubicon, or whoever).

### Issue
However in PBS bid adapter there is logic in place to map the bid response to its original bidRequest, based on the adUnitCode+bidderCode. Reference to the code in question here: https://github.com/prebid/Prebid.js/blob/master/modules/prebidServerBidAdapter/index.js#L867

Because the original request was made by the appnexus bidder alone, these incoming extra bids (that look like they came from other bidders) can't be properly mapped to the appnexus request.  We don't know all the time which bidders will respond for a request, so it could potentially change in real-world scnearios.

Given this logic is ran through the PBS bid adapter (which has it has to support its use-cases for the main PBS endpoints), I had some potential thoughts on different ways this use-case may be supported:
- alter the logic within the request payload to potentially pass in the bidId value (that's created for each bidder in each adUnit) > have the system keep a reference to this value and return it in the bid response - to signify which 'request' this bid is meant to associated to.  This approach would in theory replace the internal mapping logic done within the PBS Bid adapter file.  Perhaps we could use the ext fields somehow to pass along this data?  I'm honestly not sure if that would work in the expectations of the ORTB spec itself.
- create a retroactive bidRequest object when we can't find the original.  I did some initial tests with this approach and I can put up a draft PR to show-case what it may look like.  We may want to consider creating a proper retroactive bidderRequest object as well to have it work fully with the other features of prebid that look up bidderRequest/bidRequest to store/read values.
[...]

PR #7885, opened 2021-12-22
PBS Adapter: allow "extra" bids that are not tied to any request

dgirardi:
## Type of change

- [x] Feature

## Description of change

This allows the PBS adapter to return bids that were not explicitly requested (and do not match cleanly to any of the `bidRequest`s that were generated by the auction).

See discussion in https://github.com/prebid/Prebid.js/issues/7688
Documentation changes: https://github.com/prebid/prebid.github.io/pull/3491

The strategy is:

 - augment bid response objects to always have `auctionId` and `transactionId`. Their `requestId` is now optional (can be null).
 - the global `auctionManager` now has methods to look up auction, adUnit, and mediaType through the above. 
 - refactor every instance of "find this bid's request to look up auction, adUnit, or mediaType data". This removes the need for the request almost completely. There are two exceptions:

     - `bidderSettings.adServerTargeting` can specify custom functions that now are receiving bidRequest as an undocumented argument - this is likely unintentional.
     - the `mass` module allows [a custom render function](https://docs.prebid.org/dev-docs/modules/mass.html#mass-support-disabled-custom-renderer-enabled) that has the bidRequest as part of its `payload`. 

     In these cases the bidRequest would be undefined. For this reason they are discarded unless `s2sConfig.allowUnknownBidderCodes` is set.


modules/mass.js
```diff
@@ -78,7 +78,7 @@ export function updateRenderers() {
 /**
  * Before hook for 'addBidResponse'.
  */
-export function addBidResponseHook(next, adUnitCode, bid) {
+export function addBidResponseHook(next, adUnitCode, bid, {index = auctionManager.index} = {}) {
```
jsnellbaker, line 81: @snapwich It seems like this new `index` param doesn't exist on the original hooked function's call signature (see in the auction.js file).  

By adding this new variable here, would that cause any abnormalities with the underlying hook workflow?  I'm guessing from how this is used here (ie that it's totally local within this function), it may be fine - but wanted to double-check.

bretg:
@massadmin - please note that there's a minor change in the MASS module being done as part of this request

massadmin:
> @massadmin - please note that there's a minor change in the MASS module being done as part of this request

Thanks for the heads-up. Let me know if you need anything.

jsnellbaker (approved):
From what I can see through the code changes and from the testing I performed, this seems good.  Thanks for putting all these changes together.

Had a question for Rich to clarify, which was noted earlier.  Granting approval in the assumption that it will likely be fine.

bretg:
@dgirardi pointed out that this update doesn't fully solve the "prebid-server-only" scenario... i.e. an AdUnit without any bidders will be ignored by the PBS bid adapter.

If it's feasible, I'd suggest adding a simple(?) feature where s2sConfig.bidders could take the value '*', which means that all bidders are sent to PBS, even if all means none. It's ok if that's done in the scope of a different PR.

dgirardi:
@bretg I'd keep the "adunit with no bids" case in a separate PR, which should probably be about stored impression support. It's not just the PBS adapter but the auction itself that will ignore / reject an ad unit without bidders, so that's likely to be a bit too much complexity to add on top of this one.

jsnellbaker:
@dgirardi can you take another look here when you have the chance?  It's looks like there's some conflicts now.

aleksatr (approved):

PR #7886, opened 2021-12-23
JwPlayer RTD module: Write to oRTB content segments

karimMourra:
## Type of change

- [x] Feature

## Description of change
This PR allows the JW Player RTD module to write its contextual segments and content id on the ortb2 object and adds the JW Player segment  taxonomy code of 502.

- contact email of the adapter’s maintainer: karim@jwplayer.com

https://github.com/prebid/prebid.github.io/pull/3474

## Other information
https://github.com/prebid/prebid.github.io/pull/3474
[segment taxonomy code list](https://github.com/InteractiveAdvertisingBureau/openrtb/blob/master/extensions/community_extensions/segtax.md)

### DIFF
--- a/modules/jwplayerRtdProvider.js
+++ b/modules/jwplayerRtdProvider.js
@@ -15,6 +15,7 @@ import { ajaxBuilder } from '../src/ajax.js';
 import { logError } from '../src/utils.js';
 import find from 'core-js-pure/features/array/find.js';
 import { getGlobal } from '../src/prebidGlobal.js';
+import * as utils from '../src/utils'
 
 const SUBMODULE_NAME = 'jwplayer';
 const segCache = {};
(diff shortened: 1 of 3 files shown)


modules/jwplayerRtdProvider.js
```diff
@@ -235,6 +237,9 @@ export function getVatFromPlayer(playerID, mediaID) {
   };
 }
 
+/*
+  deprecated
+ */
```
Rothalack, line 242: What's the reasoning to mark this deprecated when it's still being called above? Not a big deal, just curious.

Rothalack:
I've attempted to test this myself but haven't found a way without asking you for some test player and media id's. Other than that and my question about the deprecated comment above, I think this is good.

modules/jwplayerRtdProvider.js
karimMourra: @Rothalack it's being called because we don't want to break the API. The call creates a targeting object which is set to`bid.rtd.jwplayer.targeting` which we are deprecating in favor of `ortb2.site`. We will remove the call once no SSP reads from `bid.rtd.jwplayer`

karimMourra:
@Rothalack you can use `2XZFlRuo` as a media id for testing.

Rothalack:
> 

Wouldn't I also need playerId? I've tried testing with just that mediaId and I am still getting ERROR: jwplayer.js was not found on page. I followed that code around and I'm pretty sure that it's looking for a playerId so that it can then load jwplayer.js. I shouldn't have to include that script manually on the test page. I'm using integrationExamples/gpt/jwplayerRtdProvider_example.html to test. Maybe this is an out of date example with more changes needed to work than I'm seeing.

I may be misunderstanding or this could be a mismatch of your documentation to the current working state.

karimMourra:
@Rothalack you can set the media id in the list of ids to be prefetched
```
pbjs.setConfig({
          realTimeData: {
            auctionDelay: 100,
            dataProviders: [{
              name: "jwplayer",
              waitForIt: true,
              params: {
                // Note: the following media Ids are placeholders and should be replaced with your Ids.
                mediaIDs: ['2XZFlRuo']
              }
            }]
          }
        });
```
We can't include a working player in the example  at the moment because our player is a commercial product that requires a license key and we don't want to leak license keys. 
We are exploring the possibility of generating demo keys, but for now rest assured, we've tested this on our end. This product generates revenue so it's in our best interest to ensure it works so that our partners continue using it. The unit tests have good coverage as well

Rothalack (approved):

Rothalack:
Fair enough, I understand!

patmmccann:
@karimMourra please see comments on #8320 ; it seems your rtd adapter is setting the segments in a rather unusual way. Was this intentional?

karimMourra:
@patmmccann the idea was to follow the standard; at the time we weren't aware that the standard was to write via `setConfig`. We will address this ASAP, thank you for reporting.

PR #7902, opened 2022-01-03
Adnuntius rtd provider: initial release

mikael-lundin:
## Type of change

- [ ] Bugfix
- [ ] Feature
- [x] New bidder adapter   
- [ ] Build related changes
- [ ] CI related changes
- [ ] Other

## Description of change

This is a pull request that will add adnuntius RTD module to prebid. 

- contact email of the adapter’s maintainer: prebid@adnuntius.com

https://github.com/prebid/prebid.github.io/pull/3433/commits/be09bd67f1cb3687462ffa85d9a8e2b65f09f4b5


modules/adnuntiusRtdProvider.js
```diff
@@ -0,0 +1,97 @@
+import { getGlobal } from '../src/prebidGlobal.js';
+import { submodule } from '../src/hook.js'
+import { logError, logInfo } from '../src/utils.js'
+import { ajax } from '../src/ajax.js';
+
+const GVLID = 855;
+
+function init(config, userConsent) {
+  if (!config.params) return false
+  if (!config.params.providers) return false
```
mmoschovas, line 10: Can we consolidate this into one conditional i.e. if (!config.params || !config.params.providers) return false;

modules/adnuntiusRtdProvider.js
```diff
@@ -0,0 +1,97 @@
...
+      success: function (res) {
+        const response = JSON.parse(res)
+        resolve(response)
+      },
+      error: function (err) { reject(err) }
+    });
+  });
+}
+
+function setGlobalConfig(config, segments) {
```
mmoschovas, line 42: This function should be merging ortb2 data not overwriting. There can very well be other useful segment data being set that this function would inadvertently remove. There should be usable mergeConfig/mergeBidderConfig functionality within src/config. You may need to merge in some commits from master though.

modules/adnuntiusRtdProvider.js
```diff
@@ -0,0 +1,97 @@
...
+  }
+}
+
+function alterBidRequests(reqBidsConfigObj, callback, config, userConsent) {
+  let allowedToRun = true
+  if (userConsent && userConsent.gdpr) {
+    logInfo('USR', userConsent.gdpr.gdprApplies)
+    logInfo('USR', userConsent.gdpr)
+    if (userConsent.gdpr.gdprApplies) {
+      if (!userConsent.gdpr.vendorData.vendorConsents[GVLID]) allowedToRun = false;
```
mmoschovas, line 70: This can be condensed a bit similar to the following:

```
function alterBidRequests(reqBidsConfigObj, callback, config, userConsent) {
  const gdpr = userConsent && userConsent.gdpr;
  let allowedToRun = true;
  if (gdpr) {
    logInfo('USR', gdprgdprApplies)
    logInfo('USR', gdpr)
    if (gdpr.gdprApplies && !gdpr.vendorData.vendorConsents[GVLID]) allowedToRun = false;
    ...
```

mmoschovas (changes requested):

modules/adnuntiusRtdProvider.js
mikael-lundin: Fixed in new pull request. :)

modules/adnuntiusRtdProvider.js
mikael-lundin: Fixed in new pull request. :)

modules/adnuntiusRtdProvider.js
mikael-lundin: Fixed in new pull request. :)

modules/adnuntiusRtdProvider.js
```diff
@@ -52,12 +51,12 @@ function setGlobalConfig(config, segments) {
     }
   }
   if (config.params && config.params.bidders) {
-    pbjsG.setBidderConfig(mergeBidderConfig({
+    pbjsG.setBidderConfig(mergeDeep(pbjsG, {
```
mmoschovas, line 54: This would merge the entire pbjs global object with ortb2 data...? This is definitely not what we want here. Was there an issue using mergeConfig/mergeBidderConfig? We want only the corresponding configs to be merged here.

mikael-lundin: I was not 100% sure on how to use it and this one seemt do be ok do you have an example on how I would use it?

mmoschovas: You would just need to pass in the data to each of the corresponding functions. They will handle retrieving the existing data, merging the two, and setting with the newly merged data. Not sure if there are docs or not on it, but it should be treated similarly to setConfig/setBidderConfig. I am seeing now the first time you implemented it was slightly incorrect because you were still calling setConfig/setBidderConfig. You should be able to just run as follows:

```
const pbjsG = getGlobal()
  const ortbSegments = {
    ortb2: {
      user: {
        data: [{
          name: 'adnuntius',
          segment: segments
        }]
      }
    }
  }
...
if (...) {
    pbjsG.mergeBidderConfig({
        bidders: config.params.bidders,
        config: ortbSegments
    })
} else {
    pbjsG.mergeConfig(ortbSegments)
}
```

mikael-lundin: That made more sense once I understood it, thank you! :D

mmoschovas (approved):
Thanks for making changes. lgtm

PR #7910, opened 2022-01-05
Kargo Bid Adapter: Fix issue with double-array for advertiserDomains

### LINKED ISSUE #7909: Kargo Bid Adapter: meta.advertiserDomains passing incorrectly
## Description
We are seeing scenarios where bid adapters are setting invalid `meta.advertiserDomains`

This field should be an array of strings

We noticed at least one adapter sending it as an double array with one string in it like so:

image

```
[['nytimes.com']]
```

I am unsure if this is happening ALL of the time for KARGO, or if there are certain scenarions where their server responds with a single adomain or it is always with the array of single element.

They get a bid response from their exchange =>

Notice that it returns `advertiserdomains` as an array of advertisers (which is exactly what it should be)

image

BUT their code wraps this inside of another array as seen here:

https://github.com/prebid/Prebid.js/blob/master/modules/kargoBidAdapter.js#L77

Which results in image

which is technically not correct!

We found this in our Analytics aggregations so I added a PR for us to handle this for now: https://github.com/prebid/Prebid.js/pull/7908

But want to let Kargo know just in case!

jsadwith:
## Type of change

- [x] Bugfix

## Description of change
This removes an extra array wrapped around advertiserDomains that was affecting the value from being read by Rubicon's analytics adapter (#7909)

## Other information
Resolves #7909 as reported by @robertrmartinez

### DIFF
--- a/modules/kargoBidAdapter.js
+++ b/modules/kargoBidAdapter.js
@@ -74,7 +74,7 @@ export const spec = {
       if (adUnit.metadata && adUnit.metadata.landingPageDomain) {
         meta = {
           clickUrl: adUnit.metadata.landingPageDomain,
-          advertiserDomains: [adUnit.metadata.landingPageDomain]
+          advertiserDomains: adUnit.metadata.landingPageDomain
         };
       }
       bidResponses.push({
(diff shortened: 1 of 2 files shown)


robertrmartinez (approved):
Awesome! Thanks for such a quick fix!

PR #7916, opened 2022-01-06
IX Bid Adapter: prefer pageUrl to refererUrl

lksharma:
## Type of change

- [ ] Bugfix
- [x] Feature
- [ ] New bidder adapter   
- [ ] Build related changes
- [ ] CI related changes
- [ ] Other

## Description of change
Prioritizes pageUrl set via setConfig() over default document referrer

### DIFF
--- a/modules/ixBidAdapter.js
+++ b/modules/ixBidAdapter.js
@@ -492,6 +492,7 @@ function buildRequest(validBidRequests, bidderRequest, impressions, version) {
   // Get ids from Prebid User ID Modules
   let eidInfo = getEidInfo(deepAccess(validBidRequests, '0.userIdAsEids'), deepAccess(validBidRequests, '0.userId.flocId'));
   let userEids = eidInfo.toSend;
+  const pageUrl = getPageUrl() || deepAccess(bidderRequest, 'refererInfo.referer');
 
   // RTI ids will be included in the bid request if the function getIdentityInfo() is loaded
   // and if the data for the partner exist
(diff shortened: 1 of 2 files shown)


modules/ixBidAdapter.js
```diff
@@ -591,7 +591,7 @@ function buildRequest(validBidRequests, bidderRequest, impressions, version) {
     }
 
     if (bidderRequest.refererInfo) {
-      r.site.page = bidderRequest.refererInfo.referer;
+      r.site.page = config.getConfig('pageUrl') || bidderRequest.refererInfo.referer;
```
patmmccann, line 594: do you want to sanity check values?

patmmccann (changes requested):
Should you sanity check values? Why is pageUrl conditional on this other object?

modules/ixBidAdapter.js
lksharma: Good point. I've made changes to set `site.page` only if pageUrl is valid.

lksharma:
@patmmccann,  I've made changes to set `site.page` only if pageUrl is valid. Let me know if there is anything else that may need updating.

lksharma:
@patmmccann, is there anything else pending for the review?

patmmccann:
@lksharma sorry for the delay here? It seems like these changes in #8059 include this?

patmmccann:
This is ready to merge if you want

lksharma:
Hey @patmmccann, 

I'll be closing this MR. I've includes these changes in another open MR here: https://github.com/prebid/Prebid.js/pull/8059

Issue #7867, opened 2021-12-16
Investigate options for reducing package size with code elimination


bretg:
## Type of issue
optimization

## Description
There have been conversations about trying to make the overall PBJS built package smaller by allowing bid adapters to tag code as having a certain optional function. Publishers could then build a package that excludes certain code. For instance, if the publisher doesn't need Native support in their wrapper, it may be possible to save several KB by not including code related to support of the Native mediatype.

There were several ideas kicked around in PR https://github.com/prebid/Prebid.js/pull/6079

Moving that conversation here.

The original proposal from @pm-harshad-mane was a pre-processor model. e.g.:

```
// removeIf(disableNativeRelatedCode)
import { processNativeAdUnitParams, nativeAdapters } from './native.js';
// endRemoveIf(disableNativeRelatedCode)
```

A counter-proposal (from @dgirardi):

The reason this approach does reduce bundle size is that it removes live code that happens to never get run. I agree that it seems hard to maintain. To me the main issue seems to be that a lot of code looks for particular settings or request/response values that keep code alive, even if you (as the publisher) know they will never be there. I think a better approach would be to programmatically put "breakpoints" around access to those values. 
For example, instead of:
```
     switch (bid.mediaType) {
           case NATIVE: doNativeStuff();
     }
```
we could have something like:
```
  bid.ifNative(doNativeStuff);
```
and define ifNative based on build settings. This would still be a major refactor, but it has the advantage that it would automatically "cascade" dead code detection to all dependencies, you wouldn't have to (for example) care about imports that only get used inside doNativeStuff like you do here.

ChrisHuie:
A webpack proposal for this is referenced in #6179

patmmccann:
merged

PR #7935, opened 2022-01-12
Build system: upgrade webpack

dgirardi:
## Type of change
- [x] Build related changes

## Description of change

Upgrade to webpack 5.

Also enables [tree shaking](https://webpack.js.org/guides/tree-shaking/) (as per https://github.com/prebid/Prebid.js/issues/7867), but it does not appear to be that effective - there is no significant difference in the final bundle size. As I understand it, tree shaking is supposed to erase unused modules - so this is not surprising given that modules are included only if they are imported.

## Notes:

- **`talkadsBidAdapter.js`** is using arrow functions with `this` bound to `null`. This upgrade changed something in the transpilation so that it now (correctly) fails. I have "fixed" it in this PR, but I did not verify my changes beyond the unit tests.
- **`plugins/RequireEnsureWithoutJsonp.js`** did not appear to have any effect; I tested it by replacing [this line](https://github.com/prebid/Prebid.js/blob/master/plugins/RequireEnsureWithoutJsonp.js#L18) with gibberish (on master) and I could not find it anywhere in the build output. It is not compatible with webpack 5 so I removed it completely.
- after the upgrade, `sinon.stub` and similar methods will refuse to work on es6 modules. This is the reason for forcing commonjs (via babel) and the refactoring of `events.js`.
- **`AnalyticsAdapter.js`** is not included in prebid-core (this was explicit in master, and here that continues to be true because it's not imported from `src/prebid.js`). This means that including multiple analytics adapters in a bundle will result in duplicated code (issue https://github.com/prebid/Prebid.js/issues/7936).


dgirardi:
Converting this back to draft because I believe `usedExports` should logically erase some needed code in a way that would not be detected by unit tests - will re-enable after some more testing.

dgirardi:
Reopening this for review. I was concerned about situations of the kind described here: https://github.com/prebid/Prebid.js/issues/7936#issuecomment-1013337454

That is an unused export for the 'prebid-core' webpack entry, so I was expecting it to get removed after this update (and unit tests wouldn't notice since they use different entries). Turns out it isn't, and it may be connected to the `mode` setting (new to webpack 4), see https://github.com/webpack/webpack/issues/6992 

For the moment I think it makes sense to have this PR focused only on the upgrade, possibly following up with tweaking tree shaking config.

robertrmartinez (approved):
This worked for me 👍 

Think we need one more review though

dgirardi:
using `modules: commonjs` in babelrc prevents webpack from doing most optimizations. I'll figure out some other way to make sinon work and update.

dgirardi:
I have updated this to transpile to commonjs modules only during tests. I don't think that's great (system under test should ideally be as close to prod as possible) but I don't know of a better option (@jsnellbaker, @robertrmartinez any suggestions?)

The issue is - I believe - that at some point the webpack runtime started making es6 modules read-only and `sinon.stub` & co cannot work on them. The options I see are:

 - transpile away es6 modules, but only for tests (more below on why we don't want to for the build);
 - a huge refactor of everything so that tests don't have to call `sinon.stub` on module objects (e.g. instead of `export thingToStub`  do `export const things = { thingToStub }` or other similar tricks)
 - try to hack into the webpack code generation to make the module objects it generates mutable.

We want to upgrade webpack and let it take in es6 modules because it does work a lot better for dead code elimination (and the bundle size did decrease by a bit after the last update in this PR). I am running some tests with [this mock module](https://gist.github.com/dgirardi/1115e0f97675a3fd360922c956f89835) to see what the effect  would be on feature tags, and there's a marked difference in the output - commonjs basically disables all dead code detection:

```javascript
// on master  (webpack 3)
pbjsChunk([3], {
  887: function (t, e, n) {
    t.exports = n(888);
  }, 888: function (t, e, n) {
    'use strict';
    Object.defineProperty(e, '__esModule', {value: !0});
    var i = n(889), u = n(890), s = n(1), c = {};
    Object(u.a)((function () {
      c.if = 'inline-function';
    })), c.of = Object(u.a)(i.a), Object(s.registerBidder)(c), window.pbjs.installedModules.push('testFeatureTags');
  }, 889: function (t, e, n) {
    'use strict';
    e.a = function () {
      return 'someOtherFunctionality';
    };
  }, 890: function (t, e, n) {
    'use strict';
    e.a = function (t) {
      0;
    };
  }
}, [887]);
```

```javascript
// webpack 5, using es6 modules (babel/preset-env has "modules": "auto")
'use strict';
(self.pbjsChunk = self.pbjsChunk || []).push([[5947], {
    2159: function (s, e, t) {
        var u = {of: void 0};
        (0, t(1609).registerBidder)(u), window.pbjs.installedModules.push('testFeatureTags');
    }
}, function (s) {
    s(s.s = 2159);
}]);
```

```javascript
// webpack 5, transpiling to commonjs (babel/preset-env has "modules": "commonjs")
'use strict';
(self.pbjsChunk = self.pbjsChunk || []).push([[5947], {
  78145: function (e, n, t) {
    var i = t(79160), u = t(34319), o = t(1609), s = {};
    (0, u.ifFeatureEnabled)((function () {
      s.if = 'inline-function';
    })), s.of = (0, u.ifFeatureEnabled)(i.someOtherFunctionality), (0, o.registerBidder)(s), window.pbjs.installedModules.push('testFeatureTags');
  }, 79160: function (e, n) {
    Object.defineProperty(n, '__esModule', {value: !0}), n.someFunctionality = function () {
      return 'someFeatureFunctionality';
    }, n.someOtherFunctionality = function () {
      return 'someOtherFunctionality';
    };
  }, 34319: function (e, n) {
    Object.defineProperty(n, '__esModule', {value: !0}), n.ifFeatureEnabled = function (e) {
    };
  }
}, function (e) {
  e(e.s = 78145);
}]);
```

jsnellbaker:
@dgirardi I'm not sure off-hand about the test-only behavior.  But I wanted to highlight something.  When I was testing some gulp commands with this PR, I was seeing the following error from webpack:
```
12033-p-mac:Prebid.js jsnellbaker$ gulp build
ReferenceError: BigInt is not defined
    at Object.<anonymous> (/Users/jsnellbaker/git/Prebid.js/node_modules/webpack/lib/ChunkGraph.js:43:22)
    at Module._compile (module.js:643:30)
    at Object.Module._extensions..js (module.js:654:10)
    at Module.load (module.js:556:32)
    at tryModuleLoad (module.js:499:12)
    at Function.Module._load (module.js:491:3)
    at Module.require (module.js:587:17)
    at require (internal/module.js:11:18)
    at Object.<anonymous> (/Users/jsnellbaker/git/Prebid.js/node_modules/webpack/lib/Compiler.js:20:20)
    at Module._compile (module.js:643:30)
```

This occurred when I was using an older version of node (8.9.4) - which is what the build server we use for releases is currently configured to use.  When I use a higher version of node, the build command works fine.  Is there any config setting that could be used within webpack to help support the older version?  Or does the upgrade to webpack 5 require the higher node?

jsnellbaker:
Can we please hold off on merging this then?  If it goes through it will interrupt/block the release processes.  

We're trying on our end to update some things, but it's going to take time.

jsnellbaker (approved):

patmmccann:
@robertrmartinez can you sign off on changes that have occurred after your review?

ChrisHuie (approved):

PR #7955, opened 2022-01-21
Richaudience Bid Adapter: uncaught error on non-gdpr locations

michae1:
## Type of change
- [ ] Bugfix

## Description of change
Richaudience causes errors on non-gdpr locations: 
```Cannot read properties of undefined (reading 'gdprApplies')```
because bidderRequest.gdprConsent is undefined


michae1:
Looks like it is related to this commit and the related comment:
https://github.com/prebid/Prebid.js/commit/fb0a73aab7a366a80db92cd06b94ff71fa24652c#r61200506

PR #7969, opened 2022-01-26
Prebid Core: add filename to pbjsGlobals module append

miguelpeixe:
## Type of change

- [x] Bugfix
- [ ] Feature
- [ ] New bidder adapter   
- [x] Build related changes
- [ ] CI related changes
- [ ] Other

## Description of change

There is a conflict while building modules if `@babel/preset-typescript` is configured through a root `babel.config.js` file. From my understanding of babel configs, adding `configFile: false` to the module rule configuration should be able to resolve, but it doesn't.

The solution I encountered was adding the `filename` property to the `api.parse()` config. This resolves the build without regression or conflicts with `@babel/preset-typescript` requirements.

This demo repository allows you to test the error, note the config inside `babel.config.js` and the `configFile: false` at `webpack.config.js`: https://github.com/miguelpeixe/prebid-config-test

Confirm the error by running `npm ci && webpack`:

```
[BABEL] unknown: Preset /* your preset */ requires a filename to be set when babel is called directly,
babel.transform(code, { filename: 'file.ts', presets: [/* your preset */] });
See https://babeljs.io/docs/en/options#filename for more information.
```

### DIFF
--- a/plugins/pbjsGlobals.js
+++ b/plugins/pbjsGlobals.js
@@ -39,7 +39,7 @@ module.exports = function(api, options) {
         const modName = getModuleName(state.filename);
         if (modName != null) {
           // append "registration" of module file to $$PREBID_GLOBAL$$.installedModules
-          path.node.body.push(...api.parse(`window.${pbGlobal}.installedModules.push('${modName}');`, {filename: modName}).program.body);
+          path.node.body.push(...api.parse(`window.${pbGlobal}.installedModules.push('${modName}');`, {filename: state.filename}).program.body);
         }
       },
       StringLiteral(path) {
--- a/plugins/pbjsGlobals.js
+++ b/plugins/pbjsGlobals.js
@@ -39,7 +39,7 @@ module.exports = function(api, options) {
         const modName = getModuleName(state.filename);
         if (modName != null) {
           // append "registration" of module file to $$PREBID_GLOBAL$$.installedModules
-          path.node.body.push(...api.parse(`window.${pbGlobal}.installedModules.push('${modName}');`).program.body);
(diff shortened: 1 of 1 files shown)


plugins/pbjsGlobals.js
```diff
@@ -39,7 +39,7 @@ module.exports = function(api, options) {
         const modName = getModuleName(state.filename);
         if (modName != null) {
           // append "registration" of module file to $$PREBID_GLOBAL$$.installedModules
-          path.node.body.push(...api.parse(`window.${pbGlobal}.installedModules.push('${modName}');`).program.body);
+          path.node.body.push(...api.parse(`window.${pbGlobal}.installedModules.push('${modName}');`, {filename: modName}).program.body);
```
dgirardi, line 42: Thanks for figuring this out! 
The actual filename is under `state.filename`. Reading through https://babeljs.io/docs/en/options#filename it seems better to use that, because I can imagine a situation where `modName` clashes with some actual file in your build and babel makes the wrong decision on it.

miguelpeixe: That makes sense! Haven't tested it yet, but what do you think about appending the `modName` to the `state.filename` so each module keeps its own name? `'${state.filename}.${modName}'`. I don't know if having all modules with the same file name creates some conflict in this case.

dgirardi: this piece of code runs once for each module file, so that should not be a problem - `state.filename` will be different for each instance.

miguelpeixe: ab779edc95032fb36940aeb2eac03ad1dc95bc11 updates the filename and the build continues to work as expected locally

dgirardi (approved):

patmmccann (approved):

PR #7975, opened 2022-01-27
Improve Digital adapter: custom creative renderer

jbartek25:
## Type of change
- [x] Feature

## Description of change
Improve Digital adapter: adding custom creative renderer

Doc PR: https://github.com/prebid/prebid.github.io/pull/3531


modules/improvedigitalBidAdapter.js
```diff
@@ -10,6 +10,70 @@ const BIDDER_CODE = 'improvedigital';
 const RENDERER_URL = 'https://acdn.adnxs.com/video/outstream/ANOutstreamVideo.js';
 const VIDEO_TARGETING = ['skip', 'skipmin', 'skipafter'];
 
+class Razr {
```
FilipStamenkovic, line 13: Why creating a class with all `static` functions?  
You could just define and use functions without defining a class.

jbartek25: It was to group related methods and variables in a single entity. It looks nicer than for example prefixing the function and variable names with the "razr". Another options would be to replace the static class with a plain JS object. What is a preferred convention?

FilipStamenkovic: I don't think there is a preferred convention. You can leave like this. I just prefer clean functions over OOP (when OOP is not necessary).

jbartek25: I agree, using a static class is a bit of a stretch here. However, we have a brand new Improve Digital adapter that will replace the existing one hopefully in the next couple of weeks. I'll have the class replaced by an object literal similar to [here](https://github.com/prebid/Prebid.js/blob/master/modules/bluebillywigBidAdapter.js#L31) so if that's ok, please review as is.

jbartek25: @FilipStamenkovic can we proceed?

FilipStamenkovic: @ChrisHuie can you assign someone else to take a look at this PR? At the moment I don't have a spare time to dig into it.

jbartek25: Hey @ChrisHuie. Would it possible to get a rough ETA for when the PR will get reviewed? Thanks

ChrisHuie: @jbartek25 added a reviewer to take a look :)

Rothalack: I don't necessarily feel qualified to make the call on this question. From my perspective, it works and you're planning to replace this in the future, so I'm failing to see any consequence really.

modules/improvedigitalBidAdapter.js
```diff
@@ -10,6 +10,70 @@ const BIDDER_CODE = 'improvedigital';
...
+    });
+
+    if (!this.loaded) {
+      const s = document.createElement('script');
+      s.type = 'text/javascript';
+      s.async = true;
+      s.src = 'https://razr.improvedigital.com/renderer.js';
+
+      const x = document.getElementsByTagName('script')[0];
+      x.parentNode.insertBefore(s, x);
```
patmmccann, line 57: please use ajax methods

patmmccann:
One thing of note, and it doesnt disqualify this pr, but #6500 may already support your use case

modules/improvedigitalBidAdapter.js
jbartek25: The script to be loaded will add logic to the page. Can you please elaborate on how and why this should be done with ajax?

jbartek25:
@patmmccann  ah, I missed the remark about https://github.com/prebid/Prebid.js/pull/6500. Yes, we looked at it before and it suits our company needs better to have a custom renderer.

modules/improvedigitalBidAdapter.js
patmmccann: because it is a prebid rule

patmmccann: you can use ajax like this https://github.com/prebid/Prebid.js/pull/7699/files or you can import the loadExternalScript function from utils. You cannot you the logic here.

jbartek25: Thank you. We'll review the options.

patmmccann:
@jbartek25 here is another reference implementation

https://github.com/prebid/Prebid.js/blob/4378f2cc8b47b331e4b94a9fa3ec5f3962b38e71/modules/oneVideoBidAdapter.js#L139

modules/improvedigitalBidAdapter.js
patmmccann: ty; I noted on the main thread, the logic in the oneVideo adapter would likely work for you as well.

jbartek25: code updated to rely on `executeRenderer`

jbartek25:
@patmmccann does the code look good to you now?

patmmccann (approved):
yes, thank you!

PR #7978, opened 2022-01-27
Prebid core & PBS adapter: POC - feature tags and optional compilation of native support

dgirardi:
## Type of change

- [x] Feature

## Description of change

This adds feature flags that can remove portions of the codebase at build time (see https://github.com/prebid/Prebid.js/issues/7867). **Consider this a proof of concept**, because it depends on (and is diffed against) https://github.com/prebid/Prebid.js/pull/7935 ([this is why](https://github.com/prebid/Prebid.js/pull/7935#issuecomment-1022546699)) - I will need to post a new PR when that is merged. Here I am looking for feedback on the general strategy and naming conventions.

The first (and only) flag defined here is "NATIVE", which is the easiest (and smallest) to manage; only core and PBS are included - with the expectation that adapters will over time mark their feature-specific code as is done here.

This is the effect on the bundle size:

```
demetrio@pbws:~/src/Prebid.js$ gulp build-bundle-prod --modules=prebidServerBidAdapter > /dev/null 2>&1 && ls -l build/dist/prebid.js
-rw-rw-r-- 1 demetrio demetrio 176791 Jan 27 11:21 build/dist/prebid.js
demetrio@pbws:~/src/Prebid.js$ gulp build-bundle-prod --modules=prebidServerBidAdapter --disable=native > /dev/null 2>&1 && ls -l build/dist/prebid.js
-rw-rw-r-- 1 demetrio demetrio 168356 Jan 27 11:21 build/dist/prebid.js
```

A readable comparison of the output: https://gist.github.com/dgirardi/c506a2aa243125cc2a8bdcfa27e10d7a/revisions

## Other information

**Testing**: the strategy is, for now, to add an additional test run with all features disabled. Once we have more flags defined I intend to revisit this - it will probably be too expensive to run the whole test suite on all feature combinations, but I am optimistic that we can find a good heuristic to identify "relevant" tests and be exhaustive only on those (most tests do not interact with the code being removed).

**Compilation**: I have decided to use our Babel plugin to resolve feature flags (as opposed to webpack's `DefinePlugin` or other options like terser arguments). The reason is that it works better for npm consumers: no change is required to keep the build working, and there's no risk of our globals (i.e. `FEATURES`) clashing with their namespace.


dgirardi:
Closing in favor of https://github.com/prebid/Prebid.js/pull/8219

Issue #7956, opened 2022-01-21
setBidderConfig is merging values into global config


bozghiyy:
## Type of issue
bug

## Description
When using `setBidderConfig` to set some custom config for a specific bidder, the values get merged into the global config. 
I've noticed it when trying to setup a custom schain for a specific bidder, but I think this applies to other config properties. (didn't tested but I assume First Party Data or other specific bidder configurations)

Seems to be in connection with bid requests. When the bid requests are made for the bidder you specified the custom config with `setBidderConfig`, then it gets merged into global config. (and, of course, the sequential requests are made with an schain with merged nodes).

## Steps to reproduce
- `pbjs.setConfig` to set global schain
- `pbjs.setBidderConfig` to set schain for a specific bidder
- this only happens when bids are requested for the specific bidder (`pbjs.requestBids`)
- verify schain config if it has only the global node (`pbjs.getConfig('schain')`)

## Test page
Load the following pages and look for `pbjs.getConfig('schain')` console.log.

Issue on Prebid 6.8.0: http://stage.adops.ro/prebid_mergedeep_issue/6.8.0.html
Working correctly on Prebid 6.4.0: http://stage.adops.ro/prebid_mergedeep_issue/6.4.0.html

### Expected results
Global config should have just one schain node
![image](https://user-images.githubusercontent.com/15686691/150559009-aa08aaff-1fc9-44d1-8376-6f99220cfd2b.png)

### Actual results
Global config has two schain nodes:
![image](https://user-images.githubusercontent.com/15686691/150559187-0b70ddf3-b212-4b42-8f22-1d26a5d5c1ef.png)

## Platform details
Prebid 6.5.0 and higher

## Other information
I think it has to do with the modification of `mergeDeep` here https://github.com/prebid/Prebid.js/commit/31cc4b42f92d187022e63d183da3c2a385d96228#diff-3274f1a37032fb0ae4e2823def0007c634e869ae0dfc304ff6a12c36513c3a52L1275-R1286

PR #7991, opened 2022-01-31
Prebid core: isolate global and bidder-specific configuration

dgirardi:
## Type of change

- [x] Bugfix

## Description of change

This fixes a bug where `getConfig` called from within a bidder "context" will sometimes merge bidder configuration into the global config.

Addresses https://github.com/prebid/Prebid.js/issues/7956

### DIFF
--- a/src/config.js
+++ b/src/config.js
@@ -305,7 +305,7 @@ export function newConfig() {
           memo[topic] = currBidderConfig[topic];
         } else {
           if (isPlainObject(currBidderConfig[topic])) {
-            memo[topic] = mergeDeep({}, config[topic], currBidderConfig[topic]);
+            memo[topic] = mergeDeep(deepClone(config[topic]), currBidderConfig[topic]);
           } else {
             memo[topic] = currBidderConfig[topic];
           }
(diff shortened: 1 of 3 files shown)


dgirardi:
@msm0504 IMO, no. The previous `mergeDeep` would merge arrays by `concat`, and the new version merges into the original array (as you'd expect `merge` to behave). But the real issue is that in `getConfig` we only clone the top-level object, `mergeDeep` would still modify nested objects with bidder configuration.

msm0504:
@dgirardi But if the key isn't already in the `target` object, the existing array reference is used:
https://github.com/prebid/Prebid.js/blob/5abf5c3af94eba9736ccf6cdc623adb0f98ed767/src/utils.js#L1276
I'm not sure if this is correct for a deep merge function. Would changing this line to the following resolve the issue? 
```
Object.assign(target, { [key]: [ ...source[key] ] });
```

dgirardi:
@msm0504 I understand what you mean now, and I agree it makes sense to clone arrays if we are also cloning objects as we are doing in there. Updated.

msm0504 (approved):

PR #7992, opened 2022-01-31
Prebid core: bidder-specific control over storage access via `bidderSettings.allowStorage`

dgirardi:
## Type of change

- [x] Feature

## Description of change

This introduces the ability to turn on or off `StorageManager` functionality for specific bidders through `bidderSettings.allowStorage`.

Resolves https://github.com/prebid/Prebid.js/issues/7280

## Other information

Docs PR: (tbd)


src/storageManager.js
```diff
@@ -10,31 +11,44 @@ export let storageCallbacks = [];
...
  * - Module type: Some modules may need these functions but are not vendor. e.g prebid core files in src and modules like currency.
  * @param {storageOptions} options
  */
-export function newStorageManager({gvlid, moduleName, moduleType} = {}) {
+export function newStorageManager({gvlid, moduleName, bidderCode, moduleType} = {}, {bidderSettings = defaultBidderSettings} = {}) {
+  function isBidderDisallowed() {
+    if (bidderCode == null) {
+      return false;
+    }
+    const storageAllowed = bidderSettings.get(bidderCode, 'storageAllowed');
```
robertrmartinez, line 33: should this also try to use the moduleName as well?

When I try this:

```js
        pbjs.bidderSettings = {
            rubicon: {
                storageAllowed: false
            }
        };
```

And use the rubiconAnalyticsAdapter, I still see the module is allowed to set its storage:

image

robertrmartinez (changes requested):

src/storageManager.js
robertrmartinez: Also, is there a way to set this globally? as to deny storageAllowed for all in bidderSettings?

```js
        pbjs.bidderSettings = {
            standard: {
                storageAllowed: false
            },
        };
```

for example

dgirardi: @robertrmartinez, my understanding is that `bidderSettings` should control bid adapters, not analytics adapters (which is where `rubicon` controls its cookie from). That was the motivation for adding the new `bidderCode` parameter to the storage layer. The `moduleName` is treated as it was before, it can disable storage but you'd have to use consent management.

and yes, `bidderSettings.standard` is the fallback  - the logic is in `bidderSettings.js` (https://github.com/prebid/Prebid.js/pull/7712)

robertrmartinez: Ok, so even when I set 

```js
        pbjs.bidderSettings = {
            standard: {
                storageAllowed: false
            },
        };
```

It is not expected to stop Analytic Adapters from storing?

Ok 👍

robertrmartinez: Though, I would say, if a pub just wants to stop storage across the board, it kind of sucks they have to use the consentManagement module, if they do not even want consentManagement installed.

But I'll leave that to the deciders!

dgirardi: Updated https://github.com/prebid/prebid.github.io/pull/3540 to be more explicit on what this setting does.

robertrmartinez (approved):
👍 We spoke about this in the PMC

Fawke (approved):

patmmccann:
@dgirardi is it possible to make localstorage only an option? Related:  https://github.com/prebid/Prebid.js/pull/8140#pullrequestreview-899375129

Publishers may wish to prevent Kargo from using cookies but allow local  storage

dgirardi:
@patmmccann yes, as long as we agree on the setting syntax. We can have an update where instead of just `true`/`false` we understand also strings like  `localStorage` or `cookies` for example?

PR #7993, opened 2022-01-31
add a configurable "bidCacheFilterFunction"

harpere:
Add a configurable "bidCacheFilterFunction" to determine whether to use a "cached" bid.   The function should take a single `bid object` argument, and return `true` to use the "cached" bid, `false` not to use the "cached" bed.  This function will only be called for "cached" bids from previous auctions, not "current" bids from the most recent auction.  This function is only relevant if "bid caching" is turned on.  For Example, if you wanted to use bid caching, but not for video bids, you could do this:

```
pbjs.setConfig({
    useBidCaching: true,
    bidCacheFilterFunction: bid => bid.mediaType !== 'video'
});
```

## Type of change

- [ ] Bugfix
- [x] Feature
- [ ] New bidder adapter   
- [ ] Build related changes
- [ ] CI related changes
- [ ] Other

## Description of change

See above

- docs PR Link:  https://github.com/prebid/prebid.github.io/pull/3541

### DIFF
--- a/src/targeting.js
+++ b/src/targeting.js
@@ -438,7 +438,15 @@ export function newTargeting(auctionManager) {
     let bidsReceived = auctionManager.getBidsReceived();
 
     if (!config.getConfig('useBidCache')) {
+      // don't use bid cache (i.e. filter out bids not in the latest auction)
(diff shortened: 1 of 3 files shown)


robertrmartinez (approved):
Looks good!

PR #8004, opened 2022-02-03
OTM bid adapter: use top origin as default domain

regulyarniy:
## Type of change
- [ ] Bugfix
- [x] Feature
- [ ] New bidder adapter
- [ ] Build related changes
- [ ] CI related changes
- [ ] Other

## Description of change
use top level origin as a default 'domain' argument for bid request

## Other information
Need @krecu review as the OTM module maintainer
Need @ChrisHuie review

### DIFF
--- a/modules/otmBidAdapter.js
+++ b/modules/otmBidAdapter.js
@@ -1,10 +1,21 @@
 import { registerBidder } from '../src/adapters/bidderFactory.js';
-import {logInfo, logError, getBidIdParameter, _each, getValue, isFn, isPlainObject} from '../src/utils.js';
+import {
+  logInfo,
+  logError,
+  getBidIdParameter,
+  _each,
+  getValue,
+  isFn,
+  isPlainObject,
+  isArray,
+  isStr,
+  isNumber,
+} from '../src/utils.js';
 import { BANNER } from '../src/mediaTypes.js';
 
 const BIDDER_CODE = 'otm';
 const OTM_BID_URL = 'https://ssp.otm-r.com/adjson';
-const DEF_CUR = 'RUB'
+const DEFAULT_CURRENCY = 'RUB'
 
 export const spec = {
 
@@ -19,7 +30,7 @@ export const spec = {
    * @return boolean True if this is a valid bid, and false otherwise.
    */
   isBidRequestValid: function (bid) {
-    return !!bid.params.tid;
+    return Boolean(bid.params.tid);
   },
 
   /**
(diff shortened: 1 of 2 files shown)


krecu (approved):

regulyarniy:
Need @ChrisHuie review

modules/otmBidAdapter.js
```diff
@@ -132,15 +136,42 @@ function _getBidFloor(bid) {
...
   return 0;
 }
 
+/**
+ * Get top level origin
+ * @return {string}
+ */
+function getTopDomain() {
+  try {
+    return window.top.location.host
```
patmmccann, line 156: you can't do any of this; you have to use our methods.

patmmccann (changes requested):
you have to use our methods

modules/otmBidAdapter.js
regulyarniy: changed to `getWindowTop()`

regulyarniy:
@patmmccann thx for review! Could you please also look why the build is failing? I think it's temporary problem with bs, nor the code problem.

modules/otmBidAdapter.js
```diff
@@ -33,43 +46,37 @@ export const spec = {
...
 
     _each(validBidRequests, (bid) => {
-      let domain = getValue(bid.params, 'domain') || ''
-      let tid = getValue(bid.params, 'tid')
-      let cur = getValue(bid.params, 'currency') || DEF_CUR
-      let bidid = getBidIdParameter('bidId', bid)
-      let transactionid = getBidIdParameter('transactionId', bid)
-      let auctionid = getBidIdParameter('auctionId', bid)
-      let bidfloor = _getBidFloor(bid)
+      const domain = isStr(bid.params.domain) ? bid.params.domain : getTopDomain()
```
patmmccann, line 53: prebid already gets this for you with bidderRequest.refererInfo, which you grab above. Why are you adding this function?

patmmccann: in other words, https://github.com/prebid/Prebid.js/blob/master/src/refererDetection.js does this work for you, the rule to use our methods is to keep build sizes trim. If every adapter reproduced, build size would grow uncontrollably.

regulyarniy: now I see what u mean earlier, thx!
changed to use URL from `bidderRequest.refererInfo.referer`

patmmccann (approved):

PR #8006, opened 2022-02-03
Synacormedia Bid Adapter: ttl and eid update

ecammit:
## Type of change

- [ ] Bugfix
- [x] Feature
- [ ] New bidder adapter   
- [ ] Build related changes
- [ ] CI related changes
- [ ] Other

## Description of change
This updates the synacormedia bid adapter to:
- remove the existing filtering for eids to pass all in the bid request to the server
- obtains the TTL expiration for bids from the ext section of the bid response from the server. If the ext value is not present, the default TTL is now 7 minutes instead of 60 seconds.

## Other information
n/a


patmmccann:
@ecammit can you please link the docs pull indicating the id support change

ecammit:
> @ecammit can you please link the docs pull indicating the id support change

I'm not sure how to best to handle this. Internally we only still support those 4 id types listed in the docs, however, all ID types are still passed downstream by us to any other SSPs or DSPs that we forward our bids along to. This change was suggested by several of our SSP/DSP partners who do support other ID types that were being filtered out prior to this change.

Should we remove the "userIds" line in the docs, leave as-is with the 4 we support internally, or mark this up in some other way?

patmmccann:
I would mark up your docs with an explanation, it never hurts to overshare.

patmmccann (approved):

Issue #8023, opened 2022-02-08
Prebid 6.10.0 doesn't send bidRequest param to floors custom fields


JulieLorin:
## Type of issue
This is a bug

## Description
I am using custom schema fields alongside the priceFloors module. 
When creating a custom schema field, we create a function that has a bidRequest and a bidResponse as parameters (cf documentation : https://docs.prebid.org/dev-docs/modules/floors.html#custom-schema-fields)
Since Prebid 6.10.0, the bidRequest param is null for bidResponse. (The associated bid request is not sent alongside the bidResponse). 

## Steps to reproduce
Run an auction with the priceFloors module enabled, some floors defined, with a custom schema. 
For example, here is a custom field we use : 
```
connector: (bidReq, bidRes) => bidReq.bidder ?? bidRes.bidderCode
```
If you have bid responses, you will see an error because bidReq is null. 

### Expected results

I would exepect to have the bidReq parameter to be filled, as it used to be. 
It would be easy to fix it for this example, but we sometimes need the bid request (for example to retrieve a video context)

## Other information
It seems to be linked to this commit : 
https://github.com/prebid/Prebid.js/pull/7885/files#diff-52e7e567c667b4e2850bccc09785b9d48c3edb905487955df6437d7fc32eb216
The search for a "matchingBidRequest" has been removed

Issue #8024, opened 2022-02-08
Prebid 6.10.0 breaks the floor module (regression - Urgent)


JulieLorin:
## Type of issue
This is a bug, very much linked to issue #8023

## Description
The floor prices module is broken for the schema fields adUnitCode (and I think for gptSlot but I didn't test)

## Steps to reproduce
I used the following floorSettings : 
```
fields: [ 'adUnitCode' ],
      values: [
        { adUnitCode: 'dfp_ban_atf', floor: 1 },
      ],
      default: 0.5
```
When I have a bid of 0.4$ for dfp_ban_atf, it is filtered by the floor module (expected)
When I have a bid of 0.6$ for dfp_ban_atf, it is not filtered by the floor module (but it should)

## Other information
The bug comes from this PR : 
https://github.com/prebid/Prebid.js/pull/7885/files#diff-52e7e567c667b4e2850bccc09785b9d48c3edb905487955df6437d7fc32eb216

This is because the bidRequest field is not populated anymore for the bidResponses. 
This has been modified this way : 
![image](https://user-images.githubusercontent.com/635021/153010934-9d86648e-d0c0-4e66-9842-442e3656f512.png)
But the issue is that the property adUnitCode doesn't exist on bidResponses, only on bidRequests. 

So when bidRequest is null, `(bidRequest || bidResponse).adUnitCode` this gives bidResponse.adUnitCode, which is undefined. So the floor matching cannot work for adUnitCode on bid responses

PR #8025, opened 2022-02-08
Price Floor Module: retrieve matching bid request

JulieLorin:
## Type of change
- [X] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes
- [ ] Other

## Description of change
Send the matching bid Request alongside bidResponse in the priceFloors module

## Other information
This will fix issues #8023 #8024

### DIFF
--- a/modules/priceFloors.js
+++ b/modules/priceFloors.js
@@ -672,8 +672,11 @@ export function addBidResponseHook(fn, adUnitCode, bid) {
   if (!floorData || !bid || floorData.skipped) {
     return fn.call(this, adUnitCode, bid);
   }
+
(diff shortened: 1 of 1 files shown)


dgirardi:
I was unable to update this PR, so I posted https://github.com/prebid/Prebid.js/pull/8027

The reason bidRequest was no longer passed in that update is that it introduced the possibility of reponses without requests (https://github.com/prebid/Prebid.js/pull/7885); however, I missed that custom priceFloor schemas exist and that they might depend on a request. I had some additional changes for the fix to make sure that when custom schemas are not in use, those responses work as expected.

patmmccann (approved):

Issue #7831, opened 2021-12-08
pbsBidAdapter should support adunit.ortb2Imp


bretg:
## Type of issue
enhancement

## Description
To support the app post-bid scenario with OM signaling, we need to allow passing imp.banner.api through to Prebid Server.

Currently, the pbsBidAdapter only looks in ortb2Imp.ext, ignoring other fields in ortb2Imp. This is how the post-bid creative would be specified:

```
    var adUnits = [
           {
               code: 'test-div',
                mediaTypes: {
                    banner: {
                        sizes: [[728,90]]
                    }
               },
               ortb2Imp: {
                 banner: {
                   api: [7]
                 }
               },
               ...
```

And for the record, the other parts of OMID signaling are already supported by the pbsBidAdapter with the global ortb2 object:
```
pbjs.setConfig({
        ortb2: {
          source: {
            ext: {
              "omidpn": "MyIntegrationPartner",
              "omidpv": "1.0"
            }
          }
        }
});
```

## Verification

The test case would be to set AdUnit.ortb2Imp.banner.api:7 and verify that the request to PBS contains that value in imp[].banner.api.

bretg:
This was discussed in the PBJS committee and approved for development.

bretg:
This would also support the `skadn` parameter as described in https://github.com/prebid/prebid-server/issues/1654

PR #8040, opened 2022-02-09
prebid Server Bid Adapter: added ortb2Imp.instl support

TheMediaGrid:
## Type of change

- [x] Feature

## Description of change

For testing instl field in s2s go adapters, we need to update Server-to-Server Testing module to make possible to sent instl field


bretg (commented):
IMO, this PR is not needed when https://github.com/prebid/Prebid.js/issues/7831 is imminent.

patmmccann:
@TheMediaGrid thanks for your contribution! Could you review and comment on #8062 

Will this solve your need?

PR #8043, opened 2022-02-09
Prebid Core: Adding support for a global return of consent metadata. 

decaffeinatedio:
## Type of change

- [ ] Bugfix
- [x] Feature
- [ ] New bidder adapter   
- [ ] Build related changes
- [ ] CI related changes
- [ ] Other

## Description of change

This includes two significant changes: 
- The introduction of a `getConsentMetadata` function on the Prebid global object.
- Two corresponding functions added in `adapterManager.js` for TCF and USP return. 

With this change, calling `pbjs.getConsentMetadata();` will return a response like: 

```
{
    "coppa": false,
    "gdpr": {
        "apiVersion": 2,
        "consentStringExists": true,
        "gdprApplies": true,
        "timestamp": 1644358143306
    },
    "usp": {
        "timestamp": 1644358143306,
        "usp": "1YYY"
    }
}
```

- Docs PR: https://github.com/prebid/prebid.github.io/pull/3561

## Other information
@bretg This is follow-up to our conversation regarding making consent information available outside of the auction/userId workflow. I made an effort to remove granular information like the full TCF string and instead focused on non-personal consent information that would still be helpful to understand the underlying consent status. Rather than introduce entirely new code, I chose to piggyback off of the well-known `adapterManager.js` interface as was done in other modules likes `bidViewability.js`. There is test coverage for both the USP and TCF cases.

### DIFF
--- a/src/adapterManager.js
+++ b/src/adapterManager.js
@@ -182,10 +182,10 @@ export let gdprDataHandler = {
     return gdprDataHandler.consentData;
   },
   getConsentMeta: function() {
-    if (gdprDataHandler.consentData && gdprDataHandler.consentData.vendorData && gdprDataHandler.consentData.vendorData.tcString && gdprDataHandler.generatedTime) {
+    if (gdprDataHandler.consentData && gdprDataHandler.consentData.vendorData && gdprDataHandler.generatedTime) {
       return {
         gdprApplies: gdprDataHandler.consentData.gdprApplies,
-        consentStringSize: gdprDataHandler.consentData.vendorData.tcString.length,
+        consentStringSize: (isStr(gdprDataHandler.consentData.vendorData.tcString)) ? gdprDataHandler.consentData.vendorData.tcString.length : 0,
         generatedAt: gdprDataHandler.generatedTime,
         apiVersion: gdprDataHandler.consentData.apiVersion
       };
--- a/src/adapterManager.js
+++ b/src/adapterManager.js
@@ -173,21 +173,42 @@ function getAdUnitCopyForClientAdapters(adUnits) {
 
 export let gdprDataHandler = {
   consentData: null,
-  setConsentData: function(consentInfo) {
+  generatedTime: null,
+  setConsentData: function(consentInfo, time = timestamp()) {
     gdprDataHandler.consentData = consentInfo;
+    gdprDataHandler.generatedTime = time;
   },
   getConsentData: function() {
     return gdprDataHandler.consentData;
+  },
+  getConsentMeta: function() {
+    if (gdprDataHandler.consentData && gdprDataHandler.consentData.vendorData && gdprDataHandler.generatedTime) {
+      return {
+        gdprApplies: gdprDataHandler.consentData.gdprApplies,
+        consentStringSize: gdprDataHandler.consentData.vendorData.tcString.length,
+        timestamp: gdprDataHandler.generatedTime,
+        apiVersion: gdprDataHandler.consentData.apiVersion
+      };
+    }
   }
 };
 
 export let uspDataHandler = {
   consentData: null,
-  setConsentData: function(consentInfo) {
+  generatedTime: null,
+  setConsentData: function(consentInfo, time = timestamp()) {
     uspDataHandler.consentData = consentInfo;
+    uspDataHandler.generatedTime = time;
   },
   getConsentData: function() {
     return uspDataHandler.consentData;
+  },
+  getConsentMeta: function() {
+    if (uspDataHandler.consentData && uspDataHandler.generatedTime) {
+      return {
+        usp: uspDataHandler.consentData,
+        timestamp: uspDataHandler.generatedTime}
+    }
   }
 };
 
--- a/src/adapterManager.js
+++ b/src/adapterManager.js
@@ -19,7 +19,8 @@ import {
   logMessage,
   logWarn,
   shuffle,
-  timestamp
+  timestamp,
+  isStr
 } from './utils.js';
 import {getLabels, resolveStatus} from './sizeMapping.js';
 import {decorateAdUnitsWithNativeParams, nativeAdapters} from './native.js';
--- a/src/adapterManager.js
+++ b/src/adapterManager.js
@@ -182,11 +182,11 @@ export let gdprDataHandler = {
     return gdprDataHandler.consentData;
   },
   getConsentMeta: function() {
-    if (gdprDataHandler.consentData && gdprDataHandler.consentData.vendorData && gdprDataHandler.generatedTime) {
+    if (gdprDataHandler.consentData && gdprDataHandler.consentData.vendorData && gdprDataHandler.consentData.vendorData.tcString && gdprDataHandler.generatedTime) {
       return {
         gdprApplies: gdprDataHandler.consentData.gdprApplies,
         consentStringSize: gdprDataHandler.consentData.vendorData.tcString.length,
-        timestamp: gdprDataHandler.generatedTime,
+        generatedAt: gdprDataHandler.generatedTime,
         apiVersion: gdprDataHandler.consentData.apiVersion
       };
     }
@@ -207,7 +207,8 @@ export let uspDataHandler = {
     if (uspDataHandler.consentData && uspDataHandler.generatedTime) {
       return {
         usp: uspDataHandler.consentData,
-        timestamp: uspDataHandler.generatedTime}
+        generatedAt: uspDataHandler.generatedTime
+      }
     }
   }
 };
--- a/test/spec/modules/consentManagement_spec.js
+++ b/test/spec/modules/consentManagement_spec.js
@@ -45,6 +45,13 @@ describe('consentManagement', function () {
         expect(userCMP).to.be.undefined;
         sinon.assert.calledOnce(utils.logWarn);
       });
+
+      it('should not produce any consent metadata', function() {
+        setConsentConfig(undefined)
+        let consentMetadata = gdprDataHandler.getConsentMeta();
+        expect(consentMetadata).to.be.undefined;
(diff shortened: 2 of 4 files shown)


bretg:
Metadata is a good compromise. Thanks @decaffeinatedio .

src/adapterManager.js
```diff
@@ -173,21 +173,42 @@ function getAdUnitCopyForClientAdapters(adUnits) {
...
-  setConsentData: function(consentInfo) {
+  generatedTime: null,
+  setConsentData: function(consentInfo, time = timestamp()) {
     uspDataHandler.consentData = consentInfo;
+    uspDataHandler.generatedTime = time;
   },
   getConsentData: function() {
     return uspDataHandler.consentData;
+  },
+  getConsentMeta: function() {
```
dgirardi, line 206: For the USP case there does not seem to be a distinction between data and metadata - this just adds a timestamp.

dgirardi (commented):
Is there an issue where I can get some more context on the intended use case? Is this related to https://github.com/prebid/Prebid.js/pull/7803#issuecomment-1038946049?

I'd like to understand:

- what exactly is metadata of interest? there's a difference here between what's returned for USP and for GDPR, and the metadata does not let the caller know the difference between "no consent was provided" and "GDPR is not in scope".
- why does this need to be in the publisher API?

decaffeinatedio:
Hi @dgirardi ! I can provide some context here if that works. 

The intended use case is to provide visibility on the consent status (though not necessarily consent content) for vendors who are outside of the core Prebid event waterfall, e.g.: https://app.sincera.io. 

> what exactly is metadata of interest? there's a difference here between what's returned for USP and for GDPR

All things being equal, I'd greedily accept everything. However, in conversation with @bretg, we've chosen to exclude any consent information that can moonlight as personal data, namely the encoded or parsed TCFv2 string. That is why you see the USP go through (not enough entropy to be personal data) but TCF is stripped. 

> and the metadata does not let the caller know the difference between "no consent was provided" and "GDPR is not in scope".

Would `gdprApplies` provide this context? 

Thanks!

dgirardi:
@decaffeinatedio OK, but what does "consent status" mean then? I don't understand the question that you are looking to answer. The timestamp seems to answer "when did the CMP respond?" but not "is there a CMP at all?" (because the metadata is not returned unless there is consent data). What utility has the consent string length for GDPR - could it be a flag like `gpdrStringExists` instead?

decaffeinatedio:
Hey @dgirardi , it boils down to wanting to know: 
- What consent was provided to Prebid?
- When did that occur? 

We _could_ look to include additional information about e.g. was there even a CMP present, even if it didn't provide consent information, but my assumption is that if the caller is interacting with the publisher API to begin with, then they could also make calls directly to any CMP for that information; what Prebid "knows" about consent is what I'm driving at. 

> What utility has the consent string length for GDPR - could it be a flag like gpdrStringExists instead?

Yeah I went back and forth a few times on the boolean versus number. I thought it could potentially be interesting as a heuristic for e.g. number of vendors included but have no concerns with flipping it to a boolean return. Would you like me to make this change?

dgirardi:
OK, thanks; I don't have strong opinions on what should or should not be included in metadata, but I'm not sure that this makes sense in some situations:

- if `getConsentMeta` is called while the CMP popup is open, it just returns null; is the caller expected to figure that out separately?
 - if the user never clicks on "accept cookies", prebid eventually times out (and calls `setConsentData(null)`). Again, `getConsentMeta` would just return null.
 - if the prebid package that is running does not include the relevant gdpr/usp module, `setConsentData` is never called metadata is `null` again.

Does this behavior work for your purpose?

decaffeinatedio:
Hey @dgirardi, yes, I think that should work! 

One potential improvement which isn't super important to me, but might resolve some of the cases you outlined, would be to pass through the[ Ping Status](https://github.com/InteractiveAdvertisingBureau/GDPR-Transparency-and-Consent-Framework/blob/master/TCFv2/IAB%20Tech%20Lab%20-%20CMP%20API%20v2.md#ping-status-codes) and/or Event Status codes in metadata. This would provide more information about e.g. whether the CMP was still open on the user's screen.

test/spec/modules/consentManagement_spec.js
```diff
@@ -667,6 +674,59 @@ describe('consentManagement', function () {
...
+          }, {});
+          let consentMeta = gdprDataHandler.getConsentMeta();
+          sinon.assert.notCalled(utils.logError);
+          expect(consentMeta.consentStringSize).to.be.above(0)
+          expect(consentMeta.gdprApplies).to.be.true;
+          expect(consentMeta.apiVersion).to.equal(2);
+          expect(consentMeta.timestamp).to.be.above(1644367751709);
+        });
+
+        it('performs lookup check and stores consentData for a valid existing user with additional consent', function () {
```
dgirardi, line 704: This looks like a copy of the test case below it?

decaffeinatedio: The setup is the same, but the assertions related to the getConsentMeta() return:

```
          let consentMeta = gdprDataHandler.getConsentMeta();
          sinon.assert.notCalled(utils.logError);
          expect(consentMeta.consentStringSize).to.be.above(0)
          expect(consentMeta.gdprApplies).to.be.true;
          expect(consentMeta.apiVersion).to.equal(2);
          expect(consentMeta.timestamp).to.be.above(1644367751709);
```

vs. 

```
          let consent = gdprDataHandler.getConsentData();
          sinon.assert.notCalled(utils.logError);
          expect(didHookReturn).to.be.true;
          expect(consent.consentString).to.equal(testConsentData.tcString);
          expect(consent.addtlConsent).to.equal(testConsentData.addtlConsent);
          expect(consent.gdprApplies).to.be.true;
          expect(consent.apiVersion).to.equal(2);
```

The checks for `consentStringSize` and `timestamp` are specifically novel.

dgirardi: I may be having a stroke, but I don't see any call to `getConsentMeta` (test case in lines 704-729). `diff` tells me it's identical to the test case from lines 730-754.

decaffeinatedio: Nope, looks like I had the stroke. You're right, that's an unnecessary copy/paste. Let me pull the 704 block out and commit it. Thanks!

test/spec/modules/consentManagement_spec.js
```diff
@@ -727,32 +727,6 @@ describe('consentManagement', function () {
           expect(consent.apiVersion).to.equal(2);
         });
 
-        it('performs lookup check and stores consentData for a valid existing user with additional consent', function () {
```
decaffeinatedio, line 670: @dgirardi This should do the trick 🤞

dgirardi (approved):

patmmccann:
At first glance, it seems difficult to argue enabling scrapers to pull more data is good for publishers. How do you make that case?

patmmccann (approved):

patmmccann:
Thanks for adding those details! Makes sense.

patmmccann:
This still needs a second review; I just wanted to indicate your reasoning LGTM

osazos (commented):
Thanks for this feature.

Some thought regarding the naming of `generatedTime` and `timestamp`. It is not clear what `timestamp` is at the first look and generatedTime may let understood a duration. What do you think to rename both to `generatedAt` (or something else).

Thanks

src/adapterManager.js
```diff
@@ -173,21 +173,42 @@ function getAdUnitCopyForClientAdapters(adUnits) {
...
+    gdprDataHandler.generatedTime = time;
   },
   getConsentData: function() {
     return gdprDataHandler.consentData;
+  },
+  getConsentMeta: function() {
+    if (gdprDataHandler.consentData && gdprDataHandler.consentData.vendorData && gdprDataHandler.generatedTime) {
+      return {
+        gdprApplies: gdprDataHandler.consentData.gdprApplies,
+        consentStringSize: gdprDataHandler.consentData.vendorData.tcString.length,
```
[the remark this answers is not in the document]
decaffeinatedio: Can't be too sure :)

src/adapterManager.js
```diff
@@ -173,21 +173,43 @@ function getAdUnitCopyForClientAdapters(adUnits) {
...
+    gdprDataHandler.generatedTime = time;
   },
   getConsentData: function() {
     return gdprDataHandler.consentData;
+  },
+  getConsentMeta: function() {
+    if (gdprDataHandler.consentData && gdprDataHandler.consentData.vendorData && gdprDataHandler.consentData.vendorData.tcString && gdprDataHandler.generatedTime) {
+      return {
+        gdprApplies: gdprDataHandler.consentData.gdprApplies,
+        consentStringSize: gdprDataHandler.consentData.vendorData.tcString.length,
```
osazos, line 188: I don't know well the behavior related to the setting of `.vendorData.tcString` but I think we don't have to change the main condition. With your change, if the tcString is `""` the function will returns nothing, and it doesn't sound what you want;

In my first comment I thought something like the code below.

```js
if (gdprDataHandler.consentData && gdprDataHandler.consentData.vendorData && gdprDataHandler.generatedTime) {
      return {
        gdprApplies: gdprDataHandler.consentData.gdprApplies,
        consentStringSize: (isStr(gdprDataHandler.consentData.vendorData.tcString)) ? gdprDataHandler.consentData.vendorData.tcString.length : 0,
        …
```

Please let me know

decaffeinatedio: @osazos Agreed, that seems more elegant. Let me update.

osazos:
@decaffeinatedio, The CI has failed. Can you fix it?

decaffeinatedio:
@osazos Sorry about that.

PR #8044, opened 2022-02-10
 orbidder Bid Adapter: add unit test for price floor module

hendrikiseke1979:
## Type of change
- [x] Build related changes

## Description of change
This PR adds a unit test for price floor module for orbidderBidAdapter


test/spec/modules/orbidderBidAdapter_spec.js
```diff
@@ -283,6 +283,27 @@ describe('orbidderBidAdapter', () => {
...
+    };
+
+    bidRequest.floors = {
+      currency: 'EUR',
+      values: {
+        'banner|640x480': 15.07
+      }
+    };
+    const request = buildRequest(bidRequest);
+    expect(request.data.params.bidfloor).to.equal(15.07);
```
patmmccann, line 304: why not an expect on the currency?

hendrikiseke1979: So far, we don't send the currency to the ssp, so we cannot expect in the data object.

patmmccann (approved):

PR #8062, opened 2022-02-14
Preebid Server: add support for adunit.ortb2Imp

ChrisHuie:
For issue #7831


modules/prebidServerBidAdapter/index.js
```diff
@@ -670,9 +670,10 @@ const OPEN_RTB_PROTOCOL = {
       }, {...deepAccess(adUnit, 'ortb2Imp.ext')});
 
       const imp = { id: impressionId, ext, secure: s2sConfig.secure };
-
       const ortb2 = {...deepAccess(adUnit, 'ortb2Imp.ext.data')};
```
bretg, line 673: We just need to copy all the contents of ortb2Imp into the imp.

PR #8071, opened 2022-02-15
Prebid core: add support for asynchronous access to consent data

dgirardi:
## Type of change

- [x] Feature

## Description of change

﻿This adds `gpdrDataHandler.promise` and `uspDataHandler.promise`, to enable access to USP/GDPR consent data from outside of an auction context (see use case: https://github.com/prebid/Prebid.js/pull/7803)


Fawke (commented):
Looks good, just one question.

Would there be any way for the caller of `gpdrDataHandler.promise` to find out if the consent module isn't loaded in the build or if it was loaded but couldn't obtain the consent data? In both cases it's returning `null`.

I don't know if that matters or not, but I'm just pointing it out.

src/consentHandler.js
```diff
@@ -0,0 +1,68 @@
...
+  get ready() {
+    return this.#ready;
+  }
+
+  /**
+   * @returns a promise than resolves to the consent data, or null if no consent data is available
+   */
+  get promise() {
+    if (!this.#enabled) {
+      this.#resolve(null);
```
Fawke, line 56: Will rejecting the promise suit better here?

dgirardi: I don't think so (but I'm not that sure). My reasoning is that unless you make it so that `data` is *never* `null` when the promise resolves, it's just asking the caller to add some complexity, because they will have to check for `null` and treat it as an error anyways. I am not against making every `null` (including, for example, USP timeouts) a rejection, but it also makes sense to me that you would want to have exactly what `getConsentData` returns, just asynchronously.

If someone is interested in whether the module is installed, they could look at `.enabled`.

Fawke (approved):

msm0504 (approved):

Issue #8032, opened 2022-02-09
Prebid server adapter incompatible with sizeMappingV2?


WalterT:
## Type of issue
Bug

## Description
The Prebid server adapter seems to be incompatible with the advanced size mapping module. Whenever ad units have a mediaTypes.banner.sizeConfig element rather than mediaTypes.banner.sizes, this error is raised when calling bids: 

_TypeError: Reduce of empty array with no initial value
    at Array.reduce (<anonymous>)
    at G.e.callBids (prebid-6.2.0.js:27)
    at prebid-6.2.0.js:9
    at Array.forEach (<anonymous>)
    at Object.w.callBids (prebid-6.2.0.js:9)
    at Object.run (prebid-6.2.0.js:9)
    at d (prebid-6.2.0.js:9)
    at Object.$ (prebid-6.2.0.js:9)
    at Object.<anonymous> (prebid-6.2.0.js:9)
    at Object.callBids (prebid-6.2.0.js:9)_

 referring to this function in the Prebid server adapter: https://github.com/prebid/Prebid.js/blob/master/modules/prebidServerBidAdapter/index.js#:~:text=out%20bid%20requests%20*/-,baseAdapter.callBids,-%3D%20function(s2sBidRequest%2C%20bidRequests

All validAdUnits seem to be filtered out because of that.

Am I analyzing this correctly?

## Steps to reproduce

- Build Prebid.js with prebidServerBidAdapter & sizeMappingV2 modules
- Configure one ad unit as follows:
```
  var adUnits = [
    {
      code: 'test1',
      mediaTypes: {
        banner: {
          sizeConfig: [
          { minViewPort: [0, 0], sizes: [[300, 250], [300, 100]] },
          { minViewPort: [768, 0], sizes: [[728, 90], [300, 250]] }
        ]
        }
      },
      bids: [{
        bidder: 'appnexus',
        params: {
          placementId: 13144370
        }
      }]
    }
  ];
```
- Add an s2s config:
pbjs.setConfig({
    s2sConfig: [{
        accountId: '10',
        bidders: ['appnexus'],
        adapter: 'prebidServer',
        enabled: true,
        endpoint: 'https://prebid.adnxs.com/pbs/v1/openrtb2/auction',
        syncEndpoint: 'https://prebid.adnxs.com/pbs/v1/cookie_sync',
        timeout: 500
    }]
})

## Platform details
Prebid.js 6.2

bretg:
[...]

PR #8084, opened 2022-02-17
PBS adapter: partial support of sizeMapping (and sizeMappingV2)

dgirardi:
## Type of change

- [x] Bugfix

## Description of change

This refactors the logic around labels and size mapping to work on the adUnit instead of individual bid requests, and allows the PBS adapter to respect size mapping config as long as it doesn't result in bidder-specific mediaTypes.

Partially fixes https://github.com/prebid/Prebid.js/issues/8032

## Other information

This is an example test page using sizeMappingV2 through the PBS adapter:

```
<html>
<head>
    <script async src="../build/dev/prebid.js"></script>
    <script async src="https://www.googletagservices.com/tag/js/gpt.js"></script>
    <script>
        var FAILSAFE_TIMEOUT = 10000;
        var PREBID_TIMEOUT = 10000;

        var adUnits = [
            {
                mediaTypes: {
                    banner: {
                        sizeConfig: [
                            {minViewPort: [0, 0], sizes: [[300, 250], [300, 100]]},
                            {minViewPort: [768, 0], sizes: [[728, 90], [300, 250]]}
                        ]
                    }
                },
                bids: [{
                    bidder: 'appnexus',
                    params: {
                        placementId: 13144370
                    },
                }, {
                    bidder: 'rubicon',
                    params: {
                        accountId: 14062,
                        siteId: 70608,
                        zoneId: 498816
                    },
                },/*{
                    bidder: 'adplus',
                    params: {
                        inventoryId: '-1',
                        adUnitId: '-3',
                    },
                }*/

                ]
            }
        ];

        var pbjs = pbjs || {};
        pbjs.que = pbjs.que || [];

    </script>

    <script>
        });

        pbjs.que.push(function() {
            pbjs.setConfig({
                debug: true,
                bidderTimeout: 10000,
                s2sConfig: {
                    accountId: '1',
                    enabled: true, //default value set to false
                    defaultVendor: 'appnexus',
                    bidders: ['appnexus', 'rubicon'],
                    timeout: 10000, //default value is 1000
                    adapter: 'prebidServer', //if we have any other s2s adapter, default value is s2s
                },
            });
            pbjs.addAdUnits(adUnits);

            pbjs.requestBids({
                bidsBackHandler: sendAdserverRequest,
                timeout: PREBID_TIMEOUT
            });
        });

                pbjs.que.push(function() {
                    googletag.pubads().refresh();
                });
            });
        }

        setTimeout(function() {
            sendAdserverRequest();
        }, FAILSAFE_TIMEOUT);

    </script>

    <script>

            googletag.enableServices();
        });
    </script>
</head>

<body>
<h2>Prebid.js Test</h2>
<h5>Div-1</h5>
    </script>
</div>
</body>
</html>
```


modules/prebidServerBidAdapter/index.js
```diff
@@ -517,6 +517,14 @@ Object.assign(ORTB2.prototype, {
...
     let impIds = new Set();
     adUnits.forEach(adUnit => {
+      // TODO: support labels / conditional bids
+      // for now, just warn about them
+      adUnit.bids.forEach((bid) => {
+        if (bid.mediaTypes != null) {
+          logWarn(`Prebid Server adapter does not (yet) support bidder-specific mediaTypes for the same adUnit. Size mapping configuration will be ignored for adUnit: ${adUnit.code}, bidder: ${bid.bidder}`);
+        }
+      })
+
```
Fawke, line 527: Can you please provided an example Ad unit setup where we use bidder-specific mediaTypes? I don't understand the meaning of this warning message.

Fawke:
Rest looks good. 

I tested out the sizeMappingV1, and sizeMappingV2 with some example test pages I had, and both seem to work. And also, it's working with Prebid Server.

Do we need to update the docs and add show an example of sizeMapping usage with Prebid Server. 

Running the sizeMapping logic on an Ad unit basis, instead of on each Bidder request simplifies things a lot in sizeMappingV2 and gets rid of a lot complexity.

If you can just add a few unit tests for the function you added in `sizeMapping.js` file, that'll be great.

dgirardi:
- bidder-specific mediaTypes: the last sizeMappingV2 test case shows how that works:
   https://github.com/dgirardi/Prebid.js/blob/e8c9f626ca91d081ac6b42eae7dec246ac730c39/test/spec/modules/sizeMappingV2_spec.js#L1480-L1489

   with the example in the OP, something like:
    ```
    mediaTypes: {
        banner: {
            sizeConfig: [
                {minViewPort: [0, 0], sizes: [[300, 250], [300, 100]]},
                {minViewPort: [768, 0], sizes: [[728, 90], [300, 250]]}
            ]
        },
        video: {
            context: 'instream',
            playerSize: [640, 480],
            mimes: ['video/mp4'],
            protocols: [1, 2, 3, 4, 5, 6, 7, 8],
            playbackmethod: [2],
            skip: 1
        }
    },
    bids: [{
        bidder: 'appnexus',
        params: {
            placementId: 13144370
        },
        sizeConfig: [{minViewPort: [0, 0], relevantMediaTypes: ['banner']}]

    }, {
        bidder: 'rubicon',
        params: {
            accountId: 14062,
            siteId: 70608,
            zoneId: 498816
        },
        sizeConfig: [{minViewPort: [0, 0], relevantMediaTypes: ['video']}]
    }]
    ```
   as far as I understand this is not possible with sizemappingV1 (but please prove me wrong if you can).

- documentation: IMO the only thing we could/should say is that support is incomplete (because the above does not work properly), there's nothing special about using sizeMapping in client or server adapters otherwise. Technically it's a bug that we are planning to eventually fix though so I'm not sure how we want to document it, I'll discuss and come back.
- sizeMappingV1 tests are (for some reason) written into the adapterManager tests: https://github.com/prebid/Prebid.js/blob/c13ed679f34823c7a401f810ed1a158b6b125dd2/test/spec/unit/core/adapterManager_spec.js#L1736-L2310
   Because they are one level of abstraction higher, they did not need to be modified (as they don't directly call the new signatures), but they are still valid. The only thing that's missing is a test similar to the one I added for v2 where the new logic populates `adUnit.bids[].mediaTypes`, but I could not find any way to write v1 configuration that would result in that.

Fawke:
@dgirardi You are correct, sizeMappingV1, built into the core doesn't support bidder level sizeConfigs.

But I don't understand why do want to build that feature into the core sizeMapping? If the publishers need that, why can't they use sizeMappingV2?

I've tested bidder level sizeConfigs and it works with PBS + sizeMappingV2. I tried this example below and it works. 

```(javascript)
	var adUnits = [
			{
				mediaTypes: {
					banner: {
						sizeConfig: [
							{ minViewPort: [0, 0], sizes: [[300, 250], [300, 100]] },
							{ minViewPort: [768, 0], sizes: [[728, 90], [300, 250]] }
						]
					}
				},
				bids: [{
					bidder: 'appnexus',
					params: {
						placementId: 13144370
					},
					sizeConfig: [
						{ minViewPort: [0, 0], relevantMediaTypes: ['none'] },
						{ minViewPort: [850, 0], relevantMediaTypes: ['none'] }
					]
				}, {
					bidder: 'rubicon',
					params: {
						accountId: 14062,
						siteId: 70608,
						zoneId: 498816
					},
				},

				]
			}
		];
```

dgirardi:
@Fawke I am not looking to build any new feature; but, the code was written in the way "for each bidder, get its mediaTypes according to sizeMapping logic", so I was not sure if I was missing some way where they would be different. If they cannot, I don't think we need any new test cases for v1 (as that's where the new logic is different). Are you asking to refactor the existing ones to work at the "right" level?

modules/prebidServerBidAdapter/index.js
bretg: @Fawke - this is the 'relevantMediaTypes feature described in https://docs.prebid.org/dev-docs/modules/sizeMappingV2.html

bretg:
docs PR https://github.com/prebid/prebid.github.io/pull/3605

Fawke (approved):

ChrisHuie (approved):

PR #8086, opened 2022-02-18
Prebid Core: switch native assets to ortb2 format

### LINKED ISSUE #7830: Prebid Native: switch to openRTB for native

musikele:
**NOTE**: this is a continuation of PR #7847. Original submitter ceased working on adtech and I am continuing his work. In order to handle change requests and future work on this, I decided (with the consent of @bretg) to move in a new PR. Original thread of comments and reviews is in the old one. 

## Type of change

- [x] Feature

## Description of change
Change from prebid 'standard' of defining native assets to [openRTB 1.2 standard](https://www.iab.com/wp-content/uploads/2018/03/OpenRTB-Native-Ads-Specification-Final-1.2.pdf).  
Instead of `sponsoredBy`, `image`, `icon`, we would have openRTB equivalents:
- `{ data: { type: 1 }}`
- `{ img: { type: 3 }}`
- `{ img: { type: 1 }}`

This will help standardize prebid native. Also, a lot of bidders are sending ortb2 requests on their back end, so at the moment, they are converting bid request from prebid to openRTB standard and later they are converting bid responses from openRTB to prebid standard. With this approach a lot of duplicated code between the bidders can be removed.

## Backwards compatibility
When this PR is merged, we will still have backwards compatibility. Which mean, nothing will break for the existing native bidders. That will give us some time until all bidders adjust their code

## Other information
This PR resolves #7830 issue.


musikele:
My latest commits are the following of this comment by @bretg on the old PR #7847:

Ok, discussed in today's PBJS meeting. We propose handling this native change in a way similar to how we handled the first party data change last year:

1. Add some transition code (e.g. toOrtbNative) that maps old-style AdUnit native definitions to new-style.
2. Create a transition function (e.g. fromOrtbNative) that bid adapters can use to translate the new-style native definition back to the old-style. This is a lossy transition and would only support whatever fields are currently covered by the proprietary format.
3. Update all 74 native-capable bid adapters to call fromOrtbNative() at the top of buildRequests. Then we communicate with them to start the process of converting their adapters to not need fromOrtbNative().

FWIW, I poked around a couple of adapters: one had a complicated bunch of code for mapping native params and the other seemed to wrap it all up and parse server-side. So this transition may take a while.

Documentation-wise, the proposal is:

- produce a new document that is the recommended way that new native implementations should be 
- leave the current doc in place but warn that new implementations should be done in the ortb-way
- hide the old-old doc away, making it more clear that no one should be using that (sendId) approach anymore.

musikele:
Prebid Universal Creative PR: https://github.com/prebid/prebid-universal-creative/pull/150

test/spec/unit/secureCreatives_spec.js
```diff
@@ -457,10 +458,13 @@ describe('secureCreatives', () => {
...
         sinon.assert.calledOnce(stubFireNativeTrackers);
-        sinon.assert.calledWith(stubEmit, CONSTANTS.EVENTS.BID_WON, adResponse);
-        sinon.assert.calledOnce(spyAddWinningBid);
+        sinon.assert.neverCalledWith(stubEmit, CONSTANTS.EVENTS.BID_WON);
+        sinon.assert.notCalled(spyAddWinningBid);
 
         expect(adResponse).to.have.property('status', CONSTANTS.BID_STATUS.RENDERED);
+
+        // revert back to 'adId = 1'
+        adResponse.adId = bidId;
```
dgirardi, line 467: This should be in some setup / teardown section, or the next guy will also have to curse for however long it took you to track it down :)

src/native.js
```diff
@@ -80,12 +183,18 @@ export const hasNonNativeBidder = adUnit =>
  * @return {Boolean} If object is valid
  */
 export function nativeBidIsValid(bid, {index = auctionManager.index} = {}) {
+  const bidRequest = index.getAdUnit(bid);
+  if (!bidRequest) { return false; }
+
+  if (deepAccess(bid, 'native.ortb') && deepAccess(bidRequest, 'nativeParams.ortb')) {
```
dgirardi, line 189: This seems to enforce that an ortb response is valid only if the request was also in ortb. But during the transition period I imagine we'll have to deal with a lot of "legacy" requests, which would force adapters to individually have logic like "if the request is legacy, respond in legacy; otherwise respond in ortb". Is it possible to decouple them? maybe by falling back to the conversion logic if the request does not have `nativeParams.ortb`.

If I understand this right with this PR no adapter is hitting this code path. Could/should the PBS adapter respond in ortb?

Also a minor point, I don't like the name `bidRequest` for an ad unit.

src/native.js
```diff
@@ -230,9 +356,16 @@ export function getAllAssetsMessage(data, adObject) {
   const message = {
     message: 'assetResponse',
     adId: data.adId,
-    assets: []
   };
 
+  if (adObject.native.ortb) {
+    Object.keys(adObject.native).forEach(key => {
+      message[key] = adObject.native[key];
```
dgirardi, line 363: Is this right? From my reading of the PUC PR it looks like it expects these under `message.ortb` ?

dgirardi: Actually disregard this, of course `ortb` is also being copied over.

test/spec/unit/pbjs_api_spec.js
```diff
@@ -2166,14 +2166,47 @@ describe('Unit: Prebid Module', function () {
...
         const nativeRequest = spyArgs.args[1][0].bids[0].nativeParams;
-        expect(nativeRequest).to.deep.equal({
-          image: {required: true},
-          title: {required: true},
-          sponsoredBy: {required: true},
-          clickUrl: {required: true},
-          body: {required: false},
-          icon: {required: false},
-        });
+        expect(nativeRequest.ortb.assets).to.deep.equal([
```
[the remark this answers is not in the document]
musikele: I don't think they should be mutually exclusive. Thousands of publishers already have native ad units set up in the old way, but in the future bidders will start to switch to ortb2, so for some time the same ad unit will contain both types.  

Then, it's the bidder's responsibility to decide which version they want to use. I think that ortb will be the preferred format since it's more expressive, but waiting for bidders to migrate can take ages. That's why we've also given the two functions to convert to and from legacy prebid native format. 

 @bretg wdyt ?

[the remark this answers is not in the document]
dgirardi: But a publisher is expected to specify either one or the other, right? otherwise it's not clear how you'd "merge" the definitions, or what to do in case they conflict. That's what I meant by validation.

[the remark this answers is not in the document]
musikele: I think publishers may specify both types in the same ad unit. This shouldn't be seen as an error. It's the single adapter that will have to adapt its code to both formats. 

Right now publishers are using only the proprietary format. When this PR is merged, the vast majority of these publishers will continue to use the proprietary format. Then, a bunch of native bid adapters will start to support the ortb format and will ask publishers to bid in this new format. So we'll have the situation where, for the same ad unit, some native bidders will understand the old way _only_, others will understand both ways, and some others will understand the new way only. It's up to them to decide how to interpret the assets object. 

If I were a bidder, i'd implement this logic: 
```
let ortb; 
if (nativeType.native && nativeType.native.ortb) {
   // use only ortb
   {ortb} = nativeType.native; 
} else {
   // publisher is using old proprietary format, it's up to them to 
   // decide wether use it straight or convert to ortb with function 
   // `toOrtbNative()` 
   ortb = toOrtbNative(nativeType.native)
}
```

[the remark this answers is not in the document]
dgirardi: Wouldn't it be better for a publisher to switch over to ortb, and let the translation shim you already added in here handle the adapters that have not yet been ported? With the goal of having every bid adapter talk only ortb at some point in the future.  Having to maintain two versions would add a lot of friction and I don't see the need for it.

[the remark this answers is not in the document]
dgirardi: To clarify, I see the transition as:

- adapters initially understand the old way only (and this PR makes sure they get it regardless of how the adunit is defined)
- gradually, adapters will understand the new way only (and they can translate old definitions as you show above)

But I don't think publishers or adapters should be interested in mixing or matching the formats; it's going to invite inconsistencies otherwise.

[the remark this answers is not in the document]
musikele: @dgirardi added the code to check if ortb is in use, no legacy native keys should be used; this has been done [here](https://github.com/prebid/Prebid.js/pull/8086/files#diff-f2f8b0820ed4ec1bd15383b439def39eb86117e358d9a14eaf5dc1ad73439dcaR132) and it's called by `checkAdUnitSetup` in both prebid and sizeMappingV2

test/spec/unit/secureCreatives_spec.js
```diff
@@ -472,6 +476,7 @@ describe('secureCreatives', () => {
         describe(`for ${test} bids`, () => {
           beforeEach(() => {
             prepBid(adResponse);
+            // throw JSON.stringify(adResponse);
```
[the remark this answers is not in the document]
musikele: solved.

test/spec/unit/secureCreatives_spec.js
```diff
@@ -332,8 +332,8 @@ describe('secureCreatives', () => {
         sinon.assert.calledWith(stubGetAllAssetsMessage, data, adResponse);
         sinon.assert.calledOnce(ev.source.postMessage);
         sinon.assert.notCalled(stubFireNativeTrackers);
-        sinon.assert.neverCalledWith(stubEmit, CONSTANTS.EVENTS.BID_WON);
-        sinon.assert.notCalled(spyAddWinningBid);
+        sinon.assert.calledWith(stubEmit, CONSTANTS.EVENTS.BID_WON, adResponse);
```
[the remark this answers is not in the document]
musikele: added new test here: https://github.com/prebid/Prebid.js/pull/8086/files#diff-216136cd3abd45d8370399ff0c8cec9a603ad18597008c38c0d54e503eafe7c3R340

src/native.js
musikele: 1. if `ortb` keyword is not present in `nativeParams`, prebid will assume that the request was made using the legacy format. What do you mean by "decouple" the two request types?  Do you mean to convert the request to OpenRTB and then validate the response against the just-created openRTB request? 

2. With this PR we will adapt all adapters to translate ortb to legacy format request. PBS will understand both types, checkout prebidServerBidAdapter/index.js . 

3. I will change bidRequest to adUnit.

dgirardi: 1 - yes, because as adapters transition, they will ideally want to deal only with ORTB - so for a legacy request they will convert to ortb and reply with ortb. This check would now be skipped if the request is legacy.
2. But the PBS adapter is still replying with legacy responses. Should it be converted over to ortb completely? (should it even deal with legacy requests, beside converting them?)

musikele: 1. All responses are converted to Ortb. 
2. Prebid Server was already responding in Ortb. I only had to remove the code that converted to legacy.

src/constants.json
```diff
@@ -118,5 +118,37 @@
...
+    "rating": "rating",
+    "address": "address",
+    "downloads": "downloads",
+    "likes": "likes",
+    "phone": "phone",
+    "price": "price",
+    "salePrice": "saleprice",
+    "displayUrl": "displayurl"
+  },  
+  "ASSET_TYPES": {
```
dgirardi, line 136: I'd put NATIVE somewhere in this name, but maybe it's clearer than it looks to me

src/secureCreatives.js
```diff
@@ -108,6 +109,13 @@ function handleNativeRequest(reply, data, adObject) {
     logError(`Cannot find ad '${data.adId}' for x-origin event request`);
     return;
   }
+
+  if (!WON_AD_IDS.has(adObject.adId)) {
```
dgirardi, line 113: `WON_AD_IDS` could be a  WeakMap using adObject as key - otherwise it's unbounded. Not a big deal as I don't think we ever throw `adObject` to the garbage collector, but one day we might.

modules/prebidServerBidAdapter/index.js
```diff
@@ -561,8 +555,8 @@ Object.assign(ORTB2.prototype, {
       this.adUnitsByImp[impressionId] = adUnit;
 
       const nativeParams = adUnit.nativeParams;
-      let nativeAssets;
-      if (nativeParams) {
+      let nativeAssets = nativeAssetCache[impressionId] = deepAccess(nativeParams, 'ortb.assets');
```
dgirardi, line 558: GH does not let me pick the right line for this - further down here, line 681, the code path for native packages only `ortb.assets` into an ORTB native request. Does that still make sense, or should it just say `request.native = nativeParams.ortb`? if no, why not? maybe explained in the form of a unit test? :)

src/secureCreatives.js
musikele: With weakmaps, i'd have to change line 114 to `WON_AD_IDS.set(adObject, adObject)` and honestly i don't like the repetition :) 

Since the complexity is still sublinear (as [written in the ecmaspec](https://262.ecma-international.org/6.0/#sec-weakset-objects)), i preferred to use WeakSet, because it maps more naturally to what we're doing here. What do you think? the only major problem was compatibility with IE, but Prebid already dropped it in version 6.

modules/prebidServerBidAdapter/index.js
musikele: You are right! Before, there was no way for publishers to pass context, plcmttype and other fields into a media type object. With my latest commit the mediaType.native.ortb will contain exactly that, so if that's present, i'm going to pass that to the request, otherwise we go the old way.

modules/prebidServerBidAdapter/index.js
```diff
@@ -685,23 +679,24 @@ Object.assign(ORTB2.prototype, {
...
         try {
           mediaTypes['native'] = {
-            request: JSON.stringify({
+            request: ortbRequest ? JSON.stringify(ortbRequest) : JSON.stringify({
               // TODO: determine best way to pass these and if we allow defaults
               context: 1,
               plcmttype: 1,
               eventtrackers: [
-                {event: 1, methods: [1]}
+                { event: 1, methods: [1] }
```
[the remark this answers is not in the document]
musikele: As discussed yesterday, I've left `{event: 1, methods: [1]}` because it's a more conservative choice, and also pixels are generally more accepted than the JS equivalent.

modules/appnexusBidAdapter.js
```diff
@@ -123,6 +124,9 @@ export const spec = {
    * @return ServerRequest Info describing the request to the server.
    */
   buildRequests: function (bidRequests, bidderRequest) {
+    // convert Native ORTB definition to old-style prebid native definition
+    bidRequests = convertOrtbRequestToProprietaryNative(bidRequests);
```
robertrmartinez, line 128: This looks like it is overwriting the incoming bidRequests. Should it be just creating a copy instead here? Not sure is this affects downstream prebid core if these input bidRequests object are changed?

For example, if the input has rendererUrl / sendTargetingKeys: false

Then this function overwrites them and they are removed, does the core still handle the bidResponse correctly when it maps back to ortb2 etc?

robertrmartinez: Looks like the function changes the incoming object anyways, so prob don't need to set it equal to it if it changes it anyways?

Maybe a copy should be returned so the original `bidRequests` object is not changed?

musikele: Hi @robertrmartinez 
You are right, it is overwriting the incoming BidRequests. I tried to create a copy and pass it around, and this would be more correct (from a functional standpoint) and may lead to less bugs, however it is unpractical. 

Some bidders in their tests rely on the fact that the `bidRequest` object they sent is exactly the one that will come back; some check the reference, other check some properties that they are attaching to the object; in the end, creating a copy of `bidRequests` instead of modifying it in place would mean to fix all their tests. 

Regarding the fact that rendererUrl/sendTargetingKeys may disappear: I'll do a change in the code so this will not happen. 

To summarize, my idea is that we *should* create a copy, but this is impossible now, because it will break all the tests that assume that the bidRequest they're testing is the same they sent in input.

musikele: @robertrmartinez check the code now!

musikele:
rebased

modules/appnexusBidAdapter.js
robertrmartinez: That all makes sense to me!

Hopefully this is a short term solution and ORTB gets fully adopted and legacy native mediatype is deprecated!

dgirardi: @musikele would a compromise of "copy-on-write" work? since tests should not be setting up native ortb2 right now, if you return the original object when no conversion is necessary, but clone it when it is, I think it might work?

modules/prebidServerBidAdapter/index.js
```diff
@@ -1045,19 +1041,19 @@ Object.assign(ORTB2.prototype, {
...
+            let ortb;
             if (typeof bid.adm === 'string') {
-              adm = bidObject.adm = JSON.parse(bid.adm);
+              ortb = bidObject.adm = JSON.parse(bid.adm);
             } else {
-              adm = bidObject.adm = bid.adm;
+              ortb = bidObject.adm = bid.adm;
             }
 
             let trackers = {
```
dgirardi, line 1051: Noticed this now (trying to mock a response for the PUC) - `trackers` does not seem to be used for anything?

musikele: Correct, it's a leftover from the previous logic. What I decided to do here is to convert the two soon-deprecated properties `imptrackers` and `jstracker` into eventtrackers.

src/native.js
```diff
@@ -22,16 +67,45 @@ const SUPPORTED_TYPES = {
...
+    7: 'salePrice',
+    8: 'phone',
+    9: 'address',
+    10: 'body2',
+    11: 'displayUrl',
+    12: 'cta',
+  }
+}
+
+export const nativeMapper = new Map();
```
dgirardi, line 95: This should be a WeakMap (using requests as key).

src/native.js
```diff
@@ -229,9 +364,31 @@ export function getAllAssetsMessage(data, adObject) {
   const message = {
     message: 'assetResponse',
     adId: data.adId,
-    assets: []
   };
 
+  // Provide a mapping from legacy asset to openRTB id, so legacy creatives can still display ads
```
dgirardi, line 369: Could this be done by providing assets in the legacy format directly? If it can be done in a way that's backwards compatible with current PUC behavior, it should make rollout significantly easier.

src/native.js
```diff
@@ -280,3 +411,299 @@ function getNativeKeys(adUnit) {
...
+
+/**
+ * Converts an OpenRTB request to a proprietary Prebid.js format.
+ * The proprietary Prebid format has many limitations and will be dropped in
+ * the future; adapters are encouraged to stop using it in favour of OpenRTB format.
+ * @param {BidRequest[]} bidRequests an array of valid bid requests
+ * @returns an array of valid bid requests where the openRTB bids are converted to proprietary format.
+ */
+export function convertOrtbRequestToProprietaryNative(bidRequests) {
+  if (!bidRequests || !isArray(bidRequests)) return bidRequests;
```
dgirardi, line 568: because a bunch of adapters are calling this now, the body of this should be guarded with `if(FEATURES.NATIVE) { ... }`

src/native.js
dgirardi: You could, then use `adapterManager.index`, as to not introduce another memory leak.

src/native.js
```diff
@@ -229,9 +345,24 @@ export function getAllAssetsMessage(data, adObject) {
   const message = {
     message: 'assetResponse',
     adId: data.adId,
-    assets: []
   };
 
+  // Pass to Prebid Universal Creative all assets, the legacy ones + the ortb ones (under ortb property)
+  const ortbRequest = nativeMapper.get(adObject.requestId);
```
dgirardi, line 351: I'm confused by this, could you add a comment outlining the message format and how you are updating it?  why does this need to update the bid response (`adObject`) ? If that is necessary I don't think it should be done here.

bretg:
As discussed in https://github.com/prebid/Prebid.js/issues/8665, there are PBS BidAdapter changes needed here. It currently creates imp.native from a hard-coded template. Believe it needs to be updated to skip that template if the ortb.native block is specified in the page.

src/native.js
dgirardi: adObject is a bid response; what doesn't seem right is that you need to request a message to update the bid response. 
If it's correct to do the last step (`bidResponse.native = ....`), then it should be done around the time it's returned by the adapter. If what matters is the message, you should be able to populate `message.assets` & `message.ortb` without modifying the bid response.

Which one it is depends on whether the legacy format matters anywhere but these messages.

dgirardi (approved):

src/native.js
```diff
@@ -280,3 +414,309 @@ function getNativeKeys(adUnit) {
...
+      legacyResponse[requestAsset.img.type === NATIVE_IMAGE_TYPES.MAIN ? 'image' : 'icon'] = asset.img.url;
+    } else if (asset.data) {
+      legacyResponse[PREBID_NATIVE_DATA_KEYS_TO_ORTB_INVERSE[NATIVE_ASSET_TYPES_INVERSE[requestAsset.data.type]]] = asset.data.value;
+    }
+  }
+  return legacyResponse;
+}
+
+/**
+ * Converts an OpenRTB request to a proprietary Prebid.js format.
```
jsnellbaker, line 674: Is the description here accurate for this function?  Per the function name, I would have thought description would have been the other way around.

src/secureCreatives.js
```diff
@@ -190,7 +193,7 @@ function resizeRemoteCreative({ adId, adUnitCode, width, height }) {
     let element = getElementByAdUnit(elmType + ':not([style*="display: none"])');
     if (element) {
       let elementStyle = element.style;
-      elementStyle.width = width + 'px';
+      elementStyle.width = width ? width + 'px' : '100%';
```
jsnellbaker, line 196: Out of curiosity - would there be scenarios where width is undefined but height is still there?  Is there a reason to not have a backup value for height as well?

src/native.js
musikele: As we settled in chat: we decided to leave the nativeMapper as is. it would take a lot of work and we'd have to change some core files in order to remove it. The goal was to not introduce memory leaks, and we temporarly "solved" by deleting native ads that are used or expired (see `removeExpiredBidsFromNativeMapper()`).

modules/appnexusBidAdapter.js
musikele: @dgirardi I've implemented copy-on-write and tests are not failing.

src/secureCreatives.js
musikele: The scenario actually existed, it was into prebid-universal creative. I am linking one of the lines that was doing exactly that, setting the height but not the width: 

https://github.com/prebid/prebid-universal-creative/pull/150/files#diff-8938ddbc8ffb985e7aeb3744e592068ea58501c022a222cb7507aabef70f8080L296 

So, a change was needed to fix this scenario: when a native ad is rendered into a GAM _regular_ creative. 

You may know that native ads can be rendered into GAM Native Creatives, that have the `fluid` size. Fluid here means that the ad will try to expand horizontally to take the size of its container. 

When you try to do that into a regular GAM creative, this is more difficult. Say we're going to render an ad in a `1x1` regular creative. In this scenario, what would happen is that a 1 pixel-large iframe is created, the ad is rendered, but since the width is 1px, it will very tall. Later, the PUC calls `resizeNativeHeight`, that tries to fix the height of the container, leaving the width ad 1px. 

With the change that was added in prebid-universal-creative there are a few things that will happen: 
1. first, the ad's width will be expanded to the size of the container. This is done even if it's a safeframe. 
2. Then, we request the ad resize, but we pass the width too. 

Other scenarios were all passing the width, so I didn't add a default here.

jsnellbaker (approved):
LGTM.   Thanks for putting this together.

I think for the 'changes' that are desired for bid adapters to adopt, it would be great if there was a 'cheat-sheet' to summarize/highlight what changes are more strongly desired vs nice-to-have (and if there are dates involved around changing the baseline in a future major release).

musikele:
@jsnellbaker will follow shortly

PR #8118, opened 2022-02-24
Yahoossp Bid Adapter: enable aliasing

slimkrazy:
## Type of change
- [x] Bugfix

## Description of change
Bid adapter was hardcoding the bidderCode in the bidResponse object, preventing aliasing.

### DIFF
--- a/modules/yahoosspBidAdapter.js
+++ b/modules/yahoosspBidAdapter.js
@@ -590,7 +590,7 @@ export const spec = {
         adId: deepAccess(bid, 'adId') ? bid.adId : bid.impid || bid.crid,
         adUnitCode: bidderRequest.adUnitCode,
         requestId: bid.impid,
-        bidderCode: spec.code,
+        bidderCode: bidderRequest.bidderCode,
         cpm: cpm,
         width: bid.w,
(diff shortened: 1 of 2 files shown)


modules/yahoosspBidAdapter.js
```diff
@@ -590,7 +590,7 @@ export const spec = {
         adId: deepAccess(bid, 'adId') ? bid.adId : bid.impid || bid.crid,
         adUnitCode: bidderRequest.adUnitCode,
         requestId: bid.impid,
-        bidderCode: spec.code,
+        bidderCode: bidderRequest.bidderCode,
```
JulieLorin, line 593: Wouldn't it be better to just remove this line ? It will be filled automatically by Prebid with the correct value (it avoids eventual incompatiblity with this method in the future)

slimkrazy: Hey @JulieLorin  - A unit test was added to ensure we don't regress here. If this line is dropped the unit test will fail. I _guess_ we could test for bidderCode to be undefined. It's a little odd, but works.

ChrisHuie (approved):

PR #8119, opened 2022-02-24
Prebid Core & Browsi RTD provider: Support Vendor Billing with Billable events

omerBrowsi:
## Type of change
- [ ] Feature

## Description of change
Support billable events, according to documentation - 
https://github.com/prebid/prebid.github.io/pull/3533


modules/browsiRtdProvider.js
```diff
@@ -332,13 +335,23 @@ export const browsiSubmodule = {
...
 
-function getTargetingData(uc) {
+function getTargetingData(uc, c, us, a) {
   const targetingData = getRTD(uc);
+  const auctionId = a.auctionId
   uc.forEach(auc => {
     if (isNumber(_ic[auc])) {
       _ic[auc] = _ic[auc] + 1;
     }
+    const transactionId = a.adUnits.find(adUnit => adUnit.code === auc).transactionId;
```
dgirardi, line 345: I believe that in general there may be more than one adUnit, and more than one transactionId, for a given code. Does it makes sense here to pick the first?

modules/browsiRtdProvider.js
```diff
@@ -332,13 +335,23 @@ export const browsiSubmodule = {
...
+    events.emit(CONSTANTS.EVENTS.BILLABLE_EVENT, {
```
dgirardi, line 346: wouldn't this emit for every adUnit code in the auction? is requesting an auction with N ad units supposed to generate N billable events regardless of the bids it gets (or does not get)?

It'd be helpful to have a test case that shows what billable events you expect an auction to generate.

modules/browsiRtdProvider.js
omerBrowsi: Shouldn't ad unit code be unique?
This is how it's documented - 

> Unique identifier that you create and assign to this ad unit. Used to set query string targeting on the ad. If using GPT, we recommend setting this to slot element ID.
https://docs.prebid.org/dev-docs/publisher-api-reference/addAdUnits.html

dgirardi: The documentation could be improved; it also says: 

https://docs.prebid.org/dev-docs/adunit-reference.html#twin-adunit-codes
> It’s ok to have multiple AdUnits with the same code. This can be useful in scenarios where bidders have different capabilities for the same spot on the page. 

If your aim is to emit an event for each adUnit, I'd just loop through them directly instead of the codes.

robertrmartinez: I believe the use case for them not being unique still winds up targeting the same slot on the page.

So, perhaps the browsi use case does in fact only want to emit for each adUnitCode?

I'll let the browsi team determine that @omerBrowsi

modules/browsiRtdProvider.js
robertrmartinez: Agree, please add actual test cases surrounding these events.

As for when the events fire, perhaps there is or needs to be some documentation on how Browsi determines it? Is it just whenever the viewability is passed into the adServer for a given adUnit?

modules/browsiRtdProvider.js
omerBrowsi: Correct. adUnitCode is the right use for us

modules/browsiRtdProvider.js
omerBrowsi: We emit event per adUnitCode, regardless of the bids

omerBrowsi: Added unit testing

robertrmartinez (approved):
thanks for adding the unit tests.

Will wait for @dgirardi to do another pass!

dgirardi (approved):

PR #8121, opened 2022-02-24
TTD Bid Adapter: Fix mediatype detection

minh-daole-ttd:
## Type of change

- [x] Bugfix
- [ ] Feature
- [ ] New bidder adapter   
- [ ] Build related changes
- [ ] CI related changes
- [ ] Other

## Description of change

This fixes the TTD prebid.js adapter's mediatype detection. It was incorrectly reading the mediatype value from the bid response from TTD bidders. The expected value should be a string rather than an int enum.

```

- prebid-maintainers@thetradedesk.com

## Other information
NA


minh-daole-ttd:
@robertrmartinez I updated the interpret response tests and they are passing now. Thank you for catching that.

minh-daole-ttd:
i've decided to cancel this and fix this on our bidder end instead to return the int value

PR #8138, opened 2022-03-03
Improve Digital Bid adapter: use the oRTB server endpoint

jbartek25:
Co-authored-by: Faisal Islam <100519197+iosfaisal@users.noreply.github.com>
Co-authored-by: Jozef Bartek <31618107+jbartek25@users.noreply.github.com>

## Type of change
- [x] other

## Description of change
A rewrite of Improve Digital adapter to
* use the new ad server endpoint based on oRTB 2.5
* removed unused legacy logic
* adding minor features like e.g. support for `gpid`
* refactoring

Adapter parameters are not changing, hence no change in the adapter documentation is needed.


modules/improvedigitalBidAdapter.js
```diff
@@ -75,187 +64,151 @@ export const spec = {
...
+        deepSetValue(request, 'regs.ext.us_privacy', bidderRequest.uspConsent);
       }
-      const genre = deepAccess(site, 'content.genre');
-      if (genre && isStr(genre)) {
-        requestParameters.genre = genre;
+      // Site
+      const url = config.getConfig('pageUrl') || deepAccess(bidderRequest, 'refererInfo.referer');
+      const urlObj = parseUrl(url);
+      deepSetValue(request, 'site.page', url);
+      deepSetValue(request, 'site.domain', urlObj && urlObj.hostname);
```
patmmccann, line 138: Please note this generates an error on lgtm; urlObj always exists

modules/improvedigitalBidAdapter.js
```diff
@@ -286,529 +223,374 @@ export const spec = {
...
-      impressionObject.banner.w = placementObject.size.w;
-      impressionObject.banner.h = placementObject.size.h;
+  buildVideoAd(bid, bidRequest, bidResponse) {
+    bid.mediaType = VIDEO;
+    bid.vastXml = bidResponse.adm;
+    if (ID_REQUEST.isOutstreamVideo(bidRequest)) {
+      bid.adResponse = { content: bid.vastXml };
+      bid.renderer = ID_OUTSTREAM.createRenderer(bidRequest);
     }
+  },
```
mmoschovas, line 464: Hmm I think this update inadvertently drops width/height definition for video bids which is a required response field as per https://docs.prebid.org/dev-docs/bidder-adaptor.html#interpreting-the-response. Can you validate that this is still being set?

jbartek25: There has never been a proper size for video bids coming from our ad server. Previously it was defaulted to 1x1 but since bidfactory defaults already to 0x0, the default size was removed and is 0x0 now in the final bid.
 https://github.com/prebid/Prebid.js/blob/master/src/bidfactory.js#L22

Screenshot 2022-03-30 at 17.49.57

mmoschovas (approved):

PR #8140, opened 2022-03-03
Kargo Bid Adapter: Removing Salesforce partner support

jsadwith:
## Type of change

## Description of change
Removing legacy partner code.


modules/kargoBidAdapter.js
```diff
@@ -172,28 +172,6 @@ export const spec = {
...
-
-    return {
-      userID: spec._getKruxUserId(),
-      segments: segments
-    };
-  },
-
   _getLocalStorageSafely(key) {
     try {
       return storage.getDataFromLocalStorage(key);
```
patmmccann, line 177: fyi, bidders will be disabled from accessing local storage by default in prebid 7

modules/kargoBidAdapter.js
```diff
@@ -172,28 +172,6 @@ export const spec = {
     return spec._getCrbFromCookie();
```
patmmccann, line 172: you should not be writing cookies in the first party domain. I know it is out of scope of this PR, but this type of behavior will be imminently disallowed.

patmmccann (approved):
This looks good, please notice commentary on setting publisher domain cookies. This behavior is extremely frowned upon and you can expect it to stop functioning in the near future.

jsadwith:
@patmmccann Thanks for your time and your notes. I'd like some further feedback on this when you have a chance - given the reliance on DSP syncing/IDs in the programmatic ecosystem, with 3P cookies going away, 1P cookies ceasing to function, and local storage being disallowed in v7, what do you recommend as a best practice to adapters for surviving in a world where the DSPs really only want to bid on synced/matched users?

PR #8143, opened 2022-03-04
cwire Bid Adapter: Add new optional parameters

kodi:
## Type of change

- [ ] Bugfix
- [x] Feature
- [ ] New bidder adapter   
- [ ] Build related changes
- [ ] CI related changes
- [ ] Other

## Description of change
~~Added 3 optional parameters to global config~~

EDIT: 
(Edited to show params in bidder params, not in global)

**1) Added 3 optional params to bidder parameters**

params are:
```javascript
var adUnits = [
  {
    code: 'target_div_id', // REQUIRED 
    bids: [{
      bidder: 'cwire',
      mediaTypes: {
        banner: {
          sizes: [[1, 1]],
        }
      },
      params: {
        ... old params ...
        cwcreative: 42,               // optional - id of creative to force
        refgroups: 'test-user',       // optional - name of group or coma separated list of groups to force
        cwapikey: 'api_key_xyz',      // optional - api key for integration testing        
      }
    }]
  }
];
```
and their main purpose is advanced debugging, or forcing single creative or group of creatives for testing and integration.

**2) Removed unused `adUnitElementId` param**

`publishers@cwire.ch` or `dragan@cwire.ch` 

docs PR: https://github.com/prebid/prebid.github.io/pull/3616

### DIFF
--- a/modules/cwireBidAdapter.js
+++ b/modules/cwireBidAdapter.js
@@ -1,6 +1,7 @@
 import {registerBidder} from '../src/adapters/bidderFactory.js';
 import {getRefererInfo} from '../src/refererDetection.js';
 import {getStorageManager} from '../src/storageManager.js';
+import {config} from '../src/config.js';
 import {BANNER, VIDEO} from '../src/mediaTypes.js';
 import {OUTSTREAM} from '../src/video.js';
 import {
@@ -13,6 +14,7 @@ import {
   isNumber,
   logError,
   logWarn,
+  isEmpty,
   parseSizesInput,
 } from '../src/utils.js';
 import {Renderer} from '../src/Renderer.js';
@@ -162,9 +164,19 @@ export const spec = {
 
     let refgroups = [];
 
-    const cwCreativeId = getQueryVariable(CW_CREATIVE_QUERY);
+    const cwCreativeId = parseInt(getQueryVariable(CW_CREATIVE_QUERY), 10) || null;
+    const cwCreativeIdFromConfig = config.getConfig('cwcreative');
+    const refGroupsFromConfig = config.getConfig('refgroups');
+    const cwApiKeyFromConfig = config.getConfig('cwapikey');
     const rgQuery = getQueryVariable(CW_GROUPS_QUERY);
+
+    if (!isEmpty(refGroupsFromConfig)) {
+      refgroups = refGroupsFromConfig.split(',');
+    }
+
     if (rgQuery !== null) {
+      // override if query param is present
+      refgroups = [];
       refgroups = rgQuery.split(',');
     }
 
@@ -173,8 +185,9 @@ export const spec = {
     const payload = {
       cwid: localStorageCWID,
       refgroups,
-      cwcreative: cwCreativeId,
+      cwcreative: cwCreativeId || cwCreativeIdFromConfig || null,
       slots: slots,
+      cwapikey: cwApiKeyFromConfig || null,
       httpRef: referer || '',
       pageViewId: CW_PAGE_VIEW_ID,
     };
--- a/modules/cwireBidAdapter.js
+++ b/modules/cwireBidAdapter.js
@@ -99,11 +99,9 @@ export const mapSlotsData = function(validBidRequests) {
     // get the pacement and page ids
     let placementId = getValue(bid.params, 'placementId');
     let pageId = getValue(bid.params, 'pageId');
-    let adUnitElementId = getValue(bid.params, 'adUnitElementId');
     // get the rest of the auction/bid/transaction info
     bidObj.auctionId = getBidIdParameter('auctionId', bid);
     bidObj.adUnitCode = getBidIdParameter('adUnitCode', bid);
-    bidObj.adUnitElementId = adUnitElementId;
     bidObj.bidId = getBidIdParameter('bidId', bid);
     bidObj.bidderRequestId = getBidIdParameter('bidderRequestId', bid);
     bidObj.placementId = placementId;
@@ -131,11 +129,6 @@ export const spec = {
    */
   isBidRequestValid: function(bid) {
     bid.params = bid.params || {};
-
-    // if ad unit elemt id not provided - use adUnitCode by default
-    if (!bid.params.adUnitElementId) {
-      bid.params.adUnitElementId = bid.code;
-    }
 
     if (!bid.params.placementId || !isNumber(bid.params.placementId)) {
       logError('placementId not provided or invalid');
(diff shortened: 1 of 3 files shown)


dgirardi:
At first glance it looks like these should be bidder parameters (https://docs.prebid.org/dev-docs/bidder-adaptor.html#designing-your-bid-params). Is there a case for them to be global config?

kodi:
Hello [dgirardi](https://github.com/dgirardi), thank you for your comment!

> At first glance it looks like these should be bidder parameters (https://docs.prebid.org/dev-docs/bidder-adaptor.html#designing-your-bid-params). Is there a case for them to be global config?

True, but since those are global, our ad server does not support different values of those for different ad units, and they are global.

For example request towards our prebid endpoint looks something like this:
```json
{
"pageViewId": "some-uuid",
"slots": [
  ... ad units here
]
}
```

and then, setting e.g. `cwcreative` parameter works like this

```json
{
"pageViewId": "some-uuid",
"cwcreative": 42
"slots": [
  ... ad units here
]
}
```

This will force that we show always creative with our internal id of 42, and it will only return one bid, even if there are multiple ad units defined for the auction. 

Since our prebid ad server does not support forcing different-creatives-per-ad-slot then it doesn't make sense to repeat this value in all of them, and it could be error prone, because it would require to fetch one or multiple values, and then extract that to top level. Other two param work in identical way, and are also global and are forcing global behaviour. 

If there is a better way to set a param just once, and not every time per slot, please advise on how to achieve that in a more clean way.

Also, this will be mostly used for integration testing and debugging purposes, or in some really custom integrations where publisher requires to just post and force a single creative id and return just a single bid.

In other, proper, integrations, those parameters will never be set.

kodi:
Some test that is totally unrelated to our adapter is not working, and I don't have permissions required to re-run it:

```  
finteza analytics adapter
    track
      bid request
        ✗ builds and sends data
	AssertionError: expected 2 to equal 1
	    at Context.<anonymous> (webpack:///test/spec/modules/fintezaAnalyticsAdapter_spec.js:79:43 <- test/test_index.js:325451:140)```

**Comment by dgirardi** [comment]
IMO, since those parameters are at least conceptually tied to the adUnit (if the backend supported it, it'd make sense to have two different sets of them), it'd be better take them in as bid parameters and "fix" the discrepancy by picking the first one / enforcing them to be identical / whatever works best for you (especially since they are meant for testing only). 

If that does not work, I'll need to investigate how we intend to support bidder-specific configuration - I am concerned about setting a precedent for using the global, top-level namespace like this.

**Comment by kodi** [comment]
@dgirardi 
Yes, you are totally correct, My bad!

Actually, I wanted to create 

```
cwire: {
   param1,
   param2,
}
```
global config, but somehow by mistake it went to a root.

But upon more reading, and actually figuring out how is this supposed to work ( 😄 ), I now agree. This should go to bidder params.

We will pick a first value from any bid, and just ignore other + print a warning to a console (using `logWarn` if that is ok ?).

Returning PR to a draft state until I fix all the docs and tests.

**Comment by kodi** [comment]
@dgirardi 
Fixed and moved, removing draft status.

Thanks for your feedback!

File: modules/cwireBidAdapter.md [hunk 1]
```diff
@@ -35,9 +39,12 @@ var adUnits = [
       params: {
         pageId: 1422,                 // required - number
         placementId: 2211521,         // required - number
-        adUnitElementId: 'other_div', // optional, div id to write to, if not set it will default to ad unit code  
+        adUnitElementId: 'other_div', // optional, div id to write to, if not set it will default to ad unit code
```
dgirardi, line 42: This does not actually do what it says, right? It looks like it's sent over to the backend but has no effect otherwise.

test/spec/modules/cwireBidAdapter_spec.js
```diff
@@ -138,12 +140,145 @@ describe('C-WIRE bid adapter', () => {
...
+      expect(requests.data.slots.length).to.equal(1);
+      expect(requests.data.cwid).to.be.null;
+      expect(requests.data.cwid).to.be.null;
+      expect(requests.data.slots[0].sizes[0]).to.equal('1x1');
+      expect(requests.data.cwcreative).to.equal(654321);
+      expect(requests.data.cwapikey).to.equal('xxx-xxx-yyy-zzz');
+      expect(requests.data.refgroups[0]).to.equal('group_2');
+    });
+
+    it('creates a valid request - if config not set null or empty array is sent', function () {
```
dgirardi, line 281: I don't see an empty array - is the test case name accurate?

kodi: Ah, bad wording

it was `if config not set null or empty array is sent`

renamed to:
`if params are not set, null or empty array are sent to the API`

and then, there is check for that, nothing changed there:

```javascript
      expect(requests.data.cwcreative).to.equal(null);
      expect(requests.data.cwapikey).to.equal(null);
      expect(requests.data.refgroups.length).to.equal(0);
```

so, if params are not set anywhere, I send null  or empty array

modules/cwireBidAdapter.js
```diff
@@ -162,9 +185,19 @@ export const spec = {
 
     let refgroups = [];
 
-    const cwCreativeId = getQueryVariable(CW_CREATIVE_QUERY);
+    const cwCreativeId = parseInt(getQueryVariable(CW_CREATIVE_QUERY), 10) || null;
```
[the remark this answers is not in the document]
kodi: 0 is not valid creative ID.

modules/cwireBidAdapter.md
kodi: This was just some old thing for testing, where we would write in some other div. It is ok that it's only sent to the backend, since then our bootstrap script knows where to write content.

But since this is confusing, I will remove any notion of this param, util someone actually requests this feature.

modules/cwireBidAdapter.js
```diff
@@ -162,9 +185,19 @@ export const spec = {
...
+    const cwApiKeyFromConfig = this.getFirstValueOrNull(slots, 'cwapikey');
     const rgQuery = getQueryVariable(CW_GROUPS_QUERY);
+
+    if (refGroupsFromConfig !== null) {
+      refgroups = refGroupsFromConfig.split(',');
+    }
+
     if (rgQuery !== null) {
+      // override if query param is present
+      refgroups = [];
```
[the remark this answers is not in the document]
dgirardi: I still think this should be removed as it's useless, but you'll be the maintainer, so your choice :)

dgirardi (approved):

PR #8146, opened 2022-03-04
TheMediaGrid & TrustX bid adapters: reformat first party data

TheMediaGrid:
## Type of change

- [x] Feature

## Description of change

Update TheMediaGrid Bid Adapter to reformat segments received from permutiveRTDProvider
Update TrustX Bid Adapter to send all ortb2.user.data in user.data

- grid-tech@themediagrid.com


test/spec/modules/gridBidAdapter_spec.js
```diff
@@ -546,6 +546,41 @@ describe('TheMediaGrid Adapter', function () {
...
+          name: 'permutive.com',
+          segment: [{ value: '1' }, { value: '2' }, { value: 'segVal' }, { value: 'segId' }, { value: 'segValue' }, { value: 'segId2', name: 'segName' }]
+        },
+        {
+          someKey: 'another data'
+        }
+      ]);
+      getConfigStub.restore();
+    });
+
```
patmmccann, line 583: Spelling right before this line as well

patmmccann:
Is the permutive module making a bad segments object, if so, why not fix that instead of work around it?

test/spec/modules/trustxBidAdapter_spec.js
```diff
@@ -401,30 +401,39 @@ describe('TrustXAdapter', function () {
...
-        ]
-      }]);
+      expect(payload.user.data).to.deep.equal([
+        {
+          name: 'someName',
+          segment: [1, 2, { anyKey: 'anyVal' }, 'segVal', { id: 'segId' }, { value: 'segValue' }, { id: 'segId2', name: 'segName' }]
+        },
+        {
+          name: 'permutive.com',
+          segment: [{ value: '1' }, { value: '2' }, { value: 'segVal' }, { value: 'segId' }, { value: 'segValue' }, { value: 'segId2', name: 'segName' }]
```
patmmccann, line 430: Why do these all say value instead of id

patmmccann (approved):

PR #8148, opened 2022-03-07
Yieldlab Bid Adapter: Add Support for User Matching

nkloeber:
## Type of change
- [x] Feature

## Description of change
Added support for user matching to the yieldlabBidAdapter.js with iframe as possible sync option


ncolletti (approved):
lgtm

PR #8159, opened 2022-03-09
Prebid Server Bid Adapter: add support for all imp parameters

ChrisHuie:
Fix for issue #7831 . Add support for all Imp object properties instead of just 'ext.data'

Also, added testing from pr #8040 to cover instl

### DIFF
--- a/test/spec/modules/prebidServerBidAdapter_spec.js
+++ b/test/spec/modules/prebidServerBidAdapter_spec.js
@@ -2910,6 +2910,25 @@ describe('S2S Adapter', function () {
       expect(requestBid.coopSync).to.be.undefined;
     });
 
+    it('should set imp banner if ortb2Imp.banner is present', function() {
+      const consentConfig = { s2sConfig: CONFIG };
+      config.setConfig(consentConfig);
+      const bidRequest = utils.deepClone(REQUEST);
+      bidRequest.ad_units[0].ortb2Imp = {
+        banner: {
+          api: 7
+        }
+      };
+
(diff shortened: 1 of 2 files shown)


patmmccann:
Can you please add @TheMediaGrid unit test from https://github.com/prebid/Prebid.js/pull/8040/files to this pr so we can confirm it solves both

test/spec/modules/prebidServerBidAdapter_spec.js
```diff
@@ -2910,6 +2910,25 @@ describe('S2S Adapter', function () {
...
+        }
+      };
+
+      adapter.callBids(bidRequest, BID_REQUESTS, addBidResponse, done, ajax);
+      const parsedRequestBody = JSON.parse(server.requests[0].requestBody);
+
+      expect(parsedRequestBody.imp).to.be.a('array');
+      expect(parsedRequestBody.imp[0]).to.be.a('object');
+      expect(parsedRequestBody.imp[0]).to.have.property('banner');
+      expect(parsedRequestBody.imp[0].banner.api).to.equal(7);
```
dgirardi, line 2930: IMO only the last assertion is needed. #2 and #3 have to be true for #4 to be true. #1 is better suited for other tests.

dgirardi (approved):

test/spec/modules/prebidServerBidAdapter_spec.js
patmmccann: Funny how GitHub linked those prs

patmmccann (approved):

harpere (approved):

PR #8168, opened 2022-03-10
DFP Ad Server Video: respect original url

karimMourra:
## Type of change

- [x] Bugfix

## Description of change
Fixes a bug where when `buildDfpVideoUrl` is called with both a url and params, the original url is modified extensively:

- the url's host, protocol and pathname are overriden. 
- if a size param is included in the url, it is overriden.
- cust_params in the url are discarded.

For example, if my ad tag url's host is `pubads.g.doubleclick.net`, it gets replaced with  `'securepubads.g.doubleclick.net'`. Instead, `'securepubads.g.doubleclick.net'` should be used as a fallback. 

With this change, the original host, protocol and pathname are kept. Additionally, sizes and cust_params are appended to any existing values in the URL.

### DIFF
--- a/modules/dfpAdServerVideo.js
+++ b/modules/dfpAdServerVideo.js
@@ -95,11 +95,7 @@ export function buildDfpVideoUrl(options) {
     derivedParams.sz = urlSzParam + '|' + derivedParams.sz;
   }
 
-  let encodedCustomParams = getCustParams(bid, options);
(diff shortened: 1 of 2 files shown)


patmmccann:
@mike-chowla could you take a glance at this one?

ChrisHuie (approved):

PR #8171, opened 2022-03-11
Prebid Core and Several Bid Adapters: fix win notification price bug

### LINKED ISSUE #8172: replaceAuctionPrice bug
## Type of issue
Bug

## Description
auction price macros appear to be replaced by the discounted cpm, not the bid price

eg
https://github.com/prebid/Prebid.js/blob/3fd3b5dca5dcc3fcd9f3c949e8e1a3238ea6b836/src/prebid.js#L449
 
and many adapters also have this behavior. Only

https://github.com/prebid/Prebid.js/blob/f011bbc78fe4b50ff9f827aa39c59ecd8be0a1bb/modules/adxcgBidAdapter.js#L326

and Outbrain adapter does it correctly from what I can tell. #8171 is an attempt to fix

patmmccann:
Some adapters appear to be sending themselves win notifications with the discounted price, not the price they bid. Publishers discount for discrepencies, then an adapter lowers the price they pay, then publishers discount more, in an endless loop. This fixes the bug. ADXCG and Outbrain and others were passing correctly. 

Fixes #8172

### DIFF
--- a/modules/apacdexBidAdapter.js
+++ b/modules/apacdexBidAdapter.js
@@ -197,9 +197,9 @@ export const spec = {
         bidResponse.dealId = dealId;
       }
       if (bid.vastXml) {
-        bidResponse.vastXml = replaceAuctionPrice(bid.vastXml, bid.originalCpm);
+        bidResponse.vastXml = replaceAuctionPrice(bid.vastXml, bid.cpm);
       } else {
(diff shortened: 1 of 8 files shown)


patmmccann:
closing while still researching this....

mmoschovas (approved):

ChrisHuie (approved):

Issue #5287, opened 2020-05-21
Considering a "consumer" version of npm package


khatibda:
We just transitioned from building an offline bundle of prebid.js (using cloned repo) to using prebid via npm, and it's almost the best thing since sliced bread.  But I have some questions / feedback on the process.

The current npm process feels geared towards "developer" and opposed to "consumers" (e.g. publishers).  Bet let's walk through the details in case I'm missing something.

First, we've include prebid in our own package.json. Then in our app we have something like this:

`// import the pbjs global.  the version pulled is determined in package.json
import pbjs from 'prebid.js';

// register the desired modules/adapters
import 'prebid.js/modules/aolBidAdapter';
import 'prebid.js/modules/appnexusBidAdapter';
import 'prebid.js/modules/consentManagement';
import 'prebid.js/modules/consentManagementUsp';
import 'prebid.js/modules/ixBidAdapter';
import 'prebid.js/modules/oneVideoBidAdapter';
import 'prebid.js/modules/openxBidAdapter';
import 'prebid.js/modules/prebidServerBidAdapter';
import 'prebid.js/modules/pubmaticBidAdapter';
import 'prebid.js/modules/rubiconBidAdapter';
import 'prebid.js/modules/spotxBidAdapter';
import 'prebid.js/modules/telariaBidAdapter';
import 'prebid.js/modules/tripleliftBidAdapter';
import 'prebid.js/modules/userId';

// not in prebid docs, but this seems required and is consistent with prior custom bundled code
pbjs.processQueue();`

Minor feedback, but maybe your README.md could show the npm example with "import pbjs" instead of "import prebid", since you can define the global any way you want on import, so might as well be consistent with the default code publishers already likely have implemented in their app (pbjs.requestBids, etc).

Also, in our particular case, we weren't able to fetch bids until we called processQueue after import.  This wasn't documented, but we tried it because we saw the same thing at the end of the prebid bundle we built offline (repo clone).  I'm not sure why, but it worked once we did that.

[...]

Issue #6661, opened 2021-04-27
Enhanced NPM Instructions for Prebid Packages


gglas:
Issue 1780 in the docs repo - https://github.com/prebid/prebid.github.io/issues/1780 - alludes to enhanced functionality for the downloads page, making it easier for publishers to save the various versions of Prebid that they've downloaded.  While the reviewer committee doesn't necessarily think this is a good idea, we do think it would be a good idea to have enhanced documentation for using NPM so that even unsophisticated users are comfortable using it to manage their Prebid packages.  This would include very basic instructions on how to use NPM (from start to finish) and potentially an explainer video to help less sophisticated users.

khatibda:
anything pro npm gets my vote!

bretg:
Taking the explainer video out of scope here. This can be resolved with some tweaks to the README.

Please add some context for the options:
- original install + copy
- as an npm dependency
- website build

patmmccann:
https://github.com/prebid/Prebid.js/issues/9333#issuecomment-1349287552 @ChrisHuie please add a warning about referencing non-public object

PR #8175, opened 2022-03-11
Build system: make compiled prebid-core and modules available for npm consumers

dgirardi:
## Type of change

- [x] Feature
- [x] Build related changes

## Description of change

This updates the build to make compiled files (both minified and not) available under 'dist', so that NPM consumers may use them without needing to set up Babel transpilation.

Resolves https://github.com/prebid/Prebid.js/issues/5287

**NOTE**: this also needs an update in the release process; please review but do not merge.


jsnellbaker (approved):
@dgirardi 

From what I can tell of how this should work (from the description, etc), this seems to be working as described.  As requested though, I will not merge until it's fully ready.

If you know anyone who should look at this change, feel free to add them.  I'm not sure who uses npm with prebid off-hand, but they'd probably have an interest here.  

@snapwich do you happen to know?  I think you may have done some work on this topic in the past (if I remember rightly).

dgirardi:
There's quite a bit of interest about this expressed in #5287, although that issue is quite old now and I'm not sure if that still applies.

dgirardi:
@khatibda does this approach match your expectation for a "consumer" npm package? The usage would look like:

```javascript
import 'prebid.js/dist/min/prebid-core.js';
import 'prebid.js/dist/min/some-module.js';
// ...
pbjs.processQueue();
```

with no transpilation necessary. There is no automagical detection of production / development flags as I'm not sure how you could do that without being opinionated about the build tools in use - which is what I think React does.

khatibda:
> @khatibda does this approach match your expectation for a "consumer" npm package? The usage would look like:
> 

no transpiling would be magical!

i'm assuming consumer npm use case is mostly a "production" case even in dev.  different from contributor npm use case.  but i could be wrong. regardless, at least for webpack, if both dev/prod import paths are available, there are ways to handle.

most major package imports look a little "nicer" in terms of syntax used, e.g. import 'prebid/core', but i think that's probably just sugar and not a big deal if complicated to solve.

patmmccann:
@khatibda we're having some trouble here in that we cannot find a second reviewer. Are you able to review?

khatibda:
@dgirardi 

final q, i recall the big hangup before was about dynamically defining the global variable name that happened at transpile time

i don't see any commentary on this thread, but i do see the inline comment example referring to "pbjs.processQueue()"

so...does this presume/mandate that the origin global is pbjs?  or do i have it wrong?

(which i suppose devs can reference themselves via variable reference if they really need to.)

khatibda:
looks like a merge conflict for gulpfile.js as well

dgirardi:
@khatibda yes - the global var is always `pbjs` unless you run the build yourself. 

I suppose we could try to transpile into es6 modules and avoid the global completely.

khatibda:
> @khatibda yes - the global var is always `pbjs` unless you run the build yourself.
> 
> I suppose we could try to transpile into es6 modules and avoid the global completely.

i think pbjs is great/fine.  it will work better than no globals.  going to approve.

khatibda (approved):
lgtm

patmmccann:
@dgirardi this appears ready to merge with a docs commit; some changes to https://github.com/dgirardi/Prebid.js/blob/consumer-npm/README.md, and the conflict resolved.

dgirardi:
@patmmccann it also needs a change in the release automation scripts (and some coordination to make them happen at the same time), so keep this labeled as `do not merge` please - I'll follow up shortly with conflict resolution /README updates (I don't think docs cover the build process, but correct me if that needs an update as well)

patmmccann:
is it possible to rely on https://github.com/prebid/Prebid.js/blob/97ae929aee9c909587ca5fce2580c910db67c2cd/src/prebidGlobal.js#L9 to get the global?

dgirardi:
@patmmccann yes, but I don't think it makes sense in this context - that will always contain just `'pbjs'` unless you ran a custom build, in which case you probably already know what you set it to.

muuki88:
We are using custom builds as well, but we quite like the Babel step in between as it allows us to compile different bundles depending on the targeted browser ( es6 module build and es5 with polyfills ).

@khatibda getting rid of Babel in your build means you rely on prebid.js output being compatible with the devices you want your bundle to be compatible. This can come at a cost of larger JS bundles.

While I think this is interesting for faster example builds, this is not something we would use for production.

patmmccann:
The docs changes @dgirardi proposes should also solve https://github.com/prebid/Prebid.js/issues/6661

mkendall07:
I know this is super old but is it on the roadmap to have dist build files published for NPM? I imagine it would make things a lot easier for folks importing via NPM.

dgirardi:
@mkendall07 it's still an open request - and I'd like to get to it - but other work has taken precedence so far. I'll close this PR (it's inadequate) to continue the dicussion in https://github.com/prebid/Prebid.js/issues/10086

Issue #8165, opened 2022-03-10
getSpec() for spotX has undefined gvlid


grajzer:
## Type of issue
Bug

## Description
When gdprEnforcement is used, getSpec() is called for each adapter. Object returned for SpotX has undefined gvlid property, although it's defined in spotxBidAdapter.js (gvlid: 165).

## Steps to reproduce
Downloaded prebid.js file and modified it by adding one console.log where that function is called.

![screen](https://user-images.githubusercontent.com/5851760/157626403-40dc697f-1133-47d2-93f5-edaab606faf5.png)

https://dejan.brid.tv/spotxgvlid/prebidpagetest.html

patmmccann:
Also, what is this undocumented parameter

https://github.com/prebid/Prebid.js/blob/38446e6f81a867b485f28e822dc3af3673367c1e/modules/spotxBidAdapter.js#L240

patmmccann:
Also, interestingly, https://github.com/prebid/Prebid.js/blob/38446e6f81a867b485f28e822dc3af3673367c1e/modules/spotxBidAdapter.js#L14 does not match the spotx doc, which has gvlid 52

dgirardi:
@harpere can you take a look at this?

robertrmartinez:
@patmmccann @dgirardi I will take this. Docs were updated to 52 which is the Rubicon / Magnite GVL ID. Will work with our team to fix and put the correct stuff in here asap!

PR #8186, opened 2022-03-16
Spotx Bid Adapter: Fix gvlid undefined + Update Magnite adapters to use proper GVLID's (Slimcut & Telaria)

robertrmartinez:
## Type of change
- [X] Bugfix
- [X] Other

## Description of change
Addresses #8165 

### SpotX gvldid undefined

This was happening because spotx bid adapter declared it's own bidder code as an alias.

And the code in [bidderFactory.js](https://github.com/prebid/Prebid.js/blob/master/src/adapters/bidderFactory.js#L163) loops through the alias array and expects it to have gvlId, and if not, then sets it to undefined.

Because spotx used same bidder code in the alias array, it was overwriting the main spec! Which may have caused other weird issues as well.

Instead of declaring spotx, we now just do not declare it and the GVL shows up!

### Magnite Adapters now use Magnite gvlid
Magnites other adapters now will use the MAgnite gvlid `52` instead of the older ones.

### DIFF
--- a/modules/slimcutBidAdapter.js
+++ b/modules/slimcutBidAdapter.js
@@ -9,7 +9,8 @@ const BIDDER_CODE = 'slimcut';
 const ENDPOINT_URL = 'https://sb.freeskreen.com/pbr';
 export const spec = {
   code: BIDDER_CODE,
-  aliases: ['scm'],
+  gvlid: 52,
+  aliases: [{ code: 'scm', gvlid: 52 }],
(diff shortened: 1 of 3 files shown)


bretg (approved):

patmmccann (approved):

patmmccann:
@robertrmartinez please put in a docs pull on https://docs.prebid.org/dev-docs/bidders/slimcut.html

PR #8187, opened 2022-03-16
Add dacid user id module

kyoya-takei:
## Type of change

- [ ] Bugfix
- [x] Feature
- [ ] New bidder adapter   
- [ ] Build related changes
- [ ] CI related changes
- [ ] Other

## Description of change

Add dacid user id module.

  - https://github.com/prebid/prebid.github.io/pull/3644

### DIFF
--- a/modules/dacIdSystem.js
+++ b/modules/dacIdSystem.js
@@ -1,0 +1,57 @@
+/**
+ * This module adds dacId to the User ID module
+ * The {@link module:modules/userId} module is required
+ * @module modules/dacIdSystem
+ * @requires module:modules/userId
+ */
+
+import { submodule } from '../src/hook.js';
+import { getStorageManager } from '../src/storageManager.js';
+
+export const storage = getStorageManager();
+
+export const cookieKey = '_a1_f';
+
+export const dacIdSystemSubmodule = {
+  /**
+   * used to link submodule with config
+   * @type {string}
+   */
+  name: 'dacId',
+
+  /**
+   * performs action to obtain id
+   * @function
+   * @returns { {id: {dacId: string}} | undefined }
+   */
+  getId: function() {
+    const newId = storage.getCookie(cookieKey);
+    if (!newId) {
+      return undefined;
+    }
+    const result = {
+      dacId: newId
+    }
+    return {id: result};
+  },
+
+  /**
+   * decode the stored id value for passing to bid requests
+   * @function
+   * @param { {dacId: string} } value
+   * @returns { {dacId: {id: string} } | undefined }
+   */
+  decode: function(value) {
+    if (value && typeof value === 'object') {
+      const result = {};
+      if (value.dacId) {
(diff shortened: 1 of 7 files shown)


kyoya-takei:
I have one question about `source` value in [modules/userId/eids.js](https://github.com/prebid/Prebid.js/pull/8187/files#diff-95dd325d8722bd7c65809e0a024011442d1f656332bb822271aa92db62bec52c).
Please tell me which source should I write:

1. Domain of the Business company responsible for this user ID
2. Domain of the company issuing this user ID

modules/dacIdSystem.js
```diff
@@ -0,0 +1,57 @@
...
+   */
+  getId: function() {
+    const newId = storage.getCookie(cookieKey);
+    if (!newId) {
+      return undefined;
+    }
+    const result = {
+      dacId: newId
+    }
+    return {id: result};
```
smenzer, line 35: why not consolidate into a single `return` statement? no issue not to...just a suggestion

smenzer:
> 1. Domain of the Business company responsible for this user ID
> 2. Domain of the company issuing this user ID

i'm not sure i understand the difference between the two, maybe you can elaborate? In general, the company supplying the ID should have their domain listed in the eids source. But this should be the tech domain for the ID. For example...I work for ID5, and our corporate domain is id5.io. But our tech domain is id5-sync.com, and this tech domain is what we put in the source field. I hope this helps, but let me know if you have further questions here.

modules/dacIdSystem.md
```diff
@@ -0,0 +1,25 @@
+## DAC User ID Submodule
+
+## Building Prebid with DAC ID Support
+
+First, make sure to add the DAC ID submodule to your Prebid.js package with:
```
smenzer, line 5: This doc is fine, but it may be valuable to explain any other dependencies to use your module. Based on the code, it seems like separate code needs to be in place for this module to pick up your ID. So you may want to talk about that requirement and how to get in touch with you if the publisher would like to work with you.

Not a requirement, just a suggestion without knowing your business model or anything.

smenzer (commented):
Overall looks good...a few comments and based on your response I'll approve and merge or hold off for you to make adjustments.

modules/dacIdSystem.js
kyoya-takei: Because it may return values other than id in the future.

kyoya-takei:
@smenzer 
I have changed the value of source in response to your answer.
Thank you.

modules/dacIdSystem.md
kyoya-takei: Thank you for comment.
I'll add document.

modules/dacIdSystem.js
smenzer: ok sounds good.

smenzer (approved):
LGTM

PR #8191, opened 2022-03-16
Vibrant Media Bid Adapter: onBidWon pixel trigger

matt-vib:
## Type of change

- [ ] Bugfix
- [x] Feature
- [ ] New bidder adapter   
- [ ] Build related changes
- [ ] CI related changes
- [ ] Other

## Description of change

Triggered Vibrant internal pixel endpoint on bidWon event handler

- contact email of the adapter’s maintainer: [kormorant@vibrantmedia.com](mailto:kormorant@vibrantmedia.com)

### DIFF
--- a/modules/vibrantmediaBidAdapter.js
+++ b/modules/vibrantmediaBidAdapter.js
@@ -13,6 +13,7 @@ import {OUTSTREAM} from '../src/video.js';
 
 const BIDDER_CODE = 'vibrantmedia';
 const VIBRANT_MEDIA_PREBID_URL = 'https://prebid.intellitxt.com/prebid';
+const VALID_PIXEL_URL_REGEX = /^https?:\/\/[a-z0-9-]+(\.[a-z0-9-]+)+([/?].*)?$/;
 const SUPPORTED_MEDIA_TYPES = [BANNER, NATIVE, VIDEO];
 
 /**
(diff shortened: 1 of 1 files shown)


modules/vibrantmediaBidAdapter.js
```diff
@@ -213,7 +213,9 @@ export const spec = {
    * @param {*} bidData the data associated with the won bid. See example above for data format.
    */
   onBidWon: function(bidData) {
-    logInfo('Bid won: ' + JSON.stringify(bidData));
+    if (bidData && bidData.meta && bidData.meta.wp) {
+      triggerPixel(`${bidData.meta.wp}${bidData.status}`);
```
patmmccann, line 217: can you make sure biddata.wp is a url?

matt-vib: bidData.meta.wp is guaranteed to be a valid url, since it's an internal endpoint we use for reporting purposes (bids/won bids ratio and similar stats), generated server-side because we use several dynamic query string parameters in it.
We could add regex-based URL validation for bidData.meta.wp in the adapter though. Would that meet your expectations?

ChrisHuie: That works :)

matt-vib: @ChrisHuie updated accordingly, thanks :)

patmmccann (approved):

Issue #8074, opened 2022-02-16
Avoid multiple n/w calls when original bidder and alias bidder configured together.


pm-harshad-mane:
## Scenario
A bidder with the name `BidderA` has an alias as `BidderAAA`.
This  bidder accepts only one parameter say `publisherId`
On the SSP server-side of the bidder, they know for a given publisher, whether they are enabled as `BidderA` or `BidderAAA` or both.
So based on the `publisherId`, the server-side can respond with separate arrays of bids that can be submitted per alias.
Hence the bidder does not need to make a separate network call per alias from the publisher page.
What is the ideal way this bidder should submit the bids in Prebid JS? 
Is it fine if the bidder sets the `bid.bidderName` based on its own logic and then submits the bid to the Prebid core logic?
I found that this way works fine.

patmmccann:
This is very interesting, we call IX adapter and IX roundel alias for example. @ix-prebid-development . If there is a better way, it could be a significant overhead savings.

Issue #8129, opened 2022-03-01
Proposal - expand "extra" bid support


bretg:
## Type of issue
enhancement

## Description

Related to:
- https://github.com/prebid/Prebid.js/issues/8074
- https://github.com/prebid/Prebid.js/issues/7688

There have been several use cases come up lately where bid adapters are multiplexing multiple demand sources behind the scenes. This is beyond the concept of "alias" in that the PBJS AdUnit may not contain the biddercode at all. Certainly in the case of the long-dreamed [Prebid.less](https://github.com/prebid/Prebid.js/issues/5001) scenario, there wouldn't be a list of bidders defined client-side.

In some scenarios, this behavior warrants caution, as it could result in "bid jamming" -- a DSP seeing a given auction more than once. However, there are valid scenarios, so this issue proposes conventions and controls.

### Conventions

We propose that two fields can manage this scenario:
- **bidResponse.bidderCode** indicates the entity making the bid. This is already set by bid adapters that support aliases and by the recent pbsBidAdapter changes allowing "extra" bids. (e.g. districtm)
- **bidResponse.adapterCode** indicates the bidder code that was in the adunit. This might be an alias of an underlying bid adapter. e.g. it would be "groupm" and not "pubmatic". This is a new field that would be set to the label used by the adunit.

### Controls

The PBS "extra" bids feature established a flag to let publishers decide whether bidders not declared in the AdUnit can bid.
```
pbjs.setConfig({
   s2sConfig: {
       ...
       allowUnknownBidderCodes: true
   }
});
```
This flag defaults to `false`.

Proposals

1. extend this control to client-side bidders on a per-adapter basis and add the ability to control which biddercodes may be added:
```
pbjs.bidderSettings = {
    standard: {
         allowAlternateBidderCodes: false,    // the default if not specified for a specific adapter
         [...]
    },
    bidderA: {
        allowAlternateBidderCodes: true,
        allowedAlternateBidderCodes: ["bidderX", "bidderY"], // bidderA may add X or Y if it wishes
        [...]
    },
    bidderB: {
        allowAlternateBidderCodes: true,
        allowedAlternateBidderCodes: ["*"],        // the default allows this bidder to respond with any biddercodes
        [...]
    }
}
```
1. Update PBJS-core to enforce these rules: if the bid response bidderCode is different than the adapterCode:
    1. check bidderSettings.adapterCode.allowAlternateBidderCodes. If it exists:
        1. If false, reject
[...]

PR #8193, opened 2022-03-17
PubMatic Bid Adapter: Added multibid support for GroupM

pm-azhar-mulla:
## Type of change

- [x] Feature

## Description of change

WRT https://github.com/prebid/Prebid.js/issues/8074
We have implemented a technique to reduce the network calls, we will not initiate network call for Alias but instead SSP will handles it using the publisherId it has and sends multiple bids in response. On receiving the response, depending upon the marketplace property we will set the bidderCode and bidder for particular bid.


pm-harshad-mane (commented):
Looks good to me

patmmccann:
@pm-harshad-mane have you looked at https://github.com/prebid/Prebid.js/issues/8129?

pm-harshad-mane:
Thank you @patmmccann for pointing out issue #8129.

pm-harshad-mane:
@pm-azhar-mulla , we will need to make changes to follow #8129

ChrisHuie (approved):

Issue #5447, opened 2020-07-01
User ID module not getting data from GDPR consent module


EskelCz:
## Description
We're using OneTrust CMP (to provide us with consent string from TCF API v2) first as a fast loaded stub (tcf.stub.js), which gets filled later on. This works great for bid requests, adapters seem to be getting the consent correctly. Problem comes with passing the string into user ID module ID5. The module is not getting the consent data, therefore it's sending an empty string.

I think it might be caused by the delayed CMP loading, but it's just a guess at the moment.

## Steps to reproduce
1. Add iab (TCF v2) compliant CMP, ideally OneTrust, which is using the stub loading method
2. Add latest Prebid with id5 module and GDPR consent module
3. Configure according to the example at the bottom of this page: https://console.id5.io/docs/public/prebid

## Test page
We have a demo page to test both the CMP and ID5 sync: [https://demo.cpex.cz/cmp/](https://demo.cpex.cz/cmp/?pbjs_debug=true)
Open it with network debugger open and (if you don't already have id5 cookie) you should see the id5 server request: `https://id5-sync.com/g/v1/250.json?1puid=&gdpr=0&gdpr_consent=`

### Expected results
The module should wait for the consent data, the same way as bidder adapters seem to do. The id5 network request should therefore contain `gdpr=1` and the consent string.

### Actual results
The consent is not present.

## Platform details
I'm testing on Prebid 3.23.0, chrome, mac os

smenzer:
Thanks @EskelCz for writing this up.

Prebid team - I've debugged this a bit and I can see that calls to [getId()](https://github.com/prebid/Prebid.js/blob/master/modules/userId/index.js#L431) are actually happening before the call to the [cmpCallMap](https://github.com/prebid/Prebid.js/blob/master/modules/consentManagement.js#L270), which means the userId module has no way of getting consent. I can't tell _why_ this is happening, though, just that it is.

EskelCz:
I should also add that we had this issue even with TCF 1.1 and our in-house built CMP which was loaded more synchronously, so it's probably a problem specifically with the consentModule->ID5 interaction.

smenzer:
I think this is more general than just ID5, I think it's the entire userId module that is impacted

SKOCHERI:
Debugging with the  https://demo.cpex.cz/cmp/ it does not seem to be an issue with Prebid.js.

[...]

PR #8201, opened 2022-03-21
UserID module: better initialization logic

dgirardi:
## Type of change

- [x] Bugfix

## Description of change

﻿This updates the userID module to avoid premature initialization (https://github.com/prebid/Prebid.js/issues/5447):

 - `pbjs.getUserIds` and `pbjs.getUserIdsAsEids` no longer force initialization of the ID system
 - ID submodules are now initialized after after GDPR consent has been resolved *and* `userSync` configuration has been set
 - added `pbjs.getUserIdsAsync()`, which returns a promise to user IDs, completing only once all ID submodules have completed init
 - `pbjs.refreshUserIds` now forces an immediate refresh regardless of `userSync` config and returns a promise that completes when all ID submodules have filled in their data. `pbjs.refreshUserIds({refresh: false})` will just return the promise without refreshing.

Note that until https://github.com/prebid/Prebid.js/pull/8185 gets merged, GDPR data cannot be resolved without an auction, so this will prevent any ID from being resolved if the consentManagement module is installed and no auction is started. No additional action should be required to fix this besides the merge.

## Other information

Fixes https://github.com/prebid/Prebid.js/issues/5447
Documentation PR: https://github.com/prebid/prebid.github.io/pull/3698

### DIFF
--- a/test/spec/modules/userId_spec.js
+++ b/test/spec/modules/userId_spec.js
@@ -18,9 +18,7 @@ import * as events from 'src/events.js';
 import CONSTANTS from 'src/constants.json';
 import {getGlobal} from 'src/prebidGlobal.js';
 import {
-  requestBidsHook as consentManagementRequestBidsHook,
   resetConsentData,
-  setConsentConfig
 } from 'modules/consentManagement.js';
 import {server} from 'test/mocks/xhr.js';
 import {find} from 'src/polyfill.js';
@@ -51,6 +49,9 @@ import {akamaiDAPIdSubmodule} from 'modules/akamaiDAPIdSystem.js'
 import {kinessoIdSubmodule} from 'modules/kinessoIdSystem.js'
 import {adqueryIdSubmodule} from 'modules/adqueryIdSystem.js';
 import * as mockGpt from '../integration/faker/googletag.js';
+import 'src/prebid.js';
+import {hook} from '../../../src/hook.js';
+import {mockGdprConsent} from '../../helpers/consentData.js';
 
 let assert = require('chai').assert;
 let expect = require('chai').expect;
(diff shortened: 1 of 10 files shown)


patmmccann:
@dgirardi please note the merge conflict

EskelCz:
Thanks for this, looks promising.
One suggestion for readability though, `pbjs.refreshUserIds({refresh: false})` is just what we want, but it's not immediately clear what it does. Can it separated into a new method, something like `pbjs.getUserIdsAsync()`?

dgirardi:
What about adding `getUserIdsAsync` but removing callbacks from the other methods' signatures? Since all of these are part of the public API I'd like to keep changes to a minimum.

EskelCz:
Sounds good, that would leave the current implementations that are working as they are... and whoever is having issues would switch to the async version.

smenzer (changes requested):
Overall this looks good. I'm not an engineer so I can't say that every bit of code is written in the best way, but the logic makes sense and I like that it doesn't break existing implementations.

I would like to see the examples page updated to show how to properly use the `pbjs.getUserIds()` method with the new callback function, as well as any related docs (I know the docs PR is still pending). Once all of that is done, I'm happy to approve. 

We will need another engineer to review and approve since this is core.

test/spec/modules/userId_spec.js
```diff
@@ -148,6 +148,11 @@ describe('User ID', function () {
   });
 
   beforeEach(function () {
+    // TODO: this whole suite needs to be redesigned; it is passing by accident
+    // some tests do not pass if consent data is available
+    // (there are functions here with signature `getId(config, storedId)`, but storedId is actually consentData)
+    // also, this file is ginormous; do we really need to test *all* id systems as one?
+    resetConsentData();
```
smenzer, line 155: I agree completely. a lot of this was from the first iteration of the user id module, so a lot of copy/paste from newer modules and signatures weren't necessarily updated as we made changes.

I think we need to have a few tests that do test *all* id systems for basic processing, but the majority of these tests could probably just be mock id systems to test core functionality

smenzer (approved):

SKOCHERI (approved):

smenzer:
We've got 2 approvals on this, so I'm going to go ahead and merge it in. Thanks for all the work on this one everyone, especially @dgirard and @EskelCz!

EskelCz:
@smenzer @dgirardi Awesome, thanks a lot for quick resolution. Looking forward to testing it out.

PR #8206, opened 2022-03-23
Biddo Bid Adapter: add new bid adapter

llays:
## Type of change

- [ ] Bugfix
- [ ] Feature
- [x] New bidder adapter   
- [ ] Build related changes
- [ ] CI related changes
- [ ] Other

## Description of change

```
{
  bidder: 'biddo',
  params: {
    zoneId: 379783,
  }
}
```

- docs PR https://github.com/prebid/prebid.github.io/pull/3659


modules/biddoBidAdapter.js
```diff
@@ -0,0 +1,92 @@
+import {registerBidder} from '../src/adapters/bidderFactory.js';
+import {BANNER} from '../src/mediaTypes.js';
+
+const BIDDER_CODE = 'biddo';
+const ENDPOINT_URL = 'https://ad.biddo.net/delivery/impress';
```
msm0504, line 5: Calling this endpoint from a test page returned `ERR_NAME_NOT_RESOLVED`. Is this set up yet?

msm0504 (changes requested):

modules/biddoBidAdapter.js
```diff
@@ -0,0 +1,92 @@
...
+        const payload = {
+          ctype: 'div',
+          pzoneid: bidRequest.params.zoneId,
+          width,
+          height,
+        };
+
+        const payloadString = Object.keys(payload).map(k => k + '=' + encodeURIComponent(payload[k])).join('&');
+
+        serverRequests.push({
```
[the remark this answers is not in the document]
llays: Hello! Thank you for review.
Our bid server production deployment is in progress.

modules/biddoBidAdapter.js
llays: Hello! Thank you for review.
Our bid server production deployment is in progress.

modules/biddoBidAdapter.js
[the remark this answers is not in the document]
llays: Our bid server is not operable in terms of multiple sizes per request.

modules/biddoBidAdapter.js
msm0504: @llays This PR can't be approved and merged if the bid server isn't set up yet.

llays: @msm0504 Updated endpoint.

msm0504: @llays I'm still not getting a response from the endpoint.
```
https://ad.adopx.net/delivery/impress?ctype=div&pzoneid=379783&width=300&height=250
```

test/spec/modules/biddoBidAdapter_spec.js
```diff
@@ -0,0 +1,172 @@
+import {expect} from 'chai';
+import {spec} from 'modules/invamiaBidAdapter.js';
```
msm0504, line 2: This should be testing the `biddoBidAdapter`.

Also, will this adapter always have the same logic as `invamia`? If so, would it be possible to make this an alias of `invamia` instead of a separate adapter?

msm0504 (changes requested):

msm0504:
@llays Thanks for updating the test zone id. I now see responses from your server.

However, I'm still not clear on why this needs to be a new adapter. Since all of the code is the same, why can't this be an alias of the `invamia` adapter? The endpoint could then be conditionally set based on a request's `bidderCode`.

llays:
@msm0504 Currently, we've decided to separate business by demand sources and Biddo will have separate team of developers and different business logic on delivery servers. To make it cheaper this adapter is based on invamia's adapter, but in the near future biddo's dev team will go on their own. Thank you!

msm0504 (approved):

PR #8210, opened 2022-03-24
Rubicon Analytics Adapter: pass along gpid

robertrmartinez:
## Type of change
- [X] Other

## Description of change
If gpid is on the adUnit, we now pass it along to our server


harpere (approved):
LGTM

ChrisHuie (approved):

PR #8214, opened 2022-03-24
IAS RTD Module: Custom key values

### LINKED ISSUE #8139: Wrong auctionDelay behavior in realTimeData module
## Type of issue
auctionDelay in realTimeData behaviour isn't correct

## Description
As per doc, auctionDelay should be an upper limit that RTD will wait for module with waitForIt at true for a response, but this is not the case.
Its managed as a fixed time the RTD will wait, also if the modules get the response before.

## Steps to reproduce
set the following config for RTD:
`realTimeData = {
                        auctionDelay: 50000,
                        dataProviders: [
                            {
                                name: "ias",
                                waitForIt: true,
                                params: {
                                    pubId: "929966"
                                }
                            }
                        ]
                    }`

set bidders timeout to 10000 and make a call.
Prebid will make IAS RTD module call, IAS give response after around 150ms, RTD module will wait for 50 seconds, after 10 seconds prebid will trigger timeout.

### Expected results
RTD IAS module will make call, after having the response bidders start make request without waiting the 50sec auctionDelay.
## Platform details
Tested with Prebid v5.20.2 and v6.13.0

rbarbini-ias:
## Type of change
- [x] Bugfix
- [x] Feature
- [ ] New bidder adapter   
- [ ] Build related changes
- [ ] CI related changes
- [ ] Other

## Description of change

- fixes #8139 
- adds the possibility for the client to rename a key-value


rbarbini-ias:
Uhm, the test that is failing is not relative to the IAS RTD adapter.

harpere (approved):

PR #8216, opened 2022-03-25
Prebid Core: Support for multiple bidder codes from a single adapter

pm-azhar-mulla:
## Type of change

- [x] Feature

## Description of change

With respect to draft [8129](https://github.com/prebid/Prebid.js/issues/8129). We have made changes to the bidderFactory.js.
 After receiving the bids from adapters, we look for eligible bids depending upon values of `allowAlternateBidderCodes ` and `allowedAlternateBidderCodes ` values in `bidderSettings`.

- If `allowAlternateBidderCodes ` is false for given bidder, we log warning on console and ignore the bid.

- If `allowAlternateBidderCodes ` is true and `allowedAlternateBidderCodes` is specified in terms of an array, we look for respective bidderCode in that array. If we find the bidderCode we continue otherwise again we log the warning on console and ignore the bid

- If `allowAlternateBidderCodes ` is true and `allowedAlternateBidderCodes` is not specified or value for `allowedAlternateBidderCodes` is `['*']` then we don't apply any restrictions.

After bid is considered as eligible, we add `adapterCode` property to the bid which represents name of the adapter through which the bid was initiated.

```
pbjs.bidderSettings = {
    standard: {
         allowAlternateBidderCodes: false,    // the default if not specified for a specific adapter
         [...]
    },
    bidderA: {
        allowAlternateBidderCodes: true,
        allowedAlternateBidderCodes: ["bidderX", "bidderY"], // bidderA may add X or Y if it wishes
        [...]
    },
    bidderB: {
        allowAlternateBidderCodes: true,
        allowedAlternateBidderCodes: ["*"],        // the default allows this bidder to respond with any biddercodes
        [...]
    }
}
```

@bretg @patmmccann This is an initial draft, we are in process of testing and adding unit test cases. Please share your views on the same.

### Checklist for the above feature

- [x]  Make changes in bidderFactory.js
- [x]  Make changes in bidderSetting.js
- [ ]  Make related changes in core files


src/adapters/bidderFactory.js
```diff
@@ -250,6 +257,17 @@ export function newBidder(spec) {
     }
   });
 
+  function isUnknownBidder(unknownBidder, bidderCode) {
+    let currentBidderSettings = !!bidderSettings.get(bidderCode, 'allowUnknownBidderCodes');
+    let unknownBiddersList = bidderSettings.get(bidderCode, 'allowedUnknownBidderCodes');
+    if (!!unknownBidder && bidderCode !== unknownBidder) {
+      if (!currentBidderSettings || (currentBidderSettings && isArray(unknownBiddersList) && (unknownBiddersList[0] !== '*' && !unknownBiddersList.includes(unknownBidder)))) {
```
pm-harshad-mane, line 264: please fix the LGTM warning

pm-azhar-mulla: Fixed

src/adapters/bidderFactory.js
```diff
@@ -250,6 +257,17 @@ export function newBidder(spec) {
     }
   });
 
+  function isUnknownBidder(unknownBidder, bidderCode) {
```
dgirardi, line 260: I am not sure this works with aliases, nor that it can achieve the aim of https://github.com/prebid/Prebid.js/pull/8193 (batching requests from multiple bidders into one network call). Looking at how [aliases work](https://github.com/prebid/Prebid.js/blob/64bd041fa29e69605086dc094dfc525d9a4e0242/src/adapterManager.js#L479), I think they'd look as completely distinct adapters from here. The request batching seems like a bigger project unless I am misreading the code.

I think it should be a requirement (and also a good test case) that aliases set up both by the bid adapters and the publisher should *not* be flagged as "unknown" here, regardless of the `allowUnknown` settings.

pm-harshad-mane:
Hi @dgirardi ,
PubMatic will revert the #8193 in the next PR. 
With this change, if the publisher allows through the settings then bidders will be able to submit bids with new bidder-names, so PubMatic will be able to submit bids as GroupM as well, hence PubMatic will no longer need to support the GroupM as an alias of PubMatic.

pm-harshad-mane:
@pm-azhar-mulla  can you please share the findings of the scenario mentioned by @dgirardi ?

src/adapters/bidderFactory.js
pm-azhar-mulla: Hi @dgirardi ,
When a publisher sets an alias, as you mentioned `pbjs.aliasBidder('pubmatic', 'foobar')`.
Inputs to `isUnknownBidder` function are `bid.bidderCode as undefined` and `bidRequest.bidder as foobar` and even if adapter sends `bid.bidderCode as foobar` our condition inside `isUnknownBidder` function handles this scenario. 
Since it checks for following conditions in sequence

1. Valid unknownBidder
2. Valid requested bidder
3. if unknownBidder is not matching requested bidder
4. Then it checks for bidderSettings for unknownBidder

In case of aliases, the 3rd condition will not satisfy and function will return `false` and continue with bid from alias.

I hope it makes sense.

pm-azhar-mulla: Hi @dgirardi can you please share your thoughts on this.

src/adapters/bidderFactory.js
```diff
@@ -250,6 +255,17 @@ export function newBidder(spec) {
...
```
dgirardi, line 258: Also, can I suggest different parameter names here - `requestBidder` and `responseBidder` instead of `unknownBidder` and `bidderCode`?

bretg:
Aliases are declared valid names that should be valid for bid adapters. This feature is about **unexpected** names in the response.

So in the example modifed from above, both scenarios are valid:

```
pbjs.aliasBidder('pubmatic', 'foobar');
pbjs.bidderSettings = {
     pubmatic: {
        allowUnknownBidderCodes: true,
        allowedUnknownBidderCodes: ['unexpected']
     }
}
pbjs.requestBids({
   adUnits: [{
       code: 'div-1'
       bids: [{
             bidder: 'foobar'    // the pubmatic adapter decides to set `bidResponse.bidderCode = 'groupm'`
        }]
   }, {
       code: 'div-2',
       bids: [{
              bidder: 'groupm'  // the groupm (pubmatic) adapter decides to set `bidResponse.biddercode = 'pubmatic'`
       }]
   }]
})
```

For div-1, PBJS calls the pubmatic adapter due to the dynamic alias from foobar. Pubmatic sets "groupm" as the biddercode, which is considered ok because the adapter declares the alias.

For div-2, PBJS calls the pubmatic adapter due to the hard-coded alias groupm. Pubmatic sets "groupm" as the biddercode, which is considered ok because that's the adapter name.

Pubmatic could also choose to return "foobar" as a biddercode because the page set that up as a dynamic alias. Pubmatic could also choose to return "unexpected" as a biddercode because the page specifically set that up as an allowed code.

Here are some things pubmatic couldn't return as a biddercode without being rejected: rubicon, unknown, avengers.

So PBJS-core should allow the following as valid biddercodes:

1) the actual biddercode
2) hard-coded aliases
3) page-defined aliases
4) bidderSettings.BIDDER.allowedUnknownBidderCodes

pm-azhar-mulla:
Hi @bretg @dgirardi ,
I have added additional check in [here](https://github.com/pm-azhar-mulla/Prebid.js/blob/unknown-bidder-support-02/src/adapters/bidderFactory.js#L261) for checking aliases.

pm-azhar-mulla:
Hi @bretg @dgirardi ,
I have added additional condition in [bidderSetting's](https://github.com/pm-azhar-mulla/Prebid.js/blob/unknown-bidder-support-02/src/bidderSettings.js#L22) get function, where in after searching for bidder level setting, it will look for adapter level setting before picking up standard setting. This was the impact mentioned in [draft](https://github.com/prebid/Prebid.js/issues/8129).
I am working on making changes in core files, where `bidderSettings.get` function is used. 
Now, along with `bidderCode`, user can also send `adapterCode` as third optional parameter. 
Priority of the search in `bidderSettings.get` function will be: 
- bidder level
- adapter level
- standard

src/adapters/bidderFactory.js
pm-azhar-mulla: Hi @dgirardi,
Thank you for your suggestion. I have made the changes. :)

src/adapters/bidderFactory.js
```diff
@@ -235,6 +236,10 @@ export function newBidder(spec) {
         onBid: (bid) => {
           const bidRequest = bidRequestMap[bid.requestId];
           if (bidRequest) {
+            if (isUnknownBidder(bid.bidderCode, bidRequest.bidder, bid.adapterCode)) {
```
dgirardi, line 239: I don't think it makes sense to look for `bid.adapterCode` - we shouldn't need to ask adapters to set that. Core can determine`adapterCode` by looking for the root of the alias tree:

1. set `adapterCode = bidRequest.bidderCode`
2. if `aliasRegistry[adapterCode]`, set `adapterCode = aliasRegistry[adapterCode]` and goto 1
3. set `bid.adapterCode = adapterCode`

I think this would work because the aliasing logic should prevent cycles.

src/adapters/bidderFactory.js
```diff
@@ -250,6 +255,17 @@ export function newBidder(spec) {
     }
   });
 
+  function isUnknownBidder(responseBidder, requestBidder, adapterCode) {
+    let currentBidderSettings = !!bidderSettings.get(requestBidder, 'allowUnknownBidderCodes', adapterCode || null);
+    let unknownBiddersList = bidderSettings.get(requestBidder, 'allowedUnknownBidderCodes', adapterCode || null);
+    if (!!responseBidder && !adapterManager.aliasRegistry[responseBidder] && !!requestBidder && requestBidder !== responseBidder) {
```
dgirardi, line 261: I read this block as:

if both `responseBidder` and `requestBidder` are defined and different from each other, and `responseBidder` is not an alias, then check that `responseBidder` is allowed by `bidderSettings`.

But this would allow for example the `pubmatic` adapter to set any alias as the `responseBidder`, like `appnexusAst` (an alias of `appnexus`). 

I think the logic should be: if `responseBidder` is defined and different from `requestBidder`, check that its adapter code is the same as the adapter code of `requestBidder`, or that it is allowed by `bidderSettings`.

src/bidderSettings.js
```diff
@@ -16,9 +16,12 @@ export class ScopedSettings {
    * @param path {String}
    * @returns {*}
    */
-  get(scope, path) {
+  get(scope, path, parentScope) {
```
dgirardi, line 19: Can you explain the new argument in the method's jsdoc?

src/adapters/bidderFactory.js
pm-azhar-mulla: Hi @dgirardi ,
Lets consider 2 scenarios
1. Without alias
2. With alias

***1. Without alias***
```
pbjs.bidderSettings = {
     pubmatic: {
        allowUnknownBidderCodes: true,
        allowedUnknownBidderCodes: ['unexpected']
     }
}
pbjs.requestBids({
   adUnits: [{
       code: 'div-1'
       bids: [{
             bidder: 'pubmatic'    // the pubmatic adapter decides to set `bidResponse.bidderCode = 'unexpected'`
        }]
   }]
})
```
In above case `pubmatic` is bidder which will respond with bidderCode as `unexpected `
So input to `isUnknownBidder` will be `responseBidder as unexpected `, `requestBidder as pubmatic` and `adapterCode as pubmatic`
Here, the function will find entry for `pubmatic` and will get `unexpected ` in the `allowedUnknownBidderCodes` array so it will return false and continue with the bid.


***2. With alias***
```
pbjs.aliasBidder('pubmatic', 'foobar');
pbjs.bidderSettings = {
     pubmatic: {
        allowUnknownBidderCodes: true,
        allowedUnknownBidderCodes: ['unexpected']
     }
}
pbjs.requestBids({
   adUnits: [{
       code: 'div-1'
       bids: [{
             bidder: 'foobar'    // the pubmatic adapter decides to set `bidResponse.bidderCode = 'unexpected'`
        }]
   }]
})
```
In above case `foobar` is alias bidder of `pubmatic` adapter which will respond with bidderCode as `unexpected `
So input to `isUnknownBidder` function will be `responseBidder as unexpected `, `requestBidder as foobar` and `adapterCode as pubmatic`
Here, the function will first search for `bidderSettings` of foobar but couldn't find any so it searches for `bidderSettings` of its adapter i.e `pubmatic`. Here the entry is found and `unexpected ` is present in the `allowedUnknownBidderCodes` array so it will return false and continue with the bid.

To search for `bidderSettings` of `adapterCode` if `bidderSettings` are not present for given `bidderCode`, i have added additional parameter `adapterCode`.

src/adapters/bidderFactory.js
pm-azhar-mulla: Using `bidRequest.bidder` makes sense. Thank you :)

src/bidderSettings.js
pm-azhar-mulla: Consider below scenario
```
pbjs.aliasBidder('pubmatic', 'foobar');
pbjs.bidderSettings = {
     pubmatic: {
        allowUnknownBidderCodes: true,
        allowedUnknownBidderCodes: ['unexpected']
     }
}
```

Here when we search for bidderSettings of `foobar` we wont find it, since `bidderSettings` are applied to its adapter i.e. `pubmatic`. To make adapter settings accessible to `bidder` i have made this change.
As explained in my comment yesterday the priority of `bidderSettings.get` function will be

1. Bidder level
2. Adapter level
3. Standard

Same thing was highlighted in IMPACTS of [8129](https://github.com/prebid/Prebid.js/issues/8129)

src/adapters/bidderFactory.js
pm-azhar-mulla: HI @dgirardi ,
Got your point. Made the related changes and also added unit test cases for it too. Please review.

src/bidderSettings.js
dgirardi: yes that makes sense, what I meant is - can you update the method's jsdoc comment? so that someone who didn't look at this pr knows what `parentScope` is :)

src/adapters/bidderFactory.js
```diff
@@ -235,6 +236,13 @@ export function newBidder(spec) {
         onBid: (bid) => {
           const bidRequest = bidRequestMap[bid.requestId];
           if (bidRequest) {
+            if (!bid.adapterCode) {
+              bid.adapterCode = adapterManager.aliasRegistry[bidRequest.bidder] || bidRequest.bidder;
```
dgirardi, line 240: Unfortunately this is not enough, because you can alias an alias (this works: `pbjs.aliasBidAdapter('pubmatic', 'foobar'); pbjs.aliasBidAdapter('foobar', 'barfoo')`, and results in `aliasRegistry['barfoo'] === 'foobar'`)

Because of that I think you do need to walk up the tree as I outlined above. It's probably best to put it close to the aliasing logic, I suggest adding `getAdapterCode(bidderCode)` in `adapterManager.js`.

src/adapters/bidderFactory.js
```diff
@@ -235,6 +236,13 @@ export function newBidder(spec) {
         onBid: (bid) => {
           const bidRequest = bidRequestMap[bid.requestId];
           if (bidRequest) {
+            if (!bid.adapterCode) {
```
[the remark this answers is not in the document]
pm-azhar-mulla: Hi @dgirardi,
Done the changes.

src/adapters/bidderFactory.js
pm-azhar-mulla: Added a function `getAdapterCode` in `adapterManager.js`

pm-azhar-mulla: `getAdapterCode` function will traverse through the `aliasRegistry` to search for respective adapter. If return the `adapterCode` if it finds the entry in `aliasRegistry` or else it returns `null`

src/adapters/bidderFactory.js
```diff
@@ -250,6 +256,17 @@ export function newBidder(spec) {
     }
   });
 
+  function isUnknownBidder(responseBidder, requestBidder, adapterCode) {
+    let currentBidderSettings = !!bidderSettings.get(requestBidder, 'allowUnknownBidderCodes', adapterCode || null);
+    let unknownBiddersList = bidderSettings.get(requestBidder, 'allowedUnknownBidderCodes', adapterCode || null);
+    if (!!responseBidder && adapterManager.getAdapterCode(responseBidder) !== adapterCode && !!requestBidder && requestBidder !== responseBidder) {
+      if (adapterManager.aliasRegistry[responseBidder] || !currentBidderSettings || (isArray(unknownBiddersList) && (unknownBiddersList[0] !== '*' && !unknownBiddersList.includes(responseBidder)))) {
```
dgirardi, line 263: Why the first condition - if the reponse bidder is an alias, then it is always unknown?

test/spec/unit/core/bidderFactory_spec.js
```diff
@@ -1022,6 +1023,166 @@ describe('validate bid response: ', function () {
...
+
+      const bidder = newBidder(spec);
+      spec.interpretResponse.returns(bids1);
+      bidder.callBids(bidRequest, addBidResponseStub, doneStub, ajaxStub, onTimelyResponseStub, wrappedCallback);
+
+      expect(addBidResponseStub.calledOnce).to.equal(false);
+      expect(logWarnSpy.callCount).to.equal(1);
+    });
+
+    it('should not accept the bid, when bidder is an alias of the other adapter and allowUnknownBidderCodes is true', function () {
```
dgirardi, line 1159: Same question - why? If I set `{allowUnknownBidderCodes: true, allowedUnknownBidderCodes: ['*']}` like this test does, shouldn't every bid always be accepted?

src/adapters/bidderFactory.js
```diff
@@ -250,6 +256,17 @@ export function newBidder(spec) {
     }
   });
 
+  function isUnknownBidder(responseBidder, requestBidder, adapterCode) {
+    let currentBidderSettings = !!bidderSettings.get(requestBidder, 'allowUnknownBidderCodes', adapterCode || null);
```
dgirardi, line 260: Sorry for not mentioning this earlier - I don't think `currentBidderSettings` is a good name; `allowUknownBidders` or similar would be a better one.

pm-azhar-mulla: Done :)

test/spec/unit/core/bidderFactory_spec.js
dgirardi: I disagree - on two fronts: first, we already have bidders that want to bid on behalf of others, so I don't think it's that implausible that they may bid for an alias of them. Second, I don't think we need to contradict what the publisher has set; if they explicitly want to allow `appnexusAST`, we should allow it, I don't see a benefit in adding exceptions to their rules.

src/adapters/bidderFactory.js
```diff
@@ -235,6 +236,11 @@ export function newBidder(spec) {
         onBid: (bid) => {
           const bidRequest = bidRequestMap[bid.requestId];
           if (bidRequest) {
+            bid.adapterCode = adapterManager.getAdapterCode(bidRequest.bidder);
```
dgirardi, line 239: this should default to `bidRequest.bidder` - if the request is for `pubmatic`, the adapter code should be `pubmatic`.

dgirardi:
@pm-azhar-mulla, there's more recent discussion in https://github.com/prebid/Prebid.js/issues/8129, and it sounds like we want to change how to deal with aliases again (sorry!). I think it's best to put this on hold until we agree on the requirements.

mike-chowla:
Looking at @bretg 's latest comments on #8129, it appears we've been able to finalize on requirements.  

@dgirardi do you agree that we are good from a requirements point of view now?

> @pm-azhar-mulla, there's more recent discussion in #8129, and it sounds like we want to change how to deal with aliases again (sorry!). I think it's best to put this on hold until we agree on the requirements.

dgirardi:
Yes, although I would have said the same before the recent changes :)

pm-azhar-mulla:
Hi @dgirardi ,
I have made the changes as per the updated proposal.

src/adapterManager.js
```diff
@@ -575,4 +575,11 @@ adapterManager.callBidderError = function(bidder, error, bidderRequest) {
   tryCallBidderMethod(bidder, 'onBidderError', param);
 };
 
+adapterManager.getAdapterCode = function(bidderCode) {
```
dgirardi, line 578: this is no longer needed; the name is also confusing as `bid.adapterCode` now means something else. Probably best to remove.

dgirardi (commented):
The logic looks good but I think it's incomplete: the proposal includes falling back to `adapterCode`  for all bidderSettings lookups, which are spread throughout the codebase. This PR adds the functionality in bidderSettings but none of the callers have been updated,  so things like `bidCpmAdjustment` for alternate bidders would not "see" the adapter code.

I am not sure we want to do that separately because it would then be a breaking change.

pm-harshad-mane:
An Org can be seen in Prebid as,
- an original registered bidder, example pubmatic-bidder, `pubmatic`
- a registered alias, example alias `pubmatic-2`
- alternate bidder, example GroupM, `groupm`

I would like to keep things simple. 
For a publisher, it will be difficult to define the bidder settings. The pub will need to know which role an Org has chosen.
Irrespective of the role an org has chosen from the above-mentioned roles, if the specific bidder setting is present for the org name then that should be applied, otherwise, the standard bidder setting should be used. 
If a pub wants to apply partial or full bidder settings of `pubmatic` to `groupm` then it can be done by defining the config in JS variables and using the variables appropriately.

mike-chowla:
My preference is keep things as simple as possible for the user to understand.  It feels complicated to me for the user to understand that there's a relationship between bidder codes that can cause settings such as `bidCpmAdjustment` to be applied from bidder code to another, especially when it doesn't happen other circumstances where there's a relationship.

If I create an alias of `bidderA` called `bidderB`, I have to create `bidCpmAdjustment` for `bidderB` is if I want to adjust the bid CPM for `bidderB`.  

Seems to me the simplest way conceptually to handle bids with a different bidder code is to only apply bid settings for the bidder code attached to the bid.  Targeting key settings should work the same way.  If you want a custom targeting key for `bidderB`, you need to set it on `bidderB`

modules/prebidServerBidAdapter/index.js
```diff
@@ -918,8 +918,8 @@ Object.assign(ORTB2.prototype, {
         (seatbid.bid || []).forEach(bid => {
           let bidRequest = this.getBidRequest(bid.impid, seatbid.seat);
           if (bidRequest == null) {
-            if (!s2sConfig.allowUnknownBidderCodes) {
-              logWarn(`PBS adapter received bid from unknown bidder (${seatbid.seat}), but 's2sConfig.allowUnknownBidderCodes' is not set. Ignoring bid.`);
+            if (!s2sConfig.allowAlternateBidderCodes) {
```
dgirardi, line 921: this is a breaking change - and not how I interpret the proposal. However, I realize now that how I'd interpret the proposal is not possible to implement.

Proposal says:

> 3. Update the pbsBidAdapter to look in bidderSettings as well as s2sConfig and enforce allowedAlternateBidderCodes.

I read it as "the pbs adapter should also enforce `bidderSettings[].allowAlternateBiddercodes`, and default to `s2sConfig.allowUnknownBidderCodes`". However, the pbs adapter cannot know what the adapter code is - it could be any of the bidders in the request - so what should it look up in `bidderSettings`? @bretg, @patmmccann what is your interpretation?

dgirardi:
@pm-azhar-mulla, if you revert the changes to `bidderSettings` (to remove the `parentScope` logic that we are not using) I'll approve this PR; hopefully by the time the second review completes we'll all have agreed that we don't need that.

The missing `s2sConfig` piece I'm not sure about, but it looks to me like something that could be added later.

pm-azhar-mulla:
Hi @dgirardi,
I have reverted the changes for `bidderSetting's parentScope`

dgirardi (approved):
LGTM, with the caveats above - we are reducing the scope to just enforcement of `allowAlternate` bidderSettings config, leaving out proposed changes to `adapterCode` fallbacks and the PBS adapter.

patmmccann:
@pm-harshad-mane we need your second review and we need docs

pm-harshad-mane:
@pm-azhar-mulla , Please link the documentation PR to this PR.

Let's say `pubmatic` is configured on the server-side in PBS, will the `pubmatic` bidder will get to know the config values for `allowedAlternateBidderCodes`? I think we should pass the newly added config params to PBS bidders as well. Create a separate PR if many changes are needed.

dgirardi:
@pm-harshad-mane, Prebid Server is not involved at all here - I don't understand what you mean. This affects exclusively client-side adapters whose backends want to bid on behalf of other bidder codes.

For PBS, there's a separate `s2sConfig.allowUnknownBidderCodes` flag that allows it to bid with any bidder. But the way you'd use that (as an SSP) would be entirely different - you'd ask clients to use a PBS instance that bids that way (I am not sure how that'd be implemented server-side). There would be no `bidderSettings` and no bidder code list.

patmmccann:
https://github.com/prebid/Prebid.js/blob/c8c326f33980f451c33577848f5ac975e73a65e7/modules/prebidServerBidAdapter/index.js#L921 for reference

pm-harshad-mane (approved):
LGTM

PR #8219, opened 2022-03-25
Prebid core & PBS adapter: Feature tags and optional compilation of native support

dgirardi:
## Type of change
- [x] Feature

## Description of change

This adds feature flags that can remove portions of the codebase at build time (see https://github.com/prebid/Prebid.js/issues/7867). 

The first (and only) flag defined here is "NATIVE", which is the easiest (and smallest) to manage; only core and PBS are included - with the expectation that adapters will over time mark their feature-specific code as is done here.

This is the effect on the bundle size:

```
demetrio@pbws:~/src/Prebid.js$ gulp build-bundle-prod --modules=prebidServerBidAdapter > /dev/null 2>&1 && ls -l build/dist/prebid.js
-rw-rw-r-- 1 demetrio demetrio 176791 Jan 27 11:21 build/dist/prebid.js
demetrio@pbws:~/src/Prebid.js$ gulp build-bundle-prod --modules=prebidServerBidAdapter --disable=native > /dev/null 2>&1 && ls -l build/dist/prebid.js
-rw-rw-r-- 1 demetrio demetrio 168356 Jan 27 11:21 build/dist/prebid.js
```

A readable comparison of the output: https://gist.github.com/dgirardi/c506a2aa243125cc2a8bdcfa27e10d7a/revisions

## Other information

**Testing**: the strategy is, for now, to add an additional test run with all features disabled. Once we have more flags defined I intend to revisit this - it will probably be too expensive to run the whole test suite on all feature combinations, but I am optimistic that we can find a good heuristic to identify "relevant" tests and be exhaustive only on those (most tests do not interact with the code being removed).

**Compilation**: I have decided to use our Babel plugin to resolve feature flags (as opposed to webpack's `DefinePlugin` or other options like terser arguments). The reason is that it works better for npm consumers: no change is required to keep the build working, and there's no risk of our globals (i.e. `FEATURES`) clashing with their namespace.

Documentation PR: TBD


patmmccann:
@pm-harshad-mane something to consider for the pubmatic adapter?

snapwich (approved):
this is awesome. the PR is a bit out-of-date now and as you stated i think there's still some things to figure out regarding testing but overall i really like this approach of using dead-code elimination as well as using the babel plugin. 

really nice

mmoschovas (approved):
This looks good. One thing I am noticing while going through this (not to do with this PR but wanted to make mention of it) is that the Object.assign within PBS bid adapter is getting excessively long i.e. 500+ lines of code. We should really think about breaking this up a bit into some smaller functions to reduce complexity and make this more readable.

dgirardi:
@mmoschovas regarding the complexity, your input is welcome in https://github.com/prebid/Prebid.js/issues/8562 - but overall I agree :)

patmmccann:
@dgirardi this is released! Very exciting. Lmk when docs avail for review

dgirardi:
@shahinrahbariasl how are you running gulp for development? I personally found all options too infuriating to use when I first dove in, so I'm not sure what people use in the wild. I work with `gulp serve-and-test --file <spec_file>` (watch mode is always on for it).

shahinrahbariasl:
I usually do `gulp test --file <path-to-test-file> --watch`. I can try the above too.

muuki88:
We use an webpack build with prebid as a NPM dependency. This change breaks our js bundle at runtime with

```
secureCreatives.js:25 Uncaught ReferenceError: FEATURES is not defined
    at ./node_modules/prebid.js/src/secureCreatives.js (secureCreatives.js:25:1)
    at __webpack_require__ (bootstrap:19:1)
    at ./node_modules/prebid.js/src/prebid.js (polyfill.js:18:2)
    at __webpack_require__ (bootstrap:19:1)
    at make namespace object:7:1
    at moli.mjs:41251:3
    at moli.mjs:41253:12
```

As the `FEATURE` variable comes out of thin air :P

wissamabyad:
We're running into the same issue, and can't upgrade beyond version 7.1, and the error is still not handled in 7.25. 

```
secureCreatives.js?4527:24 Uncaught (in promise) ReferenceError: FEATURES is not defined
    at eval (secureCreatives.js?4527:24:1)
    at ../../node_modules/prebid.js/src/secureCreatives.js (prebid.js:413:1)
    at __webpack_require__ (main.js:1349:42)
    at eval (prebid.js:12:78)
    at ../../node_modules/prebid.js/src/prebid.js (prebid.js:389:1)
    at __webpack_require__ (main.js:1349:42)
    at eval (library.ts:5:67)
    at ../../lib/advertising/Prebid/library.ts (prebid.js:23:1)
    at Function.__webpack_require__ (main.js:1349:42)
```

dgirardi:
@wissamabyad, are you passing prebid sources through our babel configuration as [in the README](https://github.com/prebid/Prebid.js#usage-as-a-npm-dependency)?

PR #8222, opened 2022-03-28
Missena Bid Adapter - allow custom endpoint.

pdamoc:
## Type of change

- [x] Other

## Description of change
The change allows custom endpoints to be configured when using the bid adapter. This is needed to allow internal testing of our staging server and it doesn't affect users of the adapter in any way.


osazos (approved):
LGTM

Issue #7796, opened 2021-12-02
Prebid 7 - a proposed list of breaking changes


patmmccann:
## Type of issue
A Proposed and working list of breaking changes to include in Prebid 7 - target release Late May 2022

## Complete:
delete userid targeting module #8226 
delete nextroll https://github.com/prebid/Prebid.js/pull/8150
Remove Appnexus analytics #8330 
delete floc #8225 
https://github.com/prebid/Prebid.js/pull/7522 Confirmation from the Yahoo team indicates to remove onevideo (aol in next version)
Adapters accepting interstitial flag must accept ortb2 instl param https://github.com/prebid/Prebid.js/issues/5917
Remove dmx #8417
change defaults on #8129 
delete halo as per https://github.com/prebid/Prebid.js/pull/7938
https://github.com/prebid/Prebid.js/issues/7651 site.content.data on specific auctions
Fold trustx into Grid #8321 
remove adlive bid adapter
delete sortable adapters #8263 
https://github.com/prebid/Prebid.js/pull/8437  -- delete appnexus aliases
#8279 bid modules will not be able to use storage manager unless whitelisted by pubs in config
Remove akamai id module
fix bcats to read from ortb2 everywhere #7925 
#7853 default usp to iab 
https://github.com/prebid/Prebid.js/pull/8235 remove stored auction response
deprecate getLegacyFpd #8283  

Adapters accepting position must accept it as per conventions defined at https://github.com/prebid/Prebid.js/issues/5578

Adapters known to be using http1 bid endpoints in violation of bidder module rules will get warnings in their adapter documentation; user sync endpoints in http1 will not incur the warning at this time. 

## PR submitted:

Drop tcf1 https://github.com/prebid/Prebid.js/issues/7781

Finish up referrer changes https://github.com/prebid/Prebid.js/issues/5637 #8450 

prevent Prebid from loading twice - https://github.com/prebid/Prebid.js/issues/8374

## To do:
https://github.com/prebid/Prebid.js/pull/7954 update pbsBidAdapter to use imp.ext.prebid.bidder.BIDDER

Update all documentation related to FPD to follow up #8283  and #7651

josepowera:
Proposal
Make universal OpenRTB and ID module that BidAdapters and ID providers can "inherit"


Make a universal  OpenRTB 2.5 bidding module, that minimizes what bidder adapter has to do.... adapter would only define:
-httpEndpoint for bidding,
-httpEndpoint for cookieSyncs
[...]

PR #8229, opened 2022-03-28
Prebid 7: Allow USP module to work without explicit configuration

patmmccann:
Part of #7796


modules/consentManagement.js
```diff
@@ -470,6 +470,7 @@ function exitModule(errMsg, hookConfig, extraArgs) {
 export function resetConsentData() {
   consentData = undefined;
   userCMP = undefined;
+  consentTimeout = undefined;
```
patmmccann, line 473: note i added this line because it not getting reset in USP was causing issues with the tests; it seemed simple to fix here as well at the same time.

modules/consentManagementUsp.js
```diff
@@ -289,6 +289,7 @@ function exitModule(errMsg, hookConfig, extraArgs) {
 export function resetConsentData() {
   consentData = undefined;
   consentAPI = undefined;
+  consentTimeout = undefined;
```
patmmccann, line 292: this one stumped me, hat tip to @dgirardi for diagnosing why my 'should disable' test was failing.

modules/consentManagementUsp.js
```diff
@@ -298,18 +299,21 @@ export function resetConsentData() {
  */
 export function setConsentConfig(config) {
   config = config && config.usp;
-  if (!config || typeof config !== 'object') {
-    logWarn('consentManagement.usp config not defined, exiting usp consent manager');
+  if (config && config.disable === true) {
```
patmmccann, line 302: this allows a publisher to restore the old behavior with configuration.

Fawke (approved):
LGTM

modules/consentManagementUsp.js
dgirardi: I don't think this is equivalent - it does not mean "disable" but "ignore". Let's say I do

```
pbjs.setConfig({
   consentManagement: {
      gpdr: {...}
   }
});
pbjs.setConfig({
  consentManagement: {
    usp: {
       disable: true
    }
  }
});
```

the first call would activate USP with the default values, the second would log a warning but otherwise do nothing - USP would still be enabled.

You could `resetConsentData` (and remove the `requestBidsHook`), but I can't think of another example where we ask for explicit *de*activation. To me it seems cleaner if we either require explicit activation (maybe that means accepting `usp: {}` to indicate the default), or just make it clear that if you install the module, it's always active.

patmmccann: if what you say is true, how is this test passing

it('should not run if disable is true', function () {
        setConsentConfig({usp: {disable: true}});
        expect(consentAPI).to.be.undefined;
        expect(consentTimeout).to.be.undefined;
        sinon.assert.callCount(utils.logWarn, 1);
      });

patmmccann:
Thanks for your review and changes @dgirardi ; I added docs PR as well

PR #8237, opened 2022-03-30
Glimpse: update api and request shape, optimize and refactor

eddieyu1998:
## Type of change

- [x] Bugfix
- [x] Feature
- [ ] New bidder adapter   
- [ ] Build related changes
- [ ] CI related changes
- [ ] Other

## Description of change

- Update Glimpse prebid api endpoint
- Pass gvlid to storage manager for gdpr use cases
- Include consent info in query params for better efficiency and privacy on the handling of user data
- Update the request body shape to optimize the size of the request
- update fpd shape and optimize fpd object
- Add appendQueryParam and optimizeObject util functions
- Remove unnecessary util functions
- Refactor and rename of some util functions for improved simplicity
- Rename variables for clarity
- Update jsdoc comments
- Reformat the files
- Update test spec: add fpd unit test and remove unnecessary tests

---

## contact email of the adapter’s maintainer
- eddie.yu@glimpseprotocol.io

## Related PR on prebid.github.io

- https://github.com/prebid/prebid.github.io/pull/3669

### DIFF
--- a/modules/glimpseBidAdapter.js
+++ b/modules/glimpseBidAdapter.js
@@ -1,18 +1,27 @@
-import { BANNER } from '../src/mediaTypes.js'
-import { config } from '../src/config.js'
-import { getStorageManager } from '../src/storageManager.js'
-import { isArray } from '../src/utils.js'
-import { registerBidder } from '../src/adapters/bidderFactory.js'
+import { registerBidder } from '../src/adapters/bidderFactory.js';
+import { config } from '../src/config.js';
+import { BANNER } from '../src/mediaTypes.js';
+import { getStorageManager } from '../src/storageManager.js';
+import {
+  isArray,
+  isEmpty,
+  isEmptyStr,
+  isStr,
+  isPlainObject,
+} from '../src/utils.js';
 
-const GVLID = 1012
-const BIDDER_CODE = 'glimpse'
-const storageManager = getStorageManager({bidderCode: BIDDER_CODE})
-const ENDPOINT = 'https://api.glimpsevault.io/ads/serving/public/v1/prebid'
+const GVLID = 1012;
+const BIDDER_CODE = 'glimpse';
+const storageManager = getStorageManager({
+  gvlid: GVLID,
+  bidderCode: BIDDER_CODE,
+});
+const ENDPOINT = 'https://market.glimpsevault.io/public/v1/prebid';
 const LOCAL_STORAGE_KEY = {
   vault: {
     jwt: 'gp_vault_jwt',
   },
-}
+};
 
 export const spec = {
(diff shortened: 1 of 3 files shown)


eddieyu1998:
After some digging, I've noticed that the reason why gulp lint is failing is probably because of the use of `??` operator in our bid adapter. The version of gulp-eslint used by Prebid currently is v4, which looking at its source code doesn't seem to support the `??` operator.

I'm wondering if Prebid can bump the version of gulp-eslint to a newer version to support new features in Javascript?

eddieyu1998:
I've made a change to replace all the `??` to `||` for now

modules/glimpseBidAdapter.js
```diff
@@ -1,230 +1,201 @@
...
-    return {
-      consentString: gdprConsent.consentString || '',
-      vendorData: gdprConsent.vendorData || {},
-      gdprApplies: hasGdprApplies ? gdprConsent.gdprApplies : true,
-    }
+  if (gdprApplies(bidderRequest)) {
+    const consentString = bidderRequest.gdprConsent.consentString;
+    url = appendQueryParam(url, 'reg', 'gdpr');
+    url = appendQueryParam(url, 'cs', consentString);
+  } else if (ccpaApplies(bidderRequest)) {
```
msm0504, line 111: Is it possible for both GDPR and CCPA to apply? If so, does your endpoint support multiple `reg` values?

msm0504 (changes requested):

modules/glimpseBidAdapter.js
eddieyu1998: Is it a requirement that we have to support both GDPR and CCPA when they both applies? We think it'd be unlikely that a user would be classified as under both GDPR and CCPA at the same time, and we currently only consider the GDPR consent if that's the case.

msm0504: I checked with other reviewers and it's ok to only consider GDPR in that case.

msm0504 (approved):

modules/glimpseBidAdapter.js
patmmccann: every user in California browsing a website in the EU has both apply. To test this, set your IP to California and go to Guardian.co.uk and see how it considers this.

patmmccann: As another datapoint in support of both frequently applying. Some publishers choose for GDPR to always apply, as considered in this document. https://github.com/InteractiveAdvertisingBureau/GDPR-Transparency-and-Consent-Framework/blob/master/TCFv2/IAB%20Tech%20Lab%20-%20CMP%20API%20v2.md

PR #8239, opened 2022-03-30
Floor price : allow having a 0$ floor

JulieLorin:
## Type of change
- [X] Bugfix
- [ ] Feature
- [ ] New bidder adapter   
- [ ] Build related changes
- [ ] CI related changes
- [ ] Other

## Description of change
While using priceFloors, I had an issue where I needed a default floor, but needed to disable it on one condition. 
By adding a floor at 0, it will be ignored and replaced by the defaultFloor, which doesn't seem to me like the expected behavior.

I made a change to check that the floor is a Number, therefore, a 0 floor will be accepted and would not be overwritten by the defaultFloor

### DIFF
--- a/modules/priceFloors.js
+++ b/modules/priceFloors.js
@@ -138,7 +138,7 @@ export function getFirstMatchingFloor(floorData, bidObject, responseObject = {})
 
   let matchingData = {
     floorMin: floorData.floorMin || 0,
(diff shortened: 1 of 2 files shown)


robertrmartinez (approved):
Nice find. Makes sense to want this use case!

Thank you

PR #8242, opened 2022-03-31
Permutive RTD module: support IAB Audience taxonomy

desbo:
## Type of change
- [x] Feature

## Description of change
Updates the Permutive RTD module to facilitate for segmentation by
the new IAB Audience taxonomy.

To achieve this, this change introduces the concept of "transformations"
on the ORT2B `user.data` object. There are two components to these
transformations: a new `transformations` property on the Prebid config,
to be set by the publisher, and logic in the module for the actual
behaviour of the transformation.

We plan to use the transformation logic in this PR, combined with
configuration we'll share with publishers, to send IAB Audience
taxonomy cohort IDs to bidders.

### DIFF
--- a/integrationExamples/gpt/permutiveRtdProvider_example.html
+++ b/integrationExamples/gpt/permutiveRtdProvider_example.html
@@ -201,6 +201,7 @@
                       data: [
                         {
                           name: 'permutive.com',
+                          ext: { segtax: 6 },
                           segment: [{ id: '1' }]
                         }
                       ]
--- a/integrationExamples/gpt/permutiveRtdProvider_example.html
+++ b/integrationExamples/gpt/permutiveRtdProvider_example.html
@@ -153,8 +153,9 @@
                         maxSegs: 500,
                         transformations: [
                           {
-                            id: 'iabAudienceTaxonomy11',
+                            id: 'iab',
                             config: {
+                              segtax: 4,
                               iabIds: {
                                 1000001: '777777',
                                 1000002: '888888'
--- a/integrationExamples/gpt/permutiveRtdProvider_example.html
+++ b/integrationExamples/gpt/permutiveRtdProvider_example.html
@@ -46,6 +46,12 @@
                 },
                 bids: [
                   {
+                    bidder: 'ix',
+                    params: {
+                      siteId: '123456',
+                    }
+                  },
+                  {
                     bidder: 'appnexus',
                     params: {
                       placementId: 13144370,
@@ -135,6 +141,7 @@
             pbjs.que.push(function() {
               pbjs.setConfig({
                 debug: true,
+                pageUrl: 'http://www.test.com/test.html',
                 realTimeData: {
                   auctionDelay: 80, // maximum time for RTD modules to respond
                   dataProviders: [
@@ -142,8 +149,19 @@
                       name: 'permutive',
                       waitForIt: true,
                       params: {
-                        acBidders: ['appnexus', 'rubicon', 'ozone', 'trustx'],
+                        acBidders: ['appnexus', 'rubicon', 'ozone', 'trustx', 'ix'],
                         maxSegs: 500,
+                        transformations: [
+                          {
+                            id: 'iabAudienceTaxonomy11',
+                            config: {
+                              iabIds: {
+                                1000001: '777777',
+                                1000002: '888888'
+                              }
+                            }
+                          }
+                        ],
                         overwrites: {
                           rubicon: function (bid, data, acEnabled, utils, defaultFn) {
                             if (defaultFn){
@@ -160,7 +178,7 @@
                 }
               });
               pbjs.setBidderConfig({
-                bidders: ['appnexus', 'rubicon'],
+                bidders: ['appnexus', 'rubicon', 'ix'],
                 config: {
                   ortb2: {
                     site: {
--- a/integrationExamples/gpt/permutiveRtdProvider_example.html
+++ b/integrationExamples/gpt/permutiveRtdProvider_example.html
@@ -199,11 +199,6 @@
                       keywords: 'a,b',
                       data: [
                         {
-                          name: 'www.dataprovider1.com',
-                          ext: { taxonomyname: 'iab_audience_taxonomy' },
-                          segment: [{ id: '687' }, { id: '123' }]
-                        },
-                        {
                           name: 'permutive.com',
                           segment: [{ id: '1' }]
                         }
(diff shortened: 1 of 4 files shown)


patmmccann:
This feels like a super clunky experience for publishers. Why doesn't your endpoint return the transformation mappings instead of asking publishers to define potentially enormous config?

integrationExamples/gpt/permutiveRtdProvider_example.html
```diff
@@ -160,7 +178,7 @@
                 }
               });
               pbjs.setBidderConfig({
-                bidders: ['appnexus', 'rubicon'],
+                bidders: ['appnexus', 'rubicon', 'ix'],
```
patmmccann, line 181: Please fix line 202 below, this syntax is out of date

modules/permutiveRtdProvider.js
```diff
@@ -260,6 +272,34 @@ function readSegments (key) {
...
+/**
+ * Functions to apply to ORT2B2 `user.data` objects.
+ * Each function should return an a new object containing a `name`, (optional) `ext` and `segment`
+ * properties. The result of the each transformation defined here will be appended to the array
+ * under `user.data` in the bid request.
+ */
+const ortb2UserDataTransformations = {
+  iabAudienceTaxonomy11: (userData, config) => ({
+    name: userData.name,
+    ext: { segtax: '4' },
```
patmmccann, line 286: If you're making pubs define all the transformations, no reason you can't let this be a param...

test/spec/modules/permutiveRtdProvider_spec.js
```diff
@@ -51,6 +51,45 @@ describe('permutiveRtdProvider', function () {
...
+
+      acBidders.forEach(bidder => {
+        expect(bidderConfig[bidder].ortb2.user.data).to.deep.include.members([
+          {
+            name: 'permutive.com',
+            segment: expectedTargetingData
+          },
+          {
+            name: 'permutive.com',
+            ext: { segtax: '4' },
```
patmmccann, line 87: It's an integer

test/spec/modules/permutiveRtdProvider_spec.js
```diff
@@ -51,6 +51,45 @@ describe('permutiveRtdProvider', function () {
...
+      setBidderRtb({}, moduleConfig)
+
+      acBidders.forEach(bidder => {
+        expect(bidderConfig[bidder].ortb2.user.data).to.deep.include.members([
+          {
+            name: 'permutive.com',
+            segment: expectedTargetingData
+          },
+          {
+            name: 'permutive.com',
```
patmmccann, line 86: Is it really permutive that decided the audience member belongs in that iab segment if permutive fails to provide the transformations? The publisher could easily transform any permutive segment to any iab segment; so I'm not sure this remains correct.

patmmccann (changes requested):

modules/permutiveRtdProvider.js
```diff
@@ -84,23 +85,34 @@ export function setBidderRtb (auctionDetails, customModuleConfig) {
...
+ * @param {Object[]} transformationConfigs - array of objects with `id` and `config` properties, used to determine
+ *                                           the transformations on user data to include the ORTB2 object
+ * @param {string[]} segmentIDs - Permutive segment IDs
  * @return {Object} Merged ortb2 object
  */
-function mergeOrtbConfig (currConfig, segmentData) {
-  const segment = segmentData.ac.map(seg => {
-    return { id: seg }
-  })
+function mergeOrtbConfig (currConfig, segmentIDs, transformationConfigs) {
```
patmmccann, line 96: There's a utils finding for this now, i think it's called mergeDeep

desbo:
@patmmccann thanks for the review so far. In terms of the overhead for publishers: I should have mentioned that this change is to support a test with a single publisher that will only involve a small selection of cohorts. For this test, Permutive will provide the publisher with the segment mapping for use in their config, and we already have agreement from the publisher on this way of working for the test.

Our plan was to go with this approach for the test and then revisit our implementation so that the segment mappings come from our SDK rather than the Prebid module in the long term. Ultimately I'd expect that publishers would only need to include a simple flag in their config. Would you accept these changes in light of this extra context?

patmmccann:
If you make segtax default to 4 but be overridable in config; i'd be happy to merge. Publishers want code that is "future-proof"; they don't want to have to suddenly upgrade prebid to support segtax i, j or k in their permutive adapter. We much prefer when adapters keep this in mind.

modules/permutiveRtdProvider.js
desbo: Thanks for the suggestion – after looking at this function I realised it's not really doing a simple merge of two objects but actually updating the `user.data` array with some new entries. So I don't think `mergeDeep` is suitable here, but I've renamed the function to clarify what it does.

desbo:
@patmmccann thanks! I've addressed your comments now, most importantly to include `segtax` as a config parameter. 
Eventually we'll be back with another PR that removes the need for publishers to include the segment ID mappings in their config too, but hopefully this is acceptable in its current form for the test I mentioned.

patmmccann (approved):

integrationExamples/gpt/permutiveRtdProvider_example.html
```diff
@@ -180,11 +199,6 @@
                       gender: 'm',
                       keywords: 'a,b',
                       data: [
-                        {
-                          name: 'www.dataprovider1.com',
-                          ext: { taxonomyname: 'iab_audience_taxonomy' },
-                          segment: [{ id: '687' }, { id: '123' }]
-                        },
```
patmmccann, line 187: two things left: circleci needs to run; add a segtax here. Can be the permutive contextual segtax or the iab contextual (don't use the audience bc that is user.data)

patmmccann (changes requested):

patmmccann:
> @patmmccann thanks! I've addressed your comments now, most importantly to include `segtax` as a config parameter. Eventually we'll be back with another PR that removes the need for publishers to include the segment ID mappings in their config too, but hopefully this is acceptable in its current form for the test I mentioned.
> 
> edit to mention that I'll be away for the next couple of weeks and my colleagues @bartholomews and @Mullefa will commit to this branch if there are any more changes to make.

Thanks so much! Did you notice circleci didnt run? Usually helps to unfollow your fork in circleci before you make the change. Just one tiny change in the example file requested.

desbo:
@patmmccann thanks again. I've added `segtax` to the example, hopefully in the right place but if not, please let me know and one of my teammates will fix. CircleCI is now running after updating this branch with master, although a lot of the tests are timing out – think these are unrelated to my changes?

bartholomews:
Hi @patmmccann circleci is now green, please let me know if there's anything else left to address! Thanks

patmmccann (approved):

Issue #7925, opened 2022-01-08
Ability to set blocked banner attributes / battr


pboisso:
## Type of issue
Feature request

## Description
I could not find any way to set battr to block some banner attributes from bids (only on videos).  How is it possible on banners?  If currently not possible, could this be considered as a feature request?  Thanks!

## Platform details
prebid.js

bretg:
Blocked Attributes/Advertisers/etc are generally defined at the SSP level. 

There are a few bid adapters that have instructions on how to specify the battr/etc. I see a few like Adagio, PulsePoint, TappX.

The definition of battr/badv/etc would be easy to make a convention:
```
pbjs.setConfig({
   ortb2: {
       battr: [...],
       badv: [...],
       bcat: [...],
       etc.
   }
});
```
Then it's up to each bid adapter to read from this location rather than their bidder-specific params.

Will leave it up to @gglas and @patmmccann to decide whether this should (A) be a convention and (B) get enforced in a future major release.

patmmccann:
It seems to me the PR here would look like this: modification to all adapters that see battr, bcat, or badv on their bidder parameters to append the ortb2 field to their params if it is set. 

The doc pull would look like this: all the bidders specifying these params would say they also read from ortb2. The first party data page would get an example of it in the massive set config. A discussion of it would go into the getting started or adops setup section.

patmmccann:
@bretg is there any simple way to specify if the bcat list should replace or append an existing list, eg one the pub configured in an ssp ui? 

cc: @mike-chowla

patmmccann:
also, it seems battr is not a top level param; it belong on imp.banner i think ? https://iabtechlab.com/wp-content/uploads/2021/12/OpenRTB-2.6.pdf

patmmccann:
Closing this, we will document it in the 7.0 docs release

PR #8246, opened 2022-04-01
PubMatic: add "acat" parameter 

kapil-tuptewar:
## Type of change

- [x] Feature

## Description of change

If Publisher add acat field on page then same will get passed to PubMaticBidAdapter request under request.ext.acat 
Documentation https://docs.prebid.org/dev-docs/pbs-bidders.html#pubmatic (https://github.com/prebid/prebid.github.io/pull/3709/files)


patmmccann:
What does acat stand for?

patmmccann:
It appears to stand for allowed categories; which seems bizarre for two reasons, one is the list would get very long, two is you could just generate it server side by taking all the categories and removing the bcats like everyone else. What's the utility here?

kapil-tuptewar:
Yes thats correct, it stands for allowed categories. There are limited number of categories that could be passed (max 25). And as publisher decides on fly which all categories needs to be allowed so we need to pass them from client side only.

kapil-tuptewar:
Updated documentation link in description.

patmmccann:
Is this for us? Are we the requesting publisher you guys are building this for? If so, why don't we work together to build a better bcat convention in the ortb2 object?

pm-nilesh-chate:
> Is this for us? Are we the requesting publisher you guys are building this for? If so, why don't we work together to build a better bcat convention in the ortb2 object? This is the sort of thing that shouldn't be specific to an adapter.

Hi,

Our publishers set `bcat` list in our DB which PubMatic uses while severing the requests but a few publishers needed a way to unblock a few entries from the `bcat` list at realtime for a few auctions and they do not want to update DB entries. So have introduced a bidder level param called `acat` whose entries will be removed from `bcat` during action.
Basically, for PubMatic AdServer `effective bcat = bcat list from db - acat list from request`.

kapil-tuptewar:
> > Is this for us? Are we the requesting publisher you guys are building this for? If so, why don't we work together to build a better bcat convention in the ortb2 object? This is the sort of thing that shouldn't be specific to an adapter.
> 
> Hi,
> 
> Our publishers set `bcat` list in our DB which PubMatic uses while severing the requests but a few publishers needed a way to unblock a few entries from the `bcat` list at realtime for a few auctions and they do not want to update DB entries. So have introduced a bidder level param called `acat` whose entries will be removed from `bcat` during action. Basically, for PubMatic AdServer `effective bcat = bcat list from db - acat list from request`.


@patmmccann Could you please let your thought on this.

patmmccann:
This is fine; but there is a standard developing. If a publisher sets the bcats via the methods described in https://github.com/prebid/Prebid.js/issues/7925 can you have the ortb2 object considered also?

modules/pubmaticBidAdapter.js
```diff
@@ -1106,6 +1126,9 @@ export const spec = {
       if (bid.params.hasOwnProperty('bcat') && isArray(bid.params.bcat)) {
         blockedIabCategories = blockedIabCategories.concat(bid.params.bcat);
       }
+      if (bid.params.hasOwnProperty('acat') && isArray(bid.params.acat)) {
```
patmmccann, line 1129: https://github.com/prebid/Prebid.js/issues/7925 please consider the ortb2 config object as well and document which you will consider first and which you will fallback to if both are set

patmmccann (changes requested):

kapil-tuptewar:
@patmmccann We will be considering bcat support using ortb2 config in separate ticket in upcoming sprints. And now for acat we will considering ortb2 config as well. Changes are pushed, please review.

modules/pubmaticBidAdapter.js
kapil-tuptewar: Changes done, considering ortb2 config as well.

patmmccann:
This logic is difficult to follow and an enormous departure from standards. Who is requesting this? Is it Cafemedia? 
`      if (bid.params.hasOwnProperty('bcat') && isArray(bid.params.bcat)) {
        blockedIabCategories = blockedIabCategories.concat(bid.params.bcat);
      }
      if (bid.params.hasOwnProperty('acat') && isArray(bid.params.acat)) {
        allowedIabCategories = allowedIabCategories.concat(bid.params.acat);
      }`

Re this section: if bcat is specified, you would want to completely ignore acat right? Acat is only to override bcats specified in your ui, not on the request. Also, please include bcats from ortb2. 

I see you mentioned you want to add this at a later date, I am hesitant to approve code that makes things muddy and off standard when you could jusst include them in one shot, particularly if you are doing this work for CafeMedia's benefit. We would rather specify bcat once in ortb2 object.

mike-chowla:
I chatted with @patmmccann about this on Slack.

PubMatic's handling of bcat is to add the categories sent in bcat to the account configured blocklist.  Changing this behavior is problematic for us since our publishers depend on it and at least some of them are not sending their full block list when sending bcat.

patmmccann:
Hi, as a final resulotion here, can you just union ortb2. Bcat with bidder params bcat and  we'll merge?

Bcat is frustratingly nuanced and i learned a bunch looking into this. Thanks for your patience.

kapil-tuptewar:
> Hi, as a final resulotion here, can you just union ortb2. Bcat with bidder params bcat and we'll merge?
> 
> Bcat is frustratingly nuanced and i learned a bunch looking into this. Thanks for your patience.

@patmmccann 
So basically e.g. if we have slot with bid.params.bcat property set with values ['IAB-1', 'IAB-2'] and publisher sets bcat using ortb2 config as {bcat: ['IAB-2', 'IAB-3', 'IAB-4']} then before sending request we should combine bcat values from bid.params.bcat and orbt2.bcat. So request has property bcat with values as ['IAB-1', 'IAB-2', 'IAB-3', 'IAB-4'].
Is my understanding correct ?

@mike-chowla I hope above implementation dose not create problems for our publishers.

kapil-tuptewar:
> > Hi, as a final resulotion here, can you just union ortb2. Bcat with bidder params bcat and we'll merge?
> > Bcat is frustratingly nuanced and i learned a bunch looking into this. Thanks for your patience.
> 
> @patmmccann So basically e.g. if we have slot with bid.params.bcat property set with values ['IAB-1', 'IAB-2'] and publisher sets bcat using ortb2 config as {bcat: ['IAB-2', 'IAB-3', 'IAB-4']} then before sending request we should combine bcat values from bid.params.bcat and orbt2.bcat. So request has property bcat with values as ['IAB-1', 'IAB-2', 'IAB-3', 'IAB-4']. Is my understanding correct ?
> 
> @mike-chowla I hope above implementation dose not create problems for our publishers.

@patmmccann suggested changes are done we doing union for ortb2.bcat and bid.params.bcat.

patmmccann (approved):

patmmccann:
Excellent, thank you!

PR #8254, opened 2022-04-04
Revert "Glimpse: update api and request shape, optimize and refactor"

msm0504:
Reverts prebid/Prebid.js#8237


bretg (approved):

bretg:
@eddieyu1998 - we're reverting your PR because it's possible that a given request is in scope for BOTH GDPR and CCPA. We're giving you a chance to update your endpoint to support both concurrently.

If your endpoint cannot support both privacy protocols concurrently, you can choose to move forward again with a new (duplicate) PR but cannot declare support for CCPA in your adapter documentation.

eddieyu1998:
@bretg thanks for the review, we will be submitting a new pr shortly to support the case where both GDPR and CCPA applies

PR #8255, opened 2022-04-05
adding PAF userId and RTD modules. test paf bid adapter

bwschmidt:
## Description of change

WIP paf modules


bwschmidt:
can be ignored for now

PR #8261, opened 2022-04-05
AudienceRun Bid Adapter: Add Support for User Matching, Referer, …

audiencerun:
## Type of change

- [ ] Bugfix
- [X] Feature
- [ ] New bidder adapter   
- [ ] Build related changes
- [ ] CI related changes
- [ ] Other

## Description of change

Add Support for User Matching, Referer, Schain, US Privacy, GLVID

## Other information

PR https://github.com/prebid/prebid.github.io/pull/3687

### DIFF
--- a/modules/audiencerunBidAdapter.js
+++ b/modules/audiencerunBidAdapter.js
@@ -1,32 +1,30 @@
-import { deepAccess, isFn, logError, getValue, getBidIdParameter, _each, isArray, triggerPixel } from '../src/utils.js';
+import {
+  deepAccess,
+  isFn,
+  logError,
+  getValue,
+  getBidIdParameter,
+  _each,
+  isArray,
+  triggerPixel,
+  formatQS,
+} from '../src/utils.js';
 import { config } from '../src/config.js';
 import { registerBidder } from '../src/adapters/bidderFactory.js';
 import { BANNER } from '../src/mediaTypes.js';
+import { createEidsArray } from './userId/eids.js';
 
 const BIDDER_CODE = 'audiencerun';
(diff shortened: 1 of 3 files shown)


ChrisHuie:
@audiencerun can you please remove changes to the package-lock file 🙏

audiencerun:
Hi @ChrisHuie, just restored the package-lock.json file. Thanks.

modules/audiencerunBidAdapter.js
```diff
@@ -44,19 +42,53 @@ function getBidFloor(bid) {
...
+    return 0;
   }
 }
 
+/**
+ * Returns the most top page referer.
+ *
+ * @returns {string}
+ */
+function getPageReferer() {
```
robertrmartinez, line 54: Is there a reason this is being duplicated?

`bidderRequest.refererInfo.referer` does this logic already, so not really sure this is needed.

https://github.com/prebid/Prebid.js/blob/master/src/refererDetection.js

robertrmartinez (changes requested):

modules/audiencerunBidAdapter.js
audiencerun: Hi @robertrmartinez, it is not a duplicate as `bidderRequest.refererInfo.referer` returns only the URL of the page where Prebid.js is called while we need the URL that linked to this page and not the URL of the current page.

Here is a screen that shows it and is available at https://7buoei.csb.app/?pbjs_debug=true

image

Feel free if you need more details.

robertrmartinez (approved):
We will be making a util function or add this to refererDetection in order to make sure adapters do not duplicate code. Thanks!

PR #8272, opened 2022-04-08
Colossus Bid Adapter: update buildRequests

HuddledMasses:


robertrmartinez (changes requested):
This looks like adding support if it is video.

Please also add in a unit test or two to cover it!

Thanks.

robertrmartinez (approved):

PR #1824, opened 2017-11-13
Emit array of objects from BID_TIMEOUT event

matthewlane:
## Type of change
- Other

## Description of change
Changes the data payload emitted from the `BID_TIMEOUT` event from an array of bidder codes strings, to an array of objects containing the bid id, bidder code, ad unit code, and auction id (this is equivalent to `requestId` in pre-1.0) for each timed out bid.

Before
```JavaScript
["bidderOne", "bidderTwo"]
```

After
```JavaScript
[
  {
    "bidId": "2baa51527bd015",
    "bidder": "bidderOne",
    "adUnitCode": "/19968336/header-bid-tag-0",
    "auctionId": "66529d4c-8998-47c2-ab3e-5b953490b98f"
  },
  {
    "bidId": "6fe3b4c2c23092",
    "bidder": "bidderTwo",
    "adUnitCode": "/19968336/header-bid-tag-0",
    "auctionId": "66529d4c-8998-47c2-ab3e-5b953490b98f"
  }
]
```

## Other information
Closes #844

PR #8288, opened 2022-04-15
Prebid Core: send all bid info within the timeout event

JulieLorin:
## Type of change
- [X] Feature

## Description of change
We are using the bids and no bids events (EVENTS.BID_TIMEOUT and EVENTS.NO_BID) and we do not retrieve the same information about the bid on both. The timeout event only has a few information. This is because most fields are not sent. 

I don't know if there is a reason to keep only little information about the timeout event, but if it's ok, it would be convenient for us to have more information.

This pull request updates the timeout event and will send the original bid in the timeout event. There will be no removal or renaming of fields. This is only an addition of field.

### DIFF
--- a/src/auction.js
+++ b/src/auction.js
@@ -848,13 +848,7 @@ function groupByPlacement(bidsByPlacement, bid) {
 function getTimedOutBids(bidderRequests, timelyBidders) {
   const timedOutBids = bidderRequests
     .map(bid => (bid.bids || []).filter(bid => !timelyBidders.has(bid.bidder)))
-    .reduce(flatten, [])
-    .map(bid => ({
-      bidId: bid.bidId,
-      bidder: bid.bidder,
-      adUnitCode: bid.adUnitCode,
-      auctionId: bid.auctionId,
-    }));
+    .reduce(flatten, []);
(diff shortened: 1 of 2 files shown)


ChrisHuie (approved):

ChrisHuie:
@patmmccann double-checking this behavior with you to make sure I'm not missing any side-effects from this changee?

dgirardi:
This is the original PR:

https://github.com/prebid/Prebid.js/pull/1824

JulieLorin:
@patmmccann The field I want is a custom one added on the bid (not something officially supported by Prebid) so I cannot really make a list. That's why I was proposing to keep all bid's fields just like it is done in the nobid event (it seems to be more consistent)

The original PR was about adding more information (before, there was only the bidderCode) but I don't see a reason to limit it to only 4 fields.

I will add a test for this

patmmccann:
i like this change, i just worry that it might be breaking to some people expecting the existing signature, we're releasing prebid 7 shortly, mind if I change the base for safety? I can add it to the #7796 release notes

patmmccann (approved):

PR #8299, opened 2022-04-20
Fix dwid convert

from20020516:
https://vg-ceg.backlog.com/view/FLUCT_SSP-15490


from20020516:
close due to mistake.

Issue #8303, opened 2022-04-20
Static GDPR-Consent 


arneschulz1984:
## Type of issue
Bug / Question

## Description
Until version 6.18.0 it was possible to set a static GDPR Consent like documentated here (https://docs.prebid.org/dev-docs/modules/consentManagement.html -> Example 2: Static CMP using custom data passing.) and the Prebid.JS requests includes the provided consentString. With Version >= 6.19.0 this behavior changed.

Until 6.18.0 setting the static GDPR Consent like this:

let tcStr = 'CO4EXq8O4EYOWADABCDEAxCsAP_AAH_AAAAAGStV_T9fb2vj-_599_t0eY1f9_63t-wzhheMs-8NyZeX_J4Wv2MyvB34JqQKGRgkunLBAQdtHGncTQgBwIlViTLMYk2MjzNKJrJEmlsbO2dYGH9Pn8XT_ZKZ70-vv__7v3___33_4GSEEmCpfAQJCWMBJNmlUKIEIVxIVAOASghGEg0sNCRQU7I4CPUACABAYAIQIAQAgohJBAAIAAElEQAgAwIBEARAIAAQAjQEIACJAEFgBIGAQACoGhYARRBKBIQYHBUcogQFSLRQTwCQgAAMAAEAABQAAYAAHAACAAAkAAKAAA.YAAAAAAAAAA'

var pbjs = pbjs || {};
pbjs.que = pbjs.que || [];
pbjs.que.push(function() {
    pbjs.setConfig({
        priceGranularity: "high", 
        consentManagement: {
            gdpr: {
                cmpApi: 'static',
                consentData: {
                    getTCData: {
                        tcString: tcStr,
                        gdprApplies: true,
                        purpose: {
                            consents: {
                                1: true,
                                2: true,
                                3: true,
                                4: true,
                                5: true,
                                6: true,
                                7: true,
                                8: true,
                                9: true,
                                10: true
                            }
                        }
                    }
                }
            }
        }
    });
});

has the impact that a Prebid.JS bid requests includes:

`{
  ......
  "gdprConsent": {
[...]

PR #8304, opened 2022-04-20
Consent management: fix bug where 'static' consent configuration does not work correctly

dgirardi:
## Type of change

- [x] Bugfix

## Description of change

https://github.com/prebid/Prebid.js/pull/8185/ introduced a bug in processic "static" consent configuration (https://github.com/prebid/Prebid.js/issues/8303) - this addresses it.

## Other information

Fix https://github.com/prebid/Prebid.js/issues/8303


harpere (approved):

wqi1972:
We had exactly the same issue for our static consentManagement Usp setup. I just apply the fix from this PR, and confirm it is working fine now.

robertrmartinez (approved):

Issue #8307, opened 2022-04-20
No floor price set in prebidServerBidAdapter


rmattis:
## Type of issue
Bug

## Description
We encountered a problem with the `prebidServerBidAdapter` where `imp.bidfloor` and `imp.bidfloorcur` in the request is not set.
We debugged this and found that the bug was added with version 6.17.0. 
If the first bidder in a request is not configured for all adUnits, the `bidfloor` and `bidfloorcur` is not set for all adUnits where this bidder is not configured.

https://github.com/prebid/Prebid.js/blob/d4c057a914ef79a1508db1daa68d27b1a3002610/modules/prebidServerBidAdapter/index.js#L710
This line is probably what causes the bug and was added with #8154 

The main problem is that only the first prebid request is used to receive the floor price for all bidders:
https://github.com/prebid/Prebid.js/blob/d4c057a914ef79a1508db1daa68d27b1a3002610/modules/prebidServerBidAdapter/index.js#L537

https://github.com/prebid/Prebid.js/blob/d4c057a914ef79a1508db1daa68d27b1a3002610/modules/prebidServerBidAdapter/index.js#L758-L773

## Steps to reproduce
1. Add 2 adUnits
2. Configure the first bidder only for the first ad unit
3. Configure the first bidder as server side

```
const adUnit1 = {
  mediaType: { banner: { sizes: [[300, 250]] } },
  ortb2Imp: { ext: { prebid: { storedrequest: { id: 'ad-slot-1' } } } },
  floors: {....},
  bids: [ { bidder: 'Bidder A', params { }, bidder: 'Bidder B', params { } }  ]
}

const adUnit2 = {
  mediaType: { banner: { sizes: [[300, 250]] } },
  ortb2Imp: { ext: { prebid: { storedrequest: { id: 'ad-slot-2' } } } },
  floors: {....},
  bids: [ { bidder: 'Bidder B', params { } }  ]
}

const s2sConfig = {
  // ...
  bidders: [ 'Bidder A']
}
```

## Test page
None

### Expected results
ad-slot-2 prebid server request should contain `bidfloor` and `bidfloorcur`

### Actual results
ad-slot-2 prebid server request doesn't contain `bidfloor` and `bidfloorcur`

## Platform details
prebid.js version > 6.17.0

## Other information
Probably changes from https://github.com/prebid/Prebid.js/pull/8154 cause this bug

dgirardi:
[...]

PR #8309, opened 2022-04-20
PBS adapter: fix bug with priceFloors sometimes not being set in request

dgirardi:
## Type of change

- [x] Bugfix

## Description of change

﻿This addresses https://github.com/prebid/Prebid.js/issues/8307 by picking what's likely to be the minimum floor for an adUnit (instead of just looking at the floor defined by the first request it finds, which may not be there)

### DIFF
--- a/modules/prebidServerBidAdapter/index.js
+++ b/modules/prebidServerBidAdapter/index.js
@@ -757,7 +757,7 @@ Object.assign(ORTB2.prototype, {
       }
 
       const floor = (() => {
-        // we have to pick a floor for the imp - here we attempt to find the minimimum floor
+        // we have to pick a floor for the imp - here we attempt to find the minimum floor
         // across all bids for this adUnit
 
         const convertCurrency = typeof getGlobal().convertCurrency !== 'function'
--- a/modules/prebidServerBidAdapter/index.js
+++ b/modules/prebidServerBidAdapter/index.js
@@ -764,20 +764,19 @@ Object.assign(ORTB2.prototype, {
           ? (amount) => amount
           : (amount, from, to) => {
             if (from === to) return amount;
-            let result;
+            let result = null;
             try {
               result = getGlobal().convertCurrency(amount, from, to);
             } catch (e) {
-              result = amount;
             }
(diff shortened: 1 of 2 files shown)


modules/prebidServerBidAdapter/index.js
```diff
@@ -755,21 +756,64 @@ Object.assign(ORTB2.prototype, {
         deepSetValue(imp, 'ext.prebid.storedauctionresponse.id', storedAuctionResponseBid.storedAuctionResponse.toString());
       }
 
-      const getFloorBid = find(firstBidRequest.bids, bid => bid.adUnitCode === adUnit.code && typeof bid.getFloor === 'function');
+      const floor = (() => {
+        // we have to pick a floor for the imp - here we attempt to find the minimimum floor
```
patmmccann, line 760: spelling on minimum

patmmccann:
@robertrmartinez could you take this one?

robertrmartinez:
Ahhh yes, makes sense this bug existed!

robertrmartinez (approved):
Looks good, Why the decision to take the min?

I don't really know to be honest what matters just curious.

Could have also just grabbed the first valid floor found I guess.

Or the max?

Does not really matter to me, most of the time it will prob always be the same!

dgirardi:
@robertrmartinez my reasoning is that if you don't pick the minimum one you might lose bids that would have otherwise been available - say that you somehow manage to set a floor of `1` for `bidderA` and `2` for `bidderB`; if the request to PBS sets `bidfloor: 2` and neither bidder can beat it, you won't get any bid. But `bidderA` might be able to beat `1`, which the publisher has indicated as acceptable.

But I'm also not sure this makes sense - maybe the revenue from when they can beat the higher floor outweigh the lost bids.

If noone thinks this is necessary, I'd rather remove the added complexity and just pick the first one.

patmmccann:
i find the reason to use the min here persuasive, seems safer

robertrmartinez:
Sounds good to me!

patmmccann (approved):

PR #8317, opened 2022-04-25
Prebid Server Adapter: add native asset id parameter

jsnellbaker:
## Type of change

- [x] Feature

## Description of change
This PR adds in the missing `id` field(s) for the `native.request.assets` object array of a PBS request.  The ORTB native spec has this field denoted as required (see PDF link below).  

While leaving out the field did not appear to be an issue for some of the PBS endpoints, it was found to be an issue for the appnexus PSP endpoint.  

Reference -  OpenRTB Native doc:
https://www.iab.com/wp-content/uploads/2018/03/OpenRTB-Native-Ads-Specification-Final-1.2.pdf
-- section 4.2 (pg 13)

### DIFF
--- a/modules/prebidServerBidAdapter/index.js
+++ b/modules/prebidServerBidAdapter/index.js
@@ -563,13 +563,15 @@ Object.assign(ORTB2.prototype, {
       const nativeParams = adUnit.nativeParams;
       let nativeAssets;
       if (nativeParams) {
+        let idCounter = 0;
         try {
           nativeAssets = nativeAssetCache[impressionId] = Object.keys(nativeParams).reduce((assets, type) => {
             let params = nativeParams[type];
 
             function newAsset(obj) {
               return Object.assign({
-                required: params.required ? 1 : 0
+                required: params.required ? 1 : 0,
+                id: idCounter++
               }, obj ? cleanObj(obj) : {});
             }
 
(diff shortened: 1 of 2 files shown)


modules/prebidServerBidAdapter/index.js
```diff
@@ -563,13 +563,15 @@ Object.assign(ORTB2.prototype, {
...
+        let idCounter = 0;
         try {
           nativeAssets = nativeAssetCache[impressionId] = Object.keys(nativeParams).reduce((assets, type) => {
             let params = nativeParams[type];
 
             function newAsset(obj) {
               return Object.assign({
-                required: params.required ? 1 : 0
+                required: params.required ? 1 : 0,
+                id: (isNumber(params.id)) ? params.id : idCounter++
```
bretg, line 574: @jsnellbaker - this doesn't cover the case where some assets have IDs and others don't. So there could wind up being duplicate IDs. I guess it might be ok to just document this behavior clearly.

jsnellbaker: I agree it should be clearly documented, but I thought of a potential alternative that maybe could better handle the `id`s.  I'll push the commit shortly, let me know if it's worthwhile.

bretg (approved):
This assumes that the user has specified increasing and non-duplicate ids, but that was the case before and this approach is a nice improvement over the original. Thanks.

patmmccann (approved):

PR #8320, opened 2022-04-26
TheMediaGrid Bid Adapter: support bid.ortb2.site.content

TheMediaGrid:
## Type of change

- [x] Feature

## Description of change

Added support of ortb2.content.site from bid, to support new format of jwpRtdAdapter.

- grid-tech@themediagrid.com


modules/gridBidAdapter.js
```diff
@@ -167,6 +176,11 @@ export const spec = {
       request.site.content = content;
     }
 
+    const ortb2SiteContent = deepAccess(bidOrtb2, 'site.content');
+    if (ortb2SiteContent) {
+      request.site.content = mergeDeep(request.site.content || {}, ortb2SiteContent);
```
patmmccann, line 181: If you get all of it, can you simplify other places in your adapter, like where you get request.site.content.genre specifically?

TheMediaGrid: Hello! But here we get site.content from bid.ortb2, and site.content.genre we get from config

patmmccann: but those are the same?

modules/gridBidAdapter.js
```diff
@@ -89,6 +90,14 @@ export const spec = {
           content = jwTargeting.content;
         }
       }
+      if (ortb2) {
```
dgirardi, line 93: I believe ortb2 is always null here - am I missing something? AFAIK the `bid` object does not have that property.

dgirardi: OK I understand now - one particular RTD provider adds that to the bids (https://github.com/prebid/Prebid.js/pull/7886). That does not make much sense to me, why would it demand an update to every bid adapter instead of calling `setConfig`?

patmmccann: fwiw #7886 caused very large revenue loss to cafemedia and we upgraded to prebid 6.7 with it carved out. we are still investigating why

TheMediaGrid: We're actually going to be super happy if the JWP RTD module can move from bid object to the regular ortb2 approach so that our standard integration would work for this

karimMourra: Hi! We will correct this ASAP, thank you for raising!

modules/gridBidAdapter.js
TheMediaGrid: Sorry, can you provide a bit more details about the source of your concern here, not quite following

modules/gridBidAdapter.js
TheMediaGrid: @patmmccann @karimMourra I guess, as soon as JW team will support upgrading their RTD module to setConfig, this part of the MediaGrid PR won't be needed anymore, right? Let's make sure the two PRs fall into the same release

karimMourra: @TheMediaGrid  please note that https://github.com/prebid/Prebid.js/pull/8338#issuecomment-1112999837 is migrating away from reading ortb2 from getConfig

karimMourra: PR for JW RTD: https://github.com/prebid/Prebid.js/pull/8354

TheMediaGrid: @patmmccann agreed, this one is no longer needed with 8354 going live

PR #8327, opened 2022-04-27
Bid Richemedia adapter : use a 1x1 creative

yki-t:
## Type of change
Adapter update

## Description of change
Update adapter bid response size


patmmccann (changes requested):
please write a test

yki-t:
@patmmccann 
> please write a test

Here is the test 
https://github.com/prebid/Prebid.js/blob/master/test/spec/modules/big-richmediaBidAdapter_spec.js

Do you mean I need to write test for the change ?

PR #8330, opened 2022-04-27
Prebid 7: remove appnexusAnalyticsAdapter

jsnellbaker:
## Type of change
- [x] Other

## Description of change
The appnexus Analytics Adapter is not supported (and hasn't been for some time). We're removing the file from the project to make it clear and to avoid confusion about its status.

It seems like there are no corresponding docs about the appnexusAnalyticsAdapter (that I could find at least), so no changes should be needed for the prebid.org site.


patmmccann:
tagging #7796  for linkage

Issue #7651, opened 2021-11-02
Site.content.data for a specific [auction]


patmmccann:
## Type of issue

Right now https://docs.prebid.org/features/firstPartyData.html describes how to set ad unit specific first party data, but not how to set seller defined contextual segments along the lines of #6057 that would normally go in site.content.data

Ad-unit specific first party data might be valid if there is a video unit on the page and the contextual segments are about the video, not the text. It also might be relevant on various single page experiences with multiple types on content that fall into different groupings. 

## Possible solution
One solution is we set it on imp.ext.data.content . The issue here is adapters would all have to start looking for this data in a second place. 

eg

```
imp": [{
		"id": "Video_Individual_Autoplay_SOff",
		"ext": {
			"data": {
				"adserver": {
					"name": "gam",
					"adslot": "/18190176,83852415/AdThrive_Video_Individual_Autoplay_SOff/53237591208f222c05b9a652"
				},
				"jwTargeting": {
					"playerID": "9F5pbYIL",
					"mediaID": "9F5pbYIL"
				},
				"content": [{
					"name": "www.iris.tv",
					"segment": [{
						"id": "ic_4146983"
					}, {
						"id": "ic_1670398"
					}, {
						"id": "ic_4260700"
					}, {
						"id": "ic_9534888"
					}, {
						"id": "ic_9954675"
					}, {
						"id": "ic_3393155"
					}],
					"ext": {
						"segtax": 501
					}
				}],
```

## Alternative solution
It seems publishers might already have a workaround, setting  the global fpd with these extra segments and then quickly unset it so other units aren't affected. Should we develop a utility function to more easily support this concept of temporary global fpd?

gglas:
Discussed today, @bretg will look into and discuss the best way to handle this.  Could be that we feature this in a second place.

bretg:
Two use cases have been offered: 
- video might have different site.content.data
- a page might have a split personality where some adunits have different site.content.data

Would it be reasonable to assume that these scenarios could be run as a separate auction? If so, we could add the ability to set ortb2 on `requestBids`:

```
pbjs.requestBids({
    bidsBackHandler: sendAdserverRequest,
    ...
    ortb2: {
       // for this auction only, deep-merge this block into any existing ortb2 settings
    }
});
```

patmmccann:
Yes, This appears to be an excellent solution!

dgirardi:
[...]

Issue #6240, opened 2021-01-29
Conditional permissioning of id's & data


patmmccann:
## Type of issue
Feature Request

In the publisher committee, it was expressed that several publishers may want to offer conditional permissioning, in which user segment and contextual segment data may be transmitted when id's are not available and vice versa, but not both together. These publishers would also prefer to specify a priority. 

Closely related to #5814 and #3687 and #4472 and the concept of allowedBidders eg at https://github.com/prebid/Prebid.js/blob/f61311769823d0431cd90a05c125d88905c0555e/test/spec/modules/prebidServerBidAdapter_spec.js#L1518 

One possible scheme, extending from #5814

`
    ext: {
        prebid: {
            data: {
                eidpermissions: [   // prebid server will use this to filter user.ext.eids
                   {"source": "sharedid.org", "bidders": ["*"], "suppressIdWithData" : true},
                   {"source": "neustar.biz", "bidders": ["bidderB"], "suppressDataWithId" : true},
                   {"source": "id5-sync.com", "bidders": ["bidderA","bidderC"]}
                ]
            }
        }
    }
}
`

bretg:
@patmmccann - would like to flesh out the requirements a little more. Want to be clear about the scope and what you mention as "priority". The example you provide implies that eids are always secondary - they never show up if FPD is present.

Prebid.js and Prebid Server, both of which control the data sent to bid adapters, may have data that is sensitive to publishers and their users. This data includes: User IDs, First Party Data, and specifically user segments and contextual segments within First Party Data. 

1) Both Prebid.js and Prebid Server must allow publishers to control which sensitive is sent to each bid adapter.
2) User segments and contextual segments must be controlled per existing FPD controls. i.e. setBidderConfig() in Prebid.js and for Prebid Server, ext.prebid.bidderconfig  and ext.prebid.data.bidders[]. Note: controlling segments in this way is an extension of what's currently protected by PBS for AMP and App, since ext.prebid.data.bidders currently only covers {site/app/user}.ext.data
3) It should be possible for a publisher to define on a per-bidder basis: A) send both FPD and IDs, B) send neither FPD nor IDs, C) send IDs only if no FPD is available and finally D) send FPD only if no IDs are available.


Questions:
- Would some IDs be allowed to go to all bidders? Would the existence of those "open" IDs affect FPD?
- Likewise, some FPD may exist in non-protected areas (e.g. AdUnit). Does that count as FPD for the purpose of suppressing IDs?
-

[...]

PR #8338, opened 2022-04-28
Prebid 7: auction-specific first party data and refactoring of FPD internals

dgirardi:
## Type of change

- [x] Feature

## Description of change
This restructures the way Prebid handles first party data to enable auction-specific FPD (https://github.com/prebid/Prebid.js/issues/7651)

﻿The changes are:

 - `pbjs.requestBids` now accepts an `ortb2` parameter, which will get merged into (and overrides) the global `ortb2` config only for that particular auction
 - `requestBids` now "freezes" FPD for a particular auction into global (`getConfig('ortb2')` + the auction-specific `ortb2`) and bidder fragments that get passed down to a new hook, `startAuction` (and further down the stack to auctionManager, adapterManager, etc)
 - FPD modules and RTD modules have been refactored to hook into `startAuction` (instead of `requestBids`), and they read from and modify its FPD parameters instead of calling `getConfig` / `setConfig`
 - bid requests and bidd*er* requests (the parameters passed to bidders' `buildRequests`) are now augmented with an `ortb2` property that replaces the use of `getConfig('ortb2')` for bid adapters. All adapters have been refactored to read from those instead of calling `getConfig`.
 - FPD submodules (currently just enrichment and validation) now run on every auction, and the global `refreshFpd` has been removed - with FPD now being potentially different for each one, I don't think the previous run-once approach makes sense. That said, both submodules can be improved to memoize some of their processing - that's not included in this PR.

The choice to put `ortb2` into every bid request is due to some adapters doing their FPD logic in the validation phase (`isBidRequestValid`) or in `transformBidParameters`, where the bidd*er* request is not available.

Note that two RTD providers - weborama and jwPlayer - were already populating `bid.ortb2` under some circumstances. Most adapters were blind to it (I know only of proxistore looking for it - related is also https://github.com/prebid/Prebid.js/pull/8320). I have refactored those two to update the bidder-level ortb2 config instead.

## Other information

Docs PR TBD


dgirardi:
Tagging:

 - @desbo &  @dreischer for changes to the permutive RTD adapter
 - @karimMourra  for jwPlayer
 - @peczenyj for weborama

peczenyj:
LGTM

I am currently working on this module, I will wait to rebase it properly

karimMourra (approved):

modules/jwplayerRtdProvider.js
```diff
@@ -158,7 +158,7 @@ export function enrichAdUnits(adUnits) {
       const contentId = getContentId(vat.mediaID);
       const contentData = getContentData(vat.segments);
       const targeting = formatTargetingResponse(vat);
-      enrichBids(adUnit.bids, targeting, contentId, contentData);
+      enrichBids(ortb2Fragments?.bidder || {}, adUnit.bids, targeting, contentId, contentData);
```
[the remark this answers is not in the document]
dgirardi: true, this or the default param value is superfluous - I missed it

dgirardi:
@peczenyj this is a breaking change for version 7; you'll probably want to fix for 6 independently, using the current `getBidderConfig` / merge / `setBidderConfig` strategy.

peczenyj:
> @peczenyj this is a breaking change for version 7; you'll probably want to fix for 6 independently, using the current `getBidderConfig` / merge / `setBidderConfig` strategy.

Well, I try to structure this refactor to be easier to rebase  https://github.com/prebid/Prebid.js/pull/8350

desbo (approved):

src/config.js
```diff
@@ -314,42 +314,52 @@ export function newConfig() {
...
+  function _getRestrictedConfig() {
+    // This causes reading 'ortb2' to throw an error; with prebid 7, that will almost
+    // always be the incorrect way to access FPD configuration (https://github.com/prebid/Prebid.js/issues/7651)
+    // code that needs the ortb2 config should explicitly use `getAnyConfig`
+    // TODO: this is meant as a temporary tripwire to catch inadvertent use of `getConfig('ortb')` as we transition.
+    // It should be removed once the risk of that happening is low enough.
+    const conf = _getConfig();
+    Object.defineProperty(conf, 'ortb2', {
+      get: function () {
+        throw new Error('invalid access to \'orbt2\' config - use request parameters instead');
```
patmmccann, line 326: this is applicable to bidders, but what if publishers what to access the config, for example, to achieve #6240

dgirardi: the publisher API does not hit this code path - the global `pbjs.getConfig` and `pbjs.readConfig` are aliased to `getAnyConfig` and `readAnyConfig`.

patmmccann (approved):

mmoschovas (approved):

PR #8341, opened 2022-04-29
add TPMN UserSync Bidder Adapter

tpmn-admin:
## Type of change

- [ ] Bugfix
- [X] Feature
- [ ] New bidder adapter   
- [X] Build related changes
- [ ] CI related changes
- [ ] Other

## Description of change

Add to script side getUserSyncs function.
>use cookie
>use generate to UUID
>userSync Type Support :  iframe or pixel
>add to getUserSyncs() test case
>keep greater than 80% code coverage.

Sample page : https://static.tpmn.co.kr/sandbox/prebid/cookieSync.html

Email : info@tpmn.co.kr


modules/tpmnBidAdapter.js
```diff
@@ -118,3 +199,12 @@ function bidToRequest(bid) {
 
   return bidObj;
 }
+
+function getUuid() {
+  const UUID_KEY = 'uuid';
+  const id = storage.getCookie(UUID_KEY);
+  if (id) return id;
+  let newId = generateUUID().replace(/-/g, '');
+  storage.setCookie(UUID_KEY, newId);
```
patmmccann, line 208: why not default to local storage?

modules/tpmnBidAdapter.js
```diff
@@ -64,7 +66,86 @@ export const spec = {
...
+      const coppa = config.getConfig('coppa') ? 1 : 0;
+      policyParam += `&coppa=${coppa}`;
+      syncArr.push({
+        type: 'iframe',
+        url: IFRAMESYNC + policyParam
+      })
+    } else {
+      const uuid = getUuid();
+      if (uuid) {
+        const bidswitchId = storage.getCookie('bidswitch');
```
patmmccann, line 94: this will check for the presence of this cookie id in the first party, right? Seems it would always be true, even if bidswitch has a cookie.

test/spec/tpmnBidAdapter_spec.js
```diff
@@ -141,4 +159,44 @@ describe('tpmnAdapterTests', function() {
       expect(bidResponses).deep.equal([mockBidResult]);
     });
   });
+
+  describe('getUserSync', function() {
+    const KEY_ID = 'uuid';
+    const TMP_UUID = generateUUID().replace(/-/g, '');
+    it('getCookie mock Test', () => {
```
patmmccann, line 166: these tests will all fail in prebid 7 where you won't have access to storage manager unless the pub has whitelisted. please make your tests pb7 compatible.

patmmccann (changes requested):

test/spec/tpmnBidAdapter_spec.js
tpmn-admin: thx. comment. fix 'getUserSyncs' test case code, And  add storageAllowed policy from pb7

modules/tpmnBidAdapter.js
tpmn-admin: Sync users without using local storage.

modules/tpmnBidAdapter.js
tpmn-admin: Yes, checking "first party cookie".

I want to use user sync with "bid-switch" based on tpmn.

tpmn-admin: I think it would be better not to uuid for user sync, so I modified the code.

tpmn-admin:
sorry, @patmmccann , i did  "git pull upstream master" and then after commit.
now fix git history.
i commit, fix user sync. static user sync without checking uuid.

PR #8351, opened 2022-04-29
adding PAF userId and RTD modules

bwschmidt:
## Type of change

- [x] Feature

## Description of change

This PR adds userId and RTD modules for the prebid addressability framework MVP found at https://github.com/prebid/paf-mvp-implementation. The demo currently uses this fork: https://github.com/openx/Prebid.js/tree/paf. More information about the specifications of the ids and data can be found at https://github.com/prebid/addressability-framework.

The design uses the RTD module to generate a seed for each auction. A userId module is used to expose the actual PAF data. Although this could have been done with an RTD module only, two modules were created so that the format of the data given to the bidders would align with the OpenRTB spec found at https://github.com/prebid/addressability-framework/blob/main/mvp-spec/ad-auction.md


patmmccann:
@bwschmidt can we break these into two PR's so the review can less intimidating?

bwschmidt:
@patmmccann sure I will split it

bwschmidt:
this has been split into https://github.com/prebid/Prebid.js/pull/8366 and https://github.com/prebid/Prebid.js/pull/8367

Issue #7781, opened 2021-11-30
Remove support for TCF1


bretg:
## Type of issue
cleanup 

## Description
The IAB deprecated TCF1 almost a year ago. Prebid Server removed support for TCF1 a while back. It's time PBJS stops supporting it.

dgirardi:
Delaying implementation of this until https://github.com/prebid/Prebid.js/pull/8185 gets merged.

PR #8356, opened 2022-05-02
Prebid 7: remove support for TCFv1

dgirardi:
## Type of change

- [x] Other

## Description of change

Remove support for TCFv1 (https://github.com/prebid/Prebid.js/issues/7781) - the consentManagement module:

- no longer looks for `window.__cmp`
- does not use the `allowAuctionWithoutConsent` config (which was respected only for v1)

## Other information

Closes https://github.com/prebid/Prebid.js/issues/7781
Docs PR: https://github.com/prebid/prebid.github.io/pull/3749


modules/consentManagement.js
```diff
@@ -147,92 +102,39 @@ function lookupIabConsent({onSuccess, onError, width, height}) {
...
-      cmpFunction('getVendorConsents', null, v1CallbackHandler.vendorConsentsCallback);
-    } else if (cmpVersion === 2) {
-      cmpFunction('addEventListener', cmpVersion, v2CmpResponseCallback);
-    }
-  } else if (cmpVersion === 1 && inASafeFrame() && typeof window.$sf.ext.cmp === 'function') {
-    // this safeframe workflow is only supported with TCF v1 spec; the v2 recommends to use the iframe postMessage route instead (even if you are in a safeframe).
-    logInfo('Detected Prebid.js is encased in a SafeFrame and CMP is registered, calling it now...');
-    callCmpWhileInSafeFrame('getConsentData', v1CallbackHandler.consentDataCallback);
-    callCmpWhileInSafeFrame('getVendorConsents', v1CallbackHandler.vendorConsentsCallback);
+    cmpFunction('addEventListener', CMP_VERSION, v2CmpResponseCallback);
```
osazos, line 105: The function named `v2CmpResponseCallback` should be renamed as `cmpResponseCallback` (as you did for  ~`checkV2Data`~  ->  `checkData`)

osazos (commented):
Hello @dgirardi

Nothing to say about the logic of module itself, I mentioned some optional consistency improvements. 

Perhaps, you can add a test to cover code at line 190 :
```
 if (consentTimeout === 0) {
      processCmpData(undefined, callbacks);
    } else {
```

Also : 
in function loadConsentData(cb), line 169 :
`if (cb != null) { `,  we should check if `cb` is a function `typeof cb === 'function'`

dgirardi:
> in function loadConsentData(cb), line 169 :
> if (cb != null) { , we should check if cb is a function typeof cb === 'function'

I don't think it's necessary as there's no "public" access to that point, and this module always passes a function or a null-y value. But it doesn't hurt either.

Fawke (commented):
Hi @dgirardi,
This change looks mostly good. Just one feedback, I know a few places in the `gdprEnforcement` module where we still have checks for tcf 1.1. It's just a few else conditions I think. 
For example, over [here](https://github.com/prebid/Prebid.js/blob/e700568fbb7d8ab3cac257fea566aeaeb657aae5/modules/gdprEnforcement.js#L323)

I think it would be better to clean up the code a bit in that module as well. Thanks.

dgirardi:
Opened separate PR for the examples - https://github.com/prebid/Prebid.js/pull/8525 - as I think they need updating anyway and we may need some more coordination to figure out which CMP to use.

osazos (approved):
LGTM, Thanks for your work @dgirardi

Issue #8283, opened 2022-04-13
Prebid 7: remove support for legacy FPD


dgirardi:
## Description

Remove support for `setConfig('fpd')`, `config.getLegacyFpd`, `config.convertAdUnitFpd` and related relics of legacy fpd

Related: https://github.com/prebid/Prebid.js/issues/8264

patmmccann:
#7796 for linkage

dgirardi:
going through adapters for an unrelated issue, I noticed some uses of legacy fpd that are not trivial to remove. For example:

https://github.com/prebid/Prebid.js/blob/2c0518f38d505a8749eafba0b34932b26b4685eb/modules/adrelevantisBidAdapter.js#L138-L145

depending on how many more like these there are, it may be better to reduce the scope of this change to removing just the portion that is public in the publisher API.

patmmccann:
taxonomy committee indicates support to include this in 7

PR #8372, opened 2022-05-04
Prebid 7: remove support for legacy FPD

dgirardi:
## Type of change

- [x] Other

## Description of change

This removes support for legacy FPD:

 - `setConfig('fpd')` is no longer converted to an equivalent `setConfig('ortb2')`
 - `adUnit.fpd` is no longer converted to an equivalent `adUnit.ortb2Imp`
 - `getLegacyFpd` and `getLegacyImpFpd`	have been moved away from `config` to standalone utils functions, and their use avoided when possible

I have kept the ortb2-to-legacy conversion logic available because of one, or possibly two adapters:

- `inmar` (@zandree-owneriq), sends legacy-formatted FPD verbatim to its backend; 
- `luponmedia` (@adxpremium) may have intended to do the same - but the implementation should not work even in version 6 (see code comments in this PR)

## Other information

Closes https://github.com/prebid/Prebid.js/issues/8283
Documentation PR: TBD, together with other Prebid 7 FPD-related changes (https://github.com/prebid/Prebid.js/issues/7651)


modules/inmarBidAdapter.js
```diff
@@ -40,7 +41,7 @@ export const spec = {
       uspConsent: bidderRequest.uspConsent,
       currencyCode: config.getConfig('currency.adServerCurrency'),
       coppa: config.getConfig('coppa'),
-      firstPartyData: config.getLegacyFpd(bidderRequest.ortb2),
+      firstPartyData: getLegacyFpd(bidderRequest.ortb2),
```
patmmccann, line 44: @zandree-owneriq we welcome your feedback on if you still need the data formatted this way or if you would prefer the newer ortb2 formatting?

modules/luponmediaBidAdapter.js
```diff
@@ -453,6 +457,8 @@ function newOrtbBidRequest(bidRequest, bidderRequest, currentImps) {
     deepSetValue(data, 'ext.prebid.bidderconfig.0', bidderData);
   }
 
+  // TODO: bidRequest.fpd is not the right place for pbadslot - who's filling that in, if anyone?
+  // is this meant to be bidRequest.ortb2Imp.ext.data.pbadslot?
```
patmmccann, line 461: @adxpremium fyi

src/utils/legacyFPD.js
```diff
@@ -0,0 +1,48 @@
+import {deepAccess, mergeDeep} from '../utils.js';
+
+/**
+ * Returns backwards compatible FPD data for modules
```
patmmccann, line 4: my only thought here is if it turns out only one adapter is using, eg inmar, would it be better to put this in the inmar adapter itself?

patmmccann (approved):

src/utils/legacyFPD.js
dgirardi: yes, that would avoid extra weight for anyone who's not using inmar. I'll update if it turns out that luponmedia does not need the legacy conversion logic. Ideally, inmar would also drop the old format eventually.

patmmccann:
#7796 for linkage

PR #8373, opened 2022-05-04
Prebid 7: change default of `allowAlternateBidderCodes` from `true` to `false`

dgirardi:
## Type of change

- [x] Other

## Description of change

Change default of `bidderSettings.allowAlternateBidderCodes` to false

## Other information

Change mentioned in https://github.com/prebid/Prebid.js/issues/7796
Documentation PR: not needed, default change already mentioned in https://github.com/prebid/prebid.github.io/pull/3721/files

### DIFF
--- a/src/adapters/bidderFactory.js
+++ b/src/adapters/bidderFactory.js
@@ -257,10 +257,10 @@ export function newBidder(spec) {
   });
 
   function isInvalidAlternateBidder(responseBidder, requestBidder) {
-    let allowAlternateBidderCodes = bidderSettings.get(requestBidder, 'allowAlternateBidderCodes');
+    let allowAlternateBidderCodes = bidderSettings.get(requestBidder, 'allowAlternateBidderCodes') || false;
     let alternateBiddersList = bidderSettings.get(requestBidder, 'allowedAlternateBidderCodes');
     if (!!responseBidder && !!requestBidder && requestBidder !== responseBidder) {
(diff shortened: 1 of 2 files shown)


patmmccann (approved):

bretg:
I think we need two more bits of documentation:
1) something in the bid adapter doc noting the need to set biddercode and providing a pointer to the allowAlternateBidderCodes setting.
2) the 7.0 release notes need to highlight this as a potential breaking change

pm-azhar-mulla:
Just curious to know, how a bidder can determine default value for `bidderSettings.allowAlternateBidderCodes`.
Since in v6.x `undefined` means `true` and v7.x `undefined` means `false`.

pm-harshad-mane:
I agree with @pm-azhar-mulla , the bidders should not receive `undefined`, can we provide a `boolean` value to bidders?

dgirardi:
I do not understand what you mean - why and how would bidders be interested in the default value? or for that matter, the value of that setting at all?

patmmccann:
the value of the setting lets them know if they can respond as groupm or not, if they do so but they are not allowed, they are wasting a response

https://github.com/prebid/Prebid.js/blob/e47b9081ef3f6283053c46dd432a6d593d9b3558/modules/pubmaticBidAdapter.js#L1152

patmmccann:
[pm-azhar-mulla](https://github.com/pm-azhar-mulla) you should do

 payload.ext.allowAlternateBidderCodes = bidderSettings.get(bidderRequest.bidderCode, 'allowAlternateBidderCodes') || true; in your 6.x adapter and 

 payload.ext.allowAlternateBidderCodes = bidderSettings.get(bidderRequest.bidderCode, 'allowAlternateBidderCodes')  || false; in your 7.x adapter

Issue #8374, opened 2022-05-05
Uncaught (in promise) TypeError: Cannot read properties of undefined (reading 'makeBidRequests')


audiencerun:
## Type of issue
Bug

## Description

We wanted to update our version of Prebid.js from **6.12.0** to the latest version **6.22.0**. Once updated, we noticed this error in the console

`Uncaught (in promise) TypeError: Cannot read properties of undefined (reading 'makeBidRequests')`

![IsawNCGP](https://user-images.githubusercontent.com/57719351/166870194-563af347-c30b-4a0f-8155-22d6a013906e.png)

After several tests, it seems that we have the error when switching to Prebid.js starting **6.17.0** and higher. The possibilities we see in the diffs may be related to the `adapterManager.js` file via [Prebid core & PBS adapter: better support for server-side stored impressions by dgirardi · Pull Request #8154 · prebid/Prebid.js](https://github.com/prebid/Prebid.js/pull/8154/files) ?

or to the webpack upgrade from **3.0.0** to **5.70.0** via [Build system: upgrade webpack by dgirardi · Pull Request #7935 · prebid/Prebid.js](https://github.com/prebid/Prebid.js/pull/7935/files) ?

we have tried some fixes, here we import `adapterManager.js` via a require 

[Prebid.js/auction.js at e07025247a81bf8c04501de8c66d817143cc30d1 · prebid/Prebid.js](https://github.com/prebid/Prebid.js/blob/e07025247a81bf8c04501de8c66d817143cc30d1/src/auction.js#L79) 

we added some logs to understand, and replaced `require` with an `import`

`adapterManager` => `import adapterManager from './adapterManager.js';`

`_adapterManager` => `const _adapterManager = require('./adapterManager.js').default;`

We see that on the second call of `callBids`, `adapterManager` becomes `undefined` when imported with a `require` while it is not the case when imported with an `import`. 

But we don't know if this is the real cause of the error.

![image](https://user-images.githubusercontent.com/57719351/166871183-4fd8a76c-2d3d-4121-b8c5-24a362efa01b.png)

## Steps to reproduce
- Go to https://7buoei.vercel.app/test.html
[...]

PR #8379, opened 2022-05-05
Core & Multiple modules: replace all usage of `require` with `import`

dgirardi:
## Type of change

## Description of change
﻿We had multiple reports of an elusive issue that seems to be related to use of `require` (https://github.com/prebid/Prebid.js/issues/8374).

Since mixed use of `require` and `import`s is not great anyway, this is a good opportunity to clean it up.

## Other information
Related: https://github.com/prebid/Prebid.js/issues/8374 (although I'd like to understand more before calling this a fix)

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -45,12 +45,19 @@ module.exports = {
     'no-throw-literal': 'off',
     'no-undef': 2,
     'no-useless-escape': 'off',
-    'no-console': 'error'
+    'no-console': 'error',
   },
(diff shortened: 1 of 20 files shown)


jsnellbaker (approved):

dgirardi:
An update in https://github.com/prebid/Prebid.js/issues/8374#issuecomment-1120050021 - the gist of it is, loading Prebid multiple times causes problems that will sometimes break `require` but not `import`.

ChrisHuie (approved):

PR #8391, opened 2022-05-09
BeOp Bid Adapter: prefer canonical URL when present & prepend protocol

cyppan:
## Type of change

- [ ] Bugfix
- [x] Feature
- [ ] New bidder adapter   
- [ ] Build related changes
- [ ] CI related changes
- [ ] Other

## Description of change

Prefer the canonical url when present when setting the payload url for the BeOp adapter.

We've noticed at least sometimes the protocol is missing (not sure where the problem is) so we prepend it when needed.

### DIFF
--- a/modules/beopBidAdapter.js
+++ b/modules/beopBidAdapter.js
@@ -36,11 +36,16 @@ export const spec = {
     */
   buildRequests: function(validBidRequests, bidderRequest) {
     const slots = validBidRequests.map(beOpRequestSlotsMaker);
-    let pageUrl = deepAccess(window, 'location.href') || deepAccess(bidderRequest, 'refererInfo.canonicalUrl') || config.getConfig('pageUrl');
-    let fpd = config.getLegacyFpd(config.getConfig('ortb2'));
-    let gdpr = bidderRequest.gdprConsent;
-    let firstSlot = slots[0];
-    let payloadObject = {
+    let pageUrl = deepAccess(bidderRequest, 'refererInfo.canonicalUrl') || config.getConfig('pageUrl') || deepAccess(window, 'top.location.href') || deepAccess(window, 'location.href');
+    // Ensure the protocol is present (looks like sometimes canonicalUrl misses it)
+    if (pageUrl != null) {
+      const defaultProtocol = deepAccess(window, 'top.location.protocol') || deepAccess(window, 'location.protocol');
+      pageUrl = ensureProtocolInUrl(pageUrl, defaultProtocol);
+    }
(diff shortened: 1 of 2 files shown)


patmmccann:
deepAccess(bidderRequest, 'refererInfo.canonicalUrl') || config.getConfig('pageUrl') doesnt make sense to me; as pageUrl affects deepAccess(bidderRequest, 'refererInfo.canonicalUrl')

sebrobert:
> deepAccess(bidderRequest, 'refererInfo.canonicalUrl') || config.getConfig('pageUrl') doesnt make sense to me; as pageUrl affects deepAccess(bidderRequest, 'refererInfo.canonicalUrl')

Is that true for older version of prebid.js ? If we want this fix deployed in 6.x and 7.x versions, was that one already available ?

dgirardi:
For 6 I would use `bidderRequest.refererInfo.canonicalUrl`, then `window.top.location.href`, then `window.location.href`. But I think you need to wrap the second in a try-catch - in my testing it will throw an exception if it crosses origins.

`bidderRequest.referer` *almost* does that, but it returns something else (the referer) if it's running from  a cross-origin frame. Better to avoid its use for now, that logic will be reworked in 7.

the reason `canonicalUrl` sometimes misses the protocol is likely that the publisher did not provide it in either `setConfig({pageUrl})` or `<link rel="canonical" href="<pageUrl>" />`.

cyppan:
Thanks @dgirardi, I've modified following your advice.

cyppan:
Not sure why the tests fail, seems to come from another adapter test file 🤔

ChrisHuie (approved):

ChrisHuie:
Sorry, ignore my review. I was changing the title and tabbed too far

modules/beopBidAdapter.js
```diff
@@ -141,4 +143,47 @@ function beOpRequestSlotsMaker(bid) {
...
+    return null;
+  }
+}
+
+function getPageUrl(refererInfo, window) {
+  refererInfo = refererInfo || getRefererInfo();
+  let pageUrl = refererInfo.canonicalUrl || safeDeepAccess(window, 'top.location.href') || deepAccess(window, 'location.href');
+  // Ensure the protocol is present (looks like sometimes the extracted pageUrl misses it)
+  if (pageUrl != null) {
+    const defaultProtocol = deepAccess(window, 'top.location.protocol') || deepAccess(window, 'location.protocol');
```
dgirardi, line 183: 'top.location.protocol' can also throw.

PR #8392, opened 2022-05-09
Prebid Server Bid Adapter: Pass floors signal to PBS if Price Floors enabled for auction

robertrmartinez:
## Type of change
- [X] Other

## Description of change
If the price floors module is active, we send a signal to PBS indicating so.

This includes both when floor enforcement is skipped or not.

### DIFF
--- a/modules/prebidServerBidAdapter/index.js
+++ b/modules/prebidServerBidAdapter/index.js
@@ -849,6 +849,11 @@ Object.assign(ORTB2.prototype, {
       }
     };
 
+    // If the price floors module is active, then we need to signal to PBS! If floorData obj is present is best way to check
+    if (typeof deepAccess(firstBidRequest, 'bids.0.floorData') === 'object') {
+      request.ext.prebid.floors = { enabled: false };
+    }
+
     // This is no longer overwritten unless name and version explicitly overwritten by extPrebid (mergeDeep)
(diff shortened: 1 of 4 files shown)


modules/prebidServerBidAdapter/index.js
```diff
@@ -849,6 +849,11 @@ Object.assign(ORTB2.prototype, {
       }
     };
 
+    // If the price floors module is active, then we need to signal to PBS! If floorData obj is present is best way to check
+    if (typeof deepAccess(firstBidRequest, 'bids.0.floorData') === 'object') {
+      request.ext.prebid.floors = { enabled: false };
```
patmmccann, line 854: if floor data exists, wouldnt enabled be true? does this mean deactivate the serverside floor bc a client side floor was chosen?

robertrmartinez: I too thought about that haha

But I think this is a signal saying "Do not run Price Floors logic on PBS for this Auction"

@bretg to confirm!

bretg: right - ext.prebid.floors.enabled:false will turn off the server-side floors feature, causing it to default to imp.bidfloor.

mmoschovas (approved):

patmmccann (approved):

Issue #7626, opened 2021-10-25
More dynamic floormin


patmmccann:
## Type of issue
Feature

## Description
https://github.com/prebid/Prebid.js/pull/5805 allowed for minimums to floors as a page level variable, but we need floorMin at slot, and potentially other granularity, etc level.

robertrmartinez:
So the change is to:

Update the price floors module to look for floorMin at `adunit.ortb2Imp.ext.prebid.floorMin` first, and then do whatever it does now to find it if not?

@bretg @patmmccann

patmmccann:
yes; this would allow for it to vary over ad units, no?

robertrmartinez:
Yep it is a good idea!

I'll try and get this and https://github.com/prebid/Prebid.js/issues/8339 done over the weekend or something!

muuki88:
This looks super useful. Does this also work if there's no floor information at all?

Currently we set our floors something like this

```js
floors: {
  currency: config.currency.adServerCurrency,
  schema: {
    delimiter: '|',
    fields: ['mediaType']
  },
  values: {
    // provided from somewhere else
    '*': priceRule.floorprice
  }
}
```

With this extension would it be possible to leaf the `floors` property undefined and only set

```js
ortb2Imp: {
  ext: {
    prebid: {
      floorMin: priceRules.floorprice
    }
  }
}
```

patmmccann:
I got so excited i merged without docs, apologies :)

we're implementing shortly.

patmmccann:
@muuki88 did you figure out the answer to that in test?

muuki88:
@patmmccann unforunately not. I'll put this on my teams agenda

patmmccann:
I haven't tested, but I suspect that would not work. You might be able to get away with just the currency or something though.

PR #8396, opened 2022-05-09
Price Floors Module: Support adUnit level floors

robertrmartinez:
## Type of change
- [X] Feature

## Description of change
Implements https://github.com/prebid/Prebid.js/issues/7626

If floor exists on adunit at `adUnit.ortb2Imp.ext.prebid.floorMin` we will use that as floorMin, otherwise use the top level one as before!


patmmccann:
tyvm!

mmoschovas (approved):

patmmccann (approved):

PR #8399, opened 2022-05-10
Improve Digital adapter: adding Extend mode

jbartek25:
## Type of change
- [x] Feature

## Description of change
Improve Digital adapter: adding Extend mode for additional demand. Currently the Extend mode sends bid requests to Improve Digital PBS. The type of the backend for the Extend mode can change in the future.

### DIFF
--- a/modules/improvedigitalBidAdapter.js
+++ b/modules/improvedigitalBidAdapter.js
@@ -9,8 +9,11 @@ import {Renderer} from '../src/Renderer.js';
 import {createEidsArray} from './userId/eids.js';
 
 const BIDDER_CODE = 'improvedigital';
-const REQUEST_URL = 'https://ad.360yield.com/pb';
 const CREATIVE_TTL = 300;
+
+const AD_SERVER_URL = 'https://ad.360yield.com/pb';
+const EXTEND_URL = 'https://pbs.360yield.com/openrtb2/auction';
+const IFRAME_SYNC_URL = 'https://hb.360yield.com/prebid-universal-creative/load-cookie.html';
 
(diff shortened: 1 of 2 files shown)


patmmccann:
please see section 2.8 if you want to respond with multiple bidder codes https://docs.prebid.org/dev-docs/publisher-api-reference/bidderSettings.html

jbartek25:
> please see section 2.8 if you want to respond with multiple bidder codes https://docs.prebid.org/dev-docs/publisher-api-reference/bidderSettings.html

@patmmccann yes, we considered this. For the initial launch we decided to always respond with `improvedigital` bidder as the Extend mode is aimed at our customers who require a simple blackbox solution. Obviously, this might change in the future based on the customer feedback.

osazos:
Hi @jbartek25 

It is ok for me regarding the code review. 

Is there a specific reason to not add the `extend` and `singleRequest` config params in the doc? They should be.

jbartek25:
thanks @osazos. good point about the doc, I opened a doc PR for `extend`: https://github.com/prebid/prebid.github.io/pull/3782
`singleRequest` was deliberately removed from the doc after we discovered performance issues with single requests. It's kept in the code for the backward compatibility for any of the older publishers who still use it.

osazos:
@jbartek25 don't you want to add the `extend` param in the .md file too?

jbartek25:
> @jbartek25 don't you want to add the `extend` param in the .md file too?

I don't see a reason why. The .md file just provide basic ad units, the full set of features is covered by unit tests.

osazos (approved):
LGTM

PR #8408, opened 2022-05-11
UserID module: do not start initialization until `pbjs.processQueue()` has been called

dgirardi:
## Type of change

- [x] Bugfix

## Description of change

After https://github.com/prebid/Prebid.js/pull/8201, the userID module will start initialization as soon as it gets configured with `setConfig`. If that happens before `pbjs.processQueue` is called (currenly only possible with a custom build), it throws an exception when it tries to use hooks that are not ready.


jsnellbaker:
@dgirardi When you get the chance, can you resync with `master` and resolve the conflicts?  I think these changes look fine otherwise.

jsnellbaker (approved):

ChrisHuie (approved):

Issue #5001, opened 2020-03-18
Prebid.less Proposal


bwschmidt:
## Type of issue
Intent to Implement

## Description
Prebid.less Proposal

MARKET SITUATION & PROBLEM STATEMENT

Over the last several years, Prebid.js has cemented itself as the defacto header bidding solution. However, maintaining and configuring Prebid has proven cumbersome for many publishers. There is a consistent need to update Prebid core code along with a multitude of adapters. Secondly, there is a huge setup and maintenance cost related to keeping bidder parameters and adunit mapping up to date with every demand source a publisher wants to use. The market’s response to this has been a myriad of managed solutions. These solutions offload much of this complexity including keeping the source up to date and managing adunit configuration.

There are other demand aggregation solutions in the space without these problems, most notably Google Open Bidding and A9 TAM. These solutions integrate completely transparently to the publisher, requiring no adapters and no adunit mapping. This offloads the vast majority of the setup work to demand sources and saves publishers critical time and resources. This is accomplished by:

- Standardizing on OpenRTB
- Requiring exchanges to handle adunit mapping

These solutions have proven that demand sources (e.g. Exchanges, SSPs) can and will implement standardized implementations in order to not lose out on monetization. 

OPPORTUNITY

By standardizing on OpenRTB and requiring demand sources to handle ad unit mapping for publishers we feel that Prebid can close a critical benefit gap between Google Open Bidding and A9 TAM ultimately driving the following benefits:

- Increased publisher adoption of Prebid from publishers lacking development and adops resources 
- Improved publisher revenue from increased experimentation
- Reduction of publisher time & resources expended on maintaining Prebid
- Reduced effort required for publishers to comply with future industry standards and regulations (e.g. CCPA, supplychain object)

CONCEPTUAL DESIGN

1. Prebid.less participants would be required to implement OpenRTB endpoints for both client and server-side requests
⋅⋅ The client-side endpoint would be required to read IP and cookies from the header
⋅⋅ The endpoint would utilize standard OpenRTB, and any applicable global extension fields used by Prebid.js
2. Prebid.js and Prebid server would no longer have bid adapters, but instead have 1 single adapter that calls each demand source in the exact same way.
3. Publishers should be able to select which demand sources are called for which adunit on the page.
⋅⋅ Publishers should be able to opt demand sources in or out of sensitive fields in the OpenRTB request, e.g. user.ext.eids
[...]

PR #8409, opened 2022-05-11
Openx Ortb2 Bid Adapter: initial release

bwschmidt:
## Type of change
- [x] New bidder adapter 

## Description of change
OpenX is working to replace its browser serving architecture and moving to an OpenRTB based protocol. We would like to offer a 2nd adapter which we will test with publishers. Once we have validated the new architecture the goal is to replace the current adapter with this one. The interface for publishers remains the same.

I am not sure if this has been done before and if there are comments or concerns. In my testing it seemed that overloading the bidder code did not result in any problems, the build will just use the one that is last alphabetically if both adapters are included in the build. Another approach would be to use a new code like `openx2` however for ease of testing we are hoping publishers would only have to change their build, not their webpages.

```javascript
            bids: [{
                    bidder: 'openx',
                    params: {
                        delDomain: 'sademo-d.openx.net',
                        unit: '540096529',
                        test: true
                    }
               }]
```

- contact email of the adapter’s maintainer prebid@openx.com

### DIFF
--- a/modules/openxBidAdapter.md
+++ b/modules/openxBidAdapter.md
@@ -8,7 +8,9 @@ Maintainer: team-openx@openx.com
 
 # Description
 
-Module that connects to OpenX's demand sources
+Module that connects to OpenX's demand sources.
+Note there is an updated version of the OpenX bid adapter called openxOrtbBidAdapter.
+Publishers are welcome to test the other adapter and give feedback. Please note you should only include either openxBidAdapter or openxOrtbBidAdapter in your build.
 
 # Bid Parameters
 ## Banner
--- a/modules/openxNewBidAdapter.js
+++ b/modules/openxNewBidAdapter.js
@@ -1,0 +1,349 @@
+import {config} from '../src/config.js';
+import {registerBidder} from '../src/adapters/bidderFactory.js';
+import * as utils from '../src/utils.js';
+import {BANNER, VIDEO} from '../src/mediaTypes.js';
+import includes from 'core-js-pure/features/array/includes.js'
+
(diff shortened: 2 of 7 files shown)


patmmccann:
Hey @bwschmidt 

I want to pitch an idea and see if you guys would be up for it. 

Right now if a publisher has, say 6 openrtb based adapters, they each have lots of redundant code. How about a generic openrtb bidder module that your new adapter would be a submodule of. The submodule would import the basic openrtb adapter and register certain things like your gvl id and any meta parameter return handling etc, but most of its code would just be imported. This way, if something gets added to the base adapter, your sub-adapter would automatically take advantage, eg regs.gpc or device.sua. This seems to go a long way towards achieving some of the goals of https://github.com/prebid/Prebid.js/issues/5001

patmmccann:
One minor point, you might now want to call this openxnew, bc then we might end up with things like openx_new_new_final_final4real_v4

bwschmidt:
Yeah that is a good idea, i'll take a stab at it and think of a better name

mmoschovas (approved):
lgtm but holding off on merging based on the conversation

bwschmidt:
This PR has been updated with a name that makes more sense and also updated with prebid 7 FPD changes. Docs have been updated at https://github.com/prebid/prebid.github.io/pull/3836.

@patmmccann I am still interested in moving some of this code into a util. It wasnt as straightforward as I hoped as adapters split requests in different ways, as an example OpenX combines banner imps and splits video into different requests. I would like to get this released and come back to that work.

@mmoschovas please take a look at the small changes and see if it still looks good

patmmccann (approved):

PR #8410, opened 2022-05-12
Prebid Core Storage Manager: add gdpr check before device write tests to ensure purpose1 gets respected

renebaudisch:
## Type of change
- [x] Other: GDPR insurance update

## Description of change
implemented getting consent before trying to write to storage

### DIFF
--- a/modules/ixBidAdapter.js
+++ b/modules/ixBidAdapter.js
@@ -17,12 +17,12 @@ import {
   parseGPTSingleSizeArray,
   parseQueryStringParameters
 } from '../src/utils.js';
-import {BANNER, VIDEO} from '../src/mediaTypes.js';
+import {BANNER, VIDEO, NATIVE} from '../src/mediaTypes.js';
 import {config} from '../src/config.js';
 import CONSTANTS from '../src/constants.json';
 import {getStorageManager, validateStorageEnforcement} from '../src/storageManager.js';
 import * as events from '../src/events.js';
-import {find, includes} from '../src/polyfill.js';
+import {find} from '../src/polyfill.js';
 import {registerBidder} from '../src/adapters/bidderFactory.js';
 import {INSTREAM, OUTSTREAM} from '../src/video.js';
 import {Renderer} from '../src/Renderer.js';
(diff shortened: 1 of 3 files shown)


renebaudisch:
I would need assistance for that. The test all access storage regardless of consent, but since tcf one cannot rely on granted access so all of the tests are some kind of incomplete as they do not respect the users choice. IMO, either the test should use a mockedup constent string or anything else, but I'm not able to take care of this.

patmmccann:
Our model has been to put stuff like this in the gdpr module(s), eg the enforcement module, as many publishers don't load gdpr related code outside of gdpr countries.

patmmccann:
@renebaudisch do you have concerns with the enforcement module or the device access flag now? It's not clear what issue this pr is trying to solve.

renebaudisch:
The problem we need to solve is that the "userSync" module initially calls "cookiesAreEnabled" from the sessionManager what right now is writing and reading the device storage regardless of consent. We need to ensure that this check respects the given laws and if needed checks consent before doing this write test.

patmmccann:
https://github.com/prebid/Prebid.js/issues/6289#issuecomment-781989848

Indicates this was solved in #6323

renebaudisch:
@patmmccann  
Please have a look of what has been changed in #6323. 
That only removes the written entry, but the privacy issue that it get's written without asking the cmp is still there, commented in this issue:
https://github.com/prebid/Prebid.js/pull/6323#issuecomment-785178095

This PR is to not let pbjs write cookieTest without having consent amd addresses this privacy part of #6323.

patmmccann:
@Fawke could you comment? Do we need to reopen 6289?

src/storageManager.js
```diff
@@ -167,13 +176,26 @@ export function newStorageManager({gvlid, moduleName, bidderCode, moduleType} =
     }
   }
 
+  /**
+   * @param {string} consentData
+   */
+  const hasPurposeOneConsent = function(consentData) {
```
jsnellbaker, line 182: Hi @renebaudisch 

There is already a similar function in `src/utils/gpdr.js` under the name `hasPurpose1Consent`.  Would you be able to use that function instead of redefining it here?

src/userSync.js
```diff
@@ -315,7 +315,9 @@ export function newUserSync(userSyncDependencies) {
   return publicApi;
 }
 
-const browserSupportsCookies = !isSafariBrowser() && storage.cookiesAreEnabled();
+const browserSupportsCookies = function() {
```
jsnellbaker, line 318: Does the type of this variable really need to change to a function, instead of its boolean resolution?

If so, can you please update the corresponding unit tests in `test/spec/userSync_spec.js` to update the signature references it has for this param (from a boolean to a function)?

renebaudisch: This way the check is done later. At the time it is done now, the cmp(-module?) isn't ready, or at least `gdprDataHandler.getConsentData()` doesn't deliver anything. When called at time needed, the module or the cmp is set and we get response.

I'll update the PR

Issue #8421, opened 2022-05-17
Multiple adapters: use our tag insertion methods


patmmccann:
## Type of issue
Bug

## Description
Several adapters appear to not be using our methods to append scripts or tags to the DOM. Please switch them over

https://github.com/prebid/Prebid.js/blob/747f1430f33f69ca683198ae7ca8c10473ff7ce9/modules/cleanioRtdProvider.js#L58
@nick-clean 

https://github.com/prebid/Prebid.js/blob/3ef3cf4d1a62a43ae314edb21b1f054ace8a764a/modules/airgridRtdProvider.js#L47 @ydennisy 

https://github.com/prebid/Prebid.js/blob/78e1bdb2dd921930e24ef1b494a7e89f84e862a3/modules/realvuAnalyticsAdapter.js#L172 @tchibirev

ftxmoJason:
Thanks @patmmccann for pointing this out. Can you direct us to the correct method definition in the docs? Thanks

patmmccann:
> Thanks @patmmccann for pointing this out. Can you direct us to the correct method definition in the docs? Thanks

https://github.com/prebid/Prebid.js/commit/2a05213fb73d88ac939c0b0dd4b5750649446372 is an example; using the loadExternalScript utility

patmmccann:
https://github.com/prebid/Prebid.js/pull/7699 is an example of a fix to a tag

parisholley:
@patmmccann we only use that to write image pixels for measuring viewability when the browser does not support the IntersectionObserver API. is there a recommended way to drop non-tracking pixels? loadExternalScript does not seem to be relevant in this case.

anthonylauzon:
@patmmccann we have a fix for this that you should expect to be submitted as a PR soon.  Please direct future bugfix requests to @abhibond or @jlaso.

jimdigriz:
What is the official method to preload a script here?

I cannot find anything obvious:

```alex@aineko:~/src/adloox/Prebid.js$ git grep "'preload'" modules/ src/
modules/adlooxAnalyticsAdapter.js:  link.setAttribute('rel', 'preload');
modules/cleanioRtdProvider.js:  linkElement.rel = 'preload';
modules/geoedgeRtdProvider.js:  link.rel = 'preload';
alex@aineko:~/src/adloox/Prebid.js$ git grep '"preload"' modules/ src/
alex@aineko:~/src/adloox/Prebid.js$ ```

[...]

PR #8432, opened 2022-05-18
Ftrack Id Module: replace native appendChild with Prebid's loadExternalScript

ftxmoJason:
## Type of change
- [x] Bugfix

## Description of change
Replacing the use of appendChild in code in favor of loadExternalScript per https://github.com/prebid/Prebid.js/issues/8421


patmmccann:
you also need to fix this asap  https://github.com/prebid/Prebid.js/blob/14ae8d5bc28780d7cd64b4913935326ada58b595/modules/ftrackIdSystem.js#L105  or we'll have to revert your module from the project

we asked that your url was configurable, but you check that it is exactly something or you dont insert

ftxmoJason:
@patmmccann We'll spin up a different PR to address this change. Working on getting our file on CDN yada yada and these two things don't need to be intertwined.

ftxmoJason:
@patmmccann Please see https://github.com/prebid/Prebid.js/pull/8473 for the fix you called out. Thanks

patmmccann:
thank you very much, I merged that. I am confused why using our method versus your own to load a script would have a cdn dependency

ftxmoJason:
Maybe we misread to Policy and such, but we thought one of the steps to get whitelisted was being on an approved CDN. What do we have to do to get https://d9.flashtalking.com/d9core approved/whitelisted? Do we just have to ask to be added to the `_approvedLoadExternalJSList` of https://github.com/prebid/Prebid.js/blob/master/src/adloader.js? @patmmccann

patmmccann:
You shouldn't refer to that code from your adapter, it should only be a configuration parameter

patmmccann:
And yes, in your pr, ad yourself to that list

ftxmoJason:
@patmmccann I think we are good now. Moved this PR out of draft.

modules/ftrackIdSystem.js
```diff
@@ -102,9 +103,9 @@ export const ftrackIdSubmodule = {
         }
 
         if (config.params && config.params.url) {
-          var ftrackScript = document.createElement('script');
-          ftrackScript.setAttribute('src', config.params.url);
-          window.document.body.appendChild(ftrackScript);
+          var ftrackOnload = function (evt) {}
```
dgirardi, line 106: I don't think this is necessary? loadExternalScript will attempt to call it only if it's defined.

ftxmoJason: Thanks @dgirardi . You are correct, I removed this condition as well as the 3rd and 4th parameters in our loadExternalScript reference as well since we aren't utilizing them.

smenzer (approved):
LGTM

PR #8435, opened 2022-05-19
Tnc id module: initial release

annavane:
## Type of change

- [x ] Feature
- [ x] Does this change affect user-facing APIs or examples documented on http://prebid.org?

## Description of change
New UserId Module: TNCIDIdSystem

Maintainer contact: anna.vane@thenewco.it


modules/tncIdSystem.js
```diff
@@ -0,0 +1,85 @@
...
+    let i = providerId;
+    let d = tncOptions;
+
+    t[c] = t[c] || {}; t[c].providerId = i; t[c].options = t[c].options || d;
+    var f = 'ready'; t[c][f] = t[c][f] || function (z) {
+      t[c][f].q = t[c][f].q || []; t[c][f].q.push(z);
+      return t[c];
+    }; var s = n.createElement('script'); s.setAttribute('global', c); s.async = !0;
+    s.defer = !0; s.id = 'tnc_route'; s.src = 'https://js.tncid.app/route.js';
+    n.querySelector('head,body').appendChild(s);
```
smenzer, line 33: I think you should be using the built in `loadExternalScript()` method to load your script... @patmmccann or @dgirardi do you have thoughts on this?

modules/.submodules.json
```diff
@@ -43,7 +43,8 @@
     "unifiedIdSystem",
     "verizonMediaIdSystem",
     "zeotapIdPlusIdSystem",
-    "adqueryIdSystem", 
+    "tncIdSystem",
```
patmmccann, line 46: most of these are alphabetical

modules/tncIdSystem.js
annavane: Thank you so much for your feedback.

We do not use loadExternalScript function or endpoint because of the particularity of our UserID script. We do not save UserID in cookies or localstorage but it is stored in its own namespace on a global window object.

As we would like to use a separate namespace for prebit script and we need to let our external script know where to look.

In modules/tncIdSystem.js:

    t[c] = t[c] || {}; t[c].providerId = i; t[c].options = t[c].options || d;
    var f = 'ready'; t[c][f] = t[c][f] || function (z) {
      t[c][f].q = t[c][f].q || []; t[c][f].q.push(z);
      return t[c];
    }; var s = n.createElement('script'); s.setAttribute('global', c); s.async = !0;
    s.defer = !0; s.id = 'tnc_route'; s.src = 'https://js.tncid.app/route.js';
    n.querySelector('head,body').appendChild(s);

Current part of the code creates a script tag with an attribute named "global" where the name of the namespace is stored for the use of an external loaded script.

Please let us know if the current approach may violate prebit best practices or if you have some better solution to propose.

annavane:
@smenzer @patmmccann Can you please re-review this PR? We would like to know if any other changes are necessary.

modules/tncIdSystem.js
patmmccann: This isn't acceptable, perhaps you should just give publishers the location of your script and assume it's on page in this adapter?

annavane: Hi @patmmccann, thank you for your feedback. 

We made changes to our TNCID generating script and we now use loadExternalScript util function for loading our external script.  

Hope it now satisfies all Prebid coding rules.

annavane: Hi @smenzer ,

Thank you for your feedback. There are few examples of scripts with similar situations. Your precise indications are really useful.

We completely **removed** this part of code that creates script tag:

> t[c] = t[c] || {}; t[c].providerId = i; t[c].options = t[c].options || d;
> var f = 'ready'; t[c][f] = t[c][f] || function (z) {
>   t[c][f].q = t[c][f].q || []; t[c][f].q.push(z);
>   return t[c];
> }; var s = n.createElement('script'); s.setAttribute('global', c); s.async = !0;
> s.defer = !0; s.id = 'tnc_route'; s.src = 'https://js.tncid.app/route.js';
> n.querySelector('head,body').appendChild(s);

Now we use loadExternalScript function but only if client doesn't have our script already on page or if prebid script was executed first:

> const loadRemoteScript = (providerId) => {
>   return new Promise((resolve) => {
>     loadExternalScript('https://js.tncid.app/remote.js?ns=' + FALLBACK_TNC_INSTANCE + '&providerId=' + providerId, MODULE_NAME, resolve);
>   })
> }
> 
> const tncCallback = function (providerId, cb) {
>   let tncNS = '__tnc';
>   let promiseArray = [];
>   if (!window[tncNS] && !disableFallbackScript) {
>     tncNS = FALLBACK_TNC_INSTANCE;
>     promiseArray.push(loadRemoteScript(providerId));
>   }
> 
>   return Promise.all(promiseArray).then(() => waitTNCScript(tncNS)).then(cb).catch(() => cb(null));
> }

Moreover we added input param to disable load of fallback script for more control by publisher
> disableFallbackScript = (config && config.params && config.params.disableFallbackScript);

annavane:
@smenzer @patmmccann Can you please let us know if code is good to go now?

modules/tncIdSystem.js
smenzer: @patmmccann are you ok with this approach?

modules/userId/eids.js
```diff
@@ -315,11 +315,18 @@ export const USER_IDS_CONFIG = {
     }
   },
 
+  // tncId
+  'tncid': {
+    source: 'thenewco.it',
+    atype: 3
+  },
+
```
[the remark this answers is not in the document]
patmmccann: Not sure why this was resolved without changes

modules/tncIdSystem.js
```diff
@@ -0,0 +1,69 @@
...
+      tnc = window[tncNS];
+      if (tnc.tncid) resolve(tnc.tncid);
+      else tnc.on('data-sent', () => resolve(tnc.tncid));
+    });
+  });
+}
+
+const loadRemoteScript = (providerId) => {
+  return new Promise((resolve) => {
+    loadExternalScript('https://js.tncid.app/remote.js?ns=' + FALLBACK_TNC_INSTANCE + '&providerId=' + providerId, MODULE_NAME, resolve);
```
patmmccann, line 27: Thank you for converting to load external script, please also do either of the below:

1) open source and lock this script to a version
2) change its location to a configuration parameter with no default ( but feel free to mention the usual location in an integration example or documentation )

Thanks!!

modules/tncIdSystem.js
patmmccann: If they want to open source and lock that script similar, sure, otherwise we risk publishers interpreting it as vetted and endorsed by the project when in fact we haven't seen it and cannot warranty its contents

smenzer: ok that's what i thought, thanks @patmmccann

annavane:
Hi @smenzer and @jlukas79, do you have any additional requests or comments?

Thank you!

modules/tncIdSystem.js
patmmccann: Thank you!

patmmccann:
@jlukas79 could you take another look? Ty!

smenzer (approved):
LGTM

PR #8437, opened 2022-05-19
Prebid 7: Remove AppNexus Alias for Brealtime and DistrictM

patmmccann:
Districtm is now sharethrough and no longer supports this. They removed DMX in another pb7 PR. Brealtime is now emxdigital, which remains an alias with a gvl id after this pr, but I believe has migrated people to their own adapter. 

#7796 for linkage


jsnellbaker:
Thanks for putting this together @patmmccann

Issue #8161, opened 2022-03-09
Fix gdprenforcement module to look for generic purpose 1 consent when enforcing legacy pubcommonid and sharedid modules


JulieLorin:
## Type of issue
I think this is a bug

## Description
When using both gdprEnforcement (with enforcement or purpose and vendor for storage) and pubCommonId, pubCommonId is denied access to storage and won't be able to retrieve an ID

## Steps to reproduce
Just import both module. 
You can use this configuration in setConfig : 
```
consentManagement: {
      gdpr: {
        cmpApi: 'iab',
        timeout: 3 000,
        allowAuctionWithoutConsent: false,
        rules: [ {
          purpose: 'storage',
          enforcePurpose: true,
          enforceVendor: true
        }, {
          purpose: 'basicAds',
          enforcePurpose: false,
          enforceVendor: false
        } ]
      }
    }
```

### Expected results

I expect to have a pubcid when consent is given

### Actual results

I have a log live saying pubcid = null

## Platform details
Chrome

## Other information
I tried some debugging and what is happening is that pubCommonId creates a storageManager with no gvlid and no moduleName.
So when pubCommonId tries to write the ID in the local storage, because of those lines in gdprEnforcement module : 
```
const vendorConsent = deepAccess(consentData, `vendorData.vendor.consents.${gvlId}`); // vendorConsent will be undefined if 
const vendorAllowed = rule.enforceVendor === false || vendorConsent === true;
```

patmmccann:
@dgirardi after review this is indeed a bug, and a rather high priority one. 

The Pubcommonid does not have a vendor id, which is correct, so I believe we should just be checking consentData for purpose one consent with no specific vendor at all. I had to decode a tcf string to see if this was even a concept, it appears to be so, that there is generalized purpose one consent.

dgirardi:
@patmmccann right now the logic is "allow if both purpose and vendor are explicitly allowed". The fix seems to be "allow if purpose is allowed, and vendor is either allowed or unknown".  Is that right? it seems like a major change in the meaning of the consent string.

https://github.com/prebid/Prebid.js/blob/02264c8b3eddd04f58b19853ac3ebaa2ddfcacc1/modules/gdprEnforcement.js#L136-L157

bretg:
@JulieLorin - PubCommon has been deprecated in favor of SharedID since PBJS 5.0. So two comments:

[...]

PR #8448, opened 2022-05-23
Gdpr Enforcement module and sharedId/pubCommonId modules: vendor consent should not be enforced for first-party-id modules

goldun:
…

## Type of change

- [x] Bugfix
- [ ] Feature
- [ ] New bidder adapter   
- [ ] Build related changes
- [ ] CI related changes
- [ ] Other

## Description of change
Fix for https://github.com/prebid/Prebid.js/issues/8161 issue.
Added additional check in the gdprEnforment module: if the module type is `fpid-module`, then skip vendor allowed check

### DIFF
--- a/modules/gdprEnforcement.js
+++ b/modules/gdprEnforcement.js
@@ -18,6 +18,8 @@ const TCF2 = {
   'purpose7': { id: 7, name: 'measurement' }
 }
 
+const VENDORLESS_MODULE_TYPES = ['fpid-module'];
+
 /*
   These rules would be used if `consentManagement.gdpr.rules` is undefined by the publisher.
 */
@@ -123,9 +125,10 @@ function getGvlidForAnalyticsAdapter(code) {
  * @param {Object} consentData - gdpr consent data
  * @param {string=} currentModule - Bidder code of the current module
  * @param {number=} gvlId - GVL ID for the module
+ * @param {string=} moduleType module type
  * @returns {boolean}
  */
-export function validateRules(rule, consentData, currentModule, gvlId) {
+export function validateRules(rule, consentData, currentModule, gvlId, moduleType) {
   const purposeId = TCF2[Object.keys(TCF2).filter(purposeName => TCF2[purposeName].name === rule.purpose)[0]].id;
 
   // return 'true' if vendor present in 'vendorExceptions'
@@ -138,12 +141,14 @@ export function validateRules(rule, consentData, currentModule, gvlId) {
   const vendorConsent = deepAccess(consentData, `vendorData.vendor.consents.${gvlId}`);
   const liTransparency = deepAccess(consentData, `vendorData.purpose.legitimateInterests.${purposeId}`);
 
+  const vendorlessModule = includes(VENDORLESS_MODULE_TYPES, moduleType);
+
   /*
     Since vendor exceptions have already been handled, the purpose as a whole is allowed if it's not being enforced
     or the user has consented. Similar with vendors.
   */
   const purposeAllowed = rule.enforcePurpose === false || purposeConsent === true;
-  const vendorAllowed = rule.enforceVendor === false || vendorConsent === true;
+  const vendorAllowed = rule.enforceVendor === false || vendorConsent === true || vendorlessModule === true;
 
   /*
     Few if any vendors should be declaring Legitimate Interest for Device Access (Purpose 1), but some are claiming
@@ -162,8 +167,9 @@ export function validateRules(rule, consentData, currentModule, gvlId) {
  * @param {Function} fn reference to original function (used by hook logic)
  * @param {Number=} gvlid gvlid of the module
  * @param {string=} moduleName name of the module
+ * @param {string=} moduleType module type
  */
-export function deviceAccessHook(fn, gvlid, moduleName, result) {
+export function deviceAccessHook(fn, gvlid, moduleName, moduleType, result) {
   result = Object.assign({}, {
     hasEnforcementHook: true
   });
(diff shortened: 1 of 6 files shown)


modules/sharedIdSystem.js
```diff
@@ -11,7 +11,8 @@ import { coppaDataHandler } from '../src/adapterManager.js';
 import {getStorageManager} from '../src/storageManager.js';
 
 const GVLID = 887;
-export const storage = getStorageManager({gvlid: GVLID, moduleName: 'pubCommonId'});
+const MODULE_TYPE = 'fpid-module';
```
patmmccann, line 14: with this fix in place, let's go ahead and remove line 13, correct? Prebid was only a vendor on the deprecated 3rd party component of this.

modules/pubCommonId.js
```diff
@@ -9,7 +9,8 @@ import * as events from '../src/events.js';
 import CONSTANTS from '../src/constants.json';
 import { getStorageManager } from '../src/storageManager.js';
 
-const storage = getStorageManager();
+const MODULE_TYPE = 'fpid-module';
```
patmmccann, line 12: can you also add this to the publisherprovidedid module?

patmmccann (changes requested):
Thanks for your contribution! We really appreciate it.

modules/gdprEnforcement.js
```diff
@@ -183,7 +189,7 @@ export function deviceAccessHook(fn, gvlid, moduleName, result) {
           gvlid = getGvlid(moduleName) || gvlid;
         }
         const curModule = moduleName || curBidder;
-        let isAllowed = validateRules(purpose1Rule, consentData, curModule, gvlid);
+        let isAllowed = validateRules(purpose1Rule, consentData, curModule, gvlid, moduleType);
```
patmmccann, line 192: is it possible to check for purpose1 consent to the publisher or purpose 1 in general (instead of on a specific vendor) here, as per https://github.com/InteractiveAdvertisingBureau/GDPR-Transparency-and-Consent-Framework/blob/master/TCFv2/IAB%20Tech%20Lab%20-%20CMP%20API%20v2.md#tcdata

SKOCHERI (approved):

jdwieland8282 (approved):

jdwieland8282:
@goldun , looks like you have some test failures, can you get these resolved so we can merge?

modules/gdprEnforcement.js
dmdabbs: I was going to suggest that @patmmccann, but with a caveat. TCF CMPs may, but are not required to, provide transparency and consent for TCF Purposes _for the Publisher_ and then appropriately reflect those signals in the tcstring and API response data. So checking those signals makessense when the site has configured their CMP to disclose the publisher's processing.

modules/pubCommonId.js
patmmccann: https://github.com/goldun/Prebid.js/blob/sharedid_vendorless_consent/modules/pubProvidedIdSystem.js

modules/gdprEnforcement.js
goldun: Not sure I follow you guys. We have a `validatRules` function, where I've added additional logic to not check the vendor consent for vendorless module: https://github.com/prebid/Prebid.js/blob/afe88d862f031631058f343622e487af6bb4d900/modules/gdprEnforcement.js#L151
Do you, mean, that this exception should be done only if `purposeId === 1`?

patmmccann: I apologize for the confusion. 

Perhaps we can see best in an additional unit test. Sharedid module (after the vendor code is deleted) should be allowed to use storage manager if there is purpose 1 consent. and should not be allowed if there is not.

modules/pubCommonId.js
patmmccann: looking at it again, this module doesnt seem to use storage manager, never mind

modules/gdprEnforcement.js
goldun: gotcha will do

patmmccann: Thanks!

src/storageManager.js
```diff
@@ -303,7 +303,7 @@ export function newStorageManager({gvlid, moduleName, bidderCode, moduleType} =
 /**
  * This hook validates the storage enforcement if gdprEnforcement module is included
  */
-export const validateStorageEnforcement = hook('async', function(gvlid, moduleName, hookDetails, callback) {
+export const validateStorageEnforcement = hook('async', function(gvlid, moduleName, moduleType, hookDetails, callback) {
   callback(hookDetails);
```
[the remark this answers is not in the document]
goldun: thanks! I knew this will cause issues. checking...

goldun:
@patmmccann tests are fixed

dgirardi:
@goldun I'm looking at this in light of #8647, and I'm wondering: why isn't this simply changing sharedId to use [getCoreStorageManager](https://github.com/prebid/Prebid.js/blob/b5c9c068048269f6dcce0f1a8c3b79b50a452442/src/storageManager.js#L314) to bypass enforcement? is it to respect `setConfig({deviceAccess})`?

dgirardi:
nevermind - I realize now this needs to bypass all of gdpr, not just storage.

dgirardi:
I posted https://github.com/prebid/Prebid.js/pull/8648 to revert this change as https://github.com/prebid/Prebid.js/issues/8647 gets addressed.

My suggestion is, rather than changing the signature of `validateRules` (which requires a lot more cascading changes through all indirect callers), update its logic to check `moduleName` against a whitelist and bypass all validation for sharedId / pubCommonId.

Issue #5637, opened 2020-08-19
Overriding referrer: publisherDomain vs pageUrl


bretg:
## Description

We have two overlapping ways for publishers to override the results of [Prebid's attempt to derive the page's URL](https://github.com/prebid/Prebid.js/blob/master/src/refererDetection.js) that's stored in bidderRequest.refererInfo as described in https://docs.prebid.org/dev-docs/bidder-adaptor.html#referrers

Another use case here is that the publisher may want to suppress the full page URL from bidders.

I'd like to propose that we fix this in 6.0:

1) Allow the pub to define the pageUrl with setConfig
2) Prebid.js core reads the pageUrl value and overrides bidderRequest.refererInfo.canonicalUrl
3) Adapters cannot call getConfig or any other function (e.g. window.location), only retrieving the values from bidderRequest.refererInfo
4) The publisherDomain config goes away. If someone needs to have just the domain, we could add it to bidderRequest.refererInfo

Here are the two current `setConfig` options:

### publisherDomain

As documented at https://docs.prebid.org/dev-docs/publisher-api-reference.html#setConfig-Publisher-Domain

Set the publisher’s domain where Prebid is running, for cross-domain iframe communication:
```
pbjs.setConfig({ publisherDomain: "https://www.theverge.com" )
```

This is used by the avocet and readpeak adapters, but they appear to expect the value to be "example.com" rather than "https://example.com"

### pageUrl

There is a setConfig value used by 16 adapters in addition to the refererInfo object or as an override.

e.g. 
```
const page = utils.deepAccess(bidderRequest, 'refererInfo.canonicalUrl') || config.getConfig('pageUrl') || utils.deepAccess(window, 'location.href');

config.getConfig('pageUrl') || bidderRequest.refererInfo.referer
```

patmmccann:
Requested not to be used at https://github.com/prebid/Prebid.js/pull/5674/files#r480260677

gglas:
Current plan is to copy over pageURL for current 6 adapters, will also explore a utility function that copies one value to the other, also for docs :

1. deprecation of pubDomain
2. Bidder Doucmentation - domain needs to be retrieved from bidderRequest.refererInfo. Nowhere else.

jbartek25:
[...]

PR #8450, opened 2022-05-23
Prebid 7: refactoring of refererInfo, pageUrl, publisherDomain

dgirardi:
## Type of change
- [x] Refactoring ~~(no functional changes, no api changes)~~

## Description of change

This refactors `refererDetection` (seen by adapters as `bidderRequest.refererInfo`) to:

- rename `referer` to `topmostLocation`, keeping it for backwards compatibility.
- add `location`  - the value of `window.top.location`, or null when that is unavailable (blocked by cross-origin protections)
- add `ref` - the value of `window.top.document.referrer`, or null when that is unavailable
- add `page` - `canonicalUrl` falling back to `location`
- add `domain` - the domain portion of `page`

Additionally, modules that use `refererDetection` (which is nearly every one)  have been refactored with approximately this logic:

- if they were just looking at `refererInfo.referer`, they are now using `refererInfo.page` instead
- if they were clearly expecting `refererInfo.referer` to be the referrer (who would blame them), they are now using `refererInfo.ref` instead. Unfortunately, the original intent is rarely clear - this mostly applies to ortb2 payloads that were filling out both `page` and `ref`.
- if they were doing "advanced" stuff, such as looking at `refererInfo.stack`, `reachedTop` etc, then `refererInfo.referer` was replaced with `refererInfo.topmostLocation` to keep the same behavior.

Because of: the ambiguity of the original name choices; the sheer number of affected modules, and the impressive variation in how they deal with this task, I am sure to have broken some adapters. **Please check how your module is affected**; I have done my best to mark ambiguities with TODO comments.

## Other information

Resolves https://github.com/prebid/Prebid.js/issues/5637
Documentation PR: https://github.com/prebid/prebid.github.io/pull/3799


modules/tripleliftBidAdapter.js
```diff
@@ -27,7 +27,8 @@ export const tripleliftAdapterSpec = {
     tlCall = tryAppendQueryString(tlCall, 'v', '$prebid.version$');
 
     if (bidderRequest && bidderRequest.refererInfo) {
-      let referrer = bidderRequest.refererInfo.referer;
+      // TODO: is 'page' the right value here?
```
nllerandi3lift, line 30: Based on [this](https://github.com/prebid/Prebid.js/issues/5637#issuecomment-1116554262), yes 'page' is correct. We are interested in the actual page where the auction is happening, not the `document.referrer`.

patmmccann:
@dgirardi @bretg is the intention to also delete this line?

https://github.com/prebid/Prebid.js/blob/e47b9081ef3f6283053c46dd432a6d593d9b3558/src/config.js#L25

modules/ttdBidAdapter.js
```diff
@@ -125,14 +125,14 @@ function getUser(bidderRequest) {
...
-    page: utils.deepAccess(bidderRequest, 'refererInfo.referer'),
+    page: utils.deepAccess(bidderRequest, 'refererInfo.page'),
     publisher: {
       id: utils.deepAccess(bidderRequest, 'bids.0.params.publisherId'),
     },
     ...firstPartyData.site
   };
 
-  var publisherDomain = config.getConfig('publisherDomain');
+  var publisherDomain = bidderRequest.refererInfo.domain;
```
patmmccann, line 135: might be helpful to rename this var so it doesnt show up in searches

src/config.js
```diff
@@ -86,12 +85,12 @@ export function newConfig() {
...
       },
 
-      // domain where prebid is running for cross domain iframe communication
-      _publisherDomain: DEFAULT_PUBLISHER_DOMAIN,
+      _publisherDomain: null,
       get publisherDomain() {
         return this._publisherDomain;
       },
       set publisherDomain(val) {
+        logWarn('publisherDomain is deprecated and has no effect since v7 - use pageUrl instead')
```
patmmccann, line 93: what about https://github.com/prebid/Prebid.js/blob/e47b9081ef3f6283053c46dd432a6d593d9b3558/src/prebid.js#L951 and https://github.com/prebid/Prebid.js/blob/130f7265cf4baa470d62102993217c3dd69998b2/test/spec/config_spec.js#L154

modules/appnexusBidAdapter.js
```diff
@@ -265,12 +265,13 @@ export const spec = {
...
       let refererinfo = {
-        rd_ref: encodeURIComponent(bidderRequest.refererInfo.referer),
+        // TODO: are these the correct referer values?
+        rd_ref: encodeURIComponent(bidderRequest.refererInfo.topmostLocation),
         rd_top: bidderRequest.refererInfo.reachedTop,
         rd_ifs: bidderRequest.refererInfo.numIframes,
         rd_stk: bidderRequest.refererInfo.stack.map((url) => encodeURIComponent(url)).join(',')
       }
-      let pubPageUrl = config.getConfig('pageUrl');
+      let pubPageUrl = bidderRequest.refererInfo.canonicalUrl;
```
jsnellbaker, line 274: Our original use for `pageUrl` config option was for some custom AMP related scenarios where the normal refer page info was not available and when AMP caches it content; it would allow the publisher to ensure they were passing a value that made sense for the page.  I'd like to keep getConfig setup for now please.

dgirardi: `canonicalUrl` picks `getConfig('pageUrl')`, then falls back to `<link rel="canonical">`. https://github.com/dgirardi/Prebid.js/blob/2ff990bbb19b78ee619468fe25c042e633fec34e/src/refererDetection.js#L84-L87

modules/ttdBidAdapter.js
dgirardi: disagree, it's a game of whack-a-mole (there are many other instances of this name) and names should be chosen for clarity, not ease of grepping through unrelated code elsewhere.

src/config.js
dgirardi: Updated the comment; the test should stay IMO as get/set of publisherDomain is still a special case that's not covered by other tests.
It was my intention to remove the default value for it, yes, as it'd be odd to have it default to something when the goal is to remove its use.

modules/appnexusBidAdapter.js
jsnellbaker: Thanks for making the updates @dgirardi I think those would work out, so the changes here should be fine.

modules/criteoBidAdapter.js
```diff
@@ -224,9 +224,9 @@ function publisherTagAvailable() {
 function buildContext(bidRequests, bidderRequest) {
   let referrer = '';
   if (bidderRequest && bidderRequest.refererInfo) {
-    referrer = bidderRequest.refererInfo.referer;
+    referrer = bidderRequest.refererInfo.page;
   }
-  const queryString = parseUrl(referrer).search;
+  const queryString = parseUrl(bidderRequest?.refererInfo?.topmostLocation).search;
```
leonardlabat, line 229: I think it should be referrerInfo.page here too
What made you think it was better to use topmostlocation ?

leonardlabat: Nop indeed it makes sense, let's keep topmostlocation.

Thanks !

modules/teadsBidAdapter.js
```diff
@@ -126,8 +126,9 @@ export const spec = {
 
 function getReferrerInfo(bidderRequest) {
   let ref = '';
-  if (bidderRequest && bidderRequest.refererInfo && bidderRequest.refererInfo.referer) {
-    ref = bidderRequest.refererInfo.referer;
+  if (bidderRequest && bidderRequest.refererInfo && bidderRequest.refererInfo.page) {
+    // TODO: is 'page' the right value here?
```
github-baptiste-haudegand, line 130: We expect the url of the current page, so indeed this looks good, thanks for the changes

modules/minutemediaBidAdapter.js
```diff
@@ -423,8 +423,9 @@ function generateGeneralParams(generalObject, bidderRequest) {
   }
 
   if (bidderRequest && bidderRequest.refererInfo) {
-    generalParams.referrer = deepAccess(bidderRequest, 'refererInfo.referer');
-    generalParams.page_url = config.getConfig('pageUrl') || deepAccess(window, 'location.href');
+    // TODO: are these the right refererInfo values - does the fallback make sense?
+    generalParams.ref = deepAccess(bidderRequest, 'refererInfo.ref');
```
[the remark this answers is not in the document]
lasloche: amazing, thanks a lot for the update!

modules/rubiconBidAdapter.js
```diff
@@ -831,11 +831,12 @@ function _getScreenResolution() {
  * @returns {string}
  */
 function _getPageUrl(bidRequest, bidderRequest) {
-  let pageUrl = config.getConfig('pageUrl');
```
robertrmartinez, line 834: why is `config.getConfig('pageUrl')` removed? Is it deprecated?

robertrmartinez (approved):
Looks good from Rubicon adapter standpoint, thanks!

PR #8471, opened 2022-05-25
ZetaGlobalSsp BidAdapter: shortname to url

asurovenko-zeta:
## Type of change
- [x] Feature

`shortname` param into url param to split requests through balancer.

Please merge with: https://github.com/prebid/prebid.github.io/pull/3797


asurovenko-zeta:
@ChrisHuie It was fixed. Thanks!

PR #8472, opened 2022-05-25
Hadron analytics adapter: fix cross-origin exception on init

dgirardi:
## Type of change

- [x] Bugfix

## Description of change

Remove access to window.top, which is causing prebid to throw an exception on load when cross-frame

## Other information

Tested in the cross-origin scenario through locally remapped hosts resolution - this seems to be the only offender of this type in 6.26.0


jsnellbaker (approved):
LGTM

Tagging the original author @jlaso so they're aware of the change.

jlaso:
thank you  @dgirardi  and @jsnellbaker  for fixing this bug

PR #8473, opened 2022-05-25
Ftrack Id Module: remove configured url validation

ftxmoJason:
## Type of change
- [X] Bugfix

## Description of change
Removing a check of a url string that in effect made the url not configurable

- this change does not require a documentation change


patmmccann:
It ran your tests, only failure was on finteza, caused by datablocks and addressed in another pr

PR #8484, opened 2022-05-26
Data Controller Module: initial release

SKOCHERI:
## Type of change

- [ ] Bugfix
- [ ] Feature
- [ ] New bidder adapter   
- [ ] Build related changes
- [ ] CI related changes
- [ ] Other

## Description of change

```
{
  bidder: '<bidder name>',
  params: {
    // ...
  }
}
```


modules/dataControllerModule/index.md
```diff
@@ -0,0 +1,30 @@
...
+
+This module will filter EIDs and SDA based on the configurations.
+The filtered EIDs are stored in 'dcUsersAsEids' configuration and filtered SDA are updated in bidder configuration.
+
+Sub module object with the following keys:
+
+|  param name | type  | Scope | Description | Params |
+| :------------ | :------------ | :------ | :------ | :------ |
+|  filterEIDwhenSDA  | function | optional | Filters user EIDs based on SDA | bidrequest |
+|  filterSADwhenEID  | function | optional | Filters SDA based on configured EIDs | bidrequest  |
```
patmmccann, line 17: typo

modules/dataControllerModule/index.js
```diff
@@ -0,0 +1,90 @@
...
+  if (dataControllerConfig.filterEIDwhenSDA && dataControllerConfig.filterSADwhenEID) {
+    processBidderRequests.getHooks({hook: processBidderRequestHook}).remove();
+    return;
+  }
+  if (dataControllerConfig.filterEIDwhenSDA) {
+    let filterEIDs = submodule.filterEIDs(bidRequest)
+    if(filterEIDs) {
+      config.setConfig({'dcUsersAsEids': filterEIDs});
+    }
+  } else if (dataControllerConfig.filterSADwhenEID) {
```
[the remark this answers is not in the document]
SKOCHERI: @patmmccann  do we want to have the param as "supressSDAwhenEID"/"supressEIDwhenSDA"? Filter would be more generic rather than suppress.

patmmccann:
SAD is a typo for SDA

modules/dataControllerModule/index.js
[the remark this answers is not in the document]
patmmccann: All references to SAD should be changed to SDA

modules/dataControllerModule/index.js
```diff
@@ -0,0 +1,98 @@
...
+  submodules.forEach(submodule => {
+    filterData(submodule, bidderRequests)
+  });
+}
+
+export function filterData(submodule, bidRequest) {
+  dataControllerConfig = config.getConfig(MODULE_NAME);
+
+  if (!dataControllerConfig) {
+    processBidderRequests.getHooks({hook: processBidderRequestHook}).remove();
```
dgirardi, line 51: I'd prefer to keep configuration and business logic separate - IMO this will get confusing fast as complexity increases. (IMO it makes more sense to turn the module off from the same place you turn it on).

src/adapters/bidderFactory.js
```diff
@@ -296,6 +296,12 @@ export function newBidder(spec) {
  * @param onCompletion {function()} invoked once when all bid requests have been processed
  */
 export const processBidderRequests = hook('sync', function (spec, bids, bidderRequest, ajax, wrapCallback, {onRequest, onResponse, onError, onBid, onCompletion}) {
+  let updatedUserAsEids = config.getConfig('dcUsersAsEids');
```
dgirardi, line 299: This looks like it's using `config` as a global variable, plus it's adding dead code for those who are not using the data controller module.

Why not do this directly from the hook added by the module? 
Also, IMO the `makeBidRequests` hook is a better place for this - I don't know if it makes a difference now, but my instinct is that the earlier they have this data the better.

test/spec/modules/dataController_spec.js
```diff
@@ -0,0 +1,378 @@
+import {expect} from 'chai';
+import {config} from 'src/config.js';
+import {filterData} from 'modules/dataControllerModule/index.js';
+import * as utils from 'src/utils.js';
+
+const MODULE_NAME = 'dataController';
+let dataControllerConfig;
+const ALL = '*';
+
+function filterEIDs(requestObject) {
```
dgirardi, line 10: This is still a work in progress I assume - this logic looks like it belongs in the module.

modules/dataControllerModule/index.js
```diff
@@ -0,0 +1,98 @@
...
+ * @name dataControllerModule#filterSDA
+ * @param {Object} bidRequest
+ */
+
+import {config} from '../../src/config.js';
+import {module, getHook} from '../../src/hook.js';
+import {processBidderRequests} from '../../src/adapters/bidderFactory.js';
+import {logError} from '../../src/utils.js';
+
+let submodules = [];
```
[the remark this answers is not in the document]
SKOCHERI: @dgirardi The proposal mentions about Publisher having ability to define the rules to filter EID and SDA hence providing submodule.

test/spec/modules/dataController_spec.js
SKOCHERI: This module development is completed. The test contains how the submodule looks like

dgirardi: the test is testing test code? that can't be right. I'm missing the point entirely, once the bundle is built this code (and therefore the tests on it) are completely irrelevant no? the module in the actual codebase is an empty shell.

modules/dataControllerModule/index.js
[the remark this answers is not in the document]
SKOCHERI: Will take a look

src/adapters/bidderFactory.js
SKOCHERI: @dgirardi  I have update this module to be self contained.

modules/dataControllerModule/index.js
```diff
@@ -0,0 +1,149 @@
...
+  let resetSDA = containsConfiguredEIDS(eidSources);
+  const allBidderConfigs = config.getBidderConfig();
+  if (resetSDA) {
+    let bidderConfig = allBidderConfigs[bidderCode];
+    bidderConfig.ortb2.user.data = [];
+    config.setBidderConfig(allBidderConfigs, false);
+  }
+}
+
+export function initHook() {
```
dgirardi, line 144: I think the better pattern is `config.getConfig(MODULE_NAME, (cfg) => {if (isValid(cfg) enable() else disable())})` - it's the approach taken by most (if not all) other modules:

 - it keeps better separation between setup logic and business logic
 - it doesn't use cpu cycles during the auction in the case when the module has not been enabled
 - it needs to worry about fewer edge cases, like what happens if configuration is set during (or after) the auction

modules/dataControllerModule/index.js
```diff
@@ -0,0 +1,149 @@
...
+const _logInfo = createLogInfo(LOG_PRE_FIX);
+
+function createLogInfo(prefix) {
+  return function (...strings) {
+    logInfo(prefix + ' ', ...strings);
+  }
+}
+
+function hasValidConfiguration() {
+  dataControllerConfig = config.getConfig(MODULE_NAME);
```
dgirardi, line 24: may I suggest something like `setupConfig` instead of `hasValidConfiguration`? to make it clear that the function has side effects.

dgirardi (changes requested):
Thanks for the update. I think the approach would not work correctly (sorry if I missed it in the previous review) - for two reasons:

 1. `processBidderRequest` does not run for Prebid Server bids - only client adapters;
 2. since prebid 7, `setBidderConfig` (or `setConfig`) is not the right way to update first party data - it will not take effect until the next auction (see [this](https://docs.prebid.org/dev-docs/add-rtd-submodule.html#getbidrequestdata))

I think the best way to do this is to hook on [`startAuction`](https://github.com/prebid/Prebid.js/blob/06e2976190aac71836f2fd0605598de38f4dc42d/src/prebid.js#L589) and:

 - perform `filterSDA` by modifying the `ortb2Fragments` parameter;
 - perform `filterEIDs` by modifying the `adUnits[].bids[].userIdAsEid` parameter - in the same way that property is set by the [userId module](https://github.com/prebid/Prebid.js/blob/06e2976190aac71836f2fd0605598de38f4dc42d/modules/userId/index.js#L493-L502)

modules/dataControllerModule/index.js
```diff
@@ -0,0 +1,149 @@
...
+import {getHook, module} from '../../src/hook.js';
+import {deepAccess, logInfo} from '../../src/utils.js';
+import {processBidderRequests} from '../../src/adapters/bidderFactory.js';
+
+const LOG_PRE_FIX = 'Data_Controller : ';
+const ALL = '*';
+const MODULE_NAME = 'dataController';
+let dataControllerConfig;
+
+const _logInfo = createLogInfo(LOG_PRE_FIX);
```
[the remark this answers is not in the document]
SKOCHERI: @dgirardi Have updated the module to be called before startAuction and fixed other issues pointed out.

modules/dataControllerModule/index.js
```diff
@@ -0,0 +1,162 @@
...
+      containsSDA = true;
+    }
+  });
+  return containsSDA;
+}
+
+export function getSegmentConfig(ortb2Fragments) {
+  let bidderSDAMap = new Map();
+
+  for (const [key, value] of Object.entries(ortb2Fragments.bidder)) {
```
dgirardi, line 71: Same here - should this look at global config?

SKOCHERI:
> Reading through https://docs.google.com/document/d/1g9GxuMtPwZyk5lRKCyoWVaYJGpqUGEbFNwEBMQswcjU/edit# again, I think the intent is - from the point of view of the publisher, "I don't want this data to be visible to the adapters / SSP".
> 
> So I think this module should be overzealous, and:
> 
> * inspect / remove global configuration as well (`ortb2Fragments.global`)
> * also remove data from `user.ext.eids` _and_ `bid.userId`, not just `bid.userIdAsEid`. otherwise, there might be a situation where for example an RTD module is populating the first, or an adapter is looking at the second (instead of `userIdAsEid`), and they still get access to the "forbidden" data.

Updating to consider the above

dgirardi:
they could be in either `bidder` or `global`, and both should be checked / filtered, like the other pieces of data. The logic is roughly speaking, when the auction starts, [`global` is from `getConfig`, `bidder` is from `getBidderConfig`](https://github.com/prebid/Prebid.js/blob/a29cafd7d0c4a50b3d8529065c92ef1b45f82cfa/src/prebid.js#L582-L585); then the object passes through [FPD](https://github.com/prebid/Prebid.js/blob/a29cafd7d0c4a50b3d8529065c92ef1b45f82cfa/modules/fpdModule/index.js#L36) and [RTD](https://github.com/prebid/Prebid.js/blob/a29cafd7d0c4a50b3d8529065c92ef1b45f82cfa/modules/rtdModule/index.js#L231) modules, then it gets here in dataController (which should be last). Any of the above may have added anything in them, so this should check `global.user` and each entry in `bidder.[code].user`. At the end each adapter sees only one object [`ortb2 = mergeDeep({},  ortb2Fragments.global, ortb2Fragments.bidder.[code])`](https://github.com/prebid/Prebid.js/blob/a29cafd7d0c4a50b3d8529065c92ef1b45f82cfa/src/adapterManager.js#L232).

SKOCHERI:
> they could be in either `bidder` or `global`, and both should be checked / filtered, like the other pieces of data. The logic is roughly speaking, when the auction starts, [`global` is from `getConfig`, `bidder` is from `getBidderConfig`](https://github.com/prebid/Prebid.js/blob/a29cafd7d0c4a50b3d8529065c92ef1b45f82cfa/src/prebid.js#L582-L585); then the object passes through [FPD](https://github.com/prebid/Prebid.js/blob/a29cafd7d0c4a50b3d8529065c92ef1b45f82cfa/modules/fpdModule/index.js#L36) and [RTD](https://github.com/prebid/Prebid.js/blob/a29cafd7d0c4a50b3d8529065c92ef1b45f82cfa/modules/rtdModule/index.js#L231) modules, then it gets here in dataController (which should be last). Any of the above may have added anything in them, so this should check `global.user` and each entry in `bidder.[code].user`. At the end each adapter sees only one object [`ortb2 = mergeDeep({}, ortb2Fragments.global, ortb2Fragments.bidder.[code])`](https://github.com/prebid/Prebid.js/blob/a29cafd7d0c4a50b3d8529065c92ef1b45f82cfa/src/adapterManager.js#L232).

@dgirardi Updated to consider global and bidder segments and update eids as mentioned above

modules/dataControllerModule/index.js
```diff
@@ -0,0 +1,195 @@
...
+  _dataControllerConfig.filterEIDwhenSDA.forEach(segment => {
+    if (bidderSegement.has(segment)) {
+      containsSDA = true;
+    }
+  });
+  return containsSDA;
+}
+
+function getSegmentConfig(ortb2Fragments) {
+  let bidderSDAMap = new Map();
```
dgirardi, line 71: there's some duplication in here, which you could avoid by reusing the same logic for both global and bidder data. Here's a possible refactor to illustrate what I mean:

```
let bidderSDAMap = new Map();
function collectSegments(key, data) {
   let segmentSet = constructSegment(deepAccess(data, 'user.data') || []);
   if (segmentSet && segmentSet.size > 0) bidderSDAMap.set(key, segmentSet);
}
collectSegments(GLOBAL, ortb2Fragments.global); 
Object.entries(ortb2Fragments.bidder).forEach(([bidder, data]) => collectSegments(bidder, data));
return bidderSDAMap;
```

dgirardi (approved):
LGTM - I think this conforms to the proposal now as I understand it. I have a few more minor suggestions, I think the first (changing from `GLOBAL = 'global'` to `GLOBAL = {}`) should be done for correctness, but the rest is not that important.

modules/dataControllerModule/index.js
SKOCHERI: Fixed

modules/dataControllerModule/index.md
```diff
@@ -0,0 +1,30 @@
+# Overview
+
+```
+Module Name: Data Controller Module
+```
+
+# Description
+
+This module will filter EIDs and SDA based on the configurations.
+The filtered EIDs are stored in 'dcUsersAsEids' configuration and filtered SDA are updated in bidder configuration.
```
mmoschovas, line 10: Is this detail no longer relevant? I am not seeing any of this stored any longer, so if that is the case, can this be removed

modules/dataControllerModule/index.js
```diff
@@ -0,0 +1,194 @@
...
+      return;
+    }
+
+    if (dataControllerConfig.dataController.filterEIDwhenSDA && dataControllerConfig.dataController.filterSDAwhenEID) {
+      _logger.logInfo(`Data Controller can be configured with either filterEIDwhenSDA or filterSDAwhenEID`);
+      startAuction.getHooks({hook: filterBidData}).remove();
+      return;
+    }
+    confListener(); // unsubscribe config listener
+    _dataControllerConfig = dataControllerConfig.dataController;
```
mmoschovas, line 187: I think we can clean this up a little to store the dataController in a const then reference that instead or referencing the full path each time. 

i.e.
```
const dataController = dataControllerConfig && dataControllerConfig.dataController;

if (!dataController) {
    ....
}

if (dataController.filterEIDwhenSDA && dataController. filterSDAwhenEID) {
    ....
}

confListener(); // unsubscribe config listener
_dataControllerConfig = dataController;

getHook('startAuction').before(filterBidData);
```

mmoschovas (approved):
Thanks for the update! lgtm

mikemangas:
@SKOCHERI 
@dgirardi
@mmoschovas

Thanks for your work. I would like to ask you some questions.
- I am not sure, how can i implement and test this module. What are the Requirements for the Data Controller Module?
- Are Seller Defined Audiences (SDA) required, in order to test/use this module? If yes, Is there a way to add SDA (provider name, segment id, taxonomy name) to my bid requests? if yes, is there a documentation for that?
- What do i pass into "filterEIDwhenSDA" or "filterSDAwhenEID"? What are the requirements for using the one or the other? 

Thank you for any helpful information.

mikemangas:
Does a bidAdapter have to change the code, in order to support this module?

PR #8491, opened 2022-05-30
Criteo - Update Publisher Tag version referenced by prebid adapter

leonardlabat:
## Type of change

## Description of change
Upgrade Criteo prebid adapter to reference the latest Criteo PubTag version.

### DIFF
--- a/modules/criteoBidAdapter.js
+++ b/modules/criteoBidAdapter.js
@@ -24,7 +24,7 @@ const LOG_PREFIX = 'Criteo: ';
   Unminified source code can be found in the privately shared repo: https://github.com/Prebid-org/prebid-js-external-js-criteo/blob/master/dist/prod.js
 */
 const FAST_BID_VERSION_PLACEHOLDER = '%FAST_BID_VERSION%';
-export const FAST_BID_VERSION_CURRENT = 117;
+export const FAST_BID_VERSION_CURRENT = 123;
 const FAST_BID_VERSION_LATEST = 'latest';
 const FAST_BID_VERSION_NONE = 'none';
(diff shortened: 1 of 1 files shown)


patmmccann:
Not yet available at the referenced repo

modules/criteoBidAdapter.js
```diff
@@ -24,7 +24,7 @@ const LOG_PREFIX = 'Criteo: ';
   Unminified source code can be found in the privately shared repo: https://github.com/Prebid-org/prebid-js-external-js-criteo/blob/master/dist/prod.js
 */
 const FAST_BID_VERSION_PLACEHOLDER = '%FAST_BID_VERSION%';
-export const FAST_BID_VERSION_CURRENT = 117;
+export const FAST_BID_VERSION_CURRENT = 123;
```
patmmccann, line 27: The repo in the line above stops at 116 currently

patmmccann (changes requested):

modules/criteoBidAdapter.js
leonardlabat: Indeed it looks like the updates were not uploaded there
We've fixed it, thanks for reporting it Patrick

patmmccann:
I read the pubtag js just now

A couple question:

Could you disclose the identity functionality in https://github.com/prebid/prebid.github.io/blob/master/dev-docs/bidders/criteo.md ? It seems like most of the functionalities are well disclosed in prebid issues but not in the primary documentation page. 

In that same file it mentions pubtag is required for floors. It seems like an oversight that hasn't been moved to your adapter yet. Could you "promote" that functionality into the main adapter soon?

patmmccann:
When you add the link to the docs pull, happy to merge

PR #8498, opened 2022-05-31
Adf adapter: avoid preflight request

braizhas:
## Type of change

- [x] Other

## Description of change
Content-type `application/json` causes preflight requests, this adds additional latency.

- contact email of the adapter’s maintainer Scope.FL.Scripts@adform.com

### DIFF
--- a/modules/adfBidAdapter.js
+++ b/modules/adfBidAdapter.js
@@ -232,9 +232,6 @@ export const spec = {
       method: 'POST',
       url: 'https://' + adxDomain + '/adx/openrtb',
       data: JSON.stringify(request),
-      options: {
-        contentType: 'application/json'
-      },
       bids: validBidRequests
     };
   },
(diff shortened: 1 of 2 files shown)


patmmccann:
This is fascinating, could you point to any documentation of this?

modules/adfBidAdapter.js
```diff
@@ -232,9 +232,6 @@ export const spec = {
       method: 'POST',
       url: 'https://' + adxDomain + '/adx/openrtb',
       data: JSON.stringify(request),
-      options: {
-        contentType: 'application/json'
```
patmmccann, line 236: why not 'text/plain' ? This is not a change request, I'm hoping to learn about this.

braizhas: 'text/plain' is the default value added by prebid core. https://github.com/prebid/Prebid.js/blob/c84cfce617b7b5128d4f34fc53087854c5b41c4b/src/adapters/bidderFactory.js#L380-L384

braizhas:
> This is fascinating, could you point to any documentation of this?

Noticed this while testing, had an issue that Adform bidder timeouts a lot, and was looking for possible solutions.

BTW, found this mentioned on MDN https://developer.mozilla.org/en-US/docs/Web/HTTP/CORS
> The only type/subtype combinations allowed for the media type specified in the Content-Type header are:
> application/x-www-form-urlencoded
> multipart/form-data
> text/plain

patmmccann (approved):

PR #8502, opened 2022-05-31
Magnite GVL ID update (7.0 branch)

bretg:
## Type of change
- [X] Other

## Description of change
At this point, the various Magnite exchanges are not in fact able to use the same GVL ID. 

## Other information
See the [IAB GVL vendors](https://vendor-list.consensu.org/v2/vendor-list.json) 52, 102, 165, and 202.


bretg:
docs PR https://github.com/prebid/prebid.github.io/pull/3810

PR #8504, opened 2022-06-01
Yieldlab Bid Adapter: support for ad unit sizes

nkloeber:
## Type of change
- [x] Feature

## Description of change
yieldlabBidAdapter.js add support for ad unit sizes

### DIFF
--- a/modules/yieldlabBidAdapter.js
+++ b/modules/yieldlabBidAdapter.js
@@ -10,12 +10,17 @@ const BID_RESPONSE_TTL_SEC = 300
 const CURRENCY_CODE = 'EUR'
 const OUTSTREAMPLAYER_URL = 'https://ad.adition.com/dynamic.ad?a=o193092&ma_loadEvent=ma-start-event'
 const GVLID = 70
+const DIMENSION_SIGN = 'x'
 
 export const spec = {
   code: BIDDER_CODE,
   gvlid: GVLID,
   supportedMediaTypes: [VIDEO, BANNER, NATIVE],
 
+  /**
+   * @param {object} bid
+   * @returns {boolean}
+   */
   isBidRequestValid: function (bid) {
     if (bid && bid.params && bid.params.adslotId && bid.params.supplyId) {
       return true
@@ -25,11 +30,13 @@ export const spec = {
 
   /**
    * This method should build correct URL
-   * @param validBidRequests
-   * @returns {{method: string, url: string}}
+   * @param {BidRequest[]} validBidRequests
+   * @param [bidderRequest]
+   * @returns {ServerRequest|ServerRequest[]}
    */
   buildRequests: function (validBidRequests, bidderRequest) {
     const adslotIds = []
+    const adslotSizes = [];
     const timestamp = Date.now()
     const query = {
       ts: timestamp,
@@ -38,6 +45,10 @@ export const spec = {
 
     _each(validBidRequests, function (bid) {
       adslotIds.push(bid.params.adslotId)
+      const sizes = extractSizes(bid)
+      if (sizes.length > 0) {
+        adslotSizes.push(bid.params.adslotId + ':' + sizes.join('|'))
+      }
       if (bid.params.targeting) {
         query.t = createTargetingString(bid.params.targeting)
       }
@@ -45,7 +56,7 @@ export const spec = {
         query.ids = createUserIdString(bid.userIdAsEids)
       }
       if (bid.params.customParams && isPlainObject(bid.params.customParams)) {
-        for (let prop in bid.params.customParams) {
+        for (const prop in bid.params.customParams) {
           query[prop] = bid.params.customParams[prop]
         }
       }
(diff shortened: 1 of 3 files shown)


modules/yieldlabBidAdapter.js
```diff
@@ -39,6 +41,15 @@ export const spec = {
 
     _each(validBidRequests, function (bid) {
       adslotIds.push(bid.params.adslotId)
+      let sizes = new Set()
```
brushmate, line 44: I think mutating function arguments is not a good idea when the same logic could be expressed as a pure function. I would prefer something like this:

```
const sizes = extractSizes(bid)

if (sizes.size > 0) {
    adslotSizes.push(bid.params.adslotId + ':' + [...sizes].join('|'))
}
```

brushmate (changes requested):

modules/yieldlabBidAdapter.js
```diff
@@ -11,6 +11,7 @@ const BID_RESPONSE_TTL_SEC = 300
 const CURRENCY_CODE = 'EUR'
 const OUTSTREAMPLAYER_URL = 'https://ad.adition.com/dynamic.ad?a=o193092&ma_loadEvent=ma-start-event'
 const GVLID = 70
+const DIMENSION_SIGN = 'x'
```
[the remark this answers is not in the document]
nkloeber: We are using the symbol as separator but it describes a multiplication and is normally called multiplication sign, also known as the times sign or the dimension sign. Since we are using it in context of a size I think dimension sign is the most fitting description.

brushmate (approved):

modules/yieldlabBidAdapter.js
```diff
@@ -382,5 +391,33 @@ function outstreamRender(bid) {
     window.document.dispatchEvent(new Event('ma-start-event'))
   });
 }
+/**
+ * Extract sizes from given bid for bid.mediaTypes or bid.sizes
+ * @param {Object} bid
+ */
+function extractSizes(bid) {
+  let sizes = new Set()
+  if (bid.mediaTypes) {
```
kippsterr, line 400: I know @nkloeber you didn't write this code in the first place but I am a bit undecided about this whole part.

1. According to the documentation about configuring for the [Yieldlab adapter](https://docs.prebid.org/dev-docs/bidders/yieldlab.html) there is no hint that `bid.sizes` is even an option. We might want to check back with Solutions to clarify where this is coming from originally (is there a global bid request object configuration page somewhere?) but I think currently the way to go is by using `bid.mediaTypes`.
2. When going down the media types branch I started wondering which types actually are multi size enabled. Turns out `VIDEO` and `native` do have some size settings but `BANNER` is the only type where different sizes are valid.

So my conclusion is the following:
1. Check with Solutions if we still need to support `bid.sizes` (hint: the adapter spec file needs to be updated as well). If not, just access `mediaTypes` directly for any following processing.
2. Check if we need to provide/extract sizes for the `NATIVE` case. If not, handle any further processing in a case style manner. If `VIDEO` do this otherwise if `BANNER` do that. I would not dynamically generate the field name (as in line 402 currently). That doesn't bring any value, it only makes it harder to read.
3. It might be easier to follow if we extract a list of all potential size value pairs first and then do the `DIMENSION_SIGN` joining once at the end only. We could also use a list first, apply the joining and to string conversion and then instantiate a set out of that list to eliminate potential duplicates.
4. Cover more cases of those size combinations in tests.
5. By the way, do we need to handle multiple sizes for `VIDEO` as well? If not, the situation would be even easier.

I wanted to provide you a brief code example of how I imagined the implementation but I am currently having issues with the tests locally. They seem to be executed more often than expected to me.

kippsterr: My current idea of this method looks like this:

```javascript
  const { mediaTypes } = bid; // see https://docs.prebid.org/dev-docs/adunit-reference.html#examples
  const sizes = [];

  if (isPlainObject(mediaTypes)) {
    Object.entries(mediaTypes)
      .forEach(([mediaType, mediaTypeConfig]) => {
        if (mediaType === VIDEO && isArray(mediaTypeConfig.playerSize)) {
          sizes.push(mediaTypeConfig.playerSize.flat());
        } else if (mediaType === BANNER && isArray(mediaTypeConfig.sizes)) {
          sizes.push(mediaTypeConfig.sizes.flat())
        }
      });
  } else if (isArray(bid.sizes)) {
    sizes.push(bid.sizes.flat())
  }

  return new Set(sizes.map(([width, height]) => width + DIMENSION_SIGN + height))
```

We still have to check back with Solutions and I wanna know why tests are run multiple times even though I am using `it.only('passes correct size to bid request for mediaType video with playerSize attribute', ...)`.

kippsterr (approved):

kippsterr:
@ChrisHuie please re-trigger the build. There is a test failure that is independent from our adapter. Review has been passed successfully now from our side, so just the Prebid approval is missing.

Thank you.

patmmccann (approved):

PR #8510, opened 2022-06-01
Kargo Analytics Adapter: Bid response time logging

jsadwith:
## Type of change
- [x] Feature

## Description of change
Adds support for bid response time logging for Kargo bids

## Other information
Related: https://github.com/prebid/prebid.github.io/pull/3818

### DIFF
--- a/modules/kargoAnalyticsAdapter.js
+++ b/modules/kargoAnalyticsAdapter.js
@@ -17,7 +17,7 @@ let _logBidResponseData = {
   responseTime: 0,
 };
 
-let _bidResponseDataSent = false;
+let _bidResponseDataLogged = [];
 
 var kargoAnalyticsAdapter = Object.assign(
   adapter({ analyticsType }), {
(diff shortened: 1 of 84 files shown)


jsadwith:
@robertrmartinez do you need anything from our end to get started on the review?

jsadwith:
@robertrmartinez thanks for pointing that out. I've updated the code to reset the bool on auction end. We don't want to send multiple events for the same auction, but this should reset it for any additional auctions on the page.

modules/kargoAnalyticsAdapter.js
```diff
@@ -1,14 +1,98 @@
...
+  }
+);
+
+function handleBidResponseData (bidResponse) {
+  if (bidResponse.bidder !== KARGO_BIDDER_CODE || _bidResponseDataSent) {
+    return
+  }
+
+  _logBidResponseData.auctionId = bidResponse.auctionId;
+  _logBidResponseData.responseTime = bidResponse.responseTimestamp - bidResponse.requestTimestamp;
```
robertrmartinez, line 50: I think prebid core sets a param on this object called `timeToRespond` which is essentially this.

Also, you want to only track when your exchange does a VALID bid response? Not when it no-bids or errors?

jsadwith:
Great idea on using a map to store the auction IDs - I decided just to use an array of auction IDs to keep it more simple. Is that okay or would you prefer the map?

modules/kargoAnalyticsAdapter.js
```diff
@@ -1,14 +1,96 @@
...
+        }
+      }
+    }
+  }
+);
+
+// handleBidResponseData: Get auction data for Kargo bids and send to server
+function handleBidResponseData (bidResponse) {
+  // Verify this is Kargo and we haven't seen this auction data yet
+  if (bidResponse.bidder !== KARGO_BIDDER_CODE || _bidResponseDataLogged.includes(bidResponse.auctionId) !== false) {
```
robertrmartinez, line 42: Two things:

1. What if somebody is using an alias name for Kargo, do you still want to track it or not? If so, you can look at the aliasRegistry in order to determine if a bidder is an alias of kargo or not.


2. This array of auctionId's definitely works. I only suggested a MAP because you can do something like:

```js
const _bidResponseDataLogged = {
  auctionId1: true,
  auctionId2: false
}
```
If NOT (data has sent AKA map at auctionId is true) 
```js
if (!_bidResponseDataLogged[bidResponse.auctionId]) {
   // log data
}
```

Which is a O(1) lookup.

But honestly, includes is prob fine as the array will not get too big for a given page load for most integrations!

robertrmartinez (approved):
Sorry for such a long delay. I had an extended holiday.

It is fine as is, just left 2 more minor comments, feel free to use them or just let me know if you are fine with how it is!

jsadwith:
Ahh - looks like I didn't push the last change using `timeToRespond` - thanks for pointing that out. All set now.

modules/kargoAnalyticsAdapter.js
```diff
@@ -1,14 +1,96 @@
...
+    return
+  }
+
+  _logBidResponseData.auctionId = bidResponse.auctionId;
+  _logBidResponseData.responseTime = bidResponse.timeToRespond;
```
robertrmartinez, line 47: so having this object at module scope we could possibly see issues if two auctions are running at same time.

The right thing would prob be to just create a local object based off your wanted params here.

Even the `auctionTimeout` part may change between auctions.

Now it is probably fine majority of the time. So I can be fine with leaving it if you want.

Let me know.

PR #8515, opened 2022-06-02
Vidazoo bid adapter: fix spurious test failure

dgirardi:
## Type of change

- [x] Bugfix

## Description of change

This fixes a test that would fail when it happened to run 15 seconds or more after its `describe` block:

https://app.circleci.com/pipelines/github/prebid/Prebid.js/11680/workflows/59aa141d-5c2b-487f-b965-3bff6029b3c6/jobs/22515

### DIFF
--- a/test/spec/modules/vidazooBidAdapter_spec.js
+++ b/test/spec/modules/vidazooBidAdapter_spec.js
@@ -322,7 +322,9 @@ describe('VidazooBidAdapter', function () {
   describe('unique deal id', function () {
     const key = 'myKey';
     let uniqueDealId;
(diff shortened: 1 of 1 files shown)


dgirardi:
@uditalias FYI - I also want to confirm that 15 seconds is what you expect for deal expiration? (maybe it's just me but I think it's odd to code that in microseconds).

dgirardi:
I am unable to spot the culprit, but I'll re-open this for review as it's still an improvement IMO - we'll have to wait and see if the failure pops up again.

patmmccann (approved):

PR #8524, opened 2022-06-06
PBjs Core Consent management: use last consent provided by CMP on timeout

dgirardi:
## Type of change

- [x] Bugfix

## Description of change

This changes the timeout behavior of the consentManagement module (for TCF v2 only).

If after `gdpr.timeout` the user has not provided consent (and therefore the CMP did not emit a `tcloaded` or `useractioncomplete` event), use the last consent data provided by any other CMP event instead of null.

According to [the TCF spec](https://github.com/InteractiveAdvertisingBureau/GDPR-Transparency-and-Consent-Framework/blob/master/TCFv2/IAB%20Tech%20Lab%20-%20CMP%20API%20v2.md#addeventlistener), CMPs should always emit at least one event regardless of what the user does.

> Note: The addEventListener callback shall be immediately called upon registration with the current TC data, even if the CMP status is loading and the CMP has incomplete TC Data, so that the calling script may have access to its registered listenerId. Furthermore, on every TC String change the callback shall be called unless it is removed via removeEventListener.

In my testing some implementations provide an empty consent string when the user does not dismiss the UI - in that case we ignore the consent object.

## Other information

Fixes https://github.com/prebid/Prebid.js/issues/7156
Same changes in 6.29.x: https://github.com/prebid/Prebid.js/pull/8540


patmmccann:
@dgirardi please resolve conflict here when you get a chance

ChrisHuie (approved):

Issue #6377, opened 2021-03-02
Proposal to use CMP in e2e tests


giniyat202:
## Type of issue
Proposal

## Description
We want to improve the consent management testing by introducing e2e tests that make use of actual CMP code, as the currently available tests only mock `__tcfapi` calls with predefined responses. The new tests could help avoid bugs such as https://github.com/prebid/Prebid.js/issues/5792.

We want to discuss the possibility of implementing these tests and the implementation details.

## Sample code
The code below provides a way to initialize iubenda CMP in the test environment with a predefined scenario:
```js
function initializeCMP(callbacks) {
  window._iub = window._iub || {};
  _iub.csConfiguration = {
    cookiePolicyId: 417383,
    siteId: 1,
    logLevel: 'error',
    lang: 'en',
    enableTcf: true,
  };

  if (callbacks) {
    _iub.csConfiguration.callback = callbacks
  }

  var stubScript = document.createElement('script');
  stubScript.src = '//cdn.iubenda.com/cs/tcf/beta/stub.js';
  document.body.insertBefore(stubScript, document.body.firstChild);

  var cmpScript = document.createElement('script');
  cmpScript.src = '//cdn.iubenda.com/cs/beta/iubenda_cs.js';
  cmpScript.setAttributeNode(document.createAttribute('async'));
  document.body.insertBefore(cmpScript, stubScript.nextSibling);
}

// Note: more scenarios can be produced by following this link: https://iabtcf.com/#/encode
const scenarios = {
  'everything rejected': 'CPAQjlJPAQlAKB7D3CENBJCgAAAAAAAAAAAAHPQAAOfgAAAA',
[...]

PR #8525, opened 2022-06-06
Consent management: update GDPR example

dgirardi:
## Type of change

- [x] Other

## Description of change

- update the GDPR example to use a modern CMP
- remove the "indefinite timeout" GDPR example. I am not sure about the history but I don't think it's a great idea to encourage users to do their own GPDR validation. (that example just waits for any consent data to show up, and disregards its contents).

@giniyat202 I have taken the liberty of using the CMP setup from #6377 - reasoning that if it's OK to use it in e2e tests, it may be OK to use it in examples. Please confirm that this is appropriate.


giniyat202:
@dgirardi It's totally OK, the example looks nice too. Thanks!

patmmccann (approved):

PR #8534, opened 2022-06-07
Triplelift Bid Adapter: get opeCloud data from local storage

nllerandi3lift:
## Type of change

- [X] Feature

## Description of change
Allows for our bid adapter to pick up the opeCloud localStorage (a Triplelift technology) and add to the bid request. This localStorage is set independently of Prebid. 

Includes 3 tests: 
- should send 1PlusX data as fpd if localStorage is available and no other fpd is defined
- should append 1PlusX data to existing user.data entries if localStorage is available
- should not append anything if getDataFromLocalStorage returns null


patmmccann:
Please grant your bidder access to storage manager before writing a test relying on access

patmmccann:
Feel free to reopen and change base to legacy

PR #8536, opened 2022-06-07
SpotX Bid Adapter: Append child fix

robertrmartinez:
## Type of change

## Description of change
Addresses https://github.com/prebid/Prebid.js/issues/8421


modules/spotxBidAdapter.js
```diff
@@ -420,12 +421,9 @@ export const spec = {
   }
 }
 
-function createOutstreamScript(bid) {
+function createOutstreamScript(bid, script) {
   const slot = getBidIdParameter('slot', bid.renderer.config.outstream_options);
   logMessage('[SPOTX][renderer] Handle SpotX outstream renderer');
-  const script = window.document.createElement('script');
-  script.type = 'text/javascript';
-  script.src = 'https://js.spotx.tv/easi/v1/' + bid.channel_id + '.js';
```
robertrmartinez, line 428: Instead of creating script in here, it is done by `loadExternalScript`

So it is passed into here after being inserted.

modules/spotxBidAdapter.js
```diff
@@ -501,15 +497,11 @@ function outstreamRender(bid) {
...
-        framedoc.body.appendChild(script);
+        script = loadExternalScript(`https://js.spotx.tv/easi/v1/${bid.channel_id}.js`, 'outstream', undefined, framedoc);
       } else {
-        const slot = getBidIdParameter('slot', bid.renderer.config.outstream_options);
-        if (slot && window.document.getElementById(slot)) {
-          window.document.getElementById(slot).appendChild(script);
-        } else {
-          window.document.getElementsByTagName('head')[0].appendChild(script);
-        }
+        script = loadExternalScript(`https://js.spotx.tv/easi/v1/${bid.channel_id}.js`, 'outstream');
```
robertrmartinez, line 502: `loadExternalScript` does not support this use case of appending under the slot element.

I do not know the use case enough to know if it matters so asking for clarification here.

modules/spotxBidAdapter.js
```diff
@@ -501,15 +497,11 @@ function outstreamRender(bid) {
...
       }
+      createOutstreamScript(bid, script);
```
robertrmartinez, line 504: This function is now run AFTER prebid fires the `loadExternalScript`

This mostly is just to set data attributes looks like, so I _think_ it should be fine to pass in the already created script and attach.

patmmccann: Are you able to get an internal review? This change seems slightly risky so just want to make sure you guys are totally comfortable

robertrmartinez: Yep! Working on getting someone more familiar with this use case to look at it! Hopefully by tomorrow morning.

patmmccann: Cool, we're releasing 7 today, so you might want to base this on master

patmmccann:
Argh! A test failure! I thought we got them all. At least we can be sure if the culprit now

bretg:
Adding @npeceniak for review.

robertrmartinez:
@ChrisHuie We are gonna wait until  @npeceniak has a chance to review, please hold off on merging!

modules/spotxBidAdapter.js
npeceniak: @robertrmartinez Does `loadExternalScript` put the script on the page during its execution? If that is the case I do not think this will work properly. We need all those data attributes set on the script when it initially loads in order to use anything other that the default options values. Adding the data attributes after the script is already loaded on the page will be ineffective since the script process those attributes on initial load.

modules/spotxBidAdapter.js
npeceniak: The "slot" use case is used when the publisher wants to load the ad in a specific element that already exists on the page. If we just insert the easi script in the head it will create a new element and render the video in that newly created element. This would have the effect of not allowing the publisher to place the ad in the correct place on the page.

npeceniak: @spotxslagle Can you confirm this is still the case. It has been a few years since I last looked at these integrations so I may be misremembering some of the configuration details.

modules/spotxBidAdapter.js
robertrmartinez: So, I am not sure but here is what I think happens, please feel free to look at the code for this function though: https://github.com/prebid/Prebid.js/blob/master/src/adloader.js#L29

If the script URL has not been loaded yet => then it will fire it async and return the script object, so we prob get into race conditions here right? Most of time the data attributes prob get attached before script is loaded, but if it is cached then maybe not?

It saves the scriptTag into a cache lookup so next time the func is called fro same URL on same doc it will just return it instead of firing the URL again.
https://github.com/prebid/Prebid.js/blob/master/src/adloader.js#L42

I had a feeling the `loadExternalScript` function would not handle all of the spotX use cases.

Now we need to know if we need to figure out what to do.

spotxslagle: I can confirm that we would be getting into a race condition here. I think the fix could be as simple as adding a param to that function that would allow us to pass an array or object allowing us to set attributes on the script. I haven't ever had to modify prebid code outside our adapter, so I don't know how that change would go over.

spotxslagle: This change would also break the custom render function. The point of the render function is for the pub to put our completed script on the page themselves. This implementation would have then trying to append an undefined object to the page, which would definitely result in an error.

modules/spotxBidAdapter.js
patmmccann: Perhaps missed showheroes, however, why not specify the script as a renderer rather than doing the append child?

ChrisHuie:
Closing this pr since it appears to moved to #8624

spotxslagle:
I believe this PR can be closed. #8624 makes this obsolete.

PR #8541, opened 2022-06-08
Sonobi - Fix gpid bid request value and add media type value

JonGoSonobi:
Fixed an issue where the key_maker value in the bid request was malformed when adding gpid. Added new media type value to the bid request.

## Type of change

- [x] Bugfix

## Description of change
This change fixes an issue where the key_maker value in the bid request was malformed. It should be a csv of values after the sizes. Additionally, added a value to determine if the bid is display or video.

- apex.prebid@sonobi.com

### DIFF
--- a/modules/sonobiBidAdapter.js
+++ b/modules/sonobiBidAdapter.js
@@ -56,15 +56,23 @@ export const spec = {
    */
   buildRequests: (validBidRequests, bidderRequest) => {
     const bids = validBidRequests.map(bid => {
+      let mediaType;
+      if (deepAccess(bid, 'mediaTypes.banner')) {
+        mediaType = 'display';
+      }
+      if (deepAccess(bid, 'mediaTypes.video')) {
+        mediaType = 'video';
+      }
+
       let slotIdentifier = _validateSlot(bid);
       if (/^[\/]?[\d]+[[\/].+[\/]?]?$/.test(slotIdentifier)) {
         slotIdentifier = slotIdentifier.charAt(0) === '/' ? slotIdentifier : '/' + slotIdentifier;
         return {
-          [`${slotIdentifier}|${bid.bidId}`]: `${_validateSize(bid)}${_validateFloor(bid)}${_validateGPID(bid)}`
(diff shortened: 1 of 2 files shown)


modules/sonobiBidAdapter.js
```diff
@@ -56,15 +56,23 @@ export const spec = {
    */
   buildRequests: (validBidRequests, bidderRequest) => {
     const bids = validBidRequests.map(bid => {
+      let mediaType;
+      if (deepAccess(bid, 'mediaTypes.banner')) {
+        mediaType = 'display';
+      }
+      if (deepAccess(bid, 'mediaTypes.video')) {
+        mediaType = 'video';
+      }
```
mmoschovas, line 65: Can this be written as the following so this does not run both conditions if it is not necessary:

```
if (deepAccess(bid, 'mediaTypes.video')) {
        mediaType = 'video';
} else if (deepAccess(bid, 'mediaTypes.banner')) {
        mediaType = 'display';
}      
```

modules/sonobiBidAdapter.js
```diff
@@ -323,11 +331,23 @@ function _validateGPID(bid) {
...
+  if (mediaType === 'video') {
+    return 'c=v,';
+  }
+
+  if (mediaType === 'display') {
+    return 'c=d,';
+  }
+
+  return '';
+}
```
mmoschovas, line 349: I would consider writing this as a ternary or setting the values to a variable to avoid the multiple return statements. Something like this:

```
function _validateMediaType(mediaType) {
     return mediaType === 'video' ? 'c=v,' : (mediaType === 'display' ?  'c=d,' : '');
}
```

or 

```
function _validateMediaType(mediaType) {
  let mediaTypeValidation = '';
  if (mediaType === 'video') {
    mediaTypeValidation = 'c=v,';
  } else if (mediaType === 'display') {
     mediaTypeValidation = 'c=d,';
  }

  return  mediaTypeValidation;
}
```

mmoschovas (changes requested):
Just a few notes here otherwise looks good

mmoschovas (approved):

PR #8546, opened 2022-06-09
Missena Bid Adapter: added getUserSyncs and testmode

pdamoc:
## Type of change
- [x] Feature

## Description of change

Added support for User Syncs.

The PR also contain a very small update to the params by adding a testmode field. This is entirely for our team's debugging needs and does not affect the users of this adapter.


pdamoc:
Unfortunate typo

PR #8550, opened 2022-06-10
YahooSSP Bid Adapter: Extend banner mimes to include image/gif for display inventory (6.29.x)

slimkrazy:
## Type of change
- [x] Other

## Description of change
- Extended banner mimes to include `image/gif` for display inventory
- Set XHR content type to text.plain to prevent preflight requests

### DIFF
--- a/modules/yahoosspBidAdapter.js
+++ b/modules/yahoosspBidAdapter.js
@@ -239,7 +239,7 @@ function generateOpenRtbObject(bidderRequest, bid) {
       cur: [getFloorModuleData(bidderRequest).currency || deepAccess(bid, 'params.bidOverride.cur') || DEFAULT_CURRENCY],
       imp: [],
       site: {
-        page: deepAccess(bidderRequest, 'refererInfo.page'),
+        page: deepAccess(bidderRequest, 'refererInfo.canonicalUrl') || deepAccess(bidderRequest, 'refererInfo.referer')
       },
       device: {
         dnt: 0,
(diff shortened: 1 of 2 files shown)


patmmccann:
Thanks @slimkrazy !

Have you noticed content type application json is causing preflight requests on this adapter? Sticking with text plain might improve performance

dgirardi:
`page` was introduced in 7; [in 6.29.x](https://github.com/prebid/Prebid.js/blob/302411b3c98fc4d5bbcaf8a292271d3acf1facfd/src/refererDetection.js#L174-L181) I suggest using `canonicalUrl || referer` (although see [this thread](https://github.com/prebid/Prebid.js/pull/8450#discussion_r882785864) for an edge case regarding AMP).

slimkrazy:
@dgirardi - thanks for that clarification. I was thinking we'd made a blunder using referrer!

I'll make the change to prevent the preflight requests when I get back to my machine

dgirardi:
I'm not sure if it's something on my end, but I get this:
yahoocors

patmmccann:
Generally, we only allow changes to legacy that are also in master. What is this only on the legacy branch?

ChrisHuie:
@slimkrazy following up on this only being added to legacy and not master? Was that intentional for a specific reason?

patmmccann:
If you plan to make these changes to legacy, please open a pull to master as well before re-opneing

PR #8559, opened 2022-06-14
cpex Id System: Fixes and cleanup, adds eids, changes cookie name

EskelCz:
## Type of change
- [X] Bugfix

## Description of change
- `getId` has to return key `id` to trigger passing the id to bidders
- removed unused params
- added eids version
- renamed cookie to not be mistaken with some Chinese id
- not adding the id when cookie is not found
- updated tests to match the changes


patmmccann:
Please note test failures

EskelCz:
@jlukas79 Can we please get an estimate when this can be merged? It's a small change concerning just our code. Thanks

EskelCz:
@patmmccann Please is there anyone else who can do the review? Should be no work at all I think.

patmmccann:
Looks almost done, could you add tests to eids spec file?

patmmccann (approved):

PR #8564, opened 2022-06-15
Concert Bid Adapter: adding referrer

andrew-a-dev:
## Type of change
- [x] Other

Specifically: adding functionality to the existing Concert adapter

## Description of change

Adding a `ref` parameter to the bid calls, using the various "referrer" properties available in Prebid. 

(The PR also contains linting-related changes to the file as a whole, to do with `;` and whitespace.)

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,6 +1,6 @@
 {
   "name": "prebid.js",
-  "version": "7.2.0",
+  "version": "7.3.0-pre",
   "lockfileVersion": 2,
   "requires": true,
   "packages": {
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "prebid.js",
-  "version": "7.2.0",
+  "version": "7.3.0-pre",
   "description": "Header Bidding Management Library",
   "main": "src/prebid.js",
   "scripts": {


modules/concertBidAdapter.js
```diff
@@ -57,8 +57,9 @@ export const spec = {
         slotType: bidRequest.params.slotType,
         adSlot: bidRequest.params.slot || bidRequest.adUnitCode,
         placementId: bidRequest.params.placementId || '',
-        site: bidRequest.params.site || bidderRequest.refererInfo.page
-      }
+        site: bidRequest.params.site || bidderRequest.refererInfo.page,
+        ref: bidderRequest.refererInfo.ref || window.document.referrer
```
dgirardi, line 61: I'm curious about the fallback (because this choice is common enough that I wonder if it should be built into `refererInfo`). It would be used when prebid is running inside an isolated iframe, and its value would usually be the location of the parent frame - why is that preferable to `null`? Not knowing how this is used in the backend it looks like pollution of the data to me. (sometimes `ref` is the referrer, sometimes not, and you don't know which is which).

dgirardi (approved):
The changes LGTM - but I am curious about it if you want to endulge me.

patmmccann:
circleci did not run

modules/concertBidAdapter.js
andrew-a-dev: @dgirardi You're right, that wasn't properly thought through; we don't want the `iframe`'s referrer. I see the pull request has been closed because of lack of CI, so I'll open another one with an update to the line above. Thanks for calling that out!

andrew-a-dev: Oh no, I just realised _I_ inadvertently closed this PR by messing around with `master` on our fork. 🤦 Sorry!

PR #8575, opened 2022-06-17
Prebid Core: Set permissions for GitHub actions

naveensrinivasan:
Restrict the GitHub token permissions only to the required ones; this way, even if the attackers will succeed in compromising your workflow, they won’t be able to do much.

- Included permissions for the action. https://github.com/ossf/scorecard/blob/main/docs/checks.md#token-permissions

https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#permissions

https://docs.github.com/en/actions/using-jobs/assigning-permissions-to-jobs

[Keeping your GitHub Actions and workflows secure Part 1: Preventing pwn requests](https://securitylab.github.com/research/github-actions-preventing-pwn-requests/)

Signed-off-by: naveen <172697+naveensrinivasan@users.noreply.github.com>

### DIFF
--- a/.github/workflows/issue_tracker.yml
+++ b/.github/workflows/issue_tracker.yml
@@ -3,8 +3,13 @@ on:
   issues:
     types:
       - opened
+permissions:
+  contents: read
+
 jobs:
   track_issue:
+    permissions:
+      contents: none
     runs-on: ubuntu-latest
     steps:
       - name: Generate token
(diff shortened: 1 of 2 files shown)


dgirardi (approved):
LGTM - @patmmccann, I think the best way to handle this is to just merge (I don't think it can have any impact on pbjs itself) and keep an eye on the release drafter to verify that it keeps working as expected. The issue tracker is easier to test because it just needs a new issue rather than a PR merge.

PR #8577, opened 2022-06-17
Criteo Bid Adapter: add iframe syncs support

afewcc:
## Type of change

- [ ] Bugfix
- [x] Feature
- [ ] New bidder adapter   
- [ ] Build related changes
- [ ] CI related changes
- [ ] Other

## Description of change

Added support for user sync through iframe as possible sync option

## Contacts
- Tachfine El Belky (t.elbelky@criteo.com)

### DIFF
--- a/modules/criteoBidAdapter.js
+++ b/modules/criteoBidAdapter.js
@@ -6,6 +6,8 @@ import { BANNER, NATIVE, VIDEO } from '../src/mediaTypes.js';
 import { find } from '../src/polyfill.js';
 import { verify } from 'criteo-direct-rsa-validate/build/verify.js'; // ref#2
 import { getStorageManager } from '../src/storageManager.js';
+import { getRefererInfo } from '../src/refererDetection.js';
+import { hasPurpose1Consent } from '../src/utils/gpdr.js';
 
 const GVLID = 91;
 export const ADAPTER_VERSION = 34;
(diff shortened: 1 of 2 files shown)


ChrisHuie:
> Hello @ChrisHuie I hope your are doing well :) Any chance to get this one reviewed in the coming days ? Thanks a lot

Should be able to review it in the next couple of days

modules/criteoBidAdapter.js
```diff
@@ -31,12 +32,108 @@ const PUBLISHER_TAG_URL_TEMPLATE = 'https://static.criteo.net/js/ld/publishertag
...
+        origin: origin,
+        requestId: requestId,
+        secureIdCookie: readFromAllStorages(SID_COOKIE_NAME),
+        tld: refererInfo.domain,
+        topUrl: refererInfo.domain,
+        uid: readFromAllStorages(IDCPY_COOKIE_NAME),
+        version: '$prebid.version$'.replace(/\./g, '_'),
+      };
+
+      window.addEventListener('message', function (event) {
```
dgirardi, line 91: this leaves a dangling listener each time it's run; imo it should detach itself when it gets the matching requestId back

modules/criteoBidAdapter.js
```diff
@@ -209,6 +306,34 @@ export const spec = {
...
+  storage.setCookie(name, value, expires);
+  storage.setDataInLocalStorage(name, value);
+}
+
+function deleteFromAllStorages(name) {
+  storage.setCookie(name, '', 0);
+  storage.removeDataFromLocalStorage(name);
+}
+
+function hasPurpose1Consent(gdprConsent) {
```
dgirardi, line 330: You could use this utility: https://github.com/prebid/Prebid.js/blob/0be4560e182e46c121c283f763cdef74b5ff9db9/src/utils/gpdr.js#L9

modules/criteoBidAdapter.js
```diff
@@ -31,12 +32,108 @@ const PUBLISHER_TAG_URL_TEMPLATE = 'https://static.criteo.net/js/ld/publishertag
...
+  getUserSyncs: function (syncOptions, _, gdprConsent, uspConsent) {
+    const fastBidVersion = config.getConfig('criteo.fastBidVersion');
+    if (canFastBid(fastBidVersion)) {
+      return [];
+    }
+
+    const refererInfo = getRefererInfo();
+    const origin = 'criteoPrebidAdapter';
+
+    if (syncOptions.iframeEnabled && hasPurpose1Consent(gdprConsent)) {
```
patmmccann, line 58: where is iframeEnabled coming from 

https://github.com/prebid/Prebid.js/blob/0be4560e182e46c121c283f763cdef74b5ff9db9/modules/pubmaticBidAdapter.js#L1300 is a more typical use of iframe syncs

patmmccann: i see, you are trying to decorate the sync call with these id's.

Can you please indicate in your docs that 'opting into iframe syncs with this bidder may set values in the first party context'

modules/criteoBidAdapter.js
dgirardi: I'm also curious: how does the sync figure out which window to message? this is the first time I see this approach and I'm wondering if it needs support from core for the general case (where prebid might be in a window other than `top`).

leonardlabat: From the sync frame, we simply try to send the message to window.parent (which should be the one listening to the message as it's the one which dropped the frame)

modules/criteoBidAdapter.js
leonardlabat: done

modules/criteoBidAdapter.js
leonardlabat: Regarding the _hasPurpose1Consent_ call
We really want the purpose1 to be granted prior to init our user sync as it involves interacting with 1PC the LS to resolve our sync endpoint arguments.

patmmccann (approved):

Issue #8337, opened 2022-04-28
Use of document.write method to write an HTML string without standard HTML declaration  


nikpil93:
https://github.com/prebid/Prebid.js/blob/061eb3c9c295d1e9bb9af62e1ba2370781f0ad17/src/prebid.js#L496

In order to write the HTML code with the best practice it would need to declare the doctype html.
For example:  
<!doctype html>
<html lang="en">
  <head>
    <meta charset="utf-8">
    <meta name="viewport" content="width=device-width, initial-scale=1">
    <title>iframe_container</title>
    ....
  </head>
  <body>
    ....
  </body>
</html>

So an idea is to use DOMParser to get the head elements and body elements as parts of the ad which is retreived by the winning bidder in order to eventually write it properly inside the iframe given in the renderAd method of pbjs library.

patmmccann:
It seems we should eliminate all use of document.write in the project? It periodically comes up as causing problems. https://github.com/prebid/Prebid.js/blob/2d14b7d53337c47b2e804f184927ed498990cc8f/src/utils.js#L582 also seems to be another place this happens.

patmmccann:
https://github.com/prebid/Prebid.js/issues/7049 is another example of this.

nikpil93:
This would be the best solution in my opinion too!

patmmccann:
@nikpil93 what is the usefulness of the doctype? We were actually considering deleting it from markup when received on a bid https://github.com/prebid/prebid-universal-creative/issues/134

nikpil93:
When the browser tries to parse and run the code if the the doctype html isn’t defined the browser is running in non standard mode and from experience and some articles I’ve read the execution of the code is not guaranteed and consistent always.

I’ve personally seen weird behavior on the subject of ad serving where the iframes I created didn’t have doctype and resulted in huge discrepancies and when I refactored it to contain doctype, everything went back to normal.

Articles: 1) https://developer.mozilla.org/en-US/docs/Web/HTML/Quirks_Mode_and_Standards_Mode
2) https://quirks.spec.whatwg.org/

patmmccann:
We'll research the feasibility of removing doc writes from the project and/or postscribe from the puc

patmmccann:
@nikpil93 we are having some difficulty reproducing an issue with doctype missing. Further, we are looking at removing doctype from markup to solve https://github.com/prebid/prebid-universal-creative/issues/134 . Adding the doctype seems to turn off 'quirks mode'

[...]

PR #8587, opened 2022-06-20
Prebid core: do not use `document.write` for rendering HTML

dgirardi:
## Type of change

## Description of change

Replace core's usage of `document.write` with `Element.innerHTML`, with some additional logic to run scripts.

According to:

https://www.w3.org/TR/2008/WD-html5-20080610/dom.html#innerhtml0
https://developer.mozilla.org/en-US/docs/Web/API/Element/innerHTML

'script' is the only major difference called out between  `document.write` and `innerHTML`, with the rest being about state of the browser's internal parsers - which should hopefully not matter for us.

## Other information

Closes https://github.com/prebid/Prebid.js/issues/8337

### DIFF
--- a/src/utils/writeHTML.js
+++ b/src/utils/writeHTML.js
@@ -1,0 +1,44 @@
+/**
+ * Replace the contents of `element` with the given `html` string, running all scripts contained within it.
+ *
+ * @param element the container element. Must be in the DOM.
+ * @param html HTML string
+ */
+export function replaceHTML(element, html) {
+  element.innerHTML = html;
+  rebuildScriptsIn(element);
+}
+
+/**
+ * Append some html to the contents of `element`, running all scripts contained within it.
+ *
+ * @param element the container element to append to. Must be in the DOM.
+ * @param html HTML string
+ */
+export function appendHTML(element, html) {
+  const container = element.ownerDocument.createElement('template');
+  container.innerHTML = html;
+
+  Array.from(container.content.childNodes).forEach((node) => {
+    if (node?.tagName === 'SCRIPT') {
+      node = cloneScript(node);
+    }
+    if (typeof node?.querySelectorAll === 'function') {
+      rebuildScriptsIn(node);
+    }
+    element.appendChild(node);
+  });
+}
+
+function cloneScript(node) {
+  const repl = node.ownerDocument.createElement('script');
+  Array.from(node.attributes).forEach((attr) => repl.setAttribute(attr.name, attr.value));
+  repl.appendChild(node.ownerDocument.createTextNode(node.innerHTML));
+  return repl;
+}
+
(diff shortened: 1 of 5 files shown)


src/utils/writeHTML.js
```diff
@@ -0,0 +1,44 @@
+/**
+ * Replace the contents of `element` with the given `html` string, running all scripts contained within it.
+ *
+ * @param element the container element. Must be in the DOM.
+ * @param html HTML string
+ */
+export function replaceHTML(element, html) {
```
mkendall07, line 7: My only concern with this approach is that I believe it makes scripts execute asynchronously, where as using `document.write` will execute synchronous. So you might hit some edge cases here.

dgirardi:
As it turns out from the test failures, it's more complicated than even the sync/async problem. I don't think this approach will work - but maybe someone has a better idea. We could try to pull in postscribe, but I'm not sure it's worth it.

In Firefox, newly created iframes behave like this:

```
 const iframe = document.createElement('iframe');
 document.appendChild(iframe);
 const doc = iframe.contentDocument;
 doc.body.appendChild(...) // does not work as expected, because:
 iframe.contentWindow.addEventListener('DOMContentLoaded', function() {
    doc === iframe.contentDocument // this is false in Firefox
    doc.defaultView.document === iframe.contentDocument // but this, somehow, is true
 })
```

I spent some time to try and code that in this PR, but I am running into other problems - I think because  scripts do not run at the expected time.

patmmccann:
Should core import postscribe?

nikpil93:
> As it turns out from the test failures, it's more complicated than even the sync/async problem. I don't think this approach will work - but maybe someone has a better idea. We could try to pull in postscribe, but I'm not sure it's worth it.
> 
> In Firefox, newly created iframes behave like this:
> 
> ```
>  const iframe = document.createElement('iframe');
>  document.body.appendChild(iframe);
>  const doc = iframe.contentDocument;
>  doc.body.appendChild(...) // does not work as expected, because:
>  iframe.contentWindow.addEventListener('DOMContentLoaded', function() {
>     doc === iframe.contentDocument // this is false in Firefox
>     doc.defaultView.document === iframe.contentDocument // but this, somehow, is true
>  })
> ```
> 
> I spent some time to try and code that in this PR, but I am running into other problems - I think because scripts do not run at the expected time.

Hello, couldn’t you use the srcdoc attribute of the iframe (srcdoc=“html code”) in order to apply the htmlcontent accordingly? It seems that low effort is needed and you could combine the src=“javascript:false” to avoid the default iframe and DOM behavior. Am I missing something?
CC: @dgirardi @mkendall07 @patmmccann 
https://developer.mozilla.org/en-US/docs/Web/API/HTMLIFrameElement/srcdoc

Issue #8588, opened 2022-06-20
gulp build/test failing on Mac OS


shahinrahbariasl:
## Type of issue
bug

## Description
Have been running into issues with gulp build/test after rebasing on latest master branch:
```
shahinrahbariasl-MacBook-Pro:prebid.js shahin.rahbariasl$ gulp test
Error: Cannot find module './index.js'
Require stack:
- /Users/shahin.rahbariasl/Desktop/IX/prebid.js/gulpHelpers.js
- /Users/shahin.rahbariasl/Desktop/IX/prebid.js/webpack.conf.js
- /Users/shahin.rahbariasl/Desktop/IX/prebid.js/gulpfile.js
- /usr/local/lib/node_modules/gulp-cli/lib/shared/require-or-import.js
- /usr/local/lib/node_modules/gulp-cli/lib/versioned/^4.0.0/index.js
- /usr/local/lib/node_modules/gulp-cli/index.js
- /usr/local/lib/node_modules/gulp-cli/bin/gulp.js
    at Function.Module._resolveFilename (internal/modules/cjs/loader.js:610:15)
    at Function.resolve (internal/modules/cjs/helpers.js:21:19)
    at /Users/shahin.rahbariasl/Desktop/IX/prebid.js/gulpHelpers.js:94:53
    at Array.map (<anonymous>)
    at Object.getLibraryFiles (/Users/shahin.rahbariasl/Desktop/IX/prebid.js/gulpHelpers.js:94:33)
    at /Users/shahin.rahbariasl/Desktop/IX/prebid.js/webpack.conf.js:48:40
    at Array.forEach (<anonymous>)
    at /Users/shahin.rahbariasl/Desktop/IX/prebid.js/webpack.conf.js:40:42
    at Object.<anonymous> (/Users/shahin.rahbariasl/Desktop/IX/prebid.js/webpack.conf.js:61:5)
    at Module._compile (internal/modules/cjs/loader.js:759:30) {
  code: 'MODULE_NOT_FOUND',
  requireStack: [
    '/Users/shahin.rahbariasl/Desktop/IX/prebid.js/gulpHelpers.js',
    '/Users/shahin.rahbariasl/Desktop/IX/prebid.js/webpack.conf.js',
    '/Users/shahin.rahbariasl/Desktop/IX/prebid.js/gulpfile.js',
    '/usr/local/lib/node_modules/gulp-cli/lib/shared/require-or-import.js',
[...]

PR #8592, opened 2022-06-21
Build system: do not use `require.resolve` for resolving library file paths

dgirardi:
## Type of change

- [x] Build related changes

## Description of change

Use `path.resolve` instead of `require.resolve` when resolving library file paths. `require.resolve` seems to behave unexpectedly in certain situations (https://github.com/prebid/Prebid.js/issues/8588)

## Other information

Closes https://github.com/prebid/Prebid.js/issues/8588

### DIFF
--- a/gulpHelpers.js
+++ b/gulpHelpers.js
@@ -91,7 +91,7 @@ module.exports = {
   },
   getLibraryFiles(name) {
     const library = libraries[name];
-    const files = library.files.map(file => require.resolve(file, {paths: ['./libraries/' + name + '/']}));
+    const files = library.files.map((file) => path.resolve('./libraries/', name, file))
     return files;
   },
   isLibrary(name) {


dgirardi:
@karimMourra could you confirm that the use of `require.resolve` was incidental - or did you want that to work with `node_modules` somehow?

ChrisHuie (approved):

karimMourra:
@dgirardi the decision was based on the steps for modules:
`var modulePath = require.resolve(module, {paths: [MODULE_PATH, LIBRARY_PATH]});` in `getModules`. I notice that the line is wrapped in a `try-catch` block, perhaps because of the node-module issues you mention.
 I am OK with using path.resolve. Thank you for fixing!

karimMourra (approved):

Issue #8561, opened 2022-06-14
pbs adapter: support bidder-specific schains


bretg:
## Type of issue
enhancement

## Description
In PBJS, the way to set bidder-specific schains is with `setBidderConfig`:

```
pbjs.setBidderConfig({
  "bidders": ['bidderA'],
  "config": {
    "schain": {
      "validation": "relaxed",
      "config": {
        "ver":"1.0",
        "complete": 1,
        "nodes": [
          {
            "asi":"example.com",
            "sid":"11111",
            "hp":1
          }
        ]
      }
    }
  }
});
```

However, Prebid Server supports bidder-specific schains on ext.prebid.schains, not ext.prebid.bidderconfig. 
```
s2sConfig: {
  ... 
  extPrebid: {
    schains: [
      { "bidders": ["bidderA"], "schain": { "ver":"1.0", "complete": 1, "nodes": [{"asi":"example.com", "sid":"11111", "hp":1 }]}}
    ]
  }
```

The PBS bid adapter currently only looks for ortb2 in bidderConfig. This issue tracks having it map bidderConfig schains to the PBS location at ext.prebid.schains.

1) Loop through bidder-specific config
2) if config.schains exists, loop through each bidder
     a) If ext.prebid.schains already exists for that bidder, skip this bidder-specific config with a informational log "bidder-specific schain for bidderA skipped due to existing entry"
     b) otherwise, create the ext.prebid.schains entry
```
{
    bidders: COPY-FROM-BIDDERCONFIG.bidders,
    schain: COPY-FROM-BIDDERCONFIG.config.schain.config
}
```

jlquaccia:
looking into this issue

jlquaccia:
Just pushed up another commit with a few tests for this issue just now

PR #8594, opened 2022-06-21
PBS Adapter: Support Bidder-Specific Schains

jlquaccia:
## Type of change

- [ ] Bugfix 

## Description of change
Added support for bidder-specific schains within the PBS adapter module

## Other information
https://github.com/prebid/Prebid.js/issues/8561


jlquaccia:
added a couple of tests for this issue just now

jlquaccia:
Just a reminder to review whenever one of you have time.  Thanks! @robertrmartinez @ChrisHuie or @musikele

modules/prebidServerBidAdapter/index.js
```diff
@@ -861,6 +861,54 @@ Object.assign(ORTB2.prototype, {
       request.ext.prebid = mergeDeep(request.ext.prebid, s2sConfig.extPrebid);
     }
 
+    // get reference to pbs config schain bidder names (if any exist)
+    let pbsSchainBidderNamesArr = [];
+    if (request.ext.prebid.schains) {
+      for (let i = 0; i < request.ext.prebid.schains.length; i++) {
```
dgirardi, line 867: `request.ext.prebid.schains.flatMap(s => s.bidders)`

modules/prebidServerBidAdapter/index.js
```diff
@@ -861,6 +861,54 @@ Object.assign(ORTB2.prototype, {
...
+        pbsSchainBidderNamesArr = pbsSchainBidderNamesArr.concat(request.ext.prebid.schains[i].bidders);
+      }
+    }
+
+    // get reference to pbjs bidder specific schain values (if any exist)
+    let pbjsSchainValuesArr = [];
+
+    // iterate over pbjs bidder specific schains (if any exist)
+    for (let i = 0; i < bidRequests.length; i++) {
```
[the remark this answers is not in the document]
jlquaccia: Thanks for the review @dgirardi.  Good point, I agree that it would make more sense to update the request itself instead.  I went back and modified my PR, I like the route you took with your suggestion above and integrated it into my PR.  Let me know if you think any other changes should be made or if the logic seems good.

modules/prebidServerBidAdapter/index.js
```diff
@@ -861,6 +861,33 @@ Object.assign(ORTB2.prototype, {
...
+    const schains = Object.fromEntries(
+      (request.ext.prebid?.schains || []).map(({bidders, schain}) => [JSON.stringify(schain), {bidders: new Set(bidders), schain}])
+    );
+
+    // compare bidder specific schains with pbs specific schains
+    request.ext.prebid.schains = Object.values(
+      bidRequests
+        .map((req) => [req.bidderCode, req.bids[0].schain])
+        .reduce((chains, [bidder, chain]) => {
+          const chainKey = JSON.stringify(chain);
```
dgirardi, line 876: I'm assuming here that as long a bidder is associated only with one chain, it's OK to have duplicates due to different serialization of equivalent schains, e.g. `[{bidders: ['A'], schain1}, {bidders: ['B'], schain2}]` works for the backend even if `schain1` and `schain2` are identical.

modules/prebidServerBidAdapter/index.js
```diff
@@ -861,6 +861,33 @@ Object.assign(ORTB2.prototype, {
...
+          if (chainKey && !chains.hasOwnProperty(chainKey)) {
+            if (pbsSchainBidderNamesArr.indexOf(bidder) !== -1) {
+              logInfo(`bidder-specific schain for ${bidder} skipped due to existing entry`);
+            } else {
+              chains[chainKey] = {bidders: new Set(), schain: chain};
+              chains[chainKey].bidders.add(bidder);
```
dgirardi, line 882: I believe the `bidders.add(bidder)` needs to be detented out of `if (!chains.hasOwnProperty(chainKey))`; otherwise, if multiple bidders have chains that serialize to the same key, you are losing the association for every one of them except the first. 

This would also be a good test case, if I understand your intent, you want two separate bid requests with identical chains to be "merged" into a single entry in ext.prebid.schains.

modules/prebidServerBidAdapter/index.js
```diff
@@ -861,6 +861,33 @@ Object.assign(ORTB2.prototype, {
...
+    );
+
+    // compare bidder specific schains with pbs specific schains
+    request.ext.prebid.schains = Object.values(
+      bidRequests
+        .map((req) => [req.bidderCode, req.bids[0].schain])
+        .reduce((chains, [bidder, chain]) => {
+          const chainKey = JSON.stringify(chain);
+          if (chainKey && !chains.hasOwnProperty(chainKey)) {
+            if (pbsSchainBidderNamesArr.indexOf(bidder) !== -1) {
```
dgirardi, line 878: this check also appears in the wrong place to me, as a consequence of the above. it should guard the `bidders.add` call.

modules/prebidServerBidAdapter/index.js
jlquaccia: adjusted the logic for this and also added a test case

jlquaccia:
hey @dgirardi, just pushed some updates can you review when you have time?

robertrmartinez (approved):
Looks good to me

modules/prebidServerBidAdapter/index.js
```diff
@@ -861,6 +861,43 @@ Object.assign(ORTB2.prototype, {
...
+              break;
+            default:
+          }
+
+          return chains;
+        }, schains)
+    ).map(({bidders, schain}) => ({bidders: Array.from(bidders), schain}));
+    // if schains evaluates to an empty array, remove it from the prebid object
+    if (request.ext.prebid.schains.length === 0) delete request.ext.prebid.schains;
+
```
dgirardi, line 900: I don't know if this is in scope for this, but I noticed something while working on a separate issue: on line 918, the PBS adapter sets `source.ext.schain` with the first schain it finds. Does that make sense? is it just ignored server side?

jlquaccia: hmm, looks as though `source.ext.schain` only gets added if an schain was configured using `pbjs.setBidderConfig` and the bidder the schain was configured for is the same as what is set under `defaultVendor` configured in the `s2sConfig`..

logic appears to be something like that, but I am not 100% sure on what the expected logic is supposed to be here.

patmmccann: let's keep that out of scope, @dgirardi did you already open another issue for this?

PR #8597, opened 2022-06-22
Appnexus Bid Adapter: support native feature flags

jsnellbaker:
## Type of change

- [x] Feature

## Description of change
Context for the underlying Feature Flags setting: #8219

Updating the appnexusBidAdapter to support the native-based Feature Flags.

### DIFF
--- a/modules/appnexusBidAdapter.js
+++ b/modules/appnexusBidAdapter.js
@@ -447,7 +447,7 @@ export const spec = {
    * @param {Bid} bid
    */
   onBidWon: function (bid) {
-    if (bid.native) {
+    if (FEATURES.NATIVE && bid.native) {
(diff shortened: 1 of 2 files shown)


dgirardi:
It looks good to me - and yes, the idea is to tag out use, not definitions (the latter should fail in tests). You can use `build-bundle-verbose` to see what happens after dead code elimination:

```
gulp build-bundle-verbose --disable NATIVE
```

result: https://gist.github.com/dgirardi/23ff993a7e09119eb32de774e8ca0caf

the only two remaining relics I see are:

 - `supportedMediaTypes` still contains NATIVE
 - `onBidWon` is still present, but empty

both should not matter, you could tag them as well but I personally think that's too much ugliness for how much you save out of it.

dgirardi (approved):

Issue #7936, opened 2022-01-13
Bundle size inflated due to intra-module dependencies


dgirardi:
## Type of issue
Bug

## Description
Prebid bundles are generated by concatenating "prebid-core.js" with a user-selected list of module files, which contain all dependencies of the module that are not in core. 

Since there is no real dependency tracking beside "core" or "not core", dependencies can be repeated across module files (if they [import from each other](https://github.com/prebid/Prebid.js/blob/4183f5d8f0875098e3041b366394ce24c43493cd/modules/adlooxRtdProvider.js#L14), or if they [import the same non-core file](https://github.com/prebid/Prebid.js/blob/master/src/AnalyticsAdapter.js)).

Fixing this requires coordination with external bundle vendors.

## Steps to reproduce

```
gulp build-bundle-dev --modules "adlooxRtdProvider,adlooxAnalyticsAdapter,id5AnalyticsAdapter"
```

Generated `prebid.js`: https://gist.github.com/dgirardi/86947f196a25fc753fbb6403094b97fe

- `AnalyticsAdapter` definition is repeated at lines #14737, #15830, #16447
- `adlooxAnalyticsAdapter` module is repeated at lines #14377, #15470

Problem persists in minified bundle, just harder to spot:

```
gulp build --modules "adlooxRtdProvider,adlooxAnalyticsAdapter,id5AnalyticsAdapter"
```

Generated `prebid.js`: https://gist.github.com/dgirardi/c3bb01de917407746950ec597cefbab2

- `"bundle"` (from [here](https://github.com/prebid/Prebid.js/blob/4183f5d8f0875098e3041b366394ce24c43493cd/src/AnalyticsAdapter.js#L28)) appears 3 times
- `window.pbjs.installedModules.push("adlooxAnalyticsAdapter")` appears twice

snapwich:
having duplicate code show up in a `prebid.js` bundle is not necessarily a sign of a problem (and with analytics adapter it's not, but not sure about adloox case). the issues you've identified are known and were addressed with what i think are acceptable solutions:

[...]

PR #8599, opened 2022-06-22
Prebid core: improve library support; make AnalyticsAdapter a library

dgirardi:
## Type of change

- [x] Feature

## Description of change

This:

 - improves library support to automatically generate libraries and track their dependencies, so that consumers of a library simply need to `import` symbols from it;
 - makes `AnalyticsAdapter.js` a library, so that it can be shared when multiple analytics adapters are included in the build.

## Other information

Closes https://github.com/prebid/Prebid.js/issues/7936
Original library PR: https://github.com/prebid/Prebid.js/pull/8527
Minor docs update: https://github.com/prebid/prebid.github.io/pull/3863


jsnellbaker (approved):
Checked the build file and saw the duplicate references for the analytics adapter code only appeared once, instead of dozens of time.  

So this appears to working from what I can tell.  Nice work :)

mmoschovas (approved):
lgtm

robertrmartinez:
@dgirardi 

Analytics Adapters have stopped initializing since this PR for me.

I am trying to pinpoint why now.

Seems the modules `registerAnalyticsAdapter` code is never executing, like the module is not even loading itself. I put breakpoints at global level of module and they do not get hit.

Bid Adapters and other modules are loaded just fine.

I have only tested with rubicon analytics adapter so far.

Wondering if it is the changes to webpack build? Will keep looking.

robertrmartinez:
Looks like `src/AnalyticsAdapter.js` is not being included in the build any longer.

Do we have to explicitly include it now?

robertrmartinez:
Looks like the `libraries/analyticsAdapter/AnalyticsAdapter.js` formerly known as `src/analyticsAdapter.js` is no longer auto included in my prebid builds using same old command as before.

Our process uses Prebid Packager https://github.com/prebid/Prebid.js-packager which I am thinking is the problem.

Since src/analyticsAdapter.js is not inside the generated rubiconAnalyticsAdapter.js, prebid packager needs to include it in the build. Guessing it does not handle libraries or something.

I'll continue to look into it.

robertrmartinez:
Yeah, so now looks like we need to update prebid packager to correctly include the necessary `library` files if modules from the build include it.

I see `analyticsAdapter.js` is output now after this PR when before it was not included in output files, it was just added into analyticsAdapter files themselves.

So this needs to be included in the build.

As I think we are the only ones working with prebid packager, I will try and fix this.

dgirardi:
@robertrmartinez the build now [generates a `dependencies.json`](https://github.com/prebid/Prebid.js/blob/73c13cde704ef396a152ed9dcb8596779473107d/gulpfile.js#L245) that links output files to the output "library" files they need. I was unfortunately completely unaware of the packager

robertrmartinez:
@dgirardi I whipped up something real quick which "solves" this. I know it could be written better / cleaner of course, but unfortunately I am slammed at the moment.

https://github.com/prebid/Prebid.js-packager/pull/4

I guess we should prob get off of this prebid packager sometime maybe, but it makes our system run quicker since we do not have to run gulp build every time we build a prebid wrapper.

Issue #8573, opened 2022-06-16
Split use of auctionid and transactionid in bidders for source.tid


patmmccann:
## Type of issue
Bug

## Description
Some bidders set auctionId, others set transactionId. Source.tid should always be from the same field. Also, the PBS adapter is missing this field. 

Discussed in the Prebid.js committee meeting. Here's the proposal:

1) Prebid will recognize 3 different IDs:
1a) auctionId (ORTB $.id). This should be unique for each request, so an auction with 5 bidders should have 5 `$.id`
1b) global transactionId (ORTB $.source.tid). This should be set to the auctionId
1c) adunit-level transactionId (ORTB $.imp[n].ext.tid).
2) All three IDs can be overridden by the publisher:
2a) auctionId - override already implemented as pbjs.requestBids as auctionId.
2b) global transactionId - override with setConfig({ortb});
2c) adunit-level transactionId - new override to be implemented as AdUnit.transactionId.

If the community agrees with this proposal:
- Prebid-core work will be done to support the reading, generation, and propagation of these 3 IDs.
- The pbsBidAdapter will be updated
- Docs will be updated
- We will update bid adapters that refer to any of these IDs and ask the maintainers to approve.

## Related
#8543 
#3190

Some bidders that appear to remain unsolved:
https://github.com/prebid/Prebid.js/blob/6e36e363aa843603a8b7cb29da3242d84e5f4b9d/modules/bizzclickBidAdapter.js#L95
@BizzClick @devBizzclick 

https://github.com/prebid/Prebid.js/blob/f32e0805a10c118d1c49e070d14ff80f9e3b7417/modules/rubiconBidAdapter.js#L187
@musikele 

https://github.com/prebid/Prebid.js/blob/6e36e363aa843603a8b7cb29da3242d84e5f4b9d/modules/gothamadsBidAdapter.js#L107
@supportGothamad 

https://github.com/prebid/Prebid.js/blob/6e36e363aa843603a8b7cb29da3242d84e5f4b9d/modules/luponmediaBidAdapter.js#L293
@adxpremium 

https://github.com/prebid/Prebid.js/blob/5ef54f7b2234cba7595ea4446f68b821af670bd6/modules/ozoneBidAdapter.js#L384
@AskRupert-DM 

https://github.com/prebid/Prebid.js/blob/6e36e363aa843603a8b7cb29da3242d84e5f4b9d/modules/adtrueBidAdapter.js#L510
@haibau 

[...]

PR #8608, opened 2022-06-23
Prebid Server Adapter: use bidRequest.auctionid as source.tid

patmmccann:
Related:

https://github.com/prebid/Prebid.js/pull/7585
https://github.com/prebid/Prebid.js/issues/8573
https://github.com/prebid/Prebid.js/pull/3190

7585 Made a new uuid for source.tid for prebid server because auction id was flawed. The first transaction id seems more appropriate, as other adapters can refer to it for source.tid as well. imp.ext.tid will be preferred by buyers in the long term, but source.tid has remaining interim utility and we want to unite their definitions

### DIFF
--- a/src/adapterManager.js
+++ b/src/adapterManager.js
@@ -341,7 +341,7 @@ adapterManager.callBids = (adUnits, bidRequests, addBidResponse, doneCb, request
   let counter = 0;
 
   // $.source.tid MUST be a unique UUID and also THE SAME between all PBS Requests for a given Auction
-  const sourceTid = bidRequests[0].transactionId;
+  const sourceTid = bidRequests.auctionId;
   _s2sConfigs.forEach((s2sConfig) => {
     if (s2sConfig && uniqueServerBidRequests[counter] && getS2SBidderSet(s2sConfig).has(uniqueServerBidRequests[counter].bidderCode)) {
       // s2s should get the same client side timeout as other client side requests.
--- a/src/adapterManager.js
+++ b/src/adapterManager.js
@@ -341,7 +341,7 @@ adapterManager.callBids = (adUnits, bidRequests, addBidResponse, doneCb, request
   let counter = 0;
 
   // $.source.tid MUST be a unique UUID and also THE SAME between all PBS Requests for a given Auction
-  const sourceTid = generateUUID();
+  const sourceTid = bidRequests[0].transactionId;
   _s2sConfigs.forEach((s2sConfig) => {
     if (s2sConfig && uniqueServerBidRequests[counter] && getS2SBidderSet(s2sConfig).has(uniqueServerBidRequests[counter].bidderCode)) {
       // s2s should get the same client side timeout as other client side requests.


patmmccann:
@bretg if PBS is to get a source.tid, 7585 and the comments in the source say it cannot be auctionId, this only leaves transactionid for the request[0]. It seems we should standardize on that.

patmmccann:
There is a strong case to make that if there is more than one impression in the request, source.tid should just be blank. If I changed it to that, Would that break anything @robertrmartinez ?

robertrmartinez:
It should just be left blank? I do not see how that could be valid if more than one impression is there.

I still think auctionId is the right approach, even with its downfalls.

From DSP perspective (Which I admit I am not knowledgeable about them much)

I would want to know an auctionId so I can say how many people are asking me for bids for a single auciton

As well as how many transactionId's came from that auction.

Either way, I still do not love the idea of setting sourct.tid to the first transactionId? They are different.

Maybe we should talk this out more but I like the idea of source.tid being the aucitonId, and if it is not a UUID (maybe we just check length of the string) then we generate one?

@bretg obvioulsy understands PBS more than I do. But from an Analytics perspective I would think having source.tid and `imp[x].tid` different is important.

patmmccann:
@robertrmartinez reopening as consensus seems to be building this is valuable (eg https://github.com/prebid/Prebid.js/pull/8679). 

Is blank even an option? Does pbs use this id for things other than shoving into source.tid ? 

I certainly understand the commentary that this id does not make sense in a multi-imp request, but when prebid server is making up its own id, it risks acting as a malicious entity, seemingly communicating these are new avails to downstream entitites, when in fact they are identical avails. 

Source.tid seems infinitely inferior to imp.ext.tid, but PRs to lower its cardinality to downstream entitites seem to be worthwhile.

bretg:
> Does pbs use this id for things other than shoving into source.tid ?

No - source.tid is just passed through to bidders.

> when prebid server is making up its own id

PBS does not make up any IDs other than a bid **response** ID, and that's placed in seatbid.bid.ext.prebid.bidid.  And that's only because we found that quite a few bidders return non-unique bid IDs that are unusable for tying auction events to impression events.

patmmccann:
> PBS does not make up any IDs other than a bid response ID,

I misspoke, the prebid server adapter makes it own one. 

> why having an auction-level ID isn't helpful.

It is, and bidRequests[0].transactionId is one and so is auctionID. The problem is that auctionid can be overridden and we want a value that is always unique, which is why this line was written in #7585

I cannot follow why having an auction id that doesn't collide with transaction id is useful. It actually seems to be a degradation of functionality from an analytics perspective.

bretg:
@patmmccann - I think I hear you saying that you want the pbsBidAdapter to accept whatever auctionId the pub defines, even if crappy. I'm ok with that.

patmmccann:
I'm confused, if that is the case, why was #7585 implemented?

bretg:
We debated what to do:
- some kind of validation (e.g. > 12 chars) // any validation rule is going to make someone unhappy
- just let it be crappy                                     // I've been burned by entities who do a bad job at creating IDs
- generate a proper UUID

There was no input about which way to go from the community. I'm willing to re-evaluate and admit having made the wrong choice.

patmmccann:
what about a fourth option, if anyone has overriden the auctionid that was generated, set it to blank?

patmmccann:
Another, simpler proposal comes to mind: always set tid zero to auctionid.

patmmccann:
@robertrmartinez I changed this to auction id

robertrmartinez:
Ok, I am good with this, with the understanding that if a pub does:

```js
pbjs.requestBids({
   auctionId: 'bobby',
});
```
or
```js
pbjs.requestBids({
   auctionId: ${some very small random id which will have many clashes},
});
```

that this ends up sucking for downstream entities.

But I consider this an implementation mistake and should be fixed by the pub to generate a true UUID or just let Prebid do its thing!

What do you all think?

@patmmccann @bretg @dgirardi

src/adapterManager.js
```diff
@@ -341,7 +341,7 @@ adapterManager.callBids = (adUnits, bidRequests, addBidResponse, doneCb, request
   let counter = 0;
 
   // $.source.tid MUST be a unique UUID and also THE SAME between all PBS Requests for a given Auction
-  const sourceTid = generateUUID();
+  const sourceTid = bidRequests.auctionId;
```
dgirardi, line 344: I don't think this works? bidRequests is an array here.

dgirardi:
On low entropy ids: because no pubs should have a reason to do that, and I'd expect it to be rare, I think it's fine. But at the same time I'm certain that whatever the goal of that feature is, there is some other way of achieving it that does not need to touch ids. I don't know the history though.

patmmccann:
this work will be incorporated into solution to #8573 pr

PR #8622, opened 2022-06-29
Adtargetme Bid Adapter: initial release

tarasmatokhniuk:
## Type of change
- [x] New bidder adapter

## Description of change
Adapter for AdserverAdtarget advertising system

```
const adUnits = [{
    code: 'your-placement',
    mediaTypes: {
            banner: {
                sizes: [[300, 250]]
            }
        },
    bids: [
        {
            bidder: 'adserveradtarget',
            params: {
                sid: 1220291391, // Site/App ID provided from SSP
            }
        }
    ]
}];
```

- info@adtarget.me

### DIFF
--- a/modules/adserveradtargetBidAdapter.js
+++ b/modules/adserveradtargetBidAdapter.js
@@ -5,7 +5,7 @@ import { config } from '../src/config.js';
 import { hasPurpose1Consent } from '../src/utils/gpdr.js';
 
 const INTEGRATION_METHOD = 'prebid.js';
-const BIDDER_CODE = 'adtargetme';
+const BIDDER_CODE = 'adserveradtarget';
 const ENDPOINT = 'https://z.cdn.adtarget.market/ssp?prebid&s=';
 const ADAPTER_VERSION = '1.0.0';
 const PREBID_VERSION = '$prebid.version$';
@@ -104,7 +104,7 @@ function validateAppendObject(validationType, allowedKeys, inputObject, appendTo
 };
 
 function getTtl(bidderRequest) {
-  const ttl = config.getConfig('adtargetme.ttl');
+  const ttl = config.getConfig('adserveradtarget.ttl');
   return ttl ? validateTTL(ttl) : validateTTL(deepAccess(bidderRequest, 'params.ttl'));
 };
 
@@ -295,7 +295,7 @@ function appendFirstPartyData(outBoundBidRequest, bid) {
 
 function generateServerRequest({payload, requestOptions, bidderRequest}) {
   return {
-    url: (config.getConfig('adtargetme.endpoint') || ENDPOINT) + (payload.site.id || ''),
+    url: (config.getConfig('adserveradtarget.endpoint') || ENDPOINT) + (payload.site.id || ''),
     method: 'POST',
     data: payload,
     options: requestOptions,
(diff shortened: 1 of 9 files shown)


ChrisHuie:
@tarasmatokhniuk can you please fix the linting errors in this pr?

tarasmatokhniuk:
@ChrisHuie, is everything ok? what else do i need to do?

tarasmatokhniuk:
Prebid Server requires the first 6 characters of an bidder/adapter name to be unique unless both adapters refer to the same company.
We change prebid js adapter name to be the same as prebid server adapter

ChrisHuie:
> @ChrisHuie, is everything ok? what else do i need to do?

modules/adserveradtargetBidAdapter.js
```diff
@@ -0,0 +1,408 @@
...
+  const mediaTypeMode = getMediaType(bid);
+
+  if (openRtbObject && bid) {
+    const impObject = {
+      id: bid.bidId,
+      secure: isSecure(bid),
+      bidfloor: getFloorModuleData(bid).floor || deepAccess(bid, 'params.bidOverride.imp.bidfloor') || 0.000001
+    };
+
+    if (bid.mediaTypes.banner && (typeof mediaTypeMode === 'undefined' || mediaTypeMode === BANNER)) {
```
mmoschovas, line 189: I believe `mediaTypeMode === BANNER` be sufficient here. Your `getMediaType` method return `'banner'` when `bid.mediaTypes.banner` exists and `false` otherwise. I don't see how this `undefined` use case is relevant here

modules/adserveradtargetBidAdapter.js
```diff
@@ -0,0 +1,408 @@
...
+        };
+        break;
+      case 'objectAllKeys':
+        if (isPlainObject(inputObject)) {
+          outputObject[objectKey] = inputObject[objectKey];
+        };
+        break;
+    };
+  };
+  return outputObject;
```
mmoschovas, line 103: Instead of using a switch statement and passing the validation type, could you simply pass the validating function and in the case of objectAllKeys pass an empty array as allowedKeys (although not sure you need this case as it is never used in your adapter)? Something like this to reduce the size and repetition in this function: 

function validateAppendObject(validationFunction, allowedKeys, inputObject, appendToObject) {
   const outputObject = {
      ...appendToObject
   };

    for (const objectKey in inputObject) {
        const hasKeys = allowedKeys.length;
        const isValid = (!hasKeys && validationFunction(inputObject)) || 
            (hasKeys && allowedKeys.indexOf(objectKey) !== -1 && validationFunction(inputObject[objectKey]);
       if (isValid) outputObject[objectKey] = inputObject[objectKey];
    }
}

modules/adserveradtargetBidAdapter.js
```diff
@@ -0,0 +1,408 @@
+import { registerBidder } from '../src/adapters/bidderFactory.js';
+import { BANNER } from '../src/mediaTypes.js';
+import { deepAccess, isFn, isStr, isNumber, isArray, isEmpty, isPlainObject, generateUUID, logWarn } from '../src/utils.js';
+import { config } from '../src/config.js';
+import { hasPurpose1Consent } from '../src/utils/gpdr.js';
+
+const INTEGRATION_METHOD = 'prebid.js';
+const BIDDER_CODE = 'adserveradtarget';
```
patmmccann, line 8: shorter bidder codes are preferred. Codes over 8 characters get truncated for some targeting keys to gam, eg hb_cache_id_adserveradtarget would truncate to hb_cache_id_adserver.

modules/adserveradtargetBidAdapter.js
tarasmatokhniuk: fixed

tarasmatokhniuk:
rename adapter as  [SyntaxNode advised to prebid server](https://github.com/prebid/prebid-server/pull/2296#issuecomment-1189213386)

mmoschovas (approved):
Thanks for making these updates! lgtm

PR #8633, opened 2022-06-30
Prebid Core: allows immediate rendering

spotxslagle:
## Type of change
- [X ] Feature

## Description of change
Allow immediate rendering by passing a flag in the options object passed through Renderer.install.

## Other information

@npeceniak @robertrmartinez

### DIFF
--- a/src/Renderer.js
+++ b/src/Renderer.js
@@ -14,7 +14,7 @@ const moduleCode = 'outstream';
  */
 
 export function Renderer(options) {
-  const { url, config, id, callback, loaded, adUnitCode } = options;
+  const { url, config, id, callback, loaded, adUnitCode, renderNow } = options;
   this.url = url;
   this.config = config;
   this.handlers = {};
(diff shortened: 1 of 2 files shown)


ChrisHuie:
@spotxslagle can you add some testing for this specific feature and pr please 🙏

spotxslagle:
Unit test has been added. Let me know if there's another aspect you'd like tested.

dgirardi:
what's the motivation for this? in what situation would the script not need to be loaded - and even if it doesn't, wouldn't the existing logic still work?

spotxslagle:
SpotX loads our script in our render function. This allows us to provide custom rendering options to publishers. We have been passing '/' as our URL, which seems to work in most circumstances, but we recently had a publisher for whom setting a src of '/' caused an error. Since the script never loaded, the callback never ran, and our render function was never called, so our ad never played. The other option I thought of was pointing to a url to load a script that did nothing, but we could feel confident would load. I didn't like that idea for obvious reasons. 

TLDR: SpotX's script needs to be loaded in our render function, not directly by Prebid, and existing logic breaks our ads for a minority of publishers.

dgirardi (approved):

Fawke (approved):

spotxslagle:
Are we waiting on anyone else before merging? If not, I don't have merge permissions. Would someone else be able to handle that?

patmmccann:
yeah we should document this and curious is there is an interplay with #8536 since spotx has some outside the rules behavior. Curious for feedback from @robertrmartinez 

@fowler446 might be able to give feedback on the best location to document this flag

spotxslagle:
#8536 has been superceded by https://github.com/prebid/Prebid.js/pull/8624. That pr is preliminary work for removing append child from the SpotX adapter. Other that dealing with merge conflicts that arise from having two stories both involving the same handful of spotx files, the stories are unrelated and there won't be interplay.

patmmccann:
In that case, this is only awaiting a link to the documentation PR

spotxslagle:
As far as I can tell, there is no appropriate place for this change to be documented. The renderer key on the bidResponse object found https://docs.prebid.org/dev-docs/bidder-adaptor.html#interpreting-the-response is missing, and the renderer object passed through that key is not defined anywhere. Since my change is to a function that generates that object, there is no documentation to change.

spotxslagle:
@patmmccann @ChrisHuie Can I get this merged if there are no other concerns?

patmmccann:
We're investigating the documentation hole you identified.

fowler446:
@patmmccann @spotxslagle 
I think the closest thing we have to documentation for the Renderer is above here:
https://docs.prebid.org/dev-docs/show-outstream-video-ads.html#step-2-show-ads-in-the-page-body

The docs direct people to the original Renderer PR at:
https://github.com/prebid/Prebid.js/pull/1082/files

That being said, I'm not sure this needs any documentation. Anyone seeking this level of info will need to be reading the code base. I can see how just passing`'/'` as the url can lead to weird behaviour. I believe in that case you would end up appending a script tag like `<script src='/'`
https://github.com/prebid/Prebid.js/blob/master/src/adloader.js#L108

spotxslagle:
Awesome! Thanks!

Issue #8628, opened 2022-06-30
Question - Prebid v6 to v7 


nicosapp:
## Type of issue
Question

## Description
We update the prebid script from 6.X to 7.X and we notice a different behaviour in the execution of our script. We had to modify it to make it works as before.

### Code version 6
```js
let ads = [...]
pbjs.setConfig(config);
pbjs.bidderSettings = settings;
pbjs.addAdUnits(getAdUnits(ads));

pbjs.requestBids({
        bidsBackHandler: () => {
            googletag.cmd.push(() => {
                pbjs.que.push(() => {
                    pbjs.setTargetingForGPTAsync();
                    this.loadAds(ads);
                });
            });
        }
    });

ads.forEach((ad) => {
      pbjs.removeAdUnit(ad.tagId);
  });
```
We use this code with version 6 and `requestBids` was executing before `removeAdUnits`
In version 7 `removeAdUnits` is executing before `requestBids`, so no bids are sent. I had to wrap the `removeAdUnit` with a `setTimeout` to make the `requestBids` call before the `removeAdUnits` call.

### Code version 7
```js
let ads = [...]
pbjs.setConfig(config);
pbjs.bidderSettings = settings;
pbjs.addAdUnits(getAdUnits(ads));

pbjs.requestBids({
        bidsBackHandler: () => {
            googletag.cmd.push(() => {
                pbjs.que.push(() => {
                    pbjs.setTargetingForGPTAsync();
                    this.loadAds(ads);
                });
            });
        }
    });
setTimeout(() => {
  ads.forEach((ad) => {
        pbjs.removeAdUnit(ad.tagId);
    });
}, 0)
```

## Platform details
Prebid version (old: 'v6.16.0', new v7.3.0)
Node version: v16.14.2
Npm version: 8.5
Linux, Ubuntu 20.04

dgirardi:
My suggestion is to use `requestBids`' [`adUnits` parameter](https://docs.prebid.org/dev-docs/publisher-api-reference/requestBids.html), which will sidestep this problem entirely:

```
pbjs.requestBids({
  adUnits: getAdUnits(ads)
})
```

[...]

PR #8637, opened 2022-07-01
Prebid core: fix adUnits for auction at the time requestBids is called

dgirardi:
## Type of change

- [x] Bugfix

## Description of change

If a publisher calls `pbjs.addAdUnits` or `pbjs.removeAdUnit` _after_ calling `requestBids`, those calls might have an effect on ongoing auctions - depending on whether any module needs to defer the start of the auction (for example, if `consentManagement` needs to look up consent data).

This changes `requestBids` to always use the adUnits that were defined at the moment it's called.

## Other information

Closes https://github.com/prebid/Prebid.js/issues/8628

### DIFF
--- a/src/prebid.js
+++ b/src/prebid.js
@@ -12,7 +12,7 @@ import { userSync } from './userSync.js';
 import { config } from './config.js';
 import { auctionManager } from './auctionManager.js';
 import { filters, targeting } from './targeting.js';
-import { hook } from './hook.js';
+import {hook, wrapHook} from './hook.js';
 import { loadSession } from './debugging.js';
 import {includes} from './polyfill.js';
 import { adunitCounter } from './adUnits.js';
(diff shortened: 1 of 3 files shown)


src/prebid.js
```diff
@@ -573,18 +573,28 @@ $$PREBID_GLOBAL$$.removeAdUnit = function (adUnitCode) {
...
+    const cbTimeout = timeout || config.getConfig('bidderTimeout');
+    logInfo('Invoking $$PREBID_GLOBAL$$.requestBids', arguments);
+    const ortb2Fragments = {
+      global: mergeDeep({}, config.getAnyConfig('ortb2') || {}, ortb2 || {}),
+      bidder: Object.fromEntries(Object.entries(config.getBidderConfig()).map(([bidder, cfg]) => [bidder, cfg.ortb2]).filter(([_, ortb2]) => ortb2 != null))
+    }
+    return startAuction({bidsBackHandler, timeout: cbTimeout, adUnits, adUnitCodes, labels, auctionId, ortb2Fragments});
+  }, 'requestBids');
+
+  return wrapHook(delegate, function requestBids(req = {}) {
```
jsnellbaker, line 588: I'm honestly still wrapping my head around these changes (though I think I've got the general thrust of it), but just to confirm something - is the `requestBids` name here just for nomenclature or is it serving a functional purpose for this new logic?

jsnellbaker:
Tagging @snapwich to possibly review the new code in the `hook.js`, to see if it would have any strange side-effects with the internals of the hook package.

jsnellbaker (approved):

dgirardi:
@snapwich there are various "before" hooks on requestBids, with many of them also falling back to the global `adUnits`, so I think to do what you say I'd need to add another before hook with higher priority than all of them. The main body of requestBids runs too late.

The high-priority hook was my first attempt, but I found it too hard to test, because a lot of "unit" tests already modify the global hook configuration.

PR #8644, opened 2022-07-05
PubxaiAnalyticsAdapter: Added extra fields to winning bid object

pnhegde:
- [ x] Refactoring (no functional changes, no api changes)

## Description of change
Added an additional field to capture more data about floors in the winning bid object.

### DIFF
--- a/modules/pubxaiAnalyticsAdapter.js
+++ b/modules/pubxaiAnalyticsAdapter.js
@@ -91,8 +91,12 @@ function mapBidResponse(bidResponse, status) {
     } else {
       Object.assign(bid, {
         bidId: bidResponse.requestId,
-        floorProvider: events.floorDetail ? events.floorDetail.floorProvider : null,
-        floorExtraData: events.floorDetail ? events.floorDetail : null,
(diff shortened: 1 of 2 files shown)


modules/pubxaiAnalyticsAdapter.js
```diff
@@ -92,6 +92,7 @@ function mapBidResponse(bidResponse, status) {
       Object.assign(bid, {
         bidId: bidResponse.requestId,
         floorProvider: events.floorDetail ? events.floorDetail.floorProvider : null,
+        floorExtraData: events.floorDetail ? events.floorDetail : null,
```
dgirardi, line 95: Could you be explicit in what data you are picking from `floorDetail`? the problem with just including it raw in your payload is that it forces Prebid to never change its structure from now on, regardless of what you are actually using from it in your backend.

pnhegde: @dgirardi Yeah makes sense. I have updated it to pick only the required fields. Thanks!

dgirardi (approved):

Issue #7968, opened 2022-01-25
Feature request: Topics api results on user.data


patmmccann:
## Type of issue
Feature request

## Description
FPD enrichment module should populate results on topics api ( see https://github.com/jkarlin/topics/issues ) onto user.data. 

Depends on a segtax being defined. see ( https://github.com/InteractiveAdvertisingBureau/openrtb/pull/81/files for an example and https://github.com/jkarlin/topics/issues/12 )

patmmccann:
I attempted to reserve 600 for this https://github.com/InteractiveAdvertisingBureau/openrtb/pull/86

patmmccann:
Copying some notes from the last taxonomy meeting: 

Publishers might not get all the topics; the browser might endeavor to hide many topics from the publisher but not from third parties. How would the do this? It seems they could isolate the frame that makes the call (but if they did this how would ssps correlate the topics to the bid request that is made from prebid?) or they could append headers to the bid request the publisher couldn't see (but then wouldnt the ssp just reflect the topics back at the publisher?) 

Still though, it seems useful to have a place for the topics the publisher does happen to get.

dmarti:
The Topics API proposal currently provides for up to five topics. But Topics API support in Prebid may need to handle more than five, since multiple parties might supply topics.

For example, a third-party caller could call Topics API, receive five topics for the current session, and reflect back to the publisher (as pointed out in @patmmccann 's comment ) and then the publisher might be able to supply additional topics for the same user, that had persisted in a first-party cookie or LocalStorage.

fbastello:
@dmarti - Prebid FPD module could handle this like this:
```
 {
   ...,
   "user": {
     "data": [
       {
         "name": "publisher.example",
         "ext": {
           "segtax": 600
         },
         "segment": [
           { "id": "243" },
           { "id": "247" }
         ]
       },
       {
         "name": "large-third-party.example",
         "ext": {
           "segtax": 600
         },
         "segment": [
           { "id": "59" },
           { "id": "129" },
           { "id": "173" }          
         ]
       }
       ]
     }
   }
 }
```
rf: https://github.com/patcg-individual-drafts/topics/issues/55#issuecomment-1112011239

patmmccann:
[...]

PR #8646, opened 2022-07-05
Topics FPD module: initial release

dgirardi:
## Type of change

- [x] Feature

## Description of change

This adds a new `topicsFpdModule` that queries the [topics api](https://developer.chrome.com/docs/privacy-sandbox/topics/) to populate `user.data`.

Since `document.browsingTopics()` always returns an empty list for me, I need volunteers to test this in the real world.

## Other information

Closes https://github.com/prebid/Prebid.js/issues/7968
Depends on https://github.com/prebid/Prebid.js/pull/8626
Documentation TBD


dgirardi:
Note that this affects anyone who's using FPD modules, by adding an additional promise resolution at the beginning of auctions - it was done with https://github.com/prebid/Prebid.js/pull/8626 in mind. IMO this should wait for that to get merged as to not needlessly affect the majority of users (who are not using topics).

modules/topicsFpdModule.js
```diff
@@ -0,0 +1,65 @@
...
+      const taxv = topic.taxonomyVersion;
+      if (!byTaxVersion.hasOwnProperty(taxv)) byTaxVersion[taxv] = [];
+      byTaxVersion[taxv].push(topic.topic);
+      return byTaxVersion;
+    }, {})
+  ).map(([taxv, topics]) => {
+    const datum = {
+      ext: {
+        segtax: TOPICS_TAXONOMY,
+        segclass: taxv
```
patmmccann, line 19: modelv

modules/topicsFpdModule.js
```diff
@@ -0,0 +1,65 @@
+import {logError, mergeDeep} from '../src/utils.js';
+import {getRefererInfo} from '../src/refererDetection.js';
+import {submodule} from '../src/hook.js';
+
+export const TOPICS_TAXONOMY = 600;
+
+export function getTopicsData(name, topics) {
+  return Object.entries(
+    topics.reduce((byTaxVersion, topic) => {
+      const taxv = topic.taxonomyVersion;
```
patmmccann, line 10: modelv = topic.modelVersion

modules/topicsFpdModule.js
```diff
@@ -0,0 +1,65 @@
+import {logError, mergeDeep} from '../src/utils.js';
+import {getRefererInfo} from '../src/refererDetection.js';
+import {submodule} from '../src/hook.js';
+
+export const TOPICS_TAXONOMY = 600;
```
patmmccann, line 5: this may change over time, tax 1 is 600, if we see future taxonomy versions beyond one, let's do formula 599 + taxonomy version. I'll reserve 601-609 with the IAB now

patmmccann: https://github.com/InteractiveAdvertisingBureau/openrtb/pull/109

dgirardi: Is there an agreed upon version scheme from topics? how do you propose this should be implemented? 

`taxonomyVersion` is a string that is apparently "1" atm, at some point it could come back as "1.1" "2" or anything, should the code just throw an error if it's not "1"? assign 601 hoping we can notice before the third one comes? Who and how is going to look for changes?

patmmccann: let's just assign 600 if it is "1" and abort otherwise. We can modify later if future returns get assigned future id's

modules/topicsFpdModule.js
```diff
@@ -0,0 +1,80 @@
...
+  } catch (e) {
+    logError('Could not call topics API', e);
+  }
+  if (topics == null) {
+    topics = GreedyPromise.resolve([]);
+  }
+  return topics;
+}
+
+const topicsData = getTopics().then((topics) => getTopicsData(getRefererInfo().domain, topics));
```
patmmccann, line 61: It seems we should potentially use the domain of the url in https://developer.mozilla.org/en-US/docs/Web/API/Document/currentScript 

Pubishers may wish to indicate the breadth of the footprint here, eg hearst.com instead of somemagazine.com or cafemedia.com instead of cafedelites.com

dgirardi: I'm not sure I understand - do you mean that topics will return different results if you use `<script src="//rubicon.com/prebid.js" />` compared `<script src="//appnexus.com/prebid.js" />`?  that seems insane.

patmmccann: That's how it is described

patmmccann: So it appears there's been some clarification:

"A design goal of the Topics API is to enable interest-based advertising without the sharing of information to more entities than is currently possible with third-party cookies. The Topics API proposes that topics can only be returned for API callers that have already observed them, within a limited timeframe.

Key Term
A Topics API caller is the entity that calls the document.browsingTopics() JavaScript method, and will use the topics returned by the method to help select relevant ads. Typically, a call to document.browsingTopics() would be from code included in a site from a third party such as an adtech platform. The browser determines the caller from the site of the current document. So, if you're a third party on a page, make sure you call the API from an iframe that your site owns."

This suggests only if you call the api from inside a frame and do the post-message will you get different behavior

Let's keep that for a follow up

patmmccann: @dgirardi testing successful

patmmccann (approved):

mikemangas:
Is there a documentation for the usage of this module?

PR #8652, opened 2022-07-06
TheMediaGrid: added ortb2.user.data support to gridNMBidAdapter

TheMediaGrid:
## Type of change

- [x] Feature

## Description of change

Added support of ortb2.user.data and ortb2.site.content.data

- grid-tech@themediagrid.com

### DIFF
--- a/modules/gridNMBidAdapter.js
+++ b/modules/gridNMBidAdapter.js
@@ -1,4 +1,13 @@
-import { isStr, deepAccess, isArray, isNumber, logError, logWarn, parseGPTSingleSizeArrayToRtbSize } from '../src/utils.js';
+import {
+  isStr,
+  deepAccess,
+  isArray,
+  isNumber,
+  logError,
+  logWarn,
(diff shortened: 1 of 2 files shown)


modules/gridNMBidAdapter.js
```diff
@@ -174,9 +181,27 @@ export const spec = {
         user.ext = userExt;
       }
 
+      const ortb2UserData = deepAccess(bidderRequest, 'ortb2.user.data');
```
dgirardi, line 184: You should, at least in intent, be able to just do `request = mergeDeep({}, bidderRequest.ortb2, request)` instead of merging it piece by piece - if you are intentionally avoiding that I'm interested in understanding why. LGTM otherwise.

dgirardi (approved):

modules/gridNMBidAdapter.js
ChrisHuie: @TheMediaGrid (just tagging for the question above)

Issue #8565, opened 2022-06-15
Price Floors module rejects valid bids ?


matthieularere-msq:
## Type of issue
bug

## Description
In prebid's console log lines I can see bid which are rejected by the price floor module but as far as I understand they are valid bids. For example see attached screenshot, where floorValue 0.25 is set in currency EUR, a 0.2886 USD bid is received, converted to 0.2761 EUR but is rejected.

## Steps to reproduce
I noticed this behavior on this page https://www.marmiton.org/?pbjs_debug=true after a few refreshs.

Is there something I missed with the reason why this bid was rejected ?

Thanks.

jlquaccia:
Starting to look into and work on this ticket today

matthieularere-msq:
Hi @dgirardi indeed that an example of rejected bids I see and where I don't understand why they are considering their cpm is higher than the floor value.

dgirardi:
@matthieularere-msq because the CPM is being adjusted lower by your (or your colleagues') `bidderSettings.standard.bidCpmAdjustment`.

dgirardi:
Indeed it should - and I missed that - it looks like it may be adjusted twice. Looking into this.

patmmccann:
@robertrmartinez could you weigh in? Do you think this can be addressed with documentation?

dgirardi:
Here's the confirmation - this is a problem about the meaning of `cpm` / `originalCpm` rather than price floors.
This bid:
![AdagioRejected](https://user-images.githubusercontent.com/10151381/174319819-e095150e-393a-44dd-a738-1830f93c4a8e.png)


Corresponds to this response from Adagio:

```
{
  "id": "904981a1-3fc3-43de-8e93-33dfe418d28b",
  "bids": [
    {
      "requestId": "20dcc4233181c8e",
      "creativeId": "392890663",
      "cpm": 1.25314785,
      "currency": "USD",
[...]

PR #8655, opened 2022-07-06
PBjs Core Price Floors Module: improve logging on bid rejections to clarify which CPM is being compared with which floor

dgirardi:
## Type of change

- [x] Other

## Description of change

When a bid is rejected because below its price floor, the logged bid object shows a `cpm` _before_ `bidCpmAdjustment`, but the comparison to the floor may have taken adjustment into account. This can appear as if a valid bid is being rejected (see https://github.com/prebid/Prebid.js/issues/8565)

This modifies the log message to explicitly show which cpm was found to be below the floor, which should help avoid confusion.

### DIFF
--- a/modules/priceFloors.js
+++ b/modules/priceFloors.js
@@ -751,14 +751,7 @@ export function addBidResponseHook(fn, adUnitCode, bid) {
     flooredBid.status = CONSTANTS.BID_STATUS.BID_REJECTED;
     // if floor not met update bid with 0 cpm so it is not included downstream and marked as no-bid
     flooredBid.cpm = 0;
-    const floorMsg = (() => {
-      try {
-        return ` (adjusted cpm: ${bid.floorData.cpmAfterAdjustments.toFixed(2)}, floor: ${floorInfo.matchingFloor.toFixed(2)})`
-      } catch (e) {
-        return ''
-      }
-    })();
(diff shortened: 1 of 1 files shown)


robertrmartinez (approved):

modules/priceFloors.js
```diff
@@ -751,7 +751,14 @@ export function addBidResponseHook(fn, adUnitCode, bid) {
     flooredBid.status = CONSTANTS.BID_STATUS.BID_REJECTED;
     // if floor not met update bid with 0 cpm so it is not included downstream and marked as no-bid
     flooredBid.cpm = 0;
-    logWarn(`${MODULE_NAME}: ${flooredBid.bidderCode}'s Bid Response for ${adUnitCode} was rejected due to floor not met`, bid);
+    const floorMsg = (() => {
+      try {
+        return ` (adjusted cpm: ${bid.floorData.cpmAfterAdjustments.toFixed(2)}, floor: ${floorInfo.matchingFloor.toFixed(2)})`
```
patmmccann, line 756: I have a feeling the lack of precision is going to generate more bug reports when the rounded values are equal; how about toFixed(4)?

patmmccann (dismissed):

modules/priceFloors.js
dgirardi: how about keeping the full precision then? (because if 2 digits are not enough why would 4 be?)

ChrisHuie (approved):

PR #8659, opened 2022-07-07
Release Drafter: change default tagging to maintenance

ChrisHuie:
Update the release drafter so that if a pr is not tagged it defaults to maintenance

### DIFF
--- a/.github/release-drafter.yml
+++ b/.github/release-drafter.yml
@@ -10,7 +10,8 @@ categories:
     labels:
       - 'fix'
       - 'bugfix'
-      - 'bug'      
+      - 'bug'
+    default: maintenance
 change-template: '- $TITLE (#$NUMBER)'
 version-resolver:
   major:


patmmccann (approved):

patmmccann:
Thanks Chris!

jsnellbaker (commented):
Just to double-check, I tried to look at the release drafter docs to confirm if the value for the default here would need quotes surrounding it (since the labels are assigned a string, and their id is not an object category - like how it's structured in the version section).  However I didn't see a direct example or reference to having a default flag under the `category` section.  I found some information on auto labeling (which is similar to what we may want here), but I'm not sure if we could make that work given the random naming schema used in pull requests/branches/etc.

Have you tested this change in a fork somewhere where the release notes get generated automatically?  If you did, just let me know (or if there was some part of the docs I missed).

Thanks.

ChrisHuie:
adding to this pr and making it more robust after reviewing more of the docs around auto-tagging and re-titling prs

Issue #8651, opened 2022-07-06
"userSync" module initially calls "cookiesAreEnabled" 


renebaudisch:
## Type of issue
Privacy - GDPR - Purpose One

## Description
"userSync" module initially calls "cookiesAreEnabled" from the sessionManager and is writing and reading the device storage regardless of consent.

## Steps to reproduce
I prepared a testpage where you can see that before the cmp is shown the LS get's written with cookieTest:
https://reports.asadcdn.com/demopages/2022/prebid.html

If it is too fast, set a breakpoint here:
<img width="460" alt="image" src="https://user-images.githubusercontent.com/15980546/174813887-936a46f0-6420-4579-822d-7bdbc3b8e3db.png">

## Test page
https://reports.asadcdn.com/demopages/2022/prebid.html

### Expected results
Not accessing cookies/LS before user interacted with cmp

### Actual results
Writes test value cookieTest regardless to CMP.

## Platform details
latest pbjs all platforms

## Other information
For more information please read:
https://github.com/prebid/Prebid.js/pull/8410
https://github.com/prebid/Prebid.js/pull/8613

dgirardi:
What is the desired behavior - storage should not be touched at all when there is not, or there might not be consent for it? (userId is not the only offender in that case)

Does it matter that a temporary "probe" cookie like this one is not personally identifiable data, and is to the best of my knowledge not protected by GPDR? are you using some auditing tool that is offended by this usage?

renebaudisch:
Our company lawyers have been contacted because of this cookieTest and they therefore pleased us to get rid of it. So at least in Germany there seems to be the opinion in law that these probes also are protected by GDPR, as it is a third party ad technology and by that term not necessary to operate the page and so has to have device access consent aka purpose one, but I'm not a lawyer. 

But it doesn't also seem logical to check for LS if the user disallows the usage of it.
Even if this would be fine by law, all further access would be protected by GDPR and so even it is accessible by test the vendors may not, so why check at all in first? Why not only check when needed to avoid being affected by a law?

patmmccann:
category translation, brand exclusion, and fpd enrichment modules have similar behavior. We'll want to put this interpretation of the law [remove device access for storage items critical to basic function before consent] behind config

dgirardi:
[...]

PR #8661, opened 2022-07-07
Core & multiple modules: strict purpose 1 consent option; do not require vendor consent for "core" storage / ID modules

dgirardi:
## Type of change

- [x] Bugfix

## Description of change

This change:

 - makes it so that `setConfig({deviceAccess: false})` always turns off _all_ storage, including "core" storage (which currently ignores that flag, contrary to [what the documentation says](https://docs.prebid.org/dev-docs/publisher-api-reference/setConfig.html#setConfig-deviceAccess)
 - introduces "vendorless" GDPR enforcement for pubCommonId/sharedId (same intent as in https://github.com/prebid/Prebid.js/pull/8448, which was later reverted)
 - introduces a new config option, `setConfig({consentManagement: {strictStorageEnforcement: true}})`, that - if set - forces _all_ storage, including "core" storage, to respect consent; core storage would follow the same rule as "vendorless" modules (just generic P1 consent required), the rest would continue to respect both purpose and vendor as it does now
 - refactors `userSync.js` and the userId module to access storage only once consent data is available, so that the above can work
 - attempts to always enforce GDPR if the consentManagement module is installed (to try and catch gdpr-protected operations that happen before consent data is resolved). 

As mentioned in https://github.com/prebid/Prebid.js/issues/8651, there are a few other consumers of `getCoreStorageManager` - I only checked that `fpdEnrichment` would fail gracefully without storage. The rest (categoryTranslation and adpod.brandCategoryExclusion) may still be accessing storage before consent is resolved.

## Other information

Closes https://github.com/prebid/Prebid.js/issues/8651
Documentation: https://github.com/prebid/prebid.github.io/pull/3889


modules/gdprEnforcement.js
```diff
@@ -12,6 +12,12 @@ import {validateStorageEnforcement} from '../src/storageManager.js';
 import * as events from '../src/events.js';
 import CONSTANTS from '../src/constants.json';
 
+// modules for which vendor consent is not needed (see https://github.com/prebid/Prebid.js/issues/8161)
+const VENDORLESS_MODULES = new Set([
+  'sharedId',
+  'pubCommonId'
```
patmmccann, line 18: https://github.com/prebid/Prebid.js/blob/master/modules/pubProvidedIdSystem.js also doesnt need consent but doesn't use device storage. Should it be on this list?

dgirardi:
@patmmccann in drafting the docs for this, I noticed that what we say now:

> You can prevent Prebid.js from reading or writing cookies or HTML localstorage by setting [`deviceAccess: false`]

Is not true; "core" storage, used as outlined above, still has access to the device regardless of that flag. Was this an oversight - should we fix it, or still require "true" disablers to also set `alwaysEnforceDeviceAccess`? (it could, conceivably, break users of category exclusion / translation, and change the domain set by fpdEnrichment).

dgirardi:
After discussing with @patmmccann  and @bretg , we agreed on the following:

 1. a vendor (bid adapter, module) wants to read/store something. In GDPR scope, requires general P1 consent and vendor-specific consent. Requires deviceAccess flag. bid adapters also require bidderSettings.storeagedAllowed. (no changes)
 2. a "vendorless" ID module wants to read/store something.  In GDPR scope, require general P1 consent. Require deviceAccess flag. (no changes)
 3. a "PBJS core" service or "vendorless module" wants to read/store something.  Require deviceAccess flag (change! may break domain calculation) If `consentManagement.strictStorageEnforcement` is true, and in GDPR scope, requires general P1 consent (change, but safe because it's behind a new config flag)

the change for case 3 should be documented (coming soon!)

patmmccann:
@dgirardi was linking 8651 an accident or does this close both 8647 and 8651

modules/gdprEnforcement.js
patmmccann: the second one is definitely redundant and removing it is in the 8.0 proposal. PPID module is different, it allows a publisher to set any arbitrary id and also say what it is. It  isn't very popular,  but it would allow a publisher to say pass hash(login) or some other identifier that was quite different than the one used by sharedid / pubcommonid

jsnellbaker (approved):

osazos (approved):
LGTM!

PR #8580, opened 2022-06-17
Concert Bid Adapter: adding referrer

andrew-a-dev:
## Type of change
- [x] Other

Specifically: adding functionality to the existing Concert adapter

## Description of change

Adding a `ref` parameter to the bid calls, using the various "referrer" properties available in Prebid. 

(The PR also contains linting-related changes to the file as a whole, to do with `;` and whitespace.)

## Note

This PR was originally opened (and approved) here: https://github.com/prebid/Prebid.js/pull/8564 -- but I unintentionally closed the PR by messing around with `master` on our forked repo. 🤦 Sorry. (It also includes an adjustment to that version, as per a suggestion from @dgirardi 🙌 )

PR #8676, opened 2022-07-12
Concert Bid Adapter: adding referer (Resubmit of Pr #8580)

ChrisHuie:
#8580 CircleCI tests didn't kickoff so resubmitted pr under my account to keep this code moving.

### DIFF
--- a/modules/concertBidAdapter.js
+++ b/modules/concertBidAdapter.js
@@ -1,6 +1,7 @@
 import { logWarn, logMessage, debugTurnedOn, generateUUID } from '../src/utils.js';
 import { registerBidder } from '../src/adapters/bidderFactory.js';
 import { getStorageManager } from '../src/storageManager.js';
+import { hasPurpose1Consent } from '../src/utils/gpdr.js';
 
 const BIDDER_CODE = 'concert';
 const CONCERT_ENDPOINT = 'https://bids.concert.io';
@@ -207,5 +208,11 @@ function consentAllowsPpid(bidderRequest) {
   /* NOTE: We can't easily test GDPR consent, without the
    * `consent-string` npm module; so will have to rely on that
    * happening on the bid-server. */
-  return !(bidderRequest.uspConsent === 'string' && bidderRequest.uspConsent.toUpperCase().substring(0, 2) === '1YY');
+  const uspConsent = !(bidderRequest?.uspConsent === 'string' &&
+    bidderRequest?.uspConsent[0] === '1' &&
+    bidderRequest?.uspConsent[2].toUpperCase() === 'Y');
+
+  const gdprConsent = bidderRequest?.gdprConsent && hasPurpose1Consent(bidderRequest?.gdprConsent);
+
+  return (uspConsent || gdprConsent);
 }
--- a/modules/concertBidAdapter.js
+++ b/modules/concertBidAdapter.js
@@ -1,6 +1,6 @@
 import { logWarn, logMessage, debugTurnedOn, generateUUID } from '../src/utils.js';
 import { registerBidder } from '../src/adapters/bidderFactory.js';
-import { getStorageManager } from '../src/storageManager.js'
+import { getStorageManager } from '../src/storageManager.js';
 
 const BIDDER_CODE = 'concert';
 const CONCERT_ENDPOINT = 'https://bids.concert.io';
@@ -45,7 +45,7 @@ export const spec = {
         uspConsent: bidderRequest.uspConsent,
         gdprConsent: bidderRequest.gdprConsent
       }
-    }
+    };
 
     payload.slots = validBidRequests.map(bidRequest => {
       let slot = {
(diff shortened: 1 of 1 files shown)


ChrisHuie:
@andrew-a-dev resubmitted this pr under my account to hopefully get this merged in this week.

modules/concertBidAdapter.js
```diff
@@ -203,9 +204,8 @@ function hasOptedOutOfPersonalization() {
  * @param {BidderRequest} bidderRequest Object which contains any data consent signals
  */
 function consentAllowsPpid(bidderRequest) {
-  /* NOTE: We cannot easily test GDPR consent, without the
+  /* NOTE: We can't easily test GDPR consent, without the
    * `consent-string` npm module; so will have to rely on that
    * happening on the bid-server. */
-  return !(bidderRequest.uspConsent === 'string' &&
-           bidderRequest.uspConsent.toUpperCase().substring(0, 2) === '1YY')
+  return !(bidderRequest.uspConsent === 'string' && bidderRequest.uspConsent.toUpperCase().substring(0, 2) === '1YY');
```
patmmccann, line 210: why are you checking for the second character https://github.com/InteractiveAdvertisingBureau/USPrivacy/blob/f8d659f58665b4e797b99b59079fd787a386eaaa/CCPA/US%20Privacy%20String.md?plain=1#L123

modules/concertBidAdapter.js
```diff
@@ -203,9 +204,8 @@ function hasOptedOutOfPersonalization() {
  * @param {BidderRequest} bidderRequest Object which contains any data consent signals
  */
 function consentAllowsPpid(bidderRequest) {
-  /* NOTE: We cannot easily test GDPR consent, without the
+  /* NOTE: We can't easily test GDPR consent, without the
```
patmmccann, line 207: what does this comment mean? why can't you look at the tcf consent string without this npm module?

andrew-a-dev:
> @andrew-a-dev resubmitted this pr under my account to hopefully get this merged in this week.

Thank you, @ChrisHuie ! 🙌

modules/concertBidAdapter.js
andrew-a-dev: This is checking the first 3 characters of the string; we do _not_ mark as consenting to PPID usage unless: (1) USPrivacy string _version_ matches what we expect, and (2) both `Notice to Opt-out` and `Opt-out Sale` are marked as consenting. (And we don't care about LSPA.)

modules/concertBidAdapter.js
andrew-a-dev: We can _view_ the consent string, yes, but if _parsing_ it is non-trivial enough that there exists an official, specialised, and sanctioned `npm` module to do it, it does not seem appropriate that we would try to add that functionality ourselves to our bid adapter; especially since we do that exact step (using `consent-string`) in our bid server.

(For the record, this change only exists in this PR because it was a `NOP` commit, to try to activate the CircleCI on a fresh push.)

modules/concertBidAdapter.js
patmmccann: If the string is '1NY', the user is also opted out

andrew-a-dev: If I understand what you're saying, we should consider the user "opted-out" if the CCPA string starts `1NY`? If so, I suspect we didn't realise that was a possible value (i.e. can they have opted-out if they weren't notified of the opt-out option?). I guess we assumed that a `Y` in the third character implied a `Y` in the second character. If that's not the case, I'll happily update the PR to ignore the second character then. (And also, your original comment makes more sense to me now; thank you for that.)

modules/concertBidAdapter.js
patmmccann: we provide a utility function, eg https://github.com/prebid/Prebid.js/blob/0be4560e182e46c121c283f763cdef74b5ff9db9/modules/appnexusBidAdapter.js#L559

modules/concertBidAdapter.js
patmmccann: some people have started pushing '1NY' if the GPC headers in the browser indicate opt out.

modules/concertBidAdapter.js
andrew-a-dev: I appreciate this, @patmmccann , thank you! (I don't think those functions existed when we first wrote our adapter, so it's good to know that so we can update it.) I've updated our use of consent strings, as per these comments, and will ask Chris if they can merge these changes to this branch (which I can't do). 🙌

andrew-a-dev:
@ChrisHuie Thanks again for creating this PR, would you be able to pull in from this commit? https://github.com/voxmedia/Prebid.js/commit/c48c1d470b3990a276044fb2d2c65ebfec2dbc75 ; this includes further changes to the consent-string handling, as per the feedback/suggestions above. Many thanks, again! 🙌

ChrisHuie:
> @ChrisHuie Thanks again for creating this PR, would you be able to pull in from this commit? [voxmedia@c48c1d4](https://github.com/voxmedia/Prebid.js/commit/c48c1d470b3990a276044fb2d2c65ebfec2dbc75) ; this includes further changes to the consent-string handling, as per the feedback/suggestions above. Many thanks, again! 🙌

Sure can 👍 Will update shortly

patmmccann (approved):

PR #8677, opened 2022-07-12
Adding cb to the callback

ftxmoJason:
## Type of change

- [X] Other: added parameter to the callback in getId so that the pbjs get ID methods work

## Description of change
added parameter to the callback in getId so that the pbjs get ID methods work


ftxmoJason:
The package-lock is causing issues, going to create this PR over again

Issue #8429, opened 2022-05-18
Return multiple IDs from same userId submodule


Nick-Merkle:
## Overview

We are looking for a way to have our User ID module return different IDs per SSP to consume. The current configuration appears to be limited to a single ID to use by anyone. For our use case each SSP we work with has their own ID to identify users so having a single ID for anyone to use is not possible.

## Issue

Each userId submodule is limited to a single configuration, but we are looking for a way to support multiple configurations within our submodule.

## Use Case

A publisher would configure to use like this:

```
pbjs.setConfig({
        "debug": true,
        "userSync": {
          "userIds": [
            {
              "name": "merkleId",
              "params": {
                "sv_pubid":  "12562",
                "ssp_ids": ["527404512","534404531"]
              },
              "storage": {
                "type": "html5",
                "name": "merkleIds",
                "expires": 30
              }
            }
          ],
          "syncDelay": 5000,
          "auctionDelay": 1000
        }
      });
```

Note: `params.ssp_ids` would correspond to their own source and id.

## Example EID data

The sub-module would construct data similar to the example below. We constructed this data similar to the examples used by `pubProvidedId`, but it can be modified to fit into something more universally acceptable for Prebid.js.

```
{
    "user": {
      "ext": {
          "eids": [
              {
                  "source": "ssp1.merkleinc.com",
                  "uids": [
                      {
                          "atype": 3,
                          "id": "unique-id-for-ssp1",
                          "ext": {
                              "enc": 1,
                              "keyId": 16,
                              "idName": "pamId"
                          }
                      }
                  ]
              },
              {
                  "source": "ssp2.merkleinc.com",
                  "uids": [
                      {
                          "atype": 3,
                          "id": "unique-id-for-spp2",
                          "ext": {
                              "enc": 1,
                              "idName": "pamId",
                              "third": 4
                          }
                      }
                  ]
              }
          ]
      }
  }
}
```

## Possible Solution

The example below illustrates one possible way to handle multiple source configs.

src/modules/userId/eids.js
```
export function createEidsArray(bidRequestUserId) {
  let eids = [];

  for (const subModuleKey in bidRequestUserId) {
    if (bidRequestUserId.hasOwnProperty(subModuleKey)) {
      if (subModuleKey === 'pubProvidedId') {
        eids = eids.concat(bidRequestUserId['pubProvidedId']);
      } 

      // ************************************************************

      // @NOTE: A more generic flag is required without hard coding subModuleKey
[...]

PR #8678, opened 2022-07-12
Ftrack Id Module: adding parameter to callback in getId return object

ftxmoJason:
## Type of change
- [X] Other

## Description of change
1. Adding a `cb` parameter to the callback property in the object returned by getId() so that more pbjs get ID methods can be used
2. Tweaking the eids.js so that a value is returned. We tried to return our IDs as an object but had to JSON.stringify() it.

### DIFF
--- a/modules/userId/eids.js
+++ b/modules/userId/eids.js
@@ -68,7 +68,20 @@ export const USER_IDS_CONFIG = {
     source: 'flashtalking.com',
     atype: 1,
     getValue: function(data) {
-      return data.uid
+      
+      var returnObject = {};
+
+      if (data.hasOwnProperty('DeviceID')) {
+        returnObject.DeviceID = data.DeviceID;
+      }
+      if (data.hasOwnProperty('HHID')) {
+        returnObject.HHID = data.HHID;
+      }
+      if (data.hasOwnProperty('SingleDeviceID')) {
+        returnObject.SingleDeviceID = data.SingleDeviceID;
+      }
+
+      return JSON.stringify(returnObject);
     },
     getUidExt: function(data) {
       if (data.ext) {
--- a/modules/userId/eids.js
+++ b/modules/userId/eids.js
@@ -370,13 +370,16 @@ export function createEidsArray(bidRequestUserId) {
         eids = eids.concat(bidRequestUserId['pubProvidedId']);
       } else if (subModuleKey === 'ftrackId') {
         // ftrack has multiple IDs so we add each one that exists
-        for (let id in bidRequestUserId[subModuleKey]) {
-          eids.push({
-            'atype': 1,
-            'ext': id,
-            'id': bidRequestUserId[subModuleKey][id]
-          });
+        let eid = {
+          'atype': 1,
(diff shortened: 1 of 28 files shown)


ftxmoJason:
To build on @TheMediaGrid 's comment. Our IDs are formatted like so:
```
{
    "HHID": ["<USERS HH ID>"],
    "DeviceID": ["<USERS DEVICE ID>"],
    "SingleDeviceID": ["USERS SINGLE DEVICE ID"]
}
```
And we can't return an object from eids.js `getValue`, so we `JSON.stringify()` the object and return it. Is that the best approach?

ftxmoJason:
@ftxmojared Just trying to ping you in this thread for awareness.

patmmccann:
have you studied this https://github.com/prebid/Prebid.js/issues/8429

ftxmoJason:
Thanks @patmmccann for the suggestion.
I tweaked the `createEidsArray` method and added a condition for 'ftrack', similar to how 'pubProvidedId' did it.

modules/userId/eids.js
```diff
@@ -367,6 +368,18 @@ export function createEidsArray(bidRequestUserId) {
...
+        let eid = {
+          'atype': 1,
+          'id': (bidRequestUserId[subModuleKey]['DeviceID'] || []).join('|'),
+          'ext': {}
+        }
+        for (let id in bidRequestUserId[subModuleKey]) {
+          eid.ext[id] = (bidRequestUserId[subModuleKey][id] || []).join('|');
+        }
+
+        eids.push(eid);
```
smenzer, line 382: this block needs tests cases, please

smenzer (changes requested):
i did an initial pass through and first thing we need are test cases for the new functionality. also, are there any changes to publisher implementation of the config or does this only impact those reading the EIDs array?

modules/userId/eids.js
TheMediaGrid: no changes to the publisher config implementation, just the eids object interaction is changed
I will also let @ftxmoJason comment on the test cases

TheMediaGrid:
@smenzer can you please take a look at the updates here? we'd really like this PR to make it to the next release if possible

smenzer (changes requested):

test/spec/modules/ftrackIdSystem_spec.js
```diff
@@ -298,4 +307,88 @@ describe('FTRACK ID System', () => {
...
+        userSync: {
+          auctionDelay: 10,
+          userIds: [{
+            name: 'ftrack', value: ids,
+          }]
+        }
+      });
+
+      getGlobal().getUserIdsAsync().then((ids) => {
+        expect(getGlobal().getUserIdsAsEids()).to.deep.equal(createEidsArray(ids));
```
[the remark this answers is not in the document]
smenzer: thanks for the updates

smenzer (approved):
LGTM

This probably warrants an update to the prebid docs repository, but I'll go ahead and approve this now since this is good to go.

PR #8679, opened 2022-07-12
Ttd Bid adapter: support for transaction id, bcat, and consolidate params

julian-burger-ttd:
## Type of change

- [X] Feature
- [X ] Does this change affect user-facing APIs or examples documented on http://prebid.org?

## Description of change
TTD Adapter Updates

Remove siteId parameter
Make placementId parameter optional if GPID is passed
Pass transactionId through to source.tid
Read bcat from ortb2

### DIFF
--- a/modules/ttdBidAdapter.js
+++ b/modules/ttdBidAdapter.js
@@ -326,6 +326,7 @@ export const spec = {
 
     const gpid = utils.deepAccess(bid, 'ortb2Imp.ext.gpid');
     if (!bid.params.placementId && !gpid) {
+      utils.logWarn(BIDDER_CODE + ': one of params.placementId or gpid (via the GPT module https://docs.prebid.org/dev-docs/modules/gpt-pre-auction.html) must be passed');
       return false;
     }
 
--- a/modules/ttdBidAdapter.js
+++ b/modules/ttdBidAdapter.js
@@ -4,7 +4,7 @@ import { createEidsArray } from './userId/eids.js';
 import { registerBidder } from '../src/adapters/bidderFactory.js';
 import { BANNER, VIDEO } from '../src/mediaTypes.js';
 
-const BIDADAPTERVERSION = 'TTD-PREBID-2022.02.18';
+const BIDADAPTERVERSION = 'TTD-PREBID-2022.06.28';
 const BIDDER_CODE = 'ttd';
 const BIDDER_CODE_LONG = 'thetradedesk';
 const BIDDER_ENDPOINT = 'https://direct.adsrvr.org/bid/bidder/';
@@ -58,7 +58,9 @@ function getBidFloor(bid) {
 }
 
 function getSource(validBidRequests) {
-  let source = {};
+  let source = {
+    tid: validBidRequests[0].transactionId
+  };
   if (validBidRequests[0].schain) {
     utils.deepSetValue(source, 'ext.schain', validBidRequests[0].schain);
   }
@@ -124,7 +126,6 @@ function getUser(bidderRequest) {
 
 function getSite(bidderRequest, firstPartyData) {
   var site = {
-    id: utils.deepAccess(bidderRequest, 'bids.0.params.siteId'),
     page: utils.deepAccess(bidderRequest, 'refererInfo.page'),
     publisher: {
       id: utils.deepAccess(bidderRequest, 'bids.0.params.publisherId'),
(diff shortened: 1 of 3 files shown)


modules/ttdBidAdapter.js
```diff
@@ -58,7 +58,9 @@ function getBidFloor(bid) {
 }
 
 function getSource(validBidRequests) {
-  let source = {};
+  let source = {
+    tid: validBidRequests[0].transactionId
```
patmmccann, line 62: @JudSpencer mentioned in ortb slack he would prefer if you guys followed Rob's proposal and implementation here https://github.com/prebid/Prebid.js/pull/8641/files to gather imp.ext.tid

modules/ttdBidAdapter.js
```diff
@@ -387,6 +382,10 @@ export const spec = {
       ext: getExt(firstPartyData)
     }
 
+    if (firstPartyData && firstPartyData.bcat) {
+      topLevel.bcat = firstPartyData.bcat;
```
patmmccann, line 386: can you document if this overwrites or appends existing settings for an account on the backend? I think append is not particularly workable, and overwrite is preferred.

Why not battr and badomain as well?

modules/ttdBidAdapter.js
patmmccann: adding to this, source.tid as the first transaction id makes sense to me, but we also seem to have debate over whether it should be the auction id, and implementors have chosen one or the other. Prebid Server seems to just make up its own id. So aside from the cardinality issue that imp.ext.tid addresses, we have other issues with adapters defining it differently (we have open issues discussing). Imp.ext.tid shares none of these concerns, only adoption, hence me encouraging you to pick it up here.

modules/ttdBidAdapter.js
patmmccann: Sorry, badv

markd-fs: We would also like to see support for battr and badv here.

julian-burger-ttd: I'll add battr/badv. We're still discussing behavior internally.

patmmccann: thanks!

julian-burger-ttd: We decided to make the blocking additive (docs will reflect this).

patmmccann: Can we discuss? This decision breaks all utility of this feature

patmmccann: @julian-burger-ttd can you confirm the badv battr changes above are planned for this pr?

julian-burger-ttd: Confirming for this PR. Haven't had a chance to get to it yet, though. As for overwrite vs. additive we having some internal discussion as there are multiple use cases to consider.

modules/ttdBidAdapter.js
julian-burger-ttd: I updated this but still new to the area so please verify it is what you expected.

modules/ttdBidAdapter.js
julian-burger-ttd: Updated the code. Docs updates (separate MR anyway) will follow.

modules/ttdBidAdapter.js
```diff
@@ -58,7 +58,9 @@ function getBidFloor(bid) {
 }
 
 function getSource(validBidRequests) {
-  let source = {};
+  let source = {
+    tid: utils.deepAccess(validBidRequests[0], 'ortb2Imp.ext.tid')
```
patmmccann, line 62: you had source.tid fine; but what I was imagining is that you would want something like 

https://github.com/prebid/Prebid.js/blob/d46661a4aa8994d57f3bea3055bd4e4c8aa50e10/modules/ixBidAdapter.js#L248 / https://github.com/prebid/Prebid.js/blob/d46661a4aa8994d57f3bea3055bd4e4c8aa50e10/modules/ixBidAdapter.js#L361

as you guys are asking other exchanges for that in other forums

modules/ttdBidAdapter.js
```diff
@@ -141,17 +142,21 @@ function getSite(bidderRequest, firstPartyData) {
 
 function getImpression(bidRequest) {
   let impression = {
-    id: bidRequest.bidId,
-    tagid: bidRequest.params.placementId
+    id: bidRequest.bidId
   };
 
-  let gpid = utils.deepAccess(bidRequest, 'ortb2Imp.ext.gpid');
+  const gpid = utils.deepAccess(bidRequest, 'ortb2Imp.ext.gpid');
```
patmmccann, line 148: right here is where you could also get ext.tid

modules/ttdBidAdapter.js
patmmccann: I commented below on where it would go

patmmccann (changes requested):
comments on imp.ext.tid vs source.tid

modules/ttdBidAdapter.js
julian-burger-ttd: Thanks I think I understand what you are getting at now.

patmmccann:
Lgtm

patmmccann (approved):

PR #8694, opened 2022-07-15
AdOcean Bid Adapter: send ids which are taking part in auction.

mmuras:
## Type of change

- [x] Bugfix

Change AdOcean bid adapter so ids which are taking part in auction will be sent in request for performance reasons.


mmuras:
I'm not sure if I understand how would we use that because communication between adapter and adserver isn't using OpenRTB protocol (adserver doesn't support it).

patmmccann (approved):

PR #8696, opened 2022-07-18
PBS Bid Adapter: Setting bidderCode and adapterCode for a bid

pm-azhar-mulla:
## Type of change

- [x] Bugfix

## Description of change

We encountered a warning message on the console, similar to [PrebidJS bug](https://github.com/prebid/Prebid.js/pull/8554) when an alternate bidder wins the auction.
Setting the bidderCode and adapterCode of a receiving bid to handle it.


modules/prebidServerBidAdapter/index.js
```diff
@@ -984,6 +984,12 @@ Object.assign(ORTB2.prototype, {
           });
           bidObject.requestTimestamp = this.requestTimestamp;
           bidObject.cpm = cpm;
+          bidObject.bidderCode = seatbid.seat;
+          if (bidRequest?.bidder) {
```
bretg, line 988: The intention is that bidObject.adapterCode is supposed to be defined by bid.ext.prebid.meta.adaptercode as a first priority, then default to seatbid.seat.

bretg (changes requested):
We're going to want test cases on the various scenarios here. Those are documented in https://github.com/prebid/Prebid.js/issues/8129

modules/prebidServerBidAdapter/index.js
pm-azhar-mulla: Hi @bretg ,
What you are pointing out is bidderCode, which should be similar to seatbid.seat. 
I have made changes to consume adapterCode from bid.ext.prebid.meta.adaptercode as first priority, then bidRequest.bidder and then seatbid.seat

modules/prebidServerBidAdapter/index.js
```diff
@@ -984,6 +984,14 @@ Object.assign(ORTB2.prototype, {
           });
           bidObject.requestTimestamp = this.requestTimestamp;
           bidObject.cpm = cpm;
+          bidObject.bidderCode = seatbid.seat;
```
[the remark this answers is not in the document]
pm-azhar-mulla: Expectations, for now, are only to handle warning messages on the console as seen in [PrebidJS  side bug](https://github.com/prebid/Prebid.js/pull/8554)
Which will be handled by setting adapterCode for the bid.

modules/prebidServerBidAdapter/index.js
bretg: this is better @pm-azhar-mulla , but I think this deserves a unit test or 6.

bretg (approved):

test/spec/modules/prebidServerBidAdapter_spec.js
```diff
@@ -2692,6 +2692,26 @@ describe('S2S Adapter', function () {
...
+
+    it('setting adapterCode for alternate bidder', function () {
+      config.setConfig({ CONFIG });
+      RESPONSE_OPENRTB.seatbid[0].bid[0].ext.prebid.meta.adaptercode = 'appnexus2'
+      adapter.callBids(REQUEST, BID_REQUESTS, addBidResponse, done, ajax);
+      server.requests[0].respond(200, {}, JSON.stringify(RESPONSE_OPENRTB));
+
+      const response = addBidResponse.firstCall.args[1];
+      expect(response).to.have.property('adapterCode', 'appnexus2');
+      delete RESPONSE_OPENRTB.seatbid[0].bid[0].ext.prebid.meta.adaptercode;
```
dgirardi, line 2712: minor point - it's better to deepClone the object, or put this in a finally block. Otherwise, if any future change causes this test to fail, the teardown logic will not run likely causing a cascade of confounding failures.

dgirardi (approved):

test/spec/modules/prebidServerBidAdapter_spec.js
pm-azhar-mulla: Done

Issue #8665, opened 2022-07-08
Prebid Server: native ad JS event tracker support


jbartek25:
## Type of issue
Question/Feature

## Description
DV360 requires support for native JS trackers as otherwise the bidding might be limited (more details [here](https://protocol.bidswitch.com/ssp-protocol/ssp-recommended-fields.html). Is there any reason why in [this code](https://github.com/prebid/Prebid.js/blob/master/modules/prebidServerBidAdapter/index.js#L695) there's only method 1 (img) but not also method 2? Since Prebid itself is a piece of JavaScript and hence it's safe to assume JS trackers are always supported. Or am I missing something?

musikele:
I don't know why it's only `1` since ages, the "general" opinion is that pixels cause less harm than javascript trackers (that in theory may inject malicious code). When we discussed this topic we thought about adding `2` but then we decided to keep existing behaviour (and keep only `1`).

bretg:
@musikele - is the intention of the new PBJS native approach to completely specify imp.native, or just imp.native.assets?

If not, then seems like imp.native.eventtrackers would just be specified there.

If so, I would propose creating an s2sConfig override for the other fields in ORTB imp.native. e.g.

```
pbjs.setConfig({
    s2sConfig: {
        ...
        ortbNative: {.     // replaces the default JSON 
              context: 1,
              plcmttype: 1,
              eventtrackers: [
                {event: 1, methods: [1]}  // so you can put whatever you want
              ]
        }
    }
});
```

musikele:
@bretg , 
the new native requires users to specify the whole `imp.native` request, comprising of `eventtrackers`. That piece of code is just to specify defaults if any value is missing. 

So, i don't think we need to change that piece of code right now... Also, not do this change in the native PR

bretg:
Ok cool. I don't see changes to the pbsBidAdapter in https://github.com/prebid/Prebid.js/pull/8086 -- agree we'd need some then, right?

 @jbartek25 - this is the proposed solution to what you're bringing up... the AdUnit would specify whatever event trackers desired. Sound ok?

bretg:
I don't understand this part of your statement @jbartek25 :

> the current default for 1.2 (JS trackers disabled) is actually breaking the behaviour established with native 1.1 where jstracker was always executed without any further checks

Aren't we talking about this specific piece of code? There's no 1.1 anywhere.
```
          mediaTypes['native'] = {
            request: JSON.stringify({
              context: 1,
[...]

PR #8701, opened 2022-07-18
Improve Digital adapter: flag support for native 1.2 JS trackers in bid request

jbartek25:
## Type of change
- [X] Feature

## Description of change
Improve Digital adapter: Include explicit declaration of native 1.2 JS imp tracker support in bid request. JS trackers were supported already but some DSPs (like DV360) now expect to see explicit confirmation in bid request.

### DIFF
--- a/modules/improvedigitalBidAdapter.js
+++ b/modules/improvedigitalBidAdapter.js
@@ -430,6 +430,9 @@ const ID_REQUEST = {
       return null;
     }
     const request = {
+      eventtrackers: [
+        {event: 1, methods: [1, 2]}
+      ],
       assets: [],
     }
     for (let i of Object.keys(nativeParams)) {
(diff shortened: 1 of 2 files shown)


patmmccann:
did you not want to allow PBS adapter to see this as well if a publisher configures it? Also it seems like we're diverging from PBS adapter here; it would seem a publisher would want to choose this in a consistent way.

patmmccann:
@jbartek25 your review / feedback of #8665 is quite welcome

patmmccann:
Any chance you could give feeedback directly on https://github.com/prebid/Prebid.js/pull/8086  ?

jbartek25:
@patmmccann let me put this PR on hold until next week.

patmmccann:
@jbartek25 8086 is now merged, I think it is fair default this config in your adapter as long as publishers have some way to specify they prefer no js trackers on native

jbartek25:
I'll look into updating the adapter to adopt the ortb native allowing the overrides via the ad unit media type native.ortb object

patmmccann:
After a bunch of discussion, i think we can merge this now, so up to you if you want to make edits or not

jbartek25:
ok. please merge then. @dgirardi is converting the Improve Digital adapter to ortb as part of https://github.com/prebid/Prebid.js/pull/8738 so I'll make sure the default is overridable as part of that PR.

patmmccann (approved):

PR #8709, opened 2022-07-19
Triplelift Bid Adapter: outstream support

nllerandi3lift:
## Type of change

- [x] Feature

## Description of change
Allows adunits configured as 'outstream' to be sent to endpoint
Allows adunits configured as 'outstream' AND 'banner' to be sent to endpoint within same impression object
exchange will return only one or the other
'banner' and 'instream' is still interpreted as instream-only
Modified banner+outstream tests to check for both objects
Added new tests for 'outstream only' and 'banner + incomplete outstream'
Outstream gets same TTL as Instream
Outstream placements:

if context=outstream and placement was not provided, set video.placement:3
if pub sets video.placement and it is not 3,4,5, set to placement:3
if pub sets as 3, 4, 5, keep as is


robertrmartinez (approved):
Looks good to me!

PR #8710, opened 2022-07-19
bidWatch Analytics Adapter : add creative endpoint

matthieularere-msq:
## Type of change

- [ ] Feature

## Description of change

- Makes a dereferenced copy of events received in order to be able to modify them without altering ad delivery
- Remove ad creative code from the payload send to our endpoint to minimize its size
- Remove gdpr vendor data for the same reason (not needed as long as we have the consent string)
- Add a new endpoint call dedicated to creatives which each bid response sent individually.


ChrisHuie:
@matthieularere-msq can you please add unit tests for this pr?

matthieularere-msq:
@ChrisHuie not that much. I've just add one line which should make sure that the lines related to gdpr vendor data is covered. However I don't really see how to add specific test for what was added as I don't see how to get access to my module's specific functions ?

matthieularere-msq:
Ok, I found the way to add test for what my changes.

modules/bidwatchAnalyticsAdapter.js
```diff
@@ -22,53 +26,108 @@ function getAdapterNameForAlias(aliasName) {
...
+  }
 }
 
 function handleBidWon(args) {
-  if (typeof allEvents.bidRequested == 'object' && allEvents.bidRequested.length > 0 && allEvents.bidRequested[0].gdprConsent) { args.gdpr = allEvents.bidRequested[0].gdprConsent; }
+  args = cleanArgObject(JSON.parse(JSON.stringify(args)), true);
+  let increment = args['cpm'];
+  if (typeof saveEvents['auctionEnd'] == 'object') {
+    for (let i = 0; i < saveEvents['auctionEnd'].length; i++) {
+      let tmpAuction = saveEvents['auctionEnd'][i];
```
dgirardi, line 83: these types of loops would look a bit cleaner as 

```
for (let tmpAuction of saveEvents['auctionEnd']) { ... }`
```

or
 
```
saveEvents['auctionEnd'].forEach((tmpAuction) => {...})
```

Although I think `auction`, `bid` would be better names than `tmpAuction`, `tmpBid` etc.

modules/bidwatchAnalyticsAdapter.js
```diff
@@ -22,53 +26,108 @@ function getAdapterNameForAlias(aliasName) {
...
-function setOriginalBidder(arg) {
+function cleanArgObject(arg, removead) {
+  if (typeof arg['bidderCode'] == 'string') { arg['originalBidder'] = getAdapterNameForAlias(arg['bidderCode']); }
+  if (typeof arg['creativeId'] == 'number') {
+    arg['creativeId'] = arg['creativeId'].toString();
+  }
+  if (removead && typeof arg['ad'] != 'undefined') {
+    arg['ad'] = 'emptied';
+  }
+  if (typeof arg['gdprConsent'] != 'undefined' && typeof arg['gdprConsent']['vendorData'] != 'undefined') {
```
dgirardi, line 37: IMO it'd be better to explicitly list what to keep rather than what to remove. As more features are added your payload will probably continue to grow, and I can't imagine new, unrecognized fields being too useful on your end. It would also make it less likely that we'd accidentally rename or remove some field that is important to you, because a simple code search will show us what you are depending on.

dgirardi (approved):
I have some suggestions below but this LGTM as it is.

modules/bidwatchAnalyticsAdapter.js
matthieularere-msq: Thanks for the suggestion @dgirardi 
I am no javascript specialist so I often go to the easiest way I know without searching for subtilities. As I have another PR to send very shortly I'll make the suggested changes with this next one.

Issue #8381, opened 2022-05-05
Intent to implement: User Agent Client Hints


patmmccann:
## Type of issue
Feature Request

## Description

https://github.com/prebid/Prebid.js/blob/3fd3b5dca5dcc3fcd9f3c949e8e1a3238ea6b836/modules/sspBCBidAdapter.js#L89 is currently the only adapter gathering user agent client hints. Before this proliferates, let's go ahead and declare we intend to implement a utils function to gather these into a single object all bidders can access, similar to refererInfo js file.

patmmccann:
Table 3.2.29 of https://iabtechlab.com/wp-content/uploads/2022/04/OpenRTB-2-6_FINAL.pdf indicates how to populate these fields into the ORTB2.6 object. We should follow that syntax and in Prebid 7, this will be part of the bid request object.

rtbh-lotani:
@patmmccann would /src/utils.js be a good place for it?

PR #8730, opened 2022-07-22
33Across Bid Adapter: Obtain display-related attributes 

carlosfelix:
## Type of change

- [ ] Bugfix
- [X] Feature
- [ ] New bidder adapter   
- [ ] Build related changes
- [ ] CI related changes
- [ ] Other

## Description of change
Obtain display-related client-side attributes and clients hints from browsers that support it 

- contact email of the adapter’s maintainer: carlos.felix@33across.com

### DIFF
--- a/modules/33acrossBidAdapter.js
+++ b/modules/33acrossBidAdapter.js
@@ -62,8 +62,8 @@ function getTTXConfig() {
   return ttxSettings;
 }
 
-const storage = getStorageManager({gvlid: GVLID, moduleName: BIDDER_CODE})
-const UA_DATA_KEY = `${BIDDER_CODE}UaReducedData`;
+export const storage = getStorageManager({gvlid: GVLID, moduleName: BIDDER_CODE})
+export const UA_DATA_KEY = `${BIDDER_CODE}UaReducedData`;
 
 function getStoredUaReducedData() {
   try {
@@ -86,7 +86,9 @@ function calculateUaReducedData() {
   ).then((uaData) => {
     uaReducedData = uaData;
 
-    storeUaReducedData(uaReducedData);
+    if (Object.values(uaData).find(value => !!value).length) {
+      storeUaReducedData(uaReducedData);
+    }
   });
 
   return uaReducedData;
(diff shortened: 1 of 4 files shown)


carlosfelix:
I've tried many different ways to reproduce the failing tests scenario that I am seeing in the CI, but no luck so far, all the specs are passing on my environment and other envs where I have tried.

patmmccann:
We're not allowing adapters to access these api's, please modify core and access from core

https://github.com/prebid/Prebid.js/issues/8381

modules/33acrossBidAdapter.js
```diff
@@ -61,6 +62,44 @@ function getTTXConfig() {
...
+function getStoredUaReducedData() {
+  try {
+    return JSON.parse(storage.getDataFromLocalStorage(UA_DATA_KEY));
+  } catch (err) {
+    return null;
+  }
+}
+
+function storeUaReducedData(uaData) {
+  storage.setDataInLocalStorage(UA_DATA_KEY, JSON.stringify(uaData));
```
patmmccann, line 77: why put this in storage?

patmmccann (changes requested):
User agent client hints should be gathered by core

patmmccann:
Update: discussed with 33across team on slack, updates pending

macinjosh32:
Hi Pat,
We're only intending to collect client hints within our adapter on a temporary basis (through end of Q3) so we can quickly understand the state of various browsers' UA deprecation. We'll be happy to switch the access over to enrichmentFpdModule thereafter if we decide we need it past Q3. It looks like several other modules, such sspBC's adapter and Just ID's User ID sub-module (via an external script) are already collecting these on a more permanent basis, so we were not anticipating any concerns as long as standards are being fairly applied across modules. Placing the values in storage is to help facilitate unit tests and ensure these were available upon initial request since the Client Hints API is asynchronous. Thanks in advance for your understanding.

patmmccann:
We noticed that as well, which is why we drew the line in the sand to make sure no more were added https://github.com/prebid/Prebid.js/issues/8381

anthonyjl92:
Hi Pat, we removed the storage and client hints from this PR - are you able to review the remaining changes? Thanks!

patmmccann (approved):

PR #8732, opened 2022-07-22
Build system: adjust babel configuration to reduce bundle size

dgirardi:
## Type of change

- [x] Build related changes

## Description of change

This updates babel configuration to:
 - not polyfill Object.assign (all browers we care about now support it)
 - share other polyfills among source files, instead of redefining them each time they are needed.

The result is ~10KB reduction in prebid-core and ~150KB overall.

Before:
```
-rw-rw-r-- 1 demetrio demetrio  147195 Jul 22 15:29 build/dist/prebid-core.js
-rw-rw-r-- 1 demetrio demetrio 2455160 Jul 22 15:29 build/dist/prebid.js
```

After:
```
-rw-rw-r-- 1 demetrio demetrio  136981 Jul 22 15:27 build/dist/prebid-core.js
-rw-rw-r-- 1 demetrio demetrio 2306056 Jul 22 15:27 build/dist/prebid.js
```

### DIFF
--- a/babelConfig.js
+++ b/babelConfig.js
@@ -24,7 +24,7 @@ module.exports = function (options = {}) {
     ],
     'plugins': [
       [path.resolve(__dirname, './plugins/pbjsGlobals.js'), options],
-      useLocal('babel-plugin-transform-object-assign'),
+      [useLocal('@babel/plugin-transform-runtime')],
     ],
   }
 }
--- a/package-lock.json
+++ b/package-lock.json
@@ -6,12 +6,13 @@
   "packages": {
     "": {
       "name": "prebid.js",
-      "version": "7.4.0-pre",
+      "version": "7.8.0-pre",
       "license": "Apache-2.0",
       "dependencies": {
         "@babel/core": "^7.16.7",
+        "@babel/plugin-transform-runtime": "^7.18.9",
         "@babel/preset-env": "^7.16.8",
-        "babel-plugin-transform-object-assign": "^6.22.0",
+        "@babel/runtime": "^7.18.9",
         "core-js": "^3.13.0",
         "core-js-pure": "^3.13.0",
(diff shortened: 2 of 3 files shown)


ChrisHuie (approved):

robertrmartinez (approved):

muuki88:
This seems to have broken our typescript build. During runtime we get the following error

```
Uncaught ReferenceError: exports is not defined
```

https://stackoverflow.com/questions/43042889/typescript-referenceerror-exports-is-not-defined

I'm trying to chase this down why it happens. Curious if others see this as well.

muuki88:
I found: https://github.com/webpack/webpack/issues/4039#issuecomment-419284940

> `@babel/plugin-transform-runtime`, and you are running Babel on CommonJS files (either your own, or in random node_modules, if you're trying to compile that). Babel assumes files are ES modules by default, meaning those transforms would insert import statements into your file, triggering case 1. above

I assume we have to detangle the babel configs in our build.

muuki88:
Now I found another bug.

https://github.com/prebid/Prebid.js/blob/a231b0da5935b474f5341a9070f7951222f2aeed/src/secureCreatives.js#L25-L29

The `FEATURES` variable is not imported or required in anyway. Our webpack build with prebid as a npm dependency works, but during runtime we now get

```
secureCreatives.js:25 Uncaught ReferenceError: FEATURES is not defined
    at ./node_modules/prebid.js/src/secureCreatives.js (secureCreatives.js:25:1)
    at __webpack_require__ (bootstrap:19:1)
    at ./node_modules/prebid.js/src/prebid.js (polyfill.js:18:2)
    at __webpack_require__ (bootstrap:19:1)
    at make namespace object:7:1
    at moli.mjs:41251:3
    at moli.mjs:41253:12
```

PR #8736, opened 2022-07-25
Permutive rtd module: update dev docs

bartholomews:
## Type of change

- [ ] Bugfix
- [ ] Feature
- [ ] New bidder adapter   
- [ ] Build related changes
- [ ] CI related changes
- [ ] Other

## Description of change

Update the md doc to be in sync with the changes in PR https://github.com/prebid/prebid.github.io/pull/3923


modules/permutiveRtdProvider.md
```diff
@@ -31,39 +32,117 @@ pbjs.setConfig({
...
 
 | Name           | ID  | Config structure                                  | Description                                                                          |
-| -------------- | --- | ------------------------------------------------- | ------------------------------------------------------------------------------------ |
+|----------------|-----|---------------------------------------------------|--------------------------------------------------------------------------------------|
 | IAB taxonomies | iab | { segtax: number, iabIds: Object<number, number>} | Transform segment IDs from Permutive to IAB (note: alpha version, subject to change) |
+
+#### Context
+
+Permutive is not listed as a TCF vendor as all data collection is on behalf of the publisher and based on consent the publisher has received from the user.
+This consent is provided to Permutive on the SDK level, rather than through the TCF framework.
```
patmmccann, line 62: Can you clarify what sdk means here?

modules/permutiveRtdProvider.md
```diff
@@ -31,39 +32,117 @@ pbjs.setConfig({
...
+Including the _ID_ of any bidder in the `acBidders` array.
+
+**Note:** If a Publisher ever needs to remove a manually-added bidder, the bidder will also need to be removed manually.
+
+### _Enabling Custom Cohort IDs for Targeting_
+
+Separately from Standard Cohorts - The Permutive RTD module also supports passing any of the **custom** cohorts created in the dashboard to some SSP partners for targeting
+e.g. setting up publisher deals. For these activations, cohort IDs are set in bidder-specific locations per ad unit (custom parameters).
+
+Currently, the supported bidders for custom cohort targeting are:
```
patmmccann, line 141: Perhaps "bidders with known support for custom cohort targeting" ?

modules/permutiveRtdProvider.md
```diff
@@ -31,39 +32,117 @@ pbjs.setConfig({
...
+The [prebid Bidders page](https://docs.prebid.org/dev-docs/bidders.html) contains instructions and a link to a list of possible bidder codes.
+
+Acbidders can be added or removed from the list using this feature, however, this will not impact any acbidders that have been applied using the manual method below.
+
+#### Option 2 - Manual
+
+As a secondary option, new demand partner bidders may be added manually.
+
+To do so, a Publisher may define which bidders should receive Standard Cohorts by
+Including the _ID_ of any bidder in the `acBidders` array.
```
patmmccann, line 132: _ID_ is not clear, should this say 'bidder code' ?

patmmccann (changes requested):
minor edits

modules/permutiveRtdProvider.md
```diff
@@ -31,39 +32,117 @@ pbjs.setConfig({
...
 | Name           | ID  | Config structure                                  | Description                                                                          |
-| -------------- | --- | ------------------------------------------------- | ------------------------------------------------------------------------------------ |
+|----------------|-----|---------------------------------------------------|--------------------------------------------------------------------------------------|
 | IAB taxonomies | iab | { segtax: number, iabIds: Object<number, number>} | Transform segment IDs from Permutive to IAB (note: alpha version, subject to change) |
+
+#### Context
+
+Permutive is not listed as a TCF vendor as all data collection is on behalf of the publisher and based on consent the publisher has received from the user.
+This consent is provided to Permutive on the SDK level, rather than through the TCF framework.
+This means that if user consent isn’t given for Permutive to fire, no cohorts will populate.
```
patmmccann, line 63: This is only true if gdpr enforcement is configured.

modules/permutiveRtdProvider.md
bartholomews: I have reworded - hopefully a bit clearer:
```
Rather than through the TCF framework, this consent is provided to Permutive when the user gives the relevant permissions on the publisher website which allow the Permutive SDK to run.
```

modules/permutiveRtdProvider.md
bartholomews: Due to the fact that we have to enable bidders for custom targeting on our side, I think we have control on this?

patmmccann: It is a rules violation to favor a bidder; all bidders should be able to see your cohorts if a publisher chooses to send them there explicitly. 

One reason for this is suppose you wanted to add support for say, Pubmatic, a publisher should not have to upgrade their Prebid software to achieve this. It is fair to say only known bidders understand the cohorts.

patmmccann: Also, you should be aware we plan to remove mentions of appnexus and magnite from your module in the near future, https://github.com/prebid/Prebid.js/issues/8596

patmmccann: To clarify, remove from your adapter but add to the other adapters, eg if magnite expects your segments on params.permutive, it needs to have that code in the magnite adapter.

patmmccann (approved):

modules/permutiveRtdProvider.md
bartholomews: I see, thanks for the info - it makes sense. I am going to share this with our product team to discuss / prioritize 👍

Issue #8562, opened 2022-06-14
Proposal: pbjs-ortb conversion library


dgirardi:
## Type of issue
Feature

## Description

There is an increasing number of adapters that interface with ORTB backends. Currently, each needs to replicate logic to convert PBJS bid objects to ORTB and back again; a dialect of that logic is also implemented in the PBS adapter.

We should extract that into one canonical implementation, and make it trivial to add a new adapter if it talks to an ORTB backend.

## Proposal

1. Move the [enrichmentFpdModule](https://github.com/prebid/Prebid.js/blob/master/modules/enrichmentFpdModule.js) logic into core and extend it to populate any ortb2 field that is not specific to the impression or bidder (for example, `device.ua`). Currently this is left to each individual adapter, but they rarely should have a reason to differ in their implementation. This is true for "traditional" (non-ORTB) adapters as well; consolidating these items into first-party data should allow some simplification in that regard too.

2. Implement utilities `bids2ortb` and `ortb2bids` such that an adapter may use them in this manner:

```
  buildRequests(bidRequests, bidderRequest) {
      return {         
           method: ...,
           url: ...,
           data: bids2ortb({bidderRequest})           
      }
  },
  interpretResponse(response, request) {
      return ortb2bids({response: response.body, request: request.data})
  }
```

these utilities should handle everything that the PBS adapter now understands, namely:

- banner, video, and native mediatypes
- coppa / gdpr / usp
- price floors
- multibid
- schain
- gpid / pbadslot
- any others?

### Adapter-specific customization

Current ortb bidders are surprisingly similar, but not identical.  Adapters are free to modify the return value they get from `bids2ortb` / `ortb2bids`, but that pattern is not ideal for some cases.

#### Bid filtering

The Rubicon adapter wants to create a separate ortb request for each one in a set of particular bids:
  https://github.com/prebid/Prebid.js/blob/48ad7190decf77cdda00827b38f334856eff4de8/modules/rubiconBidAdapter.js#L179-L183

*Proposed solution* - accept an optional list of bid requests in `bids2ortb`:
```
  const videoRequests = bidRequests.filter(bidRequest => bidType(bidRequest) === 'video')
         .map(bidRequest => bids2ortb({bidderRequest, bidRequests: [bidRequest]})
```

#### Bid-level parameters

For example, from the ADF adapter:

[...]

PR #8738, opened 2022-07-25
PBJS-ORTB conversion library

dgirardi:
## Type of change

- [x] Feature

## Description of change

This is an initial version of the pbjs-ortb conversion library [proposal](https://github.com/prebid/Prebid.js/issues/8562), an attempt to unify all ORTB-related request generation and response parsing.

The following adapters have been refactored to use the new library:

 - Prebid Server
 - openxOrtb
 - improvedigital (FYI @jbartek25)

## Other information

Closes https://github.com/prebid/Prebid.js/issues/8562
Closes https://github.com/prebid/Prebid.js/issues/8665  (missing: documentation on new `s2sConfig.ortbNative` configuration)
Library documentation is included in `libraries/ortbConverter/README.md`


modules/openxOrtbBidAdapter.js
```diff
@@ -24,6 +22,99 @@ export const spec = {
...
+  request(buildRequest, imps, bidderRequest, context) {
+    const req = buildRequest(imps, bidderRequest, context);
+    mergeDeep(req, {
+      at: 1,
+      ext: {
+        bc: `${bidderConfig}_${bidderVersion}`
+      },
+      regs: {
+        // always set regs.coppa, even when not configured
+        // TODO: is this necessary?
```
bwschmidt, line 55: yeah probably not needed to be set to 0

modules/openxOrtbBidAdapter.js
```diff
@@ -24,6 +22,99 @@ export const spec = {
...
+    const bidResponse = buildBidResponse(bid, context);
+    if (bid.ext) {
+      bidResponse.meta.networkId = bid.ext.dsp_id;
+      bidResponse.meta.advertiserId = bid.ext.buyer_id;
+      bidResponse.meta.brandId = bid.ext.brand_id;
+    }
+    const {ortbResponse} = context;
+    if (ortbResponse.ext && ortbResponse.ext.paf) {
+      // TODO: the following is likely incorrect when an ORTB response contains more than one bid
+      // all of them would be set with the same content_id (from the last bid), since they share the same `paf` object
```
bwschmidt, line 84: yes thank you you are right the paf object needs to be cloned

modules/openxOrtbBidAdapter.js
```diff
@@ -24,6 +22,99 @@ export const spec = {
...
+        utils.deepSetValue(ortbResponse, 'ext.platform', ortbRequest.ext.platform);
+      }
+    }
+    return buildResponse(bidResponses, ortbResponse, context);
+  },
+  overrides: {
+    imp: {
+      bidfloor(setBidFloor, imp, bidRequest, context) {
+        // enforce floors should always be in USD
+        // TODO: does it make sense that request.cur can be any currency, but request.imp[].bidfloorcur must be USD?
```
bwschmidt, line 107: I believe the floors module defaults to USD anyway if getFloor is called without currency so I was just making it more clear. https://docs.prebid.org/dev-docs/modules/floors.html#currency. Also yes our exchange supports mismatched floor currencies.

modules/openxOrtbBidAdapter.js
```diff
@@ -47,182 +138,22 @@ function isBidRequestValid(bidRequest) {
 function buildRequests(bids, bidderRequest) {
   let videoBids = bids.filter(bid => isVideoBid(bid));
   let bannerBids = bids.filter(bid => isBannerBid(bid));
-  let requests = bannerBids.length ? [createBannerRequest(bannerBids, bidderRequest)] : [];
+  let requests = bannerBids.length ? [createRequest(bannerBids, bidderRequest, BANNER)] : [];
   videoBids.forEach(bid => {
-    requests.push(createVideoRequest(bid, bidderRequest));
+    // TODO: why is this creating a request for each bid, instead of aggregating them together as with banner?
```
bwschmidt, line 143: this is a limitation in our exchange currently. only banner requests can be aggregated, video requests only support a single slot

bwschmidt (commented):
Thanks for the notes, I can submit a fix for the paf bug if you like or you can change it, let me know. Coppa I don't mind either way. Let me know your feedback on the other items.

modules/openxOrtbBidAdapter.js
dgirardi: I'll update it in this PR, but this might take a while to get released so if you need the fix sooner you'll want to open one yourself.

modules/openxOrtbBidAdapter.js
dgirardi: the base logic (that is overridden here) is to do - simplified:
```
  const {floor, currency} = bidRequest.getFloor({
    currency: getConfig('currency.adServerCurrency') || 'USD'
  )
  imp.bidfloor = floor;
  imp.bidfloorcur = currency;
```

this fails tests because they enforce that, if the currency returned by `getFloor` is not 'USD', the result from the floors module should be ignored.

modules/improvedigitalBidAdapter.js
```diff
@@ -29,43 +15,169 @@ const IFRAME_SYNC_URL = 'https://hb.360yield.com/prebid-universal-creative/load-
...
-  VERSION: '1.2',
-  ASSET_TYPES: {
-    TITLE: 'title',
-    IMG: 'img',
-    DATA: 'data',
+export const CONVERTER = ortbConverter({
+  context: {
+    nativeRequest: {
+      eventtrackers: [
+        {event: 1, methods: [1, 2]},
```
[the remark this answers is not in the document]
dgirardi: You're right, that behavior does not make a lot of sense. Updated.

package-lock.json
```diff
@@ -40689,6 +40703,13 @@
       "integrity": "sha1-hnrHTjhkGHsdPUfZlqeOxciDB3c=",
       "dev": true
     },
+    "typescript": {
```
patmmccann, line 40706: is this a major change? in other words, does this imply other typescript use can start to proliferate around the project?

dgirardi: no. it was pulled in by some dependency (wdio I think, didn't check too hard)

IMO it's best to ignore autogenerated files in reviews, if there's a problem, it's really about the config that generates it (package.json in this case), that's what should get attention.

patmmccann:
@jsnellbaker looks like this is ready for reviewer assignment

modules/improvedigitalBidAdapter.js
```diff
@@ -29,43 +15,169 @@ const IFRAME_SYNC_URL = 'https://hb.360yield.com/prebid-universal-creative/load-
...
+    );
+
+    return imp;
+  },
+  request(buildRequest, imps, bidderRequest, context) {
+    const request = buildRequest(imps, bidderRequest, context);
+    mergeDeep(request, {
+      id: getUniqueIdentifierStr(),
+      source: {
+        // TODO: it may be better to not set this if there's more than one bidRequest - see https://github.com/prebid/Prebid.js/issues/8543
```
jbartek25, line 72: I don't see any harm. According to imp.ext.tid spec if both source.tid and imp.ext.tid are provided then imp.ext.tid should override, with source.tid used as a fallback for any items in the impression array that don't specify a imp.ext.tid.

modules/improvedigitalBidAdapter.js
```diff
@@ -29,43 +15,169 @@ const IFRAME_SYNC_URL = 'https://hb.360yield.com/prebid-universal-creative/load-
...
+        improvedigital: {
+          sdk: {
+            name: 'pbjs',
+            version: '$prebid.version$',
+          }
+        }
+      },
+    });
+    // always set cur, even if currency module is not enabled
+    // TODO: is this necessary?
```
jbartek25, line 85: Unlike other adapters, Improve Digital ad server responds with bids in the adServerCurrency if set.

jbartek25 (changes requested):

modules/improvedigitalBidAdapter.js
```diff
@@ -29,43 +15,169 @@ const IFRAME_SYNC_URL = 'https://hb.360yield.com/prebid-universal-creative/load-
...
   }
 };
 
-const NATIVE_DATA = {
-  VERSION: '1.2',
-  ASSET_TYPES: {
-    TITLE: 'title',
-    IMG: 'img',
-    DATA: 'data',
+export const CONVERTER = ortbConverter({
```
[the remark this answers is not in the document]
dgirardi: needed in tests

modules/improvedigitalBidAdapter.js
dgirardi: if only source.tid is specified, it should be taken as the default value for imp.ext.tid (according to that spec). That means that it's never correct to set it if there is more than one imp in the request.

modules/improvedigitalBidAdapter.js
dgirardi: this affects only the case when the currency module was not included. By default `cur` is set to `currency.adServerCurrency` if the currency module is included, and omitted if not. With this logic, you would most likely set it to "USD", but it's possible for someone to `setConfig('currency.adserverCurrency')` without actually having included the currency module.

jbartek25 (changes requested):

modules/improvedigitalBidAdapter.js
dgirardi: Agreed - the matter is not as settled as I thought.

Fawke:
Hi @dgirardi,

Can you please resolve the merge conflicts. I want to clone it locally and run some tests, unable to do so till it's merged with master.

Correct me if I'm wrong, you said in your proposal [here](https://github.com/prebid/Prebid.js/issues/8562#issue-1271455018) to move enrichmentModule logic into core. So, when that happens, do we get rid of that module? 

Plus, do we wanna add the docs to Prebid.org? As a guide for Adapter creators, perhaps somewhere over [here](https://docs.prebid.org/dev-docs/bidder-adaptor.html#building-the-request)?

dgirardi:
Thank you @Fawke! Just merged master in. 
This at the moment does not attempt to move fpdEnrichments into core, I figured this PR is big enough already. there's [this other issue](https://github.com/prebid/Prebid.js/issues/8990) to track that - I'm not sure about the strategy, probably keeping an empty module until we can remove it in v8.

Agree on making this visible in the new adapter guide, I'll come back with a PR.

bretg:
Created a test page that exercises a number of features that affect ORTB output: FPD, bidder-specific FPD, GDPR, COPPA, Floors, auction ID override, s2sConfig.extPrebid, sharedId, eidPermissions, global schain, bidder-specific schain, and aliases. Compared the results from the live file on jsdelivr vs this branch.

Two items of substance:
- the library is missing tmax. This should be the value defined by s2sConfig.timeout.
- it doesn't produce a global schain entry. Specifying a global schain only results in a bidder-specific schain going to PBS. Having both a global and bidder-specific schain

Comments:
- topframe is new - nice
- the library produces device.{dnt, ua, language} - nice
- cosmetic, but I'd prefer that '$.id' be the first thing in the struct and imp[].id be the first thing in the imp

dgirardi:
Fixed `tmax`; `source.ext.schain` is now set to the schain that is used my the most bidders in the request. 

Regarding `id`, JSON object property order is  in general not defined I think. Firefox' network tab orders them alphabetically. Javascript objects do preserve insertion order, but I don't think it's worth having extra code for that.

bretg:
@dgirardi - @musikele found one more gap -- pricegranularity.

PBS should receive ext.prebid.targeting.pricegranularity so it quantizes the hb_pb* targeting values appropriately.

dgirardi:
@bretg, I don't know what that refers to - I don't see granularity in PBS requests generated from master. Is this a regression or a new requirement? What should be in that field exactly?

musikele (commented):
I've been using it while working in my adapter, and in general there are no stoppers. I left some comments regarding "how it works", speaking of how i understood it, and also some questions to the author.

libraries/ortbConverter/converter.js
```diff
@@ -0,0 +1,137 @@
...
+    return function () {
+      if (build == null) {
+        build = (function () {
+          let delegate = builderFn.bind(this, compose(processors()[slot] || {}, overrides[slot] || {}));
+          if (wrapperFn) {
+            delegate = wrapperFn.bind(this, delegate);
+          }
+          return function () {
+            try {
+              return delegate.apply(this, arguments);
```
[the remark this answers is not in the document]
dgirardi: The arguments are those passed to the return value, e.g. `buildImp`, `buildRequest`, etc. They are derived only in part from what's passed to the public `toORTB`/`fromORTB`.

libraries/ortbConverter/README.md
```diff
@@ -0,0 +1,373 @@
...
+
+If your endpoint is a Prebid Server instance, you may take advantage of the `pbsExtension` companion library, which adds a number of processors that can populate and parse PBS-specific extensions (typically prefixed `ext.prebid`); these include bidder params (with `transformBidParams`), bidder aliases, targeting keys, and others. 
+
+```javascript
+import {pbsExtensions} from '../../libraries/pbsExtensions/pbsExtensions.js'
+
+const pbsConverter = ortbConverter({
+    processors: pbsExtensions
+})
+```
```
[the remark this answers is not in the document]
dgirardi: The user (meaning the publisher) does not need to change anything about the bundle; the idea is that you get for example currency logic if you include the currency module. Unit tests do need to import if they want the side effects, but they stop being unit tests at that point, so in the long run it should not be a common pattern. I will however make this clearer in the README.

libraries/pbsExtensions/processors/mediaType.js
```diff
@@ -0,0 +1,23 @@
+import {BANNER, NATIVE, VIDEO} from '../../../src/mediaTypes.js';
+import {ORTB_MTYPES} from '../../ortbConverter/processors/mediaType.js';
+
+export const SUPPORTED_MEDIA_TYPES = {
+  // map from pbjs mediaType to its corresponding imp property
+  [BANNER]: 'banner',
```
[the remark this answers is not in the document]
dgirardi: No, but it still makes sense to reuse the constants, since they are defined. Saves a couple bits.

libraries/ortbConverter/README.md
```diff
@@ -0,0 +1,373 @@
...
+   
+## Prebid Server extensions
+
+If your endpoint is a Prebid Server instance, you may take advantage of the `pbsExtension` companion library, which adds a number of processors that can populate and parse PBS-specific extensions (typically prefixed `ext.prebid`); these include bidder params (with `transformBidParams`), bidder aliases, targeting keys, and others. 
+
+```javascript
+import {pbsExtensions} from '../../libraries/pbsExtensions/pbsExtensions.js'
+
+const pbsConverter = ortbConverter({
+    processors: pbsExtensions
```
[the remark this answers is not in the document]
dgirardi: It's everything under `ext.prebid` (note that does _not_ include currency). I am not enthusiastic about listing them here because it's a moving target, it will get out of date really fast.

modules/priceFloors.js
```diff
@@ -745,3 +748,72 @@ export const addBidResponseHook = timedBidResponseHook('priceFloors', function a
...
+/**
+ * Sets bidfloor and bidfloorcur for ORTB imp objects
+ */
+export function setOrtbImpBidFloor(imp, bidRequest, context) {
+  if (typeof bidRequest.getFloor === 'function') {
+    let currency, floor;
+    try {
+      ({currency, floor} = bidRequest.getFloor({
+        currency: context.currency || config.getConfig('currency.adServerCurrency') || 'USD',
+        mediaType: context.mediaType || '*',
```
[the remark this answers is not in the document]
bretg: The getFloor function parses the absurdly complicated Prebid price floors schema, which does support mediatype and price. It's purpose is that bid adapters can say "what's the floor for video?" and get an answer without knowing the syntax of the file.

mmoschovas (approved):
Looks good from my end. Great work here!

PR #8663, opened 2022-07-08
next Millenium Bid Adapter: add video support

mifanich:
## Type of change
- [ ] Bugfix
- [x] Feature
- [ ] New bidder adapter   
- [ ] Code style update (formatting, local variables)
- [ ] Refactoring (no functional changes, no api changes)
- [ ] Build related changes
- [ ] CI related changes
- [ ] Does this change affect user-facing APIs or examples documented on http://prebid.org?
- [ ] Other

## Description of change
Video format support has been added

Be sure to test the integration with your adserver using the [Hello World](/integrationExamples/gpt/hello_world.html) sample page.

- contact email of the adapter’s maintainer
- [x] official adapter submission

For any changes that affect user-facing APIs or example code documented on http://prebid.org, please provide:

- A link to a PR on the docs repo at https://github.com/prebid/prebid.github.io/

PR #8740, opened 2022-07-26
NextMillennium Bid Adapter: Added video support

mifanich:
## Type of change

- [ ] Bugfix
- [x] Feature
- [ ] New bidder adapter   
- [ ] Build related changes
- [ ] CI related changes
- [ ] Other

This PR is a duplicate of broken PR https://github.com/prebid/Prebid.js/pull/8663


patmmccann:
@mifanich looks like the resubmission worked, ty!

modules/nextMillenniumBidAdapter.js
```diff
@@ -39,7 +62,19 @@ export const spec = {
...
-        ...bid.ortb2
+        ...bid.ortb2,
+        'imp': [{
+          'banner': {
+            'format': (bid.sizes || []).map(s => { return {w: s[0], h: s[1]} })
+          },
+          'ext': {
+            'prebid': {
+              'storedrequest': {
+                'id': id
```
patmmccann, line 73: @bretg here is another use case of an adapter explicitly referencing stored req id

bretg: They're good - the storedrequest id comes from a bidder-specific parameter. The issue that had been discussed is when the publisher defines a storedrequest for their main PBS provider. That data is passed to bidders, that may be running their own version of PBS for which that id would mean nothing.

mifanich:
@lksharma I have pulled the changes from master to my PR.
Could you please check and approve asap if everything is OK?

modules/nextMillenniumBidAdapter.js
```diff
@@ -1,17 +1,37 @@
...
+const EXPIRENCE_WURL = 20 * 60000
+const wurlMap = {}
+
+events.on(CONSTANTS.EVENTS.BID_WON, bidWonHandler)
+cleanWurl()
+
 export const spec = {
   code: BIDDER_CODE,
-  supportedMediaTypes: [BANNER],
+  supportedMediaTypes: [BANNER, VIDEO],
```
lksharma, line 34: This property isn't currently being used

modules/nextMillenniumBidAdapter.js
```diff
@@ -200,6 +241,32 @@ function getTopWindow(curWindow, nesting = 0) {
   }
 }
 
+function getAd(bid) {
+  let ad, adUrl, vastXml, vastUrl
+
+  switch (deepAccess(bid, 'ext.prebid.type')) {
+    case VIDEO:
```
lksharma, line 248: The response is being parsed for video however, I don't see a video request being generated in `buildRequests()`. Which video formats are you looking to support (outstream or instream)? 

Ref: https://docs.prebid.org/prebid-video/video-getting-started.html

modules/nextMillenniumBidAdapter.js
```diff
@@ -97,7 +119,9 @@ export const spec = {
           contentType: 'application/json',
           withCredentials: true
         },
-        bidId: bid.bidId
+
+        bidId,
+        auctionId,
```
lksharma, line 124: The `postBody` only supports the `banner` mediaType and not `video`.  Video params must be read from `AdUnit.mediaTypes.video` when constructing a request for video.

lksharma (changes requested):
@mifanich, 
 
Please create a documentation PR to indicate all supported mediaTypes.

modules/nextMillenniumBidAdapter.js
```diff
@@ -26,6 +46,8 @@ export const spec = {
     _each(validBidRequests, function(bid) {
       window.nmmRefreshCounts[bid.adUnitCode] = window.nmmRefreshCounts[bid.adUnitCode] || 0;
       const id = getPlacementId(bid)
+      const auctionId = bid.auctionId
+      const bidId = bid.bidId
       let sizes = bid.sizes
```
lksharma, line 51: `sizes` is always referencing `banner` sizes even if the video player is a different size.

mifanich:
> @mifanich,
> 
> Please create a documentation PR to indicate all supported mediaTypes.

https://github.com/prebid/prebid.github.io/pull/3978
@lksharma Done!

modules/nextMillenniumBidAdapter.js
mifanich: https://docs.prebid.org/dev-docs/bidder-adaptor.html
According the doc from the prebid.org is property exists as example.

modules/nextMillenniumBidAdapter.js
mifanich: @lksharma Yes. Empty size is expected case in our backend

modules/nextMillenniumBidAdapter.js
mifanich: @lksharma You're right. If you look at postbody you find param ext.prebid.storedrequest. In our side all sizes are predefined. So sizes that are passed via banner size, are used by our server in special cases

lksharma: Is it possible for your servers to return a video ad for a banner-only ad unit (example: 300x250)?  My concern here is that for a purely banner ad unit you might return a video response. In that case, the ad will not render since the video ad unit requires a `renderer` to be specified (for outstream) or a cache URL to be defined  (for instream).

mifanich: > Is it possible for your servers to return a video ad for a banner-only ad unit (example: 300x250)? My concern here is that for a purely banner ad unit and the backend returns a video response then the ad will not render. This is because video media types require a `renderer` to be specified (for outstream) or a cache URL to be defined (for instream).
@lksharma 
No. It is not possible. Because every placement ID is configured for specific publisher and placement. So if placement_id is configured for video, it will be used only for video player. The same case is for banner. As I said before, banner sizes we send only for our internal logic which does not break anything on the client side.

lksharma (approved):
LGTM

PR #8742, opened 2022-07-26
Gumgum Bid Adapter: changes for parameter consistency

john-ivan:
## Type of change

## Description of change
Updated the parameter for consistency with our server side implementations


patmmccann:
Appears to be a breaking change, why not support both

modules/gumgumBidAdapter.js
```diff
@@ -370,7 +370,7 @@ function buildRequests(validBidRequests, bidderRequest) {
         data.pi = 5;
       } else if (mediaTypes.video) {
         data.pi = mediaTypes.video.linearity === 2 ? 6 : 7; // invideo : video
-      } else if (params.product && params.product.toLowerCase() === 'skin') {
+      } else if (params.product && params.product.toLowerCase() === 'skins') {
```
patmmccann, line 373: why not support both skin and skins so this doesnt break anyone

patmmccann (changes requested):

modules/gumgumBidAdapter.js
john-ivan: @patmmccann hey.

Nobody is using the parameter as we haven't done rollout for it yet. We decided to stick with one iteration "skins" in consistency with how we intend to implement s2s as well as to keep it simple.

patmmccann (approved):

PR #8743, opened 2022-07-27
Relaido Bid Adapter: Add params for hashed canonical url.

relaido:
## Type of change

- [ ] Bugfix
- [x] Feature
- [ ] New bidder adapter   
- [ ] Build related changes
- [ ] CI related changes
- [ ] Other

## Description of change
- Add params for hashed canonical url.
- I use sha1, but it doesn't matter if the security is not secure.

## Comment on the following review
- This value is correct because it is treated as a page URL inside the server.
- https://github.com/prebid/Prebid.js/blob/master/modules/relaidoBidAdapter.js#L107-L108
 
```
// TODO: is 'page' the right value here?
ref: bidderRequest.refererInfo.page,
```


modules/relaidoBidAdapter.js
```diff
@@ -104,9 +105,10 @@ function buildRequests(validBidRequests, bidderRequest) {
     uuid: getUuid(),
     pv: '$prebid.version$',
     imuid: imuid,
+    canonical_url_hash: getCanonicalUrlHash(bidderRequest.refererInfo),
```
patmmccann, line 108: why do you want the url hash? why not just hash it server side?

patmmccann (changes requested):
curious what this is for

modules/relaidoBidAdapter.js
```diff
@@ -104,9 +105,10 @@ function buildRequests(validBidRequests, bidderRequest) {
...
     // TODO: is 'page' the right value here?
     ref: bidderRequest.refererInfo.page
-  })
+  });
```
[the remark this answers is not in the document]
relaido: @patmmccann 
todo comment removed. Thanks for the review.

modules/relaidoBidAdapter.js
relaido: @patmmccann 
The reason for hashing is to reduce the computational load on the server side. It also has the advantage of data reduction since the value is a fixed length.

patmmccann (approved):

PR #8747, opened 2022-07-27
Taboola Bid Adapter: dynamic ttl

ahmadlob:
## Type of change

- [x] Bugfix

## Description of change

editing a specific attribute  passing mechanism in the adapter file + adding unit test for the change.

### DIFF
--- a/modules/taboolaBidAdapter.js
+++ b/modules/taboolaBidAdapter.js
@@ -222,9 +222,7 @@ function getBid(requestId, currency, bidResponse) {
   if (!bidResponse) {
     return;
   }
-  if (typeof bidResponse['exp'] == 'undefined') {
-    bidResponse['exp'] = 60;
-  }
+  
   const {
     price: cpm, crid: creativeId, adm: ad, w: width, h: height, exp: ttl, adomain: advertiserDomains, meta = {}
   } = bidResponse;
--- a/modules/taboolaBidAdapter.js
+++ b/modules/taboolaBidAdapter.js
@@ -222,6 +222,10 @@ function getBid(requestId, currency, bidResponse) {
   if (!bidResponse) {
     return;
   }
+  
+  if(typeof bidResponse['exp'] == 'undefined') {
+    bidResponse['exp'] = 60;
+  }
 
   const {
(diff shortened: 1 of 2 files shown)


test/spec/modules/taboolaBidAdapter_spec.js
```diff
@@ -386,6 +387,31 @@ describe('Taboola Adapter', function () {
       const res = spec.interpretResponse(serverResponse, request)
       expect(res).to.deep.equal(expectedRes)
     });
+
+    it('should set the correct ttl form the response', function () {
+      // set exp-ttl to be 125
+      const [bid] = serverResponse.body.seatbid[0].bid;
+      serverResponse.body.seatbid[0].bid[0].exp = 125
```
patmmccann, line 394: semicolon?

modules/taboolaBidAdapter.js
```diff
@@ -233,7 +233,7 @@ function getBid(requestId, currency, bidResponse) {
 
   return {
     requestId,
-    ttl: 60,
+    ttl,
```
patmmccann, line 236: do you want to fallback to a default?

ahmadlob: just how I updated it as written

test/spec/modules/taboolaBidAdapter_spec.js
ahmadlob: fixed - added semicolon

patmmccann: in the lines below as well, our lgtm ends up flagging missing semicolons

modules/taboolaBidAdapter.js
ahmadlob: added check in case of fallback with null value to set default one.

test/spec/modules/taboolaBidAdapter_spec.js
patmmccann: tyvm, please note the trailing spaces causing the failed linting checks

patmmccann (approved):

PR #8748, opened 2022-07-27
Prebid core: fix native trackers for Prebid Server; simplify native ORTB logic

dgirardi:
## Type of change

- [x] Bugfix

## Description of change

A followup to https://github.com/prebid/Prebid.js/pull/8086, this:

 - fixes several things about how native trackers are handled. The original intent was to offload their processing to the PUC (https://github.com/prebid/prebid-universal-creative/pull/150); since that has not been merged, trackers do not work for adapters that reply in native ORTB (currently that is just the Prebid Server adapter). However there are other problems with that approach  - mixing of URLs and markups; as well as a long standing problem with confusing arrays and strings in the legacy `javascriptTrackers` .
 - adUnits (and bid requests derived from it) now have a `nativeOrtbRequest` property that adapters can use instead of manually converting  `bidRequest.nativeParams` (or `bidRequest.mediaTypes.native`). I think this makes more sense because all requests derived from an ad unit will result in the same ORTB native request object.
 - native rendering now retrieves the native request from the adUnit, instead of separately associating each request with the native request.

This is also in preparation of https://github.com/prebid/Prebid.js/pull/8738, where I plan to re-use this logic for all ORTB adapters (including the PBS adapter, which currently duplicates a lot of the logic).

FYI @bretg @musikele; with this, the guidance to adapters that want to deal with native in ORTB should be to pick `bidRequest.nativeOrtbRequest` (instead of `convertLegacyNativeRequestToOrtb(validBidRequests)`).

### DIFF
--- a/modules/prebidServerBidAdapter/index.js
+++ b/modules/prebidServerBidAdapter/index.js
@@ -446,11 +446,6 @@ if (FEATURES.NATIVE) {
     });
   });
 }
-/*
- * Protocol spec for OpenRTB endpoint
- * e.g., https://<prebid-server-url>/v1/openrtb2/auction
- */
-let nativeAssetCache = {}; // store processed native params to preserve
 
 /**
  * map wurl to auction id and adId for use in the BID_WON event
@@ -553,11 +548,13 @@ Object.assign(ORTB2.prototype, {
       this.adUnitsByImp[impressionId] = adUnit;
 
       const nativeParams = adUnit.nativeParams;
-      let nativeAssets = nativeAssetCache[impressionId] = deepAccess(nativeParams, 'ortb.assets');
+      let nativeAssets = deepAccess(nativeParams, 'ortb.assets');
       if (FEATURES.NATIVE && nativeParams && !nativeAssets) {
+        // TODO: all of this should not be necessary, the same logic is in native.js
+        // will be refactored as part of https://github.com/prebid/Prebid.js/pull/8738
         let idCounter = -1;
         try {
-          nativeAssets = nativeAssetCache[impressionId] = Object.keys(nativeParams).reduce((assets, type) => {
+          nativeAssets = Object.keys(nativeParams).reduce((assets, type) => {
             let params = nativeParams[type];
 
             function newAsset(obj) {
@@ -691,7 +688,7 @@ Object.assign(ORTB2.prototype, {
           // privacy: int
           assets: nativeAssets
         };
-        const ortbRequest = deepAccess(nativeParams, 'ortb');
+        const ortbRequest = deepAccess(adUnit, 'nativeOrtbRequest');
         try {
           const request = ortbRequest ? Object.assign(defaultRequest, ortbRequest) : defaultRequest;
           mediaTypes[NATIVE] = {
--- a/modules/prebidServerBidAdapter/index.js
+++ b/modules/prebidServerBidAdapter/index.js
@@ -2,7 +2,6 @@ import Adapter from '../../src/adapter.js';
 import {createBid} from '../../src/bidfactory.js';
 import {
   bind,
-  cleanObj,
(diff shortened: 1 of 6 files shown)


patmmccann:
@jbartek25 fyi

jbartek25:
@dgirardi Apologies for a question not directly related to this PR but I'm getting confused. I'm the maintainer of the Improve Digital bid adapter. We recently migrated our adapter to oRTB and so I was eagerly waiting for https://github.com/prebid/Prebid.js/pull/8086 (thank you for all the effort)  to be merged. Now as part of https://github.com/prebid/Prebid.js/pull/8701 I already started implementing logic to allow publishers to override the default native params via `nativeParams.ortb` and see that this PR changes `nativeParams` to `nativeOrtbRequest` and there's bunch of other related PRs open (including the docs PR https://github.com/prebid/prebid.github.io/pull/3497). I'm wondering, at which point can the bidder adapters start adopting ortb native? I thought it was after https://github.com/prebid/Prebid.js/pull/8086 got merged but now I'm not so sure. Please advise.

musikele (changes requested):
One issue and a question. 

- in file `native.js` at line 483 instead of 
```
    if (key in NATIVE_KEYS_THAT_ARE_NOT_ASSETS) continue;
```
change to 
```
    if (key in NATIVE_KEYS_THAT_ARE_NOT_ASSETS) continue;
```
This is a leftover bug by me. 

--- 

Regarding the trackers: I see that trackers are fired in PUC , file `nativeORTBTrackerManager.js`... how is it supposed to work? Were your changes only related to legacy behaviour?

musikele:
@jbartek25 before _releasing_ any change I'd wait for the PUC to be relased and the documentation to be published...

dgirardi:
@musikele, with the PUC version that's currently out, it would send [a message](https://github.com/prebid/prebid-universal-creative/blob/259030726f486f161995908688f360a8befdecf1/src/nativeTrackerManager.js#L56) that triggers `fireNativeTrackers` on the Prebid side - which with this, should work again for native ortb responses. (I don't know, however, if the exising PUC native tracker logic is sound).

The [https://github.com/prebid/prebid-universal-creative/pull/150](PUC native PR) wanted to move that logic to the PUC. For that to work it had to be merged & shipped first I think (it's something I missed when we were discussing the rollout). If you still want to go that route you would need to copy the logic in this PR over, it does not seem worth it to me - considering also that apparently some publishers prefer to _not_ use the PUC (and do [this](https://github.com/prebid/Prebid.js/blob/master/integrationExamples/gpt/x-domain/creative.html) instead).

dgirardi:
@jbartek25 there's also https://github.com/prebid/Prebid.js/pull/8738 - and I am looking for adapters like yours to convert with it! ([example of what I am doing with them](https://github.com/prebid/Prebid.js/pull/8738/commits/412241ae99ed5145d666f258246d97fbf0683237), and [documentation](https://github.com/dgirardi/Prebid.js/tree/pbjs-ortb-squash/libraries/ortbConverter)). It would automatically support native with the same logic used by the Prebid Server adapter, including this PR and https://github.com/prebid/Prebid.js/pull/8086, as well as any other future updates.

If you trust your test suite, or are willing to help me with verification in that PR, I can include your adapter - I am hoping to find a couple so that they can work as reference implementations.

jbartek25:
@dgirardi 
> If you trust your test suite, or are willing to help me with verification in that PR, I can include your adapter
Yes, sure :-). Our adapter's test suite is pretty extensive + I can help with testing as needed.

musikele (approved):
Tested with jstracker, eventtracker and everything should work for legacy native (that is the one everybody's using). 
In general, you cannot use mediatype.native.ortb until the PUC changes are adapted to support ortb responses in templates.

patmmccann (approved):

PR #7449, opened 2021-09-21
clean.io RTD Module : add new RTD module

nick-clean:
## Type of change

- [x] Other

## Description of change
The clean.io Realtime module provides effective anti-malvertising solution for publishers, including, but not limited to, blocking unwanted 0- and 1-click redirects, deceptive ads or those with malicious landing pages, and various types of affiliate fraud.  Using this module requires prior agreement with clean.io to obtain the necessary distribution key.

- contact email of the adapter’s maintainer

Module Name: clean.io Rtd provider
Module Type: Rtd Provider
Maintainer: nick@clean.io

- [X] official adapter submission

For any changes that affect user-facing APIs or example code documented on http://prebid.org, please provide:

- A link to a PR on the docs repo at https://github.com/prebid/prebid.github.io/

## Other information

clean.io module will be used for malvertising prevention and thus will need to execute Javascript.  There is historical precedent for this exception with this with GeoEdge similarly purposed adapter.  Please grant a similar exception.

PR #8750, opened 2022-07-27
CleanIO RTD Module: support billable event

alexeyst:
Add billable events to [clean.io Real-time Anti-Malvertising Module](https://docs.prebid.org/dev-docs/modules/cleanioRtdProvider.html)

## Type of change

- [x] Other

## Description of change
As described in [Vendor Billing](https://docs.prebid.org/dev-docs/vendor-billing.html), clean.io will now submit a billable event for each protected impression.

Module Name: clean.io Rtd provider
Module Type: Rtd Provider
Maintainer: [nick@clean.io](mailto:nick@clean.io)

## Other information
Solution was previously merged as part of [pull request #7449](https://github.com/prebid/Prebid.js/pull/7449)


nick-clean:
@ChrisHuie 
As we expected, the CircleCI tests failed. Our update includes utilizing a billable events; however, the tests seem to be hardcoded for a specific number of billable events so they fail when we send an unexpected event.
Screen Shot 2022-07-28 at 9 15 57 AM

Can you advise?  We believe this is an issue with the tests.

nick-clean:
@mmoschovas 
We're not surprised that the CircleCI tests failed. Our update includes utilizing a billable events; however, the tests seem to be hardcoded for a specific number of billable events so they fail when we send an unexpected event.

With a previous submission, we were told to pull in recent PRs to resolve the issue, but that doesn't look to resolve the problem, as the tests seem to be hardcoded for specific values.  Can you let us know how to proceed?

nick-clean:
> In the past, for things like
> 
> `should catch all events if affiliate id present Sovrn Analytics Adapter enableAnalytics AssertError: expected track to be called 5 times but was called 6 times`
> 
> we just increase the count in the expect

@patmmccann Thanks for jumping in. Are you stating that the tests need to be updated by manually adjusting the Expected count? Is that something @mmoschovas would do?  I unclear if this is something that you are requesting of us. Could you help clarify?

mmoschovas:
@nick-clean sorry for the delay. You would need to go through the tests that fail with the introduction of your billable event and increase their expected values

mmoschovas (approved):
lgtm

PR #8759, opened 2022-07-29
TL-29693: Adding support for all universal user IDs

patrickloughrey:
## Type of change
- [x] Feature

## Description of change
Up until this point, the Triplelift bid adapter only supported PubCommon ID, Criteo ID, LiveRamp ID, and Unified ID. This feature adds additional functionality to ingest all additional user IDs that a publisher is setting on their page.


patrickloughrey:
Merging into Triplelift master branch before merging into Prebid master

PR #8760, opened 2022-07-30
Prebid Core: Handled edge cases for allowedAlternateBidderCodes functionality

pm-azhar-mulla:
## Type of change

- [x] Feature

## Description of change

We have made a few minor changes to the `allowedAlternateBidderCodes` functionality
1. Made search case-insensitive, since the bidder names are in lowercase, we have made `allowedAlternateBidderCodes` array also lowercase before comparison.
2. Trimmed the entries in `allowedAlternateBidderCodes` array before comparison.
3. To keep the functionality in sync with [prebid-server](https://github.com/prebid/prebid-server/pull/2326). It will not accept bids from the alternate bidders when the array is empty ([]).


bsardo:
Hi team, I just wanted to bring it to your attention that the PBS team is actually planning to treat an empty array as "restrict all" instead of "allow all" as mentioned in the description here.
That means all bidder codes will be allowed if the array is not defined or if the array is defined as `["*"]`.

jsnellbaker (approved):
I think this looks fine.  Will need a second review.

pm-azhar-mulla:
@ncolletti can you please review this?

ncolletti (approved):
lgtm

Issue #8689, opened 2022-07-14
Various bidders: bypass of storage manager rules violation


patmmccann:
The following modules should have these bypasses fixed or disabled:

Modules should import storageManager and use for example, storage.getDataFromLocalStorage. In this way, deviceAccess controls and gdpr enforcement modules will govern the access

https://github.com/prebid/Prebid.js/blob/846bc8d65f150b94c7534da7e36192382285e8fa/modules/kargoBidAdapter.js#L128

@jsadwith 

https://github.com/prebid/Prebid.js/blob/3fd3b5dca5dcc3fcd9f3c949e8e1a3238ea6b836/modules/widespaceBidAdapter.js#L200
https://github.com/prebid/Prebid.js/blob/3fd3b5dca5dcc3fcd9f3c949e8e1a3238ea6b836/modules/widespaceBidAdapter.js#L189

@talha08

https://github.com/prebid/Prebid.js/blob/e47b9081ef3f6283053c46dd432a6d593d9b3558/modules/fintezaAnalyticsAdapter.js#L41

@finteza 

https://github.com/prebid/Prebid.js/blob/6f4fb5b78a4bc31a1f84b59ec5b98d4e3bcf0179/modules/sspBCBidAdapter.js#L104

@wojciech-bialy-wpm 

https://github.com/prebid/Prebid.js/blob/06e2976190aac71836f2fd0605598de38f4dc42d/modules/insticatorBidAdapter.js#L42

@EugeneVigonny 

https://github.com/prebid/Prebid.js/blob/6f4fb5b78a4bc31a1f84b59ec5b98d4e3bcf0179/modules/pubwiseAnalyticsAdapter.js#L160

@GLStephen 

https://github.com/prebid/Prebid.js/blame/5534049666913829624589561cdcb51e9038a8c7/modules/invibesBidAdapter.js#L384

@invibes

PR #8762, opened 2022-08-01
sspBC bid adapter - remove storage/cookie check, add screen size to request

wojciech-bialy-wpm:
## Type of change

- [x] Feature

## Description of change

- remove storage access (see https://github.com/prebid/Prebid.js/issues/8689)
- add screen size to device object in ortb request


patmmccann (approved):

robertrmartinez (approved):
Looks fine.

Usually we would want some unit test to verify changes for each PR.

Maybe a simple one to check the screen width and height are passed into your endpoint?

Removing the cookie check definitely should have probably caused some unit tests to fail as well! But no big deal!

Issue #7873, opened 2021-12-18
Batch video cache requests 


bretg:
## Type of issue
enhancement

## Description
Some publisher pages make many calls to the video cache endpoint. The Prebid Cache Server can handle taking multiple cache entries in one request, but PBJS isn't taking advantage of that feature.

I propose enhancing the caching subsystem to support optional aggregation of cache requests. Right now caching is done in a purely event-driven way: bid response comes in, caching needed, do it. The alternative would be to set a timer, wait for more responses to come in, and batch up any additional responses that come in before the timer pops. 

Proposed config:
```
pbjs.setConfig({
        cache: {
            url: 'https://my-pbs.example.com/vtrack',
            vasttrack: true,
            batchSize: 5,
            batchTimeout: 20   // how long in ms to wait for additional responses
        }
});
```
Note: the feature should work with or without vasttrack.

spormeon:
+1 from me, i'm one of the ones that needs it

spormeon:
another potential thought here as well is, maybe only the winning bids could be sent to cache, that would dramatically cut down the cache pushes for me, e.g. if got 30 bids, but 5 slots, only the 5 winning slot bids sent to cache.  Not sure if that feasible though and easier to send all?

patmmccann:
if pbc is on http2, what problem does this solve?

spormeon:
I was getting so many bids seing to PBC that it was meaning the rrsponce times were climbing and thus timeing it out.  Had a chat with Bretg and he though batching these up in less calls would help stop the timeouts  ( i think I got the idea why right)

karimMourra:
The Video Task Force discussed this and approves.

spormeon:
any development on this one, does it move to WIP now?

jlquaccia:
@bretg Just created a PR for this issue: https://github.com/prebid/Prebid.js/pull/8765.

Had a couple of questions as well:

1. Will the response from the POST /cache call preserve the same order as the request to the /cache endpoint?  For example, would the request order like this: `{ puts: [{obj1: “val1”}, {obj2: “val2”}] }` be returned in the same order with the response: `{ responses: [ {mapToObj1: uuidForObj1}, {mapToObj2: uuidForObj2} ] }` ?
[...]

PR #8765, opened 2022-08-01
Prebid Core: Batch Video Cache Requests feature

jlquaccia:
## Type of change

- [x ] Feature

## Description of change
This PR provides integration support for videos to optionally be cached in batch requests to the Prebid cache endpoint instead of all video cache requests getting sent individually.  This new feature can be enabled by simply including the "batchSize" key within the object passed to the "setConfig" function.  Additionally, the "batchTimeout" key can also optionally be set, stating the time in milliseconds that Prebid.js will wait to gather all batch requests before sending them out (default is 0 milliseconds).

Example:
```
pbjs.setConfig({
  cache: {
    url: 'https://prebid.adnxs.com/pbc/v1/cache',
    batchSize: 2,
    batchTimeout: 20
  }
})
```

## Other information

Related Issue: https://github.com/prebid/Prebid.js/issues/7873
@bretg, could you code review when you have time?  _Note: I left the "video.html" in the PR in the meantime for visibility around how I was testing out the new integration.  I can remove that after CR_


patmmccann:
@jlquaccia please note the 3 lgtm alerts

jlquaccia:
Thanks @patmmccann, I still had a test HTML file that I was using to test out my changes for this issue on the PR.  Just removed that file.  Looks like that was what had caused the 3 lgtm alerts

jlquaccia:
Thanks @dgirardi, addressed what you mentioned above and just pushed a few more commits just now.  I created a `batchRequests.init` function that now gets invoked after all batch requests have been processed by the Prebid /cache endpoint which I think should cover your first and last comment up above.

Could you review again when you have time?

src/videoCache.js
```diff
@@ -136,12 +136,13 @@ function shimStorageCallback(done) {
  * @param {videoCacheStoreCallback} [done] An optional callback which should be executed after
  * the data has been stored in the cache.
  */
-export function store(bids, done) {
+export function store(bids, done, isBatchRequest) {
```
dgirardi, line 139: Actually, looking into this more, I don't think you need to change this at all? why couldn't you do something like (continuing from my previous comment):

```
function storeInCache(batch, auctionInstance, afterBidAdded) {
    store(batch, function(err, cacheIds){
          cacheIds.forEach(cacheId, i) {
               const bidResponse = batch[i]; // instead of the combination of `batchBidResponseRequestIdMappingObject` and `batchedBidRequestIds`
               //....
          }
    })
}
```

The way you have it here, this `batchedBidRequestIds` is just passed in and back out again - this logic does not use it at all.

src/auction.js
```diff
@@ -514,7 +514,46 @@ function tryAddVideoBid(auctionInstance, bidResponse, afterBidAdded, {index = au
...
+
+                if (!currentBidResponse.vastUrl) {
+                  currentBidResponse.vastUrl = getCacheUrl(currentBidResponse.videoCacheKey);
+                }
+                addBidToAuction(auctionInstance, currentBidResponse);
+                afterBidAdded();
+              }
+            }
+
+            if (numberOfTimesStoreCacheFuncWasProcessed === batchBidResponseArrayToPass.length && requestIdIndex === batchedBidRequestIds.length - 1) batchRequests.init()
```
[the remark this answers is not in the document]
dgirardi: correction - there can also be multiple auctions going on at the same time, so you probably need to treat each trio `{bidResponse, auctionInstance, afterBidAdded}` as a single unit. E.g.

```
// in the partitioning logic
batches[batches.length - 1].push({bidResponse, auctionInstance, afterBidAdded});
```

and
```javascript
function storeInCache(batch) {
    store(batch.map(entry => entry.bidResponse), function(err, cacheIds) {
          cacheIds.forEach((cacheId, i) => {
                 const {bidResponse, afterBidAdded, auctionInstance} = batch[i];
                 // ...
          })
    })
}
```

jlquaccia:
hey @dgirardi, thanks for the review!  always appreciate your feedback 👍 💯

i went back and updated my PR and tested things locally in the browser and the logic around batch requests appear to look good in the network tab.  however, i have one question.  even though the new logic seems to be working fine, there are a handful of tests that look to be failing and i can't seem to tell why.

this test breaks and causes others to break after:
`✖ should run auction after video bids have been cached`
`AssertionError [ERR_ASSERTION]: 0 == 2`

(for a full list of the tests that failed, see this gist i made: https://gist.github.com/jlquaccia/ced04d2f4ead17207d7c822c21053efa)

i reviewed/debugged the new code in my PR several times but wasn't able to find the root cause of the issue.  i think i might just need an extra set of eyes haha.  do you have any idea what might be causing the tests to break?  Well anyway, know it's Friday night, have a great weekend!

src/auction.js
```diff
@@ -514,28 +514,67 @@ function tryAddVideoBid(auctionInstance, bidResponse, afterBidAdded, {index = au
...
+        batches.forEach(batch => storeInCache(batch));
+        batches = [[]];
+        debouncing = false;
+      }, batchTimeout);
+    }
+  }
+})();
+
+export const callPrebidCache = hook('async', function(auctionInstance, bidResponse, afterBidAdded, videoMediaType) {
+  batchAndStore(auctionInstance, bidResponse, afterBidAdded);
```
dgirardi, line 577: It's because here you always call the batched version, which means the cache will always be delayed. (Even if the batchTimeout is 0, it still waits for the microtask queue to clear before running it; which means that all tests that expect that to have happened already, fail). Plus there's a bunch of red herring failures because they're not configured that well and can't deal with some previous setup / teardown not running.


if you change line 577 to `storeInCache([{auctionInstance, bidResponse, afterBidAdded}])` all tests pass, which is still not ideal because it means that the batching logic is untested. An alternative would be to change line 567 to only use setTimeout if batchTimeout is greater than zero.

Something extra I should mention - I don't know how the cache endpoint works, so I am expecting it to do what your code assumes it does  -  which is always return a cacheId for each item that was requested, in the same order. Does that sound correct to you?

src/auction.js
```diff
@@ -514,28 +514,67 @@ function tryAddVideoBid(auctionInstance, bidResponse, afterBidAdded, {index = au
...
   });
+};
+
+const batchAndStore = (() => {
+  let batches = [[]];
+  let debouncing = false;
+
+  return function(auctionInstance, bidResponse, afterBidAdded) {
+    const batchSize =
+      typeof config.getConfig('cache.batchSize') === 'number' && config.getConfig('cache.batchSize') > 0
```
dgirardi, line 550: the getConfig call has some overhead, and this function runs many times per auction, so it's better if we try to minimize calls to it. You could separate the config out to 

```javascript
let batchSize, batchTimeout;
getConfig('cache', function(cacheConfig) => {
    batchSize = typeof cacheConfig.cache.batchSize ...
    batchTimeout = ...;
})
```

src/auction.js
```diff
@@ -514,28 +514,67 @@ function tryAddVideoBid(auctionInstance, bidResponse, afterBidAdded, {index = au
...
         }
-        addBidToAuction(auctionInstance, bidResponse);
-        afterBidAdded();
       }
-    }
+    });
   });
+};
+
+const batchAndStore = (() => {
```
dgirardi, line 544: A trick to make testing easier is to inject `setTimeout` and `storeInCache` as dependencies here, e.g.

```javascript
export function batchingCache(timeout = setTimeout, cache = storeInCache) {
    ...
    timeout(() => {
        batches.forEach(cache);
        ....
    })
}
const batchAndStore = batchingCache();
```

so that you can write (or reuse) tests for `storeInCache` without worrying about batching, and then inject stubs here to test batching without worrying about `storeInCache`:

```
let stubCache = sinon.stub();
let mockTimeout = sinon.stub().callsFake((cb) => cb())
const batchAndStore = batchingCache(mockTimeout, stubCache);
...
sinon.assert.calledWith(mockTimeout, sinon.match.any(), expectedBatchTimeout);
sinon.assert.calledWith(stubCache, expectedBatch);
```

src/auction.js
```diff
@@ -514,28 +514,73 @@ function tryAddVideoBid(auctionInstance, bidResponse, afterBidAdded, {index = au
...
+        ? cacheConfigBatchSize
+        : 1;
+
+  const batchTimeout =
+    typeof cacheConfigBatchTimeout === 'number' && cacheConfigBatchTimeout > 0
+      ? cacheConfigBatchTimeout
+      : 0;
+
+  return function(auctionInstance, bidResponse, afterBidAdded) {
+    const batchFunc = batchTimeout > 0 ? timeout : noTimeout;
```
jlquaccia, line 562: Added this line to handle conditionally using setTimeout or not based on the presence/absence of the batchTimeout key.

src/auction.js
jlquaccia: In regards to the cache endpoint, I was wondering the same thing.  According to the README page here, https://github.com/prebid/prebid-cache, the sample responses in the sections titled `Puts Parameters` and `Overwritting Values` look to demonstrate that responses do come back in the same order.

However, I also found these lines in the test file for cache endpoint:

without custom keys:

- https://github.com/prebid/prebid-cache/blob/master/endpoints/put_test.go#L150 (looks like the same elements from the response and request are checked for here, not 100% sure if the order is preserved in this scenario though..)

with custom keys

- https://github.com/prebid/prebid-cache/blob/master/endpoints/put_test.go#L176 (looks like order is preserved here 👍 )


Do you know anyone we could ask to confirm if the ordering will work as expected or if a change is needed?

test/spec/videoCache_spec.js
```diff
@@ -297,6 +298,46 @@ describe('The video cache', function () {
...
+      });
+
+      let mockTimeout = sinon.stub().callsFake((cb) => cb());
+      let stubCache = sinon.stub();
+      const batchAndStore = batchingCache(mockTimeout, stubCache);
+      batchAndStore();
+      sinon.assert.calledWith(mockTimeout, sinon.match.any, 20);
+    });
+
+    // it('should batch requests correctly based on the value of batchSize if batched requests are enabled in the config', () => {
```
jlquaccia, line 317: Had one question here.  I can almost get this test to work.  On line 330 below I call `batchAndStore` 3 times to simulate 3 bidResponses coming in and then want to test that all 3 are put into a batchSize of 3.  However, on line 571 in auction.js, the setTimeout gets ignored during this test and line 338 below is failing:

expected:
[
  { afterBidAdded: undefined, auctionInstance: undefined, bidResponse: undefined },
  { afterBidAdded: undefined, auctionInstance: undefined, bidResponse: undefined },
  { afterBidAdded: undefined, auctionInstance: undefined, bidResponse: undefined }
]

actual:
call 1
[
  { afterBidAdded: undefined, auctionInstance: undefined, bidResponse: undefined }
]

call 2
[
  { afterBidAdded: undefined, auctionInstance: undefined, bidResponse: undefined }
]

call 3
[
  { afterBidAdded: undefined, auctionInstance: undefined, bidResponse: undefined }
]

Is there a way to get past this setTimeout issue? (or perhaps i should just re-write my test below in a different way).  I tried using fake timers in sinon but didn't have any luck

src/auction.js
dgirardi: I asked in the prebid-server slack channel and I got confirmation that the order matches for the Go version, and no reply for the Java version.

jlquaccia:
@dgirardi thank you for all the feedback!  Left a few comments above

src/auction.js
```diff
@@ -514,28 +514,73 @@ function tryAddVideoBid(auctionInstance, bidResponse, afterBidAdded, {index = au
...
+  const noTimeout = cb => cb();
+  const cacheConfigBatchSize = config.getConfig('cache.batchSize');
+  const cacheConfigBatchTimeout = config.getConfig('cache.batchTimeout');
+
+  const batchSize =
+      typeof cacheConfigBatchSize === 'number' && cacheConfigBatchSize > 0
+        ? cacheConfigBatchSize
+        : 1;
+
+  const batchTimeout =
```
dgirardi, line 556: this runs too early - before the pub had a chance to `setConfig`, so it will always result in size 1 and timeout 0.

src/auction.js
jlquaccia: Oh nice, that's good to know for the Go version!  Hopefully the same outcome for the Java one too..

test/spec/videoCache_spec.js
jlquaccia: Thanks!  Your suggestion totally makes sense.  I updated the test, and now it's almost 100% working, but have one last question.  I can see that sinon now expects the correct batch, yet the `expectedBatch` var i created seems to get formatted while i run my test and fails.  i tried using `// eslint-disable-next-line` but that didn't seem to have an effect

jlquaccia: can see the output in the circle ci test that failed

test/spec/videoCache_spec.js
```diff
@@ -297,6 +298,34 @@ describe('The video cache', function () {
...
+      let stubCache = sinon.stub();
+      const batchAndStore = batchingCache(mockTimeout, stubCache);
+      for (let i = 0; i < 3; i++) {
+        batchAndStore({}, {}, function() {});
+      }
+
+      sinon.assert.calledOnce(mockTimeout);
+      sinon.assert.calledWith(mockTimeout, sinon.match.any, 20);
+
+      const expectedBatch = [{ afterBidAdded: function () {}, auctionInstance: { }, bidResponse: { } }, { afterBidAdded: function () {}, auctionInstance: { }, bidResponse: { } }, { afterBidAdded: function () {}, auctionInstance: { }, bidResponse: { } }];
```
dgirardi, line 322: different functions have different identities; `function(){} === function(){}` is always false. Same goes for `{} === {}`. You need to pass in the same objects if you want to test them for equality.

```javascript
const mockAfterBidAdded = function() {}
// ...
batchAndStore(mockAfterBidAdded, {}, {})
// ...

expectedBatch = [{afterBidAdded: mockAfterBidAdded, ...}, ...]
```

I think `assertCalledWith` does deep equality checks, so `{}` matches `{}` even if they are different objects - it just doesn't know how to do deep equals on function objects. If that's not the case you will need to do the same with `bidResponse` and `auctionInstance`.

jlquaccia: ah, you're explanation makes sense.  just tried your fix and it worked 👍 thanks!

dgirardi (approved):

mmoschovas (approved):

spormeon:
this seems to work well for us, only thing i've spotted is if you set batchTimeout: 3000  as a test, then nothing gets sent on the "video modules" params.  Is targeting being sent before its waited 3000? no idea what/ if there is a "cut off" time?

Issue #8769, opened 2022-08-02
Native designs: Replacing placeholders for "missing" assets


anitaschiller:
## Type of issue

Question

## Description

While testing the new native creative implementation (https://docs.prebid.org/prebid/native-implementation.html), we realized that a prebid native ##macro## only gets replaced if a corresponding asset is sent. 

If the bidder e.g. doesn't send a value for `body` which would replace the `##hb_native_body##` placeholder, the string `##hb_native_body##` will be inserted in the HTML template. 

I suppose that when the native template calls `renderNativeAd(pbNativeTagData)` (with `pbNativeTagData.requestAllAssets` set to `true`), the function below is responsible to answer that request?

https://github.com/prebid/Prebid.js/blob/f0596c4366fc3cb5b12b28ea44af8b54c9b6bbc9/src/native.js#L228-L255

That would mean that a key-value pair for `body` only gets pushed in the `message.assets` array if `body` is part of `adObject.native`. If it's not, the placeholder won't be replaced.

## Question
So what exactly is the purpose of setting `requestAllAssets` to true? 😊 I'm sure there were a lot of thoughts behind this! 

## Wished behavior
We were hoping being able to use the `:empty` selector in our native design's CSS in order to set a HTML element to `display: none` in case the bidder didn't send a value that can be inserted instead of the placeholder. That - however - only works if the placeholder was replaced by an empty string.

Thank you for your help!!

PR #8773, opened 2022-08-03
Prebid Core: Replace missing native asset placeholders

anitaschiller:
## Type of change

- [x] Bugfix
- [ ] Feature
- [ ] New bidder adapter   
- [ ] Build related changes
- [ ] CI related changes
- [ ] Other

## Description of change

See the following issue: https://github.com/prebid/Prebid.js/issues/8769 
When setting `requestAllAssets` to false, missing assets are now replaced by an empty string instead of `undefined`. 

```
{
  bidder: '<bidder name>',
  params: {
    // ...
  }
}
```


dgirardi:
IMO this should be addressed from the PUC, but I have to admit that I have a hard time understanding native - so I'm not sure.

I know that it's possible to use custom renderers so I'm afraid that this might be a breaking change? because I can imagine some implementations for which it's useful to distinguish between an asset that is missing and one that is an empty string. It seems safer to me to put that fallback in the renderer (PUC in our case).

muuki88:
@dgirardi we decided to go for prebid.js, because this were the incoming bid data is processed and the simplest way to add defaults. However it's probably better to make this configurable to make it configurable like the `requestAllAssets` boolean.

E.g.

```html
<script src="https://cdn.jsdelivr.net/npm/prebid-universal-creative@latest/dist/native-render.js"></script>
<script>
    var pbNativeTagData = {};
    pbNativeTagData.pubUrl = "%%PATTERN:url%%";     // GAM specific
    pbNativeTagData.adId = "%%PATTERN:hb_adid%%";   // GAM specific
    // if you're using 'Send All Bids' mode, you should use %%PATTERN:hb_adid_BIDDER%%
    pbNativeTagData.requestAllAssets = false;
    
    // 
    pbNativeTagData.fallbackAssetValue = '';
    window.pbNativeTag.renderNativeAd(pbNativeTagData);
</script>
```

From https://docs.prebid.org/prebid/native-implementation.html

So we would extend this in PUC

https://github.com/prebid/prebid-universal-creative/blob/259030726f486f161995908688f360a8befdecf1/src/nativeAssetManager.js#L190-L196

to send the `fallbackValue` along. The `native.js` would then be able to access the fallbackValue.

https://github.com/prebid/Prebid.js/blob/dd3267dcbef849ea9ddb509967f95711e9f585d0/src/native.js#L386-L391

> but I have to admit that I have a hard time understanding native - so I'm not sure.

If you like, we can talk about this :smile:

dgirardi:
If we involve the PUC, it can manage the fallbackValue all by itself - I don't think it needs to bounce it off Prebid. It would just need to say `replace macro with the asset value, or fallbackValue if the asset value is undefined`. I think it would also be perfectly sensible to always have have a fallback of `''` in the PUC (without defining a new `fallbackValue` parameter), because I do not see how it is advantageous to show "undefined" in the rendered template. I am just unsure about doing it up here in Prebid because I do not know if any custom renderers would be affected.

muuki88:
True that _custom renderers_ maybe check of `undefined` specifically instead of `falsy` value.

In PUC it would be this code here: https://github.com/prebid/prebid-universal-creative/blob/259030726f486f161995908688f360a8befdecf1/src/nativeAssetManager.js#L340-L345

We would essentially 

1. replace all placeholders with the incoming assets
2. `scanForPlaceholders` and replace those with empty strings

```js
    (assets || []).forEach(asset => {
      const flag = (typeof win.pbNativeData !== 'undefined');
      const searchString = (adId && !flag) ? `${NATIVE_KEYS[asset.key]}:${adId}` : ((flag) ? '##'+`${NATIVE_KEYS[asset.key]}`+'##' : `${NATIVE_KEYS[asset.key]}`);
      const searchStringRegex = new RegExp(searchString, 'g');
      html = html.replace(searchStringRegex, asset.value);
    });
    
    // roughly like this
    scanForPlaceholders().forEach(placeholder => {
      html = html.replace(placeholder, '');
    })
```

anitaschiller:
I now implemented the replacement of missing assets in the PUC like @muuki88 suggested --> [PR](https://github.com/prebid/prebid-universal-creative/pull/171).

Thanks for checking!

anitaschiller:
@dgirardi I opened a new PR in the PUC: https://github.com/prebid/prebid-universal-creative/pull/171.
What do you think about this solution?

PR #8774, opened 2022-08-03
bidWatch Analytics Adapter : limit bandwidth usage + refactory

matthieularere-msq:
## Type of change

- [ ] Feature
- [ ] Other

## Description of change

- Call endpoint creatives only when needed based on another previous call response. This should avoid useless calls to be done.
- Code was cleaned as per advices in PR #8710

### DIFF
--- a/modules/bidwatchAnalyticsAdapter.js
+++ b/modules/bidwatchAnalyticsAdapter.js
@@ -12,6 +12,7 @@ const {
     BID_WON,
     BID_RESPONSE,
     BID_REQUESTED,
+    BID_TIMEOUT,
   }
 } = CONSTANTS;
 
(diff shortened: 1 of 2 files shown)


matthieularere-msq:
Once approved, could this be merged to both 7.x and 6.x ? Thanks

ChrisHuie:
@matthieularere-msq this is actually probably the last weekish we are releasing to 6.x-legacy. If you really want it in both can you please submit this change to the 6 legacy branch and we should be able to get it in.

matthieularere-msq:
@ChrisHuie sure it makes sense. I just did it, thanks.

ChrisHuie (approved):

Issue #8770, opened 2022-08-02
Increase of timeouts when using userId modules, since 6.20.0


JulieLorin:
## Type of issue
This is a bug/regression

## Description
We have noticed that we have an increase of timeouts when using a userID module (like criteoIdSystem), since 6.20.0 
I think this is linked to : https://github.com/prebid/Prebid.js/pull/8201/files
It was working fine on 6.19.0
What we consider a timeout the event `CONSTANTS.EVENTS.BID_TIMEOUT`

## Steps to reproduce
The example is made with Criteo but it seems to be global for user ID module. We have the same issue with ID5 for example.
Building with userId + criteoIdSystem is fine, as long as we do not add criteo in pbjsconfig.userSync.userIds.
We go from 3-4% timeout rate to 10% just by enabling criteo user sync in config.userSync.userIds

user Sync settings :
```
userSync: {
  userIds: [{
    name: 'criteo',
    storage: {
      type: 'html5',
      name: 'criteoid',
      expires: 365
    }
  }],
  syncDelay: 1000
},
```

All browsers are affected
Environment with and without GDPR consent are affected
We have tried to upgrade to Prebid 7.8.0 to see if things gets better but it does not (no change). 
I also tried to set the auctionDelay to 1, but this didn't have any impact

patmmccann:
we noticed as well, our hope was https://github.com/prebid/Prebid.js/pull/8626/files#diff-381b65d4141bd216f9021175532d712ce87a9c4a55985c0cf1226f03421b6014 fixed

patmmccann:
@jlustig11 confirms Rubicon DM oberved as well

JulieLorin:
Thanks, it's good to confirm it's not just us. 

I also tried to twitch the auctionDelay parameter to check if it could improve performances (but it did not). 
Let me know if you need a second set of hands to test/check things, I am available to help with that if there's anything I can do.

dgirardi:
@JulieLorin (and also @patmmccann ), are you able to see if using cookies vs local storage makes any difference?

dgirardi:
I also realized that there might be multiple issues going on - 6.20.0 had [this bug](https://github.com/prebid/Prebid.js/pull/8362) which would conceivable explain the increase in timeout. That was fixed in 6.23.0. Is that also something you could try?

JulieLorin:
[...]

PR #8786, opened 2022-08-04
removed localstorage from  adqueryIdSystem module

adquery:
## Type of change
- [ x] Bugfix

## Description of change
We gave up on storing the id in localstorage because this caused duplicate qid

### DIFF
--- a/modules/adqueryIdSystem.js
+++ b/modules/adqueryIdSystem.js
@@ -50,10 +50,6 @@ export const adqueryIdSubmodule = {
    * @returns {{qid:Object}}
    */
   decode(value) {
-    let qid = storage.getDataFromLocalStorage('qid');
-    if (utils.isStr(qid)) {
-      return {qid: qid};
-    }
     return (value && typeof value['qid'] === 'string') ? { 'qid': value['qid'] } : undefined;
   },
   /**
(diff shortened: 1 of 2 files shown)


patmmccann:
@smenzer is there something apparent to you that might help this user?

dgirardi:
@adquery, is there a chance you can help me track down this issue - where are you seeing duplicate ids? Do you have a `setConfig` that would let me reproduce it? I think this might be related to https://github.com/prebid/Prebid.js/issues/8770 but I have so far been unable to reproduce.

smenzer:
> @smenzer is there something apparent to you that might help this user?

@patmmccann I'm not sure exactly...we've always just used the built-in storage functionality of the user id module and don't recall ever having any major issues with it.

adquery:
@dgirardi the error is not related to https://github.com/prebid/Prebid.js/issues/8770.The error is in our module adquerySystemId, because I had passed wrong parametrs and I had  saved inneceserily to  localstorage.The effect was like that: `%7B%22qid%22%3A%22%257B%2522qid%2522%253A%2522%25257B%252522qid%252522%25253A%252522%2525257B%25252522qid%25252522%2525253A%25252522ea253fb7bc1f563cbb2d%25252522%2525257D%252522%25257D%2522%257D%22%7D .After decoding it reveals that there was json with qid {qid: 3418b020e761debe963e}.So, it is not that error
`

mmoschovas:
Looks like this code removal broke unit tests within the userid spec file. These will need to be corrected before merging

PR #8788, opened 2022-08-04
UserID Module: continue the auction if userId init fails

dgirardi:
## Type of change

- [x] Bugfix

## Description of change

If the userId initialization logic fails, it silently stops the auction from running. In theory with #8401 errors from ID submodules should be handled gracefully, but I think it's better to continue the auction regardless (possibly without any IDs) if some other part of the logic throws.

### DIFF
--- a/modules/userId/index.js
+++ b/modules/userId/index.js
@@ -614,9 +614,9 @@ function getPPID() {
  * @param {Object} reqBidsConfigObj required; This is the same param that's used in pbjs.requestBids.
  * @param {function} fn required; The next function in the chain, used by hook.js
  */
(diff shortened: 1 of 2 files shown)


patmmccann:
@jlustig11 fyi, might be related to #8770

jlustig11:
Thanks Patrick. @dgirardi I agree with your fix here, certainly better to continue the auction. Is the userID failing logged somewhere 

Regardless @patmmccann thanks for the mention. When this is merged Magnite can help test that version of Prebid.

patmmccann:
@jlustig11 any help on the path to merge would be much appreciated

robertrmartinez (approved):
Agree, continuing auction is definitely the right choice.

patmmccann (approved):

PR #8791, opened 2022-08-05
Prebid Core : fix issue of mergeConfig

pm-azhar-mulla:
## Type of change

- [x] Bugfix

## Description of change

We have observed following behaviour during our testing.
When `mergeConfig` is used after `setConfig`, all the properties which have object as value in the config are merged or get the latest value but the properties which are non-object, are not merged or does not get the latest value.
e.g:
```
pbjs.setConfig({
  testProp: 2000,
  testObj: {
     a1: 1
  }
});

pbjs.mergeConfig({
  testProp: 3000,
  testObj: {
     a1: 2,
     a2: 3
  }
});
```
After the above statements execution, the value of config object becomes
```
testProp: 2000,
testObj: {
  a1: 2,
  a2: 3
}
```
Here the value of `testObj` is merged but `testProp` isn't.
The reason being, when we call `mergeConfig`, it calls `mergeDeep` internally on the conflicting properties.
Since `mergeDeep` only works for object, we get above behaviour.

**Solution:**
Instead of using a `mergeDeep` on conflicting properties, we have used it on complete object.

### DIFF
--- a/src/config.js
+++ b/src/config.js
@@ -386,7 +386,11 @@ export function newConfig() {
         option = Object.assign({}, defaults[topic], option);
       }
 
-      topicalConfig[topic] = config[topic] = option;
+      try {
+        topicalConfig[topic] = config[topic] = option;
+      } catch (e) {
+        logWarn(`Cannot set config for property ${topic} : `, e)
+      }
     });
 
     callSubscribers(topicalConfig);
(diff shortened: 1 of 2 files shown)


dgirardi (approved):

ChrisHuie (approved):

patmmccann:
@pm-azhar-mulla @pm-harshad-mane @mike-chowla  we have witnessed many properties dropping from the pubmatic endpoint call in prebid 7.8 with the jw boost rtd adapter installed. Is this the fix?

PR #8792, opened 2022-08-05
Ras Bid Adapter: support for SlotSequence parameter

wsusrasp:
## Type of change

- [ ] Bugfix
- [X] Feature
- [ ] New bidder adapter   
- [ ] Build related changes
- [ ] CI related changes
- [ ] Other

## Description of change

I have added new parameter to the adapter.


patmmccann (changes requested):
Read this off the ad unit please

patmmccann:
Here's an example

https://github.com/prebid/Prebid.js/blob/6f4fb5b78a4bc31a1f84b59ec5b98d4e3bcf0179/modules/ttdBidAdapter.js#L198

wsusrasp:
Hello, I have updated the code to read the data off the adunit. However I seem to have an issue with one of the tests unrelated to the change, updating the branch with the current master did not seem to help. Do you have an idea what is the problem maybe? Thank you

ChrisHuie:
> Hello, I have updated the code to read the data off the adunit. However I seem to have an issue with one of the tests unrelated to the change, updating the branch with the current master did not seem to help. Do you have an idea what is the problem maybe? Thank you

Rerunning CircleCI. There is currently a flaky test that is popping up more often the last several days

patmmccann:
Appears to still not be read off mediatypes.banner.pos

wsusrasp:
@patmmccann It seems I have misunderstood due to naming collision - the pos that I am reading off the bid params is not the same as udnit's pos. The one here is our custom param for determining the position of adunit with the same name if they are embedded multiple times. It is impossible for me to set that parameter elsewhere and mixing it with mediaTypes pos would later yield incorrect results when generating reports, since it takes different values than we expect. How should I proceed in this case?

patmmccann:
interesting; the 'pos' is an extremely well understood field by most rtb participants; perhaps you'll want to reconsider the name to avoid the collision and the resulting confusion?

wsusrasp:
Sadly we are already using this value in multiple applications on our side and refactoring it will not be easy

patmmccann:
How about you copy it to pos on the endpoint but your publisher facing configuration has a better choice of name?

patmmccann:
Sequence seems like a candidate parameter name

wsusrasp:
That sounds reasonable, I'll update the PR shortly. Thanks

patmmccann (approved):

patmmccann:
please also copy your .md changes over to the docs repo

PR #8658, opened 2022-07-07
IX Bid Adapter: change outgoing exchange request type from GET to POST

shahinrahbariasl:
## Type of change

- [ ] Bugfix
- [x] Feature
- [ ] New bidder adapter   
- [ ] Code style update (formatting, local variables)
- [ ] Refactoring (no functional changes, no api changes)
- [ ] Build related changes
- [ ] CI related changes
- [ ] Does this change affect user-facing APIs or examples documented on http://prebid.org?
- [ ] Other

## Description of change

- Changes Adapter request to Exchange from GET to POST to have higher request size limit.
- No functional changes for users
- Tests updated

Contact [shahin.rahbariasl@indexexchange.com](mailto:shahin.rahbariasl@indexexchange.com) for more information.

No changes to user-facing APIs or docs.

PR #8794, opened 2022-08-05
draft: IX Bid Adapter: change outgoing exchange request type from GET to POST

ix-peiyuan:
## Type of change

- [ ] Bugfix
- [x] Feature
- [ ] New bidder adapter   
- [ ] Build related changes
- [ ] CI related changes
- [ ] Other

## Description of change

(The PR is a replicate of [#8658](https://github.com/prebid/Prebid.js/pull/8658), whose owned is on vacation and I am taking over)

Changes Adapter request to Exchange from GET to POST to have higher request size limit.
No functional changes for users
Tests updated

Contact [peiyuan.sun@indexexchange.com](mailto:peiyuan.sun@indexexchange.com) for more information.

### DIFF
--- a/modules/ixBidAdapter.js
+++ b/modules/ixBidAdapter.js
@@ -39,7 +39,7 @@ const BANNER_TIME_TO_LIVE = 300;
 const VIDEO_TIME_TO_LIVE = 3600; // 1hr
 const NATIVE_TIME_TO_LIVE = 3600; // Since native can have video, use ttl same as video
 const NET_REVENUE = true;
-const MAX_REQUEST_SIZE = 8000;
+const MAX_REQUEST_SIZE = 32000; // IX POST limit
 const MAX_REQUEST_LIMIT = 4;
 const OUTSTREAM_MINIMUM_PLAYER_SIZE = [144, 144];
 const PRICE_TO_DOLLAR_FACTOR = {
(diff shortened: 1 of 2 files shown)


ChrisHuie:
@ix-peiyuan looks like there is a linting error here but the code it is referring to looks to be removed?

modules/ixBidAdapter.js
```diff
@@ -39,7 +39,7 @@ const BANNER_TIME_TO_LIVE = 300;
 const VIDEO_TIME_TO_LIVE = 3600; // 1hr
 const NATIVE_TIME_TO_LIVE = 3600; // Since native can have video, use ttl same as video
 const NET_REVENUE = true;
-const MAX_REQUEST_SIZE = 8000;
+const MAX_REQUEST_SIZE = 32000; // IX POST limit
```
muuki88, line 42: So you keep the splitting logic in the adapter?

ix-peiyuan: who does splitting mean here? you mean different size limits co exist due to different version of PBJS adopted by pubs?

ix-peiyuan:
> @ix-peiyuan looks like there is a linting error here but the code it is referring to looks to be removed?

I did the fix for the linting error but it seems CI is failing for some test cases belong to `SonobiBidAdapter`?

modules/ixBidAdapter.js
ix-peiyuan: Yes, I think it is a good point. I will talk with my team on removing the redundant logic but for now we could keep this as another follow-up change after POST is successfully in live.

PR #8758, opened 2022-07-28
ShowHeroes Bid Adapter: added a new endpoint

vadim-mazzherin:
## Type of change
- [ ] Bugfix
- [x] Feature
- [ ] New bidder adapter 
- [ ] Code style update (formatting, local variables)
- [ ] Refactoring (no functional changes, no api changes)
- [ ] Build related changes
- [ ] CI related changes
- [ ] Does this change affect user-facing APIs or examples documented on http://prebid.org?
- [ ] Other

## Description of change
Added a new endpoint to a new our bidder server
P.S. Recreate PR #8635  (+ unit tests, eslint fixed, update .md)

- contact email of the adapter’s maintainer: vadim.mazzherin@showheroes.com
- [x] official adapter submission

PR #8816, opened 2022-08-10
ShowHeroes Bid Adapter: add new endpoint

ChrisHuie:
Resubmit of #8758 to kick off CircleCi @Rothalack @vadim-mazzherin


Rothalack:
Hey @vadim-mazzherin sorry, I should have noticed this earlier. Does this include new params that are not documented on your bidder reference page?

unitId, stage, contentPageUrl

If those are new params, go ahead and get a PR up to add those to your bidder reference.

This is what needs to be updated
https://docs.prebid.org/dev-docs/bidders/showheroes-bs.html

And the PR should go here https://github.com/prebid/prebid.github.io/pulls

vadim-mazzherin:
@ChrisHuie @Rothalack Thanks very much!

vadim-mazzherin:
@Rothalack, I have updated the documentation. 
stage and contentPageUrl parameters aren't public

Rothalack (approved):
This should be good to go, working in testing.

PR #6452, opened 2021-03-22
First Party Data module: Add new module and two submodules to populate defaults and validate ortb2

mmoschovas:
## Type of change

- [ ] Bugfix
- [X ] Feature
- [ ] New bidder adapter   
- [ ] Code style update (formatting, local variables)
- [ ] Refactoring (no functional changes, no api changes)
- [ ] Build related changes
- [ ] CI related changes
- [ ] Does this change affect user-facing APIs or examples documented on http://prebid.org?
- [ ] Other

## Description of change

Create new optional module to validate ortb2 as well as set ortb2 defaults if applicable.
- verify OpenRTB datatypes, remove/warn any that are likely to choke downstream readers
- verify that certain OpenRTB attributes are not specified
- suppress user FPD based on a TBD opt-out signal (_pubcid_optout)
- populate available data into object: referer, meta-keywords, cur, etc.

Issue described: https://github.com/prebid/Prebid.js/issues/6099

- test parameters for validating bids
```
{
  bidder: '<bidder name>',
  params: {
    // ...
  }
}
```

Be sure to test the integration with your adserver using the [Hello World](/integrationExamples/gpt/hello_world.html) sample page.

- contact email of the adapter’s maintainer
- [ ] official adapter submission

For any changes that affect user-facing APIs or example code documented on http://prebid.org, please provide:

- A link to a PR on the docs repo at https://github.com/prebid/prebid.github.io/

PR #8826, opened 2022-08-11
Prebid core: add utility to retrieve user agent client hints

dgirardi:
## Type of change

- [x] Feature

## Description of change

This adds a library utility to retrieve [UA client hints](https://developer.mozilla.org/en-US/docs/Web/API/NavigatorUAData) formatted as an ORTB 2.6 `device.sua` object.

Note that:

 - the browser API seems to be in the early stages, with only Chromium supporting it, and likely subject to change. There are promises involved now and maybe more in the future. For this reason I think it's too early to have Core query it by default, since it's possible for that to become too expensive / invasive at any point in the future.
 - the only potential consumer ([sspBC](https://github.com/prebid/Prebid.js/blob/3fd3b5dca5dcc3fcd9f3c949e8e1a3238ea6b836/modules/sspBCBidAdapter.js#L89)) is _not_ sending out an ORTB payload, so it's not easily refactored to use this.

## Other information
Closes https://github.com/prebid/Prebid.js/issues/8381


Fawke (commented):
Hi @dgirardi,

The change looks good to me. I know there's a [FPD Enrichment module and validation module](https://github.com/prebid/Prebid.js/pull/6452/files).

Do we wanna pre-populate `device.sua` and call this library from that module?

In future, if Core needs to access these client hints values instead of parsing the string, do we have to write another function (possibly in `utils.js`)? 

We would also need to document this.

Thanks.

dgirardi:
It makes sense to use this from FPD enrichment, I'll update this PR shortly.

I don't understand this:

> In future, if Core needs to access these client hints values instead of parsing the string,

which string? I think this could be used directly by core, although it wouldn't make sense to have it as a library then.

Fawke:
> which string? I think this could be used directly by core, although it wouldn't make sense to have it as a library then.

I meant the user agent string. For example, [here](https://github.com/prebid/Prebid.js/blob/91ce750022daedff8bc70f5fee94088d7a6cda18/src/utils.js#L790) we are checking for a safari browser.

I don't know if there'll ever be a use case for Core that makes use of userAgent information. So, I think, it's fine to have it as a library.

osazos (approved):
LGTM

Fawke:
Hi @dgirardi,

The change looks good. But do we wanna put a note in the Prebid.org docs somewhere to let adapter contributors know about this module, and how to use it?

PR #8842, opened 2022-08-16
Criteo Bid Adapter: add support for banner+native multiformat ad unit

leonardlabat:
## Type of change
- [X] Feature

## Description of change
Previously, the use of a native adunit was exclusive with the banner type.


ncolletti (approved):
lgtm

Issue #8775, opened 2022-08-03
displayio adapter: calls pubmatic get user id function


patmmccann:
@philan15  please don't call getuseridsaseids in the owpbjs library

https://github.com/prebid/Prebid.js/blob/b9fe9a2991e8383b6c121e19fe1d859fbd80365f/modules/displayioBidAdapter.js#L134

That's the wrong global and one should read the eids from the request, not this api.

PR #8847, opened 2022-08-16
Displayio Bid Adapter: custom render; fix eids payload

philan15:
## Type of change

- [x] Feature
- [x] Bugfix

## Description of change

- Allows use custom renderer for a given bid's media format. 
- Fixed bug with calls global pubmatic get user id function [#8775](https://github.com/prebid/Prebid.js/issues/8775)

Updated documentation: https://github.com/prebid/prebid.github.io/pull/3961

### DIFF
--- a/modules/displayioBidAdapter.js
+++ b/modules/displayioBidAdapter.js
@@ -1,19 +1,16 @@
 import {registerBidder} from '../src/adapters/bidderFactory.js';
 import {BANNER, VIDEO} from '../src/mediaTypes.js';
 import {Renderer} from '../src/Renderer.js';
-import {getWindowFromDocument, logWarn, deepAccess} from '../src/utils.js';
-import {createEidsArray} from './userId/eids.js';
+import {getWindowFromDocument, logWarn} from '../src/utils.js';
 
-const BIDDER_VERSION = '1.0.0';
+const ADAPTER_VERSION = '1.1.0';
 const BIDDER_CODE = 'displayio';
-const GVLID = 999;
 const BID_TTL = 300;
 const SUPPORTED_AD_TYPES = [BANNER, VIDEO];
 const DEFAULT_CURRENCY = 'USD';
 
 export const spec = {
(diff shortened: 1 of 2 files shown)


modules/displayioBidAdapter.js
```diff
@@ -42,117 +45,112 @@ export const spec = {
...
-          h: window.screen.height,
-          connection_type: connection ? connection.effectiveType : '',
+        pagecat: pageCategory ? pageCategory.split(',').map(k => k.trim()) : [],
+        keywords: keywords ? keywords.split(',').map(k => k.trim()) : [],
+        lang_content: document.documentElement.lang,
+        lang: window.navigator.language,
+        domain: window.location.hostname,
+        page: window.location.href,
+        ref: refererInfo.referer,
+        userids: deepAccess(bid, 'userId') ? createEidsArray(bid.userId) : {},
```
patmmccann, line 107: you shouldn't need to import createEidsArray. The output of this function is already on the bid request

modules/displayioBidAdapter.js
```diff
@@ -42,117 +45,112 @@ export const spec = {
...
-        device: {
-          w: window.screen.width,
-          h: window.screen.height,
-          connection_type: connection ? connection.effectiveType : '',
+        pagecat: pageCategory ? pageCategory.split(',').map(k => k.trim()) : [],
+        keywords: keywords ? keywords.split(',').map(k => k.trim()) : [],
+        lang_content: document.documentElement.lang,
+        lang: window.navigator.language,
+        domain: window.location.hostname,
+        page: window.location.href,
```
patmmccann, line 105: why not refererInfo.page

patmmccann (changes requested):

modules/displayioBidAdapter.js
philan15: @patmmccann Thank you for help! fixed

patmmccann (approved):

PR #8864, opened 2022-08-18
Prebid core: optimize getRefererInfo to run only once per page

dgirardi:
## Type of change

## Description of change

Refactor getRefererInfo to do the heavy lifting only once; profiler data shows this to be one of the more expensive pieces of logic in core.


Rothalack (approved):
This looks good to me, very straight forward. Do you want a second approval since this is core? @ChrisHuie

Issue #8749, opened 2022-07-27
Refine coordination of PBS+PBJS Floors


bretg:
## Type of issue
enhancement

## Description
The original integration between PBJS floors and PBS floors was implemented in the pbsBidAdapter to turn off server-side floors if client-side floors was active. This approach was taken to preserve the view of client-side analytics which wouldn't have access to decisions made server-side.

However, @patmmccann brought up a powerful use case: server-side floors may be better than client-side floors because they can incorporate country and possibly other aspects. So we came up with this alternate behavior:

If a publisher wants PBS floor logic to have a chance to generate a _higher_ floor, they would explicitly enable server-side floors with either of these two approaches:
```
pbjs.setConfig({
    s2sConfig: {
       ...
       extPrebid: {
           floors: {
               enabled: true
           }
       }
    }
});

OR

pbjs.setConfig({
    ortb2: {
       ext: {
           prebid: {
               floors: {
                   enabled: true
               }
           }
       }
    }
});
```

After merging s2sConfig.extPrebid and ortb2, pbsBidAdapter should do something like this:
```
   if ext.prebid.floors.enabled exists and is true
       loop through imps
           if imp[n].bidfloor exists, copy it to imp[n].ext.prebid.floors.floorMin
           keep track of the lowest floor value
       request.ext.prebid.floors.floorMin = lowestFloorValue

    // else perform the existing logic
    **else** if (typeof deepAccess(firstBidRequest, 'bids.0.floorData') === 'object') {
      request.ext.prebid.floors = { enabled: false };
    }
```

This will set the floorMins and enable server-side floors-processing.

dgirardi:
I am perplexed by the last part, about setting `imp[].ext.prebid.floors.floorMin`. Why isn't server picking it up from `imp[].bidfloor` ? isn't that the purpose of that field? I also don't see it documented in the [server docs](https://docs.prebid.org/prebid-server/features/pbs-floors.html#floor-signaling) - it mentions it only at the request level.

bretg:
The design on the server side is that `imp[].bidfloor` is the default floor, not the minFloor.

imp-level minFloor is not yet support server-side, but is in progress.

dgirardi:
isn't a floor the minimum by definition? the logic would be, as you said, to copy the value from `bidfloor`. I don't see why they would ever be different. But it's not that important, copying is simple enough.

jlquaccia:
[...]

PR #8865, opened 2022-08-18
Prebid Core: Refine coordination of PBS+PBJS Floors

jlquaccia:
## Type of change

- [x] Enhancement

## Description of change
- stronger coordination between pbs and pbjs floors
- will set the floorMins pbs if bidfloor values within imp's are detected

## Other information

Related Issue: https://github.com/prebid/Prebid.js/issues/8749
@dgirardi could you review when you have time?


modules/prebidServerBidAdapter/index.js
```diff
@@ -902,6 +902,20 @@ Object.assign(ORTB2.prototype, {
...
+      const bidfloorArray = [];
+      request.imp.forEach(imp => {
+        if (imp.bidfloor) {
+          imp.ext.prebid.floors = {
+            floorMin: imp.bidfloor
+          }
+          bidfloorArray.push(imp.bidfloor);
+        }
+      });
+      request.ext.prebid.floors.floorMin = Math.min(...bidfloorArray);
```
dgirardi, line 916: this loops over floors twice more, which is rather inefficient compared to  the zero that are required. You could keep track of the minimum as we calculate the floors, and here just append it to the request if `ext.prebid.floors.enabled` is true.

jlquaccia: thanks @dgirardi.  would the changes for my latest commit work?  also, one additional question, for each imp would i still need to assign `imp.ext.prebid.floors.floorMin = imp.bidfloor`?  (as seen in the comment above)

modules/prebidServerBidAdapter/index.js
```diff
@@ -728,6 +729,7 @@ Object.assign(ORTB2.prototype, {
       if (floor) {
         imp.bidfloor = floor.floor;
         imp.bidfloorcur = floor.currency;
+        bidfloorArray.push(floor.floor);
```
dgirardi, line 732: yes, you should copy over to floorMin here, and still spare the `Math.min` loop by just keeping the minimum here. (Instead of `bidfloorArray = []` to start, you could set `let floorMin = null` and here do `if (floorMin == null || floorMin < imp.bidfloor) { floorMin = imp.bidfloor }`.

jlquaccia: @dgirardi thanks, just made those changes to the PR

modules/prebidServerBidAdapter/index.js
```diff
@@ -728,6 +729,8 @@ Object.assign(ORTB2.prototype, {
       if (floor) {
         imp.bidfloor = floor.floor;
         imp.bidfloorcur = floor.currency;
+        imp.ext.prebid.floors = { floorMin: imp.bidfloor };
```
dgirardi, line 732: I'm not sure imp.ext.prebid always exists at this point; I would prefer `deepSetValue(imp, 'ext.prebid.floors.floorMin', imp.bidfloor)`. 

(I believe in some cases, like stored impressions where no params were set in `imp.ext.prebid.bidder`, this as it is would throw an error).

dgirardi: actually I wonder - if the pub already set `ortb2Imp.ext.prebid.floors.floorMin`, should we keep it or override it here? I think keeping it makes more sense but then that should also become the value you check for against the request-wide `floorMin`.

jlquaccia: hmm, not sure.  so the question is should `imp.ext.prebid.floors.floorMin` be set to `ortb2Imp.ext.prebid.floors.floorMin` (if exists) or `imp.bidfloor` as a fallback?

inside of the description for https://github.com/prebid/Prebid.js/issues/8749, it says `if imp[n].bidfloor exists, copy it to imp[n].ext.prebid.floors.floorMin`. also states that `imp[n].bidfloor` is the default floor and that imp-level minFloor is not yet supported server-side.  does that mean (for now at least), that the logic should be `if imp[n].bidfloor exists, copy it to imp[n].ext.prebid.floors.floorMin`?

maybe @bretg could provide some more clarity around this?

robertrmartinez: In my opinion if the pub has provided `ortb2Imp.ext.prebid.floors.floorMin` we should keep it and use that as the floorMin. But if the request-wide floor min is higher, than it should set it to the higher one.

The imp level floor mins should not overwrite the pub provided one if they are lower. Maybe I am misinterpreting though.

jlquaccia: @dgirardi @robertrmartinez could you review when you have time?  Spoke with @bretg on the issue page for this ticket (`https://github.com/prebid/Prebid.js/issues/8749`) and updated my PR just now.

dgirardi:
Does this need to take floor currency into account? See https://github.com/prebid/Prebid.js/issues/8938

jlquaccia:
@dgirardi updated the PR to address currency as we discussed.  also updated the tests on this PR

modules/prebidServerBidAdapter/index.js
```diff
@@ -728,6 +730,13 @@ Object.assign(ORTB2.prototype, {
       if (floor) {
         imp.bidfloor = floor.floor;
         imp.bidfloorcur = floor.currency;
+
+        const ortb2ImpFloorMin = imp.ext?.prebid?.floors?.floorMin;
```
dgirardi, line 734: this should fall back to `imp.ext?.prebid?.floorMin` (see https://github.com/prebid/Prebid.js/issues/8938#issuecomment-1234653789);

also, if we are condiering the currency, this should be converted to `floorMinCur` before the comparison with `floorMin`.

jlquaccia: thanks, just pushed a new commit with the changes

modules/prebidServerBidAdapter/index.js
```diff
@@ -717,18 +720,28 @@ Object.assign(ORTB2.prototype, {
...
       if (floor) {
         imp.bidfloor = floor.floor;
         imp.bidfloorcur = floor.currency;
+
+        if (floorMinCur == null) { floorMinCur = floor.currency }
+        const convertedFloorValue = convertCurrency(floor.floor, floor.currency, floorMinCur);
+        const ortb2ImpFloorMin = imp.ext?.prebid?.floors?.floorMin || imp.ext?.prebid?.floorMin;
+        const lowestImpFloorMin = ortb2ImpFloorMin && ortb2ImpFloorMin < convertedFloorValue ? ortb2ImpFloorMin : convertedFloorValue;
+
+        deepSetValue(imp, 'ext.prebid.floors.floorMin', lowestImpFloorMin);
```
dgirardi, line 743: Here's a test case: you have a bidRequest with 
```
  getFloor() {return {currency: 'JPY', floor: 2}},
  ortb2Imp: {
     ext: { prebid: { floors: { floorMinCur: 'USD', floorMin: 1 }}}
  }
```

Since 2 JPY is less than 1 USD, the expectation is that `imp[].ext.prebid.floors` is set to either 2 JPY, or its equivalent in USD. I think this implementation would set 1 JPY.

dgirardi (approved):

robertrmartinez (approved):
I think this is fine.

My only suggestion is that maybe add some comments somewhere describing the behavior?

Maybe this behavior is outlined in the docs, but for someone who is looking at code and does not know much about prebid + floors this may be a confusing section of code. Comments get removed when we browserify so I do not see any harm in adding some.

jlquaccia:
@robertrmartinez thanks, and good call.  Just added a few comments referencing the GH ticket and walking through the logic

robertrmartinez (approved):
Awesome looks great

PR #8866, opened 2022-08-18
Prebid core: optimize sizeMapping to reduce computation when not in use

dgirardi:
## Type of change

## Description of change

Refactor sizeMapping (v1) logic to optimize for the case when it's not being used. Profiler runs show this to be relatively expensive on a page without any sizeMapping configuration.


ChrisHuie (approved):

Rothalack (approved):

PR #8868, opened 2022-08-18
NextMillenium Bid Adapter: Remove ortb2 referrerInfo

JacobKlein26:
- [x] Bugfix

## Description of change

Stopped using ```bid.ortb2``` object, as it can be "frozen" (```Object.freeze```)  because of which nextMillennium threw an unexpected error,

Instead, the ```device``` and ```site``` object are created manually (using ```getRefererInfo```)

### DIFF
--- a/modules/nextMillenniumBidAdapter.js
+++ b/modules/nextMillenniumBidAdapter.js
@@ -200,12 +200,12 @@ function getTopWindow(curWindow, nesting = 0) {
   }
 }
 
-function getSiteObj(){
+function getSiteObj() {
   const refInfo = (getRefererInfo && getRefererInfo()) || {}
 
   const page = refInfo.page || refInfo.location || getWindowTop().location.href
   const ref = refInfo.ref || ''
-  const domain = refInfo.domain || (new URL(page).hostname).replace('www.','')
+  const domain = refInfo.domain || (new URL(page).hostname).replace('www.', '')
 
   return {
     page,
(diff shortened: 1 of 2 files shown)


modules/nextMillenniumBidAdapter.js
```diff
@@ -198,4 +200,25 @@ function getTopWindow(curWindow, nesting = 0) {
   }
 }
 
+function getSiteObj() {
+  const refInfo = (getRefererInfo && getRefererInfo()) || {}
+
+  const page = refInfo.page || refInfo.location || getWindowTop().location.href
```
dgirardi, line 206: I believe this would throw an error if Prebid is inside an iframe. the first two would be null and the third would throw.

Also, if `page` is missing, `location` will also be missing. `page` is `canonicalUrl || location`. Unless you want a fallback for the iframe case, you should be able to just copy over `page`, `ref`, and `domain`. 

The assumption made for `refererInfo` is that it's better to get no data rather than bad data, so if there's no good candidate for the top level URL, both `page` and `domain` will be null. If you prefer to fallback to the inner frame's URL it seems odd to do it for `domain` but not `page` as you have here.

JacobKlein26: Got it, 
I'll update with no fallbacks then.

Thanks!

dgirardi: Keep in mind also that the 2nd argument to `buildRequests` (`bidderRequest`) has the `refererInfo` object already. If you pick data from there it will be easier to mock from your tests; your last update broke them because during tests there is no referrer.

JacobKlein26: I just removed that test case, as I have domain so I think it's fine, 
Before it used to be under device, that's why I had 2...

dgirardi (approved):

PR #8870, opened 2022-08-19
BetweenBidAdapter: default value for the cur parameter

ignat-one:
## Type of change
- [x] Other

## Description of change
If the 'cur' parameter is not passed, the default setting is 'USD'.

- contact email of the adapter’s maintainer: khaylov@betweenx.com

### DIFF
--- a/modules/betweenBidAdapter.js
+++ b/modules/betweenBidAdapter.js
@@ -59,9 +59,9 @@ export const spec = {
       if (i.params.itu !== undefined) {
         params.itu = i.params.itu;
       }
-      if (i.params.cur !== undefined) {
-        params.cur = i.params.cur;
-      }
+
+      params.cur = i.params.cur || 'USD';
+
       if (i.params.subid !== undefined) {
         params.subid = i.params.subid;
       }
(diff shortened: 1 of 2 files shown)


modules/betweenBidAdapter.js
```diff
@@ -59,9 +59,9 @@ export const spec = {
       if (i.params.itu !== undefined) {
         params.itu = i.params.itu;
       }
-      if (i.params.cur !== undefined) {
-        params.cur = i.params.cur;
-      }
+
+      params.cur = i.params.cur || 'USD';
+
```
ChrisHuie, line 64: Did you want to remove the not undefined check for the currency parameter? Other params in this adapter are  still checking inside if blocks.

ignat-one: Hi, Chris. Thanks for your comment, yes you're right, I need the "cur" parameter to have a default value of "USD" if it was not defined and it should always be contained in the bid request.

ignat-one: @ChrisHuie, is everything ok? Do I need to fix something in the code?

ChrisHuie (approved):

bear-pond:
Docs PR: https://github.com/prebid/prebid.github.io/pull/3977

PR #8881, opened 2022-08-22
Permutive RTD Module: merge external submodule params

AntonioGargaro:
## Type of change

- [x] Bugfix
- [ ] Feature
- [ ] New bidder adapter   
- [ ] Build related changes
- [ ] CI related changes
- [ ] Other

## Description of change

Cohort activation within [Permutive's RTD module](https://docs.prebid.org/dev-docs/modules/permutiveRtdProvider.html#cohort-activation-with-permutive-rtd-module) is failing to be applied with [option 1](https://docs.prebid.org/dev-docs/modules/permutiveRtdProvider.html#option-1---automated), which let publishers set `acBidders` through Permutive's platform. We require this fix where the RTD module attempts to retrieve this config from LS or our Permutive SDK. This lets publishers modify their `Permutive <-> Prebid` integration without code changes.

We read from LS on initialisation in the event that the Permutive SDK has not initialised before the config is required. Every call to retrieve our config will attempt to interact with Permutive's prebid addon which will have this config available in memory, otherwise fallback to cache.

### Config Priority

  1. customModuleConfig <- set by publisher with pbjs.setConfig
  2. permutiveRtdConfig <- set by the publisher using the Permutive platform
  3. defaultConfig

The configuration module is deeply merged in reverse priority order. This means that the highest priority config (`customModuleConfig`) defined by `.setConfig` within Prebid will be merged last and overwrite other configurations.

## Other information

N/A

### DIFF
--- a/test/spec/modules/permutiveRtdProvider_spec.js
+++ b/test/spec/modules/permutiveRtdProvider_spec.js
@@ -5,9 +5,11 @@ import {
   initSegments,
   isAcEnabled,
   isPermutiveOnPage,
-  setBidderRtb
+  setBidderRtb,
+  getModuleConfig,
+  PERMUTIVE_SUBMODULE_CONFIG_KEY,
 } from 'modules/permutiveRtdProvider.js'
-import { deepAccess } from '../../../src/utils.js'
+import { deepAccess, deepSetValue, mergeDeep } from '../../../src/utils.js'
 import { config } from 'src/config.js'
 
 describe('permutiveRtdProvider', function () {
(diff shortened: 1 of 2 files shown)


test/spec/modules/permutiveRtdProvider_spec.js
```diff
@@ -29,6 +31,128 @@ describe('permutiveRtdProvider', function () {
...
+      // Read from Permutive
+      const permutiveRtdConfig = { acBidders: ['456'] }
+      const cleanupPermutive = setWindowPermutivePrebid(function () {
+        return permutiveRtdConfig
+      })
+
+      const customModuleConfig = { params: { acBidders: ['789'], maxSegs: 499 } }
+      const config = getModuleConfig(customModuleConfig)
+
+      const configMergedInPriorityOrder = mergeDeep(defaultConfig, liftToParams(cachedConfig), liftToParams(permutiveRtdConfig), customModuleConfig)
```
patmmccann, line 147: you want the configuration coming from window.permutive to override how the publisher configures the module in prebid?

patmmccann (commented):
j

test/spec/modules/permutiveRtdProvider_spec.js
AntonioGargaro: The name is misleading here, the configs are in reverse priority order as configs are merged left to right. So the priority is,

1. `customModuleConfig` <- set by publisher with `pbjs.setConfig`
2. `permutiveRtdConfig` <- set by config publisher configures inside Permutive.
3. `cacheConfig` <- Params stored in LS
4. `defaultConfig`

Therefore, anything set by the publisher will take priority over what is set within Permutive. I'll add some comments explaining this.

ChrisHuie:
@AntonioGargaro the integration tests don't appear to have ran on this pr. This is usually because the submitter is following a fork of prebid inside of CircleCI instead of the main Prebid project. You should be able to unfollow your fork in CircleCI and that should kick of testing.

AntonioGargaro:
@ChrisHuie Should be running now 👍

test/spec/modules/permutiveRtdProvider_spec.js
patmmccann: thanks!

patmmccann (approved):

PR #8893, opened 2022-08-24
Yieldlab Bid Adapter support floor module

nkloeber:
## Type of change
- [x] Feature

## Description of change
Yieldlab Bid Adapter support floor module in order to set dynamic floor prices

## Other information
1 approving review from a Yieldlab team member required

### DIFF
--- a/modules/yieldlabBidAdapter.js
+++ b/modules/yieldlabBidAdapter.js
@@ -1,9 +1,9 @@
-import {_each, deepAccess, isArray, isFn, isPlainObject, timestamp} from '../src/utils.js'
-import {registerBidder} from '../src/adapters/bidderFactory.js'
-import {find} from '../src/polyfill.js'
-import {BANNER, NATIVE, VIDEO} from '../src/mediaTypes.js'
-import {Renderer} from '../src/Renderer.js'
-import {convertOrtbRequestToProprietaryNative} from '../src/native.js';
+import { _each, deepAccess, isArray, isFn, isPlainObject, timestamp } from '../src/utils.js'
+import { registerBidder } from '../src/adapters/bidderFactory.js'
+import { find } from '../src/polyfill.js'
+import { BANNER, NATIVE, VIDEO } from '../src/mediaTypes.js'
+import { Renderer } from '../src/Renderer.js'
+import { convertOrtbRequestToProprietaryNative } from '../src/native.js';
 
 const ENDPOINT = 'https://ad.yieldlab.net'
 const BIDDER_CODE = 'yieldlab'
@@ -181,7 +181,7 @@ export const spec = {
           bidResponse.adUrl = url
           bidResponse.mediaType = NATIVE
           const nativeImageAssetObj = find(matchedBid.native.assets, e => e.id === 2)
-          const nativeImageAsset = nativeImageAssetObj ? nativeImageAssetObj.img : {url: '', w: 0, h: 0};
+          const nativeImageAsset = nativeImageAssetObj ? nativeImageAssetObj.img : { url: '', w: 0, h: 0 };
           const nativeTitleAsset = find(matchedBid.native.assets, e => e.id === 1)
           const nativeBodyAsset = find(matchedBid.native.assets, e => e.id === 3)
           bidResponse.native = {
@@ -417,11 +417,11 @@ function outstreamRender(bid) {
  * @returns {string[]}
  */
 function extractSizes(bid) {
-  const {mediaTypes} = bid // see https://docs.prebid.org/dev-docs/adunit-reference.html#examples
+  const { mediaTypes } = bid // see https://docs.prebid.org/dev-docs/adunit-reference.html#examples
   const sizes = []
 
   if (isPlainObject(mediaTypes)) {
-    const {[BANNER]: bannerType} = mediaTypes
+    const { [BANNER]: bannerType } = mediaTypes
 
     // only applies for multi size Adslots -> BANNER
     if (bannerType && isArray(bannerType.sizes)) {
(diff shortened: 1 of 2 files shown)


modules/yieldlabBidAdapter.js
```diff
@@ -441,4 +445,31 @@ function extractSizes(bid) {
...
+ * @param {string[]} sizes
+ * @returns The floor CPM of a matched rule based on the rule selection process (mediaType, size and currency),
+ *          using the getFloor() inputs. Multi sizes and unsupported media types will default to '*'
+ */
+function getBidFloor(bid, sizes) {
+  if (!isFn(bid.getFloor)) {
+    return undefined;
+  }
+  const mediaTypes = deepAccess(bid, 'mediaTypes');
+  const mediaType = mediaTypes !== undefined && spec.supportedMediaTypes.includes(Object.keys(mediaTypes)[0].toLowerCase())
```
brushmate, line 462: The result of `Object.keys(mediaTypes)[0].toLowerCase()` could be stored in a variable, otherwise it might get executed twice.

modules/yieldlabBidAdapter.js
```diff
@@ -441,4 +445,31 @@ function extractSizes(bid) {
...
+    ? Object.keys(mediaTypes)[0].toLowerCase() : '*';
+  const floor = bid.getFloor({
+    currency: CURRENCY_CODE,
+    mediaType: mediaType,
+    size: sizes.length !== 1 ? '*' : extractSizes(sizes)
+  });
+  if (isPlainObject(floor) && !isEmpty(floor) && !isNaN(floor.floor) && floor.currency === CURRENCY_CODE) {
```
brushmate, line 469: I think

```javascript
if (floor.currency === CURRENCY_CODE) {
  ...
}
```

should be enough. The docs say that `getFloor()` returns either

```
{
  currency: string,
  floor: float
}
```

or the empty object `{}` if a floor could not be found for the given input.

test/spec/modules/yieldlabBidAdapter_spec.js
```diff
@@ -531,4 +531,50 @@ describe('yieldlabBidAdapter', () => {
...
+        return {};
+      };
+      bidRequest = Object.assign(DEFAULT_REQUEST(), {getFloor})
+      const result = spec.buildRequests([bidRequest], REQPARAMS)
+      expect(result).not.to.have.nested.property('queryParams.floor')
+    });
+
+    it('should not add bid floor when value is not valid', () => {
+      getFloor = () => {
+        return undefined;
```
brushmate, line 561: According to the docs `getFloor()` will never return `undefined`.

brushmate (changes requested):

modules/yieldlabBidAdapter.js
nkloeber: I'm worried that it is possible that the currency is not the same. From the [documentation](https://docs.prebid.org/dev-docs/modules/floors.html#getfloor-response:~:text=If%20the%20Floor%20Module%20cannot%20convert%20a%20floor%20to%20the%20supplied%20currency%2C%20bid%20adapters%20will%20be%20required%20to%20handle%20the%20supplied%20floor.):
_If the Floor Module cannot convert a floor to the supplied currency, bid adapters will be required to handle the supplied floor._

brushmate: I just saw that other adapters do it the same way. So I guess it's fine.

modules/yieldlabBidAdapter.js
```diff
@@ -1,9 +1,9 @@
-import { _each, deepAccess, isArray, isPlainObject, timestamp } from '../src/utils.js'
-import { registerBidder } from '../src/adapters/bidderFactory.js'
-import { find } from '../src/polyfill.js'
-import { BANNER, NATIVE, VIDEO } from '../src/mediaTypes.js'
-import { Renderer } from '../src/Renderer.js'
-import { convertOrtbRequestToProprietaryNative } from '../src/native.js';
+import {_each, deepAccess, isArray, isEmpty, isFn, isPlainObject, timestamp} from '../src/utils.js'
```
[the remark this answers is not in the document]
nkloeber: It does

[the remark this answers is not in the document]
brushmate: Well, for some reason ESLint does not report this. But the configuration extends [standard]{https://github.com/standard/eslint-config-standard/blob/master/.eslintrc.json}, which defines `"object-curly-spacing": ["error", "always"]`. This should result in a linting error. Check this [ESLint playground example](https://eslint.org/play/#eyJ0ZXh0IjoiLyogZ2xvYmFsIGNvbnNvbGUgKi9cbi8qIGVzbGludCBvYmplY3QtY3VybHktc3BhY2luZzogW1wiZXJyb3JcIiwgXCJhbHdheXNcIl0gKi9cbmltcG9ydCB7Zm9vfSBmcm9tICdiYXInXG5cbmNvbnNvbGUubG9nKGZvbykiLCJvcHRpb25zIjp7InBhcnNlck9wdGlvbnMiOnsiZWNtYVZlcnNpb24iOiJsYXRlc3QiLCJzb3VyY2VUeXBlIjoibW9kdWxlIiwiZWNtYUZlYXR1cmVzIjp7fX0sInJ1bGVzIjp7fSwiZW52Ijp7ImVzNiI6dHJ1ZX19fQ==) for reference.

When looking into the other adapter you can find both variants. I still prefer the 'spaced' variant. In the end it's up to the Prebid.js maintainers.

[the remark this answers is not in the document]
nkloeber: Thanks for the reference I updated it.

brushmate (approved):

ncolletti (approved):
lgtm

Issue #8424, opened 2022-05-17
Support for GPC


patmmccann:
We should set ortb2.regs.ext.gpc to 1 if GPC headers indicate the user has opted out, following https://github.com/prebid/Prebid.js/issues/6270#issuecomment-780896180

 Some guidance suggests the third character of the [US Privacy string](https://github.com/InteractiveAdvertisingBureau/USPrivacy/blob/master/CCPA/US%20Privacy%20String.md) should be set to 'Y'. Publishers or bid adapters are free to modify the string set by the cmp to effectuate this or entities may choose to resolve conflicts server side as suggested in  https://github.com/InteractiveAdvertisingBureau/openrtb/issues/98

patmmccann:
#7796 for linkage

PR #8925, opened 2022-08-30
Enrichment FPD Module: Support for GPC Detection

jlquaccia:
## Type of change
- [x] Feature

## Description of change
- Added a function within the Enrichment FPD Module to detect whether or not GPC is enabled.  If it is enabled, then ortb2.regs.ext.gpc will be set to 1.

## Other information
https://github.com/prebid/Prebid.js/issues/8424
https://global-privacy-control.glitch.me

@dgirardi can you review when you have time?

### DIFF
--- a/test/spec/modules/enrichmentFpdModule_spec.js
+++ b/test/spec/modules/enrichmentFpdModule_spec.js
@@ -99,4 +99,18 @@ describe('the first party data enrichment module', function() {
     expect(validated.device).to.deep.equal({ w: 1200, h: 700 });
     expect(validated.site.keywords).to.be.undefined;
   });
+
+  it('test 2', function() {
+    let validated;
+    width = 800;
+    height = 500;
+
+    validated = processFpd({}, {}).global;
+    expect(validated.regs).to.equal(undefined);
+
+    navigator.globalPrivacyControl = { gpc: 1 };
+    validated = processFpd({}, {}).global;
+    expect(validated.regs.ext.gpc).to.equal(1);
+    delete navigator.globalPrivacyControl;
+  });
 });
--- a/test/spec/modules/enrichmentFpdModule_spec.js
+++ b/test/spec/modules/enrichmentFpdModule_spec.js
@@ -100,7 +100,7 @@ describe('the first party data enrichment module', function() {
     expect(validated.site.keywords).to.be.undefined;
   });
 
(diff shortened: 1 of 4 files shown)


test/spec/modules/enrichmentFpdModule_spec.js
```diff
@@ -99,4 +99,18 @@ describe('the first party data enrichment module', function() {
...
+
+  it('should store a reference to gpc witin ortb2.regs.ext if it has been enabled', function() {
+    let validated;
+    width = 800;
+    height = 500;
+
+    validated = processFpd({}, {}).global;
+    expect(validated.regs).to.equal(undefined);
+
+    navigator.globalPrivacyControl = { gpc: 1 };
```
dgirardi, line 111: as I read the proposal this should be `true` or `false`. However, I have two objections: 

-  aren't we doing this too early? none of my browsers set this, and it's not even mentioned on MDN.  Is the proposal final and on its way to implementation?
- Once the navigator _does_ set this, this test will not work right, because it's going to be a read-only value. I think you need to mock it here.

jlquaccia: just updated my PR.  somehow missed the `true` or `false` part thanks for pointing that out.  i think the test is ok now.

correct, from what i gathered as well, this functionality does still seem to be in it's early stages, not 100% sure when it will be fully implemented.  it does state here how detection will occur though: https://global-privacy-control.glitch.me/ (was taken from https://globalprivacycontrol.org/, from the faq at the bottom under "I’m a publisher, developer, or other service. How can I support GPC?")

i was aiming to be proactive, however, as you mentioned because this is still in the proposal phase should we hold off entirely for the time being in case specifications change?

dgirardi: I think it's OK because as it's written the worst that can happen is that we don't pick up the gpc flag. I'm just surprised that we'd get here before MDN :)

jlquaccia: haha i agree, MDN typically knows all.  least the change on our end for this is pretty minimal/low-risk like you pointed out anyway

dgirardi (approved):

patmmccann (approved):

patmmccann:
Thanks @jlquaccia !!

PR #8933, opened 2022-08-31
SpotX Bid Adapter: Support UID2

spotxslagle:
## Type of change
- [ x ] Feature

## Description of change
This change adds support for passing along UID2 id's in SpotX bid requests

## Other information
@robertrmartinez @dgirardi @patmmccann


modules/spotxBidAdapter.js
```diff
@@ -300,6 +300,21 @@ export const spec = {
         );
       }
 
+      if (bid && bid.userId && bid.userId.uid2) {
```
dgirardi, line 303: I may be missing some context, but the bid request has a `userIdAsEids` property, an array of all available user IDs formatted as ORTB `user.ext.eids`. I believe you could just copy that over instead of checking for individual id providers? although I don't know if that would set `ext.rtiPartner` as you do here.

spotxslagle: I'm honestly not sure about that. Our backend might be fine with that, but I'm not the person to ask about it. The current implementation is the one requested by our product team, and I'd prefer not to fiddle around with it unless I need to.

dgirardi: Understood, but my suggestion is to bring this up with your product team - because if that works (and I don't see why it wouldn't) that means that the next ID provider you onboard with won't need any changes at all in your adapter.

dgirardi (approved):

modules/spotxBidAdapter.js
patmmccann: the spotx adapter appears to be doing some bulky re-implementation of creating an eid. Please fix

patmmccann (changes requested):

spotxslagle:
I believe everything should be in order with this now

patmmccann (approved):

Issue #8660, opened 2022-07-07
bid adapters: support standard first party data keywords


bretg:
## Type of issue
conforming to standards

## Description
While reviewing a documentation PR from (@gchicoye), I noticed that the nexx360 is accepting a 'keywords' parameter and doesn't look for them in ortb2.site.keywords.

It's not acceptable for publishers to have to pass adapters special first party data -- rather, they should be able to set [standard FPD keywords](https://docs.prebid.org/features/firstPartyData.html) and all adapters must read from that conventional location. It's acceptable to support **both** custom params as well as the conventional location.

Adapters that need to be checked and possibly updated:

- [x] Nexx360
- [x] Appnexus

@ChrisHuie found these additional bidders that need to look into this:

- [ ] Craft Bid Adapter
- [ ] Winr Bid Adapter
- [ ] Pixfuture Bid Adapter
- [ ] Mediafuse Bid Adapter
- [ ] Beop Bid Adapter
- [ ] Displayio Bid Adapter
- [ ] Prisma Bid Adapter
- [ ] Ras Bid Adapter
- [ ] Goldbach Bid Adapter
- [ ] Deepintent Bid Adapter
- [ ] Sonobi Bid Adapter

patmmccann:
@bretg does this apply to Xandr adapter?

patmmccann:
https://github.com/prebid/Prebid.js/blob/15b688c9a4f15a7669fd1fa579f9c76e1a917949/modules/appnexusBidAdapter.js#L846

bretg:
It applies to all bidders. @jsnellbaker - please take a look at the appnexus bid adapter to make sure keywords can come from ortb2 and not just bidder-specific params. Thanks!

jsnellbaker:
To clarify the intent/context, when a publisher specifies ortb2 keywords - they intend to have them apply to all adUnits that were part of the setup?  It's not a selective thing per adUnit?

I figure it's the former, but I just want to check in case there are other possible use-cases within the ortb2 setup.

bretg:
Not clear imp-level keywords is a real use-case, but if it is, seems like AdUnit.ortb2Imp.ext.data.keywords would be the right location.

ORTB specifies site.keywords, app.keywords, user.keywords, and site.content.keywords -- those should be set with setConfig({ortb2}).

jsnellbaker:
Hi @bretg 

Thanks for the feedback.  We will take a look into these changes for our next sprint.

[...]

PR #8614, opened 2022-06-28
1plusX RTD submodule: New RTD Module

bwhisp:
## Type of change

- [ ] Bugfix
- [x] Feature (New RTD submodule)
- [ ] New bidder adapter   
- [ ] Code style update (formatting, local variables)
- [ ] Refactoring (no functional changes, no api changes)
- [ ] Build related changes
- [ ] CI related changes
- [ ] Does this change affect user-facing APIs or examples documented on http://prebid.org?
- [ ] Other

## Description of change

This PR adds a RTD submodule for 1plusX 
This module will fetch first party data from 1plusX's Profile API consisting of : 
  - the audiences the user is part of 
  - the topics related to the page
 
Then it will transmit this data to the supported & selected bid adapters through `setBidderConfig`

Contents of the PR : 
    - 1plusX RTD module: `modules/1plusXRtdProvider.js`
    - Unit tests for 1plusX RTD module: `test/spec/modules/1plusXRtdProvider_spec.js`
    - Integration example for 1plusX RTD module: `integrationExamples/gpt/1plusXRtdProviderExample.html`
    - Documentation for 1plusX RTD module: `modules/1plusXRtdProvider.md`

For any changes that affect user-facing APIs or example code documented on http://prebid.org, please provide:

- PR adding the documentation : prebid/prebid.github.io#3873

## Other information
I am not 100% sure about where to write the data in the ORTB2 format when updating the bid adapter configs.

The data I am fetching from 1plusX's Profile API consists of lists of custom-defined Ids that don't follow the standard taxonomy. 
While I found info about where to write audience segments & topics in ORTB2; it seemed that it concerned only those following the standard taxonomy. 

I reached a solution by combining the [ORTB2 specs](https://www.iab.com/wp-content/uploads/2016/03/OpenRTB-API-Specification-Version-2-5-FINAL.pdf) (pages 31 & 25) AND some examples from other RTD Providers : 
   - audience segments will be written in an element of `ortb2.user.data` array
   - topics will be written in `ortb2.site.keywords`

I will add comments to the related parts of the code, please review these parts with special care.

PR #8939, opened 2022-09-01
Appnexus Bid Adapter: add support to read ortb2 keywords

jsnellbaker:
## Type of change

- [x] Feature

## Description of change
Related to #8660

This PR adds support to read keywords from the various ortb2 fields (in the ortb2 config and the potential adUnit.ortb2Imp object.

To avoid this as a breaking change, the previous methods used by the appnexus adapter to read keyword data are still supported.  The keyword keys/values are combined from the respective sources and are assigned to the proper locations in the outgoing adserver request (ie adUnit level keywords are stored inside the `tag` object that matches the `adUnit` and the generic ortb2 keywords are applied at the request level).

When the keys from different sources match (ie bid params and adUnit.ortb2Imp), the values are joined together in the final result to help consolidate the information and ensure all values are considered.

### DIFF
--- a/test/spec/modules/appnexusBidAdapter_spec.js
+++ b/test/spec/modules/appnexusBidAdapter_spec.js
@@ -610,7 +610,7 @@ describe('AppNexusAdapter', function () {
       config.getConfig.restore();
     });
 
-    it('adds auction level keywords to request when set', function () {
+    it('adds auction level keywords and ortb2 keywords to request when set', function () {
       let bidRequest = Object.assign({}, bidRequests[0]);
       sinon
         .stub(config, 'getConfig')
@@ -618,10 +618,31 @@ describe('AppNexusAdapter', function () {
         .returns({
           gender: 'm',
           music: ['rock', 'pop'],
-          test: ''
+          test: '',
+          tools: 'power'
         });
 
-      const request = spec.buildRequests([bidRequest]);
+      const bidderRequest = {
+        ortb2: {
+          site: {
+            keywords: 'power tools, drills, tools=industrial',
+            content: {
+              keywords: 'video, source=streaming'
+            }
+          },
+          user: {
+            keywords: 'tools=home,renting'
+          },
+          app: {
+            keywords: 'app=iphone 11',
+            content: {
+              keywords: 'appcontent=home repair, dyi'
+            }
+          }
+        }
+      };
+
+      const request = spec.buildRequests([bidRequest], bidderRequest);
       const payload = JSON.parse(request.data);
 
       expect(payload.keywords).to.deep.equal([{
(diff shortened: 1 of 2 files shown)


patmmccann:
tagging #8596 as related. I believe most of the special handling for Appnexus that RTB modules do is in setting these parameters.

test/spec/modules/appnexusBidAdapter_spec.js
```diff
@@ -632,6 +653,28 @@ describe('AppNexusAdapter', function () {
...
+      }, {
+        'key': 'renting'
+      }, {
+        'key': 'app',
+        'value': ['iphone 11']
+      }, {
+        'key': 'appcontent',
+        'value': ['home repair']
+      }, {
+        'key': 'dyi'
```
patmmccann, line 677: would you consider gathering this https://github.com/prebid/Prebid.js/blob/d73944600b1e15a5a1f5c4ce5c2db8a84eb53644/modules/permutiveRtdProvider.js#L231 and https://github.com/prebid/Prebid.js/blob/d73944600b1e15a5a1f5c4ce5c2db8a84eb53644/modules/1plusXRtdProvider.js#L108 and https://github.com/prebid/Prebid.js/blob/d73944600b1e15a5a1f5c4ce5c2db8a84eb53644/modules/airgridRtdProvider.js#L76 in scope for this change?

patmmccann: these rtd modules bypass the "normal" workflow that the data controller module is able to edit

jsnellbaker: If they're populating the values under the bid.params.keywords before our request goes out, it should be gathered automatically by the logic we have already?  I'm not sure how else to pre-emptively gather the data from their systems.

If they want to change their setup to have the publisher populate the ortb2 keyword fields with their data, then I suppose their customizations shouldn't be needed (if that's what you meant by 'in scope for this change').

patmmccann: These rtd modules typically write contextual segments to site.content.data and user segments to user.data. It is in those locations that the data controller module is able to edit the segments down if necessary. We believe, since the appnexus adapter does not support either of these inputs and people seem to be using params.keywords in their place, that appnexus adapter should be responsible for copying the segments from the standard locations to the keywords instead of these modules each having special appnexus handling.

patmmccann: as reference: https://github.com/prebid/Prebid.js/pull/8614#pullrequestreview-1035918465

jsnellbaker: How would you transform the typical segment objects into key-value pairs (I'm looking at the samples in the (first party data)[https://docs.prebid.org/features/firstPartyData.html]) page from the docs, and I don't see exactly how things should be pulled together)?  

Is this really something that would okay to implement for everyone who potentially has this data in these fields, as opposed to just doing it for these rtd modules?

patmmccann: I agree it isn't obvious how this should be done, and each of these adapters seems to have made slightly different choices to support Xandr

jsnellbaker: If someone change suggest a standard approach, we can take a look and consider.  If it's going to take time, then maybe it can be a future PR and we just have this PR focus on the keywords fields specifically.

patmmccann (approved):

Fawke (approved):

Issue #8741, opened 2022-07-26
Enhancements to Topics module


patmmccann:
With the release of the topics module, https://github.com/prebid/Prebid.js/pull/8646 two enhancements immediately come to mind:

Calling the api within a frame for an entity with a larger footprint (adnxs.com, adsrvr.com, rubicon-project.com, pubmatic.com, openx.com, and other leader teir companies, as well as sharedid.org, come to mind), possibly rotating between entities or being subject to config.

Ability to specify or override user.data.name, eg cafemedia or pubmatic or hearst, if the domain calling the api is getting results over a wider footprint than pubdomain.com 

Finally, the module could, if so configured, write topics to local storage for later access, as well as set targeting to GAM.

patmmccann:
From: https://developer.chrome.com/en/docs/privacy-sandbox/topics/

"API callers only receive topics they've observed
A design goal of the Topics API is to enable interest-based advertising without the sharing of information to more entities than is currently possible with third-party cookies. The Topics API proposes that topics can only be returned for API callers that have already observed them, within a limited timeframe.

Key Term
A Topics API caller is the entity that calls the document.browsingTopics() JavaScript method, and will use the topics returned by the method to help select relevant ads. Typically, a call to document.browsingTopics() would be from code included in a site from a third party such as an adtech platform. The browser determines the caller from the site of the current document. So, if you're a third party on a page, make sure you call the API from an iframe that your site owns.

In order for document.browsingTopics()to return one or more topics, it must be called in code from the same origin as code that was on a site where those topics were observed."

patmmccann:
Discussion in the topics committee today indicated: an ssp rotation works, and that should be able to be supplemented by config. We also should put the contents of the frame in another repo.

dmdabbs:
Thanks @patmmccann. Sorry I was unable to make that discussion.
Sharedid.org would also be in rotation (subject to publisher control)?

nitin0610:
@patmmccann We (PubMatic) are interested to work on this enhancement. 
Can you help us to give more insights on SSP rotations works and config will look like to work on the solution for the same.

nitin0610:
@patmmccann  Current representation of Topics in OpenRTB protocol is in **BidRequest.user.data** like in below format specified in 
https://github.com/google/ads-privacy/tree/master/proposals/topics-rtb. 

Since this is the same spec which we use for other SDA purpose and it goes under bidrequest.user.data object. 

[...]

PR #8947, opened 2022-09-05
Topics module: Initial Topics iframe implementation

nitin0610:
## Type of change

- [ ] Bugfix
- [X] Feature
- [ ] New bidder adapter   
- [ ] Build related changes
- [ ] CI related changes

- [ ] Other

## Description of change

This change is the proposal of iframe implementation which is the enhancements of existing module under topicsFpdModule.js where different bidders will call the topic API under their domain to fetch the topics for respective domain and the segment data will be part of ORTB request under user.data object.

Enhancements are discussed in this thread https://github.com/prebid/Prebid.js/issues/8741

**Note**: This is rough draft and initial proposal of iframe implementation to discuss about the change.

### Implementation

- Configured topics under userSync object to read the iframeURL for particular bidder on publisher domain. This would override the default configuration. If these configurations are not present on the page, default configuration of bidders iframe which will call topics API. Default SSP rotation is kept as 1 as default configuration which can be override by adding configuration inside topics object having "maxTopicCaller" as key on page as shown below configuration examples

Note: IframeURL for each domain is dummy URL used for to give showcase how to configure the URL
```
pbjs.setConfig({
            userSync: {
                ...,
                topics: { 
                        maxTopicCaller: 3, // SSP rotation 
                        bidders: [{
                            bidder: 'pubmatic',
                            iframeURL: 'https://pubmatic.com:8080/topics/fpd/topic.html', // dummy URL
                            expiry: 7 // Configurable expiry days
                        },{
                            bidder: 'rubicon',
                            iframeURL: 'https://rubicon.com:8080/topics/fpd/topic.html', // dummy URL
                            expiry: 7 // Configurable expiry days
                        },{
                            bidder: 'appnexus',
                            iframeURL: 'https://appnexus.com:8080/topics/fpd/topic.html', // dummy URL
                            expiry: 7 // Configurable expiry days
                        }]
                    }
                ....
            }
})
```

- Topics data will be loaded for particular domain from iframe and sent back to first party domain or publisher website. This data will be cached in localstorage to make it available for bidders to consume in their APIs

- Data will be cached in 'prebid:topics'

- This Data will be read from cache(localstorage) and will be part of bidder's respective API under user.data object
For Eg:

```
"user":{
  "gender":"M",
  "geo":{
     "lat":40.712775,
     "lon":-74.005973
  },
  "yob":1982,
  "data":[
     {
        "ext":{
           "segtax":600,
           "segclass":"2206021246"
        },
        "segment":[
           {
              "id":"45"
           }
        ],
        "name":"timesofindia.com"
     },
     {
        "ext":{
           "segtax":600,
           "segclass":"2206021246"
        },
        "segment":[
           {
              "id":"265"
           },
           {
              "id":"45"
           }
        ],
        "name":"ads.pubmatic.com"
     }
  ]
},


```
### Open questions

1.  How SSP Rotations works ? 

- Does it means we need to call any random 3 bidders to load iframe and make topics data for that domain/api caller to make it available for bidder request. 

- What happens if we configured more than 3 bidders (for Eg: 7-8) which will load iframe for 7-8 bidders which might cause latency and concerning for publisher or first party domain.

2.  Expiry date for cached data should be added for each bidder or 1 expiry for all bidders? Should publisher be given control to refresh the cached data by setting the expiry

3.  How iframe should be maintained for each bidder? Should it be in this repository itself ?
4.  Cached Topics data in local-storage needs to be encrypted (Base64) ?


nitin0610:
@patmmccann Can you please help us review the changes and if suggestions/comments regarding the implementation of iframe. 
Note: This is WIP PR so that we can discuss about the proposal/approach if it looks good to go ahead.

modules/topicsFpdModule.js
```diff
@@ -73,6 +100,54 @@ export function processFpd(config, {global}, {data = topicsData} = {}) {
...
+      ifrm.style.display = 'none';
+      window.document.documentElement.appendChild(ifrm);
+    });
+  } else {
+    logWarn(`Topics config not defined under userSync Object`);
+  }
+}
+
+// Exposing loadTopicsForBidders function in global-name-space so that API calling from domain can get the data and send it in oRTB format.
+(getGlobal()).loadTopicsForBidders = loadTopicsForBidders;
```
dgirardi, line 149: Is this needed? If we are treating this the way we treat user syncs, I would just run them together. From the publisher's POV calling this seems like boilerplate.

modules/topicsFpdModule.js
```diff
@@ -73,6 +100,54 @@ export function processFpd(config, {global}, {data = topicsData} = {}) {
...
+      try {
+        let data = JSON.parse(e.data);
+        if (data && data.segment && !isEmpty(data.segment.topics)) {
+          let {domain, topics, bidder} = data.segment;
+          let segmentData = getTopicsData(domain, topics);
+          let updateStore = [];
+          let bidderFoundInStore = false;
+          let storedSegments = JSON.parse(storage.getDataFromLocalStorage(`${initialTopicName}${bidder}`));
+          storedSegments?.forEach((segment) => {
+            if (segment.name === segmentData[0].name && segment.ext.segclass == segmentData[0].ext.segclass) {
```
dgirardi, line 116: This means that topics can never be removed from the list. I have no opinion on whether they should, but it does seem to me that this is working around the purpose of the topics API (limiting how much privacy sensitive data is moved around).

If we want to keep each topic forever, I do think we should use a map to store them here - looping through all of them can get quite expensive since the list size can only increase.

patmmccann:
`Open questions
How SSP Rotations works ?
Does it means we need to call any random 3 bidders to load iframe and make topics data for that domain/api caller to make it available for bidder request.

What happens if we configured more than 3 bidders (for Eg: 7-8) which will load iframe for 7-8 bidders which might cause latency and concerning for publisher or first party domain.

Expiry date for cached data should be added for each bidder or 1 expiry for all bidders? Should publisher be given control to refresh the cached data by setting the expiry

How iframe should be maintained for each bidder? Should it be in this repository itself ?

Cached Topics data in local-storage needs to be encrypted (Base64) ?`

I think SSP rotation would only call one of the frames at random per pageview

Expiration should be fairly aggressive I think; as the topics API itself is fairly aggressive with expiration. 

I think the iframes in the default rotation should be in the repo likely; and publisher could override the array with their own array in config. 

I am not sure if we need to encrypt anything, the topics are meant to be passed around freely it seems. I am not sure of any security risk from exposing them.

nitin0610:
> > Open questions
> > How SSP Rotations works ?
> > Does it means we need to call any random 3 bidders to load iframe and make topics data for that domain/api caller to make it >available for bidder request.
> 
> > What happens if we configured more than 3 bidders (for Eg: 7-8) which will load iframe for 7-8 bidders which might cause >latency and concerning for publisher or first party domain.
> 
> I think SSP rotation would only call one of the frames at random per pageview; maybe with a 'callAll' override?
> Regarding "I think SSP rotation would only call one of the frames at random per page view; maybe with a 'callAll' override?"

Regarding SSP rotation, Does adding config ("maxTopicsCaller": 3) inside the setConfig under topics make sense which gives more power to publishers to provide count of bidders from where it get the topics from those bidders domain ? 
This is configurable as suggested,  this can even configure as 1 (Bidder).

For Eg. Over here, Adding 2 as count is configured for maxTopicsCaller and  3 bidders has been configured but only random 2 bidders will call the frames. If this "maxTopicsCaller" property is missing all together in config, it will load the frames for all 3 bidders ( as replacement of the one what was suggested with callAll )

```
pbjs.setConfig({
            userSync: {
                ...,
                topics: { 
                        bidders: {
                           "maxTopicsCaller": 2,
                            "pubmatic" : {
                                iframeURL: "https://ads.pubmatic.com/topics/fpd/topic-pubmatic.html" // Will be hosted on pubmatic domain
                            },
                            "rubicon" : {
                                iframeURL: "https://rubicon.com:8080/topics/fpd/topic-rubicon.html" // Will be hosted on rubicon domain
                            },
                            "appnexus" : {
                                iframeURL: "https://appnexus.com:8080/topics/fpd/topic-appnexus.html" // Will be hosted on appnexus domain
                            }
                        }
                    }
                ....
            }
})
```
Note: Also, this is being said in case of SSP rotation, random iframes of third party API caller will be loaded, but cached topics data of third party api caller will be available in user.data object(SDA data) rather than only topic data loaded for iframe for that page view at that particular time. 

@patmmccann  Please let me know your view on the config and above note if any?

> > Expiry date for cached data should be added for each bidder or 1 expiry for all bidders? Should publisher be given control to >refresh the cached data by setting the expiry
> 
> Expiration should be fairly aggressive I think; as the topics API itself is fairly aggressive with expiration.
> 
> > How iframe should be maintained for each bidder? Should it be in this repository itself ?
> 
> I think the iframes in the default rotation should be in the repo likely; and publisher could override the array with their own array in config.
> 

@patmmccann Can you help to give more details on this ? Thanks
> > Cached Topics data in local-storage needs to be encrypted (Base64) ?`
> 
> I am not sure if we need to encrypt anything, the topics are meant to be passed around freely it seems. I am not sure of any security risk from exposing them.

modules/topicsFpdModule.js
```diff
@@ -58,10 +64,31 @@ export function getTopics(doc = document) {
   return topics;
 }
 
+// function to fetch the cached topic data from storage for bidders and return it
+export function getTpsForBidderFromStorage() {
+  let cachedTopicData = [];
+  const topics = config.getConfig('userSync.topics');
+  let biddersList = Object.keys(topics.bidders || []);
+  Object.keys(window.localStorage).forEach((storeItem) => {
```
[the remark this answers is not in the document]
dgirardi: With the strategy you are using here, I would store a single map where the key is  `` `${segment.name}:${segment.ext.segclass}` `` (because I believe `segment.name` would be an alias of some kind for the bidder? otherwise some other combination that would uniquely identify the segment).

burk504:
>>Expiry date for cached data should be added for each bidder or 1 expiry for all bidders? Should publisher be given control to >refresh the cached data by setting the expiry

>Expiration should be fairly aggressive I think; as the topics API itself is fairly aggressive with expiration.

My opinion is that Topics are designed to have a 3 week lifespan by definition and that should be the default life of them here.  That said, I think a publisher should have the capability to change that timeframe easily

nitin0610:
> > > Expiry date for cached data should be added for each bidder or 1 expiry for all bidders? Should publisher be given control to >refresh the cached data by setting the expiry
> 
> > Expiration should be fairly aggressive I think; as the topics API itself is fairly aggressive with expiration.
> 
> My opinion is that Topics are designed to have a 3 week lifespan by definition and that should be the default life of them here. That said, I think a publisher should have the capability to change that timeframe easily.
> 
> This also means that each iframe should have a chance to be fired once every 7 days. That way each iframe has a rolling set of topics they have been added to storage. On the 4th 7 day period, the iframe would call again and have the older one from 4 weeks ago get deleted...leaving the storage with a rolling 3 weeks of data from each individual vendor iframe

@patmmccann  @dgirardi Any comments on the suggestion given by @burk504 ?

patmmccann:
Jeff's comments sound reasonable to me; I think the more configurable the choices, the better off we are. Prebid's goals are to discover the utility of this interface and to allow publishers and community members to participate in the feedback the browser folks are seeking.

modules/topicsFpdModule.js
```diff
@@ -73,6 +100,54 @@ export function processFpd(config, {global}, {data = topicsData} = {}) {
...
+    }
+  });
+}
+
+function loadTopicsForBidders() {
+  addListenerToFetchTopics();
+  const topics = config.getConfig('userSync.topics');
+  if (topics) {
+    const bidders = Object.keys(topics.bidders || []);
+    bidders?.forEach((bidder) => {
```
patmmccann, line 136: this generates a linting error

modules/topicsFpdModule.js
```diff
@@ -73,6 +100,54 @@ export function processFpd(config, {global}, {data = topicsData} = {}) {
...
+    if (e && e.data) {
+      try {
+        let data = JSON.parse(e.data);
+        if (data && data.segment && !isEmpty(data.segment.topics)) {
+          let {domain, topics, bidder} = data.segment;
+          let segmentData = getTopicsData(domain, topics);
+          let updateStore = [];
+          let bidderFoundInStore = false;
+          let storedSegments = JSON.parse(storage.getDataFromLocalStorage(`${initialTopicName}${bidder}`));
+          storedSegments?.forEach((segment) => {
```
patmmccann, line 115: this generates a linting error

modules/topicsFpdModule.js
```diff
@@ -73,6 +100,54 @@ export function processFpd(config, {global}, {data = topicsData} = {}) {
...
+  window.addEventListener('message', (e) => {
+    if (e && e.data) {
+      try {
+        let data = JSON.parse(e.data);
+        if (data && data.segment && !isEmpty(data.segment.topics)) {
+          let {domain, topics, bidder} = data.segment;
+          let segmentData = getTopicsData(domain, topics);
+          let updateStore = [];
+          let bidderFoundInStore = false;
+          let storedSegments = JSON.parse(storage.getDataFromLocalStorage(`${initialTopicName}${bidder}`));
```
patmmccann, line 114: please use https://github.com/prebid/Prebid.js/blob/e0ef967669ff5d4d67ff695b5326c5540ffb146b/src/utils.js#L1374

modules/topicsFpdModule.js
```diff
@@ -73,6 +100,54 @@ export function processFpd(config, {global}, {data = topicsData} = {}) {
   });
 }
 
+// Added function addListenerToFetchTopics to listen for topic data retured from iframe loaded for SSPs(bidder)
+function addListenerToFetchTopics() {
+  window.addEventListener('message', (e) => {
+    if (e && e.data) {
+      try {
+        let data = JSON.parse(e.data);
```
patmmccann, line 108: please use https://github.com/prebid/Prebid.js/blob/e0ef967669ff5d4d67ff695b5326c5540ffb146b/src/utils.js#L1374

modules/topicsFpdModule.js
nitin0610: Removed exposing the function to global. It will be called on when topics module FPD data is processed.

modules/topicsFpdModule.js
```diff
@@ -73,6 +100,54 @@ export function processFpd(config, {global}, {data = topicsData} = {}) {
   });
 }
 
+// Added function addListenerToFetchTopics to listen for topic data retured from iframe loaded for SSPs(bidder)
+function addListenerToFetchTopics() {
+  window.addEventListener('message', (e) => {
```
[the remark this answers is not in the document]
nitin0610: Added condition by checking if message received is from loaded iframe URL in latest commit.

modules/topicsFpdModule.js
nitin0610: Fixed Linting Issue

modules/topicsFpdModule.js
nitin0610: Fixed lint issues.

modules/topicsFpdModule.js
nitin0610: Using 'safeJSONParse' function to parse the json data.

modules/topicsFpdModule.js
nitin0610: Using 'safeJSONParse' function to parse the json data

abhinavsinha001:
This is what I understand and recommend for basic implementation:
1. We should cache and send only latest topics, the caching is just to reduce Iframe based communication overhead
2. In case any of the caller is not called for more than 3 weeks then it's cached topics should expire in 3 weeks or publishers can define a custom expiry.
3. If we want to cache topics over a dynamic rolling window then lets take that as separate PR 
  i. Will need to set expectations throughout the pipe that we are sending extra topics.
 ii. Will have to decide on ordering / notifying which topics are recent vs cached.

nitin0610:
@patmmccann @dgirardi  I have added the changes which Abhinav have suggested. 

1. Caching the topics for specific for default 3 weeks (21 days). Meaning if topics are older than 21 days, it will be deleted and no longer will be sent in oRTB request. Configuration maintain in topicsFPDModule.

2. Publisher can override the cache expiry time from 21 days to custom day to send the recent topics. Please find below configuration which needs to be on the page.
```
pbjs.setConfig({
            userSync: {
                ...,
                topics: { 
                        maxTopicCaller: 3, // SSP rotation 
                        bidders: [{
                            bidder: 'pubmatic',
                            iframeURL: 'https://pubmatic.com:8080/topics/fpd/topic.html', // dummy URL
                            expiry: 7 // Configurable expiry days
                        },{
                            bidder: 'rubicon',
                            iframeURL: 'https://rubicon.com:8080/topics/fpd/topic.html', // dummy URL
                            expiry: 7 // Configurable expiry days
                        },{
                            bidder: 'appnexus',
                            iframeURL: 'https://appnexus.com:8080/topics/fpd/topic.html', // dummy URL
                            expiry: 7 // Configurable expiry days
                        }]
                    }
                ....
            }
})
```
where expiry and maxTopicCaller( Numbers of iframes count which will be loaded ) can be configured which will be override the default configuration set in the topicsFpdModule.

I have also added topics_iframe.html which will be hosted on bidders domain which will call the Topics API just for reference.

Can you guys help us to review the changes and suggest if somethings need to change or needs to be added.

patmmccann:
@nitin0610 is this ready to come out of draft status?

nitin0610:
> @nitin0610 is this ready to come out of draft status?

@patmmccann @dgirardi  
Just wanted to check if the code looks good after confirmation we will work on the PR to make it available for merge by removing the draft.
Anything which you can suggest need changes in terms of business logic and code.
Let us(PubMatic) know if you have questions for the same.

integrationExamples/gpt/topics_frame.html
```diff
@@ -0,0 +1,43 @@
...
+        const params = new URLSearchParams(url.search);
+        const message = JSON.stringify({
+          segment: {
+            domain: window.location.hostname,
+            topics,
+            bidder: params.get("bidder")
+          },
+          date: Date.now(),
+        });
+        window.parent.postMessage(message, '*');
```
dgirardi, line 37: I'd prefer if this was passed the origin to use from Prebid, instead of `'*'`. If this data is sensitive then it's not a preference but a requirement, I don't know if that's the case.

modules/topicsFpdModule.js
```diff
@@ -1,13 +1,31 @@
...
   '1': 600
 }
 
+const DEFAULT_EXPIRATION_DAYS = 21;
+
+const bidderIframeList = {
+  maxTopicCaller: 1,
+  bidders: [{
+    bidder: 'pubmatic',
+    iframeURL: 'https://pubmatic.com:8080/topics/fpd/topic.html' // dummy URL for NOW
```
dgirardi, line 25: is the dummy data useful here? might be better to leave this empty and pass the test urls via config.

modules/topicsFpdModule.js
```diff
@@ -61,7 +93,9 @@ export function getTopics(doc = document) {
 const topicsData = getTopics().then((topics) => getTopicsData(getRefererInfo().domain, topics));
 
 export function processFpd(config, {global}, {data = topicsData} = {}) {
+  loadTopicsForBidders();
```
dgirardi, line 96: this runs on each auction, and will attach a new message listener each time.

dgirardi (commented):
Please add unit tests; I'd especially like to see the expiration logic tested. Also, how should this be treated for GDPR purpose 1? The implementation as it is will not require any vendor consent, is that appropriate?

patmmccann:
Seems like we need Purpose 3 & purpose 4 consent to call the topics api in Europe and calling the api constitutes a data sale in California, so we should check for an opt out prior

nitin0610:
> Seems like we need Purpose 3 & purpose 4 consent to call the topics api in Europe and calling the api constitutes a data sale in California, so we should check for an opt out prior. In europe those consents need to be to the iframe host.

@patmmccann  Can you throw some light if we have added Purpose 3 & purpose 4 consent condition for any other scenarios or feature for reference.

patmmccann:
See for example here

https://github.com/prebid/Prebid.js/blob/6e36e363aa843603a8b7cb29da3242d84e5f4b9d/modules/quantcastIdSystem.js#L94

patmmccann:
Purpose 3 is building a personalized ads profile, that's what the topics api does

modules/topicsFpdModule.js
nitin0610: Added Check @dgirardi .

integrationExamples/gpt/topics_frame.html
nitin0610: @dgirardi I have taken care of it in Prebid topic Module where i have checked if origin is from Iframe which was loaded
Added Below check when listening for topics data
```
if (includes(getLoadedIframeURL(), evt.origin) && data && data.segment && !isEmpty(data.segment.topics)) { ... }
```

Please let us know if something i am missing ?

modules/topicsFpdModule.js
```diff
@@ -73,6 +114,145 @@ export function processFpd(config, {global}, {data = topicsData} = {}) {
...
+    return checkTCFv2(gdprConsent.vendorData);
+  }
+  return true;
+}
+
+/**
+ * function to load the iframes of the bidder to load the topics data
+ */
+function loadTopicsForBidders() {
+  HAS_GDPR_CONSENT = hasGDPRConsent();
```
nitin0610, line 232: @patmmccann  Can you please review the consent logic. 
I have handled specifically Purpose [1,2,3,4] consent. If one of the consent flag is false, we don't load Iframe for any of the bidders.

We have added **topicsFpd** of 'core' module type which is neither Vendor or user id module. Same implementation as we have for **enrichmentFpd** .
```
const MODULE_NAME = 'topicsFpd';
export const coreStorage = getCoreStorageManager(MODULE_NAME);
```

patmmccann: Thanks!!

dmarti: Does this also check CCPA? (need to make sure not to call if the user is opted out of "sale")

patmmccann: Nothing in prebid checks that; publishers are currently responsible for ccpa enforcement

patmmccann (changes requested):
Please make sure the module respects the device access control flag which many publishers set in opt out scenarios

modules/topicsFpdModule.js
```diff
@@ -73,6 +114,145 @@ export function processFpd(config, {global}, {data = topicsData} = {}) {
...
+  if (!HAS_GDPR_CONSENT) {
+    logInfo('Topics Module : Consent string is required to fetch the topics from third party domains.');
+    return;
+  }
```
patmmccann, line 236: righty here can you use this function to check for dev access https://github.com/prebid/Prebid.js/blob/21712ca1f99e573c51d217ddcc4a4f08273dfb9d/src/utils.js#L820

nitin0610: @patmmccann Device access check added.

modules/topicsFpdModule.js
nitin0610: @patmmccann Device access check added. 
My understanding is we might dont need to add CCPAconsent check if deviceAccess flag is set for true/false.

patmmccann: Thanks!

patmmccann (approved):

patmmccann:
@nitin0610 could you clarify what is still in progress? This is looking very near finished to me

nitin0610:
> @nitin0610 could you clarify what is still in progress? This is looking very near finished to me

Hi @patmmccann, thanks for approving the changes/request.

Currently we are working on test these feature end to end throughly due to recent changes of GDPR, device access changes. No dev work is pending for this feature. 

We will be planning to complete this ASAP and will remove the In progress note from heading.

Also we are working on the documentation for the same.

Thanks.

modules/topicsFpdModule.js
```diff
@@ -73,6 +115,145 @@ export function processFpd(config, {global}, {data = topicsData} = {}) {
...
+  let cachedTopicData = [];
+  if (!HAS_GDPR_CONSENT || !HAS_DEVICE_ACCESS) {
+    return cachedTopicData;
+  }
+  const topics = config.getConfig('userSync.topics') || bidderIframeList;
+  const bidderList = topics.bidders || [];
+  let storedSegments = new Map(safeJSONParse(coreStorage.getDataFromLocalStorage(topicStorageName)));
+  storedSegments && storedSegments.forEach((value, cachedBidder) => {
+    // Check bidder exist in config for cached bidder data and then only retrieve the cached data
+    let isBidderConfigured = bidderList.some(({bidder}) => cachedBidder == bidder)
```
gwhigs, line 131:
```suggestion
    let isBidderConfigured = bidderList.find(({bidder}) => cachedBidder == bidder)
```
to access `isBidderConfigured?.expiry` below.

Should probably rename `isBidderConfigured` to reflect the type change to object.

gwhigs (changes requested):

modules/topicsFpdModule.js
nitin0610: Variable name isBidderConfigured is replaced **bidderConfigObj**.

gwhigs (approved):

nitin0610:
@patmmccann We have completed the testing. PR is ready to get merged. We have approvals too.

patmmccann:
Thanks! @nitin0610

I saw you mention the documentation but now I'm having trouble finding it

nitin0610:
@patmmccann  Please find below PR related to documentation
https://github.com/prebid/prebid.github.io/pull/4222

PR #8981, opened 2022-09-09
Conversant Analytics Adapter: initial release

AaronColbyPrice:
## Type of change

- [ ] Bugfix
- [x ] Feature
- [ ] New bidder adapter   
- [ ] Build related changes
- [ ] CI related changes

- [ ] Other

## Description of change

Initial release of Conversant Prebid Analytics Adapter
Documentation: https://github.com/prebid/prebid.github.io/pull/4011


AaronColbyPrice:
@robertrmartinez I'm curious if you have an ETA on the review for this.  My team is trying to figure out when we can announce some marketing material and I'm not sure what I can tell them.

robertrmartinez:
@AaronColbyPrice 

Will give it a thorough review tonight!

AaronColbyPrice:
@robertrmartinez great thanks so much.  Didn't mean to rush you, I know you are busy.  Just trying to figure out what to tell my product manager :)

integrationExamples/gpt/pbjs_video_adUnit.html
```diff
@@ -51,7 +51,15 @@
...
-
+  pbjs.que.push(function() {
+    pbjs.enableAnalytics({
+      provider: 'conversant',
+      options: {
+        site_id: '119041',
+        sampleRate: 100
+      }
+    })
+  });
```
robertrmartinez, line 62: think we can remove this from the example page.

modules/conversantAnalyticsAdapter.js
```diff
@@ -0,0 +1,564 @@
+import {ajax} from '../src/ajax.js';
+import adapter from '../libraries/analyticsAdapter/AnalyticsAdapter.js';
+import CONSTANTS from '../src/constants.json';
+import {getGlobal} from '../src/prebidGlobal.js';
+import adapterManager from '../src/adapterManager.js';
+import * as utils from '../src/utils.js';
```
robertrmartinez, line 6: Better to just import the needed utils instead of all here.

robertrmartinez: Looks to just be a couple log utils, deepAccess, and isInteger. So very small amount of the util lib! So just import those specifically.

modules/conversantAnalyticsAdapter.js
```diff
@@ -0,0 +1,564 @@
...
+  if (!Array.isArray(args.adUnits)) {
+    utils.logError(CNVR_CONSTANTS.LOG_PREFIX + 'onAuctionEnd(): adUnits not defined in arguments.');
+    return;
+  }
+
+  args.adUnits.forEach(adUnit => {
+    const cnvrAdUnit = cnvrHelper.createAdUnit();
+    // Initialize bids with bidderCode
+    adUnit.bids.forEach(bid => {
+      cnvrAdUnit.bids[bid.bidder] = cnvrHelper.initializeBidDefaults();
```
robertrmartinez, line 254: It is possible that a publisher has more than one bid for a given bidder within an adUnit.

```js
{
   code: 'some-ad-unit-code',
   mediaTypes: {
      banner: {
         sizes: [300,250, 728,90]
      }
   },
   bids: [
      { bidder: 'ix', params: { siteId: 1234, size: [300,250] } },
      { bidder: 'ix', params: { siteId: 1234, size: [728,90] } }
   ]
}

```

I think in this case, this code will always just set it to whatever the last ix bid is in here.

Same for lower in the `nobids` and `bidsrecieved` parts, it will add an event to eventCodes array, but overwrite the `bidPayload` with the last entry in it.

I guess if you do not care and do not want to track such scenario, or just want the last bid in the array to send it's data it is fine.

But maybe something to look into. I know with my experience I see many pubs having several bids in the bids array for the same bidders.

modules/conversantAnalyticsAdapter.js
```diff
@@ -0,0 +1,564 @@
...
+        bidPayload.eventCodes.push(CNVR_CONSTANTS.BID);
+        bidPayload.timeToRespond = bid.timeToRespond;
+        bidPayload.originalCpm = bid.originalCpm;
+        bidPayload.cpm = bid.cpm;
+        bidPayload.currency = bid.currency;
+        bidPayload.mediaType = bid.mediaType;
+        bidPayload.adSize = {
+          'w': bid.width,
+          'h': bid.height
+        };
```
robertrmartinez, line 323: Like here for example,

if for an adUnit you had 3 bids from bidder IX,

The last bid in the array would set it's alues to the bidPayloads object.

robertrmartinez: I think at least, I may be missing something!

robertrmartinez (changes requested):
Looks fine, just a couple minor things + a question.

modules/conversantAnalyticsAdapter.js
AaronColbyPrice: @robertrmartinez interesting edge case that we didn't realize.  I'm discussing with my team how to handle this and will follow up in the next day or so.  Thanks for your detailed look at our adapter!

AaronColbyPrice:
@robertrmartinez fixed all issues noted

robertrmartinez (approved):
Nice! Looks good

PR #8984, opened 2022-09-13
dgkeywordRtdProvider submodule: support 1plusX SDK

yki-t:
## Type of change

- [ ] Bugfix
- [x] Feature
- [ ] New bidder adapter   
- [ ] Build related changes
- [ ] CI related changes

- [ ] Other

## Description of change

added support for `fpid` param in LocalStorage: suit for 1plusX spec

### DIFF
--- a/modules/dgkeywordRtdProvider.js
+++ b/modules/dgkeywordRtdProvider.js
@@ -41,7 +41,7 @@ export function getDgKeywordsAndSet(reqBidsConfigObj, callback, moduleConfig, us
   } else {
     logMessage('[dgkeyword sub module] dgkeyword targets:', setKeywordTargetBidders);
     logMessage('[dgkeyword sub module] get targets from profile api start.');
-    ajax(getProfileApiUrl(moduleConfig?.params?.url), {
+    ajax(getProfileApiUrl(moduleConfig?.params?.url, moduleConfig?.params?.disableReadFpid), {
       success: function(response) {
         const res = JSON.parse(response);
         if (!isFinish) {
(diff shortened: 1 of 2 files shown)


patmmccann:
What is happening here? One rtd module is reading from another's endpoint? Can you explain the use case, perhaps in your docs pr?

patmmccann:
@bwhisp can you comment?

yki-t:
> What is happening here? One rtd module is reading from another's endpoint?
Yes. This module was built for connecting to 1plusx, and this patch doesn't change the endpoint.

> Can you explain the use case, perhaps in your docs pr?
If 1plusx added `fpid` to LocalStorage, this patch sends it to 1plusx

yki-t:
Made fpid optional by 4e30dab

yki-t:
@patmmccann 
Any problems ?

patmmccann:
we're waiting on comment from triplelift team. @bwhisp please let us know

bwhisp:
Hello @patmmccann, sorry for the delay
So for me the code looks good 
However I noticed the params for dgkeywordRtdProvider changed with the addition of `url` & `disableReadFpid`
So it might be a good idea to update the docs (.md file) & the integration example (if it exists) to include these parameters

patmmccann:
@yki-t we're happy to merge with the docs update, thanks!

gggiakas:
I would like to give some context to the reviewers about this PR.

Digital Garage which is a customer of 1plusX DMP, part of TripleLift,  uses its in-house developed Prebid module (https://docs.prebid.org/dev-docs/modules/dgkeywordRtdProvider.html#digital-garage-keyword-module)  which was developed before we implement ours. 

This modules uses the 1plusX endpoint which provides the targeting information. Then, the targeting information is passed to the bid request.
In this endpoint a user identifier has to be passed in order to identify the user and provide the respective targeting information. 
Until August Digital Garage was using an 1plusX snippet that used only a 3rd party cookie.
Then, they upgraded to a newer snippet which sets an ID in the local storage that is called FPID -first party Id.
As a consequence, they wanted to use this additional ID in their prebid module so that they can leverage the presence of this ID (mainly for Safari) for user addressability.

So the aim of this PR is to retrieve from the local storage this ID which is set by the DMP snippet (fpid) and pass it as a query parameter to the endpoint that they call to get the targeting info.

I hope you find this context helpful.

patmmccann (approved):

PR #8995, opened 2022-09-14
USP consent management module: fix bug with zero-conf not being enabled on first auction

dgirardi:
## Type of change

- [x] Bugfix

## Description of change

Since v7, the USP module automatically enables itself without requiring any config from the user. However, after (I believe) #8626 the timing of when auction hooks run changed so that the USP module incorrectly enables itself _after_ the first auction has already started. This bugfix makes sure USP enables itself in time.


patmmccann (approved):

jsnellbaker (approved):
LGTM

PR #8997, opened 2022-09-14
Rubicon Bid Adapter: Choose video when banner mediaType is not present

spotxslagle:
## Type of change

- [ X ] Bugfix

## Description of change
Rubicon's Adapter only classifies a bid as video when mediaTypes.video and params.video are both defined. This change causes the adapter to only check against params.video when mediaTypes.banner is also present, otherwise mediaTypes.video is enough to classify as video

## Other information
@robertrmartinez


robertrmartinez (approved):
Looks good, will you add a unit test covering classifiedAsVideo.

robertrmartinez (changes requested):
Actually the unit tests are failing, so please fix and add a couple for the new `classifiedAsVideo`

test/spec/modules/rubiconBidAdapter_spec.js
```diff
@@ -1990,6 +1990,19 @@ describe('the rubicon adapter', function () {
           expect(spec.isBidRequestValid(bidderRequest.bids[0])).to.equal(false);
         });
 
+        it('should be invalid if bidRequest.mediaTypes.video.context is instream but size_id is not defined', function () {
```
robertrmartinez, line 1993: where is the check for this exactly? I do not see `size_id` being checked if instream anywhere.

robertrmartinez (approved):
Is there a test for when BOTH banner and VIDEO are present

But also params.video object is defined?

Thus, making it classify as video?

test/spec/modules/rubiconBidAdapter_spec.js
spotxslagle: This test used to be under the hasVideoType tests, but it doesn't actually have anything to do with that function. I just moved it to where a bunch of other spec.isBidRequestValid situations were being tested. No idea if it tests what it is supposed to.

robertrmartinez (approved):
LGTM! @ChrisHuie Will you do a non Magnite review on this please?!

ChrisHuie (approved):

Issue #8955, opened 2022-09-06
Allow for configurable referrer detection frequency


patmmccann:
[(https://github.com/prebid/Prebid.js/pull/8864)](https://github.com/prebid/Prebid.js/pull/8864) allowed for an optimization where referrer detection only ran once, yet this broke some people in SPA / Infinite Scroll environments. ( #8940 )

This ticket is to track re-implementation in a non-breaking way

patmmccann:
solved in #9670

PR #9004, opened 2022-09-16
Prebid Core Referer Detection: Allow for configurable referrer detection frequency

jlquaccia:
## Type of change
- [x] Feature

## Description of change
This PR introduces new logic to support configurable referrer detection frequency.  New logic is as follows:
- By default referrer detection will occur only once
- Optionally, if the following is passed into the setConfig func, then the referrer frequency detection will be continuous:
`pbjs.setConfig({refDetectionFreq: 'ongoing'});`

## Other information
Relates to https://github.com/prebid/Prebid.js/issues/8955
@dgirardi can you review?

### DIFF
--- a/src/refererDetection.js
+++ b/src/refererDetection.js
@@ -11,6 +11,15 @@
 import { config } from './config.js';
 import {logWarn} from './utils.js';
 
+let RI = new WeakMap();
+let ongoing = false;
+config.getConfig('refDetectionFreq', (config) => {
+  if (config.refDetectionFreq === 'ongoing') {
+    ongoing = true;
+    resetRefererInfo();
+  }
+});
+
 /**
  * Prepend a URL with the page's protocol (http/https), if necessary.
  */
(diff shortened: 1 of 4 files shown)


dgirardi:
Hey @jlquaccia,

I'd prefer if we tried to optimize this without any configuration, I'm thinking of something like:

 - on the first call, run the logic, cache the result and return it
 - on subsequent calls:
    - if we are on the top frame:
        - check if `window.location.href` or `getCanonicalUrl` have changed since the last call. If they did, run the logic again and cache the result
    - otherwise, run the logic again and cache the result
    - return the cached result

I expect Prebid to be living in the top frame quite often, and because in that case it's pretty easy to check whether the cache should be invalidated, I think this approach would get us most of the way to minimizing how often we run this logic.

jsnellbaker:
Outside of the code conflicts, I agree with Patrick that the default behavior shouldn't be changed in a minor release.  We should maintain behavior as it is today and if the prebid group decides to change the default behavior later, then that can be decided and agreed upon (if we're sure it won't break most setups).  So then the config (if there will be one - see below), should be an opt-in to do the look-up once.

If not - then maybe this PR should just wait for the next major release in general.

To Demetrio's point, avoiding the config requirement can be good goal.  I'm not sure if we can do that and also maintain existing behavior; maybe that can be part of the changes we do in a major release when we also change the default logic?

dgirardi:
@jlquaccia - is this work still in progress, or do you mind if I pick it up?

jlquaccia:
@dgirardi - sorry for not following up on this one.  sure, you can go ahead and pick this one up, have a bit on my plate over here at work at the moment anyway.

dgirardi:
Closing this in favor of https://github.com/prebid/Prebid.js/pull/9670

PR #9005, opened 2022-09-16
AndBeyond.Media adapter: update bidder code

AndBeyondMediaHB:
doc - https://github.com/prebid/prebid.github.io/pull/4018


modules/andBeyondMediaBidAdapter.js
```diff
@@ -4,7 +4,7 @@ import { registerBidder } from '../src/adapters/bidderFactory.js';
 import { BANNER, NATIVE, VIDEO } from '../src/mediaTypes.js';
 import { config } from '../src/config.js';
 
-const BIDDER_CODE = 'beyondmedia';
+const BIDDER_CODE = 'andbeyond.media';
```
patmmccann, line 7: This is very long, why not andbeyond

patmmccann (changes requested):

modules/andBeyondMediaBidAdapter.js
patmmccann: It's also breaking i believe

AndBeyondMediaHB: > This is very long, why not andbeyond

The thing is that this is a full name of the Company and we cannot shorten this to andbeyond. I haven't seen any limitations before. Do you have any?

muuki88: GAM key values have a limit of 20 chars. And GET requests have a length limit. This can cause issues for publishers which use `enableSendAll` key values.

AndBeyondMediaHB: @patmmccann @muuki88 
We have changed the code to "andbeyond". This is fine ?

patmmccann: You still need the old name as an alias so as to not break existing builds

AndBeyondMediaHB: @patmmccann @muuki88 we noticed that biddercode `andbeyond` is an Adkernel adapter alias, is that normal?https://github.com/AndBeyondMediaHB/Prebid.js/blob/master/modules/adkernelBidAdapter.js#L82

AndBeyondMediaHB: @patmmccann @muuki88 
image

Are these rules valid? If so, can we use the biddercode `andbeyondmedia`. If not, why not?

patmmccann (approved):

modules/andBeyondMediaBidAdapter.js
patmmccann: well in this case you would not be unique with adkernel's alias for the first 6 characters. Did you discover why they are using your name?

AndBeyondMediaHB: @patmmccann They are using this name because this is a name of their company, they cannot change it to another one. Could you please clarify if it can cause some problems when this is not unique with adkernel's alias?

patmmccann: AdKernel's company name is also andbeyond?

AndBeyondMediaHB:
@patmmccann @muuki88 we left the bidder code as is. We just changed the [title](https://github.com/prebid/prebid.github.io/pull/4018/files) and renamed the js-files to fix the upload [issue](https://github.com/prebid/Prebid.js/issues/9037)

bretg (changes requested):
ok - changing the biddercode to "beyondmedia" for Prebid.js will solve the website download problem, but now you've introduced another problem -- you're not matching the biddercode for your server-side adapter.

I'm afraid I have to insist that your client-side and server-side bidder code match. Publishers need to be able to use the [Server-to-Server testing module](https://docs.prebid.org/dev-docs/modules/s2sTesting.html), which they wouldn't be able to do for your service if you've got different codes for these adapters. Let's get this right.

Will let you choose whether to use "beyondmedia" or "andbeyondmedia" - just settle on one please. And if desired, you can define an alias on either or both client- and server-sides to use both names!

bretg (approved):

bretg:
Question: should the adkernel adapter still be aliasing `andbeyond`?

AndBeyondMediaHB:
> Question: should the adkernel adapter still be aliasing `andbeyond`?

@bretg we are not affiliated with adkernel adapter, also we don't know if it should use `andbeyond` alias

patmmccann (approved):

PR #9012, opened 2022-09-16
Appnexus Bid Adapter: fix undefined state with adunit ortb keywords

### LINKED ISSUE #9006: Appnexus Adapter: bug during conversion of ortb2 keyword strings to object format
## Type of issue
Bug

## Description
When loading ads, the browser console shows an error in the Appnexus adapter with the information that `string.split()` cannot be used on `undefined`. 

I think this is caused by `ortb2ImpKwStr` being undefined (see screenshot below). 
https://github.com/prebid/Prebid.js/blob/c4605d582e2c9055ff7972fdad44a891d410c366/modules/appnexusBidAdapter.js#L785-L788

The function `convertStringToKeywordsObj` still gets called and `split()` is trying to be used on `keyStr` which is `undefined`.
https://github.com/prebid/Prebid.js/blob/c4605d582e2c9055ff7972fdad44a891d410c366/modules/appnexusBidAdapter.js#L1202-L1206

## Screenshot
Screenshot 2022-09-15 at 18 44 39

## Platform details
Prebid.js version 7.15

## Other information
Probably connected to this [PR](https://github.com/prebid/Prebid.js/pull/8939).

jsnellbaker:
## Type of change

- [x] Bugfix

## Description of change
Fixes #9006 

Fixes a case where the undefined state of ortb keywords in the adunit wasn't properly caught and handled.

### DIFF
--- a/modules/appnexusBidAdapter.js
+++ b/modules/appnexusBidAdapter.js
@@ -250,10 +250,10 @@ export const spec = {
     // need to convert the string values into array of strings, to properly merge values with other existing keys later
     Object.keys(anAuctionKeywords).forEach(k => { if (isStr(anAuctionKeywords[k]) || isNumber(anAuctionKeywords[k])) anAuctionKeywords[k] = [anAuctionKeywords[k]] });
     // combine all sources of keywords (converted from string comma list to object format) into one object (that combines the values for shared keys)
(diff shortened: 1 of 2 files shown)


Fawke (approved):
LGTM

Issue #8956, opened 2022-09-06
Rejected bids can sometimes still win auctions / improving distinction between rejected and accepted bids


dgirardi:
## Type of issue
Bug / Improvement

## Description

There are 3 instances where Prebid can decide to reject a bid, and all of them are done in different ways:

1. [a bid is deemed not valid by core](https://github.com/prebid/Prebid.js/blob/75818b998a683c565d31f00d47f3b96f8ad689f2/src/adapters/bidderFactory.js#L200) (does not contain required fields, some fields have invalid values, etc): in this case, the invalid bid "disappears" as soon as it's returned by the adapter - there is no `BID_RESPONSE` event for, it's not returned by `pbjs.getBidReponses()`, cannot be rendered with `pbjs.renderAd`, and so on.
2. [a bid does not meet the floor](https://github.com/prebid/Prebid.js/blob/75818b998a683c565d31f00d47f3b96f8ad689f2/modules/priceFloors.js#L740) -  in this case, it's replaced with a similar bid, except that the new bid has `{cpm: 0, status: BID_REJECTED}` and bid.getStatusCode() returns `NO_BID`.
3. [some error happens during currency conversion](https://github.com/prebid/Prebid.js/blob/75818b998a683c565d31f00d47f3b96f8ad689f2/modules/currency.js#L241) - in this case, the bid is replaced with an essentially empty bid that has `bid.getStatusCode() === NO_BID` (unlike the floors case, we do not preserve properties from the original bid such as `mediaType`, `width`, `height`, etc)

The approach taken for cases 2 and 3 makes some version of the rejected bids available to analytics (or other event listeners), but has the disadvantage of requiring every consumer to always check `getStatusCode()` for every bid. Most of the codebase does not do so; notably:

 - [our targeting logic](https://github.com/prebid/Prebid.js/blob/75818b998a683c565d31f00d47f3b96f8ad689f2/src/targeting.js#L469) does not look at the bid status, if a publisher allows zero cpm bids through `bidderSettings.standard.allowZeroCpmBids`, and _all_ bids for an  auction are rejected, we are happy to set up targeting for rejected bids.
[...]

Issue #8887, opened 2022-08-23
Targeting logic silently ignores bids that have sizes not supported by `sizeConfig`


Jordanjscott:
## Type of issue
I've just set up Minute Media as another bidder, however nothing runs as GAM is not being sent their KV.

## Description
I have MM set up the same as other partners, and have had the ID's checked by the AM, all good there. All other partners are showing KV's in GAM and serving. Bids are being received, but no KV being sent in the request to GAM. 

### Expected results

For MM key values to be sent to GAM. 

## Platform details
v. 7.0.0

patmmccann:
Please provide a test page

Jordanjscott:
> Please provide a test page
> 
> just confirming the issue:
> 
> 'I am a publisher and I receive bids from MM bid adapter but I never see gam targeting set for them'

That's correct yeah, I have test mode on the mobile mpu (mpu1-m), so bids should always be received on this unit: 

https://www.thelondoneconomic.com/wrapper-test-page/

dgirardi:
I did some debugging and I think what's happening is that bids from minutemedia are being silently ignored because they have a size that is not supported by `sizeConfig`:

![image](https://user-images.githubusercontent.com/10151381/187293638-dd3e5e6d-900a-45e1-8919-9ee1d61a945e.png)

What made this tricky to track down is that we only check against `sizeConfig` when we set up targeting; the bid is a normal (and possibly winning) bid otherwise:

https://github.com/prebid/Prebid.js/blob/ece711e73114e73a452541edcaa073281f8cd801/src/targeting.js#L452

I'm not sure if this makes sense or if it's a bug. (@bretg or @patmmccann  does it make sense that a bid may win if you call renderAd, but not if you call setTargetingForGpt?)  

For the time being, would you be able to update your `sizeConfig` to accept the sizes returned by minutemedia?

dgirardi:
Actually, this may be a problem only when using minutemedia's "test mode", it appears to reply with predefined sizes and ignore what was requested? FYI @noamtzu

noamtzu:
@dgirardi indeed. the sizes are predefined. the test mode it's only for 100% fill for integration purposes.

patmmccann:
>  (@bretg or @patmmccann does it make sense that a bid may win if you call renderAd, but not if you call setTargetingForGpt?)

We should add sizeSupported filtering to the no adserver use case

Jordanjscott:
[...]

PR #9013, opened 2022-09-16
Core & multiple modules: introduce new `bidRejected` event, and refactor bid rejection logic

dgirardi:
## Type of change

- [x] Bugfix
- [x] Feature

## Description of change

This: 
 - introduces a new `bidRejected` event, emitted when an adapter returns a bid response that core decides should not be used
 - normalizes the "NO_BID" rejections done by priceFloors and currency. Unlike the proposal in https://github.com/prebid/Prebid.js/issues/8956, I decided to _not_ generate these bids outside of those cases, because they don't always work well through the usual bid response flow (for example, sometimes we do not know which bidder it came from, or which adUnit it refers to). 
 - the above means that apart from the new 'bidRejected' event, the current behavior is left unchanged, except that:
 - banner bid responses whose size does not match any of the sizes in the request are now rejected (see https://github.com/prebid/Prebid.js/issues/8887)
 - targeting logic now does not check for bid size, but does exclude NO_BID bids

## Other information

Closes https://github.com/prebid/Prebid.js/issues/8956
Closes https://github.com/prebid/Prebid.js/issues/8887

### DIFF
--- a/modules/categoryTranslation.js
+++ b/modules/categoryTranslation.js
@@ -34,13 +34,13 @@ export const registerAdserver = hook('async', function(adServer) {
 }, 'registerAdserver');
 registerAdserver();
 
-export const getAdserverCategoryHook = timedBidResponseHook('categoryTranslation', function getAdserverCategoryHook(fn, adUnitCode, bid) {
+export const getAdserverCategoryHook = timedBidResponseHook('categoryTranslation', function getAdserverCategoryHook(fn, adUnitCode, bid, reject) {
   if (!bid) {
-    return fn.call(this, adUnitCode); // if no bid, call original and let it display warnings
+    return fn.call(this, adUnitCode, bid, reject); // if no bid, call original and let it display warnings
   }
 
   if (!config.getConfig('adpod.brandCategoryExclusion')) {
-    return fn.call(this, adUnitCode, bid);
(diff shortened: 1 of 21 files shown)


ChrisHuie:
@Fawke @ncolletti I know you guys have a few other pr reviews. Do either of you need me to reassign this one?

Fawke (approved):
Looks good at a first glance. I was able to simulate an invalid bid and could see the `bidRejected` event fire.

I agree with the removal of the banner size check from targeting, and moving it as a separate check in Bidder Factory.

Rest, all looks good. I think making a note of this new event in the somewhere in the docs will be helpful for the Analytics adapters and other third party listeners.

lgadzhev-devrix:
I'm getting this error after this update

`Exception in queued GPT command TypeError: e.getStatusCode is not a function`

dgirardi:
@lgadzhev-devrix is it similar to https://github.com/prebid/Prebid.js/issues/9245? the author there concluded that the culprit was a browser extension. If your case is different could you open an issue and provide steps to reproduce?

lgadzhev-devrix:
Yep, it was caused by the AdWizzard on my end. Thank you! I'm not sure why
I couldn't find this issue while googling the error

*--*


*Lachezar Gadzhev*Backend Developer
*DevriX Ltd.*

Issue #8953, opened 2022-09-06
Configurable ttl buffer


patmmccann:
## Type of issue
Feature

## Description
https://github.com/prebid/Prebid.js/blame/6e36e363aa843603a8b7cb29da3242d84e5f4b9d/src/targeting.js#L29 makes sure a bid is not within a certain number of ms of its expiration before considering it expired. This buffer is currently harcoded to 1000 ms. For example, if a bid has a ttl of 30 seconds, a bid is currently considered expired at 29 seconds. There is publisher demand to vary this setting at the ad unit or auction level. The use case is that a publisher might expect an ad called in GPT 'lazy' mode to render 5 to 10 seconds in the future, and for ads not called in this manner to render near immediately. Some publishers might want to fly close to the sun and allow a negative buffer for revenue reasons, or to take advantage of buffers elsewhere in the 'machine'. 

## Other information
(https://github.com/prebid/Prebid.js/pull/6300/files)

PR #9028, opened 2022-09-22
targeting.js: allow configuration of TTL_BUFFER

caseywhitmire:
## Type of change

- [ ] Bugfix
- [x] Feature
- [ ] New bidder adapter   
- [ ] Build related changes
- [ ] CI related changes

- [ ] Other

## Description of change

Update targeting.js to allow the configuration of TTL_BUFFER. Fix for issue #8953.

### DIFF
--- a/src/targeting.js
+++ b/src/targeting.js
@@ -24,7 +24,7 @@ export const TARGETING_KEYS = Object.keys(CONSTANTS.TARGETING_KEYS).map(
   key => CONSTANTS.TARGETING_KEYS[key]
 );
 
-let TTL_BUFFER = 1000;
+var TTL_BUFFER = 1000;
 
 // return unexpired bids
 const isBidNotExpired = (bid) => (bid.responseTimestamp + bid.ttl * 1000 - TTL_BUFFER) > timestamp();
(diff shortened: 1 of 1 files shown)


src/targeting.js
```diff
@@ -65,6 +66,13 @@ export const getHighestCpmBidsFromBidPool = hook('sync', function(bidsReceived,
   return bidsReceived;
 });
 
+/**
+ * Allow the configuration of TTL_BUFFER
+ */
+export function setTTLBuffer(buffer) {
```
dgirardi, line 72: The [issue](https://github.com/prebid/Prebid.js/issues/8953) is about allowing the publisher to configure this, but this function is not accessible to the publisher. (The public publisher API consists only of what's exposed under the `pbjs` global).

The pattern we use elsewhere for this type of config is accepting them as parameters in `setConfig`. [Here's an example](https://github.com/prebid/Prebid.js/blob/0a14e91db750e3549bde19c371d1d34e877b6299/src/adapterManager.js#L50-L54) of registering a listener to watch for a specific config parameter. 

You'll need to choose a setting name (probably just `ttlBuffer`) which should be documented; and it'd be good to also have some test cases - `isBidNotExpired` should behave differently after different `setConfig`s.

caseywhitmire: I missed that, thanks. I'll get on it.

PR #9032, opened 2022-09-22
IX Bid Adapter: POST Support

lksharma:
## Type of change

- [ ] Bugfix
- [x] Feature
- [ ] New bidder adapter   
- [ ] Build related changes
- [ ] CI related changes

- [ ] Other

## Description of change
This release supports the switch from GET to POST requests via features

```

## Other information
@lksharma: love.sharma@indexexchange.com

### DIFF
--- a/modules/ixBidAdapter.js
+++ b/modules/ixBidAdapter.js
@@ -13,7 +13,8 @@ import {
   logError,
   logWarn,
   mergeDeep,
-  parseQueryStringParameters
+  parseQueryStringParameters,
+  safeJSONParse
 } from '../src/utils.js';
 import {BANNER, VIDEO, NATIVE} from '../src/mediaTypes.js';
 import {config} from '../src/config.js';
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

File: modules/ixBidAdapter.js [hunk 1]
```diff
@@ -99,8 +100,46 @@ const VIDEO_PARAMS_ALLOW_LIST = [
...
+    const expiryTime = new Date();
+    const toggles = deepAccess(responseBody, 'ext.features');
+
+    if (toggles) {
+      this.featureToggles = {
+        expiry: expiryTime.setHours(expiryTime.getHours() + 1),
+        features: toggles
+      }
+      if (storage.localStorageIsEnabled()) {
+        storage.setDataInLocalStorage(LOCAL_STORAGE_FEATURE_TOGGLES_KEY, JSON.stringify(this.featureToggles));
```
patmmccann, line 132: Why is it useful to read these values from local storage instead of relying on configuration

lksharma: enabling post is dependent on our server response for the moment. We are reading from an internal config and local storage when available. The advantage of reading from local storage is we can capture requests on page load rather than waiting for a response.

lksharma: @patmmccann let me know if you have any additional questions.

jsnellbaker (changes requested):
Hi @lksharma 

While I think the overall code looks fine (baring Patrick's feedback about the localStorage question), I was hoping to test these changes to verify the POST request completes successfully.

Do you have some sample params I can use that should invoke this state?  I was trying with some of the values in the md file and hardcoding some values in localStorage, but I'd rather test with the expected natural workflow, if possible.  Please let me know when you have the chance.

Thanks.

lksharma:
@jsnellbaker,

Thanks for looking into this. You can test the code flow with the following local storage value:

`
{"expiry":<TIMESTAMP>,"features":{"pbjs_enable_post":{"activated":true}}}
`

`
TIMESTAMP = new Date().setHours(new Date().getHours() + 1)
`

Let me know if you need anything else.

jsnellbaker (approved):
Thanks for the feedback @lksharma 

With that in place in the localStorage (and the bidderSettings storageAllowed setting enabled), the logic presented here appears to be working fine.  

Will wait on merging for Patrick's feedback.

modules/ixBidAdapter.js
patmmccann: setting ix_features: {"expiry":<TIMESTAMP>,"features":{"pbjs_enable_post":{"activated":true}}}

in local storage and giving a bidder an override to enable post method seems to be a bizarre workflow for a publisher and also seems undocumented. I feel like I must be missing something here. Why couldn't the publisher opt into IX post mode via a more standard method?

patmmccann:
Generally, this use of local storage is not a rules violation, but it feels bizarre and difficult to explain to publishers, who will have to grant you access to local storage one by one. I recommend retooling to avoid these conversations and to give publishers a simpler way to enable post mode.

modules/ixBidAdapter.js
lksharma: @patmmccann 

We won't be relying on local storage to activate POST. Local storage access will allow us to generate POST requests on page load but this is a dependency and more of a nice to have. 

If reading from local storage is not allowed then POST will be limited to ad refreshes on the page. In a future PR, we will be moving completely to POST. The intention here is to roll out POST within our adapter gradually.

patmmccann: This is an unsatisfying reason; we'll discuss in committee. I recommend you plan on this not being approved.

patmmccann:
spoke with Luke at IX, merging

PR #9048, opened 2022-09-28
Module: New Teads User ID Module

github-tom-kuhnen:
## Type of change

- [x] Module

## Description of change

New User ID Module for Teads

- GDPR/CCPA compliant
- Store User ID in Cookies

## Other information

Documentation: https://github.com/prebid/prebid.github.io/pull/4026


patmmccann:
Could you let us know why you and others are building things like this? what does this do that sharedid does not?

github-tom-kuhnen:
> Could you let us know why you and others are building things like this? what does this do that sharedid does not?

Because it is an ID that we generate and use/reuse it, that we master, we have a logic on it ...

Rothalack (approved):
Looks good, working in testing for me.

Issue #9049, opened 2022-09-28
RubiconBidAdapter singleRequest bidder config is not working


elad-yosifon:
## Type of issue
A bug

## Description
RubiconBidAdapter singleRequest config is not working anymore due to caching the "initial" config in a module scope, instead of dynamically using it.

## Steps to reproduce
adding this configuration with `pbjs.que.push` will not be memoized by the `rubiconBidAdapter` current code:
```javascript
pbjs.que.push(() => {
  pbjs.setConfig({
    rubicon: {
      singleRequest: true
    }
  });
});
```

### Expected results
setting `singleRequest: true` should trigger only one request per auction

### Actual results
setting `singleRequest: true` is ignored

## Platform details
- since Prebid.js version 4.10.0

## Other information
This is a regression, it used to work before this commit 92e234c4.
https://github.com/prebid/Prebid.js/pull/5780

dgirardi:
Closing since https://github.com/prebid/Prebid.js/pull/9050 was merged.

PR #9050, opened 2022-09-28
Rubicon. bugfix for singleRequest config regression

elad-yosifon:
## Type of change

- [x] Bugfix
- [ ] Feature
- [ ] New bidder adapter
- [ ] Build related changes
- [ ] CI related changes

- [ ] Other

## Description of change
please see #9049

### DIFF
--- a/modules/rubiconBidAdapter.js
+++ b/modules/rubiconBidAdapter.js
@@ -1279,8 +1279,7 @@ export function hasValidSupplyChainParams(schain) {
 }
 
 /**
- * Creates a URL key value param, encoding the
- * param unless the key is schain
+ * Creates a URL key value param, encoding the param unless the key is schain
  * @param {String} key
  * @param {String} param
  * @returns {String}
--- a/modules/rubiconBidAdapter.js
+++ b/modules/rubiconBidAdapter.js
@@ -333,7 +333,7 @@ export const spec = {
       }
     });
 
-    if (rubiConf.singleRequest !== true) {
+    if (config.getConfig('rubicon.singleRequest') !== true) {
       // bids are not grouped if single request mode is not enabled
       requests = videoRequests.concat(bidRequests.filter(bidRequest => bidType(bidRequest) === 'banner').map(bidRequest => {
         const bidParams = spec.createSlotParams(bidRequest, bidderRequest);


ChrisHuie (approved):

robertrmartinez:
@elad-yosifon @ChrisHuie 

Not sure how this got in and merged without someone from Magnite / Rubicon reviewing or commenting at least.

But I fail to see how this was not working?

We do keep an ongoing internal cache of any update made to the `rubicon` config.

Unless you are using prebid in a different way, where you load prebid core first, then run all your que block setConfigs, and then initialize other modules.

Because the que block should be executed AFTER modules are loaded. Including our bid adapter.

And since that is the case, our module is initialized and should register to the config BEFORE anything in the queue block does.

You can test this by putting breakpoints in your first que block function and inside of a module. 

The code before the change you implemented:
```js
let rubiConf = {};
// we are saving these as global to this module so that if a pub accidentally overwrites the entire
// rubicon object, then we do not lose other data
config.getConfig('rubicon', config => {
  mergeDeep(rubiConf, config.rubicon);
});


   // ..... and in bidRequest time
   if (rubiConf.singleRequest !== true) {
```

This saves any setConfig rubicon call into rubiConf. and this if statement will evaluate to FALSE and hit the else statement if you do your example:

```js
pbjs.que.push(() => {
  pbjs.setConfig({
    rubicon: {
      singleRequest: true
    }
  });
});
```

It 100% does run singleRequest auctions and keeps the state of what the above setConfig implements. Since que blocks are executed after modules have been initialized.

Here are some screenshots which prove this (I added info logs to show how this works)

image

And we see only one call to Rubicon Fastlane endpoint:

image

And here is the reverse where we turn singleRequest off

image


image


In fact this PR introduces what I consider a bug, and the whole reason we added this logic in the commit you reference as introducing the bug.

With the change implemented by this PR the following would turn OFF single request auctions

```js
pbjs.que.push(() => {
  pbjs.setConfig({
    rubicon: {
      singleRequest: true
    }
  });

  pbjs.setConfig({
    rubicon: {
      netRevenue: true
    }
  });
});
```

Which is obviously NOT intended.

These cases where multiple setConfigs for our `rubicon` object happen is exactly why we introduced the saving of the internal state using the `rubiConf` code I posted above.

This was before the adoption of `pbjs.mergeConfig` which I still see rarely used by publishers. Which is why we still maintain our internal cache of a self merged `rubiConf`.

Here is an example test page
Please, if you are going to propose a PR for an adapter, the maintainers should be tagged first and given some time to comment and review the PR.

If I am mistaken and you can provide an example test page where my change would not work please share.

But form my testing it works just fine.

Here is a jsfiddle which points to Prebid 7.18.0 which is the version before this PR was merged.

https://jsfiddle.net/0khqzt15/12/

I am planning to revert this PR back to the old implementation.

I am not aware that there is a use case where some `pbjs.setConfigs` can be executed BEFORE modules are loaded, however, I will add in logic to handle this in the PR I will submit.

This line of code should handle if this is a valid use case.
```js
let rubiConf = pbjs.getConfig('rubicon') || {};
```

CC @bretg @harpere

robertrmartinez:
Here is the PR https://github.com/prebid/Prebid.js/pull/10332

This PR reverts back to the old method I had set, as well as solve for any potential use case where a `rubicon` config is somehow set before rubicon module is initialized.

PR #9061, opened 2022-10-01
TheMediaGrid: Support multiRequest param

TheMediaGrid:
## Type of change

- [x] Feature

## Description of change

Support multiRequest param to send no-mapping requests for each bid


musikele:
Hi @TheMediaGrid , can you add a test case for the multibid path? from the coverage view i see it was not tested. Thanks

PR #9065, opened 2022-10-02
Kueezrtb Bid Adapter: New Bid Adapter

uditalias:
## Type of change
- [x] New bidder adapter   

[prebid.github.io](https://github.com/prebid/prebid.github.io) Pull Request:
https://github.com/prebid/prebid.github.io/pull/4052

## Description of change
This is KueezRTB new bid adapter
Tests are included

- maintainer: rtb@kueez.com
```
{
  bidder: 'kueezrtb',
  params: {
    cId: '562524b21b1c1f08117fc7f9',
    pId: '59ac17c192832d0011283fe3',
    bidFloor: 0.0001,
    ext: {
      param1: 'loremipsum',
      param2: 'dolorsitamet'
    }
  }
}
```


bretg (changes requested):
You adapter supports floors, so must support the standard Prebid.js floors mechanism. See https://docs.prebid.org/dev-docs/modules/floors.html#bid-adapter-interface

You're allowed to also support a bidder-specific floors parameter, but must support the convention so publishers don't have to configure floors individually for each bidder.

uditalias:
@bretg Thanks. updated the code to use the `getFloor` function.

bretg (approved):
My review does not count here... it was limited to noticing (from the documentation) that it wasn't handling floors properly.

uditalias:
Hey @bretg @lksharma
It's been almost a month since we opened this PR. It will be appritiated if we can get a signal or an update regarding this PR.

Thanks!

ChrisHuie:
> Hey @bretg @lksharma It's been almost a month since we opened this PR. It will be appritiated if we can get a signal or an update regarding this PR.
> 
> Thanks!

I will take a look at this new bid adapter this week to keep it moving. @uditalias can you please pull in recent commit to bring in our end 2 end testing we added a couple months back 🙏

uditalias:
Thank you @ChrisHuie 
PR updated

modules/kueezRtbBidAdapter.js
```diff
@@ -0,0 +1,276 @@
...
+  'id5id': 1,
+  'idl_env': 1,
+  'lipb': 1,
+  'netId': 1,
+  'parrableId': 1,
+  'pubcid': 1,
+  'tdid': 1,
+  'pubProvidedId': 1
+};
+const storage = getStorageManager({ gvlid: null, bidderCode: BIDDER_CODE });
```
ChrisHuie, line 24: Why is gvlid specifically null here instead of just not being included? I have seen this in a few id modules but usually bid adapters either add the gvlid or omit the parameter altogether.

ChrisHuie (approved):

modules/kueezRtbBidAdapter.js
uditalias: @ChrisHuie When we submitted this PR we didn't had a gvlid.
It's there just we wont forget to put it there once we have it 😀

PR #9066, opened 2022-10-03
PubWise Bid Adapter: clean up and updates to adUnit parsing

GLStephen:
Update for PubWise Bid Adapter.

### DIFF
--- a/modules/pubwiseBidAdapter.js
+++ b/modules/pubwiseBidAdapter.js
@@ -2,7 +2,6 @@ import { _each, isStr, deepClone, isArray, deepSetValue, inIframe, logMessage, l
 import { config } from '../src/config.js';
 import { registerBidder } from '../src/adapters/bidderFactory.js';
 import { BANNER, NATIVE } from '../src/mediaTypes.js';
-import { convertOrtbRequestToProprietaryNative } from '../src/native.js';
 const VERSION = '0.1.0';
 const GVLID = 842;
 const NET_REVENUE = true;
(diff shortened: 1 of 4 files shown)


modules/pubwiseBidAdapter.js
```diff
@@ -532,9 +532,8 @@ function _cleanSlotName(slotName) {
 
 function _initConf(refererInfo) {
   return {
-    // TODO: do the fallbacks make sense here?
-    pageURL: refererInfo?.page || window.location.href,
-    refURL: refererInfo?.ref || window.document.referrer
+    pageURL: (refererInfo && refererInfo.referer) ? refererInfo.referer : window.location.href,
+    refURL: window.document.referrer
```
patmmccann, line 536: Why aren't you using refererInfo. Ref here?

GLStephen: @patmmccann not sure, I'll check, but looks like we should be

GLStephen:
> one more question, could you solve for #8573 here? We're attempting to enforce all bidders pass:
> 
> auctionId as source.tid bidderRequest[i].transationId as imp[i].ext.tid

Sure thing

modules/pubwiseBidAdapter.js
patmmccann: refererInfo.ref and refererInfo.page have a bunch of logic attempting to determine them as of Prebid 7 merge

GLStephen: This has been adjusted in the next update.

GLStephen:
@patmmccann updates have been made

ChrisHuie (approved):

PR #9074, opened 2022-10-04
Multiple bid adapters: use auctionId for `source.tid`

dgirardi:
## Type of change

- [x] Bugfix

## Description of change

This changes bid adapters that were setting `source.tid = bidRequests[0].transactionId` to use `auctionId` instead. This is _not_ exhaustive - I just touched the adapters that were easy to grep for.

Part of https://github.com/prebid/Prebid.js/issues/8573

### DIFF
--- a/modules/improvedigitalBidAdapter.js
+++ b/modules/improvedigitalBidAdapter.js
@@ -307,10 +307,10 @@ const ID_REQUEST = {
     }
     // In the single request mode, split imps between those going to the ad server and those going to extend server
     if (extendImps.length) {
-      requests.push(formatRequest(extendImps, null, true));
+      requests.push(formatRequest(extendImps, bidderRequest.auctionId, true));
     }
     if (adServerImps.length) {
-      requests.push(formatRequest(adServerImps, null, false));
+      requests.push(formatRequest(adServerImps, bidderRequest.auctionId, false));
     }
 
     return requests;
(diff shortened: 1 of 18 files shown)


ChrisHuie (approved):

patmmccann (approved):

modules/improvedigitalBidAdapter.js
```diff
@@ -307,10 +307,10 @@ const ID_REQUEST = {
     }
     // In the single request mode, split imps between those going to the ad server and those going to extend server
     if (extendImps.length) {
-      requests.push(formatRequest(extendImps, null, true));
+      requests.push(formatRequest(extendImps, bidderRequest.auctionId, true));
```
patmmccann, line 310: https://github.com/prebid/Prebid.js/blob/f407e47f96f12834bd79092bc1b5bd0203967598/modules/improvedigitalBidAdapter.js#L165 seems to still be there @dgirardi

dgirardi: @patmmccann, I missed that, but I don't think it has any effect now, because it's always overridden by this PR setting it to auctionID afterwards. Since all of that is being refactored in https://github.com/prebid/Prebid.js/pull/8738 I think it's OK to leave.

patmmccann: no problem; I changed the original issue to call out a bunch that i found where source was on one line and tid was on the next line

PR #9083, opened 2022-10-05
Nativo Bid Adapter: change QS param referrer value

jsfledd:
## Type of change

## Description of change
Changed ntv_url QS param to use referrer.location instead of referrer.page


modules/nativoBidAdapter.js
```diff
@@ -145,7 +145,7 @@ export const spec = {
         'params.url',
       )
       if (pageUrl == undefined || pageUrl === '') {
-        pageUrl = bidderRequest.refererInfo.page
+        pageUrl = bidderRequest.refererInfo.location
```
patmmccann, line 148: what's wrong with page? We spent a lot of time on the logic of it

jsfledd: Hi @patmmccann, we need the full URL including any query params which is why we opted for **location** over **page** so we would retain the entire url value

patmmccann (approved):

PR #9089, opened 2022-10-07
Smaato Bid Adapter: Native support

el-chuck:
## Type of change

- [x] Feature

## Description of change

Adding native support for Smaato bid adapter

## Other information

prebid/prebid.github.io#4064

### DIFF
--- a/modules/smaatoBidAdapter.js
+++ b/modules/smaatoBidAdapter.js
@@ -1,12 +1,42 @@
-import { deepAccess, getDNT, deepSetValue, logInfo, logError, isEmpty, getAdUnitSizes, fill, chunk, getMaxValueFromArray, getMinValueFromArray } from '../src/utils.js';
+import { cleanObj, deepAccess, getDNT, deepSetValue, logInfo, logError, isEmpty, getAdUnitSizes, fill, chunk, getMaxValueFromArray, getMinValueFromArray } from '../src/utils.js';
 import {registerBidder} from '../src/adapters/bidderFactory.js';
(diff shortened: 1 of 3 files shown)


modules/smaatoBidAdapter.js
```diff
@@ -342,6 +426,55 @@ function createVideoImp(bidRequest, videoMediaType) {
   };
 }
 
+function createNativeImp(bidRequest, nativeMediaType) {
+  const mainImageSize = nativeMediaType.image.sizes
```
dgirardi, line 430: The documentation for it is still in progress, so the timing of this PR is a bit unfortunate, but prebid now supports native ORTB out of the box with very little adapter code required:

- `bidRequest.nativeOrtbRequest` should be an object very similar to your `createNativeImp(..).native`
- bids returned by `interpretResponse` can use the ORTB native response directly (`{native: {ortb: JSON.parse(bid.adm)}}`)

@musikele FYI

el-chuck: thank you, i'm having a look and will update the PR

dgirardi (approved):
The code LGTM, but beware that I am not able to properly test this with the example config in your md file. If I use this:

```
{
        "bidder": "smaato",
        "params": {
            "publisherId": "1100042525",
            "adspaceId": "130563103"
        }
    }
```

the endpoint replies with 204, no content (and no bid). In particular I wanted to verify the server  response because this code expects `{native: [ORTB native response]}`, which is different from what I'm used to - the ORTB native response at the top level.

Issue #8295, opened 2022-04-18
PPID sync call happening too soon


patmmccann:
## Type of issue
https://github.com/prebid/Prebid.js/blob/d9465ed88c457de92f45afc06cc1bd71210358bc/modules/userId/index.js#L664 doesn't take advantage of #8201 . We want to call the gpt set ppid function as soon as the id is available; ideally before the first gpt display call. It appears the current behavior is if this function returns null, it doesnt happen again. 

This issue is to discuss possible configurations / solutions. One is to simply migrate to the new global function from 8201.

dgirardi:
As far as I know, there's currently no easy way to tell when a particular ID has become available - https://github.com/prebid/Prebid.js/pull/8201 gives an all-or-nothing signal for when *all* IDs are ready.
 
But, if that call can wait until just before ad rendering (that is, there's no advantage in making it sooner), this hook seems like a better place for it:

https://github.com/prebid/Prebid.js/blob/d9465ed88c457de92f45afc06cc1bd71210358bc/src/auction.js#L364

patmmccann:
another possibility, it could occur as part of the setkeys process

https://github.com/prebid/Prebid.js/blob/2d14b7d53337c47b2e804f184927ed498990cc8f/src/auction.js#L751

patmmccann:
The proposals seem functionally equivalent and the choice will be made by the implementer

PR #9099, opened 2022-10-11
UserID: call gpt `setPublisherProvidedId` when PPID is available

dgirardi:
## Type of change

- [x] Bugfix

## Description of change

This moves the call to gpt's `setPublisherProvidedId` from from the beginning of each auction to the moment the selected ID becomes available. This way, if the PPID is resolved during an auction it should still be made available to GAM.

## Other information

Fixes https://github.com/prebid/Prebid.js/issues/8295


ncolletti (approved):
lgtm

Issue #9110, opened 2022-10-13
Feature request: make analytics non-blocking


patmmccann:
###Description

Analytics adapters listening to events may occasionally block the auction; we'd like to rearchitect that in a better way. 
Reported by @robertrmartinez ; feel free to expand.

dgirardi:
@robertrmartinez , if you don't mind, I plan to do this together with the changes needed for https://github.com/prebid/Prebid.js/issues/9014

robertrmartinez:
Of course not! Be my guest!

PR #9113, opened 2022-10-13
Multiple analytics modules: allow pub-defined event filters; do not block auction for analytics

dgirardi:
## Type of change

- [x] Feature

## Description of change

This replaces the hardcoded list of events that are forwarded to analytics adapters with a whitelist and/or blacklist provided by the publisher:

```
pbjs.enableAnalytics([{
   provider: 'rubicon',
   includeEvents: [
       'bidWon', 
       'auctionInit', 
       // ...
   ],
   excludeEvents: [
       'auctionDebug', 
       // ...
   ],
   options: {
     /...
   }
}])
```

Additionally, event handlers for analytics are deferred until some time has passed since the last event, to avoid blocking the auction (see https://github.com/prebid/Prebid.js/issues/9110). The debounce delay is fixed at 100ms; I am not sure if there is a use case to make it configurable.

## Other information

Resolves https://github.com/prebid/Prebid.js/issues/9014
Resolves https://github.com/prebid/Prebid.js/issues/9110
Documentation: https://github.com/prebid/prebid.github.io/pull/4082


ncolletti (approved):
these changes look good to me, will wait for a second review before merge

patmmccann (approved):

PR #9114, opened 2022-10-13
PulsePoint Bid Adapter: First party data

anand-venkatraman:
## Type of change
- [x] Feature - Adding first party data lookup to PulsePoint bid adapter.

## Description of change
First party data from `site`, `user` and `impression` level, are retrieved and passed on to pulsepoint endpoint.


musikele:
@anand-venkatraman keep an eye on PR https://github.com/prebid/Prebid.js/pull/8738 . It will give you an ORTB conversion library that will greatly simplify this conversion process to ORTB and back.

Issue #9081, opened 2022-10-05
USP Privacy Module: listen for deletion event


patmmccann:
## Type of issue
Feature request

## Description

https://github.com/InteractiveAdvertisingBureau/USPrivacy/blob/master/CCPA/Data%20Deletion%20Request%20Handling.md describes an event we should potentially listen for and react to from the USP module. I am told ( by @aramzs ) that Prebid will need to fire endpoints with deletion requests upon 'hearing' this event under CPRA. Looking for confirmatory documentation.

jdwieland8282:
Thanks for sharing, so gpp-core needs an pub defined array of vendor endpoints where each endpoint is called when  Data Deletion Request is received. I'm operating off this [spec](https://docs.google.com/document/d/1dRxFUFmhh2jGanzGZvfkK_6jtHPpHXWD7Qsi6KEugeE/edit) which considers a GPP core module with various sub adapters representing various regs. 

I see this issue references the existing USP Privacy Module, do we want to continue development on that module?  If so, we could do the same, add an array of vendor endpoints that get invoked when the deletion request comes in. 

Functionally, I doubt annnnny vendor has an endpoint for this though.

patmmccann:
Yeah; suppose you are a fingerprint vendor. The user opts out two min into the page view. The user id module fingerprint vendor never knows.

jdwieland8282:
Ok, so summarizing the Identity PMC consensus. 

1. The USP module will fire a deletion signal as an internal event to the User id sub adapters, when an asynchronous "delete" signal is given by the user/CMP. 
2. Sub adapters are informed on how to listen for this event and what to do with it when they receive it.

patmmccann:
This is fairly limited in scope which is quite positive, marking ready for dev. So Prebid will simply fire an internal event that anyone vendor module in Prebid can fire an xhr from when it hears a deletion request from the uspapi. Should we wipe the shared id from local storage or cookie on that event?

jdwieland8282:
yes "shared id from local storage or cookie on that event". 

To be clear though, we expect the sub adapters to be listening for that event and do the necessary removal, or are you saying in sharedids case the usp module will do that for it?

patmmccann:
the sharedid module should listen for the event from the usp module

dgirardi:
Here's my proposal:

- Extend the interface for _all_ types of adapters (bidders, userId, analytics) with a new `onDataDeletionRequest` method;*
[...]

PR #9116, opened 2022-10-14
Multiple modules: USP data deletion event handling

dgirardi:
## Type of change

- [x] Feature

## Description of change

This defines an interface for handling USP [opt out events](https://github.com/InteractiveAdvertisingBureau/USPrivacy/blob/master/CCPA/Data%20Deletion%20Request%20Handling.md):

 - UserID submodules can define a method `onDataDeletionRequest(config, idValue)`
 - Bid adapters can define a method `spec.onDataDeletionRequest(bidderRequests)`
 - Analytics adapters can define a method `onDataDeletionRequest()`
 - The USP module attaches a 'registerDeletion' event handler with the CMP; when triggered, it will:
    - invoke the methods above on all adapters
    - delete all IDs from cookies/localStorage

Note that:

 - this was _not_ tested against a real CMP; I do not have access to a working USP CMP
 - the publisher is still responsible for checking consent and disable storage, analytics, and so on; Prebid currently does not attempt to enforce USP, so on the next session, it has no memory of the opt-out request. 

## Other information
Resolves https://github.com/prebid/Prebid.js/issues/9081
Documentation TBD


ChrisHuie (approved):

jdwieland8282:
Can we add the method to the Sharedid sub adapter and add an integration test to see if it works?  THat way we get 1) sharedid into compliance, and 2) proof it works, aside from the CMP to adapter piece.

dgirardi:
@jdwieland8282, what should the method do? stored IDs are already removed for all adapters, and there are test cases for it. The method is to allow adapters to "propagate" the request to their endpoint / do anything else that needs to be done, I'm not sure what that would be for sharedId.

jdwieland8282:
ok, I misunderstood, I though each sub adapter would need to have that method and call it when the the delete me signal occurred. It sounds like you are saying is that the modified USP module will blow those ids away.

jdwieland8282 (approved):
LGTM

jdwieland8282:
This is a problem 

```Prebid currently does not attempt to enforce USP, so on the next session, it has no memory of the opt-out request```

When users ask to be forgotten then expect to only need to do that once, and that this preference will persist.  In addition to removing the users uids, we should also set a `usp_optout=TRUE` value, and check it before allowing the userid module to set new uids.

dgirardi:
@jdwieland8282, shouldn't the consent string change after opt-out?

From https://github.com/InteractiveAdvertisingBureau/USPrivacy/blob/master/CCPA/US%20Privacy%20String.md , my understanding is that the "Opt-out sale" flag should change to `N`. IMO the missing piece is a USP enforcement module, does it make sense to do it partially for userId only?

jdwieland8282:
I'll add this to the Identity PMC meeting today. Will follow up this afternoon.

SKOCHERI (approved):

patmmccann:
@dgirardi this looks ready to merge pending a docs pull

jdwieland8282:
Unless anyone objects by 10/28/22, the USP data deletion event handling will do the following.

- delete me will persist for the session, the data deleted will be from the current session, when/if a user returns the pub may begin tracking him/her again. If the user wishes to have a more persistent opt out, he/she should do so via the publishers CMP. 
-  The deletion method needs to be adopted by all modules that store information client side including user ids, segments (RTD modules), analytics events, and bid adapters. The method is defined above (`onDataDeletionRequest`).

dgirardi:
@jdwieland8282 (or @AramZS), what should be the guidance given to adapters for what do to in `onDataDeletionRequest`? Is there a reference we can use on what constitutes compliance?

 - delete all locally stored data (or just PII)? 
 - delete all stored data downstream? how deep do you need to go?

jdwieland8282:
delete all locally stored data (or just PII), and propagate the signal downstream

We don't have such a reference we only have the reg itself. This site is a good place to start. We don't want to give legal advice, bid adapters should form their own opinion about what to do when they receive a delete me signal. 
https://oag.ca.gov/privacy/ccpa

jdwieland8282:
Docs pr: https://github.com/prebid/prebid.github.io/pull/4114

PR #9123, opened 2022-10-17
FTRACK USER ID MODULE: tweaking the createEidsArray() method to accept two schema patterns

ftxmoJason:
## Type of change
- [X] Bugfix

## Description of change
A partner was attempting to use the FTRACK User ID module and they were getting errors in the `createEidsArray` method. The issue was the JSON schema being passed into the method did not match the schema we wrote the code for. The change this PR addresses is the JSON parsing logic was tweaked to accept both schema patterns while also checking value type (string|array) before pushing to the `eids` array that is returned

### DIFF
--- a/modules/ftrackIdSystem.js
+++ b/modules/ftrackIdSystem.js
@@ -48,20 +48,36 @@ export const ftrackIdSubmodule = {
    *   similar to the module name and ending in id or Id
    */
   decode (value, config) {
-    if (!value) { return }
-    const ext = {}
-
-    for (var key in value) {
-      /** unpack the strings from the arrays */
-      ext[key] = value[key][0]
-    }
-
-    return {
+    value = value || {};
+    const DECODE_RESPONSE = {
       ftrackId: {
-        uid: value.DeviceID && value.DeviceID[0],
-        ext
+        uid: '',
+        ext: {}
       }
     }
+
+    // Loop over the value's properties:
+    // -- if string, assign value as is.
+    // -- if array, convert to string then assign value.
+    // -- If neither type, assign value as empty string
+    for (var key in value) {
+      let keyValue = value[key];
+      if (Array.isArray(keyValue)) {
+        keyValue = keyValue.join('|');
+      } else if (typeof value[key] !== 'string') {
+        // Unexpected value type, should be string or array
+        keyValue = '';
+      }
+
+      DECODE_RESPONSE.ftrackId.ext[key] = keyValue;
+    }
+
+    // If we have DeviceId value, assign it to the uid property
+    if (DECODE_RESPONSE.ftrackId.ext.hasOwnProperty('DeviceID')) {
+      DECODE_RESPONSE.ftrackId.uid = DECODE_RESPONSE.ftrackId.ext.DeviceID;
+    }
+
+    return DECODE_RESPONSE;
   },
 
   /**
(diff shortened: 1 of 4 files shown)


modules/userId/eids.js
```diff
@@ -420,16 +420,26 @@ export function createEidsArray(bidRequestUserId) {
...
         // ftrack has multiple IDs so we add each one that exists
-        let eid = {
+        const E_ID = {
           'atype': 1,
-          'id': (bidRequestUserId.ftrackId.DeviceID || []).join('|'),
           'ext': {}
+        };
+        const IDS = bidRequestUserId.ftrackId.hasOwnProperty('ext') ? bidRequestUserId.ftrackId.ext : bidRequestUserId.ftrackId;
+
+        for (let id in IDS) {
```
ftxmoJason, line 429: To whom it may review.

1. The original logic expected `bidRequestUserId.ftrackId` to be the following schema.
```
ftrackId: {
  HHID: ['household_test_id'],
  DeviceID: ['device_test_id'],
  SingleDeviceID: ['single_device_test_id']
}
```
2. But during testing, we actually were seeing:
```
ftrackId: {
  uid: 'device_test_id',
  ext: {
    HHID: 'household_test_id',
    DeviceID: 'device_test_id',
    SingleDeviceID: 'single_device_test_id'
  }
}
```
This code change allows both JSON schema's to work correctly, but its not clear to me if both are needed. Do we know why the schema's are different and/or if one is incorrect?

Thanks!

dgirardi: The ftrack ID module is what wraps the result in `ext`:

https://github.com/prebid/Prebid.js/blob/d8f25b92ed2d04ab578681076a35217548fc5df3/modules/ftrackIdSystem.js#L59-L64
https://github.com/prebid/Prebid.js/pull/8678

To me that looks odd, maybe unintentional - it'd make more sense to return `{ftrack: {hhid, deviceId, singleDeviceId}}`; but in the end what matters is that the callers (including the eid conversion logic) know what to expect.

dgirardi: then I think you can simplify this to `E_ID.ext = bidRequestUserId.ftrackId.ext`; I don't see a need to handle other cases.

ftxmoJason: Thanks @dgirardi , I concur. Our decode method is already doing that work so its def redundant at this step. I'll make those changes and ping when they are up. Thanks

modules/ftrackIdSystem.js
```diff
@@ -48,20 +48,36 @@ export const ftrackIdSubmodule = {
    *   similar to the module name and ending in id or Id
    */
   decode (value, config) {
-    if (!value) { return }
-    const ext = {}
+    value = value || {};
```
dgirardi, line 51: I think it's better to keep the short circuit (`if (!value) return`). From what I understand `value` should always be present here, but if it isn't, it's not correct to default to `{}`: it will cause this logic to return an incomplete object, which would then be translated to essentially gibberish in `eids`.

If this returns undefined, the caller will filter it out from the userID map.

ftxmoJason: Thanks @dgirardi reverted this code

dgirardi (approved):

mmoschovas (approved):

ftxmoJason:
@patmmccann @mmoschovas @dgirardi I forget, do I need to do something now or is the next step on your side? Thanks

Issue #9126, opened 2022-10-18
Bid-level renderer not working


ACannuniRP:
## Type of issue
Bug

## Description
Renderer can be inserted at ad-unit-level, bid-level (bidder-level) or media-type-level
If inserted at bid-level (bidder-level), the renderer does no longer work.
The breaking change is between version 6.9.0 and 6.10.0 and related to #7885.
The [lines](https://github.com/prebid/Prebid.js/pull/7885/files#diff-e9f987f170d6e2065c627eb6a145180770dc0921cb2c82f0c0a549ae87934a4d) causing the break are in `/src/auction`:
```
-  const bidReq = bidderRequest.bids && find(bidderRequest.bids, bid => bid.adUnitCode == adUnitCode && bid.bidId == bidObject.requestId);
-  const adUnitRenderer = bidReq && bidReq.renderer;
+  const adUnitRenderer = index.getAdUnit(bidObject).renderer;
```

This change is only affecting bid-level (bidder-level) renderers and not ad-unit-level nor media-type-level.

## Steps to reproduce
1. Open the browser console;
1. Open [the test page using Prebid.js version 6.9.0](https://magnite-acannuni.s3.eu-west-1.amazonaws.com/test-pages/prebid-interstitial-bidder-6.9.0-1.html);
1. You will notice that the renderer object is included in both "Bids requested" and "Bids Received";
1. Open [the test page using Prebid.js version 6.10.0](https://magnite-acannuni.s3.eu-west-1.amazonaws.com/test-pages/prebid-interstitial-bidder-6.10.0-1.html);
1. You will notice that the renderer object is included in both "Bids requested" but not in "Bids Received";
1. Open [the test page using Prebid.js version 6.10.0 but with line 557 replace with lines 562 and 563 from version 6.9.0](https://magnite-acannuni.s3.eu-west-1.amazonaws.com/test-pages/prebid-interstitial-bidder-6.10.0-edited.html);
1. You will notice that the renderer object is included in both "Bids requested" but not in "Bids Received";

## Test page
- [Test page using Prebid.js version 6.9.0](https://magnite-acannuni.s3.eu-west-1.amazonaws.com/test-pages/prebid-interstitial-bidder-6.9.0-1.html)
- [Test page using Prebid.js version 6.10.0](https://magnite-acannuni.s3.eu-west-1.amazonaws.com/test-pages/prebid-interstitial-bidder-6.10.0-1.html)
[...]

PR #9128, opened 2022-10-18
Prebid core: restore bid-level `renderer`

dgirardi:
## Type of change

- [x] Bugfix

## Description of change

Allow `renderer` to be specified at the bid-level. This was working, although (to the best of my knowledge) undocumented, until 6.9.

## Other information

Fixes https://github.com/prebid/Prebid.js/issues/9126

### DIFF
--- a/src/auction.js
+++ b/src/auction.js
@@ -690,7 +690,7 @@ function getPreparedBidForAuction(bid, {index = auctionManager.index} = {}) {
   events.emit(CONSTANTS.EVENTS.BID_ADJUSTMENT, bid);
 
   // a publisher-defined renderer can be used to render bids
-  const adUnitRenderer = index.getAdUnit(bid).renderer;
+  const bidRenderer = index.getBidRequest(bid)?.renderer || index.getAdUnit(bid).renderer;
 
   // a publisher can also define a renderer for a mediaType
   const bidObjectMediaType = bid.mediaType;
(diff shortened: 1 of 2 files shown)


ChrisHuie (approved):

musikele (approved):
LGTM

src/auction.js
```diff
@@ -690,7 +690,7 @@ function getPreparedBidForAuction(bid, {index = auctionManager.index} = {}) {
   events.emit(CONSTANTS.EVENTS.BID_ADJUSTMENT, bid);
 
   // a publisher-defined renderer can be used to render bids
-  const adUnitRenderer = index.getAdUnit(bid).renderer;
+  const bidRenderer = index.getBidRequest(bid)?.renderer || index.getAdUnit(bid).renderer;
```
patmmccann, line 693: Is the rendering order of priority of all four are present the following? 

Mediatype
Ad unit
Bid
Adapter

And if backup only is specified the order is:
Bid
Adapter
Media type
Ad unit

Correct?

patmmccann: @fowler446 perhaps you've spent the most time on this recently, is this your understanding as well?

dgirardi: The first one of these wins:

1. mediaType
2. adUnit.bid[]
3. adUnit
4. bid response


if any of those say `backupOnly`, they will be disregarded as long as the bid response has its own renderer.

Issue #8835, opened 2022-08-15
Publisher Analytics Adapter


patmmccann:
## Type of issue
Feature Request

## Description
Create vendorless (from gdpr perspective) analytics adapter that sends analytics events to a configurable endpoint. It would need purpose 7 consent to the publisher to not be enforced on. It should allow sampling of events. It should allow configurable event list filtering. It should allow for get or post to both work. 

Publishers would be responsible for making sense out of server access logs, or perhaps some future utility or vendor utility could be published to process the logs.

Issue #6564, opened 2021-04-09
Prebid.js GDPR feature: softVendorException


bretg:
## Type of issue
enhancement

## Description
In Prebid Server, Xandr implemented a compromise version of the original TCF2 "vendorException" setting. Adding a vendor to this list was a complete "hall pass", exempting that vendor from any checks, essentially trusting that they do the right thing on their side.

The new feature is a weaker version of that "hall pass" -- the user must still consent to the purpose, but vendor consent isn't checked. The use case here is for vendors that don't have GVLIDs -- they currently need exceptions, but the only existing exception is a strong one. This softer version allows pubs to support bidders without GVLIDs without granting the ability to bypass the user's purpose consent.

See the highlighted changes in https://docs.google.com/document/d/1fBRaodKifv1pYsWY3ia-9K96VHUjd8kKvxZlOsozm8E/edit#heading=h.hlpacpauqwkx

The proposal is to add gdpr.rules[].softVendorExceptions: "If enforceVendor is true, then any vendors in this list will ignore enforcement."

Looking for community feedback on this enhanced level of flexibility and open to suggestions about a better name for this field.

bretg:
@jsnellbaker - is this still an issue for Xandr?

jsnellbaker:
Hi @bretg From what I understand from discussing this with some others, this is still a feature on the PBS side that's still in-use.  So if there's a need to bring the functionality here into PBJS as well, I think that would good to do.

PR #7805, opened 2021-12-03
IX Bid Adapter: fix for infinite loop

lksharma:
## Type of change

- [x] Bugfix
- [ ] Feature
- [ ] New bidder adapter   
- [ ] Code style update (formatting, local variables)
- [ ] Refactoring (no functional changes, no api changes)
- [ ] Build related changes
- [ ] CI related changes
- [ ] Does this change affect user-facing APIs or examples documented on http://prebid.org?
- [ ] Other

## Description of change
This MR addresses an infinite loop bug within the IX adapter. The endless loop happens when the `gdpr enforcement` module is included in the build along with the IX adapter. 

When IX is not given permission to write to local storage an `AUCTION_DEBUG` event is emitted. IX has a handler for this particular event. The handler calls the `localStorageIsEnabled` function, which also triggers an `AUCTION_DEBUG` event. The back and forth between the handler and the `AUCTION_DEBUG` event causes an endless loop to occur. 

## Other information
Issue: https://github.com/prebid/Prebid.js/issues/7707

PR #9134, opened 2022-10-19
Generic Analytics Adapter: initial release

dgirardi:
## Type of change

- [x] Feature

## Description of change

This is a new generic analytics adapter, meant for publishers that have their own infrastrucutre for collecting analytics data. The adapter is "vendorless" for the purposes of GDPR enforcement (in the same way that sharedId is).

## Other information

Closes https://github.com/prebid/Prebid.js/issues/8835
Closes https://github.com/prebid/Prebid.js/issues/6564
Documentation: https://github.com/prebid/prebid.github.io/pull/4092

### DIFF
--- a/modules/genericAnalyticsAdapter.js
+++ b/modules/genericAnalyticsAdapter.js
@@ -15,6 +15,7 @@ const TYPES = {
   handler: 'function',
   batchSize: 'number',
   batchDelay: 'number',
+  gvlid: 'number',
 }
 
 const MAX_CALL_DEPTH = 20;
@@ -103,6 +104,9 @@ export function GenericAnalytics() {
   return Object.assign(
     Object.create(parent),
     {
+      gvlid(config) {
+        return config?.options?.gvlid
+      },
       enableAnalytics(config) {
         if (optionsAreValid(config?.options || {})) {
           options = Object.assign({}, DEFAULTS, config.options);
@@ -151,5 +155,4 @@ export function defaultHandler({url, method, batchSize, ajax = ajaxBuilder()}) {
 adapterManager.registerAnalyticsAdapter({
   adapter: GenericAnalytics(),
   code: 'generic',
-  gvlid: VENDORLESS_GVLID,
 });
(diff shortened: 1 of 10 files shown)


modules/genericAnalyticsAdapter.js
```diff
@@ -0,0 +1,155 @@
...
+    {
+      enableAnalytics(config) {
+        if (optionsAreValid(config?.options || {})) {
+          options = Object.assign({}, DEFAULTS, config.options);
+          handler = options.handler || defaultHandler(options);
+          translate = translator(options.events);
+          parent.enableAnalytics.call(this, config);
+        }
+      },
+      track(event) {
```
patmmccann, line 114: is it straightforward to see which events a particular analytics adapter subscribes to?

modules/genericAnalyticsAdapter.js
```diff
@@ -0,0 +1,155 @@
...
+
+  function processBatch() {
+    const currentBatch = batch;
+    batch = [];
+    callDepth++;
+    try {
+      // the pub-provided handler may inadvertently cause an infinite chain of events;
+      // even just logging an exception from it may cause an AUCTION_DEBUG event, that
+      // gets back to the handler, that throws another exception etc.
+      // to avoid the issue, put a cap on recursion
```
patmmccann, line 71: these infinite loops have occurred before, eg https://github.com/prebid/Prebid.js/pull/7805/files ; should we more generally defend against them?

dgirardi: There is a more general prevention mechanism in https://github.com/prebid/Prebid.js/pull/9113. That also allows the pub to set which events should be forwarded to any analytics adapter. However, adapters can choose to ignore some or circumvent the default event tracking system; so in general, the only way to find out which events are tracked is reading code. For this particular adapter though control is entirely on the publisher.

dgirardi:
Agreed offline to make this accept a gvlid to make it truly generic; we can keep the "generic" name until a better one materializes.

dgirardi:
Pausing this with the intention of tackling https://github.com/prebid/Prebid.js/issues/6564 as the solution for making this "vendorless".

Fawke (approved):
LGTM! Good work.

PR #9135, opened 2022-10-20
AAX Blockmeter RTD Module : Initial Release

product-aax:
## Type of change

- [x] New RTD Provider   

## Description of change

The module enables publishers to measure traffic coming from visitors using adblockers.

- [x] Official Adapter Submission
- contact email of the adapter’s maintainer: product@aax.media
- Link to dev-docs [AAX Blockmeter Dev Documentation prebid.github.io#4095](https://github.com/prebid/prebid.github.io/pull/4095)

### DIFF
--- a/modules/aaxBlockmeterRtdProvider.js
+++ b/modules/aaxBlockmeterRtdProvider.js
@@ -1,20 +1,35 @@
-import {isEmptyStr, isStr, logError, isFn} from '../src/utils.js';
+import {isEmptyStr, isStr, logError, isFn, logWarn} from '../src/utils.js';
 import {submodule} from '../src/hook.js';
 import { loadExternalScript } from '../src/adloader.js';
 
-const MODULE = 'aaxBlockmeter';
-const ADSERVER_TARGETING_KEY = 'atk';
-const BLOCKMETER_URL = 'https://c.aaxads.com/aax.js';
+export const _config = {
+  MODULE: 'aaxBlockmeter',
+  ADSERVER_TARGETING_KEY: 'atk',
+  BLOCKMETER_URL: 'c.aaxads.com/aax.js',
+  VERSION: '1.2'
+};
 
 window.aax = window.aax || {};
 
-function loadBlockmeter(config) {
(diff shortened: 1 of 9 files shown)


modules/aaxBlockmeterRtdProvider.js
```diff
@@ -0,0 +1,45 @@
+import {isEmptyStr, isStr, logError, isFn} from '../src/utils.js';
+import {submodule} from '../src/hook.js';
+import { loadExternalScript } from '../src/adloader.js';
+
+const MODULE = 'aaxBlockmeter';
+const ADSERVER_TARGETING_KEY = 'atk';
+const BLOCKMETER_URL = 'https://c.aaxads.com/aax.js';
```
patmmccann, line 7: Please make this a configuration param. If you fall back to it when missing, please lock it to a version. We want publishers using your module to be extremely aware there is js loaded that the project hasn't warranted or reviewed. Please also add an external js disclosure to your docs PR.

product-aax: Hi, @patmmccann, have done the changes please r[eview the changes](https://github.com/prebid/Prebid.js/pull/9135/commits/c425d7af0e8c43a1d9d52ecd873f4b1ddf39af2a) and [docs](https://github.com/prebid/prebid.github.io/pull/4095/commits/5d1b65fec6b723bc963cca038628691bad1009e6)

modules/aaxBlockmeterRtdProvider.md
```diff
@@ -0,0 +1,48 @@
+## Overview
+
+Module Name: AAX Blockmeter Realtime Data Module  
+Module Type: Rtd Provider  
+Maintainer: product@aax.media  
+
+## Description
+
+The module enables publishers to measure traffic coming from visitors using adblockers.
```
mkendall07, line 9: I think this requires an account with AAX yes? Might be better to state that here, as it doesn't look like the module itself does much more than load the AAX script.

mkendall07 (approved):
Generally I think this looks OK as long as the intention is made clear and the proper warnings are in place with loading external JS.

PR #9153, opened 2022-10-25
JW Player RTD Module: prefer segment.id to segment.value

karimMourra:
## Type of change

- [x] Bugfix

## Description of change
To reduce payload size we have decided to drop `segment.id`. `segment.id` was always identical to `segment.value` therefore there is no benefit to including it. 

Merging is blocked until all partners have communicated that this will not require changes on their end.

### DIFF
--- a/modules/jwplayerRtdProvider.md
+++ b/modules/jwplayerRtdProvider.md
@@ -101,9 +101,9 @@ Example:
                         cids: ['abc123']
                     },
                     segment: [{ 
-                        id: '123'
+                        value: '123'
                     }, { 
-                        id: '456'
+                        value: '456'
                     }]
                 }]
             }
@@ -122,7 +122,7 @@ where:
         - `ext.segtax` whose `502` value is the unique identifier for JW Player's proprietary taxonomy
         - `ext.cids` is an array containing the list of extended content ids as defined in [oRTB's community extensions](https://github.com/InteractiveAdvertisingBureau/openrtb/blob/master/extensions/community_extensions/extended-content-ids.md#example---content-id-and-seller-defined-context). 
         - `segment` is an array containing the segment taxonomy values as an object where:
-          - `id` is the string representation of the data segment value. 
+          - `value` is the string representation of the data segment. 
   
 **Example:**
 
--- a/modules/jwplayerRtdProvider.md
+++ b/modules/jwplayerRtdProvider.md
@@ -101,9 +101,9 @@ Example:
                         cids: ['abc123']
                     },
                     segment: [{ 
-                        value: '123'
+                        id: '123'
                     }, { 
-                        value: '456'
+                        id: '456'
                     }]
                 }]
             }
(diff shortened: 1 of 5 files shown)


patmmccann:
Please remove segment.value, segment.id is correct

modules/jwplayerRtdProvider.md
```diff
@@ -122,7 +122,7 @@ where:
         - `ext.segtax` whose `502` value is the unique identifier for JW Player's proprietary taxonomy
         - `ext.cids` is an array containing the list of extended content ids as defined in [oRTB's community extensions](https://github.com/InteractiveAdvertisingBureau/openrtb/blob/master/extensions/community_extensions/extended-content-ids.md#example---content-id-and-seller-defined-context). 
         - `segment` is an array containing the segment taxonomy values as an object where:
-          - `id` is the string representation of the data segment value. 
+          - `value` is the string representation of the data segment.
```
patmmccann, line 125: id is correct according to the spec.

patmmccann (changes requested):

modules/jwplayerRtdProvider.md
karimMourra: I can't tell which is better from the spec:
id: ID of the data segment specific to the data provider.
value: String representation of the data segment value.

Is `id` more commonly used ?

patmmccann: id is in the spec, value is not https://github.com/InteractiveAdvertisingBureau/openrtb/blob/master/extensions/community_extensions/segtax.md

patmmccann (approved):

jlquaccia (approved):
LGTM

TheMediaGrid:
Confirming Grid can operate using segment.id with no additional changes on the backend

patmmccann:
test page: https://cookieandkate.com/butternut-squash-mac-and-cheese/?pbjs_debug=true&commit=afa028d

site.content.data.segment.id exists but the jw user segments are on user.data.segment.value, which is incorrect.

patmmccann:
@karimMourra fyi

TheMediaGrid:
> @TheMediaGrid perhaps https://github.com/prebid/Prebid.js/pull/9061/files caused the bug
I've checked the link above and it leads to multiRequest feature, which does not affect segments format in which we send user.data.segment. So we can make additional fix, and start sending segment in format user.data.segment = [{ id, name }...] instead of [{ value, name }...]. Will this fix be enough to resolve the issue?

karimMourra:
@TheMediaGrid why would the segments be passed in `user.data` ? These segments are not generated with any user data, they are purely contextual as they are specific to the media.

TheMediaGrid:
Hey @karimMourra and @patmmccann 

Karim, you are right, there segments shouldn't be passed in user.data
Pat, I am looking at your example page here https://cookieandkate.com/butternut-squash-mac-and-cheese/?pbjs_debug=true&commit=afa028d

and I can see that the JWP segments are properly passed via site.content.data, please see the screenshot below
Screenshot 2023-02-17 at 16 39 12

I am afraid we would need to have a page example showing how segments are missing from site.content.data to better locate the issue

TheMediaGrid:
I noticed though that some subsequent calls to /hbjson indeed don't contain site.content.data.segment.id for JWP segments and they only contain site.content.data.name = "jwplayer.com"
image

Any ideas on why segments might not bee fetched between subsequent calls? Aren't they cached somewhere?

PR #9158, opened 2022-10-27
Concert Bid Adapter: Update localStorage name-spacing for Concert UID

BrettBlox:
## Type of change
- [ ] Bugfix
- [ ] Feature
- [ ] New bidder adapter   
- [ ] Build related changes
- [ ] CI related changes
- [ ] Other

## Description of change
- adds check for sharedId falls back to existing method of none exists
- swaps out localStorage key `c_uid` with `vmconcert_uid` to avoid potential collisions with vendors
- fixes broken tests related to localStorage
- updates test with new localStorage key name


patmmccann:
In general, this is bad netizenship. You generate a uuid and set it in the first party domain. Why not check for pubcommonid, which is the exact same thing and use that and fallback to this when missing?

patmmccann:
@jdwieland8282 as fyi; as these examples start to proliferate, should we crack down on them?

jdwieland8282:
Thanks for surfacing this, yes we should, absolutely.

jdwieland8282:
@BrettBlox what are your reservations about checking for and using sharedid? Using a universal open source id namespace (pubcid.org) has many advantages over a bespoke one.

BrettBlox:
@jdwieland8282 I will look into it. Thanks.

BrettBlox:
Updated to check for sharedId.

modules/concertBidAdapter.js
```diff
@@ -188,10 +188,24 @@ function getUid(bidderRequest) {
     return false;
   }
 
-  const CONCERT_UID_KEY = 'c_uid';
+  const sharedId = deepAccess(bidderRequest, 'userId._sharedid.id');
 
+  if (sharedId) {
+    return sharedId;
```
patmmccann, line 194: Awesome, ty

mmoschovas (approved):

Issue #9037, opened 2022-09-26
Prebid file not built properly


AndBeyondMediaHB:
## Type of issue
bug

## Description
When I try to download PrebidJs with the selected Compass adapter, I get an error:
`{"error":"Prebid file not built properly","requestId":"e9179e40-85cb-4ba1-a0bd-4f03152ec8c0"}`

## Steps to reproduce
1. Go to https://docs.prebid.org/download.html
2. Select Compass adapter (biddercode - `beyondmedia`)
3. Download Prebid.js

dgirardi:
@AndBeyondMediaHB, what's the relationship between Compass and AndBeyondMedia? 

There are two conflicting documentation files (which are used for the download page): both [andBeyondMedia](https://github.com/prebid/prebid.github.io/blob/master/dev-docs/bidders/andBeyondMedia.md) and [compass](https://github.com/prebid/prebid.github.io/blob/master/dev-docs/bidders/compass.md) declare a title of "Compass", so only one of them will show up - I'm not sure if the precedence is well defined. AndBeyond media also has an incorrect `biddercode`, it should be `andBeyondMedia`, this is what's causing the error.

bretg:
Docs issue https://github.com/prebid/prebid.github.io/pull/4018

AndBeyondMediaHB:
> @AndBeyondMediaHB, what's the relationship between Compass and AndBeyondMedia?
> 
> There are two conflicting documentation files (which are used for the download page): both [andBeyondMedia](https://github.com/prebid/prebid.github.io/blob/master/dev-docs/bidders/andBeyondMedia.md) and [compass](https://github.com/prebid/prebid.github.io/blob/master/dev-docs/bidders/compass.md) declare a title of "Compass", so only one of them will show up - I'm not sure if the precedence is well defined. AndBeyondMedia also has an incorrect `biddercode`, it should be `andBeyondMedia`, this is what's causing the error.

@dgirardi thanks for the answer. @bretg attached a pull request to change the title from Compass to AndBeyondMedia. Did I understand correctly that the biddercode must be the same as the adapter title in lower case? Or am I confusing with the adapter's .js file name?

dgirardi:
@AndBeyondMediaHB, for the website and download page, `biddercode` is just the file name minus `bidAdapter`. I think we recommend naming them the same as `spec.code`, hence the confusion.

AndBeyondMediaHB:
@dgirardi thanks, it's all clear now

[...]

PR #9164, opened 2022-10-28
AcuityAds adapter: fix issue with download

AcuityAdsIntegrations:
issue - `{"error":"Prebid file not built properly","requestId":"f4d9462b-43d3-401e-8cf3-8881b89340a8"}`


AcuityAdsIntegrations:
@ChrisHuie please react

ChrisHuie:
@AcuityAdsIntegrations Are you getting this error from the download page? Name capitalization shouldn't make a difference since it is consistent.

AcuityAdsIntegrations:
> @AcuityAdsIntegrations Are you getting this error from the download page? Name capitalization shouldn't make a difference since it is consistent. This would be a breaking change for anyone pointing to the current file namespace.

Yes, from [download page](https://docs.prebid.org/download.html). I followed the [following](https://github.com/prebid/Prebid.js/issues/9037) issue

patmmccann (approved):

bretg:
Rather than create a breaking change for publishers, we added a new feature to the download page that allows the bidders filename to differ from the biddercode. https://github.com/prebid/prebid.github.io/pull/4186

Issue #9142, opened 2022-10-21
Pubmatic adapter should use site.content.data instead of private dctr param


bretg:
## Type of issue
efficiency

## Description
We found that there can be many hundreds of Seller-Defined Audience segments defined in a request in site.content.data.

Further, we noticed that pubmatic is copying all of them to a bidder-specific dctr parameter before sending to Prebid Server

```
  transformBidParams: function (params, isOpenRtb, adUnit, bidRequests) {
   // wow - this doubles the size of an already XXXXXXL dataset.
    _addJWPlayerSegmentData(params, adUnit.bids[0], true);
    return convertTypes({
      'publisherId': 'string',
      'adSlot': 'string'
    }, params);
  }
};
```

Here's an example:
```
imp[0].ext.prebid.bidder:{
                        "pubmatic": {
                            "publisherId": "1111111",
                            "dctr": "iris_context=ic_1670398,ic_9534888,ic_3129871,ic_8322846,ic_4260700,ic_4962242,ic_9954675,ic_2592227,ic_8121076,ic_2001389,ic_0381735,ic_1592380,ic_2177941,ic_2416407,ic_6703731,ic_7753435,ic_1612662,ic_8328276,ic_0344266|jw-id=jw_NkSk6w4a|jw-80011026=1|jw-80011062=1|jw-80011064=1|jw-80011065=1|jw-80011079=1|jw-80011080=1|jw-80011081=1|jw-80011082=1|jw-CMTPI-001553FA=1|jw-CMTPI-0059DF08=1|jw-CMTPI-0081098E=1|jw-CMTPI-0142FEF1=1|jw-CMTPI-022AFFD7=1|jw-CMTPI-02947E6F=1|jw-CMTPI-03654CCA=1| ... 25KB more data ...
   ...
site: {
        "content": {
            "data": [
                {
                    "name": "jwplayer.com",
                    "segment": [
                        {
                            "id": "80011026",
                            "value": "80011026"
                        },
                        {
                            "id": "80011062",
                            "value": "80011062"
                        },
                        {
                            "id": "80011064",
                            "value": "80011064"
                        },
                        {
                            "id": "80011065",
                            "value": "80011065"
                        },
                        {
                            "id": "80011079",
                            "value": "80011079"
                        },
                        {
                            "id": "80011080",
                            "value": "80011080"
                        },
                        {
                            "id": "80011081",
                            "value": "80011081"
                        },
                        {
                            "id": "80011082",
[...]

PR #9169, opened 2022-10-28
Pubmatic Bid Adapter: Removed logic to append JW player data to the "dctr" param

jlquaccia:
## Type of change
- [x] Bugfix

## Description of change
Removed functionality to copy the JW player data to the dctr param in the pubmaticBidAdapter when using PBS

## Other information

Related Issue: https://github.com/prebid/Prebid.js/issues/9142
@dgirardi @patmmccann could you review?


jlquaccia:
@dgirardi @patmmccann also note that I am still waiting to here back from some of the PubMatic team who work on the server-side to confirm if there is any server-side impact on this change.  So for the time being until then, we can hold off on merging.

patmmccann (approved):

patmmccann:
Cc @bretg

patmmccann:
@jlquaccia any update?

jlquaccia:
@patmmccann sounds like there is a server-side impact to this on our end, which I'm trying to follow up with with a few people over here.  I'll keep you in the loop once I've got another update.

Issue #9159, opened 2022-10-27
Ftrack eid object incorrectly formatted


mmoschovas:
## Type of issue
Bug

## Description
createEidsArray method should return an array of objects with the format:
```
[{
    source: 'IDSYSTEM_SOURCE_NAME',
    uids: [...]
}]
```
where uids is an array of id data. Currently, the ftrack module returns the id data only without source/uids properties as seen below:

```
[{
    atype: 1,
    ext: {
        HHID: 'HHID', 
        DeviceID: 'DEVICEID', 
    },
    id: 'VALUE'
}]
```

## Steps to reproduce
  Run `pbjs.getUserIdsAsEids()` with ftrackIdSystem configured

mmoschovas:
@ftxmoJason can you review the above issue please when you have a moment

ftxmoJason:
Thanks @mmoschovas . Please see: https://github.com/prebid/Prebid.js/pull/9174 and let us know if we missed the mark, thanks.

PR #9174, opened 2022-10-31
FtrackId: correcting the eid object schema

ftxmoJason:
## Type of change
- [X] Bugfix

## Description of change
Fixing the EID schema returned for FtrackID per https://github.com/prebid/Prebid.js/issues/9159

### DIFF
--- a/modules/userId/eids.js
+++ b/modules/userId/eids.js
@@ -88,13 +88,13 @@ export const USER_IDS_CONFIG = {
     atype: 1,
     getValue: function(data) {
       let value = '';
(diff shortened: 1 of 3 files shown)


ftxmoJason:
Running the lint command locally is not showing these issues...

ftxmoJason:
Mediavine made a recommendation to change the eids.js logic in a way that we no longer needed a condition for ftrack in the  createEidsArray method so we made that change in this PR. Just added it.

mmoschovas (approved):

PR #9192, opened 2022-11-02
gumgum Bid Adapter: support coppa config

john-ivan:
## Type of change

- [x] Feature

## Description of change
Added ability to send coppa flag to gumgum ad server if set to true

### DIFF
--- a/modules/gumgumBidAdapter.js
+++ b/modules/gumgumBidAdapter.js
@@ -294,6 +294,7 @@ function buildRequests(validBidRequests, bidderRequest) {
   const gdprConsent = bidderRequest && bidderRequest.gdprConsent;
   const uspConsent = bidderRequest && bidderRequest.uspConsent;
   const timeout = config.getConfig('bidderTimeout');
+  const coppa = config.getConfig('coppa') === true ? 1 : 0;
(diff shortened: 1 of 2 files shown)


patmmccann (approved):

modules/gumgumBidAdapter.js
```diff
@@ -294,6 +294,7 @@ function buildRequests(validBidRequests, bidderRequest) {
   const gdprConsent = bidderRequest && bidderRequest.gdprConsent;
   const uspConsent = bidderRequest && bidderRequest.uspConsent;
   const timeout = config.getConfig('bidderTimeout');
+  const coppa = config.getConfig('coppa') === true ? 1 : 0;
```
patmmccann, line 297: fwiw the line above this is the wrong way to get timeout; the timeout from requestBids can be overridden in https://docs.prebid.org/dev-docs/publisher-api-reference/requestBids.html

For example, I think CafeMedia always overrides.

john-ivan: @patmmccann thanks for the review and for catching this issue. I'll make a ticket to review if we're using it or not and make the necessary corrections based on those findings.

Issue #9191, opened 2022-11-02
return a promise from requestBids causes issues along with criteoId


muuki88:
## Type of issue

prebid.js crashes while starting a second auction, because `adUnits` is `undefined`.
This _only_ happens when the `criteoIdSystem` is part of the prebid.js distribution.

## Description

Upgrading to 7.23.0 and having the `criteoIdSystem` active causes prebid.js to crash with

```
prebid.js:237 Uncaught (in promise) TypeError: Cannot read properties of undefined (reading 'forEach')
    at prebid.js:237:11
    at order.<computed> (index.js:342:1)
    at next (index.js:328:1)
    at preloadBidderMappingFile (bidderFactory.js:476:15)
    at next (index.js:328:1)
    at Object.trap [as apply] (index.js:347:1)
    at prebid.js:651:75
    at Object.measureTime (perfMetrics.js:127:16)
    at prebid.js:651:33
    at order.<computed> (index.js:342:1)
```

I suspect that there's an issue with #9106 as the hook is now `sync` and not `async` (also I do not know what this implies).

While trying to build a minimal example, I figured that removing `criteoIdSystem` fixed this issue.
Now I'm really puzzled as I thought #9106 may introduced this, but I have no idea why the criteoIdSystem changes this behaviour as well.

## Steps to reproduce

Run an "ad pipeline" (or auction pipeline) for a set of ad units. 

1. Call `pbjs.addAdUnits` for the specified adUnits
2. Call `bjs.requestBids({  adUnitCodes: adUnitCodes, timeout: ..., bidsBackHandler => /* do the gpt targeting stuff */`

## Test page

https://frag-muki.de/prebid/bucket-issue/

### Expected results

No Crash

### Actual results

![Crash](https://media.giphy.com/media/pyJsNhs040q79RNeVA/giphy.gif)

## Platform details

- prebid 7.23.0

## Other information

- #9106

dgirardi:
@muuki88, I see the issue on your test page but I'm having trouble isolating it. Would you be able to build it with unminifed Prebid? Or alternatively, if you can confirm that reverting #9106 fixes it, we can roll that back as a first step. 

I am not able to reproduce locally using this test page:

```
<!--
  This page calls a single bidder for a single ad slot. It can be considered a "hello world" example for using
  Prebid with the Google Publisher Tag.
  It also makes a good test page for new adapter PR submissions. Simply set your server's Bid Params object in the
[...]

PR #9106, opened 2022-10-12
Prebid core: return a promise from `requestBids`

dgirardi:
## Type of change

- [x] Feature

## Description of change

Return a promise from `requestBids`, resolving to the same arguments passed to `bidsBackHandler`. 

## Other information

Resolves https://github.com/prebid/Prebid.js/issues/8478
Documentation: https://github.com/prebid/prebid.github.io/pull/4073

PR #9194, opened 2022-11-02
Revert "Prebid core: return a promise from `requestBids` (#9106)"

dgirardi:
## Type of change
- [x] Other

## Description of change
This reverts commit 64aff9b1af9bb1b2761e1012f628299f16c0eb46.

Temporary fix for https://github.com/prebid/Prebid.js/issues/9191 while I track down the root cause.

### DIFF
--- a/src/prebid.js
+++ b/src/prebid.js
@@ -622,7 +622,7 @@ $$PREBID_GLOBAL$$.removeAdUnit = function (adUnitCode) {
  * @alias module:pbjs.requestBids
  */
 $$PREBID_GLOBAL$$.requestBids = (function() {
-  const delegate = hook('sync', function ({ bidsBackHandler, timeout, adUnits, adUnitCodes, labels, auctionId, ttlBuffer, ortb2, metrics } = {}) {
+  const delegate = hook('async', function ({ bidsBackHandler, timeout, adUnits, adUnitCodes, labels, auctionId, ttlBuffer, ortb2, metrics } = {}) {
     events.emit(REQUEST_BIDS);
     const cbTimeout = timeout || config.getConfig('bidderTimeout');
     logInfo('Invoking $$PREBID_GLOBAL$$.requestBids', arguments);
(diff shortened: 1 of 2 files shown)


dgirardi:
@ChrisHuie and reviewers, please leave this open until it's time to release, I'll see if I can update this to be a proper fix by then, otherwise please merge and release.

src/prebid.js
```diff
@@ -642,11 +643,13 @@ $$PREBID_GLOBAL$$.requestBids = (function() {
     req.metrics.checkpoint('requestBids');
     let adUnits = req.adUnits || $$PREBID_GLOBAL$$.adUnits;
     req.adUnits = (isArray(adUnits) ? adUnits.slice() : [adUnits]);
-    return delegate.call(this, req);
+    req.defer = defer({promiseFactory: (r) => new Promise(r)})
+    delegate.call(this, req);
+    return req.defer.promise;
```
muuki88, line 648: Maybe a few comments why this is necessary would be helpful.

As someone only looking at bid adapters mostly the fun-hook framework and the custom promise implementation this code is non-obivous why it's there.

test/spec/unit/pbjs_api_spec.js
```diff
@@ -1665,7 +1677,7 @@ describe('Unit: Prebid Module', function () {
               sinon.assert.match(bids[bid.adUnitCode].bids[0], bid)
               done();
             });
-            completeAuction([bid]);
+            setTimeout(() => completeAuction([bid]));
```
muuki88, line 1680: also a small comment here why wrapping in setTimeout is required here

src/prebid.js
```diff
@@ -622,15 +623,15 @@ $$PREBID_GLOBAL$$.removeAdUnit = function (adUnitCode) {
  * @alias module:pbjs.requestBids
  */
 $$PREBID_GLOBAL$$.requestBids = (function() {
-  const delegate = hook('sync', function ({ bidsBackHandler, timeout, adUnits, adUnitCodes, labels, auctionId, ttlBuffer, ortb2, metrics } = {}) {
+  const delegate = hook('async', function ({ bidsBackHandler, timeout, adUnits, adUnitCodes, labels, auctionId, ttlBuffer, ortb2, metrics, defer } = {}) {
```
muuki88, line 626: When I discovered the issue a little more comments would have helped what the intention of this `delegate` method was or what it is supposed to do.

muuki88 (approved):
Thanks for fixing this so fast @dgirardi :hugs:

src/prebid.js
dgirardi: If you're asking about the reason for fun-hooks or the custom promise:

 - fun-hooks is basically a plugin system (to the author: sorry if I'm oversimplifying this). if we have 5 modules that may be loaded in a random combination and order, and all plug into the same code path, fun-hook solves the problem of orchestrating those. I can't answer why we are using that over something else.
 - the custom promise - `A version of Promise that runs callbacks synchronously when it can` - came about because of [this discussion](https://github.com/prebid/Prebid.js/pull/8106#pullrequestreview-1003052163).

As for this code specifically, I tried clarifying it in the comments - it's boils down to idiosyncrasies of fun-hooks.

ChrisHuie (approved):

src/prebid.js
muuki88: Thanks for the clarification :hugs:

Issue #8990, opened 2022-09-13
Proposal: upgrade fpd enrichment module to core


patmmccann:
## Type of issue
Feature request

## Description
In committee and https://github.com/prebid/Prebid.js/pull/8738 there is much discussion of the Ortb2 object, and how much of the direction of the project from a bid adapter perspective will be on improving this object and minimizing the need for duplicate code across many adapters to form an Ortb request. Some of the enrichments we plan to add are in the enrichment module, and more are planned on being added there. So that adapters can rely on these fields existing, my proposal is to upgrade enrichment to core.

bretg:
This would add about ~2KB to core, which is about a 1.3% increase.

The tradeoff is that a number of adapters are doing this enrichment independently, which means that depending on which adapters a pub includes, moving FPD enrichment to core could potentially even out the net byte count.

PR #9205, opened 2022-11-04
Prebid core: enrich FPD by default

dgirardi:
## Type of change

- [x] Feature

## Description of change

This moves `enrichmentFpdModule` into core, and extends it to also enrich `ortb2` with coppa, GDPR and USP properties.

An example of what gets automatically filled into `ortb2` at the time of calling `requestBids`:

```
{
  "regs": {
    "ext": {
      "gdpr": 1
    }
  },
  "user": {
    "ext": {
      "consent": "CO4EXq8O4EYOWADABCDEAxCsAP_AAH_AAAAAGStV_T9fb2vj-_599_t0eY1f9_63t-wzhheMs-8NyZeX_J4Wv2MyvB34JqQKGRgkunLBAQdtHGncTQgBwIlViTLMYk2MjzNKJrJEmlsbO2dYGH9Pn8XT_ZKZ70-vv__7v3___33_4GSEEmCpfAQJCWMBJNmlUKIEIVxIVAOASghGEg0sNCRQU7I4CPUACABAYAIQIAQAgohJBAAIAAElEQAgAwIBEARAIAAQAjQEIACJAEFgBIGAQACoGhYARRBKBIQYHBUcogQFSLRQTwCQgAAMAAEAABQAAYAAHAACAAAkAAKAAA.YAAAAAAAAAA"
    }
  },
  "site": {
    "page": "http://dgirardi.prebid.org:9999/integrationExamples/gitignored/gdpr_blank.html",
    "domain": "dgirardi.prebid.org:9999",
    "publisher": {
      "domain": "dgirardi.prebid.org:9999"
    }
  },
  "device": {
    "w": 663,
    "h": 898,
    "dnt": 0,
    "ua": "Mozilla/5.0 (X11; Ubuntu; Linux x86_64; rv:104.0) Gecko/20100101 Firefox/104.0",
    "language": "en"
  }
}
```

## Other information

Closes https://github.com/prebid/Prebid.js/issues/8990
Documentation TBD


Fawke:
@dgirardi Hey, all looks good from my end. I was also able to test it using the hello world page.
Can you please resolve the merge conflicts and create a docs PR?

Also, are we gonna keep the fpdModule in the modules directory for now?

mmoschovas (approved):
lgtm

patmmccann:
@dgirardi I think https://github.com/prebid/prebid.github.io/blob/master/dev-docs/bidder-adaptor.md is likely the best place for docs?

robertrmartinez:
@dgirardi 

I have tracked down some weird circumstance where this commit never continues with the auction. I am still pinpointing the exact scenario / inputs. As I only have seen it on some of our publisher setups. Others seem to work fine.

From here to the latest prebid commit it has the same effect, the auction never continues. And "fails" silently, with no log warn or error of what has happened.

Using the commit before this `182f08e314cb1d7ab9a5fc6cf86ed0179684474b` the auction runs as usual.

I will continue digging in but as you are the expert just wanted to give you a heads up.

Perhaps it is something specific to the Magnite integration but I have tracked it down to this commit.

src/prebid.js
```diff
@@ -631,7 +632,10 @@ $$PREBID_GLOBAL$$.requestBids = (function() {
       global: mergeDeep({}, config.getAnyConfig('ortb2') || {}, ortb2 || {}),
       bidder: Object.fromEntries(Object.entries(config.getBidderConfig()).map(([bidder, cfg]) => [bidder, cfg.ortb2]).filter(([_, ortb2]) => ortb2 != null))
     }
-    return startAuction({bidsBackHandler, timeout: cbTimeout, adUnits, adUnitCodes, labels, auctionId, ttlBuffer, ortb2Fragments, metrics, defer});
+    return enrichFPD(GreedyPromise.resolve(ortb2Fragments.global)).then(global => {
+      ortb2Fragments.global = global;
```
robertrmartinez, line 636: This line here is never hit

robertrmartinez: ok the diff between a page that `works` and one that does not is 

`ortb2Fragments.global`

Woks:
```
{
    "global": {
        "regs": {
            "ext": {
                "us_privacy": "1YNY"
            }
        }
    },
    "bidder": {}
}
```

Silently Fails:

```
{
    "global": {},
    "bidder": {}
}
```

So, then it seems the `enrichFpd` funciton is not resolving 

```
GreedyPromise.all([fpd, getSUA().catch(() => null)])
```

because `fpd` here is an unfulfilled promise

image

I'll continue to look into this.

Perhaps https://github.com/prebid/Prebid.js/pull/9558 fixes this, will test now.

robertrmartinez: And it does fix it

Now `fpd` at the line is
image

So somehow this change fixed our issue here.

I am not too knowledgeable on the whole GreedyPromise thing as I should be. So would love some knowledgeable input from you @dgirardi if any of this makes sense from what I put.

Obviously I can explain more if needed.

This enrich function is the part which is never fulfilling (the `fpd` var here is the one I am mentioning above)
```js
export const enrichFPD = hook('sync', (fpd) => {
  return GreedyPromise.all([fpd, getSUA().catch(() => null)])
    .then(([ortb2, sua]) => {
      Object.entries(ENRICHMENTS).forEach(([section, getEnrichments]) => {
        const data = getEnrichments();
        if (data && Object.keys(data).length > 0) {
          ortb2[section] = mergeDeep({}, data, ortb2[section]);
        }
      });
      if (sua) {
        deepSetValue(ortb2, 'device.sua', Object.assign({}, sua, ortb2.device.sua));
      }
      return ortb2;
    });
});
```

robertrmartinez: ok, turns out it is literally because certain pubs using different versions of promise I believe!

Which is mentioned in the PR https://github.com/prebid/Prebid.js/pull/9558

Ok sorry for all the messages here, was using this as my soundboard!

PR #9207, opened 2022-11-06
Pageview billable event

omerBrowsi:
## Type of change
- [ X] Feature

## Description of change
Emit pageview billable event based on event listener


modules/browsiRtdProvider.js
```diff
@@ -59,10 +59,12 @@ export function addBrowsiTag(data) {
...
-    events.emit(CONSTANTS.EVENTS.BILLABLE_EVENT, {
-      vendor: 'browsi',
-      type: 'pageview',
-      billingId: generateUUID()
+    window.addEventListener('browsi_pageview', () => {
+      events.emit(CONSTANTS.EVENTS.BILLABLE_EVENT, {
+        vendor: 'browsi',
+        type: 'pageview',
+        billingId: generateUUID()
+      })
```
robertrmartinez, line 67: so every time a call to browsi endpoint is called `getPredictionsFromServer` it will add this event listener again.

So possible this means it can add several listeners in a given page load and then emit a billing event for each time it added.

I think a flag here to ensure this event listener is only set once is the right thing to do.

robertrmartinez (changes requested):

modules/browsiRtdProvider.js
omerBrowsi: I can add a flag if you think we need extra validation. but `getPredictionsFromServer` is only called once per instance

robertrmartinez: Ok did not know that.

What if multiple prebid instances with browsi are added to a page?

Such as an iframe setup where each wrapper is in its own iframe. Do we want to have each instance add an event listener?

Probably fine I just wanted to double check.

Sounds like this will add an event listener any time a wrapper is loaded. Which probably is the right idea?

omerBrowsi: This is a very rare case (not only do we need to have a few instances on the page, we need them to use the Browsi RTD provider), but if it happens there should be a pageview event per wrapper.

robertrmartinez (approved):

PR #9226, opened 2022-11-09
Video Module: Ad Queueing

karimMourra:
## Type of change
- [x] Feature

## Description of change
For Auctions executing at page load, a race condition occurs where the auction may resolve  before the player has been instantiated. In order to ensure that the ad is not lost, we are introducing an Ad Queue. 
The Ad Queue will attempt to load the ad when the video provider is done setting up. When a ad break ends, the next ad in the queue will be loaded, until the queue is exhausted. 

## Other information

Docs: https://github.com/prebid/prebid.github.io/pull/4138

### DIFF
--- a/integrationExamples/videoModule/jwplayer/bidMarkedAsUsed.html
+++ b/integrationExamples/videoModule/jwplayer/bidMarkedAsUsed.html
@@ -76,10 +76,6 @@
           console.log('player setup failed: ', e);
         });
 
-        pbjs.onEvent('videoPlaybackRequest', (e) => {
-          pbjs.requestBids(adUnits);
-        });
-
         pbjs.onEvent('videoBidError', e => {
           console.log('An Ad Error came from a Bid: ', e);
         });
@@ -87,6 +83,8 @@
         pbjs.onEvent('videoBidImpression', e => {
           console.log('An Ad Impression came from a Bid: ', e);
         });
+
+        pbjs.requestBids(adUnits);
       });
     </script>
 </head>
--- a/modules/videoModule/index.js
+++ b/modules/videoModule/index.js
@@ -52,6 +52,7 @@ export function PbVideo(videoCore_, getConfig_, pbGlobal_, pbEvents_, videoEvent
         const divId = provider.divId;
         videoCore.registerProvider(provider);
         adQueueCoordinator.registerProvider(divId);
+        videoCore.initProvider(divId);
         videoCore.onEvents(videoEvents, (type, payload) => {
           pbEvents.emit(getExternalVideoEventName(type), payload);
         }, divId);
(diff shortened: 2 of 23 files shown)


integrationExamples/videoModule/jwplayer/bidMarkedAsUsed.html
```diff
@@ -76,17 +76,15 @@
           console.log('player setup failed: ', e);
         });
 
-        pbjs.onEvent('videoPlaybackRequest', (e) => {
```
dgirardi, line 79: why does this not work, even if the player does not exist yet? Is the problem that you may complete multiple auctions before the video player inits? If that's the case, is it really better to play old ads instead of discarding them and running a new auction for the next one?

modules/videoModule/index.js
```diff
@@ -200,6 +211,13 @@ export function PbVideo(videoCore_, getConfig_, pbGlobal_, pbEvents_, videoEvent
   // options: adXml, winner, adUnitCode,
   function loadAdTag(adTagUrl, divId, options) {
     const payload = Object.assign({ adTagUrl }, options);
+
+    if (adQueueCoordinator.requiresQueueing(divId)) {
```
dgirardi, line 215: I think always queueing would cut down code complexity significantly. By that I mean letting the queue decide if the ad needs to wait or not, instead of forking off two code paths here.
Already I can spot a difference, whether the AD_LOAD_ATTEMPT event fires is now basically random, which makes me wonder how useful it is.

integrationExamples/videoModule/jwplayer/bidMarkedAsUsed.html
karimMourra: If you run an auction in the Header, the auction is likely to complete before the player is instantiated. The video module will attempt to load the ad into the player, but since the player is not instantiated, the ad is lost. 

I think if the publisher runs an auction in the header and configures the ad unit so that the ad is rendered by the Video Module, then he explicitly wants the ad to be queued. Otherwise he can listen to Video events to coordinate his auction.

modules/videoModule/index.js
karimMourra: I agree

integrationExamples/videoModule/jwplayer/bidMarkedAsUsed.html
dgirardi: I'd think a `videoPlaybackRequest` event would have to come from the player, it's surprising to me that it could fire when the player is not ready.

jsnellbaker (approved):
From what I can understand in the code and from the conversations that already took place, I think this would be okay.

modules/videoModule/index.js
karimMourra: @dgirardi addressed

dgirardi (approved):

PR #9234, opened 2022-11-11
UserId Module : Updated correct variable name while setting cookie

pm-azhar-mulla:
## Type of change

- [x] Bugfix

## Description of change

Hi @dgirardi @patmmccann ,
During our testing, we have observed that the cookies are not getting set for TTD, on further investigation we found that in index.js file, the original `value` is set as cookie instead of the `valueStr` (which is already calculated on [line 240](https://github.com/prebid/Prebid.js/blob/master/modules/userId/index.js#L240) but not used in `IF` statement)
We have made the related changes, please have a look at it once.


dgirardi (approved):
thanks for catching this!

jlquaccia (approved):
LGTM

PR #9236, opened 2022-11-11
Permutive RTD Module: add support for new ssp standard cohorts

zshnr:
## Type of change

- [ ] Bugfix
- [x] Feature
- [ ] New bidder adapter   
- [ ] Build related changes
- [ ] CI related changes

- [ ] Other

## Description of change
1. Adds a new kind of standard cohort to pass on to different vendors
2. Adds logic to set these cohorts as auction level keywords in Xandr for customers that use Xandr

## Other information
N/A


patmmccann:
Sorry for status change, was kicking off tests

zshnr:
@patmmccann no worries :) just waiting on some requirements to be clarified and then there might be a couple more commits to finish this off.

zshnr:
@patmmccann ready for review :) thanks!

modules/permutiveRtdProvider.js
```diff
@@ -341,6 +373,21 @@ function iabSegmentId(permutiveSegmentId, iabIds) {
   return iabIds[permutiveSegmentId] || unknownIabSegmentId
 }
 
+/**
+ * Passes Permutive's standard cohorts to Appnexus/Xandr via auction level keywords
+ * @param {Object} sspData -- {ssp?: string[], cohorts?: string[]}
+ */
+function passCohortsToAppnexusAuctionKeywords(sspData) {
```
patmmccann, line 380: This is a rules violation, you must write to the standard object and appnexus adapter should be responsible for the mutation

patmmccann (changes requested):
Rules feedback

patmmccann (approved):

modules/permutiveRtdProvider.js
```diff
@@ -222,10 +242,19 @@ function getCustomBidderFn (moduleConfig, bidder) {
...
+    const ssp = data?.ssp?.cohorts ?? []
+    return [...new Set([...ac, ...ssp])]
+  }
   const bidderMap = {
     appnexus: function (bid, data, acEnabled) {
-      if (acEnabled && data.ac && data.ac.length) {
-        deepSetValue(bid, 'params.keywords.p_standard', data.ac)
+      if (isPStandardTargetingEnabled(data, acEnabled)) {
+        const segments = pStandardTargeting(data, acEnabled)
+        deepSetValue(bid, 'params.keywords.p_standard', segments)
```
patmmccann, line 257: you guys should consider yourself on notice to remove these special handlings, move them into their respective bid adapters, expect someone else moves them for you or simply deletes them, or expect not to be in the next version of prebid

CremboC: Hi Patrick, thanks for the feedback. I'm already working on removing the appnexus special handling, I just wanted to do it as a separate PR in order to simplify reviewing.

For the other SSPs for which we have special handling - we're looking into them. I don't have the full context myself, but I understand it's not as simple as migrating to the `ortb2.user.data` field. I'll catch up internally with my team to ensure we don't break any existing targeting, as this might impact our customer's revenue!

Regardless, thanks for the heads-up that this method will stop working soon - we'll prioritise updating this to avoid any issues.

modules/permutiveRtdProvider.js
```diff
@@ -222,10 +242,19 @@ function getCustomBidderFn (moduleConfig, bidder) {
  * @return {Object} Bidder function
  */
 function getDefaultBidderFn (bidder) {
+  const isPStandardTargetingEnabled = (data, acEnabled) => {
+    return (acEnabled && data.ac && data.ac.length) || (data.ssp && data.ssp.cohorts.length)
```
robertrmartinez, line 246: @zshnr Hello!

I am seeing `data.ssp.cohorts` come back as `undefined` in some cases for our publishers prebid wrappers.

Here is an example:

image


This causes an exception to be thrown and the permutive `makeSafe` function catches it but probably does not do all of the necessary stuff!

So probably just need to add a safe check here for making sure `cohorts` is defined before accessing `length`

robertrmartinez: Seems like it is okay on the publisher site when the Perutive endpoint comes back with an actual data resonse:

image

So probably not a huge deal, but just in case might be a good idea to add a safe check!

zshnr: Ah that is bizzare 🤔 . I was sure if the `ssp` property exists then it will have a `cohorts` property even if its empty.

Thanks for flagging I'll check internally. @robertrmartinez

zshnr: Hi @robertrmartinez , I have merged a fix for this :) 

To help us understand the impact of this, can you let me know which publisher websites you saw this on?

Thanks!

PR #9243, opened 2022-11-15
Criteo Bid Adapter : fix getFloor usage issue

leonardlabat:
## Type of change
- [X] Bugfix
- 
## Description of change
getFloor call context is loss when we reference it as a callback ; we need to make sure that it is properly mapped as getFloor function assumes that "this" is the actual bid request object


mmoschovas (approved):

patmmccann:
@leonardlabat fwiw we tested versions 7.23 through 7.28 just now and we saw an enormous (circa 60%) criteo revenue gain in versions 7.23 through 7.26 which was wiped back to baseline in 7.27

Issue #9250, opened 2022-11-16
ortb2 FPD object contains a ortb 2.6 field


jbartek25:
## Type of issue
Bug

## Description
enrichmentFpdModule with the help of sua.js adds ortb 2.6 field `device.sue` into ortb2 object.
Improve Digital bidder adapter was recently migrated to benefit from the common `ortbConverter` logic which fills in all the usual fields in a oRTB bid request. Now, Improve Digital ad server endpoint is oRTB 2.5 and rejects any request containing `device.sue`.
I believe most (all) SSPs with oRTB endpoints are on oRTB 2.5 and the `ortbConverter` should only set valid 2.5 fields. This is a big issue for us. I would think our adapter is not the right place for the fix and it should be either in `ortbConverter` or elsewhere in the core/common logic. What do you think? @dgirardi

<img width="919" alt="Screenshot 2022-11-16 at 16 14 27" src="https://user-images.githubusercontent.com/31618107/202219353-bd62b5d4-9431-4244-85bb-07b746d4db6d.png">

dgirardi:
The [2.6 spec](https://iabtechlab.com/wp-content/uploads/2022/04/OpenRTB-2-6_FINAL.pdf) states:

>  implementers need not specifically treat the minor version of OpenRTB 2.x as significant, in most circumstances. For example:
> Bidders and exchanges must tolerate receiving new or unexpected fields and enumerated list
values gracefully, treating them as unknown or ignoring them

However, that section is not in the [2.5](https://www.iab.com/wp-content/uploads/2016/03/OpenRTB-API-Specification-Version-2-5-FINAL.pdf) spec, so it's not clear if that is recommended only for 2.6 onwards (the wording seems to be general to 2.x).

I am not sure of what `ortbConverter` should do, it's not trivial to enforce 2.5 - because anything provided directly by the pub is transferred almost verbatim into the request. 

As an immediate fix I believe we should start with removing `sua` from the Improve Digital adapter. Long term I think the best option would be for your endpoint to tolerate extra fields, as we plan to adopt more 2.6 fields over time; however, we can think about building a version toggle into `ortb2Converter` itself.

dgirardi:
Both of the other adapters that are currently behind the common logic - Prebid Server and OpenX - do not mind getting extra fields. It's hard to gauge now if this problem will be more common as more adapters convert.

[...]

PR #9251, opened 2022-11-16
Improve digital bid adapter: do not set `device.sua` in request

dgirardi:
## Type of change

- [x] Bugfix

## Description of change

Temporary fix for #9250 - ID endpoint does not accept 2.6 fields such as `device.sua`

@jbartek25

### DIFF
--- a/modules/improvedigitalBidAdapter.js
+++ b/modules/improvedigitalBidAdapter.js
@@ -162,6 +162,8 @@ export const CONVERTER = ortbConverter({
         }
       },
     });
+    // Remove 2.6 device.sua; see #9250
+    delete request.device?.sua;
     return request;
   },
   bidResponse(buildBidResponse, bid, context) {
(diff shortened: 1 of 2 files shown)


modules/improvedigitalBidAdapter.js
```diff
@@ -162,6 +162,8 @@ export const CONVERTER = ortbConverter({
         }
       },
     });
+    // Remove 2.6 device.sua; see #9250
+    delete request.device?.sua;
```
patmmccann, line 166: You don't prefer it in device.ext.sua?

jbartek25:
@dgirardi please don't make any change yet. https://github.com/prebid/Prebid.js/issues/9250#issuecomment-1317687029

jbartek25:
This PR can be closed. Temporary fix will be done in our ad server. thanks @dgirardi for your swift reaction

PR #9252, opened 2022-11-16
PBjs Core : send native targetings for ortb response

### LINKED ISSUE #9246: Native targeting keys are not set for bids that use ORTB native responses
## Type of issue
I think this is a bug (but I may have missed something on native changes on Prebid)

## Description
We tried to implement IX on native, but it seems that since #8853, Index Exchange adapter returns a naive bid response with openrtb format instead of Prebid native.

### Actual results
Index exchange native bid response : 
image

### Expected results
Regular native bid response : 
image

Is this an issue with IX bid adapter or should we adapt to handle this kind of native response ?

JulieLorin:
## Type of change
- Bugfix

## Description of change
Fixes #9246

Legacy prebid native fields are now added to the native object of the prebid bid response directly. 
They are now available to bid response object : 
image
They are available in adserver targetings : 
image
And they are available through the post message call to retrieve assets : 
image

### DIFF
--- a/src/auction.js
+++ b/src/auction.js
@@ -76,7 +76,7 @@ import {
   timestamp
 } from './utils.js';
 import {getPriceBucketString} from './cpmBucketManager.js';
-import {getNativeTargeting} from './native.js';
+import {getNativeTargeting, toLegacyResponse} from './native.js';
 import {getCacheUrl, store} from './videoCache.js';
 import {Renderer} from './Renderer.js';
 import {config} from './config.js';
@@ -84,7 +84,7 @@ import {userSync} from './userSync.js';
 import {hook} from './hook.js';
 import {find, includes} from './polyfill.js';
 import {OUTSTREAM} from './video.js';
-import {VIDEO} from './mediaTypes.js';
+import {VIDEO, NATIVE} from './mediaTypes.js';
 import {auctionManager} from './auctionManager.js';
 import {bidderSettings} from './bidderSettings.js';
 import * as events from './events.js';
@@ -462,9 +462,12 @@ export function auctionCallbacks(auctionDone, auctionInstance, {index = auctionM
     handleBidResponse(adUnitCode, bid, (done) => {
       let bidResponse = getPreparedBidForAuction(bid);
 
-      if (bidResponse.mediaType === 'video') {
+      if (bidResponse.mediaType === VIDEO) {
         tryAddVideoBid(auctionInstance, bidResponse, done);
       } else {
+        if (FEATURES.NATIVE && bidResponse.mediaType === NATIVE) {
+          addLegacyFieldsIfNeeded(bidResponse);
+        }
         addBidToAuction(auctionInstance, bidResponse);
         done();
       }
@@ -590,6 +593,17 @@ function tryAddVideoBid(auctionInstance, bidResponse, afterBidAdded, {index = au
   if (addBid) {
     addBidToAuction(auctionInstance, bidResponse);
     afterBidAdded();
+  }
+}
+
+// Native bid response might be in ortb2 format - adds legacy field for backward compatibility
+function addLegacyFieldsIfNeeded(bidResponse) {
+  const nativeOrtbRequest = auctionManager.index.getAdUnit(bidResponse)?.nativeOrtbRequest;
+  const nativeOrtbResponse = bidResponse.native?.ortb
+
+  if (nativeOrtbRequest && nativeOrtbResponse) {
+    const legacyResponse = toLegacyResponse(nativeOrtbResponse, nativeOrtbRequest);
+    Object.assign(bidResponse.native, legacyResponse);
   }
 }
 
--- a/src/auction.js
+++ b/src/auction.js
@@ -84,7 +84,7 @@ import {userSync} from './userSync.js';
 import {hook} from './hook.js';
 import {find, includes} from './polyfill.js';
 import {OUTSTREAM} from './video.js';
-import {VIDEO, NATIVE} from './mediaTypes.js';
+import {VIDEO} from './mediaTypes.js';
 import {auctionManager} from './auctionManager.js';
 import {bidderSettings} from './bidderSettings.js';
 import * as events from './events.js';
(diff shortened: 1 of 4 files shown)


src/auction.js
```diff
@@ -593,6 +596,17 @@ function tryAddVideoBid(auctionInstance, bidResponse, afterBidAdded, {index = au
   }
 }
 
+// Native bid response might be in ortb2 format - adds legacy field for backward compatibility
+function addLegacyFieldsIfNeeded(bidResponse) {
+  const nativeOrtbRequest = auctionManager.index.getBidRequest(bidResponse)?.nativeOrtbRequest;
```
dgirardi, line 601: this should be `index.getAdUnit` - in some situations you can have a bid response without a bid request.

src/auction.js
```diff
@@ -462,9 +462,12 @@ export function auctionCallbacks(auctionDone, auctionInstance, {index = auctionM
     handleBidResponse(adUnitCode, bid, (done) => {
       let bidResponse = getPreparedBidForAuction(bid);
 
-      if (bidResponse.mediaType === 'video') {
+      if (bidResponse.mediaType === VIDEO) {
         tryAddVideoBid(auctionInstance, bidResponse, done);
       } else {
+        if (bidResponse.mediaType === NATIVE) {
```
dgirardi, line 468: `if(FEATURES.NATIVE && bidResponse.mediaType === NATIVE)`. Without the additional check the build would pull native-specific code even when the user asks to remove it.

test/spec/native_spec.js
```diff
@@ -97,8 +97,28 @@ const ortbBid = {
       },
       privacy: 'https://privacy-link.example',
       ver: '1.2'
-    }
-  },
+    },
+    // Native bid response always still have prebid legacy fields
+    title: 'Native Creative',
```
dgirardi, line 102: with this change the tests using `ortbBid` are no longer real tests - they are satisfied from their precondition. 

I think the correct thing to do is to refactor them to test `toLegacyResponse` instead of the `getAssets` methods, now that the former is exposed. The logic should be very similar (naked ORTB response comes in, legacy response comes out).

dgirardi (commented):
Thanks for taking this on!

JulieLorin:
> Thanks for taking this on!

I treated your feedback, added trackers assets to `toLegacyResponse`, and tried to make the tests a bit more "logical", let me know if you need any more updates

src/native.js
```diff
@@ -764,6 +752,29 @@ function toLegacyResponse(ortbResponse, ortbRequest) {
       legacyResponse[PREBID_NATIVE_DATA_KEYS_TO_ORTB_INVERSE[NATIVE_ASSET_TYPES_INVERSE[requestAsset.data.type]]] = asset.data.value;
     }
   }
+
+  // Handle trackers
+  legacyResponse.impressionTrackers = [];
+  let jsTrackers = [];
+
+  if (ortbRequest?.imptrackers) {
+    legacyResponse.impressionTrackers.push(ortbRequest.imptrackers);
```
dgirardi, line 761: I think this should be `push(...ortbRequest.imptrackers)`. `imptrackers` [should be an array of URLs](https://www.iab.com/wp-content/uploads/2018/03/OpenRTB-Native-Ads-Specification-Final-1.2.pdf).

test/spec/native_spec.js
```diff
@@ -548,67 +573,54 @@ describe('native.js', function () {
...
       }
-    }
+    ],
+    eventtrackers: [
+      { event: 1, method: 1, url: 'https://sampleurl.com' },
+      { event: 1, method: 2, url: 'https://sampleurljs.com' }
+    ]
   }
   describe('getAllAssetsMessage', () => {
     it('returns assets in legacy format for ortb responses', () => {
```
dgirardi, line 614: the test names passed to `describe` /  `it` are no longer accurate.

dgirardi (commented):
Still a couple minor comments, but otherwise LGTM - I will also take some time to test this against PUC just to be sure.

JulieLorin:
> Still a couple minor comments, but otherwise LGTM - I will also take some time to test this against PUC just to be sure.

Thanks, this should be fixed. 
I didn't fully test with PUC : I just manually called the postMessage and logged the returned value. Thanks for taking the time to do so

dgirardi (approved):
LGTM. I added a small change to run `toLegacyResponse` regardless of whether the response mediatype is native - I'm not sure it's necessary, but it brings behavior in line with what's currently in master.

src/auction.js
```diff
@@ -462,9 +462,14 @@ export function auctionCallbacks(auctionDone, auctionInstance, {index = auctionM
...
       let bidResponse = getPreparedBidForAuction(bid);
 
-      if (bidResponse.mediaType === 'video') {
+      if (bidResponse.mediaType === VIDEO) {
         tryAddVideoBid(auctionInstance, bidResponse, done);
       } else {
+        if (FEATURES.NATIVE && bidResponse.native != null && typeof bidResponse.native === 'object') {
+          // NOTE: augment bidResponse.native even if bidResponse.mediaType !== NATIVE; it's possible
+          // to treat banner responses as native
+          addLegacyFieldsIfNeeded(bidResponse);
```
musikele, line 471: It might be a good thing to add a test for this branch (and test `addLegacyFieldsIfNeeded`)

musikele (changes requested):
Can you add one more test ? So we prevent problems for the future.

src/auction.js
JulieLorin: I tried to add a test for that. (I am not very experienced with prebid development, so if it's not the right way, I think I'll need some pointers on how to correctly test that)

musikele (changes requested):
@JulieLorin  There's just a linting issue and we're good to go

test/spec/auctionmanager_spec.js
```diff
@@ -1276,6 +1277,79 @@ describe('auctionmanager.js', function () {
...
+            { id: 2, title: { text: 'Sample title' } },
+            { id: 4, data: { value: 'Sample body' } },
+            { id: 3, data: { value: 'Sample sponsoredBy' } },
+            { id: 1, img: { url: 'https://www.example.com/image.png', w: 200, h: 200 } },
+            { id: 5, img: { url: 'https://www.example.com/icon.png', w: 32, h: 32 } }
+          ],
+          link: { url: 'http://www.click.com' },
+          eventtrackers: [
+            { event: 1, method: 1, url: 'http://www.imptracker.com'},
+            { event: 1, method: 2, url: 'http://www.jstracker.com/file.js'}
```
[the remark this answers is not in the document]
JulieLorin: should be all good

PR #9255, opened 2022-11-17
GravitoId System: fix issue with gravitompId not appearing in bid request

rahulgravito:
## Type of change

- [ ] Bugfix

## Description of change
Fixed the issue with Gravito User Module where Gravito id was not appearing in bid request.


ChrisHuie:
@rahulgravito can you fix the linting error here 🙏

Issue #7329, opened 2021-08-20
PBS Bid Adapter: remove site object when app is present


bretg:
## Type of issue
enhancement

## Description
Prebid Server chokes when an OpenRTB request contains both site and app sections. Since OpenRTB says this is an invalid scenario, that committee has determined in https://github.com/prebid/prebid-server/issues/1947 that the right solution is to prevent it from happening.

## Steps to reproduce
The instructions at https://docs.prebid.org/faq/prebid-server-faq.html#how-can-i-use-prebid-server-in-a-mobile-app-post-bid-scenario define how to use PBJS for App in a Post-Bid scenario.

But it doesn't actually work because the PBS Bid Adapter also includes site.

So we need to update the PBS Bid adapter to:

- First do all the merging: getConfig({app}), getConfig({site}), getConfig({ortb2})
- Then check the result. If both app and site are present, then remove site with a console warning
- it's up to the page to make sure any common fields (e.g. `cat`) are in the app object

Note: we don't consider this a breaking change because PBS is already failing on this if it's happening. We're just moving the validation to the client side to save everyone the headache.

pm-harshad-mane:
https://github.com/prebid/Prebid.js/blob/master/modules/prebidServerBidAdapter/index.js#L330
Is it already handled?

patmmccann:
might make sense for the validations module to also note this issue

ChrisHuie:
Working on tests and then submitting this pr.

patmmccann:
fix in #9258

PR #9258, opened 2022-11-17
PBS Bid Adapter : site should not exist when app is present

ChrisHuie:
#7329


test/spec/modules/prebidServerBidAdapter_spec.js
```diff
@@ -1477,6 +1477,7 @@ describe('S2S Adapter', function () {
       adapter.callBids(REQUEST, BID_REQUESTS, addBidResponse, done, ajax);
       const requestBid = JSON.parse(server.requests[0].requestBody);
       expect(requestBid.site).to.exist.and.to.be.a('object');
+      expect(requestBid.app).to.not.exist;
```
dgirardi, line 1480: I don't think this is enough to test #7329. The test should say "if both app and site are set, only app should be set in the request". The setup of this test only sets up `site` so it's not covering the conflict.

I was not aware that `app` and `site` are mutually exclusive in ORTB; I think this requires changes in the `ortbConverter` logic.

ChrisHuie: updating 👍

patmmccann:
This requires changes to choice, not just tests; we want to make it impossible for the pbs adapter to send more than one of dooh, site, app

test/spec/modules/prebidServerBidAdapter_spec.js
patmmccann: I think so as well @dgirardi ; also 'dooh' is exclusive to these

ChrisHuie:
> This requires changes to code, not just tests; we want to make it impossible for the pbs adapter to send more than one of dooh, site, app

Yes. Prior to ORTB-PBjs conversion library release there was a specific appendSiteApp function in prebidServer. Some of that

patmmccann (approved):

libraries/ortbConverter/processors/default.js
```diff
@@ -165,3 +169,17 @@ export function setSite(ortbRequest, bidderRequest) {
     ortbRequest.site = Object.assign(getDefinedParams(bidderRequest.refererInfo, ['page', 'domain', 'ref']), ortbRequest.site);
   }
 }
+
+export function onlyOneClientSection(ortbRequest) {
+  ['dooh', 'app', 'site'].reduce((found, section) => {
```
jsnellbaker, line 174: If this overall logic will simply drop the latter found objects in the PBS request, is there a desired preference for which objects should be included over the other?  

With this current logic, the order in the `['dooh', 'app', 'site']` array will prefer whatever is listed first and will most likely drop whatever is nearer the end.

I'm not sure if this semi-bias would pose to be an issue in reality, but I thought I would ask just in case.

dgirardi: That order is intentional. I know @patmmccann agreed on it but I'm not sure it's been written down anywhere; the idea is that if the pub sets `dooh`, they likely _really_ want that to be sent over, over things that are automatically generated like `site`.

jsnellbaker: Ok thanks for clarifying @dgirardi 

If the conflicts are resolved, I think this would look good.

jsnellbaker (approved):

libraries/ortbConverter/processors/default.js
patmmccann: we asked the dooh committee for input on this; this was their preference

patmmccann: I added a docs pull here https://github.com/prebid/prebid.github.io/pull/4307/files

PR #9260, opened 2022-11-18
Redtram Bid Adapter : initial adapter release

redtram-tech:
## Type of change

- [x] New bidder adapter

### DIFF
--- a/modules/redtramBidAdapter.js
+++ b/modules/redtramBidAdapter.js
@@ -1,0 +1,139 @@
+import {
+  isFn,
+  isStr,
+  deepAccess,
+  getWindowTop,
+  triggerPixel,
+  logInfo
+} from '../src/utils.js';
+import { registerBidder } from '../src/adapters/bidderFactory.js';
+import { BANNER } from '../src/mediaTypes.js';
+import { convertOrtbRequestToProprietaryNative } from '../src/native.js';
+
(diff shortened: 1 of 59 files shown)


modules/redtramBidAdapter.js
```diff
@@ -0,0 +1,158 @@
...
+export const spec = {
+  code: BIDDER_CODE,
+  supportedMediaTypes: [BANNER],
+
+  isBidRequestValid: (bid) => {
+    return Boolean(bid.bidId && bid.params && bid.params.placementId);
+  },
+
+  buildRequests: (validBidRequests = [], bidderRequest) => {
+    validBidRequests = convertOrtbRequestToProprietaryNative(validBidRequests);
```
patmmccann, line 57: I'm a little nervous about building to a legacy format from the jump. What's the goal here?

@dgirardi can you remind me if we plan long term support for this function. I seem to recall deprecation discussions

dgirardi: In this case it looks unnecessary, since the only supported media type is banner. 
The function call is useful only if you want to get native adUnits in the "legacy" prebid format.

redtram-tech: Ok, I understand you. Removed this function from the code.

jlquaccia (approved):
seems to lgtm

PR #9263, opened 2022-11-18
Prebid Core: ORTB 2.5 translation utilities

dgirardi:
## Type of change

- [x] Feature

## Description of change

This adds two library utilities for translating ORTB 2.6+ requests into 2.5:
 
- a "soft" translation (`toOrtb25`) that moves 2.x fields that have a standard extension in 2.5 (such as `regs.gdpr` -> `regs.ext.gdpr`, `user.consent` -> `user.ext.consent`, etc; I used https://github.com/prebid/prebid-server/issues/2139 as reference)
- a "strict" translation (`toOrtb25Strict`) that does the above, plus removes any fields that are not defined in 2.5, or fields that are defined as enums but have a value that is not in the 2.5 range.

## Other information

Closes https://github.com/prebid/Prebid.js/issues/9250


patmmccann:
Lgtm; pending changes to regs object

patmmccann (approved):

PR #9268, opened 2022-11-22
Prebid Core: Addition of Optional Category Targeting Key

jlquaccia:
## Type of change
- [x] Feature

## Description of change
Optional category targeting key can be sent to ad servers that stores the value of the Primary IAB category ID if present on the meta obj of a bid response.

## Other information
Hi @patmmccann, I was looking into this GitHub issue and had one question. I see in the table describing the meta obj [here](https://docs.prebid.org/dev-docs/bidder-adaptor.html#interpreting-the-response) that for category there are fields for primary and secondary cat id's.  For this new targeting key, should the value be set to the primary only? (what i did for this pr).  Or I can adjust the logic as needed.  I can also update things on the testing end after I find out.

### DIFF
--- a/integrationExamples/gpt/prebidServer_example.html
+++ b/integrationExamples/gpt/prebidServer_example.html
@@ -40,10 +40,17 @@
               } 
             },
             bids: [
+              // {
+              //   bidder: 'appnexus',
+              //   params: {
+              //      placementId: 13144370
+              //   }
+              // }
               {
-                bidder: 'appnexus',
+                bidder: 'pubmatic',
                 params: {
-                   placementId: 13144370
+                  publisherId: '156276',
+                  adSlot: 'pubmatic_test'
                 }
               }
             ]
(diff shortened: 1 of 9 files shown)


patmmccann (approved):

patmmccann:
This looks right to me! Might be useful to check prebid server adapter populating this correctly. 

Will need second review

jlquaccia:
@patmmccann Ok awesome, sounds good!  Will look into PBS and then address the docs & unit tests next.

patmmccann:
might be able to solve the prebid server check with a unit test

robertrmartinez:
@jlquaccia @patmmccann So is this now expected to send by default always?

I thought this would be an opt-in thing?

patmmccann:
Opt in, just like adomain

jlquaccia:
hey @patmmccann @robertrmartinez, can you guys review when you get a chance?  Did the following:
- updated relative targeting related tests
- updated docs -> https://github.com/prebid/prebid.github.io/pull/4192
- confirmed that this new category targeting key also gets set with pbs (when the category targeting key is enabled)

robertrmartinez (approved):

patmmccann:
thanks!

Issue #9275, opened 2022-11-22
Bid Adapters not using the floors convention


bretg:
## Type of issue
enforcing a convention

## Description
The following bid adapters are accepting a floor as a bidder-specific parameter without calling the getFloors() function:

- [x] alkimi @kalidas-alkimi 
- [x] cleanmedianet @sa1omon 
- [x] datawrkz @vishal-dw 
- [x] ebdr @jlzhangdev 
- [x] impactify @thomasdseao 
- [x] iqm @Params10 
- [x] lkqd @david-carver 
- [x] madvertise @NLopezMad 
- [x] mytarget @Denislavrov 
- [x] nexx360 @gchicoye 
- [x] quantcast @SleimanJneidi 
- [x] taboola @ahmadlob 
- [x] truereach @mmprebid 
- [x] vdoai @cpuBird 
- [x] vidoomy @sasanfarokh 

Please look in other bid adapters for how the `bidrequest.getFloor()` function is utilized.

gchicoye:
Hi @bretg , I've submitted 2 weeks ago a pull request (https://github.com/prebid/Prebid.js/pull/9229) which fixes that, I'm waiting for the approval!

patmmccann:
tagging 9.0 to boot non-compliant adapters

kalidas-alkimi:
@bretg    Alkimi Bid Adapter : We have fixed this issue long before and the changes merged. Please check the below pull request. Please let me know if we still missing anything.

https://github.com/prebid/Prebid.js/pull/9368

patmmccann:
vidoomy and cleanmedia are fine

PR #9279, opened 2022-11-23
Taboola Bid Adapter: Fixing Accepting Bid Floor Mechanism  

ahmadlob:
## Type of change

- [x] Bugfix
- [ ] Feature
- [ ] New bidder adapter   
- [ ] Build related changes
- [ ] CI related changes

- [ ] Other

## Description of change
accepting the bid floor by calling the getFloors() function instead of bidder-specific parameter.
related to this issue: https://github.com/prebid/Prebid.js/issues/9275

## Other information
@bretg

### DIFF
--- a/modules/taboolaBidAdapter.js
+++ b/modules/taboolaBidAdapter.js
@@ -186,6 +186,10 @@ function getImps(validBidRequests) {
         imp.bidfloor = parseFloat(floorInfo.floor);
         imp.bidfloorcur = CURRENCY;
       }
+    } else {
+      const {bidfloor = null, bidfloorcur = CURRENCY} = bid.params;
+      imp.bidfloor = bidfloor;
+      imp.bidfloorcur = bidfloorcur;
     }
     return imp;
   });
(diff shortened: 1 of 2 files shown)


modules/taboolaBidAdapter.js
```diff
@@ -170,15 +170,24 @@ function getSiteProperties({publisherId, bcat = []}, refererInfo) {
...
+        currency: CURRENCY,
+        mediaType: BANNER,
+        size: '*'
+      });
+      if (typeof floorInfo === 'object' && floorInfo.currency === CURRENCY && !isNaN(parseFloat(floorInfo.floor))) {
+        imp.bidfloor = parseFloat(floorInfo.floor);
+        imp.bidfloorcur = CURRENCY;
+      }
+    }
+    return imp;
```
robertrmartinez, line 190: One thing some adapters have done is fallback to use their bidder specific floor if the getFloor function is not present.

Totally up to you if you want to keep that logic or not.

So basically

```js
if (typeof bid.getFloor === 'function') {
   // get floor from getFloor function
} else {
   // get floor from your bidder params if there
}
```

ahmadlob: @robertrmartinez 

Thanks for the comment, I added the fallback to bid params as you described - thanks!

robertrmartinez: Sorry for the delay here!

Can you add a unit test or two for this change as well please.

robertrmartinez (approved):
nice thanks!

PR #9286, opened 2022-11-28
AdHash bid adapter: update to support latest version

wyand-sp:
Updating the Prebid.JS adaptor to support our platform version 3.2 and above.

## Type of change

- [ ] Bugfix
- [x] Feature
- [ ] New bidder adapter   
- [ ] Build related changes
- [ ] CI related changes

- [ ] Other

## Description of change
- Better support for brand-safety including the URL and the new formula
- Support for multiple top-level bidder URLs
- Use of local storage for recently viewed ads
- Support for ad-density


modules/adhashBidAdapter.js
```diff
@@ -98,17 +148,50 @@ export const spec = {
...
     const bidRequests = [];
     let referrer = '';
-    if (bidderRequest && bidderRequest.refererInfo) {
-      // TODO: is 'page' the right value here?
-      referrer = bidderRequest.refererInfo.page;
+    try {
+      referrer = window.top.location.href;
+    } catch (e) {
+      referrer = window.location.href;
     }
```
Fawke, line 159: You can read the referer value from bidderRequest object, `bidderRequest.refererInfo`

modules/adhashBidAdapter.js
```diff
@@ -98,17 +148,50 @@ export const spec = {
...
+      body.scrollHeight,
+      body.offsetHeight,
+      html.clientHeight,
+      html.scrollHeight,
+      html.offsetHeight
+    );
+    const pageWidth = Math.max(body.scrollWidth, body.offsetWidth, html.clientWidth, html.scrollWidth, html.offsetWidth);
+
+    for (let i = 0; i < validBidRequests.length; i++) {
+      const bidderURL = validBidRequests[i].params.bidderURL || 'https://bidder.adhash.com';
```
Fawke, line 172: Can you please ensure that the url passed through `params.bidderURL` is using `https://` and not `http://`.

Fawke (commented):
Hi @wyand-sp,

I just added a few small comments, see if it's possible to work on those. Apart from that, it looks good.

modules/adhashBidAdapter.js
wyand-sp: I've added a check in isBidRequestValid as well as unit test for that.

modules/adhashBidAdapter.js
wyand-sp: I am now using bidderRequest.refererInfo.

Fawke (approved):

PR #9289, opened 2022-11-28
fix for broken download bundle when KueezRTB is included

elad-yosifon:
fix for broken download bundle https://github.com/prebid/prebid.github.io/issues/4177


patmmccann (approved):

bretg:
Rather than create a breaking change for publishers, we added a new feature to the download page that allows the bidders filename to differ from the biddercode. https://github.com/prebid/prebid.github.io/pull/4186

PR #9295, opened 2022-11-29
[WIP] Scattered Bid Adapter: New Adapter

almendar:
## Type of change

- [ ] Bugfix
- [ ] Feature
- [x] New bidder adapter   

https://github.com/prebid/prebid.github.io/pull/4181

- [ ] Build related changes
- [ ] CI related changes

- [ ] Other

## Description of change
New adapter for Scattered DSP

<!-- For new bidder adapters, please provide the following
prebid@scattered.pl

```js
  {
                    bidder: "scattered",
                    params: {
                        bidderDomain: "prebid-test.scattered.eu/bid",
                        test: 0
                    }
}
```

### DIFF
--- a/modules/scatteredBidAdapter.js
+++ b/modules/scatteredBidAdapter.js
@@ -35,7 +35,6 @@ export function buildImps(validBidRequests) {
 
 export const spec = {
   code: BIDDER_CODE,
-  gvlid: -1,
   supportedMediaTypes: [BANNER],
 
   // 1.
--- a/modules/scatteredBidAdapter.js
+++ b/modules/scatteredBidAdapter.js
@@ -1,0 +1,176 @@
+// jshint esversion: 6, es3: false, node: true
+// import * as utils from '../src/utils';
+
+import { registerBidder } from '../src/adapters/bidderFactory.js';
+import { config } from '../src/config.js';
+import { BANNER } from '../src/mediaTypes.js';
+import { mergeDeep, deepAccess, parseSizesInput, parseUrl, deepSetValue, logInfo, createTrackPixelHtml } from '../src/utils.js';
+
+const { getConfig } = config;
+
+const BIDDER_CODE = 'scattered';
+
+export function buildImps(validBidRequests) {
+  return validBidRequests.map((bid, index) => {
+    let imp = {
+      id: `${index + 1}`
+    };
+    const bannerParams = deepAccess(bid, 'mediaTypes.banner');
+    if (bannerParams && bannerParams.sizes) {
+      const sizes = parseSizesInput(bannerParams.sizes);
+      const format = sizes.map(size => {
+        const [width, height] = size.split('x');
+        const w = parseInt(width, 10);
+        const h = parseInt(height, 10);
+        return { w, h };
+      });
+
+      imp.banner = {
+        format
+      };
+    };
+    return imp;
+  });
+}
+
+export const spec = {
+  code: BIDDER_CODE,
+  gvlid: -1,
+  supportedMediaTypes: [BANNER],
+
+  // 1.
+  isBidRequestValid: function (bid) {
+    const bidderDomain = deepAccess(bid, 'params.bidderDomain')
+    if (bidderDomain === undefined || bidderDomain === '') {
+      return false
+    }
+
+    const sizes = deepAccess(bid, 'mediaTypes.banner.sizes')
+    if (sizes === undefined || sizes.length < 1) {
+      return false
+    }
+
+    return true
+  },
+
+  // 2.
+  buildRequests: function (validBidRequests, bidderRequest) {
+    const ortb2Obj = getConfig('ortb2') || {};
+    let { user } = ortb2Obj
+
+    let site = getConfig('site') || {};
+    if (ortb2Obj.site) {
+      mergeDeep(site, ortb2Obj.site)
+    }
+    // site.page    URL of the page where the impression will be shown.
+    // site.domain  Domain of the site (e.g., “mysite.foo.com”).
+    if (!site.page && bidderRequest.refererInfo) {
+      site.page = bidderRequest.refererInfo.referer;
+      site.domain = parseUrl(bidderRequest.refererInfo.referer).hostname
+    }
+
+    const bidderDomain = getKeyOnAny(validBidRequests, 'params.bidderDomain')
+
+    const device = getConfig('device') || {};
+    device.w = device.w || window.innerWidth;
+    device.h = device.h || window.innerHeight;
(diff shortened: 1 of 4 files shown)


almendar:
Hi!
@dgirardi is there anything I can do to make this PR rolling? 😃

modules/scatteredBidAdapter.js
```diff
@@ -0,0 +1,175 @@
...
+
+    return true
+  },
+
+  // 2.
+  buildRequests: function (validBidRequests, bidderRequest) {
+    const ortb2Obj = bidderRequest.ortb2 || {};
+    let { user } = ortb2Obj
+
+    let site = getConfig('site') || {};
```
dgirardi, line 60: For a new adapter I believe it's better to not do this, since this way of setting FPD is deprecated - and incomplete (the same should be done for `app` and `device`). We have [this issue](https://github.com/prebid/Prebid.js/issues/9335) open to solve this problem in core, after which adapters would see everything directly under `bidderRequest.ortb2`.

modules/scatteredBidAdapter.js
```diff
@@ -0,0 +1,175 @@
...
+      for (let seat of seatbid) {
+        for (let bidResponse of seat.bid) {
+          const bidRequest = bidRequests[idx]
+
+          const result = {
+            requestId: bidRequest.bidId,
+            cpm: bidResponse.price,
+            creativeId: bidResponse.crid,
+            ttl: 360,
+            netRevenue: bidRequest.netRevenue === 'net',
```
dgirardi, line 134: I think `result.netRevenue` would always be false in practice - since I don't know of any situation where `bidRequest.netRevenue` would be anything other than undefined. Could you explain the reasoning - who are you expecting to fill it out?

modules/scatteredBidAdapter.js
```diff
@@ -0,0 +1,175 @@
...
+    let { user } = ortb2Obj
+
+    let site = getConfig('site') || {};
+    if (ortb2Obj.site) {
+      mergeDeep(site, ortb2Obj.site)
+    }
+    // site.page    URL of the page where the impression will be shown.
+    // site.domain  Domain of the site (e.g., “mysite.foo.com”).
+    if (!site.page && bidderRequest.refererInfo) {
+      site.page = bidderRequest.refererInfo.referer;
```
dgirardi, line 67: Since version 7 `refererInfo` has `page`, `ref` and `domain` instead of `referer` - see [this PR](https://github.com/prebid/Prebid.js/pull/8450).

modules/scatteredBidAdapter.js
```diff
@@ -0,0 +1,175 @@
...
+    const bidderDomain = getKeyOnAny(validBidRequests, 'params.bidderDomain')
+
+    const device = getConfig('device') || {};
+    device.w = device.w || window.innerWidth;
+    device.h = device.h || window.innerHeight;
+    device.ua = device.ua || navigator.userAgent;
+
+    const imps = buildImps(validBidRequests)
+
+    const tid = getKeyOnAny(validBidRequests, 'transactionId');
```
dgirardi, line 80: source.tid should be the auctionId, and kept separate from imp transaction ids - see [this issue](https://github.com/prebid/Prebid.js/issues/8573)

dgirardi (commented):
I apologize for being so late - I completely missed this until you tagged me.

For new adapters that talk to ORTB backends we recommend using our [ORTB conversion library](https://github.com/prebid/Prebid.js/blob/master/libraries/ortbConverter/README.md); besides cutting down on code size, it would give you a free implementation of new Prebid features as they come out.

Since this adapter is so simple, I took the liberty of refactoring it to use that library as an extra incentive - [here's the source tree](https://github.com/dgirardi/Prebid.js/tree/scattered-patch) and the [diff](https://github.com/dgirardi/Prebid.js/commit/c1b5bbdf72eff220d2fbb5be802264d3542d5e64).

I you choose to take in my patch, the library logic invalidates all of my comments below except for the point about  `netRevenue`, where I kept the same behavior which I am not sure makes sense.

almendar:
Sure I'll update it. Sorry for delay I was dragged by other task from this. I'll try to fix this soon!

patmmccann:
@almendar we're very much looking forward to merging the adapter with @dgirardi 's suggestions

Do you have any concerns if upon merge we were to promote this js as an example of the new 'model' of publishing a concise bid adapter?

almendar:
Sorry for being silent!
@patmmccann No, absolutley I have nothing against promoting it :) 

@dgirardi I've applies your patch. Thank you very much for going extra mile and creating it for me :)

I think its ready for second round. In the mean time we've got our own GLV ID so I've added it to the spec. Also updated the sibling docs PR https://github.com/prebid/prebid.github.io/pull/4181

modules/scatteredBidAdapter.js
```diff
@@ -0,0 +1,97 @@
...
+'use strict';
+
+import { registerBidder } from '../src/adapters/bidderFactory.js';
+import { BANNER } from '../src/mediaTypes.js';
+import { deepAccess, logInfo, parseSizesInput } from '../src/utils.js';
+import { ortbConverter } from '../libraries/ortbConverter/converter.js';
+
+const BIDDER_CODE = 'scattered';
+const GVLID = 1179;
+export function buildImps(validBidRequests) {
```
dgirardi, line 11: I think I forgot to remove this in my patch, it doesn't look to be used by anything now.

modules/scatteredBidAdapter.js
```diff
@@ -0,0 +1,97 @@
...
+    return imp;
+  });
+}
+
+export const converter = ortbConverter({
+  context: {
+    mediaType: BANNER,
+    ttl: 360
+  },
+  bidResponse(buildBidResponse, bid, context) {
```
dgirardi, line 39: This method is no longer necessary (since it's not overriding anything), but you do need to set `netRevenue` somewhere; if it's always fixed at true or false you can set that through `context: {netRevenue: [true/false], ...}`

modules/scatteredBidAdapter.js
almendar: Done!

modules/scatteredBidAdapter.js
almendar: Both done!

almendar:
@dgirardi hopefully this time I've got it right :)  Thanks for the review!
3rd round?

dgirardi (approved):
LGTM, but beware that if I try to use the configuration in the md file, I get no bids - is that expected?

almendar:
Yes, this is just an example. We aim for federation of bidders, not all being public.

patmmccann (approved):

PR #9300, opened 2022-11-29
ttd Bid Adapter: add regression test topmost domain

gielow:
## Type of change

- [ ] Bugfix
- [ ] Feature
- [ ] New bidder adapter   
- [x] Build related changes
- [ ] CI related changes

- [ ] Other


gielow:
Adding a regression test for tydBidAdapter to avoid regression.

gielow:
@patmmccann

patmmccann:
thanks for this commit! Can you describe briefly the goal here? It seems to be adding tests to confirm the logic in refererDetection works for you going forward? Is this because you are comfortable with what is being sent over the wire with the latest merge @jlquaccia did? #9241

gielow:
> thanks for this commit! Can you describe briefly the goal here? It seems to be adding tests to confirm the logic in refererDetection works for you going forward? Is this because you are comfortable with what is being sent over the wire with the latest merge @jlquaccia did? #9241

We had an internal issue, so I created a test that replicated our client's scenario that failed with the previous code. Then I used the changes from #9241 to confirm it works as intended.
We also want to avoid that future changes to the common/core code causing regression to our clients.

patmmccann (approved):

PR #9302, opened 2022-11-30
Discovery Bid Adapter & Mediago Bid Adapter: add support for test request param

BaronJHYu:
## Type of change

- [ ] Bugfix
- [x] Feature
- [ ] New bidder adapter   
- [ ] Build related changes
- [ ] CI related changes

- [ ] Other


modules/mediagoBidAdapter.js
```diff
@@ -312,7 +310,7 @@ function getParam(validBidRequests, bidderRequest) {
   if (items && items.length) {
     let c = {
       id: 'mgprebidjs_' + auctionId,
-      test: +isTest,
+      test: Number(validBidRequests[0].params.test) || 0,
```
patmmccann, line 313: why is the syntax of the change different than above? Can you also document this param?

patmmccann (changes requested):

modules/mediagoBidAdapter.js
BaronJHYu: Because we want to change test status by Publisher. To make it easier to debug，we bid more frequently for test request.
We will update the document synchronously. And the default value be able to work properly.

patmmccann (approved):

PR #9304, opened 2022-11-30
Triplelift Adapter: Update referrer logic 

patrickloughrey:
## Type of change
- [x] Refactoring 

## Description of change
Reverting changes that were made [here](https://github.com/prebid/Prebid.js/pull/9178) since referrer logic was changed


patmmccann:
thanks very much! @jlquaccia ty for the fix!

PR #9305, opened 2022-11-30
Kargo Adapter: Update referrer logic

jsadwith:
## Type of change

## Description of change
Similar to #9304 , this reverts changes after the Prebid.js referer logic was updated


patmmccann:
thanks for the follow up!

patmmccann (approved):

PR #9317, opened 2022-12-06
Bump live-connect version

wi101:
## Type of change

- [ ] Bugfix
- [ ] Feature
- [ ] New bidder adapter   
- [ ] Code style update (formatting, local variables)
- [x] Refactoring (no functional changes, no api changes)
- [ ] Build related changes
- [ ] CI related changes

- [ ] Does this change affect user-facing APIs or examples documented on http://prebid.org?
- [ ] Other

## Description of change

Use the latest live-connect version and refactor the initialiser.

PR #9320, opened 2022-12-06
Revert "Bump live-connect version"

patmmccann:
Reverts prebid/Prebid.js#9317

This deletes an important feature, live connect minimal mode. Please document why you'd like to like to do this as well as summarize changes in the build dependency 

Appears to be a breaking change?


wi101:
Hi, This would use the minimal version if the `this.moduleMode === 'minimal'` we added that check in live-connect initializer [here](https://github.com/LiveIntent/live-connect/blob/master/src/initializer.js#L18) we pass the moduleMode and check it there.

patmmccann:
thanks!

PR #8496, opened 2022-05-31
Hadron RTD module: use internal methods

jlaso:
## Type of change

- [ ] Bugfix
- [ ] Feature
- [ ] New bidder adapter   
- [X] Code style update (formatting, local variables)
- [X] Refactoring (no functional changes, no api changes)
- [ ] Build related changes
- [ ] CI related changes
- [ ] Does this change affect user-facing APIs or examples documented on http://prebid.org?
- [ ] Other

## Description of change

Changing the way our JS code is loaded, using now loadExternalScript instead of document.append

- test parameters for validating bids
```
{
  bidder: '<bidder name>',
  params: {
    // ...
  }
}
```

Be sure to test the integration with your adserver using the [Hello World](/integrationExamples/gpt/hello_world.html) sample page.

- contact email of the adapter’s maintainer    joseluis.laso@audigent.com
- [X] official adapter submission

For any changes that affect user-facing APIs or example code documented on http://prebid.org, please provide:

- A link to a PR on the docs repo at https://github.com/prebid/prebid.github.io/

PR #9325, opened 2022-12-08
Confiant RTD Module : initial release

denys-berzoy-confiant:
## Type of change
- [x] Feature

## Description of change
Added Confiant RTD Provider Module

## Other information
Relevant docs PR: https://github.com/prebid/prebid.github.io/pull/4202


garygz-confiant (approved):

garygz-confiant:
@denys-berzoy-confiant looks like you need to pull the master - see conflicts above

modules/confiantRtdProvider.js
```diff
@@ -0,0 +1,126 @@
...
+    });
+  });
+  mutationObserver.observe(document.head, { childList: true, subtree: true });
+}
+
+/**
+ * Emit billable event when Confiant integration reports that it has monitored an impression
+ */
+function reportBillableEvents (e) {
+  events.emit(CONSTANTS.EVENTS.BILLABLE_EVENT, {
```
patmmccann, line 88: why bother? doesnt your js already do this emit?

modules/confiantRtdProvider.js
```diff
@@ -0,0 +1,126 @@
...
+/**
+ * Injects the Confiant Inc. configuration script into the page, based on proprtyId provided
+ * @param {string} propertyId
+ */
+function injectConfigScript(propertyId) {
+  const scriptElement = document.createElement('script');
+  scriptElement.type = 'text/javascript';
+  scriptElement.src = `https://cdn.confiant-integrations.net/${propertyId}/gpt_and_prebid/config.js`;
+
+  insertElement(scriptElement);
```
patmmccann, line 25: you must use our methods, please import the ajax utility

patmmccann (changes requested):
couple of minor comments in line

modules/confiantRtdProvider.js
denys-berzoy-confiant: @patmmccann no, there we no use cases to do that from our codebase.

modules/confiantRtdProvider.js
denys-berzoy-confiant: @patmmccann this is pretty much the same approach that Clean.io is taking to inject their script. Do you mean we should use ajax to fetch .js and the `eval()` it?

garygz-confiant: @patmmccann just like Clean.io our client asked to implement this to cover other demand as well native, GPT as well if they choose to do so.
So we are just replicating the approach that was approved for the Clean.io prebid module, see the link above.

garygz-confiant:
@robertrmartinez see this Confiant PR

modules/confiantRtdProvider.js
patmmccann: We're a confiant customer, and you have no trouble billing is, what value is this event adding

modules/confiantRtdProvider.js
patmmccann: They are not similar and the change request stands.

patmmccann: I looked at this again; the reason I flagged them as not similar is bc you linked the wrong part of the clean adapter, preloading is not against the rules (although we are considering it). The correct part of the clean adapter that inserts has been added to this list https://github.com/prebid/Prebid.js/issues/8421

garygz-confiant: @denys-berzoy-confiant it seems that the new way of inserting scripts is this: https://github.com/prebid/Prebid.js/pull/8496/files#diff-daa702a1f45a1ffc6ff61f98bfd8b733621e125a5525b3c2927e53294f2a0fd5R14
please update the PR

patmmccann: this is resolved now ty

garygz-confiant:
> @garygz-confiant only outstanding thing is explanation of how the billing event is used. This pr largely seems to be using prebid as a tag manager and adding this billing event.
> 
> Should your existing customers be switching to this implementation so they can be billed more accurately? If not, why fire the event? Is the event optional? Please make this clear in the module doc.

@patmmccann we are developing this for Magnite who explicitly requested the billable event to be communicated to them in a standard way. All of their vendors communicate this event in a unified way. While we have our own proprietary event (aka pixel) and can bill them based on that, that's not how they prefer to operate.

This is event is optional and configurable by a feature flag on our side (see this flag setup: https://github.com/prebid/Prebid.js/pull/9325/files#diff-d09c33d6d7490d64e1725f7700d943fcb2875f189a561eebf5506c5381ec4102R41). We would not require our existing customers to switch to processing this event as you correctly noticed our proprietary pixels are adequate. 

I see your point in terms of making it more explicitly parameterized and asking Magnite to initilize it while keeping the default as false for other potential clients who choose not to enable it and rather rely on our own billable reports.

@denys-berzoy-confiant Can you change the logic in here:
https://github.com/prebid/Prebid.js/pull/9325/files#diff-d09c33d6d7490d64e1725f7700d943fcb2875f189a561eebf5506c5381ec4102R41

Add an explicit flag:
shouldEmitBillableEvent: true|false
and use it in https://github.com/prebid/Prebid.js/pull/9325/files#diff-d09c33d6d7490d64e1725f7700d943fcb2875f189a561eebf5506c5381ec4102R41

By default it should be false. We will ask Magnite to explicitly enable it.

patmmccann:
@garygz-confiant thanks so much, please document your comments in your pr and i will merge

https://github.com/prebid/prebid.github.io/pull/4202/files

garygz-confiant:
> @garygz-confiant thanks so much, please document your comments in your docs pr, change that default as you indicated, and i will merge; next release is tomorrow so you should be good
> 
> https://github.com/prebid/prebid.github.io/pull/4202/files

@patmmccann We have adjusted the code as per our discussion. The billable event is off by default and we'll be enabled only by clients who explicitly require it:
https://github.com/prebid/Prebid.js/pull/9325/files#diff-01e26e1ce25e326e331c4ddf23dd08846e4fa2cb9234b78c11f5057d6b298014R40

patmmccann:
Lgtm

patmmccann (approved):

modules/confiantRtdProvider.js
```diff
@@ -0,0 +1,128 @@
+/**
+ * This module provides comprehensive detection of security, quality, and privacy threats by Confiant Inc,
+ * the industry leader in real-time detecting and blocking of bad ads
+ *
+ * The {@link module:modules/realTimeData} module is required
+ * The module will inject a Confiant Inc. script into the page to monitor ad impressions
+ * @module modules/cleanioRtdProvider
```
muuki88, line 7: @denys-berzoy-confiant I guess this should be confiant 😅

Issue #9080, opened 2022-10-05
Prebid Server Adapter: participate in fledge auctions


patmmccann:
## Type of issue
Feature Request

## Description
The Fledge module takes an auctionConfig that must exist somewhere on the bid response. The PBS bid adapter should have a 'place' for them to be. 

https://github.com/prebid/Prebid.js/blob/b04ba3b09b35c5edc4588b569de618d0899ae432/modules/openxOrtbBidAdapter.js#L329 is where they are on the OpenXOrtb adapter response. Open to proposals on where they should be on PBS response, potentially somewhere in meta? 

Also pbs adapter should include a unit test it gets imp.ext.ae on fledge auctions just as https://github.com/prebid/Prebid.js/blob/af8fb3816be54406fd69851acc8f6fe0565f9f18/modules/openxOrtbBidAdapter.js#L72

Merging multiple auction configs in one response should be covered in the solution

patmmccann:
If we decide it exists in the meta response; there might be no server side retooling needed at all. Server just passes through meta unaltered.

patmmccann:
Correction: it appears to pass through whitelisted fields, eg https://github.com/prebid/prebid-server/issues/2157

dgirardi:
Fledge configs are independent from bids, so  `seatnonbid` is not a great place for it if it's meant to be populated only when `seatbid` is not. Counter proposal:

```
{
  ext: {
      prebid: {
           fledge: [
                 {
                    impid,  // imp.id the fledge auction refers to
                    seat,    // bidder code asking for the fledge auction. (not required, but probably useful for analytics)
                    config: {
                         // the fledge auction configuration - just a pass-through object. Examples:
                         // https://developer.chrome.com/blog/fledge-api/#ad-auction
                         // https://developers.google.com/publisher-tag/reference#googletag.config.ComponentAuctionConfig_auctionConfig                       
                    },
                    // ...
                 }
           ]
      }
  }
}
```

laurb9:
The fledge auction config is passed out-of-band with respect to the bids (or no bids), it does not contain bid information but rather a specially crafted bid request reflected back to the browser. Since `seatbid.nobid` is intended for debugging information (non-functional), `seatbid.bids` are intended to carry actual bids, and even `seatbid` could be missing entirely if `nbr` exists, I think it would be simpler to keep passing the fledge auctionconfig in the `response.ext.fledgeauctionconfig`.

[...]

PR #9342, opened 2022-12-14
Prebid Server adapter: fledge support

dgirardi:
## Type of change

- [x] Feature

## Description of change

This adds _untested_ support for fledge in the PBS adapter, following the spec described [here](https://github.com/prebid/prebid-server/issues/2411#issuecomment-1303895661).

In addition this also refactors some of the other logic around fledge:

 - the `interpretResponse` return value API has been changed from `{bids: [...], fledgeAuctionConfigs: [{bidId, ...config}] }` to `{bids: [...], fledgeAuctionConfigs: [{bidId, config}]}`
 - the `addComponentAuction` signature has been changed from `(bidRequest, fledgeAuctionConfig)` to `(adUnitCode, fledgeAuctionConfig)`
 - the logic to flag `bidderRequest.fledgeEnabled` has been moved to the fledge module
 - `ortbConverter` now automatically sets `imp.ext.ae` for all adapters; and parses `ext.prebid.fledge` for those that choose to use PBS extensions.

## Other information

Closes https://github.com/prebid/Prebid.js/issues/9080
Documentation TBD


test/spec/modules/fledge_spec.js
```diff
@@ -35,3 +42,166 @@ describe('fledgeForGpt module', function() {
     });
```
robertrmartinez, line 42: prob does not matter but this is still calling `bidRequest` instead of `adUnitCode` for `addComponentAuctionHook`

Line 40 in `fledge_spec`

JoelPM:
@laurb9  - mind giving this a review?

laurb9:
@dgirardi I tested this PR with openx adapter, both direct and via prebid-server (both varieties), and it is working for me.

Setup:
I set up three bidders on a local test page: `openx` (because it supports FLEDGE already), plus two s2s configs with custom endpoints for PBS-Go and PBS-Java, both also calling aliased `openx`.
Local test bidder that responds with a bid and fledge auction config like the `openx` endpoint would.
Started dev Prebid.js with `fledgeForGpt`,`openxOrtbBidAdapter`,`prebidServerBidAdapter` modules.
Started latest [PBS-Go](https://github.com/prebid/prebid-server) and [PBS-Java](https://github.com/prebid/prebid-server-java) from their repos, configured with that same local test bidder endpoint for `openx`.

Result:
I see all three "fledgeForGpt register component auction config" messages corresponding to each bidder response on the GPT slot.

Prebid-Fledge with FLEDGE enabled

laurb9:
I've also tested on a stable Chrome without fledge support, and I see that `imp.ext.ae` is no longer being sent and the eventual s2s fledge auction configs get the expected warning:
> Prebid WARNING: Received fledge auction configuration for an impression that was not in the request or did not ask for it {impid: 'div-gpt-ad-1460505748561-0', bidder: 'openx-via-pbs-go', adapter: 'openx', config: {…}}....

But the fledge configs coming from adapters still attempt to register the component auction, as the original code made no attempt to filter them either. I see it as more of a consistency issue now, as nothing will happen either way - or we could disable the module altogether so no fledge processing takes place.

modules/fledgeForGpt.js
```diff
@@ -31,12 +32,10 @@ export function init(cfg) {
   }
 }
 
-export function addComponentAuctionHook(next, bidRequest, componentAuctionConfig) {
+export function addComponentAuctionHook(next, adUnitCode, componentAuctionConfig) {
   const seller = componentAuctionConfig.seller;
-  const adUnitCode = bidRequest.adUnitCode;
   const gptSlot = getGptSlotForAdUnitCode(adUnitCode);
   if (gptSlot && gptSlot.setConfig) {
-    delete componentAuctionConfig.bidId;
```
laurb9, line 39: We deleted `bidId` because it was not in the schema, on the off chance the object might be checked against a schema for extra unapproved fields, given it is intended to be passed to sandboxed code.

modules/fledgeForGpt.js
```diff
@@ -31,12 +32,10 @@ export function init(cfg) {
   }
 }
 
-export function addComponentAuctionHook(next, bidRequest, componentAuctionConfig) {
+export function addComponentAuctionHook(next, adUnitCode, componentAuctionConfig) {
```
laurb9, line 35: Would a hook not benefit more from seeing the original bidRequest ?

modules/fledgeForGpt.js
```diff
@@ -48,5 +47,50 @@ export function addComponentAuctionHook(next, bidRequest, componentAuctionConfig
     logWarn(MODULE, `unable to register component auction config for: ${adUnitCode} x ${seller}.`);
   }
 
-  next(bidRequest, componentAuctionConfig);
+  next(adUnitCode, componentAuctionConfig);
 }
+
+export function markBidsForFledge(next, bidderRequests) {
+  if (navigator.runAdAuction) {
```
laurb9, line 54:
```suggestion
  if ('runAdAuction' in navigator && 'joinAdInterestGroup' in navigator) {
```
to align with what GPT looks for

modules/fledgeForGpt.js
dgirardi: possibly, but we don't always have it. In prebid server we may have multiple requests (the fledge config references an imp, but there can be any number of bidders (requests) for that imp). With `allowUnknownBidderCodes` we may also have zero.

patmmccann:
@laurb9 how do things look now?

modules/fledgeForGpt.js
```diff
@@ -48,5 +47,50 @@ export function addComponentAuctionHook(next, bidRequest, componentAuctionConfig
     logWarn(MODULE, `unable to register component auction config for: ${adUnitCode} x ${seller}.`);
   }
 
-  next(bidRequest, componentAuctionConfig);
+  next(adUnitCode, componentAuctionConfig);
 }
+
+export function markBidsForFledge(next, bidderRequests) {
```
laurb9, line 53: This function enables FLEDGE for bidderRequests, should the name be `markBidderRequestsForFledge` or just `markForFledge` ?

modules/fledgeForGpt.js
```diff
@@ -48,5 +47,50 @@ export function addComponentAuctionHook(next, bidRequest, componentAuctionConfig
...
+registerOrtbProcessor({type: IMP, name: 'impExtAe', fn: setImpExtAe});
+
+// to make it easier to share code between the PBS adapter and adapters whose backend is PBS, break up
+// fledge response processing in two steps: first aggregate all the auction configs by their imp...
+
+export function parseExtPrebidFledge(response, ortbResponse, context) {
+  (ortbResponse.ext?.prebid?.fledge?.auctionconfigs || []).forEach((cfg) => {
+    const impCtx = context.impContext[cfg.impid];
+    if (!impCtx?.imp?.ext?.ae) {
+      logWarn('Received fledge auction configuration for an impression that was not in the request or did not ask for it', cfg, impCtx?.imp);
```
laurb9, line 77: validation behavior is different between PBS and regular adapter. If adapters adopt this method it may become a non-issue but it may be confusing as is.

modules/fledgeForGpt.js
```diff
@@ -48,5 +47,50 @@ export function addComponentAuctionHook(next, bidRequest, componentAuctionConfig
     logWarn(MODULE, `unable to register component auction config for: ${adUnitCode} x ${seller}.`);
```
laurb9, line 47: Should `addComponentAuction` be skipped altogether if FLEDGE is not supported ?
On such a browser, receiving unsolicited auction configs from an adapter will print this misleading under the circumstances warning.

laurb9 (approved):
Looks good to me and it works. After playing with it I noticed a couple new and inherited  behaviors that don't impact functionality but can be misinterpreted, so I added a few comments.

Also, the log message `fledgeForGpt isEnabled: true` is emitted on a non-FLEDGE supporting browser, with no indication that we know it will not work and that `imp.ext.ae` is removed from explicit pub-supplied configuration. I've come to think it is more helpful to see the final computed state of the enablement that considers the browser support, perhaps with some detail about how it was derived.

modules/fledgeForGpt.js
dgirardi: I am not sure this validation makes sense outside of here. There are two things we are checking here - 

1. the request should have set `imp.ext.ae`; does this check make sense for an adapter that (in general) may not be using ORTB at all? 
2. the fledge config should refer to an imp that existed in the request. this is necessary for PBS because the protocol pairs a fledge config with `impid`. But, for client-side adapters,  they must provide a valid `bidId` instead - [the check is here](https://github.com/prebid/Prebid.js/blob/c05bba2920ebe3a97b231e058acc2c09db27bf10/src/adapters/bidderFactory.js#L261), although there's no warning.

I can add a warning for 2, I'm not sure about 1.

modules/fledgeForGpt.js
laurb9: The message "unable to do something" tells me the expectation was that it should have been able to, so something went wrong.
If it said "ignored unsolicited fledge configs", that would be more accurate, but I understand we may not want to have a log message for every specific condition

modules/fledgeForGpt.js
laurb9: Should be fine to leave it as it is now and update in a later iteration if it becomes a problem with usage.

mmoschovas (approved):

PR #9345, opened 2022-12-15
yahoospp bidder& aol bidder: GPP Support in bid requests

radubarbos:
## Type of change

- [ ] Bugfix
- [ X] Feature
- [ ] New bidder adapter   
- [ ] Build related changes
- [ ] CI related changes

- [ ] Other

### DIFF
--- a/modules/aolBidAdapter.js
+++ b/modules/aolBidAdapter.js
@@ -335,11 +335,6 @@ export const spec = {
       deepSetValue(openRtbObject, 'regs.ext.us_privacy', consentData.uspConsent);
     }
 
-    if (consentData.gppConsent && consentData.gppConsent.gppString) {
-      deepSetValue(openRtbObject, 'regs.ext.gpp', consentData.gppConsent.gppString);
-      deepSetValue(openRtbObject, 'regs.ext.gpp_sid', consentData.gppConsent.applicableSections);
-    }
-
     if (typeof bid.userId === 'object') {
       openRtbObject.user = openRtbObject.user || {};
       openRtbObject.user.ext = openRtbObject.user.ext || {};
--- a/modules/aolBidAdapter.js
+++ b/modules/aolBidAdapter.js
@@ -159,6 +159,7 @@ export const spec = {
     if (bidderRequest) {
       consentData.gdpr = bidderRequest.gdprConsent;
       consentData.uspConsent = bidderRequest.uspConsent;
+      consentData.gppConsent = bidderRequest.gppConsent;
     }
 
     return bids.map(bid => {
@@ -334,6 +335,11 @@ export const spec = {
       deepSetValue(openRtbObject, 'regs.ext.us_privacy', consentData.uspConsent);
     }
 
+    if (consentData.gppConsent && consentData.gppConsent.gppString) {
+      deepSetValue(openRtbObject, 'regs.ext.gpp', consentData.gppConsent.gppString);
+      deepSetValue(openRtbObject, 'regs.ext.gpp_sid', consentData.gppConsent.applicableSections);
+    }
+
     if (typeof bid.userId === 'object') {
       openRtbObject.user = openRtbObject.user || {};
       openRtbObject.user.ext = openRtbObject.user.ext || {};
@@ -371,6 +377,11 @@ export const spec = {
 
     if (consentData.uspConsent) {
       params.us_privacy = consentData.uspConsent;
+    }
+
+    if (consentData.gppConsent && consentData.gppConsent.gppString) {
+      params.gpp = consentData.gppConsent.gppString;
+      params.gpp_sid = consentData.gppConsent.applicableSections;
     }
 
     return params;
--- a/modules/aolBidAdapter.js
+++ b/modules/aolBidAdapter.js
@@ -160,6 +160,12 @@ export const spec = {
       consentData.gdpr = bidderRequest.gdprConsent;
       consentData.uspConsent = bidderRequest.uspConsent;
       consentData.gppConsent = bidderRequest.gppConsent;
+      if (!consentData.gppConsent && bidderRequest.ortb2?.regs?.gpp) {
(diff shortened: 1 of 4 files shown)


modules/yahoosspBidAdapter.js
```diff
@@ -184,46 +184,46 @@ function validateAppendObject(validationType, allowedKeys, inputObject, appendTo
...
         if (allowedKeys.indexOf(objectKey) !== -1 && isArray(inputObject[objectKey])) {
           outputObject[objectKey] = inputObject[objectKey];
-        };
+        }
         break;
       case 'object':
         if (allowedKeys.indexOf(objectKey) !== -1 && isPlainObject(inputObject[objectKey])) {
           outputObject[objectKey] = inputObject[objectKey];
-        };
+        }
```
patmmccann, line 203: please keep all these semicolons, we plan to turn on a no implied semicolon linting rule in the future

modules/yahoosspBidAdapter.js
```diff
@@ -277,21 +277,29 @@ function generateOpenRtbObject(bidderRequest, bid) {
       }
     } else {
       outBoundBidRequest.site.id = bid.params.dcn;
-    };
+    }
+
+    if (bidderRequest.gppConsent) {
+      outBoundBidRequest.regs.ext.gpp = bidderRequest.gppConsent.gppString;
+      outBoundBidRequest.regs.ext.gpp_sid = bidderRequest.gppConsent.applicableSections;
+    } else if (bidderRequest.ortb2?.regs?.gpp) {
```
patmmccann, line 285: why doesnt aol have this same logic?

patmmccann (changes requested):
In addition to the in line comments; why haven't you passed GPP on your user syncs?

modules/aolBidAdapter.js
```diff
@@ -373,6 +374,11 @@ export const spec = {
       params.us_privacy = consentData.uspConsent;
     }
 
+    if (consentData.gppConsent && consentData.gppConsent.gppString) {
```
patmmccann, line 377: why not mirror the logic below from yahoo?

modules/yahoosspBidAdapter.js
radubarbos: addressed

modules/yahoosspBidAdapter.js
radubarbos: the aol bidder adapter is the old one we are migrating off. Also there we have undocumented support for openRTB config, we don't want to add this support in the legacy adapter.

modules/aolBidAdapter.js
radubarbos: this is a legacy adapter sending the requests to some legacy endpoints as GET requests. The POST openRTB is not officially supported. We don't apply the openRTB logic here.

radubarbos:
> In addition to the in line comments; why haven't you passed GPP on your user syncs?

The user sync pixel is provided by our back-end with all the needed params, it is provided in the response.body.bidResponse.ext.pixels field.

modules/aolBidAdapter.js
patmmccann: The only current place you can expect to find the gpp string is in the location yahoo falls back to based on publisher setConfigs.

patmmccann:
> 

https://github.com/prebid/Prebid.js/blob/1b42c9a6692fa68716b216dfd2f4d15da9fd7c7a/modules/yahoosspBidAdapter.js#L634 you pass yourself consent in the syncs function but don't act on it. How does your backend become aware of consent? This seems quite unusual

modules/yahoosspBidAdapter.js
dgirardi: if `bidderRequest.gppConsent` is defined, then the `ortb2.regs` properties are as well; you can save some bytes and complexity by looking only at the latter. 

In fact, you could replace a lot of the code in here (esp. the giant `appendFirstPartyData` function) with just `mergeDeep(outBoundBidRequest, bidderRequest.ortb2)`.

radubarbos:
We are aware of this very small chance that we could use a stale consent, it has always worked like this, we'll revisit together with the back-end team how we could improve this.
How long can the user syncs pixels could be hold in memory before they are fired? how big this delay could be?

modules/aolBidAdapter.js
radubarbos: I added the fallback support similar to the yahoo adapter logic.

patmmccann:
This delay could easily be up to 60 seconds, as publishers have a configuration parameter allowing their delay. The USP spec, and GPP has been less clear, says consent must be rechecked from the cmp each time it is needed.

modules/yahoosspBidAdapter.js
patmmccann: https://github.com/prebid/Prebid.js/blob/d5746c39c437b5e04505cdd75bdef3791eb28566/modules/consentManagementGpp.js#L380

You can mergeDeep regs, site, app, and user

radubarbos: I updated it to always use the ortb2 data. For the regs we can not mergeDeep, our backend supports openRTB 2.5 and the gpp under "regs.ext.gpp".

radubarbos:
Thanks for the information. We'll discuss this with the back-end teams.

patmmccann:
@radubarbos just confirming, we're waiting for the results of your internal discussion?

radubarbos:
@patmmccann We would like to work on that as a different ticket/task. We still have to coordinate with the back-end to discuss. If there are no more comments on these particular changes we would liked them merged. 
This opinion is also shared by @slimkrazy

patmmccann (approved):

patmmccann:
@radubarbos we haven't heard from you in a while. We're planning to disclose our concerns in your documentation https://github.com/prebid/prebid.github.io/pull/4529/files

PR #9346, opened 2022-12-15
Appush Bid Adapter: initial release

nitzanappush:
doc - https://github.com/prebid/prebid.github.io/pull/4220


Rothalack (approved):
Looks good, working in testing. I appreciate that you'll return a test ad with the test placements!

PR #9350, opened 2022-12-19
Build system: set up `hook` for tests

### LINKED ISSUE #9348: Enabling local storage in unit tests via the getStorageManager()
## Type of issue
Testing local storage in unit tests does not appear to be possible.

## Description

I want to test that our use of `getStorageManager()` with `gvlid` and `bidderCode` specified is working correctly. Nevertheless, whenever I run my unit test, `isValid(cb)` within `storageManager.js` always returns undefined because the callback in `validateStorageEnforcement` is never executed.

## Steps to reproduce

In any unit test, set up the following:
```
    const storage = getStorageManager({gvlid: 855, bidderCode: 'adnuntius'});
     storage.localStorageIsEnabled(); // always returned undefined
```
You will then find that the callback in `validateStorageEnforcement` via `isValid(cb)` is never executed.

### Expected results
There should be a way to enable local storage in unit tests.

## Other information

This is what I use to run the tests: `gulp test --file ./test/spec/modules/adnuntiusBidAdapter_spec.js`

dgirardi:
## Type of change

- [x] Build related changes

## Description of change

This adds some global test setup to make functionality that depends on hooks (such as `storageManager`) work during tests. 

Normally a test run would do this already, indirectly through `pbjs.processQueue()`; but if one runs only a single spec file that does not import the Prebid machinery itself, tests fail as in #9348.

## Other information

Closes #9348

### DIFF
--- a/karma.conf.maker.js
+++ b/karma.conf.maker.js
@@ -110,7 +110,7 @@ module.exports = function(codeCoverage, browserstack, watchMode, file, disableFe
   var webpackConfig = newWebpackConfig(codeCoverage, disableFeatures);
   var plugins = newPluginsArray(browserstack);
 
-  var files = file ? ['test/test_deps.js', file].flatMap(f => f) : ['test/test_index.js'];
+  var files = file ? ['test/test_deps.js', file, 'test/helpers/hookSetup.js'].flatMap(f => f) : ['test/test_index.js'];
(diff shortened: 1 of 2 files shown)


ChrisHuie (approved):

patmmccann (commented):
LGTM

Issue #9141, opened 2022-10-21
RTD Modules: segments added to a request persist into additional requests


patmmccann:
## Type of issue
Bug

## Description
JW rtd module adding segments to ortb2 persist beyond the auction they were added on. The JW RTD module should only add segments to video requests, but when they do, it seems the method they use persists the segments into all subsequent requests.

## Steps to reproduce
Examine prebid server outgoing requests for AdThrive_Content_1 unit

## Test page
https://cafedelites.com/chicken-tikka-masala/

### Expected results
JW segments added to the video request

### Actual results
JW segments added to the video request and all other subsequent auctions

karimMourra:
@patmmccann thanks for submitting this. Could you please provide more detailed reproduction steps ? On my end i've scrolled down and interacted with the video but I don't see a AdThrive_Content_1 ad unit, i see a AdThrive_Content_1_Desktop banner ad unit. 
<img width="548" alt="Screen Shot 2022-10-21 at 4 48 22 PM" src="https://user-images.githubusercontent.com/12138319/197286445-7b36e1f7-e5fa-4122-892a-bbf647504af9.png">

Is this the right ad unit  ?

patmmccann:
Can you look for requests that occur after the video request?

dgirardi:
@patmmccann, I am also trying to reproduce and asking for better steps. That page makes a ton of requests and I am having trouble finding a video request for AdThrive_Content_1

karimMourra:
@dgirardi my guess is that the JW RTD module is writing to `ortb2Fragments.global` and those changes are persisting after the auction. Should the JW RTD module have to clean the `ortb2Fragments.global` after the auction ?

dgirardi:
@karimMourra no, `ortb2Fragments` is generated anew each auction - or at least it should.

patmmccann:
@mmoschovas can you please full steps to reproduce?

mmoschovas:
Just to clarify, the issue impacts any ad unit after the JW targeting is applied. For example, we only apply the adunit targeting to the instream video unit but the segments are applied globally to every ad unit regardless if it has video mediatype or not. Here is an example that shows the Footer ad unit configuration (via the console) and the jw segments that are being applied to the network requests

<img width="1446" alt="Screen Shot 2022-10-28 at 4 07 34 PM" src="https://user-images.githubusercontent.com/63253416/207450453-7ddf27f7-24a4-4073-8e34-27f5784fbed6.png">

[...]

PR #9355, opened 2022-12-20
Prebid core: filter adUnits (by `adUnitCodes`) before sending them to RTD and FPD modules

dgirardi:
## Type of change

- [x] Bugfix

## Description of change

This changes the input of RTD and FPD modules so that they only see adUnits that are included in the auction. This is to help prevent them from accidentally working on adUnits that are not relevant - as dissussed [here](https://github.com/prebid/Prebid.js/issues/9141).

## Other information
Closes https://github.com/prebid/Prebid.js/issues/9141


dgirardi:
Note that there will (temporarily) be an unrelated test failure that I am addressing separately in https://github.com/prebid/Prebid.js/pull/9354

karimMourra (approved):

patmmccann (approved):

Issue #9176, opened 2022-10-31
Feature request: option to timeout CMP interaction and readiness independently


EskelCz:
## Type of issue
Feature request, probably

## Description
Some of our publishers would like to run auction as soon as possible, without waiting for user interaction with the CMP banner. We do this by setting `consentManagement.gdpr.timeout: 0`.
This works fine if the CMP manages to load in time to provide stored consent. If it doesn't, prebid sends tcString `undefined`.
I think the issue is that the timeout intermingles two time intervals relevant for different use cases.

1. The CMP being ready (to provide stored tcString from cookie, if available)
2. The user`s action (to provide new consent)

We would like to propose separating it into these two distinct timeouts. Something like `cmpTimeout` and `actionTimeout`.

We can work around this by checking the CMP readiness ourselves before calling pbjs.requestBids but I think it would be cleaner if this was supported by Prebid configuration itself. Possibly also helpful to others.

## Platform details
Latest Prebid 7.23.0

## Other information
The behavior might have been caused by this change: https://github.com/prebid/Prebid.js/pull/8524

patmmccann:
Your proposal sounds reasonable to me. So by default we would have the current behavior, but if actionTimeout exists, we would have the new behavior. Sound right?

EskelCz:
@patmmccann Might be a bit confusing but I understand that you need to retain backward compatibility. So I guess it's reasonable.

PR #9365, opened 2022-12-21
Consent Management: Added config option for user action timeout

jlquaccia:
## Type of change
- [x] Feature

## Description of change
The following flow will play out on a user's initial visit to a publisher site if the action timeout is in use:
- the default timeout setting will be put to use (waiting for the cmp to load)
- if/when the cmp loads, then the action timeout will additionally kick in (overriding the first timeout) in an attempt to wait for the user to provide consent
- auction will then run either after user consent or after the action timeout expires (providing no other errors get in the way of the auction)

## Other information
https://github.com/prebid/Prebid.js/issues/9176

### DIFF
--- a/modules/consentManagement.js
+++ b/modules/consentManagement.js
@@ -25,8 +25,9 @@ export let staticConsentData;
 let consentData;
 let addedConsentHook = false;
 let provisionalConsent;
-let iabCmpHasLoaded = false;
 let onTimeout;
+let timer = null;
+let actionTimer = null;
 
 // add new CMPs here, with their dedicated lookup function
 const cmpCallMap = {
@@ -45,8 +46,7 @@ function lookupStaticConsentData({onSuccess, onError}) {
 export function setActionTimeout(timeout = setTimeout) {
   clearTimeout(timer);
   timer = null;
-  timer = timeout(onTimeout, actionTimeout);
-  iabCmpHasLoaded = true;
+  actionTimer = timeout(onTimeout, actionTimeout);
 }
 
 /**
@@ -94,7 +94,7 @@ function lookupIabConsent({onSuccess, onError}) {
         processCmpData(tcfData, {onSuccess, onError});
       } else {
         provisionalConsent = tcfData;
-        if (!iabCmpHasLoaded && timer != null) setActionTimeout();
+        if (!isNaN(actionTimeout) && actionTimer === null && timer != null) setActionTimeout();
       }
     } else {
       onError('CMP unable to register callback function.  Please check CMP setup.');
@@ -167,8 +167,6 @@ function lookupIabConsent({onSuccess, onError}) {
   }
 }
 
-let timer = null;
-
 /**
  * Look up consent data and store it in the `consentData` global as well as `adapterManager.js`' gdprDataHandler.
  *
@@ -183,6 +181,12 @@ export function loadConsentData(cb, callMap = cmpCallMap, timeout = setTimeout) 
       clearTimeout(timer);
       timer = null;
     }
+
+    if (actionTimer != null) {
+      clearTimeout(actionTimer);
+      actionTimer = null;
+    }
+
     isDone = true;
     gdprDataHandler.setConsentData(consentData);
     if (typeof cb === 'function') {
--- a/modules/consentManagement.js
+++ b/modules/consentManagement.js
@@ -25,8 +25,6 @@ export let staticConsentData;
 let consentData;
 let addedConsentHook = false;
 let provisionalConsent;
-// let timer = null;
-// let isDone;
 let loadIfMissingCb;
 let iabCmpHasLoaded = false;
 let shouldWaitForActionTimeout = false;
--- a/modules/consentManagement.js
+++ b/modules/consentManagement.js
@@ -25,10 +25,11 @@ export let staticConsentData;
 let consentData;
 let addedConsentHook = false;
 let provisionalConsent;
-let timer = null;
-let isDone;
+// let timer = null;
+// let isDone;
 let loadIfMissingCb;
 let iabCmpHasLoaded = false;
+let shouldWaitForActionTimeout = false;
 
 // add new CMPs here, with their dedicated lookup function
 const cmpCallMap = {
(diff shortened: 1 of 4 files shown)


modules/consentManagement.js
```diff
@@ -188,22 +189,29 @@ function loadConsentData(cb) {
...
       })
     }
     if (consentTimeout === 0) {
       onTimeout();
+    } else if (useActionTimeout && actionTimeout) {
+      if (timer != null) {
+        clearTimeout(timer);
+      }
+
+      timer = timeout(onTimeout, actionTimeout);
```
dgirardi, line 212: The use case for this - from how I  interpret https://github.com/prebid/Prebid.js/issues/9176 - is a pub that says, "I want to wait a bit to get a valid consent string from the CMP, but I don't want to wait any time for the user to provide consent". That would be expressed with something like `{ actionTimeout: 0, timeout: 10000}`, meaning "wait up to 10 seconds for the user to give consent, but no longer than 0 seconds after the CMP has loaded".

With this PR you would set a timeout of 0 without waiting for the CMP to load. I think the way to do this is to start the `actionTimeout` timer here:

https://github.com/prebid/Prebid.js/blob/1960ed811bd923a445aece06b20f06e11589bf1d/modules/consentManagement.js#L86

that's when we get the first message from the CMP, which should always contain valid consent data in `provisionalConsent` (although it may deny all consent if the user hasn't clicked the button yet).

patmmccann: @EskelCz is this an accurate characterization of the setting you desire? I was suspecting you wanted to lower the timeout when there is no interaction pending. 

eg { actionTimeout: 10000, timeout: 200} so that we only have the long timeout if the action is pending

EskelCz: @dgirardi @patmmccann
To clarify, most of our publishers don’t actually want to wait for any interaction (often their users don't click the banner at all), but we need to wait for the CMP to be loaded, to return some consent string (either default or read from storage). Therefore, we would probably go with something like `{ actionTimeout: 0, timeout: 200 }`.

Other publishers that are willing to wait for interaction (in case of the open banner / action pending) would use `{ actionTimeout: 10000, timeout: 200 }`, the 10s for example.

If that would be a breaking change, I think even clearer naming could be used, like gdpr.actionTimeout and gdpr.loadTimeout.

@dgirardi
I’m not sure about the internal implementation, whether the `provisionalConsent` would have the consent that was read from storage.

jlquaccia: Just refactored the actionTimeout logic here.  if the cmp loads (within the `timeout` delay) on an initial visit (no user consent) and the actionTimeout is present in the config, the actionTimeout will be used at this point.  if the cmp doesn't load in time, the auction will just proceed.  Something to note, I did notice with the provisional consent that the consent string initially returns as an empty string.  So even if we wait for the cmp to load first (when a user has not yet consented), the consent string will get set as undefined once the auction starts.

modules/consentManagement.js
```diff
@@ -84,6 +88,10 @@ function lookupIabConsent({onSuccess, onError}) {
         processCmpData(tcfData, {onSuccess, onError});
       } else {
         provisionalConsent = tcfData;
+        if (!iabCmpHasLoaded && !isNaN(actionTimeout) && shouldWaitForActionTimeout && timer != null) {
+          loadConsentData(loadIfMissingCb, true);
```
dgirardi, line 92: This adds a lot of complexity that I think is unnecessary. `loadConsentData` now needs to do two completely different things based on the argument; I think it makes more sense to not mix them.

Why not just start a timeout here, as you do in line 255? To me it looks like it would get rid of all the other changes you have here.

jlquaccia:
@dgirardi did what you said and refactored and simplified things a bit.  can you review again when you have a chance?

modules/consentManagement.js
dgirardi: @EskelCz I think it depends on the CMP, not all of them will provide a "null" string. If it's working for what you're using now it should not be affected.

modules/consentManagement.js
```diff
@@ -156,19 +167,21 @@ function lookupIabConsent({onSuccess, onError}) {
   }
 }
 
+let timer = null;
```
dgirardi, line 170: the timer for `actionTimeout` should be kept separate from the original `timeout` I think, to keep the same semantics for the latter.

modules/consentManagement.js
```diff
@@ -84,6 +94,7 @@ function lookupIabConsent({onSuccess, onError}) {
         processCmpData(tcfData, {onSuccess, onError});
       } else {
         provisionalConsent = tcfData;
+        if (!iabCmpHasLoaded && timer != null) setActionTimeout();
```
dgirardi, line 97: the check here should be on whether `actionTimeout` was set; without it I thik this would fire a timeout of 0.

jlquaccia:
thanks @dgirardi, just pushed a new commit

dgirardi (approved):

jsnellbaker (approved):
I think this looks good.  If it hasn't been made yet, this needs a docs PR.

jlquaccia:
@jsnellbaker just created a docs PR for this.  Can you review when you have time?

jsnellbaker:
Docs were created; merging at this time.

EskelCz:
Thanks for the feature to everyone involved.
We just got to testing it in production websites and ran into an issue. For some reason the `timer` never starts (stays null), therefore the action condition always fails, which is the worst outcome for us (auction not starting at all).
consent module

Should I open a new issue for it?

Reproduction steps:
1. Use resource override / requestly to replace production prebid (7.26, before this change) to the unminified 7.35:
https://cdn.cpex.cz/hb/prebid/prebid.min.js > https://cdn.cpex.cz/stage/hb/prebid/v7.35/prebid.js
2. Visit [https://www.csfd.cz/?debug](https://www.csfd.cz/?debug), preferably in anonymous window, so you are sure to get the consent banner
3. See console, the auction isn't being started, with gdpr `timeout` and `actionTimeout` set to 1000.

patmmccann:
Yes please open a new one

PR #9370, opened 2022-12-23
Yieldlab Bid Adapter: read and pass UserIdsAsEids atype information

nkloeber:
## Type of change

- [x] Feature

## Description of change

Read atype information from UserIdsAsEids and pass it as query parameter (atypes={idprovider}:{atype},{idprovider2}:{atype2},...)

## Other information

One approving review from a Yieldlab team member required. (@brushmate, @rey1128, @Mufas61 or @kippsterr)

### DIFF
--- a/modules/yieldlabBidAdapter.js
+++ b/modules/yieldlabBidAdapter.js
@@ -306,17 +306,17 @@ function createUserIdString(eids) {
 
 /**
  * Creates a string from an array of eids with ID provider and atype if atype exists
- * @param {Array} eids
+ * @param {Array.<{source: String, uids: Array.<{id: String, atype: Number, ext: Object}>}>} eids
  * @returns {String} idprovider:atype,idprovider2:atype2,...
  */
 function createUserIdAtypesString(eids) {
-  const str = []
+  const str = [];
   for (let i = 0; i < eids.length; i++) {
     if (eids[i].uids[0].atype) {
-      str.push(eids[i].source + ':' + eids[i].uids[0].atype)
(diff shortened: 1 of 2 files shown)


modules/yieldlabBidAdapter.js
```diff
@@ -303,6 +304,21 @@ function createUserIdString(eids) {
...
+ * @returns {String} idprovider:atype,idprovider2:atype2,...
+ */
+function createUserIdAtypesString(eids) {
+  const str = []
+  for (let i = 0; i < eids.length; i++) {
+    if (eids[i].uids[0].atype) {
+      str.push(eids[i].source + ':' + eids[i].uids[0].atype)
+    }
+  }
+  return str.join(',')
```
patmmccann, line 319: Use semicolons please

kippsterr: Hi @patmmccann,

we have not been using semicolons anywhere in the file so far. Also running `gulp lint` does not print any issues.

Would it make sense to add that kind of stuff to the linting rules and thus enforce consistency everywhere?

patmmccann: we've attempted to do so, but fixing all the issues is a work in progress

https://github.com/prebid/Prebid.js/pull/8846

kippsterr: That sounds reasonable. I have created #9386 to generally fix that for our adapter and will add semi colons to this PR as well.

ChrisHuie (approved):

PR #9378, opened 2022-12-27
btRecovery RTD Module: initial release

MariiaKolisnyk:
## Type of change

- [ ] Bugfix
- [x] Feature
- [ ] New bidder adapter   
- [ ] Build related changes
- [ ] CI related changes

- [ ] Other

## Description of change
This PR represents updates to support Blockthrough Realtime Data Provider for Prebid.js. The module provides effective way for Publishers to recover their revenue in AdBlock traffic with respect to users preferences and not breaking Acceptable Ads Standard.

### DIFF
--- a/modules/btRecoveryRTDProvider.js
+++ b/modules/btRecoveryRTDProvider.js
@@ -1,0 +1,30 @@
+import { isStr, logError } from '../src/utils.js';
+import { submodule } from '../src/hook.js';
+import { loadExternalScript } from '../src/adloader.js';
+
+const MODULE_NAME = 'btRecovery';
+
+function init(config) {
+  const pubID = config.params && config.params.pubID;
+
+  if (!pubID || !isStr(pubID)) {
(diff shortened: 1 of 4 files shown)


modules/btRecoveryRTDProvider.js
```diff
@@ -0,0 +1,30 @@
...
+export const btRecoveryRtdModule = {
+  name: MODULE_NAME,
+  init
+};
+
+function registerSubModule() {
+  submodule('realTimeData', btRecoveryRtdModule);
+}
+
+registerSubModule();
```
patmmccann, line 30: It appears this is a pure tag management module with no other functionality. If so, I'm curious why you register as an rtd submodule?

patmmccann: @mkendall07 curious for your thoughts

mkendall07: Hey Patrick - we are reviewing internally. Going to close the PR for now. Thanks!

PR #9380, opened 2022-12-28
Yieldmo Adapter: Add support for structured user agent 

desidiver:
## Type of change

- [ ] Bugfix
- [x] Feature
- [ ] New bidder adapter   
- [ ] Build related changes
- [ ] CI related changes

- [ ] Other

## Description of change

Adding Support for Structured User Agent.

### DIFF
--- a/modules/yieldmoBidAdapter.js
+++ b/modules/yieldmoBidAdapter.js
@@ -11,7 +11,8 @@ import {
   isStr,
   logError,
   parseQueryStringParameters,
-  parseUrl
+  parseUrl,
+  pick
 } from '../src/utils.js';
 import {BANNER, VIDEO} from '../src/mediaTypes.js';
 import {registerBidder} from '../src/adapters/bidderFactory.js';
--- a/modules/yieldmoBidAdapter.js
+++ b/modules/yieldmoBidAdapter.js
@@ -522,7 +522,7 @@ function openRtbDevice(bidRequest) {
      sua = pick(uaClientHints, ['browsers', 'platform', 'mobile', 'model']);
    }
    if (sua) {
-     deepSetValue(deviceObj, 'sua', Object.assign({}, sua, deviceObj.sua));
+     deepSetValue(deviceObj, 'sua', sua);
    }
   return deviceObj;
 }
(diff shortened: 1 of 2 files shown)


desidiver:
@patmmccann Added unit test

modules/yieldmoBidAdapter.js
```diff
@@ -515,6 +516,15 @@ function openRtbDevice(bidRequest) {
     ua: navigator.userAgent,
     language: (navigator.language || navigator.browserLanguage || navigator.userLanguage || navigator.systemLanguage),
   };
+  // Add User Agent Client hints.
+  const uaClientHints = deepAccess(bidRequest, 'ortb2Imp.device.sua');
```
dgirardi, line 520: this is under `bidderRequest.ortb2.device.sua`. Also, why not just merge in everything from `bidderRequest.ortb2.device`, or even just the whole `ortb2` object?

desidiver: We don't want to include extra fields which are not going to be used.

dgirardi: The tradeoff is some unused fields - or extra code (and slower page loads) for the pub, and extra work for you on every new feature.

But beside that, `device` is not under `bidRequest.ortb2Imp` - it's under `bidderRequest.ortb2`. (Sorry for repeating myself, I'm not sure I made it clear enough).

desidiver: We are trying to figure out if there is any downside in passing entire ortb2 object but now I have made changes to pass entire device object. Sorry for not understanding it earlier, Thank you.

dgirardi: If you find a roadblock (beside not wanting to change things until you need to, which I do think is a legitimate reason), please let us know - since currently that's the direction we are recommending. 

I am aware of one case of an exchange that would not accept fields it didn't know about (https://github.com/prebid/Prebid.js/issues/9250) - the ORTB 2.6 spec requires endpoints to ignore such fields, so in the long term that should not be a problem (but it may be for you at the moment, so beware).

You would also give up the opportunity to do validation on publisher inputs, but I don't think each adapter doing independent validation is desirable - it's code duplication in the best case scenario, and in practice I think it would be a major source of confusion if the same configuration is interpreted differently by different exchanges.

PR #9383, opened 2022-12-29
Adloox RTD: fix breakage since 7.x release

jimdigriz:
## Type of change

- [x] Bugfix
- [ ] Feature
- [ ] New bidder adapter   
- [ ] Build related changes
- [ ] CI related changes

- [ ] Other

## Description of change
`getTargetingData` results in a crash since 7.x as 'auction' is no longer passed in and we needed access to the global ORTB2 targeting.

Reworked to be a lot simpler, and removed the ATF viewability segment and instead use the results of intersectionRtdProvider if present.

### DIFF
--- a/modules/adlooxRtdProvider.js
+++ b/modules/adlooxRtdProvider.js
@@ -6,11 +6,13 @@
  * @module modules/adlooxRtdProvider
  * @requires module:modules/realTimeData
  * @requires module:modules/adlooxAnalyticsAdapter
+ * @optional module:modules/intersectionRtdProvider
  */
 
 /* eslint standard/no-callback-literal: "off" */
 /* eslint prebid/validate-imports: "off" */
 
+import {auctionManager} from '../src/auctionManager.js';
 import {command as analyticsCommand, COMMAND} from './adlooxAnalyticsAdapter.js';
 import {submodule} from '../src/hook.js';
 import {ajax} from '../src/ajax.js';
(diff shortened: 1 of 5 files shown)


patmmccann:
@giardia @jimdigriz curious if you suspect any other adapters affected similarly?

jimdigriz:
@patmmccann I had a look and the only potential one I could find was https://github.com/prebid/Prebid.js/blob/0caca74e57de50daf23854cbeccdd57ecac7a621/modules/browsiRtdProvider.js#L350

modules/adlooxRtdProvider.js
```diff
@@ -213,193 +99,119 @@ function init(config, userConsent) {
...
-  const ortb2 = auction.getFPD().global || {};
-  const dataSite = deepAccess(ortb2, `site.ext.data.${MODULE_NAME}_rtd`) || {};
-  if (!dataSite.ok) return {};
+  const targeting = {};
 
-  const dataUser = deepAccess(ortb2, `user.ext.data.${MODULE_NAME}_rtd`) || {};
-  return getGlobal().adUnits.filter(adUnit => includes(adUnitArray, adUnit.code)).reduce((a, adUnit) => {
-    a[adUnit.code] = {};
+  const ortb2base = {};
+  _each(config._fpd, function(v0, k) {
```
dgirardi, line 184: I broke this way back when - sorry! 

The issue is the `auction` argument is the "public" auction object exposed through events and not the internal one that has `getFPD`. 

Unfortunately I don't think this fix is correct, because if multiple auctions are running concurrently, they would conflict on the shared `config._fpd` handle.

You can get access to `getFPD` with:

`const ortb2 = auctionManager.index.getAuction(auction).getFPD().global || {}`

Or alternatively you could modify `config._fpd`  to be a map from auctionId to to FPD.

dgirardi:
I went through the codebase and I think this is the only incidence, no other RTD provider needs to look up first party data at the time of `getTargetingData`.

modules/adlooxRtdProvider.js
jimdigriz: > I broke this way back when - sorry!

Happens to the best of us, no sweat.

> You can get access to getFPD with:
> 
> const ortb2 = auctionManager.index.getAuction(auction).getFPD().global || {}

I prefer to do this, but I opt'ed to create `config._fpd` as 'auction' no longer (or maybe never was?) officially [passed in according to the documentation](https://docs.prebid.org/dev-docs/add-rtd-submodule.html#the-sub-module-object)

As no other RTD module seemed to use it I assumed you may be trying to get rid of it?

dgirardi: I think that's the documentation getting out of sync - the argument was added in https://github.com/prebid/Prebid.js/pull/8119 and there are no plans to remove it as of now. 

In fact, digging a little more, I think the argument is necessary. Let me move down a few lines...

modules/adlooxRtdProvider.js
```diff
@@ -213,193 +99,119 @@ function init(config, userConsent) {
...
+    const v = val(v0);
+    if (v) ortb2base[`${ADSERVER_TARGETING_PREFIX}_${k}`] = v;
+  });
 
-    _each(dataSite, (v0, k) => {
-      if (includes(SEGMENT_HISTORIC_VALUES, k)) return; // ignore site average viewability
-      const v = targetingNormalise(v0);
-      if (v) a[adUnit.code][ADSERVER_TARGETING_PREFIX + k] = v;
-    });
+  const adUnits = getGlobal().adUnits.filter(unit => adUnitArray.includes(unit.code));
```
dgirardi, line 189: this should be `const adUnits = auction.adUnits`. There's again a race condition with using the global adUnits, since a publisher could do:

```
pbjs.addAdUnits(...)
pbjs.requestBids(...)
pbjs.removeAdUnit(...)
```

And by the time the auction ends and this code runs, you are working on the wrong ad units.

In fact I think it's likely a bunch of other adapters are doing this, since the original API gave them no other option. I'll take a look and probably open a separate issue.

dgirardi: I mean that by the time this code runs (at the end of the auction), `getGlobal().adUnits` might be completely different from what it was when the auction started - and your filter would suffer from the crap in, crap out problem. In fact, if the publisher does a completely legitimate

```
pbjs.requestBids({adUnits: [...]})
```

then `getGlobal().adUnits` would be empty the whole time. I think the RTD module API was defined a time when neither the `adUnits` argument nor concurrent auctions were supported, so it's a good thing we finally spotted this.

jimdigriz: The design of our module is not sensitive to if `getGlobal().adUnits` changes between the call to `getBidRequestData` and `getTargetingData` or before the call to `getBidRequestData` runs to completion.

For adUnit segments, within `getBidRequestData` we operate on our own array of `adUnit` objects cloned from the original list. If we set the FPD on *objects* that are no longer part of the original array (ie. deleted), it does not matter as the result will just be effectively discarded.

The site/user/device segments are queried once and always the same (per-referrer/device) and as such only those go into `config._fpd`. Poking our segment endpoint will return stable site/user/device results for the same page/device/user.

By the time we get to `getTargetingData`, we grab all adUnits and only do something when `AdSlot.ortb2Imp.ext.data.adloox_rtd` is present...otherwise we NOOP. So no harm if those slots we previously queried have now been deleted.

This is why I stash and try to extract the data from the adUnits themselves so we avoid storing state within our own module; which other RTD modules look to do.

The problem our RTD module does have though is it assumes `getRefererInfo` is a constant, which it seems not to be. This will need fixing, I'll do this whilst putting in back the use of `auction` (thanks for pointing how to access that safely!).

dgirardi: > So no harm if those slots we previously queried have now been deleted.

The issue is that from the pub's POV they have not deleted them; they have used them in an auction, presumably expecting the adloox targeting logic to work on them, and then moved them out of the way to start another auction (before the targeting logic has had a chance to run). So no harm, but also no functionality :) 

All of this is nitpicking, since refactoring as we discussed will fix the problem.

jimdigriz: @dgirardi okay, thank you for the help, pointers and advice here. I have reworked the module to use the `auction` object and try to steer clear of `getGlobal().adUnits`, let me know if things look better

dgirardi (approved):

jimdigriz:
Thanks, have a great week!

Issue #9391, opened 2023-01-03
Floors module: cannot handle complex bid adjustment functions


patmmccann:
## Type of issue
Feature Request or bug, not sure

## Description
https://github.com/prebid/Prebid.js/blame/master/modules/priceFloors.js#L194 seems to try to reverse engineer a bid adjustment function by plugging the floor into a bidder's bid adjustment function and seeing how far the result is below the input. This assumes adjustment functions can be reflected in the form bid = multiplier * adjusted bid

This method only works if the only adjustment is a multiplier. 

The solution appears to be: allow the publisher to set a inverse bid adjustment function and apply that instead of the psuedo-inverse calculated here when it exists. 

## Steps to reproduce
Set the bid  adjustment function to adj_cpm = max(0,0.95*bid_cpm - .05)
Set the floor to 0.01
Observe the outcome adjusted floor sent to endpoints, eg the rubicon fastlane endpoint on the floor query string parameter

### Expected results
adjusted floor = 6 cents/ 0.95

### Actual results
adjusted floor infinity

![image](https://user-images.githubusercontent.com/1683175/210412870-61046adb-d224-4214-ac50-95cf125ff5fd.png)

patmmccann:
![image](https://user-images.githubusercontent.com/1683175/210413764-1b19f3ad-4b65-4e99-b447-c731fed9f1f1.png)
Looks like I can come up with a scenario where the floor should be 20 cents but is calculated at 72 dollars

![image](https://user-images.githubusercontent.com/1683175/210413837-9800c03f-a046-4490-8d4d-e8778e112213.png)

robertrmartinez:
We decided to first go with a new `inverseBidAdjustment` function which can be set on bidderSettings just how `bidCpmAdjustment` is.

https://github.com/prebid/Prebid.js/pull/9395

Issue #9312, opened 2022-12-02
priceFloor does not pass bid to bidCpmAdjustment


renebaudisch:
## Type of issue
Object in parameter 2 passed to user-defined bidCpmAdjustment has insuffient parameters

## Description
E.G. bid.bidder ist therefore undefined, but should be accessible regarding the [doc](https://docs.prebid.org/dev-docs/publisher-api-reference/bidderSettings.html#22-bidcpmadjustment)

![image](https://user-images.githubusercontent.com/15980546/205307029-4ffc6faa-abd1-4dfe-96e8-d042dbd526a2.png)

`function getFloor` does not pass the bid as third parameter like `addBidResponseHook` does,
so it falls back to `{}` only holding the `cpm`.

## Steps to reproduce
to trigger `getFloor` and `bidCpmAdjustment` you need to configure a custom floor price schema and bidderSettings per bidder. Then just `console.log(arguments)` in your `bidCpmAdjustment` function.

## Test page
https://www.wieistmeineip.de/

### Expected results
get the hole bid data

### Actual results
only get property `cpm`

## Platform details
independent

patmmccann:
@robertrmartinez could you take a look?

dgirardi:
This is a tricky one. The issue is that the floors module attempts to adjust the _request_ floor, when there is no bid yet. I believe the idea is that if you set a given floor, but we know that some bidders will get their cpm adjusted, then the floor should be adjusted as well before it's sent out to SSPs, or you risk asking them to bid too low.

As a workaround, in your case specifically, you could use `bidder` from the outer scope instead of looking into the second argument.

renebaudisch:
But the problem only occurs by using custom floors schema, so the bidders need to call getFloor. If you use a standard schema different arguments will be passed. To be exact, then you get the data as described in the [documentation](https://docs.prebid.org/dev-docs/publisher-api-reference/bidderSettings.html#22-bidcpmadjustment)...

Looking at the example I should get bidCpm and bid, which is true for standard schema, but not for custom scheme.

dgirardi:
the adjustment is also calculated later on the actual bid cpm, to make sure it's satisfied - I believe that's why you may see the second parameter filled out sometimes. I don't think it's related to the schema unless you mean the `enforcement.bidAdjustment` flag. If not, could you provide steps to reproduce - what should I look for in the page you linked in the OP?

renebaudisch:
[...]

PR #9395, opened 2023-01-04
PBjs Core (Price Floors) : Support inverseBidAdjustment function

robertrmartinez:
## Type of change
- [X] Bugfix

## Description of change
Addresses #9391 

### Inverse Bid Adjustment
New input on bidderSettings called `inverseBidAdjustment`

```js
        pbjs.bidderSettings = {
          rubicon: {
            bidCpmAdjustment: function (bidCpm, bidResponse) {
              bidCpm *= 0.5;
              if (bidResponse.mediaType === 'video') bidCpm -= 0.18;
              return bidCpm;
            },
            inverseBidAdjustment: function (bidCpm, bidResponse) {
              if (bidResponse.mediaType === 'video') bidCpm += 0.18;
              return bidCpm / 0.5;
            },
          }
        };
```

The floors module will now look for this `inverseBidAdjustment` and use it before falling back to the old `calculateAdjustedFloor` logic.

### Fake Bid Response
There have been grumblings of when the getFloor function adjusts floors using `bidCpmAdjustment`, it calls the adjustment function without the bidResponse object.

Well, of course it does because it does not exist yet!

But, we can do some stuff to try and eliviate this by passing in the bidRequest object as well as a "fake" bid response.

Since the `getFloor` function takes in mediaType and sizes, and we know the `bidder` code, we should pass those into the bidResponse "object" part of the adjustment call so it helps at least mitigate some issues.

### DIFF
--- a/modules/priceFloors.js
+++ b/modules/priceFloors.js
@@ -179,10 +179,10 @@ function generatePossibleEnumerations(arrayOfFields, delimiter) {
 /**
  * @summary If a the input bidder has a registered cpmadjustment it returns the input CPM after being adjusted
  */
-export function getBiddersCpmAdjustment(bidderName, inputCpm, bid) {
+export function getBiddersCpmAdjustment(bidderName, inputCpm, bid, bidRequest) {
   const adjustmentFunction = bidderSettings.get(bidderName, 'bidCpmAdjustment');
   if (adjustmentFunction) {
-    return parseFloat(adjustmentFunction(inputCpm, {...bid, cpm: inputCpm}));
+    return parseFloat(adjustmentFunction(inputCpm, { ...bid, cpm: inputCpm }, bidRequest));
   }
   return parseFloat(inputCpm);
 }
@@ -250,9 +250,8 @@ export function getFloor(requestParams = {currency: 'USD', mediaType: '*', size:
   // if cpmAdjustment flag is true and we have a valid floor then run the adjustment on it
   if (floorData.enforcement.bidAdjustment && floorInfo.matchingFloor) {
     // For cpm adjustments, they may use the "bid response" object now. So we should put as much info in it as we can
-    // But since this is at bidRequest time, we may not have everything! So put entire bidRequest and overwrite a couple things to be safe
     const fakeBid = {
-      ...bidRequest,
+      bidder: bidRequest.bidder,
       mediaType: requestParams.mediaType, // default would be `*` but it is whatever the bidder adds to getFloor
       size: requestParams.size // similaraliy, this may be * or an actual size, better than nothing!
     }
@@ -260,9 +259,9 @@ export function getFloor(requestParams = {currency: 'USD', mediaType: '*', size:
     // pub provided inverse function takes precedence, otherwise do old adjustment stuff
     const inverseFunction = bidderSettings.get(bidRequest.bidder, 'inverseBidAdjustment');
     if (inverseFunction) {
-      floorInfo.matchingFloor = inverseFunction(floorInfo.matchingFloor, fakeBid);
+      floorInfo.matchingFloor = inverseFunction(floorInfo.matchingFloor, fakeBid, bidRequest);
     } else {
-      let cpmAdjustment = getBiddersCpmAdjustment(bidRequest.bidder, floorInfo.matchingFloor, fakeBid, floorData.inverseFloorAdjustment);
+      let cpmAdjustment = getBiddersCpmAdjustment(bidRequest.bidder, floorInfo.matchingFloor, fakeBid, bidRequest);
       floorInfo.matchingFloor = cpmAdjustment ? calculateAdjustedFloor(floorInfo.matchingFloor, cpmAdjustment) : floorInfo.matchingFloor;
     }
   }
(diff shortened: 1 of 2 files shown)


robertrmartinez:
@dgirardi 

Would like your input here.

We had a discussion about the adjustment functions and how we cannot pass in bidResponse during getFloor. And how we should update to pass in bidRequest as well to help some scenarios.

What do you think about this fakeBid thing? At least passing in some information we do know to the adjustment functions?

Maybe just passing in an object with

```
{ bidder, mediaType, size, cpm }
```

as the "fake" bid is better?

Also, I do not think we did the work yet to pass in bidReq object as an option into the normal bidCpmAdjustment yet have we?

Thanks!

modules/priceFloors.js
```diff
@@ -249,8 +249,22 @@ export function getFloor(requestParams = {currency: 'USD', mediaType: '*', size:
 
   // if cpmAdjustment flag is true and we have a valid floor then run the adjustment on it
   if (floorData.enforcement.bidAdjustment && floorInfo.matchingFloor) {
-    let cpmAdjustment = getBiddersCpmAdjustment(bidRequest.bidder, floorInfo.matchingFloor);
-    floorInfo.matchingFloor = cpmAdjustment ? calculateAdjustedFloor(floorInfo.matchingFloor, cpmAdjustment) : floorInfo.matchingFloor;
+    // For cpm adjustments, they may use the "bid response" object now. So we should put as much info in it as we can
+    // But since this is at bidRequest time, we may not have everything! So put entire bidRequest and overwrite a couple things to be safe
+    const fakeBid = {
```
dgirardi, line 254: I think for the inverse it's better to just pass in the bidRequest no? since we never have a bid we don't need to fake it. This would also be the first exposure to the internal `getFloor` params we'd give to the publisher, I'm not sure that they'd help more than the request.

robertrmartinez: well in the case that the bidRequest has both banner and video

But the bidder is asking for floor specific to banner or video,

I think it is best to expose that information to them right?

Same with sizes, if there are 5 sizes, and a bidder loops through all and asks for the specific floor for that size I guess?

But yeah only makes sense if the adjustment / inverse function relys on the attribute.

Which I think bidder code, and mediaType are definitely used in adjustment functions.

dgirardi: I'm unconvinced, I think most bidders ask for star, so you still can't provide the same info you'd have in the response - and I think we are more likely to get confusion from `mediaType = '*'` than precision from when it is specified. I would be OK with passing them in separately such as `(bidRequest, floorRequestParams)`, except that the latter is still an internal API and I'd prefer to not lock that in prematurely.

robertrmartinez: ok one last ask for input from @patmmccann 

In your scenario where your regular bid adjustment function only subtracts `0.18` if the mediaType is video,

In the inverse function, if we pass you the bidRequest, is that enough?

The only way it wouldn't would be if bidRequest.mediaTypes had both banner and video right?

Also @dgirardi My thought was that the input params to the `bidCpmAdjustment` function and the `inverseBidAdjustment` functions should be aligned so it is easy for pubs to utilize them?

Or you think it is ok to just have them different.

I agree regarding exposing the `floorRequestParams` fwiw

patmmccann: Yeah seems right to me

mmoschovas (approved):
lgtm pending the update requested

robertrmartinez:
Ok I pushed the change to pass bidRequest object to the adjustment functions.

BUT I think there is still some disconnect between us.

@patmmccann for your use case in the inverseAdjustment function, you need to know the mediaType right?

And you cannot 100% know the mediaType by just looking at the bidRequest object, since an adUnit may be multiFormat

```
bidRequest.mediaTypes = {
   banner: ...
   video: ...
}
```

So, your inverse function would have to make some decision here.

Which is why I thought the `fakeBid` thing where the getFloor function passes in a contextualized "bidResponse" object into the adjustment function taking what it can from the call to getFloor, like mediaType requested and size requested.

I get this is strange, but right now the adjustment functions are not getting anything from getFloor, so it probably does not hurt to add some stuff right?

patmmccann:
I see the disconnect, I think using media type in the floor inverse is sufficiently rare by adapters (they all just pull *) and adds enough complexity that we're likely wise to not add the complexity. Seem reasonable? We can always add it back later?

robertrmartinez:
yeah ok, so what should we pass to the inverseFunction?

dgirardi:
I vote for 

> `inverseFunction(cpm, bidRequest);`

robertrmartinez:
Thy will shall be done!

Got rid of the fakeBid stuff, for the usual bidder cpm adjustment calculation I just changed it to pass empty object there, but also pass the bidRequest object so maybe that can be used in a smart way by some pubs if needed?

Thinking maybe we want to update the other placed cpmAdjustment is used as well to pass bidRequest?

dgirardi:
@robertrmartinez I'd do that separately, in fact I plan to take that on for https://github.com/prebid/Prebid.js/issues/9312, but it's easier if we merge this in first.

patmmccann (approved):

PR #9413, opened 2023-01-11
Sonobi Bid Adapter: add additional sizes to bid request

About7Sharks:
## Type of change

## Description of change

`_validateSize()` takes in a single parameter, "bid". The function takes all the possible sizes from the bid object and returns them as a string in the csv format. It accesses the different properties of the bid object to find the sizes, then parse it and throw it into a Set to uniquify it. Then it spreads it to an array and joins it into a csv of sizes.
The main purpose of this function is to check for any possible size fields within the bid object, and to parse and format that information into a single, consistent string in csv format which can be easily used by the rest of the program.


integrationExamples/gpt/hello_world.html
```diff
@@ -1,91 +0,0 @@
-<!--
```
patmmccann, line 1: please restore

patmmccann (changes requested):

integrationExamples/gpt/hello_world.html
Rothalack: Hey @About7Sharks , this file needs to be restored, not deleted.

Rothalack: I just realized that you did actually add it back, it's the files changed view on this PR that's misleading. I'll mark this resolved again.

About7Sharks:
Awesome thanks! Apologies about that, I usually use `git add .` but when  testing using the `hello_world.html` It added those changes which I fudged the `git rm` command leading to the untracked file.

integrationExamples/gpt/hello_world.html
About7Sharks: Awesome thanks! Apologies about that, I usually use `git add .` but when  testing using the `hello_world.html` It added those changes which I fudged the `git rm` command leading to the untracked file.

Rothalack (approved):
Working in testing for me, should be good to go.

patmmccann (approved):

PR #9418, opened 2023-01-12
Lotame Panorama ID Module : add safari handling

Tonsil:
## Type of change

- [ ] Bugfix
- [x] Feature
- [ ] New bidder adapter   
- [ ] Build related changes
- [ ] CI related changes

- [ ] Other

## Description of change
This change is to appease requests from any browser, including Safari and ITP browsers that does not permit setting third party cookies on designated URLs. We will now direct all such traffic to a new domain specifically tailored to avoid such cookies entirely.

### DIFF
--- a/modules/lotamePanoramaIdSystem.js
+++ b/modules/lotamePanoramaIdSystem.js
@@ -29,7 +29,7 @@ const DAY_MS = 60 * 60 * 24 * 1000;
 const MISSING_CORE_CONSENT = 111;
 const GVLID = 95;
 const ID_HOST = 'id.crwdcntrl.net';
-const SAFARI_ID_HOST = 'aWQubHRtc3BocmNsLm5ldA==';
+const SAFARI_ID_HOST = 'id.ltmsphrcl.net';
 
 export const storage = getStorageManager({gvlid: GVLID, moduleName: MODULE_NAME});
 let cookieDomain;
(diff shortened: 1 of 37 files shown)


modules/lotamePanoramaIdSystem.js
```diff
@@ -252,6 +253,13 @@ export const lotamePanoramaIdSubmodule = {
       usPrivacy = getFromStorage('us_privacy');
     }
 
+    const getRequestHost = function() {
+      if (navigator.userAgent && navigator.userAgent.indexOf('Safari') != -1 && navigator.userAgent.indexOf('Chrome') == -1) {
```
patmmccann, line 257: what about this https://chromium.googlesource.com/chromium/src/+/refs/heads/main/docs/ios/user_agent.md

modules/lotamePanoramaIdSystem.js
```diff
@@ -29,6 +29,7 @@ const DAY_MS = 60 * 60 * 24 * 1000;
 const MISSING_CORE_CONSENT = 111;
 const GVLID = 95;
 const ID_HOST = 'id.crwdcntrl.net';
+const SAFARI_ID_HOST = 'c.ltmsphrcl.net';
```
patmmccann, line 32: is this a cat and mouse game, are you rotating the domain everytime you get blacklisted?

Tonsil: This change is to isolate requests that may try to set 3rd party cookies under certain conditions from those that will not even attempt to do so, and we do not anticipate further changes to either domain. To clarify that intent, we've changed the variable name to ID_HOST_COOKIELESS

modules/lotamePanoramaIdSystem.js
Tonsil: Thanks for calling that out, but the reality is that any ITP browsers should be using the cookieless domain. This includes Chrome and Firefox on iOS. The user agent detection as written will cover both cases.

patmmccann (approved):

PR #9419, opened 2023-01-12
Sonobi Bid Adapter : now passing the entire mediaType.video object value as a new query param

About7Sharks:
## Type of change

- [x] Feature

## Description of change

Now passing the entire `mediaType.video` object value as a new query param to trinity.json called “video_options”. The expected value would be a JSON string 

## Other information

Screen Shot 2023-01-12 at 3 13 58 PM


patmmccann:
please name your pr correctly

PR #9420, opened 2023-01-13
Permutive RTD Module: adds a safety check to prevent errors in the case a nested property doesn't exist

zshnr:
## Type of change

- [x] Bugfix

## Description of change

Adds a safety check to ensure a nested property exists inside an object before we access any properties on the nested property.

This was sometimes throwing as pointed out in the discussion here: https://github.com/prebid/Prebid.js/pull/9236#discussion_r1066043015

## Other information
Relates to the work done in #9236


ChrisHuie (approved):

zshnr:
Thank you so much for the quick approve @ChrisHuie !

PR #9426, opened 2023-01-15
PBjs Core (Promises): fix static method GreedyPromise.resolve not working with Angular + Zone.js

### LINKED ISSUE #9422: Unfulfilled promise promise prevents bidding with Angular + Zone.js
I found out that in the current PrebidJS version (at the time of writing this, v7.31.0), and in combination with Angular + Zone.js, this particular line produces a pending promise that never gets fulfilled nor rejected, preventing execution from continuing and therefore not bidding:

https://github.com/prebid/Prebid.js/blob/49bf03728146399c1859f31f24368b1de09208ee/src/prebid.js#L643

Just changing `GreedyPromise.resolve` with `Promise.resolve` fixes the issue for me, and everything works just as it should.

I have not been able to find the source of the problem, but I guess it must be related to the Promise-related symbols monkey patched by Zone.js.

**Reproduction of the issue**: https://stackblitz.com/edit/typescript-ibmjdh

javiermarinros:
Fix #9422


dgirardi:
LGTM except for the linting errors.

javiermarinros:
Fixed the linting error @dgirardi

dgirardi (approved):

patmmccann (approved):

PR #9432, opened 2023-01-17
 hadronId user id submodule: force localStorage

jlaso:
## Type of change

- [ ] Bugfix
- [X] Feature
- [ ] New bidder adapter   
- [ ] Build related changes
- [ ] CI related changes

- [ ] Other

## Description of change

In order to give visitors a better user experience we have decided to store hadronId in localStorage in order to improve the timing for next website visits. Using that occasion we refactored and cleaned the code a little bit.

### DIFF
--- a/modules/hadronIdSystem.js
+++ b/modules/hadronIdSystem.js
@@ -8,8 +8,9 @@
 import {ajax} from '../src/ajax.js';
 import {getStorageManager} from '../src/storageManager.js';
 import {submodule} from '../src/hook.js';
-import { isFn, isStr, isPlainObject, logError } from '../src/utils.js';
+import {isFn, isStr, isPlainObject, logError, logInfo} from '../src/utils.js';
 
+const HADRONID_LOCAL_NAME = 'auHadronId';
 const MODULE_NAME = 'hadronId';
 const AU_GVLID = 561;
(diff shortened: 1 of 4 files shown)


patmmccann:
you're just deleting publisher choice? This appears to be an antifeature?

jlaso:
> you're just deleting publisher choice? This appears to be an antifeature?

I'm sorry, what do you mean by "publisher choice"?

patmmccann:
Publishers currently choose cookie or local storage. You appear to be deleting the configuration opportunity, likely in a way that breaks people who have chosen the former.

jlaso:
Yeah, you're absolutely right looking only at the changes I made, but if you look at the full code you will see that we haven't implemented the cookie system nor the possibility to change the name of the stored value in localStorage.
So, what I am doing in this PR is adapt the documentation to the actual status of the code.
In other words, no matter what the publisher selects in the config we are not using it for storing the hadronId. 
But, if you consider that's not a good practice I will be more than happy to change whatever is needed to follow the rules.

patmmccann:
just to confirm: if someone has the cookie configured and they upgrade to whichever 7.x version this is merged to, they wouldnt need to fix anything for hadron to keep working?

jlaso:
You are right, if someone has cookie configured gonna have exactly the same behaviour with this update, because we were always looking at the backend to get the ID and with cookie configured we will still looking at backend.

patmmccann (approved):

Issue #8539, opened 2022-06-08
Prebid 8 tracking issue


patmmccann:
List of issues or ideas for prebid 8, which unless an urgent need arises, is expected in q2 2023. Work expected to begin in q1 2023, this issue is just for organization. 

Delete legacy aol adapter
Remove duplicated of module md files from js repo? 
Delete legacy pubcommonid adapter
Targeting overhaul
Bidders should read more stuff from openrtb (tbd what)
Bidders should pass a tid back home? 
Finish moving any special bidder handling in rtd modules into the bidder module

patmmccann:
potentilly deprecate legacy native methods

patmmccann:
get rid of GA analytics adapter?

patmmccann:
https://github.com/prebid/Prebid.js/issues/8665#issuecomment-1187638257

patmmccann:
https://github.com/prebid/Prebid.js/issues/8779

patmmccann:
#8795

patmmccann:
consolidate size mapping logic?

patmmccann:
enforce https://github.com/prebid/Prebid.js/issues/8421

patmmccann:
check in on https://github.com/prebid/Prebid.js/pull/8945

patmmccann:
potentially deprecate currency file on s3

patmmccann:
potentially deprecate dual impression level floorMin location

patmmccann:
re #8956 delineate no_bid from rejected bid

patmmccann:
fix https://github.com/prebid/Prebid.js/issues/9047 in fluxt and 33across

patmmccann:
Enforce #7775

patmmccann:
https://github.com/prebid/Prebid.js/issues/9333#issuecomment-1349287552 we might want to break this behavior

patmmccann:
#9333 is relevant

patmmccann:
Closing this issue in favor of better organized issue

PR #9433, opened 2023-01-18
cpexId Id Module: Renamed to czechAdId [vendor confirms no use yet, non-breaking]

EskelCz:
## Type of change
- [X] Other

## Description of change
Simply renaming of the module. The new website (linked in readme) is about to go live in a few days.

Documentation PR: https://github.com/prebid/prebid.github.io/pull/4286


patmmccann:
appears to be breaking, tagging as #8539

EskelCz:
@patmmccann Sorry I don't understand, it's not passing tests of some future version? What should I do?

bretg:
@EskelCz  - the issue is that Prebid does not allow renaming of modules except when there's a major release. That would cause problems for publishers using the old name. We guarantee to pubs that they don't have to change their in-page or build procedures except at major releases.

So in your case, one approach would be be add an "alias" for your user ID module so that both the old name and the new name work. @dgirardi can help you determine how best to do that.

modules/czechAdIdSystem.js
```diff
@@ -1,27 +1,27 @@
...
-export const storage = getStorageManager({ gvlid: 570, moduleName: 'cpexId' })
+export const storage = getStorageManager({ gvlid: 570, moduleName: 'czechAdId' })
 
 // Returns the id string from either cookie or localstorage
 const readId = () => { return storage.getCookie('czaid') || storage.getDataFromLocalStorage('czaid') }
 
 /** @type {Submodule} */
-export const cpexIdSubmodule = {
-  version: '0.0.5',
+export const czechAdIdSubmodule = {
```
dgirardi, line 18: my strategy would be to export this as a function taking in the module name, and have a "dummy" `cpexIdSystyem.js` that re-registers under the old name until version 8. alternatively, you could delay this change until v8.

EskelCz:
@bretg I see. The thing is, our ID module is in testing phase and not yet fully public. In order for publishers to use it, they must be connected to the Czech Ad ID service - meaning they have to contact us, enter into a contract with us and go through some technical integration. As a result, we know exactly who uses this ID. Therefore this renaming will also be coordinated with everyone involved. Does the major version rule still apply, or can this be excluded, as a special case? Thanks.

bretg:
Will let @patmmccann and the committee decide if we can make an exception

EskelCz:
@patmmccann So, any verdict on this? :) Thanks

patmmccann:
Can you provide a list of existing users?

EskelCz:
@patmmccann According to the project coordinator, at the moment the only partner using the module is Czech News Center a.s. We're in close contact with them and the system is not yet in production.

patmmccann:
okay

patmmccann (approved):

Issue #9248, opened 2022-11-15
PBJS support for enhanced PBS analytics


bretg:
## Type of issue
enhancement

## Description
Client-side analytics adapters don't have access to what actually happened with PBS bids unless they're successful. The Prebid Server team is working on [an enhancement](https://github.com/prebid/prebid-server/issues/2367) that implements a new optional request flag that instructs PBS to include the status of all bid attempts in the response.

Publishers would opt-in to the additional analytics data because it could increase the size of the payload from Prebid Server by a few percent. 

They would opt-in by setting up the appropriate s2sConfig:
```
pbjs.setConfig({
  s2sConfig: [{
    ...
    extPrebid: {
      returnallbidstatus: true
    }
  }]
});
```

When this flag is true, PBS will return the additional `ext.seatnonbid` JSON as described in the [PBS issue](https://github.com/prebid/prebid-server/issues/2367).

Note that the IAB is discussing this proposed `seatnonbid` syntax, so it could possibly change somewhat in the coming months, though we don't expect any significant changes.

## Prebid.js Work

1) The pbsBidAdapter (or the new ORTB library?) needs to be able to read `ext.seatnonbid` and create the corresponding Prebid.js bid response data structures. e.g. no-bid, timeout, error. These structures should carry the detailed reason status code. Note that it should enforce allowUnknownAdapters as it would for seatbids.
    - statuscodes 1-99: No-Bid
    - statuscodes 100-199: Error
    - statuscodes 200-299: Request Blocked
    - statuscodes 300-399: Response Rejected
2) Analytics adapters should be able to read this enhanced server-side response data, including the more detailed status codes.
3) Since the IAB is discussing this extension as a standard, we should consider providing a mechanism for client-side bid adapters to report the same level of status code reporting. i.e. a PBJS bid adapter could report seatnonbid statuscode 8, which is no-bid due to unmatched user.

## Test Cases

1) PBS returns a mix of seatbid and seatnonbid

```
{
	"seatbid": [{
		"seat": "bidderB",
		"bid": [{
			"id": "613673EF-A07C-4486-8EE9-3FC71A7DC73D",
			"impid": "adunit-code-2",
			"price": 1.00,
			"adm": "<html>Hello World</html>",
			"adomain": [
				"www.addomain.com"
			],
			"ext": {
				"prebid": {
					"meta": {
						"adaptercode": "bidderB"
					}
				}
			}
		}]
	}],
	"ext": {
		"seatnonbid": [{
[...]

Issue #8629, opened 2022-06-30
Feature Request: Send bidRequested event when unknown Bidder responses occur


muuki88:
## Type of issue: feature request

Sent `bidRequest` events for unknown bidder codes returned by prebid server so analytics adapters like Pubstack keep working on client-side.

## Description

#8154 introduced the `allowUnknownBidderCodes`´setting, which makes the workaround described in #7027 obsolete. However this initial implementation comes [with some drawbacks](https://docs.prebid.org/dev-docs/modules/prebidServer.html#allowUnknownBidderCodes) most importantly that there is no `bidRequest` event sent.

This makes sense as there's no information on the client side which bidders may have been requested on the server side. Maybe there is a possibility to sent those events at a later point. The wokraround from #7027 is sufficient enough to make the our analytics adapter function properly again. 

Two options could be

- sent `bidRequest` events for all bidders defined in the `s2sCfonfig.bidders` array
- prebid server respondes with a list of bidders that have been requested for each `imp`, wich the will be turned into `bidRequest` events

## Platform details

Prebid 6.x & 7.x

## Other information

- [prebidServer docs](https://docs.prebid.org/dev-docs/modules/prebidServer.html)

dgirardi:
I think I disagree - generating fake bidRequest events might work for some scenarios, but it's bound to break others. After all the bid was _not_ requested. To me this sounds like your analytics should be looking at bidResponse instead.

A better workaround, IMO, would be a mechanism that generate fake events for particular analytics adapters only, maybe as a flag in `enableAnalytics`.

muuki88:
Thanks @dgirardi . I agree with your statement. IMHO this should be handled via a server side analytics adapter. With #7585 this should be possible through the `auctionId` being available server and client side.

I opened this issue as server side analytics is still in a very early stage and server side adoption is slowly growing. We still need to configure our ad tags if we want to add new adapters server-side, which is one huge benefits of doing server-side at all.

gpolaert:
Hello @muuki88 and @dgirardi 

Thanks for raising this issue. 

I completely agree with the fact that the adapters (and the Pubstack one) need to handle the new option `allowUnknownBidderCodes` and stored impressions without generating fake bidRequested events. 

@muuki88 I still have no ETA regarding the support of this issue in Pubstack, I will keep you informed as soon as I have any news

patmmccann:
[...]

PR #9453, opened 2023-01-23
Prebid Core: emit seatnonbid from prebid server

spotxslagle:
## Type of change

- [x ] Feature

- [ ] Other

## Description of change

Support seatnonbid events in prebid server

Solves https://github.com/prebid/Prebid.js/issues/9248


modules/prebidServerBidAdapter/index.js
```diff
@@ -543,6 +545,17 @@ export const processPBSRequest = hook('sync', function (s2sBidRequest, bidReques
   }
 }, 'processPBSRequest');
 
+function emitNoBid(bid) {
+  const status = bid.nonbid.statuscode;
```
harpere, line 549: should check that everything is defined 
 `bid && bid.nonbid && bid.nonbid.statuscode`

modules/prebidServerBidAdapter/index.js
```diff
@@ -543,6 +545,17 @@ export const processPBSRequest = hook('sync', function (s2sBidRequest, bidReques
...
+  if (status > 0 && status < 100) {
+    events.emit(CONSTANTS.EVENTS.NO_BID, bid);
+  } else if (status > 99 && status < 400) {
+    events.emit(CONSTANTS.EVENTS.BIDDER_ERROR, bid);
+  } else if (status === undefined) {
+    logError('No bid not emitted. No status defined');
+  }
```
harpere, line 556: Do we not want to use the [BID_REJECTED](https://github.com/prebid/Prebid.js/blob/master/src/constants.json#L32) event for any of the "REQUEST_BLOCKED" (200-299) or "RESPONSE_REJECTED" (300-400) statuses from the [Status Code Guidance](https://github.com/prebid/prebid-server/issues/2367#issuecomment-1299097069)?   Or is the "BIDDER_ERROR" event enough?  Or maybe add another "BID_BLOCKED" event to distinguish between **blocked** and **rejected** bids?

harpere (commented):
A couple of questions/comments, plus we should add unit tests.

modules/prebidServerBidAdapter/index.js
harpere: Also, I think we need to update the Magnite Analytics Adapter to handle these events (may require more discussion)

spotxslagle: I think we're planning to handle the analytics in a separate PR.  I'm not sure who to ask about which event to emit. We don't need to worry too much about stepping on other adapters' toes since anyone hearing this event has to opt in in their s2s conf.

libraries/ortbConverter/converter.js
```diff
@@ -127,7 +127,13 @@ export function ortbConverter({
           logError('ORTB response seatbid[].bid[].impid does not match any imp in request; ignoring bid', bid);
         })
       ).filter(Boolean);
-      return buildResponse(bidResponses, response, augmentContext(ctx.req));
+      const bidNonResposes = ((response.ext && response.ext.seatnonbid) || []).flatMap(seatnonbid =>
```
dgirardi, line 130: I don't like this; nonbids are not bids, I don't think they should be mixed together. It also requires special handling from every adapter that happens to talk to a PBS instance (like Rubicon).

I think this should be a new "response processor", part of the PBS processors, like [this one](https://github.com/prebid/Prebid.js/blob/master/libraries/pbsExtensions/processors/pbs.js#L88). This way, every adapter that wants the PBS-specific dialect (including PBS and rubicon) would get this logic without any changes.

If you want/need to map these back to the original imp you'll need some changes in here that are also already in https://github.com/prebid/Prebid.js/pull/9342 - it might be easier to merge that in first.

modules/prebidServerBidAdapter/index.js
spotxslagle: This duplicates the check that happens when this function is called. Do we want a duplicate check?

modules/prebidServerBidAdapter/index.js
patmmccann: @spotxslagle can you mark this second pr as solving #9248?

libraries/ortbConverter/converter.js
spotxslagle: I agree with this. I think the new revision is much cleaner. I don't have a super clear picture of what exact data we want to be emitting that might be missing from the server response, so I didn't mess with response processing. But if there is a lot of data we'd want to tack on, I agree adding it in a response processor would be be good.

modules/prebidServerBidAdapter/index.js
```diff
@@ -543,6 +546,17 @@ export const processPBSRequest = hook('sync', function (s2sBidRequest, bidReques
   }
 }, 'processPBSRequest');
 
+function shouldEmitNonbids(s2sConfig, response) {
+  return s2sConfig.extPrebid && s2sConfig.extPrebid.seatnonbid && response && response.ext && response.ext.seatnonbid;
```
robertrmartinez, line 550: I think it should be

`s2sConfig.extPrebid.returnallbidstatus`

So maybe

```
return s2sConfig?.extPrebid?.returnallbidstatus && response?.ext?.seatnonbid
```

robertrmartinez (approved):
I think this is a good first step.

PBS has some work to do to handle making the returned objects more interesting, and I think when that gets solidified we can enhance PBJS part to do some fancier stuff

As of now, analytics adapters can listen to the new SEAT_NON_BID event and do with it what they want.

But perhaps after PBS makes it changes, we can have PBS Bid Adapter break up the non bids into the appropriate events like nobid / timeout / error etc?

@dgirardi what do you think?

dgirardi:
(It looks like one of the new tests is failing)

I am not thrilled about this firing a new event if its express intent is to surface the existing ones... we are asking analytics adapters to do the translation themselves, only to make that work useless at some point in the future.

But I'm not sure how we would go about firing the "correct" events: if for example I involve a bidder both client and server side, would we want to get both a bid from one and a `NO_BID`  from the other? The same can happen when using multiple PBS instances. 

Some events might also be really hard to reproduce with their current interface. One of the arguments to `BIDDER_ERROR` is an instance of `XMLHttpRequest`.

robertrmartinez:
> would we want to get both a bid from one and a NO_BID from the other

I would think so, yes. We want to see each bid really is the thing right.

So if you have one client side bidder who has 10 bidRequests, and 
5 of them are no-bid
3 are error
2 are bid responses

Then you see 10 events if I am not mistaken, 5 nobids, 3 errors, 2 bidResponses

The problem we have here is that Prebid has bidRequests for all of these 'known' bids. 

So analytic adapters when listening can see 10 bid requests. 

So most start their own internal mapping of these bidRequests, and when they see a no bid, they go and update their map, if they see a bid Response, they up their map and so on.

Now, if an event comes from a bidder / bid who was not in those original mapping of bid requests, the analytic adapter has to know what to do. Which I am sure the logic to handle or not handle this is different per analytic adapter.

Really, it does not make sense for prebid to see a nonbid from PBS, and then say "ok wait I need to emit a bunch of bidRequested events real quick, and then the associated no bid or error bid right?

I think there is not some perfect solution here to solve all event listeners cases.

Which is why I think a new event that is documented and is its own case makes the most sense. And analytic adapters can choose to do with it what they want.

The "Right" approach to all of this probably requires a restructuring to how prebid handles bidRequests mapping to bidResponses itself and all of that. Which seems like too big of a task to take on for this.

dgirardi:
> And I think that if all of a sudden Prebid started to emit a bunch of no bids and errors it would probably break a lot of event listeners logic. "These nobids do not match any bidRequests help"

This would happen only if you enable `allowUnknownBidderCodes` though right? Normally PBS would not get errors and such from bidders for which there is no bidRequest. And I agree it's a problem, but we already [cannot avoid it](https://github.com/prebid/Prebid.js/issues/8629) since we fire `BID_RESPONSE`.

I'd expect the mixing of seemingly incompatible events (like getting both a bid and a no bid) to be more of a problem - because I can imagine publishers that rely on getting a single one for their business logic. Analytics adapters would be much easier to fix in comparison.

robertrmartinez:
>This would happen only if you enable allowUnknownBidderCodes though right

Well, you could have the same bidder client and also server side. So a no bid would emit that does match a bidderCode, but not a prebid bidRequest bidId (requestId)

That [issue](https://github.com/prebid/Prebid.js/issues/8629) has people mostly agreeing that generating fake events using the existing events seems like the wrong thing.

Especially when adapters may expect an order of operations for events to occur for a given auction.

I just think that these new scenarios are tough to fit into the existing event system without breaking stuff.

I agree we should come up with a more elegant approach though. Perhaps that requires some of us getting on a call and outlining all of our thoughts, use cases and ideas.

But I do still think that any new optional event like this one does not break anything and it at least somewhat solves the problem until we come up with something better. Which better likely means breaking change in 8 or 9

dgirardi:
> Well, you could have the same bidder client and also server side. So a no bid would emit that does match a bidderCode, but not a prebid bidRequest bidId (requestId)

I think you would have a matching bidId, since in this scenario core would generate two separate bidRequests for client and server. 

> But I do still think that any new optional event like this one does not break anything and it at least somewhat solves the problem until we come up with something better.

I agree except that events are also part of the publisher API, so they are not a great tool to use for "temporary" stuff. But I don't have a good alternative (maybe leave it undocumented? hey, I did said it wasn't good). Overall I think this is OK.

robertrmartinez:
> Well, you could have the same bidder client and also server side. So a no bid would emit that does match a bidderCode,> but not a prebid bidRequest bidId (requestId)
> 
> I think you would have a matching bidId, since in this scenario core would generate two separate bidRequests for client and server.

I mean in a storedRequest scenario where you 

- have `rubicon` client side
- have some stored request for PBS which has rubicon server side as well

In this case, the server side rubicon bid would not have a matching bidId in prebid right.

So you would have a `server` rubicon bid coming from PBS stored request

and a `client` rubicon bid from normal prebid bidding

> I agree except that events are also part of the publisher API, so they are not a great tool to use for "temporary" stuff. But I don't have a good alternative (maybe leave it undocumented? hey, I did say it wasn't good). Overall I think this is OK.

Right haha, the reason being that if a pub starts using the new event and all of a sudden we change or remove it they get angry! Totally get it.

I am going to ask @bretg to chime in a bit here and see how he feels. But yeah, the perfect solution is hard to find.

bretg:
I'm going to avoid getting involved in the technical architecture of PBJS core.

Here are the requirements:

1. Analytics must have access to the range of status codes supplied by Prebid Server in seatnonbid.
2. Prebid Server should not return both a bid and a nonbid on the same imp
3. If the same imp was auctioned both client-side and server-side, there might be multiple statuses. Analytics adapters need to be able to deal with that

dgirardi (approved):

PR #9457, opened 2023-01-24
AIDEM Bidder Adapter: changed required params and win notice payload

darkstarac:
## Type of change
- [x] Bugfix
- [x] Feature

## Description of change

- fix wrong key retrieve from interpretResponse
- placementId is now a required parameter
- added optional rateLimit parameter
- added publisherId, siteId, placementId in win notice payload 
- added unit tests


modules/aidemBidAdapter.js
```diff
@@ -406,10 +412,26 @@ function hasValidVideoParameters(bidRequest) {
...
 }
 
+function passesRateLimit(bidRequest) {
+  const rateLimit = deepAccess(bidRequest, 'params.rateLimit', 1);
+  if (!isNumber(rateLimit) || rateLimit > 1 || rateLimit < 0) {
+    logError('AIDEM Bid Adapter: invalid rateLimit (must be a number between 0 and 1)', { bidder: BIDDER_CODE, code: ERROR_CODES.INVALID_RATELIMIT });
+    return false
+  }
+  if (rateLimit !== 1) {
+    const randomRateValue = Math.random()
```
ChrisHuie, line 422: Why is this a random generated value?

modules/aidemBidAdapter.js
```diff
@@ -149,7 +151,11 @@ function getPageUrl(bidderRequest) {
 }
 
 function buildWinNotice(bid) {
+  const params = bid.params[0]
```
patmmccann, line 154: there's a bunch of missing semicolons in this file, please fix

modules/aidemBidAdapter.js
```diff
@@ -421,6 +443,11 @@ function hasValidParameters(bidRequest) {
     return false;
   }
 
+  if (!isStr(placementId)) {
+    logError('AIDEM Bid Adapter: placementId must valid string', { bidder: BIDDER_CODE, code: ERROR_CODES.PLACEMENT_ID_INVALID_VALUE });
```
patmmccann, line 447: this is a breaking change and must wait for prebid 8

patmmccann (changes requested):

modules/aidemBidAdapter.js
patmmccann: linking #8539

modules/aidemBidAdapter.js
gsollazzo: Hello Chris! We have a rate limiting feature, used to limit the QPS sent to AIDEM SSP. In case this option is enabled by the publisher (rateLimit !== 1), we generate a random value to decide if we will send or not the bidRequest to our SSP. The Math.random() call happens only if the publisher has enabled this feature.

modules/aidemBidAdapter.js
darkstarac: Done

gsollazzo:
@ChrisHuie all comments and relative change requests seems resolved, may we proceed with merge? Thanks!

ChrisHuie (approved):

modules/aidemBidAdapter.js
patmmccann: do you still want this?

gsollazzo: Yes indeed! What's the procedure for adding this?

PR #9459, opened 2023-01-24
ArcSpan RTD Module: Initial Release

josecu:
## Type of change
- [x] Feature

## Description of change

Submitting the RTD Module for ArcSpan for initial review.

## Other information

Docs PR: https://github.com/prebid/prebid.github.io/pull/4304


modules/arcspanRtdProvider.js
```diff
@@ -0,0 +1,71 @@
...
+  name: SUBMODULE_NAME,
+  init: init,
+  getBidRequestData: alterBidRequests,
+};
+
+function init(config, userConsent) {
+  if (typeof config.params.silo === 'undefined') {
+    return false;
+  }
+  var _s = document.createElement('script');
```
patmmccann, line 19: please use our methods to insert scripts

patmmccann: also can you check for  window.arcobj1 before appending the script and recommend that it is deployed that way in your doc, you'll get better coverage that way. also, you'll want to make the location of the script configurable and fallback to a place that is locked to a version if you decide you do want prebid sourcing your location

modules/arcspanRtdProvider.js
```diff
@@ -0,0 +1,71 @@
...
+        keywords: _v1s.toString(),
+        content: _content,
+      },
+    };
+    mergeDeep(reqBidsConfigObj.ortb2Fragments.global, _ortb2);
+  }
+  callback();
+}
+
+submodule(MODULE_NAME, arcspanSubmodule);
```
patmmccann, line 71: you should make a commit to submodules.json

modules/arcspanRtdProvider.js
```diff
@@ -0,0 +1,71 @@
...
+
+function init(config, userConsent) {
+  if (typeof config.params.silo === 'undefined') {
+    return false;
+  }
+  var scriptUrl;
+  if (config.params.silo === 'test') {
+    scriptUrl = 'https://localhost:8080/as.js';
+  } else {
+    scriptUrl = 'https://silo' + config.params.silo + '.p7cloud.net/as.js';
```
patmmccann, line 24: please handle publisher hosted js here, or cases where the publisher has already loaded the js before or in parallel with prebid.js to improve performance. Please provide instructions on this, as making your js dependent on prebid.js means less auctions will contain your enrichments

patmmccann (changes requested):
additional feedback on enabling parallel loading

modules/arcspanRtdProvider.js
josecu: Thanks for the quick review, @patmmccann! I implemented the insert script method & the check, but I'm not sure I understand the 2nd part of your last comment. Do you mean we would have to make the location configurable and implement the fallback if/when we decide to have prebid source our location? Or do you mean we should make the location configurable and implement the fallback now, in preparation for if/when we decide to have prebid source our location?

modules/arcspanRtdProvider.js
josecu: @patmmccann We don't currently support publisher-hosted JS. Is that a requirement here? I implemented the performance changes. I also added instructions to the documentation for the publisher to reference our onboarding documentation to deploy the tag outside of Prebid.js. Is that sufficient?

patmmccann: that works, ty!

patmmccann (approved):

Issue #8568, opened 2022-06-15
Transparent adapter and alias codes in auctionEnd events


patmmccann:
## Type of issue
Feature request

## Description
Third party entities using the prebid public apis eg browser extensions, publisher analytics tools, debuggers, web scrapers, and rendering libraries may often wish to know all three of the response bidder code, the adapter code, and the adapter code the adapter was aliased from. We should provide a way for these other entities to delineate all three in the auctionEnd event. One possibility is to put them in a vector where bidderCode would be of length one for normal requests, but up to length 3. Another possiblity is to bidderCode, adapterCode, origAdapterCode places in object structure. 

One alternative is we could make adapterRegistry accessible via api? 

## related
#8129

patmmccann:
Perhaps this could be of length 4 if the publisher aliases a known alias, eg suppose bidder code in the response is 'someentity' which is a alternative code for 'tx' which is a configured alias of 'trustx' which is a known alias of 'grid' ?

dgirardi:
I'm curious about why / how this would with `auctionEnd`? those three are properties of each bid - and `adapterCode` is already available. Only the alias resolution is missing.

Currently it's possible to do 

```
pbjs.onEvent('bidResponse', (bid) => {
   const {bidderCode, adapterCode} = bid;
})
pbjs.onEvent('auctionEnd', (auction) => {
   auction.bidsReceived.map(({bidderCode, adapterCode}) => ...)
   auction.winningBids.map(({bidderCode, adapterCode}) => ...)
})
```

is this about adding the alias resolution, or changing the arguments to the auctionEnd event?

patmmccann:
On further inspection, analytics adapters already have this info, eg https://github.com/prebid/Prebid.js/blob/6f4fb5b78a4bc31a1f84b59ec5b98d4e3bcf0179/modules/pubmaticAnalyticsAdapter.js#L201

patmmccann:
result of committee discussion is we plan to make adapterRegistry accessible via api; allow publishers to suppress this public api via configuration

jlquaccia:
Hey @patmmccann, I was thinking of taking this gh issue on next.  To clarify, it sounds like the req's for this ticket would be something like the following (let me know if im off, which i might be):

- create an adapterRegistry (just like bidderRegistry & aliasRegistry here: https://github.com/prebid/Prebid.js/blob/master/src/adapterManager.js#L47)
[...]

PR #9467, opened 2023-01-25
Prebid Core: Added aliasRegistry to the Public API

jlquaccia:
## Type of change
- [x] new feature

## Description of change
- Added aliasRegistry from adapterManager to the Public API

## Other information
Github Issue: https://github.com/prebid/Prebid.js/issues/8568
PR for documentation updates: https://github.com/prebid/prebid.github.io/pull/4308

### DIFF
--- a/integrationExamples/gpt/hello_world.html
+++ b/integrationExamples/gpt/hello_world.html
@@ -8,84 +8,106 @@
 -->
 
 <html>
-<head>
+  <head>
     <script async src="../../build/dev/prebid.js"></script>
-    <script async src="https://www.googletagservices.com/tag/js/gpt.js"></script>
+    <script
+      async
+      src="https://www.googletagservices.com/tag/js/gpt.js"
+    ></script>
     <script>
-        var FAILSAFE_TIMEOUT = 3300;
-        var PREBID_TIMEOUT = 1000;
+      var FAILSAFE_TIMEOUT = 3300;
+      var PREBID_TIMEOUT = 1000;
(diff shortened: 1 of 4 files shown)


patmmccann:
that was simpler than i expected it to be lol :) we want to add one feature here I think. In the committee I think we wanted to preserve the ability of publishers to keep this non-public if for some reason that was important to them, so can we add a config check, with pubs using the 'https://docs.prebid.org/dev-docs/publisher-api-reference/setConfig.html' call to do something maybe like pbjs.setConfig({ 'aliasRegistry': 'private'}) ?

patmmccann (changes requested):
awesome, ty! let's add the ability to hide it in config in case anyone is relying on private aliases

patmmccann:
i think what i am suggesting in js would look something like: `if not config.getConfig('aliasRegistry') === 'private' {$$PREBID_GLOBAL$$.aliasRegistry = adapterManager.aliasRegistry};`

jlquaccia:
@patmmccann haha yea definitely turned out to be a quick one, easier than I had originally thought.  Cool sounds good!  Refactored things a bit so that now the default is for the aliasRegistry to be public but if it is set to private the registry will no longer be accessible on the public api.  Also just updated the PR for the doc changes related to this PR: https://github.com/prebid/prebid.github.io/pull/4308

patmmccann (approved):

ChrisHuie (approved):

PR #9476, opened 2023-01-30
AdYouLike Bid Adapter : add pbjs version information

guiann:
## Type of change
- Feature

## Description of change
minor : add prebid js version information in bid request


guiann:
I'll add a unit test on the added case...

PR #9490, opened 2023-02-01
ZetaGlobalSsp bid adapter: bidfloor module

asurovenko-zeta:
## Type of change
- [x] Feature

## Other information
PR in doc repo: https://github.com/prebid/prebid.github.io/pull/4322


asurovenko-zeta:
@ChrisHuie can you look at the build? Looks like it's failed not because of my changes. Isn't it? Can't get the reason of fail

ChrisHuie (approved):

PR #9496, opened 2023-02-01
Sonobi Bid Adapter:  remove userid query param

JonGoSonobi:
## Type of change

## Description of change
This removes a legacy query param "userid" from the Sonobi Bid Request Endpoint

### DIFF
--- a/test/spec/modules/sonobiBidAdapter_spec.js
+++ b/test/spec/modules/sonobiBidAdapter_spec.js
@@ -547,7 +547,7 @@ describe('SonobiBidAdapter', function () {
       expect(bidRequests.method).to.equal('GET');
       expect(bidRequests.data.ref).not.to.be.empty;
       expect(bidRequests.data.s).not.to.be.empty;
-      expect(bidRequests.data.userid).to.be.empty;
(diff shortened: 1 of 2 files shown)


test/spec/modules/sonobiBidAdapter_spec.js
```diff
@@ -539,39 +539,6 @@ describe('SonobiBidAdapter', function () {
       ]);
     });
 
-    it('should return a properly formatted request with userid as a JSON-encoded set of User ID results', function () {
```
patmmccann, line 542: can you add a test that says basically, it doesnt choke if a publisher specifies this, even though it doesnt do anything

JonGoSonobi: Ya we can do that. I restored the test and removed the check on line 550. 

The other 2 checks for data.ref and data.s should stay the same. These are additional checks for other values the buildRequest function sets but should not be effected by the presence of a userid.

patmmccann (approved):

patmmccann:
Thanks for the changes!

PR #9522, opened 2023-02-08
nextMillennium Bid Adapter:  cookie sync URL

JacobKlein26:
## Type of change
- Bugfix

## Description of change
Cookie sync only works if there's a bid response (using ```ext.sync```), 
Added standard URL if there's no bid
(Also changed netRevenue to true)


ChrisHuie:
> Hey there, It would seem like when I updated to master, it pulled in all commits into the branch, I was just wondering if it's important that I redo it (not sure why, but rebase only makes it worse) - or if this is fine?

Looks fine to me 👍

JacobKlein26:
Hi @ChrisHuie 
Thanks so much for that, 
What about the code that doesn't have a check, is that an issue? 
And if yes, can you please help me out with how to fix that, as I don't really see any info on that.

osazos:
Hi @JacobKlein26 please add test case for your new part of code.

JacobKlein26:
@osazos Done!

Also, do you think there's any chance we can still get this in before tomorrow's release?

osazos (approved):
LGTM

Issue #9529, opened 2023-02-09
Fix for empty targeting key hb_acat being sent in DFP call


jlquaccia:
## Type of issue
Bug

## Description
https://github.com/prebid/Prebid.js/issues/9217 setup the logic pass along acat, if present with dfp calls.  However, the default for hb_acat is being set to an empty string and being sent, `hb_acat=`.  If acat doesn't exist on a page, hb_acat= should not be sent at all.

## Other information
Relates to https://github.com/prebid/Prebid.js/issues/9217

bretg:
@jlquaccia - thanks for reporting this, but can we close this issue since there's a PR open already?

jlquaccia:
@bretg, sure, if the PR for this ticket looks good we can close this one.

Issue #9217, opened 2022-11-08
New category optional targeting


patmmccann:
## Type of issue

Sometimes it is important to do competitive category exclusions on direct business in the ad server

Please make hb_acat an optional targeting key, following the model of hb_adomain, which copies it from bid.meta

patmmccann:
This was discussed in publisher committee and had remarkably high demand as a feature

PR #9530, opened 2023-02-09
Prebid Core: Only Send Optional Category Targeting Key if ACAT is Present on Page

jlquaccia:
## Type of change
- [x] Bugfix

## Description of change
- The hb_acat K/V pairing should only be sent with DFP calls if a value for ACAT is present.

## Other information
Related: https://github.com/prebid/Prebid.js/issues/9529, https://github.com/prebid/Prebid.js/issues/9217

@pm-azhar-mulla can you review?


pm-azhar-mulla:
LGTM

test/spec/auctionmanager_spec.js
```diff
@@ -232,6 +232,15 @@ describe('auctionmanager.js', function () {
       assert.deepEqual(response, expected);
     });
 
+    it('should suppress acat if undefined', function () {
+      bid.meta.primaryCatId = ''
+      let expected = getDefaultExpected(bid);
+      delete expected.hb_acat;
+      let response = getKeyValueTargetingPairs(bid.bidderCode, bid);
+      assert.deepEqual(response, expected);
+      bid.meta.primaryCatId = 'IAB-test';
```
dgirardi, line 241: this should be in the test teardown - but really, I think this shows that on test setup, line 223, we should be `deepClone(DEFAULT_BID)`.

the problem with doing it this way is that if the test fails for any reason in the future, for example we change the business logic around acat, the teardown does not run and instead of one failure you'll get many false failures from subsequent tests that expect `bid.meta` to be "clean".

jlquaccia: @dgirardi thanks, that makes total sense regarding how my test was setup.  I just refactored my test to use `deepClone` instead of using the `bid` var that was in use for multiple tests.

dgirardi (approved):

jlquaccia:
Hey @robertrmartinez, could you provide a second review for this one when you have a chance?

robertrmartinez (approved):
looks good

PR #9543, opened 2023-02-14
Prebid Core: Define "VIDEO" compile time feature flag

mbcrute:
## Type of change

- [x] Feature

## Description of change

Adds a `"VIDEO"` feature flag that can be used to remove code specific to video ads at 
compile time. This PR includes Prebid core. We've also implemented this feature flag in a 
handful of the bid adapters that we use and are prepared to submit additional PRs for those 
modules.

With just the changes included in this PR we're able to shave ~4.1 KB off the size of our build 
of Prebid.js...

```sh
$ ./node_modules/.bin/gulp build-bundle-prod --modules=modules.json > /dev/null 2>&1 && ls -l build/dist/prebid.js
-rw-r--r--  1 crutem  staff  223835 Feb 14 11:05 build/dist/prebid.js

> ./node_modules/.bin/gulp build-bundle-prod --disable=VIDEO --modules=modules.json > /dev/null 2>&1 && ls -l build/dist/prebid.js
-rw-r--r--  1 crutem  staff  219634 Feb 14 11:04 build/dist/prebid.js
```

Implementing the feature flag in the bid adapter modules obviously shrinks the bundle size further.

## Other information

**TESTING:** We have been testing these changes in our fork of Prebid on our website for approximately 2 weeks. We have seen no significant changes in any of our metrics (CPM, CTR, viewability, etc).

### DIFF
--- a/src/prebid.js
+++ b/src/prebid.js
@@ -225,12 +225,15 @@ function validateAdUnit(adUnit) {
 export const adUnitSetupChecks = {
   validateAdUnit,
   validateBannerMediaType,
-  validateVideoMediaType,
   validateSizes
 };
 
 if (FEATURES.NATIVE) {
   Object.assign(adUnitSetupChecks, {validateNativeMediaType});
+}
+
+if (FEATURES.VIDEO) {
+  Object.assign(adUnitSetupChecks, { validateVideoMediaType });
 }
 
 export const checkAdUnitSetup = hook('sync', function (adUnits) {
@@ -248,7 +251,7 @@ export const checkAdUnitSetup = hook('sync', function (adUnits) {
       if (mediaTypes.banner.hasOwnProperty('pos')) validatedBanner = validateAdUnitPos(validatedBanner, 'banner');
     }
 
-    if (mediaTypes.video) {
+    if (FEATURES.VIDEO && mediaTypes.video) {
       validatedVideo = validatedBanner ? validateVideoMediaType(validatedBanner) : validateVideoMediaType(adUnit);
       if (mediaTypes.video.hasOwnProperty('pos')) validatedVideo = validateAdUnitPos(validatedVideo, 'video');
     }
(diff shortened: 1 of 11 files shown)


src/prebid.js
```diff
@@ -995,21 +998,23 @@ $$PREBID_GLOBAL$$.getHighestCpmBids = function (adUnitCode) {
  * @alias module:pbjs.markWinningBidAsUsed
  */
 $$PREBID_GLOBAL$$.markWinningBidAsUsed = function (markBidRequest) {
```
dgirardi, line 1000: It may be just a stylistic difference, but I think it'd be better to not have this method rather than have it silently do nothing.

dgirardi (commented):
Thank you for putting this together! I think there's a few more things that can be tagged out. My strategy would be to `build-bundle-verbose --disable VIDEO` to get a somewhat legible compilation output and try to spot the stragglers. [here's what that produces with this version](https://gist.githubusercontent.com/dgirardi/01644478d8abd6442619d68926c6f882/raw/13fd2646dab499ea4a190d7dee268214c8e78174/prebid-core.js), and a few of the targets that I see:

- [custom renderers have some video-specific logic](https://github.com/prebid/Prebid.js/blob/e95769360d6ac5a3cdfe06df1bbd259d62255d1e/src/Renderer.js#L82). There's a good chance that if you don't need video you also don't need custom renderers at all, but I'm not sure that's always the case. It may be worth tagging renderers as its own feature (maybe separately from this PR).
- there's some [arcane adpod-related logic](https://github.com/prebid/Prebid.js/blob/e95769360d6ac5a3cdfe06df1bbd259d62255d1e/src/adapters/bidderFactory.js#L474) which should not be in core in the first place, and I'm almost sure you can do without if you're not interested in video.
- because the [adpod module `import`s some video symbols](https://github.com/prebid/Prebid.js/blob/e95769360d6ac5a3cdfe06df1bbd259d62255d1e/modules/adpod.js#L31), it prevents them from being removed as dead code. this should be fixable replacing the imported `callPrebidCache` with `getHook('callPrebidCache')`.
- [some special case for video in renderAd](https://github.com/prebid/Prebid.js/blob/e95769360d6ac5a3cdfe06df1bbd259d62255d1e/src/prebid.js#L535)

You may also want to [tag out ORTB conversion utitlies for video](https://github.com/prebid/Prebid.js/blob/e95769360d6ac5a3cdfe06df1bbd259d62255d1e/libraries/ortbConverter/processors/default.js#L60) (here's the [native tagging for reference](https://github.com/prebid/Prebid.js/blob/e95769360d6ac5a3cdfe06df1bbd259d62255d1e/libraries/ortbConverter/processors/default.js#L118)). This does not save any bytes in core, but it would for adapters that use the library which will hopefully become more and more common in time.

mbcrute:
@dgirardi Thank you for the feedback and suggestions. I've pushed some additional commits to address most of them.

Regarding the video-specific logic with custom renderers, I think I agree with you that that's worth tagging as its own feature in a separate PR.

src/native.js
```diff
@@ -174,7 +174,7 @@ function isOpenRTBAssetValid(asset) {
       logError(`for data asset 'type' property must be a number`);
       return false;
     }
-  } else if (asset.video) {
+  } else if (FEATURES.VIDEO && asset.video) {
```
dgirardi, line 177: I'm not sure this is right; if you turn off video but not native, does that mean that you don't want video within native?

mbcrute: Fair point. Reverted!

dgirardi:
could you "unfollow" your fork of prebid.js in circleCI? (or, if you know of a better way to convince it to use the correct account, I'm all ears).

mbcrute:
@dgirardi Sorry, but I'm not sure how to do what you're asking.

dgirardi (approved):

ChrisHuie (approved):

PR #9555, opened 2023-02-16
Permutive RTD Module: migrate magnite to ortb2

AntonioGargaro:
## Type of change

- [ ] Bugfix
- [x] Feature
- [ ] New bidder adapter   
- [ ] Build related changes
- [ ] CI related changes

- [ ] Other

## Description of change

This is the first step to migrating away from the pre-v4.30 of setting targeting values. The RTD module is now making `rubicon` targeting values available on the ortb2 object. After inspecting the `rubiconBidAdapter`, they automatically convert targeting key-values from `params.visitor` to `ortb2.user.ext.data`. So the migration for the RTD module is to just remove our mutation of `params.visitor` and write these targeting values directly to `ortb2.user.ext.data`. This data is only applicable to rubicon and follows the OpenRTB 2.5 specification of placing bidder-specific extensions in `.ext` properties.

## Other information

In a recent [PR](https://github.com/prebid/Prebid.js/pull/9236#discussion_r1043632338), our special `bid.params` handlings were put on notice of being removed if we didn't address the issue by the end of Q2. This PR is the first step in migrating these handlers. `rubicon` is the first bidder we're migrating with `appnexus` and `ozone` committed to being addressed by the end of Q2. We're working with each partner on the best place to make data available in ortb2 that each bidder expects.

### DIFF
--- a/modules/permutiveRtdProvider.js
+++ b/modules/permutiveRtdProvider.js
@@ -133,6 +133,8 @@ export function setBidderRtb (bidderOrtb2, moduleConfig, segmentData) {
  * @return {Object} Merged ortb2 object
  */
 function updateOrtbConfig(bidder, currConfig, segmentIDs, sspSegmentIDs, transformationConfigs, segmentData) {
+  const customCohortsData = deepAccess(segmentData, bidder) || []
+
   const name = 'permutive.com'
 
   const permutiveUserData = {
@@ -159,17 +161,15 @@ function updateOrtbConfig(bidder, currConfig, segmentIDs, sspSegmentIDs, transfo
   const updatedUserKeywords = (currentUserKeywords === '') ? keywords : `${currentUserKeywords},${keywords}`
   deepSetValue(ortbConfig, 'ortb2.user.keywords', updatedUserKeywords)
 
-  // Set bidder specific extensions
-  if (bidder === 'rubicon') {
-    if (segmentIDs.length > 0) {
-      deepSetValue(ortbConfig, 'ortb2.user.ext.data.' + PERMUTIVE_STANDARD_KEYWORD, segmentIDs)
-    }
+  // Set user extensions
+  if (segmentIDs.length > 0) {
+    deepSetValue(ortbConfig, `ortb2.user.ext.data.${PERMUTIVE_STANDARD_KEYWORD}`, segmentIDs)
+    logger.logInfo(`Extending ortb2.user.ext.data with "${PERMUTIVE_STANDARD_KEYWORD}"`, segmentIDs)
+  }
 
-    if (segmentData?.rubicon?.length > 0) {
-      deepSetValue(ortbConfig, 'ortb2.user.ext.data.' + PERMUTIVE_CUSTOM_COHORTS_KEYWORD, segmentData.rubicon.map(String))
-    }
-
-    logger.logInfo(`Extending ortb2.user.ext.data for ${bidder}`, deepAccess(ortbConfig, 'ortb2.user.ext.data'))
+  if (customCohortsData.length > 0) {
+    deepSetValue(ortbConfig, `ortb2.user.ext.data.${PERMUTIVE_CUSTOM_COHORTS_KEYWORD}`, customCohortsData.map(String))
+    logger.logInfo(`Extending ortb2.user.ext.data with "${PERMUTIVE_CUSTOM_COHORTS_KEYWORD}"`, customCohortsData)
   }
 
   logger.logInfo(`Updating ortb2 config for ${bidder}`, ortbConfig)
--- a/modules/permutiveRtdProvider.js
+++ b/modules/permutiveRtdProvider.js
@@ -8,12 +8,19 @@
 import {getGlobal} from '../src/prebidGlobal.js';
 import {submodule} from '../src/hook.js';
 import {getStorageManager} from '../src/storageManager.js';
-import {deepAccess, deepSetValue, isFn, logError, mergeDeep, isPlainObject, safeJSONParse} from '../src/utils.js';
+import {deepAccess, deepSetValue, isFn, logError, mergeDeep, isPlainObject, safeJSONParse, logMessage} from '../src/utils.js';
 import {includes} from '../src/polyfill.js';
 
 const MODULE_NAME = 'permutive'
+const PREFIX = MODULE_NAME + 'RTD'
+
+const logger = {
+  message: (...a) => logMessage(PREFIX, ...a),
+}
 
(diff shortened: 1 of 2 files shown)


modules/permutiveRtdProvider.js
```diff
@@ -262,17 +262,6 @@ function getDefaultBidderFn (bidder) {
...
-      if (isPStandardTargetingEnabled(data, acEnabled)) {
-        const segments = pStandardTargeting(data, acEnabled)
-        deepSetValue(bid, 'params.visitor.p_standard', segments)
-      }
-      if (data.rubicon && data.rubicon.length) {
-        deepSetValue(bid, 'params.visitor.permutive', data.rubicon.map(String))
-      }
-
-      return bid
-    },
```
AntonioGargaro, line 275: This is the first special handling removed. The others are planned in follow up PRs.

modules/permutiveRtdProvider.js
```diff
@@ -383,17 +377,57 @@ function iabSegmentId(permutiveSegmentId, iabIds) {
...
-      initSegments(reqBidsConfigObj, callback, customModuleConfig)
-    });
-    makeSafe(function () {
-      // Route for bidders supporting ORTB2
-      setBidderRtb(reqBidsConfigObj.ortb2Fragments?.bidder, customModuleConfig)
+      if (!permutiveReadyCallbackRegistered && moduleConfig.waitForIt && isPermutiveOnPage()) {
+        // Prevent multiple calls to set cohorts
+        permutiveReadyCallbackRegistered = true
+
+        window.permutive.ready(function () {
```
zshnr, line 421: Curious what difference this makes compared to previously? Is it only for updating cohorts?

modules/permutiveRtdProvider.js
```diff
@@ -283,7 +272,12 @@ function getDefaultBidderFn (bidder) {
     }
   }
 
-  return bidderMap[bidder]
+  // On no default bidder just return the same bid as passed in
+  function bidIdentity(bid) {
+    return bid
+  }
+
+  return bidderMap[bidder] || bidIdentity
```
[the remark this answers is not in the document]
AntonioGargaro: So we don't break the public API of our module. `rubicon` did return a default bidder but since it was removed it would now return undefined.  On [line 208](https://github.com/prebid/Prebid.js/pull/9555/files#diff-9476fbe01d356d6cb1e8a1a00bcdda0576f4b82641ad355f61f3252ddd821492R208), we would then pass the default handler to any publisher defined _custom_ handler. Ideally, they would check there is a default handler that exists first (i.e `if (defaultFn) { defaultFn() }`) but we can't guarantee any custom script on a publisher's page actually does this check. Hence if the invoke an undefined, it will throw an error. So we now provide a fallback which does nothing to the bid.

modules/permutiveRtdProvider.js
AntonioGargaro: I refactored this as the previously existing `initSegments` function would actually call `callback()` before the module updated the ortb2 interface. I'm not sure what the implications of that are, but the `callback` should be called once everything is finished in the RTD module according to the [docs](https://docs.prebid.org/dev-docs/add-rtd-submodule.html#getbidrequestdata).

Another issue I noted was that if the publisher had `config.waitForIt`, it would set the legacy segments twice on each call as it would keep registering a callback even after the Permutive SDK is in real-time. When this happens, the callback is executed immediately from then onwards, performing a double call. 
This now only registers through the callback until Permutive is in real-time and then once it is in realtime, just completes the RTD auction as usual.

modules/permutiveRtdProvider.js
patmmccann: awesome ty!

patmmccann:
Thanks so much for being proactive about this!

modules/permutiveRtdProvider.js
```diff
@@ -174,6 +159,21 @@ function updateOrtbConfig (currConfig, segmentIDs, sspSegmentIDs, transformation
   const updatedUserKeywords = (currentUserKeywords === '') ? keywords : `${currentUserKeywords},${keywords}`
   deepSetValue(ortbConfig, 'ortb2.user.keywords', updatedUserKeywords)
 
+  // Set bidder specific extensions
+  if (bidder === 'rubicon') {
+    if (segmentIDs.length > 0) {
+      deepSetValue(ortbConfig, 'ortb2.user.ext.data.' + PERMUTIVE_STANDARD_KEYWORD, segmentIDs)
```
dgirardi, line 165: this may be a dumb question - why is this set only for rubicon? with my naivete I'd expect it to be standardized and useful for everyone, or superfluous and the responsibility of rubicon to deduce from `user.data`.

modules/permutiveRtdProvider.js
```diff
@@ -174,6 +159,21 @@ function updateOrtbConfig (currConfig, segmentIDs, sspSegmentIDs, transformation
...
+    }
+
+    if (segmentData?.rubicon?.length > 0) {
```
dgirardi, line 168: same question - why isn't this such that each `bidder` picks `segmentData[bidder]`?

modules/permutiveRtdProvider.js
AntonioGargaro: It's specific for Rubicon because this is how they read from the ortb2 object and it falls under `ortb2.user.ext.data`, where the `.ext` being they key part of the path. Following the OpenRTB 2.5 specification, we must place bidder-specific extensions in `.ext` properties.

I hope that makes sense 😬

modules/permutiveRtdProvider.js
AntonioGargaro: I think the [above comment should answer this](https://github.com/prebid/Prebid.js/pull/9555#discussion_r1129223057), it's a specific extension for Rubicon.

modules/permutiveRtdProvider.js
AntonioGargaro: It's also worth pointing out, this data already exists under `ortb2.user.data`. We're just adding it into additional places in the ortb2 object for a bidder.

patmmccann: The goal is, the word 'rubicon' should not appear in your code.

modules/permutiveRtdProvider.js
patmmccann: This is a rules violation; you are not allowed to have a "specific extension for Rubicon."

patmmccann (changes requested):
If you want to set data for a specific bidder, read that list of bidders out of pub config or from the response on your endpoint. Do not refer to them in your js.

AntonioGargaro:
Hey @patmmccann, I've changed it to be available to all bidders in this case.

patmmccann (approved):

Issue #9551, opened 2023-02-15
Promise.race Error on UserId module on latest version 7.36.0


ialex:
## Type of issue

Bug

## Description

I found out that in the current PrebidJS version (at the time of writing this, v7.36.0), and in combination with Angular a GreedyPromise.race doesnt return a Promise causing Type Error and preventing proper prebid execution. 

Line where it happens
https://github.com/prebid/Prebid.js/blob/master/modules/userId/index.js#L514

Error
```
prebid.js:formatted:31225 Uncaught TypeError: k.Z.race(...).finally is not a function
    at f (prebid.js:formatted:31225:57)
    at $ (prebid.js:formatted:31238:29)
    at prebid.js:formatted:31656:25
    at 2695 (prebid.js:formatted:31689:18)
    at r (prebid.js:formatted:9253:28)
    at prebid.js:formatted:31798:17
    at r.O (prebid.js:formatted:9273:37)
    at prebid.js:formatted:31802:15
    at n (prebid.js:formatted:9338:37)
    at prebid.js:formatted:30676:57
    at prebid.js:formatted:33305:6
```

## Steps to reproduce

I can provide a test page where this happens but its behind login 

### Expected results
Prebids runs
## Platform details
Prebid 7.36.0 Chrome

## Other information
Similar to issue #9422

dgirardi:
Could you try this - _before_ loading either Angular or Prebid?

```
<script>
window.pbjs = window.pbjs || {};
window.pbjs.Promise = window.Promise;
</script>
```

If that doesn't work, how would we go about getting access to the test page?

dgirardi:
Confirmed off line that the above fixes the issue. Tagging @javiermarinros since you might be also affected at some point - apparently it's not enough to reimplement `resolve`  / `reject` as in #9426, although that might depend on the zone.js' version. I'd much rather bypass its monkey-patching than reimplement _everything_ in Promise.

javiermarinros:
@dgirardi snippet fixes the issue, but anyway this error is kind of obscure and hard to track, and it looks like can affect potentially all Angular users. Do you have any idea how to better fix this or at least communicate it to Angular developers?

dgirardi:
@javiermarinros  Maybe we do need to assume that Promise is broken and just not rely on it from Prebid. We could provide an independent reimplementation as an optional module and suggest it as something you'd want if you use Angular? 
[...]

Issue #8903, opened 2022-08-26
Pr #8626 breaks prebid when publisher is using the bluebirdjs module


bjorn-lw:
## Type of issue
Pr #8626 breaks prebid when publisher is using the bluebirdjs module. (https://github.com/petkaantonov/bluebird/)

## Description
The bluebirdjs doesn¨t allow "promise" to be subclassed, throwing an error and halting prebid. http://bluebirdjs.com/docs/error-explanations.html#2.-you-are-trying-to-subclass-promise

## Steps to reproduce
The easiest way would probably be to use a proxy to redirect https://content.lwadm.com/lw/6.26.0/prebid.js to a version of prebid.js >= 7.7.0 on the site https://www.gp.se.

## Test page
See above

## Other information
Unfortunately, the publisher using this module cannot remove it or fix the problem easily. I also realize it may not be in the interest of Prebid.org to fix issues specific to a publisher, but I could imagine this might be a problem for more publishers out there + there might be other modules causing similar issues. Also, the publisher is not a small one and not being able to upgrade beyond 7.6.0 will be a major issue. Though we can create a custom branch removing the new functionality, it's not really sustainable in the long run.

One suggestion is to add a config setting to control whether this extended version of promise should be used or not if the problem can't be fixed on the Prebid end?

patmmccann:
The first link says not to user bluebird on modern browsers. Prebid 6.x+ is only for modern browsers.

patmmccann:
Apologies closed by pressing the wrong button, but i don't expect we will be fixing this

dgirardi:
I believe the error wouldn't happen if Prebid is loaded before bluebird. Would that be a viable fix? 

If not, another option could be to have Prebid look for say `window.pbjsPromise` and use that if it's defined, so that you could do 

```html
<script>
window.pbjsPromise  = Promise;
</script>
<!-- load bluebird, prebid, etc -->
```

bjorn-lw:
Thank you very much for replying!

Unfortunately, the chances that the publisher can make such changes that bluebird is removed from their sites is not a possibility right now as it would requite to large rewrites (which is planned, but won't be ready for a pretty long while).

Also, sticking with older versions of Prebid is not an option either and fact is that it has worked flawlessly up until this change so if this one could be solved it would help a lot.

[...]

PR #9558, opened 2023-02-17
PBjs Core: do not rely on an extendable `window.Promise`

dgirardi:
## Type of change

- [x] Bugfix

## Description of change

This updates `GreedyPromise` to be completely independent from `window.Promise`. This is because some libraries, notably Angular, monkey-patch `Promise` in a way that breaks its extensibility. 

## Other information

Fixes https://github.com/prebid/Prebid.js/issues/9551

Other reports of the same issue:

https://github.com/prebid/Prebid.js/issues/9422
https://github.com/prebid/Prebid.js/issues/8903


ChrisHuie (approved):

robertrmartinez:
Are there unit tests that can be added since this is just removing the old ones?

dgirardi:
@robertrmartinez there's still [147 test cases left](https://github.com/prebid/Prebid.js/blob/cd3e58e71dbb4a0019ba4745394699720e0febdc/test/spec/unit/utils/promise_spec.js); I am open to suggestions if you think there's a gap, but I can't spot one.

What this removes are tests about the `unhandledRejection` event, which I am jettisoning together with the "real" `window.Promise`.

robertrmartinez:
You caught me

I did my usual initial look

 "A code change with no unit test additions - automatic ask"

I apologize, I should have realized it was a PR done by the great @dgirardi ! 😉

ialex (approved):

robertrmartinez (approved):

PR #4940, opened 2020-03-06
Support Smart Path Optimization Technology Hyperautomation & V5 Endpo…

GLStephen:
- [ X ] Other

## Description of change
This PR updates the functionality of the PubWise Prebid Module to support:

- Smart Path Optimization Technology Hyperautomation
- PubWise V5 Endpoints
- Session Level Reporting
- Less chatty even sending
- Quieter error reporting with toggle

PR #9559, opened 2023-02-17
Pubwise Analytics Adapter: Fix empty-string sessionId when loading unexpired sessions

bishpls:
## Type of change

- [x] Bugfix

## Description of change
Event metadata currently fails to properly send sessionId if the sessionId has not yet expired. Fixing that logic path + trivial 'don't hit localStorage repeatedly' var name refactor to the currently errant function.

### DIFF
--- a/modules/pubwiseAnalyticsAdapter.js
+++ b/modules/pubwiseAnalyticsAdapter.js
@@ -304,11 +304,14 @@ pubwiseAnalytics.storeSessionID = function (userSessID) {
 
 // ensure a session exists, if not make one, always store it
 pubwiseAnalytics.ensureSession = function () {
-  if (sessionExpired() === true || userSessionID() === null || userSessionID() === '') {
+  let sessionId = userSessionID();
+  if (sessionExpired() === true || sessionId === null || sessionId === '') {
     let generatedId = generateUUID();
     expireUtmData();
     this.storeSessionID(generatedId);
     sessionData.sessionId = generatedId;
+  } else if (sessionId != null) {
+    sessionData.sessionId = sessionId;
   }
   // eslint-disable-next-line
   // console.log('ensured session');


bishpls:
Small background research: it looks like sessionId was always being set as of this PR: https://github.com/prebid/Prebid.js/pull/4940/files#diff-58cbb073b0b1d6d4fa71e7cda81265ad0965e81d578ae60b9315e2c1df29e06dR312

and here, we see the problem introduced -- it's only being set on sessionData if the session is (re)generated and stored:

https://github.com/prebid/Prebid.js/pull/5677/files#diff-58cbb073b0b1d6d4fa71e7cda81265ad0965e81d578ae60b9315e2c1df29e06dR310

ChrisHuie (approved):

GLStephen:
@bishpls thanks for the fix here, I appreciate it

bishpls:
btw, @GLStephen -- do you have a preferred workflow for community contributions to the Pubwise adapter? (Github Issue submission? Going straight to submitting a PR, like this? Pinging about the problem via Twitter or the official help channels?)

I recently ran into some issues / off-spec behavior regarding ingestion of first-party data into the adapter's buildRequests function; long story short, it seems to only pull from the 'content' object set at the top level of config via setConfig/mergeConfig; this is a bit off-spec (current guidance is to read from the ortb2 object, a la: https://docs.prebid.org/features/firstPartyData.html#how-bid-adapters-should-read-first-party-data), and trying to add auction-specific first-party data (https://docs.prebid.org/features/firstPartyData.html#supplying-auction-specific-data) completely fails. 

(The Analytics adapter *is* correctly ingesting this data, it's only a problem with the bid adapter at the moment!)

I've added a quick patch on my end and the Pubwise payload appears to now be receiving the FPD (under payload.site.content), but I'd love to fully confirm that the current off-spec behavior is being passed into the bidstream on your backend!

Additionally, I'd like to double check with you before submitting a PR updating the FPD logic in the adapter, since I'd like to make sure any update plays nicely with how your backend is currently set up to ingest the payload. Any preference on whether you'd prefer your team to tackle it, or should I go ahead and make the fix? 

If you don't mind me submitting the PR: are you ingesting FPD solely on payload.site.content, or ALSO at the payload.site level?

Example payload below -- adding IAB Content Taxonomy (cattax) and content categories (cat) is valid at both site and site.content level (meaning slightly different things) per the OpenRTB spec; should the Pubwise payload continue to ONLY send FPD under payload.site.content?  

`
payload = {
  site: {
    content: {
      cattax: 2,
      cat: ['381'],
    },
    cattax: 2,
    cat: ['381'],
    domain: 'test.com'
    page: 'https://test.com/',
    ref: 'google.com',
    publisher: {
      id: 'xxxxxx'
    },
  },
  imp: {},
  source: {},
  user: {},
  ext: {},
  device: {},
  cur: ["USD"],
  at: 1,
};
`

PR #9583, opened 2023-02-23
Nativo Bid Adapter: adding UserId support

jsfledd:
## Type of change

- [x ] Feature

## Description of change
- Adding support for utilizing the UserIds module
- Added additional QS param to bid request endpoint
- Refactored request url creation
- Added / Testing a new way to handle generating data for the bid request url


ChrisHuie:
Can you please update your docs with this -> https://github.com/prebid/prebid.github.io/blob/master/dev-docs/bidders/nativo.md

jsfledd:
> Can you please update your docs with this -> https://github.com/prebid/prebid.github.io/blob/master/dev-docs/bidders/nativo.md

Added: https://github.com/prebid/prebid.github.io/pull/4395

modules/nativoBidAdapter.js
```diff
@@ -615,6 +703,61 @@ function getLargestSize(sizes, method = area) {
   })
 }
 
+/**
+ * Fire a Nativo Error AJAX request to log errors
+ * @param {String} errorMessage - The error message to send
+ */
+function fireNativoError(errorMessage) {
```
ncolletti, line 710: @ChrisHuie What's the stance on an additional endpoint call for error tracking from a module, should this be behind a functionality flag for publishers to determine it's usage? Is there another suggested pattern for this?

modules/nativoBidAdapter.js
```diff
@@ -451,6 +459,89 @@ export const spec = {
...
+    this.values = new Set()
+  }
+
+  processBidRequestData(bidRequest, bidderRequest) {
+    const eids = bidRequest.userIdAsEids
+
+    if (eids == undefined) return
+
+    for (let eid of eids) {
+      if (this.values.has(eid)) {
```
dgirardi, line 509: I believe this will in practice always be false. It's only true if the same object (meaning the same identity & memory location) appears twice in the array, which would be a bug for which a fix would make more sense than this error notification.

Stepping back, what's the intent here? even if you meant to check for duplicates by value, that's still something that I don't see happening.

jsfledd: @dgirardi The intent is to determine if eid values could differ across bidRequests. I noticed that each bidRequest has a userIdAsEids property but didn't know if they contained the same values or could ever be different. This was an initial attempt to get some insight into whether these values would ever be different but I see your point regarding the this.values.has(eid) check. 

Do you have any insight into whether userIdAsEids values can differ across bidRequests? I can adjust the logic accordingly.

dgirardi: Within one invocation to `buildRequests`, all the request objects [reference the same `userIdAsEids`](https://github.com/prebid/Prebid.js/blob/30c61f8f9bcae47f10bb37faec1c364d82059488/modules/userId/index.js#L487-L494).  So I was wrong, this would be true every time you have more than one bid request (I missed that you are reusing the same set across them).

Across different invocations (different auctions) the contents of `userIdAsEids` can change. 

If you're wondering why it's `bidRequest.userIdAsEids` instead of `bidderRequest.userIdAsEids` - good question. I have no answer.

dgirardi: Let me roll back a bit - by value they are identical, but now that I'm reading that code again I'm not sure they would be the same reference. But if I interpret your concern as "do I need to worry about different slots having different user ids", the answer is no.

jsfledd: Thank you @dgirardi for the insight. I will adjust accordingly on my end. Please stand by

modules/nativoBidAdapter.js
jsfledd: Hi @ncolletti and @ChrisHuie, I am also curious to the answer above. After speaking with @dgirardi below it seems like we may be able to use this logic as we can assume each auction's `bidRequest.userIdAsEids` will be the same. However, the idea of being able to log errors still sounds like it could be helpful in the future if this aligns with Prebid's adapter rules.

jsfledd: I have removed it for now.

modules/nativoBidAdapter.js
jsfledd: I have removed the duplicate check

ncolletti (approved):
lgtm

dgirardi:
@jsfledd, this was reverted because of problems with tests: there's a `describe.only` that was [probably left there accidentally](https://github.com/prebid/Prebid.js/pull/9691), but if I  attempt to remove it I get other test failures from the nativo suite.

jsfledd:
Apologies. I just fixed the issue and pushed the test update.

jsfledd:
I made the change to our fork: https://github.com/NativoPlatform/PrebidNativoBidder/commit/be48d28f5752cda02d8ceb320f13e5de0d25def7 but it's not pulling in here. Should I create another PR?

jsfledd:
Hello @ChrisHuie @dgirardi @ncolletti. Just following up here. Should I create a new PR for this? I've updated our forked repo but it's no longer pulling the changes into this PR

dgirardi:
@jsfledd I'd say go ahead with a new PR; I don't know if there's a better way.

PR #9586, opened 2023-02-24
AdsInteractive Bid Adapter: new bid adapter

danielpfei:
## Type of change

- [ ] Bugfix
- [ ] Feature
- [ x] New bidder adapter   
- [ ] Build related changes
- [ ] CI related changes

- [ ] Other


danielpfei:
documentation: 
https://github.com/prebid/prebid.github.io/blob/master/dev-docs/bidders/adsinteractive.md

modules/adsinteractiveBidAdapter.js
```diff
@@ -0,0 +1,100 @@
...
+    });
+  },
+
+  interpretResponse: (serverResponse, bidRequest) => {
+    if (!serverResponse || !serverResponse.body) {
+      return [];
+    }
+
+    let answer = [];
+    if (serverResponse.body.seatbid != null) {
```
mmoschovas, line 70: Can this check be combined with the previous condition?

modules/adsinteractiveBidAdapter.js
```diff
@@ -0,0 +1,96 @@
...
+        url,
+        data,
+        options,
+      };
+    });
+  },
+
+  interpretResponse: (serverResponse, bidRequest) => {
+    let answer = [];
+    if (!serverResponse || !serverResponse.body || serverResponse.body.seatbid != null) {
```
mmoschovas, line 66: The join of conditions here is incorrect. You can either keep the early return as you previously had like:

```
if (!serverResponse || !serverResponse.body || !serverResponse.body.seatbid) {
      return [];
    }
```

or you will need to revise this current check to 

```
if (serverResponse && serverResponse.body && serverResponse.body.seatbid) {
```

mmoschovas (approved):

PR #9587, opened 2023-02-24
ID5 Adapter: protect against local storage writing without consent

kzhang-id5:
## Type of change

- [x] Bugfix
- [ ] Feature
- [ ] New bidder adapter   
- [ ] Build related changes
- [ ] CI related changes

- [ ] Other

## Description of change
If a publisher is not using the GDPR enforcement module correctly, sometimes we get called getId() or extendId() even when the user doesn't have consent. In this case, we check against the consentData object in order to see if we truly do have consent, and prevent writing to local storage if we don't.


smenzer:
@jsnellbaker I'm happy to review this, but this was one of our engineers at ID5 that wrote this...just want to make sure you're ok with my review and that we don't need someone else to have a look.

jsnellbaker:
Thanks for the call-out @smenzer. I also tagged @jlukas79 to help review here.

patmmccann (approved):

patmmccann:
This doesnt directly hit the cmp, lgtm

jlukas79 (approved):

smenzer (approved):

Issue #8312, opened 2022-04-21
Feature request: Apply adapter bid adjustment to unknown bidder codes


patmmccann:
https://github.com/prebid/Prebid.js/pull/8216 carved out applying the adapter code bid adjustment function to unknown bidder codes. We would still like that feature, but since it is breaking, we would want to put it behind configuration. No known bidders currently return 'surprise' bidder codes (april 2022), but we would like to make sure adjustment functions are able to handle them should that occur.

jlquaccia:
Hey @patmmccann, was taking a look into this one but had a question about the ask.  Say we have the following bidderSettings in place:

```
        pbjs.bidderSettings = {
          standard: {
            bidCpmAdjustment: function (bidCpm, bid) {
              return bidCpm * 0.95;
            },
          },
          pubmatic: {
            catchUnknownBidderCodesWithAdapterBidAdjustment: true, ( new config option )
            bidCpmAdjustment: function (bidCpm, bid) {
              return bidCpm * 0.9;
            }
          }
        };
```

If the PubMatic response returned a bidder code of "unknownCode" then that bidderCode would normally get rejected since by default `allowAlternateBidderCodes` is set to false for `standard` and `pubmatic` above.  Would the ask for this ticket then, in this scenario, be to allow the unknown code to pass through and use the `pubmatic.bidCpmAdjustment` function? (since this unknown code came from pubmatic and the new config option above is present and set to allow this).

Basically, the `allowAlternateBidderCodes` and `allowedAlternateBidderCodes` functionality would override `catchUnknownBidderCodesWithAdapterBidAdjustment`, but `catchUnknownBidderCodesWithAdapterBidAdjustment` could optionally be used as a fallback if the other keys were absent, instead of just rejecting all bids with unknown bidder codes.

PR #9609, opened 2023-03-01
Bidder Settings: Support for option to apply adapter bid adjustment to unknown bidder codes

jlquaccia:
## Type of change
- [x] Feature

## Description of change
- Added optional configurable logic to to allow adapter bid adjustment functions to be invoked in the event a bidder code is uknown.

## Other information
https://github.com/prebid/Prebid.js/issues/8312


src/utils/cpm.js
```diff
@@ -2,9 +2,15 @@ import {auctionManager} from '../auctionManager.js';
 import {bidderSettings} from '../bidderSettings.js';
 import {logError} from '../utils.js';
 
-export function adjustCpm(cpm, bidResponse, bidRequest, {index = auctionManager.index, bs = bidderSettings} = {}) {
+export function adjustCpm(cpm, bidResponse, bidRequest, {index = auctionManager.index, bs = bidderSettings} = {}, gba = getBidAdjustment) {
   bidRequest = bidRequest || index.getBidRequest(bidResponse);
-  const bidCpmAdjustment = bs.get(bidResponse?.bidderCode || bidRequest?.bidder, 'bidCpmAdjustment');
+
+  const adapterAllowAlternateBidderCodes = bs.get(bidResponse?.adapterCode || bidRequest?.bidder, 'allowAlternateBidderCodes');
+  const catchUnknownBidderCodesWithAdapterBidAdjustment = bs.get(bidResponse?.adapterCode || bidRequest?.bidder, 'catchUnknownBidderCodesWithAdapterBidAdjustment');
```
dgirardi, line 9: 1. this setting (and variable) name is way too long, past the point of ridiculous. I'll offer `adjustAlternateBids`, or `adjustAllBids`.
2. what's the reason for looking at `allowAlternateBidderCodes`? If you have a bid from an alternate code (that is, one where `adapterCode` is different from `bidderCode`), it should mean that it was allowed already. 
3. I can't make sense of the fallback `bidResponse?.adapterCode || bidRequest?.bidder`. If you are using `bidRequest?.bidder`, are you not just using the "normal" bid adjustment rules?

src/utils/cpm.js
```diff
@@ -2,9 +2,15 @@ import {auctionManager} from '../auctionManager.js';
 import {bidderSettings} from '../bidderSettings.js';
 import {logError} from '../utils.js';
 
-export function adjustCpm(cpm, bidResponse, bidRequest, {index = auctionManager.index, bs = bidderSettings} = {}) {
+export function adjustCpm(cpm, bidResponse, bidRequest, {index = auctionManager.index, bs = bidderSettings} = {}, gba = getBidAdjustment) {
```
dgirardi, line 5: `gba` is a very thin wrapper around `bs`. it gets confusing if you can "decouple" them; if I pass in a particular instance of `bidderSettings`, I would expect it to be used throughout. to me this looks like you're asking for the same thing twice.

jlquaccia:
@dgirardi thanks for the feedback. I addressed what you brought up, could you take another look when you have the time?

src/utils/cpm.js
```diff
@@ -4,7 +4,10 @@ import {logError} from '../utils.js';
 
 export function adjustCpm(cpm, bidResponse, bidRequest, {index = auctionManager.index, bs = bidderSettings} = {}) {
   bidRequest = bidRequest || index.getBidRequest(bidResponse);
-  const bidCpmAdjustment = bs.get(bidResponse?.bidderCode || bidRequest?.bidder, 'bidCpmAdjustment');
+
+  const adjustForAlternateBids = bs.get(bidResponse?.adapterCode, 'adjustForAlternateBids');
```
dgirardi, line 8: `adjustFor` reads weird; we're adjusting the bids, not adjusting _for_ the bids.

src/utils/cpm.js
```diff
@@ -4,7 +4,10 @@ import {logError} from '../utils.js';
...
+  const scope = adjustForAlternateBids ? bidResponse?.adapterCode : bidResponse?.bidderCode || bidRequest?.bidder;
```
dgirardi, line 9: this means something else now - if I have a bid with `{adapterCode: "adapter', bidderCode: 'bidder'}`, and this config:

```
bidderSettings = {
    bidder: {
          bidCpmAdjustment: bidderAdjustment,
    },
    adapter: {
          bidCpmAdjustment: adapterAdjustment,
          adjustForAlternateBids: true
    }
}
```

then we are picking `adapterAdjustment` over `bidderAdjustment`; is that the intent? I think it's surprising behavior.

src/adapters/bidderFactory.js
```diff
@@ -286,12 +286,14 @@ export function newBidder(spec) {
   function isInvalidAlternateBidder(responseBidder, requestBidder) {
     let allowAlternateBidderCodes = bidderSettings.get(requestBidder, 'allowAlternateBidderCodes') || false;
     let alternateBiddersList = bidderSettings.get(requestBidder, 'allowedAlternateBidderCodes');
+    let adjustForAlternateBids = bidderSettings.get(requestBidder, 'adjustForAlternateBids') || false;
     if (!!responseBidder && !!requestBidder && requestBidder !== responseBidder) {
       alternateBiddersList = isArray(alternateBiddersList) ? alternateBiddersList.map(val => val.trim().toLowerCase()).filter(val => !!val).filter(uniques) : alternateBiddersList;
-      if (!allowAlternateBidderCodes || (isArray(alternateBiddersList) && (alternateBiddersList[0] !== '*' && !alternateBiddersList.includes(responseBidder)))) {
+      if ((!adjustForAlternateBids && !allowAlternateBidderCodes) || !allowAlternateBidderCodes || (isArray(alternateBiddersList) && (alternateBiddersList[0] !== '*' && !alternateBiddersList.includes(responseBidder)))) {
```
dgirardi, line 292: what's this for - why would `isInvalidAlternateBidder` care about adjustments?

src/utils/cpm.js
jlquaccia: changed the var name to be just `adjustAlternateBids`

src/utils/cpm.js
jlquaccia: Yes, believe that is the intent from https://github.com/prebid/Prebid.js/issues/8312 (or my interpretation at least), to provide Publisher's with a config option to handle a bidAdjustment by adapter.  Do you think this would be a feature that we would want to merge in?  Or would we want to pass on this one?

cc @patmmccann

src/adapters/bidderFactory.js
jlquaccia: See what you mean, reverted my change for this line.

src/utils/cpm.js
dgirardi: discussed offline - we agreed that it makes sense to pick the more specific adjustment funcition; e.g., if we find an adjustment for `bidder`, we should pick it over the adapter's adjustment funcition even if the adapter says `adjustForAlternateBids`.

patmmccann: discussed briefly, if a publisher has a specific adjustment for 'friend_of_pubmatic' than it should be used regardless of flag adjustForAlternateBids, and we should fall back to that flag applying when there is no function match

jlquaccia: thanks guys!  makes sense with what you both are saying.  Adjusted the logic just now with a new commit.

dgirardi (approved):

jlquaccia:
Hey @patmmccann, could you also review this one when you have a chance?

patmmccann (approved):

PR #8293, opened 2022-04-18
Build system: do not allow cross-module imports

dgirardi:
## Type of change

- [x] Build related changes

## Description of change

This updates the `validate-imports` eslint plugin to fix an edge case where cross-module imports would sometimes be allowed if the imported module is a directory.

PR #9617, opened 2023-03-02
Multiple Adapters : removing cross-module imports

Justin-Quinn51:
… it within corresponding functions. Added references to 'userIdAsEids' property as an alternative

## Type of change

- [ ] Bugfix
- [ ] Feature
- [ ] New bidder adapter   
- [x] Build related changes
- [ ] CI related changes

- [ ] Other

## Description of change

This updates the validate-imports eslint plugin to fix an edge case where cross-module imports would sometimes be allowed if the imported module is a directory

## Other information

Issue #8293
@patmmccann
@bretg 
@dgirardi 
@ChrisHuie


patmmccann:
Hi @Justin-Quinn51 ; all the formatting changes makes the changeset very hard to determine. Please resubmit.

ChrisHuie (approved):

modules/bluebillywigBidAdapter.js
```diff
@@ -52,10 +51,10 @@ const BB_HELPERS = {
     else if (Array.isArray(adServerCur) && adServerCur.length) request.cur = [adServerCur[0]];
   },
   addUserIds: function(request, validBidRequests) {
-    const bidUserId = deepAccess(validBidRequests, '0.userId');
-    const eids = createEidsArray(bidUserId);
+    const bidUserId = deepAccess(validBidRequests, '0.userIdAsEids');
+    const eids = bidUserId;
```
dgirardi, line 55: this line is doing nothing except renaming; you can move the name up (`const eids = deepAccess(...`)

modules/impactifyBidAdapter.js
```diff
@@ -63,8 +62,8 @@ const createOpenRtbRequest = (validBidRequests, bidderRequest) => {
   if (schain) request.source.ext = { schain: schain };
 
   // Set eids
-  let bidUserId = deepAccess(validBidRequests, '0.userId');
-  let eids = createEidsArray(bidUserId);
+  let bidUserId = deepAccess(validBidRequests, '0.userIdAsEids');
+  let eids = bidUserId;
```
dgirardi, line 66: Same as above - no need to have two names for the same thing.

modules/yieldmoBidAdapter.js
```diff
@@ -370,7 +369,7 @@ function getPageDescription() {
  * @returns an id if there is one, or undefined
  */
 function getId(request, idType) {
-  return (typeof deepAccess(request, 'userId') === 'object') ? request.userId[idType] : undefined;
+  return (typeof deepAccess(request, 'userIdAsEids') === 'object') ? request.userIdAsEids[idType] : undefined;
```
dgirardi, line 372: what was your reason for modifying this? as it is, it cannot be correct:

`request.userId` is a map; this method (before your change) returns one item from that map;
`request.userIdAsEids` is an array; this method now attempts to treat it as a map to look for an item in it. After your change, this method will *always* return undefined.

modules/adyoulikeBidAdapter.js
```diff
@@ -122,8 +121,8 @@ export const spec = {
       payload.ortb2 = bidderRequest.ortb2;
     }
 
-    if (deepAccess(bidderRequest, 'userId')) {
-      payload.userId = createEidsArray(bidderRequest.userId);
+    if (deepAccess(bidderRequest, 'userIdAsEids')) {
```
dgirardi, line 124: @guiann FYI - I believe this is currently broken and you're never getting `userId` from your adapter. This is because `userId` / `userIdAsEids` are *not* properties of `bidderRequest`, but of individual bidRequest objects (`bidRequests[].userIdAsEids` in this scope).

test/spec/modules/yieldmoBidAdapter_spec.js
```diff
@@ -218,7 +218,7 @@ describe('YieldmoAdapter', function () {
 
       it('should add pubcid as parameter of request', function () {
         const pubcid = 'c604130c-0144-4b63-9bf2-c2bd8c8d86da2';
-        const pubcidBid = mockBannerBid({crumbs: undefined, userId: {pubcid}});
+        const pubcidBid = mockBannerBid({crumbs: undefined, userIdAsEids: {pubcid}});
```
dgirardi, line 221: several of the changes here are incorrect for the same reason the yieldmo `getId` change was - at runtime, `userIdAsEids` is never a map.

dgirardi (changes requested):

modules/adyoulikeBidAdapter.js
guiann: Thank you, we'll have a look and fix it in a future PR.

modules/impactifyBidAdapter.js
```diff
@@ -63,8 +62,7 @@ const createOpenRtbRequest = (validBidRequests, bidderRequest) => {
   if (schain) request.source.ext = { schain: schain };
 
   // Set eids
-  let bidUserId = deepAccess(validBidRequests, '0.userId');
-  let eids = createEidsArray(bidUserId);
+  let eids = deepAccess(validBidRequests, '0.userIdAsEids');
   if (eids.length) {
```
dgirardi, line 66: this will throw an exception if eids is null or undefined.

modules/ttdBidAdapter.js
```diff
@@ -121,12 +120,12 @@ function getUser(bidderRequest) {
     utils.deepSetValue(user, 'ext.consent', bidderRequest.gdprConsent.consentString);
   }
 
-  if (utils.isStr(utils.deepAccess(bidderRequest, 'bids.0.userId.tdid'))) {
-    user.buyeruid = bidderRequest.bids[0].userId.tdid;
+  if (utils.isStr(utils.deepAccess(bidderRequest, 'bids.0.userIdAsEids.tdid'))) {
```
dgirardi, line 123: I don't think you want to touch this. `userIdAsEids`, when it exists, is an array; `userIdAsEids.tdid` will always be undefined, because arrays do not have a `tdid` property.

dgirardi (approved):

patmmccann (approved):

PR #9630, opened 2023-03-07
Permutive RTD Module: migrate appnexus to ortb2

AntonioGargaro:
## Type of change

- [ ] Bugfix
- [x] Feature
- [ ] New bidder adapter   
- [ ] Build related changes
- [ ] CI related changes

- [ ] Other

## Description of change

Adds additional keywords to ortb2 that enables us to migrate away from appnexus' special handlings.

## Other information

In a recent https://github.com/prebid/Prebid.js/pull/9236#discussion_r1043632338, our special bid.params handlings were put on notice of being removed if we didn't address the issue by the end of Q2. This PR migrates `appnexus` handlers We've coordinated with each partner on the best place to make data available in ortb2 that each bidder expects.


modules/permutiveRtdProvider.js
```diff
@@ -251,17 +273,6 @@ function getDefaultBidderFn (bidder) {
     return [...new Set([...ac, ...ssp])]
   }
   const bidderMap = {
-    appnexus: function (bid, data, acEnabled) {
```
patmmccann, line 254: does this make the aliasMap code above unnecessary?

ChrisHuie (approved):

PR #9632, opened 2023-03-07
PBjs Core : filter bidders from config when not server-side

r-schweitzer:
## Type of change

- [x] Bugfix
- [ ] Feature
- [ ] New bidder adapter   
- [ ] Build related changes
- [ ] CI related changes

- [ ] Other

## Description of change
Filter bidders from config when not they're not server-side. This was causing the site codes object to be duplicated in the rubicon bidder params as well as the global object.

### DIFF
--- a/modules/prebidServerBidAdapter/ortbConverter.js
+++ b/modules/prebidServerBidAdapter/ortbConverter.js
@@ -169,7 +169,8 @@ const PBS_CONVERTER = ortbConverter({
 
         const fpdConfigs = Object.entries(context.s2sBidRequest.ortb2Fragments?.bidder || {}).filter(([bidder]) => {
           const bidders = context.s2sBidRequest.s2sConfig.bidders;
-          return bidders && bidders.includes(bidder);
+          const allowUnknownBidderCodes = context.s2sBidRequest.s2sConfig.allowUnknownBidderCodes;
+          return allowUnknownBidderCodes || (bidders && bidders.includes(bidder));
         }).map(([bidder, ortb2]) => ({
           bidders: [bidder],
           config: {ortb2}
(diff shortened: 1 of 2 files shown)


dgirardi:
This trips a unit test; it's also not obvious to me that it's a bugfix rather than a breaking change. If I `allowUnknownBidderCodes`, should I expect them to get my client-side configuration? Or is that what you mean by:

> Filter bidders from config when not they're not server-side. This was causing the site codes object to be duplicated in the rubicon bidder params as well as the global object.

@bretg?

bretg:
@r-schweitzer - what does "site codes object" mean here? Would you have an example handy?

bretg:
> if I set aliasX -> bidderX, and set up aliasX for s2s; will server read ext.prebid.bidderconfig.bidderX (given it knows about the alias through ext.prebid.aliases?)

No - config for aliasX is separate from config for bidderX.

> should ext.prebid.bidderconfig continue to contain all bidder-specific configuration, or just the subset for s2sConfig.bidders

I _think_ 'the subset for s2sConfig.bidders' should be fine, but this area is sort of a hairball (at least in my mind) and am barely 80% confident this isn't going to break something for someone.

I might not be following how rubicon is ending up with "site codes" in both global and bidder-specific params -- is this because the rubicon adapter is using the ortb2 library for video? I can see that duplicating in this scenario, but seems to me the fix should be in the rubicon adapter proper. Would you agree @musikele / @robertrmartinez ?

r-schweitzer:
Sorry, to be more clear, what I'm referring to is the fact that client side bidder's configs are included in the bidder configs we send through Prebid server which are also sent client side, which isn't necessary. 

As for allowUnknownBidderCodes it should still send the bidder config for all server.

image

I'm gonna have to look into the unit test and see whats going on. Currently having some issues running unit tests in prebid

r-schweitzer:
Alright so after looking at the unit test, this actually confirms our code is working as it should because rubicon is client side in this test, we wouldn't want to pass the bidder config info for them in the server request, only appnexus.

ncolletti:
@r-schweitzer and I discussed making an update to this PR to support `allowUnknownBidderCodes`

ncolletti (approved):
lgtm

patmmccann (approved):

PR #9633, opened 2023-03-07
33Across ID System: incorporate additional EIDs that could be returned along with 33acrossId

carlosfelix:
## Type of change
- [ ] Bugfix
- [X] Feature
- [ ] New bidder adapter 
- [ ] Build related changes
- [ ] CI related changes
- [ ] Other

## Description of change
Support for additional EIDs in the response from 33across, store and decode those IDs from the 33Across ID system, return those IDs in order to make them accessible from the adapters.

## Other information
Person responsible for reviewing changes: @jdwieland8282


patmmccann:
tagging as do not merge until we discuss in @jdwieland8282 's committee

patmmccann:
we chatted about this in committee, documented some concerns in #9730 we hope you can affirm are covered before merging. Thanks!

patmmccann:
Hi @carlosfelix ; looking for your comments on the three concerns in #9730 .

Liveintent was able to successfully address those in their PR's without requiring 9730 to be solved before merge. Hoping you can do the same.

Issue #4494, opened 2019-11-22
Feature Request: Support for loading the ad but not firing the billing notifications


patmmccann:
## Type of issue
Feature Request

## Description
Many publishers can now pre-fetch ads in a Lazy Load scenario before rendering them. https://developers.google.com/doubleclick-gpt/reference#googletag.PubAdsService_enableLazyLoad
The goal of this is so that the render can be quicker and viewability is improved. 

In the case of an ad in an exit intent pop-up, conducting an auction and rendering an ad before the user exits is quite difficult.  

In prebid, if a publisher wanted the browser to start fetching an ad before the publisher is ready to show it, it would just call the adm. However, if the publisher does this too early, they will have low viewability on the unit. 

However, if an adapter were to separate billing events from the adm, as per section F of OpenRTB 3.0 updates at https://iabtechlab.com/wp-content/uploads/2017/09/OpenRTB-3.0-Draft-Framework-for-Public-Comment.pdf and  https://github.com/InteractiveAdvertisingBureau/openrtb/blob/master/OpenRTB%20v3.0%20FINAL.md  the browser could begin to fetch resources for the ad before the billing events were fired.

We could then write separate fetch and render functions and only fire the billing events on render. Another scenario is the billing events only fire on in view impressions.

A final scenario is a publisher could fetch an ad before calling dfp and render if dfp chose the ad in its response. A publisher may want to do this for all bids over a certain threshold, or other scenarios where the publisher predicts prebid is likely to be the chosen dfp order. 

Also, #2246 seems to recognize burl, but doesn't seem to add support for calling fetch methods before render methods; so I am not sure how a publisher can take advantage of the burl information now.

patmmccann:
From https://github.com/InteractiveAdvertisingBureau/openrtb/blob/master/OpenRTB%20v3.0%20FINAL.md   

burl | string; recommended | Billing notice URL called by the exchange when a winning bid becomes billable based on exchange-specific business policy (e.g., markup rendered). Substitution macros may be included.

patmmccann:
Also, https://github.com/prebid/Prebid.js/pull/2246 seems to recognize to burl, but doesn't seem to add support for calling fetch methods before render methods; so I am not sure how a publisher can take advantage of the burl information now.

bretg:
This is a big project. Getting bidders to split the body and the tracking could take a while.

What would the 'fetch' actually do? I get that the goal is having assets in browser cache, but it's unclear to me how that's feasible.

[...]

PR #9640, opened 2023-03-08
Prebid Core: Functionality to Optionally Defer Billing for an Ad

jlquaccia:
## Type of change
- [x] Feature

## Description of change
- Added the functionality for a Publisher to optionally defer billing on a specific ad unit if desired

Logic follows the proposal outlined here:

1. bid adapters can provide a new method `onBidBillable(bid)` which will be invoked by core when it deems a bid to be billable;
2. when a bid wins, it is by default also billable; that is, by default, core will invoke `onBidWon` and `onBidBillable` one after the other;
3. publisher can flag adUnits as being separately billable: pbjs.addAdUnits({deferBilling: true, ...})
4. winning bids for those adUnits trigger `onBidWon` but *not* `onBidBillable`;
5. when appropriate (e.g. interstitial is displayed), the publisher calls a new public API, pbjs.triggerBilling({adId}), which would trigger `onBidBillable`.

Note: For this feature to fully be enabled, the following is also required:
- The bid adapter a Publisher utilizes must support the `onBidBillable` method
- A Publisher would have to manually invoke the `pbjs.triggerBilling` when desired

## Other information
Related Issue: https://github.com/prebid/Prebid.js/issues/4494

### DIFF
--- a/src/adapterManager.js
+++ b/src/adapterManager.js
@@ -609,6 +609,12 @@ adapterManager.callBidWonBidder = function(bidder, bid, adUnits) {
   tryCallBidderMethod(bidder, 'onBidWon', bid);
 };
 
+adapterManager.callBidBillableBidder = function(bidder, bid, adUnits) {
+  // Adding user configured params to bidWon event data
+  bid.params = getUserConfiguredParams(adUnits, bid.adUnitCode, bid.bidder);
+  tryCallBidderMethod(bidder, 'onBidBillable', bid);
+};
+
 adapterManager.callSetTargetingBidder = function(bidder, bid) {
   tryCallBidderMethod(bidder, 'onSetTargeting', bid);
 };
--- a/src/adapterManager.js
+++ b/src/adapterManager.js
@@ -610,7 +610,7 @@ adapterManager.callBidWonBidder = function(bidder, bid, adUnits) {
 };
 
 adapterManager.callBidBillableBidder = function(bid) {
-  tryCallBidderMethod(bid.adapterCode || bid.bidder, 'onBidBillable', bid);
+  tryCallBidderMethod(bid.bidder, 'onBidBillable', bid);
 };
 
 adapterManager.callSetTargetingBidder = function(bidder, bid) {
--- a/src/adapterManager.js
+++ b/src/adapterManager.js
@@ -609,10 +609,8 @@ adapterManager.callBidWonBidder = function(bidder, bid, adUnits) {
   tryCallBidderMethod(bidder, 'onBidWon', bid);
 };
 
-adapterManager.callBidBillableBidder = function(bidder, bid, adUnits) {
-  // Adding user configured params to bidWon event data
-  bid.params = getUserConfiguredParams(adUnits, bid.adUnitCode, bid.bidder);
-  tryCallBidderMethod(bidder, 'onBidBillable', bid);
+adapterManager.callBidBillableBidder = function(bid) {
+  tryCallBidderMethod(bid.adapterCode || bid.bidder, 'onBidBillable', bid);
 };
 
 adapterManager.callSetTargetingBidder = function(bidder, bid) {
--- a/src/auction.js
+++ b/src/auction.js
@@ -367,8 +367,7 @@ export function newAuction({adUnits, adUnitCodes, callback, cbTimeout, labels, a
   }
 
   function addWinningBid(winningBid) {
-    const { transactionId } = winningBid;
-    const winningAd = adUnits.find(adUnit => adUnit.transactionId === transactionId);
+    const winningAd = adUnits.find(adUnit => adUnit.code === winningBid.adUnitCode);
     _winningBids = _winningBids.concat(winningBid);
     adapterManager.callBidWonBidder(winningBid.adapterCode || winningBid.bidder, winningBid, adUnits);
     if (winningAd && !winningAd.deferBilling) adapterManager.callBidBillableBidder(winningBid);
--- a/src/auction.js
+++ b/src/auction.js
@@ -367,6 +367,10 @@ export function newAuction({adUnits, adUnitCodes, callback, cbTimeout, labels, a
   }
 
   function addWinningBid(winningBid) {
+    // // eslint-disable-next-line no-console
+    // console.log({ winningBid });
+    // // eslint-disable-next-line no-console
+    // console.log({ adUnits });
     const { transactionId } = winningBid;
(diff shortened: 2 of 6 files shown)


src/adapterManager.js
```diff
@@ -616,6 +616,12 @@ adapterManager.callBidWonBidder = function(bidder, bid, adUnits) {
   tryCallBidderMethod(bidder, 'onBidWon', bid);
 };
 
+adapterManager.callBidBillableBidder = function(bidder, bid, adUnits) {
+  // Adding user configured params to bidWon event data
+  bid.params = getUserConfiguredParams(adUnits, bid.adUnitCode, bid.bidder);
```
dgirardi, line 621: I don't think this is necessary. It's already happened in bidWon, and even there I'd question it.

src/auction.js
```diff
@@ -363,9 +363,12 @@ export function newAuction({adUnits, adUnitCodes, callback, cbTimeout, labels, a
     }
   }
 
-  function addWinningBid(winningBid) {
+  function addWinningBid(winningBid, aUnits = adUnits, am = adapterManager) {
+    const { adUnitCode } = winningBid;
+    const winningAd = aUnits.find(adUnit => adUnit.code === adUnitCode);
```
dgirardi, line 368: I think `transactionId` is a better ID than `code`. Multiple adUnits can have the same code.

src/prebid.js
```diff
@@ -1090,4 +1090,19 @@ $$PREBID_GLOBAL$$.processQueue = function () {
   processQueue($$PREBID_GLOBAL$$.cmd);
 };
 
+/**
+ * @alias module:pbjs.triggerBilling
+ */
+$$PREBID_GLOBAL$$.triggerBilling = (bidder, bid, adUnits, am = adapterManager) => {
```
dgirardi, line 1096: this is a public method, so its interface should be kept clean - I don't think you need anything more than `bid`.

src/auction.js
```diff
@@ -363,9 +363,12 @@ export function newAuction({adUnits, adUnitCodes, callback, cbTimeout, labels, a
     }
   }
 
-  function addWinningBid(winningBid) {
+  function addWinningBid(winningBid, aUnits = adUnits, am = adapterManager) {
```
dgirardi, line 366: I think this is overdoing dependency injection - `adUnits` is already something you can pass in directly to the constructor; `adapterManger` could also be if you want it injected, it doesn't make a lot of sense to do it in just one method - the next guy will need to reason about an `auction` with tightly coupled methods but completely disjointed depedendencies.

jlquaccia:
thanks @dgirardi, went back to address your feedback and also refactored a bit.  let me know if anything else is needed.

src/prebid.js
```diff
@@ -1090,4 +1090,19 @@ $$PREBID_GLOBAL$$.processQueue = function () {
   processQueue($$PREBID_GLOBAL$$.cmd);
 };
 
+/**
+ * @alias module:pbjs.triggerBilling
+ */
+$$PREBID_GLOBAL$$.triggerBilling = (bid) => {
```
dgirardi, line 1096: what do you think about extracting [this piece of code](https://github.com/prebid/Prebid.js/blob/1dfae09702db70356ca373e085d578e0fac3314a/src/prebid.js#L998-L1013) and reusing it here? 

what I am worried about is that this call comes from the publisher and here `bid` could be anything at all; *but* the adapters will expect it to be an actual certified bid response object.

I would also improve that code a bit to look up by `bidId` as well. I don't like that it apparently accepts finding multiple matching bids, and just decides that the first one is the "correct" one.

jlquaccia: hey @dgirardi, i agree, makes sense to extract that piece of code and reuse it in the triggerBilling func, especially to your point about how a pub could pass through anything.

for the second part of your comment, if/when the `bids` var gets defined, i added an additional check to make sure that the bid we grab from that array has a matching transactionId with the bid that was passed through the triggerBilling func.  should that be sufficient?

jlquaccia:
@dgirardi do you think anything else is needed for this PR?

src/auction.js
```diff
@@ -367,8 +367,11 @@ export function newAuction({adUnits, adUnitCodes, callback, cbTimeout, labels, a
   }
 
   function addWinningBid(winningBid) {
+    const { transactionId } = winningBid;
```
dgirardi, line 370: I don't think this helps; by the time it gets here, it should already be the correct bid. The filtering happens in `markWinningBidAsUsed`.

src/prebid.js
```diff
@@ -1092,4 +1092,33 @@ pbjsInstance.processQueue = function () {
...
+$$PREBID_GLOBAL$$.triggerBilling = (winningBid) => {
+  let bids = [];
+
+  if (winningBid.adUnitCode && winningBid.adId) {
+    bids = auctionManager.getBidsReceived()
+      .filter(bid => bid.adId === winningBid.adId && bid.adUnitCode === winningBid.adUnitCode);
+  } else if (winningBid.adUnitCode) {
+    bids = targeting.getWinningBids(winningBid.adUnitCode);
+  } else if (winningBid.adId) {
+    bids = auctionManager.getBidsReceived().filter(bid => bid.adId === winningBid.adId);
```
dgirardi, line 1107: this could be extracted no? as in, re-use code instead of copy-paste.

src/auction.js
jlquaccia: I think for this part I was just aiming to link the winning bid with it's associated adunit to check if deferBilling is true or false.  Looks like that can be done by either comparing `adUnitCode` & `code` between the two or `transactionId`.  Or let me know if there is another way to do this that I may have missed.

src/prebid.js
```diff
@@ -1092,4 +1092,33 @@ pbjsInstance.processQueue = function () {
...
+  } else {
+    logWarn('Improper use of triggerBilling. It requires a bid with at least an adUnitCode or an adId to function.');
+  }
+
+  if (bids.length > 0) {
+    const triggerBillingBid = bids.find(bid => bid.transactionId === winningBid.transactionId);
```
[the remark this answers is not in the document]
jlquaccia: Addressed this as well.  Had one basic question though, I didn't see `bidId` on the winning bids but did see `requestId` which looks to have the same value.  Is it safe to you that?

jlquaccia:
thanks @dgirardi, the feedback you gave makes sense.  just addressed the changes you pointed out with a new commit.

dgirardi (approved):

patmmccann (approved):

bretg:
Docs PR?

Issue #8358, opened 2022-05-02
Consolidate Dual render ad paths


patmmccann:
## Type of issue
Two areas of the code base can render an ad

https://github.com/prebid/Prebid.js/blob/880b089bf66701d19a6284a61847fb05a059b270/src/secureCreatives.js#L56
and
https://github.com/prebid/Prebid.js/blob/2d14b7d53337c47b2e804f184927ed498990cc8f/src/prebid.js#L447

We've had various problems with these getting out of sync, including https://github.com/prebid/Prebid.js/issues/7702 and potentially recording analytics events eg bid_won. 

It seems best if we consolidate them, potentially by removing the latter and always forcing the post message into an iframe with the creative in it.

GLStephen:
I think the first path is for SafeFrames and the second is for non-safe frames.

dgirardi:
It's unclear to me how we should deal with some of the differences:

- there is partial, awkward support for CLICKTHROUGH macro substitution ([docs](https://docs.prebid.org/dev-docs/publisher-api-reference/renderAd.html), also see [this issue](https://github.com/prebid/Prebid.js/issues/9060)). It does not work, and with the current interface cannot work, on safeframe.
- there's hair-raising [scraper-targeted markup](https://github.com/prebid/Prebid.js/pull/7874) generated only from `renderAd`, again not from safeframes.
- the new video module has some [ad-hoc support only in renderAd](https://github.com/prebid/Prebid.js/blob/3e2b9fca2ab41eb8280925c27b5d970ca43cd6f9/src/prebid.js#L536). I think this might become a "normal" renderer (@karimMourra?)

Issue #9060, opened 2022-09-30
`${CLICKTHROUGH}` macro is not documented, and does not work with safeframe


dgirardi:
## Type of issue
bug?

## Description

https://github.com/prebid/Prebid.js/pull/5796 introduced the `${CLICKTROUGH}` macro for ad markup, to be substituted with the value passed as an option to `renderAd`. Both the macro and the option are undocumented, and it appears that there is no way to make it work with safeframe (since it does not go through `renderAd`).

patmmccann:
partial dupe of #6130

jlquaccia:
Created a PR for this to prebid.github.io here: https://github.com/prebid/prebid.github.io/pull/4176

Issue #6130, opened 2020-12-14
Prebid Click Tracking


AStephenson:
How can I track prebid clicks in GAM? The bidder is AppNexus. Creative is below, but I am unsure where to put the click tracker in the creative tag.

<script src = "https://cdn.jsdelivr.net/npm/prebid-universal-creative@latest/dist/creative.js"></script>
    <script>
      var ucTagData = {};
      ucTagData.adServerDomain = "";
      ucTagData.pubUrl = "%%PATTERN:url%%";
      ucTagData.adId = "%%PATTERN:hb_adid_appnexus%%";
      ucTagData.cacheHost = "%%PATTERN:hb_cache_host_appnexus%%";
      ucTagData.cachePath = "%%PATTERN:hb_cache_path_appnexus%%";
      ucTagData.uuid = "%%PATTERN:hb_cache_id_appnexus%%";
      ucTagData.mediaType = "%%PATTERN:hb_format_appnexus%%";
      ucTagData.env = "%%PATTERN:hb_env%%";
      ucTagData.size = "%%PATTERN:hb_size_appnexus%%";
      ucTagData.hbPb = "%%PATTERN:hb_pb_appnexus%%";
      // mobileResize needed for mobile GAM only
      ucTagData.mobileResize = "hb_size:%%PATTERN:hb_size_appnexus%%";
      try {
        ucTag.renderAd(document, ucTagData);
      } catch (e) {
        console.log(e);
      }
    </script>

bretg:
Someone with javascript chops might be able to add some code that listens for click events?

Maybe set a variable like
```
var gamClickTracker="%%CLICK_URL_UNESC%%"; // this will be resolved by GAM

listenToClickEvents() {
   fire gamClickTracker;
}
```

bretg:
Do you just want the clicks tracked in GAM, in the bidder reports, or in a Prebid Analytics report?

AStephenson:
Tracked in GAM. Thanks!

patmmccann:
I don't think that works @bretg as I don't think there is an event to listen to

I think #5796 is more likely to succeed,  but requires the ssp to deliver specific markup

bretg:
I was thinking something like this -- https://stackoverflow.com/questions/2381336/detect-click-into-iframe-using-javascript/32138108#32138108

Prebid creatives are wrapped in an iframe, so maybe there's something we can do to detect clicks on that iframe without affecting the rest of the page?

[...]

PR #9647, opened 2023-03-09
Core: unify rendering paths; codify (and test) cross-domain render example

dgirardi:
## Type of change

- [x] Feature

## Description of change

This attempts to unify `pbjs.renderAd` and "safe creative" rendering so that they share as much code as possible.
Additionally, this adds `build/creative/creative.js`, an optimized (and tested) version of [this cross-domain example](https://github.com/prebid/Prebid.js/blob/master/integrationExamples/gpt/x-domain/creative.html). The plan is to eventually re-use this from PUC or its successor.

## Other information

Closes https://github.com/prebid/Prebid.js/issues/8358
Also related:
 https://github.com/prebid/Prebid.js/issues/6130
 https://github.com/prebid/Prebid.js/issues/9060


libraries/creativeRender/writer.js
```diff
@@ -0,0 +1,34 @@
...
+
+export function writeAd({ad, adUrl, width, height}, cb, doc = document) {
+  if (!ad && !adUrl) {
+    // eslint-disable-next-line standard/no-callback-literal
+    cb({reason: NO_AD, message: 'Missing ad markup or URL'});
+  } else {
+    if (adUrl && !ad) {
+      mkFrame(doc, {width, height, src: adUrl})
+    } else {
+      doc.write(ad);
```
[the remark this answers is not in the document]
dgirardi: Tempted to mark this as false positive (@patmmccann ?)

This is what's protecting us from XSS: https://github.com/prebid/Prebid.js/blob/5cfce0340139f816487b00f93da57d71f7538b54/libraries/creativeRender/crossDomain.js#L34

(an attacker needs to know the adId that the creative requested. If they do, they can fake the ad with whatever markup they wish).

ChrisHuie (approved):

musikele (commented):
Some comments here and there, you may want to give a look

libraries/creativeRender/writer.js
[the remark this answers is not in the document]
patmmccann: Agreed

patmmccann:
@musikele forgive me for having trouble following, have you finished your review and awaiting changes, or are your changes addressed?

src/secureCreatives.js
```diff
@@ -69,39 +67,13 @@ export function receiveMessage(ev) {
...
-    return;
-  }
-
-  // save winning bids
-  auctionManager.addWinningBid(adObject);
-
-  events.emit(BID_WON, adObject);
+function handleRenderRequest(reply, message, bidResponse) {
+  handleRender(function (adData) {
+    resizeRemoteCreative(bidResponse);
```
mmoschovas, line 72: Would this resize for outstream renders? My assumption here is that it could if the bid has video dimensions defined as the width/height and could cause white space, but perhaps I am overlooking something

dgirardi: video doesn't go through this, both `pbjs.renderAd` and PUC throw an error if they're asked to render a video bid - since they don't know the player to use or how to interact with it. The flow for it is very different and I think it wouldn't result in a GPT frame unless you set up the player in your creative. 
Our guides for it do not use GPT at all and instead generate a VAST url that the pub is expected to submit to their player (or the new-ish video module can do it for them, but still without a separate creative iframe that could send messages here).

dgirardi: I think I misunderstood your question - if an outstream renderer is in use and the frame here is therefore useless, are we resizing it to be large and empty? I will do some testing and report back.

dgirardi: The answer is no, when bids have a renderer this code path is skipped - I added a test case for it.

mmoschovas: Ah yep I see. Thanks for confirming!

mmoschovas (approved):
lgtm

EskelCz:
@dgirardi Does this PR also solve the native template rendering through `renderAd`? We discussed it some time ago, here: https://github.com/prebid/Prebid.js/issues/10006#issuecomment-1567244132

dgirardi:
@EskelCz no, but https://github.com/prebid/Prebid.js/pull/10819 does - and it would be great to get real world feedback on it!

PR #9649, opened 2023-03-09
Js 1763 add intent iqid

About7Sharks:
## Type of change

- [x] Feature

## Description of change

 Sonobi adapter will mint a new intent iq id **Only when storageAllowed is true.** We will send this in the /trinity.json request as a separate query param called “iqid”. The value of this will be a non-encrypted intent iq user id string.

### DIFF
--- a/modules/sonobiBidAdapter.js
+++ b/modules/sonobiBidAdapter.js
@@ -440,7 +440,7 @@ function loadOrCreateFirstPartyData() {
         }
       }
     } catch (error) {
-
+      return;
     }
   };
   var firstPartyData = tryParse(readData(FIRST_PARTY_KEY));
--- a/modules/sonobiBidAdapter.js
+++ b/modules/sonobiBidAdapter.js
@@ -4,6 +4,7 @@ import { BANNER, VIDEO } from '../src/mediaTypes.js';
 import { config } from '../src/config.js';
 import { Renderer } from '../src/Renderer.js';
 import { userSync } from '../src/userSync.js';
+import { bidderSettings } from '../src/bidderSettings.js';
 const BIDDER_CODE = 'sonobi';
 const STR_ENDPOINT = 'https://apex.go.sonobi.com/trinity.json';
 const PAGEVIEW_ID = generateUUID();
@@ -49,6 +50,7 @@ export const spec = {
 
     return true;
   },
+
   /**
    * Make a server request from the list of BidRequests.
    *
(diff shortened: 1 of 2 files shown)


test/spec/modules/sonobiBidAdapter_spec.js
```diff
@@ -398,6 +405,7 @@ describe('SonobiBidAdapter', function () {
       expect(bidRequests.data.ref).not.to.be.empty
       expect(bidRequests.data.s).not.to.be.empty
       expect(bidRequests.data.pv).to.equal(bidRequestsPageViewID.data.pv)
+      expect(bidRequests.data.iqid).to.match(/^[0-9a-f]{8}-[0-9a-f]{4}-4[0-9a-f]{3}-[89ab][0-9a-f]{3}-[0-9a-f]{12}$/)
```
spotxslagle, line 408: Should the localStorage aspects be tested?

modules/sonobiBidAdapter.js
```diff
@@ -388,6 +390,69 @@ export function _getPlatform(context = window) {
...
+      }
+    } catch (error) {
+
+    }
+  };
+  var firstPartyData = tryParse(readData(FIRST_PARTY_KEY));
+  if (!firstPartyData || !firstPartyData.pcid) {
+    var firstPartyId = generateGUID();
+    firstPartyData = { pcid: firstPartyId, pcidDate: Date.now() };
+  } else if (firstPartyData && !firstPartyData.pcidDate) {
```
spotxslagle, line 450: firstPartyData will always be defined by this point and doesn't need to be checked

modules/sonobiBidAdapter.js
```diff
@@ -388,6 +390,69 @@ export function _getPlatform(context = window) {
...
+    var d = new Date().getTime();
+    return 'xxxxxxxx-xxxx-4xxx-yxxx-xxxxxxxxxxxx'.replace(/[xy]/g, function (c) {
+      var r = (d + Math.random() * 16) % 16 | 0;
+      d = Math.floor(d / 16);
+      return (c === 'x' ? r : (r & 0x3) | 0x8).toString(16);
+    });
+  };
+  var storeData = function (key, value) {
+    try {
+      if (typeof key === 'string' && key.startsWith('_iiq_fdata')) {
```
spotxslagle, line 437: This check is unnecessary. storeData is only called once, and always with the FIRST_PARTY_KEY

modules/sonobiBidAdapter.js
```diff
@@ -388,6 +390,69 @@ export function _getPlatform(context = window) {
...
+        if (value && hasLocalStorage()) {
```
spotxslagle, line 438: It is also unnecessary to check if there is a value. The code already checks if firstPartyData is defined, and defines it if not prior to this function being called.

spotxslagle (changes requested):
I have a few suggestions and a question. Otherwise it looks good to me!

spotxslagle (approved):
lgtm

Issue #9335, opened 2022-12-13
Improve FPD enrichment


dgirardi:
## Type of issue
Improvement

## Description

With https://github.com/prebid/Prebid.js/pull/9205 prebid will fill in a number of `ortb2` fields. The logic should be improved by:

 - merging values set through the "old" `setConfig({site, app, device})`
 - not filling in more than one of `site`, `app`, `dooh` (related to https://github.com/prebid/Prebid.js/pull/9258)
 - some values from `site`, such as `publisher` should be moved into `app` (or `dooh`) when appropriate
 - a number of adapters are filling in `device.devicetype`. If the logic can be made general, we should move it into core.

PR #9659, opened 2023-03-13
Core: improve FPD enrichment

dgirardi:
## Type of change

- [x] Feature

## Description of change

This improves FPD enrichment logic:

 - if `dooh` or `app` are provided by the pub, do not attempt to enrich `site`
 - put `keywords`, `domain` and `publisher.domain` in one of `dooh`, `app` or `site`
 - "legacy" setConfig({app, device, site}) are now merged in with "standard" FPD

**Note**: I did not attempt to fill in `device.devicetype`; there are several adapters that attempt to guess it based on the user agent, but that seems too fragile to me, and also not very useful (the other end could also just look at the user agent since that's in the request)

## Other information

Closes https://github.com/prebid/Prebid.js/issues/9335


ChrisHuie (approved):

ncolletti (approved):
looks good!

PR #9662, opened 2023-03-14
pairId userId submodule: initial commit for pairId submodule

congdu-kun:
## Type of change

- [ ] Bugfix
- [x] Feature
- [ ] New bidder adapter   
- [ ] Build related changes
- [ ] CI related changes

- [ ] Other

## Description of change

Add a userId submodule for PAIR Id.
Related blog post about PairId: https://blog.google/products/marketingplatform/360/engage-your-first-party-audience-in-display-video-360/
The initial commit will only try fetching id from local storage/cookie, later version might collaborate with different cleanroom providers to add api call to fetch the IDs.
Will submit a separate PR to prebid.github.io for the documentation.

### DIFF
--- a/modules/pairIdSystem.js
+++ b/modules/pairIdSystem.js
@@ -11,7 +11,7 @@ import { logError } from '../src/utils.js';
 
 const MODULE_NAME = 'pairId';
 const PAIR_ID_KEY = 'pairId';
-const LIVERAMP_PAIR_ID_KEY = '_lr_pairId';
+const DEFAULT_LIVERAMP_PAIR_ID_KEY = '_lr_pairId';
 
 export const storage = getStorageManager()
 
@@ -57,7 +57,8 @@ export const pairIdSubmodule = {
 
     const configParams = (config && config.params) || {};
     if (configParams && configParams.liveramp) {
-      const liverampValue = pairIdFromLocalStorage(LIVERAMP_PAIR_ID_KEY) || pairIdFromCookie(LIVERAMP_PAIR_ID_KEY)
+      let LRStorageLocation = configParams.liveramp.storageKey || DEFAULT_LIVERAMP_PAIR_ID_KEY
+      const liverampValue = pairIdFromLocalStorage(LRStorageLocation) || pairIdFromCookie(LRStorageLocation)
       try {
         const obj = JSON.parse(atob(liverampValue));
         ids = ids.concat(obj.envelope);
--- a/modules/pairIdSystem.js
+++ b/modules/pairIdSystem.js
@@ -7,19 +7,20 @@
 
 import { submodule } from '../src/hook.js';
 import {getStorageManager} from '../src/storageManager.js'
-import { isStr, logError } from '../src/utils.js';
+import { logError } from '../src/utils.js';
 
 const MODULE_NAME = 'pairId';
 const PAIR_ID_KEY = 'pairId';
+const LIVERAMP_PAIR_ID_KEY = 'lr_pairId';
 
 export const storage = getStorageManager()
 
-function pairIdFromLocalStorage() {
-  return storage.localStorageIsEnabled ? storage.getDataFromLocalStorage(PAIR_ID_KEY) : null;
+function pairIdFromLocalStorage(key) {
+  return storage.localStorageIsEnabled ? storage.getDataFromLocalStorage(key) : null;
 }
 
-function pairIdFromCookie() {
-  return storage.cookiesAreEnabled ? storage.getCookie(PAIR_ID_KEY) : null;
+function pairIdFromCookie(key) {
+  return storage.cookiesAreEnabled ? storage.getCookie(key) : null;
 }
 
 /** @type {Submodule} */
@@ -36,22 +37,40 @@ export const pairIdSubmodule = {
    * @returns {{pairId:string} | undefined }
    */
   decode(value) {
-    return value && isStr(value) ? {'pairId': value} : undefined
+    return value && Array.isArray(value) ? {'pairId': value} : undefined
   },
   /**
   * performs action to obtain id and return a value in the callback's response argument
   * @function
   * @returns {id: string | undefined }
   */
-  getId() {
-    const pairIdString = pairIdFromLocalStorage() || pairIdFromCookie()
-
-    if (pairIdString && typeof pairIdString == 'string') {
-      return { id: pairIdString };
+  getId(config) {
+    const pairIdsString = pairIdFromLocalStorage(PAIR_ID_KEY) || pairIdFromCookie(PAIR_ID_KEY)
+    let ids = []
+    if (pairIdsString && typeof pairIdsString == 'string') {
+      try {
+        ids = ids.concat(JSON.parse(atob(pairIdsString)))
+      } catch (error) {
+        logError(error)
+      }
     }
 
-    logError('PairId not found.')
-    return undefined;
+    const configParams = (config && config.params) || {};
+    if (configParams && configParams.liveramp) {
+      const liverampValue = pairIdFromLocalStorage(LIVERAMP_PAIR_ID_KEY) || pairIdFromCookie(LIVERAMP_PAIR_ID_KEY)
+      try {
+        const obj = JSON.parse(atob(liverampValue));
+        ids = ids.concat(obj.envelope);
+      } catch (error) {
+        logError(error)
+      }
+    }
+
+    if (ids.length == 0) {
+      logError('PairId not found.')
+      return undefined;
+    }
+    return {'id': ids};
   }
 };
 
--- a/modules/pairIdSystem.js
+++ b/modules/pairIdSystem.js
@@ -1,0 +1,58 @@
+/**
+ * This module adds PAIR Id to the User ID module
+ * The {@link module:modules/userId} module is required
+ * @module modules/pairIdSystem
+ * @requires module:modules/userId
+ */
+
+import { submodule } from '../src/hook.js';
+import {getStorageManager} from '../src/storageManager.js'
+import { isStr, logError } from '../src/utils.js';
+
+const MODULE_NAME = 'pairId';
+const PAIR_ID_KEY = 'pairId';
+
+export const storage = getStorageManager()
+
+function pairIdFromLocalStorage() {
+  return storage.localStorageIsEnabled ? storage.getDataFromLocalStorage(PAIR_ID_KEY) : null;
+}
+
+function pairIdFromCookie() {
+  return storage.cookiesAreEnabled ? storage.getCookie(PAIR_ID_KEY) : null;
+}
+
+/** @type {Submodule} */
+export const pairIdSubmodule = {
+  /**
+  * used to link submodule with config
+  * @type {string}
+  */
+  name: MODULE_NAME,
+  /**
+   * decode the stored id value for passing to bid requests
+   * @function
+   * @param { string | undefined } value
+   * @returns {{pairId:string} | undefined }
+   */
+  decode(value) {
+    return value && isStr(value) ? {'pairId': value} : undefined
+  },
+  /**
+  * performs action to obtain id and return a value in the callback's response argument
+  * @function
+  * @returns {id: string | undefined }
+  */
+  getId() {
+    const pairIdString = pairIdFromLocalStorage() || pairIdFromCookie()
+
+    if (pairIdString && typeof pairIdString == 'string') {
+      return { id: pairIdString };
+    }
+
+    logError('PairId not found.')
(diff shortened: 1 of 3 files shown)


modules/userId/eids.js
```diff
@@ -45,6 +45,12 @@ export const USER_IDS_CONFIG = {
     atype: 1
   },
 
+  // pairId
+  'pairId': {
+    source: 'google.com',
+    atype: 571187
```
patmmccann, line 51: where did you get this value?

patmmccann:
It seems possible there would be other pairids in other submodules, can rename to LiveRamp PAIRID ?

modules/userId/eids.js
congdu-kun: Hi Patrick, this is the atype value we picked to identify PAIR ID. Since PAIR is still in MVP phase and we only expect pubs/advertisers with their own user management systems to use this feature, at this moment the value is not documented publicly and mostly communicated directly with early adopter partners. I can confirm the value `571187` is correct and is the one will be used in google's ads serving system.
If you are aware of some centralized place for documenting atype, feel free to point it to me. happy to add it there.

congdu-kun:
Hi @patmmccann,
At this moment we are working closely with Liveramp as a cleanroom provider on MVP for PAIR ID. For easy user journey we are adding the logic for fetching from their cookie to the same submodule. So that either publishers choose to handle PAIR IDs themselves (put into `pairId` localstorage/cookie entry themselves) or let Liveramp do it for them, the submodule will work for them.
The potential method going forward might be ask other cleanrooms' to add logic into the same submodule with a seperate config entry in `parid.params`. If that is not feasbile, we will work with liveramp to split their specific logic out, as a liveramp PairID submodule like you suggested. But the basic logic of fetching from `pairID` local storage will still be here, for publishers that choose to maintain IDs themselves.
I will also add more info to the doc in the documentation PR. Thanks!

congdu-kun:
Added documentation here: https://github.com/prebid/prebid.github.io/pull/4462

modules/userId/eids.js
congdu-kun: Hi @patmmccann, communicated with internal team, atype: 571187 is what we have been using in our codebase for PAIR ID.  I also checked the openRTB spec, 500+ are for Vendor-specific codes. So hopefully we should be fine using this value?

patmmccann: this is the place https://github.com/InteractiveAdvertisingBureau/openrtb/blob/master/extensions/2.x_official_extensions/eids.md but your code has nothing to do with agent type, you appear to have hijacked an unrelated field and you should have used ext

patmmccann:
@congdu-kun this is not the only pair id correct? Should this not be called the LR pair id module?

congdu-kun:
> @congdu-kun this is not the only pair id correct? Should this not be called the LR pair id module?

Correct, this is not gonna be the only pair Id. This module is contributed from Google, so we cannot call this LR pair id, although I'm leaving a stub for them to facilitate MVP adopters to use. Maybe I can rename this to google pair id.

patmmccann:
That's super helpful, I thought this was only working with LR, will re-read with that in mind, no need to rename. To set additional PAIR id's, this would have to get PR'd and the storage key specified? Why not just try to achieve all that in config?

modules/userId/eids.js
patmmccann: I thought you worked at liveramp, so above comment was supposed to be google or liveramp used the wrong value, but now it is clear just google made this choice.

patmmccann: btw, I asked IABTL slack to weigh in on this question here https://openrtb-iabtechlab.slack.com/archives/C3Y6GHUTH/p1680635011100839 if you have access

test/spec/modules/pairIdSystem_spec.js
```diff
@@ -0,0 +1,56 @@
...
+  afterEach(() => {
+    sandbox.restore();
+  });
+
+  it('should log an error if no ID is found when getId', function() {
+    pairIdSubmodule.getId({ params: {} });
+    expect(logErrorStub.calledOnce).to.be.true;
+  });
+
+  it('should read pairId from local stroage if exists', function() {
```
patmmccann, line 21: typo

test/spec/modules/pairIdSystem_spec.js
```diff
@@ -0,0 +1,56 @@
...
+
+  it('should read pairId from cookie if exists', function() {
+    let pairIds = ['test-pair-id4', 'test-pair-id5', 'test-pair-id6'];
+    sandbox.stub(storage, 'getCookie').withArgs('pairId').returns(btoa(JSON.stringify(pairIds)));
+
+    let id = pairIdSubmodule.getId({ params: {} });
+    expect(id).to.be.deep.equal({id: pairIds});
+  });
+
+  it('should read pairId from liveramp envelop local storage key if configured', function() {
```
patmmccann, line 37: typo

test/spec/modules/pairIdSystem_spec.js
```diff
@@ -0,0 +1,56 @@
...
+    let pairIds = ['test-pair-id1', 'test-pair-id2', 'test-pair-id3'];
+    sandbox.stub(storage, 'getDataFromLocalStorage').withArgs('lr_pairId').returns(btoa(JSON.stringify({'envelope': pairIds})));
+    let id = pairIdSubmodule.getId({
+      params: {
+        liveramp: {}
+      }})
+    expect(id).to.be.deep.equal({id: pairIds})
+  })
+
+  it('should read pairId from liveramp envelop cookie entry if configured', function() {
```
patmmccann, line 47: typo

modules/userId/eids.js
congdu-kun: I think it depends on how do you interpret `agent type`, especially given that a person-based id (3) is considered an `agent type`, pairId is not a traditional person-based user id as its encrypted and protected combined publisher/adversiter 1p data. We are trying to use this to address various privacy concerns introduced by person-based ID. So we think it deserve a separate code.
Also, 'hijacked' is a strong word. again, given the specification that 500+ are for vendor-specific and we are only using the field for pairId, any secrutiy/privacy vulnerability you are aware of for using `atype` like this? If so, we will have to address that.

modules/pairIdSystem.js
```diff
@@ -0,0 +1,77 @@
...
+      try {
+        ids = ids.concat(JSON.parse(atob(pairIdsString)))
+      } catch (error) {
+        logError(error)
+      }
+    }
+
+    const configParams = (config && config.params) || {};
+    if (configParams && configParams.liveramp) {
+      const liverampValue = pairIdFromLocalStorage(LIVERAMP_PAIR_ID_KEY) || pairIdFromCookie(LIVERAMP_PAIR_ID_KEY)
```
patmmccann, line 60: Okay, on second pass and understanding this is not a liveramp contribution, i believe I understand this a lot better. Can you add the LIVERAMP_PAIR_ID_KEY as a config param  (yet falling back to the default string you have above) and loop over all the configured params?

My goal here is that a publisher would not need to upgrade the js library as support for new pair keys and partners are added, but could instead specify everything in config.

modules/userId/eids.js
patmmccann: I don't mean to imply any malintent with my word choice, just that you've put the value in the wrong place. Many other adapters have chosen to put their custom fields in ext.stype or something else, eg https://github.com/prebid/Prebid.js/blob/8c8a2d7d2aa9202ef0823d27ad7f5a079ac3c284/test/spec/modules/eids_spec.js#L64

If you PR IABTL with your choice and they accept it, I'm happy to merge as is. For now, it appears your value is misleading on agent type and you should have chosen ext.someotherkey

patmmccann: As further background, vendor specific codes are typically enumerated to avoid conflicts, eg https://github.com/InteractiveAdvertisingBureau/openrtb/blob/aa2ce4cfdc568e54cc4b64d9fbdec52b1e3e2654/extensions/community_extensions/segtax.md?plain=1#L233

congdu-kun: Sounds good! Thank you for the pointers Patrick! I will discuss with internal team who own this feature and get back to this thread.

patmmccann:
Summary of current status: 

1) Please switch from atype to ext.someothervalue OR pr the file `https://github.com/InteractiveAdvertisingBureau/openrtb/blob/master/extensions/2.x_official_extensions/eids.md`

2) fix a variety of typos

3) make the liveramp key location configurable

modules/pairIdSystem.js
congdu-kun: I can have LIVERAMP_PAIR_ID_KEY as a config param, but probably won't serve the user much, as it's an agreed on value between us and liveramp, and they are fetching IDs and putting them into LIVERAMP_PAIR_ID_KEY with liveramp scripts they provide to the publishers.

As for supporting more partners through config, I understand this question, but this is a non-goal for us at the moment. Because: 1) we are focusing on enabling MVP customer testing, partnering with liveramp. 2) For the other partners, i'm not sure they will follow the same pattern, probably they would prefer the publishers to do a call to their API endpoint. So other pairId provider will still need to come and provide their logic themselves, like a separate ID provider. We will not dictate how they provide PairId to publishers.

congdu-kun:
Hi Patrick, added a fix:
> 1. Please switch from atype to ext.someothervalue OR pr the file `https://github.com/InteractiveAdvertisingBureau/openrtb/blob/master/extensions/2.x_official_extensions/eids.md`
I think there is a meeting scheduled for next week for some clarification, will hold the PR until then.
> 2. fix a variety of typos
done
> 3. make the storage key location(s) configurable
done, also see the comment above.
> 4. didn't pass your ci tests
fixed.

Thanks

patmmccann:
Do you know when this meeting is scheduled for?

patmmccann:
I see your pr to IABTL

Okay to merge?

modules/pairIdSystem.js
patmmccann: Sounds reasonable

patmmccann (approved):

congdu-kun:
> Do you know when this meeting is scheduled for? You're asking us to hold merge for the outcome correct?
Hi Patrick,

I think the meeting is scheduled for Friday, but i believe it's mainly for clarification and communication. Please feel free to merge if you don't have other questions, otherwise we can hold it for the meeting.

modules/pairIdSystem.js
```diff
@@ -0,0 +1,78 @@
...
+import { logError } from '../src/utils.js';
+
+const MODULE_NAME = 'pairId';
+const PAIR_ID_KEY = 'pairId';
+const DEFAULT_LIVERAMP_PAIR_ID_KEY = '_lr_pairId';
+
+export const storage = getStorageManager()
+
+function pairIdFromLocalStorage(key) {
+  return storage.localStorageIsEnabled ? storage.getDataFromLocalStorage(key) : null;
```
robertrmartinez, line 19: `storage.localStorageIsEnabled` is a function, so this check will always evaluate to true.

This should be updated to `storage.localStorageIsEnabled()` which returns true or false.

Same with `storage.cookiesAreEnabled`

modules/pairIdSystem.js
```diff
@@ -0,0 +1,78 @@
...
+
+    const configParams = (config && config.params) || {};
+    if (configParams && configParams.liveramp) {
+      let LRStorageLocation = configParams.liveramp.storageKey || DEFAULT_LIVERAMP_PAIR_ID_KEY
+      const liverampValue = pairIdFromLocalStorage(LRStorageLocation) || pairIdFromCookie(LRStorageLocation)
+      try {
+        const obj = JSON.parse(atob(liverampValue));
+        ids = ids.concat(obj.envelope);
+      } catch (error) {
+        logError(error)
```
robertrmartinez, line 66: if some other code does not set the pair id or the liveramp id in storage or cookie then this module logs an error every single time. Which seems like overkill.

Same with the check below this, if no ID is found it throws a pretty nasty error:

image


Perhaps these can be changed to WARN or INFO messages instead?

Example of error where pair id is not found:

image


I think an error is a bit too strong here.

robertrmartinez:
@patmmccann @congdu-kun 

Found one issue as well as had a couple comments on this.

Wondering if you could address and / or fix these?

congdu-kun:
Hi @robertrmartinez, Thanks for the comment, will address the above and send a separate PR to fix these!

robertrmartinez:
@congdu-kun There is actually a PR up if you want to give it a review and make sure it looks good!

https://github.com/prebid/Prebid.js/pull/9998

PR #9684, opened 2023-03-17
Axis Bid Adapter : initial release

PyjamaWarrior:
doc - https://github.com/prebid/prebid.github.io/pull/4424

### DIFF
--- a/modules/axisBidAdapter.js
+++ b/modules/axisBidAdapter.js
@@ -1,0 +1,206 @@
+import { isFn, deepAccess, logMessage, logError } from '../src/utils.js';
+import { convertOrtbRequestToProprietaryNative } from '../src/native.js';
+
+import { registerBidder } from '../src/adapters/bidderFactory.js';
+import { BANNER, NATIVE, VIDEO } from '../src/mediaTypes.js';
+import { config } from '../src/config.js';
+
+const BIDDER_CODE = 'axis';
+const AD_URL = 'https://prebid.axis-marketplace.com/pbjs';
+const SYNC_URL = 'https://cs.axis-marketplace.com';
+
+function isBidResponseValid(bid) {
+  if (!bid.requestId || !bid.cpm || !bid.creativeId || !bid.ttl || !bid.currency) {
+    return false;
+  }
+
+  switch (bid.mediaType) {
+    case BANNER:
+      return Boolean(bid.width && bid.height && bid.ad);
+    case VIDEO:
+      return Boolean(bid.vastUrl || bid.vastXml);
+    case NATIVE:
+      return Boolean(bid.native && bid.native.impressionTrackers && bid.native.impressionTrackers.length);
+    default:
+      return false;
+  }
+}
+
+function getPlacementReqData(bid) {
+  const { params, bidId, mediaTypes } = bid;
+  const schain = bid.schain || {};
+  const { integration, token } = params;
+  const bidfloor = getBidFloor(bid);
+
+  const placement = {
+    integration,
+    token,
+    bidId,
+    schain,
+    bidfloor
+  };
+
+  if (mediaTypes && mediaTypes[BANNER]) {
+    placement.adFormat = BANNER;
+    placement.sizes = mediaTypes[BANNER].sizes;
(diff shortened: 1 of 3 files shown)


Rothalack:
Hey @PyjamaWarrior I think your code here is good. I'm just concerned with your endpoint. When I test locally I am getting the error Prebid axis ERROR: Server call for axis failed:  0. Continuing without bids. Are you able to test with the integration tests on your end?

I believe this is because my local test is on http, not https, not certain though. I am also seeing 1000+ms response time from your server as well. Is this intended?

Rothalack:
Hey @PyjamaWarrior , just wanted to check in and see if you had any questions or anything. Are you still working on things on your end?

modules/axisBidAdapter.js
```diff
@@ -0,0 +1,206 @@
...
+      deviceHeight,
+      language,
+      secure,
+      host,
+      page,
+      placements,
+      coppa: config.getConfig('coppa') === true ? 1 : 0,
+      ccpa: bidderRequest.uspConsent || undefined,
+      gdpr: bidderRequest.gdprConsent || undefined,
+      tmax: config.getConfig('bidderTimeout')
```
patmmccann, line 152: please prefer the request object to config

modules/axisBidAdapter.js
```diff
@@ -0,0 +1,206 @@
...
+    const placements = [];
+    const request = {
+      deviceWidth,
+      deviceHeight,
+      language,
+      secure,
+      host,
+      page,
+      placements,
+      coppa: config.getConfig('coppa') === true ? 1 : 0,
```
patmmccann, line 149: please prefer the request object to config

PyjamaWarrior:
@Rothalack  Can you check again, please? Now everything seems to be working fine.

Rothalack:
I'll have a look today. In the meantime, please try to address Pat's feedback above. I should have time to test things in an hour or two.

Rothalack:
@PyjamaWarrior I got a chance to test, everything looks good and is working. I still think you should address Pat's feedback about using the request instead of config above.

patmmccann:
Thanks so much @Rothalack ! I'll leave assigned to me to get this last change over the finish line.

modules/axisBidAdapter.js
patmmccann: @PyjamaWarrior here is an example 

https://github.com/prebid/Prebid.js/blob/509deaa0649824369907e6b76d0a4d045ec84df8/modules/riseBidAdapter.js#L328

modules/axisBidAdapter.js
patmmccann: @PyjamaWarrior here is an example https://github.com/prebid/Prebid.js/blob/509deaa0649824369907e6b76d0a4d045ec84df8/modules/riseBidAdapter.js#L421

patmmccann (changes requested):
Please change these two locations to look at the request object instead of the configuration.

patmmccann (changes requested):
We don't allow parameters that are just standard request fields

patmmccann (changes requested):
Two changes are currently required: 

1) iabCat: ['IAB1-1', 'IAB3-1', 'IAB4-3'] is not an allowed parameter, please read this from ortb2.site.cat or pagecat. Publishers don't enjoy setting the same field in specific ways for each adapter. Feel free to include instructions on how publishers can set this parameter or link to https://docs.prebid.org/features/firstPartyData.html

2) You are failing your own test according to the last circleci run

PyjamaWarrior:
@patmmccann I've fixed tests and iabCat. But i want to ask you about ortb2.site.cat. As you can see, i got this data from bidderRequest as you suggested and its same for all adUnits. Maybe its better to use ortb2Imp from validBidRequests to have categories for every adUnit separately? What do you think about that?

patmmccann:
IAB hangs it on the site object, but we allow it to change per request. see [ Supplying Auction-Specific Data](https://docs.prebid.org/features/firstPartyData.html#supplying-adunit-specific-data)

patmmccann (approved):

PR #9687, opened 2023-03-18
criteo bid adapter: fix for invalid domain in response 

patmmccann:
## Type of change

- [x] Bugfix
- [ ] Feature
- [ ] New bidder adapter   
- [ ] Build related changes
- [ ] CI related changes

- [ ] Other

## Description of change

Screenshot_20230318-092714

### DIFF
--- a/modules/criteoBidAdapter.js
+++ b/modules/criteoBidAdapter.js
@@ -233,7 +233,7 @@ export const spec = {
           bid.meta = Object.assign({}, bid.meta, { paf: pafResponseMeta });
         }
         if (slot.adomain) {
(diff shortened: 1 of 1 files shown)


allanjun (approved):
Hi, thank you for identifying and proposing a fix.
Criteo has actually just fixed this on the backend, meaning our endpoint will respond with adomain = ['domain.com']  so it won't break past versions already implemented.

This change adds a bit of redundancy 👍 just maybe rename the commit title to something like "reinforce adomain type in case of missmatch" to not make pubs think past versions are still broken

Issue #9689, opened 2023-03-20
Multiple cookie syncs requests


wiedaw:
## Type of issue
Question/Bug

## Description
I spotted the issue with `getUserSync`. 
When I enter the page and get first bid response after syncDelay time there is cookie sync request.
In the case when I wait until the cookie sync request is send when I scroll down the page and get more bid response I don't get more cookie syns as expected.

But in situation when I enter the page and get first bid but don't waint until cookie sync is done and scroll down the page and get new bid responses at the end I get multiple cookie sync requests.

## Steps to reproduce
Set `syncDelay` to 6000 for example.
After getting first bid request don't wait until cookie sync request will be send.
Scroll down the page to send other bid requests and you'll se multiple cookie sync requests.

### Expected results
On the website it should be only one cookie sync fired after first bid request.

### Actual results
If the user doesn't wait until first cookie sync and will scroll down the page and new bid requests are sent, at the end it will be more than one cookie sync.

## Platform details
Version 7.38.0
Browser Chrome 111.0.5563.65

dgirardi:
There is indeed a race condition with user syncs and multiple auctions. I am unsure about the right fix; the code's intent appears to be "adapters can return any number of syncs, but only once; after that, no more syncs will be run for them". There is a `syncsPerBidder` setting (default 5) that limits how many syncs a bidder can run, but that only works if the bidder returns them all at once. That is:

1 auction with one bidder that wants 5 syncs -> all 5 syncs will run
5 auctions with one bidder that wants 1 sync each time -> only the first will run (except for the bug in the OP)

I don't understand if this makes sense. Why not just limit syncs to `syncsPerBidder`? @patmmccann ?

PR #9695, opened 2023-03-21
Core: do not allow bidders to run syncs more than once

dgirardi:
## Type of change

- [x] Bugfix

## Description of change

Fix a bug where user syncs from the same bidder would be run more than once if you time the auctions just right.

## Other information

Fixes https://github.com/prebid/Prebid.js/issues/9689


ncolletti (approved):
lgtm

PR #9696, opened 2023-03-21
Magnite Analytics Adapter : add seat non bid handling

spotxslagle:
## Type of change

- [ x] Feature

## Description of change
Add support for seatnonbid events to the Magnite Analytics Adapter

### DIFF
--- a/modules/magniteAnalyticsAdapter.js
+++ b/modules/magniteAnalyticsAdapter.js
@@ -443,7 +443,7 @@ const findMatchingAdUnitFromOrderedAuctions = (matchesFunction, returnFirstMatch
 };
 
 const findMatchingAdUnitFromAllAuctions = (matchesFunction, returnFirstMatch) => {
-  return findAdUnitFromAuctions(matchesFunction, returnFirstMatch, cache.auctions);
+  return findAdUnitFromAuctions(matchesFunction, returnFirstMatch, Object.keys(cache.auctions));
 };
 
 const findAdUnitFromAuctions = (matchesFunction, returnFirstMatch, auctions) => {
@@ -982,7 +982,7 @@ const handleNonBidEvent = function(args) {
 const matchAuctionBeforeAdUnit = function(auctionId, impid) {
   return function (adUnit, auction) {
     if (auctionId !== auction.auctionId) return false;
-    return adUnit.code === impid;
+    return adUnit.adUnitCode === impid;
   }
 };
 
--- a/modules/magniteAnalyticsAdapter.js
+++ b/modules/magniteAnalyticsAdapter.js
@@ -956,7 +956,10 @@ const handleNonBidEvent = function(args) {
   const {seatnonbid, auctionId} = args;
   const auction = deepAccess(cache, `auctions.${auctionId}.auction`);
   // if no auction just bail
-  if (!auction) logWarn(`Unable to match nonbid to auction`) && return;
+  if (!auction) {
+    logWarn(`Unable to match nonbid to auction`);
+    return;
+  }
   const adUnits = auction.adUnits;
   seatnonbid.forEach(seatnonbid => {
     let {seat} = seatnonbid;
--- a/modules/magniteAnalyticsAdapter.js
+++ b/modules/magniteAnalyticsAdapter.js
@@ -955,6 +955,8 @@ magniteAdapter.track = ({ eventType, args }) => {
 const handleNonBidEvent = function(args) {
   const {seatnonbid, auctionId} = args;
   const auction = deepAccess(cache, `auctions.${auctionId}.auction`);
+  // if no auction just bail
+  if (!auction) logWarn(`Unable to match nonbid to auction`) && return;
   const adUnits = auction.adUnits;
   seatnonbid.forEach(seatnonbid => {
     let {seat} = seatnonbid;
@@ -965,10 +967,10 @@ const handleNonBidEvent = function(args) {
         const adUnit = adUnits[matchingTid];
         const statusInfo = statusMap[status] || { status: 'no-bid' };
         adUnit.bids[generateUUID()] = {
-          source: 'server',
           bidder: seat,
+          source: 'server',
           isSeatNonBid: true,
-          clientLatencyMillis: Date.now() - cache.auctions[args.auctionId].auction.auctionStart,
+          clientLatencyMillis: Date.now() - auction.auctionStart,
           ...statusInfo
         };
       } catch (error) {
--- a/modules/magniteAnalyticsAdapter.js
+++ b/modules/magniteAnalyticsAdapter.js
@@ -37,7 +37,8 @@ const {
     BIDDER_DONE,
     BID_TIMEOUT,
     BID_WON,
-    BILLABLE_EVENT
+    BILLABLE_EVENT,
+    SEAT_NON_BID
   },
   STATUS: {
     GOOD,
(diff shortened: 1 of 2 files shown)


modules/magniteAnalyticsAdapter.js
```diff
@@ -821,10 +830,16 @@ magniteAdapter.track = ({ eventType, args }) => {
...
-        logError(`${MODULE_NAME}: Could not find associated bid request for bid response with requestId: `, args.requestId);
-        break;
+        bid = adUnit.bids[generateUUID()] = {
+          bidder: 'mgnipbs',
+          source: 'server',
+          status: 'no-bid',
+          bidId: args.requestId,
+          clientLatencyMillis: args.timeToRespond || Date.now() - cache.auctions[args.auctionId].auction.auctionStart
+        };
+        return;
```
robertrmartinez, line 842: For this scenario, I think we want something a bit different.

This is a valid bid response, so in this scenario we actually want to create a new bid in our adUnit

And then just let the rest of the bid_response processing logic do what it normally does:

So something like this:

```js
      // If we still do not have a matching bid object and this bidResponse has a seatBidId, lets create a default one
      if (!bid && args.seatBidId) {
        bid = adUnit.bids[args.seatBidId] = {
          bidder: args.bidderCode,
          bidId: args.seatBidId,
          status: 'no-bid',
          source: 'server',
          unknownBid: true // just so we can see how often this happens at log level
        }
      }

      // finally if still no bid just bail like before
      if (!bid) {
        logError(`${MODULE_NAME}: Could not find associated bid request for bid response with requestId: `, args.requestId);
        break;
      }
```

So we just have a new check before this `!bid` check to see if it falls into our unknown bid response from PBS scenario. And if it does, then we can just create a bid on our adUnit, set it to its defaults, and then let the regular bid_response processing go.

modules/magniteAnalyticsAdapter.js
```diff
@@ -939,6 +957,66 @@ magniteAdapter.track = ({ eventType, args }) => {
...
+    });
+  });
+};
+
+const matchAuctionBeforeAdUnit = function(auctionId, impid) {
+  return function (adUnit, auction) {
+    if (auctionId !== auction.auctionId) return false;
+    return adUnit.adUnitCode === impid;
+  }
+};
```
robertrmartinez, line 989: think this can be removed when the refactor is done

robertrmartinez (changes requested):

modules/magniteAnalyticsAdapter.js
spotxslagle: Most of that sounds good to me, except letting regular bid response processing go. When that happens, we get this:
Screenshot 2023-03-30 at 4 14 16 PM
Some of that could be weirdness with how we mock the getStatusCode function, but do we really want any of the rest besides latency? And if we want to avoid copy pasta with latency, I could just put that code into a function that gets called twice.

robertrmartinez:
@spotxslagle Circle Ci complaining:

image

test/spec/modules/magniteAnalyticsAdapter_spec.js
```diff
@@ -2039,4 +2059,117 @@ describe('magnite analytics adapter', function () {
...
+      expect(utils.generateUUID.called).to.equal(true);
+
+      expect(message.auctions[0].adUnits[0].bids[1]).to.deep.equal(
+        {
+          bidder: 'mgnipbs',
+          source: 'server',
+          status: 'no-bid',
+          bidResponse: {
+            "bidPriceUSD": 0,
+            "conversionError": true
```
robertrmartinez, line 2098: Use single quotes instead of double quotes here

test/spec/modules/magniteAnalyticsAdapter_spec.js
```diff
@@ -160,6 +161,25 @@ const MOCK = {
...
     getStatusCode: () => 1,
   },
+  NO_BID: {
+    'requestId': 'fakeId',
+    'seatBidId': 'fakeId',
+    'bidderCode': 'mgnipbs',
+    'transactionId': '7b10a106-89ea-4e19-bc51-9b2e970fc42a',
+    'auctionId': '99785e47-a7c8-4c8a-ae05-ef1c717a4b4d',
+    'adUnitCode': 'box',
+    getStatusCode: () => 2,
```
robertrmartinez, line 171: For this scenario it is actually a valid "BID_RESPONSE" it just does not map to a bid

So we still want to use the regular fields a bidResponse would have.

Perhaps we could even remove this No_BID mock and just update the test below to clone the regular `MOCK.BID_RESPONSE` and just change the values as needed.

Something like:

```js
      const bidResponse = utils.deepClone(MOCK.BID_RESPONSE);
      bidResponse.requestId = null;
      bidResponse. seatBidId = '123-456';
      bidResponse. bidderCode = 'andrewBidder';
      bidResponse. cpm = 11.7;
```

And then the expect validation should be a normal use case where there is no conversion error and stuff.

test/spec/modules/magniteAnalyticsAdapter_spec.js
```diff
@@ -2039,4 +2059,117 @@ describe('magnite analytics adapter', function () {
...
+      expect(bid.isSeatNonBid).to.equal(true);
+    };
+    beforeEach(() => {
+      magniteAdapter.enableAnalytics({
+        options: {
+          endpoint: '//localhost:9999/event',
+          accountId: 1001
+        }
+      });
+      config.setConfig({ rubicon: { updatePageView: true } });
```
robertrmartinez, line 2142: This is probably not needed but not a big deal

test/spec/modules/magniteAnalyticsAdapter_spec.js
```diff
@@ -2039,4 +2059,117 @@ describe('magnite analytics adapter', function () {
...
+        {code: 0, status: 'no-bid'},
+        {code: 100, status: 'error', error: {code: 'request-error', description: 'general error'}},
+        {code: 101, status: 'error', error: {code: 'timeout-error', description: 'prebid server timeout'}},
+        {code: 200, status: 'rejected'},
+        {code: 202, status: 'rejected'},
+        {code: 301, status: 'rejected-ipf'}
+      ];
+      statuses.forEach((info, index) => {
+        checkStatusAgainstCode(info.status, info.code, info.error, index);
+      });
```
robertrmartinez, line 2172: 👍 nice!

test/spec/modules/magniteAnalyticsAdapter_spec.js
```diff
@@ -2039,4 +2059,117 @@ describe('magnite analytics adapter', function () {
...
+      let message = JSON.parse(server.requests[0].requestBody);
+
+      expect(message.auctions[0].adUnits[0].bids[1]).to.deep.equal(
+        {
+          bidder: 'rubicon',
+          source: 'server',
+          status: 'no-bid',
+          isSeatNonBid: true,
+          clientLatencyMillis: -139101369960
+        }
```
robertrmartinez, line 2157: So this one tests if the status is not in the status map, and then below test is all of the other ones cool.

robertrmartinez (changes requested):
Just a couple things in the tests

test/spec/modules/magniteAnalyticsAdapter_spec.js
```diff
@@ -2039,4 +2050,122 @@ describe('magnite analytics adapter', function () {
...
+    });
+
+    it('should add a no-bid bid to the add unit if it recieves one from the server', () => {
+      const bidResponse = utils.deepClone(MOCK.BID_RESPONSE);
+      const auctionInit = utils.deepClone(MOCK.AUCTION_INIT);
+
+      bidResponse.requestId = 'fakeId';
+      bidResponse.seatBidId = 'fakeId';
+
+
```
robertrmartinez, line 2072: Remove one of these new lines to solve the linting issue!

robertrmartinez (changes requested):

modules/magniteAnalyticsAdapter.js
```diff
@@ -939,6 +952,61 @@ magniteAdapter.track = ({ eventType, args }) => {
...
+      try {
+        const {status, impid} = nonbid;
+        const matchingTid = Object.keys(adUnits).find(tid => adUnits[tid].adUnitCode === impid);
+        const adUnit = adUnits[matchingTid];
+        const statusInfo = statusMap[status] || { status: 'no-bid' };
+        adUnit.bids[generateUUID()] = {
+          source: 'server',
+          bidder: seat,
+          isSeatNonBid: true,
+          clientLatencyMillis: Date.now() - cache.auctions[args.auctionId].auction.auctionStart,
```
robertrmartinez, line 971: Can be simplified to

```js
clientLatencyMillis: Date.now() - auction.auctionStart,
```

since auction is saved above

modules/magniteAnalyticsAdapter.js
```diff
@@ -939,6 +952,61 @@ magniteAdapter.track = ({ eventType, args }) => {
   }
 };
 
+const handleNonBidEvent = function(args) {
+  const {seatnonbid, auctionId} = args;
+  const auction = deepAccess(cache, `auctions.${auctionId}.auction`);
+  const adUnits = auction.adUnits;
```
robertrmartinez, line 958: WE should probably safe check in case the auction was not found for some reason:

```
  const auction = deepAccess(cache, `auctions.${auctionId}.auction`) ;
  // if no auction just bail
  if (!auction) logWarn(`Unable to match nonbid to auction`) && return;
```

PR #9699, opened 2023-03-22
Adteligent Bid Adapter : add 9 Dots Media alias

GeneGenie:
## Type of change
- [x] New bidder adapter 

## Description of change
Added new alias to adapter

## Other information
- [x] https://github.com/prebid/prebid.github.io/pull/4436
- [x] contact: support@9dotsmedia.com

### DIFF
--- a/test/spec/modules/adtelligentBidAdapter_spec.js
+++ b/test/spec/modules/adtelligentBidAdapter_spec.js
@@ -11,16 +11,16 @@ const EXPECTED_ENDPOINTS = [
   'https://ghb.adtelligent.com/v2/auction/'
 ];
 const aliasEP = {
-  appaloosa: 'https://ghb.hb.appaloosa.media/v2/auction/',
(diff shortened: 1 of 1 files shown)


ChrisHuie:
@GeneGenie can you pull in recent commits so the package.json isn't off 🙏

modules/adtelligentBidAdapter.js
```diff
@@ -24,6 +24,7 @@ const HOST_GETTERS = {
   pgam: () => 'ghb.pgamssp.com',
   ocm: () => 'ghb.cenarius.orangeclickmedia.com',
   vidcrunchllc: () => 'ghb.platform.vidcrunch.com',
+  '9dotsmedia': () => 'ghb.platform.audiodots.com'
```
patmmccann, line 27: Why is this the only one in quotes

GeneGenie: do you suggest wrapping all keys in quotes ?

ChrisHuie (approved):

PR #9706, opened 2023-03-23
IVS Bid Adapter: initial adapter release

ivs-mark:
## Type of change
- [x] New bidder adapter  

## Description of change
Add new bidder adapter for IVS

- Contact email: mark@ivs.tv
```
{
  bidder: 'ivs',
  params: {
    publisherId: '3000183'
  }
}
```

### DIFF
--- a/modules/ivsBidAdapter.js
+++ b/modules/ivsBidAdapter.js
@@ -1,0 +1,85 @@
+import { ortbConverter } from '../libraries/ortbConverter/converter.js';
+import { deepAccess, deepSetValue, getBidIdParameter, logError } from '../src/utils.js';
+import { registerBidder } from '../src/adapters/bidderFactory.js';
+import { VIDEO } from '../src/mediaTypes.js';
+import { INSTREAM } from '../src/video.js';
+
+const BIDDER_CODE = 'ivs';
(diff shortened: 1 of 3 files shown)


modules/ivsBidAdapter.js
```diff
@@ -0,0 +1,85 @@
...
+      logError(BIDDER_CODE + ': mediaTypes.video is not present in the bidder settings.');
+      return false;
+    }
+
+    if (deepAccess(bid, 'mediaTypes.video.context') !== INSTREAM) {
+      logError(BIDDER_CODE + ': only instream video context is allowed.');
+      return false;
+    }
+
+    if (!getBidIdParameter('publisherId', bid.params)) {
```
lksharma, line 44: The validation logic requires the params object to contain the `publisherId` field. However, the example configuration provided appears as the following:

`params: { site: { publisher: { id: '3001234' // required } } }`

Does this match the expected format? If not, please modify the example configuration to align with the validation criteria.

lksharma (changes requested):

modules/ivsBidAdapter.js
ivs-mark: Good catch @lksharma. The example configuration should be `params: { site: { publisherId: '3001234' // required } }`. I have updated the example configuration with the correct format. Thanks!

PR #9707, opened 2023-03-23
GumGum Bid Adapter : send gpp data in bidrequest

MartinGumGum:
Change the parameter name


MartinGumGum:
Just changing applicableSelection name to gpp_sid. Accutal support was already approved.

ChrisHuie (approved):

Issue #9709, opened 2023-03-24
Increase in timeouts since FPD enrichment


JulieLorin:
## Type of issue
I think this is a bug, or a performance issue

## Description
We are monitoring timeout events from Prebid.js and have noticed an increase of the timeouts lately (about 3 times more than usual for some publishers)
After investigation, it seems to be linked to #9205
And after further investigation (= enabling/disabling parts of prebid.js code), it seems to be linked to the call to `window.navigator?.userAgentData.getHighEntropyValues` in the `highEntropySUAAccessor` function, called when retrieving UA first party data. 

## Steps to reproduce
This is a "statistical" anomaly for us, we cannot "reproduce" the issue. 
What has been tested to conclude that the issue is the call the getHighEntropyValues is : to replace : 
```
cache[key] = uaData.getHighEntropyValues(hints).then(result => {
       return isEmpty(result) ? null : Object.freeze(uaDataToSUA(SUA_SOURCE_HIGH_ENTROPY, result))
   }).catch(() => null);
```
With : 
```
cache[key] = GreedyPromise.resolve(isEmpty(uaData) ? null : Object.freeze(uaDataToSUA(SUA_SOURCE_LOW_ENTROPY, uaData)));
```
(method from the lowEntropy function)
This single change will make our timeout rate down by 3, so I really think that the call to `userAgentData.getHighEntropyValues` is what's causing the timeouts issue. This function is experimental, and we have seen the issue only on browser that should support the feature (no change in timeouts with Firefox for example)

## Other information
There is a way to bypass this by setting the `firstPartyData.uaHints` config param to an empty array (that will make fpd enrichment use low entropy method, which causes no increase in timeouts), but maybe this should be addressed in a more global way ? (a simple change in getSUA would make low entropy the default method. I'd be happy to make a pull request for that if that seems relevant)

## Note 
I had noticed the bug fixed on Greedy Promise but this doesn't fix this issue (which I don't link to the same cause)

patmmccann:
Excellent work! We can likely make the call to getHighEntropy values opt in with publisher configuration, since they are not guaranteed to return this seems non breaking

JulieLorin:
Then I opened a PR for that, hope this will help :)

patmmccann:
Thank you!

PR #9711, opened 2023-03-24
FPD Enrichment: use low entropy method by default to fetch user agent data

JulieLorin:
## Type of change
- [X] (kind of) Bugfix

## Description of change
FPD enrichement will use "high entropy" method by default to fetch information about user agent. This causes an increase of bid timeouts (please see #9709)

We have tested to deploy this change in production and it has a positive effect on timeout (same level as it was before FPD enrichment became included by default)


JulieLorin:
I made a matching change in prebid docs : https://github.com/prebid/prebid.github.io/pull/4447

patmmccann (approved):

PR #9716, opened 2023-03-27
Yahoo ConnectId UserID Module: explicit storage management

slimkrazy:
## Type of change

- [x] Feature

- [ ] Other

## Description of change
The product team would prefer publishers to not have to specify storage options when using the connectId module. This PR shifts storage management into the module itself, as well as sending two extra parameters to our endpoint.

Note: There are not as many test changes as an initial scan of the PR would suggest. I just grouped some and added less than 10 new test cases.

## Other information
I will spend some time tomorrow setting up full e2e tests with these changes, but I want to get this PR in front of the maintainers sooner rather than later as I do have one question that I will put against a line in the PR.

Update to documentation site will follow shortly - we will just remove the requirement for the storage key as part of the configuration.


modules/connectIdSystem.js
```diff
@@ -59,7 +123,13 @@ export const connectIdSubmodule = {
       return;
     }
 
+    const storedId = getIdFromCookie() || getIdFromLocalStorage();
+    if (storedId) {
+      return {id: storedId};
```
slimkrazy, line 128: I have a question about what I'm doing here. I've noticed the documentation [here](https://github.com/prebid/Prebid.js/blob/master/modules/userId/index.js#L13) that states that if we return an `id` key, then the value will be stored in storage. It also makes it seem that both `id` and `callback` are optional. How can I prevent Prebid core from storing the value in storage, as we're already doing this ourselves, yet continue to allow the core logic to access the data?

patmmccann:
should we be defining a default error function like `function(e) { console.log(e) }`

modules/connectIdSystem.js
slimkrazy: After stepping away from this, I think I can answer this question myself: We need to explicitly state in our documentation that the storage config should be omitted, then Prebid core will not store the user ID value itself?

modules/connectIdSystem.js
```diff
@@ -8,14 +8,78 @@
...
 const VENDOR_ID = 25;
 const PLACEHOLDER = '__PIXEL_ID__';
 const UPS_ENDPOINT = `https://ups.analytics.yahoo.com/ups/${PLACEHOLDER}/fed`;
 const OVERRIDE_OPT_OUT_KEY = 'connectIdOptOut';
 const INPUT_PARAM_KEYS = ['pixelId', 'he', 'puid'];
+export const storage = getStorageManager({gvlid: VENDOR_ID, moduleName: MODULE_NAME});
+
+/**
+ * @function
+ * @param {Object} [value]
```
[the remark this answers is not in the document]
slimkrazy: Nope. I also forgot to rename when I renamed the param. Will update.

modules/connectIdSystem.js
```diff
@@ -8,14 +8,78 @@
...
+
+/**
+ * Attempts to store a cookie on eTLD + 1
+ */
+function setEtldPlusOneCookie(key, value, expiration, hostname) {
+  const subDomains = hostname.split('.');
+  for (let i = 0; i < subDomains.length; ++i) {
+    const domain = subDomains.slice(subDomains.length - i - 1, subDomains.length).join('.');
+    try {
+      storage.setCookie(key, value, expiration, null, '.' + domain);
```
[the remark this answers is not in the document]
slimkrazy: Good catch here - thanks. Updating.

modules/connectIdSystem.js
```diff
@@ -8,14 +8,78 @@
...
+  return null;
+}
+
+function getIdFromLocalStorage() {
+  if (storage.localStorageIsEnabled()) {
+    const storedIdData = storage.getDataFromLocalStorage(STORAGE_KEY);
+    if (storedIdData) {
+      if (isPlainObject(storedIdData) && storedIdData.__expires &&
+          storedIdData.__expires <= Date.now()) {
+        storage.removeDataFromLocalStorage(STORAGE_KEY);
```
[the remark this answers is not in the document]
slimkrazy: Yup. Thanks.

modules/connectIdSystem.js
slimkrazy: @jlukas79 would you mind confirming this for me?

slimkrazy: @jlukas79  - I'm just making another change. Spotted that we're looking for the consent data in the wrong place. Should update the PR shortly, please do not merge.

slimkrazy: @jlukas79  - Please review and merge whenever you are ready.

test/spec/modules/connectIdSystem_spec.js
```diff
@@ -41,250 +63,391 @@ describe('Yahoo ConnectID Submodule', () => {
...
-        he: HASHED_EMAIL,
-        puid: PUBLISHER_USER_ID,
-        pixelId: PIXEL_ID
-      }, consentData);
-      expect(result).to.be.an('object').that.has.all.keys('callback');
-      expect(result.callback).to.be.a('function');
-    });
+    describe('Low level storage functionality', () => {
+      const storageTestCases = [
+        { cookie: '{"connectId":"foo"}', localStorage: {connectId: 'bar'}, expected: {connectId: 'foo'} },
```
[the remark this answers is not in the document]
slimkrazy: @jlukas79 / @ChrisHuie  - Sorry to chase, but would you know if this PR would be able to get reviewed before the next release?

ChrisHuie (approved):

PR #9720, opened 2023-03-28
AirGrid RTD Provider: update audience segmentation data setting method for the Ortb2 bidder config

naripok:
## Type of change

- [x] Bugfix

## Description of change

Updates the way how the AirGrid RTD adapter passes audience segmentation information to the bidder config settings.
Solves a problem caused by breaking changes in the dependency modules.

## Other information

cc @ydennisy 

#### Help wanted

This PR is needed as there were some breaking changes in how the appnexus module accepts site keywords configuration. Although we have testing for our module, this breaking change passed unnoticed for a while. We would like to improve our testing suite in order to prevent this in the future.

The idea here is to add request data checking to the test, making sure that our module is passing data in the correct format to the appnexus adapter, and that appnexus is able to send requests containing the correct data after that.

Could you guys suggest a good way to achieve this, please?

I've tried the following setup, but without any luck:
```javascript
it('passes audiences to bidders correctly and generates a valid request', function () {
  const adapter = newBidder(spec); // this contains the appnexus spec imported from their module
  const adUnits = getAdUnits(); // from mocks

  getDataFromLocalStorageStub
    .withArgs(agRTD.AG_AUDIENCE_IDS_KEY)
    .returns(JSON.stringify(MATCHED_AUDIENCES));
  agRTD.passAudiencesToBidders({ adUnits }, () => {}, RTD_CONFIG, {});

  const request = {
    bidderCode: 'appnexus',
    auctionId: 'valid-uuid',
    adUnits
  };

  adapter.callBids(request)

  expect(server.requests.length).to.equal(1); // this is currently returning 0 and failing :(
  // assert on request data next...
});
```


modules/airgridRtdProvider.js
```diff
@@ -131,7 +141,7 @@ export function passAudiencesToBidders(
   const audiences = getMatchedAudiencesFromStorage();
   if (audiences.length > 0) {
     setAudiencesUsingAppNexusAuctionKeywords(audiences);
```
ydennisy, line 143: Reviewers: we plan to remove this Xandr specific method / handling in another PR once this is live and tested.

ChrisHuie (approved):

Issue #9717, opened 2023-03-27
Should aliases default to the GVL ID of the main bidder?


bretg:
## Type of issue
topic of discussion

## Description
One of our pubs is running a bidder that is one of the aliases of the appnexus adapter:

```
  aliases: [
    { code: 'appnexusAst', gvlid: 32 },
    { code: 'emxdigital', gvlid: 183 },
    { code: 'pagescience' },
    { code: 'defymedia' },
    { code: 'gourmetads' },
    { code: 'matomy' },
    { code: 'featureforward' },
    { code: 'oftmedia' },
    { code: 'adasta' },
    { code: 'beintoo', gvlid: 618 },
  ],
```

What's happening now is that that aliased bidder is not passing GDPR checks because it doesn't declare the GVL ID.

The workaround is easy enough:
```
pbjs.mergeConfig({
    gvlMapping: {
         MYALIAS: 32
    }
})
```
But this begs a few questions:
1. Is it legally correct for aliases to share GVL IDs with the main bidder? It's not clear to me that a user consenting to appnexus is also aware that they're consenting to pagescience, defymedia, adasta, etc.
2. If it is legally correct, why not have Prebid.js handle this scenario without the manual override?
3. If it isn't legally correct, is there any documentation desirable to help publishers determine what to do? I could see, for instance, a case where a 'white label' adapter doesn't share data with the alias, then it might be ok to share the GVL ID. But if the alias is a separate business with access to raw logs, then it might not be ok to assume the user is fine sharing their log-level data with this separate entity.

My proposal is that the right approach for Prebid is just what's happening today: the main adapter code should copy their own `gvlid` for aliases where they don't share data. If they share data, then the alias should have either no `gvlid` (which should result in failed GDPR validation) or their own `gvlid` so users can consent separately.

Updated req from discussion: we should do log.info (only shows in debug) if someone aliases a bidder and loses gvlid. otherwise, we'll solve the rest with docs on prebid.org

patmmccann:
I aggree with your proposal

> : the main adapter code should copy their own gvlid for aliases where they don't share data. If they share data, then the alias should have no gvlid, which should result in failed GDPR validation.

with a slight edit:

`: the main adapter code should copy their own gvlid for aliases where they don't share data. If they share data, then the alias should have an additional or no gvlid, which should result in failed GDPR validation if the vendor id does not have consent for whichever purpose(s) the enforcement module looks for`

bretg:
[...]

Issue #9546, opened 2023-02-15
GPP Phase 3 - Activity control infrastructure


bretg:
Splitting up the GPP feature tracked originally by #8991.

Phase 3 is the "Activity Infrastructure". From [the document](https://docs.google.com/document/d/1dRxFUFmhh2jGanzGZvfkK_6jtHPpHXWD7Qsi6KEugeE/edit#), this phase is described as:

Prebid.js and Prebid Server will define a set of special activities, each of which can be controlled either directly by the publisher or, in the future, via consent modules like GPP.

The activities defined for Prebid.js are in the document section 3.3.3.

The work for this phase includes:
1. Implement the call to each activity. The interface will need to be specified. It might make sense to do some performance testing for whether activity interfaces should be called once-per-bidder/module or just once and return a list of allowed/disallowed bidders/modules.
2. Documentation

This issue details a proposal for the Page and ORTB configs that control the activities.

### Page Level Config

The `condition`  logic will initially be kept simple, just modulename, which corresponds to biddercode or modulecode. However, the attributes used in conditions could expand in the future to include country, mediatype, regulation scope, or other fields.

Proposed page-level activity controls:
```
pbjs.setConfig({
  consentManagement: {
    allowActivities: {
      accessDevice: {
        default: false, // for access_device, this is only for vendors
        rules: [{ 
          priority: 1,
          condition: {
            component: { in: ["core"]} // equivalent of strictstorage flag
          },
          allow: true
        }]
      },
      syncUser: {
        default: true,
        rules: [{  // this is an array because conditions could someday be complicated
          priority: 1,
          condition: {
            component: {in: ["bidderA"]}
          },
          allow: false
        }]
      },
      transmit_ufpd: {
        default: false,
        rules: [{
          condition: {
            component: {in: ["bidderB", "module1"]}
          },
          allow: true
        }]
      },
      ...
    }
  }
});
```

### Details

Here's the general form:
```
allowactivities: {
        ACTIVITY: {
            default: true/false,
            rules: [{
              priority: INTEGER,
              condition: { CONDITION_LOGIC },
              allow: true/false
            },{
              ... more rules ...
            }]
        }
    }
```
Definitions:
- **default**: defines whether to perform this activity if no overrides or rules fire. If default is unspecified, the hardcoded default-default is true.
[...]

PR #9736, opened 2023-03-30
Multiple modules: decouple GVL IDs from storage access control

dgirardi:
## Type of change

## Description of change

This refactors `storageManager` so that it no longer deals with GVL IDs directly. Currently a module can pass it a GVL ID which gets [mostly ignored](https://github.com/prebid/Prebid.js/issues/9717#issuecomment-1487362558); in addition, with the upcoming [activity controls](https://github.com/prebid/Prebid.js/issues/9546) GVL IDs will be just one of the mechanisms that will control access to storage.

With this PR:

- modules must ID themselves fully to `storageManager` (a module type and name are now required);
- GVL IDs are tracked separately; userID will keep track of all userId submodule GVLIDs, core does the same for bid adapters, etc
- RTD modules can also now declare a `gvlid` in their spec in the same vein as all other module types.


Fawke (approved):
Hi @dgirardi,

Thanks for the PR and all the the refactoring. All looks good from my end. I agree, that the `gvlid` parameter passed to storageManager wasn't doing much. 

I don't think we'll need to make any docs change for this. I could not find any code snippet or example of initialising storageManager.

src/storageManager.js
```diff
@@ -252,31 +231,38 @@ export function newStorageManager({gvlid, moduleName, bidderCode, moduleType} =
...
+    throw new Error(`Invalid invocation for getStorageManager: must set either bidderCode, or moduleType + moduleName`)
+  }
+  if (bidderCode) {
+    if ((moduleType && moduleType !== MODULE_TYPE_BIDDER) || moduleName) err()
+    moduleType = MODULE_TYPE_BIDDER;
+    moduleName = bidderCode;
+  } else if (!moduleName || !moduleType) {
+    err()
+  }
+  return newStorageManager({moduleType, moduleName});
```
smenzer, line 256: could be a dumb question, but should this line be this?

```js
return newStorageManager({moduleName: moduleName, moduleType: moduleType});
```

smenzer (commented):
I focused only on the user id module pieces, but this LGTM

src/storageManager.js
smenzer: got it!

PR #9746, opened 2023-03-31
Pub-X Bid Adapter: adding page url support

yuki-arasaki:
## Type of change
- [x] Feature

## Description of change
Adds the page's URL information to the parameters of the request URL.

### DIFF
--- a/modules/pubxBidAdapter.js
+++ b/modules/pubxBidAdapter.js
@@ -15,9 +15,12 @@ export const spec = {
     return validBidRequests.map(bidRequest => {
       const bidId = bidRequest.bidId;
       const params = bidRequest.params;
+      const pageUrl = location.href.replace(/\?.*$/, '');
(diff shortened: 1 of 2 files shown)


modules/pubxBidAdapter.js
```diff
@@ -15,9 +15,12 @@ export const spec = {
     return validBidRequests.map(bidRequest => {
       const bidId = bidRequest.bidId;
       const params = bidRequest.params;
+      const pageUrl = location.href.replace(/\?.*$/, '');
```
patmmccann, line 18: why not use site.page on the request object?

yuki-arasaki: @patmmccann 
Thank you for letting me know about the site object. I found what I needed in the request object with `ortb2.site.page` .
Made the changes and committed the code. b2943068cbd223e0d7588d202b78a3891a30d921

patmmccann (approved):

PR #9752, opened 2023-04-03
Prebid core: fix image assets in converted legacy response

JulieLorin:
## Type of change
Bugfix

## Description of change
Bug was already there in #9252
We had some errors with native images, for Index Exchange only (which is using ortb format). 
It seems that the image and icon format of the legacy native response returned by the `toLegacyResponse` function is invalid : it is a string instead of being an object with url, w and h properties. 

Before : 
image
After : 
image

(Not sure how I haven't seen that before, I might be missing something, is it possible it is like this by design ?)

### DIFF
--- a/src/native.js
+++ b/src/native.js
@@ -750,7 +750,11 @@ export function toLegacyResponse(ortbResponse, ortbRequest) {
     if (asset.title) {
       legacyResponse.title = asset.title.text;
     } else if (asset.img) {
-      legacyResponse[requestAsset.img.type === NATIVE_IMAGE_TYPES.MAIN ? 'image' : 'icon'] = asset.img.url;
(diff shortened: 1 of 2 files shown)


jsnellbaker (approved):
LGTM

jsnellbaker:
@JulieLorin could you try to rebase with master?  The circleCI tests didn't run due to a lint error in a bid adapter (which has likely been corrected since you originally created this PR and synced with master).

JulieLorin:
> @JulieLorin could you try to rebase with master? The circleCI tests didn't run due to a lint error in a bid adapter (which has likely been corrected since you originally created this PR and synced with master).

Done but there seems to be another error (a connection timeout)

jsnellbaker:
Thanks @JulieLorin I reran the job and it seems to have run cleanly now.

jlquaccia (approved):
lgtm too

Issue #9733, opened 2023-03-30
SharedID is not set on first page view if pixelUrl is used


muuki88:
## Type of issue

Bug

## Description

We see an unexpected behaviour of sharedID on our website.
This is our sharedID config:

```js
{
  name: 'sharedId',
  params: {
    pixelUrl: '/ads/sharedid',
    create: false // server will create the cookie if it doesn't exist
  },
  storage: {
    name: '_sharedID',
    type: 'cookie',
    expires: 30
  }
}
```

 [Opening a page in private mode](https://www.gutefrage.net/frage/kann-ich-ohne-abi-aber-mit-einer-pta-ausbildung-medizin-studieren?pbjs_debug=true) `pbjs` states

> Prebid INFO: User ID: sharedId - request id responded with an empty value

Eventhough the request was sent properly `set-cookie: _sharedID=8f552993-a15f-4e23-b9f6-b86964ac0a14; Expires=Fri, 29 Mar 2024 11:18:42 GMT; Path=/` is set.

And the thing that really puzzles us `pbjs.getUserIds()`
does not contain the `pubcid`, only the other identifiers. After reloading the page, `pubcid` appears in the output.

## Steps to reproduce

- configure sharedID with `pixelUrl` and `create:false` as described [in the docs](https://docs.prebid.org/identity/sharedid.html)

### Expected results

The `sharedID` module waits for the endpoint to respond and sets the pubcid.

### Actual results

SharedID module fire-and-forgets the pixel request and returns no id.

## Platform details

- prebid 7.42.0

## Other information

- https://github.com/prebid/Prebid.js/pull/5315

PR #9756, opened 2023-04-03
SharedID: wait for ID from pixel on first page load

dgirardi:
## Type of change

- [x] Bugfix

## Description of change

This:

 - passes a new `readStoredId` parameter to ID submodule callbacks that allows them to get a fresh version of their stored ID (for situations where a pixel may update it out-of-band), and
 - fixes https://github.com/prebid/Prebid.js/issues/9733


muuki88:
thanks @ChrisHuie and @dgirardi for the super fast fix :heart:

PR #9760, opened 2023-04-04
33across Id System: Stop generating error for all cookied users

patmmccann:
33across Id system is spamming errors, generating one for all cookied users.

image


smenzer (approved):

mmoschovas (approved):

robertrmartinez:
Cool change

But looks like their endpoint just throws errors now when it detects cookies, which causes error to be logged anyways haha.

robertrmartinez:
image

PR #9769, opened 2023-04-05
CORE: disable the requirement for having a url, when enabling renderNow in Renderer

olafbuitelaar:
## Type of change

- [ x] Bugfix
- [ ] Feature
- [ ] New bidder adapter   
- [ ] Build related changes
- [ ] CI related changes

- [ ] Other

## Description of change
Right now when you specify a `Renderer` for a `Bid` instance, which shouldn't load any url, the `renderNow` can be set, however this flow currently requires you to set a dummy url, otherwise the check on `isRendererRequired` would fail. I've altered this check so either the `url` could be set or the `renderNow` flag.

### DIFF
--- a/src/Renderer.js
+++ b/src/Renderer.js
@@ -22,6 +22,7 @@ export function Renderer(options) {
   this.config = config;
   this.handlers = {};
   this.id = id;
+  this.renderNow = renderNow;
 
   // a renderer may push to the command queue to delay rendering until the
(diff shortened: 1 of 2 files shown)


elad-yosifon:
👍

Fawke (approved):

dgirardi (approved):

PR #9771, opened 2023-04-05
[Video Module] Include a type property in all events

karimMourra:
## Type of change

- [x] Bugfix

## Description of change
The `type` property is missing from certain events fired by the video module. For consistency, we have added a util.

### DIFF
--- a/libraries/video/shared/helpers.js
+++ b/libraries/video/shared/helpers.js
@@ -1,5 +1,19 @@
 import { videoKey } from '../constants/constants.js'
 
 export function getExternalVideoEventName(eventName) {
+  if (!eventName) {
+    return '';
+  }
   return videoKey + eventName.replace(/^./, eventName[0].toUpperCase());
+}
+
+export function getExternalVideoEventPayload(eventName, payload) {
+  if (!payload) {
+    payload = {};
+  }
+
+  if (!payload.type) {
+    payload.type = eventName;
+  }
+  return payload;
 }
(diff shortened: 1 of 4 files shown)


pm-harshad-mane:
I re-ran the CircleCI job and we have all the checks passed now.
@musikele, Can you please review the code changes?

karimMourra:
Hi @pm-harshad-mane and @musikele this has been open for a while. Any blockers to merging ?

musikele:
just struggled with tasks at my regular work. Will check out now.

musikele (approved):
LGTM but, do we need a second review for library code?

libraries/video/shared/helpers.js
```diff
@@ -1,5 +1,19 @@
 import { videoKey } from '../constants/constants.js'
 
 export function getExternalVideoEventName(eventName) {
+  if (!eventName) {
+    return '';
+  }
   return videoKey + eventName.replace(/^./, eventName[0].toUpperCase());
```
pm-harshad-mane, line 7: Here are we assuming that all video players have the event names have the prefix `video` ? 
Should it be configurable with the default value `video`?
Refer: https://github.com/prebid/Prebid.js/blob/master/libraries/video/constants/constants.js#L1

pm-harshad-mane: @musikele the library code has very limited functionality, looks good to me but I have one doubt (mentioned in the above comment)
@karimMourra please check the above comment.

pm-harshad-mane:
@karimMourra do we want to update some documentation along with these code changes? 
Please create /share the PR link for the same.

libraries/video/shared/helpers.js
karimMourra: @pm-harshad-mane these events are part of the module's API. When video players surface events, they surface them using the following constants https://github.com/prebid/Prebid.js/blob/master/libraries/video/constants/events.js and we append the `video` portion. It keeps the char count lower. 
This is already documented https://docs.prebid.org/prebid-video/video-module.html

karimMourra:
@pm-harshad-mane the documentation is already up to date https://docs.prebid.org/prebid-video/video-module.html , see the Events section

pm-harshad-mane:
Thanks @karimMourra and @musikele, I will merge these changes then.

PR #9773, opened 2023-04-06
Appnexus bid adapter: support for video.plcmt

fowler446:
## Type of change
- [x ] Feature

## Description of change
Support new `plcmt` field as per:
https://github.com/InteractiveAdvertisingBureau/AdCOM/blob/develop/AdCOM%20v1.0%20FINAL.md#list_plcmtsubtypesvideo

## Other information
@jaiminpanchal27 
@Fawke

### DIFF
--- a/modules/appnexusBidAdapter.js
+++ b/modules/appnexusBidAdapter.js
@@ -960,6 +960,9 @@ function bidToTag(bid) {
               tag.video[contextKey] = VIDEO_MAPPING[contextKey][context];
             }
             break;
+          case 'plcmt':
+            tag.video['plcmt'] = videoMediaType['plcmt'];
+            break;
         }
       });
   }


jsnellbaker:
This `plcmt` field does not exist on our endpoint, so it would not do anything in this state.  

The closest field that we have that seems to match this is the context field in our endpoint, which accepts values like `pre_roll`, `mid_roll`, `post_roll`, `outstream` or `in-banner`.  

We're already populating this values for the context field from the ORTB placement and startDelay, so we'd probably want to include this field in the earlier block and it as part of the potential values that we'd look at to help populate the `context` field for our endpoint.  If you want to try to make the edits along those lines, please update this PR.  If you're not sure, then please confirm and I'll try to create a ticket in our personal sprints to look at this request when we have bandwidth.

patmmccann:
context is insufficient to form an openrtb bid request; how would you differentiate placement = 1 , plcmt = 1 vs placement = 1, plcmt = 2

jsnellbaker:
Our endpoint is not an openrtb endpoint and it requires updates to conform to changes like this request.  

From what I understand based on some conversations with our product team, there are some updates planned to provide more options for the `context` field our endpoint uses.  But to clarify - I believe the expanded options would more align to the outstanding fields from the `video.placement` openrtb spec that are not currently supported (such as in-feed, video interstitial, etc).  

The `plcmt` field was only added to openrtb 2.6 earlier this month (from what I can tell), so it would need additional time to update the endpoint to properly support the differences that were introduced with the new values from the `plcmt` field.  Until then, we can try to support the plcmt by interpreting the values to the equivalent values available in the `context` (ie openrtb video.placement` field).  If that's not possible due to the different meanings that have been introduced (which may be the case from what I've tried to interpret), then this support may have to wait.

PR #9785, opened 2023-04-08
Growthcode UserId: Bug fixes & Better Error Catching

southern-growthcode:
## Type of change
- [x ] Bugfix

## Description of change
- Removed internal cookie storage
- Better error catching 
- Code clean up


Rothalack (approved):

PR #9791, opened 2023-04-10
JW Player Video Adapter: Support multiple setup listeners

karimMourra:
## Type of change

- [x] Bugfix

## Description of change
Fixes a bug where additional setup event listeners would not get called. 
This change consists of always adding setup event listeners to arrays and adding event listeners on the player. WHen the setup events are triggered, all event listeners in the array are called.


osazos (approved):
LGTM

PR #9794, opened 2023-04-11
openxBidAdapter: update to OpenRTB adapter

bwschmidt:
## Type of change

- [x] Other

## Description of change
After testing on a handful of publishers OpenX will be moving to our updated OpenRTB serving architecture. The proposal here is to simply mirror the code from the "new" adapter openxOrtbAdapter into our default adapter openxBidAdapter so that publishers will pick up the changes seamlessly. The diffs will look large, but in reality it is just a copy/paste of the openxOrtbAdapter. We would like to remove openxOrtbAdapter in Prebid 8. Let me know if that works.

### DIFF
--- a/modules/openxBidAdapter.js
+++ b/modules/openxBidAdapter.js
@@ -1,609 +1,238 @@
-import {
-  _each,
-  _map,
-  convertTypes,
-  deepAccess,
-  deepSetValue,
-  inIframe,
-  isArray,
-  parseSizesInput,
-  parseUrl
-} from '../src/utils.js';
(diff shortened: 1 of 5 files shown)


patmmccann:
Please note likely conflict with #9799

patmmccann (changes requested):
Please delete the skipped tests and open an issue to drop the second adapter and tag it 8.0

Other than that, sounds good

patmmccann (approved):

patmmccann (approved):

PR #9800, opened 2023-04-13
Brid Bid Adapter: Initial adapter release

grajzer:
## Type of change
- [x] New bidder adapter

## Description of change
Add new bidder adapter for Brid.

- Contact email: [dejan@brid.tv](mailto:dejan@brid.tv)
```
{
  bidder: 'brid',
  params: {
    placementId: '12345'
  }
}
```


fowler446:
@jsnellbaker 
Docs look good 👍 
https://github.com/prebid/prebid.github.io/pull/4482

Sovica:
Hey Team,

Any updates on when we can get a review here?

losmicar:
@fowler446 Any estimate when this should be approved ? Thanks!

ChrisHuie:
@losmicar sorry. I will make sure and review this week

losmicar:
Thanks @ChrisHuie any updates or timeline estimate about Brid adapter? Any kind of feedback will be useful. Thanks!

modules/bridBidAdapter.js
```diff
@@ -0,0 +1,225 @@
...
+      requests.push({
+        method: 'POST',
+        url: ENDPOINT_URL,
+        data: JSON.stringify(postBody),
+        options: {
+          withCredentials: true
+        },
+        bidId,
+        params,
+        auctionId,
```
patmmccann, line 122: this might be null, is that alright?

test/spec/modules/bridBidAdapter_spec.js
```diff
@@ -0,0 +1,132 @@
...
+          'adomain': ['domain.com'],
+          'w': 640,
+          'h': 480
+        }],
+        'seat': 'bidder'
+      }]
+    };
+    const bidderRequest = {
+      bidderCode: 'brid',
+      auctionId: '1d1a030790a475',
```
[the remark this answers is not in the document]
grajzer: Hello @patmmccann, tests updated. All example requests now have bidId and auctionId specified. Is that what you meant, or we should modify "isBidRequestValid" method to check if bidId is present (although I don't see other modules have that)?

sibussiso:
Hi @patmmccann ,

Let us know if anything else is needed in regards to the above or when can we expect the adapter to go live.

modules/bridBidAdapter.js
```diff
@@ -0,0 +1,224 @@
...
+      if (sizes && !Array.isArray(sizes[0])) sizes = [sizes];
+
+      const site = getSiteObj();
+
+      const postBody = {
+        sdk: {
+          source: SOURCE,
+          version: '$prebid.version$'
+        },
+        id: auctionId,
```
patmmccann, line 57: So this looks like openrtb, and in openrtb $.id should be unique to a request, not shared with other requests, so you cannot use auctionid here. You should instead use bidderrequestid. Auctionid is most fitting for source.tid, however, it might be null in the bid request object as of prebid 8, so that publishers can control downstream information leaks. This is why my previous comments said to make sure your adapter can take null auctionid.

patmmccann (changes requested):

modules/bridBidAdapter.js
grajzer: Hi @patmmccann , thank you for the explanation! Removed auctionId from requests and changed id to use bidderRequest.bidderRequestId instead.

patmmccann (approved):

PR #9804, opened 2023-04-13
PBS Bid Adapter: add context to emitted seatnonbid event

spotxslagle:
## Type of change

- [x] Feature

## Description of change

Adds additional information to the emitted seatnonbid event.

## Other information

@robertrmartinez


robertrmartinez:
ok now I am remembering that some of these in `context` are functions, so if we do this sanitize the downstream event listeners will not be able to use those functions.

Sorry, but I think we should just pass through as is @spotxslagle

spotxslagle:
@dgirardi

dgirardi (approved):
LGTM, and thanks for the ping!

robertrmartinez (approved):

Issue #9730, opened 2023-03-29
Concerns with multiple ids coming from one id submodule response


patmmccann:
## Type of issue
feature request

## Description

conflicts handled alphabetically - we need to make, as an example, uid2 conflicts default to the uid2 module

unexpected behavior might happen - we should make extra returns (eg liveintent returning bidswitch and uid2 id, or 33across id returning rubicon id) as must be config opt in -- is this breaking? [update, it is not breaking]

## Requirements:

modules that "own" an id should win in case of conflict, eg uid2 module should beat liveintent on the uid2. Modules that don't "own" an id should declare as such and always lose in case of conflict

docs / rules commit that extra ids must all be opt in

3link:
@patmmccann  Providing a way to express priorities in case of collisions will make reasoning about the final result a lot easier. I think it is the right step. Having an opt-in configuration for extra returns seems like a good idea too. LiveIntent's module follows that approach already via the module specific configuration - by default, only a single id is returned. Lifting this to a general concept in Prebid and thus unifying handling of multiple ids is a good thing. These changes would be breaking, but we talked to our customers and it should not be an issue. Also, we will adjust LiveIntent's userId module as part of this effort when required.

A couple of thoughts/questions that might help to think about a solution:
- In the case of two providers for the same id type: How would the mechanism for determining who is the primary provider for that type look like?
- In case of modules returning multiple ids: When going with the opt-in solution for multiple identifiers, will it be required to define what id is exposed by default? Will additional ids be enabled one by one or will it be a flag enabling all of them?

patmmccann:
[https://docs.google.com/document/d/1oOe8cA8NsT4JmEFs095faJpe8QxktfvGGnF9uBsKqzA/edit?usp=sharing](https://docs.google.com/document/d/1oOe8cA8NsT4JmEFs095faJpe8QxktfvGGnF9uBsKqzA/edit?usp=sharing&authuser=1)

Proposal on potential config; proposer also likes the idea of modules can declare which id's they do not 'own'

patmmccann:
solved with #9896

PR #9828, opened 2023-04-19
LiveIntent UserId module: Add support for magnite id 

3link:
## Type of change
- [x] Other

## Description of change
This pull request adds eid mapping for the magnite ids that now can be resolved via the LiveIntent UserId module.

## Other information

- repository of the backing live-connect module: https://github.com/LiveIntent/live-connect
- documentation PR: https://github.com/prebid/prebid.github.io/pull/4498

### DIFF
--- a/modules/liveIntentIdSystem.js
+++ b/modules/liveIntentIdSystem.js
@@ -196,6 +196,10 @@ export const liveIntentIdSubmodule = {
         result.medianet = { 'id': value.medianet }
       }
 
+      if (value.magnite) {
+        result.magnite = { 'id': value.magnite }
+      }
+
       return result
     }
 
--- a/modules/userId/eids.js
+++ b/modules/userId/eids.js
@@ -164,6 +164,15 @@ export const USER_IDS_CONFIG = {
     }
   },
 
+  // magnite
+  'magnite': {
+    source: 'rubiconproject.com',
+    atype: 3,
+    getValue: function(data) {
+      return data.id;
+    }
+  },
+
   // britepoolId
   'britepoolid': {
     source: 'britepool.com',
--- a/modules/userId/eids.md
+++ b/modules/userId/eids.md
@@ -115,6 +115,14 @@ userIdAsEids = [
     },
     
     {
+        source: 'rubiconproject.com',
+        uids: [{
+            id: 'some-random-id-value',
+            atype: 3
+        }]
+    },
+    
+    {
         source: 'media.net',
         uids: [{
             id: 'some-random-id-value',
--- a/test/spec/modules/eids_spec.js
+++ b/test/spec/modules/eids_spec.js
@@ -202,6 +202,21 @@ describe('eids array generation for known sub-modules', function() {
     });
   });
 
+  it('magnite', function() {
+    const userId = {
+      magnite: {'id': 'sample_id'}
+    };
+    const newEids = createEidsArray(userId);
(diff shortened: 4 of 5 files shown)


3link:
The test errors seem not to be related to changes in this PR.

deepthivenkat:
Hello @patmmccann  Could you let me know for which of the following scenarios we are implementing the tie break for:
A) Magnite cookies are present from the magnite module and LiveIntent user id module also has a magnite cookie
B) Magnite cookies are not present in the magnite module but are present in both 33Across module and LiveIntent user id module

The reason i ask this is: 
For scenario A) if the path in which the magnite bidder adapter looks for magnite cookies populated from magnite module is different from the user->ext->eids array, then it is cleaner to implement the logic to read magnite cookies from LI only if magnite cookies are not populated by magnite module from magnite's bidder adapter. If the path is same, then i need to understand a little bit more about how to check if the magnite cookies are already present

For scenario B) Which approach do you like better? Relying on the load sequence(this provides equal opportunity for both 33Across and Liveintent) or allowing for Magnite to define the priority to pick up this cookie when its own module doesnt have it?

jdwieland8282:
The id PMC recommends an optional `ext.provider` field to reference the "source of the source"
```
"user":{
   "ext":{
      "eids":[
         {
            "source":"magnite.com",
            "uids":[
               {
                  "id":"d88c96-5cb6-410d-827d-b019e476",
                  "atype":1,
                  "ext":{
                     "source":"magnite"
                  }
               }
            ]
         },
         {
            "source":"magnite.com",
            "uids":[
               {
                  "id":"d88c96-5cb6-410d-827d-b019e476",
                  "atype":1,
                  "ext":{
                     "source":"magnite"
                  }
               }
            ]
         }

deepthivenkat:
Thank you!

3link:
@patmmccann @jdwieland8282 Can you please have a look at this proposal: https://github.com/prebid/Prebid.js/pull/9896?

SKOCHERI (dismissed):

SKOCHERI:
@3link Can you update https://github.com/prebid/Prebid.js/blob/master/modules/userId/eids.md with magnite eids

3link:
@SKOCHERI Done in https://github.com/prebid/Prebid.js/pull/9828/commits/68aae9531a44cb51224eaa8c82961c657fe00885. 

@SKOCHERI Please note, there is a [concern](https://github.com/prebid/Prebid.js/issues/9730) that is blocking this PR. There is a proposal that might unblock it and requires a review: https://github.com/prebid/Prebid.js/pull/9896.

skocheri1 (approved):

dgirardi (commented):
this seems to go against https://github.com/prebid/Prebid.js/pull/9896, as in including the liveintent module can populate `magnite` at the top level instead of `libp.magnite`. is there some other integration pending?

3link:
> this seems to go against https://github.com/prebid/Prebid.js/pull/9896, as in including the liveintent module can populate magnite at the top level instead of inside libp. is there some other integration pending?

@dgirardi This is correct. This PR started the discussion about exposing the same type of identifiers via multiple user id modules. https://github.com/prebid/Prebid.js/pull/9896 is meant to be a place where possible solutions are discussed. Once a solution has been found, I will update this PR. 

It is possible to flag this PR as blocked? Alternatively, I can convert it to a draft so it does not confuse reviewers.

PR #9840, opened 2023-04-20
ortbConverter: support video.plcmt

bwschmidt:
## Type of change

- [x] Feature

## Description of change
Adding new field `plcmt` to the allowed `mediaTypes.video` fields in ortbConverter. https://github.com/InteractiveAdvertisingBureau/AdCOM/blob/develop/AdCOM%20v1.0%20FINAL.md#list_plcmtsubtypesvideo

### DIFF
--- a/libraries/ortbConverter/processors/video.js
+++ b/libraries/ortbConverter/processors/video.js
@@ -6,6 +6,7 @@ import {sizesToFormat} from '../lib/sizes.js';
 const ORTB_VIDEO_PARAMS = new Set([
   'pos',
   'placement',
+  'plcmt',
   'api',
   'mimes',
   'protocols',


dgirardi (approved):
Thank you for this! I was meaning to try to catch up at some point - do you know if it'd make sense to treat this the same way we treat `placement`? Namely, that it should be defaulted to `1` if `context === 'instream'` ([this logic](https://github.com/prebid/Prebid.js/blob/de228261a0e5eae6976168299f5e41485481db9e/libraries/ortbConverter/processors/video.js#L49-L52)). Reading [the spec](https://github.com/InteractiveAdvertisingBureau/AdCOM/blob/develop/AdCOM%20v1.0%20FINAL.md#list_plcmtsubtypesvideo) I'd say yes, but I also wouldn't get why this new field is necessary.

We can also try to figure out that separately if it needs more discussion.

bwschmidt:
The reason i didn't set it based on `context === 'instream'` is because from reading https://iabtechlab.com/blog/march-2023-update-to-openrtb-is-now-ready-for-implementation/

_We have all become accustomed to video ad experiences, more often than not, we know them to be the price of admission for free content on the web; however unless the user has navigated to the page explicitly to watch the video player, like in a Hulu or YouTube style video environment, it’s possible that these ads are not commanding the user’s attention. If the small video player only plays ads without content, it is universally understood to be outstream (in-banner, in-article, in-feed, or interstitial/slider/floating). The logical conclusion for an advertiser would be that the remaining category, in-stream, would be reserved for large format, full attention, OTT-style video content. However, the legacy OpenRTB definition of in-stream left a lot of room for interpretation: “Played before, during or after the streaming video content that the consumer has requested.” What has the consumer requested? Does the video content have to be relevant to the article? Is visiting the article enough for the video player to be considered “requested”?_

_This gray area has allowed many publishers to adapt their video players so they fall into the gray area of that in-stream definition instead of the clear outstream bucket. If the player contains content and not just ads, regardless of quality, relevance, or subject matter, they can now argue that the ads are in-stream, commanding up to 50 percent higher bid prices in today’s market and significantly increasing the resulting revenue. But do brands targeting in-stream video and getting served in small, muted players alongside articles really understand what they’re buying?_

_By re-defining video placement categories to reflect the current state of the video marketplace, we aim to allow much-needed clarity for both the sell side and the buy side so that the value of premium, full-focus web video content is preserved, and other web video advertising can be clearly described._

it seems that they really want publishers to correctly label instream video using this new field. My thought is that we will want publishers to purposely set `plcmt = 1` to be sure they have read the new guidelines and are sure the supply meets the new requirements.

patmmccann (approved):

PR #9841, opened 2023-04-20
UserId module: avoid eids passing and empty array to s2s bidder

ialex:
## Type of change

- [ X ] Bugfix
- [ X ] Refactoring (no functional changes, no api changes)

## Description of change

We started seeing an increase on server to server bidder errors on PBS (Java) server caused by Prebid sending an empty array on user.eids 
```No auction context or bid request (auction). Errors:[Invalid request format: 
request.user.eids must contain at least one element or be undefined]```
<!-- For new bidder adapters, please provide the following
```
{
  bidder: '<bidder name>',
  params: {
    // ...
  }
}
```

## Other information
<!-- References to related PR or issue #s, @mentions of the person or team responsible for reviewing changes, etc. -->

### DIFF
--- a/modules/prebidServerBidAdapter/index.js
+++ b/modules/prebidServerBidAdapter/index.js
@@ -593,7 +593,9 @@ function emitNonBids(seatnonbid, auctionId) {
  * @param {array} newEidPermissions
  */
 function setEidPermissions(newEidPermissions) {
-  eidPermissions = newEidPermissions;
+  if (newEidPermissions.length > 0) {
+    eidPermissions = newEidPermissions;
+  }
 }
 getPrebidInternal().setEidPermissions = setEidPermissions;
 
--- a/test/spec/modules/userId_spec.js
+++ b/test/spec/modules/userId_spec.js
@@ -1583,7 +1583,9 @@ describe('User ID', function () {
         setSubmoduleRegistry([sharedIdSystemSubmodule]);
         let eidPermissions;
         getPrebidInternal().setEidPermissions = function (newEidPermissions) {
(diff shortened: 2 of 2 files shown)

### CODE REVIEW

File: modules/prebidServerBidAdapter/index.js [hunk 1]
```diff
@@ -593,7 +593,9 @@ function emitNonBids(seatnonbid, auctionId) {
  * @param {array} newEidPermissions
  */
 function setEidPermissions(newEidPermissions) {
```
dgirardi, line 595: I don't think this would cause an empty user.ext.eids. This only affects `ext.prebid.data.eidpermissions` and is already [guarded against empty values](https://github.com/prebid/Prebid.js/blob/de228261a0e5eae6976168299f5e41485481db9e/modules/prebidServerBidAdapter/ortbConverter.js#L62-L69).

dgirardi (commented):
i'd like to take a step back here; as far as I can tell, an empty `user.eids` is perfectly valid ORTB - the spec does not put that restriction - and this is, very clearly IMO, a problem in server.

If this is "fixed" client side, some versions of Prebid will continue sending valid requests that will be invalidly rejected. If that's really what we want, [this is where I think the change should go](https://github.com/prebid/Prebid.js/blob/de228261a0e5eae6976168299f5e41485481db9e/modules/userId/index.js#L1086-L1091) (but I have done no testing).

ialex:
> Allow empty EIDs field prebid/prebid-server#2712

Looks like it is going to be fixed on Server side https://github.com/prebid/prebid-server/issues/2712 im fine closing this.

PR #9847, opened 2023-04-24
Prebid 8: add bidResponse before emitting the event BID_RESPONSE

olafbuitelaar:
## Type of change

- [ x] Bugfix
- [ ] Feature
- [ ] New bidder adapter   
- [ ] Build related changes
- [ ] CI related changes

- [ ] Other

## Description of change
Right now when an `auction` instance receives a bid, it triggers the event `BID_RESPONSE` and then add's it to the internal collection `_bidsReceived`. This becomes problematic when you depend on the bid already available in the auction. I think the order should be reversed, so that it first stores the bid, and then triggers the `BID_RESPONSE` event. I don't expect this to break existing functionalities.

## Other information
original changes: https://github.com/prebid/Prebid.js/pull/2067

### DIFF
--- a/src/auction.js
+++ b/src/auction.js
@@ -579,8 +579,8 @@ export function addBidToAuction(auctionInstance, bidResponse) {
   setupBidTargeting(bidResponse);
 
   useMetrics(bidResponse.metrics).timeSince('addBidResponse', 'addBidResponse.total');
-  events.emit(CONSTANTS.EVENTS.BID_RESPONSE, bidResponse);
   auctionInstance.addBidReceived(bidResponse);
+  events.emit(CONSTANTS.EVENTS.BID_RESPONSE, bidResponse);
 
   doCallbacksIfTimedout(auctionInstance, bidResponse);
 }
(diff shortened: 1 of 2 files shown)


dgirardi:
>  This becomes problematic when you depend on the bid already available in the auction

this change would be problematic when you depend on the bid not being already in the auction, which is IMO more likely, simply because that's how it worked until now. 

To be clear I'm not against this but I'm not confident it's not breaking (we have no way to tell what people do in their event handlers). What is your use case?

olafbuitelaar:
I agree, that's why i put it mildly, it was just based on the fact all tests pass, and that the remaining functionality is private. But i've no way to know for sure of course. My use case, is that on the bid-response event, i would want to be able to render the ad directly, which currently fails, because `renderAd` cannot find it.

dgirardi:
I agree that this order makes more sense, and I can't think of a good reason why you'd want to keep the current one. But I'm still worried about changing the public API - what do you (and @patmmccann) think about putting this in version 8 out of an abundance of caution? And in the meanwhile,

>  render the ad directly, which currently fails, because renderAd cannot find it

I think you should be able to work around it with a 0-delay timeout, e.g. changing `pbjs.onEvent('bidResponse', handler)` to `pbjs.onEvent('bidResponse, () => setTimeout(handler))`

patmmccann:
We're working on some stuff to render ads directly, eg in the event of a gam outage, asking @mmoschovas to take a look as he can best comment on if this might affect someone attempting a similar effort.

mmoschovas (approved):

PR #9855, opened 2023-04-25
Uol Bid Adapter: Initial release

uoladtech:
## Type of change
- [x] New bidder adapter

## Description of change
Add UOL new Bidder

- Contact email: [l-dev-techops@uolinc.com](mailto:l-dev-techops@uolinc.com)
```
{
  bidder: 'uol',
  params: {
    placementId: '13144370'
  }
}
```

### DIFF
--- a/modules/uolBidAdapter.js
+++ b/modules/uolBidAdapter.js
@@ -94,7 +94,7 @@ const SCRIPT_TAG_START = '<script';
 const VIEWABILITY_URL_START = /\/\/cdn\.adnxs\.com\/v|\/\/cdn\.adnxs\-simple\.com\/v/;
 const VIEWABILITY_FILE_NAME = 'trk.js';
 const GVLID = 32;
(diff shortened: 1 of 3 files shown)


Fawke (commented):
@uoladtech, I'm looking at the files and the code for this adapter is very similar to the Appnexus bid adapter code and you're also using the same endpoint. 

Do you want to be included as an alias of the Appnexus adapter instead of making this separate submission? I'm not against it, but I'm just trying to understand if there's a use case for having this separate file and adapter.

uoladtech:
> @uoladtech, I'm looking at the files and the code for this adapter is very similar to the Appnexus bid adapter code and you're also using the same endpoint.
> 
> Do you want to be included as an alias of the Appnexus adapter instead of making this separate submission? I'm not against it, but I'm just trying to understand if there's a use case for having this separate file and adapter.

Hello, we talked to the Xandr team (AppNexus) and we're going to use an alias for their adapter.
I will be closing this pull request and opening a new one to simplify the process and not have multiple commits associated with it.

PR #9857, opened 2023-04-26
Realvu analytics adapter: fix test failures on Edge

dgirardi:
## Type of change

- [x] Build related changes

## Description of change

The realvu analytics adapter sets up a relatively expensive interval which is not stopped when its tests are done, causing Edge to hang and fail tests runs.


dgirardi:
I am not certain that properly tearing down will be enough - because we have seen failures within the realVu test suite, that logic may be too expensive even for its own tests. Ideally the adapter would not start running until it's enabled so that the tests could at least dial down how often it runs (@tchibirev). 

It's also unclear to me if it's OK to phone home without the pub having enabled the adapter, which is what I think the adapter does right now (although the code is not that easy to decipher):

https://github.com/prebid/Prebid.js/blob/bff1f412f7e91e6eccc87b0779c8dce26288b0d2/modules/realvuAnalyticsAdapter.js#L77

https://github.com/prebid/Prebid.js/blob/bff1f412f7e91e6eccc87b0779c8dce26288b0d2/modules/realvuAnalyticsAdapter.js#L153-L167

FYI @patmmccann

dgirardi:
Updated to remove the realVu test suite entirely with the plan to remove the adapter in v8.

ChrisHuie (approved):

PR #9872, opened 2023-04-28
Prebid 8: Renaming TrustPid User ID module to Utiq

jkthomas:
## Type of change

## Description of change

Only renaming of the module, all of the interested parties were already informed about the timeline.
We would need for this change to be **LIVE** on 1st of June, 2023 because of the legal requirements. Would that be possible?

## Other information

Documentation PR at https://github.com/prebid/prebid.github.io/pull/4528


patmmccann (approved):

patmmccann:
You should modify bidders pulling your id by name eg https://github.com/prebid/Prebid.js/blob/467f78f211775702ac61e2ade86ed393ada26b9a/modules/ixBidAdapter.js#L78

patmmccann (approved):
actually you did that, my mistake

PR #9877, opened 2023-04-30
Adnuntius Bid Adaptor: Better support for Deals

richard-adnuntius:
## Type of change
- [x] Feature

## Description of change

Note: This change includes TWO commits and can be separated into two separate pull requests if required. The first commit only removes inactive commented-out code and has no functional consquences at all.

This change adds better support for deals to the Adnuntius Bid Adaptor.

It adds a new `maxDeals` parameter, which defaults to `0`, that allows the adaptor to return responses for both the highest bidding advertisement AND (when `maxDeals` > `0`) deal advertisements.

If deal advertisements are returned, they are reported using one of the new bidder code aliases:

- `adndeal1`
- `adndeal2`
- `adndeal3`
- `adndeal4`
- `adndeal5`

The usage of these aliases allows for the deals and the highest bidder to all be returned to the publisher's ad server.

## Other information

This change introduces a new **optional** parameter, which will require an update to the adaptor documentation (will be submitted shortly) but it does not affect functionality for existing users.

### DIFF
--- a/modules/adnuntiusBidAdapter.js
+++ b/modules/adnuntiusBidAdapter.js
@@ -9,7 +9,6 @@ const ENDPOINT_URL = 'https://ads.adnuntius.delivery/i';
 const ENDPOINT_URL_EUROPE = 'https://europe.delivery.adnuntius.com/i';
 const GVLID = 855;
 const DEFAULT_VAST_VERSION = 'vast4'
-// const DEFAULT_NATIVE = 'native'
 
 const checkSegment = function (segment) {
   if (isStr(segment)) return segment;
@@ -28,32 +27,6 @@ const getSegmentsFromOrtb = function (ortb2) {
   }
   return segments
 }
-
-// function createNative(ad) {
-//   const native = {};
-//   const assets = ad.assets
-//   native.title = ad.text.title.content;
-//   native.image = {
-//     url: assets.image.cdnId,
-//     height: assets.image.height,
-//     width: assets.image.width,
-//   };
-//   if (assets.icon) {
-//     native.icon = {
-//       url: assets.icon.cdnId,
-//       height: assets.icon.height,
-//       width: assets.icon.width,
-//     };
-//   }
-
-//   native.sponsoredBy = ad.text.sponsoredBy?.content || '';
-//   native.body = ad.text.body?.content || '';
-//   native.cta = ad.text.cta?.content || '';
-//   native.clickUrl = ad.destinationUrls.destination || '';
-//   native.impressionTrackers = ad.impressionTrackingUrls || [ad.renderedPixel];
-
-//   return native;
-// }
 
 const handleMeta = function () {
   const storage = getStorageManager({ bidderCode: BIDDER_CODE })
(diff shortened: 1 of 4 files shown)


modules/adnuntiusBidAdapter.js
```diff
@@ -146,50 +119,66 @@ export const spec = {
...
     const adUnits = serverResponse.body.adUnits;
+
+    function buildAdResponse(ad, adUnit, isDeal, dealCount) {
+      let destinationUrls = ad.destinationUrls || {};
+      let advertiserDomains = [];
+      for (const value of Object.values(destinationUrls)) {
+        advertiserDomains.push(value.split('/')[2])
+      }
+      let adResponse = {
+        bidderCode: isDeal ? BIDDER_DEAL_ALIAS + (dealCount || 1) : BIDDER_CODE,
```
patmmccann, line 130: This isn't allowed, it breaks aliasing

patmmccann (changes requested):

patmmccann:
Generally this whole workflow is not the story of thing we like bespoke in each adapter. What's wrong with existing configuration parameters to send multiple bids and deals?

modules/adnuntiusBidAdapter.js
richard-adnuntius: Can you clarify which part of this breaks aliasing? I have tested this code and it functionally "works", but perhaps it breaks things in a more subtle way that I don't understand. 

Is the issue that the request is made using the `adnuntius` bidder code, but the response can come back with both an `adnuntius` bidder code and also an `adndeal1` and `adndeal2` etc bidder code?

richard-adnuntius:
> Generally this whole workflow is not the sort of thing we like bespoke in each adapter. What's wrong with existing configuration parameters to send multiple bids and deals?

The issue with the current configuration to send multiple bids and deals is that it only allows for one bid per adaptor.

If I tried to send two bids for Adnuntius using the same bidder code, they would both have to use the same key to represent their associated ad (`hb_adid_adnuntius`) so it would not be possible to determine which ad to render if one of the bids was selected as a winner.

richard-adnuntius:
I have converted this to a draft for now whilst we await feedback and, simulataneously, work on a different implementation that will be less bespoke :)

richard-adnuntius:
I would still very much like to get some feedback/answers to my questions, but please also see this newly submitted pull request from my colleague: https://github.com/prebid/Prebid.js/pull/9943

modules/adnuntiusBidAdapter.js
patmmccann: as you can see in #10232, your changes broke aliasing. Publishers should be able to choose any bidder code they want. Apologies for just getting back to this, but when it was marked as draft it fell completely off my radar.

patmmccann:
> The issue with the current configuration to send multiple bids and deals is that it only allows for one bid per adaptor.

This is just not true, see https://docs.prebid.org/dev-docs/modules/multibid.html

It seems your changes have already gotten through, and I think that is my fault for falling off the conversation after it got marked as draft. We don't like bidders doing the same thing a bunch of different ways, it makes for publisher implementation nightmares.

PR #9881, opened 2023-05-01
Adloox server video: mock out blob URLs in tests

dgirardi:
## Type of change

- [x] Build related changes

## Description of change

One test case from adloox often fails on edge; mocking out object URLs seems to fix it.


dgirardi:
FYI @jimdigriz - I think this might mean that the blob workflow has performance issues on Edge, or possibly its tests do. 

Unfortunately I was never able to reproduce it consistently (and edge is the hardest browser for me to work on), but from what I could tell it behaves as if it hangs; ajax callbacks run _after_ the test has already gone over the very generous 5 second timeout.

ChrisHuie (approved):

PR #9896, opened 2023-05-03
Distinguish ids from multiple providers

3link:
## Type of change

- [X] Other

## Description of change
The goal of this PR is to sketch an implementation proposal for allowing multiple user id modules to expose the same kind of identifiers. E.g. two modules exposing UID2. The idea is that a config - like the one below - provides priority in case a collision happens. 
```
pbjs.setConfig({
    userSync: {
        idPriority: {
            uid2: ['uid2', 'liveIntentId']
        }
...
```


modules/liveIntentIdSystem.js
```diff
@@ -178,22 +178,10 @@ export const liveIntentIdSubmodule = {
...
       }
 
-      if (value.bidswitch) {
-        result.bidswitch = { 'id': value.bidswitch }
-      }
-
-      if (value.medianet) {
-        result.medianet = { 'id': value.medianet }
+      if (Object.keys(value).length > 0) {
+        result.lipb = value
```
3link, line 184: This is breaking. Now, in case there is a `lipb` field in the result, there will not be necessarily a `lipbid`. If the general approach in this PR is approved, I will adjust the bid adapters accessing `lipbid`.

patmmccann: @jdwieland8282 as fyi

jdwieland8282: I don't understand what breaks or why, can you describe the issue a little more?

3link: @jdwieland8282 The "breaking" part is that in the current implementation the existence of `lipb` implies the existence `lipb.lipbid`. This would not be the case with the changes above. I though I saw bid adapters relying on that fact. Scanned through adapters using `lipb.lipbid` one more time and it looks like they are actually not affected by this change. Probably misread the code. Sorry for confusion.

skocheri1 (approved):

modules/userId/index.js
```diff
@@ -453,7 +453,7 @@ function getSubmoduleId(submodules, sourceName) {
     return {};
   }
   const submodule = submodules.filter(sub => isPlainObject(sub.idObj) &&
-    Object.keys(sub.idObj).length && USER_IDS_CONFIG[Object.keys(sub.idObj)[0]]?.source === sourceName);
+    Object.keys(sub.idObj).length && USER_IDS_CONFIG[Object.keys(sub.idObj)[0]]?.find(conf => conf.source === sourceName));
```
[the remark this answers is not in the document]
dgirardi: actually, since this is only used in here: 

https://github.com/prebid/Prebid.js/blob/d1b228c4a6e4cf6f5fa8a7899c51acbb5a5f4115/modules/userId/index.js#L656-L658

I think what we should do instead is reverse the filtering - instead of finding the module for a given source to get its EIDs, it makes more sense now to get all EIDs and filter them by source (since they can come from different modules).

patmmccann:
I'm pretty okay with this approach; we should also remember that the user ID method is in the public api and encourage the 57 adapters not using the eids method of the advantages to changing. It isn't really discriminating against an adapter if they can all change implementations at any time.

We hope live intent may help in this, letting their ssp partners know to prefer the eids method. 

We should make sure this information is well published.

modules/userId/index.js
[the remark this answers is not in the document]
jdwieland8282: exactly, "get all EIDs and filter them by source". If there is a collision, i.e., 2 or more eids with the same source check the provider and choose the one where the provider and source are the same. The question then becomes what do we do when one eid has no provider value?

deepthivenkat:
@patmmccann I can help come up with a public facing documentation explaining this.

modules/userId/index.js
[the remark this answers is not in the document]
3link: The `eids#USER_IDS_CONFIG` could be adjusted to always have a global `provider` for each user id that is then used as `provider` for all `uids` extracted based on it. That way every `eid/uid` would have a `provider`.  For example:

```
  'lipb': {
  	provider: 'liveintent.com',
  	extractors: [
	    // liveIntentId
	    {
	      getValue: function(data) {
	        return data.lipbid;
	      },
	      source: 'liveintent.com',
	      atype: 3,
	      getEidExt: function(data) {
	        if (Array.isArray(data.segments) && data.segments.length) {
	          return {
	            segments: data.segments
	          };
	        }
	      }
	    },

	    // uid2
	    {
	      source: 'uidapi.com',
	      atype: 3,
	      getValue: function(data) {
	        return data.uid2;
	      }
	    }
	  ]
	}
```
And `eids#createEidObjects` would take care of building the `uid.ext` using the global `provider` value and output 

```
[{
    source: 'liveintent.com',
    uids: [{
        id: 'some-random-id-value',
        atype: 1
    }],
    ext: {
        provider: 'liveintent.com',
        segments: ['s1', 's2']
    }
},
{
    source: 'uidapi.com',
    uids: [{
        id: 'some-random-id-value',
        atype: 3,
        ext: { 
            provider: 'liveintent.com' 
        }
    }]
}]
```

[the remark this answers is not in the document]
3link: One related question: what if there is a collision and neither of the providers is the owner of the id? 

Maybe it would make sense to return an `eid` with multiple `uids` when `getUserIdsAsEidBySource` is called and there is a collision?

For example: `getUserIdsAsEidBySource('uidapi.com')` would then return

```
[{
    source: 'uidapi.com',
    uids: [{
        id: 'some-random-id-value',
        atype: 3,
        ext: { provider: 'uidapi.com' }
    },
    {
        id: 'some-random-id-value',
        atype: 3,
        ext: { provider: 'liveintent.com' }
    }]
}]
```
when a UID2 could be resolved via LI's and the UID2 adapter. The caller then decides which `uid` to pick if there are multiple. We also can make sure that the `uid` of the id owner is always at index `0` in the `uids` array.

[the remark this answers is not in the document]
3link: > currently getUserIdsAsEidBySource returns a single object, and it's part of the public API, so that's not really up for change. if there's more than one EID we have to pick one. 

I agree that the API should not be changed in this regard. `getUserIdsAsEidBySource` should return one single object - one `eid`. An `eid`, however, can have an array of `uids`. My proposal is rather about - in case of collisions - exposing multiple `uids` with their providers within the `eid` that is returned. By providing the "provider" field the `uids` become distinguishable and the caller can decide what id to use in case there are multiple based on the provider.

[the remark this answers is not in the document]
3link: > Can we make this configurable? Can the pub declare a list of priorities in the core user id module?
@jdwieland8282 Sounds good. 

@jdwieland8282 What do you think, should that config allow to provide priority per identifier type - e.g. `uid2` from `uidapi.com` has higher priority than `uid2` from `liveintent.com`? Or should it be a config that defines that all identifiers from one user id module have priority over same types of identifiers exposed by any other module with lower priority? I personally tend towards the first option as it is more fine grained. 

Also, I would apply this prioritization logic when building the `eids` array, not only in `getUserIdsAsEidBySource`. Are we on the same page here?

3link:
@jdwieland8282 @dgirardi @deepthivenkat 

Adjusted the PR to make submodule priority configurable per user id. Also, closed [the magnite id PR](https://github.com/prebid/Prebid.js/pull/9828) and added magnite id handling here.

3link:
The e2e test failures seem not to be caused by the changes in this PR. Ran the e2e test from local machine successfully.

jdwieland8282:
I'm not qualified to code review this, but maybe @dgirardi can when he has a free moment. I can review the docs pr does that exist yet @deepthivenkat ?

modules/userId/index.js
```diff
@@ -481,15 +482,40 @@ function getCombinedSubmoduleIdsForBidder(submodules, bidder) {
...
+function getPrioritizedCombinedSubmoduleIds(submodules) {
+  const combinedIdStates = submodules
     .filter(i => isPlainObject(i.idObj) && Object.keys(i.idObj).length)
     .reduce((carry, i) => {
       Object.keys(i.idObj).forEach(key => {
-        carry[key] = i.idObj[key];
+        const currentIdPriorityByName = idPriority[key]?.reverse().indexOf(i.submodule.name);
+        const currentIdPriorityByAlias = idPriority[key]?.reverse().indexOf(i.submodule.aliasName);
+        // eslint-disable-next-line
+        const currentIdPriority = currentIdPriorityByName ? currentIdPriorityByName : currentIdPriorityByAlias ? currentIdPriorityByAlias : - 1
```
dgirardi, line 502: I don't think this works (and I don't see it in a test case) - assuming you meant "use the name first, then the alias". 
Because of how boolean casting works, this will always use the name (every value except `0` is truthy),  except when it was the last element in the priority array (reversed becomes `0`), in which case it will always use the alias, even if it was not in the array.

I am not sure that considering aliases here is necessary, but if we do, my approach would be to "resolve" them when the id priority is passed to `setConfig`; that would also help with performance. By that I mean, instead of just taking in `idPriority` verbatim from user input, do some processing so that when you get here `idPriority[key]` is already reversed, and does not contain any aliases - only the proper names.

modules/userId/index.js
```diff
@@ -481,15 +482,40 @@ function getCombinedSubmoduleIdsForBidder(submodules, bidder) {
...
+/**
+ * @param {SubmoduleContainer[]} submodules
+ */
+function getPrioritizedCombinedSubmoduleIds(submodules) {
+  const combinedIdStates = submodules
     .filter(i => isPlainObject(i.idObj) && Object.keys(i.idObj).length)
     .reduce((carry, i) => {
       Object.keys(i.idObj).forEach(key => {
-        carry[key] = i.idObj[key];
+        const currentIdPriorityByName = idPriority[key]?.reverse().indexOf(i.submodule.name);
```
dgirardi, line 499: actually, doing the reverse only once may be a requirement and not just a suggestion - because IIRC it modifies the array in place so on the second call you'd actually be restoring it to the original order.

modules/userId/index.js
3link: @dgirardi Good catch! Addressed both concerns - flipping priority and alias handling in https://github.com/prebid/Prebid.js/pull/9896/commits/edb04989e8d9063c7a4c99410a7b308c214688b8.

patmmccann:
@deepthivenkat This looks ready to merge once we have a docs pull drafted?

deepthivenkat:
On it. Should send it today or tomorrow

Best
Deepthi

Issue #9550, opened 2023-02-15
Deprecation of adpod category translation pre-loading


dgirardi:
[This is the section of code this refers to](https://github.com/prebid/Prebid.js/blob/e986edda7bb4649670fa5519d6a62e70824425a8/src/adapters/bidderFactory.js#L474). I don't fully understand the use case; but the flow seems to work like this:

- A bid adapter can declare a `getMappingFileInfo` method, which is [partially documented here](https://docs.prebid.org/dev-docs/bidder-adaptor.html#use-case-2-i-want-to-request-my-endpoint-to-return-bids-that-exactly-match-the-durations-i-want). Essentially, this method returns a URL.
- If `adpod.brandCategoryExclusion` is set to true, core will fetch and cache the URLs returned by every adapter's `getMappingFileInfo` at the beginning of the auction.
- Adapters can then access the data.  I could not find documentation on this, but [here is an example](https://github.com/prebid/Prebid.js/blob/06dd3e4fd933cebdc35120f286739b6f454325d1/modules/appnexusBidAdapter.js#L668)

This feature is currently being used only by:

- [Appnexus](https://github.com/prebid/Prebid.js/blob/06dd3e4fd933cebdc35120f286739b6f454325d1/modules/appnexusBidAdapter.js#L432) (@jsnellbaker) 
- [Goldbach](https://github.com/prebid/Prebid.js/blob/06dd3e4fd933cebdc35120f286739b6f454325d1/modules/goldbachBidAdapter.js#L390), an appnexus clone (@dveljovicTX)
- [Mediafuse](https://github.com/prebid/Prebid.js/blob/06dd3e4fd933cebdc35120f286739b6f454325d1/modules/mediafuseBidAdapter.js#L335), another appnexus clone. Also, the [file it attempts to pre-load does not exist](https://acdn.adnxs-simple.com/prebid/mediafuse-mapping/mappings.json) (@balajimediafuse)

Since it looks like appnexus is the only _intentional_ use, I believe this should be deprecated, and moved from core into the appnexus adapter (and possibly its clones).

jsnellbaker:
Hi @patmmccann I checked internally and we'll be able to do the work here.  I'll follow-up with @dgirardi closer to when the 8.0 branch is ready to confirm scope and plans for the other adapters involved.

patmmccann:
Thanks very much! Branch is open

PR #9898, opened 2023-05-03
Prebid 8 - deprecate bidder category translation feature in core

jsnellbaker:
## Type of change
- [x] Other

## Description of change
Fix for #9550

This PR removes the category translation feature in core (used by bidders, not by the adserver which still lives in `modules/categoryTranslation.js`) and updates the corresponding missing logic in a few adapters that actually used the removed code.

Instead of loading a mapping file specified by the bidder, the mapping data was directly added to the adapter code.


modules/goldbachBidAdapter.js
```diff
@@ -86,7 +86,103 @@ const NATIVE_MAPPING = {
 };
 const SOURCE = 'pbjs';
 const MAX_IMPS_PER_REQUEST = 15;
-const mappingFileUrl = 'https://acdn.adnxs-simple.com/prebid/appnexus-mapping/mappings.json';
+const TRANSLATION_MAPPING = {
+  '1': 'IAB20-3',
```
dgirardi, line 90: If you extract this somewhere under `libraries` it would be re-used across the 3 adapters that need it. It'd make a difference only for those that include more than one appnexus clone in the same bundle, I don't know if that's common.

src/adapters/bidderFactory.js
```diff
@@ -8,22 +8,17 @@ import {isValidVideoBid} from '../video.js';
...
   logError,
   logWarn,
   parseQueryStringParameters,
   parseSizesInput,
-  timestamp,
   uniques
 } from '../utils.js';
-import {ADPOD} from '../mediaTypes.js';
-import {getHook, hook} from '../hook.js';
+import {hook} from '../hook.js';
```
dgirardi, line 21: `storage` (just below the fold) should be unnecessary now, I'd like to get rid of it since it was one of the few grey area cases where we have "lax" GDPR enforcement.

I think it's kept alive by tests, otherwise the linter should have complained.

jsnellbaker: I see what you mean.  I'll make another push to remove that and check for any tests.

modules/goldbachBidAdapter.js
jsnellbaker: Thanks for the suggestion, I'll look into this.

src/adapters/bidderFactory.js
jsnellbaker: Both changes have been pushed.  I searched through the project and I couldn't find any imported references or uses of the `storage` export from the bidderFactory file.  If you think there some references still, please let me know.

dgirardi (approved):

patmmccann (approved):

PR #9901, opened 2023-05-03
airgridRtdProvider: use provided tag insertion method `loadExternalScript`

naripok:
## Type of change

- [ ] Feature
- [ ] New bidder adapter   
- [ ] Build related changes
- [ ] CI related changes
- [ ] Other

## Description of change

Use `loadExternalScript` instead of `appendChild` to insert the ad tag.
Fixes issue https://github.com/prebid/Prebid.js/issues/8421

CC @ydennisy


naripok:
Seems like our changes are breaking an seemingly unrelated module (mass).
I looked around and can't make sense of how the changes here would affect this module in particular.
Maybe somebody has a good hint/insight about how to fix it?

patmmccann:
@dgirardi could you assist this commiter?

ydennisy:
@naripok we have had some feedback that we should be updating bid request object in a different way.

Ref: https://docs.prebid.org/dev-docs/add-rtd-submodule.html#reqBidsConfigObj

- Recommended: use one of these [First Party Data](https://docs.prebid.org/features/firstPartyData.html) conventions
- For global first party data, including bidder-specific data, modify the reqBidsConfigObj as shown [below](https://docs.prebid.org/dev-docs/add-rtd-submodule.html#reqBidsConfigObj)

dgirardi:
The MASS test failure should be fixed with https://github.com/prebid/Prebid.js/pull/9939

naripok:
> Collaborator

Thank you so much @dgirardi!

dgirardi:
Regarding how to pass audience data to bidders, I believe you had it very close in an old version:

https://github.com/AirGrid/Prebid.js/blame/0eb76d7620c3bfccd254340fbc189985373fda07/modules/airgridRtdProvider.js

where that is modifying the `ortb2Fragments` parameter in the way recommended by the [docs](https://docs.prebid.org/dev-docs/add-rtd-submodule.html#reqBidsConfigObj); the only missing piece is what to do about appnexus - since it does not understand `user.ext.data`.

My suggestion is to restore the old logic setting `user.ext.data` (*without* using `mergeBidderConfig`), and expand it to also update `user.keywords` with the appnexus-specific `perid=<audienceId>`.

naripok:
> Regarding how to pass audience data to bidders, I believe you had it very close in an old version:
> 
> https://github.com/AirGrid/Prebid.js/blame/0eb76d7620c3bfccd254340fbc189985373fda07/modules/airgridRtdProvider.js
> 
> where that is modifying the `ortb2Fragments` parameter in the way recommended by the [docs](https://docs.prebid.org/dev-docs/add-rtd-submodule.html#reqBidsConfigObj); the only missing piece is what to do about appnexus - since it does not understand `user.ext.data`.
> 
> My suggestion is to restore the old logic setting `user.ext.data` (_without_ using `mergeBidderConfig`), and expand it to also update `user.keywords` with the appnexus-specific `perid=<audienceId>`.

Thank you very much for the help, @dgirardi.
I've restored the old logic and extended it for setting `ortb2.user.keywords` with the `appnexus` values, removing `mergeBidderConfig` usage.
If you have a second, would you let me know if it looks good?
@ydennisy, please, take a look too if you can.

modules/airgridRtdProvider.js
```diff
@@ -85,24 +85,19 @@ function setAudiencesToAppNexusAdUnits(adUnits, audiences) {
  * @param {Array} audiences
  * @return {{}} a map from bidder code to ORTB2 config
  */
-export function setAudiencesAsBidderOrtb2(rtdConfig, audiences) {
+export function getAudiencesAsBidderOrtb2(rtdConfig, audiences) {
   const bidders = deepAccess(rtdConfig, 'params.bidders');
   if (!bidders || bidders.length === 0 || !audiences || audiences.length === 0) return;
 
+  const agOrtb2 = {}
+  deepSetValue(agOrtb2, 'ortb2.user.ext.data.airgrid', audiences);
```
dgirardi, line 93: this is [out of spec](https://github.com/InteractiveAdvertisingBureau/openrtb2.x/blob/main/2.6.md#3221---object-data-) - `data` should be an array, sorry for not catching this earlier.

naripok: Oh, I saw it and missed anyway...
Thank you for the catch! Let me fix that.

naripok: Wait, sorry about the confusion @dgirardi, but we're using `user.ext.data` here, not `user.data`, which is an array.
Should we be using `user.data` instead?
I saw it was present [here](https://docs.prebid.org/features/firstPartyData.html) as an obj and we were using it also in the previous implementation, but I'm not sure anyway.

dgirardi: `user.ext.data` is the safer choice, it has been "promoted" to `user.data` in ORTB version 2.6 but afaik that hasn't been widely adopted yet.

naripok: Thanks for the explanation!
Does this means that we're good, then? :sweat_smile:

dgirardi: Sorry, I misinterpreted your question - I believe they use the same format, and the spec still applies; it's just the location that is different. I am however not very knowledgeable on how community extensions work so I may be wrong (@patmmccann ?)

ydennisy: @patmmccann @dgirardi what are your thoughts on merging the changes for this PR for which it had been opened - purely to use `loadExternalScript` and we can remove the changes to data setting and happy to update that in a new PR to the most recent spec.

I would like to ensure we get Xandr to take a look to ensure that we do not have a broken module once more :)

dgirardi: No objection to forking the PR, but I'd prefer then if you rolled this back to just the `loadExternalScript` change so that the incorrect values in `user.ext` are not forgotten.

naripok: @dgirardi, I've reverted the data setting changes so this PR includes only the `loadExternalScript` changes.

Now, for the data setting method, as per last feedback, I understand that the changes introduced in [this PR](https://github.com/prebid/Prebid.js/pull/9720) are bad, and we should revert those to the previous setting method, right?
I'm just a little bit confused with all the back and forth we had, so, I'm sorry to ask, but could you please again be very specific in which should be the desired data setting method you'd like to see for the next PR?
Is [this](https://github.com/AirGrid/Prebid.js/blame/0eb76d7620c3bfccd254340fbc189985373fda07/modules/airgridRtdProvider.js) the preferred way to go forward?

dgirardi: the method is the one you just had in the [previous commit](https://github.com/prebid/Prebid.js/pull/9901/commits/b2af96037b883c1aab66df804309eb012bc82bf7), with the difference that the data should be formatted as a ORTB `user.data` segment.

dgirardi (approved):

modules/airgridRtdProvider.js
naripok: Oh, alright, got ya!
Thanks!

naripok: @dgirardi, I've pushed the data setting changes to a PR in our fork [here](https://github.com/AirGrid/Prebid.js/pull/3). I'll open a PR here once I know it is good and we have this PR merged.
Would you mind checking out the question in [this comment](https://github.com/AirGrid/Prebid.js/pull/3#discussion_r1198070689), please?
I'm doing some guesswork there, and it would be great to have some confirmation that I'm getting it right.
Thank you very much!

ydennisy (approved):

patmmccann:
@naripok please resolve conflicts

naripok:
> @naripok please resolve conflicts

Done :)

PR #9916, opened 2023-05-08
Prebid 8: stop using transactionId as source.tid on ortb2 payloads

patmmccann:
Heavyhanded enforcement / fix for #8573 

Note the fix for not sending tids by default may overlap a bit with this work, but with this merged, there should no longer be anyone setting source.tid to imp[0].transactionId


patmmccann:
@jlquaccia feel free to commit to the [use-correct-source-tid](https://github.com/prebid/Prebid.js/tree/use-correct-source-tid) branch if you don't want this behavior for pubmatic

jlquaccia (approved):
LGTM

PR #9917, opened 2023-05-08
Prebid 8: pick keywords from FPD in addition to bid params

dgirardi:
## Type of change

- [x] Bugfix

## Description of change

Update a number of adapters that accept a keywords bid param to also pick keywords from ortb2.

Note that these two adapters are called out in https://github.com/prebid/Prebid.js/issues/8660 but are _not_ included in this:

* Nexx360 - does not appear to accept a keywords parameter
* DeepIntent - it's accepting keywords only [as part of an ORTB `user` object](https://github.com/prebid/Prebid.js/blob/ca2f974ba269ce634554266bb90b91f0c873a889/modules/deepintentBidAdapter.js#L220-L230); I am not sure what to think about that, but it doesn't seem correct to expand only `keywords` within it.

## Other information

Closes https://github.com/prebid/Prebid.js/issues/8660

### DIFF
--- a/libraries/appnexusKeywords/anKeywords.js
+++ b/libraries/appnexusKeywords/anKeywords.js
@@ -1,0 +1,106 @@
+import {_each, getValueString, isArray, isStr, mergeDeep, isNumber} from '../../src/utils.js';
+import {getAllOrtbKeywords} from '../keywords/keywords.js';
+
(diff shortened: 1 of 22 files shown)


Fawke (approved):
LGTM!

Issue #9922, opened 2023-05-09
Legacy native implementation: why is aspect_ratios set to an object in fromOrtbNativeRequest?


msmeza:
## Type of issue
Question

## Description
In the function `fromOrtbNativeRequest` in `native.js`, `img.aspect_ratios` is set to an object, but a global search for `aspect_ratios` in this repository shows that most occurrences are expected to be of type array. Why is `aspect_ratios` set to an object and not an array in `fromOrtbNativeRequest`?

dgirardi:
I believe this is simply a bug (paging @musikele just in case you disagree)

musikele:
This is what can be found in previous documentation: 

<img width="899" alt="image" src="https://github.com/prebid/Prebid.js/assets/2239793/23b201f5-8e69-4756-8a39-3d3c0e6163a5">

Then, in the example, it is an array: 
<img width="875" alt="image" src="https://github.com/prebid/Prebid.js/assets/2239793/1286618e-9e1c-4263-98dc-40f70dd38e28">

Honestly i don't remember what pushed me to treat as an object. Probably unit tests, or probably the fact that in ORTB you cannot specify multiple aspect ratios...

PR #9923, opened 2023-05-09
Core: fix `aspect_ratios` in ortb -> legacy native asset conversion

dgirardi:
## Type of change

- [x] Bugfix

## Description of change

Fixes https://github.com/prebid/Prebid.js/issues/9922


Rothalack (approved):
Seems to be working, looks good to me.

patmmccann (approved):

PR #9925, opened 2023-05-09
Magnite Analytics: utilize performance metrics

robertrmartinez:
## Type of change
- [X] Other

## Description of change
The adapter used to take its own timestamp to resolve how long bidder latency took using date.now()

However, a recent change began to batch up events so they are emitted together.

So date.now() led to inaccurate / longer latencies in reporting when it is a no-bid.

This change utilizes the `metrics` on the bid objects in order to log the correct total and http lantencies.


musikele (approved):
LGTM

PR #9933, opened 2023-05-10
Magnite Analytics Adapter : do not rely on BID_RESPONSE 0cpm rejected bids

robertrmartinez:
## Type of change
- [X] Other

## Description of change
Addresses https://github.com/prebid/Prebid.js/issues/8956 where analytics adapters should not rely on bid response event to contain rejected 0 cpm bids

Adapter now uses the BID_REJECTED event.


modules/magniteAnalyticsAdapter.js
```diff
@@ -818,68 +865,11 @@ magniteAdapter.track = ({ eventType, args }) => {
...
-
-      // if pbs gave us back a bidId, we need to use it and update our bidId to PBA
-      const pbsBidId = (args.pbsBidId == 0 ? generateUUID() : args.pbsBidId) || (args.seatBidId == 0 ? generateUUID() : args.seatBidId);
-      if (pbsBidId) {
-        bid.pbsBidId = pbsBidId;
-      }
+      handleBidResponse(args, 'success');
+      break;
+    case BID_REJECTED:
+      const bidStatus = args.rejectionReason && args.rejectionReason.includes('floor') ? BID_REJECTED_IPF : 'rejected';
```
dgirardi, line 871: you could do a simple compare against [`CONSTANTS.REJECTION_REASON.FLOOR_NOT_MET`](https://github.com/prebid/Prebid.js/blob/bc6aa84ace332b582e9f125b4c8fa014e503e7a8/src/constants.json#L129). Not sure if core should provide that as an error code as well.

dgirardi (approved):
LGTM, with just one minor suggestion!

patmmccann:
@robertrmartinez please resolve conflicts

PR #9947, opened 2023-05-15
beOp BidAdapter: FirstPartyData management and ingest Permutive segments

sebrobert:
## Type of change
- [x] Feature

## Description of change
First implementation for BeOp Bid Adapter of the `ortb2` object getting the user data. We will then improve our documentation for our publishers to let them know that BeOp can now ingest segments from Permutive. We will add compatibility with others DMPs in the future.

### DIFF
--- a/modules/beopBidAdapter.js
+++ b/modules/beopBidAdapter.js
@@ -38,7 +38,7 @@ export const spec = {
   buildRequests: function(validBidRequests, bidderRequest) {
     const slots = validBidRequests.map(beOpRequestSlotsMaker);
     const firstPartyData = bidderRequest.ortb2;
-    const psegs = (firstPartyData && firstPartyData.user && firstPartyData.user.ext && firstPartyData.user.ext.data) ? firstPartyData.user.ext.data.p_standard : undefined;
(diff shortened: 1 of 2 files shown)


modules/beopBidAdapter.js
```diff
@@ -37,6 +37,8 @@ export const spec = {
     */
   buildRequests: function(validBidRequests, bidderRequest) {
     const slots = validBidRequests.map(beOpRequestSlotsMaker);
+    const firstPartyData = bidderRequest.ortb2;
+    const psegs = (firstPartyData && firstPartyData.user) ? firstPartyData.user.psegs : undefined;
```
patmmccann, line 41: are these on user.ext.psegs? hopefully permutive isn't making up mainline ortb fields

patmmccann: indeed, it appears they are here 
https://github.com/prebid/Prebid.js/blob/1a5ad0cdbcc7ca772ae9fd0f2cede09f8276cd13/modules/permutiveRtdProvider.js#L195

patmmccann (changes requested):
please correct the location of the user segments on the request

PR #9949, opened 2023-05-15
Criteo Bid Adapter: add eids to bid request sent to bidder

leonardlabat:
## Type of change
- [X] Feature

## Description of change
Ensure that eids from id modules are sent to bidder.

## Other information
Question to the reviewer: wouldn't it be interesting to automatically feed ortb2.user.ext.eids with id modules ids rather than expecting the bidder to grab them from bid level userIdAsEids? (sorry if my question looks stupid and there's an obvious reason we should really not do this)

### DIFF
--- a/modules/criteoBidAdapter.js
+++ b/modules/criteoBidAdapter.js
@@ -407,6 +407,7 @@ function checkNativeSendId(bidRequest) {
 function buildCdbRequest(context, bidRequests, bidderRequest) {
   let networkId;
   let schain;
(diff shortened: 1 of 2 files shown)


dgirardi:
You are correct that eids could be fed into `ortb2`; we have been slowly moving most things into it and didn't get to ids yes - they are a bit more complicated than the rest.

dgirardi (approved):
eids could be fed directly into ortb2 - it is not curren

PR #9952, opened 2023-05-15
1plus x RTD Adapter: remove bidder specific handling enforcement

patmmccann:
Partial fulfillment of https://github.com/prebid/Prebid.js/issues/8596

### DIFF
--- a/modules/1plusXRtdProvider.js
+++ b/modules/1plusXRtdProvider.js
@@ -14,7 +14,7 @@ const ORTB2_NAME = '1plusX.com'
 const PAPI_VERSION = 'v1.0';
 const LOG_PREFIX = '[1plusX RTD Module]: ';
 const OPE_FPID = 'ope_fpid'
-const LEGACY_SITE_KEYWORDS_BIDDERS = ['appnexus'];
+const LEGACY_SITE_KEYWORDS_BIDDERS = ['dummy'];
 export const segtaxes = {
   // cf. https://github.com/InteractiveAdvertisingBureau/openrtb/pull/108
   AUDIENCE: 526,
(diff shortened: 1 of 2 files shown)


gggiakas:
@bwhisp can you please check and approve the pull request here?
AppNexus special handling needs to be removed due to Prebid 8 rules (all special handling cases must be removed).

Ideally we would like to support this bid adaptor but I understand that this is up to AppNexus to change their implementation.. we can address this, if any workaround available, on a next release

modules/1plusXRtdProvider.js
```diff
@@ -14,7 +14,7 @@ const ORTB2_NAME = '1plusX.com'
 const PAPI_VERSION = 'v1.0';
 const LOG_PREFIX = '[1plusX RTD Module]: ';
 const OPE_FPID = 'ope_fpid'
-const LEGACY_SITE_KEYWORDS_BIDDERS = ['appnexus'];
+const LEGACY_SITE_KEYWORDS_BIDDERS = ['dummy'];
```
bwhisp, line 17:
```suggestion
```

rather than changing the value to dummy here we can completely remove this variable and all associated code in lines 155 to 164

patmmccann:
Alternatively, we could trash this pr and you can make the bidders you do 155-164 for come from config instead of being hardcoded. I think airgrid is doing something similar as a model.

modules/1plusXRtdProvider.js
```diff
@@ -216,14 +204,6 @@ export const updateBidderConfig = (bidder, ortb2Updates, bidderConfigs) => {
   return bidderConfigCopy;
 };
 
-const setAppnexusAudiences = (audiences) => {
```
dgirardi, line 219: removing this means that appnexus won't get audience data; I'm not sure that's the intent. Airgrid is replacing this with another kind of appnexus logic that's slightly more general, but they are not removing it entirely.

bwhisp:
Hello @dgirardi 

I will attempt to replace the appnexus logic the same way Airgrid does it. 

Do you know the PR number where they do that ?

patmmccann:
@bwhisp I marked this back as draft . This is the pr airgird is working on https://github.com/AirGrid/Prebid.js/pull/3

bwhisp:
Hello @patmmccann @dgirardi 
On the same model as the work done for AirGrid, I went ahead and removed bidder specificity from my module
A PR has been published here: https://github.com/prebid/Prebid.js/pull/10001

CC: @gggiakas

patmmccann:
closing in favor of #10001

PR #9957, opened 2023-05-16
RubiconBidAdapter: sync parseSize algorithm for isBidRequestValid and ortb conversion

musikele:
## Type of change
- [X] Bugfix

## Description of change
When video is outstream, we want to use the same sizes that were used in isBidRequestVaild.


robertrmartinez (approved):
nice looks good

PR #9959, opened 2023-05-16
Ortb Converter : add support for top level 'native' prop for PBS native admarkup

ckbo3hrk:
## Type of change

- [x] Bugfix

## Description of change

During pbjs-to-pbs integration test I noticed that ortbConverter library does not support native response with top-level property **native**, which is correct according to native 1.1/1.2 specs. I've checked and found that some PBS adaptors may return native admarkup in this way:

https://github.com/prebid/prebid-server/blob/master/adapters/datablocks/datablockstest/exemplary/native.json#L112

https://github.com/prebid/prebid-server/blob/master/adapters/pulsepoint/pulsepointtest/exemplary/native.json#L88

## Other information

https://github.com/prebid/Prebid.js/pull/8738

### DIFF
--- a/libraries/ortbConverter/processors/native.js
+++ b/libraries/ortbConverter/processors/native.js
@@ -26,6 +26,10 @@ export function fillNativeResponse(bidResponse, bid) {
       ortb = bid.adm;
     }
 
+    if (ortb?.native) {
+      ortb = ortb.native;
+    }
+
     if (isPlainObject(ortb) && Array.isArray(ortb.assets)) {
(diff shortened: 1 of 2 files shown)


dgirardi:
I was not aware of this, but I read the spec differently:

> Prior to VERSION 1.1, the native response’s root node was an object with a single field
“native” that would contain the object above as its value.

so the datablocks example seems out of spec to me, and if someone is actually replying with 1.0 I'd be worried about what else is different compared to 1.2.

PR #9968, opened 2023-05-18
UID2 & EUID Modules: Add support for EUID and prefer localStorage for both modules.

lionell-pack-ttd:
## Type of change

- [ ] Bugfix
- [x] Feature
- [ ] New bidder adapter   
- [ ] Build related changes
- [ ] CI related changes

- [ ] Other

## Description of change
I've extracted most of the functionality from the UID2 module to a shared file, to allow re-use with the new EUID module.
I've added a new EUID module.
Both modules now prefer local storage - the UID2 module originally preferred cookie storage. No configs need to change - it's all backward-compatible.

## Other information
Docs change PR: https://github.com/prebid/prebid.github.io/pull/4603

### DIFF
--- a/modules/euidIdSystem.js
+++ b/modules/euidIdSystem.js
@@ -1,4 +1,3 @@
-/* eslint-disable no-console */
 /**
  * This module adds EUID ID support to the User ID module. It shares significant functionality with the UID2 module.
  * The {@link module:modules/userId} module is required.
--- a/modules/euidIdSystem.js
+++ b/modules/euidIdSystem.js
@@ -5,7 +5,7 @@
  * @requires module:modules/userId
  */
 
-import { logInfo, logWarn } from '../src/utils.js';
+import { logInfo, logWarn, deepAccess } from '../src/utils.js';
 import {submodule} from '../src/hook.js';
 import {getStorageManager} from '../src/storageManager.js';
 import {MODULE_TYPE_UID} from '../src/activities/modules.js';
@@ -18,7 +18,7 @@ const MODULE_NAME = 'euid';
 const MODULE_REVISION = Uid2CodeVersion;
 const PREBID_VERSION = '$prebid.version$';
 const EUID_CLIENT_ID = `PrebidJS-${PREBID_VERSION}-EUIDModule-${MODULE_REVISION}`;
-const GVLID = 21; // The Trade Desk
+const GVLID_TTD = 21; // The Trade Desk
 const LOG_PRE_FIX = 'EUID: ';
 const ADVERTISING_COOKIE = '__euid_advertising_token';
 
@@ -37,6 +37,16 @@ const _logWarn = createLogger(logWarn, LOG_PRE_FIX);
 
 export const storage = getStorageManager({moduleType: MODULE_TYPE_UID, moduleName: MODULE_NAME});
 
+function hasWriteToDeviceConsent(consentData) {
+  const gdprApplies = consentData?.gdprApplies === true;
+  const localStorageConsent = deepAccess(consentData, `vendorData.purpose.consents.1`)
+  const prebidVendorConsent = deepAccess(consentData, `vendorData.vendor.consents.${GVLID_TTD.toString()}`)
+  if (gdprApplies && (!localStorageConsent || !prebidVendorConsent)) {
+    return false;
+  }
+  return true;
+}
+
 /** @type {Submodule} */
 export const euidIdSubmodule = {
   /**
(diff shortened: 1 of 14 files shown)


modules/euidIdSystem.md
```diff
@@ -0,0 +1,54 @@
+## UID 2.0 User ID Submodule
```
patmmccann, line 1: this appears to be docs for uid2 not euid

lionell-pack-ttd: Fixed.

modules/uid2IdSystem.js
```diff
@@ -6,13 +6,17 @@
...
 import {getStorageManager} from '../src/storageManager.js';
 import {MODULE_TYPE_UID} from '../src/activities/modules.js';
 
+// RE below lint exception: UID2 and EUID are separate modules, but the protocol is the same and shared code makes sense here.
+// eslint-disable-next-line prebid/validate-imports
+import { Uid2GetId, Uid2CodeVersion } from './uid2IdSystem_shared.js';
+
 const MODULE_NAME = 'uid2';
-const MODULE_REVISION = `1.0`;
+const MODULE_REVISION = Uid2CodeVersion;
```
patmmccann, line 19: If the uid2 needs to be replaced with the EUID in europe, why does the uid2 module continue to have a gvlid?

modules/uid2IdSystem.js
```diff
@@ -183,29 +71,24 @@ export const uid2IdSubmodule = {
    * @returns {uid2Id}
    */
   getId(config, consentData) {
-    const result = getIdImpl(config, consentData);
+    const mappedConfig = {
```
patmmccann, line 74: if this isnt supposed to work in europe, should it kick out when gdpr applies?

modules/uid2IdSystem.js
patmmccann: and why is it diff in the two modules?

modules/uid2IdSystem.js
```diff
@@ -183,29 +71,24 @@ export const uid2IdSubmodule = {
...
+      apiBaseUrl: config?.params?.uid2ApiBase ?? UID2_BASE_URL,
+      paramToken: config?.params?.uid2Token,
+      serverCookieName: config?.params?.uid2Cookie ?? config?.params?.uid2ServerCookie,
+      storage: config?.params?.storage ?? 'localStorage',
+      clientId: UID2_CLIENT_ID,
+      internalStorage: ADVERTISING_COOKIE
+    }
+    _logInfo(`UID2 configuration loaded and mapped.`, mappedConfig);
+    const result = Uid2GetId(mappedConfig, storage, _logInfo, _logWarn);
```
patmmccann, line 83: not passing consent at all to this function is a bit concerning; as we discussed in chat, you have to recheck it, you can't assume it hasnt changed since the token was generated.

patmmccann (changes requested):
Please handle consent more explicitly in both modules; also discuss why euid is different than the uid2 and why a second module is needed?

patmmccann:
> As an aside, I looked up gvl id 887, that is the prebid gvl id. You shouldnt be using that for uid2, please change that to either ttd or vendorless module; i think the latter is more appropriate.

Or just delete the gvlid from the uid2 module altogether so only euid gets called in europe

modules/uid2IdSystem.js
lionell-pack-ttd: I've removed the gvlid from the UID2 module and updated the EUID module to the gvlid for The Trade Desk (as operator of EUID).

modules/uid2IdSystem.js
lionell-pack-ttd: I've updated it to bail out if GDPR applies.

modules/uid2IdSystem.js
lionell-pack-ttd: It now bails out prior to this if GDPR applies. For the similar piece of code in the EUID module, the same - it checks consent and bails out prior to this call.

modules/euidIdSystem.js
```diff
@@ -0,0 +1,100 @@
...
+  },
+
+  /**
+   * performs action to obtain id and return a value.
+   * @function
+   * @param {SubmoduleConfig} [configparams]
+   * @param {ConsentData|undefined} consentData
+   * @returns {euidId}
+   */
+  getId(config, consentData) {
```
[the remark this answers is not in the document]
lionell-pack-ttd: Updated to check consent and bail out if it's not available.

lionell-pack-ttd:
> Please handle consent more explicitly in both modules; also discuss why euid is different than the uid2 and why a second module is needed?

I've added consent handling to both modules. I had a few internal discussions about the possibility of using a single module to handle it, but I got strong pushback - they're different ID spaces with potentially different configuration, and different legal handling as well - not all EUID participants will be UID2 participants, and vice versa.

patmmccann:
Nice work!

patmmccann (approved):

PR #9970, opened 2023-05-18
Sirdata RTD Module : bidder specific handling removal

nouchy:
move bidder specific handling to ortb2

## Type of change

- [X ] Refactoring (no functional changes, no api changes)

## Description of change
bidder specific handling removal : use ortb2 instead

### DIFF
--- a/modules/sirdataRtdProvider.js
+++ b/modules/sirdataRtdProvider.js
@@ -134,13 +134,24 @@ export function getSegmentsAndCategories(reqBidsConfigObj, onDone, moduleConfig,
     });
 }
 
-export function setGlobalOrtb2Sda(ortb2Fragments, data, segtaxid, cattaxid) {
+export function pushToOrtb2(ortb2Fragments, bidder, data, segtaxid, cattaxid) {
   try {
     if (!isEmpty(data.segments)) {
-      applyGlobalOrtb2Sda(ortb2Fragments, 'user', data.segments, segtaxid);
+      if (segtaxid) {
+        setOrtb2Sda(ortb2Fragments, bidder, 'user', data.segments, segtaxid);
+      } else {
+        setOrtb2(ortb2Fragments, bidder, 'user.ext.data', {sd_rtd: {segments: data.segments}});
+      }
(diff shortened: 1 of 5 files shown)


modules/sirdataRtdProvider.js
```diff
@@ -370,57 +369,24 @@ export function addSegmentData(reqBids, data, moduleConfig, onDone) {
           case 'msq_classic':
           case 'msq_max':
           case '366_apx':
```
dgirardi, line 371: I think I'm missing some context, if the intent of this is to remove bidder-specific handling why is there a giant switch statement for a bunch of bidders?

nouchy:
Hello, the idea here is to avoid data shared with bidders directly. First step us to work through ortb2, and when bidders are ready to use SDA… please see this comment : https://github.com/prebid/Prebid.js/issues/8596#issuecomment-1548354140

dgirardi:
> Hello, the idea here is to avoid data shared with bidders directly. First step us to work through ortb2, and when bidders are ready to use SDA… please see this comment : [#8596 (comment)](https://github.com/prebid/Prebid.js/issues/8596#issuecomment-1548354140)

I am not following; is the data different depending on the bidder? if yes, which bidders get which data should be configurable (not based in a switch statement in code); if not, you shouldn't need the switch statement. RTD modules should try to stay as agnostic as possible regarding who gets the data.

nouchy:
We will, for sure, but we need them to be ready and move step by step, sorry. We « cleaned » what we can but still work on it…

nouchy:
you’re right, most data IDs (segs, cats) are private, and we’re moving towards SDA and segtax ids. But the bidders don’t follow at same rate :(

patmmccann:
@nouchy if proxistore for example isn't moving along, feel free to move your logic specific to them into their adapter.

nouchy:
> @nouchy if proxistore for example isn't moving along, feel free to move your logic specific to them into their adapter.

Sure, I'll do it

patmmccann (approved):

PR #9977, opened 2023-05-19
Prebid 8: remove shutdown company from appnexus aliases

patmmccann:
As reported in the press, EMX and Defy Media are no longer companies.

### DIFF
--- a/modules/appnexusBidAdapter.js
+++ b/modules/appnexusBidAdapter.js
@@ -101,9 +101,7 @@ export const spec = {
   gvlid: GVLID,
   aliases: [
     { code: 'appnexusAst', gvlid: 32 },
-    { code: 'emxdigital', gvlid: 183 },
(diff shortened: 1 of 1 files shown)


jsnellbaker:
I am double-checking with some folks internally before giving the go ahead here.  Will follow-up when I get confirmation.

Thanks @patmmccann

astephensxandr:
Hi @patmmccann I don't think it's that simple in one case. In the case of Defy Media, we can definitely remove that alias. 

I'm not a bankruptcy expert, but I think that 'no longer companies' isn't an accurate characterization of the process where EMX is concerned. My understanding from internal conversations is that we're going to continue treating our relationship with them like any other member in our marketplace until the bankruptcy process provides more clarity.

patmmccann:
@astephensxandr thanks, will modify

jsnellbaker (approved):

PR #9989, opened 2023-05-23
YieldlabBidAdapter updated native asset mapping

nkloeber:
## Type of change

- [x] Feature

## Description of change

YieldlabBidAdapter updated native asset mapping of type icon for adserver responses. Only adds the `native.icon` if a value is provided in the response.

## Other information

One approving review from a Yieldlab team member required. (@rey1128, @Mufas61, @brushmate or @kippsterr)

### DIFF
--- a/modules/yieldlabBidAdapter.js
+++ b/modules/yieldlabBidAdapter.js
@@ -186,13 +186,21 @@ export const spec = {
           // there may be publishers still rely on it
           bidResponse.adUrl = `${ENDPOINT}/d/${matchedBid.id}/${bidRequest.params.supplyId}/?ts=${timestamp}${extId}${gdprApplies}${gdprConsent}${pvId}`;
           bidResponse.mediaType = NATIVE;
-          const nativeImageAssetObj = find(matchedBid.native.assets, asset => isMainImage(asset));
+          const nativeIconAssetObj = find(matchedBid.native.assets, asset => isImageAssetOfType(asset, 1));
+          const nativeImageAssetObj = find(matchedBid.native.assets, asset => isImageAssetOfType(asset, 3));
           const nativeImageAsset = nativeImageAssetObj ? nativeImageAssetObj.img : { url: '', w: 0, h: 0 };
           const nativeTitleAsset = find(matchedBid.native.assets, asset => hasValidProperty(asset, 'title'));
           const nativeBodyAsset = find(matchedBid.native.assets, asset => hasValidProperty(asset, 'data'));
           bidResponse.native = {
             title: nativeTitleAsset ? nativeTitleAsset.title.text : '',
             body: nativeBodyAsset ? nativeBodyAsset.data.value : '',
+            ...nativeIconAssetObj?.img && {
+              icon: {
+                url: nativeIconAssetObj.img.url,
+                width: nativeIconAssetObj.img.w,
+                height: nativeIconAssetObj.img.h,
+              },
+            },
             image: {
               url: nativeImageAsset.url,
               width: nativeImageAsset.w,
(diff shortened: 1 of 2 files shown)


modules/yieldlabBidAdapter.js
```diff
@@ -186,13 +186,21 @@ export const spec = {
           // there may be publishers still rely on it
           bidResponse.adUrl = `${ENDPOINT}/d/${matchedBid.id}/${bidRequest.params.supplyId}/?ts=${timestamp}${extId}${gdprApplies}${gdprConsent}${pvId}`;
           bidResponse.mediaType = NATIVE;
-          const nativeImageAssetObj = find(matchedBid.native.assets, asset => isMainImage(asset));
+          const nativeIconAssetObj = find(matchedBid.native.assets, asset => isImageAssetOfType(asset, 1));
```
kippsterr, line 189: Just a tiny thought: Having constants for the image asset types would allow us gett rid of the magic numbers and at the same time making the calling place even more specific.

`const nativeIconAssetObj = find(matchedBid.native.assets, asset => isImageAssetOfType(asset, ICON));`

Or, I know you intentionally made the previous `isMainImage` generic, we go back to having separate and specific methods for the different image asset types. Why? Because we could then have even more concise calls like e.g. `find(matchedBid.native.assets, iconAsset)`.

What do you think? At the end that is just cosmetics and not really mission critical but there is the potential that it just looks nicer :)

modules/yieldlabBidAdapter.js
```diff
@@ -186,13 +186,21 @@ export const spec = {
           // there may be publishers still rely on it
```
kippsterr, line 186: This is also rather a cosmetic suggestion but as we access `matchedBid.native` and `matchedBid.native.assets` quite a few times, we could fetch them once from the `matchedBid` object and directly use them later which cleans the calling places up a bit.

```suggestion
          const { native } = matchedBid;
          const { assets } = native;

          // there may be publishers still rely on it
```

So with this suggestion and the comment below we might transform

```
const nativeIconAssetObj = find(matchedBid.native.assets, asset => isImageAssetOfType(asset, 1));
```

to something like

```
const iconAssetObj = find(assets, iconAsset);
```

modules/yieldlabBidAdapter.js
```diff
@@ -186,13 +186,21 @@ export const spec = {
...
           const nativeImageAsset = nativeImageAssetObj ? nativeImageAssetObj.img : { url: '', w: 0, h: 0 };
           const nativeTitleAsset = find(matchedBid.native.assets, asset => hasValidProperty(asset, 'title'));
           const nativeBodyAsset = find(matchedBid.native.assets, asset => hasValidProperty(asset, 'data'));
           bidResponse.native = {
             title: nativeTitleAsset ? nativeTitleAsset.title.text : '',
             body: nativeBodyAsset ? nativeBodyAsset.data.value : '',
+            ...nativeIconAssetObj?.img && {
+              icon: {
+                url: nativeIconAssetObj.img.url,
+                width: nativeIconAssetObj.img.w,
```
kippsterr, line 200: Similar to the default for the main image above `w` and `h` instead of `width` and `height` have to be used for the response object.

modules/yieldlabBidAdapter.js
```diff
@@ -517,14 +525,14 @@ function hasValidProperty(obj, propName) {
...
+ * Checks if an asset object is of the given type.
  *
  * @param {Object} asset - The asset object to check.
- * @returns {boolean} Returns true if the object has a property img.type with a value of 3, otherwise false.
+ * @param {number} type - The type number to compare.
+ * @returns {boolean} Returns true if the object has a property img.type with a value of the given type, otherwise false.
  */
-function isMainImage(asset) {
-  return asset?.img?.type === 3
+function isImageAssetOfType(asset, type) {
```
kippsterr, line 534: If we leave it generic and still want to shorten the calling place we could do something like:

```
function isImageAssetOfType(type) { // ICON = 1, MAIN = 3
  return asset = asset?.img?.type === type;
}
```

and then call it with `const iconAssetObj = find(assets, isImageAssetOfType(ICON));`

Otherwise just two methods as suggested above.

test/spec/modules/yieldlabBidAdapter_spec.js
```diff
@@ -567,15 +576,24 @@ describe('yieldlabBidAdapter', () => {
       expect(result[0].native.image.url).to.equal('https://localhost:8080/yl-logo100x100.jpg');
       expect(result[0].native.image.width).to.equal(100);
       expect(result[0].native.image.height).to.equal(100);
+      expect(result[0].native.icon.url).to.equal('https://localhost:8080/assets/favicon/favicon-16x16.png');
+      expect(result[0].native.icon.width).to.equal(16);
```
kippsterr, line 580: Comparing with the [spec on page 22](https://www.iab.com/wp-content/uploads/2018/03/OpenRTB-Native-Ads-Specification-Final-1.2.pdf) I think this needs to be changed to `w` and `h` and also for the probably wrong expectation above.

patmmccann (approved):

PR #9990, opened 2023-05-23
Topics Module: add support for updated taxonomy

patmmccann:
This pr prepares for additional taxonomies to be released. The IABTL has already reserved these codes.


patmmccann:
already covered by this test it seems? https://github.com/prebid/Prebid.js/blob/82c23967622d0a25d9a84d56f7a6461828172950/test/spec/modules/topicsFpdModule_spec.js#L111

PR #10001, opened 2023-05-25
1plus x RTD Adapter: remove bidder specific handling enforcement (DC 3634)

bwhisp:
## Type of change

- [ ] Bugfix
- [ ] Feature
- [ ] New bidder adapter   
- [ ] Code style update (formatting, local variables)
- [ ] Refactoring (no functional changes, no api changes)
- [ ] Build related changes
- [ ] CI related changes

- [ ] Does this change affect user-facing APIs or examples documented on http://prebid.org?
- [x] Other : Fulfilling new requirement for Prebid v8

## Description of change
Updates ortb2 user data setting method to conform the spec
A draft of this work has been done here https://github.com/prebid/Prebid.js/pull/9952

PR #10005, opened 2023-05-26
Weborama RTD Module : start Bidder specific handling removal

peczenyj:
move bidder specific handling to ortb2 except appnexus

## Type of change

## Description of change

bidder specific handling removal : use ortb2 instead

related to

- sirdata pull request https://github.com/prebid/Prebid.js/pull/9970
- prebid issue https://github.com/prebid/Prebid.js/issues/8596

### DIFF
--- a/modules/weboramaRtdProvider.js
+++ b/modules/weboramaRtdProvider.js
@@ -690,7 +690,7 @@ class WeboramaRtdProvider {
     const bidder = bidderAliasRegistry[bid.bidder] || bid.bidder;
 
     if (bidder == 'appnexus') {
-      this.#handleAppnexusBid(bid, profile);
+      this.#handleAppnexusBid(reqBidsConfigObj, bid, profile);
     }
   }
 
@@ -708,16 +708,19 @@ class WeboramaRtdProvider {
   /** handle appnexus/xandr bid
    * @method
    * @private
+   * @param {Object} reqBidsConfigObj
+   * @param {Object} reqBidsConfigObj.ortb2Fragments
+   * @param {Object} reqBidsConfigObj.ortb2Fragments.bidder
    * @param {Object} bid
-   * @param {Object} bid.params
-   * @param {Object} bid.params.keyword
+   * @param {Object} bid.parameters
    * @param {Profile} profile
    * @returns {void}
    */
   // eslint-disable-next-line no-dupe-class-members
-  #handleAppnexusBid(bid, profile) {
+  #handleAppnexusBid(reqBidsConfigObj, bid, profile) {
     const base = 'params.keywords';
     this.#assignProfileToObject(bid, base, profile);
+    // this.#setBidderOrtb2(reqBidsConfigObj.ortb2Fragments?.bidder, bid.bidder, base, profile);
   }
 
   /** handle generic bid via ortb2 arbitrary data
--- a/modules/weboramaRtdProvider.js
+++ b/modules/weboramaRtdProvider.js
@@ -693,9 +693,6 @@ class WeboramaRtdProvider {
       case 'appnexus':
         this.#handleAppnexusBid(bid, profile);
         break;
-      case 'pubmatic':
-        this.#handlePubmaticBid(bid, profile);
-        break;
       case 'smartadserver':
         this.#handleSmartadserverBid(bid, profile);
         break;
@@ -729,34 +726,6 @@ class WeboramaRtdProvider {
   #handleAppnexusBid(bid, profile) {
     const base = 'params.keywords';
     this.#assignProfileToObject(bid, base, profile);
-  }
-
-  /** handle pubmatic bid
-   * @method
-   * @private
-   * @param {Object} bid
-   * @param {Object} bid.params
-   * @param {string} bid.params.dctr
-   * @param {Profile} profile
-   * @returns {void}
-   */
-  // eslint-disable-next-line no-dupe-class-members
-  #handlePubmaticBid(bid, profile) {
-    const sep = '|';
-    const subsep = ',';
-
-    bid.params ||= {};
-
-    const data = bid.params.dctr || '';
-    const target = new Set(data.split(sep).filter((x) => x.length > 0));
-
-    Object.entries(profile).forEach(([key, values]) => {
-      const value = values.join(subsep);
-      const keyword = `${key}=${value}`;
-      target.add(keyword);
-    });
-
-    bid.params.dctr = Array.from(target).join(sep);
   }
 
   /** handle smartadserver bid
--- a/modules/weboramaRtdProvider.js
+++ b/modules/weboramaRtdProvider.js
@@ -756,6 +756,7 @@ class WeboramaRtdProvider {
    * @param {Profile} profile
    * @returns {void}
    */
+  // eslint-disable-next-line no-dupe-class-members
   #setBidderOrtb2(bidderOrtb2Fragments, bidder, path, profile) {
     const base = `${bidder}.${path}`;
     this.#assignProfileToObject(bidderOrtb2Fragments, base, profile)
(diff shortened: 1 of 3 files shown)


modules/weboramaRtdProvider.js
```diff
@@ -716,94 +707,24 @@ class WeboramaRtdProvider {
...
-    if (isBoolean(metadata.user)) {
-      const section = metadata.user ? 'visitor' : 'inventory';
-      const base = `params.${section}`;
-      this.#assignProfileToObject(bid, base, profile);
+  #handleSiteLegacyKeywordsBidders(reqBidsConfigObj, bid, profile, metadata) {
+    if (metadata.user) {
+      this.#setBidderOrtb2(reqBidsConfigObj.ortb2Fragments?.bidder, bid.bidder, 'user.keywords', profile);
     } else {
-      logMessage(`SKIP bidder '${bid.bidder}', data from '${metadata.source}' is not defined as user or site-centric`);
+      this.#assignProfileToObject(bid, 'params.keywords', profile);
```
jsnellbaker, line 727: I'm trying to follow the updates that were made in this area with the recent commits.  

If this code is still primarily there for the appnexus bid adapter (as implied by the `LEGACY_SITE_KEYWORDS_BIDDERS` if check made earlier to call this `handleSiteLegacyKeywordsBidders` function), you could move it or generalize it further by adding this data to the equivalent ortb2 fields (like site.keywords) instead of the bidder.params location.  Our adapter should be able to read that data from the ortb2 location, and it should allow other bidders to read that data as well.

If this is serving some other purpose still, can you please clarify?  Thanks.

peczenyj: Hello

I just copy this `LEGACY_SITE_KEYWORDS_BIDDERS` from `1plusXRtdProvider.js`

jsnellbaker: I'm not sure I follow - the 1plusXRtdProvider appears to be writing a modified form of segment data into the appnexus auction keywords field (which they technically don't need to do anymore, since they could write it to the ortb2.site.keywords field instead).  

So while I may understand if you want to do an extra step/conversion of your data to a keyword-like format (like 1plusXRtdProvider), you shouldn't need to target the bidder keywords field like they do.  

The only reason I'm bringing up this point was that the PR seemed to be designed to move your data to the ortb2 fields.  So that should be possible even for us (appnexus), instead of using a bidder param field (which feels like it's against the grain).

peczenyj: so if I understand well, in the case of appnexis I must push to `ortb2.site.keywords` or `ortb2.user.keywords` depending if it is site-centric or user-centric, right?

jsnellbaker: For our endpoint, we just have either request level keywords or adunit/placement specific keywords.  

For our request level keywords they read the ortb2 keyword fields from various sources, collating them together based on keynames/keyvalues to send along in the auction request.  For the adunit keywords, we read from the adUnit ortb2Imp field that publishers can choose to populate, if they have unique data for that slot.  So it doesn't as much matter which of the fields are used in the ortb2 keywords properties, we'll check them and pull in the data if it's there.

peczenyj: how about now: db1fc089081660753a0a15348508c6be7aba42ff ?

patmmccann:
#10001 just merged and might be helpful, thanks so much!

modules/weboramaRtdProvider.js
patmmccann: Does it make sense to use this library and base on prebid 8 ? [libraries/appnexusKeywords/anKeywords.js](https://github.com/prebid/Prebid.js/pull/10071/files#diff-ac43bc058ac47dc8abc3602fe21d9b07ebb2adde809fac523d55099618c3166d)

dgirardi: I did some digging on this and the problems are related, `keywords` as set from this revision are removed by the fpdValidation module because of incorrect type:

image

modules/weboramaRtdProvider.js
```diff
@@ -689,19 +697,8 @@ class WeboramaRtdProvider {
...
-      case 'pubmatic':
-        this.#handlePubmaticBid(bid, profile);
-        break;
-      case 'smartadserver':
-        this.#handleSmartadserverBid(bid, profile);
-        break;
-      case 'rubicon':
-        this.#handleRubiconBid(bid, profile, metadata);
-        break;
+    if (bidder === APPNEXUS_BIDDER) {
```
dgirardi, line 700: To summarize what I think is missing:

 - appnexus (and clones) now parse keywords as `key=value` pairs. so this should serialize objects like `{ webo_ctx: ["A", "B"], webo_ds: ["C", "D"]}` down to a single string such as `webo_ctx=A,webo_ctx=B,webo_ds=C,webo_ds=D`.
​
- because appnexus has clones, this should ideally not check for `appnexus` specifically. You could take the list of bidders to set keywords for as config, or you  could even set these keywords on all the bidders you want to set segments for (non-appnexus bidders will probably have no use for them but they shouldn't hurt).

peczenyj:
Hello

Since there is a lot of changes that may be added on this MR, I decided to split the development in 3 parts

1. remove all dedicated code EXCEPT appnexus
2. fix the current appnexus implementation (to be confirmed)
3. improve the appnexis by using ortb2 configuration

I just rollback the last 4 commits.

Thanks a log guys

modules/weboramaRtdProvider.js
```diff
@@ -689,19 +689,8 @@ class WeboramaRtdProvider {
...
-      case 'pubmatic':
-        this.#handlePubmaticBid(bid, profile);
-        break;
-      case 'smartadserver':
-        this.#handleSmartadserverBid(bid, profile);
-        break;
-      case 'rubicon':
-        this.#handleRubiconBid(bid, profile, metadata);
-        break;
+    if (bidder == 'appnexus') {
```
patmmccann, line 692: can you change this to if (bidder == 'valuereadfromconfig')

modules/weboramaRtdProvider.js
```diff
@@ -689,19 +689,8 @@ class WeboramaRtdProvider {
...
+      this.#handleAppnexusBid(reqBidsConfigObj, bid, profile);
```
patmmccann, line 693: can rename this function to handleSegmentsInKeywordsBidder()

modules/weboramaRtdProvider.js
```diff
@@ -719,92 +708,19 @@ class WeboramaRtdProvider {
...
-  // eslint-disable-next-line no-dupe-class-members
-  #handleRubiconBid(bid, profile, metadata) {
-    if (isBoolean(metadata.user)) {
-      const section = metadata.user ? 'visitor' : 'inventory';
-      const base = `params.${section}`;
-      this.#assignProfileToObject(bid, base, profile);
-    } else {
-      logMessage(`SKIP bidder '${bid.bidder}', data from '${metadata.source}' is not defined as user or site-centric`);
-    }
+    // this.#setBidderOrtb2(reqBidsConfigObj.ortb2Fragments?.bidder, bid.bidder, base, profile);
```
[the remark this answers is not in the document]
patmmccann: > we can't read from the segment related ortb data fields at this time. 

@astephensxandr if this is indeed true for the foreseeable future, Microsoft should expect disruption in its ongoing compatibility with RTD modules. Are you saying you guys are comfortable with this?

patmmccann:
@peczenyj This PR achieves your first goal; thanks so much! We do plan to release prebid 8 imminently however and need all specific handling removed for you to be in it. Please let us know if you prefer us to remove your adapter or your appnexus handling.

patmmccann (approved):

modules/weboramaRtdProvider.js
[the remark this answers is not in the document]
patmmccann: @peczenyj update here! Appnexus pr should make this much easier, we're delaying prebid 8 to accommodate

Issue #4607, opened 2019-12-16
Temporarily disable certain checks in emoteevBidAdapter file 


jsnellbaker:
## Type of issue
Bug?

## Description
There have been a number of test runs with browserstack that have failed due to some of the unit tests from the `emoteevBidAdapter_spec.js` file.

Though it varies which particular test fails, the failures have occurred at the `getConfig` call checks that are part of the 'side effects' describe block.  See link below for an example of the types of failures that were observed.
https://circleci.com/gh/prebid/Prebid.js/5483

We're temporarily disabling these specific checks from these test to help improve the stability of the browserstack tests (see #4594)

@piotr-yuxuan Could you try to take a look into these errors to see why they may be failing and if there's some alternative check that you could run?

PR #10012, opened 2023-05-30
Geolocation RTD module: initial release

AdmixerTech:
## Type of change

- [x] Feature

## Description of change
add geolocation rtd provider

## Other information
prebid.github.io pull [#4607](https://github.com/prebid/prebid.github.io/pull/4607)


modules/geolocationRtdProvider.js
```diff
@@ -0,0 +1,64 @@
...
+  });
+  if (auctionDelay > 0) {
+    setTimeout(complete, auctionDelay);
+  }
+  if (!waitForIt) complete();
+  function complete() {
+    if (done) return;
+    done = true;
+    geolocation && adUnits && adUnits.forEach((unit) => {
+      unit.bids && unit.bids.forEach(bid => bid.geolocation = cleanObj(geolocation));
```
dgirardi, line 41: Instead of tacking this on to bid requests, it's IMO better to fill this out as FPD (setting the fields in `requestBidsObject.ortb2Fragments.global.device.geo`). That way a number of bidders will automatically forward that data, and [activity controls](https://github.com/prebid/Prebid.js/pull/9802) will automatically control the accuracy. 

Otherwise this relies entirely on more work in bid adapters so that they may pick this up. Looking at the [spec](https://github.com/InteractiveAdvertisingBureau/openrtb2.x/blob/main/2.6.md#objectgeo) it looks like most of the fields have a spot with the exception of those about velocity, I'd still put those somewhere under `geo.ext`.

modules/geolocationRtdProvider.js
```diff
@@ -0,0 +1,64 @@
...
+        'heading',
+        'latitude',
+        'longitude',
+        'speed'
+      ]);
+      complete();
+    });
+  });
+  if (auctionDelay > 0) {
+    setTimeout(complete, auctionDelay);
```
patmmccann, line 34: the rtd parent module handles auction delay and waitforit, no need for you to

patmmccann: see https://github.com/AdmixerTech/Prebid.js/blob/c57da4b7f2c7d4cb543e076e34aec70018443bae/modules/rtdModule/index.js#LL306C1-L308C1

modules/geolocationRtdProvider.js
patmmccann: i'd prefer the browser doesn't even ask for precise geo unless https://github.com/InteractiveAdvertisingBureau/Global-Privacy-Platform/blob/87935c4b574e974ff20ff5c1b9641374878031e2/Sections/US-National/IAB%20Privacy%E2%80%99s%20National%20Privacy%20Technical%20Specification.md?plain=1#L229 permitted it; not geo gets asked for then rounded without the proper consent. Does activityControls allow for that? we'd want to prevent invoking this altogether, not just manipulating its input to bid requests.

AdmixerTech: @patmmccann just to be clear, i need to check consent or just ensure that consent string provided ?

dgirardi: @AdmixerTech after discussing this offline, this should _not_ use the browser API if the `transmitPreciseGeo` activity is denied. Activities are very new, but the basic idea is to do something like

```javascript
if (isActivityAllowed(ACTIVITY_TRANSMIT_PRECISE_GEO, activityParams(MODULE_TYPE_RTD, 'geolocation'))) {
   // call browser geo api
}
```

Some examples of the activity API in use for reference:

https://github.com/prebid/Prebid.js/blob/bdd56a742c07df9dd9a2025e93b1649ce187d0bb/src/storageManager.js#L32-L35

https://github.com/prebid/Prebid.js/blob/bdd56a742c07df9dd9a2025e93b1649ce187d0bb/src/adapterManager.js#L281-L285

jdwieland8282:
The ID PMC agrees that the module should first check the CMP api to determine whether the user has revoked location sharing before prompting the user for permission to geo locate.

patmmccann:
Generally this is looking pretty good; we should add some discussion of interaction with activity control to the doc

dariaboyko:
Hi! Here are the docs https://github.com/prebid/prebid.github.io/pull/4680

dariaboyko:
hi! the docs are ready. can you please put a label on that this pr is ready to be merged?

patmmccann (approved):

patmmccann:
Thanks! @dgirardi could you give another sign off?

dariaboyko:
Hi @dgirardi ! Can you please check this pull request? Also I have a question when next release is planned?

dgirardi (approved):

Issue #10008, opened 2023-05-29
Video Module Doesn't Push Winning Bids to Auction Winning Bids State


fowler446:
## Type of issue
Bug

## Description
The Prebid video module doesn't add winning bids to the `_winningBids` state of an auction. This is because the `pbjsRenderAd` function doesn't run when a video ad unit plays:
https://github.com/prebid/Prebid.js/blob/master/src/prebid.js#L528

## Steps to reproduce
Implement the video module on a test page and have some bids win the auction and play. Then type `pbjs.getAllWinningBids()` and notice how it's empty.

dgirardi:
I am wondering if this is limited to the video module, or if it's a problem for video in general. The video module calls `markWinningBidAsUsed` when it knows a bid has been rendered, which is the [same flow](https://docs.prebid.org/dev-docs/publisher-api-reference/markWinningBidAsUsed.html) used before the video module was introduced.

However, that seems to only mark the bid as "rendered", and not as winning, and I don't know why - it seems to have been like this since the [beginning](https://github.com/prebid/Prebid.js/pull/2777). Paging @mkendall07 for the slim chance of some context.

karimMourra:
I had already fixed this in https://github.com/prebid/Prebid.js/pull/9612/files but it seems like my changes were deleted. the git blame for that line indicates a commit from 2018.

karimMourra:
@dgirardi i do think this fix should work. i am concerned though that other fixes might have been lost when #9612 was undone

dgirardi:
I am also confused by your commit. With some digging it looks like it was erased with a heavy handed [conflict merge](https://github.com/prebid/Prebid.js/commit/40cdc5e2e5150304ae0107d868ae03e1fa9940e7). Do you see anything else in there that may also be broken?

karimMourra:
I don't see anything else that looks broken in the conflict merge.

PR #9612, opened 2023-03-01
CORE:  add bid to winningBids when marking as used

karimMourra:
## Type of change

- [x] Bugfix

## Description of change
Add the winning bid to the list of winning bids when calling `markWinningBidAsUsed`.

Questions for reviewer: 

## Other information
Addresses #9561 and #9612

PR #10014, opened 2023-05-30
Core: fix `markWinningBidAsUsed` to set bid as winning and not just rendered

dgirardi:
## Type of change

- [x] Bugfix

## Description of change

Fixes https://github.com/prebid/Prebid.js/issues/10008

### DIFF
--- a/src/prebid.js
+++ b/src/prebid.js
@@ -1004,7 +1004,7 @@ if (FEATURES.VIDEO) {
     const bids = fetchReceivedBids(markBidRequest, 'Improper use of markWinningBidAsUsed. It needs an adUnitCode or an adId to function.');
 
     if (bids.length > 0) {
-      bids[0].status = CONSTANTS.BID_STATUS.RENDERED;
+      auctionManager.addWinningBid(bids[0]);
     }
   }
 }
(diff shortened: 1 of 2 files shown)


karimMourra:
@dgirardi it seems like I had already fixed this https://github.com/prebid/Prebid.js/pull/9612/files

src/prebid.js
```diff
@@ -1004,7 +1004,7 @@ if (FEATURES.VIDEO) {
     const bids = fetchReceivedBids(markBidRequest, 'Improper use of markWinningBidAsUsed. It needs an adUnitCode or an adId to function.');
 
     if (bids.length > 0) {
-      bids[0].status = CONSTANTS.BID_STATUS.RENDERED;
```
karimMourra, line 1007: When would the `status` have to be set to `rendered` ? Is there a place in Core, or would that have to be handled by the Video Module ?

dgirardi: it's part of the logic in `addWinningBid`: https://github.com/prebid/Prebid.js/blob/62200e3c43486e98a5b58bd2d59b0e687ab815a3/src/auctionManager.js#L46

PR #10028, opened 2023-06-01
Yahoo bid adapter:  User sync pixels, consent signals update

slimkrazy:
## Type of change
- [x] Feature

## Description of change

Addressing this issue picked up by Prebid maintainers: https://github.com/prebid/prebid.github.io/pull/4529/files#diff-d7318b0388c5171115d26fc62fb6cdcfd7387d65af9f8e013e7ba8c4828dd5df

User sync URLs will use the latest available consent data the moment they are fetched. Additionally, we will now consume GPP signals.

It's not a bugfix as such, so submitting as a feature.

I'll open a PR shortly against the documentation site and link to it.

### DIFF
--- a/modules/yahoosspBidAdapter.js
+++ b/modules/yahoosspBidAdapter.js
@@ -101,6 +101,37 @@ function extractUserSyncUrls(syncOptions, pixels) {
   return userSyncObjects;
 }
 
+/**
+ * @param {string} url
+ * @param {object} consentData
+ * @param {object} consentData.gpp
+ * @param {string} consentData.gpp.gppConsent
+ * @param {array} consentData.gpp.applicableSections
+ * @param {object} consentData.gdpr
+ * @param {object} consentData.gdpr.gdprConsent
+ * @param {object} consentData.gdpr.gdprApplies
+ * @param {string} consentData.uspConsent
+ */
+function updateConsentQueryParams(url, consentData) {
+  const parameterMap = {
+    'gdpr_consent': consentData.gdpr.gdprConsent,
+    'gdpr': consentData.gdpr.gdprApplies,
+    'us_privacy': consentData.uspConsent,
+    'gpp': consentData.gpp.gppConsent,
+    'gpp_sid': consentData.gpp.applicableSections? consentData.gpp.applicableSections.join(',') : ''
+  }
+
+  const existingUrl = new URL(url);
+  const params = existingUrl.searchParams;
+
+  for (const [key, value] of Object.entries(parameterMap)) {
+    params.set(key, value);
+  }
+
+  existingUrl.search = search_params.toString();
+  return existingUrl.toString();
+};
+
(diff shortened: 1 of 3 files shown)


patmmccann (approved):

modules/yahoosspBidAdapter.js
```diff
@@ -244,7 +275,9 @@ function generateOpenRtbObject(bidderRequest, bid) {
       regs: {
         ext: {
           'us_privacy': bidderRequest.uspConsent ? bidderRequest.uspConsent : '',
-          gdpr: bidderRequest.gdprConsent && bidderRequest.gdprConsent.gdprApplies ? 1 : 0
+          gdpr: bidderRequest.gdprConsent && bidderRequest.gdprConsent.gdprApplies ? 1 : 0,
+          gpp: bidderRequest.gppConsent.gppString,
+          gpp_sid: bidderRequest.gppConsent.applicableSections
```
robertrmartinez, line 280: @slimkrazy 

Hi, I work for Magnite and during version testing between 7.50.0 and 7.54.0 we noticed many of our pubs had `yahoossp` bids stop filling.

This is the only PR for the bid adapter during that time and I think it might point to this spot here


The object `gppConsent` is not guaranteed to be defined.

For example on one of our publisher sites we see it throws the following error:

image

An easy fix for this is to change this to be like so:

```js
          gpp: bidderRequest.gppConsent?.gppString,
          gpp_sid: bidderRequest.gppConsent?.applicableSections
```

Let me know if this makes sense.

I tested it using Chrome Overrides and can see bid requests going out now:

image

modules/yahoosspBidAdapter.js
```diff
@@ -101,6 +101,37 @@ function extractUserSyncUrls(syncOptions, pixels) {
...
+ * @param {string} consentData.uspConsent
+ */
+function updateConsentQueryParams(url, consentData) {
+  const parameterMap = {
+    'gdpr_consent': consentData.gdpr.consentString,
+    'gdpr': consentData.gdpr.gdprApplies ? '1' : '0',
+    'us_privacy': consentData.uspConsent,
+    'gpp': consentData.gpp.gppString,
+    'gpp_sid': consentData.gpp.applicableSections ? consentData.gpp.applicableSections.join(',') : ''
+  }
```
robertrmartinez, line 122: Also here as well I think you need to add safe checks because thee gdpr, gpp and uspConent objects might be undefined or null

```js
  const parameterMap = {
    'gdpr_consent': consentData.gdpr?.consentString,
    'gdpr': consentData.gdpr?.gdprApplies ? '1' : '0',
    'us_privacy': consentData?.uspConsent,
    'gpp': consentData.gpp?.gppString,
    'gpp_sid': consentData.gpp?.applicableSections ? consentData.gpp.applicableSections.join(',') : ''
  }
```

Something like that, (not tested)

PR #10047, opened 2023-06-05
Rise bid adapter: change session_id fallback to bidderRequestId

lasloche:
## Type of change

## Description of change

Changed the fallback id for the session_id request param from the **auctionId** to the **bidderRequestId**.

## Other information

Related to: https://github.com/prebid/Prebid.js/blob/4320362f4349551f6d947e28f1a3ed69476fbd93/modules/riseBidAdapter.js#L420


lasloche:
@patmmccann, we did a fallback to another value. Please, let us know if this one works!

Issue #9781, opened 2023-04-06
Setting Prebid Default for TID to off in prebid 8


jtollestrup:
Due to privacy concerns regarding back end stitching of data utilizing the TID, the publisher committee has elected to flip the default in Prebid for TID to off and allow each publisher to make their individual determination to turn this feature on.

dgirardi:
Is this referring to `imp.ext.tid`, `source.tid`, or both? 

For `source.tid` this needs, in general, an update to every adapter. `imp.ext.tid` would be easier to toggle centrally.

jtollestrup:
Both.

jtollestrup:
We met with Prebid legal today and they strongly recommended that this feature be defaulted off.

patmmccann:
TBD: should the config toggle be in activity control or its own top level flag?

dgirardi:
Many adapters include auctionId as part of their payload, separately from `source.tid`. This is especially common for those that do not use ORTB, but not limited to them. What should we do about that? replace them with a random UUID?

dgirardi:
Here's an alternative proposal:

 - refactor every usage of `auctionId` that is clearly interchangeable with `source.tid` as discussed above;
 - when it's not clear (see below), leave it untouched, and:
 - when transaction ID are turned off, add a check on outgoing network requests, and forcibly cancel requests that contain auctionId.

This way, when TIDs are turned on, adapters will continue to work as if nothing changed; when they are turned off, we can have good confidence that data won't be leaked.

Otherwise we don't have a good way to deal with a large number of adapters that use a pattern like this:

https://github.com/prebid/Prebid.js/blob/bc665aa7b6320c15df5553240d65a907837c162c/modules/proxistoreBidAdapter.js#L26-L30

in this case, it's clear that `transactionId` is meant to be the TID and should be protected by the new control we are introducing for 8. But we do not know what to do with `auctionId` - removing it has a good chance to unnecessarily break it.

I'll add that IMO it's too early to turn TIDs off by default - there's a good chance that it will break so many adapters that it the vast majority of publishers will be de-facto forced to turn them back on again.

dgirardi:
The enforcement on network requests could also be made into a module - if we are OK with the default being "TID are off, but we don't know if they are actually being sent out".

PR #10052, opened 2023-06-05
AMX RTB: remove use of auctionId

nickjacob:
## Type of change

- [ ] Bugfix
- [ ] Feature
- [ ] New bidder adapter   
- [ ] Build related changes
- [ ] CI related changes

- [X] Other

## Description of change

Remove use of bidderRequest.auctionId, replacing it with ortb2 "tid" / transaction ID. For compliance with: https://github.com/prebid/Prebid.js/issues/9781#event-9417751649

### DIFF
--- a/modules/amxBidAdapter.js
+++ b/modules/amxBidAdapter.js
@@ -1,5 +1,5 @@
-import {registerBidder} from '../src/adapters/bidderFactory.js';
-import {BANNER, VIDEO} from '../src/mediaTypes.js';
+import { registerBidder } from '../src/adapters/bidderFactory.js';
+import { BANNER, VIDEO } from '../src/mediaTypes.js';
 import {
   _each,
   deepAccess,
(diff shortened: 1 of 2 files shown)


modules/amxBidAdapter.js
```diff
@@ -302,8 +303,13 @@ export const spec = {
...
         };
 
+    const transactionId = deepAccess(
+      bidderRequest,
+      'ortb2.ext.tid',
+      deepAccess(bidRequests[0], 'ortb2Imp.ext.tid', generateUUID()));
+
     const payload = {
-      a: bidderRequest.auctionId,
+      a: transactionId,
```
patmmccann, line 312: Quick question @nickjacob ; do you plan to put this field in ortb2 bid requests to downstream parties ion either source.tid or imp.ext.tid? If so, imp.ext.tid should not fall back to a value that isn't shared by all bidders. The openrtb spec doesnt' allow for exchanges to make up their own value here. Perhaps you would want to pass yourself an indicator of if this value is shared or null if you can't take the null here?

modules/amxBidAdapter.js
```diff
@@ -302,8 +303,13 @@ export const spec = {
           bidRequestsCount: 0,
         };
 
+    const transactionId = deepAccess(
```
dgirardi, line 306: What are you using this for? I can't make sense of this if `payload.a` is supposed to have any meaning at all when it gets to your server:

 - `auctionid` and `transactionId` are not meant to be mixed, picking them interchangeably means you cannot identify either the transaction nor the auction.
 - if you're ok with falling back to a randomly generated value, why not always use a randomly generated value? what's the reason for anything more complicated?
 - `ortb2.ext.tid` should be `ortb2.source.tid` instead (if the above is not a problem).

dgirardi (approved):

patmmccann:
@nickjacob can you resolve conflicts?

nickjacob:
Sorry about the delay -- @patmmccann fixed merge conflict

PR #10064, opened 2023-06-07
Brightcom, Brightcom SSP bid adapter: add iiq first party data support

prBigBrother:
## Type of change

- [x] Feature

## Description of change
We have included support for first-party data (as an ID) in the iiq system. We will retrieve and save this data using local storage, but only if the "storageAllowed" setting is enabled

### DIFF
--- a/modules/brightcomBidAdapter.js
+++ b/modules/brightcomBidAdapter.js
@@ -2,6 +2,7 @@ import { getBidIdParameter, _each, isArray, getWindowTop, getUniqueIdentifierStr
 import { registerBidder } from '../src/adapters/bidderFactory.js';
 import { BANNER } from '../src/mediaTypes.js';
 import { config } from '../src/config.js';
+import { bidderSettings } from "../src/bidderSettings.js";
 
 const BIDDER_CODE = 'brightcom';
 const URL = 'https://brightcombid.marphezis.com/hb';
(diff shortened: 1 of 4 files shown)


modules/brightcomBidAdapter.js
```diff
@@ -287,4 +292,72 @@ function _getBidFloor(bid) {
   return null;
 }
 
+function _getFirstPartyData() {
```
patmmccann, line 295: this appears to be id module functionality; why are you adding to your bidder?

modules/brightcomSSPBidAdapter.js
```diff
@@ -319,4 +324,72 @@ function _getBidFloor(bid) {
   return null;
 }
 
+function _getFirstPartyData() {
```
patmmccann, line 327: don't copy paste huge code blocks into two adapters, use a library

prBigBrother: hi @patmmccann ! I'm not quite sure how libraries are organized in this project. Could you share a few examples with me? I'd greatly appreciate it!

thanks!

patmmccann: https://github.com/prebid/Prebid.js/blob/master/libraries/keywords/keywords.js is a code block repeated across multiple adapters. One of our goals to reduce build size is to reduce copy pasted code blocks. 

image

PR #10076, opened 2023-06-08
airgridRtdProvider: update ortb2 user data setting method

alexander-kislitsyn:
## Type of change

- [x] Bugfix

## Description of change

Updates ortb2 user data setting method to conform the spec.

## Other information

The change was requested [here](https://github.com/prebid/Prebid.js/pull/9901/commits/b2af96037b883c1aab66df804309eb012bc82bf7#r1197963663)


modules/airgridRtdProvider.js
```diff
@@ -141,7 +156,7 @@ export function passAudiencesToBidders(
   const audiences = getMatchedAudiencesFromStorage();
   if (audiences.length > 0) {
     setAudiencesUsingAppNexusAuctionKeywords(audiences);
```
dgirardi, line 158: this should be removed now - it was the initial goal after all.

PR #10082, opened 2023-06-08
prebid 8 - read ortb segment data in appnexus keywords library functions

jsnellbaker:
## Type of change

- [x] Feature

- [ ] Other

## Description of change
This change allows for a way to read ortb2 segment data (from the various possible locations in the ortb2 object) from identified recognized sources (based on `segtax` values) and have the segment data be transformed into a keypair format.  

This approach is based on the existing logic that several RtdProviders used to work for the appnexusBidAdapter; so it should not be entirely new functionality, just formalizing it into a central place (outside of the RtdProviders, which was a need for Prebid 8.0).

**NOTE** The values in the `segtax` map were guesses on my part from some conversations.  If there are other values that should be used, please clarify it here.

Making this a draft PR for now, in case there's a better approach to this change.

### DIFF
--- a/libraries/appnexusKeywords/anKeywords.js
+++ b/libraries/appnexusKeywords/anKeywords.js
@@ -3,7 +3,6 @@ import {getAllOrtbKeywords} from '../keywords/keywords.js';
 import {CLIENT_SECTIONS} from '../../src/fpd/oneClient.js';
 
 const ORTB_SEGTAX_KEY_MAP = {
-  0: null,
   526: '1plusx',
   527: '1plusx',
   541: 'captify_segments',
@@ -124,7 +123,7 @@ export function getANMapFromOrtbSegments(ortb2) {
     let ortbSegsArrObj = deepAccess(ortb2, path) || [];
     ortbSegsArrObj.forEach(segObj => {
       // only read segment data from known sources
-      const segtax = ORTB_SEGTAX_KEY_MAP[deepAccess(segObj, 'ext.segtax') || 0];
+      const segtax = ORTB_SEGTAX_KEY_MAP[deepAccess(segObj, 'ext.segtax')];
       if (segtax) {
         segObj.segment.forEach(seg => {
           // if source was in multiple locations of ortb or had multiple segments in same area, stack them together into an array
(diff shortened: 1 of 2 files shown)


libraries/appnexusKeywords/anKeywords.js
```diff
@@ -1,5 +1,17 @@
-import {_each, getValueString, isArray, isStr, mergeDeep, isNumber} from '../../src/utils.js';
+import {_each, deepAccess, getValueString, isArray, isStr, mergeDeep, isNumber} from '../../src/utils.js';
 import {getAllOrtbKeywords} from '../keywords/keywords.js';
+import {CLIENT_SECTIONS} from '../../src/fpd/oneClient.js';
+
+const ORTB_SEGTAX_KEY_MAP = {
+  0: null,
```
dgirardi, line 6: this is unnecessary (assuming `0` is never a valid segtax) - you'd get `undefined` instead of `null` but your check would still work.

libraries/appnexusKeywords/anKeywords.js
```diff
@@ -85,7 +97,7 @@ function convertKeywordsToANMap(kwarray) {
  * @param ortb2
  * @return {{}} appnexus-style keyword map using all keywords contained in ortb2
  */
-export function getANMapFromOrtb(ortb2) {
+export function getANMapFromOrtbKeywords(ortb2) {
```
dgirardi, line 100: I think it makes sense to simplify this to get rid of the `key=value` parsing now (and treat keywords as just simple keywords); since with this there's no need for the hack anymore - but if we release it and people start using it you'll be stuck with it for a while.

libraries/appnexusKeywords/anKeywords.js
jsnellbaker: Will push an update

libraries/appnexusKeywords/anKeywords.js
dgirardi: It's never been released - the only dependence on it comes from recent updates that are also still to be released with 8. But fair enough, it's better to undo this only after those are updated yet again to use segments instead.

dgirardi (approved):

PR #10084, opened 2023-06-09
Criteo Bid Adapter : remove usage of bidRequest.auctionId

leonardlabat:
## Type of change
- [X] Bugfix
- 
## Description of change
Removed usage of auctionId and use instead source.tid or a random uuid.


modules/criteoBidAdapter.js
```diff
@@ -427,6 +427,7 @@ function buildCdbRequest(context, bidRequests, bidderRequest) {
   let schain;
   let userIdAsEids;
   const request = {
+    id: bidderRequest.ortb2?.source?.tid || generateUUID(),
```
patmmccann, line 430: this looks a bit like openrtb, $.id should be unique to the request and can be  generateUUID() or bidderRequestId

modules/criteoBidAdapter.js
```diff
@@ -445,7 +446,7 @@ function buildCdbRequest(context, bidRequests, bidderRequest) {
       const slot = {
         impid: bidRequest.adUnitCode,
         transactionid: bidRequest.transactionId,
-        auctionId: bidRequest.auctionId,
+        auctionId: bidderRequest.ortb2?.source?.tid || generateUUID(),
```
patmmccann, line 449: Please make sure you're not passing this to dsps if you pass openrtb. This wouldn't comply with openrtb

leonardlabat: It won't be but indeed ; I've amended my change to make it completely obvious & transparent and rely exclusively on a random uuid

patmmccann (approved):

patmmccann (approved):

PR #10092, opened 2023-06-12
Prebid 8: improve `transmitTid` logic for PBS adapter

dgirardi:
## Type of change

- [x] Feature

## Description of change

This updates the logic around activity controls and in particular [TID suppression](https://github.com/prebid/Prebid.js/pull/10034) for the PBS adapter:

 - introduce new checks with component `prebid.pbsBidAdapter` for FPD going to PBS (including TIDs)
 - when `transmitTid` is not allowed for `prebid.pbsBidAdapter`, also set `ext.prebid.createtids: false` on outgoing PBS requests


patmmccann (approved):

lksharma (approved):
LGTM. Only seeing `ext.prebid.createtids: false` when `enableTIDs: false`

PR #10093, opened 2023-06-12
Prebid 8: remove bidder specific handling for 1plusX

dgirardi:
## Type of change

- [x] Other

## Description of change

- update 1plusX to no longer generate `key=value` keywords for appnexus as it will pick up the segments with https://github.com/prebid/Prebid.js/pull/10082/files
- ~remove the logic from weborama as well. since weborama is *not* generating segments, this will break their appnexus integration.~
- ~remove the `key=value` keyword parsing from appnexus and its clones.~


libraries/appnexusKeywords/anKeywords.js
```diff
@@ -60,24 +60,10 @@ export function convertKeywordStringToANMap(keyStr) {
...
-      // then check for existing key in result > if so add value to the array > if not, add new key and create value array
-      if (result.hasOwnProperty(key)) {
-        result[key].push(val);
-      } else {
-        result[key] = [val];
-      }
-    } else {
-      if (!result.hasOwnProperty(kw)) {
-        result[kw] = [];
-      }
```
jsnellbaker, line 78: To note - I am still unsure about removing this part of the functionality as it was previously in our adapter code, so it could have been used by someone for their setup and it would then stop sending the data properly to the endpoint.  

Since it seemingly was approved by others in the community, I'm just mentioning this thought; it is not a blocker for the PR's acceptance.

modules/1plusXRtdProvider.js
```diff
@@ -159,12 +159,7 @@ export const buildOrtb2Updates = ({ segments = [], topics = [] }) => {
     segment: topics.map((topicId) => ({ id: topicId })),
     ext: { segtax: segtaxes.CONTENT }
   }
-  // Currently appnexus bidAdapter doesn't support topics in `site.content.data.segment`
-  // Therefore, writing them in `site.keywords` until it's supported
-  // Other bidAdapters do fine with `site.content.data.segment`
-  const siteKeywords = topics.map(topic => `1plusX=${topic}`).join(',');
```
jsnellbaker, line 165: Can we clarify with 1plusX how they passed their segment keywords previously?  It seem like here it was with a keyname like `1plusX` but the new logic has `1plusx`.  We should follow the same formatting as it was before.

modules/weboramaRtdProvider.js
```diff
@@ -682,16 +673,6 @@ class WeboramaRtdProvider {
...
     this.#handleBidViaORTB2(reqBidsConfigObj, bid.bidder, profile, metadata);
-
-    /** @type {Object.<string,string>} */
-    const bidderAliasRegistry = adapterManager.aliasRegistry || {};
-
-    /** @type {string} */
-    const bidder = bidderAliasRegistry[bid.bidder] || bid.bidder;
-
-    if (bidder == 'appnexus') {
-      this.#handleAppnexusBid(reqBidsConfigObj, bid, profile);
```
jsnellbaker, line 693: Don't have a good place to insert this comment; but from testing weborama they set their data in the `site.ext.data` instead of the ortb segment fields, so it's not getting included/read by the new logic for the appnexus bid adapter.

Screen Shot 2023-06-13 at 8 32 40 AM

jsnellbaker (changes requested):
See some inline comments below.  

In addition for 1plusX - does their test page in the project (`1plusXRtdProviderExample.html`) include segment data?  I don't see anything coming back in the ortb2 field within the appnexus bid adapter (see attached), so the data (if it's there) is not getting passed along to the new logic and to the endpoint.

Screen Shot 2023-06-13 at 8 36 09 AM

bwhisp:
> @bwhisp fyi ; does this look cleaner to you?

I checked the changes, looks great ! We're no longer doing specific things for a single bidder adapter 
Thanks :)

modules/weboramaRtdProvider.js
dgirardi: Correct, this is intentional and called out in the OP. If we end up keeping the key=value parsing I'll also undo this.

dgirardi:
> does their test page in the project include segment data?

their endpoint doesn't, for me. so it tries to set no segments which gets validated out. Guessing that's what you see as well?

image
image

modules/1plusXRtdProvider.js
dgirardi: fixed!

jsnellbaker:
> > does their test page in the project include segment data?
> 
> their endpoint doesn't, for me. so it tries to set no segments which gets validated out. Guessing that's what you see as well?
> 
> image image

I am seeing the same; @bwhisp is there something we can do or change in the test page to get the segment data populated so we can make sure the flow/transformation of the data via this PR works?

dgirardi:
This is what appnexus sees if I mock out the network response with 
```
{
          s: ['segment1', 'segment2', 'segment3'],
          t: ['targeting1', 'targeting2', 'targeting3']
}
```

Ends up in the appnexus request as:

image

`s` (which is populated under `user.data`) is lost because it does not have a `segtax`:

https://github.com/prebid/Prebid.js/blob/41ee11b23597ccf58cfc05adff35e1390086beb8/modules/1plusXRtdProvider.js#L153-L161

@bwhisp is this intentional?

dgirardi:
updated to add the audience segtax to `user.data` .

bwhisp:
Hi @jsnellbaker @dgirardi 
The endpoint you mentioned is supposed to give us some data 
I will check it in the next few days.

patmmccann:
@bwhisp thanks! in the meantime, it appears from mocking your endpoint we're good to go. We appreciate the conversion over!

jsnellbaker (approved):
Approving this PR based on the mocked data that @dgirardi tested and provided.

bwhisp:
@patmmccann @jsnellbaker @dgirardi
we could create some data for the `/targeting` endpoint 

I might also need to change the extension `1plusXRTDProviderExample.html` to `xhtml`
Do you think it's feasible ?

PR #10100, opened 2023-06-14
Oxxion Rtd Module: add bid filtering

matthieularere-msq:
## Type of change
- [x] Feature

## Description of change
Add a bidder filtering functionnality.

### DIFF
--- a/modules/oxxionRtdProvider.js
+++ b/modules/oxxionRtdProvider.js
@@ -21,10 +21,10 @@ export const oxxionSubmodule = {
 };
 
 function init(config, userConsent) {
-  if (!config.params || !config.params.domain || !config.params.contexts || !Array.isArray(config.params.contexts) || config.params.contexts.length == 0) {
-    return false
-  }
-  return true;
+  if (!config.params || !config.params.domain) { return false }
+  if (config.params.contexts && Array.isArray(config.params.contexts) && config.params.contexts.length > 0) { return true; }
+  if (typeof config.params.threshold != 'undefined' && typeof config.params.samplingRate == 'number') { return true }
+  return false;
 }
 
 function getAdUnits(reqBidsConfigObj, callback, config, userConsent) {
--- a/modules/oxxionRtdProvider.js
+++ b/modules/oxxionRtdProvider.js
@@ -63,7 +63,7 @@ function getAdUnits(reqBidsConfigObj, callback, config, userConsent) {
         }
       });
     }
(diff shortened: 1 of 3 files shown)


modules/oxxionRtdProvider.md
```diff
@@ -23,14 +23,16 @@ Then add the oxxion Rtd module to your prebid configuration :
 pbjs.setConfig(
   ...
   realTimeData: {
-    auctionDelay: 200,
+    auctionDelay: 300,
```
patmmccann, line 26: typo on line 15 above

modules/oxxionRtdProvider.js
```diff
@@ -1,34 +1,71 @@
 import { submodule } from '../src/hook.js'
-import { deepAccess, logInfo } from '../src/utils.js'
+import { deepAccess, logInfo, logError } from '../src/utils.js'
+import MD5 from 'crypto-js/md5.js';
```
patmmccann, line 3: why not use something supported already?

modules/oxxionRtdProvider.md
matthieularere-msq: You meant to add the new use case in the description ?

patmmccann: you write rtbModule instead of rtdmodule

modules/oxxionRtdProvider.js
patmmccann: We just don't love the include, isnt sha supported in the browser? I forget. md5 is a rather large build side include I believe.

patmmccann: https://github.com/mbaersch/cyrb53-hasher is in python; fairly simple to implement in go just asking chatgpt to translate our js function to go

modules/oxxionRtdProvider.md
matthieularere-msq: Thanks, I hadn't seen it. It's fixed.

modules/oxxionRtdProvider.js
patmmccann: Here's an alternative we use at raptive  https://caniuse.com/?search=digest ; it has wide support for sha-1 and sha-256

matthieularere-msq: just like I said, I removed it for now as it's not needed yet.

patmmccann (approved):

PR #10107, opened 2023-06-16
Update topicsFpdModule

piwanczak:
Addition of RTB House's Topics API iframe support

## Type of change

- [ ] Bugfix
- [ ] Feature
- [ ] New bidder adapter   
- [ ] Build related changes
- [ ] CI related changes

- [ ] Other

### DIFF
--- a/modules/topicsFpdModule.js
+++ b/modules/topicsFpdModule.js
@@ -19,6 +19,9 @@ const bidderIframeList = {
   bidders: [{
     bidder: 'pubmatic',
     iframeURL: 'https://ads.pubmatic.com/AdServer/js/topics/topics_frame.html'
+  },{
+    bidder: 'rtbhouse',
+    iframeURL: 'https://topics.creativecdn.com/tapi.html'
   }]
 }
 export const coreStorage = getCoreStorageManager(MODULE_NAME);
--- a/modules/topicsFpdModule.md
+++ b/modules/topicsFpdModule.md
@@ -37,6 +37,10 @@ pbjs.setConfig({
                 iframeURL: 'https://ads.pubmatic.com/AdServer/js/topics/topics_frame.html',
                 expiry: 7 // Configurable expiry days
             },{
+                bidder: 'rtbhouse',
+                iframeURL: 'https://topics.creativecdn.com/tapi.html',
+                expiry: 7 // Configurable expiry days
+            },{
                 bidder: 'rubicon',
                 iframeURL: 'https://rubicon.com:8080/topics/fpd/topic.html', // dummy URL
                 expiry: 7 // Configurable expiry days


modules/topicsFpdModule.js
```diff
@@ -19,6 +19,9 @@ const bidderIframeList = {
   bidders: [{
     bidder: 'pubmatic',
     iframeURL: 'https://ads.pubmatic.com/AdServer/js/topics/topics_frame.html'
+  }, {
+    bidder: 'rtbhouse',
+    iframeURL: 'https://topics.creativecdn.com/tapi.html'
```
patmmccann, line 24: this frame doesnt post to top.

patmmccann (changes requested):

modules/topicsFpdModule.js
patmmccann: take a look at view-source:https://ads.pubmatic.com/AdServer/js/topics/topics_frame.html

piwanczak: Appreciate the early feedback @patmmccann ! 
Apologies for too-early post - did get roll back to WIP, but you were faster :-) 
Will update in the next few days

patmmccann: If you study the latest pr from @jlquaccia , it might make sense to branch off that and avoid the iframe altogether

piwanczak: Thanks for the pointer! 
I could be wrong here, but iframe could make sense as the call would be appropriated to the iframe's domain, rather than to the publisher's domain like in 1-party-context javascript call on pub domain. 
If that's fine with you - I'd prefer sticking with iframe for now

piwanczak: Apologies, marking as a draft again

patmmccann:
re-assigning to @jlquaccia who is much more familiar with navigating these particular test failures. Thanks!

modules/topicsFpdModule.js
```diff
@@ -15,10 +15,13 @@ let LOAD_TOPICS_INITIALISE = false;
 const HAS_DEVICE_ACCESS = hasDeviceAccess();
 
 const bidderIframeList = {
-  maxTopicCaller: 1,
+  maxTopicCaller: 2,
```
piwanczak, line 18: @jlquaccia hopefully this change of default is fine with you

patmmccann (approved):

modules/topicsFpdModule.js
jlquaccia: @piwanczak know this has already been merged, but yea seems alright to me.  Thanks for the heads up!  I took some time off last week, so was just getting around to seeing this PR today.

PR #10110, opened 2023-06-18
Vidazoo Bid Adapter: support for multiple bids on single request

saar120:
## Type of change
- [X ] Feature

## Description of change
Added support for sending multiple banner bids to the server on a single request. 

---
- To see the specific tasks where the Asana app for GitHub is being used, see below:
  - https://app.asana.com/0/0/1204691194600609


Rothalack (approved):
Looks good, working in testing. I was planning to get this out in time for the normal thursday release, but it looks like we released on tuesday this week. So this should be out next week unless there's another release tomorrow.

PR #10117, opened 2023-06-19
GPP: usnat support

dgirardi:
## Type of change

- [x] Feature

## Description of change

- extract CMP client logic into shared library code
- update GPP logic to fetch all availlable section data
- make gppConsent available as an activity parameter. this, unfortunately, broke a lot of tests that were doing their setup outside of a setup block - not all of them could be fixed cleanly (see TODO notes)
- introduce new `gppControl_usnat` module that provides a default interpretation for the `usnat` GPP section (defined in [this doc](https://docs.google.com/document/d/1yPEVx3bBdSkIw-QNkQR5qi1ztmn9zoXk-LaGQB6iw7Q))

## More info

 - the `gppControl_usnat` module is activated together with the `consentManagementGpp` module, for now. (It has no config of its own, but it needs GPP config to exist before doing anything)
 - at least with the CMP that I am using as reference (from [this test page](https://github.com/prebid/Prebid.js/blob/master/integrationExamples/gpt/gpp_us_hello_world.html)), it appears that there is no clean equivalent of the TCF "useractioncomplete" flow. On the first load consent starts out always denied (rather than pending). In that test page, settings are not persisted so there's no good way to test the "consent granted" flow.


libraries/mspa/activityControls.js
```diff
@@ -0,0 +1,70 @@
...
+  return isBasicConsentDenied(cd) ||
+    isSensitiveConsentDenied(cd) ||
+    cd.SensitiveDataProcessing[7] === 1
+}
+
+const CONSENT_RULES = {
+  [ACTIVITY_SYNC_USER]: isConsentDenied,
+  [ACTIVITY_ENRICH_EIDS]: isConsentDenied,
+  [ACTIVITY_ENRICH_UFPD]: isConsentDenied,
+  [ACTIVITY_ENRICH_UFPD]: isTransmitUfpdConsentDenied,
```
jsnellbaker, line 46: Is this supposed to share the previous keyname?  

It looks like the `isConsentDenied` function for the `enrichUfpd` key is getting eaten for the 2nd value when this code is actually running (unless I'm missing something).

patmmccann: Looks like we should just delete line 45, as the function called on line 46 incorporates that logic on line 30

libraries/mspa/activityControls.js
```diff
@@ -0,0 +1,69 @@
...
+// default interpretation for MSPA consent(s):
+// https://docs.google.com/document/d/1yPEVx3bBdSkIw-QNkQR5qi1ztmn9zoXk-LaGQB6iw7Q
+
+function isBasicConsentDenied(cd) {
+  return ['MspaServiceProviderMode', 'Gpc'].some(prop => cd[prop] === 1) ||
+    cd.PersonalDataConsents === 2 ||
+    cd.KnownChildSensitiveDataConsents[0] === 1 || cd.KnownChildSensitiveDataConsents[1] !== 0;
+}
+
+function isSensitiveConsentDenied(cd) {
```
patmmccann, line 19: misleading name, as these checks are for notice.

libraries/mspa/activityControls.js
```diff
@@ -0,0 +1,69 @@
...
+function isConsentDenied(cd) {
+  return isBasicConsentDenied(cd) ||
+    ['SaleOptOut', 'SharingOptOut', 'TargetedAdvertisingOptOut'].some(prop => cd[prop] === 1) ||
+    ['SaleOptOutNotice', 'SharingNotice', 'SharingOptOutNotice', 'TargetedAdvertisingOptOutNotice'].some(prop => cd[prop] === 2);
+}
+
+function isTransmitUfpdConsentDenied(cd) {
+  return isConsentDenied(cd) ||
+    isSensitiveConsentDenied(cd) ||
+    cd.SensitiveDataProcessing.some((val, i) => (i < 6 || i > 7) && val === 1) ||
```
patmmccann, line 32: should be (i < 6 || i > 8), right @bretg ?
image

patmmccann: or should this just be i !== 8?

bretg: Just spoke with Prebid legal counsel and we evolved the logic for transmitUfpd

```
...
SensitiveDataProcessing[1-5,11]=1 OR
SensitiveDataProcessing[6,7,9,10,12]<>0 OR
...
```

dgirardi:
In code the first element has index 0. In the doc it starts from 1 (I think
- the spec enumerates them from 1 at least).

libraries/mspa/activityControls.js
patmmccann: latest commit should reflect, please check @bretg

libraries/mspa/activityControls.js
```diff
@@ -0,0 +1,69 @@
...
+} from '../../src/activities/activities.js';
+import {gppDataHandler} from '../../src/adapterManager.js';
+
+// default interpretation for MSPA consent(s):
+// https://docs.google.com/document/d/1yPEVx3bBdSkIw-QNkQR5qi1ztmn9zoXk-LaGQB6iw7Q
+
+function isBasicConsentDenied(cd) {
+  return ['MspaServiceProviderMode', 'Gpc'].some(prop => cd[prop] === 1) ||
+    cd.PersonalDataConsents === 2 ||
+    cd.KnownChildSensitiveDataConsents[0] === 1 || cd.KnownChildSensitiveDataConsents[1] !== 0;
```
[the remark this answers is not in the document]
patmmccann: confirming i had an old reference

modules/consentManagementGpp.js
```diff
@@ -281,11 +200,10 @@ export const requestBidsHook = timedAuctionHook('gpp', function requestBidsHook(
...
  * If it's good, then we store the value and call `onSuccess`
  */
-function processCmpData(consentObject, {onSuccess, onError}) {
+function processCmpData(consentData, {onSuccess, onError}) {
   function checkData() {
-    const gppString = consentObject && consentObject.gppString;
-    const gppSection = (checkApplicableSectionsIsReady(consentObject)) ? consentObject.applicableSections
-      : (checkApplicableSectionIsReady(consentObject)) ? consentObject.applicableSection : [];
+    const gppString = consentData?.gppData?.gppString;
+    const gppSection = consentData?.gppData?.applicableSections;
```
jsnellbaker, line 206: Should we do any sanity checks here on the results of the new `sectionData` property that was added with this PR?

patmmccann:
Hopefully re-ready for review with all changes in @bretg !

modules/consentManagementGpp.js
jsnellbaker: Update - realized that the `sectionData` field here should always be an object due to the `Object.fromEntries` used to create variable.  So we probably don't need to do a surface check here for this new field.

libraries/mspa/activityControls.js
```diff
@@ -0,0 +1,91 @@
...
+    // targeted opted in but notice does not apply
+    ((cd.TargetedAdvertisingOptOut === 2 && cd.TargetedAdvertisingOptOutNotice === 0)) ||
+    // sharing opted in but notices do not apply
+    ((cd.SharingOptOut === 2 && (cd.SharingNotice === 0 || cd.SharingOptOutNotice === 0)));
+}
+
+export function isTransmitUfpdConsentDenied(cd) {
+  // SensitiveDataProcessing[1-5,11]=1 OR SensitiveDataProcessing[6,7,9,10,12]<>0 OR
+  const mustBeZero = [6, 7, 9, 10, 12];
+  const mustBeZeroSubtractedVector = mustBeZero.map((number) => number - 1);
```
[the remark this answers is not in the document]
patmmccann: sesitive processing is a 12 item vector where each element can have value 0,1, or 2. Items  [6, 7, 9, 10, 12] starting at array index 1 must be zero according to the design doc. I subtract 1 from each of them on this line to get [5,6,8,9,11] and then check in the following line that the javascript array, which starts at zero, subsetted to those positions, are all 0. In this context, 0 means not applicable.

jsnellbaker (approved):
Though I can't test everything here, it seems like it's okay from what I can tell.

libraries/mspa/activityControls.js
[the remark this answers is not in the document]
patmmccann: The unit test for this mustBeZero function is here btw https://github.com/prebid/Prebid.js/blob/67dc9f5044c801bd59cba7e2efc96c603da8a4d2/test/spec/libraries/mspa/activityControls_spec.js#L148

The idea is,when it comes to sensitive biometric data, prebid default is to not append to user first party data unless that consent field is N/A

patmmccann (approved):

patmmccann:
docs https://github.com/prebid/prebid.github.io/pull/4683

PR #10119, opened 2023-06-20
Oxxion Analytics Adapter: fix video bid stringify

matthieularere-msq:
## Type of change
- [x] Bugfix

## Description of change
JSON.stringify on bid response does not work when the object has a renderer attribute because of circular references. As I didn't find a proper way to make a clean dereferenced copy of the object, I've choose to save the render value in a temporary variable, remove it for the bid, stringify the bid and the push back the renderer inside. Not proud of it, but it would have been easier if analytics module could have received already dereferenced object so we can manipulate them without messing out with other parts of prebid.js

### DIFF
--- a/modules/oxxionAnalyticsAdapter.js
+++ b/modules/oxxionAnalyticsAdapter.js
@@ -77,7 +77,8 @@ function cleanAuctionEnd(args) {
 }
 
 function cleanCreatives(args) {
-  return filterAttributes(args, false);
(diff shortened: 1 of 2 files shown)


patmmccann:
didnt you just submit this pr? Do you have two adapters?

matthieularere-msq:
Yes, I've tried to rename it last year but I was told it would be better to create a second one to avoid breaking changes.

patmmccann (approved):

Issue #9981, opened 2023-05-22
Feature Request: add support for fetch endpoints to topics module


patmmccann:
## Type of issue
Feature Request

## Description
Topics from the fetch headers may be more richly populated than topics gathered in the iframe. This feature request is to add support for an endpoint that reflects the topics it saw on a fetch api call headers instead of an injected frame doing a post message. [Docs
](https://developer.chrome.com/docs/privacy-sandbox/topics/#use-headers-to-access-and-observe-topics) [Demo](https://topics-fetch-demo.glitch.me/)

pm-harshad-mane:
The Topics are generated as per the user's browsing history so I am assuming that the generated Topics for a user will not change drastically within a day.  
The Fetch end-point for Topics needs a server-side infrastructure. 
To save the infra cost, we can avoid frequent calls to the end-points by caching the response on the client side in the first-party local storage. 
The Topics module can take care of managing the cache and calling the SSP Fetch endpoints when the cache is stale. 
@patmmccann please share your thoughts.

patmmccann:
> The Topics are generated as per the user's browsing history so I am assuming that the generated Topics for a user will not change drastically within a day.

They change weekly

> the Topics module can take care of managing the cache and calling the SSP Fetch endpoints when the cache is stale.

this seems wise, assuming it is configurable and subject to publisher device access consent, like sharedid module

github-kevin-margueritte:
Hello,

As you may already know, Chrome 115 is now available and comes with the Topics API. I haven't closely followed the implementation of the iframe in Prebid and the limitations that led to the current proposal. However, I see some drawbacks with the solution proposed by Google.

As suggested in this ticket, it would be great to have an additional option on the adapter to use Fetch instead of XHR to send user topics during the bid request (this option could be implemented [here](https://github.com/prebid/Prebid.js/blob/master/src/adapters/bidderFactory.js#L464)). Furthermore, granting the adtech actor full control of topics and allowing them to update the classifier by returning the HTTP header `Browsing-Topics: ?1` seems to be a more flexible and performant method, as recommended by Google. Are there any plans to implement this solution, and is there any way we can help with it?

patmmccann:
@github-kevin-margueritte this comment is a dupe of another issue, #9980, recode ajax to fetch and allow topics to be appended

remysaissy:
[...]

Issue #9982, opened 2023-05-22
Topics Module: Should we maintain the SSP end-points in the module than expecting publishers to pass those?


pm-harshad-mane:
## Type of issue
Feature request / Question
Refer: https://github.com/prebid/Prebid.js/blob/master/modules/topicsFpdModule.md
```
pbjs.setConfig({
    userSync: {
        ...,
        topics: { 
            maxTopicCaller: 3, // SSP rotation 
            bidders: [{
                bidder: 'pubmatic',
                iframeURL: 'https://ads.pubmatic.com/AdServer/js/topics/topics_frame.html',
                expiry: 7 // Configurable expiry days
            },{
                bidder: 'rubicon',
                iframeURL: 'https://rubicon.com:8080/topics/fpd/topic.html', // dummy URL
                expiry: 7 // Configurable expiry days
            },{
                bidder: 'appnexus',
                iframeURL: 'https://appnexus.com:8080/topics/fpd/topic.html', // dummy URL
                expiry: 7 // Configurable expiry days
            }]
        }
        ....
    }
})
```

The Topics module needs to be configured by publishers with endpoints for each SSP. 
This may take longer to set up and could cause extra effort for publishers.
Also, the endpoints will be the same for all publishers, so It would be easier for SSPs to modify these endpoints when needed if they are maintained directly in the Prebid code.

pm-harshad-mane:
yeah, am expecting the end-points should be maintained in Prebid just like the bidders maintain their user-sync end-points.

pm-harshad-mane:
@patmmccann , the Topics are retrieved by using JS API `document.browsingTopics()`, This API makes sure that callers only receive topics they've observed. 
https://developer.chrome.com/docs/privacy-sandbox/topics/overview/#api-callers-only-receive-topics-theyve-observed
Caller's domain plays an important role here, hence we use iframes. 
I am not clear how Fetch API will be useful here, can you help me understand? Is Chrome providing some other mechanism than `document.browsingTopics()`?

patmmccann:
see #9981 and the accompanying docs ref. It seems like calling document.browsingTopics() is not going to be wise soon instead of using headers on a fetch

patmmccann:
@pm-harshad-mane doesnt this line solve your issue? https://github.com/prebid/Prebid.js/blob/8e3c263c3374040e27fb96073ddd35d5f7b7c334/modules/topicsFpdModule.js#L126

Issue #9980, opened 2023-05-22
Proposal: Recode ajax.js to use fetch


patmmccann:
## Type of issue
Feature request

## Description

Utilization of the fetch api for bidders instead of XHR would provide bidder access to Sec-Browsing-Topics headers (provided consent, publisher opt in, etc)

See: https://developer.chrome.com/docs/privacy-sandbox/topics/#use-headers-to-access-and-observe-topics

Are there any disadvantages to transitioning from xhr to fetch? Are there any implications to plans to use a web worker to offload the requests off the main thread? 

## Other information
Related: #3058

dgirardi:
From what I understand the header is filled in by the browser, using different values depending on the origin of the destination.

If that's true, it means that requests through Prebid Server will see different topics compared to requests coming directly from JS. What's our plan to deal with this - if we need to work around the concept (like we are doing now with iframes) then I don't see the value in also adding the header; if we don't work around it, and topics takes off, it will be an incentive to move off Prebid Server.

PR #10124, opened 2023-06-20
Topics Module: Support for Fetch Header Functionality

jlquaccia:
## Type of change
- [x] Feature

## Description of change
- added optional configurable support to integrate with Topics API header/fetch technique to obtain topics
- added configurable support set the duration (in days) for how often fetch calls should be made
- created a test/example server endpoint demonstrating how an ssp topics fetch header endpoint could look like (not sure if we would like to keep this, but I left it in for now)

## Other information
GitHub Issue Ticket: https://github.com/prebid/Prebid.js/issues/9981


jlquaccia:
After code review/feedback I can update the docs for this PR as well.  Also, I had 1 question while working on this ticket.  I noticed that when storing topics in local storage, we are currently only storing the first index of whatever came back in the topics array, is this expected?  I know at times the topics array can have more than 1 item in it.  See the following links for a reference.  Came across that and was curious.

https://github.com/prebid/Prebid.js/blob/d7517de11ec362026c37d7c41b9c63bfb849eef9/modules/topicsFpdModule.js#L164
https://github.com/prebid/Prebid.js/blob/d7517de11ec362026c37d7c41b9c63bfb849eef9/modules/topicsFpdModule.js#L274 (i followed the same pattern with the fetch endpoint here)

patmmccann:
> when storing topics in local storage, we are currently only storing the first index of whatever came back in the topics array, is this expected?

looks like a bug to me, good catch

modules/topicsFpdModule.js
```diff
@@ -251,6 +258,25 @@ function loadTopicsForBidders() {
...
+
+        if (!bidderLsEntry || (bidderLsEntry && isCachedDataExpired(bidderLsEntry[lastUpdated], fetchRate || DEFAULT_FETCH_RATE_IN_DAYS))) {
+          window.fetch(`${fetchUrl}?bidder=${bidder}`, {browsingTopics: true})
+            .then(response => {
+              return response.json();
+            })
+            .then(data => {
+              if (data && data.segment && !isEmpty(data.segment.topics)) {
+                const {domain, topics, bidder} = data.segment;
+                const fetchTopicsData = getTopicsData(domain, topics)[0];
```
patmmccann, line 274: if this is just the first topic, you are correct that it is a bug

jlquaccia: Just pushed a fix for the comment above.

jlquaccia:
Just wanted to follow up on a review for this one.  @patmmccann @dgirardi let me know what you guys think, or if any changes should be made.

modules/topicsFpdModule.js
```diff
@@ -232,17 +238,20 @@ export function hasGDPRConsent() {
...
   const topics = config.getConfig('userSync.topics') || bidderIframeList;
+
   if (topics) {
     listenMessagesFromTopicIframe();
     const randomBidders = getRandomBidders(topics.bidders || [], topics.maxTopicCaller || 1)
-    randomBidders && randomBidders.forEach(({ bidder, iframeURL }) => {
+    randomBidders && randomBidders.forEach(({ bidder, iframeURL, fetchUrl, fetchRate }) => {
       if (bidder && iframeURL) {
         let ifrm = document.createElement('iframe');
         ifrm.name = 'ifrm_'.concat(bidder);
```
remysaissy, line 257: I am not a prebid expert but from Google's documentation, iframe is not longer required when retrieving a topic, (see https://developer.chrome.com/docs/privacy-sandbox/topics/demo/#the-topics-api-headers-demo).

And If it is still required in the prebid context (due to CORS?) would it be possible to:
- create iframes only for bidders that do require it
- send topics to all others

modules/topicsFpdModule.js
```diff
@@ -251,6 +260,25 @@ function loadTopicsForBidders() {
         setLoadedIframeURL(new URL(iframeURL).origin);
         iframeURL && window.document.documentElement.appendChild(ifrm);
       }
+
+      if (bidder && fetchUrl) {
+        let storedSegments = new Map(safeJSONParse(coreStorage.getDataFromLocalStorage(topicStorageName)));
+        const bidderLsEntry = storedSegments.get(bidder);
+
+        if (!bidderLsEntry || (bidderLsEntry && isCachedDataExpired(bidderLsEntry[lastUpdated], fetchRate || DEFAULT_FETCH_RATE_IN_DAYS))) {
+          window.fetch(`${fetchUrl}?bidder=${bidder}`, {browsingTopics: true})
```
remysaissy, line 269: Can we also send topics send topics in the bid request using the header method?

patmmccann: yes but that is blocked by or part of https://github.com/prebid/Prebid.js/issues/9980 ; your PR would be welcomed

modules/topicsFpdModule.js
```diff
@@ -232,17 +238,20 @@ export function hasGDPRConsent() {
...
   if (!HAS_GDPR_CONSENT || !HAS_DEVICE_ACCESS) {
     logInfo('Topics Module : Consent string is required to fetch the topics from third party domains.');
     return;
   }
+
   const topics = config.getConfig('userSync.topics') || bidderIframeList;
+
   if (topics) {
     listenMessagesFromTopicIframe();
     const randomBidders = getRandomBidders(topics.bidders || [], topics.maxTopicCaller || 1)
```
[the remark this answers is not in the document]
jlquaccia: @remysaissy i agree with you that we shouldn't default to 1 (I think that was added in there during an initial PR for this, as the logic around Topics is still being figured out..) Thinking that maybe the default for this module should be off?  aka no publishers?  since this module requires config setup anyway.  Or, maybe if there is some sort of prebid general iframe/fetch url to point to as a default?  not sure how that would be exactly

modules/topicsFpdModule.js
jlquaccia: yes, fetch or iframes could be chosen.  i think ssp's would all need their own iframe/fetch endpoint configurations for the current implementation of this pr.. 

for a scenario where topics are sent back to all others might be something that prebid as a community would have to decide on/figure out

modules/topicsFpdModule.js
dgirardi, line 269: I have some dumb  questions about this - due to how poorly I understand topics.

I read this as "get some topics to add to the bid stream, while letting the remote end know the topics we already have in this origin".

I don't understand why the endpoint - which I believe is meant to be an SSP - needs to know the topics we see on our end, if all they're doing is telling us what we should add to the list of topics sent out later in the auction. What are they doing with them?

I'd understand a bit more if this change was about enabling the topics header on requests sent out by bidders - although even there, how is better than populating ORTB, which we are doing already?

modules/topicsFpdModule.js
github-xavier-bucchiotty: Hi Jason, I don't understand the need for SSPs to retrieve Topics via headers outside of the HTTP call for the bid-request.
If we focus only on this call, this would have minimal to no overhead on publisher pages.
Which use-case would not be covered here ?

modules/topicsFpdModule.js
github-xavier-bucchiotty: It's maybe because the topics one can see on the page depends on the origin of the caller?

All partners in the same page would not necessarily see the same topics for the same auction (what I mean is that, for the same browser, on the same page, at the same moment, the browser will not respond with the same topics API depending on the origin (eTLD+1) of the call.

So, every bidder has some interest to fetch the topics observed by their own origin domain (eTLD+1).

dgirardi: > It's maybe because the topics one can see on the page depends on the origin of the caller?

This may make sense for iframes (if the iframes we are doing make sense - aren't they just a way to circumvent this "in-browser only" paradigm?), but the new call in this PR is just hitting a JSON endpoint, there's no page with its own topics that may be affected by this.

patmmccann: My understanding is the xhr exception will live as long as gam needs it and not any longer

patmmccann: @dgirardi please see https://developer.chrome.com/en/docs/privacy-sandbox/topics/#use-headers-to-access-and-observe-topics and https://developer.chrome.com/docs/privacy-sandbox/topics/demo/#the-topics-api-headers-demo

In order for the browser fetch to include the topics observed over the user lifetime, the fetch must be made in this way. The lifetime topics for the users observable to that endpoint are included, not the topic of the current request. The module assumption is that the endpoint will simply reflect that header back to the user as a json.
image

modules/topicsFpdModule.js
patmmccann: @github-xavier-bucchiotty the purpose of this module is to share the topics with all participants. You are making a non-germane feature request in your comment.

jlquaccia:
@github-xavier-bucchiotty I see what you mean, and also agree that utilizing the bid request to integrate with the topics fetch header functionality sounds like it makes the most sense.

so would this then mean that:

- when bid requests go out, prebid could include the `{browsingTopics: true}` option (sounds like requests would need to be fetch because of the xhr topics header deprecation, as was noted in a comment earlier on this PR)
- the browsingTopics option would trigger the `Sec-Browsing-Topics` request header to be appended to the bid request and sent out
- the ssp can then interpret the value of `Sec-Browsing-Topics` once the bid request is received on their end (if there is even a value present at all) and act on it

Also, from https://developer.chrome.com/docs/privacy-sandbox/topics/demo/#headers, it sounds like the value for `Sec-Browsing-Topics` will list topics observed for the request URL hostname that have been observed for the current user's browser?  (if that sounds correct?) So like stated earlier, these topic values would of course vary by what a specific user has been viewing in their browser

Below is from the doc page, but want to double check if it sounds like i am interpreting it right..

`In browsers that support the API, the fetch() request will include a Sec-Browsing-Topics header that lists topics observed for the request URL hostname: in this example, topics-server.glitch.me.`

`If no topics have been observed for this hostname and this user, the header is included but the value is empty. In other words, the Sec-Browsing-Topics header on a fetch() request only includes topics that have been observed for the current user's browser by a caller whose origin matches the hostname of the request URL.`

patmmccann:
@jlquaccia I would encourage you not to take the teads team comments as feedback on your PR, which solves the issue as described, but instead a different feature request altogether, which should move to a different location on github for its own discussion. The PR appears to solve the issue as described, let's base our reviews on that.

remysaissy:
> 

It was not our intention to make feature requests in the PR as we left a comment for feature request in the related issue as well.
Thanks for pointing it out, I believe we can simply move what is considered feature request in the issue?
Thanks.

modules/topicsFpdModule.js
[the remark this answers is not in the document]
remysaissy: Following @patmmccann's remark on feature request discussion to be in issues, I've followed-up on this on the issue: https://github.com/prebid/Prebid.js/issues/9981#issuecomment-1649241119

modules/topicsFpdModule.js
github-xavier-bucchiotty: Thanks for the feedback Patrick, sorry for the noise

jlquaccia:
Thanks @patmmccann, that makes sense and I will remember that for future PR's as well going forward!

jlquaccia:
Updated the default logic around which bidders get their iframe/fetch url called.  If no bidders are configured then no bidder resource gets called by default (the general `doc.browsingTopics()` call always gets made though).  Also added my changes on top of @dgirardi 's latest change in the topicsFpdModule.js file.

Could I get another code review when everyone gets a chance?

jlquaccia:
Reverted back to previous topics bidder default logic (as we have decided here.. https://github.com/prebid/Prebid.js/issues/9982)

PR #10136, opened 2023-06-23
fledgeForGpt: consolidate publisher configuration

### LINKED ISSUE #10105: FLEDGE/PAAPI module's adUnit level config requirement needs lots of changes on publisher side
## Type of issue
Fledge Module requires Module level config, bidder level config and slot level config too.
Refer: https://github.com/prebid/Prebid.js/blob/master/modules/fledgeForGpt.md 

Module level and bidder level config can be easily applied with less effort.
But publishers will need to make changes for each slot to enable the Fledge/PAAPI on each slot.

Example to enable the Fledge on slot
```
pbjs.addAdUnits({
    code: "my-adunit-div",
    // other config here
    ortb2Imp: {
        ext: {
            ae: 1
        }
    }
});
```

The documentation says, `Absence indicates this is not FLEDGE eligible`. 

My suggestion is to set the default value of `ae` flag to `1` at the slot level internally through the Fledge module, we should not set the flag to `1` if the user has explicitly set it `0`. 

This will avoid the publisher's efforts for modifying all slot-level configs on the setup.
The overall fledge flow will execute only if the module is included and enabled (module and bidder config) by the publisher.

laurb9:
## Type of change
- [x] Feature

## Description of change
Allows enabling fledge for all adunits and specified bidders with a single global config:
```js
pbjs.setConfig({
  fledgeForGpt: {
    enabled: true,
    bidders: ['openx', 'rtbhouse'],
    defaultForSlots: 1
  }
});
```

The previous documented use of `setBidderConfig()` and `addAdUnits() setting imp.ext.ae` continue to be supported, as they offer additional control over the configuration.

- Behavior with `bidders=[]` is the same as missing `bidders`, meaning _all bidders_. If this can cause confusion, we could change it.
- imp.ext.ae is still removed as before, if it was set on adunit but fledge is disabled for bidder.

## Other information
Resolves #10105

### DIFF
--- a/modules/fledgeForGpt.js
+++ b/modules/fledgeForGpt.js
@@ -62,8 +62,8 @@ export function markForFledge(next, bidderRequests) {
       const useGlobalConfig = globalFledgeConfig?.enabled && (bidders.length == 0 || bidders.includes(req.bidderCode));
       Object.assign(req, config.runWithBidder(req.bidderCode, () => {
         return {
-          fledgeEnabled: config.getConfig('fledgeEnabled') ?? (useGlobalConfig && globalFledgeConfig.enabled),
-          defaultForSlots: config.getConfig('defaultForSlots') ?? (useGlobalConfig && globalFledgeConfig?.defaultForSlots)
+          fledgeEnabled: config.getConfig('fledgeEnabled') ?? (useGlobalConfig ? globalFledgeConfig.enabled : undefined),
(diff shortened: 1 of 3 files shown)


JoelPM:
@pm-harshad-mane - any chance you could review this?

spotxslagle (approved):
LGTM

dgirardi (approved):

PR #10141, opened 2023-06-23
A1Media  RTD module : initial release

Sagoru:
## Type of change
- [x] Feature

## Description of change
Add A1Media RTB provider

### DIFF
--- a/modules/.submodules.json
+++ b/modules/.submodules.json
@@ -77,7 +77,8 @@
       "sirdataRtdProvider",
       "timeoutRtdProvider",
       "weboramaRtdProvider", 
-      "zeusPrimeRtdProvider"
+      "zeusPrimeRtdProvider",
+      "a1MediaRtdProvider"
     ],
     "fpdModule": [
       "validationFpdModule",
--- a/modules/.submodules.json
+++ b/modules/.submodules.json
@@ -54,6 +54,7 @@
     ],
     "rtdModule": [
       "1plusXRtdProvider",
+      "a1MediaRtdProvider",
       "aaxBlockmeterRtdProvider",
       "airgridRtdProvider",
       "akamaiDapRtdProvider",
@@ -77,8 +78,7 @@
       "sirdataRtdProvider",
       "timeoutRtdProvider",
       "weboramaRtdProvider", 
-      "zeusPrimeRtdProvider",
-      "a1MediaRtdProvider"
+      "zeusPrimeRtdProvider"
     ],
     "fpdModule": [
       "validationFpdModule",
--- a/modules/a1MediaRtdProvider.js
+++ b/modules/a1MediaRtdProvider.js
@@ -1,0 +1,74 @@
+import { MODULE_TYPE_RTD } from '../src/activities/modules.js';
+import { loadExternalScript } from '../src/adloader.js';
+import { submodule } from '../src/hook.js';
+import { getStorageManager } from '../src/storageManager.js';
+import { isEmptyStr, mergeDeep } from '../src/utils.js';
+
+const REAL_TIME_MODULE = 'realTimeData';
+const MODULE_NAME = 'a1Media';
+const SCRIPT_URL = 'https://linkback.contentsfeed.com/src';
+export const A1_SEG_KEY = '__a1tg';
+export const A1_AUD_KEY = 'a1gid';
+
+export const storage = getStorageManager({moduleType: MODULE_TYPE_RTD, moduleName: MODULE_NAME});
+
+/** @type {RtdSubmodule} */
+export const subModuleObj = {
+  name: MODULE_NAME,
+  init: init,
+  getBidRequestData: alterBidRequests,
+};
+
+export function getStorageData(key) {
+  let storageValue = '';
+  if (storage.getDataFromLocalStorage(key)) {
+    storageValue = storage.getDataFromLocalStorage(key);
+  } else if (storage.getCookie(key)) {
+    storageValue = storage.getCookie(key);
+  }
+  return storageValue;
+}
+
+function loadLbScript(tagname) {
+  const linkback = window.linkback = window.linkback || {};
+  if (!linkback.l) {
+    linkback.l = true;
+
+    const d = new Date();
+    const month = `0${(d.getMonth() + 1)}`.slice(-2);
+    const day = `0${d.getDate()}`.slice(-2);
+    const todaysDate = `${d.getFullYear()}${month}${day}`;
+
+    const scriptUrl = `${SCRIPT_URL}/${todaysDate}/${tagname}.min.js`;
+    loadExternalScript(scriptUrl, MODULE_NAME);
+  }
+}
+
+function init(config, userConsent) {
+  const tagId = config.params.tagId;
+  if (tagId && !isEmptyStr(tagId)) {
+    loadLbScript(config.params.tagId);
+    return true;
+  }
+  if (!isEmptyStr(getStorageData(A1_SEG_KEY))) {
+    return true;
+  }
+  return false;
+}
+
+function alterBidRequests(reqBidsConfigObj, callback, config, userConsent) {
+  const a1seg = getStorageData(A1_SEG_KEY);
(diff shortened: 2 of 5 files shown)


modules/a1MediaRtdProvider.js
```diff
@@ -0,0 +1,74 @@
...
+  const linkback = window.linkback = window.linkback || {};
+  if (!linkback.l) {
+    linkback.l = true;
+
+    const d = new Date();
+    const month = `0${(d.getMonth() + 1)}`.slice(-2);
+    const day = `0${d.getDate()}`.slice(-2);
+    const todaysDate = `${d.getFullYear()}${month}${day}`;
+
+    const scriptUrl = `${SCRIPT_URL}/${todaysDate}/${tagname}.min.js`;
```
patmmccann, line 42: Can you add some discussion of why you need external script loaded to return segments to your documentation? It's there a reason you cannot lock this script to a version instead of this date stuff?

modules/a1MediaRtdProvider.js
```diff
@@ -0,0 +1,74 @@
...
+  }
+  return false;
+}
+
+function alterBidRequests(reqBidsConfigObj, callback, config, userConsent) {
+  const a1seg = getStorageData(A1_SEG_KEY);
+  const a1gid = getStorageData(A1_AUD_KEY);
+  const a1Ortb2 = {
+    user: {
+      buyeruid: a1gid,
```
patmmccann, line 64: This is the wrong place I'm sure; can you add some detail on this id and perhaps we can find the right place.

modules/a1MediaRtdProvider.js
```diff
@@ -0,0 +1,74 @@
...
+      ext: {
+        a1tg: a1seg
+      }
```
patmmccann, line 67: Also appears incorrect; why not use user.data object and register a segtax? Otherwise, use user.ext.data

test/spec/modules/a1MediaRtdProvider_spec.js
```diff
@@ -0,0 +1,94 @@
...
+      });
+      after(function() {
+        clock.restore();
+      });
+
+      it('successfully initialize with load script', function() {
+        expect(subModuleObj.init(configWithParams)).to.be.true;
+        expect(window.linkback.l).to.be.true;
+        expect(loadExternalScript.called).to.be.true;
+        expect(loadExternalScript.args[0][0]).to.deep.equal('https://linkback.contentsfeed.com/src/20230610/lb4test.min.js');
```
patmmccann, line 48: Fails

modules/a1MediaRtdProvider.js
Sagoru: The reason for needing an external script is not just to acquire data from the current page, but also to analyze accumulated data periodically. Instead of using a version, the script is loaded based on the date for a specific reason. The A1Media script has many customizable elements for the deployed site. Therefore, there is an issue where the script needs to be changed immediately upon request from the site. By loading the script based on the date, it ensures that the script is refreshed within a maximum of one day.

modules/a1MediaRtdProvider.js
Sagoru: It is a 1st party id that assigned to users by A1Media for the purpose of analysis and targeting.It is a 1st party user id.

test/spec/modules/a1MediaRtdProvider_spec.js
Sagoru: updated comparing tag url.

modules/a1MediaRtdProvider.js
patmmccann: Please add this to your documentation

modules/.submodules.json
```diff
@@ -77,7 +77,8 @@
       "sirdataRtdProvider",
       "timeoutRtdProvider",
       "weboramaRtdProvider", 
-      "zeusPrimeRtdProvider"
+      "zeusPrimeRtdProvider",
+      "a1MediaRtdProvider"
```
patmmccann, line 81: these are alphabetical

modules/a1MediaRtdProvider.js
patmmccann: Please see the https://github.com/InteractiveAdvertisingBureau/openrtb/blob/master/extensions/2.x_official_extensions/eids.md spec.

patmmccann (changes requested):

modules/a1MediaRtdProvider.js
ChangsikChoi: We added this to markdown file and https://github.com/prebid/prebid.github.io/pull/4672 too

modules/a1MediaRtdProvider.js
ChangsikChoi: We moved from `user.buyeruid` to `user.ext.eids`

modules/a1MediaRtdProvider.js
ChangsikChoi: We moved from `user.ext.a1tg` to `user.data`

modules/.submodules.json
ChangsikChoi: We fixed it to alphabetical

ChangsikChoi:
We fixed for requested changes.
Please tell me if there is something more to improve.

thank you

modules/a1MediaRtdProvider.js
patmmccann: Thanks! Publishers generally prefer non-dynamic scripts so you don't break them with changes. We'll have to review this requested exception in committee.

modules/a1MediaRtdProvider.js
patmmccann: If you're going to use user.data, please reserve a segtax id with a pr to segtax.md on iabtl GitHub and populate it in this object as well

modules/a1MediaRtdProvider.js
bretg: I'll go ahead and be the bad cop here -- this is simply not allowed. The [Prebid Module Rules](https://docs.prebid.org/dev-docs/module-rules.html#module-rules) clearly state:

```
The file loaded is locked to a particular version so that any upgrades must be done via Prebid pull request.
```

This is a rule specifically because many publishers do not want modules to change underneath them without notice. Your adapter is not special, and this exception will not be granted.

Here are some options:
1.  Abide by the rules and put a static version number. 
2. As your publisher partners to place your script directly on their page outside of Prebid.
3. Make the script URL a bidder-specific runtime parameter so when you urgently need to change something, you can contact the affected publisher and have them update the parameter rather than having to update the version of PBJS.
 
Note - we are fully aware that the "static version number" rule can be gamed. If you are found to be dynamically updating a supposedly static URL, there will be repercussions.

bretg: Ok - sorry- I'll admit I thought you guys were a bid adapter at first. An RTD module is more permissive and nuanced. That rule states:
```
A Real-Time Data module may load external code if it requires publisher registration and there’s a prominent disclosure on the module documentation. The idea is that a publisher will not include the module if they don’t approve of the external code, and since they’ve registered for the service, they must approve. The text of the disclosure may differ if the vendor allows Prebid to do regular reviews of a strictly versioned proprietary library.
```
However, I don't understand what benefit you gain by putting the date in the URL. We will discuss it again internally. The proposed documented disclosure message was way too complicated. If we allow RTD modules to break this rule than publishers should not have to understand the date nuance. If anything, such a disclosure should say something simple like:

```
This module loads unverified external code that could change without warning. 
```

modules/a1MediaRtdProvider.js
ChangsikChoi: We registered segtax ( https://github.com/InteractiveAdvertisingBureau/openrtb/pull/136 )
and added segtax id

modules/a1MediaRtdProvider.js
ChangsikChoi: As advised, we changed architecture to load external script to non-dynamic.
It just loads the script using the name passed from configuration.

patmmccann (approved):

PR #10150, opened 2023-06-27
iMDS Bid Adapter: support gpp

ecammit:
## Type of change

- [ ] Bugfix
- [x] Feature
- [ ] New bidder adapter   
- [ ] Build related changes
- [ ] CI related changes

- [ ] Other

## Description of change

GPP support is added to the iMDS bid adapter

### DIFF
--- a/modules/imdsBidAdapter.js
+++ b/modules/imdsBidAdapter.js
@@ -313,7 +313,7 @@ export const spec = {
       queryParams.push('us_privacy=' + encodeURIComponent(uspConsent));
     }
     if (gppConsent) {
-      queryParams.push('gpp=' + encodeURIComponent(gppConsent.gppString) + '&gppsid=' + encodeURIComponent((gppConsent.applicableSections || []).join(',')));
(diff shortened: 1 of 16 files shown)


modules/imdsBidAdapter.js
```diff
@@ -38,21 +39,23 @@ export const spec = {
...
         ua: navigator.userAgent
       },
       imp: []
-    };
+    }, bidderRequest.ortb2 || {});
 
-    const tmax = bidderRequest.timeout;
+    const callbackTimeout = bidderRequest.timeout;
+    const globalTimeout = config.getConfig('bidderTimeout');
+    const tmax = globalTimeout ? Math.min(globalTimeout, callbackTimeout) : callbackTimeout;
```
patmmccann, line 58: this is incorrect, just use the value in the request

ecammit: > this is incorrect, just use the value in the request

Okay, I was wondering how https://github.com/prebid/Prebid.js/pull/8526 somehow disappeared.

PR #10158, opened 2023-06-28
Ogury Bid Adapter: init new user sync iframe

fermeaux:
## Type of change
Feature

## Description of change
Init user sync through iframe


spotxslagle (approved):
Where you changed the adapterVersion in the spec file, I would instead pass in a constant so that you only need to change the version in one location instead of three. But that's just personal preference. Otherwise, lgtm. Approved

fermeaux:
I take into account your feedback for the next release then. Thanks for the review

PR #10162, opened 2023-06-29
VIS.X Bid Adapter : add first-party id

vfedoseev:
## Type of change

- [ ] Bugfix
- [x] Feature
- [ ] New bidder adapter   
- [ ] Build related changes
- [ ] CI related changes

- [ ] Other

## Description of change
We have added the first-party user id for VIS.X bidder.

### DIFF
--- a/modules/visxBidAdapter.js
+++ b/modules/visxBidAdapter.js
@@ -3,6 +3,7 @@ import { registerBidder } from '../src/adapters/bidderFactory.js';
 import { config } from '../src/config.js';
 import { BANNER, VIDEO } from '../src/mediaTypes.js';
 import { INSTREAM as VIDEO_INSTREAM } from '../src/video.js';
+import {getStorageManager} from '../src/storageManager.js';
 const BIDDER_CODE = 'visx';
 const GVLID = 154;
 const BASE_URL = 'https://t.visx.net';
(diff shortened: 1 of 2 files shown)


vfedoseev:
Hi!

I've checked the test report and I see that the failed test is not related to the changes in this PR.

Thank you in advance for checking.

patmmccann:
Why not just use the sharedid?

vfedoseev:
Hi @patmmccann,

Thank you for your patience.

Actually we have some reasons to use our own first-party id. 
The main reason is that the publishers could integrate our inventory not only as HB, but also via the tags. And for the tag integrations we already use our own 'vads' cookie to save the first-party user ids. So, we'd like to use the same cookie in HB in order to match the users between the different type of integrations - HB and tags.

The other reasons that we can think of are the following:
- More control against the issues from our bidder side
- Avoiding mapping of multiple IDs (saves database space)
- Independence from the 3rd party ID format and logic changes
- Easier communication with publishers as they already possess the info on our ID solution and uses of the browser storage
- Full control over compliance with user privacy regulations (GDPR)

I hope that this information the suggested solution for VIS.X bidder in more details. However, if you have any further questions, we'l be happy to discuss it. 

Thank you in advance and have a nice day.

Best regards,
Vladimir

mmoschovas (approved):

patmmccann:
@vfedoseev this is very interesting; I suggest you add similar discussion to your documentation when pubs inevitably ask the same

vfedoseev:
Hi @patmmccann ,
Thank you a lot for the suggestions. We'll definitely consider the options of how to inform the publishers about the fact that our VIS.X bidder is using our own first-party id.

PR #10167, opened 2023-06-29
Core: Fix sua low entropy check for empty objects

pycnvr:
## Type of change

- [x] Bugfix

## Description of change

Device.sua is not generated by default because window.navigator.userAgentData is mistakenly identified as empty.  On Chrome, this object has read-only properties that doesn't work with hasOwnProperty checks.  

https://jsfiddle.net/bwspyja9/

```javascript
let uaData = window.navigator.userAgentData;

function isEmpty(obj) {
   if (!obj) return true;
   for (let x in obj) {
      if (Object.prototype.hasOwnProperty.call(obj, x)) return false;
   }
   return true;
}

function isReallyEmpty(obj) {
   if (!obj) return true;
   for (let x in obj) return false;
   return true;
}

console.log(uaData);
console.log('is empty? ', isEmpty(uaData));
console.log('is really empty? ', isReallyEmpty(uaData));
```

```
 "Running fiddle"
[object NavigatorUAData] {
  brands: [{
  brand: "Not.A/Brand",
  version: "8"
}, {
  brand: "Chromium",
  version: "114"
}, {
  brand: "Google Chrome",
  version: "114"
}],
  getHighEntropyValues: function getHighEntropyValues() { [native code] },
  mobile: false,
  platform: "macOS",
  toJSON: function toJSON() { [native code] }
}
"is empty? ", true
"is really empty? ", false
```

### DIFF
--- a/src/fpd/sua.js
+++ b/src/fpd/sua.js
@@ -15,6 +15,12 @@ export const HIGH_ENTROPY_HINTS = [
   'fullVersionList'
 ]
 
+export const LOW_ENTROPY_HINTS = [
+  'brands',
+  'mobile',
+  'platform'
+]
+
 /**
  * Returns low entropy UA client hints encoded as an ortb2.6 device.sua object; or null if no UA client hints are available.
  */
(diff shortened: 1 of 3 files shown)


dgirardi (approved):

pycnvr:
Thanks @dgirardi!  
@ChrisHuie Is it ok to merge now?

patmmccann (approved):

PR #10169, opened 2023-06-30
Yahoo Bid Adapter: prevent error being thrown if GPP consent module is not enabled

slimkrazy:
## Type of change

- [x] Bugfix
- [ ] Feature
- [ ] New bidder adapter   
- [ ] Build related changes
- [ ] CI related changes

- [ ] Other

## Description of change

This bug was introduced when we started to consume GPP signals. We've been lucky in that not many publishers switched to this bid adapter from the legacy AOL bid adapter. Issue has already been addressed for v8.x, so this fix ensures that any publishers who switch to the yahoossp bid adapter without upgrading to v8 (we are actively encouraging publishers to do this now) will not have this issue.

### DIFF
--- a/modules/yahoosspBidAdapter.js
+++ b/modules/yahoosspBidAdapter.js
@@ -114,11 +114,12 @@ function extractUserSyncUrls(syncOptions, pixels) {
  */
 function updateConsentQueryParams(url, consentData) {
   const parameterMap = {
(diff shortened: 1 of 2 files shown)


ChrisHuie (approved):

kevpark02:
will this get included in the next Prebid 8 release? Right now the yahoosspBidAdapter in Prebid 8 do not have these changes Thank you!

PR #10170, opened 2023-06-30
WeboramaRTDSubmodule: add gdpr verification for user-centric modules

peczenyj:
## Type of change
 [X] Feature

## Description of change

we will start to check for user consent / gdpr on user centric modules (today we have only WAM)


patmmccann:
this is a very old draft, would require rebasing now

PR #10186, opened 2023-07-06
Connatix Bid Adapter : Initial Release

rares-mihai-preda:
## Type of change

- [ ] Bugfix
- [ ] Feature
- [x] New bidder adapter: https://github.com/prebid/prebid.github.io/pull/4713
- [ ] Build related changes
- [ ] CI related changes

- [ ] Other

## Description of change

New PrebidJS Bid Adapter

### DIFF
--- a/modules/connatixBidAdapter.js
+++ b/modules/connatixBidAdapter.js
@@ -1,0 +1,185 @@
+import {
+  registerBidder
+} from '../src/adapters/bidderFactory.js';
+
+import {
+  deepAccess,
+  isFn,
+  logError,
+  isArray,
+  formatQS
+} from '../src/utils.js';
+
+import {
+  BANNER,
+} from '../src/mediaTypes.js';
+
+const BIDDER_CODE = 'connatix';
+const AD_URL = 'https://capi.connatix.com/rtb/hba';
+const DEFAULT_MAX_TTL = '3600';
+const DEFAULT_CURRENCY = 'USD';
+
+/*
+ * Get the bid floor value from the bid object, either using the getFloor function or by accessing the 'params.bidfloor' property.
+ * If the bid floor cannot be determined, return 0 as a fallback value.
+ */
+export function getBidFloor(bid) {
+  if (!isFn(bid.getFloor)) {
+    return deepAccess(bid, 'params.bidfloor', 0);
+  }
+
+  try {
+    const bidFloor = bid.getFloor({
+      currency: DEFAULT_CURRENCY,
+      mediaType: '*',
+      size: '*',
+    });
+    return bidFloor.floor;
+  } catch (err) {
+    logError(err);
+    return 0;
+  }
+}
+
+export const spec = {
+  code: BIDDER_CODE,
+  gvlid: 143,
+  supportedMediaTypes: [BANNER],
+
+  /*
+   * Validate the bid request.
+   * If the request is valid, Connatix is trying to obtain at least one bid.
+   * Otherwise, the request to the Connatix server is not made
+   */
+  isBidRequestValid: (bid = {}) => {
+    const bidId = deepAccess(bid, 'bidId');
+    const mediaTypes = deepAccess(bid, 'mediaTypes', {});
+    const params = deepAccess(bid, 'params', {});
+    const bidder = deepAccess(bid, 'bidder');
+
+    const banner = deepAccess(mediaTypes, BANNER, {});
+
+    const hasBidId = Boolean(bidId);
+    const isValidBidder = (bidder === BIDDER_CODE);
+    const isValidSize = (Boolean(banner.sizes) && isArray(mediaTypes[BANNER].sizes) && mediaTypes[BANNER].sizes.length > 0);
+    const hasSizes = mediaTypes[BANNER] ? isValidSize : false;
+    const hasRequiredBidParams = Boolean(params.placementId);
+
+    const isValid = isValidBidder && hasBidId && hasSizes && hasRequiredBidParams;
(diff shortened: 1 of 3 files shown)


modules/connatixBidAdapter.js
```diff
@@ -0,0 +1,182 @@
...
+   * that includes the Connatix player script and related data.
+   * This HTML string will be used as the ad content.
+   */
+export function wrapAd(lineItems, requestId, playerId, customerId, isInApp) {
+  var settings = {
+    advertising: {
+      standaloneLineItems: lineItems
+    }
+  };
+  var scriptSrc = isInApp ? `//cd.connatix.com/connatix.player.omid.js?cid=${customerId}` : `//cd.connatix.com/connatix.player.js?cid=${customerId}`;
```
Fawke, line 54: Hi @rares-mihai-preda,

Can you please tell us about the script that you are loading here. What's it's supposed to do and why do you need it?

rares-mihai-preda: @Fawke If the source is from mobile applications, we utilize the first script; whereas, if it is from browsers, we employ the second script.

rares-mihai-preda: The current state of isInApp is always false, making the check unnecessary so I removed the ternary operator.

The script includes Connatix's player code and its monetization, meaning that the bid is injected via external API to the Connatix player for rendering purposes.

Fawke: @patmmccann I'm not sure if this violates rule#2 of the [Global Module Rules](https://docs.prebid.org/dev-docs/module-rules.html). (Modules must not load outside code files unless those libraries are approved and open source).

Can you please have a look? Thanks.

rares-mihai-preda: @Fawke @patmmccann Allow me to add additional details here. This script is our ads rendering core. It will be added to the bid creative after the auction ends and it doesn't affect the Prebid.js core/bundle. It will be loaded as part of the creative in the creative HTML string when our creative will be rendered. It will never be added from the bidder's code prior to our creative getting rendered.

modules/connatixBidAdapter.js
```diff
@@ -0,0 +1,180 @@
...
+   * that includes the Connatix player script and related data.
+   * This HTML string will be used as the ad content.
+   */
+export function wrapAd(lineItems, requestId, playerId, customerId) {
+  var settings = {
+    advertising: {
+      standaloneLineItems: lineItems
+    }
+  };
+  var scriptSrc = `//cd.connatix.com/connatix.player.js?cid=${customerId}`;
```
patmmccann, line 53: there are no protocol relative urls allowed in the project

modules/connatixBidAdapter.js
```diff
@@ -0,0 +1,180 @@
...
+    return 0;
+  }
+}
+
+/*
+   * Wrap the provided bid, playerId, customerId, and scriptId in an HTML string
+   * that includes the Connatix player script and related data.
+   * This HTML string will be used as the ad content.
+   */
+export function wrapAd(lineItems, requestId, playerId, customerId) {
```
patmmccann, line 47: why not just define a renderer on the response, what's the advantage of this?

modules/connatixBidAdapter.js
patmmccann: why not just specify a renderer like everyone else? https://github.com/prebid/Prebid.js/blob/bcf080d381d3f650a03e1808e89d5e517600b151/modules/rubiconBidAdapter.js#L833

modules/connatixBidAdapter.js
```diff
@@ -0,0 +1,180 @@
...
+  return `<!DOCTYPE html>
+      <html lang="en">
+      <head>
+          <meta charset="UTF-8">
+          <title></title>
+          <script>!function(n){if(!window.cnx){window.cnx={},window.cnx.cmd=[];var t=n.createElement('iframe');t.src='javascript:false'; t.display='none',t.onload=function(){var n=t.contentWindow.document,c=n.createElement('script');c.src="${scriptSrc}",c.setAttribute('async','1'),c.setAttribute('type','text/javascript'),n.body.appendChild(c)},n.head.appendChild(t)}}(document);</script>
+          <style>html, body {width: 100%; height: 100%; margin: 0;}</style>
+      </head>
+      <body>
+        <script id="${requestId}">(new Image()).src = 'https://capi.connatix.com/tr/si?token=${playerId}&cid=${customerId}';  cnx.cmd.push(function() {    cnx({      playerId: "${playerId}", settings: ${JSON.stringify(settings)} }).render("${requestId}");  });</script>
```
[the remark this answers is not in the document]
patmmccann: this fails our github code scanner 
image

modules/connatixBidAdapter.js
rares-mihai-preda: @patmmccann I updated and removed this part. We no longer needed it.

modules/connatixBidAdapter.js
rares-mihai-preda: @patmmccann I updated and removed this part. We no longer needed it.

modules/connatixBidAdapter.js
rares-mihai-preda: @patmmccann I updated and removed this part. We no longer needed it.

modules/connatixBidAdapter.js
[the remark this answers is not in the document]
rares-mihai-preda: Should be solved now

Fawke (approved):
LGTM

Issue #10147, opened 2023-06-26
Topics module: check support before inserting frame 


patmmccann:
## Type of issue
Bug

## Description
#10144 notes the frame gets inserted even when the API is not available

PR #10201, opened 2023-07-10
topicsFpdModule: replace ad-hoc GDPR logic with `transmitUfpd` acitivity check; check for topics support before loading frames

dgirardi:
## Type of change

- [x] Feature

## Description of change

- do not load `userSync.topics` URLs if the browser does not support topics (closes https://github.com/prebid/Prebid.js/issues/10147)
- replaced GDPR logic with a simple `enrichUfpd` activity check. Note that for GDPR, this means replacing the requirement for consent on all of purposes 1, 2, 3, and 4 with just purpose 4 (after https://github.com/prebid/Prebid.js/pull/10219 is merged)  and "implicit" purpose 1 (without it topics would still be retrieved but not persisted for long enough to be of use).

### DIFF
--- a/modules/topicsFpdModule.js
+++ b/modules/topicsFpdModule.js
@@ -1,18 +1,22 @@
-import {logError, logWarn, mergeDeep, isEmpty, safeJSONParse, logInfo, hasDeviceAccess} from '../src/utils.js';
+import {isEmpty, logError, logWarn, mergeDeep, safeJSONParse} from '../src/utils.js';
 import {getRefererInfo} from '../src/refererDetection.js';
 import {submodule} from '../src/hook.js';
 import {GreedyPromise} from '../src/utils/promise.js';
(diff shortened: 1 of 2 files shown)


patmmccann (approved):

modules/topicsFpdModule.js
```diff
@@ -86,10 +90,14 @@ export function getTopicsData(name, topics, taxonomies = TAXONOMIES) {
     );
 }
 
+function isTopicsSupported(doc = document) {
+  return 'browsingTopics' in doc && doc.featurePolicy.allowsFeature('browsing-topics')
```
robertrmartinez, line 94: is it guaranteed `allowsFeature` will always be there if browsingTopics is? So no need to safe check here?

robertrmartinez (approved):
looks fine, and since no safe check on allowsFeature guessing its fine and is always there

Issue #9786, opened 2023-04-10
Break out EID conversion logic into their respective ID submodules


dgirardi:
## Type of issue
Improvement

## Description

The logic to format a user ID as an EID  object is currently [centralized into a single file](https://github.com/prebid/Prebid.js/blob/master/modules/userId/eids.js). If a pub is using any ID at all, they are pulling that logic for every ID adapter in existence.

We should break that logic out so that each ID submodule can do EID conversion for itself.

PR #10229, opened 2023-07-13
userID: break out EID configuration to individual submodules

dgirardi:
## Type of change

## Description of change

Closes https://github.com/prebid/Prebid.js/issues/9786


patmmccann (approved):

patmmccann:
Nice!

smenzer:
this is great! LGTM

modules/liveIntentIdSystem.js
```diff
@@ -242,6 +242,59 @@ export const liveIntentIdSubmodule = {
     }
 
     return { callback: result };
+  },
+  eids: {
```
3link, line 246: @dgirardi I have a question related to changes in this PR: LiveIntent's user id module can also provide uid2s, but in case the `uid2idSystem` is not part of the Prebid built, it will be impossible to convert them into the eids. 

What should the intended usage pattern be in general? Is it "if you want an id of a certain type to be provided by a 'secondary' provider, you must have the 'main' provider too"? Or is it not the intention and LiveIntent's user id module should have a recipe for converting uid2s into eids?

3link: @peixunzhang FYI

PR #10236, opened 2023-07-17
Criteo Bid Adapter: add onDataDeletionRequest method

dzhang-criteo:
## Type of change

- [ ] Bugfix
- [x] Feature
- [ ] New bidder adapter   
- [ ] Build related changes
- [ ] CI related changes

- [ ] Other

## Description of change

Comply with US Privacy data deletion requests by providing spec.onDataDeletionRequest method.

### DIFF
--- a/modules/criteoBidAdapter.js
+++ b/modules/criteoBidAdapter.js
@@ -10,6 +10,7 @@ import { getRefererInfo } from '../src/refererDetection.js';
 import { hasPurpose1Consent } from '../src/utils/gpdr.js';
 import { Renderer } from '../src/Renderer.js';
 import { OUTSTREAM } from '../src/video.js';
+import { ajax } from '../src/ajax.js';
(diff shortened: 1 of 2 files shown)


Rothalack:
This has been difficult to test. To be sure I understand correctly, you are supposed to be able to trigger this via a CMP right? I've attempted with Quantcast and they don't seem to have a button for it in their CMP pop up. I see they have a way to request through their website. I assume that's not the intended flow with this function. Do you have any suggestion on how to test this? I don't have any feedback on your code, it looks good.

dzhang-criteo:
If we want to properly test this we would have to call the `performDeletion` command on a CMP supporting it but I do not know if there are currently any CMP which support this command.

Rothalack (approved):
I think this should be good then. It hasn't caused any issues with everything else functioning.

PR #10238, opened 2023-07-17
Ozone Bid Adapter : updates 2.9.0 and tid

AskRupert-DM:
## Type of change

- [x] Bugfix
- [x] Feature

## Description of change

- changes required for PB 8 https://github.com/prebid/Prebid.js/issues/8573 - removed transaction IDs
- added support for batching/limiting to 10 impression objects in a single request with a config switch - onpage and also in the adapter. Off by default. In the config use ozone.batchRequests: true
- added our own auction id as required by ozone server & also openrtb spec

https://github.com/prebid/Prebid.js/issues/8573

### DIFF
--- a/modules/ozoneBidAdapter.js
+++ b/modules/ozoneBidAdapter.js
@@ -373,7 +373,6 @@ export const spec = {
         ozoneRequest.id = ozUuid; // Unique ID of the bid request, provided by the exchange. (REQUIRED)
         ozoneRequest.imp = tosendtags.slice(i, i + 10);
         ozoneRequest.ext = extObj;
-        deepSetValue(ozoneRequest, 'source.tid', ozUuid);// RTB 2.5 : tid is Transaction ID that must be common across all participants in this bid request (e.g., potentially multiple exchanges).
         deepSetValue(ozoneRequest, 'user.ext.eids', userExtEids);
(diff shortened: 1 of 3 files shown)


modules/ozoneBidAdapter.js
```diff
@@ -355,13 +365,34 @@ export const spec = {
...
     }
+    let ozUuid = generateUUID();
+    if (this.isBatchRequests()) {
+      logInfo('going to batch the requests');
+      let arrRet = []; // return an array of objects containing data describing max 10 bids
+      for (let i = 0; i < tosendtags.length; i += 10) {
+        ozoneRequest.id = ozUuid; // Unique ID of the bid request, provided by the exchange. (REQUIRED)
+        ozoneRequest.imp = tosendtags.slice(i, i + 10);
+        ozoneRequest.ext = extObj;
+        deepSetValue(ozoneRequest, 'source.tid', ozUuid);// RTB 2.5 : tid is Transaction ID that must be common across all participants in this bid request (e.g., potentially multiple exchanges).
```
patmmccann, line 376: you shouldn't be setting source.tid as any value other than auction id

AskRupert-DM: have removed - thanks - please check latest commit

patmmccann (approved):

PR #10246, opened 2023-07-19
IX bid adapter: add screen resolution in device object

ccorbo:
## Type of change

- [ ] Bugfix
- [x] Feature
- [ ] New bidder adapter   
- [ ] Build related changes
- [ ] CI related changes

- [ ] Other

## Description of change
IX Bid Adapter: Signals device.w / device.h


modules/ixBidAdapter.js
```diff
@@ -2056,4 +2057,19 @@ function getFormatCount(imp) {
   return formatCount;
 }
 
+/**
+ * Adds device.w / device.h info
+ * @param {object} r
+ * @returns object
+ */
+export function addDeviceInfo(r) {
```
patmmccann, line 2065: why not do this in the global ortb2 object for all bidders picking up device?

patmmccann (dismissed):
seems like an opportunity to do once instead of just in one adapter to be repeated many times

modules/ixBidAdapter.js
ccorbo: Thats just going to be a longer process, as I've never touched pbjs core before. Can we merge this in for now and I can work on the core change

ChrisHuie:
@ccorbo I can add this to a utils function and just come back and update here too. There would be several adapters that could utilize this that would then need to be updated.

ChrisHuie (approved):

Issue #10245, opened 2023-07-19
GPP 1.1 CMP Cancels all auctions, do not fail if tcfapi also exists and gdpr applies


patmmccann:
Closely related to #9483 ; however, the issue here is a GPP 1.1 CMP cancels all auctions following the logic here https://github.com/prebid/Prebid.js/blob/0ae65f962d0841e38ef9f0c6eb1193b8dc6cfbeb/modules/consentManagementGpp.js#L203

The feature request is: while #9483 is being developed, if a 1.1 CMP is detected,  do not fail when gdpr applies if the __tcfapi is also available.  Please commit this feature to both legacy and 7.54.x. If #9483 is done, this becomes moot and this logic should be deleted from 8.x (but not legacy). 

Test page and screenshot: https://www.derwesten.de/?pbjs_debug=true

![image](https://github.com/prebid/Prebid.js/assets/1683175/5b090f6d-786c-4a8a-8df0-2994cb30c097)

User is aware they could choose not to configure gpp module when they have a gpp 1.1 cmp, but some users may experience a CMP upgrade without expecting it, and we're hoping they don't suddenly break.

dgirardi:
> if a 1.1 CMP is detected, do not fail when gdpr applies if the __tcfapi is also available

How about not failing if CMP version is not 1.0? I can't make sense of the gdpr condition.

PR #10247, opened 2023-07-19
consentManagementGPP: do not cancel auctions when CMP is not version 1.0

dgirardi:
## Type of change

- [x] Bugfix

## Description of change

Do not cancel auctions when CMP is using version 1.1 - https://github.com/prebid/Prebid.js/issues/10245

Note that I was not able to find a 1.0 implementation complete enough to test that this still works with 1.0 CMPs.


patmmccann (approved):

lcorrigall:
This looks like what we agreed to, warn if the version is greater than 1.0. I additionally approve.

PR #10250, opened 2023-07-19
consentManagementGPP: do not cancel auctions when CMP is not version 1.0 (7.54.x-legacy backport)

dgirardi:
## Type of change

- [x] Bugfix

## Description of change

Do not cancel auctions when CMP is using version 1.1 - https://github.com/prebid/Prebid.js/issues/10245


patmmccann (approved):

patmmccann:
legacy version of #10247

Issue #10249, opened 2023-07-19
Questions on updating from legacy native fields to ortb native fields


jsnellbaker:
## Type of issue
Question

## Description
I'm working on updating the appnexus adapter to no longer rely on the `convertOrtbRequestToProprietaryNative()` by reading the ortb native information from the `bidRequest.nativeOrtbRequest`.  

But while going through the changes and comparing the fields, I noticed some gaps from the legacy native setup to the ortb native setup.  

I was hoping for some feedback about how to address these gaps, in the case that someone might have been using them with their legacy native adUnits and I don't break part of their setup.  See below:

* the privacyLink, privacyIcon, and clickUrl fields that were present in the legacy native setup don't have a home in the ortb native spec and are simply not included in the converted nativeOrtbRequest object.
* in the legacy, the sizes and aspect_ratio fields for images and icons could support an array of sizes or aspect_ratio objects; however in the nativeOrtbRequest only the first value from the sizes or aspect_ratios array is carried forward and the rest of the values are simply dropped.  The native ortb spec implies it should be one set of values for each image asset (from what I understand), so I think this is why it's currently happening.

dgirardi:
> the privacyLink, privacyIcon, and clickUrl fields that were present in the legacy native setup don't have a home in the ortb native spec and are simply not included in the converted nativeOrtbRequest object.

those are response (not request) fields, right? I'm not sure what they'd mean in a request. `privacyLink` and `clickUrl` do have a place in ortb responses; `privacyIcon` doesn't afaik.

https://github.com/prebid/Prebid.js/blob/a318783b5ca750f7d89fec0596a6645a7b0c4a65/src/native.js#L762-L763

>  The native ortb spec implies it should be one set of values for each image asset (from what I understand), so I think this is why it's currently happening.

Correct. Looking at what we do now however, it doesn't quite match your description  IMO:

 - we pick the first `aspect_ratio` to fill out hmin/wmin, but also try to condense all other aspect ratio into something like `['3:2', '16:9']` and shove it in `ext` (this was plucked from what was initially implemented for Prebid Server)
 - we fail if there's more than one size.

It may make sense to accept any number of sizes, pick the first and put the rest under `ext` as well - not sure.

jsnellbaker:
For 1, I meant for the request.  For example, if I had a setup like:
```
mediaTypes: {
       native: {
              title: {
                required: true,
                len: 50
              },
              body: {
                required: true,
              },
              image: {
                required: true,
[...]

PR #10259, opened 2023-07-20
Core: add privacyIcon to not native asset list

jsnellbaker:
## Type of change

- [x] Bugfix

## Description of change
Added `privacyIcon` to the list of `NATIVE_KEYS_THAT_ARE_NOT_ASSETS` list in the `costants.json` file.

Previously when the `privacyIcon` asset was included in a native adUnit using the legacy setup, the `bidRequest.nativeOrtbRequest` created an additional asset that had no proper reference within the native ortb spec.  By adding this field to the list, the asset does not get created in the `nativeOrtbRequest` object.

Found while investigating #10249


musikele:
In main ticket we were also discussing if we could transform this privacyIcon to an icon asset with a special `ext` value. If we go that way, this PR will be wrong.

patmmccann (approved):

PR #10271, opened 2023-07-21
Native: privacyLink is now converted to ortb.privacy 

musikele:
## Type of change

- [X] Bugfix

## Description of change
In legacy native, users could request `privacyLink`. Current prebid ORTB implementation was not considering this value. We decided that, if `privacyLink: { required: true }`, we set `ortb.privacy = 1`. 

## Other information
Discussion: https://github.com/prebid/Prebid.js/issues/10249

### DIFF
--- a/src/native.js
+++ b/src/native.js
@@ -705,7 +705,7 @@ export function legacyPropertiesToOrtbNative(legacyNative) {
         response.jstracker = Array.isArray(value) ? value.join('') : value;
         break;
       case 'privacyLink':
-        response.privacy = 1;
+        response.privacy = value;
         break;
     }
   });
--- a/src/native.js
+++ b/src/native.js
@@ -629,7 +629,7 @@ export function fromOrtbNativeRequest(openRTBRequest) {
       }
     }
     if (openRTBRequest.privacy) {
-      oldNativeObject.privacyLink = { required: true };
+      oldNativeObject.privacyLink = { required: false };
     }
     // video was not supported by old prebid assets
   }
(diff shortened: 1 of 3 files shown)


src/native.js
```diff
@@ -480,6 +480,11 @@ export function toOrtbNativeRequest(legacyNativeAssets) {
       continue;
     }
 
+    if (key === 'privacyLink') {
+      ortb.privacy = 1;
```
dgirardi, line 484: If I say `privacyLink: {required: false}` should this still be set to 1?

dgirardi:
I believe this also needs a change in `legacyPropertiesToOrtbNative`, to address @jsnellbaker's point.

src/native.js
dgirardi: in ORTB the flag signals support and not whether the link is required, so maybe this is OK - but we should set `required: false` in the inverse logic?

musikele: My takes on this

> If I say privacyLink: {required: false} should this still be set to 1? 
I'd say the presence of the attribute means that you'd like to set privacy to 1. 

> in ORTB the flag signals support and not whether the link is required, so maybe this is OK - but we should set required: false in the inverse logic? 
Agree, will fix

src/native.js
```diff
@@ -696,8 +704,11 @@ export function legacyPropertiesToOrtbNative(legacyNative) {
         // in general, native trackers seem to be neglected and/or broken
         response.jstracker = Array.isArray(value) ? value.join('') : value;
         break;
+      case 'privacyLink':
+        response.privacy = 1;
```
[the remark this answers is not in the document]
musikele: Yeah you are definitely right ----

dgirardi (approved):

Issue #10276, opened 2023-07-24
User ID may use stale USP/GPP consent


dgirardi:
## Type of issue
Bug

## Description

The user ID module has some logic to skip UID refresh when TCF consent data has not changed since the last retrieval:

https://github.com/prebid/Prebid.js/blob/c281fc7cb022be04092c4698c2ec5643925bb039/modules/userId/index.js#L829-L830

The logic does not look at GPP or USP - just TCF. Currently, prebid may continue to use user IDs that were retrieved using GPP/USP consent that was later invalidated.

patmmccann:
solution may also fix https://github.com/prebid/Prebid.js/issues/10276

PR #10286, opened 2023-07-27
UserID: check for all consent (not just GDPR) to check if an ID needs refreshing

dgirardi:
## Type of change

- [x] Bugfix

## Description of change

Fixes https://github.com/prebid/Prebid.js/issues/10276

Also, consent hash is now stored with each ID module - meaning that userId will no longer use storage, at all, even if `strictStorageEnforcement` is not set (meaning also, however, that we need a few more bytes of storage for each ID module).


dgirardi:
the reason is to avoid using "core" storage ( = Prebid acting as the first party, and not requiring consent to use device storage). it's not clear if it's OK for us to use storage in this way; some pubs (or maybe some automated tools) expect to see nothing stored at all when consent is declined. With #8661 we introduced a new `strictStorageEnforcement` option to address that problem, but after discussing with @patmmccann we thought it'd be better if we blame a 3rd party to avoid the grey area and the need for extra config.

patmmccann:
Does this also solve #6433

smenzer:
@dgirardi can you confirm if my understanding of the basic logic is correct? instead of storing the consent hash for gdpr once, we now look at any/all consent data we have, hash them, and store them in the submodule's cookie/local storage space -- but only when "storing the id value". we still only pass in gdpr consent data to the id submodules. there are no changes required for the submodules, because the user id module still checks the submodule's consent hash to determine if a refresh is needed (based on any consent system, not just gdpr). the two main changes here are: 1) all types of consent, not just gdpr, are hashed and checked for changes before a refresh and 2) the hash is stored in each submodule's cookie/local storage space (resulting in the same hash stored several times, one per id submodule).

did i miss anything or get anything wrong?

dgirardi:
@smenzer your summary is accurate. There's no change to the submodule API, and this does not attempt to address https://github.com/prebid/Prebid.js/issues/6433

smenzer (approved):
LGTM

Rothalack (approved):
Looks good

PR #10297, opened 2023-08-01
Check if the value is in seconds, convert to milliseconds if necessary

jingyi-gao-ttd:
## Type of change

- [x] Bugfix
- [ ] Feature
- [ ] New bidder adapter   
- [ ] Build related changes
- [ ] CI related changes

- [ ] Other

## Description of change
This PR addresses an issue with the token expiration check that could lead to incorrect results when comparing token expiration times. Previously, the refresh_expires value in some cases was in seconds rather than the expected UNIX timestamp format (milliseconds since the UNIX epoch). As a result, the token expiration check would always return true, causing the token to be considered expired.

### DIFF
--- a/modules/uid2IdSystem_shared.js
+++ b/modules/uid2IdSystem_shared.js
@@ -165,6 +165,12 @@ function refreshTokenAndStore(baseUrl, token, clientId, storageManager, _logInfo
   });
 }
 
+function isExpired(timestamp) {
+  // Check if the value is in second and convert to milliseconds if necessary
+  let expiredAt = timestamp < 2000000000 ?  timestamp * 1000 : timestamp
+  return Date.now() > expiredAt
+}
+
 export function Uid2GetId(config, prebidStorageManager, _logInfo, _logWarn) {
   let suppliedToken = null;
   const preferLocalStorage = (config.storage !== 'cookie');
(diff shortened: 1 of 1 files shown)


modules/uid2IdSystem_shared.js
```diff
@@ -165,6 +165,12 @@ function refreshTokenAndStore(baseUrl, token, clientId, storageManager, _logInfo
   });
 }
 
+function isExpired(timestamp) {
+  // Check if the value is in second and convert to milliseconds if necessary
+  let expiredAt = timestamp < 2000000000 ? timestamp * 1000 : timestamp
```
lionell-pack-ttd, line 170: 2000000000 is Jan 24 1970 - there will definitely be no real tokens with that date, so I guess anything below that is kind of an invalid value. But I'm not sure about treating this as a "maybe in seconds" value - I think I'd prefer to insist it should be in ms (but open to opinions on this!)

patmmccann (approved):

patmmccann:
I was trying to get circleci to run the linter; got this run for some reason https://app.circleci.com/pipelines/github/jingyi-gao-ttd/Prebid.js/4/workflows/66db977f-4c79-48e7-99ed-5907c0778129/jobs/6

patmmccann:
@lionell-pack-ttd could you advise on the status of this?

PR #10312, opened 2023-08-02
Adquery Bid Adapter : fix getId and decode method

adquery:
## Type of change

- [x] Bugfix
- [x] Feature

## Description of change
- Method that creates bidRequests was updated, get userId was changed to  use adqueryIdSystem 
- fixed decode method in adqueryIdSystem
- fixed getId method
- fix auctionId leak: #9781


patmmccann:
please fix conflicts

ChrisHuie (approved):

PR #10321, opened 2023-08-04
appnexus bid adapter - update to work with native ortb

jsnellbaker:
## Type of change

- [x] Feature

## Description of change
This PR updates the logic in the `appnexusBidAdapter.js` file to read/write native data from/to ortb objects.

There are some caveats around the img sizes and aspect_ratios fields when building the request; this is due to multiple sets of sizes/aspect_ratios available with the prebid legacy native setup (see #10249 discussion).  

Another thing to note (which is mainly why this is still in draft), I can't seem to get the targeting keys to generate with the asset# pattern.  Prebid Core still seems to make the legacy targeting keys even though I make an adunit only use the `native.ortb` setup.  If there's any suggestions about what to do to make this work, please let me know.  I had assumed writing the data to the `bidResponse.native.ortb` would have been sufficient.

### DIFF
--- a/modules/appnexusBidAdapter.js
+++ b/modules/appnexusBidAdapter.js
@@ -33,7 +33,6 @@ import {INSTREAM, OUTSTREAM} from '../src/video.js';
 import {getStorageManager} from '../src/storageManager.js';
 import {bidderSettings} from '../src/bidderSettings.js';
 import {hasPurpose1Consent} from '../src/utils/gpdr.js';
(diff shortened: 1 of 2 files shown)


patmmccann:
@jsnellbaker I'm confused? If you're requesting review can you remove the draft status and re-assign a reviewer? Removed for now due to being in draft.

patmmccann:
@mkomorski removed the legacy keys in #13255 if that unblocks this

PR #10328, opened 2023-08-08
PubMatic Analytics Adapter : log floor values only when floor file is fetched successfully

kapil-tuptewar:
## Type of change

- [ ] Bugfix
- [x] Feature
- [ ] New bidder adapter   
- [ ] Build related changes
- [ ] CI related changes

- [ ] Other

## Description of change
With this change adapter will start logging floor field **frv** for server side partner along with client side partners. If floor type is dynamic then log floors values only if fetchstatus is success i.e. floor file gets download successfully.


jlquaccia (approved):
LGTM

PR #10330, opened 2023-08-08
gptPreAuction: fix bug where adServer object are not set in case of twin ad unit

katsuo5:
## Type of change
- [x] Bugfix

## Description of change
Fix bug where adServer object are not set in case of twin ad unit.

### DIFF
--- a/modules/gptPreAuction.js
+++ b/modules/gptPreAuction.js
@@ -15,7 +15,8 @@ export const appendGptSlots = adUnits => {
   }
 
   const adUnitMap = adUnits.reduce((acc, adUnit) => {
-    acc[adUnit.code] = adUnit;
+    acc[adUnit.code] = acc[adUnit.code] || [];
+    acc[adUnit.code].push(adUnit);
     return acc;
   }, {});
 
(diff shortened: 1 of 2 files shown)


modules/gptPreAuction.js
```diff
@@ -25,15 +26,16 @@ export const appendGptSlots = adUnits => {
...
+      adUnitMap[matchingAdUnitCode].forEach((adUnit) => {
+        adUnit.ortb2Imp = adUnit.ortb2Imp || {};
+        adUnit.ortb2Imp.ext = adUnit.ortb2Imp.ext || {};
+        adUnit.ortb2Imp.ext.data = adUnit.ortb2Imp.ext.data || {};
+
+        const context = adUnit.ortb2Imp.ext.data;
+        context.adserver = context.adserver || {};
+        context.adserver.name = 'gam';
+        context.adserver.adslot = sanitizeSlotPath(slot.getAdUnitPath());
+      });
```
dgirardi, line 38:
```suggestion
    if (matchingAdUnitCode) {
      const adserver = {
        name: 'gam',
        adslot: sanitizeSlotPath(slot.getAdUnitPath())
      };
      adUnitMap[matchingAdUnitCode].forEach((adUnit) => {
        deepSetValue(adUnit, 'ortb2Imp.ext.data.adserver', adserver);
      });
    }
```

dgirardi (approved):
LGTM. The code you moved around is quite clunky IMO, I have a suggestion to improve it.

Unfortunately I can't figure out how to add `import {deepSetValue} from '../src/utils.js';` to the GH suggestion.

modules/gptPreAuction.js
katsuo5: @dgirardi 
Thanks for review and suggestion.

I committed your GH suggestion.

dgirardi (approved):

ChrisHuie (approved):

PR #10334, opened 2023-08-09
LiveIntent UserId module: Add support for index id

3link:
## Type of change
- [x] Other

## Description of change
This pull request adds eid mapping for the index id that now can be resolved via the LiveIntent userId module.

## Other information
- repository of the backing live-connect module: https://github.com/LiveIntent/live-connect
- docs PR: https://github.com/prebid/prebid.github.io/pull/4792

### DIFF
--- a/modules/liveIntentIdSystem.js
+++ b/modules/liveIntentIdSystem.js
@@ -205,6 +205,10 @@ export const liveIntentIdSubmodule = {
         result.magnite = { 'id': value.magnite, ext: { provider: LI_PROVIDER_DOMAIN } }
       }
 
+      if (value.index) {
+        result.index = { 'id': value.index, ext: { provider: LI_PROVIDER_DOMAIN } }
+      }
+
       return result
     }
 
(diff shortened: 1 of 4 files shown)


3link:
Is the `eids_spec.js` broken?

robertrmartinez:
@3link 

Hi!

I am the lead Software Engineer at Magnite for our Prebid integrations.

I was not sure the best way to reach out but had a question for you.

When we upgrade to latest prebid version we see the AJAX call to `rp.liadm.com` timeout every single time.

Seems the `live-connect` NPM library is calling it with timeout === 0 every single time.

And setting `ajaxTimeout` in the config params seems to not change this.

This is happening on all of our pubs who have liveIntent implemented in their prebid wrappers.

image

Wondering how this timeout is configured as it seems to be always set to `0`?

3link:
@robertrmartinez Thank you for reaching out. [This default value ](https://github.com/LiveIntent/live-connect/blob/master/src/pixel/sender.ts#L62) seems to be the problem. We will submit a PR with a fix soon. 

Sorry for the inconvenience. The good news is the error you see does not have an effect on the user id resolution.

robertrmartinez:
Glad to help @3link 

Seems like this will be a PR into the NPM library which will require a Prebid PR in prebids package.json to bump the version as well correct?

3link:
@robertrmartinez 
> Seems like this will be a PR into the NPM library which will require a Prebid PR in prebids package.json to bump the version as well correct?

Correct!

PR #10340, opened 2023-08-09
Core: enable `Sec-Browsing-Topics` header on outgoing bidder requests

dgirardi:
## Type of change

- [x] Feature

## Description of change

- refactor `ajax.js` to use fetch instead of XHR
- automatically pass `browsingTopics: true` to fetch, for all bidder requests, when the `transmitUfpd` activity is allowed

## Other information

Closes https://github.com/prebid/Prebid.js/issues/9980


patmmccann (approved):

lksharma (approved):

mikemangas:
hello, is it still necessary, to fill out the form https://goo.gle/privacy-sandbox-enroll-form?
or was it meant only for the test-phase?

PR #10341, opened 2023-08-09
liveIntent Id System: fix ix eid mismatch with ix adapter

patmmccann:
https://github.com/prebid/Prebid.js/commit/b973b2d134e486c74667dcfd38e432c17440c39f looks for liveintent.indexexchange.com


patmmccann:
@DeepthiNeeladri @3link fyi

lksharma:
@patmmccann I pushed some changes to update the test cases as well.

PR #10351, opened 2023-08-11
GPP MSPA enforcement activity controls: invalidate covered = 0 returns from the CMP

patmmccann:
as per https://github.com/IABTechLab/iabgpp-java/issues/29 many CMP are returning invalid values for covered.


patmmccann:
@bretg it seems lots of people are seeing invalid covered values in the wild. While we aren't range checking every value, this check seems useful.

patmmccann:
closing in favor of https://github.com/prebid/Prebid.js/pull/10354

PR #10355, opened 2023-08-14
identityLinkSubmodule: add additional check on retrieving the envelope

mamatic:
## Type of change

- [ X ] Bugfix

## Description of change
- In case when ats library is loaded on a page after prebid, ats will be undefined and ats.retrieveEnvelope function will not get envelope and pass it to callback, but envelope may be present on a page. With this change we will cover the case when envelope is generated but ats is not present on a page, but we still can return envelope.


smenzer (approved):
LGTM

PR #10366, opened 2023-08-16
IX Bid Adapter: fix: consolidate banner format array

ccorbo:
## Type of change

- [x] Bugfix
- [ ] Feature
- [ ] New bidder adapter   
- [ ] Build related changes
- [ ] CI related changes

- [ ] Other

## Description of change
This PR fixes an issue where in a scenario where the adUnit is configured with multiple bidder configs, with each bid config having a siteID set to a different size. This would result in a multi-imp request. The request should be a single imp, with the different sizes sent out as a multi-format request.


lksharma (approved):
LGTM

PR #10375, opened 2023-08-21
Illumin Bid Adapter: initial release

saar120:
## Type of change

- [X] New bidder adapter  
## Description of change

Illumin Bid Adapter
maintainer email: [integrations@illumin.com](mailto:integrations@illumin.com)
```
{
        bidder: ‘illumin’,
        params: {
          cId: ‘562524b21b1c1f08117fc7f9’,
          pId: ‘59ac17c192832d0011283fe3’,
          bidFloor: 0.0001,
          ext: {
            param1: ‘loremipsum’,
            param2: ‘dolorsitamet’
          }
     }
 }
```

## Other information

https://github.com/prebid/prebid.github.io/pull/4813


saar120:
@mmoschovas @ChrisHuie 
Is there anything we can do to merge this one?

modules/illuminBidAdapter.js
```diff
@@ -0,0 +1,336 @@
...
+    return '';
+  }
+}
+
+export function createDomain(subDomain = DEFAULT_SUB_DOMAIN) {
+  return `https://${subDomain}.illumin.com`;
+}
+
+export function extractCID(params) {
+  return params.cId || params.CID || params.cID || params.CId || params.cid || params.ciD || params.Cid || params.CiD;
```
mmoschovas, line 29: Question regarding these three extract methods. Is there a reason we are checking multiple combinations when the params are spec'd out to be `cId`, `pId` and `subDomain`? This seems unnecessary to me and throwing a warning during validation seems appropriate here if the property name is not correct

saar120: @mmoschovas We are checking this for backward compatibility and minor typos of publishers.

mmoschovas: Maybe I am missing something. Is this not a new adapter? Im not sure where backwards compatibility is coming into play. I also understand the intent, I just do not understand why you are doing it considering it seems these are random alternative values and not comprehensive. Either way, I believe you should allow publishers who do not follow the specification to receive the error and not attempt to handle unknowns

saar120: @mmoschovas 
I've changed the extract method to support only the given properties.
Can we merge this?

mmoschovas (approved):

PR #10377, opened 2023-08-21
dynamicAdBoostModule: New module

adxpremium:
## Type of change

- [ ] Bugfix
- [x ] Feature
- [ ] New bidder adapter   
- [ ] Build related changes
- [ ] CI related changes

- [ ] Other

## Description of change

This new module requires to be whitelisted in the adloader file in order to load external JS script for it's core functionality.


adxpremium:
Thank you for reaching out @robertrmartinez and raising some valid questions about the proposed module for Prebid integration. I'd like to provide a comprehensive explanation of the module's functionality and its potential benefits. 

The module does indeed involve tag loading, however, it extends beyond that basic function. Its core purpose is to enhance ad unit viewability data retrieval, offering publishers valuable insights to optimize their revenue streams. While it does facilitate the loading of ad tags, it also incorporates a data collection mechanism that monitors how ads are being viewed by users. 
This data aims to provide a more accurate understanding of ad performance beyond mere tag loading. Please note that we have indicated a mandatory collaboration with LuponMedia in the description of our module: “In order to utilize this module, it is essential to collaborate with LuponMedia to create an account and obtain detailed guidelines on configuring your sites.”

By offering detailed viewability data, it equips publishers with the means to fine-tune their ad placements and maximize revenue potential. These insights can contribute to a more strategic approach to ad delivery.

We acknowledge your concern about modules that solely focus on tag loading without contributing to bid stream optimization or targeting. The proposed module goes beyond this by providing meaningful data that publishers can leverage and monetize. 

Considering that the loaded script already utilizes viewability data provided by the module, perhaps we could pass the viewablity data (score) to the bid stream directly in order to contribute to greater transparency and compliance with Prebid standards?

And you're absolutely right about the existing viewability partners within Prebid, such as Real Time Data Providers like Browsi etc. These partners play a vital role in supplying real-time viewability data for bidders to consider. However, our module aims to complement these efforts by offering publishers an additional layer of data that is directly monetized on their end.

Thank you very much, we appreciate your intention to consult with the Prebid Committee regarding this module's fit within the ecosystem. We value your feedback and questions. If you require any additional information or clarification, please don't hesitate to reach out.

adxpremium:
> 

Hi @robertrmartinez, we have implemented the discussed changes and have also renamed the module in accordance with the rules. As part of these modifications, our module now successfully sends viewability data in bid requests. We would like you to review and confirm if these adjustments meet Prebid.js requirements.

Thank you.

robertrmartinez:
Some tests failed in the `intersectionRtdProvider` module, which you made no change to but seems connected since you utilize the `IntersectionObserver`

I reran them to see if it was flakey. But can you please try to figure out why and fix accordingly.

You may have to update your unit tests to mock something if it is overwriting any shared objects and not resetting them.

Also, a few more unit tests to cover other cases does not hurt! We aim to be at 80% test coverage or more.

adxpremium:
> Some tests failed in the `intersectionRtdProvider` module, which you made no change to but seems connected since you utilize the `IntersectionObserver`
> 
> I reran them to see if it was flakey. But can you please try to figure out why and fix accordingly.
> 
> You may have to update your unit tests to mock something if it is overwriting any shared objects and not resetting them.
> 
> Also, a few more unit tests to cover other cases does not hurt! We aim to be at 80% test coverage or more.


Thank you for the update. I have added IntersectionObserver support check, an unobserve() function to stop watching when done, and a disconnect() at the end to clear it out. When I run locally, all tests are passing with no issues. I understand that unit tests have to cover everything, but since this module is simple, I am wondering what else we should cover. Do you have any suggestions? Thanks.

robertrmartinez:
Seems you still are causing downstream tests to fail.

You may want to update your tests to utilize a `before` and `after` functions to mock the observer.

Also, I think it may be beneficial to not initialize the observer until the init call (and if `CLIENT_SUPPORTS_IO` is true probably)

Also, you should add some unit tests which cover the `getBidBidRequests` functions code paths

adxpremium:
> Restarting tests

Hi @robertrmartinez, could you please verify if the tests have passed successfully and confirm if our code is ready for merging? If we need to make any changes, please let us know.
Thank you in advance

ChrisHuie:
@adxpremium can you pull in recent commits on master and that will probably fix the integration errors.

adxpremium:
> @adxpremium can you pull in recent commits on master and that will probably fix the integration errors.

Hi @ChrisHuie, we did that, could you please check now? Is there anything we could do on our side to speed up the module approval? Thanks :)

modules/dynamicAdBoostRtdProvider.js
```diff
@@ -0,0 +1,112 @@
...
+const SCRIPT_URL = 'https://adxbid.info';
+const CLIENT_SUPPORTS_IO = window.IntersectionObserver && window.IntersectionObserverEntry && window.IntersectionObserverEntry.prototype &&
+    'intersectionRatio' in window.IntersectionObserverEntry.prototype;
+// Options for the Intersection Observer
+const dabOptions = {
+  threshold: 0.5 // Trigger callback when 50% of the element is visible
+};
+
+// Create an Intersection Observer instance
+const observer = new IntersectionObserver(dabHandleIntersection, dabOptions);
```
robertrmartinez, line 22: You should not be trying to run this if `CLIENT_SUPPORTS_IO` is false I believe.

Am I missing something? It is possible this is what is causing downstream tests to fail perhaps...

Please try to update this to only init the instance if it is supported.

adxpremium: > You should not be trying to run this if `CLIENT_SUPPORTS_IO` is false I believe.
> 
> Am I missing something? It is possible this is what is causing downstream tests to fail perhaps...
> 
> Please try to update this to only init the instance if it is supported.

Thank you @robertrmartinez, we have have made those changes. Could you check now? Thank you

modules/dynamicAdBoostRtdProvider.js
```diff
@@ -0,0 +1,112 @@
...
+
+const MODULE_NAME = 'dynamicAdBoost';
+const SCRIPT_URL = 'https://adxbid.info';
+const CLIENT_SUPPORTS_IO = window.IntersectionObserver && window.IntersectionObserverEntry && window.IntersectionObserverEntry.prototype &&
+    'intersectionRatio' in window.IntersectionObserverEntry.prototype;
+// Options for the Intersection Observer
+const dabOptions = {
+  threshold: 0.5 // Trigger callback when 50% of the element is visible
+};
+var observer;
```
robertrmartinez, line 20: I think just using `let` is better here.

When this change was pushed to make it a var it made a lot of downstream tests fail.

Can you try changing to `let` and see what happens.

robertrmartinez (changes requested):

test/spec/modules/dynamicAdBoostRtdProvider_spec.js
```diff
@@ -0,0 +1,69 @@
...
+        expect(rtdProvider.init(configWithoutRequiredParams)).to.be.false;
+      })
+    })
+
+    describe('initialize with expected params', function() {
+      it('successfully initialize with load script', function() {
+        expect(rtdProvider.init(configWithParams)).to.be.true;
+        setTimeout(function() {
+          expect(loadExternalScript.called).to.be.true;
+        }, 5000);
```
robertrmartinez, line 32: We should not be using setTimeout in unit tests. This test is now effectively taking 5 seconds and may interrupt other tests and cause some weirdness.

An easier way to test this part would be to use sinon's `useFakeTimers`

```js
describe('dynamicAdBoost', function() {
  let clock;
  let sandbox;
  beforeEach(function () {
    sandbox = sinon.sandbox.create();
    clock = sandbox.useFakeTimers(Date.now());
  });
  afterEach(function () {
    sandbox.restore();
  });
```

The only problem with this is that you initialize a date time which is used to check your elapsed time is enough at module global scope, and sinon / jest do not mock timers until you tests begin to run!

So here is my proposal which I think will fix the flakiness of these tests:

1. Update your `dabStartDate` and `dabStartTime` variables to be initialized at the time of `init`
2. Use sinon's fake timers to run your tests and tick the clock instead of actually making the code wait.

Here is what I would do:

```js
// module scope declare your `dabStartDate` and `dabStartTime` vars
let dabStartDate;
let dabStartTime;

// in the init function set these to current date
function init(config, userConsent) {
   let dabStartDate = new Date();
   let dabStartTime = dabStartDate.getTime();
```


And in your test with the above mentioned initialization of fakeTimer

```
    describe('initialize with expected params', function() {
      it('successfully initialize with load script', function() {
        expect(rtdProvider.init(configWithParams)).to.be.true;
        clock.tick(1000); // so your setInterval timer is hit
        expect(loadExternalScript.called).to.be.true;
      })
    });
```

Let me know if that all makes sense.

robertrmartinez (changes requested):

robertrmartinez:
FWIW When I implement these requested changes on my local I no longer am getting errors, where I would intermittently get them with the current logic.

adxpremium:
> FWIW When I implement these requested changes on my local I no longer am getting errors, where I would intermittently get them with the current logic.
> 
> So I am hopeful this change will fix the unit tests from not working.

Thank you @robertrmartinez , you were right, fake timer did the trick. All checks have passed

adxpremium:
Hello @robertrmartinez, now that the module has successfully passed all the checks, could you please let us know when we can expect the merge with the master branch and the subsequent release? Thank you in advance

adxpremium:
Hello @robertrmartinez @ChrisHuie do you need anything extra from us? We would greatly appreciate it if you could speed up the process of further testing or publishing the module. Thanks in advance.

robertrmartinez (approved):

Issue #10349, opened 2023-08-10
Review and improve bid timeout logic


dgirardi:
## Type of issue
Bug

## Description

See https://github.com/prebid/Prebid.js/issues/10320#issuecomment-1673513610 for details - breaking this out into its own issue:

There are several problems with the timeout logic:

- the 2nd mechanism (using timeToRespond) appears to be entirely redundant. I suspect it's a relic.
- the 1st mechanism does not look at timeoutBuffer, which is different from what [we say in the docs](https://docs.prebid.org/dev-docs/publisher-api-reference/setConfig.html#setConfig-timeoutBuffer).
- a bid request can be considered "late" when it isn't, if it's routed to Prebid Server and there's no client side adapter for it.

PR #10379, opened 2023-08-21
Core: improve and fix bid timeout logic

dgirardi:
## Type of change

- [x] Bugfix

## Description of change

- Fix an edge case where S2S bids can be incorrectly marked as timed out
- Simplify timeout logic: currently there are two separate timeout mechanisms, one always firing after `bidderTimeout` and the other _sometimes_ firing after `bidderTimeout + timeoutBuffer`. Since the second, when it happens, always happens too late, it was removed as redundant.

## Other information

Closes https://github.com/prebid/Prebid.js/issues/10349


ChrisHuie (approved):

src/adapterManager.js
```diff
@@ -380,7 +380,7 @@ adapterManager.callBids = (adUnits, bidRequests, addBidResponse, doneCb, request
     return;
   }
 
-  let [clientBidRequests, serverBidRequests] = bidRequests.reduce((partitions, bidRequest) => {
+  let [clientBidderReqs, serverBidRequests] = bidRequests.reduce((partitions, bidRequest) => {
```
spotxslagle, line 383: Is there a reason we're changing this variable? I find the new name less readable. I'd prefer clientBidderRequests, personally.

src/auction.js
```diff
@@ -182,25 +181,21 @@ export function newAuction({adUnits, adUnitCodes, callback, cbTimeout, labels, a
...
-    if (cleartimer) {
-      clearTimeout(_timer);
+  function executeCallback(timedOut) {
+    if (!timedOut) {
+      clearTimeout(_timeoutTimer);
     }
 
     if (_auctionEnd === undefined) {
-      let timedOutBidders = [];
+      let timedOutReqs = [];
```
spotxslagle, line 193: timedOutRequests

spotxslagle (approved):
I commented a few personal preferences, but otherwise LGTM

PR #10387, opened 2023-08-23
Cadent Aperture MX Bid Adapter: rename files to match bidder code

EMXDigital:
## Type of change

- [ ] Bugfix
- [ ] Feature
- [ ] New bidder adapter   
- [x] Build related changes
- [ ] CI related changes

- [ ] Other

## Description of change

File names originally were camelCase but our `bidder_code` is snake_case.  This caused the online build to not build correctly due to the change in naming.  This PR updates the filenames to match the snake_case of the bidder code.


EMXDigital:
> The snake case adapter was new in 8.x; if we merge this it would break any build scripts including it. Are there any publishers using this adapter yet or are you saying you need this change to get it working?
> 
> Is this to fix builds from the download page or from scripts? If the former, I think @ChrisHuie has another way to alias download page names to file names

@patmmccann  Right now, we're getting reports from pubs trying to download a new version of PBJS with our adapter stating that the file generated is an error.    Our old adapter's Bidder Code (`emx_digital`) was snake case, and the file name was snake case.  When we updated our bidder code to the new name (`cadent_aperture_mx`), the filename that we used was NOT snake-cased, and we only tested using a local build, specifying the actual file name WITHOUT the snake_case, and that build worked correctly.  As far as we know, we need to change our file names to be snake_case, to match our bidder_code.  If you say that this is not necessary, and it can work a different way, we're open to that, but we'd need guidance on how to get that to work.  Right now, users are unable to create a build of PBJS from the Prebid website if it includes our adapter.

patmmccann:
We don't want to break anyone that has a build script like this https://github.com/prebid/Prebid.js/blob/4b00f6119bc4b9f639c57c02edc69003736ff662/README.md?plain=1#L173

@ChrisHuie knows of a fix that will fix the download page

PR #10391, opened 2023-08-24
AIDEM Bidder Adapter: ortbConverter

darkstarac:
## Type of change
- [x ] Feature
- 
## Description of change
Moved aidemBidAdapter implementation to comply with ortbConverter


dgirardi:
FYI, video tests are failing because the pipeline runs them once with all features disabled, and the conversion library skips the video logic. To fix it, disable video-specific tests like this:

https://github.com/prebid/Prebid.js/blob/3c70ac772f1b8620015089de778d331619d3b459/test/spec/modules/prebidServerBidAdapter_spec.js#L830

you can test locally with `gulp test-all-features-disabled`.

patmmccann (approved):

Issue #10139, opened 2023-06-23
Fledge for gpt module: add contextual win price to configs


patmmccann:
## Type of issue
Feature request

Auction config submitted by prebid should contain the price, or the floor from the floors module if all bids are excluded as below the floor or there are no bids.

patmmccann:
At https://developer.chrome.com/docs/privacy-sandbox/fledge-api/ad-auction/ ,   `auctionSignals: {...}`, is mentioned. Prebid will assume it will not conflict anyone if it writes to  `auctionSignals.pbjs`

PR #10393, opened 2023-08-24
fledgeForGpt: provide `bidfloor` in auction signals

dgirardi:
## Type of change

- [x] Feature

## Description of change

This adds two auction signals to fledge: `prebid.bidfloor` and `prebid.bidfloorcur`.

The `bidfloor` value is either the highest CPM that Prebid got for the slot, or (if there were no bids) the minimum required floor across all requests for the slot.

## Other information

Closes https://github.com/prebid/Prebid.js/issues/10139


laurb9:
I did a small "live" test -  made a bundle with fledgeForGpt and openxBidAdapter, faked an openx server response to inject some auctionSignals and intercepted `runAdAuction` in the page to print out the arguments. I can see the expected merged `auctionSignals` in the prebid debug log as well as in the `runAdAuction` configuration passed by GPT:

Screenshot 2023-09-05 at 6 23 24 PM

Fawke (approved):
LGTM

patmmccann (approved):

patmmccann (approved):
Taking @laurb9 's review as second, thanks!

pm-nitin-nimbalkar:
@patmmccann @Fawke @laurb9 

Hi Guys, Am just confused with the sentence written in description section "or (if there were no bids) the minimum required floor across all requests for the slot."

If there is no contextual bids for particular fledge slot, my assumption is that bid floor value should be 0. So what is "the minimum required floor across all requests for the slot." How this is calculated. Bit confused on this.

Can you elaborate it more on this with examples ?

TIA,
Nitin

pm-harshad-mane:
@pm-nitin-nimbalkar , the publisher sets a floor value that is passed to all bidders in the contextual-request and bidders are required to bid higher than the floor value. if partners do not return contextual bids then in this case the floor value set by the publisher will be passed in PA.

PR #10398, opened 2023-08-28
Parrable Id System: return empty EID

victorigualada:
## Type of change
- [ ] Feature

## Description of change

Parrable is temporarily pausing it's activity for a few months, so this change basically mocks the adapter to always return an empty EID to the client. Once the activity is restarted, a new PR will be made to revert this change.

### DIFF
--- a/modules/parrableIdSystem.js
+++ b/modules/parrableIdSystem.js
@@ -332,7 +332,9 @@ export const parrableIdSubmodule = {
    */
   getId(config, gdprConsentData, currentStoredId) {
     const configParams = (config && config.params) || {};
-    return fetchId(configParams, gdprConsentData);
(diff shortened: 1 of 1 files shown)


robertrmartinez:
Interesting use case so asking for some input from @smenzer @patmmccann @dgirardi 

Seems like this should be something where their endpoint just returns empty ID, stopping this in prebid seems to be a not great option depending on the "down time" since publishers stay on older versions for some time.

dgirardi:
I am also doubtful - why is having a broken client and server better than just having a broken server? 
This does nothing for people running older versions, and the rest will be required to update yet again once the service comes back online.

smenzer:
i agree. this should be paused server side and an empty id/response returned so that the module doesn't break but doesn't provide an ID any more.

PR #10408, opened 2023-08-29
Triplelift Bid Adapter: stop override on placement

### LINKED ISSUE #10407: Triplelift bid adapter: overrides publisher selections
## Type of issue
Bug

## Description
With https://prebid.org/prebid-5-0-release/ and the adoption of https://github.com/prebid/Prebid.js/issues/5237 ; we mandated bid adapters take the openRTB fields publishers set in mediatypes.video

### Expected results
Setting video.placement is passed to triplelift's openrtb payload

### Actual results
Triplelift overrides the publisher logic with https://github.com/prebid/Prebid.js/blob/e78e4b89df78dcb0578bd74be33d6c2b92e41f81/modules/tripleliftBidAdapter.js#L242C2-L242C57

If a publisher wants to set context in prebid to indicate to prebid there is a player and not a gpt unit requiring a renderer, triplelift should not override their selection of video.placement placing them in jeopardy of IVT enforcement by DSPs

patmmccann:
fixes #10407

### DIFF
--- a/modules/tripleliftBidAdapter.js
+++ b/modules/tripleliftBidAdapter.js
@@ -239,7 +239,11 @@ function _getORTBVideo(bidRequest) {
   } catch (err) {
     logWarn('Video size not defined', err);
   }
-  if (video.context === 'instream') video.placement = 1;
+  if (video.context === 'instream') {
+    if (!video.placement) {
+      video.placement = 1;
+    }
+  }
   if (video.context === 'outstream') {
     if (!video.placement) {
       video.placement = 3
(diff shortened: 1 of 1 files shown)


nllerandi3lift:
Is this necessary even with `placement:1` being the only true instream value? Is not `context:'instream', placement:2` (or any other value) just wrong? Otherwise I see no harm here, but thought I'd fire this Q off first.

patmmccann:
Some publishers have been instructed by DSPs that they must set video.placement equal to a value other than one when they have muted pre-roll. In this case video.context will remain 'instream' even though video.placement has some other value.

fowler446 (approved):

PR #10411, opened 2023-08-29
ttd Bid Adapter: use bidderRequestId instead of auctionid

khang-vu-ttd:
## Type of change
- [x] Bugfix

No

## Description of change

Notice that ttd adapter is using `bidderRequest.auctionId` instead of  `bidderRequest.bidderRequestId`.

Looking at https://github.com/prebid/Prebid.js/commit/df66cdb1b90ddde1aaa7960b9f3c09eb89242ee8 it seems like it should be  `bidderRequest.bidderRequestId`

### DIFF
--- a/modules/ttdBidAdapter.js
+++ b/modules/ttdBidAdapter.js
@@ -4,7 +4,7 @@ import { registerBidder } from '../src/adapters/bidderFactory.js';
 import { BANNER, VIDEO } from '../src/mediaTypes.js';
 import {isNumber} from '../src/utils.js';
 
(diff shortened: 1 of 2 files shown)


khang-vu-ttd:
Just to confirm
`bidderRequest.bidderRequestId` will never be null?

unlike` auctionid`/`transactionid` (`bidderRequest.ortb2.source.tid`/`ortbImp.ext.tid`) which are controlled by the `pbjs.setConfig({enableTIDs: true})` flag

patmmccann (approved):

patmmccann:
That's correct, not sure how I missed this, good catch

PR #10416, opened 2023-08-30
sspBC Bid Adapter: add support for topicsFpdModule

wojciech-bialy-wpm:
## Type of change
- [x] Feature

## Description of change
Add support for User.data segments 600-609 (topics data created by topicsFpdModule.js).
Reformat data so it's properly parsed by sspBC exchange.

### DIFF
--- a/modules/sspBCBidAdapter.js
+++ b/modules/sspBCBidAdapter.js
@@ -12,7 +12,7 @@ const SYNC_URL = 'https://ssp.wp.pl/bidder/usersync';
 const NOTIFY_URL = 'https://ssp.wp.pl/bidder/notify';
 const GVLID = 676;
 const TMAX = 450;
-const BIDDER_VERSION = '5.9';
+const BIDDER_VERSION = '5.91';
 const DEFAULT_CURRENCY = 'PLN';
 const W = window;
 const { navigator } = W;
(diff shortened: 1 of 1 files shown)


modules/sspBCBidAdapter.js
```diff
@@ -199,6 +199,22 @@ const applyClientHints = ortbRequest => {
   ortbRequest.user = { ...ortbRequest.user, ...ch };
 };
 
+const applyTopics = (validBidRequest, ortbRequest) => {
+  const userData = validBidRequest.ortb2?.user?.data || [];
+  const topicsData = userData.filter(dataObj => {
```
dgirardi, line 204: the `ortb2` object you get as input is meant to be usable as-is for your ORTB exchange, so I'm always curious to learn why the extra processing. 

From what I see here you prefer to:

 - *not* get all the available user data segments, but only topics;
 - *not* get information about who provided the topics (`data[].name`).

Both choices seem odd to me, which is why I note them here, but of course you're free to make them.

dgirardi (approved):
LGTM!

modules/sspBCBidAdapter.js
wojciech-bialy-wpm: Our exchange is, unfortunately, not fully ortb-compliant  - that's why we:
a) trim data in topics
b) custom-build other user segments (see applyClientHints etc)

The next major version of the adapter should use ortbConverter, but we're waiting for the changes on the backend (to eliminate the need for overrides on the adapter side).

PR #10425, opened 2023-09-01
LiveIntent UserId module: fix Ajax timeout when calling the collector endpoint

3link:
## Type of change
- [x] Bugfix

## Description of change
Prebid has changed the semantics of the Ajax timeout 0. It used to mean "no timeout". In Prebid version 8 it mean "immediate" timeout. This affected the calls to the collector endpoint which had a default timeout of 0. This PR introduces a default non-zero Ajax timeout for the collector endpoint, makes that timeout configurable and also makes the default Ajax timeout for the resolution endpoint more explicit (not hidden in the live-connect dependency). 

## Other information
- documentation: https://github.com/prebid/prebid.github.io/pull/4835

### DIFF
--- a/modules/liveIntentIdSystem.js
+++ b/modules/liveIntentIdSystem.js
@@ -12,6 +12,7 @@ import { gdprDataHandler, uspDataHandler } from '../src/adapterManager.js';
 import {getStorageManager} from '../src/storageManager.js';
 import {MODULE_TYPE_UID} from '../src/activities/modules.js';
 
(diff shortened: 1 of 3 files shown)


3link:
@ChrisHuie Do you have an idea why the tests are failing.

ChrisHuie:
@3link it is saying the package and package-lock file are out of sync. Maybe try running npm install and see if that syncs the files back up

3link:
@ChrisHuie Thank you! Fixed.

ChrisHuie (approved):

PR #10433, opened 2023-09-05
Draft: feat: dynamic renderer support

ccorbo:
## Type of change

- [ ] Bugfix
- [x] Feature
- [ ] New bidder adapter   
- [ ] Build related changes
- [ ] CI related changes

- [ ] Other

## Description of change
This is a draft PR that is related to github issue: 

This is an implementation of adding dynamic renderer support into the Prebid Server bid adapter. This will allow a renderer to be specified into the bid response.

In the PR this rendererurl is specified via bid.seatbid[].bid[].ext.rendererurl. This is just used as an example, and can be changed.

Since the rendererurl is passed in via the bid response, the validation of the bid was changed to look for a rendererurl on the bid response.


modules/prebidServerBidAdapter/ortbConverter.js
```diff
@@ -101,6 +101,7 @@ const PBS_CONVERTER = ortbConverter({
     }
 
     const bidResponse = buildBidResponse(bid, context);
+    bidResponse.rendererurl = bid.ext.rendererurl;
```
bretg, line 104: I've opened the server-side issue https://github.com/prebid/prebid-server/issues/3085 to discuss the location of this field. Am proposing seatbid[].bid[].ext.prebid.meta.rendererurl

bretg: Done in PBS-Go 0.274.0 and PBS-Java 2.2

ccorbo:
@patmmccann updated the PR this is ready for review

patmmccann (approved):

patmmccann:
LGTM! @lksharma or @dgirardi please take a second look

src/auction.js
```diff
@@ -740,6 +740,19 @@ function getPreparedBidForAuction(bid, {index = auctionManager.index} = {}) {
     bid.renderer.setRender(renderer.render);
   }
 
+  if (bid.rendererurl) {
```
dgirardi, line 743: I think this should be done where you updated `ortbConverter ` - since it only affects the PBS adapter, there's no reason to include the logic in core. It would also remove the need to update `video.js`.

ccorbo: updated to move into ortbconverter

dgirardi (approved):

ccorbo:
@dgirardi The following test is failing in the build step. This PR shouldnt have any affect on this test.
```
1) should initialize LiveConnect and forward the prebid version when decode and emit an event
     LiveIntentId
     TypeError: undefined is not an object (evaluating '_xhr.server.requests[0].url') (/tmp/_karma_webpack_8411/commons.js:340612)
     ```
     
    I am seeing this test fail locally on master branch as well. Could you kick off a retry in the build step, its only failing on safari it looks like.

**Review by bretg — COMMENTED** [review]
Does this approach still allow the publisher to override the bidder-supplied renderer if they wish?

**Comment by dgirardi** [comment]
> Does this approach still allow the publisher to override the bidder-supplied renderer if they wish?

It works the same way as a client adapter's renderer. Apparently that's one of three ways bids can have a renderer, in order of decreasing precedence:

1. `adUnit.mediaType.(video|banner|native).renderer`
2. `bid.renderer`
3. `adUnit.renderer`

(1) and (3) are controlled by the pub, so only (1) can override this.

**Review by bretg — APPROVED** [review]

File: modules/prebidServerBidAdapter/ortbConverter.js [hunk 3]
```diff
@@ -111,6 +113,19 @@ const PBS_CONVERTER = ortbConverter({
       bidResponse.adm = bidResponse.native.ortb;
     }
 
+    if (bidResponse.rendererurl) {
+      try {
+        bidResponse.renderer = Renderer.install({ url: bidResponse.rendererurl });
+        bidResponse.renderer.setRender((bid) => {
+          bidResponse.renderer.push(() => {
+            window.renderAd(bid);
```
dgirardi, line 121: How do the renderers look like? are they used to this name, and how do they deal with clashes?

bretg:
We ought to define what Server-Side adapters should know about this `rendererUrl`. My suggestion would be to keep it simple-from-a-Prebid-perspective like:

"The URL must point to javascript code and be completely self-sufficient - i.e. there's no way for client-side configuration."

dgirardi:
>  there's no way for client-side configuration

With just a single URL, I think this requires the renderer to "register" itself with Prebid in a standardized way. It would need to do something like

```
window._pbjsRenderers = window._pbjsRenderers || {}
window._pbjsRenderers[<rendererUrl>] = function(bid) {
  ....
}
```

so that the client can invoke the renderer. I think this needs to use a new name in the global scope since the `pbjs` global is not necessarily accessible from the renderer.

Alternatively, the response could include instructions on how to run the renderer (with something like  `rendererGlobal` alongside `rendererUrl`).

modules/prebidServerBidAdapter/ortbConverter.js
ccorbo: I mentioned this in the original issue https://github.com/prebid/Prebid.js/issues/10434 - renderers will need to have a standard api. I don't think there is avoiding this. What do you mean by clashes?
```
It's important to note that supporting a dynamic renderer and incorporating bid renderer installation into the core of Prebid.js would necessitate a standardized API for each renderer. Currently, different adapters have varying methods for rendering an outstream video player.

An example of standardization for custom renderers for native ad units is in the linked documentation. The same API could be required in supporting this feature, that is required in the native ad unit case.

https://docs.prebid.org/prebid/native-implementation.html#43-implementing-the-custom-renderer-scenario
```

bretg:
Just noticed that the path is off here -- instead of seatbid[].bid[].ext.rendererurl, I think it's seatbid[].bid[].ext.**prebid.meta**.rendererurl

bretg:
So a related topic came up in regards to mobile outstream. I'm swimming outside my lane here, but in the interest of trying to move both efforts forward, perhaps we can set up conventions like this:

1. The URL returned in seatbid[].bid[].ext.prebid.meta.rendererurl  must point to a javascript file.
2. That javascript file must contain a function called "render()" that takes an adId.
3. The rendering code in Prebid.js/PUC renderAd() recognizes this "self-contained render function scenario by the existence of just the URL and creates an iframe into which it places javascript that invokes the required render() function:

```javascript
<iframe>
<script type="javascript" src="RENDERER_URL">
render(ADID);
</script>
</iframe>
```

4. The inline render function then calls postMessage with the adId to get the ad markup
5. Prebid-core responds to the postMessage, perhaps with the same logic currently done for PUC?
6. The inline render function does what it needs to display. 

Note: this can be for outstream, richmedia, pseudo-native, whatever. We shouldn't make assumptions about what's being rendered.

Errors:
- if the URL doesn't exist or its not javascript or there's no render() function, then render will fail.  Unclear Prebid can detect this, but if so, it should throw an event.

dgirardi:
> Question: would it be possible to just pass the render() function the whole creative body so it doesn't have to do the postMessage?

Yes (and I assume SDK can do the same), by inserting a bit more logic around the call to `render`). However if we are defining a standard API and inviting everyone to rewrite their renderers we should consider:

 - what do we mean by creative body? in JS the "natural" object to render is the bid response. What I'd call "body" varies depending on the media type and even if we limit it  to say video and a vast URL I'm not sure that's enough info to render.
 - what should we do about https://github.com/prebid/Prebid.js/issues/10780 - should the renderer have access to the top window? can we run it inside the creative iframe? (I am also very unsure of how outstream works in practice).

bretg:
> what do we mean by creative body?

The ORTB 'adm' field is the obvious choice. If it was VAST that was cached, perhaps the URL? But it would be nice to push people to deal with VAST directly or createObjectURL() rather than this silly client-side caching just to get an address. 

> should the renderer have access to the top window? 

I think we should start with "no" until someone makes an ironclad case for it. Renderers should be able to display as long as they can get the creative body.

dgirardi:
> The ORTB 'adm' field is the obvious choice.

I'm not sure that's enough - how about width / height? burl / nurl ? Not to mention that we don't always have an ORTB response (not sure if that's also the case for the SDK).

bretg:
> width / height? burl / nurl ? 

Ok then, pass the whole bidresponse object and let them figure it out.

bretg:
The general idea is that the PUC could get modified to drop this iframe

```
<iframe>
<script type="javascript" src="##RENDERER_URL##">
render(##BIDRESPONSEOBJECT##);
</script>
</iframe>
```

where BIDRESPONSEOBJECT gets resolved to a JSON object

```
{bidresponse: { adid: X, width: W, height: H  }}
```

The PBC-cached bid would contain this info, but not sure if/how a webview could ask the SDK for this info?

patmmccann (changes requested):
see committee notes

patmmccann:
discussed june 5 https://docs.google.com/document/d/1AojPDLnJzXaxwmFE5U7Qsm2GAV1fVwjcx4YyI6kX9ew/edit

bretg (commented):
see comments...

bretg:
@dgirardi - any thoughts on this?

dgirardi:
Broadly I agree, but I think we need an example implementation as a first step.

> The proposal is that we just standardize the interface to the bid response object format used by Prebid.js. 

bid responses are not very stable, and only a portion of them is needed for rendering. Prebid currently extracts a stable subset for interaction with PUC / remote creatives and that's what I would use, except those never deal with video and I don't really know what's involved there. Hence why I think we need some real world example of a video renderer to get a better idea of what data needs to go through.

patmmccann:
@lcorrigall to work with team on developing ix bid adapter to work with a client side renderer that can render in frame, that example to be used for further work

bretg:
@dgirardi and I had a discussion over slack. Here's the proposal:

The interface for the javascript file would be:
```
render({width, height, adUrl, config}, win)
```

It would be placed in the page something like this:
```
<iframe>
  <script type="javascript" src="##RENDERER_URL##"></script>
  <script>
  render({width, height, adUrl, config}, window);
  </script>
</iframe>
```
But these details may vary and shouldn't matter to the author of the renderer.

Config is renderer-specific and may cover things like display details. e.g. the Rubicon renderer would support options the publisher could set like this:

```
pbjs.setConfig({
  rendererConfig:{
    rubicon: {
      closeButton: "style1",
      label: "label",
      collapse: false
    },
    ix: {
       ... any ix-specific parameters ...
    }
  }
});
```
Perhaps someone familiar with renderer display details could standardize that interface so multiple vendors could utilize the same pub-supplied config?

bretg:
@lcorrigall - any comments on the proposal here?

bretg:
Looking for further input here on the proposal that Demetrio and I put on the table.

bretg (changes requested):
Still need input on the proposed interface

dgirardi:
I'd add to https://github.com/prebid/Prebid.js/pull/10433#issuecomment-2362209337 that `render` should render into the window `win`, and it would be great if that window could be a safeframe (not sure about the feasibility of this).

ccorbo:
Sorry for the delay here - The proposed interface looks good from ix's pov.

zachsavishinsky:
Curious if we want to keep this PR open right now.

patmmccann:
@dgirardi should we keep this open or will the solution to #10434 be in a diff branch?

Issue #10184, opened 2023-07-05
Aligning TCF and Activity Controls


bretg:
## Type of issue
Enhancement

## Description
In addition to adding support for Purpose 4 as described in #10161 - the committee discussed a number of other modernizations of Prebid.js's TCF toolkit by aligning it more closely with the new Activity Controls.

1. `transmitEids` - enhance removal of EIDs in the GDPR context
    1. By default, remove user.eids and user.ext.eids if there's no legal bases for processing for any Purposes 2-10. (align with Prebid Server)
    2. With publisher config, remove EIDs if there's no consent for Purpose 4.
2. Expand the `transmitUfpd` activity to cover the following fields in the user object:  user.id, user.buyeruid, user.customdata
3. Support the `transmitGeo` activity in Prebid.js. If device.geo or user.geo is supplied, the TCF enforcement module should check for opt-in to Special Feature 1, and round any lat/long contained in that data as required, and revise device.geo.accuracy to account for the rounding

### transmitEids config

It's proposed that the 'GDPR Enforcement' module gains a new config flag that only applies to the new personalizedAds purpose: `eidsRequireP4consent`. If true, P4 must have consent or user.eids and user.ext.eids will be removed. If false, any Purpose 2-10 must have consent or user.eids and user.ext.eids will be removed.

```
 pbjs.setConfig({
   consentManagement: {
     gdpr: {
       cmpApi: 'iab',
       defaultGdprScope: true,
       rules: [{ 
         ...
       },{
         purpose: "personalizedAds",
         eidsRequireP4consent: true,
         enforcePurpose: true,
         enforceVendor: true
       }]
     }
   }
 });
```

### transmitGeo config

The proposed TCF configuration for transmitting precise lat/long is: 

```
 pbjs.setConfig({
   consentManagement: {
     gdpr: {
       cmpApi: 'iab',
       defaultGdprScope: true,
       rules: [{ 
         ...
       },{
         purpose: "transmitPreciseGeo",
         enforcePurpose: true
       }]
     }
   }
 });
```

Notes:
- this is not actually a "purpose" in TCF terms, but it's close enough that we'll use the same general config structure. 
- if not specified, this "purpose" is always enforced
- Determining consent for this purpose is done by looking up "Special Feature 1" in the TCF string. If opted-in, consent is granted.
- There are no vendor-level checks, so the `enforceVendors` config is not supported.

patmmccann:
[...]

PR #10435, opened 2023-09-05
gdprEnforcement: transmitEids and transmitPreciseGeo activity controls

dgirardi:
## Type of change

- [x] Feature

## Description of change

- By default, prevent bidders from receiving EIDs unless consent or vendor exceptions were given for purposes 2-10
- Introduce `eidsRequireP4Consent` flag; when in a 'personalizedAds' gdprEnforcement rule, replaces the complicated algorithm above with a simple check against purpose 4
- expand the scope of the 'transmitUfpd' control to cover a number of other ortb2 fields
- introduce `transmitPreciseGeo` gdpr enforcement rule, which when activated checks against "special feature 1" opt-in

## Other information

Closes https://github.com/prebid/Prebid.js/issues/10184


patmmccann:
this is looking pretty good, are there a bunch of docs changes needed?

Fawke (approved):
This looks good, but it may require some docs update.

ChrisHuie (approved):

PR #10440, opened 2023-09-06
Prebid Core: Sync endpoint macro replacement

pm-priyanka-deshmane:
## Type of change
- [x] Feature

## Description of change
For the `prebidServerBidAdapter` there is no option to pass extra information to `syncEndpoint` from `s2sConfig`.
Hence adding a function which will replace the macros in syncEndpoint, if present, from `s2sConfig.extSyncMacros`.
e.g.

>         pbjs.setConfig({
> 
>             s2sConfig: [{
>                 extSyncMacros: {
>                     'PUBID': '12345',
>                     'VERSION': '1'
>                 }
>             }]
> 
>         });
> 

If the above configs are provided on page, and the syncEndpoint is
`https://prebid.adnxs.com/pbs/v1/cookie_sync?pubId={{PUBID}}&version={{VERSION}}`
then it will be modified as
`https://prebid.adnxs.com/pbs/v1/cookie_sync?pubId=12345&version=1`


pm-priyanka-deshmane:
@ChrisHuie Thanks for the review. Unit tests added.

bretg:
@pm-priyanka-deshmane - please describe the use case for this feature. It seems unnecessary to me.

1) The Prebid Server sync endpoint goes to exactly one place. Why would you need a macro in it? Just define your values directly in s2sConfig:

```
   syncEndpoint: 
        s2sConfig: [{
            syncEndpoint: "https://prebid.adnxs.com/pbs/v1/cookie_sync?pubId=12345&version=1"
        }]
```
What's above seems so much better in every way than what's been suggested by the PR:
```
        s2sConfig: [{
            syncEndpoint: "https://prebid.adnxs.com/pbs/v1/cookie_sync?pubId=##PUBID##&version=##VERSION##",
            extSyncMacros: {
                'PUBID': '12345',
                'VERSION': '1'
            }
        }]
```

If you can produce the s2sConfig with `extSyncMacros`, I don't see why you can't just set `syncEndpoint` to whatever you want. This will override whatever's defined by `defaultVendor` data.

2) If this is about the individual sync URLs coming back from Prebid Server, it's already possible to modify those using using the `syncUrlModifier` feature.

pm-priyanka-deshmane:
@bretg Thanks for your inputs. This change is not needed. Closing the PR.

PR #10446, opened 2023-09-07
Adnuntius Bid Adaptor: Enable choosing bidType for cpm bid

antosarho:
## Type of change

- [x] Feature

## Description of change

Via a parameter in the config or bid, determines whether to use grossBid or netBid as the cpm value in the bid response.

## Other information

This change introduces a new optional parameter, `bidType`, which will require an update to the adaptor documentation but it does not affect functionality for existing users.


antosarho:
Related documentation change is here: https://github.com/prebid/prebid.github.io/pull/4849

PR #10467, opened 2023-09-11
Yieldmo Bid Adapter : dont require params.video

robertrmartinez:
## Type of change
- [X] Bugfix

## Description of change
The [Yieldmo Bidder Docs](https://docs.prebid.org/dev-docs/bidders/yieldmo.html?pbjs_debug=true) show example configurations where video ortb data can be in either `mediaTypes.video` OR `params.video`

However it breaks if at least one params.video parameter is NOT there (not safe checking Object.keys)

Not sure if this is intended or not, but it was an easy fix so I updated it.

image

image

### DIFF
--- a/modules/yieldmoBidAdapter.js
+++ b/modules/yieldmoBidAdapter.js
@@ -432,12 +432,12 @@ function openRtbImpression(bidRequest) {
     }
   };
 
-  const mediaTypesParams = deepAccess(bidRequest, 'mediaTypes.video');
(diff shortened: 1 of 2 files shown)


robertrmartinez:
@patmmccann Not sure who to tag on this one, do you have a yieldmo contact who can take a look?

patmmccann:
yep

patmmccann (approved):

PR #10473, opened 2023-09-12
Dxkulture Bid Adapter: changing default video placement value

dani-nova:
## Type of change

- [ ] Bugfix
- [ ] Feature
- [ ] New bidder adapter   
- [ ] Build related changes
- [ ] CI related changes

- [ ] Other

## Description of change

Changing default video placement value


modules/dxkultureBidAdapter.js
```diff
@@ -329,7 +329,7 @@ function buildVideoRequestData(bidRequest, bidderRequest) {
 
   // Placement Inference Rules:
   // - If no placement is defined then default to 1 (In Stream)
-  video.placement = video.placement || 2;
+  video.placement = video.placement || 1;
```
patmmccann, line 332: video.placement is deprecated; why not also support plcmt? The reason for the deprecation was actually logic like this; the buy side described instream as over-specified. 

Also pls cover in unit tests

patmmccann (changes requested):

modules/dxkultureBidAdapter.js
dani-nova: @patmmccann We are still operating on openrtb 2.5. We made suggested changes

ChrisHuie (approved):

Issue #10445, opened 2023-09-07
Floors module: default floor price does not apply


Pupis:
## Type of issue

Bug or question

## Description
Prebid.js documentation says, that it is enough to configure adUnit with the default value, but there is no example with global floor.
```
...
 floors: {
     default: 1.00     // default currency is USD
 },
 ...
```
[link to documentation](https://docs.prebid.org/dev-docs/modules/floors.html#floors-defined-in-the-adunit)

By looking into the source code, it seems that is necessary to have values in floors object [link to source code](https://github.com/prebid/Prebid.js/blob/master/modules/priceFloors.js#L381) and only `default` is insufficient, and `skipped` value to floor data is set to true, and therefore `getFloor()` does not exist.

Is this approach correct overall? If not, could you provide an example with the default floor value?
Also, is it possible to define default value while calling `setConfig`?

### Expected results
Bid adapter to be able to get the default floor value.

### Actual results
`getFloor()` does not exist.

patmmccann:
Are you saying you set a skip rate and you are surprised there is no floor when the floor logic is skipped?

patmmccann:
As an aside I generally read this logic as a bug, conflating random skip and no match; these are very different for analytics purposes. 
`
  // if we still do not have a valid floor data then floors is not on for this auction, so skip
  if (Object.keys(deepAccess(resolvedFloorsData, 'data.values') || {}).length === 0) {
    resolvedFloorsData.skipped = true;
  } else {
    // determine the skip rate now
    const auctionSkipRate = getParameterByName('pbjs_skipRate') || resolvedFloorsData.skipRate;
    const isSkipped = Math.random() * 100 < parseFloat(auctionSkipRate);
    resolvedFloorsData.skipped = isSkipped;
  }
`

dgirardi:
I believe the documentation is incorrect in this:

> Alternatively, if there’s only one mediaType in the AdUnit and a single global floor, the syntax gets easier:

from my testing both `schema` and `values` are required, and given that I'm not sure what the point of `default` is. @robertrmartinez ?

modessef1:
I have exactly the same problem. I've set up the Prebid configuration just like OP, and the SSP bids below the default floor. They don't seem to see any floor in the requests I send them at all

dgirardi:
A workaround, to achieve what I think `default` is meant to do, is to put in the adUnit:

```
            floors: {
                schema: {
                    fields: ['mediaType'],
                },
                values: {
[...]

PR #10475, opened 2023-09-12
priceFloors: fix bug where `default` does not work on adUnit-level floors

dgirardi:
## Type of change

- [x] Bugfix

## Description of change

Fixes https://github.com/prebid/Prebid.js/issues/10445


Fawke (approved):
LGTM

Issue #10419, opened 2023-08-30
Fledge for GPT module: Further auction config enrichment


patmmccann:
## Type of issue
Feature

## Description
Enrich further auction config post https://github.com/prebid/Prebid.js/pull/10393

Possible objects: user.data
user.ext.data
eids
site.content.data
site.pagecat
site.kwarray 

....

Seems we want the whole ortb2 and ortb2Imp output; with activity controls on transmit honored.

patmmccann:
interesting side note: https://github.com/WICG/turtledove/blob/main/FLEDGE.md#211-providing-signals-asynchronously; auction signals can be promises

Issue #9481, opened 2023-01-30
Fledge module without GPT


patmmccann:
## Type of issue
Feature

## Description
On https://github.com/prebid/Prebid.js/issues/7634#issuecomment-1227814549 it was clear there were two possible paths in the initial issue proposal, and Fledge for GPT was implemented. We still have publisher demand for https://github.com/prebid/Prebid.js/issues/7634#issuecomment-1227814549

JoelPM:
This was discussed at the Prebid PMC meeting on 2/15/23. It sounds like the goal is to enable Prebid.js to manage the top-level Fledge Auction for publishers who aren't using GAM as the ad server. Given the evolving nature of Fledge and the limited testing happening right now I think it makes sense to revisit this later in Q2 '23.

wojciech-bialy-wpm:
We are (as a publisher) looking into the same issue - i.e. executing top-level fledge auction outside GAM.

In test environment, we do it by emiting new CONSTANTS.EVENTS.FLEDGE_AUCTION event in onFledgeAuctionConfigs(), and storing incoming componentAuctions outside prebid.js.

But perhaps Prebid should expose getFledgeAuctionsForAdunitCode (or a similar) method?

remysaissy:
> This was discussed at the Prebid PMC meeting on 2/15/23. It sounds like the goal is to enable Prebid.js to manage the top-level Fledge Auction for publishers who aren't using GAM as the ad server. Given the evolving nature of Fledge and the limited testing happening right now I think it makes sense to revisit this later in Q2 '23.

Hello, we (Teads) are considering using Prebid without GPT for Fledge auctions. Hence I wonder what was the outcome of the revisit?

Thanks!

patmmccann:
Closed as dupe of 10690

PR #10477, opened 2023-09-13
RAS Bid Adapter: fledge support

wsusrasp:
## Type of change

- [ ] Bugfix
- [x] Feature
- [ ] New bidder adapter   
- [ ] Build related changes
- [ ] CI related changes

- [ ] Other

## Description of change

Add support for the fledge.

### DIFF
--- a/modules/rasBidAdapter.js
+++ b/modules/rasBidAdapter.js
@@ -130,23 +130,23 @@ const getGdprParams = (bidderRequest) => {
 };
 
 const parseAuctionConfigs = (serverResponse, bidRequest) => {
-  if (isEmpty(serverResponse) || isEmpty(bidRequest)) {
+  if (isEmpty(bidRequest)) {
     return null;
   }
-  let auctionConfigs = [];
-  const bidConfigs = Object.fromEntries(bidRequest.bidIds.map(x => [x.bidId, x]));
+  const auctionConfigs = [];
+  const gctx = serverResponse && serverResponse.body?.gctx;
 
-  serverResponse.ads.filter(bid => bidConfigs.hasOwnProperty(bid.id) && bidConfigs[bid.id].fledgeEnabled).forEach((bid) => {
+  bidRequest.bidIds.filter(bid => bid.fledgeEnabled).forEach((bid) => {
     auctionConfigs.push({
-      'bidId': bidConfigs[bid.id].bidId,
+      'bidId': bid.bidId,
       'config': {
         'seller': 'https://csr.onet.pl',
-        'decisionLogicUrl': `https://csr.onet.pl/${bidConfigs[bid.id].network}/v1/protected-audience-api/decision-logic.js`,
+        'decisionLogicUrl': `https://csr.onet.pl/${encodeURIComponent(bid.params.network)}/v1/protected-audience-api/decision-logic.js`,
         'interestGroupBuyers': ['https://csr.onet.pl'],
         'auctionSignals': {
-          'site': bidConfigs[bid.id].site,
-          'kvismobile': bidConfigs[bid.id].isMobile,
-          'iusizes': bidConfigs[bid.id].iusizes
+          'params': bid.params,
+          'sizes': bid.sizes,
+          'gctx': gctx
         }
       }
     });
@@ -180,10 +180,8 @@ export const spec = {
     const bidIds = bidRequests.map((bid) => ({
       slot: bid.params.slot,
       bidId: bid.bidId,
-      network: network,
-      site: bid.params.site,
-      isMobile: Boolean(bid.params.pageContext?.keyValues?.ismobile),
-      iusizes: getAdUnitSizes(bid),
+      sizes: getAdUnitSizes(bid),
+      params: bid.params,
       fledgeEnabled: fledgeEligible
     }));
 
@@ -196,12 +194,9 @@ export const spec = {
 
   interpretResponse: function (serverResponse, bidRequest) {
     const response = serverResponse.body;
-    if (!response || !response.ads || response.ads.length === 0) {
-      return [];
-    }
 
-    const fledgeAuctionConfigs = parseAuctionConfigs(response, bidRequest);
-    const bids = response.ads.map(buildBid).filter((bid) => !isEmpty(bid));
+    const fledgeAuctionConfigs = parseAuctionConfigs(serverResponse, bidRequest);
+    const bids = (!response || !response.ads || response.ads.length === 0) ? [] : response.ads.map(buildBid).filter((bid) => !isEmpty(bid));
 
     if (fledgeAuctionConfigs) {
       // Return a tuple of bids and auctionConfigs. It is possible that bids could be null.
--- a/modules/rasBidAdapter.js
+++ b/modules/rasBidAdapter.js
@@ -129,6 +129,36 @@ const getGdprParams = (bidderRequest) => {
   return queryString;
 };
 
+const parseAuctionConfigs = (serverResponse, bidRequest) => {
+  if (isEmpty(serverResponse) || isEmpty(bidRequest)) {
+    return null;
+  }
+  let auctionConfigs = [];
+  const bidConfigs = Object.fromEntries(bidRequest.bidIds.map(x => [x.bidId, x]));
+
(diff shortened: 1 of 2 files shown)


modules/rasBidAdapter.js
```diff
@@ -129,6 +129,36 @@ const getGdprParams = (bidderRequest) => {
   return queryString;
 };
 
+const parseAuctionConfigs = (serverResponse, bidRequest) => {
+  if (isEmpty(serverResponse) || isEmpty(bidRequest)) {
+    return null;
```
patmmccann, line 134: why not append fledge if it is empty?

modules/rasBidAdapter.js
```diff
@@ -129,6 +129,36 @@ const getGdprParams = (bidderRequest) => {
...
+  serverResponse.ads.filter(bid => bidConfigs.hasOwnProperty(bid.id) && bidConfigs[bid.id].fledgeEnabled).forEach((bid) => {
+    auctionConfigs.push({
+      'bidId': bidConfigs[bid.id].bidId,
+      'config': {
+        'seller': 'https://csr.onet.pl',
+        'decisionLogicUrl': `https://csr.onet.pl/${bidConfigs[bid.id].network}/v1/protected-audience-api/decision-logic.js`,
+        'interestGroupBuyers': ['https://csr.onet.pl'],
+        'auctionSignals': {
+          'site': bidConfigs[bid.id].site,
+          'kvismobile': bidConfigs[bid.id].isMobile,
```
patmmccann, line 148: we might be putting this in auction signals for you

modules/rasBidAdapter.js
wsusrasp: In the first iteration of implementation we were planning to return the configs in the serverResponse, but as you pointed out now there is no reason to do that. I'll correct the code.

modules/rasBidAdapter.js
wsusrasp: Do you mean putting in the parameters from adunit?

wsusrasp:
> On a side note about the fledge mechanism - would it be possible to add a method to retrieve the component auction configs for adunit manually instead of automatically running them? Similarly to how getAdserverTargetingForAdUnitCode works right now. The use-case for this scenario would be greater control on how the auctions are ran and in our specific case that our prebid request is being made before the gpt slots are registered, since they are configured to obtain ads from GAM immediately.

The problem is that we are loading gpt asynchronously and it's not always available when required by prebid and changing this logic to load gpt first and then prebid would make the code run slower, as now it runs in parallel.

modules/rasBidAdapter.js
patmmccann: nice, so we're filling in the prebid win price here and adding more stuff too https://github.com/prebid/Prebid.js/pull/10393

patmmccann: would love your comments on #10419

dgirardi:
>  would it be possible to add a method to retrieve the component auction configs for adunit manually instead of automatically running them? Similarly to how getAdserverTargetingForAdUnitCode

Targeting works on "complete" bids, so the control you refer to allows you (as the publisher) to pick winners. Fledge is different, only GPT has access to the auction result. The only control the publisher has is who to include in the auction, which is already possible with various combination of `fledgeEnabled` and `ortb2Imp.ext.ae` flags.

My understanding is that to do the equivalent of `getAdserverTargeting` (or, really, `bidsBackHandler`) we would need a live service running Prebid-controlled decision logic. The snag is that with this approach we cannot include GAM in the fledge auctions - GPT is gatekeeping it.

> in our specific case that our prebid request is being made before the gpt slots are registered, since they are configured to obtain ads from GAM immediately.

I'm not sure I understand - wouldn't this be a problem regardless of fledge, if your slots are not waiting for Prebid?

modules/rasBidAdapter.js
wsusrasp: Ah, I see what you mean. This does look very useful. I'll try to look into #10419 soon, but I'll need to check what data we are using now and what could possibly be included.

wsusrasp:
> @wsusrasp GPT isn't available by the time the bids are back? Can you open your request as a new issue?
> 
> @skoklowski linking #9481 to your comment. Could you elaborate on how that would enable not using GAM, perhaps in the issue above? Are you envisioning the configs are submitted to a different top level auction runner, eg Microsoft, or that Prebid or the pub runs the top level?
> 
> Aside: I'll be in Germany next week if you want to meet up

Returning raw config would make it possible to call runAdAuction from navigator, without the need for GPT

wsusrasp:
> > would it be possible to add a method to retrieve the component auction configs for adunit manually instead of automatically running them? Similarly to how getAdserverTargetingForAdUnitCode
> 
> Targeting works on "complete" bids, so the control you refer to allows you (as the publisher) to pick winners. Fledge is different, only GPT has access to the auction result. The only control the publisher has is who to include in the auction, which is already possible with various combination of `fledgeEnabled` and `ortb2Imp.ext.ae` flags.
> 
> My understanding is that to do the equivalent of `getAdserverTargeting` (or, really, `bidsBackHandler`) we would need a live service running Prebid-controlled decision logic. The snag is that with this approach we cannot include GAM in the fledge auctions - GPT is gatekeeping it.
> 
> > in our specific case that our prebid request is being made before the gpt slots are registered, since they are configured to obtain ads from GAM immediately.
> 
> I'm not sure I understand - wouldn't this be a problem regardless of fledge, if your slots are not waiting for Prebid?

Our slots are waiting for prebid, but we set their config via getAdserverTargeting and they are defined after we make a call for bids. That's why if possible we'd like to be able to set the component configs manually on the slot. Would that be possible? Perhaps something like pbjs.fledge.getComponentConfigs(auctionId)? If needed, I can provide a draft in a separate PR.

dgirardi:
> Our slots are waiting for prebid, but we set their config via getAdserverTargeting and they are defined after we make a call for bids. 

Do you mean that the GPT slots are created only after running Prebid? Otherwise I think the existing flow should work. If it doesn't I'd like to understand why - since there are several other places where Prebid looks up slots in GPT.

My understanding is that as long as the slot exists, the [fledge setConfig](https://github.com/prebid/Prebid.js/blob/e18ab5161a4edb312faf307336eba815106b4e55/modules/fledgeForGpt.js#L49-L53) should work even if done before your other configuration.

wsusrasp:
> > Our slots are waiting for prebid, but we set their config via getAdserverTargeting and they are defined after we make a call for bids.
> 
> Do you mean that the GPT slots are created only after running Prebid? Otherwise I think the existing flow should work. If it doesn't I'd like to understand why - since this is not the only place where Prebid looks up slots in GPT.
> 
> My understanding is that as long as the slot exists, the [fledge setConfig](https://github.com/prebid/Prebid.js/blob/e18ab5161a4edb312faf307336eba815106b4e55/modules/fledgeForGpt.js#L49-L53) should work even if done before your other configuration.

It is exactly as you said, the GPT slots are defined after running Prebid. I do not have the precise reason, although since we are loading GPT in parallel it would still lead to race condition. The simple fix is to force the GPT to load before prebid (in series) but that would increase the time for the ads to load.

skoklowski:
@dgirardi: Both Prebid.js and GPT scripts are asynchronous. We load them in parallel to speed up time when ads are displayed. Thanks to this approach it's more likely that user won't leave the page until ads are displayed. In addition Active View metric should be better.

I will try to describe this process. Let's assume that we need for example 2 seconds for loading GPT script and only 1 second to load Prebid.js script. Additionally we set for example 0.8 seconds as a timeout for Prebid standard auctions (in fact this is timeout for bidders to respond).

**Our current approach**
We start Prebid auction immediately without waiting for GPT script to load so we can start fetching ads from Google Ad Manager after 2 seconds - exactly the moment when GPT script is loaded because in this eample Prebid needs only 1.8 seconds to do its job.

**Your approach**
You need to wait for GPT script to be fully loaded before running Prebid auction (I don't mean Protected Audience API auction but ordinary Prebid auction) because publisher are not able to define any slot before full GPT API is available (defineSlot function is not defined before that moment). So ads can be fetched after 2.8 seconds even though Prebid script was loaded after 1 second.

patmmccann (approved):

PR #10483, opened 2023-09-13
Core: do not log errors about invalid XML when serializing XHR objects

dgirardi:
## Type of change

- [x] Bugfix

## Description of change

Since https://github.com/prebid/Prebid.js/pull/10340, serializing a (fake) XHR object will log an error for responses that are not XML-formatted. This change leaves the error message only for the case when someone is explicitly asking for XML.

### DIFF
--- a/src/ajax.js
+++ b/src/ajax.js
@@ -89,6 +89,17 @@ export function fetcherFactory(timeout = 3000, {request, done} = {}) {
 
 function toXHR({status, statusText = '', headers, url}, responseText) {
   let xml = 0;
+  function getXML(onError) {
+    if (xml === 0) {
+      try {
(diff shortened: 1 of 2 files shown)


coreyb-cbs:
@dgirardi Thanks for this! It appears to work in my local testing - the error is gone.

For context, when testing an upgrade from 7.53.0 to 8.13.0 I experienced the following new error:
`Prebid ERROR: TypeError: Failed to execute 'parseFromString' on 'DOMParser': The provided value 'undefined' is not a valid enum value of type SupportedType.` from getAdUnitCodes` in `src/targeting.js`.

ChrisHuie (approved):

PR #10503, opened 2023-09-19
update appnexus PBS default configs

jsnellbaker:
## Type of change
- [x] Other

## Description of change
This PR removes a fully deprecated appnexus PBS endpoint config and adds the syncEndpoint for the appnexus PSP default config.

Confirmed these changes with @SyntaxNode

### DIFF
--- a/modules/prebidServerBidAdapter/config.js
+++ b/modules/prebidServerBidAdapter/config.js
@@ -1,24 +1,15 @@
 // accountId and bidders params are not included here, should be configured by end-user
 export const S2S_VENDORS = {
-  'appnexus': {
-    adapter: 'prebidServer',
-    enabled: true,
-    endpoint: {
-      p1Consent: 'https://prebid.adnxs.com/pbs/v1/openrtb2/auction',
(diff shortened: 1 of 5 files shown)


modules/prebidServerBidAdapter/config.js
```diff
@@ -1,27 +1,18 @@
 // accountId and bidders params are not included here, should be configured by end-user
 export const S2S_VENDORS = {
-  'appnexus': {
```
dgirardi, line 3: this is used by some [integration examples and unit tests](https://github.com/search?q=repo%3Aprebid%2FPrebid.js+defaultVendor%3A+%27appnexus%27&type=code), those should probably also be updated.

SyntaxNode (approved):

modules/prebidServerBidAdapter/config.js
jsnellbaker: Pushed a change to get the remaining references

dgirardi (approved):

PR #10510, opened 2023-09-21
UID2: support Client-side token generation and build tag

jingyi-gao-ttd:
## Type of change
- [x] Feature

## Description of change
This PR is to introduce the Client-side Token Generation (CSTG) mode to the UID2 module. Publishers now have the flexibility to generate UID2 tokens directly from the client side, eliminating the need for initial server-side token generation.

## Other information
Since it's still backward compatible with existing functionality and we don't wanna widely advertise CSTG for now, we will not add CSTG to the UID2 module document until a few more pubs successfully use cstg inside prebid.js natively.

document PR: https://github.com/prebid/Prebid.js/pull/10629


integrationExamples/gpt/esp_example.html
```diff
@@ -3,7 +3,7 @@
 <head>
     <link rel="icon" type="image/png" href="/favicon.png">
     <script async src="//www.googletagservices.com/tag/js/gpt.js"></script>
-    <script async src="prebid.js"></script>
+    <script  type="text/javascript"  src="../../build/dev/prebid.js" async></script>
```
lionell-pack-ttd, line 6: What's the reason for this change?

jingyi-gao-ttd: image
This is the error I got when I trying to test esp_example, and I found this file hasn't been touch for a while now, so i am trying to fix the example

ChrisHuie:
@jingyi-gao-ttd it looks like our integration tests are unable to run on this pr. This is likely because you are following a different instance of prebid in circleci such as a personal fork of it. If you go into circleci and unselect the prebid instance you are currently following on the left that should kick off our code integration tests :)

jingyi-gao-ttd:
@ChrisHuie Thanks Chris! I could integration tests are running now

modules/uid2IdSystem_shared.js
```diff
@@ -160,18 +464,137 @@ function refreshTokenAndStore(baseUrl, token, clientId, storageManager, _logInfo
...
       latestToken: response.identity,
     };
+    let storedTokens = storageManager.getStoredValueWithFallback();
+    if (storedTokens?.originalIdentity) tokens.originalIdentity = storedTokens.originalIdentity;
     storageManager.storeValue(tokens);
     return tokens;
   });
 }
 
+function normalizeEmail(email) {
```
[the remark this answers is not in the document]
jingyi-gao-ttd: Hi @dgirardi 
Yes, this is a translation of our server-side code. We decided to revisit this code and refactor it, make sure this will be covered by tests

> gmail special case. gmail.com is not the only domain used by google mail

It is a decision our UID2 team made, we will keep this as it is to make it consistent with server-side

thanks.

patmmccann:
this feels like a lot of code to add for a feature you arent recommending anyone use; is there a way to ensure it is not included in the build unless someone is using it

patmmccann (changes requested):
Please make this a little more friendly to publishers building with your module and not using cstg

lionell-pack-ttd:
> this feels like a lot of code to add for a feature you arent recommending anyone use; is there a way to ensure it is not included in the build unless someone is using it

We're expecting this to quickly become the primary way to use the module - we've had a lot of feedback from publishers that the existing requirement for server-side work to generate the tokens is time-consuming and difficult, and they'd prefer a client-side-only option.

I agree with finding a way to reduce the code added for publishers who aren't using this feature.

patmmccann:
see #8593

jingyi-gao-ttd:
@patmmccann Thanks for the advice, we've added a feature flag `UID2_CSTG`  to be able to exclude CSTG code in the build. Could you please take another review?

modules/uid2IdSystem_shared.js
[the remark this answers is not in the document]
jingyi-gao-ttd: Hi @dgirardi, hope you are well. We've refactored this function and added tests to cover this. Could you please take a review? Thanks

modules/uid2IdSystem_shared.js
```diff
@@ -160,18 +173,508 @@ function refreshTokenAndStore(baseUrl, token, clientId, storageManager, _logInfo
...
+      );
+
+      return parsedAddress ? `${parsedAddress}@${domain}` : undefined;
+    }
+
+    static splitEmailIntoAddressAndDomain(email) {
+      const parts = email.split('@');
+      if (
+        !parts.length ||
+        parts.length !== 2 ||
```
dgirardi, line 359: first clause, `!parts.length`, is redundant since if it's falsy it's also always `!== 2`

dgirardi (approved):
LGTM

modules/uid2IdSystem_shared.js
jingyi-gao-ttd: Good point, I removed that, thanks

patmmccann:
Did I miss a docs pull link? If so, is the build tag documented in it?

patmmccann (approved):

jingyi-gao-ttd:
> Did I miss a docs pull link? If so, is the build tag documented in it?
We've prepared a pull request for documentation at https://github.com/prebid/Prebid.js/pull/10629. However, at this stage, we would prefer not to widely promote the CSTG feature. We'd like to proceed with merging this feature for the next release without updating the documentation for now.

@patmmccann, your insights on this matter would be appreciated.

patmmccann:
As discussed in slack, we're not able to merge sans documentation, particularly of the build flag

PR #10534, opened 2023-09-26
Missena Bid Adapter: floor implementation

pdamoc:
## Type of change

- [x] Feature
## Description of change

This is an implementation of the passing of the relevant price floor data to our server. 

## Other information

The PR also includes a small refactoring of the event notification used for `onBidWon` that is also reused for `onTimeout`.

### DIFF
--- a/modules/missenaBidAdapter.js
+++ b/modules/missenaBidAdapter.js
@@ -1,5 +1,11 @@
-import { buildUrl, formatQS, logInfo, triggerPixel } from '../src/utils.js';
-import { BANNER } from '../src/mediaTypes.js';
+import {
+  isFn,
+  deepAccess,
+  formatQS,
+  logInfo,
+  parseSizesInput,
+} from '../src/utils.js';
+import { BANNER, NATIVE, VIDEO } from '../src/mediaTypes.js';
 import { registerBidder } from '../src/adapters/bidderFactory.js';
 
 const BIDDER_CODE = 'missena';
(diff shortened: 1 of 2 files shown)


ChrisHuie:
@pdamoc can you please add units tests for these updates

pdamoc:
We will add the tests for the floor changes and remove the notification refactoring (which we will submit in a later PR).

modules/missenaBidAdapter.js
```diff
@@ -7,6 +16,48 @@ const ENDPOINT_URL = 'https://bid.missena.io/';
...
+    height: parseInt(heightStr, 10) || undefined,
+  };
+}
+
+/* Get Floor price information */
+function getFloor(bidRequest) {
+  if (!isFn(bidRequest.getFloor)) {
+    return {};
+  }
+  const sizesArray = getSizeArray(bidRequest);
```
dgirardi, line 35: I don't think this is necessary, the floors module will already pick size from the request when it makes sense. What this does is force using the first size when there are multiple instead of looking for a floor that applies to all sizes. 

Unless that was your intention I recommend removing this. (I am learning now that a lot of other adapters do this, so I suspect you're just following the pattern, which is unfortunately not a good one).

dgirardi: To clarify - if you just call `getFloor()`, it will attempt to find reasonable defaults:

 - `currency` defaults to 'USD'
 - `mediaType` defaults to the request's if it contains only one
 - `size` defaults to the request's if it contains only one (which implies it also has only one mediaType)

pdamoc: @dgirardi the functions you mentioned have been removed by @ysfbsf  (code original creator). We will rely on our partners to properly configure their bidFloors (the code was there for corner cases). 

Please let us know if we need to change anything else.

pdamoc:
@ChrisHuie is there anything else that is blocking this PR from being merged into the master?

Issue #10540, opened 2023-09-27
Access to ID of the auction won by cached bid in BID_WON event


msmeza:
## Type of issue
Feature request

## Description
As far as I can see, there's no direct way of finding the ID of an auction won by a cached bid through `BID_WON` events; the bids included in the events only contain the ID of their original auctions. 

Ideally, we'd like to be able to find the ID of the auction won by a cached bid directly from the data included in a `BID_WON` event, to be able to map them using only the data received from our analytics adapter.

## Other information
I searched the list of issues and I do not believe that this has been brought up before, though I did find an issue where there's a mention of multiple winning bids with the same auction ID coming from an analytics adapter, which can happen when cached bids win auctions (https://github.com/prebid/Prebid.js/issues/6411). However, this issue seems to be more about finding the actual winner of an auction ID with multiple winners, rather than identifying the auction IDs of auctions won by cached bids.

We believe that being able to map winning cached bids with the auctions that they actually won would be useful for others as well. For example, if I understood the code correclty, Magnite seems to have a function to try and do that in their analytics adapter (getRenderingIds).

I'm sure there are better ways to achieve our goal, but I submitted a PR with a way of doing it to help with the discussion (https://github.com/prebid/Prebid.js/pull/10541).

PR #10541, opened 2023-09-27
Prebid core: cached bids should contain the ID of the auction they actually won

msmeza:
## Type of change
- Feature

## Description of change
Cached bids should contain the ID of the auctions that they actually won so that this information is available through `BID_WON` events.

## Other information
More information about the motivation for this PR can be found here: https://github.com/prebid/Prebid.js/issues/10540

### DIFF
--- a/src/targeting.js
+++ b/src/targeting.js
@@ -464,19 +464,17 @@ export function newTargeting(auctionManager) {
       if (typeof filterFunction === 'function') {
         bidsReceived = bidsReceived.filter(bid => latestAuctionForAdUnit[bid.adUnitCode] === bid.auctionId || !!filterFunction(bid))
       }
-
-      bidsReceived = bidsReceived
-        .map(bid => {
-          if (latestAuctionForAdUnit[bid.adUnitCode] !== bid.auctionId) {
-            bid.latestAuctionForAdUnit = latestAuctionForAdUnit[bid.adUnitCode];
-          }
-          return bid;
-        });
     }
 
     bidsReceived = bidsReceived
       .filter(bid => deepAccess(bid, 'video.context') !== ADPOD)
       .filter(isBidUsable);
+
+    bidsReceived
+      .forEach(bid => {
+        bid.latestTargetedAuctionId = latestAuctionForAdUnit[bid.adUnitCode];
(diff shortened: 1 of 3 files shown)


src/targeting.js
```diff
@@ -464,6 +464,14 @@ export function newTargeting(auctionManager) {
       if (typeof filterFunction === 'function') {
         bidsReceived = bidsReceived.filter(bid => latestAuctionForAdUnit[bid.adUnitCode] === bid.auctionId || !!filterFunction(bid))
       }
+
+      bidsReceived = bidsReceived
+        .map(bid => {
```
dgirardi, line 469: this looks more suited to `forEach`. Also I think it should be done after all filtering is done, around line 480, or do you see a reason why it's better to tag them here?

src/targeting.js
```diff
@@ -464,6 +464,14 @@ export function newTargeting(auctionManager) {
...
+          if (latestAuctionForAdUnit[bid.adUnitCode] !== bid.auctionId) {
+            bid.latestAuctionForAdUnit = latestAuctionForAdUnit[bid.adUnitCode];
```
dgirardi, line 471: - IMO this should be set on all bids in the pool, not just those from a previous auction.
- I don't like the name, "forAdUnit" is redundant since the object is a bid and already tied to a particular ad unit. Maybe `latestTargetedAuctionId`? Asking also @patmmccann for opinions. I'd prefer to make targeting explicit since this only runs when the publisher uses prebid's targeting logic, which in theory they could do without.

patmmccann: On naming , I agree with a shorter name yes

src/targeting.js
msmeza: You're right, `forEach` is mor suited for this.

Yes, I agree that the property should be set after all filtering is done, provided it should be set on all the bids in the pool (as you mentioned in your comment below). The reason I added my change where I did was to avoid increasing the size of bid objects in the cases where `auctionId` would have the same value as the new property. But I think it's a good idea to set the property on all bids in the pool.

dgirardi (approved):

patmmccann (approved):

PR #10542, opened 2023-09-28
dgkeyword: use ortb2

yki-t:
## Type of change

- [x] Bugfix
- [ ] Feature
- [ ] New bidder adapter   
- [ ] Build related changes
- [ ] CI related changes

- [ ] Other

### Purpose
dgkeyword submodule to be compatible with Prebid >8.0.0

### Changes
- use ortb2 on setting keywords
- prevent to use keywords for specified websites or specified users, because to use keywords or not is related to ad-units

### DIFF
--- a/modules/dgkeywordRtdProvider.js
+++ b/modules/dgkeywordRtdProvider.js
@@ -6,8 +6,7 @@
  * @module modules/dgkeywordProvider
  * @requires module:modules/realTimeData
  */
-
-import {logMessage, deepSetValue, logError, logInfo, mergeDeep} from '../src/utils.js';
+import { logMessage, deepSetValue, logError, logInfo, isStr, isArray } from '../src/utils.js';
 import { ajax } from '../src/ajax.js';
 import { submodule } from '../src/hook.js';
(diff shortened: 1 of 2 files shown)


modules/dgkeywordRtdProvider.js
```diff
@@ -156,4 +147,32 @@ function init(moduleConfig) {
...
+      } else {
+        result += `${key},`;
+      }
+    } else if (isArray(keywords[key])) {
+      if (keywords[key][0] === '') {
+        result += `${key},`
+      } else {
+        keywords[key].forEach(val => {
+          result += `${key}=${val},`
+        });
```
ncolletti, line 168: Needs a check for the val type as String. Additional unit tests would be helpful

yki-t: @ncolletti 
We updated type check and added an unit test

yki-t: I force pushed because by mistake we included our unrelated commits on the forked repository

ncolletti (approved):

PR #10547, opened 2023-09-28
Draft: test page demo

GeneGenie:
This is a sample PR demonstrating prebid issue with userIds


test/pages/userId.html
```diff
@@ -0,0 +1,133 @@
...
+                name: 'criteo',
+                storage:{
+                  type:'html5',
+                  name:'criteo_id'
+                }
+              }
+            ]
+          }
+        })
+        pbjs.setConfig({
```
patmmccann, line 70: this is your problem; this needs to be a mergeConfig

PR #10555, opened 2023-09-29
ZetaGlobalSsp Analytics Adapter : make event object with only needed fields

asurovenko-zeta:
## Type of change
- [x] Bugfix

Due to the complex object (**args.doc**) with cyclic dependence of fields, we had a bug associated with the impossibility of turning this object into a json. Therefore, we are forced to make logic that will leave only the fields we need in the object.

Please note that **JSON.stringify(args)** may produce an exception because of that.


robertrmartinez (changes requested):
please add a unit test or two

asurovenko-zeta:
@robertrmartinez done

robertrmartinez (approved):

PR #10558, opened 2023-10-02
Core & currency module: fix bug where bid responses are collected after auction ends

dgirardi:
## Type of change

- [x] Bugfix

## Description of change

Fix a race condition where, if the currency rates file takes about as much to load as the auction does to time out, bid responses will sometimes be added to the auction after it has already ended.


Fawke (approved):
LGTM

ChrisHuie (approved):

ChrisHuie (approved):

PR #10562, opened 2023-10-03
imds Bid Adapter : fix incorrectly named user sync property

ecammit:
## Type of change

- [x] Bugfix

## Description of change

- This changes the "image" usersync from the incorrectly used "pixel" property. This was preventing the new image user sync from happening and causing a warning in the console with debug enabled of: `WARNING: User sync type "pixel" not supported`.
- Removed a warning being emitted for valid 204 no-bid responses with an empty body.
- There is also a change to the markdown documentation to match https://github.com/prebid/prebid.github.io/pull/4908

## Other information

https://github.com/prebid/prebid.github.io/pull/4908

### DIFF
--- a/modules/imdsBidAdapter.js
+++ b/modules/imdsBidAdapter.js
@@ -224,7 +224,6 @@ export const spec = {
     };
 
     if (!serverResponse.body || typeof serverResponse.body != 'object') {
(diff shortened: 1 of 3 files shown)


ecammit:
Note: Failing circleci is from some unrelated bid adapter (`lm_kiviadsBidAdapter`) that's consistently failing.

ChrisHuie (approved):

PR #10572, opened 2023-10-04
Preciso BidAdapter: UserId mapped in bid request

PrecisoSRL:
## Type of change

- [ ] Bugfix
- [ ] Feature
- [ ] New bidder adapter   
- [ ] Build related changes
- [ ] CI related changes

- [ ] Other


test/spec/modules/precisoBidAdapter_spec.js
```diff
@@ -145,7 +145,7 @@ describe('PrecisoAdapter', function () {
     })
   })
   describe('getUserSyncs', function () {
-    const syncUrl = 'https://ck.2trk.info/rtb/user/usersync.aspx?id=NA&gdpr=0&gdpr_consent=&us_privacy=&t=4';
+    const syncUrl = 'https://ck.2trk.info/rtb/user/usersync.aspx?id=12355454test&gdpr=0&gdpr_consent=&us_privacy=&t=4';
```
robertrmartinez, line 148: I am confused how this test was needing to be updated since the change is only on the `buildRequests` part, not the `getUserSyncs part.

robertrmartinez (changes requested):
Also, please add a unit test coving this userId change

test/spec/modules/precisoBidAdapter_spec.js
PrecisoSRL: Sorry for the confusion. We added the same userId into the usersync URL, so we also included it in the test case. We apologize for not mentioning that change in this PR.

modules/precisoBidAdapter.js
```diff
@@ -23,7 +23,7 @@ export const spec = {
   buildRequests: (validBidRequests = [], bidderRequest) => {
     // convert Native ORTB definition to old-style prebid native definition
     validBidRequests = convertOrtbRequestToProprietaryNative(validBidRequests);
-    // userId = validBidRequests[0].userId.pubcid;
+    userId = validBidRequests[0].userId.pubcid;
```
robertrmartinez, line 26: It is possible that the `userId` object is `undefined` / `null` which would cause this to error and not make a request to your exchange.

You can update this to

`userId = validBidRequests[0].userId?.pubcid;`

Note, if `userId.pubcid` does not exist, then your `userId` var will be set to `undefined` and sent as empty string in your userSync URL like so

`${URL_SYNC}id=&gdpr`;

test/spec/modules/precisoBidAdapter_spec.js
```diff
@@ -73,6 +73,7 @@ describe('PrecisoAdapter', function () {
 
       expect(data.city).to.be.a('string');
       expect(data.geo).to.be.a('object');
+      expect(data.userId).to.be.a('object');
       // expect(data.userId).to.be.a('string');
       // expect(data.imp).to.be.a('object');
```
robertrmartinez, line 78: probably can remove commented out code.

Also, might be good to check that userId object value is `pubcid: '27e21631-975a-4410-85f3-c08b1f1c469f'` since `null` is technically an `object`

expect(data.userId).to

```js
expect(data.userId).to.deep.equal({
      pubcid: '27e21631-975a-4410-85f3-c08b1f1c469f'
});
```

robertrmartinez (changes requested):
one fix + test suggestion

PR #10603, opened 2023-10-12
Qortex RTD Provider: add module to submodules.json

mickannese:
After our code was merged, we realized that this file was missing from the PR. In its current state, the download page does not add the required rtdModule into the generated prebid build, leading to errors when implemented on publisher pages.

Please advise if this should resolve the issue and thank you for the help!

### DIFF
--- a/modules/.submodules.json
+++ b/modules/.submodules.json
@@ -75,6 +75,7 @@
       "oneKeyRtdProvider",
       "optimeraRtdProvider",
       "permutiveRtdProvider",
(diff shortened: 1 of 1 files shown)


mickannese:
@ChrisHuie @Rothalack Hi all, just looping in those who have reviewed our code recently. Our release did not include submodules.json, so our module is not downloading with rtdModule attached. I'm not sure what could be causing this error in testing, please advise if possible

thank you

ChrisHuie (approved):

PR #10610, opened 2023-10-13
Relevant Bid Adapter : support s2s

hugolm84:
## Type of change

- [x] Feature

## Description of change

The `relevantdigital` bid adapter now supports s2s configurations and need to transform params so that server side parameters are valid.

## Other information

Related PR's

https://github.com/prebid/prebid-server/pull/3216
https://github.com/prebid/prebid.github.io/pull/4931

### DIFF
--- a/modules/relevantdigitalBidAdapter.js
+++ b/modules/relevantdigitalBidAdapter.js
@@ -98,7 +98,14 @@ export const spec = {
   isBidRequestValid: (bid) => bid.params?.placementId && getBidderConfig([bid]).complete,
 
   /** Trigger impression-pixel */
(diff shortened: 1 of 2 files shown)


spotxslagle (approved):
LGTM

patmmccann:
@bretg could you take a look; does this make your effort on another issue more difficult? 

Also this has no tests

bretg:
Talked to Patrick. The concern is that `transformBidParams` is something we'd hope to deprecate at some point so the client-side adapter is not necessary for server-side bidders. We'll bring up the topic in the next PMC meeting

bretg (changes requested):
Add unit tests please

hugolm84:
> Add unit tests please

Fixed

bretg (approved):

PR #10611, opened 2023-10-16
TPMN Bid Adapter: Change End-Point & Support Video

tpmn-admin:
## Type of change

- [ ] Bugfix
- [x] Feature
- [ ] New bidder adapter   
- [ ] Build related changes
- [ ] CI related changes

- [ ] Other

## Description of change

1. Change End-Point
2. Change Request Format (payload -> Ortb)
3. Support Video MediaType
4. Support OutStream Video
5. Update Document
6. Add Example (/integrationExamples/gpt/tpmn_example.html)

### DIFF
--- a/modules/tpmnBidAdapter.js
+++ b/modules/tpmnBidAdapter.js
@@ -115,10 +115,22 @@ function createRequest(bidRequests, bidderRequest, mediaType) {
   }
   if (bid.params.inventoryId) rtbData.ext = {};
   if (bid.params.inventoryId) rtbData.ext.inventoryId = bid.params.inventoryId
-  if (bid.params.bcat) rtbData.bcat = bid.params.bcat;
-  if (bid.params.badv) rtbData.badv = bid.params.badv;
-  if (bid.params.bapp) rtbData.bapp = bid.params.bapp;
+  
+  const ortb2Data = bidderRequest?.ortb2 || {};
+  const bcat = ortb2Data?.bcat || bid.params.bcat || [];
+  const badv = ortb2Data?.badv || bid.params.badv || [];
+  const bapp = ortb2Data?.bapp || bid.params.bapp || [];
 
+  if (bcat.length > 0) {
+    rtbData.bcat = bcat;
+  }
+  if (badv.length > 0) {
+    rtbData.badv = badv;
+  }
+  if (badv.length > 0) {
+    rtbData.bapp = bapp;
+  }
+  
   return {
     method: 'POST',
     url: BIDDER_ENDPOINT_URL + '?v=' + ADAPTER_VERSION,
--- a/modules/tpmnBidAdapter.js
+++ b/modules/tpmnBidAdapter.js
@@ -142,12 +142,6 @@ const CONVERTER = ortbConverter({
   bidResponse(buildBidResponse, bid, context) {
     const {bidRequest} = context;
     const bidResponse = buildBidResponse(bid, context);
-
-    utils.logWarn('Building bidResponse');
-    utils.logWarn('bid', bid);
-    utils.logWarn('bidRequest', bidRequest);
-    utils.logWarn('bidResponse', bidResponse);
-
     if (bidResponse.mediaType === BANNER) {
       bidResponse.ad = bid.adm;
     } else if (bidResponse.mediaType === VIDEO) {
--- a/modules/tpmnBidAdapter.js
+++ b/modules/tpmnBidAdapter.js
@@ -14,25 +14,7 @@ const SUPPORTED_AD_TYPES = [BANNER, VIDEO];
 // const BIDDER_ENDPOINT_URL = 'http://localhost:8081/ortb/pbjs_bidder';
 const BIDDER_ENDPOINT_URL = 'https://gat.tpmn.io/ortb/pbjs_bidder';
 const IFRAMESYNC = 'https://gat.tpmn.io/sync/iframe';
-const VIDEO_ORTB_PARAMS = [
-  'mimes',
-  'minduration',
-  'maxduration',
-  'placement',
-  'protocols',
-  'startdelay',
-  'skip',
-  'skipafter',
-  'minbitrate',
-  'maxbitrate',
-  'delivery',
-  'playbackmethod',
-  'api',
-  'linearity',
-  'battr',
-  'plcmt'
-];
-const BANNER_ORTB_PARAMS = [
+const COMMON_PARAMS = [
   'battr'
 ];
 
@@ -80,18 +62,22 @@ function isValidVideoRequest(bid) {
 }
 
 function buildRequests(validBidRequests, bidderRequest) {
-  if (validBidRequests.length === 0 || !bidderRequest) return [];
-  let bannerBids = validBidRequests.filter(bid => utils.deepAccess(bid, 'mediaTypes.banner'));
-  let videoBids = validBidRequests.filter(bid => utils.deepAccess(bid, 'mediaTypes.video'));
   let requests = [];
+  try {
+    if (validBidRequests.length === 0 || !bidderRequest) return [];
+    let bannerBids = validBidRequests.filter(bid => utils.deepAccess(bid, 'mediaTypes.banner'));
+    let videoBids = validBidRequests.filter(bid => utils.deepAccess(bid, 'mediaTypes.video'));
 
-  bannerBids.forEach(bid => {
-    requests.push(createRequest([bid], bidderRequest, BANNER));
-  });
+    bannerBids.forEach(bid => {
+      requests.push(createRequest([bid], bidderRequest, BANNER));
+    });
 
-  videoBids.forEach(bid => {
-    requests.push(createRequest([bid], bidderRequest, VIDEO));
-  });
+    videoBids.forEach(bid => {
+      requests.push(createRequest([bid], bidderRequest, VIDEO));
+    });
+  } catch (err) {
+    utils.logWarn('buildRequests', err);
+  }
 
   return requests;
 }
@@ -100,19 +86,7 @@ function createRequest(bidRequests, bidderRequest, mediaType) {
   const rtbData = CONVERTER.toORTB({ bidRequests, bidderRequest, context: { mediaType } })
 
   const bid = bidRequests.find((b) => b.params.inventoryId)
-  // console.log('createRequest : ', mediaType, bid);
 
-  if (!rtbData.site) rtbData.site = {}
-  rtbData.site = createSite(bidderRequest.refererInfo)
-
-  if (bidderRequest.gdprConsent) {
-    if (!rtbData.user) rtbData.user = {};
-    if (!rtbData.user.ext) rtbData.user.ext = {};
-    if (!rtbData.regs) rtbData.regs = {};
-    if (!rtbData.regs.ext) rtbData.regs.ext = {};
-    rtbData.user.ext.consent = bidderRequest.gdprConsent.consentString;
-    rtbData.regs.ext.gdpr = bidderRequest.gdprConsent.gdprApplies ? 1 : 0;
-  }
   if (bid.params.inventoryId) rtbData.ext = {};
   if (bid.params.inventoryId) rtbData.ext.inventoryId = bid.params.inventoryId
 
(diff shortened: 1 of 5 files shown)


AbdullahWins (commented):
Good Job!

modules/tpmnBidAdapter.js
```diff
@@ -1,126 +1,222 @@
...
+    if (!rtbData.user) rtbData.user = {};
+    if (!rtbData.user.ext) rtbData.user.ext = {};
+    if (!rtbData.regs) rtbData.regs = {};
+    if (!rtbData.regs.ext) rtbData.regs.ext = {};
+    rtbData.user.ext.consent = bidderRequest.gdprConsent.consentString;
+    rtbData.regs.ext.gdpr = bidderRequest.gdprConsent.gdprApplies ? 1 : 0;
+  }
+  if (bid.params.inventoryId) rtbData.ext = {};
+  if (bid.params.inventoryId) rtbData.ext.inventoryId = bid.params.inventoryId
+  if (bid.params.bcat) rtbData.bcat = bid.params.bcat;
```
patmmccann, line 117: you have accept openrtb specification of bcat as well, you can take your params as an override

modules/tpmnBidAdapter.js
```diff
@@ -131,34 +227,91 @@ function createSite(refInfo) {
...
+  if (videoSizes && videoSizes.length > 0) {
+    utils.deepSetValue(imp, 'video.w', videoSizes[0][0]);
+    utils.deepSetValue(imp, 'video.h', videoSizes[0][1]);
   }
-  return [parseSize(sizes)]; // or a single one ? (ie. [728,90])
-}
 
-function getBannerSizes(bidRequest) {
-  return parseSizes(deepAccess(bidRequest, 'mediaTypes.banner.sizes') || bidRequest.sizes);
+  VIDEO_ORTB_PARAMS.forEach((param) => {
```
patmmccann, line 243: this seems unnecessary; the ortb convertor is already doing all this for you. Is that correct @dgirardi ?

modules/tpmnBidAdapter.js
```diff
@@ -1,126 +1,222 @@
...
+const DEFAULT_CURRENCY = 'USD';
+const SUPPORTED_AD_TYPES = [BANNER, VIDEO];
+// const BIDDER_ENDPOINT_URL = 'http://localhost:8081/ortb/pbjs_bidder';
+const BIDDER_ENDPOINT_URL = 'https://gat.tpmn.io/ortb/pbjs_bidder';
+const IFRAMESYNC = 'https://gat.tpmn.io/sync/iframe';
+const VIDEO_ORTB_PARAMS = [
+  'mimes',
+  'minduration',
+  'maxduration',
+  'placement',
```
patmmccann, line 21: you're missing plcmt

tpmn-admin: [792fc2c](https://github.com/prebid/Prebid.js/pull/10611/commits/792fc2ce6e74debec8b342030393405f68ac9438)
add plcmt

[8bf7567](https://github.com/prebid/Prebid.js/pull/10611/commits/8bf75676b74ca461f4cf76b884bf99c8e720af52)
instream video is sound on default, outstream video sound off default

modules/tpmnBidAdapter.js
tpmn-admin: [da8e928](https://github.com/prebid/Prebid.js/pull/10611/commits/da8e9286419f7b9a254f9bdd0ca79ea8490d5acc)

update openrtb block attributes

modules/tpmnBidAdapter.js
tpmn-admin: i think.. this is needed to check which parameters are allowed.

dgirardi: I think [this example](https://github.com/prebid/Prebid.js/blob/master/libraries/ortbConverter/README.md#example-taking-video-parameters-from-bidrequestparamsvideo) does this with less code, as long as the parameters you're checking match the [ones used in ortbConverter](https://github.com/prebid/Prebid.js/blob/4a3fcf5b1fe843f17ba6ba19d9c5549dacf44f2f/libraries/ortbConverter/processors/video.js#L6-L29)

tpmn-admin: [2699115](https://github.com/prebid/Prebid.js/pull/10611/commits/2699115a76ed891ec7f911ad924206b399790ad9)

I modified the code in the direction you suggested. ([this example](https://github.com/prebid/Prebid.js/blob/master/libraries/ortbConverter/README.md#example-taking-video-parameters-from-bidrequestparamsvideo) )

tpmn-admin:
Hello.  @dgirardi 

I would like to request a review of the TPMN prebid.js pull request once again.

Regarding the code simplification using the previously mentioned 'ortbConverter,' we have decided to 'Revert' it.

We initially aimed to streamline the code using 'ortbConverter,' but the results generated using 'buildRequests' in the test cases behaved differently than expected.

In the test case 'should use bidder video params if they are set,' it does not perform as desired. We would like the new 'video parameter' settings to properly take effect when specifying them. The issue we encountered is that the new settings are not being reflected correctly, resulting in undesired outcomes.

The current code is working as intended at the moment.

I am requesting a code review in its current state.

Additionally, I'd like to inquire if there are any other points for consideration.

dgirardi:
@tpmn-admin we're always looking to improve so I'm interested in more details on what was not working! I posted [here](https://github.com/dgirardi/Prebid.js/commit/30ed60e67f4b6e4c543c498fd7e41327fa5ea706) the changes that I would submit for ortbConverter. Some notes:

 - I was not able to reproduce the video test failure, just following the example seems to work
 - My commit still  has 1 test failure for `✖ should build correct request for banner bid with both w, h`, because ortbConverter sets `format` instead of `w` and `h`. IMO `format` is better because it allows for multiple sizes, but of course it's your choice.
 - GDPR (and USP, GPP, etc) is also handled by ortbConverter, as well as `site`, so I removed those parts. I updated the tests as well since they need to be set up in a slightly different way.
 - Prebid does not expect to find `battr` in either `mediaTypes.video` or `mediaTypes.banner` - it's not in the [list of options for those objects](https://docs.prebid.org/dev-docs/adunit-reference.html#adunitmediatypesbanner). I updated a test case to set it in `params`.

test/spec/modules/tpmnBidAdapter_spec.js
```diff
@@ -238,35 +241,107 @@ describe('tpmnAdapterTests', function () {
...
+      //   expect(requests).to.deep.equal(null);
+
+      //   const request = requests[0].data;
+
+      //   expect(request.imp[0].video).to.deep.include({ ...check
+
+      //   });
+      // });
+
+      it('when mediaType New Video', () => {
```
tpmn-admin, line 271: @dgirardi 


I have a question.

When I run the `gulp serve-and-test --file test/spec/modules/tpmnBidAdapter_spec.js` command locally, everything works fine.

However, when I run the `gulp test --file test/spec/modules/tpmnBidAdapter_spec.js` command locally, the test fails. 
The issue is that the `const requests = spec.buildRequests([NEW_VIDEO_BID], BIDDER_REQUEST);` code fails because the 'requests.imp.video' object is not being created.

I've checked multiple times, but I can't figure out what I missed.

For banners, the tests pass successfully.

tpmn-admin: I'm reposting the code as follows after making sure it passes the test, taking into account using setTimeout in other test cases.

`
setTimeout(() => {
          expect(spec.isBidRequestValid(bid)).to.equal(true);
          const requests = spec.buildRequests([bid], BIDDER_REQUEST);
          const request = requests[0].data;
          expect(request.imp[0].video).to.deep.include({...check});
          done();
        }, 200);
`

tpmn-admin:
how Can i fix this build Fail?
```
TOTAL: 4 FAILED, 76539 SUCCESS


1) "before each" hook in "{root}"
     
     Error: done() called multiple times in hook <"before each" hook in "{root}"> (of root suite); in addition, done() received error: Error: Uncaught ReferenceError: done is not defined (/tmp/_karma_webpack_75500/commons.js:412324)

2) "before each" hook in "{root}"
     
     Error: done() called multiple times in hook <"before each" hook in "{root}"> (of root suite); in addition, done() received error: Error: Uncaught ReferenceError: done is not defined (/tmp/_karma_webpack_75500/commons.js:412324)

3) "after each" hook in "buildRequests"
     triplelift adapter buildRequests
     done() called multiple times in hook <triplelift adapter buildRequests "after each" hook in "buildRequests">; in addition, done() received error: Error: ReferenceError: done is not defined (/tmp/_karma_webpack_75500/commons.js:412324)

4) "before each" hook in "{root}"
     
     done() called multiple times in hook <"before each" hook in "{root}"> (of root suite); in addition, done() received error: Error: ReferenceError: Can't find variable: done (/tmp/_karma_webpack_75500/commons.js:412324)
[native code]
```

tpmn-admin:
> how Can i fix this build Fail?
> 
> ```
> TOTAL: 4 FAILED, 76539 SUCCESS
> 
> 
> 1) "before each" hook in "{root}"
>      
>      Error: done() called multiple times in hook <"before each" hook in "{root}"> (of root suite); in addition, done() received error: Error: Uncaught ReferenceError: done is not defined (/tmp/_karma_webpack_75500/commons.js:412324)
> 
> 2) "before each" hook in "{root}"
>      
>      Error: done() called multiple times in hook <"before each" hook in "{root}"> (of root suite); in addition, done() received error: Error: Uncaught ReferenceError: done is not defined (/tmp/_karma_webpack_75500/commons.js:412324)
> 
> 3) "after each" hook in "buildRequests"
>      triplelift adapter buildRequests
>      done() called multiple times in hook <triplelift adapter buildRequests "after each" hook in "buildRequests">; in addition, done() received error: Error: ReferenceError: done is not defined (/tmp/_karma_webpack_75500/commons.js:412324)
> 
> 4) "before each" hook in "{root}"
>      
>      done() called multiple times in hook <"before each" hook in "{root}"> (of root suite); in addition, done() received error: Error: ReferenceError: Can't find variable: done (/tmp/_karma_webpack_75500/commons.js:412324)
> [native code]
> ```

test/spec/modules/tpmnBidAdapter_spec.js
tpmn-admin: Through many trials and errors, 
I've come to realize that my test code didn't consider 'test-all-features-disabled' for the video type. 
I've made modifications to the test code and re-push it.

tpmn-admin:
@dgirardi 


1. I followed the guide([30ed60e](https://github.com/dgirardi/Prebid.js/commit/30ed60e67f4b6e4c543c498fd7e41327fa5ea706)) and used ortbConverter, removing unnecessary code in the process.
2. I realized later that I hadn't considered the 'test-all-features-disabled' case when writing the test code, so I made the necessary adjustments.
3. Now, the tests are running good.

modules/tpmnBidAdapter.js
```diff
@@ -1,164 +1,248 @@
...
-      if (gdprConsent && gdprConsent.consentString) {
-        if (typeof gdprConsent.gdprApplies === 'boolean') {
-          policyParam += `&gdpr=${Number(gdprConsent.gdprApplies)}&gdpr_consent=${gdprConsent.consentString}`;
-        } else {
-          policyParam += `&gdpr=0&gdpr_consent=${gdprConsent.consentString}`;
-        }
-      }
-      if (uspConsent && uspConsent.consentString) {
-        policyParam += `&ccpa_consent=${uspConsent.consentString}`;
+    utils.logWarn('Building bidResponse');
```
dgirardi, line 146: I'm not sure if I missed these in my previous review - I don't think they belong here, perhaps this was to help during development? Prebid already logs requests and responses.

dgirardi (approved):
I forgot to mention the feature disabled tests - sorry! This LGTM pending log cleanup that I mention below if you agree.

modules/tpmnBidAdapter.js
tpmn-admin: I added it while creating example.html, and it was reflected as is. I will remove that part.

tpmn-admin:
@patmmccann  Please request a code review once again.

tpmn-admin:
@dgirardi 
It looks like not progressing. What else should we do for the next steps?
Please let me know if I'm missing anything.

patmmccann (approved):

PR #10646, opened 2023-10-26
Adnuntius Bid Adaptor: Handle voidAuIds from server response

antosarho:
## Type of change

- [ ] Bugfix
- [x] Feature
- [ ] New bidder adapter   
- [ ] Build related changes
- [ ] CI related changes

- [ ] Other

## Description of change

- invalid ad units are put into a voidAuIds array in the ad server response
- this changeset stores those void ad units IDs into local storage and prevents calling to those same ad units.
- also stores any values from metaData from the adserver response into local storage and sends it with subsequent requests
- also some general refactoring


jlquaccia (approved):
LGTM

Issue #10657, opened 2023-10-30
getConfig cleanup of device, app, and site in bidders


patmmccann:
https://github.com/prebid/Prebid.js/blob/56292c4672a0b19bee34840dca93bd3e1e5b8afe/modules/insticatorBidAdapter.js#L110 is an example of referencing getConfig to gather a parameter about the device. The preferred implementation follows this https://github.com/prebid/Prebid.js/blob/56292c4672a0b19bee34840dca93bd3e1e5b8afe/modules/mgidBidAdapter.js#L201

There are many incorrect implementations in the project. 

https://github.com/search?q=repo%3Aprebid%2FPrebid.js+%22getConfig%28%27device%27%29%22&type=code

https://github.com/search?q=repo%3Aprebid%2FPrebid.js+%22getConfig%28%27site%27%29%22&type=code

https://github.com/search?q=repo%3Aprebid%2FPrebid.js+%22getConfig%28%27app%27%29%22&type=code

We no longer support site, device, nor app config directly accessed in bidders [core picks it up and puts it in ortb2, but it isn't documented]

patmmccann:
replaced by https://github.com/prebid/Prebid.js/issues/10747

PR #10666, opened 2023-10-31
getConfig cleanup of device

ausgomez:
Fix for https://github.com/prebid/Prebid.js/issues/10657


ausgomez:
@patmmccann What do you think?

ausgomez:
Moved to 10667

PR #10667, opened 2023-10-31
Insticator Bid Adapter : get config fix for device

ausgomez:
Fix for https://github.com/prebid/Prebid.js/issues/10657


ausgomez:
@patmmccann What do you think?

ChrisHuie (approved):

PR #10671, opened 2023-10-31
Rubicon Bid Adapter: PAAPI/Fledge support

spotxslagle:
## Type of change

- [ x ] Feature

## Description of change

Rubicon Bid Adapter support for passing PAAPI/Fledge auction configs


modules/rubiconBidAdapter.js
```diff
@@ -716,6 +720,16 @@ export const spec = {
     }, []).sort((adA, adB) => {
       return (adB.cpm || 0.0) - (adA.cpm || 0.0);
     });
+
+    let fledgeAuctionConfigs = responseObj.component_auction_config?.map(config => {
+      return { config, bidId: config.bidId }
```
robertrmartinez, line 725: Does this `bidId` returned by AE have to match the `bidId` prebid has for the bidRequest?

Where is DV+ getting this `bidId`?

modules/rubiconBidAdapter.js
```diff
@@ -520,6 +521,10 @@ export const spec = {
     if (configUserId) {
       data['ppuid'] = configUserId;
     }
+
+    if (bidderRequest.fledgeEnabled) {
```
dgirardi, line 525: I think you should look at `ortb2Imp.ext.ae` instead, the pub can enable it only for some slots.

modules/rubiconBidAdapter.js
```diff
@@ -716,6 +720,16 @@ export const spec = {
     }, []).sort((adA, adB) => {
       return (adB.cpm || 0.0) - (adA.cpm || 0.0);
     });
+
+    let fledgeAuctionConfigs = responseObj.component_auction_config?.map(config => {
+      return { config, bidId: config.bidId }
+    });
+
+    if (fledgeAuctionConfigs) {
```
dgirardi, line 728: `return {bids, fledgeAuctionConfigs}` should work, core will ignore the second property if it's not an array and still pick up the bids

modules/rubiconBidAdapter.js
spotxslagle: Yes. They're getting it from the fastlane param we pass in the adapter

modules/rubiconBidAdapter.js
spotxslagle: While that would be better, it breaks some unit tests it's not worth fixing right now. I'll plan to deal with this when we aren't trying to get this out so quickly

spotxslagle:
@dgirardi would you mind giving this another look? I'm hoping to get this merged today. Thank you!

dgirardi (approved):
LGTM!

robertrmartinez (approved):

Issue #10289, opened 2023-07-28
add catch all floor to beforeBidRequested event


patmmccann:
## Type of issue
We like the results of beforeBidRequested event to include a floor with mediatype *, the ad unit code, and size *

This would be useful for analytics adapters and publishers. 

##Current behavior
pbjs.onEvent('beforeRequestBids',(data) => console.log(data)) does not have a floor available, only the getFloor function

##Expected behavior
pbjs.onEvent('beforeRequestBids',(data) => console.log(data)) has a floor, not in the ortb place, that is numeric and not based on the getFloor function

patmmccann:
Another solution to this problem: perhaps we could somehow just identify what PBS is putting on imp.floor, which is https://github.com/prebid/Prebid.js/blob/dea49c275c2931cd689268ac7d72c6356fbaacbe/modules/prebidServerBidAdapter/ortbConverter.js#L142 or get PBS to reflect it back at us

PR #10672, opened 2023-10-31
ortbConverter: Emit Bid Floor Data

whorst:
## Type of change
- [ ] Bugfix
- [x] Feature
- [ ] New bidder adapter 

- [ ] Build related changes
- [ ] CI related changes

- [ ] Other

## Description of change
This changes adds a new [emitted event](https://docs.prebid.org/dev-docs/publisher-api-reference/getEvents.html) `pbsFloorValuesReceived` that can be used to capture bidFloor data.

## Other information
Used to resolve the issue #10289 highlighted by @patmmccann

### DIFF
--- a/test/spec/ortbConverter/bidfloor_spec.js
+++ b/test/spec/ortbConverter/bidfloor_spec.js
@@ -1,0 +1,207 @@
+/* eslint-disable no-trailing-spaces */
+import {expect} from 'chai';
+import {
+  PrebidServer as Adapter,
+  resetSyncedStatus,
+} from 'modules/prebidServerBidAdapter/index.js';
+import * as utils from 'src/utils.js';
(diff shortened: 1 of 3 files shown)


test/spec/ortbConverter/bidfloor_spec.js
```diff
@@ -0,0 +1,207 @@
+/* eslint-disable no-trailing-spaces */
```
dgirardi, line 1: this should be part of the PBS test suite, if we decide to do this only for PBS.

dgirardi (commented):
I am not sure about this approach:

- I don't think a new event is necessary, you should be able to get floors by calling `getFloor` from one of the existing events (like `bidRequested`)
- If we do need a change, why only for PBS?

Issue #10560, opened 2023-10-02
Poor error handling when loading currency rates


dgirardi:
## Type of issue
Bug

## Description

If the currency module is set up to load rates from the network (the default) and no default rates are specified, bid responses are put on hold until rates are available. This is a problem when the file does not load quickly enough or the request fails:

1. if rates are not available by the time an auction times out, bids are lost without necessarily causing an AUCTION_TIMEOUT event. Currently they'll generate NO_BID events instead; with https://github.com/prebid/Prebid.js/pull/10558 they'll generate no events. 
2. the same happens if rates fail to load - all auctions will wait until the timeout and then end without any bids.

I think it makes sense to address <span>#</span>2 by trying to fetch rates again at the start of the next auction, but I am unsure about <span>#</span>1. If an auction times out and we have pending bids but no rates, should we:

 - reject the bids?
 - emit AUCTION_TIMEOUT without any timed out bidders?
 - postpone the timeout to wait for rates?

dgirardi:
Consensus is:
 - bids should be rejected with a currency-specific reason
 - bids that do not actually need a conversion (currency is already the adserver currency) should not wait for rates

PR #10679, opened 2023-11-01
Currency module: better error handling

dgirardi:
## Type of change

- [x] Bugfix

## Description of change

This updates the logic around loading currency rates so that:

 - if they fail to load, we attempt again on the next auction
 - they take too long to load, pending bids are rejected (instead of just forgotten) when the auction times out

## Other information

Fixes https://github.com/prebid/Prebid.js/issues/10560


modules/currency.js
```diff
@@ -7,20 +7,22 @@ import {getHook} from '../src/hook.js';
 import {defer} from '../src/utils/promise.js';
 import {registerOrtbProcessor, REQUEST} from '../src/pbjsORTB.js';
 import {timedBidResponseHook} from '../src/utils/perfMetrics.js';
+import * as events from '../src/events.js';
```
robertrmartinez, line 10: IS there any reason / benefit to change this to only import the `on` and `off` functions of `events.js` since those are the only bits used I think?

robertrmartinez (approved):
Looks good! One question

Fawke (approved):

Issue #10516, opened 2023-09-21
Passing Chrome facilitated Testing modes to bidders


pm-harshad-mane:
Google Chrome will have two modes of Chrome-facilitated testing and will generate testing labels that can be accessed with a header or using JS API. 
Requesting changes in Fledge module to retrieve the testing labels using JS API and pass this information to bidders

https://developer.chrome.com/docs/privacy-sandbox/chrome-testing/
https://developer.chrome.com/docs/privacy-sandbox/chrome-testing/#accessing-the-cookiedeprecationlabel-javascript-api

patmmccann:
I'd be open to core

patmmccann:
So, to elaborate: based on the information at the link with regards to consent to device storage, I think we should add this to core enrichments subject and in gdpr be similar to vendorless modules that check purpose 1 consent to the publsiher and also that transmitufpd is allowed on the bidder before transmitting to the bidder

dgirardi:
From [here](https://developer.chrome.com/docs/privacy-sandbox/chrome-testing/#cookie-deprecation-value):

> Accessing labels involves accessing information stored on the user’s device. In some jurisdictions (for example, the EU and UK), we understand that this activity is analogous to the use of cookies and thus accessing labels likely requires end user consent. Before you begin requesting labels, we recommend that you seek legal advice as to whether this consent obligation applies to you.

@patmmccann's interpretation is that, for gdpr, prebid should check for purpose 1 consent to retrieve labels and purpose 4 consent to pass them on to a bidder. 

To me that reads like a bidder that wants those labels should be treated as if they were accessing device storage.

Either way I think this should require publisher opt-in if it's going to be in core. Otherwise we are effectively choosing what we think is the correct legal advice for publishers, the majority of which will not know that this feature even exists.

patmmccann:
as reference https://github.com/InteractiveAdvertisingBureau/openrtb/pull/143 by @dmdabbs

dmdabbs:
Ja, for sure P1 (ePrivacy).

dmdabbs:
I'm writing the IAB Europe Framework Signals Group folks to inquire whether this is something about which they would opine.

patmmccann:
Copied from a relevant email thread: 

[...]

PR #10635, opened 2023-10-20
33Across Analytics Adapter: initial release

macinjosh32:
## Type of change

- [ ] Bugfix
- [ ] Feature
- [x] New Analytics adapter   
- [ ] Code style update (formatting, local variables)
- [ ] Refactoring (no functional changes, no api changes)
- [ ] Build related changes
- [ ] CI related changes

- [ ] Does this change affect user-facing APIs or examples documented on http://prebid.org?
- [ ] Other

## Description of change

The debut of a new Prebid Analytics Adapter from 33Across.

## Other information

Documentation PR: https://github.com/prebid/prebid.github.io/pull/4944

PR #10683, opened 2023-11-02
FPD Enrichment: support for Cookie Deprecation Label

jlquaccia:
## Type of change
- [x] Feature

## Description of change
- Support for the Privacy Sandbox cookie deprecation label
- Core will access the Privacy Sandbox JS API for testing mode and place the result in device.ext.dcep
- A check for publisher purpose 1 consent before calling the Privacy Sandbox JS API will occur if the gdprEnforcement module is active

## Other information
https://github.com/prebid/Prebid.js/issues/10516

### DIFF
--- a/modules/gdprEnforcement.js
+++ b/modules/gdprEnforcement.js
@@ -282,6 +282,7 @@ function emitTCF2FinalResults() {
     eidsBlocked: formatSet(eidsBlocked),
     geoBlocked: formatSet(geoBlocked)
   };
+
   events.emit(CONSTANTS.EVENTS.TCF2_ENFORCEMENT, tcf2FinalResults);
   [storageBlocked, biddersBlocked, analyticsBlocked, ufpdBlocked, eidsBlocked, geoBlocked].forEach(el => el.clear());
 }
--- a/modules/gdprEnforcement.js
+++ b/modules/gdprEnforcement.js
@@ -192,7 +192,14 @@ export function validateRules(rule, consentData, currentModule, gvlId) {
   }
   const vendorConsentRequred = rule.enforceVendor && !((gvlId === VENDORLESS_GVLID || (rule.softVendorExceptions || []).includes(currentModule)));
   const {purpose, vendor} = getConsent(consentData, ruleOptions.type, ruleOptions.id, gvlId);
-  return (!rule.enforcePurpose || purpose) && (!vendorConsentRequred || vendor);
+
+  let validation = (!rule.enforcePurpose || purpose) && (!vendorConsentRequred || vendor);
+
+  if (gvlId === FIRST_PARTY_GVLID) {
+    validation = (!rule.enforcePurpose || !!deepAccess(consentData, `vendorData.publisher.consents.${ruleOptions.id}`));
+  }
+
+  return validation;
 }
 
 function gdprRule(purposeNo, checkConsent, blocked = null, gvlidFallback = () => null) {
@@ -202,11 +209,7 @@ function gdprRule(purposeNo, checkConsent, blocked = null, gvlidFallback = () =>
 
     if (shouldEnforce(consentData, purposeNo, modName)) {
       const gvlid = getGvlid(params[ACTIVITY_PARAM_COMPONENT_TYPE], modName, gvlidFallback(params));
-      let allow =
-        gvlid === FIRST_PARTY_GVLID
-          ? !!deepAccess(consentData, `vendorData.publisher.consents.1`)
-          : !!checkConsent(consentData, modName, gvlid);
-
+      let allow = !!checkConsent(consentData, modName, gvlid);
       if (!allow) {
         blocked && blocked.add(modName);
         return {allow};
--- a/modules/gdprEnforcement.js
+++ b/modules/gdprEnforcement.js
@@ -282,7 +282,6 @@ function emitTCF2FinalResults() {
     eidsBlocked: formatSet(eidsBlocked),
     geoBlocked: formatSet(geoBlocked)
   };
-
   events.emit(CONSTANTS.EVENTS.TCF2_ENFORCEMENT, tcf2FinalResults);
   [storageBlocked, biddersBlocked, analyticsBlocked, ufpdBlocked, eidsBlocked, geoBlocked].forEach(el => el.clear());
 }
--- a/modules/gdprEnforcement.js
+++ b/modules/gdprEnforcement.js
@@ -7,7 +7,7 @@ import {config} from '../src/config.js';
 import adapterManager, {gdprDataHandler} from '../src/adapterManager.js';
 import * as events from '../src/events.js';
 import CONSTANTS from '../src/constants.json';
-import {GDPR_GVLIDS, VENDORLESS_GVLID} from '../src/consentHandler.js';
+import {GDPR_GVLIDS, VENDORLESS_GVLID, FIRST_PARTY_GVLID} from '../src/consentHandler.js';
 import {
   MODULE_TYPE_ANALYTICS,
   MODULE_TYPE_BIDDER,
@@ -111,7 +111,7 @@ export function getGvlid(moduleType, moduleName, fallbackFn) {
     if (gvlMapping && gvlMapping[moduleName]) {
       return gvlMapping[moduleName];
     } else if (moduleType === MODULE_TYPE_PREBID) {
-      return VENDORLESS_GVLID;
+      return moduleName === 'cdep' ? FIRST_PARTY_GVLID : VENDORLESS_GVLID;
     } else {
       let {gvlid, modules} = GDPR_GVLIDS.get(moduleName);
       if (gvlid == null && Object.keys(modules).length > 0) {
@@ -166,6 +166,7 @@ export function shouldEnforce(consentData, purpose, name) {
 function getConsent(consentData, type, id, gvlId) {
   let purpose = !!deepAccess(consentData, `vendorData.${CONSENT_PATHS[type]}.${id}`);
   let vendor = !!deepAccess(consentData, `vendorData.vendor.consents.${gvlId}`);
+
   if (type === 'purpose' && LI_PURPOSES.includes(id)) {
     purpose ||= !!deepAccess(consentData, `vendorData.purpose.legitimateInterests.${id}`);
     vendor ||= !!deepAccess(consentData, `vendorData.vendor.legitimateInterests.${gvlId}`);
@@ -198,9 +199,14 @@ function gdprRule(purposeNo, checkConsent, blocked = null, gvlidFallback = () =>
   return function (params) {
     const consentData = gdprDataHandler.getConsentData();
     const modName = params[ACTIVITY_PARAM_COMPONENT_NAME];
+
     if (shouldEnforce(consentData, purposeNo, modName)) {
       const gvlid = getGvlid(params[ACTIVITY_PARAM_COMPONENT_TYPE], modName, gvlidFallback(params));
-      let allow = !!checkConsent(consentData, modName, gvlid);
+      let allow =
+        gvlid === FIRST_PARTY_GVLID
+          ? !!deepAccess(consentData, `vendorData.publisher.consents.1`)
+          : !!checkConsent(consentData, modName, gvlid);
+
       if (!allow) {
         blocked && blocked.add(modName);
         return {allow};
--- a/src/fpd/enrichment.js
+++ b/src/fpd/enrichment.js
@@ -18,6 +18,7 @@ export const dep = {
   getWindowSelf,
   getHighEntropySUA,
   getLowEntropySUA,
+  getCookieDeprecationLabel,
 };
 
 const oneClient = clientSectionChecker('FPD')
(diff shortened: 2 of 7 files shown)


jlquaccia:
@patmmccann @dgirardi, from https://github.com/prebid/Prebid.js/issues/10516, it sounded like we wanted this change in enrichment.js.  Let me know what you guys think.

src/fpd/enrichment.js
```diff
@@ -78,6 +95,34 @@ function removeUndef(obj) {
   return getDefinedParams(obj, Object.keys(obj))
 }
 
+export async function tryToGetCdepLabel(cb = getCookieDeprecationLabel) {
+  let cdep;
+  const consentData = gdprDataHandler.getConsentData();
```
dgirardi, line 100: This should do a check on the `accessDevice` activity instead (somewhat like [this](https://github.com/prebid/Prebid.js/blob/30ac7f5dff10cd5ce9841c4451641d797e888d4f/src/storageManager.js#L32-L34)). 

Normally that would be enough - the TCF logic (which is just one of the ways storage is gated) is attached to `accessDevice` [in its own module](https://github.com/prebid/Prebid.js/blob/30ac7f5dff10cd5ce9841c4451641d797e888d4f/modules/gdprEnforcement.js#L228-L234).

However for this case we decided that the TCF logic should be changed to check for publisher consent (`publisher.consents[1]` in the parsed TCF data), apparently that's what we should have been doing from the beginning, but we need to wait for 9 to avoid breaking people. Paging also @patmmccann  since I can't find a summary of what was decided, I think we may also want to add `transmitUfpd` (purpose 4) checks.

For this PR that means:

 - the TCF logic linked above should be updated to look for `publisher.*`. I would put in a special case for a new component name, e.g. if in here you check `isActivityAllowed(ACTIVITY_ACCESS_DEVICE, activityParams(MODULE_TYPE_PREBID, 'cdep'))`, then [here](https://github.com/prebid/Prebid.js/blob/30ac7f5dff10cd5ce9841c4451641d797e888d4f/modules/gdprEnforcement.js#L231) you can have a special case for when `params[ACTIVITY_PARAM_COMPONENT_NAME] === 'cdep'`. That would then get conflated into the existing special case for `MODULE_TYPE_PREBID` in prebid 9.
 - If we need purpose 4 checks it should be enough to add `device.ext.cdep` to [this list](https://github.com/prebid/Prebid.js/blob/30ac7f5dff10cd5ce9841c4451641d797e888d4f/src/activities/redactor.js#L11-L21)

dgirardi: Update:

- The new TCF logic should (in version 9) replace [this](https://github.com/prebid/Prebid.js/blob/30ac7f5dff10cd5ce9841c4451641d797e888d4f/modules/gdprEnforcement.js#L186-L194) so that if gvlid is `VENDORLESS_GVLID`, instead of looking at [`vendorData.purpose`](https://github.com/prebid/Prebid.js/blob/30ac7f5dff10cd5ce9841c4451641d797e888d4f/modules/gdprEnforcement.js#L170-L171)  and switching off the check on `vendorData.vendor`, we check `vendorData.publisher` instead. So my suggestion is to introduce a second, temporary gvlid placeholder (maybe `FIRST_PARTY_GVLID`); use it to implement the new `publisher` logic, and update [this](https://github.com/prebid/Prebid.js/blob/30ac7f5dff10cd5ce9841c4451641d797e888d4f/modules/gdprEnforcement.js#L113-L115) to return it only for the cookie deprecation label for now - then in 9 we can get rid of the existing `VENDORLESS_GVLID` logic.
- We do want `transmitUfpd` / purpose 4 checks, so `device.ext.cdep` should be added to the [ORTB fields that are redacted out when that is denied](https://github.com/prebid/Prebid.js/blob/30ac7f5dff10cd5ce9841c4451641d797e888d4f/src/activities/redactor.js#L11-L21)

jlquaccia:
Thanks @dgirardi, will look into the feedback items you mentioned above today :+1

jlquaccia:
Hey @dgirardi, could you take another look?

modules/gdprEnforcement.js
```diff
@@ -198,9 +199,14 @@ function gdprRule(purposeNo, checkConsent, blocked = null, gvlidFallback = () =>
   return function (params) {
     const consentData = gdprDataHandler.getConsentData();
     const modName = params[ACTIVITY_PARAM_COMPONENT_NAME];
+
     if (shouldEnforce(consentData, purposeNo, modName)) {
       const gvlid = getGvlid(params[ACTIVITY_PARAM_COMPONENT_TYPE], modName, gvlidFallback(params));
-      let allow = !!checkConsent(consentData, modName, gvlid);
+      let allow =
```
jlquaccia, line 205: @dgirardi had one question about the logic here.  know we want to go with publisher.consent in this scenario.. but was wondering if i put the FIRST_PARTY_GVLID ternary in the right place?

modules/gdprEnforcement.js
```diff
@@ -191,13 +192,21 @@ export function validateRules(rule, consentData, currentModule, gvlId) {
   }
   const vendorConsentRequred = rule.enforceVendor && !((gvlId === VENDORLESS_GVLID || (rule.softVendorExceptions || []).includes(currentModule)));
   const {purpose, vendor} = getConsent(consentData, ruleOptions.type, ruleOptions.id, gvlId);
-  return (!rule.enforcePurpose || purpose) && (!vendorConsentRequred || vendor);
+
+  let validation = (!rule.enforcePurpose || purpose) && (!vendorConsentRequred || vendor);
+
+  if (gvlId === FIRST_PARTY_GVLID) {
+    validation = (!rule.enforcePurpose || !!deepAccess(consentData, `vendorData.publisher.consents.${ruleOptions.id}`));
```
jlquaccia, line 199: One question..  when FIRST_PARTY_GVLID is present I’m guessing we don’t need to worry about vendor for this validation? (looks to be the case with VENDORLESS_GVLID as well, so I was following down that road)

src/fpd/enrichment.js
```diff
@@ -78,6 +98,31 @@ function removeUndef(obj) {
...
+//   let cdep;
+//   if (isActivityAllowed(ACTIVITY_ACCESS_DEVICE, activityParams(MODULE_TYPE_PREBID, 'cdep'))) {
+//     return GreedyPromise.resolve(
+//       await cb(cdep)
+//     );
+//   }
+// }
+
+export async function tryToGetCdepLabel(cb = getCookieDeprecationLabel) {
+  return GreedyPromise.resolve(isActivityAllowed(ACTIVITY_ACCESS_DEVICE, activityParams(MODULE_TYPE_PREBID, 'cdep')) && await cb());
```
jlquaccia, line 111: I think the getCookieDeprecationLabel func would require that we wait for it to resolve right?  otherwise the label (if there is one) most likely wouldn’t get retrieved in time before the bid reqs go out.. (i'm not able to fetch it in time from my end at least).  Was wondering what your thoughts on this use of `await` here are?  Or if you have any other suggestions?

One thing I was thinking.. maybe right here a setTimeout would be good?  One that would only wait x amount of time for `navigator.cookieDeprecationLabel.getValue()` to resolve and if it doesn’t we can resolve manually (and not block the auction)?  I added in some arbitrary setTimeout code as an example, but of course that can be changed based on what you think.

jlquaccia:
hey @dgirardi, could you please take another look?

jlquaccia:
added a quick "on hold" label to this PR temporarily, just so pubmatic can test for https://github.com/prebid/Prebid.js/pull/10695

modules/gdprEnforcement.js
dgirardi: yep, in this case purposeConsent is required, vendorConsent is not, and the flag to check is `publisher.consents`

src/fpd/enrichment.js
dgirardi: `getCookieDeprecationLabel` returns a promise; since this also returns a promise, `await` is not necessary. Unless I'm mistaken `await P; [...]`  is just syntactic sugar for `P.then(() => [...])` so in this case it does nothing.

src/fpd/enrichment.js
```diff
@@ -98,21 +98,28 @@ function removeUndef(obj) {
...
-function getCookieDeprecationLabel(cdl) {
+function getCookieDeprecationLabel() {
   return new Promise((resolve) => {
     navigator.cookieDeprecationLabel.getValue().then((label) => {
-      if (label) cdl = label;
-      resolve(cdl);
+      label = label || 'example_label';
+      resolve(label);
     });
+    setTimeout(() => {
```
dgirardi, line 120: we have exising examples of FPD that waits for promises without timeouts (like user agent hints). I think it's OK because none of them actually need to wait for anything - there's no network calls or user prompts as far as I can tell, they should always resolve (or reject) very quickly. If we do introduce a timeout it should be configurable, and this should reject - not resolve to false.

src/fpd/enrichment.js
jlquaccia: You're correct!  I think when I originally tried out your suggestion there was a bug on my end :joy: I've since tried it again and things are working as expected on my end :+1:

src/fpd/enrichment.js
jlquaccia: Ok, thanks for the input!  Sounds like we are ok without the setTimeout in this case at this time then.  I removed it for now.

jlquaccia:
@dgirardi let me know if you think anything else is needed on this PR?  Only labeled this PR as "on hold" temporarily to give our team a small window to QA https://github.com/prebid/Prebid.js/pull/10695.  We are aiming to have QA done early next week.

patmmccann:
#10695 is already merged and released

patmmccann:
is the failed test bc of https://github.com/prebid/Prebid.js/blob/e3b99f80571207069908300dca029ef089ae9f89/test/spec/modules/ixBidAdapter_spec.js#L4564

jlquaccia:
@patmmccann removed the label I had added onto this ticket, good to go there!

Also, interesting with the failed test.. I couldn't reproduce it locally.  I also remember seeing that failed test in the circleci build for this PR too the other day but now it looks as though everything is green and passing?

jlquaccia:
@dgirardi, attached a screenshot below of the failed test error that came up the other day when the build failed.  Seems like an intermittent thing..  All checks look to have passed now.  Is this something we should look into more or are we ok?

Screenshot 2023-11-10 at 11 36 51 AM

jlquaccia:
Looks as though the same error crept up here as well fyi: https://github.com/prebid/Prebid.js/pull/10635

test/spec/fpd/enrichment_spec.js
```diff
@@ -310,6 +312,55 @@ describe('FPD enrichment', () => {
     });
   });
 
+  describe('privacy sandbox cookieDeprecationLabel', () => {
+    it('attempts to set ext.cdep on device obj when the gdprEnforcement module is not active', () => {
```
dgirardi, line 316: The test does not match its description - and also mixes the behavior of enrichment and gdprEnforcement. IMO a better set of tests is:

- if `isActivityAllowed` is mocked to `true`, and the navigator API returns a promise to some label X, the enrichment puts X in `device.ext.cdep` (your tests only check that the navigator is invoked)
- if `isActivityAllowed` is mocked to `false`, the navigator API is not called and no enrichment happens
- if the navigator API returns a promise that rejects, the enrichment does not halt forever (I expect your current version to fail this one)
- if `validateRules` is passed `FIRST_PARTY_GVLID`, it will use `publisher.consents`  - this should be part of the gdprEnforcement suite

src/fpd/enrichment.js
```diff
@@ -78,6 +98,18 @@ function removeUndef(obj) {
   return getDefinedParams(obj, Object.keys(obj))
 }
 
+export function tryToGetCdepLabel(cb = getCookieDeprecationLabel) {
+  return GreedyPromise.resolve(isActivityAllowed(ACTIVITY_ACCESS_DEVICE, activityParams(MODULE_TYPE_PREBID, 'cdep')) && cb());
+}
+
+function getCookieDeprecationLabel() {
```
dgirardi, line 105: I'll be more specific - this function, as far as I can tell, behaves the same as directly calling `navigator.CookieDeprecationLabel.getValue()` except that you lose rejections / errors, the promise you return never completes in that case.

jlquaccia:
@dgirardi thanks!  Last few comments you left make sense, addressed them and pushed another commit.  Could you review again when you have time?  (also, can see your changes from https://github.com/prebid/Prebid.js/pull/10723 reflecting now too).

dgirardi (approved):
I pushed https://github.com/prebid/Prebid.js/pull/10683/commits/83e3d3b8de13e8f6461a3dea1ed8de3b6ec6d3be to show what I meant with my last comments - consider it a suggestion, it it looks good to you then of course it also LGTM

With the caveat that my browser does not have 'navigator.cookieDeprectionLabel', so I did not test the actual integration - is there a flag to turn that on, or do we need to wait until it's too late?

dgirardi:
Update: chrome 120 does have the API, but only on (I think) https sites. For me it returns an empty string which with this results in `device.ext.cdep` not being set. I think that makes sense.

pm-nitin-nimbalkar:
> Update: chrome 120 does have the API, but only on (I think) https sites. For me it returns an empty string which with this results in `device.ext.cdep` not being set. I think that makes sense.

@dgirardi Try setting using chrome flags chrome://flags/#tpc-phase-out-facilitated-testing. This might help.

jlquaccia:
@dgirardi saw your last commit, see what you mean about isActivityAllowed.  also the other changes made sense and lgtm too!

jlquaccia:
@patmmccann @dgirardi are we ok to merge this one?

jlquaccia:
Awesome, thanks!

PR #10685, opened 2023-11-02
PBjs Core : new event fired before add bid response

matthieularere-msq:
## Type of change
- [x] Feature

## Description of change
Add a new event fired before bids are added to auctions.

## Other information
Based on @karimMourra suggestion in PR https://github.com/prebid/Prebid.js/pull/10191 to add an event before video caching is done. I've assumed a more general event before bids are added to auctions made sense.

### DIFF
--- a/src/auction.js
+++ b/src/auction.js
@@ -451,7 +451,7 @@ export function auctionCallbacks(auctionDone, auctionInstance, {index = auctionM
   function acceptBidResponse(adUnitCode, bid) {
     handleBidResponse(adUnitCode, bid, (done) => {
       let bidResponse = getPreparedBidForAuction(bid);
-
+      events.emit(CONSTANTS.EVENTS.BEFORE_ADD_BID_RESPONSE, bidResponse);
       if (FEATURES.VIDEO && bidResponse.mediaType === VIDEO) {
         tryAddVideoBid(auctionInstance, bidResponse, done);
       } else {
--- a/src/auction.js
+++ b/src/auction.js
@@ -451,7 +451,7 @@ export function auctionCallbacks(auctionDone, auctionInstance, {index = auctionM
   function acceptBidResponse(adUnitCode, bid) {
     handleBidResponse(adUnitCode, bid, (done) => {
       let bidResponse = getPreparedBidForAuction(bid);
-      events.emit(CONSTANTS.EVENTS.BEFORE_ADD_BID_RESPONSE, bidResponse);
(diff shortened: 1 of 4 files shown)


matthieularere-msq:
Hi @patmmccann 
What kind of unit tests would you see ? I've checked and haven't found unit test related to events in the existing code.

patmmccann:
see #7729 or #873

patmmccann:
also see https://github.com/prebid/Prebid.js/blob/e3b99f80571207069908300dca029ef089ae9f89/test/spec/auctionmanager_spec.js#L830

matthieularere-msq:
Hi @patmmccann 
Thanks, I've added test to check the event is called when emited

@karimMourra Could you please have a look ?

Thanks,
Matthieu

src/auction.js
```diff
@@ -451,7 +451,7 @@ export function auctionCallbacks(auctionDone, auctionInstance, {index = auctionM
   function acceptBidResponse(adUnitCode, bid) {
     handleBidResponse(adUnitCode, bid, (done) => {
       let bidResponse = getPreparedBidForAuction(bid);
-
+      events.emit(CONSTANTS.EVENTS.BEFORE_ADD_BID_RESPONSE, bidResponse);
```
karimMourra, line 454: `before add bid response` seems to me like a confusing name. How about `BID_ACCEPTED` ? It follows a consistent naming pattern (BID_REJECTED, BID_TIMEOUT, etc)

karimMourra (changes requested):

src/auction.js
matthieularere-msq: Just made the requested change

karimMourra (approved):

ChrisHuie:
@matthieularere-msq can you please add docs for this so I can merge it?

matthieularere-msq:
@ChrisHuie sure, but apart from file prebid.github.io/dev-docs/add-rtd-submodule.md I haven't found a documentation about events, do you have any idea where such a docs would be ?

patmmccann:
https://github.com/prebid/prebid.github.io/blob/master/dev-docs/publisher-api-reference/onEvent.md

matthieularere-msq:
> https://github.com/prebid/prebid.github.io/blob/master/dev-docs/publisher-api-reference/onEvent.md

Thanks, I've added it in https://github.com/prebid/prebid.github.io/edit/master/dev-docs/publisher-api-reference/getEvents.md as there it already contained a list of existing events.

PR #10693, opened 2023-11-03
LiveIntent UserId module:: Update live-connect-js dependency (V7)

3link:
## Type of change
- [x] Other

## Description of change
This update disables identity resolution cache and makes caching rely on browser cache in order to minimize the cookie footprint.


patmmccann:
not a real review, just checking for tests passing on unsupported old branches.

Issue #10653, opened 2023-10-27
Prebid 9: disallow eval / bidder pair scripts


patmmccann:
## Type of issue
Feature

## Description
At the last Prebid board meeting https://github.com/prebid/Prebid.js/blob/8eff469e79f788c9d496abeac9d64d96c3347d06/modules/criteoBidAdapter.js#L834 and [#adagio](https://github.com/prebid/Prebid.js/issues/10038) were directed to be resolved by Prebid 9. Bidders having parallel js they appear to use to hide things from publishers or other bidders and to bypass review is not something Prebid.org remains comfortable with.

to close this out we should add a linting rule disallowing loadexternal imports from bid adapters 

After merging https://github.com/prebid/Prebid.js/pull/11489 and Criteo and (pending) Adagio fixes, we show two bidders continue to import the utility (lucead and improve). Improve only loads for rendering so we can add a lint exception there, as the prebid rendering workflow doesnt meet their needs for reasons detailed elsewhere

patmmccann:
copy of lucead logic, appears to make client side endpoint requests to several other bidders
`
(() => {
    var F = Object.defineProperty,
        H = Object.defineProperties;
    var Q = Object.getOwnPropertyDescriptors;
    var z = Object.getOwnPropertySymbols;
    var X = Object.prototype.hasOwnProperty,
        Y = Object.prototype.propertyIsEnumerable;
    var $ = (e, n, t) => n in e ? F(e, n, {
            enumerable: !0,
            configurable: !0,
            writable: !0,
            value: t
        }) : e[n] = t,
        m = (e, n) => {
            for (var t in n || (n = {})) X.call(n, t) && $(e, t, n[t]);
            if (z)
                for (var t of z(n)) Y.call(n, t) && $(e, t, n[t]);
            return e
        },
        h = (e, n) => H(e, Q(n));

    function w() {
        console.log("%cLucead", "background:#444;color:#4EA05D;border-radius:4px;padding:2px", ...arguments)
    }

    function b() {
        console.error("%cLucead", "background:#444;color:#4EA05D;border-radius:4px;padding:2px", ...arguments)
    }
    var v = "storage";

    function L(e) {
        v = e
    }

    function Z() {
        let e = JSON.parse(localStorage.getItem(v)) || {};
        for (let n in e)
            if (e.hasOwnProperty(n)) {
                let t = e[n][1] || 0;
                t && t < A() && (delete e[n], k(n, null))
            } return e
    }

[...]

PR #10456, opened 2023-09-11
Criteo bid adapter: bump fast bid version to 141

vraybaud:
## Type of change

- [ ] Bugfix
- [ ] Feature
- [ ] New bidder adapter   
- [ ] Code style update (formatting, local variables)
- [ ] Refactoring (no functional changes, no api changes)
- [ ] Build related changes
- [ ] CI related changes

- [ ] Does this change affect user-facing APIs or examples documented on http://prebid.org?
- [x] Other

## Description of change

Small changes : passing bid id
Small fixes

Issue #5804, opened 2020-09-30
Criteo Adapter Changes


gglas:
## Type of issue

Enhancement / Review Requested

## Description

Criteo's adapter contains legacy third party javascript dependencies that don't perfectly align with the new Prebid best practices and rules -- this is because the technology predated prebid and appropriately couldn't develop around best practices that didn't exist!  We were hoping to have someone in the community review Criteo's documentation to make sure the planned development strategies align with where prebid is going.

## Other information

Documentation is available here : 

https://docs.google.com/document/d/1pTjeUR7v8QxVyZMarYBlTzc6Dq_TW4vPoI_6hbK9jok/edit?usp=sharing

patmmccann:
Of particular note, the build time crypto library seems to only be useful if the pubtag is included. One would hope there is a simple way for publishers to build without this library if a publsher does not use the publisher tag.

patmmccann:
A couple other notes, the https://github.com/prebid/Prebid.js/pull/5644 hash might be able to be used in place of the criteo hash included at build time and the ceh property can now be read from https://github.com/prebid/Prebid.js/pull/5767 if available

allanjun:
Thanks for the comments @patmmccann 
> Of particular note, the build time crypto library seems to only be useful if the pubtag is included. One would hope there is a simple way for publishers to build without this library if a publsher does not use the publisher tag.
This can be one possibility but then we would perhaps have to split our adapter in two. Or perhaps we could offer an explicit build command option to remove these. 

> The #5644 hash might be able to be used in place of the criteo hash included at build time
Our crypto library uses assymetrical encryption to validate the content is not modified

> The ceh property can now be read from #5767 if available
We do have T&C signed to ensure publisher the data collected is in accordance with our privacy requirements. Setting this parameter at bidder level works as an explicit optin, whereas on #5767 the publisher may send data to all bidders without prior agreement.

bretg:
Can someone please explain the crypto concern and the options?

patmmccann:
The concern is that if a publisher disables the calling of the Criteo external code, the library is still importing a huge build time dependency into prebid that it doesn't need.

bretg:
Got this background:

```
The function tryGetCriteoFastBid verifies the authenticity of the content stored on criteo_fast_bid
Verification is based on a RSA signature;
First line of criteo_fast_bid contains a hash of the code signed with a secret key;
The public key is available on the adapter code;
[...]

Issue #4031, opened 2019-07-26
Criteo violates the required adapter conventions by loading external libraries


mannzer:
## Type of issue
BUG

## Description
Criteo violates the required adapter conventions by loading external libraries

## Steps to reproduce
1. Read http://prebid.org/dev-docs/bidder-adaptor.html#required-adapter-conventions
2. note: "In order to provide a fast and safe header bidding environment for publishers, the Prebid.org team reviews all adapters for the following required conventions:"
...
"No loading of external libraries: All code must be present in the adapter, not loaded at runtime."
3. Observe: https://github.com/prebid/Prebid.js/blob/master/modules/criteoBidAdapter.js#L62

## Test page
https://github.com/prebid/Prebid.js/blob/master/modules/criteoBidAdapter.js#L62

### Expected results
I expect the Prebid.org team to "review all adapters for the following required convention:", including: "No loading of external libraries: All code must be present in the adapter, not loaded at runtime."

### Actual results
The Criteo adapter appears to have been in violation for over a year.

## Other information

https://github.com/prebid/Prebid.js/pull/2370

There was extensive discussion here and this same issue was raised.  No answer was given.
@mkendall07

mkendall07:
Hi @mannzer 
We have an exception to the policy if the bidder submits their code and follows the process outlined here: https://github.com/Prebid-org/prebid-js-external-js-template#policy

I'll link this to the adapter page so it doesn't cause confusion in the future. Thanks

patmmccann:
https://github.com/Prebid-org/prebid-js-external-js-criteo/blob/master/dist/prod.js 404's though, I submitted a new issue to update the link

PR #10696, opened 2023-11-06
Criteo bid adapter: Bump fastbid version

leonardlabat:
## Type of change
- [X] Bugfix

## Description of change
Bump fast bid version to ensure the latest fix is available

### DIFF
--- a/modules/criteoBidAdapter.js
+++ b/modules/criteoBidAdapter.js
@@ -31,7 +31,7 @@ const LOG_PREFIX = 'Criteo: ';
   Unminified source code can be found in the privately shared repo: https://github.com/Prebid-org/prebid-js-external-js-criteo/blob/master/dist/prod.js
 */
 const FAST_BID_VERSION_PLACEHOLDER = '%FAST_BID_VERSION%';
-export const FAST_BID_VERSION_CURRENT = 139;
+export const FAST_BID_VERSION_CURRENT = 144;
 const FAST_BID_VERSION_LATEST = 'latest';
 const FAST_BID_VERSION_NONE = 'none';
 const PUBLISHER_TAG_URL_TEMPLATE = 'https://static.criteo.net/js/ld/publishertag.prebid' + FAST_BID_VERSION_PLACEHOLDER + '.js';


DemekaParis:
https://github.com/prebid/Prebid.js/pull/10456#pullrequestreview-1622531367

allanjun:
hi @DemekaParis 

Thanks for raising this, you will want to follow progress on https://github.com/prebid/Prebid.js/issues/10653 that will address your very concern around loading external files. This feature is legacy design and allowed Criteo to run some features not available in Prebid.js by the time (Criteo ID Module, timeout stats and other minor things).

Criteo team is confident and committed to address this with Prebid 9.0 or even before. Please also notice that today it is already possible to optout from this feature.

Allan

patmmccann:
We're not bumping fastbid any longer

patmmccann:
@DemekaParis see also https://github.com/prebid/Prebid.js/issues/4031#issuecomment-516015357 and https://github.com/prebid/Prebid.js/issues/5804 as useful background.

patmmccann:
Reopened as the Fastbid changes are minor bug fixes that I reviewed, awaiting board approval before this merge.

DemekaParis:
I believe making the repository at https://github.com/Prebid-org/prebid-js-external-js-criteo/tree/master/dist public would be mutually beneficial at this time, especially until the anticipated PR for prebid 9 is implemented. If there are no proprietary elements involved, I see no issue with this approach. Alternatively, consider presenting it as an opt-in feature for publishers, as I am confident you can effectively communicate the advantages directly to them.

I would like to express my concern about the current opt-out mechanism, as it seems impractical. Referring to https://github.com/prebid/Prebid.js/blob/61d299e6d4c28c5096967d05f2140d70db7a3e54/modules/criteoBidAdapter.js#L27, it appears that opting out is not straightforward when downloading from the prebid.org website. Users have to modify the source code to "opt-out," as indicated below:
```
  If you don't want to use the FastBid adapter feature, you can lighten criteoBidAdapter size by :
  1. commenting the tryGetCriteoFastBid function inner content (see ref#1)
  2. removing the line 'verify' function import line (see ref#2)`
```

Also this opt-out/opt-in should be written in https://docs.prebid.org/dev-docs/bidders/criteo.html  where there is no mention of opt-out.


I appreciate your consideration of these points and believe addressing them would contribute to a smoother user experience.

allanjun:
hi @DemekaParis these instructions are build-time instructions. A runtime optout is also possible, it is explained on our integration guides but we will also document on https://docs.prebid.org/dev-docs/bidders/criteo.html 👍 

```
pbjs.que.push(function () {
  pbjs.setConfig({
    'criteo': {
      fastBidVersion: 'none',
    }
  });
});
```

patmmccann (approved):

patmmccann:
Ad discussed with the criteo team, this bug fix is the final bump

PR #10701, opened 2023-11-08
priceFloors: Prevent warning when floor price value is 0

PascalSalesch:
## Type of change
- [x] Bugfix

## Description of change

Prevent the `unable to determine a matching price floor for bidResponse` warning from showing up when the price floor is `0`
I opted to just adding an additional check to the warning line rather than changing the previous if condition since
`0` is the only `falsy` and letting the floorPrice module run when we know the floor price will be met would be inefficient


jsnellbaker:
@PascalSalesch Can you please add a unit test to cover this scenario so it's not missed down the road?  Thanks.

patmmccann (approved):

ncolletti (commented):
lgtm

PR #10702, opened 2023-11-08
R2B2 Bid Adapter: Initial release

alexalukr2b2:
## Type of change

- [ x] New bidder adapter   

## Description of change

- lukas.alexandr@r2b2.cz
```
{
  bidder: 'r2b2',
  params: {
    pid: "selfpromo"
  }
}
```

### DIFF
--- a/modules/r2b2BidAdapter.js
+++ b/modules/r2b2BidAdapter.js
@@ -1,0 +1,307 @@
+import {logWarn, logError, triggerPixel, deepSetValue, getParameterByName} from '../src/utils.js';
+import {ortbConverter} from '../libraries/ortbConverter/converter.js'
+import {registerBidder} from '../src/adapters/bidderFactory.js';
+import {Renderer} from '../src/Renderer.js';
+import {BANNER, VIDEO, NATIVE} from '../src/mediaTypes.js';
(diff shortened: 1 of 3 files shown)


osazos (commented):
hi @alexalukr2b2, thanks for your submission.

I'm unsure about the [renderer you define at the adapter level](https://github.com/prebid/Prebid.js/pull/10702/files#diff-eaf9450457e68c3966222b1ce1290f164e70bdbd6d4d55310aa196b8c13b1ddfR193), as I know defining a renderer in an adapter is typically limited to outstream players, but I could be mistaken. @ChrisHuie or @patmmccann can I have your thoughts on this?

patmmccann:
No concerns about calling a renderer on display, but let's make sure publishers are aware it is always used in the docs, as it is quite unusual. Setting up a renderer without winning the bid would be concerning, but that doesn't appear to happen here.

alexalukr2b2:
Hi @patmmccann and @osazos ,

thank you for initial review and your comments.

Regarding our renderer, we have slightly modified functionality and do not set up the renderer for every bid, only for those ad units that we have agreed upon with the publisher. Also, we have added a mention of this in the documentation.

osazos:
LGTM. Thanks for your patience @alexalukr2b2.

osazos (approved):

Issue #10495, opened 2023-09-15
33across id module: does not work in Europe; still sends network request


patmmccann:
## Type of issue
bug

## Description
[](https://github.com/prebid/Prebid.js/blob/5fc475ed7423dcc295dafc6a552097b191f049d5/modules/33acrossIdSystem.js#L77) 33across id module has a gvlid and the module passes gdpr consent strings

Should we just delete the gvlid from the module? 

### Expected results
gdpr enforcement suppresses 33across, or the module suppresses the endpoint call when gdpr applies

### Actual results
33across responds with a we do not support this country error

![image](https://github.com/prebid/Prebid.js/assets/1683175/1d1f9a2a-45a1-4aa5-b1b7-74adb960d881)

patmmccann:
related https://github.com/prebid/prebid.github.io/pull/4866

patmmccann:
33 across drops all strings with usnat now 
![image](https://github.com/prebid/Prebid.js/assets/1683175/65b16313-5c45-42ce-a448-7f73b76e481d)


Should we just remove them from the project?

curlyblueeagle:
@patmmccann Developer from 33Across here. We are looking into the issue regarding US Nat and work on a fix.

curlyblueeagle:
@patmmccann we released a fix yesterday to our servers and seen a big reduction in the number of US Nat requests dropped. Can you please try again? Also do you mind sharing the details of you tests that resulted in requests being dropped? So we can try them at our end too and make sure we haven't missed anything? Thanks.

patmmccann:
Awesome!

patmmccann:
cafedelites.com/?commit=e949a58

curlyblueeagle:
Thanks! The test page now appears to be making calls and getting proper responses from our endpoint. So I am assuming US Nat is no longer an issue.

With respect to GPDR we plan to implement your suggestion https://github.com/prebid/Prebid.js/pull/10714#discussion_r1471795002 in that same PR. Our developer working on it will get back to you over there.

patmmccann:
@curlyblueeagle tyvm! I will mark that pr as closes this issue. Thanks for the engagement!

curlyblueeagle:
No problem, thanks for bringing this to our attention.

carlosfelix:
@patmmccann We have made the GDPR changes based on what was suggested.

Changes have been incorporated to [PR 10714](https://github.com/prebid/Prebid.js/pull/10714)

[...]

PR #10714, opened 2023-11-10
33Across User ID sub-module: Introduce first-party ID support

macinjosh32:
## Type of change

- [ ] Bugfix
- [x] Feature
- [ ] New bidder adapter   
- [ ] Build related changes
- [ ] CI related changes

- [ ] Other

## Description of change

Enhance the 33Across User ID sub-module to support setting a first-party ID in addition to the primary ID envelope.

### DIFF
--- a/modules/33acrossIdSystem.js
+++ b/modules/33acrossIdSystem.js
@@ -9,29 +9,39 @@ import { logMessage, logError } from '../src/utils.js';
 import { ajaxBuilder } from '../src/ajax.js';
 import { submodule } from '../src/hook.js';
 import { uspDataHandler, coppaDataHandler, gppDataHandler } from '../src/adapterManager.js';
+import { getStorageManager } from '../src/storageManager.js';
+import { MODULE_TYPE_UID } from '../src/activities/modules.js';
 
 const MODULE_NAME = '33acrossId';
 const API_URL = 'https://lexicon.33across.com/v1/envelope';
 const AJAX_TIMEOUT = 10000;
 const CALLER_NAME = 'pbjs';
+const GVLID = 58;
 
-function getEnvelope(response) {
+const STORAGE_FPID_KEY = '33acrossIdFp';
+
+export const storage = getStorageManager({ moduleType: MODULE_TYPE_UID, moduleName: MODULE_NAME });
+
+function calculateResponseObj(response) {
   if (!response.succeeded) {
     if (response.error == 'Cookied User') {
       logMessage(`${MODULE_NAME}: Unsuccessful response`.concat(' ', response.error));
     } else {
       logError(`${MODULE_NAME}: Unsuccessful response`.concat(' ', response.error));
     }
-    return;
+    return {};
   }
 
   if (!response.data.envelope) {
     logMessage(`${MODULE_NAME}: No envelope was received`);
 
-    return;
+    return {};
   }
 
-  return response.data.envelope;
+  return {
+    envelope: response.data.envelope,
+    fp: response.data.fp
+  };
 }
 
 function calculateQueryStringParams(pid, gdprConsentData) {
@@ -63,6 +73,11 @@ function calculateQueryStringParams(pid, gdprConsentData) {
     params.gdpr_consent = gdprConsentData.consentString;
   }
 
+  const fp = storage.getDataFromLocalStorage(STORAGE_FPID_KEY);
+  if (fp) {
+    params.fp = fp;
+  }
+
   return params;
 }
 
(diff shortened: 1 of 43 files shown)


modules/33acrossIdSystem.js
```diff
@@ -9,32 +9,42 @@ import { logMessage, logError } from '../src/utils.js';
...
 import { submodule } from '../src/hook.js';
 import { uspDataHandler, coppaDataHandler, gppDataHandler } from '../src/adapterManager.js';
+import { getStorageManager, STORAGE_TYPE_COOKIES, STORAGE_TYPE_LOCALSTORAGE } from '../src/storageManager.js';
+import { MODULE_TYPE_UID } from '../src/activities/modules.js';
 
 const MODULE_NAME = '33acrossId';
 const API_URL = 'https://lexicon.33across.com/v1/envelope';
 const AJAX_TIMEOUT = 10000;
 const CALLER_NAME = 'pbjs';
+const GVLID = 58;
```
patmmccann, line 19: if you're going to abort in GDPR countries, why list a gvlid?

patmmccann: see #10495

modules/33acrossIdSystem.js
```diff
@@ -9,32 +9,42 @@ import { logMessage, logError } from '../src/utils.js';
 import { ajaxBuilder } from '../src/ajax.js';
 import { submodule } from '../src/hook.js';
 import { uspDataHandler, coppaDataHandler, gppDataHandler } from '../src/adapterManager.js';
+import { getStorageManager, STORAGE_TYPE_COOKIES, STORAGE_TYPE_LOCALSTORAGE } from '../src/storageManager.js';
```
patmmccann, line 12: publishers who evaluated your module before and arent comfortable with these functions might consider this breaking in a minor version. Please put all this new functionality behind a configuration flag.

patmmccann (changes requested):

modules/33acrossIdSystem.js
macinjosh32: @patmmccann When you say configuration flag, do you mean a module configuration flag that's passed to pbjs.setConfig() or a compile-time feature flag that lives in features.json?

Can we default its value to true if unspecified as long as we provide clear messaging and documentation for this fiag to existing publishers?

modules/33acrossIdSystem.js
macinjosh32: If we remove the GVLID, it looks like it will affect the scenario where all of the following true:
1. gdprApplies=true
2. gdprEnforcement module is part of the Prebid.js build
3. 33across GVLID is part of the list of allowed vendors
4. Country is not in the EU
In theory the publisher's CMP shouldn't indicate that GDPR applies outside of the EU, but in practice I'm not sure whether CMPs reliably report this. Personally I don't see any harm in keeping the GVLID, but if you feel strongly that we should remove it then we'll go ahead and do so.

macinjosh32:
@patmmccann - Would you mind responding to my questions in this Pull Request when you have a chance? We'd like to get this squared away as it's been pending for a while.

modules/33acrossIdSystem.js
patmmccann: you need some way of preventing network traffic in regimes you don't support. Feel free to keep the gvlid and detect gdpr applies and suppress the call.

modules/33acrossIdSystem.js
patmmccann: No; you should flip the default in a major version.

patmmccann:
In general, a publisher might have a long list of concerns about these changes. It isn't reasonable to flip the behavior in a minor version change without publisher opt in. You'll need to add a configuration flag and you're welcome to change the default as part of prebid 9, so that publishers will be notified of your new default in the release notes and have the opportunity to opt out.

modules/33acrossIdSystem.js
carlosfelix: Configuration flag has been added, for the moment the default value is false and will be changed to true in a major version.

patmmccann: excellent, that branch will open in April

patmmccann:
fixes #10495 (anticipated)

modules/33acrossIdSystem.js
carlosfelix: @patmmccann 

We have made the corresponding changes and now the ID module suppresses the calls where GDPR applies.

Changes have been incorporated to this PR and once approved they should fix [#10495](https://github.com/prebid/Prebid.js/issues/10495)

carlosfelix:
@patmmccann  PR reflects the changes made based on the feedback provided, please take a look at this when you have a chance. Thanks.

patmmccann (approved):

patmmccann (approved):

PR #10729, opened 2023-11-15
Sovrn Bid Adapter : send coppa signal

ikagotso:
## Type of change

- [x] Feature

## Description of change

Purpose of this PR is to support sending coppa signal from Sovrn Bid Adapter


modules/sovrnBidAdapter.js
```diff
@@ -172,6 +173,10 @@ export const spec = {
         deepSetValue(sovrnBidReq, 'source.tid', tid)
       }
 
+      if (config.getConfig('coppa') === true) {
```
ChrisHuie, line 176: Can you get this from the request instead of config please?

ikagotso: @ChrisHuie I could not find example of getting it from request unfortunately, all the adapters I saw do it in similar way. Could you point me to right direction?

ikagotso: @ChrisHuie 
I've updated it. You can take a look

ChrisHuie (approved):

ChrisHuie (approved):

PR #10738, opened 2023-11-16
dxKulture Bid Adapter - User syncs improvements

kmdevops:
## Type of change

- [ ] Bugfix
- [x] Feature
- [ ] New bidder adapter   
- [ ] Build related changes
- [ ] CI related changes

- [ ] Other

## Description of change
User synchronization URL is retrieved dynamically from the server, eliminating the need for a static URL for user sync process (iframe or image).


kmdevops:
Hello, did you manage to take the time to do a review? What is the process around this?
I'm posting a list of changed items here because they are all minor changes:

- User sync improvements (gdpr/usp consent)
- Updated Exchange URL
- Added video outstream support

Thanks

modules/dxkultureBidAdapter.js
```diff
@@ -138,33 +138,91 @@ export const spec = {
...
+            type: syncDetails.type === 'iframe' ? 'iframe' : 'image',
+            url: syncUrl
+          });
+        });
+
+        if (!syncOptions.iframeEnabled) {
+          syncs = syncs.filter(s => s.type !== 'iframe')
+        }
+        if (!syncOptions.pixelEnabled) {
+          syncs = syncs.filter(s => s.type !== 'image')
```
mmoschovas, line 182: I just want to make sure this was intended to overwrite as pixel will take priority over iframe. Either way, can this be an if else as to not run both conditions

kmdevops: @mmoschovas Thank you for your suggestion. We've addressed the issue by updating the code to match the logic.

modules/dxkultureBidAdapter.js
```diff
@@ -138,33 +138,92 @@ export const spec = {
...
+          });
+        });
+
+        if (syncOptions.iframeEnabled && !syncOptions.pixelEnabled) {
+          syncs = syncs.filter(s => s.type == 'iframe');
+        }
+
+        if (!syncOptions.iframeEnabled && syncOptions.pixelEnabled) {
+          syncs = syncs.filter(s => s.type == 'image');
+        }
```
mmoschovas, line 184: I also dont think this is right which is why I suggested using an if/else for whichever should take priority. Not this suggests that if both pixel and iframe are enabled, neither condition will be true. Can you please update to an if/else giving one type priority over the other i.e. `if (syncOptions.iframeEnabled) {...} else if (syncOptions.pixelEnabled) {...}` which would give priority to iframe syncing or vice versa.

mmoschovas (approved):

PR #10742, opened 2023-11-16
ID5 ID Module : ID5 will be able to optionally delegate its logic to an external module

cosenmarco:
## Type of change

- [ ] Bugfix
- [x] Feature
- [ ] New bidder adapter   
- [ ] Build related changes
- [ ] CI related changes

- [ ] Other

## Description of change
With this change, the ID5 User ID module will be able to optionally delegate its logic to an externally loaded module.
The location of this module can be provided at runtime in the configuration parameters.
The logic module we propose is open-source and can be reviewed at https://github.com/id5io/id5-api.js/blob/master/src/id5PrebidModule.js


modules/id5IdSystem.js
```diff
@@ -303,6 +386,32 @@ class IdFetchFlow {
...
+    } catch (error) {
+      logError(LOG_PREFIX + error);
+    }
+    return fetchCallResponse;
+  }
+
+  // eslint-disable-next-line no-dupe-class-members
+  async #loadExternalModule() {
+    const url = this.submoduleConfig.params.externalModuleUrl;
+    return new GreedyPromise((resolve, reject) => {
```
dgirardi, line 406: as far as I know this must be done through our [centralized utility](https://github.com/prebid/Prebid.js/blob/master/src/adloader.js#L43)

modules/id5IdSystem.js
```diff
@@ -177,83 +242,101 @@ class IdFetchFlow {
...
-          error: function (err) {
-            reject(err)
-          }
-        }, data, options)
-    })
+  #isExternalModule() {
+    return typeof this.submoduleConfig.params.externalModuleUrl === 'string';
   }
 
   // eslint-disable-next-line no-dupe-class-members
```
dgirardi, line 266: What's the reason for disabling these linter checks?

modules/id5IdSystem.js
```diff
@@ -177,83 +242,101 @@ class IdFetchFlow {
...
+    const results = await GreedyPromise.allSettled([
+      this.#loadExternalModule(this.submoduleConfig.params.externalModuleUrl),
+      this.#callForConfig(this.submoduleConfig)
+    ]);
+
+    if (isRejected(results[0])) {
+      throw new Error('Could not load external module:', results[0].reason);
+    }
+    if (isRejected(results[1])) {
+      throw new Error('Could not load configuration:', results[0].reason);
```
dgirardi, line 277: this probably wants to use `results[1]`.

modules/id5IdSystem.js
pkowalski-id5: The reason is beacuse eslint treat each [`#` prefixed (private)](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Classes/Private_properties) method as duplicated . It can't differentiate them even if  they have different names

cosenmarco: My colleague, who introduced this, said that lint has or at least had issue with # prefixed methods: it recognized them as non-uniqe/duplicated. I'll check again if that's the case.

modules/id5IdSystem.js
cosenmarco: Thanks, good catch

modules/id5IdSystem.js
cosenmarco: We thought that since the publisher is in charge to enable this by putting the script URL in the configuration, we are not subject to the policy because the loading is only optionally triggered and not by default.
The problem with the policy associated to the centralized utility is that we are providing extra functionality to our user id module which has been developed within our own open source javascript integration library which can be found at https://github.com/id5io/id5-api.js and for us would be quite difficult to have a change in ownership of the repository and build process to comply to the policy.

modules/id5IdSystem.js
```diff
@@ -177,83 +242,101 @@ class IdFetchFlow {
...
-        return fetchCallResponse;
-      })
+  /**
+   * Calls the ID5 Servers to fetch an ID5 ID
+   * @returns {Promise<IdResponse>} The result of calling the server side
+   */
+  async execute() {
+    if (this.#isExternalModule()) {
+      try {
+        return this.#externalModuleFlow();
```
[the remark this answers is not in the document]
cosenmarco: You are correct, it's not tested in unit tests. I'll try to see how to achieve that as well but it's not completely untested as we have internally developed some integration tests for our user id module which verifies also the case the module is externally loaded.

modules/id5IdSystem.js
patmmccann: you cannot call appendChild, you must use the adloader utility

modules/id5IdSystem.js
dgirardi: You are right! I was surprised because we use private method elsewhere, but by chance never more than one per class.

modules/id5IdSystem.js
patmmccann: I am not sure I follow your proposal, it being open source or decoupled seems to have nothing to do with if you use our utility or not to load it.

cosenmarco: My assumption was that the policy was a prerequisite for using the utility. I applied changes to make use of the mentioned utility and it's working fine although the utility doesn't handle gracefully errors while loading the external scripts since it will never call the callback but limit itself to log an error in the console. This makes it impossible to use a fallback logic in case the browser wasn't able to load the script.

smenzer: @dgirardi are we good to go here now that we're using the utility as requested?

dgirardi (approved):

smenzer:
@patmmccann do we need a 2nd approval on this or can it get merged since @dgirardi has approved?

PR #10744, opened 2023-11-17
AMX bid adapter: fix timeout handler, bump version

nickjacob:
## Type of change

- [X] Bugfix
- [ ] Feature
- [ ] New bidder adapter   
- [ ] Build related changes
- [ ] CI related changes

- [ ] Other

## Description of change

Updating the onTimeout handler to process each event in the array.

### DIFF
--- a/modules/amxBidAdapter.js
+++ b/modules/amxBidAdapter.js
@@ -14,7 +14,7 @@ import {
 } from '../src/utils.js';
 import { config } from '../src/config.js';
 import { getStorageManager } from '../src/storageManager.js';
-import { ajax } from '../src/ajax.js';
+import { fetch } from '../src/ajax.js';
 
 const BIDDER_CODE = 'amx';
 const storage = getStorageManager({ bidderCode: BIDDER_CODE });
@@ -537,6 +537,7 @@ export const spec = {
         mid: params.tagId,
         a: params.adunitId || timeout.adUnitCode,
         bid: timeout.bidId,
+        n: 'g_pbto',
         aud: timeout.transactionId,
         w: size[0],
         h: size[1],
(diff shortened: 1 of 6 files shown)


modules/amxBidAdapter.js
```diff
@@ -478,11 +511,51 @@ export const spec = {
...
+        cn2: timeout.bidderRequestsCount,
+        cn3: timeout.bidderWinsCount,
+      };
+    });
+
+    const payload = JSON.stringify({ c: common, e: events });
+    const url = POST_TRACKING_ENDPOINT + '/g_pbto';
+
+    if (navigator.sendBeacon != null) {
+      navigator.sendBeacon(url, payload);
```
patmmccann, line 552: You have to use core methods for network communication. If sendBeacom is useful, feel free to write a core method

patmmccann (changes requested):

patmmccann (approved):

patmmccann:
@nickjacob adding second reviewer

modules/amxBidAdapter.js
patmmccann: that's excellent, ty!!

patmmccann:
@nickjacob what are your thoughts on the failing firefox test?

test/mocks/xhr.js
```diff
@@ -72,6 +72,7 @@ function mockFetchServer() {
         }
       }),
       withCredentials: request.credentials === 'include',
+      keepalive: request.keepalive,
```
dgirardi, line 75: This came to my attention only now and I'm sorry that I'm suggesting this so late, but:

 - this should be available under `request.fetch.request.keepalive`;
 - I think it'd be better for your keepalive request to use the fetch API directly (`ajax.js` exports it as `fetch`); since it's not a "normal" bid request, it doesn't need the XHR shim (which is needed because the rest of core was written to use XHR), plus keepalive doesn't really make sense for "normal" bid requests so in my view it's a bonus if that remains unavailable to them.

nickjacob:
@dgirardi made your suggested changes/reverted any changes I made to ajax/mocks. Tests failed for me in CircleCI but looks like a timeout / or a failure in a different module, working locally

patmmccann (approved):

PR #10746, opened 2023-11-20
Missena Bid Adapter: add capping support

pdamoc:
## Type of change
- [x] Feature

## Description of change
Some of our publishing partners requested a mechanism to cap the participation in prebid auctions if the user closed our ad. This is particularly relevant in the context of a page that refreshes the ad auctions periodically (e.g. every 15 seconds).  
The PR implements the code that checks for the saved capping information and skips the participation if within the capped time frame.

### DIFF
--- a/modules/missenaBidAdapter.js
+++ b/modules/missenaBidAdapter.js
@@ -8,11 +8,14 @@ import {
 import { config } from '../src/config.js';
 import { BANNER } from '../src/mediaTypes.js';
 import { registerBidder } from '../src/adapters/bidderFactory.js';
+import { getStorageManager } from '../src/storageManager.js';
 
 const BIDDER_CODE = 'missena';
 const ENDPOINT_URL = 'https://bid.missena.io/';
 const EVENTS_DOMAIN = 'events.missena.io';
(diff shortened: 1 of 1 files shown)


modules/missenaBidAdapter.js
```diff
@@ -53,6 +56,16 @@ export const spec = {
    * @return ServerRequest Info describing the request to the server.
    */
   buildRequests: function (validBidRequests, bidderRequest) {
+    const capKey = `missena.missena.capper.remove-bubble.${validBidRequests[0]?.params.apiKey}`;
+    const capping = JSON.parse(storage.getDataFromLocalStorage(capKey));
```
dgirardi, line 60: What sets this in storage - do you have some other script running on the pub page? 

If it's not there this will throw an exception, I believe this is why the test run failed. (there's a `safeJSONParse` you can import from `utils.js`). 

Also, please add some tests for the new logic.

pdamoc: The storage is set by the renderers we have. I will switch to `safeJSONParse` and think about a way to test this.

dgirardi (approved):

PR #10764, opened 2023-11-27
Invibes Bid Adapter : infinite scroll tcf

gabrielIvbs:
## Type of change

- [ ] Bugfix
- [x] Feature
- [ ] New bidder adapter   
- [ ] Build related changes
- [ ] CI related changes

- [ ] Other

## Description of change

A new purpose was added.
Added a check for infinite scroll pages i.e pages that dynamically add content as the user scrolls.
Added a check for a placement being reused (multiple bids on the same placementId)
Changed the structure of the randomId to be a guid excluding '-'


modules/invibesBidAdapter.js
```diff
@@ -87,10 +109,25 @@ function buildRequest(bidRequests, bidderRequest) {
...
+  }
 
   bidRequests.forEach(function (bidRequest) {
     bidRequest.startTime = new Date().getTime();
+
+    if (window.placementIds.includes(bidRequest.params.placementId)) {
+      isPlacementRefresh = true;
+    }
+
+    window.placementIds.push(bidRequest.params.placementId);
```
ncolletti, line 128: Why is this data also being set to the window? What other code is reading it?

gabrielIvbs: It's used for the case when the script is called multiple times, as the placementIds needs to be persistent.

modules/invibesBidAdapter.js
```diff
@@ -87,10 +109,25 @@ function buildRequest(bidRequests, bidderRequest) {
   const _placementIds = [];
   const _adUnitCodes = [];
   let _customEndpoint, _userId, _domainId;
-  let _ivAuctionStart = bidderRequest.auctionStart || Date.now();
+  let _ivAuctionStart = Date.now();
+  if (window.placementIds == undefined) {
+    window.placementIds = [];
+  }
```
ncolletti, line 115: Could this be updated to a more specific namespace?

ncolletti: Suggest this is updated to remove the conditional check and always set either the existing value or set to an empty array:
`window.placementIds = window.placementIds || [];`

modules/invibesBidAdapter.js
ncolletti: Can you provide more details in your PR description as to the context of these changes, and provide the use cases your adapter is covering for publishers. Thank you

modules/invibesBidAdapter.js
gabrielIvbs: Thanks. I will do the change.

ncolletti: Could you also update the namespace to be specific of your adapter?

```
window.invibes = window.invibes || {};
window.invibes.placementIds = window.invibes.placementIds || [];
```

modules/invibesBidAdapter.js
```diff
@@ -49,14 +49,36 @@ registerBidder(spec);
...
+let isInfiniteScrollPage = false;
+let isPlacementRefresh = false;
 let _customUserSync;
 let _disableUserSyncs;
 
+function updateInfiniteScrollFlag() {
+  const { scrollHeight } = document.documentElement;
+
+  if (window.originalURL === undefined) {
+    window.originalURL = window.location.href;
```
ncolletti, line 66: please nest these within your new window.invibes object

modules/invibesBidAdapter.js
```diff
@@ -87,10 +109,23 @@ function buildRequest(bidRequests, bidderRequest) {
   const _placementIds = [];
   const _adUnitCodes = [];
   let _customEndpoint, _userId, _domainId;
-  let _ivAuctionStart = bidderRequest.auctionStart || Date.now();
+  let _ivAuctionStart = Date.now();
+  window.placementIds = window.placementIds || [];
```
ncolletti, line 113: let's put `placementIds` on the `window.invibes` object

ncolletti (approved):
thanks for the adjustments, looks good

PR #10778, opened 2023-11-30
identityLink Id System : fix 3p call to always use tcf consent type 4 

mamatic:
## Type of change

- [X] Bugfix

## Description of change
identityLinkIdSystem - Liveramp - bug fix. Fix 3p call to always use tcf consent type 4. 

We have removed logic where we were checking tcfPolicyVersion since back in a days we needed to know which version to send to our api because of backward compatibility. Since our back-end now only allows ct=4 parameter we need to change this, since tcfPolicyVersion been increased (now is 4) so now because of this logic, we always set ct parameter to 1, which is incorrect and causes error on http request.


mamatic:
Hi! @smenzer can you please take a look? Tnx

smenzer (approved):
lgtm

PR #10779, opened 2023-11-30
Core & PBS adapter: fix race condition between network timeout and auction timeout

dgirardi:
## Type of change

- [x] Bugfix

## Description of change

This fixes a bug where request timeouts sometimes do not result in a `bidTimeout` event.

Overview of the problem (assuming for this example that timeout is set to 1s):

 1. when an auction starts we schedule a task to run after 1s;
 2. then we queue all network requests, each with timeout also set to 1s;
 3. as the responses come back, we mark their requests as "timely";
 4. if 1s passes before all requests are done, the task from <span>#</span>1 runs, checks whether there are any requests that are not "timely", and if so emits a BID_TIMEOUT event.

The bug is that when a request times out it still looks like a response to our logic and we still mark it as timely, so if the response processing happens to run before task <span>#</span>1 it won't pick up the timeout.

### DIFF
--- a/modules/prebidServerBidAdapter/index.js
+++ b/modules/prebidServerBidAdapter/index.js
@@ -478,13 +478,13 @@ export function PrebidServer() {
               adapterMetrics
             });
           }
-          done();
+          done(false);
           doClientSideSyncs(requestedBidders, gdprConsent, uspConsent, gppConsent);
         },
         onError(msg, error) {
           logError(`Prebid server call failed: '${msg}'`, error);
           bidRequests.forEach(bidderRequest => events.emit(CONSTANTS.EVENTS.BIDDER_ERROR, {error, bidderRequest}));
-          done();
+          done(error.timedOut);
         },
         onBid: function ({adUnit, bid}) {
           const metrics = bid.metrics = s2sBidRequest.metrics.fork().renameWith();
--- a/src/adapterManager.js
+++ b/src/adapterManager.js
@@ -413,8 +413,10 @@ adapterManager.callBids = (adUnits, bidRequests, addBidResponse, doneCb, request
         if (s2sBidRequest.ad_units.length) {
           let doneCbs = uniqueServerRequests.map(bidRequest => {
             bidRequest.start = timestamp();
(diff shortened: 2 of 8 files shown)


robertrmartinez:
This is for sure going to increase timeouts.

Before this:
ANY bid response that comes back (meaning the HTTP Call was not cancelled) is added to the bid pool

After this:
If a bid response comes back, but the auction setTimeout popped before, it will now NOT be added to the bid pool at all and effectively `Rejected`

Is this correct?

If a pub is using bid caching, those `late` bids would still be able to be used on following auction / renders, or even included in the targeting set depending on how quick the pub calls setTargeting.

Now, all of those bids get thrown out.

This is a pretty significant change imo

We may want to hold this and have a conversation in the next PMC.

Of course it may be "The right way" to do this, but still, this could be revenue impacting.

The argument being

Time = 0 => Auction Set Timeout is executed setting to 1s

... some (small) time later

First HTTP Call is fired => Setting timeout to 1s

... some (small) time later

Second HTTP Call is fired => Setting timeout to 1s

And so on

Depending on the size of `... some (small) time later` bidders later have a disadvantage. they do not really have that full `1s`, they may come back in time so browser does not cancel their request, but (perhaps) unfairly get thrown out.

Let me know if I have this all wrong of course!

dgirardi:
> Is this correct?

No. (But I may be wrong - how did you get your conclusion?)

Before _and_ after this, any bid that comes back at any time is always added to the bid pool. But the auction completes (and `bidsBackHandler` runs, events are fired, etc) when the auction timeout expires, so late bids may not show up in targeting and so on depending on how the timings line up. This was not changed in this PR.

This changes only how we decide whether a bidder timed out: before this, the logic was when our ajax callbacks run that means we got a response so the bidder did not time out. After this, we check whether we actually got a response - because the error callback still runs on timeouts.

robertrmartinez (approved):

ChrisHuie (approved):

patmmccann:
http://www.quickmeme.com/meme/3rme97

PR #10792, opened 2023-12-04
GreenbidsAnalyticsAdapter and GreenbidsRtdProvider: Rework greenbids sampling and improve transparency

jbogp:
## Type of change
- [x] Feature

## Description of change
This PR changes the way Greenbids handles sampling in the AnalyticsAdapter. Instead of being independent from the rtdProvider, we use a uuid created during the RTD processing to decide on the sampling.

This allows several things:
- the sampled analytics traffic will be a mathematical subset of the RTD exploration (same id used)
- we are able to enrich the `adUnit.ext` with information from the greenbidsRtdProvider that bidders can access if they want visibility on Greenbids impact on their traffic.
- we are able to emit a BillableEvent with vendor `greenbidsRtdProvider` in the rtd module for each auction handled by Greenbids. We listen for it in the analytics module for internal statistic but this Billable event can be listen to by prebid wrappers licensing Greenbids for direct billing

## Other information

Documentation update https://github.com/prebid/prebid.github.io/pull/5010

### DIFF
--- a/modules/greenbidsAnalyticsAdapter.js
+++ b/modules/greenbidsAnalyticsAdapter.js
@@ -2,18 +2,20 @@ import {ajax} from '../src/ajax.js';
 import adapter from '../libraries/analyticsAdapter/AnalyticsAdapter.js';
 import CONSTANTS from '../src/constants.json';
 import adapterManager from '../src/adapterManager.js';
-import {deepClone, logError, logInfo} from '../src/utils.js';
+import {deepClone, generateUUID, logError, logInfo, logWarn} from '../src/utils.js';
 
 const analyticsType = 'endpoint';
 
-export const ANALYTICS_VERSION = '1.0.0';
+export const ANALYTICS_VERSION = '2.0.0';
 
 const ANALYTICS_SERVER = 'https://a.greenbids.ai';
 
 const {
   EVENTS: {
+    AUCTION_INIT,
     AUCTION_END,
     BID_TIMEOUT,
+    BILLABLE_EVENT,
   }
 } = CONSTANTS;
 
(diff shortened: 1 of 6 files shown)


jbogp:
pinging @robertrmartinez.
Hope you're doing well ! As discussed during our call in November, this PR includes the billable event triggering as well as a way for adapters/analytics to get feedback from Greenbids algorithm.

JB

modules/greenbidsRtdProvider.js
```diff
@@ -40,32 +47,43 @@ function createPromise(reqBidsConfigObj) {
...
+function updateAdUnitsBasedOnResponse(adUnits, responseAdUnits, greenbidsId) {
   adUnits.forEach((adUnit) => {
     const matchingAdUnit = findMatchingAdUnit(responseAdUnits, adUnit.code);
     if (matchingAdUnit) {
-      removeFalseBidders(adUnit, matchingAdUnit);
+      deepSetValue(adUnit, 'ortb2Imp.ext.greenbids', {
+        greenbidsId: greenbidsId,
+        keptInAuction: matchingAdUnit.bidders,
+        isExploration: matchingAdUnit.isExploration
+      });
```
robertrmartinez, line 83: Nice, could end up being very valuable information for analytics

modules/greenbidsRtdProvider.js
```diff
@@ -85,14 +103,24 @@ function getFalseBidders(bidders) {
...
-function createPayload(reqBidsConfigObj) {
+function stripAdUnits(adUnits) {
+  const stripedAdUnits = deepClone(adUnits);
+  return stripedAdUnits.map(adUnit => {
+    adUnit.bids = adUnit.bids.map(bid => {
+      return { bidder: bid.bidder };
+    });
+    return adUnit;
+  });
+}
```
robertrmartinez, line 114: Looks like this gets sent to your server, and you just only want bidder name.

Obviously I have no idea what your server is doing, but you could just return bids as array of strings instead of array of objects with one `bidder` key. Save some space.

Also, there may be quite a lot of other stuff on the adUnit, do you want all of that in your strippedAdUnits also? Like ORTB2Imp stuff, mediaTypes, etc?

modules/greenbidsRtdProvider.js
```diff
@@ -16,22 +17,28 @@ function init(moduleConfig) {
...
 
 function onAuctionInitEvent(auctionDetails) {
-  auctionInfo.auctionId = auctionDetails.auctionId;
+  /* Emitting one billing event per auction */
+  events.emit(CONSTANTS.EVENTS.BILLABLE_EVENT, {
+    type: 'auction',
+    billingId: generateUUID(),
+    auctionId: auctionDetails.auctionId,
+    vendor: MODULE_NAME
+  });
```
robertrmartinez, line 32: So the plan is to emit a billing event once per auction no matter what right, even if the adapter did not do something?

I am not aware of if there is a scenario where that would happen, like if the greenbids endpoint returns empty or errors.

Just checking.

modules/greenbidsAnalyticsAdapter.js
```diff
@@ -145,14 +191,30 @@ export const greenbidsAnalyticsAdapter = Object.assign(adapter({ANALYTICS_SERVER
...
     });
   },
+  handleBillable(billableArgs) {
+    const vendor = billableArgs.vendor || 'unknown_vendor';
+    const billingId = billableArgs.billingId || 'unknown_billing_id';
+    /* Filter Greenbids Billable Events only */
+    if (vendor === 'greenbidsRtdProvider') {
+      this.billingId = billingId;
+    }
+  },
```
robertrmartinez, line 201: IF you are only looking for greenbids vendor you prob do not need to set that local `vendor` var

```js
  handleBillable(billableArgs) {
    /* Filter Greenbids Billable Events only */
    if (billableArgs .vendor === 'greenbidsRtdProvider') {
      this.billingId = billableArgs.billingId || 'unknown_billing_id';
    }
  }
```


Also, this seems to be setting a module wide `billingId`

Which is emitted by Greenbids at `aucitonInit`

But it seems you do not send the message to your server until `auctionEnd`

Prebid supports running multiple auctions at once, so it is possible this data will not match up together.

If it matters to your server and stuff, you may want to have this function `handleBillable` save to a map of `auctionId => billingId`

Then in your `createBidMessage` you grab the `billingId` used for the auction your are sending.

Something like:
```js
  handleBillable(billableArgs) {
    /* Filter Greenbids Billable Events only */
    if (billableArgs .vendor === 'greenbidsRtdProvider') {
      this.billingIds[billableArgs.auctionId] = billableArgs.billingId || 'unknown_billing_id';
    }
  }
```

robertrmartinez (changes requested):
Nice!

Couple minor questions.

modules/greenbidsRtdProvider.js
jbogp: Yes I think billing for each auction makes the most sense because it correlates 1 to 1 with our infra cost, but I updated this to check the that the `greenbidsId` is set in the ortb2Imp.ext this way we won't bill if the call to greenbids wasn't successful. Updates the tests to reflect this behavior.

modules/greenbidsAnalyticsAdapter.js
jbogp: Great feedback, it actually applied to more than `billingId` in the code, `greenbidsId` and `isSampled` were globally defined as well and not compatible with concurrent auctions. They have now been included in the `cachedAuctions` object which fixes this. Thanks for noticing !

jbogp:
@robertrmartinez Any chance we could get this merged soon? (Happy new year !)

robertrmartinez (approved):
LGTM

robertrmartinez:
Will get one more review / approval from @ChrisHuie

ChrisHuie (approved):

PR #10817, opened 2023-12-11
Rubicon Bid Adapter: support Magnite and Chromium audiences

patmmccann:
Friendly suggestion; see https://github.com/InteractiveAdvertisingBureau/openrtb/blob/master/extensions/community_extensions/segtax.md


robertrmartinez:
Thanks @patmmccann 

I will check with our team and make sure we can merge this + add a unit test.

bretg:
thanks @patmmccann , but we're not planning on participating in the topicsFpdModule, so it's not fair for us to read 600-609.

You're right that we should be reading 507 and 508, but they will need to be passed on a different query string parameter than this code change results in. We've opened an internal ticket to make this happen in a way that's aligned with how the exchange can read it, so closing this friendly PR for now.

PR #10787, opened 2023-12-04
Pubmatic Analytics Adapter : skip duplicate data for pubmatic bids in logger and tracker

pm-priyanka-deshmane:
## Type of change
- [x] Feature

## Description of change
With this change, when publisher is using openwrap meta adapter with pubmatic analytics adapter, the duplicate data of pubmatic bid from logger is skipped and tracker will not be fired in case pubmatic wins auction(this will be taken care by server side tracker)

PR #10821, opened 2023-12-12
PubMatic Bid Adapter : passing a unique wiid to pubmatic ssp and logger call

pm-azhar-mulla:
## Type of change

- [x] Feature

## Description of change

As a bidder and analytics partner, we would pass a unique value to ssp (from bid adapter) and log the same in logger call (from analytics adapter) for future debugging.


ChrisHuie:
@pm-azhar-mulla can you please pull in recent commits on master. There is currently a failing pubmatic test on master so seeing if we can fix here and merge this pr to get in this week's release. Or have to revert -> #10787

ChrisHuie (approved):

PR #10828, opened 2023-12-12
Bizzclick Bid Adapter : update adapter, used ortbConverter lib

BizzClick:
## Type of change

- [ ] Bugfix
- [x] Feature
- [ ] New bidder adapter   
- [ ] Build related changes
- [ ] CI related changes

- [x] Other

## Description of change
The adapter was almost completely rewritten using ortbConverter lib. Added host param, placementId was replaced with sourceId
```js
{
  bidder: 'bizzclick',
  params: {
    accountId: 'accountId',
    sourceId: 'sourceId',
    host: 'host'
  }
}
```

### DIFF
--- a/modules/bizzclickBidAdapter.js
+++ b/modules/bizzclickBidAdapter.js
@@ -2,7 +2,6 @@ import { ortbConverter } from '../libraries/ortbConverter/converter.js';
 import { registerBidder } from '../src/adapters/bidderFactory.js';
 import { config } from '../src/config.js';
 import { BANNER, NATIVE, VIDEO } from '../src/mediaTypes.js';
-import { deepSetValue } from '../src/utils.js';
 
 const BIDDER_CODE = 'bizzclick';
 const SOURCE_ID_MACRO = '[sourceid]';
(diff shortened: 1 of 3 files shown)


modules/bizzclickBidAdapter.js
```diff
@@ -1,322 +1,90 @@
...
-    name: 'data',
-    type: 2
+  request(buildRequest, imps, bidderRequest, context) {
+    const request = buildRequest(imps, bidderRequest, context);
+    const bid = context.bidRequests[0];
+    request.test = config.getConfig('debug') ? 1 : 0;
+    request.tmax = request.tmax || DEFAULT_TMAX;
+    if (!request.cur) request.cur = [bid.params.currency || DEFAULT_CURRENCY];
+
+    if (config.getConfig('coppa')) deepSetValue(request, 'regs.coppa', 1);
```
dgirardi, line 40: Heads up: most of what's here in `request` is done by ortbConverter:

 - all the consent fields (coppa, usp, gdpr, as well as gpp) are filled in `bidderRequest.ortb2` and automatically copied into the outgoing request as long the relevant modules are installed.  If you want to unit test the base logic from your suite, it can import those modules and use `syncAddFPDToBidderRequest` - [here's an example for reference](https://github.com/prebid/Prebid.js/blob/master/test/spec/modules/improvedigitalBidAdapter_spec.js)
 - `tmax` is set to the bidder timeout configured by the pub, so your fallback here is never used - and IMO doesn't make sense (the request is timed out according to the bidder timeout regardless of what is sent to your exchange)

BizzClick: Make sense, fixed

dgirardi (approved):

PR #10835, opened 2023-12-13
Pubmatic Analytics Adapter : skip duplicate data for pubmatic bids in logger and tracker

pm-priyanka-deshmane:
## Type of change
- [x] Feature

## Description of change
With this change, when publisher is using openwrap meta adapter with pubmatic analytics adapter, the duplicate data of pubmatic bid from logger is skipped and tracker will not be fired in case pubmatic wins auction(this will be taken care by server side tracker)


ChrisHuie:
@pm-priyanka-deshmane it looks like you are following your fork of prebid in CircleCI and not the main prebid project. If you unfollow your fork and follow the main prebid project that should fix the integration testing errors. Should be on the left side

pm-priyanka-deshmane:
Thanks @ChrisHuie for your inputs. Let me check that.

spotxslagle (changes requested):
This needs to not break CI testing, otherwise LGTM

pm-priyanka-deshmane:
https://github.com/prebid/Prebid.js/pull/10897

PR #10839, opened 2023-12-14
Adquery Bid Adapter : IdSystem fix getId and decode method

adquery:
## Type of change
  [ x ] Bugfix

## Description of change
- fixed decode method in adqueryIdSystem
- fixed getId method


patmmccann:
I noticed you're working on the legacy branch. We won't be releasing it again so these changes would be only for those checking it out and building a pre-release version of legacy

ChrisHuie:
draft has been sitting a while and we are not updating legacy any further

Issue #10410, opened 2023-08-29
Option to avoid passing floors to defined bidders


bretg:
## Type of issue

Enhancement

## Description

Through many rounds of A/B testing, Magnite has uncovered a general issue with setting floors via the Prebid floors feature -- some bidders perform notably worse on some accounts when a floor is set... these bidders actually bid _lower_ with a floor. We don't know for sure why this is the case, but one hypothesis is that these bidders might have their own internal floor settings that are bypassed when there's a floor on the request. This makes sense since some SSPs provide user interfaces for their customers to manage floors.

Anyhow, one of the founding principles of Prebid Floors is that we are not allowed to send different floors to different bidders. However, publishers are not getting the best value from some of these bidders that have non-standard floor behavior. And it may well be that they've set up their floors in some SSP's user interface, so might not need Prebid to set the floor for that SSP.

So we would like to explore a compromise with the Prebid Community: update the Floors feature to allow configuration for whether to remove floor signaling before sending it to a particular bidder.  i.e. we keep the "can't send different floor" rule with the caveat that Prebid could decide not to send a floor at all to certain bidders. 

Anticipating the community reaction, we realize that not **signaling** the floor raises the question about whether such a bid should have the floor should be **enforced** even if it's not signaled. It's easy to argue either side of this question, so originally we proposed adding bidder-level floor config for both signaling and enforcement, but after community discussion, use cases covering exceptions for floor enforcement were removed.

### Use Cases

1. As a publisher, I want to define which bidders which should not receive floor signals because I've already entered floors for that provider in a different interface, or some other reason.
2. As a floors provider, I want to define which bidders which should not receive floor signals because the data shows that setting a floor on that bidder results in lower bids.
3. As an analytics provider, I need to know whether floor signaling was skipped for a bidder due to this configuration so I can monitor bidder performance.

### Proposed Floors Extension

We propose that [Floors Schema 2](https://docs.prebid.org/dev-docs/modules/floors.html#schema-2) should be extended to support configuration by either the publisher or the floors provider. So we would add 2 fields to the schema at 3 levels: top, data, and modelGroups.

- enforcement.noFloorSignalBidders: ["bidderA", "bidderB", "bidderC"]
- data.noFloorSignalBidders: ["bidderA", "bidderB"]
- data.modelGroups[].noFloorSignalBidders: ["bidderA"]

### Behavior

[...]

PR #10867, opened 2023-12-21
Price Floors: Add option for no signal bidders

apukh-magnite:
## Type of change
- [X] Feature

## Description of change
Adding new price floor option to not signal floors to the configured list of bidders 
Implements #10410

### DIFF
--- a/modules/priceFloors.js
+++ b/modules/priceFloors.js
@@ -332,13 +332,29 @@ export function getFloorDataFromAdUnits(adUnits) {
   }, {});
 }
 
+function getNoFloorSignalBidersArray(floorData) {
+  const { data, enforcement } = floorData
+  // The data.noFloorSignalBidders higher priority then the enforcment
+  if (data?.noFloorSignalBidders?.length > 0) {
+    return data.noFloorSignalBidders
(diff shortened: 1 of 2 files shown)


robertrmartinez (approved):
LGTM

Needs second non-magnite review

modules/priceFloors.js
```diff
@@ -332,20 +332,37 @@ export function getFloorDataFromAdUnits(adUnits) {
   }, {});
 }
 
+function getNoFloorSignalBidersArray(floorData) {
+  const { data, enforcement } = floorData
+  // The data.noFloorSignalBidders higher priority then the enforcment
```
dgirardi, line 337: what's the reasoning for the duplicate option? I can think of why you might want to override it when you get it through dynamic `data`, but the priority should be inverted for that. What use case is this for?

robertrmartinez: From the issue:
image

If the dynamic or data object contains it, we should use what is set there, otherwise fallback.

If a pub uses some floor provider, the floor provider's dynamic endpoint may want the most control over this, which is why data is chosen first.

dgirardi: I somehow missed that - thanks!

dgirardi (approved):

PR #10870, opened 2023-12-22
Blockthrough Bid Adapter: initial release

PavloMalashnyak:
## Type of change

- [x] New bidder adapter   

## Description of change

- contact email: engsupport@blockthrough.com
```
{
  bidder: 'blockthrough',
  params: {
    pubmatic: {
       pnlisherId: 'testId'
    }
  }
}
```

## Other information
https://github.com/prebid/prebid.github.io/pull/5050


modules/BTBidAdapter.js
```diff
@@ -0,0 +1,184 @@
+import { config } from '../src/config.js';
+import { registerBidder } from '../src/adapters/bidderFactory.js';
+import { deepSetValue, isPlainObject, logWarn } from '../src/utils.js';
+import { BANNER } from '../src/mediaTypes.js';
+import { ortbConverter } from '../libraries/ortbConverter/converter.js';
+
+const BIDDER_CODE = 'blockthrough';
+const GVLID = 815;
+// reach out for approval to marketing@blockthrough.com
+const ENDPOINT_URL = 'https://pbs.sample.com';
```
patmmccann, line 10: what is this fake value?

modules/BTBidAdapter.js
```diff
@@ -0,0 +1,184 @@
...
+const SYNC_URL = 'https://cdn.sample.com';
+
+const CONVERTER = ortbConverter({
+  context: {
+    netRevenue: true,
+    ttl: 30,
```
patmmccann, line 16: why so low

modules/BTBidAdapter.md
```diff
@@ -0,0 +1,70 @@
...
+
+```javascript
+pbjs.setBidderConfig({
+  bidders: ['blockthrough'],
+  config: {
+    ortb2: {
+      site: {
+        ext: {
+          blockthrough: {
+            ab: false,
```
patmmccann, line 33: can you define these values

test/spec/modules/BTBidAdapte_spec.js
```diff
@@ -0,0 +1,192 @@
+import { expect } from 'chai';
```
patmmccann, line 1: did you misspell the file name?

modules/BTBidAdapter.js
```diff
@@ -0,0 +1,184 @@
...
+function buildRequests(validBidRequests, bidderRequest) {
+  const data = CONVERTER.toORTB({
+    bidRequests: validBidRequests,
+    bidderRequest,
+  });
+
+  return [
+    {
+      method: 'POST',
+      url: ENDPOINT_URL,
```
patmmccann, line 110: you need a real url

patmmccann (changes requested):

modules/BTBidAdapter.md
```diff
@@ -0,0 +1,70 @@
...
+
+# Bid Params
+
+| Key    | Scope    | Type   | Description                                                    |
+| ------ | -------- | ------ | -------------------------------------------------------------- |
+| bidder | Required | Object | Bidder configuration. Could configure several bidders this way |
+
+# Bidder Config
+
+Make sure to set required ab, orgID, websiteID values received after approval using `pbjs.setBidderConfig`.
```
mkendall07, line 21: we should include an example of this

modules/BTBidAdapter.js
mkendall07: is that a requirement? I think we wanted to gate it behind approval.

modules/BTBidAdapter.js
PavloMalashnyak: We want to provide real endpoint values only after publisher approval. Are they required?

modules/BTBidAdapter.js
patmmccann: this doesnt appear to fall back to config; can you have a sensible default and fallback?

modules/BTBidAdapter.js
patmmccann: am i missing the line where you overwrite this value with the value in configuration?

mkendall07: @patmmccann no we left it out intentionally. I'll talk to the team about putting in the real endpoint.

patmmccann: I must be missing something, is your intention for the publisher to fork the project and change this constant?

mkendall07: @PavloMalashnyak please update the endpoint with the real one. Thanks

mkendall07 (approved):
LGTM. @patmmccann let me know if you have any concerns before merging

PR #10872, opened 2023-12-23
Price Floors: Failure to Account for 'data.skipRate'

bbaresic:
## Type of change
- [x] Bugfix

## Description of change
According to the documentation on the [Price Floor Module](https://docs.prebid.org/dev-docs/modules/floors.html), if the "skipRate" exists in the data object within the floors module, it should take precedence over the root-level "skipRate."

This implementation includes checks to prioritize the "skipRate" in the data object, if present.

### DIFF
--- a/modules/priceFloors.js
+++ b/modules/priceFloors.js
@@ -347,7 +347,7 @@ export function updateAdUnitsForAuction(adUnits, floorData, auctionId) {
       bid.auctionId = auctionId;
       bid.floorData = {
         skipped: floorData.skipped,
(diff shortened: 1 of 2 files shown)


patmmccann:
I think this still differs from our fork by https://github.com/cafemedia/adthrive-prebid/commit/335a3afdbbc3fa9f968f211d15f97c46189d9fbb

robertrmartinez (approved):
nice catch thanks!

ChrisHuie (approved):

PR #10876, opened 2023-12-25
Yandex Analytics Adapter: initial release

enovikov11:
## Type of change
- [x] New analytics adapter

## Description of change
- Added Yandex Analytics Adapter

Docs PR: https://github.com/prebid/prebid.github.io/pull/5052


modules/yandexAnalyticsAdapter.js
```diff
@@ -0,0 +1,222 @@
...
+  EVENTS: {
+    BID_REQUESTED,
+    BID_RESPONSE,
+    BID_ADJUSTMENT,
+    BID_WON,
+    BIDDER_DONE,
+    AUCTION_END,
+    BID_TIMEOUT,
+  },
+} = CONSTANTS;
```
spotxslagle, line 52: Does this declaration work? I get a type error when I try this.

modules/yandexAnalyticsAdapter.js
```diff
@@ -0,0 +1,222 @@
...
+
+export const EVENTS_TO_TRACK = [
+  BID_REQUESTED,
+  BID_RESPONSE,
+  BID_ADJUSTMENT,
+  BID_WON,
+  BIDDER_DONE,
+  AUCTION_END,
+  BID_TIMEOUT,
```
spotxslagle, line 61: Wouldn't these all be undefined since they're declared inside the EVENTS object above?

spotxslagle (changes requested):

spotxslagle:
@ChrisHuie Are analytics adapters allowed to append scripts? I know bid adapters aren't but I'm not sure about analytics

modules/yandexAnalyticsAdapter.js
enovikov11: There would be ["bidRequested", "bidResponse", "bidAdjustment", ...]. lines 42-52 have [destruction syntax](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Destructuring_assignment), so for example BID_REQUESTED will be CONSTANTS.EVENTS.BID_REQUESTED, that is "bidRequested" as [defined here](https://github.com/prebid/Prebid.js/blob/44b7c1591616a2a9d01b96fc8640be113ac92eb6/src/constants.json#L30) and so on

modules/yandexAnalyticsAdapter.js
enovikov11: Yes, it works, in fact it passes autotests + has been tested manually. Please share error text with me, so we can debug it.

enovikov11:
@ChrisHuie we need your opinion on appending script in analytics adapter to complete review.

As I understand rules, its allowed in analytics adapters. Also I found [10 matches](https://github.com/search?q=repo%3Aprebid%2FPrebid.js+path%3Amodules+createElement%28%22script%22%29+OR+createElement%28%27script%27%29&type=code) of `createElement('script')` in Prebid modules source code

modules/yandexAnalyticsAdapter.js
spotxslagle: Uncaught TypeError: Cannot read properties of undefined (reading 'BID_REQUESTED')
    at <anonymous>:3:5
    
    You cannot nest destructuring assignment. You need to removed the EVENTS object layer.

enovikov11: Ok, thanks for improvement, done.

ChrisHuie:
@spotxslagle @enovikov11 Will bring this up today to confirm with Patrick and Bret

spotxslagle (approved):

ChrisHuie:
@enovikov11 @spotxslagle confirmed that it is a global rule to not allow the appending of scripts. I looked at the list of other adapters and looks like 5 adapters have got it in over the years (other 5 were testing). I will create an issue and ask those adapters to remove this behavior before we release Prebid 9 in the summer.

PR #10885, opened 2023-12-29
EUID Id Module : add support for client side token generation

ssundahlTTD:
## Type of change
- [X] Feature

## Description of change
Enable Client Side Token Generation for EUID.  Follow the existing pattern set for UID2.


modules/euidIdSystem.js
```diff
@@ -99,6 +99,14 @@ export const euidIdSubmodule = {
       internalStorage: ADVERTISING_COOKIE
     };
 
+    if (FEATURES.UID2_CSTG) {
```
lionell-pack-ttd, line 102: I think re-using this feature flag is fine - I expect any publisher using CSTG for either module would almost certainly use it for both modules.

lionell-pack-ttd:
Looks good to me 👍

patmmccann (approved):

skocheri1 (approved):

PR #10897, opened 2024-01-04
Pubmatic Analytics Adapter : skip duplicate data for pubmatic bids in logger and tracker

kapil-tuptewar:
## Type of change
- [x] Feature

## Description of change
With this change, when publisher is using openwrap meta adapter with pubmatic analytics adapter, the duplicate data of pubmatic bid from logger is skipped and tracker will not be fired in case pubmatic wins auction(this will be taken care by server side tracker)


jlquaccia (approved):
LGTM

PR #10916, opened 2024-01-08
Sharethrough Bid Adapter: support cdep

jefftmahoney:
## Type of change
- [ x] Feature

## Description of change
An update to the bidder adapter so that it passes along `cdep` if it is present in the request information.


jefftmahoney:
Will try again with a branch based off of a fresher copy of `master`.

PR #10924, opened 2024-01-10
OMS Adapter: add new adapter

prBigBrother:
## Type of change

- [x] New bidder adapter   

## Description of change
In line with our rebranding efforts, I've submitted a pull request introducing a new adapter, OMS, mirroring the functionality of BrightcomSSP. This addition aims to ensure a seamless transition for our partners.

Simultaneously, I've initiated a pull request for the removal of two previous adapters, Brightcom and BrightcomSSP, in the upcoming major release #10925. 

- contact email: alexandruc@onlinemediasolutions.com
- test parameters for validating bids:
```
{
      bidder: 'oms',
      params: {
        publisherId: 2141020
      }
}
```

PR #10925, opened 2024-01-10
Brightcom adapter: remove adapters

prBigBrother:
## Type of change
- [x] Other

## Description of change
I've initiated a pull request to remove two adapters, Brightcom and BrightcomSSP, in the next major release. This is in line with our ongoing rebranding efforts and related to the next pull request  #10924


patmmccann:
This should probably also setup the old names as aliases on your new adapter when that merges in

patmmccann:
sure, see https://github.com/prebid/Prebid.js/blob/2171ca4bde6e71b0889f5ba882d6b941dcdbef72/modules/adfBidAdapter.js#L23 as an example

PR #10928, opened 2024-01-10
Magnite Analytics: Check if prebid cache was called for video tracking

robertrmartinez:
## Type of change
- [X] Other

## Description of change
A downstream signal which lets listeners know if this video bid was cached client side

### DIFF
--- a/modules/magniteAnalyticsAdapter.js
+++ b/modules/magniteAnalyticsAdapter.js
@@ -77,7 +77,7 @@ const resetConfs = () => {
     eventPending: false,
     elementIdMap: {},
     sessionData: {},
-    bidWasClientSideCached: {}
+    bidsCachedClientSide: new WeakSet()
   }
   rubiConf = {
     pvid: generateUUID().slice(0, 8),
@@ -683,7 +683,7 @@ function enableMgniAnalytics(config = {}) {
     And if it was we will use the actual bidId instead of the pbsBidId override in our BID_RESPONSE handler
 */
 export function callPrebidCacheHook(fn, auctionInstance, bidResponse, afterBidAdded, videoMediaType) {
-  cache.bidWasClientSideCached[bidResponse.requestId] = true;
+  cache.bidsCachedClientSide.add(bidResponse);
   fn.call(this, auctionInstance, bidResponse, afterBidAdded, videoMediaType);
 }
 
@@ -764,7 +764,8 @@ const handleBidResponse = (args, bidStatus) => {
 
   // if pbs gave us back a bidId, we need to use it and update our bidId to PBA
   const pbsBidId = (args.pbsBidId == 0 ? generateUUID() : args.pbsBidId) || (args.seatBidId == 0 ? generateUUID() : args.seatBidId);
-  if (pbsBidId && !cache.bidWasClientSideCached[args.requestId]) {
+  if (pbsBidId && !cache.bidsCachedClientSide.has(args)) {
+    cache.bidsCachedClientSide.delete(args);
     bid.pbsBidId = pbsBidId;
   }
 }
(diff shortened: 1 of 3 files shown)


robertrmartinez:
Hello @dgirardi 

I wanted to get your input on this, I was struggling trying to find the best way to do this, so I just through this in real quick to get your feedback.

Basically I need a way for analytics to know if a bid was cached client side or not. I did not want to do it some hacky way, so thought it may be best to add a flag in the bid response, I could not find any other simple way.

Do you have any input of something better?

dgirardi (commented):
This looks OK to me. Not sure about the name, client side cache suggests the cache is on the client... cacheRequested maybe? (I don't like that either, idk).

An alternative could be to hook into [callPrebidCache](https://github.com/prebid/Prebid.js/blob/759d5ac4f0046cb3a9f940c5a677401e871adbae/src/auction.js#L642) from your analytics to see what's going through it. I don't think that's better but it can be done without touching core, if this is only useful to rubicon.

dgirardi (approved):

robertrmartinez:
@dgirardi Thanks for the suggestion seems that is the best way to do this so I updated this PR to just be the magnite analytics adapter listenign for cache calls and doing what it needs to.

modules/magniteAnalyticsAdapter.js
```diff
@@ -749,7 +764,7 @@ const handleBidResponse = (args, bidStatus) => {
 
   // if pbs gave us back a bidId, we need to use it and update our bidId to PBA
   const pbsBidId = (args.pbsBidId == 0 ? generateUUID() : args.pbsBidId) || (args.seatBidId == 0 ? generateUUID() : args.seatBidId);
-  if (pbsBidId) {
+  if (pbsBidId && !cache.bidWasClientSideCached[args.requestId]) {
```
dgirardi, line 767: Could the entry be cleared after this - to avoid having that table grow indefinitely?

robertrmartinez: for sure was thinking of that - to confirm prebid cannot emit a new bidResponse for same bidId / requestId right?

Was thinking in the scenario a bid adapter returns more than one bidResponse for a bidRequest - need to make sure that this handles that perhaps

dgirardi: There Can Be Only One.

I think `args` here is the same object named `bidResponse` in the hook so you could also use a weakset.

robertrmartinez: THERE CAN BE ONLY ONE!!!

Good call - updated to use weakset

thanks!

modules/magniteAnalyticsAdapter.js
```diff
@@ -749,7 +764,8 @@ const handleBidResponse = (args, bidStatus) => {
 
   // if pbs gave us back a bidId, we need to use it and update our bidId to PBA
   const pbsBidId = (args.pbsBidId == 0 ? generateUUID() : args.pbsBidId) || (args.seatBidId == 0 ? generateUUID() : args.seatBidId);
-  if (pbsBidId) {
+  if (pbsBidId && !cache.bidsCachedClientSide.has(args)) {
+    cache.bidsCachedClientSide.delete(args);
```
dgirardi, line 768: this works, but also defeats the purpose of the weakset :)
if you clean up, it will break when called multiple times; using a weak ref instead solves that, but it will break if the object identity changes.

This way you're vulnerable to both, but neither looks that likely to me so it's mostly an academic discussion -  LGTM any way you choose.

dgirardi (approved):

modules/magniteAnalyticsAdapter.js
robertrmartinez: This is why I value your input!

Will message you on slack to talk through it.

robertrmartinez: You learn something new anytime you talk to Demetrio - thanks sir.

patmmccann:
@robertrmartinez I can't tell if you're holding this or not; feel free to take the review and merge it in yourself if you see fit

Issue #10566, opened 2023-10-03
Support for submitting fledge auction configs before GPT is ready


patmmccann:
From #10477: Both Prebid.js and GPT scripts are asynchronous. We load them in parallel to speed up time when ads are displayed. Thanks to this approach it's more likely that user won't leave the page until ads are displayed. In addition Active View metric should be better.

I will try to describe this process. Let's assume that we need for example 2 seconds for loading GPT script and only 1 second to load Prebid.js script. Additionally we set for example 0.8 seconds as a timeout for Prebid standard auctions (in fact this is timeout for bidders to respond).

**Our current approach**
We start Prebid auction immediately without waiting for GPT script to load so we can start fetching ads from Google Ad Manager after 2 seconds - exactly the moment when GPT script is loaded because in this example Prebid needs only 1.8 seconds to do its job.

**Your approach**
You need to wait for GPT script to be fully loaded before running Prebid auction (I don't mean Protected Audience API auction but ordinary Prebid auction) because publisher are not able to define any slot before full GPT API is available (defineSlot function is not defined before that moment). So ads can be fetched after 2.8 seconds even though Prebid script was loaded after 1 second.

**Solution**
There are two ways to resolve issue mentioned above:

1. You would need to set auction config on GPT slot inside `googletag.cmd.push(callback)`. This would change your code into asynchronous call - as `callback` has to be used. As far as I know it's not possible to do it in this specific fragment of Prebid code.
2. You provide some function with all Protected Audience API auctions configuration so publisher may set auction config on desired GPT slots by oneself. Furthermore such function could be used also with with different ad servers. So you don't need to implement many different Prebid modules for every ad server which code would be almost exactly the same - the only difference would be using certain ad server SDK to set configuration of its internal Protected Audience API component auction.

_Originally posted by @skoklowski in https://github.com/prebid/Prebid.js/issues/10477#issuecomment-1725068630_

patmmccann:
Committee is favoring the second option and looking for additional feedback or consensus.

cc: @JoelPM and @skoklowski

patmmccann:
We discussed further in committee, marking ready for dev

jdwieland8282:
Notes from the Identity PMC's 12/6 mtg regarding this topic. 

```
Option 1 - Prebid creates a “fledge for other adserver” module we should move ahead with this gptUtils.js 
DM:puts constraints on what new adservers can do. 
If this is hrs worth of work why not do it?



[...]

Issue #10573, opened 2023-10-04
Rename fledge configuration options


dgirardi:
## Type of issue
Improvement

## Description

`fledgeForGpt` defines two top-level config options, `fledgeEnabled` and `defaultForSlots`. The latter has a very generic sounding name and should probably be scoped down to something like `setConfig({fledge: {enabled: true, default: true}})`. We can also consider renaming config and module to `paapi` if we are now past the codename stage.

patmmccann:
I prefer to keep GPT in the name to distinguish from #9481

PR #10930, opened 2024-01-10
PAAPI/fledgeForGpt: make auction configs available independently from GPT

dgirardi:
## Type of change

- [x] Feature

## Description of change

- Split 'fledgeForGpt' into  new 'paapi' module that contains all GPT-independent fledge logic. 'fledgeForGpt' is now a submodule of 'paapi'
- 'paapi' now provides a new API `getPAAPIConfig({auctionId, adUnitCode})`, that returns a map from ad unit code to its PAAPI auction config. currently the only populated field is 'componentAuctions'. `auctionId` and `adUnitCode` are optional filters.
- 'fledgeForGpt' now provides a new API `setPAAPIConfigForGPT({auctionId, adUnitCode})` that sets component auction on GPT slots (auctionId/adUnitCode are still optional filters)
- 'fledgeForGpt' configuration keys are renamed to 'paapi' (e.g. `setConfig({paapi: {enabled: true}})`). 'fledgeForGpt' still works but logs a warning.
- `setConfig({paapi: {gpt: {autoconfig: false}}})` (or `setConfig({fledgeForGpt: {autoconfig: false}})` disables GPT slot auto-configuration, which together with the new APIs above should allow running the prebid auction in parallel with loading of GPT (https://github.com/prebid/Prebid.js/issues/10566)

## Other information

Closes https://github.com/prebid/Prebid.js/issues/10566
Also related: https://github.com/prebid/Prebid.js/issues/10573

### DIFF
--- a/modules/fledgeForGpt.js
+++ b/modules/fledgeForGpt.js
@@ -9,6 +9,8 @@ import {getGlobal} from '../src/prebidGlobal.js';
 
 const MODULE = 'fledgeForGpt';
 
+let getPAAPIConfig;
+
 // for backwards compat, we attempt to automatically set GPT configuration as soon as we
 // have the auction configs available. Disabling this allows one to call pbjs.setPAAPIConfigForGPT at their
 // own pace.
@@ -26,39 +28,34 @@ Object.entries({
 
 export function slotConfigurator() {
   const PREVIOUSLY_SET = {};
-  return [
-    function setComponentAuction(adUnitCode, auctionConfigs, reset = true) {
-      const gptSlot = getGptSlotForAdUnitCode(adUnitCode);
-      if (gptSlot && gptSlot.setConfig) {
-        let previous = PREVIOUSLY_SET[adUnitCode] ?? {}
-        let configsBySeller = Object.fromEntries(auctionConfigs.map(cfg => [cfg.seller, cfg]));
-        const sellers = Object.keys(configsBySeller);
-        if (reset) {
-          configsBySeller = Object.assign(previous, configsBySeller);
-          previous = Object.fromEntries(sellers.map(seller => [seller, null]))
-        } else {
-          sellers.forEach(seller => {
-            previous[seller] = null;
-          })
-        }
-        Object.keys(previous).length ? PREVIOUSLY_SET[adUnitCode] = previous : delete PREVIOUSLY_SET[adUnitCode];
-        const componentAuction = Object.entries(configsBySeller)
-          .map(([configKey, auctionConfig]) => ({configKey, auctionConfig}));
-        if (componentAuction.length > 0) {
-          gptSlot.setConfig({componentAuction});
-          logInfo(MODULE, `register component auction configs for: ${adUnitCode}: ${gptSlot.getAdUnitPath()}`, auctionConfigs);
-        }
-      } else if (auctionConfigs.length > 0) {
-        logWarn(MODULE, `unable to register component auction config for ${adUnitCode}`, auctionConfigs);
+  return function setComponentAuction(adUnitCode, auctionConfigs, reset = true) {
+    const gptSlot = getGptSlotForAdUnitCode(adUnitCode);
+    if (gptSlot && gptSlot.setConfig) {
+      let previous = PREVIOUSLY_SET[adUnitCode] ?? {};
+      let configsBySeller = Object.fromEntries(auctionConfigs.map(cfg => [cfg.seller, cfg]));
+      const sellers = Object.keys(configsBySeller);
+      if (reset) {
+        configsBySeller = Object.assign(previous, configsBySeller);
+        previous = Object.fromEntries(sellers.map(seller => [seller, null]));
+      } else {
+        sellers.forEach(seller => {
+          previous[seller] = null;
+        });
       }
-    },
-    function getConfiguredSlots() {
-      return Object.keys(PREVIOUSLY_SET)
+      Object.keys(previous).length ? PREVIOUSLY_SET[adUnitCode] = previous : delete PREVIOUSLY_SET[adUnitCode];
+      const componentAuction = Object.entries(configsBySeller)
+        .map(([configKey, auctionConfig]) => ({configKey, auctionConfig}));
+      if (componentAuction.length > 0) {
+        gptSlot.setConfig({componentAuction});
+        logInfo(MODULE, `register component auction configs for: ${adUnitCode}: ${gptSlot.getAdUnitPath()}`, auctionConfigs);
+      }
+    } else if (auctionConfigs.length > 0) {
+      logWarn(MODULE, `unable to register component auction config for ${adUnitCode}`, auctionConfigs);
     }
-  ]
+  };
 }
 
-const [setComponentAuction, getConfiguredSlots] = slotConfigurator();
+const setComponentAuction = slotConfigurator();
 
 export function onAuctionConfigFactory(setGptConfig = setComponentAuction) {
   return function onAuctionConfig(auctionId, configsByAdUnit, markAsUsed) {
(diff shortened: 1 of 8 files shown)


modules/paapi.js
```diff
@@ -0,0 +1,229 @@
...
+        bidderReq.bids.forEach(bidReq => {
+          deepSetValue(bidReq, 'ortb2Imp.ext.ae', bidReq.ortb2Imp?.ext?.ae ?? ae);
+        });
+      });
+    });
+  }
+  next(bidderRequests);
+}
+
+export function setImpExtAe(imp, bidRequest, context) {
```
jlquaccia, line 184: Super minor comment.. maybe "validateImpExtAe" instead of "setImpExtAe"?

jlquaccia (approved):
Overall LGTM

patmmccann:
putting https://github.com/prebid/Prebid.js/issues/10566#issuecomment-1881762399 here.

modules/fledgeForGpt.js
```diff
@@ -59,111 +39,36 @@ function setComponentAuction(adUnitCode, auctionConfigs) {
...
-  try {
-    const allReqs = bidderRequests?.flatMap(br => br.bids);
-    Object.entries(PENDING[auctionId]).forEach(([adUnitCode, auctionConfigs]) => {
-      const forThisAdUnit = (bid) => bid.adUnitCode === adUnitCode;
-      const slotSignals = getSlotSignals(bidsReceived?.filter(forThisAdUnit), allReqs?.filter(forThisAdUnit));
-      setComponentAuction(adUnitCode, auctionConfigs.map(cfg => mergeDeep({}, slotSignals, cfg)))
+export function setPAAPIConfigFactory(
+  getPAAPIConfig = (filters) => getGlobal().getPAAPIConfig(filters),
+  setGptConfig = setComponentAuction) {
+  return function(filters) {
```
laurb9, line 53: If `autoconfig` is `true` when this is called, we will see each _"register component auction config"_ log message twice. It is harmless but potentially confusing, or we can skip the execution since it's meant for the other case.
Thoughts ?

laurb9 (approved):
Looks ok functionally. There's a potential configuration mistake we could warn about or review the defaults to save troubleshooting time - if `paapi.enabled` is set but neither `paapi.defaultForSlots` or `ortb2Imp` in adUnits are, the config does nothing useful.

This may not be in scope but it's worth mentioning: GPT slot configs seem to linger and will keep being reused if not reset, even if later auctions yield no PAAPI configs. Should we "remember" and reset previous slot configs before new ones are set ?

laurb9:
> putting [#10566 (comment)](https://github.com/prebid/Prebid.js/issues/10566#issuecomment-1881762399) here; should we combine the set targeting and pass auction configs into one api call?

Should that be the default for `autoconfig` instead of post-auction ? otherwise it's just another way of autoconfig and some explanation is needed about when either config is appropriate to use.

dgirardi:
> Should we "remember" and reset previous slot configs before new ones are set ?

`setPAAPIConfigForGPT` (re)uses the latest config that came back for each ad unit; resetting doesn't fit well with that (but we could provide a separate reset API).

Is a component auction meant to be single use? If so, I think we want both new APIs to return or set each config only once (and the set can also reset slots without a config). This is how targeting works but unlike targeting, which is about a single bid tied to a short-lived context, PAAPI configs produce fresh auctions each time, so I didn't think that we'd want to discard them. Do we?

dgirardi:
we could add a future option to disable "caching" of configs -  `getPAAPIConfig(filters, reuseConfigs = false)` would keep track of which configs were already returned / set and disregard them.

laurb9:
> > Should we "remember" and reset previous slot configs before new ones are set ?
> 
> `setPAAPIConfigForGPT` (re)uses the latest config that came back for each ad unit; resetting doesn't fit well with that (but we could provide a separate reset API).
> 
> Is a component auction meant to be single use? If so, I think we want both new APIs to return or set each config only once (and the set can also reset slots without a config). This is how targeting works but unlike targeting, which is about a single bid tied to a short-lived context, PAAPI configs produce fresh auctions each time, so I didn't think that we'd want to discard them. Do we?

What we saw with caching is multiple `win` events for the same auctionId that may span even hours, which are technically billable but look like dupes.

Some DSPs do expect the signals to be single use, having the pub control it overrides that expectation.

If GPT could discard the config after running the auction, this wouldn't be needed to be handled by prebid. But as of now setConfig with {seller: null} is the only way to clear it, and that requires remembering all the sellers previously set.

dgirardi:
Added a `reuse` filter:

 - `getPaapiConfig` keeps track of which configs it returns. `getPaapiConfig({reuse: false})` only returns configs that it never returned before.
 - `setPAAPIConfigForGPT` keeps track of which slot/seller combinations were configured in GPT. `setPAAPIConfigForGPT({reuse: false})` only sets "fresh" configs in GPT, and resets the rest.

test/spec/modules/fledgeForGpt_spec.js
```diff
@@ -97,17 +113,90 @@ describe('fledgeForGpt module', () => {
...
+    });
+
+    describe('when reuse = false', () => {
+      it('resets all previously set configs', () => {
+        getPAAPIConfig.returns({
+          au1: {
+            componentAuctions: [{seller: 's1'}, {seller: 's2'}]
+          },
+          au2: {
+            componentAuctions: [{seller: 's1'}]
```
laurb9, line 137: If `s2` is added here the test would fail. I think setGptConfig for au2 will reset both, but we actually want the new `s2`

dgirardi:
Thank you @laurb9! 

I am starting to think that `reuse` should be a global setting; maybe `setConfig({paapi: {reuseAuctionConfigs: false}})`, so that it can be turned off for the "legacy" autoconfig case as well. Thoughts? If we do make it a setting, is there a a reason to allow it on a per-call basis as well?

dgirardi:
Opted to remove the `reuse` options and treat every auction as single-use.

jlquaccia:
> > putting [#10566 (comment)](https://github.com/prebid/Prebid.js/issues/10566#issuecomment-1881762399) here; should we combine the set targeting and pass auction configs into one api call?
> 
> Should that be the default for `autoconfig` instead of post-auction ? otherwise it's just another way of autoconfig and some explanation is needed for publishers about when either config is appropriate to use.

Only thing I'm wondering about (after catching up with Slack and comments, etc.)... I kind of feel that combining set targeting + passing auction configs would make sense/be ideal for the default `autoconfig` flow (instead of separate).  But, not sure if that is getting out of scope for this PR..  Otherwise, I feel that this PR looks pretty good.

laurb9 (approved):

rdgordon-index:
@patmmccann -- just noticed that this was merged -- and still have some concerns -- capturing them all here, as they have been scattered outside of github:

* there seem to be changes the caching semantics of the `auctionConfig` applied to any slot, which seems like a behaviour change related to the current module -- which are not mentioned in the description of this issue -- and it's not clear what impact this has for bid adapters returning their component auction configurations
* is there any documentation that describes the impact of auto-config true/false, caching or examples for when you should use the default?
* by default, by virtue of the fact that it's "backwards-compatible", the race condition isn't solved automatically
* fixing the race condition requires publishers to
  * know that there is a race condition -- and many publishers load GPT library asynchronously
  * update code-on-page to call `setPAAPIConfigForGPT` everywhere that they current set async GPT targeting 
  * ensure that they don't miss any such slots -- as the only visibility is some console error

The latter increases the burden of adoption of PAAPI -- and makes it much harder to integrate with the module -- https://github.com/prebid/Prebid.js/issues/11102 seems to suggest this is only going to be tackled in prebid 9, but I see these changes directly impacting existing working integrations, and it's been merged into prebid 8.

patmmccann:
> there seem to be changes the caching semantics of the auctionConfig applied to any slot, which seems like a behaviour change related to the current module -- which are not mentioned in the description of this issue -- and it's not clear what impact this has for bid adapters returning their component auction configurations

We seemed to resolve towards @laurb9 proposal, to treat reused auction config as unintended behavior this pr fixes. You disagreed with that?

>is there any documentation that describes the impact of auto-config true/false, caching or examples for when you should use the default?


Docs are linked and merge is pending. Would love your feedback on the docs PR (which has its own healthy discussion)

> ... update code-on-page to call setPAAPIConfigForGPT everywhere that they current set async GPT targeting...  increases the burden of adoption of PAAPI -- and makes it much harder to integrate with the module.  I see these changes directly impacting existing working integrations

I am having trouble following this. how are existing integrations impacted outside of the caching ?

patmmccann:
removing needs docs that got added after merge, perhaps tagger didnt notice https://github.com/prebid/prebid.github.io/pull/5135?

dgirardi:
> there seem to be changes the caching semantics of the auctionConfig applied to any slot [...] and it's not clear what impact this has for bid adapters returning their component auction configurations

if your adapter expects configs to be re-used, so that on any given slot it returns an auction config once and then later doesn't, it would need to be updated to return fresh configs each time. when we discussed this we concluded that the opposite seems to be the case, adapters (and the whole chain including DSPs) expect their signals to be single-use, so we opted to make them so.

> how are existing integrations impacted outside of the caching

I think the request is to change `autoconfig` so that instead of keeping the existing behavior, it augments `setTargeting`. Basically fix the race condition without requiring any change. However it would break people that rely on the current behavior and do not use `setTargeting`.

mikemangas:
hello,
does the getPAAPIConfig({auctionId, adUnitCode}) returns some valid config, only if the ssp returns a config for a given ad slot?
or where does the information comes for getPAAPIConfig({auctionId, adUnitCode}) ?
best

mikemangas:
Hi @dgirardi @patmmccann ,
where does the auctionConfig comes from? From the bid response from the ssps?

PR #10934, opened 2024-01-11
Teads Bid Adapter : support fledgeEnabled boolean

github-ishara-chan-tung:
## Type of change

- [ ] Bugfix
- [x] Feature
- [ ] New bidder adapter   
- [ ] Build related changes
- [ ] CI related changes

- [ ] Other

## Description of change

This adapts Teads bid adapter to send a boolean named fledgeEnabled that describes if fledge is supported on the browser.

Related PR for documentation : https://github.com/prebid/prebid.github.io/pull/5075

### DIFF
--- a/modules/teadsBidAdapter.js
+++ b/modules/teadsBidAdapter.js
@@ -47,6 +47,8 @@ export const spec = {
     const bids = validBidRequests.map(buildRequestObject);
     const topWindow = window.top;
 
+    const fledgeEnabled = 'runAdAuction' in navigator && 'joinAdInterestGroup' in navigator;
+
     const payload = {
       referrer: getReferrerInfo(bidderRequest),
       pageReferrer: document.referrer,
@@ -63,6 +65,7 @@ export const spec = {
       hardwareConcurrency: topWindow.navigator?.hardwareConcurrency,
       deviceMemory: topWindow.navigator?.deviceMemory,
       hb_version: '$prebid.version$',
+      fledgeEnabled: fledgeEnabled,
       ...getSharedViewerIdParameters(validBidRequests),
       ...getFirstPartyTeadsIdParameter(validBidRequests)
     };
--- a/test/spec/modules/teadsBidAdapter_spec.js
+++ b/test/spec/modules/teadsBidAdapter_spec.js
@@ -407,6 +407,24 @@ describe('teadsBidAdapter', () => {
       });
     });
 
+    it('should add fledgeEnabled info to payload when runAdAuction and joinAdInterestGroup are defined', function () {
+      navigator.runAdAuction = 'Function implementation';
+      navigator.joinAdInterestGroup = 'Function implementation';
+      const request1 = spec.buildRequests(bidRequests, bidderRequestDefault);
+      const payload1 = JSON.parse(request1.data);
+
+      expect(payload1.fledgeEnabled).to.exist;
+      expect(payload1.fledgeEnabled).to.deep.equal(true);
+
+      delete navigator.runAdAuction;
+      delete navigator.joinAdInterestGroup;
+      const request2 = spec.buildRequests(bidRequests, bidderRequestDefault);
+      const payload2 = JSON.parse(request2.data);
+
+      expect(payload2.fledgeEnabled).to.exist;
+      expect(payload2.fledgeEnabled).to.deep.equal(false);
+    });
+
     it('should add timeToFirstByte info to payload', function () {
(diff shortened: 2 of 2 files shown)


modules/teadsBidAdapter.js
```diff
@@ -47,6 +47,8 @@ export const spec = {
     const bids = validBidRequests.map(buildRequestObject);
     const topWindow = window.top;
```
patmmccann, line 49: You should already be able to get this from imp.ext.ae

patmmccann (changes requested):
Why hit navigator instead of the request object?

modules/teadsBidAdapter.js
github-ishara-chan-tung: Hi Pat, indeed I see there is a fledgeEnabled in the request object which has the following definition : 

1. Chrome enablement
2. Publisher participation in the [Origin Trial](https://developer.chrome.com/docs/privacy-sandbox/unified-origin-trial/#configure)
3. Publisher Prebid.js configuration (detailed above)
 
However in our case we only want to know the Chrome enablement part hence why I need to hit navigator instead of using the fledgeEnabled from the request in order to get the information I need

muuki88:
@github-ishara-chan-tung there's also a new meta field that you can document:

```md
privacy_sandbox: no or comma separated list of `paapi`, `topics`
```

Feel free to open a PR to https://github.com/prebid/prebid.github.io/blob/master/dev-docs/bidders/teads.md

Full details here: https://docs.prebid.org/dev-docs/bidder-adaptor.html#submitting-your-adapter

modules/teadsBidAdapter.js
patmmccann: that comes in cdep right? https://github.com/prebid/Prebid.js/blob/f95f5203d62f8dfa0b7bdeddea7074b95dbec545/src/fpd/enrichment.js#L31

We want to avoid a bunch of people hitting navigator for the same information, so if there is something you need for the request, we'd love to give it to everyone.

github-ishara-chan-tung: Hi @patmmccann thank you very much for responding. 

No. What comes with cdep is not what I want, cdep is cookieDeprecationLabel. What I want and need is the definition I wrote, which tells me if Fledge is supported on the user's browser (hence why I verify if Fledge functions are supported in the browser). It is also used here : https://github.com/prebid/Prebid.js/blob/f95f5203d62f8dfa0b7bdeddea7074b95dbec545/modules/fledgeForGpt.js#L115-L117 but this function isn't exported. If it was I could reuse it so that I would not need to hit navigator that you don't want me to use. If that can be exported I'd love to use it.

github-ishara-chan-tung: Hi @patmmccann, sorry for the ping again could you have a look at my answer above from last week ?

patmmccann: seems useless, you already know this from chrome and the label; what is this adding?

patmmccann:
@github-ishara-chan-tung we don't currently plan to allow this. Why do you want it? Why can't you determine it already?

modules/teadsBidAdapter.js
github-ishara-chan-tung: Hi @patmmccann thanks for responding.
What do you mean by `you already know this from chrome`?

From our perspective, Cookie Deprecation Labels is not reliable because:
 - It will only be used for the duration of the market test which ends in June 2024, after that we don't know if Google plans to keep it, so this won't be reliable in the long term
 - Only 1% of stable traffic browsers will have this cdep, however the Fledge API is available on more browsers
 - Even within this 1% of cdep labelled traffic, we got confirmation from Google that some browsers may have not given their consent on Chrome Privacy Sandbox, which means part of the 1% cdep labeled will not have have Fledge enabled

Lastly, our need on the adapter is to ensure that the delivery that is not compliant with Fledge won't be reaching us for Fledge, which would be a problem on server-side, in term of costs for eg. It cannot be done with the cdep alone, that's I think the same need as FledgeGpt and the reason why we have, like they do, to control the availability of the methods.

Maybe we can extract this check from both fledgeGpt and our adapter to put it as a core utility function instead?

patmmccann: > Maybe we can extract this check from both fledgeGpt and our adapter to put it as a core utility function instead?

This is quite reasonable; you'd be welcome to access the field on the request if you put it there for everyone.

patmmccann:
Closing in favor of further discussion on #10998

PR #10955, opened 2024-01-17
Oxxion Rtd Module: tracking time to run

matthieularere-msq:
## Type of change
- [x] Other

## Description of change
Randomly add a pixel call to track the time elapsed between the launch of the module and its end.

## Other information
Because it's a random sample and has no direct impact and the module behavior I haven't add tests for it. I can if you request it, I just have the feeling in this case it's a bit overkill.

### DIFF
--- a/modules/oxxionRtdProvider.js
+++ b/modules/oxxionRtdProvider.js
@@ -49,9 +49,10 @@ function getAdUnits(reqBidsConfigObj, callback, config, userConsent) {
       }
       if (typeof callback == 'function') { callback(); }
       const timeToRun = new Date() - moduleStarted;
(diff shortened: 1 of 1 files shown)


patmmccann:
@matthieularere-msq did you see https://github.com/prebid/Prebid.js/pull/10191#issuecomment-1845605061 ?

We're looking very close to merge on that one

matthieularere-msq:
> @matthieularere-msq did you see [#10191 (comment)](https://github.com/prebid/Prebid.js/pull/10191#issuecomment-1845605061) ?
> 
> We're looking very close to merge on that one

I hadn't up to now. Responding on the PR. But it's not related with this PR

ChrisHuie (approved):

Issue #10858, opened 2023-12-18
Twin ad units have distinct imp.ext.tid


patmmccann:
## Type of issue
Bug

## Description
Ad unit twins generate independent tids

https://github.com/prebid/Prebid.js/pull/8591/files defines tid as a property of the ad unit, but if we have twin ad units they should share a tid, which should represent an impression opportunity

PR #10962, opened 2024-01-18
Core: use same transaction ID for twin ad units

dgirardi:
## Type of change

- [x] Bugfix

## Description of change

Fixes https://github.com/prebid/Prebid.js/issues/10858

Note that:

- `transactionId` was used internally as a unique identifier for the ad unit; this is now replaced by a new `adUnitId`;
- `adUnitId` is not protected by the `transmitTid` activity, and exchanges could in theory use it to correlate requests. My reasoning is that since adapters would have to go out of their way to do so - picking up this new field that is only useful to break rules - we can avoid the additional overhead to gate this under the activity. I can be convinced otherwise.
- there are several appnexus clones with [this copy-pasted piece of logic](https://github.com/search?q=repo%3Aprebid%2FPrebid.js+%22TODO%3A+why+does+this+need+to+iterate+through+every+ad+unit%22&type=code) that is broken by this change. However, it's also broken by turning off TIDs (which is the default since 8), and those adapters are already waiting for a fix.

### DIFF
--- a/modules/contxtfulRtdProvider.js
+++ b/modules/contxtfulRtdProvider.js
@@ -121,7 +121,7 @@ function getReceptivity() {
  * @param { [String] } adUnits
  * @param {*} _config
  * @param {*} _userConsent
-*  @return {{ code: { ReceptivityState: String } }}
+ *  @return {{ code: { ReceptivityState: String } }}
  */
 function getTargetingData(adUnits, _config, _userConsent) {
   logInfo(MODULE, 'getTargetingData');
--- a/modules/dsp_genieeBidAdapter.js
+++ b/modules/dsp_genieeBidAdapter.js
@@ -42,21 +42,21 @@ export const spec = {
   code: BIDDER_CODE,
   supportedMediaTypes: [BANNER],
   /**
-     * Determines whether or not the given bid request is valid.
-     *
-     * @param {BidRequest} - The bid params to validate.
-     * @return boolean True if this is a valid bid, and false otherwise.
-     */
+   * Determines whether or not the given bid request is valid.
+   *
+   * @param {BidRequest} - The bid params to validate.
+   * @return boolean True if this is a valid bid, and false otherwise.
+   */
   isBidRequestValid: function (_) {
     return true;
   },
   /**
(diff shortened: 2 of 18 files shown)


dgirardi:
FYI @crumbjp @hybrid-ai @dveljovicTX @balajimediafuse

Rothalack:
I see in a lot of places you replaced transactionId with adUnitId, but in other places you added adUnitId in addition to transactionId. I assume this is to make it a non-breaking change for analytics providers etc so their existing code can still grab transactionId, but wanted to make sure this was considered and what you expect the result to be.

dgirardi:
`transactionId` was used for two purposes:

1. as a unique identifier for an impression opportunity, by bidders, analytics and (probably) rtd; these are mostly just passing it on to their endpoint;
2. as a unique identifier for an adUnit in an auction, by core and some "core" modules like priceFloors, to match bid responses with properties of the ad unit such as `mediaTypes` definitions.

<span>#</span>2 cannot use `transactionId` if it's no longer unique across twin ad units; for example, if I define twin ad units for banner and video, core needs to find the specific one matching a particular response to do things like validation and targeting. I added `adUnitId` to replace this use case - it should be transparent to anyone who cares about impression opportunities; `transactionId` is unaffected.

Rothalack (approved):
I'm satisfied with it, the idea makes sense, everything I've tested seems good to go.

ncolletti (approved):
makes sense to me and the changes look good. there's a conflict to resolve, not sure if the package-lock file needs to be committed in this PR.  

should a docs PR be opened to include `adUnitId`?  ie) https://docs.prebid.org/dev-docs/bidder-adaptor.html#building-the-request

dgirardi:
I am hesitant about documenting it because of my 2nd point from the OP:

> adUnitId is not protected by the transmitTid activity, and exchanges could in theory use it to correlate requests. My reasoning is that since adapters would have to go out of their way to do so - picking up this new field that is only useful to break rules - we can avoid the additional overhead to gate this under the activity. I can be convinced otherwise.

I don't think bidders have a legitimate case to use it. Analytics might, I'm not sure. should we gate it under `transmitTid`?

PR #10970, opened 2024-01-22
Dailymotion Bid Adaptor: initial release

kvnsw:
## Type of change

- [x] New bidder adapter : docs PR (prebid.github.io) -> https://github.com/prebid/prebid.github.io/pull/5091

## Description of change

A new bidder adapter dailymotionBidAdapter added. 

- contact email of the adapter’s maintainer : **ad-leo-engineering@dailymotion.com**

By setting the following configuration options, you'll get a constant response to any request to validate your adapter integration:

```javascript
const adUnits = [
  {
    bids: [{
      bidder: 'dailymotion',
      params: {
        apiKey: 'dailymotion-testing'
      }
    }],
    code: 'test-ad-unit',
    mediaTypes: {
      video: {
        api: [2, 7],
        context: 'instream',
        playerSize: [ [1280, 720] ],
        startDelay: 0
      },
    }
  }
];
```


modules/dailymotionBidAdapter.js
```diff
@@ -0,0 +1,55 @@
...
+   *
+   * @param {BidRequest[]} validBidRequests A non-empty list of valid bid requests that should be sent to the Server.
+   * @param {BidderRequest} bidderRequest
+   * @return ServerRequest Info describing the request to the server.
+   */
+  buildRequests: (validBidRequests = [], bidderRequest) => validBidRequests.map(bid => ({
+    method: 'POST',
+    url: 'https://pb.dmxleo.com',
+    data: {
+      bidder_request: bidderRequest,
```
dgirardi, line 33: FYI, the structure of these request objects is not stable; major versions very often introduce breaking changes to them, and sometimes even minor updates do if the refactoring required is small enough. 

People are also not required to keep Prebid up to date so as those changes happen, you'd need not only to immediately update your server side logic but also somehow keep every previous version working.

dgirardi: I realize this is probably coming too late - but if your goal is to do as much as possible server side, my recommendation is to use ortb, which is an actual protocol. The client can be [about as simple](https://github.com/prebid/Prebid.js/tree/master/libraries/ortbConverter), and while I'm not familiar with server side tooling, it's an established standard so I'd expect to find some support there as well.

kvnsw: Hello @dgirardi, we can maybe continue the discussion we started on https://github.com/dailymotion-oss/Prebid.js/pull/4#discussion_r1474155769 here with the updated PR?

kvnsw:
Hello @jlquaccia, following changes related to @dgirardi's feedback, could we have a second review? :) Thanks!

modules/dailymotionBidAdapter.md
```diff
@@ -0,0 +1,74 @@
...
+# Description
+
+Dailymotion prebid adapter.
+
+# Configuration options
+
+Before calling this adapter, you need to set its configuration with a call to setConfig like this:
+
+```javascript
+config.setConfig({
```
patmmccann, line 18: Don't configure your adapter this way, see how other bidders define params on the ad unit, eg the appnexus md file, or using pbjs.setBidderConfig, eg on grid.

patmmccann (changes requested):

modules/dailymotionBidAdapter.js
```diff
@@ -0,0 +1,75 @@
...
+
+  /**
+   * Determines whether or not the given bid request is valid.
+   * The only mandatory parameters for a bid to be valid are the api_key and position configuration entries.
+   * Other parameters are optional.
+   *
+   * @return boolean True if this is a valid bid, and false otherwise.
+   */
+  isBidRequestValid: () => {
+    const dmConfig = config.getConfig('dailymotion');
```
patmmccann, line 18: as noted below; don't call getConfig

modules/dailymotionBidAdapter.md
kvnsw: Hello @patmmccann thank you for the feedback! Updated to use `setBidderConfig` on latest commit and updated documentation on https://github.com/prebid/prebid.github.io/pull/5091 as well

modules/dailymotionBidAdapter.js
sebmil-daily: Hello @ChrisHuie, we are not using `getConfig` for ad unit params anymore as recommended here https://github.com/prebid/Prebid.js/pull/10970#discussion_r1501351540.

Also, we have updated our documentation to specify that the config should be set using `pbjs.setBidderConfig`.

Is there a better way to retrieve the parameters set by the call to `pbjs.setBidderConfig`?

I saw that this type of call is already used in other adapters like adtarget or appier.

kvnsw: Hello @patmmccann, @dgirardi, @ChrisHuie, following's @sebmil-daily last week, could we have some additional informations on why `getConfig` is a blocker for this to move forward please?

dgirardi: In this case it's for consistency, publishers are used to provide this sort of data through [bid parameters](https://docs.prebid.org/dev-docs/bidder-adaptor.html#bidder-adaptor-Designing-your-Bid-Params). An example of an adapter that takes an API key: [revcontent](https://github.com/prebid/Prebid.js/blob/c27f64b1e059c83706435db0b1eeffc88e9b3559/modules/revcontentBidAdapter.js#L33-L35)

kvnsw: Thanks @dgirardi, we'll change it now

kvnsw: @dgirardi we are now using bid params. Let us know if this works for you

sebmil-daily (approved):

kvnsw:
Hello @patmmccann, @dgirardi, @ChrisHuie, sorry to bother, I think we have addressed all your feedbacks now and have a final submission. Please let us know if we can move forward :) Thank you!

modules/dailymotionBidAdapter.js
```diff
@@ -0,0 +1,112 @@
...
+        gdprConsent: {
+          apiVersion: bidderRequest?.gdprConsent?.apiVersion || 1,
+          consentString: bidderRequest?.gdprConsent?.consentString || '',
+          // Cast boolean in any case (eg: if value is int) to ensure type
+          gdprApplies: !!bidderRequest?.gdprConsent?.gdprApplies,
+        },
+        refererInfo: {
+          page: bidderRequest?.refererInfo?.page || '',
+        },
+        uspConsent: bidderRequest?.uspConsent || '',
```
patmmccann, line 74: you should pass yourself the gpp string as usp is past end of life

modules/dailymotionBidAdapter.js
```diff
@@ -0,0 +1,112 @@
...
+      },
+      config: {
+        api_key: bid.params.apiKey
+      },
+      coppa: config.getConfig('coppa'),
```
patmmccann, line 79: please get this from the request and not the confgi

modules/dailymotionBidAdapter.js
```diff
@@ -0,0 +1,112 @@
...
+      request: {
```
patmmccann, line 80: This request is quite sparse in comparison to your competitors. Why don't you want the publisher floor, the eids, or the user.data object? You also might find site.content.id will work for you as up to three vendors may be trying to set it in a race and you need the site.content ecids.

patmmccann (changes requested):

patmmccann:
@kvnsw taking over this review from Karim; looking forward to working with you. Feel free to slack me

patmmccann (approved):

karimMourra:
@patmmccann Note that the Video Taskforce had some concerns about this PR:

In an effort to simplify configuration for publishers, the Prebid Video Taskforce is requiring bid adapters to follow the oRTB spec. 

- `mediaTypes.video` should contain fields from oRTB section 3.2.7 

- For content specific fields, we encourage reading from `site.content`, see ortb 2.6 section 3.2.13

Reference: ortb 2.6 spec: https://iabtechlab.com/wp-content/uploads/2022/04/OpenRTB-2-6_FINAL.pdf](https://github.com/prebid/Prebid.js/pull/10970)

anything outside of the spec can be included in the bidder params (i.e. params.video). Though the implementation wouldn't change, we are concerned with the examples and documentation mentioning non-ortb fields in `mediaTypes.video`

m-gendi:
Thank you @patmmccann and @karimMourra we appreciate it.
We will address the remaining comments in a new PR, aiming to resolve them.

PR #10971, opened 2024-01-22
appnexus Bid Adapter : add DSA support

jsnellbaker:
## Type of change

- [x] Feature

## Description of change
This PR adds support for the Digital Services Act (DSA) feature for the appnexus bid adapter.  In specific, it allows the publisher to specify their choices for the feature via the ortb2 config as well as supplying the DSA relevant response values into the `bid.meta` field for Prebid.js core to further process/surface to the publisher/interested party listening for that data.

Reference link to the DSA spec from the OpenRTB docs:
https://github.com/InteractiveAdvertisingBureau/openrtb/blob/main/extensions/community_extensions/dsa_transparency.md

### DIFF
--- a/modules/appnexusBidAdapter.js
+++ b/modules/appnexusBidAdapter.js
@@ -346,6 +346,36 @@ export const spec = {
       }
     }
 
+    if (bidderRequest?.ortb2?.regs?.ext?.dsa) {
+      const pubDsaObj = bidderRequest.ortb2.regs.ext.dsa;
+      const dsaObj = {};
+      ['required', 'pubrender', 'datatopub'].forEach((dsaKey) => {
+        if (isNumber(pubDsaObj[dsaKey])) {
+          if (dsaKey === 'required') {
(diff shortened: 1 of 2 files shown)


modules/appnexusBidAdapter.js
```diff
@@ -346,6 +346,36 @@ export const spec = {
       }
     }
 
+    if (bidderRequest?.ortb2?.regs?.ext?.dsa) {
+      const pubDsaObj = bidderRequest.ortb2.regs.ext.dsa;
+      const dsaObj = {};
+      ['required', 'pubrender', 'datatopub'].forEach((dsaKey) => {
```
patmmccann, line 352: there's a small possibility 'required' gets renamed or aliased

patmmccann (approved):

Fawke (approved):

modules/appnexusBidAdapter.js
patmmccann: This has now happened

Issue #10957, opened 2024-01-17
DSA Control module: toss requests without dsa information


patmmccann:
## Type of issue
Feature request

## Description
basically an extentsion of https://github.com/prebid/Prebid.js/issues/6345

If dsa info is required on the request. We should block responses without it. If pub cannot render dsa info, response should render

PR #10982, opened 2024-01-24
dsaControl module: Reject bids without meta.dsa when required

dgirardi:
## Type of change

- [x] Feature

## Description of change

Update ortbConverter to to carry ORTB response `seatbid[].bid[].ext.dsa` into bid responses' `meta.dsa`; also, add a new `dsaControl` module that:

 - rejects bids that do not provide `meta.dsa` if the auction had `ortb2.regs.ext.dsa.required` set to either 2 ("Required, bid responses without DSA object will not be accepted") or 3 ("Required, bid responses without DSA object will not be accepted, Publisher is an Online Platform")
 - logs a warning for bids that do not provide `meta.dsa` if required was set to 1 ("Supported, bid responses with or without DSA object will be accepted")

## Other information

Closes https://github.com/prebid/Prebid.js/issues/10957
ORTB DSA reference: https://github.com/InteractiveAdvertisingBureau/openrtb/blob/main/extensions/community_extensions/dsa_transparency.md

### DIFF
--- a/modules/dsaControl.js
+++ b/modules/dsaControl.js
@@ -1,4 +1,3 @@
-import {BID_RESPONSE, registerOrtbProcessor} from '../src/pbjsORTB.js';
 import {config} from '../src/config.js';
 import {auctionManager} from '../src/auctionManager.js';
 import {timedBidResponseHook} from '../src/utils/perfMetrics.js';
@@ -50,16 +49,3 @@ toggleHooks(true);
 config.getConfig('consentManagement', (cfg) => {
   toggleHooks(cfg.consentManagement?.dsa?.validateBids ?? true);
 });
-
-export function setMetaDsa(bidResponse, bid) {
-  if (bid.ext?.dsa) {
-    bidResponse.meta.dsa = bidResponse.meta.dsa ?? bid.ext.dsa;
-  }
-}
-
-registerOrtbProcessor({
-  name: 'metaDsa',
-  type: BID_RESPONSE,
-  fn: setMetaDsa,
-  priority: -1 // run after 'meta' init
-})
--- a/modules/dsaControl.js
+++ b/modules/dsaControl.js
@@ -10,18 +10,34 @@ let dsaAuctions = {};
 
 export const addBidResponseHook = timedBidResponseHook('dsa', function (fn, adUnitCode, bid, reject) {
   if (!dsaAuctions.hasOwnProperty(bid.auctionId)) {
-    dsaAuctions[bid.auctionId] = auctionManager.index.getAuction(bid)?.getFPD?.()?.global?.regs?.ext?.dsa?.required
+    dsaAuctions[bid.auctionId] = auctionManager.index.getAuction(bid)?.getFPD?.()?.global?.regs?.ext?.dsa;
   }
-  const required = dsaAuctions[bid.auctionId];
-  if (!bid.meta?.dsa) {
-    if (required === 1) {
-      logWarn(`dsaControl: ${CONSTANTS.REJECTION_REASON.DSA_REQUIRED}; will still be accepted as regs.ext.dsa.required = 1`, bid);
-    } else if ([2, 3].includes(required)) {
-      reject(CONSTANTS.REJECTION_REASON.DSA_REQUIRED);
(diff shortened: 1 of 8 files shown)


robertrmartinez:
So we will not be adding DSA by default to the standard ORTB Converter?

Only when the new module is included?

So - DSA will not be attached unless pub includes dsaControl module in their bundle.

Even if a bidder responds with it (unless they add DSA overrides to their module / converter)

I thought that DSA would always be included in standard ORTB Converter?

modules/dsaControl.js
```diff
@@ -0,0 +1,65 @@
...
+  toggleHooks(cfg.consentManagement?.dsa?.validateBids ?? true);
+});
+
+export function setMetaDsa(bidResponse, bid) {
+  if (bid.ext?.dsa) {
+    bidResponse.meta.dsa = bidResponse.meta.dsa ?? bid.ext.dsa;
+  }
+}
+
+registerOrtbProcessor({
```
patmmccann, line 60: should this just be in core? It seems if not we'll be in a world where dsa info being in meta or not will be conditional on if you install this module for only a subset of your partners

dgirardi:
I remember discussing this multiple times and agreeing that we want a separate module because non-eu people do not care about dsa.

robertrmartinez:
I have been a flake in the meetings lately so was not there.

So probably always include DSA module when you are going to include GDPR modules.

patmmccann:
> So probably always include DSA module when you are going to include GDPR modules.

This seems wild, many many publishers do not have DSA apply that have european users

robertrmartinez:
I did not mean prebid will auto include, I mean more as a decision pubs / providers make to include it when they do GDPR.

So what is our conclusion here?

All DSA logic in its own optional module?

dgirardi:
Currently, with this PR, ortbConverter sets meta.dsa universally; but rejection happens only if you include the dsaControl module.

patmmccann (approved):

patmmccann:
required is changing names https://github.com/InteractiveAdvertisingBureau/openrtb/pull/158/files

patmmccann:
Also possible to toss: if pubrender = 0, toss the response if the advertiser cannot render

patmmccann:
linking https://github.com/prebid/prebid-server/issues/3367 as parallel effort in the other repo

patmmccann:
Additional requirement:  Filter responses based on consent to purposes from this table 

image

dgirardi:
> Also possible to toss: if pubrender = 0, toss the response if the advertiser cannot render

How about when pubrender = 2 ("Publisher will render"), but the response has adrender = 1 ("buyer/advertiser will render") ?

dgirardi:
About consent checks - translated into activity controls as they are now, it would mean that when a response has `dsaparams` containing:

- `1` ("Profiling") - we check both `transmitUfpd` and `transmitEids`
- `2` ("Basic ads") - we check `fetchBids`
- `3` ("Precise geolocation") - we check `transmitPreciseGeo`

However, those activities are not meant to be used when _receiving_ bids, as their names indicate. I am worried that they do not fit the use case; if I deny a particular bidder permission to use my user first party data, I am not necessarily denying them the ability to do user profiling through other means. 

I can see a scenario where TCF purpose 4 is *not* denied, but `transmitEids`/`transmitUfpd` are;  the SSP still does some profiling through for example referrers and geography, and indicates so in their dsa response. We would then mistakenly toss out the bid.

Do we need new activities?

bretg:
What is JS doing to "toss" the bid responses? What will an analytics adapter see? On the Prebid Server side, we have this shiny new "seat nonbid" system where we can provide analytics adapters a reason why the bid was rejected instead of just "error" or "no bid".

dgirardi:
An analytics adapter would see a `bidRejected` event with one of these two messages:

https://github.com/prebid/Prebid.js/blob/b296b7226441c665470d1def2d356893fe71b7da/src/constants.json#L134-L135

Something similar would happen if we check for TCF purpose consent - this would be the first instance of that causing a bid to be rejected.

ChrisHuie (approved):

PR #10983, opened 2024-01-24
User id module: adds a catch to greedy promise

mmoschovas:
## Type of change

- [X] Bugfix
- [ ] Feature
- [ ] New bidder adapter   
- [ ] Build related changes
- [ ] CI related changes

- [ ] Other

## Description of change

 Errors that occur currently throw an Uncaught promise error on page. This adds a catch to avoid the error and logs as well for debugging

### DIFF
--- a/modules/userId/index.js
+++ b/modules/userId/index.js
@@ -488,6 +488,7 @@ function idSystemInitializer({delay = GreedyPromise.timeout} = {}) {
     }
     cancel = defer();
     return GreedyPromise.race([promise, cancel.promise])
+      .catch(() => logWarn(`${MODULE_NAME} - cancelled promise`))
       .finally(initMetrics.startTiming('userId.total'))
   }
 


modules/userId/index.js
```diff
@@ -488,6 +488,7 @@ function idSystemInitializer({delay = GreedyPromise.timeout} = {}) {
     }
     cancel = defer();
     return GreedyPromise.race([promise, cancel.promise])
+      .catch(() => logWarn(`${MODULE_NAME} - cancelled promise`))
```
dgirardi, line 491: as discussed offline, I am not able to repro the issue so I am grasping at straws here - instead of this, could you try replacing 

```
    if (cancel != null) {
      cancel.reject(INIT_CANCELED);
    }
```

with

```
    ((c) => {
      setTimeout(() => {
        if (c != null) {
          c.reject(INIT_CANCELED);
        }
      })
    })(cancel);
```

and see if it makes any difference? (this change also fails tests but I think that's a problem with the tests so if it works I can go ahead and look at them as well).

patmmccann: Doesnt appear that worked, if we change from logWarn to logError is that more palatable?

coreyb-cbs: @dgirardi @patmmccann Not sure if it's helpful, but I experienced this uncaught promise error when calling `refreshUserIds` twice in quick succession. I was able to eliminate the error by removing one of those calls (it wasn't necessary anymore). It was on my to-do list to raise this up in a Git issue to have the promise fail more gracefully or provide some sort of error context. I was able to reproduce consistently in Chrome. I can probably dig up an old bundle if that's helpful, let me know.

coreyb-cbs:
@dgirardi @patmmccann Not sure if it's helpful, but I experienced this uncaught promise error when calling refreshUserIds twice in quick succession. I was able to eliminate the error by removing one of those calls (it wasn't necessary anymore). It was on my to-do list to raise this up in a Git issue to have the promise fail more gracefully or provide some sort of error context. I was able to reproduce consistently in Chrome. I can probably dig up an old bundle if that's helpful, let me know.

dgirardi:
@coreyb-cbs could you verify if the issue still happens after https://github.com/prebid/Prebid.js/pull/12246?

PR #10985, opened 2024-01-24
Rubicon Bid Adapter: Pass on carbon segtaxes

spotxslagle:
## Type of change

- [ x ] Feature

## Description of change

Collect carbon sextaxes and pass them in bid request

### DIFF
--- a/modules/rubiconBidAdapter.js
+++ b/modules/rubiconBidAdapter.js
@@ -177,6 +177,11 @@ export const converter = ortbConverter({
 
     delete data?.ext?.prebid?.storedrequest;
 
+    let carbonTopics = fetchCarbonTopics();
+    if (carbonTopics.length > 0) {
+      deepSetValue(data, 'user.data', carbonTopics);
+    }
+
     // floors
     if (rubiConf.disableFloors === true) {
       delete data.ext.prebid.floors;
@@ -527,6 +532,15 @@ export const spec = {
 
     if (bidRequest?.ortb2Imp?.ext?.ae) {
       data['o_ae'] = 1;
+    }
+
+    const topics = readCarbonTopics();
+    if (topics) {
+      const domain = bidderRequest.refererInfo.domain
+      Object.keys(topics).forEach(taxonomy => {
+        const field = `s_segs_${taxonomy}_${domain}`;
+        data[field] = topics[taxonomy];
+      });
     }
     // loop through userIds and add to request
     if (bidRequest.userIdAsEids) {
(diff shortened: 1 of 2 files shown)


patmmccann (dismissed):
You should not be using getConfig for this

modules/rubiconBidAdapter.js
```diff
@@ -1001,6 +1015,32 @@ function applyFPD(bidRequest, mediaType, data) {
   }
 }
 
+function fetchCarbonTopics() {
+  let topics = [];
+  if (rubiConf.readTopics !== false) {
+    topics = config.getConfig('user.data')?.filter(topic => {
+      const taxonomy = topic.taxonomyVersion;
```
patmmccann, line 1022: this is incorrect as well, you should look at segtax on getUserData.

modules/rubiconBidAdapter.js
```diff
@@ -528,6 +528,13 @@ export const spec = {
     if (bidRequest?.ortb2Imp?.ext?.ae) {
       data['o_ae'] = 1;
     }
+
+    const topics = getCarbonTopics(bidderRequest);
+    if (topics) {
+      Object.keys(topics).forEach(field => {
+        data[field] = topics[field];
+      });
+    }
```
robertrmartinez, line 537: Not really a huge performance increase but no need to have two loops when we can just add those keys to the data object during the initial getCarbonTopics loop.

And maybe we can generalize it to not say carbon and just "Get wanted segtax" or topics or whatever. And it just takes in list of taxonomy numbers?

So if we need to add another one we just update the list or something. - guess you can just update the if statement also so no worries.

modules/rubiconBidAdapter.js
```diff
@@ -1001,6 +1008,23 @@ function applyFPD(bidRequest, mediaType, data) {
...
+function getCarbonTopics(bidderRequest) {
+  if (rubiConf.readTopics === false) {
+    return undefined;
+  }
+  let fields = {};
+  let userData = bidderRequest.ortb2?.user?.data || [];
+  userData.forEach(topic => {
+    const taxonomy = topic.ext?.segtax;
+    if (taxonomy == 507 || taxonomy == 508) {
+      const domain = bidderRequest.refererInfo.domain;
```
robertrmartinez, line 1020: I think this should be

```js
const domain = topic.name || bidderRequest.refererInfo.domain;
```

robertrmartinez (changes requested):
one necessary change and one suggestion

modules/rubiconBidAdapter.js
```diff
@@ -1001,6 +1004,22 @@ function applyFPD(bidRequest, mediaType, data) {
   }
 }
 
+function addDesiredSegtaxes(bidderRequest, target) {
+  if (rubiConf.readTopics === false) {
+    return;
+  }
+  let iSegments = [1, 2, 5, 6, 507].concat(rubiConf.sendUserSegtax?.map(seg => Number(seg)));
+  let vSegments = [4, 508].concat(rubiConf.sendSiteSegtax?.map(seg => Number(seg)));
```
robertrmartinez, line 1011: I think these will add `undefined` to the arrays

To get around this can just add `|| []` I think at end of map so it skips it.

Other ways to handle also.

Think adding `undefined` to these arrays is prob not a big deal, but is unnecessary. 

EX:

```js
iSegments = [1, 2, 5, 6, 507].concat(rubiConf.sendUserSegtax?.map(seg => Number(seg)) || []);
```

modules/rubiconBidAdapter.js
```diff
@@ -1001,6 +1004,22 @@ function applyFPD(bidRequest, mediaType, data) {
...
+  let userData = bidderRequest.ortb2?.user?.data || [];
+  userData.forEach(topic => {
+    const taxonomy = Number(topic.ext?.segtax);
+    let char;
+    if ((iSegments.includes(taxonomy) && (char = 'i')) || (vSegments.includes(taxonomy) && (char = 'v'))) {
```
robertrmartinez, line 1016: cool way to handle the v and I thing nice! sneaky assignment!

robertrmartinez (changes requested):
one minor change

Also can we add a unit test or two which shows the `rubiConf.sendUserSegtax` and `rubiConf.sendSiteSegtax` work? maybe one for each combo

User
Site
Both

robertrmartinez (approved):

robertrmartinez:
@ChrisHuie @patmmccann Needs 2nd non magnite review. Don't want to get in trouble again! haha

patmmccann (approved):

PR #10988, opened 2024-01-25
IX Bid Adapter: Removing dependency on EID Allowlist

sajidbsk:
## Type of change

- [x] Feature

## Description of change

Remove the dependence of IXAdapter on a eid source allowlist so that we don't need to open a new prebidJS MR everytime we want to support a new eid source in the bidRequest.


lksharma (approved):
LGTM

PR #10990, opened 2024-01-25
E2E testing: Remove @wdio/sync and update @wdio/* to the latest

jamarzka:
## Type of change
- [X] Build related changes

## Description of change
The wdio/sync package depends on fibers which is obsolete and is not compatible with Node 16. This change updates to the latest versions of the wdio packages and adds support for running local e2e tests.

I tested these changes locally with `gulp e2e-test --local` but didn't confirm it works with browserstack.

Is there a reason why this project is still using circleci Node 12? The updated @wdio packages require Node 16+. I added a commit to update the circleci config but I'm assuming there is a reason for using such an old version?

If updating the Node version is out of the question, I'll revert to @wdio 7 and drop support for local e2e testing.


patmmccann:
> Is there a reason why this project is still using circleci Node 12?

image

jamarzka:
I only updated it to Node 16 because I was assumed there was a technical reason why it wasn't updated. Should I just update circleci to Node 20? Should the package.json min Node version be updated to 16?

patmmccann:
> Should the package.json min Node version be updated to 16?

Would this break a build script? That makes me a little nervous. We don't have any reason to hold back other than being careful not to break anyone building automatically.

jamarzka:
> Thank you very much for this!
> 
> I don't think updating the node version used by our pipelines is a breaking change; it was on 12 because that was the LTS version the last time we touched it. There are other systems that run the build on 12 but not the integ tests, so they should be unaffected (and should also be updated regardless). From my local testing building on 12 still works fine.

I was referring to updating the package.json minimum node version as a breaking change and agree updating circleci isn't. It would be nice to continue to run on the minimum supported version in circleci so breaking changes in the build process aren't introduced but agree these minimum versions just need to get updated.

A few observations:
 * current minimum node version in package.json is >=8.9.0 (npm v5)
 * current minimum version of webpack installed (5.70.0) requires Node >=10.13.0 (npm v6)
 * current package-lock.json uses v2 of the format but npm v5 that is used with node 8.9.0 uses version 1
 * webpack.conf.js uses Object.fromEntries which was added in Node 12
 
Since Node 12 is the true minimum version, should I update package.json to that version in this PR?

For the next major release, it probably makes sense to update the min node version to at least Node 16 but maybe the current LTS?

patmmccann:
Since Node 12 is the true minimum version, should I update package.json to that version in this PR? 
Yes, if 12 is actually required, it seems useful to reflect that in package.json

Should I just update circleci to Node 20? 
Yes, no concerns

Should the package.json min Node version be updated to 16?
No, not until prebid 9

For the next major release, it probably makes sense to update the min node version to at least Node 16 but maybe the current LTS?
Seems 16 is fine in prebid 9

patmmccann:
https://github.com/prebid/Prebid.js/issues/10999 for the pb9 change

jamarzka:
I updated the package.json min Node version to 12 and added a version check before the e2e task in case someone has been running them with Node 12.

patmmccann (approved):

PR #11008, opened 2024-01-30
setupad Bid Adapter: initial commit

pashaGhub:
## Type of change

- [x] New bidder adapter  

## Description of change

created new setupadBidAdapter

- contact email
it@setupad.com
```
{
    bidder: 'setupad',
    params: {
      placement_id: '3916',
      account_id: 'test',
   },
},
```

## Other information

SETUPAD prebid.github.io PR: https://github.com/prebid/prebid.github.io/pull/5114

### DIFF
--- a/modules/setupadBidAdapter.js
+++ b/modules/setupadBidAdapter.js
@@ -45,13 +45,12 @@ export const spec = {
     const requests = [];
     window.nmmRefreshCounts = window.nmmRefreshCounts || {};
     _each(validBidRequests, function (bid) {
-      window.nmmRefreshCounts[bid.adUnitCode] =
-        window.nmmRefreshCounts[bid.adUnitCode] || 0;
+      window.nmmRefreshCounts[bid.adUnitCode] = window.nmmRefreshCounts[bid.adUnitCode] || 0;
       const id = getBidIdParameter('placement_id', bid.params);
       const accountId = getBidIdParameter('account_id', bid.params);
       const auctionId = bid.auctionId;
       const bidId = bid.bidId;
-      const eids = getEids(bid) || [];
+      const eids = getEids(bid) || undefined;
       let sizes = bid.sizes;
       if (sizes && !Array.isArray(sizes[0])) sizes = [sizes];
 
@@ -65,12 +64,6 @@ export const spec = {
             storedrequest: {
               id: accountId || 'default',
             },
-          },
-
-          setupad: {
-            refresh_count: window.nmmRefreshCounts[bid.adUnitCode]++,
-            elOffsets: getBoundingClient(bid),
-            scrollTop: window.pageYOffset || document.documentElement.scrollTop,
           },
         },
         user: { ext: { eids } },
@@ -233,8 +226,7 @@ export const spec = {
       placementIdsArray.push(param.placement_id);
     });
 
-    const placementIds =
-      (placementIdsArray.length && placementIdsArray.join(';')) || '';
+    const placementIds = (placementIdsArray.length && placementIdsArray.join(';')) || '';
 
     if (!placementIds) return;
 
@@ -277,26 +269,6 @@ function getBidders(serverResponse) {
   if (bidders.length) {
     return encodeURIComponent(JSON.stringify([...new Set(bidders)]));
   }
-}
-
-function getAdEl(bid) {
-  // best way I could think of to get El, is by matching adUnitCode to google slots...
-  const slot =
-    window.googletag &&
-    window.googletag.pubads &&
-    window.googletag
-      .pubads()
-      .getSlots()
-      .find((slot) => slot.getAdUnitPath() === bid.adUnitCode);
-  const slotElementId = slot && slot.getSlotElementId();
-  if (!slotElementId) return null;
-  return document.querySelector('#' + slotElementId);
-}
-
-function getBoundingClient(bid) {
-  const el = getAdEl(bid);
-  if (!el) return {};
-  return el.getBoundingClientRect();
 }
 
 function getAd(bid) {
@@ -355,11 +327,10 @@ function initSendingDataStatistic() {
     disabledSending = false;
     enabledSending = false;
     auctionIds = {};
-    eventHendlers = {};
+    eventHandlers = {};
 
     initEvents() {
-      this.disabledSending =
-        !!config.getBidderConfig()?.setupad?.disabledSendingStatisticData;
+      this.disabledSending = !!config.getBidderConfig()?.setupad?.disabledSendingStatisticData;
       if (this.disabledSending) {
         this.removeEvents();
       } else {
@@ -372,11 +343,11 @@ function initSendingDataStatistic() {
 
       this.enabledSending = true;
       for (let eventName of this.eventNames) {
-        if (!this.eventHendlers[eventName]) {
-          this.eventHendlers[eventName] = this.eventHandler(eventName);
+        if (!this.eventHandlers[eventName]) {
+          this.eventHandlers[eventName] = this.eventHandler(eventName);
         }
 
-        events.on(eventName, this.eventHendlers[eventName]);
+        events.on(eventName, this.eventHandlers[eventName]);
       }
     }
 
@@ -385,9 +356,9 @@ function initSendingDataStatistic() {
 
       this.enabledSending = false;
       for (let eventName of this.eventNames) {
-        if (!this.eventHendlers[eventName]) continue;
+        if (!this.eventHandlers[eventName]) continue;
 
-        events.off(eventName, this.eventHendlers[eventName]);
+        events.off(eventName, this.eventHandlers[eventName]);
       }
     }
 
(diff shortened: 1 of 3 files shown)


modules/setupadBidAdapter.js
```diff
@@ -0,0 +1,420 @@
...
+  const bidders = serverResponse
+    .map((res) => Object.keys(res.body.ext.responsetimemillis || []))
+    .flat(1);
+
+  if (bidders.length) {
+    return encodeURIComponent(JSON.stringify([...new Set(bidders)]));
+  }
+}
+
+function getAdEl(bid) {
```
patmmccann, line 282: none of the below is allowed, what are you trying to do?

modules/setupadBidAdapter.js
```diff
@@ -0,0 +1,420 @@
...
+        events.on(eventName, this.eventHendlers[eventName]);
+      }
+    }
+
+    removeEvents() {
+      if (!this.enabledSending) return;
+
+      this.enabledSending = false;
+      for (let eventName of this.eventNames) {
+        if (!this.eventHendlers[eventName]) continue;
```
patmmccann, line 388: spelling

modules/setupadBidAdapter.js
```diff
@@ -0,0 +1,420 @@
...
+    ];
+
+    disabledSending = false;
+    enabledSending = false;
+    auctionIds = {};
+    eventHendlers = {};
+
+    initEvents() {
+      this.disabledSending =
+        !!config.getBidderConfig()?.setupad?.disabledSendingStatisticData;
```
patmmccann, line 362: make sure you document this

modules/setupadBidAdapter.js
pashaGhub: copy that, the function was removed.

modules/setupadBidAdapter.js
pashaGhub: spelling corrected

modules/setupadBidAdapter.js
pashaGhub: sure, in our docs we explain how to disable sending statistics data

robertrmartinez:
I am not sure all of this event handling `"sendingDataStatistic"` should be done in a bid adapter.

This is what Analytics Adapters are for.

As far as I can tell there is no precedence for this and I do not think we should start now unless a compelling reason is given.

The right thing to do is to have a `setupadBidAdapter` and a `setupadAnalyticsAdapter`

Will ask for feedback from the Prebid Org team

@patmmccann @dgirardi

pashaGhub:
> I am not sure all of this event handling `"sendingDataStatistic"` should be done in a bid adapter.
> 
> This is what Analytics Adapters are for.
> 
> As far as I can tell there is no precedence for this and I do not think we should start now unless a compelling reason is given.
> 
> The right thing to do is to have a `setupadBidAdapter` and a `setupadAnalyticsAdapter`
> 
> Will ask for feedback from the Prebid Org team
> 
> @patmmccann @dgirardi

actually, we have precedence already: nextMilleniumBidAdapter is sending analytics from their bid adapter.

ChrisHuie:
> > I am not sure all of this event handling `"sendingDataStatistic"` should be done in a bid adapter.
> > This is what Analytics Adapters are for.
> > As far as I can tell there is no precedence for this and I do not think we should start now unless a compelling reason is given.
> > The right thing to do is to have a `setupadBidAdapter` and a `setupadAnalyticsAdapter`
> > Will ask for feedback from the Prebid Org team
> > @patmmccann @dgirardi
> 
> actually, we have precedence already: nextMilleniumBidAdapter is sending analytics from their bid adapter.

Yeah. This is more a question of the behavior we want going forward. We would make nextMillenium update their adapter as well if the committee decides this isn't desired behavior.

pashaGhub:
> > > I am not sure all of this event handling `"sendingDataStatistic"` should be done in a bid adapter.
> > > This is what Analytics Adapters are for.
> > > As far as I can tell there is no precedence for this and I do not think we should start now unless a compelling reason is given.
> > > The right thing to do is to have a `setupadBidAdapter` and a `setupadAnalyticsAdapter`
> > > Will ask for feedback from the Prebid Org team
> > > @patmmccann @dgirardi
> > 
> > 
> > actually, we have precedence already: nextMilleniumBidAdapter is sending analytics from their bid adapter.
> 
> Yeah. This is more a question of the behavior we want going forward. We would make nextMillenium update their adapter as well if the committee decides this isn't desired behavior.

copy that. Waiting for response

robertrmartinez:
My apologies, when I did a search it did not come up for any other bid adapters using the events system.

Here is a refined search

image

I still do not agree with this and think we should not allow this in Bid Adapters as event handlers are blocking as far as I remember.

If we do allow it - it should default as off in my opinion.

Will talk with the team and get back to you. Thanks for understanding!

dimention:
@robertrmartinez @ChrisHuie both Holid and NextMillenium (and us) are using Prebid Server for the backend of these adapters. The purpose of the event handlers is to avoid having to insert tracking pixels into the creative codes and make other corrections in the Prebid Server code, as it doesn't support it out-of-the-box.

robertrmartinez:
Hey @pashaGhub and @dimention 

After discussing with the Prebid team we have deemed that bid adapters are not allowed to import the events system any longer.

It was kind of a loop hole in the rules that we are officially closing.

Here are two issues opened up:

https://github.com/prebid/Prebid.js/issues/11071
and 
https://github.com/prebid/Prebid.js/issues/11072

Bid Adapters can use the existing event handlers provided in the specs if needed and send signals using those.

But importing the general events system is not allowed.

pashaGhub:
> Hey @pashaGhub and @dimention
> 
> After discussing with the Prebid team we have deemed that bid adapters are not allowed to import the events system any longer.
> 
> It was kind of a loop hole in the rules that we are officially closing.
> 
> Here are two issues opened up:
> 
> #11071 and #11072
> 
> Bid Adapters can use the existing event handlers provided in the specs if needed and send signals using those.
> 
> But importing the general events system is not allowed.

Hey,
events system removed.

pashaGhub:
Hey @robertrmartinez, @ChrisHuie, @patmmccann 
any updates?

robertrmartinez:
Hey @pashaGhub sorry for the delay here.

Will review this today.

modules/setupadBidAdapter.js
```diff
@@ -0,0 +1,299 @@
...
+      } else if (bid.nurl) {
+        adUrl = bid.nurl;
+      }
+  }
+
+  return { ad, adUrl, vastXml, vastUrl };
+}
+
+function getSiteObj() {
+  const refInfo = (getRefererInfo && getRefererInfo()) || {};
```
robertrmartinez, line 275: You probably do not need to import `getRefererInfo` as on the `bidderRequest` object there should be the referer info you are looking for.

Example:

image


So maybe you can just pass `bidderRequest.refererInfo` into `getSiteObj` as `refInfo` in the func declaration and get rid of the import?

modules/setupadBidAdapter.js
```diff
@@ -0,0 +1,299 @@
...
+      window.innerWidth ||
+      window.document.documentElement.clientWidth ||
+      window.document.body.clientWidth ||
+      0,
+    h:
+      window.innerHeight ||
+      window.document.documentElement.clientHeight ||
+      window.document.body.clientHeight ||
+      0,
+  };
```
robertrmartinez, line 296: Both these device and site objects are probably available in the `ortb` object of your `bidderRequest`

Examples:

image


So maybe can simplify the logic to pick from there.

robertrmartinez (changes requested):
Couple minor things pointed out.

Also, your endpoint request and response seems to be in standard ORTB, you can always use the ORTB Converter from prebid libraries to generate standard ORTB which may make a more simple adapter so you do not have to do those translations yourself.

https://github.com/prebid/Prebid.js/tree/master/libraries/ortbConverter

Example implementations: 
https://github.com/prebid/Prebid.js/blob/master/modules/openxBidAdapter.js
https://github.com/prebid/Prebid.js/blob/master/modules/kimberliteBidAdapter.js

Not required just a suggestion!

pashaGhub:
> Couple minor things pointed out.
> 
> Also, your endpoint request and response seems to be in standard ORTB, you can always use the ORTB Converter from prebid libraries to generate standard ORTB which may make a more simple adapter so you do not have to do those translations yourself.
> 
> https://github.com/prebid/Prebid.js/tree/master/libraries/ortbConverter
> 
> Example implementations: https://github.com/prebid/Prebid.js/blob/master/modules/openxBidAdapter.js https://github.com/prebid/Prebid.js/blob/master/modules/kimberliteBidAdapter.js
> 
> Not required just a suggestion!

Thanks a lot for the information!
We are staying with the current solution at the moment, however, we will definitely consider updating our adapter in the future.

robertrmartinez (approved):

PR #11009, opened 2024-01-30
AppNexus Bid Adapter: stop suppressing user syncs on any gpp string with section > 5

patmmccann:
Activity controls should be sufficient to take care of this, or the AppNexus adapter should examine the strings.


jsnellbaker (approved):
I think it'll be okay to rely on the activity controls feature here.

PR #11012, opened 2024-01-30
OpenX Bid Adapter: add ortb2Imp to PAAPI auctionSignals

laurb9:
## Type of change
- [x] Feature

## Description of change
Inject the OpenRTB `imp` request object into PAAPI `auctionSignals.ortb2Imp`, to expose contextual slot information such as `banner.formats` to all PAAPI component buyers under the OpenX seller.

### DIFF
--- a/modules/openxBidAdapter.js
+++ b/modules/openxBidAdapter.js
@@ -109,7 +109,7 @@ const converter = ortbConverter({
           bidId,
           config: mergeDeep(Object.assign({}, cfg), {
             auctionSignals: {
-              ortb2Imp: context.impContext[bidId]?.ortbRequest?.imp?.[0],
+              ortb2Imp: context.impContext[bidId]?.imp,
             },
           }),
         }
--- a/modules/openxBidAdapter.js
+++ b/modules/openxBidAdapter.js
@@ -107,9 +107,11 @@ const converter = ortbConverter({
       fledgeAuctionConfigs = Object.entries(fledgeAuctionConfigs).map(([bidId, cfg]) => {
         return {
           bidId,
-          config: Object.assign({
-            auctionSignals: {},
-          }, cfg)
+          config: mergeDeep(Object.assign({}, cfg), {
+            auctionSignals: {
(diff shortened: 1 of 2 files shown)


patmmccann:
Why not do this in the PAAPI module?

patmmccann (changes requested):
Let's do this in the module, no?

laurb9:
> Why not do this in the PAAPI module?

That would be my preference also (to replace the current `prebid.ortb2Imp` with this). I had a quick chat with @dgirardi last year about this, it seemed less straightforward than I'd hoped. This is an interim solution to avoid having to populate it in the exchange, and we'd drop it if/when the PAAPI module implements it.

Since now Demetrio is working on the PAAPI module refactor, maybe it's time to take another look ?

dgirardi:
This can't be done in general because, in general, there's no `imp` object.  (We already include `ortb2Imp` as defined by the publisher, but this is looking for the complete request with `banner` and so on, which is up to each adapter).

We'd like to do it in ortbConverter but AFAIK there's still no standard way to return fledge configs in ortb, so we don't have a home for it there either.

modules/openxBidAdapter.js
```diff
@@ -107,9 +107,11 @@ const converter = ortbConverter({
       fledgeAuctionConfigs = Object.entries(fledgeAuctionConfigs).map(([bidId, cfg]) => {
         return {
           bidId,
-          config: Object.assign({
-            auctionSignals: {},
-          }, cfg)
+          config: mergeDeep(Object.assign({}, cfg), {
+            auctionSignals: {
+              ortb2Imp: context.impContext[bidId]?.ortbRequest?.imp?.[0],
```
dgirardi, line 112: you can use `context.impContext[bidId]?.imp` to get the precise imp tied to the bid id (instead of the first in the request)

patmmccann (commented):
i follow now, ty

ncolletti (approved):
looks good to me

PR #11019, opened 2024-01-31
Adquery Bid Adapter : IdSystem fix getId and decode method, added userSync with iframe type

adquery:
## Type of change
- [x] Bugfix
- [x] Feature

## Description of change

- adapter: buildRequests method refactor
- adapter: getUserSyncs added iframe support
- idSystem: chenged getId method


JulieLorin (approved):
LGTM

PR #11032, opened 2024-02-01
Kimberlite Bidder Adapter: initial commit

solta-dev:
## Type of change
- [x] New bidder adapter

## Description of change
New Kimberlite Bidder Adapter

[dev@solta.io](dev@solta.io)
```
{
      bidder: "kimberlite",
      params: {
          placementId: '123'
      }
}
```
## Other information
Kimberlite prebid.github.io PR: https://github.com/prebid/prebid.github.io/pull/5124

### DIFF
--- a/modules/kimberliteBidAdapter.js
+++ b/modules/kimberliteBidAdapter.js
@@ -1,12 +1,18 @@
 import { registerBidder } from '../src/adapters/bidderFactory.js';
 import { BANNER } from '../src/mediaTypes.js';
 import { ortbConverter } from '../libraries/ortbConverter/converter.js'
+import { deepSetValue } from '../src/utils.js';
 
 const VERSION = '1.0.0';
 
 const BIDDER_CODE = 'kimberlite';
 const METHOD = 'POST';
 const ENDPOINT_URL = 'https://kimberlite.io/rtb/bid/pbjs';
+
+const VERSION_INFO = {
+  ver: '$prebid.version$',
+  adapterVer: `${VERSION}`
+};
 
 const converter = ortbConverter({
   context: {
@@ -17,18 +23,10 @@ const converter = ortbConverter({
 
   request(buildRequest, imps, bidderRequest, context) {
     const bidRequest = buildRequest(imps, bidderRequest, context);
-    bidRequest.site ||= {};
-    bidRequest.site.publisher ||= {
-      domain: bidderRequest.refererInfo.domain
-    };
-    bidRequest.site.page ||= bidderRequest.refererInfo.page;
+    deepSetValue(bidRequest, 'site.publisher.domain', bidderRequest.refererInfo.domain);
+    deepSetValue(bidRequest, 'site.page', bidderRequest.refererInfo.page);
+    deepSetValue(bidRequest, 'ext.prebid', VERSION_INFO);
     bidRequest.at = 1;
-    bidRequest.ext = {
-      prebid: {
-        ver: '$prebid.version$',
-        adapterVer: `${VERSION}`
-      }
-    };
     return bidRequest;
   },
 
--- a/modules/kimberliteBidAdapter.js
+++ b/modules/kimberliteBidAdapter.js
@@ -25,7 +25,8 @@ const converter = ortbConverter({
     const bidRequest = buildRequest(imps, bidderRequest, context);
     deepSetValue(bidRequest, 'site.publisher.domain', bidderRequest.refererInfo.domain);
     deepSetValue(bidRequest, 'site.page', bidderRequest.refererInfo.page);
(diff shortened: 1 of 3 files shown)


solta-dev:
Hi, @robertrmartinez!
We're just wondering if there any ETA for our PR or there're some issues with it? Please let us know if any.

Regards,
@solta-dev

robertrmartinez:
Hey @solta-dev Will give this a review today / tomorrow! Thanks for the patience.

test/spec/modules/kimberliteBidAdapter_spec.js
```diff
@@ -0,0 +1,167 @@
...
+        placementId: 'test-placement'
+      }
+    };
+
+    it('pass on valid bidRequest', function () {
+      assert.isTrue(spec.isBidRequestValid(validBid));
+    });
+
+    it('fails on missed placementId', function () {
+      let bid = deepClone(validBid);
```
robertrmartinez, line 28: instead of doing a deepClone each time you could just have a `beforeEach` which sets the `validBid` object each time.

Then you just can alter it however you want in your test, not worrying about it effecting downstream tests

modules/kimberliteBidAdapter.js
```diff
@@ -0,0 +1,72 @@
...
+      domain: bidderRequest.refererInfo.domain
+    };
+    bidRequest.site.page ||= bidderRequest.refererInfo.page;
+    bidRequest.at = 1;
+    bidRequest.ext = {
+      prebid: {
+        ver: '$prebid.version$',
+        adapterVer: `${VERSION}`
+      }
+    };
```
robertrmartinez, line 31: We want to overwrite the `ext` object to just be this each time? Is that intended?

modules/kimberliteBidAdapter.js
```diff
@@ -0,0 +1,72 @@
...
+    mediaType: BANNER,
+    netRevenue: true,
+    ttl: 300
+  },
+
+  request(buildRequest, imps, bidderRequest, context) {
+    const bidRequest = buildRequest(imps, bidderRequest, context);
+    bidRequest.site ||= {};
+    bidRequest.site.publisher ||= {
+      domain: bidderRequest.refererInfo.domain
```
robertrmartinez, line 22: So we only want to set `site.publisher.domain` if the incoming `publisher` object was not defined yet?

If ORTB converter populates `publisher`, with something like

```js
{
   url: 'some url'
}
```

Then we would not send `domain` at all.

solta-dev: That's a miss yes, thank you! We definitely should fix this.

modules/kimberliteBidAdapter.js
solta-dev: You're right, we should extend it, yes. Thank you!

solta-dev:
@robertrmartinez,  thank for your review! We have a small question about further process. Shall we initiate a new PR or we can update this one?

robertrmartinez:
Just make an update to this PR directly!

modules/kimberliteBidAdapter.js
robertrmartinez: I would recommend just doing a deepSetValue of the values you want to overwrite / set.

modules/kimberliteBidAdapter.js
```diff
@@ -17,18 +23,10 @@ const converter = ortbConverter({
...
   request(buildRequest, imps, bidderRequest, context) {
     const bidRequest = buildRequest(imps, bidderRequest, context);
-    bidRequest.site ||= {};
-    bidRequest.site.publisher ||= {
-      domain: bidderRequest.refererInfo.domain
-    };
-    bidRequest.site.page ||= bidderRequest.refererInfo.page;
+    deepSetValue(bidRequest, 'site.publisher.domain', bidderRequest.refererInfo.domain);
+    deepSetValue(bidRequest, 'site.page', bidderRequest.refererInfo.page);
+    deepSetValue(bidRequest, 'ext.prebid', VERSION_INFO);
```
robertrmartinez, line 28: So if incoming ext.prebid exists this will also overwrite it!

```js
deepSetValue(bidRequest, 'ext.prebid.ver', VERSION_INFO.ver);
deepSetValue(bidRequest, 'ext.prebid.adapterVer', VERSION_INFO.adapterVer);
```

maybe is what you want?

robertrmartinez (changes requested):

modules/kimberliteBidAdapter.js
solta-dev: It was not intended that this field would be filled in by anyone, is that so?

solta-dev: Anyway it's better to be changed in a way you provided.

PR #11035, opened 2024-02-01
New bidder adapter : RixEngine

xiaochang:
## Type of change

- [ ] Bugfix
- [ ] Feature
- [x] New bidder adapter   
- [ ] Build related changes
- [ ] CI related changes

- [ ] Other

## Other information
[RixEngine bid adapter documentation](https://github.com/prebid/prebid.github.io/pull/5125)


lksharma (approved):
LGTM

patmmccann (changes requested):
Please make the user ID opt in and provide a real endpoint

patmmccann:
Can you provide a fallback or test endpoint? We just need something we can validate responds, particularly if we add additional e2e testing in the future.

BaiShuoyang:
> Can you provide a fallback or test endpoint? We just need something we can validate responds, particularly if we add additional e2e testing in the future.

@patmmccann Thanks for your reply. The ad unit parameters specified in our .md file https://github.com/prebid/Prebid.js/pull/11035/files#diff-90ac793c4a36454e4f7db480451a7cc2a69ebec68cf5b06108f8d3854b85e6c1 are workable test endpoint that can be used for validation if needed. 

Wouldn't that suffice?

patmmccann:
Generally setting an id is a very lightly enforced rules violation, in that bid modules are supposed to bid, not provide redundant functionality to shared id. Other adapters do this, but it may all get removed in a future release. It is generally a bad practice and we generally only allow it with explicit publisher opt in. Please document a check a configuration parameter before doing this, or perhaps just reconsider why you're doing something like this. It doesn't help anyone and just puts publishers in legal jeopardy; if publishers want to set a simple id like this they can do so in a variety of ways you'll find on the eids object.

modules/rixengineBidAdapter.md
```diff
@@ -0,0 +1,33 @@
...
+    sizes: [
+        [320, 50]
+    ],     
+    bids: [{
+      bidder: 'rixengine',
+      params: {
+        endpoint: 'http://demo.svr.rixengine.com/rtb', // required
+        token: '1e05a767930d7d96ef6ce16318b4ab99', // required
+        sid: 36540, // required
+        bidfloor: '0.5', // Optional
```
patmmccann, line 28: It actually seems like your adapter never looks at this; please write a unit test for it or fix your documentation

patmmccann (changes requested):

patmmccann (approved):

modules/rixengineBidAdapter.md
BaiShuoyang: @patmmccann Done, please review.

BaiShuoyang:
> Generally setting an id is a very lightly enforced rules violation, in that bid modules are supposed to bid, not provide redundant functionality to shared id. Other adapters do this, but it may all get removed in a future release. It is generally a bad practice and we generally only allow it with explicit publisher opt in. Please document and check a configuration parameter before doing this, or perhaps just reconsider why you're doing something like this. It doesn't help anyone and just puts publishers in legal jeopardy; if publishers want to set a simple id like this they can do so in a variety of ways you'll find on the eids object.

@patmmccann Noted, we have removed the usage of user ID completely. Please review.

PR #11054, opened 2024-02-05
Rayn RTD Module: initial release

aleksandar-rayn:
## Type of change

- [x] Feature

## Description of change

Rayn Real-time Data Submodule

### DIFF
--- a/modules/raynRtdProvider.js
+++ b/modules/raynRtdProvider.js
@@ -1,0 +1,198 @@
+/**
+ * This module adds the Rayn provider to the real time data module
+ * The {@link module:modules/realTimeData} module is required
+ * The module will fetch real-time audience and context data from Rayn
+ * @module modules/raynRtdProvider
+ * @requires module:modules/realTimeData
+ */
+
+import { MODULE_TYPE_RTD } from '../src/activities/modules.js';
+import { submodule } from '../src/hook.js';
+import { getStorageManager } from '../src/storageManager.js';
+import { deepAccess, deepSetValue, logError, logMessage, mergeDeep } from '../src/utils.js';
+
+const MODULE_NAME = 'realTimeData';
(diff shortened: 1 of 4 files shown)


modules/raynRtdProvider.js
```diff
@@ -0,0 +1,198 @@
...
+    } else if (window.raynJS && typeof window.raynJS.getSegtax === 'function') {
+      window.raynJS.getSegtax().then((segtaxData) => {
+        logMessage(LOG_PREFIX, `Segtax data from RaynJS: ${JSON.stringify(segtaxData)}`);
+        setSegmentsAsBidderOrtb2(reqBidsConfigObj, bidders, integrationConfig, segtaxData, checksum);
+        callback();
+      }).catch((error) => {
+        logError(LOG_PREFIX, error);
+        callback();
+      });
+    } else {
```
patmmccann, line 181: why not fall back to communicating with your endpoint from the module?

aleksandar-rayn: @patmmccann I'm not sure I understand what you mean exactly.
We have two steps, first we check for data in localStorage, then if our library is loaded we call it's function to retrieve the data and if none of the conditions are satisfied we just log a "no data" message and don't do anything.

patmmccann: I see that. How about a third step; if neither local storage nor the library exist, you communicate with an endpoint to gather the segments?

aleksandar-rayn: @patmmccann thank you for your observation. We may be adding this functionality at a later stage. For now we are relying on our customers to either run our library or populate local storage using our endpoints

patmmccann (approved):

modules/raynRtdProvider.js
patmmccann: Please make sure your documentation reflects that

PR #11061, opened 2024-02-06
GumGumBidAdapter: Fix size in the bid response for multi size slot sizes

MartinGumGum:
- [ x] Feature

## Description of change
Fix for multi-size in-slot in our bid response that says that the size of this ad is 300x600 but adapter is interpreting this as a 320x50.


MartinGumGum:
Please review @ChrisHuie

PR #11063, opened 2024-02-06
GumGumBidAdapter: Fix size in the bid response for multi size slot sizes 

MartinGumGum:
- [ x] Feature

## Description of change
In our bid response that says this size of this ad is 300x600 but Prebid adapter is interpreting this as a 320x50.


MartinGumGum:
Hi @ChrisHuie , Can you please review this PR? 
Thx in advance!

PR #11064, opened 2024-02-06
GumGum Bid Adapter: fix size in the bid response for multi size slot sizes 

MartinGumGum:
- [ x] Feature

## Description of change
In our bid response, it says the size of this ad is 300x600 but the adapter is interpreting this as a 320x50.

### DIFF
--- a/modules/gridBidAdapter.js
+++ b/modules/gridBidAdapter.js
@@ -361,6 +361,16 @@ export const spec = {
         request.regs.coppa = 1;
       }
 
+      if (ortb2Regs?.ext?.dsa) {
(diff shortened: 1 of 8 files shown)


MartinGumGum:
@ChrisHuie  Not sure why this is failing? in my Local it is not. Could you please review it? I  have opened and reopened a couple of times.

MartinGumGum:
Looks like e2e testing needs to be rerun but I can't since the button is not active for me, maybe I don't have permission in circleci?

ChrisHuie (approved):

PR #11075, opened 2024-02-07
Euid id module: cstg opt out enforcement

ssundahlTTD:
## Type of change

- [x] Feature

## Description of change
This change applies to Client Side Token Generation (CSTG) for EUID.  The EUID /client-generate endpoint now expects an optout_check parameter.  This parameter is expected to be 1 for EUID and will be defaulted as such.  Previously the endpoint returned an "optout token" for an opted out user.  It will now return an optout flag of true and no token information.

### DIFF
--- a/gulpfile.js
+++ b/gulpfile.js
@@ -88,7 +88,7 @@ function lint(done) {
     '!plugins/**/node_modules/**',
     './*.js'
   ], { base: './' })
-    .pipe(gulpif(argv.nolintfix, eslint(), eslint({ fix: true })))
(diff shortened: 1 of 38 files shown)


sunnywu:
Hi @ssundahlTTD is there any handling difference from publisher perspective and if [our integration guide](https://unifiedid.com/docs/guides/integration-prebid-server-side) need to be updated because of this functional changes? (Gen is writing up the EUID prebid.js docs now so if there are any changes we need to update the guide for we should inform her now)

lionell-pack-ttd (approved):

mmoschovas (approved):

PR #11077, opened 2024-02-08
PBjs Core Utils: fix deepEqual() to work correctly on sites where Array.prototype has been extended

samuel-palmer-relevant-digital:
## Type of change

- [x] Bugfix
## Description of change

This PR solves a problem that `utils.deepEqual` incorrectly returns false when `Array.prototype` (or `Object.prototype`) has been extended in any way - when comparing arrays/objects. A practical implication is that the **priceFloors** module doesn't work and gives errors on the form `adUnit '??' declares a different schema from one previously declared by adUnit ...`, on such sites.

The problem is that `for (let prop in obj1)` _includes_ the properties added in the prototype while the next line `obj2.hasOwnProperty(prop)` will return false. The solution is to only do the check when `obj1.hasOwnProperty(prop)` is also true.

### DIFF
--- a/src/utils.js
+++ b/src/utils.js
@@ -892,16 +892,15 @@ export function deepEqual(obj1, obj2, {checkTypes = false} = {}) {
     (typeof obj2 === 'object' && obj2 !== null) &&
     (!checkTypes || (obj1.constructor === obj2.constructor))
   ) {
-    if (Object.keys(obj1).length !== Object.keys(obj2).length) return false;
-    for (let prop in obj1) {
-      if (obj1.hasOwnProperty(prop)) {
-        if (obj2.hasOwnProperty(prop)) {
-          if (!deepEqual(obj1[prop], obj2[prop], {checkTypes})) {
-            return false;
-          }
-        } else {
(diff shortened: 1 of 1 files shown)


src/utils.js
```diff
@@ -892,8 +892,9 @@ export function deepEqual(obj1, obj2, {checkTypes = false} = {}) {
     (typeof obj2 === 'object' && obj2 !== null) &&
     (!checkTypes || (obj1.constructor === obj2.constructor))
   ) {
-    if (Object.keys(obj1).length !== Object.keys(obj2).length) return false;
-    for (let prop in obj1) {
+    const props1 = Object.keys(obj1);
+    if (props1.length !== Object.keys(obj2).length) return false;
+    for (let prop of props1) {
```
dgirardi, line 897: Take this as a suggestion (I don't know to to suggest changes for code outside the diff, so I made it a commit). Since we already run Object.keys I think this should save a couple cpu cycles.

dgirardi (approved):

patmmccann (approved):

PR #11082, opened 2024-02-08
Build system: add --no-lint-warnings option

dgirardi:
## Type of change

- [x] Build related changes

## Description of change

Add the option to turn off linter warnings with `--no-lint-warnings`.

### DIFF
--- a/gulpfile.js
+++ b/gulpfile.js
@@ -88,7 +88,7 @@ function lint(done) {
     '!plugins/**/node_modules/**',
     './*.js'
   ], { base: './' })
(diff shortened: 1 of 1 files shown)


someden:
@ChrisHuie there is no nullish coalescing in node 12, so now build is broken on `https://docs.prebid.org/download.html`
And in devcontainers:
```
$ node -v
v12.22.10
$ gulp build
/workspaces/Prebid.js/gulpfile.js:91
    .pipe(eslint({ fix: !argv.nolintfix, quiet: !(argv.lintWarnings ?? true) }))
                                                                     ^
SyntaxError: Unexpected token '?'
```

Issue #10788, opened 2023-12-04
Prebid.js client side loss notifications, adding info to future events or own event?


senaev:
## Type of issue
question

## Description

There was a question about lose notification in 2020
https://github.com/prebid/Prebid.js/issues/5960#issue-741503747

Has something been changed since then?
Does bidder have a possibility to get feedback in case of lost auction (for example like `lurl` in OpenRTB)

patmmccann:
This isn't currently a feature, partial dupe of #4493

patmmccann:
We have rules against obtaining information on other parties

eg "Bidder modules must not obtain bid information from or about any other party in the auction. E.g., they cannot listen to ad server events and forward information naming other bidders back to their endpoint - that is the job of an analytics module."

Currently the plan is to develop either #4493 or add loss notifications to future events, including via fetch with keepalive which outlives the page.

patmmccann:
We can pass in the future bid request a json with the following four fields as a new top level object on request, or somewhere in an ext.

{ bidderRequestId: 123,
auctionId: imp.ext.tid, -- might be null 
minBidToWin, the bid needed to win this auction.
rendered: 1/0 -- did prebid render a bid on this auction
}

senaev:
> We can pass in the future bid request a json with the following four fields as a new top level object on request, or somewhere in an ext.

How can I subscribe to updates about this feature request? Is there any thread?

patmmccann:
we'll keep this one open

pm-harshad-mane:
Each SSP should be able to mention how they want to receive these notifications, either in following requests or fired immediately to their end-points. 
Both approaches need different implementations... 
- to send the data in following requests we will need to store dat in first party 
- to send data asap, we can use JS Beacon API

**Why JS Beacon?**
The Beacon API is used to send an asynchronous and non-blocking request to a web server. The request does not expect a response. Unlike requests made using XMLHttpRequest or the Fetch API. **The browser guarantees to initiate beacon requests before the page is unloaded and to run them to completion.** https://developer.mozilla.org/en-US/docs/Web/API/Beacon_API

jlquaccia:
Created a PR for this ticket.  I created a new "lossNotifications" module in core that can optionally be used internally within bid adapters.  When enabled, the module supports 2 methods for issuing out loss notifications to bidders:

[...]

PR #11086, opened 2024-02-09
New Module: MinBidToWin Notifications: Created a new module to support sending minbidtowin notifications to bidders

jlquaccia:
## Type of change

- [x] Feature

## Description of change
- Created a new "lossNotifications" module in core that can optionally be used internally within bid adapters
- When enabled, the module supports 2 methods for issuing out loss notifications to bidders:
1. By default, loss notifications will be sent out to respective bidders in subsequent auctions within a bidders outgoing bid request (for example, bidder A and bidder B are participating in auction 1.  bidder A wins.  during auction 2 (or next auction that bidder B is involved in), bidder B will receive the loss notification within their bid request.
2. If a bidder provides an endpoint of their own, they can send loss notifications about their bids to their own endpoint via the JS Beacon API: https://developer.mozilla.org/en-US/docs/Web/API/Beacon_API

## Other information
https://github.com/prebid/Prebid.js/issues/10788


jlquaccia:
https://github.com/prebid/Prebid.js/issues/10788#issuecomment-1936653860

will write tests for this PR once the questions within the comment above are addressed

src/lossNotifications.js
```diff
@@ -0,0 +1,97 @@
...
+      const lossNotificationPayload = {
+        bidderRequestId: bidder.bidderRequestId,
+        auctionId: winningBidData.transactionId,
+        minBidToWin: winningBidData.cpm,
+        rendered: winningBidData.status === CONSTANTS.BID_STATUS.RENDERED ? 1 : 0,
+      };
+
+      if (bidder.beaconUrl) {
+        // use  js beacon api to fire payload immediately to ssp provided endpoint
+        navigator.sendBeacon(bidder.beaconUrl, JSON.stringify(lossNotificationPayload));
```
patmmccann, line 69: please use our methods

bretg:
Would suggest naming this "auction info" instead of trying to promise true "loss notification". https://github.com/prebid/prebid-server/issues/3333

patmmccann:
One thought in committee is keeping direct calls for the unload event; we would attempt to send all notifications on the next payload.

patmmccann:
Some notes in committee discussion https://docs.google.com/document/d/1AojPDLnJzXaxwmFE5U7Qsm2GAV1fVwjcx4YyI6kX9ew/edit

Looking forward to this feature!

src/lossNotifications.js
```diff
@@ -0,0 +1,97 @@
...
+  tidMap[winningBidData.transactionId].forEach(bidder => {
+    if (bidder.bidderCode !== winningBidData.bidderCode) {
+      const lossNotificationPayload = {
+        bidderRequestId: bidder.bidderRequestId,
+        auctionId: winningBidData.transactionId,
+        minBidToWin: winningBidData.cpm,
+        rendered: winningBidData.status === CONSTANTS.BID_STATUS.RENDERED ? 1 : 0,
+      };
+
+      if (bidder.beaconUrl) {
```
lcorrigall, line 67: Using the lurl, or any other url to automatically fire a response back needs to be removed. We don't want to double the network traffic from the page. We want to strongly encourage bidders to piggyback this data on a subsequent bid request.

mlapeyre3:
Hey, chiming in to check what is required for this PR to be approved? As a Prebid Server bidder, we are very interested with this feature as it was proven very efficient with other partners in terms of optimizing our bidding behavior and winrate.
Thanks!

patmmccann:
I think it is important to try and censure bidder B listening to scenarios they have submitted valid bids. We dont want to incentivize invalid bid submissions to get these events

src/lossNotifications.js
patmmccann: I agree; I'm not sure how to do it though, but if we somehow knew subsequent bid event was not ever going to happen, eg on page change, we could queue the beacons with keepalive to survive the unload event

bretg:
Turns out the Prebid Server committee has been discussing this general topic as well. See https://github.com/prebid/prebid-server/issues/3333 . The scenarios are different, but it might make sense to standardize on the ORTB extension here.

Haven't done a full analysis, but did note some key points of difference:
- "fire payload immediately to ssp provided endpoint" - Prebid Server can't do this. And I think doing this in Prebid.js could be a concern to the Sustainability committee since it could nearly double the amount of traffic generated by the browser. Another solution would be to piggyback previous auction data on the next auction.  Heads up @GLStephen .
- Assuming the piggyback solution is what's implemented I'd suggest not calling this thing "loss notifications" or even "minBidToWin". It's great that Prebid.js can utilize the render event. Prebid Server cannot. However, a bid might lose the first auction but later be pulled out of the bid cache and therefore be a future winner, making the term "loss" murkier. On the server-side, we've been referring to this whole thing as "previous auction info" which sidesteps any preconceived notions about "win" or "lose".
- The proposed JSON object here is missing currency. That's a must-have.

PBJS proposed payload:
```
        bidderRequestId: bidder.bidderRequestId,
        auctionId: winningBidData.transactionId,
        minBidToWin: winningBidData.cpm,
        rendered: winningBidData.status === CONSTANTS.BID_STATUS.RENDERED ? 1 : 0,
```

PBJS proposed ORTB extension at ext.prebid.previousauctioninfo
```
source: "pbjs",
auctionId: STRING, // $.id of the previous auction
impid: STRING,       // $.imp[].id of the previous auction
bidresponseid: STRING, // seatbid[].bid[].id of the previous auction
targetedbidcpm: FLOAT,          // the bid targeted as the 'winner' within PBS targeting. Not specified if includewinners flag not present
highestcpm: FLOAT,        // the highest bid seen by Prebid in the publisher's requested currency
cur: STRING,
biddercpm: FLOAT,    // the price submitted by this bidder
biddererrorcode: INTEGER,  // if the bidder's bid was rejected, let them know the seatnonbid code
timestamp: INTEGER
```

Perhaps these approaches should be discussed in next week's committee meetings?

bretg:
@jlquaccia, @patmmccann, @mlapeyre3 - any response to my concerns?

Most importantly, I believe currency is a must-have. But also I think a "piggybacking-on-the-next auction" approach is superior to the current "fire payload immediately to ssp provided endpoint" implementation. Am ready to debate this. :-)

jlquaccia:
@bretg I'm good with the ext.prebid.previousauctioninfo approach mentioned above with the additional keys you proposed, makes sense to me.  Aligning with PBS would definitely be ideal.  I can also see why immediately firing a response would have drawbacks (as also mentioned above).

What if during subsequent auctions, for ssp's that have opted in to receive previousauctioninfo, in order for them to actually receive this info, their subsequent bid request has to pass their own bid adapter isBidRequestValid check? (to help try filter out invalid submissions)

@patmmccann @mlapeyre3 thoughts on your end?

patmmccann:
@jlquaccia we'd love to make progress on this. Do you think we're in a position to start writing tests? Can we pitch in?

patmmccann:
noting the file is in /src/ not /modules

src/lossNotifications.js
```diff
@@ -0,0 +1,97 @@
...
+      } else {
+        // store the payload in an array under bidder code in local storage and send it out with the respective bidder's next bid request
+        let pbln = JSON.parse(storage.getDataFromLocalStorage('pbln'));
+        if (pbln) {
+          if (pbln[bidder.bidderCode]) {
+            pbln[bidder.bidderCode].push(lossNotificationPayload);
+          } else {
+            pbln[bidder.bidderCode] = [lossNotificationPayload];
+          }
+          storage.setDataInLocalStorage('pbln', JSON.stringify(pbln));
```
patmmccann, line 79: why use storage instead of just an object in window?

jlquaccia:
> @jlquaccia we'd love to make progress on this. Do you think we're in a position to start writing tests? Can we pitch in?

@patmmccann thanks for the recent updates above.  yes, i can pick this up again (will aim for next week).  currently wrapping up with some internal project work over here.

jlquaccia:
@patmmccann addressed the items above and wrote some new tests.  let me know if you have any other feedback

jlquaccia:
hey @patmmccann, just wanted to follow up with this one for a code review?

libraries/previousAuctionInfo/previousAuctionInfo.js
```diff
@@ -0,0 +1,130 @@
...
+    const enabledBidder = enabledBidders.find(bidder => bidder.bidderCode === bidRequest.bidderCode);
+    window.pbpai = window.pbpai || {};
+
+    if (enabledBidder && window.pbpai[bidRequest.bidderCode]) {
+      window.pbpai[bidRequest.bidderCode].forEach(prevAuctPayload => {
+        if (winningBidsMap[prevAuctPayload.transactionId]) {
+          prevAuctPayload.minBidToWin = winningBidsMap[prevAuctPayload.transactionId].cpm;
+          prevAuctPayload.rendered = 1;
+        }
+      });
```
dgirardi, line 118: The purpose of this is to update the prevAuctionInfo object with the winning bid's cpm, right? If so, I think it'd be more intuitive to put it in the BID_WON handler. But see also below on transactionId

libraries/previousAuctionInfo/previousAuctionInfo.js
```diff
@@ -0,0 +1,130 @@
...
+          bidderRequest.bids.forEach(bid => {
+            const isValidBid = enabledBidder.isBidRequestValid(bid);
+
+            if (!isValidBid) return;
+
+            const previousAuctionInfoPayload = {
+              bidderRequestId: bidderRequest.bidderRequestId,
+              minBidToWin: highestCpmBid?.cpm || 0,
+              rendered: 0,
+              transactionId: bid.ortb2Imp.ext.tid || bid.transactionId,
```
dgirardi, line 68: this leaks `transactionId` to the bidder, which is hidden by default. You could still use this to keep track of winners without exposing it, for example by using a map from transactionId -> prevAuctionInfo, which would make it easier to update the winning cpm on BID_WON, and pass only the values to the bidder on the next request.

libraries/previousAuctionInfo/previousAuctionInfo.js
```diff
@@ -0,0 +1,130 @@
...
+              source: 'pbjs',
+              auctionId: auctionDetails.auctionId,
+              impId: bid.adUnitCode,
```
dgirardi, line 71: IMO `adUnitCode` is a better name. `impId` makes me think of ORTB `imp.id`, which most of the time is different from the ad unit code.

libraries/previousAuctionInfo/previousAuctionInfo.js
```diff
@@ -0,0 +1,130 @@
...
+              // bidResponseId: FLOAT, // don't think this is available client side?
+              // targetedbidcpm: FLOAT, // don't think this is available client side?
+              highestcpm: highestCpmBid?.cpm || 0,
+              cur: bid.ortb2.cur,
+              bidderCpm: receivedBidsMap[bid.bidId] ? receivedBidsMap[bid.bidId].cpm : 'nobid',
+              biddererrorcode: rejectedBidsMap[bid.bidId] ? rejectedBidsMap[bid.bidId].rejectionReason : -1,
+              timestamp: auctionDetails.timestamp,
+            }
+
+            window.pbpai = window.pbpai || {};
```
dgirardi, line 81: why put this in `window`? it looks like internal state to me.

libraries/previousAuctionInfo/previousAuctionInfo.js
```diff
@@ -0,0 +1,130 @@
...
+let enabledBidders = [];
+export let winningBidsMap = {};
+
+export const resetPreviousAuctionInfo = () => {
+  previousAuctionInfoEnabled = false;
+  enabledBidders = [];
+  winningBidsMap = {};
+};
+
+export const enablePreviousAuctionInfo = (config, cb = initHandlers) => {
```
dgirardi, line 14: It's unclear to me how this is intended to be used. 

If it's an optional module chosen by the publisher, this should be in the `modules` folder and listen for changes to `setConfig`.

Alternatively we could let individual bidders opt in, in which case the only parameter needed should be the bidder code, and this should allow for being called multiple times with different bidders.

libraries/previousAuctionInfo/previousAuctionInfo.js
```diff
@@ -0,0 +1,130 @@
...
+
+            if (!isValidBid) return;
+
+            const previousAuctionInfoPayload = {
+              bidderRequestId: bidderRequest.bidderRequestId,
+              minBidToWin: highestCpmBid?.cpm || 0,
+              rendered: 0,
+              transactionId: bid.ortb2Imp.ext.tid || bid.transactionId,
+              source: 'pbjs',
+              auctionId: auctionDetails.auctionId,
```
dgirardi, line 70: Same issue as `transactionId` - this is hidden by default. This should either be omitted, or included only when [the `transmitTid` activity is allowed](https://github.com/prebid/Prebid.js/blob/a9958be99d7c9d431fe3d22929daf834009f3480/src/adapters/bidderFactory.js#L194)

libraries/previousAuctionInfo/previousAuctionInfo.js
```diff
@@ -0,0 +1,130 @@
...
+        const enabledBidder = enabledBidders.find(bidder => bidder.bidderCode === bidderRequest.bidderCode);
+
+        if (enabledBidder) {
+          bidderRequest.bids.forEach(bid => {
+            const isValidBid = enabledBidder.isBidRequestValid(bid);
+
+            if (!isValidBid) return;
+
+            const previousAuctionInfoPayload = {
+              bidderRequestId: bidderRequest.bidderRequestId,
```
bretg, line 65: What is bidderRequest.bidderRequestId?

libraries/previousAuctionInfo/previousAuctionInfo.js
```diff
@@ -0,0 +1,130 @@
...
+              minBidToWin: highestCpmBid?.cpm || 0,
```
bretg, line 66: How does 'minBidToWin' differ from the 'highestcpm' field? I can live with this term because it's sort of an industry standard, but it's actually not correct. Not guaranteed at all that they would have won with this bid. It just happens to be the highest bid that Prebid knows about. Which is why I prefer the term 'highestcpm' -- there's less of a promise involved. But anyhow, I get that 'minBidToWin' is a cow that's left the barn, so I would suggest removing 'highestcpm' then. Please make sure that it's positioned right next to the `cur` field though. Another pet peeve of mine is when it's not clear what currency a price is stated in.

Which I suppose opens a can of worms. The obvious minBidToWin currency is the ortb.cur, but there are going to bidders that will only accept minBidToWin in USD, EUR, or whatever. Do we need this module to convert to different currencies, so an array of minBidToWin entries?

libraries/previousAuctionInfo/previousAuctionInfo.js
```diff
@@ -0,0 +1,130 @@
...
+              rendered: 0,
+              transactionId: bid.ortb2Imp.ext.tid || bid.transactionId,
+              source: 'pbjs',
+              auctionId: auctionDetails.auctionId,
+              impId: bid.adUnitCode,
+              // bidResponseId: FLOAT, // don't think this is available client side?
+              // targetedbidcpm: FLOAT, // don't think this is available client side?
```
bretg, line 73: Yes, targetedbidcpm is available in PBJS. If deals are prioritized over open market, the `highestcpm` might not be the one that Prebid chose to send in `hb_pb`.

libraries/previousAuctionInfo/previousAuctionInfo.js
bretg: I think it should be debated whether any of this data is useful to bidders if there's no IDs. If I understand the purpose correctly, the receiver will join these logs with their bid logs so they can do stuff like "oh, when the browser is X and the country is Y, my bid has to be higher". In order to do that, I suspect that there's a minimal set of IDs or the join will be impossible.

dgirardi: we can leave in `requestId` and `bidderRequestId`. They have the same domain (allow the same join) as `transactionId` / `auctionId`, the difference is they're not shared across bidders.

Or we can do what I think the first join would have to be - each request's prevAuctionInfo only pertains bids on the same ad unit.

libraries/previousAuctionInfo/previousAuctionInfo.js
jlquaccia: Yes, that latter is what we are aiming for.  i was thinking any bid adapter that wants to enable the feature could simply import the previousAuctionInfo module and enable it with their bidder code:

```
import { enablePreviousAuctionInfo } from '../libraries/previousAuctionInfo/previousAuctionInfo.js';

enablePreviousAuctionInfo({ bidderCode: 'rise' });
```

libraries/previousAuctionInfo/previousAuctionInfo.js
jlquaccia: Just the id of an ssp's bid request: https://docs.prebid.org/dev-docs/bidder-adaptor.html#bidderrequest-parameters

libraries/previousAuctionInfo/previousAuctionInfo.js
```diff
@@ -0,0 +1,130 @@
...
+      });
+    }
+
+    if (auctionDetails.bidderRequests && auctionDetails.bidderRequests.length) {
+      auctionDetails.bidderRequests.forEach(bidderRequest => {
+        const enabledBidder = enabledBidders.find(bidder => bidder.bidderCode === bidderRequest.bidderCode);
+
+        if (enabledBidder) {
+          bidderRequest.bids.forEach(bid => {
+            const isValidBid = enabledBidder.isBidRequestValid(bid);
```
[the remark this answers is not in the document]
jlquaccia: yea to your point, i reviewed the logic around isBidRequestValid a bit more.. when fetching pbjs auction end data, invalid bidrequests will already have been filtered out right?  As seen here: https://github.com/prebid/Prebid.js/blob/master/src/adapters/bidderFactory.js#L360

so i removed the logic i had earlier where i was pulling in bid adapter isBidRequestValid functions.  the original intent was just to filter out bidders from receiving previous auction info if they had submitted invalid bids in the previous auction

libraries/previousAuctionInfo/previousAuctionInfo.js
jlquaccia: i agree with you, i think "highestcpm" makes the most sense and is the safest.  for currency, maybe just capturing the currency values from the bid stream (cur of the bidderCpm, highestBidCpm, etc.) would be good enough?  and then the bid adapters who opt into this module can just process the cur data however they need to?

jlquaccia:
@dgirardi @bretg, addressed your feedback with the latest round of commits

libraries/previousAuctionInfo/previousAuctionInfo.js
```diff
@@ -27,101 +32,102 @@ export const initHandlers = () => {
...
-            }
-
-            window.pbpai[bidderRequest.bidderCode].push(previousAuctionInfoPayload);
+            auctionState[bidderRequest.bidderCode].push(previousAuctionInfoPayload);
           });
         }
       });
     }
+    // eslint-disable-next-line no-console
+    console.log('onAuctionEndHandler: ', { auctionState });
```
dgirardi: are these intentional? there's several of them.

libraries/previousAuctionInfo/previousAuctionInfo.js
```diff
@@ -27,101 +32,102 @@ export const initHandlers = () => {
...
-    window.pbpai = window.pbpai || {};
-
-    if (enabledBidder && window.pbpai[bidRequest.bidderCode]) {
-      window.pbpai[bidRequest.bidderCode].forEach(prevAuctPayload => {
+    if (enabledBidder && auctionState[bidRequest.bidderCode]) {
+      auctionState[bidRequest.bidderCode].forEach(prevAuctPayload => {
         if (winningBidsMap[prevAuctPayload.transactionId]) {
-          prevAuctPayload.minBidToWin = winningBidsMap[prevAuctPayload.transactionId].cpm;
+          // prevAuctPayload.targetedBidCpm = winningBidsMap[prevAuctPayload.transactionId]?.adserverTargeting.hb_pb;
+          // prevAuctPayload.highestBidCpm = winningBidsMap[prevAuctPayload.transactionId]?.cpm;
```
dgirardi: how about these?

jlquaccia:
Hi @dgirardi, I pushed a few other commits after some additional feedback from the PMC call earlier today.  Can you review again when you get a chance?  Thanks!

patmmccann (approved):

test/spec/libraries/previousAuctionInfo_spec.js
```diff
@@ -0,0 +1,188 @@
...
+        // eslint-disable-next-line standard/no-callback-literal
+        callback({ previousAuctionInfo: true });
+      } else {
+        return origGetConfig.apply(config, arguments);
+      }
+    });
+
+    previousAuctionInfo.resetPreviousAuctionInfo();
+    initHandlersStub = sandbox.stub();
+  });
```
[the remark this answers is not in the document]
jlquaccia: thanks!  i switched to using setConfig instead.

libraries/previousAuctionInfo/previousAuctionInfo.js
```diff
@@ -0,0 +1,121 @@
...
+
+export const resetPreviousAuctionInfo = () => {
+  previousAuctionInfoEnabled = false;
+  enabledBidders = [];
+  auctionState = {};
+};
+
+export const enablePreviousAuctionInfo = (sspConfig, cb = initHandlers) => {
+  config.getConfig('previousAuctionInfo', (conf) => {
+    if (!conf.previousAuctionInfo) return;
```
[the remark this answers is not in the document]
jlquaccia: this makes sense.  i went back and refactored things.. i am still using this listener `config.getConfig('previousAuctionInfo', (conf) => {` but i removed it from the `enablePreviousAuctionInfo` function.  now i believe it should only get called once.  does this approach sound good?  otherwise i can refactor further if you want

dgirardi (approved):

jlquaccia:
thanks @dgirardi @patmmccann!  are we good to merge now?

krzysztofequativ:
@jlquaccia, I have issues to use the module so I have opened the issue https://github.com/prebid/Prebid.js/issues/12825.

PR #11101, opened 2024-02-14
greenbids analytics adapter force split determinist sampling

jbogp:
## Type of change

- [x] Bugfix
- [x] Feature

## Description of change
Small feature to fix a bug on Greenbids side to have a fixed split in the deterministic sampling.

A debug parameter flag has been added to to module but doesn't require documentation update as the parameter shouldn't be directly changed by end users.

### DIFF
--- a/modules/greenbidsAnalyticsAdapter.js
+++ b/modules/greenbidsAnalyticsAdapter.js
@@ -27,18 +27,21 @@ export const BIDDER_STATUS = {
 
 const analyticsOptions = {};
 
-export const isSampled = function(greenbidsId, samplingRate) {
+export const isSampled = function(greenbidsId, samplingRate, explorationSplitRate) {
+  // We want a split of analytics traffic outside rtd exploration
+  let explorationSplitSampled = Math.random() < explorationSplitRate;
   if (samplingRate < 0 || samplingRate > 1) {
     logWarn('Sampling rate must be between 0 and 1');
     return true;
   }
   const hashInt = parseInt(greenbidsId.slice(-4), 16);
-  return hashInt < samplingRate * (0xFFFF + 1);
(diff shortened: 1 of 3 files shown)


modules/greenbidsAnalyticsAdapter.js
```diff
@@ -27,18 +27,21 @@ export const BIDDER_STATUS = {
...
+  // We want a split of analytics traffic outside rtd exploration
+  let explorationSplitSampled = Math.random() < explorationSplitRate;
   if (samplingRate < 0 || samplingRate > 1) {
     logWarn('Sampling rate must be between 0 and 1');
     return true;
   }
   const hashInt = parseInt(greenbidsId.slice(-4), 16);
-  return hashInt < samplingRate * (0xFFFF + 1);
+  return hashInt < samplingRate * (0xFFFF + 1) && explorationSplitSampled;
 }
```
AlexisBRENON, line 39: suggestion: This is unclear...
1. Maybe add parens to disambiguate operator precedence.
2. Split logic in two separate steps
3. Group variables definitions together, and separate them from pre-condition checks.

```suggestion
export const isSampled = function(greenbidsId, samplingRate, explorationSplitRate) {
    if (samplingRate < 0 || samplingRate > 1 ||
        explorationSplitRate < 0 || explorationSplitRate > 1) {
      logWarn('...');
      return true;
    }

    const isRtdExploration =  parseInt(greenbidsId.slice(-4), 16) < samplingRate * (0xFFFF + 1);
    const isAnalyticsSampling = Math.random() < explorationSplitRate;  // Why not use greenbidsId here ? greenbidsId.slice(-6, -4) ?
    return (isRtdExploration && !isAnalyticsSampling) || (!isRtdExploration && isAnalyticsSampling);
}
```

Issue #5097, opened 2020-04-08
Possibility of Typescript support


snapwich:
## Description
This ticket is to track interest in Typescript types for Prebid.js. Let me know if you think types would be useful in the ways that you use Prebid.js.

My initial thought is that we could start by typing the bid adapter interface as both a complement the bid adapter unit tests as well as being additional documentation of that interface (would most likely stay up-to-date better than jsdoc types since it's verifiable with the typescript compiler).

Later on down the line we could choose type things such as the public API (methods on the `pbjs` global). I doubt we'd ever start typing internal core code as there is much less value there and it would require a lot more effort.

Some things to consider in a typescript implementation:
* Prebid.js will probably not be rewritten in typescript as that is a huge refactor.
* Types could be added as ambient declarations within the prebid repo as `.d.ts` files that live along the relevant `.js` files OR we could maintain a separate types repo within definitelytyped or in the prebid organization.
* Bid adapters could be submitted as `.ts` files and we could update our webpack build to accommodate building typescript files. This would be considered a backwards breaking change since it would break webpack builds outside of the prebid repo (such as people using prebid.js as an npm dependency). Since it's backwards breaking we probably wouldn't do this until a major release down the line, which could take awhile if we choose to go this route.
* We could accept bid adapters that are pre-compiled to `.js` files as bid adapter submissions and the we could choose to either allow the `.ts` source files to live in this repo or require you to manage your own `.ts` source files and only allow generated code in this repo. Since we're still only including the `.js` code in the webpack build, this is not a backwards breaking change and could be implemented right away.

A proof of concept of the last method (submitting generated javascript with typescript source alongside) has been opened as a pull-request here: https://github.com/prebid/Prebid.js/pull/5056

If there's some other way of using types not listed here that you think could be beneficial please feel free to comment.

## Other information
Previous ticket related to types: https://github.com/prebid/Prebid.js/issues/1122

muuki88:
Thanks @snapwich for the detailed explanation. We use typescript for all our ad tags we build and provide our own set of types for prebid.js. 

> Prebid.js will probably not be rewritten in typescript as that is a huge refactor.

Airbnb released an interesting tool that might make this more feasible.

https://github.com/airbnb/ts-migrate

> OR we could maintain a separate types repo within definitelytyped or in the prebid organization.

I would heavily favor hosting this inside the prebid repository. This ties the types much closer to the actual releases

[...]

PR #11110, opened 2024-02-19
Yandex: use ortb2 info & Core: add webdriver flag

chernodub:
## Type of change

- [ ] Bugfix
- [x] Feature
- [ ] New bidder adapter   
- [ ] Build related changes
- [ ] CI related changes

- [ ] Other

## Description of change

Used the additional user/device information provided by Prebid to be able to increase the value of the ads for the users.

### DIFF
--- a/modules/yandexBidAdapter.js
+++ b/modules/yandexBidAdapter.js
@@ -172,10 +172,7 @@ export const spec = {
         site: ortb2?.site,
         tmax: timeout,
         user: ortb2?.user,
-        device: {
-          ...ortb2?.device,
-          ext: getDeviceExt(),
-        },
+        device: ortb2?.device,
       };
 
       const eids = deepAccess(bidRequest, 'userIdAsEids');
@@ -479,13 +476,6 @@ function addRTT(url, rtt) {
   url = urlObj.toString();
 
   return url;
-}
-
-/** @returns {{webdriver: boolean}} */
-function getDeviceExt() {
-  return {
-    webdriver: self.navigator?.webdriver === true
-  };
 }
 
 registerBidder(spec);
--- a/modules/yandexBidAdapter.js
+++ b/modules/yandexBidAdapter.js
@@ -1,8 +1,8 @@
-import { formatQS, deepAccess, deepSetValue, triggerPixel, _each, _map } from '../src/utils.js';
 import { registerBidder } from '../src/adapters/bidderFactory.js';
-import { BANNER, NATIVE } from '../src/mediaTypes.js'
-import { convertOrtbRequestToProprietaryNative } from '../src/native.js';
 import { config } from '../src/config.js';
+import { BANNER, NATIVE } from '../src/mediaTypes.js';
+import { convertOrtbRequestToProprietaryNative } from '../src/native.js';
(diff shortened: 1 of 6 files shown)


src/types/ortb2.d.ts
```diff
@@ -0,0 +1,41 @@
+export namespace Ortb2 {
```
chernodub, line 1: I've allowed myself to slightly improve the types for ortb 2 object in bidder request (according to the [docs](https://docs.prebid.org/dev-docs/bidder-adaptor.html#note-on-ortb-adapters-1:~:text=across%20ad%20units.-,ortb2,-is%20the%20global)). If this change goes against your practices, I could send it via an additional PR or just discard. You are welcome to let me know, thanks

modules/yandexBidAdapter.js
```diff
@@ -493,4 +481,11 @@ function addRTT(url, rtt) {
   return url;
 }
 
+/** @returns {{webdriver: boolean}} */
+function getDeviceExt() {
+  return {
+    webdriver: self.navigator?.webdriver === true
```
patmmccann, line 487: please add this to core, not just your adaper. perhaps, device.ext.webdriver

patmmccann: please only populate if true in core, you're selcome to make it look however you want on your payload

patmmccann: perhaps https://github.com/prebid/Prebid.js/blob/ff58535df6f60c2289e73ea4b832aef8afed046e/src/fpd/enrichment.js#L112

src/types/ortb2.d.ts
```diff
@@ -0,0 +1,41 @@
...
+  type Regs = {
+    coppa?: unknown;
+    ext?: { gdpr?: unknown; us_privacy?: unknown };
+  };
+
+  type User = {
+    ext?: unknown;
+  };
+
+  type Ortb2Object = {
```
patmmccann, line 35: this isnt correct, ortb object might have site, dooh, or app, but not more than one

src/types/ortb2.d.ts
```diff
@@ -0,0 +1,41 @@
...
+    w?: number;
+    h?: number;
+    dnt?: 0 | 1;
+    ua?: string;
+    language: string;
+    sua?: {
+      source?: number;
+      platform: unknown;
+      browsers: unknown[];
+      mobile: number;
```
patmmccann, line 22: these 3 dont have the '?' ?

modules/yandexBidAdapter.js
chernodub: Will do, thanks for the suggestion!

src/types/ortb2.d.ts
chernodub: Sure, I'll add the props `dooh`, `app`. I couldn't find their exact definition based on the usage, so I'll just use `unknown`. If needed, these could definitely be extended in the future.

patmmccann: i think #5218 kind of covers it

patmmccann (approved):

src/types/ortb2.d.ts
patmmccann: #5097 actually...

PR #11133, opened 2024-02-23
Relaido Bid Adapter: add pagekvt to api request

relaido:
## Type of change

- [ ] Bugfix
- [x] Feature
- [ ] New bidder adapter   
- [ ] Build related changes
- [ ] CI related changes

- [ ] Other

## Description of change
Add pagekvt to API requests


modules/relaidoBidAdapter.js
```diff
@@ -349,6 +352,44 @@ function getBannerSizes(bidRequest) {
   return parseSizesInput(sizes).join(',');
 }
 
+function getTargeting(bidRequest) {
```
patmmccann, line 355: I think the pre-auction module does most of this work

relaido: @patmmccann 
I have some code that is a bit similar, but this is not what I want.
https://github.com/prebid/Prebid.js/blob/master/src/secureCreatives.js#L164-L171

spotxslagle (approved):
LGTM

relaido:
@patmmccann @spotxslagle 
What should I do to take the next step?

PR #11156, opened 2024-02-29
Adnuntius Bid Adapter: Bugfix - If no userId is passed the request fails.

mikael-lundin:
## Type of change

- [x] Bugfix
- [ ] Feature
- [ ] New bidder adapter   
- [ ] Build related changes
- [ ] CI related changes

- [ ] Other

## Description of change
Currently there's a bug in the bidder where the request fails if no user id is passed as a parameter in the bid. This fixes that.


mikael-lundin:
Seems to be a test that fails that's not related to us. please do let me know when I should pull the latest master. :)

PR #11159, opened 2024-02-29
lockr AIM user module: initial release

KEPlockr:
## Type of change

- [ ] Bugfix
- [x] Feature
- [ ] New bidder adapter   
- [ ] Build related changes
- [ ] CI related changes

- [ ] Other

## Description of change
The following PR adds the lockr AIM module as a part of prebid.

Alternative Identity Manager (AIM) is a unified container for identity and data management. 

With AIM’s self-service platform, publishers seamlessly integrate and activate alternative IDs like LiveRamp’s Authenticated Traffic Solution (ATS), Unified ID 2.0 (UID2), ID5 and more. The burden of due diligence and maintenance, coupled with the benefits of server-side calls result in the adoption of multiple alternative IDs, clean rooms like InfoSum and CDPs like Blueconic based on their specific needs.


test/spec/modules/lockrAIMIdSystem_spec.js
```diff
@@ -0,0 +1,64 @@
...
+
+const defaultConfig = {
+  appID: "3b5a0f6c-7e91-11ec-b9c7-e330d98440a7",
+  email: "example@test.com",
+};
+
+const LIVE_RAMP_COOKIE = "_lr_env";
+const UID2_COOKIE = "_uid2_advertising_token";
+const ID5_COOKIE = "id5id";
+
```
jdwieland8282, line 16: would you consider picking up sharedid (`_sharedid`) as well?

KEPlockr: A few points of clarity here: 

- lockr's AIM solely supports authenticated user events. Even with our integration with providers who can sometimes provide an pseudonymous identifier based on other attributes of the user agent, IP, etc. - we do not support that. Any and all data which is passed via lockr to an identity partner is solely hashed PII. As such, lockr AIM does not plan on facilitating SharedID as an option within its self-service UI at this time. Should this policy change, we will address this in the future. 

- lockr's AIM does provide pipes for publishers to pass data to various endpoints (S3 buckets, Optable, InfoSum, etc.). In these integrations we do include attributes the publisher might benefit from - i.e. a first party cookie as the root of the record. In this scenario we would be happy to pick up the SharedID as the key value, but this would require some engineering work on our end, which would have to be prioritized against other features publishers are requesting. At this point in time, this is not a function which has been requested. 

- To further clarify, lockr AIM does not interface as a bidder, so there is no need for "us" to pick up SharedID for normal programmatic.

modules/lockrAIMIdSystem.js
```diff
@@ -0,0 +1,75 @@
...
+ * This module adds lockr AIM ID support to the User ID module
+ * The {@link module:modules/userId} module is required.
+ * @module modules/lockrAIMIdSystem
+ * @requires module:modules/userId
+ */
+
+import { submodule } from "../src/hook.js";
+import { logInfo, logWarn } from "../src/utils.js";
+// eslint-disable-next-line prebid/validate-imports
+import { lockrAIMGetIds } from "./lockrAIMIdSystem_shared.js";
```
patmmccann, line 11: Do you expect other modules to call these methods?

modules/lockrAIMIdSystem_shared.js
```diff
@@ -0,0 +1,110 @@
...
+                JSON.stringify(LockrAIMApiClient.expiryDateKeys)
+              );
+              if (!settings?.dropLocalStorage) {
+                localStorage.setItem(
+                  cookieitem.key_name,
+                  cookieitem.advertising_token
+                );
+              }
+              if (!settings?.dropCookie) {
+                document.cookie = `${cookieitem.key_name}=${cookieitem.advertising_token};`;
```
patmmccann, line 69: You may not interact directly with local storage nor cookies

patmmccann (changes requested):

patmmccann:
You may notice you're not passing lint checks as well

KEPlockr:
> You may notice you're not passing lint checks as well

This was caused by us following the documentation and calling "configparams", we have updated this to "config" and it addressed the issue. It was within the lockrAIMIdSystem.js

modules/lockrAIMIdSystem_shared.js
KEPlockr: @patmmccann - this was updated in a new commit. We are now calling StorageManager from Prebid Utils

modules/lockrAIMIdSystem.js
KEPlockr: No, we do not.

KEPlockr:
@jdwieland8282 - can you expand on the "needs docs" label, as all such documentation is included in the "lockrAIMIdSystem.md" file. Thank you.

jdwieland8282:
You need a docs PR against the prebid.github.io repo. This will get your documentation on to prebid.org's website

modules/lockrAIMIdSystem.js
patmmccann: Cool, Please don't make a cross-module library for methods only one module calls

patmmccann:
@KEPlockr you're not passing the linter because you arent following our code style rules

image

modules/.submodules.json
```diff
@@ -48,7 +48,8 @@
       "gravitoIdSystem",
       "freepassIdSystem",
       "operaadsIdSystem",
-      "mygaruIdSystem"
+      "mygaruIdSystem",
+      "lockrAIMIdSystem"
```
patmmccann, line 52: these are alphabetical

modules/lockrAIMIdSystem.js
```diff
@@ -0,0 +1,77 @@
...
+ * The {@link module:modules/userId} module is required.
+ * @module modules/lockrAIMIdSystem
+ * @requires module:modules/userId
+ */
+
+import { submodule } from '../src/hook.js';
+import { logInfo, logWarn } from '../src/utils.js';
+import { getStorageManager } from '../src/storageManager.js';
+// eslint-disable-next-line prebid/validate-imports
+import { lockrAIMGetIds } from './lockrAIMIdSystem_shared.js';
```
patmmccann, line 12: please eliminate this if you're not sharing code from multiple modules

chris-lockr (commented):
Why are all these modules changing? Shouldn't we only be touching lockrAIMdSystem?

patmmccann (approved):

patmmccann:
Lgtm

You still need to pr https://github.com/prebid/prebid.github.io/pulls?q=is%3Apr+is%3Aopen+lockr it seems?

Issue #10999, opened 2024-01-29
Prebid 9: raise minimum node version from 12 to 16


patmmccann:
Since Node 12 is the true minimum version, should I update package.json to that version in this PR? 
Yes, if 12 is actually required, it seems useful to reflect that in package.json

Should I just update circleci to Node 20? 
Yes, no concerns

Should the package.json min Node version be updated to 16?
No, not until prebid 9

For the next major release, it probably makes sense to update the min node version to at least Node 16 but maybe the current LTS?
Seems 16 is fine in prebid 9

_Originally posted by @patmmccann in https://github.com/prebid/Prebid.js/issues/10990#issuecomment-1915077902_

dgirardi:
Let's update to node 20

PR #11166, opened 2024-03-03
Minimal typescript integration draft #5097

muuki88:
## Type of change

- [ ] Bugfix
- [ ] Feature
- [ ] New bidder adapter   
- [x] Build related changes
- [x] CI related changes

- [ ] Other

## Description of change

This PR tries to add typescript support requested in #5097 

I'll start as small as possible I try to get everything working through the typescript compiler without adding any types whatsoever. If everything works, I'll add the first types.

- [ ] all gulp tasks use typescript files
- [ ] all gulp tasks work as expected
- [ ] all CI tasks work as expected
- [ ] karma works
- [ ] js files are renamed to ts files

## Other information

* https://www.typescriptlang.org/docs/handbook/migrating-from-javascript.html
* https://www.npmjs.com/package/gulp-typescript
* https://www.typescriptlang.org/docs/handbook/gulp.html


.nvmrc
```diff
@@ -1 +1 @@
-12.16.1
+16.17.0
```
muuki88, line 1: this one is rather huge. Not sure why prebidjs is built against an (for frontend terms) ancient version of node.

I could also upgrade to 20.x , which is the current LTS version, but went for the minimal jump to make the typescript compiler happy. The issue was that some syntax was not supported by default in node 12.x ( `??` operator to be precise )

gulpHelpers.js
```diff
@@ -76,7 +76,7 @@ module.exports = {
     try {
       var absoluteModulePath = path.join(__dirname, MODULE_PATH);
       internalModules = fs.readdirSync(absoluteModulePath)
-        .filter(file => (/^[^\.]+(\.js)?$/).test(file))
+        .filter(file => (/^[^\.]+(\.[jt]s)?$/).test(file))
```
muuki88, line 79: There are a couple of other places where we need to consider `ts` and `js` files

package.json
```diff
@@ -77,6 +78,7 @@
     "gulp-shell": "^0.8.0",
     "gulp-sourcemaps": "^3.0.0",
     "gulp-terser": "^2.0.1",
+    "gulp-typescript": "^6.0.0-alpha.1",
```
muuki88, line 81: alpha...

I also like to live dangerously

tsconfig.json
```diff
@@ -0,0 +1,11 @@
+{
+  "compilerOptions": {
+    "noImplicitAny": false,
+    "allowJs": true,
```
muuki88, line 4: This is important for a step-by-step migration

tsconfig.json
```diff
@@ -0,0 +1,11 @@
...
+    "sourceMap": true,
+    "target": "es5"
```
muuki88, line 6: I'll probably change this to something newer, because babel takes care of downgrading the js at the end.

.nvmrc
patmmccann: Please see https://github.com/prebid/Prebid.js/pull/10990

tsconfig.json
patmmccann: Some of our testing suggests es2018 is reasonable

patmmccann:
There are already some types in the project fwiw

.nvmrc
muuki88: Ah, related to #10999

muuki88:
Build is failing if typescript can't compile. This is good. 

I haven't seen this error however. Will dig a little bit into it

.nvmrc
patmmccann: can we jettison this without breaking things?

tsconfig.json
dgirardi: This attempts to migrate the whole project though, in the sense that it pipes everything through the TS compiler. 
I think it'll be easier to manage if we only run TS on modules/adapters that choose to use TS. It may be tricky to set up configuration wise but it should be possible, in the same way that a TS project can use JS libraries from npm. 

What I'm thinking is the compilation (and most of the codebase, including all of core) should be left as it is now; except for an additional step in the beginning that converts any `modules/**/*.ts` into JS.

.nvmrc
muuki88: TIL: jettison :smile: 

Falling back to 12.16.1 , typescript won't transpile files that use newer syntax. There are babel-typescript integrations, but they don't typecheck, which defeats a big purpose and the gulp integration is actually trivial.

I guess minimal typescript support is fine to wait for the prebid 9 release, though.

Or maybe I got the meaning of jettison wrong :sweat:

tsconfig.json
muuki88: I'm guessing that piping this through `tsc` may have unintentional side effects. And I agree on that. However it is far less complex in configuration and mentally. If `tcs` outputs bascially the js files as is, because the `target` and `module` is set to the system prebidjs uses, they should bascially look the same.

Tinkering with [--noEmit](https://www.typescriptlang.org/tsconfig#noEmit) or babel-tyepscript preset, fiddling in typechecking at a later point, is IMHO not worth the effort.

The worst thing that can happen is that typescript generates additional code that would not be required, because babel is creating it. If the `import` statements are left intact, the hardest part still works (module resolution).

> What I'm thinking is the compilation (and most of the codebase, including all of core) should be left as it is now; except for an additional step in the beginning that converts any `modules/**/*.ts` into JS.

I agree with you, but I try to configure `tsc` that it does not alter the `js` files and just pipes them through.

dgirardi: IMO doing this invites treating the whole project as TS, or maybe even requires it - I see have already started with migrating `prebid.ts`. The fact that it seems to mean adding useless "any" declarations is what worries me.

My approach would be to move all ts files into a separate tree and have tsc work only on that, then merge its output with "normal" sources and have the current build start there. I'd expect this to be easier than trying to police tsc in what it does with the non-ts portion of the project; all it requires is a way to tell it "here's where you find core source". We don't even necessarily need to resolve relative imports; since all ts files would be new, we could have them do `import {registerBidder} from 'core/adapters/bidderFactory'` instead of `from '../src/adapters/bidderFactory'`.

dgirardi: I'll add that if you can do this without having to add the `any` tags in core, I'll be a lot less skeptical :)

muuki88: TL;DR: `any` is temporary. `noImplicitAny` compiler option is the goal 

`any` is a crutch that will be removed. I actually had to add this for `Object.entries` because the - [ ] type is `[string, unknown]` . Unknown is like any, but you have to check everything beforehand.

This is very sensible as you have no idea what properties you are enumerating. I'm forcing typescript to believe that the devs exactly know what they are doing.🙈

The path forward is to add proper types.

I started with `prebid.js` , because this is the only hardcoded file in the gulp build 😁

muuki88:
## Update

Update on what I found out so far

- There are 300+ warnings from eslint that the jsdoc types are invalid
- `sizeMappingV2` ad depends on `prebid.js` , which seems the wrong dependency change. The Method `adUnitSetupChecks` maybe should be refactored into a separate file imported from there
-  `adpod` ad depends on `prebid.js` , which seems the wrong dependency change. The Method `checkAdUnitSetup` maybe should be refactored into a separate file imported from there. Also `checkAdUnitSetup` and `adUnitSetupChecks` - seems duplicated :sweat_smile: 
- `dataController` module also imports from `prebid.js` - the `startAuction` method. Not sure what the relation is here

I have still a lot to uncover how the final packages are built, what is generated and what not and which module resolution should be in the final output files.

tsconfig.json
muuki88: I managed to reduce it to the `ts` files and the output looks nearly identical to the javascript before.

Due to the complexity of the build I'm wrapping my head around, this can definetly be a crutch until all js files are ported to ts. Then we may refactor the build a little built to perform linting only on ts files

muuki88:
Boy, this escalated quickly. I upgraded eslint in order to use typescript-eslint, this was follwed by

- changing gulp-eslint to gulp-eslint-new
- upgrading node 16 to node 18
- 1000+ linting errors ( a lot of them very reasonable )

patmmccann:
We're upgrading node for prebid 9 shortly

muuki88:
That's great. Hopefully I have some time this month to push this further 🙈

patmmccann:
Prebid 9 is released and Node is upgraded, I recommend you rebase

muuki88:
Oh yeah! I had a pretty long vacation and will hopefully get to this soon

patmmccann:
This might be a bit simplified if you try and merge in master? Curious if it is possible

muuki88:
Hi @patmmccann 

I had a little time and was able to merge it. I fixed a few invalid type JS type definitions and noticed that the `validator` in the `syncOrtb2` object is never used.

muuki88:
Yeiks. Some eslint stuff is not happy

```
Run npx eslint --no-inline-config --format json $(cat __changed_files.txt | xargs stat --printf '%n\n' 2> /dev/null) > __pr.json || true
  npx eslint --no-inline-config --format json $(cat __changed_files.txt | xargs stat --printf '%n\n' [2](https://github.com/prebid/Prebid.js/actions/runs/12361215535/job/34497890833?pr=11166#step:10:2)> /dev/null) > __pr.json || true
  shell: /usr/bin/bash -e {0}

Oops! Something went wrong! :(

ESLint: 8.57.0

Error [ERR_PACKAGE_PATH_NOT_EXPORTED]: Failed to load plugin 'prebid' declared in '.eslintrc.js': Package subpath './lib/rules/no-restricted-globals.js' is not defined by "exports" in /home/runner/work/Prebid.js/Prebid.js/node_modules/eslint/package.json
```

muuki88:
Damn. I tried a quick read in the eslint docs ( https://eslint.org/docs/v8.x/extend/plugins ), but I have only a slight clue how this plugin setup might work. Are there any plans to upgrade eslint anytime soon? Eslint version `9` is the latest and prebid.js is on `7.x`

patmmccann:
@muuki88 do you think we could solve your problem without upgrading the linter? 

https://github.com/prebid/Prebid.js/blob/master/plugins/eslint/package.json exists, but does have "exports" defined. 

Perhaps we could just define something to resolve error `Package subpath './lib/rules/no-restricted-globals.js' is not defined by "exports"`

muuki88:
Fair point. I should have read the error message with a little more focus. I'll try to fiddle around with the eslint exports in the package.json and see if I can make them all happy

PR #11178, opened 2024-03-06
JW Player RTD Module : populate content url, title and description

karimMourra:
## Type of change

- [x] Feature

## Description of change
Includes content url, title and description when enriching the `ortb.site.content`.

## Other information
Docs PR: https://github.com/prebid/prebid.github.io/pull/5190

### DIFF
--- a/modules/jwplayerRtdProvider.js
+++ b/modules/jwplayerRtdProvider.js
@@ -38,7 +38,7 @@ export const jwplayerSubmodule = {
   /**
    * add targeting data to bids and signal completion to realTimeData module
    * @function
-   * @param {Obj} bidReqConfig
+   * @param {object} bidReqConfig
    * @param {function} onDone
    */
   getBidRequestData: enrichBidRequest,
@@ -323,10 +323,6 @@ export function getContentData(mediaId, segments) {
 }
 
 export function addOrtbSiteContent(ortb2, contentId, contentData, contentTitle, contentDescription, contentUrl) {
-  if (!contentId && !contentData) {
-    return;
-  }
-
   if (ortb2 == null) {
     ortb2 = {};
   }
--- a/modules/jwplayerRtdProvider.js
+++ b/modules/jwplayerRtdProvider.js
@@ -58,13 +58,7 @@ config.getConfig('realTimeData', ({realTimeData}) => {
     return;
   }
   fetchTargetingInformation(params);
-  if (params.overrideContentId !== undefined) {
-    // For backwards compatibility, default to true unless overriden by Publisher.
-    overrideContentId = params.overrideContentId;
-  }
-  overrideContentUrl = !!params.overrideContentUrl;
-  overrideContentTitle = !!params.overrideContentTitle;
-  overrideContentDescription = !!params.overrideContentDescription;
+  setOverrides(params);
 });
 
 submodule('realTimeData', jwplayerSubmodule);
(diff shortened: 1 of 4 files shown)


modules/jwplayerRtdProvider.js
```diff
@@ -217,18 +217,40 @@ function loadVatForPendingRequest(playerDivId, mediaID, callback) {
...
+  if (!sources || !sources.length) {
+    return;
+  }
+
+  const validSource = sources.find(source => !!source.file);
+  if (!validSource) {
+    return;
+  }
+
+  return validSource.file;
```
dgirardi, line 251:
```suggestion
return playlistItem.sources?.find?.(source => source.file)?.file;
```

modules/jwplayerRtdProvider.js
```diff
@@ -217,18 +217,40 @@ function loadVatForPendingRequest(playerDivId, mediaID, callback) {
...
 
-  if (!segments) {
+  if (!item) {
     return null;
   }
 
+  let mediaUrl = item.file;
+  if (!mediaUrl) {
+    mediaUrl = getFileFromSources(item);
+  }
```
dgirardi, line 229:
```suggestion
const mediaUrl = item.file ?? getFileFromSources(item);
```

modules/jwplayerRtdProvider.js
```diff
@@ -329,15 +353,30 @@ export function addOrtbSiteContent(ortb2, contentId, contentData) {
     content.id = contentId;
   }
 
-  const currentData = content.data = content.data || [];
+  if (contentUrl) {
+    content.url = contentUrl;
```
dgirardi, line 357: Should this override what might be in `content` already?

karimMourra: Yes! If a publisher has requested that we enrich the bid request for a given media, then they have determined that the bid request is specific to that media, therefore we should override existing data in that field. 

Unfortunately the oRTB spec only has room for 1 content description, which is unrealistic given that multiple videos could be rendering simultaneously on the page.

dgirardi (approved):

karimMourra:
@patmmccann what does the "on hold" label mean ? Does it block this from getting merged; If so, how can we resolve?

patmmccann:
Yes this blocked from merging, it is a breaking change

patmmccann (changes requested):
This is a breaking change, please require new configuration to enable overwriting existing fields

modules/jwplayerRtdProvider.js
karimMourra: @patmmccann and @dgirardi this has been address by adding configuration params

modules/jwplayerRtdProvider.js
```diff
@@ -71,15 +81,31 @@ export function fetchTargetingInformation(jwTargeting) {
   });
 }
 
+export function setOverrides(params) {
+  // For backwards compatibility, default to always unless overridden by Publisher.
+  overrideContentId = sanitizeOverrideParam(params.overrideContentId, ENRICH_ALWAYS);
```
dgirardi, line 86: Assuming this should be updated for 9, do you have a way to track it - would a TODO comment be helpful?

dgirardi (approved):

patmmccann (approved):

modules/jwplayerRtdProvider.js
karimMourra: added!

karimMourra:
@patmmccann can we remove the `on hold` and merge ?

PR #11185, opened 2024-03-07
RTBHouse Bid Adapter: prevent buyer list overwrite

patmmccann:
Fixes a bug where full buyer list is overwritten with buyer list with pbs (in this case, can only be one)


patmmccann:
@piotrj-rtbh fyi

piwanczak:
Conditional lgtm - rtb house acts as a buyer and thus we don't support reconciliation for potential other buyers overriden on component auction level

ncolletti (approved):

PR #11191, opened 2024-03-08
Sharethrough Bid Adapter : Support for dsa transparency

pierreted90:
## Type of change
- [ x] Feature

## Description of change
Build support for DSA Transparency in Prebid to satisfy a market regulation in the EU.


modules/sharethroughBidAdapter.js
```diff
@@ -92,6 +92,26 @@ export const sharethroughAdapterSpec = {
       req.regs.ext.gpp_sid = bidderRequest.ortb2.regs.gpp_sid;
     }
 
+    if (bidderRequest?.ortb2?.regs?.ext?.dsa) {
```
patmmccann, line 95: why are you shipping so much validation code on the object structure? why not just take whatever the publisher sets and validate it server side?

pierreted90: @patmmccann  Our ad server team has requested we perform validation client-side.  It's an engineering/design decision.

patmmccann: Please move the validation into a shared library; we're not allowing it here

patmmccann (changes requested):
Extensive object validation didn't belong in bid adapters, please move to a library to avoid code duplication.

It also seems really expensive to validate something set once in publisher config billions of times

modules/sharethroughBidAdapter.js
patmmccann: https://docs.prebid.org/dev-docs/modules/validationFpdModule.html actually we chatted and this seems like a better place

jefftmahoney: @patmmccann - is a pub installing / using `validationFpdModule` with their installation enough to get validation for DSA, or are you saying that would be a good module to _add_ DSA validation to?

patmmccann: We're saying runtime validation in your adapter isn't appropriate but your welcome to assist publishers in their validation efforts by commiting to this module

patmmccann (approved):

PR #11192, opened 2024-03-10
Taboola Bid Adapter : add for support fledge

aleskanderl:
## Type of change

- [ ] Bugfix
- [ x] Feature
- [ ] New bidder adapter   
- [ ] Build related changes
- [ ] CI related changes

- [ ] Other

## Description of change
Support fledge in taboola bid adapter

### DIFF
--- a/modules/taboolaBidAdapter.js
+++ b/modules/taboolaBidAdapter.js
@@ -3,7 +3,7 @@
 import {registerBidder} from '../src/adapters/bidderFactory.js';
 import {BANNER} from '../src/mediaTypes.js';
 import {config} from '../src/config.js';
-import {deepAccess, deepSetValue, getWindowSelf, replaceAuctionPrice, isArray} from '../src/utils.js';
+import {deepAccess, deepSetValue, getWindowSelf, replaceAuctionPrice, isArray, safeJSONParse} from '../src/utils.js';
 import {getStorageManager} from '../src/storageManager.js';
 import {ajax} from '../src/ajax.js';
 import {ortbConverter} from '../libraries/ortbConverter/converter.js';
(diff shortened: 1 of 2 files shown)


modules/taboolaBidAdapter.js
```diff
@@ -151,12 +151,61 @@ export const spec = {
...
+        return [];
+      }
+    } else {
+      bids.push(...converter.fromORTB({response: serverResponse.body, request: request.data}).bids);
+    }
+    if (isArray(serverResponse.body.ext?.igbid)) {
+      serverResponse.body.ext.igbid.forEach((igbid) => {
+        let buyerdata;
+        try {
+          buyerdata = JSON.parse(igbid.igbuyer[0]?.buyerdata)
```
dgirardi, line 167: you can use [`safeJSONParse`](https://github.com/prebid/Prebid.js/blob/a0683559f87f96584c2f9aaeb556caa9ac269066/src/utils.js#L1010-L1014) to reduce the boilerplate. 

I am not sure about the purpose of doing it here as well as in the loop below: if the first buyer has invalid `buyerdata` the whole `igbid` is discarded. But if the second one does, only the buyer is discarded. Is that intended?

aleskanderl: Hi @dgirardi,
thanks for the suggestion. Changed the parse to safeParse.
Regarding the second part of comment - you got correct the intention. I want to discard only the buyer in case of failure inside of the loop and not the full action config.

aleskanderl: Hi @dgirardi,
Is there any other issues or PR can be merged?

dgirardi (approved):

Issue #10710, opened 2023-11-09
Id Modules: not respecting publisher configuration


patmmccann:
## Type of issue
Bug

## Description
![image](https://github.com/prebid/Prebid.js/assets/1683175/e354cebf-b874-4175-9727-41eefda12e35)
shows publishers can choose where to set storage, but several modules ignore

https://github.com/criteo-forks/Prebid.js/blob/a22fee00f1ab9084b297ec6d25297edef02382d1/modules/criteoIdSystem.js#L39
https://github.com/prebid/Prebid.js/blob/015f12b3a2ab62d993e2578d6725b6acf8456910/modules/connectIdSystem.js#L45

Lotame appears to not have this setting available and just sets both
https://github.com/prebid/Prebid.js/blob/015f12b3a2ab62d993e2578d6725b6acf8456910/modules/lotamePanoramaIdSystem.js#L45

LiveRamp appears to not use the parameter for html5 or cookie but instead has an additional proprietary parameter that seems related

https://github.com/prebid/Prebid.js/blob/015f12b3a2ab62d993e2578d6725b6acf8456910/modules/identityLinkIdSystem.js#L122

![image](https://github.com/prebid/Prebid.js/assets/1683175/cd9b17fb-e7c1-46be-810c-79b47cbacf8c)

dgirardi:
These are ID modules that use storage directly (grepping for `getStorageManager`). It doesn't necessarily mean that they don't respect publisher config, but in theory ID modules do not need to deal with storage - it's done for them by the base ID module.

```
        admixerIdSystem.js  
        adqueryIdSystem.js  
        adriverIdSystem.js  
        amxIdSystem.js  
        connectIdSystem.js  
        criteoIdSystem.js  
        czechAdIdSystem.js  
        dacIdSystem.js  
        deepintentDpesIdSystem.js  
        euidIdSystem.js  
        freepassIdSystem.js  
        ftrackIdSystem.js  
        gravitoIdSystem.js  
        growthCodeIdSystem.js  
        hadronIdSystem.js  
        id5IdSystem.js  
        identityLinkIdSystem.js  
        idxIdSystem.js  
        imuIdSystem.js  
        intentIqIdSystem.js  
        liveIntentIdSystem.js  
        lotamePanoramaIdSystem.js  
        merkleIdSystem.js  
        mwOpenLinkIdSystem.js  
        naveggIdSystem.js  
        novatiqIdSystem.js  
        pairIdSystem.js  
[...]

PR #11196, opened 2024-03-12
Yandex: add id system

chernodub:
## Type of change

- [ ] Bugfix
- [ ] Feature
- [ ] New bidder adapter   
- [ ] Build related changes
- [ ] CI related changes

- [x] New ID system

PR to the docs repo: https://github.com/prebid/prebid.github.io/pull/5268

## Description of change

Added yandex id system that gets the id provided by [Yandex Metrica tag](https://metrica.yandex.com/list) that is optionally installed by the publishers. In case the id is not set by the tag, it will be set from the yandexIdSystem sumbodule.

### DIFF
--- a/modules/yandexIdSystem.js
+++ b/modules/yandexIdSystem.js
@@ -10,6 +10,7 @@ import { getStorageManager } from '../src/storageManager.js';
 import { logInfo } from '../src/utils.js';
 
 const BIDDER_CODE = 'yandex';
+const BIDDER_EID_KEY = 'yandex.com';
 const YANDEX_ID_KEY = 'yandexId';
 
 const USER_ID_KEY = '_ym_uid';
@@ -48,7 +49,7 @@ export const yandexIdSubmodule = {
   },
   eids: {
     [YANDEX_ID_KEY]: {
-      source: BIDDER_CODE,
+      source: BIDDER_EID_KEY,
       atype: 1,
     },
   },
--- a/modules/yandexIdSystem.js
+++ b/modules/yandexIdSystem.js
@@ -4,28 +4,22 @@
  * @requires module:modules/userId
  */
 
-import { MODULE_TYPE_UID } from "../src/activities/modules.js";
-import { submodule } from "../src/hook.js";
-import { getStorageManager } from "../src/storageManager.js";
-import { logInfo } from "../src/utils.js";
+import { MODULE_TYPE_UID } from '../src/activities/modules.js';
+import { submodule } from '../src/hook.js';
+import { getStorageManager } from '../src/storageManager.js';
+import { logInfo } from '../src/utils.js';
 
-/**
- * @typedef {import('../modules/userId/index.js').Submodule} Submodule
- * @typedef {import('../modules/userId/index.js').SubmoduleConfig} SubmoduleConfig
- * @typedef {import('../modules/userId/index.js').ConsentData} ConsentData
- */
+const BIDDER_CODE = 'yandex';
+const YANDEX_ID_KEY = 'yandexId';
 
-const BIDDER_CODE = "yandex";
+const USER_ID_KEY = '_ym_uid';
+const USER_ID_COOKIE_EXP_MS = 31536000000; // 365 days
 
-const USER_ID_KEY = "_ym_uid";
-const USER_ID_COOKIE_EXP = 31536000000; // 365 days
-
-export const cookieStorage = getStorageManager({
+export const storage = getStorageManager({
   moduleType: MODULE_TYPE_UID,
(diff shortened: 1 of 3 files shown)


chernodub:
Hi @ChrisHuie! Our team would really appreciate it if you (or anybody from the Prebid Team) could look into this PR. If anything needs to be updated/explained — I'm here. Thanks!

ChrisHuie:
@chernodub looks like there is a CodeQL issue around insecure randomness and linting issues with single quotes. Can you please update?

chernodub:
Hi @ChrisHuie! Thanks for the opportunity to improve the code. I've added the changes you recommended 🙏

1. Added the tests.
2. Submitted the docs PR: https://github.com/prebid/prebid.github.io/pull/5268
3. Fixed the insecure randomness issue by using the Web Crypto API. I had to fall back to `Math.random` in environments that do not support the crypto module. As far as I see, several other modules also use the `Math.random` in a similar context, so probably this won't cause any issues. But I'd be happy to add the updates if you have any thoughts on this.
4. Fixed the codestyle.

Please let me know if anything else needs to be added/improved. Thanks!

bretg (commented):
the docs PR https://github.com/prebid/prebid.github.io/pull/5268 suggests there's a bidders parameter ... not seeing this in the code?

modules/yandexIdSystem.js
```diff
@@ -0,0 +1,137 @@
...
+      return;
+    }
+
+    return {
+      id: yandexUidStorage.getUid(),
+    };
+  },
+  eids: {
+    [YANDEX_ID_KEY]: {
+      source: BIDDER_CODE,
```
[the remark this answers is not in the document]
chernodub: Hi @bretg, thanks! 

1. I changed the eid key to domain name. Is it a convention? I wonder what's the reason. We at Yandex have different region-specific domains (e.g. `.com`, `.kz`, etc) so I'm not sure if it's a good idea hardcoding the TLD here.
2. Considering `bidders`. As far as I understood, this is not a userland parameter, but an option provided by Prebid SDK to control which bidders will receive the eid — [here is the link to the docs](https://docs.prebid.org/dev-docs/modules/userId.html#how-it-works:~:text=Array%20of%20Strings-,An%20array%20of%20bidder%20codes%20to%20which%20this%20user%20ID%20may%20be%20sent,-.). Are you sure we need to handle that manually in the module code? I tested this end-to-end I see it working fine without any adjustments on module's side. I also see the related references of `config.bidders` in the core code.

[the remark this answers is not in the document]
bretg: the eids source key is what DSPs look for when they code for a given ID. It's possible that you only care about your own bidder so the value of the eids source isn't as important, but yes, the convention is that it's a domain.

we're good on the bidders part. thanks.

chernodub:
Hi @ChrisHuie! All the updates are done. I'd really appreciate it if you could look into this once more. Thank you!

chernodub:
Hi! Could anybody look into this PR, please? This seems to be lost/forgotten. Thanks!

cc @ChrisHuie @bretg

modules/yandexIdSystem.js
```diff
@@ -0,0 +1,138 @@
...
+    const id = this._cookieStorage.getCookie(USER_ID_KEY);
+
+    return id;
+  }
+
+  _setUid(userId) {
+    if (this._cookieStorage.cookiesAreEnabled()) {
+      const expires = new Date(Date.now() + USER_ID_COOKIE_EXP_MS).toString();
+
+      this._cookieStorage.setCookie(USER_ID_KEY, userId, expires);
```
patmmccann, line 80: The parent module can handle setting the id accorring to publisher config, the submodules only needs to gather it

patmmccann: See #10710

chernodub: Hi @patmmccann, thanks for the recommendation! **Is it required by Prebid for us to maintain different types of storage and allow custom keys?** Can we just fall back to "do nothing"+"log error" when `storage` option is customized for our submodule?

I'm asking this because this submodule is intended to be a thin bridge between our Yandex Metrika publisher tag and our bidder adapter. And we want it to be kept as simple as possible.

We intentionally want only to use the cookie and constant `USER_ID_KEY`.

chernodub:
Hi @patmmccann! Huge thanks for pointing out the way to use the parent module for managing the cookie. I've updated the code, now it looks way simpler! I've also updated the tests and [the docs](https://github.com/prebid/prebid.github.io/pull/5268/commits/a680fc0db535f396a8429dd7feef7e6855efab43)

patmmccann (approved):

PR #11200, opened 2024-03-13
Anyclip Bid Adapter : initial release

basil79:
## Type of change

- [ ] Bugfix
- [ ] Feature
- [x] New bidder adapter   
- [ ] Build related changes
- [ ] CI related changes

- [ ] Other

## Description of change

Added new bid adapter AnyClip. It uses the external PubTag library for fast bids, first please add the following file to the `<HEAD>`

```html
<!-- AnyClip PubTag -->
<script src="https://anyclip-player.s3.amazonaws.com/anyclip-widget/pubtag/pubtag.js"></script>
```

- contact email of the adapter’s maintainer: basil.goldman@gmail.com or basil@anyclip.com
```js
var adUnits = [{
    code: 'adunit1',
    mediaTypes: {
        banner: { // support banner only
            sizes: [
                [300, 250]
            ]
        }
    },
    bids: [{
        bidder: 'anyclip',
        params: {
            publisherId: '12345', // required, string
            supplyTagId: '-mptNo0BycUG4oCDgGrU', // required, string
            floor: 0.1 // optional, floor
        }
    }]
}]
```

### DIFF
--- a/modules/anyclipBidAdapter.js
+++ b/modules/anyclipBidAdapter.js
@@ -1,6 +1,6 @@
 import {registerBidder} from '../src/adapters/bidderFactory.js';
 import {BANNER} from '../src/mediaTypes.js';
-import {deepAccess, isArray, logError, logInfo} from '../src/utils.js';
+import {deepAccess, isArray, isFn, logError, logInfo} from '../src/utils.js';
 import {config} from '../src/config.js';
 
 /**
@@ -14,6 +14,28 @@ const BIDDER_CODE = 'anyclip';
 const ENDPOINT_URL = 'https://prebid.anyclip.com';
 const DEFAULT_CURRENCY = 'USD';
 const NET_REVENUE = false;
+
+/*
+ * Get the bid floor value from the bidRequest object, either using the getFloor function or by accessing the 'params.floor' property.
+ * If the bid floor cannot be determined, return 0 as a fallback value.
+ */
+function getBidFloor(bidRequest) {
+  if (!isFn(bidRequest.getFloor)) {
+    return deepAccess(bidRequest, 'params.floor', 0);
+  }
+
+  try {
+    const bidFloor = bidRequest.getFloor({
+      currency: DEFAULT_CURRENCY,
+      mediaType: '*',
+      size: '*',
+    });
+    return bidFloor.floor;
+  } catch (err) {
+    logError(err);
+    return 0;
+  }
+}
 
 /** @type {BidderSpec} */
 export const spec = {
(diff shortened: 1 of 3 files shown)


Rothalack:
For your external PubTag code. I think the only other bidder doing something like this is Criteo and they seem to have gotten worked out where you do not need to ask the user to include an extra script in the head. I think you should attempt to do it that way, having the user have to add an extra script isn't an intended way of doing it. You should be able to see it here
https://github.com/criteo-forks/Prebid.js/blob/master/modules/criteoBidAdapter.js

Feel free to push back, I'm not absolutely certain about this.

basil79:
> For your external PubTag code. I think the only other bidder doing something like this is Criteo and they seem to have gotten worked out where you do not need to ask the user to include an extra script in the head. I think you should attempt to do it that way, having the user have to add an extra script isn't an intended way of doing it. You should be able to see it here https://github.com/criteo-forks/Prebid.js/blob/master/modules/criteoBidAdapter.js
> 
> Feel free to push back, I'm not absolutely certain about this.

@Rothalack, yes, you right, it is similar to the Criteo adapter, but Criteo uses ```loadExternalScript``` from ```'../src/adloader.js'```, which requires a code review for the external file to allow our adapter to use this feature.

https://github.com/criteo-forks/Prebid.js/blob/master/modules/criteoBidAdapter.js#L186

This is normal for us: all of our publishers must include our PubTag library in order to use our prebid adapter.

basil79:
> Hi @basil79 ; criteo is in process of removing this dependency in prebid 9

Hi @patmmccann , give me their pr I will check

basil79:
@Rothalack, any updates?

Rothalack:
There is a divide in opinion on whether or not the way you're doing this should be allowed. I've voiced my thoughts on this. I do not feel like I'm in the position to make this decision, it's a policy decision.

I'm unsure if @patmmccann s comment here was a request for you to make a change or just information for you.

basil79:
> There is a divide in opinion on whether or not the way you're doing this should be allowed. I've voiced my thoughts on this. I do not feel like I'm in the position to make this decision, it's a policy decision.
> 
> I'm unsure if @patmmccann s comment here was a request for you to make a change or just information for you.

@Rothalack , criteo has been using this method for many years, we want to use it too. I haven't found any policies or restrictions to not do it this way.

Rothalack:
They're doing it the intended way, which is to have the external JS in an open source github repo under Prebid. My understanding at least is that this is to keep the spirit of open source and have the code be publicly accountable.

basil79:
> They're doing it the intended way, which is to have the external JS in an open source github repo under Prebid. My understanding at least is that this is to keep the spirit of open source and have the code be publicly accountable.

but we don't load external JS file using Prebid, our publishers do that

Rothalack:
Right, which is where the divide in opinion occurs.

basil79:
bottom line, any publisher wishing to use any prebid adapter must contact the company (ex SSP provider) that owns the adapter to obtain connection params etc.

Rothalack (approved):
I'm going to go ahead and approve and merge this. I don't see any more discussion on this topic and it seems like Pat was in approval.

Issue #11214, opened 2024-03-14
Enable `postInstallAllowed` for submodules


olafbuitelaar:
## Type of issue
question/bug

## Description
It seems when loading modules at a later stage, the postInstallAllowed module prevents modules to register to the submodule registry.
i'm not sure what the rationale is behind preventing the loading of modules? 
the relevant code; 
https://github.com/prebid/Prebid.js/blob/fa8c030e24c0a5e193f060106de3ab6d15796c12/src/hook.js#L36-L41
i've tested this with allowing post installation and it appears to work fine.
i'm happy to create a PR if needed.

dgirardi:
The rationale is that some module types assume(d) that installation would happen at the "normal" time and may need some changes to support submodules to be added later. It looks like we never went through the changes though and now `postInstallAllowed` is always false.

User ID modules should work, because the work to support it was done before it was needed (which apparently [surprised me](https://github.com/prebid/Prebid.js/blob/fa8c030e24c0a5e193f060106de3ab6d15796c12/modules/userId/index.js#L1038-L1040), although I later forgot about it). I have no issue turning this on for user ID.

olafbuitelaar:
thank you, what would be the best way to enable this for the userId module (and maybe submodules)?

olafbuitelaar:
is there somewhere a list of modules, which don't support post-installation?

dgirardi:
Most modules were written and tested assuming that they'd be installed "normally", so none necessarily supports it. Those that are split into multiple submodules (using `module`/`submodule`) are by their nature more likely to rely on that assumption and AFAIK the only ones that are explicitly disabled through that `postInstall` flag:

- [RTD modules](https://github.com/prebid/Prebid.js/blob/d17c44bc98b88423882b4befa8baf414780cecd3/modules/rtdModule/index.js#L427)
- [video modules](https://github.com/prebid/Prebid.js/blob/d17c44bc98b88423882b4befa8baf414780cecd3/modules/videoModule/coreVideo.js#L243)
- [FPD modules](https://github.com/prebid/Prebid.js/blob/d17c44bc98b88423882b4befa8baf414780cecd3/modules/fpdModule/index.js#L50)
- [PAAPI modules](https://github.com/prebid/Prebid.js/blob/d17c44bc98b88423882b4befa8baf414780cecd3/modules/paapi.js#L25)

PR #11215, opened 2024-03-14
Userid module: allow postInstall 

olafbuitelaar:


patmmccann:
fixes https://github.com/prebid/Prebid.js/issues/11214

patmmccann (approved):

PR #11222, opened 2024-03-15
Autoplay detection library: initial release && Teads Bid Adapter: detect autoplay

github-matthieu-wipliez:
## Type of change

- [ ] Bugfix
- [x] Feature
- [ ] New bidder adapter   
- [ ] Build related changes
- [ ] CI related changes

- [ ] Other

## Description of change

This PR adds an autoplay detection library, and updates the Teads adapter to use this feature to filter out bid responses containing ads that require autoplay when autoplay is not enabled on the browser.

The autoplay detection library works by attempting to play a 1-frame, 1 second video of a 160x90 pixels black image without sound, encoded with ffmpeg as H.264 wrapped in an MP4. I reduced its size by stripping the unneeded encoder details. The `play()` promise fails immediately when the OS does not allow autoplay, but when autoplay is allowed the promise does not necessarily resolve right away, hence the choice to consider that autoplay is enabled by default unless it has been explicitly forbidden. Note that the video does not need to be in the DOM for this to work, which makes this easier. We have tested this in production on our side.

For context see preliminary discussion on Slack: https://prebid.slack.com/archives/C7TLBVAH3/p1709305698637079


github-tom-kuhnen (approved):
LGTM 👌

libraries/autoplayDetection/autoplay.js
```diff
@@ -0,0 +1,48 @@
...
+// ask ChatGPT for a 160x90 black PNG image (1/8th the size of 720p)
+//
+// encode with:
+// ffmpeg -i black_image_160x90.png -r 1 -c:v libx264 -bsf:v 'filter_units=remove_types=6' -pix_fmt yuv420p autoplay.mp4
+// this creates a 1 second long, 1 fps YUV 4:2:0 video encoded with H.264 without encoder details.
+//
+// followed by:
+// echo "data:video/mp4;base64,$(base64 -i autoplay.mp4)"
+
+const autoplayVideoUrl =
```
patmmccann, line 20: much happier this worked than sourcing an external file

patmmccann (approved):

modules/teadsBidAdapter.js
```diff
@@ -1,12 +1,16 @@
...
 import {getStorageManager} from '../src/storageManager.js';
+import {startDetection, isAutoplayEnabled} from '../libraries/autoplayDetection/autoplay.js';
 
 /**
  * @typedef {import('../src/adapters/bidderFactory.js').BidRequest} BidRequest
  * @typedef {import('../src/adapters/bidderFactory.js').Bid} Bid
  */
 
+// start autoplay detection as soon as Prebid is loading
+startDetection();
```
karimMourra, line 12: why not call this in `autoplay.js` ? Every module using this library will need to do it, so might as well do it in the library. I can't think of a case where a module would want to delay this

karimMourra (commented):
I think `startDetection` can be called in the library . 
This is great stuff!

modules/teadsBidAdapter.js
github-matthieu-wipliez: Yes good point, this makes sense since libraries are not included by default, so the detection will only run if a module includes it. I'll make the change!

github-matthieu-wipliez: By doing that, we can also remove the `detectionInProgress` logic, since the code is executed once when the module is initialized.

github-matthieu-wipliez: Done in ed3427c

karimMourra (approved):

aanton:
Hi!

We have detected a problem in iOS APP using webviews that includes a Prebid file with the Teads adapter (that includes the autoplay library which `startDetection` function runs as soon as possible). A fullscreen black 1s videoplayer is displayed (hiding the content) and the user must click on it and close it to display the page content.

It seems the webview opens this 1s videoplayer on fullscreen automatically.

Any ideas or suggestions about this issue?

Thanks!
Armando

github-matthieu-wipliez:
> Hi @github-matthieu-wipliez @karimMourra @patmmccann.
> 
> We have detected a problem in an iOS APP using webviews that includes a Prebid file with the Teads adapter (that includes the autoplay library which `startDetection` function runs as soon as possible). A fullscreen black 1s videoplayer is displayed (hiding the content) and the user must click on it and close it to display the page content.
> 
> It seems the webview opens this 1s videoplayer on fullscreen automatically.
> 
> Any ideas or suggestions about this issue?
> 
> Thanks! Armando

Hi @aanton

yes you should allow inline media playback, see this issue for details: #11438

Thanks,
Matthieu

aanton:
Thanks @github-matthieu-wipliez!

PR #11225, opened 2024-03-18
Smartadserver Bid Adapter: Update default value of startDelay video parameter

krzysztofequativ:
## Type of change
[x] Feature

## Description of change
The **startDelay** parameter is set to **1** by default.


modules/smartadserverBidAdapter.js
```diff
@@ -132,18 +132,19 @@ export const spec = {
...
+        case 0:
+          return 1;
+        case -1:
+          return 2;
+        case -2:
+          return 3;
       }
     }
-    return 2;// Default value for all exotic cases set to bid.params.video.startDelay midroll hence 2.
+    return 1; // SADR-5619
```
patmmccann, line 147: this appears wrong; what do you do if videoMediaType is say, 180? In this case, you appear to ignore the declaration and guess that it is preroll?

patmmccann (changes requested):

modules/smartadserverBidAdapter.js
krzysztofequativ: Good point, @patmmccann, I've updated the implementation. Thx.

patmmccann: If the publisher specifies midroll starting at 3 minutes (180), you are calling it preroll still?

krzysztofequativ: You are completely right. It's fixed.

krzysztofequativ: Hi @patmmccann, could you review it, please? Thx.

patmmccann (approved):

PR #11227, opened 2024-03-18
Adquery Bid Adapter: bidWon bugfix

awiackiewicz:
## Type of change

- [x] Bugfix

## Description of change

removed ad data from bidWon event handler


modules/adqueryBidAdapter.js
```diff
@@ -137,7 +137,7 @@ export const spec = {
     const bidString = JSON.stringify(bid);
     let copyOfBid = JSON.parse(bidString);
     delete copyOfBid.ad;
-    const shortBidString = JSON.stringify(bid);
+    const shortBidString = JSON.stringify(copyOfBid);
```
patmmccann, line 140: you shouldn't need to double stringify
```
copyOfBid = { ...bid }
delete copyOfBid.ad
 const shortBidString = JSON.stringify(copyOfBid);
``` 
should save some cycles

patmmccann (dismissed):

ChrisHuie (approved):

Issue #11189, opened 2024-03-08
PAAPI module: fill out auctionSignals.prebid.ortb2Imp.banner better


patmmccann:
## Type of issue
Feature request

### Expected results
Sizes are available in auctionSignals.prebid.ortb2Imp.banner, and some more fun additions, including ad unit properties from mediaTypes.video on imp.video

Let's fill out whatever we can from the ad unit, including format, w, h, pos, 

Let's not restrict these to PAAPI if we can, but we should be able to provide these into the ortb2Imp object in the bidder request, then we get them here for free.

### Actual results
Only ext are available 

Here is a full auctionSignals.prebid object we're sending

`{
  "bidfloor": 1.2068775,
  "bidfloorcur": "USD",
  "ortb2": {
    "source": {
      "tid": "fe994492-7c57-4e25-baec-0cb36b71bf6f",
      "fd": 0
    },
    "cur": [
      "USD"
    ],
    "at": 1,
    "device": {
      "w": 1512,
      "h": 150,
      "dnt": 1,
      "ua": "Mozilla/5.0 (Macintosh; Intel Mac OS X 10_14_6) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/118.0.0.0 Safari/537.36 FLEDGE_GAM_EXTERNAL_TESTER",
      "language": "en",
      "deviceType": 2,
      "js": 1,
      "sua": {
        "source": 1,
        "browsers": [],
        "mobile": 0
      }
    },
    "site": {
      "domain": "repurposingjunkie.com",
      "publisher": {
        "domain": "repurposingjunkie.com"
      },
      "page": "https://repurposingjunkie.com/cricut-heat-press-alternative-htvront-heatpress-review/?commit=ed7225c&at_features={%22fldg%22:true}&pbjs_debug=true",
      "content": {
        "language": "en"
      },
      "cat": [
        "IAB9-2"
      ],
      "pagecat": [
        "IAB9-2"
      ],
      "privacypolicy": 1,
      "mobile": 1,
      "ext": {
        "data": {
          "mcmp": [
            "BA_WSKW20"
          ],
          "site_id": [
            "600ad629adec3b6b7b1dc60f"
          ],
          "verticals": [
            "Crafts"
          ],
          "pmp_elig": [
            true
          ],
          "sens": [
            "alc",
            "ast",
            "cbd",
            "cosm",
            "dat",
            "gamc",
            "gamv",
            "rel",
            "ssr",
            "srh",
[...]

Issue #10690, opened 2023-10-31
Prebid invokes runAdAuction()


jdwieland8282:
`navigator.runAdAuction()` is a new browser function in Chrome (only) that begins a PAA (formally known as Fledge) Interest Group Auction. A publisher may invoke this function on their own, or a party with on page javascript may. GPT.js (GAM) would like to invoke this function for publishers using GAM.  Prebid should explore an alternative approach given our on page presence. 

The advantage Prebid has over GPT.js is prebid.js can invoke `RAA()` before GPT.js does. GPT.js will wait for the browser to run its contextual auction, otherwise known as the prebid auction. Prebid is effectively blocking GPT.js from calling `RAA()`.  Prebid.js can do its normal auction mechanics ie, run a contextual auction, skip the contextual auction and just call `RAA()`, or it can run/do both. 

Assumptions:
1. Publisher can pass any direct sold bid to Prebid. If the bid from the normal prebid contextual auction , or the bid from RAA() is lower than the direct sold "floor", the publisher can choose to render the direct sold ad. 

Questions:
1. What happens when prebid.js invokes RAA()? 
2. Does an ad render, if not why? 
3. Do impression beacons fire, if not why?

**Decision Logic URL**
In this architecture, Prebid becomes the top level seller (TLS). The TLS needs to provide `RAA()` with an auctionConfig.decisionLogicUrl. For the sake of simplicity the decisionLogicUrl should select the highest bid. We acknowledge that this isn't how things work today, we'll iterate on this later. 

This flow diagram represents a high level vision of what a Prebid invoked PAA auction could look like. 

<img width="745" alt="Screenshot 2023-10-31 at 12 45 56 PM" src="https://github.com/prebid/prebid.github.io/assets/14223042/14c45d71-f24e-45b9-873c-9ff9453a3048">

patmmccann:
related: https://github.com/WICG/turtledove/issues/851

dmdabbs:
You;re thinking perhaps something like this using Prebid infra as top-level seller....

```
/*
  Excerpted from 
  https://github.com/WICG/turtledove/blob/main/FLEDGE.md#21-initiating-an-on-device-auction

   All fields that specify URLs for loading scripts or JSON (decisionLogicURL and trustedScoringSignalsURL) 
    must be same-origin with seller and must point to URLs whose responses include the HTTP response header 
    Ad-Auction-Allowed: true to ensure they are allowed to be used for loading Protected Audience resources.
  */
[...]

PR #11259, opened 2024-03-26
PAAPI: add top level auction example

dgirardi:
## Type of change

- [x] Feature

## Description of change

- Add `requestedSize` to the configs returned by `pbjs.getPAAPIConfig`. For now, this is the first banner size defined by the ad unit.
- Add an example showing how to use `getPAAPIConfig` to run a top level auction using your own decision logic, using a test IG set up by openx and the debugging module.


integrationExamples/gpt/top-level-paapi/tl_paapi_example.html
```diff
@@ -0,0 +1,186 @@
...
+            });
+
+            pbjs.addAdUnits(adUnits);
+            pbjs.requestBids({
+                bidsBackHandler: sendAdserverRequest,
+                timeout: PREBID_TIMEOUT
+            });
+        });
+
+        function raa() {
```
patmmccann, line 110: does it make sense to have some of this code in a submodule, which might end up being extremely compact on initial commit, but would provide the convenient logical abstraction that a user would pick the PAAPI submodule associated with their choice of TLS?

integrationExamples/gpt/top-level-paapi/tl_paapi_example.html
```diff
@@ -0,0 +1,186 @@
...
+                        });
+                    })
+            ).then(won => won.every(el => el));
+        }
+
+        function sendAdserverRequest() {
+            if (pbjs.adserverRequestSent) return;
+            pbjs.adserverRequestSent = true;
+            raa().then((allPaapi) => {
+                if (!allPaapi) {
```
patmmccann, line 143: this calls gam if paapi doesn't fill, correct? But PAAPI will always fill if we give it the best contextual bid from prebid as its best alternative? Is the below code basically saying 'Let's compete Prebid vs PAAPI, and if neither provide any ad call gam?'

dgirardi: I don't think we can give it the best contextual bid. Some bidders may decide to provide "extra" bids for it, but those don't necessarily contain all, or even the same, contextual bids, and Prebid can't control them.

integrationExamples/gpt/top-level-paapi/decisionLogic.js
```diff
@@ -0,0 +1,87 @@
...
+
+function scoreAd(
+  adMetadata,
+  bid,
+  auctionConfig,
+  trustedScoringSignals,
+  browserSignals,
+  directFromSellerSignals,
+) {
+  var score = bid+20;
```
patmmccann, line 15: reference comparison to floor (or top prebid bid)


  var sellerSignals = auctionConfig.sellerSignals;
  var floor = sellerSignals.floor || 0;
  var modifiedBid = bid * 0.95;

  // temporary reporting: https://github.com/WICG/turtledove/blob/main/Proposed_First_FLEDGE_OT_Details.md#reporting
  const reportingBrowserSignals = {...browserSignals, bid, modifiedBid}
  const macros = '&winningBid=${winningBid}&winningBidCurrency=${winningBidCurrency}&topLevelWinningBid=${topLevelWinningBid}&topLevelWinningBidCurrency=${topLevelWinningBidCurrency}';
  globalThis.forDebuggingOnly?.reportAdAuctionLoss(
    buildReportURL('bid-loss', auctionConfig, reportingBrowserSignals) + macros);
  globalThis.forDebuggingOnly?.reportAdAuctionWin(
    buildReportURL('bid-win',  auctionConfig, reportingBrowserSignals) + macros);

  return {
    ad: null,
    desirability: (modifiedBid > floor && (browserSignals.bidCurrency == '???' || auctionConfig.sellerCurrency == browserSignals.bidCurrency)) ? modifiedBid : 0,
    allowComponentAuction: true,
    bid: modifiedBid
  };
}

integrationExamples/gpt/top-level-paapi/decisionLogic.js
```diff
@@ -0,0 +1,87 @@
...
+  console.groupCollapsed('browserSignals');
+  console.log( 'topWindowHostname: %s', browserSignals.topWindowHostname );
+  console.log( 'seller: %s', browserSignals.seller );
+  console.log( 'componentSeller: %s', browserSignals.componentSeller );
+  console.log( 'topLevelSeller: %s', browserSignals.topLevelSeller );
+  console.groupEnd();
+  console.groupEnd();
+  console.groupEnd();
+
+  log('reportResult', { auctionConfig, browserSignals, directFromSellerSignals });
```
patmmccann, line 77: could we demonstrate calling a url in sellerSingals 


eg



  const reportingURL =
    (sellerSignals.reportingURL || 'https://privacysandbox-reporting.openx.net/fledgeReporting')
    + '?'
    + Object.entries(params)
            .filter(([k,v]) => v)
            .map(([k,v]) => `${k}=${encodeURIComponent(v)}`)
            .join('&');

  return reportingURL;
}

integrationExamples/gpt/top-level-paapi/decisionLogic.js
patmmccann: citing my source ```https://privacysandbox.openx.net/fledge/decision-logic-component.js````

dgirardi:
My plan for this is to update `decisionLogic` so that it demonstrates floor enforcement, and remove the `reportResult` portion.

Pending are:
 1. moving the `runAdAuction` logic into a module. This should also define new events to fire for PAAPI won / loss.
 2. moving the rendering logic into `paapi.js`. This should define a new rendering API. It's  unclear yet how to deal with size (see https://github.com/prebid/Prebid.js/issues/11189)
 3. a reference implementation for the decision logic that includes reporting.

Let me know if you'd prefer some of the above to be included in this.

integrationExamples/gpt/top-level-paapi/tl_paapi_example.html
jdwieland8282: I think so yes. A pub makes 2 decissions. 
1. I want to run PAAPI with prebid
2. I want to do it myself (pub sub module) or I want GPT to do it for me

jdwieland8282:
I'm working on finding a pub willing to run a demo for us. If we don't find one by May 1 what are our options?  My preference would be to use Chrome flags to run the demo from prebid.org.

integrationExamples/gpt/top-level-paapi/tl_paapi_example.html
laurb9: That's `auctionSignals.prebid.bidfloor` no ? I don't know if all buyers take it into account, so it may still not fill even if everyone bid all the time and above the floor.

Anyway, I assume that deleting the `<div>` is what makes GAM skip that slot if it was filled by PAAPI. Is there a better way to disable the slot more explicitly and perhaps less permanent ? It's fine either way as a test page but I imagine it won't work if I wanted to periodically refresh the ads.

integrationExamples/gpt/top-level-paapi/decisionLogic.js
```diff
@@ -0,0 +1,44 @@
...
+    } else if (bid < bidfloor) {
+      console.log(`Bid (${bid}) lower than contextual winner/floor (${bidfloor}). Rejecting bid.`);
+      result.desirability = -1;
+      result.rejectReason = 'bid-below-auction-floor';
+    }
+  }
+  console.groupEnd();
+  console.log(...LOG_PREFIX, 'Result:', result);
+  return result;
+}
```
laurb9, line 44: `reportResult` should still be defined even if it does not do anything, to avoid
> Worklet error: https://127.0.0.1:9999/integrationExamples/gpt/top-level-paapi/decisionLogic.js `reportResult` is not a function.

Could be something like this
```javascript
function reportResult(auctionConfig, browserSignals) {
  log('reportResult', { auctionConfig, browserSignals });
  sendReportTo(`${auctionConfig.seller}/report/win?${Object.entries(browserSignals).map(([k, v]) => `${k}=${encodeURIComponent(v)}`).join('&')}`);
  return {}
}
```

dgirardi: that would replace it with an error about the report call failing :) I am already worried about people copying this demo into production so I'd rather avoid it; the plan is to work with @patmmccann to publish an actual reference implementation separately

integrationExamples/gpt/top-level-paapi/tl_paapi_example.html
dgirardi: Yes, this is not a great way to render. I am not sure yet what the render API will look like:

 - when using GAM, the use case I'm hearing is to run fledge only after GAM picks the Prebid contextual winner (to keep adx demand in GAM). In practice this would mean returning the URN back to the rendering logic running in the GAM creative (PUC or equivalent). The publisher would not call any rendering API but somehow enable this behavior.
 - when not using GAM, the publisher calls `renderAd(document, adId)`. We could do something similar and only "pretend" to render `adId`, rendering the fledge ad instead in a nested frame, if the publisher enabled the same behavior as above. Or we could provide separate `runPaapi`/`renderPaapi` APIs. 
 - I'm not sure there will be a use case for what's demoed here, only run gam if fledge has no winner. Pubs would probably prefer to let GAM run fledge and get better demand.

integrationExamples/gpt/top-level-paapi/tl_paapi_example.html
```diff
@@ -0,0 +1,189 @@
...
+
+<body>
+<h2>Standalone PAAPI Prebid.js Example</h2>
+<p>Start local server with:</p>
+<code>gulp serve-fast --https</code>
+<p>Chrome flags:</p>
+<code>--enable-features=CookieDeprecationFacilitatedTesting:label/treatment_1.2/force_eligible/true
+    --privacy-sandbox-enrollment-overrides=https://localhost:9999</code>
+<p>Join interest group at <a href="https://privacysandbox.openx.net/fledge/advertiser">https://privacysandbox.openx.net/fledge/advertiser</a>
+</p>
```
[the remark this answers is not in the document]
dgirardi: My intention was to make this as real-world as possible but I couldn't get any adapter that would both work and agree to be in here. Mocking openx' response was a compromise, hopefully temporary.

integrationExamples/gpt/top-level-paapi/decisionLogic.js
laurb9: Is this about the event url showing up as an error in the Network tab ? It's still useful to look at parameters, but if unwanted can comment out sendReportTo or make it conditional on something like sellerSignals.reportURL

integrationExamples/gpt/top-level-paapi/tl_paapi_example.html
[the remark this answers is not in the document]
laurb9: I was not sure if the demo needed to be self-contained. If so, the buyer can be local too, at a small cost of being slightly confusing to share a domain with seller. It's fine either way.

Mocking the openx adapter response is also fine - I was not aware this is possible so thanks for showing me how to do it :)

patmmccann (approved):

laurb9 (approved):

patmmccann:
> * moving the `runAdAuction` logic into a module. This should also define new events to fire for PAAPI won / loss.

tracking as #10690

piotrj-rtbh:
@patmmccann, did you have a chance to run the TL auction example on a Dev Chrome?
Mine is Version 126.0.6423.2 (Official Build) dev (64-bit) and once run (with all the appropriate flags) I get an error in the console:
```
utils.js:211 Prebid ERROR: Error executing bidsBackHandler null TypeError: Cannot read properties of undefined (reading 'apply')
    at navigator.runAdAuction (tl_paapi_example.html:10:45)
    at tl_paapi_example.html:116:42
    at Array.map (<anonymous>)
    at raa (tl_paapi_example.html:115:22)
    at sendAdserverRequest (tl_paapi_example.html:144:13)
    at Object.auctionDone (prebid.js:546:9)
    at auction.js:225:23
    at auction.js:441:5
    at executeCallback (auction.js:219:7)
    at auctionDone (auction.js:251:5)
```
I cannot figure out why is that happening, all is defined properly and this is not case for my regular Version 124.0.6367.61 (Official Build) (64-bit) - this one works like a charm.

PR #11269, opened 2024-03-28
fledgeForGpt: automatically import paapi for NPM consumers

dgirardi:
## Type of change

- [x] Bugfix

## Description of change

With https://github.com/prebid/Prebid.js/pull/10930, `fledgeForGpt` was made a submodule of `paapi`. The build automatically includes submodules' parent modules, but this is not the case for NPM consumers.

This also fixes webpack's `splitChunk` configuration to avoid other, unrelated output duplication - logic that's not pulled in from `src/prebid.js`, but imported from modules, is duplicated for each module.

### DIFF
--- a/gulpfile.js
+++ b/gulpfile.js
@@ -295,8 +295,7 @@ function bundle(dev, moduleArr) {
   [coreFile].concat(moduleFiles).map(name => path.basename(name)).forEach((file) => {
     (depGraph[file] || []).forEach((dep) => dependencies.add(helpers.getBuiltPath(dev, dep)));
   });
-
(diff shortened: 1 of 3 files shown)


osazos (approved):
LGTM

Btw, can you explain why this can be removed in Prebid 9?

dgirardi:
>  why this can be removed in Prebid 9?

We can remove it in 9 because it's a breaking change. We should remove it in 9 because it would bring it in line with how other submodules work (NPM consumers need to explicitly import the parent). Alternatively we could also do the same for all other submodules.

patmmccann (approved):

Issue #11131, opened 2024-02-22
Paapi module: support for standard interest groups object(s)


patmmccann:
see https://github.com/InteractiveAdvertisingBureau/openrtb/blob/main/extensions/community_extensions/Protected%20Audience%20Support.md

Not sure if part of ortb convertor or paapi module

patmmccann:
@laurb9 https://github.com/prebid/Prebid.js/blob/b2ace9f59d90c563209d7597b83499c68312d886/modules/openxBidAdapter.js#L104C1-L105C1 is now possible

Issue #11094, opened 2024-02-13
PAAPI Module: accumulate perBuyerSignals


patmmccann:
## Type of issue
Feature request

## Description
Some bidders that are DSPs will be more likely to respond with perBuyerSignals than AuctionConfig. We should have a utility function to assemble the perBuyerSignals for creation of the publisher auction config

PAAPI configuration would take an auctionConfig, PBS would be injected only into that one. Prebid wouldnt be able to prevent someone injecting it into a bidder provided auction Config, but that wouldnt be in scope for this. In the future, some ssps might want to provide an auction config that welcomes injections but it should be only when identified in the response somehow.

A publisher could say for example call

pbjs.accumulatePerBuyerSignalForAdUnitCode('x', myAuctionConfig) where myAuctionConfig is a json object defining a full auction config that could have decisionLogic on the publisher or some other origin.

Aside: the rtbhouse module already provides publisher defined auction config support in their vendor module

dgirardi:
Summary of offline discussions:

adapters should be able to provide just buyers and buyer signals instead of a complete auction configs;
buyers should collated into a component level auction using seller and decision logic provided by the publisher;
buyer signals should be augmented with FPD in the same way that auction signals are now.

mikemangas:
@dgirardi  @patmmccann  is this module now necessary in order to use fledgeForGpt? Or do they both work completely independent from each other?

PR #11277, opened 2024-04-01
PAAPI: add support for protected audience extensions and "direct" buyers (`igb`)

dgirardi:
## Type of change

- [x] Feature

## Description of change

This adds support for the ORTB [protected audience extensions](https://github.com/InteractiveAdvertisingBureau/openrtb/blob/main/extensions/community_extensions/Protected%20Audience%20Support.md):

- when PAAPI is enabled, bid requests' `ortb2Imp.ext.igs` is populated (similar and in addition to `ext.ae`)
- bid adapters can reply with buyers (`igb`) objects in additon to auction configs (`config`). The `fledgeAuctionConfigs` response property is migrated to `paapi`. for example:
    ```
      // current response from interpetResponse, returning an auction config
     return {bids: [...], fledgeAuctionConfigs: [{bidId: 'xxx', config: {...}}]}

     // equivalent to the above, using `paapi`
     return {bids: [...], paapi: [{bidId: 'xxx', config: {...}}]}

     // return a buyer (igb) object instead of a full auction config
     return {bids: [...], paapi: [{bidId: 'xxx', igb: {...}}]}
    ```
- ortbConverter can now parse responses' `ext.igi` to generate a `paapi` array.
- publishers can set up auctions to run igb with:
  ```
   pbjs.setConfig({
      paapi: {
         enabled: true,
         componentSeller: {
            auctionConfig: {
                  decisionLogicURL: 'publisher.example'
                  // ...
            }
         }
      }
   })
  ```
- if `paapi.componentSeller.auctionConfig` is set, we attempt to collect all IGB objects into a single auction config, which is then treated like any other auction config for the purpose of `getPAAPIConfig/setPAAPIConfigForGPT`. 
- it's possible for IGB objects to be incompatible with each other (same origin, different signals). In this case we split the conflicts into a second (or third, etc) auction config.
- Because IGB objects come from different adapters, and first-party data (`ortb2`/`ortb2Imp`) can vary by adapter, auction configs generated from IGB have it set in `perBuyerSignals` instead of `auctionSignals`. "Normal" auction configs also have first party data repeated in `perBuyerSignals`, for consistency.
- Instead of one auction config (minus conflicts), it's possible to generate one for each bidder with
    ```
     pbjs.setConfig({
         paapi: {
             enabled: true,
             componentBuyers: {
                  separateAuctions: true,
                  auctionConfig: {/* ... */}
             }
         }
     })
    ```
  In this case, `auctionSignals` is populated with first party data as usual.

## Other information
Closes https://github.com/prebid/Prebid.js/issues/11131
Closes https://github.com/prebid/Prebid.js/issues/11094


dgirardi:
Summary of PMC feedback:

 - `componentBuyers` config should be renamed to `componentSeller`
 - bidders should get a signal that tells them whether the publisher can set up a component seller (`bidRequest.paapi.componentSeller` ?)

rdgordon-index:
> when PAAPI is enabled, bid requests' ortb2Imp.ext.igi is populated (similar and in addition to ext.ae)

Shouldn't this be `BidRequest.imp.ext.igs` ?

dgirardi:
> Shouldn't this be BidRequest.imp.ext.igs ?

Yes. It looks like I used the right field in code but not the description.

pm-harshad-mane:
@jlquaccia please review the PR

modules/paapi.js
```diff
@@ -197,9 +315,30 @@ export function markForFledge(next, bidderRequests) {
...
+          const extAe = bidReq.ortb2Imp?.ext?.ae;
+          if (igsAe !== extAe && igsAe != null && extAe != null) {
+            logWarn(MODULE, `Bid request defines conflicting ortb2Imp.ext.ae and ortb2Imp.ext.igs, using the latter`, bidReq);
+          }
+          const bidAe = igsAe ?? extAe ?? ae;
+          if (bidAe) {
+            deepSetValue(bidReq, 'ortb2Imp.ext.ae', bidAe);
+            bidReq.ortb2Imp.ext.igs = Object.assign({
+              ae: bidAe,
+              biddable: 1
```
jlquaccia, line 339: Noticed on https://github.com/InteractiveAdvertisingBureau/openrtb/blob/main/extensions/community_extensions/Protected%20Audience%20Support.md, it states that the biddable field is 0 by default (indicating a buyer is not allowed).  Also saw in the prebid "fledge" slack channel (about a month back) that the biddable field default may get reversed to 1 instead.  Is there an update on the latest around this?

dgirardi: AFAIK there's no concrete plan for the spec to change.

patmmccann (approved):

PR #11292, opened 2024-04-04
Build Related : rename constants.json to constants.js

### LINKED ISSUE #10829: Clean up `constants.json`
## Type of issue
Improvement

## Description

[`constants.json`](https://github.com/prebid/Prebid.js/blob/master/src/constants.json) is included in every bundle, but its contents are not always needed:

 - A good portion of it is used only for native and should be excluded when native is disabled
 - `FLOOR_VALUES` is used exclusively by the pubmatic analytics adapter

muuki88:
## Type of change

- [x] Build related changes

## Description of change

Rename `constants.json` to `constants.js`

TODOs left

- [x] export separate consts
- [ ] if something imported only by one module, it should be moved into that module
- [ ] if something's imported only by modules, it should be moved into libraries

## Other information

- #10829


gulpfile.js
```diff
@@ -91,7 +91,7 @@ function lint(done) {
     '!plugins/**/node_modules/**',
     './*.js'
   ], { base: './' })
-    .pipe(eslint({ fix: !argv.nolintfix, quiet: !(typeof argv.lintWarnings === 'boolean' ? argv.lintWarnings : true) }))
+    .pipe(eslint({ fix: !argv.nolintfix, quiet: argv.lintWarnings === 'false' }))
```
muuki88, line 94: This setting wasn't working as `lintWarnings` was always of type `string`.

```bash
gulp lint --lintWarnings=false
```

now works as expected

muuki88:
```
'Uncaught exception:', TypeError: Cannot read properties of undefined (reading 'AUCTION_INIT')
```

nooo

gulpfile.js
dgirardi: it's a boolean if you use `--no-lint-warnings`.

muuki88:
Running a single test works

```bash
gulp test --file "test/spec/modules/yieldoneBidAdapter_spec.js" --nolint
```

running all tests, it fails. yyyyyyyyyyyyyy .... @dgirardi help plz

src/constants.js
```diff
@@ -0,0 +1,196 @@
+export default {
```
dgirardi, line 1: I think this is the same as JSON for the minifier. It wouldn't be able to remove dead definitions; I'm curious and I'd bet it also wouldn't fix the problem you were having. 

If it doesn't, what we want is 

```
export const EVENTS = {...}; 
export const TARGETING_KEYS = {...};
...
```

or even better would be 

```
export const EVENT_AUCTION_INIT = 'auctionInit';
...
export const TARGETING_KEY_BIDDER = 'hb_bidder';
...
```

but of course the latter is a lot more work. I'm sorry I didn't make this clearer earlier.

muuki88:
I found the issue why `constants` is not defined. I messed up some imports

src/constants.js
```diff
@@ -0,0 +1,214 @@
...
+  PL_SIZE: 'sizes',
+  PL_BIDS: 'bids',
+  BD_BIDDER: 'bidder',
+  BD_ID: 'paramsd',
+  BD_PL_ID: 'placementId',
+  ADSERVER_TARGETING: 'adserverTargeting',
+  BD_SETTING_STANDARD: 'standard'
+};
+
+export const FLOOR_SKIPPED_REASON = {
```
dgirardi, line 12: this should be in `priceFloors`

src/constants.js
```diff
@@ -0,0 +1,214 @@
...
+  RANDOM: 'random'
+};
+
+export const DEBUG_MODE = 'pbjs_debug';
+
+export const STATUS = {
+  GOOD: 1
+};
+
+export const CB = {
```
dgirardi, line 23: this appears to be unused

src/constants.js
```diff
@@ -0,0 +1,214 @@
...
+export const NATIVE_KEYS_THAT_ARE_NOT_ASSETS = [
+  'privacyIcon',
+  'clickUrl',
+  'sendTargetingKeys',
+  'adTemplate',
+  'rendererUrl',
+  'type'
+];
+
+export const FLOOR_VALUES = {
```
dgirardi, line 199: this should be in `pubmaticAnalyticsAdapter`

src/constants.js
muuki88: Move :+1:

src/constants.js
muuki88: Removed :wastebasket:

src/constants.js
muuki88: Move :truck:

GeneGenie:
This is such a win for my current research, i think i could have branched from here and go fully on node.

muuki88:
Awesome 🥳

Is there anything I can do to push this? Otherwise updating and resolving merge conflicts becomes cumbersome.

And we hit yet another issue with a bidder that doesn't work since the PUC refactoren, which is probably caused by some webpack minification, trimming away actually required constants 🫠

ChrisHuie (approved):

muuki88:
Thanks @ChrisHuie 
thanks

PR #11341, opened 2024-04-15
PAAPI: emit addComponentAuction event

ETNOL:
## Type of change

- [ ] Bugfix
- [x] Feature
- [ ] New bidder adapter   
- [ ] Build related changes
- [ ] CI related changes

- [ ] Other

## Description of change

Component Auction Configs appear to be entirely contained within the
PAAPI + fledgeForGPT modules. This event provides a way to access
those configs from outside of those 2 modules.

### DIFF
--- a/src/adapters/bidderFactory.js
+++ b/src/adapters/bidderFactory.js
@@ -549,6 +549,7 @@ export const registerSyncInner = hook('async', function(spec, responses, gdprCon
 }, 'registerSyncs')
 
 export const addComponentAuction = hook('sync', (request, fledgeAuctionConfig) => {
+  events.emit(CONSTANTS.EVENTS.ADD_COMPONENT_AUCTION, request, fledgeAuctionConfig);
 }, 'addComponentAuction');
 
 // check that the bid has a width and height set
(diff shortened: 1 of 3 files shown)


src/adapters/bidderFactory.js
```diff
@@ -549,6 +549,7 @@ export const registerSyncInner = hook('async', function(spec, responses, gdprCon
 }, 'registerSyncs')
 
 export const addComponentAuction = hook('sync', (request, fledgeAuctionConfig) => {
+  events.emit(CONSTANTS.EVENTS.ADD_COMPONENT_AUCTION, request, fledgeAuctionConfig);
```
dgirardi, line 552: With https://github.com/prebid/Prebid.js/pull/11277 , this is being expanded (and renamed) to also support adding single buyers.

There's also the fact that if the paapi module is not installed, this data is just dropped - I'm not sure the event makes sense here (the component auction is not really "added").

ETNOL: Does the need make sense? Prebid currently will event for things like new BidRequests or BidResponses, but I can't find any event that includes the fledgeAuctionConfigs. The paapi.js module doesn't seem ideal either, since a fledgeAuctionConfig can be returned from a BidAdapter without that module included.

Namespace-wise, I just followed with the name of the hook. If one name doesn't make sense, not sure if the other does either, right?

PR #10819, opened 2023-12-11
Core: rendering logic overhaul, PUC-less native rendering

dgirardi:
## Type of change

- [x] Feature

## Description of change

This introduces a new rendering flow, described in more detail [here](https://github.com/prebid/Prebid.js/blob/46a922de778aa6c2cdd69983207ed6a6c02dda0a/creative/README.md)

The immediate upshot is that this makes it possible to render native bids without PUC, and/or using `pbjs.renderAd`. In the long run, the goal is to (mostly) remove the necessity to update PUC every time we need to update our rendering logic (a solution different from, but addressing the same problem described [here](https://github.com/prebid/prebid-universal-creative/issues/187)).

## Testing

The PR includes some new or updated integration examples for the "no adserver" case. I did not do the same for remote creatives as the changes are entirely in GAM; if it helps, I have some line items set up for my own testing:

 - this test page uses a creative that is set up exactly as described in `integrationExamples/x-domain/creative.html`:
    <details>
    <summary>show</summary>
    <pre>

        <html>
        <head>
            <script>
                var PREBID_TIMEOUT = 60000;
        
                var adUnits = [
                    {
                        code: 'dgirardi-test',
                        mediaTypes: {
                            banner: {
                                sizes: [[300, 250]]
                            }
                        },
                        bids: [
                            {
                                bidder: 'rubicon',
                                params: {
                                    accountId: '1001',
                                    siteId: '267318',
                                    zoneId: '1861698'
                                }
                            },
                            {
                                bidder: 'adplus',
                                params: {
                                    inventoryId: '-1',
                                    adUnitId: '-3',
                                },
                            },
                            {
                                bidder: 'appnexus',
                                params: {
                                    placementId: 13144370
                                }
                            }
                        ]
                    }
                ];
        
                var pbjs = pbjs || {};
                pbjs.que = pbjs.que || [];
            </script>
            <script type="text/javascript" src="../../../build/dev/prebid.js" async></script>
        
            <script>
                var googletag = googletag || {};
                googletag.cmd = googletag.cmd || [];
                googletag.cmd.push(function () {
                    googletag.pubads().disableInitialLoad();
                });
        
                pbjs.que.push(function () {
                    pbjs.addAdUnits(adUnits);
                    pbjs.setConfig({
                        debug: true,
                    });
                });
        
                function sendAdserverRequest() {
                    if (pbjs.adserverRequestSent) return;
                    pbjs.adserverRequestSent = true;
                    googletag.cmd.push(function () {
[...]

PR #11343, opened 2024-04-15
Core: include dynamic renderer in native messages

dgirardi:
## Type of change

- [x] Feature

## Description of change

With https://github.com/prebid/prebid-universal-creative/pull/229, PUC can use the same dynamic renderes introduced in https://github.com/prebid/Prebid.js/pull/10819.

However, native uses a different flow. This includes renderers in native messages so that dynamic renderes can be used without the need to update creatives.

## Other information

https://github.com/prebid/prebid-universal-creative/issues/187


osazos (approved):
LGTM

ChrisHuie (approved):

PR #11362, opened 2024-04-19
Adnuntius Bid Adapter: allow dev end point option for adserver calls

antosarho:
## Type of change

- [x] Feature

## Description of change

Enables using the dev adserver for ad delivery for testing purposes.


modules/adnuntiusBidAdapter.js
```diff
@@ -258,10 +259,11 @@ export const spec = {
 
     const requests = [];
     const networkKeys = Object.keys(networks);
+    const devEnv = config.getConfig().env === 'dev';
```
patmmccann, line 262: Can we make this instead dependent on the request or bidder settings in someway? Either the test flag in openrtb https://github.com/InteractiveAdvertisingBureau/openrtb2.x/blob/72b31b0202ebd2ad8d27a0412a54abf6a6af712b/2.6.md?plain=1#L741 or in your bidder settings? We're hoping to avoid / curtail bidders calling getConfig

PR #11367, opened 2024-04-19
Sirdata RTD Module : fix empty bidders param array bug & various optimizations

nouchy:
## Type of change
- [X] Bugfix
- 
## Description of change
There was a bug relating to the module configuration parameters, when the publishers adds a filter for bidders and this filter is empty (empty array). We added a control on the length of the array which must be greater than 0, and optimized the code by removing parameters that had become obsolete.


osazos (approved):
LGTM

PR #11372, opened 2024-04-22
UtiqMtpIdSystem: create the Utiq MTP ID submodule

jdelhommeau:
## Type of change
- [X] Feature: Adding a new ID Submodule: UtiqMTPIdSystem

## Description of change

Creating a new ID Submodule, UtiqMTP. This ID is complementary to the other Utiq ID Module. It returns a different version of the Utiq ID (first party ID) which can be leveraged by publishers for additional use cases. This ID modules comes with its own bidder restriction, etc.


ChrisHuie:
@jdelhommeau do not see any docs around this id in our docs repo (). Can you please add https://github.com/prebid/prebid.github.io

modules/utiqMtpIdSystem.md
```diff
@@ -0,0 +1,15 @@
+## Utiq User ID Submodule
+
+Utiq MTP ID Module.
+
+First, make sure to add the utiq MTP submodule to your Prebid.js package with:
```
smenzer, line 5: This submodule seems to require additional setup by publishers to get the utiqMtpId placed in local storage. That extra installation should be made clear in the docs here as well as in the documentation repo.

smenzer (changes requested):
just the one comment from me on documentation, otherwise the module code itself looks fine.

modules/utiqMtpIdSystem.md
jdelhommeau: No there isn't any extra setup. Utiq script (external to prebid) that is installed separately by publisher already save the id in localstorage. The prebid module simply use prebid localstorage api to access the value when available.

smenzer: so that's my point - if an external script is required for this to work, you should mention that in the documentation (pbjs and prebid docs) so that publishers are aware they can't just turn this on without registering with utiq and installing additional code on their site.

jdelhommeau: ah sorry, I misunderstood your comment. Yes that makes perfect sense. I just added some extra information in the module documentation to notify publishers that some extra script configuration is required for the module to operates and provided link to utiq contact form if they are interested. Let me know if that address your point.

smenzer (approved):
LGTM

Issue #11365, opened 2024-04-19
Standalone Debugging Issue after v8.43.0


bozghiyy:
## Type of issue
Bug debugging when using standalone library after prebid 8.43.0

## Description
The standalone version of debugging hosted on jsdeliver has changed considerable and the new one doesn't work and freezes the auction. 
Check out (even if minimized, the files clearly are different):
https://cdn.jsdelivr.net/npm/prebid.js@8.42.0/dist/debugging-standalone.js
vs
https://cdn.jsdelivr.net/npm/prebid.js@8.43.0/dist/debugging-standalone.js

Probably is due the building process and the changes made in webpack.conf.js here: [https://github.com/prebid/Prebid.js/pull/11269/files#diff-49d59f0d7ac5380eec649778a04447e20fd504be848b640e42a70ab2675dc7a1](https://github.com/prebid/Prebid.js/pull/11269/files#diff-49d59f0d7ac5380eec649778a04447e20fd504be848b640e42a70ab2675dc7a1)

Probably debugging-standalone needs an exception. 

## Steps to reproduce
To test that is not working: just used a simple debugging configuration on a prebid build that uses the standalone debugging intercept, like:
`            pbjs.que.push(function() {
                pbjs.setConfig({
                    debugging: {
                        enabled: true,
                        intercept: [
                            {
                                when: {src: 'client'},
                                then: {
                                    cpm: 20
                                }
                            }
                        ]
                    }
                })
            });`

## Test page
Working with prebid version 8.42.0: https://jsfiddle.net/skruthmw/1/
Not working with prebid version 8.43.0: https://jsfiddle.net/skruthmw/0/
The prebid version is only relevant since it changes the jsdeliver URL location from where debugging-standalone.js is loaded. 

### Expected results
8.43.0 to work as on 8.42.0

### Actual results
Not only that the interceptor doesn't work but the auction freezes and nothing happens. Some ads are still displayed due to GAM FAIL SAFE timeout. 

## Platform details
All

## Other information
-

patmmccann:
Thank you for the detailed issue report!

PR #11375, opened 2024-04-22
Build system: fix standalone debugging & inclusion of node dependencies

dgirardi:
## Type of change

- [x] Bugfix

## Description of change

Fixes https://github.com/prebid/Prebid.js/issues/11365, as well as an issue where node dependencies that are only needed by some modules (e.g. live-connect) are included in core.


osazos (approved):
LGTM

tested on the JSFiddle examples provided in the issue by using a fresh build

PR #11379, opened 2024-04-22
New PAAPI module: topLevelPaapi

dgirardi:
## Type of change

- [x] Feature

## Description of change

- update PAAPI `requestedSize` handling as described in https://github.com/prebid/Prebid.js/issues/11189#issuecomment-2034716920
- introduce a new module, `topLevelPaapi`, intended for pubs running top level auctions. Currently, the only piece of configuration is a partial auction config:
   ```
   pbjs.setConfig({
     paapi: {
       topLevelSeller: {
          auctionConfig: {
              seller: 'example.seller',
              decisionLogicURL: 'https://example.seller/dl.js',
              // ...
          }
       }
     }
   })
   ```
   `auctionConfig` is reflected by `getPAAPIConfig()`.
- introduce a new public api, `pbjs.getPAAPIBids()`, which accepts the same filters as `getPAAPIConfig`, and returns a promise to a map from ad unit code to paapi auction result (obtained by calling `runAdAuction`).
  The auction result is an object with the format `{adUnitCode, auctionId, width, height, urn, frameConfig}` (where `urn` and `frameConfig` are mutually exclusive)
- introduce new events `paapiRunAuction`, `paapiBid`, and `paapiNoBid`
- update the top level auction integration example to use the above.

## Other information

Closes https://github.com/prebid/Prebid.js/issues/11189
Related: https://github.com/prebid/Prebid.js/issues/10690


pm-harshad-mane:
@jlquaccia please review the PR

integrationExamples/top-level-paapi/gam-contextual.html
```diff
@@ -0,0 +1,134 @@
...
+                    enabled: true,
+                    gpt: {
+                        autoconfig: false
+                    },
+                    topLevelSeller: {
+                        auctionConfig: {
+                            seller: window.location.origin,
+                            decisionLogicURL: new URL('shared/decisionLogic.js', window.location).toString(),
+                        },
+                        overrideWinner: true
```
jlquaccia, line 63: So this would mean, if pbjs has a tls pa winner, go with that instead of a possble gam contextual winner?  right?

dgirardi: Yes - depending on what you mean by GAM contextual winner. It means that when Prebid is asked to render a bid, it will instead render a PAAPI winning bid if it has one for the slot. 

In the typical GAM flow (at least typical in our examples, including this one) the bid that is potentially replaced is the Prebid bid that won the Prebid auction, and then the GAM auction, to be replaced by the winner of the PAAPI auction.

patmmccann (approved):

PR #11383, opened 2024-04-23
DeepIntent Id Module : fix user ids not being passed on page reload due to getValue func miss

parthshah51999:
## Type of change

- [x] Bugfix
- [ ] Feature
- [ ] New bidder adapter   
- [ ] Build related changes
- [ ] CI related changes

- [ ] Other

## Other information

On page reload, getId is not called from Id system submodule and it expects function extendId to provide the logic to read the id from storage. This is not implemented in deepintentDpesSubmodule which causes response Object to be blank. Because this is blank storedId doesn't get updated to response.id

local storage format: __key__ : {"id": "123123"}


smenzer:
the user id module _should_ pull the ID from cache automatically if using the core id module to save it (as opposed to doing it manually); extendId should not be required unless you're trying to do something extra with the cached value on every page, which in this PR, it doesn't look like you're doing. so it seems like there's an issue with the page setup or something else is going on to cause the issue you're seeing.

parthshah51999:
> the user id module _should_ pull the ID from cache automatically if using the core id module to save it (as opposed to doing it manually); extendId should not be required unless you're trying to do something extra with the cached value on every page, which in this PR, it doesn't look like you're doing. so it seems like there's an issue with the page setup or something else is going on to cause the issue you're seeing.

Thanks for your reaponse, @smenzer. Unfortunately this does not look like a setup issue. I am able to replicate this issue locally as well. Here's how the flow works:

1. DeepIntent script is responsible for setting the token in local storage. This is done in the format: __key__ : {"id": "123123123"}.
2. In the prebid code, "populateSubmoduleId" method is called in the init flow. In the intial flow idSystem.getId gets called. This returns the {id: "123123123"} object.
3. Post this we check the response is a plainObject and update the local storage to __key__: "123123123". Could you please let me know why this is done as this creates an inconsistency. On page reload, getId is no longer called but our script updates the value back to {id: "123123123"}. Hence storedId gets 2 different values. Initially it gets the string value, later it starts getting the object.
4. When we get an object, the createEidObject in file eids.js no longer is able to read the id as it expects a string. Implemented the getValue function to ensure that if we get an object read from it otherwise if it is a string (which prebid updates in local storage) return it as it is.
5. For now I have added the conditional check, and it resolves our problem of eids not being read. But why would we make an update to the local storage ? And if we need to do this can we not ensure that storedId is always a string in populateSubmoduleId ?

smenzer:
i think we probably need an engineer that knows the details of the module a bit better than myself (product manager) to help out here, @parthshah51999. i've requested that @dgirardi take a look to see if he can add some value.

dgirardi:
The issue is that storage is meant to be managed by Prebid. On the other hand, I am not able to find a developer's guide for userId modules, so it's not surprising that this generates confusion. 

This is the relevant piece of logic:
https://github.com/prebid/Prebid.js/blob/fadc303fe10a8d5327590da115a1b519f88779c4/modules/userId/index.js#L769-L791

In the "normal" flow, line 769 does not find anything in storage, `getId` is called, and the return value's `.id` is then stored in line 789. On refresh, it branches off to `extendId` if present.

In your case, an external script populates the cookie, but in a different way (stringified `JSON.stringify({id})` instead of just the id). `getId` is still called because some other pieces of metadata are missing (`consentChanged`). Then the cookie is overwritten in line 789, and on refresh it's not what you expect.

TL;DR how to fix?

The "correct" way is to not use `storage` configuration, if storage is not managed by Prebid, and instead read the cookie directly from the module. The issue with this mixed setup is that the `userSync.userIds.storage` configuration is meant to let the publisher choose when and how the ID is stored; but in your case, I don't think they have a choice, it needs to agree with what your script does, and as it is it just gives them an opportunity to get it wrong.

The easy fix is to update 
https://github.com/prebid/Prebid.js/blob/fadc303fe10a8d5327590da115a1b519f88779c4/modules/deepintentDpesIdSystem.js#L46-L48

to 

```
    if (cacheIdObj?.id) return cacheIdObj;
    return cacheIdObj ? {id: cacheIdObj} : cacheIdObj;
```

but note that this will still overwrite your cookie, so if you store anything else, or want to manage expiration, you'll still have problems.

parthshah51999:
> The issue is that storage is meant to be managed by Prebid. On the other hand, I am not able to find a developer's guide for userId modules, so it's not surprising that this generates confusion.
> 
> This is the relevant piece of logic:
> 
> https://github.com/prebid/Prebid.js/blob/fadc303fe10a8d5327590da115a1b519f88779c4/modules/userId/index.js#L769-L791
> 
> In the "normal" flow, line 769 does not find anything in storage, `getId` is called, and the return value's `.id` is then stored in line 789. On refresh, it branches off to `extendId` if present.
> 
> In your case, an external script populates the cookie, but in a different way (stringified `JSON.stringify({id})` instead of just the id). `getId` is still called because some other pieces of metadata are missing (`consentChanged`). Then the cookie is overwritten in line 789, and on refresh it's not what you expect.
> 
> TL;DR how to fix?
> 
> The "correct" way is to not use `storage` configuration, if storage is not managed by Prebid, and instead read the cookie directly from the module. The issue with this mixed setup is that the `userSync.userIds.storage` configuration is meant to let the publisher choose when and how the ID is stored; but in your case, I don't think they have a choice, it needs to agree with what your script does, and as it is it just gives them an opportunity to get it wrong.
> 
> The easy fix is to update
> 
> https://github.com/prebid/Prebid.js/blob/fadc303fe10a8d5327590da115a1b519f88779c4/modules/deepintentDpesIdSystem.js#L46-L48
> 
> to
> 
> ```
>     if (cacheIdObj?.id) return cacheIdObj;
>     return cacheIdObj ? {id: cacheIdObj} : cacheIdObj;
> ```
> 
> but note that this will still overwrite your cookie, so if you store anything else, or want to manage expiration, you'll still have problems.

Hey @dgiradi, thanks this works. Also noted that the cookie value is consistently overwritten post the fix. We are okay with this for now and will follow up if we need to make additional updates.

Quick question just for my understanding: 
I noticed that after the recommended change the populateSubmoduleId function gets called twice and in the second call, forceRefresh param is passed as true which ensures that the flow works as expected. I did not understand the flow for this could you point me to the relevant piece of code so I can understand this better.

dgirardi:
> I noticed that after the recommended change the populateSubmoduleId function gets called twice and in the second call, forceRefresh param is passed as true which ensures that the flow works as expected

`forceRefresh` should not be needed for the flow to work, and whether it's set depends on your setup (in my case it isn't and the call happens once).

It's true when:

- the user explicitly asks for a refresh with `refreshUserIds`
    https://github.com/prebid/Prebid.js/blob/51bcf81bc6212de26290168a789c6db16d53af90/modules/userId/index.js#L696-L714
- a userId module is installed after it has already been configured
    https://github.com/prebid/Prebid.js/blob/51bcf81bc6212de26290168a789c6db16d53af90/modules/userId/index.js#L1033-L1042

parthshah51999:
> You are right - my suggestion only works as long as you don't already have a cookie set.
> 
> I think (a) technically the best solution, but what you have should also work. (I was worried that it would only work for eids and not the "legacy" `userId` object, but reading through the code I think it happens to not matter).

I had implemented the extendId in the first iteration of this PR but this was suggested - https://github.com/prebid/Prebid.js/pull/11383#issuecomment-2072365055

I can update the PR if necessary to implement the extendId function. But if there is no downside you foresee with this change this seems okay to me as well. Let me know how to proceed here.

parthshah51999:
> This is OK, except for the tests - as I noted above, think most of the suite is not actually testing anything.

Cool I will update the tests, thanks for your help!

parthshah51999:
@dgirardi @ChrisHuie any specific reason this was not picked for the 8.47.0 release ?

ChrisHuie:
> @dgirardi @ChrisHuie any specific reason this was not picked for the 8.47.0 release ?

Don't believe this pr was approved when I released Wednesday morning but merging now so it is included with this week's release.

PR #11390, opened 2024-04-24
Appnexus Bid Adapter: add support for ast_override_div debug feature

astartsky:
## Type of change
- [x] Feature

## Description of change
add support for "ast_override_div" debug feature to the adapter


jsnellbaker (approved):
Thank you for putting this together; it looks correct.

Issue #11387, opened 2024-04-23
${AUCTION_PRICE} macro is not being replaced


skapoorViant:
Hello Team,

Please find the sample response and event URL. The ${AUCTION_PRICE} macro is not being replaced in event tracking URL, though it is coming as encoded. 
`{
  "dtp": "rtb_response",
  "host_node_name": "i-0de5cb59043dd69d1",
  "host_node_ip": "10.55.8.214",
  "host_node_zone": "us-east-1b",
  "host_node_version": "2.0.5147soak",
  "ep": "/d/rtb/v25/prebid/bidder",
  "ts": "1713831808093",
  "sid": "1d160339-71a4-4aa3-ab4d-84597cf785fb-559ae0e9e63b7089",
  "id": "3c5f7d42-6bc3-45cf-bb5f-e99c3c04b202",
  "seatbid": [
    {
      "bid": [
        {
          "id": "1d160339-71a4-4aa3-ab4d-84597cf785fb-559ae0e9e63b7089",
          "impid": "559ae0e9e63b7089",
          "price": 0.700001,
          "adid": "58612-424808-2152925",
[...]

PR #11394, opened 2024-04-24
Called replaceAuctionMacro in  viantOrtbBidAdapter to replace Auction_Price

skapoorViant:
Called replaceAuctionMacro in  viantOrtbBidAdapter to replace Auction_Price

## Type of change

- [x] Bugfix

## Description of change

Auction price macro was not getting expanded for bids won with Viant Adapter. Explicitly called replaceAuctionPrice method from utils.js to get the correct value.

### DIFF
--- a/modules/viantOrtbBidAdapter.js
+++ b/modules/viantOrtbBidAdapter.js
@@ -51,8 +51,10 @@ export const spec = {
   onBidWon: function (bid) {
     if (bid.burl) {
       utils.triggerPixel(bid.burl);
+      utils.replaceAuctionPrice(bid.burl,bid.price);
     } else if (bid.nurl) {
       utils.triggerPixel(bid.nurl);
+      utils.replaceAuctionPrice(bid.nurl,bid.price);
     }
   }
 }
--- a/modules/viantOrtbBidAdapter.js
+++ b/modules/viantOrtbBidAdapter.js
@@ -51,10 +51,10 @@ export const spec = {
   onBidWon: function (bid) {
     if (bid.burl) {
       utils.triggerPixel(bid.burl);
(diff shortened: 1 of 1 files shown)


skapoorViant:
@dgirardi can you please review whenever you get a chance? We have a blocker on our side related to the Auction_Price macro issue [https://github.com/prebid/Prebid.js/issues/11387](url).

dgirardi (approved):

skapoorViant:
@patmmccann can you please merge this pull request, we have a blocker on our side related to AUCTION_PRICE macro, and we need your help to merge this PR soon.

modules/viantOrtbBidAdapter.js
```diff
@@ -51,8 +51,10 @@ export const spec = {
   onBidWon: function (bid) {
     if (bid.burl) {
       utils.triggerPixel(bid.burl);
+      utils.replaceAuctionPrice(bid.burl, bid.price);
```
dgirardi, line 54: @skapoorViant, I aplogize for thinking about this only after merging.

I don't think this does what you want. it should probably be

```
utils.triggerPixel(utils.replaceAuctionPrice(bid.burl, bid.originalCpm || bid.cpm))
```

you can do the same for `nurl`, but by the time this runs it will be a duplicate call.

skapoorViant: no worries, but in the bidresponse we are sending bid.price and not bid.originalCpm or bid.cpm, so would it be good if I just write it as 
`utils.triggerPixel(utils.replaceAuctionPrice(bid.burl, bid.price||bid.originalCpm || bid.cpm))`

PR #11398, opened 2024-04-25
ViantOrtb Bid Adapter : called replaceAuctionMacro to replace Auction_Price

skapoorViant:
Called replaceAuctionMacro in viantOrtbBidAdapter to replace Auction_Price

## Type of change

- [x] Bugfix

## Description of change
Auction price macro was not getting expanded for bids won with Viant Adapter. Explicitly called replaceAuctionPrice method from utils.js to get the correct value.

### DIFF
--- a/modules/viantOrtbBidAdapter.js
+++ b/modules/viantOrtbBidAdapter.js
@@ -51,10 +51,10 @@ export const spec = {
   onBidWon: function (bid) {
     if (bid.burl) {
       utils.triggerPixel(bid.burl);
-      utils.replaceAuctionPrice(bid.burl, bid.price);
+      utils.triggerPixel(utils.replaceAuctionPrice(bid.burl, bid.originalCpm || bid.cpm || bid.price));
     } else if (bid.nurl) {
       utils.triggerPixel(bid.nurl);
-      utils.replaceAuctionPrice(bid.nurl, bid.price);
+      utils.triggerPixel(utils.replaceAuctionPrice(bid.nurl, bid.originalCpm || bid.cpm || bid.price));
     }
   }
 }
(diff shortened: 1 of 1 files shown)


skapoorViant:
@dgirardi created this new pull request as per your feedback. Thank you.

modules/viantOrtbBidAdapter.js
```diff
@@ -51,10 +51,10 @@ export const spec = {
   onBidWon: function (bid) {
     if (bid.burl) {
       utils.triggerPixel(bid.burl);
-      utils.replaceAuctionPrice(bid.burl, bid.price);
+      utils.triggerPixel(utils.replaceAuctionPrice(bid.burl, bid.originalCpm || bid.cpm || bid.price));
```
dgirardi, line 54: (re: using `bid.price`)

it makes no difference, but to clarify how this works, I believe `bid.price` here is always undefined.

When `bid` refers to the ORTB object (in `seatbid`), `price` is the right name.

Adapters however deal with a different type of bid object (that predates ORTB), where the field is named `cpm` (see [developer guide](https://docs.prebid.org/dev-docs/bidder-adaptor.html#interpreting-the-response))

In your case `interpretResponse` uses ortbConverter, which manages the translation:
https://github.com/prebid/Prebid.js/blob/0fd106dea29b7cf70be35136c5e567782166cc8c/libraries/ortbConverter/processors/default.js#L77-L91

Prebid also copies that into `originalCpm`, because there's several things that can modify `cpm` after your adapter returns it (currency conversion, pub-defined bid adjustments).

I am now also noticing that `nurl` is not copied over by ortbConverter, so the second trigger here should never run. (But that's probably for the better, as it's included in the creative and by the time this method runs it should have already been fired).

dgirardi (approved):
LGTM. let me know if you want to clean up some of what I mentioned before we merge.

modules/viantOrtbBidAdapter.js
skapoorViant: @dgirardi Thanks a lot for your input. I will just remove bid.price then.

skapoorViant (commented):
fixed changes

skapoorViant:
@dgirardi i have cleaned it up. Please merge it.

PR #11402, opened 2024-04-26
Various Adapters: Delete s2s transform bid params for adapters with no server equivalent

patmmccann:
Progress on https://github.com/prebid/Prebid.js/issues/6361


patmmccann:
Prep for prebid 9 but since none of these s2s adapters, this isn't breaking

modules/big-richmediaBidAdapter.js
```diff
@@ -112,11 +112,6 @@ export const spec = {
     return baseAdapter.getUserSyncs(syncOptions, responses, gdprConsent);
   },
 
-  transformBidParams: function (params, isOpenRtb) {
-    if (!baseAdapter.transformBidParams) { return params; }
-    return baseAdapter.transformBidParams(params, isOpenRtb);
-  },
-
```
osazos, line 119: Just to confirm, this adapter extends [appnexusBidAdapter](https://github.com/prebid/Prebid.js/blob/master/modules/appnexusBidAdapter.js) which uses this function. Was this taken into consideration?

ChrisHuie (approved):

osazos (approved):

Issue #8217, opened 2022-03-25
[Feature Request] New Module for `device` from User-Agent


maphe:
## Type of issue
Feature Request

## Description
Is there any chance for a new Prebid Module that would generate a populated `device` open RTB object (as much as possible)? As of now we can get the User Agent string from the adapter but we cannot use a third party library that would parse the UA and make a comprehensive open RTB `device` object (things like `devicetype`, `make`, `model`, `h` and `w`, etc). Also doing some sort of custom parsing in the adapter would make maintenance a bit painful.

I imagine something similar to the User module that centralizes the generation of user ids.

I know it is doable server side from the UA string but having it client side could make the bids respond faster.

patmmccann:
Part of your request (h&w) is done: https://github.com/prebid/Prebid.js/pull/6452/commits/11492e99860fa5351a46702ed52cd8bab9dbb1de

patmmccann:
It is a huge file https://github.com/ua-parser/uap-core/blob/master/regexes.yaml ; mobile / not mobile seems easy to achieve but make and model seems unlikely, especially as the world moves to making these things less detailed, not more.

maphe:
Yeah we had some back and forth internally to figure out where we wanna do it, obviously doing it within the adapter is not realistic so we might end up with some server side processing but it's a trade off too as we're delaying the bid. Having some stuff provided by the prebid core would be nice, but I understand some are trickier to obtain.

For context, buyers seem to be pretty inconsistent on what they do themselves with the User-Agent, and the ones that don't parse it on their own seem to expect a more complete `device` object. I know device type, make and model have be mentioned lately. I can probably come back with a more detailed list of valuable attributes if you think there's a chance they get managed by the prebid core.

patmmccann:
I can't see this being in core, but a module to enrich the first party data in the ortb2 object is already a thing; this could be part of that module if it is pretty terse, or its own module if the parsing is expensive with regards to the number of regex's

maphe:
Ok great, I'll keep an eye out for it. Thank you

patmmccann:
Timeout RTD provider has this functionality, but appears wrong

function getDeviceType() {
  const userAgent = window.navigator.userAgent.toLowerCase();
  if ((/ipad|android 3.0|xoom|sch-i800|playbook|tablet|kindle/i.test(userAgent))) {
    return 5; // tablet
  }
[...]

PR #11414, opened 2024-04-29
51Degrees RTD submodule: initial commit 

jwrosewell:
## Type of change
- [x] Feature
- [x] New RTD submodule

## Description of change
51Degrees module enriches an OpenRTB request with [51Degrees Device Data](https://51degrees.com/documentation/index.html).

51Degrees module sets the following fields of the device object: `make`, `model`, `os`, `osv`, `h`, `w`, `ppi`, `pixelratio` - interested bidder adapters may use these fields as needed. In addition the module sets `device.ext.fiftyonedegrees_deviceId` to a permanent device ID which can be rapidly looked up in on premise data exposing over 250 properties including the device age, chip set, codec support, and price, operating system and app/browser versions, age, and embedded features.

The module supports on premise and cloud device detection services with free options for both. 

Integration details are specified in the `modules/51DegreesRtdProvider.md`.

## Maintainer contacts:

support@51degrees.com

## Other details

The features provided by this module have been requested previously as part of: https://github.com/prebid/Prebid.js/issues/8217


integrationExamples/gpt/51DegreesRtdProvider_example.html
```diff
@@ -0,0 +1,171 @@
+<!DOCTYPE html>
+<html lang="en">
+<head>
+    <meta http-equiv="Delegate-CH" content="sec-ch-ua-full-version-list https://cloud.51degrees.com; sec-ch-ua-model https://cloud.51degrees.com; sec-ch-ua-platform https://cloud.51degrees.com; sec-ch-ua-platform-version https://cloud.51degrees.com">
+    <script async src="//www.googletagservices.com/tag/js/gpt.js"></script>
```
patmmccann, line 5: Outdated url

integrationExamples/gpt/51DegreesRtdProvider_example.html
```diff
@@ -0,0 +1,171 @@
...
+                    }]
+                }
+            ];
+
+            pbjs.setConfig({
+                debug: true,
+                cache: {
+                    url: false
+                },
+                realTimeData: {
```
patmmccann, line 91: This whole block barely adds any value beyond your documentation. Are you sure you want to maintain it?

modules/51DegreesRtdProvider.js
```diff
@@ -0,0 +1,255 @@
...
+  }
+  return `https://cloud.51degrees.com/api/v4/${pathData.resourceKey}.js`;
+}
+
+/**
+ * Injects 51Degrees script into the document head
+ * @param {string} url 51Degrees JS URL
+ * @returns {Promise} Promise that resolves when the script is loaded
+ */
+export const inject51DegreesScript = (url) => {
```
patmmccann, line 84: You have to use our injection methods

modules/51DegreesRtdProvider.js
```diff
@@ -0,0 +1,255 @@
...
+ */
+export const getBidRequestData = (reqBidsConfigObj, callback, moduleConfig, userConsent) => {
+  try {
+    // Get the required config
+    const {resourceKey, onPremiseJSUrl} = extractConfig(moduleConfig, reqBidsConfigObj);
+    logMessage('Resource key: ', resourceKey);
+    logMessage('On-premise JS URL: ', onPremiseJSUrl);
+
+    // Get 51Degrees JS URL, which is either cloud or on-premise
+    const scriptURL = get51DegreesJSURL(resourceKey ? {resourceKey} : {onPremiseJSUrl});
```
patmmccann, line 200: If you're using prebid to inject, we prefer you lock versions and pr the project when the script changes

modules/51DegreesRtdProvider.js
```diff
@@ -0,0 +1,255 @@
...
+  document.head.appendChild(script);
+
+  return new Promise((resolve, reject) => {
+    script.onload = resolve;
+    script.onerror = reject;
+  });
+}
+
+/**
+ * Check if meta[http-equiv="Delegate-CH"] tag is present in the document head and points to 51Degrees cloud
```
patmmccann, line 97: What's going on here, why not just get the high entropy hints in js

patmmccann (changes requested):
A few remarks in line

integrationExamples/gpt/51DegreesRtdProvider_example.html
BohdanVV: Thanks for the heads up, I've removed this part

modules/51DegreesRtdProvider.js
justadreamer: Thanks for the comment.  Agreed that in general it's a good idea to use `getHighEntropyValues` API.  The injected script actually uses it as a fallback in the case when navigator did not send User-Agent Client Hints as part of the first request.  Having the navigator send the User-Agent Client Hints by itself seems to be a bit more performant and also transparent to the user that they give permission to a 3rd party to process those as prescribed by the [spec](https://wicg.github.io/client-hints-infrastructure/#delegate-ch-algo)

modules/51DegreesRtdProvider.js
justadreamer: @patmmccann thanks for the comment.  Just wanted to clarify:
1. by "our injection methods" - do you mean `loadExternalScript` API from `adloader.js`?   
2. If so we'd need to add our module code into `_approvedLoadExternalJSList`, I assume it should also be part of this PR, right? 
3. by "another method" do you mean that the user may insert the `script` tag themselves at the point of embedding Prebid.js into the page when providing configuration, or is it something else?

modules/51DegreesRtdProvider.js
justadreamer: The script is generated by the [51Degrees Cloud API](https://cloud.51degrees.com/api-docs/index.html) that is version-locked through its URL:  `https://cloud.51degrees.com/api/v4/<your_resource_Key>.js`, however the script itself is generated based on resource-key and properties user selects for it - so we would not be able to store its content as part of this repo.  The javascript template and device detection library used by the cloud are open source (as mentioned in our docs): 
- https://github.com/51Degrees/javascript-templates/
- https://github.com/51Degrees/device-detection-dotnet

integrationExamples/gpt/51DegreesRtdProvider_example.html
patmmccann: I'm confused, you still have this file in your pr?

modules/51DegreesRtdProvider.js
patmmccann: Yes, yes, yes

modules/51DegreesRtdProvider.js
```diff
@@ -0,0 +1,255 @@
...
+ */
+export const is51DegreesMetaPresent = () => {
+  const meta51 = document.head.querySelectorAll('meta[http-equiv="Delegate-CH"]');
+  if (!meta51.length) {
+    return false;
+  }
+  return Array.from(meta51).some(
+    meta => !meta.content
+      ? false
+      : meta.content.includes('https://cloud.51degrees.com')
```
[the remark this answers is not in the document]
patmmccann: @BohdanVV take note

integrationExamples/gpt/51DegreesRtdProvider_example.html
justadreamer: Hi @patmmccann, sorry we thought you meant just the cache block in that configuration json,  but in general yes - we'd like to maintain this example page as it is the way to demo and test the module.  For now this is convenient as not many client-side adapters implement echoing the full oRTB device object to their servers.   We'll add also the output of the enrich oRTB object to the page itself - should be handy for testing on mobile devices.  Thanks

modules/51DegreesRtdProvider.js
bretg: FWIW - PBJS-core will add high entropy device.sua if the publisher sets some config...

```javascript
pbjs.setConfig({
       firstPartyData: {
         uaHints: ['platform', 'architecture', 'model', 'platformVersion', 'fullVersionList']
       }
});
```

This may be something you want to suggest in your doc?

justadreamer: Thanks for raising this, @bretg. A lengthy comment ahead, but we hope it will become a basis for more documentation. Breaking it down below. 

## PBS modules

The [First Party Data Enrichment](https://docs.prebid.org/dev-docs/modules/enrichmentFpdModule.html) module with uaHints mentioned above is very relevant for 51D Prebid Server Modules, and we'll actually add it to the server modules docs, because 51D PBS modules are more precise when `device.sua` is present in the request.

## Prebid.js module

51D Prebid.js module injects a dynamically generated script based on the evidence available. It also falls back to the GetHighEntropyValues API, but does not rely on it as a first (or only) choice route - please see the illustrative cases below.  Albeit it seems easier, GHEV API is not supported by all browsers (so the decision to call it should be conditional) and also even in Chrome this API will likely be a subject to the Privacy Budget in the future.

In summary we use `Delegate-CH` http-equiv as the preferred method of obtaining the necessary evidence because it is the fastest method, and future proof. `Delegate-CH` enables the Chromium-based browser to send the User-Agent Client Hints to a 3rd party right away on the very first request unlike `Accept-CH` + `Permissions-Policy` http response headers which would let it send them on subsequent requests. Hence this check in the code that outputs the warning in debug mode. 

## Illustrative Cases

* if the device is iPhone/iPad then there is no point checking for or calling GetHighEntropyValues at the moment because iOS does not support this API. However this might change in the future.  
Platforms like iOS require additional techniques to identify the model which are not covered via a single API call, and change from version to version of the operating system and browser rendering engine. **When used with iOS 51Degrees resolves the [iPhone/iPad model groups](https://51degrees.com/documentation/4.4/_device_detection__features__apple_device_table.html) using these techniques.** That is one of the benefits the module brings to the Prebid community as most solutions do not resolve iPhone/iPad model groups. More on Apple Device Detection [here](https://51degrees.com/documentation/4.4/_device_detection__features__apple_detection.html).

* if the browser is Brave on Android then there is similarly no point requesting GHEV as the API is not supported.

* if the browser is Chrome then the `Delegate-CH` if enabled by the publisher would enable the browser to provide the necessary evidence. However if this is not implemented - then the dynamic script would fall back to GHEV which is slower.

As the code for these features changes frequently the relevant snippets are in the data file rather than the core source code. As the web diverges the ability to dynamically adapt is going to be very important. Reducing the depency on Javascript and maximising server side stable controlled environments is part of the 51Degrees approach.

patmmccann: @jwrosewell in general it is bad form to mark reviewer comments as resolved. We let reviewers decide if their comments are resolved.

patmmccann: Unresolving to make easier to find for future readers.

integrationExamples/gpt/51DegreesRtdProvider_example.html
```diff
@@ -0,0 +1,185 @@
...
+                        bidder: 'appnexus',
+                        params: {
+                            placementId: 13232392,
+                        }
+                    }]
+                }
+            ];
+
+            pbjs.setConfig({
+                debug: true,
```
patmmccann, line 87: this is a tiny bit dangerous in an example; i would add a comment here to tell devs to be careful not to run debug in production

modules/51DegreesRtdProvider.js
patmmccann: Can you dm me a copy of something running; I want to check for example, you don't gather bids or identifiers from other companies and send them to your endpoint, or other obvious rules violations.

modules/51DegreesRtdProvider.md
```diff
@@ -0,0 +1,130 @@
...
+```
+gulp build --modules="rtdModule,51DegreesRtdProvider,appnexusBidAdapter,..."  
+```
+
+> Note that the 51Degrees RTD module is dependent on the global real-time data module, `rtdModule`.
+
+### Prerequisites
+
+#### Resource Key
+In order to use the module please first obtain a Resource Key using the [Configurator tool](https://configure.51degrees.com/tWrhNfY6) - choose the following properties:
```
patmmccann, line 38: Configurator is a rather awkward word, I recommend you choose a word more common in English

patmmccann (approved):
Code LGTM. please add the tiny warning about debug to your example, please add the external script disclosure prominently on yours docs PR. Please add the long discussion on entropy hints vs headers to your documentation somewhere reasonably accessible.

Issue #11399, opened 2024-04-25
Various modules: stringify and parse json is slow


patmmccann:
https://github.com/search?q=repo%3Aprebid%2FPrebid.js+json.parse%28json.stringify+path%3A%2F%5Emodules%5C%2F%2F&type=code is used by 17 modules. Several of these patterns can be improved similar to this example: https://github.com/InteractiveAdvertisingBureau/iabtcf-es/pull/446

patmmccann:
Reopening as we've had some success with the klona library we plan to commit

patmmccann:
Cc @bbaresic

dgirardi:
JSON.parse + JSON.stringify used to be the fastest way to deep clone objects. Apparently it [still is](https://www.google.com/search?channel=fs&client=ubuntu-sn&q=structuredClone+vs+JSON+parse+performance). 

We are probably deep cloning when it's not needed, but it's often not trivial to verify when it isn't, and I don't think it makes sense to after the most responsible way to do it. In general premature optimization is a bad idea, it should be focused on known (= measured) bottlenecks.

bbaresic:
Here is a [benchmark](https://github.com/bbaresic/deep-clone-benchmarks) that I did. It tries to mimic a Prebid object (several levels of nesting). 

It is a micro-benchmark, and that comes with some caveats, but as seen there, `klona` outperforms everything.

I created a [PR](https://github.com/prebid/Prebid.js/pull/11418) for it.

patmmccann:
@dgirardi Raptive measured some bottlenecks in just-clone, @bbaresic has already tested this change on live traffic to substantial improvement

dgirardi:
Closed with https://github.com/prebid/Prebid.js/pull/11418/files

PR #11418, opened 2024-04-29
PBjs Utils: Faster Deep Clone

bbaresic:
## Type of change

## Description of change

Prebid currently uses an outdated version of just-clone for deep cloning objects.

After some [benchmarking](https://github.com/bbaresic/deep-clone-benchmarks), klona seems much more efficient even compared to other libraries or even native JS (`structuredClone` and `JSON.parse/stringify`)

This PR replaces `just-clone`, but the other way of deep-cloning `JSON.parse(JSON.stringify(object))` stays unchanged.
That should be done in a different PR.

This partially addresses [#11399](https://github.com/prebid/Prebid.js/issues/11399).

### DIFF
--- a/modules/browsiRtdProvider.js
+++ b/modules/browsiRtdProvider.js
@@ -57,7 +57,7 @@ export function addBrowsiTag(data) {
   script.setAttribute('prebidbpt', 'true');
   script.setAttribute('id', 'browsi-tag');
   script.setAttribute('src', data.u);
-  script.prebidData = deepClone(data);
+  script.prebidData = deepClone(typeof data === 'string' ? Object(data) : data)
   if (_moduleParams.keyName) {
     script.prebidData.kn = _moduleParams.keyName;
   }
--- a/src/utils.js
+++ b/src/utils.js
@@ -1,5 +1,5 @@
 import {config} from './config.js';
-import clone from 'just-clone';
+import {klona} from 'klona/json';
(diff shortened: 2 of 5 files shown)


modules/browsiRtdProvider.js
```diff
@@ -57,7 +57,7 @@ export function addBrowsiTag(data) {
   script.setAttribute('prebidbpt', 'true');
   script.setAttribute('id', 'browsi-tag');
   script.setAttribute('src', data.u);
-  script.prebidData = deepClone(data);
+  script.prebidData = deepClone(typeof data === 'string' ? Object(data) : data)
```
bbaresic, line 60: Version 1 of just-clone casts everything to object, even though in this case it was a string.

The newer versions don't do that, just like klona.

This is here to preserve that behavior.

src/utils.js
```diff
@@ -609,7 +609,7 @@ export function shuffle(array) {
 }
 
 export function deepClone(obj) {
-  return clone(obj);
+  return klona(obj) || {};
```
bbaresic, line 612: Version 1 of just-clone gives back an empty object when the cloning fails.

The newer versions don't do that, just like klona.

This is here to preserve that behavior.

patmmccann (approved):

mmoschovas (approved):

PR #11425, opened 2024-04-30
PubxAi Analytics Adapter : code cleanup and additional data collection

pnhegde:
## Type of change
- [x] Bugfix
- [ ] Feature
- [ ] New bidder adapter   
- [ ] Build related changes
- [ ] CI related changes
- [ ] Other

## Description of change
Code cleanup and refactoring, renaming of fields, and additional data collection as part of analytics.


pnhegde:
@ChrisHuie All the tests have passed already. Anything else needed from our end? Thanks.

modules/pubxaiAnalyticsAdapter.js
```diff
@@ -1,208 +1,429 @@
...
+      return target[name];
+    },
+  }
+);
+
+/**
+ *
+ * @returns {boolean} whether or not the browser session supports sendBeacon
+ */
+const hasSendBeaconSupport = () => {
```
patmmccann, line 95: Please use our methods

modules/pubxaiAnalyticsAdapter.js
```diff
@@ -1,208 +1,429 @@
...
+    );
+  }
+  return {}; // TODO: support more ad servers
+};
+
+/**
+ * Access sessionStorage
+ * @returns {Storage}
+ */
+const getStorage = () => {
```
patmmccann, line 128: This is a fairly serious rules violation, please correct immediately

patmmccann: Please submit another pr fixing this issue or risk unplanned changes to your adapter, you may not bypass our stage methods

modules/pubxaiAnalyticsAdapter.js
```diff
@@ -1,208 +1,429 @@
...
-      ajax(pubxaiAnalyticsRequestUrl, undefined, JSON.stringify(data), { method: 'POST', contentType: 'text/json' });
-      auctionCache.push(data.initOptions.auctionId);
+
+    events.splice(0);
+  });
+};
+
+// register event listener to send logs when user leaves page
+if (hasSendBeaconSupport()) {
+  document.addEventListener('visibilitychange', () => {
```
patmmccann, line 401: Just use fetch keepalive

modules/pubxaiAnalyticsAdapter.js
pnhegde: We have updated this to use prebid's storage methods.

modules/pubxaiAnalyticsAdapter.js
pnhegde: `keepalive` is not supported in all the browsers. I'm unsure if switching to keepalive is a good idea.

pnhegde:
@ncolletti @patmmccann I'm not sure what to do to avoid duplication error here. I don't think we can import code from other analytics adapters or if we can move them to a common library. 

Also, we have fixed the storage issue but sendBeacon is kept as it is instead of switching to fetch-keepalive due to a firefox compatibility issue. Let me know if this works or anything else we need to do. 

Thanks!

modules/pubxaiAnalyticsAdapter.js
patmmccann: if you want to use sendBeacon, you'll need to commit that to core communication methods

It isn't in fox but well be imminently is my understanding.

patmmccann:
Please move into a common library, eg this appears to be a direct copy: 

"format": "javascript",
    "lines": 18,
    "fragment": "function getOS() {\n  if (navigator.userAgent.indexOf('Android') != -1) return 'Android';\n  if (navigator.userAgent.indexOf('like Mac') != -1) return 'iOS';\n  if (navigator.userAgent.indexOf('Win') != -1) return 'Windows';\n  if (navigator.userAgent.indexOf('Mac') != -1) return 'Macintosh';\n  if (navigator.userAgent.indexOf('Linux') != -1) return 'Linux';\n  if (navigator.appVersion.indexOf('X11') != -1) return 'Unix';\n  return 'Others';\n}\n\n/**\n * Returns floor from priceFloors module or MediaKey default value.\n *\n * @param {*} bid a Prebid.js bid (request) object\n * @param {string} mediaType the mediaType or the wildcard '*'\n * @param {string|Array} size the size array or the wildcard '*'\n * @returns {number|boolean}\n */",

modules/pubxaiAnalyticsAdapter.js
patmmccann: some activity here https://bugzilla.mozilla.org/show_bug.cgi?id=1342484 just two days ago

modules/pubxaiAnalyticsAdapter.js
```diff
@@ -235,65 +260,50 @@ export const getDeviceType = () => {
...
-    return 'Macintosh';
-  }
-  if (navigator.userAgent.indexOf('Linux') != -1) {
-    return 'Linux';
-  }
-  if (navigator.appVersion.indexOf('X11') != -1) {
-    return 'Unix';
-  }
-  return 'Others';
+  if (navigator.userAgent.indexOf('Android') != -1) return osTypes.ANDROID
```
patmmccann, line 301: this appears to be an attempt to fool the duplication check?

pnhegde: @patmmccann Almost all the analytics adapters have their own implementation of browser, platform calculation. We decide to use enum not to fool the duplicates check but to make this enum specific to us as well as to reduce the payload size.

pnhegde: Implementing import is beyond the scope of this PR, to be honest. If someone else changes the enum numbers or the values of the platform/browser etc, it will affect our data collection. So, we are keeping the platform/OS enums specific to our Analytics adapter only and don't want to introduce a new dependency on a library import atm.

patmmccann: This isn't negotiable, members to your shower are now frozen

patmmccann: Since we're having trouble getting this pr merged, please submit an independent pr for your storage fix or your adapter risks unplanned changes to enforce the rules

patmmccann: Fyi, this may merge https://github.com/prebid/Prebid.js/pull/11830 if you don't work towards getting this pr merged or so an independent pr on just this fix

modules/pubxaiAnalyticsAdapter.js
```diff
@@ -1,208 +1,440 @@
...
+      hostname:
+        (auctionData.initOptions && auctionData.initOptions.hostName) ||
+        defaultHost,
+      pathname: path,
       search: {
-        auctionTimestamp: auctionTimestamp,
+        auctionTimestamp: auctionData.auctionDetail.timestamp,
         pubxaiAnalyticsVersion: pubxaiAnalyticsVersion,
-        prebidVersion: getGlobal().version
+        prebidVersion: getGlobal().version,
```
patmmccann, line 379: You can get the version with '$prebid.version$'

modules/pubxaiAnalyticsAdapter.js
patmmccann: https://github.com/prebid/Prebid.js/blob/2efc834cc291b3c2ff9abc70e95d5b3d3eb01b63/modules/genericAnalyticsAdapter.js#L145

It's the simplest way to implement

patmmccann (approved):

PR #11434, opened 2024-05-02
beOp Bid Adapter : add DNT, SH and SW in requests

sebrobert:
## Type of change

- [x] Feature

## Description of change
Just need to get new parameters from our Prebid.js delivery as we have added them in our own tag delivery

### DIFF
--- a/modules/beopBidAdapter.js
+++ b/modules/beopBidAdapter.js
@@ -61,6 +61,14 @@ export const spec = {
     const gdpr = bidderRequest.gdprConsent;
     const firstSlot = slots[0];
     const kwdsFromRequest = firstSlot.kwds;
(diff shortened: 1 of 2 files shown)


ChrisHuie:
@sebrobert it says `navigator.doNotTrack` is deprecated in the docs -> https://developer.mozilla.org/en-US/docs/Web/API/Navigator/doNotTrack

sebrobert:
Hello, sorry for disturbing @ChrisHuie → will remove that part into the PR. I was managing to have the same payload from prebid than from our historical embedded script on pages. We were aware that this param is deprecated, but you are right, it is better for us to remove it from the old code than adding it into the newest one.

PR #11442, opened 2024-05-05
KueezRtb Bid Adapter : add `mode` param and add external script disclosure

elad-yosifon:
## Type of change

- [ ] Bugfix
- [x] Feature
- [ ] New bidder adapter   
- [ ] Build related changes
- [ ] CI related changes

- [ ] Other

## Description of change

Introduce built-in support for publishers who use KueezRTB in "Latest" mode

### DIFF
--- a/modules/kueezRtbBidAdapter.js
+++ b/modules/kueezRtbBidAdapter.js
@@ -3,7 +3,7 @@ import {registerBidder} from '../src/adapters/bidderFactory.js';
 import {BANNER, VIDEO} from '../src/mediaTypes.js';
 import {getStorageManager} from '../src/storageManager.js';
 import {config} from '../src/config.js';
-import {loadExternalScript} from "../src/adloader";
+import {loadExternalScript} from '../src/adloader.js';
 
 const GVLID = 1165;
 const DEFAULT_SUB_DOMAIN = 'exchange';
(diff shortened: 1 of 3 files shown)


modules/kueezRtbBidAdapter.js
```diff
@@ -176,6 +177,12 @@ function buildRequests(validBidRequests, bidderRequest) {
   const bidderTimeout = bidderRequest.timeout ?? config.getConfig('bidderTimeout');
   const requests = [];
   validBidRequests.forEach(validBidRequest => {
+    try {
+      if (validBidRequest.params && (validBidRequest.params.mode === 'LATEST' || validBidRequest.params.mode !== 'BASIC')) {
+        loadExternalScript('https://static.kueezrtb.com/latest.js', BIDDER_CODE, () => {});
```
patmmccann, line 182: sorry, this is not allowed

elad-yosifon: @patmmccann what is the issue?
According to the guidelines, if there is a disclosure and it can be opted-out, it should be ok.

elad-yosifon:
@patmmccann @ChrisHuie I would rally appreciate an explanation or guidance on how to allow this.
This is a real business need, which affects our publishers, and helps with our prebid integrations.

PR #11445, opened 2024-05-06
Browsi RTD : add split key

omerDotan:
## Type of change
- [ ] Feature

## Description of change
Add global key value to GPT if configured by the user


osazos (approved):
LGTM

PR #11456, opened 2024-05-07
EXADS bid adapter: replace broken logic with merely bad logic

dgirardi:
## Type of change

- [x] Bugfix

## Other information

The user agent check in EXADS throws an exception if it reaches the 4th switch statement, which happens when running tests locally.


patmmccann (approved):

giuseppe-exads:
@tfoliveira

PR #11473, opened 2024-05-09
PulsePoint bid adapter: Removing deprecated method

anand-venkatraman:
Removing the `transformBidParams` method that has been deprecated for 9.0

### DIFF
--- a/modules/pulsepointBidAdapter.js
+++ b/modules/pulsepointBidAdapter.js
@@ -1,7 +1,6 @@
 import { ortbConverter } from '../libraries/ortbConverter/converter.js';
 import {isArray} from '../src/utils.js';
 import {registerBidder} from '../src/adapters/bidderFactory.js';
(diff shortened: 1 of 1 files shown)


modules/pulsepointBidAdapter.js
```diff
@@ -58,13 +58,6 @@ export const spec = {
         url: 'https://bh.contextweb.com/visitormatch/prebid'
       }];
     }
-  },
-  transformBidParams: function(params) {
-    return convertTypes({
```
patmmccann, line 63: you need to stop importing this on line 4 above to pass the linting check

patmmccann (changes requested):
not linting

modules/pulsepointBidAdapter.js
anand-venkatraman: addressed

PR #11479, opened 2024-05-13
ZetaGlobalSsp Analytics Adapter: refactoring

asurovenko-zeta:
## Description of change
Change analytics sending model and adapter refactoring


asurovenko-zeta:
As far as I understand, the build failed not because of my changes

Issue #11274, opened 2024-03-29
Feature request: ability to store user IDs in cookie and localStorage simultaneously


macinjosh32:
## Type of issue

Feature request

## Description

Currently, when configuring a user ID sub-module, the publisher must choose between a storage method of `cookie` or `html5` (localStorage). Both of these methods individually have advantages and limitations. `cookie` size, for example, is limited to 4K but can be persisted in the root domain, whereas `html5` size is virtually unlimited (10MB) but storage is restricted is the current sub-domain. Developers working for publishers are not necessarily aware of these tradeoffs, and even if they are, it's not necessarily obvious which option will yield more optimal ID persistence.

To address this issue, I'd to introduce a third storage method that saves IDs to both cookie and localStorage. ID retrieval will re-synchronize the cookie value to localStorage if present in only the cookie (or vice-versa if present only in localStorage) before returning the value. We can name this storage method `both`, though we are open to alternatives. The net result -- assuming the ID module uses a `domainOverride` like `domainOverrideToRootDomain` -- is that lightweight IDs become global to the root domain, and the heaviest IDs are still stored but restricted to the current subdomain.

There is already some precedent for this technique among modules that manage their own storage. For example, LiveIntent uses this in LiveConnect's caching module.

jdcauley:
Hello, 

The Identity committee discussed this in the last week.

In general we are onboard for supporting this capability.

However, this is not something that will be targeted for work internally via the Prebid.js committee

If your or someone else would like to contribute this capability we would like to review a proposed implementation to ensure we have no additional concerns, and would welcome a contribution.

Thank you,

Jordan

carlosfelix:
I've started the implementation of this feature.

PR #11482, opened 2024-05-13
User id module: Ability to store user IDs in cookie and localStorage simultaneously

carlosfelix:
## Type of change
- [ ] Bugfix
- [X] Feature
- [ ] New bidder adapter 
- [ ] Updated bidder adapter 
- [ ] Build related changes
- [ ] CI related changes
- [ ] Other

## Description of change
Introduce a third storage method that saves IDs to both cookie and localStorage. Details can be found in the issue link below.

## Other information
Issue: https://github.com/prebid/Prebid.js/issues/11274
Documentation: https://github.com/prebid/prebid.github.io/pull/5315

### DIFF
--- a/modules/userId/index.js
+++ b/modules/userId/index.js
@@ -239,6 +239,29 @@ function cookieSetter(submodule, storageMgr) {
   }
 }
 
+function setValueInCookie(submodule, valueStr, expiresStr) {
+  const storage = submodule.config.storage;
+  const setCookie = cookieSetter(submodule);
+
+  setCookie(null, valueStr, expiresStr);
+  setCookie('_cst', getConsentHash(), expiresStr);
+  if (typeof storage.refreshInSeconds === 'number') {
+    setCookie('_last', new Date().toUTCString(), expiresStr);
+  }
+}
+
+function setValueInLocalStorage(submodule, valueStr, expiresStr) {
+  const storage = submodule.config.storage;
+  const mgr = submodule.storageMgr;
+
+  mgr.setDataInLocalStorage(`${storage.name}_exp`, expiresStr);
+  mgr.setDataInLocalStorage(`${storage.name}_cst`, getConsentHash());
+  mgr.setDataInLocalStorage(storage.name, encodeURIComponent(valueStr));
+  if (typeof storage.refreshInSeconds === 'number') {
+    mgr.setDataInLocalStorage(`${storage.name}_last`, new Date().toUTCString());
+  }
+}
+
 /**
  * @param {SubmoduleContainer} submodule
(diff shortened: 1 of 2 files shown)


modules/userId/index.js
```diff
@@ -305,30 +336,46 @@ function setPrebidServerEidPermissions(initializedSubmodules) {
...
-    } else if (storage.type === LOCAL_STORAGE) {
-      const storedValueExp = mgr.getDataFromLocalStorage(`${storage.name}_exp`);
-      // empty string means no expiration set
-      if (storedValueExp === '') {
-        storedValue = mgr.getDataFromLocalStorage(storedKey);
-      } else if (storedValueExp) {
-        if ((new Date(storedValueExp)).getTime() - Date.now() > 0) {
-          storedValue = decodeURIComponent(mgr.getDataFromLocalStorage(storedKey));
-        }
+    submodule.enabledStorageTypes.find(storageType => {
```
ncolletti, line 366: Any reason for the `find` instead of a loop? 

Should it be documented somewhere that the last entry of the `types` will always be set to the storedValue reported from the User Id Module in EIDs?  ie) `cookie&html5` will use the html5 value and not the cookie since it is last to be checked

modules/userId/index.js
```diff
@@ -884,6 +935,16 @@ function updateInitializedSubmodules(dest, submodule) {
   }
 }
 
+function getConfiguredStorageTypes(config) {
+  return config?.storage?.type?.trim().split(/\s*&\s*/) || [];
```
ncolletti, line 939: Any reason the storage.types isn't an array instead of string parsing? 

ie) ['cookie', 'html5'] instead of "cookie&html5"

modules/userId/index.js
carlosfelix: well, `find()` will stop the iterations as soon as the provided function returns true.. for a `cookie&html5` value, if the ID is present in the cookie, then a true value will be returned and `find()` won't check `html5`, it's going to stop at the 1st iteration.

modules/userId/index.js
carlosfelix: it could be an array, just thought about keeping the same `string` type in the documentation, an offer a value that's also a string. Since I am not seeing another option that supports more than one type.

ncolletti: Makes sense, thank you for explaining

modules/userId/index.js
ncolletti: Gotcha, OK that makes sense. I previously read the switch statement incorrectly. This makes sense for getting the id

ncolletti (approved):
lgtm

PR #11486, opened 2024-05-14
Criteo Bid Adapter: converted adapter to oRTB 2.5

Ticki84:
## Type of change

- [X] Updated bidder adapter

## Description of change

- Modified request and response models to comply with OpenRTB 2.5 (using the oRTB converter).
- Changed the endpoint to forward requests to a new oRTB 2.5 compliant endpoint.
- Dropped support of undefined sized banner trick in favor of built-in `native` media type.
- Dropped support of native callback in favor of built-in rendering fn.

## Breaking changes and how to migrate (for publishers)

- If you were using undefined sized banner to get native bids:
  - Remove the undefined sized banner from `mediaTypes`.
  - Define a `native` object in `mediaTypes`.

- If you were previously using a native callback:
  - Define the render fn by following [Prebid Documentation](https://docs.prebid.org/prebid/native-implementation.html#433-define-the-render-javascript).
  - Define a `native` object in `mediaTypes`.

## Additional notes
Even though native requests without asset requirements should be supported, we strongly advised to define some native asset requirements.

### DIFF
--- a/modules/criteoBidAdapter.js
+++ b/modules/criteoBidAdapter.js
@@ -1,14 +1,14 @@
-import { deepAccess, generateUUID, isArray, logError, logWarn, parseUrl } from '../src/utils.js';
-import { registerBidder } from '../src/adapters/bidderFactory.js';
-import { config } from '../src/config.js';
-import { BANNER, NATIVE, VIDEO } from '../src/mediaTypes.js';
-
-import { getStorageManager } from '../src/storageManager.js';
-import { getRefererInfo } from '../src/refererDetection.js';
-import { hasPurpose1Consent } from '../src/utils/gpdr.js';
-import { Renderer } from '../src/Renderer.js';
-import { OUTSTREAM } from '../src/video.js';
-import { ajax } from '../src/ajax.js';
+import {deepAccess, deepSetValue, isArray, logError, logWarn, parseUrl} from '../src/utils.js';
+import {registerBidder} from '../src/adapters/bidderFactory.js';
+import {BANNER, NATIVE, VIDEO} from '../src/mediaTypes.js';
+import {getStorageManager} from '../src/storageManager.js';
+import {getRefererInfo} from '../src/refererDetection.js';
+import {hasPurpose1Consent} from '../src/utils/gpdr.js';
+import {Renderer} from '../src/Renderer.js';
+import {OUTSTREAM} from '../src/video.js';
+import {ajax} from '../src/ajax.js';
+import {ortbConverter} from '../libraries/ortbConverter/converter.js';
+import {ortb25Translator} from '../libraries/ortb2.5Translator/translator.js';
 
 /**
  * @typedef {import('../src/adapters/bidderFactory.js').BidRequest} BidRequest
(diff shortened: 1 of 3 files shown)


modules/criteoBidAdapter.js
```diff
@@ -33,6 +32,233 @@ const BUNDLE_COOKIE_NAME = 'cto_bundle';
...
+    // site
+    if (typeof context.url !== 'undefined') {
+      deepSetValue(request, 'site.page', context.url);
+    }
+    if (typeof context.publisherId !== 'undefined') {
+      deepSetValue(request, 'site.publisher.id', context.publisherId);
+    }
+  }
+
+  if (bidderRequest.eids) {
```
dgirardi, line 146: this should probably be `request`

modules/criteoBidAdapter.js
```diff
@@ -33,6 +32,233 @@ const BUNDLE_COOKIE_NAME = 'cto_bundle';
...
+    deepSetValue(request, 'user.ext.eids', [...bidderRequest.eids]);
+  }
+
+  if (typeof deepAccess(request, 'device.sua') !== 'undefined') {
```
dgirardi, line 150: there's a utility that does most of these: https://github.com/prebid/Prebid.js/blob/master/libraries/ortb2.5Translator/translator.js

I believe the differences are:
 - `ggp_sid` is not translated (easy enough to add to [this list](https://github.com/prebid/Prebid.js/blob/0612325a0db1991a62aa92304959ee0cb6e3092f/libraries/ortb2.5Translator/translator.js#L3-L10))
 - in addition to the `ext` shuffling, the utility also transforms the 2.6 `kwarray` fields into `keywords`.

modules/criteoBidAdapter.js
```diff
@@ -33,6 +32,233 @@ const BUNDLE_COOKIE_NAME = 'cto_bundle';
...
+      deepSetValue(request, 'regs.ext.gdprversion', bidderRequest.gdprConsent.apiVersion);
+      if (typeof bidderRequest.gdprConsent.gdprApplies !== 'undefined') {
+        deepSetValue(request, 'regs.ext.gdpr', !!(bidderRequest.gdprConsent.gdprApplies));
+      }
+      if (typeof bidderRequest.gdprConsent.consentString !== 'undefined') {
+        deepSetValue(request, 'user.ext.consent', bidderRequest.gdprConsent.consentString);
+      }
+    }
+
+    if (typeof bidderRequest.tid !== 'undefined') {
```
dgirardi, line 182: I am not aware of a `tid` field in bidder requests - the closest analog is `auctionId`

modules/criteoBidAdapter.js
Ticki84: `eids` are assigned to `bidderRequest` from `bidRequest.userIdAsEids` [here](https://github.com/prebid/Prebid.js/pull/11486/files/f5098b456ff0cd6bf94951d08fb91c61ae4a06db#diff-74bafa18f954c55e3036d5a4f29828357e9f3a536c0a62e7f51b1dedc4971648R401), as it was done before.
Since `bidRequest.userIdAsEids` is not the common eids field, it won't get converted so I'm not sure why it should be `request`?

modules/criteoBidAdapter.js
Ticki84: Thanks a lot, the translator looks super useful, I'll give it a try!

modules/criteoBidAdapter.js
Ticki84: Same as eids, it gets added to bidderRequest [here](https://github.com/prebid/Prebid.js/pull/11486/files/f5098b456ff0cd6bf94951d08fb91c61ae4a06db#diff-74bafa18f954c55e3036d5a4f29828357e9f3a536c0a62e7f51b1dedc4971648R402)

Ticki84:
Pushed a change to use the 2.5 translator instead of moving each field "manually".
Huge thanks for the review @dgirardi !

Ticki84:
Pushed another change because `gdprversion` is still an extension field in 2.6.

modules/criteoBidAdapter.js
```diff
@@ -19,20 +20,241 @@ import { ajax } from '../src/ajax.js';
...
-const CDB_ENDPOINT = 'https://bidder.criteo.com/cdb';
+const CDB_ENDPOINT = 'https://bidder.criteo.com/openrtb_2_5/pbjs/auction/request';
 const PROFILE_ID_INLINE = 207;
-export const PROFILE_ID_PUBLISHERTAG = 185;
 export const storage = getStorageManager({ bidderCode: BIDDER_CODE });
 const LOG_PREFIX = 'Criteo: ';
+const TRANSLATOR = ortb25Translator(true, Object.freeze([
+  ...DEFAULT_RULES,
+  moveRule('regs.gpp_sid'),
+  moveRule('device.sua')
```
dgirardi, line 32: I think it makes sense to add these directly into the library's `EXT_PROMOTIONS` (they're missing because it was written when these fields didn't exist).

modules/criteoBidAdapter.js
dgirardi: You're right - I missed it.

modules/criteoBidAdapter.js
```diff
@@ -19,20 +20,241 @@ import { ajax } from '../src/ajax.js';
...
+    // site
+    if (typeof context.url !== 'undefined') {
+      deepSetValue(request, 'site.page', context.url);
+    }
+    if (typeof context.publisherId !== 'undefined') {
+      deepSetValue(request, 'site.publisher.id', context.publisherId);
+    }
+  }
+
+  if (bidderRequest) {
```
dgirardi, line 152: Much of this logic is already included in FPD (`ortb2` / `ortb2Imp`) and/or ortbConverter (which automatically copies everything from FPD). There's a difference in how you handle transaction IDs (more on that below). If you want to compare:

- coppa - https://github.com/prebid/Prebid.js/blob/542694585269e3796d65205283783915a8cf9f70/src/fpd/enrichment.js#L138-L140
- us_privacy - https://github.com/prebid/Prebid.js/blob/542694585269e3796d65205283783915a8cf9f70/modules/consentManagementUsp.js#L249-L255
- gdpr (but not `gdprversion`, only `regs.ext.gdpr` and `user.ext.consent`) - https://github.com/prebid/Prebid.js/blob/542694585269e3796d65205283783915a8cf9f70/modules/consentManagement.js#L307-L314 
- gpp  - https://github.com/prebid/Prebid.js/blob/542694585269e3796d65205283783915a8cf9f70/modules/consentManagementGpp.js#L510-L521
- eids - https://github.com/prebid/Prebid.js/blob/542694585269e3796d65205283783915a8cf9f70/modules/userId/index.js#L1101-L1106
- schain - https://github.com/prebid/Prebid.js/blob/542694585269e3796d65205283783915a8cf9f70/modules/schain.js#L198-L205

And on the response side, dsa: https://github.com/prebid/Prebid.js/blob/542694585269e3796d65205283783915a8cf9f70/libraries/ortbConverter/processors/default.js#L100-L102


From transaction IDs, `source.tid` is an auction ID and transaction (impression) IDs are in `imp[].ext.tid` - see [this summary](https://docs.prebid.org/dev-docs/pb8-notes.html#transaction-identifiers-are-now-reliable-and-opt-in)

If you decide to use the built-in logic you'll probably have to update unit tests - there's [some helper utilities](https://github.com/search?q=repo%3Aprebid%2FPrebid.js%20addFPDToBidderRequest&type=code) to run the FPD enrichment logic (and the test suite should import all the relevant modules).

modules/criteoBidAdapter.js
dgirardi: I thought you were intending to translate `request.user.eids` (which is filled in by ortbConverter) into `request.user.ext.eids` - I missed the intermediate step where you set `bidderRequest.eids`. This makes sense if you don't want to use the eids filled in by ortbConverter - if you do, I added a summary of what I think you're duplicating.

modules/criteoBidAdapter.js
Ticki84: Sure, done.

modules/criteoBidAdapter.js
Ticki84: Thank you for the very detailed comparisons, much appreciated!

I pushed a change to use built-in modules as it looks more convenient.
For transaction ids, it should be fixed: we'll forward both `imp[].ext.tid` from `ortb2Imp` and `source.tid` from `ortb2` and decide how to properly deal with them on our backend.

As some modules aren't enabled by default (unlike fpd), there's nothing special to do on the adapter to enable those?

Ticki84:
CI seems a bit flakky @dgirardi , let me know if I need to push another empty commit to retrigger the CI.

modules/criteoBidAdapter.js
dgirardi: > As some modules aren't enabled by default (unlike fpd), there's nothing special to do on the adapter to enable those?

No - if the publisher includes them you get the logic (and if they don't include them, but they still want to "manually" provide info such as for example gpp strings, they can only do it through FPD which is still included).

dgirardi (approved):

PR #11493, opened 2024-05-15
1plusX Rtd Provider : add First Party Cookie ID option

luca1x:
## Type of change

- [ ] Bugfix
- [x ] Feature
- [ ] New bidder adapter  
- [ ] Updated bidder adapter   
- [ ] Build related changes
- [ ] CI related changes

- [ ] Other

## Description of change

- This module used to get our first party identifier via local storage. Adding an option so that it can also read it out from the first party cookie, set by us via JavaScript. By default, we still use LocalStorage.
- Extended unit tests to cover the parsing of the config described above
- Using the prebid storeManager to obtain access to local storage and cookie storage. This seems to be a best practice and makes sense imo to be a small refactoring as part of this PR. LMK if I'm not making proper use of this abstraction in my code.

### DIFF
--- a/modules/1plusXRtdProvider.js
+++ b/modules/1plusXRtdProvider.js
@@ -1,5 +1,7 @@
 import { submodule } from '../src/hook.js';
+import { MODULE_TYPE_RTD } from '../src/activities/modules.js';
 import { ajax } from '../src/ajax.js';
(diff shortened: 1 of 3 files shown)


luca1x:
@ChrisHuie ready for a first review, thanks! :)

osazos (approved):
LGTM

PR #11508, opened 2024-05-16
Adnuntius Bid Adapter : merge targeting from multiple sources

antosarho:
## Type of change

- [x] Bugfix

## Description of change

Allows targeting from ortb and directly on the bid to merge rather than have one overwrite the other.

### DIFF
--- a/modules/adnuntiusBidAdapter.js
+++ b/modules/adnuntiusBidAdapter.js
@@ -130,31 +130,6 @@ const storageTool = (function () {
     return (meta && meta.usi) ? meta.usi : false
   }
 
-  const getSegmentsFromOrtb = function (ortb2) {
-    const userData = deepAccess(ortb2, 'user.data');
-    let segments = [];
-    if (userData) {
-      userData.forEach(userdat => {
-        if (userdat.segment) {
-          segments.push(...userdat.segment.map((segment) => {
-            if (isStr(segment)) return segment;
-            if (isStr(segment.id)) return segment.id;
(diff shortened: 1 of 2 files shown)


modules/adnuntiusBidAdapter.js
```diff
@@ -171,25 +146,69 @@ const storageTool = (function () {
...
+    return deepAccess(bidderRequest.ortb2 || {}, 'site.ext.data');
+  };
+
+  return {
+    addSegmentsToUrlData: function (validBids, bidderRequest, existingUrlRelatedData) {
+      let segments = getSegmentsFromOrtb(bidderRequest || {});
+
+      for (let i = 0; i < validBids.length; i++) {
+        const bid = validBids[i];
+        const targeting = bid.params.targeting || {};
```
JulieLorin, line 193: there seems to be a new way to send you segments through the bid.params.targeting but it is not documented in Prebid docs, is there a reason for that ?

antosarho: There is a general `targeting` option as described in the documentation: https://docs.prebid.org/dev-docs/bidders/adnuntius.html

That targeting option is quite flexible and can include many different things that our adserver supports, for instance: `kv`, `auml`, `keywords`, `segments` and, as the documentation demonstrates, `c`.

So it's not a new way to send through`segments` -- it's always been the case that segments could be passed through along with the other options our Adnuntius ad server supports.

JulieLorin (approved):
LGTM

PR #11514, opened 2024-05-18
Improve Digital Bid Adapter : remove parsing of addtlConsent

jbartek25:
## Type of change
- [x] Other

## Description of change
Improve Digital bid adapter: Remove parsing of addtlConsent, falling back to the default logic provided by ortbConverter.


jbartek25:
CircleCI test failed but `gulp test` succeeds when I run it locally. What should I do here?

ChrisHuie (approved):

PR #11520, opened 2024-05-20
optimera Rtd Module : hotfix for handling score file URIs

dgarbartopti:
## Type of change
- [x] Bugfix
- [x] Other

## Description of change
There were some client scores which weren't being handled correctly due to a quirk with CF/S3's URI scheme

## Other information
A minor problem for the most part, just needs handling.


dgarbartopti:
This PR/module shouldn't have anything to do with User ID so I'm not sure what I should do about the CircleCI test which have failed.

ChrisHuie (approved):

schanoptimera:
@ChrisHuie Thanks for the code review and merge.  When is the next Prebid.js release?

Issue #10747, opened 2023-11-20
Prebid 9: stop supporting top level `site`, `app`, and `device` configuration


dgirardi:
## Description

Currently one can `setConfig({device: X})` and/or `setConfig({ortb2: {device: Y}})`, and we have [some logic](https://github.com/prebid/Prebid.js/blob/b7744ecc5b537ba652be81904bf6273f8a7bb101/src/fpd/enrichment.js#L51-L60) that merges the two before sending them to adapters. 

Our documentation suggests using `orbt2` with a caveat that it has only been supported since version 4.29. At some point we should remove support for the top-level options altogether.

PR #11522, opened 2024-05-20
Prebid 9: stop supporting top level app/site/device FPD config

dgirardi:
## Type of change

- [x] Other

## Description of change

Closes https://github.com/prebid/Prebid.js/issues/10747


test/spec/modules/prebidServerBidAdapter_spec.js
```diff
@@ -1543,28 +1555,6 @@ describe('S2S Adapter', function () {
       });
     });
 
-    it('site should not be present when app is present', function () {
```
osazos, line 1546: Why this test is removed?

dgirardi: Not sure! added back in.

osazos (approved):
LGTM

PR #11523, opened 2024-05-20
Adxcg Bid Adapter : remove transformBidParams function

adxcgcom:
## Type of change
- [x ] Refactoring (no functional changes, no api changes)

## Description of change
Per #6361 : remove usage of transformBidParams function
Confirmed that the server-side adapters for both PBS-Go handle correct type.
```

### CODE REVIEW

**Comment by adxcgcom** [comment]
CircleCI failed with unrelated error
```
1) unifiedid callback with partner
     User ID Handle variations of config values callbacks at the end of auction
```

ChrisHuie (approved):

PR #11526, opened 2024-05-21
Topics Module: add Smartadserver topics iframe

krzysztofequativ:
## Type of change
[x] Feature


ChrisHuie:
@krzysztofequativ closing this as we are moving this into config in Prebid 9 which is releasing in the next couple of weeks.

Issue #10964, opened 2024-01-18
ortbConverter: send max tmax


bretg:
## Type of issue
enhancement

## Description
The ortbConverter lib sets the ORTB tmax value to s2sConfig.timeout. This is great, but we have a module that would like to know the actual client-side timeout -- the `timeout` passed to pbjs.requestBids.

Specifically the use case is timeout optimization. We're exploring a server-side module that would help the publisher optimize the timeout value, essentially ignoring `tmax` and experimenting with other, possibly more effective values. But we need to know the ceiling for how far these explorations should go.

We propose passing the client-side timeout value as `ext.tmaxmax`.

patmmccann:
The actual timeout higher than the ceiling, why not just ask the publisher to set an actual ceiling they are comfortable with?

bretg:
Are you suggesting that pubs just set s2sConfig.timeout to be the same as what's sent in pbjs.requestBids() and use that as the ceiling?

I would solidly prefer not to have to (A) document that and (B) get existing pubs to change the page if/when they want to try this feature when it's easy and effective to get the client-side timeout.

dgirardi:
from [here](https://docs.prebid.org/dev-docs/modules/prebidServer.html):

> If the s2sConfig timeout is greater than the Prebid.js timeout, the s2sConfig timeout will be automatically adjusted to 75% of the Prebid.js timeout in order to fit within the auction process.

AFAIK this has never been the case since at least my time with Prebid. Do we want to also implement something like it - maybe a fixed 500ms discount? (percentage doesn't make a lot of sense to me, if I increase client-side timeout it's not like the latency to server also increases).

bretg:
> this has never been the case since at least my time with Prebid

Hmm. Good topic. I think 500ms would be an overly large buffer. The number depends a lot on the user's network proximity to the nearest Prebid Server. Discuss in a future meeting?

bretg:
Discussed. Agreed that a publisher should be able to define the tmaxmax like this:

```
pbjs.setConfig({
   ortb2: {
      ext: {
         tmaxmax: 888
      }
   }
});
```

The ortbConverter library should check for the existence of `ext.tmaxmax` and if not there, set it to the timeout value passed to `requestBids()`

bmjoju:
@bretg I can also add that the below does not seem to be the working

[...]

PR #11540, opened 2024-05-22
ortbConverter: Tmax - Update logic for how it is determined

jefftmahoney:
## Type of change
- [x] Feature

## Description of change
Update the logic in `ortbConverter.js` so that if a publisher specifies `tmaxmax` as a part of their setup (i.e. specifies it in their `setConfig()` call), `tmaxmax` is what is used to determine `tmax` for a request.

### Sample setup that a publisher would do to invoke this new logic
```
pbjs.setConfig({
   ortb2: {
      ext: {
         tmaxmax: 888
      }
   }
});
```

## Other information
The work of this PR addresses [this issue](https://github.com/prebid/Prebid.js/issues/10964)


modules/prebidServerBidAdapter/ortbConverter.js
```diff
@@ -57,7 +57,7 @@ const PBS_CONVERTER = ortbConverter({
       let {s2sBidRequest, requestedBidders, eidPermissions} = context;
       const request = buildRequest(imps, proxyBidderRequest, context);
 
-      request.tmax = s2sBidRequest.s2sConfig.timeout;
+      request.tmax = s2sBidRequest.s2sConfig.ortb2?.ext?.tmaxmax || s2sBidRequest.s2sConfig.timeout;
```
dgirardi, line 60: my reading of the [issue](https://github.com/prebid/Prebid.js/issues/10964):

> This is great, but we have a module that would like to know the actual client-side timeout -- the timeout passed to pbjs.requestBids.

> The ortbConverter library should check for the existence of ext.tmaxmax and if not there, set it to the timeout value passed to requestBids()

is that this should be a new field, `request.ext.tmaxmax`, which by default should be the client timeout, optionally overridable with `ortb2.ext.tmaxmax`

which I think in code translates to

```suggestion
      request.ext.tmaxmax = request.ext.tmaxmax ?? proxyBidderRequest.timeout
```

jefftmahoney: Hmmm ... yes, I think I agree with you @dgirardi.  

Belated question: do we think that `tmaxmax` is something that a SSP endpoint (the place a bid request will get POST-ed to) needs to worry about?  Or is this just to help the server-side make a smart decision about `tmax`?  I'm assuming since `tmaxmax` is not part of oRTB the answer is "no", but thought I'd ask.

jefftmahoney: @dgirardi I'm having trouble writing a test to verify that this works.  

Where would `ortbConverter.js` look to find out if a publisher had made this call:
```
pbjs.setConfig({
   ortb2: {
      ext: {
         tmaxmax: 888
      }
   }
});
```

Would it be in `proxyBidderRequest`?  That object is undefined right now in the test I'm running, but if that is where a publisher-specified `tmaxmax` value would go I could mock one up.

dgirardi: roughly, `ortb2` comes in as `s2sBidRequest.ortb2Fragments.global` here:

https://github.com/prebid/Prebid.js/blob/e07451212d8b708fe662031f8a4a1e159b469b49/modules/prebidServerBidAdapter/index.js#L454

passed down here:

https://github.com/prebid/Prebid.js/blob/e07451212d8b708fe662031f8a4a1e159b469b49/modules/prebidServerBidAdapter/ortbConverter.js#L262

and finally here:

https://github.com/prebid/Prebid.js/blob/e07451212d8b708fe662031f8a4a1e159b469b49/modules/prebidServerBidAdapter/ortbConverter.js#L172

which is how, in the code at the top of this thread, it ends up filling `request.ext.tmaxmax` after you run `buildRequest`.

"roughly" because there's more than one way to configure `ortb2` - and the PBS adapter is peculiar in that it also has to deal with bidder-specific configuration separately from global configuration (hence the `.global`)

in practice for testing that means manipulating `ortb2Fragments.global` in the first argument to `callBids` - here's an example:

https://github.com/prebid/Prebid.js/blob/e07451212d8b708fe662031f8a4a1e159b469b49/test/spec/modules/prebidServerBidAdapter_spec.js#L2614-L2621

jefftmahoney: This is great - thank you so much!  I'll try some of this out later this afternoon and let you know.

jefftmahoney: Addressed with commit `b3b3ca4`

modules/prebidServerBidAdapter/ortbConverter.js
```diff
@@ -58,6 +58,7 @@ const PBS_CONVERTER = ortbConverter({
       const request = buildRequest(imps, proxyBidderRequest, context);
 
       request.tmax = s2sBidRequest.s2sConfig.timeout;
+      request.ext.tmaxmax = request.ext.tmaxmax || proxyBidderRequest.timeout;
```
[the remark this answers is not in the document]
jefftmahoney: Cool.  I've made the suggested changes - or what I understand the suggested changes to be - with `4bb1b47`.  Happy to answer any questions or follow up as needed!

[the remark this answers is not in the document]
dgirardi: > I think this means that I need to do additional test-data setup to mimic requestBidsTimeout being available in a real-world scenario.

Only if you add a test case that covers the fallback - which I won't object to. Currently it should pass as it only checks that the input `tmaxmax` has precedence.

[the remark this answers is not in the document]
jefftmahoney: OK, fallback test added.

dgirardi (approved):

ChrisHuie (approved):

PR #10438, opened 2023-09-06
Core: fill in `video.plcmt` when possible

dgirardi:
## Type of change

- [x] Feature

## Description of change

Closes https://github.com/prebid/Prebid.js/issues/10406

PR #11554, opened 2024-05-23
Triplelift: Adding support for video.plcmt

patrickloughrey:
## Type of change
- [x] Feature

## Description of change
Adding support for `video.plcmt` if the publisher doesn't explicitly set it

### DIFF
--- a/modules/tncIdSystem.md
+++ b/modules/tncIdSystem.md
@@ -10,6 +10,8 @@ gulp build --modules=tncIdSystem,userId
 
 ### TNCIDIdSystem module Configuration
 
+Disclosure: This module loads external script unreviewed by the prebid.js community
+
 You can configure this submodule in your `userSync.userIds[]` configuration:
 
 ```javascript


modules/tripleliftBidAdapter.js
```diff
@@ -241,6 +241,9 @@ function _getORTBVideo(bidRequest) {
     if (!video.placement) {
       video.placement = 1;
     }
+    if (!video.plcmt) {
+      video.plcmt = 1;
```
patmmccann, line 245: @fowler446 can you comment here? Generally video.context instream does not map to plcmt = 1

modules/tripleliftBidAdapter.js
```diff
@@ -249,6 +252,9 @@ function _getORTBVideo(bidRequest) {
       logMessage(`video.placement value of ${video.placement} is invalid for outstream context. Setting placement to 3`)
       video.placement = 3
     }
+    if (!video.plcmt) {
+      video.plcmt = 4;
```
patmmccann, line 256: prebid already has logic in place doing something similar

patmmccann: to support plcmt, pick up what you find in the request, this logic to infer it is not necessary. That's true for the old value as well.

patrickloughrey: in current versions? or in the upcoming v9? Because currently if `video.plcmt` isn't set, it will not be added to the bid request

patmmccann: since https://github.com/prebid/Prebid.js/pull/10438

modules/tripleliftBidAdapter.js
patrickloughrey: image (8)

patrickloughrey: This is direct from IAB so I think we have instream `plcmt` correct. But like I mentioned below, if the publisher sets this, we don't meddle with it.

patmmccann: You're gonna get in trouble here from ttd. They were very adamant no one does this. Prebid instream context basically just means preroll, there is no easy way for you to infer if plcmt should be 1 or 2

patmmccann (changes requested):
Please fix incorrect plcmt interference

Issue #11533, opened 2024-05-21
Yahoo Adapter reading video.placement and not video.plcmt (Prebid 9)


DeepthiNeeladri:
**Type of issue**
Question

**Description**
Hi,

I have a question regarding reading video.placement and not video.plcmt. our adserver does not yet support video.plcmt yet, so we will still need to make sure we can read video.placement or do a mapping if the former is sent. Will this will have any problem once user updates to Prebid 9.
https://github.com/prebid/Prebid.js/issues/10452
https://github.com/prebid/Prebid.js/blob/master/modules/yahoosspBidAdapter.js#L373
uses `placement` and no `plcmt`
https://github.com/prebid/Prebid.js/blob/master/modules/yahoosspBidAdapter.md 
docs have `placement`. Could use an update

Thanks,
Deepthi

patmmccann:
Why not have your js adapter pass it up and ignore it server side until you've completed the work to transition?

PR #11569, opened 2024-05-26
Yahoo Bid Adapter: adding plcmt param support

DeepthiNeeladri:
## Type of change
- [x] Updated bidder adapter  <!--  IMPORTANT: (1) consider whether you need to upgrade your bidder parameter 

## Description of change
Update the yahooAdsBidAdapter to add support for video.plcmt
our adserver does not yet support video.plcmt yet so we are just passing the param as suggested by https://github.com/prebid/Prebid.js/issues/11533


patmmccann:
this is great, this means publishers won't need to upgrade to receive this functionality once your backend can process it.

ChrisHuie (approved):

Issue #11503, opened 2024-05-16
Not active bidder’s iframe called by FPD module


krzysztofequativ:
## Type of issue
Bug (?)

## Description
The Topics First Party Data (FPD) Module (topicsFpdModule) calls a topics bidder iframe not paying attention to the fact that the bidder hasn't been activated in the adUnits configuration. What's more, it's not possible to control whose iframe is called or to block some bidders. It might be a bug but maybe it's an expected behaviour?

patmmccann:
The committee decided to move ALL topics module urls into configuration by the publisher. Anyone is welcome to PR the doc with the location of their frame.

PR #11579, opened 2024-05-28
Prebid 9: Move bidders iframes urls to config

mkomorski:
## Type of change

- [ ] Bugfix
- [x] Feature
- [ ] New bidder adapter  
- [ ] Updated bidder adapter   
- [ ] Build related changes
- [ ] CI related changes

- [ ] Other

## Other information

#11503

### DIFF
--- a/src/config.js
+++ b/src/config.js
@@ -60,6 +60,35 @@ const GRANULARITY_OPTIONS = {
 
 const ALL_TOPICS = '*';
 
+const DEFAULT_IFRAMES_CONFIG = {
+  maxTopicCaller: 4,
+  bidders: [{
+    bidder: 'pubmatic',
+    iframeURL: 'https://ads.pubmatic.com/AdServer/js/topics/topics_frame.html'
+  }, {
+    bidder: 'rtbhouse',
+    iframeURL: 'https://topics.authorizedvault.com/topicsapi.html'
+  }, {
+    bidder: 'openx',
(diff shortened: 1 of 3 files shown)


src/config.js
```diff
@@ -59,6 +59,35 @@ const GRANULARITY_OPTIONS = {
 
 const ALL_TOPICS = '*';
 
+const DEFAULT_IFRAMES_CONFIG = {
```
patmmccann, line 62: please make this empty and then PR this as an example configuration on https://github.com/prebid/prebid.github.io/blob/master/dev-docs/modules/topicsFpdModule.md

src/config.js
```diff
@@ -160,6 +189,11 @@ export function newConfig() {
 
       // default max nested iframes for referer detection
       maxNestedIframes: DEFAULT_MAX_NESTED_IFRAMES,
+
+      // default bidders iframes urls
+      userSync: {
+        topics: DEFAULT_IFRAMES_CONFIG
```
patmmccann, line 195: actually i think we can just leave this out altogether; this would add configuration on an optional module regardless of it it were included. 

When I said move to config, I should have been more specific, I meant we would allow publishers to choose these with their configuration, not default to anyone.

patmmccann:
please open a PR with this suggested list on the docs repo

patmmccann (changes requested):
looks like the tests need to be fixed, looking good so far!

patmmccann (approved):

PR #11581, opened 2024-05-28
Appnexus Bid Adapter: parse the currency from the bid if specified

astartsky:
## Type of change
- [x] Bugfix
- [x] Updated bidder adapter

## Description of change
Parse the currency from the bid if specified. Currently, the bid currency is ignored and is always set to USD

### DIFF
--- a/modules/appnexusBidAdapter.js
+++ b/modules/appnexusBidAdapter.js
@@ -597,7 +597,7 @@ function newBid(serverBid, rtbBid, bidderRequest) {
     cpm: rtbBid.cpm,
     creativeId: rtbBid.creative_id,
     dealId: rtbBid.deal_id,
-    currency: rtbBid.publisher_currency_code || 'USD',
+    currency: rtbBid.publisher_currency_codename || 'USD',
     netRevenue: true,
     ttl: 300,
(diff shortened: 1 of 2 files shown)


modules/appnexusBidAdapter.js
```diff
@@ -597,7 +597,7 @@ function newBid(serverBid, rtbBid, bidderRequest) {
     cpm: rtbBid.cpm,
     creativeId: rtbBid.creative_id,
     dealId: rtbBid.deal_id,
-    currency: 'USD',
+    currency: rtbBid.publisher_currency_code || 'USD',
```
jsnellbaker, line 600: @astartsky 
If you want the name of the currency (like `USD`) then you should change the field here to `publisher_currency_codename`.

The code field will return the currency symbol (like `$`), which doesn't match the backup value and the unit test.

jsnellbaker (changes requested):
See field note below

modules/appnexusBidAdapter.js
astartsky: @jsnellbaker thanks for the clarification! I made a fix

jsnellbaker (approved):

PR #11583, opened 2024-05-28
appnexusBidAdapter 9.0 - remove transform bid params and create anPspParamsConverter module

jsnellbaker:
## Type of change

- [x] Feature

## Description of change
Related to #6361

This PR moves certain functionality that was previously stored in the appnexusBidAdapter.js file within a function named transformBidParams.

This transformBidParams was a standard function in all adapters, which helped to change/modify the params and their values to a format that matched the bidder's request structure on the server-side endpoint.  In Prebid.js 9.0, this standard function was removed in all adapter files, so that the whole client-side file (eg appnexusBidAdapter.js) wouldn't have to be included in a prebid.js build file that was meant for server-side bidders.

This module is meant to be a temporary measure, until a more long-term solution can be determined on the Appnexus/Xandr/Microsoft end.

TODO Will create prebid docs PR once the module approach (and name) is reviewed/approved.


patmmccann:
@jsnellbaker thanks very much! I am sure you noticed the linting issues?

jsnellbaker:
I just pushed a fix that should address it

modules/anPspParamsConverter.js
```diff
@@ -0,0 +1,142 @@
...
+  // pbjs.aliasBidder
+  if (!result) {
+    result = !!(s2sConfig?.bidders?.find(bidder => (adapterManager.resolveAlias(bidder) === 'appnexus')));
+  }
+
+  return result;
+}
+
+// need a separate check b/c we're checking a specific bidRequest to see if we modify it, not just that we have a server-side bidder somewhere in prebid.js
+function isThisOurBidderInDisguise(tarBidder, s2sConfig) {
```
dgirardi, line 81: This can be simplified - and I'm not sure it works in the s2sTesting scenario (where I believe a bidder can be configured for s2s, but sometimes routed to the client)

bids routed to PBS have `src: CONTANTS.S2S.SRC`. so you could do something like

```javascript
bidderRequests
  .flatMap(br => br.bids)
  .filter(bid => bid.src === CONTANTS.S2S.SRC && adapterManager.resolveAlias(bid.bidder) === 'appnexus')
  .forEach((bid) => {
      // transform bid.params
  })
```

this does not look at `s2sConfig.extPrebid`, but neither does the [original logic](https://github.com/prebid/Prebid.js/blob/78f93cd29ee0955c6ddf2cf20f5251ac8e8c66e1/libraries/pbsExtensions/processors/params.js#L9-L13). Adding that is tricky because a publisher could theoretically configure multiple s2s configs where the same alias points to appnexus only on some of them.

jsnellbaker: Hi @dgirardi 
I updated the logic here with your suggestions.  Please take a look when you have the chance.  Thanks.

dgirardi (approved):

Issue #9718, opened 2023-03-27
oolo Analytics Adapter: edit down config in backend call


patmmccann:
## Type of issue
bug? unclear if intended. 

## Description
oolo sends the entire getConfig object to its backend. This contains resolved ids

### Expected results
oolo just sends publsiher configuration

### Actual results

oolo sends ids and sends publsiher configuration

patmmccann:
@aleksatr fyi

PR #11588, opened 2024-05-28
Prebid 9.0: no harvest eids

patmmccann:
removes non-documented case of harvesting eids, related #9718

Long discussion of publisher compliance risk here https://github.com/prebid/Prebid.js/pull/11524/files#r1612257023


jlaso (approved):

jlaso:
I was about to push something similar https://github.com/prebid/Prebid.js/compare/master...AuDigent:Prebid.js:HADRN-929?expand=1

Issue #9724, opened 2023-03-28
Analytics Modules must have docs


ChrisHuie:
All analytics modules need to have completed documentation inside the docs repo -> https://github.com/prebid/prebid.github.io

Issue on docs repo side -> https://github.com/prebid/prebid.github.io/issues/4093

patmmccann:
@JonGoSonobi icymi

PR #11590, opened 2024-05-28
9.0: undocumented analytics adapters

patmmccann:
from https://github.com/prebid/prebid.github.io/issues/4093

fixes https://github.com/prebid/Prebid.js/issues/9724


mdusmanalvi:
Hi @patmmccann When is this commit scheduled to be merged?
If we update Tercept analytics adapter documentation before it, can we be excused from this commit?

patmmccann:
@ChrisHuie i added tercept back to the branch

ChrisHuie (approved):

Issue #10607, opened 2023-10-13
Utiq id module: follow file naming conventions


patmmccann:
## Description
[A recent Utiq PR](https://github.com/prebid/Prebid.js/pull/10587) was to improve conformity with naming conventions. This is breaking however and this issue is to track the change for prebid 9

muuki88:
Docs PR: https://github.com/prebid/prebid.github.io/pull/4920

PR #11593, opened 2024-05-28
Prebid 9: rename utiqSystem to utiqIdSystem

dgirardi:
## Type of change
- [x] Other

## Description of change

This reverts commit 2b3426d6cc135bc9f4c903b2270234fcfaf649f1.

Closes https://github.com/prebid/Prebid.js/issues/10607


patmmccann (approved):

modules/.submodules.json
```diff
@@ -43,8 +43,7 @@
       "tapadIdSystem",
       "teadsIdSystem",
       "tncIdSystem",
-      "utiqSystem",
-      "utiqMtpIdSystem",
```
patmmccann, line 47: I think you removed one extra here. utiqMtp appears to be something else

PR #11599, opened 2024-05-29
Openweb bid adapter: Make placementId parameter mandatory

zkosanovic:
## Type of change
- [x] Updated bidder adapter   

## Description of change

Making placementId parameter mandatory for OpenWeb bid adapter.

**Note: This is OK for us to do, this change won't break any of our existing integrations.**


patmmccann:
@zkosanovic please re-open on the 9.0 branch. You cannot do this on master.

PR #11607, opened 2024-05-29
ZetaGlobalSsp Analytics Adapter : provide device object

asurovenko-zeta:
Provide **device** object through analytics adapter (in **auctionEnd** event)


asurovenko-zeta:
Looks like the tests fails not because of my changes

ChrisHuie (approved):

PR #11617, opened 2024-05-30
AdHash Bid Adapter: brand safety changes

wyand-sp:
Changing the recent ads calculations to the latest supported version.
Removed the usage of body.innerText

## Type of change
- [x] Feature
- [x] Build related changes

## Description of change
Changing the recent ads calculations to the latest supported version.
Removed the usage of body.innerText as needed in Prebid JS 9.0


ChrisHuie:
@wyand-sp this test failure should be fixed in the latest version of master if you want to pull in recent commits

modules/adhashBidAdapter.js
```diff
@@ -120,7 +121,7 @@ function brandSafety(badWords, maxScore) {
       .replaceAll(/\s\s+/g, ' ')
       .toLowerCase()
       .trim();
-    const content = window.top.document.body.innerText.toLowerCase();
+    const content = window.top.document.body.outerText.toLowerCase();
```
patmmccann, line 124: does textContent not work? that's what we used in 9.0 branch

patmmccann: https://kellegous.com/j/2013/01/26/layout-performance/ outerText just as bad

wyand-sp: Understood. We've updated the code. Thanks for your comment.

jsnellbaker (approved):

Issue #10452, opened 2023-09-07
List of adapters reading video.placement and not video.plcmt


patmmccann:
## Type of issue
Feature Request

## Description
#5237 was adopted in Prebid 5, and recently video.placement was deprecated in favor of video.plcmt.

This is a shell issue to grep out which adapters are pulling video.placement but not either using the ortb convertor nor pulling plcmt.

patmmccann:
Tagging as a possible 9.0 consideration

itaynave:
Should remove 
https://github.com/prebid/Prebid.js/tree/master/libraries/ortbConverter/processors/video.js
as placement is not 1 anymore in all instream cases

fowler446:
@ChrisHuie @patmmccann 
Here's the list of all adapters relying on `placement` without `plcmt`:

```
https://github.com/prebid/Prebid.js/blob/master/modules/e_volutionBidAdapter.js#L129
uses `placement` and no `plcmt`

https://github.com/prebid/Prebid.js/blob/master/modules/beachfrontBidAdapter.js#L29
uses `placement` and no `plcmt`

https://github.com/prebid/Prebid.js/blob/master/modules/outbrainBidAdapter.js#L357
uses `placement` and no `plcmt`

https://github.com/prebid/Prebid.js/blob/master/modules/sovrnBidAdapter.js#L30
uses `placement` and no `plcmt`

https://github.com/prebid/Prebid.js/blob/master/modules/operaadsBidAdapter.js#L528
uses `placement` and no `plcmt`

https://github.com/prebid/Prebid.js/blob/master/modules/saambaaBidAdapter.js#L125
uses `placement` and no `plcmt`

https://github.com/prebid/Prebid.js/blob/master/modules/improvedigitalBidAdapter.js#L235
uses `placement` and no `plcmt`

https://github.com/prebid/Prebid.js/blob/master/modules/admanBidAdapter.js#L146
uses `placement` and no `plcmt`

https://github.com/prebid/Prebid.js/blob/master/modules/smartxBidAdapter.js#L147
uses `placement` and no `plcmt`

https://github.com/prebid/Prebid.js/blob/master/modules/spotxBidAdapter.js#L216
uses `placement` and no `plcmt`

https://github.com/prebid/Prebid.js/blob/master/modules/advangelistsBidAdapter.js#L121
uses `placement` and no `plcmt`

https://github.com/prebid/Prebid.js/blob/master/modules/openxBidAdapter.js#L148
[...]

PR #11647, opened 2024-06-02
Gamoshi Bid Adapter : fixes use of `placement` and not `plcmt`

mmoses:
https://github.com/prebid/Prebid.js/blob/master/modules/gamoshiBidAdapter.js#L160 uses `placement` and no `plcmt`

fix bug - [10452](https://github.com/prebid/Prebid.js/issues/10452)


mmoses:
Hi Guys how can i fix this ? , I don't see any errors running lint on my branch.

ChrisHuie (approved):

Issue #11643, opened 2024-05-31
Coveralls deprecated? Update coverage badge


mkendall07:
## Type of issue
Development tooling

## Description
It looks like we no longer update the coverage badge from coveralls (latest build circa 2022 https://coveralls.io/builds/51056221) whether by intention or not. So this is just a question/request to either remove the badge or update the build pipeline to publish an updated coverage #.

patmmccann:
disabled here bc of old node version https://github.com/prebid/prebidjs-releaser/commit/49eb948339e027f2e20790f295920ea4c88e179a#diff-b7c2591cd32b0e297b6f86839df77463d07988bd3ca9c742503981acc1a2e590R16

hopefully we can reenable with 9 on #10999

patmmccann:
the coveralls badge is back

PR #11654, opened 2024-06-03
Readme : remove coveralls badge

ChrisHuie:
#11643


patmmccann:
Why remove the badge instead of fixing?

ChrisHuie:
> Why remove the badge instead of fixing?

I don't believe it is in the release pipeline any more as something we update. We could add it back but it's been gone for over a year at this point.

patmmccann:
How hard is it to add it back?

PR #11663, opened 2024-06-03
Conversant Bid Adapter: handle paapi data in the response

johnwier:
## Type of change

- [ ] Bugfix
- [X ] Feature
- [ ] New bidder adapter  
- [ ] Updated bidder adapter   
- [ ] Build related changes
- [ ] CI related changes

- [ ] Other

## Description of change
Conversant bid adapter: Add support for handling paapi data in bid response

### DIFF
--- a/modules/conversantBidAdapter.js
+++ b/modules/conversantBidAdapter.js
@@ -95,7 +95,7 @@ const converter = ortbConverter({
   },
   response(buildResponse, bidResponses, ortbResponse, context) {
     const response = buildResponse(bidResponses, ortbResponse, context);
-    return response.bids;
+    return response;
   },
   overrides: {
     imp: {
@@ -176,7 +176,8 @@ export const spec = {
    * @return {Bid[]} An array of bids which were nested inside the server.
    */
   interpretResponse: function(serverResponse, bidRequest) {
-    return converter.fromORTB({request: bidRequest.data, response: serverResponse.body});
+    const ortbBids = converter.fromORTB({request: bidRequest.data, response: serverResponse.body});
+    return ortbBids;
   },
 
   /**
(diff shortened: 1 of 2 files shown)


patmmccann (approved):

modules/conversantBidAdapter.js
```diff
@@ -176,7 +176,18 @@ export const spec = {
    * @return {Bid[]} An array of bids which were nested inside the server.
    */
   interpretResponse: function(serverResponse, bidRequest) {
-    return converter.fromORTB({request: bidRequest.data, response: serverResponse.body});
+    const ortbBids = converter.fromORTB({request: bidRequest.data, response: serverResponse.body});
+    const ortbAuctionConfigs = deepAccess(serverResponse, 'body.ext.igi') || [];
```
patmmccann, line 180: is this not handled here now? 

https://github.com/prebid/Prebid.js/commit/0ee4efdfde777045bdf5e00539efcfb878ff22d6#diff-24ee01f9467dc2a1704da10ca938150df2b3386d6b5b732870760dbc7cd1c4d7R373

cc @dgirardi

patmmccann (changes requested):
trying to validate if ortbConverter handles out of the box now since you've used the proper formatting in your response...

modules/conversantBidAdapter.js
johnwier: looks like as of today it is.  I'll go back through my changes and check it out

dgirardi: yes - this is no longer necessary. note that if you decide to keep it, `paapiAuctionConfigs` was renamed to just `paapi` (since it can also contain `igs` now) .

patmmccann (approved):

PR #11666, opened 2024-06-03
ampliffyBidAdapter.js: remove linter exceptions

patmmccann:


patmmccann:
Merging without review since it passed the linter

PR #11683, opened 2024-06-04
smartxBidAdapter: Using plcmt instead of placement

schubert-sc:
## Type of change

- [x] Other

## Description of change

smartxBidAdapter: Using plcmt instead of placement


modules/smartxBidAdapter.js
```diff
@@ -120,10 +120,10 @@ export const spec = {
       const api = getBidIdParameter('api', bid.params) || [2];
       const protocols = getBidIdParameter('protocols', bid.params) || [2, 3, 5, 6];
       var contextcustom = deepAccess(bid, 'mediaTypes.video.context');
-      var placement = 1;
+      var plcmt = 1;
```
patmmccann, line 123: this is incorrect; you cannot infer plcmt or placement, you should take it from the ad unt.

patmmccann (changes requested):

Issue #11102, opened 2024-02-14
Prebid 9: Paapi 4 gam autoconfig defaults


patmmccann:
Continuation of disussion here: https://github.com/prebid/Prebid.js/pull/10930#issuecomment-1942754697

What should the autoconfig of paapi do? are there two levels of autoconfig (1) same as prebid 8 behavior, (2) combine settargeting and setconfigs

Seems they should both be possible, debate appears to be on if (2) is the default in prebid 9 or the default behavior will be call set targeting and call set auction config independently

patmmccann:
Re-opening, as changing the default is unrelated to #11242

PR #11695, opened 2024-06-04
Prebid 9: rename fledge to paapi

dgirardi:
## Type of change
- [x] Other

## Description of change

This renames most instances of `fledge` to `paapi`:

 - `bidderRequest.fledgeEnabled` is renamed to `bidderRequest.paapi.enabled`
 - bid adapters should return `paapi` instead of `fledgeAuctionConfig`
 - the `fledgeForGpt` module is renamed to `paapiForGpt`
 - the `fledgeForGpt` configuration is renamed to `paapi`
 - the special case used for `setBidderConfig({bidders: [...], config: {fledgeEnabled}})` has been removed (as it's redundant with `setConfig({paapi: {enabled: true, bidders: [...]}})`
 - the default value for `paapi.gpt.autoconfig` is now false.

## Other information

Closes https://github.com/prebid/Prebid.js/issues/10573
Closes https://github.com/prebid/Prebid.js/issues/11102


modules/paapiForGpt.md
```diff
@@ -12,11 +12,11 @@ Publishers wishing to enable FLEDGE support must do two things. First, they must
 This is accomplished by adding the `fledgeForGpt` module to the list of modules they are already using:
 
 ```
-gulp build --modules=fledgeForGpt,...
+gulp build --modules=paapiForGpt,...
```
patmmccann, line 15: missed a variety of fledge references in this file, eg line 12 and potentially a few references above

integrationExamples/gpt/fledge_example.html
```diff
@@ -3,7 +3,7 @@
   <!--
     FLEDGE (Protected Audience API) configuration with GPT and FLEDGE-supporting adapter
 
-    gulp serve --modules=fledgeForGpt,openxBidAdapter
+    gulp serve --modules=paapiForGpt,openxBidAdapter
```
patmmccann, line 6: in general shall we set auction config before targeting? We might want to prepare for the idea it will be set before the auction is complete.

patmmccann: perhaps also rename this file

patmmccann (approved):
only remaining comment is on an md file

PR #11703, opened 2024-06-05
AdvRed Analytics Adapter : initial release

danoykin:
- [x] New analytics adapter

Maintainer: support@adv.red

pbjs.enableAnalytics({
    provider: 'advRed',
    options: {
        publisherId: '123456'
    }
});


spotxslagle (approved):
It looks like you only send events if the auction completes. It looks like this is intended, but you will be missing out on a lot of events from auctions that don't complete. Assuming you're fine with that, the rest looks good to me.

patmmccann:
@danoykin are you okay with that? Can you link the docs pr?

danoykin:
@patmmccann @spotxslagle Yes, it is ok. We are going to count only completed auctions for the current version.

PR #11709, opened 2024-06-05
UID2: Remove obsolete optout-check code for EUID

ssundahlTTD:
## Type of change

## Description of change
This change reverts the optout-check parameter that was added in https://github.com/prebid/Prebid.js/pull/11075.  This is no longer an optional parameter and its functionality is now the default for EUID.  This PR leaves the handling of the new optout return status but removes references to the CSTG parameter.

## Other information
This change reverts the optout-check parameter that was added in https://github.com/prebid/Prebid.js/pull/11075

### DIFF
--- a/modules/euidIdSystem.js
+++ b/modules/euidIdSystem.js
@@ -103,7 +103,6 @@ export const euidIdSubmodule = {
       mappedConfig.cstg = {
         serverPublicKey: config?.params?.serverPublicKey,
         subscriptionId: config?.params?.subscriptionId,
-        optoutCheck: 1,
         ...extractIdentityFromParams(config?.params ?? {})
       }
     }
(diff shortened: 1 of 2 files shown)


modules/euidIdSystem.js
```diff
@@ -103,7 +103,6 @@ export const euidIdSubmodule = {
       mappedConfig.cstg = {
         serverPublicKey: config?.params?.serverPublicKey,
         subscriptionId: config?.params?.subscriptionId,
-        optoutCheck: 1,
```
patmmccann, line 106: could you fix these jsdoc warnings too please?

image

patmmccann (changes requested):
looks good in general, also we're asking submitters to clean all the jsdocs warnings as we go

lionell-pack-ttd:
LGTM

PR #11716, opened 2024-06-05
9.0: Update idImportLibrary.js

### LINKED ISSUE #11708: idImportLibrary.js is intercepting communications without an active submission action
## Type of issue

Bug

## Description

The idImportLibrary module is [designed to scrape form fields on `change` and `blur` events](https://github.com/prebid/Prebid.js/blob/105c6620a4882dd59245b1a7c264414efbdbe3b1/modules/idImportLibrary.js#L99), meaning they will pull data from those fields without an active user action to submit it. 

This runs afoul of some pretty uncertain legal ground around intercepting communications, wiretapping and keystroke monitoring. I'm not a lawyer, but as I understand the issue this can definitely get you a publisher using this technology in court--even if the publisher might win the case.

Notable iterations on this concern in courts include the [Fullstory case](https://casetext.com/case/hasson-v-fullstory-inc-1) and the underlying legal argument [against](https://www.ca1.uscourts.gov/sites/ca1/files/opnfiles/23-1683P-01A.pdf) [many](https://www.cov.com/en/news-and-insights/insights/2022/12/wiretap-litigation-risks-when-using-website-support-technologies#layout=card&numberOfResults=12) [session](https://natlawreview.com/node/243281/printable/pdf) [replay](https://www.cbsnews.com/news/casper-mattress-startup-sued-for-wiretapping-site-visitors/) [services](https://www.bytebacklaw.com/2023/02/privacy-litigation-update-illinois-class-action-lawsuit-joins-string-of-wiretapping-claims-arising-out-of-website-operator-use-of-session-replay-technology/). 

Some of these cases have been dismissed, but not on the argument that the activity was legal, but on the argument that the person bringing the case lacked jurisdiction, making this activity very legally unclear. The legal definitions at issue here are not fully resolved in case law as far as I can find, which means while this activity may not be wiretapping, it also seems like it isn't not wiretapping, which seems a pretty dubious place to be. While this is more of a US law concern than anywhere else it raises serious concerns about the legality of using this module in the US at least and the appropriateness of hosting this module within the prebid project.  

I would argue that this module should be removed on ethical grounds alone, as scraping user inputs without an active submission is functionally misleading the user to what is occurring in those fields but even if we do not feel we need to remove it from this repository, the documentation should be updated to make it clear what is occurring here so that publishers who choose to leverage the module can make more informed choices about the accompanying legal risk.

patmmccann:
fixes #11708

## Type of change

- [ ] Bugfix
- [ ] Feature
- [ ] New bidder adapter  
- [ ] Updated bidder adapter   
- [ ] Build related changes
- [ ] CI related changes

- [ ] Other

### DIFF
--- a/modules/idImportLibrary.js
+++ b/modules/idImportLibrary.js
@@ -91,8 +91,7 @@ function addInputElementsElementListner() {
 
   for (var i = 0; i < inputs.length; i++) {
     _logInfo(`Original Value in Input = ${inputs[i].value}`);
-    inputs[i].addEventListener('change', event => processInputChange(event));
-    inputs[i].addEventListener('blur', event => processInputChange(event));
+    inputs[i].addEventListener('submit', event => processInputChange(event));
   }
 }
 
@@ -105,8 +104,7 @@ function addFormInputElementsElementListner(id) {
   }
   _logInfo('Adding input element listeners');
   const input = document.getElementById(id);
-  input.addEventListener('change', event => processInputChange(event));
-  input.addEventListener('blur', event => processInputChange(event));
+  input.addEventListener('submit', event => processInputChange(event));
 }
 
 function removeInputElementsElementListner() {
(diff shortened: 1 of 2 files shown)


patmmccann:
i dont actually think this works... can you listen to all the elements for submit or just the entire form?

PR #11719, opened 2024-06-06
ORTB2: don't rely on context to infer video.placement

### LINKED ISSUE #10452: List of adapters reading video.placement and not video.plcmt

### LINKED ISSUE #11702: Fix videojsVideoProvider.js for plcmt
part of #10452 , prebid maintains https://github.com/prebid/Prebid.js/blob/master/modules/videojsVideoProvider.js

patmmccann:
fixes #11702 

fixes #10452 for videojs player file


libraries/video/constants/ortb.js
```diff
@@ -97,6 +98,17 @@ export const PLACEMENT = {
...
 
+/**
+ * ADCOM - https://github.com/InteractiveAdvertisingBureau/AdCOM/blob/develop/AdCOM%20v1.0%20FINAL.md#list_plcmtsubtypesvideo
+ * @enum OrtbVideoParams.plcmt
+ */
+export const PLCMT = {
+  INSTREAM: 1,
+  ACCOMPANYING_CONTENT: 2,
+  INTERSTITIAL: 3,
+  NO_CONTENT: 4
```
fowler446, line 109: nice

modules/videojsVideoProvider.js
```diff
@@ -146,10 +146,7 @@ export function VideojsProvider(providerConfig, vjs_, adState_, timeState_, call
     // ~ Sort of resolved check if the player has a source to tell if the placement is instream
     // Still cannot reliably check what type of placement the player is if its outstream
     // i.e. we can't tell if its interstitial, in article, etc.
-    if (player.src()) {
-      video.placement = PLACEMENT.INSTREAM;
```
karimMourra, line 150: can we set `video.plcmt` to 2 ? If `player.src()` evaluates to `truethy` then we know there is content.

patmmccann: resolved on line below

modules/ixBidAdapter.js
```diff
@@ -258,17 +258,13 @@ export function bidToVideoImp(bid) {
 
   // if placement not already defined, pick one based on `context`
   if (context && !imp.video.hasOwnProperty('placement')) {
-    if (context === INSTREAM) {
-      imp.video.placement = 1;
```
karimMourra, line 262: Technically this is accurate, the legacy `placement` doesn't distinguish the different types of "instream" presented in `plcmt`. I think what this bid adapter is missing is `plcmt` in addition to legacy placement

modules/videojsVideoProvider.js
```diff
@@ -146,8 +146,9 @@ export function VideojsProvider(providerConfig, vjs_, adState_, timeState_, call
     // ~ Sort of resolved check if the player has a source to tell if the placement is instream
     // Still cannot reliably check what type of placement the player is if its outstream
     // i.e. we can't tell if its interstitial, in article, etc.
+    // update: cannot infer instream ever, always need declarations
     if (player.src()) {
-      video.placement = PLACEMENT.INSTREAM;
```
karimMourra, line 150: is the idea to entirely remove the use of `placement` since it is deprecated ?  Or to include `plcmt` as well ?

modules/ixBidAdapter.js
patmmccann: they handle plcmt on line 257 above

however, removing so they can handle in their own pr

modules/videojsVideoProvider.js
patmmccann: I think the idea is we want to provide information we can inder and not cross a line into invalid request territory on either field guessing too much

patmmccann: @fowler446 had flagged this adapter in https://github.com/prebid/Prebid.js/issues/10452#issuecomment-2093546373; this pr is hoping to solve it for one of the last remaining adapters on the list

PR #11721, opened 2024-06-06
Adnuntius Bid Adapter: standardise kv targeting value handling when sent to adserver

antosarho:
## Type of change

- [x] Bugfix

## Description of change

Standardises how kv targeting data is sent to the adserver. 

Kv targeting specified as an object and kvs from rtb information are converted to an array to avoid clashing with kv targeting specified as an array and to allow separate objects with the same key to be sent to the adserver via the array.

### DIFF
--- a/modules/adnuntiusBidAdapter.js
+++ b/modules/adnuntiusBidAdapter.js
@@ -1,6 +1,6 @@
 import { registerBidder } from '../src/adapters/bidderFactory.js';
 import { BANNER, VIDEO } from '../src/mediaTypes.js';
-import { isStr, deepAccess } from '../src/utils.js';
+import {isStr, isEmpty, deepAccess, getUnixTimestampFromNow, convertObjectToArray} from '../src/utils.js';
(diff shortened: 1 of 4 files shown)


modules/adnuntiusBidAdapter.js
```diff
@@ -22,6 +22,15 @@ export const misc = {
   getUnixTimestamp: function (addDays, asMinutes) {
     const multiplication = addDays / (asMinutes ? 1440 : 1);
     return Date.now() + (addDays && addDays > 0 ? (1000 * 60 * 60 * 24 * multiplication) : 0);
+  },
+  convertObjectToArray: function(obj) {
```
patmmccann, line 26: This looks like it should be a utility in utils folder

antosarho: OK -- I've pushed both functions into the utils.js file and available now for anyone else.

patmmccann (approved):

PR #11722, opened 2024-06-06
Consumable Bid Adapter: add language to request

jpiros:
- [x] Updated bidder adapter   

## Description of change
This retrieves the user's browser language from the `Navigator` object and populates the request param `lang`.

### DIFF
--- a/modules/consumableBidAdapter.js
+++ b/modules/consumableBidAdapter.js
@@ -61,7 +61,8 @@ export const spec = {
       source: [{
         'name': 'prebidjs',
         'version': '$prebid.version$'
-      }]
+      }],
+      lang: (navigator && navigator.language) ? navigator.language.split('-')[0] : ''
     }, validBidRequests[0].params);
 
     if (bidderRequest && bidderRequest.gdprConsent) {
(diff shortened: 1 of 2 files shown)


modules/consumableBidAdapter.js
```diff
@@ -61,7 +61,8 @@ export const spec = {
       source: [{
         'name': 'prebidjs',
         'version': '$prebid.version$'
-      }]
+      }],
+      lang: (navigator && navigator.language) ? navigator.language.split('-')[0] : ''
```
patmmccann, line 65: this is already on your request object, no need to hit navigator
https://github.com/prebid/Prebid.js/blob/577e8a29e709af3e7fa4464227e7cdc612b24fc9/src/fpd/enrichment.js#L122

patmmccann (changes requested):
get it from the request

modules/consumableBidAdapter.js
jpiros: Updated to grab from the `bidderRequest` object

patmmccann (approved):

PR #11746, opened 2024-06-10
SeedingAlliance Bid Adapter: internal rework to send OpenRTB native compliant requests to backend

sag-henmus:
## Type of change

- [ ] Bugfix
- [ ] Feature
- [ ] New bidder adapter  
- [x] Updated bidder adapter   
- [ ] Build related changes
- [ ] CI related changes

- [ ] Other

## Description of change
Due to work on the backend, we need clients to send properly formatted OpenRTB Native requests, as well as understand the response. This change assures that. Compability on the backend is ensured.

### DIFF
--- a/modules/oguryBidAdapter.js
+++ b/modules/oguryBidAdapter.js
@@ -1,7 +1,7 @@
 'use strict';
 
 import {BANNER} from '../src/mediaTypes.js';
-import {getWindowSelf, getWindowTop, isFn, logWarn} from '../src/utils.js';
+import {getWindowSelf, getWindowTop, isFn, logWarn, deepAccess} from '../src/utils.js';
 import {registerBidder} from '../src/adapters/bidderFactory.js';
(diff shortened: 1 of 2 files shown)


patmmccann:
It seems you have several duplicated code blocks

[filtered-jscpd-report.json](https://github.com/user-attachments/files/15765023/filtered-jscpd-report.json)

sag-henmus:
The change is only in two files, nowhere near where the report says. I'm unsure what the problem seems to be. Any advice on how to handle this?

patmmccann:
If you pull in master it should make a pretty comment flagging the blocks now. 

The idea is to simply move the duplicated code to a common location and it import it in both places it is used

patmmccann:
If your goal is to make ortb2 bid requests, why still use `convertOrtbRequestToProprietaryNative(validBidRequests)` ?

PR #11750, opened 2024-06-10
Taboola bid adapter add ortb2 device

jwrosewell:
## Type of change
- [x] Feature
- [x] Updated bidder adapter

## Description of change
This PR enhances the Taboola bidder request by incorporating device data from the global ORTB2 object. Existing values in the Taboola device object will be overridden if they are also present in the global ORTB2 object. However, values unique to Taboola or not found in the global ORTB2 object will remain unchanged.

## Motivation
The device object has previously been populated by simplistic parsers, if at all, and was inaccurate as a result. Prebid now benefits from RTD modules such as [51Degrees](https://docs.prebid.org/dev-docs/modules/51DegreesRtdProvider.html) that enrich all the device object fields including Apple iPhone model category and device ID. The PR enables Taboola users to benefit from these eco-system improvements.

## Other information
cc: @mikizi @ahmadlob @barRubi

### DIFF
--- a/modules/taboolaBidAdapter.js
+++ b/modules/taboolaBidAdapter.js
@@ -276,7 +276,7 @@ function getSiteProperties({publisherId}, refererInfo, ortb2) {
 function fillTaboolaReqData(bidderRequest, bidRequest, data) {
   const {refererInfo, gdprConsent = {}, uspConsent} = bidderRequest;
   const site = getSiteProperties(bidRequest.params, refererInfo, bidderRequest.ortb2);
-  deepSetValue(data, 'device.ua', navigator.userAgent);
+  deepSetValue(data, 'device', bidderRequest?.ortb2?.device || {ua: navigator.userAgent});
   const extractedUserId = userData.getUserId(gdprConsent, uspConsent);
   if (data.user == undefined) {
     data.user = {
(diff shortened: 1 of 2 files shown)


modules/taboolaBidAdapter.js
```diff
@@ -276,7 +276,7 @@ function getSiteProperties({publisherId}, refererInfo, ortb2) {
 function fillTaboolaReqData(bidderRequest, bidRequest, data) {
   const {refererInfo, gdprConsent = {}, uspConsent} = bidderRequest;
   const site = getSiteProperties(bidRequest.params, refererInfo, bidderRequest.ortb2);
-  deepSetValue(data, 'device.ua', navigator.userAgent);
+  deepSetValue(data, 'device', bidderRequest?.ortb2?.device || {ua: navigator.userAgent});
```
patmmccann, line 279: I don't think this works; ua isn't in the current device object so this effectively deletes it, as the fallback will almost never happen

patmmccann: Oh ua is there, so the fallback should just be deleted

https://github.com/prebid/Prebid.js/blob/66605617791f289988ca166428d348885a84f839/src/fpd/enrichment.js#L100

justadreamer: Thanks Patrick, the fallback has been removed in the latest commit.

patmmccann (approved):

Issue #11740, opened 2024-06-08
GitHub actions: improve code duplicatuon output


patmmccann:
The output should report on duplication in the changed files, fail if there is duplication warnings in the changed files, and comment on GitHub pr's. 

Continuation of #11738 and #11739

PR #11757, opened 2024-06-10
Code duplication: Update jscpd.yml to run in repo context

patmmccann:
continuation of #11740

this seems to stop the thing from running :(

How do we make it so it comments on pr's from forks?


patmmccann:
goal is to make comments like this one appear on commits from forks https://github.com/prebid/Prebid.js/pull/11758#issuecomment-2158625082

PR #11764, opened 2024-06-11
Missena Bid Adapter : add session tracking, update referer & location fields

pdamoc:
## Type of change

- [x] Feature

## Description of change

This PR implements session tracking logic and restructures the referer related fields.

## Other information

@ysfbsf  is responsible for the implementation.

### DIFF
--- a/modules/missenaBidAdapter.js
+++ b/modules/missenaBidAdapter.js
@@ -105,7 +105,8 @@ export const spec = {
   buildRequests: function (validBidRequests, bidderRequest) {
     const capKey = `missena.missena.capper.remove-bubble.${validBidRequests[0]?.params.apiKey}`;
     const capping = safeJSONParse(storage.getDataFromLocalStorage(capKey));
-    const referer = bidderRequest?.refererInfo?.referer;
+    const referer = bidderRequest?.refererInfo?.ref;
+
     if (
       typeof capping?.expiry === 'number' &&
       new Date().getTime() < capping?.expiry &&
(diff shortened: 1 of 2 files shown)


modules/missenaBidAdapter.js
```diff
@@ -148,18 +148,19 @@ export const spec = {
...
       }
-      if (bidRequest.params.placement) {
-        payload.placement = bidRequest.params.placement;
+      if (bidParams.placement) {
+        payload.placement = bidParams.placement;
       }
-      if (bidRequest.params.formats) {
-        payload.formats = bidRequest.params.formats;
+      if (bidParams.formats) {
+        payload.formats = bidParams.formats;
```
patmmccann, line 160: it seems you are just changing variable names to avoid the detection

ysfbsf: How can we avoid triggering duplication errors in this case?

patmmccann: You should import common functions from a library

patmmccann:
It appears your commits have intended to fool the duplication check instead of fixing the underlying issue.

patmmccann:
Also please note jsdoc issues: 


/home/circleci/Prebid.js/modules/missenaBidAdapter.js
   65:0  warning  Syntax error in namepath: -                                             jsdoc/valid-types
   65:0  warning  Expected @param names to be "validBidRequests, bidderRequest". Got "-"  jsdoc/check-param-names
  173:0  warning  The type 'data' is undefined                                            jsdoc/no-undefined-types
  173:0  warning  Expected @param names to be "timeoutData". Got "Containing"             jsdoc/check-param-names
  181:0  warning  Expected @param names to be "bid". Got "The"                            jsdoc/check-param-names

PR #11768, opened 2024-06-11
PBjs Utils/Module Change: Replace JSON.parse/stringify with Efficient Deep Clone Method

bbaresic:
## Type of change

## Description of change
This replaces all the `JSON.parse/stringify` instances with the faster `deepClone (klona)`.

`JSON.parse/stringify` comes with many caveats:
- it is slower than some other methods of doing deep cloning
- both `parse` and `stringifiy` can throw, and many places in code do not address this
- it has many implicit behaviors, like casting dates to strings, or removing properties that are empty

## Other information
Related: #11399, #11418

### DIFF
--- a/src/utils.js
+++ b/src/utils.js
@@ -1150,3 +1150,42 @@ export function binarySearch(arr, el, key = (el) => el) {
   }
   return left;
 }
+
+/**
+ * Checks if an object has non-serializable properties.
+ * Non-serializable properties are functions and RegExp objects.
+ *
+ * @param {Object} obj - The object to check.
+ * @param {Set} checkedObjects - A set of properties that have already been checked.
+ * @returns {boolean} - Returns true if the object has non-serializable properties, false otherwise.
+ */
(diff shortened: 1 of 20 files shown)


src/utils.js
```diff
@@ -1150,3 +1150,42 @@ export function binarySearch(arr, el, key = (el) => el) {
   }
   return left;
 }
+
+/**
+ * Checks if an object has non-serializable properties.
+ * Non-serializable properties are functions and RegExp objects.
+ *
+ * @param {Object} obj - The object to check.
+ * @param {Set} checkedObjects - A set of properties that have already been checked.
```
bbaresic, line 1159: The `Set` part could be skipped, probably netting some memory wins, but only if we can guarantee that the objects coming in are not circular (`JSON.parse/stringify` would throw anyway if that is the case.)

patmmccann (approved):

mmoschovas:
Approved in conversation but this was merged before I had a chance to submit my review. But lgtm

PR #11770, opened 2024-06-12
Setupad adapter: param change and minor fixes

eldzis:
## Type of change

- [x] Bugfix
- [ ] Feature
- [ ] New bidder adapter  
- [x] Updated bidder adapter   
- [ ] Build related changes
- [ ] CI related changes

- [ ] Other

## Description of change
Changed account_id param from optional to required, updated tests accordingly and made a minor bug fix related to URL encoding for REPORT_ENDPOINT. &currency and &timestamp params were encoded as "¤cy=" and "×tamp=" during bundling process.

## Other information
SETUPAD prebid.github.io PR: [prebid/prebid.github.io#5411](https://github.com/prebid/prebid.github.io/pull/5411)

### DIFF
--- a/modules/setupadBidAdapter.js
+++ b/modules/setupadBidAdapter.js
@@ -230,8 +230,6 @@ export const spec = {
 
     if (!placementIds) return;
 
-    let extraBidParams = '';
-
     // find the winning bidder by using creativeId as identification
     if (biddersCreativeIds.hasOwnProperty(bid.creativeId) && biddersCreativeIds[bid.creativeId]) {
       bidder = biddersCreativeIds[bid.creativeId];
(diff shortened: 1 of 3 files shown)


patmmccann:
Hi, the getAd function appears to duplicated in your adapter and another. Please modify both to import it, thanks!

eldzis:
Hi @patmmccann, after closer inspection I stripped down the function to suit only our adapter needs and refactored to make it more readable. Now there is no duplicated code and as the function is very basic I don't think there's a need for moving main nextMillenium getAd function and importing it as it would only add more unnecessary code to our adapter.

patmmccann (approved):

PR #11773, opened 2024-06-12
ZetaGlobalSsp adapter: merge ortb2.site and params.site

asurovenko-zeta:
## Type of change
- [x] Feature
Merge params.site and ortb2.site in payload.site


patmmccann:
Looks to be failing tests and requiring several code block imports to solve duplication with zmaticoo

asurovenko-zeta:
@patmmccann thanks for review. That issue has been fixed

patmmccann:
Doesn't look fixed, this adapter and zmatico should have a bunch of common imports committed

surovenko:
@patmmccann still didn't understand the issue here. Should I change not our zmaticoo Adapter? How to know what exactly lines should I move to common methods? To where? Do you have any example of this? First time saw the issue like this
thanks!

patmmccann:
#8527 demonstrates moving common code into the library folder. #11831 is an example moving something to utils

surovenko:
@patmmccann done. Now should be fine. Thanks!

surovenko:
@patmmccann
Our adapter was simply copied and I cannot reasonably avoid this Github warning.
I tried to start moving methods into the utility class. But first of all, to avoid this warning I will have to copy everything from adapter. Secondly, I expect in the utility class only functions that can be used by everyone, some general functions that do not belong to some adapter, but some general things. But since our adapter was copied almost completely, we will have to transfer functions from our adapter that do not fit the definition of utility classes.
I don't know what to do in this situation. On the one hand, I want to remove the Github warning, on the other hand, it will be extremely difficult to remove it if I have 2 almost completely identical 2 files.

patmmccann:
Are you not affiliated with zmatico? Have you seen the example in #11854 in which 20+ roughly identical adapters moved their functions into a library?

patmmccann:
Your comments are quite reasonable, we're thinking about what to do still

patmmccann (approved):

patmmccann:
One option, your adapter simply needs many updates: almost everything you take in a param needs the treatment of this pr, including the user object especially. Your usersync and constants block can reasonably be moved to /libraries/zetaUtils and we can fix zmaticoo to import them.

Issue #11310, opened 2024-04-09
minBidCacheTTL better memory optimization


andyblackwell:
## Type of issue
feature request

## Description
currently `minBidCacheTTL` is applied at the auction level, so if you get a single bid with a long TTL, like an hour, then all the bids in the auction stay around for the full hour
https://github.com/prebid/Prebid.js/blob/4577159c0ec31896e5e0bd7654638861a265c189/src/auctionManager.js#L43-L48

any reason we couldn't remove bids as they expire directly, in the auction object?
https://github.com/prebid/Prebid.js/blob/4577159c0ec31896e5e0bd7654638861a265c189/src/auction.js#L156

could leave the auction-level removal as well, as that also frees up memory from all the other stored info in an auction like: requests, nobids, non bids, winning bids, etc

## Other information
#10308

patmmccann:
@andyblackwell great Idea! Does aditude plan to submit the fix?

fowler446:
@patmmccann How about this ^

patmmccann:
closing as stale, picking up at prebid.org

PR #11776, opened 2024-06-12
PBjs Core : expire stale bids directly from auction object to prevent long TTL bid

fowler446:
## Type of change

- [x ] Feature

## Description of change

By expiring bids after their TTL has elapsed directly in the auction object we prevent bids with long TTLs from hogging memory. At the moment when there is a long TTL bid in an auction, all the other bids are kept long after their expiry.

## Relevant Issue

https://github.com/prebid/Prebid.js/issues/11310

### DIFF
--- a/test/spec/auctionmanager_spec.js
+++ b/test/spec/auctionmanager_spec.js
@@ -852,30 +852,6 @@ describe('auctionmanager.js', function () {
         config.resetConfig();
       });
 
-      it('are dropped after their last bid becomes stale (if minBidCacheTTL is set)', () => {
-        config.setConfig({
-          minBidCacheTTL: 0
-        });
-        bids = [
-          {
-            adUnitCode: ADUNIT_CODE,
-            adUnitId: ADUNIT_CODE,
-            ttl: 10
-          }, {
-            adUnitCode: ADUNIT_CODE,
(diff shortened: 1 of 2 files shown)


test/spec/auctionmanager_spec.js
```diff
@@ -852,66 +852,54 @@ describe('auctionmanager.js', function () {
...
-            ttl: 10
-          }],
-          entries: () => auctionManager.getBidsReceived()
-        },
-        'no bids': {
-          bd: [],
-          entries: () => auctionManager.getNoBids()
-        }
-      }).forEach(([t, {bd, entries}]) => {
-        it(`with ${t} are never dropped if minBidCacheTTL is not set`, () => {
```
dgirardi, line 905: would it be better to keep this behavior? expire individual bids the same way auctions are, after `Math.max(bid.ttl, getConfig('minBidCacheTTL'))`, and only if `minBidCacheTTL` is set. This way pubs can choose whether they want this / they don't have to reason about two different expiration mechanisms.

patmmccann: @fowler446 are you still working on this? Can you provide your perspective on this?

patmmccann:
minBidCacheTTL

PR #11784, opened 2024-06-12
Sovrn bid adapter add ortb2 device

jwrosewell:
## Type of change
- [x] Feature
- [x] Updated bidder adapter

## Description of change
This PR enhances the Sovrn bidder request by incorporating device data from the global ORTB2 object. Existing values in the Sovrn device object will be overridden if they are also present in the global ORTB2 object. However, values unique to Sovrn or not found in the global ORTB2 object will remain unchanged.

## Motivation
The device object has previously been populated by simplistic parsers, if at all, and was inaccurate as a result. Prebid now benefits from RTD modules such as [51Degrees](https://docs.prebid.org/dev-docs/modules/51DegreesRtdProvider.html) that enrich all the device object fields including Apple iPhone model category and device ID. The PR enables Sovrn’s users to benefit from device object improvements.

## Other information
cc: @TWrightSovrn

### DIFF
--- a/modules/sovrnBidAdapter.js
+++ b/modules/sovrnBidAdapter.js
@@ -6,7 +6,10 @@ import {
   logError,
   deepAccess,
   isInteger,
-  logWarn, getBidIdParameter, isEmptyStr
+  logWarn,
+  getBidIdParameter,
+  isEmptyStr,
+  mergeDeep
 } from '../src/utils.js';
 import { registerBidder } from '../src/adapters/bidderFactory.js'
 import {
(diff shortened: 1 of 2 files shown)


Goncakkd:
> Whoa there, partner! 🌵🤠 We wrangled some duplicated code in your PR:
> 
> * `modules/deepintentBidAdapter.js` has 9 duplicated lines with `modules/sovrnBidAdapter.js`
> * `modules/deepintentBidAdapter.js` has 14 duplicated lines with `modules/sovrnBidAdapter.js`
> * `modules/deepintentBidAdapter.js` has 12 duplicated lines with `modules/sovrnBidAdapter.js`
> 
> Reducing code duplication by importing common functions from a library not only makes our code cleaner but also easier to maintain. Please move the common code from both files into a library and import it in each. Keep up the great work! 🚀

Hello, Check for duplicated code prevents us to merge PR and touching code of the other bid adapters doesnot make sense. Can you please help us for this case?

patmmccann (approved):

PR #11785, opened 2024-06-12
Sharethrough bid adapter add ortb2 device

jwrosewell:
## Type of change
- [x] Feature
- [x] Updated bidder adapter

## Description of change
This PR enhances the Sharethrough bidder request by incorporating device data from the global ORTB2 object. Existing values in the Sharethrough device object will be overridden if they are also present in the global ORTB2 object. However, values unique to Sharethrough or not found in the global ORTB2 object will remain unchanged.

## Motivation
The device object has previously been populated by simplistic parsers, if at all, and was inaccurate as a result. Prebid now benefits from RTD modules such as [51Degrees](https://docs.prebid.org/dev-docs/modules/51DegreesRtdProvider.html) that enrich all the device object fields including Apple iPhone model category and device ID. The PR enables Sharethrough’s users to benefit from device object improvements.

## Other information
cc: @jefftmahoney

### DIFF
--- a/modules/sharethroughBidAdapter.js
+++ b/modules/sharethroughBidAdapter.js
@@ -68,6 +68,11 @@ export const sharethroughAdapterSpec = {
       req.device.ext['cdep'] = bidderRequest.ortb2.device.ext.cdep;
     }
 
+    // if present, merge device object from ortb2 into `req.device`
+    if (bidderRequest?.ortb2?.device) {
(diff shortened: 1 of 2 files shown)


jefftmahoney:
Hey @jwrosewell !  This looks like a solid request - let me run it by our ad-server team to make sure it doesn't interfere with any logic on their side for how to handle device-related info when they're processing bid requests.

jefftmahoney (approved):

PR #11786, opened 2024-06-12
Grid bid adapter add ortb2 device

jwrosewell:
## Type of change
- [x] Feature
- [x] Updated bidder adapter

## Description of change
This PR enhances the Grid bidder request by incorporating device data from the global ORTB2 object. Existing values in the Grid device object will be overridden if they are also present in the global ORTB2 object. However, values unique to Grid or not found in the global ORTB2 object will remain unchanged.

## Motivation
The device object has previously been populated by simplistic parsers, if at all, and was inaccurate as a result. Prebid now benefits from RTD modules such as [51Degrees](https://docs.prebid.org/dev-docs/modules/51DegreesRtdProvider.html) that enrich all the device object fields including Apple iPhone model category and device ID. The PR enables Grid’s users to benefit from device object improvements.

## Other information
cc: @TheMediaGrid


jwrosewell:
@TheMediaGrid - this PR will ensure Grid benefits from publishers efficiently populating the device object with enriched data including iPhone model categories and other features not obtainable via SUA and UA. As publishers progress rollout this becomes more critical.

leonardlabat:
Hello @jwrosewell - it looks good to us

patmmccann (approved):

Issue #11268, opened 2024-03-28
Feature request: pause memory cleanup operations (`minBidCacheTTL` / `eventHistoryTTL`) while window is out of focus


spormeon:
I'm more requlary seeing some prebid error for Error rendering ad (id: 2337c3ca576e2468): Cannot find ad '2337c3ca576e2468'  and usually accompanied by this one: adRenderFailed (cannotFindAd).  I'm thinking this is related to bidcache but not 100% sure, the oservation seems to be, load a page, navitage "off" page, leave for say 5 mins, go back to page an no ad will load.  I think this is due to using a "reduced"TTL of e.g 180 from #10308 functionality.  I'm using lazyload with "fetch", "render" and what I think is happening is, when going back to page the "fetch" is happening but "render" cant take place due to TTL expired, so the ad ID has been cleared out/ wiped out. Any way to either stop the "fetch" and/ or have the TTL "pause" while "off page", so TTL only "clocks down" when "on" an active page

p.s i'm not 100% sure this is whats actually happening 100% yet, but its my "best guess" so far

dgirardi:
If a bid is past its TTL, is it still valuable to render it? I thought TTL meant "once it's this old, I won't pay for this impression".

spormeon:
if its "past", its gone, non payer etc but i'm thinking and not sure if its even possible,  lets say the bid returned has TTL of 600, i'm setting minBidCacheTTL:180,  this 180 starts from when bid retunred? but if someone navigates away from page, then back on page  4 mins later the 180TTL has gone but prebid seems to be trying to still "render" it or at least looking for it and throwing this error, could the minBidCacheTTL:180 take into account "off page time", so only counts down on "live" page?

dgirardi:
with `minBidCacheTTL` set to 180,  a bid that has TTL of 30 is discarded after 180 seconds; a bid that has TTL of 600 is discarded after 600 seconds (with the caveat that TTL also includes the [TTL buffer](https://docs.prebid.org/dev-docs/publisher-api-reference/setConfig.html#setConfig-ttlBuffer)).

It's plausible that if the window went out of focus in between an auction and the ad render, the ad render may happen after the TTL expired. I think it'd be possible to "pause" the TTL in that case. But I'm not sure it makes sense, I'd expect the buyer to not pay for it.

spormeon:
the "standard"that come back are 
banner: 600
native: 300
viodeo: 3600
[...]

PR #11790, opened 2024-06-13
Prebid Core: TTL counts only when page is active

mkomorski:
## Type of change

- [ ] Bugfix
- [x] Feature
- [ ] New bidder adapter  
- [ ] Updated bidder adapter   
- [ ] Build related changes
- [ ] CI related changes

- [ ] Other

## Other information

#11268


src/utils/deferredTimeout.js
```diff
@@ -0,0 +1,31 @@
...
+    timerId = setTimeout(callback, timeLeft);
+  }
+
+  function handleFocusChange() {
+    document.hidden ? pause() : resume();
+  }
+
+  document.addEventListener('visibilitychange', handleFocusChange);
+
+  return timerId;
```
dgirardi, line 30: This is not a "true" timer ID in the sense that calling `clearTimeout(timerId)` wouldn't necessarily work as expected. It may be better to return nothing (`ttlCollection` doesn't really need the timer ID), or if you wish, return something (a function?) that can be used to properly cancel the timeout.

src/utils/deferredTimeout.js
```diff
@@ -0,0 +1,31 @@
...
+  function resume() {
+    start = Date.now();
+    timerId = setTimeout(callback, timeLeft);
+  }
+
+  function handleFocusChange() {
+    document.hidden ? pause() : resume();
+  }
+
+  document.addEventListener('visibilitychange', handleFocusChange);
```
dgirardi, line 28: I wonder if we could do this with only one event listener (for better performance). 
E.g. use this to keep a global counter of how much time is spent out of focus; and on each individual timeout, take note of the counter at the beginning, and at the end re-schedule if it increased.

src/utils/deferredTimeout.js
```diff
@@ -0,0 +1,31 @@
+/**
+ * Wraps native setTimeout function in order to count time only when page is focused
+ *
+ * @param {function(*): (number|Promise<number>)} [callback] - A function that will be invoked after passed time
+ * @param {number} [milliseconds] - Minimum duration (in milliseconds) after the callback will be executed
+ * @returns {number} A timer id
+ */
+export default function setDeferredTimeout(callback, milliseconds) {
```
dgirardi, line 8: All timeouts are deferred. Maybe `focusTimeout` ?

PR #11791, opened 2024-06-13
Magnite Analytics: Minor fix for default rule name

robertrmartinez:
## Type of change
- [X] Bugfix

## Description of change
We would not send `rule` if did not detect cookieless which is not right! Set default now.

### DIFF
--- a/test/spec/modules/magniteAnalyticsAdapter_spec.js
+++ b/test/spec/modules/magniteAnalyticsAdapter_spec.js
@@ -1758,15 +1758,38 @@ describe('magnite analytics adapter', function () {
       });
     });
     describe('cookieless', () => {
-      beforeEach(() => {
-        magniteAdapter.enableAnalytics({
-          options: {
-            cookieles: undefined
-          }
-        });
-      })
+      // beforeEach(() => {
+      //   magniteAdapter.enableAnalytics({
+      //     options: {
+      //       cookieles: undefined
+      //     }
+      //   });
+      // })
       afterEach(() => {
         magniteAdapter.disableAnalytics();
+      })
+      it('should not add cokieless and preserve original rule name', () => {
+        // Set the confs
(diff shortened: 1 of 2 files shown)


test/spec/modules/magniteAnalyticsAdapter_spec.js
```diff
@@ -1758,16 +1758,39 @@ describe('magnite analytics adapter', function () {
...
-        magniteAdapter.enableAnalytics({
-          options: {
-            cookieles: undefined
-          }
-        });
-      })
+      // beforeEach(() => {
+      //   magniteAdapter.enableAnalytics({
+      //     options: {
+      //       cookieles: undefined
```
patmmccann, line 1764: it is on the comment now, but was spelling causing an issue?

test/spec/modules/magniteAnalyticsAdapter_spec.js
```diff
@@ -1758,16 +1758,39 @@ describe('magnite analytics adapter', function () {
...
+      //     }
+      //   });
+      // })
       afterEach(() => {
         magniteAdapter.disableAnalytics();
       })
+      it('should not add cokieless and preserve original rule name', () => {
```
patmmccann, line 1771: cookieless

test/spec/modules/magniteAnalyticsAdapter_spec.js
robertrmartinez: no it was something else (mot initializing the variable), this bit of code just does not even do anything so I commented out while testing and forgot to remove, I deleted now.

test/spec/modules/magniteAnalyticsAdapter_spec.js
robertrmartinez: fixed!

patmmccann (approved):

PR #11798, opened 2024-06-14
Real Time Data Module: Mobian Brand Safety

ehb-mtk:
## Type of change
- [x ] Feature

## Description of change
This is a new RTD module which populates data from Mobian's brand safety API into the prebid environment for advertisers to use. There is a related documentation PR here: https://github.com/prebid/prebid.github.io/pull/5420

### DIFF
--- a/modules/mobianRtdProvider.js
+++ b/modules/mobianRtdProvider.js
@@ -1,19 +1,19 @@
 import { submodule } from '../src/hook.js';
 import { ajaxBuilder } from '../src/ajax.js';
-
 export const MOBIAN_URL = 'http://impact-analytics-staging.themobian.com';
 
 export const mobianBrandSafetySubmodule = {
   name: 'mobianBrandSafety',
   gvlid: null,
   init: init,
-  getTargetingData: getTargetingData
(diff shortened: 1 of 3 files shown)


modules/mobianRtdProvider.js
```diff
@@ -0,0 +1,53 @@
...
+  name: 'mobianBrandSafety',
+  gvlid: null,
+  init: init,
+  getTargetingData: getTargetingData
+};
+
+function init() {
+  return true;
+}
+function getTargetingData() {
```
patmmccann, line 16: This doesn't appear to make these data available in the bid request object

I would suggest site.ext.data.mobian

patmmccann (changes requested):

modules/mobianRtdProvider.js
ehb-mtk: Thanks for the prompt feedback! After doing some reading up, it seemed like the best way to access and insert data into `site.ext.data` was via `getBidRequestData`, which is the current approach.

patmmccann (approved):

PR #11801, opened 2024-06-14
Smaato Bid Adapter : add DOOH support

Enigo:
## Type of change

- [ ] Bugfix
- [x] Feature
- [ ] New bidder adapter  
- [ ] Updated bidder adapter   
- [ ] Build related changes
- [ ] CI related changes

- [ ] Other

## Description of change
Adding DOOH support for Smaato adapter

### DIFF
--- a/modules/smaatoBidAdapter.js
+++ b/modules/smaatoBidAdapter.js
@@ -212,15 +212,19 @@ const converter = ortbConverter({
       return bidderRequest.gdprConsent && bidderRequest.gdprConsent.gdprApplies;
     }
 
+    function setPublisherId(node) {
+      deepSetValue(node, 'publisher.id', bidRequest.params.publisherId);
+    }
+
     const request = buildRequest(imps, bidderRequest, context);
     const bidRequest = context.bidRequests[0];
-    let siteContent;
+    let content;
     const mediaType = context.mediaType;
     if (mediaType === VIDEO) {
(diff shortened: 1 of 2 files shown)


modules/smaatoBidAdapter.js
```diff
@@ -242,19 +246,26 @@ const converter = ortbConverter({
 
     if (request.site) {
       request.site.id = window.location.hostname
-      if (siteContent) {
-        request.site.content = siteContent;
+      if (content) {
+        request.site.content = content;
+      }
+      setPublisherId(request.site);
+    } else if (request.dooh) {
```
patmmccann, line 253: This feels very fragile; I would reverse the conditional and always send a dooh request if a dooh object exists, instead of putting it in the else.

Enigo: But there should be only one of `app`, `site` or `dooh` present in a request as per ortb specification

patmmccann: Right, I am saying, right now you will always choose site if both site and dooh exist. Do you prefer to always choose DOOH in that scenario and jettison the site object?

Enigo: We generally prefer `site` over `dooh`
I thought (might be confusing with prebid-server tho) that there's an internal check that prevents this kind of misconfiguration before it reaches the adapter code?

patmmccann (approved):

PR #11802, opened 2024-06-14
ID5 UserId module - integrate with TrueLink Id

abazylewicz-id5:
## Type of change
- [x] Feature

- [ ] Other

## Description of change
ID5 offers an option for publishers to integrate a TrueLink Id on the pages. If it's present on the page, we want to capture it and pass it to the server side. We also want to expose it as a EID so its later usable as a PPID


patmmccann:
Looks fine, seems like it needs docs

abazylewicz-id5:
I've added some docs about what eids are available, I can later add the same to https://docs.prebid.org/dev-docs/modules/userid-submodules/id5.html

patmmccann:
yes please make that commit

patmmccann (approved):

PR #11803, opened 2024-06-14
Prebid Core: TTL counts only when page is active

mkomorski:
## Type of change

- [ ] Bugfix
- [x] Feature
- [ ] New bidder adapter  
- [ ] Updated bidder adapter   
- [ ] Build related changes
- [ ] CI related changes

- [ ] Other

## Other information

#11268


olafbuitelaar:
isn't the bid TTL an hard cut off, from delivery? e.g. we've seen SSP's rejecting bids if the bid get's executed after delivery time + TTL expired. taking only focus time could attribute to executing the bid after it's ttl?

dgirardi:
@olafbuitelaar this is to fix an edge case (https://github.com/prebid/Prebid.js/issues/11268) where:

1. bid cache cleanup is enabled (with `minBidCacheTTL`)
2. a bid is selected as a winner and sent to the adserver
3. focus is lot, browser pauses pending tasks
4. focus comes back, the bid is past its TTL, the bid is dropped
5. the now missing bid fails to render.

in theory it shouldn't matter since the bid expired, but many bidders don't really have TTLs (a lot just hardcode some random number) or don't enforce it.

src/utils/focusTimeout.js
```diff
@@ -0,0 +1,35 @@
...
+    if (timeOutOfFocus === startTime && outOfFocusStart == null) {
+      callback();
+    } else if (outOfFocusStart != null) {
+      // case when timeout ended during page is out of focus
+      suspendedTimeouts.push({ callback, startTime })
+    } else {
+      setFocusTimeout(callback, timeOutOfFocus - startTime);
+    }
+  }, milliseconds);
+  return timerId;
```
dgirardi, line 34: the false return value came back!

dgirardi (approved):

patmmccann (approved):

PR #11820, opened 2024-06-18
Smarthub: add alias VimayX

SmartHubSolutions:
## Type of change

- [x] Updated bidder adapter   

## Description of change

Add alias VimayX


patmmccann:
Hello! We're unable to consider your submission without improvement of the duplication

PR #11829, opened 2024-06-19
SmartyadsBidAdapter remove usprivacy

rishko00:
## Type of change

- [ ] Bugfix
- [ ] Feature
- [ ] New bidder adapter  
- [ ] Updated bidder adapter   
- [ ] Build related changes
- [ ] CI related changes

- [ ] Other


modules/smartyadsBidAdapter.js
```diff
@@ -79,19 +79,19 @@ export const spec = {
...
+      'placements': placements,
       'deviceWidth': winTop.screen.width,
       'deviceHeight': winTop.screen.height,
+      'host': location?.domain,
+      'page': location?.page,
       'language': (navigator && navigator.language) ? navigator.language : '',
-      'secure': 1,
-      'host': location?.domain ?? '',
-      'page': location?.page ?? '',
       'coppa': config.getConfig('coppa') === true ? 1 : 0,
```
patmmccann, line 88: We will not be accepting tricks to fill the duplication check

patmmccann (changes requested):
Please do not attempt malicious changes designed to circumvent our rules

- [

modules/smartyadsBidAdapter.js
rishko00: @patmmccann OK, but how can I fix that? I saw the report and duplication check marks, for example, the isBidResponseValid function is used in lots of adapters and looks almost the same.

patmmccann: Check out this work in progress from an adapter structured very similar to yours

patmmccann: https://github.com/MaksymTeqBlaze/Prebid.js/commit/0249eb7a258d610e937a114676e24c1d3f6c956e

rishko00: @patmmccann I'm sorry, but I still don't get it. The previous version of the isBidResponseValid function was flagged as duplicate code by your tool, but it doesn't need any updates, right? And the same function with same code is used in other adapters

patmmccann: You and other sites have duplicate code, we need you to move common functions into libraries before making further updates to your adapter

patmmccann: #8527 has an example on how to move common functions into a library

rishko00:
@patmmccann, I've moved some functions to the libs and removed some unused fields from the request and unused functions. However, there are still common functions, such as isBidResponseValid and interpretResponse. If I move isBidResponseValid to libs, there will still be an issue with duplicated code, but I can't rewrite it because it's used for common validations. Also I didn't find example of adapter where those functions are working properly. I need your advice on how to proceed.

patmmccann:
Hi @rishko00 take a look at https://github.com/MaksymTeqBlaze/Prebid.js/commit/0249eb7a258d610e937a114676e24c1d3f6c956e#diff-a442c024395533115322fcbbfe796836b8ea37974565360dda15e9c93fbca181 or at #11849 to see examples of bidders merging common functions. The library function you added looks great but I am not sure if it would be imported in multiple adapters.

patmmccann:
#11854 has merged in and is a great example

rishko00:
@patmmccann yes, got it, I've pushed changes

patmmccann:
Looks great! Can you undo your changes to render files and we're good, thanks!!

patmmccann (approved):

PR #11840, opened 2024-06-20
WURFL RTD submodule: initial version

lucor:
## Type of change

- [x] Feature
- [x] New bidder adapter  

## Description of change

The WURFL RTD module enriches the OpenRTB 2.0 device data with [WURFL data](https://www.scientiamobile.com/wurfl-js-business-edition-at-the-intersection-of-javascript-and-enterprise/).
The module sets the WURFL data in `device.ext.wurfl` and all the bidder adapters will always receive the low entry capabilites like `is_mobile`, `complete_device_name` and `form_factor`. 
For a more detailed analysis bidders can subscribe to detect iPhone and iPad models and receive additional [WURFL device capabilities](https://www.scientiamobile.com/capabilities/?products%5B%5D=wurfl-js).

Integration details available in `modules/wurflRtdProvider.md`

## Maintainer

prebid@scientiamobile.com


modules/wurflRtdProvider.js
```diff
@@ -0,0 +1,233 @@
...
+  }
+
+  var payload = JSON.stringify({ bidders: [...enrichedBidders] });
+  if (navigator.sendBeacon !== undefined) {
+    navigator.sendBeacon(url.toString(), payload);
+    return;
+  }
+
+  if (window.fetch) {
+    fetch(url.toString(), {
```
patmmccann, line 209: Please import communication methods from ajax.js

lucor: Updated in 68e88b6

modules/wurflRtdProvider.js
```diff
@@ -0,0 +1,234 @@
...
+  if (altHost) {
+    host = altHost;
+  }
+
+  const wjsPromise = new Promise((resolve, reject) => {
+    const url = new URL(host);
+    url.pathname = WURFL_JS_ENDPOINT_PATH;
+
+    const runWjs = (src) => {
+      const wjs = document.createElement('script');
```
patmmccann, line 65: You add yourself to adloader but then you didn't import and use that method

lucor: Updated in 32d1c0e

Rothalack:
I gave this a test with your example test page. At first it was failing to load because it seems you don't have SSL set up yet to run https on your endpoint. I got my browser to ignore that and tried again. Now the https://prebid.wurflcloud.com/wurfl.js?mode=prebid endpoint is 404ing. Is that an issue you're aware of or am I missing something about setting up the test?

modules/wurflRtdProvider.js
patmmccann: Doesn't appear fixed, you have to user the loadexternalscript method

lucor:
@Rothalack sorry about that. Yes, we're aware of the issue. The prebid server code hasn't been deployed to production yet. That's why we submitted the PR as a draft; we should have clarified that in the description. We'll post an update here once the code goes live.

modules/wurflRtdProvider.js
lucor: Updated to use loadExternalScript in 1bb2df6

patmmccann:
I converted back to draft, welcome to the community!

lucor:
Thank you for the warm welcome and the helpful feedback!

patmmccann (approved):

PR #11843, opened 2024-06-20
Es6 build support

Shadialsayed:
Add es6 module support to prebidjs build

### DIFF
--- a/babelConfig.js
+++ b/babelConfig.js
@@ -23,12 +23,9 @@ module.exports = function (options = {}) {
           'modules': options.test ? 'commonjs' : 'auto',
           targets: isES6 ? {
             browsers: [
-              'Chrome >= 61',
-              'Firefox >= 57',
-              'Safari >= 10.1',
-              'iOS >= 10.3',
-              'Edge >= 15',
-              '> 1%'
+              '> 0.25%',
+              'not IE 11',
+              'not op_mini all'
             ]
           } : undefined
         }
--- a/babelConfig.js
+++ b/babelConfig.js
@@ -10,6 +10,8 @@ function useLocal(module) {
 }
 
 module.exports = function (options = {}) {
+  const isES6 = options.target === 'es6';
+
   return {
     'presets': [
       [
@@ -19,6 +21,11 @@ module.exports = function (options = {}) {
           'corejs': '3.13.0',
           // a lot of tests use sinon.stub & others that stopped working on ES6 modules with webpack 5
           'modules': options.test ? 'commonjs' : 'auto',
+          targets: isES6 ? {
+            browsers: [
+              '> 1%',
+            ]
+          } : undefined
         }
       ]
     ],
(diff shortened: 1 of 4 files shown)


babelConfig.js
```diff
@@ -19,6 +21,11 @@ module.exports = function (options = {}) {
           'corejs': '3.13.0',
           // a lot of tests use sinon.stub & others that stopped working on ES6 modules with webpack 5
           'modules': options.test ? 'commonjs' : 'auto',
+          targets: isES6 ? {
```
patmmccann, line 24: We're managing the target here

https://github.com/prebid/Prebid.js/blob/1a8b99330e53b4ae50f367d6a764a2f7dddb4f68/package.json#L29

Changing it would be breaking

patmmccann:
To confirm: you're trying to build for older browsers than the current target, not more modern, correct? The 1 percent target in your fallback doesn't match our target

babelConfig.js
muuki88: Yeah, we noticed as well. @Shadialsayed is one my engineers and we are going to use this in our infrastructure soon.

The `package.json` config is too static. And because this change only overrides the `package.json` if there's the `es6` build cli parameter, it isn't a breaking change. My argument would be that the `package.json` is the default, which can be overriden by build parameters.

patmmccann: this appears to fallback to the 1 percent target

patmmccann: this has now changed the target to match, but now keeping them in sync seems difficult. Can we import a constant?

muuki88: We are doing some more tests, but it seems that the output bundle is still a lot smaller, even the target is the same.

If it works without specifying a target here, we'll remove it here.

patmmccann: I'm a bit confused, looking forward to your tests

muuki88:
Hi @patmmccann 

We will probably close this. We assume that the `require` statements that are still quite heavily used in the code base, prohibit any ES6 modules build. However the main part, is not the esm stuff, but the polyfills that are not required.

patmmccann:
You said you were able to improve the build size. Could you quantify that? 

What was the build size:
prior to the pr,
with the pr but no option selected,
with the pr but with the es6 option selected?

muuki88:
We did improve the build size. However the browser target accounted for less than 4kb overall, which is not very much.  The main difference came from the polyfills we were able to omit in the final ad tag, that includes prebid.

The trick is to use the `nomodule` and `type="module"` attributes. See https://www.sitepoint.com/using-es-modules/ to put the same ad tag twice on the page, but the `nomodule` one has the polyfills and the other is an es6 module kinda thing. This looks like this

```html
<!-- legacy es5 bundle (IE11) -->
<script async nomodule src="https://example.com/adtag_es5.js"></script>
<!-- modern es6 bundle -->
<script async type="module" src="https://example.com/adtag.mjs></script>
```

A browser only loads one of the two. Sorry for the noise. We took a few wrong turns in the beginning and thought it looked great, but actually we misconfigured the build and not all chunks were inlcuded in the final distribution :see_no_evil:

PR #11844, opened 2024-06-20
PubMatic Analytics Adapter : Logging MultiBids from bidders

pm-azhar-mulla:
## Type of change

- [x] Feature

## Description of change

Earlier, we were not logging the multibids from bidders. With this update, we have added support for logging multibids. This enhancement ensures that all bids from bidders are recorded, providing better transparency and insights into the bidding process


modules/pubmaticAnalyticsAdapter.js
```diff
@@ -407,7 +417,7 @@ function executeBidsLoggerCall(e, highestCpmBids) {
       'au': origAdUnit.owAdUnitId || getGptSlotInfoForAdUnitCode(adUnitId)?.gptSlot || adUnitId,
       'mt': getAdUnitAdFormats(origAdUnit),
       'sz': getSizesForAdUnit(adUnit, adUnitId),
-      'ps': gatherPartnerBidsForAdUnitForLogger(adUnit, adUnitId, highestCpmBids.filter(bid => bid.adUnitCode === adUnitId)),
+      'ps': gatherPartnerBidsForAdUnitForLogger(adUnit, adUnitId, highestCpmBids.filter(bid => bid.adUnitCode === adUnitId), e),
```
patmmccann, line 420: can you also fix this? https://github.com/prebid/Prebid.js/blob/4fc7f8865e3f6971b8d880f662a4ac03e05a7109/modules/pubmaticAnalyticsAdapter.js#L404 

'$prebid.version$' works, you do not need to import the global for it

patmmccann (changes requested):
mostly lgtm, one nitpick

patmmccann (approved):

PR #11851, opened 2024-06-21
orbidderBidAdapter.js: remove mediaType video to avoid broken requests

renebaudisch:
## Type of change
- [ ] Bugfix
- [ ] Updated bidder adapter

## Description of change
The bidder seems to not support multiType bidrequests containing video. It then always throws a 400.
Our contact told us, the bidder doesn't support video at all, so we updated the bidder to remove definitions for video on bidRequests.

## Other information
Recently changes to the bidder has been done by @arneschulz1984 and @hendrikiseke1979.
To ensure this change suits the bidder it would be great if one of those could do the review please

### DIFF
--- a/modules/orbidderBidAdapter.js
+++ b/modules/orbidderBidAdapter.js
@@ -93,6 +93,9 @@ export const spec = {
       if (bidderRequest && bidderRequest.refererInfo) {
         referer = bidderRequest.refererInfo.page || '';
       }
+      if (bidRequest?.mediaTypes?.video) {
+        delete bidderRequest.mediaTypes.video;
+      }
 
       bidRequest.params.bidfloor = getBidFloor(bidRequest);
 
--- a/modules/orbidderBidAdapter.js
+++ b/modules/orbidderBidAdapter.js
@@ -94,7 +94,7 @@ export const spec = {
         referer = bidderRequest.refererInfo.page || '';
       }
       if (bidRequest?.mediaTypes?.video) {
-        delete bidderRequest.mediaTypes.video;
+        delete bidRequest.mediaTypes.video;
       }
 
       bidRequest.params.bidfloor = getBidFloor(bidRequest);


modules/orbidderBidAdapter.js
```diff
@@ -93,6 +93,9 @@ export const spec = {
       if (bidderRequest && bidderRequest.refererInfo) {
         referer = bidderRequest.refererInfo.page || '';
       }
+      if (bidRequest?.mediaTypes?.video) {
```
patmmccann, line 96: I'm confused, on line 4 and 49 this bidder already makes it clear they do not support video

@dgirardi this pr seems to imply a feature request (bug?) in the ortb2 convertor, in which a bidder receives unsupported media types

dgirardi: this is not using ortb converter. maybe it makes sense for core to remove unsupported media types, but on the other hand, adapters do not typically just dump the entirety of their input like this. Any modification of any kind to bid requests might break it for all we know.

modules/orbidderBidAdapter.js
```diff
@@ -93,6 +93,9 @@ export const spec = {
...
+        delete bidderRequest.mediaTypes.video;
```
dgirardi, line 97: does this actually work? reading the code I don't think it has any effect on the outgoing request. (also, this `bidderRequest` object is not supposed to have a `mediaType` property).

It would help to have a reproducible issue, and translate that into a unit test.

renebaudisch:
Sorry, there is indeed a wrong reference for the delete.

But, I'm not related to the company behind the bidder. We can see on our sites, that the bidder gets mediaType.video but because they do not support it the bid results in an Error 400. Deleting worked in my tests, I'll provide testpages.

We as the auctioner do have no option to define which mediaType we want a bidder to participate, don't we? Maybe this change shouldn't be done but then how would I be able to do multiType for other adapters on the same placement but not for the orbidder?

dgirardi:
I agree that this should be fixed, but it's hard for us to verify it. I understand it's difficult for you as well, hopefully we can get orbidder to chime in (@hendrikiseke1979 ?)

As the publisher you can direct only some media types to certain bidders with [twin ad units](https://docs.prebid.org/dev-docs/adunit-reference.html#twin-adunit-codes). Basically breaking out orbidder into a duplicate ad unit definition that does not have video.

renebaudisch:
Ok, but this will work for us, I can split our adults per bidder, I didn't know that's possible. I will try this on monday!

patmmccann (approved):

PR #11853, opened 2024-06-21
Audiencelogy Bid Adapter : Initial Release

hasanideepak:
## Type of change

- [ ] Bugfix
- [ ] Feature
- [x] New bidder adapter  
- [ ] Updated bidder adapter   
- [ ] Build related changes
- [ ] CI related changes

- [ ] Other

### DIFF
--- a/modules/audiencelogyBidAdapter.js
+++ b/modules/audiencelogyBidAdapter.js
@@ -1,91 +1,98 @@
 import {
-  registerBidder
-} from '../src/adapters/bidderFactory.js';
-import {
-  BANNER
-} from '../src/mediaTypes.js';
-import {
   deepAccess,
   generateUUID,
   isArray,
   logError
 } from '../src/utils.js';
+import {
+  BANNER
+} from '../src/mediaTypes.js';
+import {
+  registerBidder
+} from '../src/adapters/bidderFactory.js';
+
 const BIDDER_CODE = 'audiencelogy';
 const ENDPOINT_URL = 'https://rtb.audiencelogy.com/prebid';
 
-function buildRequests(bidRequests, bidderRequest) {
-  const requests = [];
-  let nid = 0;
-  // Loop for each bid request
-  bidRequests.forEach(bid => {
-    // Create the bid request object
-    const request = {
-      id: generateUUID(),
-      imp: [{
-        id: bid.bidId,
-        bidfloor: fetchFloor(bid),
-        banner: fetchBanner(bid)
-      }],
-      placementId: bid.params.placement_id,
-      site: fetchSite(bidderRequest),
-      user: setUser(bid)
-    };
-    // Fetch GPP Consent from bidderRequest
-    if (bidderRequest?.gppConsent?.gppString) {
-      request.gpp = bidderRequest.gppConsent.gppString;
-      request.gpp_sid = bidderRequest.gppConsent.applicableSections;
-    } else if (bidderRequest?.ortb2?.regs?.gpp) {
-      request.gpp = bidderRequest.ortb2.regs.gpp;
-      request.gpp_sid = bidderRequest.ortb2.regs.gpp_sid;
-    }
-    // Fetch coppa compliance from bidderRequest
-    if (bidderRequest?.ortb2?.regs?.coppa) {
-      request.coppa = 1;
-    }
-    // Fetch uspConsent from bidderRequest
-    if (bidderRequest && bidderRequest.uspConsent) {
-      request.us_privacy = bidderRequest.uspConsent;
-    }
-    // Push the created bid request to the requests array
-    requests.push(request);
-    // Set nid value
-    nid = bid.params.nid;
-  });
-  // Return the array of bid requests
-  return {
-    method: 'POST',
-    url: `${ENDPOINT_URL}/${nid}`,
-    data: JSON.stringify(requests),
-    options: {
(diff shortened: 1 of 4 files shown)


modules/audiencelogyBidAdapter.js
```diff
@@ -123,41 +130,32 @@ function fetchSite(bidderRequest) {
...
-    creativeId: bid && bid.crid ? bid.crid : undefined,
-    meta: {
-      advertiserDomains: bid && bid.adomain ? bid.adomain : []
-    },
-    netRevenue: false,
-    currency: bid && bid.cur ? bid.cur : 'USD',
-    dealId: bid && bid.dealId ? bid.dealId : undefined,
-    ttl: 300
+const buildResponse = (bid) => {
+  let response = {};
```
patmmccann, line 134: We can't accept changes designed to fool the duplication checker

patmmccann (changes requested):
Please import common functions from a library

modules/audiencelogyBidAdapter.js
hasanideepak: I understand, please let me know where to add common functions in library?

ChrisHuie:
Please add docs to -> https://github.com/prebid/prebid.github.io

hasanideepak:
> Please add docs to -> https://github.com/prebid/prebid.github.io

Added https://github.com/prebid/prebid.github.io/pull/5457

modules/audiencelogyBidAdapter.js
patmmccann: Sure! Thank you. See #11854 for example or #11879

modules/audiencelogyBidAdapter.js
```diff
@@ -0,0 +1,161 @@
...
+  response.currency = bid && bid.cur ? bid.cur : 'USD';
+  response.dealId = bid && bid.dealId ? bid.dealId : undefined;
+  response.ttl = 300;
+  return response;
+}
+// Function to build the user object
+const setUser = (bid) => {
+  let user = {};
+  if (bid && bid.params) {
+    user.buyeruid = localStorage.getItem('adx_profile_guid') ? localStorage.getItem('adx_profile_guid') : '';
```
patmmccann, line 154: You have to import the storage manager to do this, publishers manage bidder access to storage depending on consent or their preferences

modules/audiencelogyBidAdapter.js
```diff
@@ -0,0 +1,161 @@
...
+        logError(err);
+      }
+    }
+    return resp;
+  }
+}
+// Function to fetch bid_floor
+const fetchFloor = (bid) => {
+  if (bid.params && bid.params.bid_floor) {
+    return bid.params.bid_floor;
```
patmmccann, line 97: Make sure to check the getFloor function from the request

patmmccann (changes requested):
Looks pretty good, couple of tweaks and use of an import please

hasanideepak:
@patmmccann I have added library audiencelogyUtils/bidderUtils.js and imported common code. Please verify.

patmmccann:
@hasanideepak yes but you should also import in relevate, which has the same concern above, it uses local storage without importing storage manager

patmmccann:
>  I will need to add same library in relevate fork in order to use it, doesnt that give conflict error as same code will be there in audiencelogy and relevate?

both will import it, but the build will be more compact as the code will only be in the project once. It will clear the error. 

> tried using storage manager, but it was unable to read the value from local storage. Can you please let me know some example.

https://github.com/prebid/Prebid.js/blob/ca24c395555167e5aba7007a83332c1d49441921/modules/ixBidAdapter.js#L110

What you have in relevate is a major rules violation, as it ignores consent. We need you to take care of it in this commit. I'll go ahead and submit a bug fix to delete what you are doing there now, so you'll want to merge in master after that merges to avoid merge conflicts

hasanideepak:
> > I will need to add same library in relevate fork in order to use it, doesnt that give conflict error as same code will be there in audiencelogy and relevate?
> 
> both will import it, but the build will be more compact as the code will only be in the project once. It will clear the error.
> 
> > tried using storage manager, but it was unable to read the value from local storage. Can you please let me know some example.
> 
> https://github.com/prebid/Prebid.js/blob/ca24c395555167e5aba7007a83332c1d49441921/modules/ixBidAdapter.js#L110
> 
> What you have in relevate is a major rules violation, as it ignores consent. We need you to take care of it in this commit. I'll go ahead and submit a bug fix to delete what you are doing there now, so you'll want to merge in master after that merges to avoid merge conflicts

@patmmccann Thank you for the support. So, I will make the changes for store manager in this PR (i.e.audiencelogy) and create a new PR for relevate adapter to import common library and store manager changes. Is this right ? 
Merging both PR wont cause any conflicts ?

patmmccann:
Just do it all in one?

hasanideepak:
@patmmccann Considering we are an ad tech service provider to relevate and we also need our own prebid adapter, we would like to focus on relevate adapter at this moment to make it fully compliant with your policies (with storage manager changes), hence we would like to pause work on our own adapter (i.e. audiencelogy) at this moment and we will revisit in a few weeks.
Thank you for your support

patmmccann:
No worries, just import stage manager instead of directly accessing local storage, keeping an eye out for your commit

Issue #10594, opened 2023-10-11
GPT interstitial support - deferred rendering


ourcraig:
## Type of issue
Feature Request

## Description
Currently when the Prebid Universal Creative is returned from Google Ad Manager for an out-of-page GPT interstitial slot, it renders instantly despite not being viewable. Prebid recommends the use of `adUnit.deferredBilling: true` to defer the `onBidBillable` event from firing on ad render, and calling `pbjs.triggerBilling(<deferredBid>)` to fire the `onBidBillable` event when the interstitial becomes viewable. However, it seems that bid adapters are not currently listening for the `onBidBillable` event ([ref](https://github.com/search?q=repo%3Aprebid%2FPrebid.js+onBidBillable&type=code)).

To have wider support for GPT interstitials, I would like to request Prebid support for **deferred rendering** using something like the below:

```
<script src = "https://cdn.jsdelivr.net/npm/prebid-universal-creative@latest/dist/%%PATTERN:hb_format%%.js"></script>
<script>
  var ucTagData = {};
  var render = function() {
    try {
      ucTag.renderAd(document, ucTagData);
    } catch (e) {
      console.log(e);
    }
  }
  ucTagData.adServerDomain = "";
  ucTagData.instl = "%%PATTERN:hb_instl%%";
  ucTagData.pubUrl = (ucTagData.instl === "true") ? "*" : "%%PATTERN:url%%";
  ucTagData.adId = "%%PATTERN:hb_adid%%";
  ucTagData.cacheHost = "%%PATTERN:hb_cache_host%%";
  ucTagData.cachePath = "%%PATTERN:hb_cache_path%%";
  ucTagData.uuid = "%%PATTERN:hb_cache_id%%";
  ucTagData.mediaType = "%%PATTERN:hb_format%%";
  ucTagData.env = "%%PATTERN:hb_env%%";
  ucTagData.size = "%%PATTERN:hb_size%%";
  ucTagData.winurl = "%%PATTERN:hb_winurl%%";
  ucTagData.winbidid = "%%PATTERN:hb_bidid%%";
  ucTagData.mobileResize = "hb_size:%%PATTERN:hb_size%%";
  ucTagData.requestAllAssets = true;
  ucTagData.clickUrlUnesc = "%%CLICK_URL_UNESC%%";
  if (ucTagData.instl === "true") {
    window.addEventListener('message', function(event) {
      try {
        var data = JSON.parse(event.data);
[...]

PR #11863, opened 2024-06-25
Core: add location method for cross-frame creatives and update creatives

dgirardi:
## Type of change

- [x] Feature

## Description of change

This adds a way for creatives to find which ancestor window is running Prebid, and updates dynamic creative logic to use it (PUC changes coming soon). 

It's related to https://github.com/prebid/Prebid.js/issues/10594 - in some situations the creative logic runs in a nested iframe.


patmmccann (approved):

src/constants.js
```diff
@@ -193,3 +193,5 @@ export const MESSAGES = {
   NATIVE: 'Prebid Native',
   EVENT: 'Prebid Event'
 };
+
+export const PB_LOCATOR = '__pb_locator__';
```
patmmccann, line 197: Very cool :)

Issue #11689, opened 2024-06-04
Google pairid submodule: low quality error message


mwilsonmagnite:
## Type of issue
Confusing syntax error occurring during pbjs_debug derived from liveramp submodule.

## Description
When there are no liverampValues available in cookie storage or localStorage, the module decodes the liverampValue's null value(using atob) and passes it to JSON.parse()  which throws an error that gets logged into the console while using pbjs_debug.

## Steps to reproduce
Error is found on these pages in the console:
- https://www.birdsandblooms.com/birding/attracting-hummingbirds/hummingbirds-small-garden/?pbjs_debug=true
- https://www.businessinsider.jp/?pbjs_debug=true

<img width="1728" alt="Screenshot 2024-06-04 at 10 14 07 AM" src="https://github.com/prebid/Prebid.js/assets/167801618/da6be699-2cbf-46d7-a439-8c3742867a94">

I believe that a more verbose INFO log would be more appropriate that states no liverampValues found instead of an esoteric and possibly concerning JSON.parse syntax error.

Found in prebid.js library code which includes liveramp submodule.

patmmccann:
Cc @danielsao

patmmccann:
This code is in the pairid module fwiw

https://github.com/prebid/Prebid.js/blob/47e6561919680454d1b57b06315744852a7e07f8/modules/pairIdSystem.js#L73

patmmccann:
@mwilsonmagnite I think this is solved, lmk if you agree

mwilsonmagnite:
@patmmccann looks good, thanks!

PR #11870, opened 2024-06-26
Update pairIdSystem.js: add defensive code before json parse

patmmccann:
fixes https://github.com/prebid/Prebid.js/issues/11689


ncolletti (approved):
lgtm

PR #11854, opened 2024-06-21
Libraries: New utility file to Remove code duplication in several adapters

MaksymTeqBlaze:
## Type of change

- [ ] Bugfix
- [ ] Feature
- [ ] New bidder adapter  
- [x] Updated bidder adapter   
- [ ] Code style update (formatting, local variables)
- [ ] Refactoring (no functional changes, no api changes)
- [ ] Build related changes
- [ ] CI related changes

- [ ] Does this change affect user-facing APIs or examples documented on http://prebid.org?
- [ ] Other

## Description of change
Created a library and utilized its functions in several adapters

PR #11874, opened 2024-06-27
Smarthub bid adapter: alias vimayx

SmartHubSolutions:
## Type of change

- [x] Updated bidder adapter   

## Description of change

add VimayX alias

## Other information

docs PR here https://github.com/prebid/prebid.github.io/pull/5474


patmmccann:
Hi @SmartHubSolutions , it appears several bid adapters are clones with yours. Could you clear up the duplication warnings with imports?

patmmccann:
It looks like #11854 could be very useful to you,  getPlacementReqData(bid) for example is available now in an import

ChrisHuie:
@SmartHubSolutions can you please add docs for this new adapter into our docs repo please? 

https://github.com/prebid/prebid.github.io

SmartHubSolutions:
@ChrisHuie added docs PR https://github.com/prebid/prebid.github.io/pull/5474

SmartHubSolutions:
@patmmccann made code update for remove dublicates

patmmccann:
Hi @SmartHubSolutions it appears your changes are designed to fool the detection instead of improve the code

SmartHubSolutions:
Hi @patmmccann,
You're right, I refactored the code to pass the duplicate check.
I've looked through all the code and unfortunately I don't see an option to import functions from the Techblast library as they are incompatible with our environment.
But in the meantime, I modified some functions and some functions moved outside the _spec_ object to make the code easier to understand.

patmmccann:
@SmartHubSolutions commits of this nature are a conduct issue, you're making our code base worse to hide from our code improvement suggestion utility. Please squash out immediately and please expand on why you cannot import code that is identical to yours and already in a library

SmartHubSolutions:
Hi @patmmccann,
after your updates the test fails.
I run the tests with the following command `gulp test --file test/spec/modules/smarthubBidAdapter_spec.js`
I have also prepared an explanation for code incompatibility, please see the attached file
[Explaining the differences.pdf](https://github.com/user-attachments/files/16139646/Explaining.the.differences.pdf)

dgirardi:
@SmartHubSolutions - hoping you won't mind, I have updated this PR to reuse the teqblaze utils. Of note:

 - I modified the shared utils slightly to enable you to check for `netRevenue` on server responses, but I'm wondering if that's actually a peculiarity of your stack or only of your adapter. Should it be done for everyone? (@MaksymTeqBlaze ?)
 - your tests assumed that `bidderRequests.gdprConsent` would be a string, which is not the case. I trusted the shared utils in that your endpoint only expects to get `gdpr.consentString`.
 - I think what you were doing in `buildRequests` is splitting bids by partner, but otherwise following the same pattern that's in the shared utils for each partner.

MaksymTeqBlaze:
> * I modified the shared utils slightly to enable you to check for `netRevenue` on server responses, but I'm wondering if that's actually a peculiarity of your stack or only of your adapter. Should it be done for everyone? (@MaksymTeqBlaze ?)
I think the way you have done it is optimal; it allows us to add additional validation without possibly breaking other or older adapters.

PR #11860, opened 2024-06-25
Visx bid adapter: import utilities and retrieve data from user on ortb2

Goncakkd:
## Type of change

- [ ] Bugfix
- [x] Feature
- [ ] New bidder adapter  
- [ ] Updated bidder adapter   
- [ ] Code style update (formatting, local variables)
- [ ] Refactoring (no functional changes, no api changes)
- [ ] Build related changes
- [ ] CI related changes

- [ ] Does this change affect user-facing APIs or examples documented on http://prebid.org?
- [ ] Other

## Description of change
Retrieved and sent user data (for getting topics data)

PR #11877, opened 2024-06-27
Discovery Bid Adapter : add ext params

ecoeco163:
## Type of change

- [ ] Bugfix
- [x] Feature
- [ ] New bidder adapter  
- [ ] Updated bidder adapter   
- [ ] Build related changes
- [ ] CI related changes

- [ ] Other

## Description of change

add ext params


ecoeco163:
@ChrisHuie  Hi, discoveryBidAdapter and mediagoBidAdapter are maintained by different people. How can I solve these duplicate code tips? Thanks.

patmmccann:
Are they both Baidu employees? You can solve them similar to #11854 or #11860

patmmccann:
I see that these are both Baidu adapters. You definitely need to fix the duplication with imports. May i suggest /libraries/baiduUtils/index.js?

modules/discoveryBidAdapter.js
```diff
@@ -274,14 +228,14 @@ function transformSizes(requestSizes) {
 // Support sizes
 const popInAdSize = [
   { w: 300, h: 250 },
+  { w: 320, h: 50 },
   { w: 300, h: 600 },
   { w: 728, h: 90 },
+  { w: 320, h: 100 },
   { w: 970, h: 250 },
-  { w: 320, h: 50 },
+  { w: 336, h: 280 },
```
patmmccann, line 236: What you are doing here is obvious and will not be permitted. This is a malicious attempt to ignore our rules and fool our robot. This poor conduct. You must revert this commit and follow previous guidance.

patmmccann (changes requested):
We do not appreciate cookies designed to fool us

modules/discoveryBidAdapter.js
ecoeco163: Sorry, discovery and mediago have been separated and are no longer maintained together.
I can't contact the maintainer of Mediago Bid Adapter, so I updated our code in the hope that it can pass the robot check.
I can put the discovery code into the public utils, but I don't know whether mediago will use them.
Please tell me what I should do, thank you very much.🫶

patmmccann: please modify the mediago adapter as well, their collaboration isnt required

ecoeco163:
@patmmccann I have put the common method in the libraries, but the robot still warns that there is duplicate code. Please tell me what to do next. Thank you very much 🙏 🫶

patmmccann:
The robot is no longer mentioning your adapter nor mediavo, great work!

patmmccann (approved):

PR #11882, opened 2024-06-28
yandex ID: Use error-prone reference to crypto module

chernodub:
## Type of change

- [X] Bugfix
- [ ] Feature
- [ ] New bidder adapter  
- [ ] Updated bidder adapter   
- [ ] Build related changes
- [ ] CI related changes

- [ ] Other

## Description of change

There was an implicit issue for browsers that do not have `crypto` available globally, we would get a `ReferenceError` when trying to reference `crypto` without `window.` prefix. I've updated the code and changed the tests to reflect the changes.

### DIFF
--- a/modules/yandexIdSystem.js
+++ b/modules/yandexIdSystem.js
@@ -128,7 +128,7 @@ class YandexUidGenerator {
   }
 
   _getRandomGenerator() {
-    if (crypto) {
+    if (window.crypto) {
       return () => {
         const buffer = new Uint32Array(1);
         crypto.getRandomValues(buffer);
--- a/test/spec/modules/yandexIdSystem_spec.js
+++ b/test/spec/modules/yandexIdSystem_spec.js
@@ -110,12 +110,17 @@ describe('YandexId module', () => {
 
     describe('crypto', () => {
       it('uses Math.random when crypto is not available', () => {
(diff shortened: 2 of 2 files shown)


patmmccann:
Aside, you have a jsdoc issue

chernodub:
> Aside, you have a jsdoc issue flagged by circleci, can you include that fix here?

Hi @patmmccann. I don't see any issues reported. Could you provide more details on that, please?

Screenshot 2024-06-28 at 16 07 15

patmmccann:
from the circleci build output, 

image

chernodub:
I see, thanks @patmmccann. `@satisfies` is a TS-feature that extends available JSDoc tags.

Typescript: https://github.com/Microsoft/TypeScript/issues/51086
The support for tag tag was also released in eslint plugin recently: https://github.com/gajus/eslint-plugin-jsdoc/issues/1045 

Should I add the property to whitelist in eslint config, or maybe we could just update the plugin?

patmmccann:
Sounds great, take your pick

chernodub:
@patmmccann please take a look https://github.com/prebid/Prebid.js/pull/11909/files

I believe the mentioned PR is not a blocker for the current PR, so probably we could merge both separately. I tried the lint script with new plugin and yandex is OK now.

patmmccann (approved):

PR #11884, opened 2024-06-28
Linter checks: dom manipulation, network traffic, and direct storage access && bugfixes: unauthorized storage access

patmmccann:
related: https://github.com/prebid/Prebid.js/pull/11853 which should replace with storage manager

note: this isn't currently working, the linter isnt failing any files for any of these new checks, including relevate health, which is bypassing consent checks


spotxslagle (approved):
LGTM

PR #11901, opened 2024-07-01
[JW Player Video Provider][Video JS Video Provider]: Add missing typedefs

karimMourra:
## Type of change

## Description of change
Adds missing typedefs for types that are referenced to avoid linter warnings.
It would be nice if the types could be shared to reduce duplication

### DIFF
--- a/modules/videojsVideoProvider.js
+++ b/modules/videojsVideoProvider.js
@@ -13,6 +13,9 @@ import { submodule } from '../src/hook.js';
 import stateFactory from '../libraries/video/shared/state.js';
 import { PLAYBACK_MODE } from '../libraries/video/constants/constants.js';
 import { getEventHandler } from '../libraries/video/shared/eventHandler.js';
+/**
+ * @typedef {import('../libraries/video/shared/state.js').State} State
+ */
 
 /*
 Plugins of interest:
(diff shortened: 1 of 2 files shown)


patmmccann:
It is possible to import types, @chernodub did it here

https://github.com/prebid/Prebid.js/blob/master/src/types/ortb2.d.ts

patmmccann:
https://github.com/prebid/Prebid.js/pull/11902/files is also importing types

modules/videojsVideoProvider.js
```diff
@@ -13,9 +13,6 @@ import { submodule } from '../src/hook.js';
 import stateFactory from '../libraries/video/shared/state.js';
 import { PLAYBACK_MODE } from '../libraries/video/constants/constants.js';
 import { getEventHandler } from '../libraries/video/shared/eventHandler.js';
-/**
- * @typedef {import('../libraries/video/shared/state.js').State} State
```
patmmccann: Why revert this? looks like it worked

karimMourra: THe duplication check is buggy . I tried to see if reducing the amount of files changed could limit its scope, but that's not the case.

patmmccann:
Thanks!

patmmccann (approved):

PR #11905, opened 2024-07-02
Admixer Bid Adapter: change admixerwl to rtbstack alias

dariaboyko:
## Type of change

- [ ] Bugfix
- [ ] Feature
- [ ] New bidder adapter  
- [x] Updated bidder adapter   
- [ ] Build related changes
- [ ] CI related changes

- [ ] Other

## Description of change

Update admixerwl alias and rename it to rtbstack


modules/admixerBidAdapter.js
```diff
@@ -23,8 +23,8 @@ export const spec = {
    * Determines whether or not the given bid request is valid.
    */
   isBidRequestValid: function (bid) {
-    return bid.bidder === 'admixerwl'
-      ? !!bid.params.clientId && !!bid.params.endpointId
+    return bid.bidder === 'rtbstack'
+      ? !!bid.params.endpointId
```
patmmccann, line 27: So you have a bunch of active aliases but you only want one of them to pass validity checks,?

modules/admixerBidAdapter.js
```diff
@@ -13,7 +13,7 @@ const ALIASES = [
   {code: 'futureads', endpoint: 'https://ads.futureads.io/prebid.1.2.aspx'},
   {code: 'smn', endpoint: 'https://ads.smn.rs/prebid.1.2.aspx'},
   {code: 'admixeradx', endpoint: 'https://inv-nets.admixer.net/adxprebid.1.2.aspx'},
-  {code: 'admixerwl', endpoint: 'https://inv-nets-adxwl.admixer.com/adxwlprebid.aspx'},
```
patmmccann, line 16: This appears to break people

modules/admixerBidAdapter.js
dariaboyko: No, all aliases must have zone which is checked in the right part of operator, only rtbstack requires other params

modules/admixerBidAdapter.js
dariaboyko: do you mean renaming? if so than this alias isnt released yet on our side so it wont impact on customers.

patmmccann (approved):

PR #11926, opened 2024-07-04
CORE: check if the body is available before adding the locator iframe

olafbuitelaar:
## Type of change

- [x ] Bugfix

## Description of change
When the script is executed from the `<head>` and the `<body>` element isn't parsed by the browser yet, document.body is undefined causing the `processQue` to fail. the RAF should be executed only once the body is available. The other option would be to add the frame in the head, which should be fine for it's purposes.
image

### DIFF
--- a/src/adRendering.js
+++ b/src/adRendering.js
@@ -256,9 +256,7 @@ export function insertLocatorFrame() {
     const frame = createInvisibleIframe();
     frame.name = PB_LOCATOR;
     if (!document.body) {
-      window.requestAnimationFrame(() => {
-        document.body.appendChild(frame);
-      });
+      window.requestAnimationFrame(insertLocatorFrame);
     } else {
       document.body.appendChild(frame);
     }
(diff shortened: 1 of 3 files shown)


lksharma (approved):
LGTM

dgirardi (approved):

src/adRendering.js
```diff
@@ -253,8 +253,12 @@ export function renderAdDirect(doc, adId, options) {
  */
 export function insertLocatorFrame() {
   if (!window.frames[PB_LOCATOR]) {
-    const frame = createInvisibleIframe();
-    frame.name = PB_LOCATOR;
-    document.body.appendChild(frame);
+    if (!document.body) {
+      window.requestAnimationFrame(insertLocatorFrame);
```
muuki88, line 257: @olafbuitelaar on what assumption is this based, that this fixes the issue?

We use https://developer.mozilla.org/en-US/docs/Web/API/Document/DOMContentLoaded_event to check if the DOM is ready. Does `requestAnimationFrame` imply something similar?

muuki88: That makes sense. Thanks for the clarification

PR #11928, opened 2024-07-05
Core: Add Session Storage Manager & Contxtful RTD Provider: use session storage

sebastienrufiange:
## Type of change
- [ x ] Maintenance

## Description of change

- This PR adds sessionStorage functions in the Prebid Storage Manager
- The Contxtful RTD Provider leverages these to read from sessionStorage.

In our use case, we don't rely on localStorage because sessionStorage is more volatile and don't persist and suits perfectly as we don't want any kind of tracking.

We cannot simply override or extend the storage manager to add this functionality; so we propose to make it available in the storage manager.

## Other information
- RTD Provider Documentation does not need updating.

### DIFF
--- a/src/storageManager.js
+++ b/src/storageManager.js
@@ -179,6 +179,83 @@ export function newStorageManager({moduleName, moduleType} = {}, {isAllowed = is
   }
 
   /**
+   * @returns {boolean}
+   */
+  const sessionStorageIsEnabled = function (done) {
+    let cb = function (result) {
+      if (result && result.valid) {
+        try {
+          sessionStorage.setItem('prebid.cookieTest', '1');
+          return sessionStorage.getItem('prebid.cookieTest') === '1';
+        } catch (error) {
+        } finally {
+          try {
+            sessionStorage.removeItem('prebid.cookieTest');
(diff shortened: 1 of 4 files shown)


src/storageManager.js
```diff
@@ -178,6 +178,83 @@ export function newStorageManager({moduleName, moduleType} = {}, {isAllowed = is
     return schedule(cb, STORAGE_TYPE_LOCALSTORAGE, done);
   }
 
+  /**
+   * @returns {boolean}
+   */
+  const sessionStorageIsEnabled = function (done) {
```
dgirardi, line 184: These are a copy-paste of the local storage methods. I don't have permission to update this PR, but you can find my refactor [here](https://github.com/prebid/Prebid.js/commit/04785e9b0122c4990f108fdb143df246b0732277) (also on the `11928-refactor` branch).

sebastienrufiange: Many thanks @dgirardi!
That’s a nice refactoring. I also wondered the same thing but wasn’t sure.
Thanks for doing the branch; we just merged it.

patmmccann (approved):

PR #11931, opened 2024-07-05
Relevate Health Bid Adapter: Use Storage Manager

hasanideepak:
## Type of change

- [x] Bugfix
- [ ] Feature
- [ ] New bidder adapter  
- [ ] Updated bidder adapter   
- [ ] Build related changes
- [ ] CI related changes

- [ ] Other


hasanideepak:
@patmmccann As agreed, I have used the storage manager instead of directly using localStorage. Please review. Thanks

patmmccann:
You must stop doing things like https://github.com/prebid/Prebid.js/pull/11931/commits/a34dd3ddf119338b322577338b6c68b205a36ebe. This is a conduct issue. Revert that commit.

hasanideepak:
> You must stop doing things like [a34dd3d](https://github.com/prebid/Prebid.js/commit/a34dd3ddf119338b322577338b6c68b205a36ebe). This is a conduct issue. Revert that commit and any other changes that are designed to fool our checks and make our code worse.

I agree, but as you know these changes made to avoid duplicate code. How would I avoid it, if our adapter have some common lines with another adapter. We cant use the common library. Please guide.

patmmccann:
What is your concern with importing functions? 

It is better to let the robot complain than to change the code to give from it

PR #11932, opened 2024-07-05
Smaato Bid Adapter: Add UserSyncs

el-chuck:
## Type of change

- [x] Updated bidder adapter   

## Description of change

Adding implementation of getUserSyncs() for Smaato adapter

### DIFF
--- a/modules/smaatoBidAdapter.js
+++ b/modules/smaatoBidAdapter.js
@@ -19,10 +19,11 @@ import {ortbConverter} from '../libraries/ortbConverter/converter.js';
 
 const BIDDER_CODE = 'smaato';
 const SMAATO_ENDPOINT = 'https://prebid.ad.smaato.net/oapi/prebid';
-const SMAATO_CLIENT = 'prebid_js_$prebid.version$_3.1'
+const SMAATO_CLIENT = 'prebid_js_$prebid.version$_3.2'
 const TTL = 300;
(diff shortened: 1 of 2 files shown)


modules/smaatoBidAdapter.js
```diff
@@ -196,6 +197,22 @@ export const spec = {
    * @return {UserSync[]} The user syncs which should be dropped.
    */
   getUserSyncs: (syncOptions, serverResponses, gdprConsent, uspConsent) => {
+    if (syncOptions && syncOptions.pixelEnabled) {
```
patmmccann, line 200: you forgot about gpp? Also, why not import https://github.com/prebid/Prebid.js/blob/8f817bd187de3a1bde065f85eb5b6e0a6b034b66/libraries/teqblazeUtils/bidderUtils.js#L206 ?

el-chuck: the changes are sufficient for now, but thank you for checking.

regarding the generic implementation: both options currently don't fit our use case. if those should be used, i can of course rework one of them to make it fit. is there some preference which one should be worked on for a generic implementation?

patmmccann: No preference, I'll merge this and keep we'll study some more adapters before starting a generic implementation

patmmccann (approved):

PR #11942, opened 2024-07-08
Adkernel Bid Adapter: add voisetech alias

ckbo3hrk:
## Type of change
- [x] Other

## Description of change
Added alias for voisetech.com partner network

### DIFF
--- a/modules/adkernelBidAdapter.js
+++ b/modules/adkernelBidAdapter.js
@@ -96,7 +96,8 @@ export const spec = {
     {code: 'digiad'},
     {code: 'monetix'},
     {code: 'hyperbrainz'},
-    {code: 'globalsun'}
+    {code: 'globalsun'},
+    {code: 'voisetech'}
   ],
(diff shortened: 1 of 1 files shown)


patmmccann:
Hi there!, looks like there's an easy opportunity to import `

function getBidFloor(bid, mediaType, sizes) {
  var floor;
  var size = sizes.length === 1 ? sizes[0] : '*';
  if (typeof bid.getFloor === 'function') {
    const floorInfo = bid.getFloor({currency: 'USD', mediaType, size});
    if (typeof floorInfo === 'object' && floorInfo.currency === 'USD' && !isNaN(parseFloat(floorInfo.floor))) {
      floor = parseFloat(floorInfo.floor);
    }
  }
  return floor;
}` in two of your adapters

ckbo3hrk:
@patmmccann Extracted `getBidFloor` function into a utility file.

patmmccann (approved):

PR #11944, opened 2024-07-08
EightPod Bid Adapter + EightPod Analytic Adapter - Support multiple adUnit, updated event tracking, added UserId support

GreDiSe:
Support multiple adUnit, updated event tracking, added UserId support

### DIFF
--- a/modules/eightPodAnalyticsAdapter.js
+++ b/modules/eightPodAnalyticsAdapter.js
@@ -14,7 +14,6 @@ const MODULE = `${MODULE_NAME}AnalyticProvider`;
  * Custom tracking server that gets internal events from EightPod's ad unit
  */
 const trackerUrl = 'https://demo.8pod.com/tracker/track';
-const context = {};
 export const storage = getStorageManager({moduleType: MODULE_TYPE_ANALYTICS, moduleName: MODULE_NAME})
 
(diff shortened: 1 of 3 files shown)


modules/eightPodAnalyticsAdapter.js
```diff
@@ -176,24 +173,15 @@ eightPodAnalytics.eventsStorage = [];
...
-  };
-})(eightPodAnalytics.disableAnalytics);
-
 /**
  * Register Analytics Adapter
  */
 adapterManager.registerAnalyticsAdapter({
   adapter: eightPodAnalytics,
-  code: MODULE_NAME
+  code: 'eightPod'
```
patmmccann, line 184: on your ajax call, the keepalive method is much more polite; you may want to you it to get more events after the page dies.

GreDiSe:
Doc: https://github.com/prebid/prebid.github.io/pull/5493

modules/eightPodAnalyticsAdapter.js
GreDiSe: Thank you for great advice. I added keepalive.

Also I fixed my test, but they failed with another reason now

patmmccann (approved):

Issue #10997, opened 2024-01-29
Targeting: set pps to GAM display


patmmccann:
## Type of issue
Feature

## Description
in #10977 (closing #10687) we set PPS Json for video for IAB segtax 4 (users) and 6 (content). Turns out there is a new gpt.js function as well

`
 googletag.setConfig({
   pps: {
     taxonomies: {
      'IAB_AUDIENCE_1_1': 
        {values: ['12', '13', '14']},
      'IAB_CONTENT_2_2':
        {values: ['22', '25', '57']},
    }
  }
 });
`

Subsequent calls with new data should include all previous data to resurface, as they overwrite, not append
`
 googletag.setConfig({
   pps: {
     taxonomies: {
      'IAB_AUDIENCE_1_1': 
        {values: ['12', '13', '14','4', '5', '6']},
      'IAB_CONTENT_2_2':
        {values: ['22', '25', '57']},
    }
  }
 });
`

Should we provide a single API call similar to setTargeting, set PPID, or handoff auction configs; perhaps setAdServerFPD(), or should we just always set it? I can't think of a reason people wouldn't want to set it, so I would be in favor of always setting it and perhaps offering a do not set config flag.

patmmccann:
Some quirks: setAdServerFPD seems preferred so we don't introduce a dependency on GPT and break parallel loading of gpt and auctions. For us, FPD is an auction property, should we use the latest auction? Should we take an ad unit as an argument or an auction id? Could we offer all three depending on the arguments? () -> use latest, (auctionId =) -> use latest on the auctionid, (adunitcode = ), look up latest auction on this adunit

muuki88:
Related resources

- https://support.google.com/admanager/answer/12451124
- https://developers.google.com/publisher-tag/release-notes?hl=de#2024-02-12

dgirardi:
It's not clear how this should be done:

- GPT accepts a global `setConfig`, but in Prebid taxonomy data is potentially different for each bid. Which one do we pick? (different bids can win concurrent auctions)
- Does the global `setConfig` also set taxonomy data for video?

patmmccann:
>GPT accepts a global setConfig, but in Prebid taxonomy data is potentially different for each bid. Which one do we pick? (different bids can win concurrent auctions)

when setTargeting is called for mulitple ad units, if they have conflicting site.content.data or user.data, abort setting pps (pub provided segments). 

>Does the global setConfig also set taxonomy data for video?

[...]

PR #11946, opened 2024-07-08
gptPreAuction: pass publisher provided signals to GPT

mkomorski:
## Type of change

- [ ] Bugfix
- [x] Feature
- [ ] New bidder adapter  
- [ ] Updated bidder adapter   
- [ ] Build related changes
- [ ] CI related changes

- [ ] Other

## Other information

#10997


src/pps.js
```diff
@@ -0,0 +1,58 @@
...
+  const signals = auctionIds
+    .map(auctionId => auctionManager.index.getAuction({ auctionId })?.getFPD()?.global)
+    .map(getSignals)
+    .filter(fpd => fpd);
+
+  return signals;
+}
+
+export function getSignalsIntersection(signals) {
+  const taxonomies = ['IAB_AUDIENCE_1_1', 'IAB_CONTENT_2_2'];
```
dgirardi, line 35: these can be extracted to be reused with `getSignals` above

src/pps.js
```diff
@@ -0,0 +1,58 @@
...
+    ) : []
+    result[taxonomy] = { values: result[taxonomy] };
+  })
+  return result;
+}
+
+export function getAuctionsIdsFromTargeting(targeting) {
+  return Object.values(targeting)
+    .flatMap(x => Object.entries(x))
+    .filter((entry) => (entry[0] || '').includes(TARGETING_KEYS.AD_ID))
```
dgirardi, line 55: I think this should be `entry[0] === TARGETING_KEYS.AD_ID || entry[0].startsWith(TARGETING_KEYS.AD_ID + '_')`

That's because
1. I don't think object keys can be falsy (afaik the language requires them to be either strings or symbols), so you don't need the fallback to `''`
2. You are interested in either `hb_adid` or `hb_adid_${bidder}` ([reference](https://github.com/prebid/Prebid.js/blob/e8964d4affcdadede55489d9b02e2a29e89422b8/src/targeting.js#L702-L708
)). It's unlikely, but possible, to define custom targeting keys like `custom_hb_adid` that would be a false positive for this.

src/pps.js
```diff
@@ -0,0 +1,58 @@
...
+  const signals = Object.entries({
+    IAB_AUDIENCE_1_1: getSegments(fpd, ['user.data'], 4),
+    IAB_CONTENT_2_2: getSegments(fpd, CLIENT_SECTIONS.map(section => `${section}.content.data`), 6)
+  }).map(([taxonomy, values]) => values.length ? {taxonomy, values} : null)
+    .filter(ob => ob);
+
+  return signals;
+}
+
+export function getSignalsArrayByAuctionsIds(auctionIds) {
```
dgirardi, line 25: (regarding the test failure) my preferred method of stubbing is dependency injection. E.g. if this did something like `getSignalsArrayByAuctionIds(auctionIds, index = auctionManager.index)`, tests can swap in a mock version of `index`, and I find it a bit easier to reason about compared to sinon's stubs. To do the same with sinon I think you need an extra layer of indirection, basically instead of making `index` a parameter make it a property of some global object pointing to `auctionManager.index`, which you can then stub.

src/pps.js
```diff
@@ -0,0 +1,58 @@
+import { auctionManager } from './auctionManager.js';
```
dgirardi, line 1: since this is quite a bit of code I'm wondering if it should be in a module (maybe gptPreAuction) - what do you think @patmmccann ?

src/pps.js
```diff
@@ -0,0 +1,58 @@
...
+  })
+  return result;
+}
+
+export function getAuctionsIdsFromTargeting(targeting) {
+  return Object.values(targeting)
+    .flatMap(x => Object.entries(x))
+    .filter((entry) => (entry[0] || '').includes(TARGETING_KEYS.AD_ID))
+    .flatMap(entry => entry[1])
+    .filter(uniques);
```
dgirardi, line 57: note that these are ad ids, not auction IDs. they should be passed through [`findBidByAdId`](https://github.com/prebid/Prebid.js/blob/e8964d4affcdadede55489d9b02e2a29e89422b8/src/auctionManager.js#L131).

dgirardi (approved):

patmmccann (approved):

aanton:
Hi @patmmccann & @dgirardi,

I'm a little confused about the implementation of this functionality. It has been added to the [GPT Pre-Auction module](https://docs.prebid.org/dev-docs/modules/gpt-pre-auction.html) but its documentation has not be updated.

I thought the main responsibility of this module was to enrich the bid requests with the GPID using the GAM adunit. However, it now also enriches the GAM request with the publisher provided signals (PPS).

Best,
Armando

Issue #11934, opened 2024-07-05
Prebid 8.x: backport Prebid 9: paapiForGpt: add support for customSlotMatching,


patmmccann:
There is rather strong demand for custom slot matching for PAAPI to be in 8.52.1 

https://github.com/prebid/Prebid.js/pull/11714

cc @FlorentDancy

PR #11957, opened 2024-07-10
Core & fledgeForGpt: backport customSlotMatching

dgirardi:
## Type of change

- [x] Feature

## Description of change

Partial backport of https://github.com/prebid/Prebid.js/issues/11934. Adds support for `customSlotMatching`, but since there's no way to pass it to the legacy `autoconfig` behavior, it's only supported when used directly with `setPAAPIConfigForGpt`.

Closes https://github.com/prebid/Prebid.js/issues/11934


JulieLorin (approved):
LGTM
Thanks for the backport

PR #11965, opened 2024-07-12
ID5 UserId module : use userId storage mechanism to store request number

abazylewicz-id5:
## Type of change

## Description of change

To better adhere to Prebid user configuration (mentioned in https://github.com/prebid/Prebid.js/issues/10710) id5id module no longer stores any data on itself, and will only use the storage provided by Prebid userId module

## Other information

Docs PR: https://github.com/prebid/prebid.github.io/pull/5498

### DIFF
--- a/modules/id5IdSystem.js
+++ b/modules/id5IdSystem.js
@@ -13,14 +13,13 @@ import {
   isPlainObject,
   logError,
   logInfo,
-  logWarn,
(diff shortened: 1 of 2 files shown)


Rothalack (commented):
Looks good, working in testing for me.

Rothalack (approved):

PR #11969, opened 2024-07-12
uid2IdSystem_shared.js : better logging messages

patmmccann:
prepend uid2 system to log messages


patmmccann:
The circleci output seems to demonstrate after this PR that this module is responsible for much of the message spam about http 500 type network errors

ChrisHuie (approved):

PR #11974, opened 2024-07-15
videoHeroesAdapter: support eids

videoheroes:
## Type of change

- [ ] Bugfix
- [x] Feature
- [ ] New bidder adapter  
- [x] Updated bidder adapter   
- [ ] Build related changes
- [ ] CI related changes

- [ ] Other

## Description of change
Added support of eids array to provide our exchange servers with user data from user submodules

### DIFF
--- a/modules/videoheroesBidAdapter.js
+++ b/modules/videoheroesBidAdapter.js
@@ -41,8 +41,8 @@ export const spec = {
       device: bidderRequest.ortb2?.device || prepareDevice(),
       tmax: bidderRequest.timeout,
       cur: [ CUR ],
-      regs: { ext: {}, coppa: config.getConfig('coppa') == true ? 1 : 0 },
       source: { ext: { schain: validBidRequests[0].schain } },
+      regs: { ext: {}, coppa: config.getConfig('coppa') == true ? 1 : 0 },
       user: { ext: {} }
     };
 
--- a/modules/videoheroesBidAdapter.js
+++ b/modules/videoheroesBidAdapter.js
@@ -47,11 +47,11 @@ export const spec = {
 
     validBidRequests = convertOrtbRequestToProprietaryNative(validBidRequests);
 
-    let data = {
+    let data = {̉̉
       id: bidderRequest.bidderRequestId,
       imp: validBidRequests.map(adUnit => prepareImpression(adUnit)),
       site: prepareSite(validBidRequests[0], bidderRequest),
-      device: bidderRequest.ortb2?.device || {},
+      device: bidderRequest.ortb2?.device || prepareDevice(),
       tmax: bidderRequest.timeout,
       cur: [ CUR ],
       user: { ext: {} },
--- a/modules/videoheroesBidAdapter.js
+++ b/modules/videoheroesBidAdapter.js
@@ -54,8 +54,8 @@ export const spec = {
       device: bidderRequest.ortb2?.device || {},
       tmax: bidderRequest.timeout,
       cur: [ CUR ],
-      user: { ext: {} },
       regs: { ext: {}, coppa: config.getConfig('coppa') == true ? 1 : 0 },
+      user: { ext: {} },
       source: { ext: { schain: validBidRequests[0].schain } }
     };
 
--- a/modules/videoheroesBidAdapter.js
+++ b/modules/videoheroesBidAdapter.js
@@ -121,7 +121,7 @@ const getMediaTypeValues = {
       [w, h] = adUnit.mediaTypes.banner.sizes;
     } else if (isArray(adUnit.mediaTypes.banner.sizes)) {
       [w, h] = adUnit.mediaTypes.banner.sizes[0];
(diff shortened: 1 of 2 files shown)


patmmccann:
Hi, it appears your adapter and the brave adapter are duplicates. Please solve before working in either one further

videoheroes:
Hi guys ! Ok we will change code to avoid duplicates, its different companies, but started by same team and code

patmmccann:
Great! Can we lend a hand in any way. One example is #11951 another is #11877

modules/videoheroesBidAdapter.js
```diff
@@ -1,261 +1,286 @@
...
-    if (Array.isArray(br.mediaTypes.banner.sizes[0])) { size = br.mediaTypes.banner.sizes[0]; } else { size = br.mediaTypes.banner.sizes; }
-  } else size = [300, 250];
+  [deviceObj.w, deviceObj.h] = [screen.width, screen.height];
+  deviceObj.language = navigator.language;
+  deviceObj.dnt = navigator.doNotTrack === '1' ? 1 : 0
+  deviceObj.ua = navigator.userAgent;
 
-  return { id: br.transactionId, w: size[0], h: size[1] };
-};
+  return deviceObj;
```
patmmccann, line 236: why not just take the device object core has prepared for you on bidderRequest.ortb2.device? eg https://github.com/prebid/Prebid.js/pull/12053/files

videoheroes: never used it before on any prebid connectors, interesting, so if it can replace all device regular data, lets update it

videoheroes:
seems like  bidderRequest.ortb2 not stable and guarantee to be available, but i do not include old device prepare for else  cases - bidderRequest.ortb2?.device || {}

videoheroes:
I do not why, but there all test random browser testing errors after building

videoheroes:
Still different strange errors after added ortb2

modules/videoheroesBidAdapter.js
```diff
@@ -1,261 +1,286 @@
...
+const ENDPOINT_URL = `https://point.contextualadv.com/?t=2&partner=`;
+
+const NATIVE_ASSETS_IDS = { 1: 'title', 2: 'icon', 3: 'image', 4: 'sponsoredBy', 5: 'body', 6: 'rating', 7: 'downloads', 8: 'cta' };
+
+const NATIVE_ASSETS_TYPES = {
+  title: {id: 1, title: {len: 25}},
+  icon: {id: 2, img: {type: 1}},
+  image: {id: 3, img: {type: 3}},
+  sponsoredBy: {id: 4, data: {type: 1, len: 30}},
+  body: {id: 5, data: {type: 2, len: 100}},
```
patmmccann, line 21: It doesn't seem like you've fixed things, instead you've refactored to fool the robot. The opportunity to make the code efficient is just hidden by this commit

patmmccann (changes requested):
Please remove all your commits designed to hide your code duplication from the robot

modules/videoheroesBidAdapter.js
videoheroes: its not duplication, but it developed by same manner, does it better if i will use switch or if instead ob for per keys ?

videoheroes: maybe problem because I developed these connectors before and I still see realisation like that ... ok anyway

videoheroes:
> Please remove all your commits designed to hide your code duplication from the robot

videoheroes:
I will develop Native req / res convertor on another way, as you do not like i am using object to convert it, which is still the best variant I guess

videoheroes:
Has been updated as you asked for, now native full from scratched - without using convert object and without looking for old one, can not say I like this change, but its another way to explain logic compatible with our exchange systems and old clients

videoheroes:
About commits with changing line - I made it to rebuild projects, as there all time errors not about adapter, more situate like some adapters servers results, etc
1) calls addPaapiConfig when there is no bid in the response
     S2S Adapter response handler when the response contains ext.prebid.fledge
     AssertError: expected stub to be called with arguments 
    at Object.fail (node_modules/sinon/pkg/sinon.js:142:21)
    at failAssertion (node_modules/sinon/pkg/sinon.js:101:16)
    at assert.<computed> [as calledWith] (node_modules/sinon/pkg/sinon.js:126:13)
    at expectFledgeCalls (/tmp/_karma_webpack_716925/commons.js:395628:22)
    at Context.<anonymous> (/tmp/_karma_webpack_716925/commons.js:395659:9)

2) Logger: post-timeout check without bid response
     pubmatic analytics adapter when handling events
     AssertionError: expected 2 to equal 1
    at Context.<anonymous> (/tmp/_karma_webpack_716925/commons.js:402825:34)

3) should handle case when callback return falsy value
     weboramaRtdProvider Handle Set Targeting and Bid Request Add site-centric data (contextual)
     AssertionError: expected 3 to equal +0
    at Context.<anonymous> (/tmp/_karma_webpack_716925/commons.js:488326:48)

PR #11980, opened 2024-07-16
Insticator: Update impression ext for prebid-server params 

shubhamc-ins:


modules/insticatorBidAdapter.js
```diff
@@ -17,17 +17,17 @@ export const OPTIONAL_VIDEO_PARAMS = {
...
-  'skipafter': (value) => isInteger(value),
-  'sequence': (value) => isInteger(value),
+  'linearity': (value) => isInteger(value) && [1].includes(value),
   'battr': (value) => isArrayOfNums(value),
+  'skipafter': (value) => isInteger(value),
+  'skip': (value) => isInteger(value) && [1, 0].includes(value),
   'maxextended': (value) => isInteger(value),
-  'minbitrate': (value) => isInteger(value),
   'maxbitrate': (value) => isInteger(value),
-  'playbackmethod': (value) => isArrayOfNums(value),
```
patmmccann, line 29: Hi @shubhamc-ins this commit is poor conduct and will not be tolerated

Please resubmit

patmmccann (changes requested):
Nonsense commits to hide from the duplication detector are poor conduct

patmmccann:
Also take a look at #11970 which is working on a real solution to the duplicated code

modules/insticatorBidAdapter.js
shubhamc-ins: @patmmccann apologies, I will resubmit.

PR #11981, opened 2024-07-16
kimberliteBidAdapter: video media type support

os-solta:
## Type of change
- [ ] Bugfix
- [x] Feature
- [ ] New bidder adapter  
- [x] Updated bidder adapter https://github.com/prebid/prebid.github.io/pull/5504
- [ ] Build related changes
- [ ] CI related changes

- [ ] Other

## Description of change
Added video media type to kimberliteBidAdapter


os-solta:
Hi! I'm not sure about the reason of CI build fail. Could someone help me with the reason please?

patmmccann:
Just a flaky test, no worries

modules/kimberliteBidAdapter.js
```diff
@@ -34,19 +34,40 @@ const converter = ortbConverter({
   imp (buildImp, bidRequest, context) {
     const imp = buildImp(bidRequest, context);
     imp.tagid = bidRequest.params.placementId;
+    const mediaTypes = bidRequest.mediaTypes;
+    if (imp[VIDEO]) {
+      deepSetValue(imp, 'video.linearity', mediaTypes[VIDEO].linearity || 1);
+      deepSetValue(imp, 'video.protocols', mediaTypes[VIDEO].protocols || [3, 6]);
+      deepSetValue(imp, 'video.placement', mediaTypes[VIDEO].placement || 1);
```
patmmccann, line 41: I think the ortb convertor gives you all these for free, no need to copy them. Is that correct @dgirardi ?

dgirardi: I think this is to set default values, which the library does not provide.

patmmccann: Ah I see @os-solta these defaults aren't good guesses, you're better off requiring the publisher make a choice

patmmccann (changes requested):
Video defaults on placement and startdelay

modules/kimberliteBidAdapter.js
os-solta: Looks like i'll just skip these defaults here, thanks.

patmmccann (approved):

PR #11982, opened 2024-07-16
Insticator: Update prebid-server bidder params for impressions

shubhamc-ins:
## Type of change

- [ ] Bugfix
- [ ] Feature
- [ ] New bidder adapter  
- [ ] Updated bidder adapter   
- [ ] Build related changes
- [ ] CI related changes

- [ ] Other

### DIFF
--- a/modules/insticatorBidAdapter.js
+++ b/modules/insticatorBidAdapter.js
@@ -107,19 +107,8 @@ function buildVideo(bidRequest) {
   const playerSize = deepAccess(bidRequest, 'mediaTypes.video.playerSize');
   const context = deepAccess(bidRequest, 'mediaTypes.video.context');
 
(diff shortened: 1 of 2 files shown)


patmmccann:
My suggestion is let the robot complain, when the other pr merges, should be ready to address. Is that the same as your proposal?

shubhamc-ins:
> My suggestion is let the robot complain for now, when the other pr merges, should be ready to address. Is that the same as your proposal?

Okay @patmmccann , In that case This PR is ready. I'll address the duplication when https://github.com/prebid/Prebid.js/pull/11970 PR merges, sounds right?

Please feel free to tag me once that is done.

patmmccann (approved):

PR #11988, opened 2024-07-17
RavelTech Bid Adapter: 1st submission of new adapter

mnguyen-raveltech:
- [X] New bidder adapter

## Description of change
Ravel Technologies developed ZKAD (Zero-Knowledge ADvertising), a protocol anonymizing RTB data & PII during the RTB bidding process and during data processing. ZKAD is using a scalable proprietary Homomorphic Encryption scheme (Ravel Homormorphic Encryption, RHE), a form of encryption that allows to perform computations at scale on encrypted data without having to first decrypt it. ZKAD allows publishers to protect users’ privacy and publishers’ data while allowing advanced targeted advertising and advanced regulatory compliance.

Ravel Bidder is a “Privacy Bus” allowing to anonymize bid requests before forwarding them to SSP and DSP preventing any personal data to be transferred to SSP and DSP. PII IDs are removed or/and anonymized into RIDs (Ravelized IDs). Furthermore, User-Agent is truncated, and location is provided with a significant Radius. IP addresses are not stored by the Privacy Bus and are not provided in the bid requests.

The RavelTech Prebid Adapter first version supports only Xandr/Appnexus SSP and should be configured with your current Xandr/Appnexus existing params. For more information about the Xandr/AppNexus params can be found at [https://github.com/prebid/prebid.github.io/blob/master/dev-docs/bidders/appnexus.md]

Please contact [support@raveltech.io](mailto:support@raveltech.io) to activate your adapter after installation or for more information.

- contact email of the adapter’s maintainer: [michel.nguyen@raveltech.io] (will be replaced by a group email address later)
```
{
    'bidder': 'raveltech',
    'params': {
      'placementId': 234234
    },
    'userId': {
      'id5id': {
        'uid': '0',
        'ext': {
          'linkType': 0,
          'pba': 'K2ogG7aaimJB4/PSEuwADQ=='
        }
      },
      'pubProvidedId': [
        {
          'source': 'adnxs.com',
          'uids': [
            {
              'id': 'webo-id-1',
              'atype': 1,
              'ext': {
                'stype': 'ppuid'
              }
            }
          ]
        },
        {
          'source': 'adnxs.com',
          'uids': [
            {
              'id': 'webo-id-2',
              'atype': 1,
              'ext': {
                'stype': 'ppuid'
              }
            }
          ]
        },
        null
      ]
    },
    'userIdAsEids': [
      {
        'source': 'id5-sync.com',
        'uids': [
          {
            'id': '0',
            'atype': 1,
            'ext': {
              'linkType': 0,
              'pba': 'K2ogG7aaimJB4/PSEuwADQ=='
            }
          }
        ]
      },
      {
        'source': 'adnxs.com',
        'uids': [
          {
            'id': 'webo-id-1',
            'atype': 1,
            'ext': {
              'stype': 'ppuid'
            }
          },
          {
            'id': 'webo-id-2',
            'atype': 1,
            'ext': {
              'stype': 'ppuid'
            }
          }
        ]
      }
    ],
    'ortb2Imp': {
      'ext': {}
    },
    'mediaTypes': {
      'banner': {
        'sizes': [
          [300, 250],
          [300, 600],
          [728, 90]
        ]
      }
    },
    'adUnitCode': 'test-div',
    'transactionId': null,
    'adUnitId': 'd6f8ff69-1336-41c3-a639-20b76c6e0be1',
    'sizes': [
      [300, 250],
      [300, 600],
      [728, 90]
    ],
    'bidId': '3da7e5be5cc74',
    'bidderRequestId': '22077672ce0e1f',
    'auctionId': null,
    'src': 'client',
    'bidRequestsCount': 1,
    'bidderRequestsCount': 1,
    'bidderWinsCount': 0
  }
```

### DIFF
--- a/modules/raveltechBidAdapter.js
+++ b/modules/raveltechBidAdapter.js
@@ -46,6 +46,42 @@ export const spec = {
 
     logInfo(anonymizedBidRequests);
     return anonymizedBidRequests;
+  },
+
+  /**
+   * Determines whether or not the given bid request is valid.
+   *
+   * @param {object} bid The bid to validate.
+   * @return boolean True if this is a valid bid, and false otherwise.
+   */
+  isBidRequestValid: function (bid) {
+    if (!baseAdapter.isBidRequestValid) { return false; }
+    return baseAdapter.isBidRequestValid(bid);
+  },
+
+  /**
+   * Unpack the response from the server into a list of bids.
+   *
+   * @param {*} serverResponse A successful response from the server.
+   * @return {Bid[]} An array of bids which were nested inside the server.
+   */
+  interpretResponse: function (serverResponse, params) {
+    if (!baseAdapter.interpretResponse) { return []; }
+    return baseAdapter.interpretResponse(serverResponse, params);
+  },
+
+  getUserSyncs: function (syncOptions, responses, gdprConsent) {
+    if (!baseAdapter.getUserSyncs) { return []; }
+    return baseAdapter.getUserSyncs(syncOptions, responses, gdprConsent);
+  },
+
+  /**
+   * Add element selector to javascript tracker to improve native viewability
+   * @param {Bid} bid
+   */
+  onBidWon: function (bid) {
+    if (!baseAdapter.onBidWon) { return; }
+    baseAdapter.onBidWon(bid);
   }
 };
 
--- a/modules/raveltechBidAdapter.js
+++ b/modules/raveltechBidAdapter.js
@@ -22,7 +22,6 @@ export const spec = {
 
     let anonymizedBidRequests = baseAdapter.buildRequests(bidRequests, bidderRequest); // call the bid requests from the Appnexus adapter
     if (!anonymizedBidRequests) { return []; } // if no bid request, return empty Array
-
     if (!Array.isArray(anonymizedBidRequests)) { anonymizedBidRequests = [anonymizedBidRequests]; } // if only 1 bid request, anonymizedBidRequest will be an Object instead of an Array. Build Array with 1 bid request.
 
     // Load ZKAD runtime, used to anonymize the uids
@@ -40,11 +39,10 @@ export const spec = {
         if (!eid || !eid.id) { return; }
         logInfo('eid.source=', eid.source);
         eid.id = ZKAD.anonymizeID(eid.id, eid.source);
-        logInfo('Anonymized uid.id=', eid.id, 'as byte array of length=', eid.id.length);
+        logInfo('Anonymized as byte array of length=', eid.id.length);
       });
     });
 
-    logInfo(anonymizedBidRequests);
     return anonymizedBidRequests;
   },
 
--- a/modules/raveltechBidAdapter.js
+++ b/modules/raveltechBidAdapter.js
@@ -68,20 +68,6 @@ export const spec = {
   interpretResponse: function (serverResponse, params) {
     if (!baseAdapter.interpretResponse) { return []; }
     return baseAdapter.interpretResponse(serverResponse, params);
-  },
-
-  getUserSyncs: function (syncOptions, responses, gdprConsent) {
-    if (!baseAdapter.getUserSyncs) { return []; }
-    return baseAdapter.getUserSyncs(syncOptions, responses, gdprConsent);
-  },
-
-  /**
-   * Add element selector to javascript tracker to improve native viewability
-   * @param {Bid} bid
-   */
-  onBidWon: function (bid) {
-    if (!baseAdapter.onBidWon) { return; }
-    baseAdapter.onBidWon(bid);
   }
 };
 
--- a/modules/raveltechBidAdapter.js
+++ b/modules/raveltechBidAdapter.js
@@ -1,0 +1,88 @@
+// Import the base adapter
+import { spec as baseAdapter } from './appnexusBidAdapter.js'; // eslint-disable-line prebid/validate-imports
+import { registerBidder } from '../src/adapters/bidderFactory.js';
+import { logInfo } from '../src/utils.js';
+
+const BIDDER_CODE = 'raveltech';
+const URL = 'https://pb1.rvlproxy.net/bid/bid';
+// const URL_SIMPLE = 'https://pb1.rvlproxy.net/bid/simplebid';
+
+export const spec = {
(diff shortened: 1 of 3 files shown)


patmmccann:
couple of things: 

/home/circleci/Prebid.js/modules/raveltechBidAdapter.js
  17:1  warning  The type 'BidRequest' is undefined  jsdoc/no-undefined-types
  64:1  warning  The type 'Bid' is undefined         jsdoc/no-undefined-types

I think you can import these types from bidFactory.

It seems you have done a trick to prevent duplication with bidRichMedia. Instead you should move the functions into imports in your pr for both module files. 

Finally, why are you importing the appnexus adapter and disabling the linting rule?

modules/raveltechBidAdapter.js
```diff
@@ -0,0 +1,72 @@
...
+import { registerBidder } from '../src/adapters/bidderFactory.js';
+import { logInfo } from '../src/utils.js';
+
+const BIDDER_CODE = 'raveltech';
+const URL = 'https://pb1.rvlproxy.net/bid/bid';
+// const URL_SIMPLE = 'https://pb1.rvlproxy.net/bid/simplebid';
+
+export const spec = {
+  code: BIDDER_CODE,
+  gvlid: baseAdapter.GVLID, // use base adapter gvlid
```
patmmccann, line 12: you need written authorization from @andreiistephens for this

modules/raveltechBidAdapter.js
```diff
@@ -0,0 +1,72 @@
...
+   * @return ServerRequest[] Info describing the requests to the server.
+   */
+  buildRequests: function(bidRequests, bidderRequest) {
+    if (!baseAdapter.buildRequests) { return []; }
+
+    let anonymizedBidRequests = baseAdapter.buildRequests(bidRequests, bidderRequest); // call the bid requests from the Appnexus adapter
+    if (!anonymizedBidRequests) { return []; } // if no bid request, return empty Array
+    if (!Array.isArray(anonymizedBidRequests)) { anonymizedBidRequests = [anonymizedBidRequests]; } // if only 1 bid request, anonymizedBidRequest will be an Object instead of an Array. Build Array with 1 bid request.
+
+    // Load ZKAD runtime, used to anonymize the uids
```
patmmccann, line 27: if you are already loading js, why not just have your js call [registerBidAdapter ](https://github.com/prebid/Prebid.js/blob/40b906d17bcfda2b369924de2278ced73975edc2/src/prebid.js#L724) ?

patmmccann (commented):
several questions

patmmccann:
one more comment: 

> The RavelTech Prebid Adapter first version supports only Xandr/Appnexus SSP 

given your import, it seems to have no plan to ever support another ssp, so this seems misleading

mnguyen-raveltech:
> one more comment:
> 
> > The RavelTech Prebid Adapter first version supports only Xandr/Appnexus SSP
> 
> given your import, it seems to have no plan to ever support another ssp, so this seems misleading

I removed the line. We'll edit the documentation when we'll extend the support to other SSP.

modules/raveltechBidAdapter.js
mnguyen-raveltech: We reached our contact at Xandr asking for them to sort this out internally.

Question though: I added `gvlid: baseAdapter.GVLID` because the behaviour of the ravel adapter doesn't change anything to the AppNexus one (except anonymization).

But maybe the correct way would be to not even touch it and remove the line without providing any gvlid?

mnguyen-raveltech:
> couple of things:
> 
> /home/circleci/Prebid.js/modules/raveltechBidAdapter.js 17:1 warning The type 'BidRequest' is undefined jsdoc/no-undefined-types 64:1 warning The type 'Bid' is undefined jsdoc/no-undefined-types
> 
> I think you can import these types from bidFactory.

Fixed in 4de13aaf96e9484e99a60b11ec971911cb7592e6


> It seems you have done a trick to prevent duplication with bidRichMedia. Instead you should move the functions into imports in your pr for both module files, probably from https://github.com/prebid/Prebid.js/blob/master/libraries/appnexusUtils/anUtils.js

There was actually duplication with big-richmedia and I removed the functions that were not necessary for our adapter to work, hence the reduction in number of duplicated lines.
However, I don't understand the comment about moving functions into imports. Could you please point us to an example?


> Finally, why are you importing the appnexus adapter and disabling the linting rule? This seems more like an rtd module than a bidder in its goals.

The reason why we import the appnexus adapter is because we want to pick up the bid requests from the appnexus buildRequests function and update the uids from there.
Without disabling the linting rule, we encounter the following error when running gulp:
`2:1  error  import "./appnexusBidAdapter.js": importing from modules is not allowed  prebid/validate-imports`

We have followed the implementation of big-richmedia, which also imports the appnexus adapter.
If you think it's not the right way to do it, could you please indicate what would be a cleaner implementation with an example?

modules/raveltechBidAdapter.js
patmmccann: I'm not sure what the anonymize function below does, but wouldn't a publisher want to make sure you had a lawful basis for processing? If you check if xandr does, I'm not sure that's correct?

modules/raveltechBidAdapter.js
patmmccann: Alternatively, why not just modify the xandr bids in place with event listeners?

patmmccann:
Big rich media processes the response and adds a renderer, we're also not sure we're particularly comfortable with that particular implementation

The linter is enforcing the rules. We're not sure you have a good reason to break them yet. Please discuss why alternative implementations don't work for you
https://github.com/prebid/Prebid.js/pull/11988/files#r1687926449

patmmccann:
We discussed in committee, this does not seem to be the appropriate integration path to the project for your goals. The Xandr team is welcome to advocate for you in committee or you can join the project and tell us your reasoning. The linting rules are there for a reason and disabling them is reasonable to consider a rules violation, yet we also have the explicit rule that bidders are not supposed to make calls to other endpoints for reasons other than bidding and it seems reasonable to assume your zkad library likely does this.

Some alternative paths:
-become an appnexus alias and add the call to zkad in that module
-become an rtd module
-use the js you already have on page to modify the appnexus requests in place using the beforeBidRequested event

Issue #11994, opened 2024-07-18
Appnexus BidAdapter: It may not be possible to use currencies other USD


saitoukun:
## Type of issue

- Adapter and server mismatch

## Description

- We use `JPY` currency.
- `currency` will use `publisher_currency_codename` from  `Server's Bid`. This case is `JPY`
- But `cpm` is always seems to be `USD`
- As a result, the currency exchange process was not executed and was regarded as a very low cpm.

This means that the server may only support `USD`

## Platform details

There was no problem up to 8.50.0
I got problems since 8.51.0

> Appnexus Bid Adapter: parse the currency from the bid if specified
https://github.com/prebid/Prebid.js/releases/tag/8.51.0

## Other information

related PR
https://github.com/prebid/Prebid.js/pull/11581

saitoukun:
Related MR
https://github.com/prebid/Prebid.js/pull/11581

saitoukun:
I'm contacting appnexus for specs.

saitoukun:
Here is sample of bid responce.
```
{
  "cpm": 0.477453,
  "cpm_publisher_currency": 74.943749,
  "publisher_currency_code": "¥",
  "publisher_currency_codename": "JPY",
  "content_source": "rtb",
  "ad_type": "banner",
  "media_type_id": 1,
  "media_subtype_id": 1,
  "brand_category_id": 180,
  "is_sov": false,
  "is_roadblock": false,
  "bidder_id": 440,
  "client_initiated_ad_counting": true,
  "is_managed": false,
  "no_consent": false,
}
```

- `publisher_currency_codename` is `JPY`
- It seems to be `cpm` is  always `USD`
- I found `cpm_publisher_currency`. It seems to be `JPY` (Now `$0.477` is around 74JPN)
- Maybe we should use this param, If we use currency other USD.

saitoukun:
I am a partner, and this is just a guess from the Bid responce.
So I think It should be revert once. 
And then we should ask Appnexus engineers for imprement `newBid` function with the exact specifications.

jsnellbaker:
Hi all,

So I contacted an engineer who works on the appnexus endpoint and confirmed some details about the cpm and currency functionality (as it changed a little).

The `cpm` field in our endpoint's bid response will always be USD.
The `cpm_publisher_currency` field in the bid response will reflect and convert to the currency that was set at the publisher entity in the ad server system.  

[...]

PR #11995, opened 2024-07-18
Revert "Appnexus Bid Adapter: parse the currency from the bid if specified"

saitoukun:
## Description
AppnexusBidAdapter have possibility to not able to use any currency other USD.

Reverts prebid/Prebid.js#11581
resolve https://github.com/prebid/Prebid.js/issues/11994

### DIFF
--- a/modules/appnexusBidAdapter.js
+++ b/modules/appnexusBidAdapter.js
@@ -566,7 +566,7 @@ function newBid(serverBid, rtbBid, bidderRequest) {
     cpm: rtbBid.cpm,
     creativeId: rtbBid.creative_id,
     dealId: rtbBid.deal_id,
-    currency: rtbBid.publisher_currency_codename || 'USD',
+    currency: 'USD',
     netRevenue: true,
     ttl: 300,
     adUnitCode: bidRequest.adUnitCode,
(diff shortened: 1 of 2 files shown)


jsnellbaker (approved):

astartsky:
The initial change was intended to fix currency conversion for the case when you have non-usd default currency. Can you please explain how it is supposed to work without it now?

jsnellbaker:
Unless I'm mistaken, I think that's what the prebid currency module is meant to help address?  Can you explain your use-case more if the currency module can't help in the issue https://github.com/prebid/Prebid.js/issues/11994 where it asked for the change to be reverted?

cpcpn-emil:
Am i missing something or does this make it impossible to have setup where the appnexus bid adapter repond with non-usd currency while the adserverCurrency is also non usd?
@jsnellbaker how would the currency module convert this response:

``` js
 {
      "cpm": 1.122080,
      "cpm_publisher_currency": 7.699993,
      "publisher_currency_code": "KR",
      "publisher_currency_codename": "DKK",
      "content_source": "rtb",
      "ad_type": "banner",
}

```

With the currency config:
```
currency: {
    adServerCurrency:"DKK"
}
```

Im seeing that the bid gets converted with the USD currency (which after the revert is hardcoded in the appnexus adapter)

PR #11999, opened 2024-07-18
MobianRtdModule: Add more signals from API endpoint to first-party data

ehb-mtk:
## Type of change

- [x] Feature

## Description of change
In addition to adding Mobian's determination of brand safety to the first party data, we now also add its
determination of sentiment, emotion, and garm risk categories.

### DIFF
--- a/modules/mobianRtdProvider.js
+++ b/modules/mobianRtdProvider.js
@@ -1,10 +1,22 @@
+/**
+ * This module adds the Mobian RTD provider to the real time data module
+ * The {@link module:modules/realTimeData} module is required
+ * @module modules/anonymisedRtdProvider
+ * @requires module:modules/realTimeData
+ */
 import { submodule } from '../src/hook.js';
 import { ajaxBuilder } from '../src/ajax.js';
+import { deepSetValue } from '../src/utils.js';
+
+/**
+ * @typedef {import('../modules/rtdModule/index.js').RtdSubmodule} RtdSubmodule
+ */
+
 export const MOBIAN_URL = 'https://impact-api-prod.themobian.com/brand_safety';
 
+/** @type {RtdSubmodule} */
 export const mobianBrandSafetySubmodule = {
   name: 'mobianBrandSafety',
-  gvlid: null,
   init: init,
   getBidRequestData: getBidRequestData
 };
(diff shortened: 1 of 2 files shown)


modules/mobianRtdProvider.js
```diff
@@ -1,17 +1,30 @@
+/**
+ * This module adds the Mobian RTD provider to the real time data module
+ * The {@link module:modules/realTimeData} module is required
+ * @module modules/anonymisedRtdProvider
```
patmmccann, line 4: are you from idward / anonymised? do you mean to import their type?

ehb-mtk: no -- that's embarrassing. I've had trouble getting the build to include the real time data module without including it explicitly, and hoped that typing would help. Will remove...

patmmccann: Add yourself to  .submodules.js

patmmccann: https://github.com/prebid/Prebid.js/blob/dc6c0c35ead5f44f3fc35852a84956a6c9036bed/modules/.submodules.json#L88

patmmccann (changes requested):
Submodules file reference is missing

ehb-mtk:
(Thanks -- fixed!! In the last commit, we've fixed another bug where the browser interprets the API response as a string, in which case we need to call `JSONParse`. We're doing this conservatively in case it's browser-dependent. Let me know if you'd prefer that we make this a separate PR.)

modules/mobianRtdProvider.js
```diff
@@ -24,6 +22,21 @@ export const mobianBrandSafetySubmodule = {
 function init() {
   return true;
 }
+
+function safeJSONParse(input) {
```
patmmccann, line 26: can you import this? https://github.com/prebid/Prebid.js/blob/465096c4ab99964e9210e969c404382e13142c3b/src/utils.js#L1107 and if that is problematic, could you pr the util instead of your adpter with the extra logic

ehb-mtk: Thanks! I'll just import this, I haven't got any reason to believe that in fact the API response *ever* came in pre-parsed. It'll be a bit cause I need to fix the tests.

patmmccann (approved):

PR #12004, opened 2024-07-19
ttd bid adapter: configurable endpoint 

tongwu-sh:
## Type of change
- [x] Feature

## Description of change

We want to allow users to specify their preferred TTD endpoint instead of the default endpoint for bidding.

### DIFF
--- a/libraries/connectionInfo/connectionUtils.js
+++ b/libraries/connectionInfo/connectionUtils.js
@@ -13,7 +13,9 @@ export function getConnectionType() {
       return 1;
     case 'wifi':
       return 2;
-    case 'cellular':
+    case 'wimax':
+      return 6;
+    default:
       switch (connection.effectiveType) {
         case 'slow-2g':
         case '2g':
@@ -23,9 +25,7 @@ export function getConnectionType() {
         case '4g':
           return 6;
         default:
-          return 3;
+          return connection.type == 'cellular' ? 3 : 0;
       }
-    default:
-      return 0;
   }
 }
--- a/libraries/connectionInfo/connectionUtils.js
+++ b/libraries/connectionInfo/connectionUtils.js
@@ -1,31 +1,31 @@
 /**
  * Returns the type of connection.
  *
- * @returns {int} - Type of connection.
+ * @returns {number} - Type of connection.
  */
 export function getConnectionType() {
-    const connection = navigator.connection || navigator.webkitConnection;
-    if (!connection) {
-      return 0;
-    }
-    switch (connection.type) {
-      case 'ethernet':
-        return 1;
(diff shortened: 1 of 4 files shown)


patmmccann (changes requested):
Couple of minor requests

tongwu-sh:
> Hi @tongwu-sh ? It seems your adapter needs a tiny bit of additional maintenance: please move the common get connection logic in your adapter and prisma into a file in /libraries
> 
> It seems you also have a bug in your battr support; I think it is on imp.banner instead of imp
> 
> Please also document this endpoint change on the docs repo

Thanks for the comments; I've made the updates. How should we test this part? `Be sure to test the integration with your ad server using the [Hello World](https://github.com/integrationExamples/gpt/hello_world.html) sample page.` It seems the URL is not working.

patmmccann (approved):

patmmccann:
This looks good to me, if you pull in master you'll pass that viant test. I noticed you marked as draft so holding off on merge

Thanks!

libraries/connectionInfo/connectionUtils.js
```diff
@@ -0,0 +1,31 @@
...
+ */
+export function getConnectionType() {
+  const connection = navigator.connection || navigator.webkitConnection;
+  if (!connection) {
+    return 0;
+  }
+  switch (connection.type) {
+    case 'ethernet':
+      return 1;
+    case 'wifi':
```
Grammarian, line 14: Do we want to include "wimax" with this case?

patmmccann:
> > Hi @tongwu-sh ? It seems your adapter needs a tiny bit of additional maintenance: please move the common get connection logic in your adapter and prisma into a file in /libraries
> > It seems you also have a bug in your battr support; I think it is on imp.banner instead of imp
> > Please also document this endpoint change on the docs repo
> 
> Thanks for the comments; I've made the updates. How should we test this part? `Be sure to test the integration with your ad server using the [Hello World](https://github.com/integrationExamples/gpt/hello_world.html) sample page.` It seems the URL is not working.

Fixed, #12011

libraries/connectionInfo/connectionUtils.js
tongwu-sh: Yes, added. And checked for chrome/edge looks like connection.type is not always there but connection.effectiveType has value. Updated to handle that case.

tongwu-sh:
@patmmccann , we finished the test for this change, could you help merge it? Thanks!

tongwu-sh:
Hello @patmmccann , we finished tests for this change and ready to merge, could you help take another look at this PR? Thanks!

patmmccann (approved):

PR #12005, opened 2024-07-19
Index Exchange Bid Adapter: Add full ORTB2 device data to request payload

jwrosewell:
## Type of change
- [x] Feature
- [x] Updated bidder adapter

## Description of change
This PR enhances the Index Exchange bid request by incorporating device data from the global ORTB2 object. Existing values in the IX device object won't be replaced (e.g. screen size).

The device object has previously been populated by simplistic parsers, if at all, and was inaccurate as a result. Prebid now benefits from RTD modules such as 51Degrees that enrich all the device object fields including Apple iPhone model category and device ID. The PR enables Index Exchange’s users to benefit from device object improvements.

## Other information
cc: @ccorbo

### DIFF
--- a/modules/ixBidAdapter.js
+++ b/modules/ixBidAdapter.js
@@ -746,7 +746,7 @@ function buildRequest(validBidRequests, bidderRequest, impressions, version) {
 
     const isLastAdUnit = adUnitIndex === impKeys.length - 1;
 
-    r = addDeviceInfo(r);
+    r = addDeviceInfo(r, bidderRequest?.ortb2);
     r = deduplicateImpExtFields(r);
     r = removeSiteIDs(r);
 
@@ -2079,14 +2079,21 @@ function isValidAuctionConfig(config) {
 }
 
 /**
- * Adds device.w / device.h info
+ * Adds device info from ORTB2 data to the request
  * @param {object} r
+ * @param {object} ortb2Data
  * @returns object
  */
-export function addDeviceInfo(r) {
-  if (r.device == undefined) {
-    r.device = {};
+export function addDeviceInfo(r, ortb2Data) {
+  // add device object if not present
+  r.device = r.device || {};
+
+  // if present, merge device object from ortb2Data
+  if (ortb2Data?.device) {
+    mergeDeep(r.device, ortb2Data.device);
   }
+
+  // override screen width and height with window.screen values
   r.device.h = window.screen.height;
   r.device.w = window.screen.width;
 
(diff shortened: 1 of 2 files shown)


modules/ixBidAdapter.js
```diff
@@ -2079,14 +2079,21 @@ function isValidAuctionConfig(config) {
...
+export function addDeviceInfo(r, ortb2Data) {
+  // add device object if not present
+  r.device = r.device || {};
+
+  // if present, merge device object from ortb2Data
+  if (ortb2Data?.device) {
+    mergeDeep(r.device, ortb2Data.device);
   }
+
+  // override screen width and height with window.screen values
```
patmmccann, line 2096: @mkendall07 do you know why this override exists?

mkendall07: I guess this is the reason > https://github.com/prebid/Prebid.js/pull/10246#discussion_r1268581470

I haven't looked at the code but I guess core handles this now?

patmmccann: https://github.com/prebid/Prebid.js/blob/dc6c0c35ead5f44f3fc35852a84956a6c9036bed/src/fpd/enrichment.js#L100

patmmccann: slightly different logic it seem https://stackoverflow.com/questions/16162639/difference-between-screen-and-window-property

BohdanVV: Hi @patmmccann!

> do you know why this override exists?

I've noticed that this approach is common in many adapters. I believe the reason lies in the OpenRTB specification, where `w` and `h` are defined as the physical height and width <ins>_of the screen_</ins> in pixels. In contrast, the default implementation returns the viewport size.

If you'd like, I can create an issue to discuss this topic further. It might be something that should be addressed at the Prebid.js FPD level.

patmmccann: Please do, thanks!

mkendall07:
Hi @jwrosewell thanks for the PR. IX currently does server side population of these fields so I'm going to decline this change for now. Feel free to reach out directly if there is any specific publisher that are looking for these features. Thanks

jwrosewell:
@mkendall07 - are you on Prebid Slack?

This PR prevents IX missing out on data enriched via Prebid without restricting the option to populate the fields server side. For example; the Apple model category, which can't be determined server side, may be populated and IX and any downstream partners will loose visibility of this information. Further where the data enrichment has already be carried out by the publisher there may be no need to repeat the same processing improving efficiency.

jwrosewell:
@passani

PR #12012, opened 2024-07-22
Rise Utils: Bugfixes

### LINKED ISSUE #11289: Invalid requests generated by minuteMedia and rise bid adapters
## Type of issue
Bug

## Description
The util function [getBidIdParameter](https://github.com/prebid/Prebid.js/blob/8.42.0/src/utils.js#L105) can cause the returned value and type to change. E.g. when the value is `0` it ends up being an empty string.

I'm not sure if the bidders are not handling it correctly or if this is a not intended side effect.

There are bidder endpoints, like the ones below, which expect an integer, but end up receiving a string and result in a 400 request.
- https://github.com/prebid/Prebid.js/blob/8.42.0/modules/minutemediaBidAdapter.js#L303
- https://github.com/prebid/Prebid.js/blob/8.42.0/modules/riseBidAdapter.js#L313C30-L313C49

## Other information
- https://github.com/prebid/Prebid.js/issues/11288

zkosanovic:
## Type of change

- [x] Bugfix

## Description of change

1. Rise server expects `loop` parameter to be an integer, so it should default to 0 instead of an empty string.
2. For `page_domain` parameter, we should get the value from `refererInfo` if present.

Fixes #11289

### DIFF
--- a/libraries/riseUtils/index.js
+++ b/libraries/riseUtils/index.js
@@ -324,6 +324,7 @@ export function generateGeneralParams(generalObject, bidderRequest, adapterVersi
   if (bidderRequest && bidderRequest.refererInfo) {
     generalParams.referrer = deepAccess(bidderRequest, 'refererInfo.ref');
     generalParams.page_url = deepAccess(bidderRequest, 'refererInfo.page') || deepAccess(window, 'location.href');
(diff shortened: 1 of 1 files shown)


zkosanovic:
Fixes #11289

patmmccann (approved):

PR #12014, opened 2024-07-22
Mobian RTD Module: remove reference to garm in output of mobian brand-safety

ehb-mtk:
## Type of change

## Description of change
Our product team has asked us not to describe the given output as being created by GARM.

I also removed a loop that used to be necessary to compute the risk from the API output but is now redundant because the API output returns the risk directly.

### DIFF
--- a/test/spec/modules/mobianRtdProvider_spec.js
+++ b/test/spec/modules/mobianRtdProvider_spec.js
@@ -23,87 +23,7 @@ describe('Mobian RTD Submodule', function () {
     ajaxStub.restore();
   });
 
-  it('should return no_risk when server responds with garm_no_risk', function () {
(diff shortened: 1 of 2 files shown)


test/spec/modules/mobianRtdProvider_spec.js
```diff
@@ -23,70 +23,63 @@ describe('Mobian RTD Submodule', function () {
     ajaxStub.restore();
   });
 
-  it('should return no_risk when server responds with garm_no_risk', function () {
+  it('should return risk level when server responds with garm_risk', function () {
```
ehb-mtk, line 26: the diff on the tests looks huge, but isn't -- we're just removing the testing of the old `garm_x_risk` since we always expect `garm_risk` coming in.

patmmccann (approved):

patmmccann:
i feel your pain with this make work, approving

PR #12018, opened 2024-07-23
Uniq BID ID analytics module

slonofanya:
### Module: `uBidIdAnalyticsAdapter`

This module provides Uniq ID for every bidding request.
It prevents from duplicate bidding requests from the same user.

## Type of change
- [x ] New analytics adapter 

contact email: slonofanya@gmail.com


patmmccann:
Core already does this with imp.ext.tid and source.tid.

modules/uBidIdAnalyticsAdapter.js
```diff
@@ -0,0 +1,66 @@
...
+    return 'xxxxxxxx-xxxx-4xxx-yxxx-xxxxxxxxxxxx'.replace(/[xy]/g, function(c) {
+      var r = (d + Math.random() * 16) % 16 | 0;
+      d = Math.floor(d / 16);
+      return (c === 'x' ? r : (r & 0x3 | 0x8)).toString(16);
+    });
+  }
+};
+
+submodule(MODULE_NAME, uBidIdModule);
+
```
patmmccann, line 62: You don't call https://github.com/slonofanya/Prebid.js/blob/32383ce479f4bff5a103465baace4833aaecf20c/modules/magniteAnalyticsAdapter.js#L1082 ?

Issue #10062, opened 2023-06-07
Interest Yield on the Main Thread to optimize UX of sites using prebid.js?


gilbertococchi:
## Type of issue
Feature Request

## Description
Chrome recently announced that  [INP](https://web.dev/inp/) will become a Core Web Vitals Metric.
To optimize User Experience as INP suggest it's a good approach for 1P and 3P developers to Yield on the Main thread and reduce Long Tasks when possible.

[https://web.dev/optimize-long-tasks/](https://web.dev/optimize-long-tasks/) is a great article that covers the dfiferent methodology to Yield on the Main Thread.

Some of these methods are available (like scheduler.postTask) and some others will be shortly available via origin Trival (scheduler.yield).

It would be positive if prebid.js as well as other 3Ps libraries used by websites would Yield on the Main Thread, so in case of User Interaction, the Browser will be able to Render the Next Frame.

In the Screenshot you can see the Long Task generated by the prebid.js Example container taken [from this link](https://docs.prebid.org/dev-docs/examples/basic-example.html)

<img width="709" alt="prebid" src="https://github.com/prebid/Prebid.js/assets/329023/e03696f7-19ed-41d4-9bbe-fe52c9e8a50f">

Would there be any interest from prebid.js Library owners to consider some of this approach in order to allow Prebid.js to work as usual while reducing it's Main Thread blocking time?

happy to chat more about this.
Gilberto

mmocny:
Thanks for that extra context.  Here is one comment that I think summarizes a long thread well:

> i don't think a `if (timeout === 0) callback() else setTimeout(callback, timeout)` style is a good practice in general but i think it makes perfect sense in our case. the whole concept of header bidding is to hold an auction in the browser before the page loads with the goal of rendering with the page, and anytime we defer execution we work against that goal.

I think this makes sense to me: *if* there is explicit intent to block rendering on the results of the add auction, to prevent flash of content without adds -> with ads, that seems reasonable.

However, once the page is already rendered, and already looks interactive, I think this constraint should not be upheld.

I dont know enough about prebid.js to know how often pages uses it in the latter use case -- but it seems to me that if it is a non-trivial amount, that you should support both modes?

dgirardi:
[...]

PR #12025, opened 2024-07-23
Core: start yielding control of the main thread

dgirardi:
## Type of change

- [x] Feature

## Description of change

- Make `GreedyPromise` optional: it can be excluded with `gulp build --disable GREEDY` and/or overridden with `pbjs.Promise = window.Promise` (before Prebid loads)
- Refactor tests so that they can pass on both vanilla and greedy promises

Since promises are already used in many places, "vanilla" promises should have the effect of breaking up long running tasks (https://github.com/prebid/Prebid.js/issues/10062), but they are also potentially breaking - the timing of configuration taking effect, event handlers running, etc all change (relative to when their setup code runs).

The plan is to do some testing and, if necessary, find more tasks to split using promises - so that any such change only affects those that opt out of GreedyPromise.


gilbertococchi:
Hi @dgirardi, are you still committed to move this branch forward?

I just wanted to share that the scheduler.yield Yield method to allow 3Ps or 1P Script led Tasks to Yield while ensuring continuation (so without causing delay into Ad bids auction) is now widely available in Chrome.

I think that may be a good fit for this branch proposal for interested publishers to test it out if applicable.

patmmccann:
Raptive is planning some quantitative tests on this branch in November

gilbertococchi:
Thanks for your answer @patmmccann!

If you have feedback or  questions  about yield best practices on the Yield I am available to answer those questions.

patmmccann:
This is ready to merge when the conflicts get cleaned up

patmmccann (approved):

PR #12028, opened 2024-07-23
Lasso prebid adapter add npi support

victorlassomarketing:
## Type of change

- [ ] Bugfix
- [x] Feature
- [ ] New bidder adapter  
- [ ] Updated bidder adapter   
- [ ] Build related changes
- [ ] CI related changes

- [ ] Other

### DIFF
--- a/modules/lassoBidAdapter.js
+++ b/modules/lassoBidAdapter.js
@@ -130,10 +130,10 @@ function getBidRequestUrl(aimXR, params) {
   if (params && params.dtc) {
     path = '/dtc-request';
   }
-  if (!aimXR && !(params.npi || params.npiHash)) {
-    return GET_IUD_URL + ENDPOINT_URL + path;
+  if (aimXR || params.npi || params.npiHash) {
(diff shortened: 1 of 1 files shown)


modules/lassoBidAdapter.js
```diff
@@ -44,6 +44,8 @@ export const spec = {
         sizes,
         aimXR,
         uid: '$UID',
+        npi: bidRequest.params.npi || '',
```
patmmccann, line 47: Why not read the entire bidRequest.user.ext.data object so it isn't so clear your code is gathering pii?

modules/lassoBidAdapter.js
```diff
@@ -44,6 +44,8 @@ export const spec = {
...
+        npi_hash: bidRequest.params.npiHash || '',
```
patmmccann, line 48: This appears to be unnecessary bc the line below already catches it

patmmccann (changes requested):
This pr doesn't seem useful

PR #12032, opened 2024-07-24
Navegg UserID Submodule: conform with pub storage configuration

p-ribeiro:
## Type of change
- [x] Bugfix
- [x] Other

## Description of change
Fixing the problems described at https://github.com/prebid/Prebid.js/issues/10710.
We are now saving the ID in cookie or html5, respecting the storage.type parameter. 
We also changed our logic to get the ID from our API.

### DIFF
--- a/modules/naveggIdSystem.js
+++ b/modules/naveggIdSystem.js
@@ -6,9 +6,9 @@
  */
 import { isStr, isPlainObject, logError } from '../src/utils.js';
 import { submodule } from '../src/hook.js';
-import * as ajaxLib from '../src/ajax.js';
-import {getStorageManager} from '../src/storageManager.js';
-import {MODULE_TYPE_UID} from '../src/activities/modules.js';
+import { ajaxBuilder } from '../src/ajax.js';
+import { getStorageManager } from '../src/storageManager.js';
+import { MODULE_TYPE_UID } from '../src/activities/modules.js';
 
 /**
  * @typedef {import('../modules/userId/index.js').Submodule} Submodule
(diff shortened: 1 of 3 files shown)


p-ribeiro:
Can someone give me an explanation on why did my test failed? It seems that what failed are another modules tests. 

For example the IntentQ and imuId module...

```

166) should return expired is true
     imuId module getLocalData()
     AssertionError: expected { Object (id, ppid, ...) } to deeply equal { Object (id, ppid, ...) }
    at Context.<anonymous> (/tmp/_karma_webpack_724924/commons.js:314947:49)

167) should save data in cookie if storage type is "cookie"
     IntentIQ tests
     AssertionError: expected null to not equal null
    at Context.<anonymous> (/tmp/_karma_webpack_724924/commons.js:317159:43)

168) "before each" hook for "should set cmpData.us_privacy if uspData exists"
     IntentIQ tests IntentIQ consent management within getId
```

modules/naveggIdSystem.js
```diff
@@ -6,7 +6,7 @@
  */
 import { isStr, isPlainObject, logError } from '../src/utils.js';
 import { submodule } from '../src/hook.js';
-import { ajax } from '../src/ajax.js';
+import * as ajaxLib from '../src/ajax.js';
```
patmmccann, line 9: please only import what you need

patmmccann (changes requested):
Thanks so much, one tweak requested

p-ribeiro:
> Thanks so much, one tweak requested
The change was done as requested with my latest commit

p-ribeiro:
May I ask a question here?

Is it permitted to add a cookie saver, as a default option, if the client does not implemente the storage options?

patmmccann:
The parent module handles the storage writes

patmmccann (approved):

Issue #11391, opened 2024-04-24
Update pbsBidAdapter to pass analytics tags to client-side adapters


bretg:
## Type of issue
Enhancement

## Description
This is the Prebid.js side of https://github.com/prebid/prebid-server/issues/3615 - see that issue for a full background.

In short, we're trying to get server-side details to our client-side analytics adapter and the PBS committee has approved a generic mechanism for doing so... it will return the `ext.prebid.analytics.tags` object in the response. We would like for the pbsBidAdapter to look for this object and make it available to interested analytics adapters. 

We should emit a new event for the server side analytics object being available, and only emit the event when the library is configured by the publisher to do so (the same configuration flag would enable  ext.prebid.analytics.options.enableclientdetails: true in the pbs request) . Analytics adapters could then listen to that event.

dgirardi:
Does it make sense to re-use the SEAT_NON_BID event for this, possibly renaming it (PBS_ANALYTICS ? ) for prebid 9?

bretg:
Will let @robertrmartinez weigh in, but I suspect we're willing to follow Prebid's lead on the implementation.

PR #12015, opened 2024-07-22
PBS Index Atag changes

spotxslagle:
## Type of change

- [ x ] Feature

## Description of change

Atag changes for PBS Adapter

PR #12044, opened 2024-07-26
PBS Adapter: Add PBS_ANALYTICS Event

spotxslagle:
## Type of change

- [ x ] Feature

## Description of change

Add PBS_ANALYTICS event and pass ATAG data through it

## Other information

Supercedes https://github.com/prebid/Prebid.js/pull/12015

### DIFF
--- a/modules/prebidServerBidAdapter/index.js
+++ b/modules/prebidServerBidAdapter/index.js
@@ -475,8 +475,8 @@ export function PrebidServer() {
           if (isValid) {
             bidRequests.forEach(bidderRequest => events.emit(EVENTS.BIDDER_DONE, bidderRequest));
           }
-          const { emitSeatNonBid, emitAtag } = shouldEmitAnalyticsEvent(s2sBidRequest.s2sConfig, response)
-          if (emitSeatNonBid) {
+          const { seatNonBidData, atagData } = getAnalyticsFlags(s2sBidRequest.s2sConfig, response)
+          if (seatNonBidData) {
             events.emit(EVENTS.SEAT_NON_BID, {
               seatnonbid: response.ext.seatnonbid,
               auctionId: bidRequests[0].auctionId,
@@ -486,10 +486,10 @@ export function PrebidServer() {
             });
           }
           // pbs analytics event
-          if (emitSeatNonBid || emitAtag) {
+          if (seatNonBidData || atagData) {
             const data = {
-              seatnonbid: emitSeatNonBid ? response.ext.seatnonbid : undefined,
-              atag: emitAtag ? response.ext.prebid.analytics.tags : undefined,
+              seatnonbid: seatNonBidData,
+              atag: atagData,
               auctionId: bidRequests[0].auctionId,
               requestedBidders,
               response,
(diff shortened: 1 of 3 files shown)


spotxslagle:
fixes https://github.com/prebid/Prebid.js/issues/11391

JulieLorin:
I was reading the linked issue and I see `We should emit a new event for the server side analytics object being available, and only emit the event when the library is configured by the publisher to do so `
Not sure what `when the library is configured by the publisher to do so` exactly means but shouldn't that means there should be a configuration value ? (Or is that only for server side ?)

Otherwise, the code looks good

modules/prebidServerBidAdapter/index.js
```diff
@@ -601,10 +614,20 @@ export const processPBSRequest = hook('sync', function (s2sBidRequest, bidReques
   }
 }, 'processPBSRequest');
 
+function shouldEmitAnalyticsEvent(s2sConfig, response) {
```
osazos, line 617: The naming sounds confusing, like a function that returns a bool that allows to emit the "analytics" event. 
Can we try to find something better?
Proposition:  `allowedPbsEvents`, `getPbsEventsEmitStatus`, `emitPbsEvents`, something like that.

modules/prebidServerBidAdapter/index.js
```diff
@@ -484,6 +485,18 @@ export function PrebidServer() {
               adapterMetrics
             });
           }
+          // pbs analytics event
+          if (emitSeatNonBid || emitAtag) {
+            const data = {
+              seatnonbid: emitSeatNonBid ? response.ext.seatnonbid : undefined,
+              atag: emitAtag ? response.ext.prebid.analytics.tags : undefined,
```
osazos, line 492: consistency:

```suggestion
              seatnonbid: response?.ext?.seatnonbid,
              atag: response?.ext?.prebid?.analytics?.tags,
```

If the ternary condition is related to the "OR" condition above, there is perhaps an issue because you can still have access to the values in the `data.response` field.

patmmccann:
@spotxslagle should we also change generic analytics adapter to listen for this as in scope here?

robertrmartinez (approved):
LGTM

Regarding hiding behind config flag I do not think it is necessary since the publisher would be working with a PBS which they may expect or not to set these EXT objects. But if we feel more comfortable hiding this logic I guess we can.

These events only fire if the PBS response contains the EXTs.

patmmccann:
>Regarding hiding behind config flag I do not think it is necessary since the publisher would be working with a PBS which they may expect or not to set these EXT objects.

That's a great point, seems like a good reason to default to true?

robertrmartinez:
yeah - imo that as well is overkill.

@osazos could you give your re-review after @spotxslagle updated the function names etc? Thanks!

osazos (approved):
LGTM

@spotxslagle, seen after approved: If you can fix the name of the function ~~`getNonbidData`~~ -> `getNonBidData` in a consistency matter only. _(btw this is cosmetic)_

osazos:
@spotxslagle can you fix the function name related to my last comment?
@ChrisHuie do you want to add your review?

spotxslagle:
> @spotxslagle can you fix the function name related to my last comment? @ChrisHuie do you want to add your review?

Done!

Issue #10656, opened 2023-10-30
getConfig cleanup of currency


patmmccann:
## Type of issue
getConfig is used by adapters to gather currency.adServerCurrency, currency, and floors.data.currency

## Description
getConfig is a poor access point for bidders to a variety of fields, which may be dynamic to an ad unit and no longer match config. 

Some examples: 
https://github.com/prebid/Prebid.js/blob/56292c4672a0b19bee34840dca93bd3e1e5b8afe/modules/adfBidAdapter.js#L63
https://github.com/prebid/Prebid.js/blob/56292c4672a0b19bee34840dca93bd3e1e5b8afe/modules/beopBidAdapter.js#L140

I propose these all get cleaned up to look at the request object field `cur` and that the object is autopopulated by config

PR #12051, opened 2024-07-29
Various bid adapters: Currency config cleanup

mkomorski:
## Type of change

- [ ] Bugfix
- [ ] Feature
- [ ] New bidder adapter  
- [x] Updated bidder adapter   
- [ ] Build related changes
- [ ] CI related changes

- [ ] Other

## Other information

#10656

### DIFF
--- a/libraries/currencyUtils/currency.js
+++ b/libraries/currencyUtils/currency.js
@@ -29,3 +29,7 @@ export function currencyNormalizer(toCurrency = null, bestEffort = true, convert
 export function currencyCompare(get = (obj) => [obj.cpm, obj.currency], normalize = currencyNormalizer()) {
   return keyCompare(obj => normalize.apply(null, get(obj)))
 }
+
+export function getCurrencyFromBidderRequest(bidderRequest) {
+  return (bidderRequest?.ortb2?.cur || [])[0];
+}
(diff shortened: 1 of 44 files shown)


libraries/currencyUtils/currency.js
```diff
@@ -29,3 +29,7 @@ export function currencyNormalizer(toCurrency = null, bestEffort = true, convert
 export function currencyCompare(get = (obj) => [obj.cpm, obj.currency], normalize = currencyNormalizer()) {
   return keyCompare(obj => normalize.apply(null, get(obj)))
 }
+
+export function getCurrencyFromBidderRequest(bidderRequest) {
+  return (bidderRequest?.ortb2?.cur || [])[0];
```
dgirardi, line 34: I think this should be its own library, or even in core. the issue with putting it here is that almost none of the consumers need the rest of this library, so to get a 1 line utility you pull in Kbs of code.

modules/currency.js
```diff
@@ -334,4 +337,11 @@ export function setOrtbCurrency(ortbRequest, bidderRequest, context) {
...
 
-registerOrtbProcessor({type: REQUEST, name: 'currency', fn: setOrtbCurrency});
+registerOrtbProcessor({type: REQUEST, name: 'currency', fn: setOrtbCurrency, priority: 100});
+
+function enrichFPDHook(next, fpd) {
+  return next(fpd.then(ortb2 => {
+    deepSetValue(ortb2, 'cur', [adServerCurrency]);
+    return ortb2;
+  }))
+}
```
dgirardi, line 347: `enrichFPDHook` here always overwrites what's set by `setOrtbCurrency`.

PR #12057, opened 2024-07-30
mgid bid adapters: refactoring for trimmer code

sanychtasher:
## Type of change
- [ ] Updated bidder adapter 

## Description of change
Deduplicate code in mgid prebid adapter

### DIFF
--- a/modules/mgidXBidAdapter.js
+++ b/modules/mgidXBidAdapter.js
@@ -4,6 +4,7 @@ import { BANNER, NATIVE, VIDEO } from '../src/mediaTypes.js';
 import { config } from '../src/config.js';
 import { USERSYNC_DEFAULT_CONFIG } from '../src/userSync.js';
 import { isBidRequestValid, buildRequestsBase, interpretResponse } from '../libraries/teqblazeUtils/bidderUtils.js';
+import { getUserSyncs } from './mgidBidAdapter.js'
 
 const BIDDER_CODE = 'mgidX';
 const GVLID = 358;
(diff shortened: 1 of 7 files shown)


modules/mgidXBidAdapter.js
```diff
@@ -4,6 +4,7 @@ import { BANNER, NATIVE, VIDEO } from '../src/mediaTypes.js';
 import { config } from '../src/config.js';
 import { USERSYNC_DEFAULT_CONFIG } from '../src/userSync.js';
 import { isBidRequestValid, buildRequestsBase, interpretResponse } from '../libraries/teqblazeUtils/bidderUtils.js';
+import { getUserSyncs } from './mgidBidAdapter.js'
```
patmmccann, line 8: Generally this is looking great! However, you can import from /libraries/..., not from another adapter, so please move this function to a file in that folder for both

patmmccann:
I see you're triggering warnings when you try to move stuff out of the redtram adapter. This is very much appreciated and you don't have to solve all of redtram for your mgid changes to merge in

Thanks!!!

patmmccann:
Apologies if my last comment was misleading, we really appreciate the change to redtram and undertone, I was just saying you can ignore the robot complaining about other problems on those other adapters

sanychtasher:
@patmmccann We don't know how to fix it more correctly 
[modules/adtrueBidAdapter.js](https://github.com/prebid/Prebid.js/blob/60108e6113d5701a008acd6810044eae7880353e/modules/adtrueBidAdapter.js#L50-L55) has 8 duplicated lines with [modules/mgidBidAdapter.js](https://github.com/prebid/Prebid.js/blob/60108e6113d5701a008acd6810044eae7880353e/modules/mgidBidAdapter.js#L39-L43).
Maybe you could recommend a solution

patmmccann:
> @patmmccann We don't know how to fix it more correctly [modules/adtrueBidAdapter.js](https://github.com/prebid/Prebid.js/blob/60108e6113d5701a008acd6810044eae7880353e/modules/adtrueBidAdapter.js#L50-L55) has 8 duplicated lines with [modules/mgidBidAdapter.js](https://github.com/prebid/Prebid.js/blob/60108e6113d5701a008acd6810044eae7880353e/modules/mgidBidAdapter.js#L39-L43). Maybe you could recommend a solution
i

I would recommend not worrying about that in this pr and covering later.

sanychtasher:
@patmmccann Can you please check the PR? Thanks!

patmmccann (approved):

Issue #11270, opened 2024-03-28
Feature request: check whether bids are rendered after expiration


dgirardi:
## Description

Currently, bid expiration (or TTL) is used only to determine which bids are eligible to be potential winners. In some situations it's possible for a bid to be "fresh" at the time it's sent out to the ad server, but expired at the time that it's rendered and becomes a winner. In this case the SSP/DSP would see an impression on an expired bid.

We should check whether bids are expired at the time they are rendered. If they are, we could emit an event and/or prevent them from being rendered.

patmmccann:
Let's start with the event, disaster recovery could be for a variety of similar events and covered in different work

patmmccann:
closing to track on #12521

PR #12059, opened 2024-07-30
CORE: prevent unbound growth of suspendedTimeouts and possible NaN values

olafbuitelaar:
## Type of change
- [x ] Bugfix

## Description of change
We have publishers complain about high memory usage caused by the focusTimeout handler for cleaning up expired bids, introduced by; https://github.com/prebid/Prebid.js/pull/11803 . this is because the values in suspendedTimeouts are never cleared. in addition to this if you open a page, and really quickly tab away the value timeOutOfFocus can become NaN when outOfFocusStart isn't set. 

in addition it would be nice this feature is optional, e.g. to allow the configure the old behaviour, e.g. in case you want todo ad-reloads in the background, but still clean them up.

### DIFF
--- a/src/utils/focusTimeout.js
+++ b/src/utils/focusTimeout.js
@@ -6,7 +6,7 @@ function trackTimeOutOfFocus() {
   if (document.hidden) {
     outOfFocusStart = Date.now()
   } else {
-    timeOutOfFocus += Date.now() - (outOfFocusStart); // when the page is loaded in hidden state outOfFocusStart is undefined, which results in timeoutOffset being NaN
+    timeOutOfFocus += Date.now() - (outOfFocusStart ?? 0); // when the page is loaded in hidden state outOfFocusStart is undefined, which results in timeoutOffset being NaN
     outOfFocusStart = null;
     suspendedTimeouts.forEach(({ callback, startTime, setTimerId }) => setTimerId(setFocusTimeout(callback, timeOutOfFocus - startTime)()));
     suspendedTimeouts = [];
--- a/src/utils/focusTimeout.js
+++ b/src/utils/focusTimeout.js
@@ -1,4 +1,4 @@
-let outOfFocusStart;
+let outOfFocusStart = null; // enforce null otherwise it could be undefined and the callback wouldn't execute
 let timeOutOfFocus = 0;
 let suspendedTimeouts = [];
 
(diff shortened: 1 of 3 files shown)


src/utils/focusTimeout.js
```diff
@@ -1,16 +1,27 @@
...
 
-document.addEventListener('visibilitychange', () => {
+function trackTimeOutOfFocus() {
   if (document.hidden) {
     outOfFocusStart = Date.now()
   } else {
-    timeOutOfFocus += Date.now() - outOfFocusStart
-    suspendedTimeouts.forEach(({ callback, startTime, setTimerId }) => setTimerId(setFocusTimeout(callback, timeOutOfFocus - startTime)()))
+    timeOutOfFocus += Date.now() - (outOfFocusStart); // when the page is loaded in hidden state outOfFocusStart is undefined, which results in timeoutOffset being NaN
     outOfFocusStart = null;
```
dgirardi, line 10: You had a comment here calling this to be set before looping on suspendedTimeouts - I removed it as I don't think the order matters, it's checked from within a `setTimeout` callback.

src/utils/focusTimeout.js
```diff
@@ -1,16 +1,27 @@
...
+    suspendedTimeouts.forEach(({ callback, startTime, setTimerId }) => setTimerId(setFocusTimeout(callback, timeOutOfFocus - startTime)()));
+    suspendedTimeouts = [];
```
dgirardi, line 12: all the suspended timeouts should be dealt with here, I think the list can just be cleared instead of filtering it during the loop

dgirardi (approved):
I updated this with some tests, but ended up making a couple logic changes - please let me know if you disagree!

src/utils/focusTimeout.js
olafbuitelaar: yes you're right, it shouldn't matter.

src/utils/focusTimeout.js
olafbuitelaar: i agree, i wasn't sure about it, so took the safer route by just filtering on the callback. but prefer this is this is a safe operation

dgirardi:
> in addition it would be nice this feature is optional, e.g. to allow the configure the old behaviour, e.g. in case you want todo ad-reloads in the background, but still clean them up.

The reason for this "feature" is to avoid an edge case where:

 1. you set up `minBidCacheTTL` (to enable cleanup of expired bids)
 2. a winning bid is sent to the ad server
 3. the page goes out of focus
 4. the bid's TTL expires
 5. the page goes back in focus
 6. cleanup logic runs, the bid is forgotten
 7. the creative attempts to render a missing bid.

the intent of this is to make sure that (7) happens before (6) to keep the same behavior you'd get if you didn't enable cleanup (render the winning bid even if expired). Making it optional would mean getting a rendering error instead. We have https://github.com/prebid/Prebid.js/issues/11270 to improve the logic about rendering expired bids.

olafbuitelaar:
Ok this makes sense. this wouldn't be a problem for us as we do check this prior rendering a bid in our wrapper, but understand this is a more broad problem where such checks aren't performed. secondly we don't really do unfocused/invisible reloads, but i can imagine the edge case where somebody does, and still wants to keep tab on the memory consumption and clear expired bids even in invisible tab.

dgirardi:
AFAIK all the main browsers don't run anything while the page is out of focus (that's why the edge case happens, the rendering has to wait for focus to come back), so I'm not sure how you'd do background refreshes. Is there a way to keep the page running while out of focus?

patmmccann (commented):
LGTM

PR #12064, opened 2024-07-31
PubxAI Rtd module update: Make the endpoint call optional; read from local storage. 

pnhegde:
## Type of change

- [x] Bugfix
- [x] Feature

## Description of change
The endpoint param of the RTD module has been made optional. When the endpoint is not provided, look for available data in the session storage or in a global window object.

### DIFF
--- a/modules/pubxaiRtdProvider.js
+++ b/modules/pubxaiRtdProvider.js
@@ -115,7 +115,7 @@ export const fetchFloorRules = async (provider) => {
     } else {
       // Fetch from local storage
       try {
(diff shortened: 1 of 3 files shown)


patmmccann (approved):

zachsavishinsky:
@pnhegde @patmmccann I think this change may need to be rethought/reworked.  FLOORS_END_POINT never gets accessed unless you pass in an endpoint in the params, which then it still wouldn't use it because you pass in an endpoint manually.  This is breaking pubx rtd calls if the endpoint is missing (it's almost as if now it is NOT optional).

PR #12069, opened 2024-07-31
GumGum Bid Adapter: Send new  tpl paramter which is topmostLocation

MartinGumGum:
- [x ] Feature

## Description of change

We have introduced a new parameter named `tpl`, which captures the current page URL using `bidderRequest.refererInfo.topmostLocation `. Currently, the values returned in some cases are incorrect when using `pu = bidderRequest.refererInfo.page`. To address this issue, we need to include `bidderRequest.refererInfo.topmostLocation in the new `tpl` parameter to accurately validate the page URL values.


modules/gumgumBidAdapter.js
```diff
@@ -85,6 +85,7 @@ function _getBrowserParams(topWindowUrl) {
     sw: topScreen.width,
     sh: topScreen.height,
     pu: stripGGParams(topUrl),
+    wl: window.location.href || '',
```
patmmccann, line 88: What's wrong with other fields available in https://github.com/prebid/Prebid.js/blob/c0d565845dd8749b794807287cc28f5c5d902644/src/refererDetection.js#L238

patmmccann (changes requested):
Why not use the rest of the referral object?

modules/gumgumBidAdapter.js
MartinGumGum: This anomaly has only been noticed with the GG adapter with one of our pubs, we receive localhost as the page URL in some cases (can't provide in which exact cases) when we are getting pu = bidderRequest.refererInfo.page. This way we can also validate that window.location.href matches with what we get from refererInfo.

patmmccann: but you get that from this right? https://github.com/prebid/Prebid.js/blob/c0d565845dd8749b794807287cc28f5c5d902644/src/refererDetection.js#L243

MartinGumGum:
@patmmccann I have updated the parameter name to 'tpl' and getting it from 'topmostLocation' as you suggested above. Thank you!

patmmccann (approved):

PR #12074, opened 2024-08-01
Vidazoo lib and associated bid adapters: Add ORTB2 device data to request payload

jwrosewell:
## Type of change
- [x] Feature
- [x] Updated bidder adapter

## Description of change
This PR enhances bid requests formed within Vidazoo library by incorporating device data from the global ORTB2 object.

The device object has previously been populated by simplistic parsers, if at all, and was inaccurate as a result. Prebid now benefits from RTD modules such as 51Degrees that enrich all the device object fields including Apple iPhone model category and device ID. The PR enables users of Vidazoo and Vidazoo-dependent bid adapters to benefit from the device object.

Affected bid adapters:
- [illumin](https://github.com/prebid/Prebid.js/blob/master/modules/illuminBidAdapter.js)
- [kueezRtb](https://github.com/prebid/Prebid.js/blob/master/modules/kueezRtbBidAdapter.js)
- [shinezRtb](https://github.com/prebid/Prebid.js/blob/master/modules/shinezRtbBidAdapter.js)
- [tagoras](https://github.com/prebid/Prebid.js/blob/master/modules/tagorasBidAdapter.js)
- [twistDigital](https://github.com/prebid/Prebid.js/blob/master/modules/twistDigitalBidAdapter.js)
- [vidazoo](https://github.com/prebid/Prebid.js/blob/master/modules/vidazooBidAdapter.js)

## Other information
cc: @uditalias, @saar120


patmmccann:
This needs some work, as there is more redundant logic building the request

Eg https://github.com/prebid/Prebid.js/blob/23841fa3a2b9230106d790efe8f47b348a300ce1/libraries/vidazooUtils/bidderUtils.js#L296

patmmccann (approved):

PR #12082, opened 2024-08-02
MobianRtdModule: add mobian contextual variables directly to site.ext.data 

ehb-mtk:
## Type of change

- [x] Bugfix

## Description of change
Our client is using Magnite for importing k-v pairs from site.ext.data, and Magnite isn't able to reach into site.ext.data.mobian, so we are "bringing everything up a layer."


ehb-mtk:
We only have this one client (so, mathematically speaking, the answer to your question is "yes" 😄 )

patmmccann (approved):

PR #12089, opened 2024-08-05
Pstudio Bid Adapter : change Prebid Request Body Field

afifmahdi95:
## Type of change
- [x] Updated bidder adapter   

## Description of change
Change for P-Studio prebid body request field
- remove field 'floorPrice'
- add field 'adtagid'


modules/pstudioBidAdapter.js
```diff
@@ -417,7 +414,7 @@ function validateSizes(sizes) {
   );
 }
 
-function getBidFloor(bid) {
+/* function getBidFloor(bid) {
```
patmmccann, line 417: Why are you dropping support for floors?

afifmahdi95: Because we determine floor price on our internal SSP system based on ad tag id

patmmccann (approved):

PR #12095, opened 2024-08-06
Weborama RTD Module: BUGFIX on user-centric profile validation

peczenyj:
## Type of change

- [X] Bugfix

## Description of change

Unfortunately, the function that verifies the data that came from local storage is too restrictive and does not support the case when we have an array of numbers (ids) instead strings (labels)

### DIFF
--- a/modules/weboramaRtdProvider.js
+++ b/modules/weboramaRtdProvider.js
@@ -100,9 +100,7 @@
  * @typedef {WeboCtxConf|WeboUserDataConf|SfbxLiteDataConf} CommonConf
  */
 
-import {
-  getGlobal
-} from '../src/prebidGlobal.js';
+import { getGlobal } from '../src/prebidGlobal.js';
 import {
   deepAccess,
   deepClone,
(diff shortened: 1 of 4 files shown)


peczenyj:
I don't like to do this but ... @patmmccann do you think you can take a look? :)

patmmccann:
@peczenyj could you tag the changed cove with a GitHub comment for convenience lots of format changes

modules/weboramaRtdProvider.js
```diff
@@ -921,7 +977,18 @@ export function isValidProfile(profile) {
     return false;
   }
 
-  return Object.values(profile).every((field) => isArray(field) && field.every(isStr));
+  return Object.values(profile).every(
+    (field) => isArray(field) && field.every(isStrOrNumber)
```
peczenyj, line 981: That is the most important change: because sometimes the profile could contain not only strings but numbers, it end return false in some scenarions

modules/weboramaRtdProvider.js
```diff
@@ -1022,9 +1105,12 @@ function getDataFromLocalStorage(weboDataConf, cacheGet, cacheSet, defaultLocalS
...
         const valid = isValidProfile(profile);
         if (!valid) {
-          logWarn(`found invalid ${source} profile on local storage key ${localStorageProfileKey}, section ${targetingSection}`);
+          logger.logMessage(
+            `WARNING: found invalid ${source} profile on local storage key ${localStorageProfileKey}, section ${targetingSection}: `,
+            profile
+          );
 
-          return;
+          return [defaultProfile, true];
```
peczenyj, line 1113: also, when it considers the profile is not valid, the empty return had a negative impact in the rest of the code, now we make sure that we use the default profile and we log correctly (not using warning, but the standard log message, to be possible hide the message via configuration)

modules/weboramaRtdProvider.md
```diff
@@ -546,15 +546,15 @@ pbjs.que.push(function () {
...
-                            'appnexus': true,               // all adunits for appnexus 
+                            'appnexus': true,               // all adunits for appnexus
                             'pubmatic': ['adUnitCode1',...] // some adunits for pubmatic
                             // other bidders will be ignored
                         },
                         defaultProfile: {           // optional
-                            webo_cs: ['Red'],
-                            webo_audiences: ['bam']
+                            webo_cs: ['Red'],       // using label
+                            webo_audiences: [12345] // using id
```
peczenyj, line 557: one example of test changed, we verify if a numeric audience still works. most of the tests use a string on webo_cs (stands for custom segment) as string and audience as integer but in one test we use the opposite just in case.

test/spec/modules/weboramaRtdProvider_spec.js
```diff
@@ -2377,9 +2377,85 @@ describe('weboramaRtdProvider', function() {
         })
       });
 
+      it('should use default profile in case of something malformed on local storage', function() {
```
peczenyj, line 2380: this is a test to verify the impact of the old empty return, now fixed

peczenyj:
> @peczenyj could you tag the changed cove with a GitHub comment for convenience lots of format changes

done

patmmccann (approved):

PR #11900, opened 2024-07-01
symitriDapRtdProvider : initial release and update akamaiDapRTD to import from symitri

jpalladino84:
## Type of change
- [x] New bidder adapter  
- [x] Updated bidder adapter  

- [x] Does this change affect user-facing APIs or examples documented on http://prebid.org?

## Description of change
- Add symitriDapRtdProvider.js
- Add symitriDapRtdProvider.md
- Update akamaiDapRtdProvider.js to import from symitriDapRtdProvider.js 

## Other information
- Update document on Prebid.org to reflect new module
 https://github.com/prebid/prebid.github.io/pull/5469

PR #12097, opened 2024-08-06
symitriDap Rtd Provider: enable Secure Ad ID receipt and identifier ingestion

jpalladino84:
## Type of change
- [X] Feature

## Description of change
- Support for publishers to pass identifiers either through CSS selector or directly to module
- Hook on BidWon to record SAID receipt

### DIFF
--- a/modules/symitriDapRtdProvider.js
+++ b/modules/symitriDapRtdProvider.js
@@ -130,6 +130,31 @@ export function createRtdProvider(moduleName, moduleCode, headerPrefix) {
   }
 
   /**
+   * Listen for bidWon Events to record SAID receipt
+   * @param {string} authToken
+   */
+  function onBidWonListener (authToken) {
+    window.pbjs.onEvent('bidWon', function(bid) {
+      const url = 'https://ProdSymPrebidEventhub1.servicebus.windows.net/prebid-said-1/messages';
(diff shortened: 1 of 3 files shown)


jpalladino84:
Hi @patmmccann would it be possible to have access to the ci/circleci build? The tests and builds are passing locally for us, but the tests still fail on CircleCI and we cannot see why.

bretg:
@jpalladino84 - name changes are something that needs to be done cautiously. Generally we don't approve them - there needs to be a transition period where publishers using your service don't have to change their pages when they upgrade from 9.x to 9.x+1.  Publisher breaking changes **must** be only in major releases, e.g. 10.0.

So you need to find a way to support both 'dap' and 'symitriDap' until 10.0. e.g. leave the 'dap' module that just includes symitriDap.

jpalladino84:
Hi @bretg! Thank you for that feedback! We actually do account for this in our last PR where we allow for both our old POC adapter to be initialized with `dap`: https://github.com/prebid/Prebid.js/pull/11900/files#diff-cb8c6aa2c41236cb4721c6c232fffa84dba48338c5966f3f3248b1cfae797476R27

The rename here is just to update the new modules tests and documentation. See here for the Prebid doc changes: https://github.com/prebid/prebid.github.io/pull/5533

PR #12102, opened 2024-08-07
Various bid adapters: Currency config cleanup

mkomorski:
## Type of change

- [ ] Bugfix
- [ ] Feature
- [ ] New bidder adapter  
- [x] Updated bidder adapter   
- [ ] Build related changes
- [ ] CI related changes

- [ ] Other

## Other information

#10656

### DIFF
--- a/modules/beopBidAdapter.js
+++ b/modules/beopBidAdapter.js
@@ -12,11 +12,7 @@ import {
   logWarn,
   triggerPixel
 } from '../src/utils.js';
-import {getRefererInfo} from '../src/refererDetection.js';
-import {registerBidder} from '../src/adapters/bidderFactory.js';
 import { getStorageManager } from '../src/storageManager.js';
-import {config} from '../src/config.js';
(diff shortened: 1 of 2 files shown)


libraries/ortb2Utils/currency.js
```diff
@@ -0,0 +1,15 @@
+const additionalData = new WeakMap();
+
+export function getCurrencyFromBidderRequest(bidderRequest) {
+  return (bidderRequest?.ortb2?.ext?.cur || [])[0];
+}
+
+export function setAdditionalData(obj, key, value) {
```
dgirardi, line 7: if these are only used for adserverCurrency, they'd be better as `set/getAdServerCurrency` (they are in an currency library after all, and unlikely to be used for anything else).

But if they are only used for one adapter, I don't think they're needed at all - the adapter itself can maintain the map.

dgirardi (approved):

ChrisHuie (approved):

Issue #11395, opened 2024-04-24
Activity controls: transmit ip using geo activity, document publisher provided IP address


jdwieland8282:
## Type of issue
Feature request

## Description
Google Chrome has announced, and is actively developing, a list-based, two-hop HTTP request proxy scheme in the Privacy Sandbox [IP Protection](https://github.com/GoogleChrome/ip-protection) feature. This is similar to Apple’s iCloud Private Relay. “Limiting access to IP addresses,” as IP Protection describes it, has the potential to degrade IVT/fraud evaluation and targeting use cases.

Publishers, who as first-parties should not be restricted by Chrome’s feature, may opt to share a visitor’s IP address with trusted partners. 

### Expected results
Publishers can configure on/off adding the device's ip to `ortb.device.ip`

```
pbjs.mergeConfig({
   ortb2: {
      device: {
        ip: '1.1.1.1'
           /* or */
        ipv6: '2001:db8:3333:4444:5555:6666:7777:8888'
      }
   }
});
```

## Other information
@pm-harshad-mane has offered some server side code that can be used to return the clients IP 
https://github.com/pm-harshad-mane/cloudflare-ip-example

patmmccann:
I'll make an effort to discuss in documentation and include consent considerations

AramZS:
Two major problems here: 

1. If the user is leveraging a service that obscures their IP but somehow the publisher has access to that IP why would counteracting the user's intent make sense? 
2. Where has it been indicated that the publisher would have special access to the IP? I haven't seen that yet and it doesn't make a ton of sense for anyone to do. 

That said, transmitting an IP address across network requests seems to be a HUGE security challenge the minute the bidstream departs the hands of those with immediate access to the network request.

AramZS:
Will also note here that transmitting the IP address would be blocked by user opt outs in California and we'd need to be sure that Activity Controls could restrict that behavior on that basis.

AramZS:
Just adding an additional note that there are now new state laws that identify IP as PII specifically.

bmayd:
> there are now new state laws that identify IP as PII specifically
Can you provide references to specific sources?

mkendall07:
as an FYI - it doesn't look like clarifications have been made if subdomain is considered 3p in this context (https://github.com/GoogleChrome/ip-protection/issues/13) will have an impact on workarounds for pubs.

patmmccann:
[...]

PR #12107, opened 2024-08-08
Core: Truncating IPs using geo activity 

mkomorski:
## Type of change

- [ ] Bugfix
- [x] Feature
- [ ] New bidder adapter  
- [ ] Updated bidder adapter   
- [ ] Build related changes
- [ ] CI related changes

- [ ] Other

## Other information

#11395


src/activities/redactor.js
```diff
@@ -154,7 +155,13 @@ export function ortb2TransmitRules(isAllowed = isActivityAllowed) {
       paths: ORTB_GEO_PATHS,
       applies: appliesWhenActivityDenied(ACTIVITY_TRANSMIT_PRECISE_GEO, isAllowed),
       get(val) {
-        return Math.round((val + Number.EPSILON) * 100) / 100;
+        switch (typeof val) {
+          case 'string':
```
dgirardi, line 159: I think it's better to add another, separate, rule for ip4 and another for ipv6 fields. They have different ORTB paths so there's no need to inspect them to figure out which is which here.

libraries/ipUtils/ipUtils.js
```diff
@@ -0,0 +1,30 @@
+export function scrubIP(ip) {
+  if (!ip) {
+    return '';
```
dgirardi, line 3: would this transform nulls to empty strings? it might be a good test case.

libraries/ipUtils/ipUtils.js
```diff
@@ -0,0 +1,30 @@
+export function scrubIP(ip) {
```
dgirardi, line 1: since this is used by core, it makes more sense to include it in core. (a library used by core would be included in every bundle regardless).

test/spec/unit/utils/ipUtils_spec.js
```diff
@@ -0,0 +1,58 @@
...
+      input = '192.168.255.255';
+      output = scrubIPv4(input);
+      expect(output).to.deep.equal('192.168.255.0');
+    });
+
+    it('should convert null to blank string', () => {
+      let input = null;
+      let output = scrubIPv4(input);
+      expect(output).to.deep.equal('');
+    });
```
dgirardi, line 18: not sure if this was prompted by my comment - I meant almost the opposite, a null ip should not be transformed into an empty string (since an empty string is not a valid ip).

It doesn't matter for this because the redaction rule [only runs when the field is non-null](https://github.com/prebid/Prebid.js/blob/e0651f657bd6d1a3dfa8053df007d97cf8034b1d/src/activities/redactor.js#L53).  However if we end up re-using the scrubbing utilities in the future, I think this will prove to be an odd choice. I'll leave it to you whether you want to change it. (IMO returning null or throwing would both be better).

patmmccann (approved):

Issue #9951, opened 2023-05-15
Feature request: populate first party data with user EIDs


dgirardi:
## Type of issue
Feature request

## Description

Currently bidders are expected to pick up EIDs from `bidRequest.userIdAsEids`. We should make them available also under `bidderRequest.ortb2.user.ext.eids`; making sure the data controller module continues to filter them.

Issue #11954, opened 2024-07-10
User id module overrides ortb2.user.ext.eids content


leonardlabat:
## Type of issue
Bug

## Description
If a publisher uses the `setBidderConfig` API to set manually some identifier in the `ortb2.user.ext.eids` structure, and if the publisher also uses some user id module, then unfortunately the bidder adapters will get only the eids resolved by the user id module.

Most likely this is because this row https://github.com/prebid/Prebid.js/blob/master/modules/userId/index.js#L1199 overrides the potential existing value instead of merging the eids.

## Steps to reproduce
1. Call `setBidderConfig` and set some eids under `ortb2.user.ext.eids` structure
2. Set some user id module on the page
3. Look at the ortb2 request received by the bidder adapter

### Expected results
My bidder adapter is called with a request that contains both the eids set through the `setBidderConfig` call and the eids resolved by the user id module.

### Actual results
My bidder adapter is called with a request that contains only the eids resolved by the user id module.

dgirardi:
The line you linked affects only some adapters (those that use shared ortb2 conversion utilities). In general this needs to be done adapter by adapter.

A workaround is to use [pubProvidedId](https://docs.prebid.org/dev-docs/modules/userid-submodules/pubprovided.html), which is essentially a userId wrapper around eids. However it only allows a single set of EIDs which can be restricted to some bidders, not different EIDs for different bidders.

Related to https://github.com/prebid/Prebid.js/issues/9951 and the deprecation of `bidRequest.userId` mentioned in the [9.0 release notes](https://docs.prebid.org/dev-docs/pb9-notes.html#miscellaneous-deprecation-notices-module-maintainers). 

Proposal:

1. Update ortbConverter as a stopgap, which may be enough if you're only dealing with adapters that use it.
2. Make eids part of first party data (https://github.com/prebid/Prebid.js/issues/9951), appending and not overriding pub-provided EIDs
3. Bidders should then stop using bid requests' `userId/userIdAsEids` and pick up `ortb2.user.ext.eids` instead
4. Remove both `userId` and `userIdAsEids` from bid requests for prebid 10.

patmmccann:
@dgirardi is step 1 above simply to merge instead of set in the line @leonardlabat mentions? 

Of note @leonardlabat is an adapter maintainer of grid and criteo adapters

FlorentDancy:
> Is setBidderConfig popular, is this hypothetical, or something in between?

[...]

PR #12110, opened 2024-08-08
UserID: merge EIDs with first party data

dgirardi:
## Type of change

- [x] Feature

## Description of change

This merges `bidRequest.userIdAsEids` with `bidderRequest.ortb2.user.ext.eids`:

 - eids provided as first party data are included in `userIdAsEids`
 - eids provided by userId modules are included in `bidderRequest.ortb2.user.ext.eids`
 - the PBS adapter provides bidder-specific EIDS as `bidderconfig` (instead of setting `eidpermissions`)

## Other information
Closes https://github.com/prebid/Prebid.js/issues/9951
Related to https://github.com/prebid/Prebid.js/issues/11954


patmmccann:
Let's make sure we document ongoing use goals of of ppid module:

https://github.com/prebid/Prebid.js/blob/master/modules/pubProvidedIdSystem.js

handles storage, handles publisher purpose consent, handles bidders without eids support.

We also propose a new activity check: if tcfControl module has a flag, we would want to make sure that directly set ufpd and eids are checked for publisher purposes in the same way ppid is now, so publishers are not accidentally circumventing consent.

ncolletti (approved):
LGTM, lots of helpful changes in the user id system

ncolletti:
@dgirardi nice work! should `primaryIds` be documented anywhere in prebid.github.io for new id modules?

Issue #12125, opened 2024-08-14
adUnitCodes in auctionEnd event are duplicated for ephemeral


muuki88:
## Type of issue

Bug

## Description

The `pbjs.requestBids()` method can be called in two different ways.
In both examples I make use of the [twin ad unit feature](https://docs.prebid.org/dev-docs/adunit-reference.html#twin-adunit-codes).

### Add ad units and request adUnitCodes

```js
// create ad units
pbjs.addAdUnits([
  { code: 'div-1', /* ... */ },
  { code: 'div-1', /* ... */ },
]);

// request bids for the created ad units
pbjs.requestBids({adUnitCodes: ['div-1']});
```

With this call structure, the `adUnitCodes` array in the `auctionEnd` event has only the ad unit codes from the array.

### Ephemeral ad units

```js
// request bids 
pbjs.requestBids({
  adUnits: [
    { code: 'div-1', /* ... */ },
    { code: 'div-1', /* ... */ },
  ]
});
```

With this call structure, the `adUnitCodes` array in the `auctionEnd` event has duplicated `code` s .

## Steps to reproduce

Use Ephemeral ad units variant from the description and

```js
pbjs.getEvents().filter(event => event.eventType === 'auctionEnd')
```

and look the `adUnitCodes` array.

### Expected results

`adUnitCodes` are always unique.

### Actual results

`adUnitCodes` are unique only if explicitly specified in the `adUnitCodes` requestBids parameter.

## Platform details

- prebid 8.52.1
- prebid 9.8.0

## Other information

What is the behaviour if `adUnitCodes` and `adUnits` are provided as params?

patmmccann:
`With this call structure, the adUnitCodes array in the auctionEnd event has duplicated code s .`

This is a bug we can fix.

PR #12127, opened 2024-08-14
Core: make sure adUnitCodes are unique in auction events

dgirardi:
## Type of change

- [x] Bugfix

## Description of change

Fix https://github.com/prebid/Prebid.js/issues/12125 : in some situations, AUCTION_INIT and AUCTION_END event can contain duplicate codes in `adUnitCodes`.

### DIFF
--- a/src/prebid.js
+++ b/src/prebid.js
@@ -503,6 +503,9 @@ pbjsInstance.requestBids = (function() {
     events.emit(REQUEST_BIDS);
     const cbTimeout = timeout || config.getConfig('bidderTimeout');
     logInfo('Invoking $$PREBID_GLOBAL$$.requestBids', arguments);
+    if (adUnitCodes != null && !Array.isArray(adUnitCodes)) {
+      adUnitCodes = [adUnitCodes];
+    }
     if (adUnitCodes && adUnitCodes.length) {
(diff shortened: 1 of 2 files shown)


muuki88 (approved):
Thanks for the quick fix :heart:

src/prebid.js
```diff
@@ -503,13 +503,17 @@ pbjsInstance.requestBids = (function() {
     events.emit(REQUEST_BIDS);
     const cbTimeout = timeout || config.getConfig('bidderTimeout');
     logInfo('Invoking $$PREBID_GLOBAL$$.requestBids', arguments);
+    if (adUnitCodes != null && !Array.isArray(adUnitCodes)) {
+      adUnitCodes = [adUnitCodes];
```
muuki88, line 507: Was this another bug?

dgirardi: yes, but it's weirder than it looks. we are filtering ad units by `adUnitCodes.includes(adUnit.code)`, and strings can include each other, so passing `adUnitCodes: 'some-code'` would include `some-code` as expected, but also `some` and `code`.

muuki88: Oh 🤯that's a good catch

ChrisHuie (approved):

PR #10488, opened 2023-09-14
userId: group UIDs by source when possible

dgirardi:
## Type of change

- [x] Bugfix

## Description of change

Prebid Server requires `user.eids` to be [grouped by source](https://github.com/prebid/prebid-server/issues/3097). This updates the userID module to accomodate it when possible (all EIDs for a given source have the same values in `ext`).

## Other information

Fixes https://github.com/prebid/Prebid.js/issues/10479

Issue #10479, opened 2023-09-13
Pair Id Module: generates multiple eids with same source


patmmccann:
## Type of issue
Bug

## Description
Google Pair module produces lots of distinct eids with the same source. They should be in a single object according to https://github.com/InteractiveAdvertisingBureau/openrtb/blob/master/extensions/2.x_official_extensions/eids.md

cc: @congdu-kun 

## Test page
Hard to reproduce, you have to have a lot of retargeting segments and be a Chrome user I think
firstquarterfinance.com is a site with the Pair module installed

### Expected results

![image](https://github.com/prebid/Prebid.js/assets/1683175/6c3d751c-8f51-4734-bf0f-0cb3809082d3)

### Actual results

![image](https://github.com/prebid/Prebid.js/assets/1683175/070b98d6-ced1-4fc2-842b-003559818427)

This request is then rejected by PBS

dgirardi:
> They should be in a single object according to https://github.com/InteractiveAdvertisingBureau/openrtb/blob/master/extensions/2.x_official_extensions/eids.md

I cant' think of any way to read that spec that would warrant that conclusion. That spec does not say that there can be only one EID object per source, or that source must be unique across EID objects. What am I missing?

PR #103, opened 2015-11-12
Send referrer in /jpt call

hjeong12:

PR #12146, opened 2024-08-20
PAIR: Support for Generic TechLab Version

therevoltingx:
# Clean Room Configuration
```
{
    userSync: {
      userIds: [{
        name: 'pairId',
        params: {
          mode: "open",
          liveramp: {
            storageKey: '_lr_pairId'
          },
          habu: {
            storageKey: '_habu_pairId'
          }
       },
    }]
  }
}
```

# Bid Request
```
{
  "ext": {
    "eids": [
      "source": "iabtechlab.com",
      "uids": [
        {
          "id": "0x1234", // using clean room A
          "atype": 3
        },
        {
          "id": "0x456", // using clean room B
          "atype": 3
        }
      ]
    ]
  }
}
```

# References 
- https://github.com/InteractiveAdvertisingBureau/openrtb/blob/main/extensions/2.x_official_extensions/eids.md#object--extended-identifier-uids-
- https://github.com/prebid/Prebid.js/issues/10479
- https://github.com/prebid/Prebid.js/pull/10488

### DIFF
--- a/modules/pairIdSystem.js
+++ b/modules/pairIdSystem.js
@@ -16,7 +16,10 @@ import {MODULE_TYPE_UID} from '../src/activities/modules.js';
 
 const MODULE_NAME = 'pairId';
 const PAIR_ID_KEY = 'pairId';
-const DEFAULT_LIVERAMP_PAIR_ID_KEY = '_lr_pairId';
+
+const DEFAULT_STORAGE_PAID_ID_KEYS = {
+  liveramp: '_lr_pairId'
+};
 
 export const storage = getStorageManager({moduleType: MODULE_TYPE_UID, moduleName: MODULE_NAME});
 
@@ -36,11 +39,6 @@ export const pairIdSubmodule = {
    */
   name: MODULE_NAME,
   /**
-   * used to specify vendor id
-   * @type {number}
-   */
-  gvlid: 755,
-  /**
    * decode the stored id value for passing to bid requests
    * @function
    * @param { string | undefined } value
@@ -59,6 +57,7 @@ export const pairIdSubmodule = {
   getId(config) {
     const pairIdsString = pairIdFromLocalStorage(PAIR_ID_KEY) || pairIdFromCookie(PAIR_ID_KEY)
     let ids = []
+
     if (pairIdsString && typeof pairIdsString == 'string') {
       try {
         ids = ids.concat(JSON.parse(atob(pairIdsString)))
@@ -68,13 +67,19 @@ export const pairIdSubmodule = {
     }
 
     const configParams = (config && config.params) || {};
-    if (configParams && configParams.liveramp) {
-      let LRStorageLocation = configParams.liveramp.storageKey || DEFAULT_LIVERAMP_PAIR_ID_KEY;
-      const liverampValue = pairIdFromLocalStorage(LRStorageLocation) || pairIdFromCookie(LRStorageLocation);
+    const cleanRooms = Object.keys(configParams);
 
-      if (liverampValue) {
+    for (let i = 0; i < cleanRooms.length; i++) {
+      const cleanRoom = cleanRooms[i];
+      const cleanRoomParams = cleanRoom[i];
+
+      const cleanRoomStorageLocation = cleanRoomParams.storageKey || DEFAULT_STORAGE_PAID_ID_KEYS[cleanRoom];
+      const cleanRoomValue = pairIdFromLocalStorage(cleanRoomStorageLocation) || pairIdFromCookie(cleanRoomStorageLocation);
+
+      if (cleanRoomValue) {
         try {
-          const parsedValue = atob(liverampValue);
+          const parsedValue = atob(cleanRoomValue);
+
           if (parsedValue) {
             const obj = JSON.parse(parsedValue);
 
@@ -90,7 +95,7 @@ export const pairIdSubmodule = {
           logInfo('Pairid: Error parsing JSON: ', error);
         }
       } else {
-        logInfo('Pairid: liverampValue for pairId from storage is empty or null');
+        logInfo('Pairid: data clean room value for pairId from storage is empty or null');
       }
     }
 
@@ -98,11 +103,12 @@ export const pairIdSubmodule = {
       logInfo('PairId not found.')
       return undefined;
     }
+
     return {'id': ids};
   },
   eids: {
     'pairId': {
-      source: 'google.com',
+      source: 'iabtechlab.com',
       atype: 571187
     },
   }
--- a/modules/pairIdSystem.js
+++ b/modules/pairIdSystem.js
@@ -39,11 +39,6 @@ export const pairIdSubmodule = {
    */
   name: MODULE_NAME,
   /**
-   * used to specify vendor id
-   * @type {number}
-   */
-  gvlid: 755,
-  /**
    * decode the stored id value for passing to bid requests
    * @function
    * @param { string | undefined } value
@@ -114,7 +109,7 @@ export const pairIdSubmodule = {
   eids: {
     'pairId': {
       source: 'iabtechlab.com',
-      atype: 571187
+      atype: 3
     },
   }
 };
--- a/modules/pairIdSystem.js
+++ b/modules/pairIdSystem.js
@@ -39,6 +39,11 @@ export const pairIdSubmodule = {
    */
   name: MODULE_NAME,
   /**
+   * used to specify vendor id
+   * @type {number}
+   */
+  gvlid: 755,
+  /**
    * decode the stored id value for passing to bid requests
    * @function
    * @param { string | undefined } value
--- a/modules/pairIdSystem.js
+++ b/modules/pairIdSystem.js
@@ -71,12 +71,12 @@ export const pairIdSubmodule = {
       }
     }
 
-    const configParams = (config && config.params) || {};
+    const configParams = (config && config.params) ? config.params : {};
     const cleanRooms = Object.keys(configParams);
 
     for (let i = 0; i < cleanRooms.length; i++) {
       const cleanRoom = cleanRooms[i];
-      const cleanRoomParams = cleanRoom[i];
+      const cleanRoomParams = configParams[cleanRoom];
 
       const cleanRoomStorageLocation = cleanRoomParams.storageKey || DEFAULT_STORAGE_PAID_ID_KEYS[cleanRoom];
       const cleanRoomValue = pairIdFromLocalStorage(cleanRoomStorageLocation) || pairIdFromCookie(cleanRoomStorageLocation);
--- a/test/spec/modules/pairIdSystem_spec.js
+++ b/test/spec/modules/pairIdSystem_spec.js
@@ -13,6 +13,39 @@ describe('pairId', function () {
     sandbox.restore();
   });
 
+  it('should read pairId from specified clean room if configured with storageKey', function() {
+    let pairIds = ['test-pair-id1', 'test-pair-id2', 'test-pair-id3'];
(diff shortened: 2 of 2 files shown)


patmmccann:
your general approach is breaking and unlikely to be merged. I would recommend chatting before continuing

modules/pairIdSystem.js
```diff
@@ -90,19 +95,20 @@ export const pairIdSubmodule = {
...
       logInfo('PairId not found.')
       return undefined;
     }
+
     return {'id': ids};
   },
   eids: {
     'pairId': {
-      source: 'google.com',
+      source: 'iabtechlab.com',
```
pm-harshad-mane, line 111: In future IAB may support another Alt ID, so the source for PAIR should clearly mention what it is. 
Can we mention it as "pair-iab-tl.com"?

therevoltingx: isn't that what atype is for? like a single source can have multiple id types?

jdwieland8282: Here are the existing `atype` definitions https://github.com/InteractiveAdvertisingBureau/AdCOM/blob/main/AdCOM%20v1.0%20FINAL.md#list_agenttypes

modules/pairIdSystem.js
```diff
@@ -35,11 +38,6 @@ export const pairIdSubmodule = {
    * @type {string}
    */
   name: MODULE_NAME,
-  /**
-   * used to specify vendor id
-   * @type {number}
-   */
-  gvlid: 755,
```
therevoltingx, line 42: Make this configurable. Should be the DSP

modules/pairIdSystem.js
therevoltingx, line 111: what is this? I assumed it was the ID type which is now a techlab id type

modules/pairIdSystem.js
```diff
@@ -90,19 +95,20 @@ export const pairIdSubmodule = {
...
       return undefined;
     }
+
     return {'id': ids};
   },
   eids: {
     'pairId': {
-      source: 'google.com',
+      source: 'iabtechlab.com',
       atype: 571187
```
therevoltingx, line 112: confirm that this value was assigned to pair protocol

modules/pairIdSystem.js
pm-harshad-mane: atype identifies the user agent types a user identifier is from.

pm-harshad-mane: @therevoltingx mentioned we will have a unique atype to identify eid as a PAIR

pm-harshad-mane:
Sorry for the late update :-)

The code in the original pairIdSystem and the new module (which will have a PR created) will have a lot of duplication.

We talked about needing to change some parameter values, like switching the source from “google.com” to “iabtechlab.com” and updating or removing the GVL.

To avoid duplicating code, we can add a new optional “mode” parameter to the module. This parameter will adjust these values based on its setting.

If the “mode” parameter isn’t specified, the module will work the same as before with the original values (backward compatible). However, if the “mode” is set to “open,” the module will slightly change its behavior, like altering the source parameter to "iabtechlab.com" and the GVL reference.

modules/pairIdSystem.js
therevoltingx: this is not included in the bid request at all. this is the owner of the module. we need prebid to confirm. 
the prefered thing to do is to leave as-is.

therevoltingx:
> Sorry for the late update :-)
> 
> The code in the original pairIdSystem and the new module (which will have a PR created) will have a lot of duplication.
> 
> We talked about needing to change some parameter values, like switching the source from “google.com” to “iabtechlab.com” and updating or removing the GVL.
> 
> To avoid duplicating code, we can add a new optional “mode” parameter to the module. This parameter will adjust these values based on its setting.
> 
> If the “mode” parameter isn’t specified, the module will work the same as before with the original values (backward compatible). However, if the “mode” is set to “open,” the module will slightly change its behavior, like altering the source parameter to "iabtechlab.com" and the GVL reference.

I really like the idea. But for the eids method, how can I access the configuration? Gonna dig in but if you could point me to how I can access the config there that would be very helpful

pm-harshad-mane:
@therevoltingx the `getId` method has access t o `config` and it is returning the id in eids format.
Which eids method are you referring to?

therevoltingx:
@pm-harshad-mane I'm referring to line #103 in pairIdSystem.js. You can see the line here: https://github.com/prebid/Prebid.js/blob/master/modules/pairIdSystem.js#L103

The eids module attribute is called by modules/userId/eids.js: https://github.com/prebid/Prebid.js/blob/master/modules/userId/eids.js#L67

Which helps construct the final bid request object. That does not seem to have access to `config`.

pm-harshad-mane:
@dgirardi can you help us? we are trying to avoid code duplication

jlquaccia:
hey @therevoltingx @pm-harshad-mane, was taking a look into the code around the prebid pair module a bit from my end.  what about something like this?

pbjs.setConfig:

```
userSync: {
  userIds: [{
    name: "pairId",
    storage: {
      type: "html5",
      name: "pairId"
    },
    params: {
      mode: "open"
    }
  }],
}
```

and then the following modifications to the pairIdSystem.js file (which could be further modified as needed):

```
/**
 * This module adds PAIR Id to the User ID module
 * The {@link module:modules/userId} module is required
 * @module modules/pairIdSystem
 * @requires module:modules/userId
 */

import {getStorageManager} from '../src/storageManager.js'
import { logInfo } from '../src/utils.js';
import {MODULE_TYPE_UID} from '../src/activities/modules.js';

/**
 * @typedef {import('../modules/userId/index.js').Submodule} Submodule
 */

const MODULE_NAME = 'pairId';
const PAIR_ID_KEY = 'pairId';
const DEFAULT_LIVERAMP_PAIR_ID_KEY = '_lr_pairId';

const MODE_CONFIG = {
  default: {
    source: 'google.com',
    gvlid: 755
  },
  open: {
    source: 'iabtechlab.com',
    gvlid: 999
  },
};

export const storage = getStorageManager({moduleType: MODULE_TYPE_UID, moduleName: MODULE_NAME});

function pairIdFromLocalStorage(key) {
  return storage.localStorageIsEnabled() ? storage.getDataFromLocalStorage(key) : null;
}

function pairIdFromCookie(key) {
  return storage.cookiesAreEnabled() ? storage.getCookie(key) : null;
}

/** @type {Submodule} */
export const pairIdSubmodule = {
  /**
   * used to link submodule with config
   * @type {string}
   */
  name: MODULE_NAME,
  /**
   * used to specify vendor id
   * @type {number}
   */
  gvlid: MODE_CONFIG.default.gvlid,
  /**
   * decode the stored id value for passing to bid requests
   * @function
   * @param { string | undefined } value
   * @returns {{pairId:string} | undefined }
   */
  decode(value) {
    return value && Array.isArray(value) ? {'pairId': value} : undefined
  },
  /**
   * Performs action to obtain ID and return a value in the callback's response argument.
   * @function getId
   * @param {Object} config - The configuration object.
   * @param {Object} config.params - The parameters from the configuration.
   * @returns {{id: string[] | undefined}} The obtained IDs or undefined if no IDs are found.
   */
  getId(config) {
    const mode = (config && config.params && config.params.mode) || 'default';
    const selectedModeConfig = MODE_CONFIG[mode] || MODE_CONFIG.default;
    const source = selectedModeConfig.source;

    this.gvlid = selectedModeConfig.gvlid;
    this.eids.pairId.source = source;

    const pairIdsString = pairIdFromLocalStorage(PAIR_ID_KEY) || pairIdFromCookie(PAIR_ID_KEY)
    let ids = []
    if (pairIdsString && typeof pairIdsString == 'string') {
      try {
        ids = ids.concat(JSON.parse(atob(pairIdsString)))
      } catch (error) {
        logInfo(error)
      }
    }

    const configParams = (config && config.params) || {};
    if (configParams && configParams.liveramp) {
      let LRStorageLocation = configParams.liveramp.storageKey || DEFAULT_LIVERAMP_PAIR_ID_KEY;
      const liverampValue = pairIdFromLocalStorage(LRStorageLocation) || pairIdFromCookie(LRStorageLocation);

      if (liverampValue) {
        try {
          const parsedValue = atob(liverampValue);
          if (parsedValue) {
            const obj = JSON.parse(parsedValue);

            if (obj && typeof obj === 'object' && obj.envelope) {
              ids = ids.concat(obj.envelope);
            } else {
              logInfo('Pairid: Parsed object is not valid or does not contain envelope');
            }
          } else {
            logInfo('Pairid: Decoded value is empty');
          }
        } catch (error) {
          logInfo('Pairid: Error parsing JSON: ', error);
        }
      } else {
        logInfo('Pairid: liverampValue for pairId from storage is empty or null');
      }
    }

    if (ids.length == 0) {
      logInfo('PairId not found.')
      return undefined;
    }

    return {'id': ids};
  },
  eids: {
    'pairId': {
      source: MODE_CONFIG.default.source,
      atype: 571187
    },
  }
};

submodule('userId', pairIdSubmodule);

```

jlquaccia:
@dgirardi wondering what your thoughts on the suggestion above might be?  seems to work fine locally, but curious if there are any other elements i am not considering..

therevoltingx:
@jlquaccia 
My main concern here is that the caller must call getId() first in order for eids to return the correct values.

bmilekic:
> @dgirardi wondering what your thoughts on the suggestion above might be? seems to work fine locally, but curious if there are any other elements i am not considering..

For the open version, I think it would be better to use a correct `atype` (atype: 3) as shown in the Bid Request section at the top of [here](https://github.com/prebid/Prebid.js/pull/12146#issue-2475242977) instead of the magic number 571187. This assumes that `source: "iabtechlab.com"` is used exclusively to signal a publisher open PAIR ID, which I believe is the intention.

pm-harshad-mane:
Looks like the progress to have a sub-module to support open version of PAIR has derailed since my suggestion for avoiding the code duplication...
@therevoltingx considering the complexities that we will need to handle, I have no issues if we create a separate module...

therevoltingx:
@pm-harshad-mane @bmilekic @jlquaccia 

After some internal discussion we've gone aheand and decided not to make it backward compatible. We also chose to remove the gvlid as that didn't make much sense to have.

Please review and let me know your thoughts.

modules/pairIdSystem.js
therevoltingx: use `VENDORLESS_GVLID` here

modules/pairIdSystem.js
```diff
@@ -67,14 +66,20 @@ export const pairIdSubmodule = {
...
+    const configParams = (config && config.params) ? config.params : {};
+    const cleanRooms = Object.keys(configParams);
 
-      if (liverampValue) {
+    for (let i = 0; i < cleanRooms.length; i++) {
+      const cleanRoom = cleanRooms[i];
+      const cleanRoomParams = configParams[cleanRoom];
+
+      const cleanRoomStorageLocation = cleanRoomParams.storageKey || DEFAULT_STORAGE_PAID_ID_KEYS[cleanRoom];
+      const cleanRoomValue = pairIdFromLocalStorage(cleanRoomStorageLocation) || pairIdFromCookie(cleanRoomStorageLocation);
```
muuki88, line 77:
```suggestion
    Object.entries(configParams).forEach( ([cleanRoom, cleanRoomParams]) => { 

      const cleanRoomStorageLocation = cleanRoomParams.storageKey || DEFAULT_STORAGE_PAID_ID_KEYS[cleanRoom];
      const cleanRoomValue = pairIdFromLocalStorage(cleanRoomStorageLocation) || pairIdFromCookie(cleanRoomStorageLocation);
```

JS made this a little more convienent :smile: 
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/entries

Suggestion won't fully work as there's probably a closing brace missing

Issue #6345, opened 2021-02-22
Enforcement in Meta Object


gglas:
This is a feature request for added functionality within the meta object of bidresponses.

As a result of https://github.com/prebid/Prebid.js/issues/3115 , we've implemented a standardized taxonomy within the Meta object in order to make gathering things like adomain and exchange provided metadata about their bidresponses easier to 
ingest and index in the course of running a prebid auction.

Now, the inevitable next step is to take these values and allow publishers to apply logic across them.  Currently available meta fields include : 

```
meta: {
            networkId: NETWORK_ID,
            networkName: NETWORK_NAME
            agencyId: AGENCY_ID,
            agencyName: AGENCY_NAME,
            advertiserId: ADVERTISER_ID,
            advertiserName: ADVERTISER_NAME,
            advertiserDomains: [ARRAY_OF_ADVERTISER_DOMAINS],
            brandId: BRAND_ID,
            brandName: BRAND_NAME,
            primaryCatId: IAB_CATEGORY,
            secondaryCatIds: [ARRAY_OF_IAB_CATEGORIES],
            mediaType: MEDIA_TYPE
        }
```

It is conceivable that publishers could desire any number of functions to apply across any number of fields within this object, including but not limited to : 

- Filtering bids based on a list of approved responses
- Filtering bids based on the presence or absence of a response
- Adjusting bids based on a list of approved responses
- Adjusting bids based on the presence or absence of a response

From what I can tell, the initial use case is that we would want to support would be filtering (ie. removing from the auction) bids based on the presence of advertiserDomains -- removing bids from auction that lack the field entirely. This removes significant complexity from the project, as when logic can exist on multiple fields you need a hierarchy and ordering -- but that's not to say that this won't need to exist in the future (or is potentially a valid role for a real time module?)

patmmccann:
There may also be a use case in flooring, eg toss bids from a particular advertiser not above X.

GLStephen:
It looks like clickUrl has been dropped out of this set. Just curious if that was intentional.

patmmccann:
Check out the solution on #6453

ChrisHuie:
@gglas would this fall under #6453 and we can then close this one out?

bretg:
Status:
- There's documentation on how to do this with an event callback -- https://docs.prebid.org/dev-docs/examples/meta-bid-filtering.html
- The open item is to discuss:
    - is this sufficient? I would argue no -- the documented solution feels like a hack. 
[...]

PR #12147, opened 2024-08-20
New Module: Bid response filter

mkomorski:
## Type of change

- [ ] Bugfix
- [x] Feature
- [ ] New bidder adapter  
- [ ] Updated bidder adapter   
- [ ] Build related changes
- [ ] CI related changes

- [ ] Other

## Other information

#6345

### DIFF
--- a/modules/bidResponseFilter/index.js
+++ b/modules/bidResponseFilter/index.js
@@ -1,7 +1,7 @@
 import { config } from '../../src/config.js';
 import { getHook } from '../../src/hook.js';
 
-const MODULE_NAME = 'bidResponseFilter';
+export const MODULE_NAME = 'bidResponseFilter';
 export const PUBLISHER_FILTER_REJECTION_REASON = 'PUBLISHER_FILTER_REJECTION_REASON';
 export const BID_CATEGORY_REJECTION_REASON = 'BID_CATEGORY_REJECTION_REASON';
 export const BID_ADV_DOMAINS_REJECTION_REASON = 'BID_ADV_DOMAINS_REJECTION_REASON';
@@ -12,25 +12,27 @@ function init() {
 };
 
 export function addBidResponseHook(next, adUnitCode, bid, reject) {
-  const filterFn = config.getConfig(`${MODULE_NAME}.filterFn`)
-  const filter = typeof filterFn == 'function' ? filterFn : () => true;
   const { bcat = [], badv = [], battr = [] } = config.getAnyConfig('ortb2') || {};
-  const { primaryCatId, advertiserDomains = [], attr } = bid.meta || {};
+  const moduleConfig = config.getConfig(MODULE_NAME);
+  
+  const catConfig = moduleConfig?.cat || { enforce: true, block_unknown_cat: true };
+  const advConfig = moduleConfig?.adv || { enforce: true, block_unknown_adomain: true };
+  const attrConfig = moduleConfig?.attr || { enforce: true, block_unknown_attr: true };
 
-  // checking if bid fulfills validation rule set by publisher
-  if (!filter(bid)) {
-    reject(PUBLISHER_FILTER_REJECTION_REASON);
+  const { primaryCatId, secondaryCatIds = [], advertiserDomains = [], attr: metaAttr } = bid.meta || {};
+
+ // checking if bid fulfills ortb2 fields rules
+ if ((catConfig.enforce && bcat.some(category => [primaryCatId, ...secondaryCatIds].includes(category))) || 
+    (catConfig.block_unknown_cat && !primaryCatId)) {
+    reject(BID_CATEGORY_REJECTION_REASON);
+  } else if ((advConfig.enforce && badv.some(domain => advertiserDomains.includes(domain))) || 
+    (advConfig.block_unknown_adomain && !advertiserDomains.length)) {
+    reject(BID_ADV_DOMAINS_REJECTION_REASON);
+  } else if ((attrConfig.enforce && battr.includes(metaAttr)) || 
+    (attrConfig.block_unknown_attr && !metaAttr)) {
+    reject(BID_ATTR_REJECTION_REASON);
   } else {
-    // checking if bid fulfills ortb2 fields rules
-    if (bcat.includes(primaryCatId)) {
-      reject(BID_CATEGORY_REJECTION_REASON);
-    } else if (badv.some(domain => advertiserDomains.includes(domain))) {
-      reject(BID_ADV_DOMAINS_REJECTION_REASON);
-    } else if (battr.includes(attr)) {
-      reject(BID_ATTR_REJECTION_REASON);
-    } else {
-      return next(adUnitCode, bid, reject);
-    }
+    return next(adUnitCode, bid, reject);
   }
 }
 
--- a/modules/bidResponseFilter/index.js
+++ b/modules/bidResponseFilter/index.js
@@ -1,0 +1,37 @@
+import { config } from '../../src/config.js';
+import { getHook } from '../../src/hook.js';
+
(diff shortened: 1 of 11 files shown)


patmmccann:
let's make sure to help populate attr and categories with https://github.com/prebid/Prebid.js/blob/9e2c63ec92b3f88882d04ae2b9cbbdc80f504a32/libraries/ortbConverter/processors/default.js#L98C1-L99C1
let's filter on any response category, not just primary.

patmmccann:
https://docs.prebid.org/prebid-server/pbs-modules/ortb2-blocking.html#configuration-details

patmmccann:
lets set boolean for each of the field and add the unknown toggle in the config eg `<html>
<body>

block-unknown-adv-cat
--


<br class="Apple-interchange-newline">
</body>
</html>`

modules/bidResponseFilter/index.js
```diff
@@ -0,0 +1,39 @@
+import { config } from '../../src/config.js';
+import { getHook } from '../../src/hook.js';
+
+export const MODULE_NAME = 'bidResponseFilter';
+export const PUBLISHER_FILTER_REJECTION_REASON = 'PUBLISHER_FILTER_REJECTION_REASON';
```
dgirardi, line 5: these should be human readable

modules/bidResponseFilter/index.js
```diff
@@ -0,0 +1,39 @@
...
+export const BID_CATEGORY_REJECTION_REASON = 'BID_CATEGORY_REJECTION_REASON';
+export const BID_ADV_DOMAINS_REJECTION_REASON = 'BID_ADV_DOMAINS_REJECTION_REASON';
+export const BID_ATTR_REJECTION_REASON = 'BID_ATTR_REJECTION_REASON';
+
+function init() {
+  getHook('addBidResponse').before(addBidResponseHook);
+};
+
+export function addBidResponseHook(next, adUnitCode, bid, reject) {
+  const { bcat = [], badv = [], battr = [] } = config.getAnyConfig('ortb2') || {};
```
dgirardi, line 15: ortb2 can be different for each bid response. You can get a specific response's with `auctionManager.index.getBidderRequest(bid)?.ortb2`; however, responses can also not be associated with any request, in which case the best avialable `ortb2` is the one associated with the auction (like you used in https://github.com/prebid/Prebid.js/pull/11946)

It may be worth to add a `getOrtb2` method to `auctionManager.idex` -  similar to [getMediaTypes](https://github.com/prebid/Prebid.js/blob/95c8118583eb0737409e41c8f436f90bc134df32/src/auctionIndex.js#L35-L47), which also has a fallback for when there is not request.

modules/bidResponseFilter/index.js
```diff
@@ -0,0 +1,39 @@
...
+  const moduleConfig = config.getConfig(MODULE_NAME);
+  
+  const catConfig = moduleConfig?.cat || { enforce: true, block_unknown_cat: true };
```
dgirardi, line 18: if I 
```
setConfig({bidResponseFilter: {cat: {block_unknown_cat: false}}})
```
my expectation would be that `bcat` would still be enforced (e.g., `enforce` defaults to `true`), but I think this would disable it.

modules/bidResponseFilter/index.js
```diff
@@ -0,0 +1,39 @@
...
+  const advConfig = moduleConfig?.adv || { enforce: true, block_unknown_adomain: true };
```
dgirardi, line 19: It's probably out of scope, but I don't like these parameter names. `attr: {block_uknown_attr}` is redundant; and we do not use snake_case anywhere else. It's also inconsistent within itself (why `cat/block_unknown_cat`, but `adv/block_unknown_adomain`?) 

If we weren't following PBS this would be something like `adv: {enforce, blockUnkown}`.  Do we still prefer PBS's syntax @patmmccann ?

patmmccann: i don't know, maybe those are good reasons to abandon the hope of matching the syntax.

modules/bidResponseFilter/index.js
dgirardi: I just noticed that `battr` is not here - it's [repeated for each media type in imp](https://github.com/InteractiveAdvertisingBureau/openrtb2.x/blob/main/2.6.md). It's not clear where the publisher should set it - candidates are `ortb2Imp[mediaType].battr` and `mediaTypes[mediaType].battr`

Assuming the first one should have priority, we could:

- update [ad unit validation](https://github.com/prebid/Prebid.js/blob/master/src/prebid.js#L230-L242) for each media type to find `battr` from `ortb2Imp[type].battr`, falling back to `mediaTypes[type].battr`, and copy it to both places 
- add some unit test for ortbConverter [banner](https://github.com/prebid/Prebid.js/blob/master/test/spec/ortbConverter/banner_spec.js),  [video](https://github.com/prebid/Prebid.js/blob/master/test/spec/ortbConverter/video_spec.js) and [native](https://github.com/prebid/Prebid.js/blob/master/test/spec/ortbConverter/native_spec.js) to make sure the "right" battr wins
- this should then look for `battr` from `index.getBidRequest(bid).ortb2Imp[bid.mediaType].battr`, falling back to `index.getAdUnit(bid).ortb2Imp[bid.mediaType].battr` when there is no request for the bid

(pending confirmation on the relative priority of `ortb2Imp` vs `mediaTypes`)

src/prebid.js
```diff
@@ -100,6 +100,16 @@ function validateSizes(sizes, targLength) {
...
 }
 
+function setBattrForAdUnit(adUnit, mediaType) {
+  const battr = adUnit.ortb2Imp?.[mediaType]?.battr || adUnit.mediaType?.[mediaType]?.battr;
+  if (adUnit.ortb2Imp?.[mediaType] && battr) {
+    adUnit.ortb2Imp[mediaType].battr = battr;
+  }
+  if (adUnit.mediaTypes?.[mediaType] && battr) {
+    adUnit.mediaTypes[mediaType].battr = battr;
+  }
```
dgirardi, line 110: IMO this should always copy it over, since the goal is to make it easy to find for adapters. For example if I have `{mediaTypes: {video: {battr}}, ortbImp: {}}`, it would remain relegated to `mediaTypes`; turn it around and it would remain only in `ortb2Imp`, and adapters would still need to check both places. (They're also both decent test cases).  We should also probably log a warning if they're both specified but they differ.

src/prebid.js
```diff
@@ -100,6 +100,20 @@ function validateSizes(sizes, targLength) {
...
+  const mediaTypes = adUnit.mediaTypes || {};
+
+  if (ortb2Imp[mediaType]?.battr && mediaTypes[mediaType]?.battr && (ortb2Imp[mediaType]?.battr !== mediaTypes[mediaType]?.battr)) {
+    logWarn('battr field differ between ortb2Imp and mediaTypes');
+  }
+
+  const battr = ortb2Imp[mediaType]?.battr || mediaTypes[mediaType]?.battr;
+  
+  adUnit.ortb2Imp = {...ortb2Imp, [mediaType]: {...(ortb2Imp[mediaType] || {}), battr}};
+  adUnit.mediaTypes = {...mediaTypes, [mediaType]: {...(mediaTypes[mediaType] || {}), battr}};
```
dgirardi, line 114: I think this would add `battr: undefined` in most cases. You can also simplify a bit by using `deepSetValue`, e.g.

```suggestion
  if (battr != null) {
      ['mediaTypes', 'ortb2Imp'].forEach(target => deepSetValue(adUnit, `${target}.${mediaType}.battr`, battr))
  }
```

could you please also add some tests?

dgirardi (approved):

PR #12149, opened 2024-08-20
MediaWallah openLink User ID Submodule: GDPR and CCPA in payload

bradatefi:
## Type of change
- [ ] Compliance

## Description of change
GDPR and CCPA Compliance and Cookie Handling updates for prebid compliance


patmmccann:
@jdwieland8282 doesnt fix that https://github.com/prebid/Prebid.js/blob/ea5854505910059cfdebbc8af3f8b1bcf173b9f9/modules/mwOpenLinkIdSystem.js#L88

smenzer (changes requested):
the code looks fine, but you need unit tests for it, please.

patmmccann:
I'm closing this as stale and noting that consent is now the second argument to getid, so you don't need to import any handlers to get it

PR #12150, opened 2024-08-20
AdPlayerPro Video Module : initial module release

artemAdp:
## Type of change
- [x] Feature

## Description of change
New video module for AdPlayer.Pro

### DIFF
--- a/modules/adplayerproVideoProvider.js
+++ b/modules/adplayerproVideoProvider.js
@@ -1,0 +1,439 @@
+import {
+  API_FRAMEWORKS,
+  PLACEMENT,
+  PLAYBACK_METHODS,
+  PROTOCOLS,
+  VIDEO_MIME_TYPE,
+  VPAID_MIME_TYPE
+} from '../libraries/video/constants/ortb.js';
+import {
+  AD_CLICK,
+  AD_COMPLETE,
+  AD_ERROR,
+  AD_IMPRESSION,
+  AD_LOADED,
+  AD_PAUSE,
+  AD_PLAY,
+  AD_REQUEST,
+  AD_SKIPPED,
+  AD_STARTED,
+  DESTROYED,
+  ERROR,
+  MUTE,
+  PLAYER_RESIZE,
+  SETUP_COMPLETE,
+  SETUP_FAILED,
+  VOLUME
+} from '../libraries/video/constants/events.js';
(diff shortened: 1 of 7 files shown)


modules/adplayerproVideoProvider.js
```diff
@@ -0,0 +1,439 @@
...
+        API_FRAMEWORKS.VPAID_2_0,
+        API_FRAMEWORKS.OMID_1_0
+      ],
+    };
+
+    return video;
+  }
+
+  function getOrtbContent() {
+  }
```
karimMourra, line 118: If the player has content information such as media source url , media id, media title etc, it would be good to populate. Contextual targeting can help get higher bids

modules/adplayerproVideoProvider.js
```diff
@@ -0,0 +1,439 @@
...
+
+  function offEvent(event, callback) {
+    const playerEventName = utils.getPlayerEvent(event);
+    const eventHandler = callbackStorage.getCallback(playerEventName, callback);
+    if (eventHandler) {
+      player && player.off(playerEventName);
+    } else {
+      player && player.off(playerEventName, eventHandler);
+    }
```
karimMourra, line 175: was this meant to be inverted ?

```suggestion
    if (eventHandler) {
      player && player.off(playerEventName, eventHandler);
    } else {
      player && player.off(playerEventName);
    }
```
seems strange to check for eventHandler but only reference it in the else

karimMourra (changes requested):
Overall looks great, just one small observation

modules/adplayerproVideoProvider.js
artemAdp: Thank you for your review. Yes, this is a mistake. I fixed it and added a test for this case.

modules/adplayerproVideoProvider.js
artemAdp: This module only supports outstream format, so it does not use content.

karimMourra (approved):
@ChrisHuie LGTM!

ChrisHuie:
@artemAdp can you please add documentation for this new module in our docs repo https://github.com/prebid/prebid.github.io

artemAdp:
@ChrisHuie Thank you for your comment. I have added documentation.

PR #12156, opened 2024-08-21
UNICORN Bid Adapter: add scroll notification

ctylim:
## Type of change

- [x] Feature
## Description of change

- add scroll measurement script for interscroller ads


ctylim:
Hello @patmmccann 
Thank you for reviewing, I understand why this PR is unable to merge.
Is it also not allowed to implement this change as RTD module for the same reason?

Issue #11730, opened 2024-06-07
PAAPI: Parallel Contextual and IG auctions 


patmmccann:
Feature outline:

Bid adapters should be able to specify, as a hard coded object,  an auction config with promises that will later be filled in from the igi:igs object on their bid response. The auctionConfig gets submitted immediately (before the auction commences) to the TLS (or paapi submodule representing the TLS, eg fledgeforgpt).

The interestGroupBuyers presents a wrinkle, as [it cannot be a promise](https://github.com/WICG/turtledove/issues/1093) and could have the following logic:

Value in local storage that a bidder can set and read, falling back to publisher config list of ssp buyers, falling back to a hardcoded list in the adapter. The resulting list could then be filtered by the pub. Sellers should endeavor to list all buyers they work with. Prebid could have a constants file with all known buyers and sellers could import the constants they wish to avoid respecification of tons of long-ish strings in each adapter.

Bid responses should have an extension location the ortbConvertor can pick up for the replacement list of buyers they want to call on future auctions. Bid requests should have a new place for current known buyer list version so the endpoint can know if it wants to update the list. They should also have a new place to say: 'paapi auction style is parallel, I have already submitted an config for you, your response will be for promise filling.'

InterestGroupBuyer timeouts of buyers to exclude from the auction can be set to zero by the seller. 

For example, if dsp A doesnt need a PBS object and dsp B and C do, and there is only DSP B PBS object available, the IGB timeouts would be set to {a:X, b:X, c:0} where X is some reasonable integer. Alternatively, since sellerSignals is also a promise, sellers can set desirability of whichever bids they want to zero based on the igs response, providing a second 'boot buer' opportunity.

TLS should provide an API that allows the pub to say something like 'I am done submitting paapi auction configs'. Prebid TLS module (currently for either GAM or pub designated TLS) would call this function at the same time as requestBids, with all the auction configs that were promise based. 

Bidders unable to provide promise shell configs would be excluded from the parallel paapi auction.

Paapi module config should take a flag saying the publisher wants to run a parallel paapi auction. Bidders submitting auctionConfig too late should generate console warnings.

rdgordon-index:
A few comments following our verbal discussion on Friday:

> Bid adapters should be able to specify, as a hard coded object, an auction config with promises

Do you envision that the awareness of the parallel nature of a given PAAPI auction will be visible to the bid adapters at runtime?

> Paapi module config should take a flag saying the publisher wants to run a parallel paapi auction. 

[...]

PR #12160, opened 2024-08-21
Paapi tools: add constants for importing

patmmccann:
In order to reduce duplication among bid adapters of frequently used constants for buyer origins in the pre-communication submitted auction config after #11730 ; making a list of constants we expect to grow for bidders to import from.


modules/ringieraxelspringerBidAdapter.js
```diff
@@ -314,9 +315,9 @@ const parseAuctionConfigs = (serverResponse, bidRequest) => {
     auctionConfigs.push({
       'bidId': bid.bidId,
       'config': {
-        'seller': 'https://csr.onet.pl',
-        'decisionLogicUrl': `https://csr.onet.pl/${encodeURIComponent(bid.params.network)}/v1/protected-audience-api/decision-logic.js`,
-        'interestGroupBuyers': ['https://csr.onet.pl'],
+        'seller': BO_CSR_ONET,
```
dgirardi, line 318: is this the only buyer used at the moment? (grepping seems to suggest yes).

I understand the intent and in the long run it may make sense (a bit early to tell) but currently, the effect of this PR would be to transform this single string into a much bigger library from which we use this single string.

Should we start by only including buyers that we know are in use? or wait until there's more than one consumer for the library?

patmmccann: Yes, I think so, the expectation is the buyers are only needed to be defined in a future case

patmmccann: Update, I commented out the unused imports

dgirardi (approved):

PR #12162, opened 2024-08-21
Sharethrough Bid Adapter: support battr property in bid requests

jefftmahoney:
## Type of change
- [x] Feature

## Description of change

This update allows publishers to have `battr` included in bid requests if they specify a `battr` property in either `mediaTypes.banner` or `mediaTypes.video`.  (Native not supported.)


patmmccann:
Hmm we may have a bit of a design issue, what is the specify it on ortb2Imp

jefftmahoney:
> Hmm we may have a bit of a design issue, what if they specify it on ortb2Imp
> 
> Ideally a bidder shouldn't have to look around for it, there would be one place
> 
> Cc #7925 and #12147

Hmmm, interesting.  Thanks for the links - I'll check out the discussions on those and get back to you.

patmmccann:
if you support ortb2imp like you do here, i can merge this, but also curious for your thoughts on having more than one place 

https://github.com/prebid/Prebid.js/blob/bb43feb75fe7400088e2b2bf99c5568c8ee6a4d1/modules/sharethroughBidAdapter.js#L111

patmmccann (approved):

Issue #11541, opened 2024-05-22
LiveIntent User ID Module: Eliminating live-connect NPM Dependency.


3link:
## Type of issue
Eliminating live-connect NPM dependency.

## Description
The goal of this issue is to get rid of the LiveConnect NPM module dependency. 

Instead, the following approach should be implemented: The LiveIntent user ID module will rely on the LiveConnect script being present on the page. This LiveConnect script will accept commands and execute them according to the configuration passed along with them (e.g. what to resolve, privacy settings, callbacks, etc.).

The user ID module implementation will look like in the snippet below:

```
window.liQHub = window.liQHub || [];

decode(value, config) {
  const sync = {
	  config: config,
	  commands: [{
		command: "sync"
	  }]
  }
  window.liQHub.push(sync)
  // decode the value
}

getId(config) {
  const result = function(callback) {
  	const syncAndResolve = {
  		config: config,
  		commands: [
  			{
				command: "sync"
  			},
	  		{
	  			command: 'resolve',
	      		onSuccess: response => {callback(response)},
	      		onError: error => {callback()}
	  		}
  		]
  	}
     window.liQHub.push(syncAndResolve)
  }
  return {callback: result}
}
```

3link:
Hi @SKOCHERI @patmmccann @jdwieland8282. The proposal is as communicated in Slack. We started working on it this week. There is no PR in Prebid yet as we have to rewrite LiveConnect from scretch first.

patmmccann:
@3link excluding the import in geographies you don't function had a rather noticeable performance improvement; this speaks to the potential opportunity in the geographies where you do from the disentangling

3link:
@patmmccann Do you mean bidding performance or technical (time to load the script, time to bid etc.)?

Issue #9570, opened 2023-02-21
UserID modules: Implementation guidelines


slimkrazy:
## Type of issue
question

## Description
Hi folks, I've got some general questions regarding UserID module implementation as I'm actually not sure what rules are in play here. I see some user ID modules directly accessing storage and writing to cookies / local storage, whilst others just leave Prebid core to manage this (respecting the publisher's expiry settings etc). So here are my questions:

1. Are UserID modules ok to load external scripts without special justification? 
2. Are UserID modules free to generate say a GUID, store that for as many days as they choose?

What are the red lines?

Thanks

patmmccann:
Officially, this is the answer https://docs.prebid.org/dev-docs/module-rules.html#:~:text=In%20order%20to%20be%20listed,standard%20or%20Prebid%20JavaScript%20functions.

We have tried to be reasonable and consistent, but it isn't always easy to be both. We've noticed client side entropy gathering via external js a la flashtalking is becoming more popular. We don't want those scripts linked from the project so no pub ever thinks we review or warranty them, so please make them a configuration parameter or just assume the pub has loaded them already via their own means (preferred). Any script loading needs to be excessively disclosed and the reason for it well documented. 

2) yes this is typical behavior

PR #12167, opened 2024-08-23
LiveIntent User ID Module: Eliminating live-connect NPM Dependency

3link:
## Type of change
- [x] Feature

## Description of change
Allows to use the LiveIntent user ID module with an instance of LiveConnect v3 present on the page instead of inlining the LiveConnect NPM module into the user ID module as described here: https://github.com/prebid/Prebid.js/issues/11541.

The goal is to support both `standard`/`minimal` mode as well as the `external` mode during the migration phase. Once all our customers have migrated away from using `standard`/`minimal` mode, we will deprecate this mode and delete the LiveConnect NPM module dependency from Prebid together with all the code that depends on it. That would include moving the code from `libraries/liveIntentId` back into `modules/liveintentIdSystem` as we would not need to differentiate between two module implementations.

## Other information
- Resolves https://github.com/prebid/Prebid.js/issues/11541
- Documentation changes: https://github.com/prebid/prebid.github.io/pull/5567


3link:
@jlquaccia I have a question about possible ways to extend our user id module that depends on the external LiveConnect. Would it be possible to extend the module configuration such that a URL can be provided from which to download the LiveConnect script and the user id module would then embed this script on the page? If no such URL is provided in the configuration, nothing will be embedded.

That would make the switch for our customers smoother as they will just need to add an additional config to our user id module and it will automatically deploy the script.

jlquaccia:
> @jlquaccia I have a question about possible ways to extend our user id module that depends on the external LiveConnect. Would it be possible to extend the module configuration such that a URL can be provided from which to download the LiveConnect script and the user id module would then embed this script on the page? If no such URL is provided in the configuration, nothing will be embedded.
> 
> That would make the switch for our customers smoother as they will just need to add an additional config to our user id module and it will automatically deploy the script.

Overall, I think the changes on this PR seem fine to me.  In terms of allowing a URL to be provided via module configuration, I'm not sure if that is a direction that Prebid would like to go in since it seems it could open some security related scenarios.  @ncolletti @patmmccann what are your thoughts

jlquaccia (approved):

3link:
> > @jlquaccia I have a question about possible ways to extend our user id module that depends on the external LiveConnect. Would it be possible to extend the module configuration such that a URL can be provided from which to download the LiveConnect script and the user id module would then embed this script on the page? If no such URL is provided in the configuration, nothing will be embedded.
> > That would make the switch for our customers smoother as they will just need to add an additional config to our user id module and it will automatically deploy the script.
> 
> Overall, I think the changes on this PR seem fine to me. In terms of allowing a URL to be provided via module configuration, I'm not sure if that is a direction that Prebid would like to go in since it seems it could open some security related scenarios. @ncolletti @patmmccann what are your thoughts

Just to provide more context here: our customers would now be required to place our LiveConnect script on their domains in order for our user ID sub-module to work. We have customers with thousands of domains and it would make upgrading to the newer Prebid/LiveIntent sub-module version much easier for them if instead of setting up LiveConnect on each of their pages manually, they could configure their URL for the LiveConnect script directly in our user ID module.

libraries/liveIntentId/externalIdSystem.js
```diff
@@ -0,0 +1,173 @@
...
+  if (cachedClientRef != null) return cachedClientRef
+
+  const clientRef = {}
+
+  const clientDetails = { name: 'prebid', version: '$prebid.version$' }
+
+  const collectConfig = configParams.liCollectConfig ?? {};
+
+  let integration
+  if (collectConfig.appId != null) {
```
ncolletti, line 29: is it worth checking for the optional `publisherId` param here?

3link: @ncolletti The code here determines the type of the integration that depends on a combination of certain fields. In this case if the appId is provided, the actual value of the publisherId can be defined or undefined. Both is fine.

ncolletti:
@3link As far as I'm aware Prebid has been upfront about not allowing external code to be loaded other than from RTD Modules where a contract/registration to acknowledge the contents of the JS by the publisher is signed prior. While cumbersome, I think this configuration is preferred by Prebid. 

Somewhat recent guidance on this topic: https://github.com/prebid/Prebid.js/issues/9570

Any case, I don't know if a change should happen in this iteration of your adapter. It's best to bring this up to the Identity PMC group for further discussion

3link:
> As far as I'm aware Prebid has been upfront about not allowing external code to be loaded other than from RTD Modules where a contract/registration to acknowledge the contents of the JS by the publisher is signed prior. While cumbersome, I think this configuration is preferred by Prebid.

> Somewhat recent guidance on this topic: https://github.com/prebid/Prebid.js/issues/9570

> Any case, I don't know if a change should happen in this iteration of your adapter. It's best to bring this up to the Identity PMC group for further discussion

@ncolletti That makes sense. Thank you for the link! 

I think it makes sense to keep these two topics separate and merge this version of the module as is for now.

ncolletti (approved):

ncolletti:
@jlquaccia looks good to me, will let you give a final sign off and merge

jlquaccia:
@ncolletti yep, looks good to me too.  Merging now..

libraries/liveIntentId/idSystem.js
```diff
@@ -0,0 +1,227 @@
...
+ * @requires module:modules/userId
+ */
+import { triggerPixel, logError } from '../../src/utils.js';
+import { ajaxBuilder } from '../../src/ajax.js';
+import { gdprDataHandler, uspDataHandler, gppDataHandler } from '../../src/adapterManager.js';
+import { submodule } from '../../src/hook.js';
+import { LiveConnect } from 'live-connect-js'; // eslint-disable-line prebid/validate-imports
+import { getStorageManager } from '../../src/storageManager.js';
+import { MODULE_TYPE_UID } from '../../src/activities/modules.js';
+import { DEFAULT_AJAX_TIMEOUT, MODULE_NAME, composeIdObject, eids, DEFAULT_DELAY, GVLID, PRIMARY_IDS, parseRequestedAttributes } from './shared.js'
```
s-solodovnikov, line 14: Hi. You are importing the `DEFAULT_DELAY` const but it does not exist.

3link: https://github.com/prebid/Prebid.js/pull/12334

PR #12185, opened 2024-08-28
Dailymotion bid adapter: add publisher restrictions in consent enforcement

sebmil-daily:
## Type of change
- [x] Updated bidder adapter

## Description of change
The previous version of the consent enforcement was only looking at the consent given by the user.

This change now also looks at the publisher restrictions to disable cookie sending when the user gave consent but the publisher disallows some consents.


sebmil-daily:
Note: I rebased on the last `master` and force-pushed to relaunch CircleCI tests as I don't have access to it.
On the second try the tests passed without any change on my side.

Is there a better way to relaunch the test when this happens ?

PR #12193, opened 2024-08-30
Prebid Core : Remove duplicate event call for actionDebug event

pm-komal-kumari:
## Type of change

- [X] Bugfix

## Description of change

 
This PR removes the redundant auction debug logging that was present in the code. The debug log statement was found to be a duplicate of the implementation already handled by the logWarn function. The logWarn function is already invoked within the validateAdUnitPos function, ensuring that the necessary logging occurs without repetition.

By removing this duplicate logging, we streamline the code, reduce unnecessary verbosity in the logs, and improve overall maintainability. This change does not affect the existing functionality, as the necessary logging is still being performed through the logWarn function.

Screenshot 2024-08-30 at 4 19 38 PM

### DIFF
--- a/src/prebid.js
+++ b/src/prebid.js
@@ -194,7 +194,6 @@ function validateAdUnitPos(adUnit, mediaType) {
     let warning = `Value of property 'pos' on ad unit ${adUnit.code} should be of type: Number`;
 
     logWarn(warning);
-    events.emit(EVENTS.AUCTION_DEBUG, { type: 'WARNING', arguments: warning });
     delete adUnit.mediaTypes[mediaType].pos;
   }
 


patmmccann:
this appears to revert functionality as the event listeners would no longer fire

jlquaccia:
looks like this change seems ok as it appears the same functionality is occurring here in the logWarn function as stated: https://github.com/prebid/Prebid.js/blob/master/src/utils.js#L235.  @dgirardi can you confirm?

patmmccann (approved):

patmmccann:
Great point ty

jlquaccia:
Thanks @patmmccann !

PR #12186, opened 2024-08-29
Connatix Bid Adapter: Support Events

rares-mihai-preda:
## Type of change

- [ ] Bugfix
- [ ] Feature
- [ ] New bidder adapter  
- [ ] Updated bidder adapter   
- [ ] Code style update (formatting, local variables)
- [x] Refactoring (no functional changes, no api changes)
- [ ] Build related changes
- [ ] CI related changes

- [ ] Does this change affect user-facing APIs or examples documented on http://prebid.org?
- [x] Other

PR #12199, opened 2024-09-03
Connatix Bid Adapter: Support Events

rares-mihai-preda:
## Type of change

- [ ] Bugfix
- [X] Feature
- [ ] New bidder adapter  
- [ ] Updated bidder adapter   
- [ ] Build related changes
- [ ] CI related changes

- [ ] Other

## Description of change

Support Prebid.js events


patmmccann:
thanks for resubmitting! check your unit tests and jsdoc warnings

patmmccann:
linking with previous #12186

rares-mihai-preda:
@patmmccann 
Thank you for reviewing the pull request. We’ve added tests and addressed the 3 linter warnings.

Could you please review it once more?

modules/connatixBidAdapter.js
```diff
@@ -217,7 +234,42 @@ export const spec = {
...
       url
     }];
-  }
+  },
+
+  /**
+   * Register bidder specific code, which will execute if the server response time is greater than auction timeout
+   */
+  onTimeout: (timeoutData) => {
+    const connatixBidRequestTimeout = timeoutData.find(bidderRequest => bidderRequest.bidder === BIDDER_CODE);
```
robertrmartinez, line 243: Couple things here:

If a publisher uses an alias for `connatix` then this will not find it - you can do this by keep a list of alias names that you see come in through `buildRequests` - or you can import the aliasRegistry I think.

Although - I do think that this method will only receive an array of bids that went through connatix adapter - and therefore you might not even need to find / filter etc.

patmmccann:
Thanks! @robertrmartinez is the reviewer

modules/connatixBidAdapter.js
```diff
@@ -140,10 +155,12 @@ export const spec = {
 
     _handleEids(requestPayload, validBidRequests);
 
+    context = requestPayload;
```
robertrmartinez, line 158: So you are wanting to send your bidRequest payload in every single event you fire?

Seems like a lot of duplication but it is fine of course.

The only thing I can think of is that there can be race condition here if the publisher site is running multiple auctions within close time period.

Auction A
Auction B

Context = {};
Auction A start
Context = A

Auction B Start
Context = B

Auction A Bid WON / timeout etc
=> Event sends Context = Auction B for an event from Auction A

We can get around this by perhaps making context a map of Auction IDs => context

Then sending the right one in each event.

Or if this is enough of an edge case and you would rather keep it more simple then that is fine.

Just a suggestion!

robertrmartinez (changes requested):
Couple questions!

modules/connatixBidAdapter.js
rares-mihai-preda: @robertrmartinez Thanks for the heads-up! We’ve incorporated aliasRegistry to manage aliases, streamlining the process and eliminating the need to manually track them in buildRequests. This should ensure Connatix aliases are automatically handled. Could you please take another look at the PR when you have a moment? Thanks!

modules/connatixBidAdapter.js
rares-mihai-preda: Thanks for the thoughtful suggestion! We did discuss this model with the backend team, and they specifically need the bidRequest payload with each event for their current setup.

We understand the potential for race conditions and have considered various options. However, for now, we’ll stick with this approach to meet the backend requirements.

I appreciate your input and am grateful for your understanding. If you have any more thoughts or questions, feel free to share!

rares-mihai-preda:
Hey @robertrmartinez ,

Is there anything else you need from us to proceed with merging the PR? If you have any further questions or require additional information, just let us know.

Thanks again for your review and feedback!

cosminser:
Hei @robertrmartinez . Just following up on this, are we still blocked, or can we proceed further with the PR / merge. From my opinion the changes should suffice for the moment, no? Thanks!

robertrmartinez (approved):

PR #12206, opened 2024-09-05
adUnits.js: Fix if dots '.' are in adUnitCode

### LINKED ISSUE #12204: getBidderRequestsCounter cannot parse ad unit codes with too many dots
## BUG
getBidderRequestsCounter cannot find the "requestsCounter" when the adunit code contains too many "." (dots). Google Ad Manager GPT allows ad unit codes with more than one dot in the ad unit code. 

## Description
https://github.com/prebid/Prebid.js/blob/360c3eaf94baf6af0336abc59829baa3d66a3a5d/src/adUnits.js#L66
When it reached `getBidderRquestsCounter()`, the adunit is being parsed by ".".
So  the ad unit`gpt-ad-mayoclinic.org-top` is  broken down as 
* gpt-ad-mayoclinic > org-top > bidders > rise > requestsCounter

instead of 
* gpt-ad-mayoclinic.org-top > bidders > rise > requestsCounter

## Steps to reproduce
1. Open a chrome browser tab
2. Open Chrome Dev tools 
3. Open the page 
4. Go to console & filter for "INFO: Bids Requested for Auction with id"
5. Notice, the bidderRequestsCount for Rubicon is 1, but Rise is 0.
-- Rubicon ad unit code is "test-slot"
-- Rise ad unit code is "gpt-ad-mayoclinic.org-top.bidders.rise"
6. Notice, in the network tab there are two bid requests for Rubicon and Rise. Filter for /fastlane|yellowblue/

## Test page
https://hyunjoo-test-environment.s3.amazonaws.com/public/github/getBidderRequestsCount/testpage.html?pbjs_debug=true

### Expected results
I expected to see bidderRequestsCount to be equivalent to the number of requests in the network tab. 

### Actual results
I see uneven bidderRequestsCount against the number of requests I see under the network tab.

## Platform details
I can see this issue from v8.48.0.

## Other information
This is affecting Bidder Rise because Rise is using bids.bidderRequestCount in their bid request payload, here: https://github.com/prebid/Prebid.js/blob/ac5cc64644f6371b014cf656f2af78ef2f257c58/libraries/riseUtils/index.js#L126.
This is also affecting other bidders count.

robertrmartinez:
## Type of change
- [X] Bugfix

## Description of change
Fixes #12204 

Some sites have dots `.` in the adUnitCodes which causes some `deepAccess` calls to not do what we expect.

This PR switches those adCode accessors to use optional chaining instead.

### DIFF
--- a/src/adUnits.js
+++ b/src/adUnits.js
@@ -1,5 +1,3 @@
-import { deepAccess } from './utils.js';
-
 let adUnits = {};
 export function reset() {
   adUnits = {}
@@ -54,7 +52,7 @@ export function incrementBidderWinsCounter(adunit, bidderCode) {
  * @returns {number} current adunit count
  */
 export function getRequestsCounter(adunit) {
-  return deepAccess(adUnits, `${adunit}.requestsCounter`) || 0;
+  return adUnits?.[adunit]?.requestsCounter || 0;
 }
 
 /**
@@ -64,7 +62,7 @@ export function getRequestsCounter(adunit) {
  * @returns {number} current adunit bidder requests count
  */
 export function getBidderRequestsCounter(adunit, bidder) {
-  return deepAccess(adUnits, `${adunit}.bidders.${bidder}.requestsCounter`) || 0;
+  return adUnits?.[adunit]?.bidders?.[bidder]?.requestsCounter || 0;
 }
 
 /**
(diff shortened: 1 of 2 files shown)


robertrmartinez:
image

Test run with new tests but without code change showing the issue ^^

dgirardi (approved):

Issue #11010, opened 2024-01-30
New activity control: load external script


patmmccann:
## Type of issue
Feature Req

## Description
loadExternalScript in adLoader.js should have an activity control surrounding it.

PR #12207, opened 2024-09-05
Core: New activity control - load external script

mkomorski:
## Type of change

- [ ] Bugfix
- [x] Feature
- [ ] New bidder adapter  
- [ ] Updated bidder adapter   
- [ ] Build related changes
- [ ] CI related changes

- [ ] Other

## Other information

#11010

### DIFF
--- a/src/adloader.js
+++ b/src/adloader.js
@@ -54,8 +54,7 @@ const _approvedLoadExternalJSList = [
  * @param {object} attributes an object of attributes to be added to the script with setAttribute by [key] and [value]; Only the attributes passed in the first request of a url will be added.
  */
 export function loadExternalScript(url, moduleCode, callback, doc, attributes) {
-  if (!isActivityAllowed(LOAD_EXTERNAL_SCRIPT, activityParams(MODULE_TYPE_PREBID, 'adLoader'))) {
-    logError('cannot load external script as it\'s disabled by activity controls');
+  if (!isActivityAllowed(LOAD_EXTERNAL_SCRIPT, activityParams(MODULE_TYPE_PREBID, moduleCode))) {
     return;
   }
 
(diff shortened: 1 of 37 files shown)


src/adloader.js
```diff
@@ -50,6 +54,11 @@ const _approvedLoadExternalJSList = [
  * @param {object} attributes an object of attributes to be added to the script with setAttribute by [key] and [value]; Only the attributes passed in the first request of a url will be added.
  */
 export function loadExternalScript(url, moduleCode, callback, doc, attributes) {
+  if (!isActivityAllowed(LOAD_EXTERNAL_SCRIPT, activityParams(MODULE_TYPE_PREBID, 'adLoader'))) {
```
dgirardi, line 57: I think to be useful the module should be the one attempting to load the script (rather than prebid/adloader). Otherwise I can only disable all scripts or none - the activity parameters are always the same. It does require some refactoring to get the caller's module type.

src/adloader.js
```diff
@@ -50,6 +54,11 @@ const _approvedLoadExternalJSList = [
...
+    logError('cannot load external script as it\'s disabled by activity controls');
```
dgirardi, line 58: `isActivityAllowed` logs a warning when denied, so this is unnecessary IMO.

test/spec/adloader_spec.js
```diff
@@ -92,4 +94,11 @@ describe('adLoader', function () {
     expect(script.z).to.equal('A');
     expect(script.y).to.equal(2);
   });
+
+  it('should disable loading external script for activity rule set', function () {
+    const unregisterRule = registerActivityControl(LOAD_EXTERNAL_SCRIPT, 'loadExternalScript config', () => ({allow: false}));
+    adLoader.loadExternalScript('someURL2', 'debugging');
+    expect(utilsLogErrorStub.called).to.be.true;
+    unregisterRule();
```
dgirardi, line 102: Test teardown is better done in a teardown hook (`after`/`afterEach`), or in a `finally` block. This is fine if the test passes, but if at some point some code changes makes it fail, the teardown would not run and can potentially affect other tests, which makes it more difficult to find the failure.

dgirardi (approved):

osazos:
The public documentation should be updated accordingly.

https://docs.prebid.org/dev-docs/activity-controls.html#activities

patmmccann (approved):

PR #12214, opened 2024-09-09
Symetri RTD module: OnBidResponse method added

mp4symitri:
## Type of change

- [X] Feature

## Description of change
**OnBidResponse** listener added. This listener listens to bid responses and if it finds a response with **dealId** matching to the one stored in the local storage by the RTD adapter, it will generate a pixel and append it to **ad** such that it gets triggered upon rendering of the ad.
Also added support for already hashed identifiers or composite identifiers (array of identifiers)

Here is the PR to prebid.github.io repository with documentation update:
https://github.com/prebid/prebid.github.io/pull/5590

### DIFF
--- a/modules/symitriDapRtdProvider.js
+++ b/modules/symitriDapRtdProvider.js
@@ -151,13 +151,30 @@ export function createRtdProvider(moduleName, moduleCode, headerPrefix) {
     return true;
   }
 
+  function onBidResponse(bidResponse, config, userConsent) {
+    if (bidResponse.dealId && typeof (bidResponse.dealId) != typeof (undefined)) {
+      let membership = dapUtils.dapGetMembershipFromLocalStorage(); // Get Membership details from Local Storage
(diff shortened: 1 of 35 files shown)


modules/symitriDapRtdProvider.js
```diff
@@ -151,13 +151,30 @@ export function createRtdProvider(moduleName, moduleCode, headerPrefix) {
     return true;
   }
 
+  function onBidResponse(bidResponse, config, userConsent) {
+    if (bidResponse.dealId && typeof (bidResponse.dealId) != typeof (undefined)) {
+      let membership = dapUtils.dapGetMembershipFromLocalStorage(); // Get Membership details from Local Storage
+      let deals = membership.deals; // Get list of Deals the user is mapped to
```
patmmccann, line 157: you may want to also check for deals that match a particular pattern

patmmccann (approved):

PR #12203, opened 2024-09-04
Digital Matter Bid Adapter: Refactor Adapter

digital-matter:
## Type of change

- [ ] Bugfix
- [ ] Feature
- [ ] New bidder adapter
- [x] Updated bidder adapter
- [ ] Code style update (formatting, local variables)
- [ ] Refactoring (no functional changes, no api changes)
- [x] Build related changes
- [ ] CI related changes

- [ ] Does this change affect user-facing APIs or examples documented on http://prebid.org?
- [ ] Other

## Description of change
We have decided to change the platform we’re using for developing the Prebid adapter thus we’re changing the code fully with this PR. We would appreciate it if this PR could be accepted as soon as possible as we cannot use the current version of the submitted adapter.

<!-- For new bidder adapters, please provide the following
- prebid@digitalmatter.ai
- test parameters for validating bids:
```
{
  bidder: 'digitalMatter',
  params: {
    accountId': '1_demo_1',
    siteId': '1-demo-1',
  }
}
```

Issue #12001, opened 2024-07-18
Prebid 10: delete old modules


patmmccann:
Running list of candidates:

- [x] https://github.com/prebid/Prebid.js/blob/master/modules/akamaiDapRtdProvider.js
- [x] global sun
- [x] saambaa assuming #11992 merges in
- [x] emx as xandr alias
- [ ] id import (every year we consider it) 
- [ ] Apn transform temporary module 
- [x] telaria bid adapter
- [x] former kueez https://github.com/prebid/Prebid.js/blob/master/modules/kueezBidAdapter.js
- [x] adtrgtme appears to be a yahoo alias and needs fixing
- [x] one of either gamoshi and cleanmedia needs removing
- [x] oxxion is the new name of bidwatch?
- [x] https://github.com/prebid/Prebid.js/blob/master/modules/adsinteractiveBidAdapter.js appears abandoned in favor of https://github.com/prebid/Prebid.js/blob/master/modules/ads_interactiveBidAdapter.js
- [x] smarthub is now attekmi https://github.com/prebid/prebid.github.io/pull/5699#pullrequestreview-2436281768
- [x] konduit RTD - they're OOB.
- [ ] one of either shinez or shinezrtb is no longer in use
- [x] verizonMediaIdSystem.js

patmmccann:
Ocm should now be an alias of limelight, not adtelligent

patmmccann:
some more: 

Potential “Zombie” Modules (Very Old & No Recent Activity)
ablidaBidAdapter_spec.js

Last commit: ~5 years ago
The Ablida adapter/test code has not changed since the initial addition. No further commits for Prebid 4.x, 5.x, or 9.0 migrations, etc.
my6senseBidAdapter_spec.js

Last commit: ~5 years ago
Zero housekeeping commits or Prebid-version updates. Strong candidate for inactivity.
padsquadBidAdapter_spec.js

Last commit: ~5 years ago
No sign of any updates since around 2018–2019.
temedyaBidAdapter_spec.js

Last commit: ~5 years ago
Similarly quiet since the initial commit.
reconciliationRtdProvider_spec.js

Last commit: ~5 years ago
An older RTD module from the early Real-Time Data experiments. No code touches in half a decade.
vubleAnalyticsAdapter_spec.js

Last commit: ~6 years ago
One of the oldest test files still in the repo. If Vuble is active, they have not updated or tested their analytics code in a long time.
serverbidServerBidAdapter_spec.js

[...]

PR #12215, opened 2024-09-09
add Dichange Bid Adapter

Chucky-choo:
## Type of change
- [x] New bidder adapter

## Description of change
- contact email of the adapter’s maintainer: di-change@digitalmatter.ai
```
[
    {
        code: 'test-banner',
        mediaTypes: {
            banner: {
                sizes: [[300, 250]],
            }
        },
        bids: [
            {
                bidder: 'dichange',
                params: {
                    env: 'dichange',
                    pid: '40',
                    ext: {}
                }
            }
        ]
    },
    {
        code: 'test-video',
        sizes: [ [ 640, 480 ] ],
        mediaTypes: {
            video: {
                playerSize: [640, 480],
                context: 'instream',
                skipppable: true
            }
        },
        bids: [{
            bidder: 'dichange',
            params: {
                env: 'dichange',
                pid: '40',
                ext: {}
            }
        }]
    }
]
```
## Other information
Docs PR: https://github.com/prebid/prebid.github.io/pull/5591


bretg:
Spoke about this with @ChrisHuie .  There's an open question about what happens when PBJS sees the "bidderA" biddercode and finds both a base adapter and one or more aliases. We should document what it does: break? base adapter? first base adapter that has bidderA as an alias?

In any case, I would propose that PBJS should (A) not break, (B) choose the base adapter by default, and (C) allow the publisher to override this behavior with pbjs.aliasBidder().

That would make this kind of operation not breaking in general. Here are a couple of scenarios:

### Aliased biddercode converted to a standalone base adapter

This is what @Chucky-choo is trying to do here, though I disagree with PR #12203 -- removing the alias does create a breaking change.

Time 1

1. The publisher's adunits refer to `dichange` as a bidder
1. The publisher has built their PBJS package with digitalmatterBidAdapter.js where dichange is an alias. Life is good.

Note that if digitalmatterBidAdapter.js stops listing dichange as an alias, the publisher would be forced to change their build, causing a breaking change. 

Time 2

1. The publisher's adunits still refer to `dichange` as a bidder
1. The publisher has built their PBJS package with both dichangeBidAdapter.js AND digitalmatterBidAdapter.js where dichange is an alias. Prebid.js picks one of them (preferably dichange). digitalmatterBidAdapter might be wasted bytes in the package, but otherwise, life is ok. Yes, the request is now being processed by a different adapter, but that's fine because clealry the publisher started including the new adapter explicitly in their build. They must know what's going on.

Time 3

1. The publisher's adunits still refer to `dichange` as a bidder
1. The publisher has built their PBJS package with just dichangeBidAdapter.js. Life is good.

### Formerly standalone biddercode converted to an alias in a new base adapter

This isn't what's happening in this PR, but for sake of completeness and because Magnite is going to do this next year, thought it worth working through.

Time 1

1. The publisher's adunits refer to `rubicon` as a bidder.
2. The publisher has built their PBJS package with just rubiconBidAdapter.js. Life is good.

Time 2

1. The publisher's adunits still refer to `rubicon` as a bidder.
2. The publisher has built their PBJS package with BOTH rubiconBidAdapter.js and magniteBidAdapter.js which also aliases rubicon. PBJS chooses the base adapter. Life is ok - the Magnite adapter isn't being utilized at this point.

Time 3

1. The publisher's adunits still refer to `rubicon` as a bidder.
2. The publisher has built their PBJS package with BOTH rubiconBidAdapter.js and magniteBidAdapter.js which also aliases rubicon.
3. The publisher is ready to try the new Magnite-based rubicon, so they call pbjs.aliasBidder('magnite', 'rubicon')
4. PBJS sees this explicit signal and invokes 'magnite' for adunits that refer to rubicon.
5. Life is good.

Time 4

1. The publisher's adunits still refer to `rubicon` as a bidder.
2. The publisher has built their PBJS package with only magniteBidAdapter.js which aliases rubicon.
3. PBJS only sees rubicon as an alias of magnite, so life is good.

Chucky-choo:
@bretg Thank you for your attention to this matter. After internal discussions, we concluded that the most transparent approach is to keep the PBJS package with digitalmatterBidAdapter.js, where ‘dichange’ remains as an alias. Therefore, we’re withdrawing this PR, and in PR #12203, we won’t be removing the alias.
We hope this clarifies our plans and doesn’t cause any break. Feel free to reach out with any questions.

patmmccann:
@Chucky-choo please let us know on #12001 if you want to do a change in the next version

PR #12244, opened 2024-09-16
NextMillennium Bid Adapter : sending a request with several imp objects

mhlm:
## Type of change

## Description of change

Previously, the adapter sent its own ad request for each ad slot. The adapter now sends one ad request for all ad slots.


Rothalack (approved):

PR #12257, opened 2024-09-19
Preciso Bid Adapter : Native included

NikhilGopalChennissery:
## Type of change

- [ ] Bugfix
- [ ] Feature
- [ ] New bidder adapter  
- [ ] Updated bidder adapter   
- [ ] Build related changes
- [ ] CI related changes

- [ ] Other


bretg:
docs PR https://github.com/prebid/prebid.github.io/pull/5582

modules/precisoBidAdapter.js
```diff
@@ -37,17 +36,16 @@ export const spec = {
...
   },
   buildRequests: buildRequests(endpoint),
-  interpretResponse,
+  interpretResponse: buildBidResponse,
   onBidWon,
   getUserSyncs: (syncOptions, serverResponses, gdprConsent, uspConsent) => {
     const isSpec = syncOptions.spec;
-    if (!Object.is(isSpec, true)) {
+    if ((!Object.is(isSpec, true)) && (storage.localStorageIsEnabled())) {
       let syncId = storage.getCookie('_sharedid');
```
dgirardi, line 44: this only works if the publishers sets up sharedId to use a cookie with this name; it also bypasses consent/activity control settings (granted it's an unlikely scenario, but it's possible that a publisher may decide to not make sharedId available to Preciso in particular).

You can get sharedId (when available) as `bidRequest.userId.pubcid`, or as an EID with source 'pubcid.org' in `userIdAsEids`.

NikhilGopalChennissery: Hi @dgirardi 
we are passing the sharedId to sync and prevent the unwanted bidrequests.  we are passing this sharedId if it is available only, otherwise we send it as NA. even if the sharedid is there or not getuserSyncs will happen. if the sharedid is present, it will pass that otherwise it consider as NA.

Thank you!

libraries/precisoUtils/bidUtils.js
```diff
@@ -96,8 +77,161 @@ export function onBidWon(bid) {
...
+            result.sponsoredBy = asset.data.value;
+            break;
+          case OPENRTB.NATIVE.ASSET_ID.CTA:
+            result.cta = asset.data.value;
+            break;
+        }
+      });
+      return result;
+    } else {
+      return
```
ncolletti, line 210: A native bid w/o assets should be filtered out of your returned bid responses, could you add this?

libraries/precisoUtils/bidUtils.js
```diff
@@ -96,8 +77,161 @@ export function onBidWon(bid) {
...
+      const result = {
+        clickUrl: encodeURI(native.link.url),
+        impressionTrackers: native.eventtrackers[0].url,
+        clickTrackers: native.link.clicktrackers[0]
+
+      };
+      native.assets.forEach(asset => {
+        switch (asset.id) {
+          case OPENRTB.NATIVE.ASSET_ID.TITLE:
+            result.title = asset.title.text;
```
ncolletti, line 181: please use `deepAccess` to set these properties

modules/precisoBidAdapter.js
ncolletti: If publishers decide to set their sharedId to a different browser storage key or location, your bidder will not see that and provide bids. You can avoid those scenarios by looking in the locations mentioned above.

NikhilGopalChennissery: Hi @ncolletti  Can  I get bidRequest.userId.pubcid in getUserSyncs() ?

libraries/precisoUtils/bidUtils.js
```diff
@@ -96,8 +77,159 @@ export function onBidWon(bid) {
...
+export const OPENRTB = {
+  NATIVE: {
+    IMAGE_TYPE: {
+      ICON: 1,
+      MAIN: 3,
+    },
+    ASSET_ID: {
+      TITLE: 1,
+      IMAGE: 3,
+      ICON: 2,
```
patmmccann, line 224: Stop reordering things to avoid duplication  checks. You pr is declined.

PR #12271, opened 2024-09-25
appnexus bid adapter - initial support for image userSync

jsnellbaker:
## Type of change
- [x] Updated bidder adapter   

## Description of change
This PR adds initial support for the appnexus Bid Adapter to support image based userSyncs.  It is intended for now to use a static list of certain endpoints; this will change over time.

### DIFF
--- a/test/spec/modules/anPspParamsConverter_spec.js
+++ b/test/spec/modules/anPspParamsConverter_spec.js
@@ -51,7 +51,6 @@ describe('anPspParamsConverter', function () {
 
     const testBidderRequests = deepClone(bidderRequests);
 
-    debugger; //eslint-disable-line
     convertAnParams(function () {
       didHookRun = true;
     }, testBidderRequests);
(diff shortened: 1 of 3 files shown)


test/spec/modules/anPspParamsConverter_spec.js
```diff
@@ -51,7 +51,6 @@ describe('anPspParamsConverter', function () {
 
     const testBidderRequests = deepClone(bidderRequests);
 
-    debugger; //eslint-disable-line
```
patmmccann, line 54: whopps

patmmccann: ty

patmmccann (approved):

test/spec/modules/anPspParamsConverter_spec.js
jsnellbaker: yeah - sorry about that.  Finally caught it.

